Issue #1374, opened 2015-05-23
Port of AGM's night vision brightness adjustment is worse than CSE's implementation


ShocknAw3:
ACE3 Version: 3.0.1

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
none

**Description:**
Adjusting the brightness of the night vision in its current form is useless. Every setting that deviates from zero gives you a worse image. CSE pulled this off much better. Is it possible to port over CSE's solution instead, seeing as this is a merger between those mods?

**Steps to reproduce:**
1. Equip night vision
2. Adjust brightness using alt + page up and alt + page down

**Where did the issue occur?**
Tested in editor preview.

**RPT log file:**
n/a

esteldunedain:
@Glowbal, we were never satisfied with agm's solution, but we couldn't get it to work better; how did CSE did it?

PabstMirror:
Someone was telling me we should look into using `setAperture`, but I haven't had much luck getting any good effects with it yet.

esteldunedain:
I tried that too once without success

Glowbal:
CSE:
```sqf
_ppEffect_NVGAdjustBrightness ppEffectAdjust [1, _newBrightness, 0, [0, 0, 0, 0], [0, 0, 0, 1], [0, 0, 0, 1]];
```
ACE:
```sqf
GVAR(ppEffectNVGBrightness) ppEffectAdjust [1, 1, _brightness / 4, [0, 0, 0, 0], [0, 0, 0, 1], [0, 0, 0, 1]];
```

Glowbal:
I first used `setAperture` but I only got horrible results with it.

esteldunedain:
Well, simple enough. Definitely worth trying. If someone doesn, please make some screenshots for comparison.

PR #1429, opened 2015-05-29
NightVision adjustment - use CSE's gamma

PabstMirror:
#1374

Shots taken at -5, 0 and +5 adjustment:
Current:
old
New:
new

### DIFF
--- a/addons/nightvision/functions/fnc_changeNVGBrightness.sqf
+++ b/addons/nightvision/functions/fnc_changeNVGBrightness.sqf
@@ -22,12 +22,12 @@ PARAMS_2(_player,_changeInBrightness);
 
 _brightness = _player getVariable [QGVAR(NVGBrightness), 0];
 
-_brightness = ((round (10 * _brightness + _changeInBrightness) / 10) min 1) max -1;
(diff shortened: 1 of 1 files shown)


jonpas:
That's pretty nice! :+1:

alganthe:
For 3 lines of code that's a pretty big improvement, this looks much nicer than before.

esteldunedain:
:+1: It's definitely better

jaynus:
Thats super easy to see :/ Was this from a ticket saying our NVG's were too hard to see with at the moment?

PabstMirror:
from issue #1374

From what I understand we where just adding and subtracting a static offset.  
Which doesn't really help increase visibility at all.

CSE's way was to adjust contrast (some place called it gamma)

jaynus:
well IRL increasing/decreasing gamma is supposed to lower the image quality - your processing and mofiying the image and as a result, it looks worse. That's literally the way it is supposed to be and how it is in real life.

The NVG's in arma have always been way too horribly photorealistic.

Legolasindar:
+100

jaynus:
:thumbsup:

PR #1430, opened 2015-05-29
BFT_Devices Updated

nsgundy:
Added on-device notification system
Added interface getters (macros)
Hooked into bft_updateDeviceOwner event
Registered device map controls with ace_bft_drawing_doBFTDraw
Added Gundy to authors


alganthe:
Does this rely on an other PR ?
Because bft_device is definitely not on master yet.

Issue #1273, opened 2015-05-17
Server is forcing server language for settings values


gienkov:
ACE3 Version: 3.0.0

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *None*

**Description:**
*So it seems like even you have set your language to for example polish, while playing on the server you will see setting values in server language (in my example in english)*
http://images.akamai.steamusercontent.com/ugc/37497482120665768/9C112D557221CF612F1B4A041358358FBC42F5F3/
*As you can see on image above, only things that are translated to polish are values that are BOOL, and those that use game strings like:*
```
    class GVAR(textSize) {
        value = 2;
        typeName = "SCALAR";
        isClientSettable = 1;
        displayName = "$STR_ACE_Interact_textSize";
        values[] = {"$str_very_small", "$str_small", "$str_medium", "$str_large", "$str_very_large"};
    };    
```
This example is best:
```
    class GVAR(shadowSetting) {
        value = 2;
        typeName = "SCALAR";
        isClientSettable = 1;
        displayName = "$STR_ACE_Interact_shadowSetting";
        description = "$STR_ACE_Interact_shadowSettingDescription";
        values[] = {"$STR_A3_OPTIONS_DISABLED", "$STR_A3_OPTIONS_ENABLED", "$STR_ACE_Interact_shadowOutline"};
    };
```
***"$STR_A3_OPTIONS_DISABLED"*** and ***"$STR_A3_OPTIONS_ENABLED"*** are game strings that are being translated properly, but last option which is ***"$STR_ACE_Interact_shadowOutline"*** is using server language anyway.

**Steps to reproduce:**
* *Set game language to anything but english.*
* *Join server which has language set to english.*
* *Open settings menu and you will see that part of options (which are stringtabled) are in english, not in language you set in your game.

**Where did the issue occur?**
*Multiplayer only*

**RPT log file:**
*Not needed*

Glowbal:
We fixed this earlier but it's probably not reliable. Should be part of the  Reliable settings framework #1214.

PR #1214, opened 2015-05-14
Reliable settings framework

esteldunedain:
Apparently we have problems with settings not being reliably inited before client post init handlers are run.

This creates a `SettingsInitialized` event once every setting has been received from the server, profile,etc. This event is guaranteed to happen **AFTER** all postInits are ready. Hence it's safe to add listener for it on the postInits of the different modules to do many first time inits that require settings.

PR #1432, opened 2015-05-30
Fix ace_settings

PabstMirror:
Problem with #1214 
this should also solve issue #1273

Clients don't localize and they don't load settings from profile.

I think the problem is we never raise the `ServerSettingsReceived` on clients and GVAR(SettingsInitialized) never changes to true, so that PFEH at bottom of postInit runs forever.

Server language english
Client screenshot in spanish:

untitled

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -254,6 +254,7 @@ GVAR(commonPostInited) = true;
 [{
     // If post inits are not ready then wait
     if !(SLX_XEH_MACHINE select 8) exitWith {};
+
     // If settings are not initialized then wait
     if (isNil QGVAR(settings)) exitWith {
         diag_log text format["[ACE] Waiting on settings from server"];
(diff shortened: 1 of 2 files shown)


esteldunedain:
> I think the problem is we never raise the ServerSettingsReceived

That's because this line is wrong:

https://github.com/acemod/ACE3/blob/master/addons/common/XEH_preInit.sqf#L313

should be a globalEvent

PabstMirror:
What about JIPs or if the server triggers the event before clients load.

esteldunedain:
> What about JIPs or if the server triggers the event before clients load.

That's a good point. TBH the localEvent thing is a typo, but your implementation covers those topics better. Let me think it a minute

esteldunedain:
Right, I like you solution (whi is actually THE solution). Hold on a minute more while I fix my bloody mess

esteldunedain:
@PabstMirror, please check that last commit in case I'm derping again. But now I see it: the "ServerSettingsReceived" event was redundant, as it's much better to check isNil QGVAR(settings) as you proposed, which is JIP compatible.

addons/common/XEH_postInit.sqf
```diff
@@ -254,12 +254,25 @@ GVAR(commonPostInited) = true;
 [{
     // If post inits are not ready then wait
     if !(SLX_XEH_MACHINE select 8) exitWith {};
+
     // If settings are not initialized then wait
-    if !(GVAR(SettingsInitialized)) exitWith {};
+    if (isNil QGVAR(settings)) exitWith {
+        diag_log text format["[ACE] Waiting on settings from server"];
```
Glowbal: Do we want to print this to the rtp? What if the settings arrive late? Then we have spammed the rtp with quite a few messages. There is no need to print this multiple times I think.

esteldunedain: Well, I noticed that, but I decided leaving that spam on the first version should be good, to be able to easily check how long the settings are delaying. I propose to remove it on the next update.

PR #1433, opened 2015-05-30
Delay initialization of modules until settings are properly setup

esteldunedain:
As part of the reliable settings framework, we need to ensure that settings are not read from modules before the settings lists have arrived from the server.

So this PR does the following:
- Change all modules on ACE using settings to type ACE_Module, so we can guarantee that they wait for settings before initing.
- Command the ACE_Modules to init only once the settings from the server have arrived.
- Make ACE_Common require ACE_Modules, so we can use that functionality everywhere.


esteldunedain:
@Glowbal, @PabstMirror, please review if this makes as much sense as it seems.

Glowbal:
:+1:

PR #1442, opened 2015-05-30
BFT_Devices Updated

nsgundy:
Addition of bft_deviceClosed event
Configured devices for NATO, CSAT and AAF
Configured vanilla helicopters and planes for BFT_Devices
Interface is now 100% chosen based on available personal / vehicle device


addons/bft/CfgVehicles.hpp
```diff
@@ -1,9 +1,9 @@
...
-class CfgVehicles {
+/*class CfgVehicles {
     class MRAP_01_base_F;
     class B_MRAP_01_F: MRAP_01_base_F {
         // what devices are available?
         GVAR(vehicleDevices)[] = {"FBCB2VehDevice"};
     };
-};
\ No newline at end of file
+};/*
```
Glowbal: This should probably be `*/`

PR #1455, opened 2015-05-31
Updated all airFriction values according to: airFrictionAnalysis.txt

ulteq:
All ammo classes are now within the following error margins for the typical engagement ranges and muzzle velocities of the cartridge:

* Maximum velocity difference: 33 m/s
* Maximum drop difference: 15 cm
* Maximum time of flight difference: 34 ms

Raw analysis output: https://github.com/acemod/ACE3/blob/airFrictionUpdate/extras/airFrictionAnalysis.txt

Example: m855_dropdifference


esteldunedain:
Perfect! This is the absolutely best way of getting correcty airFriction values

PR #1456, opened 2015-05-31
Overcast > 0.7 no longer implies rain:

ulteq:
* The higher the overcast value the higher the chance for rain
* But you can end up having no rain with an overcast value of 1.0


esteldunedain:
:+1: But documentation needs to be updated

BullHorn7:
Ill update the doc tonight

PR #1458, opened 2015-05-31
Fixes random teleport when dragging/carrying inside of buildings.

ulteq:


esteldunedain:
:+1: for merge. As we discussed many times getPos is ambiguose on multi-story builidings and it's therefore pretty useless (except for determining the distance between an object and it's inmediate "floor")

PR #1460, opened 2015-05-31
HitReactions - Don't do when mounted, add setting

PabstMirror:
Disables hit reactions when mounted.

Right now they can happen from FFV seats.
But it looks silly and never resets to normal seated position.
image

Also adds setting for minimum damage to trigger, and can disable by setting to -1.


Glowbal:
:+1:

ulteq:
`GVAR(minDamageToTrigger)` :+1:

esteldunedain:
:+1: I'd say merge it

PR #1471, opened 2015-06-01
Body bag cleanup:

ulteq:
* Increased the treatment time to match the animation duration
* Fixed the body bag alignment with the dead unit


esteldunedain:
:+1:

PR #1476, opened 2015-06-02
Ported the ACE2 MX-2A thermal imaging device

ulteq:


Glowbal:
Do we want to keep this in an MX2A pbo or combine this with a few other bino's that we could port from ACE2?

alganthe:
why not put it into thermals ?

ulteq:
> why not put it into thermals ?
Thermals is not about items/weapon. It's about changing how thermal imaging looks in general.

addons/mx2a/stringtable.xml
```diff
@@ -0,0 +1,27 @@
...
+            <Czech>MX-2A</Czech>
+            <Russian>MX-2A</Russian>
+            <Italian>MX-2A</Italian>
+            <Polish>MX-2A</Polish>
+            <French>MX-2A</French>
+            <Hungarian>MX-2A</Hungarian>
+        </Key>
+        <Key ID="STR_ACE_MX2A_Description">
+            <English>Thermal Scope</English>
+            <German>Thermal Scope</German>
```
esteldunedain: The non-translated entries should be deleted (Spanish in particular, but afaik also Italian and French)

esteldunedain:
:+1: Remember to always ask for confirmation before porting p3d files from ACE2, as we can't release some of them under GPL.

PR #1477, opened 2015-06-02
flares modified

alganthe:
- all UGL flares have a 60s lifespan now
- handheld signals and 20mm flares have smaller flare sizes

flares were extinguishing too fast before to have a real use, they should be more useful now.

### DIFF
--- a/addons/grenades/CfgAmmo.hpp
+++ b/addons/grenades/CfgAmmo.hpp
@@ -3,6 +3,7 @@ class CfgAmmo {
     class FlareBase: FlareCore {
         intensity = 20000;
         flareSize = 12;
+        timeToLive = 60;
     };
     class F_40mm_White: FlareBase {
         intensity = 40000;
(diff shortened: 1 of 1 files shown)


esteldunedain:
:+1:

addons/grenades/CfgAmmo.hpp
```diff
@@ -3,6 +3,7 @@ class CfgAmmo {
     class FlareBase: FlareCore {
         intensity = 20000;
         flareSize = 12;
+        timeToLive = 60;
```
bux578: curios: based on what data?

alganthe:
wikipedia http://en.wikipedia.org/wiki/Shell_%28projectile%29#Illumination

most notably this line ``` Typically illumination flares burn for about 60 seconds. These are also known as starshell or star shell.```

Grtwatkins:
It appears that you mistakenly looked at howitzer flare shells. I found this for 40mm shells that say they burn for around 40 seconds. I could also be mistaken though

Wikipedia: https://en.wikipedia.org/wiki/United_States_40_mm_grenades#Signalling_and_illumination_cartridges

PR #1492, opened 2015-06-03
Disabled pain effect while unconscious

GitHawk:
Limited _strength of painEffect to 0.1 so a player does not have to watch a flashing screen while he can't do anything about it.
Otherwise we need to include a serious epilepsy warning! See http://en.wikipedia.org/wiki/Photosensitive_epilepsy

### DIFF
--- a/addons/medical/XEH_postInit.sqf
+++ b/addons/medical/XEH_postInit.sqf
@@ -169,6 +169,9 @@ GVAR(lastHeartBeatSound) = ACE_time;
 
         // Pain effect
         _strength = (_pain - (ACE_player getvariable [QGVAR(painSuppress), 0])) max 0;
+        if (ACE_player getVariable ["ACE_isUnconscious", false]) then {
+            _strength = _strength min 0.1;
+        };
         _strength = _strength * (ACE_player getVariable [QGVAR(painCoefficient), GVAR(painCoefficient)]);
         if (GVAR(painEffectType) == 1) then {
             GVAR(effectPainCC) ppEffectEnable false;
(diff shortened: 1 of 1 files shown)


Glowbal:
It might be better to not play any pain effects while unconscious.

jonpas:
> It might be better to not play any pain effects while unconscious.

It does sound way better, you can't see anything anyways... (practically).

ulteq:
> It might be better to not play any pain effects while unconscious.

Yes, please.

GitHawk:
Ok, will change it.

GitHawk:
@Glowbal Now the effect is disabled.

ulteq:
> The code alignment in diffview looks wierd.
You should use 4 spaces instead of tabs.

GitHawk:
> You should use 4 spaces instead of tabs.

I am using spaces but the diffview matches the wrong lines. Old line 167 is now at 174

esteldunedain:
> I am using spaces but the diffview matches the wrong lines. Old line 167 is now at 174

Yeah, don't worry about it. This is just github being weird, and happens often. Unfortunately it makes code much harder to review, but tough luck.

Glowbal:
I will take a look at this today.

GitHawk:
@Glowbal Can you merge, so I can work on other stuff?

jonpas:
@GitHawk that's the point of GitHub, make branches and you can work on infinite stuff without waiting for merges. In fact, you should really make all your pull requests on a separate (non-master) branch, do that for each feature/fix/whatever you are working on and pull request from that.

Glowbal:
Thank you @GitHawk

PR #1493, opened 2015-06-03
Initial draft of the ACE2 tactical ladder port.

ulteq:
Mandatory:
- [x] Rewrite the ladder interaction routine
- [x] Get rid of the regular action menu entries
- [x] Fix the backpack position on the players back

Optional:
- [ ] Make the ladder PhysX compatible


Legolasindar:
Please not repeat original stringtable for all languages columns, because it does not know that you have to translate this addon because it does not appear in the Tabler, thanks ;)

bux578:
^ Yep, good point.

PR #1502, opened 2015-06-04
CUP Launchers Backblast + Disposable

Winter259:
Used the same values that I had used for the AGM CUP launcher compatibility. Tested & working in game, would appreciate a sanity check.


esteldunedain:
Excellent

PR #1505, opened 2015-06-04
strobe light more bearable

alganthe:
changed the intensity of the strobe light, it's still VERY visible on a
very long distance, (around 1.6k maybe more)

Before  http://steamcommunity.com/sharedfiles/filedetails/?id=455041609

after http://steamcommunity.com/sharedfiles/filedetails/?id=455048548

RL :

### DIFF
--- a/addons/attach/CfgVehicles.hpp
+++ b/addons/attach/CfgVehicles.hpp
@@ -87,9 +87,9 @@ class CfgVehicles {
         class NVGMarker {
             diffuse[] = {0.006, 0.006, 0.006, 1};
             ambient[] = {0.005, 0.005, 0.005, 1};
-            brightness = 0.2;
+            brightness = 0.1;
             name = "pozicni blik";
-            drawLightSize = 0.2;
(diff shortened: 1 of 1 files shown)


esteldunedain:
I'm pretty sure those things in the picture are NOT strobes.

Btw, how the current strobe looks is VERY dependant on light conditions (moon, clouds, etc). Be sure to test the modification under various conditions to ensure it is still visible from the AIR (which is the whole point of this)

alganthe:
I could see this from 1.6k away, I'll do some tests in forest with a huge tempest going on.

esteldunedain:
> I could see this from 1.6k away, I'll do some tests in forest with a huge tempest going on.

I think visibility goes way down when there's some ambient light, like bright moon conditions. Check those.

alganthe:
Tested with full moon, there's barely any difference between the previous and the new brightness, for some reason at a distance the effect is the same

esteldunedain:
OK, thanks

PR #1511, opened 2015-06-04
Added Macro Guard to string macros

jonpas:
Adds a guard for xSTRING macros, as discussed with @Glowbal and @ViperMaul


Glowbal:
:+1:

PR #1521, opened 2015-06-05
Minor edits to Esc menu buttons while unconscious

AgentRev:
* Abort functionality transferred to vanilla Abort button (idc 104), instead of overriding "Save" button (idc 103)
* Respawn button now only shows in MP, all it does is kill the player so it has no purpose in SP
* Added confirmation prompts for abort & respawn, like vanilla has
* Custom text edited to use vanilla localized strings

The reasons why I transferred Abort stuff to idc 104 are because it's the vanilla game's actual Abort button, so it makes sense for it to remain like so, and also because some vanilla missions using non-ACE revive systems disable it while being unconscious to prevent combat logging. So, it makes it easier to port those missions to ACE while maintaining cross-compatibility with vanilla. The other things are just minor improvements to user experience.

### DIFF
--- a/addons/common/functions/fnc_disableUserInput.sqf
+++ b/addons/common/functions/fnc_disableUserInput.sqf
@@ -37,11 +37,12 @@ if (_state) then {
     _dlg = uiNamespace getVariable QGVAR(dlgDisableMouse);
 
     _dlg displayAddEventHandler ["KeyDown", {
-        private ["_key", "_dlg", "_ctrl", "_config", "_acc", "_index"];
+        private ["_key", "_dlgClass", "_dlg", "_abortALiVE", "_ctrl", "_config", "_acc", "_index"];
         _key = _this select 1;
 
         if (_key == 1 && {alive player}) then {
-            createDialog (["RscDisplayInterrupt", "RscDisplayMPInterrupt"] select isMultiplayer);
+            _dlgClass = ["RscDisplayInterrupt", "RscDisplayMPInterrupt"] select isMultiplayer;
+            createDialog _dlgClass;
 
             disableSerialization;
             _dlg = finddisplay 49;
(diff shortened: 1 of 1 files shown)


bux578:
Please test this in combination with ALiVE (SP/MP). IIRC there were some issues and why @commy2 used this approach.

AgentRev:
For some reason, it appears the coders of ALiVE have decided to change the IDC of the Abort button to 199 instead of the default 104, while keeping all the other ones in line with vanilla... I will figure something out.

AgentRev:
@bux578 Alright, instead of having a workaround which breaks stuff for other mods, let's make it a proper exception for the ALiVE abort button itself, since it's the odd one out.

bux578:
rgr, thanks

esteldunedain:
I like this approach.

AgentRev:
Maybe somebody else than commy2 could look at this? He hasn't been active in 2 months.

commy2:
Is this still a thing? I do have a respawn button, even without any changes while unconscious with ACE and alive

AgentRev:
This pull request was not really about the Respawn button, it's about the Abort button. My mission tries to block the abort button (idc 104) when the player is unconscious. Since you have transferred the Abort button to idc 103 because of conflicts with Alive, my anti-abort script did not work, so I decided to reorganize `fnc_disableUserInput.sqf` in a way that makes more sense, so that the original idc 104 would remain in use when Alive is not present. Also fixed localization issues.

SilentSpike:
Related to this PR, but maybe we should separate the spoofing of the interrupt menu from this function so that it can be used elsewhere. I have some similar code in spectator to initialise the display when escape is pressed, but don't require the rest of the disableUserInput code.

If we had a function just for this in common, then all compatibility and other functionality would be handled only once.

commy2:
Good idea.
Can someone than me write that function?

tupolov:
ALiVE actually deletes the original abort button and implements its own. This is necessary for us to catch any player/admin trying to exit.

From our perspective, don't mind if people want to disable the ability for a player to exit due to being unconscious. It might be annoying for an admin that wants to exit the mission.

If there is another mod-friendly way of doing this, let us know.

Glowbal:
It's been over a year since this was opened. As it has merge conflicts, please create a new PR or submit a bug report if this is still a ticket.

Thanks @AgentRev

PR #1522, opened 2015-06-06
Macroize all stringtables

PabstMirror:


jonpas:
Thanks for that, I forgot to run regex again after finishing...

PR #1525, opened 2015-06-06
ACE2 HuntIR port

ulteq:


gienkov:
:+1:

esteldunedain:
Excellent, we've meaning to port this since forever!

I guess things could maybe improved a bit for arma 3 making the HuntIR a kind of UAV, so it shares the terminal, etc. But it's not a big deal, and it's better to have something working first.

gj!

ulteq:
I like the old interface a lot.

esteldunedain:
> I like the old interface a lot.

Yeah, you are probably right. The current interface is most likely closer to the real thing for the HuntIR.

addons/huntir/CfgWeapons.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+        class ItemInfo: InventoryItem_Base_F {
+            mass = 20;
+        };
+    };
+    
+    class GrenadeLauncher;
+    class UGL_F: GrenadeLauncher {
+        //magazines[] += {"ACE_HuntIR_M203"};
+        magazines[] = {"1Rnd_HE_Grenade_shell","UGL_FlareWhite_F","UGL_FlareGreen_F","UGL_FlareRed_F","UGL_FlareYellow_F","UGL_FlareCIR_F","1Rnd_Smoke_Grenade_shell","1Rnd_SmokeRed_Grenade_shell","1Rnd_SmokeGreen_Grenade_shell","1Rnd_SmokeYellow_Grenade_shell","1Rnd_SmokePurple_Grenade_shell","1Rnd_SmokeBlue_Grenade_shell","1Rnd_SmokeOrange_Grenade_shell","ACE_HuntIR_M203"};
```
Glowbal: Does `+=` not work properly here? Are the magazines defined in `GrenadeLauncher`?

addons/huntir/functions/fnc_cam.sqf
```diff
@@ -0,0 +1,145 @@
...
+		};
+		case 3: {
+			GVAR(cam) camsetFOV 0.1;
+			GVAR(cam) camSetFocus [(GVAR(pos) select 2)/8, 1];
+		};
+	};
+
+    private ["_cam_coord_y", "_cam_coord_x", "_speed", "_cam_time", "_cam_pos"];
+    
+	GVAR(logic) setPosATL (GVAR(pos) vectorAdd [0, 0, -5]);
```
Glowbal: tabs

addons/huntir/functions/fnc_huntir.sqf
```diff
@@ -0,0 +1,93 @@
...
+                            GVAR(state) = "connected";
+                        };
+                    };
+                };
+                case "noGDS": {
+                    ctrlSetText [1, "No GDS System detected"];
+                    [{
+                        GVAR(done) = true;
+                        closedialog 0;
+                        13521 cutText ["", "PLAIN"];
```
Glowbal: We probably want to store this layer in a global var or define. Maybe use that BI function to grab a layer id?

addons/huntir/functions/fnc_huntirCompass.sqf
```diff
@@ -0,0 +1,88 @@
...
+    if (_dir >= 90 && {_dir <= 270}) then {
+        _pos set [1, (_pos select 1) - __OFFSET_Y]
+    };
+    if (_dir >= 180 && {_dir <= 360}) then {
+        _pos set [0, (_pos select 0) + __OFFSET_X]
+    };
+    _pos
+};
+
+1234 cutRsc ["ace_huntir_cam_rose", "PLAIN"];
```
Glowbal: Same thing here

addons/huntir/functions/fnc_keyPressed.sqf
```diff
@@ -0,0 +1,115 @@
...
+            if (!GVAR(IRON)) then {
+                GVAR(IRON) = true;
+                ["ace_ir_on", []] call CBA_fnc_localEvent;
+            };
+        } else {
+            setAperture -1;
+            camUseNVG false;
+            GVAR(NV) = 0;
+            GVAR(IRON) = false;
+            ["ace_ir_off", []] call CBA_fnc_localEvent; // TODO: also on death etc?
```
Glowbal: Maybe we should prefix the event names with the module name?

PR #1526, opened 2015-06-06
Update rain each frame

ulteq:
Prevents interrupted raining.


BullHorn7:
This is how Weather behaves when Overcast is set between 0.64 to 0.69 in ACE3.0.1:

https://www.youtube.com/watch?v=j_vwMBHEfr0

This is how Weather behaves when Overcast is set to 85 and Forecasted Overcast is set to 0 in 30 minutes after this particular change:

https://www.youtube.com/watch?v=RemvE5RgP5g (Upload should be done 30 minutes from this post.

tl;dw: This fix seems good.

PR #1530, opened 2015-06-07
Fixed Zeus Unconscious

legman:
I have added a fix for the way Zeus 'remote controlled' units are handled by the medical system. Originally they were treated like players which caused them to ignore the module settings that affect AI.

This means all AI are treated the same now by the medical system regardless of if Zeus is actually controlling them or not!


esteldunedain:
In fact this is something we planned to include in `ACE_common_fnc_isPlayer` at some points and then it fell through the cracks and was forgotten. I think you should fix it there instead,

legman:
OK sure will get on it!

legman:
This is quite tricky actually...
The way it is currently set up means that a unit controlled by Zeus **IS** considered a player. What I am doing above is quite the opposite so they are treated by ACE as AI.

Any ideas how I should handle this safely without causing problems?

SilentSpike:
You could introduce a second optional boolean parameter to the function that flips that behaviour

legman:
Yes thats a nice idea, so we have an additional *zeusCheck* parameter!

Lets hear some more opinions first!

legman:
This does the trick, i have tested all cases and everything is working perfect.
Let me know if you can think of a better variable name!

addons/common/functions/fnc_isPlayer.sqf
```diff
@@ -1,15 +1,21 @@
...
  *
  * Return Value:
  * Bool: is unit a player?
  */
 #include "script_component.hpp"
 
-isPlayer (_this select 0) || {_this select 0 == call FUNC(player)}
+private ["_unit", "_excludeRemote"];
+
+_unit               = [_this, 0] call BIS_fnc_param;
```
Glowbal: `BIS_fnc_param` should not be used.

legman: should i use PARAM_2, if so how do i set the default value?

Glowbal: We use this:
```sqf
_myParameter = if (count _this > 1) then { _this select 1 } else { DEFAULT_VALUE };
```

addons/medical/functions/fnc_setUnconscious.sqf
```diff
@@ -47,7 +47,7 @@ if (_unit == ACE_player) then {
 };
 
 // if we have unconsciousness for AI disabled, we will kill the unit instead
-if (!([_unit] call EFUNC(common,isPlayer)) && !_force) then {
+if (!([_unit, true] call EFUNC(common,isPlayer)) && !_force) then {
```
Glowbal: I think we should make the exclude remote controlled units a setting.

legman: ok - will get it done!

legman:
Let me know if this what you are after!

legman:
If everyone is happy with this commit is there any chance we can get it merged with 3.1.0?

I feel its quite an important change for those who use Zeus on a regular basis!

Thanks

legman:
Is it up to me to fix these merge conflicts or will that be down to the person in charge of the merge?

Thanks

SilentSpike:
Whoever is making the pull request has to resolve them before it can be pulled, do you know how to?

I can briefly explain if necessary

legman:
I would appreciate it mate if you have time!
I am very familiar with SVN as we use it at my workplace - it shouldn't be too different to that I'm guessing.

jonpas:
Update your master, merge it into your feature branch (this pull request), Git will tell you where are conflicts, you open those files and resolve them. To find them in the file search for HEAD, everything after it until `>>>>>` (maybe they point the other way) is what is in your branch, everything from that to master line is what's in what you just pulled. You can either discard one (manually remove), merge it as you like... etc.

> PS: I am using the git shell rather than the GUI

Good boy!

legman:
You are a gent!
I'll get that sorted after work and let you know if i have any problems :+1:

jonpas:
Perfect, get on Slack if you have problems, it's easiest to get help there.

SilentSpike:
There are two methods to resolve, `merge` and `rebase`. Personally I like to rebase, but I'm not entirely sure how best to explain it. So here's a brief overview of a merge (though I'm sure there are shortcuts I don't know about):

1. If you haven't already you'll want to add the ACE repository as a [new remote](https://help.github.com/articles/configuring-a-remote-for-a-fork).
2. Assuming you named the remote `upstream` you'll want to `git fetch upstream`. Which will fetch the upstream changes since you last fetched them.
3. `git merge upstream/master` will merge the latest changes from the master branch into whichever branch you currently have checked out. If there are merge conflicts you'll get a notification saying they should now be resolved and then committed.
4. From here, if you're using solely the command line you can use `git status` to get an overview of what needs to be fixed. However it's probably much easier if your git client helps out like mine does.
5. Here's [an example](http://i.imgur.com/AOT5puv.jpg) I quickly whipped up to show how this works. You can see that git has put conflict markers (`>>>`, `<<<`, `===`) where it was unsure how to merge the changes Above the `===` marker are the conflicting changes in my working branch (`HEAD`) and below are the conflicting changes from the `upstream/master` branch.
6. You simply go through these changes and edit the conflicted files to the point where they make sense again. Sometimes you'll simply need to remove one of the changes entirely and overwrite with the other. Sometimes you'll need to create a combination of the two. Note that there can be multiple conflicts within one file so make sure are no conflict markers left.
7. When a file has had its conflicts resolved you mark it as such with `git add <filepath>`. Where `filepath` is the relative path from the repository root.
8. Once they're all done you just `git commit` to commit the merge

The difference between rebase and merge is that rebase rewinds the commits and applies them on top of each other as if they were done in series rather than parallel (that's my basic understanding of it).

jonpas:
Just to fill in, DO NOT use `rebase` unless you are absolutely 200% sure what it does (flummi's words). :)

legman:
Ok well i found and fixed the conflicts in the one file. Is it ok that ALL that shit has come through though?

I only ammended a single file 'fnc_setUnconscious.sqf'!

esteldunedain:
> Is it ok that ALL that shit has come through though?

Yes, it's alright, don't worry.

When you merged master onto your branch all the changes since you originally forked were added here as well. But that was the point, so nothing to worry about. The diff looks fine.

Glowbal:
Thank you @legman

Issue #1208, opened 2015-05-14
Holster weapon key should unholster it too


esteldunedain:
Now that weapon select keys are not binded, it's cumbersome to unholster, maybe even impossible if you only have one weapon.

The "Holster Weapon" key should unholster it too if pressed again.

PR #1538, opened 2015-06-07
#1208 - Holster key also unholsters

PabstMirror:


esteldunedain:
:+1: for merging now

bux578:
It looks as if it would unholster a primary weapon (if not empty) ... and so on.

Imho, it should unholster the weapon you've holstered.

PR #1529, opened 2015-06-07
Request to add finger-pointing module to ACE 3 mod

TheDrill:
I suggest to add my finger-pointing addon to ACE 3 mod as a module as requested by some players. I tried to adopt the structure of the addon to the standards of ACE 3 as mush as possible. The operability of the resulting module was tested with ACE 3 mod version 1.0.0.3. Please do necessary modifications to the code to make fully compatible with the newest ACE 3 version if needed.

The URL to the thread of original addon:
http://forums.bistudio.com/showthread.php?188895-d_finger-quot-Point-a-finger-at-quot-addon

PR #1548, opened 2015-06-09
Fingering

PabstMirror:
Ported @TheDrill 's #1529 to cba/ace format


addons/finger/XEH_postInit.sqf
```diff
@@ -0,0 +1,23 @@
...
+
+if (!hasInterface) exitWith {};
+
+GVAR(lastFPTime) = -1;
+GVAR(fingersHash) = HASH_CREATE;
+GVAR(pfeh_id) = -1;
+
+["SettingsInitialized", {
+    //If not enabled, dont't bother adding keybind or eventhandler
+    if (GVAR(maxRange) <= 0) exitWith {};
```
esteldunedain: I like how this is setup. However, I think something like `!GVAR(enabled)` instead of `GVAR(maxRange) <= 0` setting would probably be more user friendly, even if it's technically redundant.

addons/finger/config.cpp
```diff
@@ -0,0 +1,18 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        units[] = {QGVAR(moduleSettings)};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"ace_common"};
+        author[] = {"Drill"};
```
esteldunedain: @PabstMirror, you should probably add yourself here too

addons/finger/functions/fnc_keyPress.sqf
```diff
@@ -0,0 +1,55 @@
...
+    if ((((eyePos _x) vectorDistance _playerEyePos) < GVAR(maxRange)) &&
+            {alive _x} &&
+            {(_x == (vehicle _x)) || {(vehicle _x) isKindOf "StaticWeapon"}} &&
+            {GVAR(indicatorForSelf) || {_x != ACE_player}} &&
+            {!(lineIntersects [(eyePos _x), _playerEyePos, ACE_player, _x])} &&
+            {[_x] call EFUNC(common,isPlayer)}) then {
+
+        _sendFingerToPlayers pushBack _x;
+    };
+} forEach allUnits;
```
esteldunedain: Wouldn't it be more efficient to search nearby CAManBase and StaticWeapons objects?

addons/finger/functions/fnc_keyPress.sqf
```diff
@@ -0,0 +1,55 @@
...
+//make sure player is dismounted or in a static weapon:
+if ((ACE_player != vehicle ACE_player) && {!((vehicle ACE_player) isKindOf "StaticWeapon")}) exitWith {false};
+//Check camera view (not in GUNNER)
+if !(cameraView in ["INTERNAL", "EXTERNAL"]) exitWith {false};
+//Exit if run recently (run every 1 seconds)
+if (ACE_diagTime < (GVAR(lastFPTime) + FP_ACTION_TIMEOUT)) exitWith {false};
+
+GVAR(lastFPTime) = ACE_diagTime;
+
+_fingerPosPrecise = positionCameraToWorld [0, 0, FP_DISTANCE];
```
esteldunedain: It would probably make more sense to do a recursive lineintersect search and determine the actual distance to the pointed object. That would allow pointing to nearby objects and positions too, like: "pickup that dropped magazine", "setup the static MG there", "the patient is over there", etc

The 1000m distance could be kept if no object is detected (e.g. while pointing at the horizon).

addons/finger/functions/fnc_incomingFinger.sqf
```diff
@@ -0,0 +1,35 @@
...
+
+PARAMS_2(_sourceUnit,_fingerPosPrecise);
+
+private ["_data", "_fingerPos"];
+
+//add some random float to location if it's not our own finger:
+_fingerPos = if (_sourceUnit == ACE_player) then {
+    _fingerPosPrecise
+} else {
+    _fingerPosPrecise vectorAdd [random (2*FP_RANDOMIZATION_X) - FP_RANDOMIZATION_X, random (2*FP_RANDOMIZATION_X) - FP_RANDOMIZATION_X, random (2*FP_RANDOMIZATION_Y) - FP_RANDOMIZATION_Y]
```
esteldunedain: I'm not sure adding noise is really needed. Pointing IRL is really accurate, certainly inmensely more than the 20º error we are adding (+-350m error at 1000m distance is way way too much). I think this will be even more confusing if we allow pointing at closer objects.

addons/finger/functions/fnc_keyPress.sqf
```diff
@@ -0,0 +1,55 @@
...
+            {!(lineIntersects [(eyePos _x), _playerEyePos, ACE_player, _x])} &&
+            {[_x] call EFUNC(common,isPlayer)}) then {
+
+        _sendFingerToPlayers pushBack _x;
+    };
+} forEach allUnits;
+
+TRACE_1("sending finger to",_sendFingerToPlayers);
+
+[QGVAR(fingered), _sendFingerToPlayers, [ACE_player, _fingerPosPrecise]] call EFUNC(common,targetEvent);
```
esteldunedain: We are in general not prefixing event names, but maybe it's a good idea to start doing it as here.

addons/finger/functions/fnc_perFrameEH.sqf
```diff
@@ -0,0 +1,49 @@
...
+
+private["_drawColor", "_fovCorrection", "_iconSize", "_timeLeftToShow"];
+
+if (!alive ACE_player) then {GVAR(fingersHash) = HASH_CREATE;};
+// Conditions: canInteract
+if !([ACE_player, ACE_player, ["isNotInside"]] call EFUNC(common,canInteractWith)) then {GVAR(fingersHash) = HASH_CREATE;};
+//make sure player is dismounted or in a static weapon:
+if ((ACE_player != vehicle ACE_player) && {!((vehicle ACE_player) isKindOf "StaticWeapon")}) then {GVAR(fingersHash) = HASH_CREATE;};
+
+_fovCorrection = ((worldToScreen (positionCameraToWorld [1000, 0, 10000])) select 0) - 0.5;
```
esteldunedain: This might break if the zoom level is big enough to leave that point offscreen (5º FOV)

gienkov:
Will it make it into 3.1.0 ?

Issue #1354, opened 2015-05-22
Vanilla wounding system kicks in often


gienkov:
ACE3 Version: 3.0.1

**Mods:**
* @cba_a3
* @ace3
* @acre2
* @rhs
* @leightopforpack
* @BWMod

**Placed ACE3 Modules:**
```
class ace_common_forceAllSettings {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_frag_enabled {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_interaction_EnableTeamManagement {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_magazinerepack_TimePerAmmo {
    value = 1.5;
    typeName = "SCALAR";
    force = 0;
};
class ace_magazinerepack_TimePerMagazine {
    value = 2;
    typeName = "SCALAR";
    force = 0;
};
class ace_magazinerepack_TimePerBeltLink {
    value = 8;
    typeName = "SCALAR";
    force = 0;
};
class ace_map_BFT_Interval {
    value = 1;
    typeName = "SCALAR";
    force = 0;
};
class ace_map_BFT_Enabled {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_map_BFT_HideAiGroups {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_map_mapIllumination {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_map_mapShake {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_map_mapLimitZoom {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_map_mapShowCursorCoordinates {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_maptools_EveryoneCanDrawOnBriefing {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_microdagr_MapDataAvailable {
    value = 1;
    typeName = "SCALAR";
    force = 0;
};
class ace_mk6mortar_airResistanceEnabled {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_mk6mortar_allowComputerRangefinder {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_mk6mortar_allowCompass {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
[...]

PR #1550, opened 2015-06-09
Attempt to fix #1465 and #1354

### LINKED ISSUE #1465: Bug report about ACE medical system
ACE3 Version: 3.0.1

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
hearing
explosive system
blue force tracking
respawn system
weather
advanced ballistics
mk6 settings
wind deflection
micro dagr map fill
medical settings
revive settings
advanced medical settings
set medic class
set medical vehicle
set medical facility

**Description:**
I'm not sure is it bug or not

1. sometimes when i get shot i have no injuries in ACE interaction and can see visually that my body is injured but can't run, you can run after apply PAK but if you being hit again you will walk again
i highly assume that it's happening because falling damage, i can rapdly seeing this symptom when people get falling damage.

2. and sometimes i can't revive my team mate, his pulse and blood pressure is on stable no bleeding but can't revive him he just lie down there, we can make him with PAK but we should wake him with out PAK.

3.  i can hear medical treatment sound that keep over looping, the only solution about this is respawn and come back

**Steps to reproduce:**
being shot and watch the symptom 

"Multiplay", host by in multiplay session not like TADST or other dedicated server tool 
but it's also happening in dedicated server too. 

**RPT log file:**
none

ulteq:
@gienkov, @Dobakjang, does that fix your issues?


gienkov:
Sadly, no :(.
Best way to repro now - jump from 1 floor of a building.

ulteq:
@gienkov, please test again.

gienkov:
Falling damage is now fixed and causes abrasions only which can be bandaged now by field dressing. Vehicle damage still needs fixing tho but "acupuncture" works now. By "acupuncture" I mean shooting patient once to cause any kind of damage and immediately bandaging it, thus causing to reset "vanilla damage" to 0.

gienkov:
Repro - drive onto wall few times, it will cause vanilla damage which cannot be fixed.

PR #1551, opened 2015-06-09
Removed isNotInside from Sandbag, SpottingScope and Tripod placement

jonpas:
As discussed with @ulteq this was a copy/pasta boo-boo. :grimacing:


esteldunedain:
Makes sense

PR #1553, opened 2015-06-09
Add config color support to interaction menu icons

SilentSpike:
Actions can now accept entry `color = "#RRGGBB"` or `color[] = {R,G,B}` to define the color to be applied to the action icon. They default to white when not given.

I went ahead and made the team management actions use this instead of duplicate colored icons as a proof of concept.


esteldunedain:
Is this feature really that usefull? It could be used for this and for the medical crosses, but does it have any further use?

jonpas:
What about a client-side setting for the whole interaction menu?

esteldunedain:
> What about a client-side setting for the whole interaction menu?

No, I don't like that idea. Icon color currently conveys meaning, at least for the medical crosses. We'd loose that.

SilentSpike:
It's use is open ended so that's kind of a hard question to answer.

What if someone wants to add an action using an existing icon to their mission without having to duplicate and colour it? This saves them a lot of time.

It also makes icons more useful as they can be used and coloured elsewhere too (rather than applying colour on top of an already coloured icon). This is why the majority of the vanilla icons are plain white.

So really it's a convenience thing, it saves having to open up image editing software and making duplicate files in all cases where colour is desired.

SilentSpike:
I'm going to close this (my own PR - in case you're reading this) from the activity section as I think it's edge case and could be implemented better in future anyway.

Issue #1473, opened 2015-06-02
Treatment-Log is displayed in 2 languages


DerZade:
In this case in german and english (see http://i.imgur.com/tJ7i8w9.png | Perspective: DerZade)
Ingame language settings:
DerZade: English
DiS: German

Glowbal:
Closed by #1559

PR #1559, opened 2015-06-09
#1473 - Fix Medical Log Localization

PabstMirror:
#1473

Check pluse/rate had it's log localized globably
also swaped a few EFUNCs

tested working in mixed language mp:
untitled


Glowbal:
Perfect! Thanks @PabstMirror

PR #1562, opened 2015-06-10
Fix handleDamage caching return value

Glowbal:
Should never return nil. Instead, return a 0 as it's expecting new damage. this may fix the random vanilla damage kicking in.


Glowbal:
@gienkov this may fix your issue

PR #1579, opened 2015-06-11
[Adv. Medical]  New module options

ulteq:
- [x] Add a module option for `healHitPointAfterAdvBandage`
- [x] Add a module option to chose between pain suppression and pain removal.

### DIFF
--- a/addons/medical/functions/fnc_treatmentAdvanced_medicationLocal.sqf
+++ b/addons/medical/functions/fnc_treatmentAdvanced_medicationLocal.sqf
@@ -72,13 +72,10 @@ if (alive _target) then {
 };
 
 if (_painReduce > 0) then {
-    // Reduce the pain level
-    if (GVAR(painIsOnlySuppressed)) then {
-        // Suppress the pain
-        _painSuppress = _target getvariable [QGVAR(painSuppress), 0];
-        _target setvariable [QGVAR(painSuppress), (_painSuppress + _painReduce) max 0];
-    } else {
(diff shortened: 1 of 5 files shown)


addons/medical/functions/fnc_treatmentAdvanced_medicationLocal.sqf
```diff
@@ -73,8 +73,15 @@ if (alive _target) then {
 
 if (_painReduce > 0) then {
     // Reduce the pain level
-    _painSuppress = _target getvariable [QGVAR(painSuppress), 0];
-    _target setvariable [QGVAR(painSuppress), (_painSuppress + _painReduce) max 0];
+    if (GVAR(painIsOnlySuppressed)) then {
+        // Suppress the pain
+        _painSuppress = _target getvariable [QGVAR(painSuppress), 0];
+        _target setvariable [QGVAR(painSuppress), (_painSuppress + _painReduce) max 0];
+    } else {
```
Glowbal: I would suggest to move the pain suppress above the if statement and do this instead:
```sqf
if (!GVAR(painIsOnlySuppressed)) then {
    _pain = _target getvariable [QGVAR(pain), 0];
    _target setvariable [QGVAR(pain), (_pain - _painReduce) max 0, true];
};
```
That has the benefit of still suppressing pain when it's added after morphine has been given.

ulteq: Agreed.

PR #1581, opened 2015-06-11
One PFEH for execNextFrame and waitAndExec

PabstMirror:
For each waitAndExec we create a new `CBA_fnc_addPerFrameHandler` and remove it.
They do have some overhead for stuff like handles that we really don't need.
```
x1 = [{systemChat "x"}, 0, []] call CBA_fnc_addPerFrameHandler; 
[x1] call cba_fnc_removePerFrameHandler;
```
is ~0.3 ms.  There also is an array that permanently grows each time we add: 
`count cba_common_pfhhandles` was about 3500 after an hour.

This just creates one single permanent pfeh that handles both execNextFrame and waitAndExec.  It's pretty lightweight and I'm think having it up all the time should be faster than constantly adding and removing.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -3,6 +3,27 @@
 
 //IGNORE_PRIVATE_WARNING("_handleNetEvent", "_handleRequestAllSyncedEvents", "_handleRequestSyncedEvent", "_handleSyncedEvent");
 
+//Singe PFEH to handle execNextFrame and waitAndExec:
+[{
+    private ["_entry"];
+    
+    //Handle the waitAndExec array:
+    while {((count GVAR(waitAndExecArray)) > 0) && {((GVAR(waitAndExecArray) select 0) select 0) <= ACE_Time}} do {
+        _entry = GVAR(waitAndExecArray) deleteAt 0;
+        (_entry select 2) call (_entry select 1);
+    };
+
+    //Handle the execNextFrame array:
+    {
+        (_x select 0) call (_x select 1);
+    } forEach GVAR(nextFrameBufferA);
+    //Swap double-buffer:
+    GVAR(nextFrameBufferA) = GVAR(nextFrameBufferB);
+    GVAR(nextFrameBufferB) = [];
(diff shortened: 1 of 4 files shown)


esteldunedain:
Yes, I like this!

esteldunedain:
On a side note, the `cba_common_pfhhandles` array should probably be compacted once in a while, o course that's tricky because doing it trivially would break the handlers. However it's probably feasable to maintain a second array mapping UIDs to pfh indexes.

Glowbal:
Very nice. :+1:

ViperMaul:
:+1:

Glowbal:
I am wondering if we could implement something like this for PFH as well. In ACE we almost never use the IDs and remove the PFH from inside the PFH. We could maybe apply the same solution from `waitAndExec` to the PFH and use a return value in the PFHs to determine whatever or not it should be removed?

esteldunedain:
> I am wondering if we could implement something like this for PFH as well. In ACE we almost never use the IDs and remove the PFH from inside the PFH. 

Yeah, that sounds like a good idea. If you only ever remove them from inside a return value is a good method to signal termination. As far as I see it the only gain would that the ace_pfh least could be compacted as needed, right? So we wouldn't ever have to iterate though a 3500+ array each frame.

However...thinking this a bit more. Why is the cba array growing? Doesn't it assign new pfh to old unused slots? Cause that's the obvious solution. Is it supposed to do that but broke?

PabstMirror:
The big array is just for handle lockups because each handle is unique,
It doesn't iterate. it's just a simple `select` which should be the same cost regardless of size
So it shouldn't be any slower if the array grows big (except some memory cost?)

esteldunedain:
> So it shouldn't be any slower if the array grows big (except some memory cost?)

Given that arrays are ot type std::vector in the engine, pushing back to them requires copying the whole content to a new memory position when the reserved space runs out. This doesn't happen every time, but can happen fairly often it reserved memory is not planned for. That can potentially make pushing back costly, although it's hard to say exactly how much without timing it. 

An alternative is to make this vector very big from the start and using `set` instead of `pushBack`, but that would require modifications to cba.

PR #1585, opened 2015-06-11
HuntIR parachute falling speed experiment

ulteq:
Based on: http://www.defensereview.com/1_31_2004/MEIHUNTIR.PDF


addons/huntir/functions/fnc_handleFired.sqf
```diff
@@ -33,15 +33,20 @@ if (_ammo != "F_HuntIR") exitWith {};
...
-            private ["_deltaT"];
             if (isNull _huntir) exitWith {
                 [_this select 1] call CBA_fnc_removePerFrameHandler;
             };
+            private ["_deltaT", "_velocity"];
+            _velocity = velocity _huntir;
             if (damage _huntir > 0) then {
                 _deltaT = ACE_time - (_huntir getVariable [QGVAR(lastTime), ACE_time]);
-                _huntir setVelocity (velocity _huntir vectorAdd [0, 0, -9.8066 * (damage _huntir) * _deltaT]);
+                _velocity = _velocity vectorAdd [0, 0, -9.8066 * (damage _huntir) * _deltaT];
```
esteldunedain: if the huntir get 0 < damage <= 0.1 then it will fall slower than if at full health.

ulteq: The unmodified falling speed is around 7.7 m/s

esteldunedain: But isn't z velocity set to -1 on line 46?

ulteq: Yes, but it changes back very quickly, if you stop the loop.

esteldunedain: > Yes, but it changes back very quickly, if you stop the loop. 

But once it takes damage the loop doesn't stop until the huntir dies, right?

ulteq: But only the `else` block overrides the velocity. The `if` block velocity change is relative to the current speed.

esteldunedain: Yes, totally right, sorry. I still don't understand why the else block exists at all, but that's another discussion,

ulteq: The `else` block is reasoning of this PR. It forces the parachute to decend with a reasonably slow speed.

esteldunedain: Fuck, now I finally understand it: you want to slow down the 7.7 m/s fall velocity, which it's clearly too high according to the specs. Yeah, obviously it would be nice to be able to do that through config, but I don't know how either. Maybe somethingrelated to setMass?

ulteq:
I tried `_huntir setMass 0;`, but it does not work. The mass is already 0.

addons/huntir/functions/fnc_handleFired.sqf
ulteq: I tried `_huntir setMass 0;`, but it does not work. The mass is already 0.

bux578:
:+1:

addons/huntir/functions/fnc_handleDamage.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * Nothing
+ *
+ * Return value:
+ * None
+ */
+#include "script_component.hpp"
+
+PARAMS_5(_huntir,_selectionName,_damage,_source,_projectile);
+
+systemChat format["Selection: %1; Damage: %2", _selectionName, _damage];
```
Glowbal: You merged with debug prints

PR #1586, opened 2015-06-11
PAK treatment time based on total amount of damage.

ulteq:


addons/medical/functions/fnc_treatmentAdvanced_fullHealTreatmentTime.sqf
```diff
@@ -0,0 +1,21 @@
+/**
+ * fn_healTreatmentTime.sqf
+ * @Descr: N/A
+ * @Author: Ruthberg
+ *
+ * @Arguments: []
+ * @Return:
+ * @PublicAPI: false
```
Glowbal: This is the wrong function header format - it's from CSE.

Issue #1577, opened 2015-06-11
Medical/Revive, Player standing upright while unconscious when hurt while exiting vehicle


captainblaffer:
ACE3 Version: 3.0.1
(stable)

Mods:
@CHVD;@ASDG_JR;@ASDG_JM;@AllInArmaTerrainPack;@BWA3;@CBA_A3;@dariyah;@fhq_accessories;@hlcmods;@k_mnp;@ace;@stui;@tao_foldmap_a3;@task_force_radio;@rhsafrf;@rhsusf;@3cb_baf_weapons;@rhs-hlc;

Placed ACE3 Modules:
ACE_ModuleSwitchUnits
ACE_ModuleFriendlyFire
ACE_moduleReviveSettings
ACE_moduleMedicalSettings
ACE_ModuleInteraction
ACE_moduleAssignMedicalFacility

Description:
When player goes unconscious while exiting vehicle, they will be standing upright instead of lying on the ground

You can see the issue happening here:
http://www.twitch.tv/captblaffer/v/6028830
timestamp: 38:30

Where did the issue occur?
"Multiplayer"

Issue #1575, opened 2015-06-11
Impossible to unload wounded soldier from a vehicle


Sartax:
Hi ! Awesome addon, I was waiting this since a long time !

But I have some question (because we have some problems with my team).

We are using RHS Mod (U.S Forces), and for us it's impossible to unload somebody from a vehicles if he's inconscious. If we try to unload somebody, first case, the soldier remains in the vehicle, and in the second case, he's unload, but he's teleport in the vehicle.

ulteq:
http://ace3mod.com/wiki/user/how-to-report-an-issue.html

PR #1587, opened 2015-06-11
fix animation and consistently unload units from vehicle

ViperMaul:
Fix animation and unconscious unloading from vehicles #1575 and #1577
Switch to event driven.


Glowbal:
While we are here, this just needs `setPos` to be switched for `setPosASL` and it can be merged in.

Everything else looks :+1:

PR #1589, opened 2015-06-12
Optional setting to disable laser pointer system

PabstMirror:
Adds an ACE_setting to control the laser pointer feature.

A lot of times we're not using the feature, and it does have some cost:
On each frame `ace_laserpointer_onDraw_counter: Average: 0.460449s / 1953 = 0.235765ms`

If system isn't enabled, we never call onDraw or add the nearUnits PFEH..  
If users starts with a "laser" attachement the keybind will still allow them to switch back to a "IR" pointer.


esteldunedain:
Nice

ulteq:
Can't we use nearEntities instead of nearestObjects?
```
_nearUnits = (positionCameraToWorld [0,0,0]) nearEntities ["Man", 50];
if (count _nearUnits < 10) then {
    {
        _nearUnits append crew _x;

        if (count _nearUnits > 10) exitWith {
            _nearUnits resize 10;
        };
        
    } forEach ((positionCameraToWorld [0,0,0]) nearEntities ["AllVehicles", 50]);
} else {
    _nearUnits resize 10;
};
```
This runs a lot faster for me.

SilentSpike:
Pretty sure you can use nearEntities, the command is known to be faster :+1:

Also, instead of checking if the array gets bigger than 10 and then resizing it back to 10, just check if it's 10 and then exitWith.

PabstMirror:
`nearestObjects` sorts by distance, so the results wouldn't be exactly the same.

ulteq:
We can easily sort the nearEntities array with the new `sort` command. It would still be much faster. I did the same in the HuntIR module.

PR #1593, opened 2015-06-12
fix dagr not showing after aiming

SzwedzikPL:
Fix for dagr not showing after aiming (cameraView gunner)

### DIFF
--- a/addons/dagr/XEH_postInit.sqf
+++ b/addons/dagr/XEH_postInit.sqf
@@ -3,6 +3,7 @@
 #include "initKeybinds.sqf"
 
 GVAR(run) = false;
+GVAR(hidden) = false;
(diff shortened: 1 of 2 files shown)


ulteq:
The DAGR is currently the only device which is hidden while you aim.

I'd say we either remove the `cameraView` check completely, or make all devices behave like that.

I prefer the former.

SzwedzikPL:
> The DAGR is currently the only device which is hidden while you aim.

MicroDAGR behave is ok :)

SzwedzikPL:
Merged with https://github.com/acemod/ACE3/pull/1595 :)

bux578:
> I'd say we either remove the cameraView check completely, or make all devices behave like that.

the latter :)

Issue #1018, opened 2015-05-05
Remove "Frisk person" in favour of "Open Inventory"


bux578:
![image](https://cloud.githubusercontent.com/assets/1235520/7479470/220a85da-f362-11e4-8366-1f7f73f3f4b4.png)

xJenny99x:
Frisking people is absolutely HARAM!

PR #1594, opened 2015-06-12
Gear instead of frisk (ace selectmenu)

SzwedzikPL:
https://github.com/acemod/ACE3/issues/1018


bux578:
hm


From looking at the changes I think you there'd still be two options. `Frisk` (which now opens the gear menu) and `Open Invetory`

PabstMirror:
`_player action ["Gear", _unit];`

This single line seem to do everything that all of "disarming" module does.
I thought there was a reason why "gear" action didn't work, but things may have changed.

Minor thing is you can't remove an item unless you have room to carry, but that isn't critical.
I will do some testing on this.

PR #1598, opened 2015-06-13
Fix Server not reading modules

PabstMirror:
Critical:
`InitSettingsFromModules` in only called from a PFEH bellow `if (!hasInterface) exitWith {};`,
so dedicated servers never read modules.

Also, fixes a possible issue where the server has synced GVAR(settings), but not finished syncing all the modules.  Clients could raise `SettingsInitialized` before all module setting have been recieved.  Server now sets a bool publicVariable after reading all modules.


esteldunedain:
Totally true, and definitely an improvement but...

I'm thinking now that if "SettingsInitialized" shouldn't be thrown only after the init from module thing has finished on the server. TBH this whole settings system is giving me headaches atm.

ViperMaul:
Keep your thoughts on this one. This PR is important for most communities. I'd like to see it get finished in time.

esteldunedain:
I've read it more carefully, and @PabstMirror addressed my concern already. Perfect!

Bottom line:

1 - Server inits settings from config
2 - Server inits modules
3 - Clients launch the "SettingsInitialized" event locally when the above is ready and received.

Brilliant!"

PR #1599, opened 2015-06-13
Captives fixes

PabstMirror:
Removes "escort captive" action if either is mounted.
Unloading is now a passenger action instead of on the vehicle.
Adds passenger action icon if unit is handcuffed in vehicle.


bux578:
https://github.com/acemod/ACE3/pull/1597 captives could get their own slot icon.

bux578:
:+1:

PR #1605, opened 2015-06-13
Fixed select new injuryToAdd always selects first

Glowbal:
It should not always take the first injury in the list. That would with the current configuration almost always be the same injury as in `highestSpot`.

This might need to be expanded with something that prefers higher damage types.


Glowbal:
Tested - works perfectly

PR #1608, opened 2015-06-13
Added 'glsmoke'

TheNiner:
Added 'glsmoke' which will stop the GL Smokes from bouncing through the
air.

### DIFF
--- a/addons/glsmoke/CfgAmmo.hpp
+++ b/addons/glsmoke/CfgAmmo.hpp
@@ -1,0 +1,9 @@
+class CfgAmmo
+{
+	class SmokeShell;
+	class G_40mm_Smoke: SmokeShell
+	{
+		simulation = "shotSmoke";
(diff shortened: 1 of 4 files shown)


addons/glsmoke/CfgAmmo.hpp
```diff
@@ -0,0 +1,9 @@
+class CfgAmmo
+{
+	class SmokeShell;
+	class G_40mm_Smoke: SmokeShell
+	{
+		simulation = "shotSmoke";
+		deflecting = 0;
```
ulteq: I don't think that `deflecting = 0;` is a good choice. Did you try slightly higher values as well?

TheNiner: Yeah, i tried a few, but it still behaves very weird then, sometimes it just slides on the ground and bounces uncontrollable at the end.

PR #1631, opened 2015-06-15
Fixed model.cfg and requiredAddons

ulteq:


esteldunedain:
> Does that fix the following RPT errors?

Yep, it's supposed to. Also the wrong backpack position ingame

PR #1635, opened 2015-06-15
Fixed CheckPBO Kick

jonpas:
- Removed function `adminKick` as it is now redundant
- Used close display 46 in combination with `errorMessage` function
- Actual error message is what it is until a rewrite of CheckPBOs comes, meaning no displaying of what mods in the error pop-up as that is displayed in chat from a separate script - hence why waitUntil is also needed


esteldunedain:
Thanks!

PR #1648, opened 2015-06-16
Proper CPR cycle time

gienkov:
Source: http://www.ncbi.nlm.nih.gov/pubmed/19491691
- ```The time required to perform five cycles of CPR was 115+/-5 s```

So one cycle should last about 23 +/- 1 second.
- ```The time needed to deliver the **first two** rescue breaths was between 12 and 15 s```

And you do about 100 compressions per minute, so about 2 per second, so you need about 15 seconds to do 30 compressions.


Glowbal:
Does this still match the animation?

gienkov:
Animation lasts about 6 seconds with transitions.

gienkov:
Conclusion?

PR #1652, opened 2015-06-17
Fix checking vitals anim

gienkov:
It uses animations that don't match at all what you are doing anyway so I think it's better to just disable anim for that actions which are:
* Checking pulse
* Checking BP
* Checking response

### DIFF
--- a/addons/medical/ACE_Medical_Treatments.hpp
+++ b/addons/medical/ACE_Medical_Treatments.hpp
@@ -242,6 +242,8 @@ class ACE_Medical_Actions {
             callbackProgress = "";
             animationPatient = "";
             animationCaller = ""; // TODO
+            animationCallerProne = "";
+            animationCallerSelfProne = "";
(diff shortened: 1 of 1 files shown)


esteldunedain:
I think the "Put" animation would be better than nothing. It's very short, but at least fellow soldier can see that you did something.

gienkov:
Can you tell me what name it has? "Put" may suit checking vitals well.
What about checking while lying down?

esteldunedain:
> Can you tell me what name it has? "Put" may suit checking vitals well.

TBH I don't know. We usually use `_unit playActionNow "PutDown";`, which plays the action correctly in any pose (including prone). I'm pretty sure that just plays a different anim depending on the current pose, but I don't know their names.

Glowbal:
Those should have been disabled, since the others are also all empty.

But we do need some matching animations in the future for those.

Issue #1654, opened 2015-06-17
Expand make to automate release


Glowbal:
Expand our make script to automate parts of the release process.

- [x] Update Version - README.MD
- [x] Update Download Link - README.MD
- [x] Update Core & Additional Features - README.MD
- [x] Update Version in 2 places - ace/mod.cpp
- [x] Update Version - ace/addons/main/script_mod.hpp
- [x] Create Zip file as ace3_<version>.zip
- [x] Create and Tag Pre-Release on GitHub

jonpas:
Where would the version be read from? Or would you input it in make.py run?

esteldunedain:
> Where would the version be read from? Or would you input it in make.py run?

All the require metadata could be the content of a file in the base directory.

jonpas:
> All the require metadata could be the content of a file in the base directory.

Time to change where version is currently read from then. I have the version things almost done, will PR in a bit.

ViperMaul:
@jonpas 
> Where would the version be read from? Or would you input it in make.py run? I suggest \z\ace\addons\main\script_mod.hpp

This is correct. The requirements are here:
https://github.com/acemod/ACE3/pull/1659#issuecomment-113028371

ViperMaul:
Everything looks good. Thank you @jonpas.

PR #1659, opened 2015-06-17
Automated version change and Zip acrhiving in make.py

jonpas:
#1654 

Version taken from `script_mod.hpp`
- [x] Update Version - README.MD
- [x] Update Download Link - README.MD
- [x] Update Version in 2 places - ace/mod.cpp
- [x] Create Zip file as ace3_x.y.z.zip
- [x] Preserve original files containing version


Glowbal:
It might be better if the zip file is created by a command line argument instead.

ViperMaul:
*This build now ignores key command-line params like*
1. make.py checkexternal main (it should only build ace_main.pbo) 
2. make.py checkexternal release 3.1.1 (it still shows that it build 3.1.0 zip)

*Other observations without looking at the code.* 
It seems to only build correctly when I use make.py or make.py checkexternal 
This version of make.py seems to be hard-coded to 3.1.0 when z\ace\adddons\main\script_mod.hpp shows the version to be 3.1.1

Not your fault. This issue ticket was for me to do later. Thus it did not have proper intentions.
You are more than welcome to complete this. Here are our requirements. 
 
The **requirements** are
 1. Maintain all working command-line options
 2. Keep release off by default. 
 3. Read the z\ace\adddons\main\script_mod.hpp and use as the build version number, if file does not exist then use the command-line version number as the release number. 
```make.py checkexternal release 3.1.1.5``` in this case the zip file would be <mod>_3.1.1.5.zip
 4. Allow the following command-line to only affect the zip file version number not the build version
So we could write ```make.py checkexternal release 3.1.1.5``` and the build version would remain what it says in z\ace\adddons\main\script_mod.hpp
In other words, create zip file name is based on script_mod.hpp or the command-line version#
 5. Make sure zip file does not include other previous zip files.

jonpas:
@ViperMaul

1 - Yes, version is hardcoded right now. 

2 - Will change, is enabling via command line enough, or do I add some other option to toggle it as well? 

3,4 - If you check first commit diff you'll see I hardcoded the version number for debug purposes, as not all files had version number updated yet. Possibly that's why command line version number isn't used either.

4 - Should the zip contain build number though, so far none of them did?

5 -  Already assured. 

I will look into all of it again. 

Also, no worries, I am mainly doing this to learn python, even if it doesn't get merged I got something from it.

ViperMaul:
@jonpas 

> 1 - Yes, version is hardcoded right now. 

I figured as much. I didn't have the time to review code. All the time I had was to just run it and take notes. 

> 2 - Will change, is enabling via command line enough, or do I add some other option to toggle it as well?

  This is the way it was before. At the moment we feel this is enough.Yes. I cannot think of another option to add that would make things better for our needs at present.

> 4 - Should the zip contain build number though, so far none of them did? 

That can help with zip file collisions. I think this idea to add the build number is helpful. ace3_3.1.1-bab33bf.zip

> Same fur number 3 after "ace", I though major version is enough. 

I know it would be better for all different project names if we can use the variable 'prefix'
But for now you can if you wish to just make a var called something like ```zipPrefix = "ace3"```

I am glad you are motivated by learning python. Your efforts saves us time. Thank you in advance.

jonpas:
@ViperMaul 

All done except:
> Allow the following command-line to only affect the zip file version number not the build version So we could write make.py checkexternal release 3.1.1.5 and the build version would remain what it says in z\ace\adddons\main\script_mod.hpp In other words, create zip file name is based on script_mod.hpp or the command-line version#

Not sure I understand that, do you want the zip file name to be independent of actual version number if it's supplied in the argument, and if it's not supplied then just use actual version number?

ViperMaul:
@jonpas 
Yes. If a release number is supplied on the command-line then yes. Independent.

ViperMaul:
@jonpas 
It looks good so far. Well done. All requirements met. 

I just haven't decided about another final requirement. What is appropriate to do with the changed files in the repo. See the screenshot: 
image

What do we do from the perspective of all of the people who are non-developers who want to build this and be able to PULLs in the future with any friction or errors. 

Options to consider:
- We could make backups and restore the originals after we are done. 
- Instead of making backups we could use Git Checkout readme.md for example to restore the originals
- We could leave them changed in the repo in ACE devs want to commit them. But 99% of people who build this using release may not be ACE Devs (But those who use Release would probably know what they are doing anyway)

esteldunedain:
> We could make backups and restore the originals after we are done. 

In my opinion it's the easiest solution

jonpas:
I agree, backup and restore. Additional command line option can be part of "release" for that wether it does the backup and restore or no. 

> Yes. If a release number is supplied on the command-line then yes. Independent.

I will make it so.

ViperMaul:
:+1:

PR #1678, opened 2015-06-18
Update version number in version.html

jonpas:
Reported by @jastreb http://i.imgur.com/M8MPzhZ.png?1

### DIFF
--- a/version.html
+++ b/version.html
@@ -3,7 +3,7 @@
         <title>ACE3 Version</title>
     </head>
     <body>
-        <h1>Latest: 3.1.0.0 - <a href="https://github.com/acemod/ACE3/releases/latest">Download</a></h1>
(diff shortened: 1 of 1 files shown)


version.html
```diff
@@ -3,7 +3,7 @@
         <title>ACE3 Version</title>
     </head>
     <body>
-        <h1>Latest: 3.1.0.0 - <a href="https://github.com/acemod/ACE3/releases/latest">Download</a></h1>
+        <h1>Latest: 3.1.1.1 - <a href="https://github.com/acemod/ACE3/releases/latest">Download</a></h1>
```
bux578: `3.1.1` should be enough

jonpas: Check the image, "installed version" includes the build number as well, I think it's best to have them both the same format.

PR #1163, opened 2015-05-14
Frag spall optimization

jaynus:
Changed: Entire frag system re-optimized to a single tracking/GC index-based PFH.
Changed: Spalling calculations now throttled across frames and limited; accurate but performant
Changed: All lookups are now index-based on a global projectiles array. This reduced runtime dramatically for all fragmentation.

Merges and closes #605

PR #1686, opened 2015-06-19
Fix frag

PabstMirror:
Sometime after PR #1163 frag_fnc_fired got overwritten with the old version which calls a `DFUNC(trackFragRound)` which isn't defined anymore, so frag does nothing right now.

I tried restoring it to the state of the PR, had some issues with masterPFH that I think are fixed.

Also did some cleanup and connected it to zeus module.


Cyruz143:
Running this build at the moment and not having any major issues with fragmentation, however spalling is super broken! The bullet ricochets between objects and doesn't seem to have a time to live value so just keeps bouncing endlessly.

Easy to replicate, fire a few rounds in to ```Land_Cargo_House_V1_F``` and wait for the loop.

PabstMirror:
https://github.com/acemod/ACE3/pull/1163/files#diff-f1704d8d894d0b4ae09d5a6ee3390897L11

We removed `[(_this select 1)] call cba_fnc_removePerFrameHandler;`, so it seems like that PFEH is running forever.  Re-adding seems to fix.

PR #1692, opened 2015-06-19
Fix vanilla damage effects

Glowbal:
Vanilla plays damage effects when `damage unit` returns > 0. 

The damage is increased when using return values in the `handleDamage` eventhandler. Returning 0 on the eventhandler appears to be breaking the current implementation of handleDamage in both basic and advanced (all units become invulnerable). 

Solution for advanced only at the moment is to reset damage in the caching and apply proper hitpoint damage after we are done processing the handleDamage calls. As an added side bonus this also seems to be fixing the flashing hitpoint textures.

### DIFF
--- a/addons/medical/functions/fnc_handleDamage.sqf
+++ b/addons/medical/functions/fnc_handleDamage.sqf
@@ -80,9 +80,11 @@ if (GVAR(level) < 2) then {
     } else {
         _damageReturn = _damageReturn min 0.89;
     };
+
 };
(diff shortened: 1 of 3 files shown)


gienkov:
Man this made advanced medical 1000000x more playable, it's getting more and more perfect now :).
But this #1693

esteldunedain:
So, should this be merged already @Glowbal?

Twelvecaliber:
Any chance of this being put into a hotfix release? It's becoming an increasingly frequent problem for people it seems.

PR #1700, opened 2015-06-20
Fill gaps in ball tracer config

Pashe:
Currently, some ammo classes such as B_65x39_Caseless are still laser beams. This PR fixes that.


sargken:
Any status on this?

Issue #1704, opened 2015-06-21
ACE_Box_Ammo unknown animation sources


jonpas:
ACE3 Version: 3.1.2.0 ab580dd1c7bb0dba710a42f0193bb312eceac0c0

**Mods:**
* `@cba_a3`
* `@ace3`

**Placed ACE3 Modules:**
N/A

**Description:**
```
ACE_Box_Ammo: Ammo_hide - unknown animation source Ammo_source
ACE_Box_Ammo: AmmoOrd_hide - unknown animation source AmmoOrd_source
ACE_Box_Ammo: Grenades_hide - unknown animation source Grenades_source
ACE_Box_Ammo: Support_hide - unknown animation source Support_source
```

**Steps to reproduce:**
* Put down an `[ACE] Ammo Supply Crate`
* Check RPT

**Where did the issue occur?**
* SinglePlayer - Editor
* MultiPlayer - Dedicated

**RPT log file:**
N/A

PR #1705, opened 2015-06-21
#1704 - Add missing AnimationSources

PabstMirror:
#1704 Ammo box missing AnimationSources

Also fixes the overlapping text on the side, as it seems like the animations control what's written on the box.


jonpas:
:+1:

PR #1706, opened 2015-06-21
Blue Force Tracking ACE_Settings

jonpas:
At least until new BFT is ready. Fixes https://github.com/acemod/ACE3/issues/414#issuecomment-93979783

### DIFF
--- a/addons/map/XEH_postInitClient.sqf
+++ b/addons/map/XEH_postInitClient.sqf
@@ -18,6 +18,9 @@ call FUNC(determineZoom);
     ((findDisplay 12) displayCtrl 51) ctrlAddEventHandler ["Draw", {[] call FUNC(updateMapEffects);}];
 };
 
+// Client only from here
+if !(hasInterface) exitWith {};
+
 ["SettingsInitialized", {
     // Start Blue Force Tracking if Enabled
     if (GVAR(BFT_Enabled)) then {
--- a/addons/map/XEH_postInitClient.sqf
+++ b/addons/map/XEH_postInitClient.sqf
@@ -18,7 +18,6 @@ call FUNC(determineZoom);
     ((findDisplay 12) displayCtrl 51) ctrlAddEventHandler ["Draw", {[] call FUNC(updateMapEffects);}];
 };
 
-[
 ["SettingsInitialized", {
     // Start Blue Force Tracking if Enabled
     if (GVAR(BFT_Enabled)) then {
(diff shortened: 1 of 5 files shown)


addons/map/functions/fnc_moduleMap.sqf
```diff
@@ -10,6 +10,8 @@
  */
 #include "script_component.hpp"
 
+if !(hasInterface) exitWith {};
```
PabstMirror: To work on a dedicated server this should be:
`if !(isServer) exitWith {};`

EFUNC(common,readSettingFromModule) only works when called from the server

addons/map/stringtable.xml
```diff
@@ -77,6 +77,12 @@
             <German>Blue Force Tracking</German>
             <Czech>Blue Force Tracking</Czech>
         </Key>
+        <Key ID="STR_ACE_Map_BFT_Enabled_DisplayName">
+            <English>Enable</English>
```
PabstMirror: Minor thing, but I would change this to `BFT Enabled` to make it easier to find on the server settings dialog.  I think it should still fit on the module.

addons/map/XEH_postInitClient.sqf
```diff
@@ -19,4 +18,10 @@ call FUNC(determineZoom);
     ((findDisplay 12) displayCtrl 51) ctrlAddEventHandler ["Draw", {[] call FUNC(updateMapEffects);}];
 };
 
-ADDON = true;
+["SettingsInitialized", {
+    // Start Blue Force Tracking if Enabled
+    if (GVAR(BFT_Enabled)) then {
```
PabstMirror: Probably want a `if !(hasInterface) exitWith {};` here so we don't start the PFEH on server/headless

addons/map/functions/fnc_moduleMap.sqf
jonpas: That was my bad, copy/paste from a wrong file.

addons/map/stringtable.xml
jonpas: Sure, changed.

addons/map/XEH_postInitClient.sqf
jonpas: That XEH won't even execute on Server since it's `clientInit`, but good point on Headless. I added `exitWith` in the beginning as there is nothing that should run on Headless at all in the map module afaik.

PabstMirror:
:+1: Looks good

jonpas:
Status on merge?

Issue #1713, opened 2015-06-21
ace_zeus black screen on dedicated server


SilentSpike:
It seems the changes made by ace_zeus cause the screen to go black when a player uses one of the zeus slots in the vanilla zeus mission on a dedicated server.

TODO:
- [x] Narrow down the problem circumstances. Should look into:
  - ~~Does it only occur when using the virtual zeus units (sideLogic)?~~
  - ~~Does it only occur when zeus interface is forced?~~
- [x] Fix the problem

SilentSpike:
It seems that the edits made to the module function are somehow stopping it from correctly initializing only on dedicated servers. The screen going black is just a side effect of having the interface forced.

Time to look through the code.

SilentSpike:
Potential causes I can think of:
- ACE settings variables are not defined before the function runs
- function `EFUNC(common,execNextFrame)` does not work or is not defined on dedicated server

Time to test both out.

SilentSpike:
Okay, neither of those things is the issue, in fact I commented out all the ACE changes and it still didn't initialize.

Which means the problem is possibly just part of the code that doesn't run when the settings are false is required somehow (though I'd have no idea how).

Testing now.

SilentSpike:
Okay, I set the setting to true and it still didn't initialize. :fu:

SilentSpike:
Figured it out, someone must have changed the vanilla code to use `ACE_Time` instead of `time` at one point, which somehow broke the whole module.

NemesisRE:
It seems that 3.2.0 still needs some time, would it be possible to release 3.1.2 with this fix?

SilentSpike:
3.2.0 actually releases today with this fix included :+1:

NemesisRE:
:+1:

PR #1716, opened 2015-06-21
Fix for #1713

SilentSpike:
#1713 was reported in the ACE thread. It turns out the curator module wouldn't initialize on a dedicated server because `waitUntil {time > 0};` had been changed to use `ACE_time`.

I'm PRing this from my fork rather than a branch because I wanted to get it fixed ASAP and didn't have time to reconfigure my local git first.


SilentSpike:
Also threw in README.md and moved myself in authors.txt while I was making changes.

esteldunedain:
Perfect, thanks!

PR #1723, opened 2015-06-22
View Restriction

jonpas:
Talked with @esteldunedain last time and we found out we have same scripts in our communities so we figured it would be a good addition to ACE3, and expanded with more settings.

This module provides settings to restrict view to 1st or 3rd person globally (all types of vehicles) or selectively for different restriction per type of vehicle (currently Foot, Land Vehicles, Air Vehicles and Sea Vehicles). Default is no restrictions.

There is a huge problem with using `switchCamera` on UAVs so this part is disabled. Code is commented out (since it doesn't work, duh) and a comment has been added pointing to the explanation of this bug.


addons/viewrestriction/stringtable.xml
```diff
@@ -0,0 +1,59 @@
...
+            <English>Selective mode in Air Vehicles. Default: Disabled (Requires Mode: Selective)</English>
+        </Key>
+        <Key ID="STR_ACE_ViewRestriction_ModeSelectiveSea">
+            <English>(Selective) Sea Vehicles</English>
+        </Key>
+        <Key ID="STR_ACE_ViewRestriction_ModeSelectiveSeaDesc">
+            <English>Selective mode in Sea Vehicles. Default: Disabled (Requires Mode: Selective)</English>
+        </Key>
+        <Key ID="STR_ACE_ViewRestriction_ModeSelectiveUAV">
+            <English>(Selective) UAVs</English>
```
bux578: `UAV's` Plural of abbreviations get an apostrophe

jonpas: I disagree, reference https://en.wikipedia.org/?title=Abbreviation#Plural_forms

addons/viewrestriction/XEH_postInitClient.sqf
```diff
@@ -0,0 +1,98 @@
...
+// Exit if third person view is not available
+if !(difficultyEnabled "3rdPersonView") exitWith {};
+
+["SettingsInitialized", {
+    // Exit if module is disabled
+    if (GVAR(mode) == 0) exitWith {};
+
+    // Start proper PFH based on mode (separated from one PFH with many conditions to improve performance)
+
+    // Conditions with cameraView results in visible transitions, ongoing use of switchCamera does not
```
esteldunedain: Ok, too bad, it would have been much cleaner the other way aroud.

esteldunedain:
I'm just wondering how friendly this is with third person camera scripts, like spectator mode, virtual arsenal, etc. Well VA is not a big deal as this thing does nothing by default unless settings are set specifically.

addons/viewrestriction/XEH_postInitClient.sqf
jonpas: Indeed, because cameraView will return true only once already in that view and that happens 50% into the transition. `switchCamera` itself shouldn't have a performance impact though (feel free to profile, I am not experienced with that).

jonpas:
Will that do?

esteldunedain:
Probably, IDK

jonpas:
Well, it's disabled by default anyways, if anyone uses it with some spectator script I am sure there will be reports and we can look into adding needed conditions then.

SilentSpike:
I believe this will break spectator scripts/zeus, where cameraOn will return the player unit still (and will switch the camera to view the unit instead of being free).

You can check if zeus is open using !(isNull curatorCamera)

addons/viewrestriction/XEH_postInitClient.sqf
```diff
@@ -0,0 +1,104 @@
...
+                };
+                if (GVAR(modeSelectiveFoot) == 2) exitWith {
+                    cameraOn switchCamera "External";
+                };
+            };
+
+            // Land Vehicles
+            if (cameraOn isKindOf "LandVehicle") exitWith {
+                if (GVAR(modeSelectiveLand) == 1) exitWith {
+                    cameraOn switchCamera "Internal";
```
SilentSpike: Have you tested this? Does it automatically switch the camera to the appropriate player view when applied to a vehicle or does it make everyone in the vehicle have the same default vehicle view?

jonpas: I have tested it all in Editor, but didn't get a chance to do it in multiplayer yet, though it shouldn't have any impact at all, all of it is run client-side only.

Settings are applied from via Settings Framework (ACE_Settings or module), they are not client-side settings, so it shouldn't matter if it does for all or not. Additionally BI wiki says:
> If you switch to a unit in a vehicle, this command uses the correct turret.

`cameraOn` returns the object so either vehicle player or player directly, which would be an alternative.

SilentSpike:
Just to document, cameraOn equivalents:
- zeus free cam - `(vehicle player)`
- zeus internal cam - `cameraOn`
- zeus remote control - `ACE_player` (note that it's not `(vehicle ACE_player)`)
- zeus virtual arsenal - `(vehicle player)`

addons/viewrestriction/XEH_postInitClient.sqf
SilentSpike: Yeah, I was originally confused by the word turret there. Works well :+1:

There could be some issues with the switch units stuff however, assuming it behaves the same as zeus remote control then `cameraOn` will return the unit being controlled (not the vehicle it's in) which means the vehicle detection is off. However if you were to use `(vehicle cameraOn)` then it would use the default vehicle view.

jonpas: Hmh could be the same thing for UAVs if you read the comment, it returns control to player but camera is stuck on the UAV. Though I believe that is an Arma bug.

You think replacing all `cameraOn` with `vehicle cameraOn` should work for all?

SilentSpike: > You think replacing all cameraOn with vehicle cameraOn should work for all?

Nope, precisely the opposite. I'm just warning against that as it would make things worse. As is, the "on foot" view will be applied no matter what when remote controlling a unit. However if you were to switch it to use the vehicle then, for example, passengers would see the driver view, etc.

jonpas: Ah, good point! :+1:

jonpas:
I will do some Zeus testing and then can look into some randomly picked spectator scripts. Preferably we'd want implementation that doesn't need compatibility stuff, but simply works when it has to.

Glowbal:
What is the status on this?

jonpas:
Trying to figure out a good way of determining when it should be running (player unit, not spectator... etc.).

jonpas:
This is finished and went through initial testing. `switchCamera` is crappy to the point where it doesn't work on remote controlled units (no matter what i use as the unit), so I just disabled that.

It is up to how the spectator scripts work (if they use original player unit or not - `player == ACE_player`). I think it's best if all that is added to documentation.

I am also considering another module option in which mission makers could put in a piece of code which would be used to evaluate if that specific spectator script or whatever they are running is enabled.

Glowbal:
Thank you @jonpas We will take a look at this shortly.

addons/viewrestriction/CfgVehicles.hpp
```diff
@@ -0,0 +1,85 @@
...
+class CfgVehicles {
+    class ACE_Module;
+    class GVAR(ModuleSettings): ACE_Module {
+        author = ECSTRING(common,ACETeam);
+        category = "ACE";
+        function = QFUNC(moduleInit);
+        displayName = CSTRING(ModuleDisplayName);
+        scope = 2;
+        isGlobal = 1;
+        //icon = QUOTE(PATHTOF(UI\Icon_Module_ViewRestriction_ca.paa)); //add icon
```
jonpas: Add this...

addons/viewrestriction/ACE_Settings.hpp
```diff
@@ -0,0 +1,48 @@
+#define MACRO_VALUES \
+    CSTRING(Disabled), CSTRING(FirstPerson), CSTRING(ThirdPerson)
+
+class ACE_Settings {
```
jonpas: Add category.

addons/viewrestriction/CfgVehicles.hpp
```diff
@@ -0,0 +1,85 @@
...
+        value = 1; \
+    }; \
+    class ThirdPerson { \
+        name = CSTRING(ThirdPerson); \
+        value = 2; \
+    };
+
+class CfgVehicles {
+    class ACE_Module;
+    class GVAR(ModuleSettings): ACE_Module {
```
jonpas: Rename to `GVAR(Module)`.

addons/viewrestriction/XEH_clientInit.sqf
```diff
@@ -0,0 +1,34 @@
+#include "script_component.hpp"
+
+// Exit on Headless Client
+if !(hasInterface) exitWith {};
+
+// Exit if third person view is not available
+if !(difficultyEnabled "3rdPersonView") exitWith {};
```
jonpas: Add some RPT log if it's any of the mods are enabled but not available due to this line.

addons/viewrestriction/XEH_preInit.sqf
```diff
@@ -0,0 +1,10 @@
+#include "script_component.hpp"
+
+ADDON = false;
+
+PREP(canChangeCamera);
```
jonpas: To `XEH_PREP.sqf` and add `XEH_preStart.sqf`.

addons/viewrestriction/functions/fnc_canChangeCamera.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * Can Change Camera <BOOL>
+ *
+ * Example:
+ * [] call ace_viewrestriction_fnc_canChangeCamera
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+// Remote control hates switchCamera (control returns to player, camera is left on remotely controleld object/unit), make sure remote controlled units are not impacted
```
jonpas: controlled

addons/viewrestriction/functions/fnc_moduleInit.sqf
```diff
@@ -0,0 +1,35 @@
...
+// Read selective options only if selective mode selected
+if (GVAR(mode) == 3) then {
+    [_logic, QGVAR(modeSelectiveFoot), "modeSelectiveFoot"] call EFUNC(common,readSettingFromModule);
+    [_logic, QGVAR(modeSelectiveLand), "modeSelectiveLand"] call EFUNC(common,readSettingFromModule);
+    [_logic, QGVAR(modeSelectiveAir), "modeSelectiveAir"] call EFUNC(common,readSettingFromModule);
+    [_logic, QGVAR(modeSelectiveSea), "modeSelectiveSea"] call EFUNC(common,readSettingFromModule);
+    // Disabled - Reference comment in FUNC(canChangeCamera)
+    //[_logic, QGVAR(modeSelectiveUAV), "modeSelectiveUAV"] call EFUNC(common,readSettingFromModule);
+
+    diag_log text format["[ACE]: View Restriction Module Initialized with mode: %1 (Foot: %2, Land: %3, Air: %4, Sea: %5, UAV: %6)", GVAR(mode), GVAR(modeSelectiveFoot), GVAR(modeSelectiveLand), GVAR(modeSelectiveAir), GVAR(modeSelectiveSea), /*GVAR(modeSelectiveUAV)*/ "N/A"];
```
jonpas: Use new log macro.

addons/viewrestriction/functions/fnc_moduleInit.sqf
```diff
@@ -0,0 +1,35 @@
...
+} else {
+    diag_log text format["[ACE]: View Restriction Module Initialized with mode: %1", GVAR(mode)];
```
jonpas: Same as above.

addons/viewrestriction/script_component.hpp
```diff
@@ -0,0 +1,12 @@
+#define COMPONENT viewrestriction
+#include "\z\ace\addons\main\script_mod.hpp"
+
```
jonpas: Add new debug stuff here.

jonpas:
As we decided this will go in 3.6.0 I reviewed my code again and made comments on new stuff that has to get added and some other minor things.

jonpas:
This is good to go from my side. Tagging @esteldunedain

Glowbal:
Looks good - but possibly a candidate for ACEX.

jonpas:
Candidate for ACEX. Closing.

PR #1724, opened 2015-06-22
Fixed CheckPBOs Module Whitelist defaultValue

jonpas:
And applied coding standards to `common/CfgVehicles.hpp`.


jonpas:
Just found out this also fixed `[ACE] ERROR: No array for function map in BIS_fnc_executeStackedEventHandler` coming from `fnc_checkPBOs.sqf`.

Issue #1722, opened 2015-06-22
CheckPBO doesn't kick if player joins with an ACE3 .pbo that was removed from server


BullHorn7:
Summoning @jonpas 

ACE3 Version: 3.1.1.1 official

**Mods:**
* @cba_a3
* @ace3
* EVERYTHING else :)

**Placed ACE3 Modules:**
* None

**Description:**
We had CheckPBO set to Warn Permanently and it seemed to work very well so we decided to switch to Kick.
Later we thought we noticed an issue with ACE_protection.pbo so we removed it from the server and continued testing. A few minutes later a new player that didn't know about the removal of the protection.pbo joined with that .pbo in his modpack but he wasn't kicked, nor did he receive any warning.

**Steps to reproduce:**
* Enable CheckPBO and set it to kick
* Remove some ACE .pbos from the server's modlist
* Join the server without removing the ACE .pbos from the client

**Where did the issue occur?**
* Multiplayer

**RPT log file:**
* Client: http://pastebin.com/Sfz4raw1
* Don't have access to server, sorry.

jonpas:
https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_moduleCheckPBOs.sqf#L23 is not even showing up in the RPT, investigating.

jonpas:
This is caused by https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_checkFiles.sqf

It will only check version and not if the file is actually missing.

SilentSpike:
I'm not sure there's a way to solve this without adding in a hardcoded list of ACE3 PBOs somewhere (which isn't desirable).

Shouldn't ARMA itself handle people connecting with addons that aren't running on the server?

jonpas:
> I'm not sure there's a way to solve this without adding in a hardcoded list of ACE3 PBOs somewhere (which isn't desirable).

No, it should just compare what's on the server and what's on the client, server is the base of comparing, client must be equal to server.

> Shouldn't ARMA itself handle people connecting with addons that aren't running on the server?

No Arma doesn't handle that.

On another thought, CheckPBO should check all PBOs, if they are same on server and connecting client, not sure why it doesn't for ACE's.

SilentSpike:
> No, it should just compare what's on the server and what's on the client, server is the base of comparing, client must be equal to server.

Well that's my point. If a PBO isn't on the server then you can't check if it's on the server using the PBOs that are on the server :stuck_out_tongue:

[...]

PR #1732, opened 2015-06-23
CheckPBO should ensure client isn't running any addons the server isn't

SilentSpike:
Fix for #1722

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -6,7 +6,7 @@
 //Singe PFEH to handle execNextFrame and waitAndExec:
 [{
     private ["_entry"];
-    
+
     //Handle the waitAndExec array:
     while {((count GVAR(waitAndExecArray)) > 0) && {((GVAR(waitAndExecArray) select 0) select 0) <= ACE_Time}} do {
         _entry = GVAR(waitAndExecArray) deleteAt 0;
--- a/addons/common/functions/fnc_checkFiles.sqf
+++ b/addons/common/functions/fnc_checkFiles.sqf
@@ -85,17 +85,16 @@ if (isMultiplayer) then {
                 };
             };
 
-            {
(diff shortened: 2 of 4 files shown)


SilentSpike:
As @CorruptedHeart points out in the issue, comparing these on the client isn't the most secure thing in the world.

Also I'm instinctively wary of PV'ing arrays. It might be more efficient to convert to string and back on each end.

jonpas:
This looks good to me.

bux578:
Should become more easy once `arrayIntersect` hits the floor

SilentSpike:
Well, not exactly as we'd need the set difference rather than the set intersection

bux578:
LINQ to SQF

esteldunedain:
> Well, not exactly as we'd need the set difference rather than the set intersection

You can probably compute the intersection and the use somehting like this:
```sqf
{
    _originalArray = _originalArray - [_x];
} forEach _intersectionArray;
```

SilentSpike:
Bad estse! :stuck_out_tongue_closed_eyes: 

```
_originalArray = _originalArray - _intersectionArray;
```

That works though, can just output them all as one message that way.

esteldunedain:
Ok, this seems ready to be merged. Please do once it's confirmed working.

SilentSpike:
@jonpas Any update on testing this?

jonpas:
@SilentSpike I didn't get a chance to test this, I'll do it later today.

SilentSpike:
Confirmed working, thanks again @jonpas :cookie:

jonpas:
Yay, have a :banana: !

BryanBoru:
Is this confirmed to be working? Can anyone walk me through is there is more to enabling it besides serverconfig.hpp settings?

jonpas:
@BryanBoru Yes, confirmed working, enabling it is all you have to do.

Issue #1234, opened 2015-05-15
Glitching items into the player moves them over the map


Raspu86:
CE3 Version: 3.0.0
**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *class ace_common_forceAllSettings {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_frag_enabled {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_interaction_EnableTeamManagement {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_magazinerepack_TimePerAmmo {
    value = 1.5;
    typeName = "SCALAR";
    force = 0;
};
class ace_magazinerepack_TimePerMagazine {
    value = 2;
    typeName = "SCALAR";
    force = 0;
};
class ace_magazinerepack_TimePerBeltLink {
    value = 8;
    typeName = "SCALAR";
    force = 0;
};
class ace_maptools_EveryoneCanDrawOnBriefing {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_microdagr_MapDataAvailable {
    value = 1;
    typeName = "SCALAR";
    force = 0;
};
class ace_mk6mortar_airResistanceEnabled {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_mk6mortar_allowComputerRangefinder {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_mk6mortar_allowCompass {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_nametags_showCursorTagForVehicles {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_nametags_PlayerNamesViewDistance {
    value = 5;
    typeName = "SCALAR";
    force = 0;
};
class ace_nametags_PlayerNamesMaxAlpha {
    value = 0.8;
    typeName = "SCALAR";
    force = 0;
};
class ace_respawn_SavePreDeathGear {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_respawn_RemoveDeadBodiesDisconnected {
    value = 1;
    typeName = "BOOL";
    force = 0;
};
class ace_respawn_BodyRemoveTimer {
    value = 0;
    typeName = "SCALAR";
    force = 0;
};
class ace_switchunits_EnableSwitchUnits {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
[...]

PR #1739, opened 2015-06-24
Virtual attach object

PabstMirror:
This is for attach, but the same idea could be applied to explosives which has had issues like
#1234 (Glitching items into the player moves them over the map).
Both systems want to show a virtual object when placing something.

Instead of `createVehicle` and updating it's pos every frame, this just show the p3d on the hud in the same position and scale as it would be if it were real.

Pros: Zero network traffic, No physX collisions

Cons: Looks bad in 3rd person (it's rendered separately from the main scene, so there is no z-buffer), also no lighting/nvg effects:


esteldunedain:
Great. I love this

PabstMirror:
Video: https://www.youtube.com/watch?v=iVuB8Z6OE7I
2nd half shows how it looks in 3rd person.  Like I said, it's bad.  I think the size and position of the object are right, but it should be rendered behind the player's head and obscured.

I tend to be "Function Over Form", but I'm not sure how critical of an issue it will be for others.  Given all of the issues with real objects, I think it's worth it.

If people are ok with this then I'll try doing the same for explosives.

esteldunedain:
> Given all of the issues with real objects, I think it's worth it.

I second that. @ViperMaul, @Glowbal, @KoffeinFlummi, I guess this is your call.

> If people are ok with this then I'll try doing the same for explosives.

I agree.

esteldunedain:
Hey, quick idea. For third person placement, we could offset the origin of the ray 30 cm to the right of the head, so it would be less likely the head obscure the plcing position

jonpas:
> I tend to be "Function Over Form", but I'm not sure how critical of an issue it will be for others.

Don't think it will be one at all, usually you put it down in a few seconds after selecting it anyways.

PabstMirror:
@esteldunedain - 30cm was a perfect offset, looks better now that it is not over the head.

untitled2

esteldunedain:
Yay! Over the shoulder third person action

SilentSpike:
What's the status on this?

PR #1753, opened 2015-06-27
Slideshow

jonpas:
Adds a tool for mission makers to apply interaction to presentational equipment such as whiteboards, laptops, screens and controllers such as keyboard, tablet, smartphone for switching the slides (if controllers are not available interaction with the object directly is used).

Has both, a mission module supporting both manual input with a comma as delimiter and synchronized objects (only for objects and controllers), and a public API function which takes same 4 parameters as module provides.

Also supports multiple different slideshows with both multiple modules or multiple public function uses.

- [x] Automatic slide transitions
- [x] Module icon
- [x] Add interactions dynamically
- [x] Fix issue with interaction menu closing itself #1862
- [x] Don't add `ACE_MainAction` if it already exists


ToasterBR:
THIS!
Would love to see this running :)

sargken:
Mabe next release?

alganthe:
I will try on a dedicated, see how it goes.

esteldunedain:
:+1: yay!

jonpas:
Would it be too performance overhead by adding a PFH to automatically slide through the slideshow if so enabled through the module/public function (would obviously be disabled by default)?

esteldunedain:
> Would it be too performance overhead by adding a PFH to automatically slide through the slideshow if so enabled through the module/public function (would obviously be disabled by default)?

Probably not, but if you use waitAndExecute instead of adding a custom pfh it would cost practically nothing.

jonpas:
> Probably not, but if you use waitAndExecute instead of adding a custom pfh it would cost practically nothing.

Ooh, great idea, I'll use that, thanks!

jonpas:
@esteldunedain I used a PFH, using `waitAndExecute` wouldn't work as it has to be continuous (unless I am missing something).

esteldunedain:
> unless I am missing something

At init you pass waitAndExecute a function to switch slides. And then you end the switch slides function with a new call to waitAndExecute.

Pseudo code:

```sqf
On init:
PREP(switchSlides);
[FUNC(switchSliders), 2] call EFUNC(common,waitAndExecute);

On fnc_switchSlides
// Switch slides
...
[FUNC(switchSliders), 2] call EFUNC(common,waitAndExecute);
```

jonpas:
Oooh, that's an interesting solution, thanks!

jonpas:
There, all tested and working. Added multiple modules support, replaced PFH with waitAndExecute and disabled debug.

esteldunedain:
Yay, looks good to me!

Glowbal:
I will look this over in the weekend.

jonpas:
Any update on this?

addons/slideshow/functions/fnc_createSlideshow.sqf
```diff
@@ -0,0 +1,75 @@
...
+
+// Formatted GVAR string (multiple modules support)
+private ["_varString"];
+_varString = str format [QGVAR(currentSlide%1), GVAR(slideshows)];
+
+// Set formatted GVAR to first slide
+missionNamespace setVariable [_varString, 0];
+
+// Automatic transitions handler
+[FUNC(autoTransition), [_objects, _images, _varString, _duration], _duration] call EFUNC(common,waitAndExecute);
```
Glowbal: Every slide show seems to have auto transition - Maybe this could be a setting for the created slide show?

addons/slideshow/functions/fnc_createSlideshow.sqf
```diff
@@ -0,0 +1,75 @@
...
+ * Public: Yes
+ */
+//#define DEBUG_MODE_FULL
+#include "script_component.hpp"
+
+PARAMS_5(_objects,_controllers,_images,_names,_duration);
+
+// Add controllers to objects if they support it
+{
+    if (typeOf _x in [CLASSNAMES_OBJECTS, CLASSNAMES_BOTH]) then {
```
Glowbal: I don't think hardcoding the classnames is the way to go. I'd prefer a method that's more dynamic so other content developers could very easily make a slide show object that would work with this, without anyone having to adjust the framework.

addons/slideshow/functions/fnc_createSlideshow.sqf
```diff
@@ -0,0 +1,75 @@
...
+        _objects pushBack _x;
+    };
+} forEach _controllers;
+
+// Objects synced to the module
+{
+    if (typeOf _x in [CLASSNAMES_OBJECTS, CLASSNAMES_BOTH]) then {
```
Glowbal: Same as above for this. Maybe a config attribute for the specific objects?

addons/slideshow/functions/fnc_createSlideshow.sqf
```diff
@@ -0,0 +1,75 @@
...
+
+// Exit if automatic transitions are disabled
+if (_duration == 0) exitWith {};
+
+// Number of slideshows (multiple modules support)
+GVAR(slideshows) = GVAR(slideshows) + 1;
+
+// Formatted GVAR string (multiple modules support)
+private ["_varString"];
+_varString = str format [QGVAR(currentSlide%1), GVAR(slideshows)];
```
Glowbal: You are using `str` on a `string` - doesn't seem necessary.

addons/slideshow/functions/fnc_makeList.sqf
```diff
@@ -0,0 +1,58 @@
...
+
+// Split using comma delimiter
+_splittedList = [_list, ","] call BIS_fnc_splitString;
+
+// Remove whitespace
+_listNoWhitespace = [];
+if (_removeWhitespace) then {
+    {
+        _x = [_x] call EFUNC(common,stringRemoveWhiteSpace);
+        _listNoWhitespace pushBack _x;
```
Glowbal: The above two lines could (should) be combined.

addons/slideshow/functions/fnc_moduleInit.sqf
```diff
@@ -0,0 +1,39 @@
...
+
+_logic = [_this, 0, objNull, [objNull]] call BIS_fnc_param;
+if (isNull _logic) exitWith {};
+
+// Extract variables from logic
+_objects = [_logic getVariable ["Objects", ""], true, true] call FUNC(makeList);
+_controllers = [_logic getVariable ["Controllers", ""], true, true] call FUNC(makeList);
+_images = [_logic getVariable ["Images", ""], true, false] call FUNC(makeList);
+_names = [_logic getVariable ["Names", ""], true, false] call FUNC(makeList);
+_duration = _logic getVariable ["Duration", ""];
```
Glowbal: The type of `Duration` is `NUMBER` - use a default value that matches this.

addons/slideshow/functions/fnc_moduleInit.sqf
```diff
@@ -0,0 +1,39 @@
...
+
+// Prepare with actions
+[_objects, _controllers, _images, _names, _duration] call FUNC(createSlideshow);
```
Glowbal: Please add validation of the variables here - check if there is even anything to do before we create the slide show;

- If there are no images listed, there is no point in creating the slide show
- If there is no object, same story.

addons/slideshow/CfgVehicles.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+            // Every object in this must be initialized with GVAR(Slides) Action applied, as well as 2 of below macros
+            sync[] = {CLASSNAMES_OBJECTS, CLASSNAMES_BOTH, CLASSNAMES_CONTROLLERS};
+
+            #define MACRO_SYNC \
+                optional = 1; \
+                duplicate = 1;
+            #define OBJECT description = "Object";
+            #define BOTH description = "Object AND Controller";
+            #define CONTROLLER description = "Controller";
```
Glowbal: The above 3 defines do not seem to be used.

addons/slideshow/functions/fnc_createSlideshow.sqf
jonpas: It is a setting, if duration is 0 (default), it will exit on line 62 of the above.

addons/slideshow/functions/fnc_createSlideshow.sqf
jonpas: I would as well, the problem is that most of the objects that slideshow even works on (as it requires `hiddenSelections` to change the texture) don't have ACE interaction enabled on them already. Is there some other way to go about that?

addons/slideshow/functions/fnc_createSlideshow.sqf
jonpas: Read above, requires ACE interaction which is not enabled by default on 99% of the object that slideshow looks good on.

addons/slideshow/functions/fnc_createSlideshow.sqf
jonpas: I will check that again, but I got some error with it otherwise (could be that I forgot Q in the macro at the time.

addons/slideshow/functions/fnc_moduleInit.sqf
jonpas: Objects and Controllers can be synced objects (addde to the array in `createSlideshow` function), images and names require input indeed.

addons/slideshow/CfgVehicles.hpp
jonpas: They are: https://github.com/jonpas/ACE3/blob/slideshow/addons/slideshow/CfgVehicles.hpp#L58-L70

Maybe a bit hard to notice. This is just for module "info" tab though, if a way to make it all dynamic (again ACE interaction enabling is required for the objects) then that wouldn't be needed at all.

Waffle-SS:
Screenshots, video, something? :3

jonpas:
Sure, go checkout the branch and make some. :)

addons/slideshow/CfgVehicles.hpp
Glowbal: Yes - sorry, my bad! Serves me for not using `ctrl+f`.

addons/slideshow/functions/fnc_createSlideshow.sqf
Glowbal: We can dynamically add the interaction items. The hiddenSelections we could probably check for in the config? If not - mission maker responsibility or we just define a config attribute.

Either way - let's not hardcode this.

jonpas: Interaction items yet, but what about the interaction menu itself? It requires XEH on the object, which none of those have enabled by default. Am I missing something again?

jonpas:
Updating base class errors on RC 1.48, thanks @ViperMaul 
```
11:20:02 Updating base class Furniture_base_F->ThingX, by z\ace\addons\sitting\config.bin/CfgVehicles/Land_ChairPlastic_F/
11:20:02 Updating base class Furniture_base_F->ThingX, by z\ace\addons\sitting\config.bin/CfgVehicles/Land_ChairWood_F/
11:20:02 Updating base class Furniture_base_F->ThingX, by z\ace\addons\sitting\config.bin/CfgVehicles/Land_OfficeChair_01_F/
11:20:02 Updating base class Furniture_base_F->ThingX, by z\ace\addons\sitting\config.bin/CfgVehicles/Land_RattanChair_01_F/
```

jonpas:
Everything finished, ready from my side when #1862 gets merged.

esteldunedain:
Done

jonpas:
Thank you!

esteldunedain:
Nah, thank you!

PR #1757, opened 2015-06-28
Viewdistance settingsInitialized

Winter259:
It now waits for settings to be something before it tries to do something.
Not sure why github thinks line 11 is enormous. This is what it looks like locally: https://imgur.com/p8NZ4d4
Would appreciate a sanity test that this works before merging.

### DIFF
--- a/addons/viewdistance/XEH_postInit.sqf
+++ b/addons/viewdistance/XEH_postInit.sqf
@@ -2,28 +2,33 @@
 
 if (!hasInterface) exitWith {};
 
-// Force the view distance down to the limit.
-if (GVAR(enabled) && {viewDistance > GVAR(limitViewDistance)}) then {
-    setViewDistance GVAR(limitViewDistance);
-};
-
-// Adapt view distance when the player is created or changed according to whether client is on foot or vehicle.
-["playerChanged",{
-    [false] call FUNC(adaptViewDistance);
-}] call ace_common_fnc_addEventHandler;
-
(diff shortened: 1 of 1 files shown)


Winter259:
Yes I edit in N++, don't judge me.

nsgundy:
Is there any chance for the `SettingsInitialized` event to be triggered more than once? If so, there should be some protection put in place, otherwise we will end up with multiple instances of the same EHs.

jonpas:
> Is there any chance for the SettingsInitialized event to be triggered more than once? If so, there should be some protection put in place, otherwise we will end up with multiple instances of the same EHs.

It gets called here (as local event): https://github.com/acemod/ACE3/blob/master/addons/common/XEH_postInit.sqf#L138, it only gets triggered once.

Winter259:
@jonpas so I only need to put if ! GVAR(enabled)  exitWith {}; inside the settingsInitialized?

jonpas:
And the following I guess:
```if (GVAR(enabled) && {viewDistance > GVAR(limitViewDistance)}) then {
    setViewDistance GVAR(limitViewDistance);
};```

Winter259:
Wouldn't it make sense to not apply the EHs in the first place if the module is not enabled though?

jonpas:
Hmh that's true indeed. I didn't take into account that it can be disabled fully!

Winter259:
Depends whether we want enabled to be "don't run the module at all" or "don't allow the module to operate"

esteldunedain:
:+1: Looks good to me.

PR #1759, opened 2015-06-28
Exit addEarplugs if earplugs are used

jonpas:
This solves duplicating Earplugs in persistence scenarios (eg. persistence mods). It will still add the earplugs if none are equipped, but in terms of persistence you would normally equip them and be done with it anyways.


addons/hearing/functions/fnc_addEarPlugs.sqf
```diff
@@ -15,9 +15,12 @@
  */
 #include "script_component.hpp"
 
-private ["_unit", "_launcher"];
+PARAMS_1(_unit);
 
-_unit = _this select 0;
+// Exit if soldier has earplugs already in (persistence scenarios)
+if (FUNC(hasEarPlugsIn)) exitWith {};
```
PabstMirror: "Error if: Type code, expected Bool"
probably should be:
`if ([_unit] call FUNC(hasEarPlugsIn)) exitWith {};`

jonpas: Oh crap, forgot to save the file! :grin:

alganthe:
wouldn't removing earplugs on respawn be a better solution ?

jonpas:
> wouldn't removing earplugs on respawn be a better solution ?

You mean by a mission/persistence system used? Yes of course, but this is not the only point of the PR. Right now if you have earplugs in already it will add another, which you'd normally just throw on the ground.

jonpas:
Would it be ok to add an ACE_Settings setting to disable that automatic addition?

esteldunedain:
@jonpas, please resolve the conflict generated by #1763. Besides that I think this is ready to be merged.

jonpas:
Resolved conflicts and combined both PRs for performance.

esteldunedain:
Thanks!

PR #1760, opened 2015-06-29
ACE_Spectator

voiperr:
Voiper's ported spectator suite. Initial version; changes/upgrades to come.


SilentSpike:
Exciting stuff :+1:

Will take a more thorough look through it a bit later.

addons/spectator/functions/fnc_camera.sqf
```diff
@@ -0,0 +1,1081 @@
...
+        
+        _camPos set [2, (_camPos select 2) + 2];
+        _cam = "camera" camCreate _camPos;
+        _cam setDir _camDir;
+        _cam cameraEffect ["internal", "back"];
+        _cam camSetFocus [-1, -1];
+        _cam camCommit 0;
+        showCinemaBorder false;
+        cameraEffectEnableHUD true;
+        setViewDistance 3000;
```
Cuel: save original VD and set it back once spec is exited?

addons/spectator/stringtable.xml
```diff
@@ -0,0 +1,110 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project name="ACE">
+    <Package name="Spectator">
+        <Key ID="STR_ACE_Spectator_Module_DisplayName">
+            <English>Spectator</English>
+            <Polish></Polish>
```
PabstMirror: Remove lines that aren't translated.  (just leave the English lines)

addons/spectator/XEH_postInit.sqf
```diff
@@ -0,0 +1,96 @@
...
+            if (player distance GVAR(penPos) < 200) then {
+                if ({isPlayer _x && alive _x && (_x distance GVAR(penPos)) > 200} count allUnits == 0) then {
+                    [["endDeath", false], "BIS_fnc_endMission"] call BIS_fnc_MP;
+                    [_this select 1] call CBA_fnc_removePerFrameHandler;
+                };
+            };
+        }, 2] call CBA_fnc_addPerFrameHandler;
+    };
+    
+    if (isDedicated) exitWith {};
```
SilentSpike: Should be `if !(hasInterface)`

addons/spectator/functions/fnc_camera.sqf
esteldunedain: It would probably be best not to mess with view distances at all, people may want more or less than 3000 depending on their systems.

addons/spectator/XEH_postInit.sqf
```diff
@@ -0,0 +1,96 @@
...
+        [player] joinSilent grpNull;
+        if (isClass (configFile >> "CfgPatches" >> "ace_hearing")) then {EGVAR(hearing,disableVolumeUpdate) = true};
+        _delay = getNumber (missionConfigFile >> "respawnDelay");
+        _delay fadeSound 0;
+        999999 cutText ["", "BLACK", _delay];
+    }];
+
+    player addEventHandler ["Respawn", {
+        if (!isNil QGVAR(cam)) then {["Exit"] call FUNC(camera)};
+        if (isClass (configFile >> "CfgPatches" >> "ace_hearing")) then {EGVAR(hearing,disableVolumeUpdate) = true};
```
esteldunedain: I think it should be `EGVAR(hearing,disableVolumeUpdate) = false` here

addons/spectator/functions/fnc_camera.sqf
Cuel: From what I saw in the code, there's a slider to change the VD. Do you mean that it should be removed completely?

What I meant was that for those situations say e.g PvP with respawns (players can exit spec) - the VD should be forced back to whatever it was before entering spec to keep it fair

Changing VD slider could be an option for mission makers as well.

esteldunedain: > From what I saw in the code, there's a slider to change the viewdistance. 

Oh, I missed that

> Do you mean that it should be removed completely?

IDK, is it that important to quickly change view distanced while spectating? Maybe a better solution would be to default the slider to the current distance, and not ever modify it unless the user actually drags the slider.

Cuel: Yeah I'd suggest that it's disabled by default and let the mission makers allow it through a module option for those cases where changing VD is 'okay'

esteldunedain:
Oh, I'm really exciting about having a good spectator script in ACE!

I can't tell how heavy the interfase for this is, but it would probably be welcome to have a key that hides/shows any interfase. That would probably be very appreciated by streamers and youtubers.

SilentSpike:
Also very excited by this!

There's some code in here I think will need some work, but it looks like a very good start.

esteldunedain:
I just realized this is the code used by UO @jaynus mentioned once (or at least a derivation of that). Great stuff.

A video from @voiperr of the UO version:

https://www.youtube.com/watch?t=20&v=HWrZyRpKE0M

voiperr:
The interface and the 3D/map icons are already toggleable. :innocent:

esteldunedain:
> The interface and the 3D/map icons are already toggleable. 

Excellent, I hope we can merge this soon.

SilentSpike:
While I remember: Will need to add a canInteractWithCondition `isNotSpectating` (or similar) to prevent the interaction menu while spectator is open. (Could also potentially add some cool spectator specific actions to perform spectator actions)

esteldunedain:
> Will need to add a canInteractWithCondition isNotSpectating (or similar) to prevent the interaction menu while spectator is open.

True.

> Will need to add a canInteractWithCondition isNotSpectating (or similar) to prevent the interaction menu while spectator is open.

Yeah, maybe in the future. I can think of any right now.

SilentSpike:
>Yeah, maybe in the future. I can think of any right now.

Ofc :+1:

addons/spectator/XEH_postInit.sqf
```diff
@@ -0,0 +1,96 @@
...
+            };
+        }, 2] call CBA_fnc_addPerFrameHandler;
+    };
+    
+    if (isDedicated) exitWith {};
+    
+    call FUNC(penPos);
+
+    {
+        if (getMarkerPos _x isEqualTo [0,0,0]) then {
```
SilentSpike: Assuming this code will stay the same (I have some ideas for future iterations) it should probably become `!(_x in allMapMarkers)`

addons/spectator/XEH_postInit.sqf
```diff
@@ -0,0 +1,96 @@
...
+            };
+        };
+        
+        player addEventHandler ["HandleDamage", {0}];
+        [player] joinSilent grpNull;
+        removeAllWeapons player;
+        removeAllItems player;
+        removeAllAssignedItems player;
+        removeUniform player;
+        removeVest player;
```
SilentSpike: I probably wouldn't mess with the player's gear at all when they become a spectator (as it could produce unexpected outcomes in other scripts)

SilentSpike:
Hey, I don't want to overwhelm you with feedback so I'm going to leave it there for now.

I have some flexibility improvements in mind for usage in missions with varying respawn setups. Those can all wait until the base framework is ready though. :+1:

addons/spectator/XEH_postInit.sqf
voiperr: Currently, it's set up to only work on one-time deaths only. EGVAR(hearing,disableVolumeUpdate) resets on respawn, and this respawn EH exists to restart the camera in case the player hits "Respawn" in the pause menu.

addons/spectator/functions/fnc_camera.sqf
```diff
@@ -0,0 +1,1081 @@
...
+        preloadTitleRsc [QGVAR(help), "PLAIN", 0, true];
+        
+        if (isClass (configFile >> "CfgPatches" >> "ace_nametags")) then {
+            GVAR(tags) = [EGVAR(nametags,showPlayerNames), EGVAR(nametags,showNamesForAI)];
+            EGVAR(nametags,showPlayerNames) = 0;
+            EGVAR(nametags,showNamesForAI) = false;
+        };
+        
+        if (isClass (configFile >> "CfgPatches" >> "ace_interact_menu")) then {
+            [QGVAR(interactCondition), {false}] call EFUNC(common,addCanInteractWithCondition);
```
voiperr: > Will need to add a canInteractWithCondition isNotSpectating (or similar) to prevent the interaction menu while spectator is open.
Is this enough? It just disables all interactions entirely. I feel using interaction while in a floating camera would be weird UX.

>  have some flexibility improvements in mind for usage in missions with varying respawn setups.
We'll have to deliberate on those pretty fiercely, because the hacky way it all sets up (pens, invisible players) will likely have to change.

addons/spectator/XEH_postInit.sqf
voiperr: Fair enough. Unfortunately, the vest and the goggles are necessary, because on ocean maps the player respawns over water, and vision blur/drowning occurs without those items.

addons/spectator/functions/fnc_camera.sqf
jonpas: I think a much better way to handle this is with exceptions like SilentSpike said above.

esteldunedain:
Ok, let's do this. I'll merge this an create a separate ticket with the modifications and fixes proposed above. That way we can all submit the modification collaboratively, addressing that ticket number on the commits.

bux578:
nice!

esteldunedain:
Done, please continue discussion (and fixes) on #1781.

Thank you very much @voiperr, gj!

PR #1762, opened 2015-06-29
Wake up from unconscious even without epi after a given amount of time

Winter259:
Pull requesting to see if there is interest.

In ACE2 a player would wake up after a (if I remember correctly) randomised amount of time with a lower limit, regardless of whether they have given Epinephrine.

I've written a very basic version of the above in this PR to see whether any of the team have any alternate implementations in mind. If not, I will finish doing the TODOs and update the PR for merge consideration.

I plan to make a dynamic option for _make_wake_up_time, the 300 seconds is just an example.


Kilo1-1:
I would love to see this.  AGM had this where the player would eventually wake up from consciousness after a certain amount of time.  You would wake up from conscious much faster if you were stabilized with bandages.  As it is now, you can fall from a small height and become unconscious and not wake up with basic medical, so this is a feature that I think would greatly enhance basic medical.

addons/medical/ACE_Settings.hpp
```diff
@@ -59,6 +59,16 @@ class ACE_Settings {
         typeName = "SCALAR";
         values[] = {"Disabled", "50/50", "Enabled"};
     };
+    class GVAR(autoWakeUp) {
```
Glowbal: This setting is not necessary, simply have the wake up time of `-1` disable this

addons/medical/ACE_Settings.hpp
```diff
@@ -59,6 +59,16 @@ class ACE_Settings {
...
+        typeName = "BOOL";
+        value = 1; // default 0, left 1 for testing
+        isClientSettable = 0;
+    };
+    class GVAR(autoWakeUpTime) {
+        typeName = "SCALAR";
```
Glowbal: Please add displayName and descriptions to the setting.

addons/medical/functions/fnc_unconsciousPFH.sqf
```diff
@@ -48,6 +48,10 @@ if (!alive _unit) exitwith {
     [(_this select 1)] call cba_fnc_removePerFrameHandler;
 };
 
+if (GVAR(autoWakeUp) && {(GVAR(level) == 1) && ((ACE_time - _startingTime) >= GVAR(autoWakeUpTime))}) then { // allow only if its basic medical
```
Glowbal: Move this if statement to line 141, just above the last if statement

addons/medical/ACE_Settings.hpp
Glowbal: Also please set the value to `-1` as default. And I'd suggest adjusting the setting name to:
`maxUnconsciousTime`

addons/medical/functions/fnc_unconsciousPFH.sqf
Glowbal: This does not respect the revive system:
```sqf
_unit getvariable [QGVAR(inReviveState), false]
```

Glowbal:
This is also missing the module entries.

gienkov:
> In ACE2 a player would wake up after a (if I remember correctly) randomised amount of time with a lower limit, regardless of whether they have given Epinephrine.
Nope, you would wake up in ACE2 if your wounds were not fatal enough. If you could bleed long enough then your vital status would **not** allow you to wake up regardless of anything, except being treated by medic. There was also option "Stabilize patient" but I don't remember what that was doing to casualty.

addons/medical/functions/fnc_moduleMedicalSettings.sqf
```diff
@@ -34,4 +34,6 @@ if !(_activated) exitWith {};
 [_logic, QGVAR(preventInstaDeath), "preventInstaDeath"] call EFUNC(common,readSettingFromModule);
 [_logic, QGVAR(bleedingCoefficient), "bleedingCoefficient"] call EFUNC(common,readSettingFromModule);
 [_logic, QGVAR(painCoefficient), "painCoefficient"] call EFUNC(common,readSettingFromModule);
+[_logic, QGVAR(maxUnconsciousTime), "keepLocalSettingsSynced"] call EFUNC(common,readSettingFromModule);
```
Glowbal: Incorrect name for the setting here.

addons/medical/functions/fnc_unconsciousPFH.sqf
```diff
@@ -135,6 +135,19 @@ if (!local _unit) exitwith {
     [(_this select 1)] call cba_fnc_removePerFrameHandler;
 };
 
+if (GVAR(maxUnconsciousTime) > 0 && {!(_unit getvariable [QGVAR(inReviveState), false])} && {(GVAR(level) == 1)}) then {
+    private ["_maxUnitUnconsciousTime"];
+    _maxUnitUnconsciousTime = if ((_unit getvariable [QGVAR(maxDynamicUnconsciousTime), 0]) > 0) then {
```
Glowbal: What's this dynamic unconscious time? I don't think that's necessary.

Also the above if statement could be combined with the one on line 146.

addons/medical/CfgVehicles.hpp
```diff
@@ -86,6 +86,18 @@ class CfgVehicles {
                 typeName = "NUMBER";
                 defaultValue = 1;
             };
+            class maxUnconsciousTime {
+                displayName = "Base Maximum Unconscious Time";
```
Glowbal: This should be in a stringtable

addons/medical/CfgVehicles.hpp
```diff
@@ -86,6 +86,18 @@ class CfgVehicles {
...
+                description = "The base amount of time that a unit will spend unconscious before waking up without medical intervention."; // TODO: Stringtable
+                typeName = "NUMBER";
+                defaultValue = 10; // testing value: 10
+            };
+            class maxDynamicUnconsciousTime {
```
Glowbal: I don't see a good enough use for something like this to justify adding another setting for this. I feel it would only make things more confusing for users.

addons/medical/ACE_Settings.hpp
```diff
@@ -59,6 +59,20 @@ class ACE_Settings {
         typeName = "SCALAR";
         values[] = {"Disabled", "50/50", "Enabled"};
     };
+    class GVAR(maxUnconsciousTime) {
+        displayName = "Base Maximum Unconscious Time";
```
Glowbal: This should be in a stringtable

addons/medical/CfgVehicles.hpp
```diff
@@ -86,6 +86,18 @@ class CfgVehicles {
                 typeName = "NUMBER";
                 defaultValue = 1;
             };
+            class maxUnconsciousTime {
+                displayName = "Base Maximum Unconscious Time";
+                description = "The base amount of time that a unit will spend unconscious before waking up without medical intervention."; // TODO: Stringtable
+                typeName = "NUMBER";
+                defaultValue = 10; // testing value: 10
```
Glowbal: This defaultValue should be the same as the value in the setting from ACE_Settings

addons/medical/functions/fnc_unconsciousPFH.sqf
Winter259: It was a request from the person who asked me to do the adjustment - that it will not be one set value for all the time to the point where if you know that its five minutes long and epi is not available, then you know you have enough time to go to the toilet.

addons/medical/ACE_Settings.hpp
```diff
@@ -59,6 +59,20 @@ class ACE_Settings {
...
+        description = "Time required (in seconds) before an unconscious unit will wake up without epinephrine."; // TODO: Needs stringtabling.
+        typeName = "SCALAR";
+        value = 300; // -1 for disabled.
```
Glowbal: This feature should be disabled by default.

addons/medical/CfgVehicles.hpp
Winter259: Considering default in ACE_Settings is -1, i.e. disabled, shouldn't the module default option be something different, like 300 as an example?

Winter259:
I'll remove testing values & add stringtable lines once I finish the features.

Winter259:
Despite working off master's current medical stringtable.xml, whenever I add new ones it creates a merge conflict for some reason.

Winter259:
Please ref #1803 instead.

PR #1763, opened 2015-06-29
Update fnc_addEarPlugs.sqf

naizarak:
Do not automatically add earplugs to units if combat deafness is disabled.


jonpas:
Can we add that to #1759 and some lazy evaluation to keep it clean?

Issue #1736, opened 2015-06-24
Map can't be scrolled through or zoomed while player is moving


BullHorn7:
This video is from 3.0.1 but tested and confirmed in 3.1.1.1 too. In-map FPS drops to 5-20 when moving as well.

https://youtu.be/82A6S2zhxCw?t=8m52s

Steps to reproduce:
* Open map
* Start moving
* Attempt to zoom in/out and drag the map to any other side

esteldunedain:
> Attempt to zoom in/out and drag the map to any other side

It's a known issue, which is not fixable with mapShake on. We can discuss if map shake should be enabled by default or not though

Tenga:
Map shake also makes it extremely difficult to manipulate the map when you are swimming but not moving. 

Sure, you can probably argue realism, but it's a usability nightmare. You have no feedback as to what's going on as the on sea movements are very minor and the map shake can't be seem, the map just randomly locks up or makes sudden jumps.

This can probably be solved by introducing (or increasing, if it already exists) the minimum amount movement needed for the map shake to trigger.

jonpas:
Or add an exception `isNotSwimming` and don't move it in that case.

Tenga:
@jonpas Wouldn't that prevent it from shaking at all if you're swimming, even when you **are** moving? I'd still shake while swimming and moving, just not when you're still (waves cause micro movements, hence the problem).

jonpas:
Yes it would, but on that note I'd prefer to totally prevent opening it when swimming. How can you look at the map when swimming? It's already almost impossible when just being still in the water (not to mention you'd get water all over the map).

Tenga:
I'm not sure if there are things like laminated / waterproof maps and map cases for such use. 

Considering waterproof cases for gadgets already exist, I don't think it's a stretch to assume something similar exists for maps.

jonpas:
Yep, I am of same thoughts.

SilentSpike:
>I'm not sure if there are things like laminated / waterproof maps and map cases for such use.

Yep, they definitely exist.

esteldunedain:
Fixed by the wonderfull @PabstMirror

jonpas:
:+1:

brainslush:
I actually thought this was a feature! Maybe make a module out of it?

PR #1769, opened 2015-07-01
Allow moving map while shake effect active

PabstMirror:
#1736 (Map can't be scrolled through or zoomed while player is moving)

Allow panning the lastStillPosition while mapShake is active.
Map doesn't stop shaking, just changes the "origin" that it shakes around.

I would be fine moving this to an option (like 3 settings for shake (off, on /w panning, and current-mode).


esteldunedain:
Genius!

esteldunedain:
Couldn't resist merging this, thanks!

PR #1783, opened 2015-07-03
Adding hideUnit functions to ace_common

SilentSpike:
The newly added ace_spectator needs to make use of `hideObjectGlobal` (which must be executed on the server to work).

I figured a small common system to do such a thing could be useful in ace and took cues from the similar setForceWalkStatus and muteUnit systems.


esteldunedain:
:+1:

jonpas:
Handy!

PabstMirror:
`isObjectHidden` is only on dev branch right now, so this is throwing an error at mission start
1.48 should be out soon enough though, so it's might be ok to leave for a few days

SilentSpike:
>only on dev branch right now

It might actually be best to remove that condition anyway since the object could be only hidden locally.

PR #1800, opened 2015-07-04
Small Improvement in Nametags

jokoho48:


SilentSpike:
Seems trivial to me, merging

Glowbal:
Actually I am wondering besides readability if this has any performance gains? Did you do a test?

jokoho48:
at 1000000 this is the results
Old: Frames: 2155 Time: 35.917
New: Frames: 1329 Time: 22.151
TestCode: http://pastebin.com/WFZ2tcXL

jonpas:
That's pretty nice!

jokoho48:
thats the reason why i create this PR and i think thats not so Trivial than it looks like

PR #1809, opened 2015-07-06
Fix for 'Treat at...' option

SzwedzikPL:
Fix for http://i.imgur.com/XEUhwGC.jpg


jonpas:
This can safely be merged with recent medical vanilla damage fixes I believe.

PR #1813, opened 2015-07-07
Tagging (buildings, walls, etc.)

BaerMitUmlaut:
This PR adds the possibility to tag buildings and other static objects with spray paint. It also adds a can of spray paint that is needed for tagging.

Demonstration: https://www.youtube.com/watch?v=U3I0_jDlFvA

The tags get aligned to the objects by doing a sort of hitdetection. This works great on most A3 buildings, but is not guaranteed to work on all buildings. It mainly depends on the precision of the hitbox (I'm not sure what LOD is used for lineIntersects, but I assume it's either ViewGeometry or FireGeometry).

~~TODO: Add custom spray paint object from @jokoho48 when it's done.~~ Done!


Legolasindar:
You can apply this to the terrain also?

BaerMitUmlaut:
Sure, terrain would actually be a lot easier. I don't think spray paint would work well on grass (both in real life and in game), but it could work on roads.

I don't really see a purpose for that, but I'm sure you have something in mind?

Legolasindar:
The soldiers make marks on the floor with spray to indicate where it is safe to pass, where there are no mines. Sprays generally serve to make anything you want to highlight. A door of a room cleaned. One lane where it is safe to walk, etc ...

BaerMitUmlaut:
A video or a picture would be nice. Should I just use the same black crosses for that or something different?

Legolasindar:
Is different, i see one line in vertical in purple for example. I cant find pictures or videos :(

alganthe:
merge conflicts present.

BaerMitUmlaut:
I'm

BaerMitUmlaut:
The merge conflict has been resolved.

gienkov:
It should do "spraying sound" dont you think? Its nice anyway, useful definitely.

Legolasindar:
Some samples.
http://www.freesound.org/people/bdunis4/sounds/271461/
http://www.freesound.org/people/kbnevel/sounds/119860/
http://www.freesound.org/people/stephendemaria/sounds/155519/
https://www.freesound.org/people/therentcollector/sounds/90187/
http://www.freesound.org/people/zerolagtime/sounds/231550/
https://www.youtube.com/watch?v=AjBnAvxWp5E

BaerMitUmlaut:
Playing a sound when tagging seems like a good idea. An animation would be cool, too, but I don't think we have something suitable for that (if you know of one let me know).

As for the ground tagging, for a differently coloured tag I would have to add another variant of the can of spray paint I suppose. Should I do that and if so, should I also add more colours (red, blue, green)?

gienkov:
Diversity is always good. I don't think there is animation suitable for
that. Maybe quick progress bar when tagging too? So it doesnt show up
instantly but after a second or two.
9 lip 2015 22:46 "BaerMitUmlaut" <notifications@github.com> napisał(a):

esteldunedain:
>  An animation would be cool, too, but I don't think we have something suitable for that (if you know of one let me know).

I suggest you use the "PutDown" animation. It's simple and quick and should look reasonable good:

```sqf
_player playActionNow "PutDown";
```

Glowbal:
> As for the ground tagging, for a differently coloured tag I would have to add another variant of the can of spray paint I suppose. Should I do that and if so, should I also add more colours (red, blue, green)?

Yes, different colours would be great.

Before we get this implemented I'd like some extensive testing done. How would this work on doors? What happens if the building is destroyed?

Legolasindar:
I created this spray sound in OGG for PutDown animation with 2 seconds.
www.cavallersdelcel.cat/magatzem/ArmA/edicio/addons/ace3/addons/esprai/sprai_sound_01.ogg

BaerMitUmlaut:
>How would this work on doors? What happens if the building is destroyed?
Doors can be tagged, but if you open the door again, the tag will just float in the air. I didn't find a way to prevent this unfortunately.
I should add a destroyed event handler on tagged buildings, else they might float in the air.

BaerMitUmlaut:
Multiple colors for the tags have been added, as well as the possibility to tag the ground. There's also an animation and a sound getting played now. The sound is a tweaked version of the one from @Legolasindar.

BaerMitUmlaut:
I've updated this to use the new lineInteresectsSurfaces and params commands. This allowed me to completly save the loop which ran 12 times when placing a tag and I'm getting a more precise position of the surface.

However, due to the FIRE LOD not being 100% precise it can still happen that a tag appears inside a wall - I have adjusted it so the tag gets placed 6cm in front of the surface which is enough for the majority of the walls.

commy2:
FIRE + GEO lod?

BaerMitUmlaut:
Done.

ferox113:
For tagging a clear path, here is a general idea of how we as sweepers (combat engineers/EOD techs) do things in terms of tagging: https://youtu.be/bJ0vj7m9K3s?t=6m47s

I do not know how you have made the animations/textures, but a simple vertical line would do the trick in terms of marking the route. If a POSIED/some ordnance has been found during an advance, the sweeper will generally mark the passage (a good distance from the explosive ordnance), but keep momentum -if in a spot where there is no sufficient cover, but to the objective they are headed for.

I would assume that when identifying a wall, for the tagging to become available in the interaction menu, one can trigger the "X" texture for this specific action on the object.

If, in other hands, there is no wall (or object), the " I " texture would be applied.

Is this possible?

BaerMitUmlaut:
This is absolutely possible. Currently, it's only possible to create X marks. This PR needs to be updated anyways though and if you have more suggestions for this, go ahead :+1:

PabstMirror:
Effect is pretty good.  No flickering or floating/paralax on texture.

addons/tagging/functions/fnc_tagGround.sqf
```diff
@@ -0,0 +1,53 @@
...
+
+#include "script_component.hpp"
+private ["_tagPos", "_groundPos", "_vectorDirAndUp"];
+
+params ["_color"];
+if !((toLower _color) in ["black", "red", "green", "blue"]) exitWith {
+    ["%1 is not a valid tag colour.", _color] call BIS_fnc_error;
+};
+
+_tagPos = player modelToWorld [0, 1.2, 0];
```
commy2: player

Glowbal:
Closing in favour of  #3210

Issue #1818, opened 2015-07-08
Mortar charge overlay persists


Waffle-SS:
ACE3 Version: 3.1.2

**Mods:**
* @cba_a3
* @ace3

**Description:**
The Mk6 Mortar charge UI overlay isn't removed on mission restart/abort in editor. When you restart the charge overlay can still be seen. Reloading the mission in editor, or getting in and out of the mortar fixes the issue.

Image: https://dl.dropboxusercontent.com/u/95262179/ArmA/ArmA%203/ACE3/arma3%202015-07-07%2023-47-15-16.png

**Steps to reproduce:**
1. Get in Mk6 Mortar
2. Restart/abort mission

**Where did the issue occur?**
SP

**RPT log file:**
No scripts/config errors

PabstMirror:
I have no idea why I was trying to use `ctrlCreate`, but it will be fixed with #1820

Issue #1498, opened 2015-06-04
Aiming with the Mk6 Mortar digital compass is inaccurate unless on perfectly flat ground


VKing6:
I talked about this in Slack but I forgot to ticket it.

Unless the Mk6 mortar is on flat ground the digital compass will be inaccurate when firing without using the artillery computer. The cant doesn't appear to affect the fired round because of some BIS magic or other.

More testing is required to determine if the round will land where expected if the compass error is corrected for.

alganthe:
This is a BI issue and has been fixed on dev branch

GitHawk:
@alganthe Can you post a link to dev branch change log?
I believe I have a similar problem with Artillery.

ulteq:
> This is a BI issue and has been fixed on dev branch
Are deployed rfiles also affected?

GitHawk:
@ulteq I believe they are see #890 Artillery is affected aswell - notice the difference between cannon and tracer angles.
![107410_screenshots_2015-06-04_00011](https://cloud.githubusercontent.com/assets/1816059/8021634/942ce76c-0ca7-11e5-8f06-5898f9a99c5e.jpg)

esteldunedain:
My god BIS

VKing6:
> This is a BI issue and has been fixed on dev branch

Unfortunately it appears it hasn't.
http://i.imgur.com/TdryHkD.jpg
http://i.imgur.com/rVcGsfn.jpg
http://i.imgur.com/gci9PpW.jpg

ulteq:
> My god BIS

:disappointed:

austinjames314:
@VKing6 I can confirm that if (if!) you comensate for the azimuth and elevation offset (yes elevation too!) introduced by the angle and direction of the ground's slope, then you can get rounds on target. See Codings Artillery Computer, a mod which, among other things, provides the angle added by the ground's slope, so that you can apply that offset. (Except that it gives it in degrees, making it pretty useless for me).
 Ideally of course, the mortar sight would just give you the _actual_ azimuth and elevation of the barrel, relative to the coordinate system, not relative to the piece of terrain you happen to be set up on.

VKing6:
Heh ideally the model should animate so the tube could be upright even if the base is tilted (like on the real things!), but I'll check that method out.

A mission maker from my community said he was going to try to attach the mortar to a game logic to level it. I have to ask him how that experiment went.

austinjames314:
[...]

Issue #1814, opened 2015-07-08
Mk6 Mortar charge (firing mode) resets


HTSean:
I've been messing with the Mk6 mortar and I'm loving it. There's just two issues though:

1) The charge resets when you leave the vehicle. 
It'd be great if it remembered what charge you had selected.

2) I believe there's a way to somewhat stabilize actual mortars so you can fire from slightly uneven ground. 
I'd love to see this implemented so that it can be used from more positions (I think ACE2 had it, if I remember correctly).

If you guys don't mind, I'd also like to ask if you guys are planning on adding more mortars like the M224 and M120, including ye olde ACE2 aiming method and round prepping and such?

Keep up the awesome work, guys.

austinjames314:
The mortars not firing accurately when on ground other than perfectly flat ground is discussed at length in #1498. 

You're right, the charge does seem to reset when you get in and out. I've taken it as just something to watch out for, but it would be nice if it were fixed.

If you've been using the mortar a lot, have you noticed problems with the smoke rounds not following the ranges as per the rangetable? No one else around here has responded to my mentioning that, and it would be nice to see if someone can confirm I'm not crazy.

HTSean:
Oh yeah, I just tried that out and you're right. I fired a couple HE and a couple smoke using the same ELEV and such and this was the result:


![](http://i.gyazo.com/c24c571e40daf78c9940010925444b6d.jpg)

alganthe:
smoke shells being lighter than HE wind gusts might influence them a lot more.

HTSean:
Yeah that makes sense. It'd be great to have a separate range table for smokes.

austinjames314:
Yeah that's what I've requested before in another issue. But I might make a new one and link to this. Thanks for that! I never had enough people to help me test and get pictures like that.

austinjames314:
Also @alganthe, I'm pretty sure it's not crosswind. In my experience the smoke rounds land about 5-10% _farther away_ than the HE rounds.

Waffle-SS:
I sort of question the strength of these random wind gusts...
I've had HE mortar rounds land 500+m off a 1km away target due to wind gusts...

HTSean:
You sure you were on flat ground then?

austinjames314:
Yeah that sounds like you were on uneven ground. Wind usually only affects things by a few meters at 1km. see #1820

PabstMirror:
Will now save/restore last charge setting
[...]

PR #1820, opened 2015-07-08
Mk6 fix leveling

PabstMirror:
#1498
#1818

* Show leveled elevation/azimuth.
* Fix some UI stuff

WIP - still need to do some more testing, but should be ready for 3.2


austinjames314:
Awesome! Which method was used? Does this make the mortar stay level regardless of terrain, or make the azimuth and elevation show the true values for a tilted mortar?

PabstMirror:
It shows the elevation that the gun will shoot at (what you see is what you get).
It's kind of hard to show 3d vectors in screenshots, but if you look below.
Red is how vanilla says it will shoot, blue is the weapon direction and yellow is how it actually shoots.

One fun bit of BIS logig: if the gun isn't looking at anything it will shoot via weapon direction.
So if you aim near the top of a mountain and shoot, then move your cursor 1 pixel up and aim at the sky, your next shot could be 30+ degrees off.

untitled

austinjames314:
Cool. So the mortar is still tilted, but the elevation and azimuth displayed are the actual values, relative to the coordinate system, not relative to the slope of the ground. Sounds great! Fantastic! I've been waiting for this ever since I started playing Arma, thanks!

_"One fun bit of BIS logic: if the gun isn't looking at anything it will shoot via weapon direction.
 So if you aim near the top of a mountain and shoot, then move your cursor 1 pixel up and aim at the sky, your next shot could be 30+ degrees off."_

Wait, run that by me again? You're saying that if there's nothing in the mortar's sights, the rounds will follow the blue vector, and if there is, they follow the yellow vector? If so, that's messed up BIS!

austinjames314:
Sorry to post something not quite on topic, but I figure you seem to have the mortars figured out.
Why do the smoke rounds not follow the same path as the HE rounds? #1814

PR #1826, opened 2015-07-09
AI requirements

ArcticSnowSky:
Hey guys,
I'm writing an AI mod which aims on the usage of specific ACE functionalities (like in my first step, medical treatment).

While most of the things went pretty well, I had a few smaller things to change/add in order to get up so far.

The two simplest things are https://github.com/acemod/ACE3/commit/278a86868675289b748cf9c9f84c4297bdb2195d and 
https://github.com/acemod/ACE3/commit/1ca34d4e8cb2d44b50d01e5ae318a0bf1fe6ebc4.
While the first one simply returns the selection name by a number (It's the reverse of an other function), the second one disables the progressbar on healing when an AI is doing it (otherwhise your input was blocked).

The third one was a little bit more extensive as I had to use a third helper variable.
But in my opinion it was the right decision because otherwhise the variables would state different situations with the same name.
(I was required to know when AI really is carrying someone instead of starting, in order to continue with the commands).

As far as what I've tested, everything seems fine - but I found a problem which I wrote also in 
https://github.com/acemod/ACE3/commit/0bb11268fcea00901181c08c74ee69206f784624
{Notice: is the Timer for Weapon for abortion really needed?
Because if you start carrying a unit, but abort it by switching to pistol, and then quickly drag the unit,
you will notice that after the timeOut you will lose the dragging Task}

By the way, for those who are interested in solo and coop usage, a small preview of what I did so far:
https://drive.google.com/file/d/0BzHyxtkk-Tj_VHU5bVlZQi1TXzA/view?usp=sharing

### DIFF
--- a/addons/medical/functions/fnc_selectionNumberToName.sqf
+++ b/addons/medical/functions/fnc_selectionNumberToName.sqf
@@ -1,0 +1,16 @@
+/*
+ * Author: Glowbal
+ * Get the string representation of a selection index.
+ *
+ * Arguments:
+ * 0: The selection index of a unit <STRING>
+ *
+ * ReturnValue:
+ * String representation. any if invalid. <NUMBER>
+ *
+ * Public: yes
+ */
+
+#include "script_component.hpp"
+
+(["head","body","hand_l","hand_r","leg_l","leg_r"] select (_this select 0));
--- a/addons/medical/functions/fnc_selectionNumberToName.sqf
+++ b/addons/medical/functions/fnc_selectionNumberToName.sqf
@@ -1,5 +1,5 @@
 /*
- * Author: Glowbal
+ * Author: Glowbal, ArcticSnowSky
  * Get the string representation of a selection index.
  *
  * Arguments:
--- a/addons/medical/functions/fnc_selectionNumberToName.sqf
+++ b/addons/medical/functions/fnc_selectionNumberToName.sqf
@@ -13,4 +13,4 @@
 
 #include "script_component.hpp"
 
-(["head","body","hand_l","hand_r","leg_l","leg_r"] select (_this select 0));
+(GVAR(SELECTIONS) select (_this select 0));
--- a/addons/medical/functions/fnc_treatment.sqf
+++ b/addons/medical/functions/fnc_treatment.sqf
@@ -201,32 +201,37 @@ _treatmentTime = if (isNumber (_config >> "treatmentTime")) then {
 };
 
 // Start treatment
-[
-    _treatmentTime,
-    [_caller, _target, _selectionName, _className, _items, _usersOfItems],
-    DFUNC(treatment_success),
-    DFUNC(treatment_failure),
-    getText (_config >> "displayNameProgress"),
-    _callbackProgress,
-    ["isnotinside"]
-] call EFUNC(common,progressBar);
+if (_caller == player) then {
+    [
+        _treatmentTime,
+        [_caller, _target, _selectionName, _className, _items, _usersOfItems],
+        DFUNC(treatment_success),
+        DFUNC(treatment_failure),
+        getText (_config >> "displayNameProgress"),
+        _callbackProgress,
+        ["isnotinside"]
+    ] call EFUNC(common,progressBar);
 
-// Display Icon
-_iconDisplayed = getText (_config >> "actionIconPath");
-if (_iconDisplayed != "") then {
-    [QGVAR(treatmentActionIcon), true, _iconDisplayed, [1,1,1,1], getNumber(_config >> "actionIconDisplayTime")] call EFUNC(common,displayIcon);
-};
+    // Display Icon
+    _iconDisplayed = getText (_config >> "actionIconPath");
+    if (_iconDisplayed != "") then {
+        [QGVAR(treatmentActionIcon), true, _iconDisplayed, [1,1,1,1], getNumber(_config >> "actionIconDisplayTime")] call EFUNC(common,displayIcon);
+    };
 
-// handle display of text/hints
-_displayText = "";
-if (_target != _caller) then {
-    _displayText = getText(_config >> "displayTextOther");
+    // handle display of text/hints
+    _displayText = "";
+    if (_target != _caller) then {
+        _displayText = getText(_config >> "displayTextOther");
+    } else {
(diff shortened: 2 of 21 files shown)


esteldunedain:
:+1: to the changes you propose

jonpas:
How 3rd party AI mod friendly is this, like ASR AI 3?

ArcticSnowSky:
Heya,
to me it would seem better using the if else condition as it would proceed the result value and other things that might come in the future after the ai condition.

@jonpas:
The module at the moment is purely scripted. From my knowledge ASR AI uses a nice fsm and I think they would work together (besides a lot of smoke because my functions would do it, and his fsm would do it both on hit).

Besides that, I made also some kind of Priority Manager that allows to register/unregister by a priority number if you want to take control over the unit.
It fires an event if you lost or gained priority, this way you can handle your control until you free it again.
I use that Prioritymanager also to keep the last AI command or a new player command.
example, my AI needs to go to an other unit, but afterwards it should what it was doing first:
I register the unit with the priority nr and my modulename/function
if I am in control, I give doMove and doStops etc.
when I'm done, i simply free it up by unregistering.
The Priority manager makes the ai do what it did before (returning in formation, or stay).

That's at least the wanted effect. But at the moment it won't move back from where it came or reenter a vehicle or so - at the moment everything is quite simple.

Later on I will have to investigate in using fsm to, but at the moment I don't have a clue on how I could interact on multiple units for one situation (more specific: get a closer range of units, check a few for throwing smoke, a few perhaps for suppressive fire, and one for dragging to a safe spot and healing).

I guess I will simply have to read more tutorials about fsm, but I didn't find many yet

addons/medical/functions/fnc_treatment.sqf
```diff
@@ -201,6 +201,12 @@ _treatmentTime = if (isNumber (_config >> "treatmentTime")) then {
 };
 
 // Start treatment
+if (_caller != ACE_player) exitWith {
+    // It's an AI that is doing the treatment
+    sleep _treatmentTime;
```
Glowbal: Don't use sleep. Use waitAndExec instead.

addons/medical/functions/fnc_treatment.sqf
```diff
@@ -201,6 +201,12 @@ _treatmentTime = if (isNumber (_config >> "treatmentTime")) then {
...
+    [[_caller, _target, _selectionName, _className, _items, _usersOfItems]] call DFUNC(treatment_success);
```
Glowbal: Treatment can also still fail at any moment due to any reason. That should also be handled here.

addons/medical/functions/fnc_selectionNumberToName.sqf
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * ReturnValue:
+ * String representation. any if invalid. <NUMBER>
+ *
+ * Public: yes
+ */
+
+#include "script_component.hpp"
+
+(["head","body","hand_l","hand_r","leg_l","leg_r"] select (_this select 0));
```
Glowbal: This array can be replaced with `GVAR(SELECTIONS)`. Probably the same needs to be done for the `fnc_selectionNameToNumber`.

addons/medical/functions/fnc_treatment.sqf
ArcticSnowSky: In what sense can it still fail?
From what I have seen, there would be:
alive _player (okay, I have to introduce this definitely)
"_condition" (but I have no clue about  getText (_config >> "displayNameProgress"). I guess this one is UI related and not AI.
["isnotinside"] as _exceptions in canInteractWith is needed, and I guess this is one of the things that should be checked constantly, while (alive could be checked also after waiting.

What do you think if I would introduce a new function which is only called progress, and which will be constructed close to progressBar, but instead of testing player, the caller will be tested on those conditions that aren't ui related?

ArcticSnowSky:
I worked on the progress thingy (written at https://github.com/acemod/ACE3/pull/1826#discussion_r34366046) and in my testing environment it worked well.

Regarding the exitWith I still think it would be better to use an if else (fnc_treatment).
Would you mind it if I would change that?

addons/medical/functions/fnc_treatment.sqf
Glowbal: You already quoted one;
```sqf
 alive _player
```` 
And `canInteractWith` is a very important one as well. Also we use the onProgress callback for some treatment functions. In short, you would need to mimic the progress bar without the UI elements - so it's a PFH that checks all relevant conditions + does the onProgress callback. And when stopped, it would trigger failed.

Glowbal:
I'd rather keep the `exitwith` in place.

addons/medical/functions/fnc_treatment.sqf
ArcticSnowSky: Then I guess https://github.com/ArcticSnowSky/ACE3/commit/06a3d46ac77beb62afd598717557868482ed9e1b would be an acceptable solution.
The question now would be more over if it should be the same method, but with an additional parameter (for the caller) and that it will check autonom for caller == player then add UI stuff?

ViperMaul:
Status?

ArcticSnowSky:
At the moment I am reorganizing the project.
Before it was one single file doing the whole work and by the time it became really messy and hard to debug.

The reorganized project hierarchy is completely based on ACE (same structure of modules/functions etc) principally - copy & paste.

At the moment I am working on the optionsmenu,
the user shall get the ability to activate the module only for basic/advanced/both/none of the medical level. But also he shall be able to enable/disable AI throwing smoke, carrying/dragging, and choose at what distance he well carry/drag.

Also the fsm is done a little bit, while it seems to be pretty simple, I came pretty good forward, and as I already have many of the needed functions, I have seen already some good result (from the point of view that I did my first steps yesterday with fsm).

I'd like to get the project as fast as possible on github, because I could need help on various things like pathfinding (I'm scared of that, but I really want to implement the drag/carry fully and beautiful!)

But I also need more debugging experience, because with the current project, I seem to have also various problems on debugging (in sense that I must fully close arma, build pbo, restart the game, slowing me down verry much).

bux578:
I'm closing this due to inactivity.

PR #1834, opened 2015-07-10
Removed safety from binoculars/designators

jonpas:


SilentSpike:
Trivial bugfix, merging

PR #1835, opened 2015-07-10
Exit Ear Ringing PFH when disabled during ringing

jonpas:


SilentSpike:
Another trival bugfix :+1:

PR #1857, opened 2015-07-15
Selective UI Elements

jonpas:
- [x] ~~Split UI configs~~ -> #2368
- [x] ~~Remove Vignette~~ -> #2368
- [x] Framework for selective UI
- [x] `showHUD` parameters selective UI
- [x] Use `showHud` setter
- [x] Player selective UI
- [x] ^ Firing mode
- [x] ^ Ammo type
- [x] ^ Stance indicator
- [x] Move Ammo Count to UI (default disabled on foot)
- [x] Vehicle selective UI
- ~~Toggle Icons~~ config-only
- [x] Squad Leader Command Bar selective UI
- ~~Squad Leader hexagon selective UI~~ config-only, disabled on Elite difficulty
- ~~Peripheral dots selective UI~~ config-only, disabled on Regular+ difficulty
- [x] Stamina bar selective UI
- ~~Chat windows selective UI~~ config-only
- ~~Crosshair~~ config-only
- [x] Base disable/enable UI changes setting (to get rid of the need to force all the settings)
- ~~Add "allow client to change" value to each setting, show it in client settings, don't allow changing those Enabled/Disabled in module/server settings~~ -> #3563
- [x] "Requires" descriptions to all advanced settings
- [x] Add setter (ref. #3561) 
- [x] Module icon
- [x] Disable debug

Known issues:
- Vehicle Radar doesn't do anything (using `showHud`)


jonpas:
To all ACE3 devs, there is a slight problem I encountered, specifically UI elements such as command bar and peripheral dots seem to be hard-coded into the engine. They are not actually displays (no IDDs or IDCs in sub-classes), making it impossible to manipulate in SQF (correct?).

I've been discussing on Slack about the approach, there really wasn't a perfect approach found, if they are completely hard-coded then the only way would be to decide what wants to be enabled and what disabled (in line with what ACE3 wishes to achieve), or remake those UI components so that we could toggle them. I tried adding IDD and IDC to one of them, and that doesn't do anything at all.

Are there any other ideas that could work? What would be the best approach to all this?

Cyruz143:
Aren't the peripheral dots just a .paa file somewhere, could override it in the config to point to a blank/transparent or 1x1 pixel file? Not the cleanest solution ever but it should work.

jonpas:
I want them togglable though, with settings framework. That is the point of the question. :)

SzwedzikPL:
https://community.bistudio.com/wiki/showHUD
https://community.bistudio.com/wiki/shownHUD

It may be useful :)

jonpas:
Ooh it may be indeed, thanks @SzwedzikPL ! (if not for anything else at least making sure what's hard-coded and what not.

jonpas:
On hold until next Arma 3 update (above 2 commands). If you (core devs) want vignette removal now I can do that in separate PR.

jonpas:
On hold further because `showHUD` alternative syntax is not available in 1.50 yet either.

SzwedzikPL:
Good to go i think. Can't wait to see optional Leader Command Bar :)

jonpas:
3 things to decide on for core devs.

> Move Ammo Count to UI (default disabled on foot)

Ammo count when you are on foot is currently disabled in `ace_reload` via config, which means it will never show here. Would it be good to move this here and simply disable it by default? Though, it would mean unless it's set on server/mission anyone could turn it back on, I prefer that though, more freedom for setup.

> Squad Leader hexagon selective UI (config only)

Can only be disabled by config currently, I'd vote to disable it, it's just an annoyance, at least for how we play. FYI, this is the hexagon that shows on your Squad Leader when you look at him.

> Peripheral dots selective UI (config only - maybe in vanilla 1.54)

Will possibly be disabled in 1.54 considering [this comment](http://feedback.arma3.com/view.php?id=17875#c99029), if it won't be I vote disabling them via config as well, not possible via scrip.

TheConen:
As discussed in Slack another suggestion:

Option to disable/enable the chat. I suggest the following options:

 - force enabled
 - force disabled
 - Do not force (toggable on-the-fly via a keyboard-shortcut)

jonpas:
This is feature-complete, there are most likely more UI elements that could be added, can always add them as we find them, it's a lot of looking up the config and then trial and error. Keeping a list in the PR comment to keep track.

Also implemented a "setter", which is basically a public function through which other component should toggle UI elements to prevent UI component itself changing it again (as it has to be done on vehicle changes for example). As requested by @esteldunedain for his changes in scope adjustment (and FCS IIRC).

Vehicle compass (or radar, can't remember) doesn't seem to do anything, but that works through `showHud` setter and seems to be broken even in vanilla. Otherwise all of it was tested and is working, another set of eyes never hurts however.

addons/ui/XEH_clientInit.sqf
```diff
@@ -0,0 +1,41 @@
...
+
+// Exit on Headless
+if (!hasInterface) exitWith {};
+
+["SettingsInitialized", {
+    // Initial settings
+    [true] call FUNC(setElements);
+
+    // On load and entering/exiting a vehicle
+    ["infoDisplayChanged", {
```
esteldunedain: @jonpas, where is this event raised from?

esteldunedain: oh, nvm, it's hooked in ace_common already

esteldunedain:
Looks good. Should be merged asap for 3.6.0 I think

PR #1862, opened 2015-07-16
Cache the objects actions for nearby objects and set the GVAR(objectA…

esteldunedain:
…ctionList) correctly.

Should fix various bugs related to object actions. Needs testing!


jonpas:
Confirmed fixed interaction menu disappearing when using `addActionToObject`.

esteldunedain:
> Confirmed fixed interaction menu disappearing when using addActionToObject.

Yay! Mission accomplished

Issue #1639, opened 2015-06-15
Concertina Wire - locality issue


mistergoodson:
ACE3 Version: 3.1.0 prerelease

**Mods:**
* @cba_a3
* @ace3

**Description:**
Tire damage from concertina wire is only local to the person who deployed the wire (i.e. the person who deployed the wire will see popped tires, whereas everyone else will see undamaged tires).

**Where did the issue occur?**
Multiplayer (not tested in SP).

PR #1867, opened 2015-07-17
concWire - Apply damage where vehicle is local

PabstMirror:
#1639 

handleKilled happens where wire is local,
this sends an event so the damage is applied where the car is local


SilentSpike:
:+1: All looks good to me.

It might be an idea to process the damage immediately local to the wire and then send the damage event to the vehicle (in case the event delay introduces any issues in the damage processing - but I don't think it should in its current state).

PR #1873, opened 2015-07-18
Changing Optics/Attachments

madzig21:
Can some one make it possable to take say a RCO on a M4 and swap it out with say a 8x scope with in the same animation

Example: you got a rco in bag you want to add that to your gun, but the gun already has a scope on it.
Currently: you have to drag the old scope off sit in that animation then move the new scope to the gun and sit in that animation
New: take the new scope drag it on top of the old scope and sit in one animation where you pull old one off and put the new one on 
Effect: sitting in one long animation rather then 2 animations


PabstMirror:
http://ace3mod.com/wiki/user/how-to-make-a-feature-request.html

I don't think we play animations when switching scopes, look at other mods you are running.

PR #1875, opened 2015-07-19
3D Model Update 2.0

jokoho48:
Add New LitterGloves
Add New Litter/packingbandage
Change all Litter Model and Texture to a nomenclature (littergeneric_name)

Models by @Raspu86


bux578:
:+1:

PR #1886, opened 2015-07-22
Wind deflection single pfeh

PabstMirror:
Switches fnc_updateTrajectoryPFH to be a master PFEH, bullets get added to an internal array instead of creating/deleting a new cba PFEH each time.

I believe this solves a big performance problem where the `count cba_common_perFrameHandlerArray` can grow too large.

In actual session I saw it hit >2000.  Most entries are nil, but it sill takes time to process them, which happens each frame.  (2000 is about 5.5ms each frame),

In a stress test mission with about 50 ai + disabling the isPlayer check in handleFired so all AI bullets are tracked.  Unscientific results because AI are unpredictable, but I get a huge performance difference:

image
VS
image

With PFEH debug enabled:
```
ace_windDeflection_pfeh_counter: Average: 0.812859s / 2257 = 0.36015ms
```

### DIFF
--- a/addons/winddeflection/functions/fnc_updateTrajectoryPFH.sqf
+++ b/addons/winddeflection/functions/fnc_updateTrajectoryPFH.sqf
@@ -13,12 +13,11 @@
  *
  * Public: No
  */
-#define ENABLE_PERFORMANCE_COUNTERS
+// #define ENABLE_PERFORMANCE_COUNTERS
 #include "script_component.hpp"
 
 [{
-    BEGIN_COUNTER(pfeh);
-
+    // BEGIN_COUNTER(pfeh);
     private["_accel", "_accelRef", "_bulletSpeed", "_bulletVelocity", "_deleted", "_deltaT", "_drag", "_dragRef", "_isWind", "_lastTime", "_trueSpeed", "_trueVelocity"];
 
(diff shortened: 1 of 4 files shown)


addons/winddeflection/functions/fnc_updateTrajectoryPFH.sqf
```diff
@@ -13,38 +13,44 @@
...
+    _deleted = 0;
+    _isWind = (vectorMagnitude ACE_wind > 0);
+
+    {
+        _x params ["_bullet", "_airFriction"];
+
+        _bulletVelocity = velocity _bullet;
+        _bulletSpeed = vectorMagnitude _bulletVelocity;
+
+        if ((!alive _bullet) || {(_bullet isKindOf "BulletBase") && {_bulletSpeed < 100}}) then {
```
Glowbal: `(_bullet isKindOf "BulletBase")`
We are checking this when we add the bullets to the array. Is it necessary to check it again here?

PabstMirror: We track either `BulletBase` or `GrenadeBase`
I think we still wanted to track 40mm grenades even if they're on a low speed ballistic path.

Issue #1871, opened 2015-07-18
Vector Range Finder Lag Spike


Waffle-SS:
ACE3 Version: 3.1.2

**Mods:**
* @cba_a3
* @ace3

**Description:**
Getting heading or range with the Vector Range Finder causes a lag spike. Normal range finders seem unaffected.

**Steps to reproduce:**
Press Tab or R (ACE range/azimuth keys) while using a Vector.
(spinning from left to right helps show the lag spike)

**Where did the issue occur?**
SP/MP

**RPT log file:**
No script/config errors

austinjames314:
I use the Vector all the time, and I've never noticed this.

Neax10:
Noticed it also on long range (700m+)

PabstMirror:
@Waffle-SS can you try to reproduce this again double check RPT.

I didn't realty feel any lag and the functions seemed reasonable fast, but I was getting RPT spammed with 
```
control[ACE_ScriptedDisplayControlsGroup]: Unexpected control type [15]
```
twice per frame, depending on system, that's what I think could be causing the frame drops.

PR #1897, opened 2015-07-23
#1871 - Don't do ctrlText(set) on RscControlsGroup

PabstMirror:
#1871

Doing `ctrlText` or `ctrlSetText` on a `RscControlsGroup` spams the rpt with 
```
control[ACE_ScriptedDisplayControlsGroup]: Unexpected control type [15]
```

We do both each frame so I believe it is the source of the frame lag when using the vector.

### DIFF
--- a/addons/vector/functions/fnc_illuminate.sqf
+++ b/addons/vector/functions/fnc_illuminate.sqf
@@ -7,35 +7,36 @@ disableSerialization;
 _dlgVector = GETUVAR(ACE_dlgVector,displayNull);
 
 if (_this select 0) then {
-
     {
-        private ["_string", "_index"];
+        if (ctrlIDC _x != 170) then {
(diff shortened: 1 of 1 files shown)


SilentSpike:
Is there a reason we're looping through all the controls instead of just getting control 170?

jonpas:
It seems like it has to be executed on all controls *but* 170 `!=`.

PabstMirror:
I have no clue why it was done that way.
`count allControls (uiNamespace getVariable "ACE_dlgVector")` is 23, but we really only want to change 16 of them, the rest are dummy arma controls and the control group..

SilentSpike:
>It seems like it has to be executed on all controls but 170

Another example of why you shouldn't review scripts when tired :stuck_out_tongue:

Issue #1777, opened 2015-07-01
Explosives can be activated before even being placed


Tuxe99:
It's about the M112 and probably all the other explosives.
You can do this bug only with a partner.

If you pick up a explosive like a M112 and select it in your menu the bomb will appear infront of you.
A partner can activate your bomb - your bomb will drop on the floor with his timer (example: timebomb).

The problem is... Your menu for your bomb won't dissapear but your bomb is still in front of you.
Your partner can set, thanks to your bomb infront of you, as many bombs as he wants to activate.

After that bug your menu for the bomb will block a little bit of your sight + you can't cancel your bomb anymore.

Since I know, you can only escape this bug if you reconnect to the server.

I am new on this website and you can probably tell my english is not the best, so I'm sorry if I wrote something wrong or if this bug was already reported.

Have a good day,
Tuxe

esteldunedain:
Thanks, usually we require a certain format for bug reports (that you can find on the wiki), but this is clear enough. This will be solved once the technique at #1739 is applied to explosives, which is planned.

Thank you for you report.

Issue #1734, opened 2015-06-24
Placing explosives inside walls is possible


BullHorn7:
This video was recorded in 3.0.1 and I made sure to test it again in 3.1.1.1 before opening the report.

* Placing explosives is possible inside walls so the enemy has no chance of ever finding it. 
* Sometimes explosives that are set inside a wall cannot be interacted with. 
* Explosives with fall over on their side when played inside a building

https://www.youtube.com/watch?v=82A6S2zhxCw (first few minutes)

ulteq:
Probably not so easy to fix, but worth a try.

esteldunedain:
This will be solved once the technique at #1739 is applied to explosives.

Issue #1243, opened 2015-05-16
Explosive Attachment failure in MP


Twelvecaliber:
Mods:

@cba_a3
@ace3

Placed ACE3 Modules:

None

Description: In a multiplayer environment, attaching explosives, (specifically a demoblock) to the underside of objects or the sides of them, they will not stick and the explosive just falls to the floor. In a single player environment i am able to attack a demoblock to the top of the wheel arch of the "fennek" car, in multiplayer it simply falls to the ground.

Steps to reproduce:

In multiplayer, spawn a "fennek". Attempt to attach a demoblock to the top of the wheel arch as shown below. (Screenshot taken from singleplayer editor for demonstrational purposes)

![2015-05-16_00001](https://cloud.githubusercontent.com/assets/10883422/7665916/b692affa-fbc5-11e4-8092-750cc2f86d7d.jpg)

Where did the issue occur?

MP Client and dedicated.

esteldunedain:
Solution: switch attach method to use the same method used in ace_attach.

Khaosmatic:
You can attach explosives to vehicles?! WHAAAAAAAAAAAAAAAAAT I did not know dis.

PR #1902, opened 2015-07-25
Virtual explosives place

PabstMirror:
Just like the virtual attach, except now doing per-frame placement scanning so the user can see roughly what it will look like when attached (I should backport that to attach at some point).

https://www.youtube.com/watch?v=np4135sOsso

Should solve
Glitching items into the player moves them over the map #1234
Explosive Attachment failure in MP #1243
Explosives can be activated before even being placed #1777
Placing explosives inside walls is possible #1734


alganthe:
Did some tests with a buddy on a dedicated:
- This fixes the placement issues (the explosives stick to surfaces and vehicles)
- New issues:
 - Explosives can ragdoll and kill you now (very rare, extremely hard to reproduce) if you try to place them inside vehicles
 - Explosives stay "straight" when attaching them to vehicles, see here -> http://steamcommunity.com/sharedfiles/filedetails/?id=490214873

PabstMirror:
Fixed #2018 (unconc - deadman)

The big physics problem was the satchel, which had a mass of 15.  It could kill a player if it collided with them at all.  Limiting mass to 5 seems to fix, and shouldn't really cause any problems.

alganthe:
:+1:

SilentSpike:
@PabstMirror Is #1991 still an issue in this PR?

PabstMirror:
It is, I will see if I can't fix it in this pr

Issue #1727, opened 2015-06-23
Unable to load unconscious into gunner seat


ViperMaul:

ViperMaul:
Test well on vehicles that calls the gunner seat the co-pilot seat. The AH6 little-bird is one example that does not have a gunner seat configured as a co-pilot. The AH6 doesn't have any cargo seats as well. It may have a turret seat configured as an FFV position. But didn't confirm in this test session.

PR #1921, opened 2015-07-27
Fix load unconscious into boats and co-pilot seats

ViperMaul:
Fixes boats #1729 and co-pilot seats #1727


ViperMaul:
Test well on vehicles that calls the gunner seat the co-pilot seat. The AH6 little-bird is one example that does not have a gunner seat configured as a co-pilot. The AH6 doesn't have any cargo seats as well. It may have a turret seat configured as an FFV position. But didn't confirm in this test session.

Issue #1738, opened 2015-06-24
Bring back the "head bug fix" button


ViperMaul:
Add a button in ACE Options or somewhere best appropriate for the Head Bug Fix. 
This button will fix players out of unexpected bad animations that prevents players from getting back into action.

ViperMaul:
Make sure Exploit is not there: 
> I accidently found it when I tried to parajump from a flying ladder, but after I jump of the ladder I was not in the freefall animation, so I could not open the parachute.
Fix Headbug saved my live
Source: https://ace3public.slack.com/archives/public/p1435215051006796

Cyruz143:
Got stuck when two of us accidentally tried to carry the same person, used ```playMove``` in console to get out of it, this is needed as not everyone has console access.

Glowbal:
This is already implemented.

PR #1928, opened 2015-07-28
Headbugfix

ViperMaul:
Adding HeadBugFix button to the ACE Options dialog
#1738


addons/optionsmenu/stringtable.xml
```diff
@@ -345,6 +345,22 @@
             <German>Protokolliert Debug-Informationen im RPT und speichert sie in der Zwischenablage.</German>
             <Portuguese>Envia informação de depuração para RPT e área de transferência.</Portuguese>
         </Key>
+		<Key ID="STR_ACE_OptionsMenu_headBugFix">
+            <English>Headbug Fix</English>
```
bux578: Don't add all languages if the values aren't translated. It's hard to find those instances for translators.

bux578:
Why are spectator commits in here?

ViperMaul:
Something weird that GitHub automatically. As discussed in ACE Dev Channel, It has to do with the fact that I created a branch based off of 'Master' and then created a PR for this headbugfix branch into 'Release.'  I don't understand it. It looks like a GitHub bug to me.

bux578:
"master" still has spectator whereas "release" doesn't/shouldn't.

SilentSpike:
You can fix this with a rebase:
`git rebase --onto release master headbugfix`

It will require a `git push -f` though (to the headbugfix branch)

addons/optionsmenu/stringtable.xml
ViperMaul: Good point.

gienkov: ```<Polish>Napraw headbug</Polish>```

addons/optionsmenu/stringtable.xml
```diff
@@ -345,6 +345,12 @@
...
+        </Key>
+		<Key ID="STR_ACE_OptionsMenu_headBugFixTooltip">
+            <English>Resets your animation state.</English>
+        </Key>
```
gienkov: ```<Polish>Resetuje aktualną animację.</Polish>```

gienkov:
I will be glad if you include my translations ;).

addons/common/functions/fnc_headBugFix.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * @Descr: Fixes animation issues that may get you stuck
+ * @Author: rocko
+ *
+ * @Arguments:
+ * @Return: nil
+ * @PublicAPI: true
+ */
+
+#include "script_component.hpp"
+if (player != vehicle player) exitWith {};
```
Glowbal: Should use `ACE_player`.

Glowbal: Also it is lacking any kind of check for for example the unconscious state. It is possible to use fix headbug while unconscious, even though it's a valid state. I don't think that should be the case, even though most likely you will be moved back into the unconscious animation after.

SilentSpike:
I'm thinking maybe we should just stage this for insertion into 3.3.0?

ViperMaul:
This is a direct port from ACE2 as requested by the community. We can improve on it after 3.2.0

PR #1930, opened 2015-07-28
[WIP] updated compat pbo with bwa3 version 1.2

commy2:
- old compat pbo will no longer work with the upcoming update of bwa3
- this PR should be merged once we released v1.2.0 of the bw mod
- WIP is there so this doesn't get merged now, it wouldn't work with bwa3 v1.1.2 (mainly because of fucked up class inheritance)

- also adds compatibilty for goggles


bux578:
So, BWMod won't be compatible with ACE3 by default?

commy2:
No, because of the FCS and required config changes (discreteDistance etc.). I put everything in this compat pbo, because it would be needed anyways.

commy2:
>Last time I've looked into the public git repo I've spotted some ace config values.

I moved all values in this pbo. It's easier to maintain changes. A compat PBO would be required anyways as I said. (Because of the FCS)

jonpas:
This does contradict the statement that ACE3 compatibility should be done in mods and ACE3 team is not responsible for it. Yeah I know you are part of the team, just saying it goes against what you all are saying all the time. :)

SilentSpike:
I mean, I don't think the message is that compatibility should necessarily be handled directly within mods. Just that if compatibility is to be provided it should be done so by respective mod authors, this (providing a compatibility PBO) being a prime example of one way they can do that.

jonpas:
That makes more sense wording it like that indeed.

bux578:
@commy2 With bwmod 1.2 released, Status?

commy2:
No longer needed. We have a comp file provided with 1.2 and onwards.

Issue #1932, opened 2015-07-28
Parachute cutting and reserve parachute issues


alganthe:
ACE3 Version: 3.2.0 release branch

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
doesn't matter

**Description:**
The steerable / non steerable parachutes are missing interactions for the cutting furthermore if you run the function yourself the reserve parachute isn't applied

**Steps to reproduce:**
- Place a paratrooper in the editor
- Set his altitude at 1 000 m
- Click preview
- Open the parachute and try to cut it
- Press escape and enter ```[player] call ace_parachute_fnc_cutParachute;```
- Press enter and notice that the reserve parachute isn't applied

**Where did the issue occur?**
Editor

**RPT log file:**
nothing in there

bux578:
@jokoho48

jokoho48:
i am on it

alganthe:
fixed in #1933

PR #1933, opened 2015-07-28
Parachute fix for #1932

jokoho48:
Parachute fix for #1932


alganthe:
Confirmed to fix the issues in #1932 :+1:

PR #1952, opened 2015-07-30
Hungarian translation update

Harakhti:
Blame it on me, I kept shrugging it off and it piled up rather fast.

### DIFF
--- a/addons/advanced_ballistics/stringtable.xml
+++ b/addons/advanced_ballistics/stringtable.xml
@@ -32,6 +32,7 @@
             <German>Erweiterte Ballistik</German>
             <Czech>Pokročilá balistika</Czech>
             <Portuguese>Balística avançada</Portuguese>
+            <Hungarian>Fejlett ballisztika</Hungarian>
         </Key>
         <Key ID="STR_ACE_Advanced_Ballistics_enabled_DisplayName">
             <English>Advanced Ballistics</English>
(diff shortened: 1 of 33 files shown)


addons/captives/stringtable.xml
```diff
@@ -204,6 +209,7 @@
             <Czech>Mohou hráči spoutat jednotky na své straně</Czech>
             <German>Können Spieler eigene Einheiten fesseln</German>
```
Garzog: even though not topic of this translation i would change "Kann Teamkollegen fesseln" in "Kann eigene Fraktion fesseln"

addons/advanced_ballistics/stringtable.xml
```diff
@@ -48,6 +50,7 @@
             <German>Aktiviert die erweiterte Ballistik</German>
             <Czech>Aktivuje pokročilou balistiku</Czech>
             <Portuguese>Ativa balística avançada</Portuguese>
+            <Hungarian>Engedélyezi a fejlett ballisztikát</Hungarian>
```
Garzog: I´d change "Engedélyezi" into "Aktiválya" feels better for me but i might be a bit rusty.

Harakhti: They're contextually close to make sense either way, but I might adjust it later on. "Enabling" would mean the same as "Activating" here.

PR #1954, opened 2015-07-30
Improvement for handling minor levels of pain in advanced medical

Glowbal:
I received feedback about the pain handling in advanced medical (@viperMaul). This could be a (temp) solution.


gienkov:
Interesting change. Low level of pain will reduce by itself.

Issue #1959, opened 2015-07-31
Laser Locality Issue


PabstMirror:
https://github.com/acemod/ACE3/blob/master/addons/laser/functions/fnc_laser_init.sqf#L21

`if(!(local _laserTarget)) exitWith { };`  

But when a vanilla laser comes on it seems to be local to everyone
so everyone runs FUNC(laserOn) and does the global event

In my mp testing `count (ace_laser_laserEmitters select 1)` with a single laser on was equal to the number of players.

Client 1 (shooter):
```
New Laser - 4c126400# 164378: empty.p3d [true] - LaserTargetW
laserOn globalEvent: ["23224469787",[B Alpha 1-2:1 (Pabst Mirror (2)),B Alpha 1-2:1 (Pabst Mirror (2)),"ace_laser_fnc_vanillaLaserSeekerHandler",1550,1001,1]]
```

Client 2:
```
New Laser - 49ee8800# 164039: empty.p3d [true] - LaserTargetW
laserOn globalEvent: ["23261625052",[B Alpha 1-1:1,B Alpha 1-1:2 (Pabst Mirror),"ace_laser_fnc_vanillaLaserSeekerHandler",1550,1001,1]]
```

Edit:
Playing with using
```
    _isLocal = false;
    {
        if ((laserTarget _x) == _laserTarget) exitWith {
            if (local _x) then {_isLocal = true};
        };
    } forEach allUnits;
    if(!_isLocal) exitWith { };
```

Issue #1554, opened 2015-06-09
Framedrop with RAH-66 Commanche


gienkov:
ACE3 Version: 3.0.1 Stable

**Mods:**
https://docs.google.com/spreadsheets/d/1OTjYgUqxEnQsQBAfgwmjbThDZ3unG21H1jfGkZEwO00/edit?usp=sharing

**Placed ACE3 Modules:**
http://pastebin.com/PVnhFPFF

**Description:**
*Shooting DAGR rockets is very unreliable, they often not homing into target (flying straight). They also seem to not lock onto laser designator, simply missing the target. Most annoying is framedrop you get when you try to shoot some of those rockets. I think it may be related to script error I got, as seen below and in RPT.*

http://images.akamai.steamusercontent.com/ugc/37499844262244477/FF64C479C03EC523BAADA0A55247B660E02B7D54/
Yeah, I was shooting DAGR's while on the ground, it's possible, few were homing onto target but other's kept flying straight after shot.

**Steps to reproduce:**
* *Get as gunner into Commanche*
* *Order AI or get your friend to get as pilot*
* *Try to shoot targets with laser designator turned on and off*
* *Notice framedrop*

**Where did the issue occur?**
*"Multiplayer"*

**RPT log file:**
http://pastebin.com/1zxCt5zz

esteldunedain:
@gienkov, the problem with the framedrop is two fold:

- part is genuine, due to laser an guidance
- but most of it is probably to to rpt spam

So the thing is to find why this is erroring., and we believe it's due to those Updating base class errors you are getting. The solution would be to retest with different addons and see what conflict is causing this. Could you do some tests adding/removing pbos?

bux578:
http://ace3mod.com/wiki/user/how-to-report-an-issue.html

>If you've found an issue with ACE3 please make sure that ACE3 is really the cause of the problem. To do this try to reproduce the issue with using only @cba_a3 and @ACE3 on a newly created mission.

:smile:

bux578:
Still happening. 60 -> 38 fps when unloading all DAGR's

gienkov:
Still happening on 3.2.1.

gienkov:
Additional note: DAGR's are missing very often - they don't hit point I've targeted and lased.

bux578:
> Still happening on 3.2.1.

Milestone is 3.3.0

bux578:
Close in favor of #3211

esteldunedain:
Closed in favor of #3308

PR #1960, opened 2015-07-31
Laser - Handle Bad Data Returns

PabstMirror:
This is a bandaid for issues releated to #1959

`fnc_vanillaLaserSeekerHandler.sqf` can set _laser to [-1,-1];

which causes an error with CBA_fnc_vect2polar that expects a vector.


PabstMirror:
This should also help framedrop in #1554

PR #1961, opened 2015-07-31
disable hearing & medical PP for remote controlled units

SzwedzikPL:


SilentSpike:
I think remote controlled units are intentionally supported in this way, otherwise there would be little point in using `ACE_Player` :stuck_out_tongue:

bux578:
I don't like this ... especially not as default. Maybe a setting...

SzwedzikPL:
> I don't like this ... especially not as default. Maybe a setting...
Ok, no problem. I will add an extra pbo for my community to disable this in zeus scenarios.

Thank you :)

PR #1963, opened 2015-07-31
ace_goggles changes

SilentSpike:
Adds:
- Support for alternate unit to be used as the goggles target via function `ace_goggles_fnc_changeUnit` (came about from needs of ace_spectator)
- Setting to enable/disable the goggles overlay (merged the third person setting into this)

Also includes some minor optimizations and refactoring.

Oh and bonus side effect:
- Dirt and damage effects now supported by units other than the player


jonpas:
Yay, did my job, thanks @SilentSpike ! :grin:

SilentSpike:
Re-thinking this, I'm not sure if the unit changing functionality is necessary. Might strip that out and preserve the other enhancements.

SilentSpike:
Eh, I think I'm going to close this. The goggles code could user a refactoring at some point, currently this is in too messy a state to do properly.

The including optimizations can be done when this gets refactored.

PR #1976, opened 2015-08-01
More realistic GAU-8 gun for A10

SzwedzikPL:
This is port of "GAU-8/A Avenger" mod (made with permissions from authors)
All credits go for original authors: @AlexusDE and @ZabuzaW

- [x] Gun, Mag & Ammo cfg
- [x] muzzleEffect
- [x] Custom cloudlets effect
- [x] RHS USAF compat
- [x] Correct sound


sargken:
Usaf mod combat?

alganthe:
Don't worry for the doc, i'm here.

SzwedzikPL:
> Usaf mod combat?

Yes, i will add compatibility for RHS USAF in ACE combat pbo.

> Don't worry for the doc, i'm here.

Ok :D

alganthe:
Tested it, my pc crashed with this error code: 0x0000092 BRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRT

:+1:

SzwedzikPL:
> Tested it, my pc crashed with this error code: 0x0000092

:P

jonpas:
Coding guidelines please (opening braces in same line). :)

SzwedzikPL:
> Coding guidelines please (opening braces in same line). :)

Fixed. Thank you :)

SilentSpike:
I'm being told that the GAU-8 shouldn't be that powerful again that kind of armour (video linked in the OP).

However I'm admittedly not an expert on either weapon/vehicle configs or the real life counterparts so I can't add much beyond that. Just thought it was worth noting here for ACE devs who know more about this stuff to consider.

alganthe:
no burst fire, and the shells are not penetrating anything that is more armored than an APC, making it pretty similar if you take VD into account.

ZabuzaW:
The A-10C IRL, in contrast to older versions like A-10A, does not have the two burst modes High and Low RPM, it has only one fix RPM like other A/Cs. However decided to implement this feature because we liked the fact that the player can choose which amount of ammo is depleted with each pull of the trigger to compensate him being not able to choose how long a burst continues, once pulled (in contrast to real life where you only shoot as long as holding down the trigger). This can be pretty handy if facing light or non armored targets against heavy armored targets.

A-10C mostly is equipped with a 'Combat Mix' of rounds, consisting of AP and HE ammunition. The AP rounds do penetrate light and medium armored targets with ease. Targets like a T-72 or a BMD-2 can also be penetrated by the AP rounds but you need some good aimed once to destroy such like a one second burst completely hitting the target. Modern, heavy armored vehicles like T-90 or T-14 properly won't be penetrated by the Gau-8.
However according to my experience it is not easy to different the armor of such targets in ArmA in such way that you can achieve the result you would expect. It would rather require a job to be done in adjusting all the vehicle armor correctly before you can set a damage for the Gau-8 to achieve a realistic result.

nonferrousiron:
ZabuzaW: The A-10A did not have burst modes either. What it did have was a pilot-selector to toggle between two rates of fire (Wikipedia says 2900 and 4200 rounds per minute, other sources say other values, the specifics don't really matter). The -C lost the ability to fire at anything other than a reasonably high rate (again sources differ: I'm seeing 3900, 4000, and 4200 rounds per minute).

Regardless of variant, when the trigger is held they all function identically: they fire until the pilot lets go of the trigger, it runs out of ammo, or it jams. Burst modes have never been a thing, and shouldn't be, in my opinion. Very unfortunate if it is an engine limitation that prevents continuous fire, but given the amount of ammo the thing carries, trying to save a handful of individual rounds via multiple burst modes seems an unnecessary - and most of the time not meaningfully useful - complication.

As for round penetration, even a not-monkey-model T-72 would hold up well to a burst of AP, nevermind CM, from an A-10. The only chance a pilot has against something like that is a high-angle burst from the rear, to focus as much as possible on the top and rear armor, and even then the attack needs to be distressingly-close range to ensure minimum spread and maximum concentration. I have an irrational love for the A-10, but as an *anti-tank* weapon specifically, the GAU-8 was obsolete by the 1980's. (That said, a good burst of CM would handily destroy any external equipment on said T-72. Mission kill quite possibly, actual kill or any sort of armor penetration is much harder to say.)

I suppose we'll have to agree to disagree about the burst thing, but I do humbly suggest that you may have gone overboard in terms of the damage of the cannon. Any modern MBT - nevermind the the futuretanks we have in Arma 3 - would simply laugh at that gun unless it was a burst directly into the top armor with virtually zero dispersion. External equipment broken, crew rattled, sure. Seeing them explode in a spectacular fireball from a little squirt... It's just wrong.

ZabuzaW:
As I said the A-10A would have burst modes I meant the RPM toggle, sorry for that. The A-10A had the two RPM modes, the A-10C did not. We used this mod now for half a year and our pilots liked the RPM modes very much. If you (or any other majority) don't like them, well, it is not the major point of this mod, just leave it out.

Now to the penetration thing. I think that's what I said, at least what I wanted to say. You can easily destroy some non to medium armored vehicles and with some effort also older heavy armored targets. You won't be able to make the life of a modern MBT with the GAU-8 hard, at least IRL.

If you can achieve a realistic result for the whole range of vehicles, so dead infantry and trucks with ease, smoking medium armor with some direct hits, defeated older heavy armor with some effort and no penetration of modern armor, well that would be awesome.
Please feel free to adjust the values to achieve this! ^^

Nagrach:
@nonferrousiron
I have to disagree with our oppinion on the dmg a Gau would do to a modern tank.
If you look back to the time of the tigerII the tanks where disabled by shooting them so many times that their optics aka winkelspiegel where totally demolished behind repair. 

Maybe the tank dosnt blow up but crucial equipment would be damaged or disabled. Stuff like thermals/nvg/antennas/tracks.
Overall i would say that no modern mbt would survive an encounter with its modern counterpart.

And this in mind - an 10 should be able to disable a tank with gau runs.

SilentSpike:
@Nagrach It looks like you agree to me

>That said, a good burst of CM would handily destroy any external equipment on said T-72. Mission kill quite possibly, actual kill or any sort of armor penetration is much harder to say.

SzwedzikPL:
@Nagrach can you test actual damage? I would be gratefull :)

Nagrach:
@SzwedzikPL when the dmg is the same as in mighty gau - i would say the gau is to strong. 
But i had my problems to create a valid testing scenario.

SzwedzikPL:
@Nagrach For now is about ~30% of mighty gau damage level.

AlexusDE:
How is the damage now compared to vanilla?
Btw, this is my GitHub account.

moonmanmove:
is this still being worked on?

SzwedzikPL:
> is this still being worked on?

Generally yes but i don't have much time now.

addons/aircraft/CfgWeapons.hpp
```diff
@@ -122,18 +121,116 @@ class CfgWeapons {
...
+            };
+        };
+        class autoLO: Mode_FullAuto {
+            displayName = "LO";
+            displayNameShort = "LO";
+            reloadTime = 0.012;
+            dispersion = 0.0082;
+            sounds[] = {"StandardSound"};
+            class StandardSound
+            {
```
jonpas: On same line.

addons/aircraft/CfgWeapons.hpp
```diff
@@ -122,18 +121,116 @@ class CfgWeapons {
...
+            showToPlayer = 1;
+            textureType = "burst";
+        };
+        class autoHI: autoLO {
+            displayName = "HI";
+            displayNameShort = "HI";
+            reloadTime = 0.008;
+            textureType = "fullAuto";
+            class StandardSound
+            {
```
jonpas: On same line.

addons/aircraft/functions/fnc_gau8_muzzleEffect.sqf
```diff
@@ -0,0 +1,41 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private ["_weapon", "_projectile", "_i", "_no"];
+
+_weapon =  param [1];
+_projectile = param [6];
+
+if(_weapon != "Gatling_30mm_Plane_CAS_01_F") exitWith {};
```
jonpas: Space after `if`.

Glowbal:
Moved up one version.

SzwedzikPL:
> Moved up one version.

No problem. My community will test version of this soon. I will update this PR after tests

Glowbal:
> No problem. My community will test version of this soon. I will update this PR after tests
 Hide all checks

Thanks, keep us updated :+1:

addons/aircraft/CfgCloudlets.hpp
```diff
@@ -0,0 +1,145 @@
+
+class CfgCloudlets {
+    class Gau8DefaultCldlt {
```
commy2: These need `ACE_` tags

commy2:
needs newlines at EOF

Glowbal:
Over a year later. If this is still desired, please make a new PR.

Issue #1785, opened 2015-07-03
"Do When releasing action menu key" option in OFF interferes with controls


Legolasindar:
ACE3 Version: 3.1.1

**Mods:**
* `@cba_a3`
* `@ace3`

**Placed ACE3 Modules:**
None

**Description:**
* With "Do When releasing action menu key" on, things like move and extend the telescopic ladder option is possible. But if you have that option turned off, it is no longer possible.
Same for sandbag for example.

**Steps to reproduce:**
* Create a simple mission with Virtual Arsenal in a object.
* Go ACE Options menu with ESC key, and select "Do When releasing action menu key" with OFF option.
* Get Telescopic ladder.
* Put telescopic ladder. You cant move or extend this.
* Go ACE Options menu with ESC key, and select "Do When releasing action menu key" with ON option.
* Interact with telescopic ladder. You can now can move or extend this.

**Where did the issue occur?**
* Occur in SP and MP mission.

**RPT log file:**
Not need.

austinjames314:
I can confirm this. I've been using "Do When releasing action menu key" set to 'off' so that I could actually use the menu properly, and It wasn't until someone else used the ladder that i realised that my settings meant I couldn't use it properly. My point is, there's no error message or anything, it just behaves differently.

Legolasindar:
I see with sandbag occurs the same.

Issue #1965, opened 2015-07-31
Virtual Attach doesn't work with certain ACE settings


Waffle-SS:
ACE3 Version: 3.2.0

**Mods:**
* @cba_a3
* @ace3

**Description:**
When "Do action when releasing menu key" is disabled, the Virtual Attach system doesn't run, and the object is instantly attached where you were looking.

**Steps to reproduce:**
1. Disable the ACE option: "Do action when releasing menu key"
2. Try to attach a chemlight to a vehicle

**Where did the issue occur?**
SP/MP

**RPT log file:**
No config/script errors

PR #1977, opened 2015-08-01
Mouse cooldown for click to interact option

PabstMirror:
#1965


PabstMirror:
Fix #1785 as well.

Problem is adding a "DefaultAction" actionEventHandler in the same frame as the user has clicked the action from the interaction menu,

Issue #1892, opened 2015-07-23
Wind direction arrow updates only basing on body position, not head position


gienkov:
ACE3 Version: 3.1.1

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *Advanced ballistics*

**Description:**
*Wind direction arrow (LSHIFT + K) updates wind direction only when you move whole body. It should also update basing on head look direction. You sometimes don't want to move much when on firing position to avoid being detected.*
*Compass can do this thing (rotates based on body position and head position).*

**Steps to reproduce:**
* *Go on some windy hill (to get any wind)*
* *Launch wind direction arrow (LSHIFT + K)*
* *Hold down LALT to move your head*
* *Notice how wind dir. arrow doesn't rotate when you rotate your head*
* *Move your whole body*
* *Arrow rotates*

**Where did the issue occur?**
*"Multiplayer", "Singleplayer"*

**RPT log file:**
*--*

PR #1980, opened 2015-08-02
Headdirection based wind indication

MikeMatrix:
Replaced direction detection with ```CBA_fnc_headDir``` to calculate the indicator direction off of the head direction instead of the player direction.

This solves #1892.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -107,3 +107,4 @@ VyMajoris(W-Cephei)<vycanismajoriscsa@gmail.com>
 Winter <simon@agius-muscat.net>
 zGuba
 Drill <drill87@gmail.com>
+MikeMatrix <m.braun92@gmail.com>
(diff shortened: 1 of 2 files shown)


PabstMirror:
Great job, this should be perfect.

Only thing I can think of is that someone with TrackIR might want to try this, it might effect you differently with freelook always on.

MikeMatrix:
@PabstMirror I tried with TrackIR with the ACE Test Mission and it worked just fine.

SilentSpike:
:+1: This is why CBA is so handy

jonpas:
Intruding, but is there set head direction function as well?

SilentSpike:
Not possible @jonpas

jonpas:
Thought so, I guess this concludes my search for it(cfgMoves better get ready for me lol), thanks @SilentSpike

Issue #1981, opened 2015-08-02
Can't turn combat deafness off with Hearing module.


Alextwokay:
ACE3 Version: 3.2.0
(stable)

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *Hearing*

**Description:**
*Turning combat deafness off doesn't eliminate beeping/rumbling and earplugs are needed anyway.*

**Steps to reproduce:**
* *Fire a few shots with an XM312.*

**Where did the issue occur?**
*Editor and Dedicated server*

**RPT log file:**
*N/A*

PabstMirror:
There are 2 settings, combat deafness (set via module or manual ace_settings) and disable ear ringing which is client settable.

Ear Rining will still happen even if combat deafness is off.  Which might not be clear.

Could you check your client ACE Settings for "Disable ear ringing" and make sure it is enabled and see if you still get any ringing when shooting.

Alextwokay:
That will be it then, this behaviour is different from last patch and seems a bit more unclear.
Having the option to force disable both in the module would be good.

MrSanchez:
Hope we'll be getting an option to force disable both.

PR #1985, opened 2015-08-02
Add Ear Ringing setting to module

PabstMirror:
Ref #1981

Adds "Disable Ear Ringing" setting to hearing module.
Because both can be enabled independently, I think it's important to let mission makers disable both effects.

This only allows either "Do Not Force" or "Disable Ear Ringing - Yes".  I don't think it's a good idea to force enable the effect for people (people could do it via ace_settings if they really want to)


SilentSpike:
Bonus points for adding those strings to common :+1:

I've been using the BI ones, but this is probably more reliable

jonpas:
`Do Not Force` should probably be in common as well. Great job though!

Glowbal:
This should be good to go once we fix the merge conflicts :+1:

Issue #1984, opened 2015-08-02
SetDraggable fuck interaction hierarchy on most objects


alganthe:
ACE3 Version: 3.2.0

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
Slideshow (for testing purposes)

**Description:**
When using setDraggable on most objects it fucks the interaction hierarchy making the point disappear, for the whole object class.

**Steps to reproduce:**
repro mission : https://www.dropbox.com/sh/ahzb5wjhmwnz36z/AADCowv3s0rj_R4xvRjokH4ga?dl=0
- Note that the wielding tank has the interaction and not the workbench.
- Replace the workbench by a wielding tank, notice that now both don't have any interaction points.

**Where did the issue occur?**
editor

**RPT log file:**
N/A

PR #1986, opened 2015-08-02
Always make sure ACE_MainActions exists

PabstMirror:
#1984
Made the workaround @jonpas made for slidshow work for all actions.
If ACE_Mainactions doesn't exist (true for a lot of non-vehicle objects) then add it.

Also add debuging output when adding action with no parent.


SilentSpike:
:+1: Looking good

jonpas:
Yay :+1: for Slack and :+1: for this!

alganthe:
:+1:

Issue #1821, opened 2015-07-08
Change code to use param/params


bux578:
https://community.bistudio.com/wiki/param
https://community.bistudio.com/wiki/params

PabstMirror:
Did some quick testing, for normal PARAMS it's about 50% faster, but it was fairly fast to begin with.
Much bigger improvement when replacing `DEFAULT_PARAM`, which has a function call.

```
TEST1 = {
    PARAMS_5(_a,_b,_c,_d,_e);
    DEFAULT_PARAM(5,_f,6);
};
TEST2 = {
    params ["_a", "_b", "_c", "_d", "_e", ["_f", 6]];
};
```
Test code: `[1,2,3,4,5] call testX;`
0.0496948 ms vs 0.00810547 ms.

Waffle-SS:
Holy crap, that's about 6x faster.

Glowbal:
@PabstMirror how does it compare vs a regular private + select? Since the `PARAMS` macro also does some additional checking.

SilentSpike:
Started this process (because I like using regex) on new branch: `params`

So far I've done each of the PARAMS_N macros on an individual commit to break it up and I've gone through each replacement individually to make sure there were no syntax errors.

SilentSpike:
Just to share what's been discussed in the slack:

We'll likely just update the macros themselves to use the new commands. I'll leave that branch up for the time being though in case anything changes.

Here is my PR to update the CBA macros: https://github.com/CBATeam/CBA_A3/pull/93

Tenga:
@Glowbal I've been told params outperforms select by itself.
> [1,2,3] params ["_lol"] is faster than _lol = [1,2,3] select 0

SilentSpike:
The CBA macros have been updated with these commands. Do we want to close this, or do we want the changes from https://github.com/acemod/ACE3/tree/params ?

jonpas:
I think it's better to keep using macros, that way any future changes can simply be changed there as well (just like this).

bux578:
Done in CBA. Merged in https://github.com/CBATeam/CBA_A3/pull/93

PR #1989, opened 2015-08-02
Clean Up some Code

jokoho48:
Some small Code Clean Up

Replace QUOTE(FUNC()) with QFUNC()

fnc_sendDisplayMessageTo Change von BIS_fnc_Param to New Params command


jonpas:
Clean up == :+1:

jokoho48:
found 24 hits for BIS_fnc_Param Currently Changing

addons/common/functions/fnc_unloadPersonLocal.sqf
```diff
@@ -13,7 +13,7 @@
 //#define DEBUG_MODE_FULL
 #include "script_component.hpp"
 
-#define GROUP_SWITCH_ID QUOTE(FUNC(loadPerson))
+#define GROUP_SWITCH_ID QFUNC(loadPerson)
```
SilentSpike: These are not technically the same thing, but it probably doesn't matter.

jokoho48: Return is both "ace_ADDON_fnc_function" what is ther different?

SilentSpike: Check out this file here: https://github.com/acemod/ACE3/blob/master/addons/main/script_debug.hpp

jokoho48: hmm ok thats for Debug hmm

jokoho48:
More Improvements are incomming after i take a nap

PabstMirror:
This is pretty big change to everything by swapping all the PARAMS_X() with the new `params`.

Ref #1821
@SilentSpike has pushed changes https://github.com/CBATeam/CBA_A3/pull/93 to CBA that would do some of this.

But, one issue this does solve is all the variadic functions E.G.
```
params ["_unit", "_animation", ["_priority", 0], ["_force", false]];
```
Which have been done by hand very nicely here.

I think macroing that would be very ugly.

Biggest downside to this is we lose the auto `TRACE_X` from the `PARAM_X` which has made debugging easier.  But we can always add these back in.

MikeMatrix:
@PabstMirror I agree with this in general, especially the point about `TRACE_X` being missing.
Problem I would still see is that currently from what I've seen, parameter handling is so inconsistent, that even if `TRACE_X` would be there, a quite significant amount of functions still wouldn't display any information, or partial information, because some of it was handled through some custom code anyways.
I think introducing some consistent way is a necessity, but I'm unsure if this can be solved using macro's, considering the varying nature of parameters a lot of times.

Maybe we introduce calling `TRACE_X` right after parameterization is completed.

SilentSpike:
Keep in mind the params commands do have a debugging tool built into them:

>In addition to simple parsing directly into variables, input can be tested in case it is undefined, of the wrong type or of the wrong size (if array) and substituted if necessary with default values. Substitutes can also be logged into .rpt file, if allowFunctionsLog param in description.ext is set to 1.

MikeMatrix:
So do you think it wouldn't even be necessary to have a `TRACE_X` in the parameterization?

SilentSpike:
That said, it was previously agreed that we wouldn't bother to replace every instance of the params macro with the new commands. At least not in one big change like this: #1821

MikeMatrix:
@SilentSpike The initial idea wasn't really to replace those, but to remove those cases where the method used was a manual extraction completely or partially. Since this meant we were going through files anyways, it seemed logical to adjust things to a consistent fashion.
We could simply touch the cases that are explicitly using some form of custom written code for parameterization and leave the rest alone if that's more preferred.

Glowbal:
This is to big to review effectively on github. It might be best if you split off any future work related to this on a separate branch and create a PR for that, so it's in smaller pieces.

jokoho48:
i could split this in separate PR for every PBO one if this is ok?

MikeMatrix:
Incoming PRs

PR #1994, opened 2015-08-03
Improve headbugfix

ViperMaul:
Disable headbugfix for Handcuffed/Surrender,unconscious,water surface and log the situation.

### DIFF
--- a/addons/common/functions/fnc_headBugFix.sqf
+++ b/addons/common/functions/fnc_headBugFix.sqf
@@ -12,8 +12,8 @@
 private ["_pos","_dir","_anim"];
 
 _anim = animationState ACE_player;
-["HeadbugFixUsed", [profileName, (animationState ACE_player)]] call FUNC(serverEvent);
-["HeadbugFixUsed", [profileName, (animationState ACE_player)]] call FUNC(localEvent);
+["HeadbugFixUsed", [profileName, _anim]] call FUNC(serverEvent);
+["HeadbugFixUsed", [profileName, _anim]] call FUNC(localEvent);
 
 if (ACE_player != vehicle ACE_player  || { !([ACE_player, objNull, ["isNotSitting"]] call FUNC(canInteractWith)) } ) exitWith {false};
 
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_headBugFix.sqf
```diff
@@ -10,21 +10,30 @@
...
+_dir = getDir ACE_player;
+
 titleCut ["", "BLACK"];
-_pos = getposATL player;
-_dir = getDir player;
-_anim = animationState player;
+
 // create invisible headbug fix vehicle
-_ACE_HeadbugFix = createVehicle ["ACE_Headbug_Fix", getposATL player, [], 0, "NONE"];
+_ACE_HeadbugFix = createVehicle ["ACE_Headbug_Fix", [0,0,10000], [], 0, "NONE"];
```
SilentSpike: Is it possible to do this with only a local vehicle? (Just to save on the network traffic)

gienkov:
There was a timeout implemented in ACE2 headbug fix, I don't see it in ACE3, you should add ~10 sec timeout before you can use headbug fix.

addons/common/functions/fnc_headBugFix.sqf
PabstMirror: I think this is rare enough, the traffic shouldn't be that bad.  Also, I'm not sure if this is an issue, but it's important the animation is reset on all clients and a local vehicle might?? not sync correctly.

Minor, but moving the player to [0,0,10000] could cause issues with mission triggers (eg blufor not present), what about just
`(getposATL ACE_player) vectorAdd [0,0,10000]`

Glowbal: Can we hide the player object instead of teleporting it away?

alganthe:
The transition is used as timeout, :+1: @ViperMaul

addons/common/functions/fnc_headBugFix.sqf
SilentSpike: A framework was added to hide units: `EFUNC(common,hideUnit)` and `EFUNC(common,unhideUnit)`

ViperMaul:
@gienkov show me the code you are talking about. I grabbed the code from ACE2 already.

ViperMaul:
Thanks guys those comments helped make it look much nicer in my dedi tests.

Glowbal:
Is this ready?

ViperMaul:
Yes!

PR #1999, opened 2015-08-03
removed obsolete wiki entries

alganthe:
- logistics_uavbattery.md
- logistics_wirecutter.md
- getting-started.md


PabstMirror:
Note: (_) was replaced by newer (-) versions
@alganthe looks like laser self designate has same issue. I'll delete that as well

PR #2005, opened 2015-08-04
Code cleanup Zeus module

MikeMatrix:
Code cleanup of the Zeus module.

Split PRs from #1989.

### DIFF
--- a/addons/zeus/functions/fnc_handleZeusUnitAssigned.sqf
+++ b/addons/zeus/functions/fnc_handleZeusUnitAssigned.sqf
@@ -36,7 +36,7 @@ for "_n" from 0 to (_numCfgs - 1) do {
             if !(isClass (configFile >> "CfgPatches" >> _x)) exitWith {
                 _removeAddons pushBack (configName _cfg);
             };
-        } forEach _requierdAddon;
+        } forEach _requiredAddon;
     };
 
     if (isText _cfg) then {
--- a/addons/zeus/functions/fnc_handleZeusUnitAssigned.sqf
+++ b/addons/zeus/functions/fnc_handleZeusUnitAssigned.sqf
@@ -19,11 +19,11 @@
 
 #include "script_component.hpp"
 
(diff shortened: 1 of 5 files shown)


kripto202:
I can't shake the feeling there is a typo on the requiredAddon in fnc_handleZeusUnitAssigned.sqf

MikeMatrix:
@kripto202 I don't know what you're talking about. There is no typo there.

(In all seriousness, I can't even explain how it got there in the first place O_O)

bux578:
> (In all seriousness, I can't even explain how it got there in the first place O_O)

You guys fixed the typo in the `private` line but not where the variable is used in the code.

MikeMatrix:
@bux578 actually it looks like that typo got introduced in the first commit of this PR...

bux578:
ah yes... too early

MikeMatrix:
Probably also the reason for the typo initially...

jonpas:
What's the decision for using params vs CBA PARAMS_X now? Since it's exactly the same thing in this case, shouldn't PARAMS_X be used for better debugging purposes? And params should be used directly only where replacing BIS_fnc_param/requiring more options?

SilentSpike:
I'm not sure in regards to updating old code like this, but I know for any future written code it's basically just the developers choice.

jonpas:
But PARAMS_X does give you EXPLODE macro debugging. Best solution would probably be one macro/command for all cases of course, not sure if that's doable.

bux578:
https://community.bistudio.com/wiki/params

> In addition to simple parsing directly into variables, input can be tested in case it is undefined, of the wrong type or of the wrong size (if array) and substituted if necessary with default values. Substitutes can also be logged into .rpt file, if allowFunctionsLog param in description.ext is set to 1.

jonpas:
Ooh I missed that one! Thanks bux! 

My final thought is that entire ACE3 codebase should be unified, so there isn't mix of macros abd commands, IMO best would just be to use param/params everywhere. Of course it's not up to me, I just say stuff around here. :grin:

PR #2010, opened 2015-08-04
Code cleanup Interact_menu module

MikeMatrix:
Code cleanup of the Interact_menu module.

Split PRs from #1989.


PabstMirror:
Everything looks good, only non-trivial change is createAction which appears to be fine.

Issue #1807, opened 2015-07-06
Bloody bandage litter created even though limb was perfectly healthy and has never been wounded


bux578:
ACE3 3.1.1

**Mods:**
* cba_a3
* ace3

**Placed ACE3 Modules:**
* None

**Description:**
Bloody bandage litter created even though limb was perfectly healthy and has never been wounded

**Steps to reproduce:**
* Place two units
* Bandage the limb of another unit (don't wound it)

**Where did the issue occur?**
* Editor

**RPT log file:**
n/a

jonpas:
> ACE3 3.1.0

???

bux578:
typo

jonpas:
:banana:

PR #2026, opened 2015-08-05
Clean bandages if no damage on hitpoint

PabstMirror:
#1807 (Bloody bandage litter created even though limb was perfectly healthy and has never been wounded)

New model based on `littergeneric_bandages2.p3d` with new texture


Glowbal:
I know implementing the condition for litter was a smart idea. 

Nice one, @PabstMirror  :+1:

Issue #2024, opened 2015-08-05
Disable team management module does not work for JIP players


captainblaffer:
ACE3 Version: 3.1.1

Mods:
@3cb_baf_equipment;@3cb_baf_weapons;@AllInArmaTerrainPack;@PsyLandRover;@acepa;@ASDG_JR;@ASDG_JM;@BWA3;@CBA_A3;@fhq_accessories;@hlcmods;@k_mnp;@mrt_accfncs;@rhs-hlc;@rhsafrf;@rhsusf;@stui;@tao_foldmap_a3;@task_force_radio;

Placed ACE3 Modules:
	value="0";
	parentCls="ACE_ModuleInteraction";
	typeName="EnableTeamManagement";

Description:
Disable team management module does not work for JIP players. They will be able to join other player's groups.

Steps to reproduce:
1. Make mission with disable team management module.
2. Host mission on dedicated. P1 connected at mission start.
3. Have P2 JIP, in a different squad.
4. Have P2 join P1's group using ACE interaction.

Where did the issue occur?
Dedicated

RPT log file:
http://pastebin.com/vZnDFNQB

PabstMirror:
Thanks for helpful report, fixed.

PR #2027, opened 2015-08-05
EnableTeamManagement setting for joinGroup

PabstMirror:
#2024

I'm pretty sure joinGroup should be disabled if EnableTeamManagement is false.
Unless we want separate settings for fireteam team management (different colors) and a setting for joining/leaving groups.


Glowbal:
I think the same would count for `leaveGroup`.

PabstMirror:
`leaveGroup` is a child of `ACE_TeamManagement` so it has the condition checked there

PR #2029, opened 2015-08-06
Localization of key binding categories

MikeMatrix:
Adjusted keybinding categories to be localizable.

### DIFF
--- a/addons/advanced_ballistics/initKeybinds.sqf
+++ b/addons/advanced_ballistics/initKeybinds.sqf
@@ -1,4 +1,4 @@
-["ACE3 Equipment", QGVAR(ProtractorKey), localize LSTRING(ProtractorKey),
+[localize ELSTRING(common,ACEConfigCategoryEquipment), QGVAR(ProtractorKey), localize LSTRING(ProtractorKey),
 {
     // Conditions: canInteract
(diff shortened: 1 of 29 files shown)


Glowbal:
I am pretty certain there is a reason we didn't localize those. There is a function from the CBA keybinding API to make the name look pretty but we never got around to use it.

Glowbal:
I did a quick check: I believe it was because when changing the game language, you would lose all keybinding settings because it would be classified as a different mod.

Instead, we should use `cba_fnc_registerKeybindModPrettyName`.

```sqf
[MOD_NAME, LOCALIZED_STR] call cba_fnc_registerKeybindModPrettyName;
```
ref: https://github.com/CBATeam/CBA_A3/blob/master/addons/keybinding/fnc_registerKeybindModPrettyName.sqf

MikeMatrix:
@Glowbal makes sense, couldn't really test it, since I'm running English only. I'll update to use that function instead.

MikeMatrix:
Woops wrong button

jonpas:
New PR might be better anyways. :)

MikeMatrix:
Yea true. I'll work on this and get back with a new PR

PR #2039, opened 2015-08-06
Medical menu

Glowbal:
- Initial port of the CSE medical menu.
- Works with both basic and advanced medical.
- Runs along side the current 3d interaction menu.
- Has a keybinding with same functionality as CSE had.
- Pulls treatment actions straight from the treatment config in the medical module.

Medical Menu

**To do list:**

- [x] Currently doesn't have the full log functionality as was available in CSE due to the implementation being different in our medical system. This may still require some more thought.
- [x] Also the code still requires some more love. At time of writing it's been written to be functional asap.
- [x] Some images are duplicates of the ones in the ACE medical pbo.
- [x] Ensure we got all settings we want for this.


jokoho48:
Like that :+1:

SzwedzikPL:
Awesome :+1:

addons/medical_menu/functions/fnc_handleUI_DisplayOptions.sqf
```diff
@@ -0,0 +1,88 @@
...
+    _ctrl ctrlCommit 0;
+};
+
+GVAR(LatestDisplayOptionMenu) = _name;
+
+// The triage card has no options available
+lbClear 212;
+if (_name == "triage") exitwith {
+    ctrlEnable[212,true];
+    _card = ([GVAR(INTERACTION_TARGET)] call FUNC(getTriageList));
```
MikeMatrix: `FUNC(getTriageList)` Where is that defined?

addons/medical_menu/functions/fnc_updateBodyImage.sqf
```diff
@@ -0,0 +1,41 @@
...
+_availableSelections = [50,51,52,53,54,55];
+{
+    private ["_red", "_green", "_blue"];
+    _total = _x;
+
+    _red = 1;
+    _green = 1;
+    _blue = 1;
+    if (_total > 0) then {
+        if (_damaged select _forEachIndex) then {
```
MikeMatrix: Where is `_damaged` coming from? This seems to be not defined at all.

addons/medical_menu/functions/fnc_onMenuOpen.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+[QGVAR(onMenuOpen), "onEachFrame", {
+    if (isNull GVAR(INTERACTION_TARGET)) then {
+        GVAR(INTERACTION_TARGET) = ACE_player;
+    };
+    [GVAR(INTERACTION_TARGET), _this select 0] call FUNC(updateUIInfo);
+    [GVAR(INTERACTION_TARGET)] call FUNC(updateIcons);
+    [GVAR(LatestDisplayOptionMenu)] call FUNC(handleUI_DisplayOptions);
+
+    _status = [GVAR(INTERACTION_TARGET)] call FUNC(getTriageStatus);
```
MikeMatrix: I'd imagine this should be `EFUNC(medical,getTriageStatus)` as with the other instance of this function name in `medical_menu`?

PR #2040, opened 2015-08-06
Slideshow Fixes

### LINKED ISSUE #1945: Slideshow should have failsafes when interaction is not present
This is more of a note for myself, but also one question, `ace_slideshow` doesn't require interaction, as you can set it to run automatically which doesn't add the actions and therefore no function to `ace_interaction` is called. Should I just add an if check for CfgMods (or whatever it's called) or is there another way this should be handled?

### LINKED ISSUE #2037: Slideshow does not work on dedicated server
Hey guys,

I have created a slideshow with some images on a laptop. It works very fine in the editor preview but when I play this map on a dedicated server it does not work because I do not get the option on the laptop to switch the images.

Please have a look on this, thank you :)

jonpas:
Needs testing. Fixes #1945, fixes #2037.


jonpas:
Ready for merge, tested on dedicated server, thanks @jokoho48 for help.

There is an issue which I added to documentation and mission-tools which renders non-PBOed missions incompatible with the Slideshow module due to not broadcasting the correct path (only when `setObjectTextureGlobal` is used on the server and is referencing a .paa from a non-PBOed mission) as explained in the BI feedback tracker ticket.

Glowbal:
This has merge conflicts

jonpas:
@Glowbal Conflicts resolved

Knight1981:
Thank you for fixing this and for your fantastic work. Hope it will work in the next version :) 

Do you have any information when the next release will be available? I want to test it. :D

jonpas:
When it'll be ready. :)

PR #2052, opened 2015-08-07
ACE specific logging macros

MikeMatrix:
Proposal to add these macros in, for the purpose of maintaining the same style of logging when using `diag_log` (seen in a few places), while adding the module name to generate general purpose log entries.


Cuel:
Doesn't this already happen with full debug ? From CBA

jonpas:
I believe this macros are for non-debug logging, like when a mission module loads it logs that to the PR, currently it's all written out in code directly and not really consistent either.

addons/main/script_macros.hpp
```diff
@@ -97,5 +97,9 @@
 // Time functions for accuracy per frame
 #define ACE_tickTime (ACE_time + (diag_tickTime - ACE_diagTime))
 
+#define ACE_LOG(module,level,message) diag_log text format [QUOTE([ACE] (%1) %2: %3), module, level, message];
+#define ACE_LOGERROR(message) ACE_LOG(QUOTE(COMPONENT),"ERROR",message)
```
SilentSpike: Instead of using `QUOTE` you can just use the `#` prefix

MikeMatrix: That won't work. Just tested it and it would result in it being `#common` instead of `"common"` for example.
I assume it's because it's inside another macro call? Not sure on that though.

Glowbal: How about this:

```sqf
#define ACE_LOG(module,level,message) diag_log text format [QUOTE([ACE] (module) level: %1), message];
#define ACE_LOGERROR(message) ACE_LOG(COMPONENT,ERROR,message)
```

MikeMatrix: Yea, that should work fine. Seems cleaner as well. I'll test it later and if it works, I'll update the PR

Glowbal:
@MikeMatrix What is the status on this?

MikeMatrix:
@Glowbal should be done.

ViperMaul:
@MikeMatrix any more?

MikeMatrix:
@ViperMaul The only thing I am considering, but not sure about yet, is weather or not I should add `DEBUG` log levels...

MikeMatrix:
Makes sense to have `DEBUG` level as well. So I added it.

ViperMaul:
How do you turn on/off if macro statements like ACE_LOGDEBUG is committed in code?

MikeMatrix:
I don't think it makes sense in the context of `INFO`, `WARNING` or `ERROR` for ACE itself.
That's why I was unsure about `DEBUG`, because it feels like there needs to be some way to disable that output, which would make it redundant, since we have things like `TRACE` from CBA.
I'm really not sure. Right now it really feels like we have a lot of different solutions solving their own little issues, but never fit the entire thing.

I think this might be a good step to prepare for a better solution upcoming.
I'm gonna think about making some better logging system for CBA, so we can utilize that one directly and slowly deprecate and remove all custom solutions that we have in ACE atm (which is surprisingly many...)

bux578:
Are you able to define a log level? Like show only errors, show errors and warnings, show errors, warnings and Info.

MikeMatrix:
@bux578 RIght now there is no such thing, but as I was saying, I'm thinking of improving the CBA framework with something like that.

PR #2056, opened 2015-08-08
Code cleanup of Sandbag module

MikeMatrix:
Code cleanup of the Sandbag module.


Glowbal:
This looks :+1:

SilentSpike:
RE: The empty array thing, it actually does matter since in functions with optional parameters you still need to provide at least an empty array (`params` will error if `nil` is used)

MikeMatrix:
That's good to know. Either way, I feel like this is the way to go with examples for documentation.

PR #2065, opened 2015-08-08
Setting to disable hearing for zeus RC

PabstMirror:
Feature request from @SzwedzikPL
Add setting for hearing to effect zeus RC - (part of #1961)


SzwedzikPL:
Thank you @PabstMirror ! :)

SilentSpike:
Hmm, what if we were to introduce a common setting along the lines of: "ACE systems apply to remote controlled units"?

I've noticed we introduced a similar setting for something in medical. Or should it be done on a case-by-case basis?

PR #2066, opened 2015-08-08
Fix and enhancement of two common functions

SilentSpike:
- The `disableUserInput` function allows for use of chat if user has admin access, however it didn't account for the host of listen servers (where `serverCommandAvailable` returns false).
- Expanding the `isFeatureCameraActive` function to support the curator camera and the upcoming ace_spectator camera. Also optimized the lazy evaluation and updated the function header.


jonpas:
:+1:

PR #2081, opened 2015-08-09
Fixed #1893 - Menu closes when patient is more as 10 meters away

### LINKED ISSUE #1893: Medical window stays on when you move away from patient and still hold down interaction key, also happens with self interaction menu
ACE3 Version: 3.1.1

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *Advanced medical*

**Description:**
*Medical window should turn off when you select interaction option not related to medical, or if you move away from player. It can possibly kill FPS by a little.*

**Steps to reproduce:**
* *Open medical interaction menu on someone, notice how medical window shows up on the left*
* *While you hold interaction key, try to:*
    * *Select "Interaction" with this player (to join group etc.), tap shoulder etc. any other interaction option but no medical ones*
    * *Start moving away, still holding interaction key, notice that medical window is still on*
* *In both above situations medical window will still be on.*
* *It also happens on self interaction menu - select medical, switch to any other option.*

*"Multiplayer", "Singleplayer"*

**RPT log file:**
*--*

Glowbal:


SilentSpike:
I noticed there was a function in common that seems related to this:

gienkov:
Isn't 10 meters too far?

PR #2087, opened 2015-08-09
Remove CPR action from dead people

gienkov:
You can do CPR on units that are dead - this should fix that.

### DIFF
--- a/addons/medical/ACE_Medical_Treatments.hpp
+++ b/addons/medical/ACE_Medical_Treatments.hpp
@@ -316,7 +316,7 @@ class ACE_Medical_Actions {
             requiredMedic = 0;
             treatmentTime = 15;
             items[] = {};
-            condition = "!([(_this select 1)] call ace_common_fnc_isAwake)";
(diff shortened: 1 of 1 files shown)


Glowbal:
This would give people another indication whatever or not someone is death.

bux578:
^

gienkov:
Well plain indication that someone is clearly dead is ragdoll...

BryanBoru:
Hold on, so we want it that the only way to do CPR is if there is a chance to save them? In Reality CPR has a slim success rate at 18%. However, in game it seems if you hit the magic mix of epi and CPR they almost always come back, quite unrealistic as it is. So if the concern is "people will waste time doing CPR on the dead" well, who else would get CPR but the dead? I think it will help players understand the futility of CPR. Leave it to them to call off the CPR after a few rounds of failure, especially if they don't notice the ragdall (which I wish would just happen at unconsciousness since there should never be a hint of unsurvivable besides the level of trauma a patient has).

jonpas:
:+1: for new label!

PR #2094, opened 2015-08-09
Code cleanup of Finger module.

jokoho48:
Code cleanup of Fingering module is Incomming.


addons/finger/functions/fnc_keyPress.sqf
```diff
@@ -50,7 +50,7 @@ _nearbyMen = (ACE_player nearObjects ["CAManBase", (GVAR(maxRange) + 2)]);
 
         _sendFingerToPlayers pushBack _x;
     };
-} forEach _nearbyMen;
+} count _nearbyMen;
```
MikeMatrix: AFAIK pushBack has a return value. Maybe add a `nil`?

jokoho48: ya overlooked

jokoho48:
Done

PR #2101, opened 2015-08-10
Make finger enabled by default

gienkov:
Why it's disabled in first place?

### DIFF
--- a/addons/finger/ACE_Settings.hpp
+++ b/addons/finger/ACE_Settings.hpp
@@ -1,6 +1,6 @@
 class ACE_Settings {
     class GVAR(enabled) {
-        value = 0;
+        value = 1;
         typeName = "BOOL";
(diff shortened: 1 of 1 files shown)


BryanBoru:
agreed, same with sitting.

PabstMirror:
Right now mission makers enable it as they see fit.
You are also able to make a ace_server.pbo setting to always enable it for your server.

If we flip it on by default, many will just have to go through these same steps to disable.
The giant floating virtual circles may not be right for all comunites.

austinjames314:
Yeah, but if it's enabled, you don't have to use it. And if you have it enabled and never use it, how is that different to having it disabled?
On the flipside if it's disabled, and you _want_ to use it, you can't. The two situations are not equal.
I too think it they should be enabled by default.

PR #1657, opened 2015-06-17
Trenches

Grey-Soldierman:
Ports 2 A2 trenches; Envelope Small & Envelope Big. 
Adds an entrenching tool based off an A2 shovel from the Land Rover. 
Someone smarter than me could combine the two to be able to deploy the trenches using the entrenching tool.

PR #2104, opened 2015-08-10
Trenches

Glowbal:
Based on #1657 @Grey-Soldierman 

> Ports 2 A2 trenches; Envelope Small & Envelope Big.
> Adds an entrenching tool based off an A2 shovel from the Land Rover.
> Someone smarter than me could combine the two to be able to deploy the trenches using the entrenching tool. 

I will continue work on this.


esteldunedain:
Hey @Glowbal, I hope you don't mind a little help here. I did a first code implementation of `ace_trenches`, It's practically a ripoff of @ulteq 's `ace_sandbag`, so kudos to him. It obviously can be improved, but it seems to work.

I don't have the know-how to edit the models, so I'm making this requests to anyone willing to help:
- [ ] Make a version of both models without geometry, to avoid some issues when placing.
- [ ] If possible, extend the geometry ~ 0.5 m downward, so we can make the trenches protrude a little more on slopes.
- [ ] Possibly change the texture so it doesn't look grassy (or have multiple versions)

Some screenshots:

Glowbal:
@esteldunedain not at all, thanks for the help!

esteldunedain:
I managed to do the modifications of the models myself:

- Trenches are a little bigger and deeper.
- Replaced the Roadway LOD by a Geometry LOD. It means that you cannot climb to it without pressing V, but it was neccesary to avoid the player being able to clip the head through the model to see through.
- Made versions without geometru.

I also changed the placement algorithm to align the trench with the terrain, which works much better on slopes than keeping them horizontal.

Things that could be worked on (before or after merging):
- [ ] Better textures
- [ ] A progressbar + animation so the placement is not instantaneous.

jokoho48:
> Possibly change the texture so it doesn't look grassy (or have multiple versions)
yes Posible but need a new texture

jokoho48:
@esteldunedain you miss to change the LandContact Points

jokoho48:
and if we add Normal map and re UV this look better

esteldunedain:
> @esteldunedain you miss to change the LandContact Points

Why would I do that?

esteldunedain:
> Model Config and Hiddenselection is fast make but texture not :/

I sure arma has a dirt texture somewhere we can use. I've also found OSS textures online that could work just fine. But honestly I didn't know how to swap them.

jokoho48:
i think a new uv is requried here

esteldunedain:
> i think a new uv is requried here

IDK about required, but it certainly could make the thing look much better,

jokoho48:
i mean you can basic swap this but i think the texture it self is fine but the UV is crap and a texture change is not the problem solution

esteldunedain:
> i mean you can basic swap this but i think the texture it self is fine but the UV is crap and a texture change is not the problem solution

The texture looks fine for Stratis, but it's not that appropiate for other terrains (e.g. desert). Something that looked more like dirt without grass might be better overall.

jokoho48:
no i think its look bad in generell

esteldunedain:
you: i think the texture it self is fine

me: The texture looks fine for Stratis, but it's not that appropiate for other terrains

you: no i think its look bad in generell

jokoho48:
yes because the UV is fucked up that is why the texture is so stretched at the sides

esteldunedain:
I know @jokoho48. The UV unwrapping could be improved AND also the texture could be replaced to better fit all terrains.

SilentSpike:
Soil is probably the texture that'd fit most terrains, however you'd probably also want a sand alternative too.

SzwedzikPL:
> A progressbar + animation so the placement is not instantaneous.

I think nice "animation" can be setting position of object slightly underground and changing it every 10-20% to look it like "growing" from ground.

> Soil is probably the texture that'd fit most terrains, however you'd probably also want a sand alternative too.

Can it be defined in easy to edit place? CfgWorlds maybe?

esteldunedain:
> I think nice "animation" can be setting position of object slightly underground and changing it every 10-20% to look it like "growing" from ground.

:+1:

esteldunedain:
Added animations and progressbar.

https://www.youtube.com/watch?v=wH1fyBJATJo

Unless something big comes out I'll be merging this later as per agreed with @Glowbal.

esteldunedain:
Ups, I forgot localization. I'll do that tomorrow.

esteldunedain:
Done

SzwedzikPL:
> https://www.youtube.com/watch?v=wH1fyBJATJo

Looks awesome! Great job @esteldunedain

esteldunedain:
Ticket with extra tasks to be completed eventually: #3263

licht-im-Norden87:
imho the building process is waaaay to fast. In this case you can dig out your trench in almost no time making firefights unrealisitc.

esteldunedain:
Yes, of course build time is many times faster than rl (it would take be 1/2 hr+). But do you really see people doing this in a middle of a firefight? 

If it really gets abused I would increase the duration or provide a setting.

Bummeri:
Agreed that is is way too fast. Would it be possible to require multiple interactions with the cover to gradually make it higher and higher? Also by interacting with it, would it be possible to change its texture and\or add some camouflage to it?

ArcanumSK:
At this stage, I am afraid that standard contact reaction would be to arm with shovels.

Issue #2098, opened 2015-08-10
Can't change option value on category list


SzwedzikPL:
ACE3 Version: https://github.com/acemod/ACE3/commit/0c2c71e70ea80c0c83584b4bd487ec7d1ad199ed

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
ACE Debug data in rpt file

**Description:**
https://www.youtube.com/watch?v=WupgOxNf42k

**Steps to reproduce:**
https://www.youtube.com/watch?v=WupgOxNf42k

**Where did the issue occur?**
*"Multiplayer", "Singleplayer"*

**RPT log file:**
http://pastebin.com/fy6hJemX

PR #2106, opened 2015-08-10
Option Menu Fixes for categories

PabstMirror:
#2098

We need the filtered selection index in about 6 different places, so just store it in the lbValue.

### DIFF
--- a/addons/optionsmenu/functions/fnc_onListBoxSettingsChanged.sqf
+++ b/addons/optionsmenu/functions/fnc_onListBoxSettingsChanged.sqf
@@ -16,14 +16,14 @@
 
 #include "script_component.hpp"
 
-private ["_settingsMenu", "_settingIndex", "_rightDropDownIndex"];
+private ["_settingIndex", "_rightDropDownIndex"];
 
 _rightDropDownIndex = lbCurSel 400;  //Index of right drop down
 if (_rightDropDownIndex < 0) then {_rightDropDownIndex = 0;};
 
 _settingIndex = -1;
-if (((lbCurSel 200) >= 0) && {(lbCurSel 200) < ((lbSize 200)/2)}) then {
-    _settingIndex =  lbValue [200, (lbCurSel 200)];
+if (((lnbCurSelRow 200) >= 0) && {(lnbCurSelRow 200) < ((lnbSize 200) select 0)}) then {
+    _settingIndex =  lnbValue [200, [(lnbCurSelRow 200), 0]];
 };
 if (_settingIndex == -1) exitWith {};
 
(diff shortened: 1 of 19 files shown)


addons/optionsmenu/gui/settingsMenu.hpp
```diff
@@ -277,7 +277,7 @@ class ACE_settingsMenu {
 };
 class ACE_serverSettingsMenu: ACE_settingsMenu {
     onLoad = QUOTE(uiNamespace setVariable [ARR_2('ACE_serverSettingsMenu', _this select 0)]; [] call FUNC(onServerSettingsMenuOpen););
-    onUnload = QUOTE(uiNamespace setVariable [ARR_2('ACE_serverSettingsMenu', nil)];);
+    onUnload = QUOTE(uiNamespace setVariable [ARR_2('ACE_serverSettingsMenu', nil)]; [] call FUNC(onSettingsMenuOpen););
```
Glowbal: Why call onSettingsMenuOpen when we close the menu?

addons/optionsmenu/functions/fnc_onListBoxSettingsChanged.sqf
```diff
@@ -16,26 +16,30 @@
...
-private ["_settingIndex", "_rightDropDownIndex"];
+private ["_settingsMenu", "_settingIndex", "_rightDropDownIndex"];
 
-_settingIndex =  lbCurSel 200;  //Index of left list
 _rightDropDownIndex = lbCurSel 400;  //Index of right drop down
-
 if (_rightDropDownIndex < 0) then {_rightDropDownIndex = 0;};
 
+_settingIndex = -1;
+if (((lbCurSel 200) >= 0) && {(lbCurSel 200) < ((lbSize 200)/2)}) then {
```
Glowbal:
```sqf
(lbCurSel 200) < ((lbSize 200)/2)}
```
Won't this mean we can never select an item that has the index above half of the items?

addons/optionsmenu/gui/settingsMenu.hpp
PabstMirror: The original setting menu is still open when we close the server menu, and it acts weird.  We could also just close it when we open the server settings.

Glowbal: Closing it would be prefered.

addons/optionsmenu/functions/fnc_onListBoxSettingsChanged.sqf
PabstMirror: `listNBox` is weird, the count is double the number of rows.

Glowbal: Oh right, probably because of the columns we use?

PabstMirror: We were using listBox commands on a listBoxN.
It worked, but probably safer and easier to read to swap to the proper commands.

Issue #2045, opened 2015-08-07
No "Push" Interaction Menu for SDV


ViperMaul:
@pmc-tyler reports
When SDV is touching land, you have Interaction Menu, however you do not get the push

PR #2109, opened 2015-08-11
Push SDV sub, cleanup push

PabstMirror:
#2045
Note: sub couldn't be pushed because it's `mass` is 2600 and we had a limit of 1000.
2600 is more than twice a hatchback, mass is kinda wonky in this game.
I just upped it to 2600.  The big big assault boat is 9600, so it still cannot be pushed.

Also switched to use event system.

### DIFF
--- a/addons/interaction/CfgVehicles.hpp
+++ b/addons/interaction/CfgVehicles.hpp
@@ -501,8 +501,8 @@ class CfgVehicles {
                 class ACE_Push {
                     displayName = CSTRING(Push);
                     distance = 6;
-                    condition = QUOTE(getMass _target < 1000 && {alive _target});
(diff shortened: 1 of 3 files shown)


Glowbal:
A cleaner solution would be to use a different mass limit if the object is floating in water.

bux578:
^ you can push very heavy stuff that's floating

SilentSpike:
Do configs have some sort of buoyancy value?

jonpas:
Recently BI added buoyancy values IIRC.

cptnnick:
There have been for a while, just the Buoy object didm't have a config value for buoyancy, which they fixed.

SilentSpike:
Then perhaps we could use that value and the mass value to calculate a representation of how hard it is to push

jonpas:
Well, if the object is entirely in the water, then you can really push around insanely heavy objects (just difference in speed), what this function must do is make sure yiu can't push something insanely heavy when it's eg. only halway in the water.

PR #2113, opened 2015-08-11
Code cleanup Safemode module

MikeMatrix:
Code Cleanup of the Safemode module

### DIFF
--- a/addons/safemode/functions/fnc_playChangeFiremodeSound.sqf
+++ b/addons/safemode/functions/fnc_playChangeFiremodeSound.sqf
@@ -1,9 +1,25 @@
-// by commy2
+/*
+ * Author: commy2
+ * Play weapon firemode change sound.
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ * 1: Weapon <STRING>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [ACE_player, currentWeapon ACE_player] call ace_safemode_fnc_playChangeFiremodeSound
+ *
+ * Public: No
+ */
 #include "script_component.hpp"
 
-PARAMS_2(_unit,_weapon);
+private ["_sound", "_position", "_soundCount"];
 
-private ["_sound"];
+params ["_unit", "_weapon"];
+
 _sound = getArray (configFile >> "CfgWeapons" >> _weapon >> "changeFiremodeSound");
 
 if (count _sound == 0) exitWith {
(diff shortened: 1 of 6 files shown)


Glowbal:
I think this looks good :+1:

addons/safemode/functions/fnc_playChangeFiremodeSound.sqf
```diff
@@ -16,12 +32,13 @@ if ({(toLower (_sound select 0) find _x == (count toArray (_sound select 0) - co
 };
 
 // add default volume, pitch and distance
-if (count _sound < 2) then {_sound pushBack 1};
-if (count _sound < 3) then {_sound pushBack 1};
-if (count _sound < 4) then {_sound pushBack 0};
+_soundCount = count _sound;
+if (_soundCount < 2) then {_sound pushBack 1};
+if (_soundCount < 3) then {_sound pushBack 1};
+if (_soundCount < 4) then {_sound pushBack 0};
```
commy2: This is wrong. `count _sound` increases after `pushBack`, but `_soundCount` does not.

addons/safemode/functions/fnc_playChangeFiremodeSound.sqf
```diff
@@ -16,12 +32,13 @@ if ({(toLower (_sound select 0) find _x == (count toArray (_sound select 0) - co
...
 
-private "_position";
 _position = _unit modelToWorldVisual (_unit selectionPosition "RightHand");
 _position set [2, (_position select 2) + ((getPosASLW _unit select 2) - (getPosATL _unit select 2) max 0)];
 
-playSound3D [_sound select 0, objNull, false, _position, _sound select 1, _sound select 2, _sound select 3];
+_sound params ["_filename", "_volume", "_soundPitch", "_distance"]
+playSound3D [_filename, objNull, false, _position, _volume, _soundPitch, _distance];
```
commy2: Whate the point? Isn't this slower?

addons/safemode/functions/fnc_playChangeFiremodeSound.sqf
MikeMatrix: This might be true, doesn't change the outcome of this chain of conditions though, considering that if any of them are hit, all of the following are thus logically also executing.
Example:
`count _sound` is 2. meaning L36 doesn't hit, but L37 hits. It increases `count _sound` by 1, bringing it to 3, thus L38 also hits.

addons/safemode/functions/fnc_playChangeFiremodeSound.sqf
MikeMatrix: No difference in performance, but it is self documenting.

MikeMatrix: Although there is a `;` missing... I'll fix that in a sec.

addons/safemode/functions/fnc_playChangeFiremodeSound.sqf
MikeMatrix: Actually this is completely not required at all anyways.

PR #2116, opened 2015-08-11
ST Stamina Bar integration

jonpas:
Cleaned up code, resolved "Error loading control" errors, removed 2 unnecessary controls and some config properties, support for remote controlled units - changes from ST's latest version.

Thanks @dslyecxi for the ST version obviously!


jonpas:
Any info on this?

gienkov:
Need tester?
25 sie 2015 4:35 AM "jonpas" <notifications@github.com> napisał(a):

jonpas:
Sure, go for it.

ShuraDE:
no trouble (log/game), works fine here

addons/staminabar/functions/fnc_moduleInit.sqf
```diff
@@ -0,0 +1,25 @@
...
+
+if !(isServer) exitWith {};
+
+params ["_logic", "_units", "_activated"];
+
+if (!_activated) exitWith {};
+
+[_logic, QGVAR(enabled), "enabled"] call EFUNC(common,readSettingFromModule);
+
+diag_log text "[ACE]: Stamina Bar Module Initialized.";
```
jonpas: Replace with new log macro.

bux578:
https://twitter.com/ReMuuf/status/654662671933943808

jonpas:
http://dev.arma3.com/post/oprep-core-mechanics-refinement

I guess this can be closed then? Or wait for release of it and see about it then?

MikeMatrix:
I'm pretty sure this has been rendered obsolete by the developments in the vanilla game.

jonpas:
I agree, closing.

PabstMirror:
I always appreciated the fact that the ST Stamina was very subtle and not distracting unless you were looking for it.

I don't really like the look of the new bar, and I'm not sure I like the changes from "fatigue" to "stamina" (unlimited running)

Depending on how 1.54 feels, we may want to consider looking into seriously changing fatigue on our own, and this PR could still be a part of that, so you may not want to delete it just yet.

jonpas:
How long can you still restore a branch?

I am pretty sure it'll be better to modify the vanilla bar rather than implementing another though.

MikeMatrix:
I agree with @jonpas here. If it comes down to us deciding the new stamina system is not performing to our expectations, then we can craft a better suited solution for that case. I don't think this solution with fit quite as well.

On other note, I think the new stamina bar actually makes a lot of sense for what it's trying to do.

Cyruz143:
Just a quick side note (if anyone's interested). We've kept the ST bar over the vanilla one as it's a lot less intrusive/obstructive, especially when combined with disabling a lot of the UI. I don't think it wouldn't be too bad to have the option between both.

WadeDP:
This would be cool with #1857

PR #2130, opened 2015-08-12
Code cleanup of Gforces module

jokoho48:
Code cleanup of Gforces module.

### DIFF
--- a/addons/gforces/XEH_postInit.sqf
+++ b/addons/gforces/XEH_postInit.sqf
@@ -12,4 +12,4 @@ GVAR(GForces_CC) ppEffectCommit 0.4;
 GVAR(lastUpdateTime) = 0;
 GVAR(oldVel) = [0,0,0];
 
-[FUNC(pfhUpdateGForces), 0, []] call CBA_fnc_addPerFrameHandler;
+[FUNC(pfhUpdateGForces), INTERVAL, []] call CBA_fnc_addPerFrameHandler;
--- a/addons/gforces/XEH_postInit.sqf
+++ b/addons/gforces/XEH_postInit.sqf
@@ -9,7 +9,6 @@ GVAR(GForces_CC) ppEffectForceInNVG true;
 GVAR(GForces_CC) ppEffectAdjust [1,1,0,[0,0,0,1],[0,0,0,0],[1,1,1,1],[10,10,0,0,0,0.1,0.5]];
 GVAR(GForces_CC) ppEffectCommit 0.4;
 
-GVAR(lastUpdateTime) = 0;
 GVAR(oldVel) = [0,0,0];
 
 [FUNC(pfhUpdateGForces), INTERVAL, []] call CBA_fnc_addPerFrameHandler;
--- a/addons/gforces/XEH_postInit.sqf
+++ b/addons/gforces/XEH_postInit.sqf
@@ -11,4 +11,4 @@ GVAR(GForces_CC) ppEffectCommit 0.4;
 
 GVAR(oldVel) = [0,0,0];
 
-[FUNC(pfhUpdateGForces), INTERVAL, []] call CBA_fnc_addPerFrameHandler;
+[DFUNC(pfhUpdateGForces), INTERVAL, []] call CBA_fnc_addPerFrameHandler;
--- a/addons/gforces/functions/fnc_pfhUpdateGForces.sqf
+++ b/addons/gforces/functions/fnc_pfhUpdateGForces.sqf
@@ -3,8 +3,7 @@
  * Calculates average g-forces and triggers g-effects
  *
  * Argument:
- * 0: Arguments <ARRAY>
- * 1: pfh_id <NUMBER>
+ * None
  *
  * Return value:
  * None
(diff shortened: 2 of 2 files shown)


addons/gforces/functions/fnc_pfhUpdateGForces.sqf
```diff
@@ -86,11 +80,12 @@ if (_suitCoef == 0) then {_suitCoef = 0.001};
 
 _gBlackOut = MAXVIRTUALG / _classCoef + MAXVIRTUALG / _suitCoef - MAXVIRTUALG;
 _gRedOut = MINVIRTUALG / _classCoef;
-
-["GForces", [], {format ["_g _gBO _coef: %1, %2, %3", _average, _gBlackOut, 2 * ((1.0 - ((_average - 0.30 * _gBlackOut) / (0.70 * _gBlackOut)) ^ 2) max 0) ]}] call EFUNC(common,log);
+#ifdef DEBUG_MODE_FULL
+    ["GForces", [], {format ["_g _gBO _coef: %1, %2, %3", _average, _gBlackOut, 2 * ((1.0 - ((_average - 0.30 * _gBlackOut) / (0.70 * _gBlackOut)) ^ 2) max 0) ]}] call EFUNC(common,log);
+#endif
```
jonpas: Can `TRACE_1` be used here? Would need a bit of a refactor to the debug code probably.

jokoho48: i think thats fine here

addons/gforces/functions/fnc_pfhUpdateGForces.sqf
```diff
@@ -64,16 +55,19 @@ GVAR(oldVel) = _newVel;
...
+        true
+    } count GVAR(GForces);
+    _average = _sum / _sumDif;
 };
 
-_classCoef = ACE_player getVariable ["ACE_GForceCoef",
-    getNumber (configFile >> "CfgVehicles" >> (typeOf ACE_player) >> "ACE_GForceCoef")];
+_classCoef = ACE_player getVariable "ACE_GForceCoef";
+if (isNil "_classCoef") then {
+    _classCoef = getNumber (configFile >> "CfgVehicles" >> (typeOf ACE_player) >> "ACE_GForceCoef")];
```
PabstMirror: `]` at end

addons/gforces/functions/fnc_pfhUpdateGForces.sqf
```diff
@@ -13,26 +12,18 @@
...
 
-if (isNull ACE_player) exitWith {};
+ if !(alive ACE_player) exitWith {};
 
-if !(alive ACE_player) exitWith {};
+private ["_player", "_newVel", "_accel", "_currentGForce", "_average", "_sum", "_classCoef", "_suitCoef", "_gBlackOut", "_gRedOut", "_g", "_gBO", "_coef", "_strength"];
 
 GVAR(lastUpdateTime) = ACE_time;
-
-/*if !(vehicle ACE_player isKindOf "Air") exitWith {
```
PabstMirror: If we are in the middle of a blackout, and eject from a air vehicle, the effect will never go away.

jokoho48: ah now i understand yeah change it

addons/gforces/functions/fnc_pfhUpdateGForces.sqf
```diff
@@ -64,16 +49,19 @@ GVAR(oldVel) = _newVel;
 */
 
 _average = 0;
-if (count GVAR(GForces) > 0) then {
+if (!isNil {GVAR(GForces) select 0}) then {
```
SilentSpike: This is bad, I suggest `!(GVAR(GForces) isEqualTo [])`

jokoho48: greate Idea

addons/gforces/functions/fnc_pfhUpdateGForces.sqf
```diff
@@ -64,16 +49,19 @@ GVAR(oldVel) = _newVel;
...
-    _average = _sum / (count GVAR(GForces));
+        true
+    } count GVAR(GForces);
+    _average = _sum / _sumDif;
 };
 
-_classCoef = ACE_player getVariable ["ACE_GForceCoef",
-    getNumber (configFile >> "CfgVehicles" >> (typeOf ACE_player) >> "ACE_GForceCoef")];
+_classCoef = ACE_player getVariable "ACE_GForceCoef";
+if (isNil "_classCoef") then {
```
Glowbal: I don't think this is cleaner as the previous code. As for performance increase; I expect we now got an if statement that is true more often then not. If anything, I'd expect this to be worse on average.

esteldunedain: I agree.

addons/gforces/XEH_postInit.sqf
```diff
@@ -9,7 +9,6 @@ GVAR(GForces_CC) ppEffectForceInNVG true;
 GVAR(GForces_CC) ppEffectAdjust [1,1,0,[0,0,0,1],[0,0,0,0],[1,1,1,1],[10,10,0,0,0,0.1,0.5]];
 GVAR(GForces_CC) ppEffectCommit 0.4;
 
-GVAR(lastUpdateTime) = 0;
 GVAR(oldVel) = [0,0,0];
 
-[FUNC(pfhUpdateGForces), 0, []] call CBA_fnc_addPerFrameHandler;
+[DFUNC(pfhUpdateGForces), INTERVAL, []] call CBA_fnc_addPerFrameHandler;
```
esteldunedain: Although marginally more efficient, this will break the calculation when time is accelerated.

PR #2133, opened 2015-08-12
Mk 1 eyeballs Integration

jonpas:
Integration of [Mk 1 eyeballs](https://forums.bistudio.com/topic/183606-mk-1-eyeballs/) per [granted permission](https://forums.bistudio.com/topic/183606-mk-1-eyeballs/?p=2900364).

This adds another option to View Distance setting `Dynamic Object View Distance` called `FoV Based`.
 
(OVD - Object View Distance, VD - View Distance)
When this is enabled, OVD is dynamically changed based on the Field of View (Zoom). Minimal value is OVD, maximum is VD from vanilla options, when the option is selected, you cannot change minimal value any more, but you can do so with maximum.
This means the difference between no zoom and full zoom is bigger when OVD and VD are further apart (VD much higher than OVD), or smaller if OVD and VD are closer together, or none if OVD is equal to VD.

During all this I also cleaned up the module a bit, most notably changed postInit to run only on clients (so they don't even get to the `hasInterface` check).

- [x] Add original author to AUTHORS.txt
- [x] Further view distance in `Air` vehicles


addons/viewdistance/stringtable.xml
```diff
@@ -114,7 +114,7 @@
             <Portuguese>Distância de visão dinâmica dos objetos</Portuguese>
         </Key>
         <Key ID="STR_ACE_ViewDistance_object_Description">
-            <English>Sets the object view distance as a coefficient of the view distance.</English>
+            <English>Sets the object view distance as a coefficient of the view distance or based on field of view. FoV Based's lowest and highest value is the value which is set when enabling the option.</English>
```
jonpas: Not sure what the process is for changing strings, remove the rest or hunt for translators? :grimacing:

robalo:
hasInterface is still useful for headless clients

jonpas:
> hasInterface is still useful for headless clients

It's still there, don't worry. :grinning:

Winter259:
Sorry I was radio silent on this, have too many projects on hand right now. Gave the code a one over, looks good :+1:

jonpas:
Currently, view distance limiter does NOT work for anything else but non-video setting options of the 3 vehicle type settings (this PR is not the cause). I will fix this in a separate PR if someone doesn't beat me.

jonpas:
Added separate handling for `Air` vehicles, it basically just makes you see further with FoV Based dynamic view distance setting when you are in an air vehicle. Also fixed normal zoom level (as it actually is without zooming in/out).

This is finished, feel free to test and advise if values are good.

jonpas:
Any info on this?

jokoho48:
why you using a Mix out of 2 Variable Naming

jonpas:
> why you using a Mix out of 2 Variable Naming Standarts?

Those that have a lot of undersocres are @Winter259 's, can cleanup in a separate PR.

jonpas:
What's the status on this?

addons/viewdistance/stringtable.xml
Fusselwurm: In case you're looking for translations, this is what it should it be for German:

> Passt die Objektsichtweite dynamisch der Sichtweite an oder setzt sie auf Grundlage des Sichtfeldes (FoV). 
> Wenn in Abhängigkeit vom Sichtfeld, dann wird der für "Objektsichtweite" eingestellte Wert das Minimum, und die eingestellte "Sichtweite" das Maximum der Objektsichtweite.

jonpas:
Any update?

jonpas:
If someone could test how this works with multiplayer that would be great, I know view distance is a bit funky when server is set to default. @esteldunedain @Winter259 any information about that that I can use?

Fusselwurm:
@jonpas fwiw, [I rebased your branch onto current master](https://github.com/Fusselwurm/ACE3/tree/jonpas_viewdsitance_on_Master) and [played around on a dedi server](https://youtu.be/tXv4ei5L7kI) (where you usually cannot set your view distance via video settings). Looked ok to me.

jonpas:
That's awesome @Fusselwurm, thanks for testing!

PR #2137, opened 2015-08-13
CheckPBO Fix endless waitUntil if checkAll is false

PabstMirror:
In MP clients will wait forever for `ACE_Version_ClientErrors` in a spawned loop
which can only be set if GVAR(checkPBOsCheckAll) is true and checkVersionNumber is run.


jonpas:
That module really needs a rewrite.

PR #2140, opened 2015-08-13
AddAction basic param checking

PabstMirror:
From @alganthe 

doing `[cursorTarget, 0, ["ACE_TapShoulderRight"], XX] call ace_interact_menu_fnc_addActionToObject;` will cause hard lockup if `XX` is undefined.

Params is very fast and I think it's worth it as these can be publicly used functions.

### DIFF
--- a/addons/interact_menu/functions/fnc_addActionToClass.sqf
+++ b/addons/interact_menu/functions/fnc_addActionToClass.sqf
@@ -19,7 +19,9 @@
  */
 #include "script_component.hpp"
 
-params ["_objectType", "_typeNum", "_parentPath", "_action"];
+if (!params [["_objectType", "", [""]], ["_typeNum", 0, [0]], ["_parentPath", [], [[]]], ["_action", [], [[]], 11]]) exitWith {
+    ERROR("Bad Params");
+};
 
 // Ensure the config menu was compiled first
 if (_typeNum == 0) then {
--- a/addons/interact_menu/functions/fnc_addActionToObject.sqf
+++ b/addons/interact_menu/functions/fnc_addActionToObject.sqf
@@ -19,7 +19,9 @@
  */
 #include "script_component.hpp"
 
-params ["_object", "_typeNum", "_parentPath", "_action"];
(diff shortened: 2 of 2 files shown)


alganthe:
:+1: no more deadlocks around here

SilentSpike:
Wouldn't it make more sense to just use the params command without the condition aspect?
It has debugging built into it to see when a default param was used.

On this note, we should probably update all public functions to handle bad input

bux578:
> On this note, we should probably update all public functions to handle bad input

:+1:

jonpas:
> On this note, we should probably update all public functions to handle bad input

With a standard fail message preferably (coding guidelines for error messages :grin: ). :)

SilentSpike:
> With a standard fail message preferably

See:

> It has debugging built into it to see when a default param was used.

jonpas:
> See:

Oh right, even better, so that means no error messages would actually have to be printed from ACE's side as they would all be handled internally by `params`?

bux578:
> Oh right, even better, so that means no error messages would actually have to be printed from ACE's side as they would all be handled internally by params?

If you enable debugging for params. which probably most mission makers don't have enabled. So custom error messages make sense for API functions.

SilentSpike:
Yeah, I suppose it does mean you're relying on the people using it to be competent

jonpas:
Then I'd say #2052 could help out, one of the things I noticed in ACE3 all-around is very inconsistent practices for logging (the message and the code).

PabstMirror:
I've been playing with `allowFunctionsLog = 1;` and it adds a lot of good debug,
But there are plenty of cases where we our functions are designed to use default params, so I'm not sure seeing that in the rpt would stand out enough as an error.

The type checking isn't free, but it's very minor.

One thing we could do, is make a 

`func(public_createAction);, func(public_addActionToClass),,,,,`

which are just wrappers for our internal functions with added idiot proof checking and debug output, but that might be overkill?

jonpas:
> which are just wrappers for our internal functions with added idiot proof checking and debug output, but that might be overkill?

I'd say overkill, most if not all public functions are not run many times at all, mostly at mission start and some with script calls. Wrapper might be good if the performance difference is really noticable.

PR #2143, opened 2015-08-13
Remove PFEH hack

PabstMirror:
Don't need this anymore with current CBA
Ref:
https://dev.withsix.com/issues/76243
https://github.com/CBATeam/CBA_A3/commit/b430fd076253076a8bd5910bd8c71677d0d0aa77

also our code doesn't actually do anything as `ACE_time > 0` is always true even at breifing :smile:


Glowbal:
:+1:

Issue #1988, opened 2015-08-02
Explosions have no effect on hearing loss


voiperr:
ACE3 Version: 3.2.0

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
None

**Description:**
Explosions do not have an effect on hearing loss, at any distance from the player.

**Steps to reproduce:**
* Create a mission, place the player, make the player invincible, give the player grenades.
* Throw grenades all around you. You will receive ear ringing, but no hearing loss.

**Where did the issue occur?**
Everywhere

PR #2146, opened 2015-08-13
Code cleanup of Hearing module

jokoho48:
Code cleanup of Hearing module.

### DIFF
--- a/addons/hearing/functions/fnc_firedNear.sqf
+++ b/addons/hearing/functions/fnc_firedNear.sqf
@@ -38,9 +38,9 @@ _vehAttenuation = if ((ACE_player == (vehicle ACE_player)) || {isTurnedOut ACE_p
 _distance = 1 max _distance;
 
 _silencer = call {
-    if (primaryWeapon _firer) exitWith {(primaryWeaponItems _firer) select 0};
-    if (secondaryWeapon _firer) exitWith {(secondaryWeaponItems _firer) select 0};
-    if (handgunWeapon _firer) exitWith {(handgunItems _firer) select 0};
+    if (_weapon == primaryWeapon _firer) exitWith {(primaryWeaponItems _firer) select 0};
+    if (_weapon == secondaryWeapon _firer) exitWith {(secondaryWeaponItems _firer) select 0};
+    if (_weapon == handgunWeapon _firer) exitWith {(handgunItems _firer) select 0};
     ""
 };
 
@@ -68,7 +68,7 @@ if (count _weaponMagazines == 0) then {
 
 _magazine = "";
 {
-    params ["_magazineType", "_ammoType"];
+    _x params ["_magazineType", "_ammoType"];
     if (_ammoType == _ammo) exitWith {
         _magazine = _magazineType;
     };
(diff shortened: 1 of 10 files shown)


addons/hearing/functions/fnc_firedNear.sqf
```diff
@@ -37,11 +37,11 @@ _vehAttenuation = if ((ACE_player == (vehicle ACE_player)) || {isTurnedOut ACE_p
 
 _distance = 1 max _distance;
 
-_silencer = switch (_weapon) do {
-    case (primaryWeapon _firer) : {(primaryWeaponItems _firer) select 0};
-    case (secondaryWeapon _firer) : {(secondaryWeaponItems _firer) select 0};
-    case (handgunWeapon _firer) : {(handgunItems _firer) select 0};
-    default {""};
+_silencer = call {
+    if (primaryWeapon _firer) exitWith {(primaryWeaponItems _firer) select 0};
```
PabstMirror: This has a string inside of a if statement and no comparison to _weapon

addons/hearing/functions/fnc_firedNear.sqf
```diff
@@ -68,11 +68,11 @@ if (count _weaponMagazines == 0) then {
 
 _magazine = "";
 {
-    EXPLODE_2_PVT(_x,_magazineType,_ammoType);
+    params ["_magazineType", "_ammoType"];
```
PabstMirror: needs `_x` in front

jokoho48: oh right sry miss that

addons/hearing/functions/fnc_firedNear.sqf
jokoho48: oh yeah i miss that

PabstMirror:
I tested the call/switch statements, I'm not seeing much benifit in bis's benchmark:

```
_firer = player;
_weapon = (handgunWeapon player);
_silencer = switch (_weapon) do {
    case (primaryWeapon _firer) : {(primaryWeaponItems _firer) select 0};
    case (secondaryWeapon _firer) : {(secondaryWeaponItems _firer) select 0};
    case (handgunWeapon _firer) : {(handgunItems _firer) select 0};
    default {""};
};
0.0117188 ms
0.0116699 ms

_firer = player;
_weapon = (handgunWeapon player);
_silencer = call {
    if ((primaryWeapon _firer) == _weapon) exitWith {(primaryWeaponItems _firer) select 0};
    if ((secondaryWeapon _firer) == _weapon) exitWith {(secondaryWeaponItems _firer) select 0};
    if ((handgunWeapon _firer) == _weapon) exitWith {(handgunItems _firer) select 0};
    ""
};
0.012207 ms
0.0123047 ms
```

jokoho48:
yeah right see what you mean hmm ok i try it self

Glowbal:
In fact looking at those tests, it seems to be faster to stick with the `switch`. It also is easier to read/understand without having to take a closer look imo.

jokoho48:
but if you use a swtich True call if exitWith is faster

PabstMirror:
looks good, merging before looking at #1988 to avoid conflicts.

PR #2161, opened 2015-08-15
Don't broadcast localization for settings

PabstMirror:
smaller `publicVariable`
non-dedicated server will still localize

### DIFF
--- a/addons/common/functions/fnc_loadSettingsLocalizedText.sqf
+++ b/addons/common/functions/fnc_loadSettingsLocalizedText.sqf
@@ -15,17 +15,16 @@
 private ["_parseConfigForDisplayNames", "_name"];
 
 _parseConfigForDisplayNames = {
-    private "_optionEntry";
+    private ["_optionEntry", "_values", "_text"];
     _optionEntry = _this select 0;
     if !(isClass _optionEntry) exitwith {false};
+    _values = getArray (_optionEntry >> "values");
     _x set [3, getText (_optionEntry >> "displayName")];
     _x set [4, getText (_optionEntry >> "description")];
+    _x set [5, _values];
     _x set [8, getText (_optionEntry >> "category")];
 
-    private "_values";
-    _values = _x select 5;
     {
-        private "_text";
         _text = _x;
         if (((typeName _text) == "STRING") && {(count _text) > 1} && {(_text select [0,1]) == "$"}) then {
             _text = localize (_text select [1, ((count _text) - 1)]); //chop off the leading $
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_loadSettingsLocalizedText.sqf
```diff
@@ -15,17 +15,16 @@
...
     _x set [4, getText (_optionEntry >> "description")];
+    _x set [5, _values];
     _x set [8, getText (_optionEntry >> "category")];
 
-    private "_values";
-    _values = _x select 5;
     {
-        private "_text";
         _text = _x;
         if (((typeName _text) == "STRING") && {(count _text) > 1} && {(_text select [0,1]) == "$"}) then {
```
Glowbal: Because of the change on line 21;
```sqf
_values = getArray (_optionEntry >> "values");
```
We no longer require to be looping through the values and localize each value.

addons/common/functions/fnc_loadSettingsLocalizedText.sqf
```diff
@@ -42,7 +41,9 @@ _parseConfigForDisplayNames = {
 
     if !([configFile >> "ACE_Settings" >> _name] call _parseConfigForDisplayNames) then {
         if !([configFile >> "ACE_ServerSettings" >> _name] call _parseConfigForDisplayNames) then {
-            [missionConfigFile >> "ACE_Settings" >> _name] call _parseConfigForDisplayNames;
+            if !([missionConfigFile >> "ACE_Settings" >> _name] call _parseConfigForDisplayNames) then {
+                diag_log text format ["[ACE] - Setting found, but couldn't localize [%1] (server has but we don't?)", _name];
```
Glowbal: `text` is uncessary here. `diag_log format["content"]` works without it. Is there a reason is should be here?

MikeMatrix: If you do it without text, it'll have quotes around the entry in the RPT file.

addons/common/functions/fnc_loadSettingsLocalizedText.sqf
Glowbal: Never mind, I just saw the comment about string values in arrays from `setSettingFromConfig.sqf`:

> //Strings in the values array won't be localized from the config, so just do that now:

We should probably add this comment to here as well.

addons/common/functions/fnc_loadSettingsLocalizedText.sqf
Glowbal: I did not know that, thanks!

addons/common/functions/fnc_loadSettingsLocalizedText.sqf
PabstMirror: Strings in arrays are weird, the game doesn't localize them in the config, so I think we still have to do it manually

`getArray (configfile >> "ACE_Settings" >> "ace_interact_menu_textSize" >> "values")` = 
["$str_very_small","$str_small","$str_medium","$str_large","$str_very_large"]

PR #2163, opened 2015-08-15
Add objects to all curators

Glowbal:
Add objects from mission start and placed down by other curators to all curators.

### DIFF
--- a/addons/zeus/CfgEventHandlers.hpp
+++ b/addons/zeus/CfgEventHandlers.hpp
@@ -4,3 +4,11 @@ class Extended_PreInit_EventHandlers {
         init = QUOTE(call COMPILE_FILE(XEH_preInit));
     };
 };
+
+class Extended_InitPost_EventHandlers {
+    class AllVehicles {
+        class ADDON {
+            init = QUOTE(call FUNC(addObjectToCurator));
+        };
+    };
+};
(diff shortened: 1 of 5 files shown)


PabstMirror:
Looks useful.

It is fine now as an ACE_setting, but we need to be careful if we wanted to move this to the module.
Module settings are read after unit init.

Glowbal:
It's server side only, so `initPost` should be fine? Otherwise we could check if moduleInit has been completed and if not, collect the units and execute the init function for them on the `settingsInitalized` event?

addons/zeus/CfgEventHandlers.hpp
```diff
@@ -4,3 +4,11 @@ class Extended_PreInit_EventHandlers {
         init = QUOTE(call COMPILE_FILE(XEH_preInit));
     };
 };
+
+class Extended_InitPost_EventHandlers {
+    class AllVehicles {
+        class ADDON {
+            init = QUOTE(call FUNC(addObjectToCurator));
```
SilentSpike: This will error as the function uses `params` and no argument is passed (`params` doesn't handle a `nil` input unless it's within an array)

Glowbal: Using only a call will use the `_this` from the scope where you use the call. `QFUNC` is not an option since we need to call the function here.

SilentSpike: Aha, of course, I assumed params wouldn't like that.

Also forgot this is `string`, not `code`. Looks good :+1:

SilentSpike: Also, we can actually just make this serverInit since it only need to run server side.

Glowbal: Good suggestion. It's now changed :+1:

Issue #2138, opened 2015-08-13
You respawn with Laserdesignator in your hands if you had it in your inventory


rebelvg:
ACE3 Version: 3.2.1

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *Respawn Module, Save Gear Option*

**Description:**
*If you had a Laserdesignator in your inventory before getting killed you will respawn with Laserdesignator in your hands. This is a small issue but became sort of problem because we play without ace medical and we use built in arma's revive system. And the way it works is it actually respawns you on the same place in a different animation waiting for revive. So after a fast revive in a combat zone you end up holding Laserdesignator and aren't combat ready.
This happens because you have _unit selectWeapon "Laserdesignator"; in fnc_restoreGear.sqf on 162 line. Option to fix this is probably to check whether the unit has any primary, secondary or handgun weapon and to switch to it. Or to get a weapon unit held before dying and store it in that gear array and make a selectWeapon of the weapon unit held previously.*

**Steps to reproduce:**
* *Equip Laserdesignator. Press ESC and respawn. You will respawn with Laserdesignator in your hands.*

**Where did the issue occur?**
*Multiplayer and Singleplayer*

**RPT log file:**
*No need.*

bux578:
That's actually by design. The problem is, if you have a laser designator there's no other way to reload the batteries.

rebelvg:
Yes, I understand that. I mean, after the batteries have been reloaded you can make a switch to the weapon unit held before dying (save it in that gear array) or to any other weapon the unit has.

bux578:
I'll have a look. Do you have a repro mission with the revive system set up?

rebelvg:
https://mega.co.nz/#!btZWkI4K!mho5UdtZ5ZlZaFkBGvsLyRsg8h1935C9SL1g9rCKxVI

rebelvg:
And with the revive system. https://mega.co.nz/#!n15SSJBD!v5om4LIz67rfX4XBnopD7J7jqJFhtrLnldAdw2MeLdA

rebelvg:
Fist mission - just respawn test, second for the revive system. All must be tested in MP, to activate the revive system you must setdamage 1 to yourself, respawn button will just respawn you.

bux578:
:+1:

bux578:
@rebelvg could you and are you able to test this, please. It worked quite well when I tested it.

rebelvg:
[...]

PR #2165, opened 2015-08-15
after respawn select last weapon, muzzle and mode

bux578:
#2138

### DIFF
--- a/addons/common/functions/fnc_getAllGear.sqf
+++ b/addons/common/functions/fnc_getAllGear.sqf
@@ -18,6 +18,7 @@
  * 14-16: pistol (String, Array, Array)
  *    17: map, compass, watch, etc. (Array)
  *    18: binocluar (String)
+ *    19: active weapon, active muzzle, active weaponMode (Array)
  *
  */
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_getAllGear.sqf
```diff
@@ -47,5 +49,6 @@ if (isNull _unit) exitWith {[
     secondaryWeapon _unit, secondaryWeaponItems _unit, secondaryWeaponMagazine _unit,
     handgunWeapon _unit, handgunItems _unit, handgunMagazine _unit,
     assignedItems _unit,
-    binocular _unit
+    binocular _unit,
+    [currentWeapon _unit, currentMuzzle _unit, currentWeaponMode _unit]
```
jokoho48: currentThrowable?

bux578: Good idea.

bux578: After investigating this is very complicated and probably not worth.

bux578:
Ready for merge.

PR #2168, opened 2015-08-15
Code Cleanup laser Self Designate Module

jokoho48:
Code Cleanup laser Self Designate Module.

### DIFF
--- a/addons/laser_selfdesignate/functions/fnc_findLaserSource.sqf
+++ b/addons/laser_selfdesignate/functions/fnc_findLaserSource.sqf
@@ -6,7 +6,7 @@
  * 0: Emitter <OBJECT>
  *
  * Return Value:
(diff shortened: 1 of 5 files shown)


addons/laser_selfdesignate/functions/fnc_findLaserSource.sqf
```diff
@@ -3,28 +3,21 @@
...
- * 0: Emitter
- * 1: Owner
- * 
- * Return value:
- *    [position, direction]
+ * 0: Emitter <OBJECT>
+ *
+ * Return Value:
+ * 0: position <POSTION>
+ * 1: direction <ARRAY>
```
jonpas: Capitalization and `<ARRAY>` instead of `<POSTION>`.

PR #2169, opened 2015-08-15
Code Cleanup Laserpointer Module.

jokoho48:
Code Cleanup Laserpointer Module


addons/laserpointer/functions/fnc_drawLaserpoint.sqf
```diff
@@ -1,7 +1,24 @@
-// by commy2 and esteldunedain
+/*
+ * Author: commy2
```
Glowbal: Missed @esteldunedain as an author here

PR #2172, opened 2015-08-15
Code Cleanup Logistic Wirecutter Module

jokoho48:
Code Cleanup Logistic Wirecutter Module.


addons/logistics_wirecutter/XEH_preInit.sqf
```diff
@@ -10,3 +10,29 @@ PREP(interactEH);
 PREP(isFence);
 
 ADDON = true;
+
+
+// Condition for Interaction
+DFUNC(Condition) = {
```
Glowbal: These functions are all below the `ADDON` statement on line 12. That should not be the case, as this is signaling that this modules functions are all initialized. 

I'd suggest to instead move the `progressCheck` function into separate file and `PREP` it, and the `condition` and `statement` back to their original files.

jokoho48: i think its not good to move condition and statement in the original file because both are defined in a PFH and this on each Frame

Glowbal: They are 2 line functions that do not warrant their own global function. They belong with `interactEH`. Putting them in the `preInit` does not make sense to me; they would never be used anywhere else. Any potential performance gain you could get with it is mimimal at best.

I think that placing them in the `preInit` only makes the code structure harder to read in the long run.

PR #2176, opened 2015-08-16
remove asdg message

bux578:
new cba will include asdg jr


bux578:
Once the new cba gets released

jokoho48:
but than you must load everytime the ace_asdg_comp

jonpas:
> but than you must load everytime the ace_asdg_comp

New CBA will have Joint Rails integrated.

Glowbal:
:+1:

jokoho48:
i think that we must include the ace_asdg_comp in the main ACE

jonpas:
Why @jokoho48 ? What's currently there will be moved into main config since CBA is already a requirement anyways.

PR #2179, opened 2015-08-16
Fixed Atropine autoinjector display name

jonpas:


jokoho48:
oh that is a heavy Change so Heavy :D

PR #2182, opened 2015-08-16
Repair Cleanup

jonpas:


addons/repair/functions/fnc_isEngineer.sqf
```diff
@@ -1,13 +1,16 @@
 /*
  * Author: Glowbal, KoffeinFlummi, commy2
- * Check if a unit is any engineer class
+ * Check if a unit is any engineer class.
  *
  * Arguments:
- * 0: The Unit <OBJECT>
- * 1: Class <NUMBER> <OPTIONAL>
+ * 0: Unit <OBJECT>
+ * 1: Class <NUMBER> (Optional)
```
jokoho48: better (default: "Value")?

jonpas: Meh, doesn't really matter.

jonpas: Did it anyways. :banana:

MikeMatrix: Using `default` will imply that is optional, but also give you a reference to what the default value is.
Using `optional` only gives you information that it is not required, but not about what happens when you don't specify it.

jokoho48: and we can later use a tool for auto Wiki Creation :D

jonpas: Yep yep I get it, here have a :banana: each!

PR #2183, opened 2015-08-16
Fix treatment when weapon deployed

### LINKED ISSUE #2178: Character stuck when using medical actions.
ACE3 Version: 3.x.x
3.2.1

Mods:

@cba_a3
@ace3

Placed ACE3 Modules:
None

Description:
When using the vanilla weapon resting system, deploy a bipod on a cover object, if player tries to interact with themselves to perform medical actions, their character becomes stuck to the wall after the action is complete.

Steps to reproduce:

Deploy bipod on top of a cover object, like a sandbag.
Apply bandage to a limb.
Enjoy being stuck.

Where did the issue occur?
"Single player", Virtual Arsenal
"Multiplayer", Dedicated Server.

PabstMirror:
Fix #2178 - Un-deploys if bipod is out

Also fixes a minor problem when healing with a pistol when standing, it would switch from pistol back to rifle and then back to pistol.  Picks ending animation based on starting weapon.

### DIFF
--- a/addons/medical/functions/fnc_treatment.sqf
+++ b/addons/medical/functions/fnc_treatment.sqf
@@ -181,8 +181,12 @@ if (vehicle _caller == _caller && {_callerAnim != ""}) then {
         [_caller, "", 0] call EFUNC(common,doAnimation);
     };
     
-    if (stance _caller == "STAND") then {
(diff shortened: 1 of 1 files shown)


jonpas:
What about `weaponRested`, does that work as it should natively?

PabstMirror:
weaponRested seems to be passive, with no change in animationState

PR #2185, opened 2015-08-16
Fixed Release action blending with base interaction point

jonpas:
This moves `Release` interaction menu action inside `Interactions`, otherwise they were on top of each other, blending together (well technically `Release` was over `Interactions`).

With this you can now also drag/carry and load into vehicle directly without the need of releasing the object first (animation resets due to the dragging/carrying PFH).


jokoho48:
:+1: look ready for Merge

jonpas:
Replaced by #2283

PR #2187, opened 2015-08-16
Added Spare Parts module handling

jonpas:
**Requires #2202**

- Moved spare parts adding to vehicle init
- Added module option to add spare parts or not (to `Repair Settings` module)
- Added module to add specific spare part and amount of it to specific vehicle(s) (`Add Spare Parts` module)


Glowbal:
I don't tracked vehicles should have a track part as a spare. I'd rather add them to repair vehicles, implement some option to get either a new wheel or track part?

jonpas:
Hmh, some food for thought then.

gienkov:
That should be up to Mission Maker to add them.

bux578:
Hm. Here's a picture of a Leopard AS1. There are some spare tracks on the back
image

Some WW2 tanks had spare tracks welded to their hull.

I don't know how it's now with modern tanks.

jonpas:
Maybe it would be best to add both wheels and tracks (and maybe even toolkits) on init of the vehicle, that way it can be controlled via module settings (as some people might not want it to be there hardcoded).

jonpas:
Check PR's original comment.

jonpas:
Fixed conflicts.

jonpas:
Updated with final cargo event names, resolved conflicts and tested @Glowbal

addons/repair/functions/fnc_addSpareParts.sqf
```diff
@@ -0,0 +1,45 @@
...
+private ["_part"];
+params ["_vehicle", ["_amount", 1], ["_part", ""], ["_force", false]];
+TRACE_2("params",_vehicle,_amount);
+
+// Exit if ace_cargo is not loaded
+if !(["ace_cargo"] call EFUNC(common,isModLoaded)) exitWith {};
+
+// Collect until SettingsInitialized
+if (isNil QGVAR(settingInitted)) exitWith {
+    GVAR(addSparePartsCollection) pushBack _vehicle;
```
Glowbal: Should check if the vehicle is not already in that collection

addons/repair/XEH_postInit.sqf
```diff
@@ -5,3 +5,17 @@
...
+    ["SettingsInitialized", {
+        GVAR(settingInitted) = true; // Stop collecting in FUNC(addSpareParts)
+
+        // Exit if adding spare parts disabled
+        if (!GVAR(addSpareParts)) exitWith {GVAR(addSparePartsCollection) = nil};
+
+        // Add spare parts to vehicles in collection
+        {
+            [_x] call FUNC(addSpareParts);
+        } forEach GVAR(addSparePartsCollection);
```
Glowbal: Clean up `GVAR(addSparePartsCollection)` after we are done with it (`= nil`)

Khaosmatic:
Its very common for tracks to be carried spare. They are designed so they can be (while very difficult and time consuming) replaced in the field in whatever event they are damaged.

MrSanchez:
Placing the spare parts module in the mission (var name: spare), no extra spare parts are added besides the default wheel (for a Hunter) - however when executing [spare] call ace_repair_fnc_moduleAddSpareParts in the debug console, the requested spare parts are finally added.

Perhaps a delay issue?

jonpas:
> Placing the spare parts module in the mission (var name: spare), no extra spare parts are added besides the default wheel (for a Hunter) - however when executing [spare] call ace_repair_fnc_moduleAddSpareParts in the debug console, the requested spare parts are finally added.

You need to put in either vehicle names or synchronize them with the module.

> Perhaps a delay issue?

Until modules are initialized (meaning until `moduleAddSpareParts` finishes, this one just converts module options into ACE_Settings) `addSpareParts` function will only collect the information and add it when settings do initialize.

If all this still doesn't work as you expect it to please make a new issue following the guidelines.

jonpas:
@Glowbal `discussion` label can be removed. :)

PR #2191, opened 2015-08-17
Toolkit in misc box

SzwedzikPL:


jonpas:
Isn't misc box supposed to have ACE3 items only?

SzwedzikPL:
Toolbox is used by ace so i dont see any reason why toolkit shouldn't be in misc box :)

Glowbal:
It's really a vanilla item. I'm not certain it belongs in the Misc box from ACE.

SzwedzikPL:
> It's really a vanilla item. I'm not certain it belongs in the Misc box from ACE.

Ok, so nevermind.

PR #2192, opened 2015-08-17
Simpler cargo init

SzwedzikPL:


jonpas:
I don't think that's wise, all buildings and everything will get initted with this as well (assuming I understand how XEH actually works).

SilentSpike:
Yes, this actually is less simple :stuck_out_tongue:

SzwedzikPL:
> Yes, this actually is less simple
...?

SilentSpike:
Well instead of explicitly running the code where it's needed, it'll now be ran for everything

SzwedzikPL:
Its a lot of simpler for us and addon makers to create cargo objects. Every init function have check is this cargo obj so performence cost is minimal.

jonpas:
> Every init function have check is this cargo obj so performence cost is minimal.

Which means that has to get evaluated by XEH system, it is a performance impact, doing it on `All` adds a lot more objects than are in current classes, I'd approximate at least 3 times more, think of all little object, all buildings, everything.

PabstMirror:
`if (getNumber (configFile >> "CfgVehicles" >> _type >> QGVAR(hasCargo)) != 1) exitWith {};`

that's right near the top, so even if it ran on `ALL` objects, it would have very little cost

SilentSpike:
This change would only server to add the init to all the objects that don't need it though, so the difference is entirely redundant

PR #2195, opened 2015-08-17
Fixed View Distance limit not applied on Video Settings

jonpas:
`FUNC(returnValue)` was returning `-1` when `Video Settings` option was selected in client settings, that return value caused no change when `setViewDistance` was used as it compared wanted distance with limit. I am not sure, but it appears `-1 min 2000 == 2000` (BI issue?). I changed it to return the current view distance.

Also removed unnecessary EH `playerChanged` as `playerVehicleChanged` does what is needed and removed initial view distance set to limit as it is not needed, when you load in `playerVehicleChanged` gets run and sets it anyways.

This fixes issue described in https://github.com/acemod/ACE3/pull/2133#issuecomment-131280448

@Winter259 feel free to look it over in case I missed something (mostly things I removed in postInit).


Winter259:
Looks good from my end :+1:

addons/viewdistance/XEH_postInit.sqf
```diff
@@ -5,17 +5,7 @@ if (!hasInterface) exitWith {};
...
-    // Force the view distance down to the limit.
-    if (viewDistance > GVAR(limitViewDistance)) then {
-        setViewDistance GVAR(limitViewDistance);
-    };
-    
-    // Adapt view distance when the player is created or changed according to whether client is on foot or vehicle.
-    ["playerChanged",{
-        [false] call FUNC(adaptViewDistance);
-    }] call EFUNC(common,addEventHandler);
-
```
SilentSpike: I'm assuming these are both safe to remove as they'll be handled by the code that gets ran elsewhere?

jonpas: As stated in PR's comment:
> Also removed unnecessary EH `playerChanged` as `playerVehicleChanged` does what is needed and removed initial view distance set to limit as it is not needed, when you load in `playerVehicleChanged` gets run and sets it anyways.

SilentSpike: I probably should have read the PR :+1:

And playerVehicleChanged is guaranteed to run after settings initialized?

jonpas: That I can't answer for sure, in editor it appears to be so.

SilentSpike: explicit > implicit :wink:

jonpas: Agreed, I don't want it to run twice if it is guaranteed though, maybe @esteldunedain knows for sure.

SilentSpike: If you do end up preserving it, it'd probably make more sense for it to be `[false] call FUNC(adaptViewDistance)` anyway :+1: Otherwise all looks good

jonpas: As it is you mean :grin: 

I'll try to hunt down @esteldunedain , if he won't answer in a few days (or someone else tests it) I'll readd it.

SilentSpike: I mean to say instead of lines 10-12 a simple call of the function will handle it

jonpas: Ah I see, I guess it wouldn't hurt if it got called twice on start either way.

jonpas: Added.

PR #2197, opened 2015-08-18
Map open close events / MapTools Cleanup

PabstMirror:
Switches a PFEH to be event driven and only run when the map is open.
 (really not that big a deal performance wise, but it is 3 if statements each frame we don't need)

Adds 2 local events `mapOpened` / `mapClosed` (I thought BFT might have needed these for something???)

Also cleanup / params of map tools


addons/common/XEH_postInit.sqf
```diff
@@ -371,5 +371,23 @@ GVAR(deviceKeyCurrentIndex) = -1;
 {false},
 [0xC7, [true, false, false]], false] call cba_fnc_addKeybind;  //SHIFT + Home Key
 
+//Map opened/closed Events:
+GVAR(mapOpened) = false;
+[] spawn {
+    waitUntil {(!isNull findDisplay 12)};
```
jonpas: Is there no way to handle this with PFH wait or something? Not the first time I see it being used for map https://github.com/acemod/ACE3/blob/275b0c43000709282dafd5af89c949a12d8996be/addons/map/XEH_postInitClient.sqf#L15

PabstMirror:
```
// This spawn is probably worth keeping, as pfh don't work natively on the briefing screen and IDK how reliable the hack we implemented for them is.
// The thread dies as soon as the mission start, so it's not really compiting for scheduler space.
```

Since that was written, PFEH now work natively on the briefing screen.  So we could convert all these spawns to PFEH.

But like it also says, these are "one and done", they will be gone from the scheduler once the game starts, so it really doesn't matter.

jonpas: I see, the wiki does say waitUntils and spawns are banned though.

voiperr:
My map flashlights PR creates a similar event for map opening/closing, but adds it to the main PFH that checks local events: https://github.com/voiperr/ACE3/blob/ace_map_dev/addons/common/XEH_postInit.sqf#L286. Wouldn't this be more efficient than adding another PFH to the mix?

PabstMirror:
The differences are very minor.

Yours has 2 lines of code each frame (which is nothing to worry about).
This has zero code most of the time and then creates a PFEH that runs only while the map is opened and exits when the map is closed.

nsgundy:
BFT does in deed require these events. The current temp implementation is the same as what @voiperr used above [(see here for code in BFT branch)](https://github.com/acemod/ACE3/blob/BFT/addons/bft_drawing/XEH_clientInit.sqf#L9-L23), but I don't see why we wouldn't use @PabstMirror 's implementation as it does seem slicker.

MikeMatrix:
I agree, it seems like an event that could be used throughout the codebase. Why not make it a common event?

PabstMirror:
@voiperr - in testing `visableMap` is also true on the UAV controller map, so I'm going to use your events as any speed difference is extremely minor

PR #2198, opened 2015-08-18
Added Logistics Categories

jonpas:
- `ACE Logistics` group for mission modules
- `Logistics` category for ACE Options
- `ACE Logistics` vehicle class (`ACE_Logistics_Items`)

### DIFF
--- a/addons/cargo/ACE_Settings.hpp
+++ b/addons/cargo/ACE_Settings.hpp
@@ -4,6 +4,6 @@ class ACE_Settings {
         description = CSTRING(ModuleSettings_enable_Description);
         typeName = "BOOL";
         value = 1;
-        category = CSTRING(settingsCategory);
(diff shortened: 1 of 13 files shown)


Glowbal:
I do not see the point of adding an empty module just for the `CfgFactionClasses`. I'd say put this next to the other `CfgFactionClasses` entry in I think it's the main module.

jonpas:
Can do that as well, I was thinking that modules like towing, refueling and rearming share a lot of the attach code, but now that I think about it more, it would be better to expand attach functionality in common module, talking with GitHawk we both agree there is copy pasted code in the mentioned modules and ACE3 could benefit from an attach framework and public API.

jonpas:
@Glowbal There, added it to main, including `CfgVehicleClasses` entry `ACE_Logistics_Items`.

PR #2201, opened 2015-08-18
Add vehicle towing and tilting

GitHawk:
TODO
- [ ] Add MP compatibility for towing
- [ ] Test network performance

DONE
- [x] Fix towing on bridges


GitHawk:
[Video](https://www.youtube.com/watch?v=wXiGjLAxqHM)

I could use some people testing and giving feedback.

esteldunedain:
Unfortunately setting position over the network on each frame should cause a huge amount of traffic in MP. Have you done any sort of test about this?

Doing it at larger intervals could help, but will clearly look choppier unless velocity is set too.

GitHawk:
Unfortunately, due to ground friction `setVelocity `is completely useless.

SzwedzikPL:
If i remember right in ACE2 `attachTo` was used in similar function.

https://www.youtube.com/watch?v=0wcWqb3qQPY&t=5m

commy2:
>Isn't setting position on each frame not the same thing ArmA does, when you drive a vehicle?
You should probably test the traffic on a server.( with `#monitor 1` iirc)

>attachTo
That won't pass our quality standards.

esteldunedain:
> That won't pass our quality standards.

Why?

GitHawk:
I was discussing a dynamic `attachTo` solution last night with esteldunedain.
ArmA 2 was static.
I could make a test build which modifies the attach position every (few) frames and so simulate a towing behavior, while trying to avoid the issues displayed in the video SzwedzikPL posted.

SilentSpike:
This is another feature like fastroping where I just don't see any implementation which is good enough. Using a basic attachTo is the most usable, efficient and simple solution, whether or not we care to have that in ace is another question.

commy2:
>Why?

I remember discussions about it. No idea with whom and when they took place. Might have been in the AGM peroid ... so years ago. 
Maybe I'm wrong about todays consensus.

TheConen:
Something similar has been released today, maybe it's worth contacting the author and see how he has done it? http://www.armaholic.com/page.php?id=30575

GitHawk:
Looks interesting

GitHawk:
I have stopped working on this branch since there was no clear commitment of the ace team to this topic.

jonpas:
I think the rest of the team will agree with me. Tilting is pretty nice and works well, but towing implementation is not up tk standards we want I am afraid, huge usage of setting position and velocity, network impact would be huge.

jonpas:
We have decided to use tilting from this, I will take it out in a separate PR. Towing would have to be improved with multiplayer compatibility for starters.

commy2:
- [ ] CBA events

jonpas:
Replaced by #3898 

And as said above, towing implementation is not up to ACE3 standards, lots of network usage and performance impact.

PR #2202, opened 2015-08-18
Cargo Events

jonpas:
Replaced `cargo_hideUnit` event with already existing `hideObjectGlobal` from common.

Added new events to cargo:

Type | Event Name | Passed parameter(s) `_this` | Locality
------ | --------------- | ------------------------------------ | ---------
Listenable | "cargoLoaded" | [_item, _vehicle] | Global
Listenable | "cargoUnloaded" | [_item, _vehicle] | Global
Listenable | "cargoAddedByClass" | [_item, _vehicle, _amount] | Global
Callable | "LoadObject" | [_item, _vehicle] | Local
Callable | "UnloadObject" | [_item, _vehicle] | Local
Callable | "AddCargoByClass" | [_item, _vehicle, _amount (default: 1)] | Local

### DIFF
--- a/addons/cargo/XEH_postInit.sqf
+++ b/addons/cargo/XEH_postInit.sqf
@@ -2,3 +2,4 @@
 
 ["LoadItem", {_this call FUNC(loadItem)}] call EFUNC(common,addEventHandler);
 ["UnloadItem", {_this call FUNC(unloadItem)}] call EFUNC(common,addEventHandler);
+["AddCargoItem", {_this call FUNC(addCargoItem)}] call EFUNC(common,addEventHandler);
--- a/addons/cargo/XEH_postInit.sqf
+++ b/addons/cargo/XEH_postInit.sqf
@@ -1,5 +1,5 @@
 #include "script_component.hpp"
 
-["LoadItem", {_this call FUNC(loadItem)}] call EFUNC(common,addEventHandler);
-["UnloadItem", {_this call FUNC(unloadItem)}] call EFUNC(common,addEventHandler);
-["AddCargoItem", {_this call FUNC(addCargoItem)}] call EFUNC(common,addEventHandler);
+["LoadCargo", {_this call FUNC(loadItem)}] call EFUNC(common,addEventHandler);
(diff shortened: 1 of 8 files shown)


PabstMirror:
Looks good to me. 
What about adding `ace_player` to the global events?

jonpas:
> What about adding ace_player to the global events?

It can also be called from server on vehicle init, pretty sure `ace_player == objNull` there, not sure how you guys want to handle this, if it is really `objNull` it can be added for sure and then it's up to whoever listens to the event to take care of that?

jonpas:
- Added callable event `AddCargoItem` and it's function, this takes a classname rather than object, spawns it and puts it in cargo (part of what `initVehicle` was doing), it can take optional parameter for amount of items
- Added listenable event `cargo_itemAdded` to accompany above event

This is now a requirement for #2187 (due to events for communication).

addons/cargo/XEH_postInit.sqf
```diff
@@ -0,0 +1,5 @@
+#include "script_component.hpp"
+
+["LoadItem", {_this call FUNC(loadItem)}] call EFUNC(common,addEventHandler);
+["UnloadItem", {_this call FUNC(unloadItem)}] call EFUNC(common,addEventHandler);
+["AddCargoItem", {_this call FUNC(addCargoItem)}] call EFUNC(common,addEventHandler);
```
Glowbal: I am not certain about the name of this event. Perhaps: `AddCargoItemByType` ?

jonpas: Hmm, maybe better would be? 
- `LoadObject`
- `UnloadObject`
- `AddCargoItem`/`AddCargoItemByType`

Up to you, I am bad with names.

jokoho48:
:+1:

addons/cargo/XEH_postInit.sqf
Glowbal: Yes, that would work for me.

jonpas:
@Glowbal check PR's original comment, I renamed them in the way I think makes more sense.

PR #2203, opened 2015-08-18
Replaced check for cargo function by check for ace_cargo in Repair

jonpas:


PabstMirror:
Why don't we just add the event handler in cargo's CfgEventHandlers.hpp?

jonpas:
Then you have to check if `ace_repair` is loaded since that's where the repair items are. I think it's better this way, either way you have to check if one another exists.

PabstMirror:
It should be fine to add entries to `Extended_Init_EventHandlers` for vehicles even if they don't exist.

CBA doesn't iterate though all XEH.
CBA just goes through the inheritance chain for XEH enabled vehicles.

So if `ACE_RepairItem_Base` doesn't exist, the XEH entry won't do anything.

jonpas:
Confirmed working, I've moved it over to cargo. Thanks for the insight. :)

PR #2216, opened 2015-08-20
Prevent CBA keybindings in zeus interface

### LINKED ISSUE #1951: Zeus interface climb
ACE3 Version: 3.2.0

**Mods:**
@cba_a3
@ace3

**Placed ACE3 Modules:**
Non all set via Mission variables.

**Description:**
If you have Climb on CTRL+V and you are in the forced Zeus interface triing to paste something it will tell you that you can't climb here

**Steps to reproduce:**
* Bind Climb to CTRL+V
* Press CTRL+V while in Zeus interface

"Multiplayer"

**RPT log file:**
No script errors

SilentSpike:
Checks if the key pressed matches the zeus menu keybind and only fires the code to handle CBA keybinds if true. Should fix the majority of weirdness with keybinds in zeus. However, not perfect as shared keybinds with different modifiers will still fire and it's really an un-ideal workaround.

Fixes #1951


SilentSpike:
Modifier keys now taken into account, tested and fixed :smile: 

Also, confirming that CBA_fnc_getKeybind does return the *current* key rather than the default one :+1:

PR #2234, opened 2015-08-22
Repeat CPR until success

gienkov:
This will basically run CPR action again as you would select it from interaction menu. It will be ran if CPR statement chance fails.
I did this because in RL you would repeat CPR cycles over and over until success, right now you have to select it again and it may be cumbersome for some.

Failed once:
https://youtu.be/9k70o4F8amc

### DIFF
--- a/addons/medical/functions/fnc_treatmentAdvanced_CPRLocal.sqf
+++ b/addons/medical/functions/fnc_treatmentAdvanced_CPRLocal.sqf
@@ -29,6 +29,10 @@ if ((random 1) >= 0.6) then {
     _target setvariable [QGVAR(inCardiacArrest), nil,true];
     _target setvariable [QGVAR(heartRate), 40];
     _target setvariable [QGVAR(bloodPressure), [50,70]];
+}
+else
+{
+    [ARR_4(_caller, _target, 'body', 'CPR')] call DFUNC(treatment);
 };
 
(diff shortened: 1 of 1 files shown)


addons/medical/functions/fnc_treatmentAdvanced_CPRLocal.sqf
```diff
@@ -29,6 +29,10 @@ if ((random 1) >= 0.6) then {
     _target setvariable [QGVAR(inCardiacArrest), nil,true];
     _target setvariable [QGVAR(heartRate), 40];
     _target setvariable [QGVAR(bloodPressure), [50,70]];
+}
+else
+{
```
jonpas: ACE3 standard is `} else {` in the same line.

jokoho48:
i think its better to add this as a Clientside Setting

jonpas:
> i think its better to add this as a Clientside Setting

:+1:

gienkov:
It doesn't work with medical menu "Open after treatment" option set to on. Dunno how to overcome this and how to create clientside setting for it.

SilentSpike:
I don't like this idea, how do you know it has failed unless you check the pulse?

gienkov:
Patient wakes up if you succeed.

SilentSpike:
CPR doesn't wake the patient up, it restarts their heart

gienkov:
That's what happens in game.

Glowbal:
Besides me being unsure if this is desirable behaviour, this will not work properly in MP because you start on the locality of the medic and end in the locality of the patient.

Issue #2228, opened 2015-08-22
No Self Attach Equipment Option


DevoutingChain:
**ACE3 Version:** 3.2.1
**CBA Version:** v2.0.0.150817

**Mods:**
   - @cba_a3
   -  @ace3

**Placed Modules:** N/A

**Description:**

I Open the self interaction menu, go to equipment and then there is no option to attach equipment. I can still attach equipment like an IR/Strobe to vehicles with no problem though. 

**Steps to reproduce:**

1.Load in game
2.Open self interaction menu 
3.Open Equipment option

**Where did the issue occur?**
"Multiplayer", "Singleplayer"

**RPT log file:** N/A

finesseseth:
Can confirm im getting this as well, the latest CBA update seems to cause it. If I run the old RC7 v1.1.21.150527 of CBA it works as normal.

jonpas:
Does the main "Attach" action show (just no items to select) or none at all?

finesseseth:
Doesn't show the attach menu item at all.

PabstMirror:
For `[{}, 0, ""] call CBA_fnc_defaultParam`
Old CBA: `""`
New CBA: `{}`

`{}` doesn't make any sense, so not really an error with CBA

Already fixed on master, plus I'll look into fixing a partial cause in interact menu,

PR #2236, opened 2015-08-23
Compile Self Menu, set "Default Params" to array instead of code

PabstMirror:
This is for the `6: Action parameters <ANY> (Optional)`

Compile self uses `{}`
compile external uses `[]` which I think makes more sense

ref #2228


SilentSpike:
:+1: Makes sense to me (might just be typo to be honest)

PR #2237, opened 2015-08-23
Make reserve steerable

gienkov:
Source: http://www.globalsecurity.org/military/systems/aircraft/systems/mc-5.htm

> The system's **identical main and reserve parachutes**, which can be interchanged, are 370 square feet, seven-celled, and are manufactured from 1.1 ounce F-111 nylon ripstop fabric.

MC-5 was steerable in ACE2 and I think we are simulating it here too right? There are non-steerable and steerable reserve parachutes but MC-5 has two identical chutes.
Also this fix makes so you don't magically stop while freefalling as it is with NonSteerable_Parachute_F.
I was able to open it 20 meters above ground cause all my falling speed is gone as soon as I click open parachute. Thats BS.
This fixes that.

Also MC-5 in RL, very similiar to A3 one right?
img

### DIFF
--- a/addons/parachute/CfgVehicles.hpp
+++ b/addons/parachute/CfgVehicles.hpp
@@ -94,7 +94,7 @@ class CfgVehicles {
         displayName = CSTRING(ReserveParachute);
         scope = 1;
         mass = 70;
-        ParachuteClass = "NonSteerable_Parachute_F";
+        ParachuteClass = "Steerable_Parachute_F";
         ace_reserveParachute = "";
         ace_hasReserveParachute = 0;
     };


sargken:
+1

jokoho48:
i think it was better if we have 2 parachutes one that have a steerable parachute and one with a nonsteerable parachute

jokoho48:
think about ACE_ReserveParachuteSteerable and ACE_ReserveParachuteNonSteerable and than ACE_Parachute_N and ACE_Parachute_S?

Dobakjang:
NAVER - http://www.naver.com/
--------------------------------------------

kgino1045@naver.com 님께 보내신 메일 <Re: [ACE3] Make reserve steerable (#2237)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

gienkov:
Can we get this into 3.4.0 ?

jokoho48:
i think not because i dont like the idea of making the reseve parachute steerable

bux578:
The reserve parachutes i know are not steerable. Double checked that with an airborne soldier.

gienkov:
So we're left with possibility to open it even at extreme low altitudes,
since opening reserve make you stop in place :(

jonpas:
> since opening reserve make you stop in place :(

That sounds like an issue, and separate from steerable feature.

gienkov:
Might be the engine?

bux578:
> since opening reserve make you stop in place :(
Never had that problem. Will try to repro.

headedhorseman101:
@jokoho48 Adding a steerable reserve parachute opinion addition to the static line reserve would be realistic because the arma 3 parachutes are modelled after the MC-5, which is a dual purpose (can be configured for static line [non-steerable] and free fall [steerable] drops). This reconfigurability applies for both the main and reserve chute. Sources [here] (http://soldiersystems.net/2011/09/20/us-replacing-mc-4-free-fall-parachute/) and [here]

bux578:
Combinations are:
* Steerable / Steerable
* Non-Steerable / Non-Steerable

PR #2243, opened 2015-08-23
Improved Repair Hitpoint String Localization

jonpas:
**Overview:**
Added better string(table) framework to `addRepairActions`, it will attach a number if a part has the same name without the need to have multiple strings in the stringtable with same translations (mostly notable on the Taru which has 20 `Glass_N_hitpoint` hitpoints). There is most likely room for improvement, I am not too experienced with string manipulation (and SQF isn't good for it either), it shouldn't take much performance at all though, for one it happens when the vehicle of that class is newly spawned, secondly there aren't many parts on vehicles.

**Additionally:**
- Added missing hitpoints to stringtable
- Removed same type hitpoints from stringtable (eg. `HitEngine1`, `HitEngine2`)

**To-do:**
- [x] Add support for display messages (still uses `format ["STR_ACE_Repair_%1", _hitPoint];`)
- [x] Consolidate into 1 function


jonpas:
This is ready for review.

Glowbal:
This now has a merge conflict.

jonpas:
I anticipated that, too much work on one file in multiple PRs.

jonpas:
Resolved and tested @Glowbal

Issue #2238, opened 2015-08-23
Improving wirecutter modularity


alganthe:
This is not an issue but an improvement needed for ace_logistics_wirecutter

atm the only way for a mapmaker, object-maker to add his fence to the fence array is to modify `ace_logistics_wirecutter_fnc_isFence` which is not very user friendly.

A better way to do it would be to replace the defines in `ace_logistics_wirecutter_fnc_isFence` and replace them by a config check with a new function in common and proper config params.

This way the fence array would be easier to maintain and native support would be possible.

SzwedzikPL:
:+1:

jonpas:
Config for sure :+1:

Dobakjang:
NAVER - http://www.naver.com/
--------------------------------------------

kgino1045@naver.com 님께 보내신 메일 <Re: [ACE3] Improving wirecutter modularity (#2238)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

PabstMirror:
For map/object makers:
```
class your_Fence: Wall_F { 
    ace_logistics_wirecutter_isFence = 1; 
};
```

PR #2244, opened 2015-08-23
wirecutters config for fences + cleanup

PabstMirror:
#2238 config entry for fences
dropped mass of item slightly
cleanup (I missed that #2172 already exists :rage2:)


bux578:
:+1:

alganthe:
:+1:

PR #2245, opened 2015-08-23
Custom Hitpoint Repair Positions Framework

jonpas:
**Overview:**
Adds a framework for custom hitpoint positions/selection names to be used as position to render the repair action of the specified hitpoint. Most default selections (except Winch) have position of `[0,0,0]` which obviously doesn't work, the check if it's on that location was added to prevent being over the base interaction menu (if it is on zero location, it will be grouped into the base interaction menu as before).

**Config entries:**
`GVAR(hitpointPositions[]) = { {"HitPointName", {position}} };`
Example:
`GVAR(hitpointPositions[]) = { {"HitEngine", {0, 1, -1}}, {"HitFuel", {0, 0, -0.5}} };`
(It is impossible to use array in a config array, so I decided to parse the string into an array.)

Instead of a position it can also be a selection name:
`GVAR(hitpointPositions[]) = { {"HitEngine", "engineSelection"} };`

**Side-note:**
~~`IGNORED_HITPOINTS` was also removed, since all it did was prevent glasses from being repaired (and it can be freaking nasty to drive/fly with a broken window).~~ I will be adding framework for grouping hitpoints (so you fix another hitpoint when you fix one, but doesn't show in the menu) in another PR.

**API Suggestion:**
I suggest a check of `[0,0,0]` position is made in public API for adding the actions to interaction menu, the only thing it does is cover the base interaction point.

### DIFF
--- a/addons/repair/CfgVehicles.hpp
+++ b/addons/repair/CfgVehicles.hpp
@@ -334,4 +334,9 @@ class CfgVehicles {
         GVAR(canRepair) = 1;
         transportRepair = 0;
     };
+
+    class Quadbike_01_base_F;
+    class B_Quadbike_01_F: Quadbike_01_base_F {
+        GVAR(hitpointPositions[]) = {"HitEngine", "0, 0.5, -0.7", "HitFuel", "0, 0, -0.5"};
+    };
 };
--- a/addons/repair/CfgVehicles.hpp
+++ b/addons/repair/CfgVehicles.hpp
@@ -337,6 +337,6 @@ class CfgVehicles {
 
     class Quadbike_01_base_F;
     class B_Quadbike_01_F: Quadbike_01_base_F {
-        GVAR(hitpointPositions[]) = {"HitEngine", "0, 0.5, -0.7", "HitFuel", "0, 0, -0.5"};
+        GVAR(hitpointPositions[]) = { {"HitEngine", {0, 0.5, -0.7}}, {"HitFuel", {0, 0, -0.5}} };
     };
 };
--- a/addons/repair/functions/fnc_addRepairActions.sqf
+++ b/addons/repair/functions/fnc_addRepairActions.sqf
@@ -79,7 +79,7 @@ if (_type in _initializedClasses) exitWith {};
 
         // add misc repair action
 
-        private ["_name", "_text", "_icon", "_selection", "_customSelectionsCfg", "_customSelectionsIndex", "_customSelection", "_condition", "_statement"];
+        private ["_name", "_text", "_icon", "_selection", "_customSelectionsConfig", "_currentHitpoint", "_condition", "_statement"];
 
         _name = format ["Repair_%1", _x];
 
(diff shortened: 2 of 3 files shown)


addons/repair/CfgVehicles.hpp
```diff
@@ -334,4 +334,9 @@ class CfgVehicles {
         GVAR(canRepair) = 1;
         transportRepair = 0;
     };
+
+    class Quadbike_01_base_F;
+    class B_Quadbike_01_F: Quadbike_01_base_F {
+        GVAR(hitpointPositions[]) = {"HitEngine", "0, 0.5, -0.7", "HitFuel", "0, 0, -0.5"};
```
Glowbal: I dislike this format. I'd prefer a 2d array for this instead. 

`{ {"HitEngine", {0, 0.5, -0.7} }, {"HitFuel", {0, 0, -0.5} } };`

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -96,7 +89,28 @@ _wheelHitPointSelections = _wheelHitPointsWithSelections select 1;
         };
 
         _icon = "A3\ui_f\data\igui\cfg\actions\repair_ca.paa";
+
         _selection = "";
+        // Get custom position if available
+        _customSelectionsCfg = configFile >> "CfgVehicles" >> _type >> QGVAR(hitpointPositions);
+        if (isArray _customSelectionsCfg) then {
```
Glowbal: This code block is hard to read. Please write more comments saying what's going on. Though I suspect that this will change with a different format in the config.

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -112,7 +126,11 @@ _wheelHitPointSelections = _wheelHitPointsWithSelections select 1;
         } else {
             private "_action";
             _action = [_name, _text, _icon, _statement, _condition, {}, [_x, "MiscRepair"], _selection, 4] call EFUNC(interact_menu,createAction);
-            [_type, 0, ["ACE_MainActions", QGVAR(Repair)], _action] call EFUNC(interact_menu,addActionToClass);
+            [   _type,
+                0,
+                if (_selection isEqualTo [0, 0 ,0]) then { ["ACE_MainActions", QGVAR(Repair)] } else { [] }, // Put inside main actions if no other position was found above
```
Glowbal: It is probably cleaner if you put this on it's own line above the `addActionToClass` call.

addons/repair/CfgVehicles.hpp
jonpas: Ooh, you can use {} instead of [] to havd multidimensional arrays in config as well, that's what I was missing, thanks!

addons/repair/functions/fnc_addRepairActions.sqf
jonpas: Agreed, would be most clean if that check would be made in `addActionToClass` and `addActionToObject` functions though.

jonpas:
All done @Glowbal I also reverted the `IGNORED_HITPOINTS`, will do that in the hitpoint groups PR as will be decided.

PR #2246, opened 2015-08-24
Repair Hitpoint Groups Framework

jonpas:
**Overview:**
Adds a framework to allow grouping of multiple hitpoints into one by config. This means only the first hitpoint in the config array gets the repair action, all the rest in that group get repaired simultaneously when the main one gets repaired. They repair to full even if the main repair is only partial, this should have no issues though.

**Config entries:**
`GVAR(hitpointGroups[]) = { {"hitpointName1", {"hitpointName2"}} };`
Example:
`GVAR(hitpointGroups[]) = { {"HitGlass1", {"HitGlass2"}} };`
(`HitGlass1` is the one which gets the action here and `HitGlass2` is grouped under it.)

**To-do:**
- [x] Add support when one of the grouped hitpoints is damaged but not the main
- [x] Add support for multiple groups on one vehicle
- [x] Fix check group damage returning true even if current hitpoint not in a group
- [x] Decide on `IGNORED_HITPOINTS` -> config framework in separate PR

### DIFF
--- a/addons/repair/functions/fnc_canMiscRepair.sqf
+++ b/addons/repair/functions/fnc_canMiscRepair.sqf
@@ -21,7 +21,7 @@ private ["_hitpointGroupConfig", "_hitpointGroup", "_postRepairDamage", "_return
 params ["_caller", "_target", "_hitPoint"];
 
 // Get hitpoint groups if available
-_hitpointGroupConfig = configFile >> "CfgVehicles" >> typeOf _target >> QGVAR(hitpointGroup);
+_hitpointGroupConfig = configFile >> "CfgVehicles" >> typeOf _target >> QGVAR(hitpointGroups);
 _hitpointGroup = [];
 if (isArray _hitpointGroupConfig) then {
     // Loop through hitpoint groups
@@ -46,5 +46,9 @@ _return = false;
         _return = true;
     };
 } forEach _hitpointGroup;
+
+if (typeOf _target == "B_MRAP_01_F") then {
+    diag_log format ["%1 - %2", _hitPoint, _hitpointGroup];
+};
 
 _return
--- a/addons/repair/functions/fnc_canMiscRepair.sqf
+++ b/addons/repair/functions/fnc_canMiscRepair.sqf
@@ -1,0 +1,40 @@
+/*
+ * Author: Jonpas
+ * Check if misc repair action can be done, called from callbackSuccess.
+ *
+ * Arguments:
+ * 0: Unit that does the repairing <OBJECT>
+ * 1: Vehicle to repair <OBJECT>
+ * 2: Selected hitpoint <STRING>
+ *
+ * Return Value:
+ * Can Misc Repair <BOOL>
+ *
+ * Example:
+ * [unit, vehicle, "hitpoint", "classname"] call ace_repair_fnc_canMiscRepair
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private ["_hitpointGroup", "_postRepairDamage", "_return"];
+params ["_caller", "_target", "_hitPoint"];
+
+// Check hitpoint group
+_hitpointGroup = configFile >> "CfgVehicles" >> typeOf _target >> QGVAR(hitpointGroup);
+_hitpointGroup = if (isArray _hitpointGroup) then {getArray _hitpointGroup} else {[]};
+
+if !(_hitPoint in _hitpointGroup) then {
+    _hitpointGroup pushBack _hitPoint;
+};
+
+_postRepairDamage = [_caller] call FUNC(getPostRepairDamage);
+
+_return = false;
+{
+    if ((_target getHitPointDamage _x) > _postRepairDamage) exitWith {
+        _return = true;
+    };
+} forEach _hitpointGroup;
+
+_return
--- a/addons/repair/functions/fnc_canMiscRepair.sqf
+++ b/addons/repair/functions/fnc_canMiscRepair.sqf
@@ -17,19 +17,29 @@
  */
 #include "script_component.hpp"
 
-private ["_hitpointGroup", "_postRepairDamage", "_return"];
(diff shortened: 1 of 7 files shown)


Glowbal:
- Leaving out the glass was done on purpose. It's something I considered only being repeared at an repair facility / workshop. If we do want to be able to repair glass, I'd at least go for a separate action.
- The hitpointGroup only works for a single action; I'd expand this so we can make us of it for any potential other hitpoints besides just glass.

addons/repair/functions/fnc_canMiscRepair.sqf
```diff
@@ -0,0 +1,40 @@
...
+
+private ["_hitpointGroup", "_postRepairDamage", "_return"];
+params ["_caller", "_target", "_hitPoint"];
+
+// Check hitpoint group
+_hitpointGroup = configFile >> "CfgVehicles" >> typeOf _target >> QGVAR(hitpointGroup);
+_hitpointGroup = if (isArray _hitpointGroup) then {getArray _hitpointGroup} else {[]};
+
+if !(_hitPoint in _hitpointGroup) then {
+    _hitpointGroup pushBack _hitPoint;
```
Glowbal: Wouldn't this end up collecting the hitpoint group + the glass? So if we are checking the engine it will check if the engine is damaged or if the glass is damaged. If only the glass is damaged, we will get the action to repair the engine.

addons/repair/functions/fnc_canMiscRepair.sqf
```diff
@@ -0,0 +1,40 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private ["_hitpointGroup", "_postRepairDamage", "_return"];
+params ["_caller", "_target", "_hitPoint"];
+
+// Check hitpoint group
+_hitpointGroup = configFile >> "CfgVehicles" >> typeOf _target >> QGVAR(hitpointGroup);
+_hitpointGroup = if (isArray _hitpointGroup) then {getArray _hitpointGroup} else {[]};
```
Glowbal: Using the same variable name for the config and the array does not improve readability.

addons/repair/functions/fnc_doRepair.sqf
```diff
@@ -31,9 +31,23 @@ _hitPointDamage = _hitPointDamage max ([_unit] call FUNC(getPostRepairDamage));
 // raise event to set the new hitpoint damage
 ["setVehicleHitPointDamage", _vehicle, [_vehicle, _hitPoint, _hitPointDamage]] call EFUNC(common,targetEvent);
 
+// Repair the rest in the group (no specific damage, main hitpoint is enough)
+_hitpointGroup = configFile >> "CfgVehicles" >> typeOf _vehicle >> QGVAR(hitpointGroup);
+_hitpointGroup = if (isArray _hitpointGroup) then {getArray _hitpointGroup} else {[]};
+if (count _hitpointGroup > 0) then {
```
Glowbal: This is also not going to work for any other misc repair action.

addons/repair/functions/fnc_doRepair.sqf
```diff
@@ -31,9 +31,23 @@ _hitPointDamage = _hitPointDamage max ([_unit] call FUNC(getPostRepairDamage));
...
+    ([_vehicle] call EFUNC(common,getHitPointsWithSelections)) params ["_hitpoints"];
+    _hitpointGroup deleteAt 0; // Remove main group hitpoint
+    {
+        if (_x in _hitpoints) then {
+            _vehicle setHitPointDamage [_x, 0];
```
Glowbal: Besides the comment I made above; this will not work. Use `setVehicleHitPointDamage` targetEvent for this, becase the `setHitPointDamage` is best set locally.

bux578:
> Leaving out the glass was done on purpose. It's something I considered only being repeared at an repair facility / workshop. If we do want to be able to repair glass, I'd at least go for a separate action.

Good idea

jonpas:
> Leaving out the glass was done on purpose. It's something I considered only being repeared at an repair facility / workshop. If we do want to be able to repair glass, I'd at least go for a separate action.

I see, can be done either way. What do you mean by a separate action? In the example I used (Taru, Offroad... ) I just grouped the Glasses as an example.

> The hitpointGroup only works for a single action; I'd expand this so we can make us of it for any potential other hitpoints besides just glass.

Yeah, I slightly forgot about this one, will make use of multidimensional arrays to accomplish.

addons/repair/functions/fnc_canMiscRepair.sqf
jonpas: Indeed, will fix.

addons/repair/functions/fnc_doRepair.sqf
jonpas: What do you mean? The lead hitpoint (first in the config array) will get repaired only as calculated, while everything else in it fully as it doesn't matter if they get repaired fully or not really, at least not the the user directly (you wouldn't want to group glass and engine together for example).

addons/repair/functions/fnc_doRepair.sqf
jonpas: Roger, dammit locality.

addons/repair/functions/fnc_doRepair.sqf
Glowbal: I mean that you will end up resetting the damage for any hitpoint in the hitpoint group, even if the repair action is not for any hitpoint in the hitpoint group.

jonpas: Ah ok, same as above, yeah, fixing.

jonpas:
Updated config example in PR's original comment. I've left `IGNORED_HITPOINTS` disabled for now, so you can test this. I'd like to improve that in a separate PR after this and the other PR get merged (otherwise there will be tons of conflicts) by using a config entry for "facility-only repairable hitpoints" rather than hard-coded (adds framework for mods as well).

Issue #2214, opened 2015-08-20
Captive changes animation state when in FFV seat


Ghost77788:
ACE3 Version: 3.2.1
stable

**Mods:**
* @cba_a3
* @ace3
* @RHS

**Placed ACE3 Modules:**
None

**Description:**
When loading a captive in a vehicle and the assigned cargoindex is a ffv seat then the captive will for a second be seated correctly and then switch back to its captive animation state (ACE_AmovPercMstpScapWnonDnon). He will then be standing on his seat and glitching through the roof of the vehicle.
Ive tried to add a if statement to the vehiclemoveincaptive.sqf to check what cargoindex he is in (allturrets), however with rhs the cargoindexes dont line up with the correct ffv indexes.
Only other way to check was to check the animationstate of the captive and if equal to ACE_AmovPercMstpScapWnonDnon then move him to the next seat however seeing as the change of animation isnt instantaneous, you would have to implement a sleep command which might cause issues with other functions (not tested).

**Steps to reproduce:**
place uh60m (rhs), load captive into empty into the empty helicopter, he will get cargoindex 0 and he will after half a second (+-) change back to captive stance.

**Where did the issue occur?**
"Multiplayer", "Singleplayer"

Issue #2199, opened 2015-08-18
Unable to load arrested people in vehicle


Knight1981:
ACE3 Version: 3.2.1

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
basic medic system

**Description:**
Hey guys,

yesterday we played a rescue mission on our dedicated server. Therefore we have arrested the person and escorted him to an evac helicopter (in this case it was a Little Bird). That worked very well, but than we were not able to load the arrested person in the vehicle because no menu / option for this appeared. Could you please check this?

Thank you very much. :)

**Steps to reproduce:**
1. Arrest AI
2. Escort to a Little Bird
3. Try to load the arrested AI into the Helicopter

**Where did the issue occur?**
Multiplayer, dedicated server

**RPT log file:**
n.a

PabstMirror:
You have to be Escorting the captive to have the "load in vehicle" interaction show.
The action shows on the vehicle (you have to look through the captive).

There is also an action on the captive himself while escorting, but it is extremely awkward to access.


I would break this issue into 2 parts,

* Code failure: Do you think you where using it correctly and it failed to work? 

* Design: For ease of use, would it make sense to remove the "isEscorting" requirement.  IE, make it so a captive near a vehicle will just have a "load into vehicle" interaction on him. I don't see any harm, and I think it would make things cleaner and easier to use.

Knight1981:
Thank you for your answer. It is possible that we made a mistake, I will check it later and let you know if it works as you described.

But yeah, it is a little bit confusing (4 guys didn't know how to load the captive) :D I think it would be more easier when the captive is near the vehicle and you can look at him and say "load captive into vehicle". That would be the same behaviour as for patients.

Thank you :)

Ghost77788:
This is because the check on helicopters is broken within the doloadcaptive.sqf.
Is resolved in next update for ACE 3

Knight1981:
Ah okay, thank you for this information :) We wait eagerly for the next ACE update because of some other hotfixes (e.g. camp fire issue, slide show broken etc.). 

When can we expect the next release?

bux578:
> When can we expect the next release?

You won't get any ETA's from us. The version has to be ready codewise and has to be at least somewhat tested (preferably properly tested).

Knight1981:
[...]

PR #2249, opened 2015-08-24
Captive ffv animation fix + cleanup

### LINKED ISSUE #1742: Loading captive into RHS UAZ glitch
ACE3 Version: 3.1.1 stable

**Mods:**
* @cba_a3
* @ace3
* @RHSAFRF
* @RHSUSF

**Placed ACE3 Modules:**
none

**Description:**
Loading captive into any vehicle spot which is FFV (fire from vehicle)
Observe the captive standing upright in the seat.

First observed on the RHS UAZ. The UAZ door would also constantly open & close.

**Steps to reproduce:**
Zip tie a civilian
Escort him
Load him into any vehicle seat which is FFV, e.g. RHS UAZ, AAF Wildcat, or FIA offroad.

"Multiplayer"

**RPT log file:**
n/a

PabstMirror:
* Fix #1742 / #2214 Animation weird when loading captives into FFV cargo slots.
* Removes "escorting" requirement for loading into vehicles (ref #2199) (makes it match loading uncon)
* Move RemoveHandcuffs to Main (hands and pelvis block each other from that animation)
* By doing a playMove after a switchMove on the captive animation the game will respect the canFire in the animation config and actually block grenades.  Which means we can remove the `addActionEventhandler` from units which was running all the time.

For the FFV animation, I had to pick something that would work from any seat, TBH it doesn't look great but it's better than them standing straight up,  If someone wants to make a unique captive animation for every possible FFV animation, feel free to make a PR.


Glowbal:
Awesome :+1:

PR #2253, opened 2015-08-24
pl translation of spectator

gienkov:


addons/spectator/stringtable.xml
```diff
@@ -103,6 +104,7 @@
         <!-- Interface strings -->
         <Key ID="STR_ACE_Spectator_UnitTitle">
             <English>Spectator Units</English>
+            <Polish>Obserwatorzy</Polish>
```
SilentSpike: Hmm, perhaps this string needs to eb adjusted. It's actually the heading of the spectat*able* units list in the UI. I suspect your translation here means "observers" based on the spelling

gienkov: Right I wasn't sure what it is.

SilentSpike: The word "Units" will probably do the trick :+1:

gienkov:
This is ready.

PR #2254, opened 2015-08-24
Add WaitUntilAndExecute Function

jokoho48:
Because i saw while clean up the Code very Often a CBA EH that add only for wait Until and i think that we its better if we handle this only with one this could be better

Old Method
```
[{
    if (isNil QGVAR(someAwsomeVarName)) exitWith {};
    CODE
}, 0, []] CBA_fnc_addPerFrameHandler;
```

New Method
```
[
    {isNil QGVAR(someAwsomeVarName)},
    {CODE},
    []
] EFUNC(common,waitUntilAndExecute);
```


jonpas:
:+1:

addons/common/XEH_postInit.sqf
```diff
@@ -27,6 +27,17 @@
...
     GVAR(nextFrameNo) = diag_frameno + 1;
+
+    //Handle the waitUntilAndExec array:
+    _deleted = 0;
+    {
+        _x params ["_condition", "_code", "_args"];
+        if ((_args call _condition)) then {
+            GVAR(waitUntilAndExecArray) deleteAt (_forEachIndex - _deleted);
+            _deleted = _deleted + 1;
+            _args call _code;
```
commy2: "code" is a data type and therefore a bad variable name. The condition is code too. We usually use `condition` and `statement` for a construct like this.
Also consider not saving these in variables, because
`if ((_this select 2) call (_his select 0)) then {(_this select 2) call (_this select 1)};`
is still faster in my tests than saving them in a variable first and this is a performance critical PFH.

bux578: and add comments ;)

Glowbal:
Let's sort out the comments by @commy2 and @bux, and we can merge this PR :+1:

esteldunedain:
:+1: ACE has the best helper functions ever

MikeMatrix:
@esteldunedain wait until they're in CBA...

esteldunedain:
@MikeMatrix sure

PR #2257, opened 2015-08-24
Parse List Common Function

jonpas:
Used for parsing strings in modules, has params for optional remove whitespace (default is trim) and to check for object existence. Rerouted part of `assignObjectsInList` through this as well.


MikeMatrix:
:+1:

jonpas:
Status? As soon as this gets merged (or closed) I'll replace string operations with new commands.

commy2:
I'd say something, but I have no idea at all what this does.

jonpas:
It parses modules which take strings, (eg. list of medical vehicles or list of repair facilities), checks for nil, removes or trims whitespace (all parameters). It is used in all modules that can take a list and the code was duplicated in each of them, this simply makes one function that all modules can use. It can of course be used for something else than modules, hence why it's in common.

Glowbal:
I wonder if modules isn't a better location for this.

jonpas:
I'd say no, it doesn't take module logic, it can be used for anything else that requires parsing string lists.

jonpas:
Merge conflicts resolved.

jonpas:
Status?

commy2:
I still have no idea what this does, haha.

jokoho48:
it combine a function that used in closly every module :D

commy2:
:+1:  Merge then if anyone can confirm this works.

jonpas:
If someone else can give it a test that would be great!

jonpas:
Self-note #2565

jonpas:
Added vehicle support from #2565 for `synchronizedObjects` with an optional parameter.

PR #2260, opened 2015-08-25
Update Main Config File

jokoho48:
after Bux Commit in Slack

### DIFF
--- a/addons/main/config.cpp
+++ b/addons/main/config.cpp
@@ -541,26 +541,28 @@ class CfgPatches {
             "a3_weapons_f_vests",
             "a3data",
             "map_vr",
-            "extended_eventhandlers", "CBA_UI", "CBA_XEH", "CBA_XEH_A3"
+            "extended_eventhandlers", "CBA_UI", "CBA_XEH", "CBA_XEH_A3",
+            "ace_common"
         };
-        author[] = {"ACE Team"};
-        authorUrl = "";
-        versionDesc = "A.C.E.";
-        versionAct = QUOTE(['MAIN',_this] execVM '\z\ace\addons\main\about.sqf';);
+        author[] = {ELSTRING(common,ACETeam)};
+        authorUrl = "http://ace3mod.com/";
+        versionDesc = "ACE3";
+        //versionAct = QUOTE(['MAIN',_this] execVM '\z\ace\addons\main\about.sqf';);
+        versionAct = "";
         VERSION_CONFIG;
     };
 };
 
 class CfgMods {
     class PREFIX {
-        dir = "@ACE";
-        name = "Core - Advanced Combat Environment";
+        dir = "@ACE3";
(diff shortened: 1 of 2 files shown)


addons/main/config.cpp
```diff
@@ -541,26 +541,28 @@ class CfgPatches {
...
+        dir = "@ACE3";
+        name = "Advanced Combat Environment 3";
         picture = "A3\Ui_f\data\Logos\arma3_expansion_alpha_ca";
         hidePicture = "true";
         hideName = "true";
         actionName = "Website";
-        action = "http://ace.dev-heaven.net";
-        description = "Bugtracker: ";
+        action = "http://ace3mod.com/";
+        description = "Bugtracker: https://github.com/acemod/ACE3/issues";
```
jonpas: `"Issue Tracker: <link>"` rather? To be consistent with everywhere else.

addons/main/config.cpp
```diff
@@ -541,26 +541,28 @@ class CfgPatches {
...
+        versionAct = "";
         VERSION_CONFIG;
     };
 };
 
 class CfgMods {
     class PREFIX {
-        dir = "@ACE";
-        name = "Core - Advanced Combat Environment";
+        dir = "@ACE3";
```
jonpas: Directory is actually `@ace`.

addons/main/config.cpp
```diff
@@ -541,26 +541,28 @@ class CfgPatches {
...
             "map_vr",
-            "extended_eventhandlers", "CBA_UI", "CBA_XEH", "CBA_XEH_A3"
+            "extended_eventhandlers", "CBA_UI", "CBA_XEH", "CBA_XEH_A3",
+            "ace_common"
         };
-        author[] = {"ACE Team"};
-        authorUrl = "";
-        versionDesc = "A.C.E.";
-        versionAct = QUOTE(['MAIN',_this] execVM '\z\ace\addons\main\about.sqf';);
+        author[] = {ELSTRING(common,ACETeam)};
```
jonpas: Pretty sure you'll need `ECSTRING` (difference from L is that it adds a `$` in front which is needed in configs).

addons/main/config.cpp
bux578: I think `ACE` is fine here.

jokoho48: the folder name is @ace @bux578 i think @jonpas  are right with that

addons/main/config.cpp
jokoho48: 😬

addons/main/config.cpp
bux578: yeah that's what i meant

SzwedzikPL: what is purpose of `dir` ?

jokoho48: directory short form is dir

PR #2263, opened 2015-08-25
Bandage and tourniquet on adv. medical change proposal

gienkov:
Make bandaging longer as it is not very fast process.
Make putting on tourniquet faster cause it's should be easy to put one on limb, so it's actually useful now and with above change (putting it on before was longer than bandagin).


sgtBrowncoat:
I'd actually say the bandaging isn't that slow. A well trained medic can generally apply a TQ in about 5 seconds and then (usually with an assistant) can digitally ligate and pack a wound in about 30-45 seconds

PR #2265, opened 2015-08-25
Updated make.py with changes from CBA

jonpas:
- Changed `ACE_VERSION` to `project_version`
- Changed function `get_ace_version` to `get_project_version`
- Added additional exception for optional files

I will do the same back into CBA once this and the other PR get merged.


bux578:
If cba and ace use the same make.py why not put it in to one repo? Make a generalized make.py that fits all possible mod projects with a little project specific configuration?

jonpas:
I just pitched that idea to @ViperMaul myself last night @bux578 haha. Separate PR and git submodules maybe? 

Couple of months ago I modified make.py to be project-open as I wanted to use it for my projects and that's how CBA and ACE versions of it became the same.

ViperMaul:
Actually I like what @jonpas has done here. 
if you want to go an extra mile and come up with a modular design, I would look at it. But your skills and time are better saved for other features of ACE/CBA ;)

I will merge this within the hour.

ViperMaul:
Let's bump the ```__version__``` number on line 33

PR #2268, opened 2015-08-26
HR and BP enhancement

gienkov:
* It will display 0 HR and fail to find BP on limbs that have TQ on.
* Check for TQ is run only if medic is checking pulse on arms.


addons/medical/functions/fnc_actionCheckPulseLocal.sqf
```diff
@@ -46,9 +46,14 @@ if (_heartRate > 1.0) then {
     };
 };
 
+if (_selectionName in ["hand_l","hand_r"] && [_unit, _selectionName] call FUNC(hasTourniquetAppliedTo)) then {
```
jokoho48: add here lazyEelevation that could Improve the Performance
``if (_selectionName in ["hand_l","hand_r"] && {[_unit, _selectionName] call FUNC(hasTourniquetAppliedTo)}) then {`

Glowbal:
Doesn't take all limbs into account (`leg_l` and `leg_r`)

gienkov:
You don't check BP/HR on legs that's why check for arms only.

gienkov:
Did the same for BP, it will display ***You fail to find a blood pressure*** (it's in stringtable).

Issue #2259, opened 2015-08-25
getHitpoints function doesn't support reflector hitpoints


SzwedzikPL:
During tests of https://github.com/acemod/ACE3/pull/2230 i noticed that after dragging or carrying lamp reflectors are always working (event if damaged before).

The reason is that this dont support reflector hitpoints (or selections? only `getHit` return anything)

https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_fixFloating.sqf#L24

commy2:
I'll investigate. Doubt those can be manipulated with `setHitPointDamage` and `getHitPointDamage` though.

jonpas:
@commy2 look at #2279

SzwedzikPL:
@commy2 i've prepared https://github.com/acemod/ACE3/pull/2279 but have problem with `setdamage`.

commy2:
I advice that you use the new
`getAllHitPointsDamage cursorTarget`
command

commy2:
@SzwedzikPL
What about:
`[cursorTarget] call ace_common_fnc_getReflectorsWithSelections`
?

SzwedzikPL:
The problem is not that `getHitpoints` doesn't support reflectors, after dragging some object (i.e. lamp with reflectors) we'r calling:
https://github.com/acemod/ACE3/blob/master/addons/dragging/functions/fnc_dropObject.sqf#L60
And in fixFloating fnc we do not support reflectors, so thats why reflectors are always working after drag/carry.

commy2:
k. I understand now.

I'll see that I fix this after 3.2.2. Currently cleaning up (== incorperating new commands...) all the other stuff with hitpoints.

https://github.com/acemod/ACE3/pull/2319

SzwedzikPL:
@commy2 ok. Should I close https://github.com/acemod/ACE3/pull/2279 ?

commy2:
Nah. Leave it open for now.

commy2:
Fixed in  #2601. Sry for this taking a month. Had other priorities even though this ended up being really simple.

SzwedzikPL:
This is not fixed. Please test it with https://github.com/acemod/ACE3/pull/2230

1. Place `Portable Lights (double)` from `Objects (construction)`
2. Shot to one of the reflectors
3. Carry lamp
4. Place lamp

You'll see that after placing both reflectors are working.

PabstMirror:
[...]

PR #2279, opened 2015-08-26
getHitpoints and fixFloating cleanup & enhancement

SzwedzikPL:
Fix for https://github.com/acemod/ACE3/issues/2259

Added support for hit selections in getHitpoints
Little cleanup & use of `getAllHitPointsDamage`

I dont know why but after `_object setDamage damage _object` `setHit` doesnt work. If someone knows how to fix that i will be grateful :)

### DIFF
--- a/addons/common/functions/fnc_fixFloating.sqf
+++ b/addons/common/functions/fnc_fixFloating.sqf
@@ -19,14 +19,21 @@ _object = _this;
 if (!local _object) exitWith {};
 
 // save and restore hitpoints, see below why
-private ["_hitPoints", "_hitPointDamages"];
+private ["_hitPointsAndSelections", "_hitPointDamages", "_hitSelectionDamages"];
 
-_hitPoints = [_object] call FUNC(getHitpoints);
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_fixFloating.sqf
```diff
@@ -19,14 +19,21 @@ _object = _this;
...
+
+{
+	_object setHitPointDamage [_x, _hitPointDamages select _forEachIndex];
+} forEach _hitpoints;
 
 {
-    _object setHitPointDamage [_x, _hitPointDamages select _forEachIndex];
-} forEach _hitPoints;
+	_object setHit [_x, _hitSelectionDamages select _forEachIndex];
+} forEach _hitselections;
```
commy2: This is just to trigger PhysX. No need to edit this function.

commy2:
I vote against merging this. We should make use of the new hit indices from now on.

commy2:
Implemented a better way in #2601 using new commands.

PR #2283, opened 2015-08-26
Add canInteractWith checks for cargo

### LINKED ISSUE #2281: Drag bug
ACE3 Version: https://github.com/acemod/ACE3/commit/ca991712e8f1e48541c181ac794e7101a42071d7

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *-*

**Description:**
*Interaction option "drop" is drawed on top of Interaction with object when you are carrying/dragging it.*
http://steamcommunity.com/sharedfiles/filedetails/?id=506907789

**Steps to reproduce:**
* *Drag a chest or carry a chest.*
* *Interaction menu on chest -> drop, notice how it's not "extending" as it should*

*"Singleplayer"*

**RPT log file:**
*-*

PabstMirror:
Should fix #2281 
Will probably make #2185 unnecessary 

CreateAction doesn't add the canInteractWith check, so that's why both actions are showing at the same time when they should be mutually exclusive.
Should also solve stuff like people loading objects that others are carrying.

### DIFF
--- a/addons/cargo/functions/fnc_canLoad.sqf
+++ b/addons/cargo/functions/fnc_canLoad.sqf
@@ -18,6 +18,8 @@
 
 params ["_player", "_object"];
 
+if (!([_player, _object, []] call EFUNC(common,canInteractWith))) exitWith {false};
+
 private ["_nearestVehicle"];
 _nearestVehicle = [_player] call FUNC(findNearestVehicle);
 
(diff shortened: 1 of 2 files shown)


jonpas:
How will that replace #2185 ? Carrying and dragging is not part of cargo.

jonpas:
Another thought, will that still allow direct loading of objects without dropping them first?

PabstMirror:
Right now when carrying, 2 actions show
Release
and Interactions which expands out to the "load" action from cargo.

This will block the load action on the object when carrying

I'm not sure if it is safe to load directly from carry.
fnc_dropObject_carry.sqf will do a `detach` a frame after loading into a vehicle, which attaches it to the vehicle.

jonpas:
> This will block the load action on the object when carrying

What if there are other actions, will it block all inside main actions?

> I'm not sure if it is safe to load directly from carry.

I've tried it with my PR and I saw no issues, it could be something hidden though.

PR #2286, opened 2015-08-26
Remove deprecated drop action

gienkov:


alganthe:
note: This is for carrying and is *NOT* a duplicate.

Glowbal:
I don't consider this a deprecated action. Dropping a carried/dragged object is one thing where I'd consider the scroll wheel action is to be useful/easy to use.

bux578:
What @Glowbal said!

gienkov:
Keybind would be better for this.

SilentSpike:
Disagree, over-saturating the game with keybinds is a horrible idea

jokoho48:
:-1: i dont like the idea because i personaly use the action menu and keybinding is a very horrible idea like @SilentSpike

MikeMatrix:
Same opinion as @SilentSpike. Creating an option for general purpose things is a nice idea, but don't force it.
Also I use the addAction menu for that as well.

SilentSpike:
Adding a self interaction option for it could work though (if there isn't already)

MikeMatrix:
I believe there is already

PR #2289, opened 2015-08-27
Removed ability to interact with tactical ladder while climbing

jonpas:
You could pick up the ladder (and fall) or position it (and remain in climbing animation until you redeployed it). This fixes both.

### DIFF
--- a/addons/tacticalladder/CfgVehicles.hpp
+++ b/addons/tacticalladder/CfgVehicles.hpp
@@ -1,4 +1,3 @@
-
 class CfgVehicles {
     class Man;
     class CAManBase: Man {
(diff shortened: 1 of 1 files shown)


PabstMirror:
Can check via config:

`{((getNumber (configFile >> "CfgMovesMaleSdr" >> "States" >> (animationState _player) >> "onLadder")) == 0)}`

jonpas:
@PabstMirror thanks!

commy2:
The underlying issue is that you can interact on a ladder.
This will be fixed in https://github.com/acemod/ACE3/pull/2320

This PR can be closed I think.

jonpas:
Replaced by #2320

PR #2296, opened 2015-08-28
Put vehicle mainActions on starting cursor pos

PabstMirror:
Rough WIP, just trying to get feedback on the idea

https://www.youtube.com/watch?v=cTrxSeetLzQ

Wherever you are looking on a cursorTarget when opening the interaction menu will be where the mainActions point will be.

It should save a step on things like tires and crates because you don't have to first aim at the "interaction" point and then select an option.

There are a lot of things to consider for this and this could easily be a user option.
Right now it does a basic overlap check, but that should be done by angle not distance.


jonpas:
Looks very nice!

MikeMatrix:
I agree, this behavior should be desirable. Depending on how well the implementation functions, we should consider making this the default and adding an option to turn it off in.

bux578:
How does it look/work on a unit?

PabstMirror:
I don't think it makes sense on units, it's fairly easy to find the "interactive pelvis" point and actions are already crowded in.

With big vehciles it can be tough finding the action point because the vehicle is large.  For example, trying to find unload from a HEMTT (transport truck) while at the rear of the vehicle, the action point is way in the front.

The point can also be in weird spots, like the RHS UAZ I think it is on the ground.

SzwedzikPL:
Just f*cking awesome! Pls, merge ASAP. We need this.

alganthe:
In it current state this fucks up vehicle repair interactions, they don't appear anymore.

SzwedzikPL:
Status? Really must have.

SzwedzikPL:
Any chance for 3.5.0?

esteldunedain:
@PabstMirror, you are a genius. :+1:

addons/interact_menu/functions/fnc_keyDown.sqf
```diff
@@ -95,4 +95,79 @@ if (GVAR(menuAnimationSpeed) > 0) then {
...
+    TRACE_1("Checking for cursor target",_cursorTargets);
+    if (_cursorTargets isEqualTo []) exitWith {};
+    (_cursorTargets select 0) params ["_intsectPosASL", "", "_target"]; //We need variable `_target` set for the actionPosition call
+    if (isNull _target) exitWith {TRACE_1("Ignoring Terrain (null)",_target);};
+
+    //Ignore Man (confusing, and would have trouble finding a free spot)
+    if (_target isKindOf "CAManBase") exitWith {TRACE_1("Ignoring Man",_target);};
+
+    private _interestModelPos = _target worldToModelVisual (ASLtoAGL _intsectPosASL);
+    private _classActions = missionNamespace getVariable [(format [QGVAR(Act_%1), typeOf _target]), []];
```
SzwedzikPL: this line needs to be updated becouse of https://github.com/acemod/ACE3/pull/3547
should be `private _classActions = GVAR(ActNamespace) getVariable typeOf _target;`

SzwedzikPL:
My community is testing if for more than month now. No problems so far. :+1: for merge asap

jonpas:
@PabstMirror Status on this?

The only thing I am worried about this is actions that require shorter distance, there is nothing telling you there are more actions. While that's true for current system as well, you know what to expect more than in this case (I don't know how to explain really, hope you understand).

PabstMirror:
With #3342 I'm not sure if this is necessary anymore

bux:
@PabstMirror Status on this one?

PabstMirror:
#3342 really fixed the issue (in a completely different way)
Probably close for now

PR #2297, opened 2015-08-28
Add Hearing Protection to Helmets with Peltor Headsets

bux578:
`"H_HelmetSpecB"` and `"H_HelmetB_light"` (and their variations) have visible Peltor Headsets on the 3d model. Those are electronical and will block/damp sounds that are too loud as soon as the sound appears but won't dampen environment sounds like earplugs would.

image

### DIFF
--- a/addons/hearing/CfgWeapons.hpp
+++ b/addons/hearing/CfgWeapons.hpp
@@ -47,5 +47,16 @@ class CfgWeapons {
         GVAR(lowerVolume) = 0.60;
     };
     class H_Cap_marshal: H_Cap_headphones {};
-};
 
+    class H_HelmetB_light: H_HelmetB {
+        GVAR(protection) = 0.8;
+        GVAR(lowerVolume) = 0.20;
+    };
+
+    class H_HelmetB_plain_mcamo;
+    class H_HelmetSpecB: H_HelmetB_plain_mcamo {
+        GVAR(protection) = 0.8;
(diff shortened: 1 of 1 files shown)


SzwedzikPL:
:+1:

alganthe:
:+1:

jokoho48:
:+1:

MikeMatrix:
:+1:

jonpas:
:banana: enough thumbs!

654wak654:
:+1:

654wak654:
:banana: enough thumbs but never enough bananas.

R4ppiT:
Does this work with RHS Ops Core helmets?

bux578:
@R4ppiT If they inherit from those two classes yes, if not no.

FLundin:
Could the Hearing module have a field to write in item class names for items that will apply the same effects like the helmets with Peltor Headset? (for situations where some mods may not be compatible with this feature yet)

esteldunedain:
> Could the Hearing module have a field to write in item class names for items that will apply the same effects like the helmets with Peltor Headset? (for situations where some mods may not be compatible with this feature yet)

Nope, that'd be messy. You can instead write a small pbo patch to add the required properties for your addons.

bux578:
And, please don't revive closed issues. Feel free to join our public chat: http://slackin.ace3mod.com , though, and ask if you need help creating a patch/compatibility pbo.

PR #2312, opened 2015-08-30
increase reach in 3rd person for opening doors, fix #2301

### LINKED ISSUE #2301: Opening Doors Isn't Possible in 3rd Person View
ACE3 Version: 3.2.1, Stable branch

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
*No modules*

**Description:**
The "Open Door" function isn't possible in 3rd person view.

Think it's because [interaction_fnc_openDoor](https://github.com/acemod/ACE3/blob/master/addons/interaction/functions/fnc_openDoor.sqf) calls [interaction_fnc_getDoor](https://github.com/acemod/ACE3/blob/master/addons/interaction/functions/fnc_getDoor.sqf#L24) with a distance of 2 meters, which is then used to check the distance between the camera and the door, not the player object and the door. Since the 3rd person camera will be further then 2 meters away from the door (even when the player object is right beside it), the functions will decide the player is not close enough.

Maybe instead of [interaction_fnc_openDoor line 21](https://github.com/acemod/ACE3/blob/master/addons/interaction/functions/fnc_openDoor.sqf#L21), [cameraView](https://community.bistudio.com/wiki/cameraView) could be used, it whould be 2 if internal and 2 + (distance between camera and player) for external.

**Steps to reproduce:**
* Find a door.
* Open the door in 1st person.
* Switch to 3rd person.
* Try to open the door.

"Multiplayer", "Singleplayer"

commy2:


MikeMatrix:
:+1:

PR #2313, opened 2015-08-30
fix weapon select in ffv slots not working

commy2:


addons/weaponselect/XEH_postInit.sqf
```diff
@@ -130,12 +130,12 @@ if !(hasInterface) exitWith {};
 ["ACE3 Vehicles", QGVAR(EngineOn), localize LSTRING(EngineOn),
 {
     // Conditions: canInteract
-    if !([ACE_player, (vehicle ACE_player), []] call EFUNC(common,canInteractWith)) exitWith {false};
+    if !([ACE_player, vehicle ACE_player, []] call EFUNC(common,canInteractWith)) exitWith {false};
     // Conditions: specific
     if !((ACE_player != vehicle ACE_player) && {ACE_player == driver vehicle ACE_player} && {!isEngineOn vehicle ACE_player}) exitWith {false};
 
     // Statement
-    (vehicle ACE_player) engineOn true;
```
bux578: why? this is better/faster readable with brackets

commy2: I get cataracts looking at these brackets.

jonpas: I get cataracts when looking at the no-brackets version. :grin:

PR #2315, opened 2015-08-30
fix autoriflemen spawning with 0x frag grenade bug

commy2:
This fixes the issue with unselectable grenades when spawning as autorifleman of every faction (except for FIA?).
Reason is that grenades can be added to the throw muzzle by config, even though there is not enough space in the inventory. This fix was in AGM, but probably got lost during porting...
The spare barrel still weights 1.36kg. Considering it's "usefulness" this seems more approriate anyway.


alganthe:
Where is this issue coming from ?
BI didn't changed the loadouts in ages and the last time I checked (2 weeks ago) autoriflemen wouldn't miss anything.

commy2:
Arma 1.50, ACE current dev. The issue was never fixed in ACE3.



Notice top right:
0x RGO Grenade.

PR #2316, opened 2015-08-30
fix custom cameras (for intros) being broken with ACE, fix #2248

### LINKED ISSUE #2248: "camcreate" command non-functional while 3.2.1 is loaded
ACE3 Version: 3.2.1 Stable

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *none*

**Description:**
The "camCreate" sqs command doesn't do anything while 3.2.1 is loaded.

**Steps to reproduce:**
* Create a mission on any map
* Place a "player" unit and an object that can be used as a camera (I used a gamelogic)
* In the camera object's init, execute an "example.sqs" file
* Save the mission as a user mission
* In the mission folder create an "example.sqs"
* Edit the "example.sqs" and enter a camCreate command, example:

_camera = "camera" camCreate [0,0,0]
_camera cameraEffect ["internal", "back"]
* Save the sqs file and the mission.
* Preview the mission in editor, or export it and play it in singleplayer or multiplayer.

The custom camera will not be created and will not return any script errors.

*Singleplayer, Multiplayer*

**RPT log file:**
*http://pastebin.com/cvam1MGc*

commy2:
~~The camera for the pip opctics break intros with cameras. This is because the camera effect "internal" breaks whenever a new camCreate command is used.
Unfortunately there is no command that returns all cameras. The only solution for this problem I can come up with (without adding random sleeps and hoping for the best) is for mission makers to use a predifned variable name for their cameras.~~

```
~_camera = "camera" camCreate [0,0,0];~
~_camera cameraEffect ["internal", "back"];~
```

~becomes:~

```
~ACE_CustomCamera = "camera" camCreate [0,0,0];~
~ACE_CustomCamera cameraEffect ["internal", "back"];~
```

~~We probably should add this to the wiki if this PR gets merged.
Also fixes the camera being broken when entering the VA for the first time.~~

Completely solved this by using the wonderfull new command:
```
allMissionObjects "camera"
```

### DIFF
--- a/addons/optics/XEH_postInit.sqf
+++ b/addons/optics/XEH_postInit.sqf
@@ -3,26 +3,22 @@
 
 if (!hasInterface) exitWith {};
 
-#define ALL_CAMERAS [ \
-    missionNamespace getVariable ["BIS_DEBUG_CAM", objNull], \
-    missionNamespace getVariable ["BIS_fnc_camera_cam", objNull], \
-    uiNamespace getVariable ["BIS_fnc_arsenal_cam", objNull], \
-    uiNamespace getVariable ["BIS_fnc_animViewer_cam", objNull], \
-    missionNamespace getVariable ["BIS_fnc_establishingShot_fakeUAV", objNull], \
-    missionNamespace getVariable ["ACE_CustomCamera", objNull] \
-]
+GVAR(camera) = objNull;
 
 0 = 0 spawn {
     waituntil {!isNull ACE_player};
-    waituntil {sleep 0.5; {!isNull _x} count ALL_CAMERAS == 0};
+    waituntil {sleep 0.5; !(call EFUNC(common,isfeatureCameraActive)) && isNull (missionNamespace getVariable ["ACE_CustomCamera", objNull])};
+
+    GVAR(camera) cameraEffect ["TERMINATE", "BACK"];
+    camDestroy GVAR(camera);
 
     // PiP technique by BadBenson
-    GVAR(camera) = "camera" camCreate positioncameratoworld [0,0,0];
+    GVAR(camera) = "camera" camCreate positionCameraToWorld [0,0,0];
     GVAR(camera) camSetFov 0.7;
     GVAR(camera) camSetTarget ACE_player;
     GVAR(camera) camCommit 1;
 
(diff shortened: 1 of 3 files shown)


addons/optics/XEH_postInit.sqf
```diff
@@ -3,8 +3,18 @@
...
+    missionNamespace getVariable ["BIS_fnc_camera_cam", objNull], \
+    uiNamespace getVariable ["BIS_fnc_arsenal_cam", objNull], \
+    uiNamespace getVariable ["BIS_fnc_animViewer_cam", objNull], \
+    missionNamespace getVariable ["BIS_fnc_establishingShot_fakeUAV", objNull], \
+    missionNamespace getVariable ["ACE_CustomCamera", objNull] \
+]
+
 0 = 0 spawn {
     waituntil {!isNull ACE_player};
+    waituntil {sleep 0.5; {!isNull _x} count ALL_CAMERAS == 0};
```
jonpas: PFH wait? If not possible then it can be replaced by #2254

addons/optics/XEH_postInit.sqf
```diff
@@ -3,8 +3,18 @@
 
 if (!hasInterface) exitWith {};
 
+#define ALL_CAMERAS [ \
```
jonpas: Can be replaced by `EFUNC(isFeatureCameraActive)` and on line 17 simply check boolean return.

commy2: Yup. Same in common. Have it replaced locally.

jonpas: :+1:

commy2:
jonpas
WaitAndExecute seems like an overkill here. I just have don't want to check this stuff every frame on init and it's not really time critical.

addons/optics/XEH_postInit.sqf
```diff
@@ -23,18 +29,19 @@ if (!hasInterface) exitWith {};
     };
 }] call EFUNC(common,addEventHandler);
 
-// camera has to be re-created every ACE_time another camera is created. Otherwise r2t is either black or transparent. @todo Add popular custom cameras to the event in ACE_common.
+// camera has to be re-created every time another camera is created. Otherwise r2t is either black or transparent. @todo Add popular custom cameras to the event in ACE_common.
```
commy2: GJ search and replace

PR #2319, opened 2015-08-30
make use of new getAllHitpointsDamage command

commy2:


addons/common/XEH_preInit.sqf
```diff
@@ -215,6 +215,7 @@ PREP(getConfigGunner);
 PREP(getConfigCommander);
 PREP(getHitPoints);
 PREP(getHitPointsWithSelections);
+PREP(getSelectionsWithoutHitPoints);
```
jonpas: A better name might be `getSelections` so it would mirror `getHitPoints`, that automatically implies it's without anything else.

commy2: But those are only the selections that don't have any hitpoint assigned, not all damageable selections.

MikeMatrix: Why not just make them into one function anyways and parameterize it?

commy2: >Why not just make them into one function anyways and parameterize it?
It's a completely different approach as getHitpoints. I personally don't like the overhead that comes with checking optinal parameters when nothing but one line is shared between both functions.

PR #2320, opened 2015-08-30
disable all actions while being on a ladder by default, close #2289

commy2:


jonpas:
Continuing discussion from #2289 as it's more fitting for here.

Assuming I understand the title of the PR, exception will be added to all actions globally? If so, I disagree with 2 reasons.

- Why would anyone not be able to change teams or pull up small handheld devices (just an example) on a ladder?
- Disables interaction with everything around you, not letting you extend even one arm? (Good example is that it disables ability to repair while on the ladder, which I think is pretty interesting, I was planning to extend the repair distance but if we are going simulation then you can't really repair what you can't reach and the ladder adds an interesting dynamic - that or higher ground.)

commy2:
>Why would anyone not be able to change teams

I already added exceptions for those. Actions are only disabled _by default_. Just like when you are swimming.

>or pull up small handheld devices

We could add exceptions for these aswell, although you shouldn't have a free hand for a _handheld_ devices in theory.

Draging and repairing already cause animation issues while being on a ladder on master. This would all be resolved with this PR.

PabstMirror:
@commy2  why `ACE_isLadder` instead of the arma config `onLadder`

they should be identical, plus we get more compatibility with any possible addon animations

commy2:
ACE_isLadder is probably a remnant from AGM. I was looking for a config entry like onLadder, but at that time I could find that. Either overlooked it, or it was added later.

I'll take a look if they really are identical.

jonpas:
> I already added exceptions for those. Actions are only disabled by default. Just like when you are swimming.

I mixed up exceptions again, my bad.

> We could add exceptions for these aswell, although you shouldn't have a free hand for a handheld devices in theory.

I meant small devices, not literally handheld, those you can hold in one hand and normally take out and put back in as you could in reality.

> Draging and repairing already cause animation issues while being on a ladder on master. This would all be resolved with this PR.

I tested repairing on ladder just few days ago when my repair framework PRs were merged, I had no issues with animations on it since when you are on it it won't play any animation. I tested it again just now.

commy2:
```
A = "getNumber (_x  >> 'onLadder') == 1" configClasses (configFile >> "CfgMovesMaleSdr" >> "States");
B = [A, {configName _this}] call ACE_common_fnc_map;
```

`["LadderCivilStatic","LadderCivilOn_Top","LadderCivilOn_Bottom","LadderCivilUpLoop","LadderCivilDownLoop","LadderRifleStatic","LadderRifleOn","LadderRifleUpLoop","LadderRifleDownLoop"]`


```
A = "getNumber (_x  >> 'ACE_isLadder') == 1" configClasses (configFile >> "CfgMovesMaleSdr" >> "States");
B = [A, {configName _this}] call ACE_common_fnc_map
```

`["LadderCivilStatic","LadderCivilOn_Top","LadderCivilOn_Bottom","LadderCivilUpLoop","LadderCivilTopOff","LadderCivilDownOff","LadderCivilDownLoop","LadderRifleStatic","LadderRifleOn","LadderRifleTopOff","LadderRifleDownOff","LadderRifleUpLoop","LadderRifleDownLoop"]`

ACE_isLadder is superior.

jonpas:
Wouldn't it be friendlier to rather change `onLadder` in in the missing animations?

commy2:
Might break stuff from BI. I have no idea what they use that entry for.

jonpas:
Hmh, true.

PabstMirror:
My bad commy, looks like the short transitions animations aren't the same.
I'm guessing `onLadder` is probably true when the unit is directly attached to a ladder.

commy2:
Yes. So, merge?

jonpas:
With this it will be almost impossible to repair certain hitpoints (eg. Transmission or Tail Rotor on a Taru), either allow repair on ladder (which I tested yesterday and it worked perfectly) or increase repair distance (but then you can pretty much repair all hitpoints from anywhere).

commy2:
IDK. I can't repair while being on a ladder even without this PR. Nothing happens.

commy2:
Added exception for repair. Merge?

jonpas:
:+1:

PR #2321, opened 2015-08-30
Hearing stuff

### LINKED ISSUE #1988: Explosions have no effect on hearing loss

PabstMirror:
Fix 	#1988  - the "hearing damage" switched from a 0-1 scale to 0-20.
It's still not perfect because we can only get an event when taking direct explosive damage,

Also got some nice performance improvements with new `isKindOf` command in firedNear
0.230 ms vs 0.081 ms

### DIFF
--- a/addons/hearing/CfgEventHandlers.hpp
+++ b/addons/hearing/CfgEventHandlers.hpp
@@ -29,7 +29,7 @@ class Extended_FiredNear_EventHandlers {
 class Extended_Explosion_EventHandlers {
     class CAManBase {
         class GVAR(ExplosionNear) {
-            clientExplosion = QUOTE( if (GVAR(enableCombatDeafness) && {_this select 0 == ACE_player}) then {_this call FUNC(explosionNear)}; );
+            clientExplosion = QUOTE(_this call FUNC(explosionNear););
         };
     };
 };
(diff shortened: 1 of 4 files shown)


jokoho48:
:+1: for the performance improvement

MikeMatrix:
> scale to 0-20.

Looking at the code, isn't that 0-30 there?

PabstMirror:
Explosive event is passing in the `damage` done to player, I'm not entirely sure how to map that to hearing damage.  With frag, a grenade that does 0.5 damage is probably close enough to kill, so I think we want to scale this up quite a bit.

PabstMirror:
Just checked AGM, they did `_damage * 2` for a 0-1 scale.

MikeMatrix:
Oh yea I was only refering to https://github.com/acemod/ACE3/pull/2321/files#diff-0b02185c83f52d8ff574c5705df11e31R29 where it says `(0 max damage) * 30`

esteldunedain:
:+1: Great stuff.

Issue #1342, opened 2015-05-22
Goggle effects removed when using sights in Firing from Vehicles position


SonixApache:
ACE3 Version: 3.0.1

Mods:
@cba_a3
@ace3

**Placed ACE3 Modules:**
N/A

**Description:**
When using sights in a Firing from Vehicles position, the effect of goggles are cancelled out.
This is the case for any Firing from Vehicles position, including those found in trucks, helicopters and tanks.

**Steps to reproduce:**
Attempt to use your sights while in a Firing from Vehicles position.

Repro mission: http://a.pomf.se/dbikkb.pbo

    Launch scenario
    Raise weapon
    Bring up sights

**Where did the issue occur?**
Singleplayer, multiplayer

SilentSpike:
Note to self: common,canUseWeapon

PR #2327, opened 2015-08-31
Fix gunner goggles

### LINKED ISSUE #1712: Screen get's whitewashed/gamma increase when in helicopters
ACE3 Version: 3.1.1.1
Stable

Mods:
@cba_a3
@TFR
@ace3
@ALiVE
@RHSUSAF
Excluding various maps

Placed ACE3 Modules:
http://pastebin.com/7ue3XikZ

Description:
Helicopter Pilots and crew are experiencing an increase in their screen gamma/brightness making everything very white/bright that then gets worse the longer time they spend in the helicopter http://imgur.com/a/0HzB0

The issue appears from starting the engine and only disappears when shutting the engine off, it's also worth mentioning that going into your settings and messing with your gamma will for a very short while reset/fix it.

Steps to reproduce:
Unknown

Where did the issue occur?
From entering a helicopter with its engine on

RPT log file:
Unavailable

### LINKED ISSUE #1721: Goggles effect disappears when aiming as Gunner
ACE3 Version: 3.1.1.1 Release and 3.1.2.0 ce0bbe60d39411398a26161225e0c50c977a61e0

**Mods:**
* `@cba_a3`
* `@ace3`

**Placed ACE3 Modules:**
* None

**Description:**
Goggles effect disappears in gunner  view making switching painful as it keeps changing the tint/brightness.

**Steps to reproduce:**
* Put on some Goggles (eh. Shades)
* Enter a vehicle as Gunner (eg. Offroad)
* Aim down the sights and see the effect disappear
* Look normally without aiming and see the effect reappear

Singleplayer - Editor
Multiplayer - Dedicated

**RPT log file:**
N/A

SilentSpike:
Fixes #1721 and #1342

Originally I introduced a canUseWeapon check to fix the FFV seats. Then when it came to considering the second issue I think it makes sense to just remove any special treatment for gunners anyway:
  - It simplifies the code and offers a slight performance benefit (that function is used by a PFH)
  - If you intend to be a vehicle gunner, you probably shouldn't be wearing unsuitable glasses anyway. Besides it's not the end of the world if you do, you can still see after all.
  - It makes things more consistent
    - Overlay already persists when using a UAV
    - Would be inconsistent for the weirdly configured commander seats/co-pilot seats
    - All the other effects checked by that function will now still take place when in a vehicle (if the vehicle goes underwater/etc.)
  - We will be introducing a client setting for the goggles overlay so nobody is forced to use it by default.

### DIFF
--- a/addons/goggles/functions/fnc_checkGoggles.sqf
+++ b/addons/goggles/functions/fnc_checkGoggles.sqf
@@ -18,7 +18,7 @@
 if (!alive ace_player) exitWith {};
 if (true) then {
     // Detect if curator interface is open and disable effects
-    if (!isNull(findDisplay 312)) exitWith {
+    if !(isNull curatorCamera) exitWith {
         if (GVAR(EffectsActive)) then {
             call FUNC(removeGlassesEffect);
         };
--- a/addons/goggles/functions/fnc_isGogglesVisible.sqf
+++ b/addons/goggles/functions/fnc_isGogglesVisible.sqf
@@ -15,14 +15,16 @@
  */
 #include "script_component.hpp"
 
-PARAMS_1(_unit);
-
+params ["_unit"];
 private ["_currentGlasses", "_result", "_position", "_visible"];
 
 _currentGlasses = goggles _unit;
 _result = false;
 
-if ((vehicle _unit) != _unit) exitWith {(cameraView != "GUNNER")};
+if ((vehicle _unit) != _unit) exitWith {
+    (cameraView != "GUNNER") ||
+    {[_unit] call EFUNC(common,canUseWeapon)}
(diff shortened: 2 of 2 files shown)


SilentSpike:
Labelled discussion as this contains a minor design decision that others might want to weigh in on

PabstMirror:
You can access your inventory while mounted, so this should be fine.

SilentSpike:
No objections in the past 48 hours, merging

CorruptedHeart:
I'm going to weigh in on this a bit late. But the original reason for it was due to the inventory not being accessible in a vehicle.

So now that, that is possible. No reason to keep it around.

PR #2329, opened 2015-08-31
setStructuralDamage function

commy2:
Killing a soldier by shooting him in the chest causes the whole body to get bloody. This is because `setDamage 1` is called in fnc_setDead. I wrote a function that can manipulate the structural damage without changing the hitpoints to prevent this.
Possibly needs to be used in more places. I have no idea when setDamage is intended (e.g. `setDamage 0` for a complete heal) and where it's not.
Uses the new hit index system. Pretty neat.

Plz review Glowbal.


Glowbal:
Should be good. I suppose this needs to go in with the other medical adjustments for 3.3. I will look at this more closely this week.

PR #2331, opened 2015-08-31
remove BI blood overlay

commy2:
On the medical-adjustments branch the wound visuals on the uniform are flickering when shooting someone, because some hitpoints are reset for one frame.
I was told that this was added to prevent some BI effects when getting hit by a bullet.
No idea which effects exactly, but if it's only the blood fx, then this PR does it by config and not by scripting which is the prefered way imo.

### DIFF
--- a/addons/medical/functions/fnc_showBloodEffect.sqf
+++ b/addons/medical/functions/fnc_showBloodEffect.sqf
@@ -1,0 +1,158 @@
+/*
+ * Author: KoffeinFlummi, Glowbal
+ * Does the blood screen effect, just like BIS_fnc_bloodeffect, but in non-sheduled environment.
+ *
+ * Arguments:
+ * 0: Effect multiplier <NUMBER>
+ *
+ * Return Value:
+ * None
+ *
+ * Public: Yes
+ */
+#include "script_component.hpp"
+
+params ["_bloodRemaining"];
+
+disableSerialization;
+
+// get already existing controls, or create them
+private ["_fxBloodControls", "_bloodCtrl1", "_bloodCtrl2", "_bloodCtrl3"];
+
+_fxBloodControls = GETUVAR(GVAR(FXBloodControls),[]);
+
+if (count _fxBloodControls != 3) then {
+    _bloodCtrl1 = findDisplay 46 ctrlCreate ["RscPicture", -1];
+    _bloodCtrl2 = findDisplay 46 ctrlCreate ["RscPicture", -1];
+    _bloodCtrl3 = findDisplay 46 ctrlCreate ["RscPicture", -1];
+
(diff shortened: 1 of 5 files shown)


gienkov:
:+1:

Glowbal:
This does not seem to work. I still had the damage effects.

commy2:
Can you define damage effects? The blood should be gone.

commy2:
I have no idea how to remove the blood vignette.

addons/medical/functions/fnc_showBloodEffect.sqf
```diff
@@ -0,0 +1,161 @@
+/*
+ * Author: KoffeinFlummi, Glowbal
+ * Does the blood screen effect, just like BIS_fnc_bloodeffect, but in non-sheduled environment.
+ *
+ * Arguments:
+ * 0: Effect multiplier <NUMBER>
+ *
+ * Return Value:
+ * nil
```
jokoho48: None

addons/medical/functions/fnc_showBloodEffect.sqf
```diff
@@ -0,0 +1,161 @@
...
+// reset everything
+_bloodCtrl1 ctrlSetFade 1;
+_bloodCtrl2 ctrlSetFade 1;
+_bloodCtrl3 ctrlSetFade 1;
+
+_bloodCtrl1 ctrlCommit 0;
+_bloodCtrl2 ctrlCommit 0;
+_bloodCtrl3 ctrlCommit 0;
+
+switch (true) do {
```
jokoho48: test if () exitWith {}? that could be faster

Glowbal:
@commy2 what is the status on this? I don't think it removes the blood overlay anymore, but instead fixes the blood effects, right?

commy2:
This does: 
- remove blood texture when being hit
- remove other permanent blood texture when being damaged
- adds a equvalent to BIS_fnc_bloodEffect that uses sheduled environment and therefore works way better

- it does not: remove the permant pp-effect while being damaged. It's a ColorCorrections type effect I think with unknown origin, possibly hard coded. No idea how we would remove that.

commy2:
I think it could be merged, but you said:
>This does not seem to work. I still had the damage effects.

No idea what you mean by that, except that color correction that looks similar to a vignette, but isn't one technically. Possibly provide a screenshot if I missed something?

gienkov:
Its that "red flashing" you get in vanilla A3 which is incredibly annoying.
I think there is BI function that regulates intensity of this piece of sh*t.
11 paź 2015 3:21 PM "commy2" <notifications@github.com> napisał(a):

Glowbal:
Going to try and get this in today.

Issue #2310, opened 2015-08-30
Main Menu Version dont work after 1.51


jokoho48:
RTP Issue: `Web address 'http://ace3mod.com/version.html/' won't be open. Current restrictions do not allowed it.`

Reason: `Fixed: htmlLoad restrictions when loading URL (allowedHTMLLoadURIs in CfgCommands or server.cfg)`

bux578:
Can't we add just add `http://ace3mod.com/version.html` to the `allowedHTMLLoadURIs` in `CfgCommands`?

PabstMirror:
#2428

PR #2332, opened 2015-08-31
Fix for #2310

jokoho48:
Fix for #2310

### DIFF
--- a/addons/optionsmenu/config.cpp
+++ b/addons/optionsmenu/config.cpp
@@ -32,3 +32,10 @@ class CfgAddons {
 class ACE_Extensions {
     extensions[] += {"ace_clipboard"};
 };
+
+
+class CfgCommands {
+    allowedHTMLLoadURIs = {
+        "http://ace3mod.com/version.html"
+    };
+};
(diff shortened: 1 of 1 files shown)


addons/optionsmenu/config.cpp
```diff
@@ -32,3 +32,10 @@ class CfgAddons {
 class ACE_Extensions {
     extensions[] += {"ace_clipboard"};
 };
+
+
+class CfgCommands {
+    allowedHTMLLoadURIs[] = {
```
jonpas: Shouldn't it be `+=` (if that works) so `allowedHTMLLoadURIs` from anything else wouldn't get deleted?

commy2: This doesn't influence the one in description.ext at all.
+= could actually be used here though. Would be a good idea. That way other addons could use it too.

jokoho48: ok i change it

ViperMaul:
Does this also work for Arma v1.50 as well as v1.52?

jokoho48:
@ViperMaul yes

Glowbal:
#2428

Issue #2284, opened 2015-08-26
You can stop carrying/dragging from both interaction menu and mousewheel menu


gienkov:
ACE3 Version: https://github.com/acemod/ACE3/commit/3db981a2c59250e90a39964ac29be2b16626a500

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *-*

**Description:**
*Like in title, you get action to drop crates under scrollwheel menu. I don't think this piece of code is longer needed since you can stop dragging under interaction menu.*

```
// add scrollwheel action to release object
_actionID = _unit getVariable [QGVAR(ReleaseActionID), -1];

if (_actionID != -1) then {
    _unit removeAction _actionID;
};

_actionID = _unit addAction [
    format ["<t color='#FF0000'>%1</t>", localize LSTRING(Drop)],
    QUOTE([ARR_2(_this select 0, (_this select 0) getVariable [ARR_2(QUOTE(QGVAR(draggedObject)),objNull)])] call FUNC(dropObject)),
    nil,
    20,
    false,
    true,
    "",
    QUOTE(!isNull (_this getVariable [ARR_2(QUOTE(QGVAR(draggedObject)),objNull)]))
];
```

**Steps to reproduce:**
* *Start carrying something.*
* *You get two options to drop it, in both mousewheel and interaction menu*

**Where did the issue occur?**
*"Multiplayer", "Singleplayer"*

**RPT log file:**
*-*

jonpas:
That was by design in AGM where you couldn't open interaction menu in some cases (object not directly in front of you). I believe this can now be removed since you can look at the object and use the menu normally.

alganthe:
Indeed it became redundant since you can just press the interaction key / scroll down to release the object.

PabstMirror:
When carrying a patient on your back there is no way to access his interaction point from first person.
When dragging a patient, I can't aim low enough to access the interaction point to release.

With track IR (or holding ALT/free look), or using different interaction menu options it is possible to overcome these problems, but with default options it is difficult/impossible.

Now I hate the `addAction` menu as much as anyone, so maybe we can come up with a cleaner solution, but the user needs a clear, easily accessible way to drop what they are carrying.

gienkov:
Keybind?

jonpas:
That would work, I suggest `R`. And next to that display the same icon as is used in attach?

gienkov:
R will reload your weapon. Maybe LCTRL+R ?

[...]

PR #2333, opened 2015-09-01
Add Keybinding for Drop in Dragging

jokoho48:
Add Keybinding for Drop in Dragging
based on https://github.com/acemod/ACE3/issues/2284#issuecomment-136550815 Idea

### DIFF
--- a/addons/dragging/XEH_clientInit.sqf
+++ b/addons/dragging/XEH_clientInit.sqf
@@ -32,13 +32,13 @@ if (isNil "ACE_maxWeightCarry") then {
         // End Dragging
         if !(isNull _draggedObject) then {
             [ACE_player, _draggedObject] call FUNC(dropObject);
-            _return = true
+            _return = true;
         };
         _carriedObject = (ACE_player getVariable [QGVAR(carriedObject),objNull]);
         // End Carry
         if (!isNull _carriedObject) then {
             [ACE_player, _carriedObject] call FUNC(dropObject_carry);
-            _return = true
+            _return = true;
         };
         _return
     },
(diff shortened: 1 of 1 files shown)


addons/dragging/XEH_clientInit.sqf
```diff
@@ -22,4 +22,23 @@ if (isNil "ACE_maxWeightCarry") then {
...
+        };
+        // End Carry
+        if (!isNull (ACE_player getVariable [QGVAR(carriedObject),objNull])) exitWith {
+            [ACE_player, ACE_player getVariable [QGVAR(carriedObject),objNull]] call FUNC(dropObject_carry)
+            true
+        };
+        false
+    },
+    { false },
+    [0, [false, false, false]]
```
MikeMatrix: This is an optional parameter according to [this](https://cbateam.github.io/CBA_A3/docs/files/keybinding/fnc_addKeybind-sqf.html)

MikeMatrix:
I like it.

addons/dragging/XEH_clientInit.sqf
jokoho48: and so i am sure that this have no assign

MikeMatrix: Default is `[-1, [false, false, false]]` anyways. Having this there is imo more confusing, since you don't omit setting the default itself, making it feel like this is actually a value to be considered.

gienkov:
:+1:

commy2:
I'd say we close this in favour of https://github.com/acemod/ACE3/pull/2598

commy2:
Drop can now be done with LMB which makes this obsolete.

PR #2338, opened 2015-09-01
Sitting improvements

commy2:


jonpas:
Tested it, works very nicely, great work! :+1: 

With this we no longer need the "restrict movement by GVAR(sitRotation)" in the PFH, I'll remove it once this gets merged and also cleanup some of it.

Another though, is it possible to hide the backpack like in vehicles? I think it would be preferred.

addons/sitting/CfgMoves.hpp
```diff
@@ -1,6 +1,9 @@
 // Enable visual head movement while free-looking
```
jonpas: That comment can be removed or new things added.

commy2:
head = "headDefault"; ... head turns in view direction
aimingBody = "aimingNo"; ... upper body doesn't turn in view direction (also leaning)
forceAim = 1; ... always in freelook mode
static = 1; ... whole body doesn't turn when rotating (rotating != viewing)

jonpas:
Awesome, thanks for the explanations!

commy2:
Feedback on the scroll wheel action? Any time someone wanted to sit for the first time, they were stuck, because they didn't know how to stand up again. This should be way more intuitive.

MikeMatrix:
@commy2 I think that's a good addition. Probably feels a lot more natural with the V shortcut working as well

MikeMatrix:
@commy2 In that sense, maybe it might be wiser to rename that action to `Get out` instead, so the user can intuitively make the connection that the V shortcut works on this?

commy2:
In my social experiment everyone (4 of 4) tried to press V first. The action is only there to enable the hotkey.

jonpas:
I like it! :+1:

alganthe:
:+1:

MikeMatrix:
@commy2 Alright. Although that sample size is not particularly huge :shipit:
We can always change it later if it isn't understood universally.

PR #2344, opened 2015-09-01
hide clock on map if no watch in inventory, #414

commy2:


addons/map/functions/fnc_onDrawMap.sqf
```diff
@@ -2,3 +2,8 @@
 #include "script_component.hpp"
 
 ((_this select 0) displayCtrl 1016) ctrlShow GVAR(mapShowCursorCoordinates);
+
+// hide time when no map in inventory
```
SzwedzikPL: ? :)

commy2: lol

PR #2350, opened 2015-09-02
Prevent Cargo opening if vehicle is locked

jonpas:
I think it makes sense, if a vehicle is locked you probably can't get into the trunk either. Automatically compatible with `vehiclelock` as well obviously.


bux578:
:+1:

Glowbal:
:+1:

Issue #1842, opened 2015-07-11
Random Behavior from Unconciousness function on AI (ace_medical_fnc_setUnconscious)


FatLurch:
It seems there might be a bug with the behavior of the Unconsciousness function:

I have the ACE Medical Settings Module placed, and I have the AI unconsciousness field set to enabled (not disabled, nor 50/50).

However, I seem to get getting random behavior from this function. I will put the unconsciousness function call into the init line of a unit ( [this, true] call ace_medical_fnc_setUnconscious; )  and load the scenario multiple times. Each time, there's a random chance of the unit dying instead of going unconscious. 

I also copied and pasted the unit about 20 times and ran the scenario again. The result is that a random number of the AI units wind up dead instead of unconscious.

Steps to reproduce:
1. Place ACE Medical Settings Module
2. Set "AI Unconsciousness" in Medical module to "enabled"
3. Place AI units
4. Insert the following code into the init line of each unit: [this, true] call ace_medical_fnc_setUnconscious;
5. Observe random number of AI dead instead of unconscious

PabstMirror:
Ref #148 - Events at mission start/loading 

For now, use this code in a unit's init box:
```
if (local this) then {
    nul = this spawn {
        sleep 3;
        [_this, true] call ace_medical_fnc_setUnconscious;
    };
};
```

The problem is init order:
Unit's init's run before modules are read, so the unconc function saw `GVAR(enableUnconsciousnessAI)` as "50/50" which is default and not the setting from the module.

I guess we might need something like this:
```
if (!EGVAR(common,settingsInitialized)) exitWith {
    [FUNC(setUnconscious), _this] call EFUNC(common,execNextFrame);
};
```
or
```
if (!EGVAR(common,settingsInitialized)) exitWith {
    EGVAR(common,runAtSettingsInitialized) pushBack [_this, DFUNC(setUnconscious)];
};

```
for all public functions?

FatLurch:
Gotcha, that makes perfect sense. Thank you for the quick feedback!

SilentSpike:
>```
if (!EGVAR(common,settingsInitialized)) exitWith {
    EGVAR(common,runAtSettingsInitialized) pushBack [_this, DFUNC(setUnconscious)];
};
```

Would be perfect in my opinion :+1:

Glowbal:
#2361

PR #2355, opened 2015-09-03
Setting init delay functions

PabstMirror:
From https://github.com/acemod/ACE3/issues/1842#issuecomment-120689590

Adds a way to run a function w/ parameters after settings are initialized.

We really need something like this for when we run code on objects that depends on settings, as the order isn't guaranteed in dedicated MP.

For example: @ViperMaul @Glowbal - the new medical FUNC(init) can run before settings are init
https://github.com/acemod/ACE3/blob/medical-adjustments/addons/medical/functions/fnc_init.sqf

@jonpas did something very similar in https://github.com/acemod/ACE3/pull/2187/files#diff-75f4a28dd014c00256df94dd2f6e929fR28

Helper function, otherwise you can just do something like:
```
if (!EGVAR(common,settingsInitFinished)) exitWith {
    EGVAR(common,runAtSettingsInitialized) pushBack [FUNC(init), _this];
};
```

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -150,6 +150,14 @@ call FUNC(checkFiles);
     //Event that settings are safe to use:
     ["SettingsInitialized", []] call FUNC(localEvent);
 
+    //Set init finished and run all delayed functions:
+    GVAR(settingsInitFinished) = true;
+    diag_log text format ["%1 delayed functions", (count GVAR(runAtSettingsInitialized))];
+    {
+        _x params ["_args", "_code"];
+        _args call _code;
+    } forEach GVAR(runAtSettingsInitialized);
+    
 }, 0, [false]] call CBA_fnc_addPerFrameHandler;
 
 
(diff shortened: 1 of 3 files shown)


jonpas:
:+1:

addons/common/XEH_postInit.sqf
```diff
@@ -150,6 +150,15 @@ call FUNC(checkFiles);
     //Event that settings are safe to use:
     ["SettingsInitialized", []] call FUNC(localEvent);
 
+    //Set init finished and run all delayed functions:
+    GVAR(settingsInitFinished) = true;
+    diag_log text format ["[ACE] %1 delayed functions running", (count GVAR(runAtSettingsInitialized))];
```
Glowbal: Should probably make use of the new logging macro(s) instead.

SilentSpike:
I like it, but I think we should condense the function name

SilentSpike:
All public functions that rely on settings should probably make use of this, since people can be running stuff from init fields.

commy2:
InitPostPost

SilentSpike:
PostSettings

Glowbal:
Need this to fix some issues with medical initalization.

PR #2364, opened 2015-09-05
Repair: Fix possible out of bounds setting

TheMagnetar:
Setting "engineerSetting_fullRepair" had an out of bounds default value.


jonpas:
Nice find :+1:

jokoho48:
looks very trivial

PR #2371, opened 2015-09-06
add Loudness coefficient based on Ammo Config

jokoho48:
This can be used for AT weapons with a Spotting Ammonition

### DIFF
--- a/addons/hearing/functions/fnc_firedNear.sqf
+++ b/addons/hearing/functions/fnc_firedNear.sqf
@@ -88,7 +88,7 @@ _caliber = call {
 };
 _loudness = (_caliber ^ 1.25 / 10) * (_initspeed / 1000) * _audibleFireCoef / 5;
 
-_loudnessCoef = getNumber (_ammoConfig >> QGVAR(loudnessCoef);
+_loudnessCoef = getNumber (_ammoConfig >> QGVAR(loudnessCoef));
 if (_loudnessCoef >= 0) then {
     _loudness = _loudness * _loudnessCoef;
 };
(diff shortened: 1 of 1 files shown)


addons/hearing/functions/fnc_firedNear.sqf
```diff
@@ -88,8 +87,15 @@ _caliber = call {
     if (_caliber <= 0) then { 6.5 } else { _caliber };
 };
 _loudness = (_caliber ^ 1.25 / 10) * (_initspeed / 1000) * _audibleFireCoef / 5;
+
+_loudnessCoef = getNumber (_ammoConfig >> QGVAR(loudnessCoef);
```
SilentSpike: Missing )

addons/hearing/functions/fnc_firedNear.sqf
```diff
@@ -88,8 +87,15 @@ _caliber = call {
     if (_caliber <= 0) then { 6.5 } else { _caliber };
 };
 _loudness = (_caliber ^ 1.25 / 10) * (_initspeed / 1000) * _audibleFireCoef / 5;
+
+_loudnessCoef = getNumber (_ammoConfig >> QGVAR(loudnessCoef));
```
esteldunedain: This approach has a big downside; if the loudnessCoef is not defined for a given weapon (e.g., third party) they will no longer produce hearing loss, as getNumber returns zero. 

An alternative would be to reverse the coefficient and call it `attenuationCoef`, so the default is zero (no attenuation) and the maximum is 1 (total attenuation).

esteldunedain:
Closed in favor of #4488

PR #2373, opened 2015-09-07
Move MicroDAGR 3D model from BFT branch to microdagr module

nsgundy:
BFT branch is not quite ready and probably won't release with MicroDAGR integration in the first phase. Besides, we are currently unsure what that's going to look like (MicroDAGR might stay its own module). So to get that nice model out there, it has now been moved into the `microdagr` module.


jokoho48:
@nsgundy i think the shadow LoD is a little bit to high poly but this is fixable

jokoho48:
is it ok if i reopen it from a new branch?

PR #2386, opened 2015-09-09
Fix that PPeffects throw a error on HC

jokoho48:
Fix that PPEffects Throw a eroor on HC


jokoho48:
Woops that was wrong

jonpas:
`invalid` label! :grin:

PR #2387, opened 2015-09-09
Fix that PPEffects Throw a error on HC

jokoho48:
Fix that PPEffects Throw a error on HC(this issue is PFH)


addons/common/functions/fnc_blurScreen.sqf
```diff
@@ -10,6 +10,8 @@
 
 #include "script_component.hpp"
 
+if !(hasInterface) exitWith {};
```
bux578: I don't know who came up with the idea to move the exclamation mark outside of the parentheses, but it looks ugly and wrong to me.

`if (!hasInterface) exitWith {};`

jokoho48: ok hmm yeah i can move it in if you are happy with that

jonpas:
This PR introduces a bug and a critical bug? XD

jokoho48:
na its a bug fix bug a critical bug

addons/huntir/functions/fnc_cam.sqf
```diff
@@ -13,6 +13,7 @@
  */
 #include "script_component.hpp"
 
+if !(hasInterface) exitWith {};
```
commy2: I doubt this ever runs on a dedi/hc. Seems redundant.

addons/nightvision/functions/fnc_changeNVGBrightness.sqf
```diff
@@ -16,6 +16,7 @@
  */
 #include "script_component.hpp"
 
+if !(hasInterface) exitWith {};
```
commy2: Same. This is called by a CBA key handler. Just don't add the keyhandler.

addons/goggles/functions/fnc_onEachFrame.sqf
```diff
@@ -14,6 +14,7 @@
  * Public: No
  */
 #include "script_component.hpp"
+if !(hasInterface) exitWith {};
```
commy2: Don't add the PFH on a dedi/hc. Seems better than to check something every frame that will never change.

addons/goggles/functions/fnc_applyGlassesEffect.sqf
```diff
@@ -16,6 +16,8 @@
  * Public: No
  */
 #include "script_component.hpp"
+if !(hasInterface) exitWith {};
```
commy2: >Don't add the PFH on a dedi/hc. Seems better than to check something every frame that will never change.

addons/goggles/XEH_postInit.sqf
```diff
@@ -15,7 +15,7 @@
  * Public: No
  */
 #include "script_component.hpp"
-if (!hasInterface) exitWith {};
+if !(hasInterface) exitWith {};
```
commy2: Doesn't his cover basically everything in goggles?

addons/gforces/functions/fnc_pfhUpdateGForces.sqf
```diff
@@ -13,8 +13,10 @@
  */
  #include "script_component.hpp"
 
+if !(hasInterface) exitWith {};
```
commy2: >Don't add the PFH on a dedi/hc. Seems better than to check something every frame that will never change.

commy2:
I think most of these could be done in a smarter way. Don't call functions / add pfhs in the first place instead of checking if they don't apply to dedi/hc.

addons/common/functions/fnc_blurScreen.sqf
MikeMatrix: I agree with @bux578 on this one. It makes absolutely no sense to have it outside.

SilentSpike: but...it makes more sense to have it outside! :laughing:

Glowbal:
Status?

jokoho48:
still need a review and testing but i think this work
and its importent for this version because this is a very critical bug

addons/spectator/functions/fnc_setSpectator.sqf
```diff
@@ -23,7 +23,7 @@
 params [["_set",true,[true]]];
 
 // Only clients can be spectators
-if !(hasInterface) exitWith {};
+if (!hasInterface) exitWith {};
```
SilentSpike: What's with all these dirty edits being made in this PR?

PR #2390, opened 2015-09-10
Added interaction to quickly pass a magazine

BaerMitUmlaut:
This adds an interaction that lets you quickly pass a magazine to another player (or AI unit). It automatically chooses the magazine that still has the most ammo, fits into the other units weapon and also still fits into his inventory.

I have not added the possibility to pass on ammunition for launchers, since this is more for when somebody runs out of ammo in combat, and I don't think you can't really quickly pass on a rocket. It would be easy to add though if so desired.


addons/interaction/functions/fnc_canPassMagazine.sqf
```diff
@@ -0,0 +1,31 @@
...
+
+#include "script_component.hpp"
+params ["_player", "_target", "_weapon"];
+private ["_compatibleMags", "_filterFunc", "_filteredMags"];
+
+_compatibleMags = getArray (configfile >> "CfgWeapons" >> _weapon >> "magazines");
+_filterFunc = compile format ["((_this select 0) in %1) && (!(_this select 2))", _compatibleMags];
+_filteredMags = [magazinesAmmoFull _player, _filterFunc] call EFUNC(common,filter);
+
+if (count _filteredMags > 0) then {
```
jokoho48: if (_filteredMags isEqualTo [])

bux578: 1. Has to be inverted 
2. Why? Count is fine imho

jokoho48: count is slower as isEqualTo

BaerMitUmlaut: I tested both with empty and non-empty arrays, with the added ! the performance difference is always around 0.0003ms. I think that's fine...

addons/interaction/functions/fnc_passMagazineLocal.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Public: No
+ */
+
+#include "script_component.hpp"
+params ["_unit", "_target", "_magazine", "_ammoCount"];
+
+_target addMagazine [_magazine, _ammoCount];
+if (_target == ACE_player) then {
+    [parseText format [localize LSTRING(PassMagazineHint), name _unit, getText (configFile >> "CfgMagazines" >> _magazine >> "displayName")]] call EFUNC(common,displayTextStructured);
```
commy2: [_unit] call EFUNC(common,getName)];

addons/interaction/functions/fnc_passMagazineLocal.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * [_unit, _player, "30Rnd_65x39_caseless_mag", 30] call ace_interaction_fnc_magToPassazine
+ *
+ * Public: No
+ */
+
+#include "script_component.hpp"
+params ["_unit", "_target", "_magazine", "_ammoCount"];
+
+_target addMagazine [_magazine, _ammoCount];
+if (_target == ACE_player) then {
```
commy2: ? makes no sense here

`addMagazine` implies that this is a locally run function. `hasInterface` or `EFUNC(common,isPlayer)`

addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -0,0 +1,49 @@
...
+_player removeMagazines (_magToPass select 0);
+{
+    if ((_x select 0) == (_magToPass select 0) && (_forEachIndex != _magToPassIndex)) then {
+        _player addMagazine [_x select 0, _x select 1];
+    };
+} foreach _filteredMags;
+
+_player playActionNow "PutDown";
+
+[[_player, _target, _magToPass select 0, _magToPass select 1], QUOTE(FUNC(passMagazineLocal)), _target] call EFUNC(common,execRemoteFnc);
```
commy2: `EFUNC(common,execRemoteFnc)` is deprecated. Define an event in postInit and call it with `EFUNC(common,targetEvent)`

addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -0,0 +1,49 @@
...
+_compatibleMags = getArray (configfile >> "CfgWeapons" >> _weapon >> "magazines");
+_filterFunc = compile format ["((_this select 0) in %1) && !(_this select 2)", _compatibleMags];
+
+_filteredMags = [magazinesAmmoFull _player, _filterFunc] call EFUNC(common,filter);
+
+//select magazine with most ammo
+_magToPass = _filteredMags select 0;
+_magToPassIndex = 0;
+{
+    if (((_x select 1) > (_magToPass select 1)) && ((_target canAddItemToUniform (_x select 0)) || (_target canAddItemToVest (_x select 0)) || (_target canAddItemToBackpack (_x select 0)))) then {
```
commy2: `canAdd` instead of 3x canAddX

addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+#include "script_component.hpp"
+params ["_player", "_target", "_weapon"];
+private ["_compatibleMags", "_filterFunc", "_filteredMags", "_magToPass", "_magToPassIndex"];
+
+_compatibleMags = getArray (configfile >> "CfgWeapons" >> _weapon >> "magazines");
+_filterFunc = compile format ["((_this select 0) in %1) && !(_this select 2)", _compatibleMags];
+
+_filteredMags = [magazinesAmmoFull _player, _filterFunc] call EFUNC(common,filter);
```
commy2: L23-26

No! There is absolute no reason to `compile` here.

`_filteredMags = [magazinesAmmoFull _player, {(_this select 0) in _compatibleMags && {!(_this select 2)}}] call EFUNC(common,filter);`

addons/interaction/functions/fnc_canPassMagazine.sqf
```diff
@@ -0,0 +1,31 @@
...
+    ({(_target canAddItemToUniform (_x select 0)) || (_target canAddItemToVest (_x select 0)) || (_target canAddItemToBackpack (_x select 0))} count _filteredMags) > 0
```
commy2: `canAdd` covers all three.

addons/interaction/functions/fnc_canPassMagazine.sqf
```diff
@@ -0,0 +1,31 @@
...
+ * Public: No
+ */
+
+#include "script_component.hpp"
+params ["_player", "_target", "_weapon"];
+private ["_compatibleMags", "_filterFunc", "_filteredMags"];
+
+_compatibleMags = getArray (configfile >> "CfgWeapons" >> _weapon >> "magazines");
+_filterFunc = compile format ["((_this select 0) in %1) && (!(_this select 2))", _compatibleMags];
+_filteredMags = [magazinesAmmoFull _player, _filterFunc] call EFUNC(common,filter);
```
commy2: `_filteredMags = [magazinesAmmoFull _player, {(_this select 0) in _compatibleMags && {!(_this select 2)}}] call EFUNC(common,filter);`

addons/interaction/functions/fnc_passMagazine.sqf
commy2: Why is this done here again anyway? The condition is always checked on the same frame before executing the statement in the interaction menu.

addons/interaction/functions/fnc_canPassMagazine.sqf
commy2: I love how you guys discuss `count == 0` vs `isEqualTo []` and copletely ignore the `compile`.

addons/interaction/functions/fnc_passMagazineLocal.sqf
BaerMitUmlaut: If I'd skip this check the hint would be displayed in singleplayer.

addons/interaction/functions/fnc_passMagazine.sqf
BaerMitUmlaut: I assumed I couldn't use the local variables, my bad.

You not only need to check if the player has a compatible magazine for the other unit, you also need to check if he passes one of those compatible magazines. Say an ammo carrier carries both 30rnd 5.56 mags and also 20rnd 7.62 mags for a DMR. If I'd skip that check, he'd always pass the DMR a 5.56 magazine because it has more rounds.

addons/interaction/functions/fnc_passMagazineLocal.sqf
Glowbal: Use the event version of the `displayStructuredText` and pass a target. It will only display if the target is `ACE_player`. This might be the best solution for displaying the hint.
```sqf
["displayTextStructured", [_target], ["string or format syntax", 1.5, _target]] call EFUNC(common,targetEvent);
```

addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -0,0 +1,49 @@
...
+
+//select magazine with most ammo
+_magToPass = _filteredMags select 0;
```
Glowbal: You can use `params` for this

addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -0,0 +1,49 @@
...
+{
+    if (((_x select 1) > (_magToPass select 1)) && ((_target canAddItemToUniform (_x select 0)) || (_target canAddItemToVest (_x select 0)) || (_target canAddItemToBackpack (_x select 0)))) then {
+        _magToPass = _x;
+        _magToPassIndex = _forEachIndex;
+    };
+} foreach _filteredMags;
+
+//remove all magazines and add them again, except the one to be passed
+//needed because of missing commands, see http://feedback.arma3.com/view.php?id=12782
+_player removeMagazines (_magToPass select 0);
```
Glowbal: I'd store the `_magToPass select 0` into a variable, as you use it in the `forEach` below. No point in selecting it multiple times. Also, `params`.

addons/interaction/functions/fnc_passMagazine.sqf
Glowbal: Also I don't see any handling for when there is no magazine to pass.

addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -0,0 +1,49 @@
...
+{
+    if ((_x select 0) == (_magToPass select 0) && (_forEachIndex != _magToPassIndex)) then {
```
Glowbal: Use `params` to exact the values from `_x` instead of `select`

SilentSpike:
I'm in two minds about this addition.

Passing a magazine to someone in close proximity can be done already with the inventory and being honest I can use the inventory much quicker than I can use the interaction menu on a person :stuck_out_tongue:

bux578:
> I'm in two minds about this addition.

agree


on the other hand IRL you could "throw" magazines much further

BaerMitUmlaut:
If neither you nor the other person have a backpack and you're in tall grass, dropping and picking up magazines becomes a pita. Without backpacks I'm pretty sure this would be way faster.

commy2:
>I can use the inventory much quicker than I can use the interaction menu on a person

And I can't use Arma 3's inventory system at all on remote units in MP. Half of the time nothing happens and I have to close and re-open the menu. Sometimes I have to use drag and drop, other times it only works with right-clicking. Occasionally the item disappears (especially magazines) or duplicates pop up.

With `addMagazine` and `removeMagazine` we at least know that it works if scripted correctly.

SilentSpike:
Both points seem reasonable :+1:

jonpas:
Make it optional then.

SilentSpike:
Isn't it inherently optional? :P

BaerMitUmlaut:
No need to make this optional. If you think the other way is faster, don't use the interaction menu.

jonpas:
It still clogs up the interaction menu though, not really a problem, I just like it clean. :grin:

addons/interaction/functions/fnc_canPassMagazine.sqf
```diff
@@ -0,0 +1,33 @@
...
+_filteredMags = [magazinesAmmoFull _player, {
+    params ["_className", "", "_loaded"];
+    _className in _compatibleMags && !_loaded
+}] call EFUNC(common,filter);
+
+if !(_filteredMags isEqualTo []) then {
+    ({_target canAdd (_x select 0)} count _filteredMags) > 0
+} else {
+    false
+};
```
commy2: could be simplified to:
`_filteredMags isEqualTo [] && {{_target canAdd (_x select 0)} count _filteredMags > 0}`

addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -0,0 +1,57 @@
...
+    };
+} foreach _filteredMags;
+
+_player playActionNow "PutDown";
+
+_target addMagazine [_magToPassClassName, _magToPassAmmoCount];
+
+_playerName = [_player] call EFUNC(common,getName);
+_magToPassDisplayName = getText (configFile >> "CfgMagazines" >> _magToPassClassName >> "displayName");
+["displayTextStructured", [_target], [[LSTRING(PassMagazineHint), _playerName, _magToPassDisplayName], 1.5, _target]] call EFUNC(common,targetEvent);
```
commy2: What happens if I pass this to a local AI? Reminder for me to test this.

addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -0,0 +1,57 @@
...
+    params ["_className", "", "_loaded"];
+    _className in _compatibleMags && !_loaded
+}] call EFUNC(common,filter);
+
+//select magazine with most ammo
+_magToPass = _filteredMags select 0;
+_magToPassIndex = 0;
+{
+    _x params ["_className", "_ammoCount"];
+    if ((_ammoCount > (_magToPass select 1)) && (_target canAdd _className)) then {
```
commy2: Really minor thing, but if you `exitWith` here, you don't have to iterat through the rest of the array.

addons/interaction/functions/fnc_passMagazine.sqf
BaerMitUmlaut: The AI receives the mag and the hint does not appear.

addons/interaction/functions/fnc_canPassMagazine.sqf
```diff
@@ -0,0 +1,33 @@
...
+_filteredMags = [magazinesAmmoFull _player, {
+    params ["_className", "", "_loaded"];
+    _className in _compatibleMags && !_loaded
+}] call EFUNC(common,filter);
+
+if (!(_filteredMags isEqualTo []) && {{_target canAdd (_x select 0)} count _filteredMags > 0}) then {
+    true
+} else {
+    false
+};
```
commy2: `if (BOOL) then {true} else {false}`
: - )

jokoho48: BOOL :D

BaerMitUmlaut: duh

BaerMitUmlaut:
How about this?

BaerMitUmlaut:
Merge conflicts have been resolved, also fixed an error in a header. This is no more WIP unless you see something wrong with it.

addons/interaction/CfgVehicles.hpp
```diff
@@ -35,6 +35,32 @@ class CfgVehicles {
                 icon = "\a3\ui_f\data\IGUI\Cfg\Actions\eject_ca.paa";
                 selection = "pelvis";
 
+                class ACE_PassMagazine {
+                    displayName = CSTRING(PassMagazine);
+                    condition = QUOTE([ARR_3(_player,_target,primaryWeapon _target)] call FUNC(canPassMagazine) || [ARR_3(_player,_target,handgunWeapon _target)] call FUNC(canPassMagazine));
```
PabstMirror: Parent condition can be empty string or just `true`.  
If there are no children, it won't show because statement is `{}`
Will save checking the conditions twice.

PabstMirror:
I guess I wouldn't mind this simplified down to `currentWeapon` instead of the tree, but that's just me.
Either way it looks good.

addons/interaction/CfgVehicles.hpp
BaerMitUmlaut: Fixed

Glowbal:
Thanks for the change @BaerMitUmlaut :+1:

PR #2393, opened 2015-09-11
Remove redundant rotation boundaries in Sitting

jonpas:
The PFH is only there to automatically stand up the player if a chair moves/falls, rotating is now disabled via `CfgMoves` thanks to @commy2


SilentSpike:
Seems pretty trivial, I notice the auto stand up thing doesn't account for if their chair is deleted though (it also seems kinda wasteful to have an entire function just to check if the position has changed, wouldn't `isEqualTo` suffice?).

jonpas:
It does account for deletion actually, doing `getPosASL` on an inexisting object returns `[0, 0, 0]`, just tested.

I'll look into `isEqualTo` replacement now.

SilentSpike:
What if the chair was at position [0,0,0] :wink:

Gotta be explicit

SilentSpike:
Also, it might make more sense to check the vectorDir of the chair rather than the position. I know it's probably very edge case, but there could be uses for a moving chair.

jonpas:
> What if the chair was at position [0,0,0]

That's such an edge case it can't happen in theory, there will always be some decimal numbers, if someone puts it at `[0,0,0]` explicitly... I don't think it's warrants a check really.

> Also, it might make more sense to check the vectorDir of the chair rather than the position. I know it's probably very edge case, but there could be uses for a moving chair.

Well yes and no, from my experimentation a chair can get moved in such ways that it doesn't change direction (weird Arma if you ask me), I managed to do that by throwing a grenade next to it.

SilentSpike:
>Well yes and no, from my experimentation a chair can get moved in such ways that it doesn't even change direction (weird Arma if you ask me), I managed to do that by throwing a grenade next to it.

lol, that's pretty halarious

I still think there should be an `isNull` check, again, explicit > implicit

jonpas:
Trust me, I was shocked, couldn't even think about laughing...

There, added isNull check.

SilentSpike:
:+1: the isEqualTo change doesn't cause any issues does it?

I've never actually ran a getPos command on an object that is stationary to check for changes

jonpas:
It works perfectly fine, since I want even the slightest move to get recorded. Surprisingly even 0.0000001 changes don't matter which is Arma's real-time simulation weirdness.

SilentSpike:
Cool, merging

jonpas:
This was like live PR support. :grin:

PR #2405, opened 2015-09-12
Captive on respawn fix

### LINKED ISSUE #1849: Captives persist after death
ACE3 Version: 3.1.1

Happens on Stable, Dev and RC Builds of ArmA

**Mods:**
* @cba_a3
* @ace3

If a player who is restrained by zip ties dies or respawns they remain restrained and can not be unrestrained. Requires them to re-log onto server to fix the problem

**Steps to reproduce:**
1) Zip tie another player
2) Kill the zip-tied player or have them respawn
When the zip-tied player respawns the are permanently zip tied until they log off the server

Multiplayer Bug

*edit* when my friends get on tomorrow I can record the problem if that helps*

SilentSpike:
I believe this should fix #1849 

Could use some more testing though if anyone wants to confirm it (I wasn't 100% sure what to look for)


SilentSpike:
Whoops, this was meant to be into release...damnit it github why is there no way to change the target branch.

PR #2406, opened 2015-09-12
Reset captive status on respawn for reliability

### LINKED ISSUE #1849: Captives persist after death

SilentSpike:
Previously we were resetting on death, but this caused problems when the respawnDelay was 0 as the dead unit wasn't actually the player.

To account for "group" and "side" respawn types where you change unit on respawn and could potentially become a captive unit, I am explicity checking the mission respawn type before resetting variables.

I believe this should fix #1849 

Could use more testing as I wasn't entirely sure what to look for


addons/captives/functions/fnc_handleRespawn.sqf
```diff
@@ -16,24 +16,40 @@
  */
 #include "script_component.hpp"
 
+systemChat str(_this);
```
commy2: reminder

Glowbal:
Status?

PabstMirror:
:+1:

commy2:
Fixes the respawn bug on the normal respawn. No idea how the group respawn is supposed to work.
:+1:

PR #2409, opened 2015-09-12
added the addons folder to gitignore

alganthe:
**THIS IS FOR gh-pages ONLY**
Removes the need to delete the folder every time you checkout from master (or any branch tracking it)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 _site/
 _theme_packages/
 _drafts/
+addons/
 
 Thumbs.db
 .DS_Store


bux578:
image

^ that's the way I'm doing it

alganthe:
would be simpler to just add the proper gitignores no ?

bux578:
simpler when changing, but that is just way faster for me. And "node_modules" (that is when you start to tweak the css/js files and build them) with its file name  lengths > 250 chars won't let you easily switch branches (git can't delete those files, because the names are too long).

alganthe:
Make sense, closing.

PR #2430, opened 2015-09-14
microDagr - Add ability to access draw event for minimap

PabstMirror:


Glowbal:
Sounds useful. Any specific use case you want to use this for? Just curious.

PabstMirror:
We use custom makers, drawn via `drawIcon` for our simple BFT system.
So I need this to draw them on the microDagr.

Issue #2414, opened 2015-09-12
"Keep Cursor Centered" causes interaction menu to fly off-screen


Zoxjib:
ACE3 Version: 3.3.0.1 RC - occurred all the way back since 3.0.0.1. Tested on stable 1.50

**Mods:**
- `@CBA_A3`
- `@ace`

**Placed ACE3 Modules:**
Default ACE3 testing mission, I didn't check.

**Description:**
When using the "Keep Cursor Centered" option in the ACE menu, the interaction menu flies off the screen.

**Steps to reproduce:**
- Enable "Keep Cursor Centered"
- Interact with a unit

**Where did the issue occur?**
Multiplayer listenserver

**RPT log file:**
http://pastebin.com/usvjvZpb

**Debug:**
http://pastebin.com/cCMBctxx

**Video:**
https://www.youtube.com/watch?v=rBIooeiAp-o

PabstMirror:
The video option "sampling" effects `setMousePosition`, and I don't think I can detech when it's changed.

EG put sampling at 50%, and `setMousePosition [0.5, 0.5]` will put it at [0.25, 0.25].

I can't see your mouse cursor in the video, but this may have been a cause.

PabstMirror:
Made BIS Issue: http://feedback.arma3.com/view.php?id=25874

Really isn't anything we can do to fix.

Mike-MF:
Closing due to age, inactivity or already solved. May be re-opened if the issue is still present, even better just open a new up-to-date Issue.

PR #2433, opened 2015-09-15
Deadzone for KeepCursorCentered

PabstMirror:
#2414

@BullHorn7 - I believe this should fix the issue, if you have a chance I would love if you could test this.  The deadzone value I picked seem good for me, but too high and it becomes jerky, too low and it doesn't fix the issue.

Also, I think the issue is you are playing in an odd resolution. Could you check your ingame resolution or check `getResolution`.


PabstMirror:
Won't fix the issue. Need http://feedback.arma3.com/view.php?id=25874

PR #2438, opened 2015-09-15
Add Setting for toolkit requirement for wheel repair

PabstMirror:
Add ace setting to allow replace/remove wheels without a toolkit.

### DIFF
--- a/addons/repair/stringtable.xml
+++ b/addons/repair/stringtable.xml
@@ -902,7 +902,7 @@
             <Portuguese>Número de partes sobressalentes.</Portuguese>
         </Key>
         <Key ID="STR_ACE_Repair_wheelRepairRequiredItems_name">
-            <English>Req Items for wheel repair</English>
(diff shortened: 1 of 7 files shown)


alganthe:
:+1:

Glowbal:
Very nice. I like it :+1:

addons/repair/stringtable.xml
```diff
@@ -901,5 +901,11 @@
             <Polish>Ilość wybranych części zamiennych.</Polish>
             <Portuguese>Número de partes sobressalentes.</Portuguese>
         </Key>
+        <Key ID="STR_ACE_Repair_wheelRepairRequiredItems_name">
+            <English>Req Items for wheel repair</English>
```
SilentSpike: I'd maybe make this something like "Wheel repair requirements", cleaner and actually 1 character less if space is an issue

commy2:
:+1:

Issue #2437, opened 2015-09-15
ACE3 RHS 0.3.9 No Interaction On Russian Units


moonmanmove:
ACE3 Version: 3.2.1 (stable0

Mods:
@CBA_A3
@ace
@rhs 0.3.9

Placed ACE3 Modules:
none

Description:
No self interaction on russian units, cannot interact with them either.

Steps to reproduce:
place down unit from russian opfor faction
try to self interact no interaction available

Where did the issue occur?
Singleplayer

RPT log file:
http://pastebin.com/REHewrYY

bux578:
Although we appreciate taking your time to submit an issue, please edit your ticket to follow the guidelines: http://ace3mod.com/wiki/user/how-to-report-an-issue.html

moonmanmove:
rpt up

PabstMirror:
Could you try moving the `cba_enable_auto_xeh.pbo` from CBA's optional folder to addons?

moonmanmove:
@PabstMirror 
yep, that fixes it, do you want the RPT from that?

jokoho48:
no its a mod fault we cant not fix this

alganthe:
Mod's fault check the ACE3 team official statement here [click me](http://ace3mod.com/wiki/user/information-center.html#1-3-compatibility) about hose issues.

moonmanmove:
http://i.imgur.com/SAV8SqI.png
well there's RHS response on the issue, is it only fixable on their end?

PabstMirror:
```
	class SoldierGB: CAManBase {
		class EventHandlers: EventHandlers {
			init = "this setVariable [""BIS_fnc_unitHeadgear"",false];";
		};
	};
```

So this will break both RHS and AAF (`class I_Soldier_base_F: SoldierGB`) units!

In the past they've always done 
`init = "_this call SLX_XEH_EH_Init;_this spawn RHS_fnc_t80_init;....`
which kept CBA compatibility and allowed them to use their own EventHandlers without requiring CBA.

I have no idea why they've stopped offering backwards compatibility, I hope it's just a mistake that they will fix.

In any case, we really need to stress that cba_enable_auto_xeh.pbo should be added.

SilentSpike:
Yes, only they can fix this. the auto_xeh pbo is a good workaround though :+1:

DerekSauer:
Look like the RHS team has corrected this on their end:

[...]

Issue #1171, opened 2015-05-14
The land of the mysteriously missing interact menus


jaynus:
RPT: http://pastebin.com/VvhUzYUj
Another RPT: http://pastebin.com/1NiZ57dE

Symptoms:
- This is consistently happening across all players @ UO with the modpack. It may/may not be mod pack related, but the conflict seems common enough to warrant investigation. 
- It is random, reconnects and/or reslots fix it. This means a preinit/postinit re-run fixes it.
- Units in tests are RHS units
- It cannot be file corruption or CBA issues, as the server is signed/modpacked/verified 

Thoughts on possible causes (checkbox is ruled out):
- [x] mods using the old CBA method are corrupting the keybinds (unlikely, ACRE still works)
- [] I tested excessive XEH times; that is *not* the cuplrit.
- [ ] XEH Initialization order
- [x] Somehow the objects are not tracking their interactions correctly
- [x] Interaction cache is getting corrupted or deleted 
- [x] RHS unit related issue for XEH initializers for the actions
- [ ] ????????? @esteldunedain  help me out here bud

esteldunedain:
My best bet is still XEH_postInit running on sched environment and getting cut off while compiling the menu (which is quite big). I'll try reusing the same menu for all men to cut down init time

jaynus:
I ran a test with 99999 loop XEHs. They took 10+ minutes to load but interact still worked. Is XEH scheduled only on JIP?

esteldunedain:
> Is XEH scheduled only on JIP?

As far as I understand cba code, it's only scheduled if respawn is set on the mission. There might be other conditions too.

jaynus:
Okay. I'll run all the cases and try.

alganthe:
the module backpacks is missing the whole interaction system it needs to be added.

jaynus:
That has nothing to do with this bug. File a seperate one please.

jaynus:
This is appearing to be related with CBA XEH becoming scheduled on CAManBase, and then bypassing hte player check if > 10 seconds

jaynus:
http://bit.ly/1HcwXLX RC7 build of CBA with possible fix.

Overlord00:
United Operations is still experiencing this issue.

SilentSpike:
Can you be any more specific on that?

Overlord00:
[...]

PR #2439, opened 2015-09-15
Bad xeh check

PabstMirror:
Ref #2437 (not our fault so it's not a fix)

At mission start checks for interaction menu compiles on all units and vehicles and spits out a message to users that there is a problem.

Designed to help users identify when mods break XEH (with RHS update I see this happening even more).
May help with #1171

### DIFF
--- a/addons/interact_menu/XEH_clientInit.sqf
+++ b/addons/interact_menu/XEH_clientInit.sqf
@@ -114,7 +114,7 @@ addMissionEventHandler ["Draw3D", DFUNC(render)];
         if ((isNil (format [QGVAR(Act_%1), typeOf _x])) || {isNil (format [QGVAR(SelfAct_%1), typeOf _x])}) then {
             if (!((typeOf _x) in _badClassnames)) then {
                 _badClassnames pushBack (typeOf _x);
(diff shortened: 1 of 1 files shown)


addons/interact_menu/XEH_clientInit.sqf
```diff
@@ -104,3 +104,25 @@ addMissionEventHandler ["Draw3D", DFUNC(render)];
...
+    private ["_badClassnames"];
+    _badClassnames = [];
+    {
+        if ((isNil (format [QGVAR(Act_%1), typeOf _x])) || {isNil (format [QGVAR(SelfAct_%1), typeOf _x])}) then {
+            if (!((typeOf _x) in _badClassnames)) then {
+                _badClassnames pushBack (typeOf _x);
+                ACE_LOGERROR_3("Compile checks bad for (classname: %1)(addon: %2) %3", (typeOf _x), (unitAddons (typeOf _x)), _x);
+            };
+        };
+    } forEach (allUnits + vehicles);
```
commy2: Extremly minor, but

`allUnits + allDeadMen + vehicles`

would include dead soldiers which are not covered by `allUnits` for god knows why.

Issue #2445, opened 2015-09-15
Cargo items float 100m above the vehicles


alganthe:
ACE3 Version: 3.3.0 stable

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
debug here -> [pastebin](http://pastebin.com/wScGB6LY)

**Description:**
Cargo items float 100m above the vehicles they are loaded in
screenshot here -> [puush](http://puu.sh/kcFmF/b796dfcf03.jpg)

**Steps to reproduce:**
- Place a rifleman
- Place a vehicle
- Press preview
- Look up

**Where did the issue occur?**
Editor

**RPT log file:**
RPT -> [pastebin](http://pastebin.com/L2iyWf77)

PabstMirror:
Should be fixed by e254af99640704e3ca8af141bed66573d0448a28

alganthe:
issue fixed

PR #2448, opened 2015-09-16
posible Fix  for #2445

jokoho48:


commy2:
Uhm. The issue was that the items where visible due to the event system. Now they are lying around at the bottom of the ocean. Also, `allowDamage` doesn't work on remote units (i.e. non-zeus ammo boxes on dedicated).
Also the damage is never re-enabled.

PR #2451, opened 2015-09-16
Add new zeus addObjects setting to settings module

alganthe:
The module part in CfgVehicles.hpp was missing and thus the option to turn it  on / off wasn't here.

### DIFF
--- a/addons/zeus/ACE_Settings.hpp
+++ b/addons/zeus/ACE_Settings.hpp
@@ -21,9 +21,7 @@ class ACE_Settings {
         values[] = {"$STR_A3_OPTIONS_DISABLED", CSTRING(revealMines_partial), CSTRING(revealMines_full)};
     };
     class GVAR(autoAddObjects) {
-        displayName = CSTRING(AddObjectsToCurator);
-        description = CSTRING(AddObjectsToCurator_desc);
-        value = 0;
         typeName = "BOOL";
+        value = 0;
     };
 };
--- a/addons/zeus/ACE_Settings.hpp
+++ b/addons/zeus/ACE_Settings.hpp
@@ -23,5 +23,7 @@ class ACE_Settings {
     class GVAR(autoAddObjects) {
         typeName = "BOOL";
         value = 0;
+        displayName = CSTRING(AddObjectsToCurator);
+        description = CSTRING(AddObjectsToCurator_desc);
     };
 };
--- a/addons/zeus/CfgVehicles.hpp
+++ b/addons/zeus/CfgVehicles.hpp
@@ -70,6 +70,12 @@ class CfgVehicles {
                     };
                 };
             };
+            class GVAR(autoAddObjects) {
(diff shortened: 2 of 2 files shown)


addons/zeus/ACE_Settings.hpp
```diff
@@ -21,9 +21,7 @@ class ACE_Settings {
         values[] = {"$STR_A3_OPTIONS_DISABLED", CSTRING(revealMines_partial), CSTRING(revealMines_full)};
     };
     class GVAR(autoAddObjects) {
-        displayName = CSTRING(AddObjectsToCurator);
```
Glowbal: Why are these removed?

alganthe:
The other settings don't have those lines (reveal mines and ascension) thus i though it wasn't necessary, should i add it back?

Glowbal:
Yes they should be added back. It is for the Config Export menu.

alganthe:
adding all of them now, note that some other modules are also missing them

SilentSpike:
Will have to give this one a think, this setting might not be suitable for the module.

Potentially we could retroactively add all existing objects once settings are initialized. However this wouldn't work out if it was enabled by default and the module tried to disable it.

jonpas:
> Potentially we could retroactively add all existing objects once settings are initialized. However this wouldn't work out if it was enabled by default and the module tried to disable it.

`runAtSettingsInitialized`

Glowbal:
This PR adds unrelated code as well. That should probably be in a separate PR.

Glowbal:
If we change this to use `runAtSettingsInitialized` we can merge this, I think.

SilentSpike:
My memory needs a refresher: `runAtSettingsInitialized` = run immediately if they have initialized, but wait if they haven't? (The naming is a little ambiguous)

jonpas:
@SilentSpike That's up to the function, `runAtSettingsInitialized` simply adds the function to an array for later execution, so you can relay on eg. module settings in the function and based on the setting do or don't do something.

It is not based on specific setting, the functions which are waiting for settings to initialize using this will always run, it's up to the function to exit if it shouldn't do something based on settings.

SilentSpike:
Wait, what? :confused:

I think that's a yes :laughing:

jonpas:
I think I misunderstood you a bit. :grin: 

If I understand what you are asking now, then I don't know really, the correct logic for `runAtSettingsInitialized` is https://github.com/acemod/ACE3/blob/master/addons/repair/functions/fnc_addSpareParts.sqf#L29-L31
Which means if settings are already initialized they shouldn't be added to `runAtSettingsInitialized` at all, I guess.

PabstMirror:
variable isn't read in the module sqf:
https://github.com/alganthe/ACE3ganthe/blob/zeus_fix/addons/zeus/functions/fnc_moduleZeusSettings.sqf

Glowbal:
I will take over and finish this up

Glowbal:
Closing in favour of #3768 that continues the work done in this PR.

PR #2463, opened 2015-09-16
Cleanup Nightvision

PabstMirror:


addons/nightvision/functions/fnc_blending.sqf
```diff
@@ -23,25 +23,19 @@
...
 
-private ["_vehicle", "_weapon", "_ammo", "_magazine", "_player"];
+params ["_vehicle", "_weapon", "", "", "_ammo", "_magazine"];
 
-_vehicle = _this select 0;
-_weapon = _this select 1;
-_ammo = _this select 4;
-_magazine = _this select 5;
-
-_player = ACE_player;
```
SilentSpike: This is actually quicker by the way, due to the whole global variable indexing stuff. Then again, meh :wink:

commy2: I personally prefer when a function uses `_unit` throughout the file and defines it either on the top of the file with `_unit = ACE_player` or preferably inherits it from wherever it was called from.

PR #2464, opened 2015-09-16
increase trucks cargo space

commy2:
also fixes a ubc error in the current release build that completely wrecks the hemets.


commy2:
wrong target branch

PR #2468, opened 2015-09-16
paradrop crates from helicopter or plane cargo

commy2:


PabstMirror:
this sounds really cool

Are we planning to add actions for pilots/passangers?

commy2:
No idea yet. I guess for the pilot? Or maybe depending on the type of heli / plane.

sargken:
Crew chief

addons/cargo/functions/fnc_paradropItem.sqf
```diff
@@ -0,0 +1,85 @@
...
+
+_item setPosASL _position;
+
+["hideObjectGlobal", [_item, false]] call EFUNC(common,serverEvent);
+
+// open parachute and ir light effect
+[{
+    params ["_item"];
+
+    if (isNull _item || {getPos _item select 2 < 1}) exitWith {};
```
jokoho48: if this is true the code below never gets exectuted are you shure you want to do this?

commy2: Yes. I don't want to open a parachute if the crate already landed, e.g. when droping from a very low altitude.

commy2:
- [ ] CBA events

PabstMirror:
@commy2 - integrated this into cargo

Interact with vehicle while mounted.
Shows normal cargo menu (with button name changed)
Requires level flight at a reasonable altitude when unloading.

- [ ] Who should have access to this? Anyone except pilot?
- [ ] Do we want an ace setting for this?

commy2:
I'd say pilot and copilot should be able to access it. It would be too exploitable if the cargo soldiers could do it.
No ACE setting. It already is optional. You don't have to use the action.

bux578:
The Taru has a designated "Loadmaster" seat.

jonpas:
Add a config entry for turrets (like fcs)?

esteldunedain:
I'd say pilot + copilot by default + optional config entry to specify other seats that should be able to do it.

addons/cargo/stringtable.xml
```diff
@@ -201,5 +201,13 @@
             <Portuguese>Tamanho do objeto</Portuguese>
             <Russian>Размер объекта</Russian>
         </Key>
+        <Key ID="STR_ACE_Cargo_paradropButton">
+            <English>Paradrop</English>
+            <German>Abspringen</German>
+        </Key>
+        <Key ID="STR_ACE_Cargo_unlevelFlightWarning">
+            <English>Unlevel Flight</English>
+            <German>Ungleiche Flug</German>
```
bux578: `Schieflage`

addons/cargo/stringtable.xml
```diff
@@ -201,5 +201,13 @@
             <Portuguese>Tamanho do objeto</Portuguese>
             <Russian>Размер объекта</Russian>
         </Key>
+        <Key ID="STR_ACE_Cargo_paradropButton">
+            <English>Paradrop</English>
```
bux578: Airdrop instead of Paradrop?
https://en.wikipedia.org/wiki/Airdrop

addons/cargo/stringtable.xml
```diff
@@ -201,5 +201,13 @@
...
+            <German>Abspringen</German>
```
bux578: I believe the military term is `Türlast` (which translates to "door load").

addons/cargo/stringtable.xml
```diff
@@ -202,12 +202,12 @@
...
         <Key ID="STR_ACE_Cargo_paradropButton">
-            <English>Paradrop</English>
-            <German>Abspringen</German>
+            <English>Airdrop</English>
+            <German>Türlast</German>
         </Key>
         <Key ID="STR_ACE_Cargo_unlevelFlightWarning">
             <English>Unlevel Flight</English>
-            <German>Ungleiche Flug</German>
+            <German>Schieflage Flug</German>
```
bux578: Sorry, just "Schieflage"

PR #2471, opened 2015-09-16
optional pbo to hide select weapon actions

commy2:
hides:
Weapon P07
Weapon MX
Weapon NLAW

#414


gienkov:
I like this so much

PabstMirror:
Not an error, but need to make sure to run on server as well or checkFiles will kick
image

At some point I was thinking of putting magazine selection on the ace menu which would allow removing `SwitchMagazine` as well from action menu.

Maybe instead of noactionmenu we could call it noWeaponSelectActions?

commy2:
Seems like we need an additional entry in CfgPatches and checkFiles.

>Maybe instead of noactionmenu we could call it noWeaponSelectActions?

Though about that, but it wouldn't fit anymore should we expand it to, for example, include `SwitchMagazine`.

PR #2480, opened 2015-09-17
tweak doAnimation switchMove to do no akward camera jumps whenever possible

commy2:


addons/common/functions/fnc_doAnimation.sqf
```diff
@@ -62,8 +64,19 @@ switch (_priority) do {
         };
     };
     case 2 : {
-        // Execute on all machines. SwitchMove has local effects.
-        [_unit, format ["{_this switchMove '%1'}", _animation]] call FUNC(execRemoteFnc);
+        // try playMoveNow first
+        if (_unit == vehicle _unit) then {
+            [_unit, format ["{_this playMoveNow '%1'}", _animation], _unit] call FUNC(execRemoteFnc);
```
Glowbal: Earlier in this function it states that `playMoveNow` only needs to be broadcasted by to all clients when not in a veihcle.

commy2: `execRemoteFnc` isn't broadcasted here, because the third parameter, `_unit`, is local.
It is used here to let the function work on remote units. No third parameter defaults to 0, which is similar to global event.
This will be replaced with a proper event system in the commonCleanup branch, which is why I need this before finishing it.

Glowbal: Oh right. I got confused. :+1:

Issue #2488, opened 2015-09-17
Wheel added to cargo on player connect.


JasperRab:
ACE3 Version: 3.3.1
Stable

Mods:
@CBA_A3
@ace
@RHS 
@3CB_Vehicles

Placed ACE3 Modules:
Interaction System
Medical Settings
Advanced Medical Settings
Wind Deflection
Advanced Ballistics
Respawn System
Weather
Hearing

Description:
A wheel is added to the cargo on player connect, While testing whit my friend we found out we had 4 wheels in the vehicles on our public server. 1 default 1 HeadlessClient 1 my friend 1 me I think.

Steps to reproduce:
Have a mission with multiple playable units and have everyone join at the same time, check the cargo space of a vehicle. It might be completely full or has the amount of players + 1 in side with wheels. Also happens with JIP.

Where did the issue occur?
Multiplayer

RPT log file:
Not needed

LecksC:
This fix stops wheels spawning on dedicated server, at least when the vehicles are created on clients. I'm not certain, but I think serverInit may not work like this for Init EventHandlers in XEH (I can't see this pattern in docs). 
I think a better solution may be to revert https://github.com/acemod/ACE3/commit/3b1a9ab415e36bf9735d454ffb4b09c9072e9c0d and add
`if !(local _vehicle) exitWith {};
to fn_addSpareParts.sqf .

That way it should add repair cargo wherever the vehicle is local, which makes sense.

SilentSpike:
This fix works fine. serverInit != run if vehicle is local to server. It just means run only on the server.

LecksC:
Hmm ok, it didn't add any repair cargo at all in our tests. My solution did spawn 1.

PabstMirror:
@LecksC you are right, and I can't see why besides a problem with `setVariable`

To reproduce:
Run this in Extended_Init_EventHandlers for a vehicle:
```
if (local _vehicle) then {_vehicle setVariable ["test2", 22, true];};
if (isServer) then {_vehicle setVariable ["test3", 33, true];};
```

then create a vehicles locally on a client connected to a dedicated server:
`x3 = "C_Offroad_01_F" createVehicle [1723.36,5769.56,0.00143957]; publicVariable "x3";`

and from the client's view "test3" is not defined

The effect is that cargo's loaded variable never syncs, so nothing shows up for clients.

commy2:
Related to this?
http://feedback.arma3.com/view.php?id=25565

[...]

PR #2489, opened 2015-09-17
only run addSpareParts on server

PabstMirror:
#2488 - Tag @jonpas


SilentSpike:
:+1:

jonpas:
Looks good, that was my bad!

SilentSpike:
If we are releasing a second hotfix, I'd be inclined to also throw this in there

PR #2012, opened 2015-08-04
Code cleanup of the Common module.

jokoho48:
Code cleanup of the Common module.

Split PRs from #1989.

PR #2490, opened 2015-09-17
cleanup common

commy2:
ref https://github.com/acemod/ACE3/pull/2012/files


MikeMatrix:
#2012

commy2:
^ I use that as base. (Therefore ref in the OP)
2012 is to large for me to merge with good conscience

Issue #2491, opened 2015-09-17
Script Error in "fnc_treatmentAdvanced_CPRLocal.sqf"


Th3Cr0w:
ACE3 Version: 3.3.1

**Mods:**
* `@CBA_A3`
* `@ace`
* and plenty more

**Placed ACE3 Modules:**
* ACE_moduleMedicalSettings
* ACE_moduleAdvancedMedicalSettings
* ACE_moduleReviveSettings

**Description:**
* When performing CPR a script error occurs

**Steps to reproduce:**
* Shoot AI until "Cardiac Arrest = true"
* perform CPR on unit upon finish a script-error appears

**Where did the issue occur?**
*  "Editor"

**RPT log file:**
````
18:12:47 Error in expression <ocal.sqf"


private "_reviveStartTime";
param ["_caller","_target"];

if (_targe>
18:12:47   Error position: <param ["_caller","_target"];

if (_targe>
18:12:47   Error Type String, expected Number
18:12:47 File z\ace\addons\medical\functions\fnc_treatmentAdvanced_CPRLocal.sqf, line 6
````

PR #2492, opened 2015-09-17
Fix CPR params

PabstMirror:
#2491


commy2:
`param` is a ugly trap for everyone using autocomplete with the scripting commands. Noticed that too.

SilentSpike:
Since this is targeted at master I'd go ahead and merge it, will probably want to hotfix into release too

SilentSpike:
I have no idea what happened, but that works :laughing: :+1:

Was that a cherry pick?

PabstMirror:
just branched off of release, so merge is easy to both branches

PR #2507, opened 2015-09-18
Script to search for undefined functions

PabstMirror:
Python script to generate a sqf script that checks all functions in game and makes sure they are defined.

I know there are some false positives in javelin as there functions are defined when used and not in preInit.

Output:
```
*********** Scaning for nil functions [count 857]
ace_interaction_fnc_showMenu is nil
ace_common_fnc_waitAndExec is nil
ace_common_fnc_localAnim is nil
ace_common_fnc_sendDisplayInformationTo is nil
ace_repair_fnc_isInStableCondition is nil
ace_interaction_fnc_openSubMenuSelf is nil
ace_interaction_fnc_addToTooltip is nil
ace_common_fnc_sendDisplayMessageTo is nil
ace_scopes_fnc_onDrawShortdot is nil
ace_common_fnc_makeCopyOfBody is nil
ace_common_fnc_displayMessage is nil
ace_common_fnc_disableKeyInput is nil
ace_common_fnc_sendHintTo is nil
ace_common_fnc_addInfoDisplayEventHandler is nil
ace_medical_fnc_addToTriageList is nil
ace_repair_fnc_createLitter is nil
ace_common_fnc_displayInformation is nil
ace_mk6mortar_fnc_simulateFindSolution is nil
ace_common_fnc_isModLoader is nil
ace_interaction_fnc_initialiseInteraction is nil
ace_medical_fnc_effectBlackOut is nil
ace_medical_menu_fnc_getTriageStatus is nil
ace_captives_fnc_handleVehicleChanged is nil
ace_missileguidance_fnc_checkSeekerLos is nil
ace_laser_selfdesignate_fnc_laserHudDesignatePFH is nil
ace_javelin_fnc_enableFire is nil
ace_fcs_fnc_traceBullet is nil
ace_explosives_fnc_SetTimer is nil
ace_common_fnc_formatVar is nil
ace_javelin_fnc_disableFire is nil
ace_interaction_fnc_openSubMenu is nil
ace_captives_fnc_setSurrender is nil
ace_common_fnc_getCanInteract is nil
```

PR #2509, opened 2015-09-18
Fix some nil functions

PabstMirror:
Fixing some functions that aren't defined (ref #2507)


addons/scopes/XEH_postInit.sqf
```diff
@@ -161,5 +161,3 @@ GVAR(showShortdot) = false;
         };
     };
 }] call EFUNC(common,addEventHandler);
-
-addMissionEventHandler ["Draw3D", {if (GVAR(showShortdot)) then {call FUNC(onDrawShortdot)};}];
```
commy2: wow. I'm sorry about this one.

commy2:
:+1:

PR #2527, opened 2015-09-19
Fix: Titan hitting ground instead of vehicle

alef:
We found the Titan didn't hit correctly the Punisher. In this link
http://www.15bandenere.com/forum/showthread.php?tid=619&pid=12002#pid12002 you can see the screenshots of an NLAW and Titan.

After looking for some memory points, I guess the simplest solution is to use aimPos, but I don't know if it breaks something else.

### DIFF
--- a/addons/missileguidance/functions/fnc_seekerType_Optic.sqf
+++ b/addons/missileguidance/functions/fnc_seekerType_Optic.sqf
@@ -17,7 +17,7 @@ TRACE_1("", _launchParams);
 // TODO:: Make sure the missile maintains LOS
 _foundTargetPos = [0,0,0];
 if(!isNil "_target") then {
-    _foundTargetPos = getPosASL _target;
+    _foundTargetPos = aimPos _target ;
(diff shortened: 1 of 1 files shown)


PabstMirror:
Testing with 
```
addMissionEventHandler ["Draw3D", {
    drawIcon3D ["", [1,0,0,1], ASLtoAGL (getPosASL cursorTarget), 0, 0, 0, "ASL", 1, 0.04, "PuristaMedium"];
    drawIcon3D ["", [1,0,0,1], ASLtoAGL (aimPos cursorTarget), 0, 0, 0, "aimPos ", 1, 0.04, "PuristaMedium"];
}];
```

I really like the results so far.




This should fix the issue I have where the missile will hit the ground between the 2 wheels and not impact the vehicle itself.

https://www.youtube.com/watch?v=b_JB968K0g0

Glowbal:
What is the status on this?

PabstMirror:
It's an improvement and should fix problems with missiles hitting the ground instead of vehicles.

Issue #2525, opened 2015-09-19
Wrong Interactions points for vehicles available under water


bux578:
ACE3 Version: 3.3.1 stable

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* Respawn

**Description:**
* There are only the "Remove Wheel" interaction points but not the main interaction point available under water. So divers can't unload items from a submerged vehicle (which should be possible imho).
Using the "Remove Wheel" interaction doesn't do anything though. So either this fix this (does that make sense under water?) or remove the interaction points.

**Steps to reproduce:**
* As a diver drive a truck over the pier in e.g. Stratis
* Get out of the vehicle and try to interact with it or with the wheels
![2015-09-19_00004](https://cloud.githubusercontent.com/assets/1235520/9976384/bd5056fa-5ee2-11e5-9463-2929168a1de1.jpg)

**Where did the issue occur?**
* "Editor"

**RPT log file:**
not necessary

Xerrass:
i think it woud be cool if you can get stuff out of underwater Wrecks and/or Vehicles will be a cool mission Addition to get stuf out of a crashed Plane or Boat.

PR #2529, opened 2015-09-19
Add CanInteractWith checks for repair

PabstMirror:
#2525

There is no canInterctWith check for repair actions, so right now you can remove wheels when underwater or handcuffed.

Does it make sense to do repairs underwater, if so we will want to add an exception for that??


commy2:
>Does it make sense to do repairs underwater, if so we will want to add an exception for that??

I'd say no.

SilentSpike:
Agreed :+1:

Glowbal:
Isn't this handled by the interaction actions?

PabstMirror:
createAction/addActionToXXX just takes a raw conditional

commy2:
^ This is true. Asked esterldunedain about this, but iirc it was done to leave flexibility or something.

Issue #2526, opened 2015-09-19
Weird objects (medical litter) in Zeus


bux578:
ACE3 Version: 3.3.1 stable

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* Respawn

**Description:**
* Medical litter items and other ACE3 stuff can be found in the Objects category in Zeus. They don't have a name and don't know Physx.
![2015-09-19_00005](https://cloud.githubusercontent.com/assets/1235520/9976418/a4030d72-5ee3-11e5-9007-536dde2584b2.jpg)

**Steps to reproduce:**
* Spawn as Curator and open Zeus
* Place nameless items from the Objects Tree

**Where did the issue occur?**
* "Editor"

**RPT log file:**
not necessary

SilentSpike:
Should be simple to fix

commy2:
scopeCurator = 0; it is I think. Might be 1 can't remember

SilentSpike:
Yep, 0 would be right for this case

SilentSpike:
1 only applies to the regular `scope` property

bux578:
Why not proper name them, give em physx and leave them available?

SilentSpike:
Depending on what they all are that might be desired. I'd probably hide the litter objects.

Glowbal:
I'd hide the objects from curator, see about which ones we want to add and do it properly at a later stage.

SilentSpike:
Already done locally, was going to quickly look at what the other ones are and make appropriate edits

SilentSpike:
Nevermind, looks like it was only litter that was the problem.

PR #2533, opened 2015-09-19
Fix medical litter scopeCurator values

SilentSpike:
Solves #2526 where medical litter would appear in zeus with no name.


commy2:
:+1: easy fix, merge

PR #2535, opened 2015-09-19
common code cleanup part 7

commy2:


commy2:
Do we need setProne?
There is a goKneeling that is functional.

commy2:
done otherwise

PabstMirror:
@commy2 - it's no worse than before, we can deal with it later, just merge if you want to keep going

commy2:
There is another nifty error with it that I had in similar scripts already. If you put you weapon on back or have no weapon then `currentWeapon player == primaryWeapon player == ""`

PR #2539, opened 2015-09-20
common code cleanup part 8

commy2:


addons/common/functions/fnc_setHearingCapability.sqf
```diff
@@ -11,6 +11,8 @@
  * None
  *
  * Public: Yes
+ *
+ * Note: uses player
```
jokoho48: i think better is `is Local on executed Unit`

commy2: Thats not true though. The issue with `player` is that it doesn't work with remote controled units zeus. (And drones in case of `vehicle player`)

PR #2542, opened 2015-09-20
Fix cargo setVariable on server with InitPost

### LINKED ISSUE #2488: Wheel added to cargo on player connect.

PabstMirror:
Actually fix #2488

Switching to InitPost seems to fix the problem with setVariable not working

Also fixes the same problem with `ACE_Cargo` in cargo.

### DIFF
--- a/addons/cargo/CfgEventHandlers.hpp
+++ b/addons/cargo/CfgEventHandlers.hpp
@@ -16,85 +16,73 @@ class Extended_Killed_EventHandlers {
     };
 };
 
-class Extended_Init_EventHandlers {
+//Need initPost or we have problems with setVariable with 'ACE_Cargo'
+class Extended_InitPost_EventHandlers {
     class StaticWeapon {
         class ADDON {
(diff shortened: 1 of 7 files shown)


SilentSpike:
Huh, that was weird timing...

PR #2543, opened 2015-09-20
Remove: pain diagnosis on dead patients

alef:


SilentSpike:
I'd probably just stick this into the `alive _target` check further up in the function. Unless they have to appear in a specific order.

alef:
I've thought that the information of finding someone bleed to death to be useful, so I've removed only the _pain_ message.

SilentSpike:
I mean I'd move lines 39-43 to below line 24 :smile:

alef:
I would prefer to keep the order of messages.

PR #2545, opened 2015-09-20
common code cleanup part 9

commy2:
ref: #2012

### DIFF
--- a/addons/common/functions/fnc_dumpArray.sqf
+++ b/addons/common/functions/fnc_dumpArray.sqf
@@ -25,7 +25,9 @@ for "_i" from 0 to _depth do {
 _depth = _depth + 1;
 
 if (IS_ARRAY(_var)) then {
-    if (count _var > 0) then {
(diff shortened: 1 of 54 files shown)


addons/common/functions/fnc_dumpArray.sqf
```diff
@@ -14,11 +25,14 @@ for "_i" from 0 to _depth do {
 _depth = _depth + 1;
 
 if (IS_ARRAY(_var)) then {
-    if ((count _var) > 0) then {
+    if (count _var > 0) then {
```
jokoho48: _var isEqualTo []

commy2: noted

commy2:
Ready to merge.

MikeMatrix:
:+1: for correct use of reference :D

jokoho48:
wuhu finaly gz @commy2 :D

PR #2550, opened 2015-09-21
common code cleanup part 10, close #2319

commy2:
ref: #2012

### DIFF
--- a/addons/common/functions/fnc_getHitPointsWithSelections.sqf
+++ b/addons/common/functions/fnc_getHitPointsWithSelections.sqf
@@ -1,13 +1,17 @@
 /*
  * Author: commy2
- *
- * Returns all hitpoints and their selections of any vehicle. Might contain duplicates if the turrets contain non unique hitpoints with different selection names.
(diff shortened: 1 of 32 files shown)


addons/common/functions/fnc_getHitPointsWithSelections.sqf
```diff
@@ -11,51 +11,9 @@
...
-
-        _class = inheritsFrom _class;
-    };
-
-} forEach _hitpointClasses;
-
-[_hitpoints, _selections]
+private "_hitPointsWithSelections";
+_hitPointsWithSelections = getAllHitPointsDamage _vehicle;
+_hitPointsWithSelections resize 2;
```
jokoho48: ? why you resize

commy2: So it matches the output of the previous function. This thing is deperecated just haven't added the tag yet, because it's from a different PR.

commy2:
Ready to merge.
I think I'm finally done with `common/functions` yay!

bux578:
GJ :+1:

PR #2551, opened 2015-09-21
common code cleanup part 11

commy2:
ref: #2012


addons/common/XEH_postInit.sqf
```diff
@@ -108,31 +107,72 @@ if(!isServer) then {
...
+//////////////////////////////////////////////////
+
+private ["_currentVersion", "_previousVersion"];
+
+_currentVersion = getText (configFile >> "CfgPatches" >> QUOTE(ADDON) >> "version");
+_previousVersion = profileNamespace getVariable ["ACE_VersionNumberString", ""];
+
+// check previous version number from profile
+if (_currentVersion != _previousVersion) then {
+    // do something
```
jokoho48: ?

commy2: Changed the order a bit. Removed below.

commy2:
Done !

PR #2556, opened 2015-09-21
add deprecated warnings

commy2:
ref: #2012


jonpas:
Version 1.5?

commy2:
oops. 3.5

jonpas:
:+1: Best would be to write `3.5.0` IMO, to prevent confusion.

addons/main/script_macros.hpp
```diff
@@ -132,7 +132,7 @@
 #define ACE_LOGWARNING_7(message,arg1,arg2,arg3,arg4,arg5,arg6,arg7) ACE_LOGWARNING(FORMAT_7(message,arg1,arg2,arg3,arg4,arg5,arg6,arg7))
 #define ACE_LOGWARNING_8(message,arg1,arg2,arg3,arg4,arg5,arg6,arg7,arg8) ACE_LOGWARNING(FORMAT_8(message,arg1,arg2,arg3,arg4,arg5,arg6,arg7,arg8))
 
-#define ACE_WARNINGFORMAT(message) ACE_LOGFORMAT(COMPONENT,"WARNING",message)
+#define ACE_WARNINGFORMAT(message) ACE_LOG(COMPONENT,"WARNING",message)
```
PabstMirror: I think
`ACE_LOGWARNING` is for rpt writing
`ACE_WARNINGFORMAT` just returns a string
so I don't think we wanto change it

commy2:
Example:
`21:48:03 [ACE] (common) WARNING: ace_common_fnc_isTurnedOut is deprecated. Support will be dropped in version 3.5.0. Replaced by: isTurnedOut`

addons/main/script_macros.hpp
commy2: Weird. Makes sense though

commy2:
done

MikeMatrix:
Maybe replace it by an deprecation macro?
Would clean it up a little as well as introduce a standard way of deprecating things.

commy2:
Agreed and done

jonpas:
Doesn't CBA already have deprecation macros?

commy2:
Well, they aren't ACE formated.

jonpas:
True.

MikeMatrix:
One other thing might to be considered is to use `QFUNC` instead of writing the full function name? Just a thought.

commy2:
These will be removed soonish anyway and will probably never be moved again.

MikeMatrix:
True. Maybe for future use might be easier.

commy2:
k. Anything else or merge?

MikeMatrix:
Looks good to me.
:+1:

PR #2559, opened 2015-09-21
Use CBA PFH in disableUserInput

commy2:
Also fixes that all buttons where ignored in that state. Was an leftover from AGM that is no longer needed with CBA keyhandling.


commy2:
ref: #2012

Issue #1235, opened 2015-05-15
Explosives won't rotate at placement


Raspu86:
ACE3 Version: 3.0.0

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
[...]

Issue #1675, opened 2015-06-18
Remove the modifier key


Khaosmatic:
Hi there guys, with the release of the new items in the latest update oh boy some great fun is going to be had. Something that I realized was awesome was the tactical ladder being able to just scroll to adjust the height whereas I thought it would be done in the interaction menu!

What I'm saying is, it was such easy use being able to scroll why cannot it be applied to the other rotational/adjustable placeables like explosives or the tripod. Mainly because ARMA already needs a lot of key-binds and it seems to me the modifier key cannot be bound to a lot of keys.

Would it be possible to get an option to remove the use of the modifier key?

esteldunedain:
I'm checking the code and I agree. ACE_Modifier is currently only used in conjunction with the mouse wheel (which used not to be the case):

ACE_Dragging: lowering/raising object
ACE_Explosives: Rotating explosives
ACE_Sandbag: Rotating sandbags
ACE_Tripod: Lowering/raising tripod

So yes, I think it could be safely scraped altogether.

SilentSpike:
:+1: I like it

gienkov:
It should stay since tactical ladder uses modifier to rotate the ladder and mousewheel alone to adjust height of it.

Khaosmatic:
A fair point with the tactical ladder.

commy2:
The biggest problem with the modifier key is, that the default keybind (LCtrl) doesn't work at all.
`{ace_modifier} call ace_common_fnc_monitor`

Seems like CBA keyhandling has issues with shift / ctrl / alt as sole keybind.

commy2:
resolved in https://github.com/acemod/ACE3/pull/2598

PR #2564, opened 2015-09-21
rework modifier key, Code Cleanup Sandbag, Code Cleanup Tripod

commy2:
https://github.com/acemod/ACE3/issues/1675

Currently the modifier key doesn't work at all with the default key bind.

| | ctrl not pressed | ctrl pressed
------------- | ------------- | -------------
| dragging | - | -
| carrying | lowering/raising | -
| explosives | - | rotate
| sandbag | - | rotate
| tripod | lowering/raising | -
| tactical ladder | extend | tilt

- [x] dragging
- [x] carrying
- [x] explosives
- [x] sandbag
    - [x] some missing event handling: dc, respawn, killed, unconscious, captive, surrendered
    - [x] check if sandbag still in inventory during placement
- [x] tripod
    - [x] some missing event handling: dc, respawn, killed, unconscious, captive, surrendered
    - [x] correct string
    - [x] distance check while adjusting
    - [x] move placement option out of "Equipment"
- [x] tactical ladder
    - [x] some missing event handling: dc, respawn, killed, unconscious, captive, surrendered
    - [x] correct string

- [ ] doc on class name: http://ace3mod.com/wiki/missionmaker/class-names.html#tactical-ladder
- [ ] MP testing with hard dc, unconsciousness, stealing the bag from the backpack etc.


PabstMirror:
This seems like a good solution, should also fix #1235

CBA had an issue for this, but it's closed https://github.com/CBATeam/CBA_A3/issues/77

btw `displayAddEventHandler` will take CODE or STRING.

commy2:
>btw displayAddEventHandler will take CODE or STRING.

Yup, it does for "keyDown" and "keyUp", but it doesn't for "mouseZChanged" which I copied it from. Fixed now.

addons/common/functions/fnc_handleModifierKeyInit.sqf
```diff
@@ -0,0 +1,16 @@
...
+ * None
+ *
+ * Return Value:
+ * None
+ *
+ * Public : No
+ */
+#include "script_component.hpp"
+
+(findDisplay 46) displayAddEventHandler ["KeyDown", {_this call FUNC(handleModifierKey)}];
```
jokoho48: why not using `FUNC(handleModifierKey)`

commy2:
- [ ] some rewriting in EFUNC(interaction,showMouseHint);

commy2:
I'm using CorruptedHeart's wonderfull mouse hint's to make it more clear for user what they can do with the stuff they are interacting with.



Has anyone a nice idea on how to improve this design to make it as user friendly as possilbe? Besides making the control bigger, so the text is not cut off of course.

addons/common/functions/fnc_handleModifierKeyInit.sqf
commy2: done

SilentSpike:
No ideas currently about the UI itself, but for carrying I notice you've listed `ctrl` to `raise/lower`. Isn't that inconsistent with elsewhere?

MikeMatrix:
I think an improvement could be to make it more destinct on what is keybind and what is the action it takes. A bit of spacing and the order of things (Modifier keys in front) can go a long way of making it look nicer and more understandable.

commy2:
- [ ] fix being able to throw grenades when carrying / dragging.

commy2:
Done. Needs some MP testing with getting killed/kocked out and switching units and stuff.

Issue #2583, opened 2015-09-23
40mm UGLs don't create frags


commy2:

bux578:
https://github.com/acemod/ACE3/issues/342

bux578:
Status @commy2 ?

esteldunedain:
I've taken a look. It's not a bug in the strict sense of the work; the 40mm UGL explosion parameters as defined in vanilla arma are not strong enough for the frag module to kick in.

We might want to change those parameters later. I'm pushing this back.

bux578:
IRL they produce a shitload! of frag.

PR #2584, opened 2015-09-24
Code Cleanup Frag

jokoho48:
Code Cleanup Frag Module

This is still WIP i will try to fix some bugs here too


commy2:
ref: #2583

addons/frag/functions/fnc_addBlackList.sqf
```diff
@@ -1,4 +1,4 @@
 #include "script_component.hpp"
-private ["_round"];
-_round = _this select 0;
-GVAR(blackList) set [(count GVAR(blackList)), _round];
+if (params ["_round"]) then {
+    GVAR(blackList) set [(count GVAR(blackList)), _round];
```
commy2: pushBack

addons/frag/functions/fnc_addPfhRound.sqf
```diff
@@ -1,30 +1,32 @@
...
+if((count GVAR(objects)) > GVAR(MaxTrack)) exitWith {};
+
 
-if(_gun == ACE_player) then {
-    _doFragTrack = true;
-} else {
-    if((gunner _gun) == ACE_player) then {
-        _doFragTrack = true;
+if (_doFragTrack) then {
+    _doFragTrack = if (_gun == ACE_player) then {
```
commy2: I strongly suggest
`if () then {_doFragTrack = } else {_doFragTrack = };`
over
`_doFragTrack = if () then {} else {};`

Scope should be no problem, because we private all variables.

addons/frag/functions/fnc_addPfhRound.sqf
```diff
@@ -1,30 +1,32 @@
...
+        true;
     } else {
-        if(local _gun && {!(isPlayer (gunner _gun))} && {!(isPlayer _gun)}) then {
-            _doFragTrack = true;
+        if((gunner _gun) == ACE_player) then {
+            true
+        } else {
+            if(local _gun && {!(isPlayer (gunner _gun))} && {!(isPlayer _gun)}) then {
```
commy2: EFUNC(common,isPlayer)

addons/frag/functions/fnc_addPfhRound.sqf
```diff
@@ -1,30 +1,32 @@
...
-if( (count GVAR(objects)) > GVAR(MaxTrack)) exitWith { };
+if((count GVAR(objects)) > GVAR(MaxTrack)) exitWith {};
+
 
-if(_gun == ACE_player) then {
-    _doFragTrack = true;
-} else {
-    if((gunner _gun) == ACE_player) then {
-        _doFragTrack = true;
+if (_doFragTrack) then {
```
PabstMirror: The old logic had _doFragTrack default to false and then if we matched the shooter, it set it to true.
This if statement blocks those checks.
Also never set _doFragTrack to false if it's true from the params.

TBH the old logic was fine

addons/frag/functions/fnc_addTrack.sqf
```diff
@@ -1,23 +1,17 @@
...
-_index = (count GVAR(traces));
-_obj = _this select 1;
-_origin = _this select 0;
-_color = [1,0,0,1];
-if((count _this) > 2) then {
-    _color = _this select 2;
-};
 _positions = [];
 _objSpd = vectorMagnitude (velocity _obj);
 _positions set[(count _positions), [(getPos _obj), _objSpd]];
```
Glowbal: This should probably be replaced with a `pushback` while we are at it.

addons/frag/functions/fnc_findReflections.sqf
```diff
@@ -48,7 +40,7 @@ if(_zIndex < 5) then {
             _lastPos = _testPos;
         };
     };
-    _params set[4, _zIndex+1];
+    (_this select 0) set[4, _zIndex+1];
```
Glowbal: Can now use `_args` here

Issue #2590, opened 2015-09-24
XM307A Cannot be moved, while XM307 and other static can be moved


mickeymen:
ACE3 Version: 3.3.2
Stable version

**Mods:**
* `@CBA_A3`
* `@ace`
* `@DragonFyre`

**Placed ACE3 Modules:**
None

**Description:**
Please check all static weapons. Probably not everything can be moved like in the example with XM307A (automatic-grenade launcher). 

**Steps to reproduce:**
Try deploy and move XM307 and XM307A

**Where did the issue occur?**
Editor
**RPT log file:**
Not needed

PR #2591, opened 2015-09-24
Ignore AI for dragging crew checks

PabstMirror:
#2590 

UAV / Automated Turrets have AI "crew"
The weight should be the same so we can just ignore them in the crew count check.

### DIFF
--- a/addons/dragging/functions/fnc_canDrag.sqf
+++ b/addons/dragging/functions/fnc_canDrag.sqf
@@ -21,7 +21,7 @@ _target = _this select 1;
 
 if !([_unit, _target, []] call EFUNC(common,canInteractWith)) exitWith {false};
 
-// a static weapon has to be empty for dragging
-if ((typeOf _target) isKindOf "StaticWeapon" && {count crew _target > 0}) exitWith {false};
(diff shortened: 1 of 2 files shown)


PabstMirror:
Also, AI turrets will continue to shoot while you carry it, which is kind of awesome.

untitled

commy2:
Wow. `disableAI` ?

commy2:
Why does `crew` return drone logics anyway?

addons/dragging/functions/fnc_canDrag.sqf
```diff
@@ -21,7 +21,7 @@ _target = _this select 1;
 
 if !([_unit, _target, []] call EFUNC(common,canInteractWith)) exitWith {false};
 
-// a static weapon has to be empty for dragging
-if ((typeOf _target) isKindOf "StaticWeapon" && {count crew _target > 0}) exitWith {false};
+// a static weapon has to be empty for dragging (ignore UAV AI
```
jonpas: Missing `)`

commy2:
Pulled this in https://github.com/acemod/ACE3/pull/2564

PR #2593, opened 2015-09-24
findUnloadPosition (handle piers, carriers, water)

### LINKED ISSUE #2475: Cargo: Unload locations unsuitable
ACE3 Version: 3.3.1 A3 Stable

@CBA_A3
@ace

Placed ACE3 Modules:
None

Description:
When unloading cargo from a vehicle the object unloads no where near the player who requested the cargo be unloaded. Testing with wheel I found it to be usually on another side of the vehicle (more often than not complete opposite side of vehicle) and significant distance from the vehicle or player. Shouldn't unloading items place it at the nearest available place or the requester or even better directly into the players hands. 

Steps to reproduce:
1. Set-up player and empty vehicle (hunter used)
2. Enter vehicle cargo and unload wheel.
3. Observe location of wheels unloading. (you may have to search)

Where did the issue occur?
"Singleplayer"*

RPT log file:
No Errors

### LINKED ISSUE #2525: Wrong Interactions points for vehicles available under water

PabstMirror:
WIP - should fix #2475 #1908 

Right now manually searching for a free spot is giving much better results than `findEmptyPosition` and it's still fairly fast.

Edit, also fix 	#2525


Glowbal:
@PabstMirror what is the status on this PR?

PabstMirror:
Was working fine for "normal" objects like tires/crates/static weapons, it can find a spot much closer to the player who is unloading.

But because it's possible to load/unload BIG things like a 10 ton cargo container, I was struggeling to get a good method to find the real object dimensions.

I'll try to get it working before 3.5

Feivel69:
I don't know if it's possible, but it would also be very nice to have an optional argument to give accurate position (coordinates/height/markerpos etc.) where the object should be unloaded - so the automatic place search gets ignored. Concrete example: You have a logistics center in your mission with a laptop where you create crates which get put into a cargo container and you want them to be unloaded to a table or specific ground position.

MikeMatrix:
Maybe we could do a similar system to that of the explosives, where we display a virtual version of the item being unloaded and let the person unloading the locations via that, but limit it to an area around the cargo.
Maybe as a secondary option to just straight up unloading it.

PabstMirror:
Closing for #3228

PR #2597, opened 2015-09-25
fix spectator view if called instantly

Cuel:
If you call setSpectator directly from init.sqf you'll be put into some
weird limbo mode with borders and spectator not working. This should fix
it


SilentSpike:
Yep, this seems fine, thanks for taking the initiative and fixing it :+1:

In future we can use the waitUntilAndExecute (or similar) in these cases to abolutely ensure the display exists first.

PR #2184, opened 2015-08-16
Code cleanup of respawn module

MikeMatrix:
Code Cleanup of the respawn Module.

PR #2602, opened 2015-09-26
Code cleanup respawn module, close #2184

commy2:


addons/respawn/ACE_Settings.hpp
```diff
@@ -1,3 +1,4 @@
+
```
jonpas: Empty rouge line?

commy2: Nah, I always put an empty line on the top and bottom of a config.

jonpas: *no comment* :P

PR #2605, opened 2015-09-26
Fixed Laser Update Position Bug

### LINKED ISSUE #1715: Laser target doesn't update position
ACE3 Version: 3.1.2

**Mods:**
* @cba_a3
* @ace3

**Description:**
The laser target entity that all laser designators create doesn't update position.  When you first lase, the target is created and can be locked onto, but it doesn't change its position unless you turn the laser off and on again.

SP/MP

**RPT log file:**
No config/script errors

TheNiner:
Fixed #1715

### DIFF
--- a/addons/laser/CfgVehicles.hpp
+++ b/addons/laser/CfgVehicles.hpp
@@ -5,7 +5,7 @@ class CfgVehicles {
         // @TODO: Changing the model and simulation hides it, but THEN IT DOESNT SPAWN WTF!?
         model = "\A3\Weapons_F\empty.p3d";
         destrType = "DestructNo";
-        simulation = "house";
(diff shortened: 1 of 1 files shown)


PabstMirror:
I'm not entirely sure why we changed simulation type.
I know this makes it visable in NV 

image

but given that vanilla lasers are broken/hard to use, I think making them functional is more important.

Hopefully this and #2560 should get vanilla laser weapons working as they should.

TheNiner:
Yeah i noticed that too. But as long as the simulation type is hardcoded in the engine, i guess we can't do anything about it.

TheNiner:
Any chance for getting this into 3.4.0?

PR #2607, opened 2015-09-26
Code cleanup Backpacks

commy2:


addons/backpacks/XEH_postInit.sqf
```diff
@@ -1,3 +1,3 @@
 #include "script_component.hpp"
 
-["backpackOpened", DFUNC(backpackOpened)] call EFUNC(common,addEventHandler);
+["backpackOpened", {_this call FUNC(backpackOpened)}] call EFUNC(common,addEventHandler);
```
MikeMatrix: I've seen you do this in multiple PR's now, and understanding the reason behind this it is a good thing to do.
However I feel like this is fixing the symptoms rather the issue. The issue being, that `addEventHandler` can only create copies of code, instead of being also be able to reference to existing variables.

Fixing this at the source of the problem might be the better approach.

commy2: I'll take a look at that. But in a later PR to not cause conflicts now.

PR #2613, opened 2015-09-26
Improved Slideshow API

jonpas:
- Improved public function parameter handling
- Fixed possible issue with synchronized objects
- Changed default duration to a macro

### DIFF
--- a/addons/slideshow/functions/fnc_createSlideshow.sqf
+++ b/addons/slideshow/functions/fnc_createSlideshow.sqf
@@ -10,7 +10,7 @@
  * 4: Slide Duration <NUMBER> (0 disables automatic transitions)
  *
  * Return Value:
- * Parsed List <ARRAY>
+ * None
  *
  * Example:
  * [[object1, object2, object3], [controller1], ["images\image1.paa", "images\image2.paa"], ["Action1", "Action2"], 5] call ace_slideshow_fnc_createSlideshow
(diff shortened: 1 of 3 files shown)


addons/slideshow/functions/fnc_createSlideshow.sqf
```diff
@@ -20,18 +20,21 @@
 #include "script_component.hpp"
 
 private ["_currentSlideshow", "_slidesAction", "_varString"];
-params ["_objects", "_controllers", "_images", "_names", "_duration"];
+params [
+    ["_objects", [], []],
```
commy2: If you want to only accept arrays as an element in `params` then you have to use:
`["_objects", [], [[]]],`
The third element is an array containing examples of all available data types. `[]` would be the default value and means everything is accepted.

jonpas: Yep, I realized I forgot one sub-level as soon as I made this PR.

addons/slideshow/functions/fnc_createSlideshow.sqf
```diff
@@ -20,21 +20,24 @@
...
+    ["_images", [], [[]] ],
+    ["_names", [], [[]] ],
+    ["_duration", 0, [0]]
+];
 
 // Verify data
-if (count _images != count _names || {count _images == 0} || {count _names == 0}) exitWith {
-    ACE_LOGERROR("Slideshow Images or Names fields can NOT be empty and must have equal number of items!");
+if (_objects isEqualTo []) exitWith {
+    ACE_LOGERROR("Slideshow Objects field may NOT be empty!");
```
MikeMatrix: `may` would imply it would still be valid. Use `must` instead.

addons/slideshow/functions/fnc_createSlideshow.sqf
```diff
@@ -20,21 +20,24 @@
...
+};
+if (count _images != count _names || {_images isEqualTo []} || {_names isEqualTo []}) exitWith {
+    ACE_LOGERROR("Slideshow Images or Names fields may NOT be empty and must have equal number of items!");
```
MikeMatrix: Same here

jonpas: Stupid English :p

jonpas:
Done.

PR #2614, opened 2015-09-26
Optionalized Cargo Loaded Hint

jonpas:
And disabled it when called through `addCargoItem`, for example `addSpareParts`, which happens on vehicle init, would show the hint, even if the vehicle would be spawned during the mission.

### DIFF
--- a/addons/cargo/functions/fnc_addCargoItem.sqf
+++ b/addons/cargo/functions/fnc_addCargoItem.sqf
@@ -6,6 +6,7 @@
  * 0: Item Classname <STRING>
  * 1: Vehicle <OBJECT>
  * 2: Amount <NUMBER> (default: 1)
+ * 3: Show Hint <BOOL> (default: false)
  *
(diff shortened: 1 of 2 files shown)


commy2:
^ yeah thats a bug in the current master.
I plan on reworking this anyway, but this at least fixes this bug in master for now.

jonpas:
I "created" the bug as well since I added that when adding spare parts framework.

How do you plan to rework it?

commy2:
A function that is called by action that ends in a progress bar.
A callback function called from the progress bar.
And a load function that is called by the callback function that also can be used as API.

Hint, animations etc. are done by the first two. Same for unloading.

jonpas:
Sounds good!

commy2:
:+1:

SilentSpike:
Regarding the rework discussion here, should be handled by the changes made in #2628

You may want to build upon what's there though

Issue #2307, opened 2015-08-29
Repair work on sub-turrets


PabstMirror:
Use the new "index" commands for hitpoint to get sub-turret repair points.

From #2241

>getAllHitPointsDamage / setHitIndex / getHitIndex - can make repair work on sub turrets that have identical hitpoint names

PR #2616, opened 2015-09-26
Repair all the points

PabstMirror:
#2307

addRepairActions/MiscRepair/setHitPointDamage now work on index so we can repair sub turrets.

Converted `EFUNC(common,getHitPointsWithSelections)` to `getAllHitPointsDamage`.

Gunner and commander turrets both have same name `HitTurret`, with new "index" commands we can actually fix both now.

I still need to do testing on addon vehicles.

@jonpas 
https://github.com/acemod/ACE3/blob/master/addons/repair/functions/fnc_doRepair.sqf#L45

This sets each sub member in hitPointGroups to 0,
Which can be lower than `[_unit] call FUNC(getPostRepairDamage)`
Should it be something like the minimum between it's current health and what the master hitpoint is being repaired to?


jonpas:
> https://github.com/acemod/ACE3/blob/master/addons/repair/functions/fnc_doRepair.sqf#L45
This sets each sub member in hitPointGroups to 0,
Which can be lower than [_unit] call FUNC(getPostRepairDamage)
Should it be something like the minimum between it's current health and what the master hitpoint is being repaired to?

Since it's a group I think it's simply easier to handle it this way, could add partial repair and then also add hitPointGroups sub-member repair status checks. I don't think it makes any difference to the user really. At the point when I was writing that, I simply set it to 0 and only calculated the leader point so I didn't have to worry about checking the status on sub-members when evaluating if repair can be made.

One thing that might be good to test is if a sub-member is damaged and the leader is not, can you still repair?

PabstMirror:
Going to rebase in a new PR

PR #2620, opened 2015-09-26
Sitting Framework

jonpas:
- Cleaned up rotation left-overs (now handled by config)
- Implemented better sitting framework in terms of interactions (only config values and `XEH_ENABLED;` must now be added) ~~while sacrificing a bit of init performance as InitPost happens on `class All`~~ changed to `class ThingX`
- Integrated 1 GVAR and 1 setVariable into a single setVariable
- Added carrying to chairs

Framework documentation coming if this gets merged.


addons/sitting/functions/fnc_sit.sqf
```diff
@@ -16,23 +16,14 @@
...
-
 // Overwrite weird position, because Arma decides to set it differently based on current animation/stance...
 _player switchMove "amovpknlmstpsraswrfldnon";
 
 // Add scroll-wheel action to release object
-_actionID = _player getVariable [QGVAR(StandUpActionID), -1];
-
-if (_actionID != -1) then {
-    _player removeAction _actionID;
-};
```
jonpas: @commy2 Why was/is L30-L34 needed? I don't see how it can happen that the action stays, `FUNC(Stand)` is called from all cases of standing up.

jonpas:
Status?

addons/sitting/functions/fnc_sit.sqf
commy2: These lines are needed to prevent duplicate actions from being added. Could possibly happen in certain `selectPlayer` / multiple Zeus scenarious. I'd add that back in.

jonpas: You never remove `QGVAR(StandUpActionID)` variable though, there is always something in it, it just changes as you sit next time.

PR #2622, opened 2015-09-27
Replaced isAlive function with alive command

jonpas:
`EFUNC(common,isAlive)` is deprecated.


commy2:
:+1: thanks again for the contribution. easy to miss these.

jonpas:
Just noticed it in the rpt error (thanks to your deprecation messages) and did a search on GitHub.

PR #2150, opened 2015-08-13
Code cleanup of Interaction module

jokoho48:
Code cleanup of Interaction module.

PR #2623, opened 2015-09-27
Interact menu settings after init

### LINKED ISSUE #1171: The land of the mysteriously missing interact menus

### LINKED ISSUE #2486: Settings (size, colour) of menu interaction, cannot be saved.
ACE3 Version: 3.3.1
Stable

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
None

**Description:**
Modify the text size and colors of interaction menu, always and always return to the standard indicators (default size and default colours).

**Steps to reproduce:**
Try to modify the interaction menu with yourself(color and size) via ACE Settings in the left upper corner. First you will see that the changes came into force, but later(after small time), the interaction menu will return to its default size and colour.

Editor

**RPT log file:**
Not needed

PabstMirror:
fix #2486 (might fix #1171 as well???)

* setupTextColors now runs after SettingsInitialized and only then do we install render
* settingsChanged uses count to handle possible capitalization differences
* minor optimization in fnc_render - (save 0.015ms wow !!!!!!!!)

### DIFF
--- a/addons/interact_menu/XEH_clientInit.sqf
+++ b/addons/interact_menu/XEH_clientInit.sqf
@@ -8,17 +8,19 @@ GVAR(cachedBuildingActionPairs) = [];
 
 GVAR(ParsedTextCached) = [];
 
-//Setup text/shadow/size/color settings matrix
-[] call FUNC(setupTextColors);
 ["SettingChanged", {
-    PARAMS_1(_name);
-    if (_name in [QGVAR(colorTextMax), QGVAR(colorTextMin), QGVAR(colorShadowMax), QGVAR(colorShadowMin), QGVAR(textSize), QGVAR(shadowSetting)]) then {
+    params ["_name"];
+    if (({_x == _name} count [QGVAR(colorTextMax), QGVAR(colorTextMin), QGVAR(colorShadowMax), QGVAR(colorShadowMin), QGVAR(textSize), QGVAR(shadowSetting)]) == 1) then {
         [] call FUNC(setupTextColors);
     };
 }] call EFUNC(common,addEventhandler);
 
-// Install the render EH on the main display
-addMissionEventHandler ["Draw3D", DFUNC(render)];
+["SettingsInitialized", {
+    //Setup text/shadow/size/color settings matrix
+    [] call FUNC(setupTextColors);
+    // Install the render EH on the main display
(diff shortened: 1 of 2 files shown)


commy2:
I take from this that all postInit are guaranteed to execute before SettingsInitialied ?
Pls merge soon, so I can get done with #2150 without ze merge conflicts.

PabstMirror:
I think it goes like this
```
run all preInit
run all postInit
...possibly some frames later...
run all ace_modules (server finishes before clients start)
read client saved settings
run "SettingsInitialized" event
run all runAtSettingsInitialized functions
```
#2150 should be fine, this is for interact_menu vs interaction

commy2:
Oh, right. Derp.

Glowbal:
@PabstMirror I think it goes more like this:

Server side:
```
- run preInit (server side collection of config settings)
- run init
- ACE module init
- postInit (settings initalized event is called somewhere during postInit)
```

Client side:
```
- run preInit
- run init
- ACE module init (settings are server only, so won't do much)
- postInit
- Sometime after or during post init, settings arrive. This depends on the moment of joining by the client. At start: After postInit. JIP: either after or during postInit.
```

PabstMirror:
modules get added to collection in their init
but don't get run until event `InitSettingsFromModules` which is after all other postInit

Glowbal:
Oh right, that got changed from the first implementation. I forgot about that. :+1:

PR #2628, opened 2015-09-27
Add support for virtual cargo

SilentSpike:
Allows cargo objects to be added via classname and reduces simulation/synchronization overhead. Objects become physical upon unloading and cannot be re-made virtual, only scripts can add such virtual objects.

Also separates the player feedback from the loading/unloading code and adds a progress bar.


addons/cargo/XEH_postInit.sqf
```diff
@@ -1,5 +1,45 @@
...
+    _itemName = getText (configFile >> "CfgVehicles" >> _itemClass >> "displayName");
+    _vehicleName = getText (configFile >> "CfgVehicles" >> typeOf _vehicle >> "displayName");
+
+    ["displayTextStructured", [[_hint, _itemName, _vehicleName], 3.0]] call EFUNC(common,localEvent);
+
+    if (_unloaded) then {
+        // Invoke listenable event
+        ["cargoUnloaded", [_item, _vehicle]] call EFUNC(common,globalEvent);
+    };
+
```
jonpas: Wouldn't this all be nicer in separate functions rather than all in this file?

SilentSpike:
These changes can likely be optimized/made neater. I'm not too familiar with the original code for cargo so it was a fairly messy workflow in order to get a functional implementation. Tested and working:

- Loading/unloading virtual objects
- Loading/unloading actual objects
- Cargo capacity with virtual/actual objects
- Progress bar and failure to load

SilentSpike:
Wasn't entirely sure whether to separate the virtual/object handling between functions. It seemed like it would be wasteful to do that though.

Anyone more familiar with the code feel free to add your own commits to improve this implementation.

SilentSpike:
In seperating the player feedback from the loading system:

- `fnc_startLoad` and `fnc_startUnload` initiate the progress bar (or fail immediately if the action can't be taken) which is based on the size of the item being moved
- The progress bar will raise the respective `"LoadCargo"` or `"UnloadCargo"` event (they were unused so I assumed this was to be their intended purpose)
- The event will take care of calling `fnc_(un)loadItem` and handle the player feedback based on whether it was successfully (un)loaded or not

So the loading and unloading functions themselves (the ones called by the events) contain no player feedback and are safe to use for scripting cargo manipulation

jonpas:
> The progress bar will raise the respective "LoadCargo" or "UnloadCargo" event (they were unused so I assumed this was to be their intended purpose)

I actually added them as public API, we needed a way to trace it easily for our persistence system. But any use is of course great! :grin:

SilentSpike:
Reassigning to commy since he was talking about messing with the cargo code similarly to this

PabstMirror:
I'm thinking that addSpareParts should add virtual? (just note, doesn't have to be in this PR).

Will have some problems with #2593, so lets merge this first and I can work out the conflicts.

jonpas:
addSpareParts uses events, so it's automatically virtual if I understand this PR correctly. I added AddCargoByClass event just because of addSpareParts.

SilentSpike:
Yeah, it would automatically be made virtual by this PR. It could be changed to just call the actual function, but I didn't want to mess with that because I didn't know if there was any intentional structure to it

jonpas:
> It could be changed to just call the actual function, but I didn't want to mess with that because I didn't know if there was any intentional structure to it

The intention was simply: it's cleaner if there are events, as events are made to be backward compatible (no name changes to it), while the functions below it can be changed as wished.

Glowbal:
What is the status on this @SilentSpike ?

SilentSpike:
It's done and waiting review :smile: (or if anyone wants to build on top of it, but so far nobody has spoken up).

jonpas:
Except my first comment above, it looks good to me.

commy2:
I'll merge this and other stuff after testing. Currently busy with BWMod though.

PR #2640, opened 2015-09-28
Add Extract Dependencies script

jonpas:
Extracts module dependencies to a file, for use in wiki.

Usage:
- Run the script
- Copy `temp\dependencies.md` to `_includes/` folder on `gh-pages` branch
- Watch the magic happen on all feature pages (include directive with Jekyll)


tools/extract_dependencies.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
+
+# Author: Jonpas
+
+import os
+import sys
+import re
+import shutil
```
jonpas: Remove

tools/extract_dependencies.py
```diff
@@ -0,0 +1,56 @@
...
+    # Iterate through folders in the addons directory
+    for folder in next(os.walk(addonspath))[1]:
+        writeToFile(folder)
+        print("Module: {}".format(folder))
+
+        # Open config.cpp file
+        with open(os.path.join(addonspath, folder, "config.cpp")) as file:
+            for line in file:
+                # Extract dependencies
+                match = re.match(r"\s+requiredAddons\[]\ = {(.+)};", line)
```
SilentSpike: What happens if the array spans multiple lines :wink:

jonpas: It breaks! :grin: 

Will try to fix (or call you), thanks!

jonpas: @SilentSpike I can't figure out how to do that.

jonpas: Hmh, I do it per-line, so I'll have to rework this a bit.

SilentSpike: Yeah, this is one of those situations where it'll be fine if all our addons use a one-line format. However a parser of some description is probably more suitable

jonpas:
This is finished and a PR for `gh-pages` also opened #3786 .

In the future I might add "used by <components>" if people think it would be useful.

Issue #2498, opened 2015-09-18
[Adv. Medical] Heal hitpoints doesn't work


Magirot:
ACE3 Version: 3.3.1.2

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* `Medical Settings (ACE) [Medical Level: Advanced, Medics setting: Advanced]`
* `Advanced Medical Settings (ACE) [Heal hitpoints: Yes]`

**Description:**
With the heal hitpoints parameter, bandaging wounds should heal the damage taken as well. This doesn't seem to work in the newest version.
Here's a video if it helps: https://youtu.be/5Xi2AGv3uCs

**Steps to reproduce:**
1. Add the aforementioned modules (or use an userconfig with the parameters).
2. Get shot and bandage yourself.
3. Still bloodied. :(

Pergor:
Additionally you still can't run after bandaging up some leg hits with ace_medical_healHitPointAfterAdvBandage == 1.

If I can recall it right, it's the third time this was supposed to be fixed.

Magirot:
Do you mean it didn't work in previous versions either? At least for us the damage got healed like it's supposed to in 3.2.1.

Pergor:
No, the leg damage remained and you could naught but walk after bandaging up the third leg wound, even in 3.2.1.

Glowbal:
Interesting. When I last tried this everything worked fine. I guess we need to take another look.

Pergor:
These are the settings I currently use:

class ace_medical_level {
    value = 2;
    typeName = "SCALAR";
    force = 1;
};
class ace_medical_medicSetting {
    value = 2;
    typeName = "SCALAR";
    force = 1;
};
class ace_medical_enableFor {
    value = 0;
    typeName = "SCALAR";
    force = 1;
};
class ace_medical_enableOverdosing {
    value = 1;
    typeName = "BOOL";
    force = 1;
};
class ace_medical_bleedingCoefficient {
    value = 0.3;
    typeName = "SCALAR";
    force = 1;
};
class ace_medical_painCoefficient {
    value = 1;
    typeName = "SCALAR";
    force = 1;
};
class ace_medical_enableAirway {
    value = 0;
    typeName = "BOOL";
    force = 1;
};
class ace_medical_enableFractures {
    value = 0;
    typeName = "BOOL";
    force = 1;
};
class ace_medical_enableAdvancedWounds {
    value = 1;
    typeName = "BOOL";
[...]

PR #2646, opened 2015-09-29
Fix healHitPointAfterAdvBandage - Issue #2498.

DerekSauer:
[Issue: Adv. Medical Heal hitpoints doesn't work #2498.](https://github.com/acemod/ACE3/issues/2498)

With "ace_medical_healHitPointAfterAdvBandage" enabled fully bandaging wounded body parts were not healing those body parts in the Arma engine. ACE's interface would show body parts fully bandaged and undamaged but the player's uniform would remain bloody and the effected parts would remain non-functional (E.g.: Even though ACE thinks your legs are healed, Arma doesn't and you can't run. Your pants are also full of blood).

This fix checks to see if an ACE body part is fully bandaged and then heals the respective Arma body part while keeping in mind that Arma does not have separate body parts for left & right arms and legs.

### DIFF
--- a/addons/medical/functions/fnc_treatmentAdvanced_bandageLocal.sqf
+++ b/addons/medical/functions/fnc_treatmentAdvanced_bandageLocal.sqf
@@ -102,7 +102,7 @@ if (GVAR(healHitPointAfterAdvBandage)) then {
     private["_currentWounds", "_headWounds", "_bodyWounds", "_legsWounds", "_armWounds"];
 
     // Get the list of the wounds the target is currently suffering from.
-    _currentWounds = GETVAR(_target, openWounds, []);
+    _currentWounds = _target getvariable [QGVAR(openWounds), []];
 
     // Tally of unbandaged wounds to each body part.
     _headWounds = 0;
--- a/addons/medical/functions/fnc_treatmentAdvanced_bandageLocal.sqf
+++ b/addons/medical/functions/fnc_treatmentAdvanced_bandageLocal.sqf
@@ -115,9 +115,9 @@ if (GVAR(healHitPointAfterAdvBandage)) then {
         _x params ["", "", "_bodyPart", "_numOpenWounds", "_bloodLoss"];
 
         // Use switch/case for early termination if wounded limb is found before all six are checked.
-        // Multiplying wounds by blood loss will either increment the wound count by
-        // some number (only care if greater than zero) or it will return zero indicating
-        // the body part is still undamaged.
+        // Number of wounds multiplied by blood loss will return zero for a fully
+        // bandaged body part, not incrementing the wound counter; or it will return
+        // some other number which will increment the wound counter. 
         switch (_bodyPart) do {
             // Head
             case 0: {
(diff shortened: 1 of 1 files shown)


bux578:
Thank you for your contribution! Could you please reformat your code to follow our [Coding Guidelines](http://ace3mod.com/wiki/development/coding-guidelines.html)?

Glowbal:
Thank you @DerekSauer 

This PR reintroduces the bug where injuries that does not cause bloodloss will prefent the heal hitpoints after bandaging working (the bruises bug).

To fix this, use both the injury amount and the bloodloss. Example:
```sqf
((_x select 4) * (_x select 3)) > 0
```

DerekSauer:
The code is now taking into account wounds that do not cause bleeding (bruises). Thanks for the heads up Glowbal, I would not have considered that. It should now also comply with ACE3's coding guidelines. Let me know if I missed any.

addons/medical/functions/fnc_treatmentAdvanced_bandageLocal.sqf
```diff
@@ -94,12 +94,73 @@ if (_impact > 0 && {GVAR(enableAdvancedWounds)}) then {
...
+         // Left Leg
+        if (_bodyPart == 4 && {(_numOpenWounds * _bloodLoss) > 0}) then {
+            _legsWounds = _legsWounds + 1;
+        };
+
+         // Right Leg
+        if (_bodyPart == 5 && {(_numOpenWounds * _bloodLoss) > 0}) then {
+            _legsWounds = _legsWounds + 1;
+        };
+    } foreach _currentWounds;
```
bux578: `forEach`

addons/medical/functions/fnc_treatmentAdvanced_bandageLocal.sqf
```diff
@@ -94,12 +94,73 @@ if (_impact > 0 && {GVAR(enableAdvancedWounds)}) then {
...
-    _target setHitPointDamage [_point, 0];
+// If all wounds to a body part have been bandaged, reset damage to that body part to zero
+// so that the body part functions normally and blood is removed from the uniform.
+// Arma combines left and right arms into a single body part (HitHands), same with left and right legs (HitLegs).
+// Arms are actually hands.
+if (GVAR(healHitPointAfterAdvBandage)) then {
+    private["_currentWounds", "_headWounds", "_bodyWounds", "_legsWounds", "_armWounds"];
+
+    // Get the list of the wounds the target is currently suffering from.
+    _currentWounds = _target getvariable [QGVAR(openWounds), []];
```
bux578: `getVariable`

addons/medical/functions/fnc_treatmentAdvanced_bandageLocal.sqf
```diff
@@ -94,12 +94,73 @@ if (_impact > 0 && {GVAR(enableAdvancedWounds)}) then {
...
+    _bodyWounds = 0;
+    _legsWounds = 0;
+    _armWounds  = 0;
+
+    // Loop through all current wounds and add up the number of unbandaged wounds on each body part.
+    {
+        _x params ["", "", "_bodyPart", "_numOpenWounds", "_bloodLoss"];
+
+        // Head
+        if (_bodyPart == 0 && {(_numOpenWounds * _bloodLoss) > 0}) then {
```
MikeMatrix: Create an early exit condition for `(_numOpenWounds * _bloodLoss) > 0` that way it only has to be evaluated 1 time instead of 6 times in the worst case, as well as skipping 6 value checks and making the code look cleaner.

Glowbal:
Thanks @DerekSauer :+1: 

Fix #2498

PR #2144, opened 2015-08-13
Code cleanup of Goggles module

jokoho48:
Code cleanup of Gforces module.

PR #2648, opened 2015-09-29
Code cleanup / refactoring Goggles

commy2:
ref: #2144


commy2:
- [ ] VR glasses

addons/goggles/functions/fnc_isDivingGoggles.sqf
```diff
@@ -1,22 +1,24 @@
...
-_result = [configFile >> "CfgGlasses" >> _glasses, configFile >> "CfgGlasses" >> "G_Diving"] call CBA_fnc_inheritsFrom;
-_result
+
+params ["_glasses"];
+
+local _config = configFile >> "CfgGlasses" >> _glasses;
+
+if (!isClass _config) exitWith {false};
+
+getNumber (_config >> "mode") == 1 // return
```
SilentSpike: Out of curiosity, what does the `"mode"` property represent?

commy2: `mode = 1;` is a config entry that changes the behaviour of goggles.
Usually goggles are hidden on the unit if it gets into the water even though they are equipped in that slot.
That is reversed when `mode` is set to 1. The goggles are only shown when in water. This is used for diving goggles in vanilla.

addons/goggles/functions/fnc_externalCamera.sqf
```diff
@@ -6,15 +6,13 @@
...
+ * call ace_goggles_fnc_externalCamera;
  *
  * Public: Yes
  */
 #include "script_component.hpp"
 
-if (GVAR(showInThirdPerson)) exitWith { false };
-
-(cameraView in ["EXTERNAL", "GROUP"] || {call EFUNC(common,isFeatureCameraActive)})
+cameraView in ["EXTERNAL", "GROUP"] || EFUNC(common,isFeatureCameraActive) // return
```
SilentSpike: Isn't the `call` required here?

SilentSpike: Ah wait, I see, it's automatically lazy evaluation

commy2: The implemenation of lazy evaluation is nothing more than the addition of the alternative syntax of `&&` and `||` which is `BOOL && CODE` etc.

commy2:
After some testing it seems that everything still works as expected. Merge?

PR #2650, opened 2015-09-29
Set Default Channel at mission start

PabstMirror:
From @NemesisRE

ACE_setting to set the marker channel at mission start


NemesisRE:
:+1:

licht-im-Norden87:
most likley the biggest improvement ace has to offer in PvP.

PR #2375, opened 2015-09-07
323mediccaching

commy2:
- [ ]  Leg Damage, Stuck Walking #2389 
- [ ]  [Medical] Players / AI wake up almost instantly after passing out #2494 
- [ ]  [Adv. Medical] Heal hitpoints doesn't work #2498 
- [ ]  Medical (3.3.2.0) - Multiple issues #2580 
- [ ]  Medical issues involving armor #2647 
- [ ]  Check CaManBase Hitpoints for 1.54 #2600 
- [ ]  Killed eventhandler is ran/triggered twice because of ACE3 #1859 
- [ ]  Advanced medical : crushing damage per car bug #2154

PR #2668, opened 2015-10-02
New method for handle damage

commy2:
replaces: #2375


addons/medical/XEH_postInit.sqf
```diff
@@ -2,6 +2,11 @@
 
 #include "script_component.hpp"
 
+["medical_selectionDamage", {_this call FUNC(handleSelectionDamage)}] call EFUNC(common,addEventhandler);
+["medical_fallDamage",      {_this call FUNC(handleFallDamage)}]      call EFUNC(common,addEventhandler);
+["medical_drowningDamage",  {_this call FUNC(handleDrowningDamage)}]  call EFUNC(common,addEventhandler);
+["medical_collisionDamage", {_this call FUNC(handleCollisionDamage)}] call EFUNC(common,addEventhandler);
```
MikeMatrix: Doing this might look good, but is actually bad practice to do with git in mind. Adding or changing one line and having to adjust other lines makes it so the chance of having merge conflicts is increased.

addons/medical/functions/fnc_handleFallDamage.sqf
```diff
@@ -0,0 +1,14 @@
...
+
+params ["_unit", "_newDamage"];
+
+private ["_totalDamageL", "_totalDamageR"];
+
+_totalDamageL = (_unit getHitPointDamage "HitLeftLeg") + _newDamage;
+_totalDamageR = (_unit getHitPointDamage "HitRightLeg") + _newDamage;
+
+_unit setHitPointDamage ["HitLeftLeg", _totalDamageL];
+_unit setHitPointDamage ["HitRightLeg", _totalDamageR];
```
DerekSauer: HitLeftLeg and HitRightLeg don't do anything as of Arma v1.53. Arma currently does not distinguish between Left and Right limbs but packs them into one "metalimb", HitLegs in this case. For the left and right arm it is HitHands. This will likely change in a future version if the dev team keeps their current plans.

jokoho48:
@commy2 PLEASE remove your Comment PLEASE else i could try to kill you

addons/medical/functions/fnc_handleFallDamage.sqf
commy2: Wrong. Those are custom hitpoints we add ourselves. Same as `HitLeftArm` and `HitRightArm`.

DerekSauer: That's true, those custom hitpoints are added to the Man class in Ace Medical's CfgVehicles.hpp, but those HitPoint serve no purpose in game. The Arma engine does not utilize them. The most they can do is serve as a fancy way of storing data in a character, similar to setVariable, which is perfectly fine for storing data to be used by the rest of ACE Medical; but causing damage to those HitPoints will not cause actual damage to a unit in game. 

You can test this by placing a player infantry unit in the editor and using the debug console to apply damage to _hitLeftLeg_ or _hitRightLeg_ and noting that the player is not inconvenienced at all and there is no outward appearance of damage. Damage _hitLegs_ and the player's movement will be restricted once damage exceeds 49% and their pants will become bloodied.

Glowbal:
Closing in favour of #3174

PR #2670, opened 2015-10-03
Simplify frag config

jokoho48:


jonpas:
Can't comment because the file is not shown, so here we go...
- Indent everything between L28 and L79
- Add a new line after L78
- Separate L3 into 2 lines and entire first define to proper indentation, like that:
```
#define ACE_EXPLOSION_REFLECTION(range, hit)\
	class ace_explosion_reflection_##range##_##hit : ace_explosion_reflection_base {\
		indirectHitRange = range;\
		indirectHit = hit;\
		dangerRadiusHit = range*3;\
		suppressionRadiusHit = range*2;\
	};
```

Issue #2672, opened 2015-10-03
Medicals given start at -1 minute


gienkov:
All medicals start at -1 minute, then go to 0 and 1 as seen below, build is e375dbe0.
![bug](http://images.akamai.steamusercontent.com/ugc/388790936733783541/8DC445316FAE741EA8EF21147076C4A600582238/)

jokoho48:
i realy do not think that i ever must do this but.
Please Read this: http://ace3mod.com/wiki/user/how-to-report-an-issue.html and edit your post so that we can work the the informations

PabstMirror:
QGVAR(triageCard) is using ACE_time, 

e.g. Client1: (ACE_time = 450, time = 60) Client2: (ACE_time = 200, time = 60)
Same instant, same game time, but wildly different ace_time.

I can look into fixing it

gienkov:
I've reported this before to @PabstMirror and he fixed many issues with
medical menu. This is another one @joko.
3 paź 2015 18:36 "PabstMirror" <notifications@github.com> napisał(a):

Glowbal:
Closed by #2674

PR #2674, opened 2015-10-03
Use ACE_gameTime for triage card

PabstMirror:
#2672

ACE_gameTime is just `time`, which should be synchronized


gienkov:
TY for this

Glowbal:
:+1: Nice @PabstMirror

PR #2679, opened 2015-10-03
fix missing magazines after certain scenarios

commy2:
fixes:
- respawn script adding additional weapon items if you were having an already equipped weapon class (similar to the prefilled backpack issue)

- certain scripts that add (launcher) magazines no longer worked, because BI changed the maximum load of "Bag_Base" to zero. (includes respawning with launcher magazine)

- compatibility with "binoculars" with magazines (custom laserrangefinders)

### DIFF
--- a/addons/common/CfgVehicles.hpp
+++ b/addons/common/CfgVehicles.hpp
@@ -140,4 +140,10 @@ class CfgVehicles {
         isBicycle = 1;
         XEH_DISABLED;
     };
+
+    class Bag_Base;
+    class ACE_FakeBackpack: Bag_Base {
+        scope = 1;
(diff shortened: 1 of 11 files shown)


commy2:
Also does:
- adds `FUNC(setAllGear)` as counterpart to `FUNC(getAllGear)` which seems pretty usefull for loadout scripts. It has no side effects that I am aware off and should add anything that can be added in normal gameplay too.

- adds FUNC(binocularMagazine) and FUNC(removeBinocularMagazine) functions which should be sripting commands, but aren't. `addBinocularMagazine` seems not possible for the same reason `addXYWeaponMagazine` commands are impossible.

bux578:
:+1:

PR #2680, opened 2015-10-04
Disable loading into locked vehicles.

### LINKED ISSUE #2664: Disallow loading cargo into locked vehicles
ACE3 Version: 3.3.2

Currently you cannot access the cargo space of locked vehicles (the whole interaction menu is disabled), but you can still load cargo into them. This could create unfortunate situations where you cannot unload something important you loaded into a locked vehicle, so it might be better to disable loading if the vehicle is locked?

PabstMirror:
Fix #2664


jonpas:
:+1:

PR #2681, opened 2015-10-04
Attach getIn/Out run on local unit

### LINKED ISSUE #2541: Attached items on a dedicated server
ACE3 Version: 3.3.1

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* None.

**Description:**
* When you enter a vehicle the attached item will detach but it will not disappear. It will drop on the ground and stay there. When you get out of the vehicle your attached item will be created on your shoulder but previous attached item will still stay where it detached. Happens only when playing on a dedicated server.

**Steps to reproduce:**
* Attach a chemlight to yourself.
* Enter a vehicle.
* Chemlight will detach but will not disappear and will stay on the ground.

* "Dedicated Server"

**RPT log file:**
* No need.

PabstMirror:
Fix #2541


Glowbal:
:+1:

PR #2683, opened 2015-10-04
Common cleanup x

PabstMirror:
Fix Nil Funcs (only real change is loadPerson)

Optimize 2 funcs called each frame (timePFH, getTurretIndex)


jokoho48:
why not useing ACE_LOG Macro?

PabstMirror:
@jokoho48 which line? I didn't think it made sense in fnc_dumpArray because it's more of a helper function that would be called from some other module, but I'm not sure.

MikeMatrix:
I don't see any part of there that should be replaced by the `ACE_LOG` macro

PR #2692, opened 2015-10-06
Run the travis deploy on master only

Glowbal:
- Enable sqf validation on all branches
- Have the deploy script only run on the master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,26 @@ branches:
     - master
 language: python
 python:
-  - "3.4"
+- '3.4'
 before_script:
-  - pip install pygithub
-  - pip install pygithub3
+- if [ -n "${GH_TOKEN}" ] && [ "${TRAVIS_BRANCH}" == "master" ] && [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
+    pip install pygithub;
+    pip install pygithub3;
+  fi
 script:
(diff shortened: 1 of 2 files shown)


jonpas:
Is there a way to limit it to this repo only as well? All forks try to deploy it but of course Travis returns domain already taken.

bux578:
there's the whitelist option http://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches

.travis.yml
```diff
@@ -3,13 +3,25 @@ branches:
     - master
 language: python
 python:
-  - "3.4"
+- '3.4'
 before_script:
-  - pip install pygithub
-  - pip install pygithub3
+- if [ ${TRAVIS_BRANCH} = "master" ]; then [ -z "${GH_TOKEN}" ] && exit 0; [ ${TRAVIS_PULL_REQUEST}
```
jonpas: Quotation marks around `${GH_TOKEN}` can be removed as well (works in my project using similar config).

.travis.yml
```diff
@@ -3,13 +3,25 @@ branches:
     - master
```
jonpas: Won't this make `sqf_validator.py` only ran on master as well? Or is that branch whitelist/blacklist different from PRs?

PR #2699, opened 2015-10-07
Disable nvgs in sight

BaerMitUmlaut:
This PR automatically disables the night vision goggles once you aim down the sight and re-enables them once you stop aiming. This can be disabled via a module and of course the ACE3 settings framework.

Requested in #414 by @Wishbone6061.


PabstMirror:
Acts weird with NVG optics (`optic_NVS`) or NV Binoculars (`laserDesignator`)
1. Remove NVG (`hmd player` = "")
2. Zoom into optic or laser des that has NV
3. Zoom out, NightVision persists in full screen. N key will not disable.

Oddly enough, dropping your vest/uniform will make the game correct itself.

Probably gets messy if a gun has multiple OpticsModes.  Is there a way to tell what mode you are in?

BaerMitUmlaut:
That's a really interesting BI bug! I've made a workaround so that it doesn't happen.
I've found a new bug now: when you don't wear your NVGs and aim down the sight with a night vision optic, and you stop aiming, you automatically start wearing your NVGs. I think I'll need to split the event handlers for that unfortunately.

I have also applied this to firing-from-vehicle cargo positions, which wasn't handled before.

commy2:
>Probably gets messy if a gun has multiple OpticsModes. Is there a way to tell what mode you are in?

Nope. It's only possible if you already know what OpticsModes your vehicle has (incl. FOV and visionMode).

See: https://git.koffeinflummi.de/bwmod-team/bwa3-public/blob/master/bwa3_tiger/functions/fn_drawDisplay_Tiger_Gunner.sqf#L103-116

```
_zoom = ([0.5,0.5] distance worldToScreen positionCameraToWorld [0,1.05,1]) * (getResolution select 5);

call {
    if (_zoom < 2) exitWith {
        _ctrlZoom ctrlSetText "\bwa3_tiger\data\recticles\bwa3_tiger_recticle_gunner_1.paa";
        _zoom = 1;
    };

    if (_zoom < 20) exitWith {
        _ctrlZoom ctrlSetText "\bwa3_tiger\data\recticles\bwa3_tiger_recticle_gunner_2.paa";
        _zoom = 2;
    };

    _ctrlZoom ctrlSetText "\bwa3_tiger\data\recticles\bwa3_tiger_recticle_gunner_3.paa";
    _zoom = 3;
};
```

It might be possible to apply this function to all vehicles with some config parsing, but I'm pretty sure it's possible to create a custom vehicle that easily breaks that system.

commy2:
I always thought `action "NVGogglesOff"` was completely broken. Seems like it kinda works with a lot of bugs now.
What about this?
```
_nvg = hmd _unit;
_unit unlinkItem _nvg;
_unit linkItem _nvg;
```

BaerMitUmlaut:
In this case, `"NVGoggles"` caused the error `"NVGogglesOff"` seems to work like a charm.

I have split up the event handlers now and haven't managed to break the system. Feel free to try for yourself.

addons/nightvision/functions/fnc_onVisionModeChanged.sqf
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Public: No
+ */
+
+#include "script_component.hpp"
+
+params ["_unit", "_visionMode"];
+
+if (GVAR(disableNVGsWithSights) && {(hmd _unit) != ""}) then {
+    (assignedVehicleRole _unit) params ["_role", "_turretPath"];
```
commy2: nope. `assignedVehicleRole` does not work as you expect when you have an AI group leader.
Use `[_unit] call EFUNC(common,getTurretIndex) in [vehicle _unit] EFUNC(common,getTurretsFFV)` or something like that.

addons/nightvision/functions/fnc_onCameraViewChanged.sqf
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Public: No
+ */
+
+#include "script_component.hpp"
+
+params ["_unit", "_cameraView"];
+
+if (GVAR(disableNVGsWithSights) && {(hmd _unit) != ""}) then {
+    (assignedVehicleRole _unit) params ["_role", "_turretPath"];
```
commy2: see below

commy2:
What if you die while looking through the optic? I think `GVAR(reenableNVGs)` does not work well as global variable. Probably should be an object namespace variable.

BaerMitUmlaut:
Now requires #2792.

addons/nightvision/CfgVehicles.hpp
```diff
@@ -14,4 +14,28 @@ class CfgVehicles {
...
+        functionPriority = 1;
+        isGlobal = 1;
+        isTriggerActivated = 0;
+        author = "BaerMitUmlaut";
+        class Arguments {
+            class disableNVGsWithSights {
+                displayName = CSTRING(DisableNVGsWithSights_DisplayName);
+                description = CSTRING(DisableNVGsWithSights_Description);
+                typeName = "BOOL";
+                defaultValue = 1;
```
commy2: 0

addons/nightvision/ACE_Settings.hpp
```diff
@@ -0,0 +1,8 @@
+class ACE_Settings {
+    class GVAR(disableNVGsWithSights) {
+        displayName = CSTRING(DisableNVGsWithSights_DisplayName);
+        description = CSTRING(DisableNVGsWithSights_description);
+        typeName = "BOOL";
+        value = 1;
```
commy2: 0

captainblaffer:
Is this feature added in the nightvision pbo? I personally find it very annoying gameplay wise and would like to disable it per default.o be disabled per default.

BaerMitUmlaut:
You can disable it per default serverside through the ACE settings framework if you want.

captainblaffer:
Thank you!

Issue #367, opened 2015-04-07
Translations


bux:
**[ACE3 Translation Guide](http://ace3.acemod.org/wiki/development/how-to-translate-ace3.html)**

Total number of keys: 4224

| Language | Missing Entries | Relevant Modules | % done |
|----------|----------------:|------------------|--------|
| English | 0 | - | 100% |
| Czech | 1182 | finger, missile_clgp, frag, medical_damage, interaction, noradio, minedetector, fire, vehicle_damage, logistics_rope, interact_menu, nametags, cargo, missile_gbu, hellfire, markinglaser, field_rations, scopes, medical_vitals, goggles, realisticnames, explosives, trenches, novehicleclanlogo, medical_status, dragging, hunterkiller, vehiclelock, flags, markers, advanced_throwing, gforces, overheating, missile_aim9, slideshow, milr, ai, viewdistance, aircraft, casings, medical_statemachine, hitreactions, fortify, medical, medical_ai, captives, intelitems, csw, ui, zeus, gunbag, irlight, medical_feedback, towing, marker_flags, rangecard, rearm, fastroping, map_gestures, missileguidance, missile_vikhr, medical_treatment, cookoff, missile_sdb, wardrobe, medical_engine, headless, reloadlaunchers, overpressure, viewports, missile_manpad, hearing, reload, fieldmanual, microdagr, arsenal, medical_gui, magazinerepack, ballistics, vehicles, common, nightvision, compat_sog, missile_sam, refuel, volume, maverick, grenades, parachute, laser, repair, tagging, missile_aim120, spike, killtracker, viewrestriction, maptools | 72.02% |
| French | 48 | minedetector, rearm, wardrobe, fieldmanual, medical_gui | 98.86% |
| Spanish | 16 | minedetector, milr, rearm, wardrobe, medical_gui | 99.62% |
| Italian | 116 | medical_damage, minedetector, cargo, realisticnames, markers, advanced_throwing, milr, intelitems, csw, ui, rangecard, rearm, missileguidance, cookoff, wardrobe, medical_gui, common, repair, tagging, spike | 97.25% |
[...]

PR #2707, opened 2015-10-08
Fix link and remove unneeded globals in deploy tool

jonpas:
Correct me if those globals are actually needed.


Glowbal:
The username and email are necessary for updating the stringtable issue (#367).

jonpas:
Will make a new PR as discussed on Slack.

PR #2713, opened 2015-10-08
Fix link in deploy tool

jonpas:
#2707


jonpas:
This issue is now equal to #2707, after further research... well those variables are really not needed, it was a funny day!

PR #2219, opened 2015-08-20
Hand signals enhancement 

jokoho48:
Port THS to ACE3 after #865

PR #2729, opened 2015-10-11
Setting to enable Gestures

PabstMirror:
from slack

orgoandy
>is there anyway to disable the hand signals? we're currently already using another hand signal addon


PabstMirror:
This probably isn't needed as it was requested to disable ace and use "THS" gestures; which is being merged in #2219

PR #2736, opened 2015-10-12
Map effects / flashlights for TrippleHead

### LINKED ISSUE #2734: Map darkness/Brightness does not work on three monitors
Pulling up the map during nighttime with 3 screens only makes the center monitor dark. The other ones are default brightness

PabstMirror:
Fix #2734

Should fix map darkness and map flashlights for trippleHead users.
For everyone else `safeZoneWAbs = safeZoneW` so it should have no effect.

Taging @voiperr as it messes with flashlight, but this should be good to go.


MikeMatrix:
Haven't tested it yet, but looks good :+1:

voiperr:
If it works :+1: I have no way to test triple monitor.

Issue #2706, opened 2015-10-08
Add support for privatizing variable with local keyword to tools


jonpas:
The following tools don't support `local` keyword yet (eg. `local _var = _x`):
- [ ] `search_privates.py`
- [ ] `search_unused_privates.py`

jonpas:
@PabstMirror Why closed? `search_unused_privates.py` still doesn't have local support.

PabstMirror:
if you do `local _test = 5`, it's kind of both defined and used

jonpas:
Ah, well that's true!

PR #2738, opened 2015-10-12
Local privates to search_privates.py

PabstMirror:
#2706


jonpas:
Very nice :+1:

BaerMitUmlaut:
The local keyword has been renamed to private in v1.53.132932, so this needs an update.

PabstMirror:
Tag #2755

Issue #2391, opened 2015-09-11
Sitting Modual- getting Blasted to the sky when elvation is above 0 on object


Shadowmarine578:
soooo found a bug with the siting module if a chair has a elevation any higher than 0 when you go to sit in the chair it will throw you up into the air and you will watch helplessly  as you fall to the ground.

I am placing chairs at elevation 17 on a LHD to build a briefing room. the chair is a simple camping chair Land_CampingChair_V1_F it only happens when the chair is set higher than 0 so maybe its something with the getpos on the script used

Glowbal:
Thank you for your report. We will look into it. Mainwhile could you please edit your ticket to follow the guidelines? http://ace3mod.com/wiki/user/how-to-report-an-issue.html This will help us adress the issue. Thanks!

commy2:
This most likely has nothing to do with the elevation, but with the first pathway lod below the chair being a PhysX GEO-Lod. I think I wrote a long explaination about this a while ago, but it basically is an engine bug / limitation / oversight with the PhysX implementation. The same issues occur with dragging (or would occur, I fixed them for the most part).
Basically the game thinks the PhysX chair collided with the PhysX building. AttachTo, which is used for sitting I belive is one of the triggers that cause a PhysX collision check / update / whatever.
The is a long, convoluted work around integrated in dragging using `allowDamage false` (which somehow influences PhysX, no idea it's somehow linked to collision damage), but I advice you to simply attach the chairs to the building.
tl-dr
PhysX sucks, nothing we can do, use `attachTo`

jonpas:
> AttachTo, which is used for sitting I belive

Nope, sitting uses simple setPos (not ASL, because it's relative to the model/chair) and setDir with animation change.

jonpas:
I've tested this with vanilla buildings, both those already on the map and manually placed, both work fine, this is a PhysX issue, and I don't want to add `allowDamage false` if it involves collision damage, as that would literally make people invulnerable on bigger chairs.

Shadowmarine578:
**Mods:**
* `@CBA_A3`
* `@ace`
* `@allinarmaterainpack`
* `@ATLAS_LHD_Plus`

**Modules:**
* *1.12 Make Unit Surrender:*
* *1.13 Map:*
* *1.14 MicroDAGR Map Fill:*
* *1.17 Rallypoint System:*
* *1.17 Respawn System:*
* *1.19 Sitting:*
* *2.1 Medical Settings:*


**Description:**
[...]

PR #2745, opened 2015-10-14
Use AGL for sitting setPos

PabstMirror:
#2391

Tag @jonpas

### DIFF
--- a/addons/sitting/functions/fnc_sit.sqf
+++ b/addons/sitting/functions/fnc_sit.sqf
@@ -58,8 +58,8 @@ _sitRotation = if (isNumber (_configFile >> QGVAR(sitRotation))) then {getNumber
 
 // Set direction and position
 _player setDir _sitDirection;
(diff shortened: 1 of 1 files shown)


jonpas:
What's the actual difference? Are there problems with how it was? Is the new not slower (not like it matters, just personal interest)?

PabstMirror:
https://community.bistudio.com/wiki/Position
modelToWorld returns AGL (which is the same as ASLW over water)

So on the carrier on utes, it's something like `[x,y,16.5]`
Doing `player setPos [x,y,16.5]` on the carrier, sticks you 16 meters up in the air.

I'm not entirely sure what format `setPos` uses? biki says AGL but I dont think that's right.
This explicitly converts AGL into ASL and then uses the setPosASL.

SilentSpike:
I think `setPos` is a kind of AGL, but if there's a building or something at the position it might not necessarily detect it, hence ATL.

jonpas:
Perfect. :grinning:

PR #2746, opened 2015-10-14
fix privates in Map module

jokoho48:


addons/map/functions/fnc_compileFlashlightMenu.sqf
```diff
@@ -25,12 +25,14 @@ _flashlights = [_player] call FUNC(getUnitFlashlights);
 
 //add all carried flashlight menus and on/off submenu actions
 {
-    _displayName = getText (configFile >> "CfgWeapons" >> _x >> "displayName");
-    _icon = getText (configFile >> "CfgWeapons" >> _x >> "picture");
+    private ["_cfg", "_displayName"," _icon", "_children", "_parentAction"];
```
PabstMirror: space before _icon, Fixed in #2748 which is a PR for this branch

PR #2748, opened 2015-10-14
Simplify Map Light Interactions

PabstMirror:
Currently each flashlight has 2 mutually exclusive child actions
This just shortens the depth of the action tree and puts the toggle action on the flashlight itself. 
(The 4th level can be hard to reach on some interface sizes)

@voiperr - If you were planning on adding other actions besides on/off or have any other objections to this please let me know.

before
a

after
bb

### DIFF
--- a/addons/map/functions/fnc_compileFlashlightMenu.sqf
+++ b/addons/map/functions/fnc_compileFlashlightMenu.sqf
@@ -30,7 +30,6 @@ _flashlights = [_player] call FUNC(getUnitFlashlights);
     _cfg = (configFile >> "CfgWeapons" >> _x);
     _displayName = getText (_cfg >> "displayName");
     _icon = getText (_cfg >> "picture");
-    _statement = {};
     
     _statement = if (GVAR(flashlightInUse) == _x) then {
         _displayName = format [localize LSTRING(turnLightOff), _displayName];
--- a/addons/map/functions/fnc_compileFlashlightMenu.sqf
+++ b/addons/map/functions/fnc_compileFlashlightMenu.sqf
@@ -18,47 +18,30 @@
 
 #include "script_component.hpp"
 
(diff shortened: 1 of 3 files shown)


jokoho48:
what about color switching? didnt have the KSF-1 3 color modes?

voiperr:
The intent was to eventually give colour modes (and filter switching) to the appropriate flashlights, but this was contingent on eventually getting the flashlights to be equippable as weapons.

This change is fine for now, just keep in mind it will be reverted if we ever implement equippable flashlights.

PabstMirror:
It is possible to add children but have the parent still have a working statement, like this:

image

voiperr:
I'm fine with that, but is it consistent with how the rest of ACE's actions work? The parent node never performs functions in Explosives, for instance.

jokoho48:
if i am right of cause the microDAGR have a sub action under configure if i am righ

jonpas:
> I'm fine with that, but is it consistent with how the rest of ACE's actions work? The parent node never performs functions in Explosives, for instance.

I think I saw plans to change that somewhere.

addons/map/functions/fnc_compileFlashlightMenu.sqf
```diff
@@ -18,47 +18,30 @@
...
-            {[""] call FUNC(switchFlashlight)},
-            {GVAR(flashlightInUse) == (_this select 2)},
-            {},
-            _flashlight
-        ] call EFUNC(interact_menu,createAction);
-
-        [[_onAction, [], _player], [_offAction, [], _player]]
+    _statement = {};
+    
+    _statement = if (GVAR(flashlightInUse) == _x) then {
```
MikeMatrix: This will always return something, so the previous `_statement = {};` initialization isn't required here.

commy2:
This is excellent. Simplify, simplify, simplify!

Issue #2757, opened 2015-10-16
add setAllowDamageStatus, getAllowDamageStatus functions


commy2:
-dragging fixFloating

PR #2230, opened 2015-08-22
Interaction with portable lamps

SzwedzikPL:
demo: https://www.youtube.com/watch?v=VsEnMXXO-pY :D

PR #2758, opened 2015-10-17
Allow damage handler (also block only collision damage)

PabstMirror:
#2757
Should fix issue from PR #2230

I'm not 100% sure if this is the right way to do things (like testing collision damage `_projectile == ""`)


Glowbal:
There is an `allowDamage` variable in `medical`.

commy2:
Do we need that handleDamage eventhandler? Those things don't stack well together, so we should use them only where it's absolutely necessary (once for medical).

addons/common/functions/fnc_setAllowDamage.sqf
```diff
@@ -0,0 +1,88 @@
...
+if ((_newBlockDamageNum == _blockDamageNum) && {_newBlockCollisionNum == _blockCollisionNum}) exitWith {
+    TRACE_2("No Change",_newBlockCollisionNum, _newBlockDamageNum);
+};
+
+TRACE_2("Updating",_newBlockCollisionNum, _newBlockDamageNum);
+_object setVariable [QGVAR(allowDamage), [_newBlockCollisionNum, _newBlockDamageNum], true];
+
+if ((_object isKindOf "CAManBase") && {["ace_medical"] call FUNC(isModLoaded)}) exitWith {
+    //ACE_medical already has a handleDamage installed on the Unit and we can just setVariables for it
+    TRACE_1("handled by ace medical", _object);
```
Glowbal:
```sqf
_object setvariable [QEGVAR(medical,allowdamage), false];
```

PR #2766, opened 2015-10-19
Add Error Checking to Settings + MapGesture Settings

PabstMirror:
This fixes an issue with nametags and hearing where there are settings that are clientSettable, but could be forced via a module. Modules run after GVAR(settings) is publicVariabled, so the force wasn't applied to clients.

* This moves the PV until after modules have run on the server.
* Adds a exclusive `isSingular = 1` to setting configs. A duplicate setting module would normaly have no effect, this just adds a warning.
* Adds warnings to `setSetting` for bad setting names, data types or other reasons it fails.
* @MikeMatrix - I cleaned up some map gesture stringtables, added exit code if the old pbo is already running and changed the logic on the color settings to be be optionally forced to clients.


MikeMatrix:
:+1:

PabstMirror:
Merged PR  #2826 into this branch - Read settings from paramsArray (params in description.ext)

esteldunedain:
:+1: Nice, I like #2826 and the singular flag.

Glowbal:
:+1:

captainblaffer:
Is this feature added in the nightvision pbo? I personally find it very annoying gameplay wise and would like to disable it.

PR #2767, opened 2015-10-19
Fix medical location check when set to disabled for PAK/Surgical

### LINKED ISSUE #2765: canTreat not working for disabled items
ACE3 Version: 3.3.2

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
In missionConfigFile >> "ACE_Settings":
class ace_medical_useLocation_PAK {
        typeName = "SCALAR";
        value = 4;
        values[] = {"Anywhere", "Medical vehicles", "Medical facility", "vehicle & facility", "Disabled"};
        force = 1;
 };

**Description:**
Let's assume that canTreat doesn't evaluate to false until line 78 in fnc_canTreat.sqf.
Then, the use location is checked. As _return is initialized to true, it won't be changed, if ace_medical_useLocation_PAK ist 4.
So: the item will be able to be used allthough it should be disabled.

PabstMirror:
Fix #2765

When set to 4 it should return false.

### DIFF
--- a/addons/medical/ACE_Settings.hpp
+++ b/addons/medical/ACE_Settings.hpp
@@ -154,13 +154,13 @@ class ACE_Settings {
         category = CSTRING(Category_Medical);
         typeName = "SCALAR";
         value = 3;
-        values[] = {"Anywhere", "Medical vehicles", "Medical facility", "vehicle & facility", "Disabled"};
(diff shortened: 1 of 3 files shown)


Glowbal:
The `treatment` function is missing this.

PR #2777, opened 2015-10-20
Interaction Menu Animation Setting And PreExpand First Level

PabstMirror:
Adds a setting to change the time needed to expand each level of the interaction tree.
Also pre-expands the first level for self/mounted/zeus as there is only one option (right now this is only if the setting is set to something besides normal, but could probably be default, I wasn't sure about changing default behavior)

Really makes the menu feel more responsive, especially if you are familer with the layout of actions.


Gwynblade:
These two are sorely needed, great to see it's already done!

But apart from expanding first level, could we also get rid of "Place" sub-menu of the explosives? 
Here's the picture: http://imgur.com/BTdPZyl

So under explosives, you just select which explosive you want, the "Place" menu is kinda unnecesary IMO.

PR #2778, opened 2015-10-21
Fix for missing module default defines

ArMaTeC:
noticed when loading the medical system that there where missing defines to the modules system that caused errors on the client when loaded on a server or locally

this could be a mod incompatibility but this will solve the issue any how

### DIFF
--- a/addons/captives/CfgVehicles.hpp
+++ b/addons/captives/CfgVehicles.hpp
@@ -197,6 +197,7 @@ class CfgVehicles {
                     class disable {
                         name = ECSTRING(common,No);
                         value = 0;
+						default = 0;
                     };
                     class Surrender  {
                         name = CSTRING(SurrenderOnly);
@@ -206,6 +207,7 @@ class CfgVehicles {
                     class SurrenderOrNoWeapon  {
                         name = CSTRING(SurrenderOrNoWeapon);
                         value = 2;
+						default = 0;
                     };
                 };
             };
--- a/addons/common/CfgVehicles.hpp
+++ b/addons/common/CfgVehicles.hpp
@@ -50,10 +50,12 @@ class CfgVehicles {
                     class Warn {
                         name = CSTRING(CheckPBO_Action_WarnPerm);
                         value = 1;
+						default = 0;
                     };
                     class Kick {
                         name = CSTRING(CheckPBO_Action_Kick);
                         value = 2;
+						default = 0;
                     };
                 };
             };
--- a/addons/hearing/CfgVehicles.hpp
+++ b/addons/hearing/CfgVehicles.hpp
@@ -122,11 +122,13 @@ class CfgVehicles {
                     class NotDisabled {
                         name = ECSTRING(common,No);
                         value = 0;
+						default = 0;
                     };
                     */
                     class IsDisabled {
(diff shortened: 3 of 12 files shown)


PabstMirror:
What was the error?

ArMaTeC:
6:33:22 Warning Message: No entry 'bin\config.bin/CfgVehicles/ACE_moduleMedicalMenuSettings/Arguments/allow/values/disable.default'.
 6:33:22 Warning Message: '/' is not a value
 6:33:22 Warning Message: No entry 'bin\config.bin/CfgVehicles/ACE_moduleMedicalMenuSettings/Arguments/allow/values/VehiclesOnly.default'.
 6:33:22 Warning Message: '/' is not a value

jonpas:
Looks like some other mod is trying to read the defaults of everything, without checking if it exists. I don't have this when running locally or on the server (checked client and server rpt logs).

ArMaTeC:
some times its best to fix the unknown then let it lay with errors its like a bucket full with holes works better when they are plugged

bux578:
Could you please try to figure out which mod is causing this? That would be good to know.

PabstMirror:
Yeah, a full RPT would be good to see. Even BIS configs don't have `default = 0;`.

Does putting down the chem light module (`ModuleChemlight_F`) throw an error ?

ArMaTeC:
no errors from default bis ones at all and the only other addon i use that has any ref to modules is tfar

Glowbal:
I'd like to know what is causing these errors. I do not like adding config values to fix some potential error without knowing why it is an error and what caused it.

Like @PabstMirror asked, could you please provide a full RPT log file with these issues occuring?

Also; the formatting of the `default` attributes you have added is off. It's mising a couple of spaces for most.

ArMaTeC:
http://pastebin.com/UGUC7bYH

sargken:
I also get these errors when going into the editor and starting up a missions. Or when playing online.

jonpas:
That's a lot of conflicting addons and updating of base classes for sure.

ArMaTeC:
yer that happens when you have a load of community addons

jonpas:
That happens when you have a load of incompatible mods using wrong inheritance. :)
But that's off-topic.

ArMaTeC:
yer i know im a solo modder so i do what i can do

commy2:
Shh everyone.

This issue is the same as :
https://github.com/KoffeinFlummi/AGM/issues/480
https://github.com/KoffeinFlummi/AGM/issues/481
https://github.com/KoffeinFlummi/AGM/issues/620

It is caused by any addon using class `default` as a base class (which is not valid). Same error will pop up on vanilla modules. This PR will not fix that issue. You should disable all addons and then re-enable them to figure out which one is causing it.

sargken:
So how can we fix this error?

alganthe:
Does massi, the author of mgsr and all of the other authors which content is in CL3 know that City life redistribute their content ?

ArMaTeC:
All content has original readmes and we do not redistribute as we are a private community and all content is from Armaholic that publicly available to all. We are not like other communities who steal content all attempts to contact original authors are taken

Issue #2126, opened 2015-08-12
Fire Control System is not working


MrFies:
ACE3 Version: 3.2.1

Mods:
@cba_a3
@ace3

Placed ACE3 Modules:
I changed nothing , I only used the unpacked @ace3 folder

Description:
This following Feature doesnt work.
http://ace3mod.com/wiki/feature/fcs.html 

Where did the issue occur?
"Multiplayer", "Singleplayer"

bux578:
https://www.youtube.com/watch?v=93rErjMNKjw

Can confirm!

hello12121212:
i found that it works in thermal but not in normal vision. Tested in helicopters.

bux578:
http://ace3mod.com/wiki/feature/fcs.html#2-1-engaging-moving-targets
Still doesn't work for tanks.

VKing6:
Fixed the lead issue, and I didn't notice any other problems when testing it, with any tank.

PR #2779, opened 2015-10-21
Fix lead calculation for vehicle FCS

VKing6:
Fix lead calculation aspect of #2126

### DIFF
--- a/addons/fcs/XEH_clientInit.sqf
+++ b/addons/fcs/XEH_clientInit.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 
 GVAR(enabled) = false;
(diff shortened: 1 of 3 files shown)


jokoho48:
there is the issue with regex :D

commy2:
Why next milestone? Should be merged right now.

VKing6:
Sure.

SilentSpike:
>there is the issue with regex :D

Nah, this is where regex could have prevented these edge cases :wink:

jokoho48:
yeah i am not so known how regex work but i think we now have 3 or 4 times the time issue :D

VKing6:
https://xkcd.com/208/

PR #2792, opened 2015-10-28
add FUNC(hasHatch)

commy2:


addons/common/functions/fnc_hasHatch.sqf
```diff
@@ -0,0 +1,35 @@
...
+ * Return Value:
+ * Unit has a hatch? <BOOL>
+ *
+ * Public: Yes
+ */
+#include "script_component.hpp"
+
+params ["_unit"];
+
+local _vehicle = vehicle _unit;
```
BaerMitUmlaut: `local`

addons/common/functions/fnc_hasHatch.sqf
```diff
@@ -0,0 +1,35 @@
...
+
+if (_unit == _vehicle) exitWith {false};
+
+local _config = configFile >> "CfgVehicles" >> typeOf _vehicle;
```
BaerMitUmlaut: `local`

addons/common/functions/fnc_hasHatch.sqf
commy2: It's the future.

addons/common/functions/fnc_hasHatch.sqf
```diff
@@ -0,0 +1,35 @@
+/*
+ * Author: commy2
+ * Check if unit has a hatch and can turn out.
```
Glowbal: A better description might be: 
> Check if unit in vehicle is in a vehicle position where unit can turn out.

addons/common/functions/fnc_hasHatch.sqf
BaerMitUmlaut: Not for long!

addons/common/functions/fnc_hasHatch.sqf
BaerMitUmlaut: Might as well call it canTurnOut then? Or does it return something else?

addons/common/functions/fnc_hasHatch.sqf
jonpas: There is regex ready to change it, it's still `local` until 1.54 gets rolled and that won't happen for at least 2 more weeks.

PR #2798, opened 2015-10-28
Fix broken repair point selection code

SilentSpike:
Only valid hit selection + hit point combinations are usable for repair points:
- Empty hit points mean there isn't enough information to display to the user
- Empty hit selections mean that the hit point doesn't actually exist

Fixes the random repair actions with no text that didn't work as well as the lack of actions for valid repair points on some vehicles.

This also removes the restriction on glass hit points since 3rd party content doesn't necessarily use the same names. However the amount of glass points is kind of obnoxious so maybe that should be added back in - plus they don't seem like a priority vehicle part and takes up some screen space.

Tried to make the code a bit more readable, but it looks like the repair stuff is kind of messy as is. Should probably go through and make the functions consistently use one of `hit selections/hit points/hit indexes` at some point as it's currently a mix. Indexes probably makes the most sense.

### DIFF
--- a/addons/repair/functions/fnc_addRepairActions.sqf
+++ b/addons/repair/functions/fnc_addRepairActions.sqf
@@ -1,5 +1,5 @@
 /*
- * Author: commy2
+ * Author: commy2, SilentSpike
  * Checks if the vehicles class already has the actions initialized, otherwise add all available repair options. Calleed from init EH.
  *
  * Arguments:
@@ -19,7 +19,7 @@
 params ["_vehicle"];
 TRACE_1("params", _vehicle);
 
-private ["_type", "_initializedClasses", "_condition", "_statement", "_action", "_duplicateHitpointName", "_processedHitPoints", "_selectionName", "_hitPointsAddedNames", "_hitPointsAddedStrings", "_hitPointsAddedAmount"];
+private ["_type", "_initializedClasses", "_condition", "_statement", "_action", "_processedHitPoints", "_hitPointsAddedNames", "_hitPointsAddedStrings", "_hitPointsAddedAmount"];
 
 _type = typeOf _vehicle;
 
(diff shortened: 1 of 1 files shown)


SilentSpike:
Oh, I almost forgot these too.

Before:


After:

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -29,154 +29,123 @@ _initializedClasses = GETMVAR(GVAR(initializedClasses),[]);
...
                 };
-            } forEach (getArray _customSelectionsConfig);
+            } forEach (getArray _positionsConfig);
         };
 
-        // If position still empty (not a position array or selection name) try extracting from model
-        if (typeName _selection == "STRING" && {_selection == ""}) then {
-            _selection = _vehicle selectionPosition _selectionName;
-        };
+        // Prepair the repair action
```
jonpas: `Prepare`

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -29,154 +29,123 @@ _initializedClasses = GETMVAR(GVAR(initializedClasses),[]);
...
-        // Get hitpoint groups if available
-        _hitpointGroupConfig = configFile >> "CfgVehicles" >> _type >> QGVAR(hitpointGroups);
-        _inHitpointSubGroup = false;
-        if (isArray _hitpointGroupConfig) then {
-            // Set variable if current hitpoint is in a sub-group (to be excluded from adding action)
-            _currentHitpoint = _x;
+        // Empty selections don't exist
+        // Empty hitpoints don't contain enough information
+        if (_selection isEqualTo "") exitWith {};
+        if (_hitpoint isEqualTo "") exitWith {};
```
jonpas: `if (_selection isEqualTo "" || {_hitpoint isEqualTo ""}) exitWith {};`?

SilentSpike:
Oh, this also changes the `selectionPosition` command to use the `"HitPoints"` LOD for better placement :smile:

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -29,154 +29,123 @@ _initializedClasses = GETMVAR(GVAR(initializedClasses),[]);
...
                 } forEach (_x select 1);
-            } forEach (getArray _hitpointGroupConfig);
+            } forEach (getArray _groupsConfig);
         };
-
-        // Exit if current hitpoint is in sub-group (only main hitpoints get actions)
-        if (_inHitpointSubGroup) exitWith {TRACE_3("inHitpointSubGroup",_x,_forEachIndex,_selectionName);};
-
-        // exit if the hitpoint is virtual
-        if (isText (configFile >> "CfgVehicles" >> _type >> "HitPoints" >> _x >> "depends")) exitWith {TRACE_3("Depends",_x,_forEachIndex,_selectionName);};
```
PabstMirror: Are you sure it's ok remove this check for `depends` hitpoints?

SilentSpike: Not entirely, however if a selection exists in the model and has a respective hit point then I'm presuming it's not "virtual" since it exists in the model to being with. :stuck_out_tongue:

SilentSpike: Someone who knows more about what that property represents will have to chime in

jonpas: @commy2 will probably know.

jonpas:
> However the amount of glass points is kind of obnoxious so maybe that should be added back in - plus they don't seem like a priority vehicle part and takes up some screen space.

Group them together for now I'd say, it'll be some work, but if you are going through all the vanilla vehicles already then it'd be the best timing. I think having a framework for "repairable in facility only" and "repairable in facility or repair vehicle only" would be neat, could just work through config and by default have no impact (as there is no way to check reliable what is a glass and what not for example). Having them grouped together would literally just mean changing the config entry name to whatever the repair location requirement framework would provide. This is more of a note and thought for expansion.

PR #2801, opened 2015-10-30
Remove manual team synching

BaerMitUmlaut:
This PR removes all occurences of `assignTeam` since this will be handled by the next version of CBA, see CBATeam/CBA_A3#185. The call of the ACE event has also been replaced with the CBA event.


jonpas:
That PR has some very unnecessary merge commits.

PR #2805, opened 2015-10-30
Cleanup Macro Definitions

jonpas:
Removed macros that are now in CBA (stable and released), unused macros (`EXPLODE_n`) and some redundant definitions. Also made debug ifs nicer.

### DIFF
--- a/addons/main/script_macros.hpp
+++ b/addons/main/script_macros.hpp
@@ -1,16 +1,8 @@
 #include "\x\cba\addons\main\script_macros_common.hpp"
 #include "\x\cba\addons\xeh\script_xeh.hpp"
 
-//Faster Array Unwraping (skips the IS_ARRAY check normaly found in EXPLODE_1_SYS)
-#undef EXPLODE_2_SYS
-#define EXPLODE_1_SYS_FAST(ARRAY,A) A =(ARRAY) select 0
-#define EXPLODE_2_SYS(ARRAY,A,B) EXPLODE_1_SYS_FAST(ARRAY,A); B = (ARRAY) select 1
-
 // Default versioning level
 #define DEFAULT_VERSIONING_LEVEL 2
-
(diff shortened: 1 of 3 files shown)


addons/main/script_macros.hpp
```diff
@@ -1,17 +1,9 @@
 #include "\x\cba\addons\main\script_macros_common.hpp"
 #include "\x\cba\addons\xeh\script_xeh.hpp"
 
-//Faster Array Unwraping (skips the IS_ARRAY check normaly found in EXPLODE_1_SYS)
-#undef EXPLODE_2_SYS
-#define EXPLODE_1_SYS_FAST(ARRAY,A) A =(ARRAY) select 0
-#define EXPLODE_2_SYS(ARRAY,A,B) EXPLODE_1_SYS_FAST(ARRAY,A); B = (ARRAY) select 1
-
 // Default versioning level
 #define DEFAULT_VERSIONING_LEVEL 2
```
jonpas: Why exactly is this here?

PabstMirror: Might be leftover from ACE2, doesn't seem to be used anywhere.

PabstMirror:
:+1: This should be good to go.
Just a note, people may need to update their p:\x\cba files if they haven't been updated in a long time.

jonpas:
I will update those files as well in this PR then, it's been quite some time since they were updated last.

addons/main/script_macros.hpp
jonpas: Yeah, I did a full search, found nowhere, I thought it was used by some CBA, but doesn't appear to be.

PabstMirror:
Updated CBA in https://github.com/acemod/ACE3/commit/af5b646adb29cfec7e3dc45448b3f94a0f363c24

jonpas:
:+1:

PR #2813, opened 2015-11-01
Add method to deal with new hitpoints

Glowbal:
This addition tries to deal with the new hitpoints introduced in Arma3 v1.50 and above. It converts new selection names to by the current medical system usable selection names.

This is only a temporarily patch, while we are still working on a larger overhaul to account both the new hitpoints and any potential new features / polishing.


Glowbal:
Tested it - this solves the issues quite well.

PR #2826, opened 2015-11-07
Read settings from paramsArray (params in description.ext)

PabstMirror:
Just an idea, not sure if we need
Allows you to make settings changeable by mission params, so one mission can be played with different settings

for now just limited to SCALAR/BOOL

e.g.
```
class Params {
    class ace_medical_level {
        title = "Medical Level";
        ACE_setting = 1;
        values[] = {1,2};
        texts[] = {"Basic", "Advanced"};
        default = 2;
    };
```

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -194,6 +194,9 @@ call FUNC(checkFiles);
 
     ACE_LOGINFO("Settings received from server.");
 
+    if (isServer) then { //read settings from paramsArray
+        [] call FUNC(readSettingsFromParamsArray);
+    };
     // Event so that ACE_Modules have their settings loaded:
     ["InitSettingsFromModules", []] call FUNC(localEvent);
 
--- a/addons/common/XEH_preInit.sqf
+++ b/addons/common/XEH_preInit.sqf
@@ -139,6 +139,7 @@ PREP(playerSide);
 PREP(positionToASL);
 PREP(progressBar);
 PREP(readSettingFromModule);
+PREP(readSettingsFromParamsArray);
 PREP(receiveRequest);
 PREP(removeCanInteractWithCondition);
 PREP(removeSpecificMagazine);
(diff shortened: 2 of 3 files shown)


jonpas:
Nice idea!

TheMagnetar:
I have this in my mission template. Essentially one must call this at postInit. Here the function where I set the different parameters (for the moment only Basic medical, advanced medical and advanced ballistics on/of without requiring any of the modules/missionconfig.cpp files. Everything could be dynamically set at mission start.

https://github.com/CavallersDelCel/1RA_Plantilla_Missio/blob/master/scripts/ace3/fn_ace3_config.sqf

SilentSpike:
Shouldn't it happen *after* modules? Good idea though :+1:

Glowbal:
@SilentSpike no, the settings here are forced. Anything coming after it will be ignored.

SilentSpike:
[Are you sure?](https://github.com/acemod/ACE3/pull/2826/files#diff-197da1aa4041e4449fd4e7bf0e4db3e4R29)

 It looks as though they're only forced if said so in the config, but since they're parameters they should probably take precedence over modules

Glowbal:
Ah, I was looking at the true on the `setSettings`. Wasn't paying attention. My mistake, sorry.

In that case; they are not forced. But I see no issue with them being executed before modules.

SilentSpike:
Well, if somene places a module to change a bunch of settings, but wants one of the settings handled by the module to be handled by parameter instead then the module will override it

PabstMirror:
Modules will overwrite and force all settings in the module (unless they are already forced)
- This needs to run before modules.
- You really need the force flag (maybe this should force by default?) otherwise the module will just overwrite it right after.

SilentSpike:
Yeah, I'm thinking this should just force by default :+1:

Feivel69:
Is parameters > modules still working? I seem to remember it did work before the last update, but now the modules settings are taken instead of the defaults in description.ext. Or is it just me?

Glowbal:
What do you mean with that? 

All previous options from the ACE settings framework still work.

Feivel69:
If I set up description.ext with parameters, for example ace_medical_level with default 1 and ace_medical_medicSetting with default 1 and then place the medical module in editor with both values set to 2, then the mission will take the values from module and not from description.ext. And I seem to remember, that a few days/weeks earlier the description.ext default values had highest priority.

Feivel69:
Ok, nevermind, I see, we have to use ACE_setting = 1 now. I'm sorry, thanks.

PR #2839, opened 2015-11-11
AB - Fix ACE_twistDirection

PabstMirror:
Most guns don't have a `ACE_twistDirection` and they where defaulting to 0 instead of 1.

### DIFF
--- a/addons/advanced_ballistics/functions/fnc_readWeaponDataFromConfig.sqf
+++ b/addons/advanced_ballistics/functions/fnc_readWeaponDataFromConfig.sqf
@@ -20,9 +20,11 @@ _weaponConfig = (configFile >> "CfgWeapons" >> _this);
 
 _barrelTwist = getNumber(_weaponConfig >> "ACE_barrelTwist");
 _twistDirection = 1;
-_twistDirection = getNumber(_weaponConfig >> "ACE_twistDirection");
-if !(_twistDirection in [-1, 0, 1]) then {
-    _twistDirection = 1;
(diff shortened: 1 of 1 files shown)


addons/advanced_ballistics/functions/fnc_readWeaponDataFromConfig.sqf
```diff
@@ -20,9 +20,11 @@ _weaponConfig = (configFile >> "CfgWeapons" >> _this);
 
 _barrelTwist = getNumber(_weaponConfig >> "ACE_barrelTwist");
 _twistDirection = 1;
-_twistDirection = getNumber(_weaponConfig >> "ACE_twistDirection");
-if !(_twistDirection in [-1, 0, 1]) then {
-    _twistDirection = 1;
+if (isNumber (_weaponConfig >> "ACE_twistDirection")) then {
+    _twistDirection = getNumber (_weaponConfig >> "ACE_twistDirection");
```
jonpas: getNumber returns 0 and if the value is not setyso I think it's safe to remove isNumber check in this case.

PabstMirror: -1, 0 and 1 are valid values for `ACE_twistDirection`
If no such config entry exists, it needs to be 1 not 0, so we need the isNumber check.

jonpas: Ah I see now.

PR #2093, opened 2015-08-09
Code cleanup of FCS module

jokoho48:
Code cleanup of FCS module.

PR #2842, opened 2015-11-13
Update FCS display after distance adjust

### LINKED ISSUE #2750: FCS for 35 mm AA gun not handling elevation correctly, cant adjust distance for Air-Burst ammo
FCS for the 35 mm Anti-Air Bardelas (Cheetah/Tigris) is not handling elevation correctly, currently you cant adjust distance for Air-Burst ammo without direct line of sight to target.

To replicate:
In Editor, place 35 mm AA unit on map and play as gunner. Also spawn fixed wing flying aerial unit close by.

Try to range and adjust elevation to fixed wing aircraft target against the sky.

PabstMirror:
Fix #2750

KeyUp can take a specific distance as 3rd arg when using pageUp/Down.
The hud's distance was only set in getRange with the rangefinder's raw text

### DIFF
--- a/addons/fcs/functions/fnc_keyUp.sqf
+++ b/addons/fcs/functions/fnc_keyUp.sqf
@@ -210,3 +210,11 @@ if(_playSound) then {
 if(_showHint) then {
     [format ["%1: %2", localize LSTRING(ZeroedTo), _distance]] call EFUNC(common,displayTextStructured);
 };
+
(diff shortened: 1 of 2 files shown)


commy2:
ref: https://github.com/acemod/ACE3/pull/2093

commy2:
ref: #2750

PR #2846, opened 2015-11-16
Improved revive system for advanced medical

BaerMitUmlaut:
This adds the possibility to get people out of the revive state through CPR or the use of a defibrillator. Both options can be disabled through the revive module.

The defibrillator has to be properly deployed and prepared first before it can be used. This way the patients get brought to the defibrillator and it becomes semi-stationary.

This PR includes a prototype model that is only used for demonstration purposes right now. You cannot use setVariable on it which is why I set the object texture instead.

TODO:
- [ ] Replace prototype 3D model
- [ ] Clean up some statements and conditions into external functions
- [ ] Possibly add sounds, heart rate display?

*This is more of a prototype as a base for discussion.*


Glowbal:
Could you please provide some details on what the issue is you try to adress and why you picked both CPR and a new item (Defibrillator) to do so?

BaerMitUmlaut:
As of right now the revive state is quite unintuitive in advanced medical. It's not very visible for the medic and it's hard to understand what it does for somebody who has not taken a deep look into the code or read a good explanation somewhere (which is probably the case for a lot of players). This goes that far that many people simply think it's a bug which has occured both on slack and on the issue tracker.

With prevent instant death enabled there still happen a lot of sudden deaths, because not all damage happens within a few frames - a burst from a machine gun or the splitters of a frag grenade might hit you over a longer period of time than what ACE thinks is an "instant". This is why a lot of groups enable the revive system. However, with only being able to get people out of the revive state with a personal aid kit you also fully heal the player, which is often not desired.

This is where the defibrillator comes in. For a harder setting, you could only give one to the platoon medic or a medevac unit. He would be able to save people that got into the revive state, but still have to fix their wounds. For an easier setting, you could assign one to every squad or so. And if you want, you can even allow CPR to get people out of the revive state. People would still have to deal with all wounds and the blood loss, so it's not making advanced medical too "casual" in my opinion.

SilentSpike:
I have some personal notes for the upcoming review of ace medical, which include suggestions on changes to the "revive" system :+1: We probably want to hold off on this PR until that process starts

BaerMitUmlaut:
Yep, I'm aware, just see this as a *"this is how it could work"*-PR :smiley:

jokoho48:
:-1: sorry but i dont like it because you dont use a defibrillator if someone have a asystole you only use a defibrillator if someone have ventricular fibrillation this is more unreal than you think.

BaerMitUmlaut:
I'm aware that this isn't the most realistic solution, but a PAK is far worse in terms of realism (and currently the only option for people in the revive state). I mean the whole revive system is not realistic. It's more of a gameplay mechanic.

If a group wants to play on very realistic settings they don't have to use the defibrillator.

Glowbal:
Is there an alternative item we could use for this instead?

jokoho48:
Basicly you use CPR and medication like Epi and Morphine to restart the heart

BaerMitUmlaut:
Which is possible with this - just don't assign defibrillators.

jokoho48:
i think defibrillators distroy the complett reality ambitions medical defibrillators more fit in basic medical than in advanced

Glowbal:
My issue is if we ever do want to apply defibrillators for their actual purpose, I rather not be stuck with the item already in use for something else. Sure we could change it, but that might cause confusion. So I'd rather avoid it all together and come up with a good solution.

How about a setting for the PAK to control the full heal functionality? In the field, Med facility, vehicle or both facility and vehicle?

SilentSpike:
Here are part of my suggestion notes (intended for internal discussion and development, very bare-bones) for what it's worth:

>- Strip out "revive" state, replace with a minUnconciousTime
  - If unit "dies" before minUnconciousTime has passed, enters cardiac arrest
- Add setting to tweak behaviour of CPR
  - Can rescue from cardiac arrest
  - Extend bleedout timer

It removes both diversifies the cardiac arrest state while simultaneously removing the need for the ambiguous "revive" state

BaerMitUmlaut:
Making CPR get people out of the revive state makes it a bit too easy in my opinion. Anybody can do CPR, but getting somebody out of the revive state should be limited to medics (or even better, only specific medics/doctors).

Changing the functionality of the PAK sounds a bit better, but still isn't ideal as it keeps being a magical item that way.

Glowbal:
We can probably make it stabalize in the field? That is an action from ACE2. 

Have it move a unit from revive to cardiac arrest. That way a medic can attempt to get a hold of the vitals again and it is not an instant action either.

SilentSpike:
@BaerMitUmlaut That's why my notes specify introducing a setting to allow players to decide how powerful CPR is :+1:

BaerMitUmlaut:
What do you mean by how powerful? Because everybody should be able to do CPR. Even if you set the chance for success to very low, medics still become a less important role than they are now.

SilentSpike:
"how powerful" = What it actually results in

Was in response to your comment:

jokoho48:
CPR is the only Option to become Someone out of a asystole. And everything else is unreal and not realty close. we could take a look to that a Medic can perform better CPR and have a better change to revive.

addons/medical/data/model.cfg
```diff
@@ -17,8 +17,14 @@ class CfgModels {
...
     };
-	
+
 	class IVBag_250ml: IVBagBase {};
 	class IVBag_500ml: IVBagBase {};
 	class IVBag_1000ml: IVBagBase {};
-};
\ No newline at end of file
+
+    class defibrillator: Default {
```
jokoho48: ? what did you plan with that?

addons/medical/functions/fnc_canDefibrillate.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * Public: No
+ */
+
+#include "script_component.hpp"
+params ["_caller", "_target"];
+private ["_defibrillator"];
+
+_defibrillator = nearestObject [_caller, "ACE_defibrillator_static"];
+
+!(isNull _defibrillator || {(_defibrillator distance _target) > 5} || {((getObjectTextures _defibrillator) select 0) != "#(rgb,8,8,3)color(0.4,0,0,1)"})
```
jokoho48: ((getObjectTextures _defibrillator) select 0) != "#(rgb,8,8,3)color(0.4,0,0,1)" ? what did you do there?

addons/medical/functions/fnc_treatmentAdvanced_CPRLocal.sqf
```diff
@@ -18,16 +18,24 @@ private "_reviveStartTime";
 params ["_caller","_target"];
 
 if (_target getvariable [QGVAR(inReviveState), false]) then {
-    _reviveStartTime = _target getvariable [QGVAR(reviveStartTime),0];
-    if (_reviveStartTime > 0) then {
-        _target setvariable [QGVAR(reviveStartTime), (_reviveStartTime + random(20)) min ACE_time];
+    if (GVAR(exitReviveBy) == 2 && {(random 1) >= 0.3}) then {
+        //Lower chance of success in revive state
+        _target setvariable [QGVAR(inReviveState), false, true];
```
jokoho48: capitalization

addons/medical/data/model.cfg
BaerMitUmlaut: See my comment about it in the OP.

addons/medical/functions/fnc_canDefibrillate.sqf
BaerMitUmlaut: Again, see my comment about it in the OP.

BaerMitUmlaut:
Closing this in favour of #3134.

PR #2852, opened 2015-11-19
make.py - update for new mikros tools

PabstMirror:


jonpas:
Copy of the changes recently introduced in CBA's make.py?

PabstMirror:
yup, same as https://github.com/CBATeam/CBA_A3/pull/193

jonpas:
:+1:

PabstMirror:
For reference, people will need to update their mikero tools if they are still using an older version:

https://dev.withsix.com/projects/mikero-pbodll/files

PR #2856, opened 2015-11-20
1.54 - Replace typeName with isEqualType

PabstMirror:
Use new isEqualType in 1.54

About twice as fast (on something that's so fast it doesn't really matter), 
0.051416 ms vs 0.0234131 ms on this test: https://gist.github.com/PabstMirror/7277a65d6c7f2564a861


MikeMatrix:
Nice :+1:

jokoho48:
:+1:

addons/vector/functions/fnc_convertToTexturesFOS.sqf
```diff
@@ -237,7 +237,7 @@ switch (_coordinate) do {
 };
 
 // trap errors
-if ({typeName _x == "STRING"} count [_digit0, _digit1, _digit2, _digit3, _digit4] < 5) exitWith {
+if ({_x isEqualType ""} count [_digit0, _digit1, _digit2, _digit3, _digit4] < 5) exitWith {
```
SilentSpike: This would could be replaced with `isEqualTypeAll`

PR #2859, opened 2015-11-20
Addition of compat_optre into optionals folder.

Nathan423:
Initial commit for compat_optre pbo


PabstMirror:
I noticed HALO guys couldn't take damage on arms/legs.
The core problem is they don't inherit class Hitpoints from CaManBase, so they never get the Limb hitpoints we add and need,  They just have the 4 old hitpoints instead of all the new 1.52 hitpoints.

`class HitLeftLeg { name="left leg"; }`
The selection name seems odd, but we should be able to match the hitpoint name.

Sgt-Parker:
I thought I added that to the "base" units that did not have them within cfgVehicles.hpp. That was an issue that I brought up to the OPTRE dev team and they will try to remedy the issue. It will be in here until then.

optionals/compat_optre/ACE_Medical_Treatments.hpp
```diff
@@ -0,0 +1,56 @@
+
+class ACE_Medical_Actions {
+    class Basic {
+        class Bandage;
+        class Biofoam: Bandage {
+            displayName = "Biofoam";
+            displayNameProgress = "Applying Biofoam...";
```
gienkov: Could you stringtable that?

Sgt-Parker:
https://github.com/acemod/ACE3/pull/2925

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -32,8 +32,6 @@ if (_type in _initializedClasses) exitWith {};
 // get all hitpoints and selections
 (getAllHitPointsDamage _vehicle) params [["_hitPoints", []], ["_hitSelections", []]];
 
-if (_hitSelections isEqualTo []) exitWith { ACE_LOGERROR_1("No hit selections (%1)", _type); };
```
Glowbal: Why is this line removed in this PR?

optionals/compat_optre/cfgVehicles.hpp
```diff
@@ -0,0 +1,503 @@
+class CfgVehicles {
+	class Man;
+    class CAManBase: Man
+    {
+        class ACE_SelfActions
+        {
```
jonpas: Opening brace on the same line as `class`.

gienkov: Why is this even added to optionals? Why it can't be added into said add-on?
24.02.2016 2:38 PM "jonpas" <notifications@github.com> napisał(a):

kripto202:
what still needs to be done to get this released?

Glowbal:
There have been some comments made on this PR. Once they have been resolved (answered or necessary changes made), we can include the optionals.

Issue #2742, opened 2015-10-13
Wheel repair doesn't work correctly on RHS HMMWV


MAD-BUG:
Repairing of HMMWV wheels work like they are standart parts of car. You can only repair them with toolkit without consumpting spare wheels.

jokoho48:
this is a RHS Fault that we cant fix

MAD-BUG:
can you say in which place there is a fault?

PabstMirror:
RHS configs are very odd but I can't say for certain that they are wrong.

This wheel:
```
    class LR: LF {
        boneName = "wheel_1_2_damper";
        center = "axis_wheel_3_1";
    };
```

matches with this hitpoint:
```
class HitLF2Wheel {
    name = "wheel_3";
    visual = "wheel_3_tire";
};
```
Looks like they mix `wheel_X_Y` with `wheel_N`.

I added a backup section to handle this and search for the closest hitpoint to the wheel's center.

sargken:
Anyone Report this to RHS?

MAD-BUG:
This was writen before i get confirmation about this here. http://feedback.rhsmods.org/view.php?id=1280#c2578

sargken:
So what is the problem we are facing. What needs to be changed? on RHS side.

PabstMirror:
Fixed via #2861 

The way everyother car does it is to have the hitpoint selection be a varient of the wheel's boneName. So boneName `wheel_1_1_damper` can be matched with selection `wheel_1_1_steering`.

RHS's HMMVWES have wheel selections `["wheel_1", "wheel_2", "wheel_3", "wheel_4"]`, which isn't "wrong", but is non-standard.

```
//Normal Car:
_wheelName=LF, _wheelCenter=wheel_1_1_axis, _wheelBone=wheel_1_1_damper, _wheelBoneNameResized=wheel_1_1
//will match:
_wheelHitPoint=HitLFWheel,  _wheelHitPointSelection=wheel_1_1_steering

//RHS HMMV:
_wheelName=LF, _wheelCenter=axis_wheel_1_1, _wheelBone=wheel_1_1_damper, _wheelBoneNameResized=wheel_1_1
//will match:
_wheelHitPoint=HitLFWheel,  _wheelHitPointSelection=wheel_1
```

Issue #2643, opened 2015-09-28
Can't repair car engine/body because of script error


Dystopian:
ACE3 Version: 3.3.2

**Steps to reproduce:**
1. damage car engine/body
2. try to repair it
3. nothing happens and there is "Undefined var" script error in **ace_repair\functions \fnc_canMiscRepair.sqf**

**Fix**
https://github.com/acemod/ACE3/blob/v3.3.2/addons/repair/functions/fnc_canMiscRepair.sqf#L23
```sqf
params ["_caller", "_target", "_hitPoint"];

if !([_unit, _target, ...
```
Replace "_unit" with "_caller". Or replace "_caller" with "_unit" in other 2 lines.

PabstMirror:
Bug is my fault, fixed in PR #2616

PR #2861, opened 2015-11-21
Repair Overhaul

PabstMirror:
ReBased Old PR #2616

Fixes:
#2307 - Repair Sub-turrets (commander's gun)
#2742 - Handle weird wheel configs
#2452 - Fix Body/Fuel Repair location for Hatchback
#2643 - Script error
ACE_isEngineer being scalar
Removed wheels not having damage set
Re-added normalizing "depends" hitpoints

Added:
Claim/Release repair items while repairing


SilentSpike:
To clarify:

PabstMirror:
Yeah, it's the same system that carrying/dragging uses to prevent interaction with the object.

PR #2865, opened 2015-11-21
Fix UBC for 1.54

PabstMirror:
Soldiers no longer inherit hitpoints from CaManBase
Adds a macro to deal with them


Glowbal:
:+1:

PR #2876, opened 2015-11-25
Medical Fixes

### LINKED ISSUE #1859: Killed eventhandler is ran/triggered twice because of ACE3
ACE3 Version: 3.1.1 stable PWS version
(indicate if stable or dev, if dev indicate the commit the version is based on)

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
None

**Description:**
Add a 'killed' eventhandler to a playable AI in a MP lan lobby while solo in that lobby, kill the AI and witness how the eventhandler is triggered twice.

I made a BIS thread about it before I knew that ACE3 was the problem; http://forums.bistudio.com/showthread.php?193486-Killed-Eventhandler-ran-twice

**Steps to reproduce:**
I made a repro mission: https://dl.dropboxusercontent.com/u/23922015/ai_eventhandler_bug.VR.rar

Multiplayer yet solo, LAN lobby, only player == server

**RPT log file:**
http://pastebin.com/skfgguNw

PabstMirror:
- Armor: With the 1.54 protection update, as far as I can tell most man hitpoints have armor of 1 and the actually protection is done on the Uniform/Vest/Helmet.
- Translate selections back to hitpoints for a few of the weird VR people.
- Calc newDamage using hitIndex because multiple hitpoints map to the same selection so `getHit` might not be correct. (right now _newDamage could be weird or negative)
- Fix isFalling was never being reset. So once you took falling damage that would be all you could ever take.
- Fix #1859 by delaying `setDamage 1` when inside of a handleDamage so the killedEH doesn't trigger twice

fnc_translateSelections has a sort of ugly overload

### DIFF
--- a/addons/medical/functions/fnc_handleDamage_caching.sqf
+++ b/addons/medical/functions/fnc_handleDamage_caching.sqf
@@ -8,29 +8,25 @@
  * 2: Amount Of Damage <NUMBER>
  * 3: Shooter <OBJECT>
  * 4: Projectile <STRING>
- * 5: Current damage to be returned <NUMBER>
+ * 5: HitPointIndex (-1 for structural) <NUMBER>
  *
  * Return Value:
  * <nil>
  *
  * Public: No
  */
-
 #include "script_component.hpp"
 
 private ["_hitSelections", "_hitPoints", "_impactVelocity", "_newDamage", "_cache_hitpoints", "_cache_projectiles", "_cache_params", "_cache_damages"];
-params ["_unit", "_selectionName", "_damage", "_source", "_projectile"];
-TRACE_8("ACE_DEBUG: HandleDamage_Caching Called",_unit, _selectionName, _damage, _source, _projectile,GVAR(SELECTIONS),GVAR(HITPOINTS),damage _unit);
+params ["_unit", "_selectionName", "_damage", "_source", "_projectile", "_hitPointIndex"];
+
 _hitSelections = GVAR(SELECTIONS);
-_hitPoints = GVAR(HITPOINTS);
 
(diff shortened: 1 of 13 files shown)


addons/medical/functions/fnc_handleDamage_caching.sqf
```diff
@@ -45,8 +41,14 @@ if (vehicle _unit != _unit && {!(vehicle _unit isKindOf "StaticWeapon")} && {isN
 // Handle falling damage
 _impactVelocity = (velocity _unit) select 2;
 if (_impactVelocity < -5 && {vehicle _unit == _unit}) then {
+     TRACE_1("Starting isFalling", time);
     _unit setVariable [QGVAR(isFalling), true];
     _unit setVariable [QGVAR(impactVelocity), _impactVelocity];
+} else {
+    if ((_unit getVariable [QGVAR(isFalling), false]) && {diag_frameno > (_unit getVariable [QGVAR(frameNo_damageCaching), -3]) + 2}) then {
+        TRACE_1("Ending isFalling", time);
```
MikeMatrix: Some odd indentation going on here.

Glowbal:
This looks ok. But since we are now using `hitpoint indeces`, we have to double check our manual handleDamage calls in for example overpressure. Those may not be using a proper index for each hitpoint.

Also @PabstMirror, I think the commits on this branch could do with some cleaning up.

SilentSpike:
:+1: For helper function

jonpas:
Very good start for an API. Next thing I can see for it is a sort of getDamage function of which return value could be fed right back into a sort of setDamage function.

PR #2883, opened 2015-11-26
Rough Base Port of Samples Test_LegacyFatigue

madpat3:
could you improve your climbing feature?

you can barely climb anything. only fortification's sandbags and through windows.
but there are many walls, that aren't much higher than that, and they're not climbable.
this makes the feature almost useless in combat.
good comparison is "LAxemann's Climb" mod http://www.armaholic.com/page.php?id=25996,
which looks quite similar to your animations, 
but is more useful in combat situations, since you can climb many obstacles.

speaking of climbing, from my point of view, the jump feature needs improvement, too. 
you can only jump with a rifle in hands.
not with a pistol and not even barehanded.

please :) think about it.

### DIFF
--- a/addons/LegacyFatigue/$PBOPREFIX$
+++ b/addons/LegacyFatigue/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\ace\addons\LegacyFatigue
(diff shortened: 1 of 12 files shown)


PabstMirror:
Please make an issue, this is a PR.
See http://ace3mod.com/wiki/user/how-to-report-an-issue.html

PR #2884, opened 2015-11-26
fix that in UGV dont work speedLimiter

### LINKED ISSUE #2875: Can't use speed limiter on UGV
ACE3 Version: 3.3.2

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* *None*

**Description:**
* It's not possible to enable speed limiter while driving UGV.*

**Steps to reproduce:**
* *Control UGV and try to enable speed limiter.*

*"Multiplayer", "Editor" or "Singleplayer"*

**RPT log file:**
* -

jokoho48:
posible fix #2875

### DIFF
--- a/addons/vehicles/XEH_postInit.sqf
+++ b/addons/vehicles/XEH_postInit.sqf
@@ -2,7 +2,7 @@
 #include "script_component.hpp"
 
 if (!hasInterface) exitWith {};
-
+GVAR(isSpeedLimiter) = false;
 // Add keybinds
 ["ACE3 Vehicles", QGVAR(speedLimiter), localize LSTRING(SpeedLimiter),
 {
--- a/addons/vehicles/XEH_postInit.sqf
+++ b/addons/vehicles/XEH_postInit.sqf
@@ -6,16 +6,27 @@ if (!hasInterface) exitWith {};
 // Add keybinds
 ["ACE3 Vehicles", QGVAR(speedLimiter), localize LSTRING(SpeedLimiter),
 {
-    // Conditions: canInteract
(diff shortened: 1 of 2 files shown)


addons/vehicles/XEH_postInit.sqf
```diff
@@ -2,20 +2,31 @@
...
-    // Conditions: canInteract
-    if !([ACE_player, objNull, ["isnotinside"]] call EFUNC(common,canInteractWith)) exitWith {false};
-    // Conditions: specific
-    if !(ACE_player == driver vehicle ACE_player &&
-    {vehicle ACE_player isKindOf 'Car' ||
-        {vehicle ACE_player isKindOf 'Tank'}}) exitWith {false};
+    private _connectedUAV = getConnectedUAV ACE_player;
+    private _uavControll = UAVControl _connectedUAV;
+    if ((_uavControll select 1) == "DRIVER") then {
+        if !(_connectedUAV isKindOf "UGV_01_base_F") exitWith {false};
```
commy2: No!!!

jonpas: Then we need a function in common, eg. `EFUNC(common,isUAV)`.

jokoho48: this case only need UGVs and not support UAVs

PabstMirror: `isUav = 1;`

Not many UAV's that are also a car/tank; so hard coded isn't terrible (this is a pretty odd edge case)

gienkov:
:+1:

PR #2888, opened 2015-11-27
Define custom repair points for a few vehicles

SilentSpike:
Some helicopters have the main rotor point on the tip of a single blade which can be problematic depending on the orientation of the helicopter so those have been moved to the center (to match others). The tail rotor is also in the complete wrong position for some.

Also the turret of `B_APC_Tracked_01_AA_F` was unreachable. It should be noted that there are two turret hitpoints with the same name so they are both moved to the new position (the second was overlapping with a different hitpoint anyway though).

### DIFF
--- a/addons/repair/CfgVehicles.hpp
+++ b/addons/repair/CfgVehicles.hpp
@@ -355,10 +355,24 @@ class CfgVehicles {
         transportRepair = 0;
     };
 
+    class Heli_Transport_02_base_F;
+    class I_Heli_Transport_02_F: Heli_Transport_02_base_F {
+        GVAR(hitpointPositions[]) = {{"HitVRotor", {-1,-9.4,1.8}}, {"HitHRotor", {0,1.8,1.3}}};
+    };
+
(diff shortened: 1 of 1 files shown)


SilentSpike:
I was going to commit this straight to master, but figured it's be better to let others give it the once over to make sure I didn't mess a config up.

SilentSpike:
Also if there are any other vehicles with repair position issues, feel free to mention it and I can take a look at those. I have a testing scenario for quickly visualising selection positions and new custom positions.

jonpas:
I find end-of-the-blade repair points intriguing. :grin: 

:+1:

SilentSpike:
I would like them there was a way to have it on each individual blade, but since it's for the whole thing it can be kind of annoying to search for one specific blade :laughing:

jonpas:
Yeah, unless BI would make each blade a separate hitpoint! :D

Glowbal:
Should be fine :+1:

PR #2889, opened 2015-11-28
Add mention when make fails to build all pbos.

Glowbal:
- Prints all failed modules when make has completed
- No longer prints green OK text if a single module fails
- Gives a count of how many modules have failed to build

This is done to prevent overlooking of a failed pbo in the build process, since there is a huge amount of modules being build.


jonpas:
:+1:

PR #2891, opened 2015-11-28
Legacy Fatigue

PabstMirror:
Rewrote BIS's Legacy Fatigue from A3 Samples

I'm not sure if this is something we want.  ~~The new "Stamina" isn't great but it's not completely awful anymore.~~ With 1.54 you can run forever @ 14kph, so it actually is pretty shit and not at all realistic.

Really only possible because of all the added commands in 1.54.

Being scripted we can have full control over over drain and recovery and offer more options changeable by mission makers, but it's also scripted instead of simple config changes.


jonpas:
Maybe a lead setting to toggle it on and off?

PabstMirror:
With the changes to `CfgMovesFatigue` the script part still needs to run, but it should be possible to still behave pretty much like vanilla 1.54 stamina.

addons/LegacyFatigue/README.md
```diff
@@ -0,0 +1,8 @@
+ace_x
+==========
+
+## Maintainers
+
+The people responsible for merging changes to this component or answering potential questions.
+
+- [x](https://github.com/x)
```
bux578: whoa ... who's that guy?

bux578:
Looking forwared to this. This should be released shortly after 1.54 :)

addons/LegacyFatigue/XEH_postInit.sqf
```diff
@@ -0,0 +1,26 @@
...
+if (!hasInterface) exitWith {};
+
+
+["SettingsInitialized", {
+
+    //Start on ACE_player
+    if (!isNull ACE_player) then {
+        [ACE_player] call FUNC(legacyFatigue);
+        ACE_LOGINFO("Starting fatigue on %1", _newPlayer);
+    };
```
commy2: Should be covered by the event below already.

addons/LegacyFatigue/config.cpp
```diff
@@ -0,0 +1,23 @@
...
+};
+
+#include "ACE_Settings.hpp"
+#include "CfgAnimStamina.hpp"
+#include "CfgEventHandlers.hpp"
+#include "CfgVehicles.hpp"
+
+class RscInGameUI {
+    class RscStaminaBar {
+        onLoad = "[""onLoad"",_this,""RscStaminaBar"",'IGUI'] call 	(uinamespace getvariable 'BIS_fnc_initDisplay');  uiNamespace setVariable ['test5', _this select 0];";
```
commy2: test5 
reminder to remove / rename

addons/LegacyFatigue/functions/fnc_legacyFatigue.sqf
```diff
@@ -0,0 +1,159 @@
...
+
+[{
+    BEGIN_COUNTER(all);
+    params ["_args", "_pfID"];
+    _args params ["_unit", "_lastFastRun", "_lastFullRun", "_scriptedChangePerSecond", "_scriptedChangeDone", "_lastPos", "_lastStamina", "_exhaustionEnd"];
+
+    if ((isNull _unit) || {!alive _unit} || {_unit != ACE_player}) exitWith {
+        TRACE_1("null/dead/notplayer",_unit);
+        [_pfID] call CBA_fnc_removePerFrameHandler;
+    };
```
commy2: `!alive objNull` is `true` now. And `objNull == objNull` is `false`. Something seems redundant.

commy2:
Nice.

PR #2893, opened 2015-11-29
1.55 use new selectRandom command

commy2:
https://community.bistudio.com/wiki/selectRandom
requires 1.55 dev


addons/medical/functions/fnc_playInjuredSound.sqf
```diff
@@ -59,9 +59,9 @@ if (_pain > 0 && {[_unit] call EFUNC(common,isAwake)}) exitwith {
     // Select the to be played sound based upon damage amount.
     if (_pain > 0.5) then {
         if (random(1) > 0.5) then {
-            _sound = _availableSounds_A select (round(random((count _availableSounds_A) - 1)));
+            _sound = selectRandom _availableSounds_A;
         } else {
-            _sound = _availableSounds_B select (round(random((count _availableSounds_B) - 1)));
+            _sound = selectRandom _availableSounds_B;
```
commy2: this is slightly different. should be fine though.

addons/sitting/functions/fnc_getRandomAnimation.sqf
```diff
@@ -46,4 +46,4 @@ _animations = [
 ];
 
 // Select random animation
-(_animations select (floor (random (count _animations))))
+selectRandom _animations
```
jonpas: Best to just remove L18 and L48-L49 and on L21 do `private _animations = [...`

addons/sitting/functions/fnc_getRandomAnimation.sqf
```diff
@@ -17,8 +17,8 @@
 
 private "_animations";
```
jonpas: Can remove as well.

PR #2895, opened 2015-11-29
Add Mission Config Actions

jokoho48:


addons/interact_menu/functions/fnc_compileMenu.sqf
```diff
@@ -103,12 +103,17 @@ _recurseFnc = {
 
 private ["_actionsCfg","_actions"];
 _actionsCfg = configFile >> "CfgVehicles" >> _objectType >> "ACE_Actions";
-
+_missionActionsCfg = missionConfigFile >> "ACE_MissionActions" >> _objectType >> "ACE_Actions";
```
jokoho48: missing Private

esteldunedain:
While I like the idea, this implementation is very limited. It only supports adding more interaction points to objects, but not adding more actions to existing action points.

Glowbal:
We will discuss this on Slack further and reopen if we have made a decision.

PR #71, opened 2015-01-19
Port of AGM_FastRoping

KoffeinFlummi:
Still not quite sure how to properly do the giant spawn in fnc_fastRope...

PR #2904, opened 2015-12-01
Fastroping

BaerMitUmlaut:
This is a port of AGMs fastroping and replaces #71. All functions have been completly rewritten, and pretty much only the base system remains the same. Includes a model from @KoffeinFlummi, slightly modified (added empty fire LOD).
Requires only two config entries from a mod maker to add the possibility to fast rope and can be extended with custom animations, FRIES etc.

Known issues:
- If a player is fastroping and the rope breaks below him (because it got shot at for example), he continues to fast rope just fine. This is because a broken rope cannot be retracted and it would look weird if he suddenly lets go of the rope, even though there is still a piece of rope below him.  
Maybe create a new, short rope, and retract that? Feel free to suggest a solution for this.
- You currently have to cut the ropes and redeploy the FRIES if both ropes break. That means you'll have to go through the *open door & extend hook* animation twice (which takes 8 seconds in the Ghosthawk). However, I think this is fair as you'd have to attach new ropes to the hooks, which would be hard to do with the hooks extended. IRL, a helo probably doesn't carry spare ropes either.
- The FRIES is not visible when you're sitting inside of the Ghosthawk. I think this is an LOD issue which I can do nothing about.

TODO:
- [ ] Do ? when bottom rope breaks
- [x] Add/Expand onPrepare function for all helos (e.g. KA-60 uses Czech animation names)
- [x] Add more FRIES models, texture current one - [reference for FRIES gantry](http://www.quantico.marines.mil/Portals/147/Docs/Range%20Management%20Branch/PubsInfo/3570/MCRP%203-11-4A.pdf), [reference for Chinook FRIES](https://www.flickr.com/photos/avgeekjoe/14390508271)
- [ ] Add function for automatic fast roping of AI (optional, but I've got the base of it already, might as well make some people happy :blush: )


PabstMirror:
:fries: 
 I know a lot of people have been waiting for this.

Glowbal:
Once you are done, could you clean up the commit history?

addons/fastroping/functions/fnc_canCutRopes.sqf
```diff
@@ -0,0 +1,24 @@
...
+ * [_player, _vehicle] call ace_fastroping_canCutRopes
+ *
+ * Public: No
+ */
+
+#include "script_component.hpp"
+params ["_unit", "_vehicle"];
+private ["_deployedRopes"];
+
+_deployedRopes = _vehicle getVariable [QGVAR(deployedRopes), []];
```
Glowbal: `private _DeployedRopes =`

BaerMitUmlaut:
Sure, will do.

addons/fastroping/functions/fnc_canCutRopes.sqf
```diff
@@ -0,0 +1,24 @@
...
+ * Able to cut ropes <BOOL>
+ *
+ * Example:
+ * [_player, _vehicle] call ace_fastroping_canCutRopes
+ *
+ * Public: No
+ */
+
+#include "script_component.hpp"
+params ["_unit", "_vehicle"];
```
Glowbal: `_unit` doesn't seem to be used here

addons/fastroping/functions/fnc_canDeployRopes.sqf
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Public: No
+ */
+
+#include "script_component.hpp"
+params ["_unit", "_vehicle"];
+private ["_deployedRopes", "_config"];
+
+_deployedRopes = _vehicle getVariable [QGVAR(deployedRopes), []];
+_config = configFile >> "CfgVehicles" >> typeOf _vehicle;
```
Glowbal: `_config` is unused

addons/fastroping/functions/fnc_prepareFRIES.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+#include "script_component.hpp"
+params ["_unit", "_vehicle"];
+private ["_deployedRopes"];
+
+//Stage indicator: 0 - travel mode; 1 - preparing FRIES; 2 - FRIES ready; 3 - ropes deployed
+_vehicle setVariable [QGVAR(deploymentStage), 1, true];
+
+_config = configFile >> "CfgVehicles" >> typeOf _vehicle;
+_waitTime = 0;
```
Glowbal: `private _waitTime =`

addons/fastroping/functions/fnc_prepareFRIES.sqf
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+#include "script_component.hpp"
+params ["_unit", "_vehicle"];
+private ["_deployedRopes"];
+
+//Stage indicator: 0 - travel mode; 1 - preparing FRIES; 2 - FRIES ready; 3 - ropes deployed
+_vehicle setVariable [QGVAR(deploymentStage), 1, true];
+
+_config = configFile >> "CfgVehicles" >> typeOf _vehicle;
```
Glowbal: `private _config =`

jonpas:
> The FRIES is not visible when you're sitting inside of the Ghosthawk. I think this is an LOD issue which I can do nothing about.

Yes, inside LOD rendering works differently, you can probably see it in some other position.

BaerMitUmlaut:
It's fine in anything but first person view as a crew member.

Khaosmatic:
Would it be possible to create SPIE Rigging from the same system?

addons/fastroping/functions/fnc_moduleEquipFRIES.sqf
```diff
@@ -0,0 +1,46 @@
...
+ * Equips synched helicopters with a FRIES.
+ *
+ * Arguments:
+ * 0: Module <LOGIC>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [_vehicle] call ace_fastroping_fnc_cutRopes
```
bux578: wrong

esteldunedain:
After the testing and fixing done yesterday, this should be ready for merging now. idk if it's feasable to do it for 3.5.0 or not, but I'm updating the milestone so it's reconsidered.

The bug @VKing6 happened when the invisible anchor at the end of the rope didn't touch the ground. That object is now removed.

BaerMitUmlaut:
@Khaosmatic
>Would it be possible to create SPIE Rigging from the same system?

I'm not sure, especially because of the bug we found yesterday. If I had to guess I'd say no. Something like descending a stretcher or something similar from a helo would probably be possible, but attaching multiple objects or ropes to a rope seems to cause physics bugs.

VKing6:
The ace_fastroping_helper is visible in the editor under empty->air as "Helicopter [Civilians]".

BaerMitUmlaut:
Fixed.

PabstMirror:
Mikro tools doesn't like something in config (`config_validator.py full`)
```
EQUIP_FRIES_ATTRIBUTE;
In File fastroping\CfgVehicles.hpp: Line 184 Expected Semicolon (or eol)
```
`x3 = "ace_fastropeing_helper" createVehicle [0, 0, 0];`
is `objNull` and rpt: Cannot create non-ai vehicle ace_fastropeing_helper,
(I don't think it can be scope =0 )??

jonpas:
It has to be `scope = 1` to hide it in Editor but still allow usage with `createVehicle`.

BaerMitUmlaut:
Fixed scopes and an error in the Eden macro. @PabstMirror can you run that check again? The script doesn't run for me (`AttributeError: 'module' object has no attribute 'run'`).

PabstMirror:
Configs work now (I think the tool needs a configured p:, otherwise you can just run it though pboProject which will do same checks)

Vehicles create fine now. I did have the object name spelled wrong before :smile: , but I think the scope = 0 was still a problem.

commy2:
merge this now ffs

PR #2932, opened 2015-12-04
Fix binocular related bugs

commy2:


Glowbal:
This should be going into release, not master.

PR #2948, opened 2015-12-05
remove Original entries

gienkov:
This PR removes two "Original" entries from stringtables so this:
problem
Becomes this:
noproblem
Making it easier to edit stringtables with Tabler :)


jokoho48:
very trivial :+1:

MikeMatrix:
seems trivial :+1:

PR #2949, opened 2015-12-06
ceil number of wounds

gienkov:
This fixes not round number of wounds, see image below (1.7x wound). This happens when you use wrong type of bandage on wound. Thing below that is just my experiment that failed, sadly.
img

### DIFF
--- a/addons/medical/functions/fnc_displayPatientInformation.sqf
+++ b/addons/medical/functions/fnc_displayPatientInformation.sqf
@@ -101,7 +101,7 @@ if (_show) then {
                     // Collect the text to be displayed for this injury [ Select injury class type definition - select the classname DisplayName (6th), amount of injuries for this]
                         if (_amountOf >= 1) then {
                             // TODO localization
-                            _allInjuryTexts pushback [format["%2x %1", (GVAR(AllWoundInjuryTypes) select _x1) select 6, _amountOf], [1,1,1,1]];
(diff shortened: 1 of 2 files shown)


Glowbal:
Why is showing 1.7x a bad thing?

VKing6:
Because it doesn't make sense to have fractions of a wound?
If you have a partially bandaged or treated wound it should say "Partially bandaged wound" not "0.7 wounds".

gienkov:
Let me explain, back before 3.4.0 all bandages were not correctly defined, and all of them had effectiveness of 1. Because of that you could use any bandage type and it would still heal that one wound completely. Right now bandages are defined properly and they have different effectiveness, so for example bandaging velocity wounds with basic bandage will use 4 of them because they have effectiveness of 0.3. That number is substracted from openWounds array and displayed as number of wounds after that. But I think you still want to count them as before and this PR does that:
2x wounds of same type are displayed as 2x wounds
1.7x wounds are displayed as 1x wound and partially open wound and so on.

Glowbal:
I have added this to the 3.5.0 milestone, but I might consider adding it to 3.4.2. I'd like to double check the effects before I decide to merge this (and for what version).

TheGrandVizier:
Player Input, if you care:
WE LOVE THE "0.7" WOUND SYSTEM!

Very clear to untrained personnel what is going on and why.
We like the indications. Please retain.

PR #2954, opened 2015-12-06
nozzle additonal check condition

gienkov:
Have you ever did refuel at gas station with **engine on**?

### DIFF
--- a/addons/refuel/functions/fnc_canConnectNozzle.sqf
+++ b/addons/refuel/functions/fnc_canConnectNozzle.sqf
@@ -23,4 +23,4 @@ _nozzle = _unit getVariable [QGVAR(nozzle), objNull];
 
 !(isNull _nozzle ||
     {(_target distance _unit) > REFUEL_ACTION_DISTANCE} ||
-    {!isNull (_target getVariable [QGVAR(nozzle), objNull])}) // TODO verify cant connect multiple fuel lines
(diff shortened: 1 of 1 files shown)


commy2:
Yes, I have. : - )

jokoho48:
@commy2 bad idea :D i like the idea :+1:

VKing6:
Not sure that idea applies to tanks.
They wouldn't drive up to the refuel bowser and turn their engines off just for the few minutes it takes to refuel.

gienkov:
@VKing6 you can still start up tank engine within seconds in Arma so...

SilentSpike:
Isn't this invalid syntax?

gienkov:
I'm gonna test but I used wiki https://community.bistudio.com/wiki/isEngineOn

SilentSpike:
I mean because the condition is using lazy evaluation and you've added a non-lazy condition onto the end

gienkov:
Um... tell me how to fix? :)

SilentSpike:
I'd add a line under 24 with:

`{isEngineOn _target} ||`

esteldunedain:
In any case the lazy evaluation thing should be fixed @gienkov

gienkov:
Fixed!

bux578:
:+1:

Issue #2914, opened 2015-12-02
Cannot restrict Map Marker channels anymore


adampiskorski:
In the past I used [disableChannels[]](https://community.bistudio.com/wiki/Description.ext#disableChannels) in the description.ext in order to prevent players from not just typing messages in certain channels, but also from making markers in those channels. 
With the ACE Map marker placement window, you can use channels regardless of the [disableChannels[]](https://community.bistudio.com/wiki/Description.ext#disableChannels) setting, thus overriding it.

Probably the main example of this is to prevent players from putting down global or side markers after the mission start. This increases the value of tools like cTab.

commy2:
`disableChannels` no longer works for side and another channel I can't remember anymore.
Will look into it.

SilentSpike:
There is also a new `enableChannel` [command](https://community.bistudio.com/wiki/enableChannel)

SilentSpike:
I'm not sure there's anything we can do to account for that one though

adampiskorski:
I checked it again on my side and after checking it properly on a dedicated server in Vanilla, I found out that is Vanilla Arma that brought in the channel selection for markers. 
Sorry about that, I seldom play Vanilla Arma.

Neefay:
There is a very simple fix for this issue. Currently I'm unable to submit a pull, but if anyone would be so kind, this is tested and works with no problems.

[Insert within `markers\XEH_preInit.sqf`, before `ADDON = true`.](http://pastebin.com/d9sFRBCa)

[Insert in `markers\functions\fnc_initInsertMarker.sqf`, within `EFUNC(common,execNextFrame)` .](http://pastebin.com/NEjE7swR)

As you can tell, the code is far from perfect but it's what came out after I found myself struggling with that issue for a while.

commy2:
Should work better using ACE in the next version. Classifying this as feature.

commy2:
Done in release branch, ref: #3004

PR #2990, opened 2015-12-09
Code cleanup Markers

commy2:
- disabled channels are no longer selectable
- the currently selected channel is auto selected
- changing the channel a marker should be set in also changes the current channel

closes: #2914


PabstMirror:
`x3 = radioChannelCreate [[0.96, 0.34, 0.13, 0.8], "Q-dance Radio", "%UNIT_NAME", [player]];`
Switch to channel with `,`/`.`and try placing marker:
```
Error in expression <el lbText _j];
if ((_channel lbText _j) == _currentChannelName) then {
_channel >
Error position: <== _currentChannelName) then {
_channel >
Error Generic error in expression
File z\ace\addons\markers\functions\fnc_initInsertMarker.sqf, line 153
```
I think `currentChannel` is 6 so out of bounds for `CHANNEL_NAMES`.

Also:
`disableChannels[]={0,1,2,3,4,5,6};` in description.ext and was still able to select group channel on client and group or global for admin.

commy2:
> disableChannels[]={0,1,2,3,4,5,6}; in description.ext and was still able to select group channel on client and group or global for admin.

Yes. You cannot disable the group channel or the global channel in case of the admin.
`disableChannels[]` and even `3 enableChannel false;` does not work. Same happens with the next channel and prev channel keybinds, so I think this is fine.

I'll try to fix the error with custom channels.

commy2:
I must say it's funny how the menu flies off to the right when using the custom channel, haha.

commy2:
>I think currentChannel is 6 so out of bounds for CHANNEL_NAMES.

The default value of a `param` I specifically placed for this reason was a number and not a string. Derp.

commy2:
Now supports custom radio channels as good as possible. Unfortunately there is no way to auto select these channels as there is no command to retrieve the displayed name of a channel and I have to use lbText to identify them in the first place.

PR #2996, opened 2015-12-09
fix Unexpected stringtable format

havena:
Unexpected stringtable format inside <Text ID="STR_ACE_Common_CheckPBO_Action_WarnPerm"><German>


bux578:
Thanks

PR #3008, opened 2015-12-10
Effectivness of the new Vests system.

TheMasterofBlubb:
What i would like to see is that if you are wearing a SK4 Vests you dont get ballistic injouries you get blunt wounds or pain. My suggestion would be :
fnc_handleDamage.sqf
at the point where you use the getDamageType. you add a modificator for the type of vest you are wearing. for example SK4 the highest ballistic class and say that 9mm would just give minor injuries. And if you have a Explosive resistent one you for example get additional shrapnel protection.
Like i undertood BI it should be:
SK1 .45ACP and 9mm low ammount
SK2 .45ACP and 9mm about 4-5 shots
SK3 low ammount of 5.56mm
SK4 9mm is useless, some 5.56mm and 1 or 2 7.62mm
Explosive resistent : SK4 with some protection against explosions or shrapnels


Glowbal:
Please create a feature request instead of a PR. If this is a PR containing code, please fix your branch / fork and resubmit your PR.

PR #3013, opened 2015-12-11
Fix litter for adv - Add bloody p3d for Pack/qClot

### LINKED ISSUE #2950: Bandages are always clean
ACE3 Version: 3.4.0

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* *-*

**Description:**
* When you bandage wounds, litter created is clean bandage, always.*

**Steps to reproduce:**
* *Bandage any wound.*

*"Multiplayer", "Editor" or "Singleplayer"*

PabstMirror:
Fix #2950
Get correct bloodLossOnSelection for both basic and advanced
Adds bloody versions of packing bandage and quickClot


jokoho48:
models have some issues!

jokoho48:
i think its better to add a posibilty to spawn 2 vehicles instatt of adding more and more p3ds

Glowbal:
What you mean with that @jokoho48? This PR doesn't add any new models?

jokoho48:
that we better find a method to spawn 2 Objects if the player is bleeding and not add more and more p3ds because the maintenance get harder and harder than more models we have

PabstMirror:
Not using new models, now just creating both empty packaging litter AND (clean/dirty) bandage

jokoho48:
this is way better :+1:

PR #3014, opened 2015-12-11
Sanitize less/greater-than symbol to square

### LINKED ISSUE #2912: Problem with "<" and ">" in the Tag
In first, Hellow to all the Arma and Ace community.

I'm member of Grupo Armados (Spanish military simulation group for Arma3) and our tag is `<A2>`, but if we dont put the separation this not appear in the text. We have a problem with this and the name tag of ACE3 (Yeah, the letters who appear in the top of the player). 

We don't like to change our tag becouse we has used this since OPX, soo we stay were to propose to he ACE developments solve this simple but annoying problem (That is only put some lines of code in the script).

Thanks soo much for read the propose (or problem) and we hope your answer (o solution).

Regards form Spain and Grupo Armados.

PabstMirror:
Fix #2912
The raw `&lt;` showed up in nametags and anything besides strucText
`[ / ]` is fairly close and will work in either text mode.

### DIFF
--- a/addons/common/functions/fnc_getName.sqf
+++ b/addons/common/functions/fnc_getName.sqf
@@ -5,7 +5,7 @@
  * Arguments:
  * 0: Object <OBJECT>
  * 1: Use effective commander name when used on vehicles (default: false) <BOOL>
- * 1: Get Raw Name (Don't sanatize HTML tags `</>`) (default: false) <BOOL>
+ * 2: Get Raw Name (Don't sanatize HTML tags `</>`) (default: false) <BOOL>
  *
  * Return Value:
  * Object Name <STRING>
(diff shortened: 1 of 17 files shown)


Glowbal:
But it is a different character. I'd rather use a different sanitize method for name tags instead.

commy2:
> I'd rather use a different sanitize method for name tags instead.

this

PabstMirror:
This now saves a copy of both ways of displaying html tags, and FUNc(getName) now has an option to pick which version to use depending on how the output will be displayed.

addons/common/functions/fnc_getName.sqf
```diff
@@ -5,24 +5,27 @@
  * Arguments:
  * 0: Object <OBJECT>
  * 1: Use effective commander name when used on vehicles (default: false) <BOOL>
+ * 1: Get Raw Name (Don't sanatize HTML tags `</>`) (default: false) <BOOL>
```
commy2: 2

commy2:
:+1:

PR #3018, opened 2015-12-11
Fix refuel PFH not exiting on nozzle disconnect

### LINKED ISSUE #3009: Fuel hose max length exceeded message after disconecting
ACE3 Version: 3.4.1
ArmA 1.54 stable

**Mods:**
* `@CBA_A3`
* `@ace`
* `@bwa3

**Placed ACE3 Modules:**
none

**Description:**
* After using the fuel system and putting it back on the supply vehicle, driving away with the supplied vehicle notifies me about exeededing the max fuel hose lenght like as if i would walk away with the fuel hose in my hands*

**Steps to reproduce:**
* Place a tanker a vehicle with empty tank*
*tank the vehicle"
"put the fuel hose back on the tanker"
"drive away"

*Singleplayer*

**RPT log file:**
* http://pastebin.com/2VJ3hqP3*

jonpas:
Fixes #3009


jokoho48:
:+1: look good from my side 
@GitHawk

addons/refuel/functions/fnc_refuel.sqf
```diff
@@ -29,6 +29,10 @@ _maxFuel = getNumber (configFile >> "CfgVehicles" >> (typeOf _target) >> QGVAR(f
     params ["_args", "_pfID"];
     _args params ["_source", "_sink", "_unit", "_nozzle", "_rate", "_startFuel", "_maxFuel", "_connectFromPoint", "_connectToPoint"];
 
+    if (_nozzle getVariable [QGVAR(isConnected), false]) exitWith {
```
GitHawk: Nozzle gets deleted by [fnc_returnNozzle](https://github.com/acemod/ACE3/blob/master/addons/refuel/functions/fnc_returnNozzle.sqf#L48). So I'd rather check if _source is still connected.

jonpas: I don't think it necessary, this PFH only starts when you attach the nozzle to some vehicle, and with this line it will remove the PFH when you disconnect it from the vehicle. That is before returning the nozzle, PFH will not exist when returning it anymore.

commy2: Doesn't this line need a `!` ? If the nozzle is deleted the `getVariable` will always default to `false`.

GitHawk: Sorry, your code doesn't reflect what you are saying. This PFH has to run while the `_nozzle` is connected.

`QGVAR(isConnected)` is set to `true` for `_nozzle` in [fnc_connectNozzleAction](https://github.com/acemod/ACE3/blob/master/addons/refuel/functions/fnc_connectNozzleAction.sqf#L94) which is calling `fnc_refuel`. So you would add the PFH and remove it next time it executes.

So either change it to `_source` and the PFH gets removed upon returning.
Or query `QGVAR(sink)` on `_nozzle` which is also set [here](https://github.com/acemod/ACE3/blob/master/addons/refuel/functions/fnc_connectNozzleAction.sqf#L93) but gets cleared on [disconnecting](https://github.com/acemod/ACE3/blob/master/addons/refuel/functions/fnc_disconnect.sqf#L26)

jonpas: Ah yes, that is my bad, fixing.

GitHawk: Looks better

PR #3020, opened 2015-12-12
Common code cleanup 1.54

commy2:


addons/common/XEH_postInit.sqf
```diff
@@ -251,8 +248,7 @@ GVAR(ScrollWheelFrame) = diag_frameno;
 }] call FUNC(addEventHandler);
 
 // add PFH to execute event that fires when the main display (46) is created
-private "_fnc_initMainDisplayCheck";
-_fnc_initMainDisplayCheck = {
+private _fnc_initMainDisplayCheck = {
     [{
         if !(isNull findDisplay 46) then {
```
jokoho48: replace PFH with waitUntilandExecute

addons/common/functions/fnc_errorMessage.sqf
```diff
@@ -37,35 +37,29 @@ if (_textMessage isEqualType "") then {
```
jokoho48: use in line 24 WaituntilAndExecute

commy2:
jokoho, will do that later. This is mainly for the private keyword.

Issue #3021, opened 2015-12-12
Bodybag event handler 


Tenga:
##### Feature request:
Can you please add an event when a body is being transferred into a bodybag, that triggers at some point when both the bag and the body are present and has them both as arguments?

It would be nice to use it to set some variables onto the bodybag depending on the unit, so for example, you can identify a mission critical dead body that was converted to a bodybag.

##### Optional / discussion for potential changes to the bodybag system:
Additionally, I'm not sure how bad would it be to change the way bodybags work, to just hide the dead body and attach it to it's bodybag, in a similar manner as the cargo is currently attached to the vehicle it is in, so you could even "unpack" it from the body bag down the line.

Tenga:
:+1:

commy2:
Done. 3.5.0

PR #3022, opened 2015-12-12
add event for placing units in body bags

commy2:
ref: #3021

`["placedInBodyBag", {systemChat str _this}] call ace_common_fnc_addEventhandler`

0: dead body
1: body bag object

executed on all machines

### DIFF
--- a/addons/medical/functions/fnc_actionPlaceInBodyBag.sqf
+++ b/addons/medical/functions/fnc_actionPlaceInBodyBag.sqf
@@ -28,9 +28,11 @@ _spinePos = _target modelToWorldVisual (_target selectionPosition "Spine3");
 _dirVect = _headPos vectorFromTo _spinePos;
 _direction = _dirVect call CBA_fnc_vectDir;
 
-deleteVehicle _target;
-
 _bodyBag = createVehicle ["ACE_bodyBagObject", _position, [], 0, "CAN_COLLIDE"];
+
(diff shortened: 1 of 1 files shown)


jonpas:
Does the dead body retain information about what gear was on it as well?

Glowbal:
That won't work. The target is already removed before the event gets triggered. MAYBE a local event works. Depends on the implementation - I am not certain if the event gets called right away or next frame?

commy2:
>The target is already removed before the event gets triggered. MAYBE a local event works. Depends on the implementation

Definitely needs some testing. Localy the event does works (I tested it). I don't think the deleteObject is transfered faster than the publicVar, so the chances are good.

>Does the dead body retain information about what gear was on it as well?

Yes. Except for the primary weapon of course.

jonpas:
> Yes. Except for the primary weapon of course.

Awesome, primary weapon gets dropped right? So it doesn't get deleted with the dead body either?

commy2:
Yes.

jonpas:
Thanks!

commy2:
Works in local hosted multiplayer. The event triggers with the object before it is deleted.

Glowbal:
What about on a dedicated server, where another player puts a body into a bodybag?

commy2:
Can't test that, but I see no reason why it would be different there.

PR #3023, opened 2015-12-12
add function to retrieve the control of the stamina bar

commy2:
this disables the stamina bar for the whole game sitting (not just mission):

`(call ace_common_fnc_getStaminaBarControl) ctrlShow false`

### DIFF
--- a/addons/common/RscInfoType.hpp
+++ b/addons/common/RscInfoType.hpp
@@ -107,6 +107,10 @@ class RscInGameUI {
     class RscOptics_LaserDesignator_02 {
         onLoad = QUOTE([ARR_2('infoDisplayChanged', [ARR_2(_this select 0, 'Any')])] call FUNC(localEvent););
     };
+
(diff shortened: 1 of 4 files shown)


licht-im-Norden87:
I dont know if i like it. In reality you always know in some way

licht-im-Norden87:
I don't know if i like this. In reality you always know your personal exhaustion level in some way. (knowing running for another 12km is not possible anymore without having a break etc. ) Arma should give us this information to.

commy2:
This only adds a function to return the control. The line I posted is completely optional.

licht-im-Norden87:
Nevertheless I'd like to see another method to display your stamina. The current one is way to precise or in other words to "mathematical". What could be more realistic is an indicator using colors instead of a bar. (like the wind arrows) Starting with green slowly changes to orange and finally to red if you are to exhausted.

commy2:
Sure. But all that has nothing to do with this PR.

PR #3027, opened 2015-12-13
Send string to serverLog instead of strucText

PabstMirror:
Trivial Fix for
`Performance warning: SimpleSerialization::Read
'ace_common_remoteFnc' is using type of ,'TEXT' which is not optimized
by simple serialization, falling back to generic serialization, use
generic type or ask for optimizations for these types`


MikeMatrix:
Seems pretty trivial
:+1:

PR #3041, opened 2015-12-16
Optics - Don't modify camera for ThreeDen Editor

PabstMirror:
Fix broken ThreeDen Camera when returning to editor from preview.
PIP will still work in preview.
Will have to take a look at 1.56 RC to see if command exists and eventually cleanup.

### DIFF
--- a/addons/optics/XEH_postInit.sqf
+++ b/addons/optics/XEH_postInit.sqf
@@ -6,6 +6,12 @@ if (!hasInterface) exitWith {};
 GVAR(camera) = objNull;
 
 0 = 0 spawn {
+    // Workarround for the ThreeDen Editor Camera
+    //ToDo: verify if command will exist in 1.56 release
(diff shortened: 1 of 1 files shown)


addons/optics/XEH_postInit.sqf
```diff
@@ -6,6 +6,12 @@ if (!hasInterface) exitWith {};
 GVAR(camera) = objNull;
 
 0 = 0 spawn {
+    // Workarround for the ThreeDen Editor Camera
+    //ToDo: verify if command will exist in 1.56 release
+    if (((productVersion select 2) >= 155) && {is3DEN}) exitWith {
```
commy2: Maybe `!isNil {is3DEN}` instead? Just in case they don't finish the editor for 1.56.

jokoho48:
:+1: work fine

PR #3050, opened 2015-12-18
RHS Compat fix for SMAW spotting rifle backblast

VKing6:
Fix for the SMAW's spotting rifle causing backblast


sargken:
Should the SMAW be allowed to use other optics than just the SMAW scope? Could that be another fix that could go in here?

VKing6:
That looks intentional, so there's nothing to fix really.

PR #3052, opened 2015-12-18
Add support for non local Zeus RC in common-isPlayer

PabstMirror:
Adds support for non-local zeus remote controlled units in isPlayer.
So this command will now work consistantly on any machine.

Uses a BIS setVar from the zeus module, had to add a disconnect handler to clear variable.
Slightly faster as well.

I think this is important for consistent behavior, for example wind deflection does:
`if (!([_unit] call EFUNC(common,isPlayer))) exitWith {false};`
Right now only the zeus would run the code, so the bullet's path won't appear consistent with it's effects.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -69,6 +69,7 @@ if (isServer) then {
                     ACE_LOGINFO_3("[%1] DC - Was Zeus [%2] while controlling unit [%3] - manually clearing `bis_fnc_moduleRemoteControl_owner`", [_x] call FUNC(getName), _dcPlayer, _x);
                     _x setVariable ["bis_fnc_moduleRemoteControl_owner", nil, true];
                 };
+                nil
             } count (curatorEditableObjects  _zeusLogic);
         };
     }];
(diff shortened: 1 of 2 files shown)


commy2:
Nice.

addons/common/XEH_postInit.sqf
```diff
@@ -57,6 +57,23 @@
...
+    addMissionEventHandler ["HandleDisconnect", {
+        params ["_dcPlayer"];
+        private _zeusLogic = getAssignedCuratorLogic _dcPlayer;
+        if ((!isNil "_zeusLogic") && {!isNull _zeusLogic}) then {
+            {
+                if ((_x getvariable ["bis_fnc_moduleRemoteControl_owner", objnull]) isEqualTo _dcPlayer) exitWith {
+                    ACE_LOGINFO_3("[%1] DC - Was Zeus [%2] while controlling unit [%3] - manually clearing `bis_fnc_moduleRemoteControl_owner`", [_x] call FUNC(getName), _dcPlayer, _x);
+                    _x setVariable ["bis_fnc_moduleRemoteControl_owner", nil, true];
+                };
+            } count (curatorEditableObjects  _zeusLogic);
```
bux578: iirc there has been a discussion in slack that we always should return something in count loops. `nil`, `false` etc...

SilentSpike: This will implicitly return `nil`, though it is probably good practise to do that explicitly

PR #3057, opened 2015-12-20
Fix medical menu not closing on range, litter dir

PabstMirror:
- Medical Menu didn't close when target moved out of range
- Litter was always created in the same direction


Glowbal:
:+1:

PR #3058, opened 2015-12-20
fix inverted isTurnedOut for map lighting

freghar:
A simple oneliner for what seems to be an obvious typo.

According to the comment, the idea is to provide unobscured map when
inside an enclosed vehicle (`!isTurnedOut`).

### DIFF
--- a/addons/map/functions/fnc_determineMapLight.sqf
+++ b/addons/map/functions/fnc_determineMapLight.sqf
@@ -73,7 +73,7 @@ _vehicle = vehicle _unit;
 // Do not obscure the map if the player is on a enclosed vehicle (assume internal illumination)
 if (_vehicle != _unit) then {
     // Player is in a vehicle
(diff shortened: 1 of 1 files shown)


esteldunedain:
:+1:

commy2:
merge merge merge
:shipit:

vlad8011:
Hmm, its not fixed fully - sitting in cargo positions doesnt makes illumination working, I think it will be best if every soldier will have flashlight for the map in his inventory (including RHS units...)

esteldunedain:
> Hmm, its not fixed fully - sitting in cargo positions doesnt makes illumination working, I think it will be best if every soldier will have flashlight for the map in his inventory (including RHS units...)

That's completely unrelated to this issue

freghar:
Strange, it worked for me when I tested it in some open car and a tank. Maybe something else got broken.

vlad8011:
ahh, RHS, dear RHS mod again....

PR #3060, opened 2015-12-20
Ab - 	Fix Eotvos effect

PabstMirror:
@ulteq

This could really use another set of eyes on the math, but I think this is a fix for the vertical Coriolis effecting being off.

Ref https://en.wikipedia.org/wiki/E%C3%B6tv%C3%B6s_effect

2 * OMEGA[rotationRate (radians / second)] * u [MuzzleVelocity in the eastern direction (meters / second)] * cos (PHI[latitude])

which is in m/(s^2)

m/(s^2) * deltaTime = change in velocity Over DeletaTime

It used to divide by `-32.2` which is gravity in freedom units??? 
I think this was used in a calculation for getting the unit-less gravity multiplier, but here we want the actual acceleration.

Also, I am not sure why we are using `muzzleVelocity * sin(bulletDir)`, as when the bullet slows down the effect should drop, but this difference would be minor and might just be so it matches the calculations in the atragmx


PabstMirror:
Is this ok for 3.4.2 / should I recompile dll for 3.4.2?

extensions/CMakeLists.txt
```diff
@@ -89,8 +89,8 @@ endif()
 
 string(TIMESTAMP ACE_BUILDSTAMP "%Y-%m-%dT%H:%M:%SZ")
 set(ACE_VERSION_MAJOR 3)
-set(ACE_VERSION_MINOR 2)
-set(ACE_VERSION_REVISION 0)
+set(ACE_VERSION_MINOR 4)
+set(ACE_VERSION_REVISION 1)
```
bux578: `2` if for 3.42

bux578:
:+1:
IIRC @J3aLeiNe has confirmed that this change is working.

How's the process with BattleEye if we recompile the DLL's? Do they have to whitelist them?
@ViperMaul do you happen to know that?

jonpas:
AFAIK ACE3 DLLs are automatically whitelisted.

J3aLeiNe:
This change works but i think it could be interesting to do more test with "g" in metric unit.
The AtragMx uses the same Bryan Litz's formula  with metric unit and the vertical coriolis effect's corrections are accurate in game : _eoetvoesMultiplier = 2 * (0.0000729 * _muzzleVelocity / -9.80665) * cos(_latitude) * sin(_directionOfFire);

I suppose that it will be the same with the bullet trajectory in game.

PR #3067, opened 2015-12-22
Optimize map gestures - transmit function

Sniperhid:
Better worst case complexity.

### DIFF
--- a/addons/map_gestures/functions/fnc_transmit.sqf
+++ b/addons/map_gestures/functions/fnc_transmit.sqf
@@ -28,18 +28,18 @@ if (!GVAR(EnableTransmit) || !visibleMap) exitWith {
     [_pfhId] call CBA_fnc_removePerFrameHandler;
 };
 
-{
-    _ownerID = _x getVariable QGVAR(owner_id);
-    if (isNil "_ownerID") then {
-        [EVENT_PLAYER_HAS_NO_OWNER_ID, [name _x]] call EFUNC(common,serverEvent);
-    } else {
-        _playerOwnerID = ACE_player getVariable QGVAR(owner_id);
-        if (!isNil "_playerOwnerID" && _ownerID != _playerOwnerID) then {
-            _unitUID = getPlayerUID ACE_Player;
-            _drawPosVariableName = if (!isNil "_unitUID" && _unitUID != "") then {format [QGVAR(%1_DrawPos), _unitUID]} else {nil};
-            if (!isNil "_drawPosVariableName") then {
+_unitUID = getPlayerUID ACE_Player;
+_playerOwnerID = ACE_player getVariable QGVAR(owner_id);
+if (!isNil "_unitUID" && _unitUID != "" && !isNil "_playerOwnerID") then {
+    _drawPosVariableName = format [QGVAR(%1_DrawPos), _unitUID];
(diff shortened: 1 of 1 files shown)


addons/map_gestures/functions/fnc_transmit.sqf
```diff
@@ -28,18 +28,18 @@ if (!GVAR(EnableTransmit) || !visibleMap) exitWith {
...
-        [EVENT_PLAYER_HAS_NO_OWNER_ID, [name _x]] call EFUNC(common,serverEvent);
-    } else {
-        _playerOwnerID = ACE_player getVariable QGVAR(owner_id);
-        if (!isNil "_playerOwnerID" && _ownerID != _playerOwnerID) then {
-            _unitUID = getPlayerUID ACE_Player;
-            _drawPosVariableName = if (!isNil "_unitUID" && _unitUID != "") then {format [QGVAR(%1_DrawPos), _unitUID]} else {nil};
-            if (!isNil "_drawPosVariableName") then {
+_unitUID = getPlayerUID ACE_Player;
+_playerOwnerID = ACE_player getVariable QGVAR(owner_id);
+if (!isNil "_unitUID" && _unitUID != "" && !isNil "_playerOwnerID") then {
```
commy2: `!isNil "_unitUID" && _unitUID != ""` intended to use lazy eval?

addons/map_gestures/functions/fnc_transmit.sqf
```diff
@@ -28,18 +28,18 @@ if (!GVAR(EnableTransmit) || !visibleMap) exitWith {
...
+    _drawPosVariableName = format [QGVAR(%1_DrawPos), _unitUID];
+    {
+        _ownerID = _x getVariable QGVAR(owner_id);
+        if (isNil "_ownerID") then {
+            [EVENT_PLAYER_HAS_NO_OWNER_ID, [name _x]] call EFUNC(common,serverEvent);
+        } else {
+            if (_ownerID != _playerOwnerID) then {
                 _ownerID publicVariableClient _drawPosVariableName;
```
commy2: Why do we even have this in the first place? All transmitting should be done with the ACE event system.

MikeMatrix: It was the first port. Will be working on the event based system and some other optimizations for this soon'ish.

PabstMirror:
I guess we put this on back burner, sorry Sniperhid.
But I just tested and it didn't seem to be working (could be wrong, `getPlayerUID` makes it tough to test).

```
_playerOwnerID = ACE_player getVariable QGVAR(owner_id);
```
If that is nil, then we skip the for loop, but the only way we set that var is from an event inside the loop.

I think just adding this should fix:
```
if (isNil "_playerOwnerID") exitWith {
    [EVENT_PLAYER_HAS_NO_OWNER_ID, [name ACE_Player]] call EFUNC(common,serverEvent);
};
```

commy2:
Yeah, any scripting around `getPlayerUID` is bad. It's similar to addressing machines by their `profileName` or `name player`, or objects by their `vehicleVarName`.
The correct way is to use owner Id's and object references.

Glowbal:
Closing, merge conflicts, inactivity

PR #3069, opened 2015-12-22
rearm scope changed for dummy objects

GitHawk:
I changed the scope of the dummy objects so they are no longer listed in the editor.


jonpas:
:+1:

PR #146, opened 2015-02-14
Synced weapon heating, deterministic bullet dispersion

esteldunedain:
*WIP*

Continuing the ongoing effort to get bullet trajectories to behave as identical as possible on all clients. 

- Weapon temperatures are synced regularly for clients.
- Bullet dispersion changed from random to pseudo-random, to allow deterministic computing on each client.
- The pseudo-randomness is based on a synced list of random values and uses the remaining ammo count to lookup on that table. The index could be also salted using other deterministic things like the rounded bullet coordinates.

Based on https://github.com/KoffeinFlummi/ACE3/pull/143, hence the commits to common

PR #143, opened 2015-02-12
Modified setVariablePublic 

esteldunedain:
So it published inmediately if the variable hasn't being publishes recently. Then establish an embargo time before it can be published again.

The embargo time is removed/reset through a PFH. Note that, to improve performance, the embargo is only checked every 100 ms, so that's the lowest possible embargo (well, 0 should also work).

PR #3082, opened 2015-12-29
Deterministic overheating (updated)

### LINKED ISSUE #2978: Weapons don't overheat / overheat too slow
ACE3 Version: 3.4.1 RC

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* Respawn Module

**Description:**
* Weapons don't overheat anymore
* Swapping the barrel doesn't help

**Steps to reproduce:**
* Shoot a full belt of e.g. the Stoner 99 LMG `"LMG_Mk200_F"`
* Check weapon temperatur via the self interaction menu -> equipment

* Dedicated Server

**RPT log file:**
http://pastebin.com/1j5KraUb

PabstMirror:
Contains both of esteldunedain's PRs, I've tried to finish and modernized them:

 - #143 setVariablePublic:
Broadcasts change in a variable immedietly instead of waiting for a delay. I've rewritten it slightly to use waitAndExecute instead of creating new PFEHs.
 
 - #146 deterministic Overheating - That PR's notes:

Continuing the ongoing effort to get bullet trajectories to behave as identical as possible on all clients. 

- Weapon temperatures are synced regularly for clients.
- Bullet dispersion changed from random to pseudo-random, to allow deterministic computing on each client.
- The pseudo-randomness is based on a synced list of random values and uses the remaining ammo count to lookup on that table. The index could be also salted using other deterministic things like the rounded bullet coordinates.

Other changes:
 - Simplified weapon configs:
    ```
    ACE_MRBS = 3000; //Mean Rounds Between Stoppages (this will be scaled based on the barrel temp)
    ACE_Dispersion = 1;
    ACE_SlowdownFactor = 1;
    ```
    
 - Added ace setting for overheating Dispersion. Dispersion is now running on all clients, and is semi-costly. Disabling would just run the jamming simulation on the local player.
    
 - Now runs the heatup/cooldown simulation every 3 shots `(_unit ammo _weapon) % 3) != 0`
   Difference in heat between 3 shots is minor and get some performance savings, downside is 10/25/100 round magazines will be off by a small amount (10 round mags can't be fired that fast)

 - Attempting to update thermal numbers based on real physics.
~1.333x increased heating
~1.45x less cooling
Fix #2978


addons/overheating/functions/fnc_jamWeapon.sqf
```diff
@@ -9,74 +9,58 @@
...
+        && {currentMuzzle (_this select 1) in ((_this select 1) getVariable [QGVAR(jammedWeapons), []])}
+        && {!(currentMuzzle (_this select 1) in ((_this select 1) getVariable [QEGVAR(safemode,safedWeapons), []]))}
     };
-  };
 
-  _condition2 = {
-    currentWeapon (_this select 1) in ((_this select 1) getVariable [QGVAR(jammedWeapons), []])
-  };
+    private _statement = {
+        playSound3D ["a3\sounds_f\weapons\Other\dry9.wss", _this select 0];
```
jokoho48: why we only use Dry9 all Other sound are good to use here?
```
dry_1
dry1
dry2
dry3
dry4
dry5-rifle
dry6
dry7
dry8
dry9
```

commy2: Those are not for randomization, but specific weapons use a specific sound sample. We should probably read the sample from config (possibly with caching)

esteldunedain:
@PabstMirror, thanks for this! I wasn't able to complete this PR and it seems unlikely I will be able to any time soon. IIRC, I do have some pretty heavy optimizations done on my local branch. Could you take a look at them if I push it to some branch on the repo?

PabstMirror:
@esteldunedain Yes, if you can post them somewhere I will try to see what I can take from it.

esteldunedain:
> @esteldunedain Yes, if you can post them somewhere I will try to see what I can take from it.

NVM, the code was pushed to github already. I was refering to d001131, which greatly reduces the amount of billboards spawned, which had a measurable impact over fps on my rig. I was also pretty proud of d001131.

IIRC, the piece of code missing for this to work 100% properly was broadcasting the cooling down periodically. As it is now, the current weapon temp is only sent to remote clients when the weapon is fired or its temp checked. Hence if your weapon was hot and you let it cool down, the first shot will be deflected on remote clients but not on the local client. It's not a big deal though.

As a consequence of particle effects being optimized, I also intended to try enabling them for other nearby players, which cool be usefull for AARs.

esteldunedain:
@PabstMirror, I'm proposing a few changes on the code. Feel free to review and reverse.

The main change I'm proposing is ditching the time throttling restriction for publishing temperatures in favor of a tolerance restriction. e.g. only publishing the new temps if they are more than 50ºC different than the previous published value. This should lead to far lower updates when weapons are slowly cooling down, and none at all anymore when they reach ambient temperature. Plus this restriction is quite faster to evaluate. The packet published should now be smaller too (number instead of array)

I've also added a PFH to periodically update the player weapons temp as per the rationale I've commented above. Due to the new way temps are broadcasted, this should really not generate a measurable network traffic.

Finally, I've added the possibity of enabling particle effects for other nearby players (client settable, default off at least until the performance impact can be evaluated at large scale).

I've not yet tested this in any capacity, so beware. I'll probably do so tomorrow.

esteldunedain:
Yay, aside from a few typos it seems to work really well.

The thing I'm a bit hesitant about now are the changes to weapon configs. They made total sense 10 months ago, but probably changing them now would require updating third party stuff again. I think it'd be better to at least support the old entries, although it may not be easy to pick between the new entries inherited from parent weapons and the old entries defined down the inheritance chain.

BTW, the particle effects for nearby units looks really cool.

esteldunedain:
For consistency, I prefixed the configs to:

```c++
ACE_overheating_MRBS = 3000; 
ACE_overheating_dispersion = 1;
ACE_overheating_slowdownFactor = 1;
```

bux578:
my OCD ...
`ACE_overheating_mRBS = 3000;` & `ACE_overheating_dispersion = 1;` 
or
`ACE_overheating_MRBS = 3000;` & `ACE_overheating_Dispersion = 1;` 
?

VKing6:
Nah, MBRS works because it's an abbreviation, while number two there looks like it'd then have to be `ACE_Overheating_Dispersion` in that case.

esteldunedain:
I really hate `mRBS` with passion given it's an acronym; on the other hand our config entries are generally lowerCamelCase. 

So.. I don't know. Whatever upper management decides.

esteldunedain:
Ok, I agree with @bux578 now. Searched online and the standard seems to be that acronyms should be treated as words on variable names IF they have more than two letters. So:

```c++
ACE_overheating_mrbs = 3000; 
ACE_overheating_dispersion = 1;
ACE_overheating_slowdownFactor = 1;
```

VKing6:
Swapping barrels do not appear to affect the weapon's temperature atm.

bux578:
Dropping it on the ground might help ;)

esteldunedain:
> Swapping barrels do not appear to affect the weapon's temperature atm.

On this branch? I'd swear it worked when I tested it

esteldunedain:
That should fix it.

VKing6:
Yea that's better.
Speaking of swapping barrels, would it be possible to play the animation near the end of the timer as well, or otherwise extend the animation (playing it back slower, maybe)?

esteldunedain:
> peaking of swapping barrels, would it be possible to play the animation near the end of the timer as well, or otherwise extend the animation?

Actually we are playing two animations:
´´´sqf
_player playActionNow "GestureDismountMuzzle";
_player playAction "GestureMountMuzzle";
´´´
and **then** spawning the progressBar

Yeah, certainly not ideal!

I'll try putting the progressbar in between the two, which would be a small improvement.

> (playing it back slower, maybe)

Maybe, I don't know if that's possible.

PabstMirror:
Ed, I really need to take a closer look at all of this, but I'm not crazy about starting up an extra PFEH for cooldown.
At worst ~2 bullets would have their dispersion be off by a little from what it should be.

VKing6:
Also keep in mind there appears to be a GestureDismountMuzzleProne animation. The current animation does weird things to your upper body if you swap barrels while prone.

VKing6:
(I don't remember how to comment on file lines instead of commits):
https://github.com/acemod/ACE3/blob/deterministicOverheatingX/addons/overheating/CfgWeapons.hpp#L40

I feel the MXSW shouldn't have a replaceable barrel. It's closer to an IAR than a SAW and appears to have a fixed barrel like the former.

esteldunedain:
@VKing6, there it goes: improved animations (slower and with a shorter progress bar in between). I'll take a look at the prone animation next

@PabstMirror, I get what you say. Another altenative, which might be faster is to add a self calling `waitAndExecute`. That should have no performance on the normal frames due to being pushed back by the sort you implemented. It would only have a small perf hit on one frame every five secs. What do you think?

VKing6:
You might want to inherit the animation instead of making the vanilla animation slower. Other mods may use that animation.

esteldunedain:
> You might want to inherit the animation instead of making the vanilla animation slower. Other mods may use that animation.

True, I'll try that.

esteldunedain:
@PabstMirror, I'm pretty sure this is much better now.

BTW, any high interval PFH should probably be changed to this pattern.

VKing6:
Okay having played with it a bit it feels pretty good. I think weapons in general could heat up slightly faster, especially weapons without heavy barrels like regular rifles, but it's not bad (the point of this entire thing, isn't it :p).

A bigger firefight will have to be had to make sure.

esteldunedain:
> Okay having played with it a bit it feels pretty good.

:+1:

esteldunedain:
@VKing6:

> You might want to inherit the animation

Done.

> GestureDismountMuzzleProne 

The classes exist on the config, but there's no actual animation data for them.

VKing6:
:|

esteldunedain:
I think this is pretty much done. In the future I'd like to keep track of how much dust the weapon has gathered since last cleaned and increase the jam chance based on that, but it may be better done and discussed on a sepparate PR.

PabstMirror:
This is looking good,

One minor thing to clear up is what weapon types we want to handle, right now firedEH only runs on `primaryWeapon` and `handgunWeapon` whereas updateTemperatureThread runs on all 3 types.

I assume a rpg-7 tube can heat-up but our simulations really only work for bullets, so I think that's safe to filter out.

I don't think you can really overheat a pistol, some mods put SMGs in the handgun slot, but we still might be ok filtering these out as well.

Also I think we can simplify updateTemperatureThread to only run on `currentWeapon`

esteldunedain:
> One minor thing to clear up is what weapon types we want to handle, right now firedEH only runs on primaryWeapon and handgunWeapon whereas updateTemperatureThread runs on all 3 types.

Totally right. Only handgun + primary should be enough

> Also I think we can simplify updateTemperatureThread to only run on currentWeapon

True.

I'll implement both changes.

PR #3084, opened 2015-12-30
CBA zeus integration

### LINKED ISSUE #2937: Test effects of CBA's zeus keybindings / events
Just making note for myself:

https://github.com/CBATeam/CBA_A3/pull/206

https://github.com/acemod/ACE3/blob/master/addons/interact_menu/XEH_clientInit.sqf#L83-L105

And new event may be able to replace PFEH state checker

PabstMirror:
Close #2937

CBA now adds keyDown EH to zeus interface, so #1951 (CTRL+V from climb blocking paste in zeus) is back.

 - Adds "isNotInZeus" to caninteractWith conditions.
 - Removed code to forward interaction keybind to CBA  as this is handled by CBA
 - Use `CBA_curatorOpened/Closed` events from CBA to trigger ACE's zeus event (trivial savings from removing constant checking in PFEH).

@commy2 - tested with this code and it seems to handle scripted closing fine.
```
[] spawn {sleep 2; systemChat "scipted closing zeus"; (findDisplay 312) closeDisplay 1};
```

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -1,7 +1,7 @@
 // ACE - Common
+// #define ENABLE_PERFORMANCE_COUNTERS
 #include "script_component.hpp"
 
-// #define ENABLE_PERFORMANCE_COUNTERS
 
 //////////////////////////////////////////////////
 // PFHs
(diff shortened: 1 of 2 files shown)


SilentSpike:
Looks good upon initial inspection :+1:

commy2:
>(findDisplay 312) closeDisplay 1
That works, because 1 and 2 emulate closing with either OK (1) or Cancel (2). Closing with any other exit code might however not work.
I tested with ESC (by default exits with no code / default 0), but it seems that zeus handles that with a key handler or some hard coded magic.

:+1:

SilentSpike:
>it seems that zeus handles that with a key handler or some hard coded magic

iirc it's using a key handler

PR #3085, opened 2015-12-30
Explosive enhancements

bux578:
When merged this pull request will
* remove the invalid `ACE_offset` config entry (which was ported from AGM but was never used)
* reduce the interaction distance from 5 to 1, so as an EOD you really have to get close to the explosives to arm or defuse them
* move interaction points for some explosives to more logical positions
* add a config setting where to attach the defuse helper object (same place as the interaction point in most cases, but could be different for other (e.g. modded) explosives)

Pinging the [maintainers](https://github.com/acemod/ACE3/blob/master/addons/explosives/README.md#maintainers) @esteldunedain and @CorruptedHeart

### DIFF
--- a/addons/explosives/CfgAmmo.hpp
+++ b/addons/explosives/CfgAmmo.hpp
@@ -21,7 +21,6 @@ class CfgAmmo {
 
     class DirectionalBombCore: TimeBombCore;
     class DirectionalBombBase: DirectionalBombCore;
-    class APERSTripMine_Wire_Ammo: DirectionalBombBase;
 
     class SLAMDirectionalMine_Wire_Ammo: DirectionalBombBase;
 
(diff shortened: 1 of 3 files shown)


CorruptedHeart:
Looks good from my side.

Just in case anyone was wondering what the ACE_Offset was for, it was for offsetting the visual item when placing. Because, for instance, with the tripwire it can (could, maybe it's changed) be challenging to place it properly due to not being able to see one of the the edge of the explosive when placing due to the size.

bux578:
> it was for offsetting the visual item when placing
That could be useful for ACE aswell
When placing the tripwire you currently see just wire and not the mine itself.

CorruptedHeart:
Seems to be merge conflicts now.

bux578:
@CorruptedHeart fixed

CorruptedHeart:
Since there doesn't seem to be any other objections to this, I will press the merge button then.

jonpas:
This should be merged to `release` not `master`, it's tagged for 3.4.2.

PR #3086, opened 2015-12-30
Remove obsolete hotkey entries

bux578:
This has been used in AGM and is not needed anymore. At least I wasn't able to find a place where hotkeys are being read and actively used.
Someone correct me if I'm mistaken.


jokoho48:
:+1: but its not enhancement its cleanup

PR #3099, opened 2016-01-01
Kill Spectator PFEH when exiting spectator

### LINKED ISSUE #2989: Script error from spectator
ACE3 Version: 3.4.1 stable

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:** N/A

**Description:**
The code below produces this script error
```
14:20:32 Error in expression <
ace_spectator_camPos = [_x,_y,_z max (getTerrainHeightASL [_x,_y])];
_camera>
14:20:32   Error position: <getTerrainHeightASL [_x,_y])];
_camera>
14:20:32   Error Type Number,Not a Number, expected Number
14:20:32 File z\ace\addons\spectator\functions\fnc_handleCamera.sqf, line 32
```

**Steps to reproduce:**
Run this code in editor
```
[true] call ace_spectator_fnc_setSpectator;
[player, true] call ace_spectator_fnc_stageSpectator;
[{
    [false] call ace_spectator_fnc_setSpectator;
    [player, false] call ace_spectator_fnc_stageSpectator;
}, [], 2] call ACE_common_fnc_waitAndExecute;
```

Editor

PabstMirror:
Should fix #2989

There is a race condition between these 4 PFEH and the "onUnload" event
in handleInterface.
If the PFEHs run first they will use nil variables and throw a script
error.
This sets them to nil immediately when exiting spectator

@SilentSpike - I'm not 100% sure this is the right way to fix, but it seems to work.

Debug of the race condition:
```
handleCamera Frame:35839 (normal)
end spectator Frame:35840
handleCamera Frame:35840 (throws error)
Error in expression ... File z\ace\addons\spectator\functions\fnc_handleCamera.sqf, line 35
Killing CamHandler in handleInterface: onunload Frame:35840
handleCamera Frame:35841 (exits cleanly)
```

### DIFF
--- a/addons/spectator/functions/fnc_setSpectator.sqf
+++ b/addons/spectator/functions/fnc_setSpectator.sqf
@@ -127,6 +127,12 @@ if (_set) then {
     GVAR(unitCamera) = nil;
     GVAR(targetCamera) = nil;
 
+    //Kill these PFEH handlers now because the PFEH can run before the `onunload` event is handled
+    GVAR(camHandler) = nil;
+    GVAR(compHandler) = nil;
+    GVAR(iconHandler) = nil;
+    GVAR(toolHandler) = nil;
+
     // Cleanup display variables
     GVAR(ctrlKey) = nil;
     GVAR(heldKeys) = nil;


SilentSpike:
Oh man am I glad you managed to catch this one, I'm not the greatest at spotting potential race conditions :smile:

You're on fire lately pabst, nice job :+1:

PR #3100, opened 2016-01-01
Frag - only run addPfhRound on ammo that will frag

PabstMirror:
If spall is not enabled (default) then calculate which ammo types will
actually cause a frag event, cache and then only track those.

before: `(#2): 0.0883651ms  - CODE: {_this call ace_frag_fnc_fired}`
after: `(#2): 0.0179289ms  - CODE: {_this call ace_frag_fnc_fired}`

More importantly this skips running frag simulation (tracking bullet position in a PFEH) for rounds that will never trigger a frag event.

### DIFF
--- a/addons/frag/XEH_postInit.sqf
+++ b/addons/frag/XEH_postInit.sqf
@@ -12,5 +12,5 @@ if(isServer) then {
 [FUNC(masterPFH), 0, []] call CBA_fnc_addPerFrameHandler;
 
 //Cache for ammo type configs
-GVAR(cacheRoundsTypesToTrack) = createLocation ["NameVillage", [-10000,-10000,-10000], 0, 0];
+GVAR(cacheRoundsTypesToTrack) = createLocation ["ACE_HashLocation", [-10000,-10000,-10000], 0, 0];
 GVAR(cacheRoundsTypesToTrack) setText QGVAR(cacheRoundsTypesToTrack);
--- a/addons/frag/XEH_postInit.sqf
+++ b/addons/frag/XEH_postInit.sqf
@@ -10,3 +10,7 @@ if(isServer) then {
 };
 
 [FUNC(masterPFH), 0, []] call CBA_fnc_addPerFrameHandler;
+
+//Cache for ammo type configs
+GVAR(cacheRoundsTypesToTrack) = createLocation ["NameVillage", [-10000,-10000,-10000], 0, 0];
+GVAR(cacheRoundsTypesToTrack) setText QGVAR(cacheRoundsTypesToTrack);
--- a/addons/frag/functions/fnc_fired.sqf
+++ b/addons/frag/functions/fnc_fired.sqf
@@ -1,8 +1,56 @@
+/*
+ * Author: nou, jaynus, PabstMirror
+ * Called from FiredBIS event on AllVehicles
+ * If spall is not enabled (default), then cache and only track those that will actually trigger fragmentation.
+ *
+ * Arguments:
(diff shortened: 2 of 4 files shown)


addons/frag/XEH_postInit.sqf
```diff
@@ -9,4 +9,8 @@ if(isServer) then {
     [QGVAR(frag_eh), { _this call FUNC(frago); }] call EFUNC(common,addEventHandler);
 };
 
-[FUNC(masterPFH), 0, []] call CBA_fnc_addPerFrameHandler;
\ No newline at end of file
+[FUNC(masterPFH), 0, []] call CBA_fnc_addPerFrameHandler;
+
+//Cache for ammo type configs
+GVAR(cacheRoundsTypesToTrack) = createLocation ["NameVillage", [-10000,-10000,-10000], 0, 0];
+GVAR(cacheRoundsTypesToTrack) setText QGVAR(cacheRoundsTypesToTrack);
```
bux578: Is this needed?

esteldunedain:
:+1: Looks good. I'm intriged too about that `setText` @bux578 pointed out.

addons/frag/functions/fnc_fired.sqf
```diff
@@ -1,8 +1,56 @@
...
+        TRACE_2("SettingCache[spallEnabled]",_type,_shouldAdd);
+        GVAR(cacheRoundsTypesToTrack) setVariable [_type, _shouldAdd];
+    };
+
+    //Read configs and test if it would actually cause a frag, using same logic as FUNC(pfhRound)
+    private _skip = getNumber (configFile >> "CfgAmmo" >> _type >> QGVAR(skip));
+    private _explosive = getNumber (configFile >> "CfgAmmo" >> _type >> "explosive");
+    private _indirectRange = getNumber (configFile >> "CfgAmmo" >> _type >> "indirectHitRange");
+    private _force = getNumber (configFile >> "CfgAmmo" >> _type >> QGVAR(force));
+    private _fragPower = getNumber(configFile >> "CfgAmmo" >> _type >> "indirecthit")*(sqrt((getNumber (configFile >> "CfgAmmo" >> _type >> "indirectHitRange"))));
```
jokoho48: why not caching this vales? i think that could improve the performance too

PabstMirror: I saw this as a simple way to make the common case fast (I'm guessing I shoot about 60x more bullets that throw frags).

It's an augment, not a replacement for the full overhaul (#2584).

addons/frag/XEH_postInit.sqf
PabstMirror: @NouberNou might know, I just borrowed this from ACRE / Carma

ref https://github.com/NouberNou/carma2/blob/master/rv/addons/lib/XEH_pre_init.sqf#L8-L9

addons/frag/functions/fnc_fired.sqf
jokoho48: #2584 is outdated and crap i think i will retouch it later if i have some more time

addons/frag/XEH_postInit.sqf
bux578: argh, I totall missed the Q infront of the QGVAR.
Won't hurt, then

jonpas: Doesn't that write text on the map? Carma could have it for debug purposes.

bux578: at -10000 -10000 -10000 though

jonpas: True.

Issue #734, opened 2015-04-18
Practice Ammunition


bux:
Add practice ammunition that allow proper TvT training scenarios with hit display.

http://simunition.com/en/

> ulteq [13:06] 
> I think we could add those easily
> Reduced muzzle velocity, relatively high airFriction, hit value of 0
> And for the visual feedback we would have to add a special hit eventhandler on the players
> Which checks if you got hit by it, and then shakes the camera a bit
> We could add a module for it

bux578:
@ulteq

commy2:
Flummi is gonna be so mad about this.

Heavygunner96:
I don't really get the purpose of this request but neither I understand why Flummi would be mad about it?^^

KoffeinFlummi:
https://github.com/acemod/ACE3/issues/414 ;)

bux578:
> #414 

That's the link for the public feature requests.

alganthe:
someone didn't read #414

Mike-MF:
Closing due to age, inactivity or already solved. May be re-opened if the issue is still present, even better just open a new up-to-date Issue.

PR #3103, opened 2016-01-01
Practice Ammunition

bux578:
#734


esteldunedain:
:+1: Not exactly realistic, but nice to have.

jonpas:
Rubber bullets! Very useful for us, thanks!

esteldunedain:
> Rubber bullets

I believe those go substantially slower than the real thing

jonpas:
Shhhh :grin:

VKing6:
I've never seen the point of practice ammo in a game (isn't it all practice, in the perspective of not shooting each other with real guns?), but I know some people who will find it useful.
Just don't add blanks, please?

bux578:
> I've never seen the point of practice ammo in a game

PvP CQC/CQB Training. Remove the need for `allowdamage false` or to revive/respawn after each round of training. That's what I was thinking about.

VKing6:
Do you intend to make it work with the medical system to make people react to being hit?

bux578:
If I can manage to do so, yeah. CamShake, OnScreen message... something like that. Feel free to give  ideas

esteldunedain:
> If I can manage to do so, yeah. CamShake, OnScreen message... something like that. Feel free to give ideas

What about calling `ace_hitreactions_fnc_fallDown`? It already does cam shake for the player, and a fall animation for the shooter to see. Otherwise, a simple camShake like the one on that function could work.

Anyway, I think you'll need a config an entry defining the bullet as blank (to be on the safe side, and don't rely on `hit = 0`). I think you should check for that [here](https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_handleDamage.sqf#L38).

headedhorseman101:
What might be a good idea is to knock the target out upon the impact of a certain number of hits (could be a settable value possibly?) which would allow oversight from someone in Zeus or an moderator of some sort to see what happened in the scenario easier. That way there isn't any damage done either.

Daishiknyte:
Alternative use:  Stun rounds for stopping units from running around when you're trying to capture one.

Glowbal:
@bux578 should we get this in in the current state or wait until a more polished version?

bux578:
@Glowbal this is far from being working. I haven't had the time yet to do anything here.

bux578:
I'll close this one. Even so I'd like to see that feature. I currently don't have the time.

PR #3106, opened 2016-01-01
Explosive events, fix #3104

### LINKED ISSUE #3104: Please provide a listenable event to explosives being set and removed.
This would be usefull for tracking explosive placements for example AAR purposes.

bux578:
* add 2 new events: `defuse` and `explodeOnDefuse`
* add `_unit` to place event

Can be tested with:

```sqf
["ace_explosives_place", {
    hint format ["Placed:\nexpl: %1\nunit: %2", _this select 0, _this select 3];
}] call ace_common_fnc_addEventHandler;

["ace_explosives_defuse", {
    hint format ["Defused:\nexpl: %1\nunit: %2", _this select 0, _this select 1];
}] call ace_common_fnc_addEventHandler;

["ace_explosives_explodeOnDefuse", {
    hint format ["ExplodedOnDefuse:\nexpl: %1\nunit: %2", _this select 0, _this select 1];
}] call ace_common_fnc_addEventHandler;
```

### DIFF
--- a/addons/explosives/functions/fnc_defuseExplosive.sqf
+++ b/addons/explosives/functions/fnc_defuseExplosive.sqf
@@ -22,6 +22,8 @@ TRACE_2("params",_unit,_explosive);
 if (GVAR(ExplodeOnDefuse) && {(random 1.0) < (getNumber (ConfigFile >> "CfgAmmo" >> typeOf _explosive >> "ACE_explodeOnDefuse"))}) exitWith {
     TRACE_1("exploding on defuse",_explosive);
     [_unit, -1, [_explosive, 1], true] call FUNC(detonateExplosive);
+    [QGVAR(explodeOnDefuse), [_explosive, _unit]] call EFUNC(common,globalEvent);
 };
 
 _unit action ["Deactivate", _unit, _explosive];
+[QGVAR(defuse), [_explosive, _unit]] call EFUNC(common,globalEvent);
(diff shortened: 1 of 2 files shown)


esteldunedain:
:+1: Looks good

jonpas:
:+1:

Cuel:
shouldn't it be placed/defused/explodedOnDefuse ? :smile:

CorruptedHeart:
It appears some listenable events are listed [here](http://ace3mod.com/wiki/framework/events-framework.html) this would probably need to be added there.

But there is both past tense and current tense words used in the events listed there so probably either naming scheme is fine but we might want to be consistent at some point.

jonpas:
@CorruptedHeart This list is generated with a script, events should be added to module-specific framework page, they get pulled into one big list on events framework page then.

CorruptedHeart:
Alright, I guess we still need to have a discussion about whether to keep them named like they are now or past tense.

jonpas:
I was always thinking that those that are there to signal that something happened (listenable) are named in past tense, and those that can be used to perform a certain action (callable) are like those in this PR.

Cuel:
Agree with @jonpas. 
Furthermore why not throw a "ace_explosives_detonated" in? I was just looking for that

Bummeri:
Is it possible to provide that event that Cuel was speaking about?

esteldunedain:
> Is it possible to provide that event that @Cuel was speaking about?

It would be tricky, cause it would work with explosives detonated by remote control, but not for those that explode because of proximity, being damaged, etc. IMO it's not worth it with this limitations.

PR #3109, opened 2016-01-03
fix ThingX capitalization, fix #3076

### LINKED ISSUE #3076: ThingX vs thingX
ThingX is the correct version, but we keep using the lower case for some reason.
("Bug" in CBA, fixed soon)

commy2:


jokoho48:
:+1:

PR #3110, opened 2016-01-03
fix usage of deprectated fired XEH, fix #3088

### LINKED ISSUE #3088: Deprecated Fired XEH in goggles
```
[["CAManBase","Fired",{private _this = [_this select 0, _this select 1, _this select 2, _this select 3, _this select 4, _this select 6, _this select 5];if (local (_this select 0)) then {_this call ace_goggles_fnc_handleFired};},true,[]]]
```

Should be firedBIS

commy2:


addons/goggles/CfgEventHandlers.hpp
```diff
@@ -14,15 +14,15 @@ class Extended_PostInit_EventHandlers {
...
         };
     };
 };
 
-class Extended_Fired_EventHandlers {
+class Extended_FiredBIS_EventHandlers {
     class CAManBase {
         class ADDON {
-            fired = QUOTE(if (local (_this select 0)) then {_this call FUNC(handleFired)});
+            clientFiredBIS = QUOTE(if (local (_this select 0)) then {_this call FUNC(handleFired)});
```
jonpas: Do they still need `local (_this select 0)` check if it's client only?

PabstMirror: `clientXEH_Event` still runs on all clients, it just removes the EH from dedicated servers

The effective result is something like this:
```
if (local (_this select 0)) then {
    _this call {
        params ["_unit"];
        if (_unit != ACE_player) exitWith {true};
        //CODE
    };
};
```
Right now we are doing `select 0` twice (`local` chceck and later in `param`); So we could remove the local check and I think it should be roughly the same speed?

Personally I like having the XEH lines clean and moving all code to SQF.

But all of this is a more general problem that can be taken care of at a later time.

commy2: clientX XEH is executed on every machine where `!isDedicated` is true.

A local prefix would be cool too (localFiredBIS, localGetIn, localEngine), but it doesn't really make sense on some events that are always limited to local machines. (respawn, killed, weaponRested)

jonpas: I see :+1:

jokoho48:
:+1:

PR #1219, opened 2015-05-15
Daisy Chain PR

LorenLuke:
Enables Daisy Chaining of explosives together.

PR #3119, opened 2016-01-05
Add action to detonate all explosives on a detonator

### LINKED ISSUE #2108: Simultaneous explosive detonation
Was playing with mode for a while. Found quite annoying thing about explosives - it is not possible to detonate several charges simultaneously with one player. It ruins ambushes sometimes... =(
Is it a realism feature? It would be nice if in next version there would be a menu that will pop-up on charge setup so you can assign it to different detonation groups.
The best idea is just to have a popup window after chosing detonator where you can choose radio channel used for detonation and then to have possibility to choose detonator channel.

Also have a small question about Claymore mine... is it for realism purposes that auto detonation by walking enemies is deleted?

VKing6:
This PR will add an action to the detonator menu to detonate all explosives associated with it.
Example can be seen [here](http://i.imgur.com/oGUEKTl.jpg?1)

This fixes #2108 and will probably deprecate #1219.

### DIFF
--- a/addons/explosives/XEH_preInit.sqf
+++ b/addons/explosives/XEH_preInit.sqf
@@ -28,6 +28,7 @@ PREP(canDefuse);
 PREP(canDetonate);
 PREP(defuseExplosive);
 PREP(detonateExplosive);
+PREP(detonateExplosiveAll);
 PREP(dialPhone);
 PREP(dialingPhone);
 
--- a/addons/explosives/functions/fnc_addDetonateActions.sqf
+++ b/addons/explosives/functions/fnc_addDetonateActions.sqf
@@ -43,10 +43,10 @@ _explosivesList = [];
                         {(_this select 2) call FUNC(detonateExplosive);},
                         {true},
                         {},
-                        [ACE_player,_range,_x]
+                        [_unit,_range,_x]
                     ] call EFUNC(interact_menu,createAction),
                     [],
-                    ACE_Player
+                    _unit
                 ];
         };
     };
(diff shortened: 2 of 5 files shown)


commy2:
Aren't all these `ACE_player` supposed to be `_unit`?

VKing6:
Maybe? I honestly have no idea

CorruptedHeart:
The `ACE_Player` should probably be `_unit` although for the UI part of explosives it can only ever really be done by the player. However, it was designed in such a way that it doesn't necessarily have to be.

BaerMitUmlaut:
In what case would `_unit` not be the same as `ACE_player`? I mean we're talking about the *self*-interaction menu.

CorruptedHeart:
Well if there ever becomes a case where you'd want to detonate someone's explosives or something.

BaerMitUmlaut:
The function called by the self-interaction menu still receives `ACE_player` as a param. Or am I misunderstanding something here?

Glowbal:
If it does, all code in this function should probably use `_unit`. No need to provide a unit as the input parameters and still use `ACE_player` in the code.

commy2:
Not using ACE_player inside functions also makes it easier to eventually port them to AI. That's why I'm asking.

VKing6:
At first test there doesn't appear to be any problems changing `ACE_Player` to `_unit`.

jonpas:
Do it then. :)

rlex:
Thanks for that feature, i know some of my squadmates asked for it.
Note on translation:
"Подрыв Все" is a bit wrong in russian. "Подрыв всех" is better, but still sounds a bit weird.

esteldunedain:
:+1: to merge

VKing6:
> "Подрыв Все" is a bit wrong in russian. "Подрыв всех" is better, but still sounds a bit weird.
I'll change it. If you can think of a better phrase feel free to suggest it.

Glowbal:
Thanks @VKing6

PR #3120, opened 2016-01-05
Clean up explosive interaction menu

VKing6:
Move detonate actions from "Detonate" layer to "Explosives" layer and remove inconsistent ">>" from explosive actions.
Reasoning is that the amount of detonators that someone can reasonably be expected to carry isn't large enough to warrant an extra layer to the interaction menu.


commy2:
:+1:

esteldunedain:
:+1:

Issue #3121, opened 2016-01-06
The AI is not attacking the nearest unconscious enemies.


mickeymen:
ACE3 Version: 3.4.1
Stable 1.54

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
Basic and Advanced Medical

**Description:**
Often seen such a situation. The AI hurt the player(at close range several meters) and the player loses consciousness. At this moment the AI turns his back to the unconscious player and stands like a statue or shoot other units. After a few seconds the injured player regain your consciousness and  can easily kill this "merciful" AI. I think, at least this situation is ridiculous. In order to avoid such moments the AI should always try to finish off the unconscious opponent who is close. Especially if the AI itself sent him to unconsciously condition.

**Steps to reproduce:**
Play in ACE3

**Where did the issue occur?**
Editor

**RPT log file:**
Not needed

Glowbal:
This is actually by design to avoid AI that keep targetting an unconscious player. 

I think this behaviour balances well with how unconscious AI vs players plays out imo. I've seen plenty of times where an AI went unconscious and players left him for dead.

mickeymen:
Yes probably this is not a bug, but sometimes this detail creates not a realistic situations for the player. In ACE3, the player too quickly regain your consciousness, and also AI doesn't kill him in this situation. Thus the player after the loss of consciousness has the some superiority vs AI, who wouldn't consider it as a target

Glowbal:
Perhaps a better idea is to increase the time it takes for a player to regain consciousness. Add some kind of fade in effect.

PabstMirror:
Main problem is that `setCaptive true` has fairly fast effects but `setCaptive false` can take a while for the AI to re-engage.

Also, humans generally don't stop shooting at something just because it falls over. we could make the whole captive thing a setting that we could disable.  The downside is that a enemy patrol that goes near a unconsciousness body would start engaging it until they magically know he's dead which isn't realistic. 

I do like the idea of making waking up from unconsciousness a much slower process, and make the player completely worthless at fighting for ~15 seconds, which should be enough time for the AI to see as a threat again.

esteldunedain:
> I do like the idea of making waking up from unconsciousness a much slower process, and make the player completely worthless at fighting for ~15 seconds, which should be enough time for the AI to see as a threat again.

[...]

PR #3124, opened 2016-01-06
add delay to Unconsciousness Captivity

jokoho48:
add delay to Unconsciousness Captivity(posible fix for #3121)


addons/medical/functions/fnc_setUnconscious.sqf
```diff
@@ -94,8 +94,16 @@ _unit setUnitPos "DOWN";
 if (GVAR(moveUnitsFromGroupOnUnconscious)) then {
     [_unit, true, "ACE_isUnconscious", side group _unit] call EFUNC(common,switchToGroupSide);
 };
+// Delay Unconscious so the AI dont instant stop shooting on the unit #3121
+if (GVAR(delayUnconCaptive) == 0) then {
+    [_unit, QGVAR(unconscious), true] call EFUNC(common,setCaptivityStatus);
+} else {
+    [{
+        [_this select 0, QGVAR(unconscious), true] call EFUNC(common,setCaptivityStatus);
```
PabstMirror: Need to make sure they are still unconc at this point, if delayUnconCaptive > (wake up time)

jokoho48: oh god point i missed that

addons/medical/ACE_Settings.hpp
```diff
@@ -223,4 +223,11 @@ class ACE_Settings {
         values[] = {CSTRING(useSelection), CSTRING(useRadial), "Disabled"};
         isClientSettable = 1;
     };
+
+    class GVAR(delayUnconCaptive) {
+        displayName = CSTRING(delayUnconCaptive);
+        typeName = "SCALAR";
+        value = 10;
+        values[] = {CSTRING(painEffect_Flash), CSTRING(painEffect_Chroma)};
```
esteldunedain: Not needed

jokoho48: oh right :)

captainblaffer:
I don't know if this feature is good for gameplay.

For missions without revive module, which are quite hardcore players will be less likely to survive. Which imo is not fun. They die, medics don't have anything to do.

for missions with revive, it makes 0 sense for the AI to keep shooting the player for 10 seconds as they can't die anyway. It will only cause the player to get hurt more requiring more bandages(annoying). Or the AI will keep shooting the unc. player instead of the alive player standing next to them, which will make the AI look dumb.

jokoho48:
@captainblaffer you can easy disable this if you replace the number with a 0 :D

captainblaffer:
I understand, but that is quite inconvenient for my community at least. I hoped if the general opinion is in line with my thoughts, to set the default value at 0.

addons/medical/CfgVehicles.hpp
```diff
@@ -138,6 +138,13 @@ class CfgVehicles {
                 typeName = "BOOL";
                 defaultValue = 1;
             };
+
+            class delayUnconCaptive {
+                displayName = CSTRING(MedicalSettings_delayUnconCaptive_DisplayName);
+                description = CSTRING(MedicalSettings_delayUnconCaptive_Description);
+                typeName = "NUMBER";
+                defaultValue = 10;
+            }
```
Glowbal: Missing `;`

Glowbal:
I am ok with including this feature however:
- Please finish the translations. Also do mention that value of the setting is in seconds.
- I am debating what the default value for this setting should be. We will discuss in Slack @jokoho48

esteldunedain:
> I am debating what the default value for this setting should be

My opinion is default off

jokoho48:
or we make it 3 secs

esteldunedain:
> or we make it 3 secs

Yeah, that could be about right too.

SilentSpike:
I mentioned it on the focus feature issue, but I think having a setting for this is way too specific, if the default value is set right to begin with then a setting isn't meaningful to the average user.

commy2:
^ No setting needed for this

Glowbal:
Yes, I am with you on that @SilentSpike. I think using 3 seconds is plenty. We can adjust based upon user feedback.

mickeymen:
I understand the lack of delay causes the ability of the AI to be merciful, to an unconscious opponent! 
#3121
Do I understand correctly? Thus, please add the short delay finally and solve the problem of the stupid behavior of the AI in the ACE3!  what are the obstacles to this?

Issue #1708, opened 2015-06-21
Weather propagation breaks wind


BaerMitUmlaut:
ACE3 Version: 3.1.1

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* ACE Weather

**Description:**
When setting ACE Weather to No (every other setting of the weather to yes), every few seconds when the value gets a adjusted it goes back to 0m/s before going to the new value. For example, if the wind strength is at 2.4m/s then it goes down to 0 and back up to 2.8m/s.
You can test this by executing the following code in the debug console:
`0 spawn {while {true} do {hintSilent str (vectorMagnitude wind); sleep 0.1}};`
I have also recorded it on video for better illustration: http://a.loveisover.me/rdpuij.webm
This does not happen with ACE Weather set to yes or weather propagation set to no. It also doesn't happen every time from my observations.

**Steps to reproduce:**
* Place down ACE Weather module
* Set ACE Weather to no
* Execute `0 spawn {while {true} do {hintSilent str (vectorMagnitude wind); sleep 0.1}};`

**Where did the issue occur?**
Editor & dedicated

**RPT log file:**
None

ulteq:
What was your `serverUpdateInterval` weather module setting?

BaerMitUmlaut:
`serverUpdateInterval` was left at the default 60. No setting was changed apart from ACE Weather (`useACEWeather`).

Kilo1-1:
I've also seen this happen.  I also have seen it happen where there is heavy wind yet the Kestrel says the wind is 2.4m/s or 2.5m/s when the wind is much stronger than that.  I have also seen it where the wind gets up to 70m/s or 80m/s.  This is with the same settings that is in the OP.

It just seems the wind sync is really random at times.

brokenhands:
I had major wind issues on my mission that I recently resolved. The map is Chernarus and we were noting winds at the freeway speeds mentioned. The easiest test for us detonating a vehicle and watching the flames burn parallel to the ground chasing the wind. 

The fix for was removing an ALiVE weather module.

Kilo1-1:
I thought it was the ALIVE weather module too because it happened when we used that module in our missions so we have since not used it  However the issue still comes up when that module isn't used.  I am not sure what is causing it.

ViperMaul:
Status?

PabstMirror:
@BaerMitUmlaut 
[...]

PR #1308, opened 2015-05-19
Scaled wind arrow for color blind players.

ulteq:
Scaled wind arrow for color blind players. For reference: https://github.com/acemod/ACE3/issues/1272

It works, but it's not very pretty.

PR #3126, opened 2016-01-07
Weather cleanup, fix wind inversion, add non-color based wind speed indicator

### LINKED ISSUE #1052: Potential bug in the weather module (custom settings)
Module Settings:

Watch the trees: https://www.youtube.com/embed/SWRsWKwrTBY

Can anyone reproduce the random wind direction changes?

### LINKED ISSUE #1272: Discussion: Wind Direction and Speed
Currently there's exists the wind arrow (Shift+K) for determining wind direction and speed. However, rather than using color, would it be possible to use a magnitude for the arrow for any users who may be colorblind?

PabstMirror:
- Privates, Headers, Formatting
 - Merge 4 PEFH with >1 second delay into a single one (should be faster)
 - Fix wind inverting each server controller update. (Was setting wind dir to the movement direction and not the source direction, so the wind flipped each update) Fix #1052 #1708 
 - Add non-color based indicator for wind speed (for color blind). Replace PR #1308 - Close #1272

Based on VKing6 comment https://github.com/acemod/ACE3/pull/1308#issuecomment-103847302
Dots represents the Beaufort number
image


bux578:
very :+1:

esteldunedain:
:+1:

Gwynblade:
These dots are a great addition, I didn't even notice the color changing on the arrow so had no clue it meant something. One more thing that would be great if you find the chance to add (even if not by default, but as an option) - Arrow fading off after couple of seconds by itself. You press shift+k arrow shows up, you see the wind, take aim, arrow fades away without you pressing anything again.

PR #3127, opened 2016-01-07
Reset make.py Cache on new build version

### LINKED ISSUE #1127: make.py ignores version change and skips outdated modules rebuild
ACE3 Version: 3.0.0.2 ab580dd1c7bb0dba710a42f0193bb312eceac0c0

**Mods:**
* `@cba_a3`
* `@ace3`

**Placed ACE3 Modules:**
N/A

**Description:**
Build with make.py doesn't take into account to verify ACE3 version number and does NOT rebuild the modules (PBOs) at all, which makes them unusable on the server since it will report that the modules are outdated.

**Steps to reproduce:**
* Build with make.py
* Change ACE3 version number
* Build with make.py again
* See that modules weren't updated with new version number and render "outdated" message on Servers

N/A

**RPT log file:**
N/A

PabstMirror:
Replace #3028 (that was on release branch)

Close #1127
If the version (from main) changes, reset cache
Same effect as "force"

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -975,6 +975,18 @@ See the make.cfg file for additional build options.
         print ("No cache found.")
         cache = {}
 
+    # Check the ace build version (from main) with cached version - Forces a full rebuild when version changes
+    aceVersion = get_project_version()
+    cacheVersion = "None";
+    if 'cacheVersion' in cache:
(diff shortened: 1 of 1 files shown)


Glowbal:
Awesome! :+1:

PR #3129, opened 2016-01-07
Frag - only run addPfhRound on ammo that will frag (for master)

PabstMirror:
Redo #3100 for master

If spall is not enabled (default) then calculate which ammo types will
actually cause a frag event, cache and then only track those.

before: `(#2): 0.0883651ms  - CODE: {_this call ace_frag_fnc_fired}`
after: `(#2): 0.0179289ms  - CODE: {_this call ace_frag_fnc_fired}`

More importantly this skips running frag simulation (tracking bullet position in a PFEH) for rounds that will never trigger a frag event.

Adds a `ACE_HashLocation` to `CfgLocationTypes` to common that should never be visible on the map (will still put it @ -10000, -10000)

### DIFF
--- a/addons/common/CfgLocationTypes.hpp
+++ b/addons/common/CfgLocationTypes.hpp
@@ -1,0 +1,16 @@
+//Create a location type that won't be drawn on the map
+//Ref: https://community.bistudio.com/wiki/Location
+
+class CfgLocationTypes {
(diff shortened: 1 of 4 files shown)


esteldunedain:
:+1: for merging this. I'd like to use the `ACE_HashLocation` on #3082

PR #3131, opened 2016-01-08
remove deprecated functions

commy2:


addons/common/functions/fnc_execRemoteFnc.sqf
```diff
@@ -20,6 +20,8 @@
  */
 #include "script_component.hpp"
 
+ACE_DEPRECATED("ace_common_fnc_execRemoteFnc","3.6.0","ace_common_fnc_globalEvent");
```
PabstMirror: `Search "execRemoteFnc" (59 hits in 44 files)`

3.6 might be too close?

commy2: Maybe. The goal to replace it with events was made a long time ago though.

jonpas: I'd say 3.6 is far enough, looking at previous releases it's quite some time between minor (eg. 3.4 -> 3.5) version updates. I think it's safe.

Glowbal: While I agree with you @commy2, we should go with 2 updates before a deprecated function is removed. Target 3.7.0 for removal of the `execRemoteFnc` function - we should get rid of all the usages within ACE in 3.5.0 however.

esteldunedain: `execRemoteFnc` is used in some recently deprecated functions like `fnc_setCaptivityStatus.sqf`, which we have to support until 3.6.0. Therefore, I think we should keep at least those references past 3.5.0 to avoid having to rewrite those.

Glowbal:
We will merge this when we start our release proces for 3.5.0.

Issue #2196, opened 2015-08-17
Distance in ACE_Action's sub-actions not respected


jonpas:
*This issue does NOT adhere to the issue posting guidelines, it is a framework issue!*

When you use `distance` parameter in a sub-action, it is not respected, it will always show (and even expand the parent if it's a sub-sub-action) until the interaction menu's maximum distance itself.

It works fine for `ACE_MainActions`, but for any child action distance code (eg. `_obj1 distance _unit < N`) must be added into the `condition` code to limit the distance.

NouberNou:
Do you have a use case for this? Because right now it seems like its able to be done in the current implementation and the original implementation was only for the main action.

jonpas:
Yes, it would be a lot easier when adding actions in config, rather than manually putting distances in the action's conditions.

PabstMirror:
I mentioned this in #3132, we could now easily add this in, but I don't think we want to.

 - createAction currently defaults to 2 meters, if we start enforcing distance on sub actions, old code will break or at least not work how people expected it to work
 - We only update the sub actions every 1 second (instead of 0.2 sec for the base level), so it's not as responsive when moving around.  Personally I though it felt very odd and not intuitive when you see an interaction point, but not all of it's actions,
 - Developers can still manually check distance in their own conditional code, which I think is what we should leave it at.

Can test by uncommenting: https://github.com/acemod/ACE3/blob/master/addons/interact_menu/functions/fnc_collectActiveActionTree.sqf#L38-L41

esteldunedain:
I agree. I'd leave change out.

PR #1294, opened 2015-05-18
Bandaid to prevent interact menu init from being interrupted

esteldunedain:
Bandaid for #1281, until we properly solve unreliable init handlers.

1- Menus for CAManBase are precomputed on preInit

2- Class action menus for all men are supposed to be equal to those of CAManBase

Note: we loose the capability of defining actions for specific classes from config (which we don't do on ACE annyway)

PR #3132, opened 2016-01-09
Cleanup Interaction Menu

PabstMirror:
Privates, params (~8% speedup)

I implemented but left disabled a fix for #2196 (Distance in ACE_Action's sub-actions not respected) (https://github.com/acemod/ACE3/commit/ed4cc42897fb4b00e8b31dd68f4f80b3168946d5#diff-078fc18f72c95b3cad3e8fdf7fca9d48R38)

It's a big change because most things have been written assuming that gets ignored. 
We can cleanup our stuff, but we risk breaking compatibility with missions/other addons by turing it on.

### DIFF
--- a/addons/interact_menu/XEH_clientInit.sqf
+++ b/addons/interact_menu/XEH_clientInit.sqf
@@ -79,8 +79,7 @@ GVAR(ParsedTextCached) = [];
 
 //Debug to help end users identify mods that break CBA's XEH
 [{
-    private ["_badClassnames"];
-    _badClassnames = [];
+    private _badClassnames = [];
     {
         //Only check Land objects (WeaponHolderSimulated show up in `vehicles` for some reason)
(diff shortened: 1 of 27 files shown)


esteldunedain:
:+1:

commy2:
>Privates, params (~8% speedup)
That is really nice!

PabstMirror:
I noticed that PR #1294 was having no effect anymore because `ace_interact_menu_fnc_addActionToClass` called compile with a classname before interaction_menu's compile with the object type happened.

This compiled a action for CaManBase and then copied it to all other man classes.

I fixed it, but now I'm not sure if that correct.  It apparently has been working for some time as is.  And it gives us the ability to add actions to different unit classes by config.

With https://github.com/CBATeam/CBA_A3/issues/42 fixed, maybe we should remove instead of fixing?

esteldunedain:
> With CBATeam/CBA_A3#42 fixed, maybe we should remove instead of fixing?

I think it should be removed. It was intended to reduce the compile time of actions on mission start to help prevent CBA from interrupting, which should no longer happen.

PR #3133, opened 2016-01-09
Use Function Caching in CBA

PabstMirror:
`SLX_XEH_COMPILE` doesn't cache anymore, it had been updated to a new version that takes new params.

Pretty trivial, but important to know about or the file patching won't do what you think it will. 
ACE devs need to use put this in a mission description.ext
```
class CfgSettings {
    class CBA {
        class Caching {
            compile = 0;
        };
    };
};
```
OR 
Add `#define DISABLE_COMPILE_CACHE` to the module's script component 
(or top of main/script_macros.hpp)
You NEED to use this method if you recompile functions mid-mission as the others use `compileFinal`

OR
Just add `cba_cache_disable.pbo` from optionals

PreInit goes from ~4.5 seconds to less than 0.3 (cuts mission SP Editor restart time in half)


commy2:
:+1: merge merge merge

PR #3135, opened 2016-01-10
Add Target Event to Events Framework

jonpas:
Somehow managed to miss that when rewriting that page and moving it to Frameworks.


PabstMirror:
Looks good, It's not really in the header, but the targets can be objects or numbers (from `owner`)

eg:`["displayTextStructured", [0], ["hi"]] call ace_common_fnc_targetEvent;;`

So Target could be Objects, Numbers, or a mix in an array.

jonpas:
Added, also good to know.

PR #3136, opened 2016-01-10
Fix make.py undefined variable and regex

jonpas:
- Fix possible undefined variable error when make.py exits before builds (`amountOfBuildsFailed` and `namesOfBuildsFailed` were getting initialized inside the if statement, but used outside)
- Rename `aceVersion` to `project_version` for consistency
- Update version check regex from CBA (could throw errors in edge-case PBO names)

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -697,11 +697,9 @@ def version_stamp_pboprefix(module,commitID):
         f.close()
 
         if configtext:
-            patchestext = re.search(r"version.*?=.*?$", configtext, re.DOTALL)
-            if patchestext:
+            if re.search(r"version=(.*?)$", configtext, re.DOTALL):
                 if configtext:
-                    patchestext = re.search(r"(version.*?=)(.*?)$", configtext, re.DOTALL).group(1)
-                    configtext = re.sub(r"version(.*?)=(.*?)$", "version = {}\n".format(commitID), configtext, flags=re.DOTALL)
+                    configtext = re.sub(r"version=(.*?)$", "version={}\n".format(commitID), configtext, flags=re.DOTALL)
                     f = open(configpath, "w")
                     f.write(configtext)
                     f.close()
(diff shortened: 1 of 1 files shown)


tools/make.py
```diff
@@ -697,11 +697,9 @@ def version_stamp_pboprefix(module,commitID):
         f.close()
 
         if configtext:
-            patchestext = re.search(r"version.*?=.*?$", configtext, re.DOTALL)
-            if patchestext:
+            if re.search(r"version=(.*?)$", configtext, re.DOTALL):
```
SilentSpike: Are versions expected to have a specific format? Because this seems like way too broad of a search to me.

jonpas: It works completely fine, has been used in CBA for some time now. The old way was throwing an error with some specific module name in CBA, I can look up which one exactly later.

SilentSpike: Well, based on my understanding of regex it's going to match from the first instance of the case sensitive string `version=` all the way to the end of the file

SilentSpike: That could be what you want though, I've never actually seen a version in one of the pboprefix files

jonpas: It will be to the end of the line due to `$`.

jonpas: It is most likely redundant now that CBA has one-liner pboprefix files as well.

SilentSpike: Not when you're not using the `re.M` flag

tools/make.py
```diff
@@ -697,11 +697,9 @@ def version_stamp_pboprefix(module,commitID):
...
                 if configtext:
```
SilentSpike: Unrelated to these changes, but this line seems totally redundant (as well as the respective `else` block)

tools/make.py
jonpas: Isn't it opposite? When you use the multiline flag dollar sign is used for end of string, without it end of line. I also verified it worked correctly, in old pboprefix files there were more lines after version and they were never picked up.

tools/make.py
SilentSpike: As does the one in the `else` block of the `re.search` condition

jonpas: This whole part is probably redundant, with centralised rvmake repository @glowbal started I think it's non-issue right now, can make an issue there.

SilentSpike: :+1:

tools/make.py
SilentSpike: Nah, multiline makes `^|$` match the beginning and end of every line and the overall string. Whole string only is default behaviour.

jonpas: You are right, my bad. Anyways, same comment as below and this is not used anyways. Thanks for correcting me!

Issue #3122, opened 2016-01-06
SwitchUnits conflicts with MenuPosition respawn template


JudahHarvey:
1. Run mission with respawnTemplate "menuposition" enabled for one side.
2. Create 1 unit for that side, and use ACE Switch unit to select this unit.

Upon death, you will notice the following
-Player take a ticket from that side, and will have to respawn at one of that side's respawn positions. If no positions have been setup, you will not be able to respawn and come back.

-If you manually add an extra life to the player AFTER he switches to the new Unit, this issue will resolve itself after the first death. However, you must also create a respawn position for the player to spawn on for this to work. You will notice that you can die as many times as you want after taking control of units, but you will not be prompted with the respawn menu. 

It would be better to ensure that units that die this way do not take tickets, will not send players to the respawn menu the first time.

Description.EXT
respawnTemplates[] = {"Tickets", "MenuPosition"};
respawnTemplatesWest[] = {"Tickets", "MenuPosition"};
respawnTemplatesEast[] = {"Tickets", "MenuPosition"};
respawnTemplatesGuer[] = {"Tickets", "MenuPosition"};

Using after I switch units using debug console
[player, 5] call BIS_FNC_respawntickets;
spectatePosition = [player, [0,0,150]] call BIS_fnc_addRespawnPosition;

As previously stated. Without taking these steps, players who use switch unit will be pushed into the menuposition screen without a respawn or tickets if they die in their switched unit.

Glowbal:
Pinging our respawn template wizard @SilentSpike

SilentSpike:
I'm honestly not sure there's anything we can do about this without hacking up the respawn template (which is bad for compatibility) or with an event handler of some kind that fires *before* death.

JudahHarvey:
Can't we just ammend the switch units function to add a personal lives to the switched unit as well as a respawn point? Once the unit respawns there, it's all fixed and this never needs to happen again.

SilentSpike:
Hmm, that's not a bad suggestion. Although it would mean altering the scenario without the designer's knowledge. That AI unit switched to would respawn, that side would now have a respawn marker (probably wouldn't matter in most cases, but they might intentionally not have added one for whatever reason).

bux578:
I could try removing the switch back from the `Respawn` event to the `Killed` event.
https://github.com/acemod/ACE3/blob/master/addons/switchunits/functions/fnc_switchUnit.sqf#L67

SilentSpike:
[...]

PR #3142, opened 2016-01-10
Cleanup SwitchUnits, Possible fix for #3122

bux578:
When merged this PR will:

* cleanup the SwitchUnits module
* change the `Respawn` event for the `Killed` event
* keep the killed unit and won't delete it
* show text when trying to switch (waiting for unit to become local)


bux578:
It would be better to use `setGroupOwner` instead of `setOwner`. But that would mean a different handling of the units the player is switching to.

PR #3146, opened 2016-01-11
Function to detonate explosives via script

VKing6:
This function is designed to wrap `ace_explosives_fnc_detonateExplosive` for use in triggers or scripts, to detonate explosives placed down in the editor.


bux578:
:+1:

bux578:
@VKing6 Could you add that to the wiki?

VKing6:
@bux578 Yea that's a good idea. Where, though?

Issue #2677, opened 2015-10-03
ace_spectator: Unit list: virtual player appears + duplicate squads


voiperr:
ACE3 Version: 3.3.2 Stable

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* ace_spectator, with `GVAR(filterUnits) = 3`

**Description:**
* On starting spectator, the player's virtual character briefly shows up in the unit list. The player can be selected and followed (usually invisible, out in the ocean).
* After a few seconds, it disappears, and odd behaviour can happen, usually with one of the real speccable squads being duplicated.

**Steps to reproduce:**
* Send the player into spectator mode through
```
ace_spectator_fnc_setSpectator;
[player, true] call ace_spectator_fnc_stageSpectator
```

(do not call it through the debug console; it will crash the game)
* Repro mission: https://www.dropbox.com/s/f70zd9md0vmc1js/ace_spectator_bugs.Stratis.pbo?dl=0

**Where did the issue occur?**
Everywhere

alganthe:
Can confirm I experience this issue from time to time.

SilentSpike:
:+1: Good report, will investigate

SilentSpike:
Not entirely sure what's causing this one. Looked through the code a few times and it checks out as far as I can see.

SilentSpike:
Just to update, can definitely reproduce this, still not sure what is causing it.

Cyruz143:
Just to clarify, is this why if a unit dies, you still have 4 people listed in a fireteam but it just repeats one name, for example:

```
Pre-death:
    A1 ->
        John
        Thomas
        Oliver
        Dave

Post Oliver's death:
    A1 ->
        John
        Thomas
        Dave
        Dave
```

Hit escape to force reload spec UI seems to clear it.

SilentSpike:
It's most likely related, in fact, your post is a little more specific with regards to reproduction and might help me to track down the cause

esteldunedain:
Fixed in #3147

SilentSpike:
Only partially fixed! :smile:

SilentSpike:
Since I'm quite busy at the moment and don't have much development time I'll write down my current thoughts on this issue in case anybody else wants to take a look:

- Remaining issue is that there are seemingly arbitrary duplicate units/groups in the tree
[...]

PR #3147, opened 2016-01-12
Restart specatator display if player in unitList

PabstMirror:
Possible fix for #2677

I think the core problem is that voiperr's demo mission does setSpectator before calling stage.
SetSpectator updates the unitList that has the player in it.

This PR is really just a bandAid that restarts the display when they are called out of order.


SilentSpike:
Good catch, I'm thinking about reducing the delay between list updates so might remove the display restart in a later iteration.

voiperr:
If there is a specific order to invoking those commands to work safely, the documentation should reflect that. Or, have a third function that executes the 2 functions properly from one call.

SilentSpike:
I mean...the whole point of this PR is that there isn't a specific order :stuck_out_tongue:

Glowbal:
Merge?

PR #3148, opened 2016-01-12
Nerf the amount of backblast received by the firer via reflection

### LINKED ISSUE #2890: Damage from a jet of the RPG-32 are not realistic and overly exaggerated.
ACE3 Version: 3.4.1
Stable 1.54

**Mods:**
* `@CBA_A3`
* `@ace`
* `@DragonFyre`

**Placed ACE3 Modules:**
Advanced and Basic Medical*

**Description:**
Damage from a jet of the RPG-32 are not realistic and overly exaggerated. In ACE3, it is impossible to use this weapon in the forest, in the cities, inside houses and on the hills sides. In these locations, almost every shot serious damages the player or causes him to lose consciousness. Why the hell the soldier should to lose consciousness, even if he has received burns from a jet? The soldier not thrusts his head or face into the jet, and therefore should not receive serious damages, especially to lose consciousness. At wrong using (if behind the wall or object) the soldier should has only receive light damages, nothing more. Please fix this details/

**Steps to reproduce:**
Use RPG-32

Editor

**RPT log file:**
Not needed

esteldunedain:
When calculating damage due to backblast reflection, we used the distance between the launcher and the obstacle behind it. However, we failed to acount the fact that backblast has to travel twice that distance before hitting the firer, hence spreading twice as much. Hence the exaggeration of reflection damage compared to direct damage to nearby soldiers.

Fix #2890

### DIFF
--- a/addons/overpressure/functions/fnc_fireLauncherBackblast.sqf
+++ b/addons/overpressure/functions/fnc_fireLauncherBackblast.sqf
@@ -48,7 +48,7 @@ _affected = (ASLtoAGL _position) nearEntities ["CAManBase", _backblastRange];
 
 // Damage to the firer
 private "_distance";
-_distance = [_position, _direction, _backblastRange, _firer] call FUNC(getDistance);
+_distance = 2 * ([_position, _direction, _backblastRange, _firer] call FUNC(getDistance));
 
 TRACE_1("Distance",_distance);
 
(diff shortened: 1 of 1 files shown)


commy2:
My mind is blown.

jokoho48:
Commys head :boom:

:+1:

Glowbal:
:boom:

captainblaffer:
Thanks for fixing this, but should this nerf not theoretically be even  more than *2 the distance?
I'd imagine much of the backblast power is absorbed by the wall or spread out rather than directly sent back to the launcher. 

Also, does the calculation keep this in mind? http://imgur.com/PGRpRlD

esteldunedain:
> hould this nerf not theoretically be even more than *2 the distance?

Maybe; at the very least is 2x, so it's a step in the right direction.

> Also, does the calculation keep this in mind? when shooting with an up-angle?

That thing is already considered on the code.

captainblaffer:
Awesome, thanks for your reply

Issue #1098, opened 2015-05-10
Save game time compatibility


jaynus:
This is a future work item for saved game compatability concerning time:

esteldunedain [8:07 AM] 
it's basically a if _newACE_time < ACE_time then _newACE_time = ACE_time

esteldunedain:
To support savegames we need to additionally ensure PFH with time intervals survive savegames on CBA.

jaynus:
#1099 and #1098  for future work items based on this after this is merged in for 3.1

WarezzK:
Anything progress this issue? :) Just a little info please. Thx!

commy2:
I'm hijacking this.
I think I now understand enough about how save games (and mission restarts...) work to finish this.

Requires some changes in CBA. For anyone waiting on this: my personal goal is to finish this with the release of the Expansion / Tanoa (, because the new editor changes some details).

ref: https://github.com/acemod/ACE3/pull/3150

esteldunedain:
:+1:

JoeSchr:
Is it possible to make ACE3 save game compatible by removing some lesser needed pbo's? If yes, which want are most likely responsible for breaking save games? ty

commy2:
No. It's a core issue.

commy2:
Fixed with #3516

Issue #1099, opened 2015-05-10
Convert all time/diag_tickTime instances to ACE_time or ACE_diagTime for accuracy


jaynus:
This will need to be done by hand by @esteldunedain and myself.

PR #1092, opened 2015-05-10
ACE_time variables for accurate purposeful time use

jaynus:
There was discussion recently about the fact we use 'time' everywhere in our PFH's so they detect game pausing and support saved games.

However, time is unreliable on slow machines (< 20fps) - so we *should* be using diag_tickTime for PFH functionality.

This provides the following global variables for a hybrid approach of this:
ACE_time = diag_tickTime accuracy, except it does not advance when the game is paused. It will also obviously save and continue forward in a saved game, so supports them as well.
ACE_realTime = diag_tickTime
ACE_gameTime = time
ACE_virtualTime = diag_tickTime accuracy, except taking accTime into account.
ACE_virtualGameTime = time, except takes accTime into account.
ACE_virtualPausedTime = the amount of diag_tickTime time that was paused, except accTime is taken into account. WHO THE FUCK WOULD NEED THIS!? I'm sure someone will.
ACE_pausedTime = the amount of time the game has been paused in total. 

ACE_paused = true/false, whether the game is currently paused.

These would have a seperate PR, which would *replace all uses of time with ACE_time!!!!!!!*

Completes #1091

PR #3150, opened 2016-01-12
reset interaction menu time variables on save game load

commy2:
ref: #1098

I have to set them to 0, because `ACE_diagTime` is still the value from the previous session and `execNextFrame` fails, because it relies on `GVAR(nextFrameNo)` (from common) which isn't refreshed either.

### DIFF
--- a/addons/interact_menu/XEH_preInit.sqf
+++ b/addons/interact_menu/XEH_preInit.sqf
@@ -67,6 +67,13 @@ GVAR(expanded) = false;
 
 GVAR(startHoverTime) = ACE_diagTime;
 GVAR(expandedTime) = ACE_diagTime;
+
+// reset on mission load
+addMissionEventHandler ["Loaded", {
+    GVAR(startHoverTime) = 0;
+    GVAR(expandedTime) = 0;
+}];
+
 GVAR(iconCtrls) = [];
 GVAR(iconCount) = 0;
 


esteldunedain:
Hold on a sec...

Bare with me, I know #1098 and #1099 are a mess, but let me try to explain how we intended to handle this problem:

We intended the `ACE_diagTime` variable to be equal to `ACE_diagTime` only on the *first session*. When loading a savegame we planned to do something like this:

```sqf
// On init, always
ACE_diagTimeOffset = 0;
```

```sqf
// On fnc_timePFH.sqf
if (ACE_diagTime > diag_tickTime) then {
    // A saved game was loaded
    ACE_diagTimeOffset = ACE_diagTime - diag_tickTime;
};
// ACE_diagTime continues to accumulate from the last session
ACE_diagTime = diag_tickTime + ACE_diagTimeOffset ;
```

That procedure was supposed to ensure that everything related to time would still accumulate properly between sessions, without having to patch each system independently. The only additional task would be this: https://github.com/acemod/ACE3/issues/1098#issuecomment-100653511

commy2:
PFH's do survive save games in the latest CBA dev version: https://github.com/CBATeam/CBA_A3/commit/2deccce54ebd5339dde12afb42d633118f7eb111

But even with that commit, the interaction menu wouldn't open sub menues.
Any idea why that is? Resetting these two values does fix it.

esteldunedain:
> Any idea why that is? 

The code I wrote above was never actually commited to `fnc_timePFH.sqf`

commy2:
Is that all? Looks easy enough.

`if (ACE_diagTime > diag_tickTime) then {`
That assumes that the mission was loaded right after the game was started. It could be the other way around though. A mission quickly saved and then later loaded while the game was running for an hour in the main menu.

The variables should've probably be named `ACE_tickTime` instead `ACE_diagTime`. diag is for diagnostic

esteldunedain:
> That assumes that the mission was loaded right after the game was started.

Totally right on that one. That complicates things slightly. `addMissionEventHandler ["Loaded",` could be used instead.

> The variables should've probably be named

**messy, incomplete and confusing piece of shit**

esteldunedain:
Here is a list of time variable names and their confusing meanings: https://github.com/acemod/ACE3/pull/1092

commy2:
I don't see `ACE_diagTime` in that list, even though it's used everywhere in ACE.

esteldunedain:
> I don't see ACE_diagTime in that list, even though it's used everywhere in ACE.

Once again: **messy and confusing piece of shit**

commy2:
You are making it not easy for me. Haha.

jonpas:
It was initially `ACE_tickTime` https://github.com/acemod/ACE3/commit/318b2437f9db0513ec9a42bc202d5f6f48eebc2a

Then in same PR renamed to `ACE_diagTime` https://github.com/acemod/ACE3/commit/e1c3c40cddac38d0db0565e74d83ba2c4fbde7e6

I guess it was made to fit `diag_tickTime` command.

commy2:
Both fit, but the current name makes less sense.

jonpas:
And `ACE_tickTime` is a macro apparently https://github.com/acemod/ACE3/commit/c39b5931c705e6f73ac0a129ee74c65472efb252

commy2:
Oh good, where to start?

esteldunedain:
Check the e1c3c40 commit message to see why :wink:

commy2:
Amazing.

commy2:
Tbh. All this sounds like a case for CBA.

esteldunedain:
> Tbh. All this sounds like a case for CBA.

Agreed. In that case, with proper variable names please.

commy2:
We could use this as a band aid. There is one more minor issue in the CBA keyhandling (it takes 10 seconds for them to be restored after a save game), but with this + CBA_dev save games should at least be playable.

esteldunedain:
> We could use this as a band aid.

Yeah sure. I just wanted to point out that a better solution was conceived already.

commy2:
@esteldunedain 
Would this be sufficient?
`CBA_missionTime`

pros:
- based on tickTime
- set every frame
- synched in MP/JIP
- resumes correct value when loading save games
- multiplied by accTime

cons:
- -1 on clients during preInit
- not linked to ingame day `time` (faster)
- ignores skipTime, setDate, timeMultiplier etc. (ingame day time)

Initial test:
https://gist.github.com/commy2/c01e4885b8961363ef7e
https://gist.github.com/commy2/06de40913799965ba8e5

commit: https://github.com/CBATeam/CBA_A3/commit/e69d25ec9baeae879c9ca9d312c66a48bd91b1ee

esteldunedain:
That's very nice!

I think it will work wonderfully for most things, Certainly so for the interact menu. My only (minor) dislike is that it gets multiplied by acctime, which is not ideal for ui effect things like the IM; but I guess it's better to have one or two of this variables instead of the huge number described in #1092.

commy2:
>I guess it's better to have one or two of this variables instead of the huge number described in

That's exactly what I thought. I want to keep the overhead at a minimum. I added the accTime thing, because everything else runs faster (including `uiSleep`).

Since it never comes into play in MP anyway, I could probably replace the function depending on `isMuliplayer`.

esteldunedain:
Wonderfull.

If we are only keeping one variable, it may be best to just call it `CBA_time`.

I'm looking at ACE code, and I think it would be kind of safe to batch replace `ACE_time` by this. Maybe the old system should be left for a little while longer, cause the other variables may be used in missile guidance code.

commy2:
I chose `missionTime`, because it nicely describes what it does: It returns how long the mission is running, regardless of JIP, save games, which client etc.

esteldunedain:
Fair enough

esteldunedain:
Backlogging this in favor of a proper fix on CBA coming soon (TM).

PR #3151, opened 2016-01-13
Various explosives enhancements

VKing6:
If I had known I'd be making this many changes to explosives, I would have made a single branch for it, I swear.

This PR currently does these things:
* Adds the function `fnc_connectExplosive` to add an editor placed explosive to a player's detonator.
* Adds editor placeable variants of SLAM (bottom attack) and IED (pressure plate)
* Reduces small IED pressure plate trigger distance
* Renames all of the custom classes to use the ACE_ prefix

Plus a bunch of other random stuff.

To possibly look at in the future:
* Eden editor eventhandlers for connecting to players.


VKing6:
Current problems is that the zeus modules plain don't work. No idea why.

addons/explosives/CfgVehicles.hpp
```diff
@@ -338,4 +338,80 @@ class CfgVehicles {
...
+    // Editor placed mines
+    class MineBase;
+    class SLAMDirectionalMine: MineBase {
+        displayName = CSTRING(Module_SLAMSideAttack_DisplayName);
+    };
+    class ACE_SLAMBottomMine: SLAMDirectionalMine {
+        author = ECSTRING(common,aceteam);
+        ammo = "ACE_SLAMDirectionalMine_Magnetic_Ammo";
+        displayName = CSTRING(Module_SLAMBottomAttack_DisplayName);
+        // TODO: Find a way to place the mine laying down instead of standing up
```
CorruptedHeart: This would need to be scripted, does extended eventhandlers work on this?

VKing6: It doesn't appear to, nor does the regular eventhandlers work as far as I can tell.

SilentSpike:
>Renames all of the custom classes to use the ACE_ prefix

If you're renaming classes it might be worth switching to using `GVAR` while it's still viable to change classnames without causing issues

VKing6:
Do you mean for anything specific?
I didn't notice anything that wasn't inside a class ACE_ something that didn't use GVAR.

SilentSpike:
I meant like [this](https://github.com/acemod/ACE3/blob/bcfac5db0b7b1dee270d9ce4273de07a0689be0e/addons/mk6mortar/CfgVehicles.hpp#L48).

I don't actually think it's a standard now that I've looked at some other stuff, but it's kind of nice since it makes the classnames self documenting

esteldunedain:
> I don't actually think it's a standard now that I've looked at some other stuff, but it's kind of nice since it makes the classnames self documenting

To be honest I prefer `ACE_SLAMDirectionalMine_Timer_Ammo` to `ACE_Explosives_SLAMDirectionalMine_Timer_Ammo`. Adding the module name to variables is meant mainly to avoid namespace collision, which is unlikely on class names.

VKing6:
I don't mind changing the module name to GVAR(Module), but will that have any consequences for missions where the module exist?

jonpas:
Yes it would be backwards incompatible.

SilentSpike:
Nah, dom't worry about the naming convention, I was under the impression that was our standard.

Any classname that is renamed is backwards incompatible though.

VKing6:
> I didn't notice anything that wasn't inside a `class ACE_` subclass that didn't use GVAR.

This was a dirty, dirty lie, and I've fixed all of them now.

Also added `fnc_connectExplosive`.

addons/explosives/CfgAmmo.hpp
```diff
@@ -29,12 +29,15 @@ class CfgAmmo {
...
-        ACE_Explosive = "ClaymoreDirectionalMine_Remote_Ammo_Scripted";
+        GVAR(magazine) = "ClaymoreDirectionalMine_Remote_Mag";
+        GVAR(Explosive) = "ClaymoreDirectionalMine_Remote_Ammo_Scripted";
         GVAR(defuseObjectPosition[]) = {0, 0, 0.038};
         soundActivation[] = {"", 0, 0, 0};
         soundDeactivation[] = {"", 0, 0, 0};
     };
-    //class ClaymoreDirectionalMine_Remote_Ammo_Scripted: ClaymoreDirectionalMine_Remote_Ammo;
+    class ClaymoreDirectionalMine_Remote_Ammo_Scripted: ClaymoreDirectionalMine_Remote_Ammo {
+        GVAR(magazine) = "ClaymoreDirectionalMine_Remote_Mag";
```
CorruptedHeart: Is this necessary, won't it be inherited from the parent?

VKing6: Ah, yes, you're right. I added that because I thought there was something that didn't work properly. Turns out it was the wrong config in getConfig ... -.-

VKing6:
Confirmed that this also works for AI with the dead man's switch. Evil mission makers take note.

addons/explosives/CfgVehicles.hpp
VKing6: Looks like adding a check to xeh_postInit works.
https://github.com/acemod/ACE3/pull/3151/files#diff-bdef6c2f5ffe466a001dbd949d164a83R39

PR #3154, opened 2016-01-13
Add spectator text chat

SilentSpike:
Turns out custom radio channels are kind of a pain to work with. However this should work fine assuming the `radioChannelAdd` commands will work on client machines.


SilentSpike:
I actually forgot that I had this branch ready, so I'm going to go ahead and merge it so I can move onto my other PR.

PR #3155, opened 2016-01-13
Fix some spectator errors

SilentSpike:
<ul>
<li>Storing the display in a <code>uiNamespace</code> variable allows me to explicitly close the right display and hopefully fix the interface staying open upon mission end.<br/><br/>
@Commy2's theory is that perhaps when the main display <code>46</code> is killed upon mission end, <code>findDisplay</code> can't find the spectator display to close it because the game thinks it doesn't exist.</li>
<br/>
<li>That was also a minor oversight introduced in the fix for #2989, the icon handler is a mission event handler that isn't self terminating so the index needs to be preserved until after it is removed via <code>removeMissionEventHandler</code>. It is safe to let the <code>onUnload</code> event take care of that (and preferable so that icons disappear while temporarily closed).</li>
</ul>

### DIFF
--- a/addons/spectator/XEH_postInit.sqf
+++ b/addons/spectator/XEH_postInit.sqf
@@ -10,4 +10,4 @@
 }] call EFUNC(common,addEventHandler);
 
 // Should prevent unending spectator on mission end
-addMissionEventHandler ["Ended",{ [false] call FUNC(setSpectator) }];
+addMissionEventHandler ["Ended",{ [QGVAR(EndMission)] call FUNC(interrupt) }];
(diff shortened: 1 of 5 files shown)


Cuel:
addMissionEventHandler doesn't run when pressing esc > abort, `missions`, + guessing when the server stops responding, from some earlier test. Seems to only work for `endMission`. Would that be an issue?

SilentSpike:
As much of an issue as it already is (although thank you for clarifying under what circumstances that happens). The display should automatically be terminated by the game, but for some unknown reason it isn't.

Cyruz143:
@SilentSpike just gave this a test, still has issues with mission end, users need to hit Escape to see the end mission screen, doesn't actually seem to matter if they're in spec or still alive at the point of it being called but like all good bugs, it's intermittent...

Tried it with ```endMission "end1"```` in dedicated server.

PR #3163, opened 2016-01-14
Unified status effect handler

### LINKED ISSUE #64: Unify the implementation of setForceWalk, setCapticityStatus and setDisableInputStatus
The first two come from agm, the last from cse. Unify the style on the three, or, if possible, use them as front end to a single generic function.

### LINKED ISSUE #2757: add setAllowDamageStatus, getAllowDamageStatus functions

PabstMirror:
Close #64
Close #2757 (which has been blocking progress on #2230)
Replace PR #2758 

Allows multiple sources to add and remove `forceWalk`, `setCaptive` and `allowDamage` (the event is called `blockDamage` because here any true should turn the result false).

 - Effect events are reapplied on locality switch
 - All effects are reset on respawn.
 - Only modify effects that we've actually set at some point (should maximize compatibility with missions that do not use this framework)


esteldunedain:
:+1:

MikeMatrix:
:+1:

Glowbal:
You remove some methods here. Could you replace them with a wrapper around the new method if possible and print a warning message about deprecated?

BaerMitUmlaut:
What about [allowSprint](https://community.bistudio.com/wiki/allowSprint)? This would be useful for a custom fatigue system.

commy2:
^good idea

PabstMirror:
@BaerMitUmlaut - added as `blockSprint` with about 5 lines (have to reverse some of these names because you want any true to make the result true)

@Glowbal They were listed as public so good catch, should also be useful for debugging to have a getter.

Glowbal:
They are still missing the deprecated warning log print.

esteldunedain:
> They are still missing the deprecated warning log print.

Take a look now

jokoho48:
@esteldunedain i think you fuck something up

esteldunedain:
> @esteldunedain i think you fuck something up

As far as I can tell it's github acting up. The extra changes it shows are already merged into master, so they shouldn't be shown in the diff.

jokoho48:
no that not look right sorry but i think there is go something wrong

PabstMirror:
I've seen this before
https://github.com/acemod/ACE3/branches  says it's 5 commits ahead of master
I could probably close and make pr for same branch and it would be fine.

esteldunedain:
> I could probably close and make pr for same branch and it would be fine.

Or you could just merge it :wink:

esteldunedain:
[This instructions](http://makandracards.com/makandra/13043-force-github-pull-requests-to-update-the-diff-against-its-target-branch) should solve the outdated diff problem, but I don't think they are worth it.

PR #3164, opened 2016-01-14
Move littler syncedEH install to after settingInit

PabstMirror:
Need to call this after settingInitalized for `GVAR(litterCleanUpDelay)` to be set correctly from module settings.  

From closed PR #2758 (not sure why I fixed it there)


esteldunedain:
:+1:

PR #3172, opened 2016-01-16
Overpressure cleanup and fix distance check

### LINKED ISSUE #3156: FFV backblast seems too much
ACE3 Version: 3.4.2
Arma 3 Version: 1.54

**Mods:**
* `@CBA_A3`
* `@ace` 

**Placed ACE3 Modules:**
None

**Description:**
Backblast seems too much when firing a launcher FFV turned out position. Particularly in the offroad.

Video: https://www.youtube.com/watch?v=1IJt2bTD5xo&feature=youtu.be

From a realism point of view I spoke to a military buddy of mine and whilst he advices against firing from a vehicle he said it is unlikely in the video below that the firer would be wounded from the backblast. He also mentioned that if the launcher has fins (i.e. AT4) they might expand and collide with the cabin though that is a different issue.

**Steps to reproduce:**
Get an RPG-42 and get into the back seat of an offroad, turnout, select the RPG and fire.

PabstMirror:
- Cleanup, headers, modern privates
 - Fix #3156 (`lineIntersectsSurfaces` did not filter ffv shooter)
 - Added both a3 tank weapons, I am unsure why there were not already added.

### DIFF
--- a/addons/overpressure/CfgEventHandlers.hpp
+++ b/addons/overpressure/CfgEventHandlers.hpp
@@ -14,7 +14,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_FiredBIS_EventHandlers {
     class CAManBase {
         class ADDON {
-            clientFiredBIS = QUOTE(_this call FUNC(firedEHBB););
+            clientFiredBIS = QUOTE(if (local (_this select 0)) then {_this call FUNC(firedEHBB);};);
         };
     };
 
--- a/addons/overpressure/CfgEventHandlers.hpp
+++ b/addons/overpressure/CfgEventHandlers.hpp
@@ -14,7 +14,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_FiredBIS_EventHandlers {
     class CAManBase {
         class ADDON {
-            firedBIS = QUOTE(if (local (_this select 0)) then {_this call FUNC(firedEHBB);};);
+            clientFiredBIS = QUOTE(_this call FUNC(firedEHBB););
         };
     };
 
--- a/addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
+++ b/addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
@@ -51,7 +51,7 @@ private _return = [
 ];
 
 private _varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
-missionNameSpace setVariable [format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine], _return];
+missionNameSpace setVariable [_varName, _return];
 TRACE_2("Return",_varName,_return);
 
 _return
--- a/addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
+++ b/addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
@@ -49,9 +49,9 @@ private _return = [
     (getNumber (_config >> QGVAR(range))),
     (getNumber (_config >> QGVAR(damage)))
 ];
-TRACE_2("Return",_varName,_return);
 
 private _varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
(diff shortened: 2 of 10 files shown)


addons/overpressure/CfgEventHandlers.hpp
```diff
@@ -14,7 +14,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_FiredBIS_EventHandlers {
     class CAManBase {
         class ADDON {
-            firedBIS = QUOTE(if (local (_this select 0)) then {_this call FUNC(firedEHBB);};);
+            clientFiredBIS = QUOTE(_this call FUNC(firedEHBB););
```
jokoho48: ? not sure but if i remeber right is clentFiredBIS only disalbed on server and not local

addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
```diff
@@ -9,68 +9,49 @@
...
+private _return = [
     (getNumber (_config >> QGVAR(angle))),
     (getNumber (_config >> QGVAR(range))),
     (getNumber (_config >> QGVAR(damage)))
 ];
-TRACE_1("Return",_return);
-_varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
+TRACE_2("Return",_varName,_return);
+
+private _varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
```
jokoho48: this make 0 sense

addons/overpressure/CfgEventHandlers.hpp
PabstMirror: I switched the `if !([_firer] call EFUNC(common,isPlayer)) exitWith {};` in fireLauncherBackblast
to `_firer != ACE_player` in firedEHBB  
(because local + isPlayer could only ever be `ace_player`... I think?)

In testing re-adding the local check in the EH is slightly faster than always doing the params and `_firer != ACE_player` check when non players are shooting, so I will leave as it was.

addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
```diff
@@ -9,68 +9,49 @@
...
+private _return = [
     (getNumber (_config >> QGVAR(angle))),
     (getNumber (_config >> QGVAR(range))),
     (getNumber (_config >> QGVAR(damage)))
 ];
-TRACE_1("Return",_return);
-_varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
+
+private _varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
 missionNameSpace setVariable [format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine], _return];
```
jokoho48: if you want to keep the var _varName than replace it in this line

esteldunedain:
:+1:

jokoho48:
:+1:

addons/overpressure/CfgEventHandlers.hpp
```diff
@@ -14,7 +14,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_FiredBIS_EventHandlers {
     class CAManBase {
         class ADDON {
-            firedBIS = QUOTE(if (local (_this select 0)) then {_this call FUNC(firedEHBB);};);
+            clientFiredBIS = QUOTE(if (local (_this select 0)) then {_this call FUNC(firedEHBB);};);
```
commy2: That would mean that there no longer is backblast on dedicated server AIs?

PabstMirror: We already only ran on local and isPlayer which I just switched to == ace_player.  I think the idea was that AI have no idea how backblast works, so they would not know how to use it.

The firer will still trigger the targetEvent to everyone, so player shots will still effect AI. So skipping dedi on the event is just a very small performance savings.

esteldunedain: :+1:

Issue #3171, opened 2016-01-16
PSO-1 from RHS with compat_rhs_afrf3


ZikiCZ:
Vertical axis zeroing does not seem to work for the PSO-1 sight.

Try change this:

class rhs_acc_sniper_base;
    class rhs_acc_pso1m2: rhs_acc_sniper_base {
        ACE_ScopeAdjust_Vertical[] = { 0, 0 };
        ACE_ScopeAdjust_Horizontal[] = { -10, 10 };
        ACE_ScopeAdjust_VerticalIncrement = 0.0;
        ACE_ScopeAdjust_HorizontalIncrement = 0.5;
    };

 on this: (like a HLC_AK)

class rhs_acc_sniper_base;
    class rhs_acc_pso1m2: rhs_acc_sniper_base {
        ACE_ScopeAdjust_Vertical[] = { 0, 0 };
        ACE_ScopeAdjust_Horizontal[] = { -10, 10 };
        ACE_ScopeAdjust_HorizontalIncrement = 0.5;
    };

VKing6:
AFAIK the PSO-1 doesn't have a vertical adjustment beyond zeroing.

esteldunedain:
> AFAIK the PSO-1 doesn't have a vertical adjustment beyond zeroing.

Exactly.

bux578:
Though, thank you for your report!

ZikiCZ:
Rlly? Im not sure. It was in ACE2 and now in HLC's PSO-1 from hlc_ak_pack (with ACE3). 
Look here:
(Part: 5.1 Elevation turret)
http://www.ar15.com/content/webPDF/pos1.pdf

Part: Bullet Drop Compensation elevation turret
https://en.wikipedia.org/wiki/PSO-1

https://docs.google.com/document/d/1mdaHQDCFsuiw5Vq9OWucfkvyuLN6lzX-fvINqZEhTKw/edit?pref=2&pli=1#heading=h.prgm47spffui

(We talking about PSO-1M2. And pls add PSO-1M2-1 to the compat_rhs_afrf3.)

VKing6:
I'm convinced.

VKing6:
Okay, after investigating, the compat pbo takes this into account. The vanilla 100-1000m adjusting is still enabled. There's no satisfying way at this time to implement the PSO-style adjustment fully using the ACE scope adjust system.

For the time being, this will probably not change, but you are correct that the M2-1 is missing windage adjustment.

[...]

PR #3178, opened 2016-01-18
Scopes: Don't show adjustment or play click sound if the adjustment doesn't happen

VKing6:
Fixing an issue discussed in #3171, where, even if the scope is not set to allow adjustment in a particular axis, the click sound will still play and the adjustment turret will still be displayed.


esteldunedain:
:+1:

PR #3179, opened 2016-01-18
RHS Compat: Add PSO-1M2-1 windage adjust

VKing6:
Add missing windage adjustment for one of the PSO-1 variants.
Ref #3171


esteldunedain:
:+1:

PR #3187, opened 2016-01-19
Add missing FUNC to FCS key (Rangefinders)

### LINKED ISSUE #3160: Binoculars and Optics w/ range measuring functions DO NOT WORK anymore
Binoculars and Optics w/ range measuring functions DO NOT WORK anymore that reads distance and range with 'TAB' key/

I'm guessing it stopped working since the last update(3.4.2.0) since I noticed it just today and I play nearly everyday.

Rangefinders and optics such as Nightstalker and TWS that has distance measuring functions.

I loaded the game from the launcher(STEAM) wo/ACE3 and it works just fine, but when I do load it w/ACE3, doesn't work anymore.

ADDED:
Rangefinders that doesn't work are only the ones with 'TAB' key distance measuring ones wo/laser.
The ones w/LASER measuring w/ 'left mouse button' works fine.

IT IS ONLY THE 'TAB' KEY DISTANCE MEASURING FUNCTION NOT WORKING.

Conclusion:
Whatever Optic or Binocular that measures distance w/ 'TAB' key doesn't function at all.
It is extremely stressful when you're trying to take out multiple enemies on the move, trying to measure distance changing to Vector(which seems to work fine when measuring distance since it measures it with 'R' key(and in this case, the 'TAB' key which measures elevation of the target also seems to work fine).

P.S. ALL KEY BINDING FOR BOTH ARMA3 AND ACE ARE AT ITS DEFAULT AND HAVEN'T MESSED IT WITH ANYTHING ELSE.

PabstMirror:
Fix #3160

This should match behaviour in pre-3.4.2

### DIFF
--- a/addons/fcs/functions/fnc_keyDown.sqf
+++ b/addons/fcs/functions/fnc_keyDown.sqf
@@ -16,7 +16,7 @@
 params ["_vehicle", "_turret"];
 
 private _distance = call FUNC(getRange);
-call (updateRangeHUD);
+call FUNC(updateRangeHUD);
 
 if !(!GVAR(enabled) && FUNC(canUseFCS)) exitWith {};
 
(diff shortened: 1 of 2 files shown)


Glowbal:
Nice one :+1:

commy2:
:+1: Thanks
Don't even know how these got lost.

PR #3191, opened 2016-01-19
Cargo Cleanup, Add makeLoadable

### LINKED ISSUE #2712: Cargo >> getSizeItem always return -1 if no config for cargo is present
getSizeItem will alway return -1 if object hasn't a proper config in cfgVehicle.
https://github.com/acemod/ACE3/blob/v3.3.2/addons/cargo/functions/fnc_getSizeItem.sqf

correct me if i'm wrong, but if like a object to be loadable with:
_obj setVariable ["ace_cargo_size", 1]; 
_obj setVariable ["ace_cargo_canLoad", 1];
like a Hedgehog i can't do this.

PabstMirror:
Close #2712
Adds a function to make any object loadable,
`[campChair, true, 1] call ace_cargo_fnc_makeLoadable`
Might make this callable by a module

most lines are minor cleanup, the meat is in initObject.sqf and makeLoadable.sqf


jonpas:
> Might make this callable by a module

Wouldn't be bad if it would be a callable event either.

PabstMirror:
yeah,I was thinking about that, would be much easier for missionmakers to use.

bux578:
-> http://ace3mod.com/wiki/missionmaker/useful-functions.html

PabstMirror:
jonpas, on second thought, I'm not so sure about the event, as this needs to be run on all machines, including jips.  And we don't really have a great way to do that with our event system.

I think the module and the raw function should be enough.

Glowbal:
We have JIP persistent events.

PR #3192, opened 2016-01-19
Rewrite ace_common_fnc_getTargetDistance to use LineIntersectsSurfaces.

VKing6:
Turns out it wasn't really needed after all, but better is better :p

### DIFF
--- a/addons/common/functions/fnc_getTargetDistance.sqf
+++ b/addons/common/functions/fnc_getTargetDistance.sqf
@@ -28,11 +28,9 @@ if (count _LIS > 0) then {
     _distance = _maxDistance;
 };
 
-if (_distance < _minDistance) then {
-    _distance = _minDistance;
-};
+_distance = _distance max _minDistance;
 
-_accuracy = if(_accuracy < 1) then {1} else {_accuracy};
+_accuracy = _accuracy max 1;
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_getTargetDistance.sqf
```diff
@@ -14,33 +14,25 @@
  */
 #include "script_component.hpp"
 
-params ["_interval", "_maxDistance", "_minDistance"];
+params [["_accuracy",1], ["_maxDistance",5000], ["_minDistance",0]];
 
-private _position = ATLToASL positionCameraToWorld [0, 0, 0];
-_position set [2, (_position select 2) - (getTerrainHeightASL _position min 0)];
+private _camPosition = AGLToASL positionCameraToWorld [0, 0, 0];
```
esteldunedain: I know the former code would not have worked over water. The new one probably does, but it should be tested.

VKing6: Tested with land lasing boat and boat lasing land. Works perfectly.

esteldunedain: Yep, AGLtoASL is great. I've been testing that command today.

PR #3200, opened 2016-01-21
Hearing - Cache Weapon/Ammo loudness

PabstMirror:


esteldunedain:
:+1: Gotta cache em all.

I see that these days ACE tends to store its caches on dedicated location instead of the mission namespace, as it used to. Is this to avoid slowing all mission namespace variable reads due to many entries on the hashmap? Has this effect ever been measured?

PabstMirror:
Before we just cached magazines, this now caches the combo `_weapon,_ammo` so it's not a straight comparison (this is about 2x faster overall).

Biggest "polluter" of mission namespace now is probably the interaction menu.
Benchmarking the real impact isn't easy, but I'll try to get some real numbers,

esteldunedain:
> Before we just cached magazines, this now caches the combo _weapon,_ammo so it's not a straight comparison (this is about 2x faster overall).

Ok, what I meant is: did we ever observe caching to degrade overall reading of mission namespace variables? Anyway, I'm all for keeping all the stuff separated on its own buckets .

> Biggest "polluter" of mission namespace now is probably the interaction menu.

Should be easy to move to a a location, right?

PR #3221, opened 2016-01-25
Unified ellipsis

BaerMitUmlaut:
Every occurance of `...` in stringtables now has a leading space. The leading space was chosen in favour of no leading space because the latter was less used.

### DIFF
--- a/addons/logistics_uavbattery/stringtable.xml
+++ b/addons/logistics_uavbattery/stringtable.xml
@@ -62,16 +62,16 @@
             <Russian>Используется для зарядки переносных БПЛА</Russian>
         </Key>
         <Key ID="STR_ACE_Logistics_UAVbattery_Battery_Recharge">
-            <English>Recharging ...</English>
-            <German>Aufladen ...</German>
-            <Spanish>Recargando ...</Spanish>
-            <French>Rechargement ...</French>
-            <Polish>Ładowanie ...</Polish>
-            <Hungarian>Akku feltöltése ...</Hungarian>
-            <Czech>Dobíjení ...</Czech>
-            <Portuguese>Recarregando ...</Portuguese>
-            <Italian>In ricarica ...</Italian>
-            <Russian>Заряжается ...</Russian>
+            <English>Recharging...</English>
+            <German>Aufladen...</German>
+            <Spanish>Recargando...</Spanish>
+            <French>Rechargement...</French>
+            <Polish>Ładowanie...</Polish>
+            <Hungarian>Akku feltöltése...</Hungarian>
+            <Czech>Dobíjení...</Czech>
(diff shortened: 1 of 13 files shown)


jonpas:
Add it to coding guidelines (or translation guidelines).

PabstMirror:
The extra space looks a little odd to me, but some style guides say it's good.
Just being consistent is important.

commy2:
In German, having the space left and right of `...` is correct. I don't know about other languages.

esteldunedain:
The space is not correct in spanish, but I really don't care

BaerMitUmlaut:
@commy2: There's actually no strict rule for this specific case in German or English as far as I could tell. We're not leaving a part of the sentence out, we're just indicating that something is in progress.

For that case, at least in Windows no leading space is consistently used. For example, search your C: drive for anything, and it comes up with `Suchvorgang...` or your local equivalent. As far as I know, both [Linux](http://ivanblagojevic.com/wp-content/uploads/2014/03/how-to-install-ubuntu-14-04-pic-16.png) and [Mac](http://storage.uxguide.net/wiki-imgs/macos/ct_progindexample.jpg) also usually do not use a leading space.

I personally much prefer no leading space since that seems to be most commonly used in other programs, but I adjusted this to what seemed to be most commonly used in ACE3 thus far.

commy2:
@BaerMitUmlaut 
>For example, search your C: drive for anything, and it comes up with
Windows is wrong on this.

http://www.duden.de/sprachwissen/rechtschreibregeln/auslassungspunkte

BaerMitUmlaut:
@commy2 Not really:
>Drei Auslassungspunkte zeigen an, dass in einem Wort, Satz oder Text Teile ausgelassen worden sind <§ 99>.

Not the case.

commy2:
Read the examples. No space is only used when you cut of a word.

BaerMitUmlaut:
Hmm I suppose so.

BaerMitUmlaut:
Merge conflicts have been resolved.

esteldunedain:
@BaerMitUmlaut: we discussed this PR the other day during the merge session and, while we didn't make any final decision, the concensus seems to be that the leading space is wrong for most languages; German is the only one we knew of for which this was correct.

BaerMitUmlaut:
I'll gladly change it as I prefer no leading space as well.

BaerMitUmlaut:
Changed to no leading space everywhere.

esteldunedain:
Perfect; if Germans don't complain, I'd say this is ready for merging.

Thanks @BaerMitUmlaut

Glowbal:
:+1:

bux578:
ah fck it. who cares

PR #3222, opened 2016-01-25
Wind deflection with Intercept 

esteldunedain:
Hey, this is a proof-of-concept PR to start discussing how Intercept extensions could be integrated in ACE to replace performance critical functions.

I took @NouberNou wind deflection example from Intercept and shoehorned it into ACE, enabling the code as appropiate as required by ACE_settings, etc.

A fallback sqf version of winddeflection is conserved, which runs in case Intercept is not working.

Known problems and limitations:
- [ ] `sqf::is_nil` function is still missing, which makes it impossible to implement some tests involving variables that may not be defined
- [ ] Event handlers are unfinished in Intercept. Only some are forwarded from sqf code. The `fire` event handler is only binded for the player.
- [ ] On my machine it CTD when quitting Arma. This happened with `example_frag.dll` as well.
- [ ] The ACE mod can be loaded from any path. I don't know how to find that path to the extension dll that we need to supply to intercept.

### DIFF
--- a/addons/winddeflection/XEH_postInit.sqf
+++ b/addons/winddeflection/XEH_postInit.sqf
@@ -9,7 +9,7 @@ GVAR(trackedBullets) = [];
     if (!GVAR(enabled)) exitWith {};
 
     // Decide if it should be done on sqf or on the intercept extension
-    if (missionNamespace getVariable ["intercept_core_enabled", false]) then {
+    if (missionNamespace getVariable ["intercept_invoker_ok", false]) then {
         // Use the extension
         // @todo: figure out how to solve the extension being on @ace vs z\ace, etc
         "intercept" callExtension "load_extension:z\ace\ace_wind_deflection.dll";
--- a/addons/winddeflection/XEH_postInit.sqf
+++ b/addons/winddeflection/XEH_postInit.sqf
@@ -5,9 +5,20 @@ if (!hasInterface) exitWith {};
 GVAR(trackedBullets) = [];
 
 ["SettingsInitialized", {
-    //If not enabled, dont't add PFEH
+    // If not enabled, dont't add PFEH
     if (!GVAR(enabled)) exitWith {};
 
-    [] call FUNC(updateTrajectoryPFH);
+    // Decide if it should be done on sqf or on the intercept extension
+    if (missionNamespace getVariable ["intercept_core_enabled", false]) then {
+        // Use the extension
+        // @todo: figure out how to solve the extension being on @ace vs z\ace, etc
+        "intercept" callExtension "load_extension:z\ace\ace_wind_deflection.dll";
+        GVAR(disableEventHandlers) = true;
+        ACE_LOGINFO("ace_winddeflection enabled using intercept extension.");
+    } else {
+        // Go sfq
+        [] call FUNC(updateTrajectoryPFH);
+        ACE_LOGINFO("ace_winddeflection enabled using sqf.");
+    };
 
 }] call EFUNC(common,addEventHandler);
--- a/addons/winddeflection/XEH_preInit.sqf
+++ b/addons/winddeflection/XEH_preInit.sqf
@@ -17,3 +17,5 @@ PREP(updateTrajectoryPFH);
 PREP(initModuleSettings);
 
 ADDON = true;
+
+GVAR(disableEventHandlers) = false;
--- a/addons/winddeflection/functions/fnc_handleFired.sqf
+++ b/addons/winddeflection/functions/fnc_handleFired.sqf
@@ -23,6 +23,7 @@
 
 params ["_unit", "", "", "", "_ammo", "", "_bullet"];
 
+if (GVAR(disableEventHandlers)) exitWith {};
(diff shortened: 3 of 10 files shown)


commy2:
>Intercept

It begins!

NouberNou:
Awesome!

A couple of things: 

`is_nil` should be able to be handled via the `is_null()` method on objects (and probably can be extended to other complex types like groups, etc), though is_nil is probably good for checking the state of global variables in one pass.

Loading extensions soon will be done in a different way in terms of paths, probably in a specific folder inside a `@folder` or other addon location (basically a sister folder to the `addons` folder). That way it is safe and also organized.

The majority of mission event handlers should also be in now. There are a couple that might need tweaking, for example `handleDamage` doesn't return right now, so it can't really be used effectively, and `hitPart` is not populating its event data yet.

Also I need to look into that crash, I've seen it a few times.

As far as integration with ACE goes, I think the main thing we need to do is look at how we approach the design. The best way would be to take a serious game design approach to organizing code, and really look at the best way to consolidate systems. For example, in ACE right now, all systems that need to track where an ammo object goes basically does its own tracking. Instead with C++ it is much easier and nicer, and most importantly, performant if you can consolidate all of that into a single system, so for example with wind deflection you'd have modules that can assign functionality on each frame that a ammo object is travelling, and for frag you can have one on when an ammo object explodes, but they are all being tracked and delegated from a central system. This makes the code cleaner, and also easier to debug.

So it really isn't so much of an approach we should take to where we can drop in Intercept but where we can start migrating functionality as a whole to Intercept.

esteldunedain:
> Loading extensions soon will be done in a different way in terms of paths, probably in a specific folder inside a @folder or other addon location (basically a sister folder to the addons folder).

Excellent, something like that is needed.

> The majority of mission event handlers should also be in now. 

But we need to forward them from sqf, right? I don't think the forwarding is done yet,

> Also I need to look into that crash, I've seen it a few times.

The crash is happening every time in my machine even if I don't load any module besides Intercept  core.

> As far as integration with ACE goes, I think the main thing we need to do is look at how we approach the design. The best way would be to take a serious game design approach to organizing code, and really look at the best way to consolidate systems. For example, in ACE right now, all systems that need to track where an ammo object goes basically does its own tracking. Instead with C++ it is much easier and nicer, and most importantly, performant if you can consolidate all of that into a single system

I completely agree. All the following system require bullet tracking to some degree: wind deflection, frag, spall, suppression (not in, something I'm looking at), and it's pretty reasonable to have a main bullet tracker that delegate.

Of course that approach would also make it a bit harder to replace functionality piece-wise, but it's probably worth it.

Finallly @NouberNou, is there a way for sqf code to communicate with an intercept extension? If not, I think that would probably help a lot for initializing, turning features on/off, etc. The only signal I can get right now are loading the extension (which works once) and creating GVARs (which need repeated polling in the c++ code). I'm thinking on something like this :

´´´sqf
missionNamespace setVariable ["intercept_extension_params_var", +_this];
_res = "intercept" callExtension "signal:ace_wind_deflection:intercept_extension_params_var";
´´´

probably wrapped in a nice sqf function in core:

```sqf
[_extensionName, _params] call intercept_core_fnc_signalExtension.
```

jokoho48:
:-1: 
i dont like the idea to switch to intercept. 

please dont understand me wrong intercept is a awsome tool and is a very cool idea but i think that it 1. is a to big maintainment to keep track where the issue is later if we have 2x the same code that do basicly the same thing and 2. i think that intercept never gets Battleye signed because it write and read in the memory what basicly open the door to too much people how want to do ugly thinks with that is a Major Security problem.
please dont understand me wrong intercept is a awsome tool and is a very cool idea but i think that it 1. is a to big maintainment to keep track where the issue is later if we have 2x the same code that do basicly the same thing and 2. i think that intercept never gets Battleye signed because it write and read in the memory what basicly open the door to too much people how want to do ugly thinks with that.

And one of the points is that the Base ACE team is very small and we have only a few core member can write c++ and the ACE mod is based on much community work(PRs) that i think the last few member that left are not able to handle the maintainmence of all the ACE Modules.

Glowbal:
We have not discussed the possibility of switching to intercept within the ace team @jokoho48. Perhaps once intercept is at a more mature state, we will discuss it. Until such time and perhaps even after, we stick to the way we are doing things currently, which is SQF.

dedmen:
> sqf::is_nil function is still missing, which makes it impossible to implement some tests involving variables that may not be defined
Is implemented now
>Event handlers are unfinished in Intercept. Only some are forwarded from sqf code. The fire event handler is only binded for the player.
Is implemented on `develop` branch you can bind any addEventHandler/addMissionEventhandler directly to a c++ Function/Lambda
>On my machine it CTD when quitting Arma. This happened with example_frag.dll as well.
Fixed :3
>The ACE mod can be loaded from any path. I don't know how to find that path to the extension dll that we need to supply to intercept.
All Plugins load from @mod/intercept/extensionName(_x64).dll

bux:
Do wee need to keep this PR open?

PR #3223, opened 2016-01-25
Remove VehicleLock Items from Virtual Arsenal

### LINKED ISSUE #3215: Ace 3 KEYS
Hi, I got one little request . Is there any way , I can disable Master key for players ? Or can you just give me info , in which pbo are keys located, so I can delete It myself ? I would be very thankful. 

Cheers.

PabstMirror:
Close #3215

Removes all the keys from the VA so they are not obtainable from magic VA boxes in missions.

### DIFF
--- a/addons/vehiclelock/CfgMagazines.hpp
+++ b/addons/vehiclelock/CfgMagazines.hpp
@@ -1,6 +1,7 @@
 class CfgMagazines {
     class CA_Magazine;
     class ACE_key_customKeyMagazine: CA_Magazine {
+        scopeArsenal = 0;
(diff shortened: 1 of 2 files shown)


esteldunedain:
:+1:

commy2:
Last time I checked, `scope` and `scopeArsenal` did not work in `CfgMagazines`

BreberaCZ:
nice. Just one little thing - how will we access the keys then ? through crates created by zeus ? and wont the keys be accessible from VA by just adding class names of keys into your loadout ?

aristenn:
Wouldn't this make more sense to be something that the mission maker could control?

PR #3230, opened 2016-01-28
Fix dragging moving object to random location

PabstMirror:
- modern privates / params / headers
 - functional changes in https://github.com/acemod/ACE3/commit/3f1d6ffeb73eb174331dad0585385f7a6a78a5a9

Fixes a critical bug we've seen a few times where a person or object being dragged will be teleported to a random location often in the middle of the ocean (which is not good if that person is unconscious :smile:).

Problem seems to be that `attachTo` has effects on getPos and distance for a few frames.

debug from ace_dragging_fnc_dragObject
```
attachTo [2709.28,6047.91,2.31902] @ 1946.34
tick [-2532.71,1995.72,4.07294] @ 1946.36
tick [-2532.71,1995.73,4.07294] @ 1946.38
tick [-2532.71,1995.73,4.07294] @ 1946.39
tick [-2532.71,1995.74,4.07294] @ 1946.41
tick [-2532.7,1995.74,4.07293] @ 1946.43
tick [2709.3,6047.94,2.31901] @ 1946.44
tick [2709.3,6047.95,2.31901] @ 1946.46
```

The same frame as `attachTo`, getPos is stable, then for a few frames it's some weird random location then it settles back to where it should be. dragObjectPFH could fail the distance check and permently set the object to that random position.

Only happens in mp, reproducible ~20% of all attempts done on a pier over water.

Also hides the scroll wheel action hint when carrying man.

### DIFF
--- a/addons/dragging/functions/fnc_canCarry.sqf
+++ b/addons/dragging/functions/fnc_canCarry.sqf
@@ -20,6 +20,10 @@ params ["_unit", "_target"];
 
 if !([_unit, _target, []] call EFUNC(common,canInteractWith)) exitWith {false};
 
+//#2644 - Units with injured legs cannot bear the extra weight of carrying an object
+//The fireman carry animation does not slow down for injured legs, so you could carry and run
+if ((_unit getHitPointDamage "HitLegs") >= 0.5) exitWith {false};
+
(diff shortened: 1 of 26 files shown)


esteldunedain:
:+1: Nice

bux578:
GJ!

PR #3244, opened 2016-02-01
Fix ace_scopes capturing native adjustment keys for the PSO scope. 

### LINKED ISSUE #3238: Scopes with only windage adjustment captures keybind for range adjustment
If a scope has `ACE_ScopeAdjust_Vertical[] = {0,0}` , #3178 has fixed the audiovisual cues appearing. However the adjustment keys are still being captured, which goes against the intention that if vertical adjust is disabled, vanilla range adjust should be used instead.

This isn't possible at the moment unless separate keys are bound to ACE scope adjust and vanilla range adjust.

Paging @KoffeinFlummi, @bux578, @commy2, and/or @esteldunedain.

esteldunedain:
This was first and foremost a data problem. Due to the way `ace_scopes_fnc_adjustScope` is coded,  `ACE_ScopeAdjust_VerticalIncrement` needs to be none zero to quit when adjusting isn't possible.

~~It would also be possible to make changes to that function to prevent this from happening when zero is provided. Should I do that too?~~ I just did, so it's possible to set increment to zero to disable adjusting in that direction without capturing keys.

Close #3238

### DIFF
--- a/addons/scopes/functions/fnc_adjustScope.sqf
+++ b/addons/scopes/functions/fnc_adjustScope.sqf
@@ -44,7 +44,8 @@ _maxVertical = getArray (_opticConfig >> "ACE_ScopeAdjust_Vertical");
 _maxHorizontal = getArray (_opticConfig >> "ACE_ScopeAdjust_Horizontal");
 
 if ((count _maxHorizontal < 2) || (count _maxVertical < 2)) exitWith {false};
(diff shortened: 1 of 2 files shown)


esteldunedain:
@VKing6, could you confirm this fixes the issue pease?

VKing6:
Yup that seems to have fixed it :+1:

PR #3249, opened 2016-02-02
Added a real nozzle

GitHawk:
Added code for nozzle object
Params and private update

Special thanks to @jokoho48 for the modell

nozzle in action


esteldunedain:
That looks perfect. Congrats to both of you

licht-im-Norden87:
Wow, astonishing work! Really like it. But maybe the texture does not look "used" and could need some improvement.

Glowbal:
I am happy with this. If we want to change the model, we can do that at a later stage. What @licht-im-Norden87 is suggesting does make sense. What are your thoughts on that @GitHawk and @jokoho48 ?

GitHawk:
i'm no graphics guy. Everybody with some skills can now improve the nozzle further

jokoho48:
@Glowbal there are some issues with the model import @GitHawk done i will fix that in a PR in some min

GitHawk:
Possible, was the first time I used the Object Builder.

PR #3250, opened 2016-02-02
Intercept Nametags

esteldunedain:
WIP

### DIFF
--- a/extensions/nametags/ace_nametags.cpp
+++ b/extensions/nametags/ace_nametags.cpp
@@ -1,0 +1,102 @@
+/*
+ * ace_nametags.cpp
+ * Author: esteldunedain
+ * Implements nametags using Intercept
+ *
+ * Takes:
+ * None
+ *
+ * Returns:
+ * None
+ */
+#include <Windows.h>
+#include <stdio.h>
+#include <cstdint>
+#include <atomic>
+#include <memory>
+
+#include "intercept.hpp"
+#include "logging.hpp"
+#include "client\client.hpp"
+#include "client\pointers.hpp"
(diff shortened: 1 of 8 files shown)


extensions/nametags/ace_nametags.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "intercept.hpp"
+#include "logging.hpp"
+#include "client\client.hpp"
+#include "client\pointers.hpp"
+#include "client\sqf\uncategorized.hpp"
+#include "client\sqf\config.hpp"
+
+#include "nametagger.hpp"
+
+#define PI 3.14159265f
```
bux578: `include <math.h>`
`M_PI`
?

esteldunedain: Sure. But I think I'm not even using PI or any of the math functions. It's just copy/pasted crap.

dedmen:
I worte an updated version to latest Intercept version at https://github.com/dedmen/intercept_ace_nametags
It's just a quick mockup.

bux:
Do we need to keep this open?

dedmen:
Is there still Interest in Intercept Nametags/https://github.com/acemod/ACE3/pull/3222?
If yes I would rebuild it.
The old Intercept PR's are very outdated and would need a complete rewrite.

jonpas:
Then I say close when new one is open.

I think there is still interest, I am all for starting to use Intercept, as long as we can do it optionally and not force something that might not work or be unstable (especially since Intercept is not 1.0 yet), it will help greatly with Intercept development and awareness IMO.

PR #3253, opened 2016-02-03
Improve Cargo Network Trafic

jokoho48:
This PR reduce the Network Trafic because if you attachTo something its get Synced every Frame and setPos is more static and dont create less Trafic

### DIFF
--- a/addons/cargo/functions/fnc_loadItem.sqf
+++ b/addons/cargo/functions/fnc_loadItem.sqf
@@ -35,7 +35,8 @@ _vehicle setVariable [QGVAR(space), _space - _itemSize, true];
 
 if (_item isEqualType objNull) then {
     detach _item;
-    _item attachTo [_vehicle,[0,0,-100]];
+    _item setPos [-100000,-100000,-100000];
+    _item allowDamage false;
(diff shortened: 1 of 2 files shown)


PabstMirror:
Unloaded object is invulnerable.

esteldunedain:
> if you attachTo something its get Synced every Frame

Was this confirmed? If it's implemented properly in the engine it should not cause extra updates.

addons/cargo/functions/fnc_loadItem.sqf
```diff
@@ -35,7 +35,8 @@ _vehicle setVariable [QGVAR(space), _space - _itemSize, true];
 
 if (_item isEqualType objNull) then {
     detach _item;
-    _item attachTo [_vehicle,[0,0,-100]];
+    _item setPos [-100000,-100000,-100000];
+    _item allowDamage false;
```
Glowbal: I do not like this. A better approach might be to add a `handleDamage` eventhandler and return current damage instead

Glowbal:
Merge conflicts and using allowDamage. We have no issues reported with the current approach.

Issue #3134, opened 2016-01-09
Medical Focus Feature discussion


thojkooi:
Tasks for focus feature medical:
- [x] Prevent instant death & Revive are confusing for players. Look into a single setting method instead
- [x] Unconsciousness 
  - [x] It is hard to figure out when someone is stuck in revive or just regular unconscious
  - [x] Players find it confusing how to wake up an unconscious patient
  - [x] See about improving the unconsciousness effect for players
- [x] Improve triagecard (#1658)
- [x] Medical Menu does not seem to be widely known. Make it default? - 604ba50a4a6340c545bfcd12f960e8847a4cec49
- [x] Merge basic and advanced medical injury systems
- [x] Clarify the amount of pain as feedback to a medic (#2117)
- [x] bandaging takes to much effort since latest update (3.4.0), look into improved values or alternative method ?
- [x] tourniquets should have an effect on medication (medication localized to body parts?) (#2704 #3255)
- [x] add option to treat everyone as a medic in medical facility (#1259)
- [x] Go over all treatment actions and improve triage card reporting
- [x] Fix surviving exploding vehicles

---

Please provide input on the above mentioned items.

gienkov:
Unconscious - maybe different lying down animation for both states?
09.01.2016 17:49 "Thomas Kooi" <notifications@github.com> napisał(a):

BaerMitUmlaut:
>Prevent instant death & Revive are confusing for players. Look into a single setting method instead

This sounds better, however a (optional) reason that makes medevac still necessary would be great. Something like permanent damage. Broken limbs would for example be an idea, or lower overall stamina, etc.

AGM had a maximum unconsciousness timer. Something like a proper maximum cardiac arrest time could be a replacement for the revive state.

-----

Something completly different, but I think there are a few unused functions in medical. It needs  some cleaning up.

Gwynblade:
Here's couple of things I personally would like to see, you guys ignore or implement them as you see fit :)

I really like the advanced medical, there are some really nice improvements over ACE2 done, but there are some nice stuff (from a gameplay standpoint, not sure about realism) I would like to see make a return.

-Unconsciousness (not talking about being able to see the AI/player is not dead) - It used to have a pretty big impact on gameplay, both on unconscious player and medic or other teammates. It used to take from ~10-15 seconds on light injuries, up to minute or two to wake up, if you're badly injured (sometimes you can't even wake up when you're really messed up)

[...]

PR #3255, opened 2016-02-03
Tourniquets block medication until released

### LINKED ISSUE #2704: tourniquets have effect on injected medication
ACE3 Version: 3.3.2

Happens on Stable, Dev and RC Builds of ArmA 3

Mods:

@cba_a3
@ace3

Injected drugs still have full effect when applied to a limb with a tourniquet on it. I'm pretty sure if a tourniquet is applied correctly it would cut off all blood flow to that targeted limb, any drug applied to the targeted limb would only have local effects. I know the example I am about to list is a completely different drug, but the circulatory system is the highlight. 

May 3, 1995. Missouri. Emmitt Foster. Lethal Injection. Seven minutes after the lethal chemicals began to flow into Foster's arm, the execution was halted when the chemicals stopped circulating. . Death was pronounced thirty minutes after the execution began. According to William "Mal" Gum, the Washington County Coroner who pronounced death, the problem was caused by the tightness of the leather straps that bound Foster to the execution gurney; it was so tight that the flow of chemicals into the veins was restricted. Foster did not die until several minutes after a prison worker finally loosened the straps. The coroner entered the death chamber twenty minutes after the execution began, diagnosed the problem, and told the officials to loosen the strap so the execution could proceed.

source: http://www.deathpenaltyinfo.org/some-examples-post-furman-botched-executions

Steps to reproduce:
1) tourniquet a limb
2) apply morphine/epinephrine/atropin to the limb with a tourniquet on it

PabstMirror:
Close #2704

### DIFF
--- a/addons/medical/functions/fnc_actionRemoveTourniquet.sqf
+++ b/addons/medical/functions/fnc_actionRemoveTourniquet.sqf
@@ -17,6 +17,7 @@
 
 private ["_part", "_tourniquets", "_output"];
 params ["_caller", "_target", "_selectionName"];
+TRACE_3("params",_caller,_target,_selectionName);
 
 // grab the required data
 _part = [_selectionName] call FUNC(selectionNameToNumber);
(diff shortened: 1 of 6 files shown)


addons/medical/functions/fnc_actionRemoveTourniquet.sqf
```diff
@@ -34,3 +35,20 @@ _target setVariable [QGVAR(tourniquets), _tourniquets, true];
 
 // Adding the tourniquet item to the caller
 _caller addItem "ACE_tourniquet";
+
+//Handle all injected medications now that blood is flowing:
+private _delayedMedications = _target getVariable [QGVAR(occludedMedications), []];
+private _removed = [];
+TRACE_2("meds",_part,_delayedMedications);
+{
```
jokoho48: why you dont use the new select?

commy2: Current Arma version is 1.54 and doesn't have select CODE

jokoho48: but isnt 3.5.0 targeted for 1.56?

addons/medical/functions/fnc_actionRemoveTourniquet.sqf
```diff
@@ -34,3 +35,20 @@ _target setVariable [QGVAR(tourniquets), _tourniquets, true];
...
+    _x params ["", "", "_medPartNum"];
+    if (_part == _medPartNum) then {
+        TRACE_1("delayed medication call after tourniquet removeal",_x);
+        ["medical_advMedication", [_target], _x] call EFUNC(common,targetEvent);
+        _removed pushBack _x;
```
Glowbal: Why not replace the value by `objNull` and delete those after the for each? I'm not sure i Ilike the array for removal.

Glowbal:
Nice work replacing those execRemote by events @PabstMirror :+1:

Glowbal:
#3134

PR #3276, opened 2016-02-06
replace filter and map with select and apply

commy2:


addons/inventory/XEH_postInit.sqf
```diff
@@ -42,7 +42,9 @@ GVAR(Medical_ItemList) = [];
 );
 
 // make list case insensitive
-GVAR(Medical_ItemList) = [GVAR(Medical_ItemList), {if (_this isEqualType "") then {toLower _this}}] call EFUNC(common,map);
+GVAR(Medical_ItemList) = GVAR(Medical_ItemList) apply {
+    if (_x isEqualType "") then {toLower _x};
+};
```
commy2: what if not string ?

addons/map/functions/fnc_blueForceTrackingUpdate.sqf
```diff
@@ -17,14 +17,14 @@ if (GVAR(BFT_Enabled) and {(!isNil "ACE_player") and {alive ACE_player}}) then {
...
-        _groupsToDrawMarkers = [allGroups, {
+        _groupsToDrawMarkers = allGroups select {
             _anyPlayers = {
                 [_x] call EFUNC(common,isPlayer);
-            } count units _this;
-            (side _this == _playerSide) && _anyPlayers > 0
-        }] call EFUNC(common,filter);
+            } count units _x;
+            (side _x == _playerSide) && _anyPlayers > 0
+        }; // @todo, simplify this
```
commy2: todo

addons/map/functions/fnc_determineMapLight.sqf
```diff
@@ -95,11 +95,12 @@ TRACE_1("Player is on foot or in an open vehicle","");
...
-    alive _this && {(typeOf _this == "Chemlight_red") || {
-                    (typeOf _this == "Chemlight_green") || {
-                    (typeOf _this == "Chemlight_blue") || {
-                    (typeOf _this == "Chemlight_yellow")}}}}}] call EFUNC(common,filter);
+_nearObjects = (_unit nearObjects ["SmokeShell", 4]) select {
+    alive _x && {(typeOf _x == "Chemlight_red") || {
+                    (typeOf _x == "Chemlight_green") || {
+                    (typeOf _x == "Chemlight_blue") || {
+                    (typeOf _x == "Chemlight_yellow")}}}}}; // @todo, simplify this
+
```
commy2: todo

addons/inventory/XEH_postInit.sqf
commy2:
```
ARR = [1,"A",3]
```
```
ARR apply {
    if (_x isEqualType "") then {toLower _x};
};
```
-> `[<null>,"a",<null>]`

VKing6:
Should we also add other implementations that could use select or apply to this PR or do you want to keep it for only where the functions are replaced?

commy2:
Feel free to add more of these to this PR. I was just replacing the obvious uses of the functions with it for now, but there are certainly more places (that currently use forEachIndex for example) that could benefit from these.

PR #3277, opened 2016-02-06
update cba files under tools

commy2:
latest (2.3.0+) CBA macros
also removes `XEH_ENABLED;` from `ReammoBox_F`. It has XEH enabled by default now.


Glowbal:
:+1:

PR #3278, opened 2016-02-06
remove old ragdoll tweaks

commy2:
Removes ragdoll tweaks from AGM in favour of the changes in 1.56 from BI


Glowbal:
How do the changes in 1.56 compare to our version?

PR #3279, opened 2016-02-06
Overhaul ACE_Hearing and change default settings.

esteldunedain:
This PR makes the following changes:
- Ear ringing is now not independent of hearing loss, and it's not applied unless hearing loss is simulated (like in ACE2)
- ACE2 sound effect for combat deafness has two components: a low pitch woosh and a high pitch ring
- Now, the low woosh is always played when hearing loss is simulated, cause only lowering volume sounds weird if no sound effect goes with it.
- Now the high pitch ring plays or no now depends on the "Disable Ear ringing" setting, which is client settable.
- Simplified the code logic and made minor optimizations
- Sound effect now have two volume levels each, so effects can start more gradually.
- Sound effects overlap so it's not noticeable when a clip starts/ends
- Volume is now never completely lowered to zero, so the effects are still audible. This circunvents a bug in `playSound`that prevents sounds from being forced to play indenpendently of volume.

I might still mess with the thresholds, volumes, etc and optimize the code.

Please test and provide feedback.

This is the frequency analysis for both clips (with ring on top/without in the bottom):
image


addons/hearing/XEH_postInit.sqf
```diff
@@ -9,31 +9,31 @@ GVAR(deafnessDV) = 0;
...
-GVAR(time2) = 0;
 GVAR(time3) = 0;
-GVAR(time4) = 0;
+GVAR(damageCoefficent) = 1;
 
 ["SettingsInitialized", {
+    TRACE_1("settingInit",GVAR(EnableCombatDeafness));
+    // Only run PFEH and install event handlers if combat deafness is enabled
+    if (!GVAR(EnableCombatDeafness)) exitWith {};
+
```
esteldunedain: @PabstMirror, we should probably call `fnc_updateHearingProtection` here once, in case `SettingsInitialized` arrive when the player is already fully formed.

addons/hearing/functions/fnc_updateVolume.sqf
```diff
@@ -78,10 +52,9 @@ if ([ACE_player] call FUNC(hasEarPlugsIn)) then {
 };
 
 // Headgear can reduce hearing
-if(headgear ACE_player != "") then {
-    private ["_lowerVolume"];
-    _lowerVolume = (getNumber (configFile >> "CfgWeapons" >> (headgear ACE_player) >> QGVAR(lowerVolume))) min 1;
-    if(_lowerVolume > 0) then {
+if (headgear ACE_player != "") then {
+    private _lowerVolume = (getNumber (configFile >> "CfgWeapons" >> (headgear ACE_player) >> QGVAR(lowerVolume))) min 1;
```
esteldunedain: @PabstMirror that part could also be cached on your function

addons/hearing/XEH_postInit.sqf
esteldunedain: Done

PR #3280, opened 2016-02-06
Unified fired event handler

esteldunedain:
Optimization of fired Event Handlers

Fired Event Handlers in ACE:
- [x] advanced_ballistics
- [x] disposable
- [x] fcs
- [x] frag
- [x] goggles
- [x] grenades
- [x] huntir
- [x] nightvision
- [x] optics
- [x] overheating
- [x] overpressure
- [x] recoil
- [x] scopes
- [x] weaponselect
- [x] winddeflection

I won't do:
- [ ] javelin (I don't dare)
- [ ] missileguidance (I don't dare)
- [ ] mk6mortar (it's faster not to)


esteldunedain:
Well, except for the inevitable bugs this is ready. All the listed modules should be tested. 

This should offer noticeable performance gains when AI fires. From the 16 modules, only 3 currently listen for AI fires (and one of them is default off).

esteldunedain:
Ongoing testing:
- [ ] advanced_ballistics
- [x] disposable
- [x] fcs
- [ ] frag
- [ ] goggles
- [ ] grenades
- [ ] huntir
- [ ] nightvision
- [ ] optics
- [ ] overheating
- [x] overpressure
- [ ] recoil
- [ ] scopes
- [ ] weaponselect
- [x] winddeflection

esteldunedain:
I completed preliminary testing of all modules and reviewed the diff. After fixing a couple of errors all typical cases of the features work (corner cases should, but they are hard to test systematically).

@PabstMirror, unfortunately the version you checked was running most event handlers, so performance should be measured again. If possible, I'd love to see the impact on AI.

@Glowbal, after performance is measured, I'd suggest this PR is merged well before 3.5.0 so any obvious bug can be found.

Glowbal:
:+1: Happy with that @esteldunedain

esteldunedain:
@PabstMirror measured the following performance improvement:

## Player firing

Before:
`["B_HeavyGunner_F","Extended_FiredBISEH"]: Overall Average: 0.247898s / 390 = 0.635636ms`

After: 
`["B_HeavyGunner_F","Extended_FiredBISEH"]: Overall Average: 0.190063s / 390 = 0.487342ms`

Time saving: 23%

## AI firing

Before:
`0.286349ms`

After: 
`0.127675ms`

Time saving: 55%

Based on that, I think we may merge and test @Glowbal.

addons/advanced_ballistics/XEH_postInit.sqf
```diff
@@ -22,3 +22,17 @@ if (!GVAR(extensionAvailable)) exitWith {
...
+
+if (!hasInterface) exitWith {};
+
+["SettingsInitialized", {
+    //If not enabled, dont't add PFEH
+    if (!GVAR(enabled)) exitWith {};
+
+    // Register fire event handler
+    ["firedPlayer", DFUNC(handleFired)] call EFUNC(common,addEventHandler);
+    ["firedPlayerNonLocal", DFUNC(handleFired)] call EFUNC(common,addEventHandler);
```
jokoho48: i think here we can add a what we check in  addons/advanced_ballistics/functions/fnc_handleFired.sqf Line 28

esteldunedain: Yes, I think you're right

esteldunedain: Nah, that's not possible. There'sa additional conditions inside the if of line 28, and there's no easy way to know in advance if they'll be true or false, except if GVAR(simulateForSnipers) == false and GVAR(simulateForGroupMembers) = false. I think it's a bit to hard for a too small gain

addons/common/functions/fnc_firedEH.sqf
```diff
@@ -0,0 +1,68 @@
...
+    {
+        if (_weapon in (_unit weaponsTurret _x)) exitWith {
+            _gunner = _unit turretUnit _x;
+            _turret = _x;
+        };
+        false
+    } count allTurrets [_unit, true];
+    // Ensure that at least the pilot is returned if there is no gunner
+    if (isManualFire _unit && {isNull _gunner}) then {
+        _gunner = driver _unit;
```
PabstMirror: I've never seen any actual use for ManualFire so this is really minor, but should this be `effectiveCommander` instead of `driver` as it seems like only the commander can do manual fire and driver could be null?

Possibly change FUNC(getGunner) as well?

PR #3282, opened 2016-02-06
Use location hash for ace events

PabstMirror:


esteldunedain:
:+1: 

This should help speed up #3280 more.

PabstMirror:
I don't think it's a problem (probably a good thing) but `getVariable` will ignore case whereas the old `find` would be strict on event capitalization.

e.g. Before x would be unset, now it would be set to 5.
```
["A", {x = _this}] call ace_common_fnc_addEventHandler;
["a", 5] call ace_common_fnc_localEvent;
```

esteldunedain:
> I don't think it's a problem

I agree

PR #3283, opened 2016-02-07
Make explosion reflections optional, default off

esteldunedain:
Currently all frag rounds are calculating reflections when they explode. While this is cool, it should be noted that these are many times more costly that pure frags.

This PR makes reflections independently toggable and defaults them off.

As for numbers, a single grenade exploding on the ground, clear of obstacles takes:

Without reflections:
```
23:31:02 ace_frag_frago_counter: Average: 0.0039978s / 1 = 3.9978ms
```

With reflections:
```
23:35:17 ace_frag_frago_counter: Average: 0.00299072s / 1 = 2.99072ms
23:35:17 ace_frag_fnc_findReflections_counter: Average: 0.0610046s / 10 = 6.10046ms
```

That means that time goes from 3ms in just one frame to 64 ms over 10 frames. The later produces a very noticeable frame drop. Notice that those times are code only, and don't account for the handling in the engine of the multiple explosion objects nor the network traffic it requires.

Disabling reflections could give a larger margin to finish adding frags to some vehicle projectiles for which they are missing, while improving overall performance.

--------------

Bottomline, this should take frags from producing very noticeable lag spikes to being imperceptible.


Glowbal:
While we are visiting frag, perhaps it's an idea to add the categories for the settings in that module?

esteldunedain:
> While we are visiting frag, perhaps it's an idea to add the categories for the settings in that module?

But none of the settings are client settable; do we need categories in that case?

Glowbal:
It's useful for the settings export option we have.

esteldunedain:
@Glowbal this should be ready for merging now.

Hopefully I'll PR a rewritten version of the module for the next release.

PR #3288, opened 2016-02-07
Repair & medical object attributes for eden editor

SzwedzikPL:
http://i.imgur.com/LXfNjDB.jpg //units
http://i.imgur.com/ad0qBLq.jpg //vehicles
http://i.imgur.com/j6S7Tmo.jpg //props

### DIFF
--- a/addons/medical/CfgEden.hpp
+++ b/addons/medical/CfgEden.hpp
@@ -1,0 +1,48 @@
+class Cfg3DEN {
+    class Object {
+        class AttributeCategories {
+            class StateSpecial {
+                class Attributes {
+                    class ACE_IsMedic {
+                        property = QUOTE(ACE_IsMedic);
+                        value = 0;
+                        control = "Combo";
+                        displayName = CSTRING(AssignMedicRoles_role_DisplayName);
+                        tooltip = CSTRING(Attributes_isMedic_Description);
+                        expression = QUOTE(_this setVariable [ARR_2(QUOTE(QGVAR(medicClass)),_value)];);
+                        typeName = "NUMBER";
+                        condition = "objectBrain";
+                        defaultValue = 0;
+                        class values {
+                            class none {name = CSTRING(AssignMedicRoles_role_none); value = 0; default = 1;};
+                            class medic {name = CSTRING(AssignMedicRoles_role_medic); value = 1; default = 0;};
+                            class doctor {name = CSTRING(AssignMedicRoles_role_doctor); value = 2; default = 0;};
+                        };
+                    };
+                    class ACE_isMedicalVehicle {
+                        property = QUOTE(ACE_isMedicalVehicle);
+                        value = 0;
+                        control = "CheckboxNumber";
+                        displayName = CSTRING(AssignMedicVehicle_enabled_DisplayName)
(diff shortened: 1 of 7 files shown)


jokoho48:
:+1:

jonpas:
Should we use a tag in those settings to set them apart from vanilla ones and make it clear it's an ACE setting? Eg. `[ACE] Is Medic`.

addons/medical/stringtable.xml
```diff
@@ -3891,5 +3891,13 @@
             <Portuguese>Não existe nenhum torniquete nesta parte do corpo!</Portuguese>
             <Czech>Žádné škrtidlo na této části těla!</Czech>
         </Key>
+		<Key ID="STR_ACE_Medical_Attributes_isMedic_Description">
+			<English>Medical training</English>
+			<Polish>Wyszkolenie medyczne</Polish>
+		</Key>
+        <Key ID="STR_ACE_Medical_Attributes_isMedicalVehicle_Description">
+            <English>Whatever or not the objects will be a medical vehicle.</English>
```
jonpas: Probably meant to be `Whether` and `object`.

Daishiknyte:
Jonpas is correct.  Any added commands, menus, etc., should clearly show the mod they're associated with.

esteldunedain:
:+1: Nice. I agree that some indication that those settings are ace would be great. Maybe:
`ACE: Is Medic:`

SilentSpike:
I feel like having categories for ace modules would make more sense than using the special states category. Also eliminates the need to repetitively label each attribute with ace in the name.

commy2:
^ this

addons/medical/CfgEden.hpp
```diff
@@ -0,0 +1,48 @@
+class Cfg3DEN {
+    class Object {
+        class AttributeCategories {
+            class StateSpecial {
+                class Attributes {
+                    class ACE_IsMedic {
```
commy2: lower case is

addons/repair/CfgEden.hpp
```diff
@@ -0,0 +1,48 @@
+class Cfg3DEN {
+    class Object {
+        class AttributeCategories {
+            class StateSpecial {
+                class Attributes {
+                    class ACE_IsEngineer {
```
commy2: lower case is

SzwedzikPL:
I've added ace category and used common_Options stringtable

http://i.imgur.com/Z1tI6b3.jpg //units
http://i.imgur.com/KRwpL0Y.jpg //vehicles
http://i.imgur.com/CNoeQJe.jpg //props

gienkov:
:+1:
10.02.2016 01:18 "SzwedzikPL" <notifications@github.com> napisał(a):

esteldunedain:
Looks good @SzwedzikPL

esteldunedain:
:+1: For merging this for 3.5.0

PabstMirror:
@SzwedzikPL 
These do not work in MP, only the server has the correct settings.
```
        class AttributesBase {
            class Default {
                expression = "_this setVariable ['%%s',_value,true];";
            };
```
The default uses the broadcast flag whereas these do not.
We might want to just use inheritance instead of redefining `expression`.
As far as I can tell only the server executes the code.

Also see notes in #3339

Issue #538, opened 2015-04-13
Missing nametags for cargo units


bux578:
Testing Session 2015 04 12

![2015-04-12_00014](https://cloud.githubusercontent.com/assets/1235520/7111818/02bbf272-e1c4-11e4-960b-7477256efcd3.jpg)

In case this isn't by design.

commy2:
Wouldn't work. all getPos / modelToWorld etc. commands return the position of the vehicle. All nametags would be at the same place in the center of the vehicle.

commy2:
Wouldn't work. all getPos / modelToWorld etc. commands return the position of the vehicle. All nametags would be at the same place in the center of the vehicle.

CorruptedHeart:
I know one mod that manages to show names for people in vehicles. [ZAM Show Names](http://forums.bistudio.com/showthread.php?185380-ZAM-ShowNames)

jaynus:
It looks like it uses the boundingBox of the model, which may be a visual render location instead of position. 

This could actually prove useful in other places as well.

adampiskorski:
In my mod it does show names in vehicles, and it works perfectly for stationary vehicles, but when you start moving then the positions get offset.

esteldunedain:
> In my mod it does show names in vehicles, and it works perfectly for stationary vehicles, but when you start moving then the positions get offset.

That should be fixable. You probably need to apply this corection:

```sqf_corre

adampiskorski:
Thanks for the tip!

jaynus:
Can you clarify how its being done? Are you using visual model rendering positions bounding boxes or some other factor for it?

adampiskorski:
It's been a while since I rewrote it, so I'll have to jog my memory, but my code is quite nicely commented (for me at least so that I can easily remember what I was doing) and it's nicely explained here: https://bitbucket.org/zulualpha/zam-shownames/src/fa3e5a18fd9f4cac45cd0fe121cfac29cf34941b/Addons/zam_shownames/functions/fn_showNames_OnEachFrame.sqf?at=default

esteldunedain:
`fn_showNames_OnEachFrame.sqf` just uses `modelToWorldVisual`; @commy2 said it didn't work inside vehicles; he's either wrong or some other function is used for those cases

adampiskorski:
Ill have to look into it, but if my mod doesn't show nametags in vehicles above the passengers heads, then this is the first I am hearing of it.

jaynus:
[...]

PR #3290, opened 2016-02-08
Nametags optimization

esteldunedain:
Performed a cleanup and optimization of ace_nametags, with the following changes:
- Once for each frame, current settings and button presses are converted to a set of quick flags to enable/disable code branches.
- Valid targets for nametags are computed every 0.5 secs and cached.
- Parameters for the nametag of each units are computed every 0.1 secs and cached.
- Now uses `lineIntersectsSurfaces`for LOS check, so units can block tags from units behind them (meaning less tags and less clutter).

Known issues:
- Icon caching means that soundwaves will have a delay between 0 and 0.1 sec from the time PTT button is pressed. Also the icon will change every 0.1sec instead of all frames.

For the scenario pictured below:

Before:
```
15:39:57 ace_nametags_ace_nametags_onDraw3d_counter: Average: 10.9531s / 3126 = 3.50388ms
```

After:
```
15:35:39 ace_nametags_ace_nametags_onDraw3d_counter: Average: 5.06177s / 2891 = 1.75087ms
```

Before:

After:


esteldunedain:
In frames that'd be:

Before:
60 fps -> 49 fps

After:
60 fps -> 54 fps

It's still not ideal, although it should be non-issue anywhere but in briefing.

jonpas:
Very nice! :+1:

esteldunedain:
It does use the new select syntax, so it require 1.56 I believe.

commy2:
```
Before:
60 fps -> 49 fps

After:
60 fps -> 54 fps
```

That sounds really good. You have to keep in mind that the game always chokes when bullets are flying, regardless of scripts.

addons/nametags/functions/fnc_drawNameTagIcon.sqf
```diff
@@ -7,74 +7,75 @@
...
-} else {
-    ""
-};
+    //Set Color:
+    private _color = [1, 1, 1, _alpha];
+    if ((group _target) != (group _player)) then {
+        _color = +GVAR(defaultNametagColor); //Make a copy, then multiply both alpha values (allows client to decrease alpha in settings)
+        _color set [3, (_color select 3) * _alpha];
+    } else {
+        _color = [[1, 1, 1, _alpha], [1, 0, 0, _alpha], [0, 1, 0, _alpha], [0, 0, 1, _alpha], [1, 1, 0, _alpha]] select ((["MAIN", "RED", "GREEN", "BLUE", "YELLOW"] find (assignedTeam _target)) max 0);
```
jokoho48: why not using the param? @commy2 had some code snipped

esteldunedain: > why not using the param? @commy2 had some code snipped

Why **using** param? I genuinely don't know what you're refering too.

jokoho48: i remeber something like this
`_test = colors param[[sides]find team, [default]];`

and commy2 mean that was faster than select find +1

esteldunedain:
Ok, I think I tested positively all the different combinations of settings in local MP, and it seems bug free,

I also tried a few more optimization techniques, but they didn't improve things significatively (two levels of caching for the drawIcon3D params, plane culling of nearby units, worldToScreen checks of tags).

addons/nametags/functions/fnc_drawNameTagIcon.sqf
commy2: `_color = [[1, 1, 1, _alpha], [1, 0, 0, _alpha], [0, 1, 0, _alpha], [0, 0, 1, _alpha], [1, 1, 0, _alpha]] select ((["MAIN", "RED", "GREEN", "BLUE", "YELLOW"] find (assignedTeam _target)) max 0);`

can be written as:

`_color = [[1, 0, 0, _alpha], [0, 1, 0, _alpha], [0, 0, 1, _alpha], [1, 1, 0, _alpha]] param [["RED", "GREEN", "BLUE", "YELLOW"] find assignedTeam _target, [1, 1, 1, _alpha]];`

which is slightly faster and slightly more readable (and works when assignedTeam bugs out and returns nil iirc)

commy2: Even faster might be using a hashLocation. ;D

esteldunedain: > Even faster might be using a hashLocation. ;D

Yes, if anything I'd go that way. I don't think it would have a significant impact though, so I think it'll have to wait.

jokoho48:
did you tested it with UGV/UAVs?

esteldunedain:
> did you tested it with UGV/UAVs?

Yes, they seem to work alright

esteldunedain:
I added tags to the units inside the player weapon, as per requested #538 by @bux578 a long time ago. 

The downside to this is that I don't know of any way to check LOS inside the vehicle, so you can see the tags for all the crew.

A screenshot where it works great:


An one in which the flaw is visible:


I think it's worth it, but I can reverse it if you guys disagree.

esteldunedain:
Testing only against the GEOM LOD worked:

```sqf
{lineIntersectsSurfaces [_camPosASL, eyePos _x, ACE_player, _x, true, 1, "GEOM", "NONE"] isEqualTo []} &&
```
`

bux578:
> I added tags to the units inside the player weapon

There are units inside my weapon? Gross ....

SzwedzikPL:
@commy2 check that in console
`["yes","no"] param [-1, "none"]`

commy2:
@ SzwedzikPL That is fixed in 1.56 after I complained: http://feedback.arma3.com/view.php?id=26682

commy2:
@SzwedzikPL That is fixed in 1.56 after I complained:
http://feedback.arma3.com/view.php?id=26682

jokoho48:
@commy2 tested in dev Branch sam result, it return "yes"

commy2:
Same. F Great...
@jokoho48  dev branch as in 1.57 dev or 1.56 RC. I'm on the RC

jokoho48:
1.57

commy2:
Which means the reverted it on both branches, gg.

esteldunedain:
> Which means the reverted it on both branches, gg.

BIS_fnc_KK

bux578:
Status?

esteldunedain:
@bux578 , this is ready. I don't use that params syntax.

PR #3293, opened 2016-02-08
Slightly reduce the distance for which the LOS check is skipped. 

### LINKED ISSUE #3270: Interaction through barriers
Arma 3 1.54
ACE3 most current dev 3.5.0

image

esteldunedain:
Fix #3270

I don't dare reduce this much more. Sometimes units that fall unconscious might partially clip through objects; the threshold distance prevents those units from getting inaccessible.

Anyway, the reduction seems enough to avoid the bug in common cases.


Glowbal:
:+1: Looks good to me

PR #3296, opened 2016-02-09
experimental add ace settings to 3den editor

commy2:
I tried to add mission attributes for ace settings to the 3d editor. The goal is to replace the modules if possible.
Unfortunately it seems my menu is empty and the settings refuse to show up. I have no idea why that is. Maybe it's bugged? Anyone else has experience with this?

doc:
https://community.bistudio.com/wiki/Eden_Editor:_Configuring_Attributes#Scenario
https://community.bistudio.com/wiki/Eden_Editor:_Configuring_Menu_Bar#Scenario_Attributes


commy2:
Ways to access settings:

`ace_frag_enabled`
- undefined when setting never opened

`"ACE_Settings" get3DENMissionAttribute "ace_frag_enabled";`
- only functional inside 3den. seems to return default value always while inside mission?

`getMissionConfigValue "ace_frag_enabled"`
- requires mission to be saved once (reads from description.ext on disc?)

addons/common/Cfg3DEN.hpp
```diff
@@ -0,0 +1,27 @@
...
+};
+
+class ctrlMenuStrip;
+
+class Display3DEN {
+    class Controls {
+        class MenuStrip: ctrlMenuStrip {
+            class Items {
+                class Attributes {
+                    items[] += {ACE_Settings};
```
jonpas: Missing quotes?

commy2: Yes, although the binarizer does put them around this anyway.

jonpas: Yeah it does, it's safer to do it though.

Glowbal:
Discussed this on Slack with @commy2. 

An issue is that 3DEN settings do not mimic the same behaviour mission modules have. That means they cannot replace them as is. 3DEN settings will always initalize, while mission modules only excute if placed. This is a problem in for example: Using non default and non forced settings with the server config. 3DEN settings will overwrite this regardless.

For this reason, I suggest we hold off on this for now and rethink our approach for the ACE settings framework. We intended to move it to CBA at some point, perhaps we could combine an overhaul with a move to CBA?

esteldunedain:
An alternative would be to have an extra setting for each "module" that "Enable EDEN settings"; then settings would only be tranfered from eden to the settings framework if that's enabled. Eden settings would then work like current modules do. Of course that'd have the same limitations as the current approach (e.g. not all settings being available through module/eden), but at least everything would show on the same screen.

simon84:
Why not having a default option "server settings". Then the server settings are only overridden, when the mission designer set a different setting than "server settings".

esteldunedain:
> Why not having a default option "server settings". Then the server settings are only overridden, when the mission designer set a different setting than "server settings".

That'd be an all-or-nothing approach: either all the settings would be taken from the server or non of them. What I'm proposing is a toggle setting like that for each group of settings (groups being approx. the same as current modules).

commy2:
Like a `ace_frag_useServerSettings` checkbox ? If checked, then grey out all other options and use the values from the server.hpp ?

simon84:
No, you misunderstood me. For each setting there could be an selectable option "as server settings". For example: Medical System: (as server settings, advanced, basic), advanced ballistics: (as server settings, enabled, disabled)

As said, "as server settings" is the default option. Or, are these type of option lists not possible? Is there only a possibility of checkboxes.

esteldunedain:
> Like a ace_frag_useServerSettings checkbox ? If checked, then grey out all other options and use the values from the server.hpp ?

Yes, only I'd default it on (or reverse it's name and default off). Anyway I'd love all settings to be grayed out by default, as per current behaviour.

commy2:
Checkboxes can be set to default on.

commy2:
>Medical System: (as server settings, advanced, basic), advanced ballistics: (as server settings, enabled, disabled)

Sounds good too.

esteldunedain:
> Checkboxes can be set to default on.

Great.

There's an additional small catch. Server forced settings should be never overidden; I don't know how if we could word it somehow so people know that the settings they put on 3den might not take effect in that case.

esteldunedain:
> Sounds good too.

:+1:

commy2:
> perhaps we could combine an overhaul with a move to CBA?

I think this would be the best solution.

esteldunedain:
> perhaps we could combine an overhaul with a move to CBA?
:+1: 

I think we should make it so setting variable are still prefixes according to the mod that creates them (e.g. `ACE_`). I love being able to use `if (GVAR(mySetting) ...` and i'd hate if that changes.

commy2:
noted

Glowbal:
We are moving the settings framework to CBA

PR #3298, opened 2016-02-10
replace count +array deleteAt find with array select CODE

jokoho48:


esteldunedain:
Just a comment for other cases similar to this. Using select to filter the array means that each frame the entire array is copied, even if nothing is deleted. If the array is big and there are very ocassional deletions, then using select shoud be slower than just using find+deleteAt. 

I'm not sure if these arrays are big enough for this to matter, but measuring perf wouldn't be a bad idea.

--------------------

It occured to me that, as an optimization, if `GVAR(waitUntilAndExecArray)` was sorted and iterated backward then no deletions would be necessary except at the tail, which can be handled very very fast.

jokoho48:
its faster than create a copy in every frame and than acces the array every frame for every entry thats get deleted.

and to waitUntilAndExecArray this have code as condition that you cant sort

esteldunedain:
> its faster than create a copy in every frame

The point is that the current method doesn't do that.

> and to waitUntilAndExecArray this have code as condition that you cant sort

True, I was refering to `GVAR(waitAndExecArray)`

jokoho48:
of cause. the current method do this:
```c++
{
    array deleteAt (find _x);
} count +array
```

esteldunedain:
Wow, true. Sorry, I didn't see/expect that `+_array`. Yeah of course that pattern is awfull.

:+1: to replacing

jokoho48:
yes this is because else you fuck up the array/index/skiping entrys

SzwedzikPL:
Tested on 1000 element array

http://i.imgur.com/vU5BwvK.jpg
http://i.imgur.com/tKp8o4a.jpg

SzwedzikPL:
But this is interesting...

http://i.imgur.com/JQv0P3e.jpg
http://i.imgur.com/0tdYPEb.jpg

Looks like if `select` contains something more than simple condition it works slower than `forEach`
I think it `select` should be tested with code before merge this.

addons/medical/functions/fnc_litterCleanupLoop.sqf
```diff
@@ -13,16 +13,18 @@
...
+GVAR(allCreatedLitter) = GVAR(allCreatedLitter) select {
     _x params ["_time", "_objects"];
     if (CBA_missionTime - _time >= GVAR(litterCleanUpDelay)) then {
         {
             deleteVehicle _x;
         } forEach _objects;
+        false
+    } else {
+        true
         GVAR(allCreatedLitter) set[_forEachIndex, objNull];
```
commy2: missing ; Shouldn't the `true` be the return value?

esteldunedain:
I think that this PR makes things harder to read, and it has not been demostrated it's actually faster than the current code.

PabstMirror:
My attempt at a controlled benchmark between the two methods that would behave similar to tracking bullets fired:
```
_list = [];
for "_tick" from 0 to 10000 do {
    if ((_tick % 20) == 0) then {_list pushBack (_tick + 100);};
    {
        if (_tick > _x) then {
            _dummy = _x;
        } else {
            _list deleteAt (_list find _x);
        };
        nil
    } count +_list;
};
```
Old Way: 31.0303 ms

```
_list = []; 
for "_tick" from 0 to 10000 do { 
    if ((_tick % 20) == 0) then {_list pushBack (_tick + 100);}; 
    _list = _list select { 
        if (_tick > _x) then { 
            _dummy = _x; 
            true 
        } else { 
            false 
        }; 
    }; 
};
```
New Way: 32.3548 ms

esteldunedain:
May you try something like this @PabstMirror?:

```sqf
_list = [];
for "_tick" from 0 to 10000 do {
    if ((_tick % 20) == 0) then {_list pushBack (_tick + 100);};
    {
        if (_tick > _x) then {
            _dummy = _x;
        } else {
            _list set [_forEachIndex, null];
        };
    } forEach _list;
    _list = _list - [null];
};
```

PabstMirror:
42.1667 ms

commy2:
What about replacing `{a} else {b}` with `[{a},{b}]`. If this is really that time critical that could be considered.

commy2:
test this:
```
_list = [];
for "_tick" from 0 to 10000 do {
    if ((_tick % 20) == 0) then {_list pushBack (_tick + 100);};
    {
        if (_tick > _x) then [{
            _dummy = _x;
        }, {
            _list deleteAt (_list find _x);
            nil
        }];
    } count +_list;
};
```

esteldunedain:
> 42.1667 ms

:-1:

PabstMirror:
@commy2  31.4375 ms

commy2:
alright. so, discard this?

jokoho48:
yes discard this. this is why i never continue this PR

commy2:
yay for reversed Estonian flag

PR #3299, opened 2016-02-10
Manually pre-load for StaticWeapons and Car

### LINKED ISSUE #3168: AI units in static weapons do not take any damage
**ACE3 Version:** 3.4.2
3.3.2, 3.4.0, 3.4.1 versions were tested as well. The issue occurred on all of them.

**ARMA 3 Version:** 1.54.133741

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* None

**Description:**
AI units in statics weapons do not take any damage. This does not happen with all of them and fairly rare. But it is a game-breaking bug that ruins the whole mission.

The first report the players did said that they shot the AI unit in a static weapon around 15 times with MAR-10 rifle. After that they shot him with AP Titan rocket, that did not do a thing to him.

I was able to reproduce this behavior in the Editor. But this does not happen often. I saved a simple test mission that seem to have this problem consistently.

The test mission does not have any ACE3 modules in it, my multiplayer mission does, including advanced ballistics and medicine.

Disabling ACE3 mod and running the test mission solves the problem.

**Steps to reproduce:**
* Place static weapon with AI units in it.
* Check if they take any damage (Does not happen all the time).

* Multiplayer
* Singleplayer
* Editor

The issue was reported on my multiplayer server. I was able to reproduce it the Editor and Singleplayer.

**Video:**
* https://www.youtube.com/watch?v=pkBvbSTcAug

**RPT log file:**
* http://pastebin.com/Q7uDGHZW

**Singleplayer Mission:**
* https://drive.google.com/uc?export=download&id=0B4Z2uhvLliuqSGVkZWVIRDJfdGc

PabstMirror:
Fix #3168 - thanks to killzonekid for the idea

Not pretty but it does seem to fix the issue.  Possible this could cause some stutter when zeus is spawning new statics, but I didn't notice any real problems.

### DIFF
--- a/addons/medical/XEH_preInit.sqf
+++ b/addons/medical/XEH_preInit.sqf
@@ -125,4 +125,25 @@ call FUNC(parseConfigForInjuries);
 GVAR(HITPOINTS) = ["HitHead", "HitBody", "HitLeftArm", "HitRightArm", "HitLeftLeg", "HitRightLeg"];
 GVAR(SELECTIONS) = ["head", "body", "hand_l", "hand_r", "leg_l", "leg_r"];
 
+//Hack for #3168 (units in static weapons do not take any damage):
+//doing a manual pre-load with a small distance seems to fix the LOD problems with handle damage not returning full results
+GVAR(fixedStatics) = [];
+private _fixStatic = {
+    params ["_vehicle"];
+    private _vehType = typeOf _vehicle;
+    TRACE_2("",_vehicle,_vehType);
+    if (!(_vehType in GVAR(fixedStatics))) then {
+        GVAR(fixedStatics) pushBack _vehType;
+        TRACE_1("starting preload",_vehType);
+        [{
+            1 preloadObject (_this select 0);
+        }, {
+            TRACE_1("preload done",_this);
+        }, [_vehType]] call EFUNC(common,waitUntilAndExecute);
+    };
+};
+["StaticWeapon", "init", _fixStatic] call CBA_fnc_addClassEventHandler;
+["Car", "init", _fixStatic] call CBA_fnc_addClassEventHandler;
+
+
 ADDON = true;
--- a/addons/medical/XEH_preInit.sqf
+++ b/addons/medical/XEH_preInit.sqf
@@ -144,6 +144,16 @@ private _fixStatic = {
 };
 ["StaticWeapon", "init", _fixStatic] call CBA_fnc_addClassEventHandler;
 ["Car", "init", _fixStatic] call CBA_fnc_addClassEventHandler;
+addMissionEventHandler ["Loaded",{
+    {
+        TRACE_1("starting preload (save load)",_x);
+        [{
+            1 preloadObject (_this select 0);
+        }, {
+            TRACE_1("preload done",_this);
+        }, [_x]] call EFUNC(common,waitUntilAndExecute);
+    } forEach GVAR(fixedStatics);
+}];
 
 
 ADDON = true;


esteldunedain:
:+1: Does using `CBA_fnc_addClassEventHandler` means that this requires next CBA release?

SzwedzikPL:
AFAIK `CBA_fnc_addClassEventHandler ` is present in 2.2.0

commy2:
- `CBA_fnc_addClassEventHandler` is already in 2.2.0
- KK was preloading allUnits, we are preloading the static weapons. Is this correct?
- What about tanks with hatches?

This work around has major flaws unfortunately:
- What if the game garbage collects a model?
- What if the player flushes the graphics memory? (shift num+ f, l, u, s, h)
- What if this is a savegame where the init handler run already, but the model isn't loaded in the current session?

esteldunedain:
> What about tanks with hatches?

We could preload those too. And probably also helicopters too.

> What if the player flushes the graphics memory? (shift num+ f, l, u, s, h)

That is rare enough. I'm not sure if collisions are tested in the gpu either (they certainly were not in arma 2, I'm not sure if that changed with physX).

> What if the game garbage collects a model?
> What if this is a savegame where the init handler run already, but the model isn't loaded in the current session?

Those are valid points; maybe units should be preloaded at regular intervals after they are created.

PabstMirror:
Btw, I don't really like this "fix", I would prefer BIS fix their stuff, but we don't know when or even if that will happen.  Until then, I think this is an ok band-aid to fix a critical problem.

 - Shift+NumMinus+FLUSH had no effect
 - Added save game compatibility

>What if the game garbage collects a model?

I tried spawning a bunch of new stuff with zeus and the shooting the old models which worked fine. But this still could be a problem over a long enough time?

Tank hatches seemed to be working ok without this.

Glowbal:
I dislike this workaround. I understand the need for it however and I do not have suggestions for an alternative band-aid.

Any reason we are preloading `Car`?

esteldunedain:
> Any reason we are preloading Car?

In principle, every vehicle in which you can easily shoot passengers could exibit the issue. I think cars were included because the amount of glass and the overall low armor.

jonpas:
> I think cars were included because the amount of glass and the overall low armor.

Offroads, quadbikes...

PabstMirror:
It was for the .50 cal technicals which had the same problem as the static weapons.

commy2:
>Shift+NumMinus+FLUSH had no effect
>Added save game compatibility

sweet

>Tank hatches seemed to be working ok without this.

interesting

But why are we preoading the vehicles while KK suggested preloading the soldiers?

PabstMirror:
Preloading the gunner ("B_Soldier_F") had no effect, it seems to be the vehicles that are the problem

Glowbal:
Ok thanks @PabstMirror.

I dislike it but am ok with having this go in for 3.5.0. I accept this is an issue we need to resolve and waiting for BI to do it may not be the smartest approach.

esteldunedain:
> waiting for BI to do it may not be the smartest approach.

:SMILE:

commy2:
>waiting for BI to do it may not be the smartest approach.
:kappa:

looterz:
Eagerly awaiting this or another band-aid fix until BIS fixes it, it's been affecting our games pretty frequently.

Glowbal:
If no objections, I am merging this in today.

commy2:
:+1:
@looterz
You can copy and paste this fix into your init.sqf so you don't have to wait for our update. You just have to manually resolve the macros.

esteldunedain:
@looterz, on `init.sqf`:

```sqf
ace_medical_fixedStatics = [];
private _fixStatic = {
    params ["_vehicle"];
    private _vehType = typeOf _vehicle;
    if (!(_vehType in ace_medical_fixedStatics)) then {
        ace_medical_fixedStatics pushBack _vehType;
        [{
            1 preloadObject (_this select 0);
        }, {
        }, [_vehType]] call ace_common_fnc_waitUntilAndExecute;
    };
};
["StaticWeapon", "init", _fixStatic] call CBA_fnc_addClassEventHandler;
["Car", "init", _fixStatic] call CBA_fnc_addClassEventHandler;
addMissionEventHandler ["Loaded",{
    {
        [{
            1 preloadObject (_this select 0);
        }, {
        }, [_x]] call ace_common_fnc_waitUntilAndExecute;
    } forEach ace_medical_fixedStatics;
}];
```

looterz:
@esteldunedain Thanks!

PR #3305, opened 2016-02-11
Disable ACE_Laser for choppers (aka Changes to DAGR)

GitHawk:
As discussed with @Glowbal on slack, an intermediary change to fix the DAGR missile on both Comanche and Ka-60 Kasatka Helicopter.

[Demo](https://www.youtube.com/watch?v=nzmm5ExG_14)

Allow laserLock, set weaponLockSystem to Laser and move class ADDON to ACE_Hydra70_DAGR so it can be further tested on e.g. ACE_Comanche.

### DIFF
--- a/addons/laser_selfdesignate/CfgVehicles.hpp
+++ b/addons/laser_selfdesignate/CfgVehicles.hpp
@@ -17,18 +17,18 @@ class CfgVehicles {
     class B_Heli_Attack_01_F: Heli_Attack_01_base_F {
         class Turrets: Turrets {
             class MainTurret: MainTurret {
-                GVAR(Enabled) = 1;      // Enable laser self-designation
+                GVAR(Enabled) = 0;      // Enable laser self-designation
             };
         };
     };
-    
+
     class Plane: Air {};
     class Plane_Base_F: Plane {
         class Turrets {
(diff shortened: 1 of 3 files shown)


esteldunedain:
Can you elaborate a bit more about why this change is needed?

Are we ditching some parts of ace_lasers and reverting to vanilla?

GitHawk:
Sure. Partially yes.
I moved the config values for ace_laser from DAGR to ACE_Hydra70_DAGR since they aren't working correctly.

DAGR with ace_laser only works on the Comanche with it's self designation laser. Other lasers (drones, laserpointer) are not supported.
As long as no one fixes that, this intermediary PR allows vanilla locking on all laserTargets. But no vanilla locking on IR/radar targets is enabled.

esteldunedain:
Ok, we've been talking with @GitHawk. The plan for this PR is to disable ACE_Laser and ACE_SelfDesignate usage for the moment, except for the Javelin.

The reason is that the laser code is not feature complete, as drones, laserpointers, etc are not being supported.

We've talked about cleaning up the code and trying to reenable it post 3.5, but hopefully with an approach that makes it work in all cases and be toggable on/off. Also laser difraction would be optional for performance reasons.

Issue #967, opened 2015-05-01
More realistic laser codes


gienkov:
ACE3 Version: 3.0.0.1

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
* *None*

**Description:**
*I think in reality there is a range of allowable laser codes that can be set inside SOFLAM and GBU-12 or other laser guided bombs. This is how it works in DCS: A-10C:*
* *Codes from 1111 to 1688 are valid. The first digit must be 1, the second 1-6, and the third and forth 1-8. - DCS: A-10C Manual*
* *Reading the manual with some rigor it says that the L and LSS codes for the TGP have to be of the form ABCC (A=1, B=1-7, C=1-8). That's not to say that the GBU-10/12 valid codes have to be of this same set as the TGP is capable. Valid LGB codes seem to be of the form ABBB (A=1, B=1-7). It does appear that the code 1688 is invalid according to what's possible to set in the INV page. If JTAC is lasing 1688 then one cannot drop an LGB directly and would have to LSS for it and then self-designate. Perhaps this is intentional procedure to use an "unbombable" code? - Post on DCS forum*

*Maybe it is worth to pre-enter all valid combinations and just make player switch between them using Laser Code Up/Down?*
*Right now it will feel pretty basic way its implemented.*

**Steps to reproduce:**
* *Grab SOFLAM, you can set any laser code*

**Where did the issue occur?**
*"Multiplayer", "Singleplayer"*

**RPT log file:**
*Not needed*

Ocelottr:
Relevant information from the US Army 2005 AH-64D Operators' manual.

CODE RANGE. 

The CODE RANGE indicates
the laser code ranges which are supported by the laser
keywords resident on the DTC. 

The helicopter is capable
of supporting the Tri -Service pulse repetition frequency
(PRF) laser codes and USAF, Hellfire and Copperhead
pulse interval modulation (PIM) laser codes. 

In order for
the LRFD, [ MT LST] and Hellfire subsystem to use a
PIM laser code the appropriate keyword, for the specific
code range, is required:

1111 - 1788 Tri -Service . . . . . . PRF
2111 - 2888 USAF . . . . . . . . . . . PIM
4111 - 4288 Hellfire-A . . . . . . . PIM
4311 - 4488 Hellfire-B . . . . . . . . PIM
4511 - 4688 Hellfire-C . . . . . . . PIM
4711 - 4888 Hellfire-D . . . . . . . PIM
[...]

PR #3043, opened 2015-12-16
Mortar interaction and ammunition handling

VKing6:
This is a port of code originally written by Grey and Glowbal at [16AA](https://github.com/16AAModTeam/LSR/tree/master/addons/staticweapons).

Goals of this branch is to remove the multiple-round magazines for the mortars and restricting them to loading 1 round at a time and to give assistant gunners and ammo carriers something to do. The plan is to make this module-controlled.

Current status:
- [x] Port major script components to ACE
- [x] Add interaction menu support for loading rounds
- [x] Add new mortar magazines and make them carriable by infantry
- [x] Port proper inventory images and ground models for new magazines
- [x] Implement options in module to avoid changing existing missions 
 - [x] Add module settings
 - [x] Don't show loading/unloading actions if ammunition handling is not set
 - [x] Init mortar to replace weapon and remove vanilla magazines if ammunition handling is set
- [x] Zeus testing
- [x] Unload rounds to player inventory
- [x] Update module documentation and class list
- [x] Ammo boxes
- [ ] Translations
- [x] MP testing
- ~~Make the mortar reloadable by the gunner with a time penalty~~ Eh too complicated for now; the time penalty will be having to get out and back in between shots.

Issue #380, opened 2015-04-07
Guidance Modularity Framework


jaynus:
Laser tasks pulled from PR #3308
- [x] Reenable laser marker creation for the laser_selfdesignate (Comanche).
- [ ] Prefix laser codes differently for different factions.
- [ ] Make faction specific derivatives of laser designators.
- [ ] Make laser codes indicator fall back to a hint in every case in which there's no configured way to display the code in the hud.
- [ ] Guidance framework modularity for addon makers
  - [x] Laser code and designation framework
  - [x] Configuration reading for firing mechanics
  - [x] Configurable guidance functions from parameters
  - [x] Seeker/Guidance separation
  - [ ] Locking modes
- [ ] Attack Profiles
  - [x] Linear attack (DAGR missile)
  - [x] Top-down
  - [ ] HI
  - [ ] MED
  - [x] DIR
- [ ] Seeker Profiles
  - [x] SALH
  - [ ] SARH
  - [ ] LIDAR
  - [ ] Thermal
  - [x] Optic
    - [ ] SPIKE-mode live target feed
  - [ ] GPS
  - [ ] MCLOS
  - [ ] SACLOS

Documentation:

```
class CfgAmmo {
    class ACE_ExampleMissile {

       // Begin ACE guidance Configs
        class ACE_MissileGuidance {
            enabled = 1;

            minDeflection = 0.005;      // Minium flap deflection for guidance
            maxDeflection = 0.025;       // Maximum flap deflection for guidance
            incDeflection = 0.005;      // The incrmeent in which deflection adjusts.
            //minDeflection = 0.005; 
            //maxDeflection = 0.5;
            //incDeflection = 0.005;

            // Guidance type for munitions
            defaultSeekerType = "SALH";
            seekerTypes[] = { "SALH", "LIDAR", "SARH", "Optic", "Thermal", "GPS", "SACLOS", "MCLOS" };  

            defaultSeekerLockMode = "LOAL";
            seekerLockModes[] = { "LOAL", "LOBL" }; 

            seekerAngle = 90;           // Angle in front of the missile which can be searched
            seekerAccuracy = 1;         // seeker accuracy multiplier
            seekerIntercept = 1;        // enables seeker intercept calculations (lead target)
            seekerMinRange = 1;
            seekerMaxRange = 2500;      // Range from the missile which the seeker can visually search

            // Attack profile type selection
            defaultAttackProfile = "LIN";
[...]

PR #3308, opened 2016-02-12
Laser guidance for all designators

esteldunedain:
Tasks:
- [x] Enable ace_laser to interact with vanilla lasers.
- [x] Make laser dispersion optional, default off.
- [x] Rewrite `fnc_shootRay` using `lineIntersectsSurfaces`.
- [x] Deprecate laser_selfdesignate in favor of the laser_designate?


PabstMirror:
@esteldunedain did some work on this and I think we can grab it for 3.8

Fix #4420 (someone else is trying to use the framework, so I'd like to get this merged even if we don't have some of the other stuff done)
Fix #1959
Fix #967

I merged laser_designate into laser
lase_selfDesignate really isn't nessacary so was dropped
rewrote big chunk of laser
using `lineIntersetsSurfaces` was much faster for shootRay

If anyone wants to test it out, there is a hidden vehicle that you can spawn in `ACE_Comanche_Test`

esteldunedain:
:+1: @PabstMirror, I'm glad to see this being rescued from oblivion.

PabstMirror:
This seems to be working in mp
Moved some of the laser todos to #380

We still don't have any vehicles that are using this right now besides the `ACE_Comanche_Test`
I'm not sure if we want to re-add laser/dagr weapons to the main comanche, we could make it optional (3den attribute?)

esteldunedain:
> I'm not sure if we want to re-add laser/dagr weapons to the main comanche, we could make it optional 

If it's possible to add them to the main Comanche as optional, then :+1: for that. But if not, I think we could keep it sepparate for now.

PabstMirror:
I'm thinking it would work similar to #3043
Using something like `removeWeaponTurret` and `addWeaponTurret` to just swap launchers/ ammo.

esteldunedain:
> Using something like removeWeaponTurret and addWeaponTurret to just swap launchers/ ammo.

@PabstMirror, great idea; that sounds perfect.

Issue #3309, opened 2016-02-12
Drop FPS in Vehicle


Kartaviy:
ACE3 Version: 3.4.2 stable
**Mods:**

@CBA_A3 (last one: 2.2.0.151130)
@ace
@ RHS
**Placed ACE3 Modules:**

ACE Medical
**Description:**
When use Slef interaction menu in vehicle  FPS DROP
https://youtu.be/II0S2kySCDc
**Where did the issue occur? **
Multiplayer

esteldunedain:
That is a known issue. In vehicles with many passengers the interaction menu for all passengers is constructed at all times, which is quite slow when the advanced medsys is used. We did many round of optimization for this problem and it improved greatly, but it's still not perfect. I hope we can improve it further eventually.

Glowbal:
Try the medical menu and disable 3d medical interaction. That should significantly improve your fps.

PabstMirror:
Ref #1690

esteldunedain:
Closing in favor a that

PR #3312, opened 2016-02-12
Only get passenger actions for selected unit

### LINKED ISSUE #1690: Passenger interaction while in a crowded vehicle is choppy
In a loaded transport vehicle (10+ crew) interacting with the vehicle's passengers has very noticeable pauses.  It can make selecting things difficult.

collectActiveActionTree seems to be culprit, it can take 100+ms every second.

image

PabstMirror:
Fix #1690 #3309

When interacting with vehicle passangers we collected all actions for all crew which was very expensive for large transports (>100ms stutter each second). This only collects the passenger actions for the selected crew member.

ace_interact_menu_fnc_collectActiveActionTree_counter: Average: 8.04395s / 1326 = 6.06632ms
vs
ace_interact_menu_fnc_collectActiveActionTree_counter: Average: 0.46582s / 1175 = 0.396443ms

### DIFF
--- a/addons/interaction/functions/fnc_addPassengerActions.sqf
+++ b/addons/interaction/functions/fnc_addPassengerActions.sqf
@@ -20,12 +20,10 @@
 params ["", "", "_parameters"];
 _parameters params ["_unit"];
 
-private ["_varName", "_actionTrees", "_actions"];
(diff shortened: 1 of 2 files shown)


jonpas:
:+1:

PR #3316, opened 2016-02-14
Make bodyBags vehicles that can carry respawning players

### LINKED ISSUE #1301: Player corpses cannot be successfully bodybagged while the player is respawning
ACE3 Version: 3.0.0

Mods:
@cba_a3 RC6 hotfix 1
@ace3

Placed ACE3 Modules:
N/A

Description:
When a player dies their body cannot successfully be placed in a bodybag untill that player spawns again.
Instead, a bodybag is spawned, but the corpse remains.

Steps to reproduce:
Connect two players to the server
Have one player die
Have the other played attempt to place the corpse of the first player in a bodybag
(Note how the corpse can be successfully bagged after the player has spawned again)

Repro mission:
http://a.pomf.se/ciyros.pbo
respawn = 3;
respawnDelay = 30;
respawnTemplates[] = {"Counter", "Spectator"};
respawnOnStart = 0;
respawnVehicleDelay = 240;
respawnDialog = 0;

Where did the issue occur?
Multiplayer, dedicated AND listen servers

RPT log file:
N/A

PabstMirror:
Fix #1301

Can bag re-spawning players
Can still access inventory of body
Can still track original body object

Dead bodies are odd, but this seems to be working fine. 
 - `setPos` on dead bodies seems to need to be called where local
 - `moveInX` works on bodies, but `moveOut` doesn't, so there still is no way to get a body out of a bag


esteldunedain:
Morbid bycicle :heart:

commy2:
The class name `ACE_bodyBagObject` is infuriating.

Glowbal:
This is an interesting approach. :+1: I didn't notice anything odd in the code. Has this been tested in MP?

PabstMirror:
Just very small scale local dedi.
I'm not sure about the effects of making these vehicle simulations.
There may be negative effects on something like a persistant mission with many bags and possibly having AI react to the object being a `LandVehicle`.

An alternate fix would be https://github.com/acemod/ACE3/compare/playerBodybags
which just moves the locked player bodies to a corner, hides them and then deletes when free.  I actually feel that would be safer, but this would offer more features (like not deleting the body object and having access to inventory)

Issue #1811, opened 2015-07-06
Bug in vehicle crashes (ace_medical_enableVehicleCrashes)


SzwedzikPL:
ACE3 Version: 3.1.2.0 (latest master https://github.com/acemod/ACE3/commit/86a8fb0b647afba724d6dde3e7cad08a0fdfb1f9)

Arma 3 Version: 1.46 Stable 

**Mods:**
* @cba_a3
* @ace3

**Placed ACE3 Modules:**
http://pastebin.com/AjbtdiJR

**Description:**
https://www.youtube.com/watch?v=YVXw3AL1tU4

**Steps to reproduce:**
*video*

**Where did the issue occur?**
*Multiplayer, Singleplayer*

**RPT log file:**
*n/a*

Glowbal:
So.. what is the bug? What behaviour do you expect?

SzwedzikPL:
So auto eject **after** plane crushes and pilot have **no injures** is by design?

bux578:
Can't reproduce. I'm unconcious while and after ejecting.

SzwedzikPL:
Like i told @bux578 on slack:

I think we should think about is ejecting is actually needed in 100% situations. Now i always see pilot "in the air" when some plane crash.  This vehicle crashes ale awesome for helis and land vehicles couse it push players in coop to rescue missions but what chances jet pilot have in any kind of crash?

Thats first. Second is that in my community after testing this crashes we saw that in 100% of crashes pilot and crew survives crash and some part of crew dies in flames after "eject". I think it should be changed to random chances to survive crash depend on kind of crash and medical module should eject them and "lay" in distance that is safe from fire (to actually let them have chance to survive and maybie been rescued)

bux578:
Closing in favor of https://github.com/acemod/ACE3/issues/3134

PR #3323, opened 2016-02-16
Fix vehicle crashes

### LINKED ISSUE #3269: Ace 3 Revive bug with destroyed vehicles
Description:It happened alot of times that while driving a vehicle,the vehicle gets shot and destroyed tho with ace the crew can survive and/or jump out while some crew dies too.When the module "Revive" is on as you know it moves the player to an unconcious state instead of killing him.Usually the driver is the one that dies from the explosion,atleast in my case where i used the RG33 from RHS and some default vehicles.When the vehicle was destroyed the driver was moved out of the vehicle by default and is set to unconcious.This is where the issue occurs.Its a small issue but in moments like getting shot at it means alot.When the player is moved out of the destroyed vehicle by default,not by another player,he is standing up holding his gun in his hands and looking normal even tho he is fairly injured and unconcios.This makes others not notice that the player is down and makes the player wait for the bleedout timer.I hope this is possible to fix.

Modules placed:  Ace 3 Revive "On"

Mods:
@cba_a3
@ace 
@RHS:USAF

Where did the issue happen:

Anywhere,singleplayer,multiplayer,editor

How to replicate:

Place the ace revive module,drive a vehicle over a mine or something that will destroy it(Might need more people)

RPT File: Not needed

### LINKED ISSUE #3306: Only the driver dies when the vehicle is destroyed.
Same to this report: https://github.com/acemod/ACE3/issues/3269 but described according to the rules.

ACE3 Version: 3.4.2 stable
**Mods:**
* @CBA_A3 (last one: 2.2.0.151130)
* @ace

**Placed ACE3 Modules:**
* None

**Description:**
 Only the driver dies when the vehicle is destroyed.

**Steps to reproduce:**
* Place helicopter, place few guys in it, 
* Let the pilot fly at a height of 200 m and then let helicopter swoops down and hit the ground. 
* Pilot is dead, other guys in vehicle is alive (and a bit injured). They can eject the vehicle or they are able to sit in a burning vehicle for a long time :helicopter: :fire: :smile: 

Multiplayer, 
can't test it in singleplayer and editor easely (bot don't want to crash the helicopter) but i think it's also true for these modes.

RPT log file:
Multiplayer: https://yadi.sk/d/0BieuLlpogx2M
Multiplayer (zeus): https://yadi.sk/d/bzgGtP5uoh5fe

Video:
https://youtu.be/NVTiBcx4kCU

PabstMirror:
Fix #3306
Fix #3269

Enemy RPG to MRAP (driver):
```
HD: [carrd,"",1,<NULL-object>,"",-1]
HD: [carrd,"",1,<NULL-object>,"",-1]
HD: [carrd,"",1,O Alpha 1-1:1,"R_PG32V_F",-1]
```
Enemy RPG to MRAP (gunner):
```
HD: [carrg,"",1,O Alpha 1-1:1 REMOTE,"R_PG32V_F",-1]
```

35m fall veh explodes (driver):
```
HD: [carrd,"",1,<NULL-object>,"",-1]
HD: [carrd,"",1,<NULL-object>,"",-1]
HD: [carrd,"",1,carrd,"",-1]
```

35m fall veh explodes (gunner):
```
HD: [carrg,"",1,carrd,"",-1]
```

So in crashes, the shooter/source can be the pilot for non-drivers.
Switching the order in determine if fatal will handle RPG shots that blow up vehicles as there doesn't seem to be any hit point damage.

### DIFF
--- a/addons/medical/functions/fnc_handleDamage.sqf
+++ b/addons/medical/functions/fnc_handleDamage.sqf
@@ -26,6 +26,8 @@ if !(local _unit) exitWith {
     nil
 };
 
+// if (alive _unit) then {diag_log text format ["HD: %1", _this];};
+
 private ["_damageReturn",  "_typeOfDamage", "_minLethalDamage", "_newDamage", "_typeIndex", "_preventDeath"];
 
 // bug, assumed fixed, @todo excessive testing, if nothing happens remove
@@ -78,14 +80,14 @@ _minLethalDamage = if (_typeIndex >= 0) then {
 };
 
 private _vehicle = vehicle _unit;
+private _effectiveSelectionName = _selection;
 if ((_vehicle != _unit) && {!(_vehicle isKindOf "StaticWeapon")} && {_shooter in [objNull, driver _vehicle, _vehicle]} && {_projectile == ""} && {_selection == ""}) then {
     if (GVAR(enableVehicleCrashes)) then {
-        _selection = _this select 1; //pull random selection from HDC
-        TRACE_1("Veh Crash",_selection);
+        _effectiveSelectionName = _this select 1; //pull random selection from HDC
     };
 };
 
-if ((_minLethalDamage <= _newDamage) && {[_unit, [_selection] call FUNC(selectionNameToNumber), _newDamage] call FUNC(determineIfFatal)}) then {
+if ((_minLethalDamage <= _newDamage) && {[_unit, [_effectiveSelectionName] call FUNC(selectionNameToNumber), _newDamage] call FUNC(determineIfFatal)}) then {
     if ((_unit getVariable [QGVAR(preventInstaDeath), GVAR(preventInstaDeath)])) exitwith {
         _damageReturn = 0.9;
     };
@@ -101,13 +103,8 @@ if ((_minLethalDamage <= _newDamage) && {[_unit, [_selection] call FUNC(selectio
 [_unit] call FUNC(addToInjuredCollection);
 
 if (_unit getVariable [QGVAR(preventInstaDeath), GVAR(preventInstaDeath)]) exitWith {
-    if (vehicle _unit != _unit and {damage (vehicle _unit) >= 1}) then {
-        [_unit] call EFUNC(common,unloadPerson);
-    };
-
-    private "_delayedUnconsicous";
-    _delayedUnconsicous = false;
-    if (vehicle _unit != _unit and {damage (vehicle _unit) >= 1}) then {
+    private _delayedUnconsicous = false;
+    if (_vehicle != _unit and {damage _vehicle >= 1}) then {
         [_unit] call EFUNC(common,unloadPerson);
         _delayedUnconsicous = true;
     };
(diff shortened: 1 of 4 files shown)


SzwedzikPL:
Is this also fixes https://github.com/acemod/ACE3/issues/1811 ?

PabstMirror:
@SzwedzikPL I believe so, I couldn't reproduce that issue directly, but now in-general a vehicle explosion should almost always be fatal.

addons/medical/functions/fnc_handleDamage.sqf
```diff
@@ -77,13 +79,15 @@ _minLethalDamage = if (_typeIndex >= 0) then {
     0.01
 };
 
-if (vehicle _unit != _unit && {!(vehicle _unit isKindOf "StaticWeapon")} && {isNull _shooter} && {_projectile == ""} && {_selection == ""}) then {
+private _vehicle = vehicle _unit;
+private _effectiveSelectionName = _selection;
+if ((_vehicle != _unit) && {!(_vehicle isKindOf "StaticWeapon")} && {_shooter in [objNull, driver _vehicle, _vehicle]} && {_projectile == ""} && {_selection == ""}) then {
     if (GVAR(enableVehicleCrashes)) then {
-        _selection = GVAR(SELECTIONS) select (floor(random(count GVAR(SELECTIONS))));
+        _effectiveSelectionName = _this select 1; //pull random selection from HDC
```
Glowbal: Why are we doing this? Isn't selection `""` here?

addons/medical/functions/fnc_handleDamage.sqf
```diff
@@ -26,6 +26,8 @@ if !(local _unit) exitWith {
     nil
 };
 
+// if (alive _unit) then {diag_log text format ["HD: %1", _this];};
```
Glowbal: dead debug code. Please remove.

Glowbal:
:+1:

addons/medical/functions/fnc_handleDamage.sqf
PabstMirror: On crash, handle damage caching will convert the `""` into a random selection,
I think we want the same value for passing to `FUNC(determineIfFatal)`
But we don't want to overwrite `_selection` for the `_selection in ["", "head", "body"]` checks on instantDeath/revive.

ivankrouglyi:
Works great!

ivankrouglyi:
(RHS) something strange become with AGS damage to troops and hellfire from apache to t72 tank with this change

SzwedzikPL:
This problem is not fixed. Today i had a situation, my helicopter was shot down. After helicopter explosion i was still alive in vehicle. As soon as the fire was over I got out of the vehicle without any wounds.

ivankrouglyi:
@SzwedzikPL 
did you test it on 3.5.0 version ?

SzwedzikPL:
@ivankrouglyi Yes

jonpas:
Create a new issue with app the required information please, don't reopen old closed PRs that were already released, thank you.

PR #3331, opened 2016-02-17
Fixed c4996 compiler warning/error due to use of strncpy and ace_clipboard memory leak

looterz:
Noticed the funny comment and thought I would submit a PR, strncpy_s with the _TRUNCATE flag will also ensure that the string is null-terminated whether it fits or not. Looks like ballistics and some other extensions make heavy use of strncpy as well but don't disable the warning, which on some compilers might cause an error.


PabstMirror:
Looks good, extensions parse_imagepath and fcs both have the same problem

looterz:
ace_clipboard is all over the place.

looterz:
Okay, patched every module using strcpy, strncpy or in clipboards case memcpy for strings.

looterz:
Fixed some issues I spotted with the ace_clipboard implementation, such as using memory semantics meant for GMEM_MOVEABLE, and the heap memory not being cleaned up on failure and being allocated before the clipboard was ready to be used.

extensions/clipboard/ace_clipboard.cpp
```diff
@@ -30,44 +33,49 @@ void __stdcall RVExtension(char *output, int outputSize, const char *function) {
...
-            result = "GlobalLock() failed, GetLastError=" + GetLastError();
-            gClipboardData = "";
-            EXTENSION_RETURN();
-        }
-        memcpy(pClipboardData, gClipboardData.c_str(), gClipboardData.length());
-        pClipboardData[gClipboardData.length() + 1] = 0x00;
-
-        GlobalUnlock(hClipboardData);
+    #ifndef _WIN32
+        EXTENSION_RETURN();
```
esteldunedain: Wouldn't the extension always return here before running the `strcmp` check?

looterz: Only if the pre-processor definition _WIN32 is undefined, so basically only if compiled/used on linux.

esteldunedain: Yeah, sorry, I missed the `n`

Glowbal:
This looks good, just needs some formatting changes and the merge conflict resolved. Once that is done, we will merge it :+1: Thanks @looterz 

I wil see if I can get around to do that later today.

Glowbal:
#3905

PR #3333, opened 2016-02-18
Replace execRemoteFnc in ace_medical

esteldunedain:
I hope it's right this time!


esteldunedain:
The medical event names are currently a bit inconsistent. Maybe the new one should get prefixed with `medical_` like the some of the former are. At the same time keeping the event name and the target funcion name similar really helps I think.

PR #3337, opened 2016-02-18
Substantially increase overheating when using silencers or flash suppresors

esteldunedain:

1. Substantially increase overheating when using silencers or flash suppresors. The calculations are based on how much they supress sounds and flash. For reference typical Arma silencers get x2.1 overheating and ACE flash suppresors get x1.5.

2. Optimize overheating calculations for the player by caching bullet mass and silencer parameters.

3. Replace `createLocation` by `CBA_fnc_createNamespace`


addons/overheating/XEH_postInit.sqf
```diff
@@ -13,8 +13,9 @@ if (isServer) then {
 
 if !(hasInterface) exitWith {};
 
-GVAR(cacheWeaponData) = createLocation ["ACE_HashLocation", [-10000,-10000,-10000], 0, 0];
-GVAR(cacheWeaponData) setText QGVAR(cacheWeaponData);
+GVAR(cacheWeaponData) = call CBA_fnc_createNamespace();
```
jokoho48: syntax issue?

esteldunedain: Yup, it was.

Issue #2947, opened 2015-12-05
Interaction menu flashes on and off on vehicles


gienkov:
ACE3 Version: 3.4.0

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed ACE3 Modules:**
* *-*

**Description:**
* On dedicated trying to interact with vehicles is very hard right now. Interaction menu will basically appear and disappear randomly, making it ultra hard to select any interaction with vehicles. Did not tested with other players but from what I heard from my community it seems to happen on unit's too (need to confirm that myself.*
* I'm gonna do video with it later.*

**Steps to reproduce:**
* *Try to interact with vehicles on dedicated.*

**Where did the issue occur?**
*"Multiplayer"*

**RPT log file:**
* -.*

jonpas:
It seems to be distance related, going closer seems to stop the flickering. It might be related to a lot of custom checks in refuel and rearm, just a hunch.

jonpas:
Anyone else experiencing this? Should be moved to 3.4.1 if confirmed.

PabstMirror:
I can't reproduce on dedicated.
What vehicles does it happen on.
Any specific interaction or just the main point?
Also could you post your settings?

jonpas:
It happened on refuel nozzle for us, I will try to reproduce it later today.

GitHawk:
@PabstMirror It happens as soon as an additional interact point gets close to the main interact point (e.g. in refuel)

PabstMirror:
Should be solved via #4040

PR #3342, opened 2016-02-18
Uniformize interaction distance for vehicle based on bounding box

### LINKED ISSUE #3335: Vehicle interaction distance is too big
Arma 3 1.56 stable
ACE3 3.4.2 stable

Unable to open the "cargo" at that distance, though you should be much closer before you should be able to access the interaction point.

esteldunedain:

1. Uniformize interaction distance for vehicle. Distance would ne aprox 4m from the *bounding box* of the vehicle, consistent with interaction with people. Calculations are only done for large vehicles.
This works by moving the main interact point of large vehicles (cars, tanks, boat, helicopters) to the place closest to the player inside the bounding box. The movement is done in the line between the camera pos and the original action point, so it has no visual consequences. Main action is still kept a aprox. 1m inside the vehicle, so it can be blocked by other action points (like wheel repair points).

2. Increased distance to 4m for Static Weapons for consistency.

3. Close #3335

### Known issues
- `boundingBoxReal` is not very precise for some vanilla vehicles. An easy fix would be constructing a more precise box with `lineIntersectsSurfaces` the first time and caching it.


Glowbal:
@esteldunedain what do you want to do with this? What is the status on it?

esteldunedain:
I've three outstanding issues:

1. I'm a little worried about the eventual performance impact of this PR. I remember it being fine when I tested it, but some more testing would be great.

2. Moving around the main action point can potentially make it oclude other action points for some vehicles in corner cases. More testing would also be great.

3. The commented code should probably be removed. My idea at one point was to switch between the two algorithms depending on the IM mode.

I know @PabstMirror has been doing some testing, so he may be able to tell if 1 and 2 are really things to be worried about or not.

PabstMirror:
I think this is good to merge.
1. It is a little more code each frame for every vehicle in range; but it's not that expensive and most of the time there will be a max of 1 vehicle in range.

2. The occlusion problem should be ok. Repair adds many base interaction points, so far I'm not seeing any problems.

3. The commented out code did have occlusion problems.  Basic problem is you look at a car's wheel and it might move the main interaction point on top of the "remove wheel" action. My PR #2296 had the exact same problem. I think the proper way to fix this is going to be to actually check conditions of actions and check if it is safe to move the interaction point. But it's commented out, so it's fine to leave as is for now and we can take a look at this after 3.6.

I am also thinking this may help issue #2947

esteldunedain:
Ok, thanks @PabstMirror. :+1: for merging on my part

PR #3354, opened 2016-02-19
Fix bagging respawning players (alt)

### LINKED ISSUE #1301: Player corpses cannot be successfully bodybagged while the player is respawning

PabstMirror:
Fix #1301 
Alternate fix instead of PR #3316
Much simpler, just moves the body out of the way, then deletes after player has re-spawned.

### DIFF
--- a/addons/medical/XEH_postInit.sqf
+++ b/addons/medical/XEH_postInit.sqf
@@ -11,7 +11,6 @@ GVAR(heartBeatSounds_Slow) = ["ACE_heartbeat_slow_1", "ACE_heartbeat_slow_2"];
 ["interactMenuClosed", {[objNull, false] call FUNC(displayPatientInformation); }] call EFUNC(common,addEventHandler);
 
 //Treatment EventHandlers:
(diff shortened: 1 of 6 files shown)


Glowbal:
I prefer this solution over the other one for now. We will leave #3316 open and consider it for a future version.

PR #3359, opened 2016-02-20
integrate captivity modules with 3DEN

gienkov:

1. Add captivity options to units in 3DEN:
http://images.akamai.steamusercontent.com/ugc/312242432482819918/B1DCA029C2317558D825D86339700B0DC50AF6E1/
This require some stringtabling still

WIP

---

### DIFF
--- a/addons/captives/CfgEden.hpp
+++ b/addons/captives/CfgEden.hpp
@@ -1,0 +1,30 @@
+class Cfg3DEN {
+    class Object {
+        class AttributeCategories {
+            class ace_attributes {
+                class Attributes {
+					class ace_isHandcuffed {
+                        property = QUOTE(ace_isHandcuffed);
+                        control = "Checkbox";
+                        displayName = CSTRING(ModuleHandcuffed_DisplayName);
+                        tooltip = CSTRING(ModuleHandcuffed_Description);
+						expression = QUOTE([ARR_2(_this,_value)]call FUNC(setHandcuffed));
+                        typeName = "BOOL";
+                        condition = "objectBrain";
+                        defaultValue = false;
+                    };
+					class ace_isSurrendered {
+                        property = QUOTE(ace_isSurrendered);
+                        control = "Checkbox";
+                        displayName = CSTRING(ModuleSurrender_DisplayName);
+                        tooltip = CSTRING(ModuleSurrender_Description);
+                        expression = QUOTE([ARR_2(_this,_value)]call FUNC(setSurrendered));
+                        typeName = "BOOL";
+                        condition = "objectBrain";
+                        defaultValue = false;
+                    };
+                };
+            };
+        };
+    };
+};
--- a/addons/captives/CfgEden.hpp
+++ b/addons/captives/CfgEden.hpp
@@ -3,17 +3,17 @@ class Cfg3DEN {
         class AttributeCategories {
             class ace_attributes {
                 class Attributes {
-					class ace_isHandcuffed {
+                    class ace_isHandcuffed {
                         property = QUOTE(ace_isHandcuffed);
                         control = "Checkbox";
                         displayName = CSTRING(ModuleHandcuffed_DisplayName);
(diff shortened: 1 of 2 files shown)


addons/captives/CfgEden.hpp
```diff
@@ -0,0 +1,30 @@
+class Cfg3DEN {
+    class Object {
+        class AttributeCategories {
+            class ace_attributes {
+                class Attributes {
+					class ace_isHandcuffed {
```
jonpas: Shouldn't we be using GVARs here?

esteldunedain: Yes

addons/captives/CfgEden.hpp
```diff
@@ -0,0 +1,30 @@
...
+                        property = QUOTE(ace_isHandcuffed);
+                        control = "Checkbox";
+                        displayName = CSTRING(ModuleHandcuffed_DisplayName);
+                        tooltip = CSTRING(ModuleHandcuffed_Description);
+						expression = QUOTE([ARR_2(_this,_value)]call FUNC(setHandcuffed));
```
esteldunedain: @commy2, could you briefly explain how this work?

Does `expression` get called on mission start? Are we aware about when is it called (e.g., before or after postinit)?

esteldunedain:
@gienkov apparently there's a mix of tabs and spaces, or at least the indentation is showing inconsistently.

gienkov:
From wiki:

>// Expression called when applying the attribute in Eden and at the scenario start
// Entity is passed as _this, value is passed as _value
// %s is replaced by attribute config name. It can be used only once in the expression
// In MP scenario, the expression is called only on server.
expression = "_this setVariable [%s,_value];";

> // Expression called when custom property is undefined yet (i.e., when setting the attribute for the first time)
// Entity is passed as _this
// Returned value is the default value
// Used when no value is returned, or when it's of other type than NUMBER, STRING or ARRAY
// Custom attributes of logic entities (e.g., modules) are saved always, even when they have default value

gienkov:
Can you suggest some other attribute name then?

esteldunedain:
Thanks @gienkov.

It's critical we ensure expression is run after postInit, otherwise the event calls that happen inside wouldn't work reliably. If we cannot guarantee that we'll have to wrap all the calls like this:

```sqf
["SettingsInitialized", {[_this, _value] FUNC(setHandcuffed)}] call EFUNC(common,addEventHandler);
```
(properly wrapped in macros)

esteldunedain:
> Another way to solve this is to run eden calls through the ace_common_fnc_runAfterSettingsInit function as that is guaranteed to happen after all postInit's have finished, and all event handlers should be installed.

PabstMirror:
IIRC `FUNC(setHandcuffed)` / `FUNC(setCaptive)` can't be called right at init because of an animation problem at time 0.  TBH this probably should be better handeled in those functions.

For the module I just used a small wait.
I don't see why the eden expersssion couldn't just call the module function which should handle the wait and then call the event cleanly.

https://github.com/acemod/ACE3/blob/master/addons/captives/functions/fnc_moduleSurrender.sqf

PabstMirror:
I guess a dummy call like `[objNull, [_this], true] call ace_captives_fnc_moduleHandcuffed` should work, but it also seems a little messy.

esteldunedain:
> SettingsInitialized

Yes, I think we could implement the delay mechanic directly in `FUNC(setHandcuffed) / FUNC(setCaptive)` to make them more robust.

However, if possible I'd prefer the delay to happen *after* `SettingsInitialized` has launched, just to be extra certain that nothing wil break. 

Hold on for tentative PR.

PabstMirror:
preSettingsInit / postSettingsInit ? :smile:

Glowbal:
@commy2 could you provide some input on the correctness for this PR?

commy2:
:+1: works flawlessly. Thanks for the contribution.

PR #3362, opened 2016-02-20
Delay captivity until 0.05s after settings are initialized

esteldunedain:

1. As discussed in #3359
2. It's probably more complicated than it needs to be; maybe @PabstMirror can think of a better solution
3. Support #3359


PabstMirror:
Looks great,
The delay in the modules had 3 effects
 - delay for the animation (which isn't needed anymore, but can't hurt)
 - those modules aren't "ace_modules" (because they can be trigger activated), so they happen before postInit which installs the event handlers
 - locality can be very weird right at mission start.  I think the server can think all units are local to it at early init.  In any case we need to account for the server loading faster than clients (server sends targetEvent before clients reach postInit)

PR #3365, opened 2016-02-20
Added translation for most Italian stringtable entries.

ChairborneIT:

1. *Describe what this pull request will do*
2. *Each change in a seperate line*

---
This adds italian translation for most stringtables.


PabstMirror:
I think the encoding got changed, so git think the entire file is new.

esteldunedain:
It's probably a problem with line endings

KoffeinFlummi:
I didn't want to merge this directly since that would make things like git-blame useless, so I merged it manually here: https://github.com/acemod/ACE3/commit/0021bc08dfd9012f290afa0c5c2a7c19afaf53a7

PR #3369, opened 2016-02-20
displayLoad xeh for imenu

commy2:

1. replace waitUntil loop for the interaction menu with DisplayLoad event 	
2. replace mainDisplayLoaded event with DisplayLoad XEH
3. remove unused inventoryDisplayChanged PFH check 	
4. replace inventoryDisplayLoaded with DisplayLoad event 	
5. replace zeusDisplayChanged with DisplayLoad event
6. remove unused mapDisplayLoaded event and ACE_dlgMap variable
7. remove unused configs and remove some wrong base classes in Scopes
8. use precompiled `BIS_fnc_initDisplay` on specialised RscUnitInfo
9. use RscUnitInfo DisplayLoad for turretDisplayLoaded event
10. use DisplayLoad for ACE main menu info box and to disable options when world=Empty


PabstMirror:
On first load:
```
XEH: PreInit finished.
XEH_displayLoad.sqf [Display #12]
XEH: PostInit started.
```
On restart:
```
XEH: PostInit finished.
XEH_displayLoad.sqf [Display #12]
```
Not a problem just something to keep in mind that in SP editor display events can occur before postInit has finished (e.g. make sure we don't use an undefined variable)

commy2:
Yup. I was worried about `DFUNC(render)` too, but it appears that the main map is always done after preInit at least. (No maps in main menu etc.)

commy2:
Parts of this will require an update of CBA events:
https://github.com/CBATeam/CBA_A3/blob/master/addons/events/XEH_preClientInit.sqf#L76-L79

...

esteldunedain:
:+1:

PR #3371, opened 2016-02-21
More severe jamming

esteldunedain:

1. By default, disable reloading from automatically clearing jams. We've always had a dedicated keybind (Shift-R) for the unjam action and it's currently never used because it's easier to just reload. I've kept the former behaviour optional in case people prefer it.
2. The unjam keybind may be difficult to remember due to jamming ocurring fairly sporadically. So I added an additional "Clear Jam" action in the Interact Menu. This is more consistent to how most ACE3 interactions are performed.
3. Increase the jam chance x2 for high temperatures and x4 for extreme temperatures. Jam chance on the normal cyclic rate temperature range are not affected.
4. Added a 10% (configurable) chance for unjam actions to fail, requiring it to be repeated. This imitates ACE2 behaviour.
5. The message stating if the jam was cleared or not now is shown toward the end of the clearing animation.

This PR makes jamming more frequent at high temps and a bit more demanding to handle


bux578:
Will it look like this? https://youtu.be/TJfcaCmMy7c?t=290

:kappa:

esteldunedain:
We do support custom animations, but nobody has produced one :(

PR #3375, opened 2016-02-21
Experimental: Ammo  Explosion events

esteldunedain:
WIP


commy2:
This has the downside, that this script is compiled for every grenade.

esteldunedain:
> This has the downside, that this script is compiled for every grenade.

Damn, that seems to be true: I can change the script in-mission and e.g. change the hint message.

The worst part is that there's zero reasons for this to behave like that, because the string you load for that parameter is inmutable. +1 for BI

commy2:
I thought about it. Maybe it's only compiled ones when the config is loaded or whatever. Needs some testing to figure out the exact behaviour.
Maybe compiling is still faster than tracking them, if the function that will be compiled is something like this:
"call ACE_frag_realFunction"

esteldunedain:
@commy2 I was testing exactly that!

Time went from 3ms to 0.07 ms, which is totally acceptable!

esteldunedain:
`#include "script_component.hpp"` really wreck compile times, which is quite logical given the amount of macros it defines, etc. CBA macros should therefore not be used in cases like this.

bux:
Status on this one?
Can we close the pull request?

PabstMirror:
I think it was a really neat idea, 
but because it requires config changes for every ammo type I think we would run into comparability issues

dedmen:
We have ammo eventhandlers now that could replace this. but they are config only.

Issue #3370, opened 2016-02-21
switch doAnimation to event driven approach


Glowbal:
For 3.5.0 release we need to switch ace_common_fnc_doAnimation to an event driven approach, as it is currently still using remoteExecFnc which is deprecated since 3.5.0.

PR #3382, opened 2016-02-21
ObjectEvent Func, swap execRemoteFnc in doAnim

PabstMirror:

 - adds ace_common_fnc_objectEvent which is very similar to targetEvent, but only operates on a single object, If local there is no network traffic/delay.

- swap out `execRemoteFnc` in doAnimation #3370


esteldunedain:
:+1:

The only problem I see is that when using `globalEvent` the local machine will play the animation with a roundtrip network lag. idk if that'd look bad for some action, probably not

Glowbal:
:+1:

PR #3384, opened 2016-02-21
Fix empty callbacks in medical/repair

PabstMirror:
`typeof (missionNamespace getVariable "")` = "EmptyDetector"

causes
```
Error call: Type Object, expected code
File z\ace\addons\repair\functions\fnc_repair_failure.sqf, line 46
```


esteldunedain:
:+1:

PR #3389, opened 2016-02-22
precompile on game start

commy2:

1. Create a `Extended_PreStart_EventHandlers` event for every module that uses the `PREP()` macro, which is used to precompile all functions into `uiNamespace` (caching) on game start, instead of on first mission start.
- functions cannot be overwritten with 2d editor or weird missions
- reduces loading screen time for the first mission start when `-world=empty`
```
 9:58:43 [0,85.129,0,"XEH: PreStart started."]
 9:59:03 [0,105.329,0,"XEH: PreStart finished."]
```
(potato PC though)


esteldunedain:
:+1: We should merge this ASAP to prevent conflicts.

PR #3392, opened 2016-02-22
fix error when loading save game, fix #3391

### LINKED ISSUE #3391: Error loading savegame latest master
```
17:36:20 Error in expression <= 12) then { 
(_display displayCtrl 51) ctrlAddEventHandler ["Draw", ace_interac>
17:36:20   Error position: <ctrlAddEventHandler ["Draw", ace_interac>
17:36:20   Error Type Any, expected String
17:36:20 File z\ace\addons\interact_menu\XEH_displayLoad.sqf, line 8
```

commy2:

1. fix #3391

### DIFF
--- a/addons/interact_menu/XEH_displayLoad.sqf
+++ b/addons/interact_menu/XEH_displayLoad.sqf
@@ -5,5 +5,5 @@ disableSerialization;
 params ["_display"];
 
 if (ctrlIDD _display == 12) then { // RscDisplayMainMap
-    (_display displayCtrl 51) ctrlAddEventHandler ["Draw", DFUNC(render)];
+    (_display displayCtrl 51) ctrlAddEventHandler ["Draw", {_this call FUNC(render)}];
 };


commy2:
This is because in savegames, the map display is loaded before preInit.

esteldunedain:
But how does the fix work? I can't see any difference between before and after

commy2:
DFUNC is undefined. {_this call FUNC()} is a vali function even when FUNC() is undefined at that point.
It's RV being shit again.

esteldunedain:
> DFUNC is undefined. {_this call FUNC()} is a vali function even when FUNC() is undefined at that point.

Perfect

commy2:
So merge? I need this in CBA.

commy2:
I guess the error is valid. The implementation of save game just has huge differences between normal games (and restarted ones ...)
We will have working save games eventually. Even if it means I have to turn everything upside down.

PR #3393, opened 2016-02-22
Fix 3den attributes

PabstMirror:

Fix the 3den attributes

 - Captives: expression only runs on server, so when checked on a remote unit it has no effect and throws
```
 [z\ace\addons\captives\functions\fnc_setSurrendered.sqf:23] -ERROR- ERROR running surrender on remote unit
```
Replace with `FUNC(moduleHandcuffed) which handles it.

 - Replace the broken `comboBox` control with custom ones for medical/repair. Looks nicer imho

 - Replace all setVariables with the public version


esteldunedain:
:+1: for merging this now.

commy2:
:+1:

PR #3395, opened 2016-02-22
Cleanup old cba SLX_XEH_COMPILE_NEW code

PabstMirror:
Fix 
```
Attempt to override final function - slx_xeh_compile_new
```
Commy I'm almost positive this can be removed, just wanted to check.


commy2:
IDK what I was thinking.

PR #3396, opened 2016-02-23
Add protocol version to meta.cpp

jiripolasek:

1. Add protocol version to meta.cpp file. Without protocol version line Arma 3 Server ignores publishedid attribute and it won't propagate it to the client / Launcher.


bux578:
Thank you. Much appreciated.

PR #3400, opened 2016-02-23
Fix lamp unloading

### LINKED ISSUE #3363: Fix issue cargo unloading from PR #2230 (lamps)
Fix issue cargo unloading from PR #2230

jonpas:

1. Fix #3363 
2. Make sure `hideObjectGlobal` is executed before `setPosATL` by executing both on server, to ensure lamps will get rendered correctly


PabstMirror:
:+1: 
Tested on dedicated, when unloading it takes 2-3 seconds for the model to show up (but the lights pop in right away). 
It's really not a big deal and is still completely usable; we might be able to make a feedback tracker post, but it's so minor I doubt anything would happen.

Tagging @SzwedzikPL as this should get his PR working fully.

SzwedzikPL:
great :+1:

PR #3403, opened 2016-02-23
Reimplement old parameter syntax for (Un)LoadCargo events

jonpas:

1. Reimplement the `[object, vehicle]` syntax for arguments as it initially was and as it is documented

Those events are considered public API, that's why I am tagging this with "Bug fix".


esteldunedain:
:+1:

Issue #3404, opened 2016-02-24
Cleanup unused and undefinend functions


BaerMitUmlaut:
I wrote a little script that crawls all SQF, HPP and CPP files for unused and undefined functions, similar to `search_undefinedFunctions.py` but without the use of SQF. Public functions are already filtered as well. This is its current output:

_Undefined functions:_
- [ ] `ace_common_fnc_localanim`
- [ ] `ace_missileguidance_fnc_checkseekerlos`
- [x] `ace_refuel_fnc_canattach`

_Unused functions:_
- [ ] `ace_captives_fnc_canfriskperson`
- [ ] `ace_captives_fnc_dofriskperson`
- [ ] `ace_captives_fnc_handleonunconscious`
- [ ] `ace_cargo_fnc_canload`
- [x] `ace_common_fnc_addactionmenueventhandler`
- [x] `ace_common_fnc_addcuratorunloadeventhandler`
- [ ] `ace_common_fnc_addsetting`
- [ ] `ace_common_fnc_debugmodule`
- [ ] `ace_common_fnc_dumpperformancecounters`
- [x] `ace_common_fnc_execpersistentfnc`
- [x] `ace_common_fnc_exportconfig`
- [ ] `ace_common_fnc_getdefinedvariable`
- [ ] `ace_common_fnc_getnumbermagazinesin`
- [x] `ace_common_fnc_getselectionswithouthitpoints`
- [x] `ace_common_fnc_getstaminabarcontrol`
- [ ] `ace_common_fnc_modulelsdvehicles`
- [x] `ace_common_fnc_owned`
- [x] `ace_common_fnc_removeactionmenueventhandler`
- [x] `ace_common_fnc_removecaninteractwithcondition`
- [ ] `ace_common_fnc_removesyncedeventhandler`
- [ ] `ace_common_fnc_requestsyncedevent`
- [ ] `ace_common_fnc_showuser`
- [ ] `ace_common_fnc_throttledpublicvariable`
- [ ] `ace_common_fnc_translatetomodelspace`
- [ ] `ace_common_fnc_translatetoweaponspace`
- [ ] `ace_common_fnc_waveheightat`
- [ ] `ace_fcs_fnc_getangle`
- [ ] `ace_fcs_fnc_updaterangehud`
- [ ] `ace_frag_fnc_addblacklist`
- [ ] `ace_frag_fnc_removepfhround`
- [ ] `ace_frag_fnc_removetrack`
[...]

PR #3405, opened 2016-02-24
remove unused non API functions

commy2:

1. remove unused non API functions

#3404


jonpas:
:+1:

PR #3411, opened 2016-02-25
Refuel change documentation header for API fnc

GitHawk:
Cleanup for #3404

### DIFF
--- a/addons/refuel/functions/fnc_connectNozzleAction.sqf
+++ b/addons/refuel/functions/fnc_connectNozzleAction.sqf
@@ -60,7 +60,7 @@ while {(_closeInMax - _closeInMin) > 0.01} do {
 _closeInDistance = (_closeInMax + _closeInMin) / 2;
 
 //Checks (too close to center or can't attach)
-if (((_startDistanceFromCenter - _closeInDistance) < 0.1)) exitWith {
+if ((_startDistanceFromCenter - _closeInDistance) < 0.1) exitWith {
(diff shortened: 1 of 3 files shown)


GitHawk:
This PR is for refuel @esteldunedain

jonpas:
Thank you @GitHawk

addons/refuel/functions/fnc_connectNozzleAction.sqf
```diff
@@ -60,7 +60,7 @@ while {(_closeInMax - _closeInMin) > 0.01} do {
 _closeInDistance = (_closeInMax + _closeInMin) / 2;
 
 //Checks (too close to center or can't attach)
-if (((_startDistanceFromCenter - _closeInDistance) < 0.1) || {!([_target, _unit, _itemClassname] call FUNC(canAttach))}) exitWith {
+if ((_startDistanceFromCenter - _closeInDistance) < 0.1) exitWith {
```
esteldunedain: why was this changed? Should it be reverted @GitHawk?

GitHawk: Because there is no `ace_refuel_fnc_attach` function.
Since this code was copied from `ace_attach` it was a remnant and has never thrown an error.
In `ace_attach` the` fnc_canAttach` limits the amount of attached objects.

esteldunedain: k, np then

PR #3415, opened 2016-02-25
Added Show Player Names option to Blue Force Tracking Module

looterz:

1. Add an option to the "Blue Force Tracking" module that will show individual players names instead of their group IDs, defaults to false.

### DIFF
--- a/addons/map/ACE_Settings.hpp
+++ b/addons/map/ACE_Settings.hpp
@@ -1,59 +1,69 @@
 class ACE_Settings {
     class GVAR(BFT_Interval) {
+        category = CSTRING(Module_DisplayName);
         value = 1.0;
         typeName = "SCALAR";
         displayName = CSTRING(BFT_Interval_DisplayName);
         description = CSTRING(BFT_Interval_Description);
     };
     class GVAR(BFT_Enabled) {
+        category = CSTRING(Module_DisplayName);
         value = 0;
         typeName = "BOOL";
         displayName = CSTRING(BFT_Enabled_DisplayName);
         description = CSTRING(BFT_Enabled_Description);
     };
     class GVAR(BFT_HideAiGroups) {
+        category = CSTRING(Module_DisplayName);
         value = 0;
         typeName = "BOOL";
         displayName = CSTRING(BFT_HideAiGroups_DisplayName);
         description = CSTRING(BFT_HideAiGroups_Description);
     };
     class GVAR(BFT_ShowPlayerNames) {
+        category = CSTRING(Module_DisplayName);
         value = 0;
         typeName = "BOOL";
         displayName = CSTRING(BFT_ShowPlayerNames_DisplayName);
         description = CSTRING(BFT_ShowPlayerNames_Description);
     };
     class GVAR(mapIllumination) {
+        category = CSTRING(Module_DisplayName);
         value = 1;
         typeName = "BOOL";
         displayName = CSTRING(MapIllumination_DisplayName);
         description = CSTRING(MapIllumination_Description);
     };
     class GVAR(mapGlow) {
+        category = CSTRING(Module_DisplayName);
         value = 1;
         typeName = "BOOL";
         displayName = CSTRING(MapGlow_DisplayName);
         description = CSTRING(MapGlow_Description);
     };
     class GVAR(mapShake) {
+        category = CSTRING(Module_DisplayName);
         value = 1;
         typeName = "BOOL";
         displayName = CSTRING(MapShake_DisplayName);
         description = CSTRING(MapShake_Description);
     };
     class GVAR(mapLimitZoom) {
+        category = CSTRING(Module_DisplayName);
         value = 0;
         typeName = "BOOL";
         displayName = CSTRING(MapLimitZoom_DisplayName);
         description = CSTRING(MapLimitZoom_Description);
     };
     class GVAR(mapShowCursorCoordinates) {
+        category = CSTRING(Module_DisplayName);
         value = 0;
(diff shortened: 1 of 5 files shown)


addons/map/functions/fnc_blueForceTrackingUpdate.sqf
```diff
@@ -17,6 +17,7 @@ if (GVAR(BFT_Enabled) and {(!isNil "ACE_player") and {alive ACE_player}}) then {
     _playerSide = call EFUNC(common,playerSide);
 
     _groupsToDrawMarkers = allGroups select {side _x == _playerSide};
+    _playersToDrawMarkers = allPlayers select {side _x == _playerSide};
```
esteldunedain: I'd move this inside the if below for perfomance reasons

looterz: Good catch, missed that one.

addons/map/functions/fnc_blueForceTrackingUpdate.sqf
```diff
@@ -26,6 +26,28 @@ if (GVAR(BFT_Enabled) and {(!isNil "ACE_player") and {alive ACE_player}}) then {
...
+
+            private _marker = createMarkerLocal [format ["ACE_BFT_%1", _forEachIndex], [(getPos leader _x) select 0, (getPos leader _x) select 1]];
+            _marker setMarkerTypeLocal _markerType;
+            _marker setMarkerColorLocal _colour;
+            _marker setMarkerTextLocal (name _x);
+
+            GVAR(BFT_markers) pushBack _marker;
+        } forEach _playersToDrawMarkers;
+
+        _groupsToDrawMarkers = _groupsToDrawMarkers select {
```
esteldunedain: If I undertand this correctly, this part should be removed, as it'd hide AI groups when `BFT_ShowPlayerNames` is true. The same behaviour can be achieved already by setting `BFT_HideAiGroups` to true, right?

looterz: I filter players out of the _groupsToDrawMarkers array when BFT_ShowPlayerNames is true in case BFT_HideAiGroups is false, this way you can have players with individual names and icons and AI groups shown normally. 

It probably would be best to wrap that part in a condition that checks whether or not BFT_HideAiGroups is false though.

looterz: I'll wrap it in a condition that makes its purpose a bit clearer, what it does is filter players out of the _groupsToDrawMarkers array if ShowPlayerNames is true, so if HideAiGroups is false, AI groups draw normally while players draw as intended.

addons/map/functions/fnc_blueForceTrackingUpdate.sqf
```diff
@@ -26,6 +26,31 @@ if (GVAR(BFT_Enabled) and {(!isNil "ACE_player") and {alive ACE_player}}) then {
...
+
+            private _marker = createMarkerLocal [format ["ACE_BFT_%1", _forEachIndex], [(getPos leader _x) select 0, (getPos leader _x) select 1]];
+            _marker setMarkerTypeLocal _markerType;
+            _marker setMarkerColorLocal _colour;
+            _marker setMarkerTextLocal (name _x);
+
+            GVAR(BFT_markers) pushBack _marker;
+        } forEach _playersToDrawMarkers;
+
+        if (not GVAR(BFG_HideAiGroups)) then
```
jokoho48: if i remeber right we not use not we use !

esteldunedain:
:+1: Ready for merge

addons/map/functions/fnc_blueForceTrackingUpdate.sqf
esteldunedain: ```if !(GVAR(BFG_HideAiGroups)) then```

Would be more consistant with ace styling

addons/map/functions/fnc_blueForceTrackingUpdate.sqf
looterz: Actually nevermind, I forgot it needs to filter out players regardless of HideAiGroups condition, or else the group markers will draw over the names which is unintended. I've tested it, and it works with AI groups fine, enabled or disabled.

esteldunedain:
Ok, I think I understand now. If `BFT_ShowPlayerNames `is enabled then the player groups are already drawn on that piece of code, so they need to get filtered so they are not drawn again.

:+1: Ready for merge

looterz:
Exactly, if HideAiGroups is true and ShowPlayerNames is true, the player names section takes care of the markers itself and the groups part is basically skipped due to being an empty array, and if HideAiGroups is false and ShowPlayerNames is true, players are filtered out of the groups array so they aren't drawn twice.

addons/map/ACE_Settings.hpp
```diff
@@ -17,6 +17,12 @@ class ACE_Settings {
         displayName = CSTRING(BFT_HideAiGroups_DisplayName);
         description = CSTRING(BFT_HideAiGroups_Description);
     };
+    class GVAR(BFT_ShowPlayerNames) {
```
Glowbal: Missing settings category.

esteldunedain: None of the settings have, but it could be a great oportunity for adding them.

@looterz, could you add this to all the settings?

`category = CSTRING(Module_DisplayName);`

looterz: Sure will do.

Glowbal:
Thanks @looterz :+1:

PR #3419, opened 2016-02-25
Add missing medical category to delayUnconCaptive

jonpas:

1. Ref. title


jokoho48:
@jonpas this is not a client setting and dont need a category

PabstMirror:
Categories should still work on the config exporter

jonpas:
What @PabstMirror said, descriptions should be there as well, but not that important and I don't know what exactly it does.

Glowbal:
All settings should have a category. They are used within the config export browser, same as for regular client settings.

:+1: @jonpas

Issue #3310, opened 2016-02-12
Huron fuel Container has no visible hose


Gaspasar:
ACE3 Version: 3.4.2.0

**Mods:**
@CBA_A3
@ace

**Placed ACE3 Modules:**
none

**Description:**
The Huron fuel container (B_Slingload_01_Fuel_F) has no visible hose.
If u drop the invisible hose on the ground its hard to find it again :(

**Steps to reproduce:**
place a huron fuel container in the Editor and try to refuel a car.

**Where did the issue occur?**
Editor, SP

**RPT log file:**
not necessery

jokoho48:
@GitHawk

GitHawk:
Probably bis' fault

PabstMirror:
Seems like `createRope` needs a real vehicle for the first param. So it returns nothing on the slingloads which are a reAmmoBox.

Not sure if there is a good way to fix this. Could possibly create a dummy vehicle and attach it to the container, but that will get messy.

GitHawk:
Yes, that's what I mean. Is the same problem, that keeps me from making refuel work with fuel stations.

jonpas:
On dev branch yesterday they enabled ropeCreate on static objects again.

GitHawk:
Yay, then we may be able to fix that in 1.57+ and add fuel stations.

Gaspasar:
with the Taru fuel Pod (Land_Pod_Heli_Transport_04_fuel_F) it is working correct.
isnt it also a static box?

GitHawk:
I can't check inheritance now but it seems like the Taru fuel pod is a vehicle.

jonpas:
It's a vehicle yes, you can sit in it.

jonpas:
Pushing back as it is supposed to be fixed on dev-branch.

@GitHawk can you verify on dev-branch?

GitHawk:
Not working on dev branch for me with `ropeCreate [fuel_container, [0,0,0], player, [0,0,0], 20];`

john681611:
Possible Bodge solution:  ATV is small enough to fit inside the fuel tank section of the container. Using the ATV as a surrogate attachedTo the container (0,0,1 is sufficient) this does look correct.

esteldunedain:
@john681611, the proposal was to create an invisible dummy instead. But the concensus was that it's better to wait and see if BIS fixes the issue instead,

john681611:
[...]

PR #3430, opened 2016-02-26
Refuel improvements

GitHawk:
Some minor improvements to refuel
- [x] Test if AllVehicles support ropeCreate >> Vehicles work, players and static objects don't
- [x] Test all non vehicle fuel sources

Documentation for other modders/creators on how to enable refuel will be in a different PR

### DIFF
--- a/addons/refuel/functions/fnc_takeNozzle.sqf
+++ b/addons/refuel/functions/fnc_takeNozzle.sqf
@@ -55,10 +55,13 @@ if (isNull _nozzle) then { // func is called on fuel truck
             _newNozzle attachTo [_unit, [-0.02,0.05,-0.12], "righthandmiddle1"];
             _unit setVariable [QGVAR(nozzle), _newNozzle, true];
 
-            private _rope = ropeCreate [_target, _endPosOffset, _newNozzle, [0, -0.20, 0.12], REFUEL_HOSE_LENGTH];
+            if (_target isKindOf "AllVehicles") then {
+                // Currently ropeCreate requires its first parameter to be a real vehicle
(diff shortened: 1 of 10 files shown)


addons/refuel/functions/fnc_takeNozzle.sqf
```diff
@@ -55,10 +55,13 @@ if (isNull _nozzle) then { // func is called on fuel truck
             _newNozzle attachTo [_unit, [-0.02,0.05,-0.12], "righthandmiddle1"];
             _unit setVariable [QGVAR(nozzle), _newNozzle, true];
 
-            private _rope = ropeCreate [_target, _endPosOffset, _newNozzle, [0, -0.20, 0.12], REFUEL_HOSE_LENGTH];
+            if (_target isKindOf "AllVehicles") then {
+                // Currently ropeCreate requires its first parameter to be a real vehicle
+                private _rope = ropeCreate [_target, _endPosOffset, _newNozzle, [0, -0.20, 0.12], REFUEL_HOSE_LENGTH];
+                _newNozzle setVariable [QGVAR(rope), _rope, true];
+            };
```
Glowbal: If we cannot create a rope, doesn't that mean something is wrong and we cannot properly let a user take the fuel nozzle?

GitHawk: We can either support those objects (i.e. gas pump) and the player has to "live without the rope", or we exclude them from refuel.

Since I am in hope that BIS will fix this problem, I'd like to include those objects and find ways to make refuel work.
Especially if it creates problems like mentioned in #3310, where the CSAT slingload fuel tank is working but the NATO isn't.

I'm also under the impression that calling `createRope` on a non vehicle seems to exit/break the current scope as `QGVAR(source)` was not set on the nozzle when interacting with a `B_Slingload_01_Fuel_F` object. I received the nozzle, but could not return it.

PR #3438, opened 2016-02-27
AAR actions + track temperatures of spare barrels

esteldunedain:

1. Change spare barrels from weapons to unique magazines.
2. Have the server tracks temperatures of spare barrels and calculate their cooling down.
3. Make swapping barrelsselects the coolest barrel from the inventory. Current temperature of that barrel is transfered to the weapon.
4. Add an action to check the temperature of all carried spare barrels.
5. Increase the mass of spare barrels to ~2.7 Kg
5. AAR can now interact with the AR weapon to:
    - Check its temperature.
    - Assist on swapping the barrel, taking the spare barrel from his own equipment.

### Known issues

1. ~~Same as vehicle keys, the name of the spare Barrel item cannot be localized, because it is part of the it's name as returned by `magazineDetails`. This could be worked around if deemed important enough~~ Fixed.
2. ~~I'm not sure how to add magazines in config because it'd require the += operator. Maybe @commy2 can help with that. Otherwise, they could be added through scripting like the earplugs are, or not automatically added at all~~ Spare barrels were removed from AR equipments.


esteldunedain:
Still a bit WIP. todo:
- [ ] Move some of the code out of `postInit` so it's a little cleaner.
- [ ] Add an animation and short progressbar to the action that checks the temperature of all the spare barrels.
- [ ] Add a hint with the weapon temp after swapping barrels.

PabstMirror:
Should we leave the old `ACE_SpareBarrel` as an item and add a new magazine classname, I'm worried about breaking old missions that use `addItem`

commy2:
`addItem` works with CfgMagazines. The only reason to ever use `addMagazine` over `addItem` is the alternative syntax with the ammo count. The simple syntax even fails with remote units, like `addWeapon` and `addBackpack`.

commy2:
I don't think anything prevents us from using a displayName. I don't think "Spare Barrel" will share it's name with any other item in any language.

```
FNC_getMagazineId = {

params [["_unit", objNull, [objNull]], ["_magazine", "", [""]]];

private _displayName = getText (configFile >> "CfgMagazines" >> _magazine >> "displayName");

private _return = magazinesDetail _unit select {_x find _displayName == 0} apply {_x splitString "[/:]"};

{
    reverse _x;
} forEach _return;

_return apply {_x select 1};

};
```

esteldunedain:
@commy2, that code is pretty much what I was referring with this.

> This could be worked around if deemed important enough.

It just makes the code a little more complicated.

Glowbal:
For 3.6.0?

Glowbal:
Ah @esteldunedain just moved it ahead.

esteldunedain:
Sorry, I never intended this for 3.5.0, I marked it by mistake.

commy2:
>It just makes the code a little more complicated.
Function could be moved to common or even CBA. It's just that this has overhead. Losing translations for items feels bad.

esteldunedain:
> It's just that this has overhead. 

Yeah, it's very minor though. It only needs to be done a maximum of 4~5 times each time a barrel is swapped or temperatures are checked.

esteldunedain:
@commy2, I'm more worried about the += operator for `magazines` and `respawnMagazines`. Do you think that's viable or we shoul switch to a scripted solution?

commy2:
Only way to test this is to spawn as every machine gunner with the item and then to check if any grenades shows up as 0x in the hud, or if any items are missing.
We could also simply don't add spare barrels to machine gunners. Maybe it's enough to add them to the backpacks with the assistant machine gunners instead.
The bug is also not caused by +=, but simply by the items[] etc. arrays being too full for the units containers.

addons/overheating/functions/fnc_calculateCooling.sqf
```diff
@@ -0,0 +1,58 @@
...
+//Area of a cylinder (2/r)*(Pi * r^3 + V) - for a 0.008255m radius barrel -> Area = 210(1/meters) * Volume
+//Adjusted volume for being hollowed out is ~1.1x
+//So Area = 210 * 1.1 * (mass / 7850) = mass * 0.029427 (for steel near that diameter)
+
+private _barrelSurface = _barrelMass * 0.029427;
+
+TRACE_4("cooling",_temperature,_totalTime,_barrelMass,_barrelSurface);
+
+private _time = 0;
+while {true} do {
```
jokoho48: while true in a call? are you sure?

Glowbal: It works but cleaner might be;
```sqf
while {_time >= _totalTime} do { .. };
_temperature max 0;
```

And obviously remove the `exitWith` statement on line 57.

esteldunedain: Plus it's old code, I just moved it to a sepparate function.

But I do like @Glowbal's version better.

commy2:
I cooked up these. Dunno if it's that useful for CBA or if it's superfluous.
```
CBA_fnc_getMagazineIds = {
    params [["_unit", objNull, [objNull]], ["_magazine", "", [""]]];

    private _displayName = getText (configFile >> "CfgMagazines" >> _magazine >> "displayName");

    if (_displayName isEqualTo "") exitWith {[]};

    magazinesDetail _unit select {_x find _displayName == 0} apply {_x = _x splitString "[:]"; _x select (count _x - 1)};
};
```
```
CBA_fnc_getCurrentMagazineId = {
    params [["_unit", objNull, [objNull]], ["_turret", nil, [[]]]];

    private "_magazine";

    if (isNil "_turret") then {
        _magazine = currentMagazineDetail _unit splitString "[:]";
    } else {
        _magazine = (_unit currentMagazineDetailTurret _turret) splitString "[:]";
    };

    _magazine param [count _magazine - 1, ""]
};
```

esteldunedain:
This is pretty much ready, although more testing in MP would be good.

As soon as @commy2 commit `CBA_fnc_getMagazineIds` to CBA I'll remove the fake version in `XEH_preInit.sqf` that's confusing Travis. 

BTW @commy2, your function worked great even with text with accents and non-typical letters like the spanish translation of "Spare Barrel".

commy2:
Yup. It cannot be broken with any localization.

addons/overheating/XEH_preInit.sqf
```diff
@@ -4,4 +4,14 @@ ADDON = false;
...
 #include "XEH_PREP.hpp"
 
+FakeCBA_fnc_getMagazineIds = {
+    params [["_unit", objNull, [objNull]], ["_magazine", "", [""]]];
+
+    private _displayName = getText (configFile >> "CfgMagazines" >> _magazine >> "displayName");
+
+    if (_displayName isEqualTo "") exitWith {[]};
+
+    magazinesDetail _unit select {_x find _displayName == 0} apply {_x = _x splitString "[:]"; _x select (count _x - 1)};
```
Glowbal: Wrap this in `()` and travis will no longer complain. 
```sqf
(magazinesDetail _unit select {_x find _displayName == 0}) apply {_x = _x splitString "[:]"; _x select (count _x - 1)};
```

addons/overheating/XEH_preInit.sqf
```diff
@@ -4,4 +4,14 @@ ADDON = false;
 
 #include "XEH_PREP.hpp"
 
+FakeCBA_fnc_getMagazineIds = {
```
Glowbal: Not a fan of this - rather use an `if isnil` check and use the actual name instead.

esteldunedain:
I merged the AAR actions branch into this PR to keep things simple, specialy because I made some changes to the spare barrels functions to support the new code.

@Glowbal, this should be ready for reviewing/merging. It's based on master, so target is 3.6.0.

Issue #3447, opened 2016-02-27
Filter out empty categories on the options menu


esteldunedain:
Currently some settings categories don't have any user settable settings, but are still shown in the options menu.

PR #3453, opened 2016-02-28
Hide categories that have no user settable options

PabstMirror:
Hide categories that have no user settable options
#3447


esteldunedain:
:+1:

PR #3454, opened 2016-02-28
fix reloadMutex nil, fix#3450

### LINKED ISSUE #3450: ace_gestures_reloadmutex is nill on mission restart on 3den, inited from DisplayLoad XEH
ACE3 3.5.0 RC

E.g. 
- Move forward
- Rally up
- Stop
- Cover
- etc

commy2:

1. fix #3450

GVAR(ScrollWheelFrame) is no longer needed with this approach, as no more duplicate events will appear with DisplayLoad.

All mission namespace variables have still to be defined in pre or post init. Somehow I missed GVAR(ReloadMutex).

NOTE NOTE NOTE
In single player, the keyDown event that does the reloadMutex thing will be removed on restarts by CBA thanks to this splendidTM line:
https://github.com/CBATeam/CBA_A3/blob/master/addons/events/XEH_preClientInit.sqf#L76-L79
Any CBA for A3.
This will be fixed by my CBA Events rewrite (WIP).

Alternatively we could use `CBA_fnc_addDisplayHandler`, but that would mean that it currently doesn't work with save games (incl. MP). Also fixed by the rewrite.

This should be good enough for now as all variables will be defined. Fixing CBA will do the rest.


addons/gestures/CfgEventHandlers.hpp
```diff
@@ -19,6 +19,6 @@ class Extended_PostInit_EventHandlers {
 
 class Extended_DisplayLoad_EventHandlers {
     class RscDisplayMission {
-        ADDON = QUOTE(_this call COMPILE_FILE(XEH_mainDislayLoad));
+        ADDON = QUOTE(_this call COMPILE_FILE(XEH_missionDislayLoad));
```
bux578: Dislay?

jonpas:
:+1:

Glowbal:
#3455

PR #3456, opened 2016-02-28
Standardize GVAR array brackets in configs.

VKing6:
Standardize grammar of GVAR arrays in configs to place the square brackets outside the GVAR macro.
This prevents any possible confusion when the GVARs are used in code.

Put in 3.6.0 for absolute lack of priority, but no reason it can't be merged earlier.


commy2:
:+1: very good idea

esteldunedain:
:+1: because it's more search friendly

PR #3463, opened 2016-02-29
Fast roping RC fixes

BaerMitUmlaut:

1. Remove all animations from TOH because of various errors, replace the fastroping animation with the one from AGM.
2. Cleanup the interaction conditions and fix having the ability to deploy ropes when you should not.


alganthe:
:+1: all good.

esteldunedain:
`checkVehicleThread` should be called in the server, because the machine that created the helicopter might disconnect, leaving the thread interrupted. However, I'd say we can merge the PR and make a ticket to solve that for the next version, along with the issue when deleting the chopper during fast roping.

Issue #3339, opened 2016-02-18
Module has wrong values when placed with 3D Editor


Gilatar:
**ACE3 Version:**
3.4.2.0

**Mods:**
@CBA_A3 (tested with v2.2.0 and v2.3.0, same results)
@ACE

**Description:**
There is some odd behavior with the module values in the 2D editor versus the 3D editor, specifically with the ACE Medical -> Medical Settings Module. 

[2D Editor Default Module Settings](http://puu.sh/ncLG2/d7078ddf02.png)
[3D Editor Default Module Settings](http://puu.sh/ncLbJ/65875a418d.png)

The 2D editor modules are fine and won't cause any issues, the 3D editor-placed ones, however, seem to give the wrong values for some of the settings, regardless if you make changes to the setting or not.

Looking at the mission.sqm - a 2D editor-placed [module with AI unconsciousness turned off has a value of 0](http://puu.sh/ncMa0/c8e199d486.png). A 3D editor-placed [module with AI unconsciousness turned off has a value of -1.](http://puu.sh/ncMxR/fb59a31aec.png)

Also causes some .RPT outputs listed below.

**Steps to reproduce:**
- Make a fresh mission in 3D editor
- Place player
- Place medical module (set AI unconsciousness to disabled)
- Place enemy group
- Preview
- Shoot enemies and they will still go unconscious
- Check RPT for errors below

**Where did the issue occur?**
3D Editor

**.RPT:**
http://pastebin.com/2MCdtRiY

**Excerpt from RPT:**
19:42:33 [ACE] (common) WARNING: Setting [ace_medical_medicSetting] out of bounds -1 (values[] count is 3)(
19:42:33 [ACE] (common) WARNING: Setting [ace_medical_enableUnconsciousnessAI] out of bounds -1 (values[] count is 3)(

PabstMirror:
`_logic getVariable "medicSetting"` = -1
Unbined sqm:
```
class Attribute13 {
    property="ACE_moduleMedicalSettings_medicSetting";
    expression="_this setVariable ['medicSetting',_value,true];";
    class Value {
        class data {
            class type {
                type[]= {
                    "SCALAR"
                };
            };
            value=-1;
        };
    };
};
```

This only happens for the disabled setting, basic gets 1 and adv gets 2 as it should.

PabstMirror:
http://feedback.arma3.com/view.php?id=28031
Reproduced in vanilla, seems like a fairly critical issue for our modules.

[...]

PR #3476, opened 2016-02-29
Try to handle -1 from eden modules

PabstMirror:
Help with  #3339

Bandaid for the bad mission modules in 3den.

E.G. a setting like:
```
    class GVAR(medicSetting) {
        category = CSTRING(Category_Medical);
        value = 1;
        typeName = "SCALAR";
        values[] = {"Disabled", "Normal", "Advanced"};
    };
```

It should be safe to assume that as the setting has a "values" array, which are strings shown for the server setting exporter, then a value of -1 wouldn't make any sense.

I've tried looking for cases were this might not be safe to assume, I can't find any so far.


esteldunedain:
Wouldn't stuff like this fail:

https://github.com/acemod/ACE3/blob/master/addons/nametags/CfgVehicles.hpp#L13-L44

PabstMirror:
They are handled before this func is called. - https://github.com/acemod/ACE3/blob/master/addons/nametags/functions/fnc_moduleNameTags.sqf#L26-L28

A module set to 0 (disabled) will be saved as -1, which means `FUNC(moduleNameTags)` will never make the call to `EFUNC(common,readSettingFromModule)` and the setting won't be forced off.

Because -1 and 0 are both valid options from the module; I don't see any good way to fix.
But this PR won't make those cases any worse.

esteldunedain:
Perfect :+1: for merging this and making a warning on release

commy2:
`missionVersion`
What about missions made with 3den and then converted to 2D? I know this is a common practice to avoid bugs with the HC, but the 3den values would still be wrong. I think this check should be removed for maximum compatibility.

PabstMirror:
Good point commy, the check should be safe on any mission version

PR #3477, opened 2016-02-29
Add Advanced Throwing

### LINKED ISSUE #1628: Tweak grenade modes to make them more useful
The following suggestions com from Dyslyecxi:

> The 'drop grenade' feature of the ACE3 grenades could be made useful if instead of it falling immediately on ammo creation, it instead had a small velocity forward, sufficient to clear windowsills/ledges etc. Standing right up against this ledge and using it causes the grenade to land on said ledge. 

image1

> I guess that's relevant for @commy2 based on the source for it

> "precise throw" would also be a bit more usable if it was zeroed for something like 5-10m away from the player. Right now it's aligned with weaponDirection, so by the time it gets to a practical short-range distance, it's already fallen significantly below the crosshair. The problem with the default grenade throw is that it has that zero too far away, so it goes significantly above the crosshair/point of aim. Tweaking that a bit could make it pretty usable.

> A quick test shows that normal throw has a 13 meter zero, so probably 5m would be a good compromise.

> `_projectile setVelocity (_velocity vectorMultiply 0.45);`

> seems to be pretty good for a 'drop-esque' short lob. Lands three meters away for me when aiming level, has enough velocity to clear a ledge. Aiming up/down allows you to finesse it past that point if necessary, such as if you wanted to put a chemlight exactly on a given spot for whatever reason.

> Makes it much more viable for urban usage.

> Aiming up to cause it to land closer:

image2

> Aiming level:

image3

> In MOUT this is a big thing, particularly as a defender of a structure. Being able to drop a grenade out of a window is huge when you're being assaulted. Example usage:

image4

> And on the flip side, sometimes you get in a position where you want to drop a grenade right inside a room without throwing it across the room.

image5

> Also dropping over walls. You get the idea. It's essentially taking the "drop" and turning it into a very short lob, with aiming high/low permitting you to make it behave the same as the existing drop functionality. 

image6

### LINKED ISSUE #3939: Advanced Throwing Issue
**ACE3 Version:** `3.5.1` (eedb406ae5605c2dd5663fd0b2de2bb9bf501b6b with advancedThrowing branch)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- If the grenade you're cooking disappears (by exploding or just by time of life limit), grenade throwing interface will stay, only option to remove it is to press reload which triggers some fallback mechanism.
Video.
https://www.youtube.com/watch?v=KP1906ClP_E

**Steps to reproduce:**
- Take a flashbang.
- Cook it.
- Wait a few seconds.

**Placed Modules:**
- None.

**RPT log file:**
- N/A.

jonpas:

- Integrate (90% rewritten) grenade throwing system by @Dslyecxi
- Close #1628
- Add settings to draw throw arc, show mouse hints, enable pick up, enabe attached pick up
- Improve and optimize rendering and handling of throwables
- Add support for scripted throwables (eg. flashbang, handheld flares)
- Improve 3rd person support
- Use CBA DisplayLoad and keybinds
- Add toggle lean support
- Add pick up ability (including attached)
- Add feature documentation

Todo:
- [x] Fix init failing on first game load (after loading from Eden)
- [x] Convert KeyDown EH to XEH DisplayLoad (once fixed in CBA) and DisplayLoad XEH to `CBA_fnc_addDisplayHandler` postInit
- [x] Throw `fired` event
- [x] Cleanup `waitUntilAndExecute`s and `waitAndExecute`s where possible (`FUNC(throw)` needs serious work on that)
- [x] Add button icons (like when carrying) showing controls - when #3551 merged
- [x] Add "double right click to drop when cooked" to mouse hint
- [x] `player` -> `ACE_player`/passed `_unit`
- [x] ~~Use grenade throw key (requires intercepting that key so it doesn't do its default action)~~ - Use custom CBA key instead of "Cycle Throw Items" instead.
- [x] Cancel throw when !canInteract (e.g. unconsciousness)
- [x] Cancel throw when player switch
- [x] If you pull out a throwable and throw it when your magazine is partially full, and in doing so, exhaust all of your throwables, the next time you press the 'cycle throwables' key (with no throwables available to throw) it will reset your ammo count to whatever it was during your last valid throw
- [x] Adjust position when locked in lean mode and side stances
- [x] Allow throwing with weapon down from FFV
- [x] Add better notification when unable to throw from inside vehicle
- [x] Disable leaning addition in vehicle
- [x] Fix scroll wheel up not working every second restart
- [x] 3rd person support
- [x] Improve 3rd person support
- [x] ~~`DefaultAction` doesn't get executed when in FFV with lowered weapon, so throw requires 2 LMB presses to actually throw~~ design decision to leave it like that, simpler for end-user to understand and not make mistakes because of it (can't change direction of player in FFV slot, raising weapon changes look direction to where weapon is pointing)
- [x] ~~`KeyUp` isn't guaranteed to fire (for some reason) - for example holding it down for short amount of time - apparently vanilla issue~~ fixed by changing it to "toggle drop mode" instead of holding a button
- [x] `DefaultAction` doesn't fire when `driver vehicle == unit` - fixed by changing to `mouseButtonDown` event and leaving `DefaultAction` for "fire" suppression only
- [x] Flashbang and handheld flares support
- [x] Fix "exploded in hand" not exiting throw mode
- [x] Make `weaponselect` change throwable keys compatible with from inside vehicle throwing.
- [x] Arc and mouse hint client settings
- [x] Editor module
- [x] Rename to "advancedthrowing"
- [x] Fix throw orientation being different than holding orientation
- [x] Moving (flying) rendering
- [x] Picking up primed throwables
- [x] ~~Replace animation with gesture to properly fix the moving mistakes (current fix makes grenade be thrown from behind you if you walk or run)~~ no suitable gesture, current fix will have to do
- [x] Change `GVAR`s to `setVariable`s
- [x] Convert to CBA events (required #3815)
- [x] Module icon
- [x] Fix #3939 
- [x] Use #4096 for throw gesture
- [x] Beautified component name
- [x] Fix walking sideways jiggle when weapon on back or lowered
- [x] Disable debug
- [x] Limit throwing power over shoulder
- [x] Cool-down between throws (mimic vanilla behaviour, also to prevent [issues](https://github.com/acemod/ACE3/pull/3477#issuecomment-234720437))
- [x] Disallow selecting throwable if reloading
- [x] Change pick-up settings to not be client settable
- [x] Use CBA player events
- [x] Improve pick-up action performance in multiplayer (test 0f6d54ee01881b6c548b8cb09344022881c89f19)
- [x] Add proper player detach support
- [x] Fix arc accuracy

Future:
-> #4218 

Experimentation:
- [x] ~~Rolling throwables~~ Arma 3 physics allow you to throw into the floor and it will be almost the exact same effect as ACE2 grenade rolling, not much point implementing
- [x] ~~Custom pin sound instead of regular throw sound~~ -> `ace_grenades` separate component if ever

To decide:
- ~~Remove grenade modes from `ace_grenades`?~~ Leave for anyone who wants to use it
- ~~Make Advanced Throwing mandatory or keep it optional (default enabled)? (connects with above, if we keep it it might be nice to keep old modes as well? if removed rename to `ace_throwing`?)~~ Leaving optional
- ~~Add support for picking up primed throwables to vanilla throwing? (if yes, move to `ace_grenades`?)~~ No, due to no nice way of holding them, apart from delay from pick up to throw


addons/grenadethrowing/functions/fnc_initControls.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_enable"];
+
+if (_enable) then {
+    GVAR(KeyDownHandle) = (findDisplay 46) displayAddEventHandler ["KeyDown", {_this call FUNC(onKeyDown)}];
+    GVAR(MouseButtonDownHandle) = (findDisplay 46) displayAddEventHandler ["MouseButtonDown", {_this call FUNC(onMouseButtonDown)}];
+    GVAR(MouseScrollHandle) = (findDisplay 46) displayAddEventHandler ["MouseZChanged", {_this call FUNC(onMouseScroll)}];
```
commy2: move these to a DisplayLoad event instead. Fixes your point 1.
```
class Extended_DisplayLoad_EventHandlers {
    class RscDisplayMission {
        ADDON = QUOTE(_this call COMPILE_FILE(XEH_missionDisplayLoad));
    };
};
```

addons/grenadethrowing/functions/fnc_draw3d.sqf
```diff
@@ -0,0 +1,116 @@
...
+private _newTrajASL = [];
+private _newTrajATL = [];
+GVAR(PathData) = [];
+
+for "_i" from 0.05 to 1.45 step 0.1 do {
+    private _dx = _v0x * _i;
+    private _dy = _v0z * _i - 4.905 * _i ^ 2;
+
+    _newTrajASL = [_sPosx + _flyDirSin * _dx, _sPosy + _flyDirCos * _dx, _sPosz + _dy];
+    _newTrajATL = ASLtoATL _newTrajASL;
```
esteldunedain: ASLtoAGL

addons/grenadethrowing/functions/fnc_pfh.sqf
```diff
@@ -0,0 +1,179 @@
...
+    if (GVAR(ThrowType) == "under") then {
+        _xAdjust_Bonus = -0.075;
+        _yAdjust_Bonus = 0.1;
+    };
+
+    GVAR(CameraOffset) set [0, (GVAR(CameraOffset) select 0) + GVAR(xAdjust) + _xAdjust_Bonus];
+    GVAR(CameraOffset) set [1, (GVAR(CameraOffset) select 1) + GVAR(yAdjust) + _yAdjust_Bonus];
+    GVAR(CameraOffset) set [2, (GVAR(CameraOffset) select 2) + GVAR(zAdjust)];
+
+    private _posFin = positionCameraToWorld GVAR(CameraOffset);
```
esteldunedain: `ASLtoAGL positionCameraToWorld GVAR(CameraOffset);`

addons/grenadethrowing/functions/fnc_initControls.sqf
commy2: Keep in mind that findDisplay 46 cannot be used at that point though. use:
`params ["_display"]`
instead

addons/grenadethrowing/functions/fnc_pfh.sqf
```diff
@@ -0,0 +1,179 @@
...
+    GVAR(ActiveGrenadeItem) setDir ((getDir player) + 90);
+    [GVAR(ActiveGrenadeItem), -30, 0] call BIS_fnc_setPitchBank;
+
+    if (GVAR(ThrowType) == "under") then {
+        [GVAR(ActiveGrenadeItem), -90, 0] call BIS_fnc_setPitchBank;
+    };
+
+    if (GVAR(CtrlHeld)) then {
+        if (vehicle player == player) then {
+            _posFin = positionCameraToWorld [0, 0, GVAR(TestPercArm)];
```
esteldunedain: `ASLtoAGL positionCameraToWorld [0, 0, GVAR(TestPercArm)];`

addons/grenadethrowing/functions/fnc_pfh.sqf
```diff
@@ -0,0 +1,179 @@
...
+            if (!(surfaceIsWater _posFin)) then {
+                _posFin = ATLToASL _posFin;
+            };
+
+            private _posView = positionCameraToWorld [0, 0, 0];
```
esteldunedain: `ASLtoAGL positionCameraToWorld [0, 0, 0];`

addons/grenadethrowing/functions/fnc_throw.sqf
```diff
@@ -0,0 +1,166 @@
...
+        private _p1 = AGLtoASL (GVAR(ActiveGrenadeItem) modelToWorldVisual [0, 0, 0]);
+
+        private _unitV = (vectorNormalized (_p1 vectorFromTo _p2)) vectorMultiply _velocity;
+
+        if (vehicle player == player) then {
+            // This method assumes the ability for a human to instinctively provide upper-body throw stabilization to prevent a grenade from being too influenced by how they're moving
+            _newVelocity = [0, 0, 0] vectorAdd _unitV;
+        } else {
+            // This method would be for things like the Littlebird throw-from-vehicles, where we have a vehicle-based velocity that can't be compensated for by a human
+            _newVelocity = [velocity (vehicle player), _unitV] call BIS_fnc_vectorAdd;
```
esteldunedain: `vectorAdd`

esteldunedain:
Also mass replace all `player` by `ACE_player`

addons/grenadethrowing/XEH_preInit.sqf
```diff
@@ -0,0 +1,38 @@
...
+GVAR(CancelThrow) = false;
+GVAR(ToggleThrowMode) = false;
+GVAR(ThrowGrenade) = false;
+GVAR(CookingGrenade) = false;
+
+GVAR(ActiveGrenadeItem) = objNull;
+GVAR(ActiveGrenadeType) = "";
+
+GVAR(CameraOffset) = [0, 0, 0.3];
+GVAR(zAdjust) = -0.03;
```
esteldunedain: Make this three a vector: `GVAR(adjust) = [-0.03, -0.05, -0.12];`

addons/grenadethrowing/functions/fnc_pfh.sqf
```diff
@@ -0,0 +1,179 @@
...
+    };
+
+    private _xAdjust_Bonus = 0;
+    private _yAdjust_Bonus = 0;
+    if (GVAR(ThrowType) == "under") then {
+        _xAdjust_Bonus = -0.075;
+        _yAdjust_Bonus = 0.1;
+    };
+
+    GVAR(CameraOffset) set [0, (GVAR(CameraOffset) select 0) + GVAR(xAdjust) + _xAdjust_Bonus];
```
esteldunedain: `GVAR(cameraOffset) = GVAR(cameraOffset) vectorAdd GVAR(adjust) vectorAdd [_xAdjust_Bonus, _yAdjust_Bonus, 0];`

addons/grenadethrowing/functions/fnc_isFFVAndWeaponUp.sqf
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private _return = false;
+
+if (vehicle player == player) exitWith {true};
+
+private _assignedRole = assignedVehicleRole player;
```
commy2: `assignedVehicleRole` returns what your AI group leader want's you to be and not our actual vehicle position.

addons/grenadethrowing/functions/fnc_throw.sqf
```diff
@@ -0,0 +1,166 @@
...
+        if (!GVAR(CtrlHeld)) then {
+            GVAR(ActiveGrenadeItem) setPosASL _posFin;
+        };
+
+        if (GVAR(CtrlHeld)) then {
+            _direction = [0, 200, 500];
+            _velocity = 3;
+        };
+
+        // TODO, make sure these are the right coordinates and don't mess up on buildings etc.
```
esteldunedain: This is alright. `positionCameraToWorld` return AGL coordinates

addons/grenadethrowing/functions/fnc_draw3d.sqf
```diff
@@ -0,0 +1,116 @@
...
+
+private _v0x = cos _viewAngle * _velocity;
+private _v0z = sin _viewAngle * _velocity;
+
+private _headOffest = call FUNC(getHeadOffset);
+
+private _throwDir = getDirVisual player + _headOffest;
+private _flyDirSin = sin _throwDir;
+private _flyDirCos = cos _throwDir;
+private _sPosx = (getPosASLVisual GVAR(ActiveGrenadeItem)) select 0;
```
esteldunedain: `_sPos = getPosASLVisual GVAR(ActiveGrenadeItem);`

addons/grenadethrowing/functions/fnc_draw3d.sqf
```diff
@@ -0,0 +1,116 @@
...
+
+private _newTrajASL = [];
+private _newTrajATL = [];
+GVAR(PathData) = [];
+
+for "_i" from 0.05 to 1.45 step 0.1 do {
+    private _dx = _v0x * _i;
+    private _dy = _v0z * _i - 4.905 * _i ^ 2;
+
+    _newTrajASL = [_sPosx + _flyDirSin * _dx, _sPosy + _flyDirCos * _dx, _sPosz + _dy];
```
esteldunedain: Replace by vector operations

addons/grenadethrowing/stringtable.xml
```diff
@@ -0,0 +1,14 @@
...
+<Project name="ACE">
+    <Package name="GrenadeThrowing">
+        <Key ID="STR_ACE_GrenadeThrowing_Category">
+            <English>Grenade Throwing</English>
+        </Key>
+        <Key ID="STR_ACE_GrenadeThrowing_EnableThrowingSystem_DisplayName">
+            <English>Enable Throwing</English>
+        </Key>
+        <Key ID="STR_ACE_GrenadeThrowing_EnableThrowingSystem_Description">
+            <English>Enables grenade throwing system.</English>
```
gienkov: >Enables **advanced** grenade throwing system

perhaps?

addons/grenadethrowing/XEH_postInitClient.sqf
```diff
@@ -0,0 +1,23 @@
...
+// Exit on HC
+if (!hasInterface) exitWith {};
+
+
+["SettingsInitialized", {
+    if (GVAR(Enabled)) then {
+        [true] call FUNC(init);
+    };
+
+    ["SettingChanged", {
```
esteldunedain: Technically there's no need to put this inside "SettingsInitialized". Both events can be handled by the same cose if moved to a local `_fnc_updateSetting`.

addons/grenadethrowing/functions/fnc_onMouseButtonDown.sqf
```diff
@@ -0,0 +1,46 @@
...
+    } else {
+        if (!GVAR(ThrowGrenade)) then {
+            GVAR(ThrowGrenade) = true;
+        };
+    };
+};
+
+if (_key == 2 && {GVAR(GrenadeInHand)} && {!GVAR(CookingGrenade)}) then {
+    [GVAR(ActiveGrenadeItem), GVAR(ActiveGrenadeType)] call FUNC(cook);
+    hint "Cooking!";
```
gienkov: Needs stringtabling

addons/grenadethrowing/functions/fnc_isFFVAndWeaponUp.sqf
esteldunedain: Also, don't we have a function in common to detect FFV positions?

addons/grenadethrowing/functions/fnc_pfh.sqf
```diff
@@ -0,0 +1,179 @@
...
+    GVAR(CameraOffset) set [1, (GVAR(CameraOffset) select 1) + GVAR(yAdjust) + _yAdjust_Bonus];
+    GVAR(CameraOffset) set [2, (GVAR(CameraOffset) select 2) + GVAR(zAdjust)];
+
+    private _posFin = positionCameraToWorld GVAR(CameraOffset);
+    if (!(surfaceIsWater _posFin)) then {
```
esteldunedain: The ASLtoAGL handles things correctly over water too, so this if should be removed.

addons/grenadethrowing/functions/fnc_pfh.sqf
```diff
@@ -0,0 +1,179 @@
...
+    [GVAR(ActiveGrenadeItem), -30, 0] call BIS_fnc_setPitchBank;
+
+    if (GVAR(ThrowType) == "under") then {
+        [GVAR(ActiveGrenadeItem), -90, 0] call BIS_fnc_setPitchBank;
+    };
+
+    if (GVAR(CtrlHeld)) then {
+        if (vehicle player == player) then {
+            _posFin = positionCameraToWorld [0, 0, GVAR(TestPercArm)];
+            if (!(surfaceIsWater _posFin)) then {
```
esteldunedain: Same here.

addons/grenadethrowing/functions/fnc_pfh.sqf
```diff
@@ -0,0 +1,179 @@
...
+                _posFin = ATLToASL _posFin;
+            };
+
+            private _posView = positionCameraToWorld [0, 0, 0];
+            if (!(surfaceIsWater _posView)) then {
```
esteldunedain: And here

addons/grenadethrowing/functions/fnc_onMouseButtonDown.sqf
jonpas: Needs to be replaced with ACE3 hint system as well.

addons/grenadethrowing/stringtable.xml
jonpas: I agree.

addons/grenadethrowing/functions/fnc_isFFVAndWeaponUp.sqf
jonpas: I thought we had as well, I will have a look.

addons/grenadethrowing/functions/fnc_initControls.sqf
jonpas: Will wait on fixing that when key events are fixed when used with this on CBA's side.

addons/grenadethrowing/functions/fnc_throw.sqf
```diff
@@ -0,0 +1,171 @@
...
+                _typeCount = _typeCount + 1;
+            };
+        } forEach (magazines _unit);
+
+        // Works around a bug where removing one will make them unselectable
+        if (_typeCount > 1) then {
+            TRACE_1("Removing Throwable (2)",_typeGrenCheck);
+            _unit removeMagazine _typeGrenCheck;
+            _unit removeMagazine _typeGrenCheck;
+            _unit addMagazine _typeGrenCheck;
```
commy2: pretty sure this was fixed years ago

addons/grenadethrowing/functions/fnc_onKeyDown.sqf
```diff
@@ -0,0 +1,119 @@
...
+    GVAR(ToggleThrowMode) = !GVAR(ToggleThrowMode);
+
+    // Throw mode is on
+    if (GVAR(ToggleThrowMode)) then {
+        GVAR(ThrowType) = "normal";
+        GVAR(AmmoLastMag) = ACE_player ammo (currentWeapon ACE_player);
+        ACE_player setAmmo [currentWeapon ACE_player, 0];
+        inGameUISetEventHandler ["PrevAction", "true"];
+        inGameUISetEventHandler ["NextAction", "true"];
+        inGameUISetEventHandler ["Action", "true"];
```
commy2: this, showHUD, common

addons/grenadethrowing/functions/fnc_throw.sqf
jonpas: Apparently it was not, just tried removing one with a debug console and all are gone from the vanilla throwing, can bring them back with the already existing cycle grenade keybinds from ACE though.

addons/grenadethrowing/functions/fnc_pfh.sqf
```diff
@@ -0,0 +1,167 @@
...
+
+    if (_exit) exitWith {
+        [_unit, "Trying to create a grenade we do not have in inventory (2)"] call FUNC(exitThrowMode);
+    };
+
+    GVAR(CameraOffset) = [0, 0, 0.3];
+    private _posLeftOffset = 0;
+    private _posRightOffset = 0;
+
+    if (inputAction "leanLeft" > 0.1) then {
```
jonpas: Replace with some other check to make it work when "lock leaning" as well (double press lean button).

esteldunedain: we could probably tilt the throw point based on the position of the head in relation to the hip position or something like than. That could probably be usefull for side stances too.

jonpas:
Does no funky stuff to weapon ammo anymore as it uses the action directly.

Changed controls slightly, mouse wheel up and down is change mod now, 2 right clicks is drop when cooked.

Dependent on `ace_weaponselect` due to using `selectNextGrenade` - needed because of switch to CBA key instead of vanilla "Cycle Throwables".

Some general cleanup as well, it's a start.

jonpas:
3rd person support added, leaning calculated from head position to make it work when lean is toggled and also adjust for side stances.

Someone with more math knowledge than me should probably go through the calculations to make them more robust.

esteldunedain:
@commy2, could you take a look at the bug with keyUp not firing? Might be related or not with your CBA updata, idk.

jonpas:
Throwing out of vehicles is not possible when there are no intersections with the vehicle. It will also hide the grenade object and fly arc. It doesn't respect head translation (most likely engine limitation as the head doesn't actually move, just the camera does) and drawing could be better in some cases, but overall it works pretty well.

Throwing from pilot seat is currently disabled, because `DefaultAction` doesn't get executed when `driver vehicle == unit`.

bux578:
> Throwing from pilot seat is currently disabled

How about throwing from copilot seat?

commy2:
>could you take a look at the bug with keyUp not firing? Might be related or not with your CBA updata, idk.

Should be fixed after events rewrite

jonpas:
@bux578 Yes that works, quite well from the little bird at least. With the Ghosthawk for example there is a problem of recognizing the invisible wall when you are sitting on one end of the pessenger compartment and try throwing out through the door on the other side, it just bounces of invisible wall.

@commy2 I am using latest CBA master and that still happens. The problem is not only what you described, but holding down Ctrl for some amount of time also won't trigger KeyUp, pressing in quick succession will without a problem.

jonpas:
@bux578 As requiested, arc drawing and mouse hint added to settings.

commy2:
- `addActionEventHandler` adds `confirmThrow` to LMB. Executed in scheduled env. because splendidTM
- `confirmThrow` calls `cook` (only when CTRL is held)
- `cook` calls global `throwFiredEH` event
- `throwFiredEH` global event executes all `cba_xeh_fired` events stored on the unit
- `cba_xeh_fired` calls Eds unified fired eh
- unifired fired eh calls `optics` `handleFired`
- `optics` `handleFired` stores a display in a local variable - still in scheduled env. -> Arma complains

All these are done in scheduled. I'll adjust `addActionEventHandler` with CBA_fnc_directCall to fix that. Will be another PR, because unrelated to this.

esteldunedain:
> Will be another PR, because unrelated to this.
#3560

Khaosmatic:
Is throwing back grenades something that is planned?

Because playing hot potato in ARMA would just be the biggest achievement ever!

jameslkingsley:
I thought about this a bit. You could have a check that makes sure you're close enough to the grenade and have it visible (not necessarily underneath the cursor, just in view) and then place it into the players hands as if it were his own grenade. Then if he's quick enough, he'd be able to throw it back.

jonpas:
I though about it as well and haf a discussion with Dslyecxi. We came to a conclusion it's more of "do you have enough guts to do it" IRL so it would certainly be a nice feature. However, doing an interaction on it would be a pain from technical side and also user-friendliness. I like the idea you proposed though, I will certainly look into it.

Khaosmatic:
So this is now back on track for 3.6.0 instead of 3.7.0?

jonpas:
Maybe, maybe not.

jonpas:
Moving to 3.7.0 as this needs some (not that little) more work.

Khaosmatic:
*Tiny violin*

At least it will work properly when it is release time. Very excited.

rebelvg:
Built this to try it out with my community this weekend. Any sense reporting if we find something or it's heavy WIP right now and things will change?

jonpas:
It's feature complete and pretty much finished. There are minor visual things that could be improved but I prefer usability over that. Any bugs please do report!

VKing6:
The throwing direction is tied to head direction, so at least with TrackIR you can do throws over your shoulder and the like.

jonpas:
> The throwing direction is tied to head direction, so at least with TrackIR you can do throws over your shoulder and the like.

This is intentional, however I can see that you couldn't throw with the same power in that case, I will add some limit to that or something. If not sooner than after first release of the system.

SilentSpike:
Maybe a `[0,1]` multiplier derived from the angle between the head direction and body direction to limit the throwing power (with a `(theta-30) max 0` somewhere in there to make anything in a 60 degree arc straight ahead maximum power).

SilentSpike:
Well, I suppose `[0.5,1]` since you'd never want it to be really low

SilentSpike:
Quick mock up example, totally untested:
```
_heading = eyeDirection _unit;
_heading = abs((_heading select 0) atan2 (_heading select 1));
_power = linearConversion [0, 180, _heading - 30, 1, 0.5, true];
```

jonpas:
Heh, good try, but you are not taking into account that the character can rotate as well, so `_power` will be different without moving the head at all.

rebelvg:
I think some kind of cool down is needed between throws. When grenades turn red in the UI they kinda get blocked by the engine or whatever but you still can throw them via this module. This leads to sometimes a really strange behavior, you can still have grenades in the inventory but they will not be shown by the UI in the corner and you will not be able to throw them via advanced throwing.
Same goes for reloading, if you throw the grenade mid reloading (even after latest commit the uses reload mutex), grenades get stuck in your inventory sometimes.

SilentSpike:
> Heh, good try, but you are not taking into account that the character can rotate as well, so _power will be different without moving the head at all.

I thought `eyeDirection` returned a vector representing the head direction relative to the body?

jonpas:
@SilentSpike no, it's world absolute, same as most other vectors.

SilentSpike:
Damn biki page lying to me :stuck_out_tongue_closed_eyes:

SilentSpike:
Tested code:
```
_unit = player; 
_unitDir = getDir _unit; 
_cameraDir = getCameraViewDirection _unit; 
_cameraDir = (_cameraDir select 0) atan2 (_cameraDir select 1); 
 
_phi = abs(_cameraDir - _unitDir) % 360; 
_phi = [_phi, 360 - _phi] select (_phi > 180); 

_power = linearConversion [0, 180, _phi - 30, 1, 0.3, true];
```

At the extremes the power drops to about 50%:

| `_phi` | `_power` |
| 0 | 1 |
| 15 | 1 |
| 30 | 1 |
| 45 | 0.941667 |
| 60 | 0.883333 |
| 75 | 0.825 |
| 90 | 0.766667 |
| 105 | 0.766667 |
| 120 | 0.65 |
| 135 | 0.591667 |
| 150 | 0.533333 |

I have no idea how the maths is working on those two angles since they're not normalised to the same range, but it definitely is :laughing:

jonpas:
Works brilliantly @SilentSpike !

rebelvg:
Also had a bug today, arma said I had 0 grenades in the UI (screenshot provided) but I could throw unlimited number of grenades and I didn't have any in the inventory. Have no idea how to reproduce. http://i.imgur.com/mh7XQ9Q.jpg

jonpas:
Interesting, the delay may fix that, but I don't know what the cause could be.

jonpas:
I added a 2 second delay, seems to be fairly close to what vanilla does.

rebelvg:
Just tested this, cool down works fine. But, I still managed to kill the grenades by throwing a grenade mid reloading (very easy to reproduce on a weapon with a long reload animation like a MG). What happened is that grenades disappeared from the UI section and I couldn't throw any even though I had 20 of them in the inventory. Here's a video. https://www.youtube.com/watch?v=PYR7xxhjoYw&feature=youtu.be

jonpas:
@rebelvg I can't reproduce that all, tried it with Navid (Apex weapon) and I could throw 3 of them during one reload just fine. Either way, I just disallowed preparing a throwable if currently reloading, that should fix it either way then. I don't see why it wouldn't let you cycle to the grenade in your inventory anymore though as I am using the same function as is used by the `weaponselect` component.

rebelvg:
@jonpas That's strange. I'll try this latest fix and report if anything bizarre happens.

jonpas:
Thanks for helping test this, very appreciated! :)

rebelvg:
Had an arma session yesterday. All things I mentioned are fixed now, didn't have any problems with the module. Amazing job, jonpas! One small thing, interaction point on pickup items sometimes shows but you can't pickup the object, you have to move a bit closer for it to work.

jonpas:
> One small thing, interaction point on pickup items sometimes shows but you can't pickup the object, you have to move a bit closer for it to work.

Can't seem to reproduce, was it any specific grenade, specific circumstances (your grenade or someone else's if multiplayer)?

> And maybe some delay is need on grenade switch, engine paints grenade UI red when you switch between throwables (time when it's red is different for different grenades for some reason).

That prevents you from quickly taking what you need though, I don't like the idea of the delay here. It seems to work fine, it's just red color, not exactly sure why it happens though.

rebelvg:
> Can't seem to reproduce, was it any specific grenade, specific circumstances (your grenade or someone else's if multiplayer)?
Possible it wasn't my grenade. I'll report if this happens again with more details, haven't seen this happen again.

> That prevents you from quickly taking what you need though, I don't like the idea of the delay here. It seems to work fine, it's just red color, not exactly sure why it happens though.
I meant a delay on when you try to throw, not between switching grenades. Anyway, I tested this and didn't find any problems with it so I guess it can be ignored.

commy2:
>It seems to work fine, it's just red color, not exactly sure why it happens though.
It's the `ThrowMuzzle`s `magazineReloadTime` (2.5).
Note that the `reloadTime` (0) of the muzzle refers to the delay between single shots of the __same__ magazine.

It's basically a bug that has been in A2 and will probably be in A3

commy2:
This should remove that delay:
```
class CfgWeapons {
    class GrenadeLauncher;
    class Throw: GrenadeLauncher {
        class ThrowMuzzle: GrenadeLauncher {
            magazineReloadTime = 0;
        };
    };
};
```

jonpas:
It doesn't happen when selecting grenade with vanilla though. And it causes no issues, so I won't be editing that config unnecessarily.

commy2:
I'm talking about the annoying delay when switching between different grenade types you have in vanilla. Maybe that is something else though.

jonpas:
That delay is not present with advanced throwing at all, so if we decide to remove that from vanilla throwing it should go into separate PR and into `grenades` component.

commy2:
Oh, nvm then.

jonpas:
Although that may be the cause of infinite throwables, I'll have to test that, just thought of it.

bux578:
When in drop mode, you're able to "extend" the grenade distance through a closed window (through the glass).

jonpas:
Yes, that is a limitation of the engine. There is no proper way to detect glass. You can see with vanilla grenades, you can throw them through those glass windows perfectly fine.

addons/advancedthrowing/functions/fnc_drawThrowable.sqf
```diff
@@ -15,15 +15,27 @@
...
+// Exit if throwable died primed in hand
+if (isNull _activeThrowable && {_primed}) exitWith {
+    [ACE_player, "Throwable died primed in hand"] call FUNC(exitThrowMode);
+};
+
+private _throwable = currentThrowable ACE_player;
+
+// Inventory check
+if (_throwable isEqualTo [] && {!_primed}) exitWith {
+    [ACE_player, "No valid throwables"] call FUNC(exitThrowMode);
```
bux578: "No valid throwables" -> Stringtable? (or is this just log?)

PabstMirror: It's just for debugging, won't ever be shown to user.

jonpas: Yep, just log.

commy2:
:tada:

BlackburnUTG:
What do I do, if I want to set  buton  to cancel Advanced Throwing? Where  can I do it?

jonpas:
@BlackburnUTG you can't rebind the cancel button. And please don't revive merged pull requests, join [our Slack](http://slackin.ace3mod.com/) for questions.

PR #3497, opened 2016-03-02
Fix portable light floating / dir when carrying

### LINKED ISSUE #3486: Portable lights (and other carriable objects) float when released in midair.
When you release an object that you're carrying in midair, it will fall to the ground only if it's not a building; otherwise it will float in mid air.

Add a check on release. If the object doesn't support physX, teleport it downward. Probably detect ground position with lineIntersectSurfaces.

PabstMirror:

Fix #3486
 - Set "house" type objects to have a `getPos` hegiht of 0.
 - Also update `carryDirection` after each re-attach from mouse wheel scrolling


esteldunedain:
:+1:

Issue #3499, opened 2016-03-02
Explosives M112. I can not change the menu and twist it.


Kartaviy:
**Arma 3 Version:** `1.56`

**ACE3 Version:** `3.5.0`

**Mods:**
* `@CBA_A3`
* `@ace`
* @ RHS

**Placed Modules:**
* *No modules.*

**Description:**
* *https://youtu.be/z1Xa0_5hWIo*

**Where did the issue occur?**
* *"Editor"*

**RPT log file:**
* *Add a link (pastebin.com) to the client and/or server RPT file. An instruction to find your RPT files can be [found here](https://community.bistudio.com/wiki/Crash_Files#Arma_3)*

Kartaviy:
With the same trenches

commy2:
requires RPT file

Kartaviy:
Well, what else can I show? Mina lowered to the ground I can not twist them, I can not either. Cancel operation, I can not either. When digging the trench and press ESC, the cancellation does not occur.

jonpas:
@commy2 it's the addAction hook issue we were talking about yesterday.

esteldunedain:
> @commy2 it's the addAction hook issue we were talking about yesterday.

I don't think it is

gienkov:
I can repro this.
02.03.2016 12:31 PM "Kartaviy" <notifications@github.com> napisał(a):

esteldunedain:
Please somebody upload a RPT log.

Kartaviy:
rpt - http://pastebin.com/h9bWsAqC

jonpas:
`ACE_Modifier` is always `0`, that's the cause, known CBA issue.

esteldunedain:
@Kartaviy, try rebinding the ACE modifier key once. That should prevent the bug in cba and make it work again.

This will probably be fixed soon in CBA.

jonpas:
And rotating objects will be fixed with #3507

Radeni:
This is the same issue that i had in past which i submitted but it turned out that it was only the issue in a specific mission so i tought it wasnt Ace's problem but mine.Im glad that this is going to get changed. the whole Mouse buttons and wheel system should be replaced with keybinds.

commy2:
[...]

PR #3507, opened 2016-03-02
Replace addScrollWheelEventHandler with MouseZChanged Display EH

jonpas:
- Deprecate `EFUNC(common,addScrollWheelEventHandler)`
- Use `MouseZChanged` instead of `EFUNC(common,addScrollWheelEventHandler)`
- Fixes part of #3499 

Needs some testing to make sure it all works.
Also make sure to test with `ACE_Modifier = 1` as that has it's own problems.

### DIFF
--- a/addons/common/functions/fnc_addScrollWheelEventHandler.sqf
+++ b/addons/common/functions/fnc_addScrollWheelEventHandler.sqf
@@ -15,6 +15,8 @@
 
 params ["_statement"];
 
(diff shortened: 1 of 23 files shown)


commy2:
Looks good, but eventually I'd like to replace these with `CBA_fnc_addDiplayHandler` to reduce the amount of compiles.
After it is fixed to work with save games and resumed games that is.

jonpas:
Yep, we can do that once it works, I just wanted to get it fixed for 3.5.1, still have to test all modules work as intended though.

PR #3508, opened 2016-03-02
Make the FRIES inherit from Helicopter_Base_F. Fix #3504

### LINKED ISSUE #3504: AI ignoring waypoints when FRIES is enabled/loaded
#**Arma 3 Version:** `1.56` 
**CBA Version:** `2.3.1` 
**ACE3 Version:** `3.5.0` 

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- When a vanilla helicopter is loaded with the FRIES system (both when selected in the attributes (3DEN) as with the FRIES module), the AI ignores the set waypoints and just goes straight up into the air.

**Steps to reproduce:**
- Place a Ghosthawk, open attributes and tick FRIES at the bottom or place the FRIES module and link it to the ghosthawk, give the ghosthawk waypoints, place an AI (Player) next to the helicopter. Launch the mission, board the helicopter and see what happens.

- 3DEN Editor (Singleplayer)

**Placed Modules:**
- Equip FRIES

**RPT log file:**
- N/A
- N/A

esteldunedain:
- Make the FRIES inherit from Helicopter_Base_F
- Fix #3504

I've tested it and received no UBC errors. Helicopters follow waypoints properly.

Anyway, I don't ever trust my config work, so some review would be welcome.

EDIT: tested fastroping a few times using this FRIES and it work fine, although a couple of times my unit fell the last couple of meters. I don't think it's related though.

### DIFF
--- a/addons/fastroping/CfgVehicles.hpp
+++ b/addons/fastroping/CfgVehicles.hpp
@@ -12,12 +12,69 @@
 }
 
 class CfgVehicles {
-    class Building;
-    class NonStrategic: Building {
+    class Logic;
+    class Module_F: Logic {
+        class ModuleDescription;
+    };
+    class ACE_Module: Module_F {};
+    class ACE_moduleEquipFRIES: ACE_Module {
+        scope = 2;
+        displayName = CSTRING(Module_FRIES_DisplayName);
+        icon = QUOTE(PATHTOF(UI\Icon_Module_FRIES_ca.paa));
+        category = "ACE";
+        function = QUOTE(FUNC(moduleEquipFRIES));
+        functionPriority = 10;
+        isGlobal = 0;
+        isTriggerActivated = 0;
+        isDisposable = 0;
+        author = "BaerMitUmlaut";
+
+        class ModuleDescription: ModuleDescription {
+            description = CSTRING(Module_FRIES_Description);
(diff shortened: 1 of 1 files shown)


esteldunedain:
@Gwynblade, I opted to close your ticket with this fix. Please check it out. 

Even if this PR doesn't change the fact that the distance is computed to the eye pos, the issue in your video does not longer occur while moving the mouse toward the explosive because of the interaction distance being large enough. (Normal eyepos in arma is 1.52m above the ground)

Gwynblade:
I figured the increased range might alleviate the original problem a bit, but if it disappeared, even better, thanks man, that was fast!

esteldunedain:
Obviously I commented on the wrong PR. The correct one is #3509

commy2:
I don't see how that would change the behaviour of the AI. Is that tested?
I think the AI tries to avoid collisions with other "helicopter" simulation type objects. Unfortunately the rope helper has to be a "helicopter" simulation type too for `RopeX` to work with it.

esteldunedain:
> I don't see how that would change the behaviour of the AI. Is that tested?

It is

commy2:
There is no inheritance needed in the first place. The fast rope items use no animation or ACE action of their base classes anyway.

PR #3509, opened 2016-03-02
Increase interaction distance with explosives from 1m to 1.5m. 

### LINKED ISSUE #3379: Explosives are cumbersome to interact with while standing
**Arma 3 Version:** `1.56`

**ACE3 Version:** `3.4.2`

**Mods:**
* `@CBA_A3`
* `@ace`

**Placed Modules:**
None

**Description:**
It looks like the ability to interact with something depends on proximity of player's head (or upper body) to object you're trying to interact with. With default ace interaction (show cursor disabled) it causes problems, since your entire body moves while interacting (left - right / up and down). Which causes you to appear out of range, and interaction gets aborted. Even though you're close enough to original interaction point. 

Check the video below to understand better what I'm trying to say.

**Steps to reproduce:**
https://www.youtube.com/watch?v=-5PyrVKUA_Q&feature=youtu.be

Everywhere

esteldunedain:
- Increase interaction distance with explosives from 1m to 1.5m. 
- Fix #3379

It's a nice usability enhancement, that makes placing explosives quicker and more streamlined.

### DIFF
--- a/addons/explosives/CfgVehicles.hpp
+++ b/addons/explosives/CfgVehicles.hpp
@@ -56,7 +56,7 @@ class CfgVehicles {
         class ACE_Actions {
             class ACE_MainActions {
                 selection = "";
-                distance = 1.5;
+                distance = 1;
                 condition = "true";
                 class ACE_Defuse {
                     displayName = CSTRING(Defuse);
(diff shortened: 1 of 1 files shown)


jonpas:
:+1:

esteldunedain:
Normal eyepos in arma is 1.52m, so the new distance prevents the issue that @Gwynblade described from happening at all and should allow confortable interaction.

esteldunedain:
Interaction distance for defusing was reduced again to 1m by @bux578 's request

bux578:
:kissing:

PR #3513, opened 2016-03-02
Add CBA 2.3.1 to versioning check

PabstMirror:
Will require https://github.com/CBATeam/CBA_A3/pull/290/

### DIFF
--- a/addons/common/functions/fnc_checkFiles.sqf
+++ b/addons/common/functions/fnc_checkFiles.sqf
@@ -19,6 +19,18 @@ private _version = getText (configFile >> "CfgPatches" >> "ace_main" >> "version
 
 ACE_LOGINFO_1("ACE is version %1.",_version);
 
+//CBA Versioning check - close main display if using incompatible version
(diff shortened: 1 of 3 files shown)


esteldunedain:
:+1:

Glowbal:
What will this do if someone loads an incorrect CBA version?

Are those minimum versions or does it have to match the exact version?

PabstMirror:
Just the min version, so 2.3.2 or 3.0.0 would be fine.
It just puts a little systemChat warning.
Do we want a stronger warning, like closeDisplay on a ace version mismatch?

bux578:
CheckPBO outputs a nice red box in the center of the screen.

jonpas:
What @bux578 said.

PabstMirror:
Now a hard stop if using an older version of CBA.
If we're counting on new stuff like `cba_missionTime`, there is no point trying to run on old versions of cba.

commy2:
:+1: 
This is brilliant. Old glory of A2 is coming back.

Glowbal:
Perfect, thanks @PabstMirror :+1:

PR #3516, opened 2016-03-02
replace timePFH with CBA_missionTime

commy2:
- remove timePFH
- Replace ACE_time with CBA_missionTime (about the same, but CBA is synched in MP)
- Replace ACE_diagTime with diag_tickTime (They are the same; need to change some stuff around for save games)
- ACE_gameTime with CBA_missionTime. ACE_gameTime was very close to ACE_time, but used tick time instead. Still wasn't synched in MP

REQUIRES CBA update.


PabstMirror:
`ACE_diagTime` variable was constant during the entire frame, `diag_tickTime` is a live command. But I don't see how this would effect anything.

These were never used anywhere `ACE_realTime`, `ACE_paused`, `ACE_pausedTime`, `ACE_virtualPausedTime`

commy2:
Good point. I will have to review any tickTime anyway.

Glowbal:
- ACE_time might be used by third parties. Should not be removed without proper communication about it going to be removed.
- This is a significant change. What are the effects of using a synced across MP time instead of a local variant? How does the CBA time wokr? What happens if a client has desync, does the time still increment?

esteldunedain:
> ACE_time might be used by third parties. Should not be removed without proper communication about it going to be removed.

I recommend we keep `ACE_time` around for 3.6.0 and deprecate it later. We might want to simply add a pfh that does `ACE_time = CBA_missionTime`.

> This is a significant change. What are the effects of using a synced across MP time instead of a local variant? How does the CBA time wokr? What happens if a client has desync, does the time still increment?

@Glowbal, based on my understanding of both `ACE_time` and `CBA_missionTime` I don't think your concerns are warranted. I recommend we discuss this further on Slack.

addons/common/XEH_preInit.sqf
```diff
@@ -53,19 +53,21 @@ GVAR(statusEffect_isGlobal) = [];
...
+            private _oldPlayer = ACE_player;
+
+            ACE_player = call FUNC(player);
+            uiNamespace setVariable ["ACE_player", ACE_player];
+
+            // Raise ACE event
+            ["playerChanged", [ACE_player, _oldPlayer]] call FUNC(localEvent);
+        };
+    }, 0, []] call CBA_fnc_addPerFrameHandler;
+};
```
commy2: rm

esteldunedain:
Great. Welcome CBA 2.4!

PR #3524, opened 2016-03-03
Fix Cancel (RMB) Action

jonpas:
- FIxes "Cancel" action which is right mouse button when placing things

~~**Untested** (no access to Arma-capable machine at the moment)~~

### DIFF
--- a/addons/explosives/functions/fnc_setupExplosive.sqf
+++ b/addons/explosives/functions/fnc_setupExplosive.sqf
@@ -141,6 +141,13 @@ GVAR(TweakedAngle) = 0;
     //Don't allow placing in a bad position:
     if (_badPosition && {GVAR(placeAction) == PLACE_APPROVE}) then {GVAR(placeAction) = PLACE_WAITING;};
 
+    if (_unit != ACE_player ||
+        {!([_unit, objNull, ["isNotSwimming"]] call EFUNC(common,canInteractWith))} ||
+        {!(_magClassname in (magazines _unit))}
+    ) then {
+        GVAR(placeAction) = PLACE_CANCEL;
+    };
+
     if (GVAR(placeAction) != PLACE_WAITING) then {
         [_pfID] call CBA_fnc_removePerFrameHandler;
         GVAR(pfeh_running) = false;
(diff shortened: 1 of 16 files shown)


jonpas:
Tested, seems to work fine. Tagging @commy2 for review.

commy2:
Okay this works as far as I can tell.
But.
We should probably add simple sanity checks after the != -1 check

previously these actions would only be called when the current action was active (e.g. carrying a ladder), but now all of them are executed when you RMB.

e.g.:
`_magClassname in (magazines _unit)`

addons/explosives/functions/fnc_setupExplosive.sqf
```diff
@@ -141,13 +141,6 @@ GVAR(TweakedAngle) = 0;
     //Don't allow placing in a bad position:
     if (_badPosition && {GVAR(placeAction) == PLACE_APPROVE}) then {GVAR(placeAction) = PLACE_WAITING;};
 
-    if (((inputAction "zoomTemp") > 0) || //Cancel on RMB, For some reason this works (when held) but AddActionEventHandler doesn't
-            {_unit != ACE_player} ||
-            {!([_unit, objNull, ["isNotSwimming"]] call EFUNC(common,canInteractWith))} ||
-            {!(_magClassname in (magazines _unit))}) then {
-        GVAR(placeAction) = PLACE_CANCEL;
```
PabstMirror: We don't want to remove this entire if statement, just the RMB part.
All of the other checks need to be run each frame in the PFEH so it will cancel if any of them happen.
e.g. if `canInteractWith` is false, we need to cancel right away instead of waiting for RMB

PabstMirror:
I wonder if we could simplify the event handler code to a single common
`EFUNC(common,rmbPressed) = true;`

and then in all our placement blocks we do
```
EFUNC(common,rmbPressed) = false
PFEH: if (EFUNC(common,rmbPressed)) then {cancle};
```

commy2:
there is no RMB release event though.

esteldunedain:
I still think installing only one of this EH and launching an ACE event on RMB would be more efficient and easier to use.

jonpas:
@esteldunedain I believe this is supposed to be solved with `CBA_fnc_addDisplayHandler`. @commy2 can confirm?

commy2:
Yes.

PR #3527, opened 2016-03-03
Fix empty callback for hr adjustment

### LINKED ISSUE #3526: fnc_addHeartRateAdjustment.sqf error when using morphine
Arma 3 Version: 1.56.134787
ACE3 Version: 3.5.0

Mods:
CBA
ACE
ALIVE

Description:
I get [THIS](http://i.imgur.com/Gt92kp1.jpg) when I apply morphine on myself in the Editor (Singleplayer).

Placed Modules:
ACE_ModuleMedicalSettings
ACE_ModuleBasicMedicalSettings
ACE_ModuleAdvancedMedicalSettings
ACE_ModuleAssignMedicRoles
ACE_Weather_ModuleSettings
ACE_Spectator_ModuleSettings
ACE_ModuleSitting
ACE_finger_ModuleSettings
ACE_ModuleNameTags
ACE_Map_gestures_ModuleSettings

PabstMirror:
Fix #3526
Same issue as #3384

### DIFF
--- a/addons/medical/functions/fnc_treatmentAdvanced_medicationLocal.sqf
+++ b/addons/medical/functions/fnc_treatmentAdvanced_medicationLocal.sqf
@@ -64,6 +64,7 @@ if (isNil _hrCallback) then {
 } else {
     _hrCallback = missionNamespace getVariable _hrCallback;
 };
+if (!(_hrCallback isEqualType {})) then {_hrCallback = {TRACE_1("callback was NOT code",_hrCallback)};};
 
 // Adjust the heart rate based upon config entry
 _heartRate = _target getVariable [QGVAR(heartRate), 70];


grester:
I know this has been closed but I'd like to point out that I'm not sure if this is indeed a bug or instead mod conflict. I say this because I have some simple missions where this error doesn't show up but I have some other missions with more scripts where the error does show up for all injections not just morphine. Just a heads up.

commy2:
Would be helpful to know which mod or mission causes the conflict.

grester:
I have yet had success in tracking down what's causing it in my case. However I'd like to point out though that despite the error the injection is still applied successfully and it takes effect on both pulse and pressure.

PabstMirror:
I have seen this with just CBA/ACE. I'm fairly sure it's a BIS thing.

`x = missionNamespace getVariable ""`
`typeOf x` = "EmptyDetector"

There are no medications that have a real callback, they all use `hrCallback = "";`
We could use something more explicit like `hrCallback = "#none", but I think this fix is good enough.

PR #3529, opened 2016-03-03
Throw a globalEvent when a tag is created. 

### LINKED ISSUE #3521: Add Tagging Event
When tagging something trigger an event to which mission makers can subscribe. Probably passing the tag, position, color, object as parameters.

Pinging random people: @BaerMitUmlaut @esteldunedain

esteldunedain:
- Throw a globalEvent when a tag is created. Useful for mission makers.
- Close #3521.

Parameters are:
```
0: Tag object, which can be used to get the position, orientation, etc <OBJECT>
1: Color ("black", "red", "green", "blue") <STRING>
2: Object that was tagged (objNull if it was the floor) <OBJECT>
3: Unit that created the tag <OBJECT>
```

@bux578, will you handle documentation?

### DIFF
--- a/addons/tagging/functions/fnc_createTag.sqf
+++ b/addons/tagging/functions/fnc_createTag.sqf
@@ -24,7 +24,7 @@ params ["_tagPosASL", "_vectorDirAndUp", "_color", "_object", "_unit"];
 TRACE_5("createTag:", _tagPosASL, _vectorDirAndUp, _color, _object, _unit);
 
 if !((toLower _color) in ["black", "red", "green", "blue"]) exitWith {
-    ACE_LOGERROR_1("%1 is not a valid tag colour.", _color);
+    ACE_LOGERROR_1("%1 is not a valid tag color.", _color);
 };
 
 private _tag = "UserTexture1m_F" createVehicle [0,0,0];
@@ -33,7 +33,7 @@ _tag setPosASL _tagPosASL;
 _tag setVectorDirAndUp _vectorDirAndUp;
 
 // Throw a global event for mision makers
-["tagCreated", [_tag, _color, _object, _unit]] call EFUNC(common,globalEvent);
+[QGVAR(tagCreated), [_tag, _color, _object, _unit]] call EFUNC(common,globalEvent);
 
 if (isNull _object) exitWith {};
 
--- a/addons/tagging/functions/fnc_createTag.sqf
+++ b/addons/tagging/functions/fnc_createTag.sqf
@@ -7,6 +7,7 @@
  * 1: Vector dir and up <ARRAY>
  * 2: Colour of the tag (valid colours are black, red, green and blue) <STRING>
  * 3: Object it should be tied too <OBJECT>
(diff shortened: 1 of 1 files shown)


bux578:
Will do

addons/tagging/functions/fnc_createTag.sqf
```diff
@@ -19,8 +20,8 @@
 
 #include "script_component.hpp"
 
-params ["_tagPosASL", "_vectorDirAndUp", "_color", "_object"];
-TRACE_4("createTag:", _tagPosASL, _vectorDirAndUp, _color, _object);
+params ["_tagPosASL", "_vectorDirAndUp", "_color", "_object", "_unit"];
+TRACE_5("createTag:", _tagPosASL, _vectorDirAndUp, _color, _object, _unit);
 
 if !((toLower _color) in ["black", "red", "green", "blue"]) exitWith {
     ACE_LOGERROR_1("%1 is not a valid tag colour.", _color);
```
bux578: `colour.", _color`

örgh

addons/tagging/functions/fnc_createTag.sqf
```diff
@@ -31,6 +32,9 @@ _tag setObjectTextureGlobal [0, '\z\ace\addons\tagging\UI\tags\' + _color + '\'
 _tag setPosASL _tagPosASL;
 _tag setVectorDirAndUp _vectorDirAndUp;
 
+// Throw a global event for mision makers
+["tagCreated", [_tag, _color, _object, _unit]] call EFUNC(common,globalEvent);
```
bux578: Do we have a convention on how to name events?
Maybe QGVAR it?

jonpas: We don't yet, but we should do QGVAR.

esteldunedain: Yes, QGVAR is something we'll have to do, specially if we switch to CBA events.

The bad thing is that we are not doing it almost anywhere atm.

I'll change it here though

jonpas: It will require some time to get it moved, we should do it the "deprecation" way like with functions.

esteldunedain:
@bux578, event renamed to "ace_tagging_tagCreated"

Glowbal:
Why do we all of a sudden have an ACE prefix for this event? I don't think any of the other events we use have such a prefix. We are not yet using a CBA event framework that I know off, so I do not see the reasons to prefix this event.

If you are going with "future proof it", in case we do switch over to a system were we do require to prefix all our events, every third party has to double check any event used anyway. No benefit in doing it now.

esteldunedain:
> I don't think any of the other events we use have such a prefix.

Missile_Guidance, Explosives and Mk6 use prefixed event names. The rest of modules don't.

I'm fine either way. Let's discuss it this afternoon/evening.

esteldunedain:
The event name is back to being unprefixed, for consistency with most other current event names.

PR #3534, opened 2016-03-04
Disallow refueling with running engine

GitHawk:
Extending #2954 by disallowing refueling with a running engine in both canTurnOn and the refuel PFH

### DIFF
--- a/addons/refuel/functions/fnc_canTurnOn.sqf
+++ b/addons/refuel/functions/fnc_canTurnOn.sqf
@@ -27,4 +27,5 @@ if (isNull _unit  ||
 !(_nozzle getVariable [QGVAR(isRefueling), false]) &&
     {[_nozzle getVariable QGVAR(source)] call FUNC(getFuel) != 0} &&
     {!isNull (_nozzle getVariable [QGVAR(sink), objNull])} &&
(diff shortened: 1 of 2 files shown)


bux578:
@GitHawk I had a talk about that with a Bundeswehr soldier. The bigger German vehicles all have to have their engine running for refueling. Maybe it might be better to check for civilian cars / light cars to have their engine off?

GitHawk:
I just want to avoid collisions between increasing the amount and ArmA decreasing it.

esteldunedain:
> I just want to avoid collisions between increasing the amount and ArmA decreasing it.

That shouldn't be a concern given how slowly fuel is consumed.

GitHawk:
I personally don't care. I can remove all checks if you like @bux578 @esteldunedain 

Technically you can refuel a real car/truck/tank with a running engine.
At civil fuel stations it is frowned upon (in some places maybe even illegal) to have the engine running while refueling.

Khaosmatic:
You should make it so if the engine is left on it explodes with a megaton blast radius. Dat'll teech dem kids.

sargken:
You should make it a random chance that it would catch fire or something like that.

PR #3542, opened 2016-03-05
Replace execRemoteFnc with events and increase fuel update delay in refuel

### LINKED ISSUE #3541: Replace execRemoteFnc with globalEvent and reduce to every 1s in refuel
```
[ACE] (common) WARNING: ace_common_fnc_execRemoteFnc is deprecated. Support will be dropped in version 3.7.0. Replaced by: ace_common_fnc_globalEvent
```

Possibly in other places as well, but refule is priority because it's a bunch of events in quick succession during refueling.

If possible, also make the globalEvent be triggered every 1 second only, to reduce network spam.

Tagging @GitHawk

jonpas:
- Closes #3541 
- Replaces all `execRemoteFnc` with `globalEvent` or `objectEvent`
- Increases refuel PFH to every 1 second (refuel rate stays the same as it's part of the pre-PFH calculation)
- Add `setVectorDirAndUp` and `setHitPointDamage` functions to common (both require object to be local)

Tagging @GitHawk for review. Tested locally only.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -158,6 +158,9 @@ if (isServer) then {
     ["enableSimulationGlobal", {(_this select 0) enableSimulationGlobal (_this select 1)}] call FUNC(addEventHandler);
 };
 
+["setVectorDirAndUp", {(_this select 0) setVectorDirAndUp (_this select 1)}] call FUNC(addEventHandler);
+["setHitPointDamage", {(_this select 0) setHitPointDamage (_this select 1)}] call FUNC(addEventHandler);
+
 
 //////////////////////////////////////////////////
 // Set up remote execution
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -153,7 +153,7 @@ if (isServer) then {
 ["playMoveNow", {(_this select 0) playMoveNow (_this select 1)}] call FUNC(addEventHandler);
 ["switchMove", {(_this select 0) switchMove (_this select 1)}] call FUNC(addEventHandler);
 ["setVectorDirAndUp", {(_this select 0) setVectorDirAndUp (_this select 1)}] call FUNC(addEventHandler);
-["setHitPointDamage", {(_this select 0) setHitPointDamage (_this select 1)}] call FUNC(addEventHandler);
+["setVanillaHitPointDamage", {(_this select 0) setHitPointDamage (_this select 1)}] call FUNC(addEventHandler);
 
 if (isServer) then {
     ["hideObjectGlobal", {(_this select 0) hideObjectGlobal (_this select 1)}] call FUNC(addEventHandler);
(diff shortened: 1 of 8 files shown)


GitHawk:
Looks good for me

jonpas:
Tested locally yes, but not multiplayer.

addons/common/XEH_postInit.sqf
```diff
@@ -158,6 +158,9 @@ if (isServer) then {
     ["enableSimulationGlobal", {(_this select 0) enableSimulationGlobal (_this select 1)}] call FUNC(addEventHandler);
 };
 
+["setVectorDirAndUp", {(_this select 0) setVectorDirAndUp (_this select 1)}] call FUNC(addEventHandler);
```
esteldunedain: Maybe moving this lines upward 6~7 lines would be better, so they are placed along with the rest of similar handlers.

addons/common/XEH_postInit.sqf
```diff
@@ -152,6 +152,8 @@ if (isServer) then {
 ["playMove", {(_this select 0) playMove (_this select 1)}] call FUNC(addEventHandler);
 ["playMoveNow", {(_this select 0) playMoveNow (_this select 1)}] call FUNC(addEventHandler);
 ["switchMove", {(_this select 0) switchMove (_this select 1)}] call FUNC(addEventHandler);
+["setVectorDirAndUp", {(_this select 0) setVectorDirAndUp (_this select 1)}] call FUNC(addEventHandler);
+["setHitPointDamage", {(_this select 0) setHitPointDamage (_this select 1)}] call FUNC(addEventHandler);
```
GitHawk: I would add `disableCollisionWith` as well. Useful for every attached/carried object so you won't hurt other players.

addons/refuel/functions/fnc_checkFuel.sqf
```diff
@@ -27,9 +27,9 @@ private _fuel = [_target] call FUNC(getFuel);
         params ["_args"];
         _args params [["_unit", objNull, [objNull]], ["_target", objNull, [objNull]], ["_fuel", 0, [0]]];
         if (_fuel > 0 ) then {
-            ["displayTextStructured", [_unit], [[LSTRING(Hint_RemainingFuel), _fuel], 2, _unit]] call EFUNC(common,targetEvent);
+            ["displayTextStructured", _unit, [[LSTRING(Hint_RemainingFuel), _fuel], 2, _unit]] call EFUNC(common,objectEvent);
         } else {
-            ["displayTextStructured", [_unit], [LSTRING(Hint_Empty), 2, _unit]] call EFUNC(common,targetEvent);
+            ["displayTextStructured", _unit, [LSTRING(Hint_Empty), 2, _unit]] call EFUNC(common,objectEvent);
         };
```
GitHawk: Is using `objectEvent` instead of `targetEvent` better for single target?

jonpas: Yes, it does local check, so if it already local it will just do it locally without network delay, it saves you the local check in the code.

addons/common/XEH_postInit.sqf
jonpas: That is not used in refuel though, when it is used somewhere it can be added, I just added those used in refuel.

jonpas:
Tested in locally hosted multiplayer, no issues anymore.

PR #3544, opened 2016-03-05
Fastroping: deployment API and waypoint

BaerMitUmlaut:
- Add `ace_fastroping_fnc_deployAI`, a public API function for deploying all units that sit in the cargo of a helicopter.
- Add a waypoint which does the same automatically.

Requires #3553.


addons/fastroping/functions/fnc_deployAI.sqf
```diff
@@ -0,0 +1,95 @@
...
+ * Public: Yes
+ */
+
+#include "script_component.hpp"
+params [["_vehicle", objNull, [objNull]], ["_deploySpecial", false, [true]], ["_createDeploymentGroup", true, [true]]];
+private ["_config", "_configEnabled", "_deployTime", "_unitsToDeploy", "_deployGroup"];
+
+if (isNull _vehicle || {!(_vehicle isKindOf "Helicopter")}) exitWith {
+    if (hasInterface) then {
+        ["deployAI was called with an invalid or non-existant vehicle.", QFUNC(deployAI)] spawn BIS_fnc_guiMessage;
```
BaerMitUmlaut: Note: `BIS_fnc_guiMessage` causes a CTD with call.

Glowbal: Add comment in the code please

esteldunedain: Do we really want to show a clickable error message if somebody places a waypoint of this type by mistake?

BaerMitUmlaut: I usually use `BIS_fnc_error`, but it doesn't work in Eden.

esteldunedain: So this message would only show in the editor, or may it happen too during the mission?

I'd prefer it didn't show during the mission, as it's probably non-critical (plus players can't do anything about it)

jonpas:
Tagging as WIP until that fix is PRed and merged, so no one merges it by mistake.

esteldunedain:
@BaerMitUmlaut, does #3553 contain the fix you are referring to here?

BaerMitUmlaut:
Yes, b20e433 is the fix to be precise.

Issue #3397, opened 2016-02-23
Cannot load Save Games


Dshrike:
**Arma 3 Version:** `Game version: 1.56.134627 - Main*

**ACE3 Version:** `3.4.2.0. - Stable'

**Mods:**
@CBA_A3
@ace

**Placed Modules:**
None

**Description:**
Unable to load save games with the ACE mod active. First, I receive an error: "cannot load "C:\\...\continue.arma3.save" (or whatever save file it attempts to load). Then, I receive error: "Error 0x00000001 – UNABLE_TO_INIT_DXGI"

**Steps to reproduce:**
1.) Load singleplayer/multiplayer scenario.
2.) Save while in scenario.
3.) Leave game by pressing "save & exit"
4.) Re-load scenario by pressing "resume"

**Where did the issue occur?**
Possibly after the eden editor update?

**RPT log file:**
http://pastebin.com/kfNqd4Rz

commy2:
Can reproduce with CBA 2.3.1 and ACE 3.4.2 or ACE master
Cannot reproduce with CBA 2.3.1

I have no idea what this is. Seems to me like BI fucked up serialization.

jokoho48:
i know this problem from a while ago i try to load a Save game from 1.3x to 1.4x and i had this same issue.

commy2:
They are new save games.

commy2:
Cannot repro with:
ace_common.pbo
ace_main.pbo
ace_modules.pbo

commy2:
Can reprodue with:

- [ ] ace_apl.pbo
- [x] ace_common.pbo
- [ ] ace_interaction.pbo
- [ ] ace_interact_menu.pbo
- [ ] ace_laser.pbo
- [x] ace_main.pbo
- [ ] ace_missileguidance.pbo
- [x] ace_modules.pbo
- [ ] ace_weather.pbo

`Item count not matching: +2`

commy2:
Error is in ace_interact_menu\CfgEventHandlers.hpp

commy2:
Bug caused by initPost event handler in interact_menu module. (compiling interaction menues)
BI broke nested array serialization. Nothing we can do.

Dshrike:
Just to confirm, we'll need to wait until BI decides to fix the issue on their own?

esteldunedain:
>  we'll need to wait until BI decides to fix the issue on their own?

[...]

PR #3547, opened 2016-03-06
Store Interact Menus on CBA Namespaces

commy2:
- Store action menues in CBA pseudo namespaces instead of `format`ed mission namespace variables
- I can't reproduce the save game crash with this anymore
- I don't know why though, so don't close the issue for now (requires feedback)

ref #3397, #1098

### DIFF
--- a/addons/interact_menu/XEH_preInit.sqf
+++ b/addons/interact_menu/XEH_preInit.sqf
@@ -4,6 +4,9 @@ ADDON = false;
 
 #include "XEH_PREP.hpp"
 
+GVAR(ActNamespace) = [] call CBA_fnc_createNamespace;
+GVAR(ActSelfNamespace) = [] call CBA_fnc_createNamespace;
+
 // Event handlers for all interact menu controls
 DFUNC(handleMouseMovement) = {
     if (GVAR(cursorKeepCentered)) then {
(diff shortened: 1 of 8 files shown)


PabstMirror:
Should we do the zeus actions as well?

commy2:
Zeus actions? Probably yes.

commy2:
>Should we do the zeus actions as well?
I don't think that's necessary as that one is one level less deep. There only is one zeus actions mission namespace array, while there was one for each unit type.

esteldunedain:
:+1: for merging this @commy2. I checked to see if you might have missed replacing it somewhere, but I couldn't find anything wrong.

commy2:
Yay. Let's see if that fixes the save game problem. Can't wait for testing results from the mob.

esteldunedain:
>  Can't wait for testing results from the mob.

Lol @bux578,** calling our user-base "the mob" could trigger a great PR mess.

JoeSchr:
I tried it with Campaign Mission: `The East Wind -> Win -> Resurgent West`.


I saved ~2min in, exited and couldn't load or resume the misson after. 

When disabling ACE3, restarting the mission, saving, loading it works. 

If I try to load a "naked" save game with ACE3 activated, I get 
`The instruction at 0xetc referenced memory at 0x000000. The memory could not be read`

Sadly it seems, save games are **not working**. Or do I need to disable/enable anything? 
Tested with ACE 3.5.1

commy2:
Has nothing to do with this. Please open a new issue. Keep in mind that the campaign is not supported.

PR #3551, opened 2016-03-06
Disable only action menu on mouse hint and modifier where unnecessary

jonpas:
- Allow scripted UI when using `showMouseHint` and `hideMouseHint`
- Remove `ACE_Modifier` requirement for rotating sandbags, tactical ladders, explosives and trenches


esteldunedain:
:+1:

PR #3560, opened 2016-03-08
execute addActionEvents in unscheduled env

commy2:
- execute addActionEvents in unscheduled env
- all my tears


esteldunedain:
@commy2, as always, great stuff.

PR #3561, opened 2016-03-08
FCS and Scopes with no compat pbo required

esteldunedain:
**Experimental**

- Add a correction to ACE_FCS to account for vanilla zeroing
- Allow ACE_FCS to work without requiring a compatibility PBO
- Add a correction to ACE_Scopes to account for vanilla zeroing
- Allow ACE_Scopes to work without requiring a compatibility PBO

**Rationale**
There are currently 4 ACE systems that require "ACE-Exclusive" compatibility PBOs (that cannot be included in the content without breaking it for vanilla):
- FCS (`discreteDistance[] = {}`)
- Scopes (`discreteDistance[] = {}`)
- Explosives
- Disposable

~~The technique applied in this PR could also be applied to scopes~~ (done), so the only remaining systems would be explosives and disposable, which affect only a minimum fraction of third party content (e.g., doesn't affect vehicles nor small arms).

The current fix is fast (the extension takes 0.01 ms to calculate, and only is required to do so once every time the unit changes zeroing, weapon or ammo) and seems to be as accurate. Fortunately `currentZeroing` works properly with remote units.

**TODO** 
- [ ] Hide the vanilla zeroing control when appropiate, so the end user won't even know it's still enabled. Tagging @jonpas for advice
- [ ] Restore `discreteDistance[] = {}` for vanilla content?
- [ ] Correctly compute the initSpeed from config instead of getting it from the bullet?

### DIFF
--- a/addons/scopes/functions/fnc_firedEH.sqf
+++ b/addons/scopes/functions/fnc_firedEH.sqf
@@ -27,7 +27,7 @@ if (_adjustment isEqualTo []) then {
     _adjustment = [0,0,0];
 } else {
     _adjustment = _adjustment select _weaponIndex;
-}
+};
 // Convert it from mils to degrees
 private _zeroing = _adjustment vectorMultiply 0.05625;
 _zeroing params ["_elevation", "_windage", "_zero"];
@@ -38,7 +38,7 @@ _elevation = _elevation + _zero;
 if (_adjustVertical) then {
     // Calculate the correction due to vanilla zeroing
     private _zeroDistance = (currentZeroing _unit - GVAR(defaultZeroDistance));
-    if (_zeroDistance == 0) then {
+    if (_zeroDistance != 0) then {
         private _weaponCombo = [_weapon, _magazine, _ammo, _zeroDistance];
         if !(_weaponCombo isEqualTo (_unit getVariable [QGVAR(lastWeaponCombo), []])) then {
             // Hackish way of getting initSpeed. @todo: replace it by correct calculation and caching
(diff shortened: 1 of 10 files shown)


jonpas:
> Hide the vanilla zeroing control when appropiate, so the end user won't even know it's still enabled.

I will add a public function for hiding UI parts, will also act like a setter so UI component itself won't overwrite it (as it has to be set on vehicle changes as well).

commy2:
This would be excellent! :+1:

esteldunedain:
I added a first implementation of the correction to ACE_Scopes. I didn't test this yet, so it might still contain bugs.

Khaosmatic:
Regarding vehicle FCS, how is ACE's going to work alongside BI's 1.60 enhancements?

ulteq:
> I added a first implementation of the correction to ACE_Scopes.

Won't this cause all scopes to be zeroed at 0 meters?

ulteq:
> The current fix is fast (the extension takes 0.01 ms to calculate, and only is required to do so once every time the unit changes zeroing, weapon or ammo) and seems to be as accurate. Fortunately `currentZeroing` works properly with remote units.
How about passing the `currentZeroing` value to the FCS extension and let it handle everything?

addons/scopes/functions/fnc_firedEH.sqf
```diff
@@ -15,20 +15,49 @@
...
 _zeroing params ["_elevation", "_windage", "_zero"];
-
-[_projectile, _windage, _elevation + _zero, 0] call EFUNC(common,changeProjectileDirection);
+_elevation = _elevation + _zero;
+
+// The scope supports vertical adjustment, so even if _zeroing is [0,0,0] we
+// still need to compute the correction to compensate for vanilla zeroing
+if (_adjustVertical) then {
+    // Calculate the correction due to vanilla zeroing
+    private _zeroDistance = (currentZeroing _unit - GVAR(defaultZeroDistance));
```
ulteq: I'm pretty sure that this won't work well.

esteldunedain:
Closing in favor of #4642 and #4653

PR #3563, opened 2016-03-08
3DEN - ACE_Settings Display

PabstMirror:
**WIP** - Only need design feedback

- Adds ability to set ace_setting's from 3d mission editor.

### DIFF
--- a/addons/optionsmenu/XEH_preInit.sqf
+++ b/addons/optionsmenu/XEH_preInit.sqf
@@ -15,4 +15,50 @@ GVAR(serverSideValues) = [];
 GVAR(categories) = [];
 GVAR(currentCategorySelection) = 0;
 
+recomp = {
+#include "XEH_PREP.hpp"
+};
+
+[] spawn {
+    sleep 0.1; //is3den command has problems at pre-init, need spawn/sleep
+    if (!is3DEN) exitWith {};
+    ACE_LOGINFO("3den detected, localizing settings");
+
+    //Post-init won't run in 3den, so we need to directly call some settings funcs normally run at common: post-init
+    call EFUNC(common,loadSettingsFromProfile);
+    call EFUNC(common,loadSettingsLocalizedText);
+
+    //Get Categories
+    GVAR(categories) pushBack "";  //Ensure All Catagories is at top
+    {
+        GVAR(categories) pushBackUnique (_x select 8);
+    } forEach EGVAR(common,settings);
+
+
+    disableSerialization;
+
+    waitUntil {!isNull (findDisplay 313)};
+
+    if ((findDisplay 313) getVariable [QGVAR(menuSet), false]) exitWith {TRACE_1("menu already set",true);};
+    TRACE_1("Setting up menu",false);
+    (findDisplay 313) setVariable [QGVAR(menuSet), true];
+
+    private _3denDisplay = findDisplay 313;
+
+    private _menuItemsArray = getArray (configfile >> "Display3DEN" >> "Controls" >> "MenuStrip" >> "Items" >> "items");
+    private _aceIndex = _menuItemsArray find "ACE_Settings";
+
+    TRACE_1("",_aceIndex);
+
+    {
+        private _displayName = _x;
+        if (_displayName == "") then {
+            _displayName = localize LSTRING(Uncategorized);
+        };
+        private _addedIndex = (_3denDisplay displayCtrl 120) menuAdd [[_aceIndex], _displayName];
+        TRACE_3("Adding",_x,_displayName,_addedIndex);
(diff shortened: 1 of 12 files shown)


addons/optionsmenu/XEH_preInit.sqf
```diff
@@ -15,4 +15,50 @@ GVAR(serverSideValues) = [];
 GVAR(categories) = [];
 GVAR(currentCategorySelection) = 0;
 
+recomp = {
```
Glowbal: Any reason we use `recomp` without a prefix?

addons/optionsmenu/functions/fnc_3den_openSettingsDisplay.sqf
```diff
@@ -0,0 +1,145 @@
...
+private _currentSavedSettings = QUOTE(ADDON) get3DENMissionAttribute "ACE_MissionSettings";
+if (isNil "_currentSavedSettings") then { ACE_LOGERROR("_currentSavedSettings nil"); };
+
+//ToDo: Maybe flesh out this sub-catogies stuff later
+// for now just seperate general from clientSettable settings
+private _subCatagoriesNames = ["Client", "General"];
+private _subCatagoriesSettings = [[], []];
+{
+    if ((_x select 8) == _categoryLocalized) then {
+        private _subCatIndex = if (_x select 2) then {
```
Glowbal: How about:
```sqf
private _subCatIndex = [0, 1] select (_x select 2);
```
It's a little bit cleaner as using an if else statement for this.

jonpas: :+1: for Glowbal's suggestion.

addons/optionsmenu/XEH_preInit.sqf
```diff
@@ -15,4 +15,50 @@ GVAR(serverSideValues) = [];
...
+#include "XEH_PREP.hpp"
+};
+
+[] spawn {
+    sleep 0.1; //is3den command has problems at pre-init, need spawn/sleep
+    if (!is3DEN) exitWith {};
```
jonpas: We should probably make an event for this in common.

PabstMirror:
Added sub-categories, if not explictly defined they are generated from the var name of the setting.
Helps to organize the list, and adds context for the setting display name.

commy2:
That looks amazing

addons/optionsmenu/XEH_preInit.sqf
PabstMirror: This is just for development, recompiles all funcs before opening menu.
Makes testing very fast.
Will remove when PR is ready.

esteldunedain:
:clap::clap::clap:

jonpas:
Impressive! :+1:

PabstMirror:
Should be good to review now, still want to move some settings to categories/subCategories, but that can be done on another PR

addons/optionsmenu/XEH_preInit.sqf
```diff
@@ -15,4 +15,55 @@ GVAR(serverSideValues) = [];
 GVAR(categories) = [];
 GVAR(currentCategorySelection) = 0;
 
+
+[] spawn {
+    sleep 0.1; //is3den command has problems at pre-init, need spawn/sleep
```
Glowbal: Any reason we don't run this as post-init is pre-init has issues?

Glowbal: Never mind - I just read the comment on line 24.

commy2: what about execNextFrame?

PabstMirror: execNextFrame is installed in postInit

Glowbal: A lot of things are done on postinit. Doesn't that mean all these are now broken in 3den?

commy2: postInit does fire when you press preview

Glowbal: If that is the case - why are we doing a spawn here? Is this comment false:
> //Post-init won't run in 3den, so we need to directly call some settings funcs normally run at common: post-init

commy2: by "in 3den" he means while placing objects. postInit will fire once you press "preview", but not while you actually are in the editor.

Glowbal:
@commy2 would you be able to give some input? This might be something for CBA settings framework perhaps?

Glowbal:
Leaving this open until we decide whatever or not we commit to using the CBA settings framework. If we do, this PR should be closed. Otherwise, it is good to go in.

jonpas:
Redundant with move to CBA Settings.

PR #3567, opened 2016-03-08
MicroDAGR microoverhaul

BaerMitUmlaut:
- Replace most of the current awful image files from the MicroDAGR
- Show the head direction instead of the body direction on the compass of the MicroDAGR

The dialog could use a rewrite so it doesn't stretch the images anymore but I'm not insane. If somebody does this one day, I still have the source files for all images I made and can provide unstretched images.

--- 
**Screenshots:**
Overview
Menu


esteldunedain:
Perfect; some screenshot would be welcome.

BaerMitUmlaut:
Updated with screenshots.

PabstMirror:
Original icons were from reference material: 

https://www.rockwellcollins.com/~/media/Files/Unsecure/Products/Product%20Brochures/Navigation%20and%20Guidance/GPS%20Devices/MicroDAGR%20data%20sheet.aspx (PDF)

Personally I prefer the color icons, but I suppose having higher resolution is important.
I think it might look better with a little less contrast (very light grey instead of bright white) similar to the text's color.

BaerMitUmlaut:
I can adjust them to the grey if you want. I'll try it during this week.

sargken:
Will there be an actual battery on these that will run out and signal that can change depending on where you are?

BaerMitUmlaut:
I have neither information about how long the battery would last nor knowledge about on which exact conditions the signal strength would change and how much it would change.

licht-im-Norden87:
I don't know if it is necessary to have some sort of battery life with this device. As it was made for military purpose I assume it can last for a day or longer (but I absolutly have no clue at all) and because of that it is not relevant for arma.

PR #3569, opened 2016-03-08
Add warnings for missing compat PBOs

PabstMirror:
- Add warnings for missing compat PBOs


esteldunedain:
There's a slight issue with the compat pbo warning. Most of those weapon compat mods are only needed if you use advanced_ballistics. 

One alternative would be to move the requirements to that module, although it would still through the warning if AB is disabled, which is not ideal either. 

Another alternative would be not to require them at all, given that AB works properly without them too (although oc bullet trajectories are not realistic).

commy2:
pushing back to 3.6.0, because reasons

PabstMirror:
Looks like CBA's code run's at postInit, before modules are read, so we might not have the correct value for `isABenabled`

jonpas:
Wrap it in the runAtSettingsInitialized maybe? Probably not worth it though, I just never liked direct definitions in code.

PabstMirror:
Redoing for master

Issue #3557, opened 2016-03-07
Mk6 Mortar ammo handling broken on dedicated


SzwedzikPL:
**Arma 3 Version:** `1.56` (stable - hotfixed)
**CBA Version:** `2.3.1` (stable)
**ACE3 Version:** `3.5.0` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Prepare/Remove round option is not available until you enter mortar gunner (mortar init is called by `playerVehicleChanged`)
- On first use of mortar vanilla magazines are still loaded
- After first use you can unload vanilla rounds or shoot with them
- After shooting all vanilla rouds or unloading them you cannot load ACE mortar rounds properly (round is loaded but you can't fire)

**Steps to reproduce:**
- https://www.youtube.com/watch?v=jWMgG_dari0 (recorded on dedicated multiplayer)

**Where did the issue occur?**
- Dedicated

**Placed Modules:**
`ace_mk6mortar_airResistanceEnabled` - false
`ace_mk6mortar_allowComputerRangefinder` - false
`ace_mk6mortar_allowCompass` - true
`ace_mk6mortar_useAmmoHandling` - true

**RPT log file:**
- rpt log is clear, no errors

VKing6:
It's intentional that the mortar doesn't initialize the ammo handling system before a player enters it. That it didn't initialize is unexpected.

Was the mortar set to empty (no playable unit/AI) in the editor, or did you start manning it?

SzwedzikPL:
I started recording 5 sec after mission start. Mortar was set to empty (no AI) in editor. Mission made in Eden.

OriginalHoozin:
I've done some experimenting.

This is a problem for any mortar that is owned by the dedicated server.  Spawning one via Zeus behaves as expected (enter it, it initializes properly).  Spawning one via script on server through debug menu creates a mortar that will not initialize properly.

[...]

PR #3583, opened 2016-03-14
Init the mortar in the machine in which it's local

esteldunedain:
- Init the mortar in the machine in which it's local. Ref #3557
- Possible fix for #3557


esteldunedain:
I got contacted by @SzwedzikPL on Slack. Apparently this fix won't work, so I'm closing it. His has been working on a proper solution, and the PR is incoming.

The gist of the thing is that the initialization of the turret weapon needs to be done on the machine where the *turret* is local, and that has it's complication because locality switches around the time when you enter the turret.

PR #3584, opened 2016-03-14
Fix mortar ammo handling init on dedicated

SzwedzikPL:
- Change mortarInit into global event with turret locality check to be sure that mortar init is executed on machine that owns turret.
- Fix for https://github.com/acemod/ACE3/issues/3557
- Tested on dedicated multiplayer

### DIFF
--- a/addons/mk6mortar/functions/fnc_mortarInit.sqf
+++ b/addons/mk6mortar/functions/fnc_mortarInit.sqf
@@ -18,27 +18,28 @@
 params ["_mortar"];
 
 if (_mortar getVariable [QGVAR(initialized),false] || _mortar getVariable [QGVAR(exclude),false]) exitWith {TRACE_1("Exit",_mortar)};
+if (!(_mortar turretLocal [0])) exitWith {TRACE_1("Exit - turret not local",_mortar)};
 
-// Remove all magazines
+// Remove all magazines from turret
 if (count magazines _mortar > 0) then {
     {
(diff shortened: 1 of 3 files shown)


esteldunedain:
:+1: for merging

commy2:
>Tested on dedicated multiplayer
Nice. GJ man!

addons/mk6mortar/functions/fnc_mortarInit.sqf
```diff
@@ -18,27 +18,28 @@
...
     } forEach magazines _mortar;
 };
 
-// Replace current weapon with ammo handling weapon
+// Replace current turret weapon with ammo handling weapon
 private _currentWeapon = _mortar weaponsTurret [0] select 0;
 private _newWeapon = "";
 
-if (_currentWeapon == "mortar_82mm") then {
+if (tolower _currentWeapon == "mortar_82mm") then {
```
commy2: technically not needed, because `==` compares strings case insensitive, but whatever

jonpas: It should also be `toLower` if we're being picky.

commy2: ironic, isn't it?

jonpas: :ice_cream:

commy2:
branch needs update. can't do that. @SzwedzikPL pls fix. also remove that tolower while you're at it. thanks.

PR #3592, opened 2016-03-17
Fix nametag error when rank is empty string

PabstMirror:
- Fix nametag error popup

image

`rank` seems like it can return `""` right as unit is respawning.

```
Rank qq_1 is [LIEUTENANT]
Rank qq_1 is []
Rank qq_1 is [LIEUTENANT]
```

This might be a side-effect from using cachedCall to get all near men?


commy2:
`rank objNull is ""`

esteldunedain:
> rank objNull is ""

Mmm, `!(isNull _target)`is already checked before calling this function. I guess `rank`is returning wrong values for non-null objects under certain corner-case conditions

Anyway, :+1: for merging

PR #3593, opened 2016-03-17
Fixed fastroping race condition

BaerMitUmlaut:
- Fix a network race condition that broke fastroping and got people stuck in mid air
- Fix ropes not properly detaching when cut

---

**Details on the network race condition that occured:**
Expected order on commands ran was:
- player gets moved out of the helicopter
- server PFH runs once, starts ropes unwinding
- player changes animation
- nothing really happens until the fast rope is done

What happened sometimes was:
- player gets moved out of the helicopter
- server PFH does *not* run in between here
- player changes animation
- server skips the ropes unwinding command
- ropes are stuck, player is stuck

The condition was changed and the rope unwinding command is now guaranteed to run once, which gets rid of the issue.


BryanBoru:
Excited to see the progress with this. Hope YoursTruly's input and our silliness was enough to keep your stress down while figuring this out.

PR #3600, opened 2016-03-19
Dogtags

SzwedzikPL:
- Add option to check / take dogtag from dead unit
- Every dogtag item is unique, nickname assigned to item is saved on server

Demo: https://www.youtube.com/watch?v=qMeOXQ0UYFE

Todo
- [x] Item icon
- [x] Item model
- [x] CBA Events
- [ ] Action: check all dogtags at once
- [ ] Change dogtags into magazines?
- [ ] Proper dedicated multiplayer tests


esteldunedain:
This was yet another item in my TODO list; I'm glad to see someone make progress!!

Implementation looks very good! One thing that I'd like to see implemented eventually is this: afaik, the chain contains a set of two dogtags. One is removed from casualties, but the other one is left with the remains, in order to be able to identify them later. So in a "fresh"  casualty, both dogtags should be displayed, and once the first one was removed, the remaining one should be displayed, but without the possibility of taking it.

SzwedzikPL:
> Implementation looks very good! One thing that I'd like to see implemented eventually is this: afaik, the chain contains a set of two dogtags. One is removed from casualties, but the other one is left with the remains, in order to be able to identify them later. So in a "fresh" casualty, both dogtags should be displayed, and once the first one was removed, the remaining one should be displayed, but without the possibility of taking it.

Nice idea. It would be great if someone could prepare brackgrounds for both versions of display (i mean background with both dogtags and background with only one).

addons/dogtags/functions/fnc_addDogtagActions.sqf
```diff
@@ -0,0 +1,40 @@
...
+params ["_target", "_player"];
+
+//Get all dogtags and their ids
+private _unitDogtags = [];
+private _unitDogtagIDs = [];
+{
+    private _id = getNumber (configFile >> "CfgWeapons" >> _x >> QGVAR(tagID));
+    if(_id > 0) then {
+        _unitDogtags pushback _x;
+        _unitDogtagIDs pushback _id;
```
commy2: `pushBack`

addons/dogtags/functions/fnc_addDogtagItem.sqf
```diff
@@ -0,0 +1,23 @@
...
+ * Return Value:
+ * None
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_item", "_nickname"];
+
+if(_item == "") exitWith {};
```
commy2: `if (`

addons/dogtags/functions/fnc_canCheckDogtag.sqf
```diff
@@ -0,0 +1,20 @@
...
+ * Return Value:
+ * True if dogtag can be checked <BOOL>
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_player", "_target"];
+
+if(isNull _target) exitWith {false};
```
commy2: same

addons/dogtags/script_component.hpp
```diff
@@ -0,0 +1,1029 @@
...
+        author = ECSTRING(common,ACETeam); \
+        scope = 1; \
+        scopeArsenal = 0; \
+        scopeCurator = 0; \
+        descriptionShort = QUOTE(DOGTAGID); \
+        GVAR(tagID) = DOGTAGID; \
+    };
+
+#define ACE_DOGTAGS(DOGTAGBASE) \
+    ACE_DOGTAG_CLASS(DOGTAGBASE,1) \
```
commy2: pretty sure you can make a loop with preprocessor commands, so this 1000 item list isn't necessary.

addons/dogtags/stringtable.xml
```diff
@@ -0,0 +1,89 @@
...
+            <English>Dog Tag</English>
+            <German></German>
+            <French></French>
+            <Czech></Czech>
+            <Polish>Nieśmiertelnik</Polish>
+            <Spanish></Spanish>
+            <Hungarian></Hungarian>
+            <Russian></Russian>
+            <Italian></Italian>
+            <Portuguese></Portuguese>
```
commy2: Does the localizaton script find these when they are empty?

addons/dogtags/script_component.hpp
SzwedzikPL: please send me an example :)

addons/dogtags/stringtable.xml
jonpas: Those shouldn't be here, just remove the empty ones.

SzwedzikPL:
new demo: https://www.youtube.com/watch?v=2V2tPYBqFak
@esteldunedain Is this what you had in mind?

addons/dogtags/stringtable.xml
gienkov: String table entries cannot be empty for other languages because they will
display blank name for given language then. Either fill up rest of
languages or delete blank entries and let translators do their job.
Missed that functionality too.
19.03.2016 6:09 PM "jonpas" <notifications@github.com> napisał(a):

esteldunedain:
> @esteldunedain Is this what you had in mind?

Yeah, pretty much! Thank you.

BaerMitUmlaut:
Is that image taken from Battlefield? If so I can make a custom one to avoid copyright issues.

SzwedzikPL:
> Is that image taken from Battlefield? If so I can make a custom one to avoid copyright issues.

Im not sure but it's possible. Someone gave me this about year ago. I really dont remember. If you can make your own i will be grateful :)

Orakel78:
Will there be other dogtag models ?

Quite a few armies / countries uses dog tags which you break of one piece and leave the other half behind.

esteldunedain:
> Will there be other dogtag models ?

Maybe in the future. Let's get guy finish a first version first ;)

gienkov:
That dogtag is from ACE2 I bet.

SzwedzikPL:
> That dogtag is from ACE2 I bet.

Nope

BryanBoru:
The more I look at this the more I am certain that dogtag image is taken from battlefield. I am trying to find a decent model for ya and just haven't had luck with one that would accept the text addition well.

BaerMitUmlaut:
I'm currently making a model and texture.

Khaosmatic:
Would it be possible to give the individual dog tag items their own name? This would then alleviate the need of having to look at for example 'Dog Tag #1' to find out which one is which.

jonpas:
This has the same issue as #3601 with IDs https://github.com/acemod/ACE3/pull/3601#issuecomment-224769645

commy2:
- [ ] CBA events

jonpas:
Move to 3.7.0 for possible improvements?

PR #3601, opened 2016-03-19
Unique maps for markers

SzwedzikPL:
- Add unique maps system for markers
- Can be disabled by settings module (wip)

Demo: https://www.youtube.com/watch?v=fV-k580Ld2A

Todo
- [x] Proper briefing markers handling
- [ ] Copy markers from other unit map
- [ ] Copy markers from map in player inventory
- [ ] Settings module
- [ ] Public API functions for mission-makers
- [ ] Proper MP testing (im testing on local dedicated for now)
- [ ] Proper polyline markers handling

The polyline markers for now are quite impossible to handle by scripts. There's no way to add polyline marker. It's needed becouse we need to delete and add markers based on what map you have. I hope BI will add tools (proper script support) to allow us to do this.


esteldunedain:
Thank you for making the PR @SzwedzikPL. I'll review and provide feedback soon.

Glowbal:
You have a finite map ids but never release any. This results in a growing registery of ids and the risk of long sessions running out of map ids.

SzwedzikPL:
New map is generated only if not unique map is taken by player.

esteldunedain:
> New map is generated only if not unique map is taken by player.

@SzwedzikPL, I think @Glowbal is refering to public servers, in which you may have a large number of players connecting and disconnecting between server restarts. In that case, you may very well run out of map ids. 

To be honest, I don't see any good way to if a given map will no longer be needed and releasing it's id. Maybe the best solution would be that if you need a new unique map and there's no more left, just reset and use the oldest one that was abandoned. Or something along those lines.

SzwedzikPL:
I think that public servers should just disable this option. I don't see any benefits of this solution on public server. We can make unique maps disabled by default to make this easier for them.

addons/markers/XEH_PREP.hpp
```diff
@@ -1,12 +1,20 @@
...
 PREP(mapDrawEH);
 PREP(onLBSelChangedChannel);
 PREP(onLBSelChangedColor);
 PREP(onLBSelChangedShape);
 PREP(onSliderPosChangedAngle);
 PREP(placeMarker);
-PREP(sendMarkersJIP);
-PREP(setMarkerJIP);
-PREP(setMarkerNetwork);
+PREP(initInsertMarker);
```
esteldunedain: functions should be in alphabetical order

addons/markers/XEH_postInit.sqf
```diff
@@ -1,20 +1,77 @@
...
-// recieve remote marker data
-[QGVAR(setMarkerNetwork), {_this call DFUNC(setMarkerNetwork)}] call EFUNC(common,addEventHandler);
+[QGVAR(loadMapData), DFUNC(loadMapData)] call EFUNC(common,addEventHandler);
+[QGVAR(setMarkerGlobal), DFUNC(setMarkerGlobal)] call EFUNC(common,addEventHandler);
+[QGVAR(loadBriefingMarkers), DFUNC(loadBriefingMarkers)] call EFUNC(common,addEventHandler);
+[QGVAR(onMapItemChange), DFUNC(onMapChange)] call EFUNC(common,addEventHandler);
 
-// recieve marker data for JIP
-[QGVAR(setMarkerJIP), {_this call DFUNC(setMarkerJIP)}] call EFUNC(common,addEventHandler);
+if(hasInterface) then {
```
esteldunedain: Most of the content of postInit should be wrapped in a `"settingsInitialized"` event handler, that quit inmediately if the feature is disabled:

```
["settingsInitialized", {
    if (!GVAR(enableUniqueMaps)) exitWith {};

    // All unit markers init and events here
    // ...
```

That way you wouldn't have to constantly check for `!GVAR(enableUniqueMaps)` anymore, and there would be no code running if the feature is disabled. 

Check other modules for examples on this pattern (e.g. ace_hearing)

addons/markers/XEH_postInit.sqf
```diff
@@ -1,20 +1,77 @@
...
+    //check marker changes after map is closed
+    ["visibleMapChanged", {
+        params ["", "_mapOn"];
+        TRACE_1("visibleMapChanged",_mapOn);
+
+        if (!GVAR(enableUniqueMaps)) exitWith {};
+        if (_mapOn) then {
+            GVAR(newMapMarkers) = [];
+        } else {
+            private _assignedMap = ACE_player call FUNC(getAssignedMap);
```
esteldunedain: This code branch should also run in case somehow the current map item changes while you have the map opened (probably from the `playerInventoryChanged` event above). Hence I recommend moving the code to a dedicated function and running it from both places.

addons/markers/functions/fnc_updateMapData.sqf
```diff
@@ -0,0 +1,51 @@
...
+
+params ["_map", "_mapData"];
+TRACE_2("updateMapData",_map,_mapData);
+
+_mapData params ["_newMarkers", "_deletedMarkers"];
+
+private _mapID = _map call FUNC(getMapID);
+if (_mapID <= 0) exitWith {};
+
+private _mapMarkers = missionNamespace getVariable [format ["%1_%2", QGVAR(allMapMarkers), _mapID], []];
```
esteldunedain: I recommend storing this in a CBA namespace instead of on the missionNamespace. Just search `CBA_fnc_createNamespace` in the ACE codebase for examples.

Using a namespace makes prefixing `_mapId` unnecessary, so this could be:

```sqf
private _mapMarkers = GVAR(mapNamespace) getVariable _mapID;
if (isNil "_mapMarkers") then {
    _mapMarkers = []
};
```

(that if is needed because namespace do not support the getVariable ARRAY syntax)

addons/markers/script_component.hpp
```diff
@@ -24,3 +24,1014 @@
...
+#define ACE_MAP_CLASS(MAPBASE,MAPID) \
+    class MAPBASE##_##MAPID : MAPBASE { \
+        scope = 1; \
+        scopeArsenal = 0; \
+        scopeCurator = 0; \
+        descriptionShort = QUOTE(MAPID); \
+        GVAR(mapID) = MAPID; \
+    };
+
+#define ACE_MAPS(MAPBASE) \
```
esteldunedain: These macros are usually inserted into `CfgWeapon.hpp` directly. 

Also the `ACE_MAPS` macro seems redundant to me.

addons/markers/functions/fnc_onMapChange.sqf
```diff
@@ -0,0 +1,32 @@
...
+if (_currentMap == "") then {
+    GVAR(currentMapID) = 0;
+    {deleteMarkerLocal _x} forEach GVAR(currentMapMarkers);
+    GVAR(currentMapMarkers) = [];
+    GVAR(currentMapMarkersProperties) = [];
+} else {
+    private _currentMapID = _currentMap call FUNC(getMapID);
+    if (GVAR(currentMapID) != _currentMapID) then {
+        if (isMultiplayer) then {
+            GVAR(localLogic) = (createGroup sideLogic) createUnit ["Logic", [0,0,0], [], 0, "NONE"];
```
esteldunedain: @commy2 had an idea to bypass the need of creating this object, which would be usefull. I think it requires the server to broadcast each player his ID on mission start or JIP.

addons/markers/functions/fnc_loadBriefingMarkers.sqf
```diff
@@ -0,0 +1,49 @@
...
+#include "script_component.hpp"
+
+if (!hasInterface || {!GVAR(enableUniqueMaps)}) exitWith {};
+
+private _assignedMap = ACE_player call FUNC(getAssignedMap);
+if (_assignedMap == "") exitWith {};
+
+//can't update markers data if map is not unique
+private _mapID = _assignedMap call FUNC(getMapID);
+if (_mapID <= 0) exitWith {};
```
esteldunedain: If map classname is used as _mapId then this would become something like `if (_mapId == Base_Map)`

addons/markers/functions/fnc_getMapID.sqf
```diff
@@ -0,0 +1,17 @@
...
+ * Return Value:
+ * Map ID <NUMBER>
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_map"];
+
+getNumber (configFile >> "CfgWeapons" >> _map >> QGVAR(mapID))
```
esteldunedain: Maybe we could use `_map` (string with the class name) as the map id, instead of requiring to constantly read from config what integer corresponds to that.

addons/markers/functions/fnc_isMap.sqf
```diff
@@ -0,0 +1,17 @@
...
+ * Return Value:
+ * true if given item is a map <BOOL>
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_item"];
+
+getNumber (configFile >> "CfgWeapons" >> _item >> QGVAR(isMap)) > 0
```
esteldunedain: It should be faster and simpler to use something like:

`(_item find MAPBASE) == 0`

addons/markers/functions/fnc_isMap.sqf
```diff
@@ -0,0 +1,17 @@
...
+ * Return Value:
+ * is a map <BOOL>
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_item"];
+
+_item == "ItemMap" || _item call FUNC(isUniqueMap)
```
BaerMitUmlaut: This whole function could be replaced with `isKindOf`.

addons/markers/functions/fnc_isUniqueMap.sqf
```diff
@@ -0,0 +1,17 @@
...
+ * Return Value:
+ * is a unique map <BOOL>
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_item"];
+
+((toLower _item) find "ace_map_") == 0
```
BaerMitUmlaut: Again, I'd use `isKindOf` instead.

esteldunedain: Not exactly, you don't only need to know if the item is a map, but if it's unique

SzwedzikPL:
I've added note about polyline markers

jonpas:
Would this work by making maps magazines? That way you have unique items and don't risk running out of IDs.

SzwedzikPL:
> Would this work by making maps magazines? That way you have unique items and don't risk running out of IDs.

Is this possible to make magazine a usable map? (i mean map that you can put on map slot in inventory)

jonpas:
I am not sure, but you could do something similar to ACRE2. Hide the inventory UI for map and track that in the background on inventory change events maybe.

esteldunedain:
TBH @jonpas, I much rather keep the current methodology. It's much simpler,
and running out of ids isn't really an important concern.

addons/markers/XEH_postInit.sqf
```diff
@@ -1,20 +1,59 @@
-// by commy2
```
commy2: T_T

bux:
@SzwedzikPL Status on this one?
Can we close the pull request? (your branch lives on in your fork)

BaerMitUmlaut:
Map drawing still has absolutely no script support.

bux:
I'll close this. 4 years and no updates.

synixebrett:
Hold my beer

PR #3608, opened 2016-03-20
fix script error in tagging

commy2:
- fixes a script error introduced in this commit: https://github.com/acemod/ACE3/commit/72c9e9da39dd2bbb3e96070eae482e5f3468159c


Glowbal:
If this is for 3.6.0, it should go into the master branch @commy2

PR #3611, opened 2016-03-21
Greatly improve the accuracy of the protractor display

### LINKED ISSUE #3606: AB - Protractor : wrong indication
**Arma 3 Version:** `1.56` (stable)
**CBA Version:** `2.3.1` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- The zero and the angles of the protractor ("inclinomètre" in french) are not good. This isn't the same Indicator but the zero should be like this for example : [http://www.horusvision.com/asli.php](http://www.horusvision.com/asli.php)

**Steps to reproduce:**
- Just compare the protractor with a laser designator.

- Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

**Placed Modules:**
- Advanced Ballistics manually enabled.

**RPT log file:**
- Necessary for this bug ?

esteldunedain:
- Improve accuracy of the protractor display. The movement is now linear with the sin of the angle instead of the angle itself, to account for the distorsion in the dial.
- Close #3606

For best results, the needle should match the center of the numbers in the dial.

### DIFF
--- a/addons/advanced_ballistics/functions/fnc_displayProtractor.sqf
+++ b/addons/advanced_ballistics/functions/fnc_displayProtractor.sqf
@@ -47,7 +47,7 @@ GVAR(Protractor) = true;
     __ctrl1 ctrlSetTextColor [1, 1, 1, 1];
 
     __ctrl2 ctrlSetScale 1;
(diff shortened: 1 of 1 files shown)


commy2:
math

PR #3623, opened 2016-03-24
Fix AB Terrain Extension

PabstMirror:
Fix Terrain processor never running

 - `FUNC(initializeTerrainExtension)` was called at postInit, before modules are read so `GVAR(enabled)` would be false

 - `#endIf` - That capital I breaks everything with arma preProcessor. `ace_advanced_ballistics_fnc_initializeTerrainExtension` was compiling to just `{}`


esteldunedain:
:+1:

PR #3624, opened 2016-03-24
Compat pbo warnings

PabstMirror:
- Add warnings for missing compat PBOs

Replace #3569

### DIFF
--- a/addons/advanced_ballistics/XEH_postInit.sqf
+++ b/addons/advanced_ballistics/XEH_postInit.sqf
@@ -40,13 +40,13 @@ if (!hasInterface) exitWith {};
             ACE_LOGWARNING_2("Weapon Mod [%1] missing ace compat pbo [%2] (from @ace\optionals)",_modPBO,_compatPBO);
         };
     } forEach [
-    ["RH_acc","ace_compat_rh_acc"],
-    ["RH_de_cfg","ace_compat_rh_de"],
-    ["RH_m4_cfg","ace_compat_rh_m4"],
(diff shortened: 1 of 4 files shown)


addons/advanced_ballistics/XEH_postInit.sqf
```diff
@@ -33,6 +33,21 @@ if (!hasInterface) exitWith {};
...
+            ACE_LOGWARNING_2("Weapon Mod [%1] missing ace compat pbo [%2] (from @ace\optionals)",_modPBO,_compatPBO);
+        };
+    } forEach [
+    ["RH_acc","ace_compat_rh_acc"],
+    ["RH_de_cfg","ace_compat_rh_de"],
+    ["RH_m4_cfg","ace_compat_rh_m4"],
+    ["RH_PDW","ace_compat_rh_pdw"],
+    ["RKSL_PMII","ace_compat_rksl_pm_ii"],
+    ["iansky_opt","ace_compat_sma3_iansky"],
+    ["R3F_Armes","ace_compat_r3f"]
```
commy2: should be indented 4 more spaces here.

addons/main/config.cpp
```diff
@@ -586,18 +586,6 @@ class CfgMods {
     };
 };
 
+#include "CfgDependencies.hpp"
```
commy2: Should be `CfgSettings` It's easier to find if we name things how they are actually named. and could prevent headaches when we want to add more stuff to CfgSettings later, for whatever reason.

PR #3627, opened 2016-03-27
Fix Capitalization of Many ACE_Settings classes.

654wak654:
- Fix capitalization of the following ACE_Settings classes:
 - common_SettingProgressBarLocation
 - frag_Enabled
 - frag_SpallEnabled
 - frag_ReflectionsEnabled
 - frag_MaxTrackPerFrame
 - frag_EnableDebugTrace
 - interact_menu_AlwaysUseCursorSelfInteraction
 - interact_menu_AlwaysUseCursorInteraction
 - interact_menu_UseListMenu
 - interaction_EnableTeamManagement
 - magazinerepack_TimePerAmmo
 - magazinerepack_TimePerMagazine
 - magazinerepack_TimePerBeltLink
 - map_DefaultChannel
 - maptools_EveryoneCanDrawOnBriefing
 - microdagr_MapDataAvailable
 - overheating_DisplayTextOnJam
 - reload_DisplayText
 - repair_DisplayTextOnRepair
 - repair_engineerSetting_Repair
 - repair_engineerSetting_Wheel
 - repair_repairDamageThreshold_Engineer
 - repair_consumeItem_ToolKit
 - respawn_SavePreDeathGear
 - respawn_RemoveDeadBodiesDisconnected
 - respawn_BodyRemoveTimer
 - explosives_RequireSpecialist
 - explosives_PunishNonSpecialists
 - explosives_ExplodeOnDefuse
 - hearing_EnableCombatDeafness
 - hearing_EarplugsVolume
 - hearing_UnconsciousnessVolume
 - hearing_DisableEarRinging
 - ace_weaponselect_DisplayText
 - ace_vehiclelock_DefaultLockpickStrength
 - ace_vehiclelock_LockVehicleInventory
 - ace_vehiclelock_VehicleStartingLockState
 - ace_switchunits_EnableSwitchUnits
 - ace_switchunits_SwitchToWest
 - ace_switchunits_SwitchToEast
 - ace_switchunits_SwitchToIndependent
 - ace_switchunits_SwitchToCivilian
 - ace_switchunits_EnableSafeZone
 - ace_switchunits_SafeZoneRadius
 - ace_map_gestures_GroupColorConfigurations
 - ace_map_gestures_GroupColorConfigurationMapping


addons/map/ACE_Settings.hpp
```diff
@@ -9,7 +9,7 @@ class ACE_Settings {
     class GVAR(BFT_Enabled) {
         category = CSTRING(Module_DisplayName);
         value = 0;
-        typeName = "BOOL";
+        typeName = "BOOL";d
```
Glowbal: Extra character that shouldn't be there

Glowbal:
All commits should be squashed into one before we merge this.

jonpas:
I don't really know how much I like this, I usually uppercase GVARs like that.

SilentSpike:
Yeah I'm definitely not in favour of lower case for GVARs

654wak654:
Amount of GVAR macros that start with upper case letters is 1315, inside 314 different script files. On the other hand, there are 5000+ GVAR macros in 800+ files that start with lower case letters.

jonpas:
This breaks backwards compatibility for any 3rd party mod using those GVARs to compare. For example `isEqualTo` or `find`. One such example is `"SettingChanged"` event:
```
if (_name isEqualTo QGVAR(Enabled)) then {....
```

commy2:
I see no problem with that. Code should always be made case insensitve and I doubt anyone actually does it that way.

jonpas:
Sure, it's just a thing I thought I should point out to consider.

And if we are uniforming them, then we should do it everywhere and set a standard IMO.

esteldunedain:
:+1: For this.

@Glowbal, @bux, if you agree with this I'd recommend merging asap, before conflicts start creeping in.

654wak654:
If most people agree on the way GVARs should be named, I suggest we close this, and I make a new commit with all of the **1300+** GVARs (ones start with upper-case letters) fixed. I have the proper regex tested and ready, it whould be fast enough to be done between any other commits.

Also @Glowbal I cannot squash these because of the stupid way I set it up.

jonpas:
That would be good indeed. Let's wait for @Glowbal and @bux578 for thumbs up/down.

654wak654:
https://github.com/654wak654/ACE3/commit/d8b4c172969c027d8b115e199516d81bd466a197

PR #3628, opened 2016-03-27
Italian translation

alef:
#367 Italian translation completed up to now.

Don't know what to do with the BOM in stringtable.xml file generated by tabler, so you have two commits here, same content but with and without BOM.


bux578:
Grazie!

PR #3631, opened 2016-03-28
Add support for custom textures in tagging

jonpas:
- Allow custom texture usage in tagging
- Support for 3rd party mods, with this they only need the ACE Actions definition, they can handle randomness themselves with a custom function before calling `FUNC(tag)`
- Adds a requested feature: https://github.com/acemod/ACE3/issues/3594#issuecomment-199326646 and https://github.com/acemod/ACE3/issues/3594#issuecomment-199398152

Example:
```
// Internal (helper function for random texture based on color)
statement = QUOTE([ARR_2(ACE_player,'black' call FUNC(getTexture))] call FUNC(tag));

// Full path support (texture in mission folder)
statement = QUOTE([ARR_2(ACE_player,'tags\tag.paa')] call FUNC(tag));
```

Tagging @BaerMitUmlaut


addons/tagging/functions/fnc_createTag.sqf
```diff
@@ -19,15 +19,21 @@
...
+private _customTexture = [true, false] select (_colorTexture find ".paa" == -1);
+
+if (!_customTexture && {!((toLower _colorTexture) in ["black", "red", "green", "blue"])}) exitWith {
+    ACE_LOGERROR_1("%1 is not a valid tag colour.",_colorTexture);
 };
 
 private _tag = "UserTexture1m_F" createVehicle [0,0,0];
-_tag setObjectTextureGlobal [0, '\z\ace\addons\tagging\UI\tags\' + _color + '\' + str (floor (random 3)) + '.paa'];
+if (!_customTexture) then {
+    _tag setObjectTextureGlobal [0, "\z\ace\addons\tagging\UI\tags\" + _colorTexture + "\" + str (floor (random 3)) + ".paa"];
```
esteldunedain: If we want to add this type of customization, I'd recommend changing the `_colorTexture` variable ot `_texture`, and moving all the color logic out of `fnc_createTag` and `fnc_tagDirection` and directly into the ACE_Action statement. Now it's the time to do it, before the PR that add tagging events is merged.

jonpas: Yeah I had that idea as well, will get around to it later.

esteldunedain:
When people suggested customizing tagging I thought of a completely different system than that of this PR. Idk, if it's better of worse; I'll leave some quick notes here for discussion/ideas:

1- Config entries to define available tags:

```c++
class ACE_Tags {
    class MyTag_Cross {
        name = CSTRING(tagCross);
        class Blue {
            actionName = CSTRING(tagBlueCross);
            textureNames[] = {'\z\ace\addons\tagging\UI\tags\blue\1.paa', '\z\ace\addons\tagging\UI\tags\blue\2.paa'};
        };
        class Black {
            ...  
        };
       ...
    };
    class MyTag_Killroy {      
    ...
```

2- Interaction menu action to select the "active tag" 

```
Self Interaction -> Equipment -> Select Tag -> Killroy
```

jonpas:
Seems like a way more complicated system, with this one you just need to inherit from one of the existing classes of the spray color you want then just call the function with correct arguments. I did some cleanup, pushing soon.

I don't really like the "select active tag", seems like it takes more time to select it then just more options in the interaction menu, unless you keep using the same one for an extended period of time.

esteldunedain:
> just more options in the interaction menu, 

That's debatable:

Number of options with this PR: (numColors * numTags)
Number of options with my proposal: (numColors + 1)

jonpas:
> it takes more time to select it then just more options in the interaction menu

I meant that in this PR there is more options.

esteldunedain:
Anyway, I think we should roll with your PR, given it's almost done.

The only thing that I still don't like it's the `FUNC(getTexture)`, because it doesn't help at all when adding new tags. What about rewritting it to be more usefull, eg:

`[QUOTE(PATHTOF(UI\tags\%1_%2.paa)), _color, 3] call FUNC(getTexture]`

jonpas:
I was thinking of it simply as an internal function to get the texture path, not something 3rd parties would use, if they would want randomness they can implement their own. But your way is much more friendly indeed, I'll do that. :+1:

jonpas:
Actually @esteldunedain , this just complicates the config a lot due to usage of arrays (need `ARR_n` macro). Also, it's still very specific to the folder setup that is used in this component. I don't think we benefit anything from it.

Most 3rd parties would probably have 1 texture and that's it, if they want randomness doing something like this `FUNC(getTexture)` does is very easy, maybe even less complicated depending on their setup.

addons/tagging/stringtable.xml
```diff
@@ -1,91 +1,63 @@
...
-        <Key ID="STR_ACE_tagging_tagWallGreen">
+        <Key ID="STR_ACE_Tagging_TagGreen">
             <English>Tag green</English>
             <German>Grün markieren</German>
             <Spanish>Marcar en verde</Spanish>
             <Polish>Oznakuj na zielono</Polish>
             <French>Tag vert</French>
         </Key>
-        <Key ID="STR_ACE_tagging_tagWallBLue">
+        <Key ID="STR_ACE_Tagging_TagBLue">
```
theYoursTruly: _STR_ACE_Tagging_TagB**L**ue_ should be _STR_ACE_Tagging_TagB**l**ue_

theYoursTruly:
https://github.com/acemod/ACE3/pull/3631/files#diff-509e8b391ef91d244f83412bc8f0039eR25
_STR_ACE_Tagging_TagB**L**ue_ should be _STR_ACE_Tagging_TagB**l**ue_

addons/tagging/stringtable.xml
theYoursTruly: _STR_ACE_Tagging_TagB**L**ue_ should be _STR_ACE_Tagging_TagB**l**ue_

addons/tagging/CfgVehicles.hpp
```diff
@@ -4,57 +4,31 @@ class CfgVehicles {
         class ACE_SelfActions {
             class ACE_Equipment {
                 class ACE_tagWallBlack {
```
jonpas: Should we rename the classes or rather keep backwards compatibility? They are no longer "wall" specific.

jonpas: I went ahead and did it, minor chance someone is using it (except @ColdEvul).

esteldunedain: Yeah, I agree it's not a big deal.

jonpas:
Ready for final review, @BaerMitUmlaut check it out and merge if you think it's good.

addons/tagging/CfgVehicles.hpp
jonpas: :+1:

PR #3632, opened 2016-03-28
More agressive blood loss for serious wounds

PabstMirror:
This would increase bleeding when a body part is heavily damaged.

The goal is to make wounds from large calibre wounds much more dangerous; this has a small increase for moderate wounds, but ramps up quickly for damage done from something like a heavy sniper rifle's bullet.

@Glowbal - this could also possibly be made into an optional setting if you are happy with how bleeding is balanced as is.

### DIFF
--- a/addons/medical/functions/fnc_getBloodLoss.sqf
+++ b/addons/medical/functions/fnc_getBloodLoss.sqf
@@ -24,7 +24,7 @@ params ["_unit"];
 _totalBloodLoss = 0;
 
 private _bloodLossMuliplier = _unit getVariable [QGVAR(bodyPartStatus), [0,0,0,0,0,0]];
-_bloodLossMuliplier = _bloodLossMuliplier apply { 1 + _x };  
+_bloodLossMuliplier = _bloodLossMuliplier apply { 1 + _x^2 };  
 
 // Advanced medical bloodloss handling
 if (GVAR(level) >= 2) then {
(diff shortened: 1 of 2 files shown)


addons/medical/functions/fnc_getBloodLoss.sqf
```diff
@@ -20,6 +23,9 @@ private ["_totalBloodLoss","_tourniquets","_openWounds", "_cardiacOutput", "_int
 params ["_unit"];
 _totalBloodLoss = 0;
 
+private _bloodLossMuliplier = _unit getVariable [QGVAR(bodyPartStatus), [0,0,0,0,0,0]];
```
esteldunedain: `_bloodLossMultiplier`

esteldunedain:
Do you have any numbers on how long it takes to bleed out now vs this PR? Some hard data would be nice to inform decisions like this.

PabstMirror:
Using: `[cursorTarget, 0.75, "leg_r", "bullet"] call ace_medical_fnc_addDamageToUnit`

Assuming we get a "Large Velocity Wound" each time (0.1/second base bleed rate)
If blood < 60 we can pass out.

Before:
0.75 hit: 0.1/sec - 400 seconds
1.5 hit: 0.1/sec - 400 seconds
2.25 hit: 0.1/sec - 400 seconds
After:
0.75 hit 0.156/sec - 256 seconds
1.5 hit, 0.325/sec - 123 seconds
2.25 hit, 0.6/sec -  67 seconds

Glowbal:
@PabstMirror is going to test this out and tweak based on his findings. At the moment, it looks like this is something we want but will be rendered unnecessary long term because of the planned merger between the basic and advanced wounds systems.

We will wait on @PabstMirror to report his findings before merging this.

Glowbal:
Could probably be closed in favour of #3916

Issue #3621, opened 2016-03-24
ACE_Explosives_fnc_scriptedExplosive 


C0RAX:
**Arma 3 Version:** `x.xx` (stable / rc / dev)
**CBA Version:** `2.x.x` (stable / dev + commit hash)
**ACE3 Version:** `3.x.x` (stable / dev + commit hash)

**Mods:**
- CBA_A3
- ACE
- RHS
- UOMODS
- UOMAPS
- UOMAPSCUP
- ACRE

**Description:**
- when using "call ACE_Explosives_fnc_scriptedExplosive;" as described on the aCE3wiki the function gives an error(Error Undefined variable in expression: _dettime) line 6. 

full script code:
"_ied = _this select 0;      
			
			waitUntil {
					sleep 0.5;
					{((_x distance _ied < 10)) and (side _x == west) } count allUnits > 0;
				};
				[[_ied],1] call ACE_Explosives_fnc_scriptedExplosive;
"

**Steps to reproduce:**
- place  "ACE_IEDUrbanBig_Range"
- the placed ied has "null = [this] execVM "scripts\ied.sqf";" in the init field.
- uses above script code in ied.sqf
- approach ied till triggered

**Where did the issue occur?**
- Editor (Singleplayer)

**Placed Modules:**
- none

**RPT log file:**
- http://pastebin.com/aQvEyyYc 
- line 5857 for error
- http://pastebin.com/UTDmhWwc

commy2:
The function is meant to be executed in unscheduled environment.

Try replacing:
´[[_ied],1] call ACE_Explosives_fnc_scriptedExplosive;`
with:
`[ACE_Explosives_fnc_scriptedExplosive, [[_ied],1]] call CBA_fnc_directCall`

C0RAX:
Works great, thank you for the prompt response. 

would ask for this to be added and documented in the wiki as it's not immediately clear that it wont work in a scheduled environment.

esteldunedain:
@C0RAX, scheduled environment is plain unreliable because, depending on the load your mission is having, the executed code may be interrupted at any moment. That's why ACE runs 95% of it's code in non-scheduled environment. 

Best practice would be to always call ACE API on non-sched (and the same for ACRE, TFAR, etc). You are right we should document that.

bux578:
@esteldunedain I think @C0RAX is referring to this wiki page:
http://ace3mod.com/wiki/missionmaker/useful-functions.html

[...]

PR #3633, opened 2016-03-29
make FUNC(scriptedExplosive) work in scheduled env, ref #3621

commy2:
- make FUNC(scriptedExplosive) work in scheduled env
- add option to directly add one explosive as object for convenience

the alternative syntax can do stuff like:
`_bomb call ace_explosives_fnc_scriptedExplosive`

(as well as:
`[_bomb] call ace_explosives_fnc_scriptedExplosive`
`[[_bomb]] call ace_explosives_fnc_scriptedExplosive`)

which is way easier on the eyes compared to:
`[[_bomb]] call ace_explosives_fnc_scriptedExplosive`

### DIFF
--- a/addons/explosives/functions/fnc_scriptedExplosive.sqf
+++ b/addons/explosives/functions/fnc_scriptedExplosive.sqf
@@ -4,7 +4,7 @@
  * detonate editor-placed explosives.
  *
  * Arguments:
- * 0: Explosives objects to detonate <ARRAY>
+ * 0: Explosives objects to detonate <OBJECT or ARRAY>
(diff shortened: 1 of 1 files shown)


VKing6:
Looks good :+1:

esteldunedain:
Mmm, how exactly does this adress the sched vs non-sched issue?

commy2:
The line: `private _detTime;` is wrong.
It's not the `private` keyword here, because there is no following `=` sign.
The preprocessor thinks it's either `private STRING` or `private ARRAY`, where the variable name itself is the value of another variable.
That variable is never defined, which throws an error in scheduled environment.

esteldunedain:
Aha, thanks for the explanation @commy2.

Issue #3622, opened 2016-03-24
Warning Function name unit network id - spam


TimelessInfinity:
**Arma 3 Version:** `1.56` (stable)
**CBA Version:** `2.3.1` (stable)
**ACE3 Version:** `3.5.1` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`
- (Spawning Units method)

**Description:**
- This is a little bit tricky and need a bit more to explain. I have the last days spend much time to figure out what is going on and of course Iam not sure whether it is right to post it here. I want to do so in another Issue Tracker may from ALiVE but the issue pops only up when ACE is in the modset, additional the error is not there when ACE +CBA is alone. So maybe you see the bad circumstances for me where i should post.

- Try to explain the error.
In the last weeks I was not up to date with my Server, so the Issue Iam sure dont occur bevore ACE 3.5.0.
In all MissionSetups when Units Spawn in the "Warning name Function" spam the RPT for every Unit they spawn in/out or die. When the Mission use a "caching" system and you fly with Zeus is the RPT spam heavy. The same with Zeus spawn Units or only a Spawn Script in the same Mission.

- Figure down where it comes from
Only ACE+CBA and may the Spawn Script or Zeus dont occur this "heavy" error.
But in Singleplayer/Editor the "light" one is this and this one have no server/performance issue.
`WARNING: Function 'name' - O Alpha 1-3:5 has empty name`

-  Now ACE + one of the "spawn in/out" solution above create the error.
`WARNING: Function 'name' - 637fa040# 164212: o_soldier_01.p3d has no unit
 - network id 0:0
 - person `

- Only CBA+any kind of spawn/missionmaker solutions (Alive/MCC/Zeus/SpawnScript etc.)
Dont occur this Error. I mean both errors dont pop up.

Why this broke the Mission on our Server? (I dont know whether it is this error, nothing more there to find.)

This may only to see in ASM. You need to see the Engine Simulation Cycles.
I have linked 1 Screenshoots but they need a little explain what i have done.
This is Missionstart only 7-10minutes.
When the blue line (AI) goes up i fly to cached AI, you see the simulation cycles broke down to 5 or 8 and AI line blue pops in. The CPS stay there so long the AI is active. In this case Iam alone on the Server, when there are player the Server is gone.
The only error Message is the Warning function, i found nothing more in the last 3days of work on this. When i load the same setup but without ACE then is all normal. 50CPS/FPS, no Warning error.
[...]

PR #3634, opened 2016-03-29
fix FUNC(setName) for Zeus and alive spawned units

commy2:
- artificially delay the FUNC(setName) to fix RPT spam
- fix nametags not working with zeus / alive spawned units
- close: #3622
- remove some commented out code

Should be noted that this worked before 1.56. The name was set on init, at least on local units. Not anymore. ~~The one second is arbitrary.~~ I spawned 1000 units and got no complaints in the RPT. `FUNC(setName)` itself is safe to be delayed and doesn't actually require a local unit.


esteldunedain:
:+1: Great @commy2

commy2:
Turns out one frame is enough.

PR #3644, opened 2016-03-31
fnc_getMGRSdata update1

J3aLeiNe:
fnc_getMGRSdata update1 : worldName, latitude and elevationOffset checked for all maps, new maps added.


addons/common/functions/fnc_getMGRSdata.sqf
```diff
@@ -22,27 +22,38 @@ private _long = getNumber (configFile >> "CfgWorlds" >> _map >> "longitude");
...
-if (_map in ["Chernarus", "Bootcamp_ACR", "Woodland_ACR", "utes"]) then { _lat = 50; _altitude = 0; };
+if (_map in ["chernarus", "Chernarus_Summer", "chernarus_winter", "Bootcamp_ACR", "Woodland_ACR", "utes"]) then { _lat = 50; _altitude = 0; };
 if (_map in ["Altis", "Stratis"]) then { _lat = 40; _altitude = 0; };
-if (_map in ["Takistan", "Zargabad", "Mountains_ACR"]) then { _lat = 35; _altitude = 2000; };
+if (_map in ["takistan", "zargabad", "Mountains_ACR"]) then { _lat = 35; _altitude = 2000; };
 if (_map in ["Shapur_BAF", "ProvingGrounds_PMC"]) then { _lat = 35; _altitude = 100; };
 if (_map in ["fallujah"]) then { _lat = 33; _altitude = 0; };
-if (_map in ["fata", "Abbottabad"]) then { _lat = 30; _altitude = 1000; };
+if (_map in ["fata"]) then { _lat = 33; _altitude = 1347; };
+if (_map in ["Abbottabad"]) then { _lat = 34; _altitude = 0; };
```
bux578: According to https://en.wikipedia.org/wiki/Abbottabad Abbottabad has an elevation of 1256m

J3aLeiNe: Agree with you @bux578 but seems to be a pb for the gps after :
https://ace3public.slack.com/archives/advanced_ballistics/p1459019087000072

bux578: je ne comprend pas. c quoi "pb"?

esteldunedain:
Quick question, just in case, is `in` case-sensitive?

jonpas:
Yes it is @esteldunedain 

> String values will be compared on CaSEseNsiTIve basis
`"lol" in ["Lol", "LOL", "loL"]; //false`

esteldunedain:
Ok, then I propose we lowercase `_map` before `in` and change all the comparisons to lower case. e.g.:

```sqf
_map = lower _map;
if (_map in ["chernarus", "chernarus_summer", "chernarus_winter", "bootcamp_acr", "woodland_acr", "utes"]) then { _lat = 50; _altitude = 0; };
. . . 
```

addons/common/functions/fnc_getMGRSdata.sqf
J3aLeiNe: pb = problem, sorry @bux578

jonpas: It's better that the GPS display is wrong rather than ballistics being calculated wrong.

esteldunedain: The gps should add `GVAR(mapAltitude)` here to fix this:

https://github.com/acemod/ACE3/blob/master/addons/maptools/functions/fnc_openMapGpsUpdate.sqf#L18

The fact that it's not added is a left over from AGM, in which no map offset existed

J3aLeiNe: The stand alone AB had a AB_Latitude and a AB_Altitude for all calculations,  a "acelat" and a "acealtitude" for example could be a solution ?
Sorry if it's stupid, i'm not scripter.

esteldunedain: ACE already has something similar:
```sqf
     GVAR(mapAltitude)
     GVAR(mapLatitude)
     GVAR(mapLongitude)
```

the problem is just that the gps is not adding `GVAR(mapAltitude)` as it should (the DAGR, microDAGR and Kestrel are doing it properly already!)

J3aLeiNe:
tolower added
https://github.com/acemod/ACE3/pull/3644/commits/d4a7d864aa3dfc2020f0705ab83306d6c2753f4b

Sorry in advance if i made a mistake, i'm completely lost ...

esteldunedain:
Perfect @J3aLeiNe. Please restore the 1000m or 1256m altitute for Abbottabad. After that I think this should be ready for meging.

J3aLeiNe:
I don't know exactly how the altitude influences the ballistic in-game but i know that we have a very good ballistic with  _altitude=0, on Altis for example. The altitude effect should be a little bit more tested, in my opinion, before to change all maps' altitude in this sqf.

But, anyway, altitude for Abbottabad fixed.

jonpas:
:+1:

esteldunedain:
Thanks @J3aLeiNe. 

The effect of altitude over ballistics is represented very realistically. Air is thinner the higher you go, so drag and wind drift decreases with altitude.

I created a ticket (#3650) to fix the gps display to account for map altitude. I'll do that soon.

PR #3647, opened 2016-04-02
Add Core Module Macros

jokoho48:
- add CFUNC QCFUNC CGVAR QCGVAR Macros to make the use of Core/Common functions easyer to access
- make use of the new Macros


jonpas:
Seems like majority is against, closing.

Sorry @jokoho48 :cry:

PR #3651, opened 2016-04-02
Add map altitude to GPS

### LINKED ISSUE #3650: The GPS display doesn't account for global map altitude
Ref #3644

> The gps should add GVAR(mapAltitude) here to fix this:

> https://github.com/acemod/ACE3/blob/master/addons/maptools/functions/fnc_openMapGpsUpdate.sqf#L18****

jonpas:
- Account for map altitude in GPS display
- Fixes #3650


esteldunedain:
Thanks!

PR #3658, opened 2016-04-04
Wait until postInit to install playerChanged event

PabstMirror:
Due to https://github.com/CBATeam/CBA_A3/pull/303
PFEH will be run between preInit and postInit, so the "playerChanged" event happens before most event handlers are installed in postInit.

This breaks noRadio, and other things that count on the `playerChanged` event always happening.


esteldunedain:
:+1:

PR #3666, opened 2016-04-07
Allow multiple code sources to insert sub actions

PabstMirror:
-   Allow multiple code sources to insert sub actions

example:

```
    class CAManBase: Man {
        class ACE_SelfActions {
            class ACE_Equipment {
                class additionalChildren {
                    a = "call addChildren1";
                    b = "call addChildren2";
                };
```


PabstMirror:
At one point I had an explicit use for this, but now I don't... so I guess this could be closed?

One possible use would be something like `ace_tagging` adding unlimited custom children to the equipment menu?

Glowbal:
> One possible use would be something like ace_tagging adding unlimited custom children to the equipment menu?

I think that would be useful to support @PabstMirror

jonpas:
This would allow 3rd party mods to hook into our config events as well, but we don't really need it for ourselves I guess? And it's not as clean as providing a proper API for 3rd party to hook into for various systems specifically.

PabstMirror:
Going to close for now,
if someone has an actual use for this we can reopen

PR #3677, opened 2016-04-10
BFT - Use params command and private keyword

jonpas:
- Use `params`
- Change `private ARRAY` to `private _var`
- Change `var = if-then-else` to `select BOOL`

Todo:
- [ ] bft
- [x] bft_devices
- [x] bft_drawing


Glowbal:
Thanks @jonpas. We will do the bft component separately.

PR #3678, opened 2016-04-11
Raise a local event when a weapon is jammed

Jsmuk:
- Raise an event when a player's weapon jams. 

This is a very minor (1 line!) change which will add an event upon weapon jamming to the public API. I personally have a use for it in an existing addon I am working on and I am sure that others may find it useful.

### DIFF
--- a/addons/overheating/functions/fnc_jamWeapon.sqf
+++ b/addons/overheating/functions/fnc_jamWeapon.sqf
@@ -43,6 +43,9 @@ if (_ammo > 0) then {
 // only display the hint once, after you try to shoot an already jammed weapon
 GVAR(knowAboutJam) = false;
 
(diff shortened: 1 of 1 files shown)


esteldunedain:
I think the event name is not ideal. As a rule of thumb, we are currently not prefixing the event names with the module name. Also we tend to use `camelCase` instead of `snake_case`.

Hence I propose changing the event name to `weaponJammed`.

@Glowbal, @bux, what do you think?

Jsmuk:
I was originally going to use `weaponJammed` but then I noticed an event I was already using (`medical_onSetDead`) so that confused me a little bit.

esteldunedain:
> I was originally going to use weaponJammed but then I noticed an event I was already using (medical_onSetDead) so that confused me a little bit.

Medical might be the only exception to the rules I posted :smile:

PR #3690, opened 2016-04-13
Clarify repair while engine is running

### LINKED ISSUE #3348: Player cannot repair vehicle with engine turned ON
Arma 3 Version:** `1.56`
(Stable)

ACE3 Version:** `3.4.2`
(Stable)*

**Mods:**
* @CBA_A3
* @ace
*  RHS Escalation

**Description:**
When engine of vehicle is turned on, you cannot repair any its part. Its very annoying, beacuse In very "realistic" arma world, player is very often forced to get out of tank when it have broken track (only on track!) straigt under enemy fire, but thats the other story, when player is forced out, there is action used "jump out" instead of "get out", so engine is still runing and player can only .... die, beacuse he cannot get back to vehicle (when he is not commander of group), cannot repair tank, and is on open field with his PDW against snipers, tanks, choppers.. 

**Steps to reproduce:**
Get a tank platon, set yourself as commander of one tank, but not group leader. Get damage on your tank to be forced out of vehicle. Now you will not be able to repair it.

Singleplayer

PabstMirror:
- Show repair action, but exit and show warning when selected
- Optional (default off) setting to auto shut off engine to keep

Close #3348

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -146,6 +146,7 @@ if (isServer) then {
 
 ["setDir", {(_this select 0) setDir (_this select 1)}] call FUNC(addEventhandler);
 ["setFuel", {(_this select 0) setFuel (_this select 1)}] call FUNC(addEventhandler);
+["engineOn", {(_this select 0) engineOn (_this select 1)}] call FUNC(addEventhandler);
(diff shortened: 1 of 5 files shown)


Glowbal:
:+1:

vlad8011:
So how to turn this ON to default ?
"Optional (default off) setting to auto shut off engine to keep"

PR #3693, opened 2016-04-14
Fix ship view distance using foot instead of vehicle setting

jonpas:
- Fix view distance using "on foot" setting when in boats/ships instead of "in vehicle" setting.
- Cleanup view distance component (`params`, `private` keyword, camel case, `isEqualType `)
- Move constants to `script_component.hpp`


esteldunedain:
:+1:

PR #3697, opened 2016-04-15
docs: mark waitUntilAndExecute as Public: Yes

Cuel:
- Describe what this pull request will do: update docs
- Each change in a separate line: 

-mark waitUntilAndExecute as public

Is there a reason not to? Noticed it wasn't available in Atom


jokoho48:
yes there was a reason. the reason is https://github.com/CBATeam/CBA_A3/pull/288 this PR to CBA. because after the next CBA release we can remove this function.

esteldunedain:
We plan on deprecating the ACE version of those functions. Hence I don't think making that function public makes sense at this point.

PR #3705, opened 2016-04-19
Gforces - Add ace_setting to only run in aircraft

PabstMirror:
- Add ace_setting to gforces so the PFEH only runs when player is inside of a aircraft

Default is unchanged, but we might want to make default aircraft only (`1`)?
I don't think any ground vehicle can accelerate enough to have any real effect.


esteldunedain:
Yeah, go ahead, change the default. The pfh always run because it predated the `vehicleChanged` EH. 

Otherwise, the PR looks fine. I'm reassigning it back to you.

Thanks Pabst!

jonpas:
Agreed on aircraft-only default.

PR #3710, opened 2016-04-20
Remove remaining usage of player

BaerMitUmlaut:
- Remove the remaining usage of `player` instead of `ACE_player`.

### DIFF
--- a/addons/common/functions/fnc_setHearingCapability.sqf
+++ b/addons/common/functions/fnc_setHearingCapability.sqf
@@ -54,5 +54,5 @@ private _lowestVolume = 1;
 0 fadeMusic _lowestVolume;
 
 // Set Radio mod variables.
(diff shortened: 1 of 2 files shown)


esteldunedain:
I think `fnc_setHearingCapability.sqf` came CSE, and it'is not used in ACE; maybe you should remove it altogether.

PabstMirror:
The function looks like a good idea because both hearing and medical both try to change volume (with hearing overwriting medical)

But I think we can fix conflicts in the hearing module by checking if player is Unconscious.

Glowbal:
Using such an approach would make it easier for any other kind of modification to adjust hearing. Makes it more flexible

BaerMitUmlaut:
Guys have you forgotten about the unified status effect handlers? This is basically made for it.

PR #3635, opened 2016-03-29
Fix inverted gestures Stop and Freeze and minor gestures cleanup

jonpas:
**When merged this pull request will:**
- Fix #3543 - inverted "Stop" and "Freeze" ("Stop" will still be "Freeze" in prone, it's a vanilla animation we can't change)
- Cleanup unused config
- Improve stringtable (remove duplicate)
- Replace DIK KeyCodes include and usage with decimal codes
- Improve keybind adding and figuring out if it's a vanilla gesture or not
- Add `README.md` and update author URL in `config.cpp`

PR #3713, opened 2016-04-21
Fix BIS gesture keybinds

PabstMirror:
 - From #3635 - FUNC(playSignal) needs "BIgesture" prefix on all BI gestures
 - Interaction conditions are checked on parent, don't need on every
child

### DIFF
--- a/addons/gestures/XEH_postInit.sqf
+++ b/addons/gestures/XEH_postInit.sqf
@@ -10,10 +10,10 @@ GVAR(ReloadMutex) = true;
     _x params ["_currentName", "_key", ["_vanillaKey", false] ];
 
     // Don't add "ace_gestures_" prefix to BI gestures
-     private _signalName = if (_vanillaKey) then {
-         format ["BIgesture%1", _currentName];
+    private _signalName = if (_vanillaKey) then {
+        format ["BIgesture%1", _currentName];
     } else {
-         format [QGVAR(%1), _currentName];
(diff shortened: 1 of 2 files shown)


addons/gestures/XEH_postInit.sqf
```diff
@@ -10,9 +10,10 @@ GVAR(ReloadMutex) = true;
     _x params ["_currentName", "_key", ["_vanillaKey", false] ];
 
     // Don't add "ace_gestures_" prefix to BI gestures
-    private _signalName = _currentName;
-    if (!_vanillaKey) then {
-        _signalName = format [QGVAR(%1), _currentName];
+     private _signalName = if (_vanillaKey) then {
```
jonpas: Indentation

addons/gestures/XEH_postInit.sqf
```diff
@@ -10,9 +10,10 @@ GVAR(ReloadMutex) = true;
     _x params ["_currentName", "_key", ["_vanillaKey", false] ];
 
     // Don't add "ace_gestures_" prefix to BI gestures
-    private _signalName = _currentName;
-    if (!_vanillaKey) then {
-        _signalName = format [QGVAR(%1), _currentName];
+    private _signalName = if (_vanillaKey) then {
```
jokoho48: ```private _signalName = format [[QGVAR(%1), "BIgesture%1"] select _vanillaKey, _currentName]```

PR #3714, opened 2016-04-21
Make explosives helpers hidden in eden.

PabstMirror:
- Removes duplicate entries for explosives in eden, placing in editor makes no sense as they won't have needed `setVariables`


commy2:
What about zeus? `scopeCurator = 0` ?

SilentSpike:
As long as they aren't listed in the `units[]` array it should be good

jonpas:
Removing `scopeCurator = 1;` is good for clean config though.

commy2:
:+1:

jonpas:
:+1:

PR #3725, opened 2016-04-26
Overheating - Add Master Enabled Setting

PabstMirror:
- Add Master Enabled Setting for the overheating module

Module will have no effect when set to false.


esteldunedain:
This PR looks perfectly fine to me. The only gotcha is that I'd create some nasty conflicts with #3438, which was marked for review a long time ago.

PR #3727, opened 2016-04-27
Added ffaamod compat pbo

FFAAMOD:
Added the ffaa mod compatiblity (First attempt)


jonpas:
Please change the code to follow the [ACE3 Coding Guidelines](http://ace3mod.com/wiki/development/coding-guidelines.html), mainly tabs usage and braces being on a new line.

optionals/compat_ffaamod/CfgGlasses.hpp
```diff
@@ -0,0 +1,16 @@
+#define COMBAT_GOGGLES         ACE_Overlay = z\ace\addons\goggles\textures\HUD\CombatGoggles.paa; \
+        ACE_OverlayCracked = z\ace\addons\goggles\textures\HUD\CombatGogglesCracked.paa; \
+        ACE_Resistance = 2; \
+        ACE_Protection = 1;
```
jonpas: This would look better like that:
```
#define COMBAT_GOGGLES \
    ACE_Overlay = z\ace\addons\goggles\textures\HUD\CombatGoggles.paa; \
    ACE_OverlayCracked = z\ace\addons\goggles\textures\HUD\CombatGogglesCracked.paa; \
    ACE_Resistance = 2; \
    ACE_Protection = 1;
```

bux578:
Hey @FFAAMOD thank you for your addition.
Very much appreciated. And awesome job on your update!

optionals/compat_ffaamod/CfgGlasses.hpp
FFAAMOD: Done, thanks.

FFAAMOD:
Ups, sorry I just tweaked the config files directly into the GitHub web and create many commits instead of creating a single commits with all changes :/

jonpas:
That's fine, we can squash commits during merge.

You are still putting braces on the newline, they should be on the same as the coding guidelines say.

FFAAMOD:
I really don't know why my desktop GitHub app doesn't work in my PC (It throws an error when I try to clone). Anyway, changed the files as coding guidelines says regarding to open brackets.

Mickyleitor:
How is this going? It seems this is untagged but is there something i can do in order to merge this into main faster?

Mickyleitor:
How is this going? It seems this is untagged .. Is there anything I can do in order to push this into main branch @jonpas @bux578 ?

jonpas:
I am slightly confused why this wouldn't be distributed from your (FFAA mod) side, for example like CUP does it. It would be easier to update and maintain on your side without being reliant on when we updated ACE3.

Mickyleitor:
@jonpas there is some reasons. The more relevant is our time between every update takes aprox 6-7 months so all ACE3-FFAA issues detected within this time can't be fixed. Uploading it here (as RHS, RH,R3F do) will allow us (and indepent developers) fix those issues (all that can be fixed by changing ACE3 compat pbo only). And unfortunately we can't reach all issues related to ACE3-FFAA by own because we lack developers focused in this compat addon (I mean, giving full support), which can easily resolved by allowing developers by ACE/freelance to make pull requests.

esteldunedain:
It should be clear that the ACE team may not be able to fully support this compat pbo; specially because I believe none of the regular devs uses FFAA mod. Also keep in mind that we are not doing proper releases of compat pbos; we are currently just packing the current state of the pbos when ACE is released.

On the other side, we did have internal talks about allowing third party mod makers to store their compat pbos on our repos (sepparate repos IIRC), and maybe even facilitating proper releases.

So I think a decision is required. For what it counts, I'm for merging this as-is while we figure out how to best deal with compat pbo releases.

Thoughts @Glowbal, @bux578?

jonpas:
I see, in that case I agree with estel.

Mickyleitor:
@esteldunedain Great, waiting for that decision. I'm fine if this finally is a no-go to ACE repository, just an aditional download in our webpage (or inside our mod folder) on every release. But I thought when I saw that there is currently an additional RHS compat inside ACE3 repository it could be a good idea to store it here because freelance devs can check and fix issues (or make improvements) by his own instead of our team giving full support to this compat pbo which I'm afraid we can't be forever (I'm the only one which is making changes in GitHub and I don't know if RHS are giving full support to his rhs compat in this repository by the way).

esteldunedain:
@Mickyleitor, I agree with you.

RHS is the worst possible example, because they don't care about compat at all; we took to the job of providing a compat pbo for it as a service to the Arma community, but it's against our policy, which is to have the mod makers actively contributing to their compat work, with the help of the ACE team and the community as a whole.

Glowbal:
Happy to merge this but needs to be cleaned up a bit. Indentation is all over the place.

Note to whoever merges: This should also be squashed into a single commit with a descriptive message.

We will have another internal discussion on how to handle this in the future. Thanks @FFAAMOD.

commy2:
>Removed newline at EOF.

That's bad. We always put `newline` after the last line.
http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

PR #3730, opened 2016-04-28
Cleanup reload launchers

PabstMirror:
- Cleanup reload launchers

Minor cleanup of params, and adds a fast exit for non supported weapons.

**Todo**: None of the launcher seem to have a good config for reload time and right now it's a 12 second progress bar which make this module pretty hard to use.

```
        magazineReloadTime = 12;
```

I can't really find any good config value for the real reload time (about 3 seconds), it might be linked to the reload animation??

### DIFF
--- a/addons/reloadlaunchers/functions/fnc_load.sqf
+++ b/addons/reloadlaunchers/functions/fnc_load.sqf
@@ -1,6 +1,5 @@
 /*
  * Author: commy2
- *
  * Reload a launcher
  *
  * Argument:
@@ -11,15 +10,13 @@
  *
  * Return value:
  * NONE
+ *
+ * Public: No
  */
 #include "script_component.hpp"
 
-private ["_unit", "_target", "_weapon", "_magazine"];
-
-_unit = _this select 0;
-_target = _this select 1;
-_weapon = _this select 2;
-_magazine = _this select 3;
+params ["_unit", "_target", "_weapon", "_magazine"];
+TRACE_4("params",_unit,_target,_weapon,_magazine);
 
 private "_reloadTime";
 _reloadTime = getNumber (configFile >> "CfgWeapons" >> _weapon >> "magazineReloadTime");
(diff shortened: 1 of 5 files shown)


commy2:
Hand weapons don't use the magazineReloadTime property. Magazines are reloading instantly and only the animation is preventing the player from shooting.
It is used by vehicles mainly (they obviously have no reload animation)
I think the artificial delay from addWeaponItem is unrelated to this value and the delay is some kind of magic number in the engine. Don't know for sure though.

PabstMirror:
@commy2 Should we just hard code it to be 2-3 second then?
The receiver still has to wait for the ammo load.

esteldunedain:
> Should we just hard code it to be 2-3 second then?

:+1:

commy2:
For now. I don't see them fixing this bug anytime soon.

addons/reloadlaunchers/functions/fnc_load.sqf
```diff
@@ -11,18 +10,16 @@
...
-_target = _this select 1;
-_weapon = _this select 2;
-_magazine = _this select 3;
+params ["_unit", "_target", "_weapon", "_magazine"];
+TRACE_4("params",_unit,_target,_weapon,_magazine);
 
-private "_reloadTime";
-_reloadTime = getNumber (configFile >> "CfgWeapons" >> _weapon >> "magazineReloadTime");
+// private _reloadTime = getNumber (configFile >> "CfgWeapons" >> _weapon >> "magazineReloadTime"); //Not a good config value, use a constant for now:
+private _reloadTime = 2.5;
```
bux578: Wouldn't it be better to introduce a new config value?
Maybe some launchers in the CUP or RHS weapon pack need more reload time?

PabstMirror:
Using @bux578 's suggestion, attempts to use `QGVAR(buddyReloadTime)` from weapon's config, else default to 2.5 seconds.

PR #3731, opened 2016-04-28
Update medical feature doc

alganthe:
- Update the bandages effectiveness table (thanks to daishiknyte).
- Added adenosine to the doc.


yozer247:
Hey, I've never used Github before but wanted to contribute, hopefully I'm doing this in the right place.

A while back I created a google doc table reading from the code in the ace_treatments file, same format as the table alganthe made. A few things I wanted to say:

First is that I noticed a couple values from our two tables were different, first noticed in the 'cut' effectiveness on field dressing. Might want to make a quick review over the stats on the table to make sure they are up-to-date and not translated incorrectly.

Second is that while a large table with all the stats is certainly informative, it doesn't give the information people will be looking for, which is what bandage type goes best with what wound type. In my google doc I compared each of the stats until I could create a 'final table' which had two columns for each wound type. A column for the 'most effective' bandage, and a column for the 'least re-open chance'. I think it would be appropriate to add this format of table into the documentation along with the stats table. Another thing to possibly consider is to add instead a re-worked version of the 'effectiveness table' currently on the site but re-done to accurately show the values.

Lastly, I just wanted to mention how fundamentally wrong the 'effectiveness table' is. Maybe ace was updated and values changed over time, but the information that the effectiveness table currently shows is very wrong in a lot of ways. This is a rather big problem as it is the only official guide of bandages, the core feature of advanced medical, given to the playerbase directly from an ace3 resource. This means a lot of guides and players follow this table, trusting it to be correct. The table has also created the false impression to many players that you only need to use two types of bandages, the 'packing' for velocity wounds and avulsions, and then 'elastic' for all other wounds, since combined those two bandage types cover all wounds under the 'highest effectiveness' category in the table. While in reality, elastic is actually the worst at a few of the wound types where it is labelled as 'highest' effectiveness.

Hopefully I've given some sound feedback that'll be taken into consideration on this issue, and hopefully the ace team see the importance of replacing the currently very misleading information on their site.

All the best :)
-Yozer

Link to the Google Doc : https://docs.google.com/spreadsheets/d/1AKLA_iuImpvW-sXrcmgO1PEPba0LDKOO1oWt372RT3M/edit?pref=2&pli=1#gid=0

alganthe:
It is indeed the spreadsheet that you create, @daishiknyte gave it to me, I'm currently working on a similar spreadsheet with color coding, some of the values are indeed wrong https://github.com/acemod/ACE3/blob/master/addons/medical/ACE_Medical_Treatments.hpp

wiki/feature/medical-system.md
```diff
@@ -214,9 +219,12 @@ Use the appropriate amount depending on the situation (heavy loss of blood, bloo
 
 Autoinjector | Effect
 ---------- | ---------- |
-Morphine | lower the blood pressure and heart rate of the patient, also suppress pain
-Epinephrine | raise the heart rate of the patient
-Atropine | lower the heart rate of the patient
+Morphine | Decreases the blood viscosity, suppress pain
+Epinephrine | raises the heart rate of the patient
+Adenosine | Counters the effects of epinephrine
```
Glowbal: It doesn't - it lowers heart rate.

alganthe:
bumpidy bump

bux578:
Merge conflicts :)

alganthe:
yeah, the commit was uploading when I posted

bux578:
Thanks bae!

PR #3734, opened 2016-04-28
Fix getting stuck on rope whilst fastroping

BaerMitUmlaut:
- Fix getting stuck on rope whilst fastroping
- Fix the twitching issues that appeared again with 3.5.1


jonpas:
We (4 of us, thanks to @BaerMitUmlaut @alganthe and whatever YoursTruly's GitHub name is) tested it on a dedicated server, worked smoothly with this fix.

PR #3747, opened 2016-05-02
Changes for Linux development and armake

KoffeinFlummi:
Contains various changes for building on Linux or compatibility with [armake](https://github.com/KoffeinFlummi/armake/).

TODO:
- [x] Git commit hash in bisign name
- [x] Update versions in files (auto-commit, optional push)
- [x] Include used A3 `.rvmat`s
- [x] Expand documentation with new recipes (eg. `version`)
- [x] Update binaries to armake 0.6.1 which includes https://github.com/KoffeinFlummi/armake/pull/65 and use `-e version=<git-hash>` and `-s <signature name>`
- [x] Disable excessive concatenation (when ignoring warnings from used libraries implemented or CBA fixes it)
- [x] Auto-download armake Windows binaries
- [x] Use `echo -e` for version file updates where newlines are used
- Rework Makefile to standard release/publish procedure (ref. https://github.com/CBATeam/CBA_A3/pull/730#issuecomment-411018642) -> moved to #6363 
- Fix make.ps1 on Windows 8 (ref. https://github.com/acemod/ACE3/pull/3747#discussion_r192170181) -> moved to #6363 

Notes:
- Binarization only fully works on Windows using BIS Binarize


bux:
Can we split that PR into changes for the build (armake/makefile) and code changes?
Seems those code changes would make sense even without armake.

Neviothr:
I'm assuming all of this will later be added to Armake's repo as well?

extras/userconfig/ace/serverconfig.hpp
```diff
@@ -0,0 +1,6 @@
+
```
jonpas: What the hell is this doing here.

extensions/CMakeLists.txt
```diff
@@ -90,7 +90,7 @@ endif()
 string(TIMESTAMP ACE_BUILDSTAMP "%Y-%m-%dT%H:%M:%SZ")
 set(ACE_VERSION_MAJOR 3)
 set(ACE_VERSION_MINOR 12)
-set(ACE_VERSION_REVISION 0)
+set(ACE_VERSION_REVISION 1)
```
jonpas: Nope.

KoffeinFlummi:
@dedmen 

The entire point of armake was to get the laughably complicated dev environment setup instructions down to - ideally - zero, with all that's needed being the Arma 3 Tools and the repo.

While bundling binaries is obviously not ideal, the size is negligible (the 2 binaries have the size of 1 moderately big texture) and we already do it with dlls, for precisely the same reason - making it easier for non-developers to quickly create a current build.

Having different versions of armake in circulation for various projects is not really a problem, as long as the same version is used for the same project.

If it can be reliably automated I'm all for it, but in the absence of a proper package management option for windows this is the best option.

BTW: For the BWMod i also included a make binary to allow proper partial builds for noobs, and it has definitely increased the amount of people regularly building their own versions.

dedmen:
>If it can be reliably automated I'm all for it, but in the absence of a proper package management tool for windows this is the best option.

How about using powershell instead of batch? I don't think we need to support Windows Vista or older which doesn't have powershell. That can just check if the armake binary is there. If not then download it. If it is then check if there is a newer version available. Might aswell just hardcode the download link and update it manually.
With batch, downloading files is not that nice (https://stackoverflow.com/questions/4619088/windows-batch-file-file-download-from-a-url)

Or maybe can we atleast remove the x64 armake binary? I don't see a need for a 64bit version if the 32bit runs just as well. Ideally we should just get rid of 32bit as it's time has run out long ago but there are still people running 32bit OS's.

You cooooould theoretically get BI binarize the same way. That way env setup is just a single click onto make.ps and is done automatically. But we can't redistribute BI binarize.
As alternative one could use https://developer.valvesoftware.com/wiki/Steam_browser_protocol to ask steam to install Arma 3 Tools for the user.

BaerMitUmlaut:
Funny that you say this, we've decided to do exactly that yesterday. Which is why jonpas assigned me, I'm making that PS script right now.

BaerMitUmlaut:
`make.ps1` now replaces `make.bat` and downloads the most recent version of armake if necessary.

dedmen:
Thanks baer. I love it o7.

tools/make.ps1
```diff
@@ -0,0 +1,130 @@
...
+
+    New-Item "$PSScriptRoot\temp" -ItemType "directory" -Force | Out-Null
+
+    Write-Output "Downloading armake..."
+    $client = New-Object Net.WebClient
+    $client.DownloadFile($url, "$PSScriptRoot\temp\armake.zip")
+    $client.dispose()
+
+    Write-Output "Download complete, unpacking..."
+    Expand-Archive "$PSScriptRoot\temp\armake.zip" "$PSScriptRoot\temp\armake"
```
dedmen: Doesn't exist on windows 8.
Alternative: https://stackoverflow.com/a/27768628/6216977

Works for me.

jonpas: @BaerMitUmlaut ?

tools/make.ps1
```diff
@@ -0,0 +1,130 @@
...
+        $directory
+    )
+
+    $component = $directory.Name
+    $fullPath  = $directory.FullName
+    $parent    = $directory.Parent
+    $binPath   = "$projectRoot\@ace\$parent\ace_$component.pbo"
+
+    Write-Output "  PBO  $component"
+    & $armake build -i "$projectRoot\include" -w unquoted-string -w redefinition-wo-undef -f $fullPath $binPath
```
dedmen: Missing the `version` header entry?

jonpas: This is just for quick builds on Windows, not for full release builds. But we can still add it.

jonpas:
This is now fully functional for CI. We can add other TODOs above in another PR. :)

dedmen:
Soooo.. what about https://github.com/acemod/ACE3/pull/3747#discussion_r192170181 ?

jonpas:
Read the comment above...

dedmen:
Have a hundred notifications. Commenting before I got through them all ^^

PR #3752, opened 2016-05-03
Fix "remove dead bodies on disconnect" when enabled and forced

Cuel:
When the server set "remove dead bodies on disconnect" to true and forced it, it would never actually register unless the module was placed in the mission.

Move the addMissionEventHandler to a server XEH postInit, so that it will run even though the module is not placed, allowing it to register removal when forced in server config.

Also, removed seemingly unused GVAR in fnc_module.sqf
`GVAR(Module) = true;`

I also removed some whitespace alignment in fnc_module.sqf


PabstMirror:
Just fyi - the ace_setting for has always been 1, so it will now run by default which may be unexpected.

PR #3760, opened 2016-05-05
Repair - Improve wheel detection for RHS

### LINKED ISSUE #3759: Unable to replace front right tire on RHS M1025A2.
**Arma 3 Version:** `1.58` (stable)
**CBA Version:** `2.3.1` (stable)

**Mods:**
- `@CBA_A3`
- `@ace3'
- '@RHS_USF3'

**Description:**
When attempting to replace the front right wheel on the RHS M1025A2 you will instead damage the engine and the tire will not be affected.

**Steps to reproduce:**
- Place down an RHS M1025A2 (Example class: "rhsusf_m1025_d_m2").
- Attempt to remove the front right tire.
- After the action is complete, a tire will be removed and laying near the Humvee, but the tire will still be on the Humvee.
- Get in the vehicle as the driver and notice the engine is now red and will not turn on.

- Dedicated Multiplayer / Editor

**Placed Modules:**
None.

**RPT log file:**
http://pastebin.com/MELBL1AM

PabstMirror:
- Fix #3759

The backup wheel detection just tries to find the closest hitpoint to the wheel, and for one RHS vic the engine is close than the actual wheel hitpoint. This just adds a blacklist for the engine.

image

### DIFF
--- a/addons/repair/functions/fnc_getWheelHitPointsWithSelections.sqf
+++ b/addons/repair/functions/fnc_getWheelHitPointsWithSelections.sqf
@@ -72,7 +72,7 @@ _wheelHitPointSelections = [];
                 if (_x != "") then {
                      //Filter out things that definitly aren't wheeels (#3759)
                     if ((toLower (_hitPoints select _forEachIndex)) in ["hitengine", "hitfuel", "hitbody"]) exitWith {TRACE_1("filter",_x)};
-                    _xPos = _vehicle selectionPosition [_x, 'HitPoints'];
+                    _xPos = _vehicle selectionPosition _x;
                     if (_xPos isEqualTo [0,0,0]) exitWith {};
                     _xDist = _wheelCenterPos distance _xPos;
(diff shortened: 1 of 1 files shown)


addons/repair/functions/fnc_getWheelHitPointsWithSelections.sqf
```diff
@@ -70,7 +70,9 @@ _wheelHitPointSelections = [];
             _bestIndex = -1;
             {
                 if (_x != "") then {
-                    _xPos = _vehicle selectionPosition _x;
+                     //Filter out things that definitly aren't wheeels (#3759)
+                    if ((toLower (_hitPoints select _forEachIndex)) in ["hitengine", "hitfuel", "hitbody"]) exitWith {TRACE_1("filter",_x)};
+                    _xPos = _vehicle selectionPosition [_x, 'HitPoints'];
```
jonpas: Single quotes?

PR #3761, opened 2016-05-05
Fix explosive sync on dedicated server

PabstMirror:
- Move server event for explosives sync above `if (!hasInterface) exitWith {};`
- Use `CBA_fnc_getSharedGroup`


kwinteng:
Does this fix the issue where people can't blow up explosives that aren't placed by them?

jonpas:
Yep it should.

PR #3762, opened 2016-05-06
Add key to toggle wind info display

PabstMirror:
- Add key to toggle wind info display

This would match the default behaviour of the compass, where it only shows up while the key is held down.  By default key is unbound so this will have no effect unless chosen by user.

Tag @Gwynblade


Gwynblade:
Oh wow that was quick, many thanks!

PR #3766, opened 2016-05-07
Remove difficulties component

jonpas:
- Remove `difficulties` component
- Component is redundant as of Arma 3 1.58 due to no longer using `CfgDifficulties`, new class is `CfgDifficultyPresets`
- No point moving it as it can be set in .ArmA3Profile on server and @commy2 said this was ported from AGM because server admins were lazy. :laughing:


commy2:
bye bye, old friend. you were of great use in Alpha / Beta times. :salute:

PabstMirror:
Just a reminder, because we are removing a PBO, we should make sure to include a note like:

>Make sure to delete your existing `@ace` folder when updating 

or we will get reports of version mismatch errors from next release.

jonpas:
It can't hurt to make a note indeed.

PR #3770, opened 2016-05-09
Moved IR Strobe to left shoulder when attached.

tbeswick96:
- Moves attachment position of IR strobe/grenade to left shoulder. Previously, when on the right shoulder, the strobe was visible as a reflection when looking down some optics (Elcan from 3CB) which was intrusive whilst aiming.


PabstMirror:
If no objections to this, should we move to 3.6?

esteldunedain:
I agree

PR #3780, opened 2016-05-10
Adjusted weight of bodybag

BaerMitUmlaut:
- Adjust the weight of the bodybags to ~330g ([reference](https://l3.cdnwm.com/ip/celulosas-vascas-leichensack-mit-reissverschluss-leichensack-365713.pdf))


commy2:
mass is an integer. should be 7 or 8 then.

PR #3781, opened 2016-05-11
Fix spectator unit tree refreshing

SilentSpike:

- Account for decrement in node index when culling the tree of sides/groups/units
- Fix non-zero-based index `to` value when navigating the units in the tree

### DIFF
--- a/addons/spectator/functions/fnc_handleInterface.sqf
+++ b/addons/spectator/functions/fnc_handleInterface.sqf
@@ -27,8 +27,11 @@ switch (toLower _mode) do {
         // Always show interface and hide map upon opening
         [_display,nil,nil,!GVAR(showInterface),GVAR(showMap)] call FUNC(toggleInterface);
 
+        // Initalize the unit tree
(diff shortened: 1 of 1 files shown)


SilentSpike:
Tested and the unit tree seems to refresh as expected, though it's possible I missed something (since this fix accounts for multiple nodes being removed/added simultaneously it can be hard to test every combination).

addons/spectator/functions/fnc_handleInterface.sqf
```diff
@@ -331,31 +331,40 @@ switch (toLower _mode) do {
     };
     case "onunitsupdate": {
         _args params ["_tree"];
-        private ["_cachedUnits","_cachedGrps","_cachedSides","_s","_g","_grp","_u","_unit","_side"];
+        private ["_cachedUnits","_cachedGrps","_cachedSides","_sT","_gT","_uT","_s","_g","_u","_grp","_unit","_side"];
```
Glowbal: those variable names need to be improved sometime

SilentSpike: :smile: Fair, I kept them short because they're just used for iteration

Glowbal:
Lets fix those variable names in a different PR.

PR #3785, opened 2016-05-12
Update coding guidelines

Glowbal:
Minor updates to our coding guidelines

### DIFF
--- a/wiki/development/coding-guidelines.md
+++ b/wiki/development/coding-guidelines.md
@@ -58,7 +58,7 @@ class ACE_Settings {
 All text that will be displayed to a user shall be defined in a `stingtable.xml` file for multi-language support.
 
 * There will not be any empty stringtable language values.
-* All stringtables will follow the format as specified by Tabler and the [translation guidelines](http://ace3mod.com/wiki/development/how-to-translate-ace3.html) form .
+* All stringtables will follow the format as specified by Tabler and the [translation guidelines](http://ace3mod.com/wiki/development/how-to-translate-ace3.html) form.
 
 
 ## 2. Macro Usage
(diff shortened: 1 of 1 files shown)


wiki/development/coding-guidelines.md
```diff
@@ -380,18 +382,18 @@ If a function returns error information, then that error information will be tes
...
 
 Usage of the CBA Macro `PARAM_x` or `BIS_fnc_Param` is deprecated and not allowed within the ACE project unless very specific reasons allow no other alternative.
 
 ### 6.4. Return Values
-Functions and code blocks that specific a return a value must have a meaningfull return value.
+Functions and code blocks that specific a return a value must have a meaningfull return value. If no meaningfull return value, the function should return nil.
 
 ### 6.5. Private Variables
-All private variables shall be declared in a private array.
+All private variables shall make use of the `private` keyword on initalization. When declaring a private variable before initalization, usage of the private array syntax is allowed. All private variables must be either initialized using the private key word, or declared using the private array syntax.
```
commy2: Or by using `params` I guess.
`params` is not exclusive to the functions arguments, but is sometimes also used in forEach loops (`_x params [...]`), so it might be worth noting.

jonpas: Or any array really. It's the same as `EXPLODE_N_PVT` macro was.

wiki/development/coding-guidelines.md
```diff
@@ -400,13 +402,47 @@ Any one function shall contain no more than 250 lines of code, excluding the fun
 ### 6.7. Variable declarations
 Declarations should be at the smallest feasible scope.
 
+
+Good:
+```sqf
+if (call FUNC(myCondition) then {
```
654wak654: Missing `)` ?

Glowbal:
Added both the bit about params usage for arrays and fixed the missing `)`. Thanks @654wak654.

commy2:
:+1:

wiki/development/coding-guidelines.md
```diff
@@ -380,33 +382,80 @@ If a function returns error information, then that error information will be tes
...
+   if (_areAllAboveTen) then {
+       hint "all values are above ten!";
+   };
+}
+```
+
+Bad:
+```sqf
+private _areAllAboveTen = true; // <- this is bad, because it can be initalized in the if statement
+if (call FUNC(myCondition) then {
```
654wak654: I didn't think it was worth to mention, but this line is also missing the same closing parenthesis.

PR #3787, opened 2016-05-12
Add interaction support to Motorcycle class

jonpas:
- Adds interaction support to `Motorcycle` class - for example CUP's M1030.

**Known issues:**
- Wheel repairing does not work for CUP's M1030 because it is not PhysX configured and ACE repair component expects it to be.

I probably missed a bunch of stuff that should be added, name them please!

### DIFF
--- a/addons/repair/CfgEventHandlers.hpp
+++ b/addons/repair/CfgEventHandlers.hpp
@@ -31,6 +31,12 @@ class Extended_InitPost_EventHandlers {
             serverInit = QUOTE(_this call DFUNC(addSpareParts));
         };
     };
+    class Motorcycle {
+        class ADDON {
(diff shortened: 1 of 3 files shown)


esteldunedain:
Looks good!

addons/repair/CfgEventHandlers.hpp
```diff
@@ -28,7 +28,11 @@ class Extended_InitPost_EventHandlers {
     class Tank {
         class ADDON {
             init = QUOTE(_this call DFUNC(addRepairActions));
-            serverInit = QUOTE(_this call DFUNC(addSpareParts));
```
commy2: isn't this for spare tracks?

jonpas: Oh damn, removed it from the wrong class. :laughing:

PR #3788, opened 2016-05-12
Remove redundant spare part adding and improve addSpareParts function

jonpas:
- Removes initial spare part adding for all classes except `Car` and `Tank` - it is used only to add spare wheels and tracks to those 2 classes.
- Improves `FUNC(addSpareParts)` - due to above manual checking is not needed, it just added overhead, this also allows improved checking without adding it to `runAtSettingsInitialized` even if no part name is supplied.


commy2:
I like this better.

PR #3791, opened 2016-05-14
Remove old test-mission

PabstMirror:
I added this about a year ago before the v3.0.0 release.
I think it can be cleaned up now as I don't think it gets any use and the modules are outdated.


jonpas:
Probably better to just create a new one from scratch if we really want one anyways.

PR #3792, opened 2016-05-14
use 'getUnitLoadout' instead of 'FUNC(getAllGear)' and deprecate it

commy2:
- title
- massive performance improvement every frame

```
B_Soldier_F

getUnitLoadout player
0.0251 ms
10000/10000

player call ACE_common_fnc_getAllGear
0.535412 ms
1892/10000
```

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -333,7 +333,7 @@ enableCamShake true;
 GVAR(OldPlayerVehicle) = vehicle objNull;
 GVAR(OldPlayerTurret) = [objNull] call FUNC(getTurretIndex);
 GVAR(OldPlayerWeapon) = currentWeapon objNull;
-GVAR(OldPlayerInventory) = [objNull] call FUNC(getAllGear);
+GVAR(OldPlayerInventory) = [];
 GVAR(OldPlayerVisionMode) = currentVisionMode objNull;
 GVAR(OldCameraView) = "";
 GVAR(OldVisibleMap) = false;
@@ -381,7 +381,7 @@ GVAR(OldIsCamera) = false;
     };
 
     // "playerInventoryChanged" event
-    _data = [ACE_player] call FUNC(getAllGear);
+    _data = getUnitLoadout ACE_player;
     if !(_data isEqualTo GVAR(OldPlayerInventory)) then {
         // Raise ACE event locally
         GVAR(OldPlayerInventory) = _data;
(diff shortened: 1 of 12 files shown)


SilentSpike:
:+1: Man that `switch` block was ugly. All looks good to me - assuming that `FUNC(inventoryCheck)` in scopes defaults to `ACE_Player` anyway.

Glowbal:
Perfect :+1:

commy2:
>Man that switch block was ugly.
Yeah. It really sucked that `assignedItems` reports them in a undefined order. Now with `getUnitLoadout` you always know what map/watch/gps the unit actually has or if it's `""`.

esteldunedain:
Yay!!

PabstMirror:
Just to be aware: `getUnitLoadout` includes magazines ammo, so it changes with each bullet fired

Which means we throw the `playerInventoryChanged` event each round fired:
`ace_common_playerInventoryChanged_counter: Average: 0.121994s / 256 = 0.476539ms`

It is a change in behaviour from before, and the event is expensive.

So we save ~0.07ms speedup in the common case, but get a ~0.5ms slowdown when firing.

esteldunedain:
@PabstMirror, do you think there's a quick way to exclude the current ammo of the loaded magazine from the `isEqual` test? Maybe overwriting it in the array before comparing it?

I think that should probably be good enough, because we'd still detect reloads, as in those the content of the inventory also changes.

esteldunedain:
Oh, btw. According to @commy2's number, this PR saves 0.47ms each frame, so if those are right then the impact when firing should compensate.

commy2:
We could add another step to the PFH that somehow filters out all the numbers from the array and only  do the event when something else changes.

esteldunedain:
I was thinking of just clearing out the current magazines in the array before comparing them

PabstMirror:
From same source for comparison:

0.0904 ms `ACE_common_fnc_getAllGear`
0.011 ms `getUnitLoadout`
0.47 ms playerChanged event

It might be fine as is:

600rnds/sec = ~5ms per second when full auto wasted on playerChanged event
(0.8*50 fps) = ~4ms per second saved on loadout checking (faster no matter if firing or not!)

esteldunedain:
Sure, but if we could filter out the inventory event from firing it would be even better.

PabstMirror:
Something like this would filter out currently loaded mag:
```
_data = getUnitLoadout ACE_player; 
if (!((_data select 0) isEqualTo [])) then { 
    (_data select 0) set [4, -1];
    (_data select 0) set [5, -1];
}; 
if (!((_data select 1) isEqualTo [])) then { 
    (_data select 1) set [4, -1]; 
    (_data select 1) set [5, -1];
}; 
if (!((_data select 2) isEqualTo [])) then { 
    (_data select 2) set [4, -1];
    (_data select 2) set [5, -1];
};
```

esteldunedain:
I bet that should take way less than 05 ms

commy2:
I already have made that. PR incoming.

PR #3793, opened 2016-05-14
skip 'playerInventoryChanged' after firing

commy2:
- don't do 'playerInventoryChanged' just because ammo decreased after firing

ref: #3792


esteldunedain:
@commy2, now that the `playerInventoryChanged` event performs well, it's probably the time to put it on CBA.

commy2:
@esteldunedain https://github.com/CBATeam/CBA_A3/pull/328/files

PR #3795, opened 2016-05-15
Add Deprecate components support

Glowbal:
- Add support for deprecating components
- Deprecrates ACE Sitting. Moved to ACEX Sitting.

### DIFF
--- a/addons/sitting/XEH_clientInit.sqf
+++ b/addons/sitting/XEH_clientInit.sqf
@@ -5,7 +5,7 @@ if (!hasInterface) exitWith {};
 
 GVAR(isEnabled) = false;
 ["SettingsInitialized", {
-    if ([[QUOTE(ADDON), QGVAR(enable)], ["acex_sitting", "acex_sitting_enable"], "3.7.0"] call EFUNC(common,deprecateComponent)) exitwith {};
+    if ([[QUOTE(ADDON), QGVAR(enable)], ["acex_sitting", "acex_sitting_enable"], "3.8.0"] call EFUNC(common,deprecateComponent)) exitwith {};
     //If not enabled, then do not add CanInteractWith Condition or event handlers:
     if (!GVAR(enable) || GVAR(isEnabled)) exitWith {};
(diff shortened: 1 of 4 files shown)


addons/sitting/XEH_clientInit.sqf
```diff
@@ -3,11 +3,12 @@
...
 
+GVAR(isEnabled) = false;
 ["SettingsInitialized", {
-    TRACE_1("SettingInit", GVAR(enable));
-
+    if ([[QUOTE(ADDON), QGVAR(enable)], ["acex_sitting", "acex_sitting_enable"], "3.8.0"] call EFUNC(common,deprecateComponent)) exitwith {};
     //If not enabled, then do not add CanInteractWith Condition or event handlers:
-    if (!GVAR(enable)) exitWith {};
+    if (!GVAR(enable) || GVAR(isEnabled)) exitWith {};
+    GVAR(isEnabled) = true;
```
jonpas: What is that for again?

Glowbal: Ensures that the module will not load in twice

jonpas: How could it? Unless someone put 2 of those down (in which case he should get all the RPT warnings).

Glowbal: You are quite right - this isn't necessary here, but it is for the port to ACEX. It should not have been commited with this PR.

jonpas: How are we going to handle that for components which have modules that can be placed multiple times (eg. slideshow)?

Glowbal: Same way we handle functions that we deprecate.

jonpas: Never mind, slideshow doesn't even have a XEH postInit, was thinking of something else.

jonpas:
@Glowbal Will this print warnings for sitting? If so it has to be taken out before release.

PR #3797, opened 2016-05-15
Add backwards compatability for InventoryChanged

Glowbal:
The output of `getUnitLoadout` is different as `getAllGear`. Using a different format in the event would mean that any eventhandler depending on the inventory output would now be broken.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -385,7 +385,7 @@ GVAR(OldIsCamera) = false;
     if !(_data isEqualTo GVAR(OldPlayerInventory)) then {
         // Raise ACE event locally
         GVAR(OldPlayerInventory) = _data;
-        ["playerInventoryChanged", [ACE_player, _data]] call FUNC(localEvent);
+        ["playerInventoryChanged", [ACE_player, [ACE_player, false] call FUNC(getAllGear) ]] call FUNC(localEvent);
     };
 
     // "playerVisionModeChanged" event
--- a/addons/common/functions/fnc_assignedItemFix.sqf
+++ b/addons/common/functions/fnc_assignedItemFix.sqf
@@ -29,9 +29,9 @@ GVAR(AssignedItemsShownItems) = [
 ];
 
 ["playerInventoryChanged", {
-    params ["_unit", "_assignedItems"];
+    params ["_unit"];
 
-    _assignedItems = _assignedItems param [9, ["","","","","",""]]; // ["ItemMap","ItemGPS","ItemRadio","ItemCompass","ItemWatch","NVGoggles"]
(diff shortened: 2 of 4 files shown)


PabstMirror:
#3792 marked getAllGear as deprecated

Glowbal:
Yes, but we need the output to be the same, or we will be breaking all events depending on the output.

Example:
https://github.com/acemod/ACE3/blob/master/addons/parachute/functions/fnc_storeParachute.sqf#L20

Currently throws the following script error:
```sqf
23:54:04 Error in expression <t", "_gear"];
private _backpack = _gear select 6;

if ((vehicle _unit) isKindOf >
23:54:04   Error position: <select 6;

if ((vehicle _unit) isKindOf >
23:54:04   Error Zero divisor
23:54:04 File z\ace\addons\parachute\functions\fnc_storeParachute.sqf, line 5
```

This is because the `select 6` returns the helmet classname instead of the backpack.

esteldunedain:
:+1:

Glowbal:
Another possible suggestion I have is introducing a secondairy event and deprecating `playerInventoryChanged`. Use the new syntax for the new event and remove the old event in 3.7.0.

PabstMirror:
We changed how googles handles event, so we need tor revert that as well.
https://github.com/acemod/ACE3/pull/3792/files#diff-c4ecc44c622ec3f35d4479e1014c6138L48

We could do something like:
```
["playerLoadoutChanged", _data] call FUNC(localEvent);
if (!isNil {GVAR(eventsLocation) getVariable "playerInventoryChanged"}) then {
    //deprecated warning:
    _oldData = ..FUNC(getAllGear);
    ["playerInventoryChanged", _oldData] call FUNC(localEvent);
};
```

Glowbal:
That could work @PabstMirror 

That is most likely a better solution. Replace all playerInventoryChanged by a new event within ACE, keep the old one around in case anyone used it and remove it in 3.7.0. 

Note, getAllGear also prints an deprecated warning. So we probably want to find a solution for that.

jonpas:
Optional param for internal stuff using getAllGear,  if set don't print warning.

commy2:
like this? :+1: from me

Glowbal:
This would work - However I feel like we should go down @PabstMirror's road and introduce that secondary event and make this one only trigger if it's actually used. Then we can get rid of it in the long run.

We would internally only make use of the new event, that uses the `unitLoadout` command format.

commy2:
>in the long run

You might have missed this:
https://github.com/CBATeam/CBA_A3/blob/0c3c2bdef17059f72ecf48dcebd078777bd39368/addons/events/fnc_addPlayerEventHandler.sqf

I think that will be the future for this anyway.

esteldunedain:
I agree with @commy2. The future is using CBA event with its new format.

Glowbal:
In that case, I reckon this is the best solution for now :+1: Merge?

PR #3800, opened 2016-05-16
Fix UBC errors for RHS 4.1

PabstMirror:
- Fix UBC errors for RHS 4.1


Twelvecaliber:
Is this a fix for the compatibility PBO's that now don't work with RHS 4.1 due to the changes to cfgpatches, or is this a different issue?

PabstMirror:
@Twelvecaliber - This PR seemed to fix the Updating Base Class errors from some weapon classnames changing.

I never saw any problems with CfgPatches, so would be helpful if I could get an RPT from the issue you are having.

Issue #3765, opened 2016-05-07
setHearingCapability - Both medical and Hearing modify sound levels


PabstMirror:
Ref discussion in #3710

PabstMirror:
I looked at this as part of PR #3802 and I'm not sure if this is necessary.
Medical only modifies the hearing levels of voices from radio mods (ACRE/TFR).

Glowbal:
https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_setHearingCapability.sqf

PR #3802, opened 2016-05-17
Hearing - New XEH, Fix dead effects, Medical var

PabstMirror:
- CBA/addClassEventHandlers optimization
- Fix ear-ringing playing for units in spectator
- Medical now uses hearing's `unconsciousnessVolume` if it exists

### DIFF
--- a/addons/hearing/functions/fnc_updateVolume.sqf
+++ b/addons/hearing/functions/fnc_updateVolume.sqf
@@ -19,10 +19,7 @@
 if (!alive ACE_player) then {
     if (missionNameSpace getVariable [QGVAR(disableVolumeUpdate), false]) exitWith {};
     TRACE_1("dead - removing hearing effects",ACE_player);
-    1 fadeSound 1;
-    1 fadeSpeech 1;
-    ACE_player setVariable ["tf_globalVolume", 1];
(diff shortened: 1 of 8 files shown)


esteldunedain:
@PabstMirror: As far as I can tell `ace_hearing_fnc_updateVolume` already tweaks the volume for unconscious units. I really don't see the point of medical doing the same; anyway, this PR should remove that tweak if that's what we want.

Glowbal:
In #3765 I linked this function: https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_setHearingCapability.sqf

A solution like this would be ideal imo. It would not only support us modifying hearing volume, but also give support for third party mods doing so, without conflicting with anything we do.

PabstMirror:
Update to use function - Will close #3765

PabstMirror:
@esteldunedain - I agree, I removed setting the global volume in medical.
It will still set the outgoing voice settings on unconscious.

addons/hearing/functions/fnc_updateVolume.sqf
```diff
@@ -16,6 +16,12 @@
  */
 #include "script_component.hpp"
 
+if (!alive ACE_player) then {
```
Glowbal: Can this not be `exitWith` ? If `ACE_player` is not alive, is anything that comes after it still relevant?

PR #3805, opened 2016-05-20
Improve selective UI API

jonpas:
- Use config class `ACE_UI` instead of hardcoding elements
- Improve Selective UI API (`setElementVisibility)`, can now be called by `ACE_UI` config class name from already preset options to prevent issues:
```
[true, "ammoCount", false] call ace_ui_fnc_setElementVisibility;
```
- Add `"InterfaceInitialized"` event, executed when basic UI stuff is loaded and set from `ui` component
```
["InterfaceInitialized", {
    [true, "ammoCount", false] call ace_ui_fnc_setElementVisibility;
}] call ace_common_fnc_addEventHandler;
```
- Fix elements getting overwritten due to multiple executions of `"infoDisplayChanged"` event (executed 3 times when on foot) with some logical comparisons with `GVAR(elementsSet)` - connected to API as well
- Disable ammo count UI in `reload` component using Selective UI API
- Add hint when setting has been forced by API when trying to change it via ACE Options menu
- Fixes Stance UI not respecting settings (missing `infoDisplayChanged` event) - replaces #3806 
- Separates settings that share IDCs for on-foot and in-vehicle (eg https://github.com/acemod/ACE3/pull/3805#issuecomment-220880413)
- Disable gunner weapon zeroing UI in `fcs` component (actually just marks it as it is hidden by other config which is necessary for FCS to work)


PabstMirror:
Not specific to this PR, but right now we hide ammo counts for all vehicles.
In 3.5.1 we only hide it while on foot.

image

jonpas:
It was introduced with Selective UI, I will look into that.

jonpas:
@PabstMirror should be fixed, gunner and soldier elements are now separated (were using same IDCs, because thanks amazing config).

jonpas:
I think I am finally done with this, quite a rework as well.

addons/ui/ACE_UI.hpp
```diff
@@ -0,0 +1,177 @@
+// IDD, elements[]s, Element (must be string to compare to changed setting name), Selective Type (0-both, 1-ground, 2-vehicle)
+class ACE_UI {
+    class weaponName {
+        idd = 300;
+        elements[] = {118};
+        condition = QUOTE(ACE_player == vehicle ACE_player);
+        class conditions {};
```
Glowbal: What's the purpose of condition vs conditions here?

jonpas: Condition is the main condition (vehicle/soldier UI), conditions class is for anything else to add extra conditions, such as other components or mods.

Glowbal: Why not put our main condition into the conditions class?

jonpas: I had that at first, but this makes it easier to handle, because the main condition gets evaluated before any other and if not applicable it doesn't toggle the element at all (no point as you can't see it anyways), while other conditions determine if it should be visible or hidden.

Glowbal: Is there a different effect though? Both cases the elements end up not visible? Why are those separate?

jonpas: https://github.com/acemod/ACE3/pull/3805/files#diff-fc2d593f082cd4c729bb26d91f30ede9R31

The only thing is that no other condition gets evaluated if that one is false. I could put it into conditions class and filter it out in `configProperties` as well if that would make it nicer.

Glowbal: Does that mean that if one of the other conditions is false, the rest of those are still evaluated?

jonpas: No, if one is false it will set the element to hidden, if they are all true it will show the element. The difference between the main condition and the conditions class is that when the control is shown by the engine you still need to make sure you hide/show it every time the UI gets updated by the engine. Meanwhile the main condition will exit right away if the control is not getting rendered at all.

Glowbal: So the `condition` defines when an element is available and the `conditions` if it is shown/hidden?

addons/ui/ACE_UI.hpp
```diff
@@ -0,0 +1,177 @@
+// Location (0-both, 1-ground, 2-vehicle)
```
Glowbal: A better solution would be to use macro defines instead.

PR #3806, opened 2016-05-20
UI - Delay a frame before adjusting advanced

PabstMirror:
- Delay a frame before adjusting advanced UI elements (eg stance)


jonpas:
Fixed in #3805 by adding `infoDisplayChanged` event to `RscStanceInfo`.

PR #3811, opened 2016-05-21
Replace execRemoteFnc with events

### LINKED ISSUE #3763: TODO Replace execRemoteFnc with events
`ace_common_fnc_execRemoteFnc` is deprecated and will be removed in 3.7.0. Should get rid of all occurrences in ACE3's code for 3.6.0.

jonpas:
- Replaces `ace_common_fnc_execRemoteFnc` with events
- Replaces `ace_common_fnc_remoteFnc` with events
- Closes #3763 

Tagging @commy2 for `common` and `respawn`, ~~@GitHawk for `rearm`,~~ @SilentSpike for `spectator`, @bux578 for `switchunits`, @Glowbal for `repair`, @esteldunedain for `explosives`.


GitHawk:
Already that in my own PR for rearm

jonpas:
I forgot about that. However, does this implementation without the turret owner ID work? As it is cleaner, but I haven't tested it.

GitHawk:
Not if the turret is manned. Test it and compare it please

jonpas:
Instead of checking by ID I check to which machine the turret is local and use objectEvent directly with that, it should work. I will test later.

GitHawk:
TurretOwner behaves strange. Check the Wiki. Also the turret is no object and may have a different owner from the vehicle object, which is usually owned by the driver or the server.

jonpas:
> TurretOwner behaves strange.

Aye, I will revert the changes to that in this PR, we'll use your PR for changing those to events. Rather be safe than sorry.

> Also the turret is no object and may have a different owner from the vehicle object, which is usually owned by the driver or the server. 

`turretUnit` returns the object which is controlling the turret, but I assume it can be problematic when there is no unit for example.

SilentSpike:
Spectator looks fine :+1: Though, what's the difference between and `objectEvent` and a `targetEvent`?

jonpas:
`objectEvent` takes care of locality automatically, it's just a wrapper for `targetEvent`. If the target object is local it will directly execute locally.

PR #3812, opened 2016-05-22
Use new CBA delayed execution functions and deprecate ACE functions

jonpas:
- Title


esteldunedain:
Looks perfect @jonpas. I'm glad you removed the ACE loop; without thinking about it very hard, I was afraid we'd have to keep it until we removed the deprecated functions.

jonpas:
To be fair, I almost forgot I can do that and just use ACE's functions as wrappers...

PR #3828, opened 2016-05-27
Don't show mine detector in holster

Grey-Soldierman:
Removed Valon from any vest properly configured to use a holster


PabstMirror:
Without this PR, it looks like this in a holster:
image

commy2:
Wait, I'm confused. This doesn't add a mine detector, but sets the holser scale to 0, so the mine detector isn't rendered at the holster.

Glowbal:
This has some issues that I will fix in the mine detector PR

Glowbal:
Thanks @Grey-Soldierman

PR #3840, opened 2016-05-30
Convert to LF line endings

jonpas:
- Changes default EOL to LF for EditorConfig and Git setting (in case where `core.autocrlf` is not enabled)
- Normalizes files that weren't in LF already (most were due to `core.autocrlf` being set on msysgit by default)

After checking this out (and when merged into master) people might see full files being changed, that's because in their working directories line endings have not changed. Usually, committing those files just makes Git realize it's fine and just does it behind the hood without an actual commit. Anyone who wants to get rid of that annoyance can simply run the following in bash:
```
git ls-files -z | xargs -0 rm
git checkout .
```


Glowbal:
Why do we need this?

jonpas:
Excerpt from Slack:
```
[19:02] koffeinflummi basically, having a mix of LF and CRLF fucks up all kinds of things such as diffing files etc.
[19:03] now, we could go all CRLF but that would fuck over linux
[19:03] and after all, that's why windows-git has a checkout crlf setting
[19:04] glowbal Is this going to affect outstanding pull requests / branches?
[19:04] jonpas 1% maybe
[19:05] not many files were changed actually, majority was already LF
[19:05] the change to LF itself won't impact anything
[19:05] koffeinflummi if this causes merge conflicts i'll take care of it
[19:06] just a matter of checking out the changed branch and running dos2unix there as well
[19:06] jonpas majority of changes are from extensions as well
[19:06] glowbal Ok :+1:
```

PR #3844, opened 2016-05-31
Fix showHud if set before display loaded

PabstMirror:
- Fix `showHud` command is ignored if set before display loaded

@jonpas  - This was causing `ace_ui_groupBar` to have no effect in multiplayer.


jonpas:
:+1:

PR #3847, opened 2016-05-31
Add climb height setting

### LINKED ISSUE #3845: set climbing height
Hello all !
this is not an "issue", its just an idea i got, i'm really sorry if i'm somehow violating a rule of this forum but i really didnt found any other way to open a topic here about this!

playing some CQB me and a friend noticed that while "climbing" feature is awesome, its also pretty useless, becasue often you cannot "climb" on low objects, on some walls, and in general you cannot "climb" when you really need to.
so i put hand to code, menaged just to disable the "can_climb" check, so it always climb, but with unrealistic results (like "climbing" in (not "on" i mean "IN" solid walls ecc...). i was not able to adjust height.

so i'm here to ask if there's a way to adjust climbing height (for example i wish set it around 3 meters).
thanks for your help guys !

jonpas:
- Close #3845 
- Add setting to change climb height (increase or decrease)
- Remove unused `CfgInventoryGlobalVariable`
- Make debug lines persist longer (5s)

### DIFF
--- a/addons/movement/ACE_Settings.hpp
+++ b/addons/movement/ACE_Settings.hpp
@@ -5,4 +5,10 @@ class ACE_Settings {
         isClientSettable = 1;
         displayName = CSTRING(UseImperial);
     };
+    class GVAR(climbHeight) {
+        value = 0;
+        typeName = "SCALAR";
+        displayName = CSTRING(ClimbHeight);
+        description = CSTRING(ClimbHeightDesc)
+    };
 };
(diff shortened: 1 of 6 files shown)


Glowbal:
I don't think this adds enough value to justify another setting.

No from me for this.

Khaosmatic:
Will climbing actually have a use now? In the current and previous versions you could barely climb anything. :(

jonpas:
This PR does not change how climbing works, it simply provides a setting for mission makers and server admins to increase or decrease height. However, it is most likely out of scope for ACE3.

commy2:
I can't bring myself to like this. I think this setting is bloat.
Climbing in ACE is meant to get you over small walls. Anything above ~1m should still remain impassable.
If you don't like it, get a different mod for climbing.

commy2:
I like the cleanup and debug mode stuff though.

jonpas:
Closing as wontfix as per discussion. Cleanup and debug improvements in #3876.

PR #3849, opened 2016-05-31
Fix 1.60 UBC (slingloads)

PabstMirror:
- Fix 1.60 UBC (slingloads)

### DIFF
--- a/addons/cargo/CfgVehicles.hpp
+++ b/addons/cargo/CfgVehicles.hpp
@@ -314,15 +314,16 @@ class CfgVehicles {
         GVAR(space) = 20;
         GVAR(hasCargo) = 1;
     };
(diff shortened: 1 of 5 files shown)


commy2:
This also improves the positioning of `GVAR(hooks)[]` from the refuel module, right?

PabstMirror:
I noticed that too, but it's just git diff's being weird:

Compare 
current: https://github.com/acemod/ACE3/blob/master/addons/refuel/CfgVehicles.hpp
new: https://github.com/acemod/ACE3/blob/598c3dc1812e54a36ccb2043705bab7200f99917/addons/refuel/CfgVehicles.hpp

jonpas:
It's because lines were swapped out and it doesn't see it as connected.

commy2:
roger

PR #3850, opened 2016-05-31
Fix CfgPatches author/ is not a value

### LINKED ISSUE #3839: 'Author/' is not a value
**Arma 3 Version:** `1.60` stable 
**CBA Version:** `2.3.1` stable
**ACE3 Version:** `3.5.1` (stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
When placing units or modules created/modified by ACE in a mission and starting it (MP or SP), an error message appears "'author/' is not a value".

**Steps to reproduce:**
1) Place unit or module created/modified by ACE into the editor
2) Start the mission in multiplayer or singleplayer
3) Error message appears

- Self-Hosted Multiplayer / Editor (Singleplayer) / Editor (Multiplayer)

**Placed Modules:**
- Pretty much all modules

**RPT log file:**
This is the relevant line: 
17:29:25 Warning Message: 'author/' is not a value

jonpas:
- Fix #3839 
- Convert all `author[]` to `authors[]` in all CfgPatches - [CBA is changing that](https://github.com/CBATeam/CBA_A3/pull/345) for compatibility with vanilla `author`
- Remove `"ACE-Team"` (strinbtabled) from `authors[]` in all CfgPatches - CBA will pull both `author` and `authors[]` together
- Add `author` with value `"ACE-Team"` (stringtabled) to all CfgPatches
- Add `url` with value `"http://ace3mod.com"`` to all CfgPatches
- Update `blank` component as well

#### Regex used
**Converting `author[]` to `authors[]`:**
Find:
```
author\[\] = {([\s\S]*?)};
```
Replace:
```
authors\[\] = {$1};
```

**Adding `author` and `url`:**
Find:
```
class CfgPatches([\s\S]*?);([\s]*?)authors\[\]
```
Replace:
```
class CfgPatches$1;\n        author = ECSTRING\(common,ACETeam\);$2authors\[\]
class CfgPatches$1;\n        url = "http://ace3mod.com";$2authors\[\]
```

### DIFF
--- a/addons/advanced_ballistics/config.cpp
+++ b/addons/advanced_ballistics/config.cpp
@@ -6,8 +6,9 @@ class CfgPatches {
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {"ace_ballistics", "ace_weather"};
-        author[] = {"Ruthberg"};
-        authorUrl = "https://github.com/ulteq";
+        author = ECSTRING(common,ACETeam);
+        authors[] = {"Ruthberg"};
+        url = "http://ace3mod.com";
         VERSION_CONFIG;
     };
 };
--- a/addons/advanced_ballistics/config.cpp
+++ b/addons/advanced_ballistics/config.cpp
@@ -8,7 +8,7 @@ class CfgPatches {
         requiredAddons[] = {"ace_ballistics", "ace_weather"};
         author = ECSTRING(common,ACETeam);
         authors[] = {"Ruthberg"};
-        PBO_URL
+        PBO_URL;
         VERSION_CONFIG;
     };
 };
--- a/addons/advanced_ballistics/config.cpp
+++ b/addons/advanced_ballistics/config.cpp
@@ -8,7 +8,7 @@ class CfgPatches {
         requiredAddons[] = {"ace_ballistics", "ace_weather"};
         author = ECSTRING(common,ACETeam);
         authors[] = {"Ruthberg"};
-        url = "http://ace3mod.com";
+        PBO_URL
         VERSION_CONFIG;
     };
 };
--- a/addons/main/script_mod.hpp
+++ b/addons/main/script_mod.hpp
@@ -11,6 +11,9 @@
 #define VERSION MAJOR.MINOR.PATCHLVL.BUILD
 #define VERSION_AR MAJOR,MINOR,PATCHLVL,BUILD
 
+#define WEB_URL "http://ace3mod.com/"
+#define PBO_URL url = WEB_URL;
+
 // MINIMAL required version for the Mod. Components can specify others..
 #define REQUIRED_VERSION 1.56
 #define REQUIRED_CBA_VERSION {2,3,1}
--- a/addons/main/script_mod.hpp
+++ b/addons/main/script_mod.hpp
@@ -12,7 +12,7 @@
 #define VERSION_AR MAJOR,MINOR,PATCHLVL,BUILD
 
 #define WEB_URL "http://ace3mod.com/"
-#define PBO_URL url = WEB_URL;
+#define PBO_URL url = WEB_URL
 
 // MINIMAL required version for the Mod. Components can specify others..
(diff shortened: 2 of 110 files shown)


SzwedzikPL:
You can also add `url = "http://ace3mod.com";` to have proper link in eden addons list

bux578:
Good job.

Though, the order of those entries is triggering me :)

Might be better:
```
author
authors[]
authorUrl
url
```

But it'd be just a cosmetic change.

And if no one is objecting, we could change authorUrl to http://ace3mod.com aswell.
multiple authors[] but only one authorUrl

addons/medical_menu/config.cpp
```diff
@@ -6,7 +6,9 @@ class CfgPatches {
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {"ace_medical"};
-        author[] = {$STR_ACE_Common_ACETeam, "Glowbal"};
+        author = ECSTRING(common,ACETeam);
+        url = "http://ace3mod.com";
+        authors[] = {$STR_ACE_Common_ACETeam, "Glowbal"};
```
commy2: remove it here or it will be shown twice

jonpas: Missed that one, nice spot.

jonpas:
@bux578 I'd say we get rid of authorUrl and just keep url, @commy2 just pushed a change so url only will be shown if there are both (right?).

commy2:
`url` will be shown in 2.4.1 CBA. It is also used in BIs shenanigans, so that's the way to go.

For bwc, CBA will fall back to `authorURL`, if `url` isn't there. Since that is not supported by BI, there is no reason to keep it.

I'd say replace all URLs with `http://ace3mod.com` instead of the github pages of the contributors. Possibly consider a macro (or add it to the `VERSION_CONFIG` macro)

jonpas:
I totally forgot about that macro! Will put `author` and `url` there, remove `authorUrl` and leave `authors[]` as is.

jonpas:
Done, also fuck line endings breaking my regex.

bux578:
How about making the URL into a stringtable entry or a macro?

jonpas:
`#define URL "http://ace3mod.com"` or full `#define URL url = "http://ace3mod.com";`?

bux578:
I'd say first. So we could use it somewhere else.

Glowbal:
I'd go for both.

```
#define WEB_URL "http://ace3mod.com/"
#define PBO_URL url = WEB_URL;
```

bux578:
good idea

addons/main/script_mod.hpp
```diff
@@ -1,16 +1,19 @@
...
+#define MAJOR 3
+#define MINOR 5
+#define PATCHLVL 1
+#define BUILD 0
+
+#define VERSION MAJOR.MINOR.PATCHLVL.BUILD
+#define VERSION_AR MAJOR,MINOR,PATCHLVL,BUILD
+
+#define WEB_URL "http://ace3mod.com/"
+#define PBO_URL url = WEB_URL;
```
commy2: I think the standard is to not have the ";" in the macro. Sorry for being picky.

addons/advanced_ballistics/config.cpp
```diff
@@ -6,8 +6,9 @@ class CfgPatches {
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {"ace_ballistics", "ace_weather"};
-        author[] = {"Ruthberg"};
-        authorUrl = "https://github.com/ulteq";
+        author = ECSTRING(common,ACETeam);
+        authors[] = {"Ruthberg"};
+        PBO_URL
```
commy2: see ^
would be `PBO_URL;` just like `VERSION_CONFIG;`

Glowbal: Yes, you are quite right. My bad for that one.

commy2:
Thanks!

PR #3851, opened 2016-05-31
Fixed HuntIR deployment

BaerMitUmlaut:
- Create only one HuntIR when fired (previously, there was one created for every connected player)


esteldunedain:
This is absolutely right.

The comment on the header of `fnc_handleFired.sqf`should be changed to say 

`* Handles HuntIR projectiles. Called from the unified fired EH for the local player.`

esteldunedain:
Thanks @BaerMitUmlaut :+1:

PR #3858, opened 2016-06-02
Cleanup transistion to CBA's waitAndExec

PabstMirror:
- Minor Cleanup from transistion to CBA's waitAndExec


Glowbal:
:+1:

PR #3860, opened 2016-06-02
Integrate 1.60 marker shape color map interface

PabstMirror:
1.60 adds ability to set color and marker type from the map interface (top right corner)
Luckily they use the same values and indexes as ACE :wink: 

This PR connects the two systems:
Selecting a color from the 1.60 map display will set the color in the
ace marker display and vice versa

### DIFF
--- a/addons/markers/CfgEventHandlers.hpp
+++ b/addons/markers/CfgEventHandlers.hpp
@@ -16,3 +16,9 @@ class Extended_PostInit_EventHandlers {
         init = QUOTE(call COMPILE_FILE(XEH_postInit));
     };
 };
+
+class Extended_DisplayLoad_EventHandlers {
+    class RscDiary {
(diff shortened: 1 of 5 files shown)


commy2:
Nice

jonpas:
Awesome! 

Is config displayLoad preferred over scripted addDisplayEventHandler?

commy2:
`addDisplayEventHandler` only works for the missionDisplay (46). This is the diary display.

jonpas:
Ah I see.

esteldunedain:
Is it me or BIS totally ripped out our idea?

jonpas:
They totally did.

bux578:
That's a good thing, imo :)

commy2:
Not the first time.
I like it.

jonpas:
:+1:

PR #3862, opened 2016-06-02
Spottingscope fixes

BaerMitUmlaut:
- Fix the interaction point of the spotting scope
- Correct the classname so it shows up in the miscellaneous items box

Here's the interaction point before and after:
comparison


esteldunedain:
:+1:

PR #3868, opened 2016-06-03
Realistic weights

BaerMitUmlaut:
- Tweak all weapon weights so the value shown in the inventory is correct.

The purpose of this is that all weapons have a comparable weight value, which is useful both for vanilla stamina and advanced fatigue.

Could potentially break loadouts that put weapons in backpacks, but since that is such a rarity and the majority of weights has been lowered I think this is totally fine.
The ASP-1 KIR and the Cyrus have not been adjusted since they are not based on real weapons and their weight value is close to what I would estimate.

### DIFF
--- a/addons/realisticweights/$PBOPREFIX$
+++ b/addons/realisticweights/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\ace\addons\realisticweights
--- a/addons/realisticweights/CfgWeapons.hpp
+++ b/addons/realisticweights/CfgWeapons.hpp
@@ -1,0 +1,275 @@
+class CfgWeapons {
+    class Rifle_Base_F;
+    class Rifle_Long_Base_F: Rifle_Base_F {
+        class WeaponSlotsInfo;
+    };
+    class Launcher;
+    class Launcher_Base_F: Launcher {
+        class WeaponSlotsInfo;
+    };
+    class Pistol;
+    class Pistol_Base_F: Pistol {
+        class WeaponSlotsInfo;
+    };
+
+
+// - Assault rifles -----------------------------------------------------------
+
+    // - MX - weights taken from the ACR --------------------------------------
+    class arifle_MX_Base_F: Rifle_Base_F {
+        class WeaponSlotsInfo;
+    };
+    class arifle_MX_F: arifle_MX_Base_F {
+        class WeaponSlotsInfo: WeaponSlotsInfo {
+            mass = 79;
+        };
+    };
+    class arifle_MXC_F: arifle_MX_Base_F {
+        class WeaponSlotsInfo: WeaponSlotsInfo {
+            mass = 72;
+        };
+    };
+    // 3GL weight with stock 5lb, estimated without stock 4lb
+    class arifle_MX_GL_F: arifle_MX_Base_F {
+        class WeaponSlotsInfo: WeaponSlotsInfo {
+            mass = 119;
+        };
+    };
+    // Estimated difference to MX by comparing M27 to M416 (~100g)
+    class arifle_MX_SW_F: arifle_MX_Base_F {
+        class WeaponSlotsInfo: WeaponSlotsInfo {
+            mass = 81;
+        };
+    };
+    // Estimated difference to MX by comparing M416 D16.5RS to M416 D20RS
+    class arifle_MXM_F: arifle_MX_Base_F {
+        class WeaponSlotsInfo: WeaponSlotsInfo {
+            mass = 85;
+        };
+    };
+
+    // - KH2002 ---------------------------------------------------------------
+    class arifle_Katiba_Base_F: Rifle_Base_F {
+        class WeaponSlotsInfo;
+    };
+    class arifle_Katiba_F: arifle_Katiba_Base_F {
+        class WeaponSlotsInfo: WeaponSlotsInfo {
+            mass = 81;
+        };
+    };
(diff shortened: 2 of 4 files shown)


voiperr:
The Sig's weight is off. 8.2 lbs empty should be 82, not 155.
Otherwise :ok_hand: :+1:

esteldunedain:
Do any of you guys know if it's possible to make preprocessor macros that do math?

jonpas:
> Do any of you guys know if it's possible to make preprocessor macros that do math?

It is with [`__EVAL`](https://community.bistudio.com/wiki/PreProcessor_Commands#EVAL).

PabstMirror:
```
#define KG_TO_MASS(massIngKilos) __EVAL (massIngKilos * 25.4)
mass = KG_TO_MASS(10);
```

esteldunedain:
@BaerMitUmlaut, would you mind using @PabstMirror 's awesome macro? BTW, it's CBA's material

BaerMitUmlaut:
Didn't the `__EVAL` only work with non-binarized PBOs? I'd like to avoid that if possible.

esteldunedain:
I'd assume that during binarization the expresion is evaluated and replaced by the final number, which would be fine I think.

commy2:
Mikeros tools cannot evalute SQF and the current version errors with any kind of eval.
also, don't use strings for this math. The arsenal cannot read these correctly. I tried that before with the G36 /w AG40 GL (`mass = MASS_G36 + MASS_AG40;`) and it reported strange values ingame.

BaerMitUmlaut:
@voiperr: I used the weight of the SIG 550 Sniper as a reference, seemed more appropriate.

voiperr:
Fair enough, but I contend that the amount of furniture is different than the 550 SR. It's not convincing that the 550/556 depicted in-game is heavier than the Noreen.

PabstMirror:
Unbinerized and mikro dll 5.24 handle the `__EVAL` fine 
Both are `isNumber` = true in game.

There is no sqf, just math commands.  But I'm not sure about newest mikro tools.

Glowbal:
We discussed this and feel there are potential issues and some concerns with the changes in this PR:

- Changing mass will break old load out templates.
- We have always tried to stick to vanilla standards. This would deviate from that
- An outstanding question is "what makes this realistic?"

BaerMitUmlaut:
Changing mass will not break old loadout templates unless you put a *weapon* that now weighs more in the units uniform/vest/backpack, as this PR does not change magazine weights. I've compiled a list of the mass changes below, weapons that are now heavier are marked bold:

| Weapon                  | Before | After |
| ----------------------- | ------:| -----:|
| arifle_MX_F             |    100 |    79 |
| arifle_MXC_F            |     80 |    72 |
| arifle_MX_GL_F          |    120 |   119 |
| arifle_MX_SW_F          |    120 |    81 |
| arifle_MXM_F            |    120 |    85 |
| arifle_Katiba_F         |    100 |    81 |
| arifle_Katiba_C_F       |     80 |    73 |
| arifle_Katiba_GL_F      |    120 |   111 |
| arifle_Mk20_F           |     80 |    75 |
| **arifle_Mk20C_F**      |     70 |    74 |
| **arifle_Mk20_GL_F**    |    100 |   105 |
| **arifle_TRG21_F**      |     70 |    72 |
| **arifle_TRG20_F**      |     60 |    70 |
| **arifle_TRG21_GL_F**   |     90 |   102 |
| LMG_Mk200_F             |    220 |   100 |
| LMG_Zafir_F             |    320 |   163 |
| srifle_DMR_01_F         |    100 |   100 |
| srifle_EBR_F            |    180 |   112 |
| srifle_GM6_F            |    480 |   253 |
| srifle_LRR_F            |    420 |   310 |
| **hgun_PDW2000_F**      |     30 |    33 |
| **SMG_01_F**            |     40 |    60 |
| **SMG_02_F**            |     50 |    61 |
| **launch_NLAW_F**       |    120 |   275 |
| launch_RPG32_F          |    100 |    66 |
| launch_Titan_base       |    160 |   110 |
| launch_Titan_short_base |    140 |    88 |
| **hgun_ACPC2_F**        |     20 |    24 |
| hgun_P07_F              |     20 |    13 |
| hgun_Pistol_heavy_01_F  |     30 |    21 |
| hgun_Pistol_heavy_02_F  |     30 |    20 |
| hgun_Rook40_F           |     20 |    20 |
| srifle_DMR_02_F         |    320 |   130 |
| srifle_DMR_03_F         |    220 |    90 |
| srifle_DMR_06_camo_F    |    140 |    92 |
| MMG_01_base_F           |    380 |   218 |
| MMG_02_base_F           |    340 |   240 |

This does also not deviate from vanilla standards - we're still using their mass system, we're just adjusting the weight comparatively. With one exception (the NLAW), the weight increase is what I would consider minor.

The realism part comes from exactly that - we're changing the mass so they fit to each other. For example, let's take the Stoner LMG (aka. Mk200): The original mass is 220, more than twice as much as the VS-121/Rahim (original and modified mass 100), while they should weigh about the same.
Or an even more extreme example, the Noreen ULR/MAR-10 has a mass of 320, three times as much as the F2000/Mk20 - it shouldn't even have twice as much.

Issues like these will make creating sensible loadouts way harder once we implement a new fatigue system: You're weighing more than you should, which will cause you to become fatigued unreasonably fast.

Glowbal:
lgtm

PR #3873, opened 2016-06-04
Fix NVG overbright when player throws grenades

voiperr:
- Title

### DIFF
--- a/addons/nightvision/functions/fnc_blending.sqf
+++ b/addons/nightvision/functions/fnc_blending.sqf
@@ -21,6 +21,9 @@ TRACE_10("firedEH:",_unit, _weapon, _muzzle, _mode, _ammo, _magazine, _projectil
 private "_player";
 _player = ACE_player;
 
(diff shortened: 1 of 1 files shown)


esteldunedain:
Looks good.

addons/nightvision/functions/fnc_blending.sqf
```diff
@@ -21,8 +21,8 @@ TRACE_10("firedEH:",_unit, _weapon, _muzzle, _mode, _ammo, _magazine, _projectil
 private "_player";
 _player = ACE_player;
 
-//If our vehicle didn't shoot, or we're not in NVG mode, exit
-if ((currentVisionMode _player) != 1) exitWith {};
+//If we're not in NVG mode, or it's a grenade, exit
+if (((currentVisionMode _player) != 1) || {_weapon == "Throw"}) exitWith {};
```
jonpas: Just fyi, `()` are not necessary here. `||`/`&&` will always be checked last.

PR #3883, opened 2016-06-06
Move Defuse action to main action

jonpas:
- Title
- There are no other actions apart from "Defuse" once the mine is set and it's quite stupid that you need to expand the menu
- Remove redundant entries
- Stringtable author


esteldunedain:
Yup, thank you

PR #3889, opened 2016-06-09
Convert to LF line endings gh-pages

jonpas:
- Same as #3840 but for `gh-pages`


bux578:
https://github.com/acemod/ACE3/blob/gh-pages/src/.editorconfig

jonpas:
Why exactly is that file in there in the first place? It has same settings as the one in root.

jonpas:
@bux578 ^ ?

bux578:
because reasons?

jonpas:
So it can be removed? :o

PR #3900, opened 2016-06-12
Optics - Create PIP camera only when needed

### LINKED ISSUE #3899: Ace optics causes textures flickering

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Ace optics causes textures flickering after 1.60 arma patch. My friends helped me to test this and 6 out of 7 guys had this problem. One guy didn't have the problem actually, we don't know why. He was on an AMD card, but one of the other 6 guys had AMD too and had this bug, so I'm not totally sure if that's related.
Video.
https://www.youtube.com/watch?v=8s-JsJ3N0j8&feature=youtu.be
First part of the video is with optics pbo enabled.
Second part of the video (from 0:24) is with optics pbo disabled.

**Steps to reproduce:**
- Get in vehicle.
- Turn on PIP.
- Take a look around.

**Placed Modules:**
- None.

**RPT log file:**
With optics.
https://gist.github.com/rebelvg/7be66516e8bb727b4aa8a69bf542b23e
Without optics.
https://gist.github.com/rebelvg/50c19e16440796b8c98e1191c6f9aa62
Mission file used for video.
https://gist.github.com/rebelvg/d1c064a848968ad1de95c439b6677c64

PabstMirror:
- Fix #3899

Instead of creating the PIP camera at mission start, we only create it when actually needed when in scoped view, and destroy it after leaving.

Should clear up problems we've seen with mission's using camera scripts due to us always calling `cameraEffect`.

### DIFF
--- a/addons/optics/XEH_postInit.sqf
+++ b/addons/optics/XEH_postInit.sqf
@@ -5,30 +5,9 @@ if (!hasInterface) exitWith {};
 
 GVAR(camera) = objNull;
 
-0 = 0 spawn {
-    // Workarround for the ThreeDen Editor Camera
-    if ((!isNil {is3DEN}) && {is3DEN}) exitWith {
-        ACE_LOGINFO("Eden detected: disabling Optics PIP Camera");
-    };
-
-    waitUntil {!isNull ACE_player};
-    waitUntil {sleep 1; {_x != GVAR(camera)} count allMissionObjects "camera" == 0 && {isNull curatorCamera}};
-
-    GVAR(camera) cameraEffect ["TERMINATE", "BACK"];
-    camDestroy GVAR(camera);
-
-    // PiP technique by BadBenson
-    GVAR(camera) = "camera" camCreate positionCameraToWorld [0,0,0];
-    GVAR(camera) camSetFov 0.7;
-    GVAR(camera) camSetTarget ACE_player;
-    GVAR(camera) camCommit 1;
-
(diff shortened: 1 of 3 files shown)


commy2:
I don't like this solution. It makes the PIP optics look like this for up to a few seconds (depends on FPS) when switching to them:


I'd much rather have BI fix this bug, but I guess that will not happen. This year or ever.

PabstMirror:
@commy2 - Could you give it another try, I changed it to prepare the pip camera just when you have the optic equipped and not only when zoomed in on the scope.

commy2:
There still is the same effect when switching to it for the first time, but at least every other time it looks right.
That means that you'll get the flickering when using these optics though, right?

PabstMirror:
When scoping in for the first time I do get a little black flicker, but it seems to be the same without this PR on master. For me it's only a frame or 2.

For the texture flickering that this PR should fix, I'm only really seeing it when sitting in the Hunter/M-ATV which has a PIP rear camera (check the demo mission in #3899). So it might only show up when using another PIP system?

commy2:
I don't think "Ace optics causes textures flickering" is correct. It's more like a/(anothr) render problem with PiP in general.

Only a minority is using the PiP optics while everyone can be sitting in vehicles with rear view cameras or side view mirrors. The flickering is not acceptable, so merge.
:+1:

PR #3902, opened 2016-06-12
Tanoa Weather Data

bux578:
- Add Tanoa weather data based on the close by Fiji city Suva


commy2:
:+1:

bux578:
It should even be good to go for 3.6.0, it's not messing with anything as far as i've tested

PR #3909, opened 2016-06-12
Adjust bandage configuration

### LINKED ISSUE #3732: Wounds reopening too fast?
While looking at the ACE_medical_treatments.hpp, i noticed that some wounds reopen within 10 seconds on line 749 for example. I'm sure that people would be more skilled in applying bandages that will stay on for at least 10 minutes. Is this a overlook or an actual feature?

Glowbal:
Changes the bandage configuration to match the following:

- Field dressing has average treatment
- Packing bandage has average treatment, higher reopen change, longer reopening delay
- Elastic bandage has higher treatment, higher reopen change, shorter reopen delay
- QuikClot has lower treatment, lower reopen change, longer reopening delay

Fixes #3732

### DIFF
--- a/addons/medical/ACE_Medical_Treatments.hpp
+++ b/addons/medical/ACE_Medical_Treatments.hpp
@@ -438,7 +438,6 @@ class ACE_Medical_Advanced {
                     bleedingRate = 0.0002;
                 };
             };
-
             // Occur when an entire structure or part of it is forcibly pulled away, such as the loss of a permanent tooth or an ear lobe. Explosions, gunshots, and animal bites may cause avulsions.
             class Avulsions {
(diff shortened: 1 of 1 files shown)


kripto202:
is there a way, in the treatment process, that the packing bandage stays on longer, for example, applying an elastic bandage?

Meiestrix:
Hi i just looked at this changes and made a table for all the efficiencies for each wound type. Now I noticed that the ElasticBandage is the ultimate bandage for everything when you do not play with wound reopening. It has the highest efficiency for each wound type in all three wound types.
This makes the other bandages useless when you have reopening disabled.

Meiestrix:
image

image

PR #3911, opened 2016-06-12
Gunbag cleanup

jonpas:
- Fix indentation
- Prefix model
- Shorten classnames
- Remove redundant `CLASSNAME` macro (which is the same as `ADDON`.
- Fix unquoted things
- Fix function headers

~~There is a lot more improvement to this IMO, like usage of new commands (`apply` primarily) and~~ Calls from interaction menu could be nicer, but that can be changed later without impact.


addons/gunbag/functions/fnc_toGunbagCallback.sqf
```diff
@@ -40,6 +43,6 @@ _unit removeWeapon _weapon;
 _gunbag setVariable [QGVAR(gunbagWeapon), [_weapon, _items, _magazines], true];
 
 // play sound
-if (!isNil "ACE_Backpacks") then {
+if (["ace_backpacks"] call EFUNC(common,isModLoaded)) then {
```
commy2: What's the point?

commy2:
> like usage of new commands
May I ask where?

addons/gunbag/functions/fnc_toGunbagCallback.sqf
jonpas: It checks actual CfgPatches rather than just global variable. We also use that everywhere. If it's not needed, why do we have that function in the first place?

jonpas:
> May I ask where?

Scratch that, I don't know what I was looking at, also at 5am...

addons/gunbag/functions/fnc_toGunbagCallback.sqf
esteldunedain: >  If it's not needed, why do we have that function in the first place?

Actually it comes from CSE. I believe CBA creates the `ACE_Backpacks` variable to signal that a CfgPatch with that name was loaded, so both methods serve the same purpose.

jonpas: That happens in preInit, `ADDON = false; ADDON = true;` What should we go with though. I'd say deprecate isModLoaded and mass replace it.

esteldunedain: Oh, if it happens on preInit instead of automatically we can't remove `isModLoaded`. Some modules don't have preInit on the first place.

jonpas: True. I also feel safer using isModLoaded as it checks CfgPatches directly rather than a variable which could be changed on the fly.

PR #3916, opened 2016-06-13
Merge basic and advanced injury systems

Glowbal:
#3134

This PR changes the basic injury system to make use of advanced medical. The benefits of this are that advanced medical is much more powerful in its configuration and it removes duplicate code, making it easier to maintain.

This means bugs with the injury system that occure in basic should also occure in advanced, as the only difference is the visual display and the bandage treatment types.

- The visual display is exactly the same as before this PR.
- Bandaging has some minor changes, since it now works with injuries instead of damages. It has gained its own bandage configuration item in the bandage config. This has increased effectiveness and no reopening chance.

Steps to do after this PR:

- Merge the basic and advanced settings
- Make vitals its own setting
- Introduce a queue for injury broadcasts. The aim is to stack and send out an aggegrated collection of changes rather than broadcasting everytime an injury gets updated. Local client should always have the latest values, other clients will get them at a slightly later time. Should not be noticable by clients.
- Switch out the Revive system. @BaerMitUmlaut 

Things that are not done with this PR and will be done after all else is completed:

- Further clean up and modernize the code. Certain parts could make use of new script commands. Switching over now may only provide us with obstacles while we are merging the two systems. We will do this after all else is completed. Functionality first.
- Certain treatment options could perhaps also be merged, such as IV and medications. Further investigation is required.

---

I have marked this for 3.6.0 but I realize that might be slightly unrealistic. We will see how well this merge goes and if need be, push it to 3.7.0.

This PR needs heavy testing, preferably in large scale MP environments. It is a significant change that I believe will work out for the better, but should not be merged without a proper playthrough first.


bux578:
That's awesome!!! gj @Glowbal 

> This PR needs heavy testing, preferably in large scale MP environments. It is a significant change that I believe will work out for the better, but should not be merged without a proper playthrough first.

3.7.0

Glowbal:
This PR contains some minor changes of #3915

DLEGION:
hello guys !
the plan seems great, i just wish to add a "visual" touch....in normal arma3 injured soldiers...bleed on the ground. i mean leave blood stains.
in ace3 seems there is no bleeding on the groun at all. blood stains on soldier clothes are ok, but on the ground are also useful to follow tracks ecc... and a soldier dead in a blood pool is way more realistic ! thanks !

Glowbal:
Sorry @DLEGION, this pull request is not the place for feature requests.

Glowbal:
This is largely the same as on the new rewrite. It works fine from my testing - merging this, so we can get some more testing in.

PR #3917, opened 2016-06-13
Split medical into multiple components

Glowbal:
Medical is one of our largest component. This PR splits it into seperate components with:

- Medical being responsible for common functionality and damage handling
- medical_treatment being responsbile for offering treatment functionality
- medical_menu staying as is, offering the optional medical menu

Contains some changes from other PRs to avoid merge conflicts down the line.

The goal of this PR is to decrease the amount of functionality within one component and make it easier to work with the medical components.

- This still has all settings and global variables shared with the medical component. The medical_treatment component offers only functionality ontop.
- Still need to provide deprecation wrappers for deleted functions that where marked as public in medical.


Glowbal:
Closing in favour of #4064

PR #3920, opened 2016-06-13
Fix basic UI being overwritten on RscDiary unload

### LINKED ISSUE #3907: UI / group bar re-appears on diary open
**ACE3 Version:** `3.5.1` (1136a57e4288f03bcc3f1cd5770982765acc1f2b)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Group menu from the left bottom corner re-appears when diary is opened and closed (default J key). Can be hidden by showHUD command again. If shouwHUD array exists in description ext, the showHUD command gets completely blocked (stops doing anything). Group menu can be hidden again on latest master when you switch to main weapon from any other weapon (not vise-versa).

**Steps to reproduce:**
- Place a group.
- Play as a leader.
- Open and close diary (J key).

**Placed Modules:**
- None.

**RPT log file:**
- N/A.

jonpas:
- Fix #3907 
- Add disabled message on changing basic config which has been set (forced) by mission config
- Remove redundant parameter from basic `setElements`

Todo:
- [x] Add condition source when false to hint
- [x] Add source to scripted element setting API

### DIFF
--- a/addons/ui/CfgEventHandlers.hpp
+++ b/addons/ui/CfgEventHandlers.hpp
@@ -15,3 +15,10 @@ class Extended_PostInit_EventHandlers {
         clientInit = QUOTE(call COMPILE_FILE(XEH_clientInit));
     };
 };
+
+// Closing diary resets showHUD
+class Extended_DisplayUnload_EventHandlers {
+    class RscDiary {
+        ADDON = QUOTE([{[false] call FUNC(setElements)}] call CBA_fnc_execNextFrame);
+    };
+};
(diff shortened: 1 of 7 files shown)


commy2:
Question. Does this mean that `showHud` in vanilla breaks every time you open the diary?
This would be really stupid and probably is a bug.

jonpas:
Apparently, I haven't tested myself in vanilla though.

commy2:
I guess they use the command in their scripting for the diary?

jonpas:
Could be, can't check right now.

PR #3922, opened 2016-06-14
Map + Flashlights Cleanup

voiperr:
- ace_map now handles flashlights independently of ace_flashlights.
- ace_map sources paths for flashlight beam textures directly from weapon configs (adds flexibility for other mods/beam types).
- ace_map has acc_flashlight config (so vanilla ace_map will have at least one flashlight to use).
- Flashlight and glow proxy vars are now stored globally in the unit.
- Add orange proxy light.
- Give modders options for silent flashlights (minor).


addons/flashlights/CfgWeapons.hpp
```diff
@@ -22,9 +11,11 @@ class CfgWeapons {
         scope = 2;
         class ItemInfo: InventoryItem_Base_F {
             mass = 1;
-            class FlashLight {
-                ACE_Flashlight_Colour = "red";
-                ACE_Flashlight_Size = 1.75;
+            class ACE_Map_Flashlight {
+                ACE_Map_Flashlight_Color = "red";
+                ACE_Map_Flashlight_Beam = QUOTE(PATHTOF(UI\Flashlight_beam_red_ca.paa));
```
jonpas: `QPATHTOF`

addons/flashlights/CfgWeapons.hpp
```diff
@@ -22,9 +11,11 @@ class CfgWeapons {
         scope = 2;
         class ItemInfo: InventoryItem_Base_F {
             mass = 1;
-            class FlashLight {
-                ACE_Flashlight_Colour = "red";
-                ACE_Flashlight_Size = 1.75;
+            class ACE_Map_Flashlight {
```
jonpas: I don't see why renaming this is necessary. The entries are already prefixed anyways.

addons/flashlights/CfgWeapons.hpp
```diff
@@ -52,9 +45,11 @@ class CfgWeapons {
         scope = 2;
         class ItemInfo: InventoryItem_Base_F {
             mass = 1;
-            class FlashLight {
-                ACE_Flashlight_Colour = "white";
-                ACE_Flashlight_Size = 2.15;
+            class ACE_Map_Flashlight {
+                ACE_Map_Flashlight_Color = "white";
+                ACE_Map_Flashlight_Beam = QUOTE(PATHTOF(UI\Flashlight_beam_white_ca.paa));
```
jonpas: `QPATHTOF`

addons/flashlights/CfgWeapons.hpp
```diff
@@ -37,9 +28,11 @@ class CfgWeapons {
         scope = 2;
         class ItemInfo: InventoryItem_Base_F {
             mass = 1;
-            class FlashLight {
-                ACE_Flashlight_Colour = "red";
-                ACE_Flashlight_Size = 1.5;
+            class ACE_Map_Flashlight {
+                ACE_Map_Flashlight_Color = "red";
+                ACE_Map_Flashlight_Beam = QUOTE(PATHTOF(UI\Flashlight_beam_red_ca.paa));
```
jonpas: `QPATHTOF`

addons/map/CfgWeapons.hpp
```diff
@@ -0,0 +1,16 @@
+class CfgWeapons {
+
+    class ItemCore;
+    class InventoryFlashlightItem_Base_F;
+
+    class acc_flashlight: ItemCore {
+        class ItemInfo: InventoryFlashlightItem_Base_F {
+            class ACE_Map_Flashlight {
+                ACE_Map_Flashlight_Color = "white";
+                ACE_Map_Flashlight_Beam = QUOTE(PATHTOF(UI\Flashlight_beam_white_ca.paa));
```
jonpas: `QPATHTOF`

jonpas:
Changes of entries will break backwards compatibility.
`ACE_Flashlight_Colour` -> `ACE_Map_Flashlight_Colour`

addons/flashlights/CfgWeapons.hpp
commy2: Agreed. It's pointless.

voiperr:
Implemented suggestions.

addons/flashlights/CfgWeapons.hpp
```diff
@@ -37,9 +28,11 @@ class CfgWeapons {
         scope = 2;
         class ItemInfo: InventoryItem_Base_F {
             mass = 1;
-            class FlashLight {
+            class Flashlight {
                 ACE_Flashlight_Colour = "red";
+                ACE_Flashlight_Beam = QPATHTOF(UI\Flashlight_beam_red_ca.paa);
                 ACE_Flashlight_Size = 1.5;
+                ACE_Map_Flashlight_Sound = 1;
```
jonpas: Still has old name (the one below jas a different one).

addons/flashlights/CfgWeapons.hpp
```diff
@@ -52,9 +45,11 @@ class CfgWeapons {
         scope = 2;
         class ItemInfo: InventoryItem_Base_F {
             mass = 1;
-            class FlashLight {
+            class Flashlight {
```
jonpas: This is still changed (capital L). In other places as well.

addons/map/functions/fnc_switchFlashlight.sqf
```diff
@@ -16,10 +16,18 @@
...
-GVAR(flashlightInUse) = _flashlight;
 if (GVAR(mapGlow)) then {
-    [GVAR(flashlightInUse)] call FUNC(flashlightGlow);
+    [_newFlashlight] call FUNC(flashlightGlow);
 };
-playSound "ACE_map_flashlightClick";
\ No newline at end of file
+
+if (
+    (getNumber (configFile >> "CfgWeapons" >> _newFlashlight >> "ItemInfo" >> QGVAR(Flashlight) >> QGVAR(Flashlight_Sound)) > 0) ||
```
jonpas: Also using now-inexistant classnames.

addons/map/functions/fnc_switchFlashlight.sqf
```diff
@@ -16,10 +16,18 @@
 
 #include "script_component.hpp"
 
-params ["_flashlight"];
+params ["_newFlashlight"];
+private _oldFlashlight = GVAR(flashlightInUse);
```
jonpas: I wonder if thaz QGVAR should rather be player setVariable?

voiperr: I figured since it would never be globally known, it wouldn't be needed. Would it be worthwhile to make it a PVAR?

esteldunedain: There's three "levels of detail" you can adopt when doing this kind of stuff:

A - Using a GVAR. State is shared between all local units. If you remote control a unit using zeus, team switch, etc the flashlight state carries over. No network traffic at all.

B - Using `_unit setVariable`, but privately. The state is unique to each unit, so it supports team switch, remote control, etc but only for local units. If you e.g. remote control a remote unit, the correct state won't be displayed. No network traffic at all.

C - Using `_unit setVariable` publicly. The state is unique to each unit, even if they are remote, locality changes, etc. Requires network traffic on each change of state.

Currently this uses approach A. I assume @jonpas is refering to B and @voiperr is talking about C. I consider C the ideal for this, but it's hardly worth the effort.

jonpas: Yes, I think B would be the best option, although C wouldn't cause much network traffic either as you'd use that from time to time only.

voiperr:
Taking the previous discussion into account, the flashlight and glow effect are stored in a PVAR on the unit. So, it should work persistently with switched and remote units.

I also took the liberty of cleaning up the rest of the addon (mostly privates). The biggest change is to `determineMapLight`, which runs slightly faster now, and handles any type of chemlight.

Feel free to comment on the cleanup. I don't foresee any more additions to this PR.

addons/map/XEH_postInitClient.sqf
```diff
@@ -97,35 +78,33 @@ call FUNC(determineZoom);
...
     if (GVAR(mapIllumination)) then {
-        GVAR(flashlightInUse) = "";
-        GVAR(glow) = objNull;
-
         ["ace_playerInventoryChanged", {
-            private _flashlights = [ACE_player] call FUNC(getUnitFlashlights);
-            if ((GVAR(flashlightInUse) != "") && !(GVAR(flashlightInUse) in _flashlights)) then {
-                GVAR(flashlightInUse) = "";
+            params ["_player", ""];
+            private _flashlightItems = [_player] call FUNC(getUnitFlashlights);
```
esteldunedain: `[_player] call FUNC(getUnitFlashlights)`is expensive, so it should only be called if it's really needed. 

I think `_flashlight isEqualTo ""` could be checked first when the inventory changes to quickly exit.

addons/map/functions/fnc_getUnitFlashlights.sqf
```diff
@@ -21,8 +21,8 @@ params ["_unit"];
 private _flashlights = [];
 
 {
-    if ((isText (configFile >> "CfgWeapons" >> _x >> "ItemInfo" >> "FlashLight" >> "ACE_Flashlight_Colour")) && !(_x in _flashlights)) then {
-        _flashlights pushBack _x;
+    if (isText (configFile >> "CfgWeapons" >> _x >> "ItemInfo" >> "FlashLight" >> "ACE_Flashlight_Colour")) then {
```
esteldunedain: The valid flashlight classnames could be cached for better performance.

addons/map/XEH_postInitClient.sqf
```diff
@@ -97,35 +78,33 @@ call FUNC(determineZoom);
 
     //illumination settings
     if (GVAR(mapIllumination)) then {
-        GVAR(flashlightInUse) = "";
-        GVAR(glow) = objNull;
-
         ["ace_playerInventoryChanged", {
```
esteldunedain: As far as I understand, this EH is meant to remove the current glow if the unit suddenly lost  it's flashlight. I'd appreciate if you could add a comment stating that.

esteldunedain:
> Keep in mind, this PVAR will be set every time a player opens or closes their map with the map light on--please advise on the performance hit of that.

Although it's probably not a huge deal, I'd recommend not going public with those PVARs for this case because of that. People often open/close the map too often (in part because Arma doesn't penalize it at all being instantaneous).

PabstMirror:
@voiperr - fixed merge conflicts and some of  estel's notes here: 
https://github.com/acemod/ACE3/tree/voiperr-flashlight_housekeeping

PR #3924, opened 2016-06-14
On Detach, Delete Used Chemlights

voiperr:
- Players can no longer reuse attached chemlights.
- Players can remove chemlights even if their inventory is full.
- Detach items before deleting them, to avoid objNull attaches.


esteldunedain:
:+1:

PR #3927, opened 2016-06-14
Fix for MikeRo Tools strict requirement for #undef usage

ViperMaul:
Fix for MikeRo Tools strict requirement for #undef usage


Glowbal:
This has merge conflicts

addons/main/script_macros.hpp
```diff
@@ -9,19 +9,18 @@
 #define DFUNC(var1) TRIPLES(ADDON,fnc,var1)
 #define DEFUNC(var1,var2) TRIPLES(DOUBLES(PREFIX,var1),fnc,var2)
 
-#define QFUNC(var1) QUOTE(DFUNC(var1))
-#define QEFUNC(var1,var2) QUOTE(DEFUNC(var1,var2))
```
jonpas: I just realized those are `DFUNC`, I am not sure if we want to keep that or not. @Glowbal @commy2 ?

Glowbal: This only matters when compiling with debug enabled, so it's not to worrisome. But I'd say if you use QFUNC you would not want to compile in the debug wrappers. A good example is config attributes referring to a call back. You would end up with completely different behaviour (need a compile instead of just a getvar).

PabstMirror: Good catch, when using `#define ENABLE_CALLSTACK` it throws an error when using QFUNC.
So I'm fairly sure we do want to undef and redefine QFUNC using DFUNC.

Glowbal: I think CBA should adjust the behaviour of QFUNC.

Glowbal:
I've added undefs for `QFUNC` and `QEFUNC` instead of using the CBA variants.

PR #3932, opened 2016-06-15
getVehPos - Tweak heli width, and put at eye level

PabstMirror:
@esteldunedain - Improvements for your PR #3342

I noticed that big helicopters had interaction points from far away (screenshot is 12m distance)

- For helicopters I added a max width of 3 meters, which should cover the body, but not the rotors.
- Made the max height the same as eye level.  This might just be personal preference, but it feels more natural to me to not have to look way up for tall vehicles like the big choppers, and trucks.

### DIFF
--- a/addons/interaction/functions/fnc_getVehiclePos.sqf
+++ b/addons/interaction/functions/fnc_getVehiclePos.sqf
@@ -9,7 +9,7 @@
  * Children actions <ARRAY>
  *
  * Example:
- * call ace_interaction_fnc_getWeaponPos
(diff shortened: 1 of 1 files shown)


esteldunedain:
Nice @PabstMirror.

> For helicopters I added a max width of 3 meters,

I've found the problem with helicopter rotors before, and I think that that's what I meant with the bounding box not being very good for some vehicles. I think this solution is good enough.

> Made the max height the same as eye level. 

Very clever.

PR #3940, opened 2016-06-17
Custom smoke particles

BaerMitUmlaut:
- Add custom smoke particles

The new particles are thicker, denser, last longer and produce bigger smoke clouds while using overall **less** particles, which means you can deploy more of them before Arma reaches the particle limit.

This affects:
- Smoke hand grenades
- 40mm smoke
- 81/120/155mm artillery smoke

Before:
before

After:
after

I plan to completly overhaul artillery smoke later, but that will use proper white phosphor. Since that will significantly change how smoke works (it burns, produces toxic fumes and also emits light), I'll split that into a seperate module. For now it just produces thicker and bigger smoke that can cover a larger area.


esteldunedain:
I'm a big fan of this. It can always be moved to optionals if most people prefer vanilla smoke.

addons/particles/config.cpp
```diff
@@ -0,0 +1,18 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"ace_common"};
+        author = ECSTRING(common,ACETeam);
+        authors[] = {""};
```
jonpas: Put your name here. ^^

jonpas:
> It can always be moved to optionals if most people prefer vanilla smoke.

In that case the other modules changing vanilla things should be in optionals. I'd say this fits realism perfectly and fits main ACE3 package perfectly. If people don't want it they can remove the PBO.

esteldunedain:
>  If people don't want it they can remove the PBO.

True

esteldunedain:
@BaerMitUmlaut, did you test at what distance from the camera de smoke starts spawning?

I used to have a problem with vanilla smoke, which apparently didn't start spawning until the camera was too close. This was a problem when marking targets for incoming CAS, which sometimes didn't see enough smoke to identify targets. Did you identify any parameter controling that behaviour?

System98:
I'm totally for improving the pathetic vanilla smoke grenades. Altough to me on the pictures the new effect looks too voluminous for a smoke grenade. Maybe it's just the picture, but judging from the smoke grenades I've seen I'd expect a smoke cloud ~20% less big or at least with less density at the egdes than shown above.

Glowbal:
I noticed that a lot of the colors are the same - should we put those in defines?

VKing6:
This is excellent.
As discussed on slack I have the following suggestions:
- Make coloured smoke particles slightly smaller, slightly shorter lived, and _maybe_ slightly lighter.
- Make white smoke grenades live for a bit longer
- Make custom ammo/particles for 40mm smoke, as they currently use exactly the same ammo/particles as the hand smoke grenades, which make far too much smoke for a 40mm grenade.

SilentSpike:
Maybe I'd have to see it in-game, but the vanilla smoke looks much better to me (though not perfect). The smoke in the changed screenshot looks too flat (I think - hard to say as I'm not experienced with particle effects).

VKing6:
Trust me, it looks a lot better in game. The texture doesn't come out in the picture.

john681611:
Silly Question: Will this actually have a effect on AI in MP? I remember if you used Blast core you would only get the illusion of more smoke cover the AI still could still see you once the vanilla smoke was not covering you.

jonpas:
@john681611 Yes, it has been specifically tested with this PR. Additionally part "illusion" was never an illusion afaik, it's just that AI will keep shooting in the general direction of where they last saw you, same happens with any other cover.

john681611:
@jonpas  Great thanks. When I meant 'Illusion' this would happen if the client ran Blastcore and the Server ran Vanilla effects that disconnect could cause issues.

jonpas:
Yes and that's normal, not an issue IMO. :)

BaerMitUmlaut:
This should run both server and client side, just like pretty much anything else from ACE, for the highest reliability. And since this uses less particles per second, you can safely run this without worrying about performance. Blastcore does a lot more than this does which is why it is so performance intensive.

I'll make a video to better show how it looks in game and address the adjustments from @VKing6 on sunday, the pictures from above don't really give the best impression of how it would look from a player perspective.

@esteldunedain I was still able to see the smoke from 20km away (particles are only affected by view and not by object distance btw.), although at that point you of course need to know where to look. I posted a screenshot from that in Slack. I assume the vanilla particles are simply too small.

BaerMitUmlaut:
I've made some adjustments, mainly based on the excellent feedback from @VKing6 on Slack. The smoke is now overall a bit smaller again (but still way bigger than vanilla), coloured smoke is smaller and doesn't last as long as white smoke (the latter having its burn time increased compared to vanilla) and 40mm smoke does now produce a smaller cloud that also doesn't last as long.

I've also uploaded a video with the current status: https://www.youtube.com/watch?v=qY58kTSFcJ0

PabstMirror:
I dislike being able to make out individual smoke particles.
I think the particles are too opaque and don't blend together.
Being able to see the particles rotate looks like a cartoon.

PR Example:


Compared to existing smoke:
untitled

esteldunedain:
> I think the particles are too opaque and don't blend together.

That's certainly a consequence of trying to reduce the amount of particles, while trying to increase smoke opacity at the same time. Indeed it might looks a bit uglier. I do prefer it though.

Bummeri:
Could you make colored smoke last longer but produce less smoke. 
This way they would be better for marking and signaling.

BaerMitUmlaut:
Stray particles can happen due to the added random velocity. I can reduce that factor easily and it will happen less and the particles might blend in more. The rotation of individual particles will also be less apparent.

Vanilla blends easier due to the particles barely having any opacity at all - this makes for extremely smooth edges though which neither looks good nor realistic, compare it to this:
m83
I'm not going to go for smooth edges because of this.

@Bummeri:
We're going for realism here as well, and the M18 smoke grenade lasts 50-90 seconds. The burn time of it in game is going to be 60 seconds (just like in vanilla), and each particle lasts 30 seconds, 10 seconds longer than vanilla. You're effectively going to get a good smoke signal for around ~70-80 seconds (60s burn time + 30s particle lifetime - 5-10s deployment - 5-10s faded).

BaerMitUmlaut:
I've managed to fix both the popping and stray particles. The popping was caused by particles going under the terrain and then partially under the sea level, which caused them to disappear.

rebelvg:
For some reason not all RHS grenades inherit this smoke.
"rhs_mag_rdg2_white" for example.

kerckasha:
I can't seem to get 82mm smoke to work, I see the projectile land but it doesn't emit smoke

BaerMitUmlaut:
Please create an issue if you encounter a bug.

PR #3431, opened 2016-02-26
Rearm limited ammo truck munition

GitHawk:
The rearm module will be expanded by a new option to select the ammunition amount in an ammo truck.
The two new availability levels are:
- Amount based on caliber
- Certain magazines and specific amounts
  The current "unlimited" level will remain as default value.

Supply amount based on caliber means you can for example take either two .50cal magazines or a single 120mm magazine, but not a 155mm magazine when the amount is on the level of 154. So everything can be rearmed but with a limited supply, while bigger calibers take more supply.

Supply amount with certain magazines means you are required to add certain magazines to an ammo truck. No other weapons/magazines will be able to be resupplied.
- [x] Implement the two levels
- [x] Add API
- [x] Testing

Documentation for other modders/creators on how to enable refuel will be in a different PR

PR #3944, opened 2016-06-18
Cleanup Function Headers Titles

jonpas:
- `Argument:` -> `Arguments:`
- `Return value:` -> `Return Value:`
- Redone from #3431


Glowbal:
:+1:

PR #3945, opened 2016-06-18
Fix no gray screen for advanced medical

Glowbal:
Advanced medical was never displaying the gray screen on low blood volume. This is because for advanced medical, the blood volume minimum level is around 40 before death. For basic medical, this is 0.

This PR scales the gray screen for advanced medical to the correct levels.

#3134


Glowbal:
Tested and works beautifully.

PR #3947, opened 2016-06-18
Fix ability to carry a chair someone is sitting on

jonpas:
- Use claim system to prevent carrying when someone is sitting
- Change module scope back to 2 until ACEX is released


Glowbal:
Please PR to acemod/ACEX as well.

PR #3948, opened 2016-06-18
Improve zeus auto-add XEH behaviour

SilentSpike:
- Add the InitPost XEH event to `AllVehicles` only if the `autoAddObjects` setting is enabled

The XEH InitPost event only needs to be added if the `GVAR(autoAddObjects)` setting is enabled. This change will save on overhead when the setting is disabled and also removes some complexity from the function that runs in response to the event.


Glowbal:
Very good improvement @SilentSpike

PR #3955, opened 2016-06-19
add 'name' to components

commy2:
- adds 'name' to all components by macros
- fixes some indentation issues and uses ADDON macro in parachute


Glowbal:
What is the purpose of the name attribute in CfgPatches? Is this a new BI thing?

SzwedzikPL:
> What is the purpose of the name attribute in CfgPatches? Is this a new BI thing?

Yes. Name from CfgPatches is displayed in required addons list in eden

commy2:
And this patch will change "ace_gforces" to "Advanced Combat Environment 3 - gforces"

bux578:
I don't like the long name. ACE3 is sufficient. And name should contain the file name of the pbo.

commy2:
There is enough space for that name. Vanilla Arma has even longer ones.

>And name should contain the file name of the pbo.
You already get the name of pbo if you select the "advanced" tab. This screen is for end users.

commy2:
`Arma 3 Marksmen (Multiplayer) - Mission Texts and Translations`
vs.
`Advanced Combat Environment 3 - spottingscope`

bux578:
> This screen is for end users.

Exactly.

"ACE3" is more common. And it's the end user that deletes random pbo's. So giving them the info what pbo they require would be helpful.

commy2:
Yeah.
I wouldn't use the raw pbo name though. That's what you get in the "advanced" tab already. It would defeat the purpose of this system.

jonpas:
I vote `ACE3 - component` as well.

Would be great if we could prettify components but that's some work I guess. :D

commy2:
Thanks for the help, guys. :+1:

commy2:
merging, sue me

bux578:
_sues commy_

PR #3958, opened 2016-06-19
Fix freeing prisoners after they've switched locality

### LINKED ISSUE #3950: Handcuffed unit can't be freed properly after locality change
**Arma 3 Version:** `1.60 (stable)`
**CBA Version:** `2.4.1`
**ACE3 Version:** `latest master`

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- If you handcuff playable unit without player (owned by server/headless) you can't free it properly when player enter this unit. This happens becouse variable with EH ID (AnimChaned) is local and it's stored only on machine that owned unit when it was handcuffed. I think making https://github.com/acemod/ACE3/blob/master/addons/captives/functions/fnc_setHandcuffed.sqf#L99 global will solve problem but it needs to be checked.

**Steps to reproduce:**
- Handcuff playable AI
- Make someone enter this slot
- Try to free him

- Dedicated

**Placed Modules:**
- Not relevant

**RPT log file:**
- None

esteldunedain:
- Make sure that if a captive unit switches locality, it's AnimChanged EH switches with it.
- Close #3950

I only did limited testing, but it seems to work as expected.

### DIFF
--- a/addons/captives/CfgEventHandlers.hpp
+++ b/addons/captives/CfgEventHandlers.hpp
@@ -55,3 +55,11 @@ class Extended_Respawn_EventHandlers {
         };
     };
 };
+
+class Extended_Local_EventHandlers {
+    class CAManBase {
+        class ADDON {
+            local = QUOTE(call FUNC(handleLocal));
+        };
+    };
+};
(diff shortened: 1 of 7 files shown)


esteldunedain:
I didn't have time to properly test the last two commits. i'll try to do that tomorrow.

Glowbal:
:+1:

esteldunedain:
I've done some more testing and everything seems to be working properly.

PR #3962, opened 2016-06-20
Dogtags

esteldunedain:
- Merge PR #3600 to a `dogtag` branch in our repo
- Restyle the dogtags
- Add unique service number and blood types to all dogtags

I think this is ready for 3.6.0. Improvements can be pushed for later, given they wouldn't interfere (there's no public API yet).

**Proposed by @jonpas :**
- [ ] Change the dog tag items to magazines and use their unique ID as identifier

I'm been thinking about the proposal, and although it's easy to implement. However, I think it's not needed for first release.


addons/dogtags/functions/fnc_getDogtagData.sqf
```diff
@@ -0,0 +1,32 @@
...
+private _dogTagData = _target getVariable QGVAR(dogtagData);
+if (!isNil "_dogTagData") exitWith {_dogTagData};
+
+// Create dog tag data once for the unit: nickname, code (eg. 135-13-900) and blood type
+private _dogTagData = [
+	[_target, false, true] call EFUNC(common,getName),
+	str(floor random 9) + str(floor random 9) + str(floor random 9) + "-" +
+		str(floor random 9) + str(floor random 9) + "-" +
+		str(floor random 9) + str(floor random 9) + str(floor random 9),
+	selectRandom ["O POS", "O NEG", "A POS", "A NEG", "B POS", "B NEG",
```
commy2: tab :)

commy2:
Im bringing up a silly idea again I had back in AGM.
Why don't we dynamically create the blood types using the player name or steam ui or something like that.
It would mean that these blood types are consistent from mission to mission.

esteldunedain:
Incoming

SilentSpike:
I don't think that's silly, I like it :+1:

I'd probably do a `parseNumber` on the `getPlayerUID` (ID64) and then wrap around the array of types like so: `_arr = [blablabla]; _arr select ((parseNumber (getPlayerUID _target)) % (count _arr))`

addons/dogtags/functions/fnc_getDogtagData.sqf
```diff
@@ -19,13 +19,24 @@ private _dogTagData = _target getVariable QGVAR(dogtagData);
 if (!isNil "_dogTagData") exitWith {_dogTagData};
 
 // Create dog tag data once for the unit: nickname, code (eg. 135-13-900) and blood type
+private ["_bloodTypeIndex", "_code"];
+//hint getplayeruid _target;
+if ((typename getplayeruid _target) == "SCALAR") then {
```
SilentSpike: This will never be true, also not using `isEqualType` :stuck_out_tongue_closed_eyes:

esteldunedain: Why wouldn't it work?

commy2:
>Incoming
I think I was faster, haha

addons/dogtags/functions/fnc_getDogtagData.sqf
commy2: I don't get the point of this.

SilentSpike: `getPlayerUID` always returns either a string or nil (for AI).

commy2: `getPlayerUID` is the correct case. As far as I can tell it will always report a string, so it will never be `typeOf` `SCALAR`

SilentSpike:
I don't wanna commit while you're working on stuff, but maybe this would work better:

esteldunedain:
getPlayerUID doen't work for this, because it returns "" for dead players. I'll switch to something inspired on @commy2 's code

addons/dogtags/functions/fnc_getDogtagData.sqf
```diff
@@ -20,12 +20,11 @@ if (!isNil "_dogTagData") exitWith {_dogTagData};
 
 // Create dog tag data once for the unit: nickname, code (eg. 135-13-900) and blood type
 private ["_bloodTypeIndex", "_code"];
-//hint getplayeruid _target;
-if ((typename getplayeruid _target) == "SCALAR") then {
-    _bloodTypeIndex = (getplayeruid _target) mod 8;
-    _code = (str getplayeruid _target) select [0, 3] + "-" +
-            (str getplayeruid _target) select [3, 2] + "-" +
-            (str getplayeruid _target) select [5, 3];
+if !(getplayeruid _target in ["", "_SP_AI_", "_SP_PLAYER_"]) then {
```
SilentSpike: Minor, but it would probably make sense to store the value of `getPlayerUID` in a variable instead of using the command multiple times (also the lack of camel case triggers me :camel:)

SilentSpike: Otherwise :+1:

esteldunedain:
Actually, I just merged @commy2 's PR. The service number is not persistent anymore, but that can be improved in the future.

SilentSpike:
Service number could be made persistent using:
```
_nums = ((toArray _name) select [0,8]) apply { _x % 10 };

while {count _nums < 8} do {
    _nums pushBack (floor random 10);
};
```

Then just split as necessary. I'm not familiar with the format of service numbers, so not sure if the `floor random 9` in the existing code was an oversight or if they only use 0-8.

esteldunedain:
@SilentSpike, they totally go from 0 to 8

esteldunedain:
Ok guys, do you think this should be in 3.6.0? 

:+1: :-1:

SilentSpike:
Functional (presumably) version of above snippet:
```
_nums = ((toArray _name) select [0,8]) apply { _x % 10 };

while {count _nums < 8} do {
    _nums pushBack (floor random 10);
};

_code = ([_nums select [0,3],_nums select [3,2], _nums select [5,3]] apply { _x joinString "" }) joinString "-";
```

Glowbal:
Blood type? I'm not so sure about that.

esteldunedain:
> Blood type? I'm not so sure about that.

I guess you may argue that we shouldn't give players different blood types if we don't simulate blood compatibility in medical. But you might as well say we do, only that blood bags contain O Negative blood, which is close to universally compatible.

Anyway, I just wanted dogtags to look as genuine as possible, and they usually do feature blood types on them. I could randomize religion instead, but that sounds even more controversial.

jonpas:
I like blood type, religion is kinda meh unless we put "Jedi" on all of them.

654wak654:
About `getPlayerUID` returning an empty string on dead bodies, couldn't you go around that by generating the blood type before a player dies? It only has to be done once _ever_ anyway. Could even put it in a profile var and let people play around with it.

jonpas:
I support @kripto202 's idea, should be straightfoward implementation with `ace_placedInBodyBag`.

SilentSpike:
>couldn't you go around that by generating the blood type before a player dies?

This is true, we could use an XEH PostInit or something, however the way it's done now means that code doesn't run unless it needs to. You could go a whole mission without anybody checking dogtags.

PR #3965, opened 2016-06-20
Fix issue with dogtag respawn/setVariable

PabstMirror:
Because you can take dogtags when they are just unconscios, you can set the `dogtagTaken` var before they die, so it will get copied over to respawned unit.

This PR will:
Store object instead of bool to indicate when dogtags have been taken
When object re-spawns, the the old object reference won't apply


esteldunedain:
:+1:

commy2:
I really like that method.

esteldunedain:
Yup, it's very neat

PR #3966, opened 2016-06-21
convert nametags to cba settings

commy2:
I broke the last PR by messing with rebase ...
- [x] TODO, clear up some stringtable entries


commy2:
done. works as far as I can tell.

I fixed the "showCrewInfo". Setting was called "showVehicleCrewInfo"
Also removed unused strings from AGM which make no sense anymore

esteldunedain:
@commy2, how feasable do you think it would be to add to CBA settings the possiblity to "greying out" or "disabling" settings according to the value of a master setting?

e.g. A lot of settings are valid only if  "Show Names" is true.

PabstMirror:
@esteldunedain - I've thought of something like this as well.
Could add something like a simple condition string that gets evaluated.

`condition = "ace_medical_level >= 2";`
Example, setting medical to basic would grey out all the advanced settings.

This feature could added later as well.

esteldunedain:
Yeah, if possible that'd be perfect. 

It would require setting the settings at the same time they are changed. That might be problematic while changing them in the editor (e.g. no missionNamespace). Or maybe not, IDK

commy2:
What would be the benefit of greying them out though? They still have to take a valid state (either on or off for a button for example).
All that would do, is force you to enable the master setting when attempting to change the child setting.

esteldunedain:
> What would be the benefit of greying them out though? They still have to take a valid state (either on or off for a button for example).

Well, in fact I'm mostly thinking about "client settable" settings.

E.g. in nametags, medical or interaction users are presented with a lot of settings that do nothing unless some other settings are in the correct value. That's confusing and it makes settings descriptions more convoluted:

`Show name tags only on keypress (requires "Show Names")`

Well, it's a usability feature. 

> All that would do, is force you to enable the master setting when attempting to change the child setting.

Maybe only grey them out at runtime?

commy2:
I understand, but I don't think it's that bad. You'll most likely only have one of these master settings in a hand full of categories.

esteldunedain:
> You'll most likely only have one of these master settings in a hand full of categories.

That's probably true.

esteldunedain:
I wonder; what if there was a counterpart to `CBA_settings_fnc_init`, like `CBA_settings_fnc_remove`?

That way, it would be possible to init/remove conditional settings in real time from a setting changed EH, without modifying the profile stored settings at all.

commy2:
Well, in theory. The settings menu isn't updated in real time (only "connected" controls of the same settings are)
There is a "gui refresh" function, but that can't remove settings either. You'd have to adjust the position of all controls below the deleted ones or you'd get gaps.
I mean I could do that - write a super flexible on the fly changing settings menu, but I'd have to change quite a bit.

esteldunedain:
Yeah, I imagine is quite a job. Well, it's just an idea. As @PabstMirror, it might be possible to do down the line.

PR #3969, opened 2016-06-21
Add dogtag actions to body bags

PabstMirror:
Adds dogtag actions to bodybags based on idea from @kripto202 in #3962


addons/medical/functions/fnc_serverRemoveBody.sqf
```diff
@@ -27,7 +27,7 @@ if (isNil QGVAR(bodiesToDelete)) then {GVAR(bodiesToDelete) = [];};
 GVAR(bodiesToDelete) pushBack _target;
 
 // Start up a loop to wait for bodies to be free to delete
-if ((count GVAR(bodiesToDelete)) > 0) then {
```
esteldunedain: Oops; good catch

esteldunedain:
Thanks @PabstMirror; very nicely handled.

PR #3971, opened 2016-06-21
Generate SSN of unit from their name

SilentSpike:
Similar to how we're handling blood types, uses the unit's name to generate a valid three/two/four format SSN.

For names less than 9 characters long this will currently generate a unique SSN only up to the length of the name and append valid digits as necessary. Not sure how best to handle those cases.

Output of some tests:

SilentSpike: 358-10-6325
esteldunedain: 156-18-0701
commy2: 919-91-0645 / 919-91-0774 / 919-91-0750
Glowbal: 181-98-7836 / 181-98-7837 / 181-98-7805
jonpas: 610-27-5955 / 610-27-5081 / 610-27-5438

#3962

### DIFF
--- a/addons/dogtags/XEH_PREP.hpp
+++ b/addons/dogtags/XEH_PREP.hpp
@@ -9,4 +9,5 @@ PREP(getDogtagData);
 PREP(getDogtagItem);
 PREP(sendDogtagData);
 PREP(showDogtag);
+PREP(ssn);
 PREP(takeDogtag);
--- a/addons/dogtags/functions/fnc_getDogtagData.sqf
+++ b/addons/dogtags/functions/fnc_getDogtagData.sqf
@@ -23,9 +23,7 @@ private _targetName = [_target, false, true] call EFUNC(common,getName);
 
 private _dogTagData = [
     _targetName,
-    str(floor random 10) + str(floor random 10) + str(floor random 10) + "-" +
-        str(floor random 10) + str(floor random 10) + "-" +
-        str(floor random 10) + str(floor random 10) + str(floor random 10),
+    _targetName call FUNC(ssn),
     _targetName call FUNC(bloodType)
 ];
 // Store it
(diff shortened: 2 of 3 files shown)


esteldunedain:
You could repeat the name until you get the 8 characters you need, eg: 

eg:
commycom
jonpasjo

Glowbal:
Provided they are not already used in that case.

SilentSpike:
Well, this isn't generating globally unique SSNs, that would probably be overkill for a vanity feature like this. Just unique to the names, so if two people shared the first 9 characters in their name they'd get the same SSN (but that could have happened previously anyway where it was totally random).

SilentSpike:
I could perhaps use the length of the name to influence the generation, would make things more unique and also possibly allow a method for handling short names.

voiperr:
Couldn't this be mitigated by storing the unit's `getPlayerUID`, like ACE stores the name? And just generate a random number for AI?

SilentSpike:
That would allow for global uniqueness and make length a non-issue, the question is whether it's worth storing that information just for this purpose. About to push some updated generation code either way.

SilentSpike:
Tests with new generation:

SilentSpike: 810-63-2571
esteldunedain: 807-01-0750
commy2: 919-91-0991
Glowbal: 181-98-7811
jonpas: 610-27-5607

SilentSpike:
Since I know not everybody reads commit messages:

>Names shorter than the required 9 characters are now reused until we have 9 characters - making for fully unique generation.
>
>To keep things distributed as pleasantly as possible they're itterated over with a step of two - alternating between odd and even indexes every full loop.
>
>Similarly, the slice of string used for names longer than 9 characters is now offset based on the length of the name.

commy2:
>919-91-0991
nice

esteldunedain:
Perfect @SilentSpike!

Is it me or we really ended up overengineering this little piece of metal :)

bux578:
> 919-91-0991

called that phone number, an old lady answered. Didn't know about a "commy".

SilentSpike:
If we wanted to over-engineer this further, short names could also use `BIS_fnc_phoneticalWord` to get new characters and reduce repetition :wink:

PR #3972, opened 2016-06-21
convert common events to CBA playerEvents

commy2:
TODO:
- [x] remove PFH in common
- [ ] call ACE events as wrapper, deprecate?

PR #3973, opened 2016-06-22
Convert state checking to CBA playerEH

PabstMirror:
For PR #3972

Still thinking about `ace_playerInventoryChanged` as right now this adds an extra call to `FUNC(getAllGear)` even when there are no listeners to the old event.


commy2:
>Still thinking about ace_playerInventoryChanged as right now this adds an extra call to FUNC(getAllGear) even when there are no listeners to the old event.

That doesn't worry me much. Current stable will do that every frame. Now the code won't be executed once for whole minutes. Events, yay.

PR #3975, opened 2016-06-22
convert common events to CBA playerEvents, target: release branch

commy2:
- See: https://github.com/acemod/ACE3/pull/3972


Glowbal:
CBA player events work with remote controlled units, no?

commy2:
yes. they are essentially the same, but with this I can use the same loop for BWA3, so no unnecessary overhead / strange work arounds without ACE.

commy2:
See here:
https://github.com/CBATeam/CBA_A3/blob/master/addons/events/fnc_addPlayerEventHandler.sqf#L85

It's the same thing we have in ACE, but with minor improvements.

PR #3983, opened 2016-06-24
Block FCS when controlling a UAV

PabstMirror:
Fix doing FCS keyUp calculations/`setVariable`/click sound
when controlling a UAV while also in a turret with FCS enabled.

Normal rangefinder on the UAV will still work as FUNC(canUseRangefinder) is still true.


Glowbal:
This should target the Apex branch.

addons/fcs/functions/fnc_canUseFCS.sqf
```diff
@@ -18,3 +18,4 @@
 getNumber ([configFile >> "CfgVehicles" >> typeOf vehicle ACE_player, [ACE_player] call EFUNC(common,getTurretIndex)] call EFUNC(common,getTurretConfigPath) >> QGVAR(Enabled)) == 1
 && {cameraView == "GUNNER"}
 && {!([ACE_player] call CBA_fnc_canUseWeapon)} //Not Turned Out
+&& {cameraOn != (getConnectedUAV ACE_player)} //Not Controlling a UAV
```
commy2: would this break the FCS while remote controlling a unit via zeus? Does `getConnectedUAV` include those AIs?

PabstMirror: When zues RCing another unit `getConnectedUAV player` and `getConnectedUAV ace_player` are both null

PR #3984, opened 2016-06-25
Handle "loadout" changed event when player is null

PabstMirror:
When player is null, gear is `[]`

```
 Error position: <select 6;
 Error Zero divisor
File z\ace\addons\parachute\functions\fnc_storeParachute.sqf, line 5
```


addons/parachute/functions/fnc_storeParachute.sqf
```diff
@@ -17,7 +17,7 @@
 #include "script_component.hpp"
 
 params ["_unit", "_gear"];
-private _backpack = _gear select 6;
+private _backpack = _gear param [6, ""];
```
commy2: Isnt the backpack:
`_gear param [5, []] param [0, ""]` ???

Or is this using FUNC(getAllGear) ?

Glowbal:
Can be merged when @commy2 's comment has been answered. I think it's using getAllGear but haven't checked yet.

PR #3986, opened 2016-06-25
missleGuidance - Fix seeker los check

### LINKED ISSUE #3374: Check undefined func in missile guidance
https://github.com/acemod/ACE3/blob/master/addons/missileguidance/functions/fnc_seekerType_Optic.sqf#L30

FUNC(checkSeekerLos) does not exist; look at replacing with FUNC(checkLos)
Ref #3368

PabstMirror:
Close #3374
Because func was undefined `_losOkay` was nil, so target leading was always skipped.


commy2:
3.6.0 ?

bux578:
I agree @commy2

commy2:
rhymes* : -)  

*if you leave out the "two"

PR #3989, opened 2016-06-25
disable auto weather - bugging out ACE weather

commy2:
- Disable the auto weather settings
- Which makes sure the weather SQF commands don't bug out


PabstMirror:
Some of ace_weather can be disabld via settings, so we might want to sill allow vanilla auto-weather.

Could we just keep the `onMissionNew` part, so new missions will default to good settings, but still allow people to re-enable?

commy2:
Yes, we could.

jonpas:
Neat fix! :+1:

esteldunedain:
Should we print a warning and disable ace weather simulation if the commands are bugged?

commy2:
Probably yes. I'm not sure which places are to check.

commy2:
Check would be:
```
if (["Intel" get3DENMissionAttribute "IntelRainIsForced"] param [0, true]) then {
    // warning
};
```
for each of these weather attributes

commy2:
oops, I was wrong. That only works in the editor and the previews.
Apparently it's this verbose monstrum:


`[getMissionConfigValue ["IntelRainIsForced", "intel" get3DENMissionAttribute "IntelRainIsForced"]] param [0, true]`

PR #3990, opened 2016-06-25
ACE_Finger cleanup

### LINKED ISSUE #1917: Finish Enhanchments to pointing
Finger is functional, but I didn't finish all the notes from @esteldunedain 
See #1548

esteldunedain:
- Add privates
- Replace ad-hoc function with EFUNC(common,getZoom)
- Make pointing error proportional to the distance to the target
- Close #1917
- Allow pointing to nearby positions:

"Soldier, pick that weapon up"
image


PabstMirror:
This is fantastic; I think it also matches how the 1.62 pointing will work 
(except that is global which we don't want)

esteldunedain:
>  I think it also matches how the 1.62 pointing will work 

Ouch, I didn't know they were planning such a thing. Is there any info about it?

bux578:
@esteldunedain https://www.youtube.com/watch?v=Jy68aqB2-l4

esteldunedain:
Aha, yes; this PR works exactly like that, except it's:
1- Only shown to nearby units
2- Fades after a sec
3- It has a hand animation

BIS keeps ripping out mods :grin:

jonpas:
> BIS keeps ripping out mods :grin:

Sadly making it worse in those 2 cases (pointing and squad radar).

PR #3994, opened 2016-06-25
Russian rank icons

commy2:
- Adds russian rank icons from @Skengman2 <3
- Implements them for Opfor and RHS_AFRF.


commy2:
oops !3995

PR #3995, opened 2016-06-25
add Russian rank icons

commy2:
- Adds russian rank icons from @Skengman2 <3
- Implements them for Opfor and RHS_AFRF.


esteldunedain:
Very nice!

@commy2, add @Skengman2 to authors please.

PabstMirror:
Would it make any sense to add these to `CfgFactionClasses` config?

`OPF_F` aren't russian in arma 3 (they're iran/chineese), but it should be ok to add.

RHS ARF seems to have a lot of factions (but I'm not sure if they are used/have units)

rhs_faction_msv
rhs_faction_vdv
rhs_faction_vmf
rhs_faction_vv
rhs_faction_tv
rhs_faction_vpvo
rhs_faction_vvs
rhs_faction_vvs_c
rhs_faction_rva

esteldunedain:
Ohg @PabstMirror is right. Russian are under independent in RHS

commy2:
>OPF_F aren't russian in arma 3 (they're iran/chineese), but it should be ok to add.

Currently they have american ranks, which makes even less sense. It was a good excuse to get them into non-optionals ACE. That way RHS and CUP can use them (I send message to CUP guys already)

I'll add the other RHS factions.

@esteldunedain side != faction. Doesn't matter which side these are on.

esteldunedain:
Apparently iranian use ranks very similar or equal to americans

https://en.wikipedia.org/wiki/Rank_insignia_of_the_Iranian_military

commy2:
>Apparently iranian use ranks very similar or equal to americans

Oh wow. I didn't know.
I guess it makes sense to change them back then? Can we still keep the assets in nametags pls?

commy2:
>Would it make any sense to add these to CfgFactionClasses config?

They could be there I guess, but I see no benefit. This is more flexible.

Skengman2:
Hey I've updated the rank icons a little bit so the Officer ranks weren't blurry: 
[fixed_russian_rank_icons_ace3.zip](https://github.com/acemod/ACE3/files/333331/fixed_russian_rank_icons_ace3.zip)

PR #3997, opened 2016-06-26
Add goggle setting for level of effects

PabstMirror:
- Add a ace_setting for change the level of effect simulation or goggles.

Can chose between none, tint/colorCC and full dust/dirt/rain/rotorwash.

XEH_postInit diff is a mess but it's mainly just skipping adding firedEH/PFEH depending on setting.

Not sure if this should be a client setting??

### DIFF
--- a/addons/goggles/ACE_Settings.hpp
+++ b/addons/goggles/ACE_Settings.hpp
@@ -1,11 +1,11 @@
 
 class ACE_Settings {
-    /*class GVAR(enable) { // @todo
-        value = 0;
-        typeName = "BOOL";
-        isClientSettable = 1;
-        displayName = CSTRING(enable);
-    };*/
+    class GVAR(effects) {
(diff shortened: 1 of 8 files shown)


jonpas:
I'd say that should be a client setting, it's merely visuals and who wants to enforce it can still do so.

esteldunedain:
:+1:

jonpas:
Add it as client setting @PabstMirror ?

PabstMirror:
Making it a setting was easy as it's just a reorganization of the existing init.

clientSettable would be a a bit more complex, instead of just not adding something we need to make sure it gets cleaned up. (and no way to undo `CBA_fnc_addClassEventHandler`)

jonpas:
Ah I see, no way to remove class EHs. I am good with this then, feel free to merge.

PR #4009, opened 2016-06-27
Fix rearm events

PabstMirror:
Rearm was doing some weird stuff with old execRemoteFnc:
`_turretOwnerID publicVariableClient QEGVAR(common,remoteFnc);` 

(https://github.com/acemod/ACE3/commit/0723df7dea5c0a0edbf5dd6f8698dbc93a615e87#diff-e700393f0d5b2d9446845839cb7e854bL40)

We can't do events to owners until cba 2.4.2, so I switched it to just blast a global event and filter using `turretLocal` on all recievers so it works.

Still need to test on dedicated, will do tomorrow.


Glowbal:
Can be merged when tested.

jonpas:
If only testing this would be easy and fast. :P

PabstMirror:
Seems to be working ok in mp

PR #4014, opened 2016-06-27
Handle bad target when using apex titans

### LINKED ISSUE #4007: Missileguidance error when trying to lock
**Arma 3 Version:** `1.63` dev
**CBA Version:** `2.4.1` stable
**ACE3 Version:** `3.6.0` latest RC

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- When trying to lock something with the Titan AT the RPT file will be spammed.

**Steps to reproduce:**
- Try to lock something, check the RPT

- SP / MP

**Placed Modules:**
- None

**RPT log file:**
```
if(!(terrainIntersectASL [ _seekerPos, _targe>
 3:44:55   Error position: <terrainIntersectASL [ _seekerPos, _targe>
 3:44:55   Error 0 elements provided, 3 expected
 3:44:55 File z\ace\addons\missileguidance\functions\fnc_checkLos.sqf, line 11
 3:44:55 Error in expression <etPosASL _seeker;
_return = true;

```

PabstMirror:
Fix #4007

The onFired stuff is just a bandAid until 1.62 is out. 
We need to explicitly add config's to the new weapons (it's designed to ignore inheritance)

### DIFF
--- a/addons/missileguidance/functions/fnc_checkLos.sqf
+++ b/addons/missileguidance/functions/fnc_checkLos.sqf
@@ -14,6 +14,8 @@ private["_seeker", "_seekerPos", "_target", "_targetPos", "_return", "_vectorTo"
 _seeker = _this select 0;
 _target = _this select 1;
 
+if ((isNil "_seeker") || {isNil "_target"}) exitWith {false};
+
 _targetPos = getPosASL _target;
 _seekerPos = getPosASL _seeker;
 _return = true;
--- a/addons/missileguidance/functions/fnc_onFired.sqf
+++ b/addons/missileguidance/functions/fnc_onFired.sqf
@@ -70,6 +70,12 @@ if(isNil "_target") then {
     };
 };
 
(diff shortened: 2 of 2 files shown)


Glowbal:
Do we need this? Can we prepare the fix on the Apex branch instead and release that as soon as Apex gets released?

commy2:
The script shouldn't bug out just because there are undefined config values in the first place.

Glowbal:
Agreed @commy2.

Taking a quick look at the code, it's probably among of the older pieces within ACE. If we want to do a proper fix, we should probably dust it off a bit.

PabstMirror:
`missileguidance` expects there to be variables set for the ammo type, but `javelin` only runs on weapons and ignores inheritance to solve problems with people creating addon weapons inheriting from the titans.

I would like to have this for 3.6 just for groups that don't update often or people checking out dev.
It's easy enough to remove the hard-coded classnames for 3.6.1

PR #4021, opened 2016-06-28
Added some Dutch translations

Mad-0:
Added Dutch translations for some strings.


commy2:
Dutch is not supported by the game?

bux578:
Thank you for the effort. But is Dutch even supported by the game?

bux578:
@Glowbal will know :)

jokoho48:
https://support.bistudio.com/arma-3/steam/68 no its not supported

Glowbal:
We appreciate the effort, but Dutch is not an by Arma 3 supported language. We therefore won't include it in our repository.

Note that if you are interested in having a fully Dutch translated version of ACE, there are ways to do so. iirc the Chinese community has done something like that. @bux578 will know.

PR #4023, opened 2016-06-29
Fix no attack profile on static titan launcher

PabstMirror:
Dir/Top selector didn't work on static titans as it was storing variable on player instead of staticWeapon.


Glowbal:
:+1:

PR #4024, opened 2016-06-29
convert finger to cba settings

commy2:
- title


PabstMirror:
image

@commy2 - Based on what i'm seeing, I can't set these settings at the mission level because they are global, and global only works with server settings?

I think we would want to switch cba's `get` to logic like this:
```
if (([_setting, "mission"] call FUNC(isForced)) && {!([_setting, "server"] call FUNC(isForced))}) then {
    _value = [_setting, "mission"] call FUNC(get);
} else {
    _value = [_setting, "server"] call FUNC(get);
};
```
So we can force global settings at either the server or mission level.

commy2:
>I think we would want to switch cba's get to logic like this

Wouldn't that mean that not-global settings are always server forced if they are not mission forced?

I think to fix this we should make isForced report `false` for "server" if the setting is forced by the mission.

Issue #3724, opened 2016-04-26
Click to drop system issues when dragging/carrying units in ACE


MusicalAnvil:
**Arma 3 Version:** `1.58 stable` (stable / rc / dev)
**CBA Version:** `2.3.1 stable` (stable / dev + commit hash)
**ACE3 Version:** `3.5.1 stable` (stable / dev + commit hash)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- No way to drop dragged or carried units via action menu or interaction menu, causing difficult situations when framerate drops below a certain threshold (~15 fps, common on coop missions with heavy player/AI count and ACE enabled).
- New 'click to drop' system requires spamming the left click button at these low FPS, potentially causing misfires and friendly fire situations.

**Steps to reproduce:**
- Pick up a unit
- Drop framerate to around 10 or 15
- Attempt to drop unit using the click command

**Where did the issue occur?**
- Anywhere

**Placed Modules:**
- None

**Request**
- Please add the option to drop a carried or dragged unit in the self-interaction menu and/or action menu, like what was present in previous versions.

commy2:
hold down LMB until it releases. It's a bug with the inputAction command and only BI can fix that

PabstMirror:
Added a keybind for drag drop in 3.6.1 that will also function as a release key.

PR #4034, opened 2016-07-01
Add Drag/Drop Keybind

PabstMirror:
- Adds a keybind to start dragging the cursorObject.
- If player is currently dragging/carring it will be dropped (Just like Joko's #2333)
- Unbound by default.

I had a feature request for this from someone who had trouble finding the drag interaction point on a body that was laying oddly. Given all the interaction points on people it is easy for them to overlap; so this does seem reasonable to me.

Should let us close out https://github.com/acemod/ACE3/issues/2284#issuecomment-136550815
and might work for #3724


esteldunedain:
:+1:

PR #4037, opened 2016-07-02
Fix cargo event name - spare parts not being added

PabstMirror:
Fix Spare parts not being added in 3.6
Cargo event was renamed.


esteldunedain:
Sorry, my mistake :disappointed:

PR #4040, opened 2016-07-03
Less aggressive overlap detection

PabstMirror:
This should be a fix for a side effect of #3342
I could not find the interaction point for the attached refuel nozzle.
In the cropped screenshot below, the refuel action would not be shown without this fix.

Interaction _sPos: [0.768666,0.525559,0.346336]
Refuel _sPos: [0.453124,0.457705,1.23955]

Delta: [-0.332239,-0.0714445,0.940485]
Distance2d: 0.322755

I think the vehicle's dynamic position is pushed so close to the player that it obscures things that it shouldn't.
This PR adds a 2nd requirement that the 2d screen pos distance needs to be small enough to make obscuring worthwhile.


esteldunedain:
Thanks @PabstMirror

PR #4043, opened 2016-07-03
Update showHud func for 1.62 (squad radar)

PabstMirror:
In 1.62 `showHud`/`shownHud` now has 9 arguements.
- Changed ace_common_fnc_showHud to use 9 (and be backwards compatible)
- Updated all calls to func to use 9

### DIFF
--- a/addons/captives/functions/fnc_handlePlayerChanged.sqf
+++ b/addons/captives/functions/fnc_handlePlayerChanged.sqf
@@ -21,7 +21,7 @@ params ["_newUnit","_oldUnit"];
 //set showHUD based on new unit status:
 if ((_newUnit getVariable [QGVAR(isHandcuffed), false]) || {_newUnit getVariable [QGVAR(isSurrendering), false]}) then {
     TRACE_1("Player Change (showHUD false)",_newUnit);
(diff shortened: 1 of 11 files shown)


jonpas:
:+1: 
Self-note when this gets merged: add to client UI settings.

PabstMirror:
I was going to ask you if we wanted to do that, I can add it to this PR.

jonpas:
Yeah, it's only bound to difficulty in vanilla I think, no other way to toggle it, so I'd add it. I'll add it to this PR so you don't have to search where all it has to be added. :)

jonpas:
Done from my side, I can't test it right now as I don't have RC downloaded.

PabstMirror:
Thanks for UI intergration @jonpas 

Last dev rep says squad radar won't be in 1.62
So this will probably won't be part of 3.6.1

jonpas:
Indeed, tagging with next milestone, we can make sure during the weekend, as the actual squad radar may not be there but the command may already support it.

jonpas:
Not part of 1.62, retargeted into `master` branch in #4085

PR #4045, opened 2016-07-03
Fix overheating spare barrels

### LINKED ISSUE #4044: Swap Barrel Action isn't working

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- it isn't working. neither as a selfinteraction, nor as an assistant automatic rifleman
**Steps to reproduce:**
- place a automatic rifleman on map grab an additional Barrel and try to swap barrel
- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer)
**Placed Modules:**
- zeus
**RPT log file:**
- http://pastebin.com/H76zQeWU

PabstMirror:
Fix #4044 

- Remove ACE_SpareBarrel from CfgWeapons as it is now a magazine
- Don't call updateTemperature in displayTemp, as unit may not be local
- Fix targetEvent error in loadCoolestSpareBarrel

### DIFF
--- a/addons/overheating/CfgWeapons.hpp
+++ b/addons/overheating/CfgWeapons.hpp
@@ -2,17 +2,6 @@ class CfgWeapons {
     class ACE_ItemCore;
     class InventoryItem_Base_F;
 
-    class ACE_SpareBarrel: ACE_ItemCore {
-        displayname = CSTRING(SpareBarrelName);
-        descriptionshort = CSTRING(SpareBarrelDescription);
-        //model = "";
-        picture = QPATHTOF(UI\spare_barrel_ca.paa);
-        scope = 2;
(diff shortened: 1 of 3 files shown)


esteldunedain:
The problem was introduced with this automated commit by @SilentSpike : 

108ff4f64406fd4c370a2fc146843434dabc57a8

I'm afraid the regexp might have produced other similar bugs, for example in ace_reload, as mentioned in https://github.com/acemod/ACE3/issues/4044#issuecomment-230166695

esteldunedain:
I did a quick review of that commit and couldn't find any other instance of the bug. The problem was caused by that line being the only call to targetEvent with two distinctive targets, which confused the regexp.

:+1: for this fix

PR #4047, opened 2016-07-03
Refuel - Simplify Actions, Fix Van_01 config

PabstMirror:
- Move connect to main actions as it was the only one in the tree.
- Tweak fuel nozzle interaction pos
- Fix Van_01 config
- Added debuger code to find vehicles with transportFuel

Tag @GitHawk

### DIFF
--- a/addons/refuel/CfgVehicles.hpp
+++ b/addons/refuel/CfgVehicles.hpp
@@ -52,7 +52,7 @@
 #define MACRO_CONNECT_ACTIONS \
     class ACE_Actions { \
         class ACE_MainActions { \
-            class GVAR(Refuel_Connect) { \
+            class GVAR(Connect) { \
                 displayName = CSTRING(Connect); \
                 distance = REFUEL_ACTION_DISTANCE; \
                 condition = QUOTE([ARR_2(_player,_target)] call FUNC(canConnectNozzle)); \
@@ -311,18 +311,16 @@ class CfgVehicles {
         GVAR(fuelCapacity) = 80;
     };
 
-    class Van_01_fuel_base_F: Van_01_base_F {};
-    class C_Van_01_fuel_F: Van_01_fuel_base_F {
-        transportFuel = 0; //1k
+    class Van_01_fuel_base_F: Van_01_base_F {
         MACRO_REFUEL_ACTIONS
         GVAR(hooks)[] = {{0.38,-3.17,-.7},{-0.41,-3.17,-.7}};
         GVAR(fuelCargo) = 2000;
     };
+    class C_Van_01_fuel_F: Van_01_fuel_base_F {
+        transportFuel = 0; //1k
+    };
     class I_G_Van_01_fuel_F: Van_01_fuel_base_F {
         transportFuel = 0; //1k
-        MACRO_REFUEL_ACTIONS
-        GVAR(hooks)[] = {{0.38,-3.17,-.7},{-0.41,-3.17,-.7}};
-        GVAR(fuelCargo) = 2000;
     };
     
     class Tank_F: Tank {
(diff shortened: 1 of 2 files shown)


addons/refuel/CfgVehicles.hpp
```diff
@@ -52,21 +52,13 @@
 #define MACRO_CONNECT_ACTIONS \
     class ACE_Actions { \
         class ACE_MainActions { \
-            class GVAR(Refuel) { \
-                displayName = CSTRING(Refuel); \
+            class GVAR(Refuel_Connect) { \
```
jonpas: This expands to `ace_refuel_Refuel_Connect`, seems redundant.

commy2: `GVAR(Connect)` ???

jonpas: Yeah, for example.

addons/refuel/CfgVehicles.hpp
```diff
@@ -318,13 +311,20 @@ class CfgVehicles {
...
         GVAR(hooks)[] = {{0.38,-3.17,-.7},{-0.41,-3.17,-.7}};
         GVAR(fuelCargo) = 2000;
     };
-
+    class I_G_Van_01_fuel_F: Van_01_fuel_base_F {
+        transportFuel = 0; //1k
+        MACRO_REFUEL_ACTIONS
+        GVAR(hooks)[] = {{0.38,-3.17,-.7},{-0.41,-3.17,-.7}};
+        GVAR(fuelCargo) = 2000;
+    };
```
commy2: What's the point of changing the class here? New we have two vehicles (retextures) with the same config entries.

PabstMirror: Vanilla adds transportFuel to `I_G_Van_01_fuel_F` and `C_Van_01_fuel_F` - not the base class

commy2: krgr, but our entries should still remain in the base class imo.

jonpas: Why? Not everything inheriting from it is a refuel truck.

commy2: jonpas, everything that inherits from `Van_01_fuel_base_F` is a fuel truck.
Only `C_Van_01_fuel_F` and `I_G_Van_01_fuel_F` do.
If BI adds the `transportFuel` properties to each one instead of the common base class, then we have to adjust. That does not mean that we have to imitate their strange practices.

addons/refuel/XEH_postInit.sqf
```diff
@@ -9,3 +9,16 @@ if (isServer) then {
 [QGVAR(resetLocal), {
     _this call FUNC(resetLocal);
 }] call CBA_fnc_addEventHandler;
+
+
+#ifdef DEBUG_MODE_FULL
+diag_log text format ["[ACE-refuel] Showing CfgVehicles with vanilla transportFuel"];
+private _fuelTrucks = configProperties [configFile >> "CfgVehicles", "(isClass _x) && {(getNumber (_x >> 'transportFuel')) > 0}", true];
```
commy2: Just a FYI im dropping here, because I made the suggestion to use `configProperties` the last time.
You could also use `"getNumber (_x >> 'transportFuel') > 0" configClasses (configFile >> "CfgVehicles")`
here and it would find everything while being faster.
This because `CfgVehicles` itself has no parents. Therefore it's pointless to use the 3rd parameter flag. All it would do is to additionally search through the parents of `CfgVehicles`. There are none though.

addons/refuel/CfgVehicles.hpp
jonpas: Ah, I thought not everything is a fuel truck from it.

PR #4051, opened 2016-07-04
Make ace news fit apex menu style

### LINKED ISSUE #3926: Update main menu additions for new Apex menu

PabstMirror:
Close  #3926
New clean and simple ace news that fits into apex style.
image

Will need some minor tweaks to verison.html to make it look like screenshot.


jonpas:
That's exactly how I imagined it! Great work! :+1:

addons/optionsmenu/init_loadMainMenuBox.sqf
```diff
@@ -15,17 +15,18 @@
...
-    ((_this select 0) displayCtrl 80086) ctrlShow true;
+    ((_this select 0) displayCtrl 80090) ctrlShow true;
 
-    private ["_version"];
-    _version = getText (configFile >> "CfgPatches" >> "ace_main" >> "versionStr");
-    ((_this select 0) displayCtrl 68) ctrlSetText format ["Installed Version: %1", _version];
-    ((_this select 0) displayCtrl 69) htmlLoad "http://ace3mod.com/version.html";
+    private _ace3VersionStr = (getText (configFile >> "CfgPatches" >> "ace_main" >> "versionStr")) select [0,5];
+    ((_this select 0) displayCtrl 80094) ctrlSetText format ["Version: %1", _ace3VersionStr];
+    ((_this select 0) displayCtrl 80095) htmlLoad "http://ace3mod.com/version.html";
```
esteldunedain: Should we switch to a different URL to maintain compat with non-apex or legacy versions of arma?

jonpas: ACE3 itself is not compatible with older versions, and this is not exclusive to Apex owners.

PabstMirror: Older version of ace loading new version.html would just see text aligned to the right instead of left, it should be fine.

esteldunedain: Older vesions of ACE are compatible with older versions of Arma.

bux578:
I'd even go further and do the styling in the version.html file (if possible).

bux578:
I'll try to tinker of what we can achieve with just changes in the html file. And what styling is allowed in it.

PabstMirror:
The html load is very basic, the html tag for align does seem to be respected which I think is all we need.

In configs I put `align = "right";` but that didn't seem to do anything.

bux578:
I'm wondering if even Icons and other images could be part of the html file. No "styling" (apart from placing the container) in config but all in the html file.

PR #4052, opened 2016-07-04
Mine detector optimization and cleanup

### LINKED ISSUE #4049: Enabling Mine Detector lags game

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Using the Mine Detector lags the game if multiple objects close by

- Dedicated

**RPT log file:**
- N/A atm

esteldunedain:
- Cache detected mines for 0.15 s
- Call detectorLoop every 0.05 s instead of 0.01s
- Compile a dictionary of detectable classnames from config at start
- Add the ability to switch between headphones or speakers
- Remove dummy object
- Cache getDetectorConfig for better interaction menu performance
- Change sound to mono wss
- Reduce sound duration to avoid it stacking with itself when playing through headphones. Sounds played with `say` or `say3D` wait for the previous "line of dialog" ends before playing. The minimum delay between beeps should have been is 0.5s, but it was actually controled by the sound duration, of almost 0.7s.
- Improve the positioning of the detector point.
- Move detector actions to their own submenu.
- Add clicks when activating/deactivating the detector
- Add an event when a mine is detected to allow hooking into missions
- Make the detector AI compatible
- Close #4049


bux578:
@esteldunedain What do you mean by "Switch to playSound3d for global effect"?

esteldunedain:
This module use say3D to produce the beep, which has only local effect. Now the beep can also be heard by nearby units.

bux578:
> Now the beep can also be heard by nearby units.

I don't think this is wanted. IIRC most would use headphones (especially in CQB environments)

esteldunedain:
> I don't think this is wanted. IIRC most would use headphones (especially in CQB environments)

Oh, well, maybe you are right. I can change it if that's the case.

I don't think the former behaviour was intentional though, because the code took the trouble of creating a 3d source for the sound, which doesn't make sense if we are simulating headphones.

bux578:
Maybe at some point we could allow both variants using the interaction menu. Have it audible to everyone or just yourself.

esteldunedain:
> Maybe at some point we could allow both variants using the interaction menu. Have it audible to everyone or just yourself.

Yeah, I think I'll go for that.

I'll have to change the code quite a bit again though, because there's no local equivalent to playSound3d, which requires a sound file instead of a CfgSound. Let me think a little more about how to best achieve that.

bux578:
Add an `ace_soundFilePath` property to the CfgSound?

esteldunedain:
> Add an ace_soundFilePath property to the CfgSound?

Yeah, something like that. I can also take the path to the file directly from CfgSound, but I'd need to remove the leading "\". It's quite doable too.

addons/minedetector/functions/fnc_playDetectorSound.sqf
```diff
@@ -27,6 +26,9 @@ if (!alive _unit) exitWith {
     ACE_LOGERROR_1("unit is not alive [%1]",_unit);
 };
 
-private _posASL = AGLtoASL (_unit modelToWorld (_unit selectionPosition "granat"));
-
-playSound3D [_sound, objNull, false, _posASL, 5, _pitch, 15];
+if (_unit getVariable [QGVAR(isUsingHeadphones), false]) then {
+    _unit say2D _soundClass;
```
Glowbal: Doesn't this queue the sounds and play lip animations?

bux578: https://community.bistudio.com/wiki/say2D

esteldunedain: Yeah, you are right. I'll try switching to `playSound` instead

Glowbal: That may do the same. At least the queuing. It's why I used a dummy object for the sound.

esteldunedain: Ok, I'll be sure to test if queuing happens or not with playSound, I think it doens't, cause we already use it on many places in ACE without problems (like combat deafness).

Glowbal: Ah good point. It seems I was mistaking in that case. PlaySound is probably the better solution compared with the dummy object approach.

addons/minedetector/functions/fnc_detectorLoop.sqf
```diff
@@ -35,10 +35,13 @@ if !([_unit, _type] call FUNC(isDetectorEnabled)) exitWith {
 if (ACE_player != _unit || {currentWeapon _unit != _type}) exitWith {};
 
 private _detected = [[_unit, _detectorConfig], FUNC(getDetectedObject), _unit, QGVAR(detectedObjects), 0.15] call EFUNC(common,cachedCall);
-_detected params ["_hasDetected", "", "_distance"];
+_detected params ["_hasDetected", "_mine", "_distance"];
 
 if (!_hasDetected) exitWith {};
 
+// Launch a local event stating which mine was detected for mission purposes
+[QGVAR(mineDetected), [_unit, _mine, _distance]] call CBA_fnc_localEvent;
```
bux578: Good idea!

bux578:
@esteldunedain How often is the "mine detected" event being raised for the same mine?

esteldunedain:
> @esteldunedain How often is the "mine detected" event being raised for the same mine?

With each beep. If the user wants a single event per mine it can be rather easily programmed mission side. I'd rather not add that overhead to the main code instead.

PabstMirror:
Looks good just 2 things to think about
- If we are adding sub-actions, we might want to move the detector actopms to the base level instead of under equipement as some people have trouble with deep action trees. It only shows when you have the weapon selected so it really shouldn't get in the way unless you are planning on using it, in which case moving it up would make it easier to use.

- Headphones seemed quieter than global sound, maybe boost the `CfgSounds` a little?

jonpas:
I'd say move the action up.

esteldunedain:
> If we are adding sub-actions, we might want to move the detector actions to the base level instead of under equipement as some people have trouble with deep action trees. 

To be honest this is not deeper than e.g. "bandage my left leg". If our IM can't handle that properly we should fix the IM. I'm not against moving the action up either, but I'd love this to be an excuse to improve the IM if it's needed.

> Headphones seemed quieter than global sound, maybe boost the CfgSounds a little?

True. Will do.

bux578:
It's working really well. Good job @esteldunedain !

PR #4056, opened 2016-07-06
Fix order of revive loop

### LINKED ISSUE #4035: Unconscious unit without pulse has to be treated twice with PAK

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- When a unit is knocked unconscious and has no pulse, after bandaging the wounds, the medic has to treat the patient twice with the PAK. The first treatment resets the wounds and the patient comes to very briefly, before being knocked out again. After the second treatment the patient is fine again.

We had the issue when playing normally on our server, then to investigate he issue further we removed any mods besides CBA and ACE (we did however not remove the compat.pbos we use) and set up a fresh test mission.

**Steps to reproduce:**
- Place three units on the map (the medic, a bystander to assist with CPR and the patient)
- Place medical modules for advanced treatment
- Shoot the patient
- Let the medic treat the wounds, while the other guy does CPR
- Treat the patient with the PAK

- Dedicated

**Placed Modules:**
- Medical Settings
- Advanced Medical Settings
- Revive Settings

**RPT log file:**
- https://gist.github.com/l4ndl0rd/432756959900e1f0ed3f0c252ce35584

PabstMirror:
Fix #4035
Revive loop was always killing heartrate before it exited, even if they were healed by PAK

### DIFF
--- a/addons/medical/functions/fnc_reviveStateLoop.sqf
+++ b/addons/medical/functions/fnc_reviveStateLoop.sqf
@@ -21,13 +21,6 @@ if (!local _unit) exitWith {};
 
 private _startTime = _unit getVariable [QGVAR(reviveStartTime), 0];
 
-// Remove heartbeat
-if (GVAR(level) >= 2) then {
-    if (_unit getVariable [QGVAR(heartRate), 60] > 0) then {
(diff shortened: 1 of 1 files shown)


Glowbal:
:+1: Thanks @PabstMirror

PR #4062, opened 2016-07-07
Add reload mutex as canInteractWith condition

### LINKED ISSUE #3938: Finger, AdvancedThrowing and Wipe Goggles ignore the reload mutex causing animation issues
**ACE3 Version:** `3.5.1` (eedb406ae5605c2dd5663fd0b2de2bb9bf501b6b with advancedThrowing branch)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- If you do any action mid-reloading that will trigger an animation, it will freeze the weapon animation and will not return a mag that animation took from the gun at the start of the reloading. Addons that do not check if unit is reloading so far I found advanced throwing and finger pointing. Weapon switching actually has the same problem as vanilla arma, weapon animation freezes, but the unit animation continues and it returns the mag in the end so it's not a huge deal.
Video.
https://www.youtube.com/watch?v=WwmbtaGV05I
First reload, without interupting.
Second, pointing. Interrupts reload animation, doesn't return the mag reload took at the start.
Third, grenade throw. Same.

**Steps to reproduce:**
- Pick something with a long reload animation.
- Point forward mid-reloading.
- Throw a grenade mid-reloading.

**Placed Modules:**
- Pointing Module.

**RPT log file:**
- N/A.

commy2:
- adds !GVAR(isReloading) as condition for canInteract
- fixes #3938


esteldunedain:
Hey, hold on. `GVAR(isReloading)` doesn't exist yet. It's currently called `GVAR(ReloadMutex)` and it is defined only on the `gestures` module. IMO it should be moved to common and renamed.

Also I wonder if it would be possible to set the variable from an `AnimChanged` EH instead of `actionKeys` for better reliability and joystick support.

commy2:
`AnimChanged` or any other anim event handler don't trigger for gestures like the reload animations.

esteldunedain:
> AnimChanged or any other anim event handler don't trigger for gestures like the reload animations.

Too bad; I was half expecting that reply though.

commy2:
^ this is my version of at least supporting the modifier keys. Not finished, because:

```
_a = -1207959552; (_a + 1) in [_a]
-> true
```
FML

jonpas:
Note: still needs additional canInteract checks on keybinds in all affected components.

commy2:
No, it's a canInteractWith condition.

jonpas:
Yes, whatever it's named. :P

esteldunedain:
> Yes, whatever it's named. :P

Nah, @jonpas . @commy2 means that this works automatically, without having to add any extra checks

commy2:
I think it's impossible to work with these offsets:
`(-1207959552 + 32) + 1207959552`
`0`

esteldunedain:
It's a precision issue; the numbers are being treated as floating point

commy2:
Yeah, but that means that actionKeys cannot be used for this - at least not when we want to support modifier keys.

esteldunedain:
>  at least not when we want to support modifier keys.

Let's not support it. I doubt anyone uses one for reloading

jonpas:
> Nah, @jonpas . @commy2 means that this works automatically, without having to add any extra checks

Ah, in that case - awesome!

commy2:
I don't like how easy it is to trick the reload mutex.

Pressing R in a reloading incompatible stance starts the timer and when then actually doing the reloading, the isReloading value is reset mid reload.

esteldunedain:
Yeah, it's pretty unreliable stuff. 

> incompatible stance

Those can be detected from config I guess, right?

PabstMirror:
Current interaction conditions take args `[_player, _target]`,
so you could do something like check if another player can interact with you.

This new condition would only be valid for ACE_player, 
but I really don't think this is a problem just wanted to mention it.

commy2:
Same with the old one. In the end it's just a hacky work around. : /

CorruptedHeart:
I assume this event is fired too late? (https://community.bistudio.com/wiki/Arma_3:_Event_Handlers#Reloaded)[Reloaded] and can't be used in any way?

commy2:
Reloaded is fired when the animation ended. too late...

I even thought about SoundPlayed event handler, but the only noise the player makes that it can't pick up is reloading it seems ...

commy2:
Should be good now.

Question: Do we want this as general canInteractWith condition, or would we be better of by if we add it to Finger / Advanced Throwing / Wipe Goggles directly?

jonpas:
I'd say general is fine.

esteldunedain:
@commy2, does this trigger when reloading vehicles?

commy2:
It triggers whenever you press the reload key while being in the mission display.

jonpas:
Merge @commy2 ?

PR #4064, opened 2016-07-07
Medical rewrite

thojkooi:
### Rationale

The medical component is one of our most used and obvious features. At the moment, the size and complexity of this component is large and it is quite an obstacle for other devs and contributors to contribute to medical. 

There are also a few issues with medical it self, where we have some underlaying issues and high maintance code, as well as some long outstanding issues (revive and prevent instant death being the obvious ones). 

This rewrite aims to address both these topics.
### Focus

The rewrite focus on the following:
- Split up the medical component into seperate smaller logical components.
- Combine the basic and advanced medical systems into one
- Remove any inconsistencies within the code
- A lot of the medical code has not been updated when BI introduced new script commands. The rewrite will address this as well.
#### Component diagram

pasted image at 2016_06_30 15_18
#### State Machine

The new medical rewrite will apply the medical state diagram for handling unit vitals and other tasks.

pasted image at 2016_06_30 14_54
### Backwards compatability

I will try to maintain backwards compatability as much as possible, but because I am redesigning a lot of the parts within medical, this will not be 100% possible. Once the rewrite is done, we should list all functions that have changed.

Settings will also change, but this will probably work out well if we switch to the CBA settings framework anyway.


sargken:
So there will be no more advanced medical?

bux578:
Discussions on this should be happening on Slack.

commy2:
rebase?

Glowbal:
That is needed yes. I am moving this up to 3.8, since I do not think I will be able to finish this for 3.7.0 without delaying the release further.

commy2:
- [ ] update extension to use `"Head", "Body", "LeftArm", "RightArm", "LeftLeg", "RightLeg"` instead of default BI selection names.

addons/medical/functions/fnc_handleStateDefault.sqf
```diff
@@ -22,5 +22,5 @@ systemChat format["handling state default for unit: %1", _this];
 
 private _pain = _unit getVariable [QGVAR(pain), 0];
 if (_pain > (_unit getVariable [QGVAR(painSuppress), 0])) then {
-    [_unit, _pain] call FUNC(playInjuredSound);
+    [_unit, _pain] call EFUNC(medical_engine,playInjuredSound);
```
commy2: syntax changed for those. reminder for myself to adjust these.

commy2:
Some internal functions

`EFUNC(medical_engine,damageBodyPart)`
- Sets the blood material visuals on or off.
- If used on "Arms" it will add an amount of weapon sway

```
 * 0: Unit <OBJECT>
 * 1: Selection, can be "Head", "Body", "Arms" or "Legs" <STRING>
 * 2: Damage (optional, default: true) <BOOLEAN>
```
___________

`EFUNC(medical_engine,setLimping)`
- Makes unit limp.
- Will turn on blood material visuals on "Legs"

```
 * 0: Unit <OBJECT>
 * 1: In pain (optional, default: true) <BOOLEAN>
```
___________

`EFUNC(medical_engine,setPainSway)`
- Adds set amount of weapon sway

```
 * 0: Unit <OBJECT>
 * 1: Limping (optional, default: true) <BOOLEAN>
```

commy2:
commy notes:

- [ ] handle blood textures
- [ ] sway status handler for fatigue compat

- [ ] uncomment AI, pls keep disabled for now
- [ ] custom killed event

bux578:
#4421

SilentSpike:
Instead of `enableInstantDeath` can we not implement a `minimumCardiacArrestTime`  to go along with the existing maximum?

That way instant death can be enabled if set to 0, but a minimum bleedout time can be set otherwise if desired.

SilentSpike:
So, I'm confused by something:

`(CBA_missionTime - _startTime) > (GVAR(maximumCardiacArrestTime) * 60)`

If this is the condition for death in cardiac arrest, surely it should be named `minimumCardiacArrestTime`? (I ran into this while in the process of implementing a setting of that name to replace `enableInstantDeath`)

SilentSpike:
Ah I think I see what the system is now. It **is** a minimum, but only because there is no other way to die while in cardiac arrest (currently). Cardiac arrest simply acts as an intermediate stage between fatal injury and death **only if** instant death is disabled.

My alternative proposition:

- Cardiac arrest should be an integral part of the system (always active) that acts as an intermediate stage between fatal injury and death.
- There should be a setting for both minimum and maximum time, so that if desired instead death can be enabled when `timeMin` is set to 0 (or forced when `timeMax` is set to 0).
- While in cardiac arrest, between `timeMin` and `timeMax` there should be some chance of death (based on external factors?)
- Before `timeMin` there's a 0% chance of death
- After `timeMax` there's a 100% chance of death

SilentSpike:
So looking at the component diagram here, `medical_ui` implements the display information (currently still in `medical`). Would it not make more sense for that to be in `medical_menu`?

SilentSpike:
Here's the new component structure:
image

`medical` itself is yet to be determined, but will probably require a bunch of these other components if we turn it into a proper public API.

bux:
😱

PR #4065, opened 2016-07-07
Remove redundant getMagazineIndex compatibility function

jonpas:
- Title
- Function definition was added in https://github.com/acemod/ACE3/commit/62f598b2a6bca3049438c52154e3b9c612aac952 (and later renamed in another commit) to keep compatibility with CBA at the time.


esteldunedain:
Perfect, thank you @jonpas

Issue #4066, opened 2016-07-07
Problem when trying to release a unit or place explosive. 


gbblaster:
**Arma 3 Version:** `1.60` (stable)
**CBA Version:** `2.4.1` (stable)
**ACE3 Version:** `3.6.0` (stable )

**Mods:**
           @rhs_gref |   
           @fallujah |   
           @sthud_a3 |   
        @AliveServer |   
              @ALiVE |   
@CUP_Weapons_ACE_compat |
               @Ares |   
                @mcc |   
            @A3CR1_7 |   
               @TFAR |   
            @RHSUSAF |   
            @RHSAFRF |   
               @MELB |   
             @LOPFOR |   
        @CUP_Weapons |   
       @CUP_Terrains |   
            @ASDG_JR |   
            @ASR_AI3 |   
               @ace3 |   
             @CBA_A3 |   

**Description:**
- So we've been having a problem and I want to check if it is something with our mod combo or a known issue as I haven't found anything on older threads. 

First problem is sometimes when you drag a unit the release function doesn't work, if the mouse is clicked the weapon fires. 

Second problem is, when placing explosives though the ACE menu the same thing happens instead of placing the device, the weapon fires. 

**Steps to reproduce:**
- It is kind of random. It just happens to some and not to others.

**Where did the issue occur?**
- Dedicated

**Placed Modules:**
- All ACE related are on the map. 

**RPT log file:**
- I'll post an RPT if needed.

esteldunedain:
RPTs are badly needed for this one. You need also try to repro with only CBA and ACE first.

PabstMirror:
`removeAllActions player` is one possible cause of this

You might want to try removing mods until it doesn't happen

alganthe:
10 :banana: it's due to MCC

gbblaster:
I've updated with all mods running in the mission. And here I've attached an RPT from a mission we ran last night. I will run without the other mods and try to recreate ASAP. 

[arma3server_2016-07-06_21-59-35.txt](https://github.com/acemod/ACE3/files/352879/arma3server_2016-07-06_21-59-35.txt)

commy2:
Try it in a new mission with only CBA and ACE enabled.

gbblaster:
I will, tonight. And upload the RPT. Thanks.

gbblaster:
[...]

PR #4067, opened 2016-07-07
Only use addActionEventHandler when needed

PabstMirror:
- Only install addActionEventHandler when the interaction menu is
actually open
- Less SQF running all the time (action's conditional was checked each frame, even if menu closed)
- it should help with issue #4066


esteldunedain:
I think we only need to add the EH if `GVAR(useCursorMenu) == false`

esteldunedain:
`ace_common_fnc_addActionEventHandler` has a fairly complex compile in there. It's probably not that bad to run it every time the IM is opened, but it might increase the lag spike slightly.  Theoretically that compiled code could be cached and reused, but I'm not sure if it's worth it.

commy2:
I don't like this method. We shouldn't change our ways just because a random mission removes all actions. :-1:

PabstMirror:
The compile is 0.0292 ms, so shouldn't be a problem.

I had been wanting to do this for a while for the performance improvements, the fact that this also fixes another mod's compatibility problem is just a side bonus.

esteldunedain:
> The compile is 0.0292 ms, so shouldn't be a problem.

Perfect, never mind then

PabstMirror:
Not critical for 3.6.1, will bump to 3.7

PR #4070, opened 2016-07-09
Japanese Translate on ACE3 v3.5.x

classicarma:
this is not support v3.6.0


commy2:
We need a script that replaces all tabs with 4 spaces.
Don't click on "Files changed" if you have a slow browser.

jonpas:
Replaced by updated #4473

PR #4076, opened 2016-07-10
convert interaction to cba settings

commy2:


esteldunedain:
I'd like to use this opportunity to change some settings from booleans to drop boxes mainly for settings that switch between different "modes" e.g.: 

Replace:
- Display interaction menus as lists (Yes / No)
By something like:
- Menu distribution: (Radial / List)

Replace:
- Always display cursor for self interaction (Yes / No)
By something like:
- Self interaction mode: (Free look / Cursor) 

Replace:
- Do action when releasing menu key (Yes / No)
By something like:
- Execute actions on: (Click / Release)

I think we should also discuss changing some of the default settings for the IM. From what I gathered from the dev group, most seem to use cursor mode for all interactions. From my own experience I'd go for:

- Interaction mode: (Cursor) 
- Self Interaction mode: (Cursor) 
- Menu distribution: (List)
- Do actions on: (Click)

I feel that our current default is probably the hardest combination to use.

Issue #4020, opened 2016-06-28
CfgPatches "requiredAddons" APEX


commy2:
Reminder to update requiredAddons in main after Tanoa.

kwinteng:
Does that mean we will have to have APEX in order to play with ACE3?

commy2:
3.6.1 is planned to only work with 1.62
That doesn't mean that you have to buy APEX though.

kwinteng:
Okay, thank you for your quick response commy

PabstMirror:
Weird thing to note, steam doesn't seem to download tanoa pbos unless you own the expansion.

`A3_Map_Data_Exp,A3_Map_Tanoabuka, A3_Map_Tanoabuka_Data, A3_Map_Tanoabuka_Data_Layers, A3_Map_Tanoa_Scenes_F`

These CfgPatches didn't exists before. So obviously we don't want to put these in requiredAddons.

commy2:
```
"A3_3DEN",
"A3_3DEN_Language",
"A3_Air_F",
"A3_Air_F_Beta",
"A3_Air_F_Beta_Heli_Attack_01",
"A3_Air_F_Beta_Heli_Attack_02",
"A3_Air_F_Beta_Heli_Transport_01",
"A3_Air_F_Beta_Heli_Transport_02",
"A3_Air_F_Beta_Parachute_01",
"A3_Air_F_Beta_Parachute_02",
"A3_Air_F_EPB",
"A3_Air_F_EPB_Heli_Light_03",
"A3_Air_F_EPC",
"A3_Air_F_EPC_Plane_CAS_01",
"A3_Air_F_EPC_Plane_CAS_02",
"A3_Air_F_EPC_Plane_Fighter_03",
"A3_Air_F_Exp",
"A3_Air_F_Exp_Heli_Light_01",
"A3_Air_F_Exp_Heli_Transport_01",
"A3_Air_F_Exp_Plane_Civil_01",
"A3_Air_F_Exp_UAV_03",
"A3_Air_F_Exp_UAV_04",
"A3_Air_F_Exp_VTOL_01",
"A3_Air_F_Exp_VTOL_02",
"A3_Air_F_Gamma",
"A3_Air_F_Gamma_Plane_Fighter_03",
"A3_Air_F_Gamma_UAV_01",
"A3_Air_F_Gamma_UAV_02",
"A3_Air_F_Heli",
"A3_Air_F_Heli_Heli_Attack_01",
[...]

PR #4083, opened 2016-07-11
Required addons for APEX

commy2:


jonpas:
We are waiting for Apex release to verify this right?

commy2:
exactly

PabstMirror:
Will close #4020

Glowbal:
Works fine for me. Note that I own the Apex expansion. We just need to confirm this works fine for someone without the expansion.

PR #4084, opened 2016-07-11
Add ACE javelin as required addon dependency

Glowbal:
Required because of 

```
modelOptics = "\z\ace\addons\javelin\data\reticle_titan.p3d";
```

### DIFF
--- a/optionals/compat_rhs_usf3/config.cpp
+++ b/optionals/compat_rhs_usf3/config.cpp
@@ -6,7 +6,7 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"rhsusf_c_weapons", "rhsusf_c_troops", "rhsusf_c_m1a1", "rhsusf_c_m1a2", "RHS_US_A2_AirImport", "rhsusf_c_m109", "rhsusf_c_hmmwv", "rhsusf_c_rg33", "rhsusf_c_fmtv", "rhsusf_c_m113", "RHS_US_A2Port_Armor"};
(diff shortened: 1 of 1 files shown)


bux578:
does that mean, that people who remove ace_javelin can't use the compat pbo anymore?

Glowbal:
Yes, but that should already the case with current releases.

jonpas:
I disagree, that shouldn't be the case. That compat pbo has functionality from other things than ace_javelin as well.

If `modelOptics` is used even when ace_javelin is not loaded then it should be completely separate compat pbo, if not there is no reason to add dependency.

esteldunedain:
> If modelOptics is used even when ace_javelin is not loaded then it should be completely separate compat pbo, if not there is no reason to add dependency.

A some point we could do multiple compat pbos. For the moment I suggest instead that we either duplicate the p3d in the compat pbo or that we move it to a sepparate "resources" pbo inside ACE

commy2:
Or move the p3d to common.

esteldunedain:
Or javelin_data

bux578:
> duplicate the p3d in the compat pbo

That should be the way to go. Their Javelin optic might look different and someone working on the compat could use a different p3d to reflect that.

Glowbal:
Sure, but not for 3.6.1. We will fix that for 3.7.0.

jonpas:
That breaks backwards compatibility for people using this compat but not using `ace_javelin`, I guess there aren't many of those but point stays.

PabstMirror:
In the compat/CfgWeapons we are doing `canLock = 0`, so it's more than just the missing p3d that would be the problem

Glowbal:
@jonpas no it doesn't. Our previous releases have `ace_javelin` in their required addons list. Courtesy of pboProject. We are not breaking any backwards compatability.

Here is the full list of required addons from 3.6.0:
```sqf
  requiredAddons[] = {"rhsusf_c_weapons","rhsusf_c_troops","rhsusf_c_m1a1","rhsusf_c_m1a2","RHS_US_A2_AirImport","rhsusf_c_m109","rhsusf_c_hmmwv","rhsusf_c_rg33","rhsusf_c_fmtv","rhsusf_c_m113","RHS_US_A2Port_Armor","ace_javelin","A3_UI_F","ace_goggles"};
```

esteldunedain:
Yay, nice to know pboProject has our back

jonpas:
Ah, no problem then! Can we please :armake:... :angry:

PR #4089, opened 2016-07-12
Improve SSN hash

PabstMirror:
Passes @bux578 's splendid test code: https://codeshare.io/jorDJ

Uses all chars of long names, and remixes input so that "A" does not always turn into 5

### DIFF
--- a/addons/dogtags/functions/fnc_ssn.sqf
+++ b/addons/dogtags/functions/fnc_ssn.sqf
@@ -8,24 +8,27 @@
  * Return Value:
  * A random three/two/four format social security number <STRING>
  *
+ * Example:
+ * ["AAA"] call ace_dogtags_fnc_ssn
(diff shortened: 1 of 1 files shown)


bux578:
:relaxed:

jonpas:
3.6.2?

esteldunedain:
Yay! @bux578 , nice list of names

> 3.6.2?

Nah, there's really no need to redo this PR.

Glowbal:
This is for dog tags, which is not in the current release but in 3.7.0.

jonpas:
Oh right :facepunch: (there is no picard emoji).

SilentSpike:
Very elegant :+1: An easy decision to merge!

bux578:
So, one could say this was the first test driven development we did? :smile:

Issue #4088, opened 2016-07-12
Vector 21 rangefinder issue


Hailstorm12:
**Arma 3 Version:** `1.62.137494` (stable)
**CBA Version:** `2.4.1` (stable)
**ACE3 Version:** `3.6.1` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`
- `@aceX`

**Description:**
Vector 21 range finder and azimuth no longer work after apex update. 

**Steps to reproduce:**
- Add the steps needed to reproduce the issue.

**Where did the issue occur?**
- Editor (Singleplayer) 

**Placed Modules:**
- None

**RPT log file:**
- Add a link ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the client and/or server RPT file. An instruction to find your RPT files can be found [here](https://community.bistudio.com/wiki/Crash_Files#Arma_3).
- If possible at the time the bug is encountered, go to ACE Options and select "Debug To Clipboard", this will print extensive debug information to the RPT file.

https://gist.github.com/Hailstorm12/63d348910b6acae25c2980b11af3cc3c#file-gistfile1-txt

jonpas:
Fixed by #4096

Issue #4094, opened 2016-07-12
Unable to link belts or check ammo


Hailstorm12:
**Arma 3 Version:** `1.62.137494` (stable)
**CBA Version:** `2.4.1` (stable)
**ACE3 Version:** `3.6.1` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`
- `@aceX`

**Description:**
Unable to link belts or check ammo. 

**Steps to reproduce:**
Place down assistant AR (Player) and an AR (AI) with a few unarmed csat guys because it used to be if you hadn't shot any bullets, you couldn't link belts. AI spawns in and kills csat, then try to check ammo/link belt with AI unit (you may have to leave group so he doesn't run away from you)

**Where did the issue occur?**
- Editor (Singleplayer) 

**Placed Modules:**
- None

**RPT log file:**
- Add a link ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the client and/or server RPT file. An instruction to find your RPT files can be found [here](https://community.bistudio.com/wiki/Crash_Files#Arma_3).
- If possible at the time the bug is encountered, go to ACE Options and select "Debug To Clipboard", this will print extensive debug information to the RPT file.

https://gist.github.com/Hailstorm12/5f9036516dfd8cf4539f05c5b0dec644#file-gistfile1-txt

PR #4096, opened 2016-07-12
use a function to do gestures

commy2:
- creates `doGesture` so we can apply the reload mutex to every gesture
- removes the `canInteractWith` condition for `!isReloading`

### DIFF
--- a/addons/common/XEH_PREP.hpp
+++ b/addons/common/XEH_PREP.hpp
@@ -32,6 +32,7 @@ PREP(displayText);
 PREP(displayTextPicture);
 PREP(displayTextStructured);
 PREP(doAnimation);
+PREP(doGesture);
 PREP(dropBackpack);
 PREP(endRadioTransmission);
 PREP(eraseCache);
--- a/addons/common/XEH_PREP.hpp
+++ b/addons/common/XEH_PREP.hpp
@@ -1,4 +1,5 @@
 
+PREP(actionKeysNamesConverted);
 PREP(addCanInteractWithCondition);
 PREP(addLineToDebugDraw);
 PREP(addSetting);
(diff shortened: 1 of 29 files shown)


commy2:
Should fix:
https://github.com/acemod/ACE3/issues/4088
https://github.com/acemod/ACE3/issues/4094

esteldunedain:
How is #4094 related to this?

commy2:
Shortcut for `R` blocks the actions of these. I think that is why he can't check the ammo.

esteldunedain:
> Shortcut for R blocks the actions of these. I think that is why he can't check the ammo.

Nope, those actions are done via the Interaction menu.

jonpas:
@esteldunedain Check ammo is also a keybind. Not sure about link belts.

esteldunedain:
The ticket is about checking the ammo of the AR, not your own. I dont think
we have a keybind for that

commy2:
I striked it off. Was misreading.

commy2:
will fix #4101

commy2:
^
I made a function to read vanilla keybinds

```
actionKeysNamesArray "ReloadMagazine"
["Alt links+R","Shift rechts+R","Strg rechts+R","R++","R","++R","+"]

"ReloadMagazine" call ace_common_fnc_actionKeysNamesConverted
[[19,false,false,true],[19,true,false,false],[19,false,true,false],[27,false,false,false],[19,false,false,false],[19,false,false,false],[27,false,false,false]]
```
(Note that for double keybinds, e.g. "C+R" only the last key is used - those can't be displayed with the keyDown format)

commy2:
^ CheckAmmo and UnjamWeapon use this parameter to not require `!isReloading`
The parameter should only be used for keybinds which overwrite vanilla actions via returning `true`

PR #4107, opened 2016-07-13
Fix MainMenu Ace_Options Placement to top left

PabstMirror:
Stick ace_options button placement to top left
so it doesn't interfere with apex singleplayer menu


jonpas:
Can be done later, but we were discussing of placing it under the vanilla "Options" menu, as a drop-down item. Would fit quite nicely in the main menu. However, if we'll be switching to CBA Settings it's probably not worth it.

esteldunedain:
>  However, if we'll be switching to CBA Settings it's probably not worth it.

Exactly

PR #4109, opened 2016-07-13
Port ACE2 incendiary grenades

commy2:
- adds incendiary grenades from ACE2
- also fixes m84 rvmat formating
- also fixes completely wrong ai ammo usage flags for our grenades
- some small fixes / cleanup in grenades
- adds all grenades to CfgPatches

#### AN-M14 TH3
- thermite based incendiary grenade used by the US military
- grenade is best used when dropped due to low radius of effect
- 3 seconds fuse
- burns for ~40 seconds
- destroys nearby (static) weapons, ammunition, crates and other eqippment
- lights fireplaces and barrels
- SFX include initial explosion, burning thermite, fire, smoke, light and a burning sound


addons/grenades/CfgMagazines.hpp
```diff
@@ -65,6 +65,18 @@ class CfgMagazines {
         mass = 4;
     };
 
+    class SmokeShell;
+    class ACE_Incendiary: SmokeShell {
+        author = ECSTRING(common,ACETeam);
+        displayname = CSTRING(Incendiary_Name);
+        descriptionshort = CSTRING(Incendiary_Description);
+        displayNameShort = "AN-M14";
```
bux578: stringtable

esteldunedain:
This grenade can be used IRL to permanently disable light vehicles by melting through their engines. It can allegedly burn through 1/2" armor. 

I think allowing something like that in game would be interesting. 

Maybe applying scripted damage if there's a vehicle below the grenade.

jonpas:
That would also work perfectly with advanced throwing, the only problem I can imagine is vehicle collision being wonky and grenades simply bouncing/sliding off of thin air.

esteldunedain:
For this type of usage the explosive placement system could be more appropiate than the throwing mode. IDK if implementing such a thing is feasable though.

addons/grenades/CfgAmmo.hpp
```diff
@@ -124,4 +124,26 @@ class CfgAmmo {
         soundImpactWoodExt3[] = {"A3\sounds_f\characters\footsteps\wood_run_ext_2",1,1.1,75};
         soundImpactWoodExt4[] = {"A3\sounds_f\characters\footsteps\wood_run_ext_2",1,1.15,75};
     };
+
+    class ACE_G_Incendiary: SmokeShell {
```
esteldunedain: For consistency with the flashbang this could be renamed `ACE_G_M14`. I'm not sure if it's worth it though.

commy2: Tbh, I don't like these military designations as display names. It's confusing.
Best example is `XM307` vs `XM312`. I can never tell which one is which. (we should probably change the names of those)

esteldunedain: I agree; we are calling the flashbang ACE_G_M84 though

commy2:
Damage is already scripted for consistency (particle fire damage acts strange sometimes).
I planned to trigger the engine burning feature from cookoff if the grenade is placed on a vehicle.

esteldunedain:
> I planned to trigger the engine burning feature from cookoff if the grenade is placed on a vehicle.

Sounds good

addons/grenades/CfgAmmo.hpp
commy2: It should've been `ACE_G_Flashbang`

bux578: What if we add a different Flashbang?
`ACE_G_Flashbang2` how can you then tell the difference?

bux578:
> is placed on a vehicle

anywhere, or on a specific location?

addons/grenades/CfgAmmo.hpp
commy2: `ACE_G_Flashbang2` is a bad name.
Name would depend on the situation. If it's a Soviet stun grenade then maybe: `ACE_G_Flashbang_Opfor`. If it's a `DM109` then mybe `ACE_G_Flashbang_Germany`

commy2:
IDK yet. No plans atm.

commy2:
Done. Grenade has to be placed on (near) engine.

Ready for merge

kerckasha:
these are also used to destroy weapon caches or static weapons, would that work in its current configuration?

commy2:
Yes.

Glowbal:
This PR contains ammo cook off.

commy2:
done
I also renamed it to `ACE_M14` for consistency after all.
And I added code which makes the car engine catch fire if #4116 is present.

addons/grenades/functions/fnc_incendiary.sqf
```diff
@@ -0,0 +1,173 @@
...
+            deleteVehicle _x;
+        };
+
+        // --- inflame fireplace, barrels etc.
+        _x inflame true;
+    };
+} forEach (_position nearObjects EFFECT_SIZE);
+
+// --- burn car engine
+if ("ace_cookoff" call EFUNC(common,isModLoaded)) then {
```
Glowbal: ace_cookoff should receive an enabled toggle. I imaging that will influence this as well. Unless we want to play the fire regardless of that setting?

commy2:
Now I always destroy the engine and only the effect requires cookoff to be present and to be enabled.

commy2:
Setting already renamed for #4240

PR #4125, opened 2016-07-16
Per-Ammo fired events

esteldunedain:
- Add framework to support per-ammo fired events, inspired on ACE 2 registerSimulation.
- Leverage this framework in a lot of modules (WIP)
## Rationale

In various cases we need to run special code when certain types of projectiles are fired, like the huntir, different types of grenades, missiles, heavy caliber bullets, cannon shells, explosives, etc. The way we are currently doing that is by stacking handlers to the fired EH and checking if every single projectile fired matches each criteria. The pattern is O(N) with the number of custom scripts. 

This is currently not that bad because we don't have a huge N and the unified fired event handler filters out AI from most of the checks. We've also been clever enough to attach each FEH to the minimum amount of classess we need, etc. But still, this pattern clearly doesn't scale very well if we want to keep adding a bunch of custom stuff.

ACE 2 had a framework for this called `ACE_Simulation`. That was a config entry on each ammo type which indicated which code should be called when the bullet was fired. ACE 2 had a much larger number of processes, so that made sense.  But it had the problem of only allowing a single process to be simulated per ammo type.

This PR proposes a similar framework for ACE3, but with the added flexibility of allowing any number of simulations to be carried out for each ammo. Also the criteria to determine is a process should be called is now defined via a script (instead of a fixed config entry), allowing for greater flexibility. Obviously the code is only run on mission start and then the events that should be called for each ammo are cached for later usage from the UFEH. This potentially comes at the expense of slightly increased memory usage and mission start times.

As I said, I think we don't really need this ATM. However it should provide some gain already (untested yet) and can serve as a preparation for the future.

### DIFF
--- a/addons/common/functions/fnc_registerAmmoFiredEvent.sqf
+++ b/addons/common/functions/fnc_registerAmmoFiredEvent.sqf
@@ -1,0 +1,54 @@
+/*
+ * Author: esteldunedain
+ * Register an ammo fired event that should be launched when ammo following a
+ * given criteria is fired. The criteria is analyzed on mission start and is
+ * cached.
+ *
+ * Arguments:
+ * 0: Name of the event <STRING>
+ * 1: Condition code <CODE>
+ * 2: Call for the local player on foot? <BOOL>
+ * 3: Call for non local players on foot? <BOOL>
+ * 4: Call for non players on foot? <BOOL>
+ * 5: Call for the local player vehicle? <BOOL>
+ * 6: Call for non local players vehicles? <BOOL>
+ * 7: Call for non players vehicles? <BOOL>
+ *
+ * Return Value:
+ * None
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_eventName", "_condition", "_fireForPlayer", "_fireForPlayerNonLocal", "_fireForNonPlayer", "_fireForPlayerVehicle", "_fireForPlayerVehicleNonLocal", "_fireForNonPlayerVehicle"];
+
+private _fnc_addEvent = {
+    params ["_namespace", "_condition"];
+    if (_condition) then {
+        private _events = _namespace getVariable _ammo;
+        if (isNil "_events") then {
+            _events = [];
+            _namespace setVariable [_ammo, _events];
+        };
+        _events pushBack _eventName;
+    };
+};
+
+private _cfgAmmo = configFile >> "CfgAmmo";
+for "_i" from 0 to (count _cfgAmmo) - 1 do {
+    private _ammoConfig = _cfgAmmo select _i;
+    private _ammo = configName _ammoConfig;
+
+    // The condition code receives the following as parameter:
+    // 0: Name of the ammo <STRING>
+    // 1: Config of the ammo <CONFIG>
+    if ([_ammo, _ammoConfig] call _condition) then {
+        [GVAR(ammoFiredEvents_ForPlayer), _fireForPlayer] call _fnc_addEvent;
+        [GVAR(ammoFiredEvents_ForPlayerNonLocal), _fireForPlayerNonLocal] call _fnc_addEvent;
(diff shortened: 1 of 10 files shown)


esteldunedain:
Various modules in ACE should probably be converted to use this eventually.

addons/common/functions/fnc_registerAmmoFiredEvent.sqf
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Arguments:
+ * 0: Name of the event <STRING>
+ * 1: Condition code <CODE>
+ * 2: Call for the local player on foot? <BOOL>
+ * 3: Call for non local players on foot? <BOOL>
+ * 4: Call for non players on foot? <BOOL>
+ * 5: Call for the local player vehicle? <BOOL>
+ * 6: Call for non local players vehicles? <BOOL>
+ * 7: Call for non players vehicles? <BOOL>
```
commy2: I'm a bit confused by these. "Call for non players on foot?". Does that mean: "fire for all AIs, local or not"?
What would you use if you wanted to execute it for all local soldiers, inlcuding AI?
Does "on foot" mean that it does not trigger for FFV slots? Those didn't exist in A2, so it might be an oversight.

Glowbal: I think it would be more descriptive if we ditch the negatives in the description. ie `non local players` -> `remote players`

esteldunedain: > "Call for non players on foot?". Does that mean: "fire for all AIs, local or not"?

Yes.

> Does "on foot" mean that it does not trigger for FFV slots? 

I'm actually not sure about that one.  The check is if firedEH unit is of type CAManBase, which I think should be FFV compatible. 

-------

You should keep in mind that this is just trying to mimic the Unified Fire EH logic of dividing events in six categories:
- ace_firedPlayer
- ace_firedPlayerNonLocal
- ace_firedNonPlayer
- ace_firedPlayerVehicle
- ace_firedPlayerVehicleNonLocal
- ace_firedNonPlayerVehicle

commy2: Maybe change the description to:
```
+ * Arguments:
+ * 0: Name of the event <STRING>
+ * 1: Condition code <CODE>
+ * 2: Call for local player? <BOOL>
+ * 3: Call for remote players? <BOOL>
+ * 4: Call for (local ???) AI soldiers? <BOOL>
```
The rest is still ambiguous. What is a non-local player vehicle? What if the player is the commander?

esteldunedain: > What is a non-local player vehicle? What if the player is the commander?

Actually what matters for remote vehicles is if a player is manning the turret or not. If the vehicle is in "manual fire" actually effective commander is checked, etc.

The names are a bit ambiguous, I know. But the underlying functionality is tried and true.

I think replacing "player on foot" by "player soldier" or "player unit", etc should clear some of the confusion.

Don't forget that the purpose of all these is that most things we do for shots we only do for players. But sometimes each player handles their own stuff only; and sometimes each client handles all players locally.

commy2: >Actually what matters for remote vehicles is if a player is manning the turret or not. If the vehicle is in "manual fire" actually effective commander is checked, etc.

That makes sense.

So what mask would you use to execute a fired eh on all local units, but not remote units?

esteldunedain: There's actually no combination specific to that (and I think we actually don't need that anywhere though) But anywhere you'd probably either

A- Don't use UFEH
B- Suscribe to ace_firedPlayer and ace_firedNonPlayer (but not ace_firedPlayerNonLocal) and check if local.

commy2:
Stale. I also believe that this is something for CBA... In A4.

PR #4130, opened 2016-07-16
Fix terrain LOS check for vehicles with low getPos

### LINKED ISSUE #4126: Javlin with RHS stopped working
**ACE3 Version:** `3.6.2` (stable / dev + commit hash)

**Mods:**
- `@CBA_A3`
- `@ace`
- `@RHS Escalation`
**Description:**
- Rocket shooted with ACE compatibility mode from javelin never reach its target.
**Steps to reproduce:**
- Use RHS javelin together with ACE compat. files.

- Everywhere

**Placed Modules:**
- NON
[arma3_2016-07-16_17-30-44.zip](https://github.com/acemod/ACE3/files/367358/arma3_2016-07-16_17-30-44.zip)

PabstMirror:
Fix #4126

RHS T72 `getPosASL` point is very low and can be bellow terrain level.

`ASLtoAGL getPosASL cursorObject` = [2457.5,820.846,-0.0508118]

So it would fail the terrainIntersectASL check.

### DIFF
--- a/addons/missileguidance/functions/fnc_checkLos.sqf
+++ b/addons/missileguidance/functions/fnc_checkLos.sqf
@@ -5,24 +5,24 @@
  * Arguments:
  * 0: Seeker [Object]
  * 1: Target [Object]
- * 
+ *
  * Return Value:
  *    Boolean
  */
 #include "script_component.hpp"
-private["_seeker", "_seekerPos", "_target", "_targetPos", "_return", "_vectorTo", "_searchPos"];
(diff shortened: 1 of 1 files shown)


commy2:
>RHS T72 getPosASL point is very low and can be bellow terrain level.

what?

esteldunedain:
image

Dystopian:
Not only RHS vehicles are bad. Vanilla SUV/Offroad have the same value. And many vehicles with damaged wheels.

PR #4132, opened 2016-07-18
Port of ACE2 casings

esteldunedain:
- Port of ACE2 persistent "casing" aka "cartridges" aka "staying brass"
## Notes
- Based on #4125. 
- I'm very impressed at how little performance impact having a huge amount of these objects produce. 200 seems like a very conservative number.
- The amount of objects could very well be a client setting (low, medium, high, insane). I'd prefer to wait for the new setting system for that though.
- I've tried a potentially better technique to generate the shells, but I couldn't get the XEH to fire on the casing particles the engine creates when firing, even when they are listed under CfgVehicles. @commy2, that's what I was asking about earlier today. 
## Todo
- [x] Figure out why ACE2 included custom models for typical casings (5.56, 7.62, etc). These models are already present in Arma 3, so I'm using the vanilla models. Maybe ACE2's have lower polycount, or maybe Arma 2 didn't have them, IDK.


esteldunedain:
By the way. I considered using `createObject` for this, but I think it'd be actually have worse performance:
- `createObject` is global, so it produces some network traffic when creating/deleting
- According to [this](https://forums.bistudio.com/topic/187670-new-command-createobject-for-decorative-objects/) doesn't really improve performance of "dumb" object like walls. This bullet casing inherit from `NonStrategic` and are as dumb as they get.

PabstMirror:
Vanilla cartridges have this config:
```
    class FxCartridge: ThingEffect {
        timeToLive = 5;
        disappearAtContact = 1;
    };
```
Could we just patch the config with a longer TTL?

esteldunedain:
> Vanilla cartridges have this config:

No @PabstMirror, I'm not using those classes that inherit from ThingEffect; I'm only reusing the p3ds, but inheriting from NonStrategic

esteldunedain:
> Could we just patch the config with a longer TTL?

Patching the TTL is possible, but the whole point of this modules is having a bunch of ammo to look at after a firefight, but never too much so that it doesn't slow the game. TTL doens't allow that, because you don't have any control about how many casing are actually present at any given time.

addons/casings/CfgVehicles.hpp
```diff
@@ -0,0 +1,67 @@
+class CfgVehicles {
+    class NonStrategic;
+
+    class FxCartridge_stay: NonStrategic {
```
commy2: needs ACE tag / GVAR

esteldunedain:
Heads up people:

This code requires ~ 1.00ms
```sqf 
_casing = _casingType createVehicleLocal _posATL;
_casing setposATL _posATL;
```
While this one requires ~ 0.08ms
```sqf 
_casing = _casingType createVehicleLocal [0,0,0];
_casing setposATL _posATL;
```

That's actually said [here](https://community.bistudio.com/wiki/Code_Optimisation#createVehicle.28Local.29), but I didn't quite believed it until I tried.

PabstMirror:
It does sorta make sense as `createVehicle` must do some kind of `findEmptyPosition` logic because you can create multiple cars at the same position and they usually won't collide,

[0,0,0] might be hard-coded to skip those checks or more likely that position is just free of objects on most maps.

esteldunedain:
> It does sorta make sense as createVehicle must do some kind of findEmptyPosition logic because you can create multiple cars at the same position and they usually won't collide,

That was exactly my guess. I think the findEmptyPosition only happens if the model has a geometry LOD, which the vanilla casings have. ACE2's on the other hand don't.

BTW, I measured marginally better performance with ACE2 casings vs vanilla, so I'm inclined to use them.

My only doubt is if we have permission to release those models by Rocko with our license (or APL, or some other one). Any thoughts on this @Glowbal, @bux578, @NouberNou??

jonpas:
> My only doubt is if we have permission to release those models by Rocko with our license (or APL, or some other one).

From what Glowbal said last time, we are allowed to use any of the ACE2 models.

esteldunedain:
Ok, thank you

addons/casings/XEH_postInit.sqf
```diff
@@ -0,0 +1,56 @@
...
+            false
+        };
+
+        private _casing = "ACE_casings_" + _cartridge;
+        if !(isClass (configFile >> "CfgVehicles" >> _casing)) exitWith {
+            false
+        };
+
+        private _casingModel = getText (configFile >> "CfgVehicles" >> cartridge >> "model");
+        if (_casingModel in ["\A3\weapons_f\empty", ""]) exitWith {
```
jokoho48: in is case-sensitive so maybe this should be toLower?

addons/casings/data/5_56.rvmat
```diff
@@ -0,0 +1,81 @@
...
+diffuse[] = {1.0, 1.0, 1.0, 1.0};
+forcedDiffuse[] = {0.0, 0.0, 0.0, 0.0};
+emmisive[] = {0.0, 0.0, 0.0, 1.0};
+specular[] = {0.6, 0.377931, 0.18, 1.0};
+specularPower = 100.0;
+PixelShaderID = "Super";
+VertexShaderID = "Super";
+
+class Stage1 {
+	texture = "z\ace\addons\casings\data\5_56_nohq.paa";
```
jonpas: This file has tabs.

addons/casings/data/shell.rvmat
```diff
@@ -0,0 +1,81 @@
...
+forcedDiffuse[] = {0.0, 0.0, 0.0, 0.0};
+emmisive[] = {0.0, 0.0, 0.0, 1.0};
+specular[] = {0.6, 0.377931, 0.18, 1.0};
+specularPower = 100.0;
+PixelShaderID = "Super";
+VertexShaderID = "Super";
+
+class Stage1 {
+    texture="z\ace\addons\casings\data\shell_nohq.paa";
+	uvSource = "tex";
```
jonpas: This file also has tabs.

jonpas:
Models should be named `ace_<name>.p3d`.

jokoho48:
and we should remove the Rocko Tag

esteldunedain:
> Models should be named ace_<name>.p3d.

We actually only do on approx 50% of the models. Plus I think it's dumb, because there's zero reasons to prefix them

jonpas:
> We actually only do on approx 50% of the models. Plus I think it's dumb, because there's zero reasons to prefix them.

True, however there is a good reason to prefix them, model.cfg entry requires the class to be named the same as p3d file and that class can conflict with other classes (or something along the lines). Armake will print a [warning](https://github.com/KoffeinFlummi/armake/blob/f5e16a98fccfe6d4c876f2ba1add24dc8df67001/src/model_config.c#L298) for non-prefixed models.

esteldunedain:
Ok, my bad; there's a reason then. Thank you for the information @jonpas.

jonpas:
That's why we aee a team, right. :D

commy2:
```
It does sorta make sense as createVehicle must do some kind of findEmptyPosition logic because you can create multiple cars at the same position and they usually won't collide,
```

The problem is that there is no `createVehicleLocal <ARRAY>` version, where you can use the `CAN_COLLIDE` parameter. Using it disables the hard coded "findEmptyPosition" BS.

KoffeinFlummi:
Just now catching up with my emails. @jonpas is right, we found out about this when RHS's and CUP's Huey had the same name for their CfgModels class, causing all kinds of funky animations.

commy2:
Which is hilarious.

Glowbal:
I think this is good to be merged. Could maybe do with someone taking another look at the code, since it's quite a large one. 

Please merge after that (or ping me and I will merge it). :+1:

addons/casings/functions/fnc_firedEH.sqf
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+//IGNORE_PRIVATE_WARNING ["_unit", "_weapon", "_muzzle", "_mode", "_ammo", "_magazine", "_projectile", "_vehicle", "_gunner", "_turret"];
+
+BEGIN_COUNTER(fnc_fireEH);
+
+private _unitPosATL = getposATL _unit;
```
commy2: `getPosATL`

addons/casings/functions/fnc_firedEH.sqf
```diff
@@ -0,0 +1,67 @@
...
+        // By creating it at [0,0,0] instead of _posATL the time is reduced
+        // from around 1.08 ms to 0.08 ms. This is most likely because the
+        // engine doesn't create the object exactly where it is told to, but
+        // instead looks for a suitable position. Creating at origin prevents
+        // that from happening.
+        _casing = _casingType createVehicleLocal [0,0,0];
+    };
+
+    _casing setposATL _posATL;
+    _casing setdir (random 360);
```
commy2: `setPosATL` and `setDir`

addons/casings/CfgVehicles.hpp
```diff
@@ -0,0 +1,62 @@
...
+    class GVAR(FxCartridge_9mm): GVAR(FxCartridge_Small) {
+    };
+    class GVAR(FxCartridge_556): GVAR(FxCartridge) {
+        model = QPATHTO_M(models\ace_casing_556.p3d);
+    };
+    class GVAR(FxCartridge_65): GVAR(FxCartridge) {
+        model = QPATHTO_M(models\ace_casing_556.p3d);
+    };
+    class GVAR(FxCartridge_65_caseless): GVAR(FxCartridge) {
+        model = "\A3\weapons_f\empty";
```
commy2: does it make sense to create invisible casings? shouldn't there be a better option for this, e.g. not creating them at all.

esteldunedain: Empty casing are never created. They are filtered out on mission start here:

https://github.com/acemod/ACE3/pull/4132/files#diff-b725f1f4b77e69fddfd668ff98eeae86R49

esteldunedain:
@Glowbal, @commy2, keep in mind that merging this implies merging #4125. That should probably be discussed/merged first.

moonmanmove:
will this be in for 3.7.0?

commy2:
@moonmanmove probably not, as it requires  #4125 which hasn't been tested enough.

addons/casings/functions/fnc_firedEH.sqf
```diff
@@ -0,0 +1,67 @@
...
+#include "script_component.hpp"
+
+//IGNORE_PRIVATE_WARNING ["_unit", "_weapon", "_muzzle", "_mode", "_ammo", "_magazine", "_projectile", "_vehicle", "_gunner", "_turret"];
+
+BEGIN_COUNTER(fnc_fireEH);
+
+private _unitPosATL = getposATL _unit;
+
+// Make far away units 10 times less likely to create casings
+if ((positionCameraToWorld [0,0,0]) vectorDistance _unitPosATL > 100 && {random 1 < 0.9}) exitWith {};
```
ulteq: What's the rationale behind using `positionCameraToWorld` here?

dedmen: It get's the cameraPositon instead of the unit's head. 
This will also be the View Position when viewing through a UAV for example.

bux:
Status on this one?
@commy2

Cyruz143:
Since https://github.com/acemod/ACE3/pull/8857 is merged, can probably close this off.

PR #4140, opened 2016-07-21
Fix refuel not respecting time acceleration and pause

### LINKED ISSUE #4128: ACE Refueling continues while the game is paused
**ACE3 Version:** `3.6.2` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Refueling does not take game time into account. Refueling continues while the game is paused, and does not speed up if time acceleration is changed.

**Steps to reproduce:**
- Place down an empty Littlebird in the editor with no fuel
- Place down a refuel HEMTT next to it
- Connect the HEMTT fuel hose to the Littlebird

- Editor (Singleplayer)

**Placed Modules:**
- 'None'

jonpas:
- Fix #4128 
- The PFH runs constantly while the nozzle is connected, even when not refueling, thus old time has to be set in `turnOn` function.

### DIFF
--- a/addons/refuel/functions/fnc_refuel.sqf
+++ b/addons/refuel/functions/fnc_refuel.sqf
@@ -16,11 +16,9 @@
 
 #include "script_component.hpp"
 
-#define PFH_STEPSIZE 1
-
 params [["_unit", objNull, [objNull]], ["_target", objNull, [objNull]], ["_nozzle", objNull, [objNull]], ["_connectToPoint", [0,0,0], [[]], 3]];
 
-private _rate =  getNumber (configFile >> "CfgVehicles" >> (typeOf _target) >> QGVAR(flowRate)) * GVAR(rate) * PFH_STEPSIZE;
+private _rate =  getNumber (configFile >> "CfgVehicles" >> (typeOf _target) >> QGVAR(flowRate)) * GVAR(rate);
 private _maxFuel = getNumber (configFile >> "CfgVehicles" >> (typeOf _target) >> QGVAR(fuelCapacity));
 
 [{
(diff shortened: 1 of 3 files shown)


addons/refuel/functions/fnc_refuel.sqf
```diff
@@ -46,32 +44,39 @@ private _maxFuel = getNumber (configFile >> "CfgVehicles" >> (typeOf _target) >>
...
         private _fuelInSource = [_source] call FUNC(getFuel);
         if (_fuelInSource == 0) exitWith {
             [LSTRING(Hint_SourceEmpty), 2, _unit] call EFUNC(common,displayTextStructured);
+            _nozzle setVariable [QGVAR(lastTickMissionTime), nil, true];
             _nozzle setVariable [QGVAR(isRefueling), false, true];
         };
+
+        // Calculate rate using mission time to take time acceleration and pause into account
+        private _rateTime = _rate * (CBA_missionTime - (_nozzle getVariable [QGVAR(lastTickMissionTime), CBA_missionTime]));
+        _nozzle setVariable [QGVAR(lastTickMissionTime), CBA_missionTime, true];
```
esteldunedain: I think that setting this as public is overkill. If this is only set locally and locality changes, all that'd happen is that for a single tick the _rateTime would be zero, but it'd go back to normal on the next tick; hardly an issue.

The whole function is a bit public-heavy; but that's probably a matter for a different PR.

jonpas: Agreed.

jonpas:
Works fine in self-hosted multiplayer. I noticed a bug if the one who connected the nozzle disconnects then fueling won't happen at all until another player disconnects and connects it again, however that is unrelated to this PR and is a very edge-case.

Issue #4141, opened 2016-07-21
setMass 1, still can't lift


kdjay:
I am using setMass 1;

still cannot carry box, "Item too heavy"

according to these docs it should work.
http://ace3mod.com/wiki/framework/dragging-framework.html
wtf

kdjay:
I am calling setMass before calling ace_dragging_fnc_setDraggable;

kdjay:
debug
http://pastebin.com/BQNs92x0

PabstMirror:
You can check mass of box with:
`[cursorTarget] call ace_dragging_fnc_getweight;`

and then change the limits:
```
    ACE_maxWeightCarry = 600;
    ACE_maxWeightDrag = 800;
```

The weight of things inside of a box is the biggest source of weight.

esteldunedain:
So in short using ` setMass` doesn't affect the weight of the **content** of the box. If the box is too full then you won't be able to carry it despite the command.

PR #4143, opened 2016-07-22
Update dragging framework wiki

PabstMirror:
From issue #4141

- remove notes about setMass
- add notes about weight limit variables

### DIFF
--- a/wiki/framework/dragging-framework.md
+++ b/wiki/framework/dragging-framework.md
@@ -33,7 +33,7 @@ class CfgVehicles {
     <p>The following functions are NOT public and are likely to change in the future!</p>
 </div>
 
-You will **not** be able to carry / drag objects that are too heavy, the mass is also affected by what is inside the object. To bypass this empty the object and/or use `setMass`.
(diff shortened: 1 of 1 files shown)


wiki/framework/dragging-framework.md
```diff
@@ -33,7 +33,7 @@ class CfgVehicles {
     <p>The following functions are NOT public and are likely to change in the future!</p>
 </div>
 
-You will **not** be able to carry / drag objects that are too heavy, the mass is also affected by what is inside the object. To bypass this empty the object and/or use `setMass`.
+You will **not** be able to carry / drag objects that are too heavy, the mass is also affected by what is inside the object. To bypass this empty the object. You can change the weight limits by setting `ACE_maxWeightDrag` (default 800) and `ACE_maxWeightCarry` (default 600).
```
bux578: > by setting `ACE_maxWeightDrag` (default 800) and `ACE_maxWeightCarry` (default 600).

Is this a setting? If so, could you edit it to something like `by changing the following ACE3 settings `
(Do we have a module for this? http://ace3mod.com/wiki/missionmaker/modules.html)

PabstMirror: It's just a raw variable in missionNamespace

bux578: ah, ok. Maybe an example?

esteldunedain: > It's just a raw variable in missionNamespace

We should make it a setting when the new system is adopted

Issue #3696, opened 2016-04-14
TODO list for tagging


BaerMitUmlaut:
Ready for 3.6.0:
- [x] Align tags vertically in non-horizontal surfaces

For 3.7.0+:
- [x] Scripted API for custom tags, possibly like this #4146 
  ```
  ["Circle", "path\to\circle.paa"] call ace_tagging_fnc_createCustomTag
  ```
- [x] Custom tags by config documentation #4150 
- [x] Scripted API documentation #4150 
- [x] ~~Make checkTaggable public~~ not necessary

ColdEvul:
Reference from the [help chat at slack](https://ace3public.slack.com/archives/help/p1461172389000286)
add possibility to set rotation as a param so you don't need to write a entire system for custom tagging when text is involved. :)

http://images.akamai.steamusercontent.com/ugc/278471138304886827/7EB996E32A76021DDA074369568E9A2E031BA4BA/

PR #4146, opened 2016-07-22
Tagging Framework and Improvements

jonpas:
- Add config and script tagging framework - #3696 
- Add `ACE_Tags` class to define tags and `ace_tagging_fnc_addCustomTag` to add tags during mission
- Cache tags in config on mission start
- Add actions dynamically under new interaction sub-menu
- Improve performance due to structured checking
- Quick Tag action (main "Tag" action) and keybind (unbound) and settings (Disabled, Last Used - Default, Random X, Random)

~~There is one issue I currently have with adding tags with script during mission, I am not sure what the best way is to add the action to all units currently present, the ones that might be spawned later and keep everything MP and JIP compatible.~~


esteldunedain:
I don't think this is implemented  in the way it makes the most sense. 

This adds actions the actions individually to every single CAManBase created instead of adding them to the classes. Hence the actions are potentially added a lot more times than necessary. Object actions are also slower to evaluate in the interact menu than class actions.

jonpas:
Interaction menu doesn't add actions to inherited classes with `addActionToClass` though, another solution would be @PabstMirror https://github.com/acemod/ACE3/pull/3666, which would resolve that problem. However, there is still the problem of a good way to add actions via script as I described above.

esteldunedain:
> Interaction menu doesn't add actions to inherited classes with addActionToClass

Yes right. But what we do in other modules is keep the the init EH, but add the actions to the class the first time an object from that class is created and then never again. So you only add the action once per class instead of once per unit.

jonpas:
True, will change it to that. :+1: 

Any ideas about scripted adding? If not, I will just leave it out as I see no good way of doing it atm.

esteldunedain:
> Any ideas about scripted adding?

I think the best way to do that is to add actions dynamically; adding some tagging actions on unit creation and other later could get messy. A couple of comments about that:

- It's true that we'd need #3696 to place those actions under equipment, but we could place them inside a "Tag" submenu, either inside or outside "Equipment"
- If you really want to support an API, I think you are much better off just adding all tagging actions dynamically instead of at init.

PabstMirror:
Maybe we do want #3666
Then we can dynamically add children to the equipment submenu

jonpas:
Then all there would be to do is track actions added later in an array which gets broadcasted on every tag addition and children gets compiled from data in that array, since children actions are added dynamically.

esteldunedain:
Yeah, I meant #3666.  With or without it, I think we are better of putting those actions dynamically. 

I'd prefer them under a submenu (even if it's outside equipment), because I feel it's more organized (tagging can contain a lot of actions!) and will also perform better (because you can put a condition in the parent action that checks if the unit has spray paint).

jonpas:
> because you can put a condition in the parent action that checks if the unit has spray paint

True, however there would have to be additional checks for each spray paint from CfgWeapons, as people can add more colors if they wish. Simplest would be to just check all `requiredItem` entries from `ACE_Tags` and check that on main action and then obviously the corresponding one for each tag. I think that's also most flexible.

jonpas:
This should be much better now, will also make it super simple to add tags during mission, just add it to cached array on all clients (`CBA_fnc_globalEventJIP`) - this is more of a self-note.

jonpas:
Added "spray random tag" on main interaction point usage, I think it's quite nifty when you need something sprayed quickly and don't care with what or when you only have one spray you don't have to go into the sub-menu.

Finished and tested (including MP).

addons/tagging/XEH_postInit.sqf
```diff
@@ -39,6 +39,62 @@ for "_index" from 0 to (_countOptions - 1) do {
     };
 };
 
+if (hasInterface) then {
```
esteldunedain: I think this code is now large enough to be moved to a dedicated function.

esteldunedain:
> spray random tag

I think it makes much more sense to always either:
- just spray an X
- or repeat the last tag.

We are now giving mission makers the possibility of adding custom tags because tags may have specific meanings, and IMO it doesn't make sense for the player to risk tagging something they didn't mean.

BTW, if ease of use is paramount I think it'd be better to add a keybind for that instead of an aditional action, placed somewhere in the menu it doesn't belong.

TL;DR My suggestion is adding keybind to "Repeat last tag" instead

jonpas:
Agreed, do you reckon it would be too confusing to have both the keybind and the main interaction action? I personally don't like more keybinds, although I can see main action thing could be confusing.

addons/tagging/XEH_postInit.sqf
jonpas: 9b4f565cc56c6a08495705de4540e4a1d16e3fd0

esteldunedain:
> do you reckon it would be too confusing to have both the keybind and the main interaction action?

I don't think that'd confusing; but personally I'd prefer that we didn't add the quick action. It's just a opinion though, and not a strong one by any means though.

jonpas:
Added Quick Tag action (main "Tag" action) and keybind (unbound) and settings (Disabled, Last Used - Default, Random X, Random).

jonpas:
@esteldunedain what do you think of current functionality?

BTW, this PR is considered finished from my side. :tada:

esteldunedain:
:+1:

Go merge!

PR #4149, opened 2016-07-23
Fix Cargo Framework title numbering

jonpas:
- Title - there were 2 titles in a row numbered `2.`


bux578:
Good catch. Thanks :)

PR #4152, opened 2016-07-23
Add Tripflare

### LINKED ISSUE #394: Trip Flare
Saw someone's comment on the ACE3 thread:
http://forums.bistudio.com/showthread.php?190433-ACE3-A-collaborative-merger-between-AGM-CSE-and-ACE&p=2914409&viewfull=1#post2914409

Trip flares would be a neat little idea if we get the time :)

PabstMirror:
From @VKing6 
Close #394


addons/explosives/scripts/TripflareEffect.sqf
```diff
@@ -0,0 +1,20 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * beforeDestroyScript = "\z\ace\addons\explosives\scripts\TripflareEffect.sqf;
+ *
+ * Public: no
+ */
+
+//This needs to be a script file, so it will be re-compiled each use.
```
jonpas: Is there a reason it has to be recompiled on each use? Add the reason to comments maybe.

PR #4163, opened 2016-07-26
Add tool to upload ace and compats to workshop

PabstMirror:
This should lets us automate the publishing of ace to steam workshop.
It will let us upload each compatibility to separate mods

I believe next RHS will be available on steamworkshop, 
so this will let users get ACE, RHS and ACE's RHS compat all through steam.

This will take some final adjustments like actually creating the new mods on the workshop

### DIFF
--- a/tools/publish.py
+++ b/tools/publish.py
@@ -1,10 +1,9 @@
 #!/usr/bin/env python3
 
-#Author: PabstMirror
+# Author: PabstMirror
 
-#Uploads ace relases to workshop
-#Will slice up compats to their own folders
-
+# Uploads ace relases to workshop
+# Will slice up compats to their own folders
 
 import sys
 
@@ -62,7 +61,10 @@ def buildCompatFolder(folderName, copyFileNames):
     return compatRelease_dir
 
 def publishFolder(folder,modID,changeNotes):
-    cmd = [publisherTool_path, "update", "/id:{}".format(modID), "/changeNote:'{}'".format(changeNotes), "/path:{}".format(folder)]
+    cmd = [publisherTool_path, "update", "/id:{}".format(modID), "/changeNoteFile:{}".format(changeNotes), "/path:{}".format(folder)]
+
+    print ("running: {}".format(cmd))
+
     print("")
     print("Publishing folder {} to workshop ID {}".format(folder,modID))
     print("")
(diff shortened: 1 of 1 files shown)


jonpas:
If we are going to do each compat it's own repo then we don't need to handle anything special with compats.

Glowbal:
Could you add a tiny bit of documentation for this? On this PR is ok until we figure out if we want to add this in our docs directory or some other place.

I'd like to start using this for the ACE releases from 3.7.0 and later.

I will take this from @ViperMaul, since he is busy at least the next week iirc. I should have some time to look further in this in the weekend.

tools/publish.py
```diff
@@ -0,0 +1,117 @@
...
+        raise Exception("Publisher","Publisher shit the bed")
+
+
+#GLOBALS
+release_dir = "P:\\z\\ace\\release"
+project = "@ace"
+publisherTool_path = find_bi_tools()
+ace_release_dir = os.path.join(release_dir, project)
+ace_optionals_dir = os.path.join(ace_release_dir, "optionals")
+do_publish = True #False will let you just build dirs and test without running publisher
```
jonpas: Can those global be moved to the top, just below imports? In other tools we also capitalize them, eg:
```
######## GLOBALS #########
RELEASE_DIR = "P:\\z\\ace\\release"
PROJECT = "@ace"
##########################
```

tools/publish.py
```diff
@@ -0,0 +1,117 @@
...
+    cmd = [publisherTool_path, "update", "/id:{}".format(modID), "/changeNote:'{}'".format(changeNotes), "/path:{}".format(folder)]
+    print("")
+    print("Publishing folder {} to workshop ID {}".format(folder,modID))
+    print("")
+    if (not do_publish):
+        print("Just doing test build")
+        return
+    ret = subprocess.call(cmd)
+    if ret != 0:
+        print("publisher faild with code {}".format(ret))
```
jonpas: failEd

esteldunedain:
> If we are going to do each compat it's own repo then we don't need to handle anything special with compats.

I'd be nice to keep the flexibility of being able to handle compats from the main repo too.

BTW, some optional PBOs could maybe be released as sepparate mods on SteamWorks too (e.g. ace_tracers, ace_noactionmenu).

jonpas:
> I'd be nice to keep the flexibility of being able to handle compats from the main repo too.

I was just thinking of our discussion that having it in separate repos would easier for others not familiar to ACE3 to contribute. But that would involve a very different tool set as well.

esteldunedain:
Even if the code for some compats pbos (or even ACEX) are located in sepparate repos, I think it'd probably be easier to NOT include tools on those repos. That would mostly avoid having N versions of the same files to keep up to date. 

Wouldn't it be better to have a master build & publish tool located on the main ACE repo?

jonpas:
Yes, I totally agree with that.

PR #4166, opened 2016-07-26
Make hearing component compatible with glasses

commy2:
- makes hearing component compatible with glasses


Glowbal:
I'm not entirely sure what this PR is doing ?

commy2:
This makes it so `CfgGlasses` can use `GVAR(protection)` the same way head gear can.
It was requested by someone in the public slack.

PR #4168, opened 2016-07-26
optional pbo to disable cross hair

commy2:
- Dunno if this is still needed. Someone made a request for it and I forgot about it. Found it while going through my unmerged branches.

### DIFF
--- a/addons/nocrosshair/$PBOPREFIX$
+++ b/addons/nocrosshair/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\ace\addons\nocrosshair
--- a/addons/nocrosshair/CfgInGameUI.hpp
+++ b/addons/nocrosshair/CfgInGameUI.hpp
@@ -1,0 +1,6 @@
+
+class CfgInGameUI {
+    class Cursor {
+        weapon = ""; // "\A3\ui_f\data\igui\cfg\cursors\weapon_ca.paa";
+    };
+};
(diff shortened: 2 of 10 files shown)


jonpas:
This control is hardcoded right? (no IDC/IDD)

commy2:
No IDC. It's engine behaviour. It does adjust for camera movement etc.

jonpas:
But isn't there a difficulty setting for crosshair specifically? Not sure if it's part of some other UI setting.

commy2:
I have no idea.

654wak654:
Crosshairs have their own difficulty setting, they always did.

esteldunedain:
> Crosshairs have their own difficulty setting, they always did.

Close?

jonpas:
In that case I think this PR is pointless (sorry @commy2 !).

PabstMirror:
It's a ACE_isOptional
So this allows a server to have crosshairs on, but users can disable them by adding this pbo.

commy2:
^ Yeah. I vaguely remember something about this. A shame that I can't remember who suggested this.
Also, a shame that the new settings stuff isn't made in SQF like 3den, so we can fix and tweak it where necessary.

Gwynblade:
Oh, I just saw this, it's me who requested it. When played with buddies that like crosshair on, we often had to compromise, sometimes off, sometimes on. With this, everybody can play how they like.

Thanks for making! :)

PR #4171, opened 2016-07-26
deprecate ACE hash functions in favor of CBA versions

commy2:
- Deprecates the hash functions in favour of the CBA ones.
- Part of / See: https://github.com/CBATeam/CBA_A3/issues/47
- I also nuked the hash list functions. I couldn't get them to work. They are never used, but produce script errors, even in their error handling

e.g.:
`-#define HANDLECATCH    diag_log text _exception; assert(exception=="")`

`exception` is undefined. Probably meant `_exception`. These conflict with our `ACE_LOGX macros`


PabstMirror:
`[] call CBA_fnc_hashCreate` = ["#CBA_HASH#",[],[],any]
`[] call ACE_common_fnc_hashCreate` = [[],[]]

We do some manual iteration through the old hashes with stuff like
`} count (GVAR(syncedEvents) select 0);`
which breaks with the changed functions.

I think our best bet is to switch our stuff to use CBA hashes and leave ace's hash functions alone but mark as deprecated.

addons/main/script_macros.hpp
```diff
@@ -93,17 +93,11 @@
...
-#define HASH_GET(hash, key)            ([hash, key, __FILE__, __LINE__] call EFUNC(common,hashGet))
-#define HASH_REM(hash, key)            ([hash, key, __FILE__, __LINE__] call EFUNC(common,hashRem))
-#define HASH_HASKEY(hash, key)        ([hash, key, __FILE__, __LINE__] call EFUNC(common,hashHasKey))
-
-#define HASHLIST_CREATELIST(keys)                ([keys] call EFUNC(common,hashListCreateList))
-#define HASHLIST_CREATEHASH(hashList)            ([hashList] call EFUNC(common,hashListCreateHash))
-#define HASHLIST_SELECT(hashList, index)        ([hashList, index, __FILE__, __LINE__] call EFUNC(common,hashListSelect))
-#define HASHLIST_SET(hashList, index, value)    ([hashList, index, value, __FILE__, __LINE__] call EFUNC(common,hashListSet))
-#define HASHLIST_PUSH(hashList, value)            ([hashList, value, __FILE__, __LINE__] call EFUNC(common,hashListPush))
+#define HASH_CREATE ([] call EFUNC(common,hashCreate))
```
commy2: wip, cba

commy2:
component that have to be converted to CBA hashes:

-[ ] syncedEvents
-[ ] laser
-[ ] map_gestures
-[x] finger

PR #4172, opened 2016-07-26
log file and line numbers for warnings/errors/debug msgs

commy2:
- Updates our log macros to also log file and line number

```
[ACE] (minedetector) ERROR: unit does not exist [<NULL-object>]
vs.
[ACE] (minedetector) ERROR: unit does not exist [<NULL-object>] File: z\ace\addons\minedetector\functions\fnc_playDetectorSound.sqf Line: 22
```

- except "INFO"


jonpas:
I am not sure if we want that in warning macro, as that is usually used simply to notify of some wrong configuration or similar

PR #4174, opened 2016-07-26
add component that can read BI handleDamage gibberish

commy2:
This is my attempt to contribute to our ongoing focus feature - ace medical.

The Arma vanilla health (and feedback) system is a PITA to work with. The code is ugly and convoluted and it's hard to tell which parts do what. Many things are hard coded, can only be changed by config or are defined all over the place. The absolutely worst is the mess that is called `handleDamage`.

My idea was to split off the BS into one component, `medical_engine`. This follows the reasoning behind #4064 - split off medical into sub components to make the whole system maintainable.

This component would do:
- disable vanilla heal actions
- disable hard coded behaviour of `FirstAidKit` and `Medikit` items
- disable vanilla bleeding texture overlay
- disable vanilla screams
- disable vanilla moans while injured
- instead, add a function to play these sounds manually (`FUNC(playInjuredSound)`)
- disable vanilla heartbeat while unconscious
- disable vanilla pp effects while being damaged
- add function to set structural damage (`FUNC(setStructuralDamage)`)
- add function to (visually) damage head, body, "hands" (arms) and legs, incluing increased weapon sway (arms) and limping (legs) (`FUNC(damageBodyPart)`)
- adds HitLeftArm, HitRightArm, HitLeftLeg, HitRightLeg to all units that need them
- do `handleDamage` magic: call event if local unit received a new wound.
- preload static weapons to fix that annoying bug

```
[QGVAR(woundReceived), {
    systemChat str _this;
    params ["_unit", "_woundedHitPoint", "_receivedDamage", "_shooter", "_ammo"];
```
`_ammo` will can be `#drowning`, `#falling` or `#collision` too. Damage by fire will be reported as `collision` damage, as I still haven't found a method for telling the difference.
For fall damage, a random leg will be selected.

The component will log units that don't have the necessary hitpoints (those who have invulnerable arms and legs in the current release)

The "new" parts are:
- disableing those vanilla injured moans
- disableing the heartbeat while being unconscious (which will allow us to use `setUnconscious`)
- disableing the pp effects (red vignette, blur, radial blur) while being damaged
- and obviously the nice `woundReceived` event

I woud be maintainer of this specific component.

!!please only comment in this PR if you know what you are talking about!!


addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -0,0 +1,105 @@
...
+
+// Add injury
+if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
+    // --- Head
+    private _damageFace = _unit getVariable [QGVAR(newDamage$HitFace), 0];
+    private _damageNeck = _unit getVariable [QGVAR(newDamage$HitNeck), 0];
+    private _damageHead = (_unit getVariable [QGVAR(newDamage$HitHead), 0]) max _damageFace max _damageNeck;
+
+    // --- Body
+    private _damageStructural = _unit getVariable [QGVAR(newDamage$#structural), 0];
```
esteldunedain: I wonder; should this "engine" module know about the different types of damage the "body" can have. Shouldn't that just be handled by the higher-level component? 

Maybe it should just be
```sqf
private _damageBody = (_unit getVariable [QGVAR(newDamage$HitBody), 0]);
```

commy2: This way you get low amounts of damage if a protected torso part was hit and high amounts if a unprotected part was hit.

commy2: >I wonder; should this "engine" module know about the different types of damage the "body" can have?

Like cuts vs bruises etc etc? Possibly.
possibly. I need to dig further into how we currently determine that.

esteldunedain: No that's no what  I meant. I thought you where hard coding stuff like HitDiaphragm yourself, but now I understand that's a hitpoint on the units themselves.

As for cuts vs bruises etc, I think you shouldn't add anything like that here; instead they should be managed by higher level modules.

commy2: Agreed.

commy2:
replaced

PR #4192, opened 2016-07-30
merge medical engine component into medical rewrite branch

commy2:
- integrate #4174 into #4064

@ todo

- [ ] remove things from other components that are replaced with things from this component


PabstMirror:
Shooting a guy in the leg with the MX shows as "body". I think it's because body includes `_damageStructural`.

commy2:
This should cover everything now. It doesn't work yet fully, but I think this is due `medical-rewrite` nott working fully either. I'd say we merge this and only work on one branch.

PR #4196, opened 2016-08-01
add BlueForceTracking filtering via an object variable

freghar:
add BlueForceTracking filtering via an object variable

There's no good workaround for this:

 * creating a dummy unit on [0,0,0] and making it leader prevents
   grp members from entering vehicles
 * joining group of a different side prevents grp members from seeing
   blueforce tracking markers of their original side

Usage (_obj can be group or unit, depending on BFT_ShowPlayerNames):

    _obj setVariable ["ACE_map_hideBlueForceMarker", true];

### DIFF
--- a/addons/map/functions/fnc_blueForceTrackingUpdate.sqf
+++ b/addons/map/functions/fnc_blueForceTrackingUpdate.sqf
@@ -27,7 +27,7 @@ if (GVAR(BFT_Enabled) and {(!isNil "ACE_player") and {alive ACE_player}}) then {
     };
 
     if (GVAR(BFT_ShowPlayerNames)) then {
-        _playersToDrawMarkers = allPlayers select {side _x == _playerSide};
+        _playersToDrawMarkers = allPlayers select {side _x == _playerSide && {!(_x getVariable [QGVAR(hideBlueForceMarker), false])}};
 
         {
             private _markerType = [_x] call EFUNC(common,getMarkerType);
@@ -47,6 +47,8 @@ if (GVAR(BFT_Enabled) and {(!isNil "ACE_player") and {alive ACE_player}}) then {
             } count units _x > 0;
         };
     };
+
+    _groupsToDrawMarkers = _groupsToDrawMarkers select {!(_x getVariable [QGVAR(hideBlueForceMarker), false])};
 
     {
         private _markerType = [_x] call EFUNC(common,getMarkerType);


bux578:
@esteldunedain me?

addons/map/functions/fnc_blueForceTrackingUpdate.sqf
```diff
@@ -27,7 +27,7 @@ if (GVAR(BFT_Enabled) and {(!isNil "ACE_player") and {alive ACE_player}}) then {
     };
 
     if (GVAR(BFT_ShowPlayerNames)) then {
-        _playersToDrawMarkers = allPlayers select {side _x == _playerSide};
+        _playersToDrawMarkers = allPlayers select {side _x == _playerSide && !(_x getVariable [QGVAR(hideBlueForceMarker), false])};
```
654wak654: Lazy eval?

freghar:
Rebased, added documentation in `README.md` in the style other addons use (ie. attach/captives for "Items Added").

@esteldunedain Can you please reassign this to one of the people listed in [map/Maintainers](https://github.com/acemod/ACE3/blob/master/addons/map/README.md) as the comment above would seem to indicate that the current Assignee is wrong? Thanks!

addons/map/README.md
```diff
@@ -8,6 +8,10 @@ Various tweaks to the in-game map, including:
 - Map illumination (optional)
 - Blue Force Tracker (optional)
 
+#### Variables
```
esteldunedain: We don't include documentation like this in Readme file. Please commit it to the documentation branch instead (`gh-pages`).  The ace3mod.com web site is auto generated from the content of that branch,

jonpas: We do have that in some places, I think it doesn't hurt.

esteldunedain: Ok, but line 12 is missing the leading space.

freghar: Can you please explain what do you mean? The markdown doesn't care up to 0-3 leading spaces, so it's down to code style and I followed the syntax of the rest of the file (the other lists [start without a space](https://raw.githubusercontent.com/freghar/ACE3/f930f58732a8a/addons/map/README.md)).

If you mean it should have a leading space because it's in the Variables subsection, then it's not right either - the Markdown syntax and indeed many other wiki syntaxes define levels without indent, which exists only in lists (beyond the first level).

If you don't want it there, I can remove it, though I don't think such a minor feature deserves a separate web page and a similar syntax [is used in ACE3 already](https://github.com/acemod/ACE3/blob/master/addons/attach/README.md).

Thanks.

esteldunedain: If I'm not mistaken line 12 has zero leading spaces and line 13 has one; at least that's how GitHub is showing it. Obviously it's a very minor thing, but it should be fixed if it's not a huge inconvenience.

jonpas: It's a matter of readability to others editing the file, and for consistency. Doesn't matter if there is space or there isn't, but the 2 lines are different currently.

> If you don't want it there, I can remove it, though I don't think such a minor feature deserves a separate web page and a similar syntax is used in ACE3 already.

Everything _should_ be on the wiki. This variable API should be added to [Mission tools](http://ace3mod.com/wiki/missionmaker/mission-tools.html) page.

freghar: Ah, right, line 13 has 2 leading spaces to create a new list level, to serve as a comment for line 12. See the [rich diff](https://github.com/acemod/ACE3/pull/4196/files/f930f58732a8a9f8c612ce61c8efca5decfd3654?short_path=8c8b8b6#diff-8c8b8b6450fc239d39d46113e6fa9862).

I'll remove the note from the README though, it probably shouldn't serve as a dumping ground for _possibly_ interesting hints.

Glowbal:
:+1: Merge after the stuff with the readme has been resolved.

freghar:
Rebased, `README.md` change removed.

Issue #3537, opened 2016-03-05
TODO createVehicle -> createSimpleObject


BaerMitUmlaut:
`createSimpleObject` saves on performance for non-simulated objects and could probably used for a few modules.
- [ ] Tagging (tags)
- [x] Medical (litter)

PabstMirror:
Ref https://forums.bistudio.com/topic/187670-new-command-createobject-for-decorative-objects/

For medical we need to compare synced `createVehicleLocal` with the new command.

You can actually take physX damage from bandages if they are spawned funny, so this probably is a good thing to look at.

commy2:
>You can actually take physX damage from bandages if they are spawned funny, so this probably is a good thing to look at.

Maybe could also be used for all those cases where trying to place bomb while being to near the vehicle makes it flip?

PabstMirror:
Explosives are now virtual and local only (it's just a fake 3d element on your hud).

Being local also means less network traffic (I think old system was setPos each frame)

esteldunedain:
> Being local also means less network traffic (I think old system was setPos each frame)

The old plcamente system used `createVehicleLocal`, so it didn't generate network traffic either. The new system is better though, because the virtual object doesn't collide with the objects.

esteldunedain:
So in short; @commy2 that bug you describe has been fixed for a long time now.

bux578:
@BaerMitUmlaut Status?

Glowbal:
Does `createObject` have local effect?

bux578:
[`createSimpleObject`](https://community.bistudio.com/wiki/createSimpleObject)

Glowbal:
In that case it's not a simple switch  / replace for medical litter. We allow clients to independantly define how many litter objects can exist at any given time and for how long they can stay alive. Using a global variant will mean this would no longer be managed by clients, but the server instead.

esteldunedain:
I think tagging is the only trivial case to implement

bux578:
Having the setting for each client seems unnecessary complicated. `createSimpleObject` should solve performance issues.

commy2:
^ I think we would be better off if the server created and deleted litter, even if it means the client settings about litter would have to be removed / would do nothing anymore on clients.

PR #4197, opened 2016-08-01
Use createSimpleObject for medical litter

PabstMirror:
Ref discussion in #3537

- Uses `createSimpleObject` for medical litter
- ace_medical_litterSimulationDetail is now a server setting with global effects

Benifits:
- creation and cleanup loop code only runs on server
- no physX
- Objects do not show up in `(entities "")` - e.g. CBA xeh fallback won't waste time on them!

Drawbacks:
- @Glowbal mentioned that clients lose the ability to set their own litter levels.

```
_start = diag_tickTime;
_created = 0;
for "_xPos" from -25 to 25 do { 
    for "_yPos" from 0 to 25 do { 
        _position = getPosATL player; 
        _position = _position vectorAdd [_xPos, _yPos, 0]; 
        
        // _litterObject = createSimpleObject ["z\ace\addons\medical\data\littergeneric_bandages1.p3d", [0,0,0]]; 
        _litterObject = "ACE_MedicalLitter_bandage1" createVehicleLocal [0,0,0];

        _litterObject setDir random 360; 
        _litterObject setPosATL _position; 
        _created = _created + 1;
    }; 
};
systemChat format ["took %1 to make %2", (diag_tickTime - _start), _created];
```

`createSimpleObject` had much less effect on FPS
`createVehicleLocal` also created terrible lag for about 5 second (2-4 FPS) while the objects danced around and tried to stabilize on the ground

### DIFF
--- a/addons/medical/ACE_Settings.hpp
+++ b/addons/medical/ACE_Settings.hpp
@@ -159,8 +159,6 @@ class ACE_Settings {
         value = 3;
         values[] = {"Off", "Low", "Medium", "High", "Ultra"};
         _values[] = { 0, 50, 100, 1000, 5000 };
-
-        isClientSettable = 1;
     };
     class GVAR(litterCleanUpDelay) {
         category = CSTRING(Category_Medical);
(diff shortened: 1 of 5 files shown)


commy2:
>createVehicleLocal also created terrible lag for about 5 second (2-4 FPS) while the objects danced around and tried to stabilize on the ground

PhysX

Glowbal:
Still not entirely happy with clients losing their individual settings, but I can see benefits from this PR. Thoughts from you guys on this?

commy2:
:+1: The command was basically made for this. And synchronized bandages is a benefit for me. I don't see why anyone would disable them anyway. You sure won't be able to increase performance with that ...

PR #4203, opened 2016-08-03
dragging - Don't run animChanged event on all units

PabstMirror:
- Installs the animChanged EH on demand instead of having it running all the time for all units
- Also Fix CBA_fnc_addPlayerEventHandler

animChanged runs on all units (including non-local)
for units moving it fires about 1/second/unit
With 100+ units the overhead adds up (about 1 ms / second for 100 units on my system)

This installs the EH just for the unit that starts dragging/carrying

### DIFF
--- a/addons/dragging/CfgEventHandlers.hpp
+++ b/addons/dragging/CfgEventHandlers.hpp
@@ -48,14 +48,6 @@ class Extended_Killed_EventHandlers {
     };
 };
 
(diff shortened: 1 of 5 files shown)


commy2:
very nice :+1:

esteldunedain:
Perfect. 

Even though 1ms/s is very little, every ms counts.

PR #4205, opened 2016-08-03
Fix static weapons not being loadable in cargo

### LINKED ISSUE #4165: All static weapons cannot be loaded in the vehicle cargo
**Arma 3 Version**: 1.62 (stable)
**CBA Version**: 3.0.0 (stable)
**ACE3 Version**: 3.6.2 (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- All static weapons cannot be loaded in the vehicle cargo since v3.6.0. http://i.imgur.com/C1EzaJG.jpg
Option "ace_cargo_canLoad = 1; " for cfg's also does not work.
Statics can be load in vehicle only if synchronize it with the "Make Object Loadable" module. But this is not a solution in many situations.

**Steps to reproduce:**
- Place any static weapon and vehicle in the editor.
- Try to load this weapon in the vehicle cargo.
- Nothing happens

Everywhere

**Placed Modules:**
None

**RPT log file:**
Not relevant

PabstMirror:
- Fix #4165

### DIFF
--- a/addons/cargo/CfgEventHandlers.hpp
+++ b/addons/cargo/CfgEventHandlers.hpp
@@ -67,4 +67,9 @@ class Extended_InitPost_EventHandlers {
             init = QUOTE(_this call DFUNC(initObject));
         };
     };
+    class StaticWeapon {
+        class ADDON {
+            init = QUOTE(_this call DFUNC(initObject));
+        };
+    };
 };


esteldunedain:
So what was the issue; doesn't `StaticWeapon` inherit from `ThingX`?

Also, shouldn't the Mortar be included too? IIRC it didn't inherit from `StaticWeapon`.

PabstMirror:
Mortars inherit from LandVehicle->StaticWeapon.
But we want to run `FUNC(initObject)` on them, not `FUNC(initVehicle)` 

Note that this uses the XEH "object-inheritance" system, so the staticWeapon init will replace the landVehicle init.

esteldunedain:
> But we want to run FUNC(initObject) on them, not FUNC(initVehicle)

Oh, I get it. The ThingX init line was being overwritten by LandVehicle's init line, which called initVehicle and not initObject.

It'd probably be more robust to just call both functions for all children of ThingX and let te init functions decide which objects should be inited each way.

PabstMirror:
ThingX: Thing: All
StaticWeapon: LandVehicle: Land: AllVehicles: All

esteldunedain:
Oh, ok, sorry.

commy2:
ThingX is ammo boxes, ground weapon holders and I think some map props mainly.

PR #4208, opened 2016-08-04
Update for newest JR / 1.64 UBC

PabstMirror:
Ref https://github.com/CBATeam/CBA_A3/pull/436


PabstMirror:
Adjusted for changes to 1.64 and new JR (6.5mm muzzle)

PR #4209, opened 2016-08-05
lower wirecutter item mass

freghar:
The current mass makes it take a LOT of inventory space (~half of
an assault pack) which makes it impractical to carry around as a tool.

Lower it to use the same mass as the entrenching tool.

### DIFF
--- a/addons/logistics_wirecutter/CfgWeapons.hpp
+++ b/addons/logistics_wirecutter/CfgWeapons.hpp
@@ -10,7 +10,7 @@ class CfgWeapons {
         picture = QPATHTOF(ui\item_wirecutter_ca.paa);
         scope = 2;
         class ItemInfo: InventoryItem_Base_F {
-            mass = 65;
+            mass = 25;
         };
     };
 };


esteldunedain:
Seems very low for my taste.

If I'm not mistaken mass = 10 equals 1 lb / 0.5 Kg. I'd advocate for at least twice o three times that.

PabstMirror:
https://www.amazon.com/Neiko®-00560A-18-Inch-Cutter-Molybdenum/dp/B000JIKDC8/ - This is 3.3 pounds.

I'm like dropping it below 65 as it will make it easier to carry and use, 
I think 10 might be too low, but something 20-30 should be great

kwinteng:
On the same website the smallest one is 1.8 pounds that it would have to 20 minimum

freghar:
The problem is that you can rarely relate weight ingame to real weight.

For example ACE3's own 1000ml bag of blood plasma is 0.45kg, while it should be [1.03kg](http://www.aqua-calc.com/calculate/volume-to-weight), the entrenching tool has 0.45kg where they (AFAIK) typically have 2 pounds (0.9kg), the bomb defusal kit has 0.23kg (eh), etc. On the other hand more "valuable" things like explosives are the other way (M112 0.9kg instead of [0.57kg](https://en.wikipedia.org/wiki/C-4_%28explosive%29)).

So I just followed the example of using less-than-real mass for "utility" objects Arma 3 seems to have, which makes sense when you realize the game prevents you from carrying more than ~7.3kg in the 30L assault pack.
Additionally, on a typical Arma 3 fence, something [much more lighter and smaller](https://www.amazon.com/TEKTON-3386-8-Inch-Heavy-Duty-Cutter/dp/B000NQ4OYO) (0.3kg/0.66lb) would be probably sufficient.

If you still want me to change it, I can submit a version with 25, as a compromise in the 20-30 range you mentioned.

Thanks.

PabstMirror:
The in-game model is about 600mm long and looks like something I would call a bolt-cutter

I think 25 is a good value. It should still easily fit into backpacks and can even fit into some vests.

freghar:
Adjusted to 25.

PR #4211, opened 2016-08-06
Add createVehicle(Local) performance consideration to coding guidelines

jonpas:
- Title


commy2:
Does this apply to creating #lightsource, #soundsource, #particlesource, #crater  etc. ?

jonpas:
That's a very good question.

jonpas:
It does not @commy2, nice catch.

PR #4213, opened 2016-08-07
Fix the captive module from creating attributes when not needed

AACO:
Fix default values, so we don't get two custom attributes per unit, even when the unit is not handcuffed/surrendered

Pretty simple fix, saves a lot of space when dealing with large framework scale SQMs

### DIFF
--- a/addons/captives/CfgEden.hpp
+++ b/addons/captives/CfgEden.hpp
@@ -11,7 +11,7 @@ class Cfg3DEN {
                         expression = QUOTE(if (_value) then {[ARR_3(objNull,[_this],true)] call FUNC(moduleHandcuffed)});
                         typeName = "BOOL";
                         condition = "objectBrain";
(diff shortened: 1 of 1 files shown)


jonpas:
Nice fix! :+1: 

I believe pboProject will convert that to `0`, can you change it to `"(false)"`?

AACO:
I hope it wouldn't, since that should just be a valid string, but belt and suspenders never hurts.

jonpas:
Sadly pboProject's "help" is very unhelpful in that case.

PabstMirror:
Looks good, in compiled versions we were getting
```
defaultValue = 0;
```
because of pboProject.

In my testing, this prevents the game from adding the attribute to the sqm when just opening the attirbutes on the unit.

If you open the attributes, change a default, save and then change that attribute back to default, it will still save in the SQM.  But that seems to be just a a3 bug.

PR #4217, opened 2016-08-08
Russian Translations - Update

dudakov:


bux578:
Thanks @dudakov 

pinging @Tourorist (don't know if you're still into translating :smile:  but just in case)

addons/tagging/stringtable.xml
```diff
@@ -1,45 +1,88 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project name="ACE">
     <Package name="Tagging">
-        <Key ID="STR_ACE_Tagging_TagBlack">
-            <English>Tag black</English>
```
jonpas: All this red shouldn't be red.

dudakov: That's true. But I don't understand, why they are red.
ace3/master doesn't have these lines:
 https://github.com/acemod/ACE3/pull/4217/files/29a7c0524672845d351efbddeafca243a94808bd#r74041751

May be I messed something pulling this request...

jonpas: Not sure, will deal with it when we merge.

dudakov: Fixed.

Issue #4206, opened 2016-08-03
Limping from fire damage can only be healed with a PAK


Hawkeye104:
**Arma 3 Version:** `1.62` (stable)
**CBA Version:** `3.0.0` (stable)
**ACE3 Version:** `3.6.2` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
Damage sustained to legs that cannot be bandaged (e.g. fire) causes limping and can only be healed with a PAK. Is this intended behavior?

**Steps to reproduce:**
1. Place vehicle in Zeus
2. Drag health to 0 to cause an explosion
3. Go prone and back up into the fire to damage the legs
4. Limp away

**Where did the issue occur?**
Editor/Dedicated

**Placed Modules:**
Medical Settings
Advanced Medical Settings
Revive

PabstMirror:
Fire damage will trigger handleDamage_advanced which adds damage to `bodyPartStatus` but the incoming damage is usually small enough that it does not create new `openWounds`.

Because this now effects basic medical as well, we'll need to do something for this before 3.7

aeroson:
You can also "heal" Limping by shooting person into his legs.

Glowbal:
#4223

PR #4223, opened 2016-08-09
Bundle fire damage into larger chunks

PabstMirror:
Ref #4206

Fire damage comes in as a continuous stream, but each bit is lower than the threshold to trigger unknown damage. This saves the little chunks until together they addup to something that would be enough to trigger a wound from the extension.

Without this we add damage to `bodyPartStatus` but never create any wounds, so the damage would not be removable when in basic medical or healAfterBandage.

This also makes damage type `""` into `unknown` which is consistent with the non-dll wound code.

### DIFF
--- a/addons/medical/functions/fnc_handleDamage_advanced.sqf
+++ b/addons/medical/functions/fnc_handleDamage_advanced.sqf
@@ -9,9 +9,6 @@
  * 3: Shooter <OBJECT>
  * 4: Projectile <STRING>
  * 5: Hit part index of the hit point <NUMBER>
- * 6: Current damage to be returned <NUMBER>
- *
- * //On 1.63 dev:
  * 6: Shooter? <OBJECT>
  * 7: Current damage to be returned <NUMBER>
  *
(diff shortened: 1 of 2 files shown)


esteldunedain:
Good catch @PabstMirror! :+1:

Glowbal:
How would this look like with the new handle damage approach for the medical rewrite?

commy2:
The new approach just ignores damage below a certain threshold

addons/medical/functions/fnc_handleDamage_advanced.sqf
```diff
@@ -23,12 +20,23 @@
...
+params ["_unit", "_selectionName", "_amountOfDamage", "_sourceOfDamage", "_typeOfProjectile", "_hitPointNumber", "", "_newDamage"];
 
-//Temp fix for 1.63 handleDamage changes
-if (_newDamage isEqualType objNull) then {
-    _newDamage = _this select 7;
-};
+// For burning damage we will get a ton of very small hits of damage; they are too small to create any wounds
+// Save them up in a variable and run when it is over a noticable amount
+
+if ((_typeOfProjectile == "") && {_newDamage < 0.15} && {
```
Glowbal: Are we sure that fire is the only damage source that matches this condition? iirc falling also gives you an empty projectile string. Same with being driven over and a few other cases.

addons/medical/functions/fnc_handleDamage_wounds.sqf
```diff
@@ -20,6 +20,8 @@
 params ["_unit", "_selectionName", "_damage", "_typeOfProjectile", "_typeOfDamage"];
 TRACE_6("ACE_DEBUG: HandleDamage Called",_unit, _selectionName, _damage, _shooter, _typeOfProjectile,_typeOfDamage);
 
+if (_typeOfDamage == "") then {_typeOfDamage = "unknown";};
```
Glowbal: if this is necessary here, it's also necessary in the sqf fallback method for wounds.

PabstMirror: fallback already does this

Glowbal: Ah, my bad. Awesome! :+1:

addons/medical/functions/fnc_handleDamage_advanced.sqf
PabstMirror: This only kicks in for damage < 0.15 and I think most physx damage should be more than that, but can't guarentee.

The core problem is that we are increasing `bodyPartStatus` without adding new wounds.
Without wounds, the damage can't be healed, because bandageLocal does:
`if (count _openWounds == 0) exitWith {false}; // nothing to do here!`

I could try to do the fix there; so that `GVAR(healHitPointAfterAdvBandage)` or basic medical can still heal the hitpoints even if there are no wounds?

Glowbal:
:+1:

PR #4229, opened 2016-08-11
Add ability to show kills from ace medical

PabstMirror:
Will show kills at mission end screen.
Attempts to ID killer when unit's death was scripted.

Example below, note the vanilla kill system only shows 2 kills where as this system shows 5.
x

The last damage source isn't going to be perfect; e.g. get shot once, and then fall out of a chopper 30 minutes later.

I'm not sure if this even belongs in ace or if we should just leave it as an example.

### DIFF
--- a/addons/medical/XEH_PREP.hpp
+++ b/addons/medical/XEH_PREP.hpp
@@ -70,6 +70,7 @@ PREP(setDead);
 PREP(setHitPointDamage);
 PREP(setStructuralDamage);
 PREP(setUnconscious);
+PREP(setupKillTracking);
 PREP(showBloodEffect);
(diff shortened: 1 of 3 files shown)


commy2:
>The last damage source isn't going to be perfect; e.g. get shot once, and then fall out of a chopper 30 minutes later.

Maybe also store the time of the damage happening... and discard it when it happened too long ago.
ACEX ?

Glowbal:
Nice, but does not belong in the medical component.

SilentSpike:
Not a bad idea for ACEX :+1:

esteldunedain:
:+1: for making this an enabable/disabable module within ACE.

PR #4230, opened 2016-08-11
ACE rearm/refuel/repair actions for RHS vehicles

FadiCBB:
- Add rearm/refuel/repair actions for the applicable RHS AFRF vehicles (GAZ-66s, Urals)
- Add rearm/refuel/repair actions for the applicable RHS USF vehicles (M113 supply vehicle, HEMTT A4s)

So macros are a strange and new experience for me, full of wonder and amazement. I've never really worked with them before and while I think I've done it without screwing it up, someone should look them over before this is approved.

Also because setting up those three also requires disabling the vanilla equivalent system, should I add them to the requiredAddons?

### DIFF
--- a/optionals/compat_rhs_afrf3/CfgVehicles.hpp
+++ b/optionals/compat_rhs_afrf3/CfgVehicles.hpp
@@ -1,3 +1,77 @@
+#define MACRO_REARM_TRUCK_ACTIONS \
+        class ACE_Actions: ACE_Actions { \
+            class ACE_MainActions: ACE_MainActions { \
+                class EGVAR(rearm,TakeAmmo) { \
+                    displayName = ECSTRING(rearm,TakeAmmo); \
+                    distance = 7; \
+                    condition = QUOTE(_this call EFUNC(rearm,canTakeAmmo)); \
+                    insertChildren = QUOTE(_target call EFUNC(rearm,addRearmActions)); \
+                    exceptions[] = {"isNotInside"}; \
+                    showDisabled = 0; \
+                    priority = 2; \
+                    icon = PATHTOEF(rearm,ui\icon_rearm_interact.paa); \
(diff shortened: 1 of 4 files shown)


optionals/compat_rhs_afrf3/CfgVehicles.hpp
```diff
@@ -1,3 +1,77 @@
...
+            class ACE_MainActions: ACE_MainActions { \
+                class EGVAR(rearm,TakeAmmo) { \
+                    displayName = ECSTRING(rearm,TakeAmmo); \
+                    distance = 7; \
+                    condition = QUOTE(_this call EFUNC(rearm,canTakeAmmo)); \
+                    insertChildren = QUOTE(_target call EFUNC(rearm,addRearmActions)); \
+                    exceptions[] = {"isNotInside"}; \
+                    showDisabled = 0; \
+                    priority = 2; \
+                    icon = PATHTOEF(rearm,ui\icon_rearm_interact.paa); \
```
jonpas: Use `QPATHTOEF` instead of `PATHTOEF` everywhere.

jonpas:
Apart from the comment, looks good! :+1: 

> someone should look them over before this is approved.

Don't worry, we won't merge anything blindly (at least not intentionally). :laughing: 

> Also because setting up those three also requires disabling the vanilla equivalent system, should I add them to the requiredAddons?

Yes.

PabstMirror:
Because we want to add to main actions we need to add inheritance chain for `ace_actions` all the way up to Car or Tank base.

I fixed the inheritance for both files here: https://github.com/acemod/ACE3/commit/e5d7ee5f0fbd94f1d96d182edc7bfe732bfb323f

Glowbal:
@PabstMirror merge?

PR #4234, opened 2016-08-11
Cache config UI

jonpas:
- Cache all selective UI config at mission start using CBA namespace
- Make `compile` only be called once per config on mission start instead of on every `infoDisplayChanged` and `settingChanged` event
- ~33% improvement in `fnc_setAdvancedElement` (~0.09 ms -> ~0.062 ms)


jonpas:
Just did a test and for some reason this is actually a bit slower.
Old: ~0.052 ms
New: ~0.072 ms

I am a tad bit confused honestly. Main goal was to prevent `compile` taking up precious execution time, and it's not `select` being slow either.

jonpas:
Was using wrong argument in the test case, proper test case shows (`ammoCount` which also has one condition `"(false)"`):
Old: ~0.09 ms
New: ~0.115 ms
New without `select`: ~0.07 ms

So `select` does overall make it slower, which kinda defeats the purpose of this PR. I imagine with many conditions this way would eventually pay off, but not with the 2 we have now.

esteldunedain:
I'm under the impression that BIS sped up reading from config. While fiddling the other day it seemed like caching a config entry on a namespace vs reading it from config wasn't much different anymore.

It should be properly confirmed though. I bet there're perf numbers for config reads buried somewhere among old PRs.

jonpas:
If that's true then the only reason to do any caching is due to `compile`, which in this case is minimal as we don't have many conditions that have to be evaluated all the time. I am going to try another version with just `compile` cache and add some more conditions to get somewhat realistic numbers.

jonpas:
- Changed to use CBA Namespaces
- Improved `setAdvancedElement` function speed
- Simplified scripted API function (namespaces mostly)

Over-all, I got the `setAdvancedElement` function to run in ~0.062 ms with current set of conditions (same test as above). This is good to go! :+1:

commy2:
>I'm under the impression that BIS sped up reading from config. While fiddling the other day it seemed like caching a config entry on a namespace vs reading it from config wasn't much different anymore.

That's a tough one. Configs are read from disk, so it really depends on your setup.

jonpas:
> Configs are read from disk, so it really depends on your setup.

Are you sure? Game compiles them on start, I would hope they are in memory.

esteldunedain:
> I would hope they are in memory.

Yeah, I think so too. My guess was that they were in memory, but structured in such a way that accessing them was not performant. If that's the case they might have changed the implementation.

esteldunedain:
As far as I can tell, Arma used to read all config entries and store them as a tree. Each node of the tree is a config class, that contains all their children. Children are accessed by name using a hashtable. That implementation seems to make sense, and shouldn't be too slow.

IDK why we used to measure such poor performance at some point, but I'm quite sure that was the case.

jonpas:
That's good to know, would be nice if we could get some definite answer on it, but I am not sure that's likely.

PR #4236, opened 2016-08-12
new unconsciousness ragdol

BioZ0102:
Make player/AI be in ragdol when unconsciousne. like in screenshots.

### DIFF
--- a/addons/javelin/XEH_pre_init.sqf
+++ b/addons/javelin/XEH_pre_init.sqf
@@ -7,4 +7,28 @@ ADDON = false;
 GVAR(isLockKeyDown) = false;
 GVAR(pfehID) = -1;
 
+DFUNC(disableFire) = {
+    params ["_firedEH"];
+
+    if(_firedEH < 0 && {difficulty > 0}) then {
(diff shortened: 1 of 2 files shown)


bux578:
There's something wrong with this.
Feature requests go here: https://github.com/acemod/ACE3/issues/3594/

PabstMirror:
@BioZ0102 

Adding this code to init.sqf should trigger the 1.62 ragdoll
```
["ace_unconscious", {
    params ["_unit", "_state"];
    if (local _unit) then {
        _unit setUnconscious _state;
    };
```

esteldunedain:
Why are we not doing that anymore? d0c30577afd62b8f125711a4c1700b495af32cee

Glowbal:
It was removed because I didn't get to try it out properly yet at that time and we were just before the release. It wasn't doing anything in earlier releases and I didn't feel like keeping it if I didn't know the full effects.

commy2:
There is a annoying heartbeat sfx that plays with this. This is also confusing as the rhytm of the heart beat has nothing to do with what is simulated in advanced medical.

we will come back to this I think. No point in making feature requests now.

PR #4242, opened 2016-08-13
fix bug which would cause unecessary echo network traffic

commy2:
The funtion would call itself on other machines, where it would rebound to all other machines again. n^2
It's not infinite thanks to the `exitWith` above, but still unecessary.


Glowbal:
Nice catch by you and @PabstMirror

PR #4247, opened 2016-08-14
Read git version from file rather than using Git client in make.py

jonpas:
- Title

If we want this I will update the doc as well.

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -678,28 +678,39 @@ def get_commit_ID():
     # Get latest commit ID
     global make_root
     curDir = os.getcwd()
+    commit_id = ""
 
     try:
+        # Verify if Git repository
         gitpath = os.path.join(os.path.dirname(make_root), ".git")
         assert os.path.exists(gitpath)
-        os.chdir(gitpath)
 
-        # Try to get commit ID from git file
-        commit_id = ""
-        with open("HEAD", "r") as head_file:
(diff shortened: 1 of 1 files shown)


Glowbal:
Should be tested with a detached head & on a branch with `/` in the name, ie `feature/hello-world`.

jonpas:
Detached HEAD and branches with slashes fixed. Also implemented it only as a fall-back solution (which made the code uglier but workable).

freghar:
> Should be tested with a detached head & on a branch with / in the name, ie feature/hello-world.

If you want to go to those lengths, you could add support for bare repositories and packed refs, at which point you might just use GitPython.

The real question is why would you ever parse the commit ID out yourself - if it's for build identification, git can do this for you on checkout, see "ident" in [gitattributes(5)](https://git-scm.com/docs/gitattributes).

(Sorry if this is irrelevant, I'm not that familiar with your build internals.)

jonpas:
We use commit ID for the release build, for quick identification of the commit it was based on. We use `git rev-parse HEAD` to get the current checked out commit ID. Problem is if you use Bash on Ubuntu on Windows (from Windows Subsystem for Linux) you cannot use that as it's a separate thing, thus fall-back to reading it directly from Git repo files.

Glowbal:
Not perfect, but hopefully soon we won't need this anymore.

PR #4248, opened 2016-08-16
Improvments for advanced throwing branch

PabstMirror:
For PR #3477
- FUNC(drawThrowable): Fix issue getting correct `_throwableMag` when you have a primed grenade. If you had nothing in inventory it would end up being nil otherwise it would use the wrong grenade type's initSpeed. This was a little messy to fix because we just have the ammo type and there could possibly be multiple magazines that use the same ammo. Uses a lookup hash that gets generated at the start.

- FUNC(prepare): When you have no grenades in inventory and picking up a live grenade it would fail because selectNextGrenade returned true.

- Handle detaching from ace_attach.

### DIFF
--- a/addons/advancedthrowing/CfgVehicles.hpp
+++ b/addons/advancedthrowing/CfgVehicles.hpp
@@ -50,7 +50,7 @@ class CfgVehicles {
 
     class Items_base_F;
     class GVAR(pickUpHelper): Items_base_F {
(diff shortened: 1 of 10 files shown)


SilentSpike:
:+1: All looks valid to me

jonpas:
Thanks @PabstMirror !

PR #4249, opened 2016-08-16
Use pushBackUnique in Tags cache

jonpas:
- Title
- Add maintainers
- Use `configProperties` with inheritance instead of `configClasses`
- Invert a condition to be more readable

### DIFF
--- a/addons/tagging/functions/fnc_compileConfigTags.sqf
+++ b/addons/tagging/functions/fnc_compileConfigTags.sqf
@@ -46,8 +46,6 @@
 
     if (!_failure) then {
         GVAR(cachedTags) pushBack [_class, _displayName, _requiredItem, _textures, _icon];
-        if !(_requiredItem in GVAR(cachedRequiredItems)) then {
-            GVAR(cachedRequiredItems) pushBack _requiredItem;
-        };
+        GVAR(cachedRequiredItems) pushBackUnique _requiredItem;
     };
 } forEach ("true" configClasses (configFile >> "ACE_Tags"));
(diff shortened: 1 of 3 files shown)


addons/tagging/functions/fnc_compileConfigTags.sqf
```diff
@@ -46,8 +46,6 @@
 
     if (!_failure) then {
         GVAR(cachedTags) pushBack [_class, _displayName, _requiredItem, _textures, _icon];
-        if !(_requiredItem in GVAR(cachedRequiredItems)) then {
-            GVAR(cachedRequiredItems) pushBack _requiredItem;
-        };
+        GVAR(cachedRequiredItems) pushBackUnique _requiredItem;
     };
-} forEach ("true" configClasses (configFile >> "ACE_Tags"));
+} forEach (configProperties [configFile >> "ACE_Tags"]);
```
commy2: Will also include non-class entries, which would error out if you do e.g. `_x >> "displayName"`
I guess that's fine though as the class is not supposed to have such entries.

jonpas: Doesn't `configClasses` do the same? Return both entries and sub-classes?

commy2: No, only classes. That's why it's called `configCLASSES`.

Equivalent would be:
`configProperties [configFile >> "ACE_Tags", "isClass _x"];`

jonpas: Ah, I see, I'll change it back then to be safe.

PR #4254, opened 2016-08-17
skip DLL check on non-windows machines

### LINKED ISSUE #4253: Arma 3 Linux server extension dll not install
Arma 3 Version: 1.62 (stable)
CBA Version: 3.0.0.160713 (stable)
ACE3 Version: 3.6.2 (stable)

Mods:
- @CBA_A3
- @ace

Description:
Arma3 server in ubuntu. After the role selection screen and press continue. Before the screen brifing the following error appears in the log:

```
[ACE] (common) ERROR: Extension ace_break_line.dll not installed.
[ACE] (common) ERROR: Extension ace_parse_imagepath.dll not installed.
[ACE] (common) ERROR: Extension ace_clipboard.dll not installed.
[ACE] (common) ERROR: Extension ace_fcs.dll not installed.
[ACE] (common) ERROR: Extension ace_advanced_ballistics.dll not installed.
```

This occurs since the last update of Arma.

commy2:
- title
- close #4253

### DIFF
--- a/addons/common/functions/fnc_checkFiles.sqf
+++ b/addons/common/functions/fnc_checkFiles.sqf
@@ -53,7 +53,7 @@ _addons = _addons select {_x find "ace_" == 0};
 // check dlls
 ///////////////
 if !(toLower (productVersion select 6) in ["linux", "osx"]) then {
-    ACE_LOGINFO_2(operating system does not support DLL file format);
+    ACE_LOGINFO_2("operating system does not support DLL file format");
 } else {
     {
         private _versionEx = _x callExtension "version";
--- a/addons/common/functions/fnc_checkFiles.sqf
+++ b/addons/common/functions/fnc_checkFiles.sqf
@@ -52,7 +52,9 @@ _addons = _addons select {_x find "ace_" == 0};
 ///////////////
 // check dlls
 ///////////////
-if (productVersion select 6 == "Windows") then {
+if !(toLower (productVersion select 6) in ["linux", "osx"]) then {
+    ACE_LOGINFO_2(operating system does not support DLL file format);
+} else {
     {
         private _versionEx = _x callExtension "version";
 
--- a/addons/common/functions/fnc_checkFiles.sqf
+++ b/addons/common/functions/fnc_checkFiles.sqf
@@ -52,23 +52,25 @@ _addons = _addons select {_x find "ace_" == 0};
 ///////////////
 // check dlls
 ///////////////
-{
-    private _versionEx = _x callExtension "version";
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_checkFiles.sqf
```diff
@@ -52,23 +52,25 @@ _addons = _addons select {_x find "ace_" == 0};
 ///////////////
 // check dlls
 ///////////////
-{
-    private _versionEx = _x callExtension "version";
+if (productVersion select 6 == "Windows") then {
```
bux578: I don't like the check for "Windows" here.
It would be Better to inverse it and check for the equivalent for Linux/Mac.

commy2: https://community.bistudio.com/wiki/productVersion
>String (Arma 3) - Platform ("Windows", "Linux" or "OSX")

how else would you do that?

>It would be Better to inverse it and check for the equivalent for Linux/Mac.

`.so` and the MacOS equivalent are not supported by our code and would require further changes in other places too. What do you mean by "inverse"?

bux578: That you rather check for Linux/OSX and skip the test then to check for "Windows". (It's just about the strings). My fear is that BI could change the string from `Windows` to something different (e.g. `Win`, `Windows10`, `Windows-x86`). And I would like to see it rather fail for the least amount of players if that will ever happen.

addons/common/functions/fnc_checkFiles.sqf
```diff
@@ -52,23 +52,25 @@ _addons = _addons select {_x find "ace_" == 0};
...
         };
-    } else {
-        // Print the current extension version
-        ACE_LOGINFO_2("Extension version: %1: %2",_x,_versionEx);
-    };
-    false
-} count getArray (configFile >> "ACE_Extensions" >> "extensions");
+        false
+    } count getArray (configFile >> "ACE_Extensions" >> "extensions");
+};
```
Glowbal: Log if the product version does not support our dlls?

addons/common/functions/fnc_checkFiles.sqf
esteldunedain:
````
if !(productVersion select 6 in ["Linux", "OSX"]) then {
```
?

addons/common/functions/fnc_checkFiles.sqf
commy2: done

commy2:
"Linux" and "OSX" will be detected and logged as DLL incompatible. otherwise the check will be performed as usual.

PR #4263, opened 2016-08-18
Remove never-merged rearm feature and framework documentation

jonpas:
- Title


jonpas:
Will be re-added when/if the feature is implemented.

PR #4266, opened 2016-08-19
Fix Events Framework documentation 3.1 section

jonpas:
- Title
- Use `target="_blank"` for CBA Events documentation link
- ~~Add dots after section numbers~~
- Use code for component names

### DIFF
--- a/docs/wiki/framework/events-framework.md
+++ b/docs/wiki/framework/events-framework.md
@@ -15,7 +15,7 @@ Event handlers in ACE3 are implemented through CBA's Event System. They should b
 ## 2. Events List
 ACE3 uses many events internally, this is not a complete list; but it should list most publicly usable events.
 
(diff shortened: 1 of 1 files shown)


bux578:
I don't like the closing point on the numbers.

That's how we have it everywhere:
`1.`
`1.1`
`1.1.1`

jonpas:
> That's how we have it everywhere:

Not true, we have a terrible mix of both.

jonpas:
I am all for standardizing it, don't get me wrong. :laughing: 
I like closing point on main numbers and no in subsequent. And it appears it is [documented](http://ace3mod.com/wiki/development/documentation-guidelines-and-tips.html) like that as well. I shall change it.

jonpas:
Done, and another PR for streamlining closing points opened.

PR #4268, opened 2016-08-19
Streamline dependencies extraction tool

jonpas:
- Update `extract_dependencies.py` tool to extract them directly into `docs/_includes/dependencies_list.md`, had to be copied previously due to different branch
- Ignore `main` component as that is always required (and so denoted in the wiki page)
- Add newlines between Jekyll `if` statements for better readability
- Force Python to write LF line endings


bux578:
I'm trusting you on this.

PR #4272, opened 2016-08-19
Allow inheritance for addActionToClass

PabstMirror:
Handles the mess of adding actions to all units.
E.G. can add class actions to `CaManBase`.

### DIFF
--- a/addons/interact_menu/functions/fnc_addActionToClass.sqf
+++ b/addons/interact_menu/functions/fnc_addActionToClass.sqf
@@ -8,7 +8,7 @@
  * 1: Type of action, 0 for actions, 1 for self-actions <NUMBER>
  * 2: Parent path of the new action <ARRAY>
  * 3: Action <ARRAY>
- * 4: Use Inheritance (Default: False) <BOOL><OPTIONAL>
+ * 4: Use Inheritance <BOOL> (default: false)
  *
  * Return Value:
  * The entry full path, which can be used to remove the entry, or add children entries <ARRAY>.
--- a/addons/interact_menu/functions/fnc_addActionToClass.sqf
+++ b/addons/interact_menu/functions/fnc_addActionToClass.sqf
@@ -8,6 +8,7 @@
  * 1: Type of action, 0 for actions, 1 for self-actions <NUMBER>
  * 2: Parent path of the new action <ARRAY>
  * 3: Action <ARRAY>
+ * 4: Use Inheritance (Default: False) <BOOL><OPTIONAL>
  *
  * Return Value:
  * The entry full path, which can be used to remove the entry, or add children entries <ARRAY>.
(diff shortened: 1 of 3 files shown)


commy2:
This is badly needed, thanks!

addons/interact_menu/functions/fnc_addActionToClass.sqf
```diff
@@ -8,20 +8,37 @@
  * 1: Type of action, 0 for actions, 1 for self-actions <NUMBER>
  * 2: Parent path of the new action <ARRAY>
  * 3: Action <ARRAY>
+ * 4: Use Inheritance (Default: False) <BOOL><OPTIONAL>
```
jonpas: Standardly we do this like:
` * 4: Use Inheritance <BOOL> (default: false)`

jonpas:
No more `GVAR(initializedClasses)`! :tada:

addons/interact_menu/functions/fnc_addActionToClass.sqf
```diff
@@ -8,20 +8,37 @@
...
  *
  * Return Value:
  * The entry full path, which can be used to remove the entry, or add children entries <ARRAY>.
  *
  * Example:
  * [typeOf cursorTarget, 0, ["ACE_TapShoulderRight"],VulcanPinchAction] call ace_interact_menu_fnc_addActionToClass;
  *
- * Public: No
+ * Public: Yes
```
commy2: should set `FUNC(createAction)` public=yes then too.

addons/interact_menu/functions/fnc_addActionToClass.sqf
commy2: `<OPTIONAL>` looks strange
`(optional, default: false) <BOOL(EAN)>` ?
Maybe it's just me.

commy2:
>No more GVAR(initializedClasses)! :tada:

The PR should probably convert them to use this. Will go through stuff after I get home tomorrow.

PabstMirror:
Dragging, Repair, Cargo, all do something similar but they are also reading configs, so it won't exactly be the same.

jonpas:
Ah, that's very true, so dies Sitting for example. Not sure if it would be better to leave it like this or add the config reading in condition rather though.

docs/wiki/framework/interactionMenu-framework.md
```diff
@@ -97,8 +97,10 @@ Important: `ace_common_fnc_canInteractWith` is not automatically checked and nee
  * 1: Type of action, 0 for actions, 1 for self-actions <NUMBER>
  * 2: Parent path of the new action <ARRAY>
  * 3: Action <ARRAY>
+ * 4: Use Inheritance (Default: False) <BOOL><OPTIONAL>
```
jonpas: Same here: ` * 4: Use Inheritance <BOOL> (default: false)`

esteldunedain:
For performance reasons I think it's best to NOT use inherited actions for most modules. I think it's better to take that extra time at mission start to only add the actions to classes that need them instead of mass adding them to all classes and evaluating false condition every time player tries to interact near them.

docs/wiki/framework/interactionMenu-framework.md
```diff
@@ -136,6 +138,18 @@ _action = ["Open RDF","Radio Direction Finder","pabst\RDF.jpg",_statement,_condi
 [(typeOf _unit), 1, ["ACE_SelfActions"], _action] call ace_interact_menu_fnc_addActionToClass;
 ```
 
+Using addActionToClass inheritance:
```
jonpas: `addActionToClass` (code)

PR #4273, opened 2016-08-20
Cache FUNC(canInteractWith) (?)

commy2:
- I still have this one on my disk. No idea if it works, but it sounded interesting.


Glowbal:
We should probably cache the calls to canInteractWith instead of caching the return of it. Doing this in a wrapper method might be preferable.

commy2:
Yeah. Tbh I haven't even looked at it since I made it back before 3.2 was released.

commy2:
I'll bury this.

PR #4278, opened 2016-08-21
Add setUniqueEventHandler

PabstMirror:
`setUniqueEventHandler` allows adding a event handler to a unit via a string ID which will handle cleanup if the unit respawns. It's based on `CBA_fnc_addBISEventHandler`

@BaerMitUmlaut this should fix the problem with the fatgigue anim eh (https://github.com/acemod/ACE3/blob/advanced-fatigue/addons/advanced_fatigue/functions/fnc_handlePlayerChanged.sqf#L17)

Why we need this:
Lets say there is player object `A` with AnimChanged event handler ID: 0
If `A` dies and respawns in to object `B`, all of `A`'s variables and event handlers will be copied as well.
Removing the event handler from the dead body `A` will not remove it from `B`

Added to hearing as an example:
We can replace the XEH running on every unit, and skip having to check every shot if the unit affected is the player. FiredNear can happen a lot (5 people in a fireteam each shoot once is 25 events)

Downsides to this method:
- Slightly more code; both addBISEventHandler and this add some overhead. But we can eliminate almost the amount from firedEH as we don't have to check the unit.
- Scripted event handlers could be removed via mission/other addon.

### DIFF
--- a/addons/common/XEH_PREP.hpp
+++ b/addons/common/XEH_PREP.hpp
@@ -145,6 +145,7 @@ PREP(setPitchBankYaw);
 PREP(setProne);
 PREP(setSetting);
 PREP(setSettingFromConfig);
+PREP(setUniqueEventHandler);
 PREP(setVariableJIP);
 PREP(setVariablePublic);
 PREP(setVolume);
--- a/addons/common/functions/fnc_setUniqueEventHandler.sqf
+++ b/addons/common/functions/fnc_setUniqueEventHandler.sqf
@@ -1,0 +1,64 @@
+/*
+ * Author: PabstMirror
+ * Adds a unique event handler with a unique string handle.
+ * Only stays on for the current unit and will be removed on repawn.
+ * Pass emtpy code {} to remove.
+ *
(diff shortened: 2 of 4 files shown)


esteldunedain:
I like this approach. I think the amount of overhead it adds is not that bad, but it's probably enough so that usage should be evaluated on a case by case basis. 

Seems very appropiate for ace_hearing but..

if I'm not mistaken we didn't use to install firedNear EHs on vehicles; now we do.  Wasthat a bug @PabstMirror ?

PabstMirror:
I think it was something I messed up in #3802 (https://github.com/acemod/ACE3/pull/3802/files#diff-b662c35beaa8b4d21a78842e3d0d8c02L25)

It used to be on AllVehicles, now just on CaManBase

A unit in a vehicle won't trigger a firedNear event, so it should also be on the vehicle

esteldunedain:
Perfect, that clears it then :+1:

> A unit in a vehicle won't trigger a firedNear event, so it should also be on the vehicle

Right, as suspected as much.

PR #4282, opened 2016-08-22
convert log macros to cba versions

commy2:
- converts the ACE logging macros to the CBA ones.
- CBA ones will receive our cool formatting after the next cba update

This also has the benefit of it using this new cool error pop up:
https://camo.githubusercontent.com/69f9767d74e54dbf005d5f31f1305b1a2a33e35f/687474703a2f2f692e696d6775722e636f6d2f304b66303838412e706e67

- it also deletes the format-version (e.g. `ACE_ERRORFORMAT`). I have no idea what their purpose would be. They are used nowhere.

I also used noticed that we already use the CBA macros in some places. This would make it overall more consistent.


PabstMirror:
We want to replace `// #define CBA_DEBUG_SYNCHRONOUS` with `// #define DEBUG_SYNCHRONOUS`

ACEX headless is using some of ace's log macros 

Do we really want end users seeing the big UI warning on screen? I think we should really think about when to use the ERROR macro and make sure it won't show up for trivial errors.

jonpas:
> ACEX headless/viewrestriction is using some of ace's log macros 

We should update ACEX soon anyways, 2 feature-breaking bugs in current release of it.

addons/main/script_macros.hpp
```diff
@@ -1,11 +1,4 @@
-// BWC for CBA's DEBUG_SYNCHRONOUS - https://github.com/CBATeam/CBA_A3/pull/466/
-#ifdef CBA_DEBUG_SYNCHRONOUS
-    // For New CBA:
-    #define DEBUG_SYNCHRONOUS
-    // For Old CBA:
-    #define CBA_fnc_log { params ["_file","_lineNum","_message"]; diag_log [diag_frameNo, diag_tickTime, time,  _file + ":"+str(_lineNum + 1), _message]; }
-#endif
-
+#define DEBUG_SYNCHRONOUS
```
commy2: @PabstMirror ^ debug synchronous is always enabled in ACE. I don't think it makes sense to ever have that off.

commy2:
>Do we really want end users seeing the big UI warning on screen? I think we should really think about when to use the ERROR macro and make sure it won't show up for trivial errors.

I think so, but it's true that we shouldn't use that for trivial things. There always is the warning macro for those.

PR #4285, opened 2016-08-23
Improve interaction position for air units.

### LINKED ISSUE #4233: VTOL blackfish cannot be targeted by ACE3 action key (repaired too)
**ACE3 Version:** `3.6.2` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- the nato VTOL, the "blackfish" , due to its huge mass and too high "center point" , cannot be targeted by ACE3 interaction. Not for "virtual arsenal" and for repairs with ace3 mod too!

**Steps to reproduce:**
- open the editor, place a player, a nato VTOL , and damage it a bit, then try to repair it.

- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) 

**Placed Modules:**
- none

PabstMirror:
Fix #4233

Adds a varient of getVehiclePos function that should work better for air units.
Allows us to define a custom bounding box for a vehicle to limit to the main hull of a air unit and ignore it's wings or rotors.

Example, green is boundingBoxReal, blue is the effective custom one:

this function doesn't require higher scoped variables, so we can use it externally.
For example to get get the effective distance to a vehicle:

```
private _playerPosASL = eyePos _player;
private _targetInteractionPos = [_target, _playerPosASL] call FUNC(getVehiclePosComplex);
(ASLtoAGL _playerPosASL) distance (_target modelToWorld _targetInteractionPos)
```

### DIFF
--- a/addons/interaction/functions/fnc_getVehiclePosComplex.sqf
+++ b/addons/interaction/functions/fnc_getVehiclePosComplex.sqf
@@ -1,0 +1,59 @@
+/*
+ * Author: esteldunedain, PabstMirror
+ * Return a suitable position for the action point for the given target vehicle
+ *
+ * Arguments:
+ * 0: Target <OBJECT>
+ * 1: Player's Position ASL <ARRAY>
+ *
+ * Return value:
+ * Interaction point in model cords <ARRAY>
+ *
+ * Example:
+ * [cursorTarget, eyePos player] call ace_interaction_fnc_getVehiclePosComplex
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_target", "_cameraPosASL"];
+TRACE_2("params",_target,_cameraPosASL);
+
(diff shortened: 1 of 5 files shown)


addons/interaction/functions/fnc_getVehiclePosComplex.sqf
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * Example:
+ * [cursorTarget, eyePos player] call ace_interaction_fnc_getVehiclePosComplex
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_target", "_cameraPosASL"];
+TRACE_2("params",_target,_cameraPosASL);
```
jonpas: Should probably comment it before merge.

jonpas:
Nice work as always @PabstMirror!

PR #4289, opened 2016-08-24
Fix null ace_player from early call to addPlayerEH

PabstMirror:
It is possible that `CBA_fnc_addPlayerEventHandle`r has already been called and run,
We will NOT get any events for the initial state, so manually set ACE_player

e.g. in preInit or SP mission init.sqf do
```
["loadout", {
    diag_log text format ["%1 - Loadout change - %2", time, _this];
```
and `ace_player` will never bet set because the event has already happened.

Skimmed our use of the function and fixed noradio as well, everything else seemed ok.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -295,6 +295,13 @@ enableCamShake true;
 // Set up numerous eventhanders for player controlled units
 //////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


PabstMirror:
@commy2 - I've been also thinking we could tweaks this on CBA's end.
Maybe delay adding state checker until after postInit.
Or trigger event on add if the player is already defined.

PR #4292, opened 2016-08-24
Add note about no settings class for serverconfig

Glowbal:
- Add a note that clarifies there is no need for a settings class in the serverconfig.


jonpas:
:+1:

PR #4304, opened 2016-08-27
Chemlights cleanup

PabstMirror:
Minor cleanup of @voiperr 's chemlights

For the fired event handlers:
`FUNC(throwEH)` only does anything for local units so we can remove `firedPlayerNonLocal`; 
and I don't think AI have any idea how to use chemlights so I think we can skip `firedNonPlayer` as well?

Add support for handling primed IR grenades for advanced throwing (@jonpas)

### DIFF
--- a/addons/chemlights/XEH_postInit.sqf
+++ b/addons/chemlights/XEH_postInit.sqf
@@ -2,6 +2,6 @@
 
 if (!hasInterface) exitWith {};
 
-["firedPlayer", DFUNC(throwEH)] call EFUNC(common,addEventHandler);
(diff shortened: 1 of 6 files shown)


jonpas:
Advanced Throwing support seems to work great! :+1: 
I'll add pick up support for it in a separate PR.

It also fixes 3 times:
```
[ACE] (common) WARNING: ace_common_fnc_addEventHandler is deprecated. Support will be dropped in version 3.8.0. Replaced by: CBA_fnc_addEventHandler File: z\ace\addons\common\functions\fnc_addEventHandler.sqf Line: 13
```

PR #4306, opened 2016-08-27
Increase blood bag flow rate for basic medical

PabstMirror:
- Increase flow rate for basic medical (41ml/sec in an arm vein probably doesn't feel great :smile: )
- Don't queue bags - (2 bags flow twice as fast, instead of just doubling the total to add)

Uses a single unit variable `QGVAR(ivBags)` to hold all bag info.
We can skip calculations and syncing if the value is nil (no bags) which should be the common case.

### DIFF
--- a/addons/medical/XEH_preInit.sqf
+++ b/addons/medical/XEH_preInit.sqf
@@ -5,7 +5,6 @@ ADDON = false;
 #include "XEH_PREP.hpp"
 
 GVAR(injuredUnitCollection) = [];
-GVAR(IVBags) = [];
 
 private _versionEx = "ace_medical" callExtension "version";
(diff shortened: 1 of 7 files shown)


Glowbal:
This PR removes the seperate lists for blood, plasma and saline. I think we still want to keep track of the types of IV that are providing the regen. It does not matter for basic, but could matter a lot for advanced medical in the near future.

PabstMirror:
In fnc_treatmentIVLocal:
```
_bloodBags pushBack [_volumeAdded]; // Future BagType: [_volumeAdded, _typeOf]
```
I was thinking in the future the `ivBags` array would hold arrays with the volume and type when we had a use for it. This lets us just check a single variable for speed, but we can still have all the info if we need it.

PR #4316, opened 2016-09-01
Fade nametags towards screen border

BaerMitUmlaut:
- Make nametags fade depending on distance to the screens center

PS: This component needs some cleanup.

### DIFF
--- a/addons/nametags/functions/fnc_onDraw3d.sqf
+++ b/addons/nametags/functions/fnc_onDraw3d.sqf
@@ -57,6 +57,11 @@ switch (GVAR(showPlayerNames)) do {
         _enabledTagsNearby = (GVAR(showSoundWaves) == 2);
         _enabledTagsCursor = _onKeyPressAlphaMax > 0;
     };
+    case 5: {
+        // Fade on border
+        _enabledTagsNearby = true;
+        _enabledTagsCursor = false;
+    };
 };
 
 private _ambientBrightness = ((([] call EFUNC(common,ambientBrightness)) + ([0, 0.4] select ((currentVisionMode ace_player) != 0))) min 1) max 0;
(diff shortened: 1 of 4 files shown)


jonpas:
This should be an option.

esteldunedain:
Three comments:
1 - I like it; in fact I remember programming this in the past; probably for AGM. It somehow got lost along the way,

2 - IMO it should be optional, as @jonpas said

3 - I'd like to see the compared perf of this, e.g. when 20 units are in sight. I took quite a bit of effort optimizing that loop on the last cleanup and it made quite a difference; that code can be called well over 1000 times per second when a lot of units are nearby.

addons/nametags/functions/fnc_onDraw3d.sqf
```diff
@@ -132,7 +132,15 @@ if (_enabledTagsNearby) then {
         if !(isNull _target) then {
             private _relPos = (visiblePositionASL _target) vectorDiff _camPosASL;
             private _distance = vectorMagnitude _relPos;
-            private _projDist = _relPos vectorDistance (_vecy vectorMultiply (_relPos vectorDotProduct _vecy));
+
+            private _screenPos = worldToScreen (_target modelToWorld (_target selectionPosition "head"));
+            if (_screenPos isEqualTo []) then {
```
jokoho48: why continue the render process? this is wased time. if the Nametag is not on screen.

PabstMirror:
@esteldunedain - I guess I'm the one who removed the old fading. :sob: I don't think it was intentional because I left the unused `_projDist` var. 
https://github.com/acemod/ACE3/commit/ebeba23834965ced0a7009cc1cd4cf3e662689a6#diff-8f5258d40c7670f6b1e6ae7f2f2ccbabL100

Old way did this to do edge fading.
```
 _projDist = _relPos vectorDistance (_vecy vectorMultiply (_relPos vectorDotProduct _vecy));
_alpha = ((1 - 0.2 * (_distance - _maxDistance)) min (1 - 0.15 * (_projDist * 5 - _distance - 3)) min 1) * GVAR(PlayerNamesMaxAlpha);
```

Not sure if that's any better than new?

jonpas:
Haha, in that case I'd still prefer a setting and make fading default.

BaerMitUmlaut:
Performance test with 20 AI:

Before:


After:


Now optional, I ordered it in between just show and show on cursor in the module, but to keep backwards compat I couldn't do that in the ACE settings.

Glowbal:
lgtm

esteldunedain:
:+1: Nice job @BaerMitUmlaut

PR #4334, opened 2016-09-05
Disallow radio usage on captivity and surrender

kwinteng:
- end the current radio transmission on captivity and surrender
- for both TFAR and ACRE disallow radio usage on captivity and surrender
- allow radio usage again on release/cancel surrender

### DIFF
--- a/addons/captives/functions/fnc_setHandcuffed.sqf
+++ b/addons/captives/functions/fnc_setHandcuffed.sqf
@@ -41,6 +41,11 @@ if ((_unit getVariable [QGVAR(isHandcuffed), false]) isEqualTo _state) exitWith 
 if (_state) then {
     _unit setVariable [QGVAR(isHandcuffed), true, true];
     [_unit, "setCaptive", QGVAR(Handcuffed), true] call EFUNC(common,statusEffect_set);
+    call FUNC(endRadioTransmission);
+    //TFAR disallow radio usage
+    _unit setVariable ["tf_unable_to_use_radio", true];
+    //ACRE disallow radio usage
+    _unit setVariable ["acre_sys_core_isDisabled", true, true];
 
     if (_unit getVariable [QGVAR(isSurrendering), false]) then {  //If surrendering, stop
         [_unit, false] call FUNC(setSurrendered);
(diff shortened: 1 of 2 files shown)


addons/captives/functions/fnc_setHandcuffed.sqf
```diff
@@ -41,6 +41,11 @@ if ((_unit getVariable [QGVAR(isHandcuffed), false]) isEqualTo _state) exitWith
 if (_state) then {
     _unit setVariable [QGVAR(isHandcuffed), true, true];
     [_unit, "setCaptive", QGVAR(Handcuffed), true] call EFUNC(common,statusEffect_set);
+    call FUNC(endRadioTransmission);
```
jonpas: This should probably be `EFUNC(common,endRadioTransmission)` as it's locat.ed in `common` and you are doing this in `captives`. [This](http://ace3mod.com/wiki/development/coding-guidelines.html#macro-usage) should explain it.

addons/captives/functions/fnc_setHandcuffed.sqf
```diff
@@ -41,6 +41,11 @@ if ((_unit getVariable [QGVAR(isHandcuffed), false]) isEqualTo _state) exitWith
...
+    //TFAR disallow radio usage
+    _unit setVariable ["tf_unable_to_use_radio", true];
+    //ACRE disallow radio usage
+    _unit setVariable ["acre_sys_core_isDisabled", true, true];
```
jonpas: Why is TFAR's variable local and ACRE's broadcasted?

jonpas:
Should probably do the same for `setHandcuffed`.

kwinteng:
Fixed EFUNC()
added the same lines to surrender. @jonpas 
Still needs testing - will set up my test build asap

esteldunedain:
I'm not sure if this handles properly the case in which a surrendered or handcuffed player dies and respawns.

kwinteng:
I've just tested this manually, having some issues with setting up my dev build
1) the transmission cuts out when being captured
2) respawn recovers the radio usage
3) can't continuously transmit during surrendering so removing EFUNC(endtransmission) on surrendering

Glowbal:
This change does not take into account other cases where the radio might have been disabled. This would also need to properly work with `ace_unconscious` We probably need to introduce some kind of helper method for doing this, similiar to our statusEffect solution.

jonpas:
`setSurrendered`/`setHandcuffed` are called from `ace_onUnconscious` event.

Glowbal:
@jonpas yes. It clears out the surrendered state for example; https://github.com/acemod/ACE3/blob/master/addons/captives/functions/fnc_handleOnUnconscious.sqf#L26

That would reset the radio usage variables for TFAR and ACRE.

PabstMirror:
This might be a good case to use the status effect handler.
So both medical and captive don't overwrite eachother.

commy2:
wip tag, cause it doesn't work as expected atm

PabstMirror:
Closed via #8388

PR #4336, opened 2016-09-05
Fix pick up delay and IR chemlights, prevent picking up GL and mortar smoke shells

jonpas:
- Title

Initially this was meant to fix mortar shells from being picked up, but they inherit from `GrenadeBase` as well, would be small overhead to filter them. However, it seems only smoke shell is the problem, with which I don't really have a problem with being picked up and thrown further.


Glowbal:
lgtm

PR #4349, opened 2016-09-06
Add setting for medical ai

PabstMirror:
Adds setting for medical AI
- Always Run (default)
- Only Server and HC
- Disabled


BaerMitUmlaut:
:shipit:

PR #4350, opened 2016-09-06
Grenades - disable throw mode / keybind if advThrow

PabstMirror:
Neither of these should have really caused any problems, 
but this should make things cleaner when using advanced throwing.


jonpas:
:+1:

PR #4354, opened 2016-09-06
Fix stop gesture keybind

PabstMirror:
This matches the interaction menu

stop -> gesutreFreze
freeze -> freeze

### DIFF
--- a/addons/gestures/XEH_postInit.sqf
+++ b/addons/gestures/XEH_postInit.sqf
@@ -14,6 +14,9 @@ if (!hasInterface) exitWith {};
     };
 
     private _code = compile format [QUOTE('%1' call FUNC(playSignal)), _signalName];
+    if (_currentName == "Stop") then {
+        _code = compile format [QUOTE('%1' call FUNC(playSignal)), "BIgestureFreeze"];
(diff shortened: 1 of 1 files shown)


jonpas:
I am confused, [this](https://github.com/acemod/ACE3/blob/702fc75e627f899a2c2f855a3116d540d01b6545/addons/gestures/XEH_postInit.sqf#L10-L11) should handle that.

PabstMirror:
This should now match the interaction menu.
```
class GVAR(Stop) {
    displayName = CSTRING(Stop);
    statement = QUOTE([ARR_2(_target,'gestureFreeze')] call EFUNC(common,doGesture);); // BI animation - is actualls "stop" in all stances but prone

class GVAR(Freeze) {
    displayName = CSTRING(Freeze);
    statement = QUOTE(QUOTE(QGVAR(freeze)) call FUNC(playSignal));
```

jonpas:
Ahh, so when doing it with keybind it was actually calling `BIgestureStop`, therefore not doing any gesture because that doesn't exist?

Glowbal:
:+1:

PR #4355, opened 2016-09-06
Remove voting text in Arma 3 Issues list

jonpas:
- Title
- Use `-` for points


PabstMirror:
:+1:

PR #4360, opened 2016-09-08
Fixed weather for inherited CfgWorlds

BaerMitUmlaut:
- Fix weather values for maps that inherit their CfgWorlds class from another map
- Restructure and simplify the getMapData function a little bit

### DIFF
--- a/addons/weather/functions/fnc_getMapData.sqf
+++ b/addons/weather/functions/fnc_getMapData.sqf
@@ -15,7 +15,8 @@
  */
 #include "script_component.hpp"
 
-// Assume default wind values
+// Set default values
+
 // Source: https://weatherspark.com/averages/32194/Lemnos-Limnos-North-Aegean-Islands-Greece
 GVAR(WindSpeedMax) = [[8.8, 5.5], [8.8, 5], [8.6, 4.8], [7.6, 3.4], [7.0, 3.0], [7.1, 3.0], [7.5, 3.1], [8.0, 3.2], [7.6, 3.5], [7.8, 4.6], [7.9, 5.0], [8.2, 5.5]];
 GVAR(WindSpeedMean) = [4.8, 4.9, 4.6, 4.1, 3.5, 3.5, 4.3, 4.4, 4.1, 4.5, 4.5, 5.0];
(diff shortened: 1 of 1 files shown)


esteldunedain:
lgtm

jonpas:
We should also use `toLower` everywhere, casing can change from build to build of those maps (even though unlikely, but better to be safe than sorry).

PabstMirror:
`if (toLower worldName in ["fata", "Abbottabad"]) exitWith {` - opps

addons/weather/functions/fnc_getMapData.sqf
```diff
@@ -35,23 +39,42 @@ GVAR(WindDirectionProbabilities) = [
...
-// Check if the weather data is defined in the map config
-if (isArray (configFile >> "CfgWorlds" >> worldName >> "ACE_TempDay")) exitWith {
-    GVAR(TempDay)   = getArray (configFile >> "CfgWorlds" >> worldName >> "ACE_TempDay");
-    GVAR(TempNight) = getArray (configFile >> "CfgWorlds" >> worldName >> "ACE_TempNight");
-    GVAR(Humidity)  = getArray (configFile >> "CfgWorlds" >> worldName >> "ACE_Humidity");
+// Get all non inherited arrays to filter maps that inherit from Stratis/Altis/Tanoa
+private _nonInheritedArrays = configProperties [configFile >> "CfgWorlds" >> _worldName, "isArray _x", false];
+// And check if any custom non-inherited weather is defined through config and use that if so
+if ((configFile >> "CfgWorlds" >> _worldName >> "ACE_TempDay") in _nonInheritedArrays) exitWith {
+    if (isArray (configFile >> "CfgWorlds" >> _worldName >> "ACE_TempDay")) then {
```
BaerMitUmlaut: If one weather value is defined I think it's safe to assume all of them are. I don't think these checks are necessary.

jonpas: I think they are good, this function only runs on mission start anyways.

PabstMirror: I added them just to be safe.
Before there was 2 separate `isArray` checks for wind and temp.
It's possible that some map only defines one or the other which could give us empty arrays.

PR #4367, opened 2016-09-10
Update the adv med bandage configuration values.

Drofseh:
This PR updates the adv med bandage config values based on the following:

https://docs.google.com/spreadsheets/d/1qPAL4vmpotuvCx_Td8P7z0PrAQBmBtE1OyrmTLH6eco/edit?usp=sharing

Efficiency is simply based on size of wounds compared to size of bandages.

Minor wounds could be considered either small or shallow depending on wound type, with one bandage possibly able to cover more than one.
Large wounds could be either deep, wide, or in the case of velocity or puncture wounds possibly consist of both an entrance and an exit wound, this necessitates more than one bandage to treat.

Reopening chance is based on the effectiveness of the type of bandage compared to the type of wound and its size.
Reopening Max Delay is based on reopening chance.

Wounds with a low reopening chance and wounds with a high reopening chance both have a low max delay, with wounds with a reopening chance in the middle having the longest reopening max delay.

This is based on the idea that wounds with a low reopening chance are likely to be stable and not need further treatment, and that wounds with a high reopening chance are very likely to need further treatment in the near future.

Bandage (Basic) is the simplest bandage and would normally be carried by your average soldier, used for treating smaller or shallower wounds, or for temporarily patching larger/deeper wounds before the medic is able to treat the casualty.

Bandage (Elastic) and Basic Field Dressing (Quick Clot) are similar but more effective bandages and would normally be carried by medics.

Packing Bandages are a specialized bandage for treating large or deep wounds, literally being packed into cavities in order to stem serious bleeding, and would also normally be carried by medics.


commy2:
```
$ python3 tools/config_style_checker.py
Validating Config Style
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 716
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 716
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 716
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 716
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 716
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 717
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 717
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 717
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 717
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 717
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1204
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1204
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1204
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1205
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1205
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1205
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1369
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1369
ERROR: Tab detected at addons/medical/ACE_Medical_Treatments.hpp Line number: 1369
```

addons/medical/ACE_Medical_Treatments.hpp
```diff
@@ -1155,88 +1258,115 @@ class ACE_Medical_Advanced {
...
+                    reopeningMaxDelay = 1080;
                 };
                 class punctureWoundLarge: punctureWound {
-                    effectiveness = 0.4;
+                    effectiveness = 0.5;
+                    reopeningChance = 0.2;
+                    reopeningMaxDelay = 480;
                 };
-            };
+            };
```
jonpas: Untrimmed whitespace.

Drofseh:
Removed tabs from 716, 717, 1204, 1205, and 1369.

narrowsoul:
This setup has all bandages with the same efficiency for all types of wounds. Would be better than the current setup where elastic bandages are the best for some reason, but still not ideal.

PaxJaromeMalues:
Not perfect, but definitely an improvement when implimented.

Drofseh:
I've been thinking about this a little more today, in particularly the comment from @narrowsoul about different types of bandages having the same effectiveness.

Minor	2
Medium	1
Large	0.5

The reason I have the same effectiveness for the different types was based on the idea that a bandage would be a certain size, and the bandages are roughly the same size regardless of type in real like, therefore any given bandage would only be able to cover so much area on a persons body.

However, IRL you can get bandages that are different sizes, 4 inches wide, 8 inches wide, etc, and I wonder if it would be worth having each type of bandage in game have say, two sizes.

For example, have a Large Elastic Bandage and a Small Elastic Bandage with efficiencies similar to this:

Large Elastic Bandage		Small Elastic Bandage
Minor	1				Minor	1
Medium	1				Medium	0.5
Large	0.5				Large	0.25

And then also adjust reopening chance based on the wound type and the size of bandage vs the size of wound, with the idea that a larger bandage on a smaller wound is less likely to reopen.

This would then force the players to decide if they want to use their small bandages on more minor wounds, but have the small wound be more likely to reopen, or to use a large bandage on a small wound, knowing it is more likely to stay on, but also that they won't have it later to use on larger wound if they receive those.

PaxJaromeMalues:
I guess the real problem is that we currently can not use bandages in combination and are unable to select specific wounds via the medical menu.
I.e. one putting a (in your case) medium elastic bandage above an medium bruise should be effective as the elastic bandage would put even pressure onto the bruise negating a swelling (which is honestly not important). Using elastic bandages on any other wound would be pretty ineffective as it only is used to a) hold shit in place and b) apply even pressure.
But if we now have a small puncture wound, apply a small quickclot bandage to it and then wrap it with a small elastic bandage, the elastic bandage would redurce blood flow to the puncture wound, press the quikclot onto the wound opening and therefore increase chance for the wound (due to the quickclot) to stay closed, even when slightly stressed. As were simply applying an elastic bandage to the puncture wound would be highly inefficient as it only tops the wound opening, but is not closing it up like quikclot combat gauze does.
Same goes for lets say a small to medium velocity wound on an arm. Apply field dressing to it, cool. Apply elastic bandage over fielddressing super cool. Only apply elastic bandage to velocity wound, meh, not so cool (no wound closer, not very helpful with sucking up excess blood, only applies pressure).

Type of bandages and their actual use on wounds should be reviewed.
As should the factors impacting reopening be reviewed. Wounds do not only reopen based on the applied bandage type or size they do due to: how much blood escapes through the wound? Does the bandage even contain any types of hemostasis increasing materials? Is the patient moving or being moved and if how (avoid whenever possible to carry casualties when they have wounds in the areas: abdomen, chest & head or critical wounds on limbs)?

Drofseh:
@PaxJaromeMalues 
Yah, I'd love to see a system where you can to stack bandages like you do in real life.
Stuff a deep wound with a packing bandage, put a field dressing over it, and then hold it all down with an elastic wrapped around the limb or other body part.

I think it might be a little out of scope to get that complicated though, so for now I'll be happy enough if the bandage values just make a little more sense.

Drofseh:
=(

dedmen:
Reason for close? already solved in rewrite or to be solved in rewrite?

PaxJaromeMalues:
Yeah would be nice to know the reason for the thing to be closed @thojkooi

thojkooi:
This should be adressed in the rewrite.

thojkooi:
More info can be read in an upcoming blog post.

PR #4368, opened 2016-09-10
Delete "Teleport Players" module after use

654wak654:
- Make the zeus module "Teleport Players" delete itself after first use.

The "Teleport Players" module ace adds is a very useful little utility, but the module itself isn't really needed after it's used once. It just creates clutter after repeated use and has to be manually removed.

### DIFF
--- a/addons/zeus/functions/fnc_ui_groupSide.sqf
+++ b/addons/zeus/functions/fnc_ui_groupSide.sqf
@@ -128,7 +128,7 @@ private _fnc_onConfirm = {
     private _unit = effectiveCommander (attachedTo _logic);
     private _side = [west,east,independent,civilian] select (GETVAR(_display,newSide,GETVAR(_display,oldSide,0)));
 
-    [_unit, _side, _logic] call FUNC(moduleGroupSide);
+    [_unit, _side] call FUNC(moduleGroupSide);
     deleteVehicle _logic;
 };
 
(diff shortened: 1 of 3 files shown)


654wak654:
@jonpas on slack asked if there are more modules that should do this. Most of them already do it, and there are some that don't do it by design. @SilentSpike do you think the "Group Side" module could also be deleted after use?

SilentSpike:
This is a weird one, because it can be handy to leave the module attached to a plane or whatever and repeatedly teleport waves of players to the same location. However if you don't intend to use it like that then it does just create clutter.

Group side one should probably self-delete yeah.

654wak654:
Could have another checkbox saying "delete after use", and then have `_fnc_onUnload` delete that module depending on that. Maybe even save that preference to profileNamespace.

SilentSpike:
When I first wrote it I thought of that but decided it was kind of pointless since zeus has the power to delete it anyway.

In keeping with good UX design I say if people expect it to self-delete then it should behave that way.

addons/zeus/functions/fnc_ui_groupSide.sqf
```diff
@@ -128,7 +128,7 @@ private _fnc_onConfirm = {
     private _unit = effectiveCommander (attachedTo _logic);
     private _side = [west,east,independent,civilian] select (GETVAR(_display,newSide,GETVAR(_display,oldSide,0)));
 
-    [_unit, _side] call FUNC(moduleGroupSide);
+    [_unit, _side, _logic] call FUNC(moduleGroupSide);
     deleteVehicle _logic;
```
SilentSpike: It actually looks as though the logic should be getting deleted already by this function, is that not the case?

654wak654: I should probably remove this one so it fits every other module function, where the module gets deleted at the end of the moduleDERP.sqf function.

SilentSpike: The UI modules actually get deleted at the end of their respective UI function (if there's no reason to pass the module on to the further function).

654wak654: Reverted.

PR #4371, opened 2016-09-11
Fix treatment animion with no weapon equiped

### LINKED ISSUE #4366: Unit gets stuck in animation loop after checking blood pressure
**ACE3 Version:** `3.7.0` (RC6)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Unit gets stuck in animation loop after checking blood pressure under certain circumstances. One of them being the loadout of the unit should be changed on the mission start. Second is that the unit have to have his weapons on back. I packed my loadout system to the repro mission, it's a pretty small function. https://youtu.be/UQ8YfnQDXj8

**Steps to reproduce:**
- Load repro mission. 
https://mega.nz/#!v1I1GZQI!OsTn6b06vzTIe5oOlfyS3fhQTJZtjpSKw2rlvh-QgY8
- Get weapon on back.
- Check your pulse.

**Placed Modules:**
- Medical modules.
- Other modules.

**RPT log file:**
- https://gist.github.com/rebelvg/8e7d1160e697a7a837a64ad77b4fbd96

PabstMirror:
Fix #4366

If player had no weapon equiped and no pistol then
`currentWeapon` would match `handgunWeapon` and it would play bad animation

### DIFF
--- a/addons/medical/functions/fnc_treatment.sqf
+++ b/addons/medical/functions/fnc_treatment.sqf
@@ -163,7 +163,12 @@ if (currentWeapon _caller == secondaryWeapon _caller) then {
     _caller selectWeapon (primaryWeapon _caller);
 };
 
-private _wpn = ["non", "rfl", "pst"] select (1 + ([primaryWeapon _caller, handgunWeapon _caller] find (currentWeapon _caller)));
+private _wpn = switch (true) do {
+    case ((currentWeapon _caller) == ""): {"non"};
+    case ((currentWeapon _caller) == (primaryWeapon _caller)): {"rfl"};
+    case ((currentWeapon _caller) == (handgunWeapon _caller)): {"pst"};
+    default {"non"};
+};
 private _callerAnim = [_callerAnim, "[wpn]", _wpn] call CBA_fnc_replace;
 if (vehicle _caller == _caller && {_callerAnim != ""}) then {
     if (primaryWeapon _caller == "") then {


addons/medical/functions/fnc_treatment.sqf
```diff
@@ -163,7 +163,12 @@ if (currentWeapon _caller == secondaryWeapon _caller) then {
     _caller selectWeapon (primaryWeapon _caller);
 };
 
-private _wpn = ["non", "rfl", "pst"] select (1 + ([primaryWeapon _caller, handgunWeapon _caller] find (currentWeapon _caller)));
+private _wpn = switch (true) do {
```
jokoho48: why switch True? this should be the same.
```
private _wpn = switch (currentWeapon _caller) do {
    case (primaryWeapon _caller): {"rfl"};
    case (handgunWeapon _caller): {"pst"};
    default {"non"};
};
```

PabstMirror: if (currentWeapon == "") and (handgunWeapon == "") then that will pick "pst" instead of "non"

commy2: first line should be
`case (""): {"non"};`
then?

PabstMirror: it's arma; the commands might return different strings in a different case :smile:

PabstMirror:
Will close this and add to re-write

PabstMirror:
Added to #4064

PR #4376, opened 2016-09-11
Add ammo cookoff

Glowbal:
- Add ammo cookoff from ACE2

Targetting release to hide the diff. Will target master when 3.7.0 is released.


addons/cookoff/CfgCloudlets.hpp
```diff
@@ -52,3 +52,30 @@ class CfgCloudlets {
         destroyOnWaterSurfaceOffset = 0;
     };
 };
+
+class ExploAmmoExplosion {
```
commy2: tag ?

addons/cookoff/XEH_postInit.sqf
```diff
@@ -46,14 +46,21 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 
 // secondary explosions
 ["AllVehicles", "killed", {
+    if ((_this select 0) isKindOf "StaticWeapon") exitWith {};
```
commy2: just add `"StaticWeapon"` to the array where `"Man"` already is excluded. (L52/57)

addons/cookoff/XEH_postInit.sqf
```diff
@@ -46,14 +46,21 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
...
+            [(_this select 0), magazinesAmmo (_this select 0)] call FUNC(detonateAmmunition);
+        };
     };
 }, nil, ["Man"]] call CBA_fnc_addClassEventHandler;
 
 // blow off turret effect
 ["Tank", "killed", {
     if (GVAR(enable)) then {
-        (_this select 0) call FUNC(blowOffTurret);
+        if (random (1) >= 0.85) then {
```
commy2: if you do it like this:
`random 1 < 0.15`
it's easier to see what the probability for something to happen is

commy2:
Now we have `FUNC(secondaryExplosions)` and `FUNC(detonateAmmunition)`. Could these be merged somehow? They do the same thing in theory.

SilentSpike:
Perhaps `detonateAmmunition` should simply have a chance to cause secondary explosions for explosive ammo types (instead of firing them off), removing the need for the secondary explosions function.

Then if the `enableAmmoCookoff` setting is disabled they can always cause secondary explosions.

Glowbal:
> Perhaps detonateAmmunition should simply have a chance to cause secondary explosions for explosive ammo types (instead of firing them off)

It does do that at the moment already - it fires some off, others will detonate within / near the vehicle.

SilentSpike:
So we just allow it to also create the `"SmallSecondary"` used in `fnc_secondaryExplosions` and the function is totally obsolete?

Glowbal:
We could, but we might want to look at fuel for those as well.

addons/cookoff/CfgCloudlets.hpp
Glowbal: `ExploAmmoExplosion` is from vanilla

SilentSpike:
So I've tested and it's totally possible to prevent ammo boxes from sinking into the ground by changing `destrType` to `DestructDefault` from `DestructBuilding`.

The downside is that after death the box is just left sitting there with no visual indication that it has been destroyed. Perhaps the current method of keeping the box technically alive while cook off occurs is preferable.

Going to play around some more with other types I can find, `DestructWreck` sounds interesting.

SilentSpike:
That one also does nothing. I suggest we keep it as is - as long as cookoff can be disabled per-box it shouldn't be problematic for mission makers since they can just disable it if a target box needs to be instantly destroyed or w/e.

commy2:
Couldn't you just delete the box after you are done with the fireworks?

SilentSpike:
I suppose we could, it would be more abrupt than the whole sinking into the ground thing though. I don't want to make a solo judgement call on that one.

SilentSpike:
It would also delete the box :stuck_out_tongue_closed_eyes:  As obvious as that sounds - I think the vanilla destruction just disappears the box and it still exists somewhere in the 6th dimension. So it could be unexpected with regards to scripts/triggers and stuff.

Glowbal:
This should be ready.

addons/cookoff/functions/fnc_cookOffBox.sqf
```diff
@@ -0,0 +1,73 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_box"];
+
+if (_box getVariable [QGVAR(isCookingOff), false]) exitWith {};
+_box setVariable [QGVAR(isCookingOff), true];
+
+[QGVAR(cookOffBox), _box] call CBA_fnc_remoteEvent;
```
commy2: Needs `local` check.
Same problem as: https://github.com/acemod/ACE3/pull/4242/files

Glowbal:
@commy2 done

PR #4378, opened 2016-09-11
Disable medical AI for remote controlled AI and players

Glowbal:
- Fix medical AI forcing bandage even if the AI is remote controlled.

### DIFF
--- a/addons/medical_ai/functions/fnc_healSelf.sqf
+++ b/addons/medical_ai/functions/fnc_healSelf.sqf
@@ -13,7 +13,7 @@
 #include "script_component.hpp"
 
 // Player will have to do this manually of course
-if (isPlayer _this) exitWith {};
+if ([_this] call EFUNC(common,isPlayer)) exitWith {};
 // Can't heal self when unconscious
 if (_this getVariable ["ACE_isUnconscious", false]) exitWith {};
 // Check if we're still treating


addons/medical_ai/StateMachine.hpp
```diff
@@ -1,5 +1,5 @@
 class GVAR(stateMachine) {
-    list = "allUnits select {local _x}";
+    list = "allUnits select {local _x && {_x != ACE_player} && {!isPlayer _x}}";
```
jonpas: If it's not a player then it's also not ACE_player. Are both conditions really necessary?

PabstMirror: Zeus remote controled AI return false for `isPlayer`
could also use `ace_common_fnc_isPlayer`

Glowbal: if you are remote controlling a unit, your player unit is not `ACE_player` but your remote controlled AI is.

jonpas: Ah, I see.

esteldunedain:
I'm not sure this is enough.

 iirc CBA-FSM compiles a list of units to run, and then executes actions for one unit a time over the course of many frames, even seconds; when the list has been fully iterated then it is recalculated.

 What if you remote control a unit after the list was compiled but before that unit FSM is computed?

BaerMitUmlaut:
This fix will not work. The state machine specifically runs *with* the players unit, so player will get treated by an AI medic when he's injured (for example in single player when you control a group of AI units).

Instead the fix should be applied here: https://github.com/acemod/ACE3/blob/master/addons/medical_ai/functions/fnc_healSelf.sqf#L16
The `ace_common_fnc_isPlayer` function PabstMirror mentioned should probably be used.

BaerMitUmlaut:
Fixup magic! :shipit:

BaerMitUmlaut:
Fixup magic! :shipit:

PR #4380, opened 2016-09-11
Add ammo box cookoff

SilentSpike:
- Add an option to enable ammo boxes to cook off using the new ammunition detination
- Currently only occurs when hit with explosive ammo, should look into things like tracers and the incendiary grenade
- Uses the vanilla ammo box particle effects for smooth transition to actual destruction and consistent look

Note that ammo boxes must be kept "alive" until cook off is finished, because their destruction simulation makes them sink and disappear. Hence cook off is initiated from the handle damage event - unlike vehicles where their cook off occurs after destruction has occurred.

### DIFF
--- a/addons/cookoff/ACE_Settings.hpp
+++ b/addons/cookoff/ACE_Settings.hpp
@@ -6,6 +6,12 @@ class ACE_Settings {
         value = 1;
         typeName = "BOOL";
     };
+    class GVAR(enableAmmobox) {
+        displayName = CSTRING(enableBoxCookoff_name);
+        description = CSTRING(enableBoxCookoff_tooltip);
+        value = 1;
+        typeName = "BOOL";
+    };
     class GVAR(enableAmmoCookoff) {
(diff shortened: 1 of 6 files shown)


SilentSpike:
Also, I realise the settings perhaps overlap a little bit as it currently stands. I just threw one in there while it's WIP that can be cleaned up later when we decide how to handle those.

SilentSpike:
Some thoughts I had today:

Would be nice if we could somehow stop ammoboxes from sinking into the ground on death. Would allow us to destroy them as part of the cook off process (better for triggers, EHs, etc.)

If that's not possible I reckon maybe we should add a bit of code to the handle damage EH to instantly destroy them (skipping cook off) if hit by enough damage in a single shot (you'd imagine dropping a bomb on a box would obliterate it in one go).

Glowbal:
I think both of those ideas are nice - we should put those on the backlog for a future release. I don't think it is necessary to do those right away.

PR #4383, opened 2016-09-12
Add object attribute to disable cookoff for individual vehicles

Glowbal:
When a vehicle is destroyed / killed, it will start secondairy explosives. If a mission has a single vehicle that the misson maker wishes to be destroyed, right now it will start secondairy explosives (with ammo cook off (#4376) it becomes a real party). This PR allows mission makers to disable cook off for individual vehicles.

This PR also adds initial framework documentation for cook off.

### DIFF
--- a/docs/wiki/framework/cookoff-framework.md
+++ b/docs/wiki/framework/cookoff-framework.md
@@ -1,0 +1,29 @@
+---
+layout: wiki
+title: Cook Off Framework
+description: Explains the Cook off system for developers & mission makers.
+group: framework
+order: 5
+parent: wiki
+mod: ace
+version:
+  major: 3
+  minor: 7
+  patch: 0
+---
+
+## 1. Disabling / Enabling Cook off for individual vehicles
+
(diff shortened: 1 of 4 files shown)


Glowbal:
This could probably also go in with 3.7.1.

docs/wiki/framework/cookoff-framework.md
```diff
@@ -0,0 +1,29 @@
+---
+layout: wiki
+title: Cook off Framework
+description: Explains the Cook off system for developers & mission makers
```
jonpas: Missing closing point.

docs/wiki/framework/cookoff-framework.md
```diff
@@ -0,0 +1,29 @@
+---
+layout: wiki
+title: Cook off Framework
```
jonpas: `Off` to make it look nice?

commy2:
I like it, but this definitely should be a 3den object attribute too

PR #4385, opened 2016-09-13
End path drawing when passing through Geom/Fire

PabstMirror:
Ends grenade arc drawing when throw will collide with the GEOM or FIRE LOD.

Unfortunately many CUP buildings have bulletproof windows that do not pass grenades.
There are also other cases where we could show a throw as being safe when it will be blocked such as A3 Gratings.
However this also could introduce warnings when the shot will be fine.

But I believe the worst thing we can do is to tell the user that a shot will be safe when it could fail and kill their team.  Warning about a "false negative" is far less critical.

- Neither of these shots are possible, but would have been shown as safe:

- This shot is actually fine, but we still show a warning:

### DIFF
--- a/addons/advanced_throwing/functions/fnc_drawArc.sqf
+++ b/addons/advanced_throwing/functions/fnc_drawArc.sqf
@@ -47,11 +47,15 @@ for "_i" from 0.05 to 1.45 step 0.1 do {
 
     if (_newTrajASL distance (getPosASLVisual ACE_player) <= 20) then {
         if ((ASLToATL _newTrajASL) select 2 <= 0) then {
-            _cross = 1
+            _cross = 1; // 1: Distance Limit (Green)
         } else {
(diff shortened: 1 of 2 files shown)


PabstMirror:
Also, we can get info on the actual hitpoint with 
```
private _hitpoint = [cursorObject, "FIRE"] intersect [ASLtoAGL _prevTrajASL, ASLToAGL _newTrajASL];
```
Which should give "Glass_2_hide" which would match:
```
class Glass_2_hitpoint {
    armor = 0.001;
    material = -1;
    name = "Glass_2";
    visual = "Glass_2_hide";
    passThrough = 0;
    radius = 0.4;
    convexComponent = "Glass_2_hide";
```

But I see no difference between A2 and A3 building configs. I believe the problem is with the model material, but I am not sure.

jonpas:
This looks good to me, overall I'd say it's an improvement, "false negative" is indeed better. I will add documentation for that in this PR, so marking with WIP.

Issue #4386, opened 2016-09-13
Advanced Ballistics cuts FPS in half


gienkov:
**Arma 3 Version:** `1.62`
**CBA Version:** `3.0.0`
**ACE3 Version:** `3.6.2`

**Mods:**
- `@CBA_A3`
- `@ace`
- '@CUP_Terrains'

**Description:**
- Having Adv. Bal. module enabled will cut your FPS in half even when not shooting.

**Steps to reproduce:**
- Place simple NATO rifleman on map
- Run preview and check your FPS

![](http://i.imgur.com/UHqlsbR.jpg)

- Place Adv. Bal. module and enable it
- Run preview and check your FPS and observe how low is it now

![](http://i.imgur.com/DgxzttT.jpg)

**Where did the issue occur?**
- Dedicated / Editor (Singleplayer) / Editor (Multiplayer)

**Placed Modules:**
- Advanced Ballistics

**RPT log file:**
- Nothing related inside

commy2:
RPT

gienkov:
http://pastebin.com/K25Asn5X

commy2:
Cannot confirm.

!<http://i.imgur.com/ULr6qPa.jpg>

!<http://i.imgur.com/FhH44CO.jpg>

gienkov:
Are you sure you've enabled it inside the module? It's off by default.

commy2:
Yes. Set every bar to "enabled"

gienkov:
Can you try it on one of CUP maps?

gienkov:
Altis also works for me:

- OFF: 50 FPS
- ON: 35-40 FPS

gienkov:
And CUP Takistan:

- OFF: 43 FPS
- ON: 30 FPS

commy2:
I don't have CUP and can't DL a pack that size atm.

gienkov:
Try it on Altis and closer to populated area.

commy2:
![http://i.imgur.com/kZqL059.jpg](http://i.imgur.com/kZqL059.jpg)

![http://i.imgur.com/lYzikNV.png](http://i.imgur.com/lYzikNV.png)

![http://i.imgur.com/Fod7Ikk.jpg](http://i.imgur.com/Fod7Ikk.jpg)

I guess it's to be expected that it drains frames. That's why you can turn it off.

PabstMirror:
AB has a terrain processor that runs at the beginning of a mission.
[...]

PR #4391, opened 2016-09-13
Add AB Terrain Extension Logging

PabstMirror:
Ref #4386

Adds 1 or 2 lines of debug info for the advanced ballistics terrain extension.
I don't think there is a problem, but this could help diagnose any possible issues.


addons/advanced_ballistics/functions/fnc_initializeTerrainExtension.sqf
```diff
@@ -22,6 +22,7 @@ _initStartTime = CBA_missionTime;
 _mapSize = getNumber (configFile >> "CfgWorlds" >> worldName >> "MapSize");
 
 if (("ace_advanced_ballistics" callExtension format["init:%1:%2", worldName, _mapSize]) == "Terrain already initialized") exitWith {
+    ACE_LOGINFO_1("Terrain already initialized [world: %1]", worldName);
```
Glowbal: Should be an error

PabstMirror: Terrain loading is persistent between missions.
So if you load it once, it won't have to load again.

PR #4398, opened 2016-09-14
Update ACE_Medical_Treatments.hpp

narrowsoul:
Multiple changes to dressing effectiveness and reopening chance to more realistically reflect the effectiveness of various dressings. A lot of the changes are just swapping of effectiveness between different dressing types, with some minor tweaks. None of the reopening delay values have been changed.

The basic principles for this change are as follows:

Quickclot should be the most effective dressing for stemming bleeding in puncture and velocity wounds, as that is their real life purpose and the only dressing which includes chitosan for accelerated clotting.

Elastic bandages should have relatively negligible effects for stopping bleeding, but should be excellent for keeping wounds from reopening. If things work as I hope, people should be able to use the appropriate dressing for a wound until the bleeding is controlled, and secure the dressing with an elastic bandage. Most other dressings should be pretty likely to reopen without being secured with an elastic bandage.

New values here, feedback would be greatly appreciated:

https://drive.google.com/open?id=0B3qy6tOzwzmZT0FoeVhLWmJvVUU

### DIFF
--- a/addons/medical/ACE_Medical_Treatments.hpp
+++ b/addons/medical/ACE_Medical_Treatments.hpp
@@ -686,10 +686,10 @@ class ACE_Medical_Advanced {
     };
     class Treatment {
         class Bandaging {
-            // Field dressing is normal average treatment
-            // packing bandage is average treatment, higher reopen change, longer reopening delay
-            // elastic bandage is higher treatment, higher reopen change, shorter reopen delay
-            // quickclot is lower treatment, lower reopen change, longer reopening delay
+            // Field dressing is a basic dressing
+            // packing bandage is used for packing wounds, high reopen chance, short reopen delay
+            // elastic bandage is for securing dressing, low reopen change, high reopen delay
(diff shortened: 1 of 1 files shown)


narrowsoul:
According to some folks on Slack, this setup would require a change to the code that would have the reopening chance apply on the last bandage applied, rather than the last bandage used to close a wound. I unfortunately have zero coding experience so all I can do is ask that someone who does makes that change.

bux578:
Ref: #4367

narrowsoul:
In regards to #4367, I'm of the opinion that it would only be marginally better than the current values since all of the dressings have the same effectiveness for all types of wounds, leaving us with the same issue of making a varied medical loadout and thinking carefully about treatment useless.

narrowsoul:
I'm on the go right now but when I get a chance I'll look into using the author of #4367's reopening values since he seems to have put more effort into the reopening values, and I focused more heavily on effectiveness values.

commy2:
#4064

PR #4402, opened 2016-09-15
Medical Treatment Config Changes

narrowsoul:
- Update the dressing config values to be more realistic than the current system of elastic2win.

This config could be used in the interim before my other PR that would use elastic bandages to secure dressings is workable.

In 3.7.0, elastic bandages are the best for pretty much everything. This seeks to make all the bandages useful for different wounds, as they are in real life and have been in previous versions of ACE3.

Here are the values in a table for easy assessment, none of the reopen delay numbers have been changed from 3.7.0. 

https://drive.google.com/open?id=0B3qy6tOzwzmZZUx3RGNPQU9OU3c

Feedback would be particularly useful for the reopening values, they are somewhat of a placeholder at this point, but they should be perfectly fine for public release as-is.

### DIFF
--- a/addons/medical/ACE_Medical_Treatments.hpp
+++ b/addons/medical/ACE_Medical_Treatments.hpp
@@ -686,10 +686,10 @@ class ACE_Medical_Advanced {
     };
     class Treatment {
         class Bandaging {
-            // Field dressing is normal average treatment
-            // packing bandage is average treatment, higher reopen change, longer reopening delay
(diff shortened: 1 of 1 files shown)


bux578:
What's the difference to #4402

narrowsoul:
You can compare the sheets I linked, this one has more reasonable reopening values since it's not possible to use elastic bandages to "secure" dressings on an already closed wound. #4402 is on the backburner until the code relating to reopening is changed.

commy2:
#4064

PR #4408, opened 2016-09-15
Debugging tool to watch variables on main display

PabstMirror:
Adds a dev function to easily watch multiple variables on the main display.
For numbers is can show delta change, and ratio if given min/max.
Supports any variable type.

Example showing fatigue values:

I often run out of debug watch windows and this is live updating; so you can use it to easily check vars on `cursorObject`.


esteldunedain:
Awesome

PR #4410, opened 2016-09-16
Draw Straight Lines with maptool

### LINKED ISSUE #4153: Request: Bring straight Line Drawing back
**Arma 3 Version:** 1.62 Stable 
**CBA Version:** `3.0.0` stable 
**ACE3 Version:** `3.6.2` stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Alt click straight lines are gone or at least not working, would love to see them back in the game since they are such a handy tool, the tool in arma atm is not that good to draw straight lines with and you will need it for land navigation :)

PabstMirror:
Close #4153 

As you draw lines, if you draw inside of the maptools it will snap your mouse outward to the edge.
As you draw on the edge it records the points on the line.
On finishing drawing, the polyline is converted into a rectangle.

Video has debug on it, the red dots are the corners of the map tools, blue dots are points on the line that we've draw on that are closest to the 2 corners.

https://www.youtube.com/watch?v=hXQbeKrAffg

2 downsides:
`setMousePosition` fails if resolution scaling is not 100% (keep cursor centered has same problem)
To delete you need to hover in the center of the line, it's somewhat forgiving but it is different than the normal poly lines.

Should be good, tagging wip as I still need to test in mp. Also very useful for clean tic tac toe boards.


esteldunedain:
Brilliant

PabstMirror:
Working good in mp tests.
Just like all markers in 1.66 they show the user who created it
(at the midpoint; same area as the were you need to hover to delete)
image

addons/maptools/functions/fnc_drawLinesOnRoamer.sqf
```diff
@@ -0,0 +1,143 @@
...
+_roamerPosY + (-sin GVAR(mapTool_angle)) * DIST_LEFT_TO_CENTER_PERC * _roamerWidth + (cos GVAR(mapTool_angle)) * DIST_BOTTOM_TO_CENTER_PERC * _roamerWidth,
+0];
+
+private _fnc_Distance = { // Get distance point _p is from a line made from _a to _b (uses 3d array commands, but z should be 0)
+    params ["_a", "_b", "_p"];
+    _n = _b vectorDiff _a;
+    _pa = _a vectorDiff _p;
+    _c = _n vectorMultiply ((_pa vectorDotProduct _n) / (_n vectorDotProduct _n));
+    _d = _pa vectorDiff _c;
+    sqrt (_d vectorDotProduct _d);
```
commy2: `private` ?
might also want more descriptive variable names

JohnnyShootos:
Nice work! Any chance of extending this functionality to be able to draw circles or arcs?

PabstMirror:
As far as I know there are no commands to modify the new polyLine marker type;
which is what you get when free drawing.
Drawing straight lines was possible because we can just make it a rectangle marker.

But I don't know anyway to only draw parts of a arc;
or how we would implement the user drawing part of it.

JohnnyShootos:
would there be a way to say lock the drawing cursor to the maptool when rotating the map tool to get a fixed radius from the center as the polyline is drawn?

PR #4422, opened 2016-09-17
Skip incindinary effects if projectile is null

PabstMirror:
skips creating fire/effects at [0,0,0] if projectile is dead / deleted

### DIFF
--- a/addons/grenades/functions/fnc_incendiary.sqf
+++ b/addons/grenades/functions/fnc_incendiary.sqf
@@ -34,6 +34,8 @@
 
 params ["_projectile", "_timeToLive"];
 
+if (isNull _projectile) exitWith {TRACE_1("null",_projectile);};
+
 private _position = position _projectile;
 
 // --- fire


commy2:
when would that happen?

BaerMitUmlaut:
Base protection scripts and stuff.

commy2:
then this check should be done one frame later. order or event handler execution is not guaranteed

PabstMirror:
This is already ~3 seconds after the fired event

commy2:
https://github.com/acemod/ACE3/blob/32b7ef7b8560c0a9f227bebc82f39ef3936891fc/addons/grenades/functions/fnc_throwGrenade.sqf#L59

you are right! very convenient.

commy2:
target `release` for 3.7.1 ?

PabstMirror:
I think master has most of our 3.7.1 prs

commy2:
idk what @Glowbal's plans for this are

PR #4425, opened 2016-09-18
auto add blown off turrets to curator

commy2:
- make it so that blown off turrets are added to all curators that own the wreck
- it also adds icons to the turrets. those huge ass question marks would be annoying, so I just use the corresponding tanks icon
- also OCD formatting


commy2:
hmm. For some reason I cannot move or delete the turret. Zeus says: "INSUFICIENT RESOURCES". What does it mean with this?

commy2:
must've been a fluke. I can delete them just fine now.

PR #4429, opened 2016-09-18
DLEGION-medical-settings

DLEGION:

this changes are part of a bigger "fix" for the trouble of "18 shot to limbs and soldier still alive and conscious". 
other fixes will include 2 other files : fnc_determineIFfatal.sqf and fnc_handleDamage_Advanced.sqf .
this will set the medical system to run as advanced by default and some other minor settings to allow anyone to use medical equipment if he has the personal ability of do so
(and not be forced to use a specific in-game character to do so, to not start a war-for-the-slot).

### DIFF
--- a/addons/medical/ACE_Settings.hpp
+++ b/addons/medical/ACE_Settings.hpp
@@ -3,7 +3,7 @@ class ACE_Settings {
         category = CSTRING(Category_Medical);
         displayName = CSTRING(MedicalSettings_level_DisplayName);
         description = CSTRING(MedicalSettings_level_Description);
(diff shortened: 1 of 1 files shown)


BaerMitUmlaut:
Simply changing default settings does not fix anything and is a terrible idea because it will screw up *every* mission that relied on these default settings. You can't just make everyone use advanced medical, why do you think we provide these options in the first place?

If you believe these settings are better, feel free to default to them on your server using our [settings framework](http://ace3mod.com/wiki/framework/settings-framework.html).

PR #4433, opened 2016-09-18
DLEGION-medical-handleDamageAdvanced

DLEGION:
- Respect the [Development Guidelines](http://ace3mod.com/wiki/development/)

this changes are part of a bigger "fix" for the trouble of "18 shot to limbs and soldier still alive and conscious".
other fixes will include 2 other files : ace_settings.hpp and fnc_determineIfFatalsqf .
this will set the medical system so that a shot to an arm or a leg will make you fall unconscious in pain. you can still be saved, but you need help. small / weak calibers can still injure you without causing unconsciousness.

this is FOR SURE not the best solution, but is all i can make (i hope the great ACE3 devs fix this), anyway the result is actually realistic...a wounded soldier, hit by 2-3 glancing shots will still be able to fight (despite bleeding and in pain), but a well placed hit, even on a 9mm handgun, will make you fall unconscious in pain, needing for help.


DLEGION:
Was this at least considered ?

sargken:
They are in the process of reworking the entire medical system.

ulteq:
> Was this at least considered ?

Yes, those things will be addressed in the medical-rewrite: https://github.com/acemod/ACE3/pull/4064

commy2:
#4064

PR #4435, opened 2016-09-18
Add 'uniforms' Component

654wak654:
- Add a small component to remove side restrictions from all vanilla uniforms.


654wak654:
Script used for generating CfgVehicles: https://gist.github.com/654wak654/130c456633d79c396f5c03ddc93ded24

PabstMirror:
When making loadouts we often use whats looks good and don't care about side.  So there is the awkward situation where you if you take off your uniform you can't put it back on.

I think BIS's reasoning for locking uniforms was that AI are still going to do magic target IDing.

So this won't let you do "undercover" missions or let you kill a guard and take his uniform and then sneak into a base undetected (for AI).

Against humans these tactics could work, which is good but could also possibly encourage "cheap" tactics?  It isn't clear but I think using enemy uniforms may be a wacrime.  But it is something that can and has happened, so I guess I don't see any reason to allow the game to restrict it.

jonpas:
> I think BIS's reasoning for locking uniforms was that AI are still going to do magic target IDing.

Their reasoning is one of those war conventions, they had to make the restrictions (if you remember there was no restriction in Alpha/Beta).

I think it's perfectly reasonable to do that, it's really annoying.

commy2:
this should go into ace optionals. it would destroy TVT missions like "Coopetition" (livestream yesterday)

bux578:
We have somthing similar for every TvT event we're doing.
No complaints

jonpas:
> it would destroy TVT missions like "Coopetition" (livestream yesterday)

I don't see how it would destroy it?

addons/uniforms/$PBOPREFIX$
```diff
@@ -0,0 +1 @@
+z\ace\addons\uniforms
```
Glowbal: missing new line

commy2:
enemies taking uniforms of friendlies making them indistinguishable from friendlies.
it promotes a playstyle that is not desired by everyone.

addons/uniforms/$PBOPREFIX$
654wak654: It's a PBOPREFIX ?

654wak654:
Of course it's possible for that to happen in TvT games, but it's also possible for it to happen IRL. You should deal with players who behave like that in some other way than making it impossible. We could just dump `vehiclelock` and make vehicles bound to factions too, stealing a vehicle can be more effective than stealing a uniform after all.

With that said, I have no problem with moving this to optionals or acex, but personally feel like this fix should be a default thing. Specially since it's already a popular piece of code in many private/public unit repositories.

Cuel:
I'd guess there's public servers running ACE out there and I somewhat doubt they'd like this as the default behaviour

jonpas:
Yes, let's put this into optionals @654wak654, I can see why this wouldn't be wanted as default behaviour.

654wak654:
Like I mentioned in slack, I don't think the optionals folder should have components like this anymore now that we have ACEX. Sooo ACEX?

jonpas:
No, because it's a config change and still can't be toggled in-game, so it would have to be an optional in ACEX as well and we don't want optionals in both mods.

esteldunedain:
I think moving this to ACE/Optionals is the best solution. This is one of those PBOs that some communities will love, while the general online public will probably not.

jonpas:
You are still missing https://github.com/acemod/ACE3/pull/4435#discussion_r79310748

I am thinking a name like `remove_uniform_restrictions` might be more suitable, since it's now in optionals, previously I thought things could be added to it some day but it won't happen now, and that name would probably be better for self-explanatory reasons.

optionals/uniforms/functions/fnc_exportConfig.sqf
```diff
@@ -0,0 +1,53 @@
...
+    if !(_baseClass in (_modifyClasses apply {_x select 0})) then {
+        _baseClasses pushBackUnique _baseClass;
+    };
+    false
+} count (
+    ("!isNull (_x >> 'modelSides') &&" +
+    "{!(getArray (_x >> 'modelSides') isEqualTo [6])} &&" +
+    "{!(getArray (_x >> 'modelSides') isEqualTo [0,1,2,3])}")
+    configClasses (configFile >> "CfgVehicles")
+);
```
commy2: this looks confusing. what about this method?
```sqf

private _fnc_addClasses = {
    private _sides = getArray (_x >> "modelSides");

    if (!(_sides isEqualTo [6]) && {!(_sides isEqualTo [0,1,2,3])}) then {
        private _modifyClass = {
            if (!isNull (_x >> "modelSides")) exitWith {_x};
        } forEach configHierarchy _x;

        private _baseClass = inheritsFrom _modifyClass;
        _modifyClasses pushBackUnique [_modifyClass, _baseClass];

        if !(_baseClass in (_modifyClasses apply {_x select 0})) then {
            _baseClasses pushBackUnique _baseClass;
        };
    };

};

"_x call _fnc_addClasses; false" configClasses (configFile >> "CfgVehicles");
```

654wak654: Using `configClasses` for something other than filtering? B-but you can crash the game!!

commy2: Yeah. I think there is a `isNull` missing, but it looks cleaner. and does only iterate through it once for what it's worth.

654wak654:
Forgot to take care of this PR. What is it waiting for, the component name?

jonpas:
A newline in pboprefix.

654wak654:
That's done here: https://github.com/acemod/ACE3/pull/4435/commits/a5c7ea86a7525db5d49de9d6d60b32cebc4e641b

jonpas:
Ah, then this https://github.com/acemod/ACE3/pull/4435#issuecomment-248366402 (second paragraph).

commy2:
You mean my comment, @jonpas? I think it can be merged even without that proposal. It was just a suggestion, but the current code is fine too.

jonpas:
Nah, I mean rename to `remove_uniform_restrictions`.

654wak654:
I really feel like `remove_uniform_restrictions` is too excessive. Maybe just `uniformsides`?

jonpas:
I think it describes exactly what the PBO does to the end user.

654wak654:
It does but does it have to?

jonpas:
As we have "noactionmenu" and "nocrosshair", maybe "nouniformrestrictions" might be good. "limit" could mean something else. I am fine with either though, as long as it describes more what it does.

optionals/nouniformrestrictions/functions/fnc_exportConfig.sqf
```diff
@@ -0,0 +1,52 @@
...
+    _modifyClasses pushBackUnique [_modifyClass, _baseClass];
+    if !(_baseClass in (_modifyClasses apply {_x select 0})) then {
+        _baseClasses pushBackUnique _baseClass;
+    };
+    false
+} count (
+    ("!isNull (_x >> 'modelSides') &&" +
+    "{!(getArray (_x >> 'modelSides') isEqualTo [6])} &&" +
+    "{!(getArray (_x >> 'modelSides') isEqualTo [0,1,2,3])}")
+    configClasses (configFile >> "CfgVehicles")
```
Glowbal: Can we filter out uniforms that inherit their `modelSides` value from their base class?

jonpas: `configProperties` supports that IIRC.

PR #4440, opened 2016-09-19
port litter to treatment

commy2:
- port "litter" from "medical" to "medical_treatment"
- renames functions to use "litter" as a tag (`fnc_createLitter.sqf` -> `fnc_litterCreate.sqf`)
- formatting (camel case, deleting superfluous parenthesis, space before every control structure)
- fixed multiple problems with parsing `_litterCondition` (todo: test effects of code now)
- use params to handle input validation, remove some superfluous checks (unreachable code)
- add server only event handler only on server
- switch `getPosATL` and `setPosATL` to use `ASL` for small performance boost

todo:
- [x] testing
- [x] port settings as well?

need to confirm:
- fixes a bug that would create double the amount of litter as intended


commy2:
ref: #4430

commy2:
done, works, mergeee

PR #4447, opened 2016-09-21
Handle RHS impact grenades (RGN/RGO)

PabstMirror:
Quick version of how RHS impact grenades work:
- magazine ammo is type `rhs_ammo_rgn`
- after creation rhs's script deletes and recreates as `rhs_ammo_rgn_sub` with same position and velocity
- on impact, it's deleted again and recreated as `rhs_ammo_rgn_exp`

Even without manually priming, there is delay between "pulling the pin" and throwing.
In that time the ammo is deleted and replaced.
User throws nothing and a live grenade is dropped at their feet.

Adds a simple config `ace_advanced_throwing_replaceWith` that will replace one ammo with another.

Base grenade ammo works fine, but it loses the scripted impact.

### DIFF
--- a/optionals/compat_rhs_afrf3/CfgAmmo.hpp
+++ b/optionals/compat_rhs_afrf3/CfgAmmo.hpp
@@ -163,6 +163,8 @@ class CfgAmmo {
         ace_frag_force = 1;
     };
     class rhs_ammo_rgn: rhs_ammo_rgn_base {
+        // RGN is scripted grenade that deletes itself, which will break advanced throwing, replace with it's base
+        ace_advanced_throwing_replaceWith = "rhs_ammo_rgn_base";
         ace_frag_enabled = 0;
         ace_frag_skip = 1;
         ace_frag_force = 0;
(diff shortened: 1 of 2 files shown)


optionals/compat_rhs_afrf3/CfgAmmo.hpp
```diff
@@ -163,6 +163,8 @@ class CfgAmmo {
         ace_frag_force = 1;
     };
     class rhs_ammo_rgn: rhs_ammo_rgn_base {
+        // RGN is scripted grenade that deletes itself, which will break advanced throwing, replace with it's base
+        ace_advanced_throwing_replaceWith = "rhs_ammo_rgn_base";
```
jonpas: `EGVAR(advanced_throwing,replaceWith)`

commy2:
>Base grenade ammo works fine, but it loses the scripted impact.
Is there no way to fix that? Feels to me like we are breaking their grenade...

PabstMirror:
Something like this might be possible
- after throwing start a pfeh that watches current throwable
- if it's deleted, scan for the known replacement ammo
- replace the `activeThrowable` var with the new ammo

But In any case I'd like to get this merged semi-quick as it is a rather nasty bug.

jonpas:
Why can't we handle this the same as IR chemlights though? Instead of doing that in chemlights component, do it in advanced throwing component and let the original script actually modify it?

PabstMirror:
I'm going to merge this for now to fix the bug.

@jonpas - modifying the script would mean coping RHS's script over to ace, which is going to get messy if they update it

another problem with detecting the replacement with a pfeh is that there is no guarantee deletion and creation happen in the same frame

jonpas:
I was thinking of rather calling their script, but that's another thing to bother with, which we really shouldn't.

PR #4448, opened 2016-09-21
Show damage on repair items interaction point

### LINKED ISSUE #4333: [REQUEST] Add Damage Indicator for Wheels and some ideas
**Arma 3 Version:** `1.62` stable
**CBA Version:** `Latest` stable
**ACE3 Version:** `3.6.2` stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
I and many people gets angry when see indicator in vehicle (Damaged Wheel Orange) after a fight we dont know how to chceck what wheels are still usable, idea is to add interaction "Check Wheel" and it will say its damaged, or without air in it (or something) next idea is for my unit, we will probalby Delte All HUD when 3.7.0, left only from ace, is it possible to add indicator to see what Fire Mode i have? that Custom UI sometimes get buggy with host so it will be very helpful ("Changed to Auto" or "Changed to Single")

PabstMirror:
Close #4333

Gives feedback on the damage of a repair item (wheel / track)
We don't have damaged models for these, so there is no way to tell a destroyed wheel from a good one.

Simple white->yellow->red scale; similar to medical

edit: dots are too big, I'll shrink em down to match the normal dot


jonpas:
Is it just me or is the main interaction point icon way bigger?

commy2:
I think it's supposed to be the shape of the wheel.

We should additionally look for a good rubber mosaic texture, so we can actually show if the thing is damaged via material.

commy2:
I tried adding the damage textures to these things in that ^ PR. But as usual when dealing with those things, it didn't work. Maybe @jokoho48 can check? I don't know anyone else who ever got these to work.

commy2:
Nvm. Works. I think we should this use additionally anyway though.

addons/repair/functions/fnc_modifyInteraction.sqf
```diff
@@ -0,0 +1,28 @@
...
+ */
+#include "script_component.hpp"
+
+params ["_target", "_player", "_args", "_actionData"];
+TRACE_4("params",_target,_player,_args,_actionData);
+
+// Interaction dots numbered 0..8, white to red. 
+// Convert damage to number (rounding up), using 0.8 as max
+
+private _fileName = format [QPATHTOF(ui\damage_%1_ca.paa), ceil (linearConversion [0, 0.8, damage _target, 0, 8, true])];
```
PabstMirror: My thinking was that damage above 0.8 always showed up as a destroyed wheel on a car, but that doesn't seem to be the case.

I like the rounding up as it will let you always see the damage between a damaged wheel and a perfect one.

commy2: ok

john681611:
Great Work! Just some feedback shooting the wheel object does not change the state of the wheel so i can unload a full mag into wheel on the ground and nothing will happen. The only other item of feedback is that can this be expanded to on the vehicles as well. I have no idea which wheel is in bad shape

PR #4449, opened 2016-09-22
Fix throwing from map screen - global XEH

PabstMirror:
Fixes 2 minor issues with advanced throwing
- If advanced throwing is up, and player brings up map; clicking on the map would throw a grenaded
 Now exits throw mode

- Custom firedXeh event handler was only run on clients
Renamed to XEH_postInit and moved to above the `!hasInterface` line.

### DIFF
--- a/addons/advanced_throwing/XEH_postInit.sqf
+++ b/addons/advanced_throwing/XEH_postInit.sqf
@@ -1,5 +1,8 @@
 #include "script_component.hpp"
 
+// Fired XEH
+[QGVAR(throwFiredXEH), FUNC(throwFiredXEH)] call CBA_fnc_addEventHandler;
+
 // Exit on HC
 if (!hasInterface) exitWith {};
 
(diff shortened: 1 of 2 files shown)


addons/advanced_throwing/XEH_postInit.sqf
```diff
@@ -63,6 +66,13 @@ GVAR(ammoMagLookup) = call CBA_fnc_createNamespace;
     [_this select 1, "Player changed"] call FUNC(exitThrowMode);
 }] call CBA_fnc_addPlayerEventhandler;
 
+["visiblemap", {
```
jonpas: `visibleMap`

addons/advanced_throwing/XEH_postInit.sqf
```diff
@@ -63,6 +66,13 @@ GVAR(ammoMagLookup) = call CBA_fnc_createNamespace;
...
+    if (visibleMap && {ACE_player getVariable [QGVAR(inHand), false]}) then {
```
jonpas: EH already passes `_isMapShown` as second param.

addons/advanced_throwing/XEH_postInit.sqf
```diff
@@ -63,6 +66,13 @@ GVAR(ammoMagLookup) = call CBA_fnc_createNamespace;
...
+        [ACE_player, "Opened Map"] call FUNC(exitThrowMode);
```
jonpas: EH already passes `_playerUnit` as second param.

commy2: as first*

jonpas: Oops, copy/paste.

addons/advanced_throwing/XEH_postInit.sqf
PabstMirror: Adding the params call is slower that just using command and global var
0.00286 ms vs 0.00412 ms

addons/advanced_throwing/XEH_postInit.sqf
PabstMirror: fixed

PR #4450, opened 2016-09-22
add dmg materials to spare tracks and wheel

commy2:
- title
- also changes the icon of the track to resemble a track instead of a wheel

#4448


commy2:
done : )

normal:


damaged:


dead:

kripto202:
the difference between normal and damage is almost unnoticeable and it can be confusing.

commy2:
I don't add any paint. These are vanilla mosaik textures (mc.paa and cd.paa) that are used for every object.

PR #4453, opened 2016-09-23
Use waitAndExecute for fatigue loop

### LINKED ISSUE #4379: Adv. stamina does regenerate with game paused
**Arma 3 Version:** `1.62`
**CBA Version:** `3.0.0`
**ACE3 Version:** https://github.com/acemod/ACE3/commit/a9b82513f7de0e53003bad7187a466a36613da9b

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Stamina does regenerate with game paused

**Steps to reproduce:**
- Run for a while in editor to lose some or all of your stamina
- Pause the game with ESC
- Wait for a minute
- Unpause to see most of your stamina regenerated

- Editor (Singleplayer)

**Placed Modules:**
- Advanced Stamina

**RPT log file:**
- Will upload if needed

PabstMirror:
Close #4379
- Will use CBA_missionTime so it respects pause and time acceleration
- Possible performance improvements over using a PFEH

### DIFF
--- a/addons/advanced_fatigue/XEH_PREP.hpp
+++ b/addons/advanced_fatigue/XEH_PREP.hpp
@@ -5,6 +5,6 @@ PREP(getMetabolicCosts);
 PREP(handleEffects);
 PREP(handlePlayerChanged);
 PREP(handleStaminaBar);
+PREP(mainLoop);
 PREP(moduleSettings);
-PREP(pfhMain);
 PREP(removeDutyFactor);
--- a/addons/advanced_fatigue/XEH_postInit.sqf
+++ b/addons/advanced_fatigue/XEH_postInit.sqf
@@ -37,8 +37,8 @@ if (!hasInterface) exitWith {};
         }] call FUNC(addDutyFactor);
     };
 
-    // - Add main PFH -------------------------------------------------------------
-    [FUNC(pfhMain), 1, []] call CBA_fnc_addPerFrameHandler;
+    // - Add main loop at 1 second interval -------------------------------------------------------------
+    [FUNC(pfhMain), [], 1] call CBA_fnc_waitAndExecute;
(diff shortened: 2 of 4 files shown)


commy2:
We badly need a proper doc for these on the CBA wiki. So we have a place to look up which commands use simulation time and which use tick time...

BaerMitUmlaut:
Why not exit the PFH when the time has not changed? I don't see the advantage over a PFH right now.

PabstMirror:
PFEH run on diag_tickTime or your computer's clock
So to handle pausing we would have to save CBA_missionTime in a var and only run when 1 sec has passed.
Handling time acceleration would be a little harder, could do something like
```
_deltaTime = CBA_missionTime - _lastTime;
_lastTime = CBA_missionTime;
GVAR(stamina) = (_deltaTime / 1) * CHANGE_IN_STAMINA;
```

But waitAndExec runs on CBA_missionTime. So the code only runs every 1 second of in-game time no matter if it's paused or acceleration is on.

I think estel though they would be faster, with PFEH it's internally checking "is it time to run" for each PFEH every frame.
With waitAndExec they are sorted by next execution, so only the first entry needs to be checked.

BaerMitUmlaut:
Should probably rename the function then if it's no more a PFH?

esteldunedain:
Yup, waitAndExec loops are faster than PFH for things that run seldomly.

:+1: for the function name change. I was about to suggest the same yesterday but decided to give pabst a break instead (I feel a bit gilty for not contributing)

jonpas:
> I feel a bit gilty for not contributing

Reviewing and suggesting changes is also contributing. ;)
After-all, we don't have to listen to you 😆

esteldunedain:
> After-all, we don't have to listen to you

Too bad

PR #4456, opened 2016-09-24
Convert spectator to CBA settings

SilentSpike:
Believe I've done this correctly, running a function on setting change is a much welcome addition that fits nicely with the way I've structured spectator already.


SilentSpike:
This is also what had me confused while doing this: https://github.com/acemod/ACE3/pull/4024#issuecomment-230897212

What I expect to happen is: Forced settings don't appear in the `CLIENT` section, only under `SERVER` and `MISSION` (with the `Force` checkbox locked to activated under `MISSION`).

commy2:
@SilentSpike wait for the next CBA. Pabst changed some things around about this.

SilentSpike:
Tangentially related to this (conversion of the zeus settings), but is there a way to do what we currently do [here](https://github.com/acemod/ACE3/blob/0143dd40ba0d4ee97cb36fc23405d2f5586361a0/addons/zeus/functions/fnc_bi_moduleCurator.sqf#L170-L174) using the new CBA settings.

Keep in mind this is within a module function so multiple instances could run at any point in the mission (i.e. I can't easily add a function to be ran when the setting is changed since each instance of this is associated with an individual logic initialised at an arbitrary time).

PR #4457, opened 2016-09-24
remove deprecated code for 3.8.0

commy2:
- title


addons/common/functions/fnc_removeScrollWheelEventHandler.sqf
```diff
@@ -1,31 +0,0 @@
...
- * Author: commy2
- * Remove a scroll wheel event handler.
- *
- * Arguments:
- * 0: ID of the event handler <NUMBER>
- *
- * Return Value:
- * None
- *
- * Public: Yes
```
Glowbal: public but was never deprecated?

commy2: Oops. I must've forgotten it there then. However, https://github.com/acemod/ACE3/blob/0b3d38a72f6cccb9880acb03cc446b725f51ba5d/addons/common/functions/fnc_addScrollWheelEventHandler.sqf
was deprecated and this function makes no sense without the add counterpart.

PR #4459, opened 2016-09-24
remove deprecated event handlers

commy2:
- title
- targets #4457 because not sure if we want it or if we want to keep these around a bit longer


commy2:
note: kept them in the medical components to prevent merge conflicts with medical-rewrite

PR #4461, opened 2016-09-24
Fix interaction addActionToX return values

### LINKED ISSUE #4460: AddActionToObject and AddActionToClass return number instead of array
**ACE3 Version:** `3.7.0` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- FUNC(addActionToObject) and FUNC(addActionToClass) return a number instead of an array. As the function headers state, the return value should be an array that can be used to remove the actions. Currently, the result of the last pushBack command is returned instead of the full action path. 

**Steps to reproduce:**
- Run editor
- Place playable unit
- Play scenario
- Run code in debug console     
```
testAction = ["testAction","Test Action","",{false},{true}] call ace_interact_menu_fnc_createAction;
testPath = [player, 1, ["ACE_SelfActions"],testAction] call ace_interact_menu_fnc_addActionToObject;
hint str testPath;
```

- Editor (Singleplayer)

**Placed Modules:**
- None

PabstMirror:
Fix #4460
Fix returning number from pushBack
Will still safely make a copy of the array `_parentAction`

### DIFF
--- a/addons/interact_menu/functions/fnc_addActionToClass.sqf
+++ b/addons/interact_menu/functions/fnc_addActionToClass.sqf
@@ -38,6 +38,9 @@ if (param [4, false, [false]]) exitwith {
     ', _index];
     TRACE_2("Added inheritable action",_objectType,_index);
     [_objectType, "init", _initEH, true, [], true] call CBA_fnc_addClassEventHandler;
+
+    // Return the full path
+    (_parentPath + [_action select 0])
 };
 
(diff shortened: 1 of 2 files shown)


PabstMirror:
I don't know what else to return.  Nothing gets added on error

jonpas:
I'd say it's something people can check with `isNil`, although maybe returning `false` would be easier.

PR #4466, opened 2016-09-25
fix damage materials of repair items

commy2:
- if you hit them currently they create dust
- this changes it to a "hit tyre-screech" and sparks respectively.


commy2:
I set up the hit-points LODs with hitpoints, but I don't seem to be able to damage then. Maybe it doesn't work for "thingX". PR still fixes the effects.

PR #4467, opened 2016-09-26
Fix passanger interaction

PabstMirror:
Oddly #4446 broke passenger interaction because `ACE_Dogtag` ends up being before `MainActions`.
Adding requiredAddons will fix mod load order problems.

Should also take a look at making `ace_interaction_fnc_addPassengerAction` more robust.


commy2:
>Oddly #4446 broke passenger interaction because ACE_Dogtag ends up being before MainActions.

strange. I'll protect my nerves and not think about it.

PR #4469, opened 2016-09-27
Fix medical blood locality / bleeding in vehicles

PabstMirror:
- Fix ace_setting
- Don't run on headless if enabledFor = 1
- Don't create blood for non-local units
- Don't create blood when mounted (except static weapons)
- Add Readme

### DIFF
--- a/addons/medical_blood/XEH_postInit.sqf
+++ b/addons/medical_blood/XEH_postInit.sqf
@@ -25,8 +25,17 @@ if (isServer) then {
 ["ace_settingsInitialized", {
     TRACE_1("settingsInitialized", GVAR(enabledFor));
     if (GVAR(enabledFor) == 0) exitWith {}; // 0: disabled
+    if ((GVAR(enabledFor) == 1) && {!hasInterface}) exitWith {}; // 1: enabledFor_OnlyPlayers
 
-    [configFile >> QGVAR(stateMachine)] call CBA_statemachine_fnc_createFromConfig;
+    private _listcode = if (GVAR(enabledFor) == 1) then {
+        {[ACE_player] select {[_x] call FUNC(isBleeding)}} // ace_player is only possible local player
+    } else {
+        {allUnits select {(local _x) && {[_x] call FUNC(isBleeding)}}}; // filter all local bleeding units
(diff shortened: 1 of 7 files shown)


addons/medical_blood/functions/fnc_hit.sqf
```diff
@@ -16,14 +16,13 @@
...
 if (isNull _causedBy) exitWith { // won't be able to calculate the direction properly, so instead we pick something at random
     [QGVAR(spurt), [_unit, random 360, _damage]] call CBA_fnc_serverEvent;
 };
 
 // Calculate bulletDirection
-private _unitPos = getPosASL _unit;
-private _causedByPos = getPosASL _causedBy;
-
-private _bulletDir = ((_unitPos select 0) - (_causedByPos select 0)) atan2 ((_unitPos select 1) - (_causedByPos select 1));
+private _bulletDir = _unit getDir _causedBy;
```
commy2: 👍

addons/medical_blood/XEH_postInit.sqf
```diff
@@ -25,8 +25,17 @@ if (isServer) then {
 ["ace_settingsInitialized", {
     TRACE_1("settingsInitialized", GVAR(enabledFor));
     if (GVAR(enabledFor) == 0) exitWith {}; // 0: disabled
+    if ((GVAR(enabledFor) == 1) && {!hasInterface}) exitWith {}; // 1: enabledFor_OnlyPlayers
+
+    private _listcode = if (GVAR(enabledFor) == 1) then {
+        {[ACE_player] select {[_x] call FUNC(isBleeding)}} // ace_player is only possible local player
+    } else {
+        {allUnits select {(local _x) && {[_x] call FUNC(isBleeding)}}}; // filter all local bleeding units
+    };
```
commy2: does it handle respawning if you do it here?

PabstMirror: Yeah
It's code that gets executed every-time the statemachine does a loop.

commy2: 👍

PR #4474, opened 2016-09-27
respawn - Replace unique acre radios with their baseclass

### LINKED ISSUE #4441: conflict between ace respawn with same gear and unique item mods
conflict between ace respawn with same gear and unique item mods

Here so I don't forget.

Basic summary;
- when a mod usch as acre2 replies on unique items and ace respawn with same gear is enabled, it could potentialy result in a duplicate item (when items are unique due to their classname). We should look at supporting unique items by adding their base class / parent instead of the unique id item.

PabstMirror:
Fix #4441

- Should fix ACRE error
```
ACRE WARNING: Id object relation created independently of unique ID creation process (acre_prc343_id_2)
```
- Also fix `ace_killedByFriendly` global event inside global event

### DIFF
--- a/addons/respawn/functions/fnc_restoreGear.sqf
+++ b/addons/respawn/functions/fnc_restoreGear.sqf
@@ -4,7 +4,7 @@
  *
  * Arguments:
  * 0: Unit <OBJECT>
- * 1: All Gear based on return value of ACE_common_fnc_getAllGear <ARRAY>
+ * 1: All Gear based on return value of getUnitLoadout <ARRAY>
  *
  * Return Value:
  * None
@@ -20,6 +20,22 @@ params ["_unit", "_allGear", "_activeWeaponAndMuzzle"];
 
 // restore all gear
 if (!isNil "_allGear") then {
+    // Handle Unique ACRE radios
+    {
+        private _container = _allGear select _x;
+        if (!(_container isEqualTo [])) then {
+            _container params ["","_itemsArray"];
+            {
+                _x params ["_itemClassname"];
+                if ((getNumber (configFile >> "CfgWeapons" >> _itemClassname >> "acre_isUnique")) > 0) then {
+                    private _newBase = getText (configFile >> "CfgWeapons" >> _itemClassname >> "acre_baseClass");
+                    INFO_2("replacing ACRE radio [%1] with base [%2]",_itemClassname,_newBase);
+                    _x set [0, _newBase];
+                };
+            } forEach _itemsArray;
+        };
+    } forEach [3,4,5]; // Uniform, Vest, Backpack
+
     _unit setUnitLoadout _allGear;
 };
 
--- a/addons/respawn/functions/fnc_restoreGear.sqf
+++ b/addons/respawn/functions/fnc_restoreGear.sqf
@@ -20,16 +20,24 @@ params ["_unit", "_allGear", "_activeWeaponAndMuzzle"];
 
 // restore all gear
 if (!isNil "_allGear") then {
-    // Handle Unique ACRE radios
     {
         private _container = _allGear select _x;
         if (!(_container isEqualTo [])) then {
             _container params ["","_itemsArray"];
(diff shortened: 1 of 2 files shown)


jonpas:
TFAR does it on the same principle, but not sure if it handles those things itself.

commy2:
INFO_2 needs new CBA

654wak654:
For TFAR radio classes go like `{classname}_{number}`.

https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/master/arma3/%40task_force_radio/addons/task_force_radio_items/radio_ids.hpp

PabstMirror:
Need to test with TFAR but it looks like their configs are similar.

commy2:
Wouldn't this mean that you have to readjust your radio every time you respawn?

jonpas:
@commy2 Yes, but it is the only way to handle it, as the old radio is already registered and then you get 2 of them, because cleanup of old IDs doesn't kick in that fast.

commy2:
Some people might feel like this is a regression though. Just saying.

jonpas:
But it's not, both TFAR and ACRE error out on duplicate IDs as far as I know (not 100% on TFAR) and may introduce undefined behaviour.

bux578:
Make it an option with default off. So that it behaves like now. If people want/need that, they could opt-in.

CorruptedHeart:
Just chiming in from the TFAR side, I seem to recall adding functionality that will handle duplicate IDs and copy the settings from the old radio to the new radio or something similar. Unfortunately I'm at work and can't search the repo to find exactly where that is happening on the TFAR side.

commy2:
Something like this ^ is needed.

dedmen:
I think this is the code @CorruptedHeart meant https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_radioReplaceProcess.sqf

also on respawn we are calling
https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_requestRadios.sqf
With a call in line 50 passing its undocumented parameter (true) to https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_radioToRequestCount.sqf
tells radioToRequestCount to replace all Radios. So in turn on respawn all radios should get replaced by new ID's with the same old settings.

As i just got woken up, I'm gonna go back to sleep now.

dedmen:
I'd prefer that you would just call 
```true call TFAR_fnc_requestRadios;```
which will take care of handling duplicate ID's.

If your script executes before TFAR's respawn handler (As it is now. Not copying settings), TFAR will not be able to copy settings from the old radios to the new ones.
If your Respawn Loadout script runs after TFAR's handler though you would have to take care of duplicate ID's.

bux578:
@dedmen The only problem I ever had with duplicate IDs was loading a profile from Virtual Arsenal

CorruptedHeart:
And we have a PR and issue to fix the arsenal: https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/1123

So it might not be an issue after all for the TFAR side if that is the case and no handling will be required on your side.

bux578:
@dedmen feel free to join our chat http://slackin.ace3mod.com/

Dorbedo:
Shouldn't that be fix on TFAR/ACRE side, since it is restricting the usage of arsenal an the `getUnitLoadout` command? In my opinion, you could add a possibility to modify the loadout before spawning. The modification should be done by the mod, which is using unique items.

Glowbal:
I feel this solution may be to specific towards acre. I'd like to generalize it so it can be used by other systems as well. Ig our planned bft components

jonpas:
@Sniperhid will look into it for ACRE2, to try to sort it on that side as it should be.

PabstMirror:
Looks like TFAR and ACRE will both be handling the problem on their own

PR #4479, opened 2016-09-28
Medical rewrite - treatment functions

commy2:
These need commyfication. Don't stop me.


DLEGION:
The medical need for sure a rewrite...first it cannot be turned off right now, and second its not realistic to get hit 5 times each arm and leg  for a total of 20 good shots and still be able to fight. Hope you fix that.

Issue #4326, opened 2016-09-04
ACE 3 Breaks AI's Ability to Aim (No Matter the Difficulty Settings)


Shadow6066:
**Arma 3 Version:** 1.62 (stable)
**CBA Version:** Steam Workshop
**ACE3 Version:** Steam Workshop

**Mods:**
- @CBA_A3
- @ace

**Description:**
- After some testing, it seems Ace causes the AI to lose their ability to aim at ranges past 50m. I have provided a short video highlighting the bug:
  [https://www.youtube.com/watch?v=RzH4iEs_OdQ](https://www.youtube.com/watch?v=RzH4iEs_OdQ)

Please make the AI tweaks a separate module or at least allow us to adjust the AI somehow. I love ACE but it's making my group's missions way too easy.

**Steps to reproduce:**
- Launch Arma 3 with ACE, place unit of each side down in editor (50m or more apart), play as one of them, watch as the other fails to hit you.

**Where did the issue occur?**
- Always

**Placed Modules:**
- None.

commy2:
It already is a seperate mod. Remove ace_ai.pbo and weapons ai should be like vanilla.

Shadow6066:
Thank you very much, I really wish this was made more clear. Maybe an option in game to disable it under the ACE options menu? :)

esteldunedain:
> Maybe an option in game to disable it under the ACE options menu

Sorry, that's not possible due to how this changes work.

esteldunedain:
@commy2, the video provided by @Shadow6066 does show very poor results from the tweaks in AI. Should we scrap them, or at least adjust them?

Shadow6066:
Thanks for watching it, while it does make for some epic sounding firefights once you get involved it gets awkward. I've had too many missions where I said "I shouldn't be alive right now" which was great at first but after so many times the exhilarating moments turned to eye rolls. I feel like you guys had the right idea though and I'm interested to see where this goes.

brndd:
I also tested this and didn't spot the same issue: https://www.youtube.com/watch?v=4ON7iX4TlYw

jonpas:
@Shadow6066 can you check the Skill slider and other AI options that might influence it? @brndd really does show the opposite of what you are experiencing.

Shadow6066:
Skill slider is at default 50% where it should be, same with vanilla but notice how bad the difference is.

commy2:
[...]

PR #4484, opened 2016-09-29
Removed aiDispersionCoef for vanilla weapons

theYoursTruly:
-  restore default aiDispersionCoef for vanilla weapons.

### DIFF
--- a/addons/ai/CfgWeapons.hpp
+++ b/addons/ai/CfgWeapons.hpp
@@ -2,8 +2,6 @@
 // weapon config changes, by commy2
 
 /* documentation:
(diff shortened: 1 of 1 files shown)


PabstMirror:
Ref #4326
IIRC some of these changes are old (from AGM)

commy2:
TMR actually.

https://github.com/Taosenai/tmr/blob/1bbab0f379248f0c341d7579310ef630b18168ef/tmr_ai/conf_mg_full.h#L1-L2

esteldunedain:
:+1: for this. It will make everything more uniform, as the tweaks didn't apply to Apex weapons, mods, etc.

It would be nice to have a table with before/after coefficient though

PR #4085, opened 2016-07-11
Update showHud setter for Squad Radar

jonpas:
**When merged this pull request will:**
- Replaces #4043 for `master` branch (didn't make it into 1.62)

PR #4485, opened 2016-09-29
Revert "Update showHud setter for Squad Radar"

jonpas:
Reverts acemod/ACE3#4085

Apparently it's still not in 1.64... I bet @PabstMirror told me that and I forgot.


PabstMirror:
It's weird because the `showHud` array did change, but I guess the feature didn't make it in
Ref: https://forums.bistudio.com/topic/191601-spatial-awareness-improvements/page-4#entry3095280

Like I said, we can leave this in, just make the setting non client settable to avoid confusion

jonpas:
I'd prefer if it isn't in at all until it does something, we'll have to do something about it either way.

Issue #4477, opened 2016-09-28
Map gesture group settings module errors


Fourjays:
**Arma 3 Version:** `1.64`
**CBA Version:** `3.0.0.160713`
**ACE3 Version:** `3.7.0`

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- The Map Gestures Group Settings module produces script error messages and fails to set the color, when placed in a mission and synchronised to units.
- This error message (http://steamcommunity.com/sharedfiles/filedetails/?id=771422481) appears when initially loading in to the editor preview, but not on a multiplayer dedicated server.
- This error message (http://steamcommunity.com/sharedfiles/filedetails/?id=771422511) appears on both single player preview and the dedicated server when attempting to make a map gesture.
- When no Map Gestures Group Settings module is present, the map gestures work normally with the default settings, or those set with the Map Gestures Settings module.
- This module previously didn't work in 3.6.2, but simply logged an "invalid hash" error to the RPT.

**Steps to reproduce:**
- Synchronize a Map Gesture Group Settings module to a playable unit.
- Change the color (maybe unneccessary).
- Start the mission.

**Where did the issue occur?**
- Dedicated / Editor (Singleplayer)

**Placed Modules:**
- Map Gestures - Group Settings

**RPT log file:**
- Client: http://pastebin.com/WSm4KjqK
- Server: http://pastebin.com/YVhJXJJs
- Client ACE Debug (single player editor): http://pastebin.com/HjbzA20V

Issue #4423, opened 2016-09-17
Fix ace settings array serializing


bux578:
reminder
![a](https://i.gyazo.com/7c511ce8a70fc3428353053b17506273.png)

bux578:
@commy2 @PabstMirror

PR #4486, opened 2016-09-29
Cleanup map gesture sync / color mapping

PabstMirror:
Tag @MikeMatrix 
- Adds ace_common_fnc_setPlayerOwner which syncs player's owner var
- Redo publicVarClient with CBA's ownerEvent
- Remove old ace_settings
- Make color module run globaly
- Add public ace_map_gestures_fnc_addGroupColorMapping

Will "fix" #4477 #4423 
Mainly by dropping 2 `ARRAY` ace_settings for group color mapping.

I don't think there is any way to convert them to cba_settings, so they will probably be dropped soon anyways.
Bigger problem is we broke most saved settings last patch by switching hash implementations from ace to cba.
Old modules will still work (they now run globally and do not use the setting system).
Also added a public setting function.


addons/map_gestures/functions/fnc_drawMapGestures.sqf
```diff
@@ -15,6 +15,8 @@
  */
 #include "script_component.hpp"
 
+BEGIN_COUNTER(draw);
```
Glowbal: nice

addons/map_gestures/functions/fnc_addGroupColorMapping.sqf
```diff
@@ -0,0 +1,31 @@
...
+ * Example:
+ * ["Alpha 1-1", [1,0,0,1], [0,1,0,1]] call ace_map_gestures_fnc_addGroupColorMapping
+ *
+ * Public: Yes
+ */
+#include "script_component.hpp"
+
+if (!params [["_group", "", [grpNull, ""]], ["_leadColor", [1,1,1,1], [[]], 4], ["_unitColor", [1,1,1,1], [[]], 4]]) exitWith {
+    ERROR_1("Bad Params: %1",_this);
+    -1
```
Glowbal: returning `-1` here but on the other `exitWith` statements we do not return anything.

PR #4487, opened 2016-09-29
Fix cookoff event calling

PabstMirror:
Same issue as #4242


commy2:
Not as simple as this means that the function no longer has global effects when it's called from a remote machine.
Should be fine though as it's only called by handleDamage.

PR #4489, opened 2016-09-30
Add model.cfg for medical blood

PabstMirror:
Adds model.cfg for medical_blood p3ds.

On a test mission, executing 
```
["blooddrop_3", getPos player] call ace_medical_blood_fnc_createBlood;
```
Will cause hard popup error and log to rpt
```
Warning Message: No entry 'bin\config.bin/CfgModels.default'.
Warning Message: No entry '.sections'.
Warning Message: Size: '/' not an array
Warning Message: No entry '.sectionsInherit'.
Warning Message: '/' is not a value
Warning Message: No entry '.sections'.
Warning Message: Size: '/' not an array
Warning Message: No entry '.sectionsInherit'.
Warning Message: '/' is not a value
```

I am not sure why we need it, but adding model.cfg with basic entries fixed.
Perhaps some oddity with `createSimpleObject`?


commy2:
https://forums.bistudio.com/topic/140837-development-branch-changelog/?p=3098075

>Fixed: A 'Selection missing in CfgModels' message would appear in the RPT file when using Simple Objects

PabstMirror:
1.66 could be 2 months

PR #4491, opened 2016-09-30
Update showHud setter for Squad Radar 2

jonpas:
- Add #4085 at proper time (when Squad Radar is actually working in vanilla, is not part of 1.64).


PabstMirror:
https://forums.bistudio.com/topic/191601-spatial-awareness-improvements/page-4#entry3117826
Makes it sound like this won't be in for a while, and might be completely different.

jonpas:
Nice find @PabstMirror , will keep PR open anyways until we see what it's about.

jonpas:
This made it in yet or no?

PR #4492, opened 2016-09-30
add a factor to scale fatigue up and downhill

commy2:
- pls no bully, baer
- renamed `_velocity` to `_speed` so it's not confused with the `velocity` command I use for this and one is not inclined to replace it.
- loosely based on this graph:

### DIFF
--- a/addons/advanced_fatigue/functions/fnc_getMetabolicCosts.sqf
+++ b/addons/advanced_fatigue/functions/fnc_getMetabolicCosts.sqf
@@ -16,7 +16,7 @@
  * Public: No
  */
 #include "script_component.hpp"
-params ["_unit", "_velocity"];
+params ["_unit", "_speed"];
 
 private _virtualLoad = 0;
 {
@@ -32,6 +32,14 @@ private _gearMass = ((loadAbs _unit + _virtualLoad) * 0.1 / 2.2046) * GVAR(loadF
 private _terrainFactor = 1;
 private _terrainAngle = asin (1 - ((surfaceNormal getPosASL _unit) select 2));
 private _terrainGradient = (_terrainAngle / 45 min 1) * 5 * GVAR(terrainGradientFactor);
+private _terrainAltitude = getPosASL _unit;
+private _terrainAltitudeDelta = getTerrainHeightASL _terrainAltitude - getTerrainHeightASL (_terrainAltitude vectorDiff vectorNormalized velocity _unit);
+private _terrainAltitudeFactor = 2 * abs _terrainAltitudeDelta min 1;
+
+if (_terrainAltitudeDelta < -0.1) then {
+    _terrainAltitudeFactor = _terrainAltitudeFactor / 4;
+};
+
 private _duty = GVAR(animDuty);
 
 {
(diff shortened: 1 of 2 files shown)


commy2:
`surfaceNormal` doesn't change whether you go up or down though.
good point about `getPosATL` and pathway lods

Currently this goes to zero if you walk along the slope, which we don't want.
I'll be waiting for baers feedback before doing anything though.

esteldunedain:
> doesn't change whether you go up or down though.

```sqf
_slope = (surfaceNormal _pos) vectorProduct (vectorNormalized velocity ACE_player);
```
Vector algebra 😄

esteldunedain:
> Currently this goes to zero if you walk along the slope, which we don't want.

I think that's not bad. But if you really want to take into account that walking along a slope is a bit more tiresome than walking on a flat surface you could add a coefficient based on `1 - (surfaceNormal _pos select 2)` that's independent of walking direction.

BaerMitUmlaut:
Don't quite see the need for renaming `_velocity`, there's also a `speed` command after all. You've also kept in the old slope penalty as it is right now, that should be changed.

commy2:
>You've also kept in the old slope penalty as it is right now, that should be changed.

What part of this is the "old slope penalty"? `_terrainGradient` ? Of course that should stay, because the surface normal isn't part of my factor.

commy2:
>Don't quite see the need for renaming _velocity, there's also a speed command after all.

I just think it's weird to use `velocity` when you already have a variable named `velocity`. Usually it's like this: Absolute value -> speed, Vector -> velocity (that's what the commands return, except that speed is using the 3.6 factor for km/h)
This is totally unofficial commy SQF variable naming though.

BaerMitUmlaut:
`_terrainGradient` was the main factor that penalized slopes which provided decent results. Your factor adds another penalty, but without changing the `_terrainGradient` factor - that punishes the player too much, because you're essentially punishing him twice.

commy2:
`_terrainAltitudeFactor` is always smaller or equal than 1. It's supposed to make walking downwards easier.   You said walking uphills should stay the same.

BaerMitUmlaut:
That doesn't work though. For example, let's say I walk parallel to a slope.
```
private _terrainAltitudeFactor = 2 * abs _terrainAltitudeDelta min 1;
```
In that case `_terrainAltitudeDelta` would be 0 because I don't actually climb, so `_terrainAltitudeFactor` would then also be 0. Later you do
```
_terrainGradient * _terrainAltitudeFactor
```
which would then also become 0, so no penalty for walking on slopes, which is the opposite of what we want.

commy2:
Wanna trade with the weather issue report?

BaerMitUmlaut:
Sure :smile:

addons/advanced_fatigue/functions/fnc_getMetabolicCosts.sqf
```diff
@@ -32,6 +32,14 @@ private _gearMass = ((loadAbs _unit + _virtualLoad) * 0.1 / 2.2046) * GVAR(loadF
 private _terrainFactor = 1;
```
ulteq: `_terrainFactor` -> dead code?

PR #4494, opened 2016-10-01
Use hitpoints in medical Extension instead of selections

Glowbal:


commy2:
I think they have to be `"Head", "Body", "LeftArm", "RightArm", "LeftLeg", "RightLeg"`, because that is what the woundsReceived event handler reports.
I cut away the "Hit", because it seemed superfuous to me.

Issue #2901, opened 2015-12-01
Flare with parachute


Xawery94:
I have a proposition. Can you make a flare with parachute because the vanilla flare is maintained in the air very short and falls to the ground very quickly. But I have tested a Hunt IR and it has own parachute and stay in air for very long time. So it would be nice to see this flare with a similar parachute

headedhorseman101:
@commy2 realistic values for the rounds IRL if you want the are here: [TM 43-0001-28] (http://everyspec.com/ARMY/TM-Tech-Manual/download.php?spec=TM_43-0001-28_C11_27OCT2003.013353.pdf). Page 6-37 or 606.

commy2:
thanks

commy2:
>The cavity contains a pyrotechnic
>flare candle assembly, and an integral
>ignition/ejection charge attached to a 20-
>inch diameter parachute. The projectile has a
>4- to 5-second delay ignition element crimped
>into the center opening of a metal delay carrier.
>The
>attached parachute deploys upon ejection to
>lower the candle at 7 feet per second. The candle
>burns for approximately 40 seconds.

sargken:
How about a Parachute with a flare attached for units to stay in formation at night?

Mike-MF:
Closing due to age, inactivity or already solved. May be re-opened if the issue is still present, even better just open a new up-to-date Issue.

PR #4506, opened 2016-10-03
Illumination Flares

PabstMirror:
Ref #2901
- GL and Mortar flares provide illumination
- Slows decent on GL flare (simulating a parachute)

Since 1.60/62 lighting changes arma can get very dark and flares offer almost no lighting until they are almost at ground level. Really only useful for signaling.

This make the flare provide a mild amount of  illumination.
It also slows their decent by simulating a parachute, but no change in model (based on M583's stats).

Two questions: 
Realistically- How much illumination would these types of flares produce?
Best way to implement the changes?
- Create a brand new magazine type with the changes. Would be a real pain with `magazines[]` configs on weapons.
- Add this to an existing ace addon (like grenades which already touches the intensity)
- Add it to a new optional addon, if people do not want the illumination changes.

### DIFF
--- a/optionals/flares/$PBOPREFIX$
+++ b/optionals/flares/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\ace\addons\flares
--- a/optionals/flares/CfgAmmo.hpp
+++ b/optionals/flares/CfgAmmo.hpp
@@ -1,0 +1,15 @@
+class CfgAmmo {
+    class FlareCore;
+    class FlareBase: FlareCore {};
+
+    class F_40mm_White: FlareBase {
+        intensity = 1000000; // vanilla: 10000
+        timeToLive = 40; // vanilla: 25
+        coefGravity = 0.25; // vanilla: undefined (would be 1)
+        // Makes the ammo fall the ground slower
+    };
+
+    class Flare_82mm_AMOS_White: FlareCore {
+        intensity = 1500000; // vanilla: 10000
+    };
+};
--- a/optionals/flares/README.md
+++ b/optionals/flares/README.md
@@ -1,0 +1,11 @@
+ace_flares
+============
+
+Increasing light from flares making them suitable for illumination instead of just signaling.
+Based on the M583 (burning for 40 seconds, parachute slowing decent to 2 Meters/Sec).
+
+## Maintainers
+
+The people responsible for merging changes to this component or answering potential questions.
+
+- [PabstMirror](https://github.com/PabstMirror)
--- a/optionals/flares/config.cpp
+++ b/optionals/flares/config.cpp
@@ -1,0 +1,17 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        name = COMPONENT_NAME;
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"ace_common", "ace_grenades"};
+        author = ECSTRING(common,ACETeam);
+        authors[] = {"PabstMirror"};
+        url = ECSTRING(main,URL);
+        VERSION_CONFIG;
+    };
+};
+
+#include "CfgAmmo.hpp"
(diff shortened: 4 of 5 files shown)


jonpas:
> Create a brand new magazine type with the changes. Would be a real pain with magazines[] configs on weapons.

We already do it for HuntIR, I'd say go that way, if not then optional sounds good.

esteldunedain:
> you could also say it improves realism and so adding it to existing addon would also work

:+1: for that. As far as understand illumination M203 flares should provide noticeable illumination, at least on pitch black conditions. 

From http://www.globalsecurity.org/military/library/policy/army/fm/3-22-31/chap3.htm#3-2
> - (M583A1) Star Parachute Round. This round is white impact or bar alloy aluminum, with black markings. It is used for illumination and signals and is lighter and more accurate than comparable handheld signal rounds. The parachute attached to the round deploys upon ejection to lower the candle at 7 feet per second. The candle burns for about 40 seconds. A raised letter on the top of the round denotes the color of the parachute.[8]

> - (M585) White Star Cluster Round. This round is white impact or bar aluminum alloy, with black markings. The attached plastic ogive has five raised dots for night identification. The round is used for illumination or signals. It is lighter and more accurate than comparable handheld signal rounds. The individual stars burn for about 7 seconds during free fall.[9]

sargken:
So are you making star flares also?

PabstMirror:
Because of limitations of `ARRAY[] +=` I am not sure how to add the magazine without causing problems.
Both this and HuntIR will conflict.

By just modifying the existing `UGL_FlareWhite_F` we also gain compatibility with existing mods that allow using that magazine. (e.g. HLC and RHS both use it were appropriate)

SilentSpike:
I'd say edit the existing magazines, as far as I'm concerned this seems to be a realism improvement (flares in A3 have sucked ever since they changed them from how they were in the alpha).

PabstMirror:
With #4521 we will be able to add new magazines,
so we could scrap this and redo it

jonpas:
What needs redoing? New flares instead of modifying the existing ones?

severgun:
What about aircraft flares? mk24 mk45
http://www.bits.de/NRANEU/others/amd-us-archive/FM20-60%2870%29.pdf Chapter 6 page 25
www.dtic.mil/cgi-bin/GetTRDoc?AD=AD0721697

PabstMirror:
Redone using `CfgMagazineWells`

addons/grenades/CfgAmmo.hpp
```diff
@@ -17,6 +17,18 @@ class CfgAmmo {
         intensity = 40000;
         flareSize = 12;
     };
+    class ACE_40mm_Flare_white: F_40mm_White {
+        intensity = 1250000; // vanilla: 10000
+        timeToLive = 45; // vanilla: 25, ace changes to 60
```
dedmen: >ace changes to 60

But it says 45?

dedmen:
Have you thought about IR illumination flares?
http://tikka.ws/class/index.php?b=cfgAmmo&c=UK3CB_BAF_F_40mm_CIR&conf=1
Apparently that works?

addons/grenades/CfgAmmo.hpp
PabstMirror: ace changes the upper base class to 60 from 25, I'll improve wording
```
    class FlareBase: FlareCore {
        timeToLive = 60;
```

commy2:
We have these for BWA3 for the Gustav. Should be as easy to script.

dedmen:
No need to script. Flares have a `nvgOnly` entry. Just set that to `1` and done.

commy2:
As far as I can tell, this PR could be merged, or we could add IR flares as well and merge those as well. Right? PR lgtm.

commy2:
@dedmen I set `nvgOnly` to 1, but it is still visible with the eyes.

dedmen:
Maybe they broke it? It was atleast working some time ago.

commy2:
Try this branch. Use 
```sqf
execVM "z\ace\addons\grenades\dev\dev_testFlares.sqf";
```
for a quick test.

madpat3:
delete the vanilla ones. they don't work, therefore useless and only misleading when having the ace flares next to it in the arsenal list.

dedmen:
Can just fix the vanilla flares instead.

PR #4507, opened 2016-10-04
Cleanup old common macros

PabstMirror:
`ERROR(` use of `throw` causes script error
None of the others are used.


commy2:
were*

PabstMirror:
Yeah, I dislike code in functions/script_component.hpp, I never think to look there.

commy2:
The file should only be used to include the main folder `script_component.hpp`

PR #4509, opened 2016-10-04
Disable FCS on RHS helicopters

### LINKED ISSUE #4495: RHS: USAF AH-64D Zeroing not working correctly with ACE
**Arma 3 Version:** `1.64`
**CBA Version:** `3.1.0`
**ACE3 Version:** `3.7.0`

**Mods:**
- `@CBA_A3`
- `@ace`
- `@RHS: United States Forces` v0.4.1.1

**Description:**
- I have found out that the ACE targeting system interferes with the regular Arma 3/RHS zeroing system in such a way, that it will be impossible to get any accurate shots.
One of the way ACE intervenes, is through the UI for the zeroing. (Normally found on the top right of the UI) Whenever one uses 'tab' button to change zeroing or the manual page-up and down, the UI doesn't change to the correct number (i.e. it stays at the default position of 400m for the Apache) This, while the actual zeroing does change. 
Moreover, this messes with the AUTO-zero feature the Apache has when locking onto something. (Like your own laser) The AUTO-zero doesn't compensate for the ACE zeroing, thus making your gun 'inaccurate' and shoot short or over.

As of now, I do not know if the same problem occurs with other, similar, vehicles.

**Steps to reproduce:**
-Enable RHS USAF, ACE3 and CBA3 for Arma 3
-Use any RHS Apache helicopter and get into the gunner seat.
-Use the 'tab' button or the page up and down to use the ACE zeroing mechanism.
-See how the UI for the zeroing doesn't change.
-Test it also with AUTO-zeroing.

- Editor (Singleplayer) / Multiplayer

**Placed Modules:**
RHS: USAF AH-64D, any variant

**RPT log file:**
- If possible at the time the bug is encountered, go to ACE Options and select "Debug To Clipboard", this will print extensive debug information to the RPT file.

PabstMirror:
Close #4495

### DIFF
--- a/optionals/compat_rhs_usf3/CfgVehicles.hpp
+++ b/optionals/compat_rhs_usf3/CfgVehicles.hpp
@@ -111,10 +111,7 @@ class CfgVehicles {
             };
         };
     };
-    class AllVehicles;
-    class Air: AllVehicles {
-        class Turrets;
-    };
+    class Air;
     class Helicopter: Air {
         class Turrets {
             class MainTurret;
--- a/optionals/compat_rhs_usf3/CfgVehicles.hpp
+++ b/optionals/compat_rhs_usf3/CfgVehicles.hpp
@@ -111,6 +111,15 @@ class CfgVehicles {
             };
         };
     };
+    class AllVehicles;
+    class Air: AllVehicles {
+        class Turrets;
+    };
+    class Helicopter: Air {
+        class Turrets {
+            class MainTurret;
+        };
+    };
 
     class MBT_01_base_F: Tank_F {};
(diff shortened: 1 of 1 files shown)


optionals/compat_rhs_usf3/CfgVehicles.hpp
```diff
@@ -111,6 +111,15 @@ class CfgVehicles {
             };
         };
     };
+    class AllVehicles;
+    class Air: AllVehicles {
+        class Turrets;
+    };
```
commy2: replace L114 - 117 with just:
`class Air;`

lolkij:
ACE FCS still present in the MI-24 (GSh-30)

PR #4514, opened 2016-10-05
Functions to perform arithmetic ops on sets

PabstMirror:
Allows add/removing blocks of code that return numbers.
And performing arithmetic operations on those sets.

Allows a variable number of sources to provide input into a final calcuation. Like Aim Coefficent or possibly `fadeSound`)

Inspired by Baer's AF duty system.

Example:
```
[missionNamespace, "ace_AimCoef", "ace_medical", {
    linearConversion [0,1,(ace_player getVariable "ace_medical_pain"),1,0.2,true]
}] call ace_common_fnc_arithmeticSetSource;

[missionNamespace, "ace_AimCoef", "ace_advancedFat", {
    (1.0 + _fatigue ^ 2 * 0.1);
}] call ace_common_fnc_arithmeticSetSource;

[missionNamespace, "ace_AimCoef", "min"] call ace_common_fnc_arithmeticGetResult;
```

First arg can be any type of namespace (object, location, `missionNamespace`)
Can use the namespace in the code block:
```
[player, "test", "a1", {_namespace getVariable "a1"}] call ace_common_fnc_arithmeticSetSource;
```

### DIFF
--- a/addons/common/functions/fnc_arithmeticGetResult.sqf
+++ b/addons/common/functions/fnc_arithmeticGetResult.sqf
@@ -1,0 +1,69 @@
+/*
+ * Author: PabstMirror
+ * Gets arithmetic resultfrom a set.
+ *
+ * Arguments:
+ * 0: Namespace <OBJECT><LOCATION><MISSIONNAMESPACE>
+ * 1: Number Set ID <STRING>
+ * 2: Operation (sum, product, min, max, avg) <STRING>
+ *
+ * Return Value:
+ * <NUMBER>
+ *
+ * Example:
+ * [ace_player, "ace_aimCoefficents", "product"] call ace_common_fnc_arithmeticGetResult
+ * [missionNameSpace, "ace_hearing", "min"] call ace_common_fnc_arithmeticGetResult
+ *
+ * Public: Yes
+ */
+// #define DEBUG_MODE_FULL
+#include "script_component.hpp"
+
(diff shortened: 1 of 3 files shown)


addons/common/functions/fnc_arithmeticGetResult.sqf
```diff
@@ -0,0 +1,69 @@
...
+    };
+    case ("avg"): {
+        private _result = 0;
+        {
+            _result = _result + (call _x);
+            nil
+        } count _data;
+        _result / (count _data); // return
+    };
+    default {3735928559};
```
commy2: what? ^^

PabstMirror: Magic number to indicate error

commy2:
How does it pass `_fatigue` ?

PabstMirror:
This assumes it's called inside of `ace_advanced_fatigue_fnc_handleEffects`.
`_fatigue` is defined in that scope, so it will be available to the code if called from there.

bux578:
> ace_advancedFat

gnihihi

addons/common/functions/fnc_arithmeticGetResult.sqf
```diff
@@ -0,0 +1,69 @@
+/*
+ * Author: PabstMirror
+ * Gets arithmetic resultfrom a set.
```
bux578: `result from`

PR #4516, opened 2016-10-06
fix fatigue when diving while sprinting

### LINKED ISSUE #4481: Advanced Fatigue - Diving to the ground from a full sprint uses 20% - 60% of all stamina
**CBA Version:** `3.0.0` stable
**ACE3 Version:** `3.7.0` stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
Your stamina wastes away in huge gulps as you dive prone to ground. The range of stamina loss is quite large. Most of the time you only lose 20%, but sometimes you can lose huge amounts like 60% at a time. The huge losses appear rarely and seem to perhaps be linked to doing a second dive soon after the first one.

**Steps to reproduce:**
Regardles of your gear weight or if you have a rifle on hand or not, do the following:
- Start full sprint and run for a few seconds
- Press z to dive to the ground
- Stop prone on the ground and watch your stamina plummet.

- Dedicated/ Editor (Singleplayer)

**Placed Modules:**
- None

commy2:
- Caps assumed max speed while prone.
- This ensures that moves like diving to the ground while sprinting are not calculated as crawling with running speed.
- fixes #4481

the max speed I could do while crawling is 1.33_ with rifle. Don't think there is anything faster than 1.5 aside from the transitions we try to fix here ofc.

### DIFF
--- a/addons/advanced_fatigue/functions/fnc_mainLoop.sqf
+++ b/addons/advanced_fatigue/functions/fnc_mainLoop.sqf
@@ -15,6 +15,12 @@ if (!alive ACE_player) exitWith { // Dead people don't breath, Will also handle 
 
 private _currentWork = REE;
 private _currentSpeed = (vectorMagnitude (velocity ACE_player)) min 6;
+
+// fix #4481. Diving to the ground is recorded as PRONE stance with running speed velocity. Cap maximum speed to fix.
+if (stance ACE_player == "PRONE") then {
+    _currentSpeed = _currentSpeed min 1.5;
+};
+
 if ((vehicle ACE_player == ACE_player) && {_currentSpeed > 0.1} && {isTouchingGround ACE_player || {underwater ACE_player}}) then {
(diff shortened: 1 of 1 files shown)


addons/advanced_fatigue/functions/fnc_mainLoop.sqf
```diff
@@ -15,6 +15,12 @@ if (!alive ACE_player) exitWith { // Dead people don't breath, Will also handle
 
 private _currentWork = REE;
 private _currentSpeed = (vectorMagnitude (velocity ACE_player)) min 6;
+
+// fix #4481. Diving to the ground is recorded as PRONE stance with running speed velocity. Cap maximum speed to fix.
```
bux578: 👍  for the comment

austinjames314:
Tested, and it's working most of the time. If sprinting while my rifle is slung though, it's still happening. I'm guessing that the animation while  the rifle is slung is a slightly different one?

PR #4521, opened 2016-10-08
Simplify huntIR configs using +=

### LINKED ISSUE #4199: RHS:USAF M320 GLM doesn't use the HuntIR
The M320 GLM is basically the same as the under barrel M320 so it should be able to shoot HuntIR rounds and possibly the 3Rnd 3GL rounds.

PabstMirror:
Close #4199

Based on work by @robalo https://github.com/toadie2k/NIArms/commit/365a10dcd8a51d3de358e8be2f82aacc3b84c32d

`+=` fails on inheritance in a single config; 
but it can work if done in separate configs
and we can include multiple config.cpp in a single pbo

Allows us to simplify our configs, and should help with compatibility with other mods using similar techniques.
Also will allows us to add multiple gl magazines without creating conflicts.  (ref #4506)

### DIFF
--- a/addons/huntir/CfgWeapons.hpp
+++ b/addons/huntir/CfgWeapons.hpp
@@ -19,11 +19,5 @@ class CfgWeapons {
     class UGL_F: GrenadeLauncher {
         magazines[] += {"ACE_HuntIR_M203"};
     };
-    class Rifle_Base_F;
-    class arifle_MX_Base_F: Rifle_Base_F {
-        class GL_3GL_F: UGL_F {
-            //magazines[] += {"ACE_HuntIR_M203"};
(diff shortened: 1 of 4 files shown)


esteldunedain:
Perfect! 

I think it might be cleaner to include two subconfig folders named something like: `HuntirCfg1` and `HuntirCfg2`.

jonpas:
> I think it might be cleaner to include two subconfig folders named something like: `HuntirCfg1` and `HuntirCfg2`. 

I disagree, one is base config of the thing itself, `subConfig` is simply for compatibility and shouldn't be confused with the main config.

esteldunedain:
Oh, sorry. I glanced over this and somehow asumed this was inside the RHS compat pbo. No need for two subconfigs then.

I still think that, `subConfig` as a name is non-descriptive, but meh.

PabstMirror:
There are 2 subconfigs, one for huntIR and one for the RHS compat.

I wasn't sure how to organize them.
We could do something like 

- CfgWeapons.hpp
- CfgWeapons2.hpp
- subconfig/config.cpp which includes weapons2

esteldunedain:
Well, seems like there's a subconfig in the compat pbo too.

What if next month we decide to add the cluster flares to the GL? Unless I'm mistaken, we'd suddenly need two more subconfig folders in the RHS compat pbos. That's why I think it'd be a good idea to to include  "HuntIR" in the folder name.

PabstMirror:
Subconfig is for the inherited gl weapon
```
        class M320_GL: M203_GL {
             magazines[] += {"ACE_HuntIR_M203"};
        };
```
Flares would just go in the same list. So the subconfig isn't really specific to huntir

PR #4533, opened 2016-10-09
Fix missing refuel menu on static fuel stations

### LINKED ISSUE #4077: Arma 3 Lakeside Refuel Problem (Better)

**Mods:**
- `@CBA_A3`
- `@ace`
- `@Lakeside
- `@task_force_RADIO`
- `@a3l`

**Description:**
-I have no Interaction Menu on Gas Sation also i cant Take Nozzle no ERROR message

**Steps to reproduce:**
-I have no Interaction Menu on Gas Sation also i cant Take Nozzle

- Dedicated 

**Placed Modules:**
- RefuelSettings

**RPT log file:**
-http://www.file-upload.net/download-11754563/RPTLOG2.txt.html
(need File Upload then Text to long for Pastebin xd)

Dystopian:
- Fix missing refuel menu on static fuel stations.

I think it would fix #4077.

I found that:
- to fix missing refuel menu it's enough to create vehicle of the same class (it can be deleted at once);
- it's much faster to check conditions in if-then condition inside `'true' configClasses` loop than to check conditions in configClasses parameter.

Save/load, respawn, jip compatible.

### DIFF
--- a/addons/refuel/CfgVehicles.hpp
+++ b/addons/refuel/CfgVehicles.hpp
@@ -144,13 +144,11 @@ class CfgVehicles {
     class House_F: House {};
 
     class House_Small_F: House_F {
-        class EventHandlers;
-
         class ACE_Actions {
             class ACE_MainActions {
                 displayName = ECSTRING(interaction,MainAction);
                 selection = "";
-                distance = 10;
+                distance = 5;
                 condition = "true";
             };
         };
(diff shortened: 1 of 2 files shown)


commy2:
Don't you have to rotate and tilt the object too?

commy2:
There probably is a cleaner solution.

PabstMirror:
The problem might just be that the menu's aren't being compiled on buildings without xeh.
Might be able to just run ace_interact_menu_fnc_compileMenu direcly.

654wak654:
Does this have anything to do with `ace_interact_menu_addBuildingActions`?

PabstMirror:
Updated PR to directly compile with
```
[configName _x] call EFUNC(interact_menu,compileMenu);
```

commy2:
lgtm

PR #4534, opened 2016-10-10
emulate frag grenade ai behaviour for incendiaries

commy2:
- this should emulate the hard coded behaviour pretty accurately

### DIFF
--- a/addons/grenades/functions/fnc_throwGrenade.sqf
+++ b/addons/grenades/functions/fnc_throwGrenade.sqf
@@ -56,7 +56,7 @@ if (getNumber (_config >> QGVAR(incendiary)) == 1) then {
     private _fuzeTime = getNumber (_config >> "explosionTime");
     private _timeToLive = getNumber (_config >> "timeToLive");
 
-    [FUNC(incendiary), [_projectile, _timeToLive], _fuzeTime] call CBA_fnc_waitAndExecute;
+    [FUNC(incendiary), [_projectile, _timeToLive, side _unit], _fuzeTime] call CBA_fnc_waitAndExecute; // WE WANT THE OBJECTS SIDE HERE!
 };
 
 // handle throw modes
--- a/addons/grenades/functions/fnc_incendiary.sqf
+++ b/addons/grenades/functions/fnc_incendiary.sqf
@@ -16,6 +16,8 @@
  */
 #include "script_component.hpp"
 
+#define ALERT_NEAR_ENEMY_RANGE 60
+
 #define PARTICLE_LIFE_TIME 2
 #define PARTICLE_DENSITY 20
 #define PARTICLE_SIZE 1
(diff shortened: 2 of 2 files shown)


addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -56,7 +56,7 @@ if (getNumber (_config >> QGVAR(incendiary)) == 1) then {
     private _fuzeTime = getNumber (_config >> "explosionTime");
     private _timeToLive = getNumber (_config >> "timeToLive");
 
-    [FUNC(incendiary), [_projectile, _timeToLive], _fuzeTime] call CBA_fnc_waitAndExecute;
+    [FUNC(incendiary), [_projectile, _timeToLive, side _unit], _fuzeTime] call CBA_fnc_waitAndExecute; // WE WANT THE OBJECTS SIDE HERE!
```
Glowbal: > WE WANT THE OBJECTS SIDE HERE

But we use the side of the unit?

PabstMirror:
Could we just add something like this?
```
        visibleFire = 30;
        visibleFireTime = 5;
        audibleFire = 16;
```

addons/grenades/functions/fnc_throwGrenade.sqf
commy2: `side OBJECT` (the _unit is an OBJECT) as opposed to `side GROUP` (which would be `side group _unit`)

See: https://community.bistudio.com/wiki/side
>Once dead, a unit will be on the civilian side. Query the side of the Group to get a reliable result.
(It leaves out how objects with `setCaptive` report SIDE_FRIENDLY and renegades report SIDE_ENEMY.)

This is not the wanted behaviour most of the time. We usually use `side group _object`. This is an exception though. We mimick the hard coded AI behaviour and that seems to actually use stuff like `SIDE_ENEMY` and `SIDE_FRIENDLY`. If you make yourself a captive, even enemies won't react to your grenades.

commy2:
@PabstMirror SmokeGrenade and

`class ACE_G_M14: SmokeShell` (via. `class SmokeShell: GrenadeHand`) already inherits these from `class GrenadeHand`. It doesn't seem to work with `simulation = "shotSmokeX";`

commy2:
And I need `simulation = "shotSmokeX";` for the burning sound:
```
        SmokeShellSoundHit1[] = {QPATHTOF(sounds\incen_exp1.ogg),3.1622777,1,500};
        SmokeShellSoundHit2[] = {QPATHTOF(sounds\incen_exp2.ogg),3.1622777,1,500};
        grenadeFireSound[] = {"SmokeShellSoundHit1",0.5,"SmokeShellSoundHit2",0.5};
```
and half of the particle effects:
```
        effectsSmoke = "ACE_Incendiary";
```

commy2:
Also even `GrenadeHand` doesn't use these values. They are for __firing__ (here: throwing) a weapon.
The enemies are alarmed when the grenade explodes. There are no config values for that. It's hard coded as far as I can tell.

PR #4536, opened 2016-10-10
add addPain function, cap pain

commy2:
- this adds the planned addPain function.
- it also caps the pain at 1 to avoid scenarios where > 3 morphine injectors are needed
- new received pain is scaled depending on the amount of already existing pain

this could be used to knock out or wake up a unit depending on if a certain pain threshold is reached via the state machine.

### DIFF
--- a/addons/medical/functions/fnc_addPain.sqf
+++ b/addons/medical/functions/fnc_addPain.sqf
@@ -1,0 +1,28 @@
+/*
+ * Author: commy2
+ * Adds pain to unit and plays a scream sound.
+ *
+ * Arguments:
+ * 0: Argument Name <TYPE>
+ * 0: Argument Name <TYPE>
+ *
+ * Return Value:
+ * Return Name <TYPE>
+ *
+ * Example:
+ * ["example"] call ace_[module]_fnc_[functionName]
+ *
+ * Public: [Yes/No]
+ */
+#include "script_component.hpp"
+
+params ["_unit", "_painToAdd"];
+
+_painToAdd = _painToAdd min 1;
+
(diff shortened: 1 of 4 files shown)


addons/medical/functions/fnc_addPain.sqf
```diff
@@ -0,0 +1,28 @@
...
+params ["_unit", "_painToAdd"];
+
+_painToAdd = _painToAdd min 1;
+
+private _pain = _unit getVariable [QEGVAR(medical,pain), 0];
+_pain = _pain + _painToAdd * (1 - _pain);
+
+_unit setVariable [QEGVAR(medical,pain), _pain];
+
+[_unit, "hit", PAIN_TO_SCREAM(_painToAdd)] call EFUNC(medical_engine,playInjuredSound);
```
Glowbal: I don't think we should add this to the addPain function. A wrapper that both adds pain and plays the sound would be preferable.

addons/medical/functions/fnc_addPain.sqf
```diff
@@ -0,0 +1,28 @@
+/*
+ * Author: commy2
+ * Adds pain to unit and plays a scream sound.
```
Glowbal: This function can currently also reduce pain.

Also, there is an upper limit but no lower limit enforced here.

commy2: Actually removing pain is scaled down by the already present pain as well. I'll fix that too.

addons/medical/functions/fnc_addPain.sqf
commy2: I just moved the playSound stuff back to the woundHandler(SQF). I think it will only be used there.

PR #4539, opened 2016-10-10
add function to disable dogtags for factions, and disable for civilians

commy2:
- adds function to disable dogtags for certain factions (e.g. rebels)
- use it to disable dogtags for civilians

### DIFF
--- a/addons/dogtags/functions/fnc_canTakeDogtag.sqf
+++ b/addons/dogtags/functions/fnc_canTakeDogtag.sqf
@@ -20,4 +20,7 @@ params ["_player", "_target"];
 
 if (isNull _target) exitWith {false};
 
+// check if disabled for faction
+if ([GVAR(disabledFactions) getVariable faction _target] param [0, false]) exitWith {false};
+
 (!alive _target) || {_target getVariable ["ACE_isUnconscious", false]}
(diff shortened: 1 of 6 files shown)


addons/dogtags/functions/fnc_canTakeDogtag.sqf
```diff
@@ -20,4 +20,7 @@ params ["_player", "_target"];
 
 if (isNull _target) exitWith {false};
 
+// check if disabled for faction
+if ([GVAR(disabledFactions) getVariable faction _target] param [0, false]) exitWith {false};
```
Glowbal: Only works now because we create the CBA namespace on mission start by calling the disableFactionDogtags. If we ever change that, this will break due to undefined variable.

addons/dogtags/functions/fnc_disableFactionDogtags.sqf
```diff
@@ -0,0 +1,24 @@
...
+ * Example:
+ * "CIV_F" call ace_dogtags_fnc_disableFactionDogtags
+ *
+ * Public: Yes
+ */
+#include "script_component.hpp"
+
+params [["_faction", "", [""]]];
+
+if (isNil QGVAR(disabledFactions)) then {
```
Glowbal: Maybe this should be moved to postInit?

PR #4546, opened 2016-10-12
Reset animation speed when not swimming

PabstMirror:


commy2:
This will overwrite my anim speed for the medical treatments. : /

PabstMirror:
We can fix it for 3.9, 
but I would like this for 3.8 so you are not permanently slowed down when getting out of water.

PR #4478, opened 2016-09-28
Show correct elev for mortar with no ammo

PabstMirror:
Fix #3757

Seems to be working, will want to retest after laser rewrite merged.

PR #4548, opened 2016-10-12
Fix mortar changes for laser shootRay changes

PabstMirror:
Ref #4478


austinjames314:
Tested with a Vanilla mortar, with air resistance and ammo handling on. Placed a mortar on a slope. There's still that missive jump between pointing at the ground, and pointing at the sky, but the jump when reloading is gone, which fixes the main problem.

PR #4550, opened 2016-10-14
Dockerfile

bux578:
- Will add a Dockerfile to run jekyll in a container. This will allow you to NOT install all jekyll requirements on your PC.

_Note: this might need improvements with volumes etc..._


docs/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Run:
+# docker run -p 4000:4000 -it ace3mod/jekyll
+
+FROM starefossen/github-pages:onbuild
+MAINTAINER bux
+
+COPY . /usr/src/app
+WORKDIR /usr/src/app
+
+RUN gem install --no-document \
```
Glowbal: This seems unncessary. Doesn't the base image already cover this?

docs/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# docker build -t ace3mod/jekyll .
+#
+# Run:
+# docker run -p 4000:4000 -it ace3mod/jekyll
+
+FROM starefossen/github-pages:onbuild
+MAINTAINER bux
+
+COPY . /usr/src/app
+WORKDIR /usr/src/app
```
Glowbal: Already done in the base image

docs/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN gem install --no-document \
+  github-pages:91 \
+  jekyll-github-metadata
+
+ENTRYPOINT ["jekyll", "serve", "--future", "--incremental", "--config", "_config_dev.yml", "-H", "0.0.0.0", "-P", "4000"]
```
Glowbal: Use `CMD` instead. No reason to use `ENTRYPOINT`.

docs/Dockerfile
Glowbal: Ah, you use the `onbuild` image for this. Probably want to switch to the `latest` instead?

docs/Dockerfile
Glowbal: same as below, switch to latest, this is already done.

If not, the `COPY` on line 10 is unncessary.

docs/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Build:
+# docker build -t ace3mod/jekyll .
+#
+# Run:
+# docker run -p 4000:4000 -it ace3mod/jekyll
+
+FROM starefossen/github-pages:onbuild
```
Glowbal: I think using `latest` is probably a better solution for us.

Glowbal:
Suggestion for the compose file:

```yml
version: '2'

services:
  docs:
    build: .
    restart: always
    ports:
      - "4000:4000"
    volumes:
      - .:/usr/src/app:rw

```

Note, probably double check the volume, I'm not sure if `.:/usr/src/app` works. Maybe `"$PWD":/usr/src/app` would be the better option.

docs/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Build:
+# docker build -t ace3mod/jekyll .
+#
+# Run:
+# docker run -p 4000:4000 -it ace3mod/jekyll
+
+FROM starefossen/github-pages:latest
+MAINTAINER bux
+
+VOLUME "/usr/src/app"
```
Glowbal: Should we also copy the contents of `.` to `/usr/src/app` ? Means you won't have too mount in a volume always, but still can.

bux578:
@Glowbal Should be good to go now.

bux578:
👍

PR #4551, opened 2016-10-14
fix cargo findNearestVehicle

Djzonk:
- add the ability to exclude objects from the findNearestVehicle function. For situations like in the image where a player might make an object like the quad loadable, and the findNearestVehicle is reporting the quad as still the closest even though im clipping into the truck.

- Will Also prevent below from happening if a player decides to make an object capable of being cargo as well loadable.

- As i don't relay know what happens when you attack a object to itself.
- Oh and update the files that call this function with the new parameter.


Djzonk:
Debug left on in case of possible changes

Djzonk:
- Added occupancy check when loading object

addons/cargo/functions/fnc_canLoadItemIn.sqf
```diff
@@ -19,6 +19,8 @@
 params [["_item", "", [objNull,""]], "_vehicle"];
 
 if (speed _vehicle > 1 || {((getPos _vehicle) select 2) > 3}) exitWith {TRACE_1("vehicle not stable",_vehicle); false};
+// TODO: use unitIsUAV command for uav check once out of devbuild
+if !(_item isEqualType "" || {getNumber (configFile >> "CfgVehicles" >> typeOf _item >> "isUav") == 1} || {{alive _x} count crew _item == 0}) exitWith {TRACE_1("item is occupied",_item); false};
```
commy2: sorry, I wasn't clear enough on this one.
This check is insufficient. Think of the Stomper UAV, which can carry persons. I think the check should be:
```
if (_item isEqualType objNull && {{alive _x && {getText (configFile >> "CfgVehicles" >> typeOf _x >> "simulation") != "UAVPilot"}} count crew _item > 0}) exitWith {
    TRACE_1("item is occupied",_item);
    false
};
```

PR #4552, opened 2016-10-15
fix Kestrel without adv. ballistics

### LINKED ISSUE #4544: Kestrel 4500 broken without advanced ballistics
**Arma 3 Version:** `1.64.138732` (stable)
**CBA Version:** `3.1.0.160928` (stable)
**ACE3 Version:** `3.7.0` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- The Kestrel 4500 impeller continuously spins at the same speed regardless of the direction the first person view is facing at and the instantaneous wind speed screen also displays a constant value regardless of the direction.

**Steps to reproduce:**
- Just equip the player with a Kestrel, set it to display instantaneous wind speed and rotate the player along the vertical axis.

- Editor (Singleplayer)

**Placed Modules:**
- None.

commy2:
- IDK why it uses different formulas depending on whether adv. ballisitcs is enabled or not. It clearly says "Head wind" and "Cross wind", so it makes no sense that both tabs report the same with "basic wind"

### DIFF
--- a/addons/kestrel4500/functions/fnc_generateOutputData.sqf
+++ b/addons/kestrel4500/functions/fnc_generateOutputData.sqf
@@ -132,13 +132,8 @@ if (GVAR(referenceHeadingMenu) == 0) then {
         };
         case 3: { // CROSSWIND
             if (!GVAR(MinAvgMax)) then {
-                if (missionNamespace getVariable [QEGVAR(advanced_ballistics,enabled), false]) then {
-                    _textCenterBig = Str(round(abs(sin(GVAR(RefHeading) - _playerDir) * _windSpeed) * 10) / 10);
-                    _textInfoLine1 = format["%1 m/s @ %2", round((abs(cos(_playerDir - _windDir)) * _windSpeed) * 10) / 10, round(_playerDir)];
-                } else {
-                    _textCenterBig = Str(round(abs(sin(GVAR(RefHeading)) * _windSpeed) * 10) / 10);
-                    _textInfoLine1 = format["%1 m/s @ %2", round(_windSpeed * 10) / 10, round(_windDir)];
-                };
+                _textCenterBig = Str(round(abs(sin(GVAR(RefHeading) - _playerDir) * _windSpeed) * 10) / 10);
+                _textInfoLine1 = format["%1 m/s @ %2", round((abs(cos(_playerDir - _windDir)) * _windSpeed) * 10) / 10, round(_playerDir)];
                 _textInfoLine2 = "- set heading";
             } else {
                 _textCenterLine1Left = "Max";
@@ -164,13 +159,8 @@ if (GVAR(referenceHeadingMenu) == 0) then {
         };
         case 4: { // HEADWIND
             if (!GVAR(MinAvgMax)) then {
-                if (missionNamespace getVariable [QEGVAR(advanced_ballistics,enabled), false]) then {
-                    _textCenterBig = Str(round(cos(GVAR(RefHeading) - _playerDir) * _windSpeed * 10) / 10);
-                    _textInfoLine1 = format["%1 m/s @ %2", round((abs(cos(_playerDir - _windDir)) * _windSpeed) * 10) / 10, round(_playerDir)];
-                } else {
-                    _textCenterBig = Str(round(cos(GVAR(RefHeading)) * _windSpeed * 10) / 10);
-                    _textInfoLine1 = format["%1 m/s @ %2", round(_windSpeed * 10) / 10, round(_windDir)];
-                };
+                _textCenterBig = Str(round(cos(GVAR(RefHeading) - _playerDir) * _windSpeed * 10) / 10);
+                _textInfoLine1 = format["%1 m/s @ %2", round((abs(cos(_playerDir - _windDir)) * _windSpeed) * 10) / 10, round(_playerDir)];
                 _textInfoLine2 = "- set heading";
             } else {
                 _textCenterLine1Left = "Max";


commy2:
ref: #4544

PabstMirror:
Do we also want to change https://github.com/acemod/ACE3/blob/master/addons/kestrel4500/functions/fnc_measureWindSpeed.sqf#L23

I think the only difference between AB and normal should be the call to EFUNC(weather,calculateWindSpeed);

commy2:
I think we want to use ` _windSpeed = abs(cos(_playerDir - _windDir)) * _windSpeed;` in either case

J3aLeiNe:
All Kestrel's functions are now available for the normal ballistic ?

commy2:
They are, but idk if they are supposed to.

J3aLeiNe:
The wind deflection is enabled by default, why not all wind's fonctions activated in the Kestrel ?
Imo, if something should be disabled with the normal ballistic, it's all fonctions concerning the atmospheric parameters. But anyway, thanks @commy2, @PabstMirror is happy now :p

commy2:
Now I'm confused. The Kestrel will show Head and Crosswind in "basic ballistic" even with this PR. It does the same atm too, but it actually shows the total wind speed in both modes, Head and Cross. That's all this PR fixes.

J3aLeiNe:
If i can modestly a suggestion : the AB's Kestrel for the both ballistics (if possible, i'm not scripter). 
If there's a another issue for this toy in the future, no more ballistics's problem. 
And the "set heading" function makes perfect sense in the default ballistic (used for a corrected and effective crosswind on the target).

esteldunedain:
Just for clarification: 
- IIRC the idea of `_windSpeed = abs(cos(_playerDir - _windDir)) * _windSpeed;` comes from AGM, along with basic wind deflection.
- When Ruthberg ported AB he stated that it didn't make sense to require to rotate the player 90º to measure crosswind because IRL you only rotate your hand. Rotating your whole body in Arma is slow, induces disorientation and may be very visible from afar. 

I think that's a fair argument. Moreover, it doesn't make sense to have the Kestrel be more complex in basic ballistic mode compared to AB.

Hence I think it would make sense to get rid of the `cos` on  basic ballistics and make it report headwind and crosswind sepparately on both modes.

commy2:
But advanced ballistics is usig L24 - 26. Those should be used in both systems, right?

esteldunedain:
> But advanced ballistics is usig L24 - 26. Those should be used in both systems, right?
> I think I wasn't clear enough with my comment.


Mmm, you are right. This is weird for sure. 

I think for compatibility neither mode should have the projection code on `fnc_measureWindSpeed.sqf` and, both should have it on `fnc_generateOutputData.sqf`, or viceversa. Currently AB have the cos applied twice (once on each function), and basic mode have it zero times.

PabstMirror:
Maybe I'm just not seeing it, but I don't get why we take `abs` in 
`_windSpeed = abs(cos(_playerDir - _windDir)) * _windSpeed;`

If wind is 5mps from the north,
Aim north and it reads 5
Aim east and it reads 0
Aim sound and it reads 5

How do you determine the wind direction?

esteldunedain:
> Maybe I'm just not seeing it, but I don't get why we take abs in 

Maybe the Kestrel is not able to tell which way the impeller is rotating_

commy2:
To bad Ruthberg is no longer around. : /

bux578:
The impeller's rotor is shaped that when wind comes from the front in turns into a specific direction (cw or ccw) when the wind is coming from behind it turns into the other direction. That's how it measures direction.

So it should be:

If wind is 5mps from the north,
Aim north and it reads 5
Aim east and it reads 0
Aim south and it reads -5

ulteq:
> To bad Ruthberg is no longer around. : /

I'm here. And I feel bad for being away from the project for such a long time.

> The impeller's rotor is shaped that when wind comes from the front in turns into a specific direction (cw or ccw) when the wind is coming from behind it turns into the other direction.

That is true.

> That's how it measures direction, AFAIK.

No, the Kestrel itself does not measure the wind direction at all. It always displays positive wind speed readings regardless of the direction the impeller is spinning at.

> IIRC the idea of _windSpeed = abs(cos(_playerDir - _windDir)) * _windSpeed; comes from AGM

It has always been like this: https://forums.bistudio.com/topic/171500-kestrel-4500-kestrel-4500-pocket-weather-tracker-wip/

> When Ruthberg ported AB he stated that it didn't make sense to require to rotate the player 90º to measure crosswind because IRL you only rotate your hand.

Yes, I thought about using `_headDir` instead of `_playerDir`. But I didn't really like it.

>  think we want to use `_windSpeed = abs(cos(_playerDir - _windDir)) * _windSpeed;` in either case

The basic ballistics code replicates the behavior of the Kestrel4500 in Arma2 ACE2. You only have to press one button to get a correct crosswind reading.

> Currently AB have the cos applied twice (once on each function)

At a first glance I would say that is a bug. It was probably introduced by this: https://github.com/acemod/ACE3/commit/4355df118cc4407358bd2b6bbe6affcb8756c107

J3aLeiNe:
@ulteq very, very happy to read you and i'm sure that it's the same for many people.
(as you see, my english is always so bad ...)

ulteq:
Yes, those two lines are wrong:

https://github.com/acemod/ACE3/blob/master/addons/kestrel4500/functions/fnc_generateOutputData.sqf#L137
https://github.com/acemod/ACE3/blob/master/addons/kestrel4500/functions/fnc_generateOutputData.sqf#L169
```
_textInfoLine1 = format["%1 m/s @ %2", round((abs(cos(_playerDir - _windDir)) * _windSpeed) * 10) / 10, round(_playerDir)];
```
should be:
```
_textInfoLine1 = format["%1 m/s @ %2", round(_windSpeed * 10) / 10, round(_playerDir)];
```
in both cases.

esteldunedain:
@ulteq, it's nice to hear from you man. Thank you for clarifying this.

commy2:
replaced by #4609

PR #4555, opened 2016-10-16
Add garrison zeus modules

alganthe:
- Add garrison and un-garrison functions to common.
- Add a garrison module.
- Add an un-garrison module.


addons/common/stringtable.xml
```diff
@@ -874,5 +874,21 @@
...
+        <Key ID="STR_ACE_Common_GarrisonInvalidPosition">
+            <English>Invalid position provided.</English>
+            <French>Position invalide fourni</French>
+        </Key>
+        <Key ID="STR_ACE_Common_GarrisonNoUnits">
+            <English>No units provided.</English>
+            <French>Aucune unité fourni</French>
+        </Key>
+        <Key ID="STR_ACE_Common_GarrisonNotEnoughPos">
+            <English>There isn't enough positions to place all units.</English>
```
SilentSpike: aren't

addons/zeus/functions/fnc_moduleGarrison.sqf
```diff
@@ -0,0 +1,57 @@
...
+*/
+
+#include "script_component.hpp"
+
+params ["_logic", "_pos", "_radius" ,"_mode" , "_topDownMode"];
+
+private _unit = (attachedTo _logic);
+private _building = nearestBuilding (getPosASL _unit);
+
+// Handles errors
```
SilentSpike: Validation of the module target should only be taking place in the ui function (it doesn't make sense to show zeus a UI and then tell them they placed the module wrong only after they close it).

addons/zeus/CfgVehicles.hpp
```diff
@@ -174,4 +174,16 @@ class CfgVehicles {
         function = QFUNC(moduleUnconscious);
         icon = QPATHTOF(UI\Icon_Module_Zeus_Unconscious_ca.paa);
     };
+    class GVAR(moduleGarrison): GVAR(moduleBase) {
+        curatorCanAttach = 1;
+        displayName = CSTRING(ModuleGarrison_DisplayName);
+        curatorInfoType = QGVAR(RscGarrison);
+        icon = QPATHTOF(UI\Icon_Module_Zeus_Garrison_ca.paa);
+    };
+    class GVAR(moduleUnGarrison): GVAR(moduleBase) {
```
SilentSpike: Not sure if I like having a separate module to undo the effects of another, could combine these into 1 by having a UI option to un-garrison - that way there's a single module that handles all things garrison.

Though that's just a thought, very much unsure what works better in terms of UX (if anyone else has thoughts on this please chime in).

addons/common/XEH_PREP.hpp
```diff
@@ -46,6 +46,7 @@ PREP(fixCollision);
 PREP(fixFloating);
 PREP(fixLoweredRifleAnimation);
 PREP(fixPosition);
+PREP(garrison);
```
SilentSpike: Unsure if these functions belong in common or not :stuck_out_tongue_closed_eyes:

addons/common/functions/fnc_unGarrison.sqf
```diff
@@ -0,0 +1,45 @@
...
+                _x doFollow _leader;
+            } else {
+                [QGVAR(doFollow), [_x, _leader], _leader] call CBA_fnc_targetEvent;
+            };
+
+        } else {
+            if (local _x) then {
+                _x doMove ((nearestBuilding (getPos _x)) buildingExit 0);
+            } else {
+                [QGVAR(doMove), [_x, ((nearestBuilding (getPos _x)) buildingExit 0)], _x] call CBA_fnc_targetEvent;
```
SilentSpike: Rather than sending an event for every unit from within this function, you should be running this whole function via an event from the module function.

addons/common/functions/fnc_unGarrison.sqf
```diff
@@ -0,0 +1,45 @@
...
+params [["_unitsArray", [], [[]]]];
+
+{
+    if !(isPlayer _x) then {
+        if (local _x) then {
+            _x enableAI "AUTOCOMBAT";
+            _x enableAI "PATH";
+        } else {
+            [QGVAR(enableAI), [_placedUnits, "AUTOCOMBAT"], _placedUnits] call CBA_fnc_targetEvent;
+            [QGVAR(enableAI), [_placedUnits, "PATH"], _placedUnits] call CBA_fnc_targetEvent;
```
SilentSpike: Not sure why this is being done inside this loop (also this variable doesn't exist).

addons/zeus/functions/fnc_moduleGarrison.sqf
commy2: Only true if the zeus interface updates in real time. I think it's fine to have the check in both

addons/common/functions/fnc_unGarrison.sqf
commy2: :+1:

Also to the note above:

`CBA_fnc_targetEvent` will send the event to every machine where at least one element of the array is local. If there are object with different localities in the array, it will execute the commands on each of those machines for every object, even if remote. But I guess that would still work for these.

addons/common/XEH_PREP.hpp
commy2: I'd say ace_ai, despite it not having any sqf atm.

commy2:
`pushback` should be `pushBack`, alganthe :)

alganthe:
I made some changes related to the unGarrison function, it's called as an event via the module and should only run where the units are local thus I removed the disableAI, follow, and doMove events I had added.

addons/common/XEH_PREP.hpp
jonpas: Doesn't CBA now have garrison functions that could be used instead? If not, then yeah, those should go into `ace_ai` component.

addons/zeus/CfgVehicles.hpp
jonpas: We should decide with what and stick with it. Currently there is already add/remove Virtual Arsenal which is 2 modules. Can always streamline into 1 later if it would be better.

addons/ai/XEH_PREP.hpp
```diff
@@ -0,0 +1,2 @@
+PREP(garrison);
+PREP(unGarrison);
```
jonpas: No empty newline.

addons/ai/XEH_postInit.sqf
```diff
@@ -0,0 +1,8 @@
+#include "script_component.hpp"
+
+[QGVAR(disableAI), {
+    params [["_units", [], [[]]], "_section"];
+    {_x disableAI _section} foreach (_units select {local _x});
+}] call CBA_fnc_addEventHandler;
+
+[QGVAR(unGarrison), FUNC(unGarrison)] call CBA_fnc_addEventHandler;
```
jonpas: No empty newline.

addons/ai/functions/fnc_garrison.sqf
```diff
@@ -0,0 +1,173 @@
+/*
+ * Author: alganthe
+ * Garrison function used to garrison AI inside buildings.
+ *
+ * Arguments:
+ * 0: The building(s) nearest this position are used <POSITION>
+ * 1: Limit the building search to those type of building <ARRAY>
+ * 2: Units that will be garrisoned <ARRAY>
+ * 3: Radius to fill building(s) <SCALAR> default: 50
```
jonpas: `> (default: 50)`
In other places too.

addons/ai/functions/fnc_garrison.sqf
```diff
@@ -0,0 +1,173 @@
...
+ * 4: 0: even filling, 1: building by building, 2: random filling <SCALAR> default: 0
+ * 5: True to fill building(s) from top to bottom <BOOL> default: false
+
+ * Return Value:
+ * Array of units not garrisoned
```
jonpas: ` * Units not garrisoned <ARRAY>`

addons/ai/functions/fnc_unGarrison.sqf
```diff
@@ -0,0 +1,39 @@
+/*
+ * Author: alganthe
+ * Used to un-garrison units
```
jonpas: Missing final dot.

addons/zeus/functions/fnc_moduleGarrison.sqf
```diff
@@ -0,0 +1,62 @@
+/*
+ * Author: alganthe
+ * Module calling the garrison function
```
jonpas: Missing final dot.

addons/zeus/functions/fnc_moduleGarrison.sqf
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Arguments:
+ * 0: The module logic <OBJECT>
+ * 1: Position of the module <POSITION>
+ * 2: Radius of the task <NUMBER>
+ * 3: Filling mode of the garrison function <NUMBER>
+ * 4: Enable or not top down filling <BOOLEAN>
```
jonpas: `<BOOL>`

addons/zeus/functions/fnc_moduleUnGarrison.sqf
```diff
@@ -0,0 +1,50 @@
+/*
+ * Author: alganthe
+ * Un-garrison a garrisoned group
```
jonpas: Missing final dot.

addons/zeus/functions/fnc_ui_garrison.sqf
```diff
@@ -0,0 +1,103 @@
+/*
+ * Author: alganthe
+ * Initalises the `garrison` zeus module display
```
jonpas: Missing final dot.

addons/zeus/functions/fnc_ui_garrison.sqf
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Arguments:
+ * 0: garrison controls group <CONTROL>
```
jonpas: `Garrison`

addons/zeus/functions/fnc_ui_garrison.sqf
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Return Value:
+ * Nothing
```
jonpas: ` * None`

jonpas:
@SilentSpike ?

addons/ai/functions/fnc_garrison.sqf
```diff
@@ -0,0 +1,173 @@
...
+};
+
+private _buildings = [];
+
+if (_fillingRadius < 50) then {
+    _buildings = nearestObjects [_startingPos, _buildingTypes, 50];
+} else {
+    _buildings = nearestObjects [_startingPos, _buildingTypes, _fillingRadius];
+    _buildings = _buildings call BIS_fnc_arrayShuffle;
+};
```
jonpas: This whole thing would be neater as:
```sqf
private _buildings = nearestObjects [_startingPos, _buildingTypes, _fillingRadius];
if (_fillingRadius >= 50) then {
    _buildings = _buildings call BIS_fnc_arrayShuffle;
};
```

There is also [`CBA_fnc_shuffle`](https://github.com/CBATeam/CBA_A3/blob/master/addons/arrays/fnc_shuffle.sqf) that is probably faster than BI's function.

addons/ai/functions/fnc_unGarrison.sqf
```diff
@@ -0,0 +1,39 @@
...
+        if (_leader != _x) then {
+            doStop _x;
+            _x doFollow _leader;
+
+        } else {
+            _x doMove ((nearestBuilding (getPos _x)) buildingExit 0);
+        };
+    };
+} foreach _units;
+
```
jonpas: A line too much.

addons/zeus/functions/fnc_moduleGarrison.sqf
```diff
@@ -0,0 +1,62 @@
...
+    case (alive _unit): {
+        [LSTRING(OnlyAlive)] call _fnc_errorAndClose;
+    };
+    case (_unit distance _building < 500): {
+        [LSTRING(BuildingTooFar)] call _fnc_errorAndClose;
+    };
+};
+
+private _units = units _unit;
+// Make sure all units are disembarked.
```
jonpas: No final dot.

addons/zeus/functions/fnc_moduleGarrison.sqf
```diff
@@ -0,0 +1,62 @@
+/*
+ * Author: alganthe
+ * Module calling the garrison function.
+ *
+ * Arguments:
+ * 0: The module logic <OBJECT>
```
jonpas: `The` is superfluous.

addons/zeus/functions/fnc_moduleUnGarrison.sqf
```diff
@@ -0,0 +1,50 @@
+/*
+ * Author: alganthe
+ * Un-garrison a garrisoned group.
+ *
+ * Arguments:
+ * 0: The module logic <OBJECT>
```
jonpas: `The` is superfluous.

addons/zeus/functions/fnc_ui_garrison.sqf
```diff
@@ -0,0 +1,103 @@
+/*
+ * Author: alganthe
+ * Initalises the `Garrison` zeus module display.
```
jonpas: `"Garrison"`

addons/zeus/functions/fnc_ui_garrison.sqf
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Arguments:
+ * 0: garrison controls group <CONTROL>
```
jonpas: `Garrison ...`

alganthe:
This is still very slightly WIP, I have an issue (units are sometimes stack ontop of eachothers when the function is called multiple time), thus I enabled some debug elements.

alganthe:
AI stacking on rare occasion has been fixed, debug disabled this is now ready for review.

setPos is pretty imprecise, moving units can end up within 2m of the position if they are moving at full running speed, I cannot fix this without inducing some more network usage which I don't want.

alganthe:
As @jonpas mentionned in PM zeus shouldn't teleport units by default, I'll add a "teleport" button like the top-down filling one in the dialog.

When ticked it will behave like it does right now, otherwise it will use CBA's garrison, my ungarrison module works on both (it's also needed because you have no other way of freeing units garrisoned via the defend module we already have.)

addons/zeus/stringtable.xml
```diff
@@ -1112,6 +1112,10 @@
             <English>Random filling</English>
             <French>Remplir au hasard</French>
         </Key>
+        <Key ID="STR_ACE_Zeus_ModuleGarrison_TeleportText">
```
jonpas: https://github.com/acemod/ACE3/blob/98da86b74d4089bc3403d0b0c0074e6a46ce35d8/addons/respawn/stringtable.xml#L29-L38

alganthe: What do you suggest? 
That I take the first word of the string? It could be different on asian languages.

jonpas: Just that we have some of them, I have no clue which are correct without the "base" part as well, might as well just leave it.

alganthe:
Okay, this should be ready for review.

alganthe:
Outside of some situation where the AI completely refuses to path this should now work quite nicely now.

addons/zeus/functions/fnc_moduleUnGarrison.sqf
```diff
@@ -0,0 +1,50 @@
...
+ * Example:
+ * [LOGIC] call ace_zeus_fnc_moduleUngarrison
+ *
+ * Public: No
+*/
+
+#include "script_component.hpp"
+
+params ["_logic"];
+
```
PabstMirror: `if (!local _logic) exitWith {}; // Module is global`

PR #4580, opened 2016-10-21
Javelin overhaul

### LINKED ISSUE #2686: Aqquirung the targer with javelin is almost impossible - Ai do it too fast
Aquire the target with ACE javeling is piece of art - first i need to aim little higher than target, second - frames around it are locking on all around it, makes aquire impossible in so small time (like AI). Swaying is killing it the most - beacuse of swaying i cant hold the sight on the target, so it is aquiring all round including building, objects, covers. try to aim the tank standing behind the sandabgs, and having building behind itself (2 layers of sandbags - one at 0 height, second on 0.5. Its impossible. Off course try to aim it inder fire and from the same haight as sandbags are.

### LINKED ISSUE #3241: Titan Compact Lock flickering
ACE3 Version: 3.4.2
Stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Placed ACE3 Modules:**
- None

**Description:**
Titan missile locks can flicker on and off most commonly at moving or targets surrounded by foliage.This seems to occur in MP and may be to do with Server updating the client on location. I've yet to be able to reproduce it in SP.

During this lock flickering Its impossible to launch the missile and is incredibly annoying and leads to avoidable deaths.  

**Steps to reproduce:**
1. Join ACE3 server (preferably with slight lag and non opermal performance)
2. Take Titan Compact Launcher and AT missile
3. Attempt to target moving or semi concealed vehicle 
4. Observe Flickering

 "Multiplayer"

PabstMirror:
- New target detection function won't lose lock as easily (will close #2686 - close #3241)
- Removed unneeded firedEH
- Animated track gates and re-added main crosshair
- Fixed bug with `onLoad` not always being called when switching to the weapon


esteldunedain:
:+1: Nice @PabstMirror

addons/javelin/functions/fnc_mapHelperDraw.sqf
```diff
@@ -0,0 +1,52 @@
+/*
+ * Author: PabstMirror
+ * Handles the map helper's draw event
+ * Resets arguemnts if not run recently
```
ulteq: `arguemnts` -> `arguments`

PR #4581, opened 2016-10-21
Cleanup Frag Component

654wak654:
- Do some ~~minor~~ cleanup in `ace_frag` (not so minor anymore with all the commits).
- ~~The PR looks bigger than it is because of simple whitespace changes.~~ It is as big as it looks please review carefully,  earlier on I broke spalling with one of my "optimizations".
- ~~Not much has been tested since the first commit, not sure if this can make it to 3.8.1 without proper testing. Will try to test with people from my unit this monday.~~ Update: Had a lot of fun with explosions, found & fixed some script errors,


PabstMirror:
Looks good.
Need to fix merge conflicts.

addons/frag/functions/fnc_addTrack.sqf
```diff
@@ -6,15 +6,12 @@ if (GVAR(autoTrace)) then {
...
+params ["_origin", "_obj", ["_color", [1,0,0,1]]];
+
 private _positions = [];
 private _objSpd = vectorMagnitude (velocity _obj);
-_positions set[(count _positions), [(getPos _obj), _objSpd]];
+_positions set[count _positions, [getPos _obj, _objSpd]];
 private _data = [_origin, typeOf _origin, typeOf _obj, _objSpd, _positions, _color];
 
-GVAR(traces) set[_index, _data];
+GVAR(traces) pushBack _data;
```
654wak654: Smart catch, will do.

654wak654:
Updated PR body to reflect current state.

addons/frag/functions/fnc_doExplosions.sqf
```diff
@@ -1,20 +1,20 @@
...
-if(_index >= (count _explosions)) then {
-    [(_this select 1)] call CBA_fnc_removePerFrameHandler;
-} else {
-    _params set[1, _index];
+
+ADD(_index,2);
+
+if (_index >= count _explosions) then {
+    [_pfhID] call CBA_fnc_removePerFrameHandler;
 };
```
PabstMirror: Needs to be set in the array
This will fix #4590 (in old code `_params` was undefined)
```
if (_index >= count _explosions) then {
    [_pfhID] call CBA_fnc_removePerFrameHandler;
} else {
    TRACE_1("setting arg",_index);
    _args set[1, _index];
};
```

addons/frag/XEH_postInit.sqf
```diff
@@ -19,10 +19,9 @@ if(isServer) then {
     ["ace_firedPlayerVehicle", DFUNC(fired)] call CBA_fnc_addEventHandler;
     ["ace_firedNonPlayerVehicle", DFUNC(fired)] call CBA_fnc_addEventHandler;
 
-    [FUNC(masterPFH), 0, []] call CBA_fnc_addPerFrameHandler;
-
+    [FUNC(masterPFH)] call CBA_fnc_addPerFrameHandler;
```
PabstMirror: Leave this as `[FUNC(masterPFH), 0, []] call CBA_fnc_addPerFrameHandler;`
`ENABLE_PERFORMANCE_COUNTERS` has a macro that expects 3 args.

KingZogg:
The cookoff problem is still ongoing.
Running 3.8.1
Why was the other issue closed ?

PabstMirror:
It will be fixed in next release (which should be 3.8.2)
for now you should be able to prevent issue if you disable frag relfections

bux578:
:picard:

PR #4585, opened 2016-10-22
Fix bleeding hit logic for enabledFor = 1

PabstMirror:
Logic was inverted, if player was hit it wouldn't create blood

### DIFF
--- a/addons/medical_blood/XEH_postInit.sqf
+++ b/addons/medical_blood/XEH_postInit.sqf
@@ -39,7 +39,7 @@ if (isServer) then {
 
     ["CAManBase", "hit", {
         params ["_unit"];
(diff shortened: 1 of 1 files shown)


addons/medical_blood/XEH_postInit.sqf
```diff
@@ -39,7 +39,7 @@ if (isServer) then {
 
     ["CAManBase", "hit", {
         params ["_unit"];
-        if (GVAR(enabledFor) == 1 && {!isPlayer _unit || {_unit == ACE_player}}) exitWith {};
+        if (GVAR(enabledFor) == 1 && {_unit != ACE_player}) exitWith {};
```
Glowbal: `isPlayer` check should remain. `isPlayer _unit` does not always equal `ACE_player`. Example; in cases of remote control.

commy2:
merge?

PR #4586, opened 2016-10-22
Use single queue for blood cleanup

PabstMirror:
Using a single ordered queue for blood drop cleanup 
instead of having every drop (up to 500) on a `waitAndExecute`

Every call to `CBA_fnc_waitAndExecute` does a `sort`, so it's best to keep the array small.


commy2:
Can you merge this into medical-rewrite too?

PR #4587, opened 2016-10-23
Blocks AI getting back into the vehicle on cookoff

rebelvg:
- Changes header example
- Orders crew to move 100m away from the vehicle in random direction ASAP
- https://www.youtube.com/watch?v=KX1UOjsI4bA

### DIFF
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -120,6 +120,7 @@ if (local _vehicle) then {
                 _x leaveVehicle _vehicle;
                 
                 if (leader _x == _x) then {
+                    [group _x] call CBA_fnc_clearWaypoints;
                     [group _x, _randomPosition, 0, "MOVE", "AWARE", "BLUE", "FULL", "NO CHANGE", "(group this) setBehaviour 'COMBAT'; (group this) setCombatMode 'RED';"] call CBA_fnc_addWaypoint;
                 };
             };
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -1,18 +1,18 @@
 /*
- * Author: KoffeinFlummi, commy2
- * Start a cook-off in the given vehicle.
- *
- * Arguments:
- * 0: Vehicle <Object>
- *
- * Return Value:
- * None
- *
- * Example:
- * (vehicle player) call ace_cookoff_fnc_cookOff
- *
- * Public: No
- */
+* Author: KoffeinFlummi, commy2
+* Start a cook-off in the given vehicle.
+*
+* Arguments:
+* 0: Vehicle <Object>
+*
+* Return Value:
+* None
+*
+* Example:
+* [(vehicle player)] call ace_cookoff_fnc_cookOff
+*
+* Public: No
+*/
 #include "script_component.hpp"
 
 params ["_vehicle"];
@@ -113,14 +113,27 @@ if (local _vehicle) then {
 
         [_vehicle, _fnc_FlameEffect, 12] call _fnc_FlameEffect; // recursive function
 
+        private _randomPosition = [getPos _vehicle, 100, random 360] call BIS_fnc_relPos;
+
+        private _vehicleLeaders = crew _vehicle select {leader _x == _x};
+        
+        private _disableGetInUnits = crew _vehicle select {local _x && {!(_x call EFUNC(common,isPlayer))}};
+        
+        _disableGetInUnits allowGetIn false;
+
         {
             if (local _x && {!(_x call EFUNC(common,isPlayer))}) then {
                 _x action ["Eject", _vehicle];
+                
+                if (_x in _vehicleLeaders) then {
+                    [group _vehicle, _randomPosition, 0, "MOVE", "AWARE", "WHITE", "FULL"] call CBA_fnc_addWaypoint;
+                    [group _vehicle, _randomPosition, 0, "MOVE", "COMBAT", "RED", "FULL"] call CBA_fnc_addWaypoint;
+                };
             };
         } forEach crew _vehicle;
 
         [{
-            params ["_vehicle", "_effects"];
+            params ["_vehicle", "_effects", "_disableGetInUnits"];
 
             {
                 deleteVehicle _x;
@@ -128,7 +141,9 @@ if (local _vehicle) then {
 
             if (local _vehicle) then {
                 _vehicle setDamage 1;
+                
+                _disableGetInUnits allowGetIn true;
             };
-        }, [_vehicle, _effects], 4 + random 20] call CBA_fnc_waitAndExecute;
+        }, [_vehicle, _effects, _disableGetInUnits], 4 + random 20] call CBA_fnc_waitAndExecute;
     }, [_vehicle, _effects, _positions], 3 + random 15] call CBA_fnc_waitAndExecute;
 }, _vehicle, 0.5 + random 5] call CBA_fnc_waitAndExecute;
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -113,7 +113,7 @@ if (local _vehicle) then {
 
         [_vehicle, _fnc_FlameEffect, 12] call _fnc_FlameEffect; // recursive function
 
-        private _randomPosition = [getPos _vehicle, 100, random 360] call BIS_fnc_relPos;
+        private _randomPosition = _vehicle getPos [100, random 360];
 
         private _disableGetInUnits = crew _vehicle select {local _x && {!(_x call EFUNC(common,isPlayer))}};
         
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -115,13 +115,9 @@ if (local _vehicle) then {
 
         private _randomPosition = _vehicle getPos [100, random 360];
 
-        private _disableGetInUnits = crew _vehicle select {local _x && {!(_x call EFUNC(common,isPlayer))}};
-        
-        _disableGetInUnits allowGetIn false;
-
         {
             if (local _x && {!(_x call EFUNC(common,isPlayer))}) then {
-                _x action ["Eject", _vehicle];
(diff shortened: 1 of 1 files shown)


addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -113,22 +113,35 @@ if (local _vehicle) then {
...
         {
             if (local _x && {!(_x call EFUNC(common,isPlayer))}) then {
                 _x action ["Eject", _vehicle];
+                
+                if (leader _x == _x) then {
+                    [group _x, _randomPosition, 0, "MOVE", "AWARE", "WHITE", "FULL"] call CBA_fnc_addWaypoint;
+                    [group _x, _randomPosition, 0, "MOVE", "COMBAT", "RED", "FULL"] call CBA_fnc_addWaypoint;
+                };
             };
         } forEach crew _vehicle;
```
654wak654: `array select code` could be used here.

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -113,22 +113,35 @@ if (local _vehicle) then {
 
         [_vehicle, _fnc_FlameEffect, 12] call _fnc_FlameEffect; // recursive function
 
+        private _randomPosition = [getPos _vehicle, 100, random 360] call BIS_fnc_relPos;
```
commy2: `private _randomPosition = _vehicle getPos [100, random 360]`;

- is 100 meters too much?

rebelvg: Ammo starting to cookoff and crew sometime gets killed. 100m is a pretty safe distance.

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -113,22 +113,35 @@ if (local _vehicle) then {
...
+
+        private _disableGetInUnits = crew _vehicle select {local _x && {!(_x call EFUNC(common,isPlayer))}};
+        
+        _disableGetInUnits allowGetIn false;
```
rebelvg: I tried unassignVehicle alone and it doesn't work correctly, only 100% solution is allowGetIn.

commy2: `lock` the vehicle ?

addons/cookoff/functions/fnc_cookOff.sqf
commy2: should still use the alternate syntax of `getPos` I posted instead of `BIS_fnc_relPos`

rebelvg: Oh yeah, that getPos looks better, will change.

addons/cookoff/functions/fnc_cookOff.sqf
rebelvg: Tried both, didn't work 100% reliably. Only allowGetIn worked right every time.

commy2:
Ok. Can you explain why you use `CBA_fnc_addWaypoint` twice on the same position?

rebelvg:
@commy2 Without the second one they'll just stand there, so it tells them to switch into a red combat mode on this waypoint. Same if you use red combat mode too early they'll probably get stuck near the vehicle looking for enemies.

commy2:
But why does it use engagement rule `"WHITE"` (Hold fire, engage at will) and not `"BLUE"` (Never fire)?
Why does the second waypoint set speed to `"FULL"` ?

commy2:
I think a careless, blue waypoint should be used (basically panic mode)
And the resetting of the original behaviour should be done with the onWaypoint script.

rebelvg:
I thought WHITE would help if there's someone in the way. With BLUE they'll just ignore them. I'll try careless and blue. And I'll to try replace the second waypoint with onWaypoint script thing.

commy2:
Sorry for being so petty, but I want this to be as slick as possible.

rebelvg:
With careless they're moving with their COOL walk, not running.

SilentSpike:
Not sure that I like waypoints being cleared, mission makers might be confused when AI no longer follow the path they've set in editor.

If you look at the CBA_fnc_searchNearby function you can see how a temporary self-deleting waypoint is added that resets their behaviour on completion.

bux578:
@SilentSpike I don't think that IRL a tank crew would follow their "waypoints" if the tank gets destroyed. Actually it's more realistic to have them scatter, stay safe and fall back/wait for evac.

rebelvg:
That was exactly the idea why I removed waypoints at first.

654wak654:
About my previous comment:

https://github.com/rebelvg/ACE3/pull/1

rebelvg:
So how should I go about it? Remove waypoints or leave it how it is atm?

commy2:
Keep them.

SilentSpike:
@bux578 Sure, but we're not making a reactive AI mod here. What if the mission developer has waypoints synchronised to things, etc. ?

I think it's fine to add a waypoint at the current index for them to evac the vicinity in combat mode or something, then after that they return to their business. If the mission developer wants them to evac if destroyed that can be added by other means.

654wak654:
@SilentSpike the crew whould have been dead without cookoff, never reaching their other waypoints. They should not follow their old waypoints.

bux578:
@SilentSpike I agree that we're not doing an AI mod. But we've introduced a feature that - as @654wak654 correctly pointed out and compared to vanilla - often saves the lives of the crew. 

Being a realism mod, we should take care of those poor souls that just got their life spared on the battlefield.

Cuel:
What about ejecting them and sending each member in a random direciton using `doMove`? Eventually they'll regroup?

654wak654:
Umm, I just thought of something: What if 2 tanks are in the same group? Should we seperate the crew from the group and let the other continue? Should they try to mount the other tank?

654wak654:
Maybe it can be dependent on skill. Guerilla thugs whould try to continue the mission for the motherland, but NATO crews whould detach from their group and wait.

PabstMirror:
I don't think we need an ace_setting, but we might want to add something like
`if (missionNamespace getVariable ["ace_cookoff_skipTouchingWaypoints", false]) exitWith {};`
so that other addons could disable this if it causes problems.

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -113,11 +113,20 @@ if (local _vehicle) then {
...
+            if (leader _x == _x) then {
+                private _wp = group _x addWaypoint [_randomPosition, 0, 1];
+                _wp setWaypointType "MOVE";
+                _wp setWaypointBehaviour "AWARE";
+                _wp setWaypointCombatMode "BLUE";
+                _wp setWaypointSpeed "FULL";
+                _wp setWaypointStatements ["TRUE", "(group this) setBehaviour 'COMBAT'; (group this) setCombatMode 'RED';"];
             };
-        } forEach crew _vehicle;
+        } forEach ((crew _vehicle) select {local _x && {!(_x call EFUNC(common,isPlayer))}});
```
commy2: now we are iterating through the array twice. the `if` check was totally fine.

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -113,11 +113,20 @@ if (local _vehicle) then {
...
-                _x action ["Eject", _vehicle];
+            _x leaveVehicle _vehicle;
+            
+            if (leader _x == _x) then {
+                private _wp = group _x addWaypoint [_randomPosition, 0, 1];
+                _wp setWaypointType "MOVE";
+                _wp setWaypointBehaviour "AWARE";
+                _wp setWaypointCombatMode "BLUE";
+                _wp setWaypointSpeed "FULL";
+                _wp setWaypointStatements ["TRUE", "(group this) setBehaviour 'COMBAT'; (group this) setCombatMode 'RED';"];
```
commy2: lower case `true`

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -113,11 +113,20 @@ if (local _vehicle) then {
...
 
+        private _randomPosition = _vehicle getPos [100, random 360];
+
         {
-            if (local _x && {!(_x call EFUNC(common,isPlayer))}) then {
-                _x action ["Eject", _vehicle];
+            _x leaveVehicle _vehicle;
+            
+            if (leader _x == _x) then {
+                private _wp = group _x addWaypoint [_randomPosition, 0, 1];
```
commy2: Needs optional parameter:
- index (optional): Number - can be used to insert a waypoint in between existing waypoints

https://community.bistudio.com/wiki/addWaypoint

otherwise the waypoint will be added as the last waypoint, not as the current one!
I think it should be always 0 or `currentWaypoint _group` or `currentWaypoint _group -1` or something similar

commy2: The waypoint also does not remove itself, which is what pabst proposed. very important for CYCLE behaviour

rebelvg: I'll take a look but last time I checked waypoint 0 was reserved for something and always gets put on the same position the unit is currently in, not on the position you placed it. I'll add a self-removing waypoint too.

addons/cookoff/functions/fnc_cookOff.sqf
rebelvg: Copied this from CBA. Will change to lowercase.

addons/cookoff/functions/fnc_cookOff.sqf
SilentSpike: `currentWaypount _group` Will make it their active waypoint. Also it was me who proposed that @commy2, but I'm quite happy to be be mistaken for pabst :laughing:

Oh and index 0 isn't _exactly_ reserved for anything, there's some weird behaviour regarding it, but you can definitely manipulate the waypoint at index 0 via scripting. The reason it wouldn't work as the current waypoint is because completed waypoints also take up indices - nothing decrements unless one is deleted.

commy2: Sorry, Silent. I'm bad with faces, names and apparently online aliases too it seems. :|

rebelvg:
I switched to doMove. It doesn't interrupt existing waypoints. If they had waypoints before, they will do move and continue with their waypoints. If not - they will move and stop there.

commy2:
will test again later today. looks good.

PR #4592, opened 2016-10-24
Fix cargo find nearest vehicle

Djzonk:
- Remaking of #4551 after a mess up in me trying to update my fork and being inexperienced with git 
- add the ability to exclude objects from the findNearestVehicle function. For situations like in the [image](https://camo.githubusercontent.com/e7201defd0e592011aca183596dc25bbed2ef8ce/687474703a2f2f692e696d6775722e636f6d2f42447571774c612e706e67) where a player might make an object like the quad loadable, and the findNearestVehicle is reporting the quad as still the closest even though im clipping into the truck.
- prevent object implosion. If a player decides to make an object capable of being cargo as well loadable.
- Added occupancy check when loading object.
- Update findNearestVehicle calls to pass new parameter.

### DIFF
--- a/addons/cargo/functions/fnc_canLoadItemIn.sqf
+++ b/addons/cargo/functions/fnc_canLoadItemIn.sqf
@@ -19,7 +19,7 @@
 params [["_item", "", [objNull,""]], "_vehicle"];
 
 if (speed _vehicle > 1 || {((getPos _vehicle) select 2) > 3}) exitWith {TRACE_1("vehicle not stable",_vehicle); false};
-// TODO: use unitIsUAV command for uav check once out of devbuild
+
 if (_item isEqualType objNull && {{alive _x && {getText (configFile >> "CfgVehicles" >> typeOf _x >> "simulation") != "UAVPilot"}} count crew _item > 0}) exitWith {
     TRACE_1("item is occupied",_item);
     false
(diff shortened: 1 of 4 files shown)


addons/cargo/functions/fnc_canLoadItemIn.sqf
```diff
@@ -19,6 +19,11 @@
 params [["_item", "", [objNull,""]], "_vehicle"];
 
 if (speed _vehicle > 1 || {((getPos _vehicle) select 2) > 3}) exitWith {TRACE_1("vehicle not stable",_vehicle); false};
+// TODO: use unitIsUAV command for uav check once out of devbuild
```
PabstMirror: `unitIsUAV` is ready now, but I don't think it will work for what we need here
It returns true for uav vehicles, but not for the AI pilots which is what we want.
.e.g. for the UGV Stomper the driver (`"B_UAV_AI"`) returns false 
(not sure why, I think it's meant for the vehicle not the units??)

So what you have right now is fine

Djzonk: should I remove the TODO so it do sent confuse anyone in the future

jonpas: Yeah, do it.

Glowbal:
Added to 3.8.2 branch

PR #4597, opened 2016-10-26
Add Zeus Module for FRIES

654wak654:
- Adds a zeus module to add/remove FRIES from helicopters
- It's called "add/remove" and not "toggle" because the FRIES isn't "toggled" but literally created or removed.
- Needs translation.
- I have a small zeus "cleanup" PR waiting on this.


addons/zeus/XEH_postInit.sqf
```diff
@@ -14,3 +14,4 @@ QGVAR(GlobalSkillAI) addPublicVariableEventHandler FUNC(moduleGlobalSetSkill);
 [QGVAR(modulePatrolArea), CBA_fnc_taskPatrol] call CBA_fnc_addEventHandler;
 [QGVAR(moduleSearchNearby), CBA_fnc_searchNearby] call CBA_fnc_addEventHandler;
 [QGVAR(moduleSearchArea), CBA_fnc_taskSearchArea] call CBA_fnc_addEventHandler;
+[QGVAR(equipFries), EFUNC(fastroping,equipFRIES)] call CBA_fnc_addEventHandler;
```
jonpas: Can just be added on server I guess since it won't be called outside of server.

addons/zeus/stringtable.xml
```diff
@@ -337,6 +337,15 @@
...
             <Japanese>捕虜としてトグル</Japanese>
         </Key>
+        <Key ID="STR_ACE_Zeus_ModuleAddOrRemoveFRIES_DisplayName">
+            <English>Add/Remove FRIES</English>
+        </Key>
+        <Key ID="STR_ACE_Zeus_NotFastRopeCompatible">
+            <English>%1 is not fastrope compatible.</English>
+        </Key>
+        <Key ID="STR_ACE_Zeus_CantRemoveFRIES">
+            <English>Can't remove FRIES, ropes are deployed.</English>
```
jonpas: "Unable to remove..." would be better.

addons/zeus/config.cpp
```diff
@@ -41,12 +41,18 @@ class CfgPatches {
...
+            QGVAR(moduleAddOrRemoveFRIES)
+        };
+    };
 };
 
 class ACE_Curator {
     GVAR(captives) = "ace_captives";
     GVAR(medical) = "ace_medical";
     GVAR(cargoAndRepair)[] = {"ace_cargo", "ace_repair"};
+    GVAR(fastRoping) = "ace_fastroping";
```
jonpas: I still think this should be `GVAR(fastroping)` like the module, everywhere else as well, `fastRoping` capitalization looks weird IMO. `GVAR(cargoAndRepair)` is exception because it's for 2 components.

jonpas:
Also, don't worry about translations, those come later by translators. ;)

654wak654:
I figured so, just thought I'd point it out for the record.

addons/zeus/functions/fnc_moduleAddOrRemoveFRIES.sqf
```diff
@@ -0,0 +1,53 @@
...
+} else {
+    (GETMVAR(BIS_fnc_curatorObjectPlaced_mouseOver,[""])) params ["_mouseOverType", "_mouseOverUnit"];
+
+    if (_mouseOverType != "OBJECT") then {
+        [LSTRING(NothingSelected)] call EFUNC(common,displayTextStructured);
+    } else {
+        if !(alive _mouseOverUnit) then {
+            [LSTRING(OnlyAlive)] call EFUNC(common,displayTextStructured);
+        } else {
+            _config = configFile >> "CfgVehicles" >> typeOf _mouseOverUnit;
```
jonpas: Missing private.

addons/zeus/functions/fnc_moduleAddOrRemoveFRIES.sqf
```diff
@@ -0,0 +1,53 @@
...
+            _displayName = getText (_config >> "displayName");
```
jonpas: Missing private.

addons/zeus/functions/fnc_moduleAddOrRemoveFRIES.sqf
```diff
@@ -0,0 +1,53 @@
...
+            if !(isNumber (_config >> QEGVAR(fastroping,enabled))) then {
+                [format [localize LSTRING(NotFastRopeCompatible), _displayName]] call EFUNC(common,displayTextStructured);
+            } else {
+                _fries = GETVAR(_mouseOverUnit,EGVAR(fastroping,FRIES),objNull);
```
jonpas: Missing private.

SageNTitled:
Will this allow FRIES to be equipped on helicopters which are not configured for it?

BaerMitUmlaut:
No. That's simply not possible because fastroping has no idea where to attach it to.

SageNTitled:
There's no way to assign default positions for helicopters without support? The ropes would be hanging from the hull, but at least FRIES would be usable on more than 2 helicopters. 😃

BaerMitUmlaut:
There is no way to automatically detect a good position for the ropes, ensuring that they dont clip into the helo, are near doors etc.

Every vanilla helicopter except the Little Bird supports fastroping, CUP has their helicopter properly configured and we provide a compat for RHS which adds it to their helicopters as well. I have no idea where you got your "only 2 helicopters" from.

SageNTitled:
Gotcha. There's only a couple of them who has the FRIES checkbox in 3den, so I missed a dozen. 😆

addons/zeus/XEH_postInit.sqf
```diff
@@ -2,7 +2,7 @@
 
 ["ace_settingsInitialized",{
     // Only add an InitPost EH if setting is enabled (and apply retroactively)
-    if (isServer && GVAR(autoAddObjects)) then {
+    if (isServer && {GVAR(autoAddObjects)}) then {
```
SilentSpike: It doesn't really matter, but would this not be slower in this case?

SilentSpike:
:+1: Forgot I meant to merge this one on monday

addons/zeus/XEH_postInit.sqf
commy2: It's always slower on the server, but faster on the clients. Since there are more clients than servers usually, this is overall faster.
Not that it matters.

Issue #4602, opened 2016-10-27
Vehicleammo cooking off when it's not supposed to!


Retch44:
**Arma 3 Version:** `1.64` (stable)
**CBA Version:** `3.0.0` (I can only asume from BI-Sign file name)
**ACE3 Version:** `3.7.0.6` (")

**Mods:**
- `@CBA_A3`
- `@ACE3 `
- `@RHS`
- `@other that wont interflict `

**Description:**
- Dealing critical damage to certain vehicle parts will cause the vehicleammo to start cooking off,
   I asume this happens as the commander turrets are counted as normal turrets.

**Steps to reproduce:**
- Place a vic with a commander turret, shoot the commander turret.

**Where did the issue occur?**
- Dedicated Server

**Video:**
http://plays.tv/video/5811804c36676267d3/thecourier-in-crewman-training?from=user
http://plays.tv/video/58112ba3d9cf42f93f/ty-ace-3?from=user

Retch44:
Can't provide any further Intel, watch the videos or try for youself.

jonpas:
Please update to latest CBA and ACE3 and try to reproduce. 

> @other that wont interflict

We'll decide that. Also please upload RPT.

commy2:
Needs RPT file

commy2:
I'll make this short and close it. This bug is fixed in the latest version. Please update your mods.

Retch44:
Just saying I updated to latest version, still exsists, I see what I can do about getting you some prove...

GIJed92:
not fixed

GIJed92:
This bug is not fixed in the latest version. Please Fix your mods.

jones140:
tested with latest ace and cba still happening video soon
http://pastebin.com/TXK6XsKL

jones140:
https://www.youtube.com/watch?v=EFiEJBECyDA&feature=youtu.be

commy2:
upload RPT please

Retch44:
Check the Pastebin links: 
http://pastebin.com/QkEqnZhc

jones140:
did you see this http://pastebin.com/TXK6XsKL

GIJed92:
http://plays.tv/video/58123ff782197e6f3e/ace-no-need-for-at

commy2:
Thanks

Retch44:
1: https://gyazo.com/008756b0c3017bb64280d836438b5db6
[...]

PR #4604, opened 2016-10-27
cookoff ammo selections default values

commy2:
this should be closer to vanilla behaviour and should solve issues with unconfigured vehicles taking too many hits


PabstMirror:
Add missing subTurret cache for Wheeled APCs, will fix #4602

commy2:
lgtm. Can't approve my own PR.

PR #4605, opened 2016-10-27
Disable collision while carrying, Fix floating on disabled simulation objects

jonpas:
- Fix carrying direction for `Land_WoodenBox_F`.
- ~~Fix objects with disabled simulation floating on release (`ace_common_fnc_fixPosition` improvement)~~
- Disable collision with nearby players while carrying.

Reason: I got crushed to death yesterday on a mission by a suitcase on one side and a phone on the other. 😆 

~~Needs testing in multiplayer environment, `disableCollisionWith` has documented weird requirements about locality, no idea how much of it is true as of now.~~

Changed implementation to this to the `setMass 0` method mentioned by Bär.


addons/dragging/functions/fnc_carryObjectPFH.sqf
```diff
@@ -35,10 +36,34 @@ if !(_unit getVariable [QGVAR(isCarrying), false]) exitWith {
...
+// Disable collision with nearby players
+TRACE_1("Disable collision objects",_disabledCollisionObjects);
+private _nearUnits = _target nearObjects ["CAManBase", DISABLE_COLLISION_DISTANCE];
+{
+    if !(_x in _disabledCollisionObjects) then {
+        TRACE_2("Adding disable collision object",_x,typeOf _x);
+        _target disableCollisionWith _x;
+        _disabledCollisionObjects pushBack _x;
+    };
+} count _nearUnits;
```
commy2: Error: Number expected Bool

commy2:
Not a fan of each frame `nearestObjects`, but it's not like you're carrying stuff all the time.

commy2:
The array `_disabledCollisionObjects` should be stored on the object and not on the PFH. That way `dropObject_carry` can be called from everywhere.
I also don't see this ever being executed on the remote machines. Does it even work like this in MP?
Why is it not done for dragging?

commy2:
I suggest making an independant framework for disabling collisions of an object with everything that works in MP and handles JIP.

addons/dragging/functions/fnc_carryObjectPFH.sqf
jonpas: Tested it, worked fine, hmmm...

jonpas:
> The array _disabledCollisionObjects should be stored on the object and not on the PFH. That way dropObject_carry can be called from everywhere.

Good point.

> I also don't see this ever being executed on the remote machines. Does it even work like this in MP?
Why is it not done for dragging?

Dragging doesn't cause collisions like that, I am pretty sure it's because I was testing with a small object, should try something bigger to make sure.

> I suggest making an independant framework for disabling collisions of an object with everything that works in MP and handles JIP.

👍

jonpas:
Dragging doesn't seem to be a problem at all (unless you release the object inside someone).

I still need to test how locality works on that command.

jonpas:
Will work on this tomorrow, need to use target event for `disableCollisionWith` and rerun it on each `Local` event to maintain disabled collision.

PabstMirror:
>Command has to be executed where objects are local, and as long as they don't change locality, the effect of this command will be global.
>If the 2 objects are not local to the same computer, then it has to be executed on both computers to achieve the desired effect.
>If one or both objects change locality, the command needs to be executed again on the new owner's machine(s) to maintain the effect.

https://community.bistudio.com/wiki/disableCollisionWith

Makes it sound like will need to be run globably

jonpas:
Yeah, globally and rerun on each locality change of either object. It seems like I won't have much time to work on this, setting to 3.12, can change back if I do it this weekend.

commy2:
setMass

commy2:
This works in local hosted MP. Client can't damage another client or the host. Mass has be set to >0 however, but that seems to be enough.

commy2:
Tested, works in MP.

PR #4612, opened 2016-10-29
ACE_Chemlight_Shield not public

ColdEvul:
Changed `ACE_Chemlight_Shield` to `public` from `protected`


jonpas:
Tagging @voiperr

commy2:
Same for `ACE_Chemlight_Shield_Green` which all the colored ones inherit from.

ColdEvul:
That I thought was intentional.

commy2:
Why though?

ColdEvul:
Cause I thought the `in use` one was supposed to be protected while the `in used` one was not to be protected.  To avoid people grabbing a ready one from the arsenal (or similar systems).

commy2:
So `ACE_Chemlight_Shield` is basically the empty one? I had to google what a chemlight shield is.

ColdEvul:
Hehe, yes the `ACE_Chemlight_Shield` is the empty one.

Glowbal:
Added to 3.8.2 branch

PR #4613, opened 2016-10-29
Adds realistic names for the QBZ-95 camo variants

ulteq:
Green Hex and Hex variants of:
- QBZ-95-1
- QBZ-95-1 GL
- QBZ-95-1 LSW


jonpas:
This could easily go into 3.8.2 (if we release it) in my opinion.

PR #4614, opened 2016-10-29
Update Macros

654wak654:
- Use CBA 3.1.2's RETDEF() instead of GETMVAR(), leave GETMVAR() in for bwc (Plus someone might require the actual mission namespace)
- Shorten/Encapsulate/Optimize ARR_SELECT()
- ~The zeus component has lots of GETMVAR()'s to replace, I left that for a zeus cleanup which is waiting for #4597~ Done with 72e65ad


PabstMirror:
We need to keep in mind these macro's aren't equivalent in some odd cases.
```
x5 = 5;
with uiNamespace do {
    systemChat format ["Old GETGVAR: %1", missionNameSpace getVariable ["x5", 6]]; // 5
    systemChat format ["New RETDEF: %1", [x5] param [0, 6]]; // 6
};
```

Also the old macro seems to be slightly faster on 20x calls of each (but doesn't matter):
```
// Var doesn't exist:
x1 = missionNamespace getVariable ["ace_plaxxx", objNull]; // 0.0321 ms
x1 =([ace_plaxxx] param [0,objNull]); // 0.0329 ms
// Var does exists
x1 = missionNamespace getVariable ["ace_player", objNull]; // 0.0328 ms
x1 =([ace_player] param [0,objNull]); // 0.0337 ms
```

654wak654:
The usual `GETXVAR` macros can still be used for those cases. Every current use of `GETMVAR` is using it in missionNamespace to get the same effect with RETDEF.

Maybe switching from `ARRAY param` to `ANY param` (so removing the brackets) can close the speed gap.

654wak654:
The macro is in the current cba release btw, didn't wait for 3.2

PabstMirror:
`RETDEF` was changed with https://github.com/CBATeam/CBA_A3/pull/576 - and is now even slower :frowning_face: 

10x calls of each

GETMVAR: 0.0323 ms
```
x4 = missionNamespace getVariable ["test_x_isDefined", 6];
x5 = missionNamespace getVariable ["test_x_unDefined", 7];
```

RETDEF: 0.0633 ms
```
x4 = (if (isNil {test_x_isDefined}) then [{6}, {test_x_isDefined}]);
x5 = (if (isNil {test_x_unDefined}) then [{7}, {test_x_unDefined}]);
```

Could cherry pick the `ARR_SELECT` change, but otherwise I think this could be closed.

654wak654:
So when are you porting `GETXVAR`s to CBA :P

`ARR_SELECT` could just be deprecated/ removed IMO. No need to have a `param` wrapper that isn't even used internally.

commy2:
```sqf
missionNamespace getVariable
```
cannot be used with local variables.

commy2:
>So when are you porting GETXVARs to CBA :P

Not a huge fan of those or "general purpose macros" like that at a whole. I'd rather have all of it written out. This might mainly be because of the lack of variadic macros for the SQF preprocessor.

PR #4615, opened 2016-10-29
Fix typo in Eden Options

SchwererKonigstiger:

This fixes a typo in line 25, `ace_isMeidc_temp` to `ace_isMedic_temp`

### DIFF
--- a/addons/medical/CfgEden.hpp
+++ b/addons/medical/CfgEden.hpp
@@ -22,7 +22,7 @@ class Cfg3DEN {
                     rows = 1;
                     columns = 4;
                     strings[] = {"$STR_3DEN_Attributes_Lock_Default_text", CSTRING(AssignMedicRoles_role_none), CSTRING(AssignMedicRoles_role_medic), CSTRING(AssignMedicRoles_role_doctorShort)};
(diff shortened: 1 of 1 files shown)


addons/medical/CfgEden.hpp
```diff
@@ -22,7 +22,7 @@ class Cfg3DEN {
                     rows = 1;
                     columns = 4;
                     strings[] = {"$STR_3DEN_Attributes_Lock_Default_text", CSTRING(AssignMedicRoles_role_none), CSTRING(AssignMedicRoles_role_medic), CSTRING(AssignMedicRoles_role_doctorShort)};
-                    onToolboxSelChanged = "missionnamespace setvariable ['ace_isMeidc_temp',_this select 1];";
+                    onToolboxSelChanged = "missionnamespace setvariable ['ace_isMedic_temp',_this select 1];";
```
jonpas: Used on L13 as well.

SchwererKonigstiger:
Tested locally - This fixes assigning a medic via Eden Properties dialog
Unsure if there is an open issue for this, but if so, it should be closed by this PR.

PabstMirror:
It's a good fix, 
but it shouldn't change any behaviour because the spelling mistake is the same in both places it's used.

PR #4619, opened 2016-10-30
Allow checking dogtags inside vehicles and when sitting

jonpas:
- Title
- Remove redundant `format`

I also think we should display the name in the interaction menu, currently it's only `Dog Tag, Dog Tag, Dog Tag, ...` when checking dog tags you have taken. Doesn't seem to be as easy as it is communicated from the server rather than kept on client that has it. Reverting #4558 might be preferable in that case, simply to distinguish them (maybe that was the original idea anyways?).

### DIFF
--- a/addons/dogtags/CfgVehicles.hpp
+++ b/addons/dogtags/CfgVehicles.hpp
@@ -34,6 +34,7 @@ class CfgVehicles {
                     displayName = CSTRING(checkItem);
                     condition = "true";
                     statement = "";
+                    exceptions[] = {"isNotInside", "isNotSitting"};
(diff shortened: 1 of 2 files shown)


commy2:
jonpas, the numbers where removed because they broke TvT. You could tell a minimum number of casualties by these numbers which is very bad design gameplay wise.

jonpas:
Ah, makes sense. We should get the name displayed at least then, not sure what's the best way to go about this. Should be done in separate PR anyways, merging this.

PabstMirror:
When removing the index number, I was thinking of adding the tag's names or ID number
But the data isn't available to the client until they ask for it from the server
It's doable, but it will take a little work

PR #4626, opened 2016-11-02
Allows disabling advanced ballistics after mission start

ulteq:


ulteq:
I concluded that we don't really need this.

PR #4629, opened 2016-11-02
Prevent switching to weapon while carrying fuel nozzle

### LINKED ISSUE #4601: Able to switch to weapon while carrying fuel nozzle
**CBA Version:** `3.1.1` stable
**ACE3 Version:** `3.8.1` stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
When carrying a fuel nozzle (from ACE Refuel) you are able to switch to any weapon, with the nozzle still attached to your hand.

**Steps to reproduce:**
- Place a refuel truck
- Pick up its fuel nozzle
- Switch to your primary (through action menu or keybind)

- Editor

**Placed Modules:**
- None

jonpas:
- Fix #4601 

Refuel should use a `canInteractWith` (like carrying does), but that will need a bigger rework. Because of it not using `canInteractWith` Advanced Throwing is possible while carrying a fuel nozzle or jerry can.

Aditionally vanilla throwing is possible both with refuel and carrying, a `setAmmo` framework for that would be nice, but have no time to work on it at the moment.

### DIFF
--- a/addons/refuel/XEH_PREP.hpp
+++ b/addons/refuel/XEH_PREP.hpp
@@ -14,6 +14,7 @@ PREP(dropNozzle);
 PREP(getFuel);
 PREP(handleDisconnect);
 PREP(handleKilled);
+PREP(handlePlayerWeaponChanged);
 PREP(handleUnconscious);
 PREP(makeJerryCan);
 PREP(maxDistanceDropNozzle);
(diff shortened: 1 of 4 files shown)


commy2:
lgtm, but why is `FUNC(statusEffect_set)` not part of `FUNC(dropNozzle)` but called afterwards?

jonpas:
I don't know, it's used like that in other refuel places.

PabstMirror:
@jonpas - I just flipped the PFEH exitWith condition if you want to double check, otherwise looks good to go.

jonpas:
Nice catch @PabstMirror !

Glowbal:
Added to 3.8.2 branch

PR #4631, opened 2016-11-02
Add function command-line tool

jameslkingsley:
- Add a Python command-line tool for functions

It's very basic at the moment and there are probably lots of ways to improve it, but I'm no Python guru. This tool makes it a lot easier and quicker to refactor since it will update the rest of the codebase with changes you make to function names and their addon.

-----------------------

### Options
**Create**
`fnc.py -c addon name` will create `fnc_name.sqf` in `addon\functions` with the function header and script_component included. It will also add `PREP(name);` to the PREP file and lastly open the file in the given editor (this may have to be changed to the default program for SQF files).

**Delete**
`fnc.py -d addon name` will delete `fnc_name.sqf` from `addon\functions` and also delete its PREP from the PREP file.

**Rename**
`fnc.py -r addon name newName` will rename `fnc_name.sqf` to `fnc_newName.sqf` in `addon\functions`. It will also rename the corresponding PREP. Lastly it will search through the rest of the codebase in `addons\` and update its name wherever used (`EFUNC` and `FUNC`).

**Move**
`fnc.py -m addon name newAddon` will move `fnc_name.sqf` in `addon\functions` to `newAddon\functions`, update the PREP files and update the rest of the codebase. It's almost identical to `-r`.

### DIFF
--- a/tools/fnc.py
+++ b/tools/fnc.py
@@ -149,19 +149,19 @@ def renameFunction(addon, name, newName):
 def main():
     args = sys.argv
 
-    if "-cf" in args:
+    if "-c" in args:
         createFunction(args[2], args[3])
         sys.exit(2)
 
-    if "-df" in args:
+    if "-d" in args:
         deleteFunction(args[2], args[3])
         sys.exit(2)
 
-    if "-rf" in args:
+    if "-r" in args:
         renameFunction(args[2], args[3], args[4])
         sys.exit(2)
 
-    if "-mf" in args:
+    if "-m" in args:
         moveFunction(args[2], args[3], args[4])
         sys.exit(2)
 
--- a/tools/fnc.py
+++ b/tools/fnc.py
@@ -1,0 +1,169 @@
+import os
+import sys
+import re
+import glob
+import subprocess
+import textwrap
+
+PREFIX = 'ace'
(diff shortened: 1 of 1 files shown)


esteldunedain:
Maybe take a look a this module @jameslkingsley. 

https://github.com/docopt/docopt

I used it once before and it's simply fantastic for stuff like this PR.

jonpas:
`argparse` is also good.

jonpas:
The only thing I don't like is hard-coded `EDITOR`, I think it would be better to read that from OS config and provide an argument to set the path, that way you could make a hotkey script in Sublime with one line for example.

jameslkingsley:
I kinda forgot about this PR and ended up deleting my forked repo, so now I'm not sure how to push more changes. A proper argument parser would be good, especially to support more options when creating functions and specialized files. Something along the lines of what Laravel's "artisan" CLI tool can do with creating stub files - basically just speeding up the workflow.

jameslkingsley:
I also made another option for creating a new addon using the blank template: https://github.com/ARCOMM/ARCORE/blob/master/tools/fnc.py#L150-L163

PabstMirror:
possibly pick this up for hemit

PR #4636, opened 2016-11-05
Add weapon specific data, add 20 round Lapua and Creedmoor mags for QBU-88

KaiSforza:
Adds weapon specific data for the MXM, QBU-88, and Noreen Bad News rifles. The muzzle velocities are the only real changes, but are important changes for the marksman rifles.

Changed the initSpeed for the 6.5mm Cased ammo that the QBU-88 uses from 820 down to 780, more in-line with the 6.5 caseless that the MX uses.

The velocity of the QBU has also been changed. The barrel on the QBU-88 is quite long (25in, as stated in the config), with the base speed for a 6.5mm round set for a 16in barrel (MX SW). Based on some
quick research on the [differences in muzzle velocity of 6.5mm Creedmoor shot from different barrels](http://rifleshooter.com/2016/02/6-5-creedmoor-effect-of-barrel-length-on-velocity-cutting-up-a-creedmoor/), I came up with the ~1.07 number in this change, as opposed to the reduced velocity of ~0.98
previously.

I have also added data for the 6.5mm Lapua and Creedmoor rounds to the ATragMX table for the QBU that are currently not added.


KaiSforza:
I've contemplated also changing what the zeroing is for those, but I'm not sure how I'd go about calculating that. I don't know if there's a way to bore-sight the scopes in Arma, but that would probably be the easiest way to test and figure out at least a base-line zero for 100m on most rifles.

addons/ballistics/CfgMagazines.hpp
```diff
@@ -8,6 +8,9 @@ class CfgMagazines {
     class 100Rnd_580x42_Mag_F: 30Rnd_580x42_Mag_F {
         initSpeed = 930;
     };
+    class 20Rnd_650x39_Cased_Mag_F: CA_Magazine {
+        initSpeed = 780;
```
ulteq: I think this should be 760.

addons/ballistics/CfgWeapons.hpp
```diff
@@ -28,9 +28,14 @@ class CfgWeapons {
     };
 
     class DMR_07_base_F: Rifle_Long_Base_F {
-        initSpeed = -0.982908;
+        initSpeed = -1.069461;
```
ulteq: This does not have to be changed if the magazine `initSpeed` is correct.

ulteq:
> I've contemplated also changing what the zeroing is for those

Are you talking about the iron sights of the new Apex weapons? Or about the new scopes that were introduced with Apex?

ulteq:
> Adds weapon specific data for the MXM, QBU-88, and Noreen Bad News rifles. The muzzle velocities are the only real changes, but are important changes for the marksman rifles.

I don't think that we should add weapon specific default gun profiles.

addons/ballistics/CfgMagazines.hpp
KaiSforza: I set it to 780, since it's the 6.5 cased ammo, which seemed to have higher velicities than the caseless ammo in the vanilla game. Not sure if this makes a lot of sense to do in Arma, but based on how caseless ammo works in the real world, cased ammo probably retains more of the energy as it's a lot harder for the gasses to escape from the action.

KaiSforza:
> You are right! It needs to be 1.06051.

Just curious, where did the 1.06051 come from?

> Are you talking about the iron sights of the new Apex weapons? Or about the new scopes that were introduced with Apex?

Changing the range zero in the atragmx computer to what the actual zero is for the guns when set to "100m."

KaiSforza:
Closing this in favor of #4641.

ulteq:
> Changing the range zero in the atragmx computer to what the actual zero is for the guns when set to "100m."

An ideal solution would be when Bohemia Interactive fixes their vanilla zero.

PR #4638, opened 2016-11-05
Fix ammo cook off with certain explosives

SilentSpike:
- Add a toggle to ammo boxes in 3den to disable their cook off individually
- Fix some explosives not working as part of ammo detonation
- Include shots of simulation type `shotsubmunitions` in ammo detonation

### DIFF
--- a/addons/cookoff/CfgEden.hpp
+++ b/addons/cookoff/CfgEden.hpp
@@ -11,7 +11,7 @@ class Cfg3DEN {
                         tooltip = CSTRING(enable_tooltip);
                         expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
                         typeName = "BOOL";
-                        condition = "objectVehicle";
+                        condition = "objectVehicle + objectHasInventoryCargo";
                         defaultValue = "(true)"; // fix pbo project preprocessing bug
                     };
                 };
--- a/addons/cookoff/CfgEden.hpp
+++ b/addons/cookoff/CfgEden.hpp
@@ -11,7 +11,7 @@ class Cfg3DEN {
                         tooltip = CSTRING(enable_tooltip);
                         expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
                         typeName = "BOOL";
-                        condition = "objectVehicle + objectHasInventoryCargo";
+                        condition = "objectHasInventoryCargo";
                         defaultValue = "(true)"; // fix pbo project preprocessing bug
                     };
                 };
--- a/addons/cookoff/functions/fnc_detonateAmmunition.sqf
+++ b/addons/cookoff/functions/fnc_detonateAmmunition.sqf
@@ -37,7 +37,7 @@ if (_amountOfMagazines > 0) exitWith {
     _timeBetweenAmmoDetonation = _timeBetweenAmmoDetonation max 0.1;
 
     private _ammo = getText (configFile >> "CfgMagazines" >> _magazineClassname >> "ammo");
-    private _ammoCfg = (configFile >> "CfgAmmo" >> _ammo);
+    private _ammoCfg = configFile >> "CfgAmmo" >> _ammo;
 
     private _speedOfAmmo = getNumber (configFile >> "CfgMagazines" >> _magazineClassname >> "initSpeed");
     private _simulationTime = getNumber (_ammoCfg >> "simulation");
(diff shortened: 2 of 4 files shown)


SilentSpike:
Something to consider:

The value `ace_explosives_Explosive` being retrieved to determine the script-compatible version of explosives could be either moved to a shared ace dependency (i.e. common) or added as an additional value into ace_cookoff (removing any implicit dependency on ace_explosives)

addons/cookoff/CfgEden.hpp
```diff
@@ -11,7 +11,7 @@ class Cfg3DEN {
                         tooltip = CSTRING(enable_tooltip);
                         expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
                         typeName = "BOOL";
-                        condition = "objectVehicle";
+                        condition = "objectVehicle + objectHasInventoryCargo";
```
commy2: uhm. what would be `objectHasInventoryCargo`, but not `objectVehicle` ? Doesn't the latter require the former?

addons/cookoff/functions/fnc_detonateAmmunition.sqf
```diff
@@ -101,9 +102,25 @@ if (_amountOfMagazines > 0) exitWith {
...
+            private _scripted = getNumber (_ammoCfg >> "triggerWhenDestroyed") == 1;
+            if !(_scripted) then {
+                _ammo = getText (_ammoCfg >> "ace_explosives_Explosive");
+            };
+
+            // If a scripted alternative doesn't exist use generic explosion
+            if (_ammo != "") then {
+                [_vehicle, _ammo, 0, false] call _spawnProjectile;
+            } else {
+                "SmallSecondary" createvehicle (_vehicle modelToWorld _effect2pos);
```
commy2: `createVehicle`. I think we also have a rule to alwas use the unary version of this command (`createVehicle ARRAY`)

addons/cookoff/functions/fnc_detonateAmmunition.sqf
```diff
@@ -33,15 +33,16 @@ if (_amountOfMagazines > 0) exitWith {
     } else {
         _magazine set [1, _newMagCount]; // clear out the magazine
     };
-    private _ammo = getText (configFile >> "CfgMagazines" >> _magazineClassname >> "ammo");
-
     private _timeBetweenAmmoDetonation = (random 7) * (1 / random (_amountOfMagazines)) min MAX_TIME_BETWEEN_AMMO_DET;
     _timeBetweenAmmoDetonation = _timeBetweenAmmoDetonation max 0.1;
 
+    private _ammo = getText (configFile >> "CfgMagazines" >> _magazineClassname >> "ammo");
+    private _ammoCfg = (configFile >> "CfgAmmo" >> _ammo);
```
commy2: superfluous parenthesis

addons/cookoff/functions/fnc_detonateAmmunition.sqf
SilentSpike: Fair, was just following the existing uses of `createVehicle` in the cook off code. Will change them all while I'm at it.

addons/cookoff/CfgEden.hpp
SilentSpike: Yeah I wasn't quite sure if you can get vehicles without inventory space and figured it was safer to just encompass all this way.

commy2:
I'm not sure the CAN_COLLIDE flag is the right one here. That would make the explosion spawn on the ground. I think either FLY is needed here or the good old setPosASL.

SilentSpike:
Are you sure `createVehicle` spawns things on the ground?

If anything (going by the biki comments) the old syntax from before I changed it would spawn things on the ground.

commy2:
I have no idea. I always used FLY for spawning projectiles

commy2:
do whatever works

SilentSpike:
Pretty sure `"FLY"` is only relevant to air units such that they'll spawn at their appropriate altitude with their engine on and in formation around their group leader. Should have no effect on projectiles and the likes.

I'll double check that the current code in this PR doesn't spawn things on the ground before merging, but I'm pretty confident about it :+1:

SilentSpike:
Unary `createVehicle` confirmed working for objects in the air (as does the other version - so the biki is entirely wrong). Merging!

Glowbal:
Added to 3.8.2 branch

PR #4640, opened 2016-11-05
Fix make.py checkexternal argument documentation

jonpas:
- Title

@Glowbal I am sorry, but you were properly using the wrong one. :/


jonpas:
Yeah, I messed that up in ACRE2's version, was wrong in ACRE2 wiki as well.

PR #4641, opened 2016-11-06
Ballistics - Fixed 6.5mm magazine config

ulteq:
* Add new 20rnd 6.5mm Creedmor and 6.5mm Lapua Scenar magazines
* Fix incorrect QBU-88 (DMR_07_base_F) `initSpeed`
* Fix 30rnd 6.5mm  Creedmor and 30rnd 6.5mm Lapua Scenar magazine `initSpeeds`

Closes: #4636

### DIFF
--- a/addons/ballistics/CfgMagazines.hpp
+++ b/addons/ballistics/CfgMagazines.hpp
@@ -241,12 +241,26 @@ class CfgMagazines {
         displayNameShort = CSTRING(30Rnd_65x47_Scenar_mag_NameShort);
         descriptionShort = CSTRING(30Rnd_65x47_Scenar_mag_Description);
     };
+    class ACE_20Rnd_65x47_Scenar_mag: 20Rnd_650x39_Cased_Mag_F {
+        author = ECSTRING(common,ACETeam);
+        ammo = "ACE_65x47_Ball_Scenar";
+        displayName = CSTRING(20Rnd_65x47_Scenar_mag_Name);
+        displayNameShort = CSTRING(20Rnd_65x47_Scenar_mag_NameShort);
+        descriptionShort = CSTRING(20Rnd_65x47_Scenar_mag_Description);
+    };
     class ACE_30Rnd_65_Creedmor_mag: 30Rnd_65x39_caseless_mag {
         author = ECSTRING(common,ACETeam);
         ammo = "ACE_65_Creedmor_Ball";
         displayName = CSTRING(30Rnd_65_Creedmor_mag_Name);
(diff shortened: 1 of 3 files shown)


KaiSforza:
I just have to wonder why the caseless ammo is a higher velocity than the cased ammo. Shouldn't it be the opposite?

addons/ballistics/CfgMagazines.hpp
```diff
@@ -237,17 +240,35 @@ class CfgMagazines {
...
         ammo = "ACE_65_Creedmor_Ball";
+        initSpeed = 815;
         displayName = CSTRING(30Rnd_65_Creedmor_mag_Name);
         displayNameShort = CSTRING(30Rnd_65_Creedmor_mag_NameShort);
         descriptionShort = CSTRING(30Rnd_65_Creedmor_mag_Description);
     };
+    class ACE_20Rnd_65_Creedmor_mag: 20Rnd_650x39_Cased_Mag_F {
+        author = ECSTRING(common,ACETeam);
+        ammo = "ACE_65_Creedmor_Ball";
+        initSpeed = 808;
```
KaiSforza: Did you mean for this to be 815, and the caseless to be 808, as you did for the Scenar?

KaiSforza: Also, just looking at the data from that test that was done it seems like the base speed (if we're considering a 16" barrel the base speed for the 6.5mm cartriges) should be around 760 for the creedmoor and lapua ammo.

ulteq:
The muzzle velocity vs. barrel length mapping happens here:
https://github.com/KaiSforza/ACE3/blob/6fddb787e3d05610f206e0c19c7330f3f866e521/addons/ballistics/CfgAmmo.hpp#L188-L189

This is our current config for 6.5 mm Creedmor:
```
        ACE_muzzleVelocities[]={750, 820, 840, 852, 860};
        ACE_barrelLengths[]={254.0, 406.4, 508.0, 609.6, 660.4};
```

We currently do not simulate caseless ammo. We assume that all vanilla 6.5mm rounds are cased.

> I just have to wonder why the caseless ammo is a higher velocity than the cased ammo.
The magazine `initSpeed` is multiplied by the weapon `initSpeed`.

**MXM:** `initSpeed: 1.01842` **30Rnd_65_Creedmor:** `initSpeed: 815` -> Muzzle velocity: 830
**DMR_07:** `initSpeed: 1.06946` **20Rnd_65_Creedmor:`** `initSpeed: 808` -> Muzzle velocity: 864

The net muzzle velocity is higher even though the magazine `initSpeed` is lower.

PR #4642, opened 2016-11-07
Scopes - new features, vanilla zeroing fix

ulteq:
Makes the scopes module fully compatible with any 3rd party weapon pack without the need for additional compat pbos (though their use is still recommended).

It works with both vanilla- and advanced ballistics.
It is more accurate than vanilla zeroing, because bore height is taken into account.

It allows using adjustment turrets on weapons with integrated scopes.

Fixes scope adjustments affecting under-barrel grenade launchers.

**TODO** 
- [x] Read all required data from config
- [x] Add `RailHeightAboveBore` and `SightHeightAboveRail` config entries
- [x] Calculate current BoreHeight from `RailHeightAboveBore` and `SightHeightAboveRail`
- [x] Cache the zero angle correction
- [x] Find a sweet spot for the default bore height (probably somewhere around 5 cm)
- [x] Make scope adjustment optional (~~cherry-pick this: https://github.com/acemod/ACE3/pull/3561/commits/18767cd127f81384f2ae3091f626c8d004ac6bfa ?~~)
- [x] Move bore height calculation out of the fired event handler
- [x] Move the zero angle correction calculation into the ace_advanced_ballistics.dll

Related with: https://github.com/acemod/ACE3/pull/3561


KaiSforza:
I've tested this out and the results are amazing. I can finally get rid of that file where i stored all of the default 100m range offsets for every gun. :+1:

PabstMirror:
- Adding and then removing scope throws warning and vanilla ranging doesn't return.
```
[ACE] (ui) WARNING: Element 'zeroing' already set by 'ace_scopes' - modification by 'ace_scopes' failed. z\ace\addons\ui\functions\fnc_setElementVisibility.sqf:56
```
- Adding scope to mxGL removes zeroing for GL muzzle.
- Setting `GVAR(deduceBarometricPressureFromTerrainAltitude)` only seems to have effect in the module code, but could be set from server/mission config. Probably move to `ace_settingsInitialized` EH.

ulteq:
> Adding and then removing scope throws warning and vanilla ranging doesn't return.

I think this warning is caused by the call to EFUNC(ui,setElementVisibility):
https://github.com/acemod/ACE3/blob/9885add18a5892886c43fac9368031c777c9f3b4/addons/scopes/functions/fnc_inventoryCheck.sqf#L72-L73

addons/scopes/functions/fnc_inventoryCheck.sqf
```diff
@@ -103,11 +69,71 @@ private _newGuns = [primaryWeapon _player, secondaryWeapon _player, handgunWeapo
         (GVAR(scopeAdjust) select _forEachIndex) set [3, _horizontalIncrement];
         GVAR(canAdjustElevation) set [_forEachIndex, (_verticalIncrement > 0) && !(_maxVertical isEqualTo [0, 0])];
         GVAR(canAdjustWindage) set [_forEachIndex, (_horizontalIncrement > 0) && !(_maxHorizontal isEqualTo [0, 0])];
-        private _hideVanillaZeroing = (getNumber(_opticConfig >> "ItemInfo" >> "opticType") == 2 && {GVAR(canAdjustElevation) select _forEachIndex}) || {isNumber (_opticConfig >> "ACE_ScopeZeroRange")};
+        private _hideVanillaZeroing = (GVAR(enabled) && getNumber(_opticConfig >> "ItemInfo" >> "opticType") == 2 && {GVAR(canAdjustElevation) select _forEachIndex}) || {isNumber (_opticConfig >> "ACE_ScopeZeroRange")};
```
ulteq: This should have been `GVAR(forceUseOfAdjustmentTurrets)`.

ulteq:
@PabstMirror Those issues should be fixed now.

addons/scopes/functions/fnc_inventoryCheck.sqf
```diff
@@ -17,28 +17,119 @@
...
 private _newOptics = [_player] call FUNC(getOptics);
+private _newGuns = [primaryWeapon _player, secondaryWeapon _player, handgunWeapon _player];
 
 {
     if (_newOptics select _forEachIndex != _x) then {
         // The optic for this weapon changed, set adjustment to zero
         if (!((_adjustment select _forEachIndex) isEqualTo [0, 0, 0])) then {
             _adjustment set [_forEachIndex, [0, 0, 0]];
-            [ACE_player, QGVAR(Adjustment), _adjustment, 0.5] call EFUNC(common,setVariablePublic);
+            _updateAdjustment;
```
PabstMirror: `_updateAdjustment = true;`

addons/scopes/functions/fnc_inventoryCheck.sqf
```diff
@@ -17,28 +17,119 @@
...
-if (isNil QGVAR(Optics)) then {
-    GVAR(Optics) = ["", "", ""];
-};
 private _newOptics = [_player] call FUNC(getOptics);
+private _newGuns = [primaryWeapon _player, secondaryWeapon _player, handgunWeapon _player];
 
 {
     if (_newOptics select _forEachIndex != _x) then {
         // The optic for this weapon changed, set adjustment to zero
         if (!((_adjustment select _forEachIndex) isEqualTo [0, 0, 0])) then {
```
ulteq: > Not sure, but I think we should move this down to line 72

We never write to `_newGuns`, so I don't think that the early definition can cause any trouble.

> because we should reset zeroing on weapon change, if the old weapon was zeroable

Nice catch :+1:. It would probably be easiest if any weapon change would reset the scope adjustments.

esteldunedain:
Fantastic PR @ulteq! Thank you very much.

PR #4648, opened 2016-11-09
Fix bouncing 40mm smoke

BaerMitUmlaut:
- Fix rubber 40mm smoke shells

### DIFF
--- a/addons/ballistics/CfgAmmo.hpp
+++ b/addons/ballistics/CfgAmmo.hpp
@@ -197,7 +197,7 @@ class CfgAmmo {
         typicalSpeed=833;
         hit=9;
         tracerScale = 1.2; //0.6;
-        tracerStartTime=0.073;  // Based on the British L5A1 which burns out to 1000m 
+        tracerStartTime=0.073;  // Based on the British L5A1 which burns out to 1000m
         tracerEndTime=2.15957;  // Time in seconds calculated with ballistics calculator
         ACE_caliber=7.823;
         ACE_bulletLength=28.956;
@@ -834,5 +834,10 @@ class CfgAmmo {
     };
     class Gatling_30mm_HE_Plane_CAS_01_F: BulletBase {
         tracerScale = 2.5;
+    };
+
+    class SmokeShell;
+    class G_40mm_Smoke: SmokeShell {
+        simulation = "shotSmoke";
     };
 };


commy2:
please explain what's going on here.

jonpas:
40mm smokes bounce a lot. Fire it at a hill and watch it bounce up and over. This fixes it.

PabstMirror:
@commy2 - I think I found an explanation here: 
https://forums.bistudio.com/topic/191118-howto-make-ugl-40mm-smoke-grenades-less-bouncy/

BaerMitUmlaut:
`shotSmokeX`, the simulation type for the current 40mm smoke grenades, is inherited from hand thrown smoke grenades, which are supposed to bounce. Because the 40mm ones travel at much higher speed, they bounce excessively when hitting the ground, making you miss your target _every single time_, which renders them useless for target marking - the main purpose of them irl.
This removes the bouncyness. I copied the `shotSmoke` simulation type from the artillery smoke shell.

commy2:
So this PR changes shotSmokeX to shotSmoke to remove physx bounces of UGL smoke grenades?
Are there any other consequences changing the simulation has? Are all sounds and effects still working? Or do we have to redefine some entries?
shotSmoke seems to be usable if artillery shells and mortar rounds use it

PabstMirror:
the round that comes out of arty/mortars is a a `simulation = "shotDeploy";`
which becomes the shotSmoke ammo on impact
doing it that way should guarantee no bounce.

We should double check how shotSmoke performs with air resistance as I believe it is disabled on some types. Making sure HE and Smoke rounds fly with similar flight paths in wind.

commy2:
Yeah. And I somehow don't think we should totally get rid of the bouncing. Why not modify `deflecting` or even add a custom `deflectionSlowDown` (if that works)? 

`deflecting = 0;` should disable it without changing the simulation.

BaerMitUmlaut:
A 40mm smoke shouldn't bounce unless it hits concrete or something it can't penetrate at all. In Arma that's rather the exception than the norm though.

>`deflecting = 0;` should disable it without changing the simulation.

You're right, it should. Doesn't though.

BaerMitUmlaut:
Bad news: sound doesnt work. Good point commy. I'll try tweaking the X one again then...

Neviothr:
I used this to disable the bouncing:
```class CfgAmmo {
    class SmokeShell;
    class G_40mm_Smoke : SmokeShell {
        simulation = "shotSmoke";
        deflectionSlowDown = 0.3;
    };
};```

BaerMitUmlaut:
Closing because `shotSmokeX` is simply not fixable on our side.

esteldunedain:
😢 It'd have been awesome

PR #4651, opened 2016-11-10
ATragMX - Estimate bullet length instead of using a fixed value

ulteq:
* Fix the ATragMX spin drift calculation


Glowbal:
Added to 3.8.2 branch

Issue #4424, opened 2016-09-18
Advanced fatigue - bar is not hiding when getting into vehicles.


vlad8011:
**Arma 3 Version:** `3.64` RC
**CBA Version:** `v3.0.0.160713` (stable)
**ACE3 Version:** `3.7.0` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`
- `@RHS`

**Description:**
- After getting into vehicle, stamina bar keeps uncovered, and in wrong place - HUD element with weapon info is moving into the right when getting in vehicle, beacuse its covering state indicator (useless when inside car), but stamina bar (from ace) is staying in same place, where it was before we went into car.

**Steps to reproduce:**
- Turn ON advanced fatigue, spawn yourself next to some vehicle, observe stamina bar, while getting into vehicle. 

**Where did the issue occur?**
- Everywhere

**Placed Modules:**
- Advanced Fatigue

BaerMitUmlaut:
How does vanilla behave compared to this?

commy2:
the bar goes away since the whole control is replaced with the one from the vehicle. that does only happen for driver and gunner slots though. cargo still uses the infantry weapon info box.

I guess it can get misalligned if you set different positions for these two controls in the ui editor.

PR #4655, opened 2016-11-10
Hide advFatigue bar on map / spectator

PabstMirror:


BaerMitUmlaut:
Should do vehicles as well.

PabstMirror:
Added, will close #4424

addons/advanced_fatigue/XEH_postInit.sqf
```diff
@@ -12,10 +12,11 @@ if (!hasInterface) exitWith {};
     GVAR(ppeBlackout) ppEffectCommit 0.4;
 
     // - GVAR updating and initialization -----------------------------------------
-    if !(isNull ACE_player) then {
-        [ACE_player, objNull] call FUNC(handlePlayerChanged);
-    };
-    ["unit", FUNC(handlePlayerChanged)] call CBA_fnc_addPlayerEventHandler;
+    ["unit", FUNC(handlePlayerChanged), true] call CBA_fnc_addPlayerEventHandler;
```
BaerMitUmlaut: The `handlePlayerChanged` function needs to run at least once. If the unit EH triggered already, this will completly break AF. Would recommend to just revert this to how it was.

addons/advanced_fatigue/functions/fnc_handleStaminaBar.sqf
```diff
@@ -13,6 +13,11 @@ params ["_stamina"];
 
 private _staminaBarContainer = uiNamespace getVariable [QGVAR(staminaBarContainer), controlNull];
 
+if ((vehicle ACE_player) != ACE_player) exitWith { // Hide when mounted
```
BaerMitUmlaut: Why not use the vehicle player EH?

addons/advanced_fatigue/XEH_postInit.sqf
PabstMirror: Third arg for addPlayerEventHandler was recently added to do the same thing:
```
_applyRetroactively - Call function immediately if player is defined already (optional, default: false) <BOOL>
```

addons/advanced_fatigue/functions/fnc_handleStaminaBar.sqf
PabstMirror: I think we're still going to need the if statement here or it will be overwriten

addons/advanced_fatigue/XEH_postInit.sqf
BaerMitUmlaut: Oh, I see. That works then.

addons/advanced_fatigue/functions/fnc_handleStaminaBar.sqf
BaerMitUmlaut: You could do it just like in the visibleMap EH. Both need checks if both conditions apply. I'd rather keep all the hiding/showing in one place.

BaerMitUmlaut:
Thank you @PabstMirror!

PR #4656, opened 2016-11-11
ATragMX - Added 7.62x39mm ammo to gun profile

ulteq:


ulteq:
-> https://github.com/acemod/ACE3/pull/4667

PR #4658, opened 2016-11-11
Add setting to only run adjustments on real scopes

PabstMirror:
For PR #4642
Changes enabed from bool to 
Disabled/Enabled only for adjustable scopes/Enabled for all (default)


addons/scopes/functions/fnc_firedEH.sqf
```diff
@@ -23,6 +23,7 @@ if (_weaponIndex < 0) exitWith {};
 
 private _zeroing = +(_adjustment select _weaponIndex);
 TRACE_1("Adjusting With",_zeroing);
+if (_zeroing isEqualTo []) exitWith {};
```
ulteq: This will break the entire pull request.

ulteq:
I think we also need to change this: https://github.com/acemod/ACE3/blob/b5bb3b407fbf31c2cab0f1f3cc025bce341b2fd4/addons/scopes/CfgVehicles.hpp#L31-L36

addons/scopes/stringtable.xml
```diff
@@ -7,6 +7,9 @@
         <Key ID="STR_ACE_Scopes_enabled_displayName">
             <English>Enable ACE Scope adjustment</English>
         </Key>
+        <Key ID="STR_ACE_Scopes_enabled_onlyScopes">
+            <English>Only for scopes</English>
```
ulteq: I think it should be `Only for adjustable scopes`.

addons/scopes/functions/fnc_inventoryCheck.sqf
```diff
@@ -131,10 +127,34 @@ private _newGuns = [primaryWeapon _player, secondaryWeapon _player, handgunWeapo
...
+            if ((GVAR(canAdjustElevation) select _forEachIndex) || (GVAR(canAdjustWindage) select _forEachIndex)) then {
+                if (!((_adjustment select _forEachIndex) isEqualTo [0,0,0])) then {
+                    // The optic for this weapon changed, set adjustment to zero
+                    _adjustment set [_forEachIndex, [0, 0, 0]];
+                    _updateAdjustment = true;
+                };
+            } else {
+                if (!((_adjustment select _forEachIndex) isEqualTo [])) then {
+                    // Can't change this optic, set to empty set
+                    _adjustment set [_forEachIndex, []];
```
ulteq: I like the idea, but I think it will break the code in other places:
https://github.com/acemod/ACE3/blob/98da86b74d4089bc3403d0b0c0074e6a46ce35d8/addons/scopes/functions/fnc_adjustScope.sqf#L50
https://github.com/acemod/ACE3/blob/98da86b74d4089bc3403d0b0c0074e6a46ce35d8/addons/scopes/functions/fnc_applyScopeAdjustment.sqf#L34

ulteq:
-> https://github.com/acemod/ACE3/pull/4642/commits/adc9171f11f263e3042257e48b165c5778ffdd32

PR #4576, opened 2016-10-20
Add arsenal zeus modules

alganthe:
**When merged this pull request will:**
- Add an add full arsenal curator module
- Add a remove arsenal curator module

PR #4556, opened 2016-10-16
Add toggle flashlight and NVG zeus modules

alganthe:
**When merged this pull request will:**
- Add a module to toggle flashlights on or off for a group or a side.
- Add a module to toggle NVGs on or off for a group or a side.

Both of those are added if they're not already present on the unit's weapon or head, it uses whatever the unit is supposed to have.
It works with headgear that has built in night vision / thermal.
If the units don't have flashlights on their weapons it checks if they have a compatible attachment that is a flashlight and give it to them (if none are available then so be it)

PR #4661, opened 2016-11-12
Add Zeus utility modules

RSpeekenbrink:
- **Add a Zeus Module category ACE Util**
- **Add Zeus Module: Toggle Simulation**
This module needs to be placed on an object. This will toggle the simulation of the object. Handy for base builds etc.
- **Add Zeus Module: Add Objects to Curator**
This module will ask for a radius once placed (radius in meters) and will add all objects to zeus within that radius.
- **Add Zeus Module: Remove Objects from Curator**
This module will do the same as Add but then inverted (Remove all objects in the radius from zeus, it will not remove the objects it will just make them uneditable and no icon will show up etc)


addons/zeus/stringtable.xml
```diff
@@ -644,6 +647,9 @@
             <German>Fügt jedes gespawnte Objekt allen Kuratoren der Mission hinzu</German>
             <Japanese>ミッション内で作成されたオブジェクトに全キュレーターを追加</Japanese>
         </Key>
+		<Key ID="STR_ACE_Zeus_RemoveObjectsToCurator">
```
jonpas: `STR_ACE_Zeus_RemoveObjectsFromCurator`

addons/zeus/stringtable.xml
```diff
@@ -607,6 +607,9 @@
             <Italian>Piazza su una unità</Italian>
             <Japanese>ユニットの上に設置</Japanese>
         </Key>
+		<Key ID="STR_ACE_Zeus_NoObjectSelected">
+            <English>Place on a object</English>
```
jonpas: on *an* object

addons/zeus/functions/fnc_ui_RemoveObjects.sqf
```diff
@@ -0,0 +1,64 @@
+/*
+ * Author: Fisher
+ * Removes all objects in given radius for all curators
```
jonpas: Missing final point.

addons/zeus/functions/fnc_ui_RemoveObjects.sqf
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Arguments:
+ * 0: dummy controls group <CONTROL>
```
jonpas: `Dummy` (capitalization)

addons/zeus/functions/fnc_ui_RemoveObjects.sqf
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Return Value:
+ * none <NIL>
```
jonpas: Just `None`.

addons/zeus/functions/fnc_ui_AddObjects.sqf
```diff
@@ -0,0 +1,63 @@
+/*
+ * Author: Fisher
+ * Adds all objects in given radius to all curators
+ *
+ * Arguments:
+ * 0: dummy controls group <CONTROL>
+ *
+ * Return Value:
+ * none <NIL>
```
jonpas: Just `None`.

addons/zeus/functions/fnc_ui_AddObjects.sqf
```diff
@@ -0,0 +1,63 @@
+/*
+ * Author: Fisher
+ * Adds all objects in given radius to all curators
+ *
+ * Arguments:
+ * 0: dummy controls group <CONTROL>
```
jonpas: `Dummy` (capitalization)

addons/zeus/functions/fnc_ui_AddObjects.sqf
```diff
@@ -0,0 +1,63 @@
+/*
+ * Author: Fisher
+ * Adds all objects in given radius to all curators
```
jonpas: Missing final point.

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,30 @@
+/*
+ * Author: Fisher
+ * Toggle Simulation on object
+ *
+ * Arguments:
+ * 0: The module logic <OBJECT>
+ *
+ * Return Value:
+ * None <NIL>
```
jonpas: Just `None`.

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,30 @@
+/*
+ * Author: Fisher
+ * Toggle Simulation on object
```
jonpas: Missing final point.

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Arguments:
+ * 0: The module logic <OBJECT>
+ *
+ * Return Value:
+ * None <NIL>
+ *
```
jonpas: An empty line too many.

addons/zeus/config.cpp
```diff
@@ -49,6 +52,14 @@ class ACE_Curator {
     GVAR(cargoAndRepair)[] = {"ace_cargo", "ace_repair"};
 };
 
+class CfgFactionClasses {
```
jonpas: Should be in separate file ˙`CfgFactionClasses.hpp` and included below.

addons/zeus/config.cpp
```diff
@@ -49,6 +52,14 @@ class ACE_Curator {
...
+    class ACE_Util {
```
jonpas: `ACE_UI_Util` maybe?

addons/zeus/XEH_PREP.hpp
```diff
@@ -14,17 +14,20 @@ PREP(moduleSearchNearby);
...
 PREP(moduleSetMedicalVehicle);
 PREP(moduleSetMedicalFacility);
+PREP(moduleSimulation);
 PREP(moduleSurrender);
 PREP(moduleTeleportPlayers);
 PREP(moduleUnconscious);
 PREP(moduleZeusSettings);
 PREP(ui_attributeCargo);
 //PREP(ui_attributePosition);
+PREP(ui_AddObjects);
```
jonpas: `ui_addObjects` - same for file.

addons/zeus/XEH_PREP.hpp
```diff
@@ -14,17 +14,20 @@ PREP(moduleSearchNearby);
...
 PREP(ui_attributeRadius);
 PREP(ui_defendArea);
 PREP(ui_globalSetSkill);
 PREP(ui_groupSide);
 PREP(ui_patrolArea);
+PREP(ui_RemoveObjects);
```
jonpas: `ui_removeObjects` - same for file.

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,30 @@
...
+ private ["_object","_simulationEnabled"];
+ 
+ if !(local _logic) exitWith {};
+
+ _object = attachedTo _logic;
+if (isNull _object) then {
+    [LSTRING(NoObjectSelected)] call EFUNC(common,displayTextStructured);
+} else {
+    _simulationEnabled = simulationEnabled _object;
+    _object enableSimulationGlobal (!_simulationEnabled);
```
commy2: doesn't work in MP like this.
https://community.bistudio.com/wiki/enableSimulationGlobal

Has to be executed on server:

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,30 @@
...
+ * None <NIL>
+ *
+ *
+ * Public: No
+ */
+ 
+ #include "script_component.hpp"
+ 
+ params ["_logic"];
+ private ["_object","_simulationEnabled"];
```
commy2: use `private` keyword instead.

addons/zeus/functions/fnc_ui_AddObjects.sqf
```diff
@@ -0,0 +1,63 @@
...
+ * Public: No
+ */
+ 
+#include "script_component.hpp"
+ 
+disableSerialization;
+ 
+params ["_control"];
+private _display = ctrlParent _control;
+private _logic = GETMVAR(BIS_fnc_initCuratorAttributes_target,objnull);
```
commy2: `objNull`

addons/zeus/functions/fnc_ui_AddObjects.sqf
```diff
@@ -0,0 +1,63 @@
...
+private _fnc_errorAndClose = {
+    params ["_msg"];
+    _display closeDisplay 0;
+    deleteVehicle _logic;
+    [_msg] call EFUNC(common,displayTextStructured);
+    breakOut "Main";
+};
+
+private _fnc_onUnload = {
+    private _logic = GETMVAR(BIS_fnc_initCuratorAttributes_target,objnull);
```
commy2: same

addons/zeus/functions/fnc_ui_AddObjects.sqf
```diff
@@ -0,0 +1,63 @@
...
+    if (isNull _logic) exitWith {};
+
+    private _radius = GETVAR(_display,GVAR(radius),50);
+    private _position = GETVAR(_display,GVAR(position),getPos _logic);
+    private _objectsToAdd = nearestObjects [_position, ["All"], _radius];
+
+    {
+        _y = _x;
+        { _x addCuratorEditableObjects [[_y], true]; } foreach allCurators;
+    } foreach _objectsToAdd;
```
commy2: `forEach`.
New scope in new line.
Has to be executed on server and probably doesn't work like this in MP.
missing `private` for `_y`

addons/zeus/functions/fnc_ui_RemoveObjects.sqf
```diff
@@ -0,0 +1,64 @@
...
+private _fnc_errorAndClose = {
+    params ["_msg"];
+    _display closeDisplay 0;
+    deleteVehicle _logic;
+    [_msg] call EFUNC(common,displayTextStructured);
+    breakOut "Main";
+};
+
+private _fnc_onUnload = {
+    private _logic = GETMVAR(BIS_fnc_initCuratorAttributes_target,objnull);
```
commy2: same

addons/zeus/functions/fnc_ui_RemoveObjects.sqf
```diff
@@ -0,0 +1,64 @@
...
+    };
+};
+
+private _fnc_onConfirm = {
+    params [["_ctrlButtonOK", controlNull, [controlNull]]];
+
+    private _display = ctrlparent _ctrlButtonOK;
+    if (isNull _display) exitWith {};
+
+    private _logic = GETMVAR(BIS_fnc_initCuratorAttributes_target,objnull);
```
commy2: same

addons/zeus/functions/fnc_ui_RemoveObjects.sqf
```diff
@@ -0,0 +1,64 @@
...
+    if (isNull _logic) exitWith {};
+
+    private _radius = GETVAR(_display,GVAR(radius),50);
+    private _position = GETVAR(_display,GVAR(position),getPos _logic);
+    private _objectsToAdd = nearestObjects [_position, ["All"], _radius];
+
+    {
+        _y = _x;
+        { _x removeCuratorEditableObjects [[_y], true]; } foreach allCurators;
+    } foreach _objectsToAdd;
```
commy2: same

addons/zeus/functions/fnc_moduleSimulation.sqf
jonpas: There is an event in `common` component for it.

addons/zeus/functions/fnc_ui_AddObjects.sqf
commy2: Why isn't this simply:
```sqf
    {
        _x addCuratorEditableObjects [_objectsToAdd, true];
    } forEach allCurators;
```
`addCuratorEditableObjects` is already meant to be able to add arrays of objects. No point in walking over the array just to pass each element in a size one array.

addons/zeus/functions/fnc_moduleSimulation.sqf
RSpeekenbrink: Fixed it with `BIS_fnc_MP` now, I couldn't find the event you where talking about

SilentSpike:
Like the addition of module categories, have been meaning to get around to that myself :+1:

UI stuff all looks to be handled correctly in line with the existing modules :+1: (I should probably post some explicit guidelines on standard procedure for the addition of new UI modules - I ought to fix that position attribute at some point too).

I'd like to see a setting on the add/remove modules to toggle whether it's all curators or just the current one (all by default), however I can easily add that after merge if you're not confident with the UI stuff.

Regarding the locality stuff:

Module functions will run globally by default (if inheriting from the base module), so they will run **everywhere**. The simulation module can thus just use the command without having to worry about locality (since it will be ran on all machines by the module framework).

The UI modules have their functions stem from the UI itself, so those are only run on the zeus client by default. You'll have to handle locality within those modules (hence running the add/remove object commands on the server).

addons/zeus/stringtable.xml
```diff
@@ -695,5 +701,8 @@
             <Polish>Wpisano nieprawidłowy promień</Polish>
             <German>Ungültiger Radius eingegeben</German>
         </Key>
+		<Key ID="STR_ACE_Zeus_ModuleSimulation_DisplayName">
+            <English>Toggle Simulation</English>
+        </Key>
     </Package>
 </Project>
```
commy2: newline

addons/zeus/CfgFactionClasses.hpp
```diff
@@ -0,0 +1,7 @@
+class CfgFactionClasses {
+    class ACE_UI_Util {
+        displayName = "ACE Util";
+        priority = 2;
+        side = 7;
+    };
+};
```
commy2: newline

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * None
+ *
+ * Public: No
+ */
+ 
+ #include "script_component.hpp"
+ 
+ params ["_logic"];
+ 
+ if !(local _logic) exitWith {};
```
commy2: @SilentSpike 
So this function runs everywhere... Which machine is the module local to? Server or Zeus client? Shouldn't this be a `isServer`check for simplicity?

RSpeekenbrink:
Thanks SilentSpike for the comment, in a way I was confused about the Simulation command since I also thought if you call enableSimulationGlobal it will automatically be executed from the server to all clients. I just tested the simulation script on a server without the `BIS_fnc_MP` and it works fine.

About the UI stuff, i'm currently learning how UI stuff works in ArmA so im not that confident yet.

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,28 @@
...
+
+ private _object = attachedTo _logic;
+ if (isNull _object) then {
+    [LSTRING(NoObjectSelected)] call EFUNC(common,displayTextStructured);
+ } else {
+    private _simulationEnabled = simulationEnabled _object;
+    [[_object, (!_simulationEnabled)],"enableSimulationGlobal",false] call BIS_fnc_MP;
```
commy2: that is not what @jonpas and I meant. Should probably be restored to what it was if the other comment regarding the modules locality is solved.

addons/zeus/functions/fnc_ui_addObjects.sqf
```diff
@@ -0,0 +1,65 @@
...
+
+    private _radius = GETVAR(_display,GVAR(radius),50);
+    private _position = GETVAR(_display,GVAR(position),getPos _logic);
+    private _objectsToAdd = nearestObjects [_position, ["All"], _radius];
+
+    {
+        {
+               _x addCuratorEditableObjects [_objectsToAdd, true];
+        } foreach allCurators;
+    } foreach _objectsToAdd;
```
commy2: that looks wrong

addons/zeus/functions/fnc_ui_removeObjects.sqf
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Public: No
+ */
+ 
+#include "script_component.hpp"
+ 
+disableSerialization;
```
commy2: @SilentSpike 
module scripts are run in scheduled environment, right?

addons/zeus/functions/fnc_ui_removeObjects.sqf
```diff
@@ -0,0 +1,65 @@
...
+
+    private _radius = GETVAR(_display,GVAR(radius),50);
+    private _position = GETVAR(_display,GVAR(position),getPos _logic);
+    private _objectsToAdd = nearestObjects [_position, ["All"], _radius];
+
+    {
+        {
+              _x removeCuratorEditableObjects [_objectsToAdd, true]; 
+        } foreach allCurators;
+    } foreach _objectsToAdd;
```
commy2: same as above. no need to iterate through `_objectsToAdd` at all. also, `forEach` camel case

RSpeekenbrink: Im a derp, I forgot to remove the other foreach loop

addons/zeus/functions/fnc_moduleSimulation.sqf
SilentSpike: Module will be local to zeus client. However in this case of this module, we may want to actually exploit the module framework and make it only run on the server (then use the simulationGlobal command).

To do so, in the module's cfgVehicles entry you simple set `isGlobal = 0;`

SilentSpike: (and remove this locality check yes)

RSpeekenbrink: then make the command instead of `enableSimulationGlobal` just `enableSimulation` ?

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -22,7 +22,8 @@
     [LSTRING(NoObjectSelected)] call EFUNC(common,displayTextStructured);
  } else {
     private _simulationEnabled = simulationEnabled _object;
-    [[_object, (!_simulationEnabled)],"enableSimulationGlobal",false] call BIS_fnc_MP;
+	_object enableSimulationGlobal (!_simulationEnabled);
```
commy2: tab. should be four spaces to let travis pass

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -22,7 +22,8 @@
...
  };
 
- deleteVehicle _logic;
\ No newline at end of file
+ deleteVehicle _logic;
+
```
commy2: white space here

commy2:
@SilentSpike the same problem exists for `fnc_ui_addObjects.sqf` (`addCuratorEditableObjects`) and `fnc_ui_removeObjects.sqf` (`removeCuratorEditableObjects`

SilentSpike:
Sorry for the conflict @RSpeekenbrink, thought I'd go ahead and optimise that module for you. Should have mentioned it first.

To avoid conflicts with our coding standards, I highly recommend you grab [EditorConfig](http://editorconfig.org) for whichever text editor you're using as it will automatically adopt EOF and white space conventions for the currently open project (if it has an editor config settings file - which we do).

RSpeekenbrink:
@SilentSpike I used Notepad++ but it was just a small merge error, fixed it though, no problems here :)

bux578:
https://github.com/editorconfig/editorconfig-notepad-plus-plus

addons/zeus/functions/fnc_ui_removeObjects.sqf
SilentSpike: These UI initialisation functions will stem from an `onSetFocus` event handler, which I don't believe is scheduled, not sure though.

SilentSpike:
All looking pretty good, only thing I see left to fix is locality of the add/remove modules.

However I'm inclined to perhaps merge those into a single module with a toggle for removal/addition so if you'd like I can do that and just fix the locality at the same time?

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,30 @@
...
+ } else {
+    if (isMultiplayer) then {
+        _object enableSimulationGlobal !(simulationEnabled _object);
+    } else {
+        _object enableSimulation !(simulationEnabled _object);
+    };
+ };
+
+ deleteVehicle _logic;
+
```
commy2: white space

addons/zeus/functions/fnc_ui_removeObjects.sqf
commy2: If they don't run in scheduled environment, then `disableSerialization` should be removed as unscheduled environment does not support serialization to begin with.

addons/zeus/functions/fnc_ui_addObjects.sqf
```diff
@@ -0,0 +1,63 @@
...
+disableSerialization;
+ 
+params ["_control"];
+private _display = ctrlParent _control;
+private _logic = GETMVAR(BIS_fnc_initCuratorAttributes_target,objNull);
+
+_control ctrlRemoveAllEventHandlers "setFocus";
+
+scopeName "Main";
+private _fnc_errorAndClose = {
```
commy2: function seems to be unused, remove?

addons/zeus/functions/fnc_ui_removeObjects.sqf
```diff
@@ -0,0 +1,63 @@
...
+disableSerialization;
+ 
+params ["_control"];
+private _display = ctrlParent _control;
+private _logic = GETMVAR(BIS_fnc_initCuratorAttributes_target,objNull);
+
+_control ctrlRemoveAllEventHandlers "setFocus";
+
+scopeName "Main";
+private _fnc_errorAndClose = {
```
commy2: same as above

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,30 @@
...
+ private _object = attachedTo _logic;
+ if (isNull _object) then {
+    [LSTRING(NoObjectSelected)] call EFUNC(common,displayTextStructured);
+ } else {
+    if (isMultiplayer) then {
+        _object enableSimulationGlobal !(simulationEnabled _object);
+    } else {
+        _object enableSimulation !(simulationEnabled _object);
+    };
```
commy2: I think this whole then block can be simplified to:
`_object enableSimulationGlobal !(simulationEnabled _object);`
`enableSimulation` and `enableSimulationGlobal` are synonymous in SP as the SP client is always the server.

SilentSpike: Yeah that's what I thought too, the biki seems to suggest that `enableSimulation` should be used instead of the global varient in SP so I wasn't sure

commy2: Hmm. Weird comment on the wiki. Should be tested.

RSpeekenbrink: enableSimulationGlobal seems to work in MP, enableSimulation doest

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,30 @@
...
+} else {
+    if (isMultiplayer) then {
+        _object enableSimulationGlobal !(simulationEnabled _object);
+    } else {
+        _object enableSimulation !(simulationEnabled _object);
+    };
+};
+
+deleteVehicle _logic;
+
```
commy2: white space : (

addons/zeus/functions/fnc_moduleSimulation.sqf
RSpeekenbrink: I tested enableSimulationGlobal in singleplayer aswell, it works fine on my end

SilentSpike:
I've merged the add/remove modules together, fixed their locality, added checkbox to toggle all/local curator.

All that's left now is some cleanup and this is good to go :+1:

SilentSpike:
Alright, I'm happy to merge this, but if someone else wants to give the code a quick re-review beforehand that'd be great :+1:

654wak654:
I'm assuming #4597 is going to need `category = QGVAR(Utility)` under class `moduleAddOrRemoveFRIES` because of this PR?

All of @alganthe's PRs will need it too: #4555 #4556 #4576

alganthe:
Will modify my PRs when this is merged.

commy2:
lgtm. I'd use a macro for the IDC's,but that is a general problem with the module and goes beyond this PR.

addons/zeus/XEH_postInit.sqf
```diff
@@ -14,3 +14,26 @@ QGVAR(GlobalSkillAI) addPublicVariableEventHandler FUNC(moduleGlobalSetSkill);
...
+
+        if !(isNull _curator) exitWith { _curator addCuratorEditableObjects [_objects, true]; };
+
+        {
+            _x addCuratorEditableObjects [_objects, true];
+        } forEach allCurators;
+    }] call CBA_fnc_addEventHandler;
+
+    [QGVAR(removeObjects), {
+        params ["_objects", "_curator"];
```
commy2: @SilentSpike 
Do you mean:
`params ["_objects", ["_curator", objNull]];`
so that the `exitWith` thing works as alternative syntax?
Alternativeley change `!isNull _curator` to `!isNil "_curator"`

same above

commy2: I know that this works anyway (at least in scheduled env.) due to `isNull` silently failing when `_curator` is undefined, but I still thing relying on this SQF ... "feature" is bad design.

addons/zeus/functions/fnc_ui_editableObjects.sqf
```diff
@@ -0,0 +1,67 @@
...
+ * None
+ *
+ * Public: No
+ */
+
+#include "script_component.hpp"
+
+params ["_control"];
+
+disableSerialization;
```
commy2: does nothing in unscheduled env. should be removed if these module scripts do indeed run unscheduled. otherwise ignore this

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * Public: No
+ */
+
+#include "script_component.hpp"
+
+params ["_logic"];
+
+private _object = attachedTo _logic;
+if (isNull _object) then {
+    [LSTRING(NoObjectSelected)] call EFUNC(common,displayTextStructured);
```
RSpeekenbrink: isn't it so that it runs on both server & local? @SilentSpike

SilentSpike:
@654wak654 @alganthe Modules don't necessarily need categories defined, if they inherit from the base module then it'll still default to `ACE`.

addons/zeus/XEH_postInit.sqf
SilentSpike: Good catch commy, I don't know why I totally forgot params can supply a default value when writing this :+1:

addons/zeus/functions/fnc_moduleSimulation.sqf
SilentSpike: Also a good catch, hmm, I suppose it probably should run on the client and then send an event to the server so that this client feedback can be given.

SilentSpike: Not optimal in terms of minimising network traffic, but much better UX for zeus and also these modules are only running when prompted by zeus anyway so it's not a big deal.

addons/zeus/functions/fnc_ui_editableObjects.sqf
SilentSpike: This is probably my bad if unnecessary as it's in all of my existing ui functions, not sure if I had reasoning for that when I first wrote them

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -15,11 +15,13 @@
 
 params ["_logic"];
 
+if !(local _logic) exitWith {};
```
commy2: so the module is local to the zeus? good to know

SilentSpike: Yep, this line is standard in all zeus modules (BI included) with a few exceptions.

SilentSpike:
Just tested and confirmed that `disableSerialization` is unnecessary for the ui functions. Will remove that from them all while I'm at it.

addons/zeus/functions/fnc_moduleSimulation.sqf
```diff
@@ -0,0 +1,27 @@
...
+
+params ["_logic"];
+
+if !(local _logic) exitWith {};
+
+private _object = attachedTo _logic;
+if (isNull _object) then {
+    [LSTRING(NothingSelected)] call EFUNC(common,displayTextStructured);
+} else {
+    [QEGVAR(common,enableSimulationGlobal), [_object, !(simulationEnabled _object)]] call CBA_fnc_serverEvent;
```
RSpeekenbrink: Just for learning purposes, what does QEGVAR exactly in this state? 
Thanks

commy2: The line makes use of this event defined in the `common` component:
https://github.com/acemod/ACE3/blob/master/addons/common/XEH_postInit.sqf#L135

RSpeekenbrink: Aah I get it, makes a lot of sense now, thanks 👍

PR #4668, opened 2016-11-13
Introduce small variation to length of cardiac arrest

SilentSpike:
- Add some minor variation to the length of cardiac arrest, can only vary up to 10% of the setting value with a Gaussian distribution that favours the setting itself


SilentSpike:
As suggested by @BaerMitUmlaut privately, a fixed 10% randomness to the existing cardiac arrest time might be a better idea. Has the advantage of keeping the settings simple, while keeping the predictability of cardiac arrest intact but not 100% certain.

I think I personally prefer this idea and would be in favour of re-purposing this PR to that effect.

PR #4689, opened 2016-11-18
Explosive Fixes

### LINKED ISSUE #4687: Clackers usable by only one player.
**CBA Version:** `3.1.1` (stable)
**ACE3 Version:** `3.8.1` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
ACE Explosives clackers cannot be used by anyone other than the player that armed them. This means that if the player that arms a charge with a clacker dies before setting it off, the only way to set off that explosive is to go back to it, defuse it, replace it, and rearm it. Ideally, another player could pick up the clacker and then set off the charge without having to go through that whole process.

**Steps to reproduce:**
- Have two players on a server.  
- Have Player 1 place and arm any explosive using any clacker.  
- Player 2 kills Player 1 and takes the clacker off the dead body.  
- Player 2 cannot detonate any charges armed by Player 1.

- Dedicated

**Placed Modules:**
- None.

### LINKED ISSUE #4688: Deadman's switch 'Detonate All' Infinite range
**ACE3 Version:** `3.8.1` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
The deadmans switch has an apparent range of 100 meters when detonating individual charges, but the 'detonate all' option has no apparent range limitation, tested out to 10+ kilometers.  

**Steps to reproduce:**
- Arm an explosive charge with the deadmans switch.  
- Travel more than 100 meters away from the charge.  
- From the self interact menu, try to detonate that device using its specific explosive code.  
- Note that the device does not explode.  
- Without getting any closer to the device, select the 'detonate all' function on the deadmans switch.  
- The device should now explode. 

- Editor (Singleplayer)  

**Placed Modules:**
- None.

PabstMirror:
- Fix deadman range (close #4688) in onIncapacitated.sqf
- Fix clacker transfer (close #4687) in onInventoryChanged
- Add use of `setShotParents` - should fix killed EH and correctly sets `ace_medical_lastDamageSource`

### DIFF
--- a/addons/explosives/XEH_postInit.sqf
+++ b/addons/explosives/XEH_postInit.sqf
@@ -22,6 +22,19 @@
 //When getting knocked out in medical, trigger deadman explosives:
 //Event is global, only run on server (ref: ace_medical_fnc_setUnconscious)
 if (isServer) then {
+    [QGVAR(detonate), {
+        params ["_unit", "_explosive", "_delay"];
+        TRACE_3("server detonate EH",_unit,_explosive,_delay);
+        _explosive setShotParents [_unit, _unit];
+        [{
+            params ["_explosive"];
+            TRACE_1("exploding",_explosive);
+            if (!isNull _explosive) then {
+                _explosive setDamage 1;
+            };
+        }, _explosive, _delay] call CBA_fnc_waitAndExecute;
+    }] call CBA_fnc_addEventHandler;
+
     ["ace_unconscious", {
         params ["_unit", "_isUnconscious"];
         if (!_isUnconscious) exitWith {};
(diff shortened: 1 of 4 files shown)


addons/explosives/XEH_postInit.sqf
```diff
@@ -22,6 +22,19 @@
 //When getting knocked out in medical, trigger deadman explosives:
 //Event is global, only run on server (ref: ace_medical_fnc_setUnconscious)
 if (isServer) then {
+    [QGVAR(detonate), {
+        params ["_unit", "_explosive", "_delay"];
+        TRACE_3("server detonate EH",_unit,_explosive,_delay);
+        _explosive setShotParents [_unit, _unit];
```
BaerMitUmlaut: Only in 1.65+

PabstMirror: biki must be wrong, working on 1.64

bux578:
@Glowbal would be nice to have this in 3.8.2

addons/explosives/XEH_postInit.sqf
commy2: It does work, but it's a server only global effects command.

Glowbal:
If this has been tested, we can add this to 3.8.2.

Glowbal:
Moving to 3.8.3, as this is still marked as `needs testing`

PR #4690, opened 2016-11-19
Fix camel case in Zeus addon

Neviothr:
- Change numerous commands to camel case in the Zeus addon.

### DIFF
--- a/addons/zeus/functions/fnc_bi_moduleCurator.sqf
+++ b/addons/zeus/functions/fnc_bi_moduleCurator.sqf
@@ -22,33 +22,33 @@ params ["_logic", "_units", "_activated"];
 if (_activated) then {
 
     //--- Terminate when not created on the server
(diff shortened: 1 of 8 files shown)


commy2:
As @jonpas said, these functions are amended versions of the ones found in the base game.
It's a really bad method, but very needed because Zeus severely lacks modability.
The changes made to these functions are very small tweaks and mainly add lines. The focus is not on improving them. The formating and capitalization is usually kept, but there are always eager changes made or changes made by regexp macros.
This is done so they can be updated should the original ones be changed in an Arma update. It's easier to tell the diff between the files if as few lines as possible are changed.
I'm not sure merging this would be what we want.

SilentSpike:
Appreciate the effort, but @jonpas and @commy2 have summed up the reasoning that this is intentional for maintenance purposes.

PR #4694, opened 2016-11-23
ST Nightvision Effects

### LINKED ISSUE #3886: Laserpointer is visible through NVG border
**CBA Version:** `2.4.1` (dev `f9a05d3266da57a0e5dfb7ac4aba5d326f5a92d8`)
**ACE3 Version:** `3.5.1` (dev `1c95c883e0d9d2fe93970d826c9be413320e21fa`)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Laserpointer is visible through NVG border.  
  http://i.imgur.com/RZcwXTn.jpg

**Steps to reproduce:**

- Dedicated

**Placed Modules:**
- None

**RPT log file:**
- N/A

PabstMirror:
Port Effects from ST Night Vission - Thanks to @Dslyecxi

Redone blur, grain, brightness, contrast and color correction.
Fog is now used as a visual effect to limit view distance.
NVG border is scaled dynamically based on player view's FOV.
NVG border mask is much smaller, representing a limited FOV of NVGs
But is also slightly transparent (based on nvg config)
Unlike vanilla NVG optics, this will block draw3d (e.g. nametags)
And will close #3886

This is still a WIP:
Currently looking for feedback on effects in different ambient lights and NVG Generations.
Code review can probably wait.


addons/nightvision/XEH_preInit.sqf
```diff
@@ -2,6 +2,23 @@
 
 ADDON = false;
 
+PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
+PREP_RECOMPILE_END;
```
commy2: What is this wizardry?

jonpas: Could be something from ST's framework.

addons/nightvision/XEH_preInit.sqf
```diff
@@ -2,6 +2,23 @@
...
+    (
+    [0.5,0.5]
+    distance2D
+    worldToScreen
+    positionCameraToWorld
+    [0,3,4]
+    ) * (
+    getResolution
+    select 5
+    ) / 2
```
bux578: From OP:
> Code review can probably wait.

PabstMirror: Not that much different from https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_getFov.sqf

addons/nightvision/XEH_preInit.sqf
jonpas: Or maybe it is wizardry -> #4699

addons/nightvision/XEH_preInit.sqf
Neviothr: I'm no expert, but KK's function seems more efficient.

commy2: But it's also probably very inaccurate.

commy2: What about this? https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getZoom.sqf

Neviothr: ACE3's takes **0.0083 ms**, while KK's takes **0.0100 ms**.

commy2: I don't trust the `distance` and `distance2D` commands. They're meant for terrains and not for vector math. If the terrain is sloped at the positions you are checking, it's like you are measuring the distance on a bend coordinate system, due to height differences. The command also strangely reports different values for the same position on different machines. I think some graphic settings responsible for terrain rendering affect the result too.

addons/nightvision/functions/fnc_nonDedicatedFix.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+if (isServer) exitWith {};
+
+[{
+    if (isNil QGVAR(serverPriorFog)) exitWith {
+        TRACE_1("Ending PFEH to handling non-dedicated server running effect",_this);
+        [_this select 1] call CBA_fnc_removePerFrameHandler;
+    };
+    if (!GVAR(running)) then { // If we aren't running the effect ourselves, then use the server's old fog value
+        0 setFog GVAR(serverPriorFog);
```
Glowbal: This may give conflicts with Zeus or our weather component.

addons/nightvision/XEH_preInit.sqf
Neviothr: It's probably the _terrain_ setting (the one which controls the amount of grass), the terrain geometry noticeably changes when changing it.

Neviothr:
I've noticed that the NVG overlay covers other UI elements such as stance, ammo count, etc. as can be observed [here](https://youtu.be/Lon3x0BO1bk). Is there anyway around that?

addons/nightvision/XEH_preInit.sqf
SilentSpike: I'm not sure how any of this would effect a script command that's just taking array values and working with pure maths :stuck_out_tongue_closed_eyes:

It sounds like your problems stem from `getPos` not `distance`

BryanBoru:
I kind of like that the overlay subdues the UI, it would be distracting otherwise.

Neviothr:
The action menu is a pain to use as it is, do you really want to turn it into a guessing game?

addons/nightvision/XEH_preInit.sqf
commy2: >I'm not sure how any of this would affect a script command that's just taking array values and working with pure maths 
point is, it doesn't

moonmanmove:
RHS compatibility?

jonpas:
@moonmanmove This is not a feature request topic, please keep it free of questions, you can join Slack for any information.

jameslkingsley:
> I've noticed that the NVG overlay covers other UI elements such as stance, ammo count, action menu, etc. as can be observed here. Is there any way around that?

From what I know, there is no way around that. There isn't a good way of handling the order of UI elements.

PabstMirror:
@moonmanmove - RHS compats updated in this PR

@Neviothr - Covering of the UI and Action menu is a notable downside to this.  But the more time I spent with the system, the less of a problem I thought it would be.  The border is still semi-transprent so you can still read the UI through the mask.



It does seem faint in screenshot, but I think it's different in-game as you eye's get used to the darkness.
It is a shame that this does have to be done via config. We could either add an untouched item, or leave the vanilla items alone and only modify our own.

addons/nightvision/functions/fnc_setupDisplayEffects.sqf
```diff
@@ -0,0 +1,78 @@
...
+        if (isServer) then {
+            missionNamespace setVariable [QGVAR(serverPriorFog), nil, true];
+        };
+    };
+};
+
+// Note: Using BIS_fnc_rscLayer because of bug with string syntax - https://feedback.bistudio.com/T120768
+(QGVAR(display) call BIS_fnc_rscLayer) cutText ["", "PLAIN"]; // Cleanup Old Display
+if (_activated) then { // Create New Display
+    (QGVAR(display) call BIS_fnc_rscLayer) cutRsc [QGVAR(title), "PLAIN", 0, false];
```
Neviothr: Wouldn't the game focus on the created control? Displaying a cursor and leaving the player unable to perform any actions?

ryanwoodcox:
I pulled down this branch and took a look.
Looks pretty awesome!
Two things:
1. When aiming down sight with optics such as EOTechs, EOTech + 3x magnifier, there is a 'blur' effect added. Is that intentional or an oversight?
2. Is there anything going to be added as far as a blinding effect when looking directly at light sources and/or flashbangs?

jonpas:
> 1. When aiming down sight with optics such as EOTechs, EOTech + 3x magnifier, there is a 'blur' effect added. Is that intentional or an oversight?

Intentional AFAIK, because NVGs can't adapt the lenses to focus like our eyes can, hence you get a blur effect, and that is why IR lasers are used.

> 3. Also, you probably want to remove the debugging messages that pop up when the branch ready to be merged.

Yes, that happens at the end. ;)

PabstMirror:
RE: flashbangs: 
I have heard that modern nightvision actually offers some protection from flashbangs.
They are designed to automatically limit light output, so you don't fry your eyes if someone turns on the lights.

commy2:
>Is there anything going to be added as far as a blinding effect when looking directly at light sources 
I don't think this is possible aside from obvious stuff like flashbangs, because the game does not give any feedback about the brightness of things you're looking at in any meaningful way that could be evaluated or manipulated on the SQF level.

ryanwoodcox:
Does anyone who's tried this out think its almost unplayable when walking through an open terrain, or in the jungle of Tanoa? I'm just wondering what high-end military grade NVG is capable of. IR emmitters, classified technology, etc. We obviously can compare to real life NVGs that civilians can get their hands on. I love the way this NVG enhancement feels but all in all it seems a bit too dark. I also noticed the brightness can only be set to 0 (no gain) and dimmed (negative gain).
I ran a quick mission last night using this and I was having to use an actual flashlight to see anything at all while moving through rural terrain. Flying and landing a helicopter was also nearly impossible.
Thoughts?

jonpas:
I think that's perfect. That's why flashlights are used, NVGs aren't able to enhance visibility in all conditions, I think dslyecxi's video touched on that as well. It also makes you want to use a flashlight because you might get better visibility with it, I personally like that.

ryanwoodcox:
Yeah, I mean the different generations and models of NVGs do better than overs. But yeah, if you locked yourself in a coffin, I don't think any NVGs would do anything... unless they can emit IR. I'm wondering if that's something to consider for some of the high end NVGs. Also, is there an IR flashlight? It would be nice to enhance the visibility while viewing via NVGs but not alert others in the visible light spectrum by using a normal flashlight.

bux:
> Yeah, I mean the different generations and models of NVGs do better than overs

true

> But yeah, if you locked yourself in a coffin, I don't think any NVGs would do anything

true

> Also, is there an IR flashlight?

there is

> It would be nice to enhance the visibility while viewing via NVGs but not alert others in the visible light spectrum by using a normal flashlight.

not possible in Arma

ryanwoodcox:
So if there is an IR flashlight, why is it not possible in ARMA to enhance the visibility via NVG without alerting folks viewing in the visible light spectrum?

jonpas:
Because there is no way to programmatically get the status of the light like it would be needed. All you can get is current brightness IIRC (like it's used for map), but obviously IR lights don't emit actual light.

ryanwoodcox:
Not sure how the IR Strobe and IR Grenade work in the game, but would that be a start, or am I missing something?

commy2:
Yes. The lack of scripting API in the game.

ryanwoodcox:
So there is no way to emit IR light in ARMA that can be seen via NVG and not the human eye?
So the IR Grenade and IR Strobe create just fake 'whiteness' that does not actually produce light...? I thought they would help illuminate rooms when in NVG mode.

jonpas:
No, IR lights/grenades/strobes work as intended, but there is no way to get any data about the brightness of the IR, or even if it's present or not.

PARTYzanRnD:
Hello guys, looks like fog feature do not work with "new" arma 3 fog mechanics, spawning strong fog on altitude it does not presents (setfog N instead of setfog [fogValue, fogDecay, fogBase])

PARTYzanRnD:
Also, NV goggles generation overrides rifle night scope generation
reproduce: player linkitem "nvgoggles"; player addweapon "arifle_MXC_Holo_pointer_F";player addprimaryweaponitem "optic_Nightstalker"; check rifle scope blur, aim off from the sight, player linkitem "ACE_NVG_Gen1", check blur in rifle scope view

Neviothr:
Not to rush things, but uh... any update on this?

Fusselwurm:
Concerning fog: it's really irritating to wear the NVG and *not* see lights in the distance that you *can* see with naked eye. Even with the reduced fog for air units.
Especially when flying *I've found myself continually switching* between NVG for close range (ground tracking) and Mk1 Eyeballs for higher range (see village lights that are >500m away).

I guess st_nvg_MinFog/st_nvg_MaxFog are good candidates for customizable settings :)

Personally, I reduced the NVG fog from `0.2…0.3` down to `0.075…0.125`  – this seems to me the point where I'm able to discern far-away lights well enough for my taste. 

/2c

Apart from that: I love it! It's certainly not perfect, but faaaar better than anything we've had before. 
This is the first time I believe I've created & customized my own ACE build to be able roll out a new feature for my unit :D 

:+1:

nomisum:
cant wait for this. is there any gamebreaker/help needed to merge this?

bux:
@nomisum Testing, testing, visual feedback

jonpas:
> Unlike vanilla NVG optics, this will block draw3d (e.g. nametags)

Is there anything that can be done about that? It will also block Advanced Throwing lines and anything in the mission that might rely on that.

addons/nightvision/XEH_PREP.hpp
```diff
@@ -1,7 +1,13 @@
+LOG("prep");
```
jonpas: Remove.

addons/nightvision/XEH_postInit.sqf
```diff
@@ -0,0 +1,94 @@
...
+    ["visionMode", LINKFUNC(onVisionModeChanged), false] call CBA_fnc_addPlayerEventHandler;
+    ["cameraView", LINKFUNC(onCameraViewChanged), true] call CBA_fnc_addPlayerEventHandler;
+    ["vehicle", LINKFUNC(refreshGoggleType), false] call CBA_fnc_addPlayerEventHandler;
+    ["turret", LINKFUNC(refreshGoggleType), true] call CBA_fnc_addPlayerEventHandler;
+
+    ["ace_firedPlayer", LINKFUNC(onFiredPlayer)] call CBA_fnc_addEventHandler;
+    ["ace_firedPlayerVehicle", LINKFUNC(onFiredPlayer)] call CBA_fnc_addEventHandler;
+
+
+    addMissionEventHandler ["Loaded", { // Restart UI vars on misison load
```
jonpas: `mission` in comment.

addons/nightvision/XEH_preInit.sqf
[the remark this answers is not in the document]
jonpas: Should probably use https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getZoom.sqf as stated above.

addons/nightvision/functions/fnc_nonDedicatedFix.sqf
jonpas: It is disabled in weather component at the end of this diff. Fog actually gets synced correctly by Arma which was proven not long ago.

PabstMirror:
It doesn't block all draw3d, just part that is limited from the NVG FOV overlay

addons/nightvision/XEH_PREP.hpp
PabstMirror: I really like these for verify the recompile took effect.
`LOG` macro is disabled normally
I'll remove for now but maybe we can add them to all modules

PabstMirror:
The fog stuff is just a messy workaround for non-dedicated servers.
Problem is that fog does sync correctly, so we have to handle the host turning on nvgs

jonpas:
> It doesn't block all draw3d, just part that is limited from the NVG FOV overlay

Aah, that's even better then!

addons/nightvision/functions/fnc_nonDedicatedFix.sqf
```diff
@@ -0,0 +1,30 @@
+/*
+ * Author: PabstMirror
+ * Handles the client who is the non-dedicated server turning on the fog effects
```
jonpas: Missing final dot.

addons/nightvision/functions/fnc_refreshGoggleType.sqf
```diff
@@ -0,0 +1,129 @@
+/*
+ * Author: Dslyecxi, PabstMirror
+ * Determins night vision source (player/vehicle) - Updates UI based on type
```
jonpas: `Determines` and missing final dot.

addons/nightvision/functions/fnc_scaleCtrl.sqf
```diff
@@ -0,0 +1,31 @@
+/*
+ * Author: Dslyecxi, PabstMirror
+ * Determins night vision source (player/vehicle) - Updates UI based on type
```
jonpas: Same as above.

addons/nightvision/functions/fnc_setupDisplayEffects.sqf
```diff
@@ -0,0 +1,77 @@
+/*
+ * Author: Dslyecxi, PabstMirror
+ * Handles setting up the effects: fog, ppEffects and the RscTittle
```
jonpas: Missing final dot.

addons/nightvision/functions/fnc_pfeh.sqf
```diff
@@ -0,0 +1,160 @@
+/*
+ * Author: Dslyecxi, PabstMirror
+ * Pfeh to handle refreshing effects.
```
jonpas: `PFEH`

Tittoffer:
Quick review after testing it a couple of hours yesterday. 

- The flickering effect appears in RCWS and turret optics. Intentional? 
- Action menu and general UI/HUD is hidden with vanilla NVG but not RHS.

Lastly the possibility to expand adjust brightness in-game, currently goes from minus six to zero. Also seeing that the module is quite aggressive, would it be possible to add further settings to the module itself (fogscaling, effectscaling, blur weapon in sighted perspective).

jonpas:
> Action menu and general UI/HUD is hidden with vanilla NVG but not RHS.

Not  running RHS optional.

jonpas:
With #5710 `ace_weather` will no longer touch fog, as it syncs fine on its own, which should make it fully compatible with this PR.

Trixas:
> Currently looking for feedback on effects in different ambient lights and NVG Generations.
I tried finding where I can edit vanilla generations but I couldn't. It's only those 4 different examples or rhs compat ones. Wanted to mess around it, because Gen 1 Noise is exaggerated.

Neviothr:
@Trixas -
> I tried finding where I can edit vanilla generations but I couldn't. It's only those 4 different examples or rhs compat ones. Wanted to mess around it, because Gen 1 Noise is exaggerated.
Generation 1 NVG technology dates back to the early 1960's, what do you expect it to look like?

Trixas:
>Generation 1 NVG technology dates back to the early 1960's, what do you expect it to look like?

Picture1
That pic is from my armasight spark core (ceramic intensifier tube but that's mostly a durability thing), taken at twilight before sunrise.

ZeusWrath:
Hi the ACE team!  

As seems to me, The new NightVision is the worst update of late.
At times it looks as if a mentally ill person did it. 
Several reasons at the same time:

1. Mask Closes the action menu of the AI ​​units, closes the HUD and the entire player interface! Now, the ACE3 player Can not play singleplayer missioms and control their subordinates at the night. @Neviothr said true - "The action menu is a pain to use as it is, do you really want to turn it into a guessing game?"

2. It adds a semi-transparent black mask! What for!? Really no one can understand that this looks silly, because the black mask exactly is the imitation of the not transparent parts of the night vision goggles! It should be fully black and not transparent in order to show the shape of the glasses (its transparent parts)

3. This adds completely non-logical Blur in the center of the screen, in the aiming point of ACE3 player and only when the player is aiming (without holding the right mouse button). Fuuck!  It's not at all clear why at this moment and why in the center of the screen!? Without aiming of player, this Blur does not exist! Probably someone just wanted to complicate the life of the ACE3 player. Lol!

4. Now, there is no way to increase the brightness. Now this feature is broken! ACE3 player has only a decrease in brightness! This creates a lot of inconvenience.

5. Of the large graininess and narrow field of view, the battle with the AI
 (which has an inhuman firing accuracy) become almost impossible. NVG simply complicates the vision of ACE player, but does not improve it!

ACE3 It's not at all clear why you took this system to your mod! This only creates problems for the player and looks, in places ugly and not realistic. I remember approximately about a 1.5-2 years ago in the ACE3 mod was a very good night vision! It had a large graininess, did not cover the player interface and did not complicate the life of the ACE player, as the new system does. 

The new system is not like one hundred or adequate and the player is forced to disable it (

jonpas:
Every single one of those points is addressed in Dslyecxi's video, and can you stop bringing up same issues over and over again when they have been explained to you, just stop.

I don't know what you are expecting, but calling Dslyecxi or @PabstMirror "mentally ill" surely motivates them! (sarcasm) Go make your own ACE3 if you can't even retain your complaints in their respective issues and don't want to take our explanations.

PR #4696, opened 2016-11-23
Release v3.8.2

GokouZWAR:
- Fix the minimum time on field dressing for crush wounds minimum reopen time to 200 seconds from 20 seconds.


commy2:
@devs Do not click "Delete branch" ^^

jonpas:
Can always restore it. ;)

GokouZWAR:
I screwed up when i made this pull request. Use the other one that actually has the change in it.

PR #4697, opened 2016-11-23
Update ACE_Medical_Treatments.hpp

GokouZWAR:
Fixed crush wound minimum time for field bandages from 20 seconds to 200 seconds.

### DIFF
--- a/addons/medical/ACE_Medical_Treatments.hpp
+++ b/addons/medical/ACE_Medical_Treatments.hpp
@@ -752,7 +752,7 @@ class ACE_Medical_Advanced {
                 class CrushWound: Abrasion {
                     effectiveness = 1;
                     reopeningChance = 0.2;
-                    reopeningMinDelay = 20;
+                    reopeningMinDelay = 200;
                     reopeningMaxDelay = 1000;
                 };
                 class CrushWoundMinor: CrushWound {


GokouZWAR:
seems like this was a miss-key by the original author. No other items are under 100 seconds and 20 seconds to reopen on any injury is kinda crazy since a lot of these bandages take about 5 - 10 seconds to apply already. You'd be constantly fighting reopens on this particular injury and bandage type with this low of a time.

ulteq:
> No other items are under 100 seconds

50 seconds: https://github.com/GokouZWAR/ACE3/blob/735e8cd946b65d0111bb78665de39d1dbd67dd00/addons/medical/ACE_Medical_Treatments.hpp#L1063

80 seconds: https://github.com/GokouZWAR/ACE3/blob/735e8cd946b65d0111bb78665de39d1dbd67dd00/addons/medical/ACE_Medical_Treatments.hpp#L1082

ulteq:
> seems like this was a miss-key by the original author
That's a question for @Glowbal

GokouZWAR:
Right I understand that but if you compare to field bandages... none of the other field bandages are under 100 seconds. 

Besides reopening in 20 seconds when it takes 10 seconds to apply a bandage is crazy low. If you have 3 injuries you are chasing one and while you're working on another it'll reopen. The guy hasn't even moved yet and the wound would reopen? Very unrealistic. I think you can agree on that...

ulteq:
> I think you can agree on that...

I do agree on that

Glowbal:
It's a minimum delay, meaning at the earliest it would be 20 seconds after bandaging that it would reopen. The system will take a random value somewhere between 20 seconds and 1000 seconds. Since `random` uses Gaussian for its distribution, that means the most likely value is around the 500 seconds.

This may or may not be a bug. I think this is indeed a typo, as @GokouZWAR suggested. I do not think it's a huge issue however.

Depending on the state of the medical rewrite, this fix may or may not go into the master branch. When I finished going through the current state of the medical rewrite, I will revisit this and see what is necessary.

GokouZWAR:
Never really said it was a huge issue, just wanting to get it in line with the other injury types, that's why I knocked it out real quick last month. If there were no other code conflicts I figured it could be slipped in for the time being without much trouble while we wait for medical rewrite to finish... the way this can be avoided would be to setup a config file that contains all this data so it can be tweaked by the end users. This might be a consideration to be added as a feature request outside this for the medical rewrite to enable end users to modify the data without having to go thru a code recompile each time. Just a thought.

Gokouzwar

ulteq:
We will merge it once the medical-rewrite is finished. Unless we change the way how bandaging works.

commy2:
#4064

PR #4699, opened 2016-11-23
Add dev macro to quickly recompile functions

PabstMirror:
Only effects dev enviroment

If you have `DISABLE_COMPILE_CACHE` macro set, this will add the code from `xeh_prep.hpp` to an array, which can easily be called mid mission.

So you can add a addAction/keybind/pfeh that runs 
`[] call ACE_PREP_RECOMPILE;` 
and all functions will be recompiled mid-mission.

Also adds `#define LINKFUNC(x)` which acts just like `FUNC(x)` normally, but when used with `DISABLE_COMPILE_CACHE` will become `{_this call FUNC(x)}`.
This acts sort of like "pass by reference", and will be updated with each recompile.

E.G. `player addEventHandler ["Fired", LINKFUNC(firedEH)];`
Will run updated code after each recompile.

No need to restart missions.  Save code, recompile, test.

### DIFF
--- a/addons/advanced_ballistics/XEH_preInit.sqf
+++ b/addons/advanced_ballistics/XEH_preInit.sqf
@@ -2,6 +2,8 @@
 
 ADDON = false;
 
+PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
+PREP_RECOMPILE_END;
 
 ADDON = true;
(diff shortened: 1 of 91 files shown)


commy2:
👍

jonpas:
This should be added to `blank` component in `extras` as well.

Issue #4573, opened 2016-10-19
No way to rearm static crew served weapons such as the XM312


thecuban:
**Arma 3 Version:** 1.64
**CBA Version:** 3.1.1.161012
**ACE3 Version:** 3.8.0

**Mods:**
- @CBA_A3
- @ace

**Description:**
- There exists an option for static weapons such as the XM312 to check the amount of available ammo the weapon has remaining in its magazine/inventory. However once the weapon is out of ammunition there is no way to rearm the static hmg in the field.

**Steps to reproduce: (aka the "dance")**
- place a static crew served weapon such as the XM312
- shoot all of its ammo
- sit around because your SOL due to a lack of ammo

**Where did the issue occur?**
- All game modes

**Placed Modules:**
- None

**Requested Change:**
- add ammo cans full of ammo (ex. 100 rd 12.7x99mm ammo can (for XM312)) that can be carried in your inventory to rearm static crew served weapons similar to the way ammo is used on the MK-6 mortar with all the advanced functions enabled.

kamkill01011:
You can rearm static weapons with ammo truck, as you can rearm vehicles.

thecuban:
Yes I am aware of that, however that's not how it works in the real world. In the real world a machine gun team would carry belts of ammo on foot into the field. While it is obviously possible, the machine gun (or any other crew served weapon) team would not always have a vehicle at their disposal to carry ammo and equipment.

PabstMirror:
Ideally it would work similar to the mk6 with ammo handling.
We would need to add vehicle ammo magazine items for each weapon.
E.G. A 100 round belt of 50 BMG that could then be loaded into a static mg.

thecuban:
Yes exactly! That's what I suggested as a change to fix this issue.

commy2:
feature request

PR #4701, opened 2016-11-23
Magazines for Crew Served Weapons

PabstMirror:
Ref #4573

Adds unit carry-able magazine that can be loaded into static weapons.
Ideas borrowed partialy from ACE2.

Right now it just supports .50 BMG, .50 Soviet, A3 20mm GMG, Titans and the mk6 mortars (using the shells from mk6 ammo handling)

I would love to add belt linking like ace_reload, but most a3 weapons start with multiple magazines, which causes problems with `setVehicleAmmo`/`setMagazineTurretAmmo` commands.
Still trying to think of ways around this.


addons/crewserved/CfgMagazineGroups.hpp
```diff
@@ -0,0 +1,80 @@
...
+    
+    // Soviet HMG (12.7x108mm)
+    class ACE_50Rnd_127x108_carryable {
+        rhs_mag_127x108mm_50 = 1;
+        rhs_mag_127x108mm_100 = 1;
+        rhs_mag_127x108mm_150 = 1;
+        CUP_150Rnd_127x108_KORD_M = 1;
+        CUP_50Rnd_127x108_KORD_M = 1;
+        CUP_50Rnd_TE3_LRT5_127x107_DSHKM_M = 1; // not sure why cup uses 107 for the DSHKM?
+        CUP_150Rnd_TE3_LRT5_127x107_DSHKM_M = 1;
```
VKing6: I'm pretty sure this was a typo introduced by BIS in Arma 2, or maybe even Arma 1.

addons/crewserved/CfgMagazineGroups.hpp
```diff
@@ -0,0 +1,80 @@
...
+        8Rnd_82mm_Mo_Flare_white = 1;
+    };
+    class ACE_1Rnd_82mm_Mo_HE_Guided {
+        8Rnd_82mm_Mo_guided = 1;
+    };
+    class ACE_1Rnd_82mm_Mo_HE_LaserGuided {
+        8Rnd_82mm_Mo_LG = 1;
+    };
+    
+    // A3 Titans (Spike) - just use handheld magazines (Does this make sense?)
```
VKing6: Yea, It's supposed to be the same launcher on a tripod, so popping the same missile in either makes sense.

addons/crewserved/functions/fnc_actionsUnload.sqf
```diff
@@ -0,0 +1,77 @@
...
+
+private _actions = [];
+private _handeledMagTypes = [];
+
+private _statement = {
+    params ["_target", "_player", "_params"];
+    _params params ["_vehMag", "_turretPath", "_carryMag"];
+    TRACE_5("Starting unload mag progres bar",_target,_turretPath,_player,_carryMag,_vehMag);
+
+    if (!(_player canAdd _carryMag)) exitWith {
```
VKing6: Drop magazine on the ground if there's no room in the inventory?

VKing6:
Intended that you can load magazines into a weapon with magazines in it?

PabstMirror:
@VKing6 - There is an ace setting for that.

Most guns come with multiple mags in them (e.g. the b_hmg comes with 5x `100Rnd_127x99_mag_Tracer_Red`)
I wasn't sure the best way to handle this.  After shooting 3 belts on a static weapon, one would think they could re-add belts to return it to it's original state.

Ideally we could guarantee all statics just come with a single magazine loaded, but I don't think that's possible.

VKing6:
Ah, yes, I see how the module setting works now.

Yea, that'll probably work okay.

Whipgit:
Is this still a WIP?

PabstMirror:
Merged into #5652

PR #4710, opened 2016-11-25
Jpn translation v382

classicarma:
Japanese translation for v3.8.2 and some fix


Glowbal:
Added to 3.8.3 release branch

PR #4713, opened 2016-11-26
CH53 Fastroping and other fixes

magicsh0tz:
Added fastrope points for CH53.

The getOut EH is required to keep the doors open when a unit exits the
vehicle and it uses the vehicle animation "doorHandler_R" and
"doorHandler_L" as a placeholder to determine if the doors are
open(1)/close(0).
eg. In UH1Y (Unarmed), when the right door is opened using the
useraction, "doorRB" (animateDoor) and "doorHandler_R" (animate) will be set to 1.


BaerMitUmlaut:
Pretty sure your changes to the getIn/Out EH will cause issues when on the ground. Needs testing.

magicsh0tz:
Currently, the getOut EH will only be executed if "ace_fastroping_doorsLocked" is false but the doors on the UH1Y and UH60M still closes automatically (on some occasions) if FRIES is "prepared" and when a player gets out.

So my suggestion is to allow the getOut EH to handle the door as per normal (without any conditions) while in "fnc_onPrepare.sqf" we should set animate "doorHandler_R" and "doorHandler_L" accordingly. As the "rhs_fnc_uh60_doors" function uses "doorHandler_R"/"doorHandler_L" to determine if the doors should be left open if a unit gets out.
<br>

Original <b>rhs_fnc_uh60_doors</b> (uh60_doors.sqf) code:
```
/*
a: reyhard

door handling script
while doors are opened by action make sure they stay in such state
*/

if (_this select 1 == "cargo") then {
    private["_v"];

    _v = _this select 0;

    if (_v animationPhase "doorhandler_R" == 1) then {
        _v animateDoor ["doorRB",1];
    } else {
        if (_v doorPhase "doorRB" != 0) then {
            [_v,"doorRB"] spawn rhs_fnc_doorHandler;
        };
    };

    if (_v animationPhase "doorhandler_L" == 1) then {
        _v animateDoor ["doorLB",1];
    } else {
        if(_v doorPhase "doorLB" !=0) then {
            [_v,"doorLB"] spawn rhs_fnc_doorHandler;
        };
    };
};
```

BaerMitUmlaut:
I see, interesting. Have you tested if it works as expected?

magicsh0tz:
Yup, I've tested it and the doors stay open when intended

Glowbal:
@BaerMitUmlaut could you take another look at this?

PR #4715, opened 2016-11-26
RangeCard - Take bore height into account

ulteq:
- Use the actual bore height for the range card calculation


addons/rangecard/RscTitles.hpp
```diff
@@ -192,7 +192,7 @@ class ACE_RangeCard_Dialog {
         class FOOTNOTE_TEXT_2: FOOTNOTE_TEXT_1 {
             idc=770021;
             y="safezoneY+1.72431+0.024";
-            text="and load with optic mounted 1.5'' above line of bore.";
+            text="and load with optic mounted 3.81 cm above line of bore.";
```
ulteq: Note to self, revert this change.

PR #4722, opened 2016-11-27
AdvFatigue - Remove setVar when removing eh

PabstMirror:
Removes the setVariable after removing the event handler.

Otherwise the EH would not be re-added when switching back to a previously controled unit.

Start Unit A - Add EH to A
Zeus Unit B - Remove from A, Add EH To B
Return to A - Remove from B, Would skip Adding to A because setVar was present


Glowbal:
Added to 3.8.3 release branch

PR #4732, opened 2016-11-30
RealisticNames/Weights: Add SVDK

KaiSforza:
Rename the Cyrus to the SVDK[1], change weight to more closely match the
Noreen Bad News' weights as it's only about 500g heavier than the Noreen.

[1]: https://en.wikipedia.org/wiki/SVDK

Note, the stock weight is 300, more than double what the Noreen weighs, so it puts the CSAT Marksman class at a massive weight disadvantage.


ulteq:
lgtm

J3aLeiNe:
Agreed with @BaerMitUmlaut, the Cyrus seems to be a sort of customized Saïga recalibrated by BI with the SVDK's ammos. A transgenic weapon by Doctor Bohemia.

ulteq:
`mass = 140;` looks good though, doesn't it?

BaerMitUmlaut:
Yes, the mass seems reasonable.

PR #4735, opened 2016-12-01
Body Temperature Model v2

Pasrules:
Enhances the temperature based fatigue model.

Issue: Model is correct in heating and cooling aspects yet something is unaccounted for as physical exertion has little effect against the cooling effect which is physically correct but too powerful, factored down by 200x.

Tasks:
- ~Rain based cooling, as high humidity removes cooling effect~ Fix poor theory
- Module to switch Simple/Dynamic
- Medical Diagnosis of Temperature
- Visual effects (shiver, heatstroke)
- Heat pack (item)
- Water bottle (item)

### DIFF
--- a/addons/advanced_fatigue/functions/fnc_calculateBodyFactors.sqf
+++ b/addons/advanced_fatigue/functions/fnc_calculateBodyFactors.sqf
@@ -1,0 +1,41 @@
+/*
+* Author: Pasrules
+* Calculates body factors
+*
+* Arguments:
+* 0: bodyTemperature <NUMBER>
+*
+* Return Value:
+* 0: heatStress <NUMBER>
+* 1: sweat <NUMBER>
+* 2: shiver <NUMBER>
+*
+* Example
+* [37] call ace_advanced_fatigue_fnc_calculateBodyFactors
+*/
+
+#include "script_component.hpp"
+params ["_bodyTemp"];
+
+private _tempDif = (_bodyTemp - MALE_TEMPERATURE);
+
+private _heatStress = 0;
+// Check if overheating
+if (_tempDif > 1.5) then {
+    _heatStress = (500 * _tempDif);
+};
+
+private _sweat = 0;
+// Check if sweating
+if (_tempDif > 0) then {
+    _sweat = _tempDif;
+    _sweat min 2;
+};
+
+private _shiver = 0;
+// Check if shivering
+if (_tempDif < -1.5) then {
+    _shiver = 500 * (abs _tempDif);
+};
+
+[_heatStress,_sweat,_shiver];
--- a/addons/advanced_fatigue/functions/fnc_calculateBodyFactors.sqf
+++ b/addons/advanced_fatigue/functions/fnc_calculateBodyFactors.sqf
@@ -20,21 +20,18 @@ params ["_bodyTemp"];
 private _tempDif = (_bodyTemp - MALE_TEMPERATURE);
 
 private _heatStress = 0;
-// Check if overheating
-if (_tempDif > 1.5) then {
+if (_tempDif > 1.5) then { // Check if overheating
     _heatStress = (500 * _tempDif);
 };
 
 private _sweat = 0;
-// Check if sweating
-if (_tempDif > 0) then {
+if (_tempDif > 0) then { // Check if sweating
     _sweat = _tempDif;
     _sweat min 2;
 };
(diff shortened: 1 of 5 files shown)


addons/advanced_fatigue/functions/fnc_calculateBodyTemperature.sqf
```diff
@@ -0,0 +1,56 @@
...
+    //Body
+    private _factors = [_bodyTemp] call FUNC(calculateBodyFactors);
+
+    //Environment
+    private _sun = 120 * (1 - overcast);
+    private _air = (_temperature - MALE_TEMPERATURE) * E ^ _windFactor * _airDensity;
+    private _environment = (_sun + _air) * BODY_SURFACE_AREA;
+
+    //Heating, muscles produce 80% waste heat
+    private _speed = vectorMagnitude velocity player min 10;
```
PabstMirror: Need to use `ace_player`.  
Needs to handle a player in a vehicle or free falling.

addons/advanced_fatigue/functions/fnc_calculateBodyTemperature.sqf
```diff
@@ -0,0 +1,56 @@
...
+    private _metabolicFactor = linearConversion [0, 1200, _metabolicCosts, 0, 1.4, true];
+    private _metabolism = BASAL_METABOLIC_WATTAGE * (1 + _metabolicFactor);
+    private _wasteHeat = (150 * _metabolicFactor * 4); //male output * cost * 4 
+    private _shiver = (_factors select 2);
+    private _heat = (_metabolism + _shiver + _wasteHeat);
+
+    //Cooling
+    private _sweat = _factors select 1;
+    private _wet = (_sweat + rain) min 1;
+    private _cooling = (100 - humidity) * E ^ (_wet + _windFactor) * BODY_SURFACE_AREA / _airDensity; //2.718 is Euler constant
```
PabstMirror: `humidity` seems to return 0..1

Pasrules: Good spot

addons/advanced_fatigue/functions/fnc_calculateBodyFactors.sqf
```diff
@@ -0,0 +1,41 @@
...
+// Check if sweating
+if (_tempDif > 0) then {
+    _sweat = _tempDif;
+    _sweat min 2;
+};
+
+private _shiver = 0;
+// Check if shivering
+if (_tempDif < -1.5) then {
+    _shiver = 500 * (abs _tempDif);
```
PabstMirror: Simulating low body tempature is going to be very difficult because in real life you can wear approprite clothing that provides insulation.
And I think adding configs to uniforms would just add too much micromanagement to mission makers.

addons/advanced_fatigue/functions/fnc_calculateBodyTemperature.sqf
Pasrules: `min 10` should handle free fall, I'll add in a vehicle check

addons/advanced_fatigue/functions/fnc_calculateBodyFactors.sqf
Pasrules: low body temperature only occurs in extreme conditions after 30-60mins of exposure which the shiver value will equalize causing fatigue. I found this to be the best approach as not all clothing has sleeves for example

addons/advanced_fatigue/functions/fnc_calculateBodyTemperature.sqf
AACO: I think Pabo was saying that if you're falling or in a vehicle, your speed would technically be zero, since you're not exerting any effort to fall/sit.

11RDP-LoupVert:
Gear player is wearing may have an effect on cooling, eg. wearing an helmet doesn't allow heat to be evacuated quite as effectively than while wearing a shemag for example.

Isn't CSAT unforms having an embedded cooling system btw?

Pasrules:
A complicated system which needs a complete rewrite.

Simplified more simple system:
Check if the player is wearing headgear, check the armor level on the headgear and add a fatigue factor depending on the armor level. Helmets make you hot, soft covers not so much take off your helmet but don't get hit in the head.

PR #4738, opened 2016-12-01
Realisticweights: Updated cyrus weight

KaiSforza:
Cyrus was set to weigh 300 units (~30lb?). The NATO equivalent, the Noreen
Bad News (or MAR-10) weighed 130, or 13lbs, equivalent to it's real life
counterpart. This just changes the weight to 140, similar to that of the
SVDK, a possible real-life counterpart for this rifle.

Fixes #4732


BaerMitUmlaut:
Thanks for your PR!

PR #4743, opened 2016-12-02
AdvThrow - Make sure ammo setVar is not empty

PabstMirror:
Just to be safe


jonpas:
This can go into 3.8.3 (if we do that) I think.

Glowbal:
Nice going @PabstMirror

Glowbal:
Added to 3.8.3 release branch

PR #4749, opened 2016-12-06
fix RHS air vehicles eject like in ace_aircraft

Dystopian:
- also RHS GREF compatibility addon created

Sometimes class inheritance looks like a kind of magic for me, so there can be errors.
I used modified fnc_exportConfig script from `nouniformrestrictions` to find all not ejectable classes.

### DIFF
--- a/optionals/compat_rhs_afrf3/CfgVehicles.hpp
+++ b/optionals/compat_rhs_afrf3/CfgVehicles.hpp
@@ -287,7 +287,7 @@ class CfgVehicles {
     class Heli_Light_02_base_F: Helicopter_Base_H {
         class Turrets: Turrets {
                 class CopilotTurret;
-                class MainTurret;
+                class MainTurret; // from RHS, not in vanilla
         };
     };
     class RHS_Mi8_base : Heli_Light_02_base_F {
--- a/optionals/compat_rhs_afrf3/CfgVehicles.hpp
+++ b/optionals/compat_rhs_afrf3/CfgVehicles.hpp
@@ -282,13 +282,27 @@ class CfgVehicles {
     class Helicopter_Base_F;
     class Helicopter_Base_H: Helicopter_Base_F {
         class EventHandlers;
+        class Turrets;
     };
-    class Heli_Light_02_base_F: Helicopter_Base_H {};
+    class Heli_Light_02_base_F: Helicopter_Base_H {
+        class Turrets: Turrets {
+                class CopilotTurret;
+                class MainTurret;
+        };
+    };
     class RHS_Mi8_base : Heli_Light_02_base_F {
         EGVAR(refuel,fuelCapacity) = 3700;
         EGVAR(fastroping,enabled) = 0;
         class EventHandlers: EventHandlers {
             class RHS_EventHandlers;
+        };
+        class Turrets: Turrets {
+            class CopilotTurret: CopilotTurret {
+                canEject = 1;
+            };
+            class MainTurret: MainTurret {
+                canEject = 1;
+            };
         };
     };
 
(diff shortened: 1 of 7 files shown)


PabstMirror:
I'm not against this, 
But this is more "tweaks to RHS" and not RHS/ACE compatibility.

It might be worthwhile to report not being able to eject to the RHS devs.

optionals/compat_rhs_afrf3/CfgVehicles.hpp
```diff
@@ -282,14 +282,28 @@ class CfgVehicles {
     class Helicopter_Base_F;
     class Helicopter_Base_H: Helicopter_Base_F {
         class EventHandlers;
+        class Turrets;
+    };
+    class Heli_Light_02_base_F: Helicopter_Base_H {
+        class Turrets: Turrets {
+                class CopilotTurret;
+                class MainTurret;
```
commy2: `Heli_Light_02_base_F` does not have `MainTurret` in `class Turrets` in vanilla. This causes potential problems with other helis inheriting this `Turrets` passively by introducing an invisible gunner slot.
If this really is in RHS this way, then it makes no difference. But we should make sure that this is a RHS bug and not us introucing it with the compat PBO.

```
    class Heli_Light_02_base_F: Helicopter_Base_H {
        class Turrets: Turrets {
            class CopilotTurret: CopilotTurret {
            };
        };
    };
```

commy2:
We do the same thing to vanilla helis. At this point I'm not sure though if we shouldn't have added this with `addAction` instead (and making a get-out-left and get-out-right option), just because changing this by config is pretty sensitive.

jonpas:
We just enable eject for pilots and copilots, a vanilla option.

optionals/compat_rhs_afrf3/CfgVehicles.hpp
Dystopian: `Heli_Light_02_base_F` does have `MainTurret` in vanilla. It's inherited from `Helicopter_Base_F` -> `Helicopter_Base_H`.

Dystopian:
I like `addAction` idea. IIRC ACE2 had such feature (`Jump out`?). We could remove all these config entries even for vanilla vehicles. Condition to show: if eject is disabled in current seat config.
It's also very important to support hotkey for me. I always eject with hotkey only. BTW `addAction` supports key.
Just tested `player addAction ["eject1", {moveOut player}, nil, 0, true, true, "eject"]` - hotkey works.

optionals/compat_rhs_afrf3/CfgVehicles.hpp
commy2: WRONG
You (or RHS) are introducing 
`CfgVehicles/Heli_Light_02_base_F/Turrets/MainTurret`
which does *not* exist in the base game.
The actual class is:
`CfgVehicles/Helicopter_Base_H/Turrets/MainTurret`
and you have to inherit it from `Helicopter_Base_H`, otherwise you are changing the details of class inheritance. Just because both classes are called `MainTurret` doesn't mean that they are the same.
This comes into play when you create a helicopter inheriting from `Heli_Light_02_base_F` without changing `Heli_Light_02_base_F`'s `Turrets` class. Since turrets are ignored when inherited without reference, in vanilla you would end up with one copilot turret, while with this addon you would end up with a gunner and a copilot. You created a bugged invisible (since most likely not supported by the model) ghost slot.

jonpas:
Doing the `addAction` trick would be awesome, I say go for that!

Dystopian:
`_unit action ["Eject", vehicle _unit]` doesn't work naturally.

`moveOut` has at least 2 problems:

1. it stops engine while vanilla `Eject` doesn't;
1. AI subordinate just really moves out on order while with vanilla unit doesn't until he has a parachute and then moves out.

Should I care about workaround now or wait for responses after release/review?

Dystopian:
Closed in favor of #5384.

PR #4754, opened 2016-12-09
Medical Rewrite - Post Process Effects

ulteq:


addons/medical_ui/functions/fnc_handleEffects.sqf
```diff
@@ -0,0 +1,31 @@
...
+    [false]        call FUNC(effectPain);
+    [false]        call FUNC(effectBloodVolume);
+};
+
+// - Current state info -------------------------------------------------------
+private _bleedingStrength = [ACE_player] call EFUNC(medical,getBloodloss);
+private _bleeding         = _bleedingStrength > 0;
+private _bloodVolume      = ACE_player getVariable [QEGVAR(medical,bloodVolume), DEFAULT_BLOOD_VOLUME];
+private _unconscious      = ACE_player getVariable [QEGVAR(medical,isUnconscious), false];
+private _heartRate        = ACE_player getVariable [QEGVAR(medical,heartRate), 70];
```
ulteq: The default heart rate is 80, I guess we should also put it into `script_macros_medical.hpp`.

addons/medical_ui/functions/fnc_handleEffects.sqf
```diff
@@ -0,0 +1,31 @@
...
+private _pain             = ACE_player getVariable [QEGVAR(medical,pain), 0];
+
+// - Visual effects -----------------------------------------------------------
+[_unconscious, _unconscious] call FUNC(effectUnconscious);
+[true, _pain] call FUNC(effectPain);
+[true, (DEFAULT_BLOOD_VOLUME - _bloodVolume) max 0] call FUNC(effectBloodVolume);
```
ulteq: `linearConversion [0, DEFAULT_BLOOD_VOLUME, _bloodVolume, 0, 1, true]` maybe?

PabstMirror:
Slow heartbeat is fine
But the 3 fast effects that are played randomly each time don't work well together.
Not sure which one is bad, but it might be better to just play the same file each time.

PabstMirror:
Overall I would say pain effects might be too weak compared to before??

I'm not sure how we really want pain to effect the end user.
With previously medical, extreme pain could really hinder being able to see clearly.

One of the 2 options is `RadialBlur`, the problem with this is that it can disabled via video settings.
(I usually run with all the blur effects off, so I thought the ppEffect wasn't working at first).
So it would be impossible to have the same effect on the user.

PR #4762, opened 2016-12-15
Minor wind deflection optimization

PabstMirror:


commy2:
Ha! Nifty. :)
Never expected to see an application for this in Arma.

PR #4763, opened 2016-12-15
add optionals meta.cpp files

bux578:
- Add meta.cpp files to RHS and ADR optionals

Since those mods are now Steam Workshop it just makes sense, that they get their own meta.cpp


commy2:
Hope you didn't mix up Russia and US.

bux578:
Aren't they the same faction? :joy:

I hope not

commy2:
>Aren't they the same faction?
This is WW2

PR #4770, opened 2016-12-15
Overpressure changes

### LINKED ISSUE #4691: overpressure too much on 100mm / 125mm guns & rpgs
**CBA Version:** `latest` (stable)
**ACE3 Version:** ` latest` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`
- `@RHS_AFRF`
- `@acre2`
- `Check RPT for more`

**Description:**
- If you are standing 20 meters near a tank or armored fighting vehicle and this is firing its 100 / 125mm gun then you will get overpressure damage. Even if you are behind the tank. In RealLife you need to be atleast 2 meters at front to the main gun to get that amount of damage.

- Same occurs for RPG Launchers. They overpressure damage is also too high. Since you can even fire that launcher from inside a house (e.g. all the videos from syria where soldiers are firing them inside a house with no damage)

**Steps to reproduce:**
- Place a tank or IFV on the map with advanced medical system active and let someone shot the canon while you are next to the tank (20 meters or closer). I even had a small hill between me and the tank and got damage for overpressure.

- Dedicated Server with 35 players including headless client

**Placed Modules:**
- MicroDAGR MapFill // ACE Medical System Advanced Modules.

**RPT log file:**
- https://gist.github.com/Aebian/aa61d10ce184f64edd49ee73a5565656

PabstMirror:
Close #4691

- Add ace_setting for overpressure distance
- Change overpressure config on 2a70 (used on the bmp-3/bmd-4; it has a very low muzzle velocity)
- fix rhs gl smoke causing frag
- improve debug line drawing

### DIFF
--- a/addons/common/functions/fnc_addLineToDebugDraw.sqf
+++ b/addons/common/functions/fnc_addLineToDebugDraw.sqf
@@ -7,11 +7,17 @@
  * 1: End point ASL <ARRAY>
  * 2: Color <ARRAY>
  *
- * None
+ * Return Value:
+ * Nothing
+ *
+ * Example:
+ * [[0,0,0], [1,1,0], [1,0,0,1]]] call ace_common_fnc_addLineToDebugDraw;
  *
  * Public: No
  */
 #include "script_component.hpp"
+
+params ["_startASL", "_endASL", "_color"];
 
 if (isNil QGVAR(debugLines)) then {
     GVAR(debugLines) = [];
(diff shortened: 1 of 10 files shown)


vayrynen123:
BMP-1 also uses low velocity gun and it basically uses same ammo with SPG-9 so it is rocket assisted and there should not be any overpressure effect.

PR #4779, opened 2016-12-20
Cookoff - add function to get ammo

### LINKED ISSUE #4565: Ammo Vehicles dont cook off
**CBA Version:** `3.8.0` stable
**ACE3 Version:** `3.1.1` stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- None of the ammo vehicles cook off HEMTT, Typhoon, KamAZ, Taru.
- Same for Taru Ammo pod.
- Huron ammo container and vehicle ammo smoke but nothing else

**Steps to reproduce:**
- Place listed items in editor and blow up

- Editor (Singleplayer)

**Placed Modules:**
- Add the list of modules you have placed on the map. Use 'None' if the error occurs without any modules.

**RPT log file:**
- https://gist.github.com/Djzonk/22b8c2baf8ddbf24464189e6d62e50de

PabstMirror:
- Fix #4565 - Function adds virtual ammo for rearm vehicles, also gets ammo from all turrets and cargo
- Remove FUNC(secondaryExplosions) - which duplicated detonateAmunition
- Tweak timeBetweenAmmoDetonation calculation to use all ammo, not just the current mag
- Allow GVAR(enableAmmoCookoff) effect to work with GVAR(enable)=false


addons/cookoff/functions/fnc_getVehicleAmmo.sqf
```diff
@@ -0,0 +1,58 @@
...
+#include "script_component.hpp"
+
+params ["_vehicle"];
+TRACE_1("getVehicleAmmo",_vehicle);
+
+private _ammoToDetonate = [];
+private _totalAmmo = 0;
+
+// Get ammo from turrets
+if (_vehicle isKindOf "AllVehicles") then {
```
commy2: Why `"AllVehicles"`? `"CAManBase"` is `"AllVehicles"` too.

commy2: Why `"AllVehicles"`? `"CAManBase"` is `"AllVehicles"` too.

PabstMirror: good catch, 
I was originally using `magazinesAmmo` for turret ammo and `magazinesAmmoCargo` for cargo ammo
But for ReammoBoxs they returned the same thing, so it was doubling the count.

`magazinesAllTurrets` doesn't have that issue, so we don't need that if statement

addons/cookoff/XEH_postInit.sqf
```diff
@@ -80,11 +80,9 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 // secondary explosions
 ["AllVehicles", "killed", {
     params ["_vehicle"];
-    if (_vehicle getVariable [QGVAR(enable),GVAR(enable)]) then {
```
SilentSpike: :+1: Support the removal of this dependency, can see cases where mission designer wouldn't want vehicle cook off and the damage reduction associated with it, but still want the ammo cook off on kill.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -80,11 +80,9 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
...
-        _vehicle call FUNC(secondaryExplosions);
-        if (_vehicle getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmoCookoff)]) then {
-            [_vehicle, magazinesAmmo _vehicle] call FUNC(detonateAmmunition);
-        };
+    if (_vehicle getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmoCookoff)]) then {
```
SilentSpike: Should we add a second eden attribute to toggle this too?

addons/cookoff/functions/fnc_cookOffBox.sqf
```diff
@@ -44,9 +44,9 @@ if (local _vehicle) then {
 
         // These functions are smart and do all the cooking off work
         if (local _box) then {
-            _box call FUNC(secondaryExplosions);
             if (_box getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmoCookoff)]) then {
```
SilentSpike: Almost feel like this setting dependency should either be removed for boxes, or the `GVAR(enableAmmobox)` setting homogenised with this one (see XEH_postInit.sqf line 74).

addons/cookoff/functions/fnc_detonateAmmunition.sqf
```diff
@@ -27,21 +31,22 @@ private _magazine = _magazines select _magazineIndex;
...
+    _amountOfMagazines = _amountOfMagazines - _removed;
+    if (_amountOfMagazines <= 0) then {
         _magazines deleteAt _magazineIndex;
     } else {
-        _magazine set [1, _newMagCount]; // clear out the magazine
+        _magazine set [1, _amountOfMagazines]; // clear out the magazine
     };
-    private _timeBetweenAmmoDetonation = (random 7) * (1 / random (_amountOfMagazines)) min MAX_TIME_BETWEEN_AMMO_DET;
-    _timeBetweenAmmoDetonation = _timeBetweenAmmoDetonation max 0.1;
+    private _timeBetweenAmmoDetonation = (((random 10) / (sqrt _totalAmmo)) min MAX_TIME_BETWEEN_AMMO_DET) max 0.1;
```
SilentSpike: :+1: Excellent improvement, the inverse proportionality of average delay to the square root of remaining ammo count is a very elegant solution to simulating the decaying intensity of the burn.

Should drastically reduce the length of cook off now that there's far lass chance of the max time being hit with each delay.

addons/cookoff/functions/fnc_cookOffBox.sqf
SilentSpike: Yeah on further thought I'd definitely be in favour of:

1. Removing this condition entirely
2. Changing the line 74 of `XEH_postInit.sqf` to use `getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmobox)])`
3. Adding a second eden setting that applies to boxes and vehicles to toggle their `QGVAR(enableAmmoCookoff)`
4. Making the existing eden setting only apply to vehicles again
5. Clarifying the difference between the global settings `GVAR(enableAmmoCookoff)` and `GVAR(enableAmmobox)` (the former applied to vehicles, the latter applies to boxes specifically).

I personally think that produces the most flexible settings for mission designers to use.

PabstMirror: @SilentSpike - I merged as is, we can look at settings in another pr
Should also look at incerate grenades and make sure it's using the right ace_setting?
https://github.com/acemod/ACE3/blob/master/addons/grenades/functions/fnc_incendiary.sqf#L172

Issue #4283, opened 2016-08-22
Module to assign cargo slots


tdiniz93:
I'd like to suggest a module that allows us to assign cargo slots for addon vehicles, right now the RHS LMTVs have the same cargo capacity as the HEMTT, rather than make a new conifg for every vehicle in the RHS compat, can we just get a module that would allow us to assign slots?

jonpas:
You should be able to use the following code in the vehicle's init to change its cargo size:
```
this setVariable ["ace_cargo_size", 12];
```
Number being the size.

I believe we would prefer an Eden attribute for setting cargo size rather than a module.

tdiniz93:
I have another cargo question that doesn't warrant creating another issue, I used the setCarry function, but it only works on small objects, I want to make a base building mission, and would like to use SetCarry to carry hescos, is there a way to do this?
```

_fnc_makeCarrayble = { 
    params ["_object"]; 
    diag_log text format ["Making %1 [%2] Carrable", _object, typeOf _object]; 
    [_object, true, [0,0.5,0.3], 0] call ace_dragging_fnc_setCarryable; 
};
["Land_HBarrier_01_line_3_green_F", "init", _fnc_makeCarrayble, true, [], true] call CBA_fnc_addClassEventHandler; 

```

jonpas:
Mass of the item also affects carrying, if it's too heavy you can't carry it. I guess you can try modifying the mass of that object.

SilentSpike:
Decided to take a look into adding eden settings for this. Seems that we're doing a hardcoded check against a vehicle classname (for property `ace_cargo_hasCargo`) before adding cargo actions to it - so if we wanted to add a generic cargo space setting for all vehicles that might need to be tweaked slightly.

Pasrules:
`this setVariable ["ace_cargo_size", 20]` does not work on vehicles from addons. Tested on Vanilla (works), BAF, RHS (no change in inventry size). Compatibility for those mods were not used, script was used on vehicle init.

bux:
Original question has been answered, that's why i'm closing this issue, feel free to discuss this further

markooff:
OK guys, but I have another question - does it work with objects like buildings (i.e. structures ) ? 
I'm just working on some sort of more-advanced-logistics module for my Group [PGM] and 
I've prepared some special structures (like small fortifications) which can be dragged/carried/load into the cargo (under some certain circumstances of course ) . 

[...]

PR #4780, opened 2016-12-20
Add cargo eden attributes

SilentSpike:
Inspired by #4283

- Add an `ace_cargo_space` attribute to vehicles to alter how much cargo they can carry
- Add an `ace_cargo_size` attribute to objects to alter how much cargo space they consume
- Add a `ace_cargo_setSize` public function
- Add a `ace_cargo_setSpace` public function
- Deprecate `ace_cargo_makeLoadable` public function and associated module

Known issues in current WIP implementation:
- [x] Public functions do not provide menus for future JIP players
  - [x] This is no different from current behaviour of existing public function, probably because it's a PITA to handle. Could be fixed using a `CBA_fnc_globalEventJIP`, would need to track the ID of which on the object for future function calls to overwrite the event.
- [x] Public functions used more than once on the same object (which doesn't usually have a menu) will add two action menu entries
- [x] Vehicles may have 2 cargo actions in their interaction menu (CBATeam/CBA_A3#567)


addons/cargo/CfgEden.hpp
```diff
@@ -0,0 +1,39 @@
...
+                        typeName = "NUMBER";
+                    };
+                    class GVAR(size) {
+                        displayName = CSTRING(size_edenName);
+                        tooltip = CSTRING(size_edenDesc);
+                        property = QGVAR(size);
+                        control = "Edit";
+
+                        // Expression only runs on the server, must handle actions for all machines and future JIPs (Why BI?!)
+                        expression = QUOTE(_this setVariable [ARR_3('%s',_value,true)]; [QQGVAR(EdenObjectInit),_this] call CBA_fnc_globalEventJIP;);
```
SilentSpike: Just realised there's a missing `ARR_2` here that I must have added on my local version after committing. Rest assured, that's not what was making the `initObject` call fail to work.

SilentSpike:
Good call, it actually looks like this is being handled already in `fnc_canLoadItemIn` so it can be removed from the action conditions :smile:

SilentSpike:
Fixed and tested JIP compatiblity :+1: Working as expected

SilentSpike:
Think this is now ready to go if anyone wants to review it. Assigned commy as he's a listed maintainer and pabst because he showed some interest in this.

commy2:
I like it, could be merged.

SilentSpike:
Mentioned in private slack, but looking at this now I'm not sure I like the way I've handled the locality in the public functions. I believe my intent was to allow the size to be set independently on each machine if desired (for example maybe only a specific player should be able to load the object in question). However it introduces a layer of complexity I'm not sure we want to have.

PabstMirror:
I think size vars should to be the same on all machines

SilentSpike:
Yeah I think it keeps things simpler to have values synced. I'll modify the public functions in this.

PR #4720, opened 2016-11-27
Exclude uniforms with inherited modelSides in No Uniform Restrictions optional

jonpas:
**When merged this pull request will:**
- Exclude uniforms with inherited modelSides in `ace_nouniformrestrictions_fnc_exportConfig`
- Use new dev recompile macro in optional components
- Remove `CBA_DEBUG_SYNCHRONOUS` from optional components
- Properly name debug macros in `nouniformrestrictions` optional component

PR #4782, opened 2016-12-21
Simplify config crawling in nouniformrestrictions

Dystopian:
I know it's not important for performance, but it's a good example script (e.g. for me) and I want it to be more perfect.


jonpas:
Improves #4720

PR #4787, opened 2016-12-26
Generic cycle attack profile key for missile guidance

PabstMirror:
Replace the javelin specific key with one that will work for any missileGuidance weapon.

PR #4791, opened 2016-12-28
NLAW - Predicted Line Of Sight Guidance and Overfly Attack Mode

PabstMirror:
Adds the NLAW's unique guidance system
- Launcher keeps target in sight while holding a tracking key (default Tab)
- Missile's gyroscopes record the change in angle over time
- After firing it flies a path that compensates for both gravity and the change in angle

Overfly Attack Mode
- Automatically aims missile above target
- Short range sensors detect object below and detonates warhead above target
- Payload is a shaped charge that is aimed down which is designed to penetrate the turret top

Note: This PR currently includes #4787 for testing, so diff is a bit larger than it will be.


VKing6:
Should we reduce the direct attack damage a bit more (300-350, maybe) and add some fragmentation to the warhead?

addons/nlaw/CfgAmmo.hpp
```diff
@@ -0,0 +1,53 @@
...
+    };
+    class ACE_NLAW_ShapedCharge: ACE_NLAW { // Shaped charge from rocket explosion, no effects
+        timeToLive = 1;
+        hit = 750;
+        indirectHit = 0;
+        indirectHitRange = 0;
+        explosionSoundEffect = "";
+        explosionEffects = "";
+        CraterEffects = "";
+        muzzleEffect = "";
```
VKing6: Setting `model = "";` would reduce the graphical oddity of suddenly seeing the missile flip and drop straight down.

PabstMirror:
I'm not sure exactly how much to drop the hit
Direct attack mode should still be able to cause major damage to softer targets
https://www.youtube.com/watch?v=byR9EfOTeZg&feature=youtu.be&t=261

VKing6:
It can tenderize trucks and cars pretty well at the moment. I didn't test it against helicopters, though.
I think 300-400 is the sweet spot, so it's not in a bad place. I just feel like 350 for reasons.

One other thing I thought of when I watched that video, though: Can we reduce the weapon sway (particularly when winded) to emulate the monopod thing it has?

VKing6:
I'd say this is pretty much ready, now. The only thing I'd want is for someone else to test the AI skill randomization in [fnc_fired](https://github.com/acemod/ACE3/pull/4791/files#diff-4c72a5ef7d2050f51c9c61ba427a94cdR44) to make sure the low end accuracy isn't unreasonably low.

There is one thing that I'm not sure if is reasonable to consider:
If a vehicle is driving down a straight road and you're a distance from the road somewhere to the side in front of the target tracking it, When you track true on the target and fire, the missile will pass behind the target. If the target's moving away from you, the missile will pass in front of the target.

I'm not sure if this is something that's compensated for in the software in reality (rate of change awareness, maybe?) or if it's a training thing, but it's not that hard to compensate for if you've practised a bit.

docs/wiki/feature/nlaw.md
```diff
@@ -0,0 +1,35 @@
...
+layout: wiki
+title: NLAW
+description: NLAW
+group: feature
+category: equipment
+parent: wiki
+mod: ace
+version:
+  major: 3
+  minor: 9
```
jonpas: 10?

addons/nlaw/functions/fnc_onFired.sqf
```diff
@@ -0,0 +1,63 @@
...
+// Limit Max Deflection
+_yawChange = -10 max _yawChange min 10;
+_pitchChange = -10 max _pitchChange min 10;
+
+TRACE_3("attackProfileStateParams",_firedLOS,_yawChange,_pitchChange);
+_attackProfileStateParams set [0, CBA_missionTime];
+_attackProfileStateParams set [1, _firedLOS];
+_attackProfileStateParams set [2, _yawChange];
+_attackProfileStateParams set [3, _pitchChange];
+
```
jonpas: 2 empty lines.

addons/nlaw/functions/fnc_onFired.sqf
```diff
@@ -0,0 +1,63 @@
+/*
+ * Author: PabstMirror
+ * Sets up missile guidance state arrays (called from missileGuidance's onFired)
```
jonpas: Missing final dot.

addons/nlaw/functions/fnc_attackProfile.sqf
```diff
@@ -0,0 +1,60 @@
+/*
+ * Author: PabstMirror
+ * NLAW missile guidance attack profile
```
jonpas: Missing final dot.

ZeusWrath:
> Launcher keeps target in sight while holding a tracking key (default Tab)

For which this Tab-button was mentioned,  I did not understand. After several tests I saw that NLAW hits the target without Tab. 

Try to create a visualization of a thermal imager for the NLAW also, as we see in Titan Launchers. Without thermal vision NLAW being disposable, it looks too deprived.

ZeusWrath:
Also, with an _**Overfly Attack Mode**_, there is no visualization of the Missile explosion above the tank. I could not see it. Try to add additional FX with the upper explosion, what is the type of this: https://youtu.be/byR9EfOTeZg?t=164

PabstMirror:
There is no traditional seeker like the javelin/titan's. It is just based on physical movement of the launcher prior to firing.

image

ZeusWrath:
Thanks for the info. I also noticed that, now the NLAW becomes much weaker.  It took me 4 tubes to destroy one tank Merkava MK 4 m!  (Overfly Attack Mode)

ZeusWrath:
What about the additional visualization of the explosion from above? Is it possible?

jonpas:
Please keep feature requests to the feature request issue, we won't be answering them on closed or any PRs. Sure it's possible, but that doesn't mean it'll happen.

ZeusWrath:
OK! Good luck!

madpat3:
is the range limited to about 1000 m?

Issue #4797, opened 2017-01-01
Medic AI fails to do his job after executing an unnecessary `doMove`


Skully71:
**Arma 3 Version:** `1.66.139586` (stable)
**CBA Version:** `3.1.2` (stable)
**ACE3 Version:** `3.8.3` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
Medic AI fails to do his job after executing an unnecessary `doMove`.
http://steamcommunity.com/sharedfiles/filedetails/?id=831589390

**Steps to reproduce:**
1. Make a squad with at least 1 Combat Life Saver and a Rifleman.
2. Move the Rifleman such a distance away that the Medic will not have enough time to reach him before the Rifleman heals himself.
3. Shoot the Rifleman in the foot and order him to move to another spot.
4. Observe the Medic going to the original position and then forming back into formation.
5. Medic is now "stuck".

**Where did the issue occur?**
- [ ] Dedicated
- [ ] Self-Hosted Multiplayer
- [ ] Singleplayer
- [x] Editor (Singleplayer)
- [x] Editor (Multiplayer)
- [ ] Virtual Arsenal

**Placed Modules:**
- None

**RPT log file:**
Let me know if you need anything from it. :)

Skully71:
If the injured heals during https://github.com/acemod/ACE3/blob/master/addons/medical_ai/functions/fnc_healUnit.sqf#L42  and has moved on. Then the medic arrives on an empty spot and returns to formation. https://github.com/acemod/ACE3/blob/master/addons/medical_ai/functions/fnc_healUnit.sqf#L31
`QGVAR(movingToInjured)` is however never set to `false`
Forcible setting it to false does not put the medic in a correct state though. So I think there is more to it.
For the most part it solves the bug, there is an edge where the medic goes into `Wait` instead of `Move`. I can't reproduce that.

PR #4800, opened 2017-01-02
Ensure movingToInjured is false whenever _target no longer needs healing

Skully71:
- fix issue #4797


BaerMitUmlaut:
Thanks @Skully71!

PR #4803, opened 2017-01-04
Frag Superanal Cleanup

654wak654:
- Title
- Obviously doesn't live up to #4581, just some smaller things I found


addons/frag/functions/fnc_frago.sqf
```diff
@@ -96,71 +96,71 @@ _fragArcs set [360, 0];
...
             private _target = _x;
             if (alive _target) then {
                 private _boundingBox = boundingBox _target;
+
+                private _cubic = 1;
+                for "_i" from 0 to 2 do {
+                    MUL(_cubic,(abs ((_boundingBox select 0) select _i)) + ((_boundingBox select 1) select _i));
+                };
+
+                if (_cubic <= 1) exitWith {};
```
commy2: Not the best design, but it works since it's nested in a then block and not the forEach block.

addons/frag/functions/fnc_frago.sqf
```diff
@@ -96,71 +96,71 @@ _fragArcs set [360, 0];
...
     {
         //if (random(1) > 0.5) then {
             private _target = _x;
             if (alive _target) then {
                 private _boundingBox = boundingBox _target;
+
+                private _cubic = 1;
+                for "_i" from 0 to 2 do {
+                    MUL(_cubic,(abs ((_boundingBox select 0) select _i)) + ((_boundingBox select 1) select _i));
+                };
```
commy2: This is pretty ugly. Should write the line 3 times like before and not for a for loop with constant from-to's.

commy2:
I like this, because it has perf improvements and frags needs them badly.

PR #4824, opened 2017-01-15
Fix medical typo 'to' -> 'too'

jonpas:
- Title

Not sure if I should have PRed that into medical rewrite branch.


jonpas:
Medical rewrite branch already has major conflicts with the stringtable. And the same typo is in that branch, it can be merged into medical-rewrite directly as well, doesn't matter.

PR #4836, opened 2017-01-17
Add localUnits array and use that inside medical statemachines

dedmen:
This will add an array of localUnits which is kept up-to-date by Eventhandlers and remove the need to dynamically create that list every few frames inside the medical state machines.

The EH has about a performance penalty of 0.015ms per spawned unit.

Currently the Statemachines inside medical_ai and medical_blood create this array of local Units dynamically each time they refresh their list (https://github.com/CBATeam/CBA_A3/blob/master/addons/statemachine/fnc_clockwork.sqf#L37)
by using `allUnits select {local _x}` which has about a 0.24ms impact.

This is basically moving the performance impact from a PFH thats executed every few frames to an Eventhandler and also reducing the performance cost in the meantime.

I didn't test this yet but i don't know what could go wrong. Besides the "local" EH misbehaving.

Currently the Statemachine clockwork for the 2 medical Statemachines is with 1.2ms the longest executing PFH when on a server with exactly 100 non-local AI's.
diag_captureFrame:
diag_captureFrame PFH
And because 0 of these AI's are local and the Statemachine will recreate the unit list when it iterated through it fully. It will recreate it every frame. 
For both Statemachines thats about 0.48ms per frame just for recreating that list over and over again.
This PR should cut down on the bulk of that.

This would essentially be replaced by BI adding a localUnits script function. Which i would hope they would do but.. They probably won't.


PabstMirror:
I think this is a good idea, will really shine with high unit counts.

Need to use retro init - `}, true, [], true] call CBA_fnc_addClassEventHandler;`
Or nothing gets added at start like `player`.
But that's currently bugged - https://github.com/CBATeam/CBA_A3/issues/567

Need to handle deleted units,  `Deleted` EH is a good canidate but it's for 1.68.

dedmen:
A quick workaround against CBATeam/CBA_A3#567 would probably just be using pushBackUnique
What about adding the EH's at preInit? That was actually what commy suggested. preInit should be before unit spawns right?

BaerMitUmlaut:
@pabstmirror: deleted units should become objectNull, right? The state machine skips null elements.

PabstMirror:
Think of something like ALIVE that does unit caching
Constantly creating and deleting units
The list would be constantly growing
We would waste more time filtering though a list of a 1000 objNulls

dedmen:
Then we should rather wait till 1.68 with this.
For the EH's there are three possibilities

- Cram all the code inside the config EH
- Create a seperate function for each EH and call that from the config
- Add in preInit.

I do favor the preInit one if thats reliable enough. What do you think?

commy2:
Solution for deleted units is using the "deleted" event handler (requires 1.68)
Solution for killed units is using the "killed" event handler.

I think these two should cover these potential problems.

commy2:
Retro init is not needed if the "init" event handler is installed in preInit. (Which should be the way to go anyways).

commy2:
Since the state machine is filtering null objects anyway, I think we can just add the deleted event handler and play the waiting game with the Arma update. CBA will update soon, hopefully.

PabstMirror:
Added handler for respawn.
Also for Killed - I think this is what we want `allUnits` seems to eventually filter out dead units after some delay (oddly not always the same on all machines)

Also check out https://github.com/CBATeam/CBA_A3/pull/671
I think medical_blood was slowest because the actual update code is:
```
allUnits select {(local _x) && {[_x] call FUNC(isBleeding)}}
```
With a lot of AI that call would be expensive but with no fighting it would return `[]` 
So it would be re-checked every frame.

dedmen:
As I look at this code again. It's probably not a bad Idea to replace the `pushBack`s with `pushBackUnique` It shouldn't go wrong. But just for good measure. Maybe.. Just an idea.

PabstMirror:
Unit creation and deletion works fine with this, 
but there seems to be a critical problem with vehicle crews

Placing a vehicle via zeus will produce a debug like
```
[ACE] (common) TRACE: 55543 unit init: _unit=B Alpha 1-6:1, local _unit=true z\ace\addons\common\functions\fnc_setupLocalUnitsHandler.sqf:27
[ACE] (common) TRACE: 55543 unit deleted: _unit=B Alpha 1-6:1, local _unit=true z\ace\addons\common\functions\fnc_setupLocalUnitsHandler.sqf:59
[ACE] (common) TRACE: 55543 unit init: _unit=B Alpha 1-6:2, local _unit=true z\ace\addons\common\functions\fnc_setupLocalUnitsHandler.sqf:27
[ACE] (common) TRACE: 55543 unit deleted: _unit=B Alpha 1-6:2, local _unit=true z\ace\addons\common\functions\fnc_setupLocalUnitsHandler.sqf:59
```
the deleted EH runs for some reason, even though the crew is clearly NOT deleted.

This happens with both zeus and `createVehicleCrew`

PabstMirror:
```
["CAManBase", "init", {
    params ["_unit"];
    systemChat format ["Init: %1 - %2", diag_frameNo, typeOf _unit];
    _unit addEventHandler ["Deleted", {
        systemChat format ["AEH Deleted: %1 - %2", diag_frameNo, typeOf _unit];
    }];
}, true, []] call CBA_fnc_addClassEventHandler;

["CAManBase", "deleted", {
    params ["_unit"];
    systemChat format ["XEH Deleted: %1 - %2", diag_frameNo, typeOf _unit];
}, true, []] call CBA_fnc_addClassEventHandler;
```
On placement, BOTH deleted EHs run, and they both run on real deletion
I guess it's a bis bug

dedmen:
We can CBA nextFrame and then check isNull on the unit. I guess.
Can we push parameters through nextFrame? If it is really deleted then isNull will be true.

dedmen:
Okey can confirm.
And my isNull on execNextFrame Idea seems to work. When placing vehicle with crew in Zeus the deleted EH is called. But on nextFrame the _unit is not null so we skip deleting it then.

PabstMirror:
image

It's failing when deleting via zeus
```
// Debug code:
[{isNull _this}, {diag_log text format ["%1 - %2 became null", diag_frameNo, this]}, cursorObject] call CBA_fnc_waitUntilAndExecute;

// Deleting via deleteVehicle
45596 unit deleted: _unit=10a29ec40# 164025: b_soldier_01.p3d, local _unit=true z\ace\addons\common\functions\fnc_setupLocalUnitsHandler.sqf:56
45597 unit deleted nextFrame: _unit=<NULL-object>, local _unit=false, isNull _unit=true z\ace\addons\common\functions\fnc_setupLocalUnitsHandler.sqf:61
45597 - any became null

// Deleting via zeus
53311 unit deleted: _unit=10a2bef40# 164031: b_soldier_01.p3d, local _unit=true z\ace\addons\common\functions\fnc_setupLocalUnitsHandler.sqf:56
Info: Zeus interface queued deletion of Alpha 2-2 group with 0 units inside.
53312 unit deleted nextFrame: _unit=10a2bef40# 164031: b_soldier_01.p3d, local _unit=true, isNull _unit=false z\ace\addons\common\functions\fnc_setupLocalUnitsHandler.sqf:61
53313 - any became null
```

So it takes 2+ frames for it to actually be deleted?

dedmen:
🤕 I hate Arma.

dedmen:
In the end.. Can't we just make a getLocalUnits function.
That does `localUnits = localUnits - [objNull]` on the first call every frame?
Perf wise it's crap. But still a lot better than everyone doing `allUnits select {local _x}`

addons/common/functions/fnc_setupLocalUnitsHandler.sqf
```diff
@@ -0,0 +1,76 @@
...
+["CAManBase", "deleted", {
+    params ["_unit"];
+    TRACE_2("unit deleted",_unit,local _unit);
+
+    if (local _unit) then {
+        [{
+            params ["_unit"];
+            TRACE_3("unit deleted nextFrame",_unit,local _unit,isNull _unit);
+            if (isNull _unit) then { //If it is not null then the deleted EH was Fake.
+                GVAR(localUnits) = GVAR(localUnits) - [objNull];
```
dedmen: I delete all objNull here as using `find` to find a null object will find the first null object and not the one the deleted EH belongs to.
I would prefer to just `call FUNC(getLocalUnits)` here in case we add a check to getLocalUnits so it only removes null objects once per frame.
But it would hurt readability.

addons/common/functions/fnc_getLocalUnits.sqf
```diff
@@ -0,0 +1,22 @@
...
+ * Array of local Units <ARRAY>
+ *
+ * Example:
+ * [] call ace_common_fnc_getLocalUnits
+ *
+ * Public: Yes
+ */
+
+//Remove null objects
+GVAR(localUnits) = GVAR(localUnits) - [objNull];
```
dedmen: We can check diag_frameNo against a global variable so we only remove null objects once a frame to get us a little more perf.
But I'm not sure if `GVAR(localUnits) = GVAR(localUnits) - [objNull];` is really that expensive that it's worth it. Needs profiling with atleast a hundred local units.

addons/common/functions/fnc_setupLocalUnitsHandler.sqf
```diff
@@ -0,0 +1,76 @@
...
+    params ["_unit"];
+    TRACE_2("unit init",_unit,local _unit);
+
+    if (local _unit) then {
+        if (!alive _unit) exitWith {};
+        GVAR(localUnits) pushBack _unit;
+    };
+}] call CBA_fnc_addClassEventHandler;
+
+["CAManBase", "respawn", {
```
dedmen: Shouldn't "init" also fire for respawning units?
But I guess they might be !alive at init? 

Respawning in MP should be tested.

commy2: Init does not fire for respawned objects.

addons/common/functions/fnc_getLocalUnits.sqf
dedmen: Actually we only need to run this if there was a deletedEH in the last 2-3 frames. But I don't think it hurts.
Considering this will already be about a 100x improvement vs iterating through allUnits everytime.

dedmen: also to checking `diag_frameNo` objects are deleted in simulation cycle. Draw3D runs after that in the same frame. So that could cause Draw3D to still get null objects.

dedmen:
What's currently the problem? Besides merge conflict.
The problem with Zeus deleting units that are not deleted could be solved by using a draw3D EH 1 frame after Zeus "deletes" the unit. We don't have a execNextFrame utility function for that though. And the EH fires in simulation cycle. Meaning right after that is a draw3D EH which we don't want. So if we add in the sim cycle we first have to skip one run.

As vehicles are deleted in SIM cycle this means after using Draw3D it's "impossible" to get null objects into the localUnits array. Because we clean up right after they get deleted.
But Draw3D handlers that run before our's could still see null objects. 
Others already said null-objects aren't a big problem. As they can appear in allUnits anyway.

PabstMirror:
Moved `isBleeding` check to inside state machine as checking `FUNC(isBleeding)` for all units was still expensive. Even with high unit count I think the delay between bleeding is high enough that this should not cause issues.

image

dedmen:
What about the issues we talked about?

PabstMirror:
I tested on a dedicated server and it and
`call ace_common_fnc_getLocalUnits` matched `allUnits select {local _x}`

The addition of `GVAR(localUnits) = GVAR(localUnits) - [objNull];` in the getter fixes any problems with deletedEH. I guess we could even remove the deletedEH as it still is unreliable with zeus

PabstMirror:
```
["CAManBase", "deleted", {
    params ["_unit"];
    TRACE_2("AAA unit deleted",_unit,local _unit);

    if (local _unit) then {
        [{
            params ["_unit"];
            TRACE_3("BBB unit deleted nextFrame",_unit,local _unit,isNull _unit);
            [{
                params ["_unit"];
                TRACE_3("CCC unit deleted next-nextFrame",_unit,local _unit,isNull _unit);
                if (isNull _unit) then { // If it is not null then the deleted EH was Fake.
                    GVAR(localUnits) = GVAR(localUnits) - [objNull];
                };
            }, _this] call CBA_fnc_execNextFrame;
        }, [_unit]] call CBA_fnc_execNextFrame;
    };
```
this should also work

PR #4842, opened 2017-01-20
Add ace_setting for scaling cookoff effect length

PabstMirror:


bux:
What exactly does this?
Could this be a setting?

SilentSpike:
This line determines how many magazines cook off in the current frame.

Currently it's a random int between 1 and 6, this makes the maximum possible able to be increased via an object variable.

I'm not sure it's the most intuitive setting, might be better served by a scaling factor so we could abstract it to the "ammunition burn rate"

PabstMirror:
Yeah, I just wanted a way to scale down the effect length
Maybe it should be a real ace_setting

addons/cookoff/functions/fnc_detonateAmmunition.sqf
```diff
@@ -31,7 +31,7 @@ private _magazine = _magazines select _magazineIndex;
 _magazine params ["_magazineClassname", "_amountOfMagazines"];
 
 if (_amountOfMagazines > 0) exitWith {
-    private _removed = _amountOfMagazines min floor(1 + random(missionNamespace getVariable [QGVAR(ammoScale), 6]));
+    private _removed = _amountOfMagazines min floor(1 + random(6 / GVAR(ammoCookoffDuration)));
```
SilentSpike: Only problem I have with this is that it's a divisor not a multiplier as suggested by the localised description string

addons/cookoff/XEH_postInit.sqf
```diff
@@ -81,6 +81,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 ["AllVehicles", "killed", {
     params ["_vehicle"];
     if (_vehicle getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmoCookoff)]) then {
+        if (GVAR(ammoCookoffDuration) == 0) exitWith {};
```
SilentSpike: I almost feel like we should just replace the `enableAmmoCookoff` setting with this new duration setting (can just be set to 0 for disabled ammo cook off and provides more control).

PabstMirror: Since that setting existed in 3.8, I would like to keep for now.
But when we switch to CBA settings, they could be merged.
I changed string table to indicate it.

PR #4844, opened 2017-01-20
New Function to Show BIS Style Curator Messages

654wak654:
- Add a function for zeus messages and change all displayTextStructured calls to it. New function uses `BIS_fnc_showCuratorFeedbackMessage` for zeus messages, which shows a slightly fancier message on the top bar of the zeus ui.

### DIFF
--- a/addons/zeus/functions/fnc_showMessage.sqf
+++ b/addons/zeus/functions/fnc_showMessage.sqf
@@ -4,19 +4,21 @@
  * If multiple args are given, they get formatted.
  *
  * Arguments:
- * The message <STRING>...
+ * 0: The message <STRING>
+ * N: Anything <ANY><OPTIONAL>
  *
  * Return Value:
  * Nothing
  *
  * Example:
  * ["something"] call ace_zeus_fnc_showMessage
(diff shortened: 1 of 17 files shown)


BaerMitUmlaut:
The variant with format is not documented and due to the params probably also not functional.

654wak654:
> The variant with format is not documented and due to the params probably also not functional.
Oh yes, will add that. It is functional, common's displayTextStructured works the same way.

addons/zeus/functions/fnc_showMessage.sqf
```diff
@@ -0,0 +1,30 @@
+/*
+ * Author: 654wak654
+ * Shows a zeus message through the BIS function, handles localization
```
jonpas: Missing full stop.

654wak654: Most headers doesn't have dots there?

jonpas: They should. :)

BaerMitUmlaut:
You then take an array within an array as a parameter, without a second parameter. I think just passing what you would pass `format` might be a little nicer.

PR #4847, opened 2017-01-21
Cache Nametags onDraw3D flags.

dedmen:
This will cache the Nametag Flags and _maxDistance for 2 seconds instead of reparsing them every Frame.
This will take about 0.0621ms of load from eachFrame and cache it.
After caching the load will be about 0.0323ms per frame.
0.03ms improvement call me crazy if you want.
This will also remove the unused privates from fnc_onDraw3d.

This will create lag from changing Nametag settings to them being visible. Sadly cachedCall can only use 1 Event and I used `cba_events_visionModeEvent` to be also able to cache the _maxDistance variable.
Calculating the _maxDistance variable takes 0.0267 ms so it's very worthwhile to be cached.

### DIFF
--- a/addons/nametags/functions/fnc_getCachedFlags.sqf
+++ b/addons/nametags/functions/fnc_getCachedFlags.sqf
@@ -1,0 +1,64 @@
+/*
+ * Author: <N/A>
+ * Get's flags used for onDraw3D that can be cached
+ *
+ * Arguments:
+ * None
+ *
+ * Return Value:
+ * [_drawName,_drawRank,_enabledTagsNearby,_enabledTagsCursor,_onKeyPressAlphaMax,_maxDistance]
+ *
+ * Example:
+ * call ace_nametags_fnc_getCachedFlags
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+
+// Determine flags from current settings
+private _drawName = true;
+private _enabledTagsNearby = false;
+private _enabledTagsCursor = false;
+private _onKeyPressAlphaMax = 1;
+
+switch (GVAR(showPlayerNames)) do {
+    case 0: {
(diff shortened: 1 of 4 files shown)


addons/nametags/functions/fnc_getCachedFlags.sqf
```diff
@@ -0,0 +1,64 @@
+/*
+ * Author: <N/A>
```
bux: who's that?

commy2:
`cba_events_visionModeEvent`
That is only meant as an internal, but I can see how it really helps here. I don't mind, but technically it's not API and could potentially be nuked at some point either by accident, or by me being dumb and forgetting about this.

dedmen:
I left the author from fnc_onDraw3D becaus I only copied stuff. I also
don't know who that N/A guy is.. He likes negative arrays thats for sure.
Well commy.. Then better make it official.. If its there people gonna use
it.

commy2:
Can't avoid people using internals, but I can shift the blame on them if something breaks in good conscience. Doesn't help the end user, but I can sleep well at night.

addons/nametags/functions/fnc_getCachedFlags.sqf
dedmen: https://github.com/acemod/ACE3/blob/master/addons/nametags/functions/fnc_onDraw3d.sqf#L2 That guy

PabstMirror:
@dedmen - looks good, I pushed some minor changes:
- Can't cache the keybind alpha as it's designed to fade over time
- Added manual cache reseting for both vision mode and ace_setting changes
- Fixed cursorOnly mode with soundWaves to 2 - this was present in master, not related to this PR.

PR #4848, opened 2017-01-21
Improve cachedCall

dedmen:
- Remove the need for one getVariable call.
- Add the possibility to specify more than 1 Event to listen to.

The multi-event thingy will make it possible to Improve #4847 
One could also optimize out the setVariable. Because we are getting the Array as reference anyway so we could use 2x set's on the array instead of one setVariable. But I don't think that will provide a performance benefit.

### DIFF
--- a/addons/common/functions/fnc_cachedCall.sqf
+++ b/addons/common/functions/fnc_cachedCall.sqf
@@ -22,7 +22,7 @@ params ["_params", "_function", "_namespace", "_uid", "_duration", "_event"];
 private _cacheEntry = (_namespace getVariable [_uid, [-99999]]);
 
 if (_cacheEntry select 0 < diag_tickTime) then {
-	_cacheEntry = [diag_tickTime + _duration, _params call _function];
+    _cacheEntry = [diag_tickTime + _duration, _params call _function];
     _namespace setVariable [_uid, _cacheEntry];
 
     // Does the cache needs to be cleared on an event?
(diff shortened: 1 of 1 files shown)


PabstMirror:
@dedmen - The old `getVariable` way seems to be faster. 
Old (0.144071 ms):
```
if ((_namespace getVariable [_uid, [-99999]]) select 0 < diag_tickTime) then {
    _namespace setVariable [_uid, [diag_tickTime + _duration, _params call _function]];
};
(_namespace getVariable _uid) select 1
```
New (0.15108 ms):
```
private _cacheEntry = (_namespace getVariable [_uid, [-99999]]);
if (_cacheEntry select 0 < diag_tickTime) then {
    _cacheEntry = [diag_tickTime + _duration, _params call _function];
    _namespace setVariable [_uid, _cacheEntry];
};
_cacheEntry select 1
```
New way _should_ be better, but it's arma - might be because it's ~2x as many lines.

Multi-event caching doesn't currently work, `_cacheList` needs to be done for each event, see: https://github.com/acemod/ACE3/commit/40e3f3f5aba517132b1eba5d1f524d60ec4bd8e0

But this also has a downside in that the typeChecking and forEach loop add costs to every cachedCall.
Going from 0.225479 ms to 0.267523 ms.

I think I see where you are going with the other PR.
Keep in mind we can always manually reset the cache with our own events that we add manually, e.g:
```
["visionModeEvent", {
    ACE_player setVariable [QGVAR(flagsCache), nil]; //clear cache
```
Just in general, doing our own events like that should be faster because there is no need to do a lookup and go through an array.
Do you think that would work for the nametags improvements?

dedmen:
Yeah it could be slower when the cache is rebuild because of that extra line of code.
But generally if it is cached already, assigning a local variable and reading that should be faster than calling a binary function.
should...
In my optimization i didn't really care about performance when the cache is written. I tried to optimize reading from the cache because thats what will mostly be done.

PabstMirror:
Tested with a 20:1 ratio of cached:rebuild calls, the old way is still faster.

dedmen:
Was aimed as a small performance improvement. But It failed obviously.

PR #4849, opened 2017-01-22
fix win newline in repo

Dystopian:
win newline breaks work with repo when git config core.autocrlf == true


jonpas:
> win newline breaks work with repo when git config core.autocrlf == true

It was due to GitHub Online Editor, it apparently doesn't respect `.gitattributes` for line endings. I'll look into reporting that. `core.autocrlf` has no impact on our repository setup.

PR #4855, opened 2017-01-24
Use CBA Accessory Functions for Laser mode switching

jonpas:
- Replaces ACE's laserpointer switching with CBA's accessory functions


jonpas:
> We do lose the setting check that prevented switching to a mode that wouldn't work.

You are right, this basically allows you to switch but it will just switch to a mode that won't do anything. Not sure what the best thing to do is to make it transparent to the user.

PabstMirror:
What about adding an event in the CBA handler?
We could listen for the event and handle it there

jonpas:
An event sounds good.

PR #4857, opened 2017-01-28
Fix Javelin for AI

### LINKED ISSUE #4856: ACE3 AI not firing titan or RHS Javelin
**CBA Version:** `v3.1.2.161105` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- AI are unable to use the RHS Javelin or the Vanilla Titan when ace is running.  Currently, they will acquire the target, switch to the launcher and then remain forever stuck in a staring contest with said vehicle.  I've seen then do this in the past, so something appears to have been broken lately.

**Steps to reproduce:**
- Run the editor, only ACE and CBA, place titan soldiers or RHS javelin solders and an opfor tank (unarmed)

- both in missions and in the editor

**Placed Modules:**
- Issue occurs when no modules are placed as well as when they all are

**RPT log file:**
- No mention in the logs of anything to do with this issue.

PabstMirror:
Fix #4856

Ammo seems to need this config or AI just take their launcher out and then put them away.
Missile is still guided by script

### DIFF
--- a/addons/missileguidance/CfgAmmo.hpp
+++ b/addons/missileguidance/CfgAmmo.hpp
@@ -69,7 +69,7 @@ class CfgAmmo {
     class M_Titan_AT: MissileBase {};
 
     class ACE_Javelin_FGM148: M_Titan_AT {
-        irLock = 0;
+        irLock = 1;
         laserLock = 0;
         airLock = 0;
 


kendo01:
I must be doing something wrong, but the issue seems to still be there in 3.9.  Can I just get confirmation that this now works for everyone else before I go digging for a different bug?  Same process: editor + Jav/Titan + tank = nothing.

bux:
I've removed it from the github change log

> - AI were not properly taught on how to use the Titan or a Javelin (#4856)

PR #4859, opened 2017-01-29
Optimize Laserpointer

commy2:
- title
- removes the 10 simulated units limit

todo:
- ~~will have to process the player every frame anyway to remove delay when turning on the laser with units nearby (responsiveness)~~
- ~~clearing the array at once means that the laser point is blinking atm. (bug)~~

### DIFF
--- a/addons/laserpointer/XEH_postInit.sqf
+++ b/addons/laserpointer/XEH_postInit.sqf
@@ -7,32 +7,54 @@
 if (!hasInterface) exitWith {};
 
 GVAR(nearUnits) = [];
+GVAR(index) = -1;
+GVAR(laserClassesCache) = [] call CBA_fnc_createNamespace;
+GVAR(redLaserUnits) = [];
+GVAR(greenLaserUnits) = [];
 
 #include "initKeybinds.sqf"
 
 ["ace_settingsInitialized", {
-    //If not enabled, dont't add draw eventhandler or PFEH (for performance)
-    if (!GVAR(enabled)) exitWith {};
+    // if not enabled, dont't add draw eventhandler or PFEH (for performance)
+    if !(GVAR(enabled)) exitWith {};
 
-    // @todo. Maybe move to common?
     [{
-        // handle RHS / bugged vehicle slots
-        private _camPosAGL = positionCameraToWorld [0,0,0];
-        if !((_camPosAGL select 0) isEqualType 0) exitWith {};
+        GVAR(nearUnits) = call FUNC(getNearUnits);
+    }, 5, []] call CBA_fnc_addPerFrameHandler;
 
-        private _nearUnits = [];
-        {
-            _nearUnits append crew _x;
-            if (count _nearUnits > 10) exitWith {
-                _nearUnits resize 10;
+    // custom scheduler
+    [{
+        GVAR(index) = GVAR(index) + 1;
+        private _unit = GVAR(nearUnits) param [GVAR(index), objNull];
+
+        if (isNull _unit) exitWith {
+            GVAR(index) = -1;
+            GVAR(redLaserUnits) = [];
+            GVAR(greenLaserUnits) = [];
+        };
+
+        private _weapon = currentWeapon _unit;
+        private _laser = (_unit weaponAccessories _weapon) select 1;
+
+        if (_laser != "") then {
+            private _laserID = GVAR(laserClassesCache) getVariable _laser;
+
+            if (isNil "_laserID") then {
+                _laserID = getNumber (configFile >> "CfgWeapons" >> _laser >> "ACE_laserpointer");
+                GVAR(laserClassesCache) setVariable [_laser, _laserID];
             };
-        } forEach nearestObjects [_camPosAGL, ["AllVehicles"], 50]; // when moving this, search also for units inside vehicles. currently breaks the laser in FFV
 
-        GVAR(nearUnits) = _nearUnits;
+            if (_unit isFlashlightOn _weapon) then {
+                if (_laserID isEqualTo 1) exitWith {
+                    GVAR(redLaserUnits) pushBackUnique _unit;
+                };
 
-    } , 5, []] call CBA_fnc_addPerFrameHandler;
+                if (_laserID isEqualTo 2) exitWith {
+                    GVAR(greenLaserUnits) pushBackUnique _unit;
+                };
+            };
+        };
+    }, 0.1, []] call CBA_fnc_addPerFrameHandler;
 
-    addMissionEventHandler ["Draw3D", {
-        call FUNC(onDraw);
-    }];
+    addMissionEventHandler ["Draw3D", FUNC(onDraw)];
 }] call CBA_fnc_addEventHandler;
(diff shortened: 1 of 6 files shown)


addons/laserpointer/XEH_postInit.sqf
```diff
@@ -7,32 +7,54 @@
...
+                    GVAR(greenLaserUnits) pushBackUnique _unit;
+                };
+            };
+        };
+    }, 0.1, []] call CBA_fnc_addPerFrameHandler;
 
-    addMissionEventHandler ["Draw3D", {
-        call FUNC(onDraw);
-    }];
+    addMissionEventHandler ["Draw3D", FUNC(onDraw)];
```
dedmen: Add GVAR please. Although it may not be needed. Users may still want to remove the EH.

addons/laserpointer/functions/fnc_onDraw.sqf
```diff
@@ -14,28 +14,20 @@
...
-        };
-
-        if (_laserID > 0 && {_x isFlashlightOn _weapon}) then {
-            [_x, 100, (_laserID == 2 || _isIR), _brightness] call FUNC(drawLaserpoint);
-        };
-    };
-    false
-} count GVAR(nearUnits);
+// no lasers in thermal mode
+if !(_isIR isEqualTo 2) then {
```
dedmen: instead of checking on every frame use CBA visionModeChanged EH and add onDraw EH dynamically?

commy2: CBA player events only work for the local player. I don't see a point to have network traffic here. And AI's wouldn't be able to support it either, even if that's minor.

addons/laserpointer/XEH_postInit.sqf
```diff
@@ -7,32 +7,54 @@
...
+        if (isNull _unit) exitWith {
+            GVAR(index) = -1;
+            GVAR(redLaserUnits) = [];
+            GVAR(greenLaserUnits) = [];
+        };
+
+        private _weapon = currentWeapon _unit;
+        private _laser = (_unit weaponAccessories _weapon) select 1;
+
+        if (_laser != "") then {
```
dedmen: I prefer exitWith to keep indentation a little lower. Just dropping my thoughts.

addons/laserpointer/XEH_postInit.sqf
commy2: Example?

dedmen: `GVAR(onDrawEH) =`
Like i wanted to do right now because of the performance Problem. The missionmaker knows if any player can have a laserpointer right now. If no one has one he can disable the EH and readd it once its possible for players to have a laserpointer.

addons/laserpointer/functions/fnc_onDraw.sqf
dedmen: This is local onDraw. This is only local player. ACE_Player is the same as CBA_Player (or however the CBA one is called) and we only need ACE_Player's vision mode here.

commy2: Yeah, I see what you mean. I was thinking of turning on the laser, not turning on the NVG. Good idea.

addons/laserpointer/XEH_postInit.sqf
commy2: You can just turn off the setting. I don't want people to mess with the internals in that way.

addons/laserpointer/functions/fnc_onDraw.sqf
dedmen: Also there is no reason for the PFH to run if player doesn't have a NVG.. theoretically.

dedmen: Thought error ^^ 
ACE lasers work without NVG anyway. It's only thermal thats disabled.. But still.. could disable the onDraw when Thermal is activated.

addons/laserpointer/XEH_postInit.sqf
commy2: done

addons/laserpointer/XEH_preInit.sqf
```diff
@@ -6,4 +6,11 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+["visionMode", {
```
dedmen: I was intending this to fully disable the draw EH. Because the if statement inside the draw EH is what probably costs performance.

dedmen: `if ((currentVisionMode player) == 2) exitWith {};` //0.0048
`if (visionMode == 2) exitWith {};` //0.0042 
There you go. Moving the variable outside doesn't do much

commy2: Isn't this negligible either way? I'd hate to use `removeEventhandler` for this.

dedmen:
Perf data
old:
- 5 AI's without laserpointer 0.823ms 1.105ms 0.870ms
- 5 AI's with laserpointer but not enabled 1.235ms 1.053ms 0.909ms

commy2:
>Error Type Any, expected String
maybe:
```sqf
private _laser = [_unit weaponAccessories _weapon] param [1, ""];
```

I like testing AllVehicles, because `append`'ing the `crew` is a nice way to handle units inside vehicles. But maybe filtering the animals is preferred as that would happen every five seconds only and decrease the size of the array to walk over...

```sqf
_nearUnits select {isPlayer _x}
```
maybe:
```sqf
_nearUnits select {_x isKindOf "CAManBase"}
```
The range is only 50m, not 500m, because lasers aren't that far visible, so it shouldn't pick up that many animals.

PS: You can make double posts here and it would be easier to follow

PPS:
>Thats a Pokemon
I know. That's like the only plushy I ever owned (inherited).

dedmen:
I don't know how the performance of isKindOf is.. We don't want a horrible fps drop every 5 seconds if standing in a base at the start of a huge mission.
The param way could be a solution but It's putting load into a often called unscheduled function so.... rather not.

Biggest problem on the isPlayer check is that it ignored AI's... Also I don't know if a Zeus Remote controlled unit `isPlayer`.. So your `isKindOf` solution is probably the best

commy2:
I will have to use the param method, because we also need to handle deleted units.
```sqf
objNull weaponAccessories ""
-> nil
```

addons/laserpointer/functions/fnc_getNearUnits.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * None
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+private _camPosAGL = positionCameraToWorld [0, 0, 0];
+
+// handle RHS / bugged vehicle slots
+if !((_camPosAGL select 0) isEqualType 0) exitWith {};
```
PabstMirror: Would this work any better? `isEqualTypeArray [0,0,0]`

commy2: I guess. I have an (probably irrational) disdain for `isEqualTypeArray` and `isEqualTypeParams`, because of the strange ways they handle nil and different sized arrays.

addons/laserpointer/XEH_postInit.sqf
dedmen: You will need it anyway with CBA settings now. As the setting can change mid-game.

PR #4860, opened 2017-01-29
Use CBA ACC EH to prevent switching to unusable lzr

PabstMirror:
Requires https://github.com/CBATeam/CBA_A3/pull/582


jonpas:
I was thinking it would be cleaner to just switch back by directly calling the switch function. Then we don't have to handle it ourselves at all.

`[1, "prev"] call CBA_accessory_fnc_switchAttachment;`

PabstMirror:
Done,
also fixed `gunLightOff` action

jonpas:
Nice!

Can merge into my branch then.

PR #4866, opened 2017-02-01
fix script error caused by new arguments passed to handleDamage

### LINKED ISSUE #4863: Medical HandleDamage error on dev branch
**Arma 3 Version:** `1.67` (dev)
**CBA Version:** `3.1.2` (stable + ed1feb3c)
**ACE3 Version:** `3.8.3` (stable + 0df98f79)

**Mods:**
- `@CBA_A3`
- `@ace`
- 56 other mods that I've tested without

**Description:**
- Whenever a player or AI gets shot in the vest or helmet the game throws an error and they take no damage. The error stays on screen until the entity dies. This just started happening within the past 24 hours. My friend blames the recent update to dev branch by BI.

**Steps to reproduce:**
- Load game. Go to virtual arsenal. Equip a weapon and vest. Click "try". Shoot VR entities in the vest. Note that this happens outside the virtual arsenal as well.

- Dedicated + Editor (Multiplayer) + Virtual Arsenal

**Placed Modules:**
- None

**RPT log file:**
- https://gist.github.com/Drift91/f550607bc56ebd8db82c2b9c7455f270
- "Debug To Clipboard" seems to have done nothing.

commy2:
- title
- fixes #4863

It's just a bandaid and shouldn't be an issue in the first place in the medical-rewrite branch.

### DIFF
--- a/addons/medical/functions/fnc_handleDamage.sqf
+++ b/addons/medical/functions/fnc_handleDamage.sqf
@@ -18,7 +18,7 @@
  */
 #include "script_component.hpp"
 
-_this = _this select [0, 6];
+_this = _this select [0, 7];
 params ["_unit", "_selection", "_damage", "_shooter", "_projectile", "_hitPointIndex"];
 TRACE_5("ACE_DEBUG: HandleDamage Called",_unit, _selection, _damage, _shooter, _projectile);
 
--- a/addons/medical/functions/fnc_handleDamage.sqf
+++ b/addons/medical/functions/fnc_handleDamage.sqf
@@ -18,7 +18,7 @@
  */
 #include "script_component.hpp"
 
-_this resize 6;
+_this = _this select [0, 6];
 params ["_unit", "_selection", "_damage", "_shooter", "_projectile", "_hitPointIndex"];
 TRACE_5("ACE_DEBUG: HandleDamage Called",_unit, _selection, _damage, _shooter, _projectile);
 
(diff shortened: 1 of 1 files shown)


PabstMirror:
One odd thing to keep in mind is that all `handleDamage` EH share the same input array.
So if another EH runs after ace's it would only get 6 elements as well.

But I doubt this would be a problem.

commy2:
Is that true? I somehow doubt that they share the same array reference.

PabstMirror:
```
player addEventHandler ["HandleDamage", {
    diag_log text format ["HDA: Count %1, %2", count _this, _this]; 
    _this resize 1; diag_log "resized to 1";
}];
player addEventHandler ["HandleDamage", {
    diag_log text format ["HDB: Count %1 - %2", count _this, _this];
}];
```
```
HDA: Count 7, [B Alpha 1-1:1 (Pabst Mirror),"body",0,<NULL-object>,"ace_frag_tiny_HD",6,<NULL-object>]
"resized to 1"
HDB: Count 1 - [B Alpha 1-1:1 (Pabst Mirror)]
```

commy2:
Yikes. Do we care about that? Don't think other handleDamage are supported anyway when using ACE.

PabstMirror:
I'm guessing they might be adding this for their medical thing from the campaign?
So the 2 should never be running at the same time.

addons/medical/functions/fnc_handleDamage.sqf
```diff
@@ -18,6 +18,7 @@
  */
 #include "script_component.hpp"
 
+_this resize 6;
```
commy2: Alternative:
```sqf
_this = _this select [0, 6];
```

commy2:
I changed it.

PR #4871, opened 2017-02-04
Cargo load menu overhaul

Dystopian:
- add submenu with vehicles to cargo load menu
- rewrite **Load** menu action and add condition
- clean code

load_submenu

Vehicles with alive crew are suffixed with its name (in brackets).

There are also some questions to resolve:
- should statement for Load menu be deleted? ATM it works as usually - loads to the nearest vehicle, while submenu items load to the specified vehicle;
- should Load menu be renamed to "Load to" e.g.? If yes statement should be deleted (previous question);
- should show condition be modified to take into account the nearest vehicles?


addons/cargo/functions/fnc_addCargoVehiclesActions.sqf
```diff
@@ -0,0 +1,42 @@
...
+    _name = getText (configFile >> "CfgVehicles" >> typeOf _x >> "displayName");
+    _ownerName = [_x, true] call EFUNC(common,getName);
+    if ("" != _ownerName) then {
+        _name = format ["%1 (%2)", _name, _ownerName];
+    };
+    _icon = (getText (configFile >> "CfgVehicles" >> typeOf _x >> "icon")) call BIS_fnc_textureVehicleIcon;
+    _action = [format ["%1", _x], _name, _icon, _statement, {true}, {}, [_x]] call ace_interact_menu_fnc_createAction;
+    _actions pushBack [_action, [], _target];
+    false
+} count (_player nearEntities [["Car", "Air", "Tank", "Ship", "Cargo_base_F"], MAX_LOAD_DISTANCE]);
```
commy2: Not all vehicles can load stuff. I don't see that filtered here.

addons/cargo/functions/fnc_addCargoVehiclesActions.sqf
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Example:
+ * [target, player] call ace_cargo_fnc_addCargoVehiclesActions
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_target", "_player"];
+private ["_statement", "_actions", "_name", "_ownerName", "_icon", "_action"];
```
commy2: please use `private keyword`. The variables from L31 onwards aren't even in the same script instance!

addons/cargo/functions/fnc_addCargoVehiclesActions.sqf
commy2: I mean by that, what happens if you're near a vehicle that is Car, but cannot load anything? (e.g. a Kart)

Dystopian: "could not be loaded" message appears as earlier in "Load" action. And I think it's OK, in RL you can't magically see which vehicles have enough room.
As for Kart - maybe it has sense to check vehicles cargo space config in show condition. But this behavior differs from current "Load" action, I didn't want to change it. If you insist I'll add it and remove "Load" action.

jonpas:
> should statement for Load menu be deleted? ATM it works as usually - loads to the nearest vehicle, while submenu items load to the specified vehicle;

I'd keep it.

> should Load menu be renamed to "Load to" e.g.? If yes statement should be deleted (previous question);

I don't think so, as it has it's own action as well I think it's good as is.

> should show condition be modified to take into account the nearest vehicles?

That probably wouldn't be bad if you can sort by distance.

addons/cargo/functions/fnc_startLoadIn.sqf
```diff
@@ -16,15 +16,16 @@
...
-
-if ((isNull _vehicle) || {_vehicle isKindOf "Cargo_Base_F"}) then {
-    {
-        if ([_object, _x] call FUNC(canLoadItemIn)) exitWith {_vehicle = _x};
-    } forEach (nearestObjects [_player, ["Cargo_base_F", "Land_PaperBox_closed_F"], MAX_LOAD_DISTANCE]);
+params ["_player", "_object", ["_vehicle", objNull]];
+TRACE_3("params",_player,_object,_vehicle);
+
+if (isNull _vehicle) then {
+    _vehicle = [_player, _object] call FUNC(findNearestVehicle);
```
commy2: Please don't reuse variable names in this way. Name it differently and set it to private.

addons/cargo/functions/fnc_addCargoVehiclesActions.sqf
```diff
@@ -0,0 +1,41 @@
...
+private _actions = [];
+
+{
+    private _name = getText (configFile >> "CfgVehicles" >> typeOf _x >> "displayName");
+    private _ownerName = [_x, true] call EFUNC(common,getName);
+    if ("" != _ownerName) then {
+        _name = format ["%1 (%2)", _name, _ownerName];
+    };
+    private _icon = (getText (configFile >> "CfgVehicles" >> typeOf _x >> "icon")) call BIS_fnc_textureVehicleIcon;
+    private _action = [format ["%1", _x], _name, _icon, _statement, {true}, {}, [_x]] call ace_interact_menu_fnc_createAction;
```
commy2: `EFUNC(interact_menu,createAction)`

addons/cargo/functions/fnc_startLoadIn.sqf
```diff
@@ -16,15 +16,16 @@
...
+    if ((isNull _vehicle) || {_vehicle isKindOf "Cargo_Base_F"}) then {
+        {
+            if ([_object, _x] call FUNC(canLoadItemIn)) exitWith {_vehicle = _x};
+        } forEach (nearestObjects [_player, ["Cargo_base_F", "Land_PaperBox_closed_F"], MAX_LOAD_DISTANCE]);
```
commy2: Not a fan of magic numbers in code.

addons/cargo/functions/fnc_addCargoVehiclesActions.sqf
Dystopian: my bad

addons/cargo/functions/fnc_startLoadIn.sqf
Dystopian: It's not my code. I just put it inside if-then. Anyway I don't understand about magic numbers.

addons/cargo/functions/fnc_startLoadIn.sqf
Dystopian: If you mean "_vehicle" I don't reuse it. I need it set to vehicle from parameters or from nearest vehicles.
If you mean something else I don't understand.

Dystopian: Or you mean I shouldn't reuse variable which is from params? OK, I'll fix it.

Dystopian: Also I'll add new argument description.

addons/cargo/functions/fnc_startLoadIn.sqf
commy2: Hard coded classnames like `"Cargo_base_F"` and especially `"Land_PaperBox_closed_F"`.
This will make it harder to add compatibility with new objects later, because you have to alter lines buried in the code instead of a separate config and will make it almost impossible for 3rd party mods.

Dystopian:
WIP

Dystopian:
All done, please review

SilentSpike:
Yeah I'd definitely still show the sub action when only one is in range for consistency

PR #4878, opened 2017-02-06
Handle changes to CBA_fnc_addPlayerEventHandler

PabstMirror:
Needed for CBA 3.2, but bwc with current.

"visibleMap" event is weird now: at the time the event runs the command `visibleMap` returns the previous value.
The fatigue bar was only showing on the map, but not in-game :smile: 

Similar problem with "unit", the command `name` would return the name of the previous AI.
Player reinforcements all had AI names on their nametags.

### DIFF
--- a/addons/advanced_fatigue/XEH_postInit.sqf
+++ b/addons/advanced_fatigue/XEH_postInit.sqf
@@ -14,12 +14,15 @@ if (!hasInterface) exitWith {};
     // - GVAR updating and initialization -----------------------------------------
     ["unit", FUNC(handlePlayerChanged), true] call CBA_fnc_addPlayerEventHandler;
 
-    private _fnc_showStaminaBar = {
+    ["visibleMap", {
+        params ["", "_visableMap"]; // command visibleMap is updated one frame later
         private _staminaBarContainer = uiNamespace getVariable [QGVAR(staminaBarContainer), controlNull];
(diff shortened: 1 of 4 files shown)


commy2:
Why was it using `visibleMap` to begin with when the point was to pass the new state as argument?

PabstMirror:
The point of `CBA_fnc_addPlayerEventHandler` is that we get an event on state change.
```
x = visableMap;
```
vs
```
params ["", "_visableMap"];
x = _visableMap;
```
1st would be cleaner and faster

commy2:
I could artificially delay both events via execNextFrame.

PabstMirror:
I don't think we need to change CBA.
The way it works isn't wrong, it's just odd; but we can easily handle it.

addons/advanced_fatigue/XEH_postInit.sqf
```diff
@@ -14,12 +14,15 @@ if (!hasInterface) exitWith {};
     // - GVAR updating and initialization -----------------------------------------
     ["unit", FUNC(handlePlayerChanged), true] call CBA_fnc_addPlayerEventHandler;
 
-    private _fnc_showStaminaBar = {
+    ["visibleMap", {
+        params ["", "_visableMap"]; // command visibleMap is updated one frame later
```
jonpas: Typo in `_visableMap`?

PR #4885, opened 2017-02-10
Prevent starting additional vital loops on full heal

### LINKED ISSUE #4884: Spawning new vitalLoop after each fullHeal
**CBA Version:** `latest. lazy ...` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`
- `others cuz I'm lazy`
**Description:**
After each fullHeal:
https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_treatmentAdvanced_fullHealLocal.sqf#L50
Set's the addedToUnitLoop to false although there might still be a vitalLoop running.
Then when anything happens that would cause a vitalLoop to get added 
https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_addVitalLoop.sqf#L25
this doesn't catch the duplicate anymore.

In my current game there are 101 vitalLoops running taking up to 50ms every few frames.

**Steps to reproduce:**
Get downed. Get full healed. get shot again.

- Dedicated

dedmen:
¯\_(ツ)_/¯

### DIFF
--- a/addons/medical/functions/fnc_treatmentAdvanced_fullHealLocal.sqf
+++ b/addons/medical/functions/fnc_treatmentAdvanced_fullHealLocal.sqf
@@ -47,7 +47,6 @@ if (alive _target) exitWith {
     _target setVariable [QGVAR(airwayCollapsed), false, true];
 
     // generic medical admin
-    _target setVariable [QGVAR(addedToUnitLoop), false, true];
     _target setVariable [QGVAR(inCardiacArrest), false, true];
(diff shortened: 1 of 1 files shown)


commy2:
You dropped a \/

PabstMirror:
Another possible fix would be to exit the vital loop if this var has been set to false.
But I think this PR should work fine.

PabstMirror:
Tag: Fix #4884

PR #4887, opened 2017-02-10
AdvFatigue - Fix unarmed sprint->prone + minor perf improvments

### LINKED ISSUE #4880: Exagerated stamina drop when going prone without weapon in hand

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
Massive stamina drop when going prone while running with no weapon in hands. It is exagerated compared to the stamina drop of the same action with a weapon on hands. You may lose more than half your stamina in just one action.

You can observe the animation makes a little "forward jump" just before hitting the floor that won't make when you have a gun. Maybe this has something to do with the stamina drop?

**Steps to reproduce:**
-Put your weapon on your back by pressing "0" or drop them all.
-Go sprinting with empty hands.
-While sprinting go prone.

Editor (Singleplayer)

**Placed Modules:**
- Fatigue system

PabstMirror:
Fix #4880 - Adds `GVAR(isProne)` which is set in getAnimDuty
Minor performance improvements; 
main one is setting total virtual load in movement (on `loadout` EH) so we can skip forEach loop

### DIFF
--- a/addons/advanced_fatigue/functions/fnc_getAnimDuty.sqf
+++ b/addons/advanced_fatigue/functions/fnc_getAnimDuty.sqf
@@ -21,6 +21,7 @@ private _duty = 1;
 private _animType = _animName select [1, 3];
 
 GVAR(isSwimming) = false;
+GVAR(isProne) = (stance _unit) == "PRONE";
 
 if (_animType in ["idl", "mov", "adj"]) then {
     switch (_animName select [5, 3]) do {
@@ -29,6 +30,7 @@ if (_animType in ["idl", "mov", "adj"]) then {
         };
         case ("pne"): {
             _duty = 10;
+            GVAR(isProne) = true; // #4880 - Unarmed sprint->prone has wrong `stance`
         };
         default {
             _duty = 1;
--- a/addons/advanced_fatigue/functions/fnc_getMetabolicCosts.sqf
+++ b/addons/advanced_fatigue/functions/fnc_getMetabolicCosts.sqf
@@ -18,18 +18,8 @@
 #include "script_component.hpp"
 params ["_unit", "_velocity"];
 
-private _virtualLoad = 0;
-{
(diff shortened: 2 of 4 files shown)


addons/advanced_fatigue/functions/fnc_getMetabolicCosts.sqf
```diff
@@ -18,18 +18,8 @@
...
-private _virtualLoad = 0;
-{
-    _virtualLoad = _virtualLoad + (_x getVariable [QEGVAR(movement,vLoad), 0]);
-} forEach [
-    _unit,
-    uniformContainer _unit,
-    vestContainer _unit,
-    backpackContainer _unit
-];
+private _gearMass = ((_unit getVariable [QEGVAR(movement,totalLoad), loadAbs _unit]) / 22.046) * GVAR(loadFactor);
```
commy2: Does this mean that fatigue requires movement now?

addons/advanced_fatigue/functions/fnc_getAnimDuty.sqf
```diff
@@ -21,6 +21,7 @@ private _duty = 1;
 private _animType = _animName select [1, 3];
 
 GVAR(isSwimming) = false;
+GVAR(isProne) = (stance _unit) == "PRONE";
```
commy2: Might as well use `isEqualTo` then.

addons/advanced_fatigue/functions/fnc_getMetabolicCosts.sqf
PabstMirror: Will fallback to just `loadAbs _unit` from the getVar

Glowbal:
Moved this to 3.10 milestone. We can potentially move to this to a 3.9.2 milestone if we need another patch  or 3.10 takes a while longer.

PR #4888, opened 2017-02-10
Fix frag MUL macros

PabstMirror:
Fix script errors from #4581 
It uses `MUL(` macro that doesn't seem to exist in cba/ace :confused:


654wak654:
Thought we had `*=` ¯\\\_(ツ)\_/¯

jonpas:
Actually from #4803

654wak654:
t-what now?

PR #4895, opened 2017-02-12
Convert ACE_settings to CBA_settings

PabstMirror:
Converts ACE_settings to CBA_settings
- Offers BWC with modules and missionConfig ace_settings.  
So old missions should continue to function as before.
- Dropping support for ace_server, as that can be handled by CBA server programming.

I'm sure we'll want to eventually convert them by hand to sqf.
The main downside to this is that SCALAR gets converted to SLIDER, which requires a min/max.
So max needs to be big for a distance in meters, but would make no sense for a bleeding coefficent.
But for now it will let us convert all 275 settings at once.

ToDo:
- [x] Setting Modules need to run globaly (check both config and module sqf)
- [x] Fix 3 uses of `setSetting`
- [x] Remove Options Menu (keep debug / headbug fix somehow?)
- [x] Set all settings modules to `scope = 1`
- [x] Replace [Settings Framework documentation](https://ace3mod.com/wiki/framework/settings-framework.html) with a link to [CBA Settings System documentation](https://github.com/CBATeam/CBA_A3/wiki/CBA-Settings-System)
- [x] Testing

Converting is less than ~100ms; Opening game options with all these settings is ~2 sec lockup.


jonpas:
> keep debug / headbug fix somehow?

Have that button where "ACE Options" button is current in main menu? Too open and visible?

jonpas:
Headbug fix and Debug buttons could act like the main menu vanilla buttons do:
```
ACE OPTIONS
    DEBUG
    HEADBUG FIX
```
where the indented ones show upon clicking the "ACE OPTIONS".

Similar to what @commy2 has done to the vanilla menu at https://github.com/CBATeam/CBA_A3/pull/634.

PabstMirror:
This should be mostly usable, 

@commy2 
With a lot of settings, opening the CBA "addon options" menu will cause a long hard lock
(5202.00 ms for 365 settings on my machine)
makes it difficult to tweak settings 

Looking at CBA's fnc_initDisplayGameOptions.sqf/gui_createMenu.sqf
It seems to be creating all controls at once, 
We might be able to make that more dynamic and just create controls for whatever category/machine list is open?

commy2:
ref https://github.com/CBATeam/CBA_A3/pull/739

jonpas:
Moving to 3.12.0, "CBA Settings" milestone will stay for manual conversion of components from 3.13.0 onwards.

jonpas:
Moved Debug and HeadBug Fix to Pause Menu under Options with the new CBA function and removed "ACE Options" Main Menu button. I didn't cleanup all the functions, that can be done later.

ACE3 Pause Menu Options

Too obvious / in the open?

jonpas:
Used regex to change scope:
`(?s)ACE_Module {(.+?)scope = (.);` -> `ACE_Module {$1scope = 1;`
and changed some manually.

Left modules that don't have Eden counterparts. Can be cleaned up later.

Left-Over Modules

jonpas:
Only this warning shows at the moment:
```
16:01:31 [ACE] (common) WARNING: Setting [ace_medical_enableAdvancedWounds] - value type [BOOL] is missing number z\ace\addons\common\functions\fnc_cbaSettings_loadFromConfig.sqf:65
```

PabstMirror:
isNumber (configFile >> "ACE_settings" >> "ace_medical_enableAdvancedWounds") = false

```
    class GVAR(enableAdvancedWounds) {
        value = false;
```
I think it will be ok on the make.py version, in either case the default value (`0`) is fine

PR #4897, opened 2017-02-13
More accurate Korean translate for medical

sbng-PAPA:
korean player, confused about name of wounds.
replaced it more accurate name.


PabstMirror:
@jtj5283449 - Any thoughts on these changes?

PR #4904, opened 2017-02-15
Fix multiple slideshows sub-actions, Add set name setting

### LINKED ISSUE #4901: More than One Slideshow / overlap menus
**Arma 3 Version:** 1.66.139586 (stable)
**CBA Version:** latest
**ACE3 Version:** `latest befor 3.9.0

**Mods:**
- `@CBA_A3`
- `@ace`
- BWMod

**Description:**
Hi guys...sorry for my bad english (german). I hope that you understand the Problem and can help. 
We will place more then one Slideshow on the Map. All Slideshows have the same controller. When we interact with this controller, the Slides/Dias show corecctly but the next menu is on top. You cant read the text, because the letters are above the other word fromthe other slideshow. (PICTURE)
Why is Slide1 shown, when i select Slide2...and how can i rename "Dias"?

German:
Hallo liebes ACE-Team,
Wir wollen auf unserer Karte mehr als nur eine Slideshow mit dem selben Controller platzieren. Nun werden auf dem Controller auch alle Slideshows angezeigt...aber die Untermenüs "überlappen" mit der vorhergehenden Slideshow. Warum wird mir diese noch angezeigt wenn ich doch die andere gewählt habe? Und kann man die einträge "Dias" ersetzen, so das man dort eigenen Text verwenden kann?
das würde das Handling der Slideshows sehr erleichtern!

 

**Steps to reproduce:**

- Dedicated / Self-Hosted Multiplayer / Singleplayer

**Placed Modules:**
- Slideshow

jonpas:
- Fix #4901 
- Use `private KEYWORD` instead of `private ARRAY`
- Add `Set Name` setting to allow distinguishing between multiple slideshow sets (#4901)
- Remove `Reference INFO for object support.` as there is no "INFO" in Eden
- Update documentation

### DIFF
--- a/addons/slideshow/CfgVehicles.hpp
+++ b/addons/slideshow/CfgVehicles.hpp
@@ -35,6 +35,12 @@ class CfgVehicles {
                 typeName = "STRING";
                 defaultValue = "";
             };
+            class SetName {
+                displayName = CSTRING(SetName_DisplayName);
+                description = CSTRING(SetName_Description);
+                typeName = "STRING";
+                defaultValue = "";
+            };
             class Duration {
                 displayName = CSTRING(Duration_DisplayName);
                 description = CSTRING(Duration_Description);
--- a/addons/slideshow/functions/fnc_createSlideshow.sqf
+++ b/addons/slideshow/functions/fnc_createSlideshow.sqf
@@ -19,7 +19,6 @@
  */
 #include "script_component.hpp"
 
-private ["_currentSlideshow", "_slidesAction", "_varString"];
 params [
     ["_objects", [], [[]] ],
     ["_controllers", [], [[]] ],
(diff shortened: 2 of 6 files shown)


Rocky83:
107410_screenshots_20170320184842_1

Hi...on server we have that Problem with the Slideshows...Testing in Eden or SP it will work! But we can wait to the Update 3.9.1. When it is coming?

Rocky83:
OK...Update is on the Server...same issue!!! @jonpas

jonpas:
I will need your RPT and a screenshot of the module.

Rocky83:
Hi...thx
In EDEN/Testing Szenario i see the correct way!

But on Multiplayer Server i see like on TOP

Here are my Modulsetting:


Here the correct way in SP/EDEN

PR #4905, opened 2017-02-15
Medical/Repair 3den Toolbox State Fix (Bug fix for #4902)

### LINKED ISSUE #4902: Ace Unit Attributes Medic, Doctor, Engineer, ect resetting to default or swapping settings with other units
**Arma 3 Version: 1.66.139586
**CBA Version: 3.1.2.161105
**ACE3 Version: 3.8.3

**Mods:**
- @CBA_A3
- @ace

**Description:**
- I set the ace attribute options for medics and engineers in a playable squad, after doing this i save the mission and go to double check the units and all the ACE options have ether reset to default, swapped engineers and medics around, or made the unit both a medic and an engineer. The main problem this causes is the medics dont have the the ability to give IV's.

**Steps to reproduce:**
- place down any 4 units (tried with medics and engineers as well as other units) through the attribute ace options set 2 as medics and 2 as engineers, now when going back through the units to check their settings, all the medic and engineer settings will have swapped, reset or doubled up.

- Editor, Dedicated windows server

**Placed Modules:**
- None

AACO:
Simple fix to store the toolbox state on the control level, instead of the missionNamespace level.

Resolves #4902


addons/medical/CfgEden.hpp
```diff
@@ -9,8 +9,8 @@ class Cfg3DEN {
             };
         };
         class GVAR(isMedicControl): Title {
-            attributeLoad = "(_this controlsGroupCtrl 100) lbsetcursel (((_value + 1) min 3) max 0);";
-            attributeSave = "(missionnamespace getvariable ['ace_isMedic_temp',0]) - 1;";
+            attributeLoad = "(_this controlsGroupCtrl 100) lbSetCurSel (((_value + 1) min 3) max 0);";
+            attributeSave = "(_this getVariable ['ace_isMedic_temp', 0]) - 1;";
```
commy2: Why does this exist in the first place?

PabstMirror: That's how BIS does attributes:
```
class Rank: Toolbox {
    attributeLoad = "(_this controlsGroupCtrl 100) lbsetcursel (['private','corporal','sergeant','lieutenant','captain','major','colonel'] find _value);";
    attributeSave = "['private','corporal','sergeant','lieutenant','captain','major','colonel'] select (missionnamespace getvariable ['Rank_value',0]);";
    class Controls: Controls {
        class Value: Value {
            tooltips[] = {"Private", "Corporal", "Sergeant", "Lieutenant", "Captain", "Major", "Colonel"};
            onToolboxSelChanged = "missionnamespace setvariable ['Rank_value',_this select 1];";
        };
    };
};
```
Does rank have the same issue as ace's medic setting?

commy2: Storing this on the control seems really pointless. The control itself already has a state. I'm not sure this fix is correct.

AACO: If we use a default value of 0 we can probably get away with just using the toolbox state, otherwise it'd always save to the SQM, albeit it'd be the correct value.

AACO: @PabstMirror to your point about the rank module, I'm really not sure. The rank (and stance) toolboxes are stored as bonafide unit attributes, so I'm going to speculate and say BI is cheating and treats those slightly differently.

PabstMirror:
Talking with @AACO - The way BIS does rank must use some special engine stuff as it works fine.

The current code's problem is using `ace_isMedic_temp`
It was only being changed on toolBoxChange, but not on load.
So opening and saving the attirbutes could change the value to the last settting.

commy2:
> The way BIS does rank must use some special engine stuff as it works fine.
:\

PR #4913, opened 2017-02-16
Hide earplug actions if setting disabled

### LINKED ISSUE #4912: Earplugs break if combat deafness is disabled

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Earplugs break if combat deafness is disabled. Usually when the game starts variable ace_common_setHearingCapabilityMap contains ["ace_hearing",1], and then after putting earplugs in it changes to ["ace_hearing",0.5]. But if combat deafness is disabled array ace_common_setHearingCapabilityMap is empty at the start but as soon as you put your earplugs in it set's it to ["ace_hearing",1] and after removing them it resets it too ["ace_hearing",0.5] which produces reverse effect. Possible solution is probably just block the use of the earplugs action if deafness is disabled.

**Steps to reproduce:**
- Set earplugs in.

**Placed Modules:**
- Hearing module with combat deafness disabled.

**RPT log file:**
- N/A.

PabstMirror:
Fix #4912

Hides the actions if combat deafness setting is disabled.
Also add exits on the functions in case they are manually called by a script/mission

### DIFF
--- a/addons/hearing/functions/fnc_putInEarplugs.sqf
+++ b/addons/hearing/functions/fnc_putInEarplugs.sqf
@@ -16,6 +16,7 @@
 #include "script_component.hpp"
 
 params ["_player"];
+
 if (!GVAR(EnableCombatDeafness)) exitWith {};
 
 // Plugs in inventory, putting them in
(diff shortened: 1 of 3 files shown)


addons/hearing/functions/fnc_putInEarplugs.sqf
```diff
@@ -16,6 +16,7 @@
 #include "script_component.hpp"
 
 params ["_player"];
+if (!GVAR(EnableCombatDeafness)) exitWith {};
```
jonpas: Should probably do an empty line above.

PR #4917, opened 2017-02-18
Verify params in atragmx profile gunlist

### LINKED ISSUE #4910: Script error when opening ATragMx
**ACE3 Version:** `3.9.0` stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- When opening the ATragMx a script error is shown
(possible old gun list?)

**Steps to reproduce:**
- Go into editor
- place down AB module
- take an ATragMx with you
- open the ATragMx

- Editor (Singleplayer)

**Placed Modules:**
- AB (with AB set to "Enabled")

**RPT log file:**
- http://pastebin.com/RBzfDp8a

PabstMirror:
- Delay all init until first use
- Verify each gun in profile list with `isEqualTypeArray`

Should fix #4910

@bux, if you can still reproduce the error, it would be great if you could try testing this branch.


bux:
@PabstMirror No script error when opening the ATragMx anymore.
Seems to have fixed it. 👍

PR #4931, opened 2017-02-21
Vehicle quick mount

jameslkingsley:
- Add the quick-mount keybind that when pressed will mount the player into the vehicle based on direction and distance

The idea of this module is to provide a fail-safe for when action menus fail to appear when you're trying to quickly mount a vehicle. An example is when you're running towards an extraction helicopter and you're unable to get in when in real life you know you should be able to.

Demonstration: https://www.youtube.com/watch?v=3LWhYdG02xQ

I still need to convert it to ACE, just leaving it here for now.


PabstMirror:
Ref ACEX PR: https://github.com/acemod/ACEX/pull/29

addons/quickmount/functions/fnc_getInNearest.sqf
```diff
@@ -0,0 +1,92 @@
...
+        GVAR(priority) = 0;
+    };
+
+    private _seats = ["Driver", "Gunner", "Commander", "Cargo"];
+    private _sortedSeats = [(_seats select GVAR(priority))];
+    _seats deleteAt GVAR(priority);
+    _sortedSeats append _seats;
+
+    {
+        private _unit = _target call compile format ["assigned%1 _this", _x];
```
commy2: wtf?

addons/quickmount/functions/fnc_getInNearest.sqf
```diff
@@ -0,0 +1,92 @@
...
+
+        if (_unit isEqualType objNull && {!isNull _unit} && {!alive _unit}) exitWith {
```
commy2: As opposed to what?
If it's not OBJECT, line 65 will error anyway.

addons/quickmount/functions/fnc_getInNearest.sqf
jameslkingsley: It's to avoid repeating code

commy2: `assignedDriver` does not report the driver of a vehicle etc.

addons/quickmount/functions/fnc_getInNearest.sqf
jonpas: What does line 65 have to do with `_unit`?

bux:
Well made video @jameslkingsley

addons/quickmount/CfgVehicles.hpp
```diff
@@ -0,0 +1,37 @@
+class CfgVehicles {
+    class ACE_Module;
+    class GVAR(module): ACE_Module {
+        author = ECSTRING(common,ACETeam);
+        category = "ACEX";
```
PabstMirror: "ACE"

addons/quickmount/functions/fnc_getInNearest.sqf
```diff
@@ -0,0 +1,95 @@
...
+#include "script_component.hpp"
+
+if (!GVAR(enabled) ||
+    {isNull ACE_player} ||
+    {vehicle ACE_player != ACE_player} ||
+    {!alive ACE_player} ||
+    {ACE_player getVariable ["ace_unconscious", false]}
+) exitWith {};
+
+private _start = AGLtoASL positionCameraToWorld [0, 0, 0];
```
PabstMirror: `private _start = AGLtoASL (ACE_player modelToWorldVisual (ACE_player selectionPosition "pilot"));`
I think this will work a lot better for people using 3rd person

PabstMirror:
Tweaked the seat search to use `fullCrew`
handles copilots and turret gunners better.
e.g. the ghosthawk only showed 10 `emptyPosition` but holds 12

PR #4938, opened 2017-02-22
Fix setVariablePublic reseting to old values

### LINKED ISSUE #4932: FCS ranging not working
**Arma 3 Version:** `1.66` (stable / performance v18)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
Ranging the gun of a Vehicle in MP is not really adjustable due to a forth and back jumping

**Steps to reproduce:**
Start a missions in Singleplayer. Change ranging of a Slammer/Mercava manually - it works
Do the exact same in MP and the ranging jumps back and forth

- Dedicated / Self-Hosted Multiplayer / Editor (Multiplayer) 

**Placed Modules:**
-none
Is not depending on modules

**RPT log file:**
http://pastebin.com/Kq8isc8q
From the DS (ignore the script errors in classname.sqf and classname_extension.sqf)

PabstMirror:
Fix #4932

setVariablePublic should call itself again with the current value, not the value from the first call

### DIFF
--- a/addons/common/functions/fnc_setVariablePublic.sqf
+++ b/addons/common/functions/fnc_setVariablePublic.sqf
@@ -48,6 +48,7 @@ TRACE_2("Starting Embargo", _varName, _delay);
 
     //If value at start of embargo doesn't equal current, then broadcast and start new embargo
     if (!(_value isEqualTo _curValue)) then {
-        [_object, _varName, _curValue] call FUNC(setVariablePublic);
+        _this set [2, _curValue];
+        _this call FUNC(setVariablePublic);
     };
 }, _this, _delay] call CBA_fnc_waitAndExecute;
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_setVariablePublic.sqf
```diff
@@ -48,6 +48,7 @@ TRACE_2("Starting Embargo", _varName, _delay);
 
     //If value at start of embargo doesn't equal current, then broadcast and start new embargo
     if (!(_value isEqualTo _curValue)) then {
-        [_object, _varName, _curValue] call FUNC(setVariablePublic);
+        _this set [2, _curValue];
+        _this call FUNC(setVariablePublic);
```
commy2: Why was this changed? I think it's bad practice to modify the input array.

PabstMirror: This will pass 4th arg which may or may not exist

PR #4941, opened 2017-02-23
Reload medical configs into extension when loading save

### LINKED ISSUE #4940: ACE medical system dont work after game restart
**Arma 3 Version:** stable 1.66 / dev 1.69
**CBA Version:** 3.1.2.161105
**ACE3 Version:** 3.9.0 

**Mods:**
- @CBA_A3
- @ace
- All RHS 0.4.2.1

**Description:**
- The game use the vanilla medical system after game restart. it looks like everything else works
- Loading a save game without restart works

**Steps to reproduce:**
- Start a mission with ace + cba 
- test the medical system
- save and restart game
- test medical system again

- Singleplayer

PabstMirror:
Fix #4940

### DIFF
--- a/addons/medical/XEH_preInit.sqf
+++ b/addons/medical/XEH_preInit.sqf
@@ -50,6 +50,9 @@ addMissionEventHandler ["Loaded",{
             TRACE_1("preload done",_this);
         }, [_x]] call CBA_fnc_waitUntilAndExecute;
     } forEach GVAR(fixedStatics);
+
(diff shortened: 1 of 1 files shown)


commy2:
Very nice. Didn't think it was this easy.

dedmen:
Btw this causes a lot of extension calls even if the extension is not available (Linux for example) I already reported that on Slack a few months ago and just got another report in Slack #servers. Nothing to worry about besides a few dozen messages in RPT.

PR #4942, opened 2017-02-23
Fix lost magazines when restoring gunbag mags

### LINKED ISSUE #4927: Placing rifle into gunbag consumes magazine
**Arma 3 Version:** stable
**CBA Version:** stable
**ACE3 Version:** stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- When placing a rifle into a gunbag, the magazine in the rifle disappears. When the rifle is pulled out of the gun bag, it has a magazine already loaded, but the player's magazine count is reduced by 1.  This can be prevented by placing the magazine in the rifle in the player's inventory before placing the rifle in the gun bag. The magazine is not going into the inventory nor is it being left on the ground.

I have had multiple magazines disappear at one swap, but I haven't be able to reliably reproduce it without RHS also being loaded. With no other mods other than CBA and ACE, I've been able to reliably produce 1 magazine disappearing per placement in gunbag. 

**Steps to reproduce:**
- Get rifle, several magazines in inventory, and the ace gunbag
- Note magazine count
- Place rifle in gun bag
- Get rifle out of gun bag
- Note magazine count has been reduced by 1

- Self-Hosted Multiplayer

**Placed Modules:**
- Virtual Arsenal

PabstMirror:
Fix #4927

### DIFF
--- a/addons/gunbag/functions/fnc_offGunbagCallback.sqf
+++ b/addons/gunbag/functions/fnc_offGunbagCallback.sqf
@@ -29,6 +29,16 @@ if (_state isEqualTo []) exitWith {
 _state params ["_weapon", "_items", "_magazines"];
 
 _unit addWeapon _weapon;
+
+// Game will auto add magazines from player's inventory, put these back in player inventory as they will be overwritten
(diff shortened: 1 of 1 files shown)


commy2:
lgtm, will have to test ingame

Nich10solo:
Thank you!

PR #4956, opened 2017-02-27
Fix AB water vapor pressure calc

### LINKED ISSUE #4951: AtragMx : relative humidity calculation ineffective
**ACE3 Version:** `3.9.0 / 3.8.3` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- AB enabled, AtragMx opened, the relative humidity calculation has no effect on the elevation.

**Steps to reproduce:**
- Open the AtragMx, choose a cartridge, enter a distance and play with the relative humidity by checking the elevation.

- Singleplayer / Editor (Singleplayer)

**Placed Modules:**
- Advanced Ballistics module enabled manually.

**RPT log file:**
- https://gist.github.com/J3aLeiNe/bb953c60b042f5f3eb89d35c0025c7fe

PabstMirror:
Fix #4951

This will effect both AtragMX and AB.
`_pSat` was in hPa, needs to be converted to Pa

This will require dll rebuild, so I'm thinking push to 3.10?
The effect on air density is fairly small.

Edit: We could just change all extension calls to use `100 * EGVAR(weather,currentHumidity)` to avoid having to recompile? Effect would be the same.


bux:
good find

extensions/advanced_ballistics/AdvancedBallistics.cpp
```diff
@@ -94,7 +94,7 @@ double calculateAirDensity(double temperature, double pressure, double relativeH
 
     if (relativeHumidity > 0) {
         double _pSat = 6.1078 * pow(10, ((7.5 * temperature) / (temperature + 237.3)));
```
TheMagnetar: Shouldn't 6.1078 (hPa) be converted to Pa directly here instead of converting it in `vaporPressure`? I find it more clear than later multiplying by 100.0 and having _`pSat` in hPa while `vaporPressure` in Pa

PabstMirror: Going to use `610.78` and added note

PR #4957, opened 2017-02-28
Fix misprint in maptools

Dystopian:
UTF BOM in xml removed accidentally.


PabstMirror:
Let's delay until cba settings so we don't lose saved profile settings.

Dystopian:
ignore changelog please

PR #4964, opened 2017-03-03
Fix rangecard not showing wind/lead

### LINKED ISSUE #4962: No Wind and Lead values on Rangecard (AB off)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
When advanced_ballistics is turned off Wind and Lead values aren't available on the range card anymore. I don't know if it's by design but these values were available in 3.8.3. and the projectile is still affected by the wind with advanced ballistic turned off (so wind modifications are needed for accurate fire).
3.8.3 : https://i.imgur.com/Y3jtfi7.jpg
3.9.0 : https://i.imgur.com/8ITzHKd.jpg

**Steps to reproduce:**
- Disable ace_advanced_ballistics (ace_advanced_ballistics_enabled = 0).
- Play as a sniper in the editor.
- Add yourself item "ACE_rangecard" and open it

- found during a game on Dedicated / reproduced in Editor (Singleplayer)

**Placed Modules:**
No modules but modified serverconfig.hpp
`class ace_advanced_ballistics_enabled {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_advanced_ballistics_simulateForSnipers {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_advanced_ballistics_simulateForGroupMembers {
    value = 0;
    typeName = "BOOL";
    force = 0;
};
class ace_advanced_ballistics_simulateForEveryone {
    value = 0;
    typeName = "BOOL";
    force = 0;
};`
full serverconfig : http://pastebin.com/nsjMDNQy

**RPT log file:**
http://pastebin.com/bBssyQqx

PabstMirror:
Fix #4962

I think this is right
line 229 does `} forEach [0, 3, 8];`

### DIFF
--- a/addons/rangecard/functions/fnc_updateRangeCard.sqf
+++ b/addons/rangecard/functions/fnc_updateRangeCard.sqf
@@ -174,7 +174,7 @@ if (isNil {_cacheEntry}) then {
             [_scopeBaseAngle,_boreHeight,_airFriction,_mv,_x,EGVAR(scopes,zeroReferenceBarometricPressure),EGVAR(scopes,zeroReferenceHumidity),100,4,1,GVAR(rangeCardEndRange),_bc,_dragModel,_atmosphereModel,_transonicStabilityCoef,_forEachIndex,_useABConfig] call FUNC(calculateRangeCard);
         } forEach [-15, -5, 5, 10, 15, 20, 25, 30, 35];
     } else {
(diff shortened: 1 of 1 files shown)


Glowbal:
Created an issue to one day solve that issue. Will track in #4978.

PR #4971, opened 2017-03-05
Sandbags surfaces update

jonpas:
- Add Tanoa surfaces support to Sandbags
- Check height above terrain, if more than 0.05 disallow sandbags (when on buildings and similar)
- Check if in water (ATL is negative during shallow swimmable water for some reason)
- Use private keyword

### DIFF
--- a/addons/common/XEH_PREP.hpp
+++ b/addons/common/XEH_PREP.hpp
@@ -13,6 +13,7 @@ PREP(ASLToPosition);
 PREP(binarizeNumber);
 PREP(blurScreen);
 PREP(cachedCall);
+PREP(canDig);
 PREP(canGetInPosition);
 PREP(canInteractWith);
 PREP(changeProjectileDirection);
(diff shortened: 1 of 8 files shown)


PabstMirror:
Should we update the list in ace_trenches as well?
https://github.com/acemod/ACE3/blob/master/addons/trenches/functions/fnc_canDigTrench.sqf

jonpas:
Oh, didn't even realize there is a function that's exactly the same. I'll move it to common even.

PabstMirror:
Can build trench on stratis runway
surfaceType = "#GdtStratisConcrete";   CfgSurface -> `concrete` has dust of 0.15

Maybe we should keep the blacklist and just use the dust as a backup for external maps?
Or tweak dust cutoff?

nateberkopec:
I have a crapton of mods/terrains loaded, noticing lots of concrete surfaces use 0.15 for `dust`. XCam-Taunus appears to use 0.15 for many of their road surfaces as well.

jonpas:
Will look at it and possibly just keep dust as fall-back if higher value will cause other problems.

jonpas:
Changed to fall-back for dust, as it does not work as expected for some cases.

PR #4983, opened 2017-03-10
Use parseSimpleArray

654wak654:
- Use the new [parseSimpleArray](https://community.bistudio.com/wiki/parseSimpleArray) command instead of `call compile` where possible.

### DIFF
--- a/addons/markers/XEH_preInit.sqf
+++ b/addons/markers/XEH_preInit.sqf
@@ -39,7 +39,7 @@ if (isNil QGVAR(MarkerColorsCache)) then {
 
             {
                 if !( _x isEqualType 0) then {
-                    _rgba set [_forEachIndex, call compile _x];
+                    _rgba set [_forEachIndex, parseSimpleArray _x];
                 };
             } forEach _rgba;
 
--- a/addons/common/functions/fnc_addActionEventHandler.sqf
+++ b/addons/common/functions/fnc_addActionEventHandler.sqf
@@ -42,7 +42,7 @@ _actions pushBack [_condition, _statement];
 
 // first action to add, unit needs addAction command
 if (_actionID == -1) then {
-    private _addAction = parseSimpleArray format [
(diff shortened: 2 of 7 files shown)


jonpas:
Requires 1.68.

654wak654:
On second thought, `parseSimpleArray` probably can't parse those addActions. Downloading RC branch.

addons/markers/XEH_preInit.sqf
```diff
@@ -39,7 +39,7 @@ if (isNil QGVAR(MarkerColorsCache)) then {
 
             {
                 if !( _x isEqualType 0) then {
-                    _rgba set [_forEachIndex, call compile _x];
+                    _rgba set [_forEachIndex, parseSimpleArray _x];
```
PabstMirror: throws script error, `_x` is not an array
might be able to use on _rgba, unless mixed mode arrays are possible?
e.g. `color[] = {1,0,0,"(profilenamespace getvariable ['Map_Unknown_A',0.8])"};`

PabstMirror: oh yeah, `parseSimpleArray` won't be able to handle expressions like `profileNamespace getVar`

PabstMirror:
This command is made for reading machine made strings.
It chokes pretty hard on stuff a human might write

`parseSimpleArray "[- 1.0,2.,3 ,4]"` = [0,nil,nil,4]
`call compile "[- 1.0,2.,3 ,4]"` = [-1,2,3,4]

`parseSimpleArray "[""ace_zeus"", ""ace_attach""]"` = ["ace_zeus",nil]
`call compile "[""ace_zeus"", ""ace_attach""]"` = ["ace_zeus","ace_attach"]

commy2:
You have to sanitize white space first at which point it's probably slower than `call compile`.
Benefit is not being able to cheat. Negative is that it can't handle expressions like `profileNamespace getVariable`.

PR #4984, opened 2017-03-11
Underwater actions support

jonpas:
- Allow majority of ACE3 actions to work underwater
- Allow certain actions inside vehicle as well
- Allow Advanced Throwing underwater - only drop mode
- Use custom "Point" gesture for pointing (BI ones do not work underwater) if gestures component is present - is also a feature request
- Enable vehicle lock on `Plane`, `Motorcycle` and `Ship_F`
- ~Fix vehicle lock not showing lock picking inside vehicles~ -> #4985 

Limitations:
- BI gestures do not work underwater
- ~~Some animations aren't exactly fitting~~ Solved by not playing them

Todo:
- [x] `logistics_wirecutter`
- [x] Don't play some animations (eg. kneeling) undewater
- ~~Fix repair stuck in animation if successful underwater~~ Fixed with above
- [x] Fix [getVehiclePos](https://github.com/acemod/ACE3/blob/master/addons/interaction/functions/fnc_getVehiclePos.sqf#L50) below 0 height (underwater) - it moves based on player height as that is minimal
- [x] Improve vehicle attach underwater
- ~~Use `lineIntersectsSurfaces` (works underwater) instead of `lineIntersectsWith` (doesn't work underwater) for anything else that uses it and should work underwater~~ -> #5263 
- ~~Refuel~~ Works as long as you don't go underwater


Max255PL:
Module `logistics_wirecutter` is missing from the changes.

Will be testing this more on dedicated in coming days.

addons/vehiclelock/CfgVehicles.hpp
```diff
@@ -61,6 +64,16 @@ class CfgVehicles {
     class Helicopter: Air {
         MACRO_LOCK_ACTIONS
     };
+    class Plane: Air {
```
PabstMirror: I don't think we need it on aircraft

addons/reload/functions/fnc_startLinkingBelt.sqf
```diff
@@ -56,4 +56,4 @@ private _onFailure = {
 [_player, _magazineType, _maxAmmo] call EFUNC(common,removeSpecificMagazine);
 
 // Call progress bar
-[4, [_player, _target, [_magazineType, _maxAmmo]], _onFinish, _onFailure, (localize LSTRING(LinkingBelt)), _condition] call EFUNC(common,progressBar);
+[4, [_player, _target, [_magazineType, _maxAmmo]], _onFinish, _onFailure, (localize LSTRING(LinkingBelt)), _condition, ["isNotInside", "isNotSwimming"]] call EFUNC(common,progressBar);
```
PabstMirror: linking belts and reloading rpgs while swimming ?

addons/attach/functions/fnc_placeApprove.sqf
```diff
@@ -54,9 +54,20 @@ while {(_closeInMax - _closeInMin) > 0.01} do {
...
+                    _args params ["_startingPosShifted", "_endPosShifted", "_timeAdded"];
+                    drawLine3D [_startingPosShifted, _endPosShifted, [1,0,0,1]];
+                    if (_timeAdded + 5 < CBA_missionTime) then {
+                        [_idPFH] call CBA_fnc_removePerFrameHandler;
+                    };
+                }, 0, [_startingPosShifted, _endPosShifted, CBA_missionTime]] call CBA_fnc_addPerFrameHandler;
+            #endif
+
+            // Default max results is 1, so take only first subarray and select parentObject (object itself or parent of proxy)
+            private _intersectObject = ((lineIntersectsSurfaces [_startASL, _endASL, _unit]) select 0) select 3;
```
PabstMirror: This could be nil if `lineIntersectsSurfaces` is []
But it won't throw an error so it's not really a problem

jonpas: `select 3` would throw an error though.

addons/reload/functions/fnc_startLinkingBelt.sqf
jonpas: 😆

addons/attach/functions/fnc_placeApprove.sqf
jonpas: No error, was just `nil`, but used `param` now to make sure anyways.

Max255PL:
Any chance this could make it into 3.10 or still needs more work/testing?

jonpas:
It is ready from my side, tested as much as I could, it could use more testing for sure though.

3.10 is already staged, no more features.

jonpas:
#5350 would solve the Dogtags and Captives stuff from inside a vehicle (Passenger actions), this PR adds support for that, but it won't work.

This PR is considered ready to merge now.

PR #4989, opened 2017-03-14
Fix zeus race condition

AACO:
**Fix zeus race condition:**
- Due to the fact that the bi zeus code is executed by all clients on module init, it's possible for the client to be looking for the admin variable that is not yet set/transmitted from the dedicated server, especially at heavy loads at the start of missions.

This PR simply checks if the variable is not set, retries getting the variable for 10 seconds (or until the variable is non empty). If the variable is still empty at the end of those 10 seconds it logs an error and exits the thread.

Symptom:
Error
The logged in admin will also not be attached to the curator when the error is thrown.

edit: this is very much just a bandage on top of a system that looks like it needs some work on simplification/flow.


commy2:
`fnc_bi_moduleCurator.sqf` is just a copy paste from the BI function with some added lines for us to do stuff. Dunno if it's within our scope to fix them.

jonpas:
This has not been fixed in last update either, but I would not fix it via ACE3, this is something in BI's system, they should implement the fix. It is also using `waitUntil`, which I wouldn't be against if it's just a copy of BI's function, but since this is adding to it, I am.

AACO:
I mean, fair enough, BI should fix it (and I agree `waitUntil` isn't great, but it's in form with how the file is setup). With that said I've run into this problem a lot, especially with high player counts, so if it's within the ACE project it can be bandaged on the ACE side. On the plus side my group has tools to easily deal with Zeus assignments, but anyone without the tools and is relying on the code to assign zeus modules is SOL

jonpas:
@Glowbal @commy2 ?

commy2:
Out of scope imo. But we should check if BI updated the function in the mean time.

jonpas:
I have checked for 1.68, will check in 1.70 and dev in a bit.

Glowbal:
> Out of scope imo. But we should check if BI updated the function in the mean time.

Agreed.

jonpas:
Checked stable 1.70 and dev 1.71, no change in the function from what we have (except changes we made).

The error is by no means a game breaker and is a BI issue, I say close this. Thanks for the contribution @AACO, hope you can understand the reasoning. :)

AACO:
Fair enough

PR #4990, opened 2017-03-14
Fix cookoff desynch

### LINKED ISSUE #4900: Cookoff sound and visuals desynchronize
**ACE3 Version:** `3.x.x` (dev + c0c41667)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- Sound and visual effects have a random delay/desync on the dedicated server. Visualization of the problem: [YouTube](https://www.youtube.com/watch?v=Bdn1XfyfpXA)

**Steps to reproduce:**
- Fire at an armored vehicle with an AT to cause a cookoff.

- Dedicated

**Placed Modules:**
- None

**RPT log file:**
- [RPT](https://gist.github.com/theYoursTruly/a80710513e783ad511bf6a605dcb8a24)

commy2:
- fix #4900

Randomness would be calculated on every machine, but only vehicle explosion and sound are done by the server. Smoke, light and fire sfx are done by each client.

### DIFF
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -132,6 +132,6 @@ if (local _vehicle) then {
             if (local _vehicle) then {
                 _vehicle setDamage 1;
             };
-        }, [_vehicle, _effects], 4 + random 20] call CBA_fnc_waitAndExecute;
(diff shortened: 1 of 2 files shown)


654wak654:
Q: Could a better solution be to calculate and broadcast a number from a single machine? Is it not worth the traffic?

commy2:
That would be the full solution without having to sacrifice any functionality (randomness), but I think it would require a lot of rewriting.

PR #5010, opened 2017-03-20
Frag - Cleanup and Performance

### LINKED ISSUE #5001: Several GL automatic weapons - always noticeably destroys the game performance.
**CBA Version:** `3.1.2.161105` (stable)

**Mods:**
- CBA_A3
- ace
- JSRS Sound Mod

**Description:**
Hi!
I really like the feature of ACE, in which almost all the weapons with the explosives have fragments. This is cool feature! However, if it is an automatic GL-weapon, then it noticeably slows down the game performance.  An automatic weapon has a great fire rate and thus gives rise to many fragments. Yes, the problem may not be noticeable if you only use a one automatic weapons and if the number of involved units is not high. But if player create a mission with several static automatic GL-turrets, several APC-Otokar ARMA, several MPAP Punisher(GMG), then the game is very much losing performance.  I don't mean a huge number of such units. Just add >=3 automatic units in the one battle and the problem can be visible.  I have a not weak PC, with CPU in 12 threads and 16 GB of RAM, but even I, in such moments notice a significant performance drop!

**Probably this problem can be solved by some tricks? Can be to add a smaller number of pieces for automatic weapons?  Or any other tricks.  Any solution to improve performance would significantly mitigate the problem.**

**Steps to reproduce:**
- place in the map (near big city) several automatic GL turrets (OPFOR)
- place in the map (near big city) several MRAPS (punisher) with GL turrets (OPFOR)
- place in the map (near big city) several APC (Otokar) with GL turrets (OPFOR)
- place several enemy Blufor vehicles (APC-Badger-IFV, MRAP-M-ATV(HMG) 
- go into battle

-  Singleplayer / Editor (Singleplayer)

**Placed Modules:**
- none

**RPT log file:**
- not needed

PabstMirror:
- Scales down frag creation count when busy
- Skip some tracking for full-auto GLs
- Smaller array for server event
- Removed some broken functions
- Renamed debug tracking functions

Close #5001


addons/frag/functions/fnc_addBlackList.sqf
```diff
@@ -1,5 +1,21 @@
+/*
+ * Author: Jaynus, NouberNou
+ * Adds a round to the blacklist (will be ignored)
```
jonpas: Missing final dot (in other new/moved/updated functions too).

PR #5014, opened 2017-03-21
Add custom CfgUIGrids for microDagr

### LINKED ISSUE #5012: Not possible to change microdagr size and position.
Arma 3 Version:1.68 stable
CBA Version:3.2.1.170227
ACE3 Version:3.9.0

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
Microdagr gps opens up very large at the center of the screen. Minimap element is no longer resizable in the arma layout options. Not possible to change microdagr size and position.

**Steps to reproduce:**
New clean installation of arma. Open the microdagr.

PabstMirror:
Fix #5012

Looks like in 1.68 they removed the gps in layout options.
This adds a custom layout just for the microDagr.


jonpas:
Should the same be done for Kestrel and other devices that are shown minimized?

PabstMirror:
This same method could work for anything.
But I'm not sure how much it's needed for the kestrel

jonpas:
It's really not I guess, doubt anyone uses it apart from when sniping with Advanced Ballistics and it's pretty much in the perfect place for that.

jonpas:
@dedmen wants to move the Kestrel too. 😆

PR #5017, opened 2017-03-22
French loc improvement

11RDP-LoupVert:
(French) text as it is clutters the interface interface, making it
uneasy for medic to use the it smoothly.
 

Hopefully suggested changes fixes that.

There were a few typos and, in my opinion, unaccurate or missing
translation.
Based on French Army Field Manual, actual paramedical personnel and past
work on XMedSys.

- Improve Medical usability in French version


11RDP-LoupVert:
How can I fix that Jon?

jonpas:
Your editor has to be changed to use spaces instead of tabs, then you have to mass convert it back (as your commit converts everything to tabs).

jonpas:
@alganthe could we get you to take a quick look at this?

alganthe:
French is quite verbose, the best way to avoid that issue is to use the line display instead of the circular one.
There's also the option to remove the verbs and simply add the name of the item that is to be used, it may lead to confusion.

jonpas:
Doesn't fix the root issue, but improves translations by shortening them and keeping them on point (alganthe's words), I'd say merge.

PR #5020, opened 2017-03-23
[Medical Rewrite] Adding respawn compatibility

Zakant:
- Make medical rewrite compatible with multiplayer respawn, by adding a respawn transition and XEH to handle respawn of a unit.
- Will also take care of resetting the medical values.


BaerMitUmlaut:
I'd recommend using the `CBA_statemachine_fnc_manualTransition` function to reset the unit's state instead. It keeps the state machine a little cleaner.

Zakant:
Ah okay, didn't know such a function existed. However I do not like to set the state from outside the state machine. I fear that this will lead to hard-to-find bugs when components start to move units around the state machine on their own. However if you insit i will change the code to work with `manualTransition`.

BaerMitUmlaut:
I totally agree that components usually shouldn't do that, it's definitely not ideal. However for such a full reset which doesn't happen regularly this is fine. It saves us 4 transition definitions cluttering the state machine, imo the benefits outweigh the downsides here.

Zakant:
Okay, I can agree with that. Will change it once I got time again.

Zakant:
Changed it to utilize `manualTrabsition`. Would appreciate if somebody could take a look if it's okay like that.

Arcanum417:
I cannot call `[ace_medical_STATE_MACHINE, player call ace_medical_fnc_getUnitState, "DEFAULT"] call CBA_statemachine_fnc_manualTransition;` from devconsole. I however can call `[ACE_player,ace_medical_STATE_MACHINE, player call ace_medical_fnc_getUnitState, "DEFAULT"] call CBA_statemachine_fnc_manualTransition;`

PR #5038, opened 2017-03-26
Optionals moved into launchable mod folders

### LINKED ISSUE #5036: Format Optional pbos to be loadable in A3 Launcher
**Arma 3 Version:** `1.68` (stable / rc / dev)
**CBA Version:** `3.2.1` (stable / dev + commit hash)
**ACE3 Version:** `3.9.1` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
As per chat on slack #dev
A3 launcher seems to search for any folder named `addons` within your mods folder (no matter the level).  This means that formatting the optional mods  for example `optionals/ACE_RHSUSF_Compat/addons` would show. The 3CB guys already do a basic form of this.

+No more messing with compat files every update.
+If you don't use the A3 launcher the mods are prepacked to copy and paste out
+ Offical compat mod names makes server listings easier
-Potential spam of mods list (may want to limit to popular ones) 

**Steps to reproduce:**
_For example_
1. Create File structure `optionals/ACE_RHSUSF_Compat/addons`
2. Add pbo's in.
3. Load launcher and see it in the list

- Launcher

john681611:
- `make.py` now builds optionals within folders that makes them launchable in the vanilla editor
- Fixes #5036


Glowbal:
If we wish to go down this road, optionals should probably become seperate release packages, like we have done with the RHS compats on Steam.

I'm not sure having a bunch of `@*` directories in the optionals folder is desirable, expect for maybe the default Arma3 launcher. I don't see it gaining us much, but I am happy for this to be discussed further.

john681611:
This does not stop the separate packages on steam at all. The `@` symbol isn't required it works without (I've added it so it works with publish). For those we decide not to publish to the variety of locations it will give players nice prepacked mods which means its less work for players even if they don't use the vanilla launcher.

bux:
TBH I'd like to see the optionals moved to separate repos.

Dystopian:
I vote for this. Managing optionals on every update is annoying.

jonpas:
Let's do this for 3.13.0 then if no other concerns are found. Don't want to break people's setups (especially where automated) shortly before release.

john681611:
*rises from grave
Any work needed to be done on this?

Dystopian:
no **john681611**, waiting for 3.13

jonpas:
@john681611 as @Dystopian said, no, it looks good, but we want to make doubly sure nothing will break with it when we want to release, and also warn people in advance that this will change.

commy2:
Merge now?

PR #5044, opened 2017-03-28
Update documentation for Vector21

Don42:
Add explanation on how to switch range and azimuth units


bux:
Perfect! Thanks @Don42

Issue #4978, opened 2017-03-08
Remove magic value occurances in the code base


thojkooi:
We have some magic value occurances in various places in our code base. 

One of such examples is #4964. 

We should replace those magic values by macro defines. This would be an ongoing issue.

commy2:
We almost exclusively use magic numbers without macros for IDC's in ACE:
https://github.com/acemod/ACE3/blob/98da86b74d4089bc3403d0b0c0074e6a46ce35d8/addons/optionsmenu/functions/fnc_onServerListBoxShowSelectionChanged.sqf#L33-L43

We also never use the DIK table for key codes:
`#include "\a3\ui_f\hpp\defineDIKCodes.inc"`

Mike-MF:
Closing due to age, inactivity or already solved. May be re-opened if the issue is still present, even better just open a new up-to-date Issue.

PR #5054, opened 2017-04-03
Increase max weapon index for large modsets

### LINKED ISSUE #5042: Weapon on Back and Holster dont work All Weapons
**Arma 3 Version:** `1.68` and sometimes 1.66 stable
**CBA Version:** `3.2.1` stable
**ACE3 Version:** `3.9.1` stable

**Mods:**

- CBA_A3
- ace
RHS AFRF and USA
SMA 
and many Other. 
All Weapons...

**Description:**
Frist Time its some one with Arma version 1.66 and the Last RHS and ACE versions come. The Problem comes in the time not always. But since the Last Arma3 Update 1.68 Always. 
I and my Clan try to take our Rilfe on the Back. but thats dont work all have Standard Key Config. we Press 0 but nothing happend.

**Steps to reproduce:**
1. Take a Rifle
2. Take a Pistol
3. Take a binocular
Do you have the Rifle in youre hand you press 0 "Nothing work" do you change to the Pistol and press 0 nothing happend. If you have a binocular, you can change with B thats workes normal but sometimes if you press 0 he Change to the binocular. But the First steps are more impotante. 

Its happend on All

**Placed Modules:**
its happend with and without Moduls on ALL maps. 

**RPT log file:**
Nothing Errors in RPT and Debug to find. No more Infos about that to find i am sorry.

PabstMirror:
Fix #5042

### DIFF
--- a/addons/respawn/functions/fnc_restoreGear.sqf
+++ b/addons/respawn/functions/fnc_restoreGear.sqf
@@ -44,7 +44,7 @@ if (!isNil "_activeWeaponAndMuzzle") then {
         private _index = 0;
 
         while {
-            _index < 100 && {currentWeaponMode _unit != _activeWeaponMode}
+            _index < 299 && {currentWeaponMode _unit != _activeWeaponMode}
         } do {
             _unit action ["SwitchWeapon", _unit, _unit, _index];
             _index = _index + 1;
--- a/addons/dragging/functions/fnc_startCarry.sqf
+++ b/addons/dragging/functions/fnc_startCarry.sqf
@@ -51,7 +51,7 @@ if (_target isKindOf "CAManBase") then {
 } else {
 
     // select no weapon and stop sprinting
-    _unit action ["SwitchWeapon", _unit, _unit, 99];
+    _unit action ["SwitchWeapon", _unit, _unit, 299];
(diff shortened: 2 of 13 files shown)


Dystopian:
#4978 doesn't agree

commy2:
It's not just macros. Lots of duplicated code. Could use https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_selectWeapon.sqf
?

addons/respawn/functions/fnc_restoreGear.sqf
```diff
@@ -44,7 +44,7 @@ if (!isNil "_activeWeaponAndMuzzle") then {
         private _index = 0;
 
         while {
-            _index < 100 && {currentWeaponMode _unit != _activeWeaponMode}
+            _index < 299 && {currentWeaponMode _unit != _activeWeaponMode}
```
dedmen: you added 200 to the other indexes. Why only add 199 here?

PabstMirror: Nothing special about the specific number, just needs to be bigger than biggest index.
This way they all use the same number `299`

Issue #5035, opened 2017-03-26
ace3 and infistar


neohlee:
**Arma 3 Version:** `x.xx` (stable / rc / dev)
**CBA Version:** `3.1.2` (stable / dev + commit hash)
**ACE3 Version:** `3.9.0` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- ace
-infistar
```

**Description:**
I installed ace3 on my server. I already had cba and other mods + infistar woking fine. But now i have this error : "BadButton on ESC: ACE Options (v0246)". And i probably guess it is what causes the infistar menu to not load. i cant use admin menu anymore.

**Steps to reproduce:**
Have a nice server with mods.
Try to enhanced the gameplay with ace3.
infistar admin menu not loading anymore.

**Where did the issue occur?**
- Dedicated 

my english is bad im sorry about it.
has any had this issue ? there must be a quick solution. i have seen many issue like this on forum. if anyone tells me how to fix this i can sharw with the french community.!
thanks in advance

PabstMirror:
RPT might help, but it sounds like the issue is from infistar.
I think you should be able to run without `ace_optionsMenu.pbo` if that's the cause of the conflict.

neohlee:
[gist](https://gist.github.com/neohlee/f20ccaff51c191f02148aee41dc39788)

neohlee:
i also tried running serv/client without ace_optionsmenu, same thing :(

commy2:
```
 7:25:45 Error in expression <isNull (findDisplay 46))};

if((__GETC__(life_medicLevel)) < 1) exitWith {
["Not>
 7:25:45   Error position: <(life_medicLevel)) < 1) exitWith {
["Not>
 7:25:45   Error Missing )
 7:25:45 File core\fn_initMedic.sqf [life_fnc_initMedic], line 71
 7:25:45 Error in expression <isNull (findDisplay 46))};

if((__GETC__(life_medicLevel)) < 1) exitWith {
["Not>
 7:25:45   Error position: <(life_medicLevel)) < 1) exitWith {
["Not>
 7:25:45   Error Missing )
 7:25:45 File core\fn_initMedic.sqf [life_fnc_initMedic], line 71
```

```
 7:25:43 Error in expression <aticModels setVariable [toLower (_array select ((count _array) - 2)), _cfgClass]>
 7:25:43   Error position: <select ((count _array) - 2)), _cfgClass]>
 7:25:43   Error Zero divisor
[...]

PR #5055, opened 2017-04-03
Fix tagging model cache

PabstMirror:
- postInit wasn't saving p3d name to cache (wrong index)
- other functions had inverted logic for exitWith (so it still worked)
- config class isn't used, so just save bool

Could possibly cause array bounds error if a building had an odd model path (see #5035)
```
 7:25:43 Error in expression <aticModels setVariable [toLower (_array select ((count _array) - 2)), _cfgClass]>
 7:25:43   Error position: <select ((count _array) - 2)), _cfgClass]>
 7:25:43   Error Zero divisor
```
If array only had 1 element you could get something like `["x"] select -1`

### DIFF
--- a/addons/tagging/XEH_postInit.sqf
+++ b/addons/tagging/XEH_postInit.sqf
@@ -16,7 +16,7 @@ for "_index" from 0 to (_countOptions - 1) do {
             private _model = getText (_cfgClass >> "model");
             if (_model != "") then {
                 private _array = _model splitString "\";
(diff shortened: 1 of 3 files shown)


commy2:
`param` can handle negative indices and reports default value or `nil` instead without error.

PabstMirror:
The problem was that the index was just wrong,
if `model = "\A3\Structures_F\Ind\PowerLines\HighVoltageColumn_F.p3d";`
it was picking `PowerLines` instead of the p3d filename.

PR #5063, opened 2017-04-08
Fix FCS Vehicle Init

PabstMirror:
Related to findings in https://github.com/CBATeam/CBA_A3/pull/628

vehicleInit.sqf runs several `setVariables [,,true]` that were not synced


commy2:
Not synched for vehicles created mid mission (not editor placed) for anyone trying to reproduce.

PR #5065, opened 2017-04-08
Lower minimum speed limiter speed to 5 km/h

SchmittAkaSmallfly:
Minimum speed limit of 10 km/h was needed in the past due to engine limitations. Multiple user tests have shown that the minimum speed is not needed anymore. The new minimum of 0 km/h allows for example setting walking speed for vehicles (<10 km/h).

### DIFF
--- a/addons/vehicles/functions/fnc_speedLimiter.sqf
+++ b/addons/vehicles/functions/fnc_speedLimiter.sqf
@@ -28,7 +28,7 @@ if (GVAR(isSpeedLimiter)) exitWith {
 playSound "ACE_Sound_Click";
 GVAR(isSpeedLimiter) = true;
 
-private _maxSpeed = speed _vehicle;
+private _maxSpeed = speed _vehicle max 3;
 
 [{
     params ["_args", "_idPFH"];
--- a/addons/vehicles/functions/fnc_speedLimiter.sqf
+++ b/addons/vehicles/functions/fnc_speedLimiter.sqf
@@ -28,7 +28,7 @@ if (GVAR(isSpeedLimiter)) exitWith {
 playSound "ACE_Sound_Click";
 GVAR(isSpeedLimiter) = true;
 
-private _maxSpeed = speed _vehicle; // "max 10" entfernt, da kein Grund mehr für Mindestgeschwindigkeit 10 km/h besteht
+private _maxSpeed = speed _vehicle;
 
 [{
     params ["_args", "_idPFH"];
--- a/addons/vehicles/functions/fnc_speedLimiter.sqf
+++ b/addons/vehicles/functions/fnc_speedLimiter.sqf
@@ -28,7 +28,7 @@ if (GVAR(isSpeedLimiter)) exitWith {
 playSound "ACE_Sound_Click";
 GVAR(isSpeedLimiter) = true;
 
-private _maxSpeed = speed _vehicle max 10;
+private _maxSpeed = speed _vehicle; // "max 10" entfernt, da kein Grund mehr für Mindestgeschwindigkeit 10 km/h besteht
 
 [{
     params ["_args", "_idPFH"];
--- a/addons/vehicles/functions/fnc_speedLimiter.sqf
+++ b/addons/vehicles/functions/fnc_speedLimiter.sqf
@@ -28,7 +28,7 @@ if (GVAR(isSpeedLimiter)) exitWith {
 playSound "ACE_Sound_Click";
 GVAR(isSpeedLimiter) = true;
 
-private _maxSpeed = speed _vehicle max 3;
+private _maxSpeed = speed _vehicle max 5;
 
 [{
     params ["_args", "_idPFH"];


jonpas:
> dedmen: What is the reason for the 10kmh limit?
commy2: Handbrake
commy2: PhysX Handbrake stops vehicles abruptly below 10 kmph

derbuur:
Please see also discussion on armaworld forum [https://armaworld.de/index.php?thread/2892-tempomat-f%C3%BCr-schrittgeschwindigkeit/&pageNo=1](url).
It seems that the original reason (Handbrake) is solved by BI so that the 10km/h limit is not longer necessary.

commy2:
What happens when you drive backwards after this PR?

derbuur:
Maybe change the minimum speed limit from 10 km/h to 5 km/h or 3 km/h to avoid any problems with driving backwards.

jonpas:
Don't speak German, so that forum is useless to me (and apparently everyone speaks badly enough for Google Translate to not make any sense of it). 😆

derbuur:
sorry jonpas, let me give you a brief result of the forum post.
We tested the minimal speed with following function:
`oneachframe {
_vehicle = vehicle player;
_maxSpeed = 1;
_speed = speed _vehicle;
if (_speed > _maxSpeed) then {
  _vehicle setVelocity ((velocity _vehicle) vectorMultiply ((_maxSpeed / _speed) - 0.00001));
 };};`
Six month ago the result of this function with slow speed was very bad and not playable, it was like a driver without knowledge of coupling.
But tested again a couple of days ago the result was very good. It seams that BI has changed the behavior of coupling and handbrake.
This is all written in the forum.

jonpas:
Ah, nice! Thank you.

SchmittAkaSmallfly:
If you feel more comfortable with min speed of 3 km/h you can of course change from 10 to 3. This also solves the problem I want to avoid with the pull request. With minimum 3 km/h it is easily possible to limit to walking speed (~6 km/h). Do you want me to generate a new pull request with minimum speed 3?

jonpas:
3 sounds good, you can just commit to the same branch, it will show on PR automatically.

SchmittAkaSmallfly:
Done. So what are the next steps? Who needs to approve this change?

commy2:
One of the maintainers of this component and the author of the original function, who incidentally assigned this pull request to himself - me.

commy2:
This pull request breaks the speed limiter for cars.
For example, enter a quadbike and set the speed limiter to 3 kmph. The driving is extremely choppy in curves and the worst thing is, that the engine triggers the brake sound.

SchmittAkaSmallfly:
Who ever wants to set 3 km/h speed limit for a quadbike?! Nobody in history has ever tried and nobody will ever try. Much more important is to support those players who want to synchronize tank and soldier speed when approaching cities. And if someone really tries to set it for quad bikes, he will notice that 3 km/h does not work really fine and this guy will raise his speed.

commy2:
Disagreed. The feature has to work in any circumstance. And as for tanks, the vanilla ones might not use brake sounds, but I know of certain ones that do.

SchmittAkaSmallfly:
so what about a new keyboard combination like SHIFT+DEL for limiting to any speed limit. this new keyboard combination will be hidden in the code and not being officially supported?

commy2:
Sounds like something you should put in your mission or community addon packet and not in ACE.

bux:
> The feature has to work in any circumstance.

I agree with @commy2.

SchmittAkaSmallfly:
Ok. But could you at least transform the current value 10 to a global variable, so that others have the chance to influence the script from outside? Hard coded numbers in a script are not the nicest way of coding so that you could solve two issues at the same time: 1) High quality code 2) Chance for others to influence minimum required speed in their environment

bux:
Could be a solution.

Though, generally speaking this might be true, but in this case it's utterly wrong:
> Hard coded numbers in a script are not the nicest way of coding

The reason why 10 was chosen is explained above.

SchmittAkaSmallfly:
As I said. Fine for me when 10 needs to be the official ACE value. But there should be some way to configure this minimum speed value.

PabstMirror:
What about something like this?
```
// Set `ace_vehicles_speedLimiterMin` in mission to change limit (may cause problems if set too low)
private _limit = missionNamespace getVariable [QGVAR(speedLimiterMin), 10];
private _maxSpeed = speed _vehicle max _limit;
```
Would let people change it via mission or a little addon.

jonpas:
In that case I'd just make it a setting to be honest.

PabstMirror:
If we make it a setting, then people can set it to values that could break things.
I would rather just make it a semi-hidden mission var.
If you need to change it, you still can.

SchmittAkaSmallfly:
By the way: I asked again what is the absolute minimum speed needed for tank/soldier groups. Minimum of 5 km/h would still be fine. So if the car issue that came up with 3 km/h is no problem with 5 km/h, we can easily solve the current issue by setting 5 km/h. Otherwise a semi-hidden mission var could do the job like @PabstMirror depicted.

SilentSpike:
Is there no minimum speed value that can be read from the vehicle's config?

commy2:
No, but I guess it's hard coded for the different simulation types.

PR #5067, opened 2017-04-09
Use getPosASLVisual for icons and camera position.

rebelvg:
- Fix stuttering of the camera and icons when camera is following the unit.
- Moves icons directly to the unit position.

It uses ```getPosASLVisual``` and ```ASLToAGL``` which simulates what ```modelToWorldVisual``` does but it updates properly. Tested this in the spectator we use in my community, works properly for any surface (water, underwater, above ground).

Example video.
Old - https://youtu.be/j7L0hg4POtY?t=0
New - https://youtu.be/j7L0hg4POtY?t=16

### DIFF
--- a/addons/spectator/functions/fnc_handleCamera.sqf
+++ b/addons/spectator/functions/fnc_handleCamera.sqf
@@ -64,7 +64,7 @@ if (GVAR(camMode) == 0) then {
     _vector = _vector vectorAdd [0,0,_distance*sin(-_tilt)];
 
     // Update the position of the target camera (used for smooth unit tracking)
-    _target camSetPos ((_unit modelToWorldVisual [0,0,0]) vectorAdd [0,0,1.5]);
+    _target camSetPos ((ASLToAGL getPosASLVisual _unit) vectorAdd [0,0,1.5]);
     _target camCommit 0;
 
     // Update the relative position of the unit camera
--- a/addons/spectator/functions/fnc_handleIcons.sqf
+++ b/addons/spectator/functions/fnc_handleIcons.sqf
@@ -29,7 +29,7 @@ _drawVehicles = [];
         _color = GETVAR(_x,GVAR(gColor),[ARR_4(0,0,0,0)]);
         _txt = groupID _x;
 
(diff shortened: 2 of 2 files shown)


commy2:
So you're saying that `modelToWorldVisual` is b0rked or used wrong?

rebelvg:
Some commands can't be used in per frame since BI fixed "teleporting" bug by introducing interpolation to the network code long time ago in arma 2 ao, that's why they came up with these getPosVisual and visiblePosition commands, they work properly if you want to do some per frame stuff.

commy2:
And why doesn't it work for `modelToWorldVisual`? Broken?

rebelvg:
No idea.

commy2:
```sqf
_unit modelToWorldVisual [0,0,0]
```
AGL from model center, render scope

```sqf
ASLToAGL getPosASLVisual
```
AGL from lowest road contact point, render scope

I don't think these are identical. What would be needed is a `getPosWorldVisual`, but that does not exist. Manually add the distance from model center to lowest land contact point? That could be determined once via `getPosASL select 2 - getPosWorld select 2`.
Refresher: `getPosWorld` reports PosASL from model center.

rebelvg:
I'm not sure what I'm doing wrong but they report 0 difference for ground units and flying units.
http://i.imgur.com/Hi37S7o.jpg

SilentSpike:
I admit I'm not currently familiar with the intricacies of this PR, but based on the video alone it seems like a good fix. However, it's also worth noting I'm currently undertaking a fairly significant refactor/overhaul of the spectator code based on BIs end game spectator code (which has already enlightened me to some things that can definitely be handled better - I suspect their method of icon rendering should also prove foolproof).

As for moving the icon to the exact position of the unit, personally I preferred it above the unit so that it isn't obscuring anything, but if this is preferable to most people I see no reason not to make the change.

SilentSpike:
This PR will be made redundant by #5171 (which I expect to have ready to merge fairly soon). However, it could be merged in as a temporary fix if a release occurs before that PR is merged.

PR #5069, opened 2017-04-10
AtragMx: BC G1 .338LM API526

Laid3acK:
- Update the BC G1 for the API526

Lapua  API526:  BC G1 approximately 0.56
http://www.lapua.com/upload/tuotteet/specialpurpose/lapuaspecialpurpose2011eng.pdf#page=2

In-game, as close as possible the ICAO conditions, the BC seems to be near 0.58 but 0.56 gives some better results.
(maybe a typo error : 0.76 / 0.56 ?)


ulteq:
You are right. It seems to be a typo and the ideal C1 coefficient is `0.58`.

Laid3acK:
Glad to see you around. :)

PR #5070, opened 2017-04-10
Filter headless clients from zeus teleportation

Cuel:
Title.


commy2:
>Title

It catches on! :)

PR #5073, opened 2017-04-11
tweak ACE main menu info box

commy2:
- simplified config by using inheritance from other buttons
- macro defines for IDC's
- fixed no spacing between our button and the vanilla ones
- added hover highlighting like the other buttons
- fixed how version number is generated (old method: 3.10.X would be shown as "3.10.")
- changed url to link directly to the /releases/ page of github.

Pabst, you made this originally. Please review. Thanks.

### DIFF
--- a/addons/optionsmenu/config.cpp
+++ b/addons/optionsmenu/config.cpp
@@ -22,7 +22,6 @@ class CfgAddons {
     };
 };
 
-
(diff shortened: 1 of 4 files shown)


bux:
btw. you could use `https://ace3mod.com` now.

PR #5083, opened 2017-04-15
Spectator Improvements

### LINKED ISSUE #5081: Spectator public functions are not pre-initialisation safe
Should be a simple fix, functions `updateVisionModes` and `updateCameraModes` are not safe to run before settings have initialised as the GVARs they manipulate are set to defaults upon setting initialisation (overriding any changes applied).

Need to simply use the current system of adding to `ace_common_runAtSettingsInitialized` if used before `ace_common_settingsInitFinished`.

Brought to my attention by user anfo in slack :+1:

SilentSpike:
- Fix #5081 
- Disable searching on the spectator unit tree
- Add stringtables for the eden respawn template dialog
- Optimise some spectator functions

~~This can be merged now, but I'll leave it open for a little bit in case anyone wants to review.~~
Forgot the new system requires a review from someone else :+1: It's been a while :laughing:

### DIFF
--- a/addons/spectator/config.cpp
+++ b/addons/spectator/config.cpp
@@ -21,7 +21,9 @@ class CfgPatches {
 
 class CfgRespawnTemplates {
     class ADDON {
+        displayName = CSTRING(DisplayName);
         onPlayerKilled = QFUNC(respawnTemplate);
         onPlayerRespawn = QFUNC(respawnTemplate);
+        respawnTypes[] = {2,3};
     };
 };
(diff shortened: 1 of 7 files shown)


addons/spectator/config.cpp
```diff
@@ -21,7 +21,9 @@ class CfgPatches {
 
 class CfgRespawnTemplates {
     class ADDON {
+        displayName = CSTRING(DisplayName);
         onPlayerKilled = QFUNC(respawnTemplate);
         onPlayerRespawn = QFUNC(respawnTemplate);
+        respawnTypes[] = {2,3};
     };
```
commy2: maybe disable conflicting templates in the eden ui?

SilentSpike: Wasn't aware there is a setting for that, will look into it

commy2: There isn't afaik, but that has never stopped me. I just don't know which modes if any.

Issue #5090, opened 2017-04-18
CfgAiSkill.hpp is limiting AI skill


xJenny99x:
**Arma 3 Version:** `yes` (stable / rc / dev)
**CBA Version:** `yes` (stable / dev + commit hash)
**ACE3 Version:** `yes` (stable / dev + commit hash)

**Mods:**
```
yes
```

**Description:**
- so this file limits ai skill to a certain degree. And I think ai skill is definitely not a mod specific setting. It's mission specific and should remain like it.

**Steps to reproduce:**
- Play with _unit setSkill 1; 
- Ai can't hit shit and are dumb.

**Where did the issue occur?**
- ACE3/addons/ai/CfgAISkill.hpp

**Placed Modules:**
- none

**RPT log file:**
no?

xJenny99x:
With the new AI module for zeus that ace3 introduces, it should not force shitty ai through this config.

xJenny99x:
I've just confirmed this. I had my ai `_unit setskill ["aimingAccuracy",0.9];`

Before chaning the CfgAiSkill the ai would not kill me in 10 seconds (300 m) I was able to kill a whole squad (10 units):

Now I change the cfgaiskill and remove limits. Now with same skill lvl, the ai kill me in ~3 seconds. This will allow me as a mission maker to better customize my missions. 

Btw as reference with default ace3 config, we took ~18 players and killed ~200 ai with `_unit setskill ["aimingAccuracy",0.9];` in 1 hour and no players died.

This is really bad and boring and forces mission makers to use 100s of AI to make it challanging. PLEASE FIX

PR #5091, opened 2017-04-18
Remove CfgAiSkill config changes

PabstMirror:
Ref #5090 #4326

`general[]        = {0,0, 1,0.9};`
is something like `linearConversion [0, 1, skill, 0, 0.9];`

The benefit is that it makes the default AI a little less skilled.
The negative is that it hard locks the range skill can be set to via command.

e.g. even if you do `setSkill ["General", 1]`
`skillFinal "general"` will always be less than 1

I think this should be removed and left to dedicated AI mods to change.
We also already have a zeus module that allows easy changing of skill.


PabstMirror:
We could also re-add this feature via SQF
The effect should be the same, but it could be disabled via settings.
```
if (GVAR(adjustSkills)) then {
    ["CAManBase", "init", {
        params ["_unit"];
        if (local _unit) then {
            {
                _unit setSkill linearConversion [0, 1, _unit skill _x, 0, 0.9];
            } forEach [Skills...]
        };
    }, nil, nil, true] call CBA_fnc_addClassEventHandler;
};
```

Cuel:
Imo modifying CfgAiSkill does not belong with ACE. It's for AI overhaul mods. It's already hard enough to work with vanilla AI, without mods playing a part in the skill modification

docs/wiki/feature/ai.md
```diff
@@ -14,21 +14,16 @@ version:
 
 ## 1. Overview
 
-### 1.1 Adjusted AI skill values
-The idea here is to reduce the AI's godlike aiming capabilities while retaining its high intelligence. The AI should be smart enough to move through a town, but also be 'human' in their reaction time and aim.
```
Glowbal: On the feature page, we should probably mention that this feature was dropped in version 3.10.

Issue #5093, opened 2017-04-20
G Forces apply to uav units


Zakant:
**Arma 3 Version:** `stable` , tested on `dev`
**CBA Version:** `latest stable`
**ACE3 Version:** `latest sable`

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
G Forces apply to uav controlled units as well. This means the uav ai could have a black or red out.

**Steps to reproduce:**
- Place a uav plane (new uvac is the best, as it have the highest top speed)
- Remote contoll the uav either from zeus or per UAV Terminal
- Fly a tight corner at high speed
- Get a black out

**Where did the issue occur?**
- Everywhere 

**RPT log file:**
Not relevant for this.

commy2:
They are poor quality Chinese knock-offs.

SilentSpike:
:rofl: Should be an easy fix

Zakant:
Question who is faster... I already tried a few things... So I could probably provide a fix tomorrow. Unless you want to fix it ;)

PabstMirror:
PR to fix: #5094

PR #5094, opened 2017-04-21
Disable g-forces for UAV AI

PabstMirror:
A zeus can remote control UAV
Doing so means that typeOf ace_player = B_UAV_AI
(has some other weird effects)

This exits the gfroce pfeh if the unit is virtual or AI

### DIFF
--- a/addons/gforces/XEH_postInit.sqf
+++ b/addons/gforces/XEH_postInit.sqf
@@ -3,22 +3,26 @@
 if (!hasInterface) exitWith {};
 
 GVAR(pfID) = -1;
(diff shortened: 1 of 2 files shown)


Zakant:
How about moving the virtual unit detection to a more central component such as ace_common? Only one unit event handler would be needed an all components could grab the data. I think it would be useful for a few other cases as well.

commy2:
So #5093 only applies when remote controlling the drone, not when managing it as drone operator?

Zakant:
Can confirm. Just happening from zeus. UAV operators are uneffected.

PR #5095, opened 2017-04-21
Add keybind to toggle interaction menu

PabstMirror:
@HeadTraumaZA
Ref https://github.com/acemod/ACE3/issues/3594#issuecomment-296169526

Adds seperate keybind that just toggles the interaction menu.
When combined with the "Do action when releasing key: No" setting, it should be much easier to use.

Should we add new keybinds, or just add a setting to change behaviour of the existing keybinds?


CorruptedHeart:
Considering Arma's method of having toggle as a separate keybind that would be consistent with Arma.

rebelvg:
Tried this, pressing esc on self-interaction hides the menu but leaves the black-screen. In interaction it just opens esc-menu. Would be nice if esc just closed the action menu.

Issue #3095, opened 2015-12-31
Pain not shown in the optical mode and when map is open 


ghost:
ACE3 Version: 3.4.1
Stable 1.54

**Mods:**
- `@CBA_A3`
- `@ace`

**Placed ACE3 Modules:**
Advanced and Basic Medical modules

**Description:**
Pain(Colour flashing) not shown in the optical mode and when map is open.
Also inside vehicles!

**Steps to reproduce:**
Get wound and to use optic or map

**Where did the issue occur?**
Editor

**RPT log file:**
Not needed

mickeymen:
Also inside vehicles I noticed this, when I use Otokar ARMA

Glowbal:
Will be done after #3916 and #3917

Glowbal:
Possible solution is to close the map on hit.

This can possibly be integrated into hit reactions.

Mike-MF:
Closing due to age, inactivity or already solved. May be re-opened if the issue is still present, even better just open a new up-to-date Issue.

PR #5099, opened 2017-04-22
hitreactions - Close map on hit

PabstMirror:
#3095


madpat3:
maybe you should also do this for, when controlling a drone.

PR #5100, opened 2017-04-22
Med rw handle fire damage

PabstMirror:
Redo #4223 for rewrite


commy2:
I think 0.15 is too high for this.

PR #5107, opened 2017-04-26
Fix jerry can refuel interaction overlapping

### LINKED ISSUE #5106: Interaction icon overlap for object with custom fuel AND loadable; also custom fuel object location
**Arma 3 Version:** 1.68 stable
**CBA Version:** 3.2.1 stable
**ACE3 Version:** 3.9.1 stable

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
When using both Refuel and cargo loading functions on a single custom object, the Refuel and Interaction icons overlap at close range to the object, with Interaction taking priority so that load/carry show up, but refuel is inaccessible. I suggest it would be more uniform to format Refuel as an additional option alongside Load and Carry, branching from Interaction, instead of competing with Interaction.

However, when staying slightly farther away from the object, Refuel option becomes available about a meter farther away, allowing that action before getting close enough to be able to load/carry. Inconsistent UI distance may need to be addressed for uniformity/standardization of interaction.

Disregard and delete this issue if that is a design feature and not a bug or inconsistency.

I added only custom refuel init to a metal barrel. Was able to refuel vehicles with it (barrel looks funny, sticking out of vehicle at a 45 degree angle), no problem.

Added cargo loading ability separately and individually via ace_cargo_fnc_makeLoadable and synced cargo loadable ACE module. Both work as expected via Interaction > Load.

(Ideally wanted to do so via config, but cargo framework documentation was not very clear.)

Also added carrying ability to barrel via ace_dragging_fnc_setCarryable. Worked as expected via Interaction > Carry.

But adding both custom fuel and any one or more normal Interaction choices cause Interaction icon to superimpose on and block access to Refuel.

**Steps to reproduce:**
- In 3den editor, place an object Land_MetalBarrel_F
- Also place a vehicle w virtual cargo enabled (HEMTT box or transport)
- In init:
[this, true, 2] call ace_cargo_fnc_makeLoadable; [this, true] call ace_dragging_fnc_setCarryable; [this, 200] call ace_refuel_fnc_makeJerryCan;
- In mission, interact with barrel (Refuel
-Note Refuel option, then note superimposed Interaction and Refuel options.
- Additionally, when refueling the HEMTT, barrel placement can be hilariously wonky or buried inside the HEMTT. Finding the right spot to "attach nozzle" near the driver side fuel tank can be tricky. Just an FYI. This is to be expected with a custom fuel object.

- Editor (Multiplayer)

**Placed Modules:**
- Occurs with both no modules or with ACE Make Object Loadable module.

jonpas:
- Fix #5106 
- Add jerry can actions under `"ACE_MainActions"`, mimicking the standard Refuel actions on vehicles.

### DIFF
--- a/addons/refuel/functions/fnc_makeJerryCan.sqf
+++ b/addons/refuel/functions/fnc_makeJerryCan.sqf
@@ -36,7 +36,7 @@ private _action = [QGVAR(Refuel),
     [],
     [0, 0, 0],
     REFUEL_ACTION_DISTANCE] call EFUNC(interact_menu,createAction);
-[_target, 0, [], _action] call EFUNC(interact_menu,addActionToObject);
+[_target, 0, ["ACE_MainActions"], _action] call EFUNC(interact_menu,addActionToObject);
 
 // Add pickup
 _action = [QGVAR(PickUpNozzle),
@@ -48,7 +48,7 @@ _action = [QGVAR(PickUpNozzle),
     [],
     [0, 0, 0],
     REFUEL_ACTION_DISTANCE] call EFUNC(interact_menu,createAction);
-[_target, 0, [QGVAR(Refuel)], _action] call EFUNC(interact_menu,addActionToObject);
+[_target, 0, ["ACE_MainActions", QGVAR(Refuel)], _action] call EFUNC(interact_menu,addActionToObject);
 
 // Add turnOn
 _action = [QGVAR(TurnOn),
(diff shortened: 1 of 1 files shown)


jonpas:
Also, this really needs a rework like refuel has, to dynamically add actions, would make it much easier to work with and also not require insane setup for custom refuel vehicles (I still have plans!).

PR #5113, opened 2017-04-29
Tweak randomness for wound bleeding and pain

PabstMirror:
Limits the range of randomness for wound bleeding/pain
I feel like it was way to random before, and most shots ended up providing very little bleeding (~0.01). You could also end up with 0.2 damage causing more bleeding than a 20 damage hit depending on the dice rolls.

I do see some use for randomness (e.g. a .22lr in an artery is going to do serious damage)
But I think we want some consistency in this as well.

image


ulteq:
Two things that come into my mind:
* `_bloodiness` and `_painfullness` were designed to be between 0 and 1
* `_pain` was not supposed to add up anymore, instead we used `_painLevel max _pain`

Hard to tell if this PR breaks anything without actually testing it.
But I guess it is a good idea to make the bleeding rates more consistent.

PabstMirror:
We can tweak these number later based on feedback, but I thinking decreasing the range is going to useful for testing.
The pain only sums on multipart wounds (grenades that make several small wounds)

PR #5115, opened 2017-04-30
Explosives - Add code handler for detonation

### LINKED ISSUE #5101: Add ability to disable remote detonation for explosives
Hi guys!

I'm currently working on a jammer script that should add the ability to jam remote detonation of explosives/IEDs in a certain radius around a jammer vehicle. Unfortunately either I can't find a function to disable remote detonation (without actually defusing or detonating the explosive) or there actually isn't any.

That might be a feature to add, so jammers can be made.

PabstMirror:
Close #5101

Allows mission/addon makers to block detonations in the explosive system.
See example in addDetonateHandler.sqf

### DIFF
--- a/addons/explosives/XEH_PREP.hpp
+++ b/addons/explosives/XEH_PREP.hpp
@@ -2,6 +2,7 @@
 PREP(addCellphoneIED);
 PREP(addClacker);
 PREP(addDetonateActions);
+PREP(addDetonateHandler);
 PREP(addExplosiveActions);
 PREP(addToSpeedDial);
 PREP(addTransmitterActions);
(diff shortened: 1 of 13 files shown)


bux:
Having the examples in the wiki would be splendid 😄

CorruptedHeart:
During testing it seems that cellphone triggers don't trigger the DetonateHandlers. Investigating further.

CorruptedHeart:
Okay, I found what the issue is. Timer based explosives and Cellphone explosives use -1 as range. The example code has a check for range initially and allows the explosion if range is < 0.

So I guess maybe just change the example code to have the cellphone check before the initial range check.

PR #5116, opened 2017-04-30
Add Paradrop Time Coefficient Setting

654wak654:
- Add a new setting so it's possible to modify how long it takes for cargo to unload while paradropping. If the setting is 0 the progress bar is skipped and cargo is dropped immediately.
- Show how long it's going to take to unload something in the list dialog by it's name.

Counter point to any realism concerns people might have (even though it's a setting using the current value as default): IRL the aircraft's crew would have to unstrap the cargo container, slowly and safely move it to the ramp area, attach the drop hook etc... I know that. This setting is so _filthy casuals_ don't have to make 2 minutes of flight planning and fail the first attempt anyway because the random ammo box zeus chose that day took 17.8 seconds to drop, and there was no way for the pilot to know that.

### DIFF
--- a/addons/cargo/ACE_Settings.hpp
+++ b/addons/cargo/ACE_Settings.hpp
@@ -6,4 +6,11 @@ class ACE_Settings {
         value = 1;
         category = ECSTRING(OptionsMenu,CategoryLogistics);
     };
+    class GVAR(paradropTimeCoefficent) {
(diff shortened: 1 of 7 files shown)


jonpas:
I like this, dropping something should be able to be timed, in defense of realism, this is what this PR does, IRL you would know when something would be dropped and plan accordingly.

PR #5119, opened 2017-05-01
Fix refuel nozzle dropping to ground

### LINKED ISSUE #5112: Fuel nozzle drops at terrain level / in sea
**CBA Version:** `3.2.1` (stable)

**Mods:**
```
- CBA_A3
- ACE
- CUP
- other stuff
```

**Description:**
Fuel nozzle is dropped at terrain level. On a carrier or mission build base above terrain level the fuel nozzle cannot be found because it is at terrain level. Therefore the fuel nozzle cannot be returned and a zeus must copy/paste the fuel truck to work again.

**Steps to reproduce:**
- Place carrier / build base above terrain level
- Take fuel nozzle
- Drop fuel nozzle / move away from truck
- Fuel nozzle cannot be found, because it is at terrain level

- Dedicated 

**Placed Modules:**
- Many: https://wiki.tacticalteam.de/index.php?title=Modulvorgaben#Serverseitig_gesetzt

**RPT log file:**
- None

**Code:**
https://github.com/acemod/ACE3/blob/master/addons/refuel/functions/fnc_dropNozzle.sqf#L28
The height / z position should be relative to the carrying unit, shouldn't it?

PabstMirror:
Fix #5112

### DIFF
--- a/addons/refuel/functions/fnc_dropNozzle.sqf
+++ b/addons/refuel/functions/fnc_dropNozzle.sqf
@@ -19,13 +19,26 @@
 #include "script_component.hpp"
 
 params [["_unit", objNull, [objNull]], ["_nozzle", objNull, [objNull]], ["_disconnectOnly", false, [false]]];
+TRACE_3("dropNozzle",_unit,_nozzle,_disconnectOnly);
 
 detach _nozzle;
 _nozzle setVariable [QGVAR(isRefueling), false, true];
 
 if (_disconnectOnly) exitWith {};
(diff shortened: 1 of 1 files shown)


jonpas:
Should make this a common function maybe? Same is used in dropping carried items right?

PabstMirror:
https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_findUnloadPosition.sqf
is for cargo and people unloading
but that is fairly complex because it need to avoid collisions
the gas handle is pretty small and shouldn't be a physx problem

PR #5144, opened 2017-05-10
Handle escape menu with toggle iteractMenu

### LINKED ISSUE #5136: Toggle interaction menu and esc button issues
**ACE3 Version:** `3.9.1` (dev + c001be5b09d749737551acf281de70d5c8c32073)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- With cursor enabled: pressing esc hides the menu (if show cursor is enabled) but leaves the black-screen background (if enabled, of course). Without cursor enabled: esc just opens esc-menu. Would be nice if esc just closed the action menu in both cases.

**Steps to reproduce:**
- Set to toggle. Open menu. Press esc.

**Placed Modules:**
- None.

**RPT log file:**
- N/A.

PabstMirror:
Close #5136

### DIFF
--- a/addons/interact_menu/functions/fnc_handleEscapeMenu.sqf
+++ b/addons/interact_menu/functions/fnc_handleEscapeMenu.sqf
@@ -1,0 +1,25 @@
+/*
+ * Author: PabstMirror
+ * Handle the escape key being pressed
+ *
+ * Arguments:
+ * 0: Escape menu display that should be closed <DISPLAY><OPTIONAL>
+ *
+ * Return Value:
+ * Nothing
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+if (GVAR(openedMenuType) < 0) exitWith {};
+
+params [["_display", displayNull, [displayNull]]];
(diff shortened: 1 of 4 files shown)


rebelvg:
Just tried this. Interaction with cursor on closes properly now, but interaction without curson still opens esc menu.

PabstMirror:
This should work on both display options now.


There is a tiny-micro stutter when hitting escape because it partialy loads the escape menu before closing, not ideal but it's barley noticeable.

rebelvg:
Yep, seems to work.
Strange behavior can be noticed when you open interaction with cursor and while it's open you try to open interaction without cursor. It will kinda close previous interaction but won't open other interaction and still have black overlay enabled with the cursor.

addons/interact_menu/functions/fnc_handleEscapeMenu.sqf
```diff
@@ -0,0 +1,25 @@
+/*
+ * Author: PabstMirror
+ * Handle the escape key being pressed
```
jonpas: Missing final dot.

PR #5146, opened 2017-05-10
Update helmet hearing protection for RHS-US

PabstMirror:


bux:
The part about
>amplifies weak sounds

is so very true and quite amazing once you've tried a ComTac for the first time.

PabstMirror:
Do all peltors have this tech?

In any case we should look into standardizing the values.
Earplugs offer the equivalent of 
```
        GVAR(protection) = 0.75;
        GVAR(lowerVolume) = 0.5;
```
but some helmet `H_Cap_headphones` has
```
        GVAR(protection) = 0.5;
        GVAR(lowerVolume) = 0.6;
```

bux:
> Do all peltors have this tech?

It seems so. I can see the external mics on the first generation too.

> but some helmet H_Cap_headphones has

yep, that seems like a mistake

bux:
> In any case we should look into standardizing the values.

I'd suggest something like:

```c++
    GVAR(protection) = 0.7;
    GVAR(lowerVolume) = 0.0;
```

PabstMirror:
@bux - Let me know if you can ID these headsets
http://i.imgur.com/6R2HFgq.png

bux:
First two are Peltors. The last one is unknown to me.

addons/hearing/CfgWeapons.hpp
```diff
@@ -13,42 +13,42 @@ class CfgWeapons {
         };
     };
 
+    #define HEARING_PROTECTION_VICCREW EGVAR(hearing,protection) = 0.85; EGVAR(hearing,lowerVolume) = 0.6;
+    #define HEARING_PROTECTION_EARMUFF EGVAR(hearing,protection) = 0.75; EGVAR(hearing,lowerVolume) = 0.5;
+    #define HEARING_PROTECTION_PELTOR EGVAR(hearing,protection) = 0.75; EGVAR(hearing,lowerVolume) = 0.2;
```
bux: I'd reduce the `lowerVolume` value to `0.0` because of the sound enhancing features / microphones

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -87,24 +87,20 @@ class CfgWeapons {
         ace_reloadlaunchers_enabled = 1;
     };
 
+    #define HEARING_PROTECTION_VICCREW EGVAR(hearing,protection) = 0.85; EGVAR(hearing,lowerVolume) = 0.6;
+    #define HEARING_PROTECTION_EARMUFF EGVAR(hearing,protection) = 0.75; EGVAR(hearing,lowerVolume) = 0.5;
+    #define HEARING_PROTECTION_PELTOR EGVAR(hearing,protection) = 0.75; EGVAR(hearing,lowerVolume) = 0.2;
```
bux: same here

bux:
These behave the same as Peltors
a
_MSA Sordin_

Drofseh:
With ear muffs that do audio pass through from external microphones (like the Peltors or those MSA Sordins), when the the circuit detects a loud noise (82 db on a pair I have) like a gunshot it cuts the audio pass through and then acts like normal ear muffs for 0.25 seconds or so.

I don't know if it's worth doing, but it would be neat to see that simulated as well.

bux:
@Drofseh that's not part of this PR
This should go into the Feature Request issue

Drofseh:
> This should go into the Feature Request issue
Yah fair enough

PR #5147, opened 2017-05-10
hide 1.71 mine detector display

commy2:
- All our items are mine detectors and removing this one is no loss


jonpas:
Fixed by using `CBA_MiscItem` thanks to CBA and @PabstMirror.

PR #5148, opened 2017-05-10
hide 1.71 mine detector actions

commy2:
- Alternative to #5147. Hide the actions instead.

### DIFF
--- a/addons/common/CfgActions.hpp
+++ b/addons/common/CfgActions.hpp
@@ -1,0 +1,12 @@
+
+// hide mine detector actions
+class CfgActions {
+    class None;
+    class NextModeRightVehicleDisplay: None {
+        show = 0;
+    };
+
+    class ListRightVehicleDisplay: None {
+        show = 0;
+    };
+};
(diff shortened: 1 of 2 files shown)


CorruptedHeart:
Will this not affect the new airplane UI stuff?

jonpas:
As far as I understand this config, it will remove both left and right action menu entries, still allowing you to use them via keybinds, but that has a huge downside of those actions not being available inside vehicles (eg. airplanes) where they should be.

commy2:
I thought these are for the mine detector. All they do is bring up that thing and I have the actions outside of vehicles.

jonpas:
They are used for many things starting with new sensor stuff all the way to mine detectors. Thre are basically 2 screens you can select what you want displayed on.

BlueNexus:
I strongly recommend not merging this, due to the fact that it will make accessing critical parts of the new plane UI difficult.

jonpas:
Yeah, no worries, this won't get merged. :)

PR #5150, opened 2017-05-10
Add Arma 3 Issues breaking ace_rearm

Dystopian:
ACE rearm has some issues like it can't rearm vehicle with different localities of vehicle and turret, or it can't fully rearm turret with multiple magazines of the same type. It's all because of mentioned Arma 3 issues.


PabstMirror:
ace_rearm has 3 levels, do you know if it effects all of them?

Dystopian:
I didn't test all of them, at least "Entire Vehicle" level is impacted.

PR #5151, opened 2017-05-10
Refuel for non-AllVehicle objects

tbeswick96:
- Add helper.p3d from fastroping
- Create helper object for non-AllVehicles objects, allowing refuel to be configured for more stuff (sling-loadable barrels for example)
- Un-magic refuel progress timer value

This adds a workaround which makes it possible to have a fuel hose (an arma rope) coming from a non-physx object (limitation of ropeCreate) by using an invisible helper object which the rope is connected to instead - such that it looks like it connects to the actual object.

This has been used for sling-loadable fuel barrels for a number of weeks with no issues. These can be found here: https://github.com/uksf/modpack/blob/master/addons/vehicles/CfgVehicles.hpp#L102

### DIFF
--- a/addons/refuel/CfgVehicles.hpp
+++ b/addons/refuel/CfgVehicles.hpp
@@ -512,6 +512,19 @@ class CfgVehicles {
         GVAR(fuelCargo) = REFUEL_INFINITE_FUEL;
     };
 
+    // Helper object for non-AllVehicles objects
+    class GVAR(helper): Helicopter_Base_F {
+        scope = 1;
+        displayName = "Refuel Helper";
+        model = "\A3\Weapons_f\empty";
+        delete ACE_Actions;
+        damageEffect = "";
+        destrType = "";
+        class HitPoints {};
+        class Turrets {};
+        class TransportItems {};
+    };
+
     /* // Barrels found in config  \
         BarrelHelper: Misc_thing 100
         BarrelBase: BarrelHelper 100
--- a/addons/refuel/CfgVehicles.hpp
+++ b/addons/refuel/CfgVehicles.hpp
@@ -517,7 +517,10 @@ class CfgVehicles {
         scope = 1;
         displayName = "Refuel Helper";
         model = "\A3\Weapons_f\empty";
-        delete ACE_Actions;
+        class ACE_Actions {};
+        class ACE_SelfActions {};
+        EGVAR(cargo,hasCargo) = 0;
+        EGVAR(cargo,space) = 0;
         damageEffect = "";
         destrType = "";
         class HitPoints {};
--- a/addons/refuel/functions/fnc_takeNozzle.sqf
+++ b/addons/refuel/functions/fnc_takeNozzle.sqf
@@ -39,7 +39,7 @@ if (isNull _nozzle) then { // func is called on fuel truck
         _endPosOffset = _endPosOffset select 0;
     };
     [
-        2,
+        REFUEL_PROGRESS_DURATION,
         [_unit, _target, _endPosOffset],
         {
             params ["_args"];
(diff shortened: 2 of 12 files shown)


PabstMirror:
I think this should let us fix #3310

tbeswick96:
Not sure if damage from collisions would be an issue. Didn't BIS disable collisions of attachTo'd objects recently?

tbeswick96:
arma3_x64 2017-05-10 23-01-04

addons/refuel/CfgVehicles.hpp
```diff
@@ -512,6 +512,16 @@ class CfgVehicles {
         GVAR(fuelCargo) = REFUEL_INFINITE_FUEL;
     };
 
+    // Helper object for non-AllVehicles objects
+    class GVAR(helper): Helicopter_Base_F {
+        author = "Beswick.T";
+        scope = 1;
+        model = QPATHTOF(data\helper.p3d);
```
commy2: Is this an actual model or just an empty dummy? One can use `model = "-";` instead for no model.

tbeswick96: It's an empty dummy. Tried and it doesn't work with "-", just gives an error.

tbeswick96:
Have done a little more playing around, and although it fixes the issue for things like the Huron fuel container, it doesn't fix it for buildings like the fuel bladder. It seems when the helper is attached to the fuel bladder, it loses simulation. Detach it and it's fine. Unsure how to get around that...is there a way to make an object static in a position without disabling simulation?

tbeswick96:
Think I've solved it!
arma3_x64 2017-05-11 02-04-38

If the simulation type is thingX, which handles attachTo just fine with rope simulation, it will use that, otherwise it will set the help position, direction, and vectorUp to the same as the target's, resulting in proper positioning. This should be fine, as it's not everyday a house or fuel pump gets up and walks away.

addons/refuel/CfgVehicles.hpp
commy2: What about:
```
model = "\A3\Weapons_f\empty";
```
?

tbeswick96: Will give that a shot 1 min

tbeswick96: That seems to work juuuust fine

bux:
Could you explain a little more in your OP what this does and what's being improved?
I'd like to know more.

And it would be helpful when writing the change logs.

addons/refuel/functions/fnc_takeNozzle.sqf
```diff
@@ -56,7 +56,14 @@ if (isNull _nozzle) then { // func is called on fuel truck
             private _ropeTarget = _target;
             if (!(_target isKindOf "AllVehicles")) then {
                 private _helper = QGVAR(helper) createVehicle [0,0,0];
-                _helper attachTo [_target, [0,0,0]];
+                hideObjectGlobal _helper;
```
SilentSpike: Keep in mind this has to run on the server, there is an event in common that can handle the locality:

`[QEGVAR(common,hideObjectGlobal), [_helper, true]] call CBA_fnc_serverEvent;`

addons/refuel/functions/fnc_takeNozzle.sqf
```diff
@@ -53,11 +53,22 @@ if (isNull _nozzle) then { // func is called on fuel truck
...
             _unit setVariable [QGVAR(nozzle), _newNozzle, true];
 
-            if (_target isKindOf "AllVehicles") then {
-                // Currently ropeCreate requires its first parameter to be a real vehicle
-                private _rope = ropeCreate [_target, _endPosOffset, _newNozzle, [0, -0.20, 0.12], REFUEL_HOSE_LENGTH];
-                _newNozzle setVariable [QGVAR(rope), _rope, true];
+            private _ropeTarget = _target;
+            if (!(_target isKindOf "AllVehicles")) then {
+                private _helper = QGVAR(helper) createVehicle [0,0,0];
+                hideObjectGlobal _helper;
```
SilentSpike: Keep in mind this has to run on the server, there is an event in common that can handle the locality:

`[QEGVAR(common,hideObjectGlobal), [_helper, true]] call CBA_fnc_serverEvent;`

SilentSpike:
To give @bux some context: This adds a workaround which makes it possible to have a fuel hose (an arma rope) coming from a non-physx object (limitation of `ropeCreate`) by using an invisible helper object which the rope is connected to instead - such that it looks like it connects to the actual object.

PabstMirror:
Pushed some config changes to disable actions on the helper
I tested this in 2 player mp and it seemed to be working fine

IIRC it's a little dangerous to `setVariable` on map objects because they are technically "streaming".
And might be deleted and come back with no variables.
But I don't think it would cause any serious issues, and the functionality gains are worth it.

addons/repair/CfgEventHandlers.hpp
```diff
@@ -37,7 +37,7 @@ class Extended_InitPost_EventHandlers {
     class Helicopter {
         class ADDON {
             init = QUOTE(_this call DFUNC(addRepairActions));
-            exclude[] = {QEGVAR(fastroping,helper), "ACE_friesBase"};
+            exclude[] = {QEGVAR(fastroping,helper), "ACE_friesBase", QEGVAR(refuel,helper)};
```
jonpas: I believe this fastroping helper does not need to be here anymore.

tbeswick96:
@PabstMirror just so I can understand better, when you say streaming is that to do with view distance? So the vars are there when in view but if you go far away and back the vars might not be there?

jonpas:
@tbeswick96 Correct.

PabstMirror:
http://killzonekid.com/arma-scripting-tutorials-variables-part-3/
>Houses for example like other map objects (and I mean map objects that have been added during map creation, not mission objects) are streamed. Think about it, the maps in ArmA are huge so even if you can load the whole map you will be using too much computer power. This is unnecessary, so map objects are streamed to you as you go. This means that if you attach a variable to an object on your map and try to make it global, on other computers the same object might not exist yet. And when the object appears later on other computers, it might not have variable you attached to it, so remember this.

tbeswick96:
Can just change the setVar to be on the nozzle instead if that is preferred?

PR #5152, opened 2017-05-11
1.70 FCS - Remove ACE_FCS from most vics

PabstMirror:
Right now 1.70 RC has the new FCS system
In some quick testing it's fairly close to ace's system.
New "Laze Target" button 
- just like ace you now have to hit the button to get a range in most things like the handheld RangeFinder
- Their moving target calculations seem to be using some magic based off of looking at the velocity of the cursorTarget instead of using real relative movement of the aim point like ace does, but it's not that bad imho.

There is a major conflict from fcs doing
```
discreteDistance[] = {};
discreteDistanceInitIndex = 0;
```
which seems to mess up their FCS and delete the projectile and cause some bizarre script errors in overpressure and winddeflection.
```
getPosASL _projectile=[2492.92,5692.51,-1.#IND], vectorDir _projectile=[-1.#IND,-1.#IND,-1.#IND]
```

This PR removes ace_fcs from all vanilla vics that have the new system.
The only 4 left that don't are the 2 AA canons and 2 attack choppers.

The other option is disabling vanilla FCS on vanilla weapons and keeping ace's fcs, but that cause more compatibility issues.

### DIFF
--- a/addons/fcs/CfgOptics.hpp
+++ b/addons/fcs/CfgOptics.hpp
@@ -11,110 +11,13 @@
         h = 0; \
     };
 
+class RscText;
 class RscControlsGroup;
 class RscMapControl;
 
 class RscInGameUI {
     class RscUnitInfo;
-    class RscUnitInfo_AH64D_gunner {
-        controls[] = {"CA_Distance","ACE_CA_Distance"};
-        MACRO_RANGEFINDER
-    };
-    class RscWeaponRangeFinder {
-        controls[] = {"CA_Distance","ACE_CA_Distance"};
-        MACRO_RANGEFINDER
-    };
 
-    class RscWeaponRangeFinderPAS13 {
-        MACRO_RANGEFINDER
-    };
-    class RscOptics_Rangefinder: RscUnitInfo {
-        class CA_IGUI_elements_group: RscControlsGroup {
-            class controls {
-                MACRO_RANGEFINDER
-            };
-        };
-    };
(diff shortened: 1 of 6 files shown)


commy2:
I'd say get rid of it. It served it's purpose as the superior system for years, but now it's time to make room for the new engine based solution.

jonpas:
Can we just add vanilla FCS to the AA cannons or does it not work for cases like that? We should for the attack helos though.

PabstMirror:
Attack choppers have the CCIP for the autocannon which doesn't currently work with ace_fcs.
The main problem is the drop is so great that the CCIP reticule goes off screen when at the highest zoom level.

I'm leaning towards removing ace_fcs from the 2 aa vics, as in vanilla they have the CCIP drop calc + a radar based lead-indicator. But that will mean breaking their airburst rounds feature.

bux:
Would it be possible to still use the "Tab" key but skip the FCS and apply the range only to the airburst feature?

PabstMirror:
Yesterday's dev changelog
>Tweaked: Gun zeroing now again works for gunships and AA platforms 

For those 2 we can wait and see what actually is in 1.70 final

tbeswick96:
> Tweaked: Gun zeroing now again works for gunships and AA platforms

This is in 1.70 if that changes anything

CCRcmcpe:
Well, vanilla FCS sometimes sucks. It doesn't take drag into consideration. To high drag projectiles vanilla FCS is so broken as they always land shorter at long distance.

PR #5154, opened 2017-05-11
MapGesutres - Fix dependency on maptools

### LINKED ISSUE #5153: Map gestures break without map tools loaded
**ACE3 Version:** `3.9.1` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- If you remove the ace_maptools pbo, ace_gesture markers don't hide on mouse key release and keep the last dragged position. On mission start they create at 0,0,0 and aren't hidden.

**Steps to reproduce:**
- Delete ace_maptools.pbo, click map in SP/Dedi mission.

**Placed Modules:**
- None

**Cause:**
- https://github.com/acemod/ACE3/blob/master/addons/map_gestures/functions/fnc_transmitterInit.sqf#L29

@jonpas pointed this out these vars will be null without maptools loaded

PabstMirror:
Close #5153


jonpas:
This might be confusing, why not use `["ace_maptools"] call EFUNC(common,isModLoaded)` instead?

PR #5157, opened 2017-05-13
SQF Lint Cleanup Pass

PabstMirror:
Did a quick pass with the SQF Lint tool. (https://github.com/LordGolias/sqf)

Most are just private cleanups which aren't a big deal, but it did find some actual errors.
- Fixed: getDoorTurrets (Not used, was broken, could drop?)
- Fixed: getMGRSdata (Was ignoring getMapData altitude)
- Fixed: throttledPublicVariable (Not used anywhere, should probably test if we use it in the future)
- Fixed: handleRespawn (Using bad var name, escort setVar would not be reset)
- Fixed dropNozzle
- magazineRepackProgress (magically not broken because progressbar uses same var name)
- readWeaponDataFromConfig (same thing, `_weapon` always existed from upper scope)

### DIFF
--- a/addons/advanced_throwing/XEH_postInit.sqf
+++ b/addons/advanced_throwing/XEH_postInit.sqf
@@ -115,7 +115,7 @@ addMissionEventHandler ["Draw3D", { // Blue is predicted before throw, red is re
         drawIcon3D ["\a3\ui_f\data\gui\cfg\hints\icon_text\group_1_ca.paa", [0,0,1,1], _newTrajAGL, 1, 1, 0, "", 2];
     } forEach GVAR(predictedPath);
     {
-        _newTrajAGL = _x;
+        private _newTrajAGL = _x;
         drawIcon3D ["\a3\ui_f\data\gui\cfg\hints\icon_text\group_1_ca.paa", [1,0,0,1], _newTrajAGL, 1, 1, 0, "", 2];
     } forEach GVAR(flightPath)
 }];
--- a/addons/advanced_throwing/XEH_postInit.sqf
+++ b/addons/advanced_throwing/XEH_postInit.sqf
@@ -115,8 +115,7 @@ addMissionEventHandler ["Draw3D", { // Blue is predicted before throw, red is re
         drawIcon3D ["\a3\ui_f\data\gui\cfg\hints\icon_text\group_1_ca.paa", [0,0,1,1], _newTrajAGL, 1, 1, 0, "", 2];
     } forEach GVAR(predictedPath);
     {
-        private _newTrajAGL = _x;
-        drawIcon3D ["\a3\ui_f\data\gui\cfg\hints\icon_text\group_1_ca.paa", [1,0,0,1], _newTrajAGL, 1, 1, 0, "", 2];
+        drawIcon3D ["\a3\ui_f\data\gui\cfg\hints\icon_text\group_1_ca.paa", [1,0,0,1], _x, 1, 1, 0, "", 2];
     } forEach GVAR(flightPath)
 }];
 #endif
--- a/tools/sqf_linter.py
+++ b/tools/sqf_linter.py
@@ -1,0 +1,63 @@
+﻿#!/usr/bin/env python3
+
+# Requires: https://github.com/LordGolias/sqf
+
+import fnmatch
+import os
+import sys
+import argparse
+from sqf.parser import parse
+import sqf.analyzer
+from sqf.exceptions import SQFParserError
+
+
+def analyze(filename, writer=sys.stdout):
+    with open(filename, 'r') as file:
+        code = file.read()
+        try:
+            result = parse(code)
+        except SQFParserError as e:
+            print("{}:".format(filename))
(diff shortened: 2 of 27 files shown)


addons/advanced_throwing/XEH_postInit.sqf
```diff
@@ -115,7 +115,7 @@ addMissionEventHandler ["Draw3D", { // Blue is predicted before throw, red is re
         drawIcon3D ["\a3\ui_f\data\gui\cfg\hints\icon_text\group_1_ca.paa", [0,0,1,1], _newTrajAGL, 1, 1, 0, "", 2];
     } forEach GVAR(predictedPath);
     {
-        _newTrajAGL = _x;
+        private _newTrajAGL = _x;
         drawIcon3D ["\a3\ui_f\data\gui\cfg\hints\icon_text\group_1_ca.paa", [1,0,0,1], _newTrajAGL, 1, 1, 0, "", 2];
```
CorruptedHeart: Could probably just use `_x` here without `_newTrajAGL`

jonpas:
Oh, commy also opened the PRs for getDoorTurret and publicVariable. :D

commy2:
missed this 🤦‍♀️

commy2:
This PR closes #5159 and #5160.

LordGolias:
Glad to know this was so helpful! Please share relevant issues to the SQF parser github please.

tools/sqf_linter.py
```diff
@@ -0,0 +1,63 @@
...
+    all_warnings = 0
+    all_errors = 0
+    
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-m','--module', help='only search specified module addon folder', required=False, default=".")
+    args = parser.parse_args()
+ 
+    for root, dirnames, filenames in os.walk('../addons' + '/' + args.module):
+      for filename in fnmatch.filter(filenames, '*.sqf'):
+        sqf_list.append(os.path.join(root, filename))
```
jonpas: 2 space instead of 4 space indentation.

tools/sqf_linter.py
```diff
@@ -0,0 +1,66 @@
...
+    args = parser.parse_args()
+ 
+    for root, dirnames, filenames in os.walk('../addons' + '/' + args.module):
+        for filename in fnmatch.filter(filenames, '*.sqf'):
+            sqf_list.append(os.path.join(root, filename))
+        
+    for filename in sqf_list:
+        warnings, errors = analyze(filename)
+        all_warnings = all_warnings + warnings
+        all_errors = all_errors + errors
```
SilentSpike: `+=`

tools/sqf_linter.py
```diff
@@ -0,0 +1,66 @@
...
+            print("{}:".format(filename))
+            writer.write('    [%d,%d]:%s\n' % (e.position[0], e.position[1] - 1, e.message))
+            return 0, -1
+
+        exceptions = sqf.analyzer.analyze(result).exceptions
+        if (exceptions): 
+            print("{}:".format(filename))
+            for e in exceptions:
+                if (e.message.startswith("error")):
+                    errors = errors + 1
```
SilentSpike: `+=`

tools/sqf_linter.py
```diff
@@ -0,0 +1,66 @@
...
+                else:
+                    warnings = warnings + 1
```
SilentSpike: `+=`

tools/sqf_linter.py
```diff
@@ -0,0 +1,66 @@
...
+    parser.add_argument('-m','--module', help='only search specified module addon folder', required=False, default=".")
+    args = parser.parse_args()
+ 
+    for root, dirnames, filenames in os.walk('../addons' + '/' + args.module):
+        for filename in fnmatch.filter(filenames, '*.sqf'):
+            sqf_list.append(os.path.join(root, filename))
+        
+    for filename in sqf_list:
+        warnings, errors = analyze(filename)
+        all_warnings = all_warnings + warnings
```
SilentSpike: `+=`

tools/sqf_linter.py
```diff
@@ -0,0 +1,66 @@
...
+    warnings = 0
+    errors = 0
+    with open(filename, 'r') as file:
+        code = file.read()
+        try:
+            result = parse(code)
+        except SQFParserError as e:
+            print("{}:".format(filename))
+            writer.write('    [%d,%d]:%s\n' % (e.position[0], e.position[1] - 1, e.message))
+            return 0, -1
```
SilentSpike: What's the reason this returns a `-1` for number of errors? I feel like it should be 0, but perhaps I'm missing something

jonpas:
SQF Lint now has directory analyze option https://github.com/LordGolias/sqf/blob/92130853eadeec524ca641b2c4bde34f5fc18f52/sqflint.py with help of @PabstMirror 's script. It doesn't count the amount of warnings and errors as of now though.

tools/sqf_linter.py
PabstMirror: -1 used to mean error, but I changed it, thanks for catching

PR #5162, opened 2017-05-13
fix script error in ACE_common_fnc_throttledPublicVariable, fix #5160

### LINKED ISSUE #5160: Wrong syntax in `params` of `fnc_throttledPublicVariable.sqf`
Hi guys and thank you for this great mod.

In `ACE3-3.9.1`, `common/functions/fnc_throttledPublicVariable.sqf`, line 30, we read

```
_x params [_unit, _varName];
```

I believe it should read

```
_x params ["_unit", "_varName"];
```

since the signature of the right hand side of `params` is `[STRING,STRING,...]`.

This was found while I was testing the [SQF parser](https://github.com/LordGolias/sqf) in the ACE base code.

commy2:
- title
- fix #5160


jonpas:
Close, as already included in #5157?

PR #5163, opened 2017-05-13
1.70 CMs - Remove countermeasure mode switching

jonpas:
- Remove our version of countermeasure mode switching (burst - single) as vanilla will have it in 1.70.


PabstMirror:
Works great, old code pretty much matches what they added in 1.70

PR #5165, opened 2017-05-15
Realistic names for 1.70 pylon magazines

PabstMirror:
3den pylon interface uses CfgMagazines name


PabstMirror:
Realistic weapons in general could use a lookover

- We rename the "macer" to the maverick, jets expansion adds a "macer II"
- Some Renamed weapons are too long for some HUDs (e.g. "AIM-132 ASRAAM")
- Seems weird putting "9K121 Vikh" on a blufor vehicles compared to vanilla "Scalpel"
- Both vanilla "DAR" and "Shrieker" are now called Hydra

PR #5167, opened 2017-05-15
Disable aircraft flightmodel changes

### LINKED ISSUE #4817: The plane A-164 Wipeout / A-10D Thunderbolt is very hard to land
**Arma 3 Version:** `1.66` stable
**CBA Version:** `3.1.2.161105` stable
**ACE3 Version:** `3.8.3.14` stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
The A-164 Wipeout / A-10D is very hard to land. Autoland hardly manages to land or fails with catastrophic failure.
As it looks like the file https://github.com/acemod/ACE3/blob/master/addons/aircraft/flightmodel_thunderbolt.hpp is the culprit configuring some parameters. In Vanilla the Wipeout can be landed very easily with keyboard+mouse as well as throttle+stick, e.g. slow down to 180 km/h, glide off, touch down, no problems. With @cba_a3+@ace one needs to stay above around 194 km/h at least and be very gentle. With throttle+stick (Saitex X45) this is almost impossible for me to manage as the error margin is very small.

I propose to adapt the settings so that one has a better chance to land without hurting the pilot or damaging the airframe.

Maybe the problem was introduced because the parameters come from ACE2/AGM and are not really applicable to the current flight model?

**Steps to reproduce:**
- Try to land A-164 Wipeout and compare to vanilla

- Regardless of player mode, reproducible in single player editor

**Placed Modules:**
- None

### LINKED ISSUE #5104: plane AI lands in front of the runway end crashes
**Arma 3 Version:** `i dont know wat the version was but tha date  24/04/2017` 
**CBA Version:** `i dont know wat the version was but tha date  24/04/2017` 
**ACE3 Version:** `i dont know wat the version was but tha date  24/04/2017` 
**mods used:**
- CBA_A3
- ace
```

**Description:**
-it does not give an error. the AI it lands infront of the runway end i dont kow why. it looks like it goes to slow to land end then it stalls. i tested it with only CBA if the problem was in thare but it was a smooth landing 

**Steps to reproduce:**
-maybe increase the speed of the AI wile landing 

-wile an AI tries to land a plane on a runway 

**Placed Modules:**
- none only a plane with a land waypoint end 1 unit to look if he lands right (he does not)

**RPT log file:**
-none

PabstMirror:
Fix #5104
Fix #4817

These changes are ancient and unreliable.
Using autopilot landing a A-164 you will crash short of the runway and die.


commy2:
Should delete the files.

PR #5168, opened 2017-05-17
Use correct TransportX type for ace items

PabstMirror:
Fix #5068

As far as I can tell arma can handle the wrong entries.
This does seem to put he mine detectors at the top of the "misc items" box, but I noticed no other changes.

PR #5170, opened 2017-05-17
Common func to run config tests

PabstMirror:
Adds a simple common function to run tests.
Mostly just config checkers right now (e.g. test_vehicleInventory.sqf for #5168)

- Makes them easier to call.
- Gets them out of `#ifdef DEBUG_MODE_FULL` so they can be used without file patching
- Very basic test pass / fail reporting

Edit: Some actual code changes in this:
- `latitude` loaded from map configs is now reversed (seems to be an arma convention, matches in-game world map display)
This only effects MGRS and advBallistics (ab uses sin/cos, so it will have zero effect)
- Added some map configs (1.72 Malden)
- Removed `Map Grid Warning` warnings (maps configs will probably never be changed)
Warnings now part of the test system


Glowbal:
I don't like the `dev/` sub directory for tests. I propose to rename it to `test/`.

PabstMirror:
I was thinking it could hold things like debuging functions like 
https://github.com/acemod/ACE3/blob/medical-rewrite/addons/medical/functions/fnc_dev_watchMedicalStats.sqf

but we could keep those in `\functions` and just put test stuff in `\test`

PabstMirror:
I guess these cross the line between debuging and actual unit testing.

I write a lot of debuging and testing code like this, and like to keep it in the repo, but don't want it cluttering up the main code.

So where should these go, and what do we want them named as?

Cuel:
Will tests be included in release pbos?

PabstMirror:
Yes, for example the rearm/reammo tests will be useful for testing compatibility with other mods.

Those tests weren't usable because they are inside of a `#ifdef DEBUG_FULL` block, which isn't accessible on a release build of ace

PabstMirror:
So what do we want to do with these?
- `\dev`
- `\test`
- keep in `\functions`

addons/common/functions/fnc_runTests.sqf
```diff
@@ -0,0 +1,48 @@
+/*
+ * Author: PabstMirror
+ * Run test functions
```
jonpas: Missing final dot.

addons/common/functions/fnc_runTests.sqf
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Arguments:
+ * 0: Specifc test to run (default is to run all) <STRING><OPTIONAL>
```
jonpas: `Specific` and `<STRING> (default: #all)`

addons/common/functions/fnc_runTests.sqf
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call ace_common_fnc_runTests;
```
jonpas: I think we decided to do those without `;`.

Issue #4265, opened 2016-08-18
Screeching seagulls give away the presence of spectators when using ACE_Spectator


brndd:
**Arma 3 Version:** `1.62.137494` (stable)
**CBA Version:** `3.0.0.160713` (stable)
**ACE3 Version:** `3.6.2.0` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
As per the default Arma3 spectator functionality, spectators appear as screeching seagulls visible to players. This is quite immersion breaking, especially when using a respawn template which causes players to enter spectator mode when they die; one can immediately tell if a nearby player has died dead from the screeching of the ghost seagull released from their corpse. This means that there's no point in eg. checking the body of the player to see if they're still alive and reviveable with ACE medical, because if they turned into a seagull they're dead for good.

**Steps to reproduce:**
1) Create a multiplayer mission with 2 or more slots using the following respawn settings in description.ext:

```
respawn = "GROUP";
respawnOnStart = 0;
respawnTemplates[] = {"GROUP", "ace_spectator"};
respawndelay = 3;
```

2) Enter the mission with 2 or more players.

3) Kill one of the players.

4) Upon death, the player will spawn as a seagull, visible and audible to the living players.

**Where did the issue occur?**
- Multiplayer, both dedicated and P2P hosted

**Placed Modules:**
- None

**RPT log file:**
- Sorry I don't have one, nor do I feel it's relevant to the issue at hand.

SilentSpike:
Good report, I actually thought we had code in already to deal with seagulls but I guess I must have revised that or something.

brndd:
Sweet. In the mean time putting this in onPlayerRespawn.sqf for the mission seems to work for migrating the seagulls away:

```{
	if (_x isKindOf "seagull") then
	{
		_x setPos [0, 0, 15000];
	};
} forEach nearestObjects [player, [], 250];```

commy2:
@brndd remove `"GROUP"` from `respawnTemplates[]`

SilentSpike:
The group template should do nothing really, it's an outdated leftover from A2 as far as I can tell.

Confirming that seagulls are still a thing - unfortunately it's kind of a pain to develop with respawn types `group` and `side` because without more than one player the mission instantly ends when you die (regardless of the `"Mission fail when everyone is dead"` - aka `EndMission` - template being active).

brndd:
[...]

PR #5171, opened 2017-05-17
Spectator overhaul

SilentSpike:
**Objective:**
Overhaul the spectator component to match the UX of BI's End Game Spectator, while retaining any additional functionality of the current implementation of ace spectator.

**Todo:**
- [x] New UI
- [x] Free camera
- [x] Follow camera
- [x] First person camera
- [x] Unit focus system:
  - [x] Can focus on units in free camera
  - [x] Can click on units in 3D world
  - [x] Can click on units in 2D map
  - [x] Can click on units in UI list
  - [x] Focused unit information widget
    - [x] Toggle
    - [x] ~~Add support for ACEX kill tracker feed~~ (Scrapped for now)
- [x] UI toggle
- [x] Camera mode buttons
  - [x] Update the buttons properly when camera mode switches
  - [x] Replace with more generalised button dock for potential future camera modes?
- [x] Map functionality
  - [x] Unit drawing and highlighting
  - [x] Camera icon
  - [x] Camera teleport on click
- [x] Help UI element
  - [x] Toggle
  - [x] Update entries dynamically
- [x] Locations functionality
  - [x] List tab switching
  - [x] Populate list with locations
  - [x] Functions to add/remove locations
  - [x] Location drawing
- [x] Camera light toggle
- [x] Adjustable view distance
- [x] Unit/Group visualisation
  - [x] Toggle
  - [x] Draw loop
  - [x] Support for revive icons (BIS and ACE)
- [x] Projectile visualisation
  - [x] Toggle
  - [x] Draw loop
  - [x] Add fired EHs
- [x] Usable virtual spectator unit
  - [x] Add the actual unit (used in respawn template to replace seagulls (#4265))
  - [x] Make it work when set to playable (remember to set scope to 2 when done)
- [x] ~~Setting to optionally replace BI's spectator with ACE spectator~~
  - [x] Not feasible. Just look into respawn 1,4 & 5 compatibility.
- [x] Everything marked with `TODO` in the code
- [x] Probably a bunch more stuff

**Retained:**
- [x] Vision mode system
  - [x] Fine control over available vision modes
  - [x] Can add extra [vision modes](https://community.bistudio.com/wiki/setCamUseTI)
- [x] Camera mode system
  - [x] Fine control over available camera modes
  - [x] Possibility to add extra camera modes in future
- [x] Non-forced UI capability
- [x] Unit whitelisting/blacklisting
- [x] Control over spectatable sides
- [x] Compatibility with BI `counter` respawn template (use [UI event](https://github.com/acemod/ACE3/blob/f7d9a165f6adef07a4ecfd7be7931a5d5c1750a6/addons/markers/CfgEventHandlers.hpp#L20-L24))
 - [x] Free camera heading publicly adjustable
 - [x] Free camera position publicly adjustable

**Removed:**
- [x] Compass UI element
- [x] Some camera attributes
  - [x] Free camera zoom functionality
    - [x] ~~Replace with `+/-` incremental zoom keys~~
    - Engine limitation prevents this. Need to use `camcurator` for engine driven controls, which doesn't support `camSetFOV`.
  - [x] Camera speed functionality
    - [x] Replace with `2x, 4x & 8x` speed hotkeys
  - [x] Free camera tilt publicly adjustable
  - [x] Add obsolete warnings to public function
- [x] Interrupt system?
- [x] "Side filter" setting (same results can be achieved by manipulation of spectatable sides)
- [x] "Unit filter" setting (replace with "spectate AI" setting, aim is to simplify setup process)
- [x] Obsolete stringtable entries

**Add:**
- [ ] Extra optional camera modes?
  - [ ] Bullet cam
  - [ ] Top down
- [x] ~~Extra respawn templates?~~
  - [x] ~~Number of lives then permadeath?~~
- [ ] Compatibility with other respawn templates?

**Things to Test:**
- [x] JIP into a virtual spectator unit initialises correctly
- [x] Location removal updates the list correctly
- [x] Unit whitelisting/blacklisting works correctly
- [x] Addition/removal of vision modes still works as expected
- [x] Addition/removal of camera modes still works as expected
- [x] Adjusting camera attributes all work as expected

**Special thanks:**
Significant amount of code contributed by @AACO who had already performed a lot of optimisation and conversion work for https://github.com/BourbonWarfare/POTATO. Added to function headers where appropriate and will be added to authors too.


SilentSpike:
So thankfully the hard part of this where I had to mash two code bases together and break arma a bunch of times is over (which is why a load of stuff is done in that one "overhaul spectator" commit).

Progress from here on should be relatively smooth (though 1.7 arma has started to semi-regularly crash my display driver) as I can actually test changes as I make them, but everything's looking good so far :+1:

addons/spectator/functions/fnc_cam_draw3D.sqf
```diff
@@ -0,0 +1,193 @@
...
+            // Draw icon
+            drawIcon3D _icon;
+
+            nil // Speed loop
+        } count GVAR(thingsToDraw);
+        END_COUNTER(drawTags);
+    };
+
+    // Draw projectiles and grenades paths
+    if (GVAR(drawProjectiles)) then {
```
bux: For drawing projectiles there's now:
https://github.com/CBATeam/CBA_A3/blob/master/addons/diagnostic/fnc_addUnitTrackProjectiles.sqf

SilentSpike: Problem with this is that it's generalised and doesn't meet the needs of this specific use case (optimised for mass application, drawing when near camera only, easily toggled, etc.). However, I might be able to borrow some of the drawing code if applicable. :+1:

bux: Makes sense to me.

Though, you might still [add the unit](https://github.com/CBATeam/CBA_A3/blob/master/addons/diagnostic/fnc_addUnitTrackProjectiles.sqf) when in proximity and [remove it](https://github.com/CBATeam/CBA_A3/blob/master/addons/diagnostic/fnc_removeUnitTrackProjectiles.sqf) when too far from it.
And for performance you could [reduce the amount of lines that are drawn at the same time](https://github.com/CBATeam/CBA_A3/blob/master/addons/diagnostic/XEH_preInit.sqf#L15)

SilentSpike: Hmm, I'll try and experiment with this when I'm working on the projectile drawing

addons/spectator/functions/fnc_cam_draw3D.sqf
```diff
@@ -0,0 +1,193 @@
...
+                ["_projectile", objNull, [objNull]],
+                ["_segments", [], [[]]]
+            ];
+
+            if !(isNull _projectile) then {
+                private _newestIndex = _segments pushBack [
+                    getPosVisual _projectile,
+                    (vectorMagnitude velocity _projectile) call {
+                        if (_this < 250) exitWith { [0,0,1,1] };
+                        if (_this < 250) exitWith { [0,1,0,1] };
```
AACO: oops, these should probably not both be 250

SilentSpike: Good catch, should be `500` :+1: I'm just about to look into finishing off the projectile drawing now

nomisum:
Removal of cam zoom why? One of the features I really love/use and miss in other spec cams. Makes for great screenshots. 

Somehow related: both BIS/ACE spec lack smooth (inertia) cam movement options for a nice spec video :)

SilentSpike:
Mostly because it adds an extra layer of complexity that most users don't use - plus you can always just move the camera closer :stuck_out_tongue:

nomisum:
intertia toggle would be greatly appreciated!

i beg your pardon, but the outcome of zooming and moving closer is way different.

something like this is not possible with moving closer: http://imgur.com/a/f8buo (you wont see any details in the background at all). so combining foreground and distant background into one composition is my goal. even during mission i often find myself making screenshots only while "zoomed in" via cam focus because the output is nicer.

SilentSpike:
Apologies, I didn't intend to imply that they were the same thing, just that for the majority of users the desired outcome is to get a closer view rather than a zoomed in one.

I could quite easily add incremental zoom capability on the numpad `+` and `-` buttons now that I think about it. Would that meet your needs?

nomisum:
I was about to write I prefer smooth zoom, but a quick test with splendid cam showed no real benefits (its quite sluggish/imprecise). Would you have to limit the steps?

Btw thanks for working on the speccam, we use it quite extensively.

SilentSpike:
Not sure what you mean by limit the steps? The increments would have to be set, but could be pretty fine and then you can just hold the button to zoom as needed

nomisum:
you could have had the idea of limiting the amount of zoom steps, e.g. 10 steps from 1x - 10x. an unnecessary limitation imo, glad you are ruling this out.

nomisum:
if I could throw in even more suggestions (you got me started, lol):

for creating beautiful sequences when spectating a slower camera movement could come quite handy, so maybe you could extend the movement speed multipliers (2x, 4x & 8x) in the other direction as well (1/2, 1/4 and 1/8). (you dont need to preset those keys as target group might be limited)

SilentSpike:
I'd say those are almost certainly unnecessary as the default camera movement is pretty slow to begin with (to the point I've been considering upping it in the back of my mind - but that would make x8 super fast)

jonpas:
The idea is that you can move really close to someone without jumping from one side to the other I assume.

SilentSpike:
I tried adding those zoom buttons to the + and - keys btw, but because the camera control is now engine driven it seems to override those keys and they're used for a weird diagonal movement or something.

Any good alternative suggestions for keys to use? Page up and down are already used for view distance, I suppose I could make them accept a shift modifier for zoom?

SilentSpike:
So after doing some testing it actually seems that for some reason the `camSetFOV` and `camCommand "inertia on"` commands are having no effect...

nomisum:
"inertia on" doesnt work with `camcurator`, if that is what you are now using. drove me mad a few times already. the same goes for `camSetFOV` probably, needs type "camera".

and yes, initial cam movement is slow already but some things just need a tiny, subtle movement :D

SilentSpike:
Hmm, in that case I'm not sure if I can add those features since I think `camcurator` might be needed for the controls to work. I can do some tests using a regular `camera`, but have paused work on this for now since arma is giving me terrible crashes that sometimes make me have to hard reset my computer.

SilentSpike:
Another option would be to somehow define our own camera that supports the controls and the commands. However, idk if those are in the config somewhere or what

SilentSpike:
Yeah so it definitely has to be a `CamCurator` otherwise the controls are totally different and it doesn't support the rest of the `camCommand` arguments

Drofseh:
RE Camera control.

I don't know if this is something you've considered, but I used to use the CSSA3 Spectator script, and one of the things I and a few other people in the unit really liked about it was that you could use the mouse to move the free camera around.

Holding LMB and RMB at the same time would allow exceptionally fluid continuous X/Y camera panning based on the speed and direction of mouse movement until you let go of the buttons or moved the mouse in a different.

nomisum:
CSSA3 had been used by us as well but suffered some bugs iirc. It uses type "camera" though and moving was great.

Locking camera view vector to mouse movement [in free cam] would be great indeed. (pressing RMB currently doesnt result in 1 to 1 movement it seems, plus its tiring over time).

SilentSpike:
I would love to incorporate these requests, but the problem is that in order to achieve such results the reliable engine solution provided for camera movement (`camCommand "manual on"`) would have to be replaced by PFH overhead which is what was being used previously. This greatly increases the complexity of the code, adds extra overhead and also lends itself to causing unforeseen issues.

If I can find workarounds then I'll definitely do what I can, but my priority right now is to make the overhaul feature complete and ready to merge into an upcoming release build.

SilentSpike:
So this is basically done. There are extra goodies on the wishlist, but the system itself is feature complete and just requires a bunch of testing to find any unexpected behaviour or non-obvious bugs.

I have some things on my to-test list which I am aware might be problem areas, but if people want to build this branch and mess around with it let me know if you find anything :+1:

addons/spectator/functions/fnc_compat_spectatorBI.sqf
```diff
@@ -22,6 +22,10 @@ if !(QUOTE(ADDON) in _templates) exitWith {};
 // Start our spectator
 [true] call FUNC(setSpectator);
 
+// Delete the seagull that spawns (not actually the player, a CfgNonAIVehicles object)
+// This could delete seagulls created by a wildlife module (TODO: Find a more accurate method of getting the exact seagull)
```
AACO: If the respawn type is bird, the seagull should be passed in as a parameter.

addons/spectator/XEH_postInit.sqf
```diff
@@ -12,3 +12,15 @@ if (isServer) then {
 };
 
 [QGVAR(stageSpectator), FUNC(stageSpectator)] call CBA_fnc_addEventHandler;
+
+// Initalise virtual spectators
+[QGVAR(virtual),"initpost",{
+    if !(GVAR(isSet)) then {
+        // Delay until local player is fully initalized
+        [
+            { !isNil { player } && { !isNull player } },
```
AACO: Should this be ACE_PLAYER?

SilentSpike: No it should always be `player`. Good question though, I'll add a comment that says this explicitly

addons/spectator/functions/fnc_compat_spectatorBI.sqf
SilentSpike: I'll do some more testing related to this. if so then I could have the template take care of the seagull while this display XEH handles replacing BI's spectator

SilentSpike: Testing shows that the template is only passed the seagull with respawn type `1`, so for types `4` and `5` I still have to do a workaround. It's an improvement though.

SilentSpike:
Yay, first review :smile:

Two of these are already on my todo list:
- Proper Zeus compatibility
- When clicking on same tab list it gets funky

Some sound like engine weirdness:
- Help scrollbar
- Scroll speed not influenced by modifier keys
- Player unit moving while in first person view (I can maybe workaround this one if I selectively block the key input or something)

Nametags overlapping:
- Can investigate, but it will add overhead to sort the units by distance. The way it's done here is the same as BI's system.

Slow default camera movement:
- Yeah I actually thought this myself, I left the speed values all the same as BI's, but they can be easily adjusted. Good suggestion.

All of the list expanding stuff:
- Noted, but low priority as changing the behaviour is a **lot** of work for little gain.

Can't focus in camera mode other than freelook:
- You should be able to, it was working fine when I tested it :thinking:

Projectile lines:
- There is no timer on these, they simply draw until the projectile becomes null (same system used in BI's spectator). With an adjustment to the code I can make them last 1 frame longer though, hopefully that is enough..

Should be noted that the focus system follows the same behaviour as BI's spectator and also the help window is currently using all of their localisations.

addons/spectator/functions/fnc_compat_zeus.sqf
```diff
@@ -12,8 +12,8 @@
 params ["_display"];
 
 _display displayAddEventHandler ["Unload",{
-    // Only re-open if still a spectator
-    if (GVAR(isSet)) then {
+    // Only re-open if still a spectator (and not remote-controlling)
+    if (GVAR(isSet) && {isNull (GETMVAR(bis_fnc_moduleRemoteControl_unit,objNull))}) then {
```
jonpas: Maybe check `player != ace_player`? That would catch other ways similar to remote control that we bypass with `ace_player`.

SilentSpike: This is zeus specific though as it occurs when the zeus display closes

addons/spectator/functions/fnc_ui_updateHelp.sqf
```diff
@@ -36,22 +37,40 @@ if (_cameraMode == MODE_FREE) then {
...
+    format ["[%1]", [DIK_F1] call CBA_fnc_localizeKey],
+    localize "STR_A3_Spectator_Helper_Controls"
+];
 
 if (_cameraMode == MODE_FREE) then {
-    _controls pushback ["[LSHIFT]", localize "STR_A3_Spectator_Helper_Shift"];
-    _controls pushback ["[LALT]", localize "STR_A3_Spectator_Helper_Alt"];
-    _controls pushback ["[LALT + LSHIFT]", localize "STR_A3_Spectator_Helper_ShiftAlt"];
+    _controls pushBack ["[LSHIFT]", localize "STR_A3_Spectator_Helper_Shift"];
+    _controls pushBack ["[LSHIFT]", localize "STR_A3_Spectator_Helper_Alt"];
```
SilentSpike: Whoops, this is a mistake. Problem is if I use the CBA_fnc_localizeKey on these keys the result is too long to fit within the bounds of the control (i.e. the 3rd one becomes "Left Alt + Left Shift"). Idk if these DIKs ever represent different buttons on any international keyboard though so could be a non-issue.

SilentSpike:
I consider this PR no longer WIP and ready for testing. That's not to say it's finalised, but that all of the functionality required for an initial release is complete.

However, feedback is needed on the UX (things like camera speed or intuitiveness of how the focusing system works) and whether there are any scenarios that things behave unexpectedly.

SilentSpike:
Oh man, some of these bugs are so strange :laughing:

Some things (like the unit list tooltips) are just brought over directly from BI's system. I'm unsure about changing those as I quite like to keep the UX the same (I think in the tooltip case the idea is to be able to see really long player names).

SilentSpike:
Great feedback though, my responses:

- Camera speed, yeah shift is engine driven and alt is not. I liked the idea of slow mode being a toggle, but also found that it made things awkwardly inconsistent. Will make it have to be held again
- The help labels, I have no idea what's going on there. Should be a fun one to debug as they work fine in my testing :laughing:
-  Order of the list was implicitly changed because I made it dynamic so it can support additional sides. However I think I can fix this by hard-coding the 4 main sides.
- The gap in the UI is weird. The UI again is ported 1:1 from BI's system. I'll look into the configs as it's not there for me.
- The character *should* be filtered yeah. Weird again as I've never experienced that in testing :laughing: 
- Tooltip - see above comment.
- Will centre :+1: These are the little intuitive UX feedback things I like to hear about
- Sorting by rank, I think should be doable. Not sure what kind of overhead it will add, but will give it a go.

bux:
As confirmed on Slack, I am using the current release version of CBA v3.3.1

SilentSpike:
Yeah, just to clarify to anyone wishing to test this branch: You should be running the development version of CBA (at least for the keybind's to display correctly in the UI, that might be the only thing that needs it).

SilentSpike:
So the sorting by rank thing might prove too complicated. Basically, the problem with altering the behaviour of the list is related to how it is populated. Once an entry is in the list, it remains untouched until it is time to remove it (if ever). This is vital to maintain correct list behaviour (retaining current selection, expansion states, etc.).

However, I can make it so that when groups are first added to the list their units are sorted by rank. It just won't accommodate for addition of more units or future changes in rank. The caveat with this, is that by doing so the order in which Left and Right switch through the unit list will jump around instead of being logical as it currently is.

SilentSpike:
Rank might actually be a good thing to display in the tooltip, instead of the group ID which is already visible

SilentSpike:
I'm not a huge fan of the group textures used by BI's dynamic groups stuff, so implemented this common function to get an appropriate group icon. Currently only used by the list, but would like to also try and use it in icon rendering too (however because the default group icon used by spectator is it's own special .paa with a different size it's going to be a pain to adjust the code to position and scale these icons).

Someone with a valid squad.xml logo should also test that the code can use it without requiring file patching to be enabled. Steps to test:
- Run game without file patching
- Launch a test mission where you are a group leader
- Add yourself to the spectator whitelist (`[[player]] call ace_spectator_fnc_updateUnits`)
- Start spectator while alive (`[true] call ace_spectator_fnc_setSpectator`)
- Check whether there are any script errors and whether your squad.xml logo is used in the unit list for your group

AACO:
+1 on getting rid of the random icons

One small issue I've seen with Squad.XML logos is (especially with the BI Units setup) is the squad's paa never downloads locally, so trying to display it shows scripting errors, so I strongly suggest a test with using a units based logo.

jonpas:
Squad's `.paa` always downloads locally, and that's what is used in Zeus as well.

AACO:
Not in all cases @jonpas I've seen more than a few issues on our sessions with players using the units page spitting RPT errors due to missing textures when they've had unit page setups.

jonpas:
Ah I know what you mean, that's usually caused by an old upload of the units logo back when A3 Units was first released.

SilentSpike:
I was under the impression that zeus handled the squad.xml stuff engine side rather than using the script commands. I might just forgo the inclusion of those if it makes life simpler

AACO:
@jonpas yeah, the ones I've seen issue with are pretty old setups.

@SilentSpike it is handled engine side for Zeus, but it's basically looking for the texture specified by units/XML but the file either fails to download or doesn't download properly and will spam the RPT logs with missing texture errors. I'm not sure how the scripting way would handle this, it might just throw a single click through popup for the spectator, but I'm not sure if there's a reliable way to reproduce it.

SilentSpike:
I'll probably pull out the squad.xml support, it has potential for errors with malformed squad.xml data and also makes the code a bit awkward to adjust so that they aren't coloured unintentionally.

I think for group drawing I'm going to have to just include some custom .paa files, otherwise changing the code to find the right scaling and positioning is gonna be a pain in the ass.

AACO:
Yeah that's what I ended up doing silent, if you look at how the default icon is setup it's a super easy conversion. It's basically just putting the icon on a large transparent sheet. I have a set of icons you can use, but a few of them are custom to our framework setup, so you probably wouldn't want to use them.

SilentSpike:
I'm just gonna set up a batch conversion process using the default ones as I think there are some extra categories I'm using :+1:

SilentSpike:
Would have got those changes pushed a while ago if arma wasn't crashing my display drivers again. I don't want to waste time restarting my PC anymore so they're totally untested and my work on this PR is halted until I can be bothered putting up with the issues again

addons/spectator/functions/fnc_getGroupIcon.sqf
```diff
@@ -4,56 +4,51 @@
...
  * Arguments:
  * 0: Group to get the icon of <GROUP>
- * 1: Allow leader's squad.xml logo <BOOL> (Default: false)
+ * 1: Return icons for draw3D use <BOOL> (Default: false)
  *
  * Return Value:
  * Icon of group <STRING>
  *
  * Examples:
  * [group player] call ace_common_fnc_getGroupIcon;
```
AACO: update header for move

addons/spectator/functions/fnc_ui.sqf
```diff
@@ -34,18 +34,19 @@ while {dialog} do {
...
 if (_init) then {
     // UI visibility tracking
-    GVAR(uiVisible) = true;
-    GVAR(uiHelpVisible) = true;
-    GVAR(uiMapVisible) = true;
-    GVAR(uiWidgetVisible) = true;
+    GVAR(uiVisible)         = true;
+    GVAR(uiHelpVisible)     = true;
+    GVAR(uiMapVisible)      = true;
+    GVAR(uiWidgetVisible)   = true;
```
Phyma: Looks pretty but I don't like it. Should be only one space before = .

AACO:
Just curious, do you have any calculations on the time saved with the squared distances?

SilentSpike:
> Just curious, do you have any calculations on the time saved with the squared distances?

Nope not at all, but because these functions are running frequently I like to optimise them as much as humanly possible. Quick comparison in the debug console shows no appreciable difference in speed.

nomisum:
One thing came to my mind:

Would it be possible to provide a script hook for additional camera modes? 

So I could add my own custom cinematic camera with an add-function via script? Thus I dont need to intercept/override keys and it would be much cleaner than hacking something on top of ace. Should hide spectator UI but let the unit stay as an ace spectator regardless. Does that make sense?

SilentSpike:
*I think* that's possible, I've set up the camera button dock to be easily extensible and on the back end the camera modes available are just integers in an array so that's also easy to extend.

The difficulty I foresee is that new camera modes need to integrate cleanly into the whole system (various aspects of the UI/UX are influenced by the camera mode and changes in camera mode) and it could be challenging to generalise some of the functionality enough to support arbitrary addition of new camera modes (I could be totally wrong, it might already largely take care of itself).

It's certainly not something within the scope of this PR and personally I probably wouldn't spent time developing it, but anyone can PR that kind of framework if they want.

nomisum:
i see, will take a look after your PR is finished 👍

PR #5175, opened 2017-05-18
Fix Zeus teleport players module

SilentSpike:
- Fix zeus teleport module on non-local units (i.e all other players)
- Remove BI error message when using zeus UI modules that have no function

### DIFF
--- a/addons/zeus/functions/fnc_moduleTeleportPlayers.sqf
+++ b/addons/zeus/functions/fnc_moduleTeleportPlayers.sqf
@@ -37,7 +37,7 @@ if (_group) then {
     // Must be ran where local
     private _attached = attachedTo _logic;
     if (isNull _attached) then {
-        [QGVAR(moveToRespawnPosition), [_x, _logic], _x] call CBA_fnc_targetEvent;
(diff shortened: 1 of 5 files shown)


addons/zeus/functions/fnc_moduleTeleportPlayers.sqf
```diff
@@ -34,10 +34,10 @@ if (_group) then {
 {
     moveOut _x;
 
-    // Must be ran where local
     private _attached = attachedTo _logic;
     if (isNull _attached) then {
-        [QGVAR(moveToRespawnPosition), [_x, getPos _logic], _x] call CBA_fnc_targetEvent;
+        // Function takes position AGL and must be ran where local
+        [QGVAR(moveToRespawnPosition), [_x, _logic modelToWorld [0,0,0]], _x] call CBA_fnc_targetEvent;
```
Neviothr: Spaces after `,` in `modelToWorld`? `[0,0,0]` -> `[0, 0, 0]`.

SilentSpike: I'm not sure if we have a set standard for that, but personally I only use spaces where it helps with readability. In this case I'd argue that adding spaces actually has the opposite effect.

PR #5177, opened 2017-05-18
Just warn for old compat pbos

PabstMirror:
Having old compat pbos is less critical
Depending on how people update, they may get `@ace` way before `@ace_compat_X` and would be unable to play at all.

### DIFF
--- a/addons/common/functions/fnc_checkFiles.sqf
+++ b/addons/common/functions/fnc_checkFiles.sqf
@@ -41,18 +41,28 @@ private _addons = "true" configClasses (configFile >> "CfgPatches");//
 _addons = _addons apply {toLower configName _x};//
 _addons = _addons select {_x find "ace_" == 0};
 
+private _oldCompats = [];
 {
     if (getText (configFile >> "CfgPatches" >> _x >> "versionStr") != _version) then {
         private _errorMsg = format ["File %1.pbo is outdated.", _x];
 
         ERROR(_errorMsg);
 
-        if (hasInterface) then {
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_checkFiles.sqf
```diff
@@ -41,18 +41,28 @@ private _addons = "true" configClasses (configFile >> "CfgPatches");//
...
+            };
+        } else {
+            _oldCompats pushBack _x;  // Don't block game if it's just an old compat pbo
         };
     };
     false
 } count _addons;
+if (!(_oldCompats isEqualTo [])) then {
+    [{
+        systemChat format ["Warning: The Following ACE Compatiblity PBOs are Outdated %1", _this];
```
bux: Could we use one of the big red error boxes from CBA and have them fade out / hide after e.g. 10 seconds?

addons/common/functions/fnc_checkFiles.sqf
```diff
@@ -41,18 +41,29 @@ private _addons = "true" configClasses (configFile >> "CfgPatches");//
...
+        } else {
+            _oldCompats pushBack _x;  // Don't block game if it's just an old compat pbo
         };
     };
     false
 } count _addons;
+if (!(_oldCompats isEqualTo [])) then {
+    [{
+        // Lasts for ~10 seconds
+        ERROR_WITH_TITLE_1("The Following ACE Compatiblity PBOs are Outdated", "%1", _this);
```
bux: only `are` is written in lower case

PR #5191, opened 2017-05-22
[Medical Rewrite] Options to enable insta-death for AI

Arcanum417:
- Add setting to force AI die on FatalInjury
- Add setting to force AI die instead of going unconcious
- Fix setting module fail
- Add myself to authors


commy2:
Merge?

addons/medical/stringtable.xml
```diff
@@ -3397,12 +3397,24 @@
...
         </Key>
-         <Key ID="STR_ACE_Medical_MedicalSettings_fatalInjuryCondition_InCardiacArrest">
+        <Key ID="STR_ACE_Medical_MedicalSettings_fatalInjuryConditionAI_DisplayName">
+            <English>Instant death AI</English>
+        </Key>
+        <Key ID="STR_ACE_Medical_MedicalSettings_fatalInjuryConditionAI_Description">
+            <English>Defines if AI will be able to die instantly</English>
+        </Key>
+        <Key ID="STR_ACE_Medical_MedicalSettings_unconsciousConditionAI_DisplayName">
+            <English>Forbid unconscious AI</English>
```
SilentSpike: I feel like this setting is unnecessarily inverted, it should be "Enable AI Unconsciousness"

(However yeah we could merge and fix this in the main PR)

Arcanum417: Yup, descriptions and such are inverted and funky, I added the conditions in statemachine and those are kinda notes to self to remember what does it really do. The descriptions are logically correct and describe what the condition really does. The values in code should be inverted so the options could be described sensibly.

PR #5205, opened 2017-05-26
Add Ability to Change Repair Times of Vehicles

654wak654:
- Allow repair times of vehicles to be changed by script or config. Currently there is a single global value used for every repair operation, which isn't convenient if you're trying to amp up your saturday go kart nights but still keep convoy ops realistic.
- I just added a simple check for `GVAR(repairTime)` both as a variable on the vehicle and a config property. If either is found the function skips the regular method of determining time and goes with the new value.

### DIFF
--- a/addons/repair/functions/fnc_repair.sqf
+++ b/addons/repair/functions/fnc_repair.sqf
@@ -189,7 +189,7 @@ if (_repairTime < 0) then {
         getNumber (_config >> "repairingTime");
     } else {
         if (isText (_config >> "repairingTime")) exitWith {
-            _repairTimeConfig = getText(_config >> "repairingTime");
+            _repairTimeConfig = getText (_config >> "repairingTime");
             if (isNil _repairTimeConfig) then {
                 _repairTimeConfig = compile _repairTimeConfig;
             } else {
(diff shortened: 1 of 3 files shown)


654wak654:
This makes every type of repair take the same amount of time 🤔

PabstMirror:
Maybe make it a coefficent, defaulting to 1
Also I think a setVar should have higher priority than the new config.

654wak654:
setVar _did_ have a higher priority, It's gone for now.

SilentSpike:
Shouldn't repair time be based on amount of damage somehow? With a coefficient to modify how fast you can repair (which could also be setVar on individual players).

654wak654:
The time can be dependent on your engineer level (none, engineer, repair specialist), I thought of it too but it's not really in the scope of this small PR. Additionally, the damage level doesn't change the repair time for all repair types. Changing a wheel takes the same amount of time weather it has 2 bullets in it or 20.

addons/repair/functions/fnc_repair.sqf
```diff
@@ -177,23 +177,31 @@ if (vehicle _caller == _caller && {_callerAnim != ""}) then {
...
-            _repairTimeConfig;
+// Get repair time
+_repairTime = [
+    configFile >> "CfgVehicles" >> typeOf _target >> QGVAR(repairTimes) >> configName _config,
+    "number",
+    -1
+] call CBA_fnc_getConfigEntry;
+
+if (_repairTime < 0) then {
+    _repairTime = if (isNumber (_config >> "repairingTime")) then {
```
Phyma: Personal preference maybe but I think its better to have the if statement alone and set the _repairTime in each section.
So in the if statement you have _repairTime = getNumber() and in the else you have something else. It's hard to read what is happening like it is now.

PR #5206, opened 2017-05-26
Linking belt refactoring

Phyma:
startLinkingbelt now uses canLinkBelt so if condition needs to be changed then you can do it on one position.

Nothing will change with this, only easier to edit condition in the future.
First time edit so please point out malpractice.

### DIFF
--- a/addons/reload/functions/fnc_startLinkingBelt.sqf
+++ b/addons/reload/functions/fnc_startLinkingBelt.sqf
@@ -16,15 +16,10 @@ params ["_player", "_target"];
 if (vehicle _target != _target) exitWith {false};
 
 private _magazineType = currentMagazine _target;
-private _magazineCfg = configFile >> "CfgMagazines" >> _magazineType;
 
-if (getNumber (_magazineCfg >> "ACE_isBelt") == 0) exitWith {false};
+private _canLink = [_player, _target] call FUNC(canLinkBelt);
 
-// Check if the ammo is not empty or full
-private _ammoCount = _target ammo currentWeapon _target;
-
-// Exit if the belt is full or empty
-if ((_ammoCount == 0)  || (getNumber (_magazineCfg >> "count") - _ammoCount) == 0) exitWith {false};
+if(!_canLink) exitWith{};
 
 // Check if the player has any of the same same magazines
 // Calculate max ammo it can link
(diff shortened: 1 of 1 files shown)


addons/reload/functions/fnc_startLinkingBelt.sqf
```diff
@@ -16,15 +16,10 @@ params ["_player", "_target"];
...
 
-if (getNumber (_magazineCfg >> "ACE_isBelt") == 0) exitWith {false};
+private _canLink = [_player, _target] call FUNC(canLinkBelt);
 
-// Check if the ammo is not empty or full
-private _ammoCount = _target ammo currentWeapon _target;
-
-// Exit if the belt is full or empty
-if ((_ammoCount == 0)  || (getNumber (_magazineCfg >> "count") - _ammoCount) == 0) exitWith {false};
+if(!_canLink) exitWith{};
```
jonpas: Space before `(` and `{`.

Phyma: Fixed thanks for input :)

addons/reload/functions/fnc_startLinkingBelt.sqf
```diff
@@ -16,15 +16,10 @@ params ["_player", "_target"];
...
 
-if (getNumber (_magazineCfg >> "ACE_isBelt") == 0) exitWith {false};
+private _canLink = [_player, _target] call FUNC(canLinkBelt);
 
-// Check if the ammo is not empty or full
-private _ammoCount = _target ammo currentWeapon _target;
-
-// Exit if the belt is full or empty
-if ((_ammoCount == 0)  || (getNumber (_magazineCfg >> "count") - _ammoCount) == 0) exitWith {false};
+if ( !_canLink ) exitWith {} ;
```
Phyma: Okey!

addons/reload/functions/fnc_startLinkingBelt.sqf
```diff
@@ -16,15 +16,10 @@ params ["_player", "_target"];
...
 
 // Check if the player has any of the same same magazines
 // Calculate max ammo it can link
```
Phyma: Okey but I wanted to keep that FUNC(canLinkBelt) returned a bool and not a number.

So thats why i kept the lower part in both cases. But I could redo both of the files to keep them 100% unique?

Just say the word nad ill do it :)

jonpas: I think it would be a good idea to cut down on duplicate code if nothing else.

Phyma: #5213

PR #5208, opened 2017-05-27
Fixed zero divisor when throwing Chemlight(IR)

Phyma:
Fixed zero divisor when throwing Chemlight(IR).
Will fix zero divisor bug because the function throwFiredXEH was called with VAR instead of FUNC.


PabstMirror:
Don't think this is right, the event name needs to match the handler added in postInit:
https://github.com/acemod/ACE3/blob/master/addons/advanced_throwing/XEH_postInit.sqf#L4

jonpas:
Yeah, no, this breaks the XEH event actually.

Phyma:
Yeah noticed it aswell will continue look for the problem

PR #5209, opened 2017-05-27
Advanced throw zero divisor

Phyma:
Missing call.

All other places have call except for that when adding a event.

Bug does not appear. Not tested on multiplayer.


PabstMirror:
Passing code is fine.
Sometimes we will do `{_this call FUNC(x)}` which is really almost the same thing (tiny overhead)

`CBA_fnc_addEventHandler` needs code as 2nd arg
`call FUNC(x)` isn't code, it could return anything

I think I've got a fix for it in #5210

PR #5212, opened 2017-05-27
Usage of BIS HandleDamage Array

Phyma:
https://community.bistudio.com/wiki/Arma_3:_Event_Handlers#HandleDamage
Has hitPoint as a input so use that instead of try and find it by ourself.

This changed the behaviour of damage pretty hard, more insta explosions and only cook off some of the time more than i expected so please try and verify if this is proper behaviour.

Test mission was RHS Launchers and RHS vehicles.

### DIFF
--- a/addons/cookoff/functions/fnc_handleDamage.sqf
+++ b/addons/cookoff/functions/fnc_handleDamage.sqf
@@ -16,7 +16,7 @@
 #include "script_component.hpp"
 
 params ["_simulationType", "_thisHandleDamage"];
-_thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIndex", "_shooter", "_hitPoint"];
+_thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIndex", "_shooter", "_hitLoc"];
 
 // it's already dead, who cares?
 if (damage _vehicle >= 1) exitWith {};
@@ -25,7 +25,7 @@ if (damage _vehicle >= 1) exitWith {};
 private _hitpoint = "#structural";
 
 if (_hitIndex != -1) then {
-    _hitpoint = toLower (_hitPoint);
+    _hitpoint = toLower (_hitLoc);
 };
 
 // get change in damage
--- a/addons/cookoff/functions/fnc_handleDamage.sqf
+++ b/addons/cookoff/functions/fnc_handleDamage.sqf
@@ -16,21 +16,15 @@
 #include "script_component.hpp"
 
 params ["_simulationType", "_thisHandleDamage"];
-_thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIndex", "_shooter", "_hitLoc"];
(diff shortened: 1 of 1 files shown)


addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -16,13 +16,13 @@
...
 
 params ["_simulationType", "_thisHandleDamage"];
-_thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIndex", "_shooter"];
+_thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIndex", "_shooter", "_hitPoint"];
 
 // it's already dead, who cares?
 if (damage _vehicle >= 1) exitWith {};
 
-// get hitpoint name
-private _hitpoint = "#structural";
```
PabstMirror: `"#structural"` is a special var that would be lost by this
you might be able to replace line 28, but IIRC there might be some problem with sub-turret hitpoints

addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -16,16 +16,16 @@
...
 // it's already dead, who cares?
 if (damage _vehicle >= 1) exitWith {};
 
-// get hitpoint name
+// set default value
 private _hitpoint = "#structural";
 
 if (_hitIndex != -1) then {
-    _hitpoint = toLower ((getAllHitPointsDamage _vehicle param [0, []]) select _hitIndex);
+    _hitpoint = toLower (_hitPoint);
```
PabstMirror: need to account for line 25 which overwrites the `params` var
I remember @commy2 saying this was a bad idea for some reason, but I forget why

Phyma: but vars are casesensitive. arent they?
Sure it could be clarified with using another var that _hitPoint in the param.
params use _hitPoint the rest is _hitpoint.

SilentSpike: `_hitpoint = ["#structural", toLower _hitpoint] select (_hitIndex != -1);`

SilentSpike: To clarify:

- Variables in sqf are case insensitive
- There should be no issue overriding the params variable, it's already private though so no need for `private`

addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -16,21 +16,15 @@
 #include "script_component.hpp"
 
 params ["_simulationType", "_thisHandleDamage"];
-_thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIndex", "_shooter"];
+_thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIndex", "_shooter", "_hitpoint", "_oldDamage"];
```
commy2: Don't like `_oldDamage` here. It's not a parameter of the event and should not be abused to `private` a variable. There could come a point where the parameter is used by the event!

bux:
Can we close this?

PR #5214, opened 2017-05-28
Fix zeus unit and group interactions locality

SilentSpike:
- The `setUnitPos`, `setFormation`, `setSpeedMode` and `setBehaviour` commands require local arguments
- The waypoint equivalents require server execution

### DIFF
--- a/addons/interaction/XEH_postInit.sqf
+++ b/addons/interaction/XEH_postInit.sqf
@@ -60,7 +60,6 @@ GVAR(isOpeningDoor) = false;
 },
 [57, [false, true, false]], false] call CBA_fnc_addKeybind; //Key CTRL+Space
 
-
 ["ACE3 Common", QGVAR(tapShoulder), localize LSTRING(TapShoulder), {
     // Conditions: canInteract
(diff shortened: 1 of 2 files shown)


SilentSpike:
So I went ahead and tested the locality of all the commands not marked on the biki. Some interesting results, apparently the waypoint commands need to be executed on the server, rather than where the group is local. Will update code and biki with this information.

addons/interaction/XEH_postInit.sqf
```diff
@@ -79,3 +78,33 @@ GVAR(isOpeningDoor) = false;
 
 ["isNotSwimming", {!underwater (_this select 0)}] call EFUNC(common,addCanInteractWithCondition);
 ["isNotOnLadder", {getNumber (configFile >> "CfgMovesMaleSdr" >> "States" >> animationState (_this select 0) >> "ACE_isLadder") != 1}] call EFUNC(common,addCanInteractWithCondition);
+
+// Zeus action events
```
PabstMirror: These EH are below line 31 - `if (!hasInterface) exitWith {};`

SilentSpike: Good catch :+1:

PR #5216, opened 2017-05-29
AdvThrow - Handle bugged currentThrowable

PabstMirror:
image

`currentThrowable` can get glitched and show a magazine that you don't have in inventory
Advanced throwing lets you throw unlimited grenades.
Effects vanilla throw as well, although there you can only do a single throw of a `x0` grenade.

This also breaks `ace_weaponselect_fnc_selectNextGrenade` because removing has no effect

### DIFF
--- a/addons/advanced_throwing/functions/fnc_drawThrowable.sqf
+++ b/addons/advanced_throwing/functions/fnc_drawThrowable.sqf
@@ -36,6 +36,11 @@ if (_throwable isEqualTo [] && {!_primed}) exitWith {
 
 private _throwableMag = _throwable param [0, "#none"];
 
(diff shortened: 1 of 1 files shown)


PabstMirror:
https://feedback.bistudio.com/T125196

jonpas:
Should add to https://ace3mod.com/wiki/development/arma-3-issues.html

commy2:
>Unit B opens their backpack and removes it
You cannot open the backpack of another unit in the base game.

PabstMirror:
my video was taken with no mods

jonpas:
Of course you can @commy2, we only add camera shake and sound on open of your backpack.

PR #5219, opened 2017-05-30
Fix zeus RC units using player damage treshold

### LINKED ISSUE #5218: RC Units Inherit Player Damage Threshold
Is it possible to stop units that zeus remote controls from inheriting player medical settings, makes for some really awkward situations

Cuel:
Closes #5218

Use the boolean parameter for common_isPlayer to exclude remote controlled units


PabstMirror:
This will work differently in rewrite, but for now I think this is reasonable.
I think we might want to use the setting `GVAR(remoteControlledAI)` (default: true) instead.

```
<English>Treat remote controlled units as AI not players?</English>
```

PR #5222, opened 2017-05-31
Fix ammo count hidden in UAVs

PabstMirror:
Fixes UAVs having no ammo count displayed

@jonpas - this could also be fixed in
https://github.com/acemod/ACE3/blob/master/addons/reload/ACE_UI.hpp#L4
wasn't sure which was best place to put it


jonpas:
Should indeed be put in reload condition, setAdvancedElement is aan abstracted function, some UI element might not be wanted on them either (all vehicle info for example), this would counteract it.

PR #5223, opened 2017-05-31
Only show zeus interactions when thing is selected

PabstMirror:
Hides actions that aren't appropriate based on what is selected.


SilentSpike:
I'd also add a condition to the markers actions (`select 3`) even though there are no actions in there. It's possible that someone somewhere has made an addon which adds some (also just for consistency).

PR #5226, opened 2017-05-31
Open glass and CUP doors

Phyma:
Will make sure that you can Ace slow open glass doors.
Also a bit flexible animation gathering instead of fixed strings.
CUP houses now works to incremental open!

The reason behind getGlassDoor is that if you do the regular door check you will get the hide and unhide of the glass pane. So you then need to find the closest doorobject because thats the door.

### DIFF
--- a/addons/interaction/functions/fnc_getDoor.sqf
+++ b/addons/interaction/functions/fnc_getDoor.sqf
@@ -39,9 +39,9 @@ _door = _intersections select 0 select 0;
 
 //Check if door is glass because then we need to find the proper location of the door so we can use it
 if (["glass", _door] call BIS_fnc_inString) then {
-	_glassDoor = [_house, _door] call FUNC(getGlassDoor);
-	_house = _glassDoor select 0;
-	_door = _glassDoor select 1;
+    _glassDoor = [_house, _door] call FUNC(getGlassDoor);
+    _house = _glassDoor select 0;
+    _door = _glassDoor select 1;
 };
 
 if (isNil "_door") exitWith {[_house, ""]};
--- a/addons/interaction/functions/fnc_getDoor.sqf
+++ b/addons/interaction/functions/fnc_getDoor.sqf
@@ -41,7 +41,7 @@ if (isNil "_door") exitWith {[_house, ""]};
 
 //Check if door is glass because then we need to find the proper location of the door so we can use it
 if ((_door find "glass") != -1) then {
-    _door = [_distance, _house, _door] call FUNC(getGlassDoor);   
+    _door = [_distance, _house, _door] call FUNC(getGlassDoor);
 };
 
 if (isNil "_door") exitWith {[_house, ""]};
--- a/addons/interaction/functions/fnc_getDoor.sqf
+++ b/addons/interaction/functions/fnc_getDoor.sqf
@@ -1,5 +1,5 @@
 /*
- * Author: commy2
+ * Author: commy2, Phyma
  * Find door.
  *
  * Arguments:
@@ -36,6 +36,50 @@ if (typeOf _house == "") exitWith {[objNull, ""]};
 _intersections = [_house, "GEOM"] intersect [_position0, _position1];
 
 _door = _intersections select 0 select 0;
+
+private _doorParts = [];
+
+//Check if door is glass because then we need to find the proper location of the door so we can use it
+if (["glass", _door] call BIS_fnc_inString) then {
+
+	private ["_animName", "_splitStr", "_newString", "_dist", "_animate"];
+	private _config = _house call CBA_fnc_getObjectConfig;	
+	
+	private _animate = configProperties [_config >> "AnimationSources", "true", false];
+	
+	// calculate all animation names so we know what is there	
+	{			
+		_animName = configName _x;
+		if ((["door", _animName] call BIS_fnc_inString) && !(["locked", _animName] call BIS_fnc_inString) && !(["disabled", _animName] call BIS_fnc_inString) && !(["handle", _animName] call BIS_fnc_inString)) then {			
+			_splitStr = _animName splitString "_";			
+			_doorParts pushBack ((_splitStr select 0) + "_" + (_splitStr select 1) + "_trigger");
+		};
+	} forEach _animate;	
+	
+	private _glassDoor = _door splitString "_";	
+	private _glassPos = (_house selectionPosition [(_glassDoor select 0) + "_" + (_glassDoor select 1) + "_effects", "Memory"]);	
+	
+	private _doorPos = [];
+	{
+		_doorPos pushBack (_house selectionPosition [_x, "Memory"]);
+	} forEach _doorParts;
+	
+	private _lowestDistance = 0;	
+	{		
+		private _dist = _glassPos distance  _x;
+		if (_lowestDistance == 0) then {
+			_lowestDistance = _dist;
+			_newString = (_doorParts select _forEachIndex) splitString "_";			
+			_door = ((_newString select 0) + "_" + (_newString select 1) + "_rot");
+		} else {
+			if (_dist < _lowestDistance) then {
+				_lowestDistance = _dist;
+				_newString = (_doorParts select _forEachIndex) splitString "_";			
+				_door = ((_newString select 0) + "_" + (_newString select 1) + "_rot");				
+			};			
+		};							
+	} forEach _doorPos;		
+};
 
 if (isNil "_door") exitWith {[_house, ""]};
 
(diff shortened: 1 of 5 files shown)


BaerMitUmlaut:
You need to use 4 spaces for indentation instead of tabs. If you use a text editor like Sublime Text, Atom or Notepad++ you should be able to change that in your settings.

jonpas:
Or just install EditorConfig plugin for the editor you use and it will be automatic (we have configuration for it).

Phyma:
Changes to Pullrequest:
CUP houses now works to incremental open!

Phyma:
Known issue when on CUP houses if you have a door that is both glass and door then if you aim on the glass it wont open. but if you aim on the doorframe then it works. it wont make a click sound when on the glass.

jonpas:
Should use `STRING find SUBSTRING != -1` instead of BI function.

addons/interaction/functions/fnc_getDoorAnimations.sqf
```diff
@@ -22,102 +22,19 @@ params ["_house", "_door"];
...
-    ["", ""],
-    ["", ""],
-    ["", ""]
-] select _index;
+private _animate = animationNames _house;
+private _animations = [];
+private _lockedVariable = [];
+
+{
+    _animName = _x;
```
commy2: missing `private`

addons/interaction/functions/fnc_getDoorAnimations.sqf
```diff
@@ -22,102 +22,19 @@ params ["_house", "_door"];
 
 private ["_index", "_animations", "_lockedVariable"];
```
commy2: delete this

addons/interaction/functions/fnc_getDoor.sqf
```diff
@@ -39,4 +39,11 @@ _door = _intersections select 0 select 0;
 
 if (isNil "_door") exitWith {[_house, ""]};
 
+//Check if door is glass because then we need to find the proper location of the door so we can use it
+if (["glass", _door] call BIS_fnc_inString) then {
+    _door = [_distance, _house, _door] call FUNC(getGlassDoor);   
+};
+
+if (isNil "_door") exitWith {systemChat "Wrong"; [_house, ""]};
```
commy2: debug doesn't use `systemChat`

addons/interaction/functions/fnc_getGlassDoor.sqf
```diff
@@ -0,0 +1,67 @@
...
+
+private ["_animName", "_splitStr", "_newString", "_objDist", "_animate"];
+private _doorParts = [];
+private _doorPos = [];
+private _animate = animationNames _house;
+private _glassDoor = _door splitString "_"; 
+private _glassPos = (_house selectionPosition [(_glassDoor select 0) + "_" + (_glassDoor select 1) + "_effects", "Memory"]);     
+// Calculate all animation names so we know what is there   
+{           
+    _animName = _x;
```
commy2: `private` here. The variable is not used outside this scope

addons/interaction/functions/fnc_getGlassDoor.sqf
```diff
@@ -0,0 +1,67 @@
...
+    if ((["door", _animName] call BIS_fnc_inString) && !(["locked", _animName] call BIS_fnc_inString) && !(["disabled", _animName] call BIS_fnc_inString) && !(["handle", _animName] call BIS_fnc_inString)) then {         
+        _splitStr = _animName splitString "_";
```
commy2: `private` here. The variable is not used outside this scope

addons/interaction/functions/fnc_getGlassDoor.sqf
```diff
@@ -0,0 +1,67 @@
...
+// Calculate what door that is closest to the glass door
+private _lowestDistance = 0;    
+{       
+    private _objDist = _glassPos distance  _x;
+    
+    if (_objDist <= _distance) then {    
+        //Need to set the value in the beginning
+        if (_lowestDistance == 0) then {
+            _lowestDistance = _objDist;
+            _newString = (_doorParts select _forEachIndex) splitString "_";
```
commy2: `private` here. The variable is not used outside this scope. No reason not to reuse the same identifier for a temporary value as `_splitStr` uses a few blocks before this one.

addons/interaction/functions/fnc_getGlassDoor.sqf
```diff
@@ -0,0 +1,67 @@
...
+ * Example:
+ * [player, target] call ace_interaction_fnc_getGlassDoor
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_distance", "_house", "_door"]; 
+
+private ["_animName", "_splitStr", "_newString", "_objDist", "_animate"];
```
commy2: delete this
`_objDist` and `_animate` are already set to private in their scopes.

addons/interaction/functions/fnc_getGlassDoor.sqf
Phyma: changed _newString to _splitStr and decaled `private ["_splitStr"];` just over  `if (_objDist <= _distance) then {`. Okey?
Will try so everything works later will push after tested.

commy2: Just use `private _splitStr =` here and in the if block below, as well as in L32 above.
They are unrelated and only temporary, so there is no need for them to be set to the same scope. Yet the concept is the same so they should use the same identifier.

PabstMirror:
This seems to be working pretty good
- opening altis/stratis glass doors 
(except the sliding ones on the airport terminal, but I think they are extra weird)
- working on CUP_Terrain buildings

addons/interaction/functions/fnc_getGlassDoor.sqf
```diff
@@ -42,26 +41,26 @@ private _glassPos = (_house selectionPosition [(_glassDoor select 0) + "_" + (_g
 // Calculate what door that is closest to the glass door
 private _lowestDistance = 0;    
 {       
-    private _objDist = _glassPos distance  _x;
-    
+    private _objDist = _glassPos distance  _x;
```
commy2: white space :P

addons/interaction/functions/fnc_getDoor.sqf
```diff
@@ -35,7 +35,14 @@ if (typeOf _house == "") exitWith {[objNull, ""]};
...
 _intersections = [_house, "GEOM"] intersect [_position0, _position1];
 
-_door = _intersections select 0 select 0;
+_door = toLower (_intersections select 0 select 0);
+
+if (isNil "_door") exitWith {[_house, ""]};
+
+//Check if door is glass because then we need to find the proper location of the door so we can use it
+if ((_door find "glass") != -1) then {
+    _door = [_distance, _house, _door] call FUNC(getGlassDoor);
```
commy2: white space

addons/interaction/functions/fnc_getDoorAnimations.sqf
```diff
@@ -20,104 +20,19 @@
...
-    ["", ""],
-    ["", ""]
-] select _index;
+private _animate = animationNames _house;
+private _animations = [];
+private _lockedVariable = [];
+
+{
+    private _animName = toLower _x;    
+    if ((_animName find (toLower _door)) != -1) then {
```
commy2: white space

addons/interaction/functions/fnc_getGlassDoor.sqf
```diff
@@ -0,0 +1,66 @@
...
+            };          
+        };
+    };        
+} forEach _doorPos;
+
+// Check if we have a door or if it is the glass part
+if ((isNil "_door") || ((_door find "glass") != -1)) exitWith {};
+
+_door
+
```
commy2: many lines with white space.
Sublime Text 3 is the best editor

Phyma: Well white lines are nice for readability. Segmenting the code, there is not a single white line in a segment. A segment could be a forEach and it wont affect the performance of the code. So I'm hesitating to do this change.

jonpas: Commy means spaces (literally) after text. There are a few lines in this file that have that. 

I suggest you install EditorConfig plugin for your editor, it will automatically clean that up. Otherwise you can probably also manually set it.

PabstMirror: Some text editors let you visualize whitespace, 

Notepad++: View -> Show Symbol -> Show White Space and Tab
Sublimb Text: `"draw_white_space": "all"`
I think Atom is "Show Inadvisables"

EditorConfig should also clean these up automatically and prevent things like tabs from being saved

Phyma: Okey I fixed sublime with EditorConfig and all that setup so should not be a problem now. It should say something about this in the Developer wiki.

bux: could be added somewhere to:
https://ace3mod.com/wiki/development/coding-guidelines.html

PR #5234, opened 2017-06-02
Raise lower carry object

Phyma:
- Will fix  bug where you cant raise or lower some objects
- Doesn't use setPosATL but uses the internal ace_dragging_carryingPos

### DIFF
--- a/addons/dragging/functions/fnc_dropObject_carry.sqf
+++ b/addons/dragging/functions/fnc_dropObject_carry.sqf
@@ -31,6 +31,11 @@ private _inBuilding = [_unit] call FUNC(isObjectOnObject);
 // release object
 detach _target;
 
+// Set to default value
+_parentConfig = [_target] call CBA_fnc_getObjectConfig;
+_defaultPos = getArray(_parentConfig >> QGVAR(carryPosition));
+_target setVariable [QGVAR(carryPosition), _defaultPos, true];
+
 // fix anim when aborting carrying persons
 if (_target isKindOf "CAManBase" || {animationState _unit in CARRY_ANIMATIONS}) then {
     if (vehicle _unit == _unit && {!(_unit getVariable ["ACE_isUnconscious", false])}) then {
--- a/addons/dragging/functions/fnc_dropObject_carry.sqf
+++ b/addons/dragging/functions/fnc_dropObject_carry.sqf
@@ -31,11 +31,6 @@ private _inBuilding = [_unit] call FUNC(isObjectOnObject);
 // release object
 detach _target;
 
-// Set to default value
-private _parentConfig = [_target] call CBA_fnc_getObjectConfig;
-private _defaultPos = getArray(_parentConfig >> QGVAR(carryPosition));
-_target setVariable [QGVAR(carryPosition), _defaultPos, true];
-
 // fix anim when aborting carrying persons
 if (_target isKindOf "CAManBase" || {animationState _unit in CARRY_ANIMATIONS}) then {
     if (vehicle _unit == _unit && {!(_unit getVariable ["ACE_isUnconscious", false])}) then {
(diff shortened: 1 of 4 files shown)


PabstMirror:
This has some problems with being able to scroll the item below ground and breaks on objects manually set with `ace_dragging_fnc_setCarryable`.

But it shows that the real problem is `setPosATL` doesn't have immediate effects when object is not local.
So I think a better solution would be to just replace handleScrolWheel 40,41 with
```
private _positionChange =  _position vectorDiff (getPosATL _carriedItem);
private _selectionPosition = _unit worldToModel (ASLtoAGL getPosWorld _carriedItem);
_selectionPosition = _selectionPosition vectorAdd _positionChange;
_carriedItem attachTo [_unit, _selectionPosition];
```

Phyma:
Is that code tested? Dosent work at all. The item just disappears.
Can check later what is wrong with that code.

addons/dragging/functions/fnc_dropObject_carry.sqf
```diff
@@ -31,6 +31,11 @@ private _inBuilding = [_unit] call FUNC(isObjectOnObject);
 // release object
 detach _target;
 
+// Set to default value
+private _parentConfig = [_target] call CBA_fnc_getObjectConfig;
+private _defaultPos = getArray(_parentConfig >> QGVAR(carryPosition));
+_target setVariable [QGVAR(carryPosition), _defaultPos, true];
```
PabstMirror: This breaks `ace_dragging_fnc_setCarryable` because the object has no config, so it stores `[]`

PabstMirror:
https://github.com/acemod/ACE3/commit/df147b8dc605faa8263aae84b99e0557ed4856a2
Seems to work in MP

Phyma:
Sorry. I didnt clean all the lines of code. So forgot to change back to the original _position declaration.
Works on Dedicated aswell.

Phyma:
Still real new to the 100 different ways of taking out a position and what space it is in when you take it out.

SilentSpike:
I know your pain :smile:

Phyma:
Will post when this is ready to be reviewed! still making changes

jonpas:
One day you'll think you've grasped it all, just to realize you aren't even close.

addons/dragging/functions/fnc_handleScrollWheel.sqf
```diff
@@ -37,8 +37,10 @@ _position set [2, ((_position select 2) + _scrollAmount min (_maxHeight + 1.5))
 
 // move up/down object and reattach at current position
 detach _carriedItem;
-_carriedItem setPosATL _position;
-_carriedItem attachTo [_unit];
+private _positionChange =  _position vectorDiff (getPosATL _carriedItem);
```
jonpas: 2 spaces on the right side of `=`.

addons/dragging/functions/fnc_handleScrollWheel.sqf
```diff
@@ -37,7 +37,9 @@ _position set [2, ((_position select 2) + _scrollAmount min (_maxHeight + 1.5))
 
 // move up/down object and reattach at current position
 detach _carriedItem;
-private _positionChange =  _position vectorDiff (getPosATL _carriedItem);
+
+// Uses this method of selecting position because setATL dosen't work on dedicated server
```
jonpas: `doesn't`

That's not entirely true though, it only doesn't work correctly when object is not local, if I understood the core issue correctly.

Phyma: Tha is true sorry

PabstMirror: `setPosATL` does work on remote objects, but from what I saw it didn't have immediate effects
At least in the same frame:
```
x setPosATL y;
(getPosATL x) != y
```

PR #5246, opened 2017-06-04
Use getUnitTrait for medics, engineers and EOD

SilentSpike:
- Make the functions which check whether units are medics, engineers or explosive specialists use the newer `getUnitTrait` command which makes them compatible with use of `setUnitTrait` and not purely checking the config

### DIFF
--- a/addons/medical/functions/fnc_isMedic.sqf
+++ b/addons/medical/functions/fnc_isMedic.sqf
@@ -19,7 +19,7 @@
 
 params ["_unit", ["_medicN", 1]];
 
-private _class = _unit getVariable [QGVAR(medicClass), _unit getUnitTrait "medic"];
+private _class = _unit getVariable [QGVAR(medicClass), [0, 1] select (_unit getUnitTrait "medic")];
 
 if (_class >= _medicN min GVAR(medicSetting)) exitWith {true};
 if (!GVAR(increaseTrainingInLocations)) exitWith {false};
--- a/addons/medical/functions/fnc_isMedic.sqf
+++ b/addons/medical/functions/fnc_isMedic.sqf
@@ -19,7 +19,7 @@
 
 params ["_unit", ["_medicN", 1]];
 
-private _class = _unit getVariable [QGVAR(medicClass), getNumber (configFile >> "CfgVehicles" >> typeOf _unit >> "attendant")];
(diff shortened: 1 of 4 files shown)


addons/medical/functions/fnc_isMedic.sqf
```diff
@@ -19,7 +19,7 @@
 
 params ["_unit", ["_medicN", 1]];
 
-private _class = _unit getVariable [QGVAR(medicClass), getNumber (configFile >> "CfgVehicles" >> typeOf _unit >> "attendant")];
+private _class = _unit getVariable [QGVAR(medicClass), _unit getUnitTrait "medic"];
```
PabstMirror: _class should be number, not bool

someone in public slack was talking about using config `attendant =2` to set doctors
it's not the prettiest code, but something like this should support that
```
private _class = _unit getVariable QGVAR(medicClass);
if (isNil "_class") then {
    if (_unit getUnitTrait "medic") then {
        _class = 1 max getNumber (configFile >> "CfgVehicles" >> typeOf _unit >> "attendant");
    } else{
        _class = 0;
    };
};
```

SilentSpike: Are we sure we want to support that since afaik the game itself treats the `attendant` property as a boolean?

SilentSpike: It might be better to introduce an extra config property to support doctors that way (although, are doctors even relevant in the current medical code?)

Phyma: ACE_common_fnc_getUnitTrait that will check all traits? Input of what you need to check. This will also help later if we decide to implement more types of medics, engineers and more.
Thoughts?

jonpas: I think that should be done in medical-rewrite, as we can afford to change things there that would otherwise be backwards-incompatible.

SilentSpike:
I didn't add the code for `attendant` being set to 2 as I'm unsure about whether we want to do that, but fixed the incorrect boolean

PabstMirror:
Could replace the `attendant = 2;` config with 
```
    class x_doctorUnitType {
        class makeDoctor {
            init = "(_this select 0) setVariable ["ace_medical_medicClass", 2]";
        };
    };
};
```

PR #5253, opened 2017-06-06
Add public function documentation script

SilentSpike:
- Remove obsolete event_rename.py script
- Introduce new python script to the docs/tools directory to generate public function documentation

Current Status:
- [x] Crawls all function files and identifies which are public
  - [x] Supports optional argument to specify a specific addon to crawl
- [x] Reports function files with malformed headers
  - [x] Supports optional `--debug` argument to only check for header issues
  - [x] Missing or misplaced header
  - [x] Missing `Public` value
  - [x] Invalid `Public` value
- [x] Processes the header information
  - [x] Split the header into sections
  - [x] Process the author and description
  - [x] Process the arguments
  - [x] Process the return value
  - [x] Process the example
- [x] Writes to HTML/whatever


SilentSpike:
So, this is a bit of a weird one as I know there are some plans involving the use of http://www.sphinx-doc.org/

I've spent some time messing with that and ran into a few walls so for the time being decided to work on this python script. I've yet to look at what output format we need the data in, but am currently mulling over whether there's a better approach to splitting the header into sections than `re.split` as it results in the following: 
```python
["useless stuff", "Author", "(authors) \n (description text)", "Argument", "(arguments text)", "Return Value", "(return value text)", "Example", "(examples text)", "Public", "(public text)"]
```

SilentSpike:
So this is basically almost finished, the only question is how we want the data to actually be output. Currently it will spit out .md files (named after component, i.e. `medical.md`) in the `wiki/functions` folder of the form:

## ace_common_fnc_addSyncedEventHandler
__Description__

Register an event handler for an ACE synced event

__Parameters__

Index | Description | Datatype(s) | Default Value
--- | --- | --- | ---
0 | Name  | STRING | 
1 | Handler  | CODE | 
? | Malformed | ? | ?

__Return Value__

Description | Datatype(s)
--- | ---
Boolean of success  | BOOL

__Example__

```sqf
["myEvent", {_this call x}, 0] call ace_common_fnc_addSyncedEventHandler
```


__Authors__

- jaynus

---

## ace_common_fnc_addToInventory
__Description__

Adds an item, weapon, or magazine to the unit's inventory or places it in a weaponHolder if no space.

__Parameters__

Index | Description | Datatype(s) | Default Value
--- | --- | --- | ---
0 | Unit  | OBJECT | 
1 | Classname  | STRING | 
2 | Container (uniform, vest, backpack) (default: "")  | STRING | 
3 | Magazine Ammo Count (default: -1)  | NUMBER | 

__Return Value__

Description | Datatype(s)
--- | ---
0: Added to player  | BOOL

__Example__

```sqf

```


__Authors__

- Garth 'L-H' de Wet

SilentSpike:
It currently just appends to the output files instead of overwriting them, but that should be solvable easily enough

jonpas:
I see you are opening the file for each function, so you'll just have to open it with `w` flag once before writing first function.

I am not sure what we want to do about removed functions though, I think we're better off just removing them from the doc anyways.

Would this run on Travis? Maybe only on `release` branch so when we merge `release` back into `master` it gets nicely updated at correct time?

SilentSpike:
Yeah, just a question of how best to do the control flow for it. I guess I can just add a quick and dirty solution for now, code can all be refined later.

SilentSpike:
Oh man this finally got merged? 😆

I can't remember if I wrote this with any semblance of logic, but hopefully it's in a relatively good state. Someone once contacted me on slack to ask for permission to adapt it to their project, may be worth looking into what they did if I can find who that was.

dedmen:
Can we get this into CI too? to make sure everyone pushes valid headers?

jonpas:
Make me a separate issue so I don't forget.

PR #5254, opened 2017-06-06
Fix some event names on wiki

PabstMirror:


docs/wiki/framework/cargo-framework.md
```diff
@@ -46,19 +46,18 @@ class staticBananaLauncher {
...
 
-Event Name | Description | Passed Parameter(s) | Locality
+Event Name | Passed Parameter(s) | Locality | Description
 ---------- | ----------- | ------------------- | --------
-`LoadCargo` | Load object into vehicle. | `[_object, _vehicle, _showHint]` | Local
-`UnloadCargo` | Unload object from vehicle. | `[_object, _vehicle]` | Local
-`AddCargoByClass` | Adds a cargo item to the vehicle. | `[_itemClass, _vehicle, _amount, _showHint]` | Local
+`ace_addCargo` | [_item (CLASSNAME or OBJECT), _vehicle, _cargoCount] | Target | Scripted way to add cargo to vehicle
+`ace_loadCargo` | `[_object, _vehicle, _showHint]` | Target | Load object into vehicle.
+`ace_unloadCargo` | `[_object, _vehicle]` | Target | Unload object from vehicle.
```
jonpas: I'd really say get rid of those "Callable" events, we should promote public functions instead (I wrote this back when I wasn't totally sure what I was doing :P).

SilentSpike: I agree with this, I think events should only be publicly listenable and public functions should take care of calling things where appropriate

PabstMirror:
Switched callable events to functions.
Also note that directly calling the functions won't trigger the listenable events (like `ace_cargoLoaded`)

PR #5255, opened 2017-06-06
Conform function headers to coding guidelines

Phyma:
- Fix every header on all files that i have on my fork
- Tested with Silentspikes python script #5253
- Even if the other thing is not accepted this is still good
- Alot of things that i have done is switch to current header, add example, add public and fixed "ReturnValue" to "Return Value"

I believe this is a label "impact/trivial" because it's only headers


addons/fcs/functions/fnc_onForceUpdate.sqf
```diff
@@ -1,3 +1,18 @@
+/*
+ * Author: ACE
+ *
+ *
+ * Arguments:
+ * Nothing
```
jonpas: This should be `None`, I noticed it was wrong in some other files before as well.

SilentSpike: The script will now complain about this too after latest commit 😄

Phyma: Yeay. Well we can put this as WIP so I can fix this in the same tempo that silentspike updates the script and we just do both at the same time :)

I was just bored and didnt feel like scripting so did this.

Phyma:
Done! No more errors! Plus new version of the software only public functions

bux:
DO IT NAOW!

PR #5261, opened 2017-06-08
Fix Large IED defuse distance

### LINKED ISSUE #5258: Large IEDs (pressure plate) exploding before defuse action is shown
**Arma 3 Version:** `1.70` (stable)
**ACE3 Version:** `3.9.2` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- Large IEDs (both urban and land versions) with pressure plates can currently not be defused. The range in which the explosive is triggered appears to be larger than the range in which the defuse action is shown. Looking at the relevant code leads me to think it is an issue with defuse action being shown in a 1 meter range from the camera (same as for small IEDs) and the explosive trigger going off in a 2 meter range from presumably the player object (opposed to 1 meter for small IED).
- Recommend either increasing the defuse action range for large IEDs or making the large IED ammo use the smaller mineTrigger like the small IEDs

**Steps to reproduce:**
- Spawn a unit with defuse kit
- Spawn either a large IED urban (pressure plate) or large IED land (pressure plate)
- Crawl toward IED slowly while checking if the defuse option is available
- Get blown up before the action is shown
- (optional) Try with small IED to show that it is specifically large IEDs

- Dedicated 
- Editor (Singleplayer)
- Editor (Multiplayer)

**Placed Modules:**
- None

**RPT log file:**
- https://pastebin.com/JAiFjiYu

Phyma:
- IED will get size variable. 0 for small, 1 for large.
- This size will determine how far the defuse range should be, might be used for other things aswell.
- Increase defusal range for large IEDs to replicate behaviour of small IEDs, you can now crouch to big IEDs but be aware!
- Will add Tip on wiki about defusing 
- Will fix #5258 

PS I know _size == 1 might be redundant but its how I write because _size is formally a number not a bool then you have to compare it and not just if (_size)

### DIFF
--- a/addons/explosives/CfgAmmo.hpp
+++ b/addons/explosives/CfgAmmo.hpp
@@ -27,10 +27,13 @@ class CfgAmmo {
     class PipeBombCore: TimeBombCore;
     class PipeBombBase: PipeBombCore;
     */
+    // GVAR(size) = 0; is small size
+    // GVAR(size) = 1; is large size
     class DirectionalBombBase;
     class ClaymoreDirectionalMine_Remote_Ammo: DirectionalBombBase {
         GVAR(magazine) = "ClaymoreDirectionalMine_Remote_Mag";
         GVAR(Explosive) = "ClaymoreDirectionalMine_Remote_Ammo_Scripted";
+        GVAR(size) = 0;
         GVAR(defuseObjectPosition)[] = {0, 0, 0.038};
         soundActivation[] = {"", 0, 0, 0};
         soundDeactivation[] = {"", 0, 0, 0};
@@ -39,6 +42,7 @@ class CfgAmmo {
 
     class APERSTripMine_Wire_Ammo: DirectionalBombBase {
         GVAR(defuseObjectPosition)[] = {-1.415, 0, 0.12};
+        GVAR(size) = 0;
     };
 
     class ACE_FlareTripMine_Wire_Ammo: APERSTripMine_Wire_Ammo {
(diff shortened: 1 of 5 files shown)


Phyma:
Last commit is ofc. Fixed typo hehe

PabstMirror:
I still couldn't defuse the big ied without exploding
```
		class RangeTrigger: Default {
			mineTriggerType = "radius";
			mineTriggerRange = 2;
```
is more than the large's 1.5m

so maybe we need a huge size?

Phyma:
I could do it. But it was hard. Maybe also increase defusalrange from
1.5->2 for large?

PabstMirror:
I think it explodes at 2, so probably at least 2.5

Phyma:
But if I could defuse it at 1.5m with my code then 2 should be easier but
not to easy right?

PabstMirror:
Adding notes from slack
```
`cursorObject distance (player modelToWorld (player selectionPosition "pilot"))` = 1.35
`cursorObject distance player` = 2.02676
```
So that's why prone is safer
and it's possible to defuse large IED while crouched

addons/explosives/CfgAmmo.hpp
```diff
@@ -27,10 +27,13 @@ class CfgAmmo {
     class PipeBombCore: TimeBombCore;
     class PipeBombBase: PipeBombCore;
     */
+    // GVAR(size) = 0; is small size
+    // GVAR(size) = 1; is large size
```
jonpas: This should be put into the [framework page of documentation](https://ace3mod.com/wiki/framework/explosives-framework.html).

PabstMirror: added

PR #5262, opened 2017-06-08
Remove maximum distance of attach to vehicle action

jonpas:
- Title

Max distance of 7 (4.5 in config had no impact) prevents attaching things to large vehicles such as VTOLs, unless you crawl under them. There is no real reason to have a max distance check here, as you can only attach to a vehicle you interacted with after picking the item to attach.


jonpas:
Good point, changed it to 10 meters now, will still end placement system, just 3 meters further away.

PR #5264, opened 2017-06-09
Added parameter to use default damage processing. 

gandjustas:
Fatal damage will kill infantry with new setting. Partialy fixing kill counters.

### DIFF
--- a/addons/medical/stringtable.xml
+++ b/addons/medical/stringtable.xml
@@ -4901,5 +4901,13 @@
             <Japanese>AI は気絶している人へ、ためらってから射撃します</Japanese>
             <Korean>기절할 경우 인공지능이 발사를 지연합니다</Korean>
         </Key>
+        <Key ID="STR_ACE_MedicalSettings_useDefaultDamageProcessing_DisplayName">
+            <English>Use default damage processing.</English>
+            <Russian>Испоользовать обработку урона по-умолчанию.</Russian>
+        </Key>
+        <Key ID="STR_ACE_MedicalSettings_useDefaultDamageProcessing_Description">
+            <English>Don not reduce damage to non-fatal value.</English>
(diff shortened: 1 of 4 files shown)


addons/medical/stringtable.xml
```diff
@@ -4901,5 +4901,13 @@
             <Japanese>AI は気絶している人へ、ためらってから射撃します</Japanese>
             <Korean>기절할 경우 인공지능이 발사를 지연합니다</Korean>
         </Key>
+        <Key ID="STR_ACE_MedicalSettings_useDefaultDamageProcessing_DisplayName">
+            <English>Use default damage processing.</English>
```
Glowbal: The name `default damage processing` is confusing and I am not sure by reading this what `default` means. Judging by the code, it means use Vanilla and the name for the setting should reflect that.

addons/medical/stringtable.xml
```diff
@@ -4901,5 +4901,13 @@
...
+            <Russian>Испоользовать обработку урона по-умолчанию.</Russian>
+        </Key>
+        <Key ID="STR_ACE_MedicalSettings_useDefaultDamageProcessing_Description">
+            <English>Don not reduce damage to non-fatal value.</English>
```
Glowbal: This is misleading, as this is not what is happening.

gandjustas:
- preventInstaDeath work well with my option. It's handled in fnc_handleDamage lines 110-134.
- fatal injury logic works well with my option because my fix only for non-fatal ( by determineIfFatal function)

gandjustas:
Checked `_returnDamage`. It contains correct damage after call to `handleDamage_caching`.

Only issue is to call `setDead` - inside `handleDamage` there is no way to decide will unit die or not. May be set variables inside killed event?

PR #5269, opened 2017-06-11
Fix isFeatureCameraActive for Spectator enhacements

Fusselwurm:
At the moment, isFeatureCameraActive does *not* detect ACE spectator mode, because `ace_spectator_camera` as a variable was removed in 77c2b99ee542e3825c4f25e7b5a399597fd436e2 .
spectator module initializes three separate cameras instead, of which for example `ace_spectator_freeCamera` could be used to check if spec mode is active.

### DIFF
--- a/addons/common/functions/fnc_isFeatureCameraActive.sqf
+++ b/addons/common/functions/fnc_isFeatureCameraActive.sqf
@@ -25,7 +25,7 @@
 
 !(
     isNull curatorCamera && // Curator
-    {isNull (GETMVAR(EGVAR(spectator,camera),objNull))} && // ACE Spectator
+    {isNull (GETMVAR(EGVAR(spectator,freeCamera),objNull))} && // ACE Spectator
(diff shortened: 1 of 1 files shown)


SilentSpike:
This could be changed to check `EGVAR(spectator,isSet)` as it will also support the overhaul once it is merged for 3.11, but otherwise 👍

jonpas:
Let's do what @SilentSpike said, and we can merge. `missionNamespace getVariable [nil, objNull]` will return `objNull` btw? If not, we need to make sure that function still works if spectator is not present (as in removed PBO).

SilentSpike:
@jonpas Isn't that the whole point of the second `getVariable` argument? 😆 

Also note that `isSet` is a boolean, just to clarify. Will be true when spectator is open.

jonpas:
Uh, I could bet I wanted to write something else there after writing the rest... but I forgot now.

PR #5278, opened 2017-06-13
New medical box

BaerMitUmlaut:
- Replace the current medical box with a new model made by @ElTyranos (thanks again!)
- Add open and close lid interaction for it
- Add @ElTyranos to authors


addons/medical/stringtable.xml
```diff
@@ -4901,5 +4901,13 @@
             <Japanese>AI は気絶している人へ、ためらってから射撃します</Japanese>
             <Korean>기절할 경우 인공지능이 발사를 지연합니다</Korean>
         </Key>
+        <Key ID="STR_ACE_Medical_openLid">
+            <English>Open lid</English>
```
jonpas: This might be better in `common` since it's a very general string.

PR #5279, opened 2017-06-14
Refuel - change nozzle position calc

### LINKED ISSUE #5143: UAV Yabhon refueling issue
**ACE3 Version:** `3.9.1` (stable)
 
**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- Cannot connect the fuel nozzle with the Yabhon. 

**Steps to reproduce:**
- Place a fuel truck and a Yabhon UAV with bingo fuel on map. try to refuel the drone.

**Placed Modules:**
- None

**RPT log file:**
- not needed

PabstMirror:
Fix #5143

Removes the old binary search and now uses `lineIntersectsSurfaces`
Default is to aim at player's viewing angle (what you see is what you get) instead of using model center

### DIFF
--- a/addons/refuel/functions/fnc_checkFuel.sqf
+++ b/addons/refuel/functions/fnc_checkFuel.sqf
@@ -21,7 +21,7 @@ params [["_unit", objNull, [objNull]], ["_target", objNull, [objNull]]];
 private _fuel = [_target] call FUNC(getFuel);
 
 [
-    REFUEL_PROGRESS_DURATION * 2.5,
+    TIME_PROGRESSBAR(REFUEL_PROGRESS_DURATION * 2),
     [_unit, _target, _fuel],
     {
         params ["_args"];
--- a/addons/refuel/script_component.hpp
+++ b/addons/refuel/script_component.hpp
@@ -5,6 +5,7 @@
 // #define DEBUG_MODE_FULL
 // #define DISABLE_COMPILE_CACHE
 // #define ENABLE_PERFORMANCE_COUNTERS
+// #define FAST_PROGRESSBARS
 
 #ifdef DEBUG_ENABLED_REFUEL
     #define DEBUG_MODE_FULL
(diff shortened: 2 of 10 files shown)


addons/refuel/functions/fnc_checkFuel.sqf
```diff
@@ -21,7 +21,7 @@ params [["_unit", objNull, [objNull]], ["_target", objNull, [objNull]]];
 private _fuel = [_target] call FUNC(getFuel);
 
 [
-    REFUEL_PROGRESS_DURATION * 2.5,
+    TIME_PROGRESSBAR(REFUEL_PROGRESS_DURATION * 2),
```
Neviothr: Shouldn't `REFUEL_PROGRESS_DURATION` be a `GVAR()`?

addons/refuel/script_component.hpp
```diff
@@ -29,3 +30,9 @@
     _weaponSelect = _unit getVariable QGVAR(selectedWeaponOnRefuel); \
     _unit selectWeapon _weaponSelect; \
     _unit setVariable [QGVAR(selectedWeaponOnRefuel), nil];
+
+#ifdef FAST_PROGRESSBARS
+    #define TIME_PROGRESSBAR(X) ((X) * 0.075)
+#else
+    #define TIME_PROGRESSBAR(X) (X)
+#endif
```
jonpas: We could do that globally, and just set it to 0 or 0.5 (so it's visible that it happens at least) when defined?

addons/refuel/functions/fnc_checkFuel.sqf
jonpas: No, this is not flow rate, it's how long it takes to check current fuel level of a vehicle. Macro is 2s, so this makes it 4s.

addons/refuel/script_component.hpp
PabstMirror: `TIME_PROGRESSBAR` is already in rearm
Rearm, Refuel are the main ones with some long progress bars that make testing tedious.

Might actually cause problems on some things like repacking and medical surgery, not sure.

jonpas: I see. Would be done separate of this PR anyways.

PR #5284, opened 2017-06-16
Added ability to abort module action

chris579:
- Add the ability to abort a module action via ace_zeus_fnc_getModuleDestination
- Action can be aborted by pressing ESC
- Current position is still passed to the code callback

### DIFF
--- a/addons/zeus/functions/fnc_getModuleDestination.sqf
+++ b/addons/zeus/functions/fnc_getModuleDestination.sqf
@@ -5,7 +5,14 @@
  * Arguments:
  * 0: The souce object <OBJECT>
  * 1: Code to run when position is ready <CODE>
- * - Code is passed [0: Successful <BOOL>, 1: Object <OBJECT>, 2: Position ASL <ARRAY>]
+ * - Code is passed [
+                        0: Successful <BOOL>, 
+                        1: Object <OBJECT>, 
+                        2: Position ASL <ARRAY>, 
+                        3: State of Shift <BOOL>, 
+                        4: State of Ctrl <BOOL>,
+                        5: State of Alt <BOOL>
+                    ]
  * 2: Text <STRING><OPTIONAL>
  * 3: Icon image file <STRING><OPTIONAL>
  * 4: Icon color <ARRAY><OPTIONAL>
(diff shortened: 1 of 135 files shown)


bux:
If not happened already don't forget to add yourself to the authors file. :+1:

SilentSpike:
Very nice 😃

chris579:
What about a param for a format text that is shown as a BIS_fnc_curatorHint (like what is happening when placing while pressing shift, ctrl or alt)? Should I add it?

jonpas:
Can you show us how that would look on a screenshot or a short video @chris579 ?

addons/zeus/functions/fnc_getModuleDestination.sqf
```diff
@@ -5,7 +5,14 @@
...
  * 0: The souce object <OBJECT>
  * 1: Code to run when position is ready <CODE>
- * - Code is passed [0: Successful <BOOL>, 1: Object <OBJECT>, 2: Position ASL <ARRAY>]
+ * - Code is passed [
+                        0: Successful <BOOL>, 
+                        1: Object <OBJECT>, 
+                        2: Position ASL <ARRAY>, 
+                        3: State of Shift <BOOL>, 
+                        4: State of Ctrl <BOOL>,
+                        5: State of Alt <BOOL>
```
jonpas: I don't like this indentation, I assume this is stuff passed on from higher scope? I think this is clearer either way:
```
 * - Code is passed <ARRAY>
 *   0: Successful <BOOL>
 *   1: Object <OBJECT>
 *   2: Position ASL <ARRAY>
 *   3: State of Shift <BOOL>
 *   4: State of Ctrl <BOOL>
 *   5: State of Alt <BOOL>
```
@SilentSpike How would you handle that in your doc script?

addons/zeus/functions/fnc_getModuleDestination.sqf
```diff
@@ -5,7 +5,14 @@
...
+                    ]
  * 2: Text <STRING><OPTIONAL>
  * 3: Icon image file <STRING><OPTIONAL>
  * 4: Icon color <ARRAY><OPTIONAL>
```
jonpas: Not changed in this PR, but those 3 liens should be:
```
 * 2: Text <STRING> (default: "")
 * 3: Icon image file <STRING> (default: "\a3\ui_f\data\IGUI\Cfg\Cursors\select_target_ca.paa")
 * 4: Icon color <ARRAY> (default: [1,0,0,1])
```

addons/zeus/functions/fnc_getModuleDestination.sqf
chris579: I just indented this because it got messy with more than 3 params. This are the params that are passed to the code that is run when the user finished placing. I don't know how you handle such things in your docs

PR #5287, opened 2017-06-18
Modify JoinGroup action to show group name

PabstMirror:
Shows group name in action menu.

image

Edit:
I see this as a small usability improvement. 
But I guess it does give some _meta_ information about the specific group a unit is part of??
There is an ace setting `GVAR(EnableTeamManagement)` that could disable the interaction if it's a problem.

### DIFF
--- a/addons/interaction/functions/fnc_modifyJoinGroupAction.sqf
+++ b/addons/interaction/functions/fnc_modifyJoinGroupAction.sqf
@@ -1,0 +1,25 @@
+/*
+ * Author: PabstMirror
+ * Modifies the ACE_JoinGroup action to show group name.
+ *
+ * Arguments:
+ * 0: Target <OBJECT>
+ * 1: Player <OBJECT>
+ * 2: Args <Any>
+ * 3: Action Data <ARRAY>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [cursorObject, player, [], []] call ace_repair_fnc_modifyInteraction;
+ *
+ * Public: No
+ */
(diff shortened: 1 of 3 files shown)


jonpas:
I don't see why anyone would have anything against that. It is information that you'd have no reason to hold back from people on the same side, and people from the opposite side shouldn't get that close to you anyways (if they can even join group from opposite side, not sure how that works).

addons/interaction/functions/fnc_modifyJoinGroupAction.sqf
```diff
@@ -0,0 +1,26 @@
+/*
+ * Author: PabstMirror
+ * Modifies the ACE_JoinGroup action to show group name.
+ *
+ * Arguments:
+ * 0: Target <OBJECT>
+ * 1: Player <OBJECT>
+ * 2: Args <Any>
```
jonpas: `<ANY>`? Not sure what our standard should be for that.

addons/interaction/functions/fnc_modifyJoinGroupAction.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * 3: Action Data <ARRAY>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [cursorObject, player, [], []] call ace_interaction_fnc_modifyJoinGroupAction;
```
jonpas: No semi-colon at the end (I think that's what we decided to do... but no one put it on the wiki yet?).

PR #5294, opened 2017-06-21
Changed hellfire interaction check & added N variant

tbeswick96:
- Change the check in the interaction condition to check for `ace_hellfire_enabled` config value instead of just the raw `ace_hellfire_launcher` weapon.
- This allows other launchers to be made and be compatible (for example, N-variant hellfires)
- Added November (Thermobaric) hellfire ammo, magazines, and weapon. (https://en.wikipedia.org/wiki/AGM-114_Hellfire#Variants) Have used the settings committed for some time with no balance issue. Not as good against tanks, better against infantry and bunkers etc.

### DIFF
--- a/addons/hellfire/CfgAmmo.hpp
+++ b/addons/hellfire/CfgAmmo.hpp
@@ -19,7 +19,7 @@ class CfgAmmo {
         laserLock = 0;
         manualControl = 0;
         maxSpeed = 450;
-        
+
         thrustTime = 2.5; // motor burn 2-3 sec
         thrust = 250;
         timeToLive = 40;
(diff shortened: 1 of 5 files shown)


tbeswick96:
I don't think the `[ACE]` in the naming of magazines etc such as `AGM-114K [ACE]` needs to be there really. Just extra text with no real meaning.

PabstMirror:
Here is another mod's pylon loadout which has their own hellfires, adding the `[ACE]` makes it clear which system you will get.

image

tbeswick96:
@PabstMirror Gotcha makes sense

tbeswick96:
No problem :)

PR #5304, opened 2017-06-22
Add XEH to ace_module

PabstMirror:
Ref https://github.com/CBATeam/CBA_A3/pull/708


jonpas:
Won't it automatically have XEH when `Module_F` will get it in CBA?

PabstMirror:
I don't think so, no inheritance on 
```
class EventHandlers {
```

jonpas:
Ah, good to know!

PR #5305, opened 2017-06-22
Add EH for UAV control, Add Hellfire support

### LINKED ISSUE #5301: Hellfire UI not working with UAVs
**Arma 3 Version:** Latest as of 22 Jun 17
**CBA Version:** Latest as of 22 Jun 17
**ACE3 Version:** Latest as of 22 Jun 17

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
1- Whilst using a UAV, there is no option to select the mode (LOBL/LOAL-H/LOAL-D/LOAL-L) or change the laser code. 
2- If you are sitting in an aircraft with a Hellfire pylon and using the UAV terminal, you can select different modes and codes however this affects the vehicle you sit in rather than the UAV.

**Steps to reproduce:**
1- Fly a UAV with Hellfires attached.
2 - Sit in an aircraft with hellfires attached then fly a UAV with hellfires attached.

- Editor (singleplayer)

**Placed Modules:**
- Zeus (Gamemaster)

**RPT log file:**
- Not present

PabstMirror:
Fix #5301

Needs testing on dedicated, as `PlayerViewChanged` was unreliable in the past.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -320,25 +320,44 @@ GVAR(OldIsCamera) = false;
 // Add event handler for UAV control change
 ACE_controlledUAV = [objnull, objnull, [], ""];
 addMissionEventHandler ["PlayerViewChanged", {
-    params ["", "", "", "", "", "_UAV"];
+    // On non-server client this command is semi-broken
+    // arg index 5 should be the controlled UAV, but it will often be objNull (delay from locality switching?)
+    // On PlayerViewChanged event, start polling for new uav state for a few seconds (should be done within a few frames)
     
-    private _position = (UAVControl _UAV) param [1, ""];
-    private _seatAI = objnull;
-    private _turret = [];
-    switch (toLower _position) do {
-        case ("driver"): {
-            _turret = [-1];
-            _seatAI = driver _UAV;
+    params ["", "", "", "", "_newCameraOn", "_UAV"];
+    TRACE_2("PlayerViewChanged",_newCameraOn,_UAV);
+    
+    [{
+        if (isNull player) exitWith {true};
+        private _UAV = getConnectedUAV player;
+        if (!alive player) then {_UAV = objNull;};
+        private _position = (UAVControl _UAV) param [1, ""];
+        private _seatAI = objnull;
+        private _turret = [];
+        switch (toLower _position) do {
+            case (""): {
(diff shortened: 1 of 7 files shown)


tbeswick96:
Used this earlier on a dedicated server. I had the display just fine, but had no option to change the hellfire mode.

PabstMirror:
@tbeswick96 - Thanks for feedback, I've fixed the cycle guidance keybind for UAVs (Ctrl+Tab)
Interaction menu working directly on UAVs might be possible, but probably won't get to it in this PR

When testing in MP the `PlayerViewChanged` was broken, but I've written something that uses it to start polling for an actual state change, but only runs for a limited time. Could use a second look on it (common/postInit code)

addons/common/XEH_postInit.sqf
```diff
@@ -317,6 +317,30 @@ GVAR(OldIsCamera) = false;
     END_COUNTER(stateChecker);
 }, 0.5, []] call CBA_fnc_addPerFrameHandler;
 
+// Add event handler for UAV control change
+ACE_controlledUAV = [objnull, objnull, [], ""];
```
Neviothr: Camel case in `objNull`.

addons/common/XEH_postInit.sqf
```diff
@@ -317,6 +317,30 @@ GVAR(OldIsCamera) = false;
...
+addMissionEventHandler ["PlayerViewChanged", {
+    params ["", "", "", "", "", "_UAV"];
+    
+    private _position = (UAVControl _UAV) param [1, ""];
+    private _seatAI = objnull;
```
Neviothr: Same.

addons/common/XEH_postInit.sqf
```diff
@@ -320,25 +320,44 @@ GVAR(OldIsCamera) = false;
...
-            _seatAI = gunner _UAV;
+    params ["", "", "", "", "_newCameraOn", "_UAV"];
+    TRACE_2("PlayerViewChanged",_newCameraOn,_UAV);
+    
+    [{
+        if (isNull player) exitWith {true};
+        private _UAV = getConnectedUAV player;
+        if (!alive player) then {_UAV = objNull;};
+        private _position = (UAVControl _UAV) param [1, ""];
+        private _seatAI = objnull;
```
Neviothr: And here.

PR #5306, opened 2017-06-23
Handle ace_cargo_canLoad set to scalar

PabstMirror:
var `ace_cargo_canLoad` used to be 0/1
some addons are still setting to scalar

e.g. r3f-logistics does
```
		_object setVariable ['ace_dragging_canCarry', false];
		_object setVariable ['ace_dragging_canDrag', false];
		_object setVariable ['ace_cargo_canLoad', 0];
```

Throws error
```
  Error ||: Type Number, expected Bool,code
 File z\ace\addons\cargo\functions\fnc_initObject.sqf, line 13
```

### DIFF
--- a/addons/cargo/functions/fnc_initObject.sqf
+++ b/addons/cargo/functions/fnc_initObject.sqf
@@ -21,10 +21,13 @@ TRACE_2("params",_object,_type);
 
 // If object had size given to it via eden/public then override config canLoad setting
 private _canLoadPublic = _object getVariable [QGVAR(canLoad), false];
+if (!(_canLoadPublic isEqualType false)) then {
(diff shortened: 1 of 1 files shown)


addons/cargo/functions/fnc_initObject.sqf
```diff
@@ -21,10 +21,13 @@ TRACE_2("params",_object,_type);
 
 // If object had size given to it via eden/public then override config canLoad setting
 private _canLoadPublic = _object getVariable [QGVAR(canLoad), false];
+if (!(_canLoadPublic isEqualType false)) then {
+    WARNING_4("%1[%2] - Variable %3 is %4 - Should be bool",_object,_type,QGVAR(canLoad),_canLoadPublic);
```
jonpas: `Should be number` you mean?

PR #5308, opened 2017-06-23
Fixed viewdistance error while controlling UAV

wanderrful:
[Fixes the issue mentioned here](https://github.com/acemod/ACE3/issues/3594#issuecomment-219291214)

When you take control of UAV drones, the view distance does not properly update to your setting because it thinks that you are still the land infantry guy.  So I've experimented a bit in the editor and found a condition that will properly return true when you are in control of a drone aircraft.  I've also adapted this for use with land-based UAV stuff (like the Stomper).

To accomplish this, I created a private __isControllingDrone_ variable that will detect if the player is currently in control of any UAV.  I also edited the _landVehicle and _airVehicle scripts so if and only __isControllingDrone_ returns true, then it will check whether the drone is a subclass of "LandVehicle" or an "Air".

That's all I did.  Please accept my pull request!


addons/viewdistance/functions/fnc_adaptViewDistance.sqf
```diff
@@ -22,8 +22,10 @@ if (!GVAR(enabled) || isNull ACE_player) exitWith {};
 
 private _vehicle = vehicle ACE_player;
 
-private _landVehicle = _vehicle isKindOf "LandVehicle" || {_vehicle isKindOf "Ship_F"};
-private _airVehicle = _vehicle isKindOf "Air";
+private _isControllingDrone = !( (UAVControl (getconnectedUAV _vehicle) select 1) isEqualTo "" );
```
PabstMirror: I think this should be `getconnectedUAV ACE_player` in all 3 instances.
If player is riding in a vehicle while controlling a UAV `getconnectedUAV vehicle player` will return `objNull`

PabstMirror:
`fnc_adaptViewDistance` is only called on events (vehicle or setting change)
PR #5305 will add an EH for UAV control, so I think this will have to use that

wanderrful:
I agree that it'd be better if it were ACE_Player.  I only kept it because I thought I'd have a better chance of being accepted if I kept my changes as small as possible.

So does this mean my PR won't be accepted?

PabstMirror:
Overall the PR looks good and fixes a bug
but lets wait for #5305 to be merged so we can add the uav event

PabstMirror:
Updated PR to use new uav event.

jonpas:
Neatly solved! 👌

PR #5318, opened 2017-06-28
Add: Optional parameter to useEffects during setdamage/sethit

Vdauphin:
Hello,
Since 1.67, you can  skip destruction effects if you want (see  : https://community.bistudio.com/wiki/setDamage and https://community.bistudio.com/wiki/setHitPointDamage https://community.bistudio.com/wiki/setHitIndex).
This is usefull in some cases where you don't want to see the vehicle explode and damage objects around (During a database loading for exemple).

This PR add an optional parameter  to ace_repair_fnc_setHitPointDamage to allow this new choice.

I would know if this change is possible ?
Thanks

- Add: Optional parameter `_useEffects`  to ace_repair_fnc_setHitPointDamage
- update the header
- change `params` with an optional param `"_useEffects"`
- `setDamage ` and `setHitIndex` have the new param `_useEffects`
-  https://community.bistudio.com/wiki/setDamage and https://community.bistudio.com/wiki/setHitPointDamage https://community.bistudio.com/wiki/setHitIndex
- Respect the [Development Guidelines](https://ace3mod.com/wiki/development/) --> tell me if anything is wrong

### DIFF
--- a/addons/repair/functions/fnc_setHitPointDamage.sqf
+++ b/addons/repair/functions/fnc_setHitPointDamage.sqf
@@ -7,6 +7,7 @@
  * 0: Local Vehicle to Damage <OBJECT>
  * 1: Selected hitpoint INDEX <NUMBER>
  * 2: Total Damage <NUMBER>
+ * 3: Skip destruction effects <BOOL>
  *
  * Return Value:
  * None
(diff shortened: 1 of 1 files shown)


PabstMirror:
This looks good, do you have a use case for the new param?

jonpas:
It's not a public function, is that param used within ACE3 anywhere?

Vdauphin:
Hello,

> do you have a use case for the new param?

`[vehicle, 1, 0.5, true] call ace_repair_fnc_setHitPointDamage`
Should I change the actual [exemple](https://github.com/Vdauphin/ACE3/blob/f3269b1b496fd1ed1110ad8d5ff6942720cbf3d7/addons/repair/functions/fnc_setHitPointDamage.sqf#L16) with this one ?

> is that param used within ACE3 anywhere?

@jonpas  I don't know. I am using this function to set damage to vehicle saved from a [database](https://github.com/Vdauphin/HeartsAndMinds/blob/master_daily/%3DBTC%3Dco%4030_Hearts_and_Minds.Altis/core/fnc/db/load.sqf#L260).

PR #5320, opened 2017-06-29
Music compatibility  

fishykins:
This is a small change that adds the option to stop music volume being set. If  "ace_setMusicVolume" is false, music volume will not be set. Default value is true, so base functionality is unchanged.


rebelvg:
Maybe you can add it as an ace setting so everyone can set whatever they want instead of what mission maker set.

PR #5327, opened 2017-07-04
Cookoff - Improve ammo detonation sounds

jokoho48:
- Replace the old System that used playSound3D with an approach
that I used within Enhanced SoundScape.
- add new Sounds with distance steps
- Remove Old Sound because they Clipped.
Demo:
https://www.youtube.com/watch?v=WyjK5lZIuD4&feature=youtu.be

the Version in the video not 100% of the same random values because the current values fit better

Sounds by LAxemann


bux:
But the cook off sounds have been redone for last release, haven't they?

jokoho48:
I am currently working to Implement an Object Pooling Solution and LAxemann is not 100% happy with the sounds and may change them a bit.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -108,3 +108,40 @@ if (isServer) then {
...
+        [{
+            params ["_obj", "_sound"];
+
+            private _soundSource = "#particlesource" createVehicleLocal [0,0,0]; // We maybe should switch to a pool of objects that we dont need to create every time a sound plays a new source
+            _soundSource setPos (getPos _obj);
+            _soundSource say3D [_sound, 3000, 0.9 + (random 0.2)];
+            [{
+                params ["_soundObj"];
+                deleteVehicle _soundObj;
+            }, _soundSource, 2] call CBA_fnc_waitAndExecute;
```
commy2: The game already handles the speed of sound without this, no?

jokoho48: say3D don't handle that at all. say3D is a direct command. to use SoS engine side you would need to use SoundSets or SoundShader. what means you need to use Ammo or Weapon Firing. else you dont have any SoS

jonpas:
Pushing back to 3.12.0, would like to see more discussion on this versus old system.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -108,3 +108,59 @@ if (isServer) then {
...
+        _obj; // return
+    }] call CBA_fnc_compileFinal;
+
+    [QGVAR(playCookoffSound), {
+        params ["_obj", "_sound", "_maxDis"];
+        if (isNull _obj) exitWith {};
+        private _distance = _obj distance (positionCameraToWorld [0,0,0]);
+        if (_distance > _maxDis) exitWith {};
+        _distance = _distance + (random (50) - 25);
+        private _dis = switch (true) do {
```
ulteq:
``` 
private _dis = if (_distance < 325) then {
    "close"
} else {
    if (_distance < 952) then {
        "mid"
    } else {
        "far"
    }
};
```
Less code and 33% faster.

ulteq:
The new variant sounds good, but I do not know how the old one was.

addons/cookoff/XEH_postInit.sqf
jonpas: This might be even faster:
```sqf
private _dis = [["far", "mid"] select (_distance < 952), "close"] select (_distance < 325);
```
Can do multi-line for readability.

BTW, you can use `sqf` as highlighter, it works on GitHub.

PabstMirror: `["close", "mid", "far"] select linearConversion [11.5, 1265.5, _distance, 0, 2, true]`
:smile:

jokoho48:
The old.system had no sound modulation on distances so it sounded like the explosion was close even when that was not the case

jokoho48:
And no SOS calculation so a rocket explosion sound was by the player at the moment of creation, of the sound object. The new system now have 3 steps for modulation of the sound and also uses a correct sos timing

jokoho48:
The old sounds also clipped what created  kind of white noice

addons/cookoff/XEH_postInit.sqf
ulteq: Jonas wins. His code is almost twice as fast as mine and a breath faster than PabstMirror's.

jokoho48: Ok i am currently not able to change it feel free to change

SilentSpike:
Pushing this milestone back in favour of getting 3.12.0 out the door, significant changes that require some further review.

Is there any performance impact associated with this? I've always been a little dubious about the whole creating objects for sound addons thing, but never properly looked into it.

jokoho48:
the old system was even more worse with creating Objects.
the old system used PlaySound3D what creates a global Object. i only do that when the player/camera is near the position. also i reuse local Objects so that the create object process is way less then the old system.
for that the system not only respects Speed of Sound but also have distance steps in it what the old system could not even do.

so this system not only creates less Network Trafic but also handles all it with less objects created and better sound. i realy dont get why we pushed it now back for the 3rd time

bux:
Sample: https://youtu.be/8rXpi4NImZQ

bux:
Current Sounds: https://youtu.be/_cJGtrondhw

SilentSpike:
Nice one bux 👍 I find it hard to give feedback on whether the sounds themselves are realistic, but I do prefer the new sounds (although not shown are the new cookoff sounds of shells/rockets you'd find in vehicles).

I definitely like the distance steps!

Cyruz143:
I prefer the new sounds, less jarring on the ear, probably the slightly lower pitch. I assume just audio feedback was wanted on this.

SilentSpike:
For reference, real life: https://www.youtube.com/watch?v=3SlOXowwC4c&t=792

SilentSpike:
A lot of feedback I'm getting (from linking bux's video) is that the sounds could be a bit "sharper" (whatever that really means)

jokoho48:
the old sound where more "Sharper" because they cliped alot over the max Volume and later only create basicly when they got fired in rapid speeds more of a white noice then a good sound that is why the new sounds are more balanced
also i change the sounds a bit and how the volume on distance behave:
https://www.youtube.com/watch?v=kjM8WItJZj0 video of the changes

thojkooi:
we should take the sounds but not the code from this PR.

jokoho48:
> we should take the sounds but not the code from this PR.

I do not think that that is possible because the sounds are made to change at different distances and playSound3D does not support that at all.
Also, playSound3D creates a Globally Synced Object what is more stress on the network than a CBA Event with data attached.
so when we would switch back to the old code base it would create the same jarring sounds.
playSound3D is just not made for Effect sounds that are meant to reach more then 100m

AirShark07:
the new ones sounds way much better is this still WIP?

commy2:
I wanted to implement this using https://github.com/CBATeam/CBA_A3/pull/1174 which is not finished.

Dystopian:
> I wanted to implement this using [CBATeam/CBA_A3#1174](https://github.com/CBATeam/CBA_A3/pull/1174) which is not finished.

@commy2 maybe this PR should be merged and when CBA implementation is ready it can be edited?

commy2:
But I don't like this implementation.

jonpas:
Anyone wants to do anything with this?

johnb432:
> Anyone wants to do anything with this?

I think it would be neat. I need to take a better look at it once I have cleared my backlog.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -58,3 +58,50 @@
...
+            private _sDistance = _distance + ((random 5) - 2.5);
+            private _disStep = [[["far", 1.5], ["mid", 1.3]] select (_sDistance < 952), ["close", 1]] select (_sDistance < 235);
+            _sound = format [QGVAR(%1_%2_%3), _sound, _disStep select 0, ceil (random 3)];
+            hintSilent _sound;
+            // fix for Vanilla Say3d Falloff
+            private _value = 1 - (_distance / _maxDist);
+            private _fakeValue = linearConversion ([[0, 0.7, _value, 1, 0.8], [0.7, 1, _value, 0.8, 0]] select (_value >= 0.7));
+            private _soundSource = call FUNC(poolGetObject);
+            _soundSource setPos (getPos _obj);
+            _soundSource say3D [_sound, (_distance / _fakeValue) * (_disStep select 1), 0.9 + (random 0.2)];
```
johnb432, line 95: Why not use [playSound3D](https://community.bistudio.com/wiki/playSound3D)? In my testing (in single player) it obeys the speed of sound, as in it takes time for the sound to propagate. According to the wiki page it also takes the Doppler effect into account, but I did not verify that.

jokoho48: Doppler Effect != Speed of Sound and https://www.youtube.com/watch?v=fhsKUGzLHog
and https://github.com/acemod/ACE3/issues/9547 does prove that playSound3D does not use Speed Of Sound

johnb432: > Doppler Effect != Speed of Sound
I know that, I just thought it's something nice to have if possible.
I'll test `playSound3D` in MP on my end to see what it gives.

johnb432: Well.. both the bug report and my findings are correct.

`playSound3D` does obey speed of sound on the machine where `playSound3D` is called from. For others it's immediate.

However, if the following is executed, it plays fine:

```sqf
[["z\ace\addons\cookoff\sounds\heavy_crack_close.wss", objNull, false, getPosASL player, 5, 1, 1000, 0, true]] remoteExecCall ["playSound3D", 0]
```

We can make an event that calls it locally, would be the easiest solution imo.

jokoho48: Yeah, that is definitely new, as this PR is over 6 years old. playSound3D didn't have SoS, nor did it have a local tag.

Additionally, we still can't blindly play the sound effects but still need to do a local measuring of the distance and choose the correct sound to play.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -58,3 +58,50 @@
...
+    }] call CBA_fnc_addEventHandler;
+
+    [{
+        if (GVAR(soundObjPool) isEqualTo []) exitWith {};
+        private _element = GVAR(soundObjPool) param [0, [time, objNull]];
+        if (((_element select 0) + 120) < time) then {
+            GVAR(soundObjPool) deleteAt 0;
+            deleteVehicle (_element select 1);
+        };
+    }, 0.5] call CBA_fnc_addPerFrameHandler;
```
johnb432, line 106: I don't know much about the calculations behind sound waves, but my suggestion is as follows:

```suggestion
    // Plays a sound locally, so that different sounds can be used for various distances
    [QGVAR(playCookoffSound), {
        params ["_object", "_sound", "_maxDistance"];

        if (isNull _object) exitWith {};

        private _distance = _object distance (positionCameraToWorld [0, 0, 0]);

        if (_distance > _maxDistance) exitWith {};

        TRACE_3("",_object,_sound,_maxDistance);

        // 3 classes of distances: close, mid and far, each having different sound files
        _distance = ([["far", "mid"] select (_distance < 952), "close"] select (_distance < 235));
        _sound = format ["sounds\%1\%2_%3.wss", _sound, _distance, floor (random 3) + 1];

        TRACE_1("",_sound);

        _sound = FORMAT_1(QPATHTOF(%1),_sound);

        // Make sure file exists, so RPT isn't spammed with non-existent entry errors
        if (!fileExists _sound) exitWith {};

        // Obeys speed of sound and, according to wiki, takes doppler effects into account
        playSound3D [_sound, objNull, insideBuilding _object >= 0.5, getPosASL _object, 1, 0.9 + (random 0.2), _maxDistance, 0, true];
    }] call CBA_fnc_addEventHandler;
```

It's a very basic implementation which I detailed in https://github.com/acemod/ACE3/issues/9547#issuecomment-1799740572.

I dropped the fall-off calculations made in the PR, as `playSound3D` already accounts for that. I also removed the volume calculations, as I did not know how to re-implement them into my suggestion.

I don't really have the knowledge and setup to judge if the quality is acceptable, so if anyone knowledgeable could find calculations for the volume (and fall-off if necessary) that would be much appreciated.

jokoho48: This does not cover the shotmissile and shotsubmunitions. those where done via a https://github.com/acemod/ACE3/pull/5327/files#diff-73a0f49fbad6f9b061d350aa515eef040577f4c7b959bda32d5c654c643651d3R24-R42

johnb432: I have updated my suggestion to include `shotmissile` and `shotsubmunitions`.

johnb432:
I went ahead and pushed my suggestions after testing. The sounds might not be loud enough in its current implementation, but that can be easily tweaked via `playSound3D`.
Further feedback is needed for the above, but for me this is good to go.

addons/cookoff/CfgSounds.hpp
```diff
@@ -0,0 +1,59 @@
...
+#define SHOTSOUNDCLASSTYPE(type,maxDistance) \
+SHOTSOUNDCLASS(type,close,maxDistance);\
+SHOTSOUNDCLASS(type,mid,maxDistance);\
+SHOTSOUNDCLASS(type,far,maxDistance)
+
+class CfgSounds {
+    SHOTSOUNDCLASSTYPE(shotbullet,1250);
+    SHOTSOUNDCLASSTYPE(shotrocket,1600);
+    SHOTSOUNDCLASSTYPE(shotshell,1300);
+
```
LinkIsGrim, line 21: Add comment?

addons/cookoff/XEH_postInit.sqf
```diff
@@ -58,3 +58,37 @@
...
+        params ["_object", "_sound"];
+
+        if (isNull _object) exitWith {};
+
+        private _distance = _object distance (positionCameraToWorld [0, 0, 0]);
+
+        TRACE_3("",_object,_sound,_maxDistance);
+
+        // 3 classes of distances: close, mid and far, each having different sound files
+        private _classDistance = ([["far", "mid"] select (_distance < 952), "close"] select (_distance < 235));
```
LinkIsGrim, line 74: double select might be faster but this just looks ugly.

also replace the magic numbers with defines?

addons/cookoff/CfgSounds.hpp
johnb432: Not sure if the one I added suffices.

johnb432:
In the latest commit I bumped the volume from 1 to 2.
Feedback on the volume in general is needed.

addons/cookoff/CfgSounds.hpp
LinkIsGrim: They do

LinkIsGrim (approved):

PR #5335, opened 2017-07-06
Fix the issue that SFX variants are not global 

jokoho48:
what means that different players can get different pressure Cookoff sounds

to prevent that I split up the sounds in 3 types and used a Weighted select to have the before used values back

### DIFF
--- a/addons/cookoff/CfgSFX.hpp
+++ b/addons/cookoff/CfgSFX.hpp
@@ -1,12 +1,26 @@
 
 class CfgSFX {
-    class GVAR(CookOff) {
-        name = QGVAR(cookoff);
+    class GVAR(CookOff_low) {
+        name = QGVAR(cookoff_low);
         // Index 4 is percentage chance to play, in theory high pressure is way more likely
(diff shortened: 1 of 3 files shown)


addons/cookoff/CfgSFX.hpp
```diff
@@ -1,12 +1,26 @@
...
+        titles[] = {};
+        empty[] = {"",0,0,0,0,0,0,0};
+    };
+    class GVAR(CookOff_mid) {
+        name = QGVAR(cookoff_mid);
+        // Index 4 is percentage chance to play, in theory high pressure is way more likely
+        sound[] = {QPATHTOF(sounds\cookoff_mid_pressure.ogg),6,1,400,1,0,0,0};
+        sounds[] = {"sound"};
+        titles[] = {};
+        empty[] = {"",0,0,0,0,0,0,0};
```
SilentSpike: Probably make sense to have these classes inherit and just change the `name` and `sound[]` properties

jokoho48: I thought I'd done that. wired.

jonpas: What wires? 😆

PR #5336, opened 2017-07-06
Fix adv. throw key handling when reloading

Dystopian:
I use one key for vanilla and ACE throwing (if ACE is loaded adv. one is used, if not - vanilla one).
If current weapon reloading time is not correctly calculated, you can suddenly throw vanilla way instead of ACE. This can be easily reproduced with `MX_3GL` or `KH2002 Sama 3GL` UGL reloading because UGL reloading time is less than calculated one.
This PR adds UGL reloading time considering to `ace_common_isReloading` var calculating.
Also this PR adds blocking of adv. throwing key when reloading (if once another reloading time is not correctly calculated).

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -374,11 +374,8 @@ GVAR(isReloading) = false;
 
         if (_weapon != "") then {
             private _muzzle = currentMuzzle ACE_player;
-            private _gesture = if (_weapon isEqualTo _muzzle) then {
-                getText (configfile >> "CfgWeapons" >> _weapon >> "reloadAction");
-            } else {
-                getText (configfile >> "CfgWeapons" >> _weapon >> _muzzle >> "reloadAction");
-            };
+            private _wpnConfig = configFile >> "CfgWeapons" >> _weapon;
+            private _gesture = getText ([_wpnConfig >> _muzzle, _wpnConfig] select (_weapon isEqualTo _muzzle) >> "reloadAction");
             if (_gesture == "") exitWith {}; //Ignore weapons with no reload gesture (binoculars)
             private _isLauncher = _weapon isKindOf ["Launcher", configFile >> "CfgWeapons"];
             private _config = ["CfgGesturesMale", "CfgMovesMaleSdr"] select _isLauncher;
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -374,11 +374,10 @@ GVAR(isReloading) = false;
 
         if (_weapon != "") then {
             private _muzzle = currentMuzzle ACE_player;
-            private ["_gesture"];
-            if(_weapon isEqualTo _muzzle) then {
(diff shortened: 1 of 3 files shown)


addons/common/XEH_postInit.sqf
```diff
@@ -373,7 +373,13 @@ GVAR(isReloading) = false;
         private _weapon = currentWeapon ACE_player;
 
         if (_weapon != "") then {
-            private _gesture  = getText (configfile >> "CfgWeapons" >> _weapon >> "reloadAction");
+            private _muzzle = currentMuzzle ACE_player;
+            private ["_gesture"];
+            if(_weapon isEqualTo _muzzle) then {
```
jonpas: Space after `if`.

654wak654:
Too big for review, too small for PR, so I'm just going to mention it here:

Instead of lines 377-382:
```sqf
private _gesture = if (_weapon isEqualTo _muzzle) then {
    getText (configFile >> "CfgWeapons" >> _weapon >> "reloadAction");
} else {
    getText (configFile >> "CfgWeapons" >> _weapon >> _muzzle >> "reloadAction");
};
```

jonpas:
> your code will throw an undefined variable

Either the `then` or `else` statement will execute, they both do `getText`, which returns `""` by default. I don't like that `private` much either though.

654wak654:
They will be executed but in their own scope, won't carry over.

Dystopian:
> They will be executed but in their own scope, won't carry over

No, I defined `private ["_var"]` before `if` for that.
Anyway I fixed like you say though I don't really like this syntax.

jonpas:
With this, I am wondering if we should change the default keybind to whatever the default keybind is for vanilla throwing. Would cause less confusion for new players and also prevent them from throwing grenades by mistake, because they would instead be greeted by adv. throwing.

@commy2 That wouldn't cause a reset for people who already changed it, correct?

SilentSpike:
Regarding the use of private, something like this would also work:
```
private _wpnConfig = configFile >> "CfgWeapons" >> _weapon;
private _gesture = getText ([_wpnConfig >> _muzzle, _wpnConfig] select (_weapon isEqualTo _muzzle) >> "reloadAction");
```

As for default key, I wouldn't be against making it the same as vanilla by default. It's not often you'd want to use a throwable instantly (you can still do it pretty damn quickly with advanced and more accurately too).

jonpas:
There is still one problem with having the same key, eg. if you have "G" on both (vanilla bind) and press "Shift+G" it will still throw the vanilla way, due to the modifier not getting intercepted.

jonpas:
I'd leave "Shift+G" as default for now, as that way anyone using same keybind for it is more likely to be aware of modifiers not getting intercepted, and it retains the possibility of using vanilla throwing alongside Advanced Throwing for whatever reason.

jonpas:
On another though, that's the least safe way of throwing (having both available). Maybe we should just do this + completely intercept vanilla key (we can read that and block it IIRC) for 3.11.

PR #5338, opened 2017-07-07
fix the delay timing when a unit switch to Captive

jokoho48:
- fix a timing issue that the setCaptive delay can be incorrect on a rapid switching of unconscious or long Captive Delays


jonpas:
@BaerMitUmlaut @Glowbal ?

PR #5351, opened 2017-07-10
Improve cook off settings UX

SilentSpike:
- Add second eden attribute to boxes/vehicles to toggle the ammunition cook off independently
- Remove the general cook off toggle from boxes (only ammunition cook off is relevant)
- Remove the `GVAR(enableAmmoCookoff)` setting condition from boxes as they already have their own setting `GVAR(enableAmmobox)`
- Make the eden default values respect the ace_setting value

The aim here is to change the way boxes respect cook off settings and make the UX cleaner.

Old behaviour:
Boxes are sharing both possible attributes vehicles use to toggle their individual cook off components (`GVAR(enableAmmoCookoff)` and `GVAR(enable)`) as well as the global setting for `GVAR(enableAmmoCookoff)` and their own setting `GVAR(enableAmmobox)`.

New behaviour:
Boxes only care about the global box setting `GVAR(enableAmmobox)` and the attribute `GVAR(enableAmmoCookoff)`.

We probably want to do this alongside the change to CBA settings for 3.11.0.

### DIFF
--- a/addons/cookoff/CfgEden.hpp
+++ b/addons/cookoff/CfgEden.hpp
@@ -11,6 +11,16 @@ class Cfg3DEN {
                         tooltip = CSTRING(enable_tooltip);
                         expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
                         typeName = "BOOL";
+                        condition = "objectVehicle";
+                        defaultValue = "(true)"; // fix pbo project preprocessing bug
+                    };
+                    class GVAR(enableAmmoCookoff) {
+                        property = QGVAR(enableAmmoCookoff);
+                        control = "Checkbox";
+                        displayName = CSTRING(enableAmmoCookoff_name);
+                        tooltip = CSTRING(enableAmmoCookoff_tooltip);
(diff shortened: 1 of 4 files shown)


SilentSpike:
Should probably be looked at now that CBA setting conversion is merged. Not sure who's best to review. @commy2 ?

jonpas:
We probably want to do it with 3.13, convert the rest of the settings in that component to CBA Settings in-code (by hand, currently it's converted on the fly), as that's when backwards compatibility will be lost.

SilentSpike:
That's why this should be merged before they are converted by hand, so that this PR doesn't need to be converted separately to retain the intended behaviour.

PabstMirror:
https://github.com/acemod/ACE3/blob/master/addons/grenades/functions/fnc_incendiary.sqf#L172
```
            if ("ace_cookoff" call EFUNC(common,isModLoaded) && {EGVAR(cookoff,enable)}) then {
                _x call EFUNC(cookoff,cookOffBox);
```

Should this be changed to enableAmmobox ?

SilentSpike:
Good catch, you're right it should actually match the condition for box cookoff: 
```sqf
(_this select 0) getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmobox)]
```

PR #5354, opened 2017-07-10
Added rating settings

chris579:
- Add a module + settings to control whether the rating system should be enabled for players or ai
- Mainly purpose is to prevent friendly ai shooting at players that crash helicopters or destroy friendly equipment


addons/rating/functions/fnc_initManLocal.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+// Don't add if unit is not local
+if(!local _unit) exitWith {};
+
+// Check settings
+if((isPlayer _unit && GVAR(allowNegativeRatingPlayers)) || (!(isPlayer _unit) && GVAR(allowNegativeRatingAi))) exitWith {};
+
+TRACE_1("adding rating eh",_unit);
+
+_unit addEventHandler ["HandleRating", {0}];
```
Cuel: How about this to still keep engine behaviour but not allow it to go negative? No idea if there's any use of this old OFP remnant
```
_unit addEventHandler ["HandleRating", {
    if ((_this select 1) < 0) exitWith {0};
}];
```

SilentSpike:
Not sure this warrants it's own entire addon, might fit better into ace_ai or something

654wak654:
>Not sure this warrants it's own entire addon, might fit better into ace_ai or something

Maybe optionals/ or acex?

chris579:
> Not sure this warrants it's own entire addon, might fit better into ace_ai or something
I started building this whole thing into ace_ai at the beginning but this ended up misleading as you would player specific behaviour in an ai component.

I was quite unsure where to put it so I made a new component.

addons/rating/functions/fnc_initManLocal.sqf
chris579: I don't think there is any purpose for this except ai behaviour for friendly fire.

This code would be better:

`_unit addEventHandler ["HandleRating", {
    (_this select 1) max 0
}];`

addons/rating/CfgEventHandlers.hpp
```diff
@@ -0,0 +1,25 @@
+class Extended_Init_EventHandlers {
+    class CAManBase {
+        class ADDON {
+            serverInit = QUOTE(_this call FUNC(initMan));
```
AACO: Can this be added only once the ace settings are initialized on the server and either GVAR bool is false via https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_addClassEventHandler.sqf (apply retroactively set to true)?

chris579: `initMan` checks if the settings init is finished and adds all pending requests to the settings finished execution queue. Don't really see a benefit adding this on the script side.

AACO: It may be a personal preference thing, but I don't think any config based XEHs should be defined if there is a case where it should never run (especially when the default case is that it shouldn't run).

My proposed solution:
On the server postInit you wait until the settings are finished initializing, only where there is a use case (one or both GVARs are false) then you install the class based event handler with the retroactive flag.

The current solution:
Even if there's no use case for it (both GVARs are true) the event handler gets installed, and will run (and exit) for every unit spawned.

Just seems needlessly wasteful.

addons/rating/functions/fnc_initManLocal.sqf
```diff
@@ -0,0 +1,30 @@
...
+#include "script_component.hpp"
+
+params["_unit"];
+TRACE_1("params",_unit);
+
+// Don't add if unit is not local
+if(!local _unit) exitWith {};
+
+// Check settings
+if((isPlayer _unit && GVAR(allowNegativeRatingPlayers)) || (!(isPlayer _unit) && GVAR(allowNegativeRatingAi))) exitWith {};
```
AACO: Could this have race issues? I.E. the server's settings are initialized, but the clients are not (especially for JIP cases)?

addons/rating/functions/fnc_initManLocal.sqf
AACO: This setup seems very `local` sensitive. What happens if a unit changes locality (not likely for clients, but could be a very real issue for AI between headless client setups like werthless and RCing units changing group locality)?

jonpas:
What's the status on this now @chris579 ? I think we should go with a lot more minimalistic approach if you still want to work on it, as in just reset rating on whatever event.

chris579:
I didn't worked on it any more since it isn't clear where to put it and how to handle it. Personally I would remove the AI option since it's quite useless and only focus on players. 

If you say which addon this belongs to I will move it there. A new addon just for rating is truly pointless.

commy2:
Either put this into interaction, or put this into another fitting addon and also move "Pardon" together with this. Disable "Pardon" if this is enabled. my 2ct

addons/interaction/functions/fnc_canPardon.sqf
```diff
@@ -21,3 +21,4 @@ params ["_unit", "_target"];
 alive _target
 && {rating _target < -2000}
 && {side group _unit == side group _target}
+&& !{GVAR(disableNegativeRating)}
```
jonpas: I think this is not necessary since L22 will never be `true` anyways due to minimum rating being `0` when this setting is active.

addons/interaction/stringtable.xml
```diff
@@ -927,6 +927,14 @@
             <Chinesesimp>允许玩家使用小队管理选单? 预设: 是</Chinesesimp>
             <Chinese>允許玩家使用小隊管理選單? 預設: 是</Chinese>
         </Key>
+        <Key ID="STR_ACE_Interaction_DisableNegativeRating_DisplayName">
+            <English>Disable negative rating</English>
+            <German>Negative Bewertung deaktivieren</German>
+        </Key>
+        <Key ID="STR_ACE_Interaction_DisableNegativeRating_Description">
+            <English>Should players receive negative rating?</English>
```
jonpas: This could do with a longer explanation, as in what it can cause if not disabled and how it changes it.

addons/interaction/functions/fnc_canPardon.sqf
chris579: Removed it

addons/interaction/stringtable.xml
chris579: Expanded it a bit more

addons/interaction/CfgVehicles.hpp
```diff
@@ -7,7 +7,7 @@ class CfgVehicles {
         author = ECSTRING(common,ACETeam);
         category = "ACE";
         displayName = CSTRING(Module_DisplayName);
-        function = "ACE_Interaction_fnc_moduleInteraction";
+        function = FUNC(moduleInteraction);
```
SilentSpike: `QFUNC`

addons/interaction/XEH_postInit.sqf
```diff
@@ -108,3 +108,11 @@ GVAR(isOpeningDoor) = false;
 
 ["isNotSwimming", {!underwater (_this select 0)}] call EFUNC(common,addCanInteractWithCondition);
 ["isNotOnLadder", {getNumber (configFile >> "CfgMovesMaleSdr" >> "States" >> animationState (_this select 0) >> "ACE_isLadder") != 1}] call EFUNC(common,addCanInteractWithCondition);
+
+["ace_settingsInitialized", {
+    if (GVAR(disableNegativeRating)) then {
+        ACE_player addEventHandler ["HandleRating", {
```
SilentSpike: I think either this should be `player` (as you'd want it to apply to the actual player and not something they're controlling) or you should be using CBA's `"unit"` [player event](https://github.com/CBATeam/CBA_A3/wiki/Player-Events) to add the same rating handling whenever the player character changes.

jonpas: Are you sure? I am not sure how rating works with remote control, doesn't it get applied to the remote controlled unit? I guess we still wouldn't want to impact that if you are Zeusing or something though.

chris579: No, @SilentSpike is right. The handler gets assigned to the unit that is `ACE_player` at the time the settings get initialized. Therefore it's not safe to use `ACE_player` because there is a slight risk that this EH could get assigned to a remote unit.
Btw I also changed this to `player` already

Issue #2894, opened 2015-11-29
Cannot Disable Medical System


jameslkingsley:
This is seriously not right. There needs to be an option in the mission to disable the medical system completely. Hard coding the system to be enabled it just not right and is poor coding practice.

jokoho48:
delete the ace_medical.pbo and its disabled ;)

jameslkingsley:
Did you not see the "option in the mission" part of what I wrote?

jonpas:
-> #414 

> Hard coding the system to be enabled it just not right and is poor coding practice.

You are always free to write your own or even pull request to ACE3. :)

jameslkingsley:
Or it should just be in the base ACE 3 mod... It's not like I'm asking for much...

SilentSpike:
It's not a coding practise, we can't add an option because the medical system requires config changes.

sargken:
Why would you want to remove the medical system regardless? Its a mainstay of the ACE mod.

jameslkingsley:
Because I don't want to use it in a mission...

jameslkingsley:
If there was API for handling a unit's damage then this wouldn't be an issue, but at the moment any BIS damage commands do nothing with ACE.

licht-im-Norden87:
I know this request is a little bit old.. But i must say: jameslkingsley , you behave like a jerk

commy2:
ref #414

PR #5358, opened 2017-07-11
Remove the unsupported option of disabling medical level

Wakbub:
Setting `Medical Level` to `0` (`Disabled`) in config entries (`serverconfig.hpp`) or mission parameters (`description.ext`) will unexpectedly default it to `1` (`Basic`), despite `Medical Level` under `ACE Options -> Export Menu` having a `Disabled` option. This is because only values `1` (`Basic`) and `2` (`Advanced`) are supported.

- Improve expectancy: In `ACE Options -> Export Menu` the `Medical Level` can now only be set to the supported options `Basic` or `Advanced` ([CfgVehicles.hpp](https://github.com/acemod/ACE3/blob/master/addons/medical/CfgVehicles.hpp#L31)).
- Improve consistency: In the `Medical Settings [ACE]` module the `Medical Level` can only be set to `Basic` or `Advanced` ([fnc_moduleMedicalSettings.sqf](https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_moduleMedicalSettings.sqf#L25)), as supported settings are read from [CfgVehicles.hpp](https://github.com/acemod/ACE3/blob/master/addons/medical/CfgVehicles.hpp).
- Fulfill design decisions: An option to disable `Medical Level` was previously considered invalid (https://github.com/acemod/ACE3/issues/2894).
- Not conflict with anyone's feature request (https://github.com/acemod/ACE3/issues/414) to be able to set `Medical Level` to `Disabled` (despite it not being supported).
- Not conflict with anyone's ACE settings or future implementation of being able to disable `Medical Level`. The alternative solution of changing the values of `normal` (`Basic`) and `full` (`Advanced`) from `1` and `2` to `0` and `1` in [CfgVehicles.hpp](https://github.com/acemod/ACE3/blob/master/addons/medical/CfgVehicles.hpp) as well, instead of only removing `Disabled`, would however cause conflict.

### DIFF
--- a/addons/medical/ACE_Settings.hpp
+++ b/addons/medical/ACE_Settings.hpp
@@ -5,7 +5,7 @@ class ACE_Settings {
         description = CSTRING(MedicalSettings_level_Description);
         value = 1;
         typeName = "SCALAR";
-        values[] = {"Disabled", "Basic", "Advanced"};
+        values[] = {"Basic", "Advanced"};
     };
     class GVAR(medicSetting) {
         category = CSTRING(Category_Medical);


PabstMirror:
This will make 
basic = 0
advanced = 1
which will break all the code that expects `ace_medical_level` to be either 1 or 2

Wakbub:
Even if `typeName` is `SCALAR`?

PabstMirror:
If there is a `values[]` entry, it will only allow selecting values that are indexes from the list
` values[] = {"Disabled", "Basic", "Advanced"};`
it's zero indexed, so 0 = disabled, 1 = basic, 2 = advanced

Wakbub:
Just realized ACE does not have a texts[] to values[] mapping. With [Arma 3's mission parameters](https://community.bistudio.com/wiki/Arma_3_Mission_Parameters), the following would have worked without breaking anything.

```
values[] = {1, 2};
texts[] = {"Basic","Advanced"};
```

PabstMirror:
Hopefully we move to CBA settings sooner than later,
which does support a name/value setup for list entries

dedmen:
Reopen as issue and fix now that we have CBA settings?

PR #5363, opened 2017-07-12
Update the Wiki for the Settings Framework

Wakbub:

- Delete the invalid shortcut for Modules (F7) and add the shortcut for Systems (F5).
- Add that you need to place a playable character (to be able to interact with the `ACE Options` dialog, otherwise the game will become unresponsive when opening it).
- Add that the scenario must be played in single-player (for the `Open Export Menu` button to be displayed).
- Change the non-existent `Config Export` button to `Open Export Menu`.
- Change the wording in other parts for clarity.

### DIFF
--- a/docs/wiki/framework/settings-framework.md
+++ b/docs/wiki/framework/settings-framework.md
@@ -71,15 +71,15 @@ The server config setting entries are done through our optional `ace_server.pbo`
 
 ACE3 contains a lot of settings, for that reason tweaking everything to your liking manually can be quite a task. We have provided the option to export all settings in the editor (single-player). For this, follow these simple steps:
 
-- Open the `Eden Editor`.
-- Place a playable character.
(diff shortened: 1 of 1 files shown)


Laid3acK:
Thank you, i think that it's much comprehensible for all non English speaking people.
The server.pbo is an important part of this, maybe a sort of "Don't forget the server.pbo in your ACE3/Addons folder and you're done" ?

Wakbub:
The “and you’re done” part implies done exporting the settings (the heading is “3.1 Exporting the settings”). How to load the server config is described later (in “3.2 Loading up the server config”).

In my opinion “and you’re done” should be omitted as reaching the end of the list implies completion. Further, the bulleted list should be replaced with a numbered list as the steps are in chronological order. The pull request is updated to reflect these opinions.

jonpas:
Agreed with all of the above comment. 👍

PR #5364, opened 2017-07-12
Maverick L (laser guided AGM)

xrufix:
- Add AGM-65 Maverick L type laser guided missiles.
-- Utilizes ACE Advanced Missile Guidance.
-- Uses the HUD which is also used in ace_hellfire.
-- Pylon magazines included - available to all racks on which vanilla Macers can be used.
-- DIR attack profile.

- Rename IR-guided Maverick to "AGM-65 Maverick G".
- Add names of pylon magazines of Macer/Maverick to ace_realisticnames. 

My aim was to provide a laser guided AGM as an addition to vanilla's IR-guided missiles.
The Maverick L is such a weapon, being a somewhat modernized version of the Maverick E/E2 used by US Navy since 2012.


addons/maverick/functions/fnc_attackProfile.sqf
```diff
@@ -0,0 +1,58 @@
...
+ * [[1,2,3], [], []] call ace_maverick_fnc_attackProfile;
+ *
+ * Public: No
+ */
+#define DEBUG_MODE_FULL
+#include "script_component.hpp"
+
+params ["_seekerTargetPos", "_args"];
+_args params ["_firedEH"];
+_firedEH params ["_shooter","","","","","","_projectile"];
```
Neviothr: Space after the commas.

BaerMitUmlaut:
It's there anything special about the attack profile? Because there is already a linear one in missile guidance.

xrufix:
The attack profile is a 90% copy of the linear profile. I only removed the upwards arc at the start to prevent missiles climbing after being fired in low altitude flight.

I just checked the differences again, though, and found it will only make a difference in altitudes lower than 5m AGL, so I might as well remove it and use AMGs LIN profile instead.

I have yet to research if Mavericks have a top-attack-mode or similar alternative attack profiles in real-life. Could not find any information about this yet.

PabstMirror:
Looks pretty good to me,

To avoid code duplication I'm thinking we should make the showHud code more generic and just move it to the laser component so that it could work for any weapon system.
There will be some changes to it in #5305 and #5365, so don't worry about doing anything yet.

tecak:
Please make a attackProfile ace_missileguidance for gbu bombs

BaerMitUmlaut:
@tecak This is not the place to do feature requests.

xrufix:
With #5305 and #5365 merged, is there any news on moving the showHud code?

PabstMirror:
Sorry for the delay xrufix, I switched this to use the showHud that's now in ace_laser

PabstMirror:
Self designation is a little tricky due to plane's speed and view distance limits in arma
but is possible (tested in the a-10, using the targeting pod)

jonpas:
> Self designation is a little tricky due to plane's speed and view distance limits in arma

FoV view distance can help a lot I found out, so when you zoom in with the targeting pod view distance increases automatically.

PR #5366, opened 2017-07-14
Switch Units - Remove use of BIS_fnc_addStackedEventHandler

PabstMirror:
BIS_fnc_addStackedEventHandler throws error on dev branch
they are now compiling the passed args.
also `_pos, _shift, _alt` are no longer present

### DIFF
--- a/addons/switchunits/functions/fnc_addMapFunction.sqf
+++ b/addons/switchunits/functions/fnc_addMapFunction.sqf
@@ -3,25 +3,21 @@
  * Adds a mapClick Eventhandler
  *
  * Arguments:
- * 0: unit <OBJECT>
- * 1: sides <ARRAY<OBJECT>>
+ * None
  *
  * Return Value:
  * None
  *
  * Example:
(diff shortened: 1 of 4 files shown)


commy2:
MapSingleClick mission eh also does not support overwriting the input by returning true, which onMapSingleClick stacked/command does. I don't think this is used here though.

jonpas:
From dev-branch update 17-07-2017:
> Fixed: The BIS_fnc_addStackedEventHandler function did not accept object argument or function name 
> Fixed: The BIS_fnc_addStackedEventHandler function was missing some native parameters, such as _pos in onMapSingleClick

dedmen:
going through `addMissionEventhandler` is more performant because it doesn't have to go through the Logic of BIS Stacked. But.. https://feedback.bistudio.com/T123355

commy2:
That doesn't matter at all for MapSingleClick tbqh.

PabstMirror:
Guess this isn't needed anymore

PR #5367, opened 2017-07-15
Add hellfire support to CUP vehicles

PabstMirror:


bux:
Shouldn't this be in CUP's compat?

PR #5374, opened 2017-07-19
Fix missing string with stringtable

GodofMonkeys:
Fix Advanced Fatigue & QuickMount missing, add string into stringtable, and update Chinese localization with this two pbo.

### DIFF
--- a/addons/quickmount/ACE_Settings.hpp
+++ b/addons/quickmount/ACE_Settings.hpp
@@ -36,6 +36,6 @@ class ACE_Settings {
         description = CSTRING(PriorityDescription);
         isClientSettable = 1;
         force = 0;
-        values[] = {"Driver", "Gunner", "Commander", "Passenger"};
+        values[] = {CSTRING(PriorityDriver), CSTRING(PriorityGunner), CSTRING(PriorityCommander), CSTRING(PriorityPassenger)};
     };
 };
--- a/addons/quickmount/ACE_Settings.hpp
+++ b/addons/quickmount/ACE_Settings.hpp
@@ -36,6 +36,6 @@ class ACE_Settings {
         description = CSTRING(PriorityDescription);
         isClientSettable = 1;
         force = 0;
-        values[] = {CSTRING(PriorityDriver), CSTRING(PriorityGunner), CSTRING(PriorityCommander), CSTRING(PriorityPassenger)};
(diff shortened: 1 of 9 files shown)


addons/quickmount/ACE_Settings.hpp
```diff
@@ -36,6 +36,6 @@ class ACE_Settings {
         description = CSTRING(PriorityDescription);
         isClientSettable = 1;
         force = 0;
-        values[] = {"Driver", "Gunner", "Commander", "Passenger"};
```
jonpas: There must be vanilla translations for those already.

GodofMonkeys: It is vanilla translations? But I can't see it translation in my game.

Image
alt text

GodofMonkeys: Oh, I see. You mean I should use vanilla game's string to replace these string?

GodofMonkeys:
Add more, fix Gestures & DAGR missing string

addons/quickmount/ACE_Settings.hpp
jonpas: Yeah, but I only found 2:
- Driver: `str_a3_cfgvehicles_c_driver_1_random_base_f0`
- Commander: `str_a3_cfgvehicles_highcommand_moduledescription_anybrain_0`

Might be better to just translate them but put them in `common` component so they can be used in more places.

GodofMonkeys: I find it. Waite me update.

addons/quickmount/ACE_Settings.hpp
```diff
@@ -36,6 +36,6 @@ class ACE_Settings {
         description = CSTRING(PriorityDescription);
         isClientSettable = 1;
         force = 0;
-        values[] = {CSTRING(PriorityDriver), CSTRING(PriorityGunner), CSTRING(PriorityCommander), CSTRING(PriorityPassenger)};
+        values[] = {"$str_getin_pos_driver", "$str_getin_pos_gunn", "$str_getin_pos_comm", "$STR_GETIN_POS_PASSENGER"};
```
GodofMonkeys: This is might be better :)

jonpas: Awesome!

GodofMonkeys:
I merge with @Dissyu.

PR #5376, opened 2017-07-19
Re enable all markers (except civ and quantity)

chris579:
- Re-enable the placement of the medical markers on the ingame map

Marking a medical FOB ingame would be quite nice so this could be re-enabled.

### DIFF
--- a/addons/map/CfgMarkers.hpp
+++ b/addons/map/CfgMarkers.hpp
@@ -13,7 +13,6 @@ class CfgMarkers {
     class b_hq: b_unknown {scope = 1;};
     class b_installation: b_unknown {scope = 1;};
     class b_maint: b_unknown {scope = 1;};
-    class b_med: b_unknown {scope = 1;};
     class b_service: b_unknown {scope = 1;};
     class b_support: b_unknown {scope = 1;};
 
@@ -22,7 +21,6 @@ class CfgMarkers {
     class n_installation: n_unknown {scope = 1;};
     class u_installation: n_unknown {scope = 1;}; // i have no idea...
     class n_maint: n_unknown {scope = 1;};
-    class n_med: n_unknown {scope = 1;};
     class n_service: n_unknown {scope = 1;};
     class n_support: n_unknown {scope = 1;};
 
@@ -30,7 +28,6 @@ class CfgMarkers {
     class o_hq: o_unknown {scope = 1;};
     class o_installation: o_unknown {scope = 1;};
     class o_maint: o_unknown {scope = 1;};
-    class o_med: o_unknown {scope = 1;};
     class o_service: o_unknown {scope = 1;};
     class o_support: o_unknown {scope = 1;};
 
--- a/addons/map/CfgMarkers.hpp
+++ b/addons/map/CfgMarkers.hpp
@@ -4,8 +4,6 @@ class CfgMarkers {
 
     // Reenable NATO symbols ...
     class b_unknown: Flag {scope = 2;};
-    class n_unknown: b_unknown {};
-    class o_unknown: b_unknown {};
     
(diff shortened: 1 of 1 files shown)


jonpas:
Considering [this comment](https://github.com/chris579/ACE3/blob/ef41d66776b18a2859710a00c87459edb116fece/addons/map/CfgMarkers.hpp#L8-L12) I would say this entire `CfgMarkers` config should be removed, as there is a better way to place markers now.

jonpas:
I guess everyone agrees with me.

SilentSpike:
Yeah I agree 👍

chris579:
Re enabled all markers except civ markers and quantity indicators.

jonpas:
I think that's good. Is `Flag` hidden by default or was it enabled because it inherits from some other class?

chris579:
I think `Flag` is hidden by default and `b_unknown` inherits from it.

SilentSpike:
You can simplify the whole file:

```
class b_unknown;

// disable all civy markers (harbor etc.)
class c_unknown: b_unknown {scope = 1;};

// disable quantity indicators (fire team/squad/platoon ...)
class group_0: b_unknown {scope = 1;};
```

Though personally I don't see why the size and civilian markers should remain disabled. I'd just enable those too and delete this file.

chris579:
Is there a reason why you should be able to create harbor markers while being ingame? But yes, we could delete the whole file.

SilentSpike:
Looking at the vanilla config file now, I'm actually a little confused as there are no markers given `scope = 2`. Surely that would mean in vanilla you cannot place any markers? 😕

chris579:
Afaik you're only able to place the hand drawn markers in vanilla so you have at least to set `scope = 2` to one of the base classes. Therefore deleting the whole config is not possible.

jonpas:
> Afaik you're only able to place the hand drawn markers in vanilla

Which makes sense, given you only have a board marker and no flags on pins.

chris579:
Anyway, is there any reason why somebody should be able to place harbor or hospital markers? I think this would mess up the marker type selection when trying to quickly place a marker.

jonpas:
I personally don't see why, I'd only leave markers that look hand-drawn.

addons/map/CfgMarkers.hpp
```diff
@@ -4,36 +4,9 @@ class CfgMarkers {
...
     class n_unknown: b_unknown {};
-    class n_hq: n_unknown {scope = 1;};
-    class n_installation: n_unknown {scope = 1;};
-    class u_installation: n_unknown {scope = 1;}; // i have no idea...
-    class n_maint: n_unknown {scope = 1;};
-    class n_med: n_unknown {scope = 1;};
-    class n_service: n_unknown {scope = 1;};
-    class n_support: n_unknown {scope = 1;};
-
     class o_unknown: b_unknown {};
```
SilentSpike: This line can also be deleted (it inherits from `b_unknown` already)

addons/map/CfgMarkers.hpp
```diff
@@ -4,36 +4,9 @@ class CfgMarkers {
...
-    // important markers here.
-    // Keep in mind that all of these can still be placed in the editor.
-    class b_hq: b_unknown {scope = 1;};
-    class b_installation: b_unknown {scope = 1;};
-    class b_maint: b_unknown {scope = 1;};
-    class b_med: b_unknown {scope = 1;};
-    class b_service: b_unknown {scope = 1;};
-    class b_support: b_unknown {scope = 1;};
-
     class n_unknown: b_unknown {};
```
SilentSpike: This line can also be deleted (it inherits from `b_unknown` already)

jonpas:
@SilentSpike All good for you?

PR #5378, opened 2017-07-22
Allow explosives to be placed on buildings

tbeswick96:
- Title


PabstMirror:
I just tried this an the explosive doesn't stick to walls, it's just spawned nearby on the ground.
I don't think `attachTo` works on buildings.

tbeswick96:
I actually only tested with a MGB killhouse spawned in zeus, perhaps it does not work with map buildings

PabstMirror:
It probably works with mission placed objects, but terrain buildings won't work.
Besides, attachTo is needed for vehicles because they move, most buildings have a low max speed.

tbeswick96:
In that case scrap this. I'll keep it on my custom branch for use with a breaching charge on the MGB Kill houses doors, but otherwise it doesn't work.
Shoulda tested it more lol

654wak654:
Could still place the explosive in the right relative location and freeze it in the air.

tbeswick96:
Yeah I might revisit it and reopen if I get anywhere

PR #5379, opened 2017-07-22
UAV Self-interactions

tbeswick96:
- Add condition to interact_menu allowing for UAV self-interactions
- Tested with https://github.com/uksf/modpack/blob/drone/addons/reaper/CfgVehicles.hpp#L335 and all works as intended, including when sat in other vehicles.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -300,7 +300,6 @@ TRACE_1("adding unit playerEH to set ace_player",isNull cba_events_oldUnit);
 ["unit", {
     ACE_player = (_this select 0);
 }, true] call CBA_fnc_addPlayerEventHandler;
-ACE_player_previous = ACE_player;
 
 GVAR(OldIsCamera) = false;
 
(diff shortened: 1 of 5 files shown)


commy2:
This looks very strange to me.

PabstMirror:
We use `ace_player` over 1000 times, this changes it with no corresponding event.
A lot of weird things happen, like medical menu opening for the AI.
It also limits things like setting off explosives linked to the real player while viewing camera feed.

There is a keybind for fire mode cycling, so the interaction menu really isn't needed.

Another option would be to add an interaction node to player for the UAV and add them there.
Something like this
```
    class CAManBase: Man {
        class ACE_Actions {
            class ACE_MainActions {
               class ACE_UAV {
                  condition = "!isNull (ACE_controlledUAV param [0, objNul])";

**Comment by tbeswick96** [comment]
OK ignore this, this was the wrong approach.
I'm going to try one more thing that will not touch ACE_Player, and hopefully that will be satisfactory.

**Comment by tbeswick96** [comment]
@commy2 @PabstMirror How does that look? It no longer interferes with ACE_Player at all.

**Comment by commy2** [comment]
Looks good. Not tested. But why use:
```sqf
ACE_controlledUAV param [0, objNull]
```
when it could just be:
```sqf
ACE_controlledUAV select 0
```

Some of these are called quite often and it would be a shame to waste computing time.

tbeswick96:
Updated with that improvement

commy2:
lgtm

PR #5384, opened 2017-07-23
Replace Helicopter eject configs with universal scripted action

Dystopian:
- remove `*canEject` config entries;
- add Eject action with vanilla `addAction` command to appropriate helicopters.

This is replacement of #4749.

- [x] I would like to wait for CBA 3.4.1, there is a little change in `CBA_fnc_getTurret`. But not if it's later then next ACE minor release.

There may be another approach with getIn/getOut EH, but I decided it's more difficult.

~Question: can it happen that initPost EH fires more then once? If yes, many actions will be added.~ OK

Problems with used `moveOut` comand:
1. ~It stops engine while vanilla `Eject` doesn't (driver only).~ Fixed
1. AI subordinate just really moves out by order while with vanilla AI unit doesn't do it until he has a parachute and then he moves out.

### DIFF
--- a/addons/aircraft/CfgEventHandlers.hpp
+++ b/addons/aircraft/CfgEventHandlers.hpp
@@ -1,4 +1,3 @@
-
 class Extended_PreStart_EventHandlers {
     class ADDON {
         init = QUOTE(call COMPILE_FILE(XEH_preStart));
--- a/addons/aircraft/CfgEventHandlers.hpp
+++ b/addons/aircraft/CfgEventHandlers.hpp
@@ -1,0 +1,18 @@
+
+class Extended_PreStart_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preStart));
+    };
+};
+
+class Extended_PreInit_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preInit));
+    };
+};
+
+class Extended_PostInit_EventHandlers {
+    class ADDON {
+        clientInit = QUOTE(call COMPILE_FILE(XEH_postInitClient));
+    };
+};
--- a/addons/aircraft/XEH_PREP.hpp
+++ b/addons/aircraft/XEH_PREP.hpp
@@ -1,3 +1,2 @@
-
 PREP(initEjectAction);
 PREP(canShowEject);
--- a/addons/aircraft/XEH_PREP.hpp
+++ b/addons/aircraft/XEH_PREP.hpp
@@ -1,0 +1,3 @@
+
+PREP(initEjectAction);
+PREP(canShowEject);
--- a/addons/aircraft/XEH_postInitClient.sqf
+++ b/addons/aircraft/XEH_postInitClient.sqf
@@ -20,7 +20,7 @@ GVAR(ejectActionParams) = [
         false,
         true,
         getText (_cfgAction >> "shortcut"),
-        QUOTE([ARR_2(_this,_target)] call DFUNC(canShowEject))
+        '[_this, _target] call DFUNC(canShowEject)'
     ],
     getText (_cfgAction >> "text"),
     getText (_cfgAction >> "textDefault")
--- a/addons/aircraft/XEH_postInitClient.sqf
+++ b/addons/aircraft/XEH_postInitClient.sqf
@@ -20,7 +20,7 @@ GVAR(ejectActionParams) = [
         false,
         true,
         getText (_cfgAction >> "shortcut"),
-        QUOTE([ARR_2(_this,_target)] call FUNC(canShowEject))
+        QUOTE([ARR_2(_this,_target)] call DFUNC(canShowEject))
     ],
     getText (_cfgAction >> "text"),
     getText (_cfgAction >> "textDefault")
--- a/addons/aircraft/XEH_postInitClient.sqf
+++ b/addons/aircraft/XEH_postInitClient.sqf
@@ -6,7 +6,15 @@ private _cfgAction = configFile >> "CfgActions" >> "Eject";
 GVAR(ejectActionParams) = [
     [
         "", // will be set with setUserActionText
-        {moveOut (_this select 1)},
+        {
+            params ["_vehicle", "_unit"];
+            private _preserveEngineOn = (_unit == driver _vehicle) && {isEngineOn _vehicle};
+            moveOut _unit;
+            if (_preserveEngineOn) then {
+                // vehicle is local to last driver, no need to care
+                _vehicle engineOn true;
+            };
+        },
         nil,
         getNumber (_cfgAction >> "priority"),
         false,
--- a/addons/aircraft/XEH_postInitClient.sqf
+++ b/addons/aircraft/XEH_postInitClient.sqf
@@ -1,0 +1,21 @@
+#include "script_component.hpp"
+
+if (!hasInterface) exitWith {};
+
+private _cfgAction = configFile >> "CfgActions" >> "Eject";
+GVAR(ejectActionParams) = [
+    [
+        "", // will be set with setUserActionText
+        {moveOut (_this select 1)},
+        nil,
+        getNumber (_cfgAction >> "priority"),
+        false,
+        true,
+        getText (_cfgAction >> "shortcut"),
+        QUOTE([ARR_2(_this,_target)] call FUNC(canShowEject))
+    ],
+    getText (_cfgAction >> "text"),
+    getText (_cfgAction >> "textDefault")
+];
+
+["Helicopter", "initPost", LINKFUNC(initEjectAction)] call CBA_fnc_addClassEventHandler;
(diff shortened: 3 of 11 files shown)


tbeswick96:
> It stops engine while vanilla Eject doesn't (driver only).

I personally dislike that. If i'm the pilot of a heli and need to hop out to quickly sort something out, I don't want the aircraft to spool down. Maybe add something that checks if the engine is on before moving the unit out, and if so, turn it back on once out?

jonpas:
What @tbeswick96 says, that's one the main points of having the "Eject" action. Is there a way to work around this in a scripted way?

commy2:
>Question: can it happen that initPost EH fires more then once?

No. Once per vehicle.

commy2:
>Problems with used moveOut comand

What about this?
https://community.bistudio.com/wiki/Arma_3_Actions#Eject

Dystopian:
> What about this?

[my last comment at #4749](https://github.com/acemod/ACE3/pull/4749#issuecomment-315598560)
>`_unit action ["Eject", vehicle _unit]` doesn't work naturally.

Maybe because action is absent.

commy2:
OK, makes sense I guess. ACE2 had jump out left and right actions. Just something I remember.

jonpas:
Someone with ACE2 source should take a look (I would, but not going to browse code on phone).

Dystopian:
> ace_sys_eject\jumpout.sqf

```
// New implementation brings new issues...
_engine = isEngineOn _vehicle;

// Make jumper get out of a vehicle (without parachute).
moveOut _jumper;

// ...solution for moveOut stopping the engine
if (_engine && {(!isEngineOn _vehicle)}) then { _vehicle action ["EngineOn", _vehicle] };
```
I'll implement engine workaround.

addons/aircraft/CfgEventHandlers.hpp
```diff
@@ -0,0 +1,18 @@
+
```
Neviothr: Redundant whitespace at the top of the file.

addons/aircraft/XEH_PREP.hpp
```diff
@@ -0,0 +1,3 @@
+
```
Neviothr: Same.

commy2: I like those for included files personally. Makes it easier to read the decompiled code later sometimes.

Dystopian: One of the most difficult thing for me in this PR was creating new ACE-specific files without templates but with many samples in the source tree. And unfortunately these samples are different in some aspects like these spaces at file start.

commy2: Yes, because people delete them with the same argument as Neviothr made. Do what you like best. It really doesn't matter.

jonpas: There is a basic template in extras. Whitespace at the top is  redundant, you should never read decompiled code unless debugging tools or something.

jonpas:
ACE2 just hid the vanilla "Eject" action with the following. Unsure what's better, one keeps vanilla where it is the other makes it consistent, so if there is a problem it will be everywhere (problem in either vanilla or ACE3's).
```
class CfgActions {
	class None;
	class Eject: None {
		show = 0;
		shortcut = "";
	};
};
```

commy2:
A custom action added with `addAction` can use a shortcut.
List: https://community.bistudio.com/wiki/inputAction/actions
It would be `"Eject"`.

Dystopian:
done

addons/aircraft/functions/fnc_canShowEject.sqf
```diff
@@ -0,0 +1,36 @@
...
+            _x params ["_owner", "_role", "", "_turretPath"];
+            if (_unit == _owner) exitWith {
+                format ["%1_%2_%3", QGVAR(ejectAction), _role, _turretPath]
+            };
+            false
+        } count fullCrew _vehicle;
+        // getVariable ["", ...] returns default value
+        ["", _return] select (_return isEqualType "")
+    },
+    false
```
commy2: My brain hurts.

addons/aircraft/functions/fnc_initEjectAction.sqf
```diff
@@ -0,0 +1,44 @@
...
+ * Example:
+ * _vehicle call ace_aircraft_fnc_initEjectAction
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_vehicle"];
+
+if (unitIsUAV _vehicle) exitWith {};
```
commy2: If anyone wonders, this command is right. It reports true for unmanned vehicles, not UAV dummy units like the name implies.

addons/aircraft/functions/fnc_initEjectAction.sqf
```diff
@@ -0,0 +1,44 @@
...
+    _addAction = true;
+};
+
+{
+    _x params ["", "_role", "", "_turretPath"];
+    if (0 == getNumber (([_config, _turretPath] call CBA_fnc_getTurret) >> "canEject")) then {
+        _vehicle setVariable [format ["%1_%2_%3", QGVAR(ejectAction), _role, _turretPath], true];
+        _addAction = true;
+    };
+} forEach (fullCrew [_vehicle, "gunner", true] + fullCrew [_vehicle, "turret", true]);
```
commy2: commander?

addons/aircraft/functions/fnc_canShowEject.sqf
Dystopian: Performance, not Readability, sorry. All I can do for this implementation is to add more comments and/or correct indents.

addons/aircraft/functions/fnc_initEjectAction.sqf
Dystopian: There are no `Air` vehicles ATM in vanilla + RHS + CUP with `commander` in `fullCrew` (just checked). Copilot is gunner. And I think there will not.

addons/aircraft/functions/fnc_canShowEject.sqf
commy2: You could get rid of the type validation of the first params for performance.

commy2:
```sqf

params ["_unit", "_vehicle"];

_vehicle == vehicle _unit && {
    private _return = "";

    {
        _x params ["_owner", "_role", "", "_turretPath"];

        if (_unit isEqualTo _owner) exitWith {
            _return = format ["%1_%2_%3", QGVAR(ejectAction), _role, _turretPath]
        };
    } count fullCrew _vehicle;

    _vehicle getVariable [_return, false] // return
};
```
???

commy2: Btw, shouldn't the turret path already make this unique? Do we need the role at all?

addons/aircraft/functions/fnc_initEjectAction.sqf
jonpas: I think we should still add it, it can't hurt, right?

addons/aircraft/functions/fnc_canShowEject.sqf
Dystopian: Your code is faster and more readable, thank you.
The role is deleted too.

addons/aircraft/functions/fnc_initEjectAction.sqf
Dystopian: sure, why not.

addons/aircraft/functions/fnc_canShowEject.sqf
Dystopian: The role was to distinguish driver and cargo man (both have `[]` turret). I have to restore it.

Dystopian:
done

addons/aircraft/XEH_postInitClient.sqf
```diff
@@ -0,0 +1,29 @@
...
+                // vehicle is local to last driver, no need to care
+                _vehicle engineOn true;
+            };
+        },
+        nil,
+        getNumber (_cfgAction >> "priority"),
+        false,
+        true,
+        getText (_cfgAction >> "shortcut"),
+        QUOTE([ARR_2(_this,_target)] call DFUNC(canShowEject))
```
commy2: Just a tip, you could've written this as:
```sqf
'[_this, _target] call DFUNC(canShowEject)'
```
This is fine too though.

Dystopian: that's amazing

jonpas:
All good @commy2 ?

PabstMirror:
As far as I know you simply can't eject from some types of helicopter like the Apache.
I think there might be a Russian attack chopper that actually has an ejection seat but it destroys the rotors in the process (so if copilot ejects, the pilot pretty much has to as well)

jonpas:
Yeah, that's all true. But the main reason we really have this is to allow getting out of helicopters without spooling it down, because in Arma you often just have to run to something for like a second, can't really mimic real life. Also ACE2 did the same AFAIK.

Maybe there is a good way to limit it to only be doable on the ground? And the actual action through config can be for air then, so only those with config action support ejection?

PR #5397, opened 2017-07-31
added movable markers option

chris579:
- Add the possibility to move markers placed by players (default: disabled)
- Add a settings module

Markers can be moved after placing by holding the left mouse button. As indication for the movement the marker gets an alpha of 0.5 and the map cursor is changing. The final position is globally applied when the movement process has been stopped. You can only move markers on the main map and nowhere else like briefing or zeus (could be added when needed).

[Demo on YT](https://www.youtube.com/watch?v=bAQTgsNzf4I)


bux:
Nice idea. Needs restrictions though on who can move markers (Admin, Group leaders, all, ...) and what markers (maybe by channel).

chris579:
I already thought about restricting this. The only harm you can do by moving a marker is that it is on a wrong position. If somebody would do harm to markers he could just delete it. So I don't think that this should be restricted.

bux:
This might hold true for smaller private groups. Imagine bigger public events or even public servers where it would make sense to use the functionality, there a restriction is needed.

chris579:
I agree to your point but then it should also be considered that there should be a limitation for deleting markers

bux:
With arguments like 
> because it's already bad in vanilla
there wouldn't be ACE

Imho we have an opportunity to do it better than it's done in vanilla 😄

bux:
Some ideas that have been raised in chat:

* an event fired for a loved marker 
* accidental move protection (e.g. modifier key)

chris579:
Sorry, I think my statement was misleading. It was meant the way that such a system to prevent marker deletion should also be added to prevent abuse of markers on both ways.

Dystopian:
Maybe just on mouse hover show name of player who moved marker? Like vanilla does with created markers.
Anyway I think restrictions are not for this PR. This feature is not enabled by default and can be used by those people who understand where it can go.

chris579:
> an event fired for a moved marker

Added a local event
Also added some basic restrictions (All, Admins, Group leaders, Group leaders & Admins).

> accidental move protection (e.g. modifier key)

Not sure which one to pick. Shift is used for placing the target marker, Ctrl for drawing and Alt for teleporting in debug mode

addons/markers/functions/fnc_canMoveMarker.sqf
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+switch (GVAR(moveRestriction)) do {
+    case MOVE_RESTRICTION_ALL: { true };
+    case MOVE_RESTRICTION_ADMINS: { IS_ADMIN };
+    case MOVE_RESTRICTION_GROUP_LEADERS: {
+        leader group ACE_player == leader ACE_player
```
Dystopian: `leader group ACE_player == ACE_player`?

addons/markers/ACE_Settings.hpp
```diff
@@ -0,0 +1,16 @@
+class ACE_Settings {
+    class GVAR(MovableMarkersEnabled) {
```
jonpas: Lower-case GVARs (movableMarkers...).

addons/markers/functions/fnc_canMoveMarker.sqf
chris579: Whats wrong with it?

addons/markers/XEH_postInit.sqf
```diff
@@ -18,3 +18,24 @@ GVAR(currentMarkerPosition) = [];
...
 GVAR(currentMarkerConfigName) = "";
+
+// set marker pos local on every computer (prevent markers visible for everyone)
+[QGVAR(applyMarkerPosLocal), {
+    params["_marker", "_pos"];
+    _marker setMarkerPosLocal _pos;
+}] call CBA_fnc_addEventHandler;
+
+["ace_settingsInitialized", {
+    if(GVAR(movableMarkersEnabled)) then {
```
jonpas: Space after 'if' as per coding guidelines.

addons/markers/functions/fnc_canMoveMarker.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * Checks whether the player can move markers.
+ *
+ * Arguments:
+ * None
+ *
+ * Return Value:
+ * Whether the player can move markers <BOOL>
+ *
+ * Example:
+ * [] call ACE_markers_fnc_canMoveMarker
```
jonpas: Lower-case "ACE".

addons/markers/functions/fnc_movePFH.sqf
```diff
@@ -0,0 +1,34 @@
+/*
+ * Author: chris579
+ * When the marker is being moved.
+ *
+ * Arguments:
+ * 0: Marker data <ARRAY>
+ * 1: PFH id <NUMBER>
```
jonpas: "ID"

addons/markers/functions/fnc_canMoveMarker.sqf
Dystopian: I thought `leader == leader` is not what you wanted.

addons/markers/functions/fnc_onMouseButtonDownMap.sqf
```diff
@@ -0,0 +1,51 @@
+/*
+ * Author: chris579
+ * Triggered when a mouse button is pressed on the map.
+ *
+ * Arguments:
+ * 0: Map Control the evh was assigned to <CONTROL>
```
jonpas: "EVH" or "event handler".

addons/markers/functions/fnc_canMoveMarker.sqf
chris579: Oh snap, you're right

jonpas:
Unless we can also protect who can delete markers there is little point in protecting who can move them. If we can we should protect both.

addons/markers/XEH_postInit.sqf
```diff
@@ -18,3 +18,24 @@ GVAR(currentMarkerPosition) = [];
...
+// set marker pos local on every computer (prevent markers visible for everyone)
+[QGVAR(applyMarkerPosLocal), {
+    params["_marker", "_pos"];
+    _marker setMarkerPosLocal _pos;
+}] call CBA_fnc_addEventHandler;
+
+["ace_settingsInitialized", {
+    if (GVAR(movableMarkersEnabled)) then {
+        addMissionEventHandler ["Map", {
+            params["_isOpen"];
```
Dystopian: space

addons/markers/functions/fnc_movePFH.sqf
```diff
@@ -0,0 +1,34 @@
...
+ */
+#include "script_component.hpp"
+
+params ["_args", "_idPFH"];
+_args params ["_marker"];
+
+if (isNull (findDisplay 12 displayCtrl 51) || !(player getVariable [QGVAR(moveInProgress), false])) exitWith {
+    [_idPFH] call CBA_fnc_removePerFrameHandler;
+
+    _marker = (player getVariable[QGVAR(movedMarker), ""]);
```
Dystopian: space and no need for (). And maybe `_marker` param changing

addons/markers/functions/fnc_movePFH.sqf
```diff
@@ -0,0 +1,34 @@
...
+ * [[MARKER], 5] call ace_markers_fnc_movePFH
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_args", "_idPFH"];
+_args params ["_marker"];
+
+if (isNull (findDisplay 12 displayCtrl 51) || !(player getVariable [QGVAR(moveInProgress), false])) exitWith {
```
Dystopian: lazy?

addons/markers/XEH_postInit.sqf
```diff
@@ -18,3 +18,24 @@ GVAR(currentMarkerPosition) = [];
...
+
+            if (_isOpen) then {
+                (findDisplay 12 displayCtrl 51) ctrlAddEventHandler ["MouseButtonDown", FUNC(onMouseButtonDownMap)];
+                (findDisplay 12 displayCtrl 51) ctrlAddEventHandler ["MouseButtonUp", FUNC(onMouseButtonUpMap)];
+            } else {
+                player setVariable [QGVAR(moveInProgress), false];
```
Dystopian: I don't really know but maybe `player` -> `ACE_player` here and everywhere

bux:
>Unless we can also protect who can delete markers there is little point in protecting who can move them.

I disagree

654wak654:
>Unless we can also protect who can delete markers there is little point in protecting who can move them.
>>I disagree

I disagree back. Why try to introduce limitations to an already simple system? You out of everyone know giant arma events already have 50 other problems before it comes to "who can move markers ((Even though everyone can still place / delete them))"

addons/markers/functions/fnc_movePFH.sqf
chris579: Tbh your comments on my code are not really helpful unless. Feel free to criticize it directly but such comments are just suboptimal.

addons/markers/XEH_postInit.sqf
chris579: Like already stated I can't work out the meaning of your single word comments.

addons/markers/functions/fnc_movePFH.sqf
CorruptedHeart: He's meaning you should use lazy evaluation.
```
if ( (statement) || { another statement })
```
Arma doesn't lazy evaluate if statements, so both sides of `or` will always execute if the subsequent checks aren't wrapped in `{ }`

chris579: Thanks for the explanation but I'm aware of lazy evaluations. My point is that the statement can be interpreted different ways. "Use lazy evaluations" would be more straightforward.

Dystopian: No, thank you. Interpret it the most suitable way. Nobody is interested if _you_ are lazy but _evaluation_.

chris579: Ok, let's be more straightforward:
Throwing contextless words into the room is bs. And I can't read your dystopian mind.

chris579:
Whats about this pr? The requested changes are done.

addons/markers/ACE_Settings.hpp
```diff
@@ -0,0 +1,16 @@
...
+        typeName = "BOOL";
+        displayName = CSTRING(MovableMarkers_DisplayName);
+        description = CSTRING(MovableMarkers_Description);
+    };
+    class GVAR(moveRestriction) {
+        category = CSTRING(Module_DisplayName);
+        value = 0;
+        typeName = "SCALAR";
+        displayName = CSTRING(MoveRestriction_DisplayName);
+        description = CSTRING(MoveRestriction_Description);
```
jonpas: Should provide values, as such https://github.com/acemod/ACE3/blob/21502d7b7bc2516fdf3d674355d7ae1e892c65af/addons/missileguidance/ACE_Settings.hpp#L7

addons/markers/XEH_postInit.sqf
jonpas: `params ["_isOpen"];`

addons/markers/XEH_postInit.sqf
jonpas: Yeah, `ACE_player` should be used, as you might use that as a remote controlled unit as well.

addons/markers/functions/fnc_movePFH.sqf
jonpas: `if (isNull (findDisplay 12 displayCtrl 51) || {!(player getVariable [QGVAR(moveInProgress), false])}) exitWith {`

addons/markers/ACE_Settings.hpp
chris579: Added them

SilentSpike:
So, I'm kind of surprised nobody else has mention it, but why reinvent the wheel?

Markers can already be moved by deleting and remaking them so I question whether this is really worth all the effort of adding restrictions and settings.

chris579:
I definitely see a difference between the possibility of moving a marker and deleting and remaking it.

Especially when in combat it's really helpful when you can move a marker quickly without the need of remaking it. Think of a situation where you are a spotter and you failed to place the tango marker on the right position and you want to move it 100 meters to the west.

I don't know why there is so much trouble about restrictions and such stuff. It is an optional thing just like BFT and can be enabled by players that want to move markers. Nobody is forced to enable it.

@654wak654 is completely right stating that the vanilla marker system is degenerated and has to be overhauled. But this is not the responsibility of ACE or other mods, it's the responsibility of BI to do needed changes.

Like I already said: this feature was developed for a single community after we were annoyed by redoing markers just because somebody missed the right destination for it. I thought this could be neat and other people could profit from this.

*I can't really understand the trouble that came up because of this optional (!) feature.*
Of course you are right that there should be restrictions for it. But first there have to be restrictions for deleting markers before you even should think about restricting who can move markers. It's like a castle mound that has a big hole and you are about to fix a small hole in the gate.

SilentSpike:
I think you misunderstand, I agree with yourself and @654wak654 that it's not worth implementing restrictions when they can just be side-stepped by the engine behaviour 😄

You also make a good point about the spotter scenario, that's something I hadn't considered. However, I will say nobody is giving you trouble. Discussion is entirely the purpose of github's pull request and issue system. Part of the process is to figure out whether the code fits into the scope of the project as a whole - PR's aren't intended as a means for anything and everything to be eventually merged into the project.

chris579:
@SilentSpike my post was not a reaction to yours, more a reaction to the whole discussion. I'm aware of the fact that not all PR's are making it to the final product. I can live with that. 
My point is that the discussion missed the essential part that markers can be sabotaged by everyone. The focus was on protecting who can move markers and not how this feature could be extended so that it could be usable in more situations and not only milsim.

Imo ACE is a collection of nice features and gimmicks (think of LSD vehicles 😄) so optional features are an essential part of it. This is such an optional feature.

jonpas:
I think this perfectly fits ACE3, it's a real hassle to remove and remake the marker, simply because you lose the text associated with it. You could say in real life you also have to erase it and mark it again, but then again you could use pins! 😆

chris579:
> Also, has Ctrl modifier been implemented yet? Or have we decided not to, I forgot?

Not sure which one to pick. Shift is used for placing the target marker, Ctrl for drawing and Alt for teleporting in debug mode

SilentSpike:
Code all looks sound to me. Still not convinced on the necessity of the restriction stuff, but it's not a huge deal 👍

jonpas:
Well it's there now, so if BI ever adds restrictions for deleting markers, we got it sorted! I still wonder if it's possible to add deletion restrictions ourselves though.

chris579:
If it's not engine based this should be possible.

jonpas:
I'd say use Alt as modifier, it only works in single-player and is so rarely used anyways.

chris579:
Set alt as modifier key

jonpas:
How far is this tested? Looks good to me overall.

chris579:
This feature was used for months in our community. After integrating it to ACE I tested it quickly whether the settings module works and if the move conditions are applied. For me it works good but it would be nice if somebody else could test it too.

addons/markers/XEH_postInit.sqf
```diff
@@ -18,3 +18,24 @@ GVAR(currentMarkerPosition) = [];
...
+    _marker setMarkerPosLocal _pos;
+}] call CBA_fnc_addEventHandler;
+
+["ace_settingsInitialized", {
+    if (GVAR(movableMarkersEnabled)) then {
+        addMissionEventHandler ["Map", {
+            params ["_isOpen"];
+
+            if (_isOpen) then {
+                (findDisplay 12 displayCtrl 51) ctrlAddEventHandler ["MouseButtonDown", FUNC(onMouseButtonDownMap)];
```
PabstMirror: This adds a new map EH every-time the map is opened, which means multiple PFEHs get started on mouse down

addons/markers/functions/fnc_movePFH.sqf
```diff
@@ -0,0 +1,38 @@
...
+
+params ["_args", "_idPFH"];
+_args params ["_marker"];
+
+if (isNull (findDisplay 12 displayCtrl 51) || {!(player getVariable [QGVAR(moveInProgress), false])}) exitWith {
+    [_idPFH] call CBA_fnc_removePerFrameHandler;
+
+    private _origin = ACE_player getVariable [QGVAR(movedMarkerOrigin), getMarkerPos _marker];
+
+    if !([QGVAR(markerMoveEnded), [ACE_player, _marker, _origin, getMarkerPos _marker]] call CBA_fnc_localEvent) exitWith {
```
PabstMirror: Event doesn't exist in ace, is this just so something external can cancel marker movement?
`if (nil) ...` always worries me a little, but it should be fine as is

jonpas: Event was requested in Slack chat yeah.

addons/markers/XEH_postInit.sqf
chris579: I think the display (or the control) get's destroyed when you close the map

PabstMirror: We usually only add the EH once, see https://github.com/acemod/ACE3/blob/master/addons/map/XEH_postInitClient.sqf#L24

This causes problems with the marker alpha because it saves the 0.5, also sends multiple events over the network

```
[{
    !isNull (findDisplay 12)
}, {
    (findDisplay 12 displayCtrl 51) ctrlAddEventHandler ["MouseButtonDown", FUNC(onMouseButtonDownMap)];
    (findDisplay 12 displayCtrl 51) ctrlAddEventHandler ["MouseButtonUp", FUNC(onMouseButtonUpMap)];
}] call CBA_fnc_waitUntilAndExecute;
```
Should work nicely

addons/markers/XEH_postInit.sqf
```diff
@@ -18,3 +18,24 @@ GVAR(currentMarkerPosition) = [];
 GVAR(currentMarkerAngle) = 0;
 GVAR(currentMarkerColorConfigName) = "";
 GVAR(currentMarkerConfigName) = "";
+
+// set marker pos local on every computer (prevent markers visible for everyone)
+[QGVAR(applyMarkerPosLocal), {
+    params["_marker", "_pos"];
+    _marker setMarkerPosLocal _pos;
```
PabstMirror: To make this jip compatible, add the following here:
```
    if (isServer) then {
        private _index = (GETGVAR(allMapMarkers,[])) find _marker; // case-sensitive, but should be fine
        if (_index < 0) exitWith {ERROR_1("Could not find data for %1", _marker);};
        private _data = GVAR(allMapMarkersProperties) select _index;
        _data set [2, _pos];
    };
```

PabstMirror: All units will run marker's `FUNC(setMarkerJIP)` which will set marker color, dir, pos and shape. The pos would be from when it was originally placed, this code will update the data to have the moved position.

chris579: Isn't your sample missing the line `GVAR(allMapMarkersProperties) set [_index, _data]` after `_data set [2, _pos];`. The changes I make to `_data` are not set in `GVAR(allMapMarkersProperties)`.

PabstMirror: It's an array so this should be fine. (arrays are pass by reference)

chris579: Oh well, nice to know, then nvm.

chris579: Added this

PR #5402, opened 2017-08-02
Fix cars loaded as cargo taking water damage

### LINKED ISSUE #5398: Vehicles in cargo break after unloading
**Arma 3 Version:** 1.72 stable 
**CBA Version:** 3.4.0 stable 
**ACE3 Version:** 3.10.2 stable 

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- Anytime you load a vehicle into the ace cargo of another vehicle and unload it, the vehicle bugs out. 

**Steps to reproduce:**
- Spawn ATV set so the ATV can be loaded 
- Spawn Truck with cargo space
- Unload the ATV and get in and try to drive

- All

**Placed Modules:**
- None

**RPT log file:**
- Not needed

PabstMirror:
Fix #5398

Seems like cars are hard-coded to take water damage,
`_item attachTo [_vehicle,[0,0,-100]];` can put the car below 0 ASL which triggers the effect
(tested on a 75m high hill, -70 was fine, -80 caused damage)

Once water logged the vehicle seems to be permently disabled (fixing HitEngine does not fix)

This just disables damage (`allowDamage`) on the car while in cargo

### DIFF
--- a/addons/cargo/XEH_postInit.sqf
+++ b/addons/cargo/XEH_postInit.sqf
@@ -50,6 +50,11 @@
 
     _item hideObjectGlobal false;
     _item setPosASL (AGLtoASL _emptyPosAGL);
+
+    if ((getText (configFile >> "CfgVehicles" >> (typeOf _item) >> "simulation")) == "carx") then {
+        TRACE_1("re-enabling car damage",_item);
+        [_item, "blockDamage", "ACE_cargo", false] call EFUNC(common,statusEffect_set);
+    };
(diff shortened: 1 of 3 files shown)


bux:
What happens to the loaded car if the truck gets blown up?

PabstMirror:
https://github.com/acemod/ACE3/blob/master/addons/cargo/functions/fnc_handleDestroyed.sqf

Should just get deleted like normal

PR #5238, opened 2017-06-03
Add Aircraft Loadout Dialog

654wak654:
**When merged this pull request will:**
- This: http://imgur.com/a/dkP4i

**TODO List:**
- [ ] Blacklist / Whitelist functionality.
- [ ] Whatever the fuck that'll please the `rearm` overlords get this merged before ACE 3.15.0
- [ ] Test locality.

PR #5411, opened 2017-08-08
Rearm bugfixing

### LINKED ISSUE #5289: Rearming only rearms the first magazine to a vehicle
**Arma 3 Version:** `1.70.141838` (stable)
**ACE3 Version:** `3.10.0` (stable), also happened in `3.9.2` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
Rearming only rearms the first magazine to a vehicle with multiple magazines. For example, the M-ATV (HMG) has 2 HMG magazines. If you shoot them empty, then rearm the vehicle, only one magazine is rearmed. There is no way to rearm it more after that.

**Steps to reproduce:**
- Create an M-ATV (HMG).
- Create an HEMTT Ammo next to it.
- Go to the M-ATV as gunner and shoot all its ammo until empty. (Note, you have to expend the ammo by shooting it. Simply setting "ammunition" to 0 in the editor does not trigger the bug.)
- Rearm the M-ATV from the ammo truck.
- Observe how the M-ATV has only one magazine in its HMG and how the "Rearm"-interaction is no longer available in the ammo truck.

Everywhere, both multi- and singleplayer.

**Placed Modules:**
None

Tuupertunut:
### Major changes

- Rearming now correctly fills all magazines with Rearm amount: "Entire vehicle" setting. (partially fixes #5289)
- Empty pylon ammo selection ([this feature](https://ace3mod.com/img/news/170714_jet_rearm.jpg)) is no longer supported in favor of the upcoming "pylons" ACE-module (#5238).
  - It was in the way of fixing some bugs and I didn't see it as necessary to maintain just to be replaced with a better module soon anyway.
  - It never was a complete feature. You couldn't use it if "Rearm amount" setting was "Entire vehicle" (the default setting). You could only add missiles to empty pylons but not change missiles in already populated ones. You also couldn't choose which pylon something gets attached to.

### Minor changes

- Turrets are now retrieved dynamically from the vehicle. No need to maintain a list of possible turret paths manually anymore.
- Empty pylons are now recognized correctly and ACE no longer tries to set their ammo count on entire vehicle rearm.

### Development-only changes

- Moved all partial magazine checking code to one function; `fnc_getNeedRearmMagazines`.
- Replaced a few failing BIS commands with workaround functions.
  - These are `fnc_getAllTurrets`, `fnc_getTurretMagazineAmmo` and `fnc_setTurretMagazineAmmo`.
  - More info in those functions' comment headers.
- Renamed `fnc_getVehicleMagazines` to `fnc_getTurretConfigMagazines`.
  - The new name describes much better what the function actually does.

### Current progress

\* | **Rearm amount: Entire vehicle** | **Entire magazine** | **Amount based on caliber**
--- | --- | --- | ---
**Unlimited ammo supply** | Fully fixed | Vehicle can't be fully rearmed in some cases. Still uses `setMagazineTurretAmmo`. | Vehicle can't be fully rearmed in some cases. Still uses `setMagazineTurretAmmo`.
**Limited ammo supply based on caliber** | Takes wrong amount from the ammo truck. | Takes wrong amount from the ammo truck. Vehicle can't be fully rearmed in some cases. Still uses `setMagazineTurretAmmo`. | Takes wrong amount from the ammo truck. Vehicle can't be fully rearmed in some cases. Still uses `setMagazineTurretAmmo`.
**Only specific magazines** | Not tested. | Not tested. | Not tested.


jonpas:
> no longer supported in favor of the upcoming "pylons" ACE-module
> to be replaced with a better module soon anyway

#5238 is NOT a rearm thing! The pylons menu is for changing what is on the pylons, we decided to have 3 settings for them (disabled, enabled at specific locations/vehicles, enabled everywhere). That way people who want realism (default setting - at locations) will only be able to change what's on the pylons at a proper facility, while others can enable it everywhere. Rearm still has to be fully compatible with pylons on its own, and pylons component is rather going to use rearm function to accomplish compatibility with limited rearm.

> You could only add missiles to empty pylons but not change missiles in already populated ones.

Rearm is not supposed to change what's on the pylons, only rearm them (add missiles in this case). Same explanation as above.

addons/rearm/functions/fnc_getAllTurrets.sqf
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params [
+    ["_vehicle", objNull, [objNull]]
+];
+
+if (isNull _vehicle) exitWith {[]};
```
jonpas: Can that ever happen in the first place? It's an internal function, so double checking things is redundant and will make it slower.

addons/rearm/functions/fnc_rearm.sqf
```diff
@@ -27,11 +27,16 @@ if (isNil "_magazineClass") exitWith {ERROR_1("magazineClass nil",_attachedDummy
...
+private _needRearmMagsOfClass = _needRearmMags select {(_x select 0) isEqualTo _magazineClass};
 
 // Exit if no magazines need rearming
-if (!_needRearm) exitWith {ERROR_2("Could not find turret for %1 in %2", _magazineClass, typeOf _target);};
+if ((count _needRearmMagsOfClass) == 0) exitWith {ERROR_2("Could not find turret for %1 in %2", _magazineClass, typeOf _target);};
+
+private _currentRearmableMag = _needRearmMagsOfClass select 0;
+private _turretPath = _currentRearmableMag select 1;
+private _pylon = _currentRearmableMag select 3;
+private _magazineCount = _currentRearmableMag select 5;
```
jonpas: Maybe the following to shorten it a bit (I think it's also cleaner what is from what)?
```
_needRearmMagsOfClass params ["_currentRearmableMag"];
_currentRearmableMag params ["", "_turretPath", "", "_pylon", "", "_magazineCount"];
```

Tuupertunut:
In this PR, rearm does support pylons. It will fill empty pylon magazines, meaning that if there was originally  a missile but it was shot, it will be refilled. What I mean by "empty pylon" here is a pylon that never had any missile configured. The way I see it, filling (configuring) **those** pylons is the job of the "pylons" module, while rearming spent missiles is the job of "rearm" module.

jonpas:
Ah, I misunderstood then! That is perfect then, as you said, pylons component is supposed to do that, as IRL it is not as simple as just throwing up another missile, go to change the rack as well!

addons/rearm/functions/fnc_getAllTurrets.sqf
Tuupertunut: It probably can not. I only ever added it because some other functions (like [fnc_getMaxMagazines](https://github.com/acemod/ACE3/blob/master/addons/rearm/functions/fnc_getMaxMagazines.sqf)) had it too. I'd be happy to remove it from all of those.

jonpas: It is redundant, as well as the extensive `params` type checking. General rule of thumb for such checking is for public functions where you really want to check everything to make sure someone didn't pass wrong data.

addons/rearm/functions/fnc_rearm.sqf
Tuupertunut: I would retain the row 36 as `select 0` because it literally means "get the first magazine from an array of many magazines". On the rows 37-39 I agree with you, they could be changed.

Anyway, going to sleep now. You can suggest changes to me all night. 😄

jonpas: `params` is a bit faster though, since it's `private` + assignment + `select`. 😆

Tuupertunut: So it's the classic case of performance vs. readability. Usually readability wins in most cases, unless the performance difference is really big. Any objections?

jonpas: It's not big I assume, since it doesn't do anything heavy. I'd say readability is better with `params` myself, but up to you how you want to keep it.

Dystopian: https://ace3mod.com/wiki/development/coding-guidelines.html#65-private-variables

Tuupertunut: In SQF, arrays can be used as lists or pseudo-objects. Pseudo-object is when you use an array to group together multiple related values (like function parameters). There is always a fixed number of elements and they all have a specific purpose. List is when you use an array in the more traditional sense, to hold 0-n values of the same type.

On pseudo-objects, it's clearly better to use `params` to get the individual values. On lists, there rarely is even a need to get the element in a specific index. Those cases are mostly getting either the first or the last element.

I'd say that when getting the first element from a list type array, `_arr select 0` is clearer than `_arr params ["_firstElem"]` When using `params` on a list type array, it makes me think that it is a pseudo-object and the first element has some specific meaning.

654wak654: You are correct on the last paragraph but I think we can all agree `_currentRearmableMag` is better off as a struct.

addons/rearm/functions/fnc_getAllTurrets.sqf
Tuupertunut: Fixed.

addons/rearm/functions/fnc_rearm.sqf
Tuupertunut: Fixed.

addons/rearm/functions/fnc_getAllTurrets.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_vehicle"];
+
+private _turrets = allTurrets _vehicle;
+
+// Adding the driver turret "[-1]".
+_turrets pushBack [-1];
```
Dystopian: instead of creating new function, you could use `[[-1]] + allTurrets _vehicle` syntax, as said at [allTurrets Notes](https://community.bistudio.com/wiki/allTurrets)

addons/rearm/functions/fnc_rearmEntireVehicleSuccessLocal.sqf
```diff
@@ -17,60 +17,42 @@
  */
 #include "script_component.hpp"
 
-params [["_truck", objNull, [objNull]], ["_vehicle", objNull, [objNull]], ["_turretPath", [], [[]]]];
-TRACE_3("rearmEntireVehicleSuccessLocal",_truck,_vehicle,_turretPath);
+params ["_truck", "_vehicle", "_turretPath"];
+TRACE_3("rearmEntireVehicleSuccessLocal", _truck, _vehicle, _turretPath);
```
Dystopian: spaces inside macro are not used on purpose

addons/rearm/functions/fnc_rearmEntireVehicleSuccessLocal.sqf
```diff
@@ -17,60 +17,42 @@
...
         for "_idx" from 1 to (_maxMagazines - _currentMagazines) do {
-            if ((GVAR(supply) == 0) || {[_truck, _magazine, _maxRounds] call FUNC(removeMagazineFromSupply)}) then {
-                _vehicle addMagazineTurret [_magazine, _turretPath];
+            if ((GVAR(supply) == 0) || {[_truck, _magazineClass, _maxRoundsPerMag] call FUNC(removeMagazineFromSupply)}) then {
+                _plannedRounds pushBack _maxRoundsPerMag;
             };
         };
+    };
+    
+    TRACE_2("rearming", _x, _plannedRounds);
```
Dystopian: also spaces and anywhere else

addons/rearm/functions/fnc_getAllTurrets.sqf
jonpas: `+` is rather slow, it's better to use `append`/`pushBack` (`pushBack` should be faster, but not sure). I don't see much point in having a separate function for that though, in how many places is it used?

Tuupertunut: It's used in three different files, so it is pretty widely used. But even if it wasn't, it's supposed to be a fixed drop-in replacement function for `allTurrets`. You can use it everywhere where you would normally use `allTurrets`, it just works better.

addons/rearm/functions/fnc_rearmEntireVehicleSuccessLocal.sqf
Tuupertunut: Oh, I didn't know. What does it cause?

jonpas: To have extra spaces logged as well.

Tuupertunut: So is that not wanted? And what would it cause in other types of macro, like say
`EFUNC(interact_menu, createAction)`?

654wak654: That would turn into `ace_interact_menu_fnc_ createAction`. So probably an undefined variable error.

Tuupertunut: Oh I see. So macro behaviour in undefined when used with spaces. They were not designed with that in mind.

addons/rearm/functions/fnc_getAllTurrets.sqf
Dystopian: I think such `allTurrets` behavior is intentional. Driver is not a turret in fact. So I would either replace function with suggested methods or rename function, maybe `getAllTurretsWithDriver`.

addons/rearm/functions/fnc_rearmEntireVehicleSuccessLocal.sqf
jonpas: That's just how they work, for logging it's not a big problem, but it makes logs longer for no reason.

addons/rearm/functions/fnc_getAllTurrets.sqf
Tuupertunut: Driver turret may not be a turret from some other perspectives, but from a rearm perspective, driver turret is always a turret. That function is in the rearm module, and there is not a single case where the rearm module would want `allTurrets` without the driver turret.

Dystopian: ok, I have the last suggestion :) Instead of function you can use macro.

Tuupertunut: I don't like macros. 😦  They don't behave in the same elegant way as code should, as the space issue shows.

But yeah, even if I made a macro **or** used just pure code in the place of the function call, it wouldn't be simple. As @jonpas said, `[[-1]] + allTurrets _vehicle` is unnecessarily slow, so it would have to be `pushBack`. Problem is, the `pushBack` method takes al least 3 lines, as you have to first get the array and save it to a variable, then edit it with `pushBack` and then use it.

Dystopian: `[[-1]] append allTurrets _vehicle`

Tuupertunut: [`append`](https://community.bistudio.com/wiki/append) returns nothing.

jonpas: @Dystopian that wouldn't work because `append` modifies the array and doesn't return anything.

Just leave it as a separate function but rename it to be clearer what it does, `getAllRearmSeats` or something similar.

Tuupertunut: So if `getAllTurrets` is considered bad, it would probably be the @Dystopian's `getAllTurretsWithDriver` or `getAllRearmTurrets`.

Tuupertunut: Can't say fixed but... Changed.

addons/rearm/functions/fnc_addVehicleMagazinesToSupply.sqf
```diff
@@ -29,16 +29,17 @@ if (_vehicle isEqualType objNull) then {_vehicle = typeOf _vehicle};
 if (_vehicle == "") exitWith {
     ERROR_1("VehicleType [%1] is empty in ace_rearm_fnc_addVehicleMagazinesToSupply",_string);
 };
+private _turrets = [_vehicle] call FUNC(getAllRearmTurrets);
```
Dystopian: it was so interesting for me, so I decided to test ;)
on my computer `[[-1]] + allturrets t` is **twice faster** then function call.

Tuupertunut: Yeah, it's the function call itself that causes the performance overhead.

```
private _a = allTurrets veh;
_a pushBack [-1];
_a
```
is as fast as
`[[-1]] + allTurrets veh`

But since it's better to have a function anyway (cleaner code, used in many places, intended as replacement for `allTurrets`), it doesn't really matter what is inside it.

Dystopian: it looks strange...
> -Hey man, you can do it shorter
-I know but I want function
-You can do it faster
-I know but I want function
-You can do it even shorter and faster
-I know but I WANT FUNCTION

I don't know what to say else. I just disagree.

Tuupertunut: I still have a few arguments for function left in my pocket.
- Cleaner code. It hides implementation details. `[_veh] call FUNC(getAllRearmTurrets)` tells the reader immediately what the code does, whereas `[[-1]] + allTurrets veh` leaves him to wonder "What is that number -1 doing there?" and it requires a comment to explain what is happening.
- Replacement for `allTurrets`. There really should be a command for retrieving all turrets including the driver turret. I think most people would expect such a command to exist, because it is so common use case in Arma. It however doesn't, so this function is provided as a substitute for how the Arma API should have been.

Now you could argue that a macro would satisfy those needs as well, but I don't really like macros because they work with different rules than the rest of the code.

Dystopian: Nobody forces you to like macros. They work as documented. They are used everywhere in ACE, CBA and many other projects. Just because somewhere it is useful to use macros not functions.

jonpas: Ok guys that's enough. Separate function is fine, this doesn't run per-frame or that often at all.

Tuupertunut: 😜

SilentSpike: I feel like a macro is most suitable in this case, function is overkill. Though it's also not worth arguing over :)

Tuupertunut:
So, moving forward, I noticed a lot of rearm bugs, but also features that don't seem to work as intended. I'll be asking about them as I move on, but here's the first.

When there is limited ammo supply based on caliber, the ammo cost is defined as the caliber of a shell. For example, taking a magazine of 9mm rounds costs 9 ammo points and a magazine of 82mm rounds costs 82 points. **This does not take into account the number of shells there is in the magazine.** Currently a **30** round magazine of 7.62mm costs just as much as a **2000** round magazine of 7.62mm. It seems that the cost should be **per round**, not **per magazine** as it currently is.

Tuupertunut:
Any comment?

PabstMirror:
Sorry for the delay, this looks good
I'd like to merge this now to fix the partial magazine reloading problems.
And then maybe look the the supply problems in another PR.

To be honest, I really don't understand the caliber cost logic in removeMagazineFromSupply.
We shouldn't be afraid to just scrap it and rewrite a new cost calcuation method that makes sense.

jonpas:
> Empty pylon ammo selection (this feature) is no longer supported in favor of the upcoming "pylons" ACE-module (#5238). 

We might want to wait for pylons to actually be done.

Tuupertunut:
In the meantime, I could fix the ammo cost logic if I only knew how it's supposed to work.

My best guess is that it's meant to be **per-round** cost. Therefore:
- 30 round mag of 7.62mm would cost 30 * 7.62 = 228.6
- 2000 round mag of 7.62mm would cost 2000 * 7.62 = 15240
- 8 round mag of 82mm mortar would cost 8 * 82 = 656

In the current (broken) logic they cost 7.62, 7.62 and 82 points respectively.

Switching to the per-round logic would of course mean that the initial 2000 point supply in ammo trucks would need to be raised.

PabstMirror:
We could try to use volume as the basis for points.
One 82mm shell is going to be much more than 10 bullets

7.62 NATO (12x70) - 7.9 ml
81mm Shell (81x528) = 2720 ml

Weight should also be good. Either way it doesn't have to be perfect

I thinking we can just precompute these and stick them in the lookup table in scriptComponent

Tuupertunut:
Volume sounds good. So we need to know the volume of every ammunition in Arma. How do we do that? Is there some way to parse them straight from the game files?

jonpas:
I just want to confirm,
> Empty pylon ammo selection (this feature) is no longer supported in favor of the upcoming "pylons" ACE-module (#5238).

means that you can still rearm pylons, just not change them for other ammo types, correct?

Tuupertunut:
@jonpas Correct.

PabstMirror:
Unlimited/Entire vehicle are the default settings, so I'm going to merge this for 3.11
We can continue work in other PRs

PR #5414, opened 2017-08-09
Delete medical litter on building explosion

### LINKED ISSUE #5412: floating medical junk
On all versions without no mods and all mods the medical junk does not drop/destroy when a building is destroyed.
Medical junk missing physx

http://i.imgur.com/VjsKqbO.jpg

PabstMirror:
Based on commy's idea
Should fix #5412


SilentSpike:
Code looks good, will have to test in-game how reliable/effective it is

addons/medical/XEH_postInit.sqf
```diff
@@ -31,6 +31,20 @@ GVAR(heartBeatSounds_Slow) = ["ACE_heartbeat_slow_1", "ACE_heartbeat_slow_2"];
 if (isServer) then {
     ["ace_placedInBodyBag", FUNC(serverRemoveBody)] call CBA_fnc_addEventHandler;
     [QGVAR(createLitterServer), FUNC(handleCreateLitter)] call CBA_fnc_addEventHandler;
+    addMissionEventHandler ["BuildingChanged", {
+        if (isNil QGVAR(allCreatedLitter)) exitWith {};
+        params ["_buildingOld", "_buildingNew", "_isRuin"];
```
thojkooi: Shouldn't `_isRuin` need to be checked before cleaning up?

addons/medical/XEH_postInit.sqf
```diff
@@ -31,6 +31,20 @@ GVAR(heartBeatSounds_Slow) = ["ACE_heartbeat_slow_1", "ACE_heartbeat_slow_2"];
...
+        TRACE_3("BuildingChanged",_buildingOld,_buildingNew,_isRuin);
+        private _radius = ((0.75 * getNumber (configFile >> "CfgVehicles" >> (typeOf _buildingOld) >> "mapsize")) max 5) min 50;
```
thojkooi: What's the `50` based on? Seems like a rather large and specific number for some buildings.

addons/medical/XEH_postInit.sqf
```diff
@@ -31,6 +31,20 @@ GVAR(heartBeatSounds_Slow) = ["ACE_heartbeat_slow_1", "ACE_heartbeat_slow_2"];
...
+        TRACE_1("",_radius);
+        {
+            _x params ["", "_objects"];
+            if (({(_x distance2d _buildingOld) < _radius} count _objects) > 0) then {
```
thojkooi: This also cleans up anything that is not inside the building. Perhaps check for anything that's up in the air?

addons/medical/XEH_postInit.sqf
SilentSpike: I'd suggest we just use the largest bounding box dimension

addons/medical/XEH_postInit.sqf
SilentSpike: Collateral damage on valid medical litter isn't very important, but a quick height check isn't a bad idea 👍

addons/medical/XEH_postInit.sqf
commy2: sizeOf typeOf _buildingOld

addons/medical/XEH_postInit.sqf
commy2: `getPos select 2` is perfect for this. Reports the height above the first roadway LOD /or the terrain.

addons/medical/XEH_postInit.sqf
commy2: Some buildings might change appearance enough to make some things float before they are ruins.

PR #5415, opened 2017-08-09
Change "Remove Wheel" Action's Color Based on Damage

654wak654:
- Title.
- Sometimes I have to remove all 6 wheels of the Otokar Arma to find which 2 were effected by an RPG hit 😢.

Screenshots: http://imgur.com/a/N1SgY


addons/repair/functions/fnc_modifySelectionInteraction.sqf
```diff
@@ -0,0 +1,32 @@
+/*
+ * Author: 654wak654
+ * Modifies interaction color of vehicle selection to show it's current damage
```
jonpas: Missing final dot. `its current damage`.

654wak654:
> I am assuming you used same colors as for wheels themselves?
Yup, exact same color as the damage_X stuff

jonpas:
Do other repair points use similar levels of damage that this could be added to? I am thinking all points could use this in that case.

654wak654:
> Do other repair points use similar levels of damage that this could be added to? I am thinking all points could use this in that case.

Thought of the same thing but it's less useful with other hitpoints, since their action doesn't show up unless they're damaged anyway.

jonpas:
You are right, little point since they show up only when damaged, additionally their hitpoints are shown in when you sit in the vehicle.

PR #5418, opened 2017-08-11
Add fast mouse actions to refuel, add refuel eden attributes

### LINKED ISSUE #5173: Make Refuel capabilities dynamic
Entails:
- Dynamic interaction adding on init
- Ability to set any vehicle as ~~rearm~~/refuel vehicle easily (config and run-time)
- Proper API for 3rd party mods adding ~~rearm~~/refuel via simple config entries (currently have to copy all actions, which is just a mess)
- Follow Repair component way of doing it

Dystopian:
- title
 
- remove mouse wheel actions added with `addAction`;
- prevent nozzle connecting to static weapons and electric vehicles (e.g. darter, SDV);
- add new `isNotRefueling` interaction condition;
- allow refuel on ladder;
- fix AI unit can't run if player at its slot disconnected while nozzle carrying;
- fix #5173;
- add refuel Eden attributes.

Also:
- `fnc_takeNozzle` is rewritten (no code duplicating);
- some event handlers are replaced with PFH checking (no code dup and faster);
- some code is optimized;
- `DefaultAction` workaround is not used. Unit can't shoot with nozzle in hands, so I decided it will be easier to handle mouse clicking. Tell me if I'm wrong;
- take a look at recent [getCursorObjectParams command](https://community.bistudio.com/wiki/getCursorObjectParams) using, it allowed me to avoid some `*intersect*` code. Maybe it could be useful in other modules.

Here I have some points I would like (or be able) to implement but I need your opinion. At least I want you to say "yes", "no", "yes in another PR".
- [x] add `AnimChanged` handler, e.g. to drop nozzle when car pushes unit and he falls without dying;
- [x] replace `setVar [..., objNull]` with `setVar [..., nil]`. I think it's more logical to remove variable than set it to `objNull`;
- [x] unify variable names in all functions (`_unit`, `_nozzle`, `_source` (not `_target`), `_sink` look good to me);
- [x] remove ACE `Connect` action along with `fnc_connectNozzle`. ATM it runs another PFH with fast mouse actions, so it's not fully compatible with my PR. At least my way is faster and more obvious;
- ~move all refuel actions to `Interactions` menu not `Refuel` submenu. In most cases all I want from refuel objects (truck, station, nozzle) is using `Refuel` actions, and submenu is overkill~; 
- [x] remove all `reset` functions. I think it's not good to transfer problems from devs to users with such functions, it's better to fix issues;
- [x] replace static menu for refuel vehicles with only some GVARs (like existing `fuelCargo`) to simplify 3rd party mods support and allow mission maker to assign his own refueler;
- (next time) rework jerrycans, because attaching it to vehicle looks very odd;
- ~maybe rename `fnc_startNozzleInHandsPFH` (I have problems with names)~;
- ~maybe handle map/gear/etc opening and then RMB clicking~ - atm unit drops the nozzle on RMB clicking despite any dialog is opened;
- ~maybe implement `showHint` fast updating.~ ATM there is a small fade in on hint changing (0.5s), I think it's OK. In explosive attach code there is a piece which do it immediately with direct control updating;
- (someday) maybe implement digital display on fuel station :);
- [ ] inform CUP devs about changes.


Dystopian:
I assume there are no objections for all points.

jonpas:
> move all refuel actions to Interactions menu not Refuel submenu. In most cases all I want from refuel objects (truck, station, nozzle) is using Refuel actions, and submenu is overkill;

I don't think that would be too wise, especially not on fuel trucks where you have a few refuel actions (taking nozzle, checking amount left, etc).

Dystopian:
@jonpas maybe fuel objects/stations not trucks?

jonpas:
They are using same set of actions, splitting them up would cause more issues than solve problems IMO, especially since any objects might get more actions in the future.

addons/refuel/functions/fnc_handleRespawn.sqf
```diff
@@ -0,0 +1,27 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_unit"];
+TRACE_1("respawn",_unit);
+
+if (isNull (_unit getVariable [QGVAR(nozzle), objNull])) exitWith {};
+
+[_unit, "forceWalk", "ACE_refuel", false] call EFUNC(common,statusEffect_set);
```
PabstMirror: statusEffect should take care of respawns automatically
or was this explicitly needed?

Dystopian: it was in the original `XEH_respawn.sqf`, I didn't check. I will check now and delete if yes

jonpas:
@Dystopian Is this considered finished from your side? Also, are old interactions defined in config still compatible with this PR (thinking of 3rd party mods that use them for example) or will they have to update for compatibility?

PabstMirror:
I had an idea to prevent duplicate actions shown, will post something later today

Dystopian:
@jonpas it's done.
Old interactions aren't compatible atm. I thought nobody use ACE refuel configs except CUP (I'll inform them).
I can add additional check if @PabstMirror doesn't do it earlier.
Will be at my pc in 1 day.

jonpas:
It is likely some other 3rd party mod also uses it. I am sure CUP will update it ASAP, not worried about them. But if possible we should try to do backwards compatibility, and then remove it after 2 versions as is standard with ACE3.

Dystopian:
@jonpas done

jonpas:
That just prints deprecation message to RPT, but is the old interaction menu still compatible with the rest of the rework done in this PR? There is no point saying it's deprecated and will be removed in 3.13.0 if it doesn't work with 3.11.0 in the first place.

jonpas:
Clarified on Slack, old interaction menu still works, so backwards compatibility is there. 👍

jonpas:
What @PabstMirror said, also dynamic actions! 👍 

Would be great if we can get framework documentation for that up as well. [This](https://ace3mod.com/wiki/framework/refuel-framework.html) is what we have currently. Then that can just be linked to CUP and whoever else uses it.

Issue #1335, opened 2015-05-21
Break out disposables to a framework for addon makers to use


jaynus:
@headedhorseman101 commented 4 hours ago
Would it be possible for the addition of a template or a function that would allow addon makers to make their launchers disposable (i.e. ace_missile_guidance)? The disposable module in its current state works really well for the NLAW , but doing this would allow addon makers to, as an example, make a javelin that you had to assemble and has a disposable tube, kinda like in ACE 2.

sargken:
Like the RHS M136 it could use the functionality

PabstMirror:
Handled in CBA
https://github.com/CBATeam/CBA_A3/pull/1116

PR #5422, opened 2017-08-12
Disposable module expansion

CoxLedere:
### Changes
- Preloaded Missiles can be stored in any inventory slot (Uniform, Vest, Backpack) and have 0 mass to prevent issues of not having inventory space for a missile that should be loaded into a launcher already.
- modelSpecial changed to emptry string as it was turning the model of any weapon loaded with a Preloaded Missile into an NLAW.
- NLAW and Used NLAW now have their models set to a loaded and unloaded NLAW respectively since modelSpecial has been changed.
- Changed NLAW and Used NLAW weight to reflect the NLAW missile's weight being set to 0.
- Added in "preparation" handling. This emulates how RHS has setup their M136 and M72 where the launcher must be "reloaded" before it can be fired.
- Players and AI now "throw" launchers off their back. In essence, what was originally a GroundWeaponHolder has been changed to a WeaponHolderSimulated, and applied to Players as well as AI.
- Removal of Preloaded Missiles is now done dynamically based on whether the missile is a child of ACE_PreloadedMissileDummy_Base instead of using the static type check for ACE_PreloadedMissileDummy and ACE_FiredMissileDummy.
- RHS USAF and AFRF compatibility updated to use this new framework for the disposable launchers RHS has. RHS's disposable system is also disabled with these changes. (GREF has no launchers)
- Documentation updated to reflect the new changes.

### Minor Changes
- Removal of Preloaded Missiles moved from updateInventoryDisplay.sqf to takeLoadedATWeapon.sqf

### Discussion point
This pull was originally intended to work off of CA_LauncherMagazine in CfgMagazines. However, this caused issues with ace_nlaw, specifically, the attack type could not be selected. While the current solution has been to turn ACE_PreloadedMissileDummy_Base into a child of NLAW_F and essentially replicate CA_LauncherMagazine for ACE_PreloadedMissileDummy_Base. While this does work in it's current state, in my entirely personal opinion, I believe this is far from ideal and would _prefer_ that ace_disposable be made a dependency for ace_nlaw.

But, far be it from me to tell you guys how to setup dependencies, thus, I welcome your opinions on this matter.


addons/disposable/XEH_postInit.sqf
```diff
@@ -9,6 +9,11 @@ if (!hasInterface) exitWith {};
     [_unit] call FUNC(updateInventoryDisplay);
 }] call CBA_fnc_addPlayerEventHandler;
 
+["weapon", {
+    params ["_unit","_weapon"];
```
Dystopian: space here and anywhere between arguments

addons/disposable/functions/fnc_handlePreparation.sqf
```diff
@@ -0,0 +1,30 @@
...
+ * [fromWeaponEH] call ace_disposable_fnc_handlePreparation;
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_unit","_weapon"];
+
+//If it's an AI, if our current weapon is not a launcher, if our current launcher is not a hasPreparation launcher, or if our launcher is actually a used launcher, exit.
+if (!local _unit || !([_unit] call EFUNC(common,isPlayer)) || {_weapon != secondaryWeapon _unit} || {getNumber (configFile >> "CfgWeapons" >> _weapon >> QGVAR(hasPreparation)) != 1} || {getNumber (configFile >> "CfgWeapons" >> _weapon >> "ACE_isUsedLauncher") == 1})  exitWith {};
```
Dystopian: lazy eval on the 2nd condition; maybe use `_config` var instead of `configFile >> "CfgWeapons" >> _weapon`

addons/disposable/functions/fnc_replaceATWeapon.sqf
```diff
@@ -34,33 +34,24 @@ _unit selectWeapon _replacementTube;
...
-            _unit removeWeaponGlobal _tube;
-        };
-    }, 1, [_unit, _replacementTube, _projectile]] call CBA_fnc_addPerFrameHandler;
-};
+[{
+    params ["_args","_idPFH"];
+    _args params ["_unit", "_replacementTube", "_projectile"];
+
+    //We want to wait for the missile to explode if it's an AI. If it's a player we want to wait until they switch weapons.
+    if ((!([_unit] call EFUNC(common,isPlayer)) && isNull _projectile) || ([_unit] call EFUNC(common,isPlayer) && currentWeapon _unit != _replacementTube)) then {
```
Dystopian: lazy

addons/disposable/functions/fnc_takeLoadedATWeapon.sqf
```diff
@@ -20,15 +20,25 @@ TRACE_1("params",_unit);
 
 if (!local _unit) exitWith {};
 
+//Remove any preloaded missiles, unless we're currently wielding a hasPreparation launcher
+if(getNumber (configFile >> "CfgWeapons" >> currentWeapon _unit >> QGVAR(hasPreparation)) != 1) then {
```
Dystopian: space after if

addons/disposable/functions/fnc_takeLoadedATWeapon.sqf
```diff
@@ -20,15 +20,25 @@ TRACE_1("params",_unit);
...
+            _unit removeMagazine _x;
+        };
+    } forEach magazines _unit;
+};
+
 private _launcher = secondaryWeapon _unit;
 private _config = configFile >> "CfgWeapons" >> _launcher;
 
+//Players with hasPreparation launchers stop here, but AI need to have ammo for their launchers in order to select them as weapons.
+if ([_unit] call EFUNC(common,isPlayer) && {getNumber (_config >> QGVAR(hasPreparation)) == 1}) exitWith{};
```
Dystopian: space after exitWith

addons/disposable/XEH_postInit.sqf
```diff
@@ -9,6 +9,11 @@ if (!hasInterface) exitWith {};
...
+    [_unit,_weapon] call FUNC(handlePreparation);
```
Dystopian: no need to use param, replace `{code}` with `LINKFUNC(handlePreparation)`

addons/disposable/functions/fnc_handlePreparation.sqf
CoxLedere: Using a _config var would require evaluation of the configFile for every weapon that is passed through, and according to CBA's documentation the weaponEH is called everytime a weapon is switched. I'm fine with using a _config var, I just felt that this method would be faster in the long term.

Dystopian: could split condition to 2 lines, the 1st checks unit and weapon, the 2nd uses `_config`

addons/disposable/XEH_postInit.sqf
SilentSpike: No need to use `LINKFUNC`, use `FUNC`

SilentSpike:
Interesting changes, I'll have to give this a test as I'm not familiar enough with the disposable module to have an opinion from looking at the code alone :)

addons/disposable/XEH_postInit.sqf
CoxLedere: So what exactly is the difference between LINKFUNC and FUNC then?

addons/disposable/functions/fnc_takeLoadedATWeapon.sqf
```diff
@@ -16,19 +16,29 @@
 #include "script_component.hpp"
 
 params ["_unit"];
-TRACE_1("params",_unit);
+TRACE_1("params", _unit);
```
Dystopian: macros don't use spaces, here and anywhere

addons/disposable/XEH_postInit.sqf
jonpas: [`LINKFUNC`](https://github.com/acemod/ACE3/blob/master/addons/main/script_debug.hpp#L7-L15) is used so you can recompile those functions on the fly when developing.

addons/disposable/functions/fnc_takeLoadedATWeapon.sqf
CoxLedere: Okay, is there any documentation on how spaces work? Because the only reference [here](https://ace3mod.com/wiki/development/coding-guidelines.html) is to have 4 spaces instead of tabs.

Dystopian: macros are not commands, they are for preprocessor. Just don't use spaces if it's not intended

jonpas: Spaces in macros will cause them to be logged as well, adding additional spaces in the logs (you'd get double spaces).

addons/disposable/XEH_postInit.sqf
SilentSpike: `LINKFUNC` does actually make sense here, I misread our wiki 😆

CoxLedere: Is it fine as of the most recent commit, or should I change it?

Dystopian: no real need

CoxLedere:
Stumbled onto #1335, this pull should close that.

jonpas:
Moving to 3.12 as we still have to go over it and don't want to rush such a big change.

optionals/compat_rhs_usf3/CfgWeapons.hpp
```diff
@@ -1,8 +1,10 @@
 
+class Eventhandlers;
 class CfgWeapons {
     class Pistol_Base_F;
     class Rifle_Base_F;
     class srifle_EBR_F;
+    class Launcher_Base_F;
```
Dystopian: this is already here after #5589, need to merge master and fix

SilentSpike:
Finally got around to testing this, ran into this error due to the config changes:



Also here is how the launcher dropping looks from 3PP just for reference (obviously a little janky, but I personally think it's okay):

[](https://giant.gfycat.com/SoggyHideousCanadagoose.webm)

Pushing the milestone back to allow time to discuss and resolve.

CoxLedere:
@SilentSpike by the "janky" part, I assume you mean the AI reloading the launchers and then dropping them?

I made a slight change that might fix it, but I've never seen that behavior before. This is what I'm currently seeing as of the most recent commit with default NATO Rifleman (AT) units:

https://giant.gfycat.com/DiscretePhysicalGeese.webm

SilentSpike:
Yeah the little invisible reload was what I was referring to, what you're seeing there looks a lot better (since the invisible holstering is very quick).

thojkooi:
Most of the functionality should already be in ACE - closing.

jonpas:
Anything missing, please open new PR.

Dystopian:
> Most of the functionality should already be in ACE - closing.

Could you tell which ones exactly are in ACE?

> Anything missing, please open new PR.

Everything?

CoxLedere:
I have absolutely no idea why this was closed, other than just not wanting to merge this for some nebulous reason.

This PR has been open for 19 months, been pushed back from milestones three times, and the last time it seems like an ACE team member even looked at this was 15 months ago.

As far as functionality that is included in this and is missing from ACE proper, just five posts up from this comment is [this](https://giant.gfycat.com/DiscretePhysicalGeese.webm) link to a video of a section of this.

As far as other parts of this:

- Preloaded missiles can be stored in a backpack; which isn't a problem because this PR didn't get merged.
- Preloaded missiles still have mass; which isn't a problem because this PR didn't get merged
- The entire disposable "framework" is still built almost entirely around the NLAW requiring quite a bit of work to add other launchers to it. You'd have to do about the half the work I've done in this PR, hence why I created this PR to begin with.
- I mean, just pick a bullet point from the PR, and it's not included in the disposable framework.

addons/disposable/CfgWeapons.hpp
```diff
@@ -1,8 +1,13 @@
+class WeaponSlotsInfo;
 class CfgWeapons {
     class Launcher_Base_F;
     class launch_NLAW_F: Launcher_Base_F {
         ACE_UsedTube = "ACE_launch_NLAW_Used_F";      // The class name of the used tube.
-        magazines[] = {"ACE_PreloadedMissileDummy"};  // The dummy magazine
+        magazines[] = {"ACE_PreloadedMissileDummy_NLAW"};  // The dummy magazine
+        model = "\A3\weapons_f\launchers\nlaw\nlaw_loaded_F";
+        class WeaponSlotsInfo : WeaponSlotsInfo{
```
commy2: Formatting broken.

addons/disposable/CfgWeapons.hpp
```diff
@@ -12,8 +17,11 @@ class CfgWeapons {
         displayName = CSTRING(UsedTube);
         descriptionShort = CSTRING(UsedTubeDescription);
         magazines[] = {"ACE_FiredMissileDummy"};  // This will disable the used launcher class from being fired again.
+        model = "\A3\weapons_F\launchers\nlaw\nlaw_F.p3d";
         //picture = "";              @todo
-        //model = "";                @todo
         weaponPoolAvailable = 0;
+        class WeaponSlotsInfo : WeaponSlotsInfo{
```
commy2: Formatting broken.

addons/disposable/functions/fnc_handlePreparation.sqf
```diff
@@ -0,0 +1,33 @@
...
+private _config = configFile >> "CfgWeapons" >> _weapon;
+if (getNumber (_config >> QGVAR(hasPreparation)) != 1 || {getNumber (_config >> "ACE_isUsedLauncher") == 1}) exitWith {};
+
+private _magazine = getArray (_config >> "magazines") select 0;
+_unit removeSecondaryWeaponItem _magazine;
+_unit addMagazine _magazine;
+
+[{currentWeapon (_this select 0) != (_this select 1)},
+    {(_this select 0) removeMagazines (_this select 2)}, //Remove the magazine if the player doesn't load the launcher so it's not just hanging around in their inventory.
+[_unit, _weapon, _magazine]] call CBA_fnc_waitUntilAndExecute;
```
commy2: Why ask this every frame when this function itself uses the exact event for this purpose?

addons/disposable/functions/fnc_handlePreparation.sqf
```diff
@@ -0,0 +1,33 @@
...
+ * Example:
+ * [fromWeaponEH] call ace_disposable_fnc_handlePreparation;
+ *
+ * Public: No
+ */
+
+params ["_unit", "_weapon"];
+
+//If it's an AI, or if our current weapon is not a launcher, exit.
+if (!local _unit || {!([_unit] call EFUNC(common,isPlayer))} || {_weapon != secondaryWeapon _unit}) exitWith {};
```
commy2: Local and player checks are pointless, since this is a player event handler.

addons/disposable/functions/fnc_replaceATWeapon.sqf
```diff
@@ -34,33 +34,24 @@ _unit selectWeapon _replacementTube;
...
+        //If (tube is dropped) OR (is dead) just exit
+        if (secondaryWeapon _unit != _replacementTube || {!alive _unit}) exitWith {};
+
+        //Remove the weapon and "throw" it on the ground.
+        _unit removeWeaponGlobal _replacementTube;
+        private _container = createVehicle ["WeaponHolderSimulated", _unit modelToWorld ((_unit selectionPosition ["rightshoulder", "Memory"]) vectorAdd [0, 0.2, 0.1]), [], 0, "CAN_COLLIDE"];
+        _container addWeaponCargoGlobal [_replacementTube, 1];
+        _container setDir (getDir _unit - 90);
+        //We get the current velocity of the soldier, then apply it to the launcher. Position 0 and 1 in the array are reversed because the model is rotated by 90 degrees.
+        _container setVelocityModelSpace [(velocityModelSpace _unit select 1) + 0.2, (velocityModelSpace _unit select 0) - 1.5, velocityModelSpace _unit select 2];
```
commy2: Should use `vectorAdd`.

addons/disposable/functions/fnc_replaceATWeapon.sqf
```diff
@@ -34,33 +34,24 @@ _unit selectWeapon _replacementTube;
...
+    };
+}, 0, [_unit, _replacementTube, _projectile]] call CBA_fnc_addPerFrameHandler;
```
commy2: Every frame?

optionals/compat_rhs_afrf3/CfgMagazines.hpp
```diff
@@ -14,4 +16,14 @@ class cfgMagazines {
     class rhs_mag_127x108mm_1470 : rhs_mag_127x108mm_50 {
         ace_isbelt = 0;
     };
+    class ACE_PreloadedMissileDummy_RPG26: ACE_PreloadedMissileDummy_Base
+    {
```
commy2: Broken formatting.

optionals/compat_rhs_afrf3/CfgMagazines.hpp
```diff
@@ -14,4 +16,14 @@ class cfgMagazines {
...
+        ammo = "rhs_rpg26_rocket";
+        type = "6 * 256";
+        initSpeed = 149;
+    };
+    class ACE_PreloadedMissileDummy_RSHG2: ACE_PreloadedMissileDummy_RPG26
+    {
```
commy2: Broken formatting.

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -1,4 +1,5 @@
-
+class WeaponSlotsInfo;
```
commy2: This class does not exist. This breaks inheritance of pretty much every weapon.

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -1,4 +1,5 @@
...
+class Eventhandlers;
```
commy2: Same as above, except for vehicles and weapons.

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -113,6 +114,39 @@ class CfgWeapons {
     class rhs_weap_rpg7: Launcher_Base_F {
         ace_reloadlaunchers_enabled = 1;
     };
+    class rhs_weap_rpg26: Launcher_Base_F {
+        ACE_UsedTube = "rhs_weap_rpg26_used";
+        magazines[] = {"ACE_PreloadedMissileDummy_RPG26"};
+        class WeaponSlotsInfo : WeaponSlotsInfo {
```
commy2: Broken formatting.

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -113,6 +114,39 @@ class CfgWeapons {
...
+        class Eventhandlers: Eventhandlers {
+            class RHS_DisposableWeapon {
+                fired = "";
+            };
+        };
+    };
+    class rhs_weap_rshg2: rhs_weap_rpg26 {
+        ACE_UsedTube = "rhs_weap_rshg2_used";
+        magazines[] = {"ACE_PreloadedMissileDummy_RSHG2"};
+        class WeaponSlotsInfo : WeaponSlotsInfo {
```
commy2: Broken formatting.

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -113,6 +114,39 @@ class CfgWeapons {
...
+            mass = 90.6;
+        };
+    };
+    class rhs_weap_rpg26_used: rhs_weap_rpg26 {
+        ACE_isUsedLauncher = 1;
+        magazines[] = {"ACE_FiredMissileDummy"};
+        class WeaponSlotsInfo : WeaponSlotsInfo {
```
commy2: Broken formatting.

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -113,6 +114,39 @@ class CfgWeapons {
...
+            mass = 57.2;
+        };
+    };
+    class rhs_weap_rshg2_used: rhs_weap_rshg2 {
+        ACE_isUsedLauncher = 1;
+        magazines[] = {"ACE_FiredMissileDummy"};
+        class WeaponSlotsInfo : WeaponSlotsInfo {
```
commy2: Broken formatting.

optionals/compat_rhs_usf3/CfgWeapons.hpp
```diff
@@ -1,4 +1,5 @@
 
+class Eventhandlers;
```
commy2: See above.

optionals/compat_rhs_usf3/CfgWeapons.hpp
```diff
@@ -220,7 +221,39 @@ class CfgWeapons {
         lockingTargetSound[] = {"",0,1};
         lockedTargetSound[] = {"",0,1};
     };
+    class rhs_weap_M136 : Launcher_Base_F {
```
commy2: Broken formatting.

optionals/compat_rhs_usf3/CfgWeapons.hpp
```diff
@@ -220,7 +221,39 @@ class CfgWeapons {
...
+        ACE_UsedTube = "rhs_weap_M136_used";
+        EGVAR(disposable,hasPreparation) = 1;
+        magazines[] = {"ACE_PreloadedMissileDummy_M136HEAT"};
+        class Eventhandlers: Eventhandlers {
+            class RHS_DisposableWeapon {
+                fired = "";
+            };
+        };
+    };
+    class rhs_weap_M136_hedp:  rhs_weap_M136 {
```
commy2: White space.

commy2:
The config inheritance is completely broken, but I can't be arsed to download RHS to fix it.

commy2:
>Preloaded Missiles can be stored in any inventory slot (Uniform, Vest, Backpack) and have 0 mass to prevent issues of not having inventory space for a missile that should be loaded into a launcher already.

This change is pointless. You need to add a backpack anyway if you have no uniform vest and backpack. So either you have a backpack, then you can always add the mass 0 mags, or you don't, in which case you just add a temporary dummy backpack.

CoxLedere:
> > Preloaded Missiles can be stored in any inventory slot (Uniform, Vest, Backpack) and have 0 mass to prevent issues of not having inventory space for a missile that should be loaded into a launcher already.
> 
> This change is pointless. You need to add a backpack anyway if you have no uniform vest and backpack. So either you have a backpack, then you can always add the mass 0 mags, or you don't, in which case you just add a temporary dummy backpack.

I've been messing around with giving the player the dummy backpack, and it's not working out very well.

1. The ACE_FakeBackpack class has an actual model, making it visible to player the player, and it shows up in their inventory, and I don't see any other backpacks that are invisible.
2. This isn't like the take loaded weapon script, where we're adding the missile to the backpack, spawning then immediately removing it, we need to wait for the player to consciously reload the weapon, meaning they could have this fake backpack on for an indefinite amount of time. Which means we may as well just give the player a free backpack.

commy2:
I don't mind reopening it. If you have a problem with the dummy backpacks model, then why not replace it? You need this backpack no matter what, because the launchers have to work even when wearing any uniform, vest or backpack, so the change is still absolutely pointless.

CoxLedere:
But the main problem, as I stated above, is that we're giving the player a free backpack, that they can potentially use indefinitely. Because, as long as the launcher is equipped, we need the backpack to remain on them.

Honestly, if we really need to have guys in their skivvies be able to use these launchers, I'd much prefer to just add the missile directly to the launcher.

Which somewhat removes the point, but it's less open to abuse and/or problems than giving the player a backpack to wear forever.

commy2:
The dummy backpack gets removed immediately. It only gets add for adding the launcher magazine.

CoxLedere:
No it doesn't! The missile needs to in the player's inventory until the player manually reloads the launcher themselves.

To explain the full process with a hasPreparation launcher:
1. Player equips launcher
2. The missile class to load the launcher is added
3. The player reloads the launcher, this plays the preparation animation
4. The missile is now moved from the players inventory into the launcher, and we can safely delete it.

The problem here is step 3. The player can do this at any time they want to. So if the player suddenly discovers that he gets a free backpack when he pulls out his launcher, he might not be so keen on surrendering it, and now there's _a lot_ of code that would need to be added to prevent any case of abuse of this.

654wak654:
Couldn't this be solved by removing that feature? Needing a reload to "prepare" a launcher is one of the most annoying RHS hacks anway.

CoxLedere:
I mean, it can always be removed. But I'd really rather just shortcut the process if the missile can't be added normally.

commy2:
>Player equips launcher

check

>The missile class to load the launcher is added

This fails, because you are not wearing any uniform, vest or backpack to hold the magazine.

>The player reloads the launcher, this plays the preparation animation

Can't reload, because no magazine :(

>The missile is now moved from the players inventory into the launcher, and we can safely delete the backpack.

nope

commy2:
>Couldn't this be solved by removing that feature? Needing a reload to "prepare" a launcher is one of the most annoying RHS hacks anway.

Agreed. It's trash.

CoxLedere:
I've found a better method to add the magazine when the player tries to reload. I'm cleaning it all up currently, but I'm able to add the backpack now, so this shouldn't be a problem.

And yeah, I probably should've just created this method when I first opened this...

commy2:
There is no better method than a dummy backpack. addWeaponItem and addSecondaryWeaponItem are bugged.

CoxLedere:
> ... but I'm able to add the backpack now, so this shouldn't be a problem.

I am adding the backpack, I just found a significantly better place of where it should be added so it's used on demand.

CoxLedere:
Okay, everything @commy2 brought up should now be addressed.

Can we re-open this PR now?

commy2:
I don't like any of this, sorry. I will be porting the BWA3 system to CBA and subsequently make use of that in ACE (ironing out all the bugs that accumulated in ACE since BI added stuff like modelSpecial).

I see no reason to touch RHS.

I don't like using a physx weapon holder for this and I likewise don't like players throwing the weapon away on their own (at least not without setting).

Your changes to the magazines like mass and container compatibility are born out of ignorance how this stuff works. It is admittedly strange and poorly documented.

addons/disposable/functions/fnc_replaceATWeapon.sqf
```diff
@@ -34,33 +34,24 @@ _unit selectWeapon _replacementTube;
...
+    if (currentWeapon _unit != _replacementTube) then {
+        //Remove PFEH:
+        [_idPFH] call CBA_fnc_removePerFrameHandler;
+
+        //If (tube is dropped) OR (is dead) just exit
+        if (secondaryWeapon _unit != _replacementTube || {!alive _unit}) exitWith {};
+
+        //Remove the weapon and "throw" it on the ground.
+        _unit removeWeaponGlobal _replacementTube;
+        private _container = createVehicle ["WeaponHolderSimulated", _unit modelToWorld ((_unit selectionPosition ["rightshoulder", "Memory"]) vectorAdd [0, 0.2, 0.1]), [], 0, "CAN_COLLIDE"];
```
commy2: `createVehicle` needs ATL pos (except for boat or amphibious vehicles, where it needs AGL), but m2w reports AGL.

This breaks on aircraft carriers.

CoxLedere: @commy2 I appreciate the code review, but now I'm confused as to the status of this.

1. The PR itself is still closed, so the code here is stale by at least two weeks as can be seen [here](https://github.com/CoxLedere/ACE3/commits/DisposableExpansion)
2. Your comment from [here](https://github.com/acemod/ACE3/pull/5422#issuecomment-482780446) seems incredibly damning on this ever getting into ACE.

Ultimately, I appreciate you pointing out the problem here, I'm just confused why you're spending time on something that is effectively "dead"

commy2: I am doing something similar in CBA and these are notes for me.

PR #5426, opened 2017-08-13
Refuel: Turn on nozzle after connection

654wak654:
- Automatically start fueling after the "connect nozzle" action.
- Remove the 2 second timer for the "turn on", "turn off", and "read fuel counter" actions. IMO The 2 second delay "simulated" by the progress bar is already simulated when the player goes through the action menu, and doesn't contribute anything to gameplay.


jonpas:
Some strings can be cleaned up as well.

addons/refuel/functions/fnc_connectNozzleAction.sqf
```diff
@@ -128,6 +128,7 @@ private _attachPosModel = _target worldToModel (ASLtoAGL _bestPosASL);
         };
 
         [_unit, _target, _nozzle, _endPosTestOffset] call FUNC(refuel);
```
PabstMirror: Add the following line here:
`_unit setVariable [QGVAR(tempFuel), nil];`

Before there was a pause before connecting and refueling when this var would be reset.
Without clearing it, it will use the tempFuel from the last fueled vic (which would probably be 1), so it would instantly fill vics.

ref https://github.com/acemod/ACE3/blob/master/addons/refuel/functions/fnc_refuel.sqf#L91

PabstMirror:
Probably best to hold off on this until we merge #5418

addons/refuel/functions/fnc_connectNozzleAction.sqf
```diff
@@ -128,6 +128,8 @@ private _attachPosModel = _target worldToModel (ASLtoAGL _bestPosASL);
...
+        _unit setVariable [QGVAR(tempFuel), nil];
+        [_unit, _nozzle] call FUNC(turnOn);
```
Dystopian: Here should be `canTurnOn` check before `turnOn` call like it is in interaction menu.

654wak654: Should also communicate to the player why the muzzle didn't turn on too then.

Dystopian: Then maybe it's easier to make sure that `FUNC(refuel)` manages all situations checked in `canTurnOn` and not call it :)

addons/refuel/functions/fnc_readFuelCounter.sqf
```diff
@@ -18,24 +18,11 @@
...
-    localize LSTRING(CheckFuelCounterAction),
-    {true},
-    [INTERACT_EXCEPTIONS]
-] call EFUNC(common,progressBar);
+private _currentFuel = [_source] call FUNC(getFuel);
+if (_currentFuel == REFUEL_INFINITE_FUEL) then {
+    private _fuelCounter = 0.01 * round (100 * (_source getVariable [QGVAR(fuelCounter), 0]));
+    [[LSTRING(Hint_FuelCounter), _fuelCounter], 1.5, _unit] call EFUNC(common,displayTextStructured);
+} else {
+    private _fuelCounter = 0.01 * round (100 * ((_source getVariable [QGVAR(fuelCounter), _currentFuel]) - _currentFuel));
```
jonpas: This is the only different thing in the entire statement. I suggest writing it as:
```sqf
private _currentFuel = [_source] call FUNC(getFuel);
private _fuelCounter = if (_currentFuel == REFUEL_INFINITE_FUEL) then {
    _source getVariable [QGVAR(fuelCounter), 0]
} else {
    (_source getVariable [QGVAR(fuelCounter), _currentFuel]) - _currentFuel
};

private _fuelCounter = 0.01 * round (100 * _fuelCounter);
[[LSTRING(Hint_FuelCounter), _fuelCounter], 1.5, _unit] call EFUNC(common,displayTextStructured);
```

PR #5431, opened 2017-08-14
Remove locations tab from spectator

SilentSpike:
- Doesn't add much value, adds complexity and the implementation is half
baked. Would rather add back in at a later date (if at all) with a
better implementation.
- I have an idea to replace the locations tab with a meta tab so users
can toggle things like projectile drawing via the UI and are not forced
to use a hotkey. Might also be a good place to display the extended controls.


AACO:
Yeah, I had pretty much the exact same experience. Makes sense for end game where it'll have very key defined locations, but doesn't make a ton of sense for a more open ended spectator. Only way I considered having it work was a module where you could just plop it down and have it freestanding, or sync it to an object to automatically add it to the list. (But didn't want to put the effort into it as it would be underutilized in my setting)

addons/spectator/ui.hpp
```diff
@@ -84,9 +84,9 @@ class GVAR(display) {
 
             fade = 0.8;
             rows = 1;
-            columns = 2;
-            strings[] = {"$STR_A3_Spectator_Entities","$STR_A3_Spectator_Locations"};
-            values[] = {0,1};
+            columns = 1;
+            strings[] = {"$STR_A3_Spectator_Entities"};
+            values[] = {0};
```
AACO: Should 0 be replaced with LIST_ENTITIES for consistency? (It doesn't matter for the current state, but if you introduce the meta tab might help)

SilentSpike: Yes it probably should, good catch!

addons/spectator/functions/fnc_setSpectator.sqf
```diff
@@ -109,8 +109,8 @@ if (alive player) then {
     [player, _hidden, QGVAR(isSet), side group player] call EFUNC(common,switchToGroupSide);
 
     // Ghosts can't talk
-    [_hidden, QGVAR(isSet)] call EFUNC(common,hideUnit);
-    [_hidden, QGVAR(isSet)] call EFUNC(common,muteUnit);
+    [player, QGVAR(isSet)] call EFUNC(common,hideUnit);
```
Neviothr: Not `ACE_player`?

Cuel: it's the player that goes into spectator, not the currently controlled unit

addons/spectator/script_component.hpp
```diff
@@ -50,8 +50,7 @@
 #define DISTANCE_NAMES_SQR 30625 // Names are rendered within 175m, squared for `distanceSqr` speed
 #define NAME_MAX_CHARACTERS 17
 
-#define LIST_ENTITIES localize "STR_A3_Spectator_Entities"
-#define LIST_LOCATIONS localize "STR_A3_Spectator_Locations"
+#define LIST_ENTITIES 0
```
bux: no string needed here anymore?

SilentSpike: Nah, it was only ever used to check the current tab (and localised for some reason). So I may as well just use the tab indices.

PR #5432, opened 2017-08-14
Cargo: Add removeCargoItem Function

654wak654:
- Adds a function called `ace_cargo_fnc_removeCargoItem` to do the opposite of what `ace_cargo_fnc_addCargoItem` does.
- Adds appropriate event and documentation.

### DIFF
--- a/addons/cargo/functions/fnc_removeCargoItem.sqf
+++ b/addons/cargo/functions/fnc_removeCargoItem.sqf
@@ -11,7 +11,8 @@
  * None
  *
  * Example:
- * ["item", vehicle] call ace_cargo_fnc_removeCargoItem
+ * ["ACE_Wheel", vehicle, 2] call ace_cargo_fnc_removeCargoItem
+ * [crate_7, truck] call ace_cargo_fnc_removeCargoItem
  *
  * Public: Yes
  */
@@ -28,7 +29,7 @@ private _addedSpace = 0;
 {
     if (_itemsRemoved == _amount) exitWith {};
 
-    if (_x isEqualTo _item || {_x isEqualType objNull && {typeOf _x == _item}}) then {
+    if (_x isEqualTo _item || {_x isEqualType objNull && {typeOf _x isEqualTo _item}}) then {
         INC(_itemsRemoved);
         ADD(_addedSpace,[_x] call FUNC(getSizeItem));
 
@@ -44,6 +45,11 @@ _vehicle setVariable [QGVAR(loaded), _loaded, true];
 
 private _space = [_vehicle] call FUNC(getCargoSpaceLeft);
 _vehicle setVariable [QGVAR(space), _space + _addedSpace, true];
+
+// No point in returning a deleted object in the event
+if (_item isEqualType objNull) then {
+    _item = typeOf _item;
+};
 
 // Invoke listenable event
 ["ace_cargoRemoved", [_item, _vehicle, _amount]] call CBA_fnc_globalEvent;
--- a/addons/cargo/functions/fnc_removeCargoItem.sqf
+++ b/addons/cargo/functions/fnc_removeCargoItem.sqf
@@ -25,16 +25,17 @@ private _loaded = _vehicle getVariable [QGVAR(loaded), []];
 
 private _addedSpace = 0;
 private _itemClass = _item;
+private _itemsRemoved = 0;
 
 private _continue = if (_item isEqualType objNull) then {
-    if (_item in _loaded) exitWith {false};
+    if !(_item in _loaded) exitWith {false};
     _addedSpace = [_item] call FUNC(getSizeItem);
     _loaded deleteAt (_loaded find _item);
     _itemClass = typeOf _item;
     deleteVehicle _item;
+    _itemsRemoved = 1;
     true
 } else {
-    private _itemsRemoved = 0;
     {
         if (_itemsRemoved == _amount) exitWith {};
 
@@ -44,7 +45,7 @@ private _continue = if (_item isEqualType objNull) then {
         ) then {
             INC(_itemsRemoved);
             ADD(_addedSpace,[_x] call FUNC(getSizeItem));
-    
+
             if (_x isEqualType objNull) then {
                 deleteVehicle _x;
             };
--- a/addons/cargo/functions/fnc_removeCargoItem.sqf
+++ b/addons/cargo/functions/fnc_removeCargoItem.sqf
@@ -1,0 +1,48 @@
+/*
+ * Author: 654wak654
+ * Reamoves a cargo item from the vehicle.
+ *
+ * Arguments:
+ * 0: Item Classname <STRING>
+ * 1: Vehicle <OBJECT>
+ * 2: Amount <NUMBER> (default: 1)
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * ["item", vehicle] call ace_cargo_fnc_removeCargoItem
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_itemClass", "_vehicle", ["_amount", 1]];
+TRACE_3("params",_itemClass,_vehicle,_amount);
+
+private _loaded = _vehicle getVariable [QGVAR(loaded), []];
+
+private _itemsToRemove = [];
+private _addedSpace = 0;
+
+{
+    if (count _itemsToRemove == _amount) exitWith {};
+
+    if (_x == _itemClass || {typeOf _x == _itemClass}) {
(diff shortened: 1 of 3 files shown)


addons/cargo/functions/fnc_removeCargoItem.sqf
```diff
@@ -0,0 +1,48 @@
+/*
+ * Author: 654wak654
+ * Reamoves a cargo item from the vehicle.
```
jonpas: That spelling! :D

SilentSpike:
Code all looks good - though I wonder whether the function should also be able to accept a specific object as well as a classname

docs/wiki/framework/cargo-framework.md
```diff
@@ -141,3 +142,20 @@ Note first arg can be a in-game object or a classname of an object type.
  * Example:
  * [object, vehicle] call ace_cargo_fnc_unloadItem
 ```
+
+### 4.6 Remove/ Delete cargo from vehicle 
+
```
SilentSpike: Maybe we should add notes to documentation that doesn't apply to the public build yet?

654wak654: Could hold the documentation part until release.

SilentSpike: A possibility for sure, but we should probably decide on (and document) an official workflow for this as I ran into the same issue when I added new cargo functions previously. It would be kind of nice to be able to merge stuff in without having to wait to reduce conflicts and so on.

654wak654: This is probably too big to just suggest in a PR comment, but ace3 could switch to a "pre-release" and "post-release" milestone model.

markooff: 1) Thank You for such a quick reaction in my request and 
2) thank you for the function ace_cargo_fnc_removeCargoItem
and also 
3) it's a great idea to specify that the  "function should also be able to accept a specific object as well as a classname" :) It may helps a lot (esp all the people who are new in scripts/sqf writing )

bux:
What you can do, is to add something like `"Added in ACE3 3.1x.x"`

docs/wiki/framework/cargo-framework.md
bux: What you can do, is to add something like `"Added in ACE3 3.1x.x"`

jonpas: Do that and we can merge it.

addons/cargo/functions/fnc_removeCargoItem.sqf
```diff
@@ -0,0 +1,48 @@
...
+        _itemsToRemove pushBack _forEachIndex;
+        ADD(_addedSpace,[_x] call FUNC(getSizeItem));
+
+        if (_x isEqualType objNull) then {
+            deleteVehicle _x;
+        };
+    };
+} forEach _loaded;
+
+{_loaded deleteAt _x} forEach _itemsToRemove;
```
PabstMirror: Every-time it deletes sometime, all the indexes will shift., e.g.
```
_loaded = ["ace_wheel", "ace_wheel"];
_itemsToRemove = [0,1];
// will only remove the first one::
_loaded deleteAt 0; // ["ace_wheel"];
_loaded deleteAt 1; // ["ace_wheel"];
```
I think this should work:
`{_loaded deleteAt (_x - _forEachIndex)} forEach _itemsToRemove;`

addons/cargo/functions/fnc_removeCargoItem.sqf
```diff
@@ -0,0 +1,48 @@
...
+
+private _loaded = _vehicle getVariable [QGVAR(loaded), []];
+
+private _itemsToRemove = [];
+private _addedSpace = 0;
+
+{
+    if (count _itemsToRemove == _amount) exitWith {};
+
+    if (_x == _itemClass || {typeOf _x == _itemClass}) then {
```
PabstMirror: This fails if objects are in the loaded array
```
   Error position: <== _itemClass || {typeOf _x == _itemClas>
   Error Generic error in expression
 File z\ace\addons\cargo\functions\fnc_removeCargoItem.sqf, line 15
```
Need to check type, then compare, something like
```
(((_x isEqualType "") && {_x == _itemClass}) || {(_x isEqualType objNull) && {typeOf _x == _itemClass}})
```

addons/cargo/functions/fnc_removeCargoItem.sqf
SilentSpike: This seems like a prime use case for `[] select {}`

SilentSpike: Scratch that, forgot you can't use the index with that syntax

654wak654: Could set the indexes to nil in the first loop, then remove all nils.

SilentSpike: That sounds like the cleanest solution 👍 Less iteration required

addons/cargo/functions/fnc_removeCargoItem.sqf
```diff
@@ -17,32 +17,33 @@
...
 
         if (_x isEqualType objNull) then {
             deleteVehicle _x;
         };
+        _loaded set [_forEachIndex, nil];
     };
 } forEach _loaded;
 
-{_loaded deleteAt _x} forEach _itemsToRemove;
+FILTER(_loaded,_x != nil);
```
SilentSpike: A quicker way to do this would be using `arrayIntersect` which removes all `nil` entries if used on the same array 👍

addons/cargo/functions/fnc_removeCargoItem.sqf
```diff
@@ -17,32 +17,33 @@
...
+private _itemsRemoved = 0;
 private _addedSpace = 0;
 
 {
-    if (count _itemsToRemove == _amount) exitWith {};
+    if (_itemsRemoved == _amount) exitWith {};
 
-    if (_x == _itemClass || {typeOf _x == _itemClass}) then {
-        _itemsToRemove pushBack _forEachIndex;
+    if (_x isEqualTo _item || {_x isEqualType objNull && {typeOf _x == _item}}) then {
```
SilentSpike: Would this not fail in the case where `_x` is an object, but not the same object as `_item`?

SilentSpike:
Hate to keep requesting changes, but now that I think about it:

It would make way more sense that in the case of an object being input, you just check if the object is in the cargo array and then delete it if true. No need to iterate and also simplifies that condition again.

docs/wiki/framework/cargo-framework.md
```diff
@@ -50,6 +50,7 @@ Event Name | Passed Parameter(s) | Locality | Description
 ---------- | ----------- | ------------------- | --------
 `ace_cargoLoaded` | [_item, _vehicle] | Global | Cargo has been Loaded into vehicle
 `ace_cargoUnloaded` | [_item, _vehicle] | Global | Cargo has been Unloaded from vehicle
+`ace_cargoRemoved` | [_itemClass, _vehicle, _amount, _itemsRemoved] | Global | Cargo has been removed (deleted) from vehicle. `_amount` is the requested number of items, `_itemsRemoved` is the actual amount that was removed.
```
654wak654: This column looks terrible with that much text. Is there any other way I could deliver this info without making it cumbersome? @bux

jonpas:
```
`ace_cargoRemoved` | [_itemClass, _vehicle, _amountRequested, _itemsRemoved] | Global | Cargo has been removed (deleted) from vehicle
```
Simply changed argument to `_amountRequested` and I think it's all self-explanatory like that.

addons/cargo/functions/fnc_removeCargoItem.sqf
```diff
@@ -23,33 +23,47 @@ TRACE_3("params",_item,_vehicle,_amount);
...
 private _loaded = _vehicle getVariable [QGVAR(loaded), []];
 
-private _itemsRemoved = 0;
 private _addedSpace = 0;
+private _itemClass = _item;
 
-{
-    if (_itemsRemoved == _amount) exitWith {};
+private _continue = if (_item isEqualType objNull) then {
+    if (_item in _loaded) exitWith {false};
```
SilentSpike: Surely this should be `!(_item in _loaded)`

654wak654: 🤦‍♂️

jonpas:
@PabstMirror Good to merge?

PR #5433, opened 2017-08-14
Made AI medic use epipen on unconscious patients

LordGolias:
- Make medic AIs use epipen on unconscious patients. They will use it only after stabilizing blood loss and pain.


BaerMitUmlaut:
This needs to handle revive state I believe.

LordGolias:
@BaerMitUmlaut, why? doesn't adding time to `treatmentOverAt` forces the unit to remain in the same state, just like with morphine?

BaerMitUmlaut:
Actually I was thinking of something else in advanced medical, my bad.

You should adjust the isInjured function to include unconsciousness though.

LordGolias:
@BaerMitUmlaut you are right. I just updated the PR adding that.

jonpas:
@BaerMitUmlaut Looks good?

addons/medical_ai/functions/fnc_healUnit.sqf
```diff
@@ -82,4 +83,13 @@ switch (true) do {
             systemChat format ["%1 is giving %2 morphine", _this, _target];
         #endif
     };
+    case _needsEpinephrine: {
+        [_target] call EFUNC(medical,treatmentBasic_epipen);
```
PabstMirror: Has no effect, should be `[_caller, _target]`

PabstMirror:
Tested working,
I also changed `healUnit` to use the non-local versions of morphine and bandage.

Issue #5404, opened 2017-08-04
Build ACE Problem


br1zey:
One addon make Problems 
the fastroping addon can't Pack

Error
```
Bis Binarise...
 Checking for binarised rtm...

Bis binarise failed to binarise some or all rtm files.
There was no skeleton that could be used.
Read the documentation

```

PabstMirror:
What version of mikro tools do you have?

br1zey:
pboProoject is 2.23

tbeswick96:
The latest version have these issues. 
Use older versions (1.73, 1.92 respectively). You can find these versions [here](https://armaservices.maverick-applications.com/Products/MikerosDosTools/FileBrowserFree)

br1zey:
ok THX

mikerobrooks:
This faulty is caused by bis binarise being unable to find a skeleton to match the contents of the rtm file. This is caused when the rtm file uses something other than the standard ofp2_manskeleton. Destruct sequences for buildings are an example.  The bones used in _that_ rtm must be in a model.cfg, and that model.cfg _must be adjacent to the rtms in the anims\folder.

PR #5442, opened 2017-08-21
Compatibility with newest mikro tools

PabstMirror:
Proper fix for #5404

- Remove all use of `access` in configs
- Add skeleton cfg for rtms
- Remove use of `-L` in makePbo for non-bin


jonpas:
Skeleton CFGs are using tabs instead of spaces.

jonpas:
Why don't we want `-L` again? (Haven't updated my tools in a while, not following updates.)

PabstMirror:
It was for lint checking the config
now it just says `-L deprecated`
if used it fails

PR #5444, opened 2017-08-21
Missile guidance for RHS hellfire and DAGR

TheNiner:
- Add the hellfire missile guidance to the RHS hellfire and DAGR


PabstMirror:
Sorry for the delayed response on this.
RHS is suppose to be adding pylon support, hopefully soon

> will all aircraft eventually transition to have jets DLC compatible loadouts?
>this is already done and part of the next update.
From their facebook, aug 21st

We could merge this now and update to plyons when they do,
but I think I'd rather wait for them to release the pylon update; 
ace_hellfire is somewhat complex to use and I'm worried about forcing it's use on their helicopters

shukari:
0.4.3 is out

cych:
need an ACE fix for the just released dynamic weapon loadout. when selected via this new system the hellfire in the weapon cycle lost the ACE supported feature (missile trajectory mode, laser code)

PabstMirror:
Looks like hardpoint `RHS_HP_HELLFIRE_RACK` will work nicely.

bux:
#5577

PR #5447, opened 2017-08-23
Nametag Changes (1)

Whalen207:
> NAMETAGS
- Nametags under cursor can now be seen from further away.
- New GVAR added to settings to configure nametag cursor distance.
- Nametag alpha affected by distance from player to target instead of camera to target. This prevents players in third person from seeing fewer / fainter tags than players in first person. On the flip side, it should greatly increase the tags seen when, for instance, a player is in a truck full of people in third person.
> OTHER
- Added Whale to AUTHORS.txt


Whalen207:
psst tabs are better than spaces.
resubmitting.

PR #5448, opened 2017-08-23
Nametag Changes

Quailsnap:
**Eventual Goals of Nametag Update:**

Currently many units use STUI Nametags and ACE nametags together, which ends up creating many ugly redundancies I *dislike*. People use ACE for 3D nearby tags that display voice, and STUI for checking identities at a distance and recognizing groups. To reduce the necessity for STUI nametags, features of ACE nametags will have to be brought up to par with the script I created [here](https://github.com/Quailsnap/WH-NT). That means:
- Nametags under cursor can be seen from further away.
- Nametags under cursor now feature the name of the target's group.
- Nametags under cursor fade slowly when you mouse away.
- Properly displaying nametags in/on vehicles -- or at least the tags for vehicle commanders. This will require using different location functions if targets are mounted.
- Distance fading starts later but falls off faster (using linearConversion). Barely-visible "squinty" levels of transparency are to be avoided, since you can't 'half-recognize' a face. You either recognize it or you don't.
- Change the default font color to not something that blends in with so much of Arma's palette.
- Change the default font weight OR size to be more legible without squinting.
- Prevent tags from changing position if there's an icon displayed above (ie: rank or soundwaves).
- (?Maybe) Zoom level (e.g. holding right click, or looking through a scope) can affect how far you can see nametags under cursor.
- (? Maybe) Zoom level affects nametag spacing (to maintain consistency) and font size.
- (? Maybe) Targets under cursor display their role and/or allow custom text.
- (? Maybe) Give players an option to display tags above target heads (as it is right now) or on their torsos (ala F3, looks better at very close or very far ranges).
- (? Maybe) Let players customize the font (in font sets), color (in color sets), and size (with precision) of the nametag text.
- (??Maybe) Display vehicle name and available seating info on mouseover.
Also, maybe:
https://github.com/acemod/ACE3/issues/3594#issuecomment-314247005

<hr>

> NAMETAGS

1 
- Nametags for cursorTarget can be seen further away (3x) default.
- A new setting (playerNamesViewDistanceCursor) has been added to accomodate that.
- The existing setting is now (playerNamesViewDistanceNearby).
- Distance checks have been changed to be between player and target instead of camera and target. This means players in third person will no longer see fewer / fainter tags than players in first person. This also means that players riding in full trucks in third person will see lots and lots of tags.

2 
- Prevent height from changing when soundwaves or ranks are in use.
> OTHER
- Added Whale to AUTHORS.txt
- Remember not to use any TABS

### DIFF
--- a/addons/nametags/functions/fnc_getCachedFlags.sqf
+++ b/addons/nametags/functions/fnc_getCachedFlags.sqf
@@ -18,12 +18,13 @@
 // Determine flags from current settings
 private _drawName = true;
 private _enabledTagsNearby = false;
-private _enabledTagsCursor = false;
+private _enabledTagsCursor = true;
 
 switch (GVAR(showPlayerNames)) do {
     case 0: {
         // Player names Disabled [Note: this should be unreachable as the drawEH will be removed]
         _drawName = false;
+		_enableTagsCursor = false;
         _enabledTagsNearby = (GVAR(showSoundWaves) == 2);
     };
     case 1: {
@@ -51,6 +52,7 @@ switch (GVAR(showPlayerNames)) do {
 };
 
 private _ambientBrightness = ((([] call EFUNC(common,ambientBrightness)) + ([0, 0.4] select ((currentVisionMode ace_player) != 0))) min 1) max 0;
-private _maxDistance = _ambientBrightness * GVAR(PlayerNamesViewDistance);
+private _maxDistanceNearby = _ambientBrightness * GVAR(playerNamesViewDistanceNearby);
+private _maxDistanceCursor = _ambientBrightness * GVAR(playerNamesViewDistanceCursor);
 
-[_drawName, GVAR(showPlayerRanks),_enabledTagsNearby,_enabledTagsCursor,_maxDistance]
+[_drawName, GVAR(showPlayerRanks),_enabledTagsNearby,_enabledTagsCursor,_maxDistanceNearby,_maxDistanceCursor]
--- a/addons/nametags/functions/fnc_getCachedFlags.sqf
+++ b/addons/nametags/functions/fnc_getCachedFlags.sqf
@@ -24,7 +24,7 @@ switch (GVAR(showPlayerNames)) do {
     case 0: {
         // Player names Disabled [Note: this should be unreachable as the drawEH will be removed]
         _drawName = false;
-        _enableTagsCursor = false;
+        _enabledTagsCursor = false;
         _enabledTagsNearby = (GVAR(showSoundWaves) == 2);
     };
     case 1: {
--- a/addons/nametags/functions/fnc_getCachedFlags.sqf
+++ b/addons/nametags/functions/fnc_getCachedFlags.sqf
@@ -24,7 +24,7 @@ switch (GVAR(showPlayerNames)) do {
     case 0: {
         // Player names Disabled [Note: this should be unreachable as the drawEH will be removed]
         _drawName = false;
-		_enableTagsCursor = false;
+        _enableTagsCursor = false;
         _enabledTagsNearby = (GVAR(showSoundWaves) == 2);
     };
     case 1: {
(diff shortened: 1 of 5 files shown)


addons/nametags/functions/fnc_getCachedFlags.sqf
```diff
@@ -18,12 +18,13 @@
...
 private _drawName = true;
 private _enabledTagsNearby = false;
-private _enabledTagsCursor = false;
+private _enabledTagsCursor = true;
 
 switch (GVAR(showPlayerNames)) do {
     case 0: {
         // Player names Disabled [Note: this should be unreachable as the drawEH will be removed]
         _drawName = false;
+        _enableTagsCursor = false;
```
PabstMirror: should be `_enabledTagsCursor`

addons/nametags/functions/fnc_onDraw3d.sqf
```diff
@@ -142,4 +117,48 @@ if (_enabledTagsNearby) then {
...
 };
 
+// W- moved tagsCursor down to avoid double tagging easier
+// Show nametag for the unit behind the cursor or its commander
+if (_enabledTagsCursor) then {
+
+    private _target = cursorTarget;
+    
+    // W- exit to prevent double tagging.
+    if (_enabledTagsNearby && {_target in _nearMen}) exitWith {};
```
PabstMirror: `_nearMen ` is from a different scope and will be undefined here

addons/nametags/functions/fnc_onDraw3d.sqf
```diff
@@ -104,10 +69,17 @@ if (_enabledTagsNearby) then {
...
-            private _distance = vectorMagnitude _relPos;
+            
+            // W- disabled below
+            //if (_enabledTagsCursor && {!_drawSoundwave}) exitWith {}; 
+            // (Cursor Only && showSoundWaves==2) - quick exit
+
+            private _distance = ACE_Player distance _target;
+            
+            // W- Changed distance to be absolute distance from player to target.
+            // This prevents players seeing less tags (or fainter ones) in third person.
```
dedmen: That change also prevents from seeing Nametags in Zeus interface doesn't it?

Quailsnap: Correct. Remote controlled units will still be able to see tags, but the Zeus camera will not. That said, tag distance by default is 5m and it's very, very infrequent for the Zeus camera to ever see tags anyways.

Quailsnap:
Updated header with more info on what I intend to do.

sargken:
A idea for you to do is make a cba color slider for the color of the nametags.

Quailsnap:
@sargken That's actually already been there for a while -- but it's "hidden" behind the colors tag. When ACE moves to CBA settings it'll be more obvious.

( I didn't even know it was there until I started working on this system. )

Quailsnap:
Made font height dependent on font size. This avoids a little issue where smaller sizes (ie: very small) would appear much higher than their large counterparts.

Next change is going to be implementing zoom -- font size, font height, and visible tag distance under cursor all adjusting depending on the player's FOV. This is going to be the biggest change so far for a few reasons:
1. Changes in zoom will have to be tracked nearly every frame if you guys want transitions to be smooth. This could mean a possible performance hit.
2. If someone can zoom in far enough to see a face, they can see a tag. This includes from miles away.

Quailsnap:
Putting this on the backburner for myself until some other tasks get done. Will revisit and revise after I revamp the system I'm using.

Quailsnap:
still on backlog as of 1/4

addons/nametags/functions/fnc_onDraw3d.sqf
Cuel: I didn't even know nametags were enabled in Zeus camera tbh

DeliciousJaffa: The position is already in the `_camPosAGL` var anyway, so may as well change it to use that and support Zeus.

thojkooi:
Thanks for your PR. Closing due to merge requests. This can be reopened once the conflict has been resolved.

PR #5453, opened 2017-08-24
Allow digging in malden grass

### LINKED ISSUE #5450: Entrenching Tool/ Sandbags not working on Malden 2035
Fully Updated

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
-Sandbags and Entrenching tool don't seem to be working on the new Malden 2035 map. Everywhere else it works perfectly fine.

- Singleplayer Malden 2035

PabstMirror:
Close #5450

Malden uses `GdtGrassGreen` which has a dust of 0.05


jonpas:
This is just... wow. I wonder if that's plastic grass or how does it grow out of almost-concrete.

SilentSpike:
Could we not have also increased the dust value? Seems stupid that it's so low

PR #5468, opened 2017-08-30
WIP: Add SQFLint to Travis

jonpas:
- Title

Todo:
- [x] Make it work without crashing (possibly needs waiting for SQFLint updates)
- [ ] Return an error code for warnings (warnings-as-errors) for Travis
- [ ] Run only on changed `.sqf` files


SilentSpike:
Is there no way to run for only the changed files?

jonpas:
@SilentSpike As said in OP:
> [ ] Run only when the diff includes an `.sqf` file change (otherwise build times will go up significantly for other PRs as well)

So yes, there is a way by diffing and pulling out file name extensions, have done it before.

SilentSpike:
I meant more, it looks like it's running for the entire addons folder: `sqflint -d addons`, is there no way to run the linter over *only* the changed files?

jonpas:
Ah, yes, good idea!

LordGolias:
Let me know if you need help here. I can modify SQFLint so that it works well here.

jonpas:
@LordGolias Right now it crashes as seen [here](https://travis-ci.org/acemod/ACE3/builds/270160569).

LordGolias:
You are right. I just pushed a fix of that to master of sqf. I also updated the pip version so that you can install the latest using pip install...

jonpas:
@LordGolias Very nice! Travis ran to the end now.

Next step is making it return an error code instead. I am not sure this should be configurable somehow as some might not want warnings reported as errors.

jonpas:
@SilentSpike We might still want to run full lint on merges into master, just to identify anything that may get through in whatever way maybe?

SilentSpike:
I think it would be unnecessary unless the linter is updated in a way that it cross-examines function calls for return value types or something. Might not be a bad idea to run it fully for release branch though?

thojkooi:
Why not run it for every file? Does the linting take a long time to complete?

SilentSpike:
Nah not necessarily, not sure how it would scale if being run for every commit though. I'm committing the sin of premature optimization 😆

jonpas:
I recommend:
- Changed files lint on PR/branch commits (I believe that's what @thojkooi meant?)
- Full lint on merges to `master` and `release`

Full lint does take a bit, we have A LOT of files.

thojkooi:
But is it an issue? How long does a full lint take?

We almost never merge anything straight away. Just lint everything is my suggestion.

jonpas:
```
real    1m19.454s
user    0m57.734s
sys     0m0.922s
```

thojkooi:
I'd just run the full lint.  We almost never merge anything that fast anyway.

PR #5470, opened 2017-09-02
7,62 x 54 R ammo update

kOepi123:
the actual file to be merged, will follow without comments and uncollapsed

**adds new barrel lengths for PK/PKM, PKT, Mosin nagant**
- it will allow weapon configs to have more accurate muzzle speeds

- sidenote: for reasons of type of weapon differences ( probably isolation and self-reloading energy )
the longer barrel is not meaning, that the muzzle velocity increases across different types of weapons:
e.g.: the mosin nagant rfile has a ( much ) higher muzzle velocity than the PKT, though the barrel is 2.5 inches shorter.

- I am not sure whether this issue can or should be set by altering the initspeed coefficent ( is there already something in place? ). In that case it would disqualify this pull request.

[mousin barrel experiment.pdf](https://github.com/acemod/ACE3/files/1272583/mousin.barrel.experiment.pdf)


addons/ballistics/CfgAmmo.hpp
```diff
@@ -307,8 +307,9 @@ class CfgAmmo {
         ACE_velocityBoundaries[]={};
         ACE_standardAtmosphere="ICAO";
         ACE_dragModel=1;
-        ACE_muzzleVelocities[]={700, 800, 820, 833};
-        ACE_barrelLengths[]={406.4, 508.0, 609.6, 660.4};
+        ACE_muzzleVelocities[]={
```
jonpas: Keep it on the same line.

AUTHORS.txt
```diff
@@ -131,3 +132,4 @@ voiper
 VyMajoris(W-Cephei)<vycanismajoriscsa@gmail.com>
 Winter <simon@agius-muscat.net>
 zGuba
+kOepi <ickekoep@hotmail.com>
```
bux: file is sorted by alphabet

ulteq:
> it will allow weapon configs to have more accurate muzzle speeds

In that case you should tweak the config values in the weapon class itself.
We're using the [`initSpeed`](https://github.com/acemod/ACE3/blob/master/addons/ballistics/CfgWeapons.hpp#L517) to fine tune the dependency between ammo and weapon.

I do like the changes you made to the minimum and maximum values of the muzzle velocity array, but I think it is a bad idea to make it non-monotonic.

Changing the muzzle velocity array requires a reassessment of the muzzle velocities of all weapons that use this type of ammo.

It usually also means that you have to reevaluate the `initSpeed` values of all related weapons and magazines as well as the `airFriction` value, if you want to make it perfect again.

Different muzzle velocities require different `airFriction` values for a perfect trajectory match between vanilla- and advanced ballistics.

kOepi123:
@ulteq 
I see. I thought it would not require that much work. Other than making advanced ballistics independent from basic, I dont see anything to do here then.

ulteq:
We could probably get away with changing the muzzle velocity array like this:
```
ACE_muzzleVelocities[]={760, 795, 835, 865, 855};
ACE_barrelLengths[]={406.4, 508.0, 604.5, 736.6, 772.2};
```
But it would be better to also adjust the `initSpeed` values of the two vanilla weapons that use this type of ammunition.

kOepi123:
I cannot change it according to any standard other than seeing if it "looks ok".
I could probably make a good guess in a non-AB, with basic ballistics, environment for the initspeed though.

ulteq:
> it will allow weapon configs to have more accurate muzzle speeds
> In that case you should tweak the config values in the weapon class itself.

I have to correct myself. That will not work, because the Advanced Ballistics module is not taking the `initSpeed` values into account. It is solely based on the muzzle velocity array and the actual barrel length.

We would have to add a new config entry into the weapon class, that modifies the muzzle velocity based on the gas system configuration.

> I cannot change it according to any standard other than seeing if it "looks ok".

It is relatively simple to come up with the required `initSpeed` values in the weapon and magazine classes.

a) Set the `initSpeed` value in the magazine class to the typical muzzle velocity of that ammo type.
b) Calculate the `initSpeed` value of the weapon class so that it fits the following equation:

`initSpeed (magazine class)` * - `initSpeed (weapon class)` = Desired muzzle velocity based on `ACE_muzzleVelocities`, `ACE_barrelLengths` and `ACE_barrelLength`

Lets say you want calculate the `initSpeed` values for the [Rahim](https://github.com/acemod/ACE3/blob/master/addons/ballistics/CfgWeapons.hpp#L512-L520):

1. The [`initSpeed`](https://github.com/acemod/ACE3/blob/master/addons/ballistics/CfgMagazines.hpp#L381) value of the magazine class is set to 800.
2. The muzzle velocity with Advanced Ballistics is:
```
lowerVelocity = 835 (from ACE_muzzleVelocities)
upperVelocity = 865 (from ACE_muzzleVelocities)
lowerBarrelLength = 604.5 (from ACE_barrelLengths)
barrelLength = 609.6 (from ACE_barrelLength)
upperBarrelLength = 736.6 (from ACE_barrelLengths)
slope = (upperVelocity - lowerVelocity) / (upperBarrelLength - lowerBarrelLength)
muzzleVelocity = lowerVelocity + slope * (barrelLength - lowerBarrelLength)
-> desiredMuzzleVelocity ~= 836
```

So we get the following `initSpeed` value for the weapon class:
```
initSpeed (weapon class) = -desiredMuzzleVelocity / initSpeed (magazine class)
initSpeed (weapon class) = -836 / 800
initSpeed (weapon class) = -1.045
```
We should probably write a small SQF program that automates this process.

dedmen:
Could you please comment here where you got these values from? for future reference.

kOepi123:
@dedmen 
I am happy to see that you follow my suggestion, and it indeed could help you with your calculations.

ulteq:
- [ ] Update `B_762x54_Ball` ammo with:
```
ACE_muzzleVelocities[]={760, 795, 835, 865, 855};
ACE_barrelLengths[]={406.4, 508.0, 604.5, 736.6, 772.2};
```
- [ ] Update `srifle_DMR_01_F` weapon `initSpeed`
- [ ] Update `10Rnd_762x54_Mag` magazine `initSpeed`

- [ ] Update muzzle velocity (and probably also the scope base angle) in ATragMX gun list:
https://github.com/acemod/ACE3/blob/master/addons/atragmx/functions/fnc_initGunList.sqf#L57

PR #5488, opened 2017-09-10
Add setting to allow the transfer of objects from a casualty to a body bag

AACO:
**What it does:**
- Simple concept, a bit convoluted in execution, but it adds a body bag 'crate' and copies the targeted unit's loadout into the body bag.
- This is ACE_Settings driven, when turned off (default) the only added complexity/behavior is a quick boolean check.

**Questions/Concerns:**
- Should the new functions live somewhere else? I can only see their use for medical right now, but they're very generic.
- Primary and secondary weapons are dropped on death, and handled in weapon holders. This becomes a problem when the unit's body is deleted as the holders get cleaned up as well (unless the holders are interacted with before the body is removed). Does anyone have a tried/true method to making the weapon holders stick around after death?

**Short comings:**
- Weapons have to be broken down to components, there's no good way to add a configured weapon to a container (without extensive modding or hacking)
- A lot of global cargo calls, but it's the only way I could think of simulating the 1:1 gear setup of Mr. Dead Dude


jonpas:
I built a similar thing for my community (back in April it seems), at the time I decided not to PR it to ACE3 after discussing with other devs, but it hasn't  been discussed much.
https://github.com/Theseus-Aegis/tac-a3-mods/tree/master/addons/bodybag

I handle weapons (as much as possible with those holders sometimes not getting removed, sometimes getting removed...). There is also % chance that can be set on destroying each item getting moved, otherwise seems pretty identical, except a lot shorter.

There is no need for additional holder object, scripting commands ignore `maximumLoad`. That's why my component only has 2 functions and a short config file.

AACO:
> I handle weapons (as much as possible with those holders sometimes not getting removed, sometimes getting removed...). There is also % chance that can be set on destroying each item getting moved, otherwise seems pretty identical, except a lot shorter.
Grabbing it on death is a great idea, although in my mind I only wanted to allow what was on the body to be accessible leaving everything else to itself. Also deleting all the nearest weapon holders (required to avoid duplication I know) seems a tad aggressive and could wipe out other dropped equipment near by.

 >There is no need for additional holder object, scripting commands ignore maximumLoad. That's why my component only has 2 functions and a short config file.
Not a terrible idea, but I wanted to leave open putting stuff back into the bag in case it was accidentally removed/you need to move the bag to the ground to open it. I can probably ratchet down some of the limits to avoid having it used as a go to container.

>At the time I decided not to PR it to ACE3 after discussing with other devs, but it hasn't been discussed much (I kinda gave up, because I personally didn't think it would really fit, even without item destroying chance).
I'd like to see some discussion on this. It's off by default so only people who would want to use it would turn it on. I personally don't understand weapon destruction on body bagging, however I could see it being standalone thing on death.

> Realistically looking you wouldn't load someone with full gear in (or would you?).
Well ArmA realities being ArmA realities, if you want to move a dead body it has to go in the bag, which this would help out with.

jonpas:
> Also deleting all the nearest weapon holders (required to avoid duplication I know) seems a tad aggressive and could wipe out other dropped equipment near by.

Agreed, but that's the best I could come up with, it only deletes very close holders. I didn't find any better way to handle those holders.

> Not a terrible idea, but I wanted to leave open putting stuff back into the bag in case it was accidentally removed/you need to move the uniform/vest/backpack to the ground to open it. I can probably ratchet down some of the limits to avoid having it used as a go to container.

By how I do it any containers (uniform, vest, backpack) get put it empty and all items from them into the body bag directly, thus overcoming what you just described.

> I'd like to see some discussion on this. It's off by default so only people who would want to use it would turn it on. I personally don't understand weapon destruction on body bagging, however I could see it being standalone thing on death.

Yes definitely. We have item being destroyed chance because we don't want full gear to ever be returned, plays in with our persistent gear + armory we are running, so kind of specific to us. We can still do that if there is an event thrown after moving items with the body bag object (or holder) as argument.

AACO:
@jonpas would listening for ace_placedInBodyBag work (it seems exactly what you would need, besides the race for body cleanup), or would you need a separate event?

jonpas:
That's what I am using now, but it wouldn't be enough in this case because items might not be moved yet when I'd receive the event.

Overall, I still think this PR over-complicates things a lot with additional holder and adding containers and all that. I just updated [my component](https://github.com/Theseus-Aegis/tac-a3-mods/pull/316/files) with the things it was still missing (thanks to you) and also added support for weapons/backpacks with preset attachments/contents. Again, this might be just me, but I think my 2 files do exactly what yours do, but in a simpler way.

AACO:
Maybe simpler is better, but my goal was to get as 1:1 in terms of gear. It doesn't look like this got any traction, so to POTATO it goes

jonpas:
You can leave it open for a while longer, it hasn't been open for even a week, hard to get traction that fast. ^^

You mean 1:1 in gear to keep them in same containers (stuff in backpack) instead of dropping them directly in? I think it makes it easier to drop directly in. Also I am now thinking this would fit ACEX more, but I'd prefer to go with my approach, or a mix of yours and mine.

What do other ACE3 Devs think?

bux:
imho it wouldn't hurt to have such functionality. Especially if it's off by default.

ulteq:
I like the idea and I like the new code.

thojkooi:
fyi; ci/circleci fails because it requires a configuration file from master. Updating to latest origin/master should solve it (but is not required for now).

AACO:
Fixed a merge issue ("" -> "NONE" in create vehicle)

PR #5489, opened 2017-09-10
Kestrel4500 - Fixed head- and crosswind calculation

### LINKED ISSUE #5181: Kestrel headwind and crosswind computation wrong
**Arma 3 Version:** `1.70` stable
**CBA Version:** `3.3.1` stable
**ACE3 Version:** `3.9.2` stable

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
The calculation for headwind and crosswind on the Kestrel is wrong.

**Steps to reproduce:**
For example, without advanced ballistics, set the runway/target direction to straight into the wind. The crosswind should be close to zero, and the headwind close to max, yet this is not the case. In fact from the calculation: `_textCenterBig = Str(round(cos(GVAR(RefHeading)) * _windSpeed * 10) / 10);` you can notice that the direction of the wind itself is not even used! Set the reference heading straight north, and you will get maximum headwind, set it straight east, you will get maximum crosswind; it does not matter where the wind blows, only how much.

Assume "unit circle" used by Arma going clockwise, and with 360 degrees. Assume the reference direction, R, is 0, and the wind is coming from relative X.

Then the head wind is:  `cos(X) * windSpeed`
And the cross wind is: `-sin(X) * windSpeed`

For example, lets say relative X is 30 and windspeed is 10,
then headwind becomes =  8.66 which fits because wind is coming from close to the reference direction,
and crosswind becomes = -5.00 which also fits since the wind is coming a bit to the right relative to the reference direction.

Now if the reference direction is not 0, then simply subtract R from X and modulo with 360 (which you technically don't have to), making the calculation:

headwind :  `cos(windFromDir - RefHeading) * windSpeed`
crosswind: `-sin(windFromDir - RefHeading) * windSpeed`

Currently the measurements are shown in absolute values. From the manual linked to on the ACE Features site, it seems at least tailwind is shown negative, so the absolute value should not be shown for that. Maybe the same applies for crosswind.

When running with advanced ballistics, the intention seems to be the player's direction also affects the measurement, which agree with the manual that states that you should turn the kestrel into the wind. So rather than always getting the true windspeed it should scaled by how well the player looks into the wind (the players direction is opposite where the wind is headed). And rather than the true wind direction the player's current direction (the player's guess of where the wind comes from) should be used. So for advanced ballistics the calculations become. 

headwind :  `cos(playerDir - RefHeading) * windSpeed * cos(playerDir - windFromDir)`
crosswind: `-sin(playerDir - RefHeading) * windSpeed * cos(playerDir - windFromDir)`

Editor (Singleplayer)

**Placed Modules:**
- Add the list of modules you have placed on the map. Use 'None' if the error occurs without any modules.

None

**RPT log file:**
- If possible at the time the bug is encountered, go to ACE Options and select "Debug To Clipboard", this will print extensive debug information to the RPT file.

ulteq:
- The wind direction was not taken into account when calculating head- and crosswinds.
- The output of the wind direction was formatted incorrectly.
- Those changes are only relevant when AdvancedBallistics is disabled.
- Fixes #5181

### DIFF
--- a/addons/kestrel4500/functions/fnc_generateOutputData.sqf
+++ b/addons/kestrel4500/functions/fnc_generateOutputData.sqf
@@ -136,7 +136,7 @@ if (GVAR(referenceHeadingMenu) == 0) then {
                     _textCenterBig = Str(round(abs(sin(GVAR(RefHeading) - _playerDir) * _windSpeed) * 10) / 10);
                     _textInfoLine1 = format["%1 m/s @ %2", round(_windSpeed * 10) / 10, round(_playerDir)];
                 } else {
-                    _textCenterBig = Str(round(abs(sin(GVAR(RefHeading)) * _windSpeed) * 10) / 10);
+                    _textCenterBig = Str(round(abs(sin(GVAR(RefHeading) - _windDir) * _windSpeed) * 10) / 10);
                     _textInfoLine1 = format["%1 m/s @ %2", round(_windSpeed * 10) / 10, round(_windDir)];
                 };
                 _textInfoLine2 = "- set heading";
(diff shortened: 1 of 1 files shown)


QuickDagger:
With Adv Bal module on, everything works right. Higher wind when standing, lower when crouch, lower when prone. High value when measured 12 o'clock wind.

Problem is the majority of servers don't activate Adv. Bal. And the Kestrel always reads the same wind values, no mater head or cross wind, standing or prone.

PR #5496, opened 2017-09-11
AdvancedBallistics - Utilize new 'parseSimpleArray' command

ulteq:
- Decent performance improvement compared to the old `call compile` approach
- Requires DLL rebuild.

frametime_degradation


bux:
reminder #5497

PR #5498, opened 2017-09-11
Move handle on carrier doors

jonpas:
- Title

There are 3 animations for carrier doors:
```
"door_1_rot","door_handle_1_rot_1","door_handle_1_rot_2"
```
`rot_2` does not seem to be used with normal action menu opening at all, so using `animationSource` does not work, because it will activate both of those animations as well and handle will end up in original position (and also be broken for action menu usage).

I didn't find any other good way to extract those handle animations, so it's just checking for the USS Freedom carrier directly and adding the handle based on the door number.

https://i.imgur.com/HzJZ058.gifv

### DIFF
--- a/addons/interaction/XEH_postInit.sqf
+++ b/addons/interaction/XEH_postInit.sqf
@@ -83,7 +83,7 @@ GVAR(isOpeningDoor) = false;
     call EFUNC(interaction,openDoor);
     true
 }, {
-    //Probably don't want any condidtions here, so variable never gets locked down
+    //Probably don't want any conditions here, so variable never gets locked down
     // Statement
     GVAR(isOpeningDoor) = false;
     true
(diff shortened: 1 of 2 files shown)


PabstMirror:
btw at some point we might want to take a look at https://community.bistudio.com/wiki/getCursorObjectParams
as I think it can get the named selections for door interaction
(but might not help with the specific carrier issue)

jonpas:
> tested working 👍

I did attach a gif of it working, but good to know it works for someone else as well. :D

jonpas:
```
getCursorObjectParams
-> [2a1f2704b80# 22: carrier_01_island_01_f.p3d,["door_1"],2.58767]
```
It can, `["door_1"]` above, but how does a selection name help us? Can we get animation/animationSource name from it?

PabstMirror:
I think we might be able to replace getDoor, but nothing to do with this PR i guess

PR #5501, opened 2017-09-11
Fastrope animations and in-line privatization

jonpas:
- Add animations from older Arma games (support for eg. CUP which are still using those - not including any special new animations like fancy ramps)
- Macroize animations names
- Change to in-line privatization
- Fix inverted comments for public `onPrepare(/Cut)Common` functions

### DIFF
--- a/addons/fastroping/script_component.hpp
+++ b/addons/fastroping/script_component.hpp
@@ -16,3 +16,8 @@
 #endif
 
 #include "\z\ace\addons\main\script_macros.hpp"
+
+
+#define ANIMS_HOOK ["extendHookRight", "extendHookLeft"]
+#define ANIMS_DOOR ["door_R", "door_L", "CargoRamp_Open", "Door_rear_source", "Door_6_source", "CargoDoorR", "CargoDoorL"]
+#define ANIMS ["dvere1_posunZ", "dvere2_posunZ", "doors"]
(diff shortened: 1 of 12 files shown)


addons/fastroping/script_component.hpp
```diff
@@ -16,3 +16,8 @@
 #endif
 
 #include "\z\ace\addons\main\script_macros.hpp"
+
+
+#define ANIMS_HOOK ["extendHookRight", "extendHookLeft"]
+#define ANIMS_DOOR ["door_R", "door_L", "CargoRamp_Open", "Door_rear_source", "Door_6_source", "CargoDoorR", "CargoDoorL"]
+#define ANIMS ["dvere1_posunZ", "dvere2_posunZ", "doors"]
```
BaerMitUmlaut: Could we find a different name for this macro? "Anims" is so generic.

jonpas: `ANIMS_DOOR` -> `animateDoor` (requires "door" source)
`ANIMS` -> `animate` ("user" source)

jonpas: The only reason they are split in two is because they require different commands (above comment).

I have no idea what else to name it, suggestions?

654wak654: `ANIMS_USER` then?

jonpas: I named them `ANIMS_ANIMATEDOOR` and `ANIMS_ANIMATE` to make clear what command is which. Is that good or should I go with `ANIMS_DOOR` and `ANIMS_USER` rather? One is what command is used, the other is what type is used in config.

PR #5502, opened 2017-09-11
Use CBA Misc Items

PabstMirror:
For https://github.com/CBATeam/CBA_A3/pull/744

```
class ACE_ItemCore: CBA_MiscItem {};
```
Should we keep `ACE_ItemCore` or just drop it and use the CBA item base for everything?


jonpas:
Might be wise to keep it, in case we add something else specific to ACE3 to all items in the future.

jonpas:
Tested with CBA 3.4.1, all looks good. 👍

PR #5503, opened 2017-09-12
Move the hellfire hud to the laser module

PabstMirror:
Needed for #5364

Should be no functionality change, just adding a new config ` EGVAR(laser,showHud)` 
so that the hud can be used by other modules.


jonpas:
Are `requiredAddons` good?

PabstMirror:
yeah, 
hellfire requires missileGuidance, which requires laser

PR #5519, opened 2017-09-16
Medical & Captive Load Menu Overhaul

654wak654:
- Add child actions to "Load Patient" and "Load Captive" for all near vehicles with seats.
- Make captives component use more person loading code from ace_common. Although the code could still use some refactoring, I left it alone because I don't have a grip over all of it.
- Shamelessly steal most code from #4871


addons/medical/functions/fnc_addLoadPatientActions.sqf
```diff
@@ -0,0 +1,42 @@
+/*
+ * Author: 654wak654
+ * Adds child actions to the "load patient" action for near vehicles
```
jonpas: Missing final dot.

addons/medical/functions/fnc_addLoadPatientActions.sqf
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Arguments:
+ * 0: The medic <OBJECT>
+ * 1: The patient <OBJECT>
+ *
+ * Return Value:
+ * The child actions
```
jonpas: ` <ARRAY>`

Also "The" is superfluous everywhere.

addons/captives/functions/fnc_addLoadCaptiveActions.sqf
```diff
@@ -0,0 +1,42 @@
+/*
+ * Author: 654wak654
+ * Adds child actions to the "load captive" action for near vehicles
```
jonpas: Missing final dot.

addons/captives/CfgVehicles.hpp
```diff
@@ -51,6 +51,7 @@ class CfgVehicles {
                     showDisabled = 0;
                     icon = QPATHTOF(UI\captive_ca.paa);
                     priority = 2.2;
+                    insertChildren = QUOTE([ARR_2(_player, _target)] call DFUNC(addLoadCaptiveActions));
```
jonpas: Just `_this call...` works here as well, so you don't need to do macro shenanigans.

PabstMirror:
I will try to test this laster, but there are some delicate issue when loading a person into a turret or FFV "gunner" position

addons/captives/CfgVehicles.hpp
Dystopian: `_this` is `[_target, _player]` and @654wak654 wants reversed array (I think).
Also space in `ARR_2` macro. Also see @commy2 comment https://github.com/acemod/ACE3/pull/5384#pullrequestreview-62922317

jonpas: He is calling his own function, may as well revers the order there.

Single quotes are fine but you lose syntax highlighting because it's all a long string then, personal preference. I prefer single quotes if it really makes the thing shorter, but in this case you don't need anything but QUOTE and DFUNC.

addons/medical/functions/fnc_addLoadPatientActions.sqf
654wak654: I copied most things from other medical headers so I'm keeping all the "the"s.

addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * Example:
+ * [bob] call ace_common_fnc_nearestVehiclesFreeSeat
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_unit"];
+
+private _nearVehicles = nearestObjects [_unit, ["Car", "Air", "Tank", "Ship_F", "Pod_Heli_Transport_04_crewed_base_F"], 10];
```
thojkooi: Making the distance configurable by allowing it to be passed along as a parameter would be more flexible.

```sqf
params ["_unit", ["_distance", 10]];

private _nearVehicles = nearestObjects [_unit, ["Car", "Air", "Tank", "Ship_F", "Pod_Heli_Transport_04_crewed_base_F"], _distance];
```

654wak654: Because of this function's very specific purpose it's good that everyone gets the same search distance IMO.

addons/medical/functions/fnc_addLoadPatientActions.sqf
jonpas: We should go for no "the"s, majority (I hope) is not using them, it's superfluous and quiet silly.

addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
jonpas: It's a common function though. Name of it implies it finds nearest no matter the distance. As a common function I agree with @thojkooi, can always simply call it with distance of 10 from medical/captives.

addons/medical/functions/fnc_addLoadPatientActions.sqf
Dystopian: at least for non-medical files

jonpas:
Handcuffing a unit and loading (via captives) ✔️ 
Surrendering a unit and loading (via captives) ✔️ 
Killing/downing a unit and loading (via medical) ✔️ 
Handcuffing a unit, then killing/downing a unit and loading (via captives) ✔️ 
Handcuffing a unit, then killing/downing a unit, then un-handcuffing a unit and loading (via captives or medical) ❌

654wak654:
> Handcuffing a unit, then killing/downing a unit, then un-handcuffing a unit and loading (via either) ❌

This has been fixed by #5544

PR #5521, opened 2017-09-16
Fix Underwater interactions (magrepack, loading, refuel, medical legs, dragging)

### LINKED ISSUE #5513: Underwater repacking doesn't show hint when it's done
**Arma 3 Version:** `1.76` stable
**CBA Version:** `3.4.1` stable
**ACE3 Version:** `3.10.2` dev df739bd36cd36f5a63d6cb8e56f5b2fda4cce770

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- Repacking underwater doesn't show the mags repacked when it's done

**Steps to reproduce:**
- Go underwater and repack

-Editor singleplayer

**Placed Modules:**
- None

**RPT log file:**
- Nothing of interest there.

### LINKED ISSUE #5515: Loading patients underwater does not work
**Arma 3 Version:** `1.76` stable
**CBA Version:** `3.4.1` stable
**ACE3 Version:** `3.10.2` dev df739bd36cd36f5a63d6cb8e56f5b2fda4cce770

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- Load patient does not work underwater

**Steps to reproduce:**
- Try to load a patient in a SDV inland, then try underwater.

-Editor singleplayer

**Placed Modules:**
- None

**RPT log file:**
- Nothing of interest there.

jonpas:
- Fix Magazine Repack underwater - fix #5513 
- Fix Medical loading patients underwater - fix #5515
- Fix Cargo load object underwater
- Fix Refuel Take Nozzle underwater (will drop nozzle when starting to walk out - has to be picked up again, current limitation due to missing animation transitions underwater)
- Fix Dragging Drag underwater (no Carry due to animation timing issues and transition shenanigans, doesn't make sense either)
- Fix Medical Legs self-interactions underwater (were missing `"isNotSwimming"` exception)
- Use new `FUNC(common,isSwimming)` which uses animation name to determine swimming, instead of `underwater` for certain actions (all except Advanced Throwing), reports `true` even when swimming on surface (same limitations apply as when diving) instead of just when diving (Advanced Throwing is different as it allows proper throws on surface, but only drop in water to show you can't throw far because water would stop it)
- Fix `FUNC(common,fixPosition)` to work underwater (use `getPosATL` instead of `getPos`) and correctly apply slope adjustment for objects without gravity/physics as well, also do the same for objects with disabled simulation (from #4605)


jonpas:
```sqf
_unit action ["SwitchWeapon", _unit, _unit, 299];
```
does not work underwater, thus refuel does not work underwater, because it relies on not having the weapon in hand. This would be easily solvable by just ignoring the "no weapon in hand" condition when underwater, but it would also mean you could then walk outside with nozzle and weapon both in hands.

I have no clue how animations work in Arma, so any help around this would be great.

addons/refuel/script_component.hpp
```diff
@@ -30,4 +30,5 @@
     #define TIME_PROGRESSBAR(X) (X)
 #endif
 
-#define INTERACT_EXCEPTIONS "isNotInside", "isNotOnLadder", "isNotRefueling"
+#define INTERACT_EXCEPTIONS "isNotInside", "isNotOnLadder", "isNotRefueling", "isNotSwimming"
```
Dystopian: maybe `#define INTERACT_EXCEPTIONS INTERACT_EXCEPTIONS_REFUELING, "isNotRefueling"`?

jonpas: Can improve, want to figure out this refuel thing first.

alganthe:
I can confirm that magazine repack and cargo have been fixed with this PR, loading / unloading patients is also much more consistent now.

jonpas:
Did the best on refuel until proper transition can be made if it even can.

Also using `isTouchingGround` to also catch surface swimming and not just diving (it's "isNotSwimming" after-all).

All done from my side.

addons/interaction/XEH_postInit.sqf
```diff
@@ -106,7 +106,7 @@ GVAR(isOpeningDoor) = false;
 {false},
 [20, [true, false, false]], false] call CBA_fnc_addKeybind;
 
-["isNotSwimming", {!underwater (_this select 0)}] call EFUNC(common,addCanInteractWithCondition);
+["isNotSwimming", {isTouchingGround (_this select 0)}] call EFUNC(common,addCanInteractWithCondition);
```
PabstMirror: `isTouchingGround  player` is false when in aircraft
this breaks the microDagr actions and button

maybe `(isTouchingGround _unit) || {(vehicle _unit) != _unit}` ?

Still broken during free-fall, but I don't think you could realistically do much during that.

jonpas: Hmh, I thought this does checks differently. Will look into it.

jonpas: Would checking height be sufficient?
```sqf
{isTouchingGround (_this select 0) || {(getPosASL (_this select 0)) select 2 > 0}}
```
Unsure if that would work with CUP ponds for example though, as they are above sea level, no clue if `surfaceIsWater` reports `true` for those either, will test tomorrow.

jonpas: Ooh just got an idea, animation state checking. Swimming animations start with one of those (Actions):
```sqf
[bdv","bsw","dve","sdv","ssw","swm"]
```

alganthe: surfaceIsWater reports true for inland water, I just checked on chernarus' black lake

PabstMirror: maybe if (animation->"actions" inheritsFrom "SwimmingActions")

jonpas: @alganthe ah, good to know! However I prefer the animation solution as animation is actually what prevents all those interactions for executing normally like they would on land. That way there is probably less inconsistency.

jonpas: @PabstMirror Doesn't seem like they inherit from that at all, maybe a select few?
``` sqf
[configFile >> "CfgMovesMaleSdr" >> "States" >> "abdvpercmstpsnonwrfldnon", true] call BIS_fnc_returnParents;
-> ["AbdvPercMstpSnonWrflDnon","AidlPercMstpSrasWrflDnon_G0S","AmovPercMstpSrasWrflDnon","AmovPercMstpSlowWrflDnon","StandBase","Default"]
```

jonpas:
@commy2 can you take a look at updated `FUNC(common,fixPosition)`? In case I missed some case.

PR #5523, opened 2017-09-17
Fix spectator onLoad

PabstMirror:
Fix crash on start
```
 File z\ace\addons\spectator\ui.hpp, line 20: '/ace_spectator_display.onLoad': Missing ';' at the end of line
 Error context , _this select 0]";
ErrorMessage: File z\ace\addons\spectator\ui.hpp, line 20: '/ace_spectator_display.ace_spectator_display': '"' encountered instead of '='
```

I guess the macro doesn't like being in a quote macro

### DIFF
--- a/addons/spectator/ui.hpp
+++ b/addons/spectator/ui.hpp
@@ -18,7 +18,7 @@ class GVAR(display) {
     movingEnable = 0;
     closeOnMissionEnd = 1;
     
-    onLoad = QUOTE(SETUVAR(GVAR(display),_this select 0));
+    onLoad = QUOTE(with uiNameSpace do {GVAR(display) = _this select 0};);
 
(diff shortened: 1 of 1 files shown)


alganthe:
I can confirm the crash has been fixed.

addons/spectator/ui.hpp
```diff
@@ -18,7 +18,7 @@ class GVAR(display) {
     movingEnable = 0;
     closeOnMissionEnd = 1;
     
-    onLoad = QUOTE(SETUVAR(GVAR(display),_this select 0));
+    onLoad = QUOTE(with uiNameSpace do {GVAR(display) = _this select 0};);
```
jonpas: Shouldn't it rather be `uiNamespace setVariable...`?

PabstMirror: same effect either way,
I like this method because there are no arrays to deal with

PR #5524, opened 2017-09-17
showMouseHint - Allow adding additional icon/text lines

PabstMirror:
Want this for https://github.com/acemod/ACEX/pull/76

image

Not really thrilled with how the icons look, 
and they might have problems with localization (eg german "STUG")
Might drop them for a wider text bar and just use `localize "str_dik_shift"`


Dystopian:
vote for simple rectangles with text inside. I think Shift should be the widest, then Ctrl, and Alt is the narrowest.

654wak654:
Could do what BIS does and add color too.

commy2:
>STUG

Sturmgeschütz?
I think you meant "Strg".

commy2:
```
localize "str_dik_control"
localize "str_dik_shift"
localize "str_dik_alt"
```

jonpas:
I'd personally add some spacing between icon and text as well, even for mouse buttons.

PabstMirror:
image

jonpas:
Could use a key-shape background? Like on the image in OP, just no text on them.

PabstMirror:
It has very limited use for now, I'd rather keep it simple for now
we can make it pretty later if it gets used more

Issue #5215, opened 2017-05-28
Add object support to zeus fnc_getModuleDestination


kymckay:
**Description:**
- This issue is basically just a reminder for myself or anyone interested to look into. Currently `fnc_getModuleDestination` returns only position ASL, but it could be nice if it would (optionally?) return the object under the curator cursor when the destination is clicked. For example, `doSuppressiveFire` command used by the respective module supports the target being a unit and not just a position, would allow for more accurate suppression of moving units.

chris579:
How does the AI behave when you assign an object for `doSuppressiveFire`? Will it still suppress the target even if it is not in line of sight (e.g. hidden by an house)? Would support the cheaty AI. However, a object param would indeed be nice.

My approach for the object param would be using nearestObjects or iterating through all players which is more hacky than stylish. Do you have a different approach for this?

SilentSpike:
Yeah, zeus caches the target (object, group, waypoint or marker) currently under the cursor into a variable you can read:

`GETMVAR(bis_fnc_curatorObjectPlaced_mouseOver,[""])`

I believe the form is `["TYPE", data]` (e.g. `["OBJECT", objNull]`)

SilentSpike:
To expand this issue further, we could actually allow the getModuleDestination function to take an array of allowed types. That way you could also target a group.

chris579:
This could be integrated quite easily. For the callback I would return both the position and the object under the cursor.

Mike-MF:
Closing due to age, inactivity or already solved. May be re-opened if the issue is still present, even better just open a new up-to-date Issue.

PR #5528, opened 2017-09-18
Add "Load Into Cargo" Zeus Module

654wak654:
- Add a utility module that allows loading of objects into ace cargo.
- Remove a really old "load cargo" icon from cargo component. Helicopters and Apex brought some nice new icons regarding cargo
- Butcher out alganthe's strings for the arsenal module. It's already implied that the module will do the named action to the object, and no other modules were named that way.

Talked about this in slack sometime ago, figured I'd at least do a WIP PR to motivate myself.


addons/zeus/CfgVehicles.hpp
```diff
@@ -98,14 +98,14 @@ class CfgVehicles {
         category = QGVAR(Repair);
         displayName = CSTRING(ModuleAddSpareTrack_DisplayName);
         function = QFUNC(moduleAddSpareTrack);
-        icon = QPATHTOF(UI\Icon_Module_Zeus_Medic_ca.paa);//@todo
```
jonpas: Remove that icon as well, or is it used elsewhere?

654wak654: It is used in actual medical modules like `moduleSetMedic`. Though ace's medic cross could be replaced with files from ui_f\data too. I'll handle other outdated icons in a separate PR.

654wak654:
Done my tests, not WIP anymore. Could even stage for 3.11 _pretty please_.

addons/zeus/functions/fnc_moduleLoadIntoCargo.sqf
```diff
@@ -0,0 +1,61 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [LOGIC, [bob, kevin], true] call ace_zeus_fnc_moduleLoadIntoCargo
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
```
PabstMirror: Got popup error:
``` 
Warning Message: Variable '_thing' does not support serialization. Call 'disableSerialization' in the current script (maybe 'CBA_fnc_addBISEventHandler') if you need to use it.
```
Add the following
```
if (canSuspend) exitWith {[FUNC(moduleLoadIntoCargo), _this] call CBA_fnc_directCall;};

if (!(missionNamespace getVariable [QEGVAR(cargo,enable), false])) exitWith {
    [LSTRING(RequiresAddon)] call FUNC(showMessage);
};
```

654wak654: Should I also add it to `moduleSuppressiveFire`, because I used that as a template for this function.

PabstMirror: already in that one
https://github.com/acemod/ACE3/blob/master/addons/zeus/functions/fnc_moduleSuppressiveFire.sqf#L21

654wak654: Also when the function calls itself through `CBA_fnc_directCall` after logic is deleted, `_logic` parameter will be null for the new call so I'm going to put the first `exitWith` above the deletion.

SilentSpike:
The only thing I don't like about this is that it uses the nearest vehicle instead of explicitly targeting a vehicle. However, we can enhance that later a la #5215

PR #5529, opened 2017-09-18
Added captive hotkey

chris579:
- Add a hotkey for toggling the captive status of a unit under the cursor

I was unsure about the default hotkey for this so I chose `Shift+F1`. Maybe we should leave out the default hotkey completely for this.


jonpas:
Finally someone does this! :D

I think you have to check distance in the action as well, interaction menu takes care of it otherwise.

I also think there is no point in having a "toggle" keybind, just having the keybind for handcuffing would be better, since that's the hard part to do with the interaction menu.

addons/captives/XEH_postInit.sqf
```diff
@@ -35,6 +35,22 @@ if (isServer) then {
...
+["ACE3 Common", QGVAR(captives), [(localize LSTRING(KeyComb)), (localize LSTRING(KeyComb_description))],
+{
+    if !(cursorObject isKindOf "Man") exitWith {};
+
+    if([ACE_player, cursorObject] call FUNC(canApplyHandcuffs)) exitWith {
+        [cursorObject, true] call FUNC(setHandcuffed);
+    };
+
+    if([ACE_player, cursorObject] call FUNC(canRemoveHandcuffs)) exitWith {
+        [cursorObject, false] call FUNC(setHandcuffed);
```
commy2: Don't use `cursorObject`. It has really bad accuracy.

addons/captives/XEH_postInit.sqf
```diff
@@ -35,6 +35,22 @@ if (isServer) then {
 
 if (!hasInterface) exitWith {};
 
+    //Add Keybind:
+["ACE3 Common", QGVAR(captives), [(localize LSTRING(KeyComb)), (localize LSTRING(KeyComb_description))],
+{
+    if !(cursorObject isKindOf "Man") exitWith {};
```
commy2: CAManBase
Man includes rabbits and snakes and dog and fish etc.

addons/captives/XEH_postInit.sqf
chris579: What should I use instead of `cursorObject`?

654wak654: `screenToWorld [0.5, 0.5]` maybe?

addons/captives/stringtable.xml
```diff
@@ -428,12 +428,12 @@
             <Chinesesimp>投降或无武器状态</Chinesesimp>
         </Key>
         <Key ID="STR_ACE_Captives_KeyComb">
-            <English>Toggle unit captive</English>
-            <German>Einheit festnehmen umschalten</German>
+            <English>Set unit captive</English>
```
jonpas: You can just use the strings from above, the ones used for interaction menu.

addons/captives/XEH_postInit.sqf
```diff
@@ -35,6 +35,18 @@ if (isServer) then {
 
 if (!hasInterface) exitWith {};
 
+    //Add Keybind:
+["ACE3 Common", QGVAR(captives), [(localize LSTRING(KeyComb)), (localize LSTRING(KeyComb_description))],
+{
+    if !(cursorObject isKindOf "CAManBase" || {(cursorObject distance player) > getNumber (configFile >> "CfgVehicles" >> "CAManBase" >> "ACE_Actions" >> "ACE_ApplyHandcuffs" >> "distance")}) exitWith {};
```
PabstMirror: use `ace_player` not player

this code should also return a bool, generally false if nothing happens, and true if it will
so false if out of range, true if we run `setHandcuffed` and false at the bottom

PabstMirror: also add
```
if !([ACE_player, cursorObject, []] call EFUNC(common,canInteractWith)) exitWith {false};
```

addons/captives/XEH_postInit.sqf
```diff
@@ -1,4 +1,5 @@
 #include "script_component.hpp"
+#include “\a3\editor_f\Data\Scripts\dikCodes.h”
```
jonpas: Those are some wrong quotes.

addons/captives/XEH_postInit.sqf
```diff
@@ -36,16 +37,20 @@ if (isServer) then {
...
+["ACE3 Common", QGVAR(captives), [(localize LSTRING(SetCaptive)), (localize LSTRING(KeyComb_description))],
 {
-    if !(cursorObject isKindOf "CAManBase" || {(cursorObject distance player) > getNumber (configFile >> "CfgVehicles" >> "CAManBase" >> "ACE_Actions" >> "ACE_ApplyHandcuffs" >> "distance")}) exitWith {};
+    private _target = cursorObject;
+    if !([ACE_player, _target, []] call EFUNC(common,canInteractWith)) exitWith {false};
+    if !(_target isKindOf "CAManBase" || {(_target distance ACE_player) > getNumber (configFile >> "CfgVehicles" >> "CAManBase" >> "ACE_Actions" >> "ACE_ApplyHandcuffs" >> "distance")}) exitWith {false};
 
-    if([ACE_player, cursorObject] call FUNC(canApplyHandcuffs)) exitWith {
-        [cursorObject, true] call FUNC(setHandcuffed);
+    if([ACE_player, _target] call FUNC(canApplyHandcuffs)) exitWith {
```
jonpas: Space missing in `if (`.

addons/captives/XEH_postInit.sqf
chris579: Whoops, this happens when directly copying from the cba docu

addons/captives/XEH_postInit.sqf
jonpas: @commy2 ?

addons/captives/stringtable.xml
```diff
@@ -427,6 +427,10 @@
             <Chinese>投降或無武器狀態</Chinese>
             <Chinesesimp>投降或无武器状态</Chinesesimp>
         </Key>
+        <Key ID="STR_ACE_Captives_KeyComb_Description">
+            <English>Sets the unit under the cursor captive.</English>
+            <German>Nimmt die Einheit vor dem Cursor fest.</German>
+        </Key>¸¸¸¸
```
chris579: This points got added with the marge, didn't them?

jonpas: Woops, remove them, yeah. 😆

addons/captives/XEH_postInit.sqf
```diff
@@ -35,6 +36,22 @@ if (isServer) then {
 
 if (!hasInterface) exitWith {};
 
+    //Add Keybind:
```
jonpas: Many spaces. This comment is kind of superfluous anyways, it's seen immediately it's a keybind thing.

PabstMirror:
Forgot FUNC(setHandcuffed needs to run were target is local, pushed small change to use targetEvent

PR #5532, opened 2017-09-19
Replace UI Icons With Vanilla/ Common Ones

654wak654:
- Replace UI icons with vanilla/ common ones.

Also removes some _very_ old paa files that belong to AGM/CSE.

Reason for every change:
- Removed `apl/functions/script_component.hpp`: The apl component doesn't have any functions.
- Removed `common/UI/blank_CO.paa`: There is a vanilla blank texture with the same resolution.
- Moved `magazinerepack/UI/repack_ca.paa` to `common/UI/repack_ca.paa`: The same texture was also used in `overheating` under a different name.
- Removed `quickmount/UI/Icon_Module_QuickMount_ca.paa`: It was a texture copied from ui_f_data.
- Removed `interaction/UI/CenterIcon_ca.paa`, `interaction/UI/backArrow_ca.paa`, and `interaction/UI/command_rose_ca.paa`: They're relics of another time.


addons/quickmount/CfgVehicles.hpp
```diff
@@ -9,7 +9,7 @@ class CfgVehicles {
         isGlobal = 1;
         isTriggerActivated = 0;
         isDisposable = 0;
-        icon = QPATHTOF(UI\Icon_Module_QuickMount_ca.paa);
+        icon = "a3\ui_f\data\IGUI\Cfg\Actions\Obsolete\ui_action_getin_ca.paa";
```
bux: Obsolete?

654wak654: The non obsolete ones are all for a specific seat `getincommander`, `getindriver`, `getingunner`, `getinpilot`, `getincargo`, so I'm using the generic "arrow in box" icon which is not used by the game anymore, but in the obsolete folder since 2014.

bux: i see, thanks for explaining :)

PR #5533, opened 2017-09-19
Add Stow FRIES interaction

jonpas:
- Allow stowing FRIES (retracting the hooks, closing doors) without the need to deploy and cut ropes, useful for when you pull out of a fastroping position before deploying the ropes, aka so you don't get shot in the ass, but rather have the door get shot for you.


alganthe:
I couldn't break it, so it works.
I did notice that when cutting ropes the FRIES automatically get stowed, maybe change that behavior now that we can do it manually ?

jonpas:
> I did notice that when cutting ropes the FRIES automatically get stowed, maybe change that behavior now that we can do it manually ?

Could be done, I don't really have a preference. @BaerMitUmlaut ?

BaerMitUmlaut:
No real preference. Can see how both are useful:
- automatic stowing means less fiddling with the system
- manual stowing could be used when a rope breaks off, so you redeploy it without stowing

jonpas:
> manual stowing could be used when a rope breaks off, so you redeploy it without stowing

That makes me like the manual way more.

What would happen if only 1 rope breaks, just can't deploy until the other one is also broken, correct?

jonpas:
@alganthe could use another trying-everything-to-break-it test.

BaerMitUmlaut:
>What would happen if only 1 rope breaks, just can't deploy until the other one is also broken, correct?
Rope gets marked as occupied, other rope will have to be cut too to redeploy. You'd have to retract the FRIES to attach new ropes. I'm not sure how realistic it is to just redeploy with the FRIES outside.

>Is that because of some limitation?
Yes, helper objects and the ropes need to get deleted.

alganthe:
tested again, usability wise it's better now, it you cut the rope or they get shot you don't have to wait for the animations to play.

jonpas:
> I'm not sure how realistic it is to just redeploy with the FRIES outside.

I think it's fine with the limitations Arma gives us. I still want having to pack the ropes into a bag to use them. ^^

654wak654:
Maybe for a different PR, but could we also add a "FRIES backpack" so people can add/ remove fries to/ from helicopters from the interact menu.

BaerMitUmlaut:
I'd rather not add that. They're usually rather large metal constructions that require modifications to the helicopter.

jonpas:
Which is why it makes sense to be part of `ace_pylons`. Could even make FRIES a pylon attachment, but that would break backwards compatibility and is probably not worth it. :P

If anything, a "ropes backpack" that would have to be loaded into the helicopter is something that I think we all liked.

PR #5534, opened 2017-09-19
Block adv. throwing when player is busy

Dystopian:
- add new `blockThrow` status effect to block ~both vanilla and~ advanced throwing;
- add this status effect to appropriate places;
- disable effect sending more than once.

To check:
- [x] concertina_wire
- [x] tripod

To discuss:
- [x] when launcher is in hands (I think throw should be blocked);
- [x] when dragging/refueling **vanilla throw** shouldn't be blocked (one hand is free and can somehow throw) (I think it isn't worth it).

~WIP~


BaerMitUmlaut:
>one hand is free and can somehow throw

You can't pull the pin with your teeth, no matter what Hollywood tells you.

Dystopian:
but you can do it with the busy hand thumb, can't you? (But why do I argue, it's harder for me to implement it)

ColdEvul:
If GVAR(oneThumbAvaliableTask) then {...}

Dystopian:
done

addons/common/XEH_postInit.sqf
```diff
@@ -55,6 +56,23 @@
     _vehicle setVariable [QGVAR(blockEngine), _set > 0, true];
     _vehicle engineOn false;
 }] call CBA_fnc_addEventHandler;
+[QGVAR(blockThrow), {
+    params ["_object", "_set"];
+    if (_set > 0) then {
+        GVAR(blockThrow_EH) = ["KeyDown", {(_this select 1) in actionKeys "Throw"}] call CBA_fnc_addDisplayHandler;
```
jonpas: Does this work if you bind it to a mouse (or a controller/joystick)?

Dystopian: no naturally

addons/common/XEH_postInit.sqf
```diff
@@ -19,11 +19,12 @@
 
 //Status Effect EHs:
 [QGVAR(setStatusEffect), {_this call FUNC(statusEffect_set)}] call CBA_fnc_addEventHandler;
-["forceWalk", false, ["ACE_SwitchUnits", "ACE_Attach", "ACE_dragging", "ACE_Explosives", "ACE_Ladder", "ACE_Sandbag", "ACE_refuel", "ACE_rearm", "ACE_dragging"]] call FUNC(statusEffect_addType);
+["forceWalk", false, ["ACE_SwitchUnits", "ACE_Attach", "ACE_dragging", "ACE_Explosives", "ACE_Ladder", "ACE_Sandbag", "ACE_refuel", "ACE_rearm", "ACE_Trenches"]] call FUNC(statusEffect_addType);
```
Dystopian: @PabstMirror Is it correct to delete the 2nd `ACE_dragging` here? Honestly I didn't manage to understand `statusEffect` types logic, but from what I understood it seems correct.

addons/common/XEH_postInit.sqf
```diff
@@ -55,6 +56,24 @@
     _vehicle setVariable [QGVAR(blockEngine), _set > 0, true];
     _vehicle engineOn false;
 }] call CBA_fnc_addEventHandler;
+[QGVAR(blockThrow), {
+    params ["_object", "_set"];
+    TRACE_2("blockThrow EH",_object,_set);
+    if (_set > 0) then {
+        GVAR(blockThrow_EH) = ["KeyDown", {(_this select 1) in actionKeys "Throw"}] call CBA_fnc_addDisplayHandler;
```
Dystopian: Due to what @jonpas said about joystick and how `actionKeys` handles keys with modifiers, it looks like I have to change vanilla throw blocking way. Maybe the best is not to block vanilla throw but advanced throw only. Will think some time about it.

Dystopian:
Vanilla throw blocking is deleted. I don't think it's worth it to implement vanilla blocking with grenade removing from gear (looks like it's the only way to achieve this).
At least Advanced Throwing will not break other actions hint now.

addons/common/XEH_postInit.sqf
```diff
@@ -360,6 +376,12 @@ addMissionEventHandler ["PlayerViewChanged", {
     }, {}, [], 3, {TRACE_1("timeout",_this);}] call CBA_fnc_waitUntilAndExecute;
 }];
 
+// block throw when launcher is in hands
+["weapon", {
+    params ["_unit", "_weapon"];
+    [_unit, "blockThrow", QUOTE(ADDON), _weapon == secondaryWeapon _unit] call FUNC(statusEffect_set);
```
PabstMirror: This will return true when player has no weapon ("" == "")

I also don't like this calling `statusEffect_set` as there are global setVars
We can just directly check this in advThrow/canPrepare

* basically everything else is fairly rare, so it won't really matter, but weapon switching is probably the most common thing to happen

addons/common/functions/fnc_statusEffect_set.sqf
```diff
@@ -62,9 +61,13 @@ if (_set isEqualTo (_effectBoolArray select _statusIndex)) exitWith {
...
-_effectNumber = _effectBoolArray call FUNC(toBitmask); //Convert array back to number
+private _newEffectNumber = _effectBoolArray call FUNC(toBitmask); //Convert array back to number
 
-TRACE_2("Saving globaly",_effectVarName,_effectNumber);
-_object setVariable [_effectVarName, _effectNumber, true];
+TRACE_2("Saving globally",_effectVarName,_newEffectNumber);
+_object setVariable [_effectVarName, _newEffectNumber, true];
 
-[_object, _effectName] call FUNC(statusEffect_sendEffects);
+if (_effectNumber == 0 || {_newEffectNumber == 0}) then {
```
PabstMirror: I'm 99% sure this is a safe optimzation
All of our use is `_set > 0` or `_set == 0`
so going from forceWalk _set=1 [missionSet] to _set=3 [missionSet + dragging] would be the same

In theory someone could be doing something weird and handling the effect based on the number of sources
But I doubt it

addons/common/XEH_postInit.sqf
```diff
@@ -55,6 +56,21 @@
     _vehicle setVariable [QGVAR(blockEngine), _set > 0, true];
     _vehicle engineOn false;
 }] call CBA_fnc_addEventHandler;
+[QGVAR(blockThrow), {
+    params ["_object", "_set"];
+    TRACE_2("blockThrow EH",_object,_set);
```
PabstMirror: I really don't like messing with the setting var directly

I would move this to advThrowing with
```
GVAR(allowed) = true;
[QEGVAR(common,blockThrow), {
    params ["_object", "_set"];
    TRACE_2("blockThrow EH",_object,_set);

    GVAR(allowed) = (_set <= 0);
```

and then in canPrepare.sqf check `GVAR(enabled) && GVAR(allowed)`

addons/common/XEH_postInit.sqf
PabstMirror: good catch, they should only be listed once
it's doing `_statusReasons find _ID` to get the index

addons/common/XEH_postInit.sqf
PabstMirror: even better, just check `(ace_player getVariable ["ace_common_effect_blockthrow", 0]) <= 0` in can prepare

addons/common/XEH_postInit.sqf
Dystopian: I chose `statusEffect` framework exactly because it allows to set effect from several sources at the same time. If we directly check variables then they can be overwritten when 2 sources are changed at the same time.
Though launcher check is really better to check in `advThrow`.
Maybe add `global` parameter to `FUNC(statusEffect_set)` (true by default) and use it here for all calls?

PabstMirror:
rpg-26 is only ~3 kg, so you might be able to throw with something like that, but it would be awakward

PR #5544, opened 2017-09-22
Fix Load Patient/Captive actions

### LINKED ISSUE #5525: Cannot unload injured from vehicle
**Arma 3 Version:** `1.76,142872` (stable)
**CBA Version:** `3.4.1.170912` (stable)
**ACE3 Version:** `medical-rewrite bdea0fcdbab53d4d5a5d67652f41d37733523447` (dev + commit hash)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- You always get error there is no space if you try to unload injured player from vehicle.
- inconsistent behaviour when trying to unload AI

**Steps to reproduce:**
- Create mission with two slots and one vehicle
- Start two instances of Arma3
- Host the mission on one instance, join with the other
- Shoot the player until he falls
- Load injured player into vehicle
- Mount the vehicle and try to unload the player
- Or try to unload the player from outside of vehicle

-  Self-Hosted Multiplayer

**Placed Modules:**
- none

**RPT log file:**
- ([gist](https://gist.github.com/f7d285d4b6e2f5d2d922a44292cfdf4d))

jonpas:
- Disable "Load Patient" action if unit is dead but was unconscious before (we don't want loading of dead units do their weirdness) - `ACE_isUnconscious` flag doesn't get reset on death
- Remove handcuffs on death to prevent ability to remove them later which causes the unit to stand up and run the ragdoll again (looks quite hilarious though)
- Change "Load Captive" to show only if unit is not unconscious ("Load Patient" and "Load Captive" do the same thing and it's easier to notice someone is unconscious)
- Disable "Unload Patient" action when inside a vehicle - fix #5525 (more of a bandaid, but it makes more sense anyways)

### DIFF
--- a/addons/captives/CfgEventHandlers.hpp
+++ b/addons/captives/CfgEventHandlers.hpp
@@ -63,3 +63,11 @@ class Extended_Local_EventHandlers {
         };
     };
 };
+
+class Extended_Killed_EventHandlers {
+    class CAManBase {
+        class ADDON {
+            killed = QUOTE(_this call FUNC(handleKilled));
+        };
+    };
+};
--- a/addons/captives/XEH_PREP.hpp
+++ b/addons/captives/XEH_PREP.hpp
@@ -1,4 +1,3 @@
-
 PREP(canApplyHandcuffs);
 PREP(canEscortCaptive);
 PREP(canFriskPerson);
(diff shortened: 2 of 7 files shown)


jonpas:
Ready.

PR #5548, opened 2017-09-24
Add ace_common_fnc_getPylonTurret func

PabstMirror:
Needed for ace_pylons and rearm
It can fail if aircraft is configured with same weapon on pilot and gunner but those configurations don't really make any sense and probably won't be an issue.

### DIFF
--- a/addons/common/functions/fnc_getPylonTurret.sqf
+++ b/addons/common/functions/fnc_getPylonTurret.sqf
@@ -1,13 +1,13 @@
 /*
  * Author: PabstMirror
- * Finds turret owner of a pylon
+ * Finds turret owner of a pylon.
  *
  * Arguments:
  * 0: Vehicle <OBJECT>
  * 1: Pylon Index (starting at 0) <NUMBER>
  *
  * Return Value:
- * <ARRAY> (turret Index, either [-1] or [0])
+ * * Turret index (either [-1] or [0]) <ARRAY>
  *
  * Example:
  * [cursorObject, 0] call ace_common_fnc_getPylonTurret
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_getPylonTurret.sqf
```diff
@@ -0,0 +1,64 @@
+/*
+ * Author: PabstMirror
+ * Finds turret owner of a pylon
```
jonpas: Missing final dot.

addons/common/functions/fnc_getPylonTurret.sqf
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Arguments:
+ * 0: Vehicle <OBJECT>
+ * 1: Pylon Index (starting at 0) <NUMBER>
+ *
+ * Return Value:
+ * <ARRAY> (turret Index, either [-1] or [0])
```
jonpas: ` * Turret index (either [-1] or [0]) <ARRAY>`

addons/common/functions/fnc_getPylonTurret.sqf
```diff
@@ -0,0 +1,64 @@
...
+#include "script_component.hpp"
+
+params ["_vehicle", "_pylonIndex"];
+
+// See if index is in ace_pylonTurrets setVar on vehicle
+private _pylonTurrets = _vehicle getVariable ["ace_pylonTurrets", []];
+private _returnValue = _pylonTurrets param [_pylonIndex, []];
+
+if (!(_returnValue isEqualTo [])) then {
+    TRACE_1("Using ace_pylonTurrets value",_savedValue);
```
Dystopian: savedValue?

addons/common/functions/fnc_getPylonTurret.sqf
```diff
@@ -0,0 +1,64 @@
...
+} else {
+    // Attempt to determine turret owner based on magazines in the vehicle
+    private _pyMags = getPylonMagazines _vehicle;
+    private _pylonConfigs = configProperties [configFile >> "CfgVehicles" >> typeOf _vehicle >> "Components" >> "TransportPylonsComponent" >> "Pylons", "isClass _x"];
+    if (_pylonIndex >= (count _pyMags)) exitWith {ERROR("out of bounds"); []};
```
Dystopian: no need to return `[]` inside this if-else because it exits to big if-else.
IMO it's better to remove big if-else above (`if (!(_returnValue isEqualTo []))`) and use `exitWith` there.

jonpas: This is nicer for debugging due to TRACE in the if-then.

PR #5558, opened 2017-09-26
Toggle modules standardization

alganthe:
- Change the message type to use the standard zeus message instead of ace hint.
- Delete the module on dialog unload.


jonpas:
`fnc_ui_ToggleNvg.sqf`, this file also has weird capitalization.

PR #5571, opened 2017-09-29
Add CBA setting for noRadio

### LINKED ISSUE #2346: disable noVoice
Hi all,

is it possible to disable the "NoRadio" Funktion? I need de AI and my own Player to use the radio function. But it seems like this function ist in the basic code of ace. Is it possible to disable the noradio function?

Thanks alot!

PabstMirror:
Close #2346


AACO:
Just curious, why is the DC event handler added, even if the module is disabled? It's not going to fire a lot, but I'm always a fan of not adding handlers if the component is disabled.

jonpas:
To support changing it on the fly. More and more things that don't it like that will likely be converted as we start manually moving components to CBA Settings. It's not any overhead really on an event handler like that. For things that execute more often it might be wiser to add/remove handler as required.

AACO:
@jonpas the main player changed EH won't get added if it's changed on the fly. I guess I don't see how handling these two items separately will add (besides confusion ;) )

 I'm not sure how ACE will convert to CBA settings, but (as you know) CBA settings can have a chunk of code executed on change, which should allow adding/removing these kinds of event handlers without too much pain (only problem will probably come from handling class event handlers, and the messiness of having to track the EHIDs through GVARs).

jonpas:
Ah I didn't notice that, good point!

And yes, that's why I said it will probably only be used for performance heavy EHs. We'll see how we'll pursue that, for starters ACE Settings will be converted on the fly to CBA Settings, there will definitely be a post going live with the update about that.

commy2:
I'd like to make this the first CBA setting and have it support changing it on the fly :)

PabstMirror:
I think we can pick this up for 3.12
It's new setting so there should be no bwc to worry about

PR #5578, opened 2017-10-01
advanced throwing spin grenades

### LINKED ISSUE #5567: Advanced Throwing throwable torque
Make grenades rotates in air, similar or better than in vanilla. Especially useful for WW2 German stick grenades. However some grenades have air stabilizer fins, so have to support both.

Opening this to discuss config entry requirements, so far thought of this (with ideas from below comments):
```cpp
class CfgAmmo {
    class SmokeShell;
    class ACE_G_Handflare_White: SmokeShell {
        // {} or {0,0,0} - disabled, {x, y, z} - torque vector
        ace_advanced_throwing_rotation = {100, 0, 0};
    };
};
```

It would also be enabled for all vanilla grenades with some default rotation, since they are all similar enough.

I didn't do any testing yet, but `addTorque` is applied as impulse, which means it should stop rotating relatively fast on very low torque vectors, thus simulating stabilizer fins.

commy2:
- title
- fix #5567

### DIFF
--- a/addons/advanced_throwing/XEH_postInit.sqf
+++ b/addons/advanced_throwing/XEH_postInit.sqf
@@ -3,6 +3,18 @@
 // Fired XEH
 [QGVAR(throwFiredXEH), FUNC(throwFiredXEH)] call CBA_fnc_addEventHandler;
 
+// make grenades spin
+["ace_throwableThrown", { // local event
+    params ["_unit", "_grenade"];
+
+    private _config = _grenade call CBA_fnc_getObjectConfig;
+    systemChat str [_config, getArray (_config >> QGVAR(torqueDirection)), getNumber (_config >> QGVAR(torqueMagnitude))];
+    private _torque = vectorNormalized getArray (_config >> QGVAR(torqueDirection)) vectorMultiply
+        getNumber (_config >> QGVAR(torqueMagnitude));
+
+    _grenade addTorque (_unit vectorModelToWorld _torque);
+}] call CBA_fnc_addEventHandler;
+
 // Exit on HC
 if (!hasInterface) exitWith {};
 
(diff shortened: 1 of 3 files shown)


addons/advanced_throwing/XEH_postInit.sqf
```diff
@@ -3,6 +3,17 @@
 // Fired XEH
 [QGVAR(throwFiredXEH), FUNC(throwFiredXEH)] call CBA_fnc_addEventHandler;
 
+// make grenades spin
+["ace_throwableThrown", { // local event
+    params ["_unit", "_grenade"];
+
+    private _config = _grenade call CBA_fnc_getObjectConfig;
+    private _torque = vectorNormalized getArray (_config >> QGVAR(torqueDirection)) vectorMultiply
+        getNumber (_config >> QGVAR(torqueMagnitude));
```
PabstMirror: This works fine.
Has getNumber always worked on strings or was this something recent?
I thought I remember having to do some `call compile getText` to evaluate stuff like this.

commy2: It worked like this since A2.

jonpas:
I already did that: https://github.com/acemod/ACE3/compare/advThrowingRotation :P

I like this implementation more in terms of config entries, I didn't really think of randomness. However, I'd prefer to do `addTorque` directly in `fnc_throw.sqf`, attaching to an event in same component where we throw it looks weird and separates code for no reason.

commy2:
I really like the modularity of this, but I guess if everything was made this way, a third of the computation time would go into frameworks.

jonpas:
It's 3 lines, I think it's more readable and concise if in the function directly.

commy2:
Put it in https://github.com/acemod/ACE3/compare/master...advThrowingRotation

PR #5581, opened 2017-10-03
weaponSelect CBA Setting

commy2:
This is an easy one.


PabstMirror:
Not really related to this PR, just something I saw in testing
```
[
    "test1", "CHECKBOX",
    ["test1", LSTRING(SettingDisplayTextDesc)],
     localize "STR_ACE_common_ACEKeybindCategoryWeapons"
] call CBA_settings_fnc_init;

[
    "test2", "CHECKBOX",
    ["test2"],
    "STR_ACE_common_ACEKeybindCategoryWeapons"
] call CBA_settings_fnc_init;
```

Running both of these produces weird results.
It opens on the localized category with only test1
Category lists both "ACE3 Weapons" and "STR_ACE_common_ACEKeybindCategoryWeapons"
The "STR" has nothing in it, reselecting the localized one then shows both options.

I'm thinking we should just have CBA localize categories on init so there's no problems on category filtering?

commy2:
lgtm, can't approve what I opened.

addons/weaponselect/initSettings.sqf
```diff
@@ -0,0 +1,8 @@
+// CBA Settings [ADDON: ace_weaponselect]:
+
+[QGVAR(displayText), "CHECKBOX",
+[LSTRING(SettingDisplayTextName), LSTRING(SettingDisplayTextDesc)],
+"ACE Uncategorized",
```
SilentSpike: Is there some sort of inventory category this would be more suited to?

commy2:
UI

PabstMirror:
Requires #5752

PR #5585, opened 2017-10-04
Cook-off probability and cook-off tweaks

tbeswick96:
- Add cook-off probability for vehicles. Configurable via config value `ace_cookoff_cookoffProbability` for tanks and APCs. ACE setting `ace_cookoff_probabilityCoef ` is used as a multiplier for the probability. Higher values mean cook-off is more likely.
- Probability values (up for discussion, figured tanks less likely to cook-off):
  - `Tank_F: 0.5 (50%)`
  - `Wheeled_APC_F: 0.8 (80%)`
  - `ace_cookoff_probabilityCoef 1 (multiplier)`
- Tweak the way cook-off works somewhat (major impact on end-effect)
  - Increased damage threshold for other hit locations on tanks, but a high enough hit will cause instant destruction. This means high-damage munitions (Javelin, Hellfire, GBU, etc) are capable of instantly destroying heavy armour like they should be able to.

### DIFF
--- a/addons/cookoff/ACE_Settings.hpp
+++ b/addons/cookoff/ACE_Settings.hpp
@@ -24,10 +24,10 @@ class ACE_Settings {
         value = 1;
         typeName = "SCALAR";
     };
-    class GVAR(probability) {
-        displayName = CSTRING(probability_name);
-        description = CSTRING(probability_tooltip);
-        value = 0.4;
+    class GVAR(probabilityCoef) {
+        displayName = CSTRING(probabilityCoef_name);
+        description = CSTRING(probabilityCoef_tooltip);
+        value = 1;
         typeName = "SCALAR";
     };
 };
--- a/addons/cookoff/ACE_Settings.hpp
+++ b/addons/cookoff/ACE_Settings.hpp
@@ -24,9 +24,9 @@ class ACE_Settings {
         value = 1;
         typeName = "SCALAR";
     };
-    class GVAR(cookoffProbability) {
-        displayName = CSTRING(cookoffProbability_name);
-        description = CSTRING(cookoffProbability_tooltip);
+    class GVAR(probability) {
+        displayName = CSTRING(probability_name);
+        description = CSTRING(probability_tooltip);
         value = 0.4;
         typeName = "SCALAR";
     };
--- a/addons/cookoff/ACE_Settings.hpp
+++ b/addons/cookoff/ACE_Settings.hpp
@@ -24,4 +24,10 @@ class ACE_Settings {
         value = 1;
         typeName = "SCALAR";
     };
+    class GVAR(cookoffProbability) {
+        displayName = CSTRING(cookoffProbability_name);
+        description = CSTRING(cookoffProbability_tooltip);
+        value = 0.4;
+        typeName = "SCALAR";
+    };
 };
--- a/addons/cookoff/functions/fnc_handleDamage.sqf
+++ b/addons/cookoff/functions/fnc_handleDamage.sqf
@@ -69,7 +69,7 @@ if (_simulationType == "tank") exitWith {
         // get cookoff probability for vehicle
         private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(probability), "number", 0.7] call CBA_fnc_getConfigEntry;
         // probability multiplied by coef for global probability control (higher coef = more probable cookoff)
-        if (_damage > 0.1 && {random 1 < (_probability * GVAR(probabilityCoef))}) then {
+        if (_damage > 0.5 && {random 1 < (_probability * GVAR(probabilityCoef))}) then {
             _vehicle call FUNC(cookOff);
         };
     } else {
--- a/addons/cookoff/functions/fnc_handleDamage.sqf
+++ b/addons/cookoff/functions/fnc_handleDamage.sqf
@@ -67,7 +67,7 @@ if (_simulationType == "tank") exitWith {
     // ammo was hit, high chance for cook-off
     if (_hitpoint == _ammoLocationHitpoint) then {
         // get cookoff probability for vehicle (default to global setting if not found)
-        private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(cookoffProbability), "number", GVAR(cookoffProbability)] call CBA_fnc_getConfigEntry;
+        private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(probability), "number", GVAR(probability)] call CBA_fnc_getConfigEntry;
         if (_damage > 0.1 && {random 1 < _probability}) then {
             _vehicle call FUNC(cookOff);
         };
--- a/addons/cookoff/functions/fnc_handleDamage.sqf
+++ b/addons/cookoff/functions/fnc_handleDamage.sqf
@@ -66,17 +66,18 @@ if (_simulationType == "tank") exitWith {
 
     // ammo was hit, high chance for cook-off
     if (_hitpoint == _ammoLocationHitpoint) then {
-        // get cookoff probability for vehicle
-        private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(probability), "number", 0.7] call CBA_fnc_getConfigEntry;
-        // probability multiplied by coef for global probability control (higher coef = more probable cookoff)
-        if (GVAR(probabilityCoef) > 1) then {
-            _probability = 1 - (1 - _probability) / GVAR(probabilityCoef);
-        } else {
-            _probability = _probability * GVAR(probabilityCoef);
-        };
-
-        if (_damage > 0.5 && {random 1 < _probability}) then {
-            _vehicle call FUNC(cookOff);
+        if (_damage > 0.5) then {
+            // get cookoff probability for vehicle
+            private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(probability), "number", 0.7] call CBA_fnc_getConfigEntry;
+            // probability multiplied by coef for global probability control (higher coef = more probable cookoff)
+            if (GVAR(probabilityCoef) > 1) then {
+                _probability = 1 - (1 - _probability) / GVAR(probabilityCoef);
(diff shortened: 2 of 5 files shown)


addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -66,17 +66,23 @@ if (_simulationType == "tank") exitWith {
 
     // ammo was hit, high chance for cook-off
     if (_hitpoint == _ammoLocationHitpoint) then {
-        if (_damage > 0.5 && {random 1 < 0.7}) then {
+        // get cookoff probability for vehicle (default to global setting)
+        private _probability = if (isNumber (_vehicle call CBA_fnc_getObjectConfig >> QGVAR(cookoffProbability))) then {
```
Dystopian: `private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(cookoffProbability), "number", GVAR(cookoffProbability)] call CBA_fnc_getConfigEntry;`

addons/cookoff/ACE_Settings.hpp
```diff
@@ -24,4 +24,10 @@ class ACE_Settings {
         value = 1;
         typeName = "SCALAR";
     };
+    class GVAR(cookoffProbability) {
```
Dystopian: maybe just `GVAR(probability)`?

addons/cookoff/functions/fnc_handleDamage.sqf
tbeswick96: Nice

addons/cookoff/ACE_Settings.hpp
tbeswick96: Yes, good idea

addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -66,17 +66,19 @@ if (_simulationType == "tank") exitWith {
 
     // ammo was hit, high chance for cook-off
     if (_hitpoint == _ammoLocationHitpoint) then {
-        if (_damage > 0.5 && {random 1 < 0.7}) then {
+        // get cookoff probability for vehicle (default to global setting if not found)
+        private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(cookoffProbability), "number", GVAR(cookoffProbability)] call CBA_fnc_getConfigEntry;
```
Dystopian: `cookoffProbability` -> `probability` :)

tbeswick96: *facepalm*

addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -66,17 +66,19 @@ if (_simulationType == "tank") exitWith {
 
     // ammo was hit, high chance for cook-off
     if (_hitpoint == _ammoLocationHitpoint) then {
-        if (_damage > 0.5 && {random 1 < 0.7}) then {
+        // get cookoff probability for vehicle (default to global setting if not found)
+        private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(probability), "number", GVAR(probability)] call CBA_fnc_getConfigEntry;
```
PabstMirror: I think the ace setting will always be ignored because it will just use the vehicle's config first

tbeswick96: Maybe could use the setting as a multiplier instead? So setting the setting higher results in more common cook-off etc.

BlueNexus:
Status on this?

PabstMirror:
probability stuff looks ok now
not sure about changes to `hitfuel`, I thought that hitpoint was critical to prevent vanilla explosion
 @commy2 would know more

tbeswick96:
I'll revert the `hitFuel` locally and do tests against how it behaves now to see

tbeswick96:
Tested a variety of vehicles and munitions, and didn't notice any change in behaviour with `hitFuel` in that check or not

tbeswick96:
Have reverted the removal of `hitFuel` from the destruction blocking hitpoint check, as in my testing I found it made no difference either way.

tbeswick96:
## Test:
- `10 T100 Tanks & 10 BM-2T APCs`
- `1x Merkava`
- 1 shot AP round into the same place on each vehicle
- Count how many cook-off

#### Results Before Changes:
- `8/10 T100 Cook-off`
- `10/10 BM-2T Cook-off`

#### Results After Changes (Coef 1):
- `5/10 T100 Cook-off`
- `9/10 BM-2T Cook-off`

#### Results After Changes (Coef 0.5):
- `2/10 T100 Cook-off`
- `5/10 BM-2T Cook-off`

#### Results After Changes (Coef 2):
- `9/10 T100 Cook-off`
- `10/10 BM-2T Cook-off`

tbeswick96:
## Test:
- `1 T100 Tank & 1 BM-2T APC`
- `2x Javelin (3CB with ACE Guidance), 2x NLAW (Vanilla with ACE Guidance, Direct & Overfly)`
- 1 shot Javelin round into each vehicle, top-down mode

#### Results Before Changes:
- `Javelin T100 Cook-off`
- `NLAW Direct T100 Nothing`
- `NLAW Overfly T100 Nothing`
- `Javelin BM-2T Cook-off`
- `NLAW Direct BM-2T Nothing`
- `NLAW Overfly BM-2T Cook-off`

#### Results After Changes:
- `Javelin T100 Instant destruction`
- `NLAW Direct T100 Nothing`
- `NLAW Overfly T100 Cook-off`
- `Javelin BM-2T Instant destruction`
- `NLAW Direct BM-2T Nothing`
- `NLAW Overfly BM-2T Instant destruction`

tbeswick96:
This is ready for review

addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -66,12 +66,15 @@ if (_simulationType == "tank") exitWith {
...
+        private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(probability), "number", 0.7] call CBA_fnc_getConfigEntry;
+        // probability multiplied by coef for global probability control (higher coef = more probable cookoff)
+        if (_damage > 0.5 && {random 1 < (_probability * GVAR(probabilityCoef))}) then {
             _vehicle call FUNC(cookOff);
         };
     } else {
-        if (_hitpoint in ["hithull", "hitturret", "#structural"] && {_newDamage > 0.6 + random 0.3}) then {
-            _vehicle call FUNC(cookOff);
+        if (_hitpoint in ["hithull", "hitturret", "#structural"] && {_newDamage > 0.8 + random 0.2}) then {
+            _vehicle setDamage 1;
```
PabstMirror: Why change `call FUNC(cookOff);` to `setDamage`
`FUNC(cookOff)` eventualy calls setDamage, after a delay, this makes it instant

ulteq:
I would assume that `probabilityCoef = 1.5` is supposed to increase the chance of triggering a 'Cook Off' event by 50%. Right?

Let us now assume that the vehicle has a `probability` of `0.7` and our `probabilityCoef` is set to `1.5`:

```random 1 < _probability * GVAR(probabilityCoef)```
```random 1 < 0.7 * 1.5```
```random 1 < 1.05```
```true```

In this case we would always get a 'Cook Off' event.

**Comment by tbeswick96** [comment]
What would you expect the value to be in that example?

**Comment by ulteq** [comment]
You have a 30% chance of not triggering a 'Cook Off' event. You want to decrease that chance by 50%.
The expected value would be `0.8` in this case.

But it might be better to build around  normal distribution: `random [min, mid, max]`

**Comment by tbeswick96** [comment]
`random 1 < (1 - ((1 - probability) / probabilityCoef))`
Seems to do what you're suggested should be the case

**Comment by ulteq** [comment]
Yes, but only if `GVAR(probabilityCoef) >= 0`. Try this:

```
if (GVAR(probabilityCoef) > 1) then {
    _probability = 1 - (1 - _probability) / GVAR(probabilityCoef);
} else {
    _probability = _probability * _probabilityCoef;
};

if (random 1 < _probability) then {
    ...
};
```

File: addons/cookoff/functions/fnc_handleDamage.sqf [hunk 6]
```diff
@@ -66,12 +66,21 @@ if (_simulationType == "tank") exitWith {
...
+        // get cookoff probability for vehicle
+        private _probability = [_vehicle call CBA_fnc_getObjectConfig >> QGVAR(probability), "number", 0.7] call CBA_fnc_getConfigEntry;
+        // probability multiplied by coef for global probability control (higher coef = more probable cookoff)
+        if (GVAR(probabilityCoef) > 1) then {
+            _probability = 1 - (1 - _probability) / GVAR(probabilityCoef);
+        } else {
+            _probability = _probability * GVAR(probabilityCoef);
+        };
+
+        if (_damage > 0.5 && {random 1 < _probability}) then {
```
ulteq: You could do the `_damage > 0.5` check in line 69 for performance reasons.

SilentSpike:
I like these changes 👍 Can't speak for how well the instant destruction aspect applies, but it sounds reasonable to me (seems like the kind of change we should just push and see if feedback is positive/negative).

The only thing I could possibly see missing from this is the possibility for mission maker to control probability on individual vehicles (e.g. an objective vehicle that should always cook-off for effect, or a friendly vehicle they'd never want to cook-off, etc.). I don't think it's a necessity though and could always be added later.

addons/cookoff/stringtable.xml
```diff
@@ -106,6 +106,12 @@
             <Chinese>設定彈藥殉爆效果會持續多久時間 [輸入0來關閉殉爆效果]</Chinese>
             <Chinesesimp>设定弹药殉爆效果会持续多久时间 [输入0来关闭殉爆效果]</Chinesesimp>
             <Korean>쿡오프 지속 시간의 배수 [0 이면 비활성]</Korean>
-        </Key>
+        </Key>        
+		<Key ID="STR_ACE_CookOff_probabilityCoef_name">
```
Dystopian: tabs :)

PR #5587, opened 2017-10-04
Short R3F compat update according with the new AB drag models

Laid3acK:
- Revert to the default G7 drag model for the 5.56x45 and 7.62x51 bullets.
- Update all `airFriction ` according with the new ACE3's ones.


jonpas:
Merge after #5566?

dedmen:
Doesn't have to be after #5566 But wouldn't hurt.
No dependencies here.

jonpas:
Yes that's what I meant, it will work better after #5566, in case of any issues that might prevent merge of it (hope not!).

Laid3acK:
The R3F `airFriction` are designed to work with the vanilla ballistic (R3F SCROME scopes and BDC reticles). 
The ballistic in-game is more accurate with the old or the new ACE3 `airFriction` in all cases.
So, you can merge it when you want : now, after [#5566](https://github.com/acemod/ACE3/pull/5566) or with the ACE3 3.12.
It's not really a problem, the result in-game will be necessarily better with the ACE3 values. :)

PR #5588, opened 2017-10-04
AB : same values for the both 127x108 bullets

Laid3acK:
Accord the `typicalSpeed`, `initSpeed` and `airFriction` for the both bullets due to the difference of the ballistic trajectory in-game :

- 127x108 : `typicalSpeed = 820;` /  `initSpeed = 820;` / `airFriction = -0.00065098;`

- APDS default values: `typicalSpeed = 1060;` / `initSpeed = 1060;` / `airFriction = -0.00036;`

### DIFF
--- a/addons/ballistics/CfgAmmo.hpp
+++ b/addons/ballistics/CfgAmmo.hpp
@@ -553,6 +553,10 @@ class CfgAmmo {
         ACE_muzzleVelocities[]={820};
         ACE_barrelLengths[]={728.98};
     };
+    class B_127x108_APDS: B_127x108_Ball {
+        typicalSpeed = 820;
+        airFriction = -0.00065098;
+    };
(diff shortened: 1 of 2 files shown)


AACO:
APDS = Armor Piercing Discarding Sabot so the higher speed and lower friction makes sense because it's a smaller lighter bullet. Now, I don't know how well this translates to ArmA as much since I don't think it takes the bullet caliber/mass into account.
(also this looks like the normal ballistics component, not advanced)

Laid3acK:
It would makes sens if the 127x108 APDS had his own ACE3 entries but it's the same bullet currently and the bullet trajectory in-game should be also the same with AB enabled.
I didn't find a decent documentation about this bullet and the correct GM Lynx's muzzle velocity.
If someone has the information ... :)

PR #5589, opened 2017-10-04
[compat_rhs_usf3] Reload Launchers and Overpressure

Whigital:
* Add ace_reloadlaunchers capability and ace_overpressure config to Mk 153 SMAW launcher
* Add ace_reloadlaunchers capability and ace_overpressure config to M3 MWAAS launcher

Overpressure for the Gustaf based on the vanilla `launch_RPG32_F` values and increased a bit. SMAW got default inherited values from `Launcher_Base_F`, but with a narrower cone angle.

Let me know if the values are unacceptable and i'll scale them back a bit.

### DIFF
--- a/optionals/compat_rhs_usf3/CfgWeapons.hpp
+++ b/optionals/compat_rhs_usf3/CfgWeapons.hpp
@@ -212,7 +212,17 @@ class CfgWeapons {
         lockedTargetSound[] = {"",0,1};
     };
 
-
+    class Launcher_Base_F;
+    class rhs_weap_smaw: Launcher_Base_F {
+        ace_reloadlaunchers_enabled = 1;
+        ace_overpressure_angle = 45;
+    };
+    class rhs_weap_maaws: Launcher_Base_F {
+        ace_reloadlaunchers_enabled = 1;
+        ace_overpressure_range = 15;
+        ace_overpressure_angle = 70;
+        ace_overpressure_damage = 0.75;
+    };
 
     #define HEARING_PROTECTION_VICCREW EGVAR(hearing,protection) = 0.85; EGVAR(hearing,lowerVolume) = 0.6;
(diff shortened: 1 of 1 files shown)


Whigital:
Dug through `rhsusf_c_weapons.pbo` and it turns out both the `rhs_weap_smaw` and `rhs_weap_maaws` has `ace_overpressure` values set on their classes. Im guessing they're coming from the source (TF47 ?).

No `ace_reloadlaunchers` there though, so maybe just remove the `ace_overpressure` data from the pull ?

sargken:
Leave a note on the rhs feedback tracker, they would like to remove those most likely.

jonpas:
> Leave a note on the rhs feedback tracker, they would like to remove those most likely.

Which is just straight up said, it would be so much easier and faster if those things would be in RHS itself...

Whigital:
Hehe, yeah, sad indeed. Was surprised to find those there, only explanation it's remnants from the source.

sargken:
Yeah, I think they don't offer compatibility because it would be more for them to maintain, with their scope of the mod I can see why it would be more beneficial for ace to run the compatibility.

jonpas:
> Yeah, I think they don't offer compatibility because it would be more for them to maintain, with their scope of the mod I can see why it would be more beneficial for ace to run the compatibility.

No, there would be less for them to maintain. How many times has it happened that we HAD to update RHS compats because of ACE3 changes compared to RHS changes? Almost all changes because of ACE3 changes were non-game-breaking things. It's also them who could use ACE3 scripts instead of developing their own and save some time, and also them who know exactly how they want things working.
It would be more beneficial for RHS users if they did it, like CUP does, but such is the world we live in, can't have nice things. 😞

j0zh94:
Most recent RHS patch (0.4.3.1) has removed those lines from the MAAWs and SMAW weapons. We're not opposed to ACE or anything but it makes more sense for us to approach compatibility as an "all or nothing" approach, hence the removal. Perhaps in the future, when we have time to sit down and discuss this we can.

PR #5606, opened 2017-10-10
Added fastrope keybind

BaerMitUmlaut:
- Add a keybind to fastrope

### DIFF
--- a/addons/fastroping/XEH_postInit.sqf
+++ b/addons/fastroping/XEH_postInit.sqf
@@ -10,6 +10,8 @@
 
 // Keybinds
 ["ACE3 Vehicles", QGVAR(fastRope), localize LSTRING(Interaction_fastRope), {
+    if ((vehicle ACE_player) == player) exitWith {false};
+    if (!([ACE_player, vehicle ACE_player, []] call EFUNC(common,canInteractWith))) exitWith {false};
     if ([ACE_player, vehicle ACE_player] call FUNC(canFastRope)) then {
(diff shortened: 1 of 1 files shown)


addons/fastroping/XEH_postInit.sqf
```diff
@@ -8,6 +8,16 @@
     [FUNC(fastRopeServerPFH), 0, _this] call CBA_fnc_addPerFrameHandler;
 }] call CBA_fnc_addEventHandler;
 
+// Keybinds
+["ACE3 Vehicles", QGVAR(fastRope), localize LSTRING(Interaction_fastRope), {
+    if ([ACE_player, vehicle ACE_player] call FUNC(canFastRope)) then {
```
PabstMirror: I think we might want the following (probably don't need the in-vehicle check, but can't hurt)
```
if ((vehicle ACE_player) == player) exitWith {false};
if (!([ACE_player, vehicle ACE_player, []] call EFUNC(common,canInteractWith))) exitWith {false};
```

BaerMitUmlaut: Didn't you mean `(vehicle ACE_player) == ACE_player`?

dansta:
I don't mean to feature creep, but could it possibly be made so that helicopter pilots are able to initiate a fast roping sequence and force people out? There's always that guy in the back who can't find the menu option when scrolling or who hasn't set the fast roping keybind.

BaerMitUmlaut:
No.

PabstMirror:
oops, thanks bär

PR #5607, opened 2017-10-11
RHS 3.4 Goggles and AB Ammo compat update

dedmen:
Adds the new Oakley SI Ballistics googles and the googles integrated into the new shemags.
Also add's M855 Ammo which is the same as M855A1 because our A1 data is incorrect. But it doesn't matter that it's incorrect. The difference is too small to matter for such a short-range round.
The `rhs_ess_black` is not new in 3.4 as far as I can see. But it was missing.

### DIFF
--- a/optionals/compat_rhs_usf3/CfgAmmo.hpp
+++ b/optionals/compat_rhs_usf3/CfgAmmo.hpp
@@ -25,7 +25,7 @@ class CfgAmmo {
         ACE_dragModel = 1;
         ACE_muzzleVelocities[] = {900};
         ACE_barrelLengths[] = {736.6};
-    };    
+    };
     class B_762x54_Ball;
     class rhsusf_B_300winmag: B_762x54_Ball { // ACE_762x67_Ball_Mk248_Mod_1 (ballistics/CfgAmmo.hpp)
         ACE_caliber = 7.823;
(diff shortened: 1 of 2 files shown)


optionals/compat_rhs_usf3/CfgAmmo.hpp
```diff
@@ -40,7 +40,19 @@ class CfgAmmo {
         ACE_barrelLengths[] = {508.0, 609.6, 660.4};
     };
     class B_556x45_Ball;
-    class rhs_ammo_556x45_M855A1_Ball: B_556x45_Ball { // B_556x45_Ball (ballistics/CfgAmmo.hpp)
```
dedmen: Sometimes I wish diffing algo's were a little better.

commy2: You changed the line break.

jonpas: It's supposed to be LF. Either it was CRLF or you changed it to it, either way .gitattributes apparently doesn't catch all configurations. :/

dedmen: Everything is CRLF in my file \(o_o)/

dedmen: btw .gitattributes doesn't say that it will convert to anything.
I see editorConfig says LF.. But the .editorconfig is using CRLF itself..
I guess that's just autocrlf doing stuff...

jonpas: Yes that's `autocrlf`, Git will automatically convert to LF as by `.gitattributes`, you won't see those messages if you are using `autocrlf` because that's the point of `autocrlf`.

And yes, it obviously does say: `* text=auto`. The whole point is that it doesn't matter what your local files are using, they should always be LF on the server (GitHub), although we had an occasion in the past that it somehow slipped through, but it seemed to be because of editing directly on GitHub.

PR #5609, opened 2017-10-11
Remove legacy medical settings

BaerMitUmlaut:
- Remove legacy airway and fractures medical settings

Both have no effect and only confuse users, see Slack.

### DIFF
--- a/addons/medical/ACE_Settings.hpp
+++ b/addons/medical/ACE_Settings.hpp
@@ -49,16 +49,6 @@ class ACE_Settings {
         typeName = "SCALAR";
         value = 1;
     };
-    class GVAR(enableAirway) {
-        category = CSTRING(Category_Medical);
-        typeName = "BOOL";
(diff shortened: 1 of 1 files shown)


jonpas:
Also used (but commented out) at:
- https://github.com/acemod/ACE3/blob/ffaa195fe576625f052eeee04b1af9f3f377434f/addons/medical/functions/fnc_moduleAdvancedMedicalSettings.sqf#L28
- https://github.com/acemod/ACE3/blob/ffaa195fe576625f052eeee04b1af9f3f377434f/addons/medical/functions/fnc_handleDamage_advanced.sqf#L59

BaerMitUmlaut:
There's more. Just wanted to remove the settings for now. I could purge all occurances of airway and fractures, but that should be job of the rewrite.

jonpas:
Sure, I just did a quick search, that's what turned out.

PR #5610, opened 2017-10-11
Disable RHS openable doors

Dystopian:
- add function for exporting CfgVehicles config;
- add separate configs for driver and cargo openable doors disabling to RHS compats.

Some RHS vehicles have configured `driverDoor` and `cargoDoors` parameters. If use `GetIn*` actions on them (like `GetInGunner` or `GetInCargo`) the driver door (hatch) is opened. It's especially bad in MP with driver in.

`cargoDoors` parameter gives sometimes (I don't really know when) bug with `GetInCargo`: unit gets stuck in standing animation and can't move anymore until mission restart.

It looks like it's BI's fault, maybe that's why they don't have any vehicles with configured openable doors.

ACE quickmount uses exactly those actions. So we have to disable bugged openable doors to use quickmount.

CUP already removed those parameters from their vehicles (see http://dev.cup-arma3.org/T2200).

### DIFF
--- a/optionals/compat_rhs_afrf3/nodoors/CfgVehicles.hpp
+++ b/optionals/compat_rhs_afrf3/nodoors/CfgVehicles.hpp
@@ -1,0 +1,112 @@
+class CfgVehicles {
+    class MRAP_02_base_F;
+    class Offroad_01_base_F;
+    class Truck_F;
+    class O_Truck_02_covered_F;
+    class Heli_Attack_02_base_F;
+    class RHS_Mi8_base;
+    class RHS_Mi8mt_vvs;
+    class RHS_Mi8MTV3_vvs;
+    class Tank_F;
+    class rhs_bmp1_vdv;
+    class rhs_bmp2_vdv;
+    class Wheeled_APC_F;
+    class APC_Tracked_02_base_F;
+
+    class rhs_tigr_base: MRAP_02_base_F {
+        driverDoor = "";
+        cargoDoors[] = {};
+    };
+    class RHS_UAZ_Base: Offroad_01_base_F {
+        driverDoor = "";
+        cargoDoors[] = {};
+    };
+    class rhs_uaz_open_Base: RHS_UAZ_Base {
+        driverDoor = "";
+        cargoDoors[] = {};
+    };
+    class RHS_Ural_BaseTurret: Truck_F {
+        driverDoor = "";
+        cargoDoors[] = {};
+    };
+    class rhs_kamaz5350: O_Truck_02_covered_F {
+        driverDoor = "";
+        cargoDoors[] = {};
+    };
+    class RHS_Mi24_base: Heli_Attack_02_base_F {
+        driverDoor = "";
+        cargoDoors[] = {};
+    };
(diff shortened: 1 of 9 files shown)


optionals/compat_rhs_usf3/nodoors/config.cpp
```diff
@@ -0,0 +1,18 @@
+#include "\z\ace\addons\compat_rhs_usf3\script_component.hpp"
+#undef COMPONENT
+#define COMPONENT compat_rhs_usf3_nodoors
```
commy2: ?

optionals/compat_rhs_afrf3/nodoors/config.cpp
```diff
@@ -0,0 +1,18 @@
+#include "\z\ace\addons\compat_rhs_afrf3\script_component.hpp"
+#undef COMPONENT
+#define COMPONENT compat_rhs_afrf3_nodoors
```
commy2: ?

optionals/compat_rhs_afrf3/nodoors/CfgVehicles.hpp
```diff
@@ -0,0 +1,112 @@
...
+    class O_Truck_02_covered_F;
+    class Heli_Attack_02_base_F;
+    class RHS_Mi8_base;
+    class RHS_Mi8mt_vvs;
+    class RHS_Mi8MTV3_vvs;
+    class Tank_F;
+    class rhs_bmp1_vdv;
+    class rhs_bmp2_vdv;
+    class Wheeled_APC_F;
+    class APC_Tracked_02_base_F;
```
commy2: I'd rather have the base classes sorted before their child classes.

optionals/compat_rhs_usf3/nodoors/config.cpp
Dystopian: https://github.com/acemod/ACE3/blob/c8255960d619df3456f4c3ac78967ae5e09cfb3e/optionals/compat_rhs_usf3/subConfig/config.cpp#L3

optionals/compat_rhs_afrf3/nodoors/CfgVehicles.hpp
Dystopian: it's generated config and sorting your way is not really necessary. I you insist I can implement it, but I think it's not worth it. The same is in [nouniformrestrictions](https://github.com/acemod/ACE3/blob/master/optionals/nouniformrestrictions/CfgVehicles.hpp) component.

optionals/compat_rhs_afrf3/nodoors/config.cpp
Dystopian: if you have a better way, please say

Dystopian:
RHS `Door toggle` feature is partly broken with this PR, I'm working on workaround.

Dystopian:
https://feedback.bistudio.com/T126920

Dystopian:
`Door toggle` is for some Hummers only, so it's much easier to remove it.
Need someone to check if class deleting is correct.
I think this PR is done.

Dystopian:
fixed in #5626.

PR #5619, opened 2017-10-14
Add compat for RHS 0.4.3 new vehicles

Dystopian:
- add Ural Repair and FMTV Rearm/Refuel/Repair;
- replace `era_*_hitpoint` elements with macros;
- fix `t72bd` ERA config;
- add missing T-80 and M2A2 ERA grouping configs.

It looks like it's better/simpler to create 1 big enough ERA grouping array (e.g. 100 elements) and not mess with exact ERA hitpoints count.

In fact `rhsusf_M1078A1R_SOV_M2_D_fmtv_socom` is not real refuel vehicle, so I made refuel hook at most appropriate (for me) place.


Neviothr:
Shouldn't `#define`s be in script_component.hpp?

Dystopian:
these macros are very specific for corresponding classes, so I decided not to move them far.

PR #5621, opened 2017-10-14
Add Advanced Throwing torque (rotation)

jonpas:
- Add new config entries `ace_advanced_throwing_torqueDirection[]` and `ace_advanced_throwing_torqueMagnitude`
- Add torque to the throwable, by default some random spin to all grenades
- Allow mods with eg. stick grenades to imply proper spin to make it look more realistic and immersive
- Add torque to debug info


addons/advanced_throwing/functions/fnc_throw.sqf
```diff
@@ -49,9 +49,15 @@ if (!(_unit getVariable [QGVAR(primed), false])) then {
         _newVelocity = _newVelocity vectorAdd (velocity (vehicle _unit));
     };
 
+    private _config = configFile >> "CfgAmmo" >> typeOf _activeThrowable;
+    private _torqueDir = getArray (_config >> QGVAR(torqueDirection));
+    private _torqueMag = getNumber (_config >> QGVAR(torqueMagnitude));
+    private _torque = vectorNormalized (_torqueDir vectorMultiply _torqueMag);
```
PabstMirror: Why is torqueMagnitude a complex random number if we normalize it?
`(50 + random 100)`  - 50 or 150 would produce the same end value

jonpas: @commy2 ?

commy2: Because this change is wrong and doesn't do what the previous code did.
Should be:
```sqf
    private _torqueDir = vectorNormalized getArray (_config >> QGVAR(torqueDirection));
    private _torqueMag = getNumber (_config >> QGVAR(torqueMagnitude));
    private _torque = _torqueDir vectorMultiply _torqueMag;
```

jonpas: Oh, my bad! 😆

PR #5622, opened 2017-10-14
Remove rain, lightnings and wind sync

### LINKED ISSUE #5465: (weather) wind and rain sync should not be a option
**Arma 3 Version:** stable
**CBA Version:** v3.4.0.170627
**ACE3 Version:** 3.10.2

**Mods:**
```
- CBA_A3
- ace
```

In some recent patch, rain and wind synchronization seems to be done by the engine
Clients will overtime start using the value the server has.

Missions or servers with `useAceWeather` off needs to turn `syncRain` and `syncWind` off. As for `syncMisc`, it requires some further investigation. Most likely Ace Weather (world data) shouldn't be synchronizing either, only setting the values on the server

This is due to https://github.com/acemod/ACE3/blob/master/addons/weather/functions/fnc_serverController.sqf#L30 using vectorMagnitude on `wind`, but `updateWind` and `getWind` will jump between engines wanted values and `ACE_WIND_PARAMS`, most likely. On some missions, reasons unknown, the value will eventually get extremely high 

On those missions, doing `[ace_weather_fnc_serverController, 1] call CBA_fnc_addPerFrameHandler;` will increase the wind extremely fast, reaching unplayable values. By default the update timer is 60s, so it can take up to 20 minutes before it gets very high

https://feedback.bistudio.com/T61395
this seems to suggest BI worked on improving weather sync about the same time as the weather module was created

<details>
<summary>#help chatlog</summary>
cuel [12:46] 
yea it just feels like it's all connected somehow. wind changes slowly at start, but once serverController has run a couple of times it's very off and extreme winds

[12:46] 
and from what i can tell it only happens when useaceweather is off

[12:47] 
calling updateaceweather seems to stabilize it

[12:47] 
just an observation

silentspike 
[12:48] 
Yeah I believe you're right

[12:48] 
And I triple checked the maths and it's right :stuck_out_tongue:

[12:48] 
Forgot my trig for a second there

cuel [12:58] 
ok it should be since useAceWeather is off, ACE_WIND_PARAMS is only set in fnc_serverController to current wind
updateWind gets  called every sec and uses ACE_wind (from getWind, that uses ACE_WIND_PARAMS) but since serverController uses the current wind it will slowly increase

[12:59] 
something something

silentspike 
[13:00] 
So it's like a mismatch in timing between updating the GVAR and reading it?

cuel [13:00] 
yeah something like that. shouldn't the serverController use fnc_getWind instead?

[13:00] 
not sure how that plays along with forecast set in eden

silentspike 
[13:01] 
Would that not just be a feedback loop? Since the server controller is feeding fnc_getWind

[13:03] 
My understanding is server controller handles syncing all clients. With useAceWeather off, it just reads the actual wind and sets that as the params. Then on the clients updateWind sets the wind to those params

cuel [13:03] 
but the server also runs `updateWind`

silentspike 
[13:04] 
Perhaps that's the issue :slightly_smiling_face:

cuel [13:04] 
the wind is constant at mission start until the first serverController fires, that's when it starts jumping in small values

[13:04] 
and eventually it'll reach high values because of the jumping

silentspike 
[13:05] 
I can see no need to run updateWind on the server unless useAceWeather is active

cuel [13:05] 
but will it update along with the forecast then?

silentspike 
[13:06] 
You mean the mission forecast thing?

cuel [13:06] 
yeah

[13:06] 
or is that ignored now even

silentspike 
[13:06] 
I think that is overridden by useAceWeather, so it all hinges on that setting

[13:07] 
If that setting is on, the server will need updateWind so that ace weather is applied. If not, then the server is just syncing the engine value to all clients

cuel [13:07] 
i'll do some tests

silentspike 
[13:08] 
I think maybe there's some sort of feedback loop between it reading the engine value and then re-applying it on the server again

[13:09] 
Have wanted to squash this bug for a while, good to have another pair of eyes on it :slightly_smiling_face:

silentspike 
[13:15] 
If what I think is true is true, I'd just add `if !(isServer)` around lines 21+22 of updateWind to fix it

cuel [13:16] 
aye

[13:16] 
i'll check it on a dedi

silentspike 
[13:18] 
It also kind of makes sense why disabling automatic engine whether usually fixes it but not always

[13:18] 
I think it's reading the value at mission start then constantly resetting that value, but fluctuating between that and the engine value applied by the forecast/auto weather

cuel [13:21] 
oh yeah

cuel [13:48] 
it seems to work

it's jumping between these values
[0,0.147803,0]
[1.16967e-008,0.133795,0]

and i guess that's when ace_weather is fighting vanilla weather?

[13:49] 
it matches ace_wind_params

silentspike 
[13:53] 
It jumps with the `!isServer` added?

[13:53] 
Or it works with that?

cuel [13:55] 
it doesn't increase in an extreme matter but it's still behaves odd. it's jumping between ace_wind_params values and something else (vanilla?) (edited)

silentspike 
[13:56] 
:smile: It's a start I suppose

cuel [13:58] 
is the servercontroller even needed anymore? when useAceWeather is off?
`NOTE: The effect is global only if command is executed on the server. Wind set locally will sync back to server value in a while.` (edited)

silentspike 
[13:59] 
I wasn't involved with ace when weather was made, but supposedly there's desync between server and clients

cuel [14:00] 
actually yeah one of the values it's hopping between seems to match the server's `wind` (edited)

silentspike 
[14:00] 
Personally I just disable it because who cares unless you're doing super long range shots :stuck_out_tongue:

[14:01] 
I feel like maybe it's just an intrinsic problem with trying to force weather sync non-engine side

cuel [14:01] 
yeah, vanilla weather sync was probably broken earlier but fixed lately (edited)

silentspike 
[14:02] 
Wouldn't be against considering the removal of syncing but keeping the custom weather stuff

cuel [14:09] 
yeah seems like at least rain and wind sync needs to go

[14:09] 
uncertain about lightnings gusts and waves

[14:09] 
biki doesn't say

cuel [15:13] 
i'll make an issue, not entirely sure what works or not or what the rewrite would include
</details>

jonpas:
- Close #5465 
- Move `setRain` to server (correctly synced in vanilla), we still want proper map data calculation
- Removed lightnings sync (synced correctly in vanilla), we didn't do anything else with it
- Moved `setWind` to only execute on server (correctly synced in vanilla), `FUNC(updateWind)` runs globally (I think), we still need to sync gusts and waves

This should resolve extreme wind bug (assuming it works). Completely untested.

### DIFF
--- a/addons/weather/ACE_Settings.hpp
+++ b/addons/weather/ACE_Settings.hpp
@@ -11,12 +11,6 @@ class ACE_Settings {
         typeName = "BOOL";
         value = 1;
     };
-    class GVAR(syncRain) {
-        displayName = CSTRING(syncRain_DisplayName);
-        description = CSTRING(syncRain_Description);
-        typeName = "BOOL";
-        value = 1;
-    };
     class GVAR(syncWind) {
         displayName = CSTRING(syncWind_DisplayName);
         description = CSTRING(syncWind_Description);
--- a/addons/weather/CfgVehicles.hpp
+++ b/addons/weather/CfgVehicles.hpp
@@ -24,12 +24,6 @@ class CfgVehicles {
                 typeName = "BOOL";
                 defaultValue = 1;
             };
-            class syncRain {
-                displayName = CSTRING(syncRain_DisplayName);
-                description = CSTRING(syncRain_Description);
-                typeName = "BOOL";
-                defaultValue = 1;
-            };
             class syncWind {
                 displayName = CSTRING(syncWind_DisplayName);
                 description = CSTRING(syncWind_Description);
--- a/addons/weather/XEH_postInit.sqf
+++ b/addons/weather/XEH_postInit.sqf
@@ -12,13 +12,12 @@ if (!isServer) then {
     "ACE_MISC_PARAMS" addPublicVariableEventHandler {
         TRACE_1("MISC PARAMS PVEH",ACE_MISC_PARAMS);
         if (GVAR(syncMisc)) then {
-            30 setLightnings (ACE_MISC_PARAMS select 0);
-            30 setRainbow    (ACE_MISC_PARAMS select 1);
-            30 setFog        (ACE_MISC_PARAMS select 2);
+            30 setRainbow    (ACE_MISC_PARAMS select 0);
+            30 setFog        (ACE_MISC_PARAMS select 1);
         };
-        GVAR(temperatureShift) = (ACE_MISC_PARAMS select 3);
-        GVAR(badWeatherShift)  = (ACE_MISC_PARAMS select 4);
-        GVAR(humidityShift)    = (ACE_MISC_PARAMS select 5);
+        GVAR(temperatureShift) = (ACE_MISC_PARAMS select 2);
+        GVAR(badWeatherShift)  = (ACE_MISC_PARAMS select 3);
+        GVAR(humidityShift)    = (ACE_MISC_PARAMS select 4);
         call FUNC(updateTemperature);
         call FUNC(updateHumidity);
     };
--- a/addons/weather/XEH_postInit.sqf
+++ b/addons/weather/XEH_postInit.sqf
@@ -6,14 +6,10 @@ GVAR(badWeatherShift) = (random 1) ^ 2 * 10;
 GVAR(humidityShift) = (5 - random 10) / 100;
 
 GVAR(wind_period_start_time) = CBA_missionTime;
-GVAR(rain_period_start_time) = CBA_missionTime;
-
-GVAR(ACE_rain) = rain;
 
 "ACE_WIND_PARAMS" addPublicVariableEventHandler { GVAR(wind_period_start_time) = CBA_missionTime; };
-"ACE_RAIN_PARAMS" addPublicVariableEventHandler { GVAR(rain_period_start_time) = CBA_missionTime; };
-"ACE_MISC_PARAMS" addPublicVariableEventHandler {
-    if (!isServer) then {
+if (!isServer) then {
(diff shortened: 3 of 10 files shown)


jonpas:
[ace_weather_lessSync_v1.zip](https://github.com/acemod/ACE3/files/1384943/ace_weather_lessSync_v1.zip)

PabstMirror:
I think we can drop fog sync as well

jonpas:
https://github.com/acemod/ACE3/issues/5465#issuecomment-326263243 says fog doesn't synchronize properly in any way, I guess our sync does fix it.

jonpas:
Tested locally, no errors, but that's all I can do atm.

Cuel:
>#5465 (comment) says fog doesn't synchronize properly in any way, I guess our sync does fix it.

`just tried 10 setfog [0,0,0] and the client did indeed synchronize. setFog requires further investigation.`
This was referring to server exec. It was syncing when using the fogParams array, not when using an integer.

jonpas:
Ah, missed that! That's problematic for ST NVGs still though, since it uses local fog as a view distance limiting factor (because you can't set that low view distance itself).

Cuel:
Someone should double check that setting `fog` with the array did indeed sync. I can check it in a couple of hours. Because that sounds rather problematic.

jonpas:
I'd honestly prefer our own fog sync then, so we can selectively sync it when NVGs are off for example. @PabstMirror?

Cuel:
Yes, but using `setFog` on the server with `fogParams` would sync to clients and make NVG look odd? Effectively it would break anything that isn't using the integer version, including existing missions. It would render fog decay and fog base useless.

PR #5624, opened 2017-10-14
Lock object while loading to cargo

Dystopian:
- title;
- disable static weapon disassembling while it's claimed.

This will restrict interaction with static weapons which are being dragged/carried/(loaded to cargo).

I would also like to disable `Set-up the tripod` feature while it's claimed, but suddenly it's a vanilla feature and there are some troubles.

### DIFF
--- a/addons/common/functions/fnc_claim.sqf
+++ b/addons/common/functions/fnc_claim.sqf
@@ -30,10 +30,17 @@ _target setVariable [QGVAR(owner), _unit, true];
 
 // lock target object
 if (_lockTarget) then {
+    private _canBeDisassembled = !([] isEqualTo getArray (_target call CBA_fnc_getObjectConfig >> "assembleInfo" >> "dissasembleTo"));
     if (!isNull _unit) then {
         [QGVAR(lockVehicle), _target, _target] call CBA_fnc_targetEvent;
+        if (_canBeDisassembled) then {
+            _target enableWeaponDisassembly false;
+        };
     } else {
         [QGVAR(unlockVehicle), _target, _target] call CBA_fnc_targetEvent;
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_claim.sqf
```diff
@@ -30,10 +30,17 @@ _target setVariable [QGVAR(owner), _unit, true];
...
+    private _canBeDisassembled = !([] isEqualTo getArray (_target call CBA_fnc_getObjectConfig >> "assembleInfo" >> "dissasembleTo"));
     if (!isNull _unit) then {
         [QGVAR(lockVehicle), _target, _target] call CBA_fnc_targetEvent;
+        if (_canBeDisassembled) then {
+            _target enableWeaponDisassembly false;
+        };
     } else {
         [QGVAR(unlockVehicle), _target, _target] call CBA_fnc_targetEvent;
+        if (_canBeDisassembled) then {
+            _target enableWeaponDisassembly true;
```
jonpas: Is there any way to check if weapon can be disassembled (real-time, not config default I mean)? This would break it if anyone would set it as not disassemble-able on purpose.

Dystopian: I supposed you'll ask this. When working on this PR I thought absolutely the same. I didn't find any way.

Then I reasoned this way: `fnc_claim` is used in cargo and dragging components. So if somebody disabled disassembling on purpose, he should disable cargo loading and dragging too. If not - I can just move the tripod without disassembling with loading or dragging. So I don't think it's a problem.

jonpas: I see, not perfect but I guess it works.

PR #5626, opened 2017-10-15
QuickMount - fix locked seats, getIn EH

### LINKED ISSUE #5611: Quick Mount MP issues
Mods:
CBA - 3.4.1.170912
RHS 0.4.2.2
ACE - 3.10.2
Arma - Stable 
Windows Dedi Cloud Server

Think there may be an issue with ace quick mount and vehicles that have two gunner positions/weapons like the modern GAZ from RHS with the pkm and 40mm GL for the turret, the drivers seat, gunner seat, and front passenger/commander were all taken when I tried to quick mount the GAZ as they drove past and instead of putting me in a passenger seat at the back, I crossed the ethereal plane and merged my soul with the person who was already in the gunner seat, effectively I was put in first-person spectator mode of the gunner, only thing I could do was watch or hit ESC and DC.

Only been able to trigger it a few times since the feature was added to ace, seems to only happen in ops

PabstMirror:
- Handle locked turrets (fix #5611)
- Switch to use `GetInTurret` which fixes "GetIn" event handler args
- Pass cargo index to `GetInCargo`

This should fix the issues from PR #5610

Any vic with a  gunner:
`player action ["GetInGunner", cursorObject];` // GetIn EH: `[z_1,"gunner",q,[]]`
`player action ["GetInTurret", cursorObject, [0]];` // GetIn EH: `[z_1,"gunner",q,[0]]`
The wrong turret index is probably why the driver's door gets left open.

RHS UAZ:
`player action ["GetInCargo", cursorObject];` player becomes glitched (headbug fix does nothing)
`player action ["GetInCargo", cursorObject, 4];` works as expected

### DIFF
--- a/addons/quickmount/functions/fnc_getInNearest.sqf
+++ b/addons/quickmount/functions/fnc_getInNearest.sqf
@@ -63,6 +63,12 @@ if (!isNull _target &&
 
                 if ((_effectiveRole in ["driver", "gunner"]) && {unitIsUAV _target}) exitWith {}; // Ignoring UAV Driver/Gunner
                 if ((_effectiveRole == "driver") && {(getNumber (([_target] call CBA_fnc_getObjectConfig) >> "hasDriver")) == 0}) exitWith {}; // Ignoring Non Driver (static weapons)
+
+                // Seats can be locked independently of the main vehicle
+                if ((_role == "driver") && {lockedDriver _target}) exitWith {TRACE_1("lockedDriver",_x);};
+                if ((_cargoIndex >= 0) && {_target lockedCargo _cargoIndex}) exitWith {TRACE_1("lockedCargo",_x);};
+                if ((!(_turretPath isEqualTo [])) && {_target lockedTurret _turretPath}) exitWith {TRACE_1("lockedTurret",_x);};
+
                 if (_effectiveRole == "turret") then {
                     if ((getNumber (([_target, _turretPath] call CBA_fnc_getTurret) >> "isCopilot")) == 1) exitWith {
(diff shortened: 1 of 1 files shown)


addons/quickmount/functions/fnc_getInNearest.sqf
```diff
@@ -75,12 +81,18 @@ if (!isNull _target &&
...
-                    TRACE_3("Geting In",_x,_role,_turretPath);
+                if (!(_turretPath isEqualTo [])) then {
+                     // Using GetInTurret seems to solve problems with incorrect GetInEH params when gunner/commander
+                    ACE_player action ["GetInTurret", _target, _turretPath];
+                    TRACE_3("Geting In Turret",_x,_role,_turretPath);
                 } else {
-                    ACE_player action ["GetIn" + _role, _target];
-                    TRACE_3("Geting In",_x,_role);
+                    if (_cargoIndex > -1) then {
+                        ACE_player action ["GetInCargo", _target, _cargoIndex];
```
commy2: Have you double checked this cargo index for vehicles with both FFV cargo turret slots and classic cargo slots? Like that one jeep from Apex? The cargo indices as used by some, but not all commands are shifted by the number of FFV slots.

commy2: By Jeep I mean that buggy one. With the co driver machine gun.

PabstMirror: seems to work fine

commy2: OK. Good!

PR #5628, opened 2017-10-16
Fix garrison header and use showMessage

PabstMirror:
@alganthe


alganthe:
I swear I thought I had done it, :+1:

PabstMirror:
u did it in ace_zeus_fnc_ui_garrison, which checks most of the same checks, so it really wouldn't have been a problem

PR #5631, opened 2017-10-17
Fix server rain PFEH

PabstMirror:
`GVAR(nextUpdateRain) = 0;` was inside the pfeh

### DIFF
--- a/addons/weather/XEH_postServerInit.sqf
+++ b/addons/weather/XEH_postServerInit.sqf
@@ -12,27 +12,21 @@ GVAR(ACE_rain) = rain;
 call FUNC(initWind);
 
 ["ace_settingsInitialized", {
-    TRACE_2("ace_settingsInitialized eh",GVAR(enableServerController),GVAR(serverUpdateInterval));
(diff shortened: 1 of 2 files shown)


addons/weather/XEH_postServerInit.sqf
```diff
@@ -12,27 +12,21 @@ GVAR(ACE_rain) = rain;
...
-        addMissionEventHandler ["EachFrame", {0 setRain GVAR(ACE_rain)}];
+        GVAR(nextUpdateRain) = 0;
 
-        // Create a 1 sec delay PFEH to update rain
-        [{
-            BEGIN_COUNTER(weatherPFEHserver);
-
-            GVAR(nextUpdateRain) = 0;
+        addMissionEventHandler ["EachFrame", {
             if (CBA_missionTime >= GVAR(nextUpdateRain)) then {
```
dedmen: https://feedback.bistudio.com/T123355
And you are executing every 2 second? Why are you replicating what CBA PFH already does?

PR #5639, opened 2017-10-19
Allow rearm, refuel and pylons to function while vehicle engine is running

tbeswick96:
- Remove pylons and refuel target vehicle engine blocking/checks.
- Add distance checks for pylons and rearm to close dialog and stop progress if the vehicle moves away.


addons/rearm/functions/fnc_rearm.sqf
```diff
@@ -48,6 +48,10 @@ if (_magazineDisplayName == "") then {
     {(_this select 0) call FUNC(rearmSuccess)},
     "",
     format [localize LSTRING(RearmAction), getText(configFile >> "CfgVehicles" >> (typeOf _target) >> "displayName"), _magazineDisplayName],
-    {true},
+    {
+        params ["_args"];
+        _args params ["_target", "_unit"];
```
Neviothr: `param [0] params ["_target", "_unit"];` perhaps?

addons/rearm/functions/fnc_rearmEntireVehicle.sqf
```diff
@@ -17,14 +17,18 @@
...
     TIME_PROGRESSBAR(10),
-    [_truck, _vehicle],
+    [_truck, _player, _vehicle],
     FUNC(rearmEntireVehicleSuccess),
     "",
     format [localize LSTRING(BasicRearmAction), getText(configFile >> "CfgVehicles" >> (typeOf _vehicle) >> "displayName")],
-    {true},
+    {
+        params ["_args"];
+        _args params ["", "_player", "_vehicle"];
```
Neviothr: Same?

addons/rearm/functions/fnc_rearm.sqf
tbeswick96: 👍

Dystopian:
I think there is no need for settings, distance check is enough.
Or 1 setting for all supplies.

tbeswick96:
@Dystopian are you suggesting removing the engine checks altogether and just use distance checks?

Dystopian:
Yes, that's my opinion, this restriction have nothing similar to RL.

tbeswick96:
Yeah I agree tbh, no reason other than to stop the code from breaking to have the engine disabled.

jonpas:
Introduce failure chance if engine is not off. 😆

tbeswick96:
Fuel nozzle already disconnects if the vehicle is driven away, and with the distance checks in the right places, the pylons dialog closes and progress bars fail, and rearm progress bars fail. That's all it really needs to be imo

jonpas:
It was a joke. :P

I am not sure though, input from other devs?

PabstMirror:
I'm fine with removing engine checks as long as distance is good (maybe check velocity < 0.1 as well?)
I've seen videos of an apache being rearmed/refueld with engines running

654wak654:
I'm also for keeping just the distance check.

Though Instead of introducing a new `PYLONS_DISTANCE_CHECK` macro, you should use the already existing 
`GVAR(searchDistance)` setting: https://github.com/acemod/ACE3/blob/master/addons/pylons/XEH_postInit.sqf#L18

Also there is a very small "bug":
The progress bar fail condition is `!(distance < PYLONS_DISTANCE_CHECK)`, but the dialog fail condition is `distance > PYLONS_DISTANCE_CHECK`. There is a very unlucky player out there who will switch from the menu to the progress bars when `distance == PYLONS_DISTANCE_CHECK` and get a seemingly random failure.

addons/pylons/functions/fnc_showDialog.sqf
```diff
@@ -137,3 +139,12 @@ _edit ctrlAddEventHandler ["KeyDown", LINKFUNC(onNameChange)];
 
 private _checkbox = _display displayCtrl ID_CHECKBOX_MIRROR;
 _checkbox ctrlAddEventHandler ["CheckedChanged", {[(_this select 1) == 1] call FUNC(onPylonMirror)}];
+
+// PFH for distance check to vehicle
+[{
+    params ["", "_idPFH"];    
+    if ((ace_player distance GVAR(currentAircraft)) > PYLONS_DISTANCE_CHECK) exitWith {
```
dedmen: `distanceSqr`
waitUntilAndExecute?

tbeswick96: @dedmen What's the advantage of using `distanceSqr`?

Realised I forgot to make this exit if the dialog is closed as well.

PabstMirror: distance = sqrt (xDist^2+yDist^2+zDist^2)
distanceSqr = (xDist^2+yDist^2+zDist^2)
If PYLONS_DISTANCE_CHECK is known static, you can precompute a `PYLONS_DISTANCE_CHECK_SQR` constant
then just check
`(ace_player distanceSqr GVAR(currentAircraft)) > PYLONS_DISTANCE_CHECK_SQR`

you save having to do a extra square root
normally it would be a nice optimization, but in sqf the performance difference pretty small

tbeswick96: Ok thanks for that, is that something that just be applied in general to distance comparisons?

dedmen: It can be. You can perf profile that yourself and see if it's actually worth it.
Also you might get precision issues because the number is then twice as big. and the bigger it is the lower the precision.. but considering the engine first does `xDist^2+yDist^2+zDist^2` which produces a huge number and then does sqrt... precision shouldn't really be different.

mr-bo-jangles:
Would this allow situations like this https://youtu.be/m0vr6OfNKTM?t=185 in arma?

tbeswick96:
Theoretically yes, if you can reach with the hose.
Better example is https://www.youtube.com/watch?v=Z-vweWwA9TI

addons/rearm/script_component.hpp
```diff
@@ -19,6 +19,7 @@
 
 
 #define REARM_ACTION_DISTANCE 7
+#define REARM_ACTION_DISTANCE_SQR REARM_ACTION_DISTANCE ^ 2
```
dedmen: You are now squaring the number at runtime every time... It doesn't matter no need to fix.

addons/pylons/XEH_postInit.sqf
```diff
@@ -43,4 +43,13 @@
 
         addMissionEventHandler ["HandleDisconnect", LINKFUNC(handleDisconnect)];
     };
+
+    GVAR(searchDistanceSqr) = GVAR(searchDistance) ^ 2;
+}] call CBA_fnc_addEventHandler;
+
+["ace_settingChanged", {
```
654wak654: This EH may be needed with CBA settings but currently you have to restart the mission/ server to change `GVAR(searchDistance)` anyway.

addons/pylons/functions/fnc_showDialog.sqf
```diff
@@ -137,3 +136,10 @@ _edit ctrlAddEventHandler ["KeyDown", LINKFUNC(onNameChange)];
 
 private _checkbox = _display displayCtrl ID_CHECKBOX_MIRROR;
 _checkbox ctrlAddEventHandler ["CheckedChanged", {[(_this select 1) == 1] call FUNC(onPylonMirror)}];
+
+[{
+    isNull (GVAR(currentAircraft) getVariable [QGVAR(currentUser), objNull]) ||
+    {(ace_player distanceSqr GVAR(currentAircraft)) > GVAR(searchDistanceSqr)}
+}, {
+    call FUNC(onButtonClose);
```
654wak654: Could add a hint about distance here.

addons/pylons/XEH_postInit.sqf
tbeswick96: No harm in having it though?

654wak654: It's as harmful as any block of code that's _never_ going to be called.

654wak654: But really, it's better to not add it. The CBA settings PR should take care of this part.

addons/pylons/functions/fnc_configurePylons.sqf
```diff
@@ -69,6 +69,6 @@ if (_currentPylon == count _pylonsToConfigure) exitWith {};
         [format [localize LSTRING(Stopped), _currentPylon + 1], false, 5] call EFUNC(common,displayText);
     },
     format [localize LSTRING(ReplacingPylon), _currentPylon + 1, count _pylonsToConfigure],
-    {true},
+    {(ace_player distanceSqr GVAR(currentAircraft)) <= GVAR(searchDistanceSqr)},
```
PabstMirror: This can be opened via a zeus module so it probably needs to be something like
`(!isNull curatorCamera) || (ace_player distanceSqr GVAR(currentAircraft)) <= GVAR(searchDistanceSqr)}`

addons/pylons/functions/fnc_showDialog.sqf
```diff
@@ -137,3 +136,11 @@ _edit ctrlAddEventHandler ["KeyDown", LINKFUNC(onNameChange)];
 
 private _checkbox = _display displayCtrl ID_CHECKBOX_MIRROR;
 _checkbox ctrlAddEventHandler ["CheckedChanged", {[(_this select 1) == 1] call FUNC(onPylonMirror)}];
+
+[{
+    isNull (GVAR(currentAircraft) getVariable [QGVAR(currentUser), objNull]) ||
+    {(ace_player distanceSqr GVAR(currentAircraft)) > GVAR(searchDistanceSqr)}
```
PabstMirror: same bit here, handle being called from zeus

addons/pylons/functions/fnc_configurePylons.sqf
654wak654: Could just use `GVAR(isCurator)` like line 25 does, the function already relies on it.

PR #5646, opened 2017-10-20
Collision lights keybind

tbeswick96:
- Add collision lights keybind
- Unbound by default (any suggestions for good default?)

### DIFF
--- a/addons/weaponselect/XEH_postInit.sqf
+++ b/addons/weaponselect/XEH_postInit.sqf
@@ -196,6 +196,18 @@ if (!hasInterface) exitWith {};
 {false},
 [10, [false, false, false]], false] call CBA_fnc_addKeybind; //9 Key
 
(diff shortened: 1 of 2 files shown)


jonpas:
I don't think a default keybind is necessary for this. But if we want it, some modifier on top of the vanilla "Lights" action.

tbeswick96:
Only issue with a modifier on top of L I found is that they are already bound to other things, or aren't great for vehicles.
Ctrl+L is something already (Change sight mode in CBA I think)
Shift+L not sure if it's something, however this will start the engine unwanted if Shift is bound as throttle
Alt+L seems to turn normal lights on as well (could be CBA bug?)

PabstMirror:
> Note that the vehicle has to be local, for global variant use Arma 3 Actions "CollisionLightOn"/"CollisionLightOff"

Could use the `action` version, but vehicle should always be local to driver.

jonpas:
Can finally bind that to my joystick!
_Next step is BI giving us joystick API so I don't have to use 3rd party software._

PR #5647, opened 2017-10-21
Zeus AI Static Weapon Modules

PabstMirror:
Adapted BIS function to make AI tear down and setup static weapons.
would have just used their funcs but they have some weird limitations and problems
Uses getModuleDestination to set weapon direction / watch target.


addons/zeus/functions/fnc_moduleStaticWeaponPlace.sqf
```diff
@@ -0,0 +1,65 @@
...
+if (({[_x] call EFUNC(common,isPlayer)} count units _group) != 0) exitWith { ["str_a3_cfgvehicles_moduleremotecontrol_f_errorPlayer"] call FUNC(showMessage); };
+
+private _gunner = objNull;
+private _assistant = objNull;
+private _availableUnits = [_unit]; // try to seach attached unit first
+_availableUnits append ((units _group) select {(!isNull _x) && {_unit != _x}});
+_availableUnits = _availableUnits select {([_x] call EFUNC(common,isAwake)) && {(vehicle _x) == _x}};
+
+{
+    if ((unitBackpack _x) isKindOf "Weapon_Bag_Base") exitWith {_gunner = _x};
```
AACO: This will trigger for UAV bags (how ever UAVs have a blank value for `assembleInfo >> base` so they will exit there, but with a confusing message. Maybe check if the assemble to string contains UAV?

Sub note: while not in vanilla crew served weapons, what happens when the `base` is an empty string (a single backpack setting up a turret)? Should that case be handled at all?

addons/zeus/functions/fnc_moduleStaticWeaponPlaceLocal.sqf
```diff
@@ -0,0 +1,91 @@
+/*
+ * Author: PabstMirror
+ * Adapted from BIS_fnc_packStaticWeapon (Dean "Rocket" Hall, reworked by Killzone_Kid)
+ * Place on a unit in a static weapon to make them dissasemble and pickup the backpacks parts
```
AACO: This is the assembly, not disassembly

addons/zeus/functions/fnc_moduleStaticWeaponPlaceLocal.sqf
```diff
@@ -0,0 +1,91 @@
...
+private _compatibleBases = if (isText _cfgBase) then {[getText _cfgBase]} else {getArray _cfgBase};
+
+private _availableUnits = (units group _gunner) select {(!isNull _x) && {_gunner != _x} && {[_x] call EFUNC(common,isAwake)} && {(vehicle _x) == _x}};
+private _assistant = objNull;
+{
+    private _unit = _x;
+    if (({(unitBackpack _unit) isKindOf _x} count _compatibleBases) > 0) exitWith {_assistant = _x};
+} forEach _availableUnits;
+TRACE_2("",_gunner,_assistant);
+if (!local _assistant) exitWith {WARNING("Split locality");};
```
AACO: You already calculate the base/assistant in the calling function, should those just be passed in, instead of recalculating it?

addons/zeus/functions/fnc_moduleStaticWeaponTake.sqf
```diff
@@ -0,0 +1,66 @@
...
+if (({[_x] call EFUNC(common,isPlayer)} count units _group) != 0) exitWith { ["str_a3_cfgvehicles_moduleremotecontrol_f_errorPlayer"] call FUNC(showMessage); };
+
+private _staticWeapon = vehicle _unit;
+if (!(_staticWeapon isKindOf "StaticWeapon")) then {
+    _staticWeapon = (_unit nearEntities ["StaticWeapon", 50]) param [0, objNull];
+};
+if ((!(_staticWeapon isKindOf "StaticWeapon")) ||
+        {(locked _staticWeapon) >= 2} ||
+        {(!isNull gunner _staticWeapon) && {(group gunner _staticWeapon) != _group}} ||
+        {(count getArray(configFile >> "CfgVehicles" >> (typeOf _staticWeapon) >> "assembleInfo" >> "dissasembleTo")) != 2}) exitWith {
```
AACO: Again, what happens if this is a single component static weapon?

addons/zeus/functions/fnc_moduleStaticWeaponTakeLocal.sqf
```diff
@@ -0,0 +1,118 @@
...
+    case (2): {
+            if ((!_gunnerReady) && {unitReady _gunner}) then {
+                TRACE_1("gunner in pos",_gunner distance _staticWeapon);
+                doStop _gunner;
+                _args set [5, true];
+            };
+            if ((!_assistantReady) && {unitReady _assistant}) then {
+                TRACE_1("assistant in pos",_assistant distance _staticWeapon);
+                doStop _assistant;
+                _args set [6, true];
```
AACO: Should update the local ready vars here, so if they're both ready they can start to disasemble without another iteration of the PFH

AACO:
Just curious, besides belts/suspenders checks, can the gunner/assistant ever be split? If they're AI in a group shouldn't they have the same owner?

addons/zeus/functions/fnc_moduleStaticWeaponPlace.sqf
```diff
@@ -0,0 +1,65 @@
...
+} forEach _availableUnits;
+
+if (isNull _gunner) exitWith { [localize LSTRING(StaticWeapon_NotCarrying)] call FUNC(showMessage); };
+private _cfgBase = configFile >> "CfgVehicles" >> (backpack _gunner) >> "assembleInfo" >> "base";
+private _compatibleBases = if (isText _cfgBase) then {[getText _cfgBase]} else {getArray _cfgBase};
+{
+    private _xUnit = _x;
+    if (({(unitBackpack _xUnit) isKindOf _x} count _compatibleBases) > 0) exitWith {_assistant = _xUnit};
+} forEach _availableUnits;
```
AACO: For the assistant checks, it could potentially select a player, should not is player be checked?

PabstMirror: handled by line 34, if any member of the group is a player it will exit

PabstMirror:
I have seen split locality groups, but I think it's only a tempoary thing (like when a player connects to a AI slot, the group members might be local to the server for a little bit).  Really don't think it would be be an issue, but better to be safe.

I will look into single backpack things like the uav.

Mike-MF:

PR #5648, opened 2017-10-21
Scopes - BoreHeight config revamp

ulteq:
Special thanks to @dedmen for providing the tools to extract the required data.

* Added rail heights for all vanilla guns
* Improved default rail height
* Updated all scope heights

| Before  | After |
|  |  |

**TODO:**
- [x] Verify the correctness of the newly added config (class inheritance, ...)
- [x] Update the default bore heights in the ATragMX gun list
- [x] Update the default scope base angles in the ATragMX gun list


bux:
This should be part of 3.11 imho

@ulteq do the other compats need updates too?
Do we need to inform the cup guys?

dedmen:
\(O_O)/
@bux No. Not really. This is nothing new. Just improved values for the old stuff.
Would be good if we also updated to compats. but not a must.

ulteq:
> do the other compats need updates too?
Those 3rd party mods who already use their own bore height config don't need to change anything.
And those who don't yet use custom bore heights will probably benefit from the new defaults values.

dedmen:
@bux https://docs.google.com/spreadsheets/d/1d5O_nC0Cq26CWGI4sqjZU9jdhArV5pkSm-VMDT7uvM4/edit#gid=0 The CUP guys could compare against that to see if that matches their compats.
That list doesn't contain all CUP weapons though.

Laid3acK:
About the RHS compats, i'm not a RHS user, someone should update my "empiric" values with the new ones.
Not sure there is a big difference in-game but it will be surely a big improvement for all RHS players. :)

QuickDagger:
Hi guys,
Oh, good to see you have fixed the scope height (ace_RailHeightAboveBore + ACE_ScopeHeightAboveRail) showing up on the Range Card. It was affecting the close Zero. far zero was so so but now it all looks good.

PR #5650, opened 2017-10-21
AB compat rail/scope height update

dedmen:
Compat variant of #5648
I only updated the weapons we already had in compat and only added new ones were really necessary.
I noticed we are missing all RHS AK variants with NPZ which have a different railheight that the non-NPZ ones.
If someone else want's to add the missing stuff here is a table of all rail heights: https://docs.google.com/spreadsheets/d/1d5O_nC0Cq26CWGI4sqjZU9jdhArV5pkSm-VMDT7uvM4/edit?usp=sharing

### DIFF
--- a/optionals/compat_r3f/CfgWeapons.hpp
+++ b/optionals/compat_r3f/CfgWeapons.hpp
@@ -4,7 +4,7 @@ class CfgWeapons {
     class Pistol_Base_F;
     class Rifle_Base_F;
     class R3F_Famas_F1: Rifle_Base_F {
-        ACE_RailHeightAboveBore = 10.6;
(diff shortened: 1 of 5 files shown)


dedmen:
@ulteq What is WIP on this? Besides adding all the NPZ classes which I don't plan on doing.

ulteq:
> Besides adding all the NPZ classes 

This, but I guess we could also insert them later with a separate PR.

dedmen:
Yeah. This is just a update to improve the existing compat. IMO the RHS compat should be completly overhauled. checking if inheritance is still correct. and really adding all the missed classes

PR #5652, opened 2017-10-22
Adds Crew Served Static Weapons

TheCandianVendingMachine:
- Add custom crew-served static weapons
- Add the ability to load/unload ammo from static weapons
- Adds carry-weapons and carry-tripods that are needed to assemble a static weapon.

The way I implemented this is by adding a few config values to the static weapons, and adding new "Launcher" weapons to allow the user to carry both a backpack and a gun-bag.

The config values added are in the carry weapon, the static vehicle, and the weapon that the static vehicle uses. I decided that all data about the time it takes to un/deploy and un/load ammo would go into the vehicle weapon as for most of the other timings logically go into there, and it reduces the amount of foreign config values added. The only downside I see to this is that there are some areas in the code where I travel through the config values more than once to get the data, and that could possibly lead to some performance issues. The only other config values added are meta-data about what the static weapon will dis/assemble to, what tripod it can deploy on, and if it is enabled for that particular static weapon to satisfy the issue PabstMirror brought up.

I also added base classes to allow for easy additions of static weapons for the end-user. Essentially wrappers for the base classes everything should use with some defaults defined so nothing goes wrong if a config is messed up.

As I said before, everything that is needed to assemble a weapon are defined as launchers. I decided that was the best implementation as for it allows for a minimum of a 2 man CSW with weight issues, rather than a forced minimum 3 man CSW crew with weight issues. I found that was the best balance between realism and game balance, as well as implementing a real way that militaries carry static weapons according to [this picture.](https://i.imgur.com/t4CM1W9.jpg) The only issue is that custom weapons will have to implement this specifically for ACE while possibly adding bloat for their own mods. I don't know what the consensus on this implementation is, and would like discussion about it

The way that the code actually checks if the weapon is supported and is enabled is via a config check if the current launcher has the config value that defines which tripod it can be mounted on.

There are two key events that I defined that fix some issues regarding locality. One is defined only on the server, and is called when a weapon is spawned. When spawned the weapon changes locality to the server so I can `setAmmo` easily on it without worrying if the user that spawned it disconnected, and just generally having faith that it will exist forever. The other event defined is one that all clients have and it is a wrapper for `setAmmo`. All it does is guarantee that the ammo will be set if weapon locality changes for whatever reason.

I also added a new ACE action on all static weapons. `Get In`. I added this due to an ARMA issue where the action to get in to the vehicle does not appear due to a memory point issue. I don't think I can fix this, so I made the work around

I have tested this in single-player and multi-player on a dedicated server and found no glaring issues.

In the documentation, have `version::minor` as `?` since I don't know when this will be merged.

This is also incompatible with #5647 as for it uses a completely different system.


addons/csw/functions/fnc_tripodDeploy.sqf
```diff
@@ -0,0 +1,39 @@
...
+ * Example:
+ * [player] call ace_csw_fnc_tripodDeployConfirm
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params["_player", "_key"];
+
+if ((_key != 0) || {isNull GVAR(cswTripod})) exitWith {};
```
commy2:
```sqf
if ((_key != 0) || {isNull GVAR(cswTripod})) exitWith {}; 
```
should be:
```sqf
if ((_key != 0) || {isNull GVAR(cswTripod)}) exitWith {}; 
```

commy2:
Travis complains about tabs. Tabs should be spaces in ACE code.

addons/csw/CfgWeapons.hpp
```diff
@@ -0,0 +1,250 @@
+class CfgWeapons {
+    class Launcher;
+    class Launcher_Base_F: Launcher {
+        class WeaponSlotsInfo;
+    };
+    class GVAR(m3CarryTripod): Launcher_Base_F {
+        class GVAR(cswOptions) {
```
BaerMitUmlaut: I'd name that class just options - the GVAR macro would resolve to ace_csw_cswOptions, kind of redundant.

addons/csw/README.md
```diff
@@ -0,0 +1,11 @@
...
+===============
+
+Crew Served Weapons - Static weapons that are served by multiple people
+
+
+## Maintainers
+
+The people responsible for merging changes to this component or answering potential questions.
+
+- [TCVM (Brandon)](https://github.com/TheCandianVendingMachine)
```
BaerMitUmlaut: Maintainer is somebody from the ACE team.

PabstMirror:
This breaks UAV backpacks from vanilla and all static weapons from other addons

It might be better to make this system optional, so that it can work on top of the vanilla assemble system, instead of replacing it entirely.

TheCandianVendingMachine:
I think that all of the issues brought up have been fixed. Is there anything else I need to do?

addons/csw/CfgMagazines.hpp
```diff
@@ -0,0 +1,89 @@
...
+    class 100Rnd_127x99_mag;
+    class 1Rnd_GAT_missiles;
+    class 40Rnd_20mm_G_belt;
+    class 1Rnd_GAA_missiles;
+    
+    class GVAR(100Rnd_127x99_mag): 100Rnd_127x99_mag {
+        author = ECSTRING(common, ACETeam);
+        displayName = CSTRING(HMGBelt_displayName);
+        model = "\A3\Structures_F_EPB\Items\Military\Ammobox_rounds_F.p3d";
+        scope=2;
```
thojkooi: Could you adjust the spacing so there is a space infront and behind each `=`? 

For example on line 11, change it to:
```
scope = 2;
```

This will keep everything uniform and clean. This occures on multiple locations in this file.

addons/csw/XEH_postInit.sqf
```diff
@@ -0,0 +1,36 @@
+#include "script_component.hpp"
+
+if (isServer) then {
+    [QGVAR(addObjectToServer), {
+        params["_object"];
+        _object setOwner 2;
```
thojkooi: It's not immedially clear that `2` is the server's owner id. Could you change this to be a defined macro?

```sqf
#define SERVER_OWNER_ID 2

....
[QGVAR(addObjectToServer), {
     params ["_object"];
    _object setOwner SERVER_OWNER_ID;
```

addons/csw/functions/fnc_deployTripod.sqf
```diff
@@ -0,0 +1,55 @@
...
+ */
+#include "script_component.hpp"
+
+params ["_player"];
+
+[{
+    params["_player"];
+    // Remove the tripod from the launcher slot
+    private _secondaryWeaponClassname = secondaryWeapon _player;
+    _player removeWeaponGlobal (secondaryWeapon _player);
```
thojkooi: Does this also remove magazines loaded in the secondary weapon? If so, re-adding any removed magazines may need to be handled at line 49.

addons/csw/functions/fnc_deployWeapon.sqf
```diff
@@ -0,0 +1,63 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params["_tripod", "_player"];
+
+[{
+    params["_tripod", "_player"];
+    private _carryWeaponClassname = secondaryWeapon _player;
+    _player removeWeaponGlobal _carryWeaponClassname;
```
thojkooi: Same as above, secondary weapon magazines may need to be handled in this file as well if `removeWeaponGlobal` removes those.

addons/csw/functions/fnc_loadAmmo.sqf
```diff
@@ -0,0 +1,62 @@
...
+
+private _cswWeapon = getArray(configFile >> "CfgVehicles" >> typeof(_csw) >> "Turrets" >> "MainTurret" >> "weapons") select 0;
+private _cswMagazines = getArray(configFile >> "CfgWeapons" >> _cswWeapon >> "magazines");
+
+private _magazineSelected = "";
+private _magazineAmmoCount = 0;
+private _maxAmmo = [_csw] call FUNC(getMaxAmmo);
+
+{
+    if ((_x select 0) in _cswMagazines && (_csw ammo _cswWeapon) + (_x select 1) <= _maxAmmo) exitWith {
```
thojkooi: Please make use of lazy evaluation for all conditions. For example, this if statement would need to be changed to:

```
if ((_x select 0) in _cswMagazines && {(_csw ammo _cswWeapon) + (_x select 1) <= _maxAmmo}) exitWith {
```
You would add `{..}` for all tests that have multiple conditions. One for each condition after the first.

See also: https://community.bistudio.com/wiki/a_%26%26_b

addons/csw/functions/fnc_unloadAmmo.sqf
```diff
@@ -0,0 +1,93 @@
...
+ * Example:
+ * [CSW, false, false] call ace_csw_fnc_unloadAmmo
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params["_csw", "_slowUnload", "_allAmmo"];
+
+private _weaponPos = _csw getRelPos[1.5, 270];
```
thojkooi: These seem very specific values. Could you move `[1.5, 270]` to a macro so we can label it, just like a variable?

addons/csw/functions/fnc_deployTripod.sqf
TheCandianVendingMachine: Since the secondaryWeapon in this case is the tripod, there should be zero ammo in the secondaryWeapon. I think it's not needed, but if you really think its should exist I will add it

addons/csw/functions/fnc_deployWeapon.sqf
TheCandianVendingMachine: Again, because of how the CSW's are implemented ammo in the secondaryWeapon should not be an issue. I clarified how I implemented them in the edited PR fourth paragraph.

SilentSpike:
Pushing this milestone back while still WIP

PabstMirror:
I'm wondering if we could combine this with https://github.com/acemod/ACE3/pull/4701
Have this focus on weapon assembly and transport

TheCandianVendingMachine:
@SilentSpike As far as I am concerned this PR in the current state isn't WIP, instead being feature complete. As far as I know all changes that have been requested were addressed, and the times I have inquired for clarification on a requested change have been ignored. I don't know if you meant this because the reviewers have not dismissed the changes that I have addressed, or if there is another reason that I am unaware of

@PabstMirror I think that system is better than what I currently have implemented for ammo loading. I think that combining the two would be worthwhile.

bux:
I'll take a look at this today

bux:
Got this error when placing the CSW tripod

image

TheCandianVendingMachine:
@bux The lack of textures is due to me, sadly, not being a texture artist. The error in the first picture is because I accidentally gave an old texture to an LOD and neglected to fix it. The current texture for all data is just some noise on a black background.

If you're interested, [here is the texture used](https://i.imgur.com/kozkLg9.png)

bux:
@TheCandianVendingMachine have you [joined our slack](https://slackin.ace3mod.com/) yet?
If yes check out the #feature_csw channel

TheCandianVendingMachine:
Ready for review, again

TheCandianVendingMachine:
Tested in Dedicated MP and it worked for me. Ready for review again, again

addons/csw/CfgMagazineGroups.hpp
```diff
@@ -0,0 +1,88 @@
...
+    };
+    class ACE_1Rnd_82mm_Mo_HE_Guided : GVAR(default) {
+        8Rnd_82mm_Mo_guided = 1;
+    };
+    class ACE_1Rnd_82mm_Mo_HE_LaserGuided : GVAR(default) {
+        8Rnd_82mm_Mo_LG = 1;
+    };
+    
+    // Titan Missiles
+    class GVAR(1Rnd_GAT_missiles) : GVAR(default) {
```
PabstMirror: I think we can just use `Titan_AT` which is the mag for the human launcher

addons/csw/CfgWeapons.hpp
```diff
@@ -0,0 +1,260 @@
...
+        model = QPATHTOF(data\ACE_CSW_WeaponBag.p3d);
+        modes[] = {};
+        picture = QPATHTOF(UI\StaticHGMG_Icon.paa);
+    };
+    
+    // STATIC WEAPONS
+    class HMG_Static;
+    class GVAR(HMG_Static): HMG_Static {
+        class GVAR(options) {
+            deployTime = 15;
```
PabstMirror: imho, some of the progress bars are a little long. I know some MGs are just put down tripod, put down gun and push in a pin. I think ace's progress make this a little worse because you can't move or look around.

addons/csw/functions/fnc_weaponInit.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params["_weapon"];
+
+if (isNull(configFile >> "CfgVehicles" >> typeof(_weapon) >> QGVAR(options) >> "enabled") || {getNumber(configFile >> "CfgVehicles" >> typeof(_weapon) >> QGVAR(options) >> "enabled") != 1}) exitWith {};
+if (_weapon getVariable[QGVAR(initialized), false]) exitWith {};
+if (!(_weapon getVariable[QGVAR(enableCSW), false])) exitWith {};
```
PabstMirror: default value on 3den attribute is true, only gets save if changed so default value should match.  
 .e.g. I put down static, check it's attribute is pre-checked, but the var wasn't set in-game

addons/csw/CfgWeapons.hpp
```diff
@@ -0,0 +1,260 @@
...
+        displayName = CSTRING(StaticAutoGMGBag_displayName);
+        author = ECSTRING(common, ACETeam);
+        scope = 2;
+        model = QPATHTOF(data\ACE_CSW_WeaponBag.p3d);
+        modes[] = {};
+        picture = QPATHTOF(UI\StaticHGMG_Icon.paa);
+    };
+    class GVAR(staticGMGTallCarry): Launcher_Base_F {
+        class GVAR(options) {
+            assembleTo = QGVAR(staticGMGTallWeapon);
```
PabstMirror: For high/low the Tripods are different, not the weapons
will get a little messy, can worry about this later

addons/csw/functions/fnc_positionTripod.sqf
```diff
@@ -0,0 +1,34 @@
...
+ * [player, tripod] call ace_csw_fnc_positionTripod
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params["_player", "_tripod"];
+
+[_player select 1, "forceWalk", "ACE_csw", true] call EFUNC(common,statusEffect_set);
+_tripod attachTo [_player select 1, [0, 2, 0.5]];
```
PabstMirror: There is some overlap with ace_dragging.
We could use that instead, but it's also more restrictive and wouldn't allow moving while prone.

If we keep our own method, we'll need to add some more scripts to handle player dying, unconscious, getting in vehicles and something to claim the tripod.

dedmen:
I see you are adding models. Are they GPL or do we have the same problem as with ace_apl's models?

TheCandianVendingMachine:
@dedmen From what I was told from @PabstMirror ACE 2 models are OK to be added and have been added in the past with bf7f1e8. The only other things ported were some ARMA 2 paa's and an rvmat which I thought was OK to be ported to other ARMA games

If the ACE 2 stuff isn't OK to be ported, is it possible for them to be ported at all or is it a dead end?

dedmen:
If they are APL we have to put them into ace_apl or put crew served under APL license. I'd go the first route.
Other models we ported from ACE2 are APL too that's why I ask. @PabstMirror probably knows.

severgun:
CfgMagazineGroups config or ace_csw_fnc_reload_canLoadMagazine are broken.
There is no config entry that code trying to get.
https://github.com/TheCandianVendingMachine/ACE3/blob/acc0bb0906d86fa9749d3337eb269fd66a21caad/addons/csw/functions/fnc_reload_canLoadMagazine.sqf#L44

Also wiki have nothing similar with addon configs.

severgun:
1) Addon disabled in settings still allow operations with mags. (So disabling not bring vanilla back as is)
2) Wiki outdated
3) Would be nice to have tripod placing process implemented as move action. So you can see model, choose height by scroll wheel and place. Only after that start progressbar.

severgun:
Trick with proxy classes brake reload and shooting animations, and cause reloading sound desync

addons/csw/functions/fnc_assemble_pickupWeapon.sqf
```diff
@@ -0,0 +1,82 @@
...
+        private _cswTripod = createVehicle [_turretClassname, [0, 0, 0], [], 0, "NONE"];
+        _cswTripod setPosATL _weaponPos;
+        _cswTripod setDir _weaponDir;
+        _cswTripod setVelocity [0, 0, -0.05];
+        _cswTripod setVectorUp (surfaceNormal _weaponPos);
+
+        _weaponRelPos = _cswTripod getRelPos RELATIVE_DIRECTION(90);
+        private _weaponHolder = createVehicle ["groundWeaponHolder", [0, 0, 0], [], 0, "NONE"];
+        _weaponHolder setPosATL [_weaponRelPos select 0, _weaponRelPos select 1, _weaponPos select 2];
+        _weaponHolder setDir random [0, 180, 360];
```
severgun: I found weird behaviour of this command on some islands. Weaponholder spawn with negative Z ATL coordinates.
Quote from BIKI:

> Make sure you setDir BEFORE you set position. Setting direction after set position could lead to unpredictable behaviour. For example main part of the hospital building in Arma 3 can lose collision detection near both side entrances. AI will also get confused and will stop detecting obstacles if setDir is called after setPos.

So I suppose it should be swapped

severgun:
Would be nice to see non static vehicles support too.
SPG/MG offroads as an example.

addons/csw/functions/fnc_canGetIn.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * Example:
+ * [weapon] call ace_csw_fnc_canGetIn
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_staticWeapon"];
+
+alive _staticWeapon && (((crew _staticWeapon) isEqualTo [] || {!(alive (gunner _staticWeapon))}))
```
PabstMirror: check if locked

Cuel: Why check both crew and gunner?

TheCandianVendingMachine: According to the biki `crew` returns both alive and dead members

Cuel: So the alive gunner check should be enough? :p

PabstMirror:
AI compatibility https://github.com/acemod/ACE3/commit/581155f73d5637a6885b42f17803db9885d3debb
Should fix #6173

addons/csw/functions/fnc_reload_actionsLoad.sqf
```diff
@@ -0,0 +1,56 @@
...
+};
+
+{
+    _x params ["_carryMag", "_turretPath", "_loadInfo"];
+    _loadInfo params ["", "", "", "_isBeltLinking"];
+
+    private _displayName = getText (configFile >> "CfgMagazines" >> _carryMag >> "displayName");
+    private _picture = getText (configFile >> "CfgMagazines" >> _carryMag >> "picture");
+    private _text = if (_isBeltLinking) then {
+        format ["Link: %1", _displayName];
```
severgun: Not localized action names and progressbars

addons/csw/functions/fnc_reload_handleAddTurretMag.sqf
```diff
@@ -0,0 +1,56 @@
...
+private _ammoRemaining = _ammoRecieved;
+if (_canAdd) then {
+    private _ammoUsed = _neededAmmo min _ammoRecieved;
+    _ammoRemaining = _ammoRemaining - _ammoUsed;
+
+    if (_isBeltLinking) then {
+        private _currentAmmo = _vehicle magazineTurretAmmo [_loadedMag, _turret];
+        _currentAmmo = _currentAmmo + _ammoUsed;
+        TRACE_2("Setting mag ammo",_loadedMag,_currentAmmo);
+        _vehicle setMagazineTurretAmmo [_loadedMag, _currentAmmo, _turret];
```
severgun: `setMagazineTurretAmmo` broken by BIS.
It is not working if turret **and** vehicle not local to issuer.
Because this is **CREW** served weapons system it must allow to link and load ammo by assistant player while other player sitting on gunner slot.
There is linking code in ace_reload module. Maybe it will help.
Also if CSW will be applieble to vehicles `setMagazineTurretAmmo` also will cause problems because vehicle is local to driver and not gunner.

addons/csw/functions/fnc_reload_getLoadableMagazines.sqf
```diff
@@ -0,0 +1,52 @@
...
+
+params ["_vehicle", "_player"];
+
+private _carriedMagazines = [];
+
+{
+    if (isClass (configFile >> QGVAR(groups) >> _x)) then {
+        _carriedMagazines pushBackUnique _x;
+    };
+} forEach (magazines _player);
```
severgun: Would be nice to see ability to load magazines from groundWeaponholder.

TheCandianVendingMachine: This PR is for all intents and purposes feature complete

addons/csw/functions/fnc_assemble_canDeployTripod.sqf
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Return Value:
+ * Can deploy <BOOL>
+ *
+ * Example:
+ * [player] call ace_csw_fnc_canDeployTripod
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
```
severgun: As much as I know include must be before header to fix line numbers reported in case of errors.

severgun:
Also seems like ace_rearm must be disabled for entire StaticWeapons class.

addons/csw/functions/fnc_reload_handleAddTurretMag.sqf
PabstMirror: seems like the normal `setAmmo` command will work
https://github.com/acemod/ACE3/pull/5652/commits/386d6aec75e4be9fe919037cca1632a36a44c871#diff-f2d09afb4c14c8d9fc63da546b5a0d9bR46

I had planned to expand this to cover full vehicles, (like m2 turrets on tanks) at some point and I'm worried how this will work with multi turreted vehicles
but it seems to work just fine for static weapons
and, it's really only required for belt linking

PabstMirror:
I think this is pretty close to being ready for a final review

severgun:
Found little "bug" on heavy loaded server.
assemble/disassemble action need some kind of delay between vehicle spawn.
In lags sometimes gun can be spawned on top of tripod and cause collision.
https://clips.twitch.tv/InterestingPlumpClintLeeroyJenkins

PR #5660, opened 2017-10-24
Fix 5.56 Reload Tracer mags are not tracer

### LINKED ISSUE #5656: 5.56mm Reload. Tracers don't work
Arma 3 Version: 1.76 stable
CBA Version: 3.4.1 stable
ACE3 Version: 3.10.2 stable

Mods:
- CBA_A3
- ace
- task_force_radio

Description:
Reload. Tracers for every 5.56mm assault rifle don't have tracers

Steps to reproduce:
- take any 5.56mm assault rifle;
- load Reload. Tracers of any color;
- start shooting.

Where did the issue occur?
- I tested it in a Dedicated server and in Editor (SP)

Placed Modules:
tested with no modules;
tested with:

- Adv. Fatigue;
- Captives Settings;
- Medical Settings;
- Respawn Settings;
- Revive Settings;
- Set Medic Class;

RPT log is not available.

Dystopian:
- title;
- fix #5656.

5.56 mags have correct names in vanilla ATM. There is no need to disable reload tracers there.

### DIFF
--- a/addons/smallarms/CfgMagazines.hpp
+++ b/addons/smallarms/CfgMagazines.hpp
@@ -6,36 +6,6 @@ class CfgMagazines {
     // 2. Remove tracers at bottom of magazine.
     // 3. Do string updates.
 
-    // 5.56mm ////////////////////////////////////
-
-    class 30Rnd_556x45_Stanag : CA_Magazine {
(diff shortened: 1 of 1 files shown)


PabstMirror:
I'm not sure what we were originally fixing with these configs, but this cleanup looks good

PR #5661, opened 2017-10-24
Fix Pylons' Combo Box Cut off When There Are Too Many Magazines to List

### LINKED ISSUE #5649: Pylon interface dropdowns cut off at bottom
**Arma 3 Version:** `stable `
**CBA Version:** `stable `
**ACE3 Version:** `3.11.0-rc2`

**Mods:**
```
- CBA_A3
- ace
- CUP
- 3CB (all except units)
- custom mods that add a few pylon magazines
```

**Description:**
- Dropdown boxes for pylon magazines cuts off magazines at the bottom of the list when there are a lot of options.

654wak654:
- Fix combo boxes cutting off. Goodbye #5649 
- 😢 

Using the old `RscCombo` type but making it look like `ctrlCombo` looks like the ideal way to get a proper combo box that doesn't cut off at an arbitrary control height limit.

### DIFF
--- a/addons/pylons/functions/fnc_showDialog.sqf
+++ b/addons/pylons/functions/fnc_showDialog.sqf
@@ -57,7 +57,7 @@ if (["ace_fastroping"] call EFUNC(common,isModLoaded) && {_hasFRIES > 1}) then {
 
 GVAR(comboBoxes) = [];
 {
-    private _combo = _display ctrlCreate ["ctrlCombo", -1];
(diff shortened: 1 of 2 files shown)


tbeswick96:
Can confirm this works well enough for a quick fix, but with some layout sizes there is still some clipping.

Maybe a full fix in a later version using `pixelGrid` as recommended by @alganthe would be the best option.

jonpas:
Agreed with above. Can you open a new issue for that?

PR #5672, opened 2017-10-24
Ability for ace engineer to configure pylons

shukari:
- allow ace engineers to configure pylons


654wak654:
`isEngineer` [already checks](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_isEngineer.sqf#L20) for unit trait too, so we might as well remove that part. Also lazy eval.

`if (GVAR(requireEngineer) && {!([_unit] call EFUNC(common,isEngineer))}) exitWith {false};`

PR #5675, opened 2017-10-25
Fix ACE_ItemCore order

dedmen:
`ACE_ItemCore` is always above `CBA_MiscItem_ItemInfo` except here.


dedmen:
https://github.com/acemod/ACE3/blob/master/addons/common/CfgWeapons.hpp Seems reasonable to group CBA thingies together

PR #5687, opened 2017-10-26
Tagging - For lastUsed setting, use random X tag if none selected yet

### LINKED ISSUE #5662: "Quick Tag" keybind only works after using ace menu first
Arma3 - stable
CBA - 3.4.1.170912
ACE - 3.10.2

**Mods:**
TFR - 1.-1.0.264
Lots of other shite...

**Description:**
- Quick-Tag keybind only works after you have placed a tag using the ace interaction menu first, after that it works fine

**Steps to reproduce:**
- Try placing a tag using the keybind method (it will not work) place a tag using the ace menu (it will work), now place another tag using your keybind (it will work)

- Dedicated Windows Cloud

PabstMirror:
Close #5662

When using the "Last Used" setting, nothing would happen when hitting the hot key.
This changes it to just do a random X

`GVAR(quickTag == 2)` :smile:

### DIFF
--- a/addons/tagging/functions/fnc_quickTag.sqf
+++ b/addons/tagging/functions/fnc_quickTag.sqf
@@ -22,19 +22,22 @@ if (GVAR(quickTag) == 0) exitWith {};
 params ["_unit"];
 
 private _possibleTags = [];
+private _useRandom = false;
 
 // Last Used
 if (GVAR(quickTag) == 1) then {
     private _lastUsedTagClass = _unit getVariable [QGVAR(lastUsedTag), nil];
 
-    if (!isNil "_lastUsedTagClass") then {
+    if (isNil "_lastUsedTagClass") then {
+        _useRandom = true;
+    } else {
         private _lastUsedTag = GVAR(cachedTags) select {(_x select 0) == _lastUsedTagClass};
         _possibleTags = _lastUsedTag;
     };
 };
 
 // Random X
-if (GVAR(quickTag == 2)) then {
+if ((GVAR(quickTag) == 2) || _useRandom) then {
     private _xTags = GVAR(cachedTags) select {(_x select 0) in ["ACE_XBlack", "ACE_XRed", "ACE_XGreen", "ACE_XBlue"]};
     _possibleTags = _xTags;
 };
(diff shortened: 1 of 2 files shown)


addons/tagging/stringtable.xml
```diff
@@ -26,7 +26,7 @@
             <Chinese>定義噴漆系統預設設定</Chinese>
         </Key>
         <Key ID="STR_ACE_Tagging_QuickTag">
-            <English>Quick Tag</English>
+            <English>Spray Paint - Quick Tag</English>
```
jonpas: Remove other translations?

addons/tagging/functions/fnc_quickTag.sqf
```diff
@@ -22,19 +22,22 @@ if (GVAR(quickTag) == 0) exitWith {};
...
+        _useRandom = true;
+    } else {
         private _lastUsedTag = GVAR(cachedTags) select {(_x select 0) == _lastUsedTagClass};
         _possibleTags = _lastUsedTag;
     };
 };
 
 // Random X
-if (GVAR(quickTag == 2)) then {
+if ((GVAR(quickTag) == 2) || _useRandom) then {
```
jonpas: Lazy eval?

commy2: Might be not worth it for plain booleans.

jonpas: That's actually a good point, I know it's not always worth it especially with `&&`, but that would be interesting to test. And we should really add this more specifically to our guidelines.

Dystopian: it's mentioned at [Code Optimisation](https://community.bistudio.com/wiki/Code_Optimisation#Conditions)

jonpas: Yes, I meant our wiki.

addons/tagging/stringtable.xml
PabstMirror: I just wanted to clarify what the keybind was for
I don't think we should drop the other languages because they might not be updated in time for next release

PR #5688, opened 2017-10-26
Weather - Properly sync overcast value

ulteq:
* Makes sure that all clients share the same (numeric) overcast value
* Only affects the barometric pressure calculation
* Does not affect the actual weather on the clients

Avoids barometric pressure synchronization issues between clients due to overcast drift.

Reference: https://cuel.github.io/

### DIFF
--- a/addons/weather/XEH_postInit.sqf
+++ b/addons/weather/XEH_postInit.sqf
@@ -11,13 +11,14 @@ GVAR(wind_period_start_time) = CBA_missionTime;
 if (!isServer) then {
     "ACE_MISC_PARAMS" addPublicVariableEventHandler {
         TRACE_1("MISC PARAMS PVEH",ACE_MISC_PARAMS);
+        GVAR(currentOvercast) = (ACE_MISC_PARAMS select 0);
         if (GVAR(syncMisc)) then {
-            30 setRainbow    (ACE_MISC_PARAMS select 0);
-            30 setFog        (ACE_MISC_PARAMS select 1);
+            30 setRainbow    (ACE_MISC_PARAMS select 1);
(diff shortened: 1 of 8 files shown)


addons/weather/XEH_postInit.sqf
```diff
@@ -11,13 +11,14 @@ GVAR(wind_period_start_time) = CBA_missionTime;
 if (!isServer) then {
     "ACE_MISC_PARAMS" addPublicVariableEventHandler {
         TRACE_1("MISC PARAMS PVEH",ACE_MISC_PARAMS);
+        GVAR(currentOvercast) = (ACE_MISC_PARAMS select 0);
```
jonpas: I don't know why there are `()` on the right side in the rest of the params, but they are really pointless.

ulteq: We can fix it in the following cleanup PRs.

PabstMirror: This only runs on non server's
So GVAR(currentOvercast) is always 0 on server?

PabstMirror:
imho, instead of syncing with a pub variable "event" in ACE_MISC_PARAMS 
we should just do `missionNamespace setVariable ["ace_weather_overcast", overcast, true]`
skip the need for any sqf to run on the clients

ulteq:
Yes, I'm going to remove those public event handlers entirely in the following cleanup pull request.

addons/weather/XEH_postInit.sqf
ulteq: The server never uses:
```GVAR(currentTemperature), GVAR(currentHumidity) and GVAR(currentOvercast)```

PR #5689, opened 2017-10-26
AB - Revert default simulation interval

ulteq:
* Point of impact shifts measurably (at long ranges) if we skip frames in our velocity update routine
* New players might never touch the config and suffer from less than perfect accuracy without knowing why
* Reverts: #4707

### DIFF
--- a/addons/advanced_ballistics/ACE_Settings.hpp
+++ b/addons/advanced_ballistics/ACE_Settings.hpp
@@ -69,7 +69,7 @@ class ACE_Settings {
         displayName = CSTRING(simulationInterval_DisplayName);
         description = CSTRING(simulationInterval_Description);
         typeName = "SCALAR";
-        value = 0.05;
+        value = 0.0;
     };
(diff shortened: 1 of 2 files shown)


PabstMirror:
0.05 = 20fps which all machines should be capable of hitting
Now that it's unlimited, more fps = more accuracy

Could consistency be better than accuracy here?

e.g.
Shooter has slow machine, spotter has fast
Before they still should have seen similar results

ulteq:
I guess that's a valid argument.

My last test at 2000 m showed the following results (worst case scenario):

AB (0.00): 0.20 Mil high at 60 FPS | 0.05 Mil high at 140 FPS
AB (0.05): 0.75 Mil high at 60 FPS | 0.70 Mil high at 140 FPS

So yes, there seems to be less POI shift with a simulation interval of `0.05`, even though the error is much higher.

I'll try to find another solution first.

ulteq:
Found a better solution

Issue #1062, opened 2015-05-08
Vehicle damage work items


jaynus:
Vehicle damage is coming to fruition in extP3dParser branch. This ticket is to log work items needed for VD

- [x] Code Framework
   - [x] model/vehicle/instance framework
   - [x] background threaded initialization
   - [x] dispatching and messaging arma<->ext
   - [x] SQF result work items
- [x] Model Reading
   - [x] p3d parsing
   - [x] pbo searching/in-game vehicle:model association, registration and loading
   - [x] Animation handling for vehicle state (@noubernou)
- [x] 3d physics world initiation and management
    - [x] model positional transformations (bullet, all models in same world)
- [x] Extension framework and dispatching
- [ ] penetration methods @esteldunedain 
   - [ ] long rod equations
   - [ ] missile two-stage penetration
   - [ ] bullet/shell (conical) round penetration
   - [ ] malformed fragment penetration
   - [ ] HEAT round penetration
- [ ] damage @jaynus
   - [x] vehicle-based damage logging and compounding
   - [x] damage types->hitpoint mappings
   - [ ] damage effects WIP
- [ ] crew kills
- [ ] extended ACE hitpoint definitions 
- [ ] lots of documentation and optionals compat

- [x] Optimization work items
   - [x] bullet-model optimizations
   - [x] GPU offloading vector math
   - [x] animations
   - [x] async dispatching framework

Bummeri:
Just out of curiosity, Why isnt the [RAM](http://forums.bistudio.com/showthread.php?171586-Real-Armor-Mod) vehicle damage system an acceptable way of doing things ?

esteldunedain:
> Just out of curiosity, Why isnt the RAM vehicle damage system an acceptable way of doing things ?

AFAIK the main issue with RAM's approach is that it requires vehicles models to be setup in an specific way, thus being incompatible with most vehicles currently out there. The second issue is that it requires changing projectile properties, which creates compatibility issues too.

jaynus:
work items updated.

john681611:
Is this a dead Milestone Its over a year since any development seems to have been done.

PR #5694, opened 2017-10-27
Extensions: Removed/disable unused stuff

dedmen:
Disabled unused remnants of #1062
C++17 compat by removing stuff that was deprecated in C++14

Buildtime with my changes is 9 seconds and total size of all dlls 1.088.512 Bytes
Before it was 44 seconds and output of 1.093.632 Bytes

This will be required for my upcoming Intercept Plugins because they need C++17 and C++17 doesn't support the stuff that was deprecated in C++14. 

The deprecated stuff is the ptr_fun stuff btw. That was just replaced by 2 lambdas which btw are also more performant than the ptr fun stuff.


dedmen:
@ruthberg for test compile and review if there is anything to review
@jonpas

jonpas:
Looks good to me if it compiles.

PR #5699, opened 2017-10-28
(spectator) unhide/unmute unit on exiting

Cuel:
- unmute/unhide player when exiting spectator

previously doing `[true] call ace_spectator_fnc_setSpectator` and then `[false] call ace_spectator_fnc_setSpectator` would result in a hidden player

I'm not entirely sure this is the best way. What if the player is dead when exiting?


SilentSpike:
Good catch 👍 You should be using the `_hidden` variable though

PR #5702, opened 2017-10-28
Scopes - UI improvement

ulteq:
| Before  | After |
| before1wk9i | after7mjid |

### DIFF
--- a/addons/scopes/stringtable.xml
+++ b/addons/scopes/stringtable.xml
@@ -210,6 +210,14 @@
             <Chinesesimp>在不同高度上会有不同的大气压力</Chinesesimp>
             <Chinese>在不同高度上會有不同的大氣壓力</Chinese>
         </Key>
+        <Key ID="STR_ACE_Scopes_useLegacyUI_displayName">
+            <English>Use legacy UI</English>
+            <German>Vorheriges UI verwenden</German>
+        </Key>
+        <Key ID="STR_ACE_Scopes_useLegacyUI_description">
+            <English>Displays elevation and windage with signed numbers</English>
+            <German>Anzeige der Absehenverstellungen mit vorzeichenbehafteten Zahlen</German>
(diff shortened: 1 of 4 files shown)


PabstMirror:
Stringtables for the letters?

ulteq:
Right!

bux:
That's printed on the RL turrets?

ulteq:
Yes, it's very common that some kind of direction indicator is printed on the turret itself. Together with a little arrow.

Common letterings are, `UP`, `RIGHT`, `LEFT`, `R`, `L`.

ulteq:
But I don't know if we should translate them. At first I thought it's obvious that we should. But I'm not sure anymore.

Laid3acK:
French military SCROME scopes have an `H` (up), `B` (down), `G` (left) and `D` (right) on the turrets:
http://www.scrome.com/assets/templates/flexibility/pdf/Scrome_Marksman_Scope_LTE_Datasheet_GB.pdf

Russian scopes have also their own Cyrillic indications.

ulteq:
I will add string table entries for all four directions.

ulteq:
Should we optionally offer the old "signed" version?

addons/scopes/stringtable.xml
```diff
@@ -365,5 +365,25 @@
             <Chinesesimp>此模块可为高倍率瞄准镜新增归零风偏，距离用的调整纽。</Chinesesimp>
             <Chinese>此模塊可為高倍率瞄準鏡新增歸零風偏，距離用的調整紐。</Chinese>
         </Key>
+        <Key ID="STR_ACE_Scopes_DisplayAdjustmentUp">
```
bux: There seems something wrong with the translations.
https://github.com/acemod/ACE3/pull/5702#issuecomment-340248564

PR #5711, opened 2017-10-30
Ballistics - 7.62x54R muzzle velocity correction

ulteq:
* Closes: https://github.com/acemod/ACE3/pull/5470
* Reference: https://github.com/acemod/ACE3/files/1272583/mousin.barrel.experiment.pdf

@kOepi123, thanks for discovering the issue.


kOepi123:
(y)

PR #5723, opened 2017-11-02
Advanced Fatigue - Various improvements

ulteq:
* Fixes fatigue calculation when walking across structures (bridges, ...)
* Prevents sprinting on very steep terrain
* Adds more in-depth debug output
* Fixes the terrain gradient calculation
* Distinguishes uphill and downhill movements
* Closes #4492

- [ ] Decide if we want to keep the respiratory system
- [ ] Decide if we want to tweak the terrain gradient penalty
- [ ] Optimize the performance of the new code
- [ ] Remove duplicate code

3218m run at VO2max 43.5 ml/kg/min and AnT 75% (live data captured in-game):
ace3_advanced_fatique_test_run

### DIFF
--- a/addons/advanced_fatigue/XEH_postInit.sqf
+++ b/addons/advanced_fatigue/XEH_postInit.sqf
@@ -1,6 +1,32 @@
 #include "script_component.hpp"
 if (!hasInterface) exitWith {};
 
+#ifdef DEBUG_MODE_FULL
+onEachFrame {
+    private _normal = surfaceNormal (getPosWorld ACE_player);
+    private _beg = (getPosWorld ACE_player) vectorAdd (_normal vectorMultiply 0.5);
+    private _end = _beg vectorAdd (_normal vectorMultiply 2.0);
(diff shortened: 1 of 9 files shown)


addons/advanced_fatigue/XEH_postInit.sqf
```diff
@@ -1,6 +1,32 @@
 #include "script_component.hpp"
 if (!hasInterface) exitWith {};
 
+#ifdef DEBUG_MODE_FULL
+onEachFrame {
```
bux: could this be a separate function?
```#ifdef DEBUG_MODE_FULL
call FUNC(renderDebugLines);
#endif
```

so one could run this without having a debug build?

ulteq: I guess we can do that. But the other debug info (`systemChat` and `hintSilent`) is more helpful.

Should we try to move all of the debug output into a function?

Bummeri:
This ticket is relevant to fatigue changes.

thojkooi:
We are happy for this to be merged - will need testing before though.

Mike-MF:

PR #5728, opened 2017-11-04
Don't recompile XEH EH's before every call

dedmen:
Medical XEH_init is always recompiled before being used
Before
First spawn: slack_2017-11-04_18-13-39
Second spawn: arma3_2017-11-04_18-14-53

After:
First spawn: 
Second spawn: arma3_2017-11-04_18-53-49

I'm checking whether I find more stuff that makes the same mistake.

TODO

- [x] ACE_movement inventory displayLoad.
- [x] ACE_Parachute respawn
- [x] ACE_Rearm respawn

inventory displayLoad causes the same problem. 70ms compiling a function everytime you open inventory.


addons/medical/CfgEventHandlers.hpp
```diff
@@ -33,7 +33,7 @@ class Extended_Init_EventHandlers {
 class Extended_Respawn_EventHandlers {
     class CAManBase {
         class ADDON {
-            respawn = QUOTE(call COMPILE_FILE(XEH_respawn));
+            respawn = QUOTE(call FUNC(handleRespawn));
```
commy2: CBA_fnc_addClassEventHandler both in preInit.
This would've avoided this problem in the first place.

dedmen: Should I also use addCEH for displayLoad?

commy2:
lgtm

dedmen:
Done.
Theoretically can do the same with all postInit handlers. They only run once.. But could improve mission start time?

commy2:
Yes. If you do, separate PR please.

dedmen:
I'll leave that to whoever has more time than me to do this :D

addons/medical/XEH_preInit.sqf
```diff
@@ -41,6 +41,13 @@ private _fixStatic = {
 };
 ["StaticWeapon", "init", _fixStatic] call CBA_fnc_addClassEventHandler;
 ["Car", "init", _fixStatic] call CBA_fnc_addClassEventHandler;
+["CAManBase", "Init", FUNC(handleInit)] call CBA_fnc_addClassEventHandler;
+["CAManBase", "Respawn", FUNC(handleRespawn)] call CBA_fnc_addClassEventHandler;
+["CAManBase", "Killed", FUNC(handleKilled)] call CBA_fnc_addClassEventHandler;
+["CAManBase", "Local", FUNC(handleLocal)] call CBA_fnc_addClassEventHandler;
+
+["ACE_bodyBagObject", "init", EFUNC(dragging,initObject)] call CBA_fnc_addClassEventHandler;
```
PabstMirror: dragging does 
```
    class ThingX {
        class ADDON {
            init = QUOTE(_this call DFUNC(initObject));
```
and `"ACE_bodyBagObject" isKindOf "ThingX" = true`
so I think we can drop this

commy2: OK

dedmen: I didn't see a double init of dragging when profiling.. Maybe I just didn't look close enough though.

PR #5630, opened 2017-10-17
Add ability to pull dead body out of vehicle

Dystopian:
**When merged this pull request will:**
- title;
- disable interaction with crew of dead vehicle;
- disable `Unload patient` medical action when he's dead.

Action is added to `Passengers` submenu.
Getting in right seat is used to pull out dead body - looks similar to RL.
Tested also on RHS vehicles with openable doors.

As usually English mistake remarks are welcomed.
Thanks to #5626, old dreams come true.

PR #5733, opened 2017-11-07
QuickMount - Use correct cargo index

PabstMirror:
ref @Dystopian / https://github.com/acemod/ACE3/pull/5630#discussion_r149043888


commy2:
Pretty sure you can just subtract the amount of person turrets from the command cargo index to get the action cargo index.

PabstMirror:
Maybe, might break if they're in a weird order like `[FFV,Cargo,FFV,Cargo]`

Dystopian:
I wanted to fix it :(

As for PR, I wouldn't use another `forEach fullCrew` inside the same cycle. External cycle is enough to count cargo action index.

jonpas:
@Dystopian If you would like to, feel free to PR the improvement, merged that as it fixes things.

PR #5735, opened 2017-11-09
Spectator - Use ACRE2 add display passthrough keys API

jonpas:
- ACRE2 v2.5.1 will have `acre_api_fnc_addDisplayPassthroughKeys` function (https://github.com/IDI-Systems/acre2/pull/437) for better compatibility with all spectator systems. It will hook all required buttons (only "Mute Spectators" as of this time) and also support CBA's new multiple keybinds.

### DIFF
--- a/addons/spectator/functions/fnc_ui_handleLoad.sqf
+++ b/addons/spectator/functions/fnc_ui_handleLoad.sqf
@@ -1,0 +1,25 @@
+/*
+ * Author: SilentSpike, Jonpas
+ * Function used to handle load event.
+ *
+ * Arguments:
+ * 0: Spectator display <DISPLAY>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * _this call ace_spectator_fnc_ui_handleLoad
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_display"];
+
+uiNamespace setVariable [QGVAR(display), _display];
+
+// Handle ACRE2 Toggle Spectator (if present)
+if (["acre_sys_radio"] call EFUNC(common,isModLoaded)) then {
+    [_display] call acre_api_fnc_addDisplayPassthroughKeys;
+};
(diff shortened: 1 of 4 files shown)


jonpas:
**Requires ACRE2 v2.5.1 (not yet released)!**

addons/spectator/functions/fnc_ui_handleLoad.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_display"];
+
+uiNamespace setVariable [QGVAR(display), _display];
+
+// Handle ACRE2 Toggle Spectator (if present)
+if (["acre_sys_radio"] call EFUNC(common,isModLoaded)) then {
```
commy2: `!isNil "acre_api_fnc_addDisplayPassthroughKeys"` and merge now, not having to wait for ACRE update + bwc?

jonpas: Have to add it here and `isNil` where it was handled manually before.
Previous manual handling will continue to work after ACRE2 update, without any change from how it currently works (only supports first keybind).

bux:
I'd find it cleaner to have a global var where other mods could register their functions. Instead of calling mod specific functions ourselves.

Something like this:

```sqf
if ( count GVAR(uiLoadedHooks) > 0 ) then {
    {
        // assert if function
        call _x;
    } forEach GVAR(uiLoadedHooks);
};
```

and in ACRE 
```sqf
ace_spectator_uiLoadedHooks pushBack (/*<code to be executed>*/);
```

SilentSpike:
We're already calling ACRE/TFAR functions in `fnc_setSpectator` anyway

jonpas:
We can easily have ACE Spectator support in ACRE with that API function. I simply put it here, because we already handle both TFAR and ACRE, so it made sense to keep it like that.

jonpas:
API function is live with ACRE v2.5.1.

addons/spectator/functions/fnc_ui_handleLoad.sqf
jonpas: Do we want that backwards compatibility? I don't really see a reason to be honest, people should switch to updated version of ACRE2.

PR #5738, opened 2017-11-10
Scopes - Rifle zeroing improvements

ulteq:
* Fixed vertical point of impact shift with different initSpeeds / airFrictions (vanilla bug)
* Rifle/Ammo precision is now following a bivariate normal distribution

| Vanilla | ACE3 |
| -------- | -------- |
| dispersion_vanilla | dispersion_ace3 |

Instead of trying to undo the vanilla zeroing we now overwrite the velocity vector of the projectile with the current `weaponDirection` and apply an angular offset (`ACE_RailBaseAngle`).

*TODO:*
- [ ] Make the `dispersion` config lookup more robust
- [ ] Figure out how to handle non local units
- [ ] Test with 3rd party rifles
- [ ] Rebuild AB DLLs

### DIFF
--- a/addons/main/script_macros.hpp
+++ b/addons/main/script_macros.hpp
@@ -111,6 +111,8 @@
 
 #define GRAVITY 9.8066
 
+#define SD_TO_MIN_MAX(d) ((d) * 3.371) // Standard deviation -> min / max of random [min, mid, max]
+
 // Angular unit conversion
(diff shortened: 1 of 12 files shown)


PabstMirror:
I think `weaponDirection` is very inaccurate in multiplayer on non-local units.

ulteq:
That's bad news. I'll look into it. One solution would be to use the old method for non local units.

ulteq:
We can't use `weaponDirection` as reference, since it needs to work for both local and non local units.

PR #5741, opened 2017-11-11
Wirecutter - Add global event when cutting fences

nomisum:
This hook could be used to script things happening when player 'touches' fence when wirecutting. Would reduce overhead if you want to script electric fences (no need for distance checks).

- add global event to wirecut start

### DIFF
--- a/addons/logistics_wirecutter/functions/fnc_cutDownFence.sqf
+++ b/addons/logistics_wirecutter/functions/fnc_cutDownFence.sqf
@@ -56,6 +56,6 @@ private _progressCheck = {
     ((!isNull _fenceObject) && {(damage _fenceObject) < 1} && {("ACE_wirecutter" in (items _unit))})
 };
 
-["ace_wireCuttingStarted", [_unit, _fenceObject]] call CBA_fnc_globalEvent;
-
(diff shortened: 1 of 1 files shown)


Cuel:
Why not local event?

nomisum:
i thougt about that. for one i think the server should handle stuff happening plus there wont be frequent calls as wirecutting is done few times a mission.

but please correct me if local is more plausible.

addons/logistics_wirecutter/functions/fnc_cutDownFence.sqf
```diff
@@ -56,4 +56,6 @@ private _progressCheck = {
     ((!isNull _fenceObject) && {(damage _fenceObject) < 1} && {("ACE_wirecutter" in (items _unit))})
 };
 
+["ace_wireCutStart", [_unit, _fenceObject]] call CBA_fnc_globalEvent;
```
jonpas: `"ace_wireCuttingStarted` to fit the naming of other events?

commy2:
Why's the event before the progress bar starts? Would be more useful afterwards.

nomisum:
It just seemed easier to implement, I would have had to define a timespan after which the EH would fire and the only thing I need is a "this guy begins to cut". 

Actually I could live with a beginning progress bar as long as there is no way someone could start and exit before the EH fires. I assume no one would see the progress bar for that split second before he gets roasted anyway ;)

PR #5746, opened 2017-11-12
Remove options menu

PabstMirror:


commy2:
Travis died?

bux:
> Done. Your build exited with 0.

jonpas:
Seems like problems with webhooks returning, takes a while since yesterday.

thojkooi:
I think this should end up in the change log.

PR #5749, opened 2017-11-12
Restore some modules set to scope = 1

PabstMirror:
These aren't setting modules and shouldn't be disabled


addons/fastroping/CfgVehicles.hpp
```diff
@@ -18,7 +18,7 @@ class CfgVehicles {
     };
     class ACE_Module: Module_F {};
     class ACE_moduleEquipFRIES: ACE_Module {
-        scope = 1;
+        scope = 2;
```
jonpas: I disabled that one because there is the Eden attribute for it. We decided to scrap modules in favor of Eden attributes some time ago.

PR #5751, opened 2017-11-12
Scopes - Hide 'up' direction lettering

ulteq:
* Improves readability


ulteq:
That is how I did it before we agreed on having a lettering for each case. But I can change it if you want.

jonpas:
I see no point in having `%1` for each language, unless you plan to add a setting in the future. But I think that's not needed at all.

PR #5752, opened 2017-11-12
Dev helper function to convert settings to code

PabstMirror:
Should make hand conversion a little easier, still have to manually match stringtables

produces code like this:
```
[QGVAR(reflectionsEnabled), "CHECKBOX",
[LSTRING(), LSTRING()], // Explosion Reflections Simulation, Enable the ACE Explosion Reflection Simulation
localize LSTRING(), // ACE - Fragmentation Simulation
false, // // default value
true, // isGlobal
{["ace_frag_reflectionsEnabled", _this] call EFUNC(common,cbaSettings_settingChanged)}] call CBA_settings_fnc_init;

[QGVAR(maxTrack), "SLIDER",
[LSTRING(), LSTRING()], // Maximum Projectiles Tracked, This setting controls the maximum amount of projectiles the fragmentation and spalling system will track at any given time. If more projectiles are fired, they will not be tracked. Lower this setting if you do not want FPS drops at high-count projectile scenarios ( >200 rounds in the air at once)
localize LSTRING(), // ACE - Fragmentation Simulation
[-1,5000,10,1], // [min, max, default value, trailing decimals (-1 for whole numbers only)]
true, // isGlobal
{["ace_frag_maxTrack", _this] call EFUNC(common,cbaSettings_settingChanged)}] call CBA_settings_fnc_init;
```


jonpas:
`toString [13, 10]` intsead of `endl` might be better, as `endl` gives `\r\n`. Add indentations maybe?

We should also decide on common style for this to begin with.

PabstMirror:
Example conversion: https://github.com/acemod/ACE3/pull/5581/files

PR #5761, opened 2017-11-13
Categorize Overheating

ulteq:
Should we call the category `ACE - Weapon Overheating`?

### DIFF
--- a/addons/overheating/ACE_Settings.hpp
+++ b/addons/overheating/ACE_Settings.hpp
@@ -1,5 +1,6 @@
 class ACE_Settings {
     class GVAR(displayTextOnJam) {
+        category = CSTRING(DisplayName);
         typeName = "BOOL";
         isClientSettable = 1;
         value = 1;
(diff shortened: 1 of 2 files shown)


PabstMirror:
Right now the converter will add the `ACE - ` if it doesn't exist.
For final settings I think we'll want it in a single string for simplicity;
so might as well add it now

ulteq:
> Right now the converter will add the ACE - if it doesn't exist.

Yes. But the module is currently called `Overheating`. `ACE - Overheating` is very ambiguous in my opinion.

jonpas:
Wouldn't that generally fit under `ACE - Weapons`?

ulteq:
And what else would you put under `ACE - Weapons`?

jonpas:
Anything to do with a weapon? 😆 
I'd have to check what settings there are, I think we're fine with whatever you do, that can always be tweaked later.

ulteq:
> I think we're fine with whatever you do, that can always be tweaked later.

I'd merge it as is.

PR #5762, opened 2017-11-13
Tweaks for Linux extensions

BaerMitUmlaut:
- Make some minor tweaks to allow compiling the extensions on Linux
- Conform SQF extensions check for Linux server extensions support
  - Sub-class extensions with `windows`, `linux`, `client` and `server` properties
- Add `*.so` support to build tools 

~~I had to disable the advanced ballistics extension, the errors I was getting were in regards to a constexpr which I assume was because the lack of sufficient C++17 support in my gcc version (5.4). If somebody else wants to give it a try, be my guest.~~


extensions/common/shared.hpp
```diff
@@ -20,6 +20,12 @@
 #include "ace_version.hpp"
 #include <algorithm>
 #include <cctype> //std::isspace
+#include <cstring>
+#include <istream>
+
+#ifndef _WIN32
+#define strncpy_s(dest, size, src, ignored) strncpy(dest, src, size)
```
jonpas: I would prefer if we just disable MSVC "safe" warnings and use proper C++ functions everywhere.

dedmen: strncpy_s is C.
I guess the proper C++ std::strncpy function should work crossplatform.

BaerMitUmlaut: Does std::strncpy throw warnings in MSVC too or are they okay with that one?

jonpas: I think it might throw warnings as well, but I am not sure.

dedmen: It might throw unchecked iterators warning/critical error. But as it takes char* and not iterators it really shouldn't.

dedmen:
I'd still recommend Git LFS.
The Linux extensions add 5.6MB to the repo.
And the Linux extensions are all bigger than the Windows ones. I wonder why.

jonpas:
Probably wrong compiler flags, I did not look into that, just built them for testing.

extensions/common/shared.hpp
dedmen: On Intercept I use `std::copy_n(src, len, dest)` which doesn't automatically add the ending null char which I think strncpy does. I just set it manually. like `dest[ignored] = 0;`

dedmen:
Just got told by a dev working on the port that extension support is not there yet? I thought it was 🤔
Neither dll nor .so work on ports.

BaerMitUmlaut:
Which is weird, they definitely work to some degree, I've tested them.

fbt:
The linux dedicated server has .so support, but not the client, apparently.
Were you testing on the server, by chance?

thojkooi:
This will allow us to drop the sqf implementation for ACE medical wounds.

dedmen:
For reference on compiler flags to push the size down:
https://github.com/dedmen/ArmaScriptProfiler/blob/master/src/CMakeLists.txt#L104

jonpas:
Built extensions work:
```
"ace_medical" callExtension "version"
-> "3.12.0-af807b6"
```

Didn't manage to reduce file size yet though. Statically linking `libgcc` and `libstdc++` takes majority of space but is pretty much required.

@dedmen's solution should work but I had trouble adding it to our extension setup so far.

dedmen:
>Statically linking libgcc and libstdc++ takes majority of space but is pretty much required.

I assume the linker doesn't optimize them and just takes them as a whole.

I'll do some testing, see if I can get size reduction.

TheMagnetar:
Yes. At the moment is only used for medical. However, if we can get rid of microsoft C/C++ extensions it would be better.

dedmen:
>I had to disable the advanced ballistics extension, the errors I was getting were in regards to a constexpr which I assume was because the lack of sufficient C++17 support in my gcc version (5.4). If somebody else wants to give it a try, be my guest.

compiled just fine for me. on GCC 9 trunk. Not even a single warning.

Wow.. without static libc and libc++ medical goes down to 54K.


>However, if we can get rid of microsoft C/C++ extensions it would be better.

What do you mean by that?

TheMagnetar:
functions like strncpy_s DWORD, BYTE and so on

jonpas:
Yeah Dedmen, I told you that site difference on Slack.

Compiling all is just easier and future proof. So I would like if we do that.

dedmen:
Okey I got a compile down with my own libstdc++ build.
58K. And seems to work
putty_2019-03-26_12-36-30

The dev-env setup for that is quite extensive. But it seems to work.
I can only test on my own server on debian squeeze so here:

[ace_medical.zip](https://github.com/acemod/ACE3/files/3007950/ace_medical.zip)

dedmen:
```
wolf@ns3032530 ~/projects/ACE3/extensions $ ls -lh * | grep .so
-rwxr-xr-x 1 wolf wolf  50K Mar 26 12:40 ace_advanced_ballistics.so
-rwxr-xr-x 1 wolf wolf  14K Mar 26 12:40 ace_break_line.so
-rwxr-xr-x 1 wolf wolf 5.2K Mar 26 12:40 ace_clipboard.so
-rwxr-xr-x 1 wolf wolf  18K Mar 26 12:40 ace_fcs.so
-rwxr-xr-x 1 wolf wolf  58K Mar 26 11:18 ace_medical.so
-rwxr-xr-x 1 wolf wolf 5.3K Mar 26 12:40 ace_parse_imagepath.so
```
Seems good enough for me :D 

>Compiling all is just easier and future proof. So I would like if we do that.

Future proof? you mean for the suspended linux client port while running the windows version via proton already works perfectly fine and while arma is already at end of life?


For example clipboard does absolutely nothing. all the functionality is in a #ifdef WIN32. So we are including a extension that does nothing, doesn't make much sense.

jonpas:
I was also thinking doing stuff server-side, but some extensions won't ever be used for sure, so we could ignore those.

Have to add additional checks in SQF extension check tool, but that's easy.

---

About this libc++ compiling, I really don't want anyone having to do that, not even for release builds, 1 MB is still nothing, and if we only do some extensions (or even only 1) then that's a non-issue IMO.

Good to know it can be done though, interesting stuff.

dedmen:
>About this libc++ compiling

Doesn't really have to be done everytime.

This is the magic line in my buildscript:
`set(CMAKE_SHARED_LINKER_FLAGS "-static-libgcc -fdata-sections -ffunction-sections -Wl,--gc-sections /home/wolf/projects/ACE3/extensions/libstdc++.a -nodefaultlibs -Wl,-trace")`

https://github.com/acemod/ACE3/blob/0e8728129f62f7890f81c2341f240a7fb115bb95/extensions/CMakeLists.txt#L25

`/home/wolf/projects/ACE3/extensions/libstdc++.a` That's the kicker. 
We could just ship our own libstdc++ build in the repo if we wanted to. That's 9MB though..
But I might be able to get that down a notch by removing all debug info.

jonpas:
So you'd rather add 9 MB of yet another file than have 1 MB extension with library inside of it?

dedmen:
Only if you'd still stay on wanting all the libraries built for linux.
With all extensions present, we would have the 9MB back in after the second update of the extensions.

jonpas:
> we would have the 9MB back in after the second update of the extensions.

Final zip size would not change. And git history is really not a concern, run `git gc --aggressive` from time to time and you are good. GitHub reports it as less than 400 MB in size.

jonpas:
Meeting note: Compile only medical extension, ignore unbuilt ones on Linux server in SQF checker.

extensions/common/shared.hpp
jonpas: Will disable MSVC "unsafe" warnings for all, but in separate PR.

jonpas: -> #7171

jonpas:
**Note:** Build extension in `medicalExtensionWork` after all other changes are part of that as well.

jonpas:
Confirmed working on Linux server with latest changes:
```
20:29:38 CallExtension loaded:   ace_medical (/home/jonpas/Data/Games/SteamLib/steamapps/common/arma3/mods/ACE3/ace_medical.so) [������]
20:29:38 [ACE] (common) INFO: Extension version: ace_medical: 3.12.3-14307c7
```
Version in `[]` is not showing, because Linux doesn't have metadata like Linux which Arma reads.

addons/medical_damage/config.cpp
```diff
@@ -19,5 +19,8 @@ class CfgPatches {
 #include "CfgEventHandlers.hpp"
 
 class ACE_Extensions {
-    extensions[] += {"ace_medical"};
+    class ace_medical {
+        windows = 1;
```
PabstMirror:
```suggestion
        windows = 0;
```

addons/medical_damage/config.cpp
```diff
@@ -19,5 +19,8 @@ class CfgPatches {
...
+        linux = 1;
```
PabstMirror: not used for now
```suggestion
        linux = 0;
```

addons/advanced_ballistics/config.cpp
```diff
@@ -20,5 +20,8 @@ class CfgPatches {
 #include "ACE_Settings.hpp"
 
 class ACE_Extensions {
-    extensions[] += {"ace_advanced_ballistics"};
+    class ace_advanced_ballistics {
+        windows = 1;
+        client = 1;
```
jonpas: Is this correct?

addons/fcs/config.cpp
```diff
@@ -28,7 +28,10 @@ class CfgPatches {
 #include "CfgOptics.hpp"
 
 class ACE_Extensions {
-    extensions[] += {"ace_fcs"};
+    class ace_fcs {
+        windows = 1;
+        client = 1;
```
jonpas: Is this correct?

addons/interact_menu/config.cpp
```diff
@@ -23,5 +23,12 @@ class CfgPatches {
 #include "ACE_Settings.hpp"
 
 class ACE_Extensions {
-    extensions[] += {"ace_break_line", "ace_parse_imagepath"};
+    class ace_break_line {
+        windows = 1;
+        client = 1;
```
jonpas: Is this correct? Could this be used on server theoretically?

addons/interact_menu/config.cpp
```diff
@@ -23,5 +23,12 @@ class CfgPatches {
...
+    };
+    class ace_parse_imagepath {
+        windows = 1;
+        client = 1;
```
jonpas: Is this correct? Could this be used on server theoretically?

addons/optionsmenu/config.cpp
```diff
@@ -30,7 +30,10 @@ class CfgAddons {
 #include "gui\pauseMenu.hpp"
 
 class ACE_Extensions {
-    extensions[] += {"ace_clipboard"};
+    class ace_clipboard {
+        windows = 1;
+        client = 1;
```
jonpas: Is this correct? Could this be used on server theoretically?

dedmen: yes. you could copy to the servers clipboard if it were running on a windows server.
But I'd see that more as a security issue than a feature.

jonpas: How is that more of a security issue than on client?

jonpas:
One change this PR makes is it disables checks on server, something we currently actually do. I guess that's not a problem as anyone can still use it if they really want to, but the mission load is still faster due to less extension loading. So what @dedmen mentions above, we don't actually fix a security issue, we just officially don't support it (and won't build them for that platform).

This is ready to merge into `medicalExtensionWork` (changes from that branch included and required for this PR). No requirement to rebuild DLLs at this time.

PR #5764, opened 2017-11-13
Change key items mass from 0 to 0.1

alganthe:
- Change ace keys and lockpick to have 0.1 mass instead of 0.

### DIFF
--- a/addons/vehiclelock/CfgWeapons.hpp
+++ b/addons/vehiclelock/CfgWeapons.hpp
@@ -11,7 +11,7 @@ class CfgWeapons {
         picture = QPATHTOF(ui\keyBlack.paa);
         scope = 2;
         class ItemInfo: CBA_MiscItem_ItemInfo {
-            mass = 0;
+            mass = 0.1;
         };
     };
     class ACE_key_lockpick: ACE_key_master {


PabstMirror:
The reason they were 0 is because they get added to player's inventory and I didn't want to worry about a unit having no space left.

But I guess 0.1 is still small enough it shouldn't be a problem

alganthe:
thing is with 0 you can have thousands of them, it's not an issue with BIS arsenal because adding items so slow but with ACE3 arsenal I can have 10 000 under a minute.

commy2:
I'd say you should fix your arsenal instead and cap the number of items for the "fill all" option or whatever you're doing. There could be many more items with no mass in 3rd party mods out there and this won't help at all for those.

alganthe:
Then you'd require to do the same for vanilla arsenal, it's slow but still possible.

commy2:
The slowness makes it not feasible though.

jonpas:
Why is that a problem either way, tbh if people are so stupid to do that they deserve a crashed server. For public servers and whatnot you can add a setting (maybe even default enabled) which caps amount of one item to something like 100.

commy2:
No! Not every small thing needs a setting. Just limit all items to idk. 100 or whatever and you fix it very elegantly. There is no use in having more than enough of any type of item, so a setting would be bloat.

commy2:
That said, all items should have a weight. And if you worry about the inventory being full then consider something like: https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_addItem.sqf which spills the item on the ground if needed.

PabstMirror:
This uses magazineDetail ID stuff, so I can't just add it to a ground container

PabstMirror:
Still needed, or fix in arsenal?

alganthe:
still "needed", it doesn't break anything per se, it's just a really bad idea to have items with 0 mass.

alganthe:
keys are considered unique items in ace arsenal now as far as i'm aware, closing.

PR #5765, opened 2017-11-14
Remove ace_server.pbo

PabstMirror:
no longer supported in 3.12


654wak654:
o7 RIP old friend.

thojkooi:
This should receive extra attention in the change log, as it will require additional action to take place on the user's side.

Do we have a migration guide for switching from ACE server to CBA server settings?.

jonpas:
We will have to put a post up with all the information. There is no general guide, basically go to CBA Settings on the server and set things (and overwrite mission if wanted), that's it.

PR #5774, opened 2017-11-15
Make.py Mikero tools detection fix

Djzonk:
- Changes the find_depbo_tools function to look inside both root keys for each tool individually instead of as a group.
    - Allowing for situations where some keys for the tools would be in different root keys 

Maybe an edge case issue but I don't see a reason not to fix it.
_Will also be making PR to project template_

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -228,57 +228,31 @@ def find_bi_tools(work_drive):
         raise Exception("BadTools","Arma 3 Tools are not installed correctly or the P: drive needs to be created.")
 
 
-def find_depbo_tools(regKey):
+def find_depbo_tools():
     """Use registry entries to find DePBO-based tools."""
-    stop = False
+    requiredToolPaths = {"pboProject": None,"rapify": None,"MakePbo": None}
+    failed = False
 
-    if regKey == "HKCU":
-        reg = winreg.ConnectRegistry(None, winreg.HKEY_CURRENT_USER)
-        stop = True
-    else:
-        reg = winreg.ConnectRegistry(None, winreg.HKEY_LOCAL_MACHINE)
-
-    try:
+    for tool in requiredToolPaths:
         try:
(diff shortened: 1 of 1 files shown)


tools/make.py
```diff
@@ -228,57 +228,31 @@ def find_bi_tools(work_drive):
...
-            print_error("Could not find rapify.")
-
-        try:
-            k = winreg.OpenKey(reg, r"Software\Wow6432Node\Mikero\MakePbo")
-        except FileNotFoundError:
-            k = winreg.OpenKey(reg, r"Software\Mikero\MakePbo")
-        try:
-            makepbo_path = winreg.QueryValueEx(k, "exe")[0]
+            print_error("Could not find {}".format(tool))
+            failed = True
```
SilentSpike: Any reason you aren't just raising the exception here? It's been a while since I've used python in any depth

Djzonk: it raises an exception on line 252 for the failure after the 3 checks for the tools are made. It's more for ease of debugging why it won't run as it will tell you exactly what it can/cant find

SilentSpike: Fair 👍 I can't see any issues with the code, will let @jonpas handle approval to put a second set of eyes over it

tools/make.py
```diff
@@ -228,57 +228,31 @@ def find_bi_tools(work_drive):
...
-            k = winreg.OpenKey(reg, r"Software\Wow6432Node\Mikero\pboProject")
-        except FileNotFoundError:
-            k = winreg.OpenKey(reg, r"Software\Mikero\pboProject")
-        try:
-            pboproject_path = winreg.QueryValueEx(k, "exe")[0]
-            winreg.CloseKey(k)
-            print("Found pboproject.")
-        except:
-            print_error("Could not find pboProject.")
+    requiredToolPaths = {"pboProject": None,"rapify": None,"MakePbo": None}
```
jonpas: Space after `,`.

jonpas:
We'll see if it breaks when I try to do a build. 😆

Issue #5039, opened 2017-03-26
Scopes are locked when not using Adv. Ballistics


JBartlett108:
**Arma 3 Version:** `1.68` (stable)
**CBA Version:** `3.2.1` (stable)
**ACE3 Version:** `3.9.1` (stable)

**Mods:**
- CBA_A3
- ace
- RHS USAF 4.2.1 (sometimes used and tested)
- RHS AFRF 4.2.1 (sometimes used and tested)
- RHS GREF 4.2.1 (sometimes used and tested)
- RHS SAF 4.2.1 (sometimes used and tested)

**Description:**
- With advanced ballistics module placed, but everything on OFF, high power scopes cannot range and are locked at 100m. Same thing happens when there are absolutely NO MODULES on the mission, or when it is configured to my unit's liking (medical, logistics, etc.)

**Steps to reproduce:**
- Add playable player with any sniper rifle (M107, M200, M24, etc.) and give any high power scope (M84A1, AMS, LPRS, etc.). Add in Advanced Ballistics module and turn EVERYTHING off. Off for everyone, off for snipers, off on full-auto, everything. No matter what way you try to play it, you cant zero your scope. The number stays locked at 100m, and there's no movement when ADSing. Try it with any combination of weapon, scope... hell, even ammo. It stays locked. The same happens if you delete the module, save the mission, open another mission, and then re-open the original (to clear it from Eden). Also tried with and without RHS and their compats, to no avail.

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Editor (Singleplayer) / Editor (Multiplayer)

It occurred on our dedicated "Operations" server, with and without a headless client, when I exported it to MP and hosted it on LAN, and when I ran it from Eden in both singleplayer and multiplayer.

**Placed Modules:**
- Advanced Ballistics, none, and nearly all of them (tested with each config, same result).

**RPT log file:**
- Will get ASAP

AACO:
When you're ADS do you hear clicking? And when you're not zoomed in, do you get an adjustment knob with a number readout? If so, that's part of the scopes framework (not advanced balistics) https://github.com/acemod/ACE3/tree/master/addons/scopes

You'll have to place the scopes module down and disable those settings from there.

JBartlett108:
Negative on both, there's simply no adjustment period

PabstMirror:
Check your keybinds for scope adjust
Controls -> Configure Addons -> Ace Scope Adjust
Should default to PageUp/Down

[...]

PR #5787, opened 2017-11-17
Scopes - Adds simplified zeroing subsystem

ulteq:
* Client side setting to choose the preferred zeroing method
* Replicates the behavior of the vanilla zeroing system

This allows you to zero your scope in 50 meter increments if you don't like the current adjustment turrets.


PabstMirror:
Close #5039 (Gives the user an experience much more like vanilla zeroing)

ulteq:
> Also, I think this shouldn't be user settable as it's more of a difficulty setting
Agreed.

PR #5794, opened 2017-11-20
Remove parachute landing animation

### LINKED ISSUE #5596: Rework ace_parachute (landing-animation / vanilla altimeter)
**Arma 3 Version:** `1.76.143187` (stable)
**CBA Version:** `v3.4.1.170912` (stable)
**ACE3 Version:** `v3.10.2.22` (stable)

**Mods:**
```
- CBA_A3
- ACE3
- ACEX
- ACRE2
- HeadRangePlus
```

**Description:**
- The ACE Animation for landing a parachute was implemented to replace the horrible version of the vanilla game where you allways landed with your face in the dirt. Since then, the vanilla animation was overhauled and is now very nice and usable.
- Despite having the ACE_Altimeter linked in the inventory, the vanilla altitude and speed infobar at the top of the screen still is visible

**Steps to reproduce:**
- Place a playable unit with:
  - a parachute
  - sufficient height to freefall and open the chute
  - an altimeter watch
- preview the mission
  - recognize the vanilla altitude and speed infobar at the top of the screen
  - open the parachute and land
- repeat without mods
  - see how the parachute collapses behind the unit instead of disappearing
  - no bellyflop, "proper" landing 

- tested on Dedicated and in Editor (Singleplayer)

**Placed Modules:**
- None

SilentSpike:
The vanilla animation has since been reworked and is an improvement.

I believe the PFH used to track when the player lands could also be removed by adjusting how the variables that track chute cutting and reserve chute are used. ~~Beyond the scope of this PR though.~~

Closes #5596


addons/parachute/XEH_postInit.sqf
```diff
@@ -33,7 +33,7 @@ GVAR(PFH) = false;
 ["vehicle",{
     if (!GVAR(PFH) && {(vehicle ACE_player) isKindOf "ParachuteBase"}) then {
         GVAR(PFH) = true;
-        [FUNC(onEachFrame), 0.1, []] call CALLSTACK(CBA_fnc_addPerFrameHandler);
+        [FUNC(onEachFrame), 0.1, []] call CBA_fnc_addPerFrameHandler;
     };
 }] call CBA_fnc_addPlayerEventHandler;
```
commy2: The whole thing can be deleted I think.

commy2:
Btw. this is horrible: https://github.com/acemod/ACE3/blob/master/addons/parachute/functions/fnc_handleRespawn.sqf#L17

addons/parachute/XEH_postInit.sqf
SilentSpike: It needs to run to determine when the player lands to reset the chute cut variable. Which is what I mention in the OP about if we change the way those variables are handled we can totally remove the PFH.

SilentSpike:
Yeah I also saw that and actually think it might be slightly broken since it's running via a CBA XEH on CAManBase and not just on the player respawn

addons/parachute/XEH_postInit.sqf
SilentSpike: (At least, I'm assuming that variable being reset is necessary. It's quite hard to determine what the original intent is with some of the code)

commy2:
I still don't get what we need `chuteIsCut` for.

SilentSpike:
I think it's so that a reserve chute is not added if the players uses a reserve chute which is configured to have a reserve, but not entirely sure myself

ulteq:
> I still don't get what we need `chuteIsCut` for.

https://github.com/acemod/ACE3/blob/98da86b74d4089bc3403d0b0c0074e6a46ce35d8/addons/parachute/CfgVehicles.hpp#L18

SilentSpike:
Aha so that's where it's used! I did a ctrl+shift+f for the function but it didn't turn up that instance. Thanks 😄

SilentSpike:
Think this should now work as intended now without the arbitrary code complexity, but haven't had a chance to test yet

SilentSpike:
Got around to testing this, seems the cut parachute action is broken because I overlooked that the `hasReserve` variable is reset to false once the reserve is added. Fixing now.

SilentSpike:
Fixed and tested, this should be good to go 👍

SilentSpike:
Actually 🤔 I think if player was to pull chute and never cut it, then the `canCut` variable would still be true on landing. If they were to then jump out of another plane with their reserve still on then they would then be able to cut that too and fall to their death.

The simple way to avoid all this nonsense would be to let players cut any parachute, but then we'd have to provide some indication that they have a reserve because there's no way to check while in the air.

SilentSpike:
Okay fixed all the issues now and massively reduced the complexity of the implementation 👍

Will let @commy2 handle the vanilla UI stuff. Only thing I did notice in testing is that when you cut your chute you don't seem to return to the falling animation, but that isn't a result of this PR and must be a longstanding bug.

commy2:
UI stuff is fixed here: ref #5836

addons/parachute/functions/fnc_showAltimeter.sqf
```diff
@@ -17,32 +17,33 @@
...
     private _hour = floor daytime;
     private _minute = floor ((daytime - _hour) * 60);
 
     private _height = ((getPosASL _unit) select 2) + EGVAR(common,mapAltitude);
     private _curTime = CBA_missionTime;
     private _timeDiff = _curTime - _prevTime;
-    private _descentRate = if(_timeDiff > 0) then {floor((_oldHeight - _height) / _timeDiff)} else {0};
+    private _descentRate = if (_timeDiff > 0) then {floor((_oldHeight - _height) / _timeDiff)} else {0};
 
-    _TimeText ctrlSetText (format ["%1:%2",[_hour, 2] call EFUNC(common,numberToDigitsString),[_minute, 2] call EFUNC(common,numberToDigitsString)]);
```
PabstMirror: We want the old way, it pads zeros to the left
[5, 2] call ace_common_fnc_numberToDigitsString = "05"
5 toFixed 2 ="5.00"

PR #5796, opened 2017-11-20
Minor performance tweaks

ulteq:
* Medical_ai - Optimize 'fnc_isInjured'
* Laserpointer - Optimize 'fnc_onDraw'
* Overheating - firedEH: Skip ground detection if possible
* Overheating - firedEH: Replace function call with inline code
* Overheating - firedEH: Nest abort conditions

### DIFF
--- a/addons/laserpointer/functions/fnc_onDraw.sqf
+++ b/addons/laserpointer/functions/fnc_onDraw.sqf
@@ -15,8 +15,7 @@
  */
 #include "script_component.hpp"
 
-// no lasers in thermal mode
-if !(GVAR(isTI)) then {
+if (count GVAR(redLaserUnits) + count GVAR(greenLaserUnits) > 0 && {!GVAR(isTI)}) then {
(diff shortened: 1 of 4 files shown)


addons/laserpointer/functions/fnc_onDraw.sqf
```diff
@@ -16,7 +16,7 @@
 #include "script_component.hpp"
 
 // no lasers in thermal mode
-if !(GVAR(isTI)) then {
+if (!GVAR(isTI) && {(count GVAR(redLaserUnits)) + (count GVAR(greenLaserUnits)) > 0}) then {
```
commy2: Might
```sqf
&& {!([GVAR(redLaserUnits), GVAR(greenLaserUnits)] isEqualTo [[],[]])}
```
be even faster?

ulteq: I will try.

PR #5798, opened 2017-11-20
Scopes - Added new config entry to fix the iron sight zeroing

ulteq:
Fixes the zeroing of all vanilla weapon iron sights.


ulteq:
Might not be the best possible solution.

PR #5799, opened 2017-11-20
Scopes - wiki/framework update

ulteq:
Documents #5747 and #5804


bux:
Any info on how to measure those values?

ulteq:
Trial and error, unless you write yourself a helper script.

This mission also helps a lot: http://www.fastshare.org/dlgo/AB_Verification_ACE.VR.zip

PR #5814, opened 2017-11-24
 Advanced Ballistics - Removed all unnecessary settings

ulteq:
Replaced `simulateForSnipers`, `simulateForGroupMembers`, `simulateForEveryone`, `disabledInFullAutoMode` and `simulationRadius` with some automagic.

Advanced Ballistics is now only running if:
a) The projectile is local
b) The projectile is a tracer (or has a visible bullet trace effect) and is reasonably close to the player

### DIFF
--- a/addons/advanced_ballistics/functions/fnc_handleFired.sqf
+++ b/addons/advanced_ballistics/functions/fnc_handleFired.sqf
@@ -21,22 +21,45 @@ TRACE_10("firedEH:",_unit, _weapon, _muzzle, _mode, _ammo, _magazine, _projectil
 
 if (!(_ammo isKindOf "BulletBase")) exitWith {};
 if (!alive _projectile) exitWith {};
-if (_unit distance ACE_player > GVAR(simulationRadius)) exitWith {};
 if (underwater _unit) exitWith {};
 
-private _abort = GVAR(disabledInFullAutoMode) && {getNumber(configFile >> "CfgWeapons" >> _weapon >> _mode >> "autoFire") == 1};
-if (!local _unit && {!_abort && !GVAR(simulateForEveryone)}) then {
-    // The shooter is non local -> abort unless we match an exception rule        
-    _abort = !GVAR(simulateForGroupMembers) || {group ACE_player != group _unit};
-    if (_abort && GVAR(simulateForSnipers)) then {
-        if (currentWeapon _unit == primaryWeapon _unit && count primaryWeaponItems _unit > 2) then {
-            private _opticsName = (primaryWeaponItems _unit) select 2;
+private _abort = !local _unit;
+if (_abort) then {
+    private _bulletVelocity = velocity _projectile;
+    private _muzzleVelocity = vectorMagnitude _bulletVelocity;
+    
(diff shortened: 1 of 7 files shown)


addons/advanced_ballistics/functions/fnc_handleFired.sqf
```diff
@@ -21,22 +21,45 @@ TRACE_10("firedEH:",_unit, _weapon, _muzzle, _mode, _ammo, _magazine, _projectil
...
+        private _maxRange = if (_airFriction < 0) then {
+            -ln(1 - _airFriction * _muzzleVelocity * _maxTime) / _airFriction
+        } else {
+            _muzzleVelocity * _maxTime
+        };
+        uiNamespace setVariable [format[QGVAR(maxRange_%1), _ammo], _maxRange];
+    };
+    if (ACE_player distance _unit > _maxRange && {ACE_player distance ((getPosASL _unit) vectorAdd ((vectorNormalized _bulletVelocity) vectorMultiply _maxRange)) > _maxRange}) exitWith {};
+    
+    private _ammoCount = _unit ammo _muzzle;
```
PabstMirror: I think we want to add +1 here as first shot is 29 and last is 0 (the ammo is already removed by the time firedEH runs)

addons/advanced_ballistics/functions/fnc_handleFired.sqf
```diff
@@ -21,22 +21,45 @@ TRACE_10("firedEH:",_unit, _weapon, _muzzle, _mode, _ammo, _magazine, _projectil
...
+    private _tracersEvery = getNumber(configFile >> "CfgMagazines" >> _magazine >> "tracersEvery");
+    private _lastRoundsTracer = getNumber(configFile >> "CfgMagazines" >> _magazine >> "lastRoundsTracer");
+    if (_ammoCount <= _lastRoundsTracer || {_tracersEvery > 0 && {(_ammoCount - _lastRoundsTracer) % _tracersEvery == 0}}) exitWith { _abort = false };
+    
+    if (GVAR(bulletTraceEnabled) && {_muzzleVelocity > BULLET_TRACE_MIN_VELOCITY} && {cameraView == "GUNNER"}) then {
+        if (currentWeapon ACE_player == binocular ACE_player) exitWith { _abort = false };
```
PabstMirror: I guess a civilian driving a tank could have `cameraView` gunner and `currentWeapon ACE_player == binocular ACE_player` (both "")
but I wouldn't worry about it

ulteq: We could check if the player is currently sitting in a vehicle. But then we would need extra code for the spotting scope.

An ideal solution would be to only display the bullet trace if the view is magnified at least 4 times.

addons/advanced_ballistics/functions/fnc_handleFired.sqf
ulteq: Right!

addons/advanced_ballistics/functions/fnc_handleFired.sqf
PabstMirror: We can always refine this later, I wouldn't bother changing it for now

PR #5815, opened 2017-11-24
Unify settings and keybinds categories prefix

Dystopian:
- replace `ACE3 ` and `ACE -` prefixes and suffixes with `ACE ` prefix, as discussed in Slack.

### DIFF
--- a/addons/common/stringtable.xml
+++ b/addons/common/stringtable.xml
@@ -999,94 +999,94 @@
             <Chinesesimp>不要强行</Chinesesimp>
         </Key>
         <Key ID="STR_ACE_Common_ACEKeybindCategoryEquipment">
-            <English>ACE3 Equipment</English>
-            <German>ACE-Ausrüstung</German>
-            <Polish>ACE3 Wyposażenie</Polish>
-            <Portuguese>Equipamentos ACE3</Portuguese>
-            <Russian>ACE3 Снаряжение</Russian>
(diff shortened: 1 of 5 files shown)


commy2:
I like this style very much.

addons/common/stringtable.xml
```diff
@@ -999,94 +999,94 @@
...
         </Key>
         <Key ID="STR_ACE_Common_ACEKeybindCategoryVehicles">
-            <English>ACE3 Vehicles</English>
-            <German>ACE3 Fahrzeuge</German>
-            <Polish>ACE3 Pojazdy</Polish>
-            <Portuguese>Veículos ACE3</Portuguese>
-            <Russian>ACE3 Транспорт</Russian>
-            <Czech>ACE3 Vozidla</Czech>
-            <Spanish>ACE3 Vehículos</Spanish>
-            <Italian>Veicoli ACE3</Italian>
```
SilentSpike: My only concern is whether this is no longer localised correctly?

jonpas: Think of it as: `<mod> <category>`.

PabstMirror: It's combining 2 nouns, I don't think the order could be that bad. 
("Equipment ACE3" sounds a little weird, but can be understood just fine)
Having the prefix first will be a big help for categories as they get sorted alphabetically and I think we want all of ace together.

PR #5821, opened 2017-11-25
Add persistent laserpointer

Dystopian:
- add automatically enabled laserpointer/flashlight after weapon switch or vehicle enter/exit.

It's very useful in coop modes (where laserpointer are used IMO) when you often have to switch to launcher or enter/exit vehicles. In vanilla laserpointer is disabled after these actions and you have to switch it on each time.

Option is added to `ACE Weapons` category (after #5815), which I plan to extend with some uncategorized settings.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -446,9 +446,4 @@ GVAR(deviceKeyCurrentIndex) = -1;
 {false},
 [0xC7, [true, false, false]], false] call CBA_fnc_addKeybind;  //SHIFT + Home Key
 
-
-// init persistent laser
-if (GVAR(persistentLaserEnabled)) then {true call FUNC(switchPersistentLaser)};
-
-
 GVAR(commonPostInited) = true;
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -446,4 +446,9 @@ GVAR(deviceKeyCurrentIndex) = -1;
 {false},
 [0xC7, [true, false, false]], false] call CBA_fnc_addKeybind;  //SHIFT + Home Key
 
+
+// init persistent laser
+if (GVAR(persistentLaserEnabled)) then {true call FUNC(switchPersistentLaser)};
+
+
 GVAR(commonPostInited) = true;
--- a/addons/common/functions/fnc_switchPersistentLaser.sqf
+++ b/addons/common/functions/fnc_switchPersistentLaser.sqf
@@ -18,6 +18,7 @@
 params ["_enabled"];
 
 if (!_enabled) exitWith {
+    if (isNil QGVAR(laserKeyDownEH)) exitWith {};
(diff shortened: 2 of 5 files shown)


addons/common/functions/fnc_switchPersistentLaser.sqf
```diff
@@ -0,0 +1,64 @@
...
+
+if (!_enabled) exitWith {
+    ["KeyDown", GVAR(laserKeyDownEH)] call CBA_fnc_removeDisplayHandler;
+    ["weapon", GVAR(laserWeaponEH)] call CBA_fnc_removePlayerEventHandler;
+    ["turret", GVAR(laserTurretEH)] call CBA_fnc_removePlayerEventHandler;
+    ["vehicle", GVAR(laserVehicleEH)] call CBA_fnc_removePlayerEventHandler;
+};
+
+GVAR(laserKeyDownEH) = ["KeyDown", {
+    if !((_this select 1) in actionKeys "headlights") exitWith {false};
```
commy2: Rebind `headlights` to something with modifier key, e.g. `Shift + L` and try this.

Dystopian: Here we have to choose: key with modifier or persistent laser. I think this implementation is most compromise. We can add much more complicated way to handle this, but I don't think it's worth it.

commy2: But `actionKeys` is really bad even apart from that. It reports 0 on very low FPS on keyDown, despite the key being pressed and the vanilla action executing. I just don't think this makes for reliable code.

addons/common/XEH_postInit.sqf
```diff
@@ -446,4 +446,9 @@ GVAR(deviceKeyCurrentIndex) = -1;
 {false},
 [0xC7, [true, false, false]], false] call CBA_fnc_addKeybind;  //SHIFT + Home Key
 
+
+// init persistent laser
+if (GVAR(persistentLaserEnabled)) then {true call FUNC(switchPersistentLaser)};
```
PabstMirror: I don't think we need this as the function should be run when cba settings are ready

addons/common/functions/fnc_switchPersistentLaser.sqf
```diff
@@ -0,0 +1,64 @@
...
+ * true call ace_common_fnc_switchPersistentLaser
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_enabled"];
+
+if (!_enabled) exitWith {
+    ["KeyDown", GVAR(laserKeyDownEH)] call CBA_fnc_removeDisplayHandler;
```
PabstMirror: these GVARs will all be nil if it's false on first run

addons/common/XEH_postInit.sqf
Dystopian: will check this

Dystopian: it's true, didn't know that, fixed both comments

commy2:
lgtm :+1:

PR #5773, opened 2017-11-15
Add sway factor option to advanced fatigue

jlillis:
**When merged this pull request will:**
- Add a sway factor option to advanced fatigue that allows the amount of weapon sway to be customized.

PR #5828, opened 2017-11-29
Add ability to set aiming coefficient for mission scripting.

xrufix:
- Add function **ace_common_fnc_setAimCoef**. This is a partial copy of the [setHearingCapability](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_setHearingCapability.sqf) function and allows for scripts to raise the aiming coefficient (weapon sway) above the current value. The highest value set will be applied. It is designed so that scripts can simulate stress, fear, suppression and other mental or physical conditions that will likely reduce a soldier's ability to aim correctly.
- Make ace_advanced_fatigue_fnc_handleEffects use the above-mentioned function instead of `setCustomAimCoef`.

I would however prefer a framework for similar functionality (having a max/min setting that can be set by different functions/scripts).

### DIFF
--- a/addons/common/functions/fnc_setAimCoef.sqf
+++ b/addons/common/functions/fnc_setAimCoef.sqf
@@ -1,0 +1,50 @@
+/*
+ * Author: xrufix, Glowbal
+ * Handle set AimCoef calls. Will use highest available setting.
+ *
+ * Arguments:
+ * 0: id <STRING>
+ * 1: settings <NUMBER>
+ * 2: add [true] OR remove [false] (default: true) <BOOL>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player,"ace_advanced_fatigue", 1, true] call ace_common_fnc_setAimCoef
+ *
+ * Public: Yes
+ */
+#include "script_component.hpp"
+
+params ["_unit","_id", "_setting", ["_add", true]];
+
+private _exists = false;
+private _highestCoef = 1;
+private _map = _unit getVariable [QGVAR(setAimCoefMap), []];
+
+_map = _map select {
+    _x params ["_xID", "_xSetting"];
+    if (_id == _xID) then {
+        _exists = true;
+        if (_add) then {
+            _x set [1, _setting];
+            _highestCoef = _highestCoef max _setting;
+            true
+        } else {
+            false
+        };
+    } else {
+        _highestCoef = _highestCoef max _xSetting;
+        true
+    };
+};
+
+if (!exists && _add) then {
+	_highestCoef = _highestCoef max _setting;
+	_map pushBack [_id, _setting];
+};
+
+// Update the value
+_unit setVariable [QGVAR(setAimCoefMap), _map];
+_unit setCustomAimCoef _highestCoef;
--- a/addons/common/functions/fnc_setAimCoef.sqf
+++ b/addons/common/functions/fnc_setAimCoef.sqf
@@ -18,7 +18,7 @@
  */
 #include "script_component.hpp"
 
-params ["_unit","_id", "_setting", ["_add", true]];
+params ["_unit", "_id", "_setting", ["_add", true]];
 
 private _exists = false;
 private _highestCoef = 1;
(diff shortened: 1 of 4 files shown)


addons/common/functions/fnc_setAimCoef.sqf
```diff
@@ -0,0 +1,51 @@
...
+ * None
+ *
+ * Example:
+ * [player,"ace_advanced_fatigue", 1, true] call ace_common_fnc_setAimCoef
+ *
+ * Public: Yes
+ */
+#include "script_component.hpp"
+
+params ["_unit","_id", "_setting", ["_add", true]];
```
Neviothr: Space after `["_unit",`.

xrufix: Done.

PabstMirror:
@jlillis @xrufix any thoughts on this PR and https://github.com/acemod/ACE3/pull/5773

They both do similar but different things to the same system.
Do we need both or would just adding the linear scaling factor in 5773 be enough?

I don't think it's a problem to add both (will be some minor merge conficts but should be easy to resolve)

jlillis:
I think it would be better to modify #5773 so that when swayFactor is 0, advanced fatigue does not update setCustomAimCoef anymore. This would allow mission makers to customize the aiming coefficient in their own scripts.

xrufix:
I think #5773 does something different. It will enable mission editors to change the impact of fatigue on weapon sway, not weapon sway itself (at least, not directly). It adds a factor which is multiplied with fatigues current setting. E.g. when you want to `setCustomAimCoef 6`, while standing still with no fatigue you would use a factor of 4. But when the player runs a bit this high factor will produce ridiculy high values.

My proposed solution does instead affect weapon sway directly. By checking for the highest set weapon sway, it produces more reliable and predictable results if a certain AimCoef is desired.

Merging and keeping both could be easily done and would imho be best.

addons/common/functions/fnc_setAimCoef.sqf
```diff
@@ -0,0 +1,51 @@
+/*
+ * Author: xrufix, Glowbal
+ * Handle set AimCoef calls. Will use highest available setting.
+ *
+ * Arguments:
+ * 0: unit <OBJECT>
```
jonpas: `Unit`

addons/common/functions/fnc_setAimCoef.sqf
```diff
@@ -0,0 +1,51 @@
...
+ * 1: id <STRING>
```
jonpas: `ID`

addons/common/functions/fnc_setAimCoef.sqf
```diff
@@ -0,0 +1,51 @@
...
+ * 2: settings <NUMBER>
```
jonpas: `Settings` and maybe add a note in `()` describing what exactly is meant.

addons/common/functions/fnc_setAimCoef.sqf
```diff
@@ -0,0 +1,51 @@
...
+ * 3: add [true] OR remove [false] (default: true) <BOOL>
```
jonpas: `Add (true) or remove (false) <BOOL> (default: true)`

xrufix:
@jonpas Requested changes addressed. I also changed the file where i copy-pasted from.

xrufix:
@commy2 This really needs #5773 to be merged (keeping my changes to fnc_handleEffects.sqf), otherwise GVAR(swayFactor) is undefined.

commy2:
Easy fix :)

Tee-ee-ee:
Sorry for digging the old post, but has this issue been solved? If so, what is currently the correct way to set the custom aim coefficient?
I pull the "Sway Factor" in ACE advanced fatigue all the way down to 0; then in the editor I put this in my player's init:
this setCustomAimCoef 0;
But I still get sway. The only valid way for the time being seems to completely disable the advanced fatigue to get no sway. Wondering if this is supposed to be the way to do it?

xrufix:
@Tee-ee-ee Please open a new issue to adress this.

PR #5834, opened 2017-12-01
Fix enum errors

commy2:
- title


Neviothr:
Might as well fix `createUnit` in these:
https://github.com/acemod/ACE3/blob/1f987c0f3dd0da7d8ecb281e6dcc6cd758778796/addons/spectator/functions/fnc_compat_spectatorBI.sqf
https://github.com/acemod/ACE3/blob/cbe06e5bc716332fc4023f2abcaccf5d2430aca5/addons/interaction/ACE_ZeusActions.hpp

PR #5836, opened 2017-12-01
add setting to hide freefall altimeter, ref #5596

commy2:
- Currently broken, because `cadetMode` is synonymous with `true` now
- solves part of #5596

### DIFF
--- a/addons/parachute/XEH_preInit.sqf
+++ b/addons/parachute/XEH_preInit.sqf
@@ -21,4 +21,12 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+[
+    QGVAR(hideAltimeter),
+    "CHECKBOX",
+    [LSTRING(HideAltimeter), LSTRING(HideAltimeter_tooltip)],
+    format ["ACE %1", localize ELSTRING(common,DisplayName)],
+    true
+] call cba_settings_fnc_init; 
+
 ADDON = true;
--- a/addons/parachute/XEH_preInit.sqf
+++ b/addons/parachute/XEH_preInit.sqf
@@ -26,7 +26,9 @@ PREP_RECOMPILE_END;
     "CHECKBOX",
     [LSTRING(HideAltimeter), LSTRING(HideAltimeter_tooltip)],
     format ["ACE %1", localize ELSTRING(common,DisplayName)],
-    true
+    true,
+    false,
+    {[QGVAR(hideAltimeter), _this, false] call EFUNC(common,cbaSettings_settingChanged)}
 ] call cba_settings_fnc_init; 
 
 ADDON = true;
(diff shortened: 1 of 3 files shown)


addons/parachute/XEH_preInit.sqf
```diff
@@ -21,4 +21,12 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+[
+    QGVAR(hideAltimeter),
+    "CHECKBOX",
+    [LSTRING(HideAltimeter), LSTRING(HideAltimeter_tooltip)],
+    format ["ACE %1", localize ELSTRING(common,DisplayName)],
+    true
```
PabstMirror: We might want to add
```
    true,
    {[QGVAR(hideAltimeter), _this, false] call EFUNC(common,cbaSettings_settingChanged)}
```
to make it global and then to add a note that a restart might be required for this setting change to take effect.

In testing the soldier UI didn't always reload after jumping, so once those UI are hidden they won't come back if setting is changed back to easy.

But this would be very rare issue.

commy2: Why make it a global setting needlessly? That would just be a restriction and this one works perfectly fine as individual setting.
Good point about the warning though.

addons/parachute/stringtable.xml
```diff
@@ -97,5 +97,13 @@
             <Chinesesimp>备用降落伞</Chinesesimp>
             <Chinese>備用降落傘</Chinese>
         </Key>
+        <Key ID="STR_ACE_Parachute_HideAltimeter">
+            <English>Hide Freefall Altimeter</English>
+            <German>Freifall-Höhenmesser verstecken</German>
```
bux: > Höhen- und Geschwindigkeitsmesser

Welcher wird da jetzt ausgeblendet?

commy2: Beide.

commy2: Wobei hier bei Geschwindigkeit allein die Fallgeschwindigkeit gemeint ist. Ist quasi Teil vom Altimeter.

bux: >Freifall-Höhenmesser

vs.

>Höhen- und Geschwindigkeitsmesser

commy2: Naja, da ist nur soviel Platz für Text. Wenn es passen sollte dann ists okay.

commy2: Es blendet ja auch nicht den normalen Geschwindigkeitsmesser aus. Nur den für den freien Fall. Der Tooltip spezifiziert es dann nochmal genauer.

Hamster2k:
The option is currently turned off, but missions and clients can override. Check your personal config and/or mission config!

PR #5839, opened 2017-12-02
Wind Deflection - Only track remote bullets if they are tracers

ulteq:
* Do not run the PFH for remote projectiles unless they are tracers

### DIFF
--- a/addons/winddeflection/functions/fnc_handleFired.sqf
+++ b/addons/winddeflection/functions/fnc_handleFired.sqf
@@ -18,9 +18,18 @@
 //IGNORE_PRIVATE_WARNING ["_unit", "_weapon", "_muzzle", "_mode", "_ammo", "_magazine", "_projectile", "_vehicle", "_gunner", "_turret"];
 TRACE_10("firedEH:",_unit, _weapon, _muzzle, _mode, _ammo, _magazine, _projectile, _vehicle, _gunner, _turret);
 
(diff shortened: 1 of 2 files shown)


addons/winddeflection/functions/fnc_handleFired.sqf
```diff
@@ -18,9 +18,18 @@
...
 
-if (missionNamespace getVariable [QEGVAR(advanced_ballistics,enabled), false] && {_projectile isKindOf "BulletBase"} && {_unit isKindOf "Man"}) exitWith {false};
+if (missionNamespace getVariable [QEGVAR(advanced_ballistics,enabled), false] && {_projectile isKindOf "BulletBase" && {_unit isKindOf "Man"}}) exitWith {false};
 
 if (!((_projectile isKindOf "BulletBase") || {_projectile isKindOf "GrenadeBase"})) exitWith {false};
 if (_unit distance ACE_player > GVAR(simulationRadius)) exitWith {false};
 
+private _abort = false;
+if (!local _unit && {_projectile isKindOf "BulletBase"}) then {
+    private _ammoCount = (_unit ammo _muzzle) + 1;
```
PabstMirror: Just adding a note, don't think it will be a problem
This function is called for vehicles as well and `ammo` doesn't seem to work for non-primary turret.
e.g. Right door gunner of ghost hawk

`(vehicle player) ammo "LMG_Minigun_Transport2"` is always 0

PR #5845, opened 2017-12-02
scale grenade torque depending on throw mode

### LINKED ISSUE #5680: Grenade torque should vary with throw type
**Arma 3 Version:** `1.76.143187` (stable)
**CBA Version:** `3.4.1` (stable)
**ACE3 Version:** `3.11.0` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- The new grenade torque is always the same regardless of the type of throw.

**Steps to reproduce:**
- Play as any loadout with a grenade, try the three different ACE throws, watch grenade spin.

I was talking to @jonpas about how the new grenade torque looks great for the stick grenades in the various WW2 mods and suggested to him that each throw type should have it's own rotation speed, he asked me to make an issue for it.

My suggestion would be:
Normal - quick rotation
High - slow to moderate rotation
Drop - very little rotation

commy2:
- close #5680
- dunno about the numbers. Could be tweaked.


Drofseh:
I would lower it further.
Something like
```
#define THROWSTYLE_HIGH_TORQUE_COEF .6
#define THROWSTYLE_DROP_TORQUE_COEF .2
```

PR #5848, opened 2017-12-03
Overpressure - Don't add backblast damage if blocked

PabstMirror:
@Neviothr 

Blocks backblast damage if not using ace medical AND damage is disabled via `allowDamage`


Neviothr:
Thanks!

PR #5855, opened 2017-12-05
Interaction - Fix setting zeus group behaviour to safe

PabstMirror:


PabstMirror:
>when in Zeus if you try to set a groups combat mode via ace interactions to "Safe" it sets it to "Careless"

PR #5860, opened 2017-12-05
add function to create zeus like compass control

commy2:
- title
- automatically animated
- can be added for any display


commy2:
Added darker background and indicator for the middle of the control, so the user knows where he's looking at exactly.
I moved the whole thing to config, which means all subcontrols can now be moved as one entity by moving only the controlgroup, and to also make it more neat.

The control can still be created by script:
```sqf
findDisplay 46 ctrlCreate ["ace_common_CompassControl", -1]
```

Or now also added by config:
```cpp
class ace_common_CompassControl;
class MyDisplay {
    class controls {
        compass: ace_common_CompassControl {};
    };
};
```

PR #5863, opened 2017-12-06
Generate isTracer array for magazines

PabstMirror:
Minor optimization for determining if a shot is a tracer
Makes a array of bools for which shots are tracers `[true, false, true...]`
Performance gains are minor (0.0184 ms -> 0.0153 ms)

- [ ] Add to AB
- [ ] Add to new NVG after merge

Not thrilled with creating big arrays for the 2000 rnd mags, but it's really only going to be a few kilobytes. Could just cap count at ~500.
Most of those end up being all tracers, which are represented by an empty array.

### DIFF
--- a/addons/common/XEH_PREP.hpp
+++ b/addons/common/XEH_PREP.hpp
@@ -52,6 +52,7 @@ PREP(fixCollision);
 PREP(fixFloating);
 PREP(fixLoweredRifleAnimation);
 PREP(fixPosition);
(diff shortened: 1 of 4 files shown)


ulteq:
I love the idea, but ```0.0184 ms -> 0.0153 ms``` is somewhat disappointing.

Did you test the performance with both small and big magazines?

PabstMirror:
Closing for now, probably not worth it

PR #5864, opened 2017-12-06
replace usage of setCaptive with camouflageCoef trait

commy2:
- title
- same effect, same locality (AL, EG)
- same ai behavior
- better mission compatibility (many use setCaptive, very unlikely they use setUnitTrait CamoCoef mid mission)
- no moving to the civ side which breaks map group marker and probably many other things


PabstMirror:
Tested a bit in mp
- Tried effect as non-global, `unitTrait` still synced globally as it says it should, but it took several seconds; so I think keeping it as a global is a good idea just like setCaptive.

- Don't really like getting the enum errors, but probably most people don't run with show script errors. For ace_movement I added `INFO("Note: getUnitTrait / loadCoef enum error can be ignored [present in ARMA 1.78 RC]");` which we might want here to prevent false issue reports.

- Tried `[player, "setHidden", "testing", true] call ace_common_fnc_statusEffect_set` and while walking around an AI he threw a grenade at me (really surprised me). So they can still hear you and may continue to engage.

With setCaptive, it's like you turn into a friendly civ, so they probably actively try not to shoot at you or splash you with frag.
With this it's just that they can't see you and just like hiding in smoke they may keep firing or launch GPs.

I actually think that might be an improvement and be more realistic.

commy2:
The hearing you part could maybe be fixed by also setting `audibleCoef` in the same way. I don't know if that makes them stop lobbing grenades though.

SilentSpike:
For the cause of unconscious I don't think hearing you is an issue since you can't move anyway

carlosleung:
This made me still get shoot by the enemy,When I lost consciousness in front of the him

bux:
@carlosleung  Please don't revive old issues and pull requests.
Searching the current or resolved issues would have shown you that this is already known and even already fixed.

PR #5866, opened 2017-12-07
Remove dash from NV category name

SilentSpike:
- Title

### DIFF
--- a/addons/nightvision/stringtable.xml
+++ b/addons/nightvision/stringtable.xml
@@ -1,8 +1,8 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project name="ACE">
     <Package name="NightVision">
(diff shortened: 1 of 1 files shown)


commy2:
:banana:

addons/nightvision/stringtable.xml
```diff
@@ -1,8 +1,8 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project name="ACE">
     <Package name="NightVision">
         <Key ID="STR_ACE_NightVision_Category">
-            <English>ACE - Nightvision</English>
+            <English>ACE Nightvision</English>
```
Dystopian: would be more consistent not to have prefix here, like in other modules. It would added automatically.

PabstMirror: I think if we eventually convert to calling `cba_setting_init` directly, we'll want string-tables that have both parts combined

PR #5873, opened 2017-12-08
Add endRadioTransmissions Event

dedmen:
Add's ACE_endRadioTransmissions Event that can be hooked by other Radio mods.

The TFAR code causes Problems in combination with spectator. TFAR 1.0 has a new function that handles that but I don't want to keep adding a TFAR 1.0 check into here and modify this everytime something changes. So I just added one Event for all Radio mods.

Also calling `TFAR_fnc_onDDTangentReleased` will throw a deprecation warning with 1.0. So the old code should be removed once the new version is released and properly gained traction.

### DIFF
--- a/addons/common/functions/fnc_endRadioTransmission.sqf
+++ b/addons/common/functions/fnc_endRadioTransmission.sqf
@@ -1,7 +1,7 @@
 /*
  * Author: commy2
  *
- * End radio transmissions of addons TFAR and ACRE2. TFAR v0.9.7, ACRE Public Beta 2.0.3.571
+ * End radio transmissions of addons TFAR and ACRE2. TFAR v0.9.x, ACRE Public Beta 2.0.3.571
  *
  * Arguments:
  * None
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_endRadioTransmission.sqf
```diff
@@ -16,6 +16,8 @@
  */
 #include "script_component.hpp"
 
+[QGVAR(endRadioTransmissions)] call CBA_fnc_localEvent;
```
jonpas: Public events should be `"ace_X"`.

PabstMirror:
could add something like this if the calls would be a problem
```
private _tfrVersionAr = getArray (configFile >> "CfgPatches" >> "task_force_radio" >> "versionAr");
if ([[1], _tfrVersionAr] call cba_versioning_fnc_version_compare) exitWith {};
```

dedmen:
Can detect 1.0 with CfgPatches class `tfar_core` It's just a warning message in RPT.

PR #5876, opened 2017-12-09
Attempt to make enum errors more clear

PabstMirror:
image

Adds a big `IGNORE THIS` to start of on screen error

I had to change var name to make it fit

Also bump required cba/arma ver

### DIFF
--- a/addons/main/script_mod.hpp
+++ b/addons/main/script_mod.hpp
@@ -9,7 +9,7 @@
 #define VERSION_AR MAJOR,MINOR,PATCHLVL,BUILD
 
 // MINIMAL required version for the Mod. Components can specify others..
-#define REQUIRED_VERSION 1.80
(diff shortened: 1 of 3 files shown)


addons/main/script_mod.hpp
```diff
@@ -9,8 +9,8 @@
 #define VERSION_AR MAJOR,MINOR,PATCHLVL,BUILD
 
 // MINIMAL required version for the Mod. Components can specify others..
-#define REQUIRED_VERSION 1.76
-#define REQUIRED_CBA_VERSION {3,4,1}
+#define REQUIRED_VERSION 1.80
```
jonpas: Doing 1.78 here is fine I think, maybe Linux port reaches 1.78 and not 1.80 for whatever reason. :P

PR #5877, opened 2017-12-09
make ammoCount ui element changeable despite ace_reload

commy2:
I think making this not changeable when a component is present is unnecessarily confusing.

### DIFF
--- a/addons/reload/ACE_UI.hpp
+++ b/addons/reload/ACE_UI.hpp
@@ -1,7 +1,0 @@
-class ACE_UI {
-    class ammoCount {
-        class conditions {
-            ADDON = "cameraOn == (getConnectedUAV ACE_player)";
-        };
-    };
-};
--- a/addons/reload/config.cpp
+++ b/addons/reload/config.cpp
@@ -19,4 +19,3 @@ class CfgPatches {
 #include "CfgEventHandlers.hpp"
 #include "CfgActions.hpp"
 #include "ACE_Settings.hpp"
-#include "ACE_UI.hpp"


PabstMirror:
We need this to show ammo count while controlling UAVs

jonpas:
I wrote something like this on Slack, but then forgot to paste it here. The point of Selective UI is it being selective yes, but ACE3's goal is still realism, and by accepting this PR we would allow anyone to just show the Reload UI, which goes against why that component even exists:
> Hides the default reload indicators, making it necessary to manually check your magazine.

This still allows people to remove the PBO and get ability to set it themselves. Additionally using that config it could (and probably should) be made into a setting (enable/disable component) instead, so you wouldn't even have to remove the PBO, and it will all nicely hook with Selective UI as well.

I believe @commy2 agreed after I explained that, and is required for reason @PabstMirror stated above, so closing this.

Brojustuserandomgenerationusername:
If this is forced why even have the option to enable or disable it in options? And why force it if its outright removing an UI element that was in the base game?

jonpas:
Because you can remove the reload PBO.

Brojustuserandomgenerationusername:
Wouldn't that make incompatibility with people that have the reload PBO on servers? And why not just disable it through the ingame options?

jonpas:
What incompatibility? You should run the same PBOs on server and all clients.

I am not sure what you mean with your last question.

Brojustuserandomgenerationusername:
What i mean is that if having no reload PBO on a server people with the reload PBO would be able to join. Since i cant force an UI setting on the players.
The question was about why even have the option to enable/disable the ammo counter if it doesn't work and i have to delete a PBO to make it appear.

jonpas:
Enable CheckPBOs. I already explained why, ACE is modular, you can do whatever you like with it, we only provide sensible defaults.

Now let's stop resurrecting old tickets and move questions like this to Slack.

Brojustuserandomgenerationusername:
I realize that its modular, i just dont understand why enabling and disabling PBOs are required just because a setting in the options for the UI is bugged.

jonpas:
It is not bugged.

Brojustuserandomgenerationusername:
Then why does changing the obvious setting in ACE UI to show me the ammo left in the mag doesn't work?

commy2:
This should've just been merged imo.

jonpas:
Because you have the reload PBO loaded, which works on top of UI, like any mod can. This has been explained in this topic a few times now, end of discussion.

Brojustuserandomgenerationusername:
And there is no plan to fix the obvious issue that with the reload PBO loaded the settings wont display the ammo counter yet the option wont be removed because its obviously a bug?

jonpas:
It's not a bug, it is by design. Remove the reload PBO if you don't want it. With standard ACE package you can't change it, with custom variations (without reload PBO) you can - by design.

PR #5879, opened 2017-12-09
Add category for laserpointer and reload settings

Dystopian:
- move uncategorized `Laser Pointer` and `Check ammo on weapon reload` settings to `ACE Weapons` category. `Display text on grenade throw` will be moved there after #5581 (when?).

Should these uncategorized settings be moved to `ACE Vehicles`?
- `Cook-off probability coefficient`
- `Gforces Effects`
- `Overpressure Distance Coefficient`


PabstMirror:
Not sure about those as they are global settings. I feel like it's more important to get client settings into categories

PR #5882, opened 2017-12-10
Transfer old client settings to CBA

PabstMirror:
Transfer old ace setting to the cba setting profile var
Should make the conversion smoother for normal users.

Example output:
```
[ACE] (common) INFO: -Transfering old ACE_Settings to CBA-
[ACE] (common) INFO: Transfering setting [ace_interact_menu_alwaysUseCursorSelfInteraction: true] returned true
[ACE] (common) INFO: Transfering setting [ace_interact_menu_menuBackground: 2] returned true
``


PabstMirror:
@commy2 - Do you see any problems with this?
As far as I could tell `CBA_settings_fnc_set` should be smart enough and not let me break things

commy2:
Get and set shouldn't let you break anything.
I knew `ACE_VersionNumberString` would be useful for something one day.

PR #5889, opened 2017-12-10
Fix some undefined funcs / strings

PabstMirror:


bux:
interesting... good catch

PR #5897, opened 2017-12-12
Adjusted dig surface whitelist for jungle maps

chris579:
- Allow digging trenches on some jungle maps (e.g. Prei Khmaoch Luong)

Probably this whitelist and blacklist should be included in CfgSurfaces to allow 3rd party mods to whitelist their surfaces on purpose.


jonpas:
I like `CfgSurfaces` idea! 👍

chris579:
Will do a separate pr for this idea.

PR #5898, opened 2017-12-12
Added config params for white/blacklisting digging

chris579:
- Add config switches to allow 3rd party mods to white/blacklist single surfaces for digging.

Current black/whitelist macros could be preserved otherwise a lot of config entries have to be made. Plus some maps could be misconfigured.

### DIFF
--- a/addons/common/functions/fnc_canDig.sqf
+++ b/addons/common/functions/fnc_canDig.sqf
@@ -24,8 +24,11 @@ if ((getPosATL _unit) select 2 > 0.05 || // Walking on objects, such as building
 ) exitWith {false};
 
 private _surfaceClass = (surfaceType _posASL) select [1];
-private _surfaceType = getText (configFile >> "CfgSurfaces" >> _surfaceClass >> "soundEnviron");
-private _surfaceDust = getNumber (configFile >> "CfgSurfaces" >> _surfaceClass >> "dust");
+private _config = configFile >> "CfgSurfaces" >> _surfaceClass;
+private _surfaceType = getText (_config >> "soundEnviron");
+private _surfaceDust = getNumber (_config >> "dust");
+private _isWhitelisted = getNumber (_config >> "ACE_digWhitelisted") == 1;
+private _isBlacklisted = getNumber (_config >> "ACE_digBlacklisted") == 1;
 TRACE_2("Surface",_surfaceType,_surfaceDust);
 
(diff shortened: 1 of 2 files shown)


commy2:
ACE_canDig, with !isNumber being the indeterminate state?
Two entries makes it overdefined.

addons/common/functions/fnc_canDig.sqf
```diff
@@ -24,8 +24,10 @@ if ((getPosATL _unit) select 2 > 0.05 || // Walking on objects, such as building
...
-private _surfaceType = getText (configFile >> "CfgSurfaces" >> _surfaceClass >> "soundEnviron");
-private _surfaceDust = getNumber (configFile >> "CfgSurfaces" >> _surfaceClass >> "dust");
+private _config = configFile >> "CfgSurfaces" >> _surfaceClass;
+private _surfaceType = getText (_config >> "soundEnviron");
+private _surfaceDust = getNumber (_config >> "dust");
+private _canDig = getNumber (_config >> "ACE_canDig") == 1;
 TRACE_2("Surface",_surfaceType,_surfaceDust);
 
-!(_surfaceType in DIG_SURFACE_BLACKLIST) && {(_surfaceDust >= 0.1) || {_surfaceType in DIG_SURFACE_WHITELIST}}
+_canDig || !{(_surfaceType in DIG_SURFACE_BLACKLIST) && {(_surfaceDust >= 0.1) || {_surfaceType in DIG_SURFACE_WHITELIST}}}
```
commy2: This is what I meant:
```sqf
if (isNumber (_config >> "ACE_canDig")) then {
    getNumber (_config >> "ACE_canDig") // return
} else {
    !(_surfaceType in DIG_SURFACE_BLACKLIST) && {(_surfaceDust >= 0.1) || {_surfaceType in DIG_SURFACE_WHITELIST}} // return
};
```
Which makes the explicit setting of the entry overwrite whatever the script would estimate in case the entry is missing.

commy2: This is the same logic that the description.ext and the mission.sqm operate under btw.

chris579:
Changed it

commy2:
Hope that made sense to you.

chris579:
Yeah got a thinking barrier and wanted to stuff everything into one large condition

PabstMirror:
Looks good,
we could block inheritance with `configProperties [ , , false]` thing but from what I can see most thing just inherit from `Default`

PR #5900, opened 2017-12-12
Make attach/detach hint look uniform

mharis001:
- Add space at the bottom of hint to make it look nicer
- Before: 
attach_before
- After:
attach_after


jonpas:
What if the item name is longer or shorter?

mharis001:
The `STR_ACE_Attach_Item_Attached` string always places the attached/detached text on the second line and the item name on the first.
The items that are able to be attached/detached is when they have `ACE_Attachable != ""` defined.
Tested it with all items and even the longest names have some padding on the sides.

bux:
IMO, It would be better to add a general inner padding to the box instead of adding a blank line as a spacer.

jonpas:
@bux I am not sure how that works in Arma UI, but it is no CSS. 😆

PR #5906, opened 2017-12-14
Trench improvements

chris579:
**WIP**  
Migration of [GRAD Trenches](https://github.com/gruppe-adler/grad_trenches) into ACE.  

Things that need to be addressed:
* [ ] Trenches that are too huge stuck out of the ground when placed
* [ ] Define Settings for camouflage
* [ ] Remove intercept on finish
* [x] Edit names displayed in 3DEN and Zeus


addons/trenches/functions/fnc_initTrench.sqf
```diff
@@ -0,0 +1,31 @@
...
+params [
+    ["_object", objnull, [objNull]]
+];
+
+if (is3DEN) exitWith {
+    [_object] call FUNC(initTrench3DEN);
+};
+
+if (local _object) then {
+    // Has to be spawned to ensure MP compatibility
```
commy2: Can vouch for this. Same reason setVariable public fails from init. initPost is no option without creating even more event handlers, because 3den compat.

addons/trenches/functions/fnc_initTrench3DEN.sqf
```diff
@@ -0,0 +1,50 @@
...
+ * [TrenchObj] call ace_trenches_fnc_initTrench3DEN
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+params [
+    ["_object", objnull, [objNull]]
+];
+
+_initTrench = {
```
commy2: private

addons/trenches/functions/fnc_initTrench3DEN.sqf
```diff
@@ -0,0 +1,50 @@
...
+    params["_object"];
```
commy2: space

addons/trenches/functions/fnc_initTrench3DEN.sqf
```diff
@@ -0,0 +1,50 @@
...
+    _object addEventHandler ["Dragged3DEN", {
+        params["_object"];
```
commy2: same

addons/trenches/functions/fnc_initTrench3DEN.sqf
```diff
@@ -0,0 +1,50 @@
...
+
+    if (((_object get3DENAttribute QUOTE(ACE_trenches_camouflageTrench)) select 0) isEqualTo 1) then {
+        [_object] call FUNC(placeCamouflage);
+    };
+};
+
+// If no object is given apply this to all trenches in 3den
+if (isNull _object) then {
+    {
+        if(getNumber(configFile >> "CfgVehicles" >> typeOf _x >> QGVAR(isTrench)) == 1) then {
```
commy2: space

addons/trenches/functions/fnc_placeCamouflage.sqf
```diff
@@ -0,0 +1,61 @@
...
+
+private _fnc_onFinish = {
+    (_this select 0) params ["_unit", "_trench"];
+
+    private _camouflageObjects = getArray (configFile >> "CfgWorldsTextures" >> worldName >> "camouflageObjects");
+    private _placedObjects = [];
+
+    {
+        private _object = createSimpleObject [selectRandom _camouflageObjects, [0,0,0]];
+        _object attachTo [_trench, getArray(_x)];
```
commy2: space

addons/trenches/functions/fnc_placeTrench.sqf
```diff
@@ -87,6 +88,13 @@ GVAR(digPFH) = [{
     _trench setVectorDirAndUp [_v1, _v3];
     GVAR(trenchPos) = _basePos;
 
+    if(surfaceType (position _trench) != GVAR(currentSurface)) then {
```
commy2: space

addons/trenches/functions/fnc_removeCamouflage.sqf
```diff
@@ -0,0 +1,37 @@
...
+
+private _fnc_onFailure = {
+    (_this select 0) params ["_unit"];
+    // Reset animation
+    [_unit, "", 1] call ace_common_fnc_doAnimation;
+};
+
+[CAMOUFLAGE_DURATION, [_unit, _trench], _fnc_onFinish, _fnc_onFailure, localize LSTRING(removeCamouflageProgress)] call ace_common_fnc_progressBar;
+
+[_unit, "AinvPknlMstpSnonWnonDnon_medic4"] call ace_common_fnc_doAnimation;
```
commy2: EFUNC

addons/trenches/functions/fnc_getSurfaceTexturePath.sqf
```diff
@@ -0,0 +1,60 @@
...
+    private _fileNameArr = _fileWildcard splitString "";
+    if(_fileNameArr find "*" > -1) then {
+        _fileNameArr deleteAt (_fileNameArr find "*");
+    };
+
+    format["%1%2%3", _basePath, (_fileNameArr joinString ""), _filePrefix];
+};
+
+private _basePath = getText (configFile >> "CfgWorldsTextures" >> "Altis" >> "surfaceTextureBasePath");
+if((_surfaceType find "#Gdt" == -1) || {worldName == "Tanoa"}) then {
```
commy2: space

addons/trenches/functions/fnc_getSurfaceTexturePath.sqf
```diff
@@ -0,0 +1,60 @@
...
+    // check for overridden surface paths
+    private _overrideCfg = configFile >> "CfgWorldsTextures" >> worldName >> "Surfaces" >> _parsedSurfaceType >> "texturePath";
+    if(isText (_overrideCfg)) exitWith {
+        getText(_overrideCfg)
+    };
+    // get config file wildcard
+    private _fileWildcard = getText(configfile >> "CfgSurfaces" >> _parsedSurfaceType >> "files");
+    // remove * in file wildcard
+    private _fileNameArr = _fileWildcard splitString "";
+    if(_fileNameArr find "*" > -1) then {
```
commy2: space

addons/trenches/functions/fnc_getSurfaceTexturePath.sqf
```diff
@@ -0,0 +1,60 @@
...
+
+private _getTexturePath = {
+    params["_surfaceType", "_basePath", "_filePrefix"];
+
+    // remove leading #
+    private _parsedSurfaceType = _surfaceType select [1, count _surfaceType];
+    // check for overridden surface paths
+    private _overrideCfg = configFile >> "CfgWorldsTextures" >> worldName >> "Surfaces" >> _parsedSurfaceType >> "texturePath";
+    if(isText (_overrideCfg)) exitWith {
+        getText(_overrideCfg)
```
commy2: spaces

addons/trenches/functions/fnc_getSurfaceTexturePath.sqf
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+params [
+    ["_object", objnull, [objNull]]
+];
+if(isNull _object) exitWith {};
+
+if(!isText (configFile >> "CfgWorldsTextures" >> worldName >> "surfaceTextureBasePath")) exitWith {DEFAULT_TEXTURE};
```
commy2: spaces

chris579:
I'll look over all spaces, you don't have to comment all 1337 evidences 😛

addons/trenches/CfgVehicles.hpp
```diff
@@ -59,41 +41,121 @@ class CfgVehicles {
...
+    class ACE_envelope_big: ACE_envelope_small {
         author = ECSTRING(common,ACETeam);
         displayName = CSTRING(EnvelopeBigName);
         descriptionShort = CSTRING(EnevlopeBigDescription);
         model = QPATHTOEF(apl,ace_envelope_big4.p3d);
-        scope = 2;
         GVAR(diggingDuration) = 25;
         GVAR(removalDuration) = 15;
-        GVAR(noGeoClass) = "ACE_envelope_big_NoGeo";
+        GVAR(noGeoClass) = QUOTE(ACE_envelope_big_NoGeo);
```
commy2: Why?

addons/trenches/CfgVehicles.hpp
```diff
@@ -59,41 +41,121 @@ class CfgVehicles {
         scope = 2;
         GVAR(diggingDuration) = 20;
         GVAR(removalDuration) = 12;
-        GVAR(noGeoClass) = "ACE_envelope_small_NoGeo";
+        GVAR(noGeoClass) = QUOTE(ACE_envelope_small_NoGeo);
```
commy2: Why?

addons/trenches/CfgVehicles.hpp
chris579: I just like macros

commy2:
Don't want to miss any of them^^

chris579:
I think you will find a lot of missing spaces in my code 😄

addons/trenches/functions/fnc_placeCamouflage.sqf
chris579: what do u mean exactly?

chris579:
Most spaces should be fixed now. Also found some missing ones in old code, fixed them too.

addons/trenches/functions/fnc_placeCamouflage.sqf
jonpas: getArray _x

Drofseh:
Normally when a trench is first dug it's the colour of whatever the soil is, would it be possible to have a medium to dark dirt trench dug first and then have the terrain matching done as a second "camouflage the position" action on the trench itself?

SilentSpike:
Cool, I like it 👍 I also kind of like @Drofseh's suggestion, but I'm not sure it adds much to the experience of trench digging

addons/trenches/CfgWorldsTextures.hpp
```diff
@@ -0,0 +1,280 @@
+class CfgWorldsTextures {
```
SilentSpike: Usually I'd say class structures added by ace components should use the prefix, however I'm not sure here since I can see that retaining this one would be backwards compatible with the original addon 🤔

jonpas: I don't think any map actually adds compatibility for this in the map mod, so it wouldn't be a problem to change. Additionally backwards compatibility can be simpler by inheriting the class from the new class.

commy2: Tag everything custom in a global namespace.

addons/trenches/stringtable.xml
```diff
@@ -1,6 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project name="ACE">
     <Package name="trenches">
+      <Key ID="STR_ACE_Trenches_ActionCategory">
+          <English>Trenches</English>
+          <German>Gräben</German>
+      </Key>
```
commy2: 6 spaces, should be 8.

addons/trenches/script_component.hpp
```diff
@@ -2,9 +2,9 @@
 #define COMPONENT_BEAUTIFIED Trenches
 #include "\z\ace\addons\main\script_mod.hpp"
 
-// #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
-// #define ENABLE_PERFORMANCE_COUNTERS
+#define DEBUG_MODE_FULL
+#define DISABLE_COMPILE_CACHE
+#define ENABLE_PERFORMANCE_COUNTERS
```
commy2: Reminder to disable debug mode.

addons/trenches/CfgWorldsTextures.hpp
```diff
@@ -0,0 +1,280 @@
...
+            class travajih_detail_nohq {
+                texturePath = "ca\sara\Data\travajih_detail_nohq.paa";
+            };
+            class travajih_mco {
+                texturePath = "ca\sara\Data\travajih_mco.paa";
+            };
+        };
+    };
+
+  //Shapur
```
commy2: +2 spaces

addons/trenches/functions/fnc_placeConfirm.sqf
```diff
@@ -36,8 +36,10 @@ _unit setVariable [QGVAR(isPlacing), false, true];
 params ["_unit"];
 if (isNull GVAR(trench)) exitWith {};
 
+private _trenchTexture = (getObjectTextures ace_trenches_trench) select 0;
 deleteVehicle GVAR(trench);
 private _trench = createVehicle [GVAR(trenchClass), [0, 0, 0], [], 0, "NONE"];
+_trench setObjectTextureGlobal [0,_trenchTexture];
```
commy2: space

addons/trenches/functions/fnc_placeConfirm.sqf
```diff
@@ -36,8 +36,10 @@ _unit setVariable [QGVAR(isPlacing), false, true];
 params ["_unit"];
 if (isNull GVAR(trench)) exitWith {};
 
+private _trenchTexture = (getObjectTextures ace_trenches_trench) select 0;
```
commy2: GVAR(trench)

addons/trenches/functions/fnc_removeCamouflage.sqf
```diff
@@ -0,0 +1,37 @@
...
+
+private _fnc_onFailure = {
+    (_this select 0) params ["_unit"];
+    // Reset animation
+    [_unit, "", 1] call EFUNC(common,doAnimation);
+};
+
+[CAMOUFLAGE_DURATION, [_unit, _trench], _fnc_onFinish, _fnc_onFailure, localize LSTRING(removeCamouflageProgress)] call EFUNC(common,progressBar);
+
+[_unit, "AinvPknlMstpSnonWnonDnon_medic4"] call ace_common_fnc_doAnimation;
```
commy2: EFUNC

addons/trenches/CfgWorldsTextures.hpp
chris579: Maybe this data should be included into `CfgWorlds` as it's referring to single maps.What about that?

chris579:
Currently waiting for @AdlerSalbei to finish models for envelope small and big. Beside that I have to figure out how trench placement works when digging start. May have to contact Grey for this.

Drofseh:
@SilentSpike yah, I'm not sure my suggestion is necessary or even practical, but I do get a little annoyed with the current trenches I dig coming out of the ground looking like they are made of straw.

I think this PR is great and I'm really happy that we'll have terrain matched trenches, I just wish it went a little bit further.

On the topic of the current implementation: Does this support Eden or Zeus placed trenches?

chris579:
> On the topic of the current implementation: Does this support Eden or Zeus placed trenches?

Yes, this supports both. However Zeus has never been tested but as the texture is also set on object init it should work.

Drofseh:
It'd just like it to start with a generic earth texture, which then gets changed to the map appropriate texture if the player chooses to camouflage it with an interaction after the initial dig.

I dig my trench, the texture is dirt, I want to hide it, I interact with the trench and spend some time camouflaging it, at which point the texture gets changed the the appropriate choices for the map.

This is both more realistic, and I think makes for better TvT gameplay, making the players take more time to get a better position.

SilentSpike:
If we were going to go that route, remember to account for sand terrains too! (Where I guess it would be camouflaged immediately).

However, that's also something that could be added in a later PR once this implementation is complete.

chris579:
Intercept and Intercept_CBA have been added. Will be moved to optionals when development is finished. 
Also new trench models for big and small have been added. Will rewrite trench placement to fix placement for bigger trench models.

commy2:
intercept_cba and intercept_core should be their own PR, no?

chris579:
Could remove it if you want to but including it here would allow to directly implement it into our changes.

dedmen:
Intercept core itself probably shouldn't be in ACE. It is a optional dependency and a seperate mod.
Also the Intercept core stuff in here is already outdated and the new stuff is needed for the latest intercept_cba stuff.
Also Intercept cba is also a seperate mod.. But the PBO content will not change besides adding a certificate some time in the future.
Also considering the intercept-cba dll will get relatively frequent updates compared to other extensions it could considerably increase the repo size. I can get the intercept-cba build server up and let make.py download the latest build automagically when you build ACE.
Maybe you have a better idea.

jonpas:
What's the point of having Intercept CBA in ACE3 repo in the first place? It's optional to use, it should have own releases in its own repo, there is no reason for it to be here or us including it in release packages.

We don't even include CBA in release package and that's a hard dependency. Who cares  will get it, who doesn't care we don't care about. A simple post with release that informs people that is possibility is enough.

addons/trenches/XEH_postInit.sqf
```diff
@@ -23,4 +23,14 @@ GVAR(digDirection) = 0;
 // handle waking up dragged unit and falling unconscious while dragging
 ["ace_unconscious", {_this call FUNC(handleUnconscious)}] call CBA_fnc_addEventHandler;
 
+// Add zeus evh to handle texture on trench move
+(getAssignedCuratorLogic player) addEventHandler ["CuratorObjectEdited", {
```
dedmen: Guessing this won't work properly at all if Zeus logic get's assigned after postInit.

chris579: Yeah, that’s right but I don’t see a way to detect when a player has become curator.

PabstMirror: https://community.bistudio.com/wiki/Arma_3:_Event_Handlers#CuratorObjectEdited
The event is local/local, so you could probably just add them to all curators

chris579: Is it possible to do this EH with XEH? Then it would work all the time however this EH might has to be attached to the curator unit.

chris579: XEH would at least allow support when somebody gets Zeus while the mission is running already.

chris579: Ok there is no XEH support for curator events.

PabstMirror:
```
["ModuleCurator_F", "init", {
    params ["_logic"];
    _logic addEventHandler ["CuratorObjectEdited", {
    ...
    }];
```

chris579: I added an InitPost xeh for curator modules that adds this evh. Have no possibility for testing currently but I assume that this could work.

addons/trenches/functions/fnc_isTrench.sqf
```diff
@@ -0,0 +1,22 @@
...
+ * Arguments:
+ * 0: Object <OBJECT>
+ *
+ * Return Value:
+ * Is trench <BOOL>
+ *
+ * Example:
+ * [myObj] call ace_trenches_fnc_isTrench
+ *
+ * Public: No
```
SilentSpike: Since it's not a public function anyway, you could just make this a macro

chris579: Or this could be public. Dunno if there is any useful case for such a public function.

chris579:
Most things (except for CBA Settings) are done now, however trenches will stuck out of the ground as positioning isn't done with consideration of the bounding box. I already tried to rewrite the whole positioning (as it is calculated multiple times and so on) but [Arma is fooling me](https://www.youtube.com/watch?v=NnbY7OUlZlU). Would appreciate some help with this as I don't understand the current calculation.

addons/trenches/functions/script_component.hpp
```diff
@@ -1 +1 @@
-#include "\z\ace\addons\trenches\script_component.hpp"
\ No newline at end of file
+#include "..\script_component.hpp"
```
jonpas: Keep that as it was, we said we won't do that back when it was introduced.

chris579:
Added CBA settings that should cover almost every edge case. Allows to:
* disable single trench types (default: all enabled)
* disable trench digging completely (default: disabled)
* disable camouflaging completely (default: disabled)
* disable the need for an entrenching tool for digging trenches (default: disabled)
* enable the need for an entrenching tool for camouflaging (default: disabled)
* specify the dig time for every single trench type
* specify the removal time for every single trench type

AdlerSalbei:
This PR can be closed.

PabstMirror:
Continued in #6745

PR #5909, opened 2017-12-14
Update how weight is displayed in ACE arsenal

alganthe:
- Weight is now indicated as follows: "kilos (lbs)".

new weight

### DIFF
--- a/addons/arsenal/functions/fnc_onSelChangedLeft.sqf
+++ b/addons/arsenal/functions/fnc_onSelChangedLeft.sqf
@@ -410,4 +410,4 @@ switch (GVAR(currentLeftPanel)) do {
     };
 };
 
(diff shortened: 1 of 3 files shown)


PabstMirror:
Why not just show a single weight that matches the user setting for inventory display?

alganthe:
Because that way you can tell your squadmates using the other system how much you weight without having to go on a google search.

Cuel:
If merged the change should be reflected in inventory aswell imo

Dystopian:
I know how make it suitable for everyone: make the option not just `kg` and `lb`, but also `Show both (kg first)` and `Show both (lb first)`

alganthe:
The order literally doesn't matter., for inventory this is out of the scope of this PR

commy2:
Done in #5918.

jonpas:
The PRs work hand-in-hand right @commy2 ? As in merging them both it should already work.

commy2:
Yes.

PR #5910, opened 2017-12-14
Add Ctrl+F search shortcut for ACE Arsenal

alganthe:
- Add Ctrl+F, Ctrl+A, and Ctrl+X support for ACE Arsenal

Pressing it while no search field are in focus: select left field
Pressing it while a search field is in focus: select the other field.

### DIFF
--- a/addons/arsenal/functions/fnc_onKeyDown.sqf
+++ b/addons/arsenal/functions/fnc_onKeyDown.sqf
@@ -46,6 +46,10 @@ if (!GVAR(leftSearchbarFocus) && {!GVAR(rightSearchbarFocus)}) then {
         case (_keyPressed == DIK_V && {_ctrlState}): {
             [_display] call FUNC(buttonImport);
         };
+        // Search fields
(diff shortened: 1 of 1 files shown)


commy2:
This is very nice!

addons/arsenal/functions/fnc_onKeyDown.sqf
```diff
@@ -99,6 +103,14 @@ if (!GVAR(leftSearchbarFocus) && {!GVAR(rightSearchbarFocus)}) then {
         case (_keyPressed == DIK_V && {_ctrlState}): {
             _return = false;
         };
```
PabstMirror: Maybe ignore ctrl+x as well?
```
        case (_keyPressed == DIK_X && {_ctrlState}): {
            _return = false;
        };
```

alganthe: You mean allow cut and paste?

alganthe:
Added Ctrl+A and Ctrl+X support as requested.

PR #5911, opened 2017-12-14
Revert #5887 because it breaks JIP

alganthe:
- This reverts commit 5fa0179d063800fd59ddce382a19cd533200e839.

In my testing I did not check if JIP still worked, and in hindsight I should've realized it wouldn't, here's why I did the change and why I need to revert it.

Implementation 1 (before global fix):
- Local changes are ignored except for the machine running the add / remove func with the global flag.
- JIP compatibility

- Local changes are respected even when running add / remove with the global flag
- No JIP compatibility


PabstMirror:
Tag #5887

PR #5912, opened 2017-12-14
Add Kh25ML Laser Guided AGMs

xrufix:
* CSAT was sad that NATO had it's new Maverick L, so they decided they needed a semi-active laser homing air-to-ground missile themselves. Seems that Russian KH-25ML fits the job.

### DIFF
--- a/addons/maverick/CfgAmmo.hpp
+++ b/addons/maverick/CfgAmmo.hpp
@@ -6,7 +6,7 @@ class CfgAmmo {
     class Missile_AGM_02_F : MissileBase {};
     
     class GVAR(L) : Missile_AGM_02_F {
-        author = "rufix";
+        author = "xrufix";
         autoSeekTarget = 0;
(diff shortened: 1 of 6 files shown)


xrufix:
I added that to ace_maverick. I'm not happy with that, because, obviously, its not a maverick.

addons/maverick/CfgAmmo.hpp
```diff
@@ -6,7 +6,7 @@ class CfgAmmo {
     class Missile_AGM_02_F : MissileBase {};
     
     class GVAR(L) : Missile_AGM_02_F {
-        author = "rufix";
+        author = "xrufix";
```
bux: so close

xrufix: Question: Should i remove my name and replace it with `ECSTRING(common,ACETeam);`?

PR #5913, opened 2017-12-14
Add brightness control when NV effects disabled

Dystopian:
- get back brightness control from 3.11 when NV effects disabled with `ace_nightvision_effectScaling`==0.

### DIFF
--- a/addons/nightvision/XEH_postInit.sqf
+++ b/addons/nightvision/XEH_postInit.sqf
@@ -26,11 +26,15 @@ GVAR(ppEffectCCMuzzleFlash) = -1;
 ["ace_settingsInitialized", {
     TRACE_3("settingsInitialized",GVAR(disableNVGsWithSights),GVAR(fogScaling),GVAR(effectScaling));
 
-    // Disable ALL effects if ace_nightvision_effectScaling is zero
-    if (GVAR(effectScaling) == 0) exitWith {};
+    ["visionMode", LINKFUNC(onVisionModeChanged), false] call CBA_fnc_addPlayerEventHandler;
+
+    // handle only brightness if effects are disabled
+    if (GVAR(effectScaling) == 0) exitWith {
+        GVAR(ppEffectNVGBrightness) = ppEffectCreate ["ColorCorrections", 1236];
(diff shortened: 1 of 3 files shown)


addons/nightvision/XEH_postInit.sqf
```diff
@@ -26,11 +26,15 @@ GVAR(ppEffectCCMuzzleFlash) = -1;
...
 
-    // Disable ALL effects if ace_nightvision_effectScaling is zero
-    if (GVAR(effectScaling) == 0) exitWith {};
+    ["visionMode", LINKFUNC(onVisionModeChanged), false] call CBA_fnc_addPlayerEventHandler;
+
+    // handle only brightness if effects are disabled
+    if (GVAR(effectScaling) == 0) exitWith {
+        GVAR(ppEffectNVGBrightness) = ppEffectCreate ["ColorCorrections", 1236];
+        GVAR(ppEffectNVGBrightness) ppEffectForceInNVG true;
+    };
```
Dystopian: should be adjusted and committed here, will fix ASAP

PabstMirror:
Not a huge fan of this, because I liked being able to say that if `effectScaling = 0` then we don't touch nvg at all (besides config changes).

This will adjust nvgs to max brightness (higher than normal) at mission start.
Just use the old scaling code
```
_brightness = ((round (10 * _brightness + _changeInBrightness) / 10) min 0.5) max -0.5;
```
so that it can start at 0 (no change).

Dystopian:
What's the point to change ranges for enabled/disabled effects?

PabstMirror:
Old range was -5 to 5, new range is -6 to 0
you re-map that so that -3 is base level `(_brightness+3)/5 + 1`
But this means that the starting value of 0, is brighter than normal.

I just want it so that the default brightness, with no key-presses, is 1

Dystopian:
Yes, I wanted to preserve new range and get back old style allowing make picture brighter and dimmer than default. So new `-3` became old `0`, `-6` -> `-5` and `0` -> `5`. If you want I'll make `-3` as default value. Or you insist on range changing depending on setting?

PabstMirror:
maybe change the default in the `getVariable` to -3 if using old method, that should work

PR #5917, opened 2017-12-15
Scopes - Fix 'boreHeight' and 'baseAngle' MP synchronization

ulteq:
* Unifies the 'getBaseAngle' & 'getBoreHeight' API
* Fixes the issue that the `boreHeight` and `baseAngle` information is always used from the local player

We should probably cache the calculation or use publicVariables instead at some point.


bux:
> We should probably cache the calculation or use publicVariables instead at some point.

could you create a task?

ulteq:
I think if we need a task for it, we should do it right away, if not we should just forget it.

ulteq:
:+1:

PR #5920, opened 2017-12-15
Add slider settings

PabstMirror:
Adds proper min/max for slider settings
Without it would use `[-1, 5000, 0, 1]` because it needed to handle long distances, but that's pretty bad for settings in the 0-1 range.

These values were a little tricky to come up with, some of them are a little subjective.
I tried to give plenty of room for the min/max as they are hard limits.
e.g. medical_bleedingCoefficient has max of 25, which I think should be ok

There also are some special cases where -1 means something else, I tried to check everything, but this could use a thorough review.

### DIFF
--- a/addons/advanced_fatigue/ACE_Settings.hpp
+++ b/addons/advanced_fatigue/ACE_Settings.hpp
@@ -20,7 +20,7 @@ class ACE_Settings {
         description = CSTRING(PerformanceFactor_Description);
         typeName = "SCALAR";
         value = 1;
-        sliderSettings[] = {0, 10, 1, 1};
+        sliderSettings[] = {0, 5, 1, 1};
     };
     class GVAR(recoveryFactor) {
         category = CSTRING(DisplayName);
@@ -28,7 +28,7 @@ class ACE_Settings {
         description = CSTRING(RecoveryFactor_Description);
         typeName = "SCALAR";
         value = 1;
-        sliderSettings[] = {0, 10, 1, 1};
+        sliderSettings[] = {0, 5, 1, 1};
     };
     class GVAR(loadFactor) {
         category = CSTRING(DisplayName);
@@ -36,7 +36,7 @@ class ACE_Settings {
         description = CSTRING(LoadFactor_Description);
         typeName = "SCALAR";
         value = 1;
-        sliderSettings[] = {0, 10, 1, 1};
+        sliderSettings[] = {0, 5, 1, 1};
     };
     class GVAR(terrainGradientFactor) {
         category = CSTRING(DisplayName);
@@ -44,6 +44,6 @@ class ACE_Settings {
         description = CSTRING(TerrainGradientFactor_Description);
         typeName = "SCALAR";
         value = 1;
-        sliderSettings[] = {0, 10, 1, 1};
+        sliderSettings[] = {0, 5, 1, 1};
     };
 };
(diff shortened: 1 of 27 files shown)


addons/advanced_fatigue/ACE_Settings.hpp
```diff
@@ -20,26 +20,30 @@ class ACE_Settings {
         description = CSTRING(PerformanceFactor_Description);
         typeName = "SCALAR";
         value = 1;
+        sliderSettings[] = {0, 10, 1, 1};
```
ulteq: Should be `sliderSettings[] = {0, 3, 0.1, 1};` which translates to `GVAR(VO2Max) >= 35 && GVAR(VO2Max) <= 95`.

addons/advanced_fatigue/ACE_Settings.hpp
```diff
@@ -20,26 +20,30 @@ class ACE_Settings {
...
     };
     class GVAR(loadFactor) {
         category = CSTRING(DisplayName);
         displayName = CSTRING(LoadFactor);
         description = CSTRING(LoadFactor_Description);
         typeName = "SCALAR";
         value = 1;
+        sliderSettings[] = {0, 10, 1, 1};
```
ulteq: I think we should go for a lower limit. Maybe 5 or even 3.

addons/advanced_fatigue/ACE_Settings.hpp
```diff
@@ -20,26 +20,30 @@ class ACE_Settings {
...
     };
     class GVAR(terrainGradientFactor) {
         category = CSTRING(DisplayName);
         displayName = CSTRING(TerrainGradientFactor);
         description = CSTRING(TerrainGradientFactor_Description);
         typeName = "SCALAR";
         value = 1;
+        sliderSettings[] = {0, 10, 1, 1};
```
ulteq: I would chose `sliderSettings[] = {0, 5, 1, 1};`.

addons/scopes/ACE_Settings.hpp
```diff
@@ -47,20 +48,23 @@ class ACE_Settings {
         value = 15;
         displayName = CSTRING(zeroReferenceTemperature_displayName);
         description = CSTRING(zeroReferenceTemperature_description);
+        sliderSettings[] = {-55, 55, 15, 1};
```
ulteq: We could use `sliderSettings[] = {-55, 55, 15, 0};` here.

addons/advanced_fatigue/ACE_Settings.hpp
PabstMirror: I have it at 5 max, just to keep it consistent with the other settings in the module.

ulteq: Would have to ask @BaerMitUmlaut what he thinks about it.

BaerMitUmlaut:
AF should be fine with maximum values of 2.

ulteq:
If you want the same value for each slider in advanced fatigue then I'd go for a maximum of 3.

But I don't understand why it is so important to have the same limits for all sliders.

PabstMirror:
My worry is the limits are hard

being more narrow gives a little more precision on the slider (kinda nice to have)
but if they are too narrow it prevents someone from setting it to what they want (really bad)

PR #5922, opened 2017-12-16
overhaul movable markers

### LINKED ISSUE #5921: Make movable markers work on briefing map
This #5397 don't work.

### LINKED ISSUE #5980: after move a marker marking on map is toggled on
**Arma 3 Version:** stable
**CBA Version:** stable
**ACE3 Version:** stable

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
after move a marker with ALT+mouseclick... marking on map is toggled on you need to mouseclick to turn off again

**Steps to reproduce:**
- Place marker
- move the marker with ALT+mouse
- the yellow map marking is active after moving
- need to click on map to disable

commy2:
- close #5921
- [x] cut back on the global use, objectively better variable / function names
- [x] fix ended event not resetting correct alpha etc.
- [x] cba_settings_fnc_init
- [x] work in briefing
- close #5980
- [x] map gestures interferes


chris579:
There should be an extra options for enabling/disabling moving in briefing.

commy2:
Why?

chris579:
In big missions this could end up in a clusterfuck.

commy2:
Then only allow it for group leaders? The setting can be changed mid mission too.

chris579:
Ok, you are right.

PR #5926, opened 2017-12-18
Fix FRIES being applied when attribute unchecked

SilentSpike:
Attributes that are specific to objects are saved even when the default value is set, so the helicopters using this one would always be equipped with FRIES if the checkbox had ever been toggled.


commy2:
These attributes are wonky.

PR #5932, opened 2017-12-19
Add setting to replace BIS Arsenal

Dystopian:
- add setting to replace BIS Arsenal with ACE Arsenal ~~both~~ in missions ~~and main game menu~~;
- add `condition` parameter to `openBox` function (like in vanilla);
- allow open arsenal with just player gear.

This PR allows to use new ACE Arsenal in existing missions with BIS Arsenal enabled without mission editing.

~~Still needs fix from #5887 (`false` parameter) reverted in #5911. @alganthe~~

Not all features are wrapped. E.g. `getVirtual*Cargo` are not handled at all (does anybody use it?).


addons/arsenal/config.cpp
```diff
@@ -6,7 +6,7 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"ace_common"};
+        requiredAddons[] = {"ace_common", "A3_Functions_F_Bootcamp"};
```
Cuel: not needed, already in ace_main

addons/arsenal/functions/fnc_wrapper_arsenal.sqf
```diff
@@ -0,0 +1,43 @@
...
+        _params params [["_full", false, [false]], ["_box", ACE_player, [objNull]], ["_center", ACE_player, [objNull]]];
+        [_box, _center, _full] call FUNC(openBox);
+    };
+    case "AmmoboxInit": {
+        _params params [["_box", objNull, [objNull]], ["_full", false, [false]], ["_condition", {true}, [{}]]];
+        private _interactionCondition = if (_condition isEqualTo {true}) then {
+            {true}
+        } else {
+            _box setVariable [QGVAR(wrapperCondition), _condition];
+            {_player call (_target getVariable QGVAR(wrapperCondition))}
```
Dystopian: probably most odd code in PR. It works though.

PabstMirror:
Will the setting be ready at time the functions are called?
We might need to use `ace_common_fnc_runAfterSettingsInit`

Dystopian:
looks like setting can be not ready, will fix it

alganthe:
I'm worried that people might think ACE Arsenal (AA) is a "mod" of BI Virtual Arsenal (VA) instead of a standalone, especially with the function wrappers, this could lead to misunderstandings.

I'm split on this.

jonpas:
@Dystopian We might not want to merge this due to confusion it would cause, and would rather go forward with ACE Arsenal overwriting Eden button and Zeus module and solve the loadout issue with port loadouts functions. That is we consider ACE Arsenal to be an improvement of Vanilla Arsenal and basically deprecates Vanilla Arsenal.

This is not final yet, but to let you know, if we will indeed not merge it, we thank you for your contribution! :)

Dystopian:
It sounds very strange for me. Most missions my group plays have `addAction`-to-box style Arsenal which you can't replace without function substitution. This PR adds **optional** disabled by default way mainly for server admins to use ACE Arsenal instead of BIS Arsenal without mission edit messing.

If you don't like that this option also replaces arsenal in main game menu, I can add exception for that.

I wish I understood what `confusion` and `misunderstandings` you are talking about. User has option. User can use this option. Everything is clear, isn't it?

Dystopian:
no more conflict with vanilla arsenal mission in main menu

addons/arsenal/script_component.hpp
```diff
@@ -14,4 +14,21 @@
...
+#ifdef DISABLE_COMPILE_CACHE
+    #define RECOMPILE
+#endif
+
 #include "\z\ace\addons\main\script_macros.hpp"
+
+
+// if use just GVAR here, "Undefined variable" error occurs
+#define WAIT_AND_CHECK_SETTING(func) \
+    if (missionName isEqualTo 'Arsenal') exitWith {_this call DOUBLES(BIS_fnc,func)BIS}; \
```
dedmen: call BIS_fnc_arsenalBIS ? The BIS at the end looks weird.

addons/arsenal/functions/fnc_wrapper_addVirtualItemCargo.sqf
```diff
@@ -0,0 +1,34 @@
...
+ * Public: No
+*/
+#include "script_component.hpp"
+
+WAIT_AND_CHECK_SETTING(addVirtualItemCargo);
+
+params [["_box", ACE_player, [objNull]], ["_items", [], ["", [], true]], ["_isGlobal", false, [false]], ["_initBox", true, [true]]];
+
+if (_items isEqualType "") then {_items = [_items]};
+if (_items isEqualType [] && {0 < count _items} && {"%All" == _items select 0}) then {
```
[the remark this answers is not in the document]
Dystopian: not much more weird than `"%All" == _items select 0` :P If you insist I'll fix sure

addons/arsenal/script_component.hpp
```diff
@@ -14,4 +14,21 @@
...
 #endif
 
+#ifdef DISABLE_COMPILE_CACHE
+    #define RECOMPILE
+#endif
+
 #include "\z\ace\addons\main\script_macros.hpp"
+
+
+// if use just GVAR here, "Undefined variable" error occurs
```
dedmen: Where?

addons/arsenal/script_component.hpp
Dystopian: what would you suggest to use as prefix/suffix? Because `BIS_fnc_arsenal` is already wrapper.

addons/arsenal/functions/fnc_wrapper_addVirtualItemCargo.sqf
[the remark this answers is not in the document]
dedmen: I prefer statements to be like you would say them in english.
If more than 0 items. And not if 0 is less than items we have. :U

addons/arsenal/script_component.hpp
Dystopian: instead of `missionNamespace getVariable QGVAR()`. Looks like there is changed namespace in BIS code

dedmen: Ah yeah. BIS code runs in uiNamespace for whatever reason.

addons/arsenal/script_component.hpp
dedmen: Ah understand now. Btw the ACE Arsenal mission is also called `Arsenal` But I guess you don't call BIS_fnc_arsenal in there

PabstMirror:
@alganthe - Is this needed after https://github.com/acemod/ACE3/pull/6111 ?
That added separate zeus modules

Dystopian:
@PabstMirror these are different things

Dystopian:
Implemented as standalone addon in [v7 ace additions](https://steamcommunity.com/sharedfiles/filedetails/?id=2025187921).

PR #5943, opened 2017-12-21
Add a function to port BI VA loadouts to ACE Arsenal

alganthe:
- Adds a function that can be called to port over BI VA loadouts to ACE Arsenal.

### DIFF
--- a/addons/arsenal/functions/fnc_portVALoadouts.sqf
+++ b/addons/arsenal/functions/fnc_portVALoadouts.sqf
@@ -1,9 +1,9 @@
 /*
  * Author: alganthe
- * Port VA loadouts to ACE Arsenal
+ * Port VA loadouts to ACE Arsenal.
  *
  * Arguments:
- * 0: Write over similarly named loadouts in ACE Arsenal <BOOL>
+ * 0: Write over similarly named loadouts in ACE Arsenal <BOOL> (default: false)
  *
  * Return Value:
  * None
--- a/addons/arsenal/functions/fnc_portVALoadouts.sqf
+++ b/addons/arsenal/functions/fnc_portVALoadouts.sqf
@@ -3,7 +3,7 @@
  * Port VA loadouts to ACE Arsenal.
  *
  * Arguments:
- * 0: Write over similarly named loadouts in ACE Arsenal <BOOL> (default: false)
+ * None
  *
  * Return Value:
  * None
@@ -12,7 +12,6 @@
 */
 #include "script_component.hpp"
 
-params [["_override", false, [false]]];
 private _VALoadouts = +(profilenamespace getvariable ["bis_fnc_saveInventory_data",[]]);
 private _aceLoadouts = +(profileNamespace getVariable [QGVAR(saved_loadouts),[]]);
 
@@ -28,9 +27,7 @@ for "_i" from 0 to (count _VALoadouts - 1) step 2 do {
     private _loadout = getUnitLoadout player;
 
     if (count _sameNameLoadoutsList > 0) then {
-        if (_override) then {
-            _aceLoadouts set [_aceLoadouts find (_sameNameLoadoutsList select 0), [_name, _loadout]];
-        };
+        _aceLoadouts set [_aceLoadouts find (_sameNameLoadoutsList select 0), [_name, _loadout]];
 
     } else {
         _aceLoadouts pushBack [_name, _loadout];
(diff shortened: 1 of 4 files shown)


addons/arsenal/functions/fnc_portVALoadouts.sqf
```diff
@@ -0,0 +1,40 @@
+/*
+ * Author: alganthe
+ * Port VA loadouts to ACE Arsenal
```
jonpas: Missing final dot.

addons/arsenal/functions/fnc_portVALoadouts.sqf
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Arguments:
+ * 0: Write over similarly named loadouts in ACE Arsenal <BOOL>
```
jonpas: `<BOOL> (default: false)`

jonpas:
Maybe add a button somewhere?

alganthe:
it need to load the loadout before saving it, that could be easily abused.

bux:
Force it to only work in SP

alganthe:
now that I think about it I could add a button in 3DEN instead of changing the ACE Arsenal interface.

alganthe:
If I add such a button, should I place it under tools or make an ACE category for the main bar?

alganthe:
I removed the override argument (it always does it now) and added a button under tools in 3DEN:
https://imgur.com/a/wWsfn

alganthe:
added an error message for both:
- No VA loadouts to import
- No player unit

should be good to go.

addons/arsenal/stringtable.xml
```diff
@@ -433,5 +433,17 @@
             <Italian>Impossibile aprire l'arsenale ACE</Italian>
             <Japanese>ACE 武器庫を開けません</Japanese>
         </Key>
+        <Key ID= "STR_ACE_Arsenal_portLoadoutsText">
+            <English>import BI VA loadouts to ACE Arsenal</English>
```
jonpas: `Import` (capitalization).

I think something like `Import Vanilla Arsenal loadouts to ACE Arsenal` would be more understandable, but longer.

addons/arsenal/stringtable.xml
```diff
@@ -433,5 +433,17 @@
...
+            <French>Importe les loadouts de BI VA dans ACE Arsenal</French>
+        </Key>
+        <Key ID= "STR_ACE_Arsenal_portLoadoutsPlayerError">
+            <English>No player unit available!</English>
```
jonpas: I would add instruction on what to do to make it work, such as `No player unit available! Place a unit and mark it as "Player"` or something.

addons/arsenal/stringtable.xml
```diff
@@ -433,5 +433,17 @@
...
+            <French>Aucune unité joueur disponible !</French>
+        </Key>
+        <Key ID= "STR_ACE_Arsenal_portLoadoutsLoadoutError">
+            <English>No loadouts to import</English>
```
jonpas: No `!` or `.` here?

addons/arsenal/stringtable.xml
alganthe: The issue is that it translates poorly into other languages, on the other hand "BI VA" is pretty much universal.

addons/arsenal/stringtable.xml
alganthe: huh, surprisingly you can fit a lot more characters than I thought in those messages

addons/arsenal/stringtable.xml
```diff
@@ -434,16 +434,16 @@
...
         <Key ID= "STR_ACE_Arsenal_portLoadoutsText">
-            <English>import BI VA loadouts to ACE Arsenal</English>
+            <English>Import BI VA loadouts to ACE Arsenal</English>
             <French>Importe les loadouts de BI VA dans ACE Arsenal</French>
         </Key>
         <Key ID= "STR_ACE_Arsenal_portLoadoutsPlayerError">
-            <English>No player unit available!</English>
-            <French>Aucune unité joueur disponible !</French>
+            <English>No player unit available! Place a unit and mark it as "Player".</English>
+            <French>Aucune unité joueur disponible ! Placez une unité et marquez la en tant que "joueur"</French>
```
jonpas: Missed final dot? ^^

alganthe: dammit !

Issue #5410, opened 2017-08-07
Incendiary grenades are unreliable on ammo boxes


CoxLedere:
**Arma 3 Version:** `1.72` (stable)
**CBA Version:** `3.4.0` (stable)
**ACE3 Version:** `3.10.2` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
 AN-M14 Incendiary grenades are unreliable when it comes to destroying ammo boxes. Some of the larger ammo boxes such as NATO Supply Box (B_supplyCrate_F) or an FIA Ammo box with camo netting (Box_FIA family) will often times not be destroyed by incendiary grenades even when mere inches away from the box. The only way I've been able to increase the reliability of their use is to use advanced throwing to drop the grenade on top of the ammo box, but even this isn't a surefire way, as sometimes the grenade will clip through the model in odd ways, or will roll off of what appears to be an otherwise flat surface. (the NATO Supply Box is a perfect example of this, it's collision box appears to have an odd cone at the top of it instead of being flat)

With the Supply Box you can throw an incendiary grenade on one side, run to the other side and take damage from the incendiary grenade burning you, but the ammo box can come out untouched. You can roll a grenade underneath the camo netting of the FIA Boxes and the box will be untouched as the grenade burns.

**Steps to reproduce:**
1. Place down an Ammo Box, preferably a NATO Supply Box or one of the FIA variants with camo netting.
2. Throw an incendiary grenade at the box.

**Where did the issue occur?**
- Editor (Singleplayer)

**Placed Modules:**
- None

**RPT log file:**
No errors in the RPT, but I've included it for completeness.
[rpt.txt](https://github.com/acemod/ACE3/files/1206282/rpt.txt)

JoramD0:
This was improved with the above referenced PR.

PR #5971, opened 2017-12-24
Tweaked effective area of incendiary grenades

JoramD0:
- This pull will tweak the effect size of the AN-M14 incendiary grenade. 

EDIT:
It will only change the range it area it damages for weaponholders, ammoboxes and static weapons (Not for vehicles).

This is to improve effectiveness on bigger supplyboxes.
ref: #5410


commy2:
Isn't that macro just for the visual effect?

JoramD0:
Now that I look at it, yea (how did I miss that...).

commy2:
Renamed the radius to radius, because it's not an area.

PR #5984, opened 2017-12-28
Change ace_map settings to use CBA init directly

mharis001:
- Convert all ace_map settings to CBA settings directly
- Give BFT settings their own category
- Make default channel setting a list


jonpas:
I think this is perfect to change the rest of the settings in this component.

@PabstMirror should we remove reading the setting from module (call to the function that applies it)?

mharis001:
@jonpas I can convert the settings in map component if that is decided

jonpas:
@mharis001 thank you! Let's wait to hear from others first.

mharis001:
Should BFT be its own component, or maybe a more generalized unit tracking? I dont know in terms of realism but i'd assume some OPFOR factions would have BFT like technology.

commy2:
I think the BFT settings should be in their own Category.

mharis001:
Remove question mark at the end of displayName of `BFT_ShowPlayerNames` and `BFT_HideAiGroups`?

commy2:
YES! Please do that! Also remove the `:` if present. Some other settings use that and it also doesn't belong.
Also worth looking into how to capitalize the English.

mharis001:
I think sentence case would be better than title case for settings.

PabstMirror:
Take a look at https://github.com/acemod/ACE3/pull/5581/files#diff-e1cf3ccb88bb6158f0e98e8cbd4f9483R4

We'll want to leave the ace_settings in, but empty except for `movedToSQF = 1;`

(Because we need to know at preInit if a mission config setting is a change, or a brand new setting for the mission)

mharis001:
Thats my bad. I'll fix it.

jonpas:
> Do you want to also place all settings in a `initSettings.sqf` file?

Yes, please.

PabstMirror:
We could keep BFT stuff in same category and just put them in a sub category
image

CBA 3.6 isn't out yet, so we can always do this in a later PR
`[format ["ACE %1", localize LSTRING(Module_DisplayName)], localize LSTRING(BFT_Module_DisplayName)],`

mharis001:
Sub category sounds good once CBA 3.6 is released.

Issue #5979, opened 2017-12-26
sqf_validator doesn't support multi-line statements


Dystopian:
**ACE3 Version:** `3.12.0` (dev + 4cfa64d)

**Description:**
SQF preprocessor supports multi-line statements with `\` at the line end, including inside strings and excluding comments. E.g. this code
```sqf
a=1 \
 + 2;
b="q\
s";
c="z
x"; \
// c1
// c2 \
// c3
// c4 \
a
```
is preprocessed to
```
a=1  + 2;

b="qs";

c="z
x"; 




a
```
It looks like `tools\sqf_validator.py` doesn't support such multi-line statements and produces false errors in some cases.

**Steps to reproduce:**
- create `test.sqf` in `addons\ai` folder with this correct sqf content:
```sqf
[{hint "1"} \
] call CBA_fnc_waitAndExecute;
```
- run `sqf_validator.py -m ai`;
- see false error `ERROR: Possible missing semi-column ';' detected at addons/ai\test.sqf Line number: 1`

**Where did the issue occur?**
- tools

SilentSpike:
That python script will become obsolete with the introduction of #5468

PR #5985, opened 2017-12-28
Use engine command to select weighted spall size

SilentSpike:
- Update spalling function to use `selectRandomWeighted`

This slightly changes distribution of size selected, previously the random selection wasn't uniform anyway because it was using the `round` command. The new distribution will select "small" more often than "medium" whereas previously it was the other way around, but I figure it makes sense to decrease probability with size in the context of spalling.

### DIFF
--- a/addons/frag/functions/fnc_doSpall.sqf
+++ b/addons/frag/functions/fnc_doSpall.sqf
@@ -15,12 +15,7 @@
  */
 #include "script_component.hpp"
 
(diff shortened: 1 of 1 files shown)


commy2:
Travis doesn't like multiline macros in SQF. Could be moved to .hpp

SilentSpike:
Just made it single-line instead

bux:
https://github.com/acemod/ACE3/issues/5979

PR #5987, opened 2017-12-29
make setHidden use forgetTarget

commy2:
- title
Question is if this breaks anything else? Maybe just apply it to enemy groups?

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -54,7 +54,9 @@
             _object setVariable [QGVAR(oldVisibility), _vis];
             _object setUnitTrait ["camouflageCoef", 0];
             {
-                _x forgetTarget _object;
+                if (side _x != side group _object) then {
+                    _x forgetTarget _object;
+                };
             } forEach allGroups;
         };
     } else {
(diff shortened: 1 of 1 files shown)


PabstMirror:
What is locality on forgetTarget?

commy2:
I just assumed it's AG EL, since it's the anti-`reveal`. This effect is a global one.

SilentSpike:
I have a feeling it may be global since it was added in A3

PabstMirror:
Should we `revel` to previous knowsAbout when they wake up?

commy2:
They'd be revealed over time. I think it's better not to.

addons/common/XEH_postInit.sqf
```diff
@@ -54,7 +54,9 @@
             _object setVariable [QGVAR(oldVisibility), _vis];
             _object setUnitTrait ["camouflageCoef", 0];
             {
-                _x forgetTarget _object;
+                if (side _x != side group _object) then {
```
Dystopian: this will not work if `side _object == ENEMY`

SilentSpike: We actually just discussed this internally, it's quite edge case. One fix would just be to do:
```
if (side _x != side _object)
```

SilentSpike: The other would be to just have friendlies forget the unit too (with the exception of their own group for healing).

PR #5995, opened 2017-12-30
Add parentheses to GETVAR macros

Dystopian:
- add parentheses to `GETVAR` and `ARR_SELECT` macros;
- remove superfluous parentheses in some scripts.

### DIFF
--- a/addons/main/script_macros.hpp
+++ b/addons/main/script_macros.hpp
@@ -20,11 +20,11 @@
 #define SETPVAR_SYS(var1,var2) setVariable [ARR_3(QUOTE(var1),var2,true)]
 
 #undef GETVAR
-#define GETVAR(var1,var2,var3) var1 GETVAR_SYS(var2,var3)
-#define GETMVAR(var1,var2) missionNamespace GETVAR_SYS(var1,var2)
-#define GETUVAR(var1,var2) uiNamespace GETVAR_SYS(var1,var2)
-#define GETPRVAR(var1,var2) profileNamespace GETVAR_SYS(var1,var2)
-#define GETPAVAR(var1,var2) parsingNamespace GETVAR_SYS(var1,var2)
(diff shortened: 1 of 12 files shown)


addons/main/script_macros.hpp
```diff
@@ -37,7 +37,7 @@
 #define GETGVAR(var1,var2) GETMVAR(GVAR(var1),var2)
 #define GETEGVAR(var1,var2,var3) GETMVAR(EGVAR(var1,var2),var3)
 
-#define ARR_SELECT(ARRAY,INDEX,DEFAULT) if (count ARRAY > INDEX) then {ARRAY select INDEX} else {DEFAULT}
+#define ARR_SELECT(ARRAY,INDEX,DEFAULT) (if (count ARRAY > INDEX) then {ARRAY select INDEX} else {DEFAULT})
```
commy2:
```sqf
(ARRAY param [INDEX, DEFAULT])
```

commy2: ^ Just an idea.

654wak654: Could also remove `ARR_SELECT` which only has 1 use:
https://github.com/acemod/ACE3/search?utf8=%E2%9C%93&q=ARR_SELECT&type=

Dystopian: what about BWC?

654wak654: ACE's macros aren't supposed to be used by others, specially not one this archaic.

jonpas: Actually they are: https://github.com/acemod/arma-project-template/blob/master/addons/main/script_macros.hpp

But this one can safely be removed as it plays no role anymore.

commy2: I mean, it can also stay as is. Doesn't have to be removed with this PR. I'm for merging this PR.

Dystopian: OK, will not update it

PR #5996, opened 2017-12-30
Various ACE Arsenal fixes (see description)

### LINKED ISSUE #5998: ACE Arsenal switches camera to edited unit on exit when fnc_openBox is used on not player unit
**Arma 3 Version:** `1.80` (stable)
**CBA Version:** `3.5` (stable)
**ACE3 Version:** `3.12.0.29` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
Vanilla Arsenal can be opened on AI unit (actually on other players too) using next syntax:
`["Open",[true,objNull, AI_Unit_1]] call bis_fnc_arsenal;`
And it is extremly useful in Zeus games.

But if I open ACE Arsenal on AI using:
`[player, AI_Unit_1, true] call ace_arsenal_fnc_openBox` or `[AI_Unit_1, AI_Unit_1, true] call ace_arsenal_fnc_openBox`
Then after Close button is pressed - camera sticks to edited AI unit.

It will be awesome, to make ACE Arsenal (which is much faster than vanilla) usable on AI.

**Steps to reproduce:**
- Start mission with AI unit name AI_Unit_1
- Execute `[AI_Unit_1, AI_Unit_1, true] call ace_arsenal_fnc_openBox`
- Click 'Close' button

- Editor (Singleplayer) / Editor (Multiplayer)

alganthe:
- Fix the cached array not being copied properly in openBox when using true as third argument.
- Fix add / remove virtualItems erroring out when passing false as second argument.
- Disable and fade out faces, voices, and insignias tabs while in 3DEN 
- close #5998 _~commy2_
- Add ability to exit arsenal and not apply a loadout in 3DEN

disabled preview


addons/arsenal/functions/fnc_addVirtualItems.sqf
```diff
@@ -47,25 +47,27 @@ private _cargo = _object getVariable [QGVAR(virtualItems), [
 
 private _configCfgWeapons = configFile >> "CfgWeapons"; //Save this lookup in variable for perf improvement
 
-if (_items isEqualType true && {_items}) then {
-    private _configItems = uiNamespace getVariable QGVAR(configItems);
+if (_items isEqualType true) then {
+    if (_items) then {
```
dedmen: I don't see the improvement here :D

alganthe: passing false would error out since it would try to execute the else part, now it just silently end.

dedmen: Ah didn't see the else 👍

addons/arsenal/functions/fnc_scanConfig.sqf
```diff
@@ -127,20 +127,18 @@ private _magazineGroups = [[],[]] call CBA_fnc_hashCreate;
...
 
-    private _putList = [];
-    {
-        _putList append getArray (_configCfgWeapons >> "Put" >> _x >> "magazines");
-        false
-    } count getArray (_configCfgWeapons >> "Put" >> "muzzles");
+private _putList = [];
+{
+    _putList append getArray (_configCfgWeapons >> "Put" >> _x >> "magazines");
+} foreach getArray (_configCfgWeapons >> "Put" >> "muzzles");
```
dedmen: How about removing duplicates with arrayIntersect? 
Won't be that big of a difference as the lookup is done in-engine anyway

alganthe: is it even possible to have duplicates in there?

dedmen: Nooo Idea. But I'd say bad Modders can do anything?

jonpas: Why should we be fixing what bad modders do then? Show duplicates, so they are more likely to remove them even.

alganthe:
That's all the fixes I wanted to push.
I wanted to add "right click on a listbox item to copy the classname" but apparently there's no UI EH for right clicking a listbox item.

commy2:
What about shift click?

SilentSpike:
Alternatively, just add a generic mouseButtonDown EH and check for right click, the retrieve the current item

addons/arsenal/functions/fnc_scanConfig.sqf
SilentSpike: It wouldn't even matter if there were duplicates, since you're just checking below if the classname is present in the list

PR #6001, opened 2018-01-02
add allowFadeMusic setting

commy2:
- close #5320
- same idea, but with the requested setting

Couldn't push to remote branch, so 2nd PR it is...


addons/common/stringtable.xml
```diff
@@ -1151,5 +1151,13 @@
             <Chinesesimp>重量:</Chinesesimp>
             <Chinese>重量:</Chinese>
         </Key>
+        <Key ID="STR_ACE_Common_AllowFadeMusic">
+            <English>Allow muting music</English>
```
SilentSpike: Are we muting the music though? I think it'd be a bit clearer like "Allow music volume adjustment"

commy2: You're right. I read `0 fadeMusic` wrong.

PR #5755, opened 2017-11-13
Scopes - Auto convert rifle/ammo precision to dispersion

ulteq:
Converts rifle/ammo precision (a) into dispersion config values (b).

Meets the following criteria:
* 68.3 % of the shots land within a circle of diameter (0.5 * a)
* 99.6 % of the shots land within a circle of diameter (1.0 * a)

- [ ] Needs discussion since it might impact gameplay for some people.

| Vanilla | ACE3 |
| -------- | -------- |
| dispersion_vanilla | dispersion_ace3 |
| dispersion_vanilla | dispersion_ace3 |

PR #6003, opened 2018-01-02
R3F compat: dispersion overhaul & AtragMx presets

Laid3acK:
- Dispersion overhaul for all R3F rifles :
    - Assuming that a dispersion depends of many, many parameters (rifle/ammo harmonisation, position, supports or not, shooter, etc ...).
    - Assuming that a dispersion is not a constant value but increases with the distance.
    - Assuming that a bipod is not a Bench Rest support.
    - Assuming that all muzzles velocities are correct with RL values.
    - According with https://github.com/acemod/ACE3/pull/5755 and the `0.562` value (should be increased a little bit imo).

For all rifles, I got the corresponding distance at the end of the supersonic domain (approximately Mach 1.2 or 408 m/s at the normal conditions) with Strelok Pro and applied a dispersion's square of 50/50 cm for the assault rifles, a dispersion's square of 30/30 cm for the Marksman and Sniper rifles.

By this way, i obtain a dispersion around 3 MOA for the assault rifles, 1.5 MOA for the Marksman rifles, 1 MOA for the Sniper rifles.

> Example with the added HK416 S (11" barrel) dispersion : 
408 m/s at the normal conditions = 455 meters, with a dispersion's square of 50/50 cm -> 3.78 MOA.
3.78 x 0.562 =  2.12 MOA, the effective dispersion in-game.

The goal isn't an utopian "real" dispersion but something coherent with military shooting and the (splendid) ArmA 3 dispersion system.

- AtragMx presets according with the [ATragMX Framework](https://ace3mod.com/wiki/framework/atragmx.html).
- Vertical and horizontal increments updated for the Zeiss scope.
- Fix a "ghost" magazine with the FAMAS FELIN.
- Minor changes for sniper scopes.
- And Space Wars ! 

*May the Force be with you !*

### DIFF
--- a/optionals/compat_r3f/CfgWeapons.hpp
+++ b/optionals/compat_r3f/CfgWeapons.hpp
@@ -176,52 +176,12 @@ class CfgWeapons {
     class R3F_HK416M_M203: R3F_HK416M {
         muzzles[] = {"this","Lance_Grenades"};
     };
-    class R3F_HK416M_HG: R3F_HK416M {};
-    class R3F_HK416S_HG: R3F_HK416M_HG {
+    class R3F_HK416S_HG: R3F_HK416M {
         ACE_barrelLength = 279.4;
-        class Single: Mode_SemiAuto {
-            sounds[] = {"StandardSound","SilencedSound"};
-            class BaseSoundModeType {
-                weaponSoundEffect = "DefaultRifle";
-                closure1[] = {"A3\sounds_f\weapons\closure\closure_rifle_6",0.38999999,1,30};
-                closure2[] = {"A3\sounds_f\weapons\closure\closure_rifle_7",0.38999999,1,30};
-                soundClosure[] = {"closure1",0.5,"closure2",0.5};
-            };
-            class StandardSound: BaseSoundModeType {
-                begin1[] = {"\r3f_armes\sons\hk416_1",3.5,1,1200};
-                begin2[] = {"\r3f_armes\sons\hk416_2",3.5,1,1200};
-                begin3[] = {"\r3f_armes\sons\hk416_3",3.5,1,1200};
-                soundBegin[] = {"begin1",0.333,"begin2",0.333,"begin3",0.333};
-            };
-            class SilencedSound: BaseSoundModeType {
-                begin1[] = {"A3\sounds_f\weapons\silenced\silent-18",0.80000001,1,300};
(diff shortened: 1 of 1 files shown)


optionals/compat_r3f/CfgWeapons.hpp
```diff
@@ -75,56 +108,156 @@ class CfgWeapons {
...
+                begin2[] = {"\r3f_armes\sons\hk416_2",3.5,1,1200};
+                begin3[] = {"\r3f_armes\sons\hk416_3",3.5,1,1200};
+                soundBegin[] = {"begin1",0.333,"begin2",0.333,"begin3",0.333};
+            };
+            class SilencedSound: BaseSoundModeType {
+                begin1[] = {"A3\sounds_f\weapons\silenced\silent-18",0.80000001,1,300};
+                begin2[] = {"A3\sounds_f\weapons\silenced\silent-19",0.80000001,1,300};
+                begin3[] = {"A3\sounds_f\weapons\silenced\silent-11",0.80000001,1,300};
+                soundBegin[] = {"begin1",0.333,"begin2",0.333,"begin3",0.333};
+            };
```
commy2: Sound config :confused:

Laid3acK: Yep, i know ...
Without this, i have a beautiful pop up in-game ...

commy2: But does it belong here? I don't mind it personally. Is R3F still maintained?

Laid3acK: I'm not a R3F team member and i don't know if they will make a ACE3 compat.
But, in all cases, at your own convenience guys. :)

commy2: I mean, this error pop up would also happen without ACE, right?

Laid3acK: Yes, due to the added dispersion for this rifle.

commy2: Why don't you inherit the e.g. class Single from the rifles base class then, instead of inheriting Mode_SemiAuto? That'd inherit the sounds too.

Laid3acK:
Strangely, no pop-up and sound bug anymore.
Thank you. :)

PR #6007, opened 2018-01-02
Deprecate common function numberToDigitsString

SilentSpike:
- Deprecate common function `numberToDigitsString` in favour of `CBA_fnc_formatNumber`
- Optimise common function `numberToDigits`

### DIFF
--- a/addons/common/functions/fnc_numberToDigitsString.sqf
+++ b/addons/common/functions/fnc_numberToDigitsString.sqf
@@ -16,6 +16,6 @@
  */
 #include "script_component.hpp"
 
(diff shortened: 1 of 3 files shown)


addons/common/functions/fnc_numberToDigitsString.sqf
```diff
@@ -16,20 +16,6 @@
  */
 #include "script_component.hpp"
 
-params ["_number", "_minLength"];
+ACE_DEPRECATED(QFUNC(numberToDigitsString),"3.14.0",CBA_fnc_formatNumber);
```
SilentSpike: Realised I forgot to put CBA func in strings here

PR #6011, opened 2018-01-03
Add custom repair positions to all quadbikes

Dystopian:
NATO quadbike was added in #2245, don't know why not all.


jonpas:
Because I am stupid and forgetful and used NATO one for testing and then probably just assumed that's it when I looked over it at the end. Nice catch.

Shows that almost nobody uses quad bikes though!

PR #6012, opened 2018-01-03
Add spectator handling of grenades

AACO:
- Add spectator handling of grenades (vanilla and advanced)

This is needed since grenades and other large/slow projectiles are actually handled globally instead of locally, and the projectile returned to the handler is null.

~~Note: we can probably drop special handling of advanced throwing grenades if we change the BI's fired event handler to use CBA's (since https://github.com/acemod/ACE3/blob/master/addons/advanced_throwing/functions/fnc_throwFiredXEH.sqf *should* handle it)~~


jonpas:
> Note: we can probably drop special handling of advanced throwing grenades if we change the BI's fired event handler to use CBA's

Yes, XEH should be used, that's the whole reason of us throwing it.

AACO:
@jonpas do you have a good starting point for that? It looks like my options in CBA are:
1. add a BIS event handler to the object, which from what I can see wont do what's needed here
or
2. Add a class based XEH that would have to be handled every fire, which seems very clunky for the usage here.

Am I missing something, or is there no way to handle adding / removing object based extended events?

PabstMirror:
xeh doesn't really have a clean way to remove EH, 
so I think the current system is the best we can do

AACO:
>As for the advanced throwing part, I'm not sure that will work because that function is using the _thisEventHandler variable.
The advanced throwing event handler should get removed before the conditions trigger the fired event removal, with that said having dedicated grenade handling code for advanced throwing might not be the worst thing ever, my only concern is an update to grenade handling code has to be tracked in two areas. (For example if you wanted to separate grenade icons by grenade types, you'd have to update the advanced throwing and the fired event handler code). Not the worst thing in the world.

>I think I'd personally just add the event in post init (so it's always there)
I'm personally not a fan of adding handlers until they're needed, so I do want to keep the add/remove flow in

jonpas:
> I'm personally not a fan of adding handlers until they're needed, so I do want to keep the add/remove flow in

I am not sure about this, but adding/removing a handler might be more expensive overall than adding it on start and then just checking a simple boolean.

SilentSpike:
> The advanced throwing event handler should get removed before the conditions trigger the fired event removal

Keep in mind that units could be removed from the viewable list which were previously in it (via public API or even mid-mission settings change). So this is only true for when spectator is closed, I think dedicated code for advanced throwing is needed.

> my only concern is an update to grenade handling code has to be tracked in two areas

We could make a local event to add projectiles into the system, that way it could also be called independently in other components too.

AACO:
@SilentSpike something like that? (also why does circle ci keep failing?)

jonpas:
> also why does circle ci keep failing?

Just not setup yet from where you branched off, don't worry about it.

addons/spectator/functions/fnc_ui.sqf
```diff
@@ -93,6 +93,39 @@ if (_init) then {
...
+    ] call CBA_fnc_addEventHandler;
+
+    // register advanced throwing EH
+    GVAR(advancedThrowingEH) = [
+        QEGVAR(advanced_throwing,throwFiredXEH), {
+            // Fire time used for unit icon highlighting
+            (_this select 0) setVariable [QGVAR(highlightTime), time + FIRE_HIGHLIGHT_TIME];
+
+            // add grenade to tracking
+            [QGVAR(addToGrenadeTracking), [_this select 6]] CBA_fnc_localEvent;
```
Dystopian: `call` is missing

AACO: Good catch

PR #6019, opened 2018-01-03
Interaction - Add Flip and Push actions to Quad Bike, Kart and Static Weapons

Dystopian:
- title;
- add `Flip` action to static weapons;
- fix `Push` action Russian translation.

Inspired by ACE2 `Flip ATV` action.

There is `str_a3_cfgvehicles_quadbike_01_base_f_useractions_pressxtoflipthequad0` string (`Flip the quad bike`) in vanilla already, but it's not implemented and it's not universal (my one can be used for any vehicle) and it has very bad Russian translation.

Also in ACE2 there was `abs (speed _target) < 3` in condition, but I can't see any sense to use it with ACE3 interaction menu.


commy2:
Add this also to the Karts?

Dystopian:
Karts already have `Set-up the Kart` vanilla useraction, but it's not so good. Besides `Push` action would be useful for karts. So yes, will add this.

What do you think about `Flip` for static weapons? I think this would be useful too.

Drofseh:
> What do you think about `Flip` for static weapons? I think this would be useful too.
I'd love to see a `Flip` for statics.

bux:
Can I have one for tables?
`(╯°□°）╯︵ ┻━┻`

shukari:
You can "Flip" static weapons allready by drag/carry them and place them again.

Dystopian:
@bux looks like we want small framework with `canBeFlipped` and `canBePushed` config and var values and PostInit menu adding :) Do we?
@shukari dragging can be disabled and iirc it doesn't align object to terrain.

SilentSpike:
My only fear with this is that a vehicle might get arma'd and clip something and explode 😆 Looks good though!

commy2:
As long as he does `setVectorUp surfaceNormal`, it at least can't collide with the terrain.
Collisions with anything else are just as bad as dragging etc.

addons/interaction/script_component.hpp
```diff
@@ -17,3 +17,6 @@
 #include "\z\ace\addons\main\script_macros.hpp"
 
 #define MACRO_DOOR_REACH_DISTANCE (AGLToASL positionCameraToWorld [0,0,0] vectorDistance AGLToASL (ACE_player modelToWorld (ACE_player selectionPosition "Head"))) + 2
+
+// here is most suitable mass for all vanilla, CUP and RHS statics ATM (ZU-23 suits, D-30 and M119 don't suit)
```
jonpas: Ditch "here is", it's obviously here if the comment is here. ^^

addons/interaction/CfgVehicles.hpp
```diff
@@ -368,6 +368,40 @@ class CfgVehicles {
...
+                };
+            };
+        };
+    };
+    class Kart_01_Base_F: Car_F {
+        class ACE_Actions: ACE_Actions {
+            class ACE_MainActions: ACE_MainActions {
+                class GVAR(flip) {
+                    displayName = CSTRING(Flip);
+                    condition = QUOTE(1 < (vectorUp _target) vectorDistance (surfaceNormal getPosATL _target) && {0 == {alive _x} count crew _target});
```
jonpas: First condition should be made into a common function, as it repeats.

Dystopian: I didn't want to move such a small piece of code to function.

Dystopian:
@jonpas No, thank you :) I think bux just joked.

Snak3Doc:
Possibly add push actions to boat class? a fix for the beached RHIB

Dystopian:
@Snak3Doc already in my todo list

PR #6021, opened 2018-01-05
Optimise spectator loops

SilentSpike:
`count` is only quicker than `forEach` when the final command returns a
boolean or nil. `forEach` is quicker if the final command returns any
other data type.

### DIFF
--- a/addons/spectator/functions/fnc_ui_draw3D.sqf
+++ b/addons/spectator/functions/fnc_ui_draw3D.sqf
@@ -72,8 +72,6 @@ if !(GVAR(uiMapVisible)) then {
 
             // Draw icon
             drawIcon3D _icon;
-
-            nil // Speed loop
         } count GVAR(iconsToDraw);
         END_COUNTER(drawTags);
     };
@@ -119,13 +117,9 @@ if !(GVAR(uiMapVisible)) then {
                         drawLine3D [_oldLoc, _locNew, _colorNew];
                     };
                     _oldLoc = _locNew;
-
-                    nil // Speed loop
                 } count _segments;
             };
-
-            nil // Speed loop
-        } count GVAR(projectilesToDraw);
+        } forEach GVAR(projectilesToDraw);
         GVAR(projectilesToDraw) = _projectilesNew;
 
         {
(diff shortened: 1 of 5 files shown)


addons/spectator/functions/fnc_ui_draw3D.sqf
```diff
@@ -138,9 +132,7 @@ if !(GVAR(uiMapVisible)) then {
                 // Store grenade for next frame
                 _grenadesNew pushBack _x;
             };
-
-            nil // Speed loop
-        } count GVAR(grenadesToDraw);
+        } forEach GVAR(grenadesToDraw);
```
dedmen: pushBack returns nil so you can still use count

addons/spectator/functions/fnc_ui_updateIconsToDraw.sqf
```diff
@@ -139,9 +139,7 @@ private _camPos = AGLToASL positionCameraToWorld [0,0,0];
             SETVAR(_vehicle,GVAR(firedEH),_vehicle addEventHandler [ARR_2("Fired",{_this call FUNC(handleFired)})]);
         };
     };
-
-    nil // Speed loop
-} count ([] call FUNC(getTargetEntities));
+} forEach ([] call FUNC(getTargetEntities));
```
dedmen: I'm not sure what `if` returns when the then didn't run and there is no `else`. But I think there is a nil on the stack here meaning you can use `count`

addons/spectator/functions/fnc_ui_draw3D.sqf
dedmen: If you swap the pushBack and drawIcon3D you will have a nil on the stack and can then use `count`

dedmen:
>`count` is only quicker than `forEach` when the final command returns a
boolean or nil. `forEach` is quicker if the final command returns any
other data type.
Did you write that yourself or did you copy that from me somewhere?

the benefit from count>forEach is only really important on very big loops that don't do much work. Considering all the time that is spent in each iteration using count here might not even be worth it. The time saved relative to the total time spent.

addons/spectator/functions/fnc_ui_updateIconsToDraw.sqf
dedmen: Just looked it up. `if` returns typeless nil when the condition didn't match and there was no else.

SilentSpike:
I paraphrased my understanding of what you'd said in slack the other day. Though yeah I might just change them all to `forEach` as I imagine the performance difference is going to be negligible (until we start using intercept for stuff).

SilentSpike:
Just made them all `forEach` because the performance impact is meaningless. If we eventually start using intercept for stuff, I believe @dedmen has a much faster iteration method we can adopt.

PR #6024, opened 2018-01-05
Removing unused classname from wiki

Cyruz143:
- Remove a classname that's no longer in use from the wiki


jonpas:
Travis is dying it seems...

PR #6029, opened 2018-01-06
Add interaction to inventory holders

Dystopian:
- add `Inventory` interaction to inventory holders: dead bodies and dropped weapons/gear;
- add `Inventory` interaction to unconscious men;
- add `Drag`/`Carry` actions to dropped gear and `Carry` to dead body weapons;
- reformat and optimize `ace_dragging_fnc_canCarry` and `ace_dragging_fnc_canDrag`.

Now it's much easier to take gear from several bodies/weapons lying close to each other.

TODO:
- [x] take weapon/~~mag~~
- [x] adjust action position
- [x] drag/carry
- ~~cargo load~~
- ~~setting to disable ?~~

~~WIP~~


Dystopian:
Unfortunately dropped from dead body weapons (class `WeaponHolderSimulated`) are bad when drag/carry: they can just disappear (looks like they go far under ground), so I had to enable drag menu for dropped gear only.

BTW possibility to take weapon from unconscious enemy is amazing ;)

Dystopian:
Dead body weapons can be carried now (with small workaround).
Cargo loading and moving gear to another vehicle gear have too many troubles to fix so I decided not implement them.

This is done, please review.

addons/interaction/CfgVehicles.hpp
```diff
@@ -648,12 +648,11 @@ class CfgVehicles {
             class ACE_MainActions {
                 displayName = CSTRING(MainAction);
                 distance = 3;
-                position = QUOTE(_target worldToModel getPos _target);
+                position = QUOTE(_target worldToModel ASLToAGL ATLToASL getPosATL _target);
```
dedmen: How about getPosASL and just ASLToAGL?

Dystopian: will check

addons/interaction/XEH_postInit.sqf
```diff
@@ -116,3 +116,18 @@ GVAR(isOpeningDoor) = false;
...
+private _action = [
+    // action display name will be overwritten in modifier function
+    QGVAR(takeWeapon), "take", "\A3\ui_f\data\igui\cfg\actions\take_ca.paa",
+    {if (1 == count weaponCargo _target) then {_player action ["TakeWeapon", _target, weaponCargo _target select 0]}},
+    {1 == count weaponCargo _target},
+    nil, nil, nil, nil, nil,
+    {
+        params ["_target", "", "", "_actionData"];
+        if (1 != count weaponCargo _target) exitWith {};
+        _actionData set [1, format [localize "STR_ACTION_TAKE_BAG", getText (configfile >> "CfgWeapons" >> weaponCargo _target select 0 >> "displayName")]];
```
Dystopian: There are also `str_a3_cfgactions_takeitem0`, `str_action_takeweapon`, `str_action_takemagazine` strings, all of them have some differences in translations. I chose `STR_ACTION_TAKE_BAG` because of correct Russian translation.

addons/interaction/CfgVehicles.hpp
dedmen: Why was getPos even a problem here?

Dystopian: bad position under water and/or on building, I don't remember. In building `getPos` Z=0 IIRC.

dedmen: I think that's just the nature of AGL. In buildings it uses the ground under your feet so height is almost always 0. I don't understand how converting ASL to AGL makes a difference then

Dystopian: `ASLToAGL` doesn't take building height into account unlike `getPos`

dedmen: I think you just want ATL then.. I guess.

Dystopian: ATL didn't work under water. It really works now with AGL.

654wak654:
Moving towards a future without the action menu, one PR at a time 👍

addons/interaction/XEH_postInit.sqf
Cuel: Why yoda conditions? :p

Dystopian: `short == very_long_expression_with_not_observed_end` trying to improve code look :)

severgun:
Moving whole weaponholder seems weird for me and far from "realism" theme of ACE3 mod.
You can drop tonns of ammo on a ground and then carry it around in hands.

dedmen:
^ I agree. Also picking up a invisible force matrix that holds the items together seems.. Not right.
I can see it looking acceptable for a single weapon. But even then you might aswell just pick it up.

Also the last couple games I played I tried several times to find the "open inventory" action on a dead body and wondered why I couldn't find it. Completely forgot that this isn't merged and released already.

Dystopian:
@severgun you can't `drop tonns of ammo on a ground and then carry it`

@dedmen I think I do can hold some magazines and weapons in hands. I think this is not really big difference from other objects carrying.

dedmen:
>you can't `drop tonns of ammo on a ground and then carry it`

What prevents you from doing that?

Dystopian:
https://github.com/acemod/ACE3/blob/master/addons/dragging/functions/fnc_getWeight.sqf

dedmen:
So.. Nothing does?
You still can carry as much as with boxes, just without the box now then.
Hm. Okey.

severgun:
> you can't drop tonns of ammo on a ground and then carry it

tonns not as value of mass but of quantity here.

Dystopian:
Max dragged items count is restricted to 3.
Sorry for large commit, functions reformat was really necessary.

654wak654:
This needs some performance testing to see the impact of all these new actions, and regardless of that impact, a setting similar to that of building interactions would be nice.

dedmen:
I don't see any performance impact here. If you don't hold the action button then they do nothing. And if you do I don't think a handful of new dots will make a big difference, in the rare cases where you have many inventories infront of you.

Though, why that performance regression in canDrag/canCarry

`{vehicle _target isEqualto _target}`
->
`{vehicle _target == _target}`
? What's the reasoning behind that?

Dystopian:
> Though, why that performance regression in canDrag/canCarry
> 
> `{vehicle _target isEqualto _target}`
> ->
> `{vehicle _target == _target}`
> ? What's the reasoning behind that?

It was done with `isEqualto` in #6780 after this PR was done. So it was just replaced while upstream merging.
Besides `isEqualto` doesn't give real speed up for objects (tested) and also looks worse.

Neviothr:
Use https://github.com/CBATeam/CBA_A3/pull/1228?

addons/interaction/XEH_postInit.sqf
```diff
@@ -135,3 +135,20 @@ GVAR(isOpeningDoor) = false;
...
+    {1 == count weaponCargo _target},
+    nil, nil, nil, nil, nil,
+    {
+        params ["_target", "", "", "_actionData"];
+        _actionData set [1, format [localize "STR_ACTION_TAKE_BAG", getText (configfile >> "CfgWeapons" >> weaponCargo _target select 0 >> "displayName")]];
+    }
+] call EFUNC(interact_menu,createAction);
+{
+    [_x, 0, ["ACE_MainActions"], _action, true] call EFUNC(interact_menu,addActionToClass);
+} forEach ["WeaponHolder", "WeaponHolderSimulated"];
```
Neviothr: Use https://github.com/CBATeam/CBA_A3/pull/1228?

Dystopian: Looks like I wanted not to create extra function for action modifying. So no changes are required here. Anyway I removed XEH creating lines.

Dystopian:
Implemented as standalone addon in [v7 ace additions](https://steamcommunity.com/sharedfiles/filedetails/?id=2025187921).

Dystopian:
Can't repeat anymore bug with weapon carry-drop when it falls under ground. Maybe because in 1.98:
```Fixed: Simulated things should no longer fall too fast and fall through the ground (e.g. weapons)```
I removed workaround.

10Dozen:
Is it possible to allow player to run when e.g. only 2 items is in hands and it's not weights much? Like - to pick 2 AT-4 from truck and run back to positions, instead of walking/finding big backpack. Stamina drain may be at the same level, as it's harder to keep balance.

Dystopian:
> Is it possible to allow player to run when e.g. only 2 items are in hands and total weight is about <=10-15 kg?

I will try to implement this in other PR.

LinkIsGrim:
Works, including MP. Functionality and performance are fine, but see https://github.com/Dystopian/ACE3/pull/2 for code suggestions.

Also, there's a double-up when using this with disarming:


But that should be fixed after merging.

LinkIsGrim (approved):

PR #6030, opened 2018-01-06
Fix back blast if damage not allowed

Dystopian:

### DIFF
--- a/addons/overpressure/functions/fnc_firedEHBB.sqf
+++ b/addons/overpressure/functions/fnc_firedEHBB.sqf
@@ -54,8 +54,8 @@ if (_distance < _backblastRange) then {
     if (isClass (configFile >> "CfgPatches" >> "ACE_Medical") && {([_unit] call EFUNC(medical,hasMedicalEnabled))}) then {
         [_unit, _damage, "body", "backblast"] call EFUNC(medical,addDamageToUnit);
     } else {
-        TRACE_1("",isDamageAllowed _x);
-        if (!isDamageAllowed _x) exitWith {}; // Skip damage if not allowed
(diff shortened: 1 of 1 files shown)


PabstMirror:
good catch

addons/overpressure/functions/fnc_firedEHBB.sqf
```diff
@@ -54,8 +54,8 @@ if (_distance < _backblastRange) then {
     if (isClass (configFile >> "CfgPatches" >> "ACE_Medical") && {([_unit] call EFUNC(medical,hasMedicalEnabled))}) then {
         [_unit, _damage, "body", "backblast"] call EFUNC(medical,addDamageToUnit);
     } else {
-        TRACE_1("",isDamageAllowed _x);
-        if (!isDamageAllowed _x) exitWith {}; // Skip damage if not allowed
+        TRACE_1("",isDamageAllowed _unit);
+        if (!isDamageAllowed _unit) exitWith {}; // Skip damage if not allowed
         _unit setDamage (damage _unit + _damage);
```
commy2: Any reason this is an `exitWith` when it could be a `then` for simplicity?

Dystopian: I like `exitWith` :)

commy2: Same. But I like classical control structures even more.

Dystopian: classical one makes too many indents

PabstMirror: I think I might have just been trying to minimize the git diff, which isn't really a good reason

Issue #69, opened 2015-01-19
Improve flashbang effect


bux578:
bux578:
> IRL if you're hit by a military grade flashbang you'll suffer a (in germany called: h90/3 type) sudden hearing loss, serious loss of orientation and eye damage like if you'd do arc welding without a safety mask. This usually leads to two weaks of being unable to fight (with potential hospital stay). That wouldn't be much fun in arma ... so this needs to be balanced to still be fun.

Kristian/VKing
> Flashbangs should knock you out for at least long enough for someone to clear the room you're in
it really needs to be limited to Line of Sight, though
if there's someway to make it work, it should have a greater effect indoors, even if you don't see it, than outdoors

esteldunedain:
Done in #418

CG-Delta:
My experiment doesn't quite match what I read here. I'm thinking of one called 6-bang which I believe is the one linked here:
http://www.angelopodesta.com/documenti/07_d380e0111_6bang.pdf

> bux578:

> IRL if you're hit by a military grade flashbang you'll suffer a (in germany called: h90/3 type) sudden hearing loss, serious loss of orientation and eye damage like if you'd do arc welding without a safety mask. This usually leads to two weaks of being unable to fight (with potential hospital stay). That wouldn't be much fun in arma ... so this needs to be balanced to still be fun.

Apart from throwing it on room clearing exercise on various occasions, we did one time try it on our selves. 10 people cramped together in a small, dim room, normal clear goggles, double hearing protection and dropped one at our feet. Some had eyes closed some not, either way it is not something that will damage your eyes at all. More like if you're looking into a camera flash in a dark room, if you're looking straight at the thing. 
Thing is. When clearing a room you'll go in after the first bang (5 to follow). The people going in are not blinded. That would be impractical. As for hearing. You'll go in with hearing protection (which you use anyway) and that will drastically reduce it's effect.

> Pain would perhaps be appropriate if the flash bang explosion was close enough to a unit. Let's say within 1 meter? Once the medical system is more developed, we could even go for burn injuries.

About damage. Just no. At least the six bang I'm talking about. It has been jumping between our feet. Apart from pressure/sound shock there's no pain involved at all.

> Military grade flashbangs even contain a small portion of CS gas.

[...]

PR #6034, opened 2018-01-07
Grenades - Make flashbang reaction more authentic

BaerMitUmlaut:
- Make players flinch and AI try to look away as a reaction to a flashbang explosion - previously both players and AI turned 180° away from the explosion, making it very confusing and weird for players to be suddenly turned around
- Improve effect strength calculation for AI using eye position instead of ground position

### DIFF
--- a/addons/grenades/functions/fnc_flashbangExplosionEH.sqf
+++ b/addons/grenades/functions/fnc_flashbangExplosionEH.sqf
@@ -135,6 +135,7 @@ if (hasInterface && {!isNull ACE_player} && {alive ACE_player}) then {
     };
 
     // Make player flinch
+    if (_strength <= 0.2) exitWith {};
     private _minFlinch = linearConversion [0.2, 1, _strength, 0, 60, true];
     private _maxFlinch = linearConversion [0.2, 1, _strength, 0, 95, true];
     private _flinch    = (_minFlinch + random (_maxFlinch - _minFlinch)) * selectRandom [-1, 1];
--- a/addons/grenades/functions/fnc_flashbangExplosionEH.sqf
+++ b/addons/grenades/functions/fnc_flashbangExplosionEH.sqf
@@ -46,7 +46,7 @@ private _affected = (ASLtoAGL _grenadePosASL) nearEntities ["CAManBase", 20];
 _affected = _affected - [ACE_player];
 {
     if (local _x && {alive _x}) then {
-        private _strength = 1 - (((getPosASL _x) vectorDistance _grenadePosASL) min 20) / 20;
+        private _strength = 1 - (((eyePos _x) vectorDistance _grenadePosASL) min 20) / 20;
 
         TRACE_3("FlashBangEffect Start",_x,((getPosASL _x) vectorDistance _grenadePosASL),_strength);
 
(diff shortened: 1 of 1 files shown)


bux:
https://github.com/acemod/ACE3/issues/69#issue-54737954

Proper flashbang effect is very important to me.

BaerMitUmlaut:
No change in effectiveness imo, because you're still blinded and your aim will still be off. The 180° rotation is just unintuitive, because you can't give feedback to the player that he did that until the blinding effect goes away. Unless you have a heavy concussion you should still somewhat know what movement you just did, and that is not the case with a full rotation.

Effects with this PR are:
- AI turns away between 40° and 135° from the explosion.
- Players get an offset added to their aim direction, which is randomized but also has a minimum, so you will always flinch somewhat. The maximum effect strength now will be turning away between 45° and 90°. I would add a vertical offset to the aim, but I don't think that's possible.

SilentSpike:
Yeah I don't believe we'd get a better implementation without being able to fake mouse input (`setMousePosition` does nothing when used during gameplay right?).

I like this aproach 👍 Without being able to give the feedback that would be given by equilibrioception  in real life (i.e. being able to tell they'd turned around even though they can't see), I think it's bad design to turn players around completely.

bux:
I've tried it and I like the randomization. I'd like to see a higher minimum offset of maybe 60°. For AI this is fine for me. Players though shouldn't have a reduced penalty compared to AI.

BaerMitUmlaut:
The idea of the higher offset for AI is that they try to turn away from the flashbang. The player might do that intuitively, the AI won't, thus we force them.

jonpas:
@BaerMitUmlaut @bux ?

thojkooi:
@bux, did the change. Also increased max flinch a little bit.

addons/grenades/functions/fnc_flashbangExplosionEH.sqf
```diff
@@ -134,8 +134,10 @@ if (hasInterface && {!isNull ACE_player} && {alive ACE_player}) then {
         }, [], 17 * _strength] call CBA_fnc_waitAndExecute;
     };
 
-    if (_strength > 0.2) then {
-        ACE_player setVectorDir (_eyePos vectorDiff _grenadePosASL);
-    };
+    // Make player flinch
```
PabstMirror:
```suggestion
    // Make player flinch
    if (_strength <= 0.2) exitWith {};
```

PabstMirror: Less for performance, more just to prevent doing extra `setDir`
I figure we need to think about how it will work with that 9-flash going off

CG-Delta:
https://github.com/acemod/ACE3/issues/69#issuecomment-585476981

commy2:
Are you considering doing that or is this a low key request?

CG-Delta:
I have no such ability no. It's a suggestion/request/correction.

commy2:
Okay. Feel free to repost your comment. But it really did not belong on a 5 year old closed issue.

CG-Delta:
Better there than here, I thought.

commy2:
Nope.

Issue #6040, opened 2018-01-10
AtragMX Bullet Weight 


SgtKrunch:
**Arma 3 Version:** `x.xx` (stable / rc / dev)
**CBA Version:** `3.x.x` (stable / dev + commit hash)
**ACE3 Version:** `3.12.0` stable

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- When entering bullet weight into the AtragMX for custom entries, I've noticed that when switching between gr and gm, gr usually loses a little weight because of the conversion between the two. It doesn't allow for decimal entries, it just rounds up to the next whole number.

**Steps to reproduce:**
- Enter the gr of a bullet when under the Emperial system in the AtragMX, then switch to the the Metric system, then back to the emperial system and the bullet will have lost weight.

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

**Placed Modules:**
- Advanced Ballistics, wind deflection

**RPT log file:**
- Add a link ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the client and/or server RPT file. An instruction to find your RPT files can be found [here](https://community.bistudio.com/wiki/Crash_Files#Arma_3).
- If possible at the time the bug is encountered, go to Options and select "ACE Debug To Clipboard", this will print extensive debug information to the RPT file and copy it to clipboard.

ulteq:
I will look into it.

PR #6041, opened 2018-01-10
ATragMX - Allow one decimal place for the bullet mass

ulteq:
The bullet mass (in [grains](https://en.wikipedia.org/wiki/Grain_(unit))) is still going to be limited to integers.

Closes: #6040

### DIFF
--- a/addons/atragmx/functions/fnc_update_gun.sqf
+++ b/addons/atragmx/functions/fnc_update_gun.sqf
@@ -24,7 +24,7 @@ if (GVAR(currentUnit) != 2) then {
 if (GVAR(currentUnit) != 2) then {
     ctrlSetText [110, Str(Round((GVAR(workingMemory) select 12) * 15.4323584))];
 } else {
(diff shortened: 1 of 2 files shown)


ulteq:
There are more than 60 other places in the code where I would also have to use `toFixed` ...

Can we do that in a separate pull request please?

SilentSpike:
I'm not sure I understand the objection, why do it in a separate pull request when we could just do it in this one? 😕

It seems a little silly to implement something one way just to turn around and immediately change it

ulteq:
> why do it in a separate pull request when we could just do it in this one?

The massive cleanup would hide the actual code change.

SilentSpike:
> The massive cleanup would hide the actual code change.

Oh I see now, I just meant on the one line where you'd changed the code. On the other hand, if we were going to change them all you'd probably want to do the cleanup in a separate commit anyway - thus the code change would remain visible.

PR #6042, opened 2018-01-10
Documentation: extending header guidelines

Dorbedo:
- extends the header guidlines
- multiple input types
- default value(s)

### DIFF
--- a/docs/wiki/development/coding-guidelines.md
+++ b/docs/wiki/development/coding-guidelines.md
@@ -178,6 +178,10 @@ Every function should have a header of the following format as the start of thei
  * Arguments:
  * 0: The first argument <STRING>
  * 1: The second argument <OBJECT>
+ * 2: Multiple input types <STRING|ARRAY|CODE>
+ * 3: Optional input <BOOL> (Default: true)
+ * 4: Optional input with multiply types <CODE|STRING> (Default: {true})
+ * 5: Not mendatory input <STRING> (Default: nil)
  *
  * Return Value:
  * The return value <BOOL>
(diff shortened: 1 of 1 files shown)


docs/wiki/development/coding-guidelines.md
```diff
@@ -178,6 +178,10 @@ Every function should have a header of the following format as the start of thei
  * Arguments:
  * 0: The first argument <STRING>
  * 1: The second argument <OBJECT>
+ * 2: Multiple input types <STRING|ARRAY|CODE>
```
Dystopian: this doesn't look consistent with https://github.com/acemod/ACE3/pull/5253/files#diff-24615c452e26c6f9032ec7e9bc660c61R156 (`\<([\s\w]+?)\>`)

bux: which could be changed

docs/wiki/development/coding-guidelines.md
```diff
@@ -178,6 +178,10 @@ Every function should have a header of the following format as the start of thei
...
+ * 3: Optional input <BOOL> (default: true)
+ * 4: Optional input with multiple types <CODE|STRING> (default: {true})
+ * 5: Not mandatory input <STRING> (default: nil)
```
SilentSpike: This is the only syntax I'm not sure about here, I think default should always be an actual value.

Dorbedo: What about e.g.: [fnc_getWeaponState](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getWeaponState.sqf)

SilentSpike: Yeah I think `(optional)` is bad, I guess maybe it would be better as `None` instead of `nil` to be consistent with functions which return `nil` or have no input parameters

SilentSpike: 🤔 Actually, `None` doesn't really fit right in this context. I think you're right to go with `nil` on consideration.

Dorbedo: I must add, this is a very special case. I think I've found only 2 functions, where this could be used inside ace.
It's more like to have this case defined instead of using it on a regular basis.

SilentSpike:
I would happily approve this now, but will leave it to @bux as the documentation overlord

PR #6044, opened 2018-01-11
Match case for magazines in fillRightPanel (ACE Arsenal)

alganthe:
- Match case for magazines in fillRightPanel.
 
For some reason you're able to have different case between the entry in the weapon magazines array and its actual configName, meaning that some mags won't show up in the currentMag and currentMag2 tabs even tho they should.


dedmen:
What about my proposed fix of actually using the proper case in scanConfig instead of toLowering at runtime each time the right panel is refreshed?

alganthe:
Replace my lazy ass fix with ded's, I also removed closeOnMissionEnd after noticing some weird behaviors in 3DEN.

PR #6049, opened 2018-01-14
Fix house-type object carrying in building

Dystopian:
If you carry "house" type object (like `Portable light`) inside building and drop it, it will be aligned to terrain level instead of building floor. This can be avoided with `lineIntersectsSurfaces` using.

This is #3497 partial rewrite.


jonpas:
Tested underwater? It was specifically changed to use ATL position for that reason.

Dystopian:
Tested, works.
And exactly when it was changed to ATL it was broken :P

jonpas:
It was a lot more broken before, because it didn't work at all underwater! :p

Cuel:
Wasn't it changed for ships?

Dystopian:
for underwater actions https://github.com/acemod/ACE3/pull/5521/files#diff-a8b7a0c8987449950215a723aa6874b6L27

PR #6059, opened 2018-01-17
Add seperate setting for NVG noise intensity

BaerMitUmlaut:
- Add a seperate setting for NVG noise intensity

A seperate setting for this makes sense as it can cause headaches with long time use. The other effects do not cause this issue. Was also requested in #3594 by @ghost.

### DIFF
--- a/addons/nightvision/ACE_Settings.hpp
+++ b/addons/nightvision/ACE_Settings.hpp
@@ -14,6 +14,14 @@ class ACE_Settings {
         value = 1;
         sliderSettings[] = {0, 2, 1, 1};
     };
+    class GVAR(noiseScaling) {
+        category = CSTRING(Category);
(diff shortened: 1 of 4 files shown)


jonpas:
Should it not be a client setting in that case?

Should also be added via CBA Settings directly.

BaerMitUmlaut:
No, because it's also a difficulty option.
I don't think it's a good idea to have half CBA, half ACE settings. Either convert all or none.

jonpas:
> Either convert all or none.

All, and use `movedToSQF` like in markers.

BaerMitUmlaut:
That's not the point of this PR and should be done in a seperate one. Small, granular changes.

jonpas:
Sure, but before next release then. Less `movedToSQF` the better. Doesn't matter during development as we obviously don't provide backwards compatibility that time.

PR #6078, opened 2018-01-24
Add Camping Light interaction

Dystopian:
- add interaction to Camping Light (Turn On/Off, Drag/Carry, etc.);
- cleanup Portable Light interaction config;
- add icon to both lights actions.

land_camping_light_f
Camping light (lantern) is undestroyable light source which presents in some missions and sometimes it's good to turn it on/off. This object doesn't have any hitpoints and workaround should be used. Fortunately it can be handled with `CollisionLightOff` action.

TODO
- [x] drag/carry
- [x] cargo load
- [x] handle `Land_Camping_Light_off_F` objects - need to replace it with `Land_Camping_Light_F`

~~WIP~~


Dystopian:
This PR is done but cargo loading requires little cargo init rework, PR is coming.

Dystopian:
Cargo loading will work after #6085.

Dystopian:
Yeah, I should have checked this.
One correction: not server but owner. I have checked this :)

addons/interaction/XEH_postInit.sqf
```diff
@@ -28,6 +28,11 @@ ACE_Modifier = 0;
     {_lamp setHit [_x select 0, (_x select 1) max _disabledLampDMG];nil} count _hitPointsDamage;
 }] call CBA_fnc_addEventHandler;
 
+[QGVAR(switchCollisionLight), {
+    params ["_target", "_player", "_state"];
+    _player action [["CollisionLightOff", "CollisionLightOn"] select _state, _target];
```
PabstMirror: Would https://community.bistudio.com/wiki/setCollisionLight be simpler?

Dystopian: now I see why I chose action instead of this command. Because of this note:
```
Note that the vehicle has to be local, for global variant use Arma 3 Actions "CollisionLightOn"/"CollisionLightOff"
```
I didn't want to use CBA events and just used these actions.
Could you remove this note from wiki please? I'll replace action with command.

Dystopian:
@PabstMirror https://github.com/acemod/ACE3/pull/6078#discussion_r170463314
> Could you remove this note from wiki please?

addons/interaction/XEH_postInit.sqf
dedmen: Why remove the note? is it invalid?

Dystopian: yes, vehicle must be local for both command and action, see https://github.com/acemod/ACE3/pull/6078#pullrequestreview-99120289

PabstMirror: It might only be a problem for this object simulation type.

Dystopian: just checked Buoy, Road Cone, civil plane - all of them require local vehicle

PR #6087, opened 2018-01-26
Map - Add Camping Light support

Dystopian:
- title.

Camping Light and some other vehicles have `MarkerLights` config instead of `Reflectors` with hitpoints. Their lights should be considered when determining map light.

Unfortunately I don't know how to consider `blinking` parameter. So blinking lights will light a map like continuous light. Not a big deal I think.

### DIFF
--- a/addons/common/functions/fnc_lightIntensityFromObject.sqf
+++ b/addons/common/functions/fnc_lightIntensityFromObject.sqf
@@ -79,6 +79,21 @@ if (_lightSource isKindOf "CAManBase") then {
 
     } forEach _lights;
 
+    if (isCollisionLightOn _lightSource) then {
+        private _markerLights = [
+            _lightSource,
+            {configProperties [configFile >> "CfgVehicles" >> typeOf _this >> "MarkerLights", "isClass _x", true]},
+            uiNamespace,
+            format [QEGVAR(cache,MarkerLights_%1), typeOf _lightSource],
+            1E11
+        ] call FUNC(cachedCall);
+        {
+            private _position = _lightSource modelToWorld (_lightSource selectionPosition getText (_x >> "name"));
+            private _distance = _unitPos distance _position;
+            _lightLevel = _lightLevel max (linearConversion [0, 10, _distance, 1, 0, true]);
+        } forEach _markerLights;
+    };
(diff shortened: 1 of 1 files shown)


jonpas:
> Unfortunately I don't know how to consider `blinking` parameter. So blinking lights will light a map like continuous light. Not a big deal I think.

Would be nice to see it blinking, but yes, it's not a big deal.

addons/common/functions/fnc_lightIntensityFromObject.sqf
```diff
@@ -79,6 +79,21 @@ if (_lightSource isKindOf "CAManBase") then {
...
+            _lightSource,
+            {configProperties [configFile >> "CfgVehicles" >> typeOf _this >> "MarkerLights", "isClass _x", true]},
+            uiNamespace,
+            format [QEGVAR(cache,MarkerLights_%1), typeOf _lightSource],
+            1E11
+        ] call FUNC(cachedCall);
+        {
+            private _position = _lightSource modelToWorld (_lightSource selectionPosition getText (_x >> "name"));
+            private _distance = _unitPos distance _position;
+            _lightLevel = _lightLevel max (linearConversion [0, 10, _distance, 1, 0, true]);
```
PabstMirror: I think we should account for light brightness as there are several objects with collision lights that wouldn't be useful
maybe something like:
```
_lightLevel = _lightLevel max (linearConversion [0, 10, _distance, 1, 0, true] * linearConversion [25, 2000, getNumber (_x >> "intensity"), 0, 1, true]);
```

PabstMirror: e.g. ghosthawk collision lights

Dystopian: I didn't add `intensity` check because of this comment in `Reflectors` code above: `@todo intensity affects range?`. But it looks like you are right and intensity of `MarkerLights` is more important than for `Reflectors`. I'll add it.

Dystopian: I slightly modified `linearConversion` range to suit all found light sources.

PR #6094, opened 2018-01-28
Add setting to disable Vehicle Freelook in aircraft

Dystopian:
- add setting to disable `Vehicle Freelook` game option when pilot an aircraft.

In vanilla `Vehicle Freelook` game option can be only enabled or disabled. There are some tickets at BI tracker asking to disable it for air vehicles because it's more comfortable to control them with mouse than with keyboard.

This setting goes the most hackish way: opens Esc-menu, presses `Game` button, then switches `Vehicle Freelook` value and closes both displays. I tried to create just `Game options` display without Esc-menu but it opened with empty values. I will be glad if anyone suggest better solution.

During game this looks like small delay when enter/exit air vehicle.

I moved the setting to `ACE Uncategorized` category which should be changed to some subcategory when CBA 3.6 is released.

To use this feature you have to re-enter the vehicle when:
- pilot and activate the setting;
- get in not as pilot and then change seat to pilot.

It does work when start mission in pilot seat.

Bug: if you end mission in pilot seat, option will remain disabled until you manually enable it or enter/exit aircraft in next mission.
Also I don't know how it will go if some other display/dialog is opened and player enters/exits aircraft.

There are really many checks in code to avoid possible problems if BI changes some classes/values. Error messages are not localized because of very small probability of errors.

I will not be sorry if you say this PR doesn't suit ACE by any reason :)

### DIFF
--- a/addons/aircraft/XEH_PREP.hpp
+++ b/addons/aircraft/XEH_PREP.hpp
@@ -1,2 +1,3 @@
-PREP(initEjectAction);
 PREP(canShowEject);
+PREP(initEjectAction);
+PREP(switchVehicleFreelook);
--- a/addons/aircraft/XEH_preInit.sqf
+++ b/addons/aircraft/XEH_preInit.sqf
@@ -6,4 +6,6 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+#include "initSettings.sqf"
+
 ADDON = true;
--- a/addons/aircraft/functions/fnc_switchVehicleFreelook.sqf
+++ b/addons/aircraft/functions/fnc_switchVehicleFreelook.sqf
@@ -22,6 +22,8 @@
 
 params ["_player", "_vehicle"];
 
+if (!hasInterface) exitWith {};
+
 disableSerialization;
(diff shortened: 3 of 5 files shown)


BaerMitUmlaut:
You use a lot of hints for debugging and error logging, use ERROR macros instead, as they log to the RPT.

Dystopian:
I use hints on purpose, because it's exclusively client feature and it's not really critical and exactly player should see if something goes wrong when he enters aircraft. If error goes to RPT he won't see anything in game.
As you wish though.

jonpas:
Doesn't double tapping Alt (Freelook toggle) do exactly that?

BaerMitUmlaut:
Those hints don't explain what's going on to the user at all and will just confuse them. If there's an actual bug, tracing them through the RPT is much easier than relying on the user to remember what hint was shown.

Dystopian:
@jonpas exactly. But if option is disabled it's very annoying to Alt-Alt in every land vehicle, and if it's enabled Alt-Alt doesn't work.

jonpas:
I see. I don't know if this fits ACE3, especially due to the rather hacky nature it has to be done with. It would probably fit nicer with ACEX View Restriction.

Dystopian:
Replaced by #6097.

PR #6097, opened 2018-01-30
Add Toggle Vehicle Freelook option keybind

Dystopian:
- add keybind to toggle `Vehicle Freelook` game option.

`Vehicle Freelook` is a game option which enables constant freelook in vehicles. It's in `Esc-menu` - `Options` - `Game options` dialog. It can be handy when you often enable freelook when get in vehicle - you don't have to press `Alt-Alt` or `*` each time you get in vehicle.

But sometimes you find vehicle control with mouse better than with keyboard e.g. in air vehicles. Unfortunately you can't use `Alt-Alt` or `*` keys to temporarily disable freelook (BI feature). In these cases you have to open options dialog and disable the option and enable it back after flight.

To automate this routine keybind can be used. By default it's not set to any key.


jonpas:
This is shorter! :D

Why not bind `Alt-Alt`/`*` to it directly? So at the end all we'd do is allow what that "BI feature" prevents.

Dystopian:
> Why not bind Alt-Alt/* to it directly?
- `Alt-Alt`/`*` freelook toggling (temporary, per vehicle) is not the same as `Vehicle Freelook` option (permanent) and it's replacing could bring more problems than benefits IMO;
- `*` key is not fully intercepted with `true` return in `addDisplayHandler` (both vanilla and CBA EHs fire).

addons/vehicles/functions/fnc_toggleVehicleFreelook.sqf
```diff
@@ -13,7 +14,6 @@
  *
  * Public: No
  */
-#include "script_component.hpp"
 
 #include "\a3\ui_f\hpp\defineResincl.inc"
```
dedmen: That one also needs to be moved to the top.

Dystopian:
Implemented as standalone addon in [v7 ace additions](https://steamcommunity.com/sharedfiles/filedetails/?id=2025187921). Looks like it's feature for one man (me:) so closing.

PR #6109, opened 2018-02-01
Arsenal - Pass arsenal dispay to displayOpened event

PabstMirror:
Tag @Alganthe
This pretty much matches the args passed by the vanilla event:
```
[missionnamespace,"arsenalOpened",[_display,_toggleSpace]] call bis_fnc_callscriptedeventhandler;
```

e.g. we add extra buttons to allow exporting gear to the format we want


alganthe:
^ I was planning to imrpove all the EHs / add more in my next PR after stats, but this works too :D

PR #6124, opened 2018-02-06
CBA Settings Conversion - Fix movedToSQF

PabstMirror:
`_settingsConfig >> _settingName >> "movedToSQF"`

_settingsConfig was the missionConfig, not main config

### DIFF
--- a/addons/common/functions/fnc_cbaSettings.sqf
+++ b/addons/common/functions/fnc_cbaSettings.sqf
@@ -85,17 +85,18 @@ for "_index" from 0 to (_countOptions - 1) do {
     };
 };
 
-_settingsConfig = missionConfigFile >> "ACE_Settings";
-_countOptions = count _settingsConfig;
+private _missionSettingsConfig = missionConfigFile >> "ACE_Settings";
+_countOptions = count _missionSettingsConfig;
 TRACE_1("Reading settings from missionConfigFile",_countOptions);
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_cbaSettings.sqf
```diff
@@ -85,17 +85,18 @@ for "_index" from 0 to (_countOptions - 1) do {
...
+_countOptions = count _missionSettingsConfig;
 TRACE_1("Reading settings from missionConfigFile",_countOptions);
 for "_index" from 0 to (_countOptions - 1) do {
-    private _optionEntry = _settingsConfig select _index;
+    private _optionEntry = _missionSettingsConfig select _index;
     private _settingName = configName _optionEntry;
     if ((toLower _settingName) in GVAR(cbaSettings_forcedSettings)) then {
         WARNING_1("Setting [%1] - Already Forced - ignoring missionConfig",_varName);
     } else {
         if ((isNil _settingName) && {(getNumber (_settingsConfig >> _settingName >> "movedToSQF")) == 0}) then {
```
commy2: Also `_missionSettingsConfig`?

PabstMirror: this was the bug, 
while checking mission config settings we want to check `movedToSQF` on the configFile configs

PR #6128, opened 2018-02-06
Scopes - Use real zero for non-adjustable scopes

### LINKED ISSUE #6126: Simplified zeroing breaks pistols
**Arma 3 Version:** `1.80`
**CBA Version:** `3.5.0`
**ACE3 Version:** current master

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
Enabling simplified zeroing causes pistols to shoot high.

**Steps to reproduce:**
- Enable ace_scopes_simplifiedZeroing
- Shoot at a persons head - you will miss
- Shoot center mass - you will hit the head/neck area

- Editor

PabstMirror:
Fix #6126

Minor performance improvments in firedEH
Fix in getCurrentZeroRange, was using adjustment elev (defaulted to 300) when gun was non-scope-adjustable
Had major effect on pistols, but effected all guns

### DIFF
--- a/addons/scopes/functions/fnc_firedEH.sqf
+++ b/addons/scopes/functions/fnc_firedEH.sqf
@@ -24,11 +24,8 @@ private _weaponIndex = [_unit, currentWeapon _unit] call EFUNC(common,getWeaponI
 if (_weaponIndex < 0) exitWith {};
 
 private _adjustment = _unit getVariable [QGVAR(Adjustment), [[0, 0, 0], [0, 0, 0], [0, 0, 0]]];
-private _zeroing = +(_adjustment select _weaponIndex);
+private _zeroing = (_adjustment select _weaponIndex) vectorMultiply 0.05729578; // Convert zeroing from mils to degrees (value of MRAD_TO_DEG(1))
 TRACE_1("Adjusting With",_zeroing);
-
-// Convert zeroing from mils to degrees
-_zeroing = _zeroing vectorMultiply MRAD_TO_DEG(1);
 
 if (GVAR(correctZeroing) || GVAR(simplifiedZeroing)) then {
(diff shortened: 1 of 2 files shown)


commy2:
`currentZeroing`  may be broken for FFV slots.

addons/scopes/functions/fnc_firedEH.sqf
```diff
@@ -24,12 +24,9 @@ private _weaponIndex = [_unit, currentWeapon _unit] call EFUNC(common,getWeaponI
 if (_weaponIndex < 0) exitWith {};
 
 private _adjustment = _unit getVariable [QGVAR(Adjustment), [[0, 0, 0], [0, 0, 0], [0, 0, 0]]];
-private _zeroing = +(_adjustment select _weaponIndex);
+private _zeroing = (_adjustment select _weaponIndex) vectorMultiply 0.05729578; // Convert zeroing from mils to degrees (value of MRAD_TO_DEG(1))
```
ulteq: Would `(_adjustment select _weaponIndex) vectorMultiply MRAD_TO_DEG(1)` really be that much slower?

dedmen: No.

commy2:
Uhm, tested what happens if you use the hangun from a vehicle's ffv slot?

PabstMirror:
Non-simplified zeroing does the same
I guess the command might be broken from FFV (returns 0)
but it shouldn't really matter as fired does
```
    private _oldZeroRange = currentZeroing _unit;
    private _newZeroRange = [_unit] call FUNC(getCurrentZeroRange);
```
and compares the difference, so the diff should be 0

I think the only effect will be the loss of the zero correction, but scopes should still work correctly

PR #6144, opened 2018-02-11
Interaction - Add actions based on animations

Dystopian:
- add animation based actions framework;
- add `Take off` action to:
  - backpacks on Offroad and some armored vehicles (since 1.82);
  - spare wheel on Wrangler and Van;
  - spare wheel on RHS AFRF cars (UAZ, Ural, KamAZ, Tigr, GAZ-66, ZIL-131) and BTRs;
  - spare wheel on RHS USF trucks (FMTV and HEMMT);
  - spare tracks on Kamysh and Tigris;
  - fuel canisters on Marid and arty;
  - entrench tool on some armored vehicles (since 1.82);
- disable RHS AFRF vehicles Retread system ~(except BTRs which is not so simple)~.

Some vehicles have animations which could be useful for in-game interaction. At first I thought to add all possible animations like remove doors and some vehicle parts. But then I decided to stop on really useful animations which can give you spare wheel e.g.

Animation based actions Framework allows to use any vehicle animations, which can give some items or not. I tried to make it universal enough but not too overloaded with all possible parameters. It can be easily used in 3rd party mods because it's based on configs and not on code.

Any ideas about other useful animations are welcomed.

If this PR is merged I'll add documentation in another PR.

Also at least CUP has Datsun with animated spare wheel.

Testing missions with modified vehicles:
[ace_interaction_anims.VR.zip](https://github.com/acemod/ACE3/files/1936127/ace_interaction_anims.VR.zip)
[ace_interaction_anims_RHS.VR.zip](https://github.com/acemod/ACE3/files/1944541/ace_interaction_anims_RHS.VR.zip)


jonpas:
Neat!

I think some more API allowing disabling of such actions (a setting maybe) would be good.

Dystopian:
> I think some more API allowing disabling of such actions (a setting maybe) would be good.

You want settings for each feature :) User can easily disable such actions with switching these anims to disabled state (even in editor via Vehicle Customization). So I don't think another setting is really needed.

> Additionally also setting items[] array via SQF in addition to config, so mission makers can set different types of eg. backpacks on offroads.

Can't imagine any real use cases for that. Only some vehicles have working anims and only few of them are really useful. Please don't think I try to avoid additional coding, I just need to know if it worth it.
I didn't add any user API exactly because of that.

> Those animated parts, aren't they just proxies? In that case it should be possible to get their location programmatically maybe?

I don't understand about proxies. Some anims have config values `source = "Proxy"` but I don't know how to use this info. Some vehicles have named selections in appropriate places (Wrangler) and it's much easier; some have them too dislodged (Van) and some don't have any (Offroad) - so you have to use absolute position.

jonpas:
Toggling one setting is a lot faster than disabling animations on all the vehicles.

API is probably not worth it, but I was already seeing possibilities for future, ability to put those items back on trucks and so on. :D

Dystopian:
1.82 brought some new animations so I added more backpacks and fuel canisters, spare tracks, entrench tool.
Also now backpacks and items go to WeaponHolder not to own vehicle.

Still not sure about hybrid config entries (which can be string or array) - it's not so good for mods.

Testing mission with all modified vehicles included.
[ace_interaction_anims.VR.zip](https://github.com/acemod/ACE3/files/1936127/ace_interaction_anims.VR.zip)

addons/interaction/functions/fnc_switchAnimActions.sqf
```diff
@@ -0,0 +1,141 @@
...
+                private ["_weaponHolder"];
+                {
+                    switch true do {
+                        case (1 == getNumber (configFile >> "CfgVehicles" >> _x >> "isBackpack")): {
+                            if (isNil "_weaponHolder") then {
+                                _weaponHolder = createVehicle ["GroundWeaponHolder", _emptyPosAGL, [], 0, "CAN_COLLIDE"];
+                            };
+                            _weaponHolder addBackpackCargoGlobal [_x, 1];
+                        };
+                        case (getNumber (configfile >> "CfgWeapons" >> _x >> "type") in [4096,131072]): {
```
Dystopian: reminder: should be macro #6487

Dystopian:
Is it a good idea to move anim configs to subfolder because they are independent enough (at least in `interaction` module)?

addons/interaction/functions/fnc_switchAnimActions.sqf
```diff
@@ -81,61 +81,60 @@ private _condition = {
...
+        if (
+            !isClass _animConfigParent
+            || {_x != _animConfigParent && {0 == [_animConfigParent >> "inherit", "number", 1] call CBA_fnc_getConfigEntry}}
+        ) then {
+            private _positions = [_x >> "position"] call BIS_fnc_getCfgData;
+            if (!isNil "_positions") then {
+                if (_positions isEqualType "") then {
                     _positions = [compile format [
-                        QUOTE(_target selectionPosition getText (configFile >> 'CfgVehicles' >> typeOf _target >> QQGVAR(anims) >> '%1' >> 'selection')),
+                        QUOTE(call compile getText (configFile >> 'CfgVehicles' >> typeOf _target >> QQGVAR(anims) >> '%1' >> 'position')),
```
dedmen: Is this really needed? there is a compile in L105. Can't we just do
call {+text+} instead?

dedmen: Also why call compile here but not in L116?

Dystopian: L105 is for `position = "[1,1.4,0]"` case. It produces this position code:
`{call compile getText (configFile >> 'CfgVehicles' >> typeOf _target >> "ace_interaction_anims" >> 'hide_spare' >> 'position')}`
so any child class with different `position` value will use exactly that value and not from parent class.

L116 is for `selection = "spare"` case. It produces this position code:
`{_target selectionPosition getText (configFile >> 'CfgVehicles' >> typeOf _target >> "ace_interaction_anims" >> 'spare' >> 'selection')}`
so child class with different selection position will use its own value and not position from parent class.

dedmen: >L105 is for position = "[1,1.4,0]" case.

parseSimpleArray.
`call compile` is almost always a mistake and just bad code. Only valid reason would be if position is in-config code string that's intended to be evaluated.

Dystopian: it was implemented this way to reflect `position` in [ace interaction config ](https://ace3mod.com/wiki/framework/interactionMenu-framework.html#2-adding-actions-via-config) and is really intended to be evaluated. `parseSimpleArray` could be used here but it will restrict only `[1,2,3]`-like values.

Dystopian:
Found CUP anims: Datsun, RG31, UAZ (spare wheel); Land Rover (spare wheels, jerrycans, Entrenching Tool, wire cutter (what?)). Can't start implement them if this PR is not merged.

addons/interaction/functions/fnc_switchAnimActions.sqf
jonpas: Yeah, interaction menu can take code to evaluate, I think this is fine.

jonpas:
> Still not sure about hybrid config entries (which can be string or array) - it's not so good for mods.

Make it just array, it can be array with one item.

> Is it a good idea to move anim configs to subfolder because they are independent enough (at least in `interaction` module)?

Could do that but you will quickly run into issues with base classes, eg. where multiple files reference same base class, it just goes all over the place.

Dystopian:
- all config entries are arrays now and are easy to use, no more BIS functions used;
- setting can not be enabled without mission restart but init is replaced from config loop to vehicle init events;
- no more those strange `call compile getText` lines, they are not so necessary. @dedmen will be satisfied :)
- config entries checks moved from `condition` to init function - faster;
- some code was simplified and now it's easier to read;
- RHS HEMTTs were fixed.

Dystopian:
Someone please check RHS AFRF class `rhs_kamaz5350` parent, looks like there was an error in #8308.

johnb432 (approved):
LGTM, but given I've redone some of the code, having another person take a look would be good.

PabstMirror (approved):

johnb432:
Sorry to add this many changes, but I felt they were necessary:
- Added option to give the progress bar a title
- Used better names & progress bar titles than the default
- Added some missing interactions (`rhs_kraz255b1_base`'s spare wheel, 2x base game HEMTTs' spare wheel)
- Restructured compats, so that interactions load only if the dependencies are there
- Improved interaction positions on turrets (previously some of them were static and didn't follow the turret, now they do)

addons/compat_rhs_afrf3/compat_rhs_afrf3_repair/CfgEventHandlers.hpp
```diff
@@ -0,0 +1,5 @@
+class Extended_PreInit_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_SCRIPT(XEH_preInit));
```
PabstMirror, line 3: path macros and `addon` won't work because it's a subaddon
I think the easiest fix is to just keep the code behind the `isModLoaded` check in upper

PabstMirror, line 3: This might also work? 🤢 

```suggestion
    class SUBADDON {
        init = QUOTE(call COMPILE_SCRIPT(SUBADDON\XEH_preInit));
```

johnb432: Nah, I'll revert the change.

johnb432: Reverted.

PR #6152, opened 2018-02-17
Interaction - Add Smash Windshield action

Dystopian:
When windshield is partly damaged it becomes less clear so you may want to break it entirely.

Strings are copied from ACE2, I don't know if they are correct.

### DIFF
--- a/addons/interaction/CfgVehicles.hpp
+++ b/addons/interaction/CfgVehicles.hpp
@@ -365,6 +365,11 @@ class CfgVehicles {
                 statement = "";
                 insertChildren = QUOTE(_this call DFUNC(addPassengersActions));
             };
+            class GVAR(smashWindshield) {
+                displayName = CSTRING(SmashWindshield);
+                condition = QUOTE(_player == driver _target && {private _damage = _target getHitPointDamage 'HitGlass1'; _damage > 0.5 && {_damage < 1}});
+                statement = QUOTE(playSound3D [ARR_2('A3\Sounds_F\weapons\hits\glass_2.wss',_target)]; _target setHitPointDamage [ARR_2('HitGlass1',1)];);
+            };
         };
     };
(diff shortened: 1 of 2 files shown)


jonpas:
Do all vehicles have `HitGlass1`? Are none just `HitGlass`? What about vehicles that have multiple of them `HitGlassX`?

I am not sure what to do about last, I guess you could get the position from it using `selectionPosition` with `HitPoints` LOD. Then make those actions added to root model (like repair/wheel actions are). That would solve all problems.

As for hit-point names, you'd solve that by parsing them in string form, just compare `HitGlass`, I am pretty sure everyone uses that standard so that would be pretty safe to go with.

PabstMirror:
Check locality issues on `setHitPointDamage`

Dystopian:
> Do all vehicles have HitGlass1? Are none just HitGlass? What about vehicles that have multiple of them HitGlassX?

At least vanilla, RHS, CUP. I can't check all of them, if there are issues we'll fix them. It looks like `HitGlass1` is almost always driver windshield.
Moreover `getHitPointDamage` returns 0 if there is no such hitpoint (unlike `getHit`), so it's safe to check it in Quad Bike or Kart.
BTW this worked in ACE2 and I don't remember any issues with it.

> I am not sure what to do about last, I guess you could get the position from it using selectionPosition with HitPoints LOD. Then make those actions added to root model (like repair/wheel actions are). That would solve all problems.

This action is `SelfAction` and intended to work inside only.

> Check locality issues on setHitPointDamage

It's driver-only action, so vehicle is always local.

PabstMirror:
possible issue with this is a very heavily armored glass window that wouldn't be remove-able
(possibly checking `armor` of the hitpoint?)
but I think it's ok for now and we can tweak later if there are problems

PR #6154, opened 2018-02-19
Weather - Add a 'check air temperature' action

LorenLuke:
Because you should feel generally how warm/cold it is without a Kestrel.

Add a 'Check Air Temperature' action to the self-interaction menu, showing roughly what the air temperature is.


addons/advanced_ballistics/CfgVehicles.hpp
```diff
@@ -1,4 +1,16 @@
 class CfgVehicles {
+    class Man;
+    class CAManBase: Man {
+        class ACE_SelfActions {
+            class ACE_CheckAirTemperature {
+                displayName = CSTRING(CheckTemperature);
+                condition = QUOTE( GVAR(enabled));
```
jonpas: No space before `GVAR`.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
+/*
+ * Author: LorenLuke
+ *
```
jonpas: No empty line.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+ * Returns an approximate representation of temperature graphically;
```
jonpas: End with `.` instead of `;`.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * Arguments:
+ * 0: Unit- Man <OBJECT>
```
jonpas: `* 0: Unit <OBJECT>`

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+ *
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [ACE_player] call ace_advanced_ballistics_fnc_checkAirTemp;
```
jonpas: No final `;`.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
+/*
+ * Author: LorenLuke
+ *
+ * Returns an approximate representation of temperature graphically;
+ *
+ * Arguments:
+ * 0: Unit- Man <OBJECT>
+ *
+ *
```
jonpas: Only one empty line.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+
+/*
+// Windchill calc with wind vars; https://en.wikipedia.org/wiki/Wind_chill#Australian_Apparent_Temperature
+private _windspeed = [getPosASL _unit, true, true, true] call EFUNC(weather,calculateWindSpeed);
+private _water_vapor_pressure = _humidity * 6.105 *(_e ^ ((17.27 * _temperature)/(237.7 + _temperature))) / 100;
+private _chill = _temperature + (0.33 * _water_vapor_pressure) - (0.70 * _windspeed) - 400;
+*/
+
+// sigmoid = f(x) = L / (1 + e^(-k * (x - x_0)));
+// L max value, e = euler, k = steepness, x_0 = midpoint
```
jonpas: Just write them out in above formula.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+
+// L = 1;
```
jonpas: Put those comments next to lines that actually do that.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+    if (( _i <= 0 && _i >= _apparent_temperature && _apparent_temperature <= 0) || ( _i >= 0 &&_i <= _apparent_temperature && _apparent_temperature >= 0) || abs(_i - _apparent_temperature) < 7) then {
+        _temp_color = [
+        //Colors obtained by quartic regression formula of RGB values at corresponding temperatures as marked on advanced_ballstics rangecard;
+            ( (((-0.000238348 * (_i ^ 4)) - (0.00865642 * (_i ^ 3)) + (0.794548 * (_i ^2)) - (0.323314 * _i) +6.78445) min 255) /255) max 0,
+            ( (((-0.000308515 * (_i ^ 4)) + (0.000245732 * (_i ^ 3)) + (0.308531 * (_i ^2)) - (02.78916 * _i) +164.326) min 255) /255) max 0,
+            ( (((-0.000107648 * (_i ^ 4)) + (0.00106365 * (_i ^ 3)) - (0.0989707 * (_i ^2)) - (9.66256 * _i) +171.402) min 255) /255) max 0
+        ];
+    };
+
+    if(_i == 0) then {
```
jonpas: Space after `if`.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+
+// Weighted average between windchill and heat index based on sigmoid levels, plus +- 4deg randomisation
+private _apparent_temperature = ((_temperature_sigmoid * _heatIndex) + ((1 - _temperature_sigmoid) * _chill) + random(8) - 4);
+
+private _string = "  -40C "; //extra spaces to balance zero point to be middle
+
+private _text = [_string, [0.1,0.4,1]] call EFUNC(common,stringToColoredText);
+private _color = [0.5, 0.5, 0.5];
+
+
```
jonpas: No need for double newline.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+for [{_i=-40}, {_i <= 40}, {_i = _i + 4}] do {
+
+    _temp_color = _color;
+    if (( _i <= 0 && _i >= _apparent_temperature && _apparent_temperature <= 0) || ( _i >= 0 &&_i <= _apparent_temperature && _apparent_temperature >= 0) || abs(_i - _apparent_temperature) < 7) then {
+        _temp_color = [
+        //Colors obtained by quartic regression formula of RGB values at corresponding temperatures as marked on advanced_ballstics rangecard;
```
jonpas: Space after `//`.
`advanced_ballistics` typo.
No final `;`.

addons/advanced_ballistics/stringtable.xml
```diff
@@ -275,5 +275,8 @@
             <Chinese>該模塊實現先進的彈道仿真 - 這意味著子彈的軌跡是由空氣溫度、大氣壓力、濕度、重力、彈藥類型以及射擊的武器所影響</Chinese>
             <Chinesesimp>该模块实现先进的弹道仿真 - 这意味着子弹的轨迹是由空气温度、大气压力、湿度、重力、弹药类型以及射击的武器所影响</Chinesesimp>
         </Key>
+        <Key ID="STR_ACE_Advanced_Ballistics_CheckTemperature">
```
jonpas: Add `Air` as well to distinguish fully.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+        _string = "|";
+    };
+
+    _text = composeText [_text, [_string, [_temp_color select 0, _temp_color select 1, _temp_color select 2]] call EFUNC(common,stringToColoredText)];
+
+};
+
+_text = composeText [_text, [" +40C", [1,0,0]] call EFUNC(common,stringToColoredText)];
+
+private _picture = QPATHTOEF(overheating,UI\temp_ca.paa);
```
jonpas: Just put into the call directly below.

addons/advanced_ballistics/functions/fnc_checkAirTemp.sqf
```diff
@@ -0,0 +1,92 @@
...
+// Weighted average between windchill and heat index based on sigmoid levels, plus +- 4deg randomisation
+private _apparent_temperature = ((_temperature_sigmoid * _heatIndex) + ((1 - _temperature_sigmoid) * _chill) + random(8) - 4);
+
+private _string = "  -40C "; //extra spaces to balance zero point to be middle
+
+private _text = [_string, [0.1,0.4,1]] call EFUNC(common,stringToColoredText);
+private _color = [0.5, 0.5, 0.5];
+
+
+for [{_i=-40}, {_i <= 40}, {_i = _i + 4}] do {
```
jonpas: `for "_i" from -40 to 40 step 4 do {` (indices might be off)

LorenLuke: I double checked the indices. They're working as intended with this formatting.

jonpas: Yes, but that format is a lot worse performance-wise than the one I provided.

LorenLuke:
@jonpas 
Review ready.

addons/advanced_ballistics/functions/fnc_displayAirTemp.sqf
```diff
@@ -0,0 +1,51 @@
...
+ * None
+ *
+ * Example:
+ * [ACE_player] call ace_advanced_ballistics_fnc_displayAirTemp
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_temperature", "_unit"];
```
PabstMirror: Should be `_apparent_temperature`, unit not needed, also change header

addons/advanced_ballistics/functions/fnc_displayAirTemp.sqf
```diff
@@ -0,0 +1,51 @@
...
+        _temp_color = [
+        // Colors obtained by quartic regression formula of RGB values at corresponding temperatures as marked on advanced_ballistics rangecard
+            ( (((-0.000238348 * (_i ^ 4)) - (0.00865642 * (_i ^ 3)) + (0.794548 * (_i ^2)) - (0.323314 * _i) +6.78445) min 255) /255) max 0,
+            ( (((-0.000308515 * (_i ^ 4)) + (0.000245732 * (_i ^ 3)) + (0.308531 * (_i ^2)) - (02.78916 * _i) +164.326) min 255) /255) max 0,
+            ( (((-0.000107648 * (_i ^ 4)) + (0.00106365 * (_i ^ 3)) - (0.0989707 * (_i ^2)) - (9.66256 * _i) +171.402) min 255) /255) max 0
+        ];
+    };
+
+    if (_i == 0) then {
+        _text = composeText [_text, ["[", [_temp_color select 0, _temp_color select 1, _temp_color select 2]] call EFUNC(common,stringToColoredText)];
```
PabstMirror: could just use `_temp_color` directly

addons/advanced_ballistics/CfgVehicles.hpp
```diff
@@ -1,4 +1,16 @@
 class CfgVehicles {
+    class Man;
+    class CAManBase: Man {
+        class ACE_SelfActions {
+            class ACE_CheckAirTemperature {
+                displayName = CSTRING(CheckAirTemperature);
+                condition = QUOTE(GVAR(enabled));
+                statement = QUOTE([_player] call FUNC(getApproximateAirTemp));
+                icon = QPATHTOEF(overheating,UI\temp_ca.paa);
```
PabstMirror: this does create a requirement between ab and overheating
might want to copy icon over to ab

BaerMitUmlaut:
How accurate is this? It should be a very very rough estimation. I'd maybe even go for just words like freezing, cold, chilly, warm, hot etc.

ulteq:
I haven't had time to look at the actual code yet, but I think this feature should be added to the weather module.

LorenLuke:
@BaerMitUmlaut 
@ulteq 

Temperature Gauge

LorenLuke:
Should I add a bias to each individual unit that uses this, just so that it's one more confounding variable (if one thinks this is too accurate), i.e. 'this unit always will think it's 3C colder than whatever they'd normally predict'?

SilentSpike:
Personally I don't think human thermoception is accurate enough to even consider guessing exact temperatures like that. I'd just have it split into ranges (-5 to 5, 5 to 15, 15 to 25, etc.) and then have strings to represent ("below freezing", "freezing", "cold", "mild", "warm", "hot", etc.)

SilentSpike:
Actually, even just show the temperature range being estimated by the character

Drofseh:
I like the idea of string based output that SilentSpike suggested.
Something like this:
```
>40         "extremely hot"
35 to 40    "exceptionally hot"
30 to 35    "very hot"
25 to 30    "hot"
20 to 25    "warm"
10 to 20    "mild"
5 to 10     "cool"
-5 to 5     "cold"
-15 to -5   "very cold"
-15 to -30  "exceptionally cold"
<-30         "extremely cold"
```
I don't know if you'd want that much granularity, although I am tempted to suggest another level of <-40 cold to the bottom, I've been in such cold air before and I could feel it freezing my nose from the inside, somewhere between -40 and -45.

bux:
It's a very very subjective scale.
Ask someone in Central Africa what he would define as hot and then ask someone from Finnland.

SilentSpike:
Yeah that's why I reconsidered and thought just showing the estimated range might be better, also more useful in this context anyway since you'd be guessing that in real life, rather than telling yourself "it's cold"

LorenLuke:
@SilentSpike 
I thought about doing it as a range where the color blob is a thing. I'll see how that looks.

Drofseh:
>It's a very very subjective scale.

true

bux:
feeled temperature is also influenced by rain/humidity and wind

LorenLuke:
Version 2.

Temperature Gauge

the0utsider:
This is amazing

Laid3acK:
Why not an arcadish "Check the Temp" but this is nothing to do in AB.

Drofseh:
Would it be possible to colour the thermometer to match the coloured range on the scale? 
ace_temp_01

ulteq:
Can we please move this into the weather module?

I would also either remove the thermometer, the numbers or the whole temperature scale. The visualization is currently too cluttered.

mharis001:
I agree, the UI definitely looks cluttered/busy and this should be moved to the weather module.

I think just a centered thermometer with the center being the current temperature colour would be fine with potentially having the entire colour scale drawn below it.

KarlMoebius:
I suggest for the text, instead of gradations like 'hot' and 'cold', have something more normal sounding, like "High 30's" or "Low 20's", like a weatherman would say for the forecast. You get a 5 degree granularity, and it sounds more natural without having to remember what temp range 'hot' is.

LorenLuke:
Version 3

Temperature Gauge

bux:
That’s definitely better 👍
Could need more vertical centering

bux:
> I do agree that the thermometer always being blue is a little odd because it seems like it's presenting conflicting information.

good point, maybe a white fill color would be better here

LorenLuke:
@SilentSpike 
@bux 

If you happen to know a way to separate the interior from the rest of the thermometer *and* a way to color it (or the whole thermometer) to match the temperature below, and recombine the icon into one, I'm all ears, but honestly, if it's viewed as more a 'standard' icon that's independent of the actual temperature, I'm all ears.

I tried putting in a white one previous but it appears my wizardry was no match as it kept showing the blue one.

Drofseh:
Doing the thermometer colour as a gradient work work I think. Better than solid blue or white.
Here's a quick example from photoshop.
More yellow and less green would probably be good if actually implemented.

ulteq:
> maybe a white fill color would be better here

That would have been my suggestion as well.

KarlMoebius:
A suggestion? Like subtitles, if there could be an option to outline the letters in a different color, like white or yellow, like other parts of ace? It would make reading the numbers against different or chaotic backgrounds (like inside vehicles, against trees, buildings, moving people) a little easier to read.

Laid3acK:
>A suggestion?

Should be an option in the Weather module.

KarlMoebius:
There is a part of me that thinks all font related chicanery (color, brightness, outline) should be collected under one umbrella instead of buried in multiple places. But if I'm going to search for the font adjustments for a specific thing, I'll start in it's home folder. So... I dunno? Both?

jonpas:
Some left-over files need to be moved into the `weather` module. Otherwise this looks good to be merged!

addons/weather/functions/fnc_displayAirTemp.sqf
```diff
@@ -0,0 +1,72 @@
...
+            ( (((-0.000107648 * (_i ^ 4)) + (0.00106365 * (_i ^ 3)) - (0.0989707 * (_i ^2)) - (9.66256 * _i) +171.402) min 255) /255) max 0
+        ];
+    };
+
+    _text = composeText [_text, [_string, [_temp_color select 0, _temp_color select 1, _temp_color select 2]] call EFUNC(common,stringToColoredText)];
+};
+
+_text = composeText [_text, [" +40C", [1,0,0]] call EFUNC(common,stringToColoredText)];
+*/
+[_text, QPATHTOF(UI\temp_ca.paa),[1,1,1], ACE_player, 2] call EFUNC(common,displayTextPicture);
```
jonpas:
```suggestion
[_text, QPATHTOF(UI\temp_ca.paa),[1,1,1], ACE_player, 2] call EFUNC(common,displayTextPicture);

```

PR #6159, opened 2018-02-20
Arsenal - Compile stats on first use

PabstMirror:
Tag @Alganthe

Compiles the stats arrays on first use in a mission
Prevents stats added during one mission from having persistent effects


alganthe:
Looks good.

> Prevents stats added during one mission from having persistent effects

That was intentional, however it also means you can abuse it to execute code as you pointed out.
Compiling on mission start also means you can't add stats in 3DEN (nothing past preInit runs there) but that's a lesser evil I can bear with.

PR #6162, opened 2018-02-22
Add cursor movement API

LorenLuke:
- Add extensions capable of mimicking player mouse input.
- Adds a function to 'common' to call the extension easily.

Notes:
- Cursor movement in loops persists in dialog and pause menu.
Be sure to have safeguards in place before using in loops.


extensions/cursor_movement/cursorMovement.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+void __stdcall RVExtension(char *output, int outputSize, const char *function) {
+    INPUT ip;
+    ip.type = INPUT_MOUSE;
+    ip.mi.mouseData = 0;
+    ip.mi.dwFlags = MOUSEEVENTF_MOVE;
+    ip.mi.time = 0;
+
+    int x, y;
+    if (strncmp(function, "x", 1)) {
```
dedmen: use the AB like format x:y so you can just send both.

addons/common/functions/fnc_moveCursor.sqf
```diff
@@ -0,0 +1,30 @@
...
+ */
+#include "script_component.hpp"
+
+params ["_x", "_y"];
+
+if (dialog) exitWith {};
+
+// "bluefcs" callExtension "x100"; "bluefcs" callExtension "y-100"
+// Yes, they're backwards. No, I can't fix it because I don't know how to DLL. -Luke
+call compile format ['"ace_cursor_movement" callExtension "y%1"',_x];
```
dedmen: call compile format? really?
`"ace_cursor_movement" callExtension format ["%1:%2",_x,_y]`

addons/common/functions/fnc_moveCursor.sqf
```diff
@@ -0,0 +1,30 @@
+/*
+ * Author: LorenLuke (function), BaerMitUmlaut (Extension)
+ * Simulates mouse input by number of pixels (not consistent/smooth)
```
dedmen: Optional animated movement? linear or even specific curves?

extensions/cursor_movement/cursorMovement.cpp
```diff
@@ -0,0 +1,34 @@
...
+        ip.mi.dx = atoi(&function[1]);
+        ip.mi.dy = 0;
+    }
+    else if (strncmp(function, "y", 1)) {
+        ip.mi.dy = atoi(&function[1]);
+        ip.mi.dx = 0;
+    }
+    else {
```
dedmen: All other extensions use `} else {` else on same line.

extensions/cursor_movement/cursorMovement.cpp
```diff
@@ -0,0 +1,34 @@
...
+        strncpy_s(output, outputSize, "error", _TRUNCATE);
+        return;
+    }
+
+    SendInput(1, &ip, sizeof(INPUT));
```
dedmen: Shouldn't it also return "error" if SendInput fails?

dedmen:
Could also call this `mouse` extension and use it to click. But that needs even better protection that it can't be missused.

addons/common/functions/fnc_moveCursor.sqf
```diff
@@ -0,0 +1,30 @@
...
+ * Note- using with open dialogs may cause issues when in loops.
```
dedmen: What kind of issues? Moving out of game window?

extensions/cursor_movement/cursorMovement.cpp
```diff
@@ -0,0 +1,34 @@
...
+    ip.mi.mouseData = 0;
+    ip.mi.dwFlags = MOUSEEVENTF_MOVE;
+    ip.mi.time = 0;
+
+    int x, y;
+    if (strncmp(function, "x", 1)) {
+        ip.mi.dx = atoi(&function[1]);
+        ip.mi.dy = 0;
+    }
+    else if (strncmp(function, "y", 1)) {
```
dedmen: strncmp for just one character is a little overkill..
`function[0] == 'y'`

LorenLuke:
@dedmen 

https://github.com/LorenLuke/ACE3/blob/191f8cf0534f8608749c2e9ce58980ea1a019009/addons/common/functions/fnc_moveCursor.sqf#L2
https://github.com/LorenLuke/ACE3/blob/191f8cf0534f8608749c2e9ce58980ea1a019009/addons/common/functions/fnc_moveCursor.sqf#L25

The second half of both those lines are probably the best response I can give to like, 90% of your review.

addons/common/functions/fnc_moveCursor.sqf
LorenLuke: ~April, April.~ Yes, really.
I do stuff the way I know how that it works. If there's a better way, tell instead of condescending for not knowing the better way.

addons/common/functions/fnc_moveCursor.sqf
LorenLuke: Rotating a foot-mobile is all in one go, but vehicles had some sort of acceleration and cap on how fast you can traverse, meaning it has to be inputting for more than one frame to do much of anything.

addons/common/functions/fnc_moveCursor.sqf
LorenLuke: And pinning your cursor to the corner of the desktop/monitor like it's a professional wrestler waiting for a three-count.

LorenLuke:
> not precise motion.

The intended use is to call it within a PFH. Check that the game isn't paused and a dialogue isn't open, then use a constant angle check in to tandem with a PID controller to direct a correspondence towards an orientation.

addons/common/functions/fnc_moveCursor.sqf
dedmen: > If there's a better way, tell instead of condescending for not knowing the better way.

Uhm...
>`"ace_cursor_movement" callExtension format ["%1:%2",_x,_y]`

And my other comment about the format in the extension?

addons/common/functions/fnc_moveCursor.sqf
dedmen: What is a  `foot-mobile`?

dedmen:
>then use a constant angle check in to tandem with a PID controller to direct a correspondence towards an orientation.

Ehhh.... What?

Cuel:
what's the purpose after merge? just wondering

addons/common/functions/fnc_moveCursor.sqf
jonpas: A person. Don't you play proper Arma, that's standard callout. :p

addons/common/functions/fnc_moveCursor.sqf
LorenLuke: _*I didn't write the extension. I don't know how to .DLL.*_

LorenLuke:
@dedmen 
> angle check in to tandem with a

Should have been 'in tandem with'. Also not proofread semi-pseudocode, because I'm on my phone.
```
Player setVariable ["autotrack", [0,0]];
onEachFrame { 
(player getVarable "autotrack") call func(move_cursor);
};
```
and just check angles between it and a desired aimpoint from there and feed them into the unit variable.

@Cuel I recall in the slack regarding <https://github.com/acemod/ACE3/pull/6034> that someone said they wouldn't mind if a vertical component could be added somehow. This could do just that. However original plans were for this to be used with a to-be-written Shtora defensive system framework, one of the components of which involves rotating the entire turret automatically to face an incoming rocket. With an AI or even nobody in the turret, that's doable; with a player in this, less so. Additional capabilities of this can include laying off some automatic artillery pieces and the creation of auto-tracking, radar-guided guns like the Shilka would be.

commy2:
Hunter-Killer for tanks.

Dystopian:
Some time ago I was thinking about systems like Shilka: maybe it could be done with temporary moving player from main gunner to some other vehicle seat and creating UAV-like AI at gunner seat. AI could be ordered not fire till player orders.

addons/common/functions/fnc_moveCursor.sqf
dedmen: And? Does that mean problems in your PR don't need to be fixed and can just be ignored?

addons/common/functions/fnc_moveCursor.sqf
dedmen: I intended this to mean adding animation handled in a second thread.
Which also solves:
>meaning it has to be inputting for more than one frame to do much of anything.

extensions/cursor_movement/cursorMovement.cpp
dedmen: Or maybe just use `callExtensionArgs`.

commy2:
State?

dedmen:
See all comments. Currently Stale it seems because author is unable to change stuff on the extension.
Unless you are just fine with ignoring all comments and just merging it.

PabstMirror:
I think we should wait until we have an actual use for this

LorenLuke:
@dedmen 
@commy2 
As someone of variable and limited free time (_why are there only **24** hours in a day?!_), fluctuating mental health, and biting off more than I can chew with many other projects (missile guidance rework, MANPADS realistic functionality, Javelin realistic functionality introduction, MiniSpike television guidance, UAV vanilla module replacement/improvement) this PR has indeed gone somewhat stale, but it's not gone from my memory.

@PabstMirror 
Uses for it would include commy's Hunter-Killer/slew-to-cue system, as well as auto tracking radar guided guns for the Tigris/Bardelas AA as well as 3rd party AA (ZSU Shilka), and Shtora active protection systems (intended to be compatible with the missile guidance rewrite).

PR #6183, opened 2018-03-05
Allow goggle effects to be set locally per client

DeliciousJaffa:
From what I can see goggles are all handled locally and do not need to be globally enabled/disabled. If a group wants to force the setting, CBA provides optional force functionality.

### DIFF
--- a/addons/goggles/ACE_Settings.hpp
+++ b/addons/goggles/ACE_Settings.hpp
@@ -5,6 +5,7 @@ class ACE_Settings {
         displayName = CSTRING(effects_displayName);
         typeName = "SCALAR";
         value = 2;
+        isClientSettable = 1;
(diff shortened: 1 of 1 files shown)


PabstMirror:
Right now the setting isn't designed to be changed mid-mission.
It will take some work to make it work.
https://github.com/DeliciousJaffa/ACE3/blob/b6a2315525886e49c9e066f0d2f2514e02114ec8/addons/goggles/XEH_postInit.sqf#L17-L156

DeliciousJaffa:
That is irrelevant to the purpose of this change, currently clients are forced to have the same setting as the server, this change will make that optional (can still be forced using "Overwrite Clients")

DeliciousJaffa:
To clarify, you can currently change the option mid-mission anyway as a server admin or in single player, all this change would do is allow clients to use their own settings by default for an entirely visual feature instead of being forced to use the goggle effects by default.

Currently, the only way to disable this as a client on a server with default settings is to delete the PBO

PR #6192, opened 2018-03-14
More various ace arsenal fixes

alganthe:
- Fix case issue in verifyLoadout
- Remove uneeded lines related to the perf profiler (doesn't need them anymore)
- Move that one line this PR was about where it should be;
- Fix the ace arsenal camera weird behavior underwater or around rocks, the camera will now clip through rocks or other objects to avoid those issues.
- Gunbags will keep their weapons if you switch to another backpack then back to a gunbag in ace arsenal.

### DIFF
--- a/addons/gunbag/XEH_preInit.sqf
+++ b/addons/gunbag/XEH_preInit.sqf
@@ -24,4 +24,22 @@ PREP_RECOMPILE_END;
     }, _this] call CBA_fnc_execNextFrame;
 }] call CBA_fnc_addClassEventHandler;
 
+[QEGVAR(arsenal,displayOpened), {
+
+    private _center = EGVAR(arsenal,center);
+
+    if (_center call FUNC(hasGunBag)) then {
+        GVAR(arsenalCache) = (backpackContainer _center) getVariable [QGVAR(gunbagWeapon), []];
+    };
+}] call CBA_fnc_addEventHandler;
+
+[QEGVAR(arsenal,displayClosed), {
+
+    if !(isNil QGVAR(arsenalCache)) then {
+        (backpackContainer EGVAR(arsenal,center)) setVariable [QGVAR(gunbagWeapon),GVAR(arsenalCache), true];
+    };
+
+    GVAR(arsenalCache) = nil;
+}] call CBA_fnc_addEventHandler;
+
 ADDON = true;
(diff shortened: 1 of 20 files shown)


PabstMirror:
I don't see why this is necessary,
compileStats already does `if (!isNil QGVAR(statsListLeftPanel)) exitWith {};`
so it should have been safe to call at anytime

alganthe:
Nearly forgot about this, since it ended up being a single line I added a bug fix with it :D

PabstMirror:
`findIf`?

alganthe:
new command added in dev branch: https://community.bistudio.com/wiki/findIf

jonpas:
Is it on RC? No merge if it isn't.

alganthe:
it was added before the RC, I didn't check if it was on RC and assumed it was, good point.
I'll check.

alganthe:
Just checked, I can confirm findIf is present on the current RC build (1.82.144616)

dedmen:
why use `findIf` if it's slower than `in` and everything is already built to be lowercase and work fine with `in`?

alganthe:
because, like I said, findIf is case insensitive while in is case sensitive.

alganthe:
There, included all the fixes for the issues i'm aware off.

addons/gunbag/XEH_preInit.sqf
```diff
@@ -24,4 +24,22 @@ PREP_RECOMPILE_END;
     }, _this] call CBA_fnc_execNextFrame;
 }] call CBA_fnc_addClassEventHandler;
 
+[QEGVAR(arsenal,displayOpened), {
+
+    private _center = EGVAR(arsenal,center);
+
+    if (_center call FUNC(hasGunBag)) then {
+        GVAR(arsenalCache) = (backpackContainer _center) getVariable [QGVAR(gunbagWeapon), []];
```
dedmen: Maybe use a more descriptive varname. As this always stores the gunbag. Why not gunbagCache?
You might want to use arsenalCache for something like an actual arsenalCache later on.

alganthe: the full var is:
ace_gunbag_arsenalCache, I don't really know if there's a more descriptive varname :/

dedmen: ooooohhhh. duh. Should've looked at what file that is. Expected it to be in arsenal

PR #6193, opened 2018-03-14
Add No Default Action Icon optional

jonpas:
- Remove icon in the middle of the screen when player is looking at something intractable with vanilla action menu
- Optional to mimic `noactionmenu`
- Integrated from [Gruppe Adler Mod](https://github.com/gruppe-adler/gruppe_adler_mod/tree/master/optionals/noDefaultActionIcon)
- [x] Add notice of integration (attribute original authors)
- [x] Name `No Default Action Icon`?
- [x] Documentation


Drofseh:
Does this allow whatever the current default action is to still be performed?

I could see it being a little confusing if I accidentally performed an action that I wasn't expecting due to their being no icon to tell me about it.

jonpas:
> Does this allow whatever the current default action is to still be performed?

I think so, I literally just ported it so far, didn't even run the game with it yet. Anyways, that's why it's an optional, but I'll definitely update the description when I am sure, no worries. Should probably tag this as WIP until then.

Laid3acK:
Personally, i play "No icon" since ArmA1,
with ArmA3, i recommend to check the `Use default action` keys in the "Common" CONTROLS option
and, eventually, deleted it if necessary.
Without this precaution, a current default action can be performed accidentally.

jonpas:
Disabling that key won't hide the icon though. You mean you ve been using something like this PR since then?

commy2:
Like the idea, but not the current name.

Laid3acK:
>You mean you ve been using something like this PR since then?

Since the beginning off ArmA3 for my personal use.
It was also a part off the [UIcorrections Suite](https://forums.bohemia.net/forums/topic/168353-uicorrections-suite/?page=2&tab=comments#comment-2656185) by **beazley** (2014) and a tiny [mod](https://forums.bohemia.net/forums/topic/190020-levelite-mini-clean-hud-mod/?tab=comments#comment-3010787) i made after (2016).

But it's not a problem.
I just wanted to draw your attention to those keys assigned to the default action.

jonpas:
Ah, that key, already deleted on my side, haha. But cool, good note, I'll even add it to doc. :) 

@commy2 got a better name? It was "no default action icon" in the gruppe adler mod. "no action menu" is not exactly good or descriptive either.

commy2:
nodefaultactionicon would be more descriptive. And a different component having a bad name doesn't mean this one should have a bad one too.

jonpas:
> nodefaultactionicon would be more descriptive.

Can agree. I just wrote "nodefaultaction" because I wasn't sure if it just disables the icon or not at the time.

> And a different component having a bad name doesn't mean this one should have a bad one too.

I was just kidding about that. :P

jonpas:
Done, would like to hear from @Fusselwurm if he is fine with the way I attributed him/them.

optionals/nodefaultactionicon/CfgInGameUI.hpp
```diff
@@ -0,0 +1,5 @@
+class CfgInGameUI {
+    class DefaultAction {
+        size = "0";
```
PabstMirror: this could be something like
`size = "profileNamespace getVariable ['actionSizeSetting', 0.025];";`
but it seems to require full game restart to take effect

commy2: Yep, this is something that could be a client / local only setting (not forcable). It would require a game restart, but that is still a million times better than optional pbos.
It may need a dummy entry, because numbers may not be read correctly if the expression contains a comma. This needs more research though and might not apply to all config entries.
The helper technique I posted on Slack once.

jonpas: What an old thing revived! :+1: 

Would definitely be great, going to leave it to @commy2 or someone else to test though, I probably won't have time that soon.

Schwaggot: FYI, the changes in CfgInGameUI break BIS_fnc_holdActionAdd (the hold action introduced by APEX), the symbol is not shown.

commy2: You mean this pull request or something else in ACE?

Schwaggot: The code referenced above
```class CfgInGameUI {
    class DefaultAction {
        size = "0";```

**Comment by commy2** [comment]
- [ ] BIS_fnc_holdAction apparently uses this default size.

Maybe it's fine since this is optional and anyone who would ever use it is unlikely to have a scipted hold action in their mission...

File: docs/wiki/feature/nodefaultactionicon.md [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+  major: 3
+  minor: 13
+  patch: 0
+---
+
+## 1. Overview
+
+"Default action" icon shows up automatically in Arma 3 when something in front of the cursor is interactable, disturbing the player. This component prevents that icon from ever showing.
+
+It does **NOT** remove the availability or activation of that action. It is advisable to also unbind "Use default action" key to prevent unwanted interactions.
```
jonpas: Must also add: Makes `BIS_fnc_holdActionAdd` actions invisible.

jonpas:
Replaced with #8883.

PR #6201, opened 2018-03-18
Arma 1.82 compat

PabstMirror:
Based on current RC


jonpas:
Require next CBA with 1.82 compatibility as well?

PR #6203, opened 2018-03-19
Add possibility to create zeus during mission

Dystopian:
- add `Create Zeus` and `Delete Zeus` self-interaction menus, controlled with setting.

Player gets his own zeus object with all addons activated.

Still thinking about `Debug console` option at setting (means menu available if console is available), if it's needed at all and if its translation is OK next to `None`, `Available for the host or logged-in admin` and `Available for everyone`.

This feature is more for debugging and tests, not for production use.


jonpas:
Why not just add a button to pause menu, next to debug console?

Dystopian:
will check

Dystopian:
- there is no really good place for such button;
- too much work for such simple feature.

PabstMirror:
>This feature is more for debugging and tests, not for production use.
It my experience, when creating a zeus module it causes a noticeable stutter on all machines
not that bad for most, but could cause issues for pilots

Dystopian:
yeah, I have those lags too. This feature really should be used with care.

SilentSpike:
Not sure how I feel about this one. It seems kind of out of place in ace to me.

Dystopian:
I can move it to optionals if you want.

Dystopian:
Also init code can be moved to setting change script, say if it's needed.

shukari:
Cool feature! Is this still in dev or abandoned?
Would be really cool if admin and user with debugconsole access have a button in the esc menu.

Dystopian:
I deleted workaround for missed 3rd party vehicles in zeus because I can't reproduce it anymore. If anyone test the PR, check if vehicles are not missed in vehicles tree.

Dystopian:
Workaround is still necessary. I just load game (without filepatching) and load editor preview.

addons/zeus/XEH_postInit.sqf
```diff
@@ -39,4 +41,75 @@ if (isServer) then {
...
+        _zeus setVariable ["Addons", 3, true];
+        _zeus setCuratorCoef ["Place", 0];
+        _zeus setCuratorCoef ["Delete", 0];
+        _group deleteGroupWhenEmpty true;
+        if (!isMultiplayer && {!isNull findDisplay IDD_DISPLAY3DEN}) then {
+            // if loaded from editor (but not after restart), addons are not activated so we do it manually
+            private _addons = ('true' configClasses (configFile >> "CfgPatches")) apply {configName _x};
+            activateAddons _addons;
+            removeAllCuratorAddons _zeus;
+            _zeus addCuratorAddons _addons;
```
dedmen: I think adding a huge amount of addons causes a major lag on all clients. I never actually tried if the lag spike is gone if I remove the addons but it is my biggest suspect.
Should test that out and if it's really this the addons should be spread over a longer time. Many small lags over 10 seconds are better than half a second game freeze.

Dystopian: If you talk about `addCuratorAddons` in multiplayer then ^^ `if (!isMultiplayer`.
If about `_zeus setVariable ["Addons", 3, true]` then nothing we can do and it's mentioned in comments.

dedmen: First one yeah. I always do `addCuratorAddons` in my curator script. Didn't know it wasn't needed.

jonpas:
I'd still prefer this in pause menu. Why add more clutter to self-interaction? I know it's only for admins, but I as an admin don't want it there.

https://github.com/CBATeam/CBA_A3/blob/master/addons/ui/fnc_addPauseMenuOption.sqf this function wasn't the case back in March, would this make it easier to add for you @Dystopian ?

Dystopian:
`addPauseMenuOption` requires dialog in params so I don't think it suits.

We use this feature for some months and it never was clutter in admin interaction menu. But it was VERY useful for debugging or testing something all the time.

I really don't mind if someone implements it in Esc menu any way, easy or complicated.

jonpas:
:+1:

Issue #6163, opened 2018-02-22
Advanced Fatigue - Increased performance factor reduces stamina recovery rate.


Coryf88:
**Arma 3 Version:** `1.80` (stable)
**CBA Version:** `3.6.1` (stable)
**ACE3 Version:** `3.12.0` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
Increasing performance factor reduces stamina recovery rate.

Performance factor increases the rate at which [`anFatigue`](https://github.com/acemod/ACE3/blob/master/addons/advanced_fatigue/functions/fnc_mainLoop.sqf#L56) recovers, however `anReserve` uses [`anFatigue ^ 2`](https://github.com/acemod/ACE3/blob/master/addons/advanced_fatigue/functions/fnc_mainLoop.sqf#L64) as a multiplier. So, a higher performance factor also reduces the `anReserve` recovery rate.

**Steps to reproduce:**
1. Take note of the amount of time it takes to recover stamina.
2. Increase performance factor, repeat step 1.
3. Observe increased stamina recovery time with a higher performance factor.

**Where did the issue occur?**
Dedicated / Self-Hosted Multiplayer / Editor (Singleplayer) / Editor (Multiplayer)

**Placed Modules:**
None

**RPT log file:**
N/A

BaerMitUmlaut:
According to the original documentation I think everything is correct here, though they mention `(0.057 / GVAR(peakPower))` being a "constant". I'm wondering if it is actually a constant in the sense that it doesn't adjust to the peakPower, which results in the expected behaviour. I'll patch it like that for the next release.

ulteq:
> Performance factor increases the rate at which [`anFatigue`](https://github.com/acemod/ACE3/blob/master/addons/advanced_fatigue/functions/fnc_mainLoop.sqf#L56) recovers

No. The performance factor decreases the rate at which `anFatique` increases.

NyteMyre:
Sorry for kicking up this 2 year old ticket, but last weekend I hosted a mission where i wanted to increase the stamina/recovery for the players since they had a lot of walking to do. So i set the following in ACE Advanced Fatigue options:

![rd2nzh8yxrl61](https://user-images.githubusercontent.com/72631377/110325962-51825b00-8018-11eb-896b-a61d284a8217.png)

In the debrief, people complained about having worse stamina than usual and that recovery seemed to take longer than before

SekiAlpha:
[...]

PR #6204, opened 2018-03-19
Fix performance factor breaking stamina recovery

BaerMitUmlaut:
- Fix performance factor breaking stamina recovery

From the original VBS documentation:

---

**Acidosis accumulation**
`anaerobicFatigue += anaerobicDwork * maxPowerFatigueRatio * 1.1;`
[...]
`maxPowerFatigueRatio` is a constant designed to scale the accumulation and recovery of anaerobicFatigue: `maxPowerFatigueRatio = 0.057 / peakPower;`

---

They might mean that it's a constant regardless of the actual peakPower, which is calculated from the VO₂max (which in turn is adjusted by the performance factor). This seems to fix issue #6163.

### DIFF
--- a/addons/advanced_fatigue/functions/fnc_mainLoop.sqf
+++ b/addons/advanced_fatigue/functions/fnc_mainLoop.sqf
@@ -53,7 +53,9 @@ GVAR(ae1Reserve) = GVAR(ae1Reserve) - _ae1Power / WATTSPERATP;
 GVAR(ae2Reserve) = GVAR(ae2Reserve) - _ae2Power / WATTSPERATP;
 GVAR(anReserve)  = GVAR(anReserve)  -  _anPower / WATTSPERATP;
 // Increase anearobic fatigue
-GVAR(anFatigue)  = GVAR(anFatigue) + _anPower * (0.057 / GVAR(peakPower)) * 1.1;
+// 1264.64 = 0.057 / peakPower with default VO2 max
+// See #6163
+GVAR(anFatigue)  = GVAR(anFatigue) + _anPower * 1264.64 * 1.1;
 
 // Aerobic ATP reserve recovery
 GVAR(ae1Reserve) = ((GVAR(ae1Reserve) + OXYGEN * 6.60 * (GVAR(ae1PathwayPower) - _ae1Power) / GVAR(ae1PathwayPower) * GVAR(recoveryFactor)) min AE1_MAXRESERVE) max 0;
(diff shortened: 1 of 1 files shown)


BaerMitUmlaut:
I don't think it's how it's meant either, this is more of a hotfix tbh. If you have an idea where else the problem might be let me know.

ulteq:
I think part of the problem is that we derive `_perceivedFatigue` off `anReserve`. Which is not how it is meant to be done.

Instead we should use `anFatique` directly: https://github.com/acemod/ACE3/pull/5723/files#diff-0bdf0f3e6417972d88363532f5354a5aR114

I'll try to do some tests tomorrow.

BaerMitUmlaut:
That's definitely not the underlying issue. The stamina bar displays only the anaerobic reserve which very visibly shows that it does essentially not recover at all. `_perceivedFatigue` is only used for feedback effects and has no influence on the recovery.

ulteq:
> That's definitely not the underlying issue.

Right.

New attempt:

What if the performance factor would also modify `ANTPERCENT`.

addons/advanced_fatigue/functions/fnc_mainLoop.sqf
```diff
@@ -53,7 +53,9 @@ GVAR(ae1Reserve) = GVAR(ae1Reserve) - _ae1Power / WATTSPERATP;
 GVAR(ae2Reserve) = GVAR(ae2Reserve) - _ae2Power / WATTSPERATP;
 GVAR(anReserve)  = GVAR(anReserve)  -  _anPower / WATTSPERATP;
 // Increase anearobic fatigue
-GVAR(anFatigue)  = GVAR(anFatigue) + _anPower * (0.057 / GVAR(peakPower)) * 1.1;
+// 1264.64 = 0.057 / peakPower with default VO2 max
+// See #6163
+GVAR(anFatigue)  = GVAR(anFatigue) + _anPower * 1264.64 * 1.1;
```
ulteq: `1264.64` seems to be very high. Where did you get peakPower `0.00004507211` from?

PabstMirror:
Just to clarify my changes `GVAR(peakPower) = 1264` under default values.

But I'm not sure if this is correct, line 58 increases fatigue and line 69 handles recovery.
Before both lines used the term `(0.057 / GVAR(peakPower)` to scale; this goes down as performance goes up.
This means that you had more reserves, but it also took longer to recover.

With this PR we change fatigue increase to ignore peakPower, but recovery still is negatively effected by it.

ulteq:
I'm still not convinced by this PR. But I sadly don't have time to take a closer look.

One solution might be to finish this PR: https://github.com/acemod/ACE3/pull/5723

It should be almost ready to go if we leave the [Experimental movement speed limiter](https://github.com/acemod/ACE3/pull/5723/commits/18661ebe02014d6f2c91350e341cf01eeddc8f0e) out.

Mike-MF:

PR #6207, opened 2018-03-22
Repair - Fix Typhoon rear wheel change

### LINKED ISSUE #6196: change rear tire of typhoon is impossible
**Arma 3 Version:1.80 `` (stable)
**CBA Version:** `3.6.1` (stable)
**ACE3 Version:** `3.12.1` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- chaging rear tire of the typhoon (truck) is not possible.

**Steps to reproduce:**
- place typhoon on map, destroy the rear tire (3rd axis) and try to change the tire.

- Dedicated

**Placed Modules:**
- None.

Dystopian:
- fix `Truck_03` double wheel hitpoint;
- fix double selection check code in Repair conponent;
- close #6196.

Typhoon (`Truck_03_base_F`) has 6 configured wheels (`1_1`, `1_2`, `1_3`, `2_1`, `2_2`, `2_3`) but 8 hitpoints:
`["HitLFWheel","HitLF2Wheel","HitLMWheel","HitLBWheel","HitRFWheel","HitRF2Wheel","HitRMWheel","HitRBWheel"]`
and corresponding selections: `["wheel_1_1_steering","wheel_1_2_steering","wheel_1_3_steering","wheel_1_3_steering","wheel_2_1_steering","wheel_2_2_steering","wheel_2_3_steering","wheel_2_3_steering"]`
Note the same `1_3` and `2_3` selections.

BI added this strange config overwriting original hitpoint selection name:
```
class HitLBWheel: HitLBWheel {
    name = "wheel_1_3_steering";
};
```
Typhoon is the only vanilla vehicle with such config. If someone wants to report this issue - welcome.

It's strange but CUP has some vehicles with similar double wheel hitpoints. I'll report it.


bux:
Could you copy the link to the issue/s in here? ... for reference

addons/repair/CfgVehicles.hpp
```diff
@@ -480,7 +482,22 @@ class CfgVehicles {
...
-    class Truck_03_base_F;
+    class Truck_F: Car_F {
+        class HitPoints: HitPoints {
+            class HitLBWheel;
+            class HitRBWheel;
+        };
+    };
+    class Truck_03_base_F: Truck_F {
+        class HitPoints: HitPoints {
+            class HitLBWheel: HitLBWheel {
```
commy2: `bin\config.bin/CfgVehicles/Car_F/HitPoints/HitLBWheel` NOT Truck_F HitLBWheel!

addons/repair/CfgVehicles.hpp
```diff
@@ -480,7 +482,22 @@ class CfgVehicles {
         GVAR(canRepair) = 0;
     };
 
-    class Truck_03_base_F;
+    class Truck_F: Car_F {
+        class HitPoints: HitPoints {
+            class HitLBWheel;
+            class HitRBWheel;
```
commy2: These two don't exist in the base game.

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -66,6 +66,8 @@ private _icon = QPATHTOF(ui\repair_0_ca.paa);
         _statement = {[_this select 1, _this select 0, _this select 2 select 0, "ReplaceWheel"] call DFUNC(repair)};
         _action = [_name, _text, _icon, _statement, _condition, {}, [_hitpoint], _position, 2] call EFUNC(interact_menu,createAction);
         [_type, 0, [], _action] call EFUNC(interact_menu,addActionToClass);
+
+        _processedSelections pushBack _selection;
```
commy2: Now we do it twice for some selections.

commy2:
>Could you copy the link to the issue/s in here? ... for reference

https://github.com/acemod/ACE3/issues/6196

To not change the order of hit indices, you have to reference all existing subclasses of `class HitPoints` and do that in the correct order. (This also means they have to be maintained a lot.)

bux:
> #6196

The BI / CUP ones ...

addons/repair/functions/fnc_addRepairActions.sqf
Dystopian: without this line double wheel hitpoints (with same selection) get 2 actions with same position

addons/repair/CfgVehicles.hpp
Dystopian: it's still Car_F, checked in config dumper, no worries

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -36,7 +36,7 @@ if (_type in _initializedClasses) exitWith {};
 private _hitPointsAddedNames = [];
 private _hitPointsAddedStrings = [];
 private _hitPointsAddedAmount = [];
-private _processedHitpoints = [];
+private _processedSelections = [];
```
Dystopian: I also changed hitpoints to selections here because hitpoints are always unique (I hope) and there is no sense to check them. Selections as we see can be the same for different hitpoints and should be checked.

Dystopian:
`setHit` doesn't work too (just repairs 1 of hitpoints) but seems to work because uses selection name which is the same for 2 hitpoints. So config fix looks reasonable.

Dystopian:
I fixed CUP cars in their latest trunk.

jonpas:
Some more comments have to be addressed it seems.

Dystopian:
@jonpas I think both @commy2's comments are about the same thing.

Once again: after patch 
`inheritsFrom (configfile >> "CfgVehicles" >> "Truck_03_base_F" >> "HitPoints" >> "HitLBWheel")`
and
`configfile >> "CfgVehicles" >> "Truck_F" >> "HitPoints" >> "HitLBWheel"`
both equal to
`bin\config.bin/CfgVehicles/Car_F/HitPoints/HitLBWheel`
in debug console.

Or if I didn't understand correctly, @commy2 please correct me.

addons/repair/CfgVehicles.hpp
PabstMirror: I would think it should be 
```
    class Car_F: Car {
        class HitPoints {
            class HitLBWheel;
            class HitRBWheel;
        };
    };
    class Truck_F: Car_F {
        class HitPoints: HitPoints {};
    };
    class Truck_03_base_F: Truck_F {
        class HitPoints: HitPoints {
            class HitLBWheel: HitLBWheel {
```
but this works fine I think
`(configProperties [configfile >> "CfgVehicles" >> "Truck_F" >> "HitPoints", "isClass _x", false]) apply {configName _x}` = []

PR #6208, opened 2018-03-22
Add IR beam to visible laser mode

Max255PL:
- Add an IR beam to a visible laser mode for vanilla laser pointers

Visible laser is low enough power that the beam leading to the dot  is not visible to the naked eye but should still be visible for light magnification devices like NVGs. 
This PR enables the vanilla IR pointer module in addition the previously displayed laser dot which uses the vanilla flashlight module and drawIcon3D.


LorenLuke:
Could this be able to be added to laser designators?

commy2:
No.

PR #3749, opened 2016-05-03
added gunbag for DMRs

ir0n1e:
Gunbag for DMRs
~~Known issues: Mass calculation of full gunbag - cause of ArmA restrictions~~
Screenshots (tan)
http://imgur.com/a/r9l36

PR #6212, opened 2018-03-26
Add Switching of Weapon from gunbag.

jokoho48:
- Add Action to Switch Weapons from/to gun bag
- Add keybinding for switching Weapons from/to gun bag


addons/gunbag/functions/fnc_switchWeaponCallback.sqf
```diff
@@ -0,0 +1,75 @@
...
+ * [player, target] call ace_gunbag_fnc_offGunbagCallback
+ *
+ * Public: No
+ */
+#include "script_component.hpp"
+
+params ["_unit", "_target"];
+
+private _gunbag = backpackContainer _target;
+
```
dedmen: Does this really need a empty line after each line of script?

jonpas:
What is the goal of this? Switching weapon takes time anyways, and using interaction menu for that kind of plays into that.

Cuel:
> What's the goal of this?
Seems to me that instead of having to drop your weapon on the ground, use ace interact to get your gunbag weapon out, drop that on the ground, pick up the previous weapon, and put that in the gunbag, it's combined into one action

jonpas:
Ah, there was a big discussion about that a while ago IIRC.

jokoho48:
Currently to "switch" a weapon with a Gunbag you need to Open your Inventory and drop your weapon, take your weapon out of your gun bag, drop the new weapon and take the other one, put it into the gun bag and then finally get the last weapon from the ground.

and I think that is currently a very tedious progress. this PR was the idea to improve that.

addons/gunbag/functions/fnc_switchWeaponCallback.sqf
jokoho48: I just combined the other 2 functions in that module and there are written the same way.

jonpas:
IIRC from the original PR which added this feature, it was decided against adding this specifically because you can't add a weapon in same perfect order as you had it in hands (with attachments, preset attachments and all those shenanigans). Even with [`CBA_fnc_removeWeaponCargo`](https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_removeWeaponCargo.sqf) and [`CBA_fnc_addWeaponCargo`](https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_addWeaponCargo.sqf) it's not perfect due to some combinations which can't be worked around.

@commy2 probably knows more details since he was vocal about it in the original PR.
Ref. original PR: #3749

jokoho48:
@jonpas the weapon is not stored in the backpack as normal Arma weapon its stored as a virtual item. so there is no problem with weapons cargo.

jonpas:
I see, however that doesn't solve the problem of preset weapons. Say you have a preset weapon (that's one of those with attachments already on them without adding them yourself), and a player takes off one of those preset attachments, it will just reappear with your method if I am reading this correctly. `[CBA_fnc_getNonPresetClass]`(https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_getNonPresetClass.sqf) does solve it for some, but there is no guarantee every weapon even has a base class without preset attachments.

jokoho48:
but then the general system also have that issue. it is then not a issue this PR introduces

jonpas:
> but then the general system also have that issue. it is then not a issue this PR introduces.

No, it "exploits" the issues, as in introduces a way to make it happen.

PR #6219, opened 2018-03-28
Zeus - Toggle NVGs and Flashlights improvements

mharis001:
- Improve Toggle NVGs module UI and function
    - Before: 
nvg_old
    - After: 
nvg_new
- Improve Toggle Flashlights module UI and function
    - Before: 
flashlight_old
    - After: 
flashlight_new
- Add Icon for Toggle Flashlights module
- Clean up ace_ai `XEH_postInit.sqf`


addons/zeus/stringtable.xml
```diff
@@ -1097,18 +1097,14 @@
...
-            <Chinese>切換目標的夜視鏡狀態</Chinese>
-            <Chinesesimp>切换目标的夜视镜状态</Chinesesimp>
-            <Korean>목표의 야시경 토글</Korean>
-        </Key>
-        <Key ID="STR_ACE_Zeus_moduleToggleNVG_SelectedGroup">
-            <English>Selected group</English>
+        <Key ID="STR_ACE_Zeus_ToggleTarget">
+            <English>Toggle Target</English>
+        </Key>
+        <Key ID="STR_ACE_Zeus_ToggleTarget_tooltip">
```
bux: upper casing on `_Tooltip`?

jones140:
Would it be possible to add modded flashlights to a selection either via class name or list 
For example this module becomes useless with rhs opfor units as they can't equip vanilla flashlight on most weapons but have there own rhs flashlight

jonpas:
@jones140 but it does that https://github.com/acemod/ACE3/pull/6219/files#diff-f9b1874ef83cc0f4109d45c63b090e05R48?

mharis001:
@jones140 The old function checked it a compatible item was a flashlight by checking if the `Pointer` class was null, now we check if `FlashLight` class has `size > 0`. This should ensure that if a weapon has a compatible flashlight item that it is added as some attachments have both the `Pointer` & `FlashLight` class defined.

PR #6226, opened 2018-04-05
Zeus - Add ability to unload cargo

mharis001:
- Add ability to unload cargo from a vehicles zeus attribute display
cargo_unload


addons/zeus/functions/fnc_ui_attributeCargo.sqf
```diff
@@ -19,18 +19,75 @@
...
+// Init unload button
+private _button = _control controlsGroupCtrl 80087;
+
+private _fnc_onButtonUnload = {
+    params ["_button"];
+
+    // Validate vehicle
+    private _vehicle = GETMVAR(BIS_fnc_initCuratorAttributes_target,objNull);
+    if (isNull _vehicle || {!alive _vehicle}) exitWith {
+        LOG("Vehicle deleted or killed, cannot unload");
```
SilentSpike: It's a been a while, but is log a valid macro? I ask because it's an sqf command 🤔

jonpas: Yes it is.

dedmen: @SilentSpike Don't remember
`#define true false` fun? ^^
The preprocessor doesn't know what SQF is. It just does it's job and hands the result off to others.

SilentSpike: 😆 Yeah I remembered, that's why I wasn't sure if we'd actually hijacked that command

PR #6231, opened 2018-04-06
Fixes damage redirection issues

derbismarck:
Units accumulate damage in the local variables
"ACE_Medical_Engine_$HitXXXX", but these are never reset. The way these
vars are used is as a single-damage-event tracker, and the end result of
not resetting their value when they're done being used is that a
specific $HitXXXX can accumulate enough damage to cause the "select
wound with highest damage" code to always select _it_ for the next wound
event, even if it wasn't the primary target of the last wound event.

I.E. I get shot in the right leg badly. If I then get winged in the left
arm, that wound will end up on my right leg, because when the code runs,
it finds QGVAR($HitRightLeg) > QGVAR($HitLeftArm), and so it applies the
wound to the right leg.

- Fix the above-described issue

### DIFF
--- a/addons/medical_engine/functions/fnc_handleDamage.sqf
+++ b/addons/medical_engine/functions/fnc_handleDamage.sqf
@@ -141,6 +141,17 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
         [QGVAR(woundReceived), [_unit, _woundedHitPoint, _receivedDamage, _shooter, _ammo]] call CBA_fnc_localEvent;
     };
 
+    // resetting these single-damage-event tracker vars, if we don't do this then
+    // subsequent wounds will be piled onto the selection which has accumulated
+    // the most wounding
+    {
+        _unit setVariable [_x, 0];
+    } forEach [
+        QGVAR($HitFace),QGVAR($HitNeck),QGVAR($HitHead),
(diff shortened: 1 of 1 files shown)


commy2:
This is a terrible idea.

SilentSpike:
After looking closely at the handle damage code, I'm interpreting it the same way as @derbismarck and this seems to be a correct fix to me.

It's just resetting the cached value that's used to determine where a wound should be applied between injuries.

Entirely possible the code doesn't work like we're reading it though (assuming handler first fires for the normal hitpoint, then fires afterwards for the "ace_hdbracket" hitpoint which triggers the logic to apply a wound to whichever normal hitpoint took the most damage).

SilentSpike:
In image of the issue in action:



All of the shots listed in the bottom left were put where my cursor is, clearly the torso

thojkooi:
Merging into the restructure branch.

SilentSpike:
Bad news is that I can still reproduce this issue on latest restructure branch where this was merged 😢

TheMagnetar:
I tested it some time ago in rewrite and it was working fine. Maybe variables have been moved?

SilentSpike:
Yeah that's a possibility, I can't see any reason it shouldn't be working

SilentSpike:
Nevermind, I just forgot to pull before testing 😆 It's been a while...

PR #6238, opened 2018-04-08
Fixed airdrops sink in the ground

### LINKED ISSUE #6237: helicopter airdrop bugs out of the world
**Arma 3 Version:** stable
**CBA Version:** stable
**ACE3 Version:** stable

**Mods:**
```
- CBA_A3
- ace
- rhsusaf
- ares
```

**Description:**
- drop a container out of a helicopter and look how it goes to the ground
- Video: https://youtu.be/0ADEBPhaOS0
- Information: also my own script airdrop bugs out of the world.... so i suppose in the last update something changed!

**Steps to reproduce:**
- heli with cargo
- drop while in the air
- follow the cargo
- the cargo goes in the ground

shukari:
- fix #6237

### DIFF
--- a/addons/cargo/functions/fnc_paradropItem.sqf
+++ b/addons/cargo/functions/fnc_paradropItem.sqf
@@ -65,7 +65,7 @@ _itemObject setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vecto
 
     private _velocity = velocity _item;
 
-    _item attachTo [_parachute, [0,0,-1]];
+    _item attachTo [_parachute, [0,0,1]];
     _parachute setVelocity _velocity;
 
     private _light = "Chemlight_yellow" createVehicle [0,0,0];


commy2:
Wouldn't this mean the cargo is stuck inside the parachute model?

shukari:
Yes... look here -> https://youtu.be/IYS29dWDe5s?t=1m20s

But the problem is... the driangle thing is the bottom of the parachute... and because the supply is attached is has no physics so... the parent model must be the bottom...

It no works for every supply box!

dedmen:
>It no works for every supply box!

How about only doing it for the boxes that actually don't have physx and need this "fix"?

shukari:
The problem is the attachTo to the parachute.... then the supplybox is there but has no collision. So then the parachute then hit the ground and the box is in the ground if it is not attached a littlebit over the bottom of the parachute.

shukari:
> It no works for every supply box!
That was a fail... i mean NOW it works for everything.

shukari:
Vanilla is doing the same... if you place a supplybox in zeus and then change the height to 100m (with ALT) it will automatic attach a parachute to it with attachTo [0,0,1] ^^

Cuel:
Seems logical to attach it above the obejct instead of under

SilentSpike:
I can confirm that this is how curator does it (and it works):
```sqf
case "Object": {
	switch _objectType do {
		case "AmmoBox": {
			//--- Create parachute for vehicles in the air
			if ((position _object select 2) > 20 && alive _object) then {
				_para = createvehicle ["B_Parachute_02_F",_objectPos,[],0,"none"];
				_object attachto [_para,[0,0,1]];
			};
		};
	};
};
```

SilentSpike:
Question though, why do we not just create the parachute where the object is? (Instead of at `[0,0,0]` then moving it)

shukari:
> Question though, why do we not just create the parachute where the object is? (Instead of at [0,0,0] then moving it)

I dont understand what you mean... with this PR it works for all container... do you mean dynamically?

SilentSpike:
Not specific to this PR, I was just looking at the code we have above these changes:

```sqf
    private _parachute = createVehicle ["B_Parachute_02_F", [0,0,0], [], 0, "CAN_COLLIDE"];
 
     // cannot use setPos on parachutes without them closing down
     _parachute attachTo [_item, [0,0,0]];
     detach _parachute;
 
     private _velocity = velocity _item;
 
    _item attachTo [_parachute, [0,0,1]];
     _parachute setVelocity _velocity;
```

Bosenator:
Before this fix, my solution was the following:

```
if (isServer) then {
	["ace_cargoUnloaded", {
		params ["_item", "_vehicle", ["_type", ""]];

		if (_type == "paradrop") then {
			[_item] spawn {
				params ["_crate"];
				
				if (isNull _crate) exitWith {};				
				waitUntil {(getPosVisual _crate select 2) < 1};
				_groundPosition = getPos _crate;
				detach _crate;
				_crate setPos ([_groundPosition select 0, _groundPosition select 1, 0]);
			};
		};
};
```

It's messy by all means, but perhaps the logic could be applied?

PR #6240, opened 2018-04-09
Fixed mortar double loading

### LINKED ISSUE #5671: Two rounds in mortar (single loading setting)
**Arma 3 Version:** stable
**CBA Version:** stable
**ACE3 Version:** stable

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
Two people can load the mortar at the same moment and the mortar had then 2 rounds. Is it possible to make a mutex for this

**Steps to reproduce:**
- 2 guys load mortar at the same moment
- 2 rounds in mortar

shukari:
- fix #5671
- tested and all seams ok


PabstMirror:
I think this will help a lot, but with network delay it would still be possible to have a collision.
Still an improvement that will work most of the time.

PR #6241, opened 2018-04-09
Make common_fnc_stringToColoredText accept hex string as color.

dedmen:
It seems kinda unintelligent to always convert constants like in
https://github.com/acemod/ACE3/blob/master/addons/reload/functions/fnc_displayAmmo.sqf#L98
into string dynamically. We already know it's constant so wouldn't it be nice if it didn't have to re-calculate the same result each time it's called?

Also
brofiler_2018-04-09_13-24-10
Looks like some uses could be quite a bit faster without always converting it.
Yeah that's only 0.1ms but.. why not get rid of it.

### DIFF
--- a/addons/common/functions/fnc_stringToColoredText.sqf
+++ b/addons/common/functions/fnc_stringToColoredText.sqf
@@ -4,7 +4,7 @@
  *
  * Arguments:
  * 0: Text <ANY>
- * 1: Color <ARRAY>
+ * 1: Color <ARRAY, STRING>
  *
  * Return Value:
  * Text <STRING>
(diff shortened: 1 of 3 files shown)


addons/common/functions/fnc_stringToColoredText.sqf
```diff
@@ -20,12 +20,14 @@ params ["_string", "_color"];
...
 _string = format ["%1", _string];
 
-_color = (
-    [255 * (_color select 0), 2] call FUNC(toHex)
-) + (
-    [255 * (_color select 1), 2] call FUNC(toHex)
-) + (
-    [255 * (_color select 2), 2] call FUNC(toHex)
-);
+if (_color isEqualType 0) then {
```
SilentSpike: Should this not be `isEqualType []`?

dedmen: duh... thanks.

PR #6242, opened 2018-04-09
Improve Reload TakeEH

dedmen:
No reason to call `_this select 0` 7 times in total.

Improvement should be around 0.03ms per call. Yeah. Not much.

Gonna take a look at taking items from containers in general. Maybe I can find more improvements.

Biggest cause of lag when taking items from containers in my setup seems to be ASR AI. config lookup for uniform/vest/backpack each time you take whatever item even if your clothing didn't change.

### DIFF
--- a/addons/explosives/CfgEventHandlers.hpp
+++ b/addons/explosives/CfgEventHandlers.hpp
@@ -24,7 +24,7 @@ class Extended_Killed_EventHandlers {
 
 class Extended_Take_EventHandlers {
     class CAManBase {
-        GVAR(takeHandler) = QUOTE([ARR_3(_this select 0, _this select 1, _this select 2)] call FUNC(onInventoryChanged));
+        GVAR(takeHandler) = QUOTE(call FUNC(onInventoryChanged));
     };
 };
 class Extended_Put_EventHandlers {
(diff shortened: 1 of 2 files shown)


dedmen:
Ref: https://github.com/robalo/mods/pull/8

Next biggest "issue" is ace_explosives_fnc_onInventoryChanged. Which runs on every take/put 
https://github.com/acemod/ACE3/blob/master/addons/explosives/CfgEventHandlers.hpp#L25-L34
Does that have to run on every take/put? Or would CBA loadout changed EH be sufficient?

SilentSpike:
Looks like the explosives one needs the giver/receiver

dedmen:
>Looks like the explosives one needs the giver/receiver
Yep. It needs to make sure that it's the local player. Which CBA inventory changed EH implicitly does.

SilentSpike:
The inventory changed EH doesn't track who gave the item though? (Which that function needs to read the variable from)

dedmen:
Ah I see now. Yeah probably not an option then

dedmen:
Didn't find any more optimizations that would be worth it. Atleast not on default ACE configuration.

PabstMirror:
```
GVAR(takeHandler) = QUOTE([ARR_3(_this select 0, _this select 1, _this select 2)] call FUNC(onInventoryChanged));
```
In explosives, I think we could just pass `_this` and save 3 selects

dedmen:
Now I feel bad for not having seen that.

PR #6246, opened 2018-04-10
Remove arsenal when already present

oOKexOo:
- Fix the issue that the `Add full Arsenal` Zeus module fails when the box already has an Arsenal.


alganthe:
Approved.

Issue #5510, opened 2017-09-15
Fatal Damage - only headshots would kill a downed unit


ViperMaul:
This was brought up in some past conversations.

**Observations:**
https://www.youtube.com/watch?v=wD8ZKuD-5D4 
Note that the player survives many bullets.

The victim gets hit.
He seems to receive a non-instantly fatal wound. 
He goes unconscious.
Then he just bleeds. 

**Notes:**
- This was not on our internal ACE Dev test server however. 
- This was an AI - but it is believed this should not matter 

**Expected:**
Player is expected to die faster in this video.

derbismarck:
https://github.com/acemod/ACE3/blob/medical-rewrite/addons/medical/functions/fnc_conditionCardiacArrestTimer.sqf - x60 multiplication is probably not meant to be there. Unit will wait 30 minutes (in default configuration) before dying in cardiac arrest.

https://github.com/acemod/ACE3/blob/32e73fa1b314180dc436985a461bf25f5cda7881/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf#L127 - Controls when the `FatalInjury` event occurs (only headshots based on the condition) that will cause the state machine to enter the `FatalInjury` state from `Default`, `Injured`, or `Unconscious`. Also the same event that causes the transition from `CardiacArrest` to `Dead`.

SilentSpike:
> x60 multiplication is probably not meant to be there. Unit will wait 30 minutes (in default configuration) before dying in cardiac arrest.

Yeah it looks like this setting used to be in minutes (see 39582914f743d586efcc361e9d01d97cd8e0ab77), but was later changed to seconds without updating this piece of code to match.

DocScarle:
A body dies from shock. Septic shock caused by bacteria, Anaphylactic shock caused by hypersensitivity or allergic reaction, Cardiogenic shock from heart damage, Hypovolemic shock from blood or fluid loss, and Neurogenic shock from spinal cord/brain trauma.
Therefore in slow reality it would mean that the AI you shot would only die once these criteria have been met.
Would it not be right to have each wound cause an amount of Hypovolemia that is added to a total once that total reaches a certain level the patient suffers injuries not conducive with life and expires.

SilentSpike:
In addition to something along the lines of #6277 we might want to think about removing the cap on blood loss and letting units bleed out incredibly fast if wounds stack up

PR #6277, opened 2018-04-15
Add possibility for lethal heart damage

SilentSpike:
- ~~Changes the fatal injury logic to account for all wounds across the body. This way if you take many wounds it will eventually become fatal (with weighting applied to each body part so that the head still takes one shot, the body takes quite a few and the limbs take many).~~
- ~~This should also mean that if you get hit in the chest with a tank shell; stand on a grenade; jump from very high; etc. then you'll probably die instantly (which seems more realistic to me, but should also be controllable via the fatal injury setting).~~

~~Not sure if this is the ideal solution and haven't tested extensively (just enough to confirm it's one form of solution).~~

Outdated, see comments below.

Reference: #5510


thojkooi:
This seems to work pretty well. It's a huge improvement over current behaviour.

SilentSpike:
I think it's fine as a temporary implementation, maybe once we introduce more death conditions it won't be necessary

SilentSpike:
PR has now been updated to add the possibility for lethal heart damage to torso damage only.

This more correctly reflects the purpose of the lethal damage system. Down the line I could see us adding further cases for other organs which aren't lethal, but could contribute to causes of death - akin to https://github.com/acemod/ACE3/issues/5510#issuecomment-361489263.

Hopefully we can also make #6470 work such that enough bullets will cause sufficient bloodloss for a bleedout death to fix #5510.

addons/medical_engine/script_macros_medical.hpp
```diff
@@ -3,6 +3,12 @@
 #define ALL_SELECTIONS ["head", "body", "hand_l", "hand_r", "leg_l", "leg_r"]
 #define ALL_HITPOINTS ["HitHead", "HitBody", "HitLeftArm", "HitRightArm", "HitLeftLeg", "HitRightLeg"]
 
+// Damage threshold above which fatal organ damage can occur
+#define ORGAN_DAMAGE_THRESHOLD 1
+// Chance to hit heart based on ratio of 70kg (approx. 70L) body to 70mL stroke volume of heart
+// Assuming torso is 50% of the body volume (35L)
+#define HEART_HIT_CHANCE 0.005
```
TheMagnetar: Shouldn't this be 70/(35*1000) = 0.002? You have hit the torso already so, the 35L is your 100%

TheMagnetar: It's a pitty we cannot really check were the hit is relative to the centre of masses of the given unit...

SilentSpike: I rounded it up to the nearest half percent because it's such a small chance already.

Yeah, even more ideally units would have organs modelled inside where we could detect hits on them

thojkooi:
This does not fix the issue.

image

This was 8 mags. Also done on multiple other body parts.

I think the previous implementation was much better @SilentSpike.

thojkooi:
I did some changes and combined with the bug fixes in #6481 this should now work decently enough. I'd opt to merge this PR and change if necessary based on feedback.

SilentSpike:
Yeah this seems good to me 👍 We can always scale it back later if we want to more specifically differentiate between organs and their effects.

PR #6278, opened 2018-04-16
(repair) - isRepairVehicle handling values according to wiki

Cuel:
The wiki states that `ACE_isRepairVehicle` is a boolean but this function only checked for integer values, resulting in a script error for whoever used the wiki value. 

-  Make isRepairVehicle handle both types

### DIFF
--- a/addons/repair/functions/fnc_isRepairVehicle.sqf
+++ b/addons/repair/functions/fnc_isRepairVehicle.sqf
@@ -20,4 +20,5 @@ TRACE_1("params",_vehicle);
 
 if (_vehicle isKindOf "CAManBase") exitWith {false};
 
-((_vehicle getVariable ["ACE_isRepairVehicle", getNumber (configFile >> "CfgVehicles" >> typeOf _vehicle >> QGVAR(canRepair))]) > 0);
+// Value can be integer or boolean
+([false, true] select (_vehicle getVariable ["ACE_isRepairVehicle", getNumber (configFile >> "CfgVehicles" >> typeOf _vehicle >> QGVAR(canRepair))])) // return
--- a/addons/repair/functions/fnc_isRepairVehicle.sqf
+++ b/addons/repair/functions/fnc_isRepairVehicle.sqf
@@ -21,4 +21,5 @@ TRACE_1("params",_vehicle);
 if (_vehicle isKindOf "CAManBase") exitWith {false};
 
 // Value can be integer or boolean
-([false, true] select (_vehicle getVariable ["ACE_isRepairVehicle", getNumber (configFile >> "CfgVehicles" >> typeOf _vehicle >> QGVAR(canRepair))])) // return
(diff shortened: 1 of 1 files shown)


jonpas:
Hmh, I'd rather change wiki, that leaves open possibilities for repair levels or something.

dedmen:
So does adding the numbering back when we want repair levels.
a `isRepairVehicle` being a boolean makes much more sense. If it was a repair level it should be called differently anyway

BaerMitUmlaut:
I may be nitpicking here, but why the outer parenthesis? Why the `// return` comment? It's the last statement, this being the return value should be clear for somebody who knows SQF.
Additionally, right now it's a 160 character line. Please consider splitting it.

Disregarding that, I agree with dedmen. The is prefix implies that it's a boolean and it's fully backwards compatible. I don't see us moving to multiple repair levels this late into A3.

Cuel:
At least in CBA `//return` comments have been popping up here and there so I figured I'd add it. If that doesn't apply to ace I'll remove it

commy2:
I do that whenever it's not clear what a return value is. It's debateable here, because on the one hand, this expression is huge, on the other, it is the last line of the script.

What bugs me more about this PR is:
```sqf
[false, true] select 
```

Cuel:
That's what I had at first
https://github.com/Cuel/ACE3/commit/74f5b99dd85b836cebc4c96109b07fb97f7458df

SilentSpike:
```sqf
[false, true] select
```

😱

Cuel:
So would you prefer [74f5b99](https://github.com/Cuel/ACE3/commit/74f5b99dd85b836cebc4c96109b07fb97f7458df) or something else

SilentSpike:
Nah it's best as is because the `getVariable` can return an index. I was just being silly 😉

Cuel:
Yeah I was talking to everyone who's commenting here 👍

dedmen:
My preference would be a slightly modified variant of https://github.com/Cuel/ACE3/commit/74f5b99dd85b836cebc4c96109b07fb97f7458df
```
private _value = _vehicle getVariable ["ACE_isRepairVehicle", -1];
if (_value == -1) exitWith { getNumber (configFile >> "CfgVehicles" >> typeOf _vehicle >> QGVAR(canRepair)) > 0 };
(_value isEqualTo true || {value > 0})  // return
```

Dystopian:
@dedmen
to handle `false` in `_value`
`(_value isEqualTo true || {value > 0})  // return`
->
`if (_value isEqualType true) then {_value} else {value > 0}`

dedmen:
^ That is slower though...

Cuel:
But you'd be checking if `{false > 0}`

Cuel:
Any decision?

SilentSpike:
Honestly I don't think we need to handle values greater than one for the time being. The variable name doesn't imply that anyway and I really don't foresee a need for us to add it.

If we want to add repair levels in future we can PR it separately. I say keep this code simple `in [1, true]`

Cuel:
For only handling 0/false/1/true I'd say the current PR is good enough

Cuel:
:(

PR #6288, opened 2018-04-20
Improve medical body image/interaction colouring

### LINKED ISSUE #4347: Damage required to turn a body part red is too high in advanced medical
**Arma 3 Version:** `1.64` RC
**CBA Version:** `3.0.0`
**ACE3 Version:** `3.7.0` release; commit: f3dc1afdbfc8dc05c6057a5d6806c1a7cc10e6db

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
- The amount of damage required to turn a body part icon (the crosses) red is too high, 2 large wounds are still shown as yellow when before a single large wound was red.
- http://puu.sh/r1mTx/671444d92a.jpg

SilentSpike:
This makes the blood loss calculation used for the interaction menu icons match the calculation used for the body image.

Closes #4347

Edit: This has since become an implementation of the method used by ShackTac to colour the icons and body image selections. It fixes the issue and adds some granularity to the interaction icons. It does change the behaviour slightly of the body image colouring (now matches the icons from yellow through to red instead of pink through to red).

The new icons follow a ["white, yellow, red" colour scale](https://gka.github.io/palettes/#colors=white,yellow,red) with 10 steps, Bezier interpolation and Correct lightness gradient.

Edit 2: This also now includes a tourniquet icon overlay (rather than colouring the limb blue - which is being used to now show damage again since orange is taken by blood loss).


SilentSpike:
The `20` scaling factor seems somewhat arbitrary, but it works for the colour calculations done on the body image (`fnc_updateBodyImage.sqf` in `medical_menu`) so essentially this just adjusts the interaction icons to match.

Seems ideal for a macro instead of a magic number, but I don't know what to call it 🤔

I'll also have to do some maths to check the output colour for the body image and adjust the threshold for the icon to match when it turns a darker red.

SilentSpike:
The now adjusted threshold will turn the icon red when the body image is approximately #FF4D4D

SilentSpike:
Okay I tested this now and it goes instantly to red, something seems wrong with that scaling factor. I wonder now if the body image is what's actually wrong.

SilentSpike:
Yeah so it seems the issue here is actually the other way around, the body image calculation was wrong because of the extra factor of 20 (it would turn instantly to the darkest red).

Which brings us back to what the threshold should be set at on the action menu icon to turn it from yellow to red.

SilentSpike:
This is now in a state where everything is consistent and the interaction icons turn red when the body image icons start to turn to a dark red (at 0.5 blood loss). However, this may not reflect what is actually a potentially lethal blood loss - in which case we need to adjust the icon threshold and the body image calculations.

SilentSpike:
Here's how it now looks with a tourniquet applied:



and without:



I kind of don't like the discrepancy between the icon and the body image here, but it's how it was behaving beforehand too.

SilentSpike:
Okay so looking further into this:

1. It seems that in the rewrite the body image colours were updated so that limbs went from pale to dark red based on the blood loss they had. If there was no blood loss, they then went from pale to dark orange based on the damage (previously this was done with blue - which was changed to show a limb with tourniquet applied).
2. The new behaviour in this PR (based on ShackTac's implementation) changes the interaction icons and limbs to step through yellow, orange, red as the blood loss increases. This definitely looks better for the icons and personally I like to have the consistency with this on the body image (it is confusing to convey the same information in two different ways based on where you're looking).
3. However, this now means that damage isn't displayed on the body image once blood loss is stopped (and keeping the orange would be confusing because of the blood loss changes).

So the question is, do we revert the body image colour changes (and accept inconsistency with the interaction icons)? Or, is there another colour we would want to to show damage? Another option is to adopt ShackTac's method of showing a tourniquet and apply a little overlaying icon on the limb and reverting to blue to show damage.

Drofseh:
I'm more concerned with blood loss than with damage, and I think colour consistency with the interaction icons is very important.

If there is going to be an overlay icon (which I think is a good idea) I'd rather it showed damage while the limb was coloured for blood loss.


Or maybe do it like Escape from Tarkov and have coloured bars for damage.
I wouldn't use numbers like Tarkov does though.

jonpas:
That makes a lot of sense, since tourniquet is on the limb image as well, comme ter with bleeding. And if damage is a bar this also directly conveys its damage.

SilentSpike:
Currently I think I'm going to add the tourniquet image provided by ShackTac and then make it behave like the current release version where once blood loss is stopped the damage is shown in blue. That way it's consistent with older versions and there's no issue with colour overlap.

The bar is not a bad idea, I just feel that it's too accurate in comparison.

addons/medical/functions/fnc_displayPatientInformation.sqf
```diff
@@ -158,30 +158,35 @@ if (_show == 1) then {
...
-                    if (_damage > 0.1) then {
-                        _blue = 0 max (1 - _damage);
-                        _green = 0.5 max (1 - _damage / 2);
-                    };
-                };
+
+                private _frBL = 0 max (_bloodLoss / BLOOD_LOSS_RED_THRESHOLD) min 1;
+                private _colorInt = ceil ((_frBL * (BLOOD_LOSS_TOTAL_COLORS - 1)));
+
+                _colorSelection = [
```
mharis001: Should this array of colors be a macro?

Drofseh:
Yah, I don't really care about seeing damage at all tbh, well, unless treatment for damage is significantly different in the rewrite.

Right now in 3.12.2 you either require a PAK or you have to have bandages heal damage turned on right?
Is that changed in the rewrite?

addons/medical/functions/fnc_displayPatientInformation.sqf
SilentSpike: I'm going to tidy a bunch of stuff up once the functionality is finalised

SilentSpike:
The final result:



As you can see one leg is damaged but not bleeding, while the other was bleeding. Both have a tourniquet applied to show how the colour looks.

addons/medical_menu/CfgVehicles.hpp
```diff
@@ -48,7 +48,7 @@ class CfgVehicles {
                 displayName = ECSTRING(medical,Actions_Medical);
                 runOnHover = 1;
                 exceptions[] = {"isNotInside", "isNotSitting"};
-                statement = QUOTE([ARR_3(_target,1,0)] call EFUNC(medical,displayPatientInformation));
+                statement = QUOTE([ARR_3(_target,1,0)] call EFUNC(medical_ui,displayPatientInformation));
                 condition = "true";
                 icon = QPATHTOEF(medical,UI\icons\medical_cross.paa);
```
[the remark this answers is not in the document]
SilentSpike: Well spotted, thanks 👍

SilentSpike:
If #6290 gets merged in future we can just use a single basic cross icon and update the colour of it. Then we only need multiple of the tourniquet icons since they retain the white outer ring.

SilentSpike:
One thing I'm not totally sure about is at what level of a damage a selection should reach the bluest blue. Currently it's set to 0.8 ([here](https://github.com/acemod/ACE3/blob/5eb6a389c0b279d5fbe758aa230663724af96622/addons/medical/script_macros_medical.hpp#L42)), but that's just because I mistakenly thought damage was in the range `[0,1]` when actually it seems to just keep going up (a single pistol shot seems to instantly make a limb the highest level of blue).

mharis001:
Included in #6831 .

PR #6290, opened 2018-04-20
Interaction Menu - Ability to modify icon color

mharis001:
- Add ability to modify action icon color without multiple icon files
- Quick demo: https://gfycat.com/SmoggySlimyGazelle

### DIFF
--- a/addons/interact_menu/functions/fnc_renderIcon.sqf
+++ b/addons/interact_menu/functions/fnc_renderIcon.sqf
@@ -1,10 +1,10 @@
 /*
- * Author: NouberNou and esteldunedain
- * Render a single interaction icon
+ * Author: NouberNou, esteldunedain, mharis001
+ * Render a single interaction icon.
  *
  * Arguments:
  * 0: Text <STRING>
- * 1: Icon <STRING>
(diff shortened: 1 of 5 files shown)


addons/interact_menu/functions/fnc_renderIcon.sqf
```diff
@@ -33,14 +33,15 @@ if(GVAR(iconCount) > (count GVAR(iconCtrls))-1) then {
 };
 private _ctrl = GVAR(iconCtrls) select GVAR(iconCount);
 
-if(_icon == "") then {
-    _icon = DEFAULT_ICON;
+_icon params [["_iconFile", "", [""]], ["_iconColor", "#FFFFFF", [""]]];
+if (_iconFile isEqualTo "") then {
+    _iconFile = DEFAULT_ICON;
```
jonpas: Just put it in default value in `params` above?

mharis001: That would not work because `params` will not use `DEFAULT_ICON` if `_icon` is `""`.

jonpas: Ah, I see.

PR #6294, opened 2018-04-23
Zeus - Improve Editable Objects module and Cleanup

mharis001:
- Improve Update Editable Objects UI and add Icon
    - Before:
editobj_before
    - After:
editobj_after
    - Icon: 
editobj_icon_zeus
- Remove `Attribute Position`: it was not functional since map controls cannot be in controls groups
- Cleanup AI module functions and use `QQGVAR()` instead of `QUOTE(QGVAR())`


mharis001:
The one issue with that I see is selecting additional options. Either the `curatorInfoType` would have to be displayed manually after position selection or the user selects all the options, clicks ok, selects position and then the action goes through. Good feature to expand and use more often nevertheless.

SilentSpike:
Yeah that's the UX I was hoping to investigate and figure out if it feels too awkward. Would be good to find an ideal solution as there are a few modules that could use both a UI and position selection.

PR #6295, opened 2018-04-23
Add option to remove action from all child classes

### LINKED ISSUE #5009: ace_interact_menu_fnc_removeActionFromClass issue
**CBA Version:** `3.2.1` (stable)
**ACE3 Version:** `3.9.1` (rc1)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- The command doesn't delete action from the unit.

**Steps to reproduce:**
- Add this code into your mission.
```
_action = ["custom_action", "TEST ACTION", "", {systemChat "test action";}, {true}] call ace_interact_menu_fnc_createAction;
["CAManBase", 1, ["ACE_SelfActions"], _action, true] call ace_interact_menu_fnc_addActionToClass;

["CAManBase", 1, ["ACE_SelfActions", "custom_action"]] call ace_interact_menu_fnc_removeActionFromClass;
```
only this one works
```[typeOf player, 1, ["ACE_SelfActions", "custom_action"]] call ace_interact_menu_fnc_removeActionFromClass;```

**Placed Modules:**
- None.

**RPT log file:**
- N/A.

### PR DESCRIPTION
**Pull request by xrufix**
- Currently, `ace_interact_menu_fnc_removeActionFromClass` doesn't support inheritance. This PR fixes that issue.
- Allows excluding classes from fnc_addActionToClass.

fixes #5009

To do:
- [ ]  Prevent initEH adding the action after calling fnc_removeActionFromClass

### DIFF
--- a/addons/interact_menu/functions/fnc_removeActionFromClass.sqf
+++ b/addons/interact_menu/functions/fnc_removeActionFromClass.sqf
@@ -17,7 +17,17 @@
  */
 #include "script_component.hpp"
 
-params ["_objectType", "_typeNum", "_fullPath"];
+params ["_objectType", "_typeNum", "_fullPath", ["_inherit", false, [false]]];
+
+// Call this function for all child classes if inheritance is enabled.
+if (_inherit) then {
+    private _children = (
+        (format ["_class = configName _x; (_class isKindOf '%1') && (_class != '%1')", _objectType]) configClasses (configFile >> "CfgVehicles")
+    ) apply {configName _x};
+    {
+        [_x, _typeNum, _fullPath] call FUNC(removeActionFromClass)
+    } forEach _children;
+};
 
 private _res = _fullPath call FUNC(splitPath);
 _res params ["_parentPath", "_actionName"];
--- a/addons/interact_menu/functions/fnc_removeActionFromClass.sqf
+++ b/addons/interact_menu/functions/fnc_removeActionFromClass.sqf
@@ -21,13 +21,9 @@
 params ["_objectType", "_typeNum", "_fullPath", ["_inherit", false, [false]]];
 
 // Call this function for all child classes if inheritance is enabled.
-if (_inherit) then {
-    private _children = (
-        (format ["_class = configName _x; (_class isKindOf '%1') && (_class != '%1')", _objectType]) configClasses (configFile >> "CfgVehicles")
-    ) apply {configName _x};
-    {
-        [_x, _typeNum, _fullPath] call FUNC(removeActionFromClass)
-    } forEach _children;
+if (_inherit) exitWith {
+    private _children = (format ["(configName _x) isKindOf '%1'", _objectType]) configClasses (configFile >> "CfgVehicles");
+    _children apply { [configName _x, _typeNum, _fullPath] call FUNC(removeActionFromClass) };
 };
 
 private _res = _fullPath call FUNC(splitPath);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: addons/interact_menu/functions/fnc_removeActionFromClass.sqf [hunk 1]
```diff
@@ -17,7 +17,17 @@
  */
 #include "script_component.hpp"
 
-params ["_objectType", "_typeNum", "_fullPath"];
+params ["_objectType", "_typeNum", "_fullPath", ["_inherit", false, [false]]];
```
PabstMirror: Update header

PabstMirror:
```
private _action = ["VulcanPinch","Vulcan Pinch","",{_target setDamage 1;},{true},{},[], [0,0,0], 100] call ace_interact_menu_fnc_createAction;
private _path = ["CaManBase", 0, ["ACE_TapShoulderRight"], _action, true] call ace_interact_menu_fnc_addActionToClass;

["CaManBase", 0, ["ACE_TapShoulderRight", "VulcanPinch"], true] call ace_interact_menu_fnc_removeActionFromClass;
```
Run that, then spawn in a new unit (of new unique classname); it will still have the action because addActionToClass adds a init EH.

addons/interact_menu/functions/fnc_removeActionFromClass.sqf
```diff
@@ -6,18 +6,29 @@
...
+
+// Call this function for all child classes if inheritance is enabled.
+if (_inherit) then {
+    private _children = (
+        (format ["_class = configName _x; (_class isKindOf '%1') && (_class != '%1')", _objectType]) configClasses (configFile >> "CfgVehicles")
```
Dystopian: it will be faster and pretty to check condition in count/forEach loop for `'true' configClasses` array

dedmen: Why are two loops faster than one loop?

xrufix: Is this better?
```
if (_inherit) then {
    {
        private _class = configName _x;
        if ((_class isKindOf _objectType) && {_class != _objectType}) then {
            [_class, _typeNum, _fullPath] call FUNC(removeActionFromClass);
        };
    } forEach ("true" configClasses (configFile >> "CfgVehicles"));
};
```

dedmen: Nope way worse. Adding a if and then for each element. And as mentioned the second loop
Actually looking at the current code. you have 3 loops but you only two.
```
private _children = (format ["_class = configName _x; (_class isKindOf '%1') && (_class != '%1')", _objectType]) configClasses (configFile >> "CfgVehicles");
_children apply { [configName _x, _typeNum, _fullPath] call FUNC(removeActionFromClass) }
```
Instead of apply you could use any loop type you want. count/forEach/select/apply 
But I'd say apply is the most readable of all.
count might be faster though if removeActionFromClass returns nil.

Dystopian: @dedmen yeah, looks like BI optimized `configClasses` much, it was much slower months ago.
@xrufix sorry for bothering

xrufix: It should be better now.

Mike-MF:

PR #6296, opened 2018-04-23
Add ace hearing support to the tank DLC helmet

alganthe:
- `H_Tank_black_F` doesn't have hearing support yet, so here it is.


SilentSpike:
As an aside, I hate that these macros include the semicolon

PR #6297, opened 2018-04-23
Medical Rewrite - Fix missing function errors

mharis001:
- Get rid of missing function errors


mharis001:
Agreed, hard to work on it when there's 2 different _versions_ of the rewrite. Should merge the restructure and bring it up to being able to launch so working on it is easier.

PR #6299, opened 2018-04-24
Hellfire - Update to use 1.82 submunition

tbeswick96:
- Change hellfire to use Vikhr (`M_Scalpel_AT`) as base missile along with its submunition configuration
- Remove damage changes to Kilo as damage is dealt by submunition
- Remove submunition from November
- Increase hellfire seeker max range to 8km
- Update frag base class


PabstMirror:
Update base class here as well
https://github.com/acemod/ACE3/blob/master/addons/frag/CfgAmmo.hpp#L137

tbeswick96:
@PabstMirror updated frag base class as well

PR #6301, opened 2018-04-24
Restructure medical rewrite

SilentSpike:
Just opening this PR so we can see the progress of it compared to the main rewrite branch


addons/medical/script_macros_medical.hpp
```diff
@@ -56,3 +56,14 @@
 
 // Minimum body part damage required for blood effect on uniform
 #define VISUAL_BODY_DAMAGE_THRESHOLD 0.35
+
+// - Status macro functions ---------------------------------------------------
+// These macros provide the same functionality as the functions in the status
+// component, but are slightly faster (because most are just object variables)
+#define GET_PAIN_TOTAL(unit)     (unit getVariable [QEGVAR(status,pain), 0])
```
SilentSpike: Should these not all be `QEGVAR(medical_status,XXX)`

SilentSpike:
Huh, how the heck did 5206b63d5aeb6b8913c47be543d5531ca7f31070 get onto this branch 😕

commy2:
176 files changed O_O

SilentSpike:
Writing out my current understanding of this restructure to check that it makes sense before I start moving things around:

- medical: TBD
- medical_ai: Handles AI using the medical system
- medical_blood: Handles blood drops/pools
- medical_damage: Handles scripted damage (wound allocation, etc.)
- medical_engine: Handles engine based damage and status (`setUnconscious`, handleDamage EH, etc.)
- medical_feedback: SFX and FX that provide feedback to the player (heartbeat, blood overlay)
- medical_gui: Body part interaction points, medical menu, patient information display
- medical_statemachine: Tracks state of units (unconscious, cardiac arrest, normal, etc.) and handles transitions
- medical_status: Handles scripted status (blood volume, pain, medical ability, etc.)
- medical_treatment: Adds treatment items and actions (bandage, CPR, etc.)
- medical_vitals: Handles vitals loop of units (constantly updates unit status)

@BaerMitUmlaut Does this seem correct to you? I'm going to move some functions and things around to tidy things up more in line with this if so.

BaerMitUmlaut:
Yes, I think that's all correct. Medical core could contain API maybe. Everything else can be distributed into the other packages I think.

SilentSpike:
Currently thinking we move moans to into feedback and move the medical vehicle/medic unit stuff into treatment.

mharis001:
Could be worth splitting `medical_menu` from `medical_gui`. Also, the `ui\ui\..` paths should be fixed?

mharis001:
True, before information display injury list and medical menu injury list didn't use the same function, even though they were the same.

SilentSpike:
Looking closer at the use of blood pressure, I'm not actually sure why the vitals loop is setting it as a variable on the unit. Everywhere that blood pressure is used just calls the function to calculate it anyway.

Bummeri:
I applaud your efforts!
A long time ago I remember reading that medical-rewrite was going to deprecate the B-medical system. Is this information still correct?

BaerMitUmlaut:
All differences between what is currently considered basic and advanced are individual settings, so you can still have something like the current basic system. Might enable us to do something even more basic.

addons/medical/functions/fnc_addDamageToUnit.sqf
```diff
@@ -40,14 +40,14 @@ if (!isNull _instigator) then {
 };
 
 #ifdef DEBUG_TESTRESULTS
-private _startDmg = +(_unit getVariable [QGVAR(bodyPartDamage), [-1]]);
+private _startDmg = +(_unit getVariable [QGVAR(medical,bodyPartDamage), [-1]]);
```
TheMagnetar: Why is QGVAR(medical,bodyPartDamage) used in here? should not it be QEGVAR(medical,bodyPartDamage) or since we are in medical module leave it as QGVAR(bodyPartDamage)?

SilentSpike: Good catch! For the time being EGVAR is fine as things may be moved around a bit more and it saves them breaking

SilentSpike:
Here's how the current dependency graph looks for the restructure (excluding `medical` which I think we should gut and reserve for public functionality):

image

This was generated by searching for extended function calls between components, it's possible that there are some implicit dependencies beyond these.

TheMagnetar:
One implicit dependency, and the most important one, is that state-machine depends strongly on ace_medical. Without it, the states will not change nor  trigger properly.

SilentSpike:
@TheMagnetar Yeah, hoping that component can be cleared out into the others appropriately without breaking the dependency structure

PR #6309, opened 2018-04-27
Add basic medical framework documentation

DeliciousJaffa:
I've started some medical documentation based on the functions labelled as public and a var that is useful for mission makers that wasn't documented.
This documentation will probably need more adding to it.
If there's any specifics to adding the page to the listing, that also needs to be done.


jonpas:
Superseded by Medical Rewrite.

PR #6314, opened 2018-04-29
optimize suicide bomber zeus module

commy2:
- increase check delay from 0/each frame to 1 seconds (reset to 0 in debug mode for line drawing)
- use more appropriate `skillFinal`
- changed spotting distance formula to increase distance linearly starting from 0 skill instead of .5 skill
- remove some very redundant parenthesis
- re-formated if control structure for readability

### DIFF
--- a/addons/zeus/functions/fnc_moduleSuicideBomber.sqf
+++ b/addons/zeus/functions/fnc_moduleSuicideBomber.sqf
@@ -71,7 +71,7 @@ if (_autoSeek) then {
     private _memory = _unit getVariable [QGVAR(suicideBomber_memory), [nil, CBA_missionTime]];
     _memory params ["_lastMove", "_lastTime"];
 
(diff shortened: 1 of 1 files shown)


commy2:
On second thought, I don't think the math turns out to be the same. But the original one is kind of weird, I'm not sure what was intended.

Original:

| skill  | range |
| ------- | ---------------- |
| 0  | 100 |
| 0.25 | 100 |
| 0.5 | 100 |
| 0.75 | 150 |
| 1 | 200 |

This PR:

| skill  | range |
| ------- | ---------------- |
| 0  | 100 |
| 0.25 | 125 |
| 0.5 | 150 |
| 0.75 | 175 |
| 1 | 200 |

jonpas:
@mharis001

SilentSpike:
I think the maths makes more sense after this

commy2:
Also I'm pretty sure one wants to use `skillFinal`, not just `skill`.

commy2:
Done.

PR #6324, opened 2018-05-04
Zeus - Don't trigger suicide bomber if unconsicous

Vdauphin:
Not sure if it is intended but this will prevent the detonation if the suicide bomber is unconscious. From my point of view, the suicider is not supposed to activate his bomb when he is unconscious. Let me know ...

- check  if `[_unit] call EFUNC(common,isAwake)` before trigger the bomb.


Vdauphin:
I see!

SilentSpike:
However, your point is right. You can add another `exitWith` below this one that just exists the current frame and doesn't remove the PFH 👍

Vdauphin:
I was also thinking about this solution ;)

SilentSpike:
I'm probably going to extend the module to include a dead man's switch option anyway, so this will work for that once I get around to it 👍

Vdauphin:
I am pushing but do what you want with this!

dedmen:
Well if it's a dead-man switch. He would trigger when unconcious.

PR #6326, opened 2018-05-05
fix a changeProjectileDirection bug

commy2:
funny what things you find sometimes


PabstMirror:
image

`atan2` is higher than `+`, so I don't think this is needed
https://community.bistudio.com/wiki/SQF_syntax#Rules_of_Precedence

commy2:
OK, thanks. I was worried :P

PR #6330, opened 2018-05-07
Add destroyed boat ejecting

Dystopian:
- ~~add destroyed boat ejecting framework;~~
- make `Rubber Boat`, `Water Scooter`, `RHIB` eject crew if destroyed.

Some boats like rubber boat or scooter are very easy to destroy even with 1-2 bullets or if they are just collided with other object (especially by engine). It's OK except it kills its crew (Arma feature) even if it's not really possible. This PR adds `HandleDamage` event handler which ejects crew when boat is destroyed.

~~This way has at least one defect: ejected crew is not damaged with any projectile type. You can destroy boat with grenade, HE rocket or even explosive charge but crew will survive when eject. From my point of view it's not a big deal in comparison with dying from stone collision but maybe you think otherwise.~~

~~Config values are safe for 3rd-party mods and can be easily added to them. E.g. CUP has suitable Zodiac and RHIB boats.~~ Suitable CUP boats inherit vanilla boats and are fixed automatically.

I also tested solution with `allowDamage` command but it has more bugs without any advantages.


addons/vehicles/XEH_postInit.sqf
```diff
@@ -1,7 +1,42 @@
...
+// eject destroyed vehicle init
+[QGVAR(ejectDestroyed),{
+    TRACE_2("ejectDestroyed event",_this,typeOf _this);
+    _this removeEventHandler ["HandleDamage", _this getVariable QGVAR(ejectDestroyedHDEH)];
+    _this setVariable [QGVAR(ejectDestroyedHDEH), nil];
+}] call CBA_fnc_addEventHandler;
+private _ejectDestroyedClasses = [];
+{
+    private _ejectDestroyedClass = configName _x;
+    if (-1 == _ejectDestroyedClasses findIf {_ejectDestroyedClass isKindOf _x}) then {
```
SilentSpike: How does this not error with array to number comparison?

addons/vehicles/XEH_postInit.sqf
```diff
@@ -1,7 +1,42 @@
...
+                    TRACE_1("ejectDestroyed ejectDestroyedHDEH",_this);
+                    {
+                        if (alive _x) then {
+                            moveOut _x;
+                        };
+                    } forEach crew _vehicle;
+                    [QGVAR(ejectDestroyed), _vehicle] call CBA_fnc_globalEvent;
+                };
+            }];
+            _vehicle setVariable [QGVAR(ejectDestroyedHDEH), _eh];
```
SilentSpike: You don't need to store this variable, just use `_thisEventHandler` in the event to remove itself

SilentSpike:
Regarding the defect you mention, we could surely have it check for explosives (etc.) and let the units be killed in those cases.

Also not sure whether this belong here or in cookoff (as it has damage handling for other vehicles which don't actually cook off)

addons/vehicles/XEH_postInit.sqf
SilentSpike: This wasn't meant as a criticism, I'm just a curious boi 😆

My headcanon interpretation of sqf doesn't fit with this line because it would error as soon as it hit the array value after the `==`

Dystopian: `findIf` returns index or -1 if not found

addons/vehicles/XEH_postInit.sqf
Dystopian: I need it because `HandleDamage` is local and doesn't hit on all machines, so I don't have `_thisEventHandler` value everywhere

addons/vehicles/XEH_postInit.sqf
SilentSpike: Yeah I get that, I'm just so used to interpreting SQF left to right. It seems `findIf` has higher operator precedence than `==` 👍

Dystopian:
> we could surely have it check for explosives (etc.) and let the units be killed in those cases.

But we don't know how big explosion is, if it's small grenade or big satchel. Also distance is important. ATM I don't think it's worth it to check all of these parameters.

addons/vehicles/XEH_postInit.sqf
SilentSpike: 🤔 Huh? Of course you have `_thisEventHandler` everywhere, it's a block of code used in an event handler added with `addEventHandler`.

Personally I would much rather have some leftover event handlers on some machines which will remove themselves if a dead vehicle ever changes locality and takes a single hit of damage again (which, let's be honest, isn't likely). It will make the code more elegant and remove unnecessary network traffic.

addons/vehicles/XEH_postInit.sqf
dedmen: @SilentSpike Actually == has lower precedence than the rest, same as all other comparison operators
&&/and are one lower than ==
and ||/or are the lowest.

addons/vehicles/XEH_postInit.sqf
dedmen: But why do you need it? https://github.com/acemod/ACE3/pull/6330/files#diff-71d8e1c52a77144d92094e89a77b7a55R6
That?
The eventhandler and variables are removed anyway when the vehicle is destroyed aren't they?

dedmen: @SilentSpike Ref: https://github.com/acemod/ACE3/pull/6330#discussion_r186632443

dedmen:
>But we don't know how big explosion is, if it's small grenade or big satchel.
HandleDamage get's the amount of damage though. Wouldn't that be enough?

addons/vehicles/XEH_postInit.sqf
SilentSpike: Thanks dedmen, TIL 👨‍🎓

addons/vehicles/XEH_postInit.sqf
Dystopian: I have `_thisEventHandler` only where vehicle is local. EH is fired only on one machine so I have to clean up EHs on all other machines. I always thought cleanup is necessary. If you insist I can get rid of it, no problem.

> The eventhandler and variables are removed anyway when the vehicle is destroyed aren't they?

No, they aren't removed.

SilentSpike: They will clean themselves up in the event that a dead vehicle switches locality and receives more damage (again, not likely). It's totally harmless to leave them sitting there doing nothing.

Dystopian:
> HandleDamage get's the amount of damage though. Wouldn't that be enough?

Here is HDEH `[alive _vehicle, _this]` output for 30mm AP (which shouldn't kill crew) and 40mm HEDP (which should do) hits:
```
"[true,[s,"""",11.7853,p,""B_30mm_APFSDS_Tracer_Green"",-1,p,""""]]"
"[false,[s,""hull"",7.54619,p,""B_30mm_APFSDS_Tracer_Green"",0,p,""hitbody""]]"
"[false,[s,""engine"",0.919199,p,""B_30mm_APFSDS_Tracer_Green"",1,p,""hitengine""]]"

"[true,[s,"""",0.854012,p,""G_40mm_HEDP"",-1,p,""""]]"
"[true,[s,"""",10.1193,p,""G_40mm_HEDP"",-1,p,""""]]"
"[false,[s,""hull"",7.14085,p,""G_40mm_HEDP"",0,p,""hitbody""]]"
"[false,[s,""engine"",1.37552,p,""G_40mm_HEDP"",1,p,""hitengine""]]"
```
Note AP damage is more than HE damage. So looks like out of luck.

dedmen:
I guessss...... One could look into the ammo config to look for explosiveness and damage

SilentSpike:
We already do this in cook off so it's nothing new 😛

dedmen:
^ https://github.com/acemod/ACE3/blob/master/addons/cookoff/script_component.hpp#L19

Dystopian:
OK it works.
How do you think about component? If cookoff then config value will be `ace_cookoff_ejectDestroyed`. Isn't it strange?

SilentSpike:
Good point, I'm happy to leave it in vehicles 👍

Can see what other maintainers say

dedmen:
If in vehicles then maybe change https://github.com/Dystopian/ACE3/blame/master/addons/vehicles/README.md#L4
to 
>Various tweaks to vehicles and vehicle weapons.

addons/vehicles/XEH_postInit.sqf
```diff
@@ -1,7 +1,38 @@
...
+            if (!alive _vehicle) exitWith {};
+            TRACE_2("init ejectDestroyed vehicle",_vehicle,typeOf _vehicle);
+            _vehicle addEventHandler ["HandleDamage", {
+                params ["_vehicle", "", "", "", "_ammo"];
+                if (!alive _vehicle) then {
+                    TRACE_2("ejectDestroyed HDEH",typeOf _vehicle,_this);
+                    if (0.5 >= getNumber (configFile >> "CfgAmmo" >> _ammo >> "explosive")) then {
+                        {
+                            if (alive _x) then {
+                                moveOut _x;
```
dedmen: wasn't there a problem that moveOut didn't work with dead bodies? Or was that fixed recently?

addons/vehicles/XEH_postInit.sqf
```diff
@@ -1,7 +1,38 @@
...
+    if (-1 == _ejectDestroyedClasses findIf {_ejectDestroyedClass isKindOf _x}) then {
+        TRACE_1("init ejectDestroyed class",_ejectDestroyedClass);
+        _ejectDestroyedClasses pushBack _ejectDestroyedClass;
+        [_ejectDestroyedClass, "init", {
+            params ["_vehicle"];
+            if (!alive _vehicle) exitWith {};
+            TRACE_2("init ejectDestroyed vehicle",_vehicle,typeOf _vehicle);
+            _vehicle addEventHandler ["HandleDamage", {
+                params ["_vehicle", "", "", "", "_ammo"];
+                if (!alive _vehicle) then {
```
dedmen: `if (alive _vehicle) exitWith`?
get's rid of indentation (readability)
and get's rid of the `!` (performance)

addons/vehicles/XEH_postInit.sqf
Dystopian: but I don't move out them ^^ `if (alive _x)`.
Or you want to move out them also? I don't think it's logical.

dedmen: lol duh... Must read. Read good for thinking.
I was thinking about unconscious units but this already handles that fine. Then my brain moved from unconscious to dead.

addons/vehicles/XEH_postInit.sqf
```diff
@@ -1,7 +1,38 @@
 // by esteldunedain
 #include "script_component.hpp"
 
+// eject destroyed vehicle init
+private _ejectDestroyedClasses = [];
+{
+    private _ejectDestroyedClass = configName _x;
+    if (-1 == _ejectDestroyedClasses findIf {_ejectDestroyedClass isKindOf _x}) then {
```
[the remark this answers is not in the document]
Dystopian: well ACE uses some of such config loops and there were no problems yet AFAIK.

PabstMirror:
Current method doesn't check inheritance, e.g. it still runs on this child:
```
    class C_Boat_Transport_02_F: Boat_Transport_02_base_F {
        GVAR(ejectDestroyed) = 0;
    };
```
`{} forEach ("1 == getNumber (_x >> ""ace_vehicles_ejectDestroyed"")" configClasses (configFile >> "CfgVehicles"));` is about 25ms with rhs+cup

I think adding like this will be cleaner and faster.
```
    class Boat_Transport_02_base_F {
        class GVAR(ejectDestroyed) {
            init = QUOTE(call FUNC(ejectDestroyed));
        };
    };    
    class Rubber_duck_base_F {
        class GVAR(ejectDestroyed) {
            init = QUOTE(call FUNC(ejectDestroyed));
        };
    };
```
We could even add RHS vehicles directly in here as there is no load order/inheritance problems with xeh.

dedmen:
Do RHS vehicles still have the same issue?

Also please stop recompiling the whole eventhandler code on every shot that hits the boat.

Dystopian:
> Current method doesn't check inheritance, e.g. it still runs on this child

It's intended because I didn't find suitable classes to exclude. I didn't want to add unnecessary logic.

> I think adding like this will be cleaner and faster

I agree with `faster`. But I didn't agree with `cleaner`. Until I have known that CUP boats inherit from `Rubber_duck_base_F` :laughing:. So now I don't have real examples of 3rd-party addons which need this PR. If anybody know some or think this PR *should* rely on config value - say. Otherwise I'll rewrite PR like @PabstMirror says. Also in this case I don't object moving code to cookoff.

> Do RHS vehicles still have the same issue?

RHS don't have suitable boats.

> Also please stop recompiling the whole eventhandler code on every shot that hits the boat.

will do :ok_hand:

dedmen:
Found a solution for the config scan problem.
Do it like ASDG_JR. Just make a global class where everyone enters
`classname = 1`
That way we only scan over that.

Dystopian:
@dedmen really good solution. Could be used if it's needed.

Dystopian:
hmm working on static weapons - they should be handled another way

Dystopian:
Static weapons don't fire HDEH when become dead and can be alive if `_damage == 0.950505`. I don't want to mess with them. I think this PR is done.

jonpas:
Needs testing if this still happens with medical, specifically on boats.

PabstMirror:
This was needed on old medical, but now units won't take damage from vehicle deaths.



Pilot doing fine and staying on a sinking ship 😄 

We actually need to do work on making vehicle crashes/dying more dangerous to units, keeping in mind to not just always make it lethal

Dystopian:
I don't completely understand why you link unit HD and vehicle HD. This PR is about vehicle HD only. It was made with disabled ACE Medical at all. The purpose was just to eject unit from shot **rubber** boat.
If you think this is OK:
> Pilot doing fine and staying on a sinking ship smile

then I can add somewhat like `!IS_MOD_LOADED(medical)`.

PabstMirror:
Merging now, will keep this in mind while working on #7305

PR #6334, opened 2018-05-09
Only switch side if necessary on loading

### LINKED ISSUE #6333: Load/unload person function will broke slot if loaded person will disconnect
**Arma 3 Version:** `1.82` (stable)
**CBA Version:** `3.6.1` (old stable)
**ACE3 Version:** `3.12.2` (stable)

**Description:**
- If person loaded by `[bob, kevin, vehicle] call ace_common_fnc_loadPerson;` was disconnected from server (manually/net drop/game crash) then his bot is broken and on reconnect person will go to spectator.

**Steps to reproduce:**
- 2 players 1 vehicle.
- load person with ace_common_fnc_loadPerson
- loaded person need to go to lobby and toggle slot in/out or reconnect to server entirely
- welcome to spectator

- Dedicated. May be on all MP.

**Placed Modules:**
- none

Seems like loaded person change group. Then on dissconnect arma thinks that group is empty and delete it.
Bot without group = dead unit. That is why user go to spectator. (May be I'm wrong)

BaerMitUmlaut:
- When loading units into vehicles, only switch side if
  - vehicle is empty and *loader* has a different side than the target or
  - vehicle is crewed and *vehicle* has a different side than the target
- Fix #6333 

Previously the unit would always switch to the side of the person loading the unit into the vehicle. That most likely did also not work when the loader has a different side than the crew of the vehicle. This *should* cover all cases.

Needs multiplayer testing.

### DIFF
--- a/addons/common/functions/fnc_loadPerson.sqf
+++ b/addons/common/functions/fnc_loadPerson.sqf
@@ -29,7 +29,15 @@ if (isNull _vehicle) then {
 };
 
 if (!isNull _vehicle) then {
-    [_unit, true, GROUP_SWITCH_ID, side group _caller] call FUNC(switchToGroupSide);
+    switch (true) do {
+        case ((crew _vehicle isEqualTo []) && {side _caller != side _unit}): {
+            [_unit, true, GROUP_SWITCH_ID, side _caller] call FUNC(switchToGroupSide);
+        };
+        case (side _vehicle != side _unit): {
+            [_unit, true, GROUP_SWITCH_ID, side _vehicle] call FUNC(switchToGroupSide);
+        };
+    };
+
     ["ace_loadPersonEvent", [_unit, _vehicle, _caller], _unit] call CBA_fnc_targetEvent;
 };
 


addons/common/functions/fnc_loadPerson.sqf
```diff
@@ -29,7 +29,15 @@ if (isNull _vehicle) then {
...
 if (!isNull _vehicle) then {
-    [_unit, true, GROUP_SWITCH_ID, side group _caller] call FUNC(switchToGroupSide);
+    switch (true) do {
+        case ((crew _vehicle isEqualTo []) && {side _caller != side _unit}): {
+            [_unit, true, GROUP_SWITCH_ID, side _caller] call FUNC(switchToGroupSide);
+        };
+        case (side _vehicle != side _unit): {
+            [_unit, true, GROUP_SWITCH_ID, side _vehicle] call FUNC(switchToGroupSide);
+        };
+    };
```
commy2: `side OBJECT` will report stuff like sideEnemy and sideFriendly. `side GROUP` will only report west, east, resistance, civilian. Are you sure you don't want to keep `side group _something`?

BaerMitUmlaut: Oh yes, my bad. I'll fix it.

severgun: So will it be fixed and merged? We playing with this "fixed fix" for more than a year. But it is pain in the ass to track our temporary fixes on every ACE release. Would be nice to see solved issue in next release.

jonpas:
Still relevant I assume? Looks like just needs a bit of a move.

Issue #6186, opened 2018-03-07
sqf_validator.py fails on remoteExec variant


bux:
The `sqf_validator.py` script fails when encountering a certain variant of `remoteExec`.

https://community.bistudio.com/wiki/remoteExec#Examples
Example 8
`{player setAmmo [primaryWeapon player, 1];} remoteExec ["bis_fnc_call", 0];`

When validating this will throw the following error:
`ERROR: Possible missing semi-column ';' detected` (_also note the typo -> semicolon_)

commy2:
Why not put two newlines? It would look better  thaz way.

bux:
It doesn't matter to the validator.

PR #6338, opened 2018-05-10
Support # command in SQF Validator 

jonpas:
- Support `#` command in SQF Validator (instead of ignore because we don't care about preprocessor directives)
- Will miss missing semi-colons on actual `#` command


bux:
if you're on it: https://github.com/acemod/ACE3/issues/6186

PR #6345, opened 2018-05-16
Attach - Optimize creating children actions

mharis001:
- Optimize function that returns children actions using `CBA_fnc_uniqueUnitItems` and `arrayIntersect`
- Rename the function: `ace_attach_fnc_getChildrenAttachActions` -> `ace_attach_fnc_getChildrenActions`


addons/attach/functions/fnc_getChildrenAttachActions.sqf
```diff
@@ -1,52 +0,0 @@
...
-        _listed pushBack _x;
-        private _item = ConfigFile >> "CfgWeapons" >> _x;
-        if (getText (_item >> "ACE_Attachable") != "") then {
-            private _displayName = getText(_item >> "displayName");
-            private _picture = getText(_item >> "picture");
-            private _action = [_x, _displayName, _picture, {[{_this call FUNC(attach)}, _this] call CBA_fnc_execNextFrame}, {true}, {}, [_x]] call EFUNC(interact_menu,createAction);
-            _actions pushBack [_action, [], _target];
-        };
-    };
-} forEach (items _player);
```
dedmen: Oh FFS the old script iterated through `items`?! wtf.

commy2:
>One minor further improvement would be to move configFile >> "CfgMagazines" and configFile >> "CfgWeapons" into local variables outside the loop to spare a bit of runtime.

If you go that far, why not put these into globals at preInit? Guess that's something to consider for A4 if nothing significant changes.

dedmen:
>If you go that far, why not put these into globals at preInit?

Because at that point the overhead of global var vs local var makes it not worth it anymore.

Dystopian:
`git mv` please :(

PR #6346, opened 2018-05-16
Fix minor typo in fnc_cachedCall

dedmen:
Sorry. But I just couldn't stop thinking about it.
Also as I see _event is a string.. Shouldn't that default to empty string and test for empty string? as users might pass in empty string to show "no event"


SilentSpike:
I hate comments that are questions 😛

dedmen:
@SilentSpike make another PR :D

PR #6348, opened 2018-05-16
Med menu zeus

mgStingray:
- Allow Zeus to open the medical menu on any unit in the mission via `ace_medical_menu_fnc_openMenu`
  - Adds a condition check to `fnc_canOpenMenu` that returns true if the player has the curator interface open
  - Adds exception to `fnc_onMenuOpen` that allows `fnc_canInteractWith` to return true if player is zeus.

**Rationale**
- I'd like to allow game master to better assess damage using the medical menu on players in the mission. This could either be done via the function call listed above or in a zeus module to be developed later.

### DIFF
--- a/addons/interaction/ACE_ZeusActions.hpp
+++ b/addons/interaction/ACE_ZeusActions.hpp
@@ -37,8 +37,7 @@ class ACE_ZeusActions {
         };
 
         class medicalMenu {
-          displayName = "Open Medical Menu"; // TODO - localize
-          //icon = QPATHTOEF(medical_menu\data\icons\triage_card_small.paa);
+          displayName = ECSTRING(Medical_Menu,OpenMenu);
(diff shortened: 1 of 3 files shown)


jonpas:
Won't this break if you try to use anything inside the medical menu? I'd rather show the menu that's used in 3D interaction mode, you can't click anything there and it's smaller.

mgStingray:
Have a preference? Interaction/module? I think I like zeus interaction better as modules can be clunky to drop on correct unit.

mgStingray:
I've added a zeus interaction. per @SilentSpike's comment. I had to add a `CBA_fnc_waitUntilAndExecute` to allow for the interaction display (idd 919191) to close. I believe it not being fully closed was preventing the medical menu from opening. 

image

mgStingray:
I just realized I never localized display name for the interaction. This has been fixed in last commit. Any outstanding issues with this PR that I can address?

jonpas:
Replaced by #9019.

PR #6350, opened 2018-05-17
Improve checking of unit items

mharis001:
- Add `uniqueItems` common function - same concept as `CBA_fnc_uniqueUnitItems` however, caches items if unit is `ACE_player` (improves performance when checking units with lots of items)
- Optimize `ace_dogtags_fnc_addDogtagActions` by using `uniqueItems`
- Optimize `ace_explosives_fnc_getDetonators` by using `uniqueItems`

Performance differences for `ace_dogtags_fnc_addDogtagActions`:

Items | Old | New
--- | --- | ---
10 | 0.0683 ms | 0.0826 ms
25 | 0.122249 ms | 0.082 ms
50 | 0.212993 ms | 0.0867 ms
75 | 0.308928 ms | 0.0887 ms
100 | 0.400641 ms | 0.0924 ms
300 | 1.13636 ms | 0.0978 ms

Performance differences for `ace_explosives_fnc_getDetonators`:

Items | Old | New
--- | --- | ---
5 | 0.0532 ms | 0.0945 ms
50 | 0.189609 ms | 0.107308 ms
100 | 0.335233 ms | 0.123047 ms
200 | 0.626959 ms | 0.132293 ms


mharis001:
Based on slack discussion with @dedmen , unsure whether checking `"my_item" in items _unit` should be changed as well.

dedmen:
>unsure whether checking "my_item" in items _unit should be changed as well. 

replacing `items` has a noticeably positive effect once you go over about 100 item's I'd say. Question is just if we think that the average user carries more than 100 items.

addons/common/functions/fnc_uniqueItems.sqf
```diff
@@ -0,0 +1,38 @@
...
+ * Public: No
+ */
+#include "script_component.hpp"
+
+#define ITEMS_CACHE_TIME 3600
+
+params ["_unit"];
+
+private _fnc_getItems = {
+    params ["_unit"];
```
dedmen: That is just the CBA thing without assignedItems right? https://github.com/CBATeam/CBA_A3/pull/922 add's a parameter to disable assignedItems. But this specific implementation in here will be faster. Than the CBA one with it's conditionals.

Btw you don't need `params ["_unit"]` the local variable implicitly carries over.

addons/explosives/functions/fnc_getDetonators.sqf
```diff
@@ -1,25 +1,21 @@
...
 
 params ["_unit"];
-TRACE_1("params",_unit);
+TRACE_1("Getting detonators",_unit);
 
-private _result = (items _unit) select {getNumber (ConfigFile >> "CfgWeapons" >> _x >> QGVAR(Detonator)) == 1};
-_result = _result arrayIntersect _result;
-
-_result
+(_unit call EFUNC(common,uniqueItems)) select {getNumber (configFile >> "CfgWeapons" >> _x >> QGVAR(Detonator)) == 1};
```
dedmen: here you can again move the `configFile >> "CfgWeapons"` outside the loop.
How often does this function run? That's a interaction menu condition right? so probably worth it.

mharis001: Keep forgetting to do that..

addons/common/functions/fnc_uniqueItems.sqf
mharis001: And you are right about `_unit`, so I guess `cachedCall` doesn't need anything passed to it either.

mharis001: And you are right about `_unit`, i guess just pass `0` to `cachedCall`

Monsen007:
I'm sorry but speak no English ..: / only German .. can you tell me somehow where I have to replace it so where exactly? would be helpful because I do not see it at the janzen ace because it is the steam version


ps translator: P
________________________________
Von: mharis001 <notifications@github.com>
Gesendet: Donnerstag, 30. August 2018 03:19:04
An: acemod/ACE3
Cc: Subscribed
Betreff: Re: [acemod/ACE3] Improve checking of unit items (#6350)


Replacing the cachedCall part (0.0118 ms):

[_unit, _fnc_getItems, _unit, QGVAR(uniqueItemsCache), ITEMS_CACHE_TIME, "cba_events_loadoutEvent"] call FUNC(cachedCall);

with something like this (0.0065 ms):

private _items = _unit getVariable QGVAR(uniqueItemsCache);
if (isNil "_items") then {
    _items = call _fnc_getItems;
    _unit setVariable [QGVAR(uniqueItemsCache), _items];
};
+_items

and a loadout event added in postInit to clear the cache seems worth it, considering we are going for overkill.

This was tested with a low item count (4 unique items), the situation in which the items (~0.0013 ms) command is faster than this solution. This change would help decrease the difference.

And for similar reasons, I think _fnc_getItems should be used rather than CBA_fnc_uniqueUnitItems.

dedmen:
Keep in mind that loadout event only works for local player. the `_unit` there makes it look like it works everywhere

mharis001:
With no caching at all (4 items): `_fnc_getItems` (0.0085 ms) vs `[_unit, false, true, true, true, false] call CBA_fnc_uniqueUnitItems` (0.0241 ms), about 3x faster.

I think the `uniqueItems` function is as optimized as it will get now, just have to go through and replace any `items _player` that are left.

mharis001:
I think this PR is complete (replaced most/all `items` command uses), ready for review.

PR #6360, opened 2018-05-22
Fix AF sway factor

### LINKED ISSUE #6359: Empty swayFactor in module advanced fatigue
**Arma 3 Version:** `x.xx` (stable / rc / dev)
**CBA Version:** `3.x.x` (stable / dev + commit hash)
**ACE3 Version:** `3.12.2` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- The swayFactor in advanced fatigue module was never assigned a value, which leads to very weak weapon sway effect.

- Singleplayer / Editor (Singleplayer) tested, but presumably, it will occur everywhere

**Placed Modules:**
- advanced fatigue

BaerMitUmlaut:
- Fix #6359 by switching to CBA settings
- Update slider ranges to the same ranges as the ones from ACE settings
- Switch to stringtable entry for setting category, add ACE prefix to it


addons/advanced_fatigue/ACE_Settings.hpp
```diff
@@ -1,49 +0,0 @@
-class ACE_Settings {
```
mharis001: movedToSQF = 1 for settings moved to use CBA_settings_fnc_init?

BaerMitUmlaut: @jonpas do we still provide BWC in 3.13?

PR #6362, opened 2018-05-22
Update development environment documentation for armake

jonpas:
- Update documentation for setup and building with armake.

Currently just split from #3747, as using armake in ACE3 is still experimental. Will probably be rewritten as an additional section to the current wiki and only fully replaced when we start using armake only.

Important notes:
- Binarization only fully works on Windows using BIS Binarize!
- Current (from #3747) Makefile builds optionals the old way (not as launchable mod folders) - #6363


jonpas:
Superseded by HEMTT #6900.

PR #6364, opened 2018-05-23
Cargo - Support of object without `ace_cargo_size` defined in config or `-1` config value

### LINKED ISSUE #6357: Offest for large containers too small when being unloaded from trucks
Arma 3 Version: 1.82 stable
CBA Version: 3.7.0
ACE3 Version: current to Steam WS

**Mods:**
- CBA_A3
- ace

**Description:**
- we have large containers (Taru and Huron pods) loadable into any kind of big truck (HEMMT, Zamak etc), when being unloaded they are put too close to the person unloading them or the truck causing critical damage to the person / the container / the vehicle

**Steps to reproduce:**
- set cargo size for containers as well as the adequate capacity to the trucks
- load container into truck -> unload container

- Dedicated / Editor (Singleplayer) / Editor (Multiplayer)

**Placed Modules:**
- None

short video
https://youtu.be/t5-_39Krvqs?t=43

Vdauphin:
close #6357 

Actually, if `_typeofcargo` does not have `ace_cargo_size` define in config, the `_radiusOfItem` of the object is `1`. This inconsistency can be avoid by using the `ace_cargo_fnc_getSizeItem` which take an object as parameter and get the size of the object defined manually by user ( `ace_cargo_fnc_setsize`). 

By having a `_radiusOfItem` more accurate, the `ace_common_fnc_findUnloadPosition` will find better position for unloading and avoid vehicle explosion when big objects, without `ace_cargo_size` defined in config, are unloaded.

I don't know if it is the best way to handle this case but at least it is a work around. Cheers.

- `ace_common_fnc_findUnloadPosition` can take object as parameter
- Prefer the config way instead to determine the size of `_cargo` directly with  `ace_cargo_fnc_getSizeItem` (this avoid inconsistency when player `ace_cargo_fnc_setsize` of an object lower from the `ace_cargo_size` defined in config)
- Also check if the config value is not `-1` to avoid the issue: https://github.com/acemod/ACE3/issues/6357#issuecomment-393525374
- If `ace_cargo_size` is not define in config, use the `ace_cargo_fnc_getSizeItem` to get object size set by player with `ace_cargo_fnc_setsize`
- Update the `ace_cargo_fnc_unloadItem` accordingly (use the cargo object instead of the class name).

### DIFF
--- a/addons/common/functions/fnc_findUnloadPosition.sqf
+++ b/addons/common/functions/fnc_findUnloadPosition.sqf
@@ -48,10 +48,10 @@ if (_cargo isKindOf "CAManBase") then {
     private _itemSize = if (isNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) then {
         getNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size));
     } else {
-        if (isNil QEFUNC(cargo,getSizeItem)) then {
-            _radiusOfItem;
-        } else {
+        if (["ace_cargo"] call EFUNC(common,isModLoaded)) then {
             [_cargo] call EFUNC(cargo,getSizeItem);
+        } else {
+            _radiusOfItem;
         };
     };
     if !(_itemSize isEqualTo -1) then {
--- a/addons/common/functions/fnc_findUnloadPosition.sqf
+++ b/addons/common/functions/fnc_findUnloadPosition.sqf
@@ -5,7 +5,7 @@
  *
  * Arguments:
  * 0: Source Vehicle <OBJECT>
- * 1: Cargo Classname <STRING>
+ * 1: Cargo <OBJECT> or <STRING>
  * 2: Unloader (player) <OBJECT> (default: objNull)
  * 3: Max Distance (meters) <NUMBER> (default: 10)
  * 4: Check Vehicle is Stable <BOOL> (default: true)
@@ -27,8 +27,8 @@
 //Manual collision tests (count and radius):
 #define COL_TEST_COUNT 12
 
-params ["_vehicle", "_typeOfCargo", ["_theUnloader", objNull], ["_maxDistance", 10], ["_checkVehicleIsStable", true]];
-TRACE_5("params",_vehicle,_typeOfCargo,_theUnloader,_maxDistance,_checkVehicleIsStable);
+params ["_vehicle", "_cargo", ["_theUnloader", objNull], ["_maxDistance", 10], ["_checkVehicleIsStable", true]];
+TRACE_5("params",_vehicle,_cargo,_theUnloader,_maxDistance,_checkVehicleIsStable);
 
 scopeName "main";
 
@@ -40,12 +40,18 @@ if (_checkVehicleIsStable) then {
 };
 
 private _radiusOfItem = 1;
-if (_typeOfCargo isKindOf "CAManBase") then {
+if (_cargo isKindOf "CAManBase") then {
     _radiusOfItem = 1.1;
 } else {
     //`sizeOf` is unreliable, and does not work with object types that don't exist on map, so estimate size based on cargo size
-    if (isNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) then {
-        _radiusOfItem = (((getNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) ^ 0.35) max 0.75);
+    private _typeOfCargo = if (_cargo isEqualType "") then {_cargo} else {typeOf _cargo};
+    private _itemSize = if (isNumber (configFile >> "CfgVehicles" >> [_typeOfCargo] select (_cargo isEqualType "") >> QEGVAR(cargo,size))) then {
+        getNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size));
+    } else {
+        [_cargo] call EFUNC(cargo,getSizeItem);
+    };
+    if !(_itemSize isEqualTo -1) then {
+        _radiusOfItem = (_itemSize ^ 0.35) max 0.75;
     };
 };
 
--- a/addons/common/functions/fnc_findUnloadPosition.sqf
+++ b/addons/common/functions/fnc_findUnloadPosition.sqf
@@ -48,7 +48,7 @@ if (_cargo isKindOf "CAManBase") then {
     private _itemSize = if (isNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) then {
         getNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size));
     } else {
-        [_cargo] call EFUNC(cargo,getSizeItem);
+        [[_cargo] call EFUNC(cargo,getSizeItem)] param [0, _radiusOfItem];
     };
     if !(_itemSize isEqualTo -1) then {
         _radiusOfItem = (_itemSize ^ 0.35) max 0.75;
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_findUnloadPosition.sqf
```diff
@@ -40,12 +40,18 @@ if (_checkVehicleIsStable) then {
...
     _radiusOfItem = 1.1;
 } else {
     //`sizeOf` is unreliable, and does not work with object types that don't exist on map, so estimate size based on cargo size
-    if (isNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) then {
-        _radiusOfItem = (((getNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) ^ 0.35) max 0.75);
+    private _typeOfCargo = if (_cargo isEqualType "") then {_cargo} else {typeOf _cargo};
+    private _itemSize = if (isNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) then {
+        getNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size));
+    } else {
+        [_cargo] call EFUNC(cargo,getSizeItem);
```
Dystopian: there is error here if cargo module is no loaded

PabstMirror:
For safety we could take the `max` of both numbers

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: `[[_cargo] call EFUNC(cargo,getSizeItem)] param [0, _radiusOfItem];` could resolve that issue.

Dystopian: I meant if cargo module is not loaded there is no `EFUNC(cargo,getSizeItem)` and you call non-existing function

Vdauphin: I just learned in scheduled EFUNC(cargo,getSizeItem) will trow an undefined variables which is not the case in unscheduled.
If you guys are using an other approach, let me know, I will use it for consistency.

addons/common/functions/fnc_findUnloadPosition.sqf
```diff
@@ -40,12 +40,22 @@ if (_checkVehicleIsStable) then {
...
 private _radiusOfItem = 1;
-if (_typeOfCargo isKindOf "CAManBase") then {
+if (_cargo isKindOf "CAManBase") then {
     _radiusOfItem = 1.1;
 } else {
     //`sizeOf` is unreliable, and does not work with object types that don't exist on map, so estimate size based on cargo size
-    if (isNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) then {
-        _radiusOfItem = (((getNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) ^ 0.35) max 0.75);
+    private _typeOfCargo = if (_cargo isEqualType "") then {_cargo} else {typeOf _cargo};
+    private _itemSize = if (isNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size))) then {
```
Dystopian: this code piece already exists in `EFUNC(cargo,getSizeItem)`, so can be optimized well

addons/common/functions/fnc_findUnloadPosition.sqf
```diff
@@ -40,12 +40,22 @@ if (_checkVehicleIsStable) then {
...
+        getNumber (configFile >> "CfgVehicles" >> _typeOfCargo >> QEGVAR(cargo,size));
+    } else {
+        if (["ace_cargo"] call EFUNC(common,isModLoaded)) then {
```
Dystopian: `EFUNC(common,isModLoaded)` -> `FUNC(isModLoaded)`

addons/common/functions/fnc_findUnloadPosition.sqf
```diff
@@ -40,12 +40,22 @@ if (_checkVehicleIsStable) then {
...
+            [_cargo] call EFUNC(cargo,getSizeItem);
+        } else {
+            _radiusOfItem;
+        };
+    };
+    if !(_itemSize isEqualTo -1) then {
```
Dystopian: just `_itemSize != -1`

addons/common/functions/fnc_findUnloadPosition.sqf
```diff
@@ -40,12 +40,22 @@ if (_checkVehicleIsStable) then {
...
+        _radiusOfItem = (_itemSize ^ 0.35) max 0.75;
```
Dystopian: most significant change: before this PR if cargo module is not loaded `_radiusOfItem` was 1 and now it becomes `(1 ^ 0.35) max 0.75`.

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: I did a choice here as explain in the PR description :
> Prefer the config way instead to determine the size of `_cargo` directly with  `ace_cargo_fnc_getSizeItem` (this avoid inconsistency when player `ace_cargo_fnc_setsize` of an object lower from the `ace_cargo_size` defined in config)

I can add a parameter in `EFUNC(cargo,getSizeItem)` to prefer the config if exist or prefer varaible if exist

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: `EFUNC(common,isModLoaded)` -> ace_common_fnc_isModload
`FUNC(isModLoaded)` -> ace_cargo_fnc_isModload

Because the second doesn't exist I am using the first one as suggested by @jonpas  here: https://github.com/acemod/ACE3/pull/6364#discussion_r190597398

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: With `isEqualTo` you can compare  different type like: `nil isEqualTo -1` so the function will never fail
> It doesn't throw error when comparing different types, i.e. ("eleven" isEqualTo 11)
But, if ace use it as convention I will change it.

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: Ahah! I didn't know we could make joke in review!
Thanks a lot for the time you take to review this :-), I appreciate your feedback!

addons/common/functions/fnc_findUnloadPosition.sqf
Dystopian: here we are in `common` module (`addons/common/functions/fnc_findUnloadPosition`) so FUNC is `ace_common_fnc`

addons/common/functions/fnc_findUnloadPosition.sqf
Dystopian: `_itemSize` here is always number

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: `(1 ^ 0.35) max 0.75 = 1 max 0.75 = 1`

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: I am supposing that `ace_cargo_size` in config is correct so if I am calling `EFUNC(cargo,getSizeItem)` and if player has `ace_cargo_fnc_setsize` an other value (a smaller one for exemple) the function will return a bad position base on a wrong value changed by player. This avoid that case and also cover the objective of this PR where `ace_cargo_size` in config is not defined.

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: Sorry I thought you were ironical :/

addons/common/functions/fnc_findUnloadPosition.sqf
Vdauphin: woups!!!

PabstMirror:
looks good, it does require that people set reasonable cargo sizes for items they load

PR #6368, opened 2018-05-26
Pass functions by reference to BI event handlers

mharis001:
- Pass functions by reference to `addEventHandler` and `addMissionEventHandler`
Should provide a slight performance improvement


mharis001:
Also, the formatting on the [coding guidelines](https://ace3mod.com/wiki/development/coding-guidelines.html#73-event-driven) for this is weird

654wak654:
Isn't this what the `LINKFUNC` macro already does?

dedmen:
https://github.com/acemod/ACE3/blob/81e02a73369d341ef21d0b8fc1ad9c62e1688f2e/addons/main/script_debug.hpp#L7-L15

commy2:
Btw, `LINKFUNC()` should be changed to `{call FUNC()}`.

PR #6370, opened 2018-05-26
Declare ACE_player in a local scope

Wakbub:
- Declare `ACE_player` in a local scope.

**Motivation:**
- Speed.
- Resilience.
- Consistent with code in [other functions](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_headBugFix.sqf).

**Notes:**
- Repeated `ACE_player` global lookups can also be seen in other functions, but this function has most of them.

### DIFF
--- a/addons/common/functions/fnc_switchPersistentLaser.sqf
+++ b/addons/common/functions/fnc_switchPersistentLaser.sqf
@@ -1,5 +1,5 @@
 /*
- * Author: Dystopian
+ * Author: Dystopian, optimized by Anton
  * Controls persistent laser state.
  *
  * Arguments:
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_switchPersistentLaser.sqf
```diff
@@ -28,16 +28,17 @@ if (!_enabled) exitWith {
...
 GVAR(laserKeyDownEH) = ["KeyDown", {
     if !((_this select 1) in actionKeys "headlights") exitWith {false};
-    private _weapon = currentWeapon ACE_player;
+    private _unit = ACE_player;
+    private _weapon = currentWeapon _unit;
     [
         {
             params ["_weapon", "_laserWasEnabled"];
-            private _laserEnabled = ACE_player isIRLaserOn _weapon || {ACE_player isFlashlightOn _weapon};
+            private _laserEnabled = _unit isIRLaserOn _weapon || {_unit isFlashlightOn _weapon};
```
PabstMirror: I think `_unit` will be undefined here (inside of a CBA_fnc_execNextFrame call)

Wakbub:
You are probably right as you are one of the authors to `CBA_fnc_execNextFrame`.

PR #6381, opened 2018-06-01
Pylons - Update configure pylons action

mharis001:
- Update configure pylons action to use `addActionToClass` instead of `addActionToObject` with `CBA_fnc_addClassEventHandler`
- Only use `createAction` once
- Update action condition to use `findIf` instead of `count` for checking rearm vehicles (should provide small performance benefit)

### DIFF
--- a/addons/pylons/XEH_postInit.sqf
+++ b/addons/pylons/XEH_postInit.sqf
@@ -1,34 +1,38 @@
 #include "script_component.hpp"
 
-private _filter = "isClass (_x >> 'Components' >> 'TransportPylonsComponent') && {(getNumber (_x >> 'scope')) > 0}";
-GVAR(aircraftWithPylons) = (_filter configClasses (configFile >> "CfgVehicles")) apply {configName _x};
-{
-    [_x, "init", {
-        params ["_aircraft"];
+GVAR(aircraftWithPylons) = [];
 
-        private _loadoutAction = [
-            QGVAR(loadoutAction),
-            localize LSTRING(ConfigurePylons),
-            "",
-            {[_target] call FUNC(showDialog)},
-            {
-                if (!GVAR(enabledFromAmmoTrucks)) exitWith {false};
+GVAR(loadoutAction) = [ // create action
+    QGVAR(loadoutAction),
+    localize LSTRING(ConfigurePylons),
+    "",
+    {[_target] call FUNC(showDialog)},
+    {
+        if (!GVAR(enabledFromAmmoTrucks)) exitWith {false};
 
-                private _vehicles = nearestObjects [_target, ["Air", "LandVehicle", "Slingload_base_F", "ReammoBox_F"], GVAR(searchDistance) + 10];
-                private _isRearmVehicle = if (["ace_rearm"] call EFUNC(common,isModLoaded)) then {
-                    _vehicles findIf {[_x] call EFUNC(rearm,isSource)} != -1;
-                } else {
-                    private _cfgVehicle = configFile >> "CfgVehicles";
-                    _vehicles findIf {getNumber (_cfgVehicle >> typeOf _x >> "transportAmmo") > 0} != -1;
-                };
-                
-                (_isRearmVehicle && {[ace_player, _target] call FUNC(canConfigurePylons)})
-            }
-        ] call EFUNC(interact_menu,createAction);
+        private _vehicles = nearestObjects [_target, ["Air", "LandVehicle", "Slingload_base_F", "ReammoBox_F"], GVAR(searchDistance) + 10];
+        private _isRearmVehicle = if (["ace_rearm"] call EFUNC(common,isModLoaded)) then {
+            _vehicles findIf {[_x] call EFUNC(rearm,isSource)} != -1;
+        } else {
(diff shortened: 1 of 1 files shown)


mharis001:
@654wak654  Not sure if there was a special reason for using `addActionToObject` with init EH, but since it doesn't use inheritance it seems clearer to use `addActionToClass` and avoid the class EH.

addons/pylons/XEH_postInit.sqf
```diff
@@ -2,28 +2,26 @@
 
 private _filter = "isClass (_x >> 'Components' >> 'TransportPylonsComponent') && {(getNumber (_x >> 'scope')) > 0}";
 GVAR(aircraftWithPylons) = (_filter configClasses (configFile >> "CfgVehicles")) apply {configName _x};
+
+GVAR(loadoutAction) = [
```
Dystopian: why global not local?

mharis001: oversight, fixed

654wak654:
@mharis001 Don't remember any special reason for using `addActionToObject`. It's probably just a decision I made while sleepy 😃.

PabstMirror:
I don't really like how we're running `addActionToClass` on all possible classes (99% of witch won't exist in the mission).
With RHS+CUP `count ace_pylons_aircraftWithPylons = 316` and the section takes 200ms.

I think it would be better to just check aircraft on init and add it then
```
GVAR(aircraftWithPylons) = [];

GVAR(loadoutAction) = [
    QGVAR(loadoutAction),
    localize LSTRING(ConfigurePylons),
    "",
    {[_target] call FUNC(showDialog)},
    {
        if (!GVAR(enabledFromAmmoTrucks)) exitWith {false};

        private _vehicles = nearestObjects [_target, ["Air", "LandVehicle", "Slingload_base_F", "ReammoBox_F"], GVAR(searchDistance) + 10];
        private _filter = ["transportAmmo", QEGVAR(rearm,defaultSupply)] select (["ace_rearm"] call EFUNC(common,isModLoaded));
        private _cfgVehicles = configFile >> "CfgVehicles";
        private _rearmFound = _vehicles findIf {getNumber (_cfgVehicles >> typeOf _x >> _filter) > 0};

        _rearmFound > -1 && {[ACE_player, _target] call FUNC(canConfigurePylons)}
    }
] call EFUNC(interact_menu,createAction);

["Air", "init", {
    params ["_vehicle"];
    private _typeOf = typeOf _vehicle;

    if (_typeOf in GVAR(aircraftWithPylons)) exitWith {};
    if (!isClass (configFile >> "CfgVehicles" >> _typeOf >> 'Components' >> 'TransportPylonsComponent')) exitWith {};

    GVAR(aircraftWithPylons) pushBack _typeOf;
    [_typeOf, 0, ["ACE_MainActions"], GVAR(loadoutAction)] call EFUNC(interact_menu,addActionToClass);
}, true, ["ParachuteBase"], true] call CBA_fnc_addClassEventHandler;
```

Dystopian:
@PabstMirror your code makes vehicle init a little slower. Not sure what is better: slow mission init or little slower vehicle init.

dedmen:
^ Slow mission init. half a second longer loading screen is not a problem. But lags when placing vehicles are baaad.

PabstMirror:
```
private _filter = "isClass (_x >> 'Components' >> 'TransportPylonsComponent') && {(getNumber (_x >> 'scope')) == 2}";
pTypes = (_filter configClasses (configFile >> "CfgVehicles")) apply {configName _x};
{
    _x createVehicle [0,50 * _forEachIndex, 0];
} forEach pTypes;
```
ace_pylons_init_counter: Average: 0.0168457s / 197 = 0.0855112ms
ace_pylons_init_counter: Average: 0.0128784s / 197 = 0.0653727ms

For duplicate vehicles, it would be even faster. IMHO, this isn't a problem

Compiling all actions for all vehicles not only takes time at start, but it also makes the interaction menu slower by adding cost to `getVariable` lookups if `count allVariables (ace_interact_menu_ActNamespace)` has hundreds of extra entries

dedmen:
>slower by adding cost to `getVariable` lookups

Hashmap lookup is O(1) anyway.

addons/pylons/XEH_postInit.sqf
```diff
@@ -2,28 +2,26 @@
 
 private _filter = "isClass (_x >> 'Components' >> 'TransportPylonsComponent') && {(getNumber (_x >> 'scope')) > 0}";
 GVAR(aircraftWithPylons) = (_filter configClasses (configFile >> "CfgVehicles")) apply {configName _x};
```
dedmen: These two lines need 160ms to run (with my modset) 
Might be worth moving that to preStart and caching it.

dedmen:
This stuff just showed up in profiler. 160ms for the config lookup and 370ms for the addClassEHs
Actually causes pylons postInit to be the worst of all,
tracy_2019-02-07_21-20-15

I'd say just implement @PabstMirror's idea.
We can later just profile what takes the most time at vehicle spawning and optimize that away, or revert this if it turns out so bad that it matters, though 0.1ms doesn't sound like it would matter much.

PR #6388, opened 2018-06-10
Pylons - Add ACE prefix to settings category

PabstMirror:


654wak654:
#6381 does this too, but the other way around:

https://github.com/acemod/ACE3/pull/6381/files#diff-c0360faf7e7639f194e7844c41e266a7

mharis001:
This is the better way to do it from now on, will change my PR.

PR #6395, opened 2018-06-17
Cargo load ignore interaction distance parameter

tbeswick96:
- Add an optional parameter to `ace_cargo_loadItem` & `ace_cargo_canLoadItemIn`
- Parameter allows interaction distance check in `ace_cargo_canLoadItemIn` to be ignored
- I needed this for a persistence implementation where supply boxes with inventories as ace cargo are made persistent, and couldn't previously be loaded in due to distance
- Maybe someone else will find this useful


jonpas:
Nice! I wanted this. :D

PabstMirror:
Would we want to use this to remove distance check on zeus cargo loading?

PR #6396, opened 2018-06-19
Zeus - Ignore interaction distance check with cargo load module

tbeswick96:
- Uses ignore parameter for `loadItem` to ignore interaction distance check when loading cargo through zeus


PabstMirror:
Ref #6395

PR #6397, opened 2018-06-19
Nightvision - small improvments and a  new "ninetube" mask

Mezilsa:
- Fix an error that made the "Grey" ENVG use the "small" tube-mode
- create and add a new "ninetube" screen for the "spidereye NVGs" (may need a few size changes)
- add all Generations of NVGs in the black, green and brown (1, 2, 4 as 3 is allready present)


xrufix:
I think it's unlikely that a NVG with nine lenses would also use nine seperate screens (and if, they would not likely overlapp as with your suggested overlay). It would rather compute two images and use only two screens, creating wide field-of-view image similar to VR headsets.

Thumbs up on the other changes, though.

Mike-MF:

PR #6402, opened 2018-06-22
Explosives - New Timer UI

mharis001:
- Add new explosives timer UI [(video preview)](https://gfycat.com/HandsomeSoupyFlatfish)
- Time on the new UI can be changed with the scroll wheel (normal +/- 1s, with CTRL held down +/- 10s)
- Improve `ACE_Triggers` explanation comment
timer_3


PabstMirror:
Looks great
added a small check in case the explosive was picked up by someone else (problem existed before this pr)

PR #6403, opened 2018-06-23
Medical Menu - Fix annoyance of mouse position change when opening

mharis001:
- title #medicalrewritegoals


mharis001:
I have not experienced any random mouse jumps when opening the UI with this change. It stays where it was before the UI is opened.

PR #6405, opened 2018-06-24
don't make Explosives close one display at death

commy2:
Explosives closes one display at killed.
This breaks e.g. the Life Respawn menu.
This PR makes the component no longer kill a display indiscriminately, but only the Cellphone one.

Debug script:
```sqf
player addEventHandler ["Killed", {
    createDialog "RscDisplayEmpty";
}];
```
Before this PR: no display created, after this PR: display created successfully.

Note: don't run the code twice. Only one display is closed, so executing it twice seemingly fixes the issue, but doesn't really.

### DIFF
--- a/addons/explosives/functions/fnc_onIncapacitated.sqf
+++ b/addons/explosives/functions/fnc_onIncapacitated.sqf
@@ -24,7 +24,7 @@ if (_unit == ace_player) then {
     findDisplay 8855 closeDisplay 0;
 };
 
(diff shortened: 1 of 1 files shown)


dedmen:
https://github.com/acemod/ACE3/commit/54320eb451795f87761b6ad69580a484e08517b2

commy2:
I shouldn't have doubted CorruptedHeart for a second.

PR #6407, opened 2018-06-25
[WIP] Fix script errors reporting wrong line numbers

dedmen:
analogue to https://github.com/CBATeam/CBA_A3/pull/937

Todo:
- [ ] check if documentation generator can handle this

Shouldn't need to worry about merge conflicts. unless someone edited the first line of some script.

Doing all commits for automated replacement first. Then manual commits that need closer review later.
Automatic updates use notepad++ replace in files
`\s*(/\*[\s\S]+?\*/)\s*#include\s*"script_component\.hpp"`
-> `#include "script_component.hpp"\n$1`


dedmen:
All automatic changes are done to this point. Now to the manual ones... Which a quick search tells me will be 28!

dedmen:
>check if documentation generator can handle this

We don't even have a documentation generator as far as I can see :D 
I'd like to add a check to the SQF validator that there are no #include's after block comments. Which is just a simple regex. But for some files like https://github.com/acemod/ACE3/blob/master/addons/zeus/functions/fnc_zeusAttributes.sqf#L23
That would always throw errors.

SilentSpike:
@dedmen Exclude `.sqf` extension from the filter?

dedmen:
@SilentSpike We only want to check sqf files though. ^^

dedmen:
need someone to test the sqf_validator.py change

Dystopian:
> need someone to test the sqf_validator.py change

tested, works

dedmen:
@Dystopian Working means it should throw an error about https://github.com/acemod/ACE3/blob/master/addons/zeus/functions/fnc_zeusAttributes.sqf#L23 currently. 
Does it do that? Or does it work in the sense that it isn't completely broken?

Dystopian:
It really triggers if `#include` is not before `/*comment*/` but right after it.
It doesn't throw error if **anything except whitespace**  is before `/*comment*/` and I don't know why :disappointed:.
Slack for more.

dedmen:
>check if documentation generator can handle this

🤔 Which documentation generator even? Does ACE have one?

Dystopian:
#5253?

jonpas:
Ah, @SilentSpike !

I think we should merge this, documentation script can easily be changed in order afterwards.

dedmen:
Somehow I thought we already had a documentation generator merged :thinking:

Issue #5758, opened 2017-11-13
Can't speak when being dragged


Avokadomos:
**Arma 3 Version:** `1.76` (stable)
**CBA Version:** `3.4.1` (stable)
**ACE3 Version:** `3.11.0` (stable)

**Mods:**
```
- CBA_A3
- ace
- Task Force Radio
```

**Description:**
If you happen to wake up when you are unconscious and being dragged, you still can't be heard when trying to speak.

**Steps to reproduce:**
- One player becomes unconscious.
- Another player drags the player. 
- The unconscious player wakes up and yells "I'M ALIVE, RELEASE ME" repeatedly.
- The player who drags doesn't hear him and continues dragging.
- A couple of minutes later they arrive at a safe spot and release the now conscious player.
- The player who couldn't be heard can now be heard.

**Where did the issue occur?**
Dedicated & Self-Hosted Multiplayer

**Placed Modules:**
None.

commy2:
Sounds like a problem with tfar/acre, not ace.

SageNTitled:
Isn't it ACE that mutes the player when unconscious?

jonpas:
We handle that in ACE3 (https://github.com/acemod/ACE3/blob/master/addons/medical/XEH_postInit.sqf#L36-L51), but I don't see how it could cause that. Must be something else.

commy2:
I think it has to do with attachTo.

PabstMirror:
We delay waking up when being dragged
```
    // Wait until the unit isn't being carried anymore, so we won't end up with wierd animations
    if !(([_unit] call FUNC(isBeingCarried)) || ([_unit] call FUNC(isBeingDragged))) then {
```
https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_unconsciousPFH.sqf#L56

kripto202:
happens when being carried i've seen

PR #6411, opened 2018-06-27
Fix typo in med and issue related to unconsciousness

alganthe:
- Fix adenosine displayname being wrong in some medical interactions 
- Fix unconscious units waking up while being carried staying muted 

For the unconscious units, turns out the global EH isn't called if the unit is being carried, and the muting / unmutting relies on it.

### DIFF
--- a/addons/medical/ACE_Medical_Actions.hpp
+++ b/addons/medical/ACE_Medical_Actions.hpp
@@ -561,7 +561,7 @@ class ACE_LegLeft {
         icon = QPATHTOF(UI\icons\autoInjector.paa);
     };
     class Adenosine: Morphine {
-        displayName = CSTRING(Inject_Atropine);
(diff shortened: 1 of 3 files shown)


dedmen:
Ref: https://github.com/acemod/ACE3/blob/2bde515aaaf974b77c50d5b71ae54ff08958f5b5/addons/medical/XEH_postInit.sqf#L50

PabstMirror:
Fix #5758
I think this is an improvement, but there still might be an edge case, (not sure)

Unit is uncon and being carried
Unit wakes up and can now talk
Unit passes out again but won't be muted?

alganthe:
The unit will be muted again because the EH is called when passing out, I think.

I tested this fix by making an AI carry me, making myself pass and and wake up again.

PR #6413, opened 2018-06-30
Explosives - Optimize creating explosive place actions

mharis001:
- Optimize explosive place action (remove useless condition) and creating its children (optimized function and return is cached)

All of these were tested with `count magazines = 28` and `placeable explosives => 6, 2, 1, 1` (where each number represents the amount of a unique type of explosive).
- The old function took ~0.31 ms
- Useless condition on `ACE_Place` action took ~0.1 ms (time it takes `hasExplosives`)
- New function takes ~0.20 ms (without any caching)
- New function first time (with caching) takes ~0.23 ms (cachedCall adds ~0.03 ms)
- New function (when return is cached) takes ~0.009 ms

**Overall**: Instead of ~0.41 ms every second, the new method is ~0.23 ms after a change in the player's loadout and then ~0.009 ms (accessing the cached value) every second after that.

For an inventory with no placeable explosives and `count magazines = 17`: it is ~0.11 ms vs ~0.10 ms (first time after loadout change) and ~0.009 ms every time after that.

### DIFF
--- a/addons/explosives/functions/fnc_hasExplosives.sqf
+++ b/addons/explosives/functions/fnc_hasExplosives.sqf
@@ -21,4 +21,4 @@ TRACE_1("params",_unit);
 private _cfgMagazines = configFile >> "CfgMagazines";
 private _magazines = magazines _unit;
 
-(_magazines arrayIntersect _magazines) findIf {getNumber (_cfgMagazines >> _x >> QGVAR(Placeable)) == 1} > -1
+((_magazines arrayIntersect _magazines) findIf {getNumber (_cfgMagazines >> _x >> QGVAR(Placeable)) == 1}) > -1
--- a/addons/explosives/functions/fnc_hasExplosives.sqf
+++ b/addons/explosives/functions/fnc_hasExplosives.sqf
@@ -9,7 +9,7 @@
  * Has explosives <BOOL>
  *
  * Example:
- * [player] call ace_explosives_fnc_hasExplosives
+ * player call ace_explosives_fnc_hasExplosives
  *
  * Public: Yes
  */
--- a/addons/explosives/functions/fnc_hasExplosives.sqf
+++ b/addons/explosives/functions/fnc_hasExplosives.sqf
@@ -21,4 +21,4 @@ TRACE_1("params",_unit);
 private _cfgMagazines = configFile >> "CfgMagazines";
 private _magazines = magazines _unit;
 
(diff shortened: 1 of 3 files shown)


mharis001:
The `hasExplosives` function is not used anywhere else after this change. The header says that it is a public function but it is not on the wiki. Safe to remove it?

TheMagnetar:
I would not remove a public function. Some people do look at the code and by removing it you may break somethinget like other mode or missions

dedmen:
@TheMagnetar explain?

dedmen:
hasExplosives still sounds useful. But atleast add that arrayIntersect fix to it

jonpas:
@TheMagnetar are you using it? If it's not marked "Public: Yes" it's not meant for public use. Of course if someone uses it still, it's up to them to pay attention and update it or request it to be made public API properly.

alganthe:
^ hasExplosive is indeed marked public jonpas.

imho if public funcs are removed a core dev should make an announcement on the forums and website to have the maximum visibility.

As to the usefulness of said func, i'd say it has it's place.

mharis001:
That use is taken out in this PR, if the player has no explosives, there wont be any children actions and the action wont show anyway. Will update `hasExplosives` to be more efficient since it not being removed.

jonpas:
Oh maybe I should've read the other comments properly... Yeah, as a Public function we have to keep backwards compatibility. If it's essentially deprecated, then mark it as so with the `ACE_DEPRECATED` macro (+2 major versions from the one it was added in). No need to improve, only remove when the time comes! :D

mharis001:
The function isn't getting a better alternative, I just thought about removing it since it isn't used anywhere is ace. Still deprecate?

addons/explosives/functions/fnc_hasExplosives.sqf
```diff
@@ -18,12 +18,7 @@
...
+private _cfgMagazines = configFile >> "CfgMagazines";
 private _magazines = magazines _unit;
-{
-    if (getNumber (ConfigFile >> "CfgMagazines" >> _x >> QGVAR(Placeable)) == 1) exitWith {
-        _result = true;
-    };
-} count _magazines;
 
-_result
+_magazines arrayIntersect _magazines findIf {getNumber (_cfgMagazines >> _x >> QGVAR(Placeable)) == 1} > -1
```
dedmen: `(_magazines arrayIntersect _magazines) findIf {getNumber (_cfgMagazines >> _x >> QGVAR(Placeable)) == 1} > -1`

commy2:
¯\\\_(ツ)_/¯

PR #6421, opened 2018-07-09
Add CPR Upper and Lower limits 

654wak654:
- Adds CPR lower and upper limits. This allows people to configure how long CPR adds to a unit's revive timer, while still keeping the randomness element.
- **WIP!** I'm not sure what's the best way to enforce `upperLimit >= lowerLimit`. Advice on this? (Check the TODO in XEH_postInit.sqf)


shukari:
if `lowerLimit > upperLimit` then turn them that the old upper is the new lower.
So it is no problem if the values are 10 and 20 or 20 and 10 it will have the same effect.

PR #6422, opened 2018-07-09
ace_infoDisplayChanged will now trigger also for mods

shukari:
- add 'RscUnitInfo' to Extended_DisplayLoad_EventHandlers
- because 90% of the RscInGameUIs are use BIS_fnc_initDisplay with RscUnitInfo
- instead of adding onLoad to all mod RscInGameUIs (example: RHS) it will go over the cba xeh

This works now perfect for my desire, the rhs jets now also get a limited UI by the Ace_UI addon. But i dont know what side effects this triggers.

### DIFF
--- a/addons/common/CfgEventHandlers.hpp
+++ b/addons/common/CfgEventHandlers.hpp
@@ -23,6 +23,9 @@ class Extended_DisplayLoad_EventHandlers {
     class RscDisplayMission {
         ADDON = QUOTE(_this call COMPILE_FILE(XEH_missionDisplayLoad));
     };
+    class RscUnitInfo {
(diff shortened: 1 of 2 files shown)


jonpas:
There shouldn't be any side effects from ACE UI itself, only added support (which is great!), I am not sure about other things though, @commy2 ?

PabstMirror:
`uiNamespace setVariable [ARR_2('ACE_dlgSoldier',     _this select 0)];`
I can't find any use of the `ACE_dlg*` variables in our code, anyone know what they were for?

jonpas:
Reads like "dialog" to me.

PabstMirror:
If it's just a old thing from AGM we might be able to drop the entire line, but it doesn't really hurt.

PR #6424, opened 2018-07-09
eden rearm vehicles and pylons can changed by non-config rearm vehicles

### LINKED ISSUE #6401: this setVariable ["ace_rearm_isSupplyVehicle", true] doesn't work
**Arma 3 Version:** `1.82` (stable)
**CBA Version:** `3.7.1` (stable)
**ACE3 Version:** `3.12` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
Adding 'this setVariable ["ace_rearm_isSupplyVehicle", true]' to vehicle does nothing. Vehicle still can't rearm other vehicles.

shukari:
- eden attribute vor supply/rearm vehicles
- pylons can now changed with scripted rearm vehicles
- an easy public make rearmvehicle function
- an easy is this vehicle a rearm vehicle function

resolve #6401

IMPORTANT
the jip functionality is not tested... im not a pro at this... i need a second pair of eyes there


addons/rearm/functions/fnc_makeSource.sqf
```diff
@@ -0,0 +1,55 @@
...
+// already initialized because this is a config rearm vehicle
+if (_rearmCargoConfig > 0 || _source getVariable [QGVAR(isSupplyVehicle), false]) exitWith {};
+
+_source setVariable [QGVAR(isSupplyVehicle), true, true];
+
+// check if menu already exists
+if (!isNil {_source getVariable QGVAR(initSupplyVehicle_jipID)}) exitWith {};
+
+private _jipID = [QGVAR(initSupplyVehicle), [_source]] call CBA_fnc_globalEventJIP;
+[_jipID, _source] call CBA_fnc_removeGlobalEventJIP;
```
dedmen: This is confusing. Why are you adding a JIP event and remove it right away?

shukari: that is the point where i have no clue... the makeSource function is a template copy of the makeSource refuel function and @Dystopian done that this way ^^
https://github.com/acemod/ACE3/blob/29726946f689d921762ac77a847dbbebc7bf4a8c/addons/refuel/functions/fnc_makeSource.sqf#L59

Dystopian: @dedmen see CBATeam/CBA_A3#741, `CBA_fnc_removeGlobalEventJIP` has ambiguous name IMHO

dedmen: >Either immedietly or if passed an object it will remove the jipEvent if object is deleted.

Okey yeah.. Kinda bad name

PabstMirror:
JIPs could get double actions added because both globalEventJIP and xeh ran
GlobalEventJIP also runs before settings are init (which have a minor effect on a string)
And disabled (`-1`) had no effect (e.g. setting an ammo truck to -1.

I think the issues should be resolved by this commit: https://github.com/acemod/ACE3/commit/d5b3696c02b82116c330f26a181fdab70fce9f5a

PR #6426, opened 2018-07-11
 Respawn Addon reworked

shukari:
- settings converted to CBA
- friendlyfire fixed and is now a CBA setting
- friendlyfire modul disabled (scope=1)
- fixed problem with respawnDelay and SavePreDeathGear


addons/respawn/functions/fnc_showFriendlyFireMessage.sqf
```diff
@@ -1,25 +1,35 @@
...
+
 if (_unit != _killer && {side group _unit in [side group ACE_player, civilian]} && {side group _killer == side group ACE_player}) then {
-    systemChat format ["%1 was killed by %2", [_unit, false, true] call EFUNC(common,getName), [_killer, false, true] call EFUNC(common,getName)];
+    private _msg = format ["[FriendlyFire] " + localize LSTRING(FriendlyFire_Message), 
+            ([_unit, false, true] call EFUNC(common,getName)) + ([" (AI)", ""] select (isPlayer _unit)),
+            ([_killer, false, true] call EFUNC(common,getName)) + ([" (AI)", ""] select (isPlayer _killer))
+        ];
+    
+    systemChat _msg;
+    diag_log _msg;
```
commy2: `diag_log text` to get rid of the quote marks.

addons/respawn/stringtable.xml
```diff
@@ -333,6 +333,10 @@
             <Chinesesimp>摆放此模块后，当有发生友军误击致死的事件，会显示提示讯息在聊天视窗中。</Chinesesimp>
             <Chinese>擺放此模塊後，當有發生友軍誤擊致死的事件，會顯示提示訊息在聊天視窗中</Chinese>
         </Key>
+        <Key ID="STR_ACE_Respawn_FriendlyFire_Message">
+            <English>%1 was killed by %2</English>
+            <German>%1 wurde durch %2 getötet</German>
+        </Key>
```
commy2: This string is already in the base game. Let me look it up...

commy2:
```cpp
        <Key ID="str_killed_by">
          <Original>%s was killed by %s</Original>
          <French>%s a été tué par %s</French>
          <Spanish>%s asesinado por %s</Spanish>
          <Italian>%s è stato ucciso da %s</Italian>
          <Polish>%s został zabity przez %s</Polish>
          <Russian>%s убит игроком %s</Russian>
          <German>%s wurde von %s getötet</German>
          <Czech>%s byl zabit %s</Czech>
          <English>%s was killed by %s</English>
          <Chinese>%s 被 %s 擊殺了</Chinese>
          <Portuguese>%s foi morto por %s</Portuguese>
          <Korean>%s 사망(공격자: %s)</Korean>
          <Chinesesimp>%s 被 %s 杀死</Chinesesimp>
          <Turkish>%s, %s tarafından öldürüldü</Turkish>
          <Japanese>%sは%sに殺された</Japanese>
        </Key>
```
May have to make your own String and use %1, %2.

addons/respawn/XEH_postInit.sqf
```diff
@@ -1,7 +1,11 @@
...
 
 ["ace_rallypointMoved", FUNC(updateRallypoint)] call CBA_fnc_addEventHandler;
 ["unit", FUNC(handlePlayerChanged)] call CBA_fnc_addPlayerEventHandler; // hide enemy rallypoint markers
 
 [QGVAR(showFriendlyFireMessageEvent), FUNC(showFriendlyFireMessage)] call CBA_fnc_addEventHandler;
+
+// fix, because killed EH will not get called at time before respawn EH
+if (GVAR(SavePreDeathGear) && getMissionConfigValue ["respawnDelay", 0] == 0) then {
+    ["CBA_loadingScreenDone", {setPlayerRespawnTime 0.01}] call CBA_fnc_addEventHandler;
+};
```
commy2: This should be the setting changed code block, so that this fix is applied when changing the setting mid session.

shukari: It is here https://github.com/shukari/ACE3/blob/0f51226521d3afd6df81d7e2fdd647372df49579/addons/respawn/initSettings.sqf#L9

It needs to execute on startup when SavePreDeathGear is true and also if setting was false and then is set mid mission to true

Miss i something?

commy2: Is this really needed in postInit then? I also think that even if you use loadingScreenDone, the EH should be added in preInit (with the check for the setting inside the EH code block).

addons/respawn/stringtable.xml
```diff
@@ -334,8 +334,21 @@
             <Chinese>擺放此模塊後，當有發生友軍誤擊致死的事件，會顯示提示訊息在聊天視窗中</Chinese>
         </Key>
         <Key ID="STR_ACE_Respawn_FriendlyFire_Message">
-            <English>%1 was killed by %2</English>
-            <German>%1 wurde durch %2 getötet</German>
+          <Original>%1 was killed by %2</Original>
```
commy2: We don't use Original. We just put English as first entry, so that is chosen as default automatically by virtue of Original being undefined.

addons/respawn/XEH_postInit.sqf
shukari: Is it important if it is postInit or preInit?

shukari: changed

shukari:
@commy2 😄 it was because no other file had that type of author branding and i had the code before there... now it looks... yeah ;D

addons/respawn/functions/fnc_showFriendlyFireMessage.sqf
```diff
@@ -1,25 +1,35 @@
...
  *
  * Public: No
  */
-#include "script_component.hpp"
-
 params ["_unit", "_killer"];
 
+// fix, because ace medical cause _unit == _killer
+if (_unit == _killer) then {
+   _killer = _unit getVariable [QEGVAR(medical,lastDamageSource), _killer];
```
PabstMirror: This functions runs globally, but the medical var lastDamageSource is only set where unit is local.
Will probably want to determine this in fnc_handleDamage
ref https://github.com/acemod/ACEX/blob/master/addons/killtracker/XEH_postInit.sqf

addons/respawn/functions/fnc_showFriendlyFireMessage.sqf
```diff
@@ -1,25 +1,35 @@
...
+    private _msg = format ["[FriendlyFire] " + localize LSTRING(FriendlyFire_Message), 
+            ([_unit, false, true] call EFUNC(common,getName)) + ([" (AI)", ""] select (isPlayer _unit)),
+            ([_killer, false, true] call EFUNC(common,getName)) + ([" (AI)", ""] select (isPlayer _killer))
+        ];
+    
+    systemChat _msg;
+    diag_log text _msg;
 
     // Raise ACE globalEvent
     ["ace_killedByFriendly", [_unit, _killer]] call CBA_fnc_globalEvent;
```
PabstMirror: Nothing you changed, but this functions is called by a global event, and raises another global event

addons/respawn/initSettings.sqf
```diff
@@ -0,0 +1,30 @@
+[
+    QGVAR(savePreDeathGear),
+    "CHECKBOX",
+    [LSTRING(SavePreDeathGear_DisplayName), LSTRING(SavePreDeathGear_Description)],
+    LSTRING(DisplayName),
+    true,
+    true, {
+        if (_this && {getMissionConfigValue ["respawnDelay", 0] == 0}) then {
+            0.01 remoteExec ["setPlayerRespawnTime"];
```
PabstMirror: Why `remoteExec`? The code block should run on all machines

addons/respawn/initSettings.sqf
```diff
@@ -0,0 +1,30 @@
+[
+    QGVAR(savePreDeathGear),
+    "CHECKBOX",
+    [LSTRING(SavePreDeathGear_DisplayName), LSTRING(SavePreDeathGear_Description)],
+    LSTRING(DisplayName),
+    true,
```
PabstMirror: default value should be same as before: false

addons/respawn/initSettings.sqf
```diff
@@ -0,0 +1,30 @@
...
+    true
+] call CBA_Settings_fnc_init;
+
+[
+    QGVAR(showFriendlyFireMessage),
+    "CHECKBOX",
+    [LSTRING(FriendlyFire_Module_DisplayName), LSTRING(FriendlyFire_Module_Description)],
+    LSTRING(DisplayName),
+    true,
```
PabstMirror: default value should be false

PabstMirror:
ace_respawn_fnc_moduleFriendlyFire should be something like
```
["CBA_settings_setSettingMission", [QGVAR(showFriendlyFireMessage), true, true]] call CBA_fnc_localEvent;
```
(should run on all machines) to maintain compatiblity with cba settings
removing the old ` publicVariable QGVAR(showFriendlyFireMessage);`

addons/respawn/initSettings.sqf
shukari: If this code is exec global than this is pointless... i didnt know that,

addons/respawn/initSettings.sqf
shukari: copy paste fail

shukari:
@PabstMirror do you can fix this... i cant follow all your points.

addons/respawn/stringtable.xml
```diff
@@ -351,6 +351,22 @@
...
+            <Russian>%1 убит игроком %2</Russian>
+            <German>%1 wurde von %2 getötet</German>
+            <Czech>%1 byl zabit %2</Czech>
+            <Chinese>%1 被 %2 擊殺了</Chinese>
+            <Portuguese>%1 foi morto por %2</Portuguese>
+            <Korean>%1 사망(공격자: %2)</Korean>
+            <Chinesesimp>%1 被 %2 杀死</Chinesesimp>
+            <Turkish>%1, %2 tarafından öldürüldü</Turkish>
+            <Japanese>%1は%2に殺された</Japanese>
+        </Key>
```
commy2, line 369: We have this message in vanilla?

shukari: https://github.com/acemod/ACE3/pull/6426#discussion_r201852965

Mike-MF:
This was done in #8461

PR #6437, opened 2018-07-17
add microdagrwaypoint precision as an option

Mezilsa:
- Remove the magic number `3` of the microdagr precision and adds the option to choose it.

Available Options are "1m", "10m", "100m". The default option will be 1m (cause there are people who don't want to calculate mortar stuff)

I am not shure if `(GVAR(waypointPrecision) +1)` is the style of Ace. I ask kindly to correct me if I messed up.


addons/microdagr/stringtable.xml
```diff
@@ -497,6 +497,27 @@
             <Chinesesimp>无 (无法检视地图)</Chinesesimp>
             <Chinese>無 (無法檢視地圖)</Chinese>
         </Key>
+        <Key ID="STR_ACE_MicroDAGR_WaypointPrecision_DisplayName">
+            <English>Waypoint Precision</English>
+            <German>Wegpunkt Genauigkeit</German>
+        </Key>
+        <Key ID="STR_ACE_MicroDAGR_WaypointPrecision_Description">
+            <English>Controls how precise the waypointdistance can be displayed.</English>
```
dedmen: remove the punctuation at the end. Same for all the other new strings.

Mezilsa: Done. Can you show me how to make "code snippets" ?

addons/microdagr/ACE_Settings.hpp
```diff
@@ -7,4 +7,12 @@ class ACE_Settings {
         isClientSettable = 0;
         values[] = {CSTRING(MapFill_None), CSTRING(MapFill_OnlyRoads), CSTRING(MapFill_Full)};
     };
+        class GVAR(waypointPrecision) {
```
mharis001: New settings should use `CBA_settings_fnc_init` directly

Mezilsa: You mean like here?

https://github.com/acemod/ACE3/blob/master/addons/advanced_fatigue/ACE_Settings.hpp
https://github.com/acemod/ACE3/blob/master/addons/advanced_fatigue/initSettings.sqf

mharis001: Yes, but since this is a new setting you would not use `movedToSQF`

addons/microdagr/stringtable.xml
dedmen: which one? You mean the code displayed above my comment? Yeah that's straight from the diff.

Mezilsa:
I seems to be working, but I am not shure that the the micro dagr shouldn't get it's own option section and the `"ACE uncategorized"` seems wrong. 
Pretty shure I botched that one.

PabstMirror:
I still don't think this one needs it's own category, but I moved both settings to a microdagr subcategory so they should be a little easier to find.

PR #6438, opened 2018-07-17
Move scream and moan sounds to medical feedback

SilentSpike:
Also moving the vanilla feedback changes over.


addons/medical_feedback/XEH_preInit.sqf
```diff
@@ -26,4 +26,741 @@ PREP_RECOMPILE_END;
...
+], [
+    ["A3\sounds_f\characters\human-sfx\P04\Hit_Max_1",3.1622777,1,80],
+    ["A3\sounds_f\characters\human-sfx\P04\Hit_Max_2",3.1622777,1,80],
+    ["A3\sounds_f\characters\human-sfx\P04\Hit_Max_3",3.1622777,1,80],
+    ["A3\sounds_f\characters\human-sfx\P04\Hit_Max_4",3.1622777,1,80],
+    ["A3\sounds_f\characters\human-sfx\P04\Hit_Max_5",3.1622777,1,80]
+]]];
+
+GVAR(HitScreamNamespace) setVariable ["Male03GRE", [[
+    ["A3\sounds_f\characters\human-sfx\P05\Hit_Low_1",3.1622777,1,80],
```
SilentSpike: I feel like we should be able to cut down on the amount of repeat code here with macros

bux:
The title sounds dirty

addons/medical_feedback/XEH_preInit.sqf
SilentSpike: or can we not even pull this data from config somehow? It's done preInit so any extra overhead is imo worth reducing the mass of lines.

SilentSpike: We should actually even move it to preStart so it only gets done once ever! 👍

SilentSpike:
The only downside to this PR is that it introduces a dependency on `medical_feedback` for both `medical_damage` and `medical_statemachine` (see dependency graph [here](https://github.com/acemod/ACE3/pull/6301#issuecomment-405662793)).

However, I think this code definitely makes more sense in this component and we could update the code so that those components check if this function exists before consuming it.

SilentSpike:
Yeah it's missing simply because I made this PR before looking into all that. Ideally though I'd actually like to not introduce that dependency as it doesn't make too much sense.

We could use events to trigger the screams/moans or check if the function exists first before calling it? The other option is to leave this functionality in `medical_engine`.

dedmen:
I'd say events make sense here. Also add's a nice API for users. More sense atleast than always checking if a function that exists for over 90% of users, does exist.

PR #6439, opened 2018-07-17
RHS Compatibility - g force adjustments and MH6M can now fastrope

shukari:
- g force adjustments
- RHS g force adjustments
- MH6M can now fastrope


commy2:
Called g force for "gravity force", and not Gforce like those GPUs from NVIDIA.

PR #6440, opened 2018-07-17
Move medical facility, vehicle and unit functions

SilentSpike:
They seem to make more sense here to me as these functions are used in treatment. Not sure if everyone agrees though so feel free to weigh in.


TheMagnetar:
I think that the `isMedic`, `isInMedicalFacility` or `isMedicalVehicle` functions should be moved to medical. I sort of like what you propose in #6301 
> Here's how the current dependency graph looks for the restructure (excluding medical which I think we should gut and reserve for public functionality):

thojkooi:
I wonder if `isBeingCarried` and `isBeingDragged` should also be moved out of medical_status.

SilentSpike:
@TheMagnetar My plan was to leave these version in treatment and then just add public wrappers for them in `medical`

SilentSpike:
@thojkooi Good idea, I'll check where they are used to see if it's a suitable move 👍

PR #6441, opened 2018-07-17
Fix required addons in medical restructure

SilentSpike:
- Title


addons/medical_status/config.cpp
```diff
@@ -6,7 +6,7 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"ace_medical"};
+        requiredAddons[] = {"ace_medical_engine"};
```
TheMagnetar: You need ace_medical for the state machine to work.

PR #6442, opened 2018-07-17
Strip medical component

SilentSpike:
The idea here is to move everything out of medical into its appropriate component of medical (and reserve this component for public functionality only).


SilentSpike:
Looking at some of the remaining functions in `medical`:
- `displayTriageCard`, the display used by this function is in gui and it makes sense to put this function there too, but there's no call it it anywhere in the code currently
- `addStateHandler`, unsure what this is for specifically, there's only one commented out line that uses it in `medical_damage`
- `setUnconsciousStatemachine`, can move this one down to `medical_status` as it's only used there and in `medical_statemachine`
- `handleIncapacitation`, can move this one down to `medical_damage`

SilentSpike:
That's the bulk of this component stripped out. We should probably move `script_macros_medical.hpp` to `medical_engine` since this component is no longer at the top of the dependency hierarchy

SilentSpike:
Also it seems the paths to textures and stuff in the item p3d's need to be updated to reflect the new location

PR #6446, opened 2018-07-18
Add script to sort all stringtables

kymckay:
- Add a python script `sort_stringtables.py` to the tools directory. This sorts all `Key` tags alphabetically by their `ID` attribute (or `name` for `Container` and `Package` tags).

Would be nice if we could set this up to automatically run at some point.

### DIFF
--- a/tools/sort_stringtables.py
+++ b/tools/sort_stringtables.py
@@ -56,12 +56,12 @@ def main():
 
         # Verify that stringtable is structured as expected
         if xml_root.tag != 'Project':
-            print('Missing "Project" root tag: {}'.format(filepath))
+            #print('Missing "Project" root tag: {}'.format(os.path.join(folder, 'stringtable.xml')))
             continue
 
         sort_children(xml_root)
         indent_pretty(xml_root)
-        print('Sorted: {}'.format(filepath))
+        #print('Sorted: {}'.format(os.path.join(folder, 'stringtable.xml')))
 
         tree.write(filepath, encoding='utf-8', xml_declaration=True, method='xml')
 
--- a/tools/sort_stringtables.py
+++ b/tools/sort_stringtables.py
@@ -20,28 +20,51 @@ def indent_pretty(parent, level=0):
     for child in parent:
         indent_pretty(child, level+1)
         if child == last_child:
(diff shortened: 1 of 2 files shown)


alganthe:
This should run and commit right after changes are pushed to master, otherwise this is going to be merge conflict galore.

tools/sort_stringtables.py
```diff
@@ -0,0 +1,48 @@
...
+    parent.tail = '\n' + "".join(['    '] * level)
+
+    if len(parent) is 0:
+        return
+
+    last_child = parent[len(parent) - 1]
+    for child in parent:
+        indent_pretty(child, level+1)
+        if child == last_child:
+            child.tail = '\n' + "".join(['    '] * level)
```
SilentSpike: Note to self, this line should just be:
```python
child.tail = parent.tail
```

jonpas:
You can easily automatically run it on pushes to master, check tools/deploy.py for handling diff and pushes, and just add another section for this.

SilentSpike:
Thanks for the info, I'll take a look at that 👍

SilentSpike:
As far as I can tell there's no way to commit multiple files in a single commit via PyGithub, so this will create a commit for each stringtable file changed.

Not entirely sure how to test the new deploy.py code either, but I *think* I've done it right 🤔

alganthe:
^ stage the files, commit once done?

SilentSpike:
There's no method to do that

jonpas:
Good point, you can't with PyGithub, and they still didn't implement it. But you can use git directly, using given token maybe?

SilentSpike:
There might be a way to just run the script and commit via the circle.yml, but I find their documentation pretty confusing and can't figure that out

kymckay:
Looking through my old PRs, could probably do this now with a GitHub action running on pushes to master. May look into getting this one finally resolved if I have time.

jonpas:
> Looking through my old PRs, could probably do this now with a GitHub action running on pushes to master. May look into getting this one finally resolved if I have time.

Should be done via GitHub action indeed. Closing this as it is not really that important, I welcome a PR though.

PR #6447, opened 2018-07-18
Fix mouse getting repositioned when opening medical menu

thojkooi:
Change from @mharis001 in PR #6403, done for the rewrite branch

Needs testing.


thojkooi:
Original issue is still present;

- open menu, observe position of the mouse
- close menu, open again. See position of the mouse changed to some random spot.

mharis001:
I am not sure what is going on. When the menu re-opens automatically after treatment the issue doesn't occur. But when closing and re-opening manually, it does.

thojkooi:
Ok I understand now. Yes I agree, that should be fixed. Need to figure out how to detect the reopen bit. Maybe pass a flag for centralizing the mouse?

mharis001:
That should work, will fix half of the problem at least.

PR #6454, opened 2018-07-20
Fix ladder dismount animation bug

### LINKED ISSUE #6227: "Unconscious" interpolation to "ACE_UnconsciousOutProne" causes unexpected animation behaviors
**Arma 3 Version:** `1.80`
**CBA Version:** `3.6.1`
**ACE3 Version:** `medical-rewrite` (state of the branch, as of https://github.com/acemod/ACE3/commit/306c4499864e7d4de283f8a0555419743f9b9f1e)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- Currently, the method used at https://github.com/acemod/ACE3/blob/medical-rewrite/addons/medical_engine/CfgMoves.hpp#L16 which interpolates the `Unconscious` animation to `ACE_UnconsciousOutProne` is flawed. This implementation results in, for one example, players being thrown to the ground when leaving a ladder.

**Steps to reproduce:**
- On the `medical-rewrite` branch, climb onto a ladder and then get off of it. You will transition into prone (`ACE_UnconsciousOutProne`) unexpectedly. If you edit the aforementioned line back to the default: `InterpolateTo[] = {"DeadState",0.1};` you'll note that the issue with diving to prone upon leaving ladders is no longer present. I have personally been able to find a few other animations that result in an unexpected/undesired transition to `ACE_UnconsciousOutProne`, but this one with ladders is by far the most impactful to gameplay.

- N/A, all environments

**Placed Modules:**
- N/A

**RPT log file:**
- N/A

SilentSpike:
Not sure of the original reason we interpolate to this ACE unconscious animation, but removing it seemingly has no effect and fixes all the other animations that were broken by it. Thanks to Bismarck for the fix 👍

Fixes #6227

### DIFF
--- a/addons/medical_engine/CfgMoves.hpp
+++ b/addons/medical_engine/CfgMoves.hpp
@@ -13,7 +13,7 @@ class CfgMovesMaleSdr: CfgMovesBasic {
 
         // wake me up inside
         class Unconscious: Default {
-            InterpolateTo[] = {"DeadState",0.1,"ACE_UnconsciousOutProne",0.2};
+            InterpolateTo[] = {"DeadState",0.1};
         };
 
         class AmovPpneMstpSnonWnonDnon;


commy2:
Sine this is now the same as vanilla, one can simplify:
https://github.com/acemod/ACE3/pull/4064/commits/78ee7d46775aa7aec269d9657f57095966d1e36c

PR #6455, opened 2018-07-20
Improve unconscious volume handling

SilentSpike:
- Lower the player's hearing when they go unconscious (both game sounds and addon local/radio voices).
  - Perhaps this should only affect voices?
- Re-evaluate player's ability to talk and hear upon changing unit
  - Fix potential for players to be able to talk if they transition into an already unconscious AI unit that was previously local to dedicated/headless.
  - Fix potential for players to respawn and not be able to hear or talk.
    - [x] We should probably be resetting these on death anyway because players could be in an external spectator system during that time.
  - This also removes the large number of variables being set on AI and never used.

### DIFF
--- a/addons/medical_feedback/XEH_postInit.sqf
+++ b/addons/medical_feedback/XEH_postInit.sqf
@@ -21,6 +21,7 @@ GVAR(heartBeatEffectRunning) = false;
 ["ace_unconscious", {
     params ["_unit", "_unconscious"];
 
+    // Toggle unit's ability to talk in radio addons
     if (local _unit) then {
         _unit setVariable ["tf_voiceVolume", [1, 0] select _unconscious, true];
         _unit setVariable ["tf_unable_to_use_radio", _unconscious, true];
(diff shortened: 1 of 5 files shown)


addons/medical_feedback/XEH_postInit.sqf
```diff
@@ -21,21 +21,42 @@ GVAR(heartBeatEffectRunning) = false;
...
-        _unit setVariable ["tf_unable_to_use_radio", _unconscious, true];
-        _unit setVariable ["acre_sys_core_isDisabled", _unconscious, true];
-    };
-
     if (_unit != ACE_player) exitWith {};
 
+    // Toggle unconscious player's ability to talk in radio addons
+    _unit setVariable ["tf_voiceVolume", [1, 0] select _unconscious, true];
+    _unit setVariable ["tf_unable_to_use_radio", _unconscious, true];
+    _unit setVariable ["acre_sys_core_isDisabled", _unconscious, true];
```
SilentSpike: @dedmen @jonpas 

Do ACRE/TFAR need these variables to be publicly set to prevent the local player from talking? I suspect the old code was only doing it to handle the potential for the player to change unit, but they're now updated whenever that happens anyway.

dedmen: voiceVolume yes.. sadly.
`tf_unable_to_use_radio` no. Local only

jonpas: Yeah, it appears it needs to be global.

SilentSpike: @dedmen In this PR I lower the hearing volume while unconscious, but muffled would be good to add.

and thanks guys!

PR #6462, opened 2018-07-22
Wire cutter - Allow use when unit has RHS Engineer UMBTS backpack

Dystopian:
- title.

~~Class name `rhs_assault_umbts_engineer` is hard-coded on purpose because I didn't find any analogues in vanilla, RHS and CUP gear.~~

### DIFF
--- a/addons/logistics_wirecutter/script_component.hpp
+++ b/addons/logistics_wirecutter/script_component.hpp
@@ -59,5 +59,6 @@
 
 #define HAS_WIRECUTTER(unit) (\
     "ACE_wirecutter" in ([ARR_6(unit, false, true, true, true, false)] call CBA_fnc_uniqueUnitItems) \
-    || {backpack unit isKindOf "rhs_assault_umbts_engineer"} \
+    || {1 == getNumber (configFile >> "CfgVehicles" >> (backpack unit) >> QGVAR(hasWirecutter))} \
+    || {1 == getNumber (configFile >> "CfgWeapons" >> (vest unit) >> QGVAR(hasWirecutter))} \
 )
--- a/optionals/compat_rhs_afrf3/CfgVehicles.hpp
+++ b/optionals/compat_rhs_afrf3/CfgVehicles.hpp
@@ -218,4 +218,9 @@ class CfgVehicles {
     class OTR21_Base: Truck_F {
         EGVAR(refuel,fuelCapacity) = 500;
     };
+
+    class rhs_assault_umbts;
+    class rhs_assault_umbts_engineer: rhs_assault_umbts {
+        EGVAR(logistics_wirecutter,hasWirecutter) = 1;
+    };
 };


Blutze:
Gives it a 5 mass capacity advantage over standardised Assault Packs. Considering most of that is carried on the outside, I'd be ok with that.

Not sure what to think about the precedent of adding functionality to other mods simply based on cosmetic aspects of their models though. And how well does that code work when you don't have RHS loaded? Does it still work if it is replaced with some kind of macro in case other examples crop up?

addons/logistics_wirecutter/script_component.hpp
```diff
@@ -56,3 +56,8 @@
 #define SOUND_CLIP_TIME_SPACING 1.5
 #define CUT_TIME_DEFAULT 11
 #define CUT_TIME_ENGINEER 7.5
+
+#define HAS_WIRECUTTER(unit) (\
+    "ACE_wirecutter" in ([ARR_6(unit, false, true, true, true, false)] call CBA_fnc_uniqueUnitItems) \
+    || {backpack unit isKindOf "rhs_assault_umbts_engineer"} \
```
dedmen: Backpacks that inherit from this are not guaranteed to have a model with a wirecutter on it.

Dystopian:
> And how well does that code work when you don't have RHS loaded?

It works OK. I thought about init checking if RHS is loaded and then replacing code with just `false` but I found it superfluous.
> Does it still work if it is replaced with some kind of macro in case other examples crop up?

If there are more examples it would be very easy to add their support. Until then there is no sense to do it.

Blutze:
> I would implement it if there was at least very small probability for other such backpacks.

@jokoho48 mentioned VSM and another mod that even had them on vests. Haven't touched either in ages, but even if there isn't anything in those - I'd still try and keep this future proof.

addons/logistics_wirecutter/script_component.hpp
```diff
@@ -59,5 +59,6 @@
 
 #define HAS_WIRECUTTER(unit) (\
     "ACE_wirecutter" in ([ARR_6(unit, false, true, true, true, false)] call CBA_fnc_uniqueUnitItems) \
-    || {backpack unit isKindOf "rhs_assault_umbts_engineer"} \
+    || {1 == getNumber (configFile >> "CfgVehicles" >> (backpack unit) >> QGVAR(hasWirecutter))} \
+    || {1 == getNumber (configFile >> "CfgWeapons" >> (vest unit) >> QGVAR(hasWirecutter))} \
```
dedmen: Vests are CfgVehicles.

PabstMirror: They are Items/CfgWeapons.
There is a `vestContainer` command, that returns the vest "object"
but `typeOf vestContainer player` = `"Supply140"`

Dystopian: [Vests are CfgWeapons](https://community.bistudio.com/wiki/Arma_3_Characters_And_Gear_Encoding_Guide#Vest_configuration)
[Vest ground holders are CfgVehicles](https://community.bistudio.com/wiki/Arma_3_Characters_And_Gear_Encoding_Guide#Vest_ground_holder)

PR #6465, opened 2018-07-22
Minor fixes

Dystopian:
- Fix example in `ace_map_fnc_simulateMapLight`;
- Make `fnc_addActionToObject` public;
- Repair - Delete unnecessary condition (since #4198).

### DIFF
--- a/addons/repair/functions/fnc_addRepairActions.sqf
+++ b/addons/repair/functions/fnc_addRepairActions.sqf
@@ -128,11 +128,7 @@ private _icon = QPATHTOF(ui\repair_0_ca.paa);
         if (_hitpoint in TRACK_HITPOINTS) then {
             // Tracks should always be unique
             if (_hitpoint in _processedHitpoints) exitWith {TRACE_3("Duplicate Track",_hitpoint,_forEachIndex,_selection);};
-            if (_hitpoint == "HitLTrack") then {
-                _position = compile format ["private _return = _target selectionPosition ['%1', 'HitPoints']; _return set [1, 0]; _return", _selection];
-            } else {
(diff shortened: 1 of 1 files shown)


addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -128,11 +128,7 @@ private _icon = QPATHTOF(ui\repair_0_ca.paa);
         if (_hitpoint in TRACK_HITPOINTS) then {
             // Tracks should always be unique
             if (_hitpoint in _processedHitpoints) exitWith {TRACE_3("Duplicate Track",_hitpoint,_forEachIndex,_selection);};
-            if (_hitpoint == "HitLTrack") then {
-                _position = compile format ["private _return = _target selectionPosition ['%1', 'HitPoints']; _return set [1, 0]; _return", _selection];
-            } else {
-                _position = compile format ["private _return = _target selectionPosition ['%1', 'HitPoints']; _return set [1, 0]; _return", _selection];
-            };
+            _position = compile format ["private _return = _target selectionPosition ['%1', 'HitPoints']; _return set [1, 0]; _return", _selection];
```
dedmen: `private _return = _target selectionPosition ['%1', 'HitPoints']; _return set [1, 0]; _return`
->
`(_target selectionPosition ['%1', 'HitPoints']) vectorMultiply [1,0,1]`?
Just an idea. Not sure if it's a good idea.

Dystopian: Then it will not be minor fix made in browser :) I'll test it and change.

Dystopian: `vectorMultiply` uses number as 2nd parameter, so no changes

dedmen: Duh.. Sometimes I hate BI :D

PR #6466, opened 2018-07-23
cargo stringtable german [WIP]

shukari:
- some german stringtable entries (September 21, 2017 o.O)

INFO: ok know it shows a conflict will fix that

### DIFF
--- a/addons/cargo/stringtable.xml
+++ b/addons/cargo/stringtable.xml
@@ -275,6 +275,7 @@
         </Key>
         <Key ID="STR_ACE_Cargo_space_edenName">
             <English>Cargo Space</English>
+            <German>Frachtraum</German>
             <Italian>Spazio Cargo</Italian>
             <Japanese>カーゴ スペース</Japanese>
             <Chinese>貨物空間</Chinese>
@@ -283,6 +284,7 @@
         </Key>
         <Key ID="STR_ACE_Cargo_space_edenDesc">
             <English>The cargo space available in this vehicle/container</English>
+            <German>Verfügbarer Frachtraum für Fahrzeuge und Container</German>
             <Italian>Lo spazio disponibile in questo veicolo/container</Italian>
             <Japanese>この車両/コンテナでカーゴ スペースを使えるようにします</Japanese>
             <Chinese>設定此載具/集裝箱可裝載多少貨物</Chinese>
@@ -291,6 +293,7 @@
         </Key>
         <Key ID="STR_ACE_Cargo_size_edenName">
             <English>Cargo Size</English>
+            <German>Frachtgröße</German>
             <Italian>Dimensioni Cargo</Italian>
             <Japanese>カーゴ サイズ</Japanese>
             <Chinese>貨物的大小</Chinese>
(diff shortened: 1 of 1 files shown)


addons/cargo/stringtable.xml
```diff
@@ -299,6 +302,7 @@
         </Key>
         <Key ID="STR_ACE_Cargo_size_edenDesc">
             <English>The cargo space required to hold this object (-1 for unloadable)</English>
+            <German>Frachtgröße zum Einladen dieses Objektes (-1 nicht einladbar)</German>
```
dedmen: the `required` part is missing there.
Maybe..
`Benötigter Frachtraum zum...`?

addons/cargo/stringtable.xml
```diff
@@ -329,6 +333,7 @@
         </Key>
         <Key ID="STR_ACE_Cargo_paradropTimeCoefficent">
             <English>Paradrop Time Coffecient</English>
+            <German>Türlast Zeitkoeffizient</German>
```
dedmen: "Türlast"? Never heard of that and no google results. But I see that's already used in `STR_ACE_Cargo_paradropButton`

bux: Ausdruck wird in der BW verwendet.

Blutze: Das is entschieden länger als der längste Text in der englischen Übersetzung (welcher da wäre der erste Eintrag in Nametags): turlastabsetzungsdaueranpassung
Glaub "Türlastabsetzungsdaueranpassung" wär eh besseres Deutsch.


Und `<English>Paradrop Time Coffecient</English>` müsste `<English>Paradrop Time Coefficient</English>` sein.

commy2: The correct translation for "paradrop" is "Fallschirmabwurf" and not "Türlast".
Terms used in the Bundeswehr are very bad for a translation like this. They're not common use in language. This is a German translation and not a Bundeswehrdeutsch translation.

commy2: "Fallschirmabwurf-Zeitkoeffizient"

shukari: Im ok with Fallschirmabwurf.

@dedman my google find thousands of results for Türlast in conection with Flugzeug ^^

dedmen: dunno who you mentioned there but not me :D 
Okey yeah with Flugzeug I have 51 results. Including one where a guy talks about ACE and that  `Türlast` is a bad translation :D

addons/cargo/stringtable.xml
xetra11: "Benötigter Frachtraum um dieses Objekt einzuladen (-1 nicht einladbar)"

*clever clogs* 🥇

addons/cargo/stringtable.xml
xetra11: "Tuerlast" is the right term for dropped cargo out of a plane for the german army. We used this term when we did paradrops and our heavy equipment needed to be dropped separately.

commy2: >"Tuerlast" is the right term for dropped cargo out of a plane for the german army.
See: https://github.com/acemod/ACE3/pull/6466#discussion_r204638133

xetra11: The english translation uses army terms as well. So where is the problem doing the same in german?

commy2: The English version of this string is: "Paradrop Time Coffecient", which is, while very technical, standard English. If there is a bad (=not day to day English) translation anywhere, then it should be changed as well.

>This is a military simulation after all.

And it is also a video game used by civilians. A tooltip is there to be understood by everyone.

bux: @commy2 I disagree

commy2: I'd be fine with the term being used, and defined in an extra sentence, even if that were a feature of the German string only. Tooltips can be quite lengthy and still fit. Aside from that, I rest my case.

PR #6467, opened 2018-07-23
Vary time taken to bandage based on circumstances

SilentSpike:
- Change the time it takes to bandage based on:
  - The size of wound to be targeted (and how much has already been treated)
  - ~~If the head or torso is being bandaged~~
  - Whether the bandage user is a trained medic
  - Whether the patient ~~is unconscious or~~ self treating

Note:
- There are a few goals to this PR:
  - To reward use of medically trained units and teamwork
  - To reduce annoyance of how long it takes to treat minor wounds
  - To improve realism
- The times I've set are arbitrary and just based on what I figured makes sense, probably needs to be tested to check if it feels right.
- This isn't an outright penalty because it can make treatment faster if a wound is already mostly closed up.
- We might also want to factor in wound type (an abrasion is surely quicker to bandage than a velocity wound).
- There is potential for some weirdness if two people bandage the same bodypart of the same unit at the same time. This is due to how locality is currently handled in treatment (the time to bandage is calculated local to the medic, then the actual treatment occurs local to the patient - so the first medic could close the target wound, then the second medic would be bandaging a different wound, but the time was calculated from the now closed wound).

### DIFF
--- a/addons/medical_treatment/functions/fnc_getBandageTime.sqf
+++ b/addons/medical_treatment/functions/fnc_getBandageTime.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: SilentSpike
- * Calculates the time to bandage a wound based on it's location, size, the patient and the medic.
+ * Calculates the time to bandage a wound based on it's size, the patient and the medic.
  *
  * Arguments:
  * 0: The medic <OBJECT>
@@ -28,17 +28,21 @@ if (_wound isEqualTo EMPTY_WOUND) exitWith { 0 };
 
 _wound params ["", "", "", "_amountOf", "_bloodloss", "_damage", "_category"];
 
-// Base bandage time is based on wound category
-private _bandageTime = ([4, 6, 8] select _category) * _amountOf;
+// Base bandage time is based on wound size and remaining percentage
+private _bandageTime = ([
+    BANDAGE_TIME_S,
+    BANDAGE_TIME_M,
+    BANDAGE_TIME_L
+] select _category) * _amountOf;
 
-// Medically unskilled units aren't so practised at applying bandages
-if !([_medic] call FUNC(isMedic)) then {
-    _bandageTime = _bandageTime + 3;
+// Medics are more practised at applying bandages
+if ([_medic] call FUNC(isMedic)) then {
+    _bandageTime = _bandageTime + BANDAGE_TIME_MOD_MEDIC;
 };
 
 // Bandaging yourself requires more work
 if (_medic == _patient) then {
-    _bandageTime = _bandageTime + 2;
+    _bandageTime = _bandageTime + BANDAGE_TIME_MOD_SELF;
 };
 
 // Nobody can bandage instantly
--- a/addons/medical_treatment/functions/fnc_getBandageTime.sqf
+++ b/addons/medical_treatment/functions/fnc_getBandageTime.sqf
@@ -31,18 +31,13 @@ _wound params ["", "", "", "_amountOf", "_bloodloss", "_damage", "_category"];
 // Base bandage time is based on wound category
 private _bandageTime = ([4, 6, 8] select _category) * _amountOf;
 
-// Head and torso wounds take slightly longer to bandage
-if (_bodypart in ["head", "body"]) then {
-    _bandageTime = _bandageTime + ([2, 4] select (_bodypart == "body"));
-};
-
 // Medically unskilled units aren't so practised at applying bandages
 if !([_medic] call FUNC(isMedic)) then {
     _bandageTime = _bandageTime + 3;
 };
 
-// Bandaging yourself or an unconscious person requires more work
-if (_medic == _patient || {IS_UNCONSCIOUS(_patient)}) then {
+// Bandaging yourself requires more work
+if (_medic == _patient) then {
     _bandageTime = _bandageTime + 2;
 };
 
--- a/addons/medical_treatment/functions/fnc_getBandageTime.sqf
+++ b/addons/medical_treatment/functions/fnc_getBandageTime.sqf
@@ -23,7 +23,7 @@ _targetWound params ["_wound", "_woundIndex", "_effectiveness"];
 // Everything is patched up on this body part already
 if (_wound isEqualTo []) exitWith { 0 };
 
-_wound params ["", "", "", "_amountOf", "_bloodloss", "_damage", "_category"]
+_wound params ["", "", "", "_amountOf", "_bloodloss", "_damage", "_category"];
 
 // Base bandage time is based on wound category
 private _bandageTime = ([4, 6, 8] select _category) * _amountOf;
--- a/addons/medical_treatment/functions/fnc_getBandageTime.sqf
+++ b/addons/medical_treatment/functions/fnc_getBandageTime.sqf
@@ -1,0 +1,47 @@
+#include "script_component.hpp"
+/*
+ * Author: SilentSpike
+ * Calculates the time to bandage a wound based on it's location, size, the patient and the medic.
+ *
+ * Arguments:
+ * 0: The medic <OBJECT>
+ * 1: The patient <OBJECT>
+ * 2: Body part <STRING>
+ * 3: Treatment class name <STRING>
+ *
+ * Return Value:
+ * Time in seconds <NUMBER>
+ *
+ * Public: No
+ */
+
+params ["_medic", "_patient", "_bodypart", "_bandage"];
+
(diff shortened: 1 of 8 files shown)


bux:
increase time during rain and being outside?
increase time when in tight space (passenger seats vs. stretcher in the back of a car / chopper)?

SilentSpike:
Good suggestions.

I have some design questions we should probably answer:

1. What's the maximum time we'd want players to wait before it becomes unreasonable and not enjoyable (keeping in mind player input is removed for the duration)? Should there even be a max?
2. Should we be displaying feedback that tells the player what is influencing their treatment time?
3. Should we perhaps add a setting to influence this (a direct coefficient or maybe "realistic" vs "balanced")?

nk3nny:
Without an interface to **clearly** indicate why treatment times vary, I would be opposed to tacking too many variables to healing time.  Why there may be realism factors at work, without these being easily understood and accessible-- the change in time will only be arbitrary and opaque. 

Perhaps the modifiers can be indicated with some sort iconography or text?

Drofseh:
I would prefer not to bother with variable treatment times, especially if it could cause locality issues.

Drofseh:
If you guys are really set on it though, I would have the order of relative time/difficulty be
`Contusion -> Abrasion -> Puncture -> Velocity -> Cut -> Laceration -> Crush -> Avulsion`
Based on how much needs to be held together while bandaging.

SilentSpike:
After thinking about it, I'm against adding factors beyond those in this PR to treatment time because the effect on gameplay isn't engaging enough to be worth the effort it would take to communicate them with the player.

I'd maybe even scale this back slightly - removing the penalty for unconscious patients and head/torso wounds as they are flat penalties that can't be escaped. Whereas penalising self-treatment encourages buddy healing, penalising non-medics encourages medic use and scaling with wound size removes some of the drag associated with treating many small wounds.

I now wonder whether we should be reducing the base time for medics/others rather than increasing it for non-medics/self.

addons/medical_treatment/functions/fnc_getBandageTime.sqf
```diff
@@ -0,0 +1,50 @@
...
+private _targetWound = [_patient, _bandage, _partIndex] call FUNC(findMostEffectiveWound);
+_targetWound params ["_wound", "_woundIndex", "_effectiveness"];
+
+// Everything is patched up on this body part already
+if (_wound isEqualTo EMPTY_WOUND) exitWith { 0 };
+
+_wound params ["", "", "", "_amountOf", "_bloodloss", "_damage", "_category"];
+
+// Base bandage time is based on wound category
+private _bandageTime = ([4, 6, 8] select _category) * _amountOf;
```
TheMagnetar: I think these numbers should be declared as defines for easy adjustment.

addons/medical_treatment/functions/fnc_getBandageTime.sqf
```diff
@@ -0,0 +1,50 @@
...
+
+// Head and torso wounds take slightly longer to bandage
+if (_bodypart in ["head", "body"]) then {
+    _bandageTime = _bandageTime + ([2, 4] select (_bodypart == "body"));
+};
+
+// Medically unskilled units aren't so practised at applying bandages
+if !([_medic] call FUNC(isMedic)) then {
+    _bandageTime = _bandageTime + 3;
```
TheMagnetar: Same as above. It could be even a setting.

addons/medical_treatment/functions/fnc_getBandageTime.sqf
dedmen: Also maybe some comment about which category has which values?

addons/medical_treatment/functions/fnc_treatmentBandageLocal.sqf
```diff
@@ -14,92 +14,33 @@
...
-        if (_woundEffectiveness * _amountOf * _bleeding > _effectivenessFound * (_mostEffectiveInjury select 3) * (_mostEffectiveInjury select 4)) then {
-            _effectivenessFound = _woundEffectiveness;
-            _mostEffectiveSpot = _forEachIndex;
-            _mostEffectiveInjury = _x;
-        };
-    };
-
-    if (_exit) exitWith {};
-} forEach _openWounds;
+private _targetWound = [_target, _bandage, _partIndex] call FUNC(findMostEffectiveWound);
```
thojkooi: This change breaks being able to bandage a specific type of injury, due to the removal of:

```sqf
-        TRACE_2("Wound classes: ", _specificClass, _classID);
-        if (_specificClass == _classID) exitWith {
-            _effectivenessFound = _woundEffectiveness;
-            _mostEffectiveSpot = _forEachIndex;
-            _mostEffectiveInjury = _x;
-            _exit = true;
-        };
```

addons/medical_treatment/functions/fnc_treatmentBandage.sqf
```diff
@@ -8,21 +8,20 @@
...
  *
  * Return Value:
  * Succesful treatment started <BOOL>
  *
- * Public: Yes
+ * Public: No
  */
 #include "script_component.hpp"
 
-params ["_caller", "_target", "_bodyPart", "_className", "_items", "", ["_specificSpot", -1]];
```
thojkooi: Same here, removing ` ["_specificSpot", -1]` means we are dropping support for selecting the injury to bandage in the UI.

addons/medical_treatment/functions/fnc_getBandageTime.sqf
```diff
@@ -0,0 +1,50 @@
...
+};
+
+// Bandaging yourself or an unconscious person requires more work
+if (_medic == _patient || {IS_UNCONSCIOUS(_patient)}) then {
```
thojkooi: I'd remove the unconscious condition. I am not sure how much value this would add.

addons/medical_treatment/functions/fnc_treatmentBandageLocal.sqf
SilentSpike: So before removing this I looked back through the function calls and couldn't find anywhere that was using that parameter 🤔

thojkooi: You are right - it's also not functional on the master branch. Seems like this is a bug we introduced somewhere along the way, though I couldn't find out at what point it broke, if it ever worked at all.. I do know it's intended to work through the medical menu. It was added as part of #172 in f6fbef980a.

Seeing as it's currently broken, I'm ok with removing this code and creating a seperate issue to investigate implementing the correct functionality for that (or decide on not doing it at all).

SilentSpike: Sounds good, my thoughts exactly

SilentSpike:
Fulfilled requested changes and re-balanced the values for medics and self treatment. The current PR produces the following result:

  |   | Self Time | Self Time | Buddy Time | Buddy Time
-- | -- | -- | -- | -- | --
Wound Size | Base Time | Non-Medic | Medic | Non-Medic | Medic
S | 4s | 8s | 6s | 4s | 2s
M | 6s | 10s | 8s | 6s | 4s
L | 8s | 12s | 10s | 8s | 6s

Drofseh:
Are large wounds still going to require multiple bandages to close?
Likewise, will a bandage still be able to close multiple small wounds?

---

If yes, then that will significantly skew the total time in which a given number of wounds can be bandaged.

For example, if a large wound takes 2 bandages to close, and each bandage takes 8 seconds, then closing one large wound takes 16 seconds.

By comparison, eight small wounds could be bandaged in the same 16 seconds (assuming one bandage will close two small wounds (16 seconds / 4 seconds per bandage * 2 wounds fixed per bandage).

Note that this is based on Base Time and doesn't include time spent in the UI selecting the wound and bandage.

This seems a little disproportionate to me.

thojkooi:
>  I don't know if this was ever implemented in the old medical system

No, it was not. This was a feature part of the CSE medical system that never made it into ACE.

Drofseh:
> The first would take 8s and the second would take 4s (assuming first closes 50% exactly).

That is cool then.

> Currently bandages don't close multiple wounds.

Are you talking about in the rewrite, because in the current stable release bandages in adv medical absolutely can treat more than one would at a time.

Anything with an `effectiveness` of >1 for a given wound type.
For example, one field dressing here will completely bandage three minor abrasions.
https://github.com/acemod/ACE3/blob/master/addons/medical/ACE_Medical_Treatments.hpp#L709

It's something I address in my PR #4367

PR #6470, opened 2018-07-24
Introduce bleedout condition and fix IV hemorrhage

### LINKED ISSUE #5510: Fatal Damage - only headshots would kill a downed unit

SilentSpike:
- Add a death condition for units who've lost more than 50% of their original blood volume
- Change the blood volume cardiac arrest condition to occur after 40% is lost (class IV hemorrhage) to emulate hypovolemic shock

Source on those values:
https://www.medicaldaily.com/breaking-point-how-much-blood-can-human-body-lose-350792

This PR relates to #6277 in that the fatal injury system really is designed to emulate damage to vital organs and so allowing it to occur on any bodypart isn't using it as intended. Instead, if we introduce a bleedout death condition we fix #5510 via another means which makes more sense. #6277 can then be re-purposed to allow fatal injury on the torso to emulate damage to organs located there.

We probably need to investigate the current bleed rates in the rewrite and tweak the factors on those to be balanced to these realistic limits on blood volume.

### DIFF
--- a/addons/medical_statemachine/Statemachine.hpp
+++ b/addons/medical_statemachine/Statemachine.hpp
@@ -105,6 +105,10 @@ class ACE_Medical_StateMachine {
             condition = QFUNC(conditionExecutionDeath);
             events[] = {QEGVAR(medical,FatalInjury)};
         };
+        class Bleedout {
+            targetState = "Dead";
+            events[] = {QEGVAR(medical,Bleedout)};
+        };
     };
     class Dead {
         // When the unit is killed it's no longer handled by the statemachine
--- a/addons/medical_engine/script_macros_medical.hpp
+++ b/addons/medical_engine/script_macros_medical.hpp
@@ -28,7 +28,7 @@
 #define BLOOD_VOLUME_CLASS_2_HEMORRHAGE 5.100 // lost more than 15% blood, Class II Hemorrhage
 #define BLOOD_VOLUME_CLASS_3_HEMORRHAGE 4.200 // lost more than 30% blood, Class III Hemorrhage
 #define BLOOD_VOLUME_CLASS_4_HEMORRHAGE 3.600 // lost more than 40% blood, Class IV Hemorrhage
-#define BLOOD_VOLUME_FATAL 3.0 // Lost more than 50% blood, Fatal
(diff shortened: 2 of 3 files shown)


addons/medical_statemachine/Statemachine.hpp
```diff
@@ -105,6 +105,10 @@ class ACE_Medical_StateMachine {
             condition = QFUNC(conditionExecutionDeath);
             events[] = {QEGVAR(medical,FatalInjury)};
         };
+        class Bleedout {
```
thojkooi: This transition needs to be added to the other states as well. Otherwise if blood volume drops significantly you could potentially by pass the cardiac arrest state and end up with bleedout events without transitioning into a different state.

SilentSpike:
Good catch, I've now updated both places where hemorrhage value is used in the UI (moved the strings associated over while I was at it). 👍 

I'll probably change the way those are done in another PR as currently it seems a little off.

addons/medical_gui/stringtable.xml
```diff
@@ -844,6 +844,34 @@
             <Chinesesimp>大量失血</Chinesesimp>
             <Chinese>大量失血</Chinese>
         </Key>
+        <!--
+            Strings above and below this seem to differ in some languages, determine which is best to use
+        -->
+        <Key ID="STR_ACE_Medical_GUI_Lost_Blood1">
```
TheMagnetar: After seeing the text strings, I would rename this one as "GUI_Lost_BloodGeneric" to avoid confusion with Blood2 message. Blood1 is "Lost a lot of blood", Blood2 is "Lost some blood".

This is a minor wish. so PR approved

SilentSpike: Yeah this is what I meant about them being off, I'll make a separate PR to fix these

Issue #6458, opened 2018-07-20
Differentiate PAK and Medkit


SilentSpike:
An area of the medical UX that could use improvement. These item names are ambiguous and so similar that their use is not immediately self explanatory.

What ShacTac have done:
- Personal Aid Kit -> Basic Aid Kit
- Medkit -> Advanced Aid Kit

I quite like this personally, but perhaps other people feel that it's not realistic enough? Please weigh in with any thoughts!

BaerMitUmlaut:
There's no medkit item?

jonpas:
Surgical Kit maybe?

BaerMitUmlaut:
That does something completely different though? I'm confused now.

thojkooi:
I think there is a medkit in vanilla? Though that is switched out by medical supplies on init.

dedmen:
It's called "First Aid Kit" in vanilla.

SilentSpike:
I admittedly didn't check the item list and just brought this up based on ShacTac's feedback. However, I just looked now and see this:

```

    class ACE_medKit: ACE_ItemCore {
        scope = 2;
        displayName = CSTRING(Med_Kit_Display);
        picture = QPATHTOF(ui\items\personal_aid_kit_x_ca.paa); // todo: new picture
        descriptionShort = CSTRING(Med_Kit_Desc_Short);
        descriptionUse = CSTRING(Med_Kit_Desc_Use);
        class ItemInfo: CBA_MiscItem_ItemInfo {
            mass = 10;
        };
    };
    class ACE_personalAidKit: ACE_ItemCore {
        scope = 2;
        author = ECSTRING(common,ACETeam);
        displayName = CSTRING(Aid_Kit_Display);
        picture = QPATHTOF(ui\items\personal_aid_kit_x_ca.paa);
        descriptionShort = CSTRING(Aid_Kit_Desc_Short);
        descriptionUse = CSTRING(Aid_Kit_Desc_Use);
        class ItemInfo: CBA_MiscItem_ItemInfo {
            mass = 10;
        };
    };
```

thojkooi:
So it looks like it's something that was newly introduced in the medical rewrite? I don't remember any thing about it.

It apparently is used for a partial heal treatment action. I don't think we should keep this in the rewrite. Judging from the code, it defeats some of the other gameplay mechanics.

SilentSpike:
I would be on board with adding back a "Bandages Heal Damage" setting and removing the medkit

[...]

PR #6473, opened 2018-07-25
 refactor(medical): remove medkit due to confusing functionality

thojkooi:
Due to be replaced by an alternative treatment option such as a split.

PR made to support the effort in #6458.


SilentSpike:
Do we want to stick to removing this? One of the suggestions on that issue which I kind of like:

> Introduce single-use and multi-use full heal items and remove the setting to toggle PAK behaviour

I personally quite like this diegetic approach (as nk3nny puts it). I'm definitely in favour of not naming it a Medkit though

SilentSpike:
Yeah I'm merging this now and we can look into the multi-use stuff as part of the settings review 👍

PR #186, opened 2015-03-10
add repair PBO, fix #185

commy2:
- [x] Replacing wheels logic
- [x] Tracks
- [x] Localizations

- [x] Toolkit, engineer, disable default repair
- [x] progress bar abort conditions (correct wheel is in nearest objects)

PR #6478, opened 2018-07-29
Repair - Fix facility is like vehicle

Dystopian:
- (1) fix config-based facility acts as repair vehicles;
- fix repair at dead facility (if it's possible);
- optimize `fnc_isInRepairFacility` a little.

(1) Config-based facilities have the same config value name as repair vehicles (`GVAR(canRepair)`). When there is a dead config-based repair vehicle near it becomes repair facility and can affect repair routines. This PR excludes `AllVehicles` child classes from facilities. Maybe simulation check would be more efficient but I think it's OK.

### DIFF
--- a/addons/repair/functions/fnc_isInRepairFacility.sqf
+++ b/addons/repair/functions/fnc_isInRepairFacility.sqf
@@ -21,19 +21,22 @@ TRACE_1("params",_object);
 private _position = getPosASL _object;
 private _isInBuilding = false;
 
-private _objects = (lineIntersectsWith [_object modelToWorldVisual [0, 0, (_position select 2)], _object modelToWorldVisual [0, 0, (_position select 2) +10], _object]);
-{
-    if (_x getVariable ["ACE_isRepairFacility", getNumber (configFile >> "CfgVehicles" >> typeOf _x >> QGVAR(canRepair))] > 0) exitWith {
+private _checkObject = {
+    if (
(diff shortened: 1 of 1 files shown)


PabstMirror:
I'm not sure, but should we also change the default setting to `vehicleAndFacility`?
This would ensure consistent behaviour with previous versions

https://github.com/acemod/ACE3/blob/master/addons/repair/initSettings.sqf#L61
2->3

Dystopian:
Looks like it was always default `repairVehicleOnly` and now it's the same. It was added by @thojkooi in #186:
https://github.com/acemod/ACE3/blob/83f46be990f8af12af3abfc66d7b6d449e8f9a63/addons/repair/ACE_Settings.hpp#L47-L48
and at the same time module had `repairFacilityOnly` as default: https://github.com/acemod/ACE3/blob/83f46be990f8af12af3abfc66d7b6d449e8f9a63/addons/repair/CfgVehicles.hpp#L80
So I think at least his opinion is important.

SilentSpike:
@Dystopian You mean `repairFacilityOnly` in the first sentence, right?

Dystopian:
@SilentSpike No, I meant exactly `repairVehicleOnly`. I decided value is counted from `1` because near https://github.com/acemod/ACE3/blob/83f46be990f8af12af3abfc66d7b6d449e8f9a63/addons/repair/ACE_Settings.hpp#L54-L55

Dystopian:
And it was fixed in #2364 :laughing: Then `repairFacilityOnly` was default.

PR #6483, opened 2018-07-29
WIP: Have viscosity matter when calculating bloodloss

thojkooi:
This is an experimental change. When using medication, it will decrease visocity. This in turn will now affect bloodloss due to two factors; 
- Heart rate changes.
- Blood viscosity changes.

Now the more morphine a player uses, the more increase in blood loss. This is an insentive to bandage before applying medication.


SilentSpike:
I like this change, however I have no idea how the calculations for peripheral resistance work in our code (where do the `2`, `100` and `0.1` come from? Is it a percentage value that's scaled by 100 in the variable?).

I also find it confusing that our code refers to peripheral resistance rather than viscosity. Viscosity does influence peripheral resistance, but so do vasoconstrictors and vasodilators. It would be good to clarify what the code is actually simulating (I believe it's actually vasoconstrictors and vasodilators).

PabstMirror:
will look into how to balance this

PabstMirror:
- Should blood/plasma/saline have an effect on this as well?
- We should standardize between viscosity or peripheralResistance
- `min` seems to lead to odd results

use morphine and 
player getVariable ["ace_medical_peripheralResistance", 100] = 90.1781

example (r = 90.1781)
(2 - (r/100)) = 1.09822
(2 - (r/100) min 0.1) = 0.1

Mike-MF:

PR #6485, opened 2018-07-30
[medical] Code cleanup using SQFLint

dedmen:
This is some minor cleanup of SQFLint warnings. And one actual bug of a wrong variable being used.
I also want to cleanup the spacing after the header.

Tasks left that I want in here.

**1.** 

We have mixed
```
<header>
params []
```
```
<header>

params []
```

```
<header>


params []
```
Through the files. I'd clean it up to a single newline after the header. Then maybe file local #defines, another newline and then params. But for that I need #6484 merged first.

**2.** 
Also there are tons of unused variables. Especially `params` with all entries defined although not all are needed.
I'd clean them up like this
`params ["_var1", "_var2" /*, "_var3", "_var4" */]` in case where the last ones aren't needed.
Or
`params ["" /*_var1*/, "" /*_var2*/, "_var3", "_var4"]`
in case where ones in the middle aren't needed. What do you think about that?

**3.** 
https://github.com/acemod/ACE3/blob/c9ad92e92d26f929e3cb78a53743585e5ffa20d8/addons/medical_treatment/functions/fnc_checkItems.sqf#L17
This is horrible. As I already wrote in Slack.
>Maybe add a ACE common function like. "getNumberOfItem" the returns the count of how many of a specific item type the unit has in inventory?
Would we want weaponItems or assignedItems for that too? Maybe only on condition like in the CBA func?

**4.** 
https://github.com/acemod/ACE3/blob/c9ad92e92d26f929e3cb78a53743585e5ffa20d8/addons/medical_treatment/functions/fnc_actionUnloadUnit.sqf#L20
the `_drag` thing is not implemented. Shouldn't such things be marked as `//#TODO` in the code to make them easier to find and harder to forget?


dedmen:
Last commit has my proposed solution for **3.**
But it's very specialized. It could be generalized further by adding weaponItems/assignedItems as I said in first post.

addons/medical_treatment/functions/fnc_getTriageStatus.sqf
```diff
@@ -14,10 +14,10 @@
  * Public: Yes
  */
 
-private ["_unit","_return","_status"];
 params ["_unit"];
-_status = _unit getVariable [QEGVAR(medical,triageLevel), -1];
-_return = switch (_status) do {
+
+private _status = _unit getVariable [QEGVAR(medical,triageLevel), -1];
+private _return = switch (_status) do {
```
SilentSpike: The use of a return variable is pointless in this function since the switch is at the end anyway 👍

dedmen: How about replacing the switch by an `ARRAY param [_status, default]` ?

dedmen: Nah. Would then call `localize` for all entries everytime

SilentSpike:
I like the common function solution 👍

addons/common/functions/fnc_getCountOfItem.sqf
```diff
@@ -0,0 +1,31 @@
+#include "script_component.hpp"
+/*
+ * Author: Dedmen
+ * Return how many items of type _itemType the player has in his containers (Uniform, Vest, Backpack)
+ * Doesn't count assignedItems, weapons, weapon attachments, magazines in weapons
+ *
+ * Arguments:
+ * 0: unit <OBJECT>
```
bux: `Unit`

addons/common/functions/fnc_getCountOfItem.sqf
```diff
@@ -0,0 +1,31 @@
...
+ * 1: Classname of item (Case-Sensitive) <STRING>
+ *
+ * Return Value:
+ * itemCount <NUMBER>
```
jonpas: Item Count

jonpas:
I don't like the empty line between header and file local defines. Include above header has no newline, amd header separates it nicely already.

addons/common/functions/fnc_getCountOfItem.sqf
dedmen: Thought the same actually. But I just copied it from another function :D

dedmen:
@jonpas so.. #define directly after header without newline? Or before header?

jonpas:
After, without the newline I'd say.

SilentSpike:
@dedmen Is this ready for review?

dedmen:
yes. I'll fix the conflicts now. Nothing else I want to do here.

PR #6487, opened 2018-07-30
Fix some item type magic numbers

Dystopian:
- title.

These macros are used: https://github.com/acemod/ACE3/blob/2767fddd92bd48199ba580601dd148f77e001f00/addons/main/script_macros.hpp#L65-L68

### DIFF
--- a/addons/arsenal/functions/fnc_addVirtualItems.sqf
+++ b/addons/arsenal/functions/fnc_addVirtualItems.sqf
@@ -81,37 +81,37 @@ if (_items isEqualType true) then {
                         /* Weapon acc */
                         case (
                                 isClass (_configItemInfo) &&
-                                {(getNumber (_configItemInfo >> "type")) in [101, 201, 301, 302]} &&
+                                {(getNumber (_configItemInfo >> "type")) in [TYPE_MUZZLE, TYPE_OPTICS, TYPE_FLASHLIGHT, TYPE_BIPOD]} &&
                                 {!(_x isKindOf ["CBA_MiscItem", (_configCfgWeapons)])}
                             ): {
                             switch (getNumber (_configItemInfo >> "type")) do {
-                                case 201: {
+                                case TYPE_OPTICS: {
                                     (_cargo select 1) select 0 pushBackUnique _x;
                                 };
-                                case 301: {
+                                case TYPE_FLASHLIGHT: {
                                     (_cargo select 1) select 1 pushBackUnique _x;
                                 };
-                                case 101: {
+                                case TYPE_MUZZLE: {
                                     (_cargo select 1) select 2 pushBackUnique _x;
                                 };
-                                case 302: {
+                                case TYPE_BIPOD: {
                                     (_cargo select 1) select 3 pushBackUnique _x;
                                 };
                             };
                         };
                         /* Headgear */
                         case (isClass (_configItemInfo) &&
-                            {getNumber (_configItemInfo >> "type") == 605}): {
+                            {getNumber (_configItemInfo >> "type") == TYPE_HEADGEAR}): {
                             (_cargo select 3) pushBackUnique _x;
                         };
                         /* Uniform */
                         case (isClass (_configItemInfo) &&
-                            {getNumber (_configItemInfo >> "type") == 801}): {
+                            {getNumber (_configItemInfo >> "type") == TYPE_UNIFORM}): {
                             (_cargo select 4) pushBackUnique _x;
                         };
                         /* Vest */
                         case (isClass (_configItemInfo) &&
-                            {getNumber (_configItemInfo >> "type") == 701}): {
+                            {getNumber (_configItemInfo >> "type") == TYPE_VEST}): {
                             (_cargo select 5) pushBackUnique _x;
                         };
                         /* NVgs */
@@ -145,7 +145,7 @@ if (_items isEqualType true) then {
                         };
                         /* UAV terminals */
                         case (isClass (_configItemInfo) &&
-                            {getNumber (_configItemInfo >> "type") == 621}): {
+                            {getNumber (_configItemInfo >> "type") == TYPE_UAV_TERMINAL}): {
                             (_cargo select 14) pushBackUnique _x;
                         };
                         /* Weapon, at the bottom to avoid adding binos */
@@ -166,9 +166,9 @@ if (_items isEqualType true) then {
                         /* Misc items */
                         case (
                                 isClass (_configItemInfo) &&
-                                ((getNumber (_configItemInfo >> "type")) in [101, 201, 301, 302] &&
+                                ((getNumber (_configItemInfo >> "type")) in [TYPE_MUZZLE, TYPE_OPTICS, TYPE_FLASHLIGHT, TYPE_BIPOD] &&
                                 {(_x isKindOf ["CBA_MiscItem", (_configCfgWeapons)])}) ||
-                                {(getNumber (_configItemInfo >> "type")) in [401, 619, 620]} ||
+                                {(getNumber (_configItemInfo >> "type")) in [TYPE_FIRST_AID_KIT, TYPE_MEDIKIT, TYPE_TOOLKIT]} ||
                                 {(getText (_configCfgWeapons >> _x >> "simulation")) == "ItemMineDetector"}
                             ): {
                             (_cargo select 17) pushBackUnique _x;
--- a/addons/arsenal/functions/fnc_addVirtualItems.sqf
+++ b/addons/arsenal/functions/fnc_addVirtualItems.sqf
@@ -120,7 +120,7 @@ if (_items isEqualType true) then {
                         };
                         /* Binos */
                         case (_simulationType == "Binocular" ||
-                            {(_simulationType == 'Weapon') && {(getNumber (_configCfgWeapons >> _x >> 'type') == 4096)}}): {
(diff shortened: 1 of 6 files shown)


SilentSpike:
Good stuff, way more readable 👍

No idea about those other types personally

dedmen:
Looking into it..
Interesting 🤔 
```
"PrimaryWeapon"
"SecondaryWeapon"
"RightHand"
"LeftHand"
"Backpack"
"Binocular"
"EarPhones"
"Map"
"Compass"
"Watch"
"Radio"
"GPS"
"Para"
"Scuba"
"Goggles"
"HMD"
"MineDetector"
"Medikit"
"Toolkit"
"Backpack2"
"Headgear"
"Vest"
"Hair"
"NVG"
"Launcher"
"Pistol"
"PistolHolstered"
"ForceTracker"
"Handgun"
"FinL"
"FinR"
```
Left and right hand? ForceTracker? Ear phones?

dedmen:
Found it!
If class has no "ItemInfo" and no "WeaponSlotsInfo" entry.
Then `type` is a mask.
```
#define MaskSlotPrimary     0x00000001  // primary weapons
#define MaskSlotHandGun     0x00000002  // hand gun
#define MaskSlotSecondary   0x00000004  // secondary weapons
#define MaskSlotItem        0x00000F00  // items
#define MaskSlotBinocular   0x00003000  // binocular
#define MaskHardMounted     0x00010000  // hard mounted
#define MaskSlotHandGunItem 0x000000F0  // hand gun magazines
#define MaskSlotInventory   0x001E0000  // hand gun magazines
```
But 4096 is missing 🤔 
The first 3 you can see 
https://github.com/acemod/ACE3/blob/2767fddd92bd48199ba580601dd148f77e001f00/addons/arsenal/functions/fnc_scanConfig.sqf#L105-L113

ah here. 
https://github.com/acemod/ACE3/blob/2767fddd92bd48199ba580601dd148f77e001f00/addons/arsenal/functions/fnc_scanConfig.sqf#L74
Binocular is 4096 but.. According to my stuff it's 0x3000 not 0x1000 🤔 

I'd say just make a `#define TYPE_BINOCULAR 4096` and also the primary/secondary/handgun defines for 1,4,2

Dystopian:
> I'd say just make a #define TYPE_BINOCULAR 4096 and also the primary/secondary/handgun defines for 1,4,2

what about these? https://github.com/acemod/ACE3/blob/2767fddd92bd48199ba580601dd148f77e001f00/addons/common/functions/fnc_getItemType.sqf#L45-L47 https://github.com/acemod/ACE3/blob/2767fddd92bd48199ba580601dd148f77e001f00/addons/common/functions/fnc_getItemType.sqf#L69-L74
The second one looks not like binocular only.

dedmen:
`2^4` -> 0x10 -> 16
`2^8` -> 0x100 -> 256
`2^9` -> 0x200 -> 512
`2^12` -> 0x1000 -> 4096
These should also be replaced by the appropriate macros btw. We are doing constant calculations at runtime here.

I'm failing to find the actual 4096 anywhere... Maybe.. The second one is actually only binocular? I think this data just comes from BIS_fnc_itemType. Maybe these items actually don't exist at all. One should run with tons of mods and check if that part of the code is ever executed.


Duh. I found the 4096. Mask for Binocular is 0x3000. That means it's binary AND'ed together.
And 0x1000 & 0x3000 -> 0x1000 -> true
also 0x2000 & 0x3000 -> 0x2000 -> true

and https://github.com/Dystopian/ACE3/blob/ceb544bd3d8f384cfb540f33f02efe52d47cb825/addons/common/functions/fnc_getItemType.sqf#L80
is `MaskSlotInventory` because
`0x20000 & 0x001E0000` -> 0x20000 -> true.

Meh.. complex stuff.

addons/arsenal/functions/fnc_addVirtualItems.sqf
```diff
@@ -81,37 +81,37 @@ if (_items isEqualType true) then {
...
-                                case 302: {
+                                case TYPE_BIPOD: {
                                     (_cargo select 1) select 3 pushBackUnique _x;
                                 };
                             };
                         };
                         /* Headgear */
                         case (isClass (_configItemInfo) &&
-                            {getNumber (_configItemInfo >> "type") == 605}): {
+                            {getNumber (_configItemInfo >> "type") == TYPE_HEADGEAR}): {
```
commy2: isEqualTo

addons/arsenal/functions/fnc_scanConfig.sqf
```diff
@@ -46,23 +46,23 @@ private _configCfgWeapons = configFile >> "CfgWeapons"; //Save this lookup in va
...
                 {!(configName _x isKindOf ["CBA_MiscItem", (_configCfgWeapons)])}
             ): {
 
             //Convert type to array index
-            (_cargo select 1) select ([201,301,101,302] find _itemInfoType) pushBackUnique _className;
+            (_cargo select 1) select ([TYPE_OPTICS,TYPE_FLASHLIGHT,TYPE_MUZZLE,TYPE_BIPOD] find _itemInfoType) pushBackUnique _className;
         };
         /* Headgear */
-        case (_itemInfoType == 605): {
+        case (_itemInfoType == TYPE_HEADGEAR): {
```
commy2: isEqualTo

addons/common/functions/fnc_getItemType.sqf
```diff
@@ -47,24 +47,24 @@ switch (true) do {
...
-    case (_type == 605): {["item", "headgear"]};
-    case (_type == 611): {["item", "radio"]};
-    case (_type == 616): {["item", "hmd"]};
-    case (_type == 617): {["item", "binocular"]};
-    case (_type == 619): {["item", "medikit"]};
-    case (_type == 620): {["item", "toolkit"]};
-    case (_type == 621): {["item", "uav_terminal"]};
-    case (_type == 701): {["item", "vest"]};
-    case (_type == 801): {["item", "uniform"]};
+    case (_type == TYPE_MUZZLE): {["item", "muzzle"]};
```
commy2: isEqualTo

addons/common/functions/fnc_getItemType.sqf
```diff
@@ -47,24 +47,24 @@ switch (true) do {
...
+    case (_type == TYPE_RADIO): {["item", "radio"]};
+    case (_type == TYPE_HMD): {["item", "hmd"]};
+    case (_type == TYPE_BINOCULAR): {["item", "binocular"]};
+    case (_type == TYPE_MEDIKIT): {["item", "medikit"]};
+    case (_type == TYPE_TOOLKIT): {["item", "toolkit"]};
+    case (_type == TYPE_UAV_TERMINAL): {["item", "uav_terminal"]};
+    case (_type == TYPE_VEST): {["item", "vest"]};
+    case (_type == TYPE_UNIFORM): {["item", "uniform"]};
 
     case (_type == 2^12): {
```
commy2: Looks good so far. Please replace these as well.

addons/arsenal/functions/fnc_addVirtualItems.sqf
Dystopian: these functions can't gain noticeable performance (nanoseconds in comparison to overall microseconds) from this replacement but become really less readable

addons/main/script_macros.hpp
```diff
@@ -60,6 +60,18 @@
...
+#define TYPE_WEAPON_HANDGUN 2
+#define TYPE_WEAPON_SECONDARY 4
+// magazine types
+#define TYPE_MAGAZINE_HANDGUN_AND_GL 16 // mainly
+#define TYPE_MAGAZINE_PRIMARY_AND_THROW 256
+#define TYPE_MAGAZINE_SECONDARY_AND_PUT 512 // mainly
+// more types
+#define TYPE_BINOCULAR_AND_NVG 4096
+#define TYPE_WEAPON_VEHICLE 65536
+#define TYPE_ITEM 131072
```
dedmen: Engine name `WeaponSlotInventory` Is less helpful though.

addons/arsenal/functions/fnc_scanConfig.sqf
```diff
@@ -142,7 +142,7 @@ private _putList = [];
     switch true do {
         // Rifle, handgun, secondary weapons mags
         case (
-                (getNumber (_x >> "type") in [256,512,1536,16]) &&
+                (getNumber (_x >> "type") in [TYPE_MAGAZINE_PRIMARY_AND_THROW,TYPE_MAGAZINE_SECONDARY_AND_PUT,1536,TYPE_MAGAZINE_HANDGUN_AND_GL]) &&
```
dedmen: Can't find out what 1536 is. It's 0x600 and thus falls under `MaskSlotItem`
Would be interesting to know what items fall under that.

Dystopian: some rockets which are not `TYPE_MAGAZINE_SECONDARY_AND_PUT` IIRC

dedmen: The number behind there defines the `SlotCount`.. Don't really understand what it means though. Certainly not it's size.

commy2: Can confirm that A2 inventory size was changed by the type value.

commy2: E.g. 1x1 mag vs 2x1 mag vs 6x2 mag etc.

dedmen: yeah then that's it. This should really be implemened using binary & then... But we don't have that so meh

PR #6492, opened 2018-08-01
[medical-rewrite] Fix cpr condition

Zakant:
- Add a `alive` check to the cpr treatment condition.
- `fnc_treatmentCPR_progress.sqf` checks if the target is alive. If its dead cpr is stopped.
- This leads to the situation where you have cpr as an option but it stops right after the start.

Question here is what is intended behaviour: Should cpr be always available (and the alive checks therefor removed) or should it be only available when the target is alive (current state).
The first option has the downside that it allow players to fast check if somebody is alive (cpr option there => alive, cpr missing => dead).

### DIFF
--- a/addons/medical_treatment/functions/fnc_treatmentCPR_progress.sqf
+++ b/addons/medical_treatment/functions/fnc_treatmentCPR_progress.sqf
@@ -24,5 +24,3 @@ if (_target call EFUNC(common,isAwake)) exitWith {false};
 if !IN_CRDC_ARRST(_target) exitWith {false};
 
 [_target] call FUNC(calculateBlood); // Calculate blood volume. If their is no pulse, nothing happens!
-
-(alive _target) // CPR may only proceed if the patient is not dead


dedmen:
I'd say have CPR always available. Otherwise it would be a cheaty way to check if target is alive.

SilentSpike:
Yeah I'd say let it always be available, I actually want to investigate improving the CPR experience at some point down the line

Zakant:
Im okay with this. However I feel like a medic / doctor should have the possibility to detect if somebody is deceased (maybe a really long taking action, toggleabel by a setting). Otherwise I see already the players that perfectly patch up a corpse.

SilentSpike:
There is a "diagnose" action which tells you

ArwynFr:
@SilentSpike :
> I actually want to investigate improving the CPR experience

You might be interested in getting in touch with https://github.com/Pergor/ADV_Medical

TheMagnetar:
> You might be interested in getting in touch with https://github.com/Pergor/ADV_Medical

I believe this goes in another direction...

Zakant:
I removed the alive checks on both files. Cpr is now always available.

SilentSpike:
Yeah my bad for missing that, just pushed a fix directly

Issue #6452, opened 2018-07-19
Improve medical settings UX


SilentSpike:
The current list of settings is bloated and confusing. As part of the rewrite I intend to:
- [x] Convert to CBA settings
- [x] Add settings to their respective component categories
- [x] Remove any unnecessary settings to streamline UX
- [x] Clarify any confusing/ambiguous settings (whether due to wording or implementation)

SilentSpike:
I've started my review process of the current settings, will be posting comments on this issue with changes I'd like to propose for others to vote or comment on 👍

SilentSpike:
First to the chopping block:

>**Medics setting**
>What is the level of detail prefered for medics?

My thoughts:
- I don't think this setting adds any value to the system. The only place it's used in code is within the `fnc_isMedic` function to change which units are recognised as medics by default (all, vanilla or none).
  - This is redundant because we have an editor attribute to set whether a unit should be considered a medic or not. This provides finer control and no confusion because it's explicit.
  - Having both of these at play can cause unexpected results especially because the setting is very vaguely worded.

Proposal:
- Remove it.

SilentSpike:
Suggesting for removal:
- `GVAR(showPainInMenu)`
  - **Show pain**: Show pain status in the medical menu
  - What it actually does: Determines if the little "In pain" message is shown in the patient preview/overview
  - Why it should be removed: We already have `GVAR(medicSetting_PainVisualization)` which let's you control who can see this message (nobody/medics only/everybody), so this is a duplicate setting.

SilentSpike:
Suggesting for removal:
- GVAR(medicSetting_PainVisualization)
  - **Allow Pain visualization**: Who can see pain in the menu?
  - What it actually does: Let's you control who (nobody/medics only/everybody) can see the little "In pain" message is shown in the patient preview/overview
  - Why it should be removed: Personally I think we should decide on a default behaviour for this and have it be consistent for everyone (my opinion: always show the message if the patient is conscious, only for medics if unconscious)

SilentSpike:
Suggesting for removal:
- `GVAR(enableScreams)`
  - **Enable Screams**: Enable screaming by injured units
  - What it actually does: Nothing, it isn't used anywhere in code.
[...]

PR #6493, opened 2018-08-01
Move medical settings into their respective components

SilentSpike:
Reference: #6452

This PR just handles moving the settings appropriately (and removing the old medic setting + neutering the old settings module).


SilentSpike:
I'm leaving `GVAR(spontaneousWakeUpChance)` where it is for now because there's some weird inter-component control flow going on with `FUNC(setUnconsciousStatemachine)` in `medical_status`. We probably want to move that into `medical_statemachine` and change the way those two components interact to avoid a dependency.

addons/medical/XEH_PREP.hpp
```diff
@@ -1,6 +1,4 @@
 PREP(addDamageToUnit);
-PREP(adjustPainLevel);
```
SilentSpike: One thing of note in this PR is that I had to move this function to `medical_status` and remove the direct calls to it in `medical_damage` to avoid a dependency (it's now hooked into the `[QEGVAR(medical,injured), [_unit, _painLevel]]` event which is already being raised for screaming).

addons/medical_treatment/ACE_Settings.hpp
```diff
@@ -1,11 +1,47 @@
...
+        // todo: this setting just disables some treatment options, remove?
+    };
+    class GVAR(advancedMedication) {
+        displayName = CSTRING(advancedMedication_DisplayName);
+        description = CSTRING(advancedMedication_Description);
+        typeName = "BOOL";
+        value = 1;
+    };
+    // todo: the strings for all three advanced bandages/medication/diagnose settings are terribly ambigious
     class EGVAR(medical,allowLitterCreation) {
```
TheMagnetar: If we have to use EGVAR() for a setting, it either means that the setting is misplaced, or the functionality the setting is repsonsible for is misplaced.  I would suggest moving the allowLiterCreation and all other EGVAR() settings to the respective folder.

Since this is more a suggestion than an actual issue: PR approved.

SilentSpike: All the ones with `EGVAR` are that way for backwards compatibility as they will become `movedToSQF` once I convert everything to CBA settings

SilentSpike: If we want to throw it out the window we can change them all though

TheMagnetar: Understood. I did not consider this.

PR #6495, opened 2018-08-01
[medical] remove collision of bandage litter

dedmen:
I've just observed a car crashing into a floating bandage litter and loosing it's front wheels.
All other litter items don't have collision. Just these.


TheMagnetar:
Why? Is this not a realistic behaviour?

JonBons:
@dedmen I'm not sure if I'm testing this properly but it looks like littergeneric.p3d still has collision.

https://i.imgur.com/XgJdw8p.png
https://i.imgur.com/zetUxTu.jpg

dedmen:
Oh yeah. Must've overlooked that one.

dedmen:
Also removed shadow/collision from normal items as they are too small anyway and performance.
funFact. bodyBagItem's shadowLOD was 90° rotated versus res lod.

PR #6497, opened 2018-08-03
Wiki - Add new Arsenal EHs and stat indexes to arsenal framework doc

alganthe:
- Add the 4 new EHs for ace arsenal to the wiki (Added in 3.12.3)
- Add stat tab index to the wiki.


jonpas:
Maybe add a note on when they were added?

alganthe:
good point.

PR #6509, opened 2018-08-08
[medical-rewrite] Fix wrong macro

Zakant:
- Nightly build failed because the macro was broken (either `EGVAR` with two arguments or `GVAR` with one)
- For backwards compatibility it should be `EGVAR`.

However i would plead to not maintain backwards comp with the rewrite and so change it to `GVAR`, but thats up to you.


SilentSpike:
Good fix 👍 I'm neither for/against backwards compatibility so will leave that decision to @thojkooi

PR #6511, opened 2018-08-08
Fix Japanese stringtable

jonpas:
- Title - RPT error


Dystopian:
maybe some XML checker should be added to CI

PR #6512, opened 2018-08-08
Add param for settings that require mission restart

PabstMirror:
Ref https://github.com/CBATeam/CBA_A3/pull/894

Use the new cba_settings_init param to mark settings as requiring restart
For converted settings the logic is just any global setting
there are some global settings that can be changed mid-mission, and we could think about it when converting.


addons/advanced_fatigue/initSettings.sqf
```diff
@@ -4,7 +4,9 @@
     [LSTRING(Enabled), LSTRING(Enabled_Description)],
     LSTRING(DisplayName),
     true,
-    true
+    true,
+    {[QGVAR(enabled), _this] call EFUNC(common,cbaSettings_settingChanged)},
+    true // Needs mission restart
```
Dystopian: this comment looks quite weird in each file because other params are not commented

dedmen: TBH I would prefer the other `true`'s to be commented too. But on the other end we cannot comment the same things over and over again everywhere.
Or better we probably shouldn't.

PR #6514, opened 2018-08-10
Minor syntax and variable reference fixes for medical rewrite

redbery:
- Fix double setting type for ace_medical_ai_enabledFor setting
- Fix injury instead of injured event called in handelUnitVitals
- Fix multiple instances of functions pointing to old settings locations (before they were moved to separate addons)
- Fix pain being reset and continuous screaming from injured event called by handleUnitVitals
- Fix mistyped macro for isMedicalFacility eden attribute

Note: I'm unsure if the handleUnitVitals is expected to keep sending injured events. I assume this is needed to change the state of the statemachine, but seeing how it caused issues in multiple functions which were not expecting the regular trigger without a pain amount I realize I might have misunderstood the purpose of the event call in handleUnitVitals. It is my first time submitting a pull request so please let me know if I messed something up!

### DIFF
--- a/addons/medical_ai/initSettings.sqf
+++ b/addons/medical_ai/initSettings.sqf
@@ -1,6 +1,5 @@
 [
     QGVAR(enabledFor),
-    "CHECKBOX",
     "LIST",
     [LLSTRING(enableFor_title), LLSTRING(enableFor_desc)],
     LLSTRING(settingCategory),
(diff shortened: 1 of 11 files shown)


thojkooi:
Thank you for your PR @redbery. We will review shortly.

redbery:
I reverted the Injured changes and renamed it to LoweredVitals instead (not a great name but it is somewhat in line with CriticalVitals). I'd agree with the idea of leaving this as a separate event as the causes are very different as well (existing blood loss/pain vs actively being injured).

redbery:
Added the fix for a macro causing a script error I just ran into. Just added it on here as it is still open for review anyway, hope that's ok.

SilentSpike:
Yeah it's a minor fix so that's fine 👍  Merging this as I'm currently working on some stuff and need to test without these bugs in the code 😄

PR #6515, opened 2018-08-11
Fix environmental damage sources

SilentSpike:
- Tidy, optimise and comment handle damage code.
- Fix some environmental damage conditions:
  - ~~There was no sanity check on fall damage to see if the unit was in a vehicle to avoid weird edge cases.~~ Outdated, we now first check for a `_shooter` to handle this.
  - Vehicle crashes would never trigger because they don't fire the EH for all hitpoints like most damage sources do.
  - ~~Being hit by a vehicle would work (unintentionally), but was not explicitly checked for - which is now done by checking if a shooter exists.~~ Outdated, we now only check if a shooter exists to know collision damage has occurred, it's irrelevant if a vehicle caused it.

I tested all of this in the editor and can confirm the added traces for each damage source now fire when expected.

Point of interest: as you can see in the comments, vehicle collisions do fire the EH multiple times and that seems to scale to crash intensity so I think we can just let it happen and avoid complicated code to work around that.

### DIFF
--- a/addons/medical_engine/functions/fnc_handleDamage.sqf
+++ b/addons/medical_engine/functions/fnc_handleDamage.sqf
@@ -88,42 +88,48 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
         _receivedDamage = _damageStructural;
         _woundedHitPoint = "Body";
     };
-    TRACE_2("received",_receivedDamage,_woundedHitPoint);
 
     // Environmental damage sources all have empty ammo string
     // No explicit source given, we infer from differences between them
     if (_ammo isEqualTo "") then {
-        // Fall damage and being hit by vehicle always have a shooter
-        // Check this first because fire damage could happen while falling
+        // Any collision with terrain/vehicle/object has a shooter
+        // Check this first because burning can happen at any velocity
         if !(isNull _shooter) then {
-            // Shooter can be _unit if hit by empty vehicle so check velocity too
-            // A piloted vehicle could hit falling unit so still check shooter
+            _ammo = "#collision";
+
+            /*
+              If shooter != unit then they hit unit, otherwise it could be:
+               - Unit hitting anything at speed
+               - An empty vehicle hitting unit
+               - A physX object hitting unit
+               Assume fall damage for downward velocity because it's most common
+            */
             if (_shooter == _unit && {(velocity _unit select 2) < -2}) then {
-                // Low damage indicates < ~5m fall
-                _woundedHitPoint =  if (_receivedDamage < 0.35) then {
-                    selectRandom ["LeftLeg", "RightLeg"];
-                } else {
-                    selectRandom ["LeftLeg", "RightLeg", "Body", "Head"];
-                };
-                _ammo = "#falling";
-                TRACE_5("Fall Damage",_unit,_shooter,_instigator,_damage,_receivedDamage);
+                _woundedHitPoint = selectRandom ["RightLeg", "LeftLeg"];
+                TRACE_5("Fall",_unit,_shooter,_instigator,_damage,_receivedDamage);
             } else {
-                _woundedHitPoint = "Body";
-                _ammo = "#vehiclecrash";
-                TRACE_5("Collision Damage",_unit,_shooter,_instigator,_damage,_receivedDamage);
+                _woundedHitPoint = selectRandom ["RightArm", "LeftArm", "RightLeg", "LeftLeg"];
+                TRACE_5("Collision",_unit,_shooter,_instigator,_damage,_receivedDamage);
+            };
+
+            // Significant damage suggests high relative velocity
+            // Momentum transfers to body/head for worse wounding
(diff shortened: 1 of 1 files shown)


addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -97,15 +83,17 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
...
         _woundedHitPoint = "Body";
     };
     TRACE_2("received",_receivedDamage,_woundedHitPoint);
 
-    // Check for falling damage.
+    // Check for environmental damage sources (EH doesn't explicitly know)
     if (_ammo isEqualTo "") then {
-        if (velocity _unit select 2 < -2) then {
+        // Downward velocity indicates falling damage
+        if ((vehicle _unit == _unit) && {velocity _unit select 2 < -2}) then {
```
SilentSpike: I think I actually need to swap the fall damage check with the hit by vehicle check (to handle the extreme edge case of being hit by a plane/helicopter while skydiving).

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -157,19 +146,34 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
...
 };
 
-// Handle vehicle crashes
-if (_isCrash) exitWith {
+// Crashing a vehicle doesn't fire the EH for each hitpoint so the "ace_hdbracket" code never runs
+// It does fire the EH multiple times, but this seems to scale with the intensity of the crash
+private _vehicle = vehicle _unit;
+if (
+    _hitPoint isEqualTo "#structural" &&
+    _ammo isEqualTo "" &&
```
TheMagnetar: lazy evaluation also on the second condition {_ammo isEqualTo ""} &&

addons/medical_engine/functions/fnc_handleDamage.sqf
TheMagnetar: This is more a suggestion for code legibility. I like putting brackets to separate commands, etc...: `{((velocity _unit) # 2) < -2 }`

We can  use also `vehicle _unit isEqualTo _unit`

SilentSpike: The quickest way to check if a unit is on foot is actually `_unit in _unit`, we should probably have added a macro for that to CBA ages ago

commy2: `OBJECT in OBJECT` is pretty slow.

SilentSpike:
Did some more testing and debugging and have updated the code to better handle situations where damage could be classified incorrectly (burning while falling and hit by vehicle while falling)

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -88,42 +88,48 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
...
-                _ammo = "#vehiclecrash";
-                TRACE_5("Collision Damage",_unit,_shooter,_instigator,_damage,_receivedDamage);
+                _woundedHitPoint = selectRandom ["RightArm", "LeftArm", "RightLeg", "LeftLeg"];
+                TRACE_5("Collision",_unit,_shooter,_instigator,_damage,_receivedDamage);
+            };
+
+            // Significant damage suggests high relative velocity
+            // Momentum transfers to body/head for worse wounding
+            if (_receivedDamage > 0.35) then {
+                 _woundedHitPoint = selectRandom ["Body", "Head"];
```
SilentSpike: I feel like we should actually raise an additional wounding event in these cases (if you hit the ground at high speed the injury wouldn't fully transfer to your body - you'd still break your legs).

SilentSpike:
So we can probably merge this now, but it definitely needs testing in multiplayer environment (in particular vehicles crashing with multiple players inside - I'm confident everything other than this will behave the same).

addons/medical_engine/functions/fnc_handleDamage.sqf
SilentSpike: Oh really? I figured because it was one less command it'd be faster 🤷‍♂️

thojkooi:
We need to do various test sessions in MP anyway, better we merge this in now, so we can combine that effort.

Anything in particular that makes you say we need to pay attention to those scenario's specifically in MP?

SilentSpike:
Yeah just because a vehicle crashing is the only kind of damage where I could see potential for locality to introduce a change in behaviour. If you're not the driver then the vehicle isn't local to you (I believe) so it's possible that the handle damage EH could fire differently (e.g. maybe it does fire for all hitpoints and not just structural) for such passengers.

ShackTac have also reported that they've observed issues with passengers not taking damage when vehicles crash (example being a helicopter crash) in the dedicated environment - admittedly that's on the old code which in my testing never correctly fired for vehicle crashes at all.

PR #6523, opened 2018-08-12
Medical rewrite bug fixes

redbery:
- Fix medical menu reopen setting
- Fix second delay on execution condition which had a broken macro causing players to fall through cardiac arrest to death immediately 
- Fix people reporting 40 HR (set from leaving cardiac arrest state) when dead
- Fix players having old stats on respawn (including references to arrays causing wounds to the old corpse to apply to the new player too)
- Fix locality issue on unit init causing JiP to reset all variables (wounds etc) on all units
- Fix updateHeartRate being allowed to go passed the target heart rate. This is an especially large issue as it made it completely impossible to perform CPR. During CPR the vitals would not be checked leading to the maximum deltaT of 10. On the first tick of handle vitals once CPR was successfully completed the function would be entered with a HR of 40 and the target HR would be 80. This would make the function 
40 +  ((80 - 40) / 2 )*10 = 240, which would be a HR high enough to push the unit back into cardiac arrest.

Note: I realize the respawn issue has been noted as fixed in #6518, but I can't find the respawn XEH noted in that PR. If I missed or misunderstood this I will of course happily take out my proposed fix. Also my apologies for the seemingly random mix of fixes. These were all the problems we ran into when running a test operation with the rewrite branch today.

### DIFF
--- a/addons/medical_statemachine/functions/fnc_enteredStateDeath.sqf
+++ b/addons/medical_statemachine/functions/fnc_enteredStateDeath.sqf
@@ -19,4 +19,7 @@ params ["_unit"];
 // Send a local event before death
 [QEGVAR(medical,death), [_unit]] call CBA_fnc_localEvent;
 
+_unit setVariable [VAR_HEART_RATE, 0, true];
+_unit setVariable [VAR_BLOOD_PRESS, [0, 0], true];
+
 _unit setDamage 1;
--- a/addons/medical_status/CfgEventHandlers.hpp
+++ b/addons/medical_status/CfgEventHandlers.hpp
@@ -19,7 +19,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_Init_EventHandlers {
     class CAManBase {
         class ADDON {
-            onRespawn = true;
+            onRespawn = 1;
             init = QUOTE(call FUNC(initUnit));
         };
     };
--- a/addons/medical_status/CfgEventHandlers.hpp
+++ b/addons/medical_status/CfgEventHandlers.hpp
@@ -19,6 +19,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_Init_EventHandlers {
     class CAManBase {
         class ADDON {
+            onRespawn = true;
             init = QUOTE(call FUNC(initUnit));
         };
     };
--- a/addons/medical_vitals/functions/fnc_updateHeartRate.sqf
+++ b/addons/medical_vitals/functions/fnc_updateHeartRate.sqf
@@ -67,7 +67,7 @@ if !IN_CRDC_ARRST(_unit) then {
         };
         _targetHR = (_targetHR + _hrTargetAdjustment) max 0;
 
-        _hrChange = round(_targetHR - _heartRate) / 2; 
+        _hrChange = round(_targetHR - _heartRate) / 2;
     } else {
         _hrChange = -round(_heartRate / 10);
     };
--- a/addons/medical_vitals/functions/fnc_updateHeartRate.sqf
+++ b/addons/medical_vitals/functions/fnc_updateHeartRate.sqf
@@ -46,6 +46,7 @@ private _heartRate = GET_HEART_RATE(_unit);
 
 if !IN_CRDC_ARRST(_unit) then {
     private _hrChange = 0;
+    private _targetHR = 0;
     private _bloodVolume = GET_BLOOD_VOLUME(_unit);
     if (_bloodVolume > BLOOD_VOLUME_CLASS_4_HEMORRHAGE) then {
         GET_BLOOD_PRESSURE(_unit) params ["_bloodPressureL", "_bloodPressureH"];
@@ -57,7 +58,7 @@ if !IN_CRDC_ARRST(_unit) then {
             _targetBP = _targetBP * (_bloodVolume / DEFAULT_BLOOD_VOLUME);
         };
 
-        private _targetHR = DEFAULT_HEART_RATE;
+        _targetHR = DEFAULT_HEART_RATE;
         if (_bloodVolume < BLOOD_VOLUME_CLASS_2_HEMORRHAGE) then {
             _targetHR = _heartRate * (_targetBP / (45 max _meanBP));
         };
@@ -68,7 +69,6 @@ if !IN_CRDC_ARRST(_unit) then {
 
         _hrChange = round(_targetHR - _heartRate) / 2; 
     } else {
-        _targetHR = 0;
         _hrChange = -round(_heartRate / 10);
     };
     if (_hrChange < 0) then {
@@ -81,4 +81,3 @@ if !IN_CRDC_ARRST(_unit) then {
 _unit setVariable [VAR_HEART_RATE, _heartRate, _syncValue];
 
 _heartRate
-
--- a/addons/medical_vitals/functions/fnc_updateHeartRate.sqf
+++ b/addons/medical_vitals/functions/fnc_updateHeartRate.sqf
@@ -64,13 +64,18 @@ if !IN_CRDC_ARRST(_unit) then {
         if (_painLevel > 0.2) then {
             _targetHR = _targetHR max (80 + 50 * _painLevel);
         };
-        _targetHR = _targetHR + _hrTargetAdjustment;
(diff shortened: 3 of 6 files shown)


addons/medical_status/CfgEventHandlers.hpp
```diff
@@ -19,6 +19,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_Init_EventHandlers {
     class CAManBase {
         class ADDON {
+            onRespawn = true;
```
commy2: This only accepts number 1, no strings. Booleans don't exist in config.
https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_compileEventHandlers.sqf#L155

addons/medical_vitals/functions/fnc_updateHeartRate.sqf
```diff
@@ -57,23 +58,26 @@ if !IN_CRDC_ARRST(_unit) then {
...
         };
-        _targetHR = _targetHR + _hrTargetAdjustment;
+        _targetHR = (_targetHR + _hrTargetAdjustment) max 0;
 
         _hrChange = round(_targetHR - _heartRate) / 2;
     } else {
         _hrChange = -round(_heartRate / 10);
     };
-    _heartRate = (_heartRate + _deltaT * _hrChange) max 0;
+    if (_hrChange < 0) then {
```
TheMagnetar: You can put the whole new If statement directly under the _hrChange calculation and spare it since it is redundant:
```
   _hrChange = round(_targetHR - _heartRate) / 2;
   _heartRate = (_heartRate + _deltaT * _hrChange) min _targetHR;
} else ...

File: addons/medical_vitals/functions/fnc_updateHeartRate.sqf [hunk 3]
```diff
@@ -46,6 +46,7 @@ private _heartRate = GET_HEART_RATE(_unit);
 
 if !IN_CRDC_ARRST(_unit) then {
```
TheMagnetar: Isn't this preventing the unit from entering when in cardiac arrest?

> On the first tick of handle vitals once CPR was successfully completed the function would be entered with a HR of 40 and the target HR would be 80. This would make the function
>40 + ((80 - 40) / 2 )*10 = 240, which would be a HR high enough to push the unit back into cardiac arrest.

redbery: That is only resetting the statemachine though and not resetting the variables such as openWounds right? My worry is that if the new unit does not override the variables from the old unit on respawn it will have a reference to the actual arrays meaning they will receive any updates to the old arrays (and therefor receive wounds when shooting the old corpse).

addons/medical_status/CfgEventHandlers.hpp
redbery: Sorry about that. Took it from the XEH wiki and didn't notice the define statement 4 lines above it... Should be fixed now.

addons/medical_vitals/functions/fnc_updateHeartRate.sqf
```diff
@@ -57,23 +58,26 @@ if !IN_CRDC_ARRST(_unit) then {
...
-        private _targetHR = DEFAULT_HEART_RATE;
+        _targetHR = DEFAULT_HEART_RATE;
         if (_bloodVolume < BLOOD_VOLUME_CLASS_2_HEMORRHAGE) then {
             _targetHR = _heartRate * (_targetBP / (45 max _meanBP));
         };
         if (_painLevel > 0.2) then {
             _targetHR = _targetHR max (80 + 50 * _painLevel);
         };
-        _targetHR = _targetHR + _hrTargetAdjustment;
+        _targetHR = (_targetHR + _hrTargetAdjustment) max 0;
```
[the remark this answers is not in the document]
redbery: The hrTargetAdjustment can be a rather large negative value depending on current medication I believe. I don't know if it could ever go below 0 because of that but I'd think its worth ensuring it won't. Will remove it if that's really wanted but it seems like taking a bit of a gamble just to save 6 characters.

addons/medical_vitals/functions/fnc_updateHeartRate.sqf
redbery: I'm sorry, I don't think I understand what you are suggesting. The heart rate is currently doing a min or max with _targetHR depending on if we are increasing or decreasing our heart rate. Wouldn't the change you are suggesting here cause any gradual change lowering the HR to happen immediately instead?

addons/medical_status/CfgEventHandlers.hpp
redbery: Just looked in the config viewer as I was wondering why this mistake didn't cause unwanted behavior and noticed that it already got converted from true to 1.  I'll leave the change in for clarity.

addons/medical_status/CfgEventHandlers.hpp
```diff
@@ -19,6 +19,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_Init_EventHandlers {
     class CAManBase {
         class ADDON {
+            onRespawn = 1;
```
SilentSpike: 👍 I overlooked that this wasn't a respawn XEH in my comment on #6518

SilentSpike:
All looks good except I need to look more closely at the last point (there's definitely an unhandled scenario in the code there though) 👍

addons/medical_status/CfgEventHandlers.hpp
commy2: Not all binarizers do this. Mikero converts it despite no define, which causes a lot of other problems.

addons/medical_statemachine/functions/fnc_enteredStateDeath.sqf
```diff
@@ -19,4 +19,7 @@ params ["_unit"];
 // Send a local event before death
 [QEGVAR(medical,death), [_unit]] call CBA_fnc_localEvent;
 
+_unit setVariable [VAR_HEART_RATE, 0, true];
```
thojkooi: This should be done in the medical vitals component through an eventhandler I think.

SilentSpike: Well spotted, I would agree that an EH is better here

redbery: Would we also want to change leftStateCardiacArrest then?

thojkooi: We probably also want to do it for that function, yes. That may require introducing some new events though. Does not have to be done in this PR, imo. But we should create an issue for it if we do not. I will leave that to your discretion @redbery.

redbery: I don't think I'll have the time to do that today as it is somewhat involved which would mean it would need to wait until next weekend. Think it is better to create a new issue for it if you don't mind.

addons/medical_vitals/functions/fnc_updateHeartRate.sqf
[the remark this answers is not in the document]
SilentSpike: This is good, we're just moving the `max 0` up from line 73 where it was previously 👍

addons/medical_vitals/functions/fnc_updateHeartRate.sqf
SilentSpike: After going through the code more closely these lines look good to me 👍

This doesn't fix the issue of incorrectly large `deltaT` first tick after cardiac arrest, but you're right that we don't want the heart rate to pass the target anyway so it should stay because it does fix that issue (theoretically `deltaT` could reach 10 at any time).

For the exiting cardiac arrest `deltaT` issue, I think we should still be running `handleUnitVitals` while in cardiac arrest because it appears to handle other things not related to heart rate too, if we don't then we should refresh the stored `lastTimeUpdated` value on the unit when exiting cardiac arrest to avoid similar issues to the one revealed here.

thojkooi:
@SilentSpike yes, sounds good.

SilentSpike:
I've opened two issues so we don't forget the issues touched upon here:
#6530
#6531

PR #6524, opened 2018-08-12
Prevent treatment animation on dead bodies

SilentSpike:
This was causing bodies to glitch out on dedicated servers according to feedback.

### DIFF
--- a/addons/medical_treatment/functions/fnc_treatment.sqf
+++ b/addons/medical_treatment/functions/fnc_treatment.sqf
@@ -69,7 +69,7 @@ if (IS_UNCONSCIOUS(_target) && EGVAR(medical,allowUnconsciousAnimationOnTreatmen
 
 private _isSelf = _caller isEqualTo _target;
 
-if (!_isSelf && {vehicle _target == _target} && {_patientAnim != ""}) then {
+if (!_isSelf && {vehicle _target == _target} && {_patientAnim != ""} && {alive _target}) then {
     if IS_UNCONSCIOUS(_target) then {
         [_target, _patientAnim, 2, true] call EFUNC(common,doAnimation);
     } else {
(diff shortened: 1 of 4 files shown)


addons/medical_treatment/functions/fnc_treatment.sqf
```diff
@@ -69,7 +69,7 @@ if (IS_UNCONSCIOUS(_target) && EGVAR(medical,allowUnconsciousAnimationOnTreatmen
 
 private _isSelf = _caller isEqualTo _target;
 
-if (!_isSelf && {vehicle _target == _target} && {_patientAnim != ""}) then {
+if (!_isSelf && {vehicle _target == _target} && {_patientAnim != ""} && {alive _target}) then {
```
thojkooi: we should probably do an `isAwake` check instead.

commy2: May only apply to actually dead bodies, not ragdolls?

SilentSpike: To be honest I've never actually noticed treatment animations before 😄 Maybe that means they're doing their job though

I'm in favour of an isAwake check 👍

thojkooi: Treatment animations have always been disabled by default iirc. Due to some very valid reasons documented in a pr way back. Or maybe just slack?

SilentSpike:
Since we're now using an isAwake check the unconscious treatment animation setting is now obsolete (I suggested it for removal [here](https://github.com/acemod/ACE3/issues/6452#issuecomment-409706676)), do you want me to remove that in this PR also?

thojkooi:
@SilentSpike separate PR I think. All changes in this PR now look good.

PR #6525, opened 2018-08-12
Add treatment notifications

SilentSpike:
One of ShackTac's added features, these provide the player with feedback to show that they are being treated.

I imagine we will want to provide a setting for these, but I think for development's sake we should worry about adding new settings at the end when we know everything is working without them.

### DIFF
--- a/addons/medical_treatment/XEH_postInit.sqf
+++ b/addons/medical_treatment/XEH_postInit.sqf
@@ -26,6 +26,9 @@ if (isServer) then {
 [QGVAR(initiated), {
     params ["_caller", "_target"];
 
+    // Don't track self-treatment
+    if (_caller isEqualTo _target) exitWith {};
+
     // Add the new responder
     private _responders = _target getVariable [QGVAR(responders), []];
     _responders pushBackUnique _caller;
@@ -40,6 +43,9 @@ if (isServer) then {
 // Remove tracked players when they stop treating
 private _treatmentEnded = {
     params ["_caller", "_target"];
+
+    // Don't track self-treatment
+    if (_caller isEqualTo _target) exitWith {};
 
     // Remove the old responder
     private _responders = _target getVariable [QGVAR(responders), []];
@@ -64,12 +70,6 @@ private _treatmentEnded = {
 };
 [QGVAR(success), _treatmentEnded] call CBA_fnc_addEventHandler;
 [QGVAR(failure), _treatmentEnded] call CBA_fnc_addEventHandler;
-
-// We stop tracking who is treating a dead patient
-[QEGVAR(medical,death), {
-    params ["_unit"];
-    _unit setVariable [QGVAR(responders), nil];
-}] call CBA_fnc_addEventHandler;
 
 // Update the responders array upon switching unit
 ["unit", {
--- a/addons/medical_treatment/XEH_postInit.sqf
+++ b/addons/medical_treatment/XEH_postInit.sqf
@@ -73,21 +73,26 @@ private _treatmentEnded = {
 
 // Update the responders array upon switching unit
 ["unit", {
-    params ["_unit"];
+    params ["_newUnit", "_oldUnit"];
+
+    // In case of respawn nobody is treating you
+    if !(alive _oldUnit) exitWith {
+        _newUnit setVariable [QGVAR(responders), nil];
+    };
 
     // See comments in _treatmentEnded for details
-    private _responders = _unit getVariable [QGVAR(responders), []];
+    private _responders = _newUnit getVariable [QGVAR(responders), []];
     _responders = _responders select {
         !isNull _x &&
         {[_x] call EFUNC(common,isPlayer)}
     };
-    _unit setVariable [QGVAR(responders), _responders];
+    _newUnit setVariable [QGVAR(responders), _responders];
 
     // Raise a notification event if the unit switched to is already being treated
     if !(_responders isEqualTo []) then {
         [
             QEGVAR(medical_feedback,treatment_initiated),
-            [_responders select 0, _unit]
+            [_responders select 0, _newUnit]
         ] call CBA_fnc_localEvent;
     };
 }] call CBA_fnc_addPlayerEventHandler;
--- a/addons/medical_treatment/XEH_postInit.sqf
+++ b/addons/medical_treatment/XEH_postInit.sqf
@@ -27,21 +27,24 @@ if (isServer) then {
     params ["_caller", "_target"];
 
     // Notifications don't make sense for self-healing or when dead
-    if (_target != ACE_player || {_target == _caller} || {!alive _target}) exitWith {};
+    if (_target == _caller || {!alive _target}) exitWith {};
 
-    // Can't tell who is providing aid when you're unconscious
-    private _name = if IS_UNCONSCIOUS(_target) then {
-        LLSTRING(UnknownPerson);
-    } else {
-        name _caller
-    };
-
-    QGVAR(notification) cutText [format [LLSTRING(Notification), _name], "PLAIN DOWN"];
-
-    // We track who is currently treating the unit
+    // Tracking who is responding lets us show notifications in future too
     private _responders = _target getVariable [QGVAR(responders), []];
-    _responders pushBack _caller;
+    _responders pushBackUnique _caller;
     _target setVariable [QGVAR(responders), _responders];
+
+    // Player only cares if they are the patient
+    if (_target == ACE_player) then {
+        // Can't tell who is providing aid when you're unconscious
+        private _name = if IS_UNCONSCIOUS(_target) then {
+            LLSTRING(UnknownPerson);
+        } else {
+            name _caller
+        };
+
+        QGVAR(notification) cutText [format [LLSTRING(Notification), _name], "PLAIN DOWN"];
+    };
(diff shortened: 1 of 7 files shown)


addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,60 @@ if (isServer) then {
...
+// Notification may no longer apply when treatment ends
+private _treatmentEnded = {
+    params ["_caller", "_target"];
+
+    // Notifications don't make sense for self-healing or when dead
+    if (_target != ACE_player || {_target == _caller} || {!alive _target}) exitWith {};
+
+    // Remove from tracking (in place)
+    private _responders = _target getVariable [QGVAR(responders), []];
+    _responders deleteAt (_responders find _caller);
```
thojkooi: This logic is missing an edge case for handling `_responders` items that have never raised an event for `_treatmentEnded`; i.g. on network issues. Do we wish to handle this here as well?

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,60 @@ if (isServer) then {
...
+    if (_target != ACE_player || {_target == _caller} || {!alive _target}) exitWith {};
+
+    // Can't tell who is providing aid when you're unconscious
+    private _name = if IS_UNCONSCIOUS(_target) then {
+        LLSTRING(UnknownPerson);
+    } else {
+        name _caller
+    };
+
+    QGVAR(notification) cutText [format [LLSTRING(Notification), _name], "PLAIN DOWN"];
```
thojkooi: I'm not to fond of the `cutText` bit. Do we have alternatives for displaying the notification?

Maybe the `displayIcon` framework might work, as an optional setting? This is a bit less in your face, which might be nice.

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,60 @@ if (isServer) then {
...
+
+    // We track who is currently treating the unit
+    private _responders = _target getVariable [QGVAR(responders), []];
+    _responders pushBack _caller;
```
thojkooi: This is lacking a sanity check against duplicate entries. Also another edge case that should almost never happen, but may be nice to handle here.

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,60 @@ if (isServer) then {
...
+    _target setVariable [QGVAR(responders), _responders];
+}] call CBA_fnc_addEventHandler;
+
+// Notification may no longer apply when treatment ends
+private _treatmentEnded = {
+    params ["_caller", "_target"];
+
+    // Notifications don't make sense for self-healing or when dead
+    if (_target != ACE_player || {_target == _caller} || {!alive _target}) exitWith {};
```
thojkooi: in that case, it should wrap the notifications logic in this conditional. I see a bunch of other logic that may be used for more generic purposes that should run regardless of this condition. i.g. the `deleteAt` part.

addons/medical_treatment/XEH_postInit.sqf
SilentSpike: Yeah this is something we definitely want to handle, I've assumed that the medic being killed mid-treatment would trigger the treatment failure condition, but we probably need to filter out any `null` units too for disconnections.

I'm not entirely sure how best to handle a player disconnection who is then taken over by an AI (as they'd then no longer be treating).

addons/medical_treatment/XEH_postInit.sqf
SilentSpike: `pushBackUnique` nicely fixes this

addons/medical_treatment/XEH_postInit.sqf
SilentSpike: I'm not familiar with the `displayIcon` stuff and will need to take a look, I do like the idea of allowing icons vs text.

As for an alternative to `cutText` the only other options I can think of are to add a custom `cutRsc` that's more subtle or using a hint/displayStructuredText (which I'd be against as they're even more in your face).

addons/medical_treatment/XEH_postInit.sqf
SilentSpike: `deleteAt` doesn't need to run for dead or self healing since self healing doesn't add the unit and in the first place and upon death the whole array is deleted. However, I'll split out the player check and put that before the notification

addons/medical_treatment/XEH_postInit.sqf
thojkooi: We'd probably have to integrate a timeout for it?

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,92 @@ if (isServer) then {
...
+    // Notifications don't make sense for self-healing or when dead
+    if (_target == _caller || {!alive _target}) exitWith {};
+
+    // Tracking who is responding lets us show notifications in future too
+    private _responders = _target getVariable [QGVAR(responders), []];
+    _responders pushBackUnique _caller;
+    _target setVariable [QGVAR(responders), _responders];
+
+    // Player only cares if they are the patient
+    if (_target == ACE_player) then {
```
thojkooi: One further suggestion;

let's move the parts from l38 through l46, and the related logic in the other event handlers into the feedback component by raising a local event in here, and listing for that event in the feedback component.

I realise that we are shuffling around a bit of logic here,  but this will make the `responders` functionality introduced generic and makes the notifications bit part of a component that is more suitable for such functionality.

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,92 @@ if (isServer) then {
...
+    if (_target == _caller || {!alive _target}) exitWith {};
+
+    // Update the responder tracking
+    private _responders = _target getVariable [QGVAR(responders), []];
+    _responders deleteAt (_responders find _caller);
+    _responders = _responders - [objNull];
+    _target setVariable [QGVAR(responders), _responders];
+
+    // If someone else is still treating it's useful to know
+    if (_target == ACE_player) then {
```
thojkooi: That means we have to move this towards that component as well.

addons/medical_treatment/XEH_postInit.sqf
SilentSpike: I kind of hate that it makes it more complicated for such an edge case, but I can't think of a better way to handle that scenario

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,92 @@ if (isServer) then {
...
+    };
+};
+[QGVAR(success), _treatmentEnded] call CBA_fnc_addEventHandler;
+[QGVAR(failure), _treatmentEnded] call CBA_fnc_addEventHandler;
+
+// A dead patient no longer needs a notification
+[QEGVAR(medical,death), {
+    params ["_unit"];
+    _unit setVariable [QGVAR(responders), nil];
+    if (_unit == ACE_player) then {
```
thojkooi: And introduce the `death` event handler with these 3 lines as well, except the `_unit setVariable [QGVAR(responders), nil];` bit of course.

addons/medical_treatment/XEH_postInit.sqf
SilentSpike: 🤔 Actually we could store the value of `EFUNC(common,isPlayer)` and then check if it's still true in future

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,92 @@ if (isServer) then {
...
+        QGVAR(notification) cutFadeOut 1;
+    };
+}] call CBA_fnc_addEventHandler;
+
+// Need to update treatment notification when player switches unit
+["unit", {
```
thojkooi: And all the logic in here as well, except the bit on line 100, I suppose.

addons/medical_treatment/XEH_postInit.sqf
SilentSpike: Actually actually, it looks like `medical_ai` is written such that AI never treat player controlled characters, so we could just check if the unit is still player controlled

thojkooi: Perfect. If we need to have this work for AI as well, we can always implement a timeout based system around this.

SilentSpike:
I'm noticing some postInit files are quite large and handling initialisation of lots of different systems. I wonder if we should start splitting those up logically (either via `#include` or using block comment sub headers) so that it's easier to find what code handles what.

SilentSpike:
This is still semi-broken since `alive` and self-treatment checks were moved to feedback

SilentSpike:
The logic in this should now be correct, except I think the variable won't be reset on respawn which will produce an erroneous notification.

SilentSpike:
I think we could check if `oldUnit` is dead in the CBA unit playerEventHandler which should be a respawn event and reset the variable then. Not sure if that would ever be true in other cases though.

PabstMirror:
proposal, 3 way settings
- disabled
- simple cam shake (ie, same as opening backpacks)
- full hints

jonpas:
Closing pull request as it's out of date. Simple settings should be the way to implement this as per the proposal above or even more simplified.

PR #6538, opened 2018-08-17
Make state change status updates consistent

### LINKED ISSUE #6531: Move status variable updates out of statemachine
Some unit status variables are being updated directly in statemachine transition functions (exiting cardiac arrest resets heart rate, entering death sets it to 0), we should move these into event handlers in either the status or vitals component (whichever is more appropriate).

SilentSpike:
State transitions now call their respective status update functions in the status component where unit variables and status are actually handled.

We don't actually need to use events here since the statemachine component depends on the status component. For public consumption we already raise an `ace_unconscious` and `ace_cardiacArrest` event within the status update functions (the latter I changed to make consistent with existing unconscious event so it fires for both entry and exit).

This PR also fixes an additional behaviour where on death a unit would become bloodied even if they weren't before (untested, but going by the `setStructuralDamage` function header that's what it's for).

I've also taken the chance to rename the `setUnconsciousStatemachine` function to simply `setUnconscious` since it resides in the status component and thus has the prefix to differentiate it from the public `ace_medical_fnc_setUnconscious`.

Closes #6531


addons/medical_statemachine/functions/fnc_leftStateCardiacArrest.sqf
```diff
@@ -1,7 +1,8 @@
 #include "script_component.hpp"
 /*
- * Author: BaerMitUmlaut
- * Handles a unit entering cardiac arrest.
+ * Author: RedBery
```
TheMagnetar: Author: BaerMitUmlaut, RedBery

SilentSpike: Baer's name was only here because the enteredStateCardiacArrest file was copied over and edited so the header was for that function.

TheMagnetar: ok. I got confused :)

PR #6552, opened 2018-08-27
RHS GREF Compat - MG42

### LINKED ISSUE #6550: Spare Barrel Compatibility with RHS:GREF MG42
**Arma 3 Version:** `1.84` stable
**CBA Version:** `3.8.0` stable
**ACE3 Version:** `3.12.3` stable

**Mods:**
```
- CBA_A3
- ace
- RHS: GREF
```

**Description:**
- The MG42 lacks ACE spare barrel compatibility. The MG42 heats up very fast and switching out barrels is a necessity. Please add this feature.

**Steps to reproduce:**
- N/A

- Virtual Arsenal/ Ace Arsenal

**Placed Modules:**
- none

**RPT log file:**
- N/A

PabstMirror:
Close #6550

### DIFF
--- a/optionals/compat_rhs_gref3/CfgMagazines.hpp
+++ b/optionals/compat_rhs_gref3/CfgMagazines.hpp
@@ -1,0 +1,9 @@
+class CfgMagazines {
+    class CA_Magazine;
+    class rhsgref_50Rnd_792x57_SmE_drum: CA_Magazine { // drum, but still just a belt
+        ACE_isBelt = 1;
+    };
+    class rhsgref_296Rnd_792x57_SmE_belt: CA_Magazine {
+        ACE_isBelt = 1;
+    };
+};
--- a/optionals/compat_rhs_gref3/CfgWeapons.hpp
+++ b/optionals/compat_rhs_gref3/CfgWeapons.hpp
@@ -77,4 +77,7 @@ class CfgWeapons {
         ACE_barrelTwist=240;
         ACE_barrelLength=658;
     };
+    class rhs_weap_mg42_base: Rifle_Base_F {
+        ACE_Overheating_allowSwapBarrel = 1;
+    };
 };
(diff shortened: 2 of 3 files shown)


optionals/compat_rhs_gref3/CfgWeapons.hpp
```diff
@@ -77,4 +77,7 @@ class CfgWeapons {
         ACE_barrelTwist=240;
         ACE_barrelLength=658;
     };
+    class rhs_weap_mg42_base: Rifle_Base_F {
+        ACE_Overheating_allowSwapBarrel = 1;
```
mharis001: `EGVAR(overheating,allowSwapBarrel) = 1;`?

optionals/compat_rhs_gref3/CfgMagazines.hpp
```diff
@@ -0,0 +1,9 @@
+class CfgMagazines {
+    class CA_Magazine;
+    class rhsgref_50Rnd_792x57_SmE_drum: CA_Magazine { // drum, but still just a belt
+        ACE_isBelt = 1;
```
PabstMirror: On 2nd thought, 
magazine repack should be ok with these "assault drums"
but linking ammo via assistant gunner probably doesn't make sense???

Drofseh: Linking ammo makes as much sense as any other belt with a starter tab.
The assistant just needs to remove the tab from the to be linked belt and then use that round to link to the end of the old belt.

The belt being inside of the drum might make it a bit of a headache, but ACE allows belt linking for other belts that feed from boxes.

PabstMirror: As far as I can tell the drum is attached to the weapon by a clip on the drum's door
so opening the door would either be impossible or cause the magazine to fall free

You would also be working on the bullets and links from the rear

But I'm not really sure, I'll just merge as is for now and we can adjust if someone who knows more about the weapon has some thoughts.

Drofseh: The drum is removed every time the weapon is reloaded.

The assistant could just detach the drum, the remaining belt would slip out of the drum, and another belt would be linked to it and its drum attached.

PR #6558, opened 2018-09-02
Arsenal - Add Default Loadouts API

654wak654:
- Add a `addDefaultLoadout` function to arsenal so addon and mission makers can add default loadouts through script easier.
- Change other loadout related arsenal functions to use `findIf` like the new function.


addons/arsenal/XEH_preInit.sqf
```diff
@@ -8,6 +8,7 @@ PREP_RECOMPILE_START;
 PREP_RECOMPILE_END;
 
 // Arsenal
+GVAR(defaultLoadoutsList) = [];
```
alganthe: You're breaking default loadouts, this variable already exists whatever happens due to https://github.com/acemod/ACE3/blob/master/addons/arsenal/Cfg3DEN.hpp#L11

addons/arsenal/functions/fnc_addDeafultLoadout.sqf
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Example:
+ * ["Squad Leader", getUnitLoadout sql1] call ace_arsenal_fnc_addDefaultLoadout
+ *
+ * Public: Yes
+*/
+
+params [["_name", "", [""]], ["_loadout", [], [[]], 10]];
+
+GVAR(defaultLoadoutsList) append [_name, _loadout];
```
dedmen: Maybe some check to not add same name twice? Not sure if that matters.

addons/arsenal/functions/fnc_buttonExport.sqf
```diff
@@ -16,7 +16,7 @@ params ["_display"];
 
 if (GVAR(shiftState)) then {
 
-    if (isNil QGVAR(defaultLoadoutsList) || {GVAR(defaultLoadoutsList) isEqualTo []}) exitWith {
```
dedmen: This is unscheduled anyway so nil shouldn't matter right?

addons/arsenal/functions/fnc_addDeafultLoadout.sqf
alganthe: true, completely forgot about that too, you definitely need name checking in there.

addons/arsenal/functions/fnc_onArsenalOpen.sqf
```diff
@@ -34,12 +34,8 @@ if (isNil {GVAR(sharedLoadoutsNamespace) getVariable QGVAR(sharedLoadoutsVars)})
...
 };
 
-if (isNil QGVAR(defaultLoadoutsList)) then {
-    if (is3DEN) then {
-        GVAR(defaultLoadoutsList) = (QGVAR(DummyCategory) get3DENMissionAttribute QGVAR(DefaultLoadoutsListAttribute));
-    } else {
-            GVAR(defaultLoadoutsList) = [];
-    };
+if (is3DEN) then {
+    GVAR(defaultLoadoutsList) append (QGVAR(DummyCategory) get3DENMissionAttribute QGVAR(DefaultLoadoutsListAttribute));
```
dedmen: parenthesis for readability.

addons/arsenal/functions/fnc_buttonExport.sqf
alganthe: it can't be nil, so no worries. (my mistake)

dedmen:
<https://github.com/acemod/ACE3/pull/6558#discussion_r214539698>
No it's not. preInit runs before the 3DEN code. Meaning either the 3DEN code doesn't run and the list is empty. or it runs and overwrites the old value.

addons/arsenal/functions/fnc_addDeafultLoadout.sqf
alganthe: I can't identify loadouts by magic, I had to use something (plus it makes sense)

bux:
lacks documentation :)

addons/arsenal/Cfg3DEN.hpp
```diff
@@ -8,7 +8,7 @@ class Cfg3DEN {
                         class GVAR(DefaultLoadoutsListAttribute) {
                             property = QGVAR(DefaultLoadoutsListAttribute);
                             value = 0;
-                            expression = "if !(is3DEN) then {ace_arsenal_defaultLoadoutsList = _value};";
```
654wak654: @alganthe Was there a reason this was like this? Could I break something by using QUOTE?

commy2:
```cpp
expression = QUOTE(if !(is3DEN) then {GVAR(defaultLoadoutsList) = _value});
```
works, and so does
```cpp
expression = if !(is3DEN) then {GVAR(defaultLoadoutsList) = _value};
```

654wak654: I'll keep the verbose one with `QUOTE`, but good point. Might abuse it later.

jonpas: Second one just gets auto-quoted, and will throw missing quotations warning with armake (unless that warning is disabled).

addons/arsenal/functions/fnc_buttonLoadoutsSave.sqf
```diff
@@ -155,7 +155,7 @@ switch (GVAR(currentLoadoutsTab)) do {
 
         if (is3DEN) then {
 
-            private _sameNameLoadoutsList = _data select {_x select 0 == _editBoxContent};
+            private _loadoutIndex = (+(_data)) findId {(_x select 0) == _editBoxContent};
```
Dystopian: `findId` -> `findIf` ?

654wak654: RIP

dedmen: Why `(+(_data))` ? Why copy the array. Don't see any need.

654wak654: @alganthe had me deep copy everything

dedmen: *slaps @alganthe*

alganthe: hey no, I didn't told you to deep copy EVERYTHING, just the namespace var that is used to store loadouts (depends on which tab is active).

dedmen:
Just tried to use this and wondered why it doesn't work. Turns out, it's not released yet :U I need this tho

PR #6559, opened 2018-09-02
Movement pace configure

Dobakjang:
Lower gun and change moving pace, one of running pace is too slow. would like to make it same as not slow but there are no guidelines to alternate it or take alternate addon. let us choose the option.


Neviothr:
> Files changed [5,231]
Uhhhhh....

PabstMirror:
Post Feature Requests in #3594

I think you're asking for a setting so that non-lowered gun has less of an effect on stamina drain.
Should actually be fairly easy to implement

Dobakjang:
sorry for my lack of explaining. here is the thing
you lower youre gun
start runing (not sprinting with shift key)
and press c (which is change stance)

there are two pace and one of em is too slow. in vanilla they are both same speed. would like to let us choice to make this moving speed to slow or fast

PR #6564, opened 2018-09-05
Add the ability to edit user placed markers

Timi007:
- Add the ability to edit user placed markers by double clicking at a marker
- Remove the direct channel as a valid channel because selecting it ingame will not select the default (group) channel but instead the global
- Optimize the method to get the current map display
- To some extend also allow to copy markers by e.g. starting to edit a marker, canceling it and then creating a new marker with the same properties as the edited marker

### DIFF
--- a/addons/markers/functions/fnc_getEnabledChannels.sqf
+++ b/addons/markers/functions/fnc_getEnabledChannels.sqf
@@ -1,5 +1,5 @@
 /*
- * Author: commy2
+ * Author: commy2, Timi007
  * Return enabled channels.
  *
  * Arguments:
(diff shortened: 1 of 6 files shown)


jonpas:
I like this! Editing markers!

commy2:
Nice, this is why I love open source.

addons/markers/functions/fnc_getEnabledChannels.sqf
```diff
@@ -40,12 +40,8 @@ if (_localize) then {
...
-
-    if (setCurrentChannel 5) then {
-        _enabledChannels pushBack localize "str_channel_direct";
-    };
 } else {
-    for "_i" from 0 to 5 do {
+    for "_i" from 0 to 4 do {
         if (setCurrentChannel _i) then {
             _enabledChannels pushBack _i;
         };
```
commy2: If I understand correctly, this means that using this getter while in direct channel, the channel gets changed to global (or some other one), which would violate the no side effects principle of a getter.

Timi007:
@bux Corrected all spelling mistakes (at least those which I could find)

Issue #5408, opened 2017-08-06
Map illumination breaks after restarting mission


yarhooty:
**Arma 3 Version:** `1.72` (stable / rc / dev)
**CBA Version:** `3.4.0` (stable / dev + commit hash)
**ACE3 Version:** `3.10.2` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- ace
- RHS
- CUP
```

**Description:**
Upon restarting a single-player mission, or even loading a save, the map illumination feature does not seem to work at all. There is no dimming at night, and flashlights+map tools cannot be used. 

**Steps to reproduce:**
1. Play a single-player mission, such as Dynamic Recon Ops, at night. 
2. Verify the map is dimmed.
3. Restart the mission or load a previous save.
4. Observe that the map is no longer dimmed. 

**Where did the issue occur?**
- Single-player missions

yarhooty:
Update: After removing NIArms Core (which I forgot to include) the issue MAY have been resolved

yarhooty:
This issue continues to occur in Dynamic Recon Ops

commy2:
To make sure, does this happen with only CBA and ACE on an unscripted mission?
If not, then what is the minimum requirement of mods and / or what mission does this occur?

Xon2:
So this is happening on my side for the longest time also. I just tested this again with only CBA3 and ACE (both steam versions). If you start a mission, make a savegame and load that savegame from within the mission (without quitting to arma's main menu or quitting arma 3 alltogether, map illumination still works. But as soon as you quit the mission and resume it from arma's scenario menu, map illumination stops working. Its a really nice and immersive feature and it would be great to have this working even when finishing a mission the next day or whatever.

PR #6566, opened 2018-09-06
Map - Fix effects breaking mid mission

commy2:
- title

The map effects are only applied once, but if the map display is destroyed mid mission (loaded savegame, respawn etc.) the effects are never added again.


addons/map/functions/fnc_initMainMap.sqf
```diff
@@ -0,0 +1,44 @@
+#include "script_component.hpp"
+
+params ["_display"];
+private _control = _display displayCtrl 51;
```
Dystopian: magic number

addons/map/config.cpp
```diff
@@ -89,6 +89,11 @@ class RscDisplayMainMap {
             onDraw = QUOTE([ctrlParent (_this select 0)] call DFUNC(onDrawMap));
             #include "MapTweaks.hpp"
         };
+        class InitScript: RscText {
+            onLoad = QUOTE(ctrlParent (_this select 0) call (uiNamespace getVariable 'FUNC(initMainMap)'));
```
mharis001: Use `Extended_DisplayLoad_EventHandlers`, AFAIK `RscDisplayMainMap` uses `BIS_fnc_initDisplay`

addons/map/functions/fnc_initMainMap.sqf
```diff
@@ -0,0 +1,45 @@
...
+}];
+_control ctrlAddEventHandler ["MouseButtonUp", {
+    if ((_this select 1) == 1) then {
+        GVAR(rightMouseButtonLastPos) = [];
+    };
+}];
+
+//get mouse position on map
+_control ctrlAddEventHandler ["MouseMoving", {
+    GVAR(mousePos) = (_this select 0) ctrlMapScreenToWorld [_this select 1, _this select 2];
```
mharis001: Rather than 3x `select`, use `params`?

addons/map/functions/fnc_initMainMap.sqf
```diff
@@ -0,0 +1,45 @@
...
+}];
+_control ctrlAddEventHandler ["MouseHolding", {
+    GVAR(mousePos) = (_this select 0) ctrlMapScreenToWorld [_this select 1, _this select 2];
```
mharis001: Same as above

addons/map/functions/fnc_initMainMap.sqf
commy2: True. The orig didn't use it either, but could be updated.

addons/map/config.cpp
commy2: True. It's called `RscDiary` there however, so I should filter all idd != 12 (`IDD_MAIN_MAP`) though.

commy2:
- close #5408

addons/map/functions/fnc_initMainMap.sqf
```diff
@@ -0,0 +1,54 @@
...
+//map sizes are multiples of 1280
+GVAR(worldSize) = worldSize / 1280;
+GVAR(mousePos) = [0.5, 0.5];
+
+//Allow panning the lastStillPosition while mapShake is active
+GVAR(rightMouseButtonLastPos) = [];
+
+_control ctrlAddEventHandler ["Draw", {_this call FUNC(updateMapEffects)}];
+_control ctrlAddEventHandler ["MouseMoving", {
+    params ["_control", "_x", "_y"];
```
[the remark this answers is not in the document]
commy2: Don't think that's worth it, tbh.

addons/map/CfgEventHandlers.hpp
```diff
@@ -17,3 +17,9 @@ class Extended_PostInit_EventHandlers {
         serverInit = QUOTE(call COMPILE_FILE(XEH_postInitServer));
     };
 };
+
+class Extended_DisplayLoad_EventHandlers {
+    class RscDiary {
+        GVAR(initMainMap) = QUOTE(ctrlParent (_this select 0) call (uiNamespace getVariable 'FUNC(initMainMap)'));
```
PabstMirror:
```
 Error in expression <ctrlParent (_this select 0) call (uiName>
 Error position: <ctrlParent (_this select 0) call (uiName>
 Error ctrlparent: Type Display (dialog), expected Control
```

addons/map/XEH_postInitClient.sqf
```diff
@@ -8,52 +8,6 @@ LOG(MSG_INIT);
 // Calculate the maximum zoom allowed for this map
 call FUNC(determineZoom);
 
-[{
-    if (isNull findDisplay 12) exitWith {};
```
PabstMirror: I think we still want this line or the effects get added to the briefing maps (maps shown before mission starts)
e.g. night mission, before there were no lighting effects on map
I think to filter them out we need to do something like
```
[{
    params ["_display"];
    private _control = _display displayCtrl IDC_MAP;

    if (_display != findDisplay 12) exitWith {};
    ...
}, _this] call CBA_fnc_execNextFrame; // need to delay a frame or findDisplay doesn't work
````

Comment (commy2) [reply addons/map/XEH_postInitClient.sqf]:
```sqf
if (ctrlIDD _display != IDC_MAP) exitWith {};
```
?

PabstMirror: little easier with that command :laughing:, `IDD_MAIN_MAP`

commy2: Done.

PR #6569, opened 2018-09-09
Zeus - Group Side Module: preserve group id and assigned team

### LINKED ISSUE #6568: Group Side Module changes group attributes
**ACE3 Version:** `3.12.3` stable

**Description:**
- Using the Group Side module changes the group's name and the assigned team of each member is reset.

**Steps to reproduce:**
- Be in a group with a custom callsign
- Set your team color
- Have Zeus use the Group Side module
- Callsign has been changed to a vanilla callsign and everyone is on the main team.

I am willing to write the code preserving these details for groups with human players, if the ACE team thinks that will be a good addition and would merge the PR.

I know it would be for my group, as our modded radios are linked to group names and team colors. You'll go from being on Block 2 Channel 4 as Bravo-3 (Blue Team) to being Block 16 Channel 1 as Alpha 1-1 (White).

synixebrett:
- Close #6568 
- Preserve the group id when switching the group's side if it is not taken
- Preserve the assigned team color of units in the group being switched

**Notes:**
If the `Move units from group on unconscious` setting is enabled and a unit is unconscious when the module is used they will join the new group when they are woken up as expected. They will not keep their assigned team however and will join the white team when woken up. That would require changing code related to the `moveUnitsFromGroupOnUnconscious` setting and I don't think it is worth it for this edge case.

### DIFF
--- a/addons/zeus/functions/fnc_moduleGroupSide.sqf
+++ b/addons/zeus/functions/fnc_moduleGroupSide.sqf
@@ -1,5 +1,5 @@
 /*
- * Author: SilentSpike
+ * Author: SilentSpike, Brett
  * Zeus module function to change side of a group on dialog confirmation
  *
  * Arguments:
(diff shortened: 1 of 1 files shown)


addons/zeus/functions/fnc_moduleGroupSide.sqf
```diff
@@ -25,20 +25,36 @@ if (_side == _newSide) exitWith {};
 private _oldGroup = group _unit;
 private _newGroup = createGroup _newSide;
 
+// Preserve groupid from the previous group if doesn't already exist
```
synixebrett: If there is a better way to do this please let me know so I can update it

mharis001: Could use `findIf`

synixebrett: From a quick test it looks like this method is faster, and in my opinion more readable than `findIf`.
I'll do a few more tests though since it would make more sense for `findIf` to be faster.

addons/zeus/functions/fnc_moduleGroupSide.sqf
```diff
@@ -25,20 +25,30 @@ if (_side == _newSide) exitWith {};
...
+if ((allGroups findIf {side _x isEqualTo _newSide && {(groupId _oldGroup) isEqualTo (groupId _newGroup)}}) == -1) then {
+    _newGroup setGroupIdGlobal [groupId _oldGroup];
+};
```
commy2: Couldn't you just use `setGroupIdGlobal` without check, and if it is already taken, it would just fail silently anyway?

synixebrett: I just thought it would be better to avoid potentially undefined behavior, but I guess I can test it to make sure nothing odd happens and then get rid of the check.

commy2: I wonder if these even have to be unique. I would do the check if:
- the game bugs out with duplicated ids
- the game complains about it in RPT, but still changes the id
I would maybe do the check if, depending on the time of day if:
- the game complains about it in RPT, but doesn't change the id

synixebrett: I'm not exactly sure what to do here. Here is what I discovered:
- You can have multiple groups with the same ID if they are on different sides
- You can not have multiple groups with the same ID if they are on the same side
- If you try to set a group id to something already taken on that side, they switch IDs

This means the following is possible:
 - BLUFOR has a group "Alpha-1"
 - OPFOR has a group "Alpha-1"
 - The Group Side module is used on the BLUFOR group and set to OPFOR
 - The former BLUFOR group is now an OPFOR group with an empty id
 - The former BLUFOR group takes "Alpha-1"
 - The original OPFOR group now has an empty id

Personally I'm not a fan of that scenario, since it is modifying a different group. I think the check is worth keeping, especially since it is relatively quick and infrequently used. I'm also thinking I'll add a new module to set a group ID from Zeus. I know this stuff is kinda niche but I think it'll be useful enough to add.

commy2: So setting to an already used id (on the same side) removes the id from the group that already had it? I'd say it's worth to keep the check in that case. The new group should be empty, not the old one that wasn't even touched by the module.
Alternatively some way to add (2) etc. to the id, though idk if you think that is worth the trouble.

synixebrett: I don't think adding (2) is worth it, either you don't care and will leave it empty, or you do care and are going to change it anyway.

PR #6571, opened 2018-09-10
don't hide stamina twice when changing the setting

commy2:
- title


commy2:
Changed what this PR is about. The deleted code is already written here: https://github.com/acemod/ACE3/blob/master/addons/advanced_fatigue/initSettings.sqf#L17-L21

PR #6572, opened 2018-09-13
Arsenal - Filter firemodes for precision display by showToPlayer

Blutze:
- Change the precision display in the ACE Arsenal to ignore AI-only firemodes

There is a number of weapons, such as the NIArms M14DMR, RHS M249s and NIArms SiG 553s, that have an AI-only firemode with a lower `dispersion` value than any mode available to players. The current implementation shows those values on the Arsenal interface, painting a rather misleading picture.

I have considered checking for existence of the `showToPlayer` property first, but it is already defined on the base classes for weapons (`Default`) and firemodes (`Mode_SemiAuto`).

### DIFF
--- a/addons/arsenal/functions/fnc_statTextStatement_accuracy.sqf
+++ b/addons/arsenal/functions/fnc_statTextStatement_accuracy.sqf
@@ -23,7 +23,9 @@ private _fireModes = getArray (_config >> "modes");
 private _dispersion = [];
 
 {
-    _dispersion pushBackUnique (getNumber (_config >> _x >> "dispersion"));
+    if (getNumber (_config >> _x >> "showToPlayer") != 0) then {
+        _dispersion pushBackUnique (getNumber (_config >> _x >> "dispersion"));
+    };
 } foreach _fireModes;
 
 _dispersion sort true;


commy2:
Sounds like the weapon's config is broken if for some reason AI has different dispersion than players.

dedmen:
>Sounds like the weapon's config is broken if for some reason AI has different dispersion than players.

You don't want AI insta headshotting you generally.

Blutze:
> Sounds like the weapon's config is broken if for some reason AI has different dispersion than players.

In most cases, it is just caused by not adjusting all firemodes when you have your own gun inherit from `Rifle_Base_F` or every time you create a shortbarreled version. Usually though, the offending firemode is "long range, with optics" or something of the sort, and the discrepancy might actually be intentional.

commy2:
>In most cases, it is just caused by not adjusting all firemodes when you have your own gun inherit from Rifle_Base_F or every time you create a shortbarreled version.

Like I said, broken then. Doesn't mean I am against this PR / fix btw.

>Usually though, the offending firemode is "long range, with optics" or something of the sort, and the discrepancy might actually be intentional.

Sounds fine if it is intentional, although you could just change aiDispersionX and aiDispersionY instead for those (ai) modes. Asset makers (artists) are bad at logic I guess, or I am just full of myself and there is something I am missing.

alganthe:
> The current implementation shows those values on the Arsenal interface, painting a rather misleading picture.

Which is false, I fixed this months ago by changing what firemodes are used.
https://github.com/acemod/ACE3/blob/master/addons/arsenal/functions/fnc_statTextStatement_accuracy.sqf#L27

Only non-AI firemodes should be listed in there.

commy2:
@alganthe, you're wrong about that `modes[] = {"manual","close","short","medium","far"};` from `LMG_RCWS`.

commy2:
>Can you even remove a subclass after inheriting it?

No.

PR #6573, opened 2018-09-14
Add Feature Camera Player EH and Handling

jonpas:
- Replace ACE3's Feature Camera handling with CBA's new Player EH
- Remove `"ace_activeCameraChanged"` event.
- Register Spectator and HuntIR Cameras with CBA

**Requires CBA 3.9.0 (unreleased)!** (possibly higher) - https://github.com/CBATeam/CBA_A3/pull/982


addons/nightvision/functions/fnc_pfeh.sqf
```diff
@@ -22,9 +22,9 @@ if ((currentVisionMode ACE_player) != 1) exitWith {
     [GVAR(PFID)] call CBA_fnc_removePerFrameHandler;
     GVAR(PFID) = -1;
 };
-if (EGVAR(common,OldIsCamera)) exitWith {
+if (call CBA_fnc_getActiveFeatureCamera != "") exitWith {
```
PabstMirror: This runs every frame, I'd rather just use the var `cba_events_OldIsCamera`

PR #6575, opened 2018-09-15
RHS Compat - Fix GREF canoes refuel interaction

Wakbub:
- Fix RHS GREF canoes so that they cannot receive fuel
- Add authors

### DIFF
--- a/optionals/compat_rhs_gref3/CfgVehicles.hpp
+++ b/optionals/compat_rhs_gref3/CfgVehicles.hpp
@@ -1,0 +1,12 @@
+class CfgVehicles {
+    class All;
+    class AllVehicles: All {};
+    class Ship: AllVehicles {};
+    class Ship_F: Ship {};
+    class Boat_F: Ship_F {};
+    class Rubber_duck_base_F: Boat_F {};
+    class rhsgref_canoe_base: Rubber_duck_base_F {
+        // Canoes are propelled by paddlers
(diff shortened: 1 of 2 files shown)


optionals/compat_rhs_gref3/config.cpp
```diff
@@ -6,8 +6,9 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"rhsgref_main", "rhsgref_c_weapons"};
+        requiredAddons[] = {"ace_refuel", "rhsgref_main", "rhsgref_c_weapons"};
```
PabstMirror: don't want `ace_refuel` here as we want to keep it optional/modular

optionals/compat_rhs_gref3/CfgVehicles.hpp
```diff
@@ -0,0 +1,12 @@
+class CfgVehicles {
+    class All;
+    class AllVehicles: All {};
+    class Ship: AllVehicles {};
+    class Ship_F: Ship {};
+    class Boat_F: Ship_F {};
+    class Rubber_duck_base_F: Boat_F {};
+    class rhsgref_canoe_base: Rubber_duck_base_F {
```
mharis001: Just need the parent class of this.

PR #6576, opened 2018-09-16
Use STRING macros

Wakbub:
- Use STRING macros
- Fix minor documentation typos related to STRING macros

Related, there is a problem with the naming of the strings in the `concertina_wire` module. They don't include the module name, so the string macros cannot be used.
```
"STR_ACE_CONCERTINA_WIRE"
"STR_ACE_CONCERTINA_WIRECOIL"
"STR_ACE_UNROLLWIRE"
"STR_ACE_ROLLWIRE"
```
If changing these strings, any mods using the current strings will be incompatible. But maybe change this in a bigger update for coherency?


commy2:
You can also write `LLSTRING(var)` for `localize LSTRING(var)`

Wakbub:
Great. In that case the [Coding Guidelines](https://ace3mod.com/wiki/development/coding-guidelines.html) can be updated to reflect that.

SilentSpike:
I don't think we've ever worried about backwards compatibility for our stringtables, other mods could always copy them out if they want.

Wakbub:
Maybe that will do.

docs/wiki/development/coding-guidelines.md
```diff
@@ -126,22 +126,23 @@ These macros are allowed but are not enforced.
...
 
-Script strings (still require `localize` to localize the string):
+Script strings:
 
 | Macro | Expands to |
 | -------|---------|
-|`LSTRING(banana)` | `"STR_ACE_balls_banana"` |
-|`ELSTRING(leg,banana)` | `"STR_ACE_leg_banana"` |
+|`LLSTRING(banana)` | `localize "STR_ACE_balls_banana"` |
+|`LELSTRING(leg,banana)` | `localize "STR_ACE_leg_banana"` |
```
commy2: Should list all four imo.

Wakbub:
Couldn't find any place where the string itself (e.g., "STR_ACE_balls_banana") has been of interest, but there we go.

commy2:
We don't universally use those as opposed to LLSTRING etc, and I personally think they offer no benefit in i.e. readability, and only obscure the code.

commy2:
By that I mean that I wouldn't advertise for that family of macros, since I wouldn't use them myself for various reasons. There is no reason why the wiki should be a complete list of all macros that have been conceived of in the past that didn't establish themselves, because they offered nothing valuable enough for the price of being a macro.

Wakbub:
Certainly. I never said complete however, just asked if the goal is for it to be more comprehensive. It can be tricky to know when the coding guidelines state one way to write, the first comment being there is another way to write, and then it is suggested that both ways should be listed. If both ways should be listed, it should be ok to use either way? Maybe your comment was just trivia rather than a suggestion.

As no one in ACE3 has ever used those macros mentioned though and you strongly advise against them, maybe they should not be defined? It would make sense to only have macro definitions that the core people use and support. ACE3 and CBA are as far as I can tell the top places to learn what beautiful SQF code looks like, and anything to make it easier to learn is welcome.

commy2:
That's just because LLSTRING is relatively recent, newer than the coding guidelines. I was for keeping LSTRING as well as LLSTRING, because sometimes you don't want to immediately translate strings via `localize`, for example when doing MP stuff where the string may be translated differently on multiple machines.

commy2:
>maybe they should not be defined?
They have to be kept for bwc, and are probably used somewhere in legacy code. Or maybe someone else finds them useful and disagrees with me on wheter they should be used or not.

Wakbub:
Ok, in case both macros serving a purpose and there's concerns for backwards compatibility all is good.

PR #6582, opened 2018-09-18
Tagging - Cache required items check and use uniqueItems

mharis001:
- title

### DIFF
--- a/addons/tagging/functions/fnc_checkTaggable.sqf
+++ b/addons/tagging/functions/fnc_checkTaggable.sqf
@@ -20,7 +20,7 @@
 
     // Exit if no required item in inventory
     if ([_unit, {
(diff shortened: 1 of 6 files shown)


addons/tagging/functions/fnc_checkTaggable.sqf
```diff
@@ -36,7 +38,6 @@
     // Exit if trying to tag a non static object
     TRACE_1("Obj:",_intersections);
 
-    // Exit if trying to tag a non static object
```
dedmen: Why remove?

mharis001: Duplicate comment as above

dedmen: 🤦‍♂️ oof

PabstMirror:
Added a commit that ensures `_requiredItem` is always config-case, should let us drop all the `apply {toLower _x}`

PR #6585, opened 2018-09-19
UI - Fix selective UI in cargo

### LINKED ISSUE #6547: Disabling Ammo Count interface - not affect player inside vehicles, if player uses - "this moveInCargo"
**Arma 3 Version:** `1.84` (stable)
**CBA Version:** `3.8.0.180801` (stable )
**ACE3 Version:**  `3.12.3` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
-In the settings - OPTIONS/ADDON OPTIONS/ACE User interface/Ammo Count I turn off the display of the number of current ammunition for infantryman, but if I use for player - "this moveInCargo", then inside any vehicle,  it does not work. In this case I again see the number of ammo in the my rifle. 
This is not a big mistake, but a lot of old missions are contain the function "this moveInCargo" and this creates confusion in the player's interface
Please correct this if possible.

**Steps to reproduce:**
- a) Place in the editor civilian vehicle Van (transport) and call it in the init as "Car1"
  b) Place in the editor squad with player and place in init of every soldier  -this moveInCargo Car1;
After starting the game you will see, that inside this vehicle  the player again has an Ammo Count interface.

-Editor

**Placed Modules:**
- none

**RPT log file:**
- none

Dystopian:
- fix #6547;
- improve `ace_ui_fnc_setAdvancedElement`:
  - fix doc;
  - add some debug;
  - fix magic numbers;
  - fix bad variable name (`show` -> `fade`).

Maybe now `VEHICLE_ONLY` and `GROUND_ONLY` macros should be renamed e.g. to `VEHICLE_NOT_CARGO` and `GROUND_AND_CARGO`.

### DIFF
--- a/addons/ui/functions/fnc_setAdvancedElement.sqf
+++ b/addons/ui/functions/fnc_setAdvancedElement.sqf
@@ -6,7 +6,7 @@
  * Arguments:
  * 0: Element Name <STRING>
  * 1: Show/Hide Element <BOOL>
- * 2: Show Hint <BOOL>
+ * 2: Show Hint <BOOL> (default: false)
  * 3: Force change even when disallowed <BOOL> (default: false)
  *
  * Return Value:
@@ -18,12 +18,13 @@
  * Public: No
  */
 
-params ["_element", "_show", ["_showHint", false, [true]], ["_force", false, [true]] ];
+params ["_element", "_show", ["_showHint", false, [true]], ["_force", false, [true]]];
 
 private _cachedElement = GVAR(configCache) getVariable _element;
-if (isNil "_cachedElement") exitWith {};
+if (isNil "_cachedElement") exitWith {TRACE_1("nil element",_this)};
 
 if (!_force && {!GVAR(allowSelectiveUI)}) exitWith {
+    TRACE_1("not allowed",_this);
     [LSTRING(Disallowed), 2] call EFUNC(common,displayTextStructured);
     false
 };
@@ -31,8 +32,15 @@ if (!_force && {!GVAR(allowSelectiveUI)}) exitWith {
 _cachedElement params ["_idd", "_elements", "_location", "_conditions"];
 
 // Exit if main vehicle type condition not fitting
-private _canUseWeapon = ACE_player call CBA_fnc_canUseWeapon;
-if ((_canUseWeapon && {_location == 2}) || {!_canUseWeapon && {_location == 1}}) exitWith {false};
+private _cargoIndex = vehicle ACE_player getCargoIndex ACE_player; // nil if not in vehicle
+private _isOnFootOrInCargo = isNil "_cargoIndex" || {_cargoIndex > -1};
+if (
+    (_isOnFootOrInCargo && {_location == VEHICLE_ONLY})
+    || {!_isOnFootOrInCargo && {_location == GROUND_ONLY}}
+) exitWith {
+    TRACE_3("skip location",_this,_isOnFootOrInCargo,_location);
+    false
+};
 
 // Get setting from config API
 {
@@ -59,7 +67,7 @@ if (!_force) then {
     };
 };
 
-_show = [1, 0] select _show;
+private _fade = [1, 0] select _show;
 
 // Disable/Enable elements
 private _success = false;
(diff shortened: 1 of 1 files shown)


jonpas:
I don't like `fade`, it reminds me of slow fading away. Like vanilla option to fade after not being used, and show when needed again. Which is something Selective UI is missing btw.

commy2:
show and hide

addons/ui/functions/fnc_setAdvancedElement.sqf
```diff
@@ -18,21 +18,29 @@
...
     [LSTRING(Disallowed), 2] call EFUNC(common,displayTextStructured);
     false
 };
 
 _cachedElement params ["_idd", "_elements", "_location", "_conditions"];
 
 // Exit if main vehicle type condition not fitting
-private _canUseWeapon = ACE_player call CBA_fnc_canUseWeapon;
-if ((_canUseWeapon && {_location == 2}) || {!_canUseWeapon && {_location == 1}}) exitWith {false};
+private _cargoIndex = vehicle ACE_player getCargoIndex ACE_player; // nil if not in vehicle
```
dedmen: Why check cargo specifically? Instead of just in vehicle check?

Dystopian: because these controls (Soldier/Vehicle/Weapon Information, `location = GROUND_ONLY`) should be set when on foot or in vehicle cargo including FFV.

dedmen: Yeah but why cargo? Why exclude driver/gunner?

Dystopian: @jonpas should know. I think because that info is shown while player is in cargo and isn't shown when in other seats.

jonpas: Why exactly does old way break with moveInCargo?

jonpas: I see, I will test this soon.

addons/ui/functions/fnc_setAdvancedElement.sqf
```diff
@@ -18,21 +18,29 @@
...
+private _isOnFootOrInCargo = isNil "_cargoIndex" || {_cargoIndex > -1};
```
PabstMirror:
```suggestion
private _isOnFootOrInCargo = isNil "_cargoIndex" || {_cargoIndex > -1} || {ACE_player call CBA_fnc_canUseWeapon};
```
Need this to handle non-cargo FFV turrets.
Example with merkel tank with commander gun,
inside it shows ammo (gunner ammo) fine
turn out and player's gun ammo shows, and exiting while in this state it will continue to show dismounted.

While turned out in turret::
ACE_player call CBA_fnc_canUseWeapon = false
vehicle ACE_player getCargoIndex ACE_player = -1

Could possibly also use `isTurnedOut player`, but cba func isn't that expensive

Dystopian:
@PabstMirror I added non-cargo FFV check with variable renaming.

Also I'm not sure now if we should add more `location` types to config instead such `GROUND_ONLY` and FFV/cargo combination. Something like `GROUND_FFV_CARGO` etc.

PabstMirror:
I guess ground and vehicle aren't perfect, but I think the intent is clear enough
Ground = ground weapons, even if mounted
vehicle = vehicle's own weapons

addons/ui/functions/fnc_setAdvancedElement.sqf
```diff
@@ -18,21 +18,28 @@
...
-private _canUseWeapon = ACE_player call CBA_fnc_canUseWeapon;
-if ((_canUseWeapon && {_location == 2}) || {!_canUseWeapon && {_location == 1}}) exitWith {false};
+private _canUseWeaponOrInCargo = ACE_player call CBA_fnc_canUseWeapon || {-1 < vehicle ACE_player getCargoIndex ACE_player};
+if (
+    (_canUseWeaponOrInCargo && {_location == VEHICLE_ONLY})
+    || {!_canUseWeaponOrInCargo && {_location == GROUND_ONLY}}
+) exitWith {
+    TRACE_3("skip location",_this,_canUseWeaponOrInCargo,_location);
+    false
+};
```
jonpas: I'd prefer this whole thing like this:
```sqf
if ((_canUseWeaponOrInCargo && {_location == VEHICLE_ONLY}) ||
    {!_canUseWeaponOrInCargo && {_location == GROUND_ONLY}}
) exitWith {
    TRACE_3("skip location",_this,_canUseWeaponOrInCargo,_location);
    false
};
```
Probably just because I am more used to it, either is mergable.

PR #6589, opened 2018-09-19
Fixes to unconsciousness and damageTypes systems

redbery:
- Ensure units go unconscious CriticalVitals when bleeding 0.5 of max regardless of bleedingCoefficient (without this fix a bleeding bleedingCoefficient of 0.5 or less prevents units from going KO due to blood loss in any circumstances)
- Fix unit not having lastWakeUpCheck set due to incorrect variable macro
- Fix the detection for unknown damageTypes (should now fire with damageType "Unknown")
- Add collision damageType used by the damageHandler to the config (same settings as vehicleCrash)
- Changes collision wound location selection to be weighted to increase lethality on collisions with high momentum

### DIFF
--- a/addons/medical_damage/ACE_Medical_Injuries.hpp
+++ b/addons/medical_damage/ACE_Medical_Injuries.hpp
@@ -7,7 +7,7 @@ class ACE_Medical_Injuries {
         // Source: Scarle
         //  Also called scrapes, they occur when the skin is rubbed away by friction against another rough surface (e.g. rope burns and skinned knees).
         class Abrasion {
-            causes[] = {"falling", "ropeburn", "vehiclecrash", "unknown"};
+            causes[] = {"falling", "ropeburn", "vehiclecrash", "collision", "unknown"};
             bleeding = 0.001;
             pain = 0.4;
             minDamage = 0.01;
@@ -15,7 +15,7 @@ class ACE_Medical_Injuries {
         };
         // Occur when an entire structure or part of it is forcibly pulled away, such as the loss of a permanent tooth or an ear lobe. Explosions, gunshots, and animal bites may cause avulsions.
         class Avulsion {
-            causes[] = {"explosive", "vehiclecrash", "grenade", "shell", "bullet", "backblast", "bite"};
+            causes[] = {"explosive", "vehiclecrash", "collision", "grenade", "shell", "bullet", "backblast", "bite"};
             bleeding = 0.25;
             pain = 1.0;
             minDamage = 0.01;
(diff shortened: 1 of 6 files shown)


addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -114,8 +114,10 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
 
             // Significant damage suggests high relative velocity
             // Momentum transfers to body/head for worse wounding
+            // Higher momentum results in higher chance for head to be hit for more lethality
             if (_receivedDamage > 0.35) then {
-                 _woundedHitPoint = selectRandom ["Body", "Head"];
+                private _headHitWeight = (_receivedDamage / 2) max 1;
+                _woundedHitPoint = selectRandomWeighted ["Body", (1 - _headHitWeight), "Head", _headHitWeight];
```
SilentSpike: 👍 Good change

addons/medical_vitals/functions/fnc_handleUnitVitals.sqf
```diff
@@ -126,7 +126,7 @@ switch (true) do {
             [QEGVAR(medical,CriticalVitals), _unit] call CBA_fnc_localEvent;
         };
     };
-    case (_bloodLoss > BLOOD_LOSS_KNOCK_OUT_THRESHOLD * _cardiacOutput): {
+    case (_bloodLoss / EGVAR(medical,bleedingCoefficient) > BLOOD_LOSS_KNOCK_OUT_THRESHOLD * _cardiacOutput ): {
```
TheMagnetar: extra space before bracket. 

All in all nice changes!

addons/medical_damage/ACE_Medical_Injuries.hpp
```diff
@@ -115,6 +119,10 @@ class ACE_Medical_Injuries {
             thresholds[] = {{0.1, 1}};
             selectionSpecific = 1;
         };
+        //No related wounds as drowning should not cause wounds/bleeding. Can be extended for internal injuries if they are added.
```
TheMagnetar: Should this be left out until (if) airway management gets implemented?

redbery: For now it is mainly there to prevent the wounds getting fired with Unkown damage type as that will cause wounds that make no sense for drowning. This was no issue in the current branch as the detection for a non specified damageType was not working correctly (the conversion to Unkown did not fully take place). This PR fixes that so it suddenly does become an issue.

redbery:
Actually didn't realize I still had this PR open when pushing the new batch of changes. Changed the title and description to match all of the changes now.

addons/medical_vitals/functions/fnc_handleUnitVitals.sqf
redbery: Thanks! Space should be fixed now.

TheMagnetar:
Tested.

PR #6590, opened 2018-09-20
Realistic Names - Rename RGN to V40 Mini-Grenade, fix some trailing spaces

Blutze:
- Rename vanilla "RGN" grenade to "V40 Mini-Grenade"
- Add German translation for V40
- Remove trailing spaces in Czech strings for M67 grenade, M15 AT mine and M112 demo block; and in Italian string for HEMTT (Container).

**This pull request will not:**
- Add any other translation for V40
- Attempt to do any fixes on strings in non-Latin charsets.

And while we're at it: Shouldn't it be "Nebel" instead of "Rauch" for the German smokes?

### DIFF
--- a/addons/realisticnames/CfgMagazines.hpp
+++ b/addons/realisticnames/CfgMagazines.hpp
@@ -396,6 +396,10 @@ class CfgMagazines {
         displayName = CSTRING(HandGrenade_Name);
         displayNameShort = "M67";
     };
+    class MiniGrenade: CA_Magazine {
+        displayName = CSTRING(MiniGrenade_Name);
+        displayNameShort = "V40";
+    };
     class SmokeShell: HandGrenade {
         displayName = CSTRING(SmokeShell_Name);
     };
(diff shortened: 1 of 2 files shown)


bux:
Don't put the english translation into every language. Just remove the languages you're not able to translate the text to. If they have a value, they won't show up as missing translation in our overview.

Blutze:
Wasn't sure, so I rather took that risk than creating null/undefined exceptions.

bux:
Wouldn't happen. Missing translations will fall back to the english entry

commy2:
>Missing translations will fall back to the english entry

They fall back to the \<Original\> entry, and if that doesn't exist, the first entry, which is always \<English\> in ACE by design, so we don't have to write the same thing twice.

Blutze:
Anything else you want me to change/add/remove? There'd be at least one more Czech space for instance.

Blutze:
Fixed more trailing spaces (not that that has anything to do with the original goal of this PR). The ACE names for the `lsv_01_at` have them as well, but I'm soooooo not gonna touch characters I don't understand.

bux:
Thanks for czeching... (badum tsss)

PR #6594, opened 2018-09-23
WIP: uniqueItemsPlayer

PabstMirror:
Stripped down fast version that's just for ace_player
20 items, 4 unique (ammo bearer unit placed in editor)
`"ACE_CableTie" in (items player)` 0.0034 ms
`"ACE_CableTie" in (player call ace_common_fnc_uniqueItems)` 0.0056 ms
`"ACE_CableTie" in (call ace_common_fnc_uniqueItemsPlayer)` 0.0024 ms

### DIFF
--- a/addons/common/XEH_PREP.hpp
+++ b/addons/common/XEH_PREP.hpp
@@ -179,6 +179,7 @@ PREP(toNumber);
 PREP(unhideUnit);
 PREP(uniqueElements);
 PREP(uniqueItems);
+PREP(uniqueItemsPlayer);
 PREP(unloadPerson);
 PREP(unloadPersonLocal);
 PREP(unmuteUnit);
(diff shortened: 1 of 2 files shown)


dedmen:
You are only creating the cache once. Never clearing it?

PabstMirror:
Actually implementing this is going to be a little tricky, and might not be worth it due to making functions less usable.

Example (https://github.com/acemod/ACE3/blob/master/addons/captives/functions/fnc_canApplyHandcuffs.sqf#L23):
```
{"ACE_CableTie" in (_unit call EFUNC(common,uniqueItems))} &&
```
Function is private and in every use we have it's only called when `_unit` is player.
Changing to `in (call ace_common_fnc_uniqueItemsPlayer)` would break the function if it was ever used on a remote `_unit`.

So it should be safe (assuming no one else is using this function in a unintended way), but I'm not sure if the trivial performance gains are worth it.

PabstMirror:
probably not worth it, closing for now

dedmen:
> Actually implementing this is going to be a little tricky

Could add a _unit isEqualTo ACE_Player check in fnc_uniqueItems, and switch to the extra fast cached variant in that case.

PR #6596, opened 2018-09-24
Disable marker editing for everything non-icon

Timi007:
- Disable marker editing for everything non-icon

[Comment from PabstMirror](https://github.com/acemod/ACE3/pull/6564#issuecomment-423861538)

### DIFF
--- a/addons/markers/functions/fnc_initInsertMarker.sqf
+++ b/addons/markers/functions/fnc_initInsertMarker.sqf
@@ -55,7 +55,7 @@
 
     //check if entity under mouse is a user marker
     if (_mouseOverType isEqualTo "marker") then {
-        if !((_marker find "_USER_DEFINED") isEqualTo -1) then {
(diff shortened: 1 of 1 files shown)


commy2:
What else are there for shapes aside from "ICON" and "POLYLINE" (?).

Timi007:
According to the [Wiki](https://community.bistudio.com/wiki/setMarkerShape): "ICON", "RECTANGLE", "ELLIPSE" and "POLYLINE"

commy2:
So actually this also disables editing for everything non-ICON? (POLYLINE, RECTANGLE and ELLIPSE)?

jonpas:
Can you even place them in game? Only Editor IIRC

commy2:
Well, one could've easily written `!= "POLYLINE"` if one wanted to make these types still editable.
Or if one doesn't want to, one would name the PR accordingly.

commy2:
Thanks. These are the things that bother me.

PabstMirror:
Can't test now, but I think the "straight lines" feature ace still has might be type rectangle. and probably don't want to convert those to icons either

PR #6598, opened 2018-09-26
Arsenal - Fix attribute blacklist not working in multiplayer

mharis001:
- title


bux:
Are two frames really really enough? Or just enough on your machines?

mharis001:
Should be enough. `initBox` happens one frame after postInit, so `removeVirtualItems` one frame after than should work.

alganthe:
InitBox is re-ran via a global EH in MP, thus you need to run any arsenal related funcs 2 frames after it's called @bux

PR #6604, opened 2018-09-29
Add MagazineWells support

dedmen:


PabstMirror:
Could make a vog version of the huntIR (seperate magazine)

addons/ballistics/CfgMagazineWells.hpp
```diff
@@ -0,0 +1,186 @@
...
+    class CBA_65x39_MX {
+        ADDON[] = {
+            "ACE_30Rnd_65x39_caseless_mag_Tracer_Dim",
+            "ACE_30Rnd_65x39_caseless_green_mag_Tracer_Dim"
+        };
+    };
+    class CBA_65x39_MX_XL {
+        ADDON[] = {
+            "ACE_100Rnd_65x39_caseless_mag_Tracer_Dim",
+            "ACE_200Rnd_65x39_cased_Box_Tracer_Dim"
```
PabstMirror: I think this is a 200 rnd belt and probably not right for mx
was used in Stoner 99 / LMG_Mk200_F

PabstMirror:
I guess we could also cleanup some of the `magazine[]` arrays.
Although I don't understand some of them like
```
    class arifle_MXM_F: arifle_MX_Base_F {
        magazines[] = {
            "30Rnd_65x39_caseless_mag",
            "30Rnd_65x39_caseless_mag_Tracer",
            "ACE_30Rnd_65x39_caseless_mag_Tracer_Dim",
            "ACE_30Rnd_65x47_Scenar_mag",
            "ACE_30Rnd_65_Creedmor_mag"
```

addons/ballistics/CfgMagazineWells.hpp
Drofseh: yah, the 200rd is a belt for the Mk200.
I will make a PR to cba to add a magwell for it.

Drofseh: https://github.com/CBATeam/CBA_A3/pull/1007

Drofseh:
> I guess we could also cleanup some of the `magazine[]` arrays.
> Although I don't understand some of them like
> 
> ```
>     class arifle_MXM_F: arifle_MX_Base_F {
>         magazines[] = {
>             "30Rnd_65x39_caseless_mag",
>             "30Rnd_65x39_caseless_mag_Tracer",
>             "ACE_30Rnd_65x39_caseless_mag_Tracer_Dim",
>             "ACE_30Rnd_65x47_Scenar_mag",
>             "ACE_30Rnd_65_Creedmor_mag"
> ```

I don't understand that either, I don't think the MX should shoot 6.5 Creedmoor or 6.5x47 Lapua (Scenar)

addons/ballistics/CfgMagazineWells.hpp
PabstMirror: moved to that well
and it won't cause any problems to use the well config before cba is updated

addons/concertina_wire/XEH_PREP.hpp
```diff
@@ -6,3 +6,4 @@ PREP(handleDamage);
 PREP(handleInit);
 PREP(handleKilled);
 PREP(vehicleDamage);
+x
```
dedmen: Wat

Blutze:
The magazines themselves also have the mass and inventory icons, potentially even 3d models, of normal 6.5x39mm Bohemia 30 rounders. Always thought they are just there as a showcase for those particular ACE ammo types, and advanced ballistics in general.

Something I noticed missing in the CBA code: Why is the SiG 556 (`dmr_03`) not getting any magwells? Shouldn't it be AR-10/SR-25, with the vanilla 20 rounders?

Drofseh:
I'm really not sure what to do with `srifle_DMR_03_F`.
The SIG 556 is a 5.56x45 rifle, but in game it's using 7.62x51.

CBA discussion here: https://github.com/CBATeam/CBA_A3/issues/108#issuecomment-429584424

Issue #4814, opened 2017-01-11
Tactical Ladder script error


dedmen:
**Arma 3 Version:** `1.66.139.630` (profv3)
**CBA Version:** `3.8.3` (stable)
**ACE3 Version:** `3.1.2` (stable)

**Mods:**
- `@CBA_A3`
- `@ace`
- `@acex`

**Description:**
- Script error even when no Ladder anywhere near
- Probably related to #4017
- First that script shouldn't even be triggered by just right-clicking without a Ladder anywhere near
- Second `animate` should always return Nothing according to BIKI so the error should always be triggered. I don't know why it sometimes apparently returns something that count accepts as a bool.

**Steps to reproduce:**
- Open a completly empty 3DEN mission.
- Start the mission without any playable unit
- When in a camera hovering over the map press right-click.
- Error is triggered

**Where did the issue occur?**
- Editor (Singleplayer)

**RPT log file:**
too lazy
```
20:05:16 Error in expression <tate", 0];

{
ace_tacticalladder_ladder animate [_x, 0];
} count ["extract_1","e>
20:05:16   Error position: <animate [_x, 0];
} count ["extract_1","e>
20:05:16   Error Type Nothing, expected Bool
20:05:16 File z\ace\addons\tacticalladder\functions\fnc_cancelTLdeploy.sqf, line 18
```

PabstMirror:
Full RPT would be helpful.
Just like #4017, this is probably caused by a initPost failure and unrelated to this module.
Error will occur if `ace_tacticalLadder_ladder` is nil. Could you check what this variable is?

dedmen:
But even if it is unrelated.. according to BIKI animate always returns Nothing.. So it should never be a bool type.. so the error should always happen. Why count on undefined behavior returning a null type instead of a Nil type like it's supposed to.
Also I don't think that function should be called if I wasn't actually trying to deploy a ladder.
Very likely caused by MCC. Though i have never experienced that error before.
But yeah.. its correct.. According to RPT no postInit is happening.
ace_tacticalLadder_ladder is nil till the first right-click. after that it is objNull.
So yeah.. the final cause is a postInit fail. But the other points still remain.

RPT: http://pastebin.com/xawHpSXk
["ace_tacticalladder_ladder=",any] is my debug print that prints the current value of the var
There is no other script error besides that.. I have no clue why PostInit is not running.
Also I'm using a heavily stripped down version of MCC with most of the crap removed.

[...]

PR #6607, opened 2018-10-01
Some Refactoring

dedmen:
`findIf` instead of `forEach` loop.
`select` instead of `forEach` with `pushBack`
Lazy eval and reordering conditions by performance impact
more params

### DIFF
--- a/addons/tacticalladder/functions/fnc_cancelTLdeploy.sqf
+++ b/addons/tacticalladder/functions/fnc_cancelTLdeploy.sqf
@@ -32,7 +32,7 @@ GVAR(ladder) animate ["rotate", 0];
 
 {
     GVAR(ladder) animate [_x, 0];
-} count __ANIMS;
+} forEach __ANIMS; //Don't "optimize" this to a count. See #6607
 
 // remove mouse buttons and hint
 call EFUNC(interaction,hideMouseHint);


addons/switchunits/functions/fnc_handleMapClick.sqf
```diff
@@ -21,16 +21,11 @@
...
-private _sideNearest = [];
+private _nearestObjects = nearestObjects [_pos, ["Man"], 15];
 
-{
-    if ([_x] call FUNC(isValidAi) && (side group _x in _sides)) then {
-        _sideNearest pushBack _x;
-    };
-    nil
-} count (nearestObjects [_pos, ["Man"], 15]);
+private _nearestValidUnit = _nearestObjects findIf {(side group _x in _sides) && {[_x] call FUNC(isValidAi)}};
```
Dystopian: maybe `_nearestValidUnitIndex` name would be better

addons/vehiclelock/functions/fnc_hasKeyForVehicle.sqf
```diff
@@ -22,20 +22,16 @@ TRACE_2("params",_unit,_veh);
...
 
 //Check custom keys
 private _customKeys = _veh getVariable [QGVAR(customKeys), []];
-{
-    if (_x in (magazinesDetail _unit)) then {_returnValue = true;};
-} forEach _customKeys;
+private _magazines = magazinesDetail _unit;
 
-_returnValue
+(_customKeys findIf {_x in _magazines}) != -1
```
Dystopian: newline

addons/switchunits/functions/fnc_handleMapClick.sqf
dedmen: That's actually what I was thinking when writing that.. Apparently the thoughts didn't get to my fingers 🤔

addons/tacticalladder/functions/fnc_cancelTLdeploy.sqf
```diff
@@ -32,7 +32,6 @@ GVAR(ladder) animate ["rotate", 0];
 
 {
     GVAR(ladder) animate [_x, 0];
-    true
```
PabstMirror: https://github.com/acemod/ACE3/pull/4815
This was added by you to fix some mcc thing, is this no longer needed?

dedmen: Actually looking back at it.. I don't know how that's supposed to have fixed that bug.
See https://github.com/acemod/ACE3/issues/4814#issue-200180402 
>Second animate should always return Nothing according to BIKI so the error should always be triggered. I don't know why it sometimes apparently returns something that count accepts as a bool.

The `true` was added to make the `count` not error when anything non-bool got to it. Replacing the count by a forEach made it faster (because we don't need the call to `true`) and well.. removed the need for the `true`.

And for whatever reason `animate` would return crap when fed with a nullObj 
https://github.com/acemod/ACE3/issues/4814#issuecomment-272030822
This just removes all constraints on the return values.

dedmen: Only problem now is.. That I didn't actually swap out the `count` for a `forEach` 🤔

dedmen: We probably should fix the core issue instead. Right-clicking while not currently deploying a ladder, shouldn't execute this function. Maybe that was fixed already?
If we fix the underlying issue, `ladder` will not be nil/null here and we won't have this problem.

dedmen: https://github.com/acemod/ACE3/search?q=cancelTLdeploy&unscoped_q=cancelTLdeploy

https://github.com/acemod/ACE3/blob/e5a15d200f44df5fccc0bc5575d18d80b35538dd/addons/tacticalladder/XEH_missionDisplayLoad.sqf#L6
This seems to be the problem. The rightclick and scrollwheel handlers are always registered. Instead of only when you're trying to place a ladder.

Executed from https://github.com/acemod/ACE3/blob/e5a15d200f44df5fccc0bc5575d18d80b35538dd/addons/tacticalladder/CfgEventHandlers.hpp#L30 at mission start.

dedmen: @jonpas added that in https://github.com/acemod/ACE3/pull/3524 

@commy2 already found the problem back then https://github.com/acemod/ACE3/pull/3524#issuecomment-192816064
>previously these actions would only be called when the current action was active (e.g. carrying a ladder), but now all of them are executed when you RMB every time.


Seems like `explosives` and `sandbag` have the same underlying problem, but they are not affected because they exit if the placement PFH is not running.


https://github.com/acemod/ACE3/pull/3524/files#diff-963d8341c69f48d73203fbd72841ef32R23 This check fails. Because the ladder is nil, causing the isNull check to return nil. Causing a `if (nil) exitWith {}` causing a `nil exitWith {}` and the exitWith also being ignored because of the nil parameter.
So either add a isNil check there, or stop always adding the EH and only add it when it's needed.
Should be fairly simple as sandbag/explosives/trenches all check for their placement PFH running, so we can just add the EH when the placement PFH starts, and remove it when it ends.

PabstMirror: I don't see anyway for `GVAR(ladder)` to become nil. 
Was it mcc messing with things?
I'd say just swap back to `forEach`, performance of this code does not matter

dedmen: `performance of this code does not matter` it's executed every time the player makes a right-click. :U 
But if postInit runs properly. It will exit at the isNull check at the start of the file.. So maybe not thaat bad. But the problem of the unused eventhandlers always being added still stands.

PR #6612, opened 2018-10-03
Safemode - Fix indicator when switching units

### LINKED ISSUE #3807: Safe mode indicator broken when remote controlling a unit
**Arma 3 Version:** `1.58` stable
**CBA Version:** `2.3.1` stable
**ACE3 Version:** `3.5.1` stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
When remote controlling a unit after putting your weapon on safe the firing mode indicator will stay empty, even though the weapon is not on safe.

**Steps to reproduce:**
- Be Zeus
- Put your weapon on safe
- Remote control a unit
- Press F, firing mode indicator stays empty even though you can fire your weapon
- Press keybind to put your weapon on safe _twice_
- Firing mode indicator reappears

- Dedicated

**Placed Modules:**
- On demand

**RPT log file:**
- On demand

BaerMitUmlaut:
- Fix #3807

### DIFF
--- a/addons/safemode/XEH_postInit.sqf
+++ b/addons/safemode/XEH_postInit.sqf
@@ -3,13 +3,7 @@
 
 if (!hasInterface) exitWith {};
 
-// IGNORE_PRIVATE_WARNING(_player)
-//["Soldier", {_player = ACE_player; if (currentWeapon _player in (_player getVariable [QGVAR(safedWeapons), []])) then {[false] call FUNC(setSafeModeVisual)}] call EFUNC(common,addInfoDisplayEventHandler);
-//@todo addEventHandler infoDisplayChanged with select 1 == "Soldier"
-
-// add keybinds
-["ACE3 Weapons", QGVAR(safeMode), localize LSTRING(SafeMode),
-{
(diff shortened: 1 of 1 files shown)


BaerMitUmlaut:
Oh also I removed the todo because that seems to be a leftover from AGM about this exact problem.

addons/safemode/XEH_postInit.sqf
```diff
@@ -18,6 +12,9 @@ if (!hasInterface) exitWith {};
     // Statement
     [ACE_player, currentWeapon ACE_player, currentMuzzle ACE_player] call FUNC(lockSafety);
     true
-},
-{false},
-[41, [false, true, false]], false] call CBA_fnc_addKeybind;
+}, {false}, [41, [false, true, false]], false] call CBA_fnc_addKeybind;
```
Dystopian: Magic variable, should be something like `DIK_*`.

BaerMitUmlaut:
Okay, so when would that actually happen? I can't find it ever being raised with "Soldier".

commy2:
>Okay, so when would that actually happen?
Loading a savegame for example.

PR #6613, opened 2018-10-04
Quickmount - Default enable quick mount (empty keybind) and Move to CBA Settings

### LINKED ISSUE #6516: Enabling quick mount should be a one-step process
Simple little suggestion for quick mount: at the moment to enable using quick mount the client has to bind the hotkey _and_ enable it in their ACE settings. Most people I run into have the issue of not knowing they need to enable it. Ideally it should just work off of the hotkey - dunno why I didn't make it like that from the start, but nevertheless it would be sweet.

jonpas:
- Close #6516 

Hotkey is unset by default, no reason to have a setting in that case as described in #6516.

@jameslkingsley review please.

### DIFF
--- a/addons/quickmount/initSettings.sqf
+++ b/addons/quickmount/initSettings.sqf
@@ -1,0 +1,36 @@
+[
+    QGVAR(enabled),
+    "CHECKBOX",
+    localize ELSTRING(common,Enabled),
+    format ["ACE %1", localize LSTRING(Category)],
+    true,
+    true
+] call CBA_settings_fnc_init;
+
+[
+    QGVAR(distance),
+    "SLIDER",
+    [localize LSTRING(Distance), localize LSTRING(DistanceDescription)],
+    format ["ACE %1", localize LSTRING(Category)],
+    [0, 10, DEFAULT_DISTANCE, 0],
+    true
+] call CBA_settings_fnc_init;
+
+[
+    QGVAR(speed),
+    "SLIDER",
+    [localize LSTRING(Speed), localize LSTRING(SpeedDescription)],
+    format ["ACE %1", localize LSTRING(Category)],
+    [0, 30, DEFAULT_SPEED, 0],
+    true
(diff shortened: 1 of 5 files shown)


Dystopian:
What if some server admin would like to disable this feature? I think jameslkingsley meant just enabling by default.

jonpas:
But he also said that's a hacky fix. :D

jameslkingsley:
Looks good to me 👍

Dystopian:
I just think Quick Mount can be not appropriate for some milsim PvP games because it's cheating a little. Setting would be helpful in these cases.
BTW where is that jonpas who asked me to add setting for some features I implemented? ;)

jonpas:
I don't know, not my feature, I'll do what others think is best.

jonpas:
`movedToSQF = 1;`

PabstMirror:
ace_common preinit runs first
if someone has a modified ace_settings in either mission config or configFile, the settings converted would think it was just an unconverted setting and try to add it first

addons/quickmount/initSettings.sqf
```diff
@@ -0,0 +1,36 @@
+[
+    QGVAR(enabled),
+    "CHECKBOX",
+    localize ELSTRING(common,Enabled),
+    format ["ACE %1", localize LSTRING(Category)],
```
Dystopian: `LLSTRING` and `LELSTRING` could be used here and anywhere

commy2: Could simply delete the `localize` , because the settings system can translate the strings on its own.

Dystopian: btw just checked, subcategory is not automatically localized

commy2: Issue report then.

PabstMirror: fixed in https://github.com/CBATeam/CBA_A3/pull/1003

PR #6614, opened 2018-10-04
Force lf line endings for docker entrypoint file

bux:
Docker Container won't start, because the linux image can't execute the entrypoint file with crlf line endings

### DIFF
--- a/docs/.gitattributes
+++ b/docs/.gitattributes
@@ -2,3 +2,4 @@
 *.png binary
 *.jpg binary
 *.paa binary
+*.sh eol=lf


bux:
On windows default git setting is `autocrlf = true` by forcing it to be `lf` we can circumvent such settings.

jonpas:
`autocrlf = true` just auto-converts to CRLF when pulling and back to LF when pushing. There should be no way in having CRLF upstream with `* text = auto`.

But I see what you are saying, I'd blame the user though. :P

bux:
It's not about having `lf` upstream, but having ``lf` locally too. Because docker needs it.

jonpas:
Yeah I get it.

PR #6615, opened 2018-10-04
Consitently capitalize first letters of wiki titles

bux:

### DIFF
--- a/docs/wiki/development/how-to-translate-ace3.md
+++ b/docs/wiki/development/how-to-translate-ace3.md
@@ -1,6 +1,6 @@
 ---
 layout: wiki
-title: How to translate ACE3
+title: How To Translate ACE3
 description: This page describes in short how you can help translating ACE3 into your language. 
 group: development
 parent: wiki
--- a/docs/wiki/development/how-to-translate-ace3.md
+++ b/docs/wiki/development/how-to-translate-ace3.md
@@ -1,6 +1,6 @@
 ---
 layout: wiki
-title: How To Translate ACE3
+title: How to Translate ACE3
 description: This page describes in short how you can help translating ACE3 into your language. 
 group: development
 parent: wiki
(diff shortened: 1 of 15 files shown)


docs/wiki/development/how-to-translate-ace3.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: wiki
-title: How to translate ACE3
+title: How To Translate ACE3
```
synixebrett: `To` should probably remain `to`

docs/wiki/user/ace3-and-acex.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: wiki
-title: ACE3 and ACEX
+title: ACE3 And ACEX
```
synixebrett: `And` should probably remain `and`

docs/wiki/user/how-to-report-an-issue.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: wiki
-title: How to report an issue
+title: How To Report An Issue
```
synixebrett: `to` and `an` should probably remain lowercase

bux:
@synixebrett is that some kind of "rule" in the english language? I wasn't sure if every first letter should be capitalized or not.

Drofseh:
English has `Title Case` where only principal words are capitalized.
Articles (a, an, the, etc.), conjunctions (and, but, or, etc.), prepositions (to, on, in, with, etc.) are not capitalized unless they are the first word in the title.

docs/wiki/user/ace3-and-acex.md
dedmen: Also felt wrong to me but didn't know if it was supposed to be like that

commy2:
The Lord of the Rings

PR #6619, opened 2018-10-05
Fix slow motion falling of SLAM mines

### LINKED ISSUE #2770: SLAM Mine after deploying slowly falls on the ground.
ACE3 Version: 3.3.2
Stable Version 1.52

**Mods:**
- `@CBA_A3`
- `@ace`
- `@DragonFyre`

**Placed ACE3 Modules:**
None

**Description:**
SLAM Mine after laying on the ground slowly falls on the ground. While Claymore is installed without such  glitches. Such physics is not from our planet. Please fix this nonsense.

**Steps to reproduce:**
Put the SLAM mine on the ground

Editor

**RPT log file:**
Not needed

BaerMitUmlaut:
- Fix #2770

Center of mass was previously not centered but too far back. It's now a little lower than the actual model's center. Mass was increased to 5 imaginary units, default is 0. Looks more realistic when falling now.

### DIFF
--- a/addons/explosives/CfgEventHandlers.hpp
+++ b/addons/explosives/CfgEventHandlers.hpp
@@ -38,3 +38,9 @@ class Extended_DisplayLoad_EventHandlers {
         ADDON = QUOTE(_this call COMPILE_FILE(XEH_missionDisplayLoad));
     };
 };
+
+class Extended_Init_EventHandlers {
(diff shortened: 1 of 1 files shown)


commy2:
This similar init script has a locality check: https://github.com/acemod/ACE3/blob/master/addons/vehicles/CfgEventHandlers.hpp#L31

PR #6623, opened 2018-10-06
fix engine start delay turning viv laoded vehicle when starting the engine

commy2:
- fixes: https://gfycat.com/FlimsyHealthyGreendarnerdragonfly


Cyruz143:
Tested in SP editor, fix works.

PR #6628, opened 2018-10-07
Medical- Use patient items first setting

synixebrett:
- Create a setting for using the patient's medical items first. Default is true to preserve current behavior

### DIFF
--- a/addons/medical/functions/fnc_useItem.sqf
+++ b/addons/medical/functions/fnc_useItem.sqf
@@ -24,13 +24,20 @@ if (isNil QGVAR(setting_allowSharedEquipment)) then {
     GVAR(setting_allowSharedEquipment) = true;
 };
 
-if (GVAR(setting_allowSharedEquipment) && {[_patient, _item] call EFUNC(common,hasItem)}) exitWith {
-    if (local _patient) then {
-        ["ace_useItem", [_patient, _item]] call CBA_fnc_localEvent;
-    } else {
-        ["ace_useItem", [_patient, _item], _patient] call CBA_fnc_targetEvent;
+private _use_patiten_item = {
+    params ["_patient", "_item"];
+    if ([_patient, _item] call EFUNC(common,hasItem)) exitWith {
+        if (local _patient) then {
+            ["ace_useItem", [_patient, _item]] call CBA_fnc_localEvent;
+        } else {
+            ["ace_useItem", [_patient, _item], _patient] call CBA_fnc_targetEvent;
+        };
+        [true, _patient];
     };
-    [true, _patient];
+};
+
+if (GVAR(usePatientItemsFirst) && {GVAR(setting_allowSharedEquipment)} && {[_patient, _item] call EFUNC(common,hasItem)}) exitWith {
+    [_patient, _item] call _use_patiten_item;
 };
 
 if ([_medic, _item] call EFUNC(common,hasItem)) exitWith {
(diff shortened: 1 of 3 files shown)


xetra11:
nice one - adds immersion

dedmen:
Add to rewrite too, But might aswell add it to current too until rewrite is ready.
About @Drofseh's thing, if we make it a clientside setting that can be changed mid-mission that would solve that problem already I'd say. The medic can choose by himself.

addons/medical/functions/fnc_useItem.sqf
```diff
@@ -24,13 +24,20 @@ if (isNil QGVAR(setting_allowSharedEquipment)) then {
...
 
-if (GVAR(setting_allowSharedEquipment) && {[_patient, _item] call EFUNC(common,hasItem)}) exitWith {
-    if (local _patient) then {
-        ["ace_useItem", [_patient, _item]] call CBA_fnc_localEvent;
-    } else {
-        ["ace_useItem", [_patient, _item], _patient] call CBA_fnc_targetEvent;
+private _use_patiten_item = {
+    params ["_patient", "_item"];
+    if ([_patient, _item] call EFUNC(common,hasItem)) exitWith {
+        if (local _patient) then {
```
dedmen: All this check adds is a bit of performance
targetEvent already does the check
https://github.com/CBATeam/CBA_A3/blob/fbd59d2ac1f85da79ea18a53b2d14e3ea285f60a/addons/events/fnc_targetEvent.sqf#L35
targetEvent is just a little less efficient.

addons/medical/functions/fnc_useItem.sqf
```diff
@@ -24,13 +24,20 @@ if (isNil QGVAR(setting_allowSharedEquipment)) then {
...
+            ["ace_useItem", [_patient, _item]] call CBA_fnc_localEvent;
+        } else {
+            ["ace_useItem", [_patient, _item], _patient] call CBA_fnc_targetEvent;
+        };
+        [true, _patient];
```
dedmen: Generally we have the convention to leave away the `;` on return values don't we?
But this whole file isn't doing that.

addons/medical/functions/fnc_useItem.sqf
```diff
@@ -24,13 +24,20 @@ if (isNil QGVAR(setting_allowSharedEquipment)) then {
...
     };
-    [true, _patient];
+};
+
+if (GVAR(usePatientItemsFirst) && {GVAR(setting_allowSharedEquipment)} && {[_patient, _item] call EFUNC(common,hasItem)}) exitWith {
+    [_patient, _item] call _use_patiten_item;
```
dedmen: You check `[_patient, _item] call EFUNC(common,hasItem)` one line above.
But you check again inside `_use_patiten_item`
duplicate check. and hasItem isn't the cheapest function.

addons/medical/functions/fnc_useItem.sqf
```diff
@@ -42,6 +49,10 @@ if ([_medic, _item] call EFUNC(common,hasItem)) exitWith {
     [true, _medic];
 };
 
+if (GVAR(setting_allowSharedEquipment) && {[_patient, _item] call EFUNC(common,hasItem)}) exitWith {
+    [_patient, _item] call _use_patiten_item;
```
dedmen: See last comment. Duplicate check.

mharis001:
Closing as this is now included in rewrite. [reference](https://github.com/acemod/ACE3/blob/master/addons/medical_treatment/functions/fnc_useItem.sqf#L24)

PR #6632, opened 2018-10-08
Zeus - Add Repair and Rearm interactions

Dystopian:
- title;
- optimize conditions for zeus submenus a little;
- add conditions for `Stance` and `Remote Control` menus;
- make `Remote Control` statement multi-line;
- simplify `ace_rearm_fnc_rearmEntireVehicleSuccess` params (remove nested array).

### DIFF
--- a/addons/interaction/ACE_ZeusActions.hpp
+++ b/addons/interaction/ACE_ZeusActions.hpp
@@ -36,13 +36,14 @@ class ACE_ZeusActions {
             icon = "\A3\Modules_F_Curator\Data\portraitRemoteControl_ca.paa";
             condition = QUOTE(-1 < (curatorSelected select 0) findIf {_x isKindOf 'CAManBase'});
             statement = QUOTE( \
-                private _unit = objNull; \
-                { \
-                    if ( \
+                private _units = curatorSelected select 0; \
+                private _unit = _units param [ARR_2( \
+                    _units findIf { \
                         side _x in [ARR_4(east,west,resistance,civilian)] \
                         && !(isPlayer _x) \
-                    ) exitWith { _unit = _x; }; \
-                } forEach (curatorSelected select 0); \
+                    }, \
+                    objNull \
+                )]; \
                 bis_fnc_curatorObjectPlaced_mouseOver = [ARR_2('OBJECT',_unit)]; \
                 group _target createUnit [ARR_5('ModuleRemoteControl_F',[ARR_3(0,0,0)],[],0,'NONE')]; \
             );
--- a/addons/interaction/ACE_ZeusActions.hpp
+++ b/addons/interaction/ACE_ZeusActions.hpp
@@ -3,10 +3,11 @@ class ACE_ZeusActions {
     class ZeusUnits {
         displayName = "$STR_A3_RscDisplayCurator_ModeUnits_tooltip";
         icon = "\A3\UI_F_Curator\Data\Displays\RscDisplayCurator\modeUnits_ca.paa";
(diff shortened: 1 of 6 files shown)


addons/interaction/ACE_ZeusActions.hpp
```diff
@@ -33,14 +34,32 @@ class ACE_ZeusActions {
         class remoteControl {
             displayName = "$STR_A3_CfgVehicles_ModuleRemoteControl_F";
             icon = "\A3\Modules_F_Curator\Data\portraitRemoteControl_ca.paa";
-            statement = "_unit = objNull; { if ((side _x in [east,west,resistance,civilian]) && !(isPlayer _x)) exitWith { _unit = _x; }; } forEach (curatorSelected select 0); bis_fnc_curatorObjectPlaced_mouseOver = ['OBJECT',_unit]; (group _target) createUnit ['ModuleRemoteControl_F',[0,0,0],[],0,'NONE'];";
+            condition = QUOTE(-1 < (curatorSelected select 0) findIf {_x isKindOf 'CAManBase'});
+            statement = QUOTE( \
+                private _unit = objNull; \
+                { \
+                    if ( \
```
dedmen: this looks like it should be `findIf`

Dystopian: Can't imagine how. Could you explain please?

dedmen:
```
{
    if (
        side _x in [east,west,resistance,civilian]
        && !(isPlayer _x)
    ) exitWith { _unit = _x; };
} forEach (curatorSelected select 0); 
```
All this does is find the first unit where the condition returns true. Which is also what findIf does, returns the index of the first element where condition returns true..
```
private _unit = (curatorSelected select 0) param [
    (curatorSelected select 0) findIf { side _x in [east,west,resistance,civilian] && !(isPlayer _x) }
    , objNull];
```

Dystopian: I didn't know `param` can handle `-1` as index. Without it this code wouldn't be so elegant.

dedmen: param/params are literally the best commands ever.

PabstMirror:
Works great
I'm wondering if these actions might be better as modules?
Like medical's Full Heal

Dystopian:
> might be better as modules?

I want it to work with many vehicles at the same time, module can handle one vehicle only (as I know).

PabstMirror:
Also, I generally dislike big code in configs because it's hard to read and maintain (`ARR_2(` and quote mess), if this needs work in the future maybe we could think about moving to sqf

Bummeri:
@Dystopian Please do not add any rearm or repair actions to the interact menu. To my knowledge they cannot be easily removed by the scenario designer from that menu. I want to limit the curator actions to enable more versus oriented approach to curator actions. Adding them there limits the way zeus can be used.

@PabstMirror  They would be much better if they were modules because that way I as a scenario designer can limit them. I am able to choose what modules get added to which curator.

I hope you please consider this point now and also in the future.

Dystopian:
@Bummeri really, I didn't think about it. What do you think about setting limiting those menus which mission designer can restrict in scenario?

Bummeri:
@Dystopian If I understand you correctly then that would be a better way to go, since I could disable them from such scenarios.

Dystopian:
I'll make a PR.

Bummeri:
If it is not done in another way then please revert this.

PR #6636, opened 2018-10-11
Quick Mount - Fix turrets without weapon treated as gunner

Dystopian:
- title.

Some vehicles have turrets without weapons. They should be treated as cargo not gunners because they can influence (Quick)mount order.
Vanilla vehicles: `B_T_VTOL_01_infantry_F`, `B_T_VTOL_01_vehicle_F` (passengers in cabin).
There are some vehicles in RHS and quite many in CUP. Especially CUP SUVs (like `CUP_I_SUV_Armored_ION`) where you first get in external cargo seats on SUV sides.

### DIFF
--- a/addons/quickmount/functions/fnc_getInNearest.sqf
+++ b/addons/quickmount/functions/fnc_getInNearest.sqf
@@ -70,13 +70,17 @@ if (!isNull _target &&
                 if ((!(_turretPath isEqualTo [])) && {_target lockedTurret _turretPath}) exitWith {TRACE_1("lockedTurret",_x);};
 
                 if (_effectiveRole == "turret") then {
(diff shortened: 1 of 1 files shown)


addons/quickmount/functions/fnc_getInNearest.sqf
```diff
@@ -93,7 +97,7 @@ if (!isNull _target &&
                         {
                             if ((_x select 2) == _cargoIndex) exitWith {_cargoActionIndex = _forEachIndex};
                         } forEach (fullCrew [_target, "cargo", true]);
```
dedmen: This should be a `findIf` ;)

Dystopian: :laughing: yeah but out of this PR scope :tongue:

Issue #210, opened 2015-03-21
weaponselect turns off laser when cycling through fire modes


bux:
ACE_Dev: e7d7fa8e6c6c2a234429e8800ae1a9b010e4686c

**Mods:**
- @cba_a3_dev (RC6_v1.20.150317)
- @ace_dev

**Modules:**
- None

**Steps to reproduce:**
- Start a night mission
- Equip a MX Rifle with the "IR Laser Pointer" attachment and add some NVG to the unit
- Turn on Night Vision and the Laser
- Press the (default) <kbd>2</kbd> key to switch fire mode
- Laser is now turned off

That's different to the default behavior of the <kbd>F</kbd> key. And it just doesn't make sense. :smile:

Mike-MF:
Closing due to age, inactivity or already solved. May be re-opened if the issue is still present, even better just open a new up-to-date Issue.

PR #6639, opened 2018-10-13
Safe Mode - Fix disabled laser on safety switch

Dystopian:
- title.

### DIFF
--- a/addons/safemode/functions/fnc_lockSafety.sqf
+++ b/addons/safemode/functions/fnc_lockSafety.sqf
@@ -64,7 +64,13 @@ if (_muzzle isEqualType "") then {
 
     _unit selectWeapon _muzzle;
 
-    if (_laserEnabled) then {
+    if (
+        _laserEnabled
+        && {
+            0 == [_unit, currentWeapon _unit] call EFUNC(common,getWeaponIndex)
+            || {"" == primaryWeapon _unit} // Arma switches to primary weapon if exists
+        }
+    ) then {
         {_unit action [_x, _unit]} forEach ["GunLightOn", "IRLaserOn"];
     };
 };
--- a/addons/safemode/functions/fnc_lockSafety.sqf
+++ b/addons/safemode/functions/fnc_lockSafety.sqf
@@ -67,7 +67,7 @@ if (_muzzle isEqualType "") then {
     if (
         _laserEnabled
         && {
-            0 == [_unit, currentWeapon _unit] call EFUNC(common,getWeaponIndex)
+            _muzzle == primaryWeapon _unit // prevent UGL switch
             || {"" == primaryWeapon _unit} // Arma switches to primary weapon if exists
         }
     ) then {
--- a/addons/safemode/functions/fnc_lockSafety.sqf
+++ b/addons/safemode/functions/fnc_lockSafety.sqf
@@ -60,7 +60,13 @@ if (_unit getVariable [QGVAR(actionID), -1] == -1) then {
 };
 
 if (_muzzle isEqualType "") then {
+    private _laserEnabled = _unit isIRLaserOn _weapon || {_unit isFlashlightOn _weapon};
+
     _unit selectWeapon _muzzle;
+
+    if (_laserEnabled) then {
+        {_unit action [_x, _unit]} forEach ["GunLightOn", "IRLaserOn"];
+    };
 };
 
 // play fire mode selector sound
--- a/addons/safemode/functions/fnc_unlockSafety.sqf
+++ b/addons/safemode/functions/fnc_unlockSafety.sqf
@@ -30,7 +30,13 @@ if (_safedWeapons isEqualTo []) then {
     _unit setVariable [QGVAR(actionID), -1];
 };
 
+private _laserEnabled = _unit isIRLaserOn _weapon || {_unit isFlashlightOn _weapon};
+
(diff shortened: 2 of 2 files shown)


Dystopian:
It works when you switch safety with hotkey. It doesn't work when safety is turned off with weapon mode key (related to #210).

commy2:
Pretty sure this breaks something else. UGL muzzles?

Dystopian:
On my PC safety doen't correctly work with UGL before patch too. I don't know about anything else.

addons/safemode/functions/fnc_lockSafety.sqf
```diff
@@ -60,7 +60,13 @@ if (_unit getVariable [QGVAR(actionID), -1] == -1) then {
 };
 
 if (_muzzle isEqualType "") then {
+    private _laserEnabled = _unit isIRLaserOn _weapon || {_unit isFlashlightOn _weapon};
```
alganthe: No need for an extra variable.

addons/safemode/functions/fnc_unlockSafety.sqf
```diff
@@ -30,8 +30,14 @@ if (_safedWeapons isEqualTo []) then {
     _unit setVariable [QGVAR(actionID), -1];
 };
 
+private _laserEnabled = _unit isIRLaserOn _weapon || {_unit isFlashlightOn _weapon};
```
alganthe: Same as above, no need to store that condition, it's used once.

addons/safemode/functions/fnc_lockSafety.sqf
Dystopian: There is need. Laser is disabled right after `selectWeapon` and we need to check if it was enabled.

alganthe: You only use this condition once, you do not need to store it.

Dystopian: I store state not condition. Or I don't understand you.

alganthe: this is a local var that is only used once (the check below) and then is destroyed.

dedmen: >You only use this condition once, you do not need to store it.

You cannot evaluate the condition after LIne 65 though. So how else would you do it?

alganthe: true, I may or may not have seen that selectWeapon ...
wooops.

alganthe:
Bare my comments above, this has a core issue, GunLightOn and IRLaserOn switches you to your primary, there is currently no way to toggle flashlights and IR lasers for players that doesn't have an issue.

Dystopian:
I already met this bug in https://github.com/acemod/ACE3/blob/52abed1d074325b5b4ab9f9dedfc4083667d875e/addons/common/functions/fnc_switchPersistentLaser.sqf#L51 should add workaround here too.

addons/safemode/functions/fnc_unlockSafety.sqf
```diff
@@ -30,8 +30,20 @@ if (_safedWeapons isEqualTo []) then {
...
 _unit selectWeapon _muzzle;
 
+if (
+    _laserEnabled
+    && {
+        _muzzle == primaryWeapon _unit // prevent UGL switch
+        || {"" == primaryWeapon _unit} // Arma switches to primary weapon if exists
+    }
+) then {
+    {_unit action [_x, _unit]} forEach ["GunLightOn", "IRLaserOn"];
```
PabstMirror: I think this should be bellow the `action [SwitchWeapon`

PabstMirror:
MX-GL with modes [single, auto, gl]

Single:
Safe on -> Safe off with keybind: laser on, in semi
Safe on -> Safe off with F (next weapon): laser on, in auto
Auto:
Safe on -> Safe off with keybind: laser on, in semi
Safe on -> Safe off with F (next weapon): laser **disabled**, in auto
GL:
Safe on: laser **disabled**
restart laser
Safe off with keybind: laser **disabled**, in GL
Safe off with F (next weapon): laser **disabled**, in auto

This PR is an improvement over current, so I'd say we merge and fix it later.

Notes for me:
I think the problem is in the `addActionEventHandler` and the weapon swap happens before we think it should. So the currentMuzzle changes first and then the eh runs when we check inputAction.
Also, I can't see any reason why we are doing `selectWeapon` in fnc_lockSaftey
```
private _modes = [];
private _weaponCfg = configFile >> "CfgWeapons" >> _weapon;
{
    private _muzzle = _x;
    private _muzzleCfg = _weaponCfg;
    if (_x == "this") then {
        _muzzle = _weapon;
    } else {
        _muzzleCfg = _muzzleCfg >> _x;
    };
    {
        if (getNumber (configFile >> "CfgWeapons" >> _weapon >> _x >> "showToPlayer") == 1) then {
            _modes pushBack [_muzzle, _x];
        };
        if (_x == "this") then {
            _modes pushBack [_muzzle, _weapon];;
        };
    } forEach getArray (_muzzleCfg >> "modes");
} forEach getArray (_weaponCfg >> "muzzles");


// _previousModeIndex = (_modes find [currentMuzzle, currentMode]) - 1
// (_modes select _targetModeIndex) params [["_targetMuzzle", "#"], ["_targetMode", "#"]];
```

PR #6646, opened 2018-10-18
Vehicle Lock - Enable for planes

### LINKED ISSUE #6637: Vehicle Lock Assign Module  Clarification
Hey guys. Just wanted to ask how exactly do you use this feature? Now what I am trying to accomplish is I'm making a coop multiplayer type map with Alive. I want to restrict a certain pilot to use a certain plane. So that if anybody comes in that is not that specific pilot it would be locked to them.  I have seen you have a Vehicle Lock Assign module and just need a little help understanding how to use this. 

Looking at the "usage" on the framework link it says " Sync the module **with** vehicles and players. Custom keys will be handed to players for every synced vehicle. It is only valid for objects present at mission start."

    -I understood this as to sync the vehicle to the module as well as the specific player(s) to the module then the specific keys will be given out? Am I right on that? When I start the mission there is an error bottom left of my screen that the module has not been synced to a vehicle and no keys are givin out.

Thanks for any info.

Fletcher

https://ace3mod.com/wiki/feature/vehiclelock.html

https://ace3mod.com/wiki/framework/vehiclelock-framework.html

commy2:
- close #6637

### DIFF
--- a/addons/vehiclelock/CfgEventHandlers.hpp
+++ b/addons/vehiclelock/CfgEventHandlers.hpp
@@ -33,4 +33,9 @@ class Extended_InitPost_EventHandlers {
             serverInit = QUOTE(_this call FUNC(handleVehicleInitPost));
         };
     };
+    class Plane {
+        class ADDON {
+            serverInit = QUOTE(_this call FUNC(handleVehicleInitPost));
+        };
+    };
 };
--- a/addons/vehiclelock/CfgEventHandlers.hpp
+++ b/addons/vehiclelock/CfgEventHandlers.hpp
@@ -16,26 +16,3 @@ class Extended_PostInit_EventHandlers {
         init = QUOTE(call COMPILE_FILE(XEH_postInit));
     };
 };
-
-class Extended_InitPost_EventHandlers {
-    class Car {
-        class ADDON {
-            serverInit = QUOTE(_this call FUNC(handleVehicleInitPost));
-        };
-    };
-    class Tank {
-        class ADDON {
-            serverInit = QUOTE(_this call FUNC(handleVehicleInitPost));
-        };
-    };
-    class Helicopter {
-        class ADDON {
-            serverInit = QUOTE(_this call FUNC(handleVehicleInitPost));
-        };
-    };
-    class Plane {
-        class ADDON {
-            serverInit = QUOTE(_this call FUNC(handleVehicleInitPost));
-        };
-    };
-};
--- a/addons/vehiclelock/functions/fnc_handleVehicleInitPost.sqf
+++ b/addons/vehiclelock/functions/fnc_handleVehicleInitPost.sqf
@@ -27,7 +27,7 @@ TRACE_1("params",_vehicle);
 
     params ["_vehicle"];
 
-    if ((_vehicle isKindOf "Car") || {_vehicle isKindOf "Tank"} || {_vehicle isKindOf "Helicopter"}) then {
+    if ((_vehicle isKindOf "Car") || {_vehicle isKindOf "Tank"} || {_vehicle isKindOf "Helicopter"} || {_vehicle isKindOf "Plane"}) then {
         //set lock state (eliminates the ambigious 1-"Default" and 3-"Locked for Player" states)
         private _lock = switch (GVAR(VehicleStartingLockState)) do {
             case (0): { (locked _vehicle) in [2, 3] };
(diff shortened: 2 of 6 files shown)


addons/vehiclelock/CfgEventHandlers.hpp
```diff
@@ -33,4 +33,9 @@ class Extended_InitPost_EventHandlers {
             serverInit = QUOTE(_this call FUNC(handleVehicleInitPost));
         };
     };
+    class Plane {
+        class ADDON {
+            serverInit = QUOTE(_this call FUNC(handleVehicleInitPost));
```
TheMagnetar: `QFUNC(handleVehicleInitPost)`?

commy2: That would break it, because it removes the `call`.

addons/vehiclelock/functions/fnc_handleVehicleInitPost.sqf
```diff
@@ -27,7 +27,7 @@ TRACE_1("params",_vehicle);
 
     params ["_vehicle"];
 
-    if ((_vehicle isKindOf "Car") || {_vehicle isKindOf "Tank"} || {_vehicle isKindOf "Helicopter"}) then {
+    if (_vehicle isKindOf "Car" || {_vehicle isKindOf "Tank" || {_vehicle isKindOf "Helicopter" || {_vehicle isKindOf "Plane"}}}) then {
```
dedmen: Can't we combine them into IsKindOf Air?

Dystopian:
```suggestion
    if (-1 < ["Car", "Tank", "Helicopter", "Plane"] findIf {_vehicle isKindOf _x}) then {
```
github suggestion is cool :)

addons/vehiclelock/functions/fnc_moduleSync.sqf
```diff
@@ -28,7 +28,7 @@ if !(_activated) exitWith {WARNING("Vehicle Lock Sync Module - placed but not ac
     params ["_syncedObjects"];
 
     private _listOfVehicles = _syncedObjects select {
-        (_x isKindOf "Car") || {(_x isKindOf "Tank") || {_x isKindOf "Helicopter"}}
+        _x isKindOf "Car" || {_x isKindOf "Tank" || {_x isKindOf "Helicopter" || {_x isKindOf "Plane"}}}
```
Dystopian:
```suggestion
        private _vehicle = _x;
        -1 < ["Car", "Tank", "Helicopter", "Plane"] findIf {_vehicle isKindOf _x}
```
Also I think this array of classes should become macro because it's used at least in 2 places.

addons/vehiclelock/functions/fnc_handleVehicleInitPost.sqf
commy2: Like, except the -1 should be at the end and the < be a >=.

PabstMirror: this whole line is pointless
func is only called from xeh added to specific vehicle types
https://github.com/acemod/ACE3/blob/master/addons/vehiclelock/CfgEventHandlers.hpp#L23

addons/vehiclelock/CfgEventHandlers.hpp
TheMagnetar: oops... too much time with CBA state machine

addons/vehiclelock/functions/fnc_moduleSync.sqf
```diff
@@ -28,7 +28,8 @@ if !(_activated) exitWith {WARNING("Vehicle Lock Sync Module - placed but not ac
     params ["_syncedObjects"];
 
     private _listOfVehicles = _syncedObjects select {
-        (_x isKindOf "Car") || {(_x isKindOf "Tank") || {_x isKindOf "Helicopter"}}
+        #define CLASSNAMES ["Car", "Tank", "Air"]
```
Dystopian: macro is useless here

addons/vehiclelock/functions/fnc_moduleSync.sqf
```diff
@@ -28,7 +28,8 @@ if !(_activated) exitWith {WARNING("Vehicle Lock Sync Module - placed but not ac
...
+        IS_KIND_OF_ANY(_x,CLASSNAMES)
```
Dystopian: you now have `findIf {(_x) isKindOf _x}` after preprocessing

addons/vehiclelock/functions/fnc_moduleSync.sqf
commy2: Commas break the args in a macro, because the preprocessor cannot handle the parentheses syntax. You'd have to use the ARR_3 macro which is ugly.

PR #6647, opened 2018-10-23
Make small UAVs battery life more realistic

alganthe:
- Balance the darter battery life to be around 30 minutes.
- Balance the AL-6 Pelican battery life to be around 25 minutes.

Quadcopters and hexacopters can currently fly for hours, not only this is wildly unrealistic, it removes the need for UAV batteries altogether and makes them able to fly absolutely insane distances, this PR fixes this.

Fuel capacity was calculated by giving each drone 1 fuel capacity and making them hover for a minute then multiply the result by the desired battery life.

### DIFF
--- a/addons/logistics_uavbattery/CfgVehicles.hpp
+++ b/addons/logistics_uavbattery/CfgVehicles.hpp
@@ -11,6 +11,7 @@ class CfgVehicles {
         };
     };
     class UAV_01_base_F: Helicopter_Base_F {
+        fuelCapacity = 19; // Around 30 minutes hovering
         class ACE_Actions: ACE_Actions{
             class ACE_MainActions: ACE_MainActions {
                 class GVAR(RefuelUAV) {
@@ -23,6 +24,7 @@ class CfgVehicles {
         };
     };
     class UAV_06_base_F: Helicopter_Base_F {
+        fuelCapacity = 16; // Around 25 minutes hovering
         class ACE_Actions: ACE_Actions{
             class ACE_MainActions: ACE_MainActions {
                 class GVAR(RefuelUAV) {


bux:
lgtm

addons/logistics_uavbattery/CfgVehicles.hpp
```diff
@@ -11,6 +11,7 @@ class CfgVehicles {
         };
     };
     class UAV_01_base_F: Helicopter_Base_F {
+        fuelCapacity = 19; // Around 30 minutes hovering
```
Dystopian: Is there a formula to calculate these values? Could it be more useful than magic numbers?

alganthe: I couldn't find info on the formula anywhere, it's a BI secret apparently.

We could try asking them, but from my findings it's engine RPM + some other factor I couldn't identify.

jonpas:
You're using the wrong mod then!

addons/logistics_uavbattery/CfgVehicles.hpp
commy2: So fuelConsumptionRate is ignored for carx, tankx, motorcycle?

jokoho48: no fuelConsumptionRate is (0.01+rpm*0.02) for cars, as far as i understood it.

alganthe: ^ this seems close to what I observed.

Also I don't see anything relating to UAV RPMs in configs, so i'm not sure what's up with that.

commy2: fuelConsumptionRate is a config entry. It cannot be a formula like that.

dedmen: Exactly. fuelConsumptionRate is a float.
I'm too lazy to check again now. But I think what I wrote about Car==Carx

jonpas: Anything to be done here or good to merge?

dedmen: Good to merge. We might be able to calculate more exact numbers so that they last 30.0 minutes instead of 29.8 minutes or so.. Nobody cares about that.

Dystopian: I thought these numbers was calculated somehow and not just guessed. And then formula here instead magic number would be better for future.

dedmen: They were calculated
>Fuel capacity was calculated by giving each drone 1 fuel capacity and making them hover for a minute then multiply the result by the desired battery life.

https://github.com/acemod/ACE3/pull/6647#issue-225022928

PR #6651, opened 2018-10-23
AI - Fix units not being removed from the garrison unitMove list consistently

alganthe:
- Fix an issue where units would keep trying to path until it fails even if they're ungarrisoned or garrisoned via teleport, this would also keep drawing the red lines to the old positions.

Not sure how I missed this during testing.

### DIFF
--- a/addons/ai/functions/fnc_unGarrison.sqf
+++ b/addons/ai/functions/fnc_unGarrison.sqf
@@ -43,7 +43,7 @@ _units = _units select {local _x};
             _unit doMove ((nearestBuilding (getPos _unit)) buildingExit 0);
         };
 
-        if ({(_unit getVariable [QGVAR(garrisonned), false]) && !isPlayer _unit} count (units _unit) == 0) then {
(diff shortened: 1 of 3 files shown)


addons/ai/functions/fnc_unGarrison.sqf
```diff
@@ -21,27 +21,31 @@ params [["_units", [], [[]]]];
...
+            doStop _unit;
+            _unit doFollow _leader;
 
         } else {
-            _x doMove ((nearestBuilding (getPos _x)) buildingExit 0);
+            _unit doMove ((nearestBuilding (getPos _unit)) buildingExit 0);
         };
 
-        if ({(_x getVariable [QGVAR(garrisonned), false]) && !isPlayer _x} count (units _x) == 0) then {
+        if ({(_unit getVariable [QGVAR(garrisonned), false]) && !isPlayer _unit} count (units _unit) == 0) then {
```
dedmen: count == 0 could be a findIf == -1 instead.

PR #6652, opened 2018-10-24
Interaction - Improve mouse hint

mharis001:
- Improve spacing of mouse hint
- Change font of extra icon sets
- Before: 
mouse_hint_before
- After: 
mouse_hint_after

### DIFF
--- a/addons/interaction/functions/fnc_hideMouseHint.sqf
+++ b/addons/interaction/functions/fnc_hideMouseHint.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: Garth de Wet (LH)
- * Hides the interaction helper text with the mouse buttons at the bottom middle of the screen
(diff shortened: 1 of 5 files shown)


mharis001:
Currently have no fade for showing the hint and added one for hiding it. No fade when showing IMO improves the experience, for example, refuel's updating mouse hints. Thoughts?

addons/interaction/functions/fnc_hideMouseHint.sqf
```diff
@@ -10,14 +10,12 @@
  * None
  *
  * Example:
- * call ace_interaction_fnc_hideMouseHint
+ * [] call ace_interaction_fnc_hideMouseHint
```
Dystopian: `[]` is not necessary

PabstMirror:
spacing looks much better and I like the font change to differentiate keys

PR #6609, opened 2018-10-02
Split interaction menu settings between interaction and self interaction

synixebrett:
**When merged this pull request will:**
 - Title

This is something that's always annoyed me about self interaction, so I figured I'd finally do something about it. Pretty minor but personally not seeing people wave their guns in the air as they navigate self interaction is a decent improvement.

Edit note: By default both backgrounds are set to disabled, my personal setting is the blur screen.

PR #6656, opened 2018-10-28
Interact menu - Convert 'use list menu' setting to 'use wheel menu'

alganthe:
- Convert "use list menu" setting to "use wheel menu"  which is false by default.

Lists are superior, can display more elements at once and are easier to read.

### DIFF
--- a/addons/interact_menu/ACE_Settings.hpp
+++ b/addons/interact_menu/ACE_Settings.hpp
@@ -21,12 +21,12 @@ class ACE_Settings {
         category = CSTRING(Category_InteractionMenu);
         displayName = CSTRING(AlwaysUseCursorInteraction);
     };
-    class GVAR(UseWheelMenu) {
+    class GVAR(UseWRadialMenu) {
         value = 0;
         typeName = "BOOL";
         isClientSettable = 1;
         category = CSTRING(Category_InteractionMenu);
-        displayName = CSTRING(UseWheelMenu);
+        displayName = CSTRING(UseRadialMenu);
     };
     class GVAR(colorTextMax) {
         value[] = {1, 1, 1, 1};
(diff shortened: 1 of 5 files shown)


Drofseh:
I do think `circle` sounds better than `wheel` in English.
"Use circular menu"

alganthe:
it's all synonyms, radial sound good too but it might be harder to translate, so I kept it simple

Drofseh:
It doesn't work grammatically in English because wheel is a noun.
It would have to be "Use wheel shaped menu".
Circular is an adjective, so "Use circular menu" works.

PabstMirror:
Is this a real PR or a shitpost?

alganthe:
That's not very nice pabst, and yes this is a real PR, list are superior to radial and should be made default, they can display more entries with longer text and are easier to navigate, as such they should be made the default behavior of the interaction menu.

bux:
Let's not change years old defaults.

jonpas:
I am undecided, if most agree it should not be merged, let's just close it.

PabstMirror:
Sorry for my rude comment
https://github.com/acemod/ACE3/pull/6609 was just merged 
so if we wanted to change any defaults now would be a good time
but I still think keeping defaults the same is for the best

commy2:
I would've just changed value = 0; to value = 1;

Drofseh:
I would still like to see this changed.
List is vastly superior to radial for ease of use.
Sub-nodes appear in more predictable places, you can have a much larger number of sub-nodes before you run into space constraints, and you can have longer node name strings before the end of the string runs into another node.

jonpas:
I am still in favour of this change, and @commy2's comment:

> I would've just changed value = 0; to value = 1;

SilentSpike:
I personally think list should be the default, however would just change the default setting value rather than create a new setting for backwards compatibility.

jonpas:
Replaced by #8217.

PR #6657, opened 2018-10-29
Medical - optimize handle damage caching

commy2:
- title
- cuts back on excessive PFH use, making missions that run long reach the PFH limit


commy2:
>        if (diag_frameno >= _frameno + 2) then {
Shouldn't this be actually two frames later, despite the comment saying 3?

PabstMirror:
On a single frame, does handleDamage come before or after PFEHs are run?
Not sure if it matters or not, but it might be why we thought it was 3

PabstMirror:
Could also use `CBA_fnc_waitUntilAndExecute` if we wanted to keep identical logic

commy2:
>Could also use CBA_fnc_waitUntilAndExecute if we wanted to keep identical logic

But execNextFrame is extremely simple. Just an if and a pushBack. I really like stacking execNextFrame if it is ~2-4 frames delay.

commy2:
> Not sure if it matters or not, but it might be why we thought it was 3

If you would've used diag_log diag_frameNo in HandleDamage and this one, it would still be a difference of two frames, regardless which comes first.

PR #6663, opened 2018-11-06
Add Event for tripflare triggering

nomisum:
- add a CBA global event to tripflare triggering
- _flare might be useful to delete / replace flare
- _position is already available, so why not send it too (no need to getpos on the other side)

### DIFF
--- a/addons/explosives/functions/fnc_spawnFlare.sqf
+++ b/addons/explosives/functions/fnc_spawnFlare.sqf
@@ -22,4 +22,6 @@ private _flare = "ACE_TripFlare_FlareEffect" createVehicle [_posX,_posY,_posZ];
 
 TRACE_1("",_flare);
 
+["ace_tripflareTriggered", [_flare, [_posX,_posY,_posZ]]] call CBA_fnc_globalEvent;
+
 nil


TheMagnetar:
Can this be also documented?

nomisum:
Added documentation, also added missing wirecutter event documentation

Renamed Logistics section to Cargo, as it contains cargo specific stuff only and "logistics" is now only wirecutter and uav batteries afaik

docs/wiki/framework/events-framework.md
```diff
@@ -47,7 +47,7 @@ MenuType: 0 = Interaction, 1 = Self Interaction
 |`ace_interactMenuOpened` | [_menuType] | Local | Listen | Interaction Menu Opened
 |`ace_interactMenuClosed` | [_menuType] | Local | Listen | Interaction Menu Closed
 
-### 2.4 Logistics (`ace_cargo`)
+### 2.4 Cargo (`ace_cargo`)
```
bux: careful, this will make some potential incoming links kaputt!
Could you please do a string search for pages that link to that section (if there are any) and rewrite that link?

nomisum: 0 matches in 5251 files it seems 👍

dedmen: Only thing that would make sense to link here would be https://ace3mod.com/wiki/framework/cargo-framework.html#21-listenable
But as you can see it has it's own table. Maybe we could just link here instead of having a copy of the table?
Would be bad UX tho.

nomisum: I had the same irritation. Would be great if the events site could aggregate all the event sections from the single framework pages.

Buuuut I dont think reworking the entire structure is in the scope of this PR ;)

jonpas: Could write a script to aggregate and link them all across. I like them on framework pages, each component gets one and is harder to miss something outdated.

Not in scope of this PR as you said.

PR #6665, opened 2018-11-07
Interaction - Add condition to zeus actions

Dystopian:
- title.

Zeus modules can be disabled/enabled with `removeCuratorAddons`/`addCuratorAddons` commands. This PR adds condition which checks if addon is enabled for curator and thus zeus interaction menu are controlled the same way.

### DIFF
--- a/addons/main/script_macros.hpp
+++ b/addons/main/script_macros.hpp
@@ -134,4 +134,6 @@
 #define MRAD_TO_DEG(d) ((d) / 17.45329252) // Conversion factor: 9 / (50 * PI)
 #define MOA_TO_RAD(d) ((d) * 0.00029088) // Conversion factor: PI / 10800
 
+#define ZEUS_ACTION_CONDITION (QUOTE(QUOTE(ADDON)) in curatorAddons _target)
+
 #include "script_debug.hpp"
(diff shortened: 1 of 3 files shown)


addons/main/script_macros.hpp
```diff
@@ -134,4 +134,6 @@
 #define MRAD_TO_DEG(d) ((d) / 17.45329252) // Conversion factor: 9 / (50 * PI)
 #define MOA_TO_RAD(d) ((d) * 0.00029088) // Conversion factor: PI / 10800
 
+#define ZEUS_ACTION_CONDITION (QUOTE(QUOTE(ADDON)) in curatorAddons _target)
```
Dystopian: not sure about macro name because double `QUOTE` can't be used in actions added by script not config

PabstMirror: `"ace_zeus" in curatorAddons z` is ~0.2ms with just ace or 0.5 with my full modset
so we might want to use cached call instead
```
#define ZEUS_ACTION_CONDITION ([nil, {QUOTE(QUOTE(ADDON)) in curatorAddons _target}, missionNamespace, QUOTE(QUOTE(GVAR(zeusCheck))), 60] call ace_common_fnc_cachedCall)
```

Dystopian: Maybe add event to clear this cache, e.g. that one on interaction menu closed (don't remember exactly name)? 60s timeout can be not so effective if `removeCuratorAddons`/`addCuratorAddons` are used.

PabstMirror: `ace_interactMenuClosed`

PR #6667, opened 2018-11-09
Ballistics/Realistic Names - Integrate P90 Compat

PabstMirror:


jonpas:
Remove compat?

Why not put in fire mode restrictions? They are realistic.

PabstMirror:
Wasn't sure about the fire modes, I'll re-add them

PabstMirror:
It might make sense to release 1 last dummy compat, so we can delete it after release
giving more time for people to unsubscribe?

Drofseh:
Add it's magwell in this PR?

jonpas:
That would be CBA, no?

Drofseh:
lmao, yes, yes, it would.

addons/ballistics/CfgWeapons.hpp
```diff
@@ -70,6 +70,18 @@ class CfgWeapons {
         };
     };
 
+    // P90 (1.86)
+    class SMG_03_TR_BASE: Rifle_Base_F {
+        ACE_barrelTwist = 228.6; // 1:9 inch twist
+        ACE_barrelLength = 407;
+        ACE_twistDirection = 1;
+        modes[]= {"Single"};
```
jonpas:
```suggestion
        modes[] = {"Single"};
```

addons/ballistics/CfgWeapons.hpp
```diff
@@ -70,6 +70,18 @@ class CfgWeapons {
...
+    };
+    class SMG_03C_BASE: SMG_03_TR_BASE {
+        ACE_barrelLength = 264;
+        modes[]= {"Single", "FullAuto"};
```
jonpas:
```suggestion
        modes[] = {"Single", "FullAuto"};
```

addons/main/config.cpp
```diff
@@ -8,6 +8,7 @@ class CfgPatches {
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {
             "A3_Data_F_Tank_Loadorder",
+            "A3_Data_F_Mod_Loadorder",
```
jonpas: Does this one not require `A3_Data_F_Tank_Loadorder`?

addons/realisticnames/CfgWeapons.hpp
```diff
@@ -93,6 +93,60 @@ class CfgWeapons {
     class arifle_Mk20_GL_plain_F: arifle_Mk20_GL_F {
         displayName = CSTRING(arifle_Mk20_GL_plain_Name);
     };
+    
+    // P90 (1.86)
+    class SMG_03_TR_BASE;
+    class SMG_03_TR_black: SMG_03_TR_BASE{
```
jonpas:
```suggestion
    class SMG_03_TR_black: SMG_03_TR_BASE {
```

addons/realisticnames/stringtable.xml
```diff
@@ -3545,6 +3545,294 @@
             <Chinesesimp>C-More Railway SMG (绿色)</Chinesesimp>
             <Japanese>C-More レイルウェイ SMG (緑)</Japanese>
             <Italian>C-More Railway SMG (Verde)</Italian>
+        </Key>
+                <Key ID="STR_ACE_RealisticNames_P90_TR_Black_Name">
```
jonpas:
```suggestion
        <Key ID="STR_ACE_RealisticNames_P90_TR_Black_Name">
```

jonpas:
@PabstMirror don't people get automatically unsubscribed if the mod doesn't exist anymore? Anyone using it on Workshop will get the main ACE3 update anyways at the same time (or rather earlier since we do compats last).

Blutze:
Very minor (and rather old) issue: Alphabetical sorting in the arsenal puts a variety of Russian weaponry between P90 and PS90. How about adding "FN" to the names?

And on the topic of things that should be changed in CBA, the current release on the SWS isn't JointRails compatible.

jonpas:
Agreed, adding FN would make sense.

Doesn't our P90 compat make them joint rails compatible? I thought it did. It should in that case, but side rails are smaller and idk what would fit without adding new rail definition (which would have to be done in CBA).

Blutze:
> Doesn't our P90 compat make them joint rails compatible?

Don't see anything in the code, and using the two Workshop mods didn't do it either. Other vanilla weapons get their JR compat in CBA as well: https://github.com/CBATeam/CBA_A3/blob/master/addons/jr/cfgweapons.hpp.

I wouldn't be concerned about the pure length of light rails - if anything, combo modules with a white light on the side and a laser on top are more problematic, and are so for other weapons too. Short optics rails should already be a thing.

jonpas:
Ah cool, guess CBA can add it for next release then if someone wants to PR it.

addons/main/config.cpp
PabstMirror: it seems to just be itself, unless I'm reading it wrong

dedmen: https://cdn.discordapp.com/attachments/441242389823225876/519888830179049483/1.JPG

Fail

PR #6671, opened 2018-11-11
Arsenal - Add Night Vision support stat for scopes

dedmen:


addons/arsenal/functions/fnc_statTextStatement_scopeVisionMode.sqf
```diff
@@ -0,0 +1,68 @@
...
+ * Return Value:
+ * String to display
+ *
+ * Public: No
+ */
+
+params ["", "_config"];
+TRACE_1("statTextStatement_scopeVisionMode",_config);
+
+private _opticsModes = ("true" ConfigClasses (_config >> "ItemInfo" >> "OpticsModes")) apply {
```
Dystopian:
```suggestion
private _opticsModes = ("true" configClasses (_config >> "ItemInfo" >> "OpticsModes")) apply {
```

dedmen: Wow I "fixed" from configclasses to ConfigClasses :man_facepalming:

dedmen:
Currently the stat is that the top priority wise. Should it be moved to the bottom of the stack?

alganthe:
would be better at the bottom so more important stats don't overflow to the next page.

dedmen:
Is that correct? @alganthe

alganthe:
yes

PR #6673, opened 2018-11-11
ACE Arsenal - Fix NVG overlay appearing in ace arsenal

alganthe:
- Fix the NVG overlay appearing in ace arsenal by toggling off the NVGs when entering ace arsenal.
Works on remote units and AI.

### DIFF
--- a/addons/arsenal/functions/fnc_onArsenalOpen.sqf
+++ b/addons/arsenal/functions/fnc_onArsenalOpen.sqf
@@ -296,6 +296,10 @@ if (is3DEN) then {
 } else {
     GVAR(centerNotPlayer) = (GVAR(center) != player);
 
+    if (currentVisionMode GVAR(center) == 1) then {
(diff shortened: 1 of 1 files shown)


addons/arsenal/functions/fnc_onArsenalOpen.sqf
```diff
@@ -296,6 +296,10 @@ if (is3DEN) then {
 } else {
     GVAR(centerNotPlayer) = (GVAR(center) != player);
 
+    if (currentVisionMode ACE_Player == 1) then {
+        GVAR(center) action ["NVGogglesOff", GVAR(center)];
```
PabstMirror: Should this line also use `ace_player`?

alganthe: both are the same in that case, it doesn't really matter outside readability

PR #6677, opened 2018-11-13
Wirecutters - Add compatibility with more fences from CUP

madpat3:
- add three more fences from CUP-Core, to be cut down with Ace-Wirecutter.

### DIFF
--- a/addons/logistics_wirecutter/script_component.hpp
+++ b/addons/logistics_wirecutter/script_component.hpp
@@ -50,7 +50,10 @@
     "wall_fen1_5.p3d",\
     "plot_provizorni.p3d",\
     "plp_ctm_partitioningfencegrey.p3d",\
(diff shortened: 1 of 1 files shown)


addons/logistics_wirecutter/script_component.hpp
```diff
@@ -50,7 +50,10 @@
     "wall_fen1_5.p3d",\
     "plot_provizorni.p3d",\
     "plp_ctm_partitioningfencegrey.p3d",\
-    "fence.p3d"\
+    "fence.p3d,"\
```
kamkill01011: I think it should be `"fence.p3d",` (coma at the end of line) no ?

madpat3: yes, after the " not before ;)

PR #6679, opened 2018-11-15
Medical - Make vehicle explosions kill players even with prevent instant death enabled

Cyruz143:
**When prevent instant death is enabled, vehicles exploding will now kill occupants**

I'm not quite sure if this is correct for prevent instant death but it seems a bit silly for your vehicle to explode and you just flop out unconscious. Also if you're playing with cookoff, you generally die to cookoff after falling out anyway.

This is slightly counter to the actual name of the option however so open to input on this, I did ask in Slack a while ago but didn't get any responses.

### DIFF
--- a/addons/medical/functions/fnc_handleDamage.sqf
+++ b/addons/medical/functions/fnc_handleDamage.sqf
@@ -1,3 +1,4 @@
+#include "script_component.hpp"
 /*
  * Author: KoffeinFlummi, Glowbal, commy2
  * Main HandleDamage EH function.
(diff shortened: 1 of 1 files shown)


alganthe:
I'm not sure about this one, the current behavior leads to interesting gameplay.

xrufix:
If at all, this should be optional.

In Arma, vehicles often explode in cases in which in reality they would only be disabled or heavily damaged. In my opinion in such cases it should be possible to rescue injured crew instead of just abandoning them.

SilentSpike:
I believe our design intent was that prevent instant death should also prevent instant death on vehicle explosion. If the explosion is severe enough, death should come shortly after from secondary damage events if the system is designed correctly.

Cyruz143:
You've convinced me, this is a bit outside of scope for PID. Closing down.

PR #6680, opened 2018-11-15
Medical - Adding option to convert vanilla medical gear

Cyruz143:
**Give an option to convert vanilla medical items to ACE ones**
Adds a setting to disable the conversion from vanilla medical to ACE medical but still removes the original item. Defaults to true to keep current behaviour the same.

Figured it might be useful for people running missions with vanilla gear in and they don't want them being auto-converted to ACE meds, rather just having them removed.

### DIFF
--- a/addons/medical/stringtable.xml
+++ b/addons/medical/stringtable.xml
@@ -5665,5 +5665,11 @@
             <Chinese>當隊員對嗎啡/腎上腺素/阿托品過量時，會使他的傷害更重</Chinese>
             <Italian>Rende il paziente suscettibile di overdose da morfina, epinefrina o atropina.</Italian>
         </Key>
+        <Key ID="STR_ACE_Medical_MedicalSettings_convertVanillaMedicalItems_DisplayName">
+            <English>Convert vanilla medical items to ACE medical items</English>
+        </Key>
+        <Key ID="STR_ACE_Medical_MedicalSettings_convertVanillaMedicalItems_Description">
+            <English>Enables or disables if vanilla medical items are converted to ACE medical items or removed</English>
+        </Key>
     </Package>
 </Project>
--- a/addons/medical/ACE_Settings.hpp
+++ b/addons/medical/ACE_Settings.hpp
@@ -310,4 +310,11 @@ class ACE_Settings {
         value = 3;
         sliderSettings[] = {0, 30, 3, 0};
     };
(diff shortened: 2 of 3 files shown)


addons/medical/stringtable.xml
```diff
@@ -5665,5 +5665,11 @@
             <Chinese>當隊員對嗎啡/腎上腺素/阿托品過量時，會使他的傷害更重</Chinese>
             <Italian>Rende il paziente suscettibile di overdose da morfina, epinefrina o atropina.</Italian>
         </Key>
+        <Key ID="STR_ACE_Medical_MedicalSettings_convertVanillaMedicalItems_DisplayName">
+            <English>Convert vanilla medical items to ACE medical items</English>
+        </Key>
+        <Key ID="STR_ACE_Medical_MedicalSettings_convertVanillaMedicalItems_Description">
+            <English>Enables or disables if vanilla medical items are converted to ACE medical items or removed</English>
```
dedmen: Reviewed before reading the translations
`Enables or disables whether vanilla medical items are converted to ACE medical items or just removed`

jonpas:
Setting name isn't really fitting, because instead of converting them, it deletes them. Maybe 3 values (convert, keep, delete)? Or rename the setting.

Setting should also be CBA setting since it's new.

Is the setting title not a bit too long?

Cyruz143:
> 
> 
> Setting name isn't really fitting, because instead of converting them, it deletes them. Maybe 3 values (convert, keep, delete)? Or rename the setting.
> 

I don't see the point in keeping them, having ACE medical loaded means you can't actually use them as we remove the action for the player.

> Setting should also be CBA setting since it's new.
> 

I wasn't sure on this, none of the existing medical ones are done, I couldn't find any initSettings.sqf for them anywhere. I kinda assume this is done in rewrite and these are being left as-is. I didn't want to do just add one CBA_Setting for a single option and also didn't want to convert all the existing ones.

> Is the setting title not a bit too long?

Yes it was, fixed it now!

Blutze:
> I don't see the point in keeping them, having ACE medical loaded means you can't actually use them as we remove the action for the player.

There is at least one mission in the base game that expects you to use that action on an NPC to progress, and having ACE Medical loaded currently breaks it. I guess adding at action to manually split them into ACE items, instead of getting split automatically on pickup, would increase the complexity of this PR (and lead to balancing issues where players could just bring vanilla items in ACE missions).

PR #6681, opened 2018-11-16
Medical - Show tourniquets in Medical and Patient information

diwako:
Show applied tourniquets in the medical menu and Patient information view
tourniquets

Adds 4 new images for each tourniquet
Adds code to handle the opacity of all 4 new UI entries respectively for the medical menu and the Patient information view

I know the medical rewrite is in the works, but this might be months and maybe like a year away. That is why I thought adding this in the mean time has some benefits even if it will be removed in the future.


jonpas:
Is that using rewrite pictures or your custom ones? (If I am not mistaken rewrite does it a bit differently anyways?)

diwako:
Correct, I use the rewrite picture for the 4 tourniquets. In the rewrite iirc it uses the same image for all 4 but with rotation and so on. In mine I just made 4 images all like the separate body part images.

TheMagnetar:
Why don't work on rewrite instead?

diwako:
> Why don't work on rewrite instead?

I do not have any real knowledge about the rewrite so far, right now I am just reacting so some things that I think are missing and a temp solution would be neat until the big update hits.

addons/medical_menu/ui/menu.hpp
```diff
@@ -309,6 +309,22 @@ class GVAR(medicalMenu) {
             idc = 55;
             text = QPATHTOEF(medical,ui\body_leg_right.paa);
         };
+        class bodyImgTournAL:bodyImgHead {
```
jonpas: Space after :.

addons/medical_menu/functions/fnc_updateTourniquets.sqf
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+params ["_display", "_tourniquets"];
+
+// Handle tourniquets
+private _availableSelections = [0, 0, 56, 57, 58, 59];
+for "_i" from 2 to 5 do {
+    private _tourn = _tourniquets select _i;
+    (_display displayCtrl (_availableSelections select _i)) ctrlSetTextColor [0, 0, 0.8, _tourn];
+};
```
jonpas: Missing empty last line.

addons/medical_menu/functions/fnc_updateTourniquets.sqf
```diff
@@ -0,0 +1,26 @@
+#include "script_component.hpp"
+/*
+ * Author: diwako
+ * Update tourniquet on the body image on the menu
+ *
+ * Arguments:
+ * 0: display <DISPLAY>
+ * 1: tourniquets <ARRAY>
```
jonpas: Tourniquets

addons/medical_menu/functions/fnc_updateTourniquets.sqf
```diff
@@ -0,0 +1,26 @@
+#include "script_component.hpp"
+/*
+ * Author: diwako
+ * Update tourniquet on the body image on the menu
+ *
+ * Arguments:
+ * 0: display <DISPLAY>
```
jonpas: Display

addons/medical_menu/functions/fnc_updateTourniquets.sqf
```diff
@@ -0,0 +1,26 @@
+#include "script_component.hpp"
+/*
+ * Author: diwako
+ * Update tourniquet on the body image on the menu
```
jonpas: Missing final dot.

addons/medical/functions/fnc_displayPatientInformation.sqf
```diff
@@ -145,6 +145,13 @@ if (_show) then {
             (_display displayCtrl (_availableSelections select _foreachIndex)) ctrlSetTextColor [_red, _green, _blue, 1.0];
         } foreach _selectionBloodLoss;
 
+        private _tourniquets = _target getVariable [QGVAR(tourniquets), [0,0,0,0,0,0]];
```
dedmen: This is just a copy of `ace_medical_menu_fnc_updateTourniquets`
Call that instead of having duplicate code.

diwako: See latest commit as reference

dedmen: Well if a medical function needs it. Then it seems to make sense to be in medical.

diwako:
Anything else needs changing?

severgun:
Images are so weird.

Issue #5592, opened 2017-10-05
ACE mortar rounds not in arsenal


shukari:
**Arma 3 Version:** stable
**CBA Version:** stable
**ACE3 Version:** stable

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
All this mortar rounds dont show up in the arsenal... so if you add them to a supplybox per mod and then edit the box in eden... the mortar rounds are not shown because the arsenal dont know this rounds...
"ACE_1Rnd_82mm_Mo_HE",
"ACE_1Rnd_82mm_Mo_Smoke",
"ACE_1Rnd_82mm_Mo_Illum",

**Steps to reproduce:**
- get full arsenal
- try endless to get the mortar rounds

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

AACO:
If the magazine config `type` is set to magazine (multiple of 256) it should show up in the arsenal, I'll test it out and try to create a PR for this.

kripto202:
mortar rounds are considered as vehicle ammo

jarckn:
Hey,

whats the status of this issue? It, seems, that the ammo is still not viewable in 3.12.
Tested in arsenal inside a mission and in the virtual arsenal.

Cheers

shukari:
this issue seems to stall

dedmen:
>All this mortar rounds dont show up in the arsenal... so if you add them to a supplybox per mod and then edit the box in eden... the mortar rounds are not shown because the arsenal dont know this rounds

Arsenal doesn't work on boxes.


Is this a ACE Arsenal issue too?

shukari:
Then is is more a feature request, it would be nice to have the mortar rounds in the (ace) arsenal.
ATM the only way to get the mortar rounds is per supplyboxes.

dedmen:
I agree. But quite sure @alganthe already said something about that.

alganthe:
They're type 0, vehicle mags aren't added automatically for obvious reasons.
You can add them manually by doing:
```sqf
["ace_arsenal_displayOpened", {
    (ace_arsenal_virtualItems select 2) append ["ACE_1Rnd_82mm_Mo_HE","ACE_1Rnd_82mm_Mo_Smoke","ACE_1Rnd_82mm_Mo_Illum",
[...]

PR #6694, opened 2018-11-18
Arsenal - Add config to force show (mortars)

PabstMirror:
#5592
@alganthe


alganthe:
👍 forgot we could do that too.
While we're at it, any reason why not do it for the laser guided / illum rounds?

PabstMirror:
There is another mortar PR, I want to get that done first and then I'll come back to this.
Are you ok with the config name, (reuse the `ace_arsenal_hide`)?

alganthe:
yep, it's fine as this is pretty much the same idea as the original intended use of it.

addons/arsenal/functions/fnc_addVirtualItems.sqf
```diff
@@ -193,7 +193,8 @@ if (_items isEqualType true) then {
                     switch true do {
                         // Rifle, handgun, secondary weapons mags
                         case (
-                                (getNumber (configFile >> "CfgMagazines" >> _x >> "type") in [TYPE_MAGAZINE_PRIMARY_AND_THROW,TYPE_MAGAZINE_SECONDARY_AND_PUT,1536,TYPE_MAGAZINE_HANDGUN_AND_GL]) &&
+                                ((getNumber (_x >> "type") in [TYPE_MAGAZINE_PRIMARY_AND_THROW,TYPE_MAGAZINE_SECONDARY_AND_PUT,1536,TYPE_MAGAZINE_HANDGUN_AND_GL]) ||
```
Dystopian: looks like error here: was `configFile >> "CfgMagazines" >> _x >> "type"`, became `_x >> "type"`

alganthe: both work, the full path that is being replaced is not necessary here.

PR #6697, opened 2018-11-19
Fix Armake problems

dedmen:
Don't specify skeletonName if you don't have a skeleton.
Add missing model.cfg entries even if they just refer to the default.

Btw model.cfg files are still mixed tabs/spaces.
I even found tabs and spaces on the same line, with the spaces being 3 space indent.
But I won't fix that in here. I just fixed that one because I noticed it there.


thojkooi:
Nice, thanks for making this PR @dedmen!

jonpas:
No, your additions are using tabs while stuff around it mostly uses spaces.

dedmen:
Builds with 0 warnings/errors now.

thojkooi:
Remaining warnings are all missing prefix for model.

dedmen:
>Remaining warnings are all missing prefix for model.

I broke that then on my end? Thanks. I'll have to fix that.

PR #6704, opened 2018-11-25
Custom Materials for tagging

dedmen:
Allows to use custom materials (rvmat) for tag's.
Adds additional "materials" array to the config.
When there are the same number of materials as textures, it will select a random texture and the corresponding material. If the numbers are different it will select random texture and random material.

Allows for things like this:

arma3_x64_2018-11-25_16-34-57

Glow-in-the-dark tags.

Next question would be if we also want to directly implement a glow-in-the-dark spraycan directly in ACE.

### DIFF
--- a/addons/tagging/functions/fnc_addCustomTag.sqf
+++ b/addons/tagging/functions/fnc_addCustomTag.sqf
@@ -9,6 +9,7 @@
  * 2: Required Item <STRING>
  * 3: Textures Paths <ARRAY>
  * 4: Icon Path <STRING> (default: "")
+ * 5: Material Paths <ARRAY> (optional)
  *
  * Return Value:
  * Sucessfully Added Tag <BOOL>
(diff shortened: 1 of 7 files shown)


dedmen:
ah crap. the model cfg's belong into armake pr.

dedmen:
For some reason the ace action doesn't do anything when I pack unbinarized.
Not even diag logs are printed. No idea wtf is going on there :D

PabstMirror:
problem with `find` is if there are multiple similar textures 
example
textures = ["A", "A", "B"]
mats = ["X", "Y", "Z"]
will never pick combo A/Y

dedmen:
okey yeah didn't think of that. But we still only want to select matching ones if they have the same size.
As you can't select material 5 when there are only 2. And a additional selectRandom as default value might be more overhead

PR #6706, opened 2018-11-26
Javelin - Can disable thermal requirement via code

PabstMirror:


Nemesi-BDR:
It would be very nice if it was possible to disable only the thermal camera requirement while leaving the thermal lock requirement. That's because thermal view is often disabled especially in PvPs to maintain a high level of challenge, but of course it would be nice if the javelin could work realistically as intended by ACE mod.

Nemesi-BDR:
Basically and in short, just leave the "press N" requirement!

PR #6728, opened 2018-12-10
Adapt to armake2

jonpas:
- Update Makefile with nicer definitions
- Fix Makefile not copying additional files when doing local build (eg. extensions)
- Fix an issue with preprocessor define and comment being on same line (not handled in armake2 as of this time)
- Use armake2 in Makefile and on CircleCI (using `circleci/rust` image)
- Make armake CI job longer, longest process is installing armake binary via crates.io, but it's still pretty fast (below 2min)

Warning: Auto-download script is untouched as of right now! @BaerMitUmlaut


jonpas:
Closing in favour of HEMTT.

PR #6734, opened 2018-12-15
markers - add setting allowing only marker creator to move

McDiod:
- add setting to ace_markers for only allowing a marker's creator to move it
- provide local CBA event on new marker placed for external scripts to hook into


McDiod:
Also, shouldn't the moveRestriction setting be global?

commy2:
>Also, shouldn't the moveRestriction setting be global?

I guess it makes no sense to have this setting differently everywhere, but all in all global settings are just restrictions that may limit when you actually want the setting to be different on different machines without providing any real benefit, as the server owner or mission creator could just force the settings.

addons/markers/functions/fnc_isMarkerOwner.sqf
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Public: No
+ */
+
+params [["_marker",""]];
+
+// user has never created a marker
+if (isNil QGVAR(directPlayID)) exitWith {false};
+
+if !((_marker find "_USER_DEFINED") isEqualTo 0) exitWith {false};
```
commy2: This is equivalent to:
`_marker find "_USER_DEFINED" != 0`
I think. Isn't that easier to read as well as faster?

Why 0 and not -1?

addons/markers/functions/fnc_placeMarker.sqf
```diff
@@ -16,15 +16,27 @@
...
 TRACE_2("params",_display,_closeNum);
 
 if (_closeNum isEqualTo 1) then {
     if !(GVAR(editingMarker) isEqualTo "") then {
         //delete "old" marker
         deleteMarker GVAR(editingMarker);
+    } else {
+        [{
+            // provide hook for external scripts
+            private _newestMarker = allMapMarkers select (count allMapMarkers - 1);
```
commy2: Are we sure that only one marker is created? This seems very error prone.

addons/markers/functions/fnc_isMarkerOwner.sqf
McDiod: Some joker might call his editor marker "myMarker_USER_DEFINED", unlikely but possible. I don't mind if you prefer -1 though.

commy2:
I don't like how this retrieves the marker-player-id thingy. What if two markers are created in the same frame? 50% chance to record the wrong id.

addons/markers/functions/fnc_isMarkerOwner.sqf
commy2: I don't get how 0 vs -1 helps in that case.

addons/markers/functions/fnc_placeMarker.sqf
McDiod: No, in fact my first thought was that this might very well return an ACE_BFT_x marker since they are updated so often. On the other hand I have never had any issues while moving markers, which works the same way:

https://github.com/acemod/ACE3/blob/master/addons/markers/functions/fnc_placeMarker.sqf#L31

addons/markers/functions/fnc_isMarkerOwner.sqf
McDiod: 0 vs -1:
```sqf
// not a user defined marker
_marker = createMarker ["myMarker_USER_DEFINED",[0,0,0]];

// true
!((_marker find "_USER_DEFINED") isEqualTo 0)

// false
((_marker find "_USER_DEFINED") isEqualTo -1)
```

As for the `!=` vs. `isEqualTo`: Can be changed. I just used isEqualTo to be in line with
https://github.com/acemod/ACE3/blob/master/addons/markers/functions/fnc_initInsertMarker.sqf#L58

commy2: How are these different from a mission created marker? I.e. `57298579257_USER_DEFINED`?

McDiod:
See above, it seems to work when editing/moving markers. I guess the consequences if it does happen are way worse here though.

Unfortunately I don't see any other way of doing this 🤔

addons/markers/functions/fnc_isMarkerOwner.sqf
McDiod: Mission created markers are `_USER_DEFINED #712781623786/2/5` or something.

addons/markers/functions/fnc_isMarkerOwner.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+private _markerData = (_marker splitString " ") param [1,""];
+private _ownerDirectPlayID = (_markerData splitString "/") param [0,""];
```
dedmen: DirectPlayID? shouldn't this be the playerID aka steamID?

McDiod: See [onPlayerConnected](https://community.bistudio.com/wiki/onPlayerConnected). Apparently this is only used for markers.

PabstMirror:
Instead of the _USER_DEFINED/directPlay ID stuff, what about just saving all our markers in an array?

onPlace: `GVAR(myMarkers) pushBack allMapMarkers select (count allMapMarkers - 1);`
canMove: `if (markerName in GVAR(myMarkers))`

McDiod:
> Instead of the _USER_DEFINED/directPlay ID stuff, what about just saving all our markers in an array?
> onPlace: `GVAR(myMarkers) pushBack allMapMarkers select (count allMapMarkers - 1);`

Same problem that @commy2 brought up. The user placed marker is not necessarily the newest marker.
I don't think there is a way to solve this though. Can be closed if no one else has an idea.

commy2:
Since DirectPlay Id is passed to the OPC event, we could just synch back the players DirectPlay Id at the mission start / JIP and use that.

addons/markers/functions/fnc_isMarkerOwner.sqf
commy2: This does not make any sense to me.
>Why 0 and not -1?
>Some joker might call his editor marker "myMarker_USER_DEFINED", unlikely but possible.

They could just name it `_USER_DEFINED 123`. My question is still not answered.

McDiod:
onPlayerConnected knows the ID as a number. I haven't found a way to convert it to string for the marker name.

>Quite useless as the number is too big for in-built string representation and gets rounded.

commy2:
Ah, so it is a number to big for SQF NUMBER floats. Don't bother then, it is impossible then.

PabstMirror:
https://github.com/acemod/ACE3/blob/master/addons/markers/functions/fnc_placeMarker.sqf#L31

markers has used ` allMapMarkers select (count allMapMarkers - 1)` to pick placed marker and has never had much of a problem
but we can add a safety check and filter based on closest marker:
```
    [{
        params ["_startingMarkers"];
        private _bestMarker = "#";
        private _bestMarkerDist = 1e10;
        private _newMarkers = allMapMarkers - _startingMarkers;
        
        {
            private _distX = GVAR(currentMarkerPosition) distance2d (getMarkerPos _x);
            if (_distX < _bestMarkerDist) then {
                _bestMarker = _x;
                _bestMarkerDist = _distX;
            };
        } forEach _newMarkers;
        
        TRACE_3("",_bestMarker,_bestMarkerDist,_newMarkers);
        
        if (_bestMarker == "#") exitWith {ERROR("could not find user placed marker");};
        GVAR(userPlacedMarkers) pushBack _bestMarker;
        GVAR(userPlacedMarkers) = GVAR(userPlacedMarkers) select {!((getMarkerPos _x) isEqualTo [0,0,0])};
        
        
        [QGVAR(setMarkerNetwork), [
            _bestMarker, [
                GETGVAR(currentMarkerConfigName,""),
                GETGVAR(currentMarkerColorConfigName,""),
                GETGVAR(currentMarkerPosition,[]),
                GETGVAR(currentMarkerAngle,0)
            ]
        ]] call CBA_fnc_globalEvent;

    }, [allMapMarkers]] call CBA_fnc_execNextFrame;
```

PabstMirror:
Made changes in commit https://github.com/acemod/ACE3/commit/a039ce07d3838c7a365d07ba706361c42bba4453

Let me know if this looks ok and your or me can add to this PR

PR #6746, opened 2018-12-27
Changed: Remove ear ringing from flashbang explosions when wearing protective gear (e.g. ComTacs)

bux:
- title
- also reduces ear ringing strength by damageCoefficent if above a certain threshold (not that it matters too much with our current ear ringing implementation)

### DIFF
--- a/addons/grenades/functions/fnc_flashbangExplosionEH.sqf
+++ b/addons/grenades/functions/fnc_flashbangExplosionEH.sqf
@@ -92,8 +92,9 @@ if (hasInterface && {!isNull ACE_player} && {alive ACE_player}) then {
     _strength = _strength * _losCoefficient;
 
     // Add ace_hearing ear ringing sound effect
-    if (isClass (configFile >> "CfgPatches" >> "ACE_Hearing") && {_strength > 0}) then {
-        [40 * _strength] call EFUNC(hearing,earRinging);
+    if (isClass (configFile >> "CfgPatches" >> "ACE_Hearing") && {_strength > 0 && {EGVAR(hearing,damageCoefficent) > 0.25}}) then {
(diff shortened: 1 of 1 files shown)


addons/grenades/functions/fnc_flashbangExplosionEH.sqf
```diff
@@ -92,8 +92,9 @@ if (hasInterface && {!isNull ACE_player} && {alive ACE_player}) then {
     _strength = _strength * _losCoefficient;
 
     // Add ace_hearing ear ringing sound effect
-    if (isClass (configFile >> "CfgPatches" >> "ACE_Hearing") && {_strength > 0}) then {
-        [40 * _strength] call EFUNC(hearing,earRinging);
+    if (isClass (configFile >> "CfgPatches" >> "ACE_Hearing") && {_strength > 0 && {EGVAR(hearing,damageCoefficent) > 0.25}}) then {
+        [40 * _strength * EGVAR(hearing,damageCoefficent)] call EFUNC(hearing,earRinging);
```
jonpas: Split calculation into separate variable for readability and logging without double calculations.

PabstMirror:
https://github.com/acemod/ACE3/blob/master/addons/hearing/functions/fnc_earRinging.sqf
already does
`_strength * GVAR(damageCoefficent`
so now we are effectively doing ` _strength * EGVAR(hearing,damageCoefficent)^2`

bux:
`40 * _strength * EGVAR(hearing,damageCoefficent)^2`
I agree, it looks redundant.

I can change it back. Despiste the one more operation that is performed, it doesn't matter much because  of the way we handle ear ringing
https://github.com/acemod/ACE3/blob/e2ac18a05d5f646bc7cbc043bcc148fde4c0f5dd/addons/hearing/functions/fnc_updateVolume.sqf#L39-L43
It would be better to offer some more different ear ringing sounds/volumes for different strength levels.

PR #6756, opened 2019-01-02
Medical - Add new setting to hide self-interaction actions

### LINKED ISSUE #6735: Medical options on the Windows Keys Menus made Optional. (Toggle able on/off on the Config Menu)
**Arma 3 Version: Stable, latest.
**CBA Version:** Stable, latest.
**ACE3 Version:** Stable, latest.

**Mods:**
- CBA_A3
- ace

**Description:**
- Since the implementation of the Medical Menu (H) the need for both the self interaction medical menus and the interactions with other units' medical options should be made optional.

**Steps to reproduce:**
- Not an issue, but a suggestion. 

- All Scenarios. 

**Additional information:**
- If you check out most of the streamers videos in twitch/youtube etc you can confirm that a great percentage of them just use the Medical Menu (H) since it is faster and easier to understand and also allows for both self medical attention as well as giving medical attention to others.

Because of that I want to suggest to make the self interact and regular menu (windows keys) medical options... optional. Just like Group Management and Gestures can also be disabled from the Ace Config Menu.

That way we will have less clutter and we can add more actions to both the windows keys menus.

Thanks and keep up the amazing work!

PabstMirror:
Fix #6735

### DIFF
--- a/addons/medical/ACE_Medical_SelfActions.hpp
+++ b/addons/medical/ACE_Medical_SelfActions.hpp
@@ -3,7 +3,7 @@ class Medical {
     runOnHover = 1;
     exceptions[] = {"isNotInside", "isNotSwimming", "isNotSitting"};
     statement = QUOTE([ARR_3(_target, true, 0)] call DFUNC(displayPatientInformation));
-    condition = "true";
+    condition = QUOTE(GVAR(menuTypeStyle) != 2);
     icon = QPATHTOF(UI\icons\medical_cross.paa);
 
    class ACE_Head {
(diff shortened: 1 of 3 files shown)


jonpas:
> also "disabled" doesn't make sense in this case, maybe change that to "medical menu only" ?

Medical menu is a separate setting, you can run both disabled (I think only keybind for medical menu will work then, if even). Disabled is fine in this case, it disables Medical Interactions.

alganthe:
yes but the setting is ambiguous.

PR #6768, opened 2019-01-05
Fix # command in sqf_validator.py

neilzar:
- Fix the sqf validator to correctly detect the use of # as select or preprocessor command.


jonpas:
Same as https://github.com/CBATeam/CBA_A3/pull/1043

PR #6770, opened 2019-01-07
Dragging - Fix player getting locked with isCarrying

PabstMirror:
Player could get bugged with `player getVariable "ace_dragging_isCarrying"` set to true
(then cannot use ace interaction menu)

Reproduce:
Try to carry unconscious X
During pickup anim, X gets shot and dies

FUNC(dropObject) sets QGVAR(isDragging)
FUNC(dropObject_carry) sets QGVAR(isCarrying)


Drofseh:
I think this has happened to me before.
I had to interact with the dead body and "put it down" before I could interact with other things again.

PR #6780, opened 2019-01-20
Additional weight override parameter for setCarryable/setDraggable

PiZZAD0X:
- Add an additional optional parameter to the `ace_dragging_fnc_setCarryable` and `ace_dragging_fnc_setDraggable` functions that sets a object variable that can skip the weight check in `ace_dragging_fnc_startCarry`  and `ace_dragging_fnc_startCarry` respectively. 

- Existing function parameter format is `[object, true, [0,1,1], 0] call ace_dragging_fnc_setCarryable;` which is compatible with this PR.

eg. `[object, true, [0,1,1], 0, true] call ace_dragging_fnc_setCarryable;`
would override the carry weight check on the object.

- This would be useful if mission makers want the ability to carry/drag a specific heavy object without disabling/nullifying the weight checks globally, similar to how mission makers could use `setMass` in the past.

-  Addresses https://github.com/acemod/ACE3/issues/4141

### DIFF
--- a/addons/dragging/functions/fnc_canCarry.sqf
+++ b/addons/dragging/functions/fnc_canCarry.sqf
@@ -25,6 +25,6 @@ if !([_unit, _target, []] call EFUNC(common,canInteractWith)) exitWith {false};
 if ((_unit getHitPointDamage "HitLegs") >= 0.5) exitWith {false};
 
 // a static weapon has to be empty for dragging (ignore UAV AI)
-if (((typeOf _target) isKindOf "StaticWeapon") && {!({!((getText (configFile >> "CfgVehicles" >> (typeOf _x) >> "simulation")) isEqualto "UAVPilot")} count crew _target isEqualto -1)}) exitWith {false};
+if (((typeOf _target) isKindOf "StaticWeapon") && {!({!((getText (configFile >> "CfgVehicles" >> (typeOf _x) >> "simulation")) isEqualto "UAVPilot")} count crew _target isEqualto 0)}) exitWith {false};
 
 alive _target && {vehicle _target isEqualto _target} && {_target getVariable [QGVAR(canCarry), false]} && {animationState _target in ["", "unconscious"] || (_target getVariable ["ACE_isUnconscious", false]) || (_target isKindOf "CAManBase" && {(_target getHitPointDamage "HitLegs") > 0.4})}
--- a/addons/dragging/functions/fnc_canCarry.sqf
+++ b/addons/dragging/functions/fnc_canCarry.sqf
@@ -25,6 +25,6 @@ if !([_unit, _target, []] call EFUNC(common,canInteractWith)) exitWith {false};
 if ((_unit getHitPointDamage "HitLegs") >= 0.5) exitWith {false};
 
 // a static weapon has to be empty for dragging (ignore UAV AI)
-if (((typeOf _target) isKindOf "StaticWeapon") && {{(getText (configFile >> "CfgVehicles" >> (typeOf _x) >> "simulation")) != "UAVPilot"} count crew _target > 0}) exitWith {false};
+if (((typeOf _target) isKindOf "StaticWeapon") && {!({!((getText (configFile >> "CfgVehicles" >> (typeOf _x) >> "simulation")) isEqualto "UAVPilot")} count crew _target isEqualto -1)}) exitWith {false};
 
-alive _target && {vehicle _target == _target} && {_target getVariable [QGVAR(canCarry), false]} && {animationState _target in ["", "unconscious"] || (_target getVariable ["ACE_isUnconscious", false]) || (_target isKindOf "CAManBase" && {(_target getHitPointDamage "HitLegs") > 0.4})}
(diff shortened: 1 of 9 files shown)


addons/dragging/functions/fnc_canDrag.sqf
```diff
@@ -21,6 +21,6 @@ params ["_unit", "_target"];
 if !([_unit, _target, ["isNotSwimming"]] call EFUNC(common,canInteractWith)) exitWith {false};
 
 // a static weapon has to be empty for dragging (ignore UAV AI)
-if ((typeOf _target) isKindOf "StaticWeapon" && {{(getText (configFile >> "CfgVehicles" >> (typeOf _x) >> "simulation")) != "UAVPilot"} count crew _target > 0}) exitWith {false};
+if ((typeOf _target) isKindOf "StaticWeapon" && {!({!((getText (configFile >> "CfgVehicles" >> (typeOf _x) >> "simulation")) isEqualto "UAVPilot")} count crew _target isEqualto 0)}) exitWith {false};
```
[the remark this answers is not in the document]
PiZZAD0X: Interesting, I was under the impression not/! was a binary modifier. The case sensitivity addition is not worth it.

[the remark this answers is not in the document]
commy2: `!` is just another SQF command that reports `true` if input was `false` and vice versa.

PiZZAD0X:
for ```if (!GETVAR(_target,GVAR(ignoreWeightCarry),false) && {
    private _weight = [_target] call FUNC(getWeight);
    _weight > (GETMVAR(ACE_maxWeightCarry,1E11))
    })```

Should I skip the variable and do ```(!GETVAR(_target,GVAR(ignoreWeightCarry),false) && {
    ([_target] call FUNC(getWeight)) > (GETMVAR(ACE_maxWeightCarry,1E11))
    })``` or keep the var?

commy2:
Either is fine I think, but I prefer more locals (with descriptive names) instead of long lines that are hard to read.

addons/dragging/functions/fnc_canDrag.sqf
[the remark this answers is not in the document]
PiZZAD0X: Hmm so the scope in which one can see performance improvements from using isEqualTo are more limited than one would expect from a `===` to `==` understanding in other languages. Good to know.

addons/dragging/functions/fnc_startCarry.sqf
```diff
@@ -19,10 +19,12 @@
...
 
-// check weight
-private _weight = [_target] call FUNC(getWeight);
-
-if (_weight > missionNamespace getVariable ["ACE_maxWeightCarry", 1E11]) exitWith {
+// exempt from weight check if object has override variable set
+if (!GETVAR(_target,GVAR(ignoreWeightCarry),false) && {
+    private _weight = ([_target] call FUNC(getWeight));
+    _weight > (GETMVAR(ACE_maxWeightCarry,1E11))
+    }) exitWith {
```
commy2:
```suggestion
    private _weight = [_target] call FUNC(getWeight);
    _weight > GETMVAR(ACE_maxWeightCarry,1E11)
}) exitWith {
```
formatting. Also testing GitHub `suggestion`.

PiZZAD0X: Hmm, replaced only 1 line with multiple

commy2: Yeah, apparently you can only replace 1 line as of now. :(

addons/dragging/functions/fnc_startDrag.sqf
```diff
@@ -19,16 +19,18 @@
...
 TRACE_2("params",_unit,_target);
 
-// check weight
-private _weight = [_target] call FUNC(getWeight);
-
-if (_weight > missionNamespace getVariable ["ACE_maxWeightDrag", 1E11]) exitWith {
+// exempt from weight check if object has override variable set
+if (!GETVAR(_target,GVAR(ignoreWeightDrag),false) && {
+    private _weight = [_target] call FUNC(getWeight);
+    _weight > (GETMVAR(ACE_maxWeightDrag,1E11))
```
commy2:
```suggestion
    _weight > GETMVAR(ACE_maxWeightDrag,1E11)
```

addons/dragging/functions/fnc_canCarry.sqf
```diff
@@ -27,4 +27,4 @@ if ((_unit getHitPointDamage "HitLegs") >= 0.5) exitWith {false};
 // a static weapon has to be empty for dragging (ignore UAV AI)
 if (((typeOf _target) isKindOf "StaticWeapon") && {{(getText (configFile >> "CfgVehicles" >> (typeOf _x) >> "simulation")) != "UAVPilot"} count crew _target > 0}) exitWith {false};
 
-alive _target && {vehicle _target == _target} && {_target getVariable [QGVAR(canCarry), false]} && {animationState _target in ["", "unconscious"] || (_target getVariable ["ACE_isUnconscious", false]) || (_target isKindOf "CAManBase" && {(_target getHitPointDamage "HitLegs") > 0.4})}
+alive _target && {vehicle _target isEqualto _target} && {_target getVariable [QGVAR(canCarry), false]} && {animationState _target in ["", "unconscious"] || (_target getVariable ["ACE_isUnconscious", false]) || (_target isKindOf "CAManBase" && {(_target getHitPointDamage "HitLegs") > 0.4})}
```
dedmen: `isEqualTo` uppercase `To`.

PR #6788, opened 2019-01-23
Common - optimize toHex

commy2:
- Old one takes about 13ms each mission start (postInit)
- this should be many times faster and looks simpler to me
- untested


PabstMirror:
```
GVAR(hexArray) = [
"00","01","02","03","04","05","06","07","08","09","0A","0B","0C","0D","0E","0F",
"10","11","12","13","14","15","16","17","18","19","1A","1B","1C","1D","1E","1F",
"20","21","22","23","24","25","26","27","28","29","2A","2B","2C","2D","2E","2F",
"30","31","32","33","34","35","36","37","38","39","3A","3B","3C","3D","3E","3F",
"40","41","42","43","44","45","46","47","48","49","4A","4B","4C","4D","4E","4F",
"50","51","52","53","54","55","56","57","58","59","5A","5B","5C","5D","5E","5F",
"60","61","62","63","64","65","66","67","68","69","6A","6B","6C","6D","6E","6F",
"70","71","72","73","74","75","76","77","78","79","7A","7B","7C","7D","7E","7F",
"80","81","82","83","84","85","86","87","88","89","8A","8B","8C","8D","8E","8F",
"90","91","92","93","94","95","96","97","98","99","9A","9B","9C","9D","9E","9F",
"A0","A1","A2","A3","A4","A5","A6","A7","A8","A9","AA","AB","AC","AD","AE","AF",
"B0","B1","B2","B3","B4","B5","B6","B7","B8","B9","BA","BB","BC","BD","BE","BF",
"C0","C1","C2","C3","C4","C5","C6","C7","C8","C9","CA","CB","CC","CD","CE","CF",
"D0","D1","D2","D3","D4","D5","D6","D7","D8","D9","DA","DB","DC","DD","DE","DF",
"E0","E1","E2","E3","E4","E5","E6","E7","E8","E9","EA","EB","EC","ED","EE","EF",
"F0","F1","F2","F3","F4","F5","F6","F7","F8","F9","FA","FB","FC","FD","FE","FF"
];
```

dedmen:
@PabstMirror your method would be 690 bytes longer!!!111!!11!
We need more benchmarks. Tho commy's new code was measure to be in the sub half millisecond region (probably on a Intel CPU) and it looks soo elegant :D

addons/common/functions/fnc_toHex.sqf
```diff
@@ -22,33 +22,18 @@ _number = ((round abs _number) max 0) min 255;
...
+    {
+        private _x1 = _x;
+
+        {
+            GVAR(hexArray) pushBack (_x1 + _x);
+        } forEach _digits;
+    } forEach _digits;
 };
 
 GVAR(hexArray) select _number // return
```
SilentSpike: `// return`

Quality comment here

commy2: I've been using this for forever. Whenever the last statement's return value is the return value of the whole function and it is not something trivial like a single variable, I write // return in addition to leaving out the semi colon. This makes it clear that when someone happens to edit the function for whatever reason, it is not good to add lines below this one. It is kind of obvious here, but not necessarily so. E.g. for functions that do stuff as well as returning something instead of being a pure getter like this one.

PabstMirror:
Thoughts on which way we want?
I say put the LUT in preInit and we can also remove the `if (isNil QGVAR(hexArray)) then {` check

commy2:
By "LUT" you mean the big array in your previous comment? I am fine with that.

PR #6789, opened 2019-01-24
Fixed Typo from change to chance

chaot3ch:
This change will correct the typo when describing bandage types.


jonpas:
This change will change change to chance!

chaot3ch:
> This change will change change to chance!

Gotta start somewhere :D

PR #6790, opened 2019-01-24
Copy inherited compiled self actions instead of recompiling

dedmen:
Previously it would compile the selfActions 11 times (for 11 different CAManBase types)
even though they all have the exact same config actions, all inherited from CAManBase.

This PR makes it skip compiling actions if the class itself has no unique self actinons and will instead just copy the parent (if the parent doesn't have actions either, it will already be a copy of it's parent)

The problem I see is that if someone add's scripted actions to CAManBase with inherit disabled.
If the sub class is compiled later it will copy these scripted actions that were added to CAManBase, and they will be inherited even though inherit parameter (addActionToClass) was false.

Before: 
The 624 number is 624 calls to `compile`. They are compiling the exact same stuff 11 times.
After: 

Traces if you wanna look at the data yourself.

BeforeFix trace is coming in an hour.

[AfterFix_ASP_Tracy.zip](https://github.com/acemod/ACE3/files/2792379/AfterFix_ASP_Tracy.zip)

### DIFF
--- a/addons/interact_menu/functions/fnc_compileMenuSelfAction.sqf
+++ b/addons/interact_menu/functions/fnc_compileMenuSelfAction.sqf
@@ -22,9 +22,21 @@ if (_target isEqualType objNull) then {
     _objectType = typeOf _target;
 };
 private _namespace = GVAR(ActSelfNamespace);
-
+diag_log _objectType;
 // Exit if the action menu is already compiled for this class
 if !(isNil {_namespace getVariable _objectType}) exitWith {};
+
+if (("configName _x == 'ACE_SelfActions'" configClasses (configFile >> "CfgVehicles" >> _objectType)) isEqualTo []) exitWith {
+    //This class doesn't have self actions, it just inherits them. So we don't have to recompile and can just copy from parent
+
+    private _parentType = configName inheritsFrom (configFile >> "CfgVehicles" >> _objectType);
+
+    //Make sure the parent is compiled
+    [_parentType] call FUNC(compileMenuSelfAction);
+
+    //Copy classes from parent
+    _namespace setVariable [_objectType, _namespace getVariable _parentType];
+};
 
 
 private _recurseFnc = {
--- a/addons/interact_menu/functions/fnc_compileMenuSelfAction.sqf
+++ b/addons/interact_menu/functions/fnc_compileMenuSelfAction.sqf
@@ -22,7 +22,7 @@ if (_target isEqualType objNull) then {
     _objectType = typeOf _target;
 };
 private _namespace = GVAR(ActSelfNamespace);
-diag_log _objectType;
+
 // Exit if the action menu is already compiled for this class
 if !(isNil {_namespace getVariable _objectType}) exitWith {};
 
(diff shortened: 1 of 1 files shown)


addons/interact_menu/functions/fnc_compileMenuSelfAction.sqf
```diff
@@ -22,10 +22,22 @@ if (_target isEqualType objNull) then {
     _objectType = typeOf _target;
 };
 private _namespace = GVAR(ActSelfNamespace);
-
+diag_log _objectType;
```
bux: needed?

dedmen: oops :D

dedmen:
I created a stresstest. I placed down all NATO unit types in vanilla in a mission and started it.

zeus postInit
Before: 1060ms
After: 58.52ms

Before on top:


Traces:

[profilerRecordingAceAddActionToClass.zip](https://github.com/acemod/ACE3/files/2792797/profilerRecordingAceAddActionToClass.zip)

PabstMirror:
With module placed it goes into unrecoverable loop
```
[ACE] (interact_menu) TRACE: 1191 : _objectType=ModuleCurator_F z\ace\addons\interact_menu\functions\fnc_compileMenuSelfAction.sqf:24
[ACE] (interact_menu) TRACE: 1191 : _objectType=Module_F z\ace\addons\interact_menu\functions\fnc_compileMenuSelfAction.sqf:24
[ACE] (interact_menu) TRACE: 1191 : _objectType=Logic z\ace\addons\interact_menu\functions\fnc_compileMenuSelfAction.sqf:24
[ACE] (interact_menu) TRACE: 1191 : _objectType=All z\ace\addons\interact_menu\functions\fnc_compileMenuSelfAction.sqf:24
[ACE] (interact_menu) TRACE: 1191 : _objectType= z\ace\addons\interact_menu\functions\fnc_compileMenuSelfAction.sqf:24
[ACE] (interact_menu) TRACE: 1191 : _objectType= z\ace\addons\interact_menu\functions\fnc_compileMenuSelfAction.sqf:24
[ACE] (interact_menu) TRACE: 1191 : _objectType= z\ace\addons\interact_menu\functions\fnc_compileMenuSelfAction.sqf:24
```
I think we can fix endless recursion with
```
    private _parentActions = if (_parentType == "") then {
        []
    } else {
        [_parentType] call FUNC(compileMenuSelfAction);
        +(_namespace getVariable _parentType)
    };
```
but another problem is duplicate actions showing on inheritance
e.g.
```
private _act = ["test","Redfor Action","",{systemChat "x"},{true}] call ace_interact_menu_fnc_createAction;
["O_Soldier_F", 1, ["ACE_SelfActions"], _act, true] call ace_interact_menu_fnc_addActionToClass; // inhertiable action
```
then place a red rifleman and then a red unarmed rifleman
because ` class O_Soldier_unarmed_F: O_Soldier_F {`
he will get double actions (depending on init order)

dedmen:
I got this with the current ace build WITHOUT my change.
```
  Error position: <setVariable [_objectType, _actions];
>
  Error Reserved variable in expression
```
who the hell calls that func with empty string :U

So yeah we definitely need empty string check. Or maybe just isNull check on the class?

PabstMirror:
ref https://github.com/acemod/ACE3/pull/1294

PabstMirror:
Close?

dedmen:
I guess so.. If there are still performance issues then it would most likely be a different fix anyway so I'd just make a new PR. If not I can just reopen this one and update it I guess.

PR #6791, opened 2019-01-25
Move SelfActions from postInit to config

dedmen:
Also swapped around the condition on the create, it would do the expensive switch before doing the cheap isNil.


addons/zeus/CfgVehicles.hpp
```diff
@@ -317,4 +317,35 @@ class CfgVehicles {
...
+
+    class Man;
+    class CAManBase: Man {
+        class ACE_SelfActions {
+            class GVAR(create) {
+                displayName = CSTRING(CreateZeus);
+                condition = QUOTE(\
+                isNil QQGVAR(zeus) && {\
+                    switch (GVAR(canCreateZeus)) do {\
+                        case CAN_CREATE_ADMIN:   {isServer || {IS_ADMIN_LOGGED}};\
```
dedmen: `IS_ADMIN_LOGGED` is not config compatible as it has `"#shutdown"` inside it. Fix incoming

dedmen: Fixed in https://github.com/CBATeam/CBA_A3/pull/1048/files
Will apply that in my PR

Dystopian:
I object.
- postInit actions are much more flexible than config one. Config actions testing requires game restart (release, not dev) while postInit wants just mission restart;
- config multiline code looks quite ugly and moving it to separate functions is overkill.

We save just some hundreds ms at mission start but I don't think it's really worth it. That's why I added both actions with postInit.

Maybe `addActionToClass` should be modified to speed up mission start instead. E.g. set some namespace variables to init actions on first using instead `addClassEventHandler` call.

dedmen:
>postInit actions are much more flexible than config one. Config actions testing requires game restart (release, not dev) while postInit wants just mission restart;

Why does this action have to be flexible? Has nothing in there that one would want to change.

Dystopian:
> Why does this action have to be flexible? Has nothing in there that one would want to change.

Any code can be improved. And it's easier to improve it if testing is simple.

dedmen:
Then improve it and put it into config when it's improved.

addons/zeus/CfgVehicles.hpp
```diff
@@ -317,4 +317,35 @@ class CfgVehicles {
     class ModuleArsenal_F: Module_F {
         function=QFUNC(bi_moduleArsenal);
     };
+
+    class Man;
+    class CAManBase: Man {
+        class ACE_SelfActions {
+            class GVAR(create) {
+                displayName = CSTRING(CreateZeus);
+                condition = QUOTE(\
```
PabstMirror:
```
14:09:37 File z\ace\addons\zeus\CfgVehicles.hpp, line 325: '/CfgVehicles/CAManBase/ACE_SelfActions/ace_zeus_create.condition': Missing ';' at the end of line
14:09:37 ErrorMessage: File z\ace\addons\zeus\CfgVehicles.hpp, line 325: /CfgVehicles/CAManBase/ACE_SelfActions/ace_zeus_create/: Wrong # directive
14:09:37 Application terminated intentionally
ErrorMessage: File z\ace\addons\zeus\CfgVehicles.hpp, line 325: /CfgVehicles/CAManBase/ACE_SelfActions/ace_zeus_create/: Wrong # directive
```
I think complex code like this should just be in a function, I hate dealing with ARR_X and string escaping

dedmen: Is that mikeros tools? Might be a bug in his parser ^^
Yeah I'll move it.. Name?
zeus_fnc_actionCreateCondition ?

jonpas: It's a condition, so `zeus_fnc_canCreate` (something with `canX`).

PabstMirror: it was crashing arma on game start
the way it's now handled in master is good enough imho (https://github.com/acemod/ACE3/blob/master/addons/zeus/XEH_postInit.sqf#L85-L121)
otherwise `zeus_fnc_actionCreateCondition`/`zeus_fnc_actionCreateStatement` isn't bad because they are sorted alphabetically

dedmen: The code in current master will still execute code when you spawn units in Zeus right? Even if just once per type. Whereas the config one wouldn't even do that?
Also it seems like with the script code we'll be adding it to tons of CAManBase based classes, even though we really only need CAManBase once.

I checked the config code using the make script, Don't remember whether mikero or armake.

addons/zeus/functions/fnc_canCreateModule.sqf
```diff
@@ -0,0 +1,26 @@
...
+
+isNil QQGVAR(zeus) && 
+{
+	switch (GVAR(canCreateZeus)) do {
+		case CAN_CREATE_ADMIN:   {isServer || {IS_ADMIN_LOGGED}};
+		case CAN_CREATE_CONSOLE: {call BIS_fnc_isDebugConsoleAllowed};
+		case CAN_CREATE_ALL:     {true};
+		default {false};
+	}
+}
```
commy2:
```suggestion
}

```

addons/zeus/functions/fnc_canCreateModule.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Return Value:
+ * Allowed to create module <BOOL>
+ *
+ * Example:
+ * call ace_zeus_fnc_canCreateModule
+ *
+ * Public: No
+ */
+
+isNil QQGVAR(zeus) &&
```
commy2:
```suggestion
isNil QQGVAR(zeus) && {
```

addons/zeus/functions/fnc_canCreateModule.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Allowed to create module <BOOL>
+ *
+ * Example:
+ * call ace_zeus_fnc_canCreateModule
+ *
+ * Public: No
+ */
+
+isNil QQGVAR(zeus) && 
+{
```
commy2:
```suggestion
```

addons/zeus/functions/fnc_canCreateModule.sqf
dedmen: That was what I had at first but I thought readability is better with it on seperate line.

addons/zeus/CfgVehicles.hpp
PabstMirror: no code happens on object init
only fires once each time the player controls a new man/vehicle type

include/x/cba/addons/main/script_macros_common.hpp
```diff
@@ -478,8 +478,8 @@ Group: General
 
 // *************************************
 // Internal Functions
-#define DOUBLES(var1,var2) var1##_##var2
-#define TRIPLES(var1,var2,var3) var1##_##var2##_##var3
+#define DOUBLES(var1,var2) ##var1##_##var2
```
jonpas: This was changed in ACE3 (and branch on CBA) for armake, as leading concatenation is pointless.

dedmen: Should get these fixes merged into CBA, instead of waiting for full armake merge on CBA.

jonpas: Yes, indeed.

PR #6792, opened 2019-01-25
WIP: Interaction Menu - Only compile self actions when needed

PabstMirror:
Replace #6790 #6791

best of both worlds
adds new event `QEGVAR(interact_menu,newControlableObject)` that will run once for each unique thing a player controls, so we can add actions via code with no need to compile everything

### DIFF
--- a/addons/zeus/XEH_postInit.sqf
+++ b/addons/zeus/XEH_postInit.sqf
@@ -82,7 +82,12 @@ if (hasInterface) then {
         [localize "str_a3_cfgvehicles_moduletasksetstate_f_arguments_state_values_created_0"] call EFUNC(common,displayTextStructured);
     }] call CBA_fnc_addEventHandler;
 
-    private _action = [
+    [QEGVAR(interact_menu,newControlableObject), {
+        params ["_type"];
+        if (!(_type isKindOf "CaManBase")) exitWith {};
+        TRACE_1("Adding zeus actions",_type);
+        
+        private _action = [
         QGVAR(create),
         LLSTRING(CreateZeus),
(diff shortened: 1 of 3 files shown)


jonpas:
`newControlableObject` should probably be `newControllableObject`?

addons/interact_menu/XEH_preInit.sqf
```diff
@@ -8,6 +8,10 @@ PREP_RECOMPILE_END;
 
 #include "initSettings.sqf"
 
+if (!hasInterface) exitWith { ADDON = true; };
```
dedmen: Why did noone think of this before. This is so trivial and logical..

addons/zeus/XEH_postInit.sqf
```diff
@@ -82,7 +82,12 @@ if (hasInterface) then {
         [localize "str_a3_cfgvehicles_moduletasksetstate_f_arguments_state_values_created_0"] call EFUNC(common,displayTextStructured);
     }] call CBA_fnc_addEventHandler;
 
-    private _action = [
+    [QEGVAR(interact_menu,newControllableObject), {
+        params ["_type"];
+        if (!(_type isKindOf "CaManBase")) exitWith {};
```
dedmen:
```suggestion
        if (!(_type isKindOf "CAManBase")) exitWith {};
```

addons/interact_menu/XEH_preInit.sqf
jonpas: Yeah we have.

Dystopian: > Why did noone think of this before. This is so trivial and logical..

maybe because it didn't really hurt?

PR #6773, opened 2019-01-14
Implement M47 Dragon

TheCandianVendingMachine:
**When merged this pull request will:**
- Port NouberNou's M47 Dragon to the Missile Guidance framework as well as the upcoming CSW framework
- Fix M47 Dragon memory points
- Change the optic model so the stadia lines actually measure the range of the vehicle
- Change the optic zoom to the real life 6 degree 6x FOV
- Change the Missile Guidance framework to disable the flaps feature under certain circumstances
- Change the Missile Guidance framework so that seekers as well as attack profiles can have their own `onFired` functions
- Update the HOT missile to update to the changes above
- Add a `feature` Wiki page

The Dragon uses the HOT Missile SACLOS guidance and implements a custom attack profile. While the Dragon is in flight is uses a set of charges alongside the edge of the missile to guide it in flight. This makes the missile inherently inaccurate and obvious as to when the missile is approaching. The max range of the Dragon implemented is 1.5 kilometers under the "Super Dragon" variant

Showcasing the M47 against a CUP T-55
https://youtu.be/Dj4U4ojJiU8

The whole system in action (including CSW building)
https://youtu.be/74zMfotVue0

Requires #5652 
Overwrites #3701

PR #6796, opened 2019-01-31
cfgSkeletons indentation formatting fixes

TheCandianVendingMachine:
- Change all instances of `class OFP2_ManSkeleton` to correct indentation of bone layout


addons/tacticalladder/data/model.cfg
```diff
@@ -26,125 +26,118 @@ class CfgSkeletons {
...
+                                                "Face_EyelidUpperRight","Face_Hub",
+                                                "Face_EyelidUpperLeft","Face_Hub",
+                                                "Face_EyelidLowerRight","Face_Hub",
+                                                "Face_EyelidLowerLeft","Face_Hub",
+                                                "EyeLeft","Face_Hub",
+                                                "EyeRight","Face_Hub"
+        };// end of skeleton array
         // location of pivot points (local axes) for hierarchical animation
-        pivotsModel="A3\anims_f\data\skeleton\SkeletonPivots.p3d";
+        pivotsModel="Samples_F\TemplateRTM\ManSkeleton_Pivots.p3d";
```
dedmen: Changing pivots model. The old one was correct. Samples_F doesn't exist in normal play

dedmen: That is also wrong in the fastroping one and https://github.com/acemod/ACE3/pull/6773

Though afaik that entry isn't even used in-game, it's only needed for creation of rtm animations.

jonpas:
We could symlink them, git supports that.

dedmen:
>But there are small differences in each eh?

shouldn't. The pivot;Model is different, but that's because some use the samples one which will not be loaded in release version Arma. We should use the path that's always there.

jokoho48:
that is wrong @dedmen they should be different in some places where models or Animations have a different behavior in that file or a model is referenced in the model.cfg

addons/tacticalladder/data/model.cfg
```diff
@@ -26,125 +26,118 @@ class CfgSkeletons {
     class OFP2_ManSkeleton {
         isDiscrete = 0;
         skeletonInherit = "";
-        skeletonBones[] = {
+        skeletonBones[]= {
```
dedmen: whitespace before `=` should proably stay

dedmen:
>they should be different in some places where models or Animations have different behavior in that file or a model is referenced in the model.cfg

Which is not the case here.
Or if that should be the case, it isn't. So I guess it doesn't matter.

One difference I just saw is that gunbag has some names lowercase that are uppercase in Arma 3 samples. Are bone names case sensitive or not? Cuz if they are then we have a problem, either the gunbag model.cfg is wrong, or Arma 3 samples is wrong.

addons/gunbag/data/model.cfg
```diff
@@ -7,119 +7,118 @@ class CfgSkeletons {
...
+                                                "Face_Eyelids","Face_Hub",
+                                                "Face_EyelidUpperRight","Face_Hub",
+                                                "Face_EyelidUpperLeft","Face_Hub",
+                                                "Face_EyelidLowerRight","Face_Hub",
+                                                "Face_EyelidLowerLeft","Face_Hub",
+                                                "EyeLeft","Face_Hub",
+                                                "EyeRight","Face_Hub"
+        };// end of skeleton array
         // location of pivot points (local axes) for hierarchical animation
-        pivotsModel="A3\anims_f\data\skeleton\SkeletonPivots.p3d";
```
dedmen: Should stay the `A3\anims_f` path, for ever instance of the OFP2_ManSkeleton that we have

dedmen:
@PabstMirror what do you think about the idea of putting the OFP skeleton into a single header file in ace common and just including that in the model.cfg? Would create a PR for that.

PR #6797, opened 2019-02-01
Wiki - Add documentation for ace_interact_menu_newControllableObject

PabstMirror:


654wak654:
Maybe add a `(Added in ACE3 3.13.0)` note? Since this feature is about performance, people might rush to use it soon as they see the PRs :sweat_smile:

PR #6799, opened 2019-02-08
Compile minedetector scripts at preStart

dedmen:
Move 100ms of preInit code to preStart

tracy_2019-02-08_14-55-57


commy2:
Why is this not already in wtf.

PR #6800, opened 2019-02-08
Common - checkfiles use cba's list of CfgPatches

PabstMirror:
for full modset
`count cba_common_addons = 3275`

most of the cost is in `configClasses`

```
44.7826 ms
private _addons = "true" configClasses (configFile >> "CfgPatches");
_addons = _addons apply {toLower configName _x};
z = _addons select {_x find "ace_" == 0};

4.5045 ms
z = (cba_common_addons select {(_x select [0,4]) == "ace_"}) apply {toLower _x};
```


dedmen:
More speed for my beloved postInit

PR #6801, opened 2019-02-09
add 100Rnd mags magwell to all MX variants

commy2:
- keeps 100Rnd mags magwell for all MX variants after this CBA PR: https://github.com/CBATeam/CBA_A3/pull/1058


ZeusWrath:
Hello ACE! In the 3.12.6 you added 100Rnd Maggs to all MX rifles, but after reloading this magazine outwardly looks like a standard 30Rnd magazine. This is normal or bug?

dedmen:
@ZeusWrath BI's fault if it really does.

ZeusWrath:
@dedmen thanks for the answer. Try to fix it, it looks not good/

PR #6804, opened 2019-02-09
Tagging - Cache tagging static models

dedmen:
Picture > words

### DIFF
--- a/addons/tagging/XEH_preStart.sqf
+++ b/addons/tagging/XEH_preStart.sqf
@@ -1,3 +1,31 @@
 #include "script_component.hpp"
 
 #include "XEH_PREP.hpp"
+
+
+private _cacheStaticModels = [];
+
+private _vehicleClasses = "isClass _x && (configName _cfgClass) isKindOf 'Static'" configClasses (configFile >> "CfgVehicles");
+
+{
+    private _model = getText (_x >> "model");
+    if (_model != "") then {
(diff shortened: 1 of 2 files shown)


addons/tagging/XEH_preStart.sqf
```diff
@@ -1,3 +1,34 @@
...
+// This include houses (which we don't need), but also walls, that we do
+{
+    private _model = getText (_x >> "model");
+    if (_model != "") then {
+        private _array = _model splitString "\";
+        _cacheStaticModels pushBackUnique (_array select ((count _array) - 1));
+    };
+} forEach _vehicleClasses;
+
+
```
commy2:
```suggestion
```

addons/tagging/XEH_preStart.sqf
```diff
@@ -1,3 +1,34 @@
 #include "script_component.hpp"
 
 #include "XEH_PREP.hpp"
+
+
```
commy2:
```suggestion
```

PabstMirror:
With CUP and RHS:
`count  call (uiNamespace getVariable ["ace_tagging_cacheStaticModels", {[]}]);` = 6927
`count allVariables ace_tagging_cacheStaticModels` = 5758

We're using `pushBackUnique`, but I guess there must be different capitalizations

PR #6809, opened 2019-02-12
Fix Medical HandleDamageWounds memory leak

dedmen:
Do what title says.
`compile` command caches string indefinitely. Meaning every combination of wounds and pain will stay in memory forever till the game exits or crashes.


dedmen:
Btw while we're at it.
https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_handleDamage_wounds.sqf#L29 what is up with _woundID? It's never changed, but 
https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_handleDamage_wounds.sqf#L62 set onto the unit and causing network traffic.
Actually that variable is never used. It's only initialized to 1 in medical init and then never used again..

PabstMirror:
IIRC, we've talked about dropping unique wound ids

SilentSpike:
> IIRC, we've talked about dropping unique wound ids

Correct, as they would have been used only to do individually targeted treatment which I think we decided is unnecessarily detailed.

dedmen:
>IIRC, we've talked about dropping unique wound ids

Makes sense, as there currently doesn't seem to be any use nor sense for it. Unless the extension isn't there, which only can happen on linux dedicated server.

AUTHORS.txt
```diff
@@ -40,7 +40,9 @@ Aleksey EpMAK Yermakov <epmak777@gmail.com>
 Alganthe <alganthe@live.fr>
 Andrea "AtixNeon" Verano <veranoandrea88@gmail.com>
 Anthariel <Contact@storm-simulation.com>
+Arcanum417 <lubos.len@gmail.com>
 Anton
+Arcanum417 <lubos.len@gmail.com>
```
dedmen: Why twice?

PR #6810, opened 2019-02-12
WIP: Get rid of call compile's

dedmen:
The fnc_bi_moduleCurator change might be problematic as parseSimpleArray doesn't allow spaces after comma's. But users might try to use spaces in the textbox in the module.
Same for ace_map_gestures_fnc_moduleGroupSettings

### DIFF
--- a/addons/common/functions/fnc_debugModule.sqf
+++ b/addons/common/functions/fnc_debugModule.sqf
@@ -17,5 +17,5 @@
 
 params ["_entity"];
 
-GVAR(LOGDISPLAY_LEVEL) = call compile (_entity getVariable ["logDisplayLevel","4"]);
-GVAR(LOGLEVEL) = call compile (_entity getVariable ["logLevel","4"]);
+GVAR(LOGDISPLAY_LEVEL) = (parseSimpleArray format ["[%1]", _entity getVariable ["logDisplayLevel","4"]]) select 0;
+GVAR(LOGLEVEL) = (parseSimpleArray format ["[%1]", _entity getVariable ["logLevel","4"]]) select 0;
--- a/addons/mk6mortar/functions/fnc_turretDisplayLoaded.sqf
+++ b/addons/mk6mortar/functions/fnc_turretDisplayLoaded.sqf
@@ -45,13 +45,7 @@ private _fnc_hideControl = {
     if (_hideCtrl) then {
         _pos = [-9,-9,0,0];
     } else {
-        {
-            if (isNumber (_path >> _x)) then {
-                _pos pushBack (getNumber (_path >> _x));
-            } else {
-                _pos pushBack (call compile (getText (_path >> _x)));
-            };
-        } forEach ["x", "y", "w", "h"];
+        _pos = ["x", "y", "w", "h"] apply {getNumber (_path >> _x)};
     };
     (_display displayCtrl _idc) ctrlSetPosition _pos;
     (_display displayCtrl _idc) ctrlCommit 0;
--- a/addons/zeus/functions/fnc_bi_moduleCurator.sqf
+++ b/addons/zeus/functions/fnc_bi_moduleCurator.sqf
@@ -240,7 +240,7 @@ if (_activated) then {
         _addons = [];
         {
             if (typeof _x == "ModuleCuratorAddAddons_F") then {
-                _paramAddons = call compile ("[" + (_x getvariable ["addons",""]) + "]");
+                _paramAddons = parseSimpleArray ("[" + (_x getvariable ["addons",""]) + "]");
                 {
                     if !(_x in _addons) then {_addons set [count _addons,_x];};
                     {
--- a/addons/common/functions/fnc_addActionEventHandler.sqf
+++ b/addons/common/functions/fnc_addActionEventHandler.sqf
@@ -45,22 +45,24 @@ _actions pushBack [_condition, _statement];
 
 // first action to add, unit needs addAction command
 if (_actionID == -1) then {
-    private _addAction = call compile format [
-        "[
-            '',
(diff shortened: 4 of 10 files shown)


dedmen:
ace_common_fnc_sendRequest actually didn't conform to the parameters described in the header?

commy2:
Scary PR. Pls don't break anything.

addons/common/functions/fnc_debugModule.sqf
```diff
@@ -17,5 +17,5 @@
 
 params ["_entity"];
 
-GVAR(LOGDISPLAY_LEVEL) = call compile (_entity getVariable ["logDisplayLevel","4"]);
-GVAR(LOGLEVEL) = call compile (_entity getVariable ["logLevel","4"]);
+GVAR(LOGDISPLAY_LEVEL) = (parseSimpleArray format ["[%1]", _entity getVariable ["logDisplayLevel","4"]]) select 0;
+GVAR(LOGLEVEL) = (parseSimpleArray format ["[%1]", _entity getVariable ["logLevel","4"]]) select 0;
```
commy2: PSA breaks when there are spaces.

addons/mk6mortar/functions/fnc_turretDisplayLoaded.sqf
```diff
@@ -45,13 +45,7 @@ private _fnc_hideControl = {
...
         _pos = [-9,-9,0,0];
     } else {
-        {
-            if (isNumber (_path >> _x)) then {
-                _pos pushBack (getNumber (_path >> _x));
-            } else {
-                _pos pushBack (call compile (getText (_path >> _x)));
-            };
-        } forEach ["x", "y", "w", "h"];
+        _pos = ["x", "y", "w", "h"] apply {getNumber (_path >> _x)};
```
commy2: No longer supports globals, as they are read from parsing namespace, unless you change it to the current namespace via call compile.

addons/common/functions/fnc_debugModule.sqf
dedmen: Yes. But there are none here. Input should be a number. Spaces in there are nonsense.

addons/mk6mortar/functions/fnc_turretDisplayLoaded.sqf
PabstMirror: I have no clue why I wrote it like that, this change should be fine

dedmen: The config entries are for Arma Engine. And Arma reads it with parsingNamespace too.  And it works for Arma too. So no problem here.

dedmen:
As I wrote in first comment.
We need to check module stuff whether the user can generate input that would fail. Should've marked as WIP.

addons/zeus/functions/fnc_bi_moduleCurator.sqf
```diff
@@ -240,7 +240,7 @@ if (_activated) then {
         _addons = [];
         {
             if (typeof _x == "ModuleCuratorAddAddons_F") then {
-                _paramAddons = call compile ("[" + (_x getvariable ["addons",""]) + "]");
+                _paramAddons = parseSimpleArray ("[" + (_x getvariable ["addons",""]) + "]");
```
SilentSpike: Optimising BI code, I like it 😆

PabstMirror:
```
test getvariable ["addons",""]
"""ace_medical"", ""ace_zeus"""

call compile ("[" + (test getvariable ["addons",""]) + "]");
["ace_medical","ace_zeus"]

parseSimpleArray ("[" + (test getvariable ["addons",""]) + "]");
["ace_medical",<null>]
```
> Error Undefined variable in expression: _x 
 File z\ace\addons\zeus\functions\fnc_bi_moduleCurator.sqf, line 245

dedmen: The whitespace after comma 🤦‍♂️ 
I asked BI whether they'll remove that. See my comment below

dedmen:
This PR has become kinda obsolete now that BI will have fixed the compile memory leak.
All this would do is a performance improvement now, and none of the changes here are on a critical path.

addons/zeus/functions/fnc_bi_moduleCurator.sqf
dedmen: I'd say I revert the non 100% sure changes and we only leave the addAction changes from common?

TheMagnetar:
@dedmen should we close this?

dedmen:
No memory leak anymore.
These changes are still performance optimizations though:
https://github.com/acemod/ACE3/pull/6810/files#diff-b406794a3deef0d051b1c023f6f5230aL48
https://github.com/acemod/ACE3/pull/6810/files#diff-6344ea9932824bd79a1f90f9e02b8cd9L57
https://github.com/acemod/ACE3/pull/6810/files#diff-5ea6697c61f916ce450e5117a2f77ea2R26
https://github.com/acemod/ACE3/pull/6810/files#diff-00295e150cdf69d75dfe3d1cc813425cR48
https://github.com/acemod/ACE3/pull/6810/files#diff-1897f2bb4dbe630cd59a18322f8e27c4R55

But not worth the effort for me to update this PR to remove the parseSimpleArray stuff and test the rest.
Saving like a quarter of a millisecond per call on these compiles.

addons/mk6mortar/functions/fnc_turretDisplayLoaded.sqf
commy2: You're right. I was thinking about config entries read by the engine and not getNumber. OOps.

PR #6815, opened 2019-02-14
Cleanup spacing on call COMPILE_FILE

PabstMirror:
match expected pattern in https://github.com/CBATeam/CBA_A3/pull/1056/files


dedmen:
Didn't even notice these in my debug logs when testing that cba PR.

PR #6821, opened 2019-02-17
Cargo - Add condition checking during load/unload

PabstMirror:
Fix:
Car could drive off during loading and progress bar would continue
Showed load actions for blownup/dead vehicles 

postInit - use `findIf`
addCargoVehiclesActions - check alive and not locked
startLoadIn/startUnload - add checks during progressBar


bux:
Could all those requirements be moved to a separate function?
They get repeated more than once.

PabstMirror:
it's similar to `FUNC(canUnloadItem)` but that also checks find unloadPos which is expensive

severgun:
Repair need same checks. You can dismount wheel on the go.

PR #6825, opened 2019-02-19
Rewrite rectify convert option

TheMagnetar:
- Rewrite rectify: Adding option to convert vanilla medical gear #6680


addons/medical_treatment/functions/fnc_checkItems.sqf
```diff
@@ -20,6 +20,8 @@ _unit removeItems "FirstAidKit";
 private _countMedikit = [_unit, "Medikit"] call EFUNC(common,getCountOfItem);
 _unit removeItems "Medikit";
 
+if (GVAR(convertItems) == 0) exitWith {};
```
dedmen: Doesn't match with the PR #6680?

PabstMirror: ^

PR #6827, opened 2019-02-20
Explosives - Change SLAM magnetic to detonate upon defusal

alganthe:
- Make SLAM mines set to bottom attack explode upon defusal attempt.

SLAM mines anti-tamper is also active on the bottom attack mode as seen here:

https://en.wikipedia.org/wiki/Selectable_Lightweight_Attack_Munition#Overview

### DIFF
--- a/addons/explosives/CfgAmmo.hpp
+++ b/addons/explosives/CfgAmmo.hpp
@@ -65,6 +65,7 @@ class CfgAmmo {
         GVAR(size) = 0;
     };
 
+    // "The SLAM has an anti-tamper feature that is only active in the bottom- and side-attack modes."
     class SLAMDirectionalMine_Wire_Ammo: DirectionalBombBase {
         indirectHitRange = 20;
         GVAR(explodeOnDefuseChance) = 1;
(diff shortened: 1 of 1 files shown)


dedmen:
We don't have the self destruct feature implemented do we?

alganthe:
^ If you mean mines exploding upon a failed defusal, yes we do.

dedmen:
>^ If you mean mines exploding upon a failed defusal, yes we do.
No, the self destruct feature where the SLAM mine detonates itself after 4/8/24 hours even if it's not triggered.
That's not implemented as far as I can see.

bux:
Feel free to implement 🌮

alganthe:
from the looks of it the variant we have in arma is the M2, solid green, disable itself silently after a set amount of time.

as to if it's possible to disarm it once the anti-tamper is active, I couldn't find any info on it.

PabstMirror:
I believe the `GVAR(explodeOnDefuseChance) = 0;` was intended to be on the timed mode?
that would match the wikipedia description

PR #6830, opened 2019-02-22
Repair - Add condition check to progressBar

PabstMirror:
also removes a `compile`


commy2:
Slightly off topic, but we need a check like this in medical. Currently you can begin treatment and it still works even if the patient walked away while you're looking at the loading bar.

PabstMirror:
medical and repair use the same framework
but a few medial treatments actually use the callback

PR #6832, opened 2019-02-23
Medical Add Variable PAK Times

RulingCommander:
- Introduce new settings in the ACE Medical settings: PAK Time Minimum and PAK Time Coefficient.
- Use variable Min and Max PAK time settings in place of hard-coded values for PAK time. Max PAK time will be derived from Min * Coefficient.
- Add stringtable.xml entries for newly defined Min and Coefficient PAK time settings.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -126,6 +126,7 @@ ramius86 <pasini86@hotmail.com>
 Raspu86
 Riccardo Petricca <petriccarcc@gmail.com>
 Robert Boklahánics <bokirobi@gmail.com>
+RulingCommander <rulingcommander@gmail.com>
 ruPaladin <happyworm24@rambler.ru>
 simon84 <badguy360th@gmail.com>
 Skengman2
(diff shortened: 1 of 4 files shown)


SilentSpike:
I'm not as active here these days so take my opinion with a grain of salt, but personally I think a coefficient is more suitable here.

TheMagnetar:
I agree, you avoid problems with assigning a higher value to min than to max.

RulingCommander:
Using a coefficient is a good idea, hadn't occurred to me. Updated to now use minimum time and a coefficient settings, where the maximum time comes from minimum * coefficient.

PabstMirror:
Rewrite has changed quite a bit,
https://github.com/acemod/ACE3/blob/master/addons/medical_treatment/functions/fnc_healTime.sqf

current setting is "How long should PAK take to apply? Use 0 for default (based on damage)"

RulingCommander:
I had the minimum also be adjustable instead of remaining 10 seconds to allow for a bit more customizability. You can force a single time by setting the coefficient to 1, or keep damage scaling with different mins and maxes. I can look into combining my changes with yours, if you're interested in that.

PabstMirror:
#6933 added a setting for fullHeal scaling coefficent (with a constant 10sec min)
https://github.com/acemod/ACE3/blob/master/addons/medical_treatment/functions/fnc_getHealTime.sqf

PabstMirror:
since 3.13 medical now contains a setting inspired by this PR
see https://github.com/acemod/ACE3/blob/master/addons/medical_treatment/functions/fnc_getHealTime.sqf#L29

PR #6837, opened 2019-02-27
Update Jekyll documentation packages and fix lodash vulnerability

jonpas:
- Update Jekyll packages
- Fix `lodash` security vulnerability (require 4.17.11 or higher)


jonpas:
Ran it in docker, was fine, so should be here too.

PR #6843, opened 2019-03-03
Captives - Use the doApplyHandcuffs function for keybind

C0kkie:
- The problem is that the shortcut checks if the unit has cabletie but if you use the shortcut it will not remove the cabletie from the unit. Better use the thing i changed to do this.

### DIFF
--- a/addons/captives/XEH_postInit.sqf
+++ b/addons/captives/XEH_postInit.sqf
@@ -44,7 +44,7 @@ if (!hasInterface) exitWith {};
     if ((_target distance ACE_player) > getNumber (configFile >> "CfgVehicles" >> "CAManBase" >> "ACE_Actions" >> "ACE_ApplyHandcuffs" >> "distance")) exitWith {false};
 
     if ([ACE_player, _target] call FUNC(canApplyHandcuffs)) exitWith {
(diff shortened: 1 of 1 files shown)


addons/captives/XEH_postInit.sqf
```diff
@@ -44,7 +44,7 @@ if (!hasInterface) exitWith {};
     if ((_target distance ACE_player) > getNumber (configFile >> "CfgVehicles" >> "CAManBase" >> "ACE_Actions" >> "ACE_ApplyHandcuffs" >> "distance")) exitWith {false};
 
     if ([ACE_player, _target] call FUNC(canApplyHandcuffs)) exitWith {
-        [QGVAR(setHandcuffed), [_target, true], _target] call CBA_fnc_targetEvent;
+        [player, _target] call ACE_captives_fnc_doApplyHandcuffs;
```
mharis001: Use `FUNC(doApplyHandcuffs)`.

commy2: Also, what is `player` doing there?

jokoho48:
```suggestion
        [ACE_player, _target] call FUNC(doApplyHandcuffs);
```

C0kkie: Changed

PR #6845, opened 2019-03-04
Add rksl pm II 5-25x56 compat

dedmen:
^ title. Essentially same as the PM II, But I re-measured the scope height and it turned out to be the same.

Should I name it rksl_pmii_525 maybe?
The rksl pbo name is just pm_525


PabstMirror:
They come in the same pack, 
ok if I just combine it together with the existing `compat_rksl_pm_ii`?

Laid3acK:
In this case, PabstMirror, why not with the good elevations values ?
- the single turn values (as modeled) with the S&B 3-12x50 PM II:
https://www.schmidtundbender.de/en/products/police-military-forces/3-12x50-pm-ii.html

- the double turn values with the S&B 5-25x56 PM II/LP:
https://www.schmidtundbender.de/en/products/police-military-forces/5-25x56-pm-iilp.html

PR #6848, opened 2019-03-07
Add support for N-bangs

bovine3dom:
- Look for an `nBangs` entry in `CfgAmmo` if a thrown grenade is a flashbang
- Make it go bang `n` times with a semi-random delay between each bang
- Flashbangs without an `nBangs` entry are treated as 1-bangs.

Questions:

- I wasn't sure where or if I should document this - I don't see a "Grenade Framework" [here](https://ace3mod.com/wiki/framework/) for example.
- I didn't bother to make the fuse separation configurable, as I don't think people will care very much.
- `nBangs` is a slightly silly name but I couldn't think of anything else. Anyone have any better ideas?


PabstMirror:
I think `nBangs` should have an ace prefix
We could reuse the existing config, existing ammo will already have 1 configured
```
private _nBangs =getNumber (_config >> QGVAR(flashbang)) ;
if (_nBangs > 1) then {
```

jonpas:
Agreed with pabst, expand configuration as possible before adding new entries. 

Make a grenade framework page. ;)

bux:
And in the best of all cases, already add a 9-bang flashbang to ace :)

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,15 @@ if (local _unit) then {
     };
 
     if (getNumber (_config >> QGVAR(flashbang)) == 1) then {
-        private _fuzeTime = getNumber (_config >> "explosionTime");
-
-        [FUNC(flashbangThrownFuze), [_projectile], _fuzeTime] call CBA_fnc_waitAndExecute;
+        private _nBangs = [_config >> "nBangs", "number", 1] call CBA_fnc_getConfigEntry;
```
commy2:
```suggestion
private _nBangs = 1;

if (isNumber (_config >> QGVAR(nBangs)) then {
    _nBangs = getNumber (_config >> QGVAR(nBangs);
};
```
The config entry name written out would be `ace_grenades_nBangs`.

bovine3dom: Why not leave the line as it was but wrap nBangs in QGVAR? Does it throw an error if it exists but is the wrong type?

bovine3dom:
> We could reuse the existing config, existing ammo will already have 1 configured

Hmm, I think `...flashbang = 3` is less readable than nBangs. To me, it would suggest a magic number specifying the type of flashbang rather than the number of bangs. I'm happy to be outvoted though.

addons/grenades/functions/fnc_throwGrenade.sqf
commy2: There is no error. 

I just don't like `CBA_fnc_getConfigEntry`.

bovine3dom:
I'd be happy to add a 9 bang if someone donated the model and all the other gubbins. I'm more interested in 6 bangs because that's what the UK uses (any more than 6 explosions just isn't sporting) - http://tgrm.foxed.ca/Britishgrenades/l%20series/L107A1.jpg

bux:
I'd ask in Slack if someone could pull a 6 and/or a 9 bang model out of the hat. There are some talented modellers lurking around.

dedmen:
>There are some talented modellers lurking around there.
Shouldn't be a problem. Get me some more pictures please. I suggest doing that in a seperate PR though.

bux:
> I suggest doing that in a seperate PR though.
I've created the #flashbang-project channel on our slack.

dedmen:
What does our "randomly turn player around on a bang" code do with nBangs?
I'd say it should only disorient people once.

Also this creates a slight problem. We turn players around to disorient them, which is all fine.
But with nBangs if the player get's blinded he will instinctively turn away to not get blinded again by the next blast.
But with our random movement, the player might end up turning and thinking he turned away, but will be staring directly into the next blast of the nBang. While he expected to be turned away and not blinded.

Do we even check if the bang is in front of the player, to not blind him if he doesn't look into it?

bux:
>But with our random movement, the player might end up turning and thinking he turned away, but will be staring directly into the next blast of the nBang.

I doubt you can predict the random flight pattern of an nBang

dedmen:
>I doubt you can predict the random flight pattern of an nBang
If it's 10m infront of you. You can just turn into the opposite direction. It won't hop past you to the other side of you.

bux:
The effect will turn you away from the flash bang and not randomly around.

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -36,10 +36,18 @@ if (local _unit) then {
...
 
-    if (getNumber (_config >> QGVAR(flashbang)) == 1) then {
+    private _flashbangParams = (_config >> QGVAR(flashbang)) call BIS_fnc_getCfgDataArray;
+    private _nBangs = _flashbangParams param [0,0];
+    if (_nBangs > 0) then {
         private _fuzeTime = getNumber (_config >> "explosionTime");
+        private _std = _flashbangParams param [2,0.05];
+        // variance of symmetric Bates distribution: sigma^2 = (max - min)^2/(12 * 4);
+        private _max = _std * sqrt(12*4)/2;
+        private _interval = _flashbangParams param [1,0.25];
```
commy2: `params`, not 3x `param`.

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -36,10 +36,18 @@ if (local _unit) then {
...
     };
 
-    if (getNumber (_config >> QGVAR(flashbang)) == 1) then {
+    private _flashbangParams = (_config >> QGVAR(flashbang)) call BIS_fnc_getCfgDataArray;
+    private _nBangs = _flashbangParams param [0,0];
+    if (_nBangs > 0) then {
         private _fuzeTime = getNumber (_config >> "explosionTime");
+        private _std = _flashbangParams param [2,0.05];
+        // variance of symmetric Bates distribution: sigma^2 = (max - min)^2/(12 * 4);
+        private _max = _std * sqrt(12*4)/2;
```
commy2:
```suggestion
        private _max = _std * sqrt (12*4) / 2;
```

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -36,10 +36,18 @@ if (local _unit) then {
         playSound3D [_file, objNull, false, getPosASL _projectile, _volume, _pitch, _distance];
     };
 
-    if (getNumber (_config >> QGVAR(flashbang)) == 1) then {
+    private _flashbangParams = (_config >> QGVAR(flashbang)) call BIS_fnc_getCfgDataArray;
```
commy2: Use isNumber, getNumber, getArray, isArray; anything but this function.

addons/grenades/CfgAmmo.hpp
```diff
@@ -103,6 +103,7 @@ class CfgAmmo {
 
     class ACE_G_M84: SmokeShell {
         GVAR(flashbang) = 1;
+
```
bovine3dom: Was this added on purpose? Is it to ensure that the file is marked as changed by this commit?

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -36,17 +36,31 @@ if (local _unit) then {
         playSound3D [_file, objNull, false, getPosASL _projectile, _volume, _pitch, _distance];
     };
 
-    private _flashbangParams = (_config >> QGVAR(flashbang)) call BIS_fnc_getCfgDataArray;
-    private _nBangs = _flashbangParams param [0,0];
-    if (_nBangs > 0) then {
+    if (getNumber (_config >> QGVAR(flashbang)) == 1) then {
+        private _amount = 1;
+        if (isNumber (_config >> QGVAR(flashbangAmount))) then {
+            _amount = getNumber (_config >> QGVAR(flashbangAmount));
```
bovine3dom: I think this variable name is ambiguous - it could refer to the amount of powder, for example, and thus the strength of the effect. If it's the `n` in `nBangs` you didn't like, can we just do `_bangs` and `flashbangBangs` since it's a getNumber, so it's obvious it is a number?

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -36,17 +36,31 @@ if (local _unit) then {
...
+        };
+
         private _fuzeTime = getNumber (_config >> "explosionTime");
-        private _std = _flashbangParams param [2,0.05];
+
+        private _standardDeviation = 0.05;
+        if (isNumber (_config >> QGVAR(flashbangDeviation))) then {
+            _standardDeviation = getNumber (_config >> QGVAR(flashbangDeviation));
```
bovine3dom: I don't think `flashbangDeviation` is especially ideal, as it could refer to inaccuracy in how it was thrown? Perhaps `flashbangIntervalStd` or `flashbangIntervalStandardDeviation`?

docs/wiki/framework/flash-framework.md
```diff
@@ -0,0 +1,52 @@
...
+---
+
+## 1. Overview
+
+ACE provides a simple interface for creating flashbang grenades and specifying their properties.
+
+
+## 2. Config Values
+
+There is only one config entry that is specific to ACE flashbangs: `ace_grenades_flashbang`. All successive values can be left undefined and they will be given the defaults shown in the first example below.
```
bovine3dom: As commy2 pointed out on Slack,

> commy2 [9:48 PM]
> For this to work you have to make sure that lifeTime is sufficiently longer than fuzeTime plus amount-1)*interval plus randomness.
> Otherwise the last explosions get lost.
> I think they spawn at 0,0,0 if the grenade projectiles becomes null.

addons/grenades/CfgAmmo.hpp
commy2: Didn't delete the newline after debugging. I usually put a space between vanilla and addon config entries.

addons/grenades/functions/fnc_throwGrenade.sqf
commy2: flashbangBangs is fine.

addons/grenades/functions/fnc_throwGrenade.sqf
commy2: flashbangIntervalDeviation preferably, or flashbangIntervalStandardDeviation if you insist.

bovine3dom: I called it `flashbangIntervalSpread` to make it clear I was being imprecise, and documented it more precisely in the wiki. I couldn't stomach `flashbangIntervalStandardDeviation`...

docs/wiki/framework/flash-framework.md
```diff
@@ -14,39 +14,57 @@ version:
...
 
-There is only one config entry that is specific to ACE flashbangs: `ace_grenades_flashbang`. All successive values can be left undefined and they will be given the defaults shown in the first example below.
+There are several config entries specific to ACE flashbangs. All successive values can be left undefined and they will be given the defaults shown in the first example below.
 
 ```cpp
 class CfgAmmo {
     class MyDefaultFlashbang {
-        ace_grenades_flashbang[] = {0,0.25,0.05}; // Not actually a flashbang
-        //ace_grenades_flashbang[] = {NUMBER_OF_BANGS,BANG_INTERVAL,BANG_DEVIATION};
+        ace_grenades_flashbang = 0;                 // Not a flashbang
```
jonpas: But class is flashbang, would also make more sense since you continue comments with "with one bang".

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
+
+        private _standardDeviation = 0.1;
+        if (isNumber (_config >> QGVAR(flashbangIntervalSpread))) then {
+            _standardDeviation = getNumber (_config >> QGVAR(flashbangIntervalSpread));
+        };
+
+        // variance of symmetric Bates distribution: sigma^2 = (max - min)^2/(12 * 4);
+        private _max = _standardDeviation * sqrt (12*4) / 2;
+
+        for "_i" from 0 to (_amount - 1) do {
```
commy2:
```suggestion
        for "_i" from 0 to (_bangs - 1) do {
```

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
     };
 
     if (getNumber (_config >> QGVAR(flashbang)) == 1) then {
+        private _amount = 1;
```
commy2:
```suggestion
        private _bangs = 1;
```

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
+        if (isNumber (_config >> QGVAR(flashbangBangs))) then {
+            _amount = getNumber (_config >> QGVAR(flashbangBangs));
```
commy2:
```suggestion
            _bangs = getNumber (_config >> QGVAR(flashbangBangs));
```

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
+
         private _fuzeTime = getNumber (_config >> "explosionTime");
 
-        [FUNC(flashbangThrownFuze), [_projectile], _fuzeTime] call CBA_fnc_waitAndExecute;
+        private _interval = 0.5;
+        if (isNumber (_config >> QGVAR(flashbangInterval))) then {
+            _interval = getNumber (_config >> QGVAR(flashbangInterval));
+        };
+
+        private _standardDeviation = 0.1;
```
commy2:
```suggestion
        private _intervalSpread = 0.1;
```

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
+        if (isNumber (_config >> QGVAR(flashbangIntervalSpread))) then {
+            _standardDeviation = getNumber (_config >> QGVAR(flashbangIntervalSpread));
```
commy2:
```suggestion
            _intervalSpread = getNumber (_config >> QGVAR(flashbangIntervalSpread));
```

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
+        };
+
+        // variance of symmetric Bates distribution: sigma^2 = (max - min)^2/(12 * 4);
+        private _max = _standardDeviation * sqrt (12*4) / 2;
```
commy2:
```suggestion
        private _max = _intervalSpread * sqrt (12*4) / 2;
```

docs/wiki/framework/flash-framework.md
commy2:
```suggestion
        ace_grenades_flashbang = 1;                 // A flashbang
```

jonpas:
Make sure to test this with advanced throwing and especially holding it in hand / cooking. _It would make sense to force drop it on first explosions._

docs/wiki/framework/flash-framework.md
bovine3dom: I wanted to list the defaults. I think I'll need to reword a little of the surrounding docs to change this.

addons/grenades/functions/fnc_throwGrenade.sqf
bovine3dom: I'd rather leave this as standardDeviation as it's more precise. I used spread as an imprecise shorthand for standardDeviation.

bovine3dom:
I tested it with advanced throwing. My understanding is that these grenades don't actually explode - the casing remains in tact. The 9-bang in particular is vented only at the top and bottom. I suspect one would get burnt and traumatised, but I don't think it is impossible that someone mad enough could hold onto it.

Once/if we make the power configurable, I want to leave open the possibility for people to use this to make, e.g, fire crackers, which could be held.

addons/grenades/functions/fnc_throwGrenade.sqf
commy2: Then we change the config entry to that as well.

docs/wiki/framework/flash-framework.md
jonpas: Just add: ` (default: 0 - not flashbang)`

jonpas:
Ah neat, didn't know how they work, interesting stuff.

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
         private _fuzeTime = getNumber (_config >> "explosionTime");
 
-        [FUNC(flashbangThrownFuze), [_projectile], _fuzeTime] call CBA_fnc_waitAndExecute;
+        private _interval = 0.5;
+        if (isNumber (_config >> QGVAR(flashbangInterval))) then {
+            _interval = getNumber (_config >> QGVAR(flashbangInterval));
+        };
+
+        private _standardDeviation = 0.1;
+        if (isNumber (_config >> QGVAR(flashbangIntervalSpread))) then {
```
commy2:
```suggestion
        if (isNumber (_config >> QGVAR(flashbangIntervalStandardDeviation))) then {
```

bovine3dom:
From looking at https://github.com/acemod/ACE3/milestone/42, it looks like the convention is for doc changes to be in separate PRs, so that they can be merged on release. Should I split it off into a separate PR?

---

I agree that the standard deviation question is much more wide ranging than this issue. I very strongly believe that it makes more sense for the user. It would make sense for it to be included in CBA so it could be used elsewhere, too.

We could agree to disagree and put both in, with one being overridden by the other. I am slightly concerned that if we don't get this sorted sensibly before merging, it will be harder to change as add-ons may start to depend on it.

When I have time, I will make a PR to CBA and I will try to explain why it makes more sense.

commy2:
We'll only put one in and it will not be the STD.

jonpas:
> Should I split it off into a separate PR?

No, keep it in this one, easier to track.

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
     };
 
     if (getNumber (_config >> QGVAR(flashbang)) == 1) then {
+        private _bangs = 1;
```
Dystopian: `private _bangs = [_config >> QGVAR(flashbangBangs), "NUMBER", 1] call CBA_fnc_getConfigEntry;`

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
+        if (isNumber (_config >> QGVAR(flashbangBangs))) then {
+            _bangs = getNumber (_config >> QGVAR(flashbangBangs));
+        };
+
         private _fuzeTime = getNumber (_config >> "explosionTime");
 
-        [FUNC(flashbangThrownFuze), [_projectile], _fuzeTime] call CBA_fnc_waitAndExecute;
+        private _interval = 0.5;
```
Dystopian: as above `CBA_fnc_getConfigEntry`

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
+        if (isNumber (_config >> QGVAR(flashbangInterval))) then {
+            _interval = getNumber (_config >> QGVAR(flashbangInterval));
+        };
+
+        private _standardDeviation = 0.1;
```
Dystopian: as above `CBA_fnc_getConfigEntry`

addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -37,9 +37,31 @@ if (local _unit) then {
...
+        if (isNumber (_config >> QGVAR(flashbangIntervalSpread))) then {
+            _standardDeviation = getNumber (_config >> QGVAR(flashbangIntervalStandardDeviation));
+        };
+
+        // variance of symmetric Bates distribution: sigma^2 = (max - min)^2/(12 * 4);
+        private _max = _standardDeviation * sqrt (12*4) / 2;
+
+        for "_i" from 0 to (_bangs - 1) do {
+            private _fuzeTime = _fuzeTime + _i*_interval + random [- _max, 0, _max];
```
Dystopian: looks like `private` is misused in here. Maybe `_fuzeTime` var name should be changed.

addons/grenades/functions/fnc_throwGrenade.sqf
commy2: No.

addons/grenades/functions/fnc_throwGrenade.sqf
commy2: Okay, but it works anyway like this.

addons/grenades/functions/fnc_throwGrenade.sqf
Dystopian: cache?

Dystopian: I didn't see your comments above about it. Is it because it's in `Fired` EH? Then caching at game start (like in some recent @dedmen's PRs) could be useful.

commy2: No need to cache, unless you add `CBA_fnc_getConfigEntry` which has many if's and == checks.

bovine3dom:
We tried these out last night on a night op. The flashes lit up the whole town in a several hundred metre radius, even when used indoors. I think we need to add more parameters to make this tuneable, and perhaps turn down the M84 in a separate PR. I imagine it looks fine in the daytime but at night it's bonkers.

addons/grenades/functions/fnc_throwGrenade.sqf
commy2: What's with the face? It's true. You don't cache single config lookups. You do it for long(ish) scripts and CBA_fnc_getConfigEntry would qualify maybe, while this currently doesn't.

commy2:
All light goes through obstacles in Arma.

addons/grenades/functions/fnc_throwGrenade.sqf
dedmen: FiredEH perf isn't that important here. A millisecond or two lag on grenade throw is tolerable, you don't throw grenades that often.
Cache is beneficial for things that are looked up often. How many n-bangs do you throw in a mission? A dozen, half a dozen?
So for one cache wouldn't be that useful here. Same way the overhead doesn't matter much here.

PR #6849, opened 2019-03-08
Arsenal - Improvements to 3DEN attribute

mharis001:
- Add various QOL improvements to 3DEN attribute
- Ability to automatically add compatible attachments or magazines to list for all currently added weapons
    - Useful when playing with a lot of mods resulting in hundreds of magazines
    - Button is only visible for applicable categories
- Ability to import an exported list
- Last selected category is automatically selected on load
- Right clicking search bar will clear it and set focus on it
- Clicking the search button will set focus on the search bar
- Preview:
arsenal_attribute_5


addons/arsenal/functions/fnc_attributeAddCompatible.sqf
```diff
@@ -0,0 +1,75 @@
...
+    private _cfgMagazines = configFile >> "CfgMagazines";
+
+    {
+        private _weaponConfig = _cfgWeapons >> _x;
+
+        {
+            private _muzzleConfig = if (_x == "this") then {_weaponConfig} else {_weaponConfig >> _x};
+
+            // Only add existent magazines and ensure correct classname case
+            private _magazines = getArray (_muzzleConfig >> "magazines") select {isClass (_cfgMagazines >> _x)};
```
PabstMirror: CBA_fnc_compatibleMagazines?

dedmen: We have to. For magazineWells.

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
```diff
@@ -0,0 +1,75 @@
...
+// Get list of all weapons in attribute items
+(_configItems select 0) params ["_primaryWeapons", "_secondaryWeapons", "_handgunWeapons"];
+private _attributeWeapons = _attributeItems select {_x in _primaryWeapons || {_x in _secondaryWeapons} || {_x in _handgunWeapons}};
+
+// Add compatible attachments or magazines to attribute
+private _cfgWeapons = configFile >> "CfgWeapons";
+private _itemsToAdd = [];
+
+if (_category == 8) then {
+    private _magazineGroups = +(uiNamespace getVariable QGVAR(magazineGroups));
```
dedmen: What is the copy for? you are not modifying this value.
L54 copies the elements into _itemsToAdd. So no problem there.

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
mharis001: Can change to avoid duplicate code. Will need to test the performance difference for a large weapons list.

PabstMirror: I guess this does match code in fillRightPanel
cba func does cache results, but on first call it probably will be more expensive

mharis001: Did a quick test with ~400 weapons (RHS) and `diag_tickTime`, here are the results for the first and second clicks.
Current:
```
14:08:41 [ACE] (arsenal) INFO: Added compatible magazines (current) - Time: 36.9873 ms
14:09:05 [ACE] (arsenal) INFO: Added compatible magazines (current) - Time: 37.9944 ms
```
CBA_fnc_compatibleMagazines:
```
14:10:59 [ACE] (arsenal) INFO: Added compatible magazines (cba func) - Time: 67.0166 ms
14:11:13 [ACE] (arsenal) INFO: Added compatible magazines (cba func) - Time: 14.0076 ms
```
So not too big of a difference and this is also the extreme case of all weapons. Thoughts?

Drofseh:
Does the add compatible items button always set it to infinite?

mharis001:
@Drofseh It just adds the items to the list, the mode controls whether the items added are whitelisted or blacklisted.

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
dedmen: That only happens once when you open the UI right?
Perf of both is fine for that.
CBA func uses additional memory for it's cache. But reduces code duplication.
So don't really know.
CBA func also filters out non existent magazines (which can occur in magazinewells) which you don't really do (configName will probably return "" so you'll have a bunch of empty strings)

mharis001: The arsenal magazine wells cache also filters out non-existent magazines aswell: https://github.com/acemod/ACE3/blob/master/addons/arsenal/functions/fnc_scanConfig.sqf#L180

Yeah, the performance is fine for both (and I can't see the button being clicked very often). The slowest part is L72 which in the above situation takes 900-1000 ms.

dedmen: Don't really need to worry about 35ms if you have a 1000ms laying around a couple lines lower.

It really doesn't matter. Whatever you prefer. Less code duplication is the only reason that I have left as a reason to change this.

mharis001: Yeah, I think in that case it is fine as is.

PR #6850, opened 2019-03-08
fix to weaponMuzzleVelocity

PabstMirror:


PabstMirror:
other commits were already squash merged,
not sure why it's showing all the files changed, it should just be picking up beb800e and changes in statTextStatement_weaponMuzzleVelocity

Issue #4080, opened 2016-07-11
Update mk6mortar as mortar framework


bux578:
To easily get compatibility with third party mods (e.g. RHS/CUP/Iron front/...) to use the more realistic approach we took with the mk6 mortar

PabstMirror:
Rangetables are fairly easy to make generic.
The problem with mortars in arma is they don't shoot how they say they will.

We have to 
- Modify the TurretInfoDisplay to hide the bad data
- Figure how how they actually shoot, based on where the camera is looking, ground level...
- Modify the display to show the real data

So it's not just some quick config changes, it requires code running to update the displays.
I can try to think about ways to make it easier to port.

PR #6853, opened 2019-03-08
Artillery rangletables (universal)

PabstMirror:
Ref #4080

Universal rangetables for any artillery
Generated by new extension
Some bits based on mk6

3 main components to the addon
- Show accurate elevation and azimuth
 This has been a real pain to deal with, (e.g. aim at sky vs aim at ground and you get radically different results). Short mortars also seem to need a different approach to calculate, elevation calc has improved from the mk6 code.
- Show rangetables for weapons
 New extension to handle calculating all values at runtime, should be multithreaded, non-blocking.
- Apply air resistance (optional)
There is no config air resistance so it's not that bad, I tried using `CBA_fnc_waitUntilAndExecute` because it's pretty lightweight and actual round count should be low.

WIP for now so don't worry about the small stuff
Possible issues:
- Keep testing mods for compatibility issues
- Rocket artillery the ammo deploys a submunition in flight, have to think about applying air resistance to the child?
- Vehicle's huds will still show degrees, Mk6 replaced the hud, but I can't do that universally. Right now it will just display the angles in the upper right and the (incorrect) degrees will remain on the hud.
ex


Blutze:
Aren't rocket artillery submunitions deployed only like, 200-300 meters slope distance from the target? I'd rather worry about heavily scripted special ammo like https://en.wikipedia.org/wiki/SMArt_155... if you have to respawn the projectile at the vector and velocity determined by this system, those scripts might not get triggered. Similar to RHS impact hand grenades with Advanced Throwing.

Semi-related: Couldn't Frag be rewritten with the Jets DLC submunition system?

extensions/artillerytables/artillerytables.cpp
```diff
@@ -0,0 +1,356 @@
...
+        std::tie(std::ignore, testResultDist, std::ignore) = simulateShot(testAngle, _muzzleVelocity, 0, 0, 0, 15, 1, _airFriction);
+        if (testResultDist > bestDistance) {
+            bestAngle = testAngle;
+            bestDistance = testResultDist;
+        }
+    }
+    return { bestAngle, bestDistance };
+}
+
+std::tuple<double, double, double> simulateFindSolution(const double& _rangeToHit, const double& _heightToHit, const double& _muzzleVelocity, const double& _airFriction, const double& _minElev, const double& _maxElev, const bool& _highArc) {
```
dedmen: `const bool&` is less effective than just bool. Need to transmit a full 32/64 bit memory address instead of just a single byte.

extensions/artillerytables/artillerytables.cpp
```diff
@@ -0,0 +1,356 @@
...
+    }
+    ss << std::fixed << std::setw(_widthInt) << std::setprecision(_widthDec) << _num;
+}
+
+std::string simulateCalcRangeTableLine(const double& _rangeToHit, const double& _muzzleVelocity, const double& _airFriction, const double& _minElev, const double& _maxElev, const bool& _highArc) {
+    auto [actualDistance, lineElevation, lineTimeOfFlight] = simulateFindSolution(_rangeToHit, 0, _muzzleVelocity, _airFriction, _minElev, _maxElev, _highArc);
+    if (lineTimeOfFlight < 0) {
+        return "";
+    }
+    auto [actualDistanceHeight, lineHeightElevation, lineHeightTimeOfFlight] = simulateFindSolution(_rangeToHit, -100, _muzzleVelocity, _airFriction, _minElev, _maxElev, _highArc);
```
dedmen: Wasn't there some reason in the past why we didn't want to use C++17?

jonpas: Yes, reason was it just came out and compiler support was not ideal. I still don't see a reason to use it for one line though.

extensions/artillerytables/artillerytables.cpp
TheMagnetar: Well in theory & should only be used for classes/structs. In that case, types like double or int should just be passed by value... but there is nothing set on stone except maybe on MISRA ruleset

extensions/artillerytables/artillerytables.cpp
dedmen: gcc support for C++17 is still abysmal. And we compile our linux plugins with gcc.
Although this plugin should be clientside only, so windows only. So it is kinda fine.
I didn't check if there is more C++17 in here.

extensions/artillerytables/artillerytables.cpp
dedmen: Yeah I thought the same. But technically on 32bit you still get benefit if you pass a 64bit double by ref because it's bigger than the pointer and if you don't have xmm registers available to hold it.
But not really worth it.

extensions/artillerytables/artillerytables.cpp
PabstMirror: but it's 2019 :smile: 
I think this was the only 17 feature, should compile anywhere now

dedmen:
Have to remove your debug printf's before merging.
And someone in Slack talked about checking if there is a solvable algo for that?

PabstMirror:
`// printf` will be removed by the preprocessor :wink: 

I don't think there is a solution for v^2 wind resistance
ref `https://www.youtube.com/watch?v=uq4cg0MtzVo&t=792`
keep in mind I need 3d, not 2d, time of flight and handle stopping at a given height

but I could be wrong
I didn't have much to base this on, so it's written naively and pretty much just brute forces the solution

dedmen:
Not really PR relevant.
But we should probably change to Ox
https://github.com/acemod/ACE3/blob/master/extensions/CMakeLists.txt#L32
and to O3
https://github.com/acemod/ACE3/blob/master/extensions/CMakeLists.txt#L23
and OPT:REF/OPT:ICF
https://github.com/acemod/ACE3/blob/master/extensions/CMakeLists.txt#L29


And we can get the size down further with GCC/Clang
https://github.com/intercept/intercept/blob/master/src/host/intercept_dll/CMakeLists.txt#L42
Though we don't build linux anyway so doesn't matter.

Should get the size down a bit for all future extensions. Needs to be tested tho

addons/mk6mortar/functions/fnc_rangeTableCanUse.sqf
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Example:
+ * [bob, bob] call ace_mk6mortar_fnc_rangeTableCanUse
+ *
+ * Public: No
+ */
+
+params ["_vehicle", "_player"];
+
+"ACE_RangeTable_82mm" in (_player call EFUNC(common,uniqueItems));
```
dedmen: missing newline 😗

addons/common/functions/fnc_checkFiles.sqf
```diff
@@ -105,6 +105,7 @@ if (!isServer && {_platform in ["linux", "osx"]}) then {
         };
     } forEach ("true" configClasses (configFile >> "ACE_Extensions"));
 };
+if (isArray (configFile >> "ACE_Extensions" >> "extensions")) then { WARNING("extension[] array no longer supported"); };
```
jonpas:
```suggestion
if (isArray (configFile >> "ACE_Extensions" >> "extensions")) then {
    WARNING("extensions[] array no longer supported");
};
```

PR #6854, opened 2019-03-09
Fix M84 lods

### LINKED ISSUE #6851: M84 looks gash in advanced throwing
look, it's awful

dedmen and jonpas suggest it might be an LOD issue.

this is less bad

Opening it in Oxygen seems to confirm it - it seems to be using LOD 7:

oh god why

dedmen:
Fixes #6851


bux:
how does it look now?

PabstMirror:
PR on the left,
I don't think those lod names are right?
image

commy2:
Pabst, you have to delete the distance lods, otherwise the farhest LOD is displayed for grenades, because the game is bugged. This was acknowledged by reyhard on the discord.

Dedmen kept them as Edit LODs, in case they ever fix this, or if this model will get recycled in another game.

dedmen:
>I don't think those lod names are right?

They are

dedmen:
>how does it look now?

arma3_x64_2019-03-09_19-37-26

PR #6856, opened 2019-03-09
Grenades - Add CTS 7290 9-bang flashbang

dedmen:
Requires #6848


dedmen:
arma3_x64_2019-03-09_19-42-46

addons/grenades/config.cpp
```diff
@@ -10,7 +10,8 @@ class CfgPatches {
             "ACE_HandFlare_Green",
             "ACE_HandFlare_Yellow",
             "ACE_M84",
-            "ACE_M14"
+            "ACE_M14",
+            "ACE_CTS9"
```
bovine3dom: Unsure what the ACE guidelines are, but some projects prefer items to be prepended to lists like this to prevent unrelated lines being edited (i.e, the M14 line here).

addons/grenades/stringtable.xml
```diff
@@ -144,6 +144,12 @@
             <Korean>플래시뱅이라고도 알려져있습니다. 사용즉시 섬광으로 인한 시력장애, 청각장애, 이명, 내이기관방해를 유발합니다.</Korean>
             <Chinesesimp>也被称为闪光弹，会造成暂时性失明，耳聋，耳鸣等效果。</Chinesesimp>
             <Chinese>也被稱為閃光彈，會造成暫時性失明，耳聾，耳鳴等效果</Chinese>
+        </Key>
+		<Key ID="STR_ACE_Grenades_CTS9_Name">
+            <English>CTS 7290-9 9-bang Flash Grenade</English>
+        </Key>
+        <Key ID="STR_ACE_Grenades_CTS9_NameShort">
+            <English>CTS 7290-9</English>
```
bovine3dom: Couldn't this also be used for all the other languages?

addons/grenades/CfgAmmo.hpp
```diff
@@ -119,6 +119,13 @@ class CfgAmmo {
         effectsSmoke = "ACE_M84FlashbangEffect";
         whistleDist = 0;
     };
+    
+    class ACE_G_CTS9: ACE_G_M84 {
```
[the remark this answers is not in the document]
bovine3dom: It probably makes more sense to do that in #6848.

addons/grenades/config.cpp
commy2: It's fine.

addons/grenades/CfgAmmo.hpp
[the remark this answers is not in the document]
commy2: If someone already changes the M84, but changing the inheritance you would break it even more.

addons/grenades/stringtable.xml
commy2: Yes.

addons/grenades/CfgAmmo.hpp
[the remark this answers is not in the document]
bovine3dom: jonpas pointed out on Slack that very few people would want to use our pre-made class as it would introduce a direct dependency on ACE for their mods. In combination with what commy2 is saying, there is little point to a base class.

So, the question is: should we copy and paste the whole `ACE_G_M84` class here or inherit from it? Copying the whole class is probably safer, but this is your PR, dedmen, so I'll let you decide :)

bux:
@dedmen Could you make it so, that the label is slightly rotated to the left or to the right.
Imho it's too perfectly centered right now. It feels to me like either a reporter making sure that his channel's logo is visible on his mic or someone holding a coke with the label visible to the camera straight out of a coke ad

addons/grenades/CfgAmmo.hpp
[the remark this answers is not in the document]
dedmen: The base class would be just to make it easy for us. Not for others to inherit from.
I don't like code duplication so I did it like this.

addons/grenades/stringtable.xml
dedmen: Need to change displayName from 7290-9 to 7290. The 9-bang part is duplicate otherwise. And the name looks crap in arsenal

dedmen:
>And the label is too comical for my taste. Can't we just copy the original label?

Commy wanted the banana :D
The labels on the original are also perfectly centered I think? It's a machine printing them afterall.
Or do you mean not rotate the label but rotate the grenade in advanced throwing?
Yeah can do that. It was actually an accident that it's rotated so perfectly :D

bux:
> Commy wanted the banana

he is also raging about sexual transmitted diseases, so i'm not sure what to make of this 😄

commy2:
Keep the banana.

alganthe:
agreed bux, the Rheinmetall one looks nice.

bovine3dom:
The CTS 9-bang this is based on has the manufacturer's address where the banana is. The label already has all the details that the more sober military spec ones have.

One alternative would be safety instructions - you're not supposed to throw them at muddy ground as it can clog the vents and turn it into a rocket, for example.

I'd like to see the serial number being used, too. One number from the top and bottom just be needs to be blacked out.

jokoho48:
I like the idea of easter eggs in models/textures but they need to be hard to find. I think in close to all models I have done/ where done in my time have small easter eggs, but they are not so open and bland as this one. so when it would be smaller somewhere I am totally fine with it but that big, please no.

dedmen:
Could just move the banana to the bottom ^^

alganthe:
maybe use a friendlier banana, this one belongs in ABE.

jonpas:
Keep it realistic as possible with small hard to find easter eggs. This is what ACE3 is about after-all.

TheMagnetar:
+1 on Rheinmetall

dedmen:
arma3_x64_2019-03-16_12-49-09

That's 20° rotated now.

commy2:
You probably shouldn't use a grenade to peal either.

dedmen:
The triangle on the top is not visible ingame. Only on the screenshot :D

jokoho48:
I don't like the "made by dedmen" text there that breaks it I think again and I don't think any other model in ace currently has that type of marking by the model maker.

dedmen:
>that breaks it I think again

?
>I don't think any other model in ace currently has that type of marking

I guess none of the other ace models say "Made in ..." on the original then?

jokoho48:
> I guess none of the other ace models say "Made in ..." on the original then?

maybe there are some but then the creator did not put there name in there but did things like "Made on Altis" or "Made in Greece" or maybe even "Made in ..." where ever it does get made

commy2:
Is this done apart from merging the branch that implements the scripting support for it?
I'd like to finish that branch and then also merge this.

dedmen:
If nothing changed on the config-entry syntax since it was pushed into here then yeah done.

bux:
Did you remove the "Made by dedmen"?

dedmen:
>Did you remove the "Made by dedmen"?

no

addons/grenades/CfgAmmo.hpp
```diff
@@ -119,6 +119,16 @@ class CfgAmmo {
...
+        GVAR(flashbang) = 1;
+        GVAR(flashbangBangs) = 9;
+        GVAR(flashbangInterval) = 0.5;
+        GVAR(flashbangIntervalStandardDeviation) = 0.1;
```
bovine3dom:
```suggestion
        GVAR(flashbangIntervalMaxDeviation) = 0.35;
```
The variable was renamed to max deviation and now corresponds to the half width of the n=4 Bates distribution, so this is equivalent via `0.1 * sqrt (12*4) / 2  ≈ 0.35`.

jonpas:
> Did you remove the "Made by dedmen"?

That'll have to be done before merge, name does not fit, write what's on the actual thing but modified for Arma.

dedmen:
> write what's on the actual thing 
Actual thing says "Made in U.S.A."

jonpas:
Then use that or a deviation such as "Made in Greece".

bux:
Can we get that finsihed for 3.13?

dedmen:
MADE IN RASMAN

now says "dedmen <3" on bottom

jonpas:
What/Where is `RASMAN`?

> now says "dedmen <3" on bottom

Why? Please remove. Banana references are fine, personal ego boosters are not.

jokoho48:
> What/Where is RASMAN?
RASMAN is the Northern city on Takistan, with the Airfield

jonpas:
> now says "dedmen <3" on bottom
> Why? Please remove. Banana references are fine, personal ego boosters are not.

Do so and we can merge it.

PR #6859, opened 2019-03-11
More trivial code cleanup

dedmen:
Related to #6858
pboprefixes are for both compile and for armake (as armake can also scan for pboprefix, atleast armake++)


addons/common/functions/fnc_checkFiles.sqf
```diff
@@ -69,8 +69,8 @@ if (!(_oldCompats isEqualTo [])) then {
 ///////////////
 // check dlls
 ///////////////
-if (toLower (productVersion select 6) in ["linux", "osx"]) then {
-    INFO_2("Operating system does not support DLL file format");
```
dedmen: How did this not get noticed before? and why didn't arma crash at this invalid macro usage? and why did our linter not pick that up?
so many questions.

dedmen:
defineResincl.inc required by map/functions/fnc_initMainMap.sqf

addons/common/functions/fnc_checkFiles.sqf
bux: per se linux doesn't support *.dll files, does ist?

dedmen:
defineResinclDesign.inc
optionsmenu/init_loadMainMenu.sqf

dedmen:
All typos/bugs/errors detected by my script profiler.
We should run that on CI.
I'll add a argument to recursively compile a folder with multithreading. Should only take a couple seconds for whole ace.

addons/common/functions/fnc_checkFiles.sqf
PabstMirror: none of our releases have ever included `.so` files
with this line, people on linux can at least attempt to use all ace features that do not rely on extensions
now it will block access entirely

dedmen: I thought I have seen them on linux server :thinking:

dedmen:
SQF-VM can now compile all of ace without errors, besides one file in zeus that's really just a redirect to a vanilla file, because the include is not available. And I won't copy that script file into our includes as it doesn't make sense.

include/a3/data_f/$PBOPREFIX$
```diff
@@ -0,0 +1 @@
+a3\data_f
```
jonpas: Why are those necessary? Those files are already in folders representing prefix, and armake doesn't need them.

dedmen: Because some tools (compiler) search for pboprefix to be sure that it's what it should be.
Otherwise the prefix might aswell be include\a3\data_f or ace\include\a3\data_f or whateve

TheMagnetar: EOF?

jonpas: Can be present, any normal tool will parse it just fine.

dedmen: Some bad tools just take the whole $PBOPREFIX$ as prefix.
So if you pack with these you end up with prefix\n 
soo stupid.. 
There is no reason to have newline at the end here really anyway.

jonpas: Forget those tools, just leave it as is. ;)

dedmen: cpbo is one of them. And I use that to build pbo's to quickly test things :D

jonpas: Judging.

dedmen: Solution would be to write a script that let's you add armake pbo packing to context menu.
Context menu stuff is a chore to do though.

PR #6868, opened 2019-03-14
Aircraft - add setting to hide Eject action from action menu

commy2:
- title
- requires a game restart, because the game caches the config entries value once at game start :(
- setting can not be overwritten / is always local, due to the config caching stuff
- PR is mainly meant to show off and one up the other mod from Steam

todo:
- [x] Scripted eject should respect the setting too


veteran29:
And... my addon is not needed 2h after it was released. 👍

Dystopian:
Scripted eject should respect the setting too https://github.com/acemod/ACE3/blob/9d5b936d7c3cb49c300f898845a8a87d45cb17ef/addons/aircraft/functions/fnc_initEjectAction.sqf#L46

Dystopian:
Did I miss something? Why everybody is so happy when disabling this useful action?

Drofseh:
@Dystopian Because it's a useless thing to have in the interaction menu.
Too slow when you need it, but too easy to accidentally select when you meant to do something else.
There's a keybinding already `2xV` that's much quicker and safer.

Dystopian:
Does key work if action is disabled? I often use key and never menu action, that's why asking.

commy2:
Of course the key works. This just hides the action from the action menu. It does not remove the action as keybind. Sadly you cannot remove bindable actions at all.

commy2:
>Scripted eject should respect the setting too 

Hmm, I don't see how this will possible, because actions added by script have no `show` parameter. Therefore you cannot make actions that don't appear in the menu, but still works by the key bound to Eject.

Dystopian:
For scripted action we could change priority and/or color if setting is active.

commy2:
Make it always appear at the bottom and gray?

Dystopian:
Yes. Also I think red is better not to select it accidentally. But this should be chosen by those who need this setting.

commy2:
Red stands out, as if it wants to be executed.

commy2:
Done.

addons/aircraft/XEH_preInit.sqf
```diff
@@ -6,4 +6,41 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+if (hasInterface) exitWith {
```
PabstMirror:
```suggestion
if (hasInterface) then {
```
so the final `ADDON = true;` gets run

PR #6872, opened 2019-03-16
version check handle minor versions below 10

### LINKED ISSUE #6871: PBO Checking doesn't handle two-digit sub numbers correctly
**Arma 3 Version:** `1.90.145381` (stable / rc / dev)
**CBA Version:** `3.10.0.190314 (client) / 3.9.1.181229 (server)`
**ACE3 Version:** `3.12.5`

**Mods:**
```
- CBA_A3
- ace
- https://steamcommunity.com/sharedfiles/filedetails/?id=1373175187
- 3cb Weapons ACE compat, VPHUD, dzn Vehicle on Fire, ACE Interaction Menu Expansion, some NIArms compat mods
```

**Description:**
CBA 3.10 release on Steam happened ~15 minutes before our scheduled mission start. The server is running essentially 24/7 and needs to be manually kicked into grabbing mod updates from the SWS.
After loading in, those who got their clients updated got slapped by PBO Checking for having "CBA 3.1" which was claimed to be older than CBA 3.9; I checked CfgPatches in the config viewer pretty much immediately and it was showing `version = "3.10.0.190314";` for all CBA classes that had a version defined.

**Steps to reproduce:**
- Have CBA 3.10 loaded
- Join a server running CBA 3.9 and PBO Checking
- Observe the chat message dropping the zero in the version number and saying the newer version is older

- Dedicated

**Additional information:**
- For us, this time, it was only a display issue. But if the system ever has to compare `x.1.x` to `x.10.x`, it might end up calling them identical.

**RPT log file:**
- https://gist.github.com/Blutze/d3ce1a84efae33d815a0b14a5c437102
- Chat message is in line 3393, two lines below that is the log entry also claiming my version is older.

commy2:
- fix #6871

Note that a big rewrite of this function, that would for example handle `1.99 < 1.100` etc. correctly is not desirable. The function has to be cross compatible between all versions due to its nature and therefore this transferred variable should stay a number and not become an array or string.

### DIFF
--- a/addons/common/scripts/checkVersionNumber.sqf
+++ b/addons/common/scripts/checkVersionNumber.sqf
@@ -10,7 +10,7 @@ private _files = CBA_common_addons select {
 
 private _versions = [];
 {
-    getText (configFile >> "CfgPatches" >> _x >> "version") splitString "." params ["_major", "_minor"];
+    getText (configFile >> "CfgPatches" >> _x >> "version") splitString "." params [["_major", "0"], ["_minor", "0"]];
     private _version = parseNumber _major + parseNumber _minor/100;
     _versions set [_forEachIndex, _version];
 } forEach _files;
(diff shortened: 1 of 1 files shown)


commy2:
3.9.123456

Before:
3.9

After:
3.09

_____

3.10.123456

Before:
3.1

After:
3.1

_______

3.9 > 3.1 (bugged)
3.1 > 3.09 (correct)

PR #6876, opened 2019-03-17
Fastroping - Add user setting to confirm cutting ropes

PabstMirror:
When you are in queue to fast rope the "Fast rope" action fades in and out as other people are using it and it is possible to accidentally select the "cut ropes". This adds a user setting to move the action to a 2nd level.

Could also just make this default and drop the setting.


Drofseh:
I think this should be the default as well.

bux:
Default... and needs an icon ;)

alganthe:
agreed, it should be default.

imho confirm shouldn't have an icon, but fastrope and cut ropes definitely should, it would help people differentiate the two at a quick glance.

bux:
We should have an icon for cutting the primary parachute, iirc

PR #6880, opened 2019-03-18
WIP: Fix medical extension ignoring allowed selections for injury types

thojkooi:
- fix the medical extension ignoring the allowed selections for injury types by only including injury types that match selections

I haven't been able to test this yet and the extensions should be recompiled for this. I unfortunately do not have a windows machine available at this time.


commy2:
On Slack they said it is also broken for the fallback SQF version. That should be fixed as well.

extensions/medical/handleDamage.cpp
```diff
@@ -54,11 +54,12 @@ namespace ace {
             return stream.str();
         }
 
-        std::vector<ace::medical::injuries::OpenWound> handleDamage::GetInjuryInfoFor(const std::string& typeOfDamage, double amountOfDamage, int selection, int woundID)
+        std::vector<ace::medical::injuries::OpenWound> handleDamage::GetInjuryInfoFor(const std::string& typeOfDamage, double amountOfDamage, std::string& selectionName, int woundID)
         {
             std::vector<ace::medical::injuries::OpenWound> injuriesToAdd;
             std::vector<std::shared_ptr<ace::medical::injuries::InjuryType>> information;
             std::shared_ptr<ace::medical::injuries::InjuryType> highestSpot = nullptr;
+            int selection = SelectionToNumber(selectionName);
```
TheMagnetar: const int32_t

extensions/medical/handleDamage.cpp
```diff
@@ -68,13 +69,15 @@ namespace ace {
...
                         {
-                            if (highestSpot == NULL)
-                                highestSpot = possibleInjury;
-
-                            if (possibleInjury->minDamage > highestSpot->minDamage)
-                                highestSpot = possibleInjury;
-
-                            information.push_back(possibleInjury);
+                            if (InjuryAllowedOnSelection(possibleInjury->allowedSelections, selectionName)) {
+                                if (highestSpot == NULL) {
```
TheMagnetar: nullptr

extensions/medical/handleDamage.cpp
```diff
@@ -54,11 +54,12 @@ namespace ace {
             return stream.str();
         }
 
-        std::vector<ace::medical::injuries::OpenWound> handleDamage::GetInjuryInfoFor(const std::string& typeOfDamage, double amountOfDamage, int selection, int woundID)
+        std::vector<ace::medical::injuries::OpenWound> handleDamage::GetInjuryInfoFor(const std::string& typeOfDamage, double amountOfDamage, std::string& selectionName, int woundID)
```
TheMagnetar: amountOfDamage and woundID are also const values

PabstMirror:
@thojkooi - does the sqf fallback function need work as well?
or is this just making the extension work the same?

extensions/medical/handleDamage.cpp
dedmen: selectionName should be const to not accidentally modify it in here.

extensions/medical/handleDamage.cpp
```diff
@@ -68,13 +69,15 @@ namespace ace {
...
                         if (amountOfDamage >= possibleInjury->minDamage && (amountOfDamage <= possibleInjury->maxDamage || possibleInjury->maxDamage <= 0))
                         {
-                            if (highestSpot == NULL)
-                                highestSpot = possibleInjury;
-
-                            if (possibleInjury->minDamage > highestSpot->minDamage)
-                                highestSpot = possibleInjury;
-
-                            information.push_back(possibleInjury);
+                            if (InjuryAllowedOnSelection(possibleInjury->allowedSelections, selectionName)) {
```
TheMagnetar: `allowedSelections` is undefined

extensions/medical/handleDamage.cpp
```diff
@@ -217,6 +220,11 @@ namespace ace {
             }
         }
 
+        bool handleDamage::InjuryAllowedOnSelection(const std::vector<std::string>& allowedSelections, const std::string& selectionName)
```
TheMagnetar: Not modified in the header file

extensions/medical/handleDamage.cpp
```diff
@@ -217,6 +220,11 @@ namespace ace {
...
+        {
+            return selectionName == "All" || find(allowedSelections.begin(), allowedSelections.end(), selectionName) != selections.end();
```
TheMagnetar: `selections` is undefined

thojkooi:
>  does the sqf fallback function need work as well?

That should already work as intended for so far I can tell. I still haven't had the opportunity to validate this myself though..

extensions/medical/handleDamage.cpp
dedmen: Just `!highestSpot`

dedmen:
Did you just forget to push changes to the other files? Explains most of the build problems here.
I made a PR to fix this.
When this is merged I'll make a cleanup PR to remove unused stuff like woundID's.
Btw.. What's up with the `{` on seperate line stuff in medical? We don't do that in SQF nor in other Extensions. Should I fix that in my cleanup PR?

TheMagnetar:
>Btw.. What's up with the { on seperate line stuff in medical?
>We don't do that in SQF nor in other Extensions. Should I fix that in my cleanup PR?

Yes please. If you are at it and if you don't mind can you add const correctness as well to function declaration and the use of base types (int32_t)?

extensions/CMakeLists.txt
```diff
@@ -92,6 +92,7 @@ set(ACE_VERSION_MAJOR 3)
 set(ACE_VERSION_MINOR 12)
 set(ACE_VERSION_REVISION 3)
 EXECUTE_PROCESS(COMMAND git rev-parse --verify HEAD
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
```
dedmen: Btw this didn't work with CMake integrated into visual studio, because the build and workdir would be in a temp directory somewhere else and won't be a git dir.
Forcing work directory to be in source directory fixes that.

PabstMirror:
I think this can be closed as we're probably not going back to the extension

PR #6885, opened 2019-03-22
Add keep engine of vehicles running after getting out

brainslush:
- Adds a feature which leaves the vehicle engine running after getting out. So, it requires the driver to switch off the engine of the vehicle.
I tested it locally and have tested it numerous times as part of another mod before moving it to ACE3. Please retest as I could made have mistakes while adapting the code for ACE3.


addons/vehicles/XEH_postInit.sqf
```diff
@@ -1,6 +1,8 @@
 // by esteldunedain
 #include "script_component.hpp"
 
+[QGVAR(engineOnLocal), {_this engineOn true;}] call CBA_fnc_addEventHandler;
```
jonpas: Already exists https://github.com/acemod/ACE3/blob/4ff01967b6b553ad55b8360aa210617e0a035839/addons/common/XEH_postInit.sqf#L128

addons/vehicles/XEH_PREP.hpp
```diff
@@ -1,3 +1,4 @@
 
 PREP(speedLimiter);
 PREP(startEngine);
+PREP(keepEngineRunning);
```
jonpas:
```suggestion
PREP(keepEngineRunning);

```

addons/vehicles/CfgEventHandlers.hpp
```diff
@@ -23,6 +23,19 @@ class Extended_Engine_EventHandlers {
             clientEngine = QUOTE(if (local driver (_this select 0)) then {_this call FUNC(startEngine)};);
         };
     };
+    class AllVehicles {
+        class ACE_EngineState {
+            engine = QUOTE(if (local (_this select 0)) then {(_this select 0) setVariable [ARR_3(QQGVAR(engineState), _this select 1, true)]};);
```
jonpas: Make function? It's long and ugly. ARR also doesn't like spaces too much AFAIK.

addons/vehicles/CfgEventHandlers.hpp
```diff
@@ -32,4 +45,9 @@ class Extended_Init_EventHandlers {
             exclude[] = {"TargetSoldierBase","Static_Designator_01_base_F","Static_Designator_02_base_F","Pod_Heli_Transport_04_base_F"};
         };
     };
+    class AllVehicles {
+        class ACE_EngineState {
+            init = QUOTE(if (local (_this select 0)) then {(_this select 0) setVariable [ARR_3(QQGVAR(engineState),isEngineOn (_this select 0),true)]});
```
jonpas: Same thing as above with different value. Function!

addons/vehicles/functions/fnc_keepEngineRunning.sqf
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Public: No
+ */
+
+params ["_vehicle","_position","_unit"];
+
+if (!GVAR(keepEngineRunning)) exitWith {};
+if (_unit == ACE_player && {_position == "driver"} && {_vehicle getVariable [QGVAR(engineState),false]}) then {
+    [{[QGVAR(engineOnLocal),_this,_this] call CBA_fnc_targetEvent}, _vehicle] call CBA_fnc_execNextFrame;
+};
```
jonpas:
```suggestion
};

```

addons/vehicles/functions/fnc_keepEngineRunning.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * [car, "driver", player] call ace_vehicles_fnc_keepEngineAlive
+ *
+ * Public: No
+ */
+
+params ["_vehicle","_position","_unit"];
+
+if (!GVAR(keepEngineRunning)) exitWith {};
+if (_unit == ACE_player && {_position == "driver"} && {_vehicle getVariable [QGVAR(engineState),false]}) then {
+    [{[QGVAR(engineOnLocal),_this,_this] call CBA_fnc_targetEvent}, _vehicle] call CBA_fnc_execNextFrame;
```
jonpas:
```suggestion
    [{
        [QGVAR(engineOnLocal), _this, _this] call CBA_fnc_targetEvent;
    }, _vehicle] call CBA_fnc_execNextFrame;
```
And change to the other event.

addons/vehicles/initSettings.sqf
```diff
@@ -0,0 +1,7 @@
+
+[QGVAR(keepEngineRunning), "CHECKBOX",
+[LSTRING(SettingKeepEngineRunningName), LSTRING(SettingKeepEngineRunningDesc)],
+localize ELSTRING(common,ACEKeybindCategoryVehicles),
+true, // default value
+true, // isGlobal
+{[QGVAR(keepEngineRunning), _this] call EFUNC(common,cbaSettings_settingChanged)}] call CBA_settings_fnc_init;
```
jonpas: Prettier formatting please. You took one of the old auto-generated files probably.

addons/vehicles/stringtable.xml
```diff
@@ -67,5 +67,13 @@
             <Japanese>速度制限を減らす</Japanese>
             <Polish>Zmniejsz ograniczenie prędkości</Polish>
         </Key>
+        <Key ID="STR_ACE_Vehicles_SettingKeepEngineRunningName">
+            <English>Keep engine runing after geting out</English>
```
jonpas:
```suggestion
            <English>Keep engine running after getting out</English>
```

Maybe this is better?
```suggestion
            <English>Disable automatic engine shut-off</English>
```

addons/vehicles/stringtable.xml
```diff
@@ -67,5 +67,13 @@
...
+            <German>Motor nach Verlassen laufen lassen</German>
+        </Key>
+        <Key ID="STR_ACE_Vehicles_SettingKeepEngineRunningDesc">
+            <English>Prevent the automatic shut off of the engine when exiting vehciles</English>
```
jonpas:
```suggestion
            <English>Prevent the automatic shut-off of the engine when exiting vehicles.</English>
```

addons/vehicles/CfgEventHandlers.hpp
```diff
@@ -23,6 +23,19 @@ class Extended_Engine_EventHandlers {
             clientEngine = QUOTE(if (local driver (_this select 0)) then {_this call FUNC(startEngine)};);
         };
     };
+    class AllVehicles {
+        class ACE_EngineState {
```
commy2:
```suggestion
        class ACE_EngineState {
```

addons/vehicles/CfgEventHandlers.hpp
```diff
@@ -23,6 +23,19 @@ class Extended_Engine_EventHandlers {
...
+            engine = QUOTE(if (local (_this select 0)) then {(_this select 0) setVariable [ARR_3(QQGVAR(engineState), _this select 1, true)]};);
+        };
+    };
+};
+
+class Extended_GetOut_EventHandlers {
+    class AllVehicles {
+        class ACE_KeepEngineRunning {
```
commy2:
```suggestion
        class GVAR(KeepEngineRunning) {
```

addons/vehicles/CfgEventHandlers.hpp
```diff
@@ -32,4 +45,9 @@ class Extended_Init_EventHandlers {
             exclude[] = {"TargetSoldierBase","Static_Designator_01_base_F","Static_Designator_02_base_F","Pod_Heli_Transport_04_base_F"};
         };
     };
+    class AllVehicles {
+        class ACE_EngineState {
```
commy2:
```suggestion
        class GVAR(EngineState) {
```

addons/vehicles/CfgEventHandlers.hpp
commy2: Do we even need this? I am pretty sure the engine is always off on a created vehicle.

commy2: This is an Init event. The script uses `setVariable public`. `setVariable public` does not work on the Init event: the variable will not be synchronized in multiplayer.

brainslush: Is it always off? Think of vehicles placed in the air. They don't have any spin-up but are at full power from the start. I'm also not quite sure and just added this as we always start our missions from the ground. But I can test is.

addons/vehicles/CfgEventHandlers.hpp
brainslush: Can do but then you should change the other classes from speed limiter as well but in a different PR

addons/vehicles/CfgEventHandlers.hpp
commy2: In any case, setVariable public does not work from Init.

addons/vehicles/CfgEventHandlers.hpp
jonpas: No, that's already released. This isn't. Don't follow bad old examples. :laughing:

addons/vehicles/CfgEventHandlers.hpp
brainslush: I got that and will fix that. Probably just move the locality check to the config.

commy2:
Do we even need setVariable public for this?

I would do the following things: 
- remove the event handler from config into preInit, because config is butt ugly
- change all setVar pub to setVar
- change Init to InitPost
- use engineOn event from common

This should handle the correct state for JIP, because it is either the state when you connect as client, or whicher was last toggled by Engine event. InitPost and Engine are local, so no strain on the network.

addons/vehicles/initSettings.sqf
```diff
@@ -0,0 +1,7 @@
+
+[QGVAR(keepEngineRunning), "CHECKBOX",
+[LSTRING(SettingKeepEngineRunningName), LSTRING(SettingKeepEngineRunningDesc)],
+localize ELSTRING(common,ACEKeybindCategoryVehicles),
+true, // default value
```
Dystopian: is it really necessary to make this by default? You can always exit with `eject` action to keep engine off. With this setting `get out` would became the same as `eject`.

Drofseh: Eject does very different things for aircraft.

jonpas: I tend to agree with @Dystopian unless we also remove the Eject action in ACE3. Then we could have 1 setting that toggles between:
- Vanilla
- Removed Eject action & Disabled automatic shut-off

Dystopian: > Eject does very different things for aircraft.

make it for aircraft only, why force for any vehicles?

Drofseh: I'd be happy to have Eject as an aircraft only action.

brainslush: I can agree on the by default disabled setting. But I don't see why it should be aircraft only. This constant turning-off the engine stuff is a disease in the arma mill-sim world. I can agree on wheeled vehicles with a regular piston engine that you constantly turn them on and off but turning off the gas turbine of a tank is just not that common. Maybe we can agree that we make this a setting as well, either all vehicles or just aircrafts?
Also I see a difference between eject and get out even with the keep the engine running feature. Eject allows you to get out of a moving vehicle, get out doesn't. For someone like me who uses keybinds for both it is crucial difference. The get out keybind is close to the often used keys, the eject is on the far end of the keyboard.

jonpas: @brainslush Eject means ejection seat in aircraft, not just keeping the engine running. I am not sure if you missed that part.

addons/vehicles/CfgEventHandlers.hpp
```diff
@@ -32,4 +32,4 @@ class Extended_Init_EventHandlers {
             exclude[] = {"TargetSoldierBase","Static_Designator_01_base_F","Static_Designator_02_base_F","Pod_Heli_Transport_04_base_F"};
         };
     };
-};
+};
```
commy2:
```suggestion

```

addons/vehicles/initSettings.sqf
```diff
@@ -0,0 +1,10 @@
+
+[
+    QGVAR(keepEngineRunning),
+    "CHECKBOX",
+    [LSTRING(SettingKeepEngineRunningName), LSTRING(SettingKeepEngineRunningDesc)],
+    localize ELSTRING(common,ACEKeybindCategoryVehicles),
+    false, // default value
+    true, // isGlobal
+    {[QGVAR(keepEngineRunning), _this] call EFUNC(common,cbaSettings_settingChanged)}
```
commy2: This does not add the actions when off at mission start and then turned on. `cbaSettings_settingChanged` not correctly implemented I think.

brainslush: Yes, it will not. Question is, should it be changeable mid mission?

commy2: Whenever it is possible settings should be live editable and take effect immediately.

addons/vehicles/stringtable.xml
```diff
@@ -67,5 +67,13 @@
             <Japanese>速度制限を減らす</Japanese>
             <Polish>Zmniejsz ograniczenie prędkości</Polish>
         </Key>
+        <Key ID="STR_ACE_Vehicles_SettingKeepEngineRunningName">
+            <English>Disable automatic engine shut-off</English>
+            <German>Motor nach Verlassen laufen lassen</German>
+        </Key>
+        <Key ID="STR_ACE_Vehicles_SettingKeepEngineRunningDesc">
+            <English>Prevent the automatic shut-off of the engine when exiting vehicles.</English>
+            <German>Verhindere das automatische Abschalten des Motors beim Verlassen des Fahrzeugs</German>
```
commy2:
```suggestion
            <German>Verhindere das automatische Abschalten des Motors beim Verlassen des Fahrzeugs.</German>
```

addons/vehicles/XEH_preInit.sqf
```diff
@@ -6,4 +6,30 @@ PREP_RECOMPILE_START;
...
+] call CBA_fnc_addClassEventHandler;
+[
+    "AllVehicles",
+    "getout",
+    {
+        params ["_vehicle","_position","_unit"];
+        if (!GVAR(keepEngineRunning)) exitWith {};
+        if (_unit == ACE_player && {_position == "driver"} && {_vehicle getVariable [QGVAR(engineState), false]}) then {
+            [{
+                [QEGVAR(common,engineOn), _this, _this select 0] call CBA_fnc_targetEvent;
```
commy2: params

addons/vehicles/initSettings.sqf
commy2: Not needed as far as I can tell.

addons/vehicles/XEH_preInit.sqf
```diff
@@ -6,4 +6,30 @@ PREP_RECOMPILE_START;
...
+
+[
+    "AllVehicles",
+    "initPost",
+    {(_this select 0) setVariable [QGVAR(engineState), isEngineOn (_this select 0)]}
+] call CBA_fnc_addClassEventHandler;
+[
+    "AllVehicles",
+    "engine",
+    {(_this select 0) setVariable [QGVAR(engineState), _this select 1]}
```
commy2: params

addons/vehicles/XEH_preInit.sqf
```diff
@@ -6,4 +6,30 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+#include "initSettings.sqf"
+
+[
+    "AllVehicles",
+    "initPost",
+    {(_this select 0) setVariable [QGVAR(engineState), isEngineOn (_this select 0)]}
```
commy2: params

addons/vehicles/XEH_preInit.sqf
```diff
@@ -6,4 +6,37 @@ PREP_RECOMPILE_START;
...
+    {
+        params ["_vehicle", "_state"];
+        _vehicle setVariable [QGVAR(engineState), _state];
+    }
+] call CBA_fnc_addClassEventHandler;
+[
+    "AllVehicles",
+    "getout",
+    {
+        params ["_vehicle","_position","_unit"];
```
commy2:
```suggestion
        params ["_vehicle", "_position", "_unit"];
```

addons/vehicles/XEH_preInit.sqf
```diff
@@ -6,4 +6,37 @@ PREP_RECOMPILE_START;
...
+        if (!GVAR(keepEngineRunning)) exitWith {};
+        if (_unit == ACE_player && {_position == "driver"} && {_vehicle getVariable [QGVAR(engineState), false]}) then {
+            [{
+                params ["_unit"];
+                [QEGVAR(common,engineOn), _this, _unit] call CBA_fnc_targetEvent;
```
commy2: Variable should probably be `_vehicle` and not `_unit`.

commy2:
```suggestion
                [QEGVAR(common,engineOn), _this, _vehicle] call CBA_fnc_targetEvent;
```

addons/vehicles/XEH_preInit.sqf
```diff
@@ -6,4 +6,37 @@ PREP_RECOMPILE_START;
...
+] call CBA_fnc_addClassEventHandler;
+[
+    "AllVehicles",
+    "getout",
+    {
+        params ["_vehicle","_position","_unit"];
+        if (!GVAR(keepEngineRunning)) exitWith {};
+        if (_unit == ACE_player && {_position == "driver"} && {_vehicle getVariable [QGVAR(engineState), false]}) then {
+            [{
+                params ["_unit"];
```
commy2:
```suggestion
                params ["_vehicle"];
```

addons/vehicles/XEH_preInit.sqf
```diff
@@ -6,4 +6,37 @@ PREP_RECOMPILE_START;
...
+        _vehicle setVariable [QGVAR(engineState), _state];
+    }
+] call CBA_fnc_addClassEventHandler;
+[
+    "AllVehicles",
+    "getout",
+    {
+        params ["_vehicle", "_position", "_unit"];
+        if (!GVAR(keepEngineRunning)) exitWith {};
+        if (_unit == ACE_player && {_position == "driver"} && {_vehicle getVariable [QGVAR(engineState), false]}) then {
```
commy2:
```suggestion
        if (_unit == ACE_player && {_position == "driver" && {_vehicle getVariable [QGVAR(engineState), false]}}) then {
```
optimize

dedmen: isEqualTo on the string? Not sure where the position string comes from and if casing is guaranteed.

commy2: I don't trust BI not the change the capitalization of this.

PR #6890, opened 2019-03-24
Handle old public funcs, fix strings

PabstMirror:


addons/medical/XEH_preInit.sqf
```diff
@@ -8,4 +8,10 @@ PREP_RECOMPILE_END;
 
 #include "initSettings.sqf"
 
+// Add warning for old functions that were technically public, Remove at 3.14.0
+{
+    missionNamespace setVariable [_x, compileFinal format ['diag_log text "ACE_Medical: Warning Formally public function [%1] has no effect in medical rewrite"; nil', _x]];
+} forEach ["ace_medical_fnc_actionPlaceInBodyBag","ace_medical_fnc_actionRemoveTourniquet","ace_medical_fnc_addHeartRateAdjustment","ace_medical_fnc_addToLog","ace_medical_fnc_addToTriageCard
```
mharis001: All on same line or one line per function?

addons/medical/XEH_preInit.sqf
```diff
@@ -8,4 +8,10 @@ PREP_RECOMPILE_END;
 
 #include "initSettings.sqf"
 
+// Add warning for old functions that were technically public, Remove at 3.14.0
+{
+    missionNamespace setVariable [_x, compileFinal format ['diag_log text "ACE_Medical: Warning Formally public function [%1] has no effect in medical rewrite"; nil', _x]];
```
mharis001:
```suggestion
    missionNamespace setVariable [_x, compileFinal format ['diag_log text "ACE Medical WARNING: Formerly public function [%1] has no effect in medical rewrite."; nil', _x]];
```

addons/medical/XEH_preInit.sqf
PabstMirror: word wrap should be fine :smile: 
most of these functions really don't make sense to be called externally, 
but I thought it was better to be safe

PR #6892, opened 2019-03-25
 Cookoff - Reword settings and move to cba

PabstMirror:
```
class GVAR(enableAmmoCookoff) { 
// For CBA Setting Switch: we can eliminate and just use (ammoCookoffDuration == 0)
```
are we ok actually doing this, I think it might have been written back when we though we would lose all saved setting when converting?

### DIFF
--- a/addons/cookoff/ACE_Settings.hpp
+++ b/addons/cookoff/ACE_Settings.hpp
@@ -1,40 +1,18 @@
 
 class ACE_Settings {
     class GVAR(enable) {
-        category = CSTRING(displayName);
-        displayName = CSTRING(enable_name);
-        description = CSTRING(enable_tooltip);
-        value = 0;
-        typeName = "BOOL";
+        movedToSqf = 1;
     };
     class GVAR(enableAmmobox) {
-        category = CSTRING(displayName);
-        displayName = CSTRING(enableBoxCookoff_name);
-        description = CSTRING(enableBoxCookoff_tooltip);
-        value = 1;
-        typeName = "BOOL";
+        movedToSQF = 1;
     };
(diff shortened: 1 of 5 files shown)


jonpas:
> are we ok actually doing this, I think it might have been written back when we though we would lose all saved setting when converting?

I think it's clearer to be kept as it is now. But could also be more confusing, I am not sure.

SilentSpike:
> I think it might have been written back when we though we would lose all saved setting when converting

I believe it may have been me who added that comment and this was the case. Neither strongly for/against keeping it.

PabstMirror:
(not directly related to this PR, just cookoff in general)
It is a shame that a lot of the good turret effects have been disabled by default.
IMHO, the problem was some of the damage handling might be based on realism, but just never felt right to me
```
// Would never blow up with AP rounds (part of this problem is crew didn't take enough damage)
if (_hitpoint in ["hithull", "hitfuel", "#structural"] && {!IS_EXPLOSIVE_AMMO(_ammo)}) then { 
...
// _newDamage had to be very large to do anything
if (_hitpoint in ["hithull", "hitturret", "#structural"] && {_newDamage > 0.8 + random 0.2}) then { 
```

We could add a third option for handle damage, that would use the vanilla damage model, so tanks would die when they would in vanilla, 
but use a quick version of cookoff effects instead of just blowing up

handleDamageNew = {
    if ((_hitpoint in ["hithull", "hitfuel", "#structural"]) && {_damage > 0.89}) then {
        [_veh, true] call FUNC(cook-offX) // quickly add efects and eventually explode
        0.89 // don't exlode yet
    } else {
        _damage
    };
};

PR #6893, opened 2019-03-25
RHS Compat - Add 2 round hellfire to pylon options

### LINKED ISSUE #5957: RHS AH6 Littlebird can't be armed with dual-pylon-Ace3 Hellfire
**Arma 3 Version:** `1.80` (stable)
**CBA Version:** `3.5.0` (stable)
**ACE3 Version:** `3.12.0` (stable)

**Mods:**
```
- CBA_A3
- ace
- ace-rhs-compats
```

**Description:**
- the rhs littlebirds can carry 4 rhs hellfire rockets (two on each side), but with ace hellfire rockets, you only have the option of two (one on each side).
**Steps to reproduce:**
- place rhs littlebird, open attributes and arm with the ace-hellfire.

- Dedicated / Editor (Singleplayer).

**Placed Modules:**
- none.

PabstMirror:
Close #5957

Using RHS model, adds new mag that uses ace's guidance for pylon options RHS choppers.

### DIFF
--- a/addons/hellfire/CfgMagazineWells.hpp
+++ b/addons/hellfire/CfgMagazineWells.hpp
@@ -1,0 +1,4 @@
+class CfgMagazineWells {
+    class GVAR(magwell_K) {};
+    class GVAR(magwell_N) {};
+};
--- a/addons/hellfire/CfgMagazineWells.hpp
+++ b/addons/hellfire/CfgMagazineWells.hpp
@@ -1,4 +1,4 @@
 class CfgMagazineWells {
-    class GVAR(magwell_K) {};
-    class GVAR(magwell_N) {};
+    class GVAR(K) {};
+    class GVAR(N) {};
 };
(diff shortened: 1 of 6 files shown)


addons/hellfire/CfgMagazineWells.hpp
```diff
@@ -0,0 +1,4 @@
+class CfgMagazineWells {
+    class GVAR(magwell_K) {};
```
jonpas: It's in a magwell config and alway use in magwell related entries. CBA also doesn't prepend "magwell". Just "K" or "N" will do ("ace_hellfire_k") unless that's already taken (does that even matter in this case)?

PabstMirror: `ace_hellfire_K` should be fine

dedmen: >unless that's already taken (does that even matter in this case)

No. Doesn't matter if already taken.

optionals/compat_rhs_usf3/CfgMagazineWells.hpp
```diff
@@ -0,0 +1,8 @@
+class CfgMagazineWells {
+    class ace_hellfire_K {
+        GVAR[] = {QGVAR(pylon_mag_2rnd_hellfire_k)};
```
dedmen: Actually.... GVAR[] ?
Does that work? Shouldn't work in my mind.
`ADDON` seems better?

PabstMirror:
I was pleasantly surprised magwells worked fine for both vehicles and pylons

PR #6894, opened 2019-03-26
Interaction - Pass Magazine fixes

PabstMirror:
Use CBA_fnc_compatibleMagazines
Fix nil return in canPassMagazine if player has no mags (`forEach` never ran)
Ensure target canAdd (`private _magToPass = _filteredMags select 0;` could be a 200rnd belt that target can't add, but it's ammo count would be larger than smaller mags that the target could add)

### DIFF
--- a/addons/interaction/functions/fnc_passMagazine.sqf
+++ b/addons/interaction/functions/fnc_passMagazine.sqf
@@ -12,21 +12,24 @@
  * None
  *
  * Example:
- * [_player, _target, "arifle_MX_F"] call ace_interaction_fnc_magToPassazine
+ * [_player, _target, "arifle_MX_F"] call ace_interaction_fnc_passMagazine
  *
  * Public: No
  */
 params ["_player", "_target", "_weapon"];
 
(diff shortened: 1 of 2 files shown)


addons/interaction/functions/fnc_passMagazine.sqf
```diff
@@ -12,21 +12,24 @@
...
  * Public: No
  */
 params ["_player", "_target", "_weapon"];
 
-private _compatibleMags = getArray (configfile >> "CfgWeapons" >> _weapon >> "magazines");
-private _filteredMags = magazinesAmmoFull _player select {(_x select 0) in _compatibleMags && {!(_x select 2)}};
+private _compatibleMags = [_weapon] call CBA_fnc_compatibleMagazines;
+private _filteredMags = magazinesAmmoFull _player select {
+    _x params ["_className", "", "_loaded"];
+    (_className in _compatibleMags) && {!_loaded} && {_target canAdd _className}
```
dedmen: Order by speed? !_loaded is faster to check than `in`

PabstMirror: `!_loaded` will be true most of the time, so I don't think we should bother checking it first

PR #6897, opened 2019-03-26
Use function documentation script

### LINKED ISSUE #6896: Add function documentation script to CI
Add script from #5253 to CI to make sure everyone pushes valid documentation.

@jonpas

jonpas:
- Close #6896 
- Make use of #5253 

TODO:
- [x] Fix all the current errors (1)

TODO (can be split to separate PR):
- [x] Populate and link on wiki
  - Do we keep separate page for this? Where do we put the button to access that page ~~(we'll want to run it for ACEX as well, so 2 buttons)~~?
  - Do we want to include component functions in their respective framework pages instead, like we do with dependencies? Too much clutter that way?


dedmen:
`Argument index 4 does not match listed order`
Leaving out unused arguments that are just sent to the function for convenience, not sure if that should warn.
`Error: Missing "Example" header section`
Do we need examples for private functions? I'm thinking no.

Do we want to include private functions in the public docs?

jonpas:
> Leaving out unused arguments that are just sent to the function for convenience, not sure if that should warn.

Yes, should be written likes o:
`* 4: Name <TYPE> (unused)`

> Do we need examples for private functions? I'm thinking no.

Maybe we can skip that requirement.

> Do we want to include private functions in the public docs?

No, and they are already filtered out when generating the docs (without `--debug`).

jonpas:
56 errors left to fix.

dedmen:
I'll try to fix atleast some errors on weekend.

SilentSpike:
From the looks of things, quite a few of them are functions that should be private

jonpas:
@dedmen did you do anything on this?

jonpas:
Only 18 errors left!

jonpas:
@Dahlgren Please don't force push!

One last error remaining:

> Processing... common/functions/fnc_toBitmask.sqf
    Error: Malformed argument "N: Booleans <ARRAY>"

I think the example in this function is wrong. Or the arguments is wrong and first parameter is actually array.

If former, we should probably just add support for either no number in front of argument or `N: `.

jonpas:
This is finally done!

- All public headers are now in correct format.
- New Functions page is on the wiki, similar to [ACRE2's](https://acre2.idi-systems.com/wiki/frameworks/functions-list) but split by component.
- Function documentation script received some updates (support run from root or subfolder, read `COMPONENT_BEAUTIFIED` for the wiki page title, frontmatter for Jekyll).
- Framework and Functions index pages are now split into 2 columns (bye ugly long list).
- Dependencies List and Documented Functions are now generated in CI before the wiki gets deployed via GitHub Pages, this removes committing changes back to the repository. _Pages via GitHub Actions is in Beta._

PabstMirror (approved):

PR #6909, opened 2019-03-31
Extension changes for Medical Rewrite and Linux compilation

PabstMirror:
~~Track in one PR so we don't need to constantly rebuild dlls~~ No rebuilds required for now anyways.

Contains:
- #6809
- #7171
- #5762


dedmen:
We don't use woundID internally in the dll. So we might aswell just remove that stuff?

PabstMirror:
Merging as is to let linux changes become part of master

PR #6910, opened 2019-03-31
Medical - Improve adjustment calcs / wound blood loss / medications

PabstMirror:
Fix pain not reducing over time (`_unit setVariable [QEGVAR(medical_status,pain)`)

Add helper funcs for adding adjustments
lets us skip `if (_value != 0 && {_maxTimeInSystem > 0}) then {` checking each run, also skip a `max`

fix removal from arrays, as nil didn't work
```
x = [1,2,3];
x set [1, nil];
x = x - [nil]

[1,any,3]
```
Edit:
this PR also
- Adds `GET_WOUND_BLEEDING` getVar,
On any change in wound bleeding (openWound or tourniquite change), we update that var.
This improves the speed of `GET_BLOOD_LOSS` function
It also prevents `isBleeding` from switching to false when entering cardiac arrest.
ToDo: merge into medical_ai PR

- moves `#include "\z\ace\addons\medical_engine\script_macros_medical.hpp"` up, so we can use debug macros in medical_engine instead of changing them in 11 different modules


addons/medical_status/functions/fnc_addPainSuppressionAdjustment.sqf
```diff
@@ -0,0 +1,30 @@
+#include "script_component.hpp"
+/*
+ * Author: BaerMitUmlaut, PabstMirror
+ * Adds a heart rate adjustment that will take effect over time.
```
TheMagnetar: Wrong description

addons/medical_status/functions/fnc_addViscosityAdjustment.sqf
```diff
@@ -0,0 +1,30 @@
+#include "script_component.hpp"
+/*
+ * Author: BaerMitUmlaut, PabstMirror
+ * Adds a heart rate adjustment that will take effect over time.
```
TheMagnetar: Wrong description

addons/medical_vitals/functions/fnc_handleUnitVitals.sqf
```diff
@@ -53,7 +53,7 @@ if (_hemorrhage != GET_HEMORRHAGE(_unit)) then {
 
 private _bloodLoss = GET_BLOOD_LOSS(_unit);
 if (_bloodLoss > 0) then {
-    _unit setVariable [QGVAR(bloodloss), _bloodLoss, _syncValues];
+    _unit setVariable [QGVAR(bloodloss), _bloodLoss, _syncValues]; // ToDo: nothing uses this, remove?
```
TheMagnetar: Maybe it should be used to enter unconscious state due to blood loss?

PabstMirror: `#define GET_BLOOD_LOSS(unit)        ([unit] call EFUNC(medical_status,getBloodLoss))`
that function has fairly complex calcs, and is called in 10 different places
so maybe we could swap it to use the setVar

SilentSpike: I remember this is something I noticed way back too. If I remember right, I determined that that function was probably intended for the vitals loop only and everywhere else was probably meant to be reading the variable.

PabstMirror: See 7a87aad
We can calculate wound blood loss only on events that would change it

addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
```diff
@@ -265,6 +265,7 @@ class GVAR(Actions) {
         icon = QPATHTOEF(medical_gui,ui\surgical_kit.paa);
         category = "advanced";
         items[] = {"ACE_surgicalKit"};
+        condition = "";
```
TheMagnetar: This would case conflict with the actual master branch when building.

addons/medical_treatment/functions/fnc_onMedicationUsage.sqf
```diff
@@ -6,69 +6,69 @@
...
 
-private _usedMeds = _target getVariable [_variable, 0];
-if (_usedMeds >= floor (_maxDosage + round(random(2))) && {_maxDosage >= 1}) then {
-    [QEGVAR(medical,CriticalVitals), _target] call CBA_fnc_localEvent;
+private _overdosedMedications = [];
+
+// Check for overdose from current medication
+private _currentDose = [_target, _className] call _fnc_getMedicationCount;
+if (_currentDose >= floor (_maxDosage + round(random(2))) && {_maxDosage >= 1}) then {
+    TRACE_1("exceded max dose",_currentDose);
```
dedmen:
```suggestion
    TRACE_1("exceeded max dose",_currentDose);
```

PabstMirror:
Before: ace_medical_vitals_Vitals_counter: Average: 0.652908s / 2696 = 0.242177ms
After: ace_medical_vitals_Vitals_counter: Average: 0.609032s / 3422 = 0.177975ms

addons/medical_engine/script_macros_medical.hpp
```diff
@@ -1,3 +1,7 @@
+// #define DEBUG_MODE_FULL
```
TheMagnetar: "Defined" in medical_engine/script_component.hpp

addons/medical_treatment/functions/fnc_handleBandageOpening.sqf
```diff
@@ -79,26 +80,30 @@ if (!_exist) then {
 
 _target setVariable [QEGVAR(medical,bandagedWounds), _bandagedWounds, true];
 
+// _reopeningChance = 1;
```
TheMagnetar: Remove comments

addons/medical_treatment/functions/fnc_onMedicationUsage.sqf
```diff
@@ -6,69 +6,69 @@
...
-    } forEach _allUsedMedication;
+    _x params ["_xMed", "_xLimit"];
+    private _inSystem = [_target, _xMed] call _fnc_getMedicationCount;
+    if (_inSystem> _xLimit) then {
+        _overdosedMedications pushBackUnique _xMed;
+    };
 } forEach _incompatabileMeds;
 
-if (_hasOverDosed > 0) then {
+if ((count _overdosedMedications) > 0) then {
```
TheMagnetar:
```suggestion
if !(_overdosedMedications isEqualTo []) then {
```

addons/medical_engine/script_macros_medical.hpp
PabstMirror: I'd rather keep them in both places
script_macros_medical.hpp lets you enable for all 11 medical modules in one place
but you can still go for individual modules with script_component.hpp

a lot of times you have to balance too much vs too little debuging info

addons/medical_treatment/functions/fnc_handleBandageOpening.sqf
PabstMirror: this just makes it easier to debug wound opening
I'd prefer to keep for now, maybe remove when we get closer to release

PR #6911, opened 2019-03-31
Advanced Fatigue - Solid stamina bar option

Blutze:
- Add CBA option to make stamina bar always non-transparent, regardless of stamina status
- With translation entries
- ~Hopefully~ Definitely off by default
- Fix existing bug where disabling AF and directly restarting a mission wouldn't clear out the AF bar

~Keep in mind that I never managed to get Mikero's set up properly, so I can't compile or test this myself.~ @jokoho48 gave it a quick glance back when I made it, but IIRC he was in a bit of a hurry.

Intent is to show stamina consumption early, allowing players to correct their stance and allowing easier testing of stamina settings - at the moment I have to sometimes walk a few klicks to verify a certain combination is not stamina negative.


addons/advanced_fatigue/ACE_Settings.hpp
```diff
@@ -5,6 +5,9 @@ class ACE_Settings {
     class GVAR(enableStaminaBar) {
         movedToSQF = 1;
     };
+    class GVAR(solidStaminaBar) {
+        movedToSQF = 1;
+    };
```
brainslush: Not needed for new settings.

BaerMitUmlaut:
There's no need to set the fade every time if you don't change it.

Don't really see the point of this PR because if you don't see it you're good but okay.

Blutze:
> There's no need to set the fade every time if you don't change it.

So that function in initSettings.sqf already hides it until I change the setting or restart, and in the handler I just need to stop the calculated fade from applying?

> Don't really see the point of this PR because if you don't see it you're good

When the bar becomes visible, you already lost at least half a block, or ~10% of your stamina. And as I mentioned this is also intended for testing loadout weights and settings - "do I need to drop another mag for non-combatpace running to become stamina neutral" or something along those lines. Might sound overly nerdy to most people, but doing my homework has always been how I am not outright failing at videogames.

PR #6917, opened 2019-04-03
Fix issues in #6880

dedmen:
#6880


extensions/medical/handleDamage.cpp
```diff
@@ -70,7 +70,7 @@ namespace ace {
                         if (amountOfDamage >= possibleInjury->minDamage && (amountOfDamage <= possibleInjury->maxDamage || possibleInjury->maxDamage <= 0))
                         {
                             if (InjuryAllowedOnSelection(possibleInjury->allowedSelections, selectionName)) {
-                                if (highestSpot == NULL) {
+                                if (!highestSpot) {
```
TheMagnetar: please switch it to `nullptr` so that the conditional has a basic boolean type

extensions/medical/handleDamage.cpp
```diff
@@ -80,7 +80,7 @@ namespace ace {
                             }
                         }
                     }
-                    if (highestSpot == NULL) {
+                    if (!highestSpot) {
```
TheMagnetar: same as above.

extensions/medical/handleDamage.cpp
dedmen: shared_ptr already has a boolean operator

TheMagnetar: The more explicit a code is the better...

Issue #6913, opened 2019-04-01
Medical Rewrite - TODO


alganthe:
### Needs Testing:
- [x] **Needs testing**. Add / fix vitals conditions for unconsciousness, currently only damage results in it.

### To Do (Unclaimed):
- [ ]  **?** Unconsciousness mutes sounds in zeus (sound is not reset when entering curator after going unconscious, see [here](https://github.com/acemod/ACE3/blob/master/addons/medical_feedback/XEH_postInit.sqf#L49))
- [ ] **Medium**.  Remove bleedout event in favor of fatal injury event (remove from state machine, too)
- [ ] **Discussion**. More heavy injuries increase condition of the next one being fatal ("stacked damage")
- [ ] **?** Fix people running with 4 tourniquets without consequences.

### Think about:
- [ ] **After Release**. Look into reducing target events with treatments. Reference: https://github.com/acemod/ACE3/pull/6925#issuecomment-480618638
- [ ] Medical Blood - https://github.com/acemod/ACE3/pull/7069#discussion_r297099846 - extra setPos?
- [ ] Should do a final pass once the above are merged and settings are finalized.

### Fixed:
- [x] **Trivial**. Pain / color grading effects are too tame. (#7215)
- [x] Look into AI stop shooting setting (ref #7074) old setting `delayUnconCaptive`
- [x] **?** Bandages and medication are twice as heavy as previous medical (0.09kg vs 0.05kg) (#7200)
- [x] **trivial** Increase carry pickup speed #7216
- [x] Remove Atropine (please) #7212 
- [x] **Critical**. Vitals loop should run at cardiac arrest with a minimum heart rate of ~20 to simulate gravity effects (bleeding, blood bags) - affects CPR (whoever wants to claim this, just work on top of #7013) PR #7060
- [x] **Medium**. Sort medical settings into appropriate subcategories. (PR #7057)
    - AI subcategory will be corrected by #6902 
    - A lot of settings, including GUI, will be corrected by #6933 
- [x] **Medium**. Fix surgical kit treatment condition. (#6916, #7044) 
- [x] Limping Todo: Issue #6940 - PR #7045
- [x] AI does not stop shooting unconscious players (#7074)
- [x] **Critical**. Medications have no effect (including IVs) after #6933 (PR #7048)
- [x] **Critical**. Fix unconscious UI effects in copilot camera / turrets. (#7003)
- [x]  Fix patient unloading failing to put the patient back in unconscious anim on dedicated
[...]

PR #6918, opened 2019-04-03
Medical - Change animation for cpr

Katalam:
- #6913
- Change animation for cpr action


SilentSpike:
If I remember right, I believe we wanted to go ahead with this 👍

The animation isn't perfect, but it's better

PR #6922, opened 2019-04-04
Fix trivial bugs

PabstMirror:
- pass mag (if weapon == "" then CBA_fnc_compatibleMagazines will throw warning
- med-engine - add ace tag to debug line
- overheat, `_energyIncrement ` would be nil on first run because of inner private

### DIFF
--- a/addons/interaction/functions/fnc_canPassMagazine.sqf
+++ b/addons/interaction/functions/fnc_canPassMagazine.sqf
@@ -19,6 +19,7 @@
 params ["_player", "_target", "_weapon"];
 
 if (!GVAR(enableMagazinePassing)) exitWith {false};
+if (_weapon == "") exitWith {false};
 if (((vehicle _target) != _target) && {(vehicle _target) != (vehicle _player)}) exitWith {false};
 
 private _compatibleMags = [_weapon] call CBA_fnc_compatibleMagazines;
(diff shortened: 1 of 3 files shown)


addons/medical_engine/script_component.hpp
```diff
@@ -19,11 +19,11 @@
 #include "\z\ace\addons\medical_engine\script_macros_config.hpp"
 
 #define PRELOAD_CLASS(class) \
-    diag_log format["Starting preload for ""%1""",class];\
+    diag_log text format["[ACE] (medical_engine) Starting preload for (%1)",class];\
```
jonpas:
```suggestion
    INFO_1("Starting preload for (%1)",class);\
```

addons/medical_engine/script_component.hpp
```diff
@@ -19,11 +19,11 @@
...
     [{\
         1 preloadObject _this;\
     }, {\
-        diag_log format ["Preload done for ""%1""",_this];\
+        diag_log text format ["[ACE] (medical_engine) Preload done for (%1)",_this];\
```
jonpas:
```suggestion
        INFO_1("Preload done for (%1)",_this);\
```

addons/interaction/functions/fnc_canPassMagazine.sqf
```diff
@@ -19,6 +19,7 @@
 params ["_player", "_target", "_weapon"];
 
 if (!GVAR(enableMagazinePassing)) exitWith {false};
+if (_weapon == "") exitWith {false};
```
dedmen: isEqualTo :smirk:

commy2:
```suggestion
if (_weapon isEqualTo "") exitWith {false};
```

PR #6925, opened 2019-04-07
Medical activity to non local units is not logged

redbery:
- Fix medical events not appearing in MP (when unit is non local)
- Add my name to AUTHORS.txt (forgot to do this in other PRs)


addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -21,3 +21,6 @@ if (isServer) then {
 [QGVAR(actionCheckPulseLocal), FUNC(actionCheckPulseLocal)] call CBA_fnc_addEventHandler;
 [QGVAR(actionCheckBloodPressureLocal), FUNC(actionCheckBloodPressureLocal)] call CBA_fnc_addEventHandler;
 [QGVAR(actionPlaceInBodyBag), FUNC(actionPlaceInBodyBag)] call CBA_fnc_addEventHandler;
+
+// log events
+[QGVAR(addToMedicalLog), FUNC(addToLog)] call CBA_fnc_addEventHandler;
```
TheMagnetar: EOF

PabstMirror: Should add triageCard as well
```suggestion
[QGVAR(addToMedicalLog), FUNC(addToLog)] call CBA_fnc_addEventHandler;
[QGVAR(addToTriageCard), FUNC(addToTriageCard)] call CBA_fnc_addEventHandler;
```

PabstMirror:
this fixes issue, but we might want to look at how we do events
e.g. ivTreatment does
```
[QGVAR(treatmentIVLocal), [_target, _className, _bodyPart], _target] call CBA_fnc_targetEvent;

[_target, _removeItem] call FUNC(addToTriageCard);
[_target, "activity", ELSTRING(medical_treatment,Activity_gaveIV), [[_caller, false, true] call EFUNC(common,getName)]] call FUNC(addToLog);
[_target, "activity_view", ELSTRING(medical_treatment,Activity_gaveIV), [[_caller, false, true] call EFUNC(common,getName)]] call FUNC(addToLog); // TODO expand message
```
that's 4 targetEvents,
I see no reason to not just move logging inside treatmentIVLocal and then pass _caller as well in a single targetEvent

AUTHORS.txt
```diff
@@ -1,4 +1,4 @@
-# ACE3 CONTRIBUTOR LIST
+﻿# ACE3 CONTRIBUTOR LIST
```
Dahlgren: BOM added, `0xEF0xBB0xBF` which can't be seen

https://en.wikipedia.org/wiki/Byte_order_mark

PR #6932, opened 2019-04-10
Fastroping - Remove FRIES on vehicle deletion / adjust Ghosthawk Fries

veteran29:
- Move FRIES removal logic to separate function
- Remove FRIES when vehicles are deleted
- Adjust FRIES bar position on Ghosthawk to not clip with right door (https://imgur.com/a/s4zAcZ2)
- Mark `equipFRIES` as public function, TODO documentation? Where?


addons/fastroping/XEH_PREP.hpp
```diff
@@ -6,6 +6,7 @@ PREP(canPrepareFRIES);
 PREP(canStowFRIES);
 PREP(checkVehicleThread);
 PREP(cutRopes);
+PREP(deequipFRIES);
```
bux: unequipFRIES

jonpas:
Documentation on Fastroping Framework wiki page.

veteran29:
I assume this can be safely deleted as I've added `deleted` event handler:
https://github.com/acemod/ACE3/blob/master/addons/fastroping/functions/fnc_checkVehicleThread.sqf
?

dedmen:
If we can trust the Deleted handler.. Yes.
In the past we only had problems with the deleted EH firing too often, while testing I didn't find a single occurence of it not firing. So it should be fine

veteran29:
I think I will add `true/false` return to `equipFRIES `, if it is public it should give feedback if fries were added or not. Opinions?

veteran29:
Added docs.

PR #6933, opened 2019-04-11
Medical Treatment - Modernize and Cleanup

mharis001:
- Modernizes and cleans up the medical_treatment component


addons/medical_treatment/XEH_preInit.sqf
```diff
@@ -34,4 +34,23 @@ GVAR(animDurations) = [] call CBA_fnc_createNamespace;
     ["AinvPknlMstpSnonWnonDnon_medic1", 10]
 ];
 
+// class names of medical facilities
+// global variable so it can be accessed by mission makers
+GVAR(facilityClasses) = [
```
dedmen: Wouldn't it make more sense to have the default list in config? So that mods like CUP/RHS can add compat?
Iterating over all vehicles is not an option though. So maybe a seperate config class for these? Maybe in CfgMagazineWells style to make adding onto it easy?

addons/medical_treatment/functions/fnc_addToLog.sqf
```diff
@@ -1,56 +1,45 @@
...
-    {
-        // ensure the first element will not be added
-        if (_forEachIndex > 0) then {
-            _newLog pushBack _x;
-        };
-    } forEach _log;
-
-    _log = _newLog;
+if (count _log >= MAX_LOG_ENTRIES) then {
+    _log deleteAt 0;
```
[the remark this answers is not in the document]
mharis001: Not intended and yes the old code did not but, either behaviour should be fine and I think this is cleaner.

addons/medical_treatment/functions/fnc_litterCleanupLoop.sqf
```diff
@@ -10,31 +10,28 @@
...
 
 {
-    _x params ["_time", "_objects"];
+    _x params ["_object", "_timeCreated"];
 
-    // Older elements are always at the begining of the array
-    if (CBA_missionTime - _time < GVAR(litterCleanUpDelay)) exitWith {};
+    // Litter array has older objects at the beginning
+    // Can exit on first element that still has lifetime remaining
+    if (CBA_missionTime - _timeCreated < GVAR(litterCleanUpDelay)) exitWith {};
```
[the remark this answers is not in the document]
mharis001: Will change to `litterCleanupDelay`.

addons/medical/XEH_preInit.sqf
```diff
@@ -11,7 +11,6 @@ PREP_RECOMPILE_END;
...
...ustment","ace_medical_fnc_addToLog","ace_medical_fnc_addToTriageCard","ace_medical_fnc_addUnconsciousCondition","ace_medical_fnc_addVitalLoop","ace_medical_fnc_canAccessMedicalEquipment","ace_medical_fnc_canTreat","ace_medical_fnc_displayTriageCard","ace_medical_fnc_dropDownTriageCard","ace_medical_fnc_getTriageStatus","ace_medical_fnc_getUnconsciousCondition","ace_medical_fnc_hasItem","ace_medical_fnc_hasItems","ace_medical_fnc_hasTourniquetAppliedTo","ace_medical_fnc_isInMedicalFacility","ace_medical_fnc_isInMedicalVehicle","ace_medical_fnc_isMedic","ace_medical_fnc_isMedicalVehicle","ace_medical_fnc_itemCheck","ace_medical_fnc_selectionNameToNumber","ace_medical_fnc_setCardiacArrest","ace_medical_fnc_setDead","ace_medical_fnc_setHitPointDamage","ace_medical_fnc_showBloodEffect","ace_medical_fnc_treatment","ace_medical_fnc_treatmentAdvanced_bandage","ace_medical_fnc_treatmentAdvanced_CPR","ace_medical_fnc_treatmentAdvanced_CPRLocal","ace_medical_fnc_treatmentAdvanced_medication","ace_medical_fnc_treatmentAdvanced_medicationLocal","ace_medical_fnc_treatmentIV","ace_medical_fnc_treatmentIVLocal","ace_medical_fnc_unconsciousPFH","ace_medical_fnc_useItem","ace_medical_fnc_useItems"];
```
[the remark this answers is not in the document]
SilentSpike: `QFUNC` exists 😃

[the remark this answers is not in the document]
thecogdoctor: Addressed in 1acfff81d7200cf7084155b7cc192845509f709f.

addons/medical_treatment/functions/fnc_litterCleanupLoop.sqf
```diff
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
- * Author: Glowbal, esteldunedain
- * Loop that cleans up litter
+ * Author: Glowbal, esteldunedain, mharis001
+ * Handles cleaning up litter objects that have reached their lifetime.
```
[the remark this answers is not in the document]
thecogdoctor: Addressed in b4f91960ce6054a8a78d0e35823c4ef7d44f5642.

addons/medical_treatment/functions/fnc_litterCleanupLoop.sqf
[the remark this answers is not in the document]
thecogdoctor: Addressed in b4f91960ce6054a8a78d0e35823c4ef7d44f5642.

addons/medical_treatment/functions/fnc_hasTourniquetAppliedTo.sqf
```diff
@@ -1,23 +1,23 @@
...
+ * Checks if the unit has a tourniquet applied on the specified body part.
  *
  * Arguments:
- * 0: The Unit <OBJECT>
+ * 0: Unit <OBJECT>
  * 1: Body Part <STRING>
  *
  * Return Value:
- * Has tourniquet applied <BOOL>
+ * Has Tourniquet Applied <BOOL>
```
thecogdoctor:
```suggestion
 * Has tourniquet been applied <BOOL>
```

PabstMirror:
Fixed merge conflicts with limping: https://github.com/acemod/ACE3/tree/medical-treatment-work-rebaseLimping

PabstMirror:
Because of squash merge stuff the commits are a mess but it shouldn't cause any problems going forward

addons/medical_treatment/XEH_preStart.sqf
```diff
@@ -1,3 +1,11 @@
...
 
 #include "XEH_PREP.hpp"
+
+private _replacementItems = [];
+
+{
+    _replacementItems pushBack [configName _x, getArray _x];
+} forEach configProperties [configFile >> QEGVAR(medical,replacementItems), "isArray _x"];
+
+uiNamespace setVariable [QGVAR(replacementItems), _replacementItems];
```
PabstMirror: need to use `compileFinal`
or I think it would be fine to just move to preInit, this is going to take ~0.1ms to run

addons/medical_treatment/XEH_preInit.sqf
```diff
@@ -34,4 +34,30 @@ GVAR(animDurations) = [] call CBA_fnc_createNamespace;
...
+    "RU_WarfareBFieldhHospital",
+    "USMC_WarfareBFieldhHospital"
+];
+
+// array of medical items to replace and their ACE equivalents
+GVAR(replacementItems) = [];
+
+{
+    GVAR(replacementItems) pushBack [configName _x, getArray _x];
+} forEach configProperties [configFile >> QEGVAR(medical,replacementItems), "isArray _x"];
```
commy2:
```sqf
GVAR(replacementItems) = configProperties [configFile >> QEGVAR(medical,replacementItems), "isArray _x"] apply {
    [configName _x, getArray _x]
};
```
Saves a command and a variable per iteration.

addons/medical_treatment/functions/fnc_getHealTime.sqf
```diff
@@ -0,0 +1,28 @@
...
+
+params ["_patient"];
+
+private _bodyPartDamage = 0;
+
+{
+    _bodyPartDamage = _bodyPartDamage + _x;
+} forEach (_patient getVariable [QEGVAR(medical,bodyPartDamage), []]);
+
+10 max (_bodyPartDamage * DAMAGE_SCALING_FACTOR * GVAR(timeCoefficientPAK)) min 180
```
SilentSpike: Based on the way you've worded the setting, I'd be inclined to make this
```suggestion
10 max (((_bodyPartDamage * DAMAGE_SCALING_FACTOR) min 180) * GVAR(timeCoefficientPAK))
```

addons/medical_treatment/XEH_preInit.sqf
mharis001: We add CfgVehicles config entry and then compile the list at preStart? 
Or for your suggestion, would something like this work?
```cpp
class ACE_Medical_Facilities {
    BI[] = {
        "VanillaMedFac01_F",
        "VanillaMedFac02_F"
    };
    CUP[] = {
        "TK_GUE_WarfareBFieldhHospital_Base_EP1",
        "TK_GUE_WarfareBFieldhHospital_EP1"
    };
};
```

SilentSpike: I'd probably do a separate config like that, same as we're doing for the healing item replacement

addons/medical_treatment/CfgWeapons.hpp
```diff
@@ -1,32 +1,31 @@
...
-        ace_arsenal_hide = 1;
+        EGVAR(arsenal,hide) = 1;
         class ItemInfo: InventoryFirstAidKitItem_Base_F {
             mass = 4;
         };
     };
     class Medikit: ItemCore {
         type = 0;
-        ace_arsenal_hide = 1;
+        EGVAR(arsenal,hide) = 1;
```
SilentSpike: I think we should port over the removal of these entries from #7017

My reasoning is that there are a number of reasons people might still want to add vanilla items to units that will be only replaced if ACE is present (e.g. it's possible to make missions addon compatible but not dependent). Additionally it will be consistent with any other addons that add to the replacement config because they won't be hiding their own items.

addons/medical_treatment/XEH_preInit.sqf
dedmen: Yes that's what I thought about.

Not sure if we really need preStart. The problem with preStart is that we always have to add overhead of compileFinal to make it hacking safe.
preInit would be good enough here.
Or.. We could argue that it's not gameplay relevant enough to be worth hacking, and thus we don't care and don't need to make it compileFinal'ed?

PR #6941, opened 2019-04-13
Medical - Reduce bleeding rates

mharis001:
- Scale back bleeding rates for wounds by 5 times

Should help correct the currently very fast bleed out times with even just one large wound. These values will likely need to be further adjusted as more testing is done.

### DIFF
--- a/addons/medical_damage/ACE_Medical_Injuries.hpp
+++ b/addons/medical_damage/ACE_Medical_Injuries.hpp
@@ -5,10 +5,10 @@ class ACE_Medical_Injuries {
     // Defines all the possible injury types
     class wounds {
         // Source: Scarle
-        //  Also called scrapes, they occur when the skin is rubbed away by friction against another rough surface (e.g. rope burns and skinned knees).
+        // Also called scrapes, they occur when the skin is rubbed away by friction against another rough surface (e.g. rope burns and skinned knees).
         class Abrasion {
             causes[] = {"falling", "ropeburn", "vehiclecrash", "collision", "unknown"};
-            bleeding = 0.001;
+            bleeding = 0.0002;
             pain = 0.4;
             minDamage = 0.01;
             maxDamage = 0.30;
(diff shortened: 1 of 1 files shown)


TheMagnetar:
Ideally real values (publication) should be used to adjust the bleeding rates instead of taking a value that we thing that more or less fits

mharis001:
Agreed. I will try to search for real values however, it is difficult to find reference material.

TheMagnetar:
There is also the maximum blood loss limit depending on the body part. There we can also do some adjustments... I also have something but I must check the authenticity of the data.

severgun:
> Ideally real values (publication) should be used to adjust the bleeding rates instead of taking a value that we thing that more or less fits

I think it will be even faster and no playable

PabstMirror:
I belive this is from 2 things, the swich from total blood from a 0-100 scale to 0-6 (liters)
I think I attempted to solve this by adjusting bleed rates in code:
https://github.com/acemod/ACE3/blob/master/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf#L102-L104
```
private _nastinessLikelihood = linearConversion [0, 20, (_woundDamage / _thresholdWoundCount), 0.5, 30, true];
private _bleedingModifier = 0.25 + 8 * exp ((random [-4.5, -5, -6]) / _nastinessLikelihood);
private _painModifier = 0.05 + 2 * exp (-2 / _nastinessLikelihood);
```
We might want to adjust those values (effecting randomness) as well.

SilentSpike:
If we want to keep the bleeding proportional to how it was on a 0-100 scale in a 0-6 scale these values should be updated as `x/100 * 6`.

I did a quick test of this and it seems somewhat reasonable, unit was stable with a class I hemorrhage with two 6.5mm shots to the body. Third shot killed him.

I think in order to correctly solve this bleeding issue, we first need to choose time periods that are reasonable to survive one of each given wound type and then set the bleeding values and modifier calculations so that they average those time scales.

PabstMirror:
Handled by #7010

Issue #3169, opened 2016-01-16
[optimization] ace medical init advanced and basic JIP variable


havena:
ACE3 Version: 3.4.2
ARMA 3 Version: 1.54.133741

Mods:
@CBA_A3
@ace

Placed ACE3 Modules:
None

Description:
ace_medical init all medical variable and spam JIPqueue on server startup / server network for nothing. This create a huge amazing jipqueue with 100/200 players.
https://github.com/acemod/ACE3/blob/master/addons/medical/functions/fnc_init.sqf

maybe add conditions for init advanced jip var only if ace_medic is set to advanced, and same for basic var.

Steps to reproduce:
use ace_medical basic and look at #exportJIPqueue

exemple : 
---------------------------------------------------------------------------------------------
	Type_272 - id (time = 0.150)
		name (String) = ace_medical_triagelevel
		object (Ref) = creator:7 networkId:142
---------------------------------------------------------------------------------------------
	Type_272 - id (time = 0.150)
		name (String) = ace_medical_fractures
		object (Ref) = creator:7 networkId:142
---------------------------------------------------------------------------------------------
	Type_272 - id (time = 0.150)
		name (String) = ace_medical_lastuniquewoundid
		object (Ref) = creator:7 networkId:142
---------------------------------------------------------------------------------------------
	Type_272 - id (time = 0.150)
		name (String) = ace_medical_internalwounds
		object (Ref) = creator:7 networkId:142
---------------------------------------------------------------------------------------------
	Type_272 - id (time = 0.150)
		name (String) = ace_medical_bandagedwounds
		object (Ref) = creator:7 networkId:142
---------------------------------------------------------------------------------------------
	Type_272 - id (time = 0.150)
		name (String) = ace_medical_openwounds
		object (Ref) = creator:7 networkId:142
---------------------------------------------------------------------------------------------
	Type_272 - id (time = 0.150)
		name (String) = ace_medical_tourniquets
		object (Ref) = creator:7 networkId:142

esteldunedain:
Possible idea for optimization: launch a global `medicalReset` event on unit init, and have each client set all this variables locally. That way only one `publicVariable`is required instead of ~20.

commy2:
^ This has to consider JIP though. Another effect of setVariable public is that the last value set will be set on JIP as well, which wouldn't happen with a global setVariable.

esteldunedain:
> Though, thank you for your report!

[...]

PR #6949, opened 2019-04-18
Medical - Skip unneeded setVars on initUnit

PabstMirror:
For old issue https://github.com/acemod/ACE3/issues/3169

As long as we have good default values for all of our `getVariable` we should be able to skip broadcasting default values on unit init.

e.g. 200 AI * 20 setVars = useless network traffic

### DIFF
--- a/addons/medical_status/CfgEventHandlers.hpp
+++ b/addons/medical_status/CfgEventHandlers.hpp
@@ -19,8 +19,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_Init_EventHandlers {
     class CAManBase {
         class ADDON {
-            onRespawn = 1;
-            init = QUOTE(call FUNC(initUnit));
+            init = QUOTE([ARR_2((_this select 0), false)] call FUNC(initUnit));
         };
     };
 };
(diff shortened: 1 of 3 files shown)


addons/medical_status/CfgEventHandlers.hpp
```diff
@@ -19,8 +19,7 @@ class Extended_PostInit_EventHandlers {
 class Extended_Init_EventHandlers {
     class CAManBase {
         class ADDON {
-            onRespawn = 1;
```
thojkooi: How is respawn handled now?

thojkooi:
I think this breaks respawn handling.

addons/medical_status/CfgEventHandlers.hpp
commy2: initUnit L22

PabstMirror:
On first run it adds the respawn handler
` _unit addEventHandler ["Respawn", {[(_this select 0), true] call FUNC(initUnit)}];`
this should match the same behaviour as XEH's `onRespawn = 1;`

PR #6955, opened 2019-04-21
Disposable - switch NLAW to CBA disposable system and remove ace_disposable functionality

commy2:
- title
- see: https://github.com/CBATeam/CBA_A3/pull/1116


PabstMirror:
to maintain compatibility with CUP, we probably want to go in stages
- move nlaw over to cba
- then remove code after next cupdate

jonpas:
Standard 2 version deprecation sounds the best to me.

commy2:
I don't understand why we need to put the NLAW into CBA.

PabstMirror:
I mean make it compatible with cba_disposable (like how the configs are changed in this PR)
but leave the ace_disposable code for other addons to use

commy2:
All we have to do is to update CBA first, then CUP and at a later point ACE.

commy2:
Merge?

PabstMirror:
this needs to be after cup

PR #6957, opened 2019-04-22
Refuel - Drop nozzle at unit's feet if no valid pos found

### LINKED ISSUE #6944: Problem with fuel nozzle (when it fall uder some objects)
**Arma 3 Version:** 1.90 (stable)
**CBA Version:** 3.9.1 (stable)
**ACE3 Version:** 3.12.5 (stable)

**Mods:**
- CBA_A3
- ace
- TFAR
- RHS USAF, AFRF, GREF
- CUP Core, maps
- Enchanced Movements

**Description:**
- Error in positioning of the fuel nozzle (after it has been dropped) on such things like carrier board , narbour pier, etc..

**Steps to reproduce:**
- It is simple - just take the fuel nozzle , go a few steps and then drop it. 
Then You cannot take it back because it has fallen under the deck (or through the deck)

ALL OF THEM 

**Additional information:**
- Here you 've got some screens ilustrating this issue: 
1) after I have dropped previously taken fuel nozzle (taken from zeus) 
http://markooff.net/obrazki/PGM/bledy_army_ACE_itp/blad_ginaca_wylewka/Arma3_x64_2019-04-15_18-01-26-86.jpg
2) when I try to catch it anyway (using ACE_Interaction Menu - from the first person view)
http://markooff.net/obrazki/PGM/bledy_army_ACE_itp/blad_ginaca_wylewka/Arma3_x64_2019-04-15_18-01-39-63.jpg
3) I cannot bring it back using the ACE_Interaction_Menu on the object (fuel_bladder) 
http://markooff.net/obrazki/PGM/bledy_army_ACE_itp/blad_ginaca_wylewka/Arma3_x64_2019-04-15_18-01-49-36.jpg

The solution could be 
- a bit better positioning of the dropped down fuel nozzle
or
- adding an extra function to the fyel menu - for arbitrally bringing it back (when it is lost or hidden behind any object) 

I've tried to solve the problem using Zeus BIND VALUE + EXECUTE CODE Module and one of ACE3 function like 
[player, fuelTruck] call ace_refuel_fnc_returnNozzle
[player, cursorObject] call ace_refuel_fnc_takeNozzle
bot none of them can help at this time.

PabstMirror:
Fix #6944

### DIFF
--- a/addons/refuel/functions/fnc_dropNozzle.sqf
+++ b/addons/refuel/functions/fnc_dropNozzle.sqf
@@ -29,11 +29,16 @@ _nozzle setVelocity [0, 0, 0];
 
 private _groundPosition = getPosASL _nozzle;
 private _posA = (getPosASL _nozzle) vectorAdd [0,0,0.05];
-private _posB = (getPosASL _nozzle) vectorAdd [0,0,-1000];
+private _posB = (getPosASL _nozzle) vectorAdd [0,0,-4];
 private _intersections = lineIntersectsSurfaces [_posA, _posB, _unit, _nozzle, true, 1, "GEOM"];
 TRACE_1("",_intersections);
 if (_intersections isEqualTo []) then {
-    _groundPosition set [2, (getTerrainHeightASL _groundPosition) + 0.005];
+    WARNING_1("no ground intersections for nozzle drop @ %1",_groundPosition);
+    if (!isNull _unit) then {
+        _groundPosition = getPosASL _unit; // place at unit's feet
+    } else {
+        _groundPosition set [2, (getTerrainHeightASL _groundPosition) + 0.005];
+    };
 } else {
     _groundPosition = ((_intersections select 0) select 0) vectorAdd [0,0,0.005];
 };
--- a/addons/refuel/functions/fnc_dropNozzle.sqf
+++ b/addons/refuel/functions/fnc_dropNozzle.sqf
@@ -29,7 +29,7 @@ _nozzle setVelocity [0, 0, 0];
 
 private _groundPosition = getPosASL _nozzle;
 private _posA = (getPosASL _nozzle) vectorAdd [0,0,0.05];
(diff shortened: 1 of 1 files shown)


addons/refuel/functions/fnc_dropNozzle.sqf
```diff
@@ -29,11 +29,16 @@ _nozzle setVelocity [0, 0, 0];
 
 private _groundPosition = getPosASL _nozzle;
 private _posA = (getPosASL _nozzle) vectorAdd [0,0,0.05];
-private _posB = (getPosASL _nozzle) vectorAdd [0,0,-1000];
+private _posB = (getPosASL _nozzle) vectorAdd [0,0,-4];
```
commy2: Why? Someone could intentionally drop the nozzle from a roof, no?

PabstMirror: If nozzle drops too far then we end up with weird condition where it's further away than the hose length

dedmen: make the vector max hose length in the Z instead of 4m

addons/refuel/functions/fnc_dropNozzle.sqf
```diff
@@ -29,7 +29,7 @@ _nozzle setVelocity [0, 0, 0];
 
 private _groundPosition = getPosASL _nozzle;
 private _posA = (getPosASL _nozzle) vectorAdd [0,0,0.05];
-private _posB = (getPosASL _nozzle) vectorAdd [0,0,-4];
+private _posB = (getPosASL _nozzle) vectorAdd [0,0,-1 * GVAR(hoseLength)];
```
commy2:
```suggestion
private _posB = (getPosASL _nozzle) vectorAdd [0,0,- GVAR(hoseLength)];
```
unary minus

PR #6966, opened 2019-04-27
Medical Feedback - Fix DISABLE_VANILLA_DAMAGE_EFFECTS

PabstMirror:
`DISABLE_VANILLA_DAMAGE_EFFECTS` was only defined in medical_feedback
so it wasn't running, and I'm pretty sure this is where it belongs
it's also now under a `hasInterface` check


PabstMirror:
ai would be handled inside the init:
```
        if (getText (_config >> "simulation") == "UAVPilot") exitWith {TRACE_1("ignore UAV AI",typeOf _unit);};
```
this just skips adding the initEH to AI classes

also pushed a small change to keep the old var name
` _unit setVariable [QEGVAR(medical,lastDamageSource), _shooter];`
to ensure compatiblity with acex_killtracker and possibly other missions that would use this var to determine killer

PR #6976, opened 2019-05-03
Repair - Fix taking wheels from invincible vehicles

### LINKED ISSUE #6923: When disabling the "Enable Damage" switch or using "allowDamage false", you can remove wheels indefinitely from a vehicle
**Arma 3 Version:** 1.90
**CBA Version:** `3.10.1`
**ACE3 Version:** `3.12.6`

**Mods:**
```
CBA_A3
ACE3
ACRE2
RHSAFRF
RHSUSAF
ace_compat_rhs_afrf3
ace_compat_rhs_usf3
```

**Description:**
- One is able to use the Interaction "Remove Wheel" multiple times (more than 4) on that vehicle on the same wheel

- Dedicated 

**RPT log file:**
https://hastebin.com/raw/ayirihizof

commy2:
- fix #6923
- setHit(PointDamage) fails when used on vehilces with damage disabled! The wiki is wrong on this, but they kept reverting my edit because it apparently worked 5 years ago in Arma 2. Just reset after use
- command has the same locality as setHit(PointDamage) so everything should work out.

### DIFF
--- a/addons/repair/XEH_postInit.sqf
+++ b/addons/repair/XEH_postInit.sqf
@@ -6,15 +6,15 @@
 // wheels
 [QGVAR(setWheelHitPointDamage), {
     params ["_object", "_hitPoint", "_damage"];
-    private _isDamageAllowed = isDamageAllowed _object;
+    private _damageDisabled = !isDamageAllowed _object;
 
-    if !(_isDamageAllowed) then {
+    if (_damageDisabled) then {
         _object allowDamage true;
     };
 
     _object setHitPointDamage [_hitPoint, _damage];
 
-    if !(_isDamageAllowed) then {
+    if (_damageDisabled) then {
         _object allowDamage false;
     };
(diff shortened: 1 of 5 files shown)


commy2:
@bux Can you test if this fixes your infinite wheel bug?

addons/repair/XEH_postInit.sqf
```diff
@@ -4,4 +4,17 @@
 [QGVAR(setVehicleHitPointDamage), {_this call FUNC(setHitPointDamage)}] call CBA_fnc_addEventHandler;
 
 // wheels
-[QGVAR(setWheelHitPointDamage), {(_this select 0) setHitPointDamage [_this select 1, _this select 2]}] call CBA_fnc_addEventHandler;
+[QGVAR(setWheelHitPointDamage), {
```
PabstMirror: We could replace use of this event with `QEGVAR(common,setVanillaHitPointDamage)`

PR #6942, opened 2019-04-14
NightVision - Fix NVS blur

kamkill01011:
**When merged this pull request will:**
- Fix #6666 by removing blur in NVS.

Issue #6666, opened 2018-11-08
NVS scope has blur


kamkill01011:
**Arma 3 Version:** `1.84` (stable)
**CBA Version:** `3.9.0` (stable)
**ACE3 Version:** `3.12.3` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- When you aim down the NVS (vanilla scope with integrated night vision) the screen is blurred like when you aim down any normal sight with NVG, however when looking in binoculars with night vision (like the Vector nite) the screen is not blurred; ithink NVS shouldn't be blurred.

Vector : https://i.imgur.com/DPakw0u.jpg
NVS : https://i.imgur.com/QNccjJ2.jpg

**Steps to reproduce:**
- Place a unit, give him the NVS, aim down the sight.

**Where did the issue occur?**
- Dedicated / Editor (Singleplayer)

**Placed Modules:**
- 

**RPT log file:**
-

jonpas:
Same happens with `pilotCamera` cameras in planes and helicopters.

xrufix:
@jonpas `pilotCamera` seems to be fine for me, compare the following screenshots (`ace_nightvision_aimdownsightsblur == 2`):

![20181112085644_1](https://user-images.githubusercontent.com/29610495/48333864-2a3e3d80-e659-11e8-93e5-de600dceae76.jpg)

![20181112085553_1](https://user-images.githubusercontent.com/29610495/48333842-214d6c00-e659-11e8-8a01-476e50890f44.jpg)

PR #6977, opened 2019-05-03
isUsingOptic function to fix night vision optics

commy2:
- Proof of concept function to report if optics mode is used
- `cameraView == "GUNNER"` reports true for CQB optics on top of the optic or weapon, but this really checks for the optics mode and only then reports `true`.
- for #6666 and #6942

```sqf
player addPrimaryWeaponitem "optic_nightstalker";
```

```sqf
call ace_nightvision_fnc_isUsingOptic
```

Obviously only works with comptaible optics, but support for it can be relatively easily added to anything in particular.


Mike-MF:

PR #6979, opened 2019-05-04
Cookoff - Fix shell detonation

### LINKED ISSUE #6978: ace ammo cook off  in vehicles bug (floating rockets)
**Arma 3 Version:*latest* `1.92` (stable)
**CBA Version:** `3.11 latest` (stable)
**ACE3 Version:** `3.12.6 latest` (stable)
[https://imgur.com/oGcNLI0](url)

**Mods:**
- CBA_A3
- ace
- RHS mod pack

**Description:**
- sometimes it happens sometimes not...Random

**Steps to reproduce:**
- Destroy a tank that carry ATGM or rocket or something like that

-  Singleplayer

**Additional information:**
- Provide any additional information that will help us solve this issue.

**RPT log file:**
- If possible at the time the bug is encountered, go to Options and select "ACE Debug To Clipboard", this will print extensive debug information to the RPT file and copy it to clipboard.

TheCandianVendingMachine:
- Fix #6978

ShotShell requires velocity to properly simulate itself, so when detonating ammo always give it velocity. The RHS vehicle example in the issue is just a ShotShell that looks like a missile

Example:
https://youtu.be/GinftkYl5Mo


AirShark07:
Great this must be it !

addons/cookoff/functions/fnc_detonateAmmunition.sqf
```diff
@@ -87,7 +87,7 @@ if (_amountOfMagazines > 0) exitWith {
         playSound3D [_sound, objNull, false, (getPosASL _vehicle), 2, 1, 1300];
 
         if (random 1 < 0.15) then {
-            [_vehicle, _ammo, _speed, random 1 < 0.15] call _spawnProjectile;
+            [_vehicle, _ammo, _speed, true] call _spawnProjectile;
```
dedmen: Don't need the _flyAway parameter in _spawnProjectile if we always set it to true.

commy2: For `shotgrenade` and a few others it is still randomized. See L97ff

SilentSpike:
Only thing missing here is I would have added a comment above this line explaining as much for future reference

PR #6988, opened 2019-05-11
Medical Menu - Hide categories if they don't have any actions

PabstMirror:
Effectively it just hides the "Airway Management" button to reduce confusion

This could also just be done statically in gui.hpp, but this way might be better if we or others implement airway stuff (I think someone has done this?)


SilentSpike:
Hmm, this is an interesting one. Personally I don't think it's necessary to do dynamically like this. If airway management is implemented later then I think it'd be easier to just show/hide it using the setting to enable/disable it.

PabstMirror:
ref https://github.com/Katalam/kat_aceAdvMedical/blob/master/addons/airway/ACE_Medical_Actions.hpp
addon using ace medical that adds `category = "airway";` actions

so I think this makes sense

PR #6994, opened 2019-05-12
Fix volume level of max injury sounds

SilentSpike:
- Pointed out by https://github.com/acemod/ACE3/issues/6452#issuecomment-409697516


commy2:
Interesting. These were exported by script, so it seems to be an error in BI config.

I probably should go through the new voices they added in the last two years :P

SilentSpike:
It's possible that the volume adjustment was right and BI made the sound files **extremely** loud 😆

PR #6995, opened 2019-05-12
Treatment - cap max animation speed

PabstMirror:
part of #6913

caps max anim speed to 3x, which still seems fast but not terrible
and add lock to prevent advanced fatigue from reseting setAnimSpeedCoef

needs mp testing, but I don't think we need global events for the setters (AF doesn't use it)


PabstMirror:
in mp `setAnimSpeedCoef` seems to have global effects

PabstMirror:
On second look, the animations visually look better with the command run on all machines

PR #6708, opened 2018-11-30
SACLOS and HOT Missiles

TheCandianVendingMachine:
**When merged this pull request will:**
- Implement SACLOS simulation alongside HOT missiles

The system works by calculating the weapons direction and multiplying that by the projectiles distance. This creates an "aim point" in front of the missile that represents the cross-hair.

In the attack profile the missile calculates the equivalent to what the real systems electronic impulse would be, and off of that it adjusts itself to move toward the simulated cross-hair. Eventually the missile will be able to slow its movement enough so that it stays on the cross-hair for the duration of the flight

Features:
 SACLOS system
 HOT 1/2/3 Missiles
 Wiesel AWC (ATGM) converted to carry HOT Missiles
 Addition of HOT missile pylons for helicopters
 Wire snapping when missile goes out of range

Overview of system: https://youtu.be/P_m86USv0V0
Helicopter Firing: https://youtu.be/nwwnVtKyJyg
Reference Video: https://youtu.be/7bzHFioJbrc

PR #6997, opened 2019-05-12
Implement Metis SACLOS guidance

TheCandianVendingMachine:
- Implement SACLOS guidance for default Vorona missile
- Move SACLOS and Wire guidance to Missile Guidance framework


addons/missileguidance/script_component.hpp
```diff
@@ -16,3 +16,14 @@
 #endif
 
 #include "\z\ace\addons\main\script_macros.hpp"
+
+#define RANDOM_VECTOR_3D (call {\
+    private _z = random 2 - 1;\
+    private _r = sqrt (1 - _z^2);\
+    private _theta = random 360;\
+    [_r * cos _theta, _r * sin _theta, _z]\
+})
```
commy2: Archimedes!

addons/missileguidance/functions/fnc_SACLOS_onFired.sqf
```diff
@@ -0,0 +1,35 @@
...
+ * Public: No
+ */
+params ["_firedEH", "", "", "", "_stateParams"];
+_firedEH params ["_shooter","_weapon","","","","","_projectile"];
+_stateParams params ["", "_seekerStateParams"];
+
+private _config = ([_projectile] call CBA_fnc_getObjectConfig) >> "ace_missileguidance";
+private _distanceAheadOfMissile = [_config >> "missileLeadDistance", "NUMBER", DEFAULT_LEAD_DISTANCE] call CBA_fnc_getConfigEntry;
+
+if (_shooter isKindOf "Plane") then { WARNING("SACLOS fired from planes unsupported"); };
```
TheMagnetar: if unsupported, should then the function exit?

addons/missileguidance/script_component.hpp
TheCandianVendingMachine: What do you mean?

commy2: This random distribution function is based on the fact that a sphere and a cylinder with height = radius have the same surface area. Something that Archimedes figured out.

TheCandianVendingMachine: Well don't give me the credit for coming up with it, give it to yourself :P https://github.com/acemod/ACE3/pull/6708#discussion_r237926302

commy2: Hmm, maybe I should create it as a CBA function. It seems to be used in multiple places now.

SilentSpike: @commy2 Create it as a function, but generalise it to `n` dimensions ;)

PR #7003, opened 2019-05-17
Medical - Exit turret gunner view when going unconscious

PabstMirror:
Was this the only problem with turrets?

#6913 
>Fix unconscious UI effects in copilot camera / turrets.


alganthe:
> Was this the only problem with turrets?

As far as i'm aware yes.

Issue #6906, opened 2019-03-30
look into bleeding rate and pain in rewrite


PabstMirror:

alganthe:
A large avulsion or velocity wound to the leg results in death after exactly 40s, it's extreme and results in an insane casualty rate in actual gameplay, it definitely needs to be toned down.

SilentSpike:
Commented on #6941 for how I think we can definitively solve the bloodloss issues.

SilentSpike:
As I commented on PR:

>I think in order to correctly solve this bleeding issue, we first need to choose time periods that are reasonable to survive one of each given wound type and then set the bleeding values and modifier calculations so that they average those time scales.

Following the bleeding calculations through, the critical path to death is:

1. Wound bleeding is updated upon wounding/bandaging/etc. This is where bleeding is scaled per body part and a maximum value of 1 is possible (so it's a percentage). See `medical_status_fnc_updateWoundBloodLoss`.
2. Bloodloss is calculated as a product of wound bleeding, cardiac output and the bleeding coefficient setting (this must be litres/second for following calculations to work). If heart rate is normal, cardiac output has a maximum value of 5.25 when blood volume is default 6 litres and this decreases with blood volume. See `medical_status_fnc_getBloodLoss` and `medical_status_fnc_getCardiacOutput`.
3. Unit vitals loop periodically calculates the change in blood volume as a product of time between loops and bloodloss value (hence should be litres). This is assuming the absence of bloodbags, they're factored in here. See `medical_status_fnc_getBloodVolumeChange`.
4. If blood volume is less than 50% unit will die via bleeding out.

With this, we can take the value of a given wound type and follow through the calculations (assuming default heart rate, no bloodbags, and wound is applied to torso) to figure out how long it would take to cause bleedout.

SilentSpike:
Based on the calculations, I can produce the following plot which shows the average time it would take to bleedout for each wound type when applied to the torso (assuming wound is fully open, heart rate is normal throughout, bleeding coefficient setting is 1, time between vital loop iterations is 1s, nastiness modifier is taken as worst case average of `0.25 + 8 * exp(-5/0.5)`).

![image](https://user-images.githubusercontent.com/5459452/57969396-5a4efc00-796e-11e9-964d-54cee9b32f3b.png)

When adjusted for the values in PR #6941 these become

![image](https://user-images.githubusercontent.com/5459452/57969501-7dc67680-796f-11e9-8a47-8e140d738bb9.png)

[...]

PR #7010, opened 2019-05-20
Medical - Fix severity of wound bleeding and adjust cardiac output calculations

SilentSpike:
My attempt at fixing the bleeding in current medical rewrite. If we go this route it will deprecate #6941. My researching steps before implementing these changes can be found on #6906.

I'm simplifying the nastiness calculations so that the wound config specifies the worst wound and we scale it between 25% to 100% based on the wound damage and number of wounds received. Worth noting that the damage value is larger when limbs are damaged, so I'm scaling wounds to those half as much.

Similarly I've updated the wound configs to more reasonable maximum bleeding values based on the fact that they're percentages of cardiac output being bled.

~~In my testing, these values tend to give a bleeding of roughly 0.075 for a typical single gunshot wound which translates to approximately 5 minutes before entering cardiac arrest.~~

Unrelated to this PR, something I noticed during testing:
- If you shoot a unit in the face with a tank, it actually produces a nicer wound than a gunshot. I think this is due to the way we're processing incoming damage not recognising the severity of tank shells.

### DIFF
--- a/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
+++ b/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
@@ -78,7 +78,7 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
     if (_damage > _thresholdMinDam) exitWith {
         private _woundDamage = _damage / (_thresholdWoundCount max 1); // If the damage creates multiple wounds
         for "_i" from 1 to _thresholdWoundCount do {
-            // Find the injury we are going to add. Format [ classID, allowdSelections, bleedingRate, injuryPain]
+            // Find the injury we are going to add. Format [ classID, allowedSelections, bleedingRate, injuryPain]
             private _oldInjury = if (random 1 < 0.15) then {
                 _woundTypes select _highestPossibleSpot
             } else {
--- a/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
+++ b/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
@@ -93,14 +93,16 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
             _bodyPartVisParams set [[1,2,3,3,4,4] select _bodyPartNToAdd, true]; // Mark the body part index needs updating
 
 
-            // Config specifies bleeding and pain for worst possible wound
-            // Worse wound correlates to higher damage, damage is not capped at 1
             // Damage to limbs is scaled higher than body/torso by engine
             // Anything above this is guaranteed worst wound possible
-            private _worstDamage = [2, 4] select (_bodyPartNToAdd > 1);
+            private _worstDamage = [1.8, 4] select (_bodyPartNToAdd > 1);
 
             // More wounds means more likely to get nasty wound
-            private _bleedModifier = linearConversion [0.1, _worstDamage, _woundDamage * _i, 0.25, 1, true];
+            private _countModifier = 1 + random(_i - 1);
+
+            // Config specifies bleeding and pain for worst possible wound
+            // Worse wound correlates to higher damage, damage is not capped at 1
+            private _bleedModifier = linearConversion [0.1, _worstDamage, _woundDamage * _countModifier, 0.25, 1, true];
             private _painModifier = (_bleedModifier * random [0.7, 1, 1.3]) min 1; // Pain isn't directly scaled to bleeding
 
             private _bleeding = _injuryBleedingRate * _bleedModifier;
--- a/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
+++ b/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
@@ -93,9 +93,9 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
             _bodyPartVisParams set [[1,2,3,3,4,4] select _bodyPartNToAdd, true]; // Mark the body part index needs updating
 
 
-            // Damage to limbs is scaled higher than body/torso by engine
-            // Anything above this is guaranteed worst wound possible
-            private _worstDamage = [1.8, 4] select (_bodyPartNToAdd > 1);
+            // Damage to limbs/head is scaled higher than torso by engine
+            // Anything above this value is guaranteed worst wound possible
+            private _worstDamage = [2, 1, 4, 4, 4, 4] select _bodyPartNToAdd;
 
             // More wounds means more likely to get nasty wound
             private _countModifier = 1 + random(_i - 1);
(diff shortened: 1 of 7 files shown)


addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -93,17 +93,22 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
...
-            private _painModifier = 0.05 + 2 * exp (-2 / _nastinessLikelihood);
+            // Config specifies bleeding and pain for worst possible wound
+            // Worse wound correlates to higher damage, damage is not capped at 1
+            // Damage to limbs is scaled higher than body/torso by engine
+            // Anything above this is guaranteed worst wound possible
+            private _worstDamage = [2, 4] select (_bodyPartNToAdd > 1);
 
-            private _bleeding = _injuryBleedingRate * _bleedingModifier;
+            // More wounds means more likely to get nasty wound
+            private _bleedModifier = linearConversion [0.1, _worstDamage, _woundDamage * _i, 0.25, 1, true];
```
SilentSpike: Something we may want to change here is
```suggestion
            private _bleedModifier = linearConversion [0.1, _worstDamage, _woundDamage * (1 + random (_i - 1)), 0.25, 1, true];
```

which just makes it so that beyond 1 wound there's an increased *chance* for more nasty wounds, rather than steadily increasing nastiness. This would mean if the wound damage is 0.66 there a chance you can get 3 medium wounds best case, or 1 medium wound and 2 large wounds worst case. Whereas currently it would result in guaranteed 2 medium wounds and 1 large wound.

SilentSpike: Something else to consider is we could remove the upper limit on the linear conversion. This would mean wound damage values greater than the `worstDamage` can actually cause wounds which bleed more than the configs.

SilentSpike:
After some internal discussion about the cardiac output calculation, I've implemented a new calculation based on just looking at the units of the quantities and the definition of cardiac output.

I've used an assumption that reduced blood volume results in less blood entering the heart with each beat (or stroke) and an approximate relation via `linearConversion`. Perhaps we'd want to make the relation nonlinear.

SilentSpike:
I decided to leave the ventricle volume the same and test this out with the change to heart rate skyrocketing (because it's much more significant). Very happy with the results here, if anything blood loss might even be a little on the slow side now (we could tone down the blood volume scaling in cardiac output a bit if we want to change that).

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -93,17 +93,24 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
...
-            private _nastinessLikelihood = linearConversion [0, 20, (_woundDamage / _thresholdWoundCount), 0.5, 30, true];
-            private _bleedingModifier = 0.25 + 8 * exp ((random [-4.5, -5, -6]) / _nastinessLikelihood);
-            private _painModifier = 0.05 + 2 * exp (-2 / _nastinessLikelihood);
+            // Damage to limbs/head is scaled higher than torso by engine
+            // Anything above this value is guaranteed worst wound possible
+            private _worstDamage = [2, 1, 4, 4, 4, 4] select _bodyPartNToAdd;
 
-            private _bleeding = _injuryBleedingRate * _bleedingModifier;
+            // More wounds means more likely to get nasty wound
+            private _countModifier = 1 + random(_i - 1);
```
PabstMirror: I'm not sure I understand this line, if I test multi hit wound like
`[cursorTarget, 0.4, "body", "shell", player] call ace_medical_fnc_addDamageToUnit`
I get
```
test: _i=1, _countModifier=1 z\ace\addons\medical_damage\functions\fnc_woundsHandlerSQF.sqf:102
test: _i=2, _countModifier=1.44389 z\ace\addons\medical_damage\functions\fnc_woundsHandlerSQF.sqf:102
test: _i=3, _countModifier=1.91001 z\ace\addons\medical_damage\functions\fnc_woundsHandlerSQF.sqf:102
test: _i=4, _countModifier=3.46463 z\ace\addons\medical_damage\functions\fnc_woundsHandlerSQF.sqf:102
test: _i=5, _countModifier=2.4834 z\ace\addons\medical_damage\functions\fnc_woundsHandlerSQF.sqf:102
```
if we create 5 random wounds, why do the later ones bleed more?

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -75,11 +75,11 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
 
 {
     _x params ["_thresholdMinDam", "_thresholdWoundCount"];
-    if (_thresholdMinDam <= _damage) exitWith {
+    if (_damage > _thresholdMinDam) exitWith {
         private _woundDamage = _damage / (_thresholdWoundCount max 1); // If the damage creates multiple wounds
         for "_i" from 1 to _thresholdWoundCount do {
             // Find the injury we are going to add. Format [ classID, allowdSelections, bleedingRate, injuryPain]
```
PabstMirror:
```suggestion
            // Find the injury we are going to add. Format [ classID, allowedSelections, bleedingRate, injuryPain]
```

PabstMirror:
master 
master

pr 
pr

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
SilentSpike: It's basically just making it so that if you receive multiple wounds all at once, subsequent wounds have a higher chance of being worse.

addons/medical_status/functions/fnc_getCardiacOutput.sqf
```diff
@@ -20,13 +20,18 @@
...
 params ["_unit"];
 
-private _bloodVolumeRatio = (GET_BLOOD_VOLUME(_unit) / DEFAULT_BLOOD_VOLUME);
+private _bloodVolumeRatio = GET_BLOOD_VOLUME(_unit) / DEFAULT_BLOOD_VOLUME;
 private _heartRate = GET_HEART_RATE(_unit);
-private _cardiacOutput = ((_bloodVolumeRatio / MODIFIER_CARDIAC_OUTPUT) + ((_heartRate / DEFAULT_HEART_RATE) - 1)) / 60;
 
-(0 max _cardiacOutput)
+// Blood volume ratio dictates how much is entering the ventricle (this is an approximation)
+private _entering = linearConversion [0.5, 1, _bloodVolumeRatio, 0, 1, true];
```
PabstMirror: I'm worried this will block CO and bleeding when blood gets too low (e.g. CPR)

PabstMirror:
In my opinion, blood should effect HR between 4 and 5 to make the ramp up more continuous
Merging for now, as these are things we can tweak later

SilentSpike:
@PabstMirror Yeah after seeing those graphs I've been thinking the same thing

addons/medical_status/functions/fnc_getCardiacOutput.sqf
SilentSpike: I set the conversion based on the bleedout condition at 50% volume so it should never become 0 currently. We will presumably need to change it when dealing with the vitals loop during cardiac arrest stuff (if we remove that death condition).

PR #7013, opened 2019-05-21
Run vitals loop during cardiac arrest

### LINKED ISSUE #6530: time delta is incorrectly large when exiting cardiac arrest
Issue revealed and discussed as part of #6523

Basically because vitals loop isn't ran as part of the cardiac arrest state, when you recover the time delta becomes large because the previously stored time was back before you entered cardiac arrest. This leads to some values trying to incorrectly change massively.

I see two potential solutions:
1. Run the vitals loop while in cardiac arrest (I think we might need to do this anyway because it looks like it updates some values unrelated to the heart rate)
2. Update the stored `GVAR(lastTimeUpdated)` value on the unit when exiting cardiac arrest

SilentSpike:
The bleedout condition will never occur currently since bleeding stops during cardiac arrest unless CPR is administered and the vitals loop doesn't even run so the condition is never checked.

This fixes that by running the loop during cardiac arrest and also allows bleeding to continue if heart rate is 0, just at a fixed reduced rate.

Additionally this fixes #6530

### DIFF
--- a/addons/medical_statemachine/functions/fnc_leftStateCardiacArrest.sqf
+++ b/addons/medical_statemachine/functions/fnc_leftStateCardiacArrest.sqf
@@ -21,7 +21,7 @@ params ["_unit"];
 _unit setVariable [QGVAR(cardiacArrestTime), nil];
 _unit setVariable [QEGVAR(medical,cardiacArrestStart), nil];
 
-// Temporary fix for vitals loop on cardiac arrest exit
-_unit setVariable [QGVAR(lastTimeUpdated), CBA_missionTime];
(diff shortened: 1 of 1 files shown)


SilentSpike:
I think I want to fix this by having the loop running and then making bleeding continue to occur but at quarter speed or something if heart rate is 0.

Need to investigate how CPR can continue to work if the loop is still running (either having it act as a heart rate modifier or just skip heart rate updates if in cardiac arrest).

Also need to prevent the status change events from firing a bunch (e.g. low blood puts you into cardiac arrest and because the loop keeps running the event keeps firing).

Issue #6744, opened 2018-12-22
Titan launcher can lock on through IR Smoke


Mezilsa:
**Arma 3 Version:** `1.84` (stable)
**CBA Version:** `3.x.x` (stable)
**ACE3 Version:** `3.12.4` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- The Titan-Launcher can lock onto target behind IR-Smoke. As the target isn't visible it shouldn't be lock-able.

**Steps to reproduce:**
- Place a unit with a Titan
- Place a vehicle
- Throw a smoke in front of the vehicle
- Take your launcher, switch to IR and try to lock on
- Watch the missle fly straight through the smoke

**Where did the issue occur?**
-Editor(Singleplayer)

**Additional information:**
- https://i.imgur.com/nRxUGVo.jpg
- I assume this is a problem with the lineIntersect command used to find a target. Maybe to scan for smokesheels within 20 meter to the target would prove a viable fix for the problem.

houndog129:
That's not a bug, it's a fairly common misconception but infrared cameras can see thermal signatures through smoke, it may have reduced effectiveness but it still works the same. That's why firefighters can use infrared cameras in a building filled with smoke to find victims.

PabstMirror:
Some vehicle smoke grenades will be opaque in infrared, I think they use some nasty stuff like brass powder and graphite (don't breath this).
I think `checkVisibility` might be able to be used to check smoke, but probably can't differentiate between infantry smoke and IR smoke.

PR #7014, opened 2019-05-22
Missileguidance - Use checkVisibility to test LOS for missile seekers

### LINKED ISSUE #7012: AT-Counter Measures not effective (in vanilla game it looks more effective)
**Arma 3 Version:** `1.92` (stable)
**CBA Version:** `3.10.1.190316` (stable)
**ACE3 Version:** `3.12.6` (stable)

**Mods:**
```
- CBA_A3
- ace
- JSRS sound mod
```

**Description:**

At first I thought it was a problem in the vanilla game. However, when I turned off ACE3, I saw that it worked more effective in a vanilla game. 

**When I try to protect my vehicle from Guided AT-Missiles, I launch a smoke screen, however this does not help me. Always and always the missile will find my tank or APC and I will be destroyed.** 

I tried to run a smoke screen and change the position of my vehicle. In this case, the AT-missiles should lose the target and continue fly to the place where my vehicle was originally, but the always and always AT-missile still finds my vehicle, even if I shifted the location! I tried dozens of times, but it seems to me that the effectiveness of such counter measures is zero. Please make counter measures effective.

**Steps to reproduce:**
 Try defend your vehicle (Merkava Tank or Badger APC) from Guided AT-Misseles:

_a) Missile Specialist (man) - Titan Launcher
b) APC AT-Missile (OPFOR APC - BM 2T Stalker) AT-Missile (Mini-Spike)
c) Independent Jet (JAS 39 Gripen) AT-Missile (Maverick)
d) OPFOR Helicopter Mi-48 Kaiman AT-Missile (9k 121 Vikhr)_

repeat test without ACE3

- Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

BaerMitUmlaut:
- Make missiles using optical guidance (for example the Javelin) now lose lock once the target gets behind smoke

0.001 as a visibility limit seemed reasonable with my testing with a smoke grenade, I'd rather go higher (missiles would then easier lose lock) than lower. If a tank pops a smoke wall, it should be _well_ within that limit anyways.

Not tested with actually firing a missile yet, since the tank would have to pop smoke at just the right time and also be moving, because the missile will still fly straight if it loses lock.

Closes #7012, also ref. #6744 which still needs additonal code when actually locking on to fix completely.

### DIFF
--- a/addons/missileguidance/functions/fnc_checkLos.sqf
+++ b/addons/missileguidance/functions/fnc_checkLos.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: jaynus
- * Returns whether the seeker object can see the target position with lineIntersect
+ * Returns whether the seeker object can see the target position with checkVisibility
  *
  * Arguments:
  * 0: Seeker <OBJECT>
@@ -18,22 +18,5 @@
 
 params ["_seeker", "_target"];
 
-if ((isNil "_seeker") || {isNil "_target"}) exitWith {
-    ERROR_2("nil",_seeker,_target);
-    false
-};
-
-private _targetPos = getPosASL _target;
-private _targetAimPos = aimPos _target;
-private _seekerPos = getPosASL _seeker;
-private _return = true;
-
-if (!((terrainIntersectASL [_seekerPos, _targetPos]) && {terrainIntersectASL [_seekerPos, _targetAimPos]})) then {
-    if (lineIntersects [_seekerPos, _targetPos, _seeker, _target]) then {
-        _return = false;
-    };
-} else {
-    _return = false;
-};
-
-_return;
+private _visibility = [_seeker, "VIEW", _target] checkVisibility [getPosASL _seeker, aimPos _target];
+_visibility > 0.001


Drofseh:
Most thermal sensors see through smoke, unless that smoke is specifically designed to be opaque to thermal EM.

severgun:
> unless that smoke is specifically designed to be opaque to thermal EM.

And all of vehicle smokes are such kind of. Isn't it?

Drofseh:
Only the more modern ones, since around the mid-80's.

BaerMitUmlaut:
In Arma all smoke is thermally opaque afaik. I don't know any optically guided fire and forget ATGMs that were developed before the 80s either.

Drofseh:
Right, but you could be firing a modern weapon at an older tank.
Say a Javelin at a T-55.

jonpas:
Could add a config entry to at least give the option to change it per tank/missile.

TheCandianVendingMachine:
I don't think that's feasible. Some missiles don't lock on to targets but would still be affected by thermal smoke

BaerMitUmlaut:
@jonpas no, you couldn't. How would you know that the smoke that's blocking LOS was caused by that vehicle?

Again, since all smoke is thermally opaque in Arma and you'd have the same problem in vanilla, I'd rather go for this option. The T-55 might have been equipped with newer smoke ammo after all.

jonpas:
Yes, I tend to agree. This is the best we can do at the moment.

PabstMirror:
We could check normal `lis` and `checkVisibility`, if checkVis is blocked then check for presence of vehicle based smoke (`SmokeShellVehicle`)

I guess to be perfect you would have to check for smoke at multiple points on a line from missile to target, but missile should be up high and I think just checking at target is fine.

Vanilla/CUP/RHS both seem to be using `SmokeLauncherMag`, which is a scripted ammo that launches `SmokeShellVehicle`

```
params ["_seeker", "_target"];

if (!((lineIntersectsSurfaces [getPosASL _seeker, aimPos _target, _seeker, _target]) isEqualTo [])) exitWith { false };

private _visibility = [_seeker, "VIEW", _target] checkVisibility [getPosASL _seeker, aimPos _target];
if (_visibility > 0.001) exitWith { true };

// Check if visibility is blocked due to vehicle based smoke launcher
// Note: bis_fnc_effectfiredsmokelauncher spawns SmokeShellVehicle, but particle effects are scripted?
private _shells = _target nearObjects ["SmokeShellVehicle", 30];
_shells isEqualTo [] // return
```

BaerMitUmlaut:
`lis` is not necessary, checkVisibility takes care of that and will do multiple ray casts.
What if visibility is low because a bush, but there is no vehicle smoke involved? This would also require special magazines for non-thermal blocking smoke. Really not a fan of making this more complex than necessary.

Gwynblade:
Speaking of thermally opaque smoke, is it possible to make the smoke not be thermally opaque?
Having a variant of smoke grenades that don't block thermals would add an interesting gameplay mechanic (for infantry smokes I mean).

I wanted to post this in feature requests ages ago, but figured something like this is probably not possible.

PR #7017, opened 2019-05-22
Unblacklist vanilla medical items from arsenal

### LINKED ISSUE #6971: Handling of custom FirstAidKits and MedKits
**Arma 3 Version:** `1.92` (stable / rc / dev)
**CBA Version:** `3.10.1` (stable / dev + commit hash)
**ACE3 Version:** `3.12.6` (stable / dev + commit hash)

**Mods:**
```
- Global Mobilization
- CBA_A3
- ace
```

**Description:**
- GM has two new FAK's, gm_ge_army_burnBandage & gm_ge_army_gauzeBandage. These need to be changed into medical supplies

**Steps to reproduce:**
- Load GM, go into campaign or play as a unit with vanilla loadout.

**Additional information:**
`while {({_x == "gm_ge_army_burnBandage"} count items _unit) > 0} do {
    _unit removeItem "gm_ge_army_burnBandage";
    if (GVAR(convertItems) == 0) then {
        if (GVAR(level) >= 2) then {
            _unit addItem "ACE_fieldDressing";
            _unit addItem "ACE_packingBandage";
            _unit addItem "ACE_morphine";
            _unit addItem "ACE_tourniquet";
        } else {
            _unit addItem "ACE_fieldDressing";
            _unit addItem "ACE_fieldDressing";
            _unit addItem "ACE_morphine";
        };
    };
};`
`while {({_x == "gm_ge_army_gauzeBandage"} count items _unit) > 0} do {
    _unit removeItem "gm_ge_army_gauzeBandage";
    if (GVAR(convertItems) == 0) then {
        if (GVAR(level) >= 2) then {
            _unit addItem "ACE_fieldDressing";
            _unit addItem "ACE_packingBandage";
            _unit addItem "ACE_morphine";
            _unit addItem "ACE_tourniquet";
        } else {
            _unit addItem "ACE_fieldDressing";
            _unit addItem "ACE_fieldDressing";
            _unit addItem "ACE_morphine";
        };
    };
};` https://github.com/acemod/ACE3/blob/v3.12.6/addons/medical/functions/fnc_itemCheck.sqf

SilentSpike:
- The replacement setting may be disabled so we shouldn't prevent their addition
- If it isn't, they will be replaced with suitable items which is fine (and someone may want to design a mission that works either way)

### DIFF
--- a/addons/medical_treatment/CfgWeapons.hpp
+++ b/addons/medical_treatment/CfgWeapons.hpp
@@ -8,14 +8,12 @@ class CfgWeapons {
 
     class FirstAidKit: ItemCore {
         type = 0;
-        EGVAR(arsenal,hide) = 1;
         class ItemInfo: InventoryFirstAidKitItem_Base_F {
             mass = 4;
         };
     };
     class Medikit: ItemCore {
         type = 0;
-        EGVAR(arsenal,hide) = 1;
         class ItemInfo: MedikitItem {
             mass = 60;
         };


addons/medical_treatment/XEH_preStart.sqf
```diff
@@ -1,3 +1,12 @@
...
 #include "XEH_PREP.hpp"
+
+// Cache configured item conversions
+private _itemHash = [] call CBA_fnc_hashCreate;
+
+{
+    [_itemHash, configName _x, getArray _x] call CBA_fnc_hashSet;
+} forEach configProperties [configFile >> QGVAR(replaceItems)];
+
+uiNamespace setVariable [QGVAR(replaceItems), _itemHash];
```
dedmen: Might throw the security argument here.
Actually yes, I think that's relevant. Manipulating that array you can also inject things like "replace bandage with a M134 minigun"
You can replace any item in your inventory with anything. Def needs compileFinal.

addons/medical_treatment/ACE_Medical_ReplaceItems.hpp
```diff
@@ -0,0 +1,17 @@
+class GVAR(replaceItems) {
```
dedmen: Don't like the idea of making root-level classes. Would prefer if everything were in a `ace_medical` class.
But just personal preference because I don't like scrolling in config viewer :D

addons/medical_treatment/functions/fnc_checkItems.sqf
```diff
@@ -16,27 +16,22 @@ if (EGVAR(medical,convertItems) == 2) exitWith {};
...
 
-if (EGVAR(medical,convertItems) != 0) exitWith {};
+    if (EGVAR(medical,convertItems) != 0) exitWith {};
 
-for "_i" from 1 to _countFirstAidKit do {
-    _unit addItem "ACE_fieldDressing";
-    _unit addItem "ACE_packingBandage";
-    _unit addItem "ACE_morphine";
-    _unit addItem "ACE_tourniquet";
+    for "_i" from 1 to _itemCount do {
```
dedmen: for loop 🤢 
Could just do one for loop.
```sqf
{
    for "_j" from 1 to ((_x select 1) * _itemCount) do {
        _unit addItem (_x select 0);
    };
} forEach _value
```

addons/medical_treatment/functions/fnc_checkItems.sqf
```diff
@@ -16,27 +16,22 @@ if (EGVAR(medical,convertItems) == 2) exitWith {};
...
-private _countFirstAidKit = [_unit, "FirstAidKit"] call EFUNC(common,getCountOfItem);
-_unit removeItems "FirstAidKit";
+// Item conversions are cached at preStart
+private _itemHash = uiNamespace getVariable QGVAR(replaceItems);
 
-private _countMedikit = [_unit, "Medikit"] call EFUNC(common,getCountOfItem);
-_unit removeItems "Medikit";
+private _replaceItem = {
+    private _itemCount = [_unit, _key] call EFUNC(common,getCountOfItem);
+    _unit removeItems _key;
```
dedmen: No need if _itemCount is 0.
Don't know how expensive it is. You could try packing a backpack full with hundreds of ace bandages,
and running `player removeItems "FirstAidKit"` in debug console to see how expensive it is to remove a item that player doesn't even have.
Knowing Arma, I would assume it iterates over all items in inventory and grabs the ones that match. Will will be hella expensive with 200 items in inventory and trying to remove an item that doesn't exist.

addons/medical_treatment/ACE_Medical_ReplaceItems.hpp
jonpas: Make a better config viewer.

dedmen: Really need a online config viewer that actually is maintained -.-
There are good viewers, but all only show a subset (weapons,vehicles,ammo) and/or are outdated and closed-source.

addons/medical_treatment/functions/fnc_checkItems.sqf
SilentSpike: I knew I was missing an obvious way to simplify these loops! 😷

SilentSpike:
It did come to attention that this is already implemented in #6933

I'm not sure which implementation is preferable so will leave this open for now (there are also some minor changes here we should port over there if we go with that one).

SilentSpike:
Looking at the other implementation I expect it should be faster because there aren't calls to the CBA hash functions.

dedmen:
There are also less calls in the other impl and it fixes my comments too.
Well doesn't need compileFinal as it's in preInit, which is fine too. Not big enough to benefit from preStart

PabstMirror:
6933 merged, but we still may want to intergrate parts of this PR
ref: https://github.com/acemod/ACE3/pull/6933#discussion_r288023248

PR #7023, opened 2019-05-28
Refuel - Add events for start, stop and progress

dedmen:
@synixebrett requested this.

`ace_refuel_refuelTick` might be renamed to just `ace_refuel_tick` ?
We also want a `refuelDone` event, but for that we need to store the amount of fuel that was transferred in the previous interaction.

I think local event is enough for this?

Todo:
- [x] Write documentation for eventhandlers

### DIFF
--- a/addons/refuel/functions/fnc_turnOff.sqf
+++ b/addons/refuel/functions/fnc_turnOff.sqf
@@ -21,3 +21,4 @@ params [["_unit", objNull, [objNull]], ["_nozzle", objNull, [objNull]]];
 _nozzle setVariable [QGVAR(lastTickMissionTime), nil];
 _nozzle setVariable [QGVAR(isRefueling), false, true];
 [LSTRING(Hint_Stopped), 1.5, _unit] call EFUNC(common,displayTextStructured);
+[QGVAR(refuelStopped), [_nozzle getVariable QGVAR(source), _nozzle getVariable QGVAR(sink)]] call CBA_fnc_localEvent;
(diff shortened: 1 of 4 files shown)


jonpas:
> I think local event is enough for this?

Global would take a lot of network resources for something people might not even need.

dedmen:
There are many places where refueling might end.
And we'd need a variable and keep track of transferred fuel for a proper refuelEnded event.
Seems like maybe too much work?
Or we might aswell just do a refuelEnded that doesn't pass the total amount of transferred fuel as argument, if the user needs it. He can do that by himself by using the start/tick/end events.

synixebrett:
This is going to be extremely niche, asking them to track the ticks isn't too much to ask imho

addons/refuel/functions/fnc_turnOn.sqf
```diff
@@ -21,3 +21,4 @@ params [["_unit", objNull, [objNull]], ["_nozzle", objNull, [objNull]]];
 _nozzle setVariable [QGVAR(lastTickMissionTime), CBA_missionTime];
 _nozzle setVariable [QGVAR(isRefueling), true, true];
 [LSTRING(Hint_Started), 1.5, _unit] call EFUNC(common,displayTextStructured);
+[QGVAR(refuelStarted), [_nozzle getVariable QGVAR(source), _nozzle getVariable QGVAR(sink)]] call CBA_fnc_localEvent;
```
commy2: newline @ eof missing

addons/refuel/functions/fnc_turnOff.sqf
```diff
@@ -21,3 +21,4 @@ params [["_unit", objNull, [objNull]], ["_nozzle", objNull, [objNull]]];
 _nozzle setVariable [QGVAR(lastTickMissionTime), nil];
 _nozzle setVariable [QGVAR(isRefueling), false, true];
 [LSTRING(Hint_Stopped), 1.5, _unit] call EFUNC(common,displayTextStructured);
+[QGVAR(refuelStopped), [_nozzle getVariable QGVAR(source), _nozzle getVariable QGVAR(sink)]] call CBA_fnc_localEvent;
```
commy2: newline @ eof missing

Dystopian:
> `ace_refuel_refuelTick` might be renamed to just `ace_refuel_tick`

I vote for this. And other events too.

Issue #6861, opened 2019-03-12
Overheating. Loud and unused sounds


severgun:
**Arma 3 Version:** `1.90` (stable)
**CBA Version:** `3.9` (prev stable)
**ACE3 Version:** `3.12.5` (stable)

**Mods:**
```
- CBA_A3
- ace
- RHS
...
```

**Description:**
1.  ACE Overheating module have very loud(far audible) jamming sound which can be heared I think ~100m away.
2.  There is unused files and config defined sounds

**Steps to reproduce:**
- Go multiplayer. Get weapon jam event. Listen from far away.

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

**Additional information:**
Define 10m or so?
https://github.com/acemod/ACE3/blob/842f9e5d3bedf18d69974950908c328ddb10bcb0/addons/overheating/functions/fnc_jamWeapon.sqf#L58

dedmen:
Easy fix. Set distance parameter to 10-15m

```sqf
playSound3D ["a3\sounds_f\weapons\Other\dry9.wss", _this select 0, false, [0,0,0], 1, 1, 15]; 
```

severgun:
Yes. What about dead files and code?

PR #7024, opened 2019-05-29
Fixing playsound3d limits

Cyruz143:
- Adhere to the new 5x sound limit in MP with playsound3d

I've bumped the ranges on some of the sounds to counter the volume lowering. I've not tested it in-game, just used some guess work on the numbers.


severgun:
Any chance to include fix for https://github.com/acemod/ACE3/issues/6861 ?

PabstMirror:
Tested working
This pr puts the params into spec for the `playSound3d` command
some sounds are still a little quieter than they used to be before the command changed 
(I remember the hot wirebreak being loud)
I guess the only way to fix would be to modify the sound file to increase loudness

PR #7029, opened 2019-06-03
Grenades - Don't flashbang spectators

PabstMirror:


bux:
Does this include ace_spectator spectators?

PR #7040, opened 2019-06-07
Logistics enable/disable settings (no more compat 🎉)

BaerMitUmlaut:
- Add enable settings for each component
- Remove the need for compat addons for the logistic components

Status:
- [x] Refuel
- [x] Repair
- [ ] Rearm


addons/refuel/CfgVehicles.hpp
```diff
@@ -451,31 +437,26 @@ class CfgVehicles {
...
         GVAR(fuelCargo) = REFUEL_INFINITE_FUEL;
     };
 
     class Land_FuelStation_01_pump_F: House_F {
-        transportFuel = 0; //50k
         GVAR(hooks)[] = {{0, 0.4, -0.5}, {0, -0.4, -0.5}};
         GVAR(fuelCargo) = REFUEL_INFINITE_FUEL;
     };
     class Land_FuelStation_01_pump_malevil_F: House_F {
-        transportFuel = 0; //50k
```
PabstMirror: `setFuelCargo` won't work on map objects
e.g. the fuel station on stratis (`[2713, 5796]`) will fill a empty car

jonpas: I knew this command is somehow broken.

BaerMitUmlaut: Will check that. Hiding and replacing it with a simple object might be an option.

addons/repair/script_component.hpp
```diff
@@ -17,5 +17,6 @@
 #include "\z\ace\addons\main\script_macros.hpp"
 
 #define TRACK_HITPOINTS ["hitltrack", "hitrtrack"]
-
 #define DAMAGE_COLOR_SCALE ["#FFFFFF", "#FFFF7E", "#FFEC4D", "#FFD52C", "#FCB121", "#FF9916", "#FF7D16", "#FF4400", "#FF0000"]
+#define ITEM_TYPE_TOOLKIT 620
```
mharis001: https://github.com/acemod/ACE3/blob/master/addons/main/script_macros.hpp#L93

addons/repair/script_component.hpp
```diff
@@ -17,5 +17,6 @@
...
+#define ITEM_TYPE_BIPOD 302
```
mharis001: https://github.com/acemod/ACE3/blob/master/addons/main/script_macros.hpp#L80

BaerMitUmlaut: Oh, duh, thanks 😄

addons/repair/XEH_postInit.sqf
```diff
@@ -18,3 +20,65 @@
...
+};
+
+
+// Item replacement EH
+
+["loadout", {
+    params ["_unit"];
+
+    private _cfgWeapons = configFile >> "CfgWeapons";
+    private _toolKits = (items _unit) select {
```
dedmen: VERY bad. `(items _unit)` might be several hundreds of items that we don't care about, and we do a config lookup for EACH of them, EVERY time the players loadout changes.

Was this done in old code or is this a new feature?
Alternatively use uniqueUnitItems if we really need to do this. But I really dislike this happening on every loadout change, how is medical doing item replacement?

https://github.com/acemod/ACE3/blob/44050df98b00e579e5b5a79c0d76d4d1138b4baa/addons/medical_treatment/functions/fnc_checkItems.sqf

Seems like medical is also doing it on every loadout change.. oof.
uniqueUnitItems should be good enough I guess. But if we find items we need to combine with common,getCountOfItem to get the number to replace.
Also needs a setting to disable this.

dedmen:
> addons/repair/.vscode/ipch/1ce1e1282607b219/mmap_address.bin

*burp*
Maybe we should add `.vscode` to repository .gitignore?

addons/repair/XEH_postInit.sqf
dedmen: We could (in general, not just this) implement a "delayed item replacement" system.
When a loadout EH is fired we just remember it in a waitAndExecute, if another loadout EH fires we reschedule that waitAndExecute and bump up it's time.

That way if you do a spam loadout change (like grabbing a hundred bandages from a box) it won't be affected. And this item replacement is not a time critical thing, no issue making it happen 5/10 seconds late, same for medical.

Just make a global "delayedItemReplacement" EH that we can then use for that.

addons/refuel/CfgVehicles.hpp
Dystopian: > `setFuelCargo` won't work on map objects
> e.g. the fuel station on stratis (`[2713, 5796]`) will fill a empty car

It does work, you must just handle all 3 pumps at the station.
The real problem here is that fuel pumps are `Static` children and they don't trigger `InitPost` EH. This can be handled with `nearestObjects` or `nearObjects` call with each pump class at mission init. But how to handle pumps created during mission with `createVehicle` after mission init?
At least fuel pumps `class CBA_Extended_EventHandlers` removed in #4533 can be restored. But this must be done also for all 3rd party pumps so some compat can be needed.

addons/repair/initSettings.sqf
```diff
@@ -1,4 +1,16 @@
-// CBA Settings [ADDON: ace_repair]:
+[
+    QGVAR(enabled),
+    "CHECKBOX",
+    ELSTRING(common,Enabled),
+    [LELSTRING(OptionsMenu,CategoryLogistics), localize "str_state_repair"],
+    true,
+    true,
+    {
+        parsingNamespace setVariable [QGVAR(actionEnabled), parseNumber !_this];
```
Dystopian, line 9: what is this line for?

addons/refuel/CfgVehicles.hpp
dedmen: terrain objects can also stream in/out, and might not store their fuel state inbetween.

Dystopian: I already tested this here https://github.com/acemod/ACE3/pull/7613#issuecomment-612692751

LinkIsGrim:
If something does `setUnitTrait` it'll override, but IMO this is better than replacing every toolkit.

I'd rather defer the refuel/rearm compats to another PR, they'll need lots more special handling from the looks of it.

BaerMitUmlaut:
> but IMO this is better than replacing every toolkit

Why? We have a framework for item replacement already and it stays compatible with the vanilla attributes. This is a regression.

PR #7042, opened 2019-06-08
Crew Served Weapons - Final Integration

PabstMirror:
Contains: 
CSW #5652 
Dragon #6773 
RHS CSW compatability (#7082) 
CSW/Dragon cleanup and rearm compat (#7068)

Artillery Tables #6853 (WIP) is based on this PR, but can be merged later

Do not squash merge this PR


severgun:
As much as I can see CSW still does not handle "cheating" with ace_rearm and does not support loading from ground.

bux:
What does that mean @severgun ?

severgun:
More detailed comments were in original CSW pull request.
For now, you can get ammo truck and load all magazines in static weapon even if by CSW you limit it to one at a time. 
Now ace_rearm allow to disable rearming only by script. I think config parameter would be better here.

Loading from ground more like feature request, but I think it can be in initial release. Just check not only players inventory, but groundWeapholder in near field of weapon.
IRL you will take heavy MG ammoboxes from ground where your bearer droped them and not from back pocket of jeans.

addons/csw/functions/fnc_assemble_deployTripod.sqf
```diff
@@ -0,0 +1,66 @@
...
+        private _tripodClassname = getText(configFile >> "CfgWeapons" >> _secondaryWeaponClassname >> QUOTE(ADDON) >> "deploy");
+
+        // Create a tripod
+        private _cswTripod = createVehicle [_tripodClassname, [0, 0, 0], [], 0, "NONE"];
+
+        private _posATL = _player getRelPos [2, 0];
+        _posATL set [2, ((getPosATL _player) select 2) + 0.5];
+
+        _cswTripod setDir (direction _player);
+        _cswTripod setPosATL _posATL;
```
severgun: Some small tripods can suffer from rollover even on slight slope.
```suggestion
        _cswTripod setPosATL _posATL;
        _cswTripod setCenterOfMass ((getCenterOfMass _cswTripod) vectorAdd [0,0,-0.3]);
```

addons/csw/functions/fnc_assemble_deployTripod.sqf
```diff
@@ -0,0 +1,66 @@
...
+
+        private _tripodClassname = getText(configFile >> "CfgWeapons" >> _secondaryWeaponClassname >> QUOTE(ADDON) >> "deploy");
+
+        // Create a tripod
+        private _cswTripod = createVehicle [_tripodClassname, [0, 0, 0], [], 0, "NONE"];
+
+        private _posATL = _player getRelPos [2, 0];
+        _posATL set [2, ((getPosATL _player) select 2) + 0.5];
+
+        _cswTripod setDir (direction _player);
```
severgun: Some small tripods can suffer from rollover even on slight slope.
```suggestion
        _cswTripod setDir (direction _player);
        _cswTripod setCenterOfMass ((getCenterOfMass _cswTripod) vectorAdd [0,0,-0.3]);
```

dedmen: Should we really fix a bug in the 3D model of some tripods?

PabstMirror: ace still does this in ace_vehicles
```
    class StaticWeapon {
        class ACE_FixMass {
            init = QUOTE(if (local (_this select 0)) then {(_this select 0) setMass (getMass (_this select 0) max 250)};);
```
so for full static weapons's we try to fix them, so it wouldn't be out of line to fix the tripod models as well
we control most of the p3ds for tripods, so we might be able to fix by moving mass
and perhaps this command would work better than setMass for the full weapon?

veteran29: FYI setting center of mass has one issue, if you can pack the weapon into the backpacks it will reset after it is assembled again.

severgun: > Should we really fix a bug in the 3D model of some tripods?
This happens on small and narrow third party models. RHS/CUP SPG and Metis have very narrow models. So I don't think that it can be/will be fixed.

> FYI setting center of mass has one issue, if you can pack the weapon into the backpacks it will reset after it is assembled again.
That is why it should be done on each createVehicle. Not shure about Drag/Carry actions after placement.

veteran29: > That is why it should be done on each createVehicle. 

Weapon assembly/disassembly does not fire `createVehicle`, the weapon is hidden at some coordinates until assembled again.

TheCandianVendingMachine: @veteran29 In the way that CSW works `createVehicle` does fire

addons/apl/data/m3tripod.rvmat
```diff
@@ -0,0 +1,98 @@
...
+};
+class Stage3
+{
+	texture = "#(argb,8,8,3)color(0,0,0,0,MC)";
+	uvSource = "tex";
+	class uvTransform
+	{
+		aside[] = {1.0,0.0,0.0};
+		up[] = {0.0,1.0,0.0};
+		dir[] = {0.0,0.0,0.0};
```
dedmen: should be 0,0,1, it's a matrix.

it's wrong everywhere -.-

PR #7044, opened 2019-06-08
Medical - Only stitch wounds on not bleeding body parts

mharis001:
- Change stitching to only be possible for bandaged wounds on not bleeding body parts
- Fix issue where another stitched wound would be added instead of increasing the amount


TheMagnetar:
I think it should be a setting and default turned off.

alganthe:
Same, should be a setting like in current.

commy2:
I think it should be a setting... and default turned on :P

dedmen:
>I think it should be a setting... and default turned on :P

But muh realism, I can stitch that bleeding guy if I want to!

SilentSpike:
Now that you mention realism, why should stable condition be a requirement? (I may have missed a meeting where this setting was discussed).

Shouldn't it just be `GET_WOUND_BLEEDING(_unit) == 0`? Otherwise you can't stitch someone who's unconscious.

dedmen:
>why should stable condition be a requirement

Good point too. I thought of "stable" as not bleeding, but it's actually alot more.

https://github.com/acemod/ACE3/blob/847d2d41792b9d0f9ab27bb7b6ebf7eedaaa3907/addons/medical_status/functions/fnc_hasStableVitals.sqf

I can stitch someone with severe bloodloss or in cardiac arrest or with low heart rate or with low blood pressure.
It might not make sense, but nothing really prevents me from doing it.

jonpas:
Agreed, only bleeding should be stopped before stitching.

SilentSpike:
I'd maybe even go as far as saying a setting isn't needed if we just say bleeding should always be stopped before stitching

dedmen:
That opens a problem tho, you cannot stitch specific body parts can you? If you have problems with constant wound reopening (Left arm is now closed, but right arm already reopened) you can never get him dry.
Currently in old medical we just stitch in-between while someone else is bandaging, to close the already bandaged bodyparts and prevent reopening.

Maybe we can just allow stitching if there is atleast one bodypart that can be stitched (has wounds, but not bleeding) and then let it only stitch the parts that are able to (Needs to also adjust the timer).

mharis001:
> Maybe we can just allow stitching if there is atleast one bodypart that can be stitched (has wounds, but not bleeding) and then let it only stitch the parts that are able to (Needs to also adjust the timer).

This could work. 

We could also change stitching to be body part specific if we go this route. Thoughts?

Drofseh:
I would like that

dedmen:
>We could also change stitching to be body part specific if we go this route. Thoughts?

If we adjust the timer accordingly I think that's good.
But I think people will complain as you usually get multiple wounds.

alganthe:
The stitching kit removes wounds as the bar progresses, there's no need to make it work on a per limb basis.

addons/medical_treatment/functions/fnc_surgicalKitProgress.sqf
```diff
@@ -22,28 +22,42 @@
...
 private _bandagedWounds = GET_BANDAGED_WOUNDS(_patient);
 private _stitchedWounds = GET_STITCHED_WOUNDS(_patient);
 
-// Stop treatment if there are no wounds that can be stitched remaining
-if (_bandagedWounds isEqualTo []) exitWith { false };
+// Remove the first stitchable wound from the bandaged wounds
+private _treatedWound = _bandagedWounds deleteAt (_bandagedWounds find (_stitchableWounds select 0));
+_treatedWound params ["_treatedID", "_treatedBodyPartN", "_treatedAmountOf"];
+
+// Check if we need to add a new stitched wound or increase the amount of an exiting one
```
SilentSpike:
```suggestion
// Check if we need to add a new stitched wound or increase the amount of an existing one
```

SilentSpike:
All looks good to me

PR #7046, opened 2019-06-08
Speed and Altitude units

synixebrett:
- Add additional speed units
  - km/h
  - mph
  - kn
  - m/s

- Add additional altitude units
  - meters (m)
  - feet (f)

### DIFF
--- a/addons/arsenal/missions/Arsenal.VR/script_component.hpp
+++ b/addons/arsenal/missions/Arsenal.VR/script_component.hpp
@@ -1,4 +1,4 @@
 #include "\z\ace\addons\arsenal\script_component.hpp"
 
-#undef PREP // make.py can't redefine already defined macros
+#undef PREP
 #define PREP(var) FUNC(var) = compileFinal preprocessFileLineNumbers format ["fnc_%1.sqf", QUOTE(var)]
--- a/addons/units/CfgEventHandlers.hpp
+++ b/addons/units/CfgEventHandlers.hpp
@@ -1,0 +1,15 @@
+class Extended_PreStart_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preStart));
+    };
+};
+class Extended_PreInit_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preInit));
+    };
+};
+class Extended_PostInit_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_postInit));
+    };
+};
--- a/addons/units/config.cpp
+++ b/addons/units/config.cpp
@@ -1,0 +1,51 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        name = COMPONENT_NAME;
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"ace_common"};
+        author = ECSTRING(common,ACETeam);
+        authors[] = {"SynixeBrett"};
+        url = ECSTRING(main,URL);
+        VERSION_CONFIG;
+    };
+};
+
+class RscText;
+class RscInGameUI {
+    class RscUnitInfo {
+        class CA_Alt: RscText {
+            onLoad = QUOTE(call FUNC(onAltLoad));
+        };
+        class CA_Speed: RscText {
+            onLoad = QUOTE(call FUNC(onSpeedLoad));
+        };
+    };
+};
+
+class GVAR(alt): RscText {
+    style = 1;
+    colorText[] = {"(profilenamespace getvariable ['IGUI_TEXT_RGB_R',0])","(profilenamespace getvariable ['IGUI_TEXT_RGB_G',1])","(profilenamespace getvariable ['IGUI_TEXT_RGB_B',1])","(profilenamespace getvariable ['IGUI_TEXT_RGB_A',0.8])"};
+    text = "9999";
+    x = "6.3 * (((safezoneW / safezoneH) min 1.2) / 40) + (profilenamespace getvariable [""IGUI_GRID_VEHICLE_X"", (safezoneX + 0.5 * (((safezoneW / safezoneH) min 1.2) / 40))])";
+    y = "2.3 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25) + (profilenamespace getvariable [""IGUI_GRID_VEHICLE_Y"", (safezoneY + 0.5 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25))])";
+    w = "2 * (((safezoneW / safezoneH) min 1.2) / 40)";
+    h = "1 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25)";
+    sizeEx = "0.8 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25)";
+};
+
+class GVAR(speed): RscText {
+    style = 1;
+    colorText[] = {"(profilenamespace getvariable ['IGUI_TEXT_RGB_R',0])", "(profilenamespace getvariable ['IGUI_TEXT_RGB_G',1])", "(profilenamespace getvariable ['IGUI_TEXT_RGB_B',1])", "(profilenamespace getvariable ['IGUI_TEXT_RGB_A',0.8])"};
+    text = "9999";
+    x = "6.3 * (((safezoneW / safezoneH) min 1.2) / 40) + (profilenamespace getvariable [""IGUI_GRID_VEHICLE_X"", (safezoneX + 0.5 * (((safezoneW / safezoneH) min 1.2) / 40))])";
+    y = "1.3 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25) + (profilenamespace getvariable [""IGUI_GRID_VEHICLE_Y"", (safezoneY + 0.5 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25))])";
+    w = "2 * (((safezoneW / safezoneH) min 1.2) / 40)";
+    h = "1 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25)";
+    sizeEx = "0.8 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25)";
+};
+
+#include "CfgEventHandlers.hpp"
--- a/addons/units/functions/fnc_altInfo.sqf
+++ b/addons/units/functions/fnc_altInfo.sqf
@@ -1,0 +1,7 @@
+#include "script_component.hpp"
+
+params ["_unit"];
+
+if (_unit isEqualTo ALT_F) exitWith {["f", 0.3047999902464003]};
+
+["m", 1]
--- a/addons/units/functions/fnc_onAltLoad.sqf
+++ b/addons/units/functions/fnc_onAltLoad.sqf
@@ -1,0 +1,40 @@
+#include "script_component.hpp"
+
+params ["_ctrl"];
+
+private _alt = (ctrlParent _ctrl) ctrlCreate [QGVAR(alt), -1];
+
+private _units = (vehicle ace_player) call FUNC(altUnits);
+private _altInfo = [_units select 0] call FUNC(altInfo);
+
+systemChat "alt time";
+
+((ctrlParent _alt) displayCtrl 1005) ctrlSetText (_altInfo select 0);
+GVAR(ratio) = _altInfo select 1;
+
+private _settingEH = ["CBA_SettingChanged", {
+    params ["_setting", "_value"];
+    _thisArgs params ["_settingName", "_alt"];
+    if (_setting isEqualTo _settingName) then {
+        private _altInfo = [_value] call FUNC(altInfo);
+        ((ctrlParent _alt) displayCtrl 1005) ctrlSetText (_altInfo select 0);
+        GVAR(ratio) = _altInfo select 1;
+    };
+}, [_units select 1, _alt]] call CBA_fnc_addEventHandlerArgs;
+
+[{
+    (_this select 0) params ["_ctrl", "_alt", "_settingEH"];
+    if (isNull _ctrl) exitWith {
+        [_this select 1] call CBA_fnc_removePerFrameHandler;
+        ["CBA_SettingChanged", _settingEH] call CBA_fnc_removeEventHandler;
+    };
+    _ctrl ctrlSetFade 1;
+    _ctrl ctrlCommit 0;
+    private _altValue = (((getPosATL vehicle ace_player) select 2) / GVAR(ratio));
+    if (_altValue < 10) then {
+        _altValue = _altValue toFixed 1;
+    } else {
+        _altValue = str round _altValue;
+    };
+    _alt ctrlSetText _altValue;
+}, 0, [_ctrl, _alt, _settingEH]] call CBA_fnc_addPerFrameHandler;
--- a/addons/units/functions/fnc_onSpeedLoad.sqf
+++ b/addons/units/functions/fnc_onSpeedLoad.sqf
@@ -1,0 +1,33 @@
+#include "script_component.hpp"
+
+params ["_ctrl"];
+
+private _speed = (ctrlParent _ctrl) ctrlCreate [QGVAR(speed), -1];
+
+private _units = (vehicle ace_player) call FUNC(speedUnits);
+private _speedInfo = [_units select 0] call FUNC(speedInfo);
+
+((ctrlParent _speed) displayCtrl 1004) ctrlSetText (_speedInfo select 0);
+GVAR(ratio) = _speedInfo select 1;
+
+private _settingEH = ["CBA_SettingChanged", {
+    params ["_setting", "_value"];
+    _thisArgs params ["_settingName", "_speed"];
+    systemChat format ["%1 = %2", _setting, _value];
+    if (_setting isEqualTo _settingName) then {
+        private _speedInfo = [_value] call FUNC(speedInfo);
+        ((ctrlParent _speed) displayCtrl 1004) ctrlSetText (_speedInfo select 0);
+        GVAR(ratio) = _speedInfo select 1;
+    };
+}, [_units select 1, _speed]] call CBA_fnc_addEventHandlerArgs;
+
+[{
+    (_this select 0) params ["_ctrl", "_speed", "_settingEH"];
+    if (isNull _ctrl) exitWith {
+        [_this select 1] call CBA_fnc_removePerFrameHandler;
+        ["CBA_SettingChanged", _settingEH] call CBA_fnc_removeEventHandler;
+    };
+    _ctrl ctrlSetFade 1;
+    _ctrl ctrlCommit 0;
+    _speed ctrlSetText str (round (speed vehicle ace_player / GVAR(ratio)));
+}, 0, [_ctrl, _speed, _settingEH]] call CBA_fnc_addPerFrameHandler;
--- a/addons/units/stringtable.xml
+++ b/addons/units/stringtable.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ACE">
+    <Package name="Units">
+        <Key ID="STR_ACE_Units_Module_DisplayName">
+            <English>Units</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Speed_DisplayName">
+            <English>Speed</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_Description">
+            <English>Units used to display Speed</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Alt_DisplayName">
+            <English>Altitude</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Alt_Description">
+            <English>Units used to display Altitude</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Land_DisplayName">
+            <English>Land Vehicles</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Speed_KMH">
+            <English>km/h</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_MPH">
+            <English>mph</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_KNOT">
+            <English>kn</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_MS">
+            <English>m/s</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Alt_M">
+            <English>Meters</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Alt_F">
+            <English>Feet</English>
+        </Key>
+    </Package>
+</Project>
--- a/addons/units/$PBOPREFIX$
+++ b/addons/units/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\ace\addons\units
--- a/addons/units/XEH_PREP.hpp
+++ b/addons/units/XEH_PREP.hpp
@@ -1,0 +1,6 @@
+PREP(altInfo);
+PREP(altUnits);
+PREP(onAltLoad);
+PREP(onSpeedLoad);
+PREP(speedInfo);
+PREP(speedUnits);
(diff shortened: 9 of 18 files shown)


addons/arsenal/missions/Arsenal.VR/script_component.hpp
```diff
@@ -1,4 +1,4 @@
 #include "\z\ace\addons\arsenal\script_component.hpp"
 
-#undef PREP // make.py can't redefine already defined macros
+#undef PREP
```
commy2: Why delete this comment?

addons/units/CfgEventHandlers.hpp
```diff
@@ -0,0 +1,15 @@
+class Extended_PreStart_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preStart));
+    };
+};
+class Extended_PreInit_EventHandlers {
```
commy2:
```suggestion

class Extended_PreInit_EventHandlers {
```

addons/units/CfgEventHandlers.hpp
```diff
@@ -0,0 +1,15 @@
...
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preInit));
+    };
+};
+class Extended_PostInit_EventHandlers {
```
commy2:
```suggestion

class Extended_PostInit_EventHandlers {
```

addons/units/functions/fnc_onAltLoad.sqf
```diff
@@ -0,0 +1,40 @@
+#include "script_component.hpp"
+
+params ["_ctrl"];
+
+private _alt = (ctrlParent _ctrl) ctrlCreate [QGVAR(alt), -1];
+
+private _units = (vehicle ace_player) call FUNC(altUnits);
+private _altInfo = [_units select 0] call FUNC(altInfo);
+
+systemChat "alt time";
```
commy2: remove debug

addons/units/functions/fnc_altInfo.sqf
```diff
@@ -0,0 +1,7 @@
+#include "script_component.hpp"
+
+params ["_unit"];
+
+if (_unit isEqualTo ALT_F) exitWith {["f", 0.3047999902464003]};
```
commy2: Should probably be a macro with descriptive name. (It is the ratio of meters per feet).

addons/units/config.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+class GVAR(alt): RscText {
+    style = 1;
+    colorText[] = {"(profilenamespace getvariable ['IGUI_TEXT_RGB_R',0])","(profilenamespace getvariable ['IGUI_TEXT_RGB_G',1])","(profilenamespace getvariable ['IGUI_TEXT_RGB_B',1])","(profilenamespace getvariable ['IGUI_TEXT_RGB_A',0.8])"};
+    text = "9999";
+    x = "6.3 * (((safezoneW / safezoneH) min 1.2) / 40) + (profilenamespace getvariable [""IGUI_GRID_VEHICLE_X"", (safezoneX + 0.5 * (((safezoneW / safezoneH) min 1.2) / 40))])";
+    y = "2.3 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25) + (profilenamespace getvariable [""IGUI_GRID_VEHICLE_Y"", (safezoneY + 0.5 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25))])";
+    w = "2 * (((safezoneW / safezoneH) min 1.2) / 40)";
+    h = "1 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25)";
+    sizeEx = "0.8 * ((((safezoneW / safezoneH) min 1.2) / 1.2) / 25)";
```
commy2: `(((safezoneW / safezoneH) min 1.2) / 40)` = GUI_GRID_W
`((((safezoneW / safezoneH) min 1.2) / 1.2) / 25)` = GUI_GRID_H

Defined in `#include "\a3\ui_f\hpp\defineCommonGrids.inc"`.

commy2: Also, camel case is broken for profileNamespace, getVariable etc.

addons/units/functions/fnc_onAltLoad.sqf
```diff
@@ -0,0 +1,40 @@
...
+
+((ctrlParent _alt) displayCtrl 1005) ctrlSetText (_altInfo select 0);
+GVAR(altRatio) = _altInfo select 1;
```
commy2: `params` would still be benefitial, even though you have two more locals. Same below.

addons/units/functions/fnc_onAltLoad.sqf
```diff
@@ -0,0 +1,40 @@
+#include "script_component.hpp"
+
+params ["_ctrl"];
+
+private _alt = (ctrlParent _ctrl) ctrlCreate [QGVAR(alt), -1];
+
+private _units = (vehicle ace_player) call FUNC(altUnits);
+private _altInfo = [_units select 0] call FUNC(altInfo);
```
commy2: `[_units select 0]` is the same as `_units`, because FUNC(altInfo) only has one arg and as internal function with this also being unlikely to change, this line should be simplified.

addons/units/functions/fnc_onAltLoad.sqf
```diff
@@ -0,0 +1,40 @@
+#include "script_component.hpp"
+
+params ["_ctrl"];
+
+private _alt = (ctrlParent _ctrl) ctrlCreate [QGVAR(alt), -1];
+
+private _units = (vehicle ace_player) call FUNC(altUnits);
```
commy2: ACE_player

addons/units/functions/fnc_onAltLoad.sqf
```diff
@@ -0,0 +1,40 @@
...
+        [_this select 1] call CBA_fnc_removePerFrameHandler;
+        ["CBA_SettingChanged", _settingEH] call CBA_fnc_removeEventHandler;
+    };
+    _ctrl ctrlSetFade 1;
+    _ctrl ctrlCommit 0;
+    private _altValue = (((getPosATL vehicle ace_player) select 2) / GVAR(altRatio));
+    if (_altValue < 10) then {
+        _altValue = _altValue toFixed 1;
+    } else {
+        _altValue = str round _altValue;
```
commy2: `toFixed 0` would be the same result as far as I can tell and be one command less.
1 binary + number should trump 2 unary.

addons/units/functions/fnc_onSpeedLoad.sqf
```diff
@@ -0,0 +1,32 @@
+#include "script_component.hpp"
+
+params ["_ctrl"];
+
+private _speed = (ctrlParent _ctrl) ctrlCreate [QGVAR(speed), -1];
+
+private _units = (vehicle ace_player) call FUNC(speedUnits);
+private _speedInfo = [_units select 0] call FUNC(speedInfo);
```
commy2: Both lines, same as above.

addons/units/functions/fnc_onSpeedLoad.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+((ctrlParent _speed) displayCtrl 1004) ctrlSetText (_speedInfo select 0);
+GVAR(speedRatio) = _speedInfo select 1;
```
commy2: Same as above.

addons/units/functions/fnc_onSpeedLoad.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+[{
+    (_this select 0) params ["_ctrl", "_speed", "_settingEH"];
+    if (isNull _ctrl) exitWith {
+        [_this select 1] call CBA_fnc_removePerFrameHandler;
+        ["CBA_SettingChanged", _settingEH] call CBA_fnc_removeEventHandler;
+    };
+    _ctrl ctrlSetFade 1;
+    _ctrl ctrlCommit 0;
+    _speed ctrlSetText str (round (speed vehicle ace_player / GVAR(speedRatio)));
```
commy2: toFixed 0 probably.

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_MS">
+            <English>m/s</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Alt_M">
+            <English>Meters</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Alt_F">
+            <English>Feet</English>
```
commy2:
```suggestion
            <English>Feet</English>
            <German>Fuß</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_KNOT">
+            <English>kn</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_MS">
+            <English>m/s</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Alt_M">
+            <English>Meters</English>
```
commy2:
```suggestion
            <English>Meters</English>
            <German>Meter</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+            <English>km/h</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_MPH">
+            <English>mph</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_KNOT">
+            <English>kn</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_MS">
+            <English>m/s</English>
```
commy2:
```suggestion
            <English>m/s</English>
            <German>m/s</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+        </Key>
+
+        <Key ID="STR_ACE_Units_Speed_KMH">
+            <English>km/h</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_MPH">
+            <English>mph</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_KNOT">
+            <English>kn</English>
```
commy2:
```suggestion
            <English>kn</English>
            <German>kn</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+
+        <Key ID="STR_ACE_Units_Land_DisplayName">
+            <English>Land Vehicles</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Speed_KMH">
+            <English>km/h</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_MPH">
+            <English>mph</English>
```
commy2:
```suggestion
            <English>mph</English>
            <German>mph</German>
```
I suppose it is technically mi/h, but no one uses this here so it is only known as "mph" from Ami cars and Hollywood movies.

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+        <Key ID="STR_ACE_Units_Alt_Description">
+            <English>Units used to display Altitude</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Land_DisplayName">
+            <English>Land Vehicles</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Speed_KMH">
+            <English>km/h</English>
```
commy2:
```suggestion
            <English>km/h</English>
            <German>km/h</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+
+        <Key ID="STR_ACE_Units_Alt_DisplayName">
+            <English>Altitude</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Alt_Description">
+            <English>Units used to display Altitude</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Land_DisplayName">
+            <English>Land Vehicles</English>
```
commy2:
```suggestion
            <English>Land Vehicles</English>
            <German>Bodenfahrzeuge</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ACE">
+    <Package name="Units">
+        <Key ID="STR_ACE_Units_Module_DisplayName">
+            <English>Units</English>
```
commy2:
```suggestion
            <English>Units</English>
            <German>Einheiten</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+        </Key>
+
+        <Key ID="STR_ACE_Units_Speed_DisplayName">
+            <English>Speed</English>
```
commy2:
```suggestion
            <English>Speed</English>
            <German>Geschwindigkeit</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_Description">
+            <English>Units used to display Speed</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Alt_DisplayName">
+            <English>Altitude</English>
```
commy2:
```suggestion
            <English>Altitude</English>
            <German>Flughöhe</German>
```
There may be a more anglicized expression in the German Arma script, but I cba to check that out.

On the other hand, this may be a string in the game already, so I will probally check later.

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+    <Package name="Units">
+        <Key ID="STR_ACE_Units_Module_DisplayName">
+            <English>Units</English>
+        </Key>
+
+        <Key ID="STR_ACE_Units_Speed_DisplayName">
+            <English>Speed</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Speed_Description">
+            <English>Units used to display Speed</English>
```
commy2:
```suggestion
            <English>Units used to display Speed.</English>
            <German>Maßeinheiten um Geschwindigkeit darzustellen.</German>
```

addons/units/stringtable.xml
```diff
@@ -0,0 +1,46 @@
...
+        </Key>
+
+        <Key ID="STR_ACE_Units_Alt_DisplayName">
+            <English>Altitude</English>
+        </Key>
+        <Key ID="STR_ACE_Units_Alt_Description">
+            <English>Units used to display Altitude</English>
```
commy2:
```suggestion
            <English>Units used to display Altitude</English>
            <German>Maßeinheit, um Flughöhe darzustellen.</German>
```

commy2:
Is this worth it to have two (or more) PFH running each frame?
Can this be optimized to have no PFH running in case the settings match Arma default units?

addons/arsenal/missions/Arsenal.VR/script_component.hpp
synixebrett: Ah I was building with armake2 and this breaks it. I just used `commit -a` so it got tracked.

synixebrett:
It could be modified to use no PFH at the cost of complication. When in a land vehicle there will be a single PFH running, in aircraft there will be 2.

addons/units/stringtable.xml
synixebrett: It wasn't one that I could find with https://synixebrett.github.io/stringtables. I used all the existing strings I found

commy2:
Is the control actually deleted when leaving the vehicle, therefore is there the PFH only running when inside the vehicle and it stops when leaving the vehicle?

addons/arsenal/missions/Arsenal.VR/script_component.hpp
commy2: Then move the comment above the line. That should work.

addons/units/config.cpp
```diff
@@ -0,0 +1,52 @@
...
+        };
+    };
+};
+
+class GVAR(alt): RscText {
+    style = 1;
+    colorText[] = {"(profileNamespace getVariable ['IGUI_TEXT_RGB_R',0])","(profileNamespace getVariable ['IGUI_TEXT_RGB_G',1])","(profileNamespace getVariable ['IGUI_TEXT_RGB_B',1])","(profileNamespace getVariable ['IGUI_TEXT_RGB_A',0.8])"};
+    text = "9999";
+    x = 6.3 * GUI_GRID_W + (profileNamespace getVariable ["IGUI_GRID_VEHICLE_X", (safezoneX + 0.5 * GUI_GRID_W)]);
+    y = 2.3 * GUI_GRID_H + (profileNamespace getVariable [""IGUI_GRID_VEHICLE_Y"", (safezoneY + 0.5 * GUI_GRID_H)]);
```
commy2: Double quotes? It is different than the line above.

synixebrett:
> Is the control actually deleted when leaving the vehicle, therefore is there the PFH only running when inside the vehicle and it stops when leaving the vehicle?

Yes, when entering the vehicle the PFHs are created, 1 for land and 2 for air. Upon exiting the controls are destroyed and the PHFs and EHs are removed

commy2:
Was this confirmed in a test? It might be that the controls linger and are just hidden, but still exist, until a new vehicle is entered.

synixebrett:
It was confirmed with `systemChats` :D

The controls were null upon leaving a vehicle and the PFHs are being removed without issue.

addons/arsenal/missions/Arsenal.VR/script_component.hpp
jonpas: What commy2 said, I'll get that branch merged today.

Dystopian:
I think `units` component name is too similar to `units` arma word as `men`. Maybe it should be more descriptive.

commy2:
Probably be moved to the ui component.

addons/units/config.cpp
dedmen:
```suggestion
    y = 2.3 * GUI_GRID_H + (profileNamespace getVariable ["IGUI_GRID_VEHICLE_Y", (safezoneY + 0.5 * GUI_GRID_H)]);
```

synixebrett:
I didn't put it in the UI addon since in my opinion, it is serving a different purpose. It is possible someone wouldn't want the config changes made to the UI but still want the different units this addon provides. I do agree that `units` isn't the best name but it's all I could come up with at the time.

addons/ui_units/stringtable.xml
```diff
@@ -1,56 +1,56 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project name="ACE">
     <Package name="Units">
-        <Key ID="STR_ACE_Units_Module_DisplayName">
+        <Key ID="STR_ACE_UI_UNITS_Module_DisplayName">
```
jonpas: `STR_ACE_UI_Units_Module_DisplayName`

synixebrett:
@commy2 Ready for a re-review

addons/ui_units/config.cpp
```diff
@@ -0,0 +1,52 @@
...
+        };
+        class CA_Speed: RscText {
+            onLoad = QUOTE(call FUNC(onSpeedLoad));
+        };
+    };
+};
+
+class GVAR(alt): RscText {
+    style = 1;
+    colorText[] = {"(profileNamespace getVariable ['IGUI_TEXT_RGB_R',0])","(profileNamespace getVariable ['IGUI_TEXT_RGB_G',1])","(profileNamespace getVariable ['IGUI_TEXT_RGB_B',1])","(profileNamespace getVariable ['IGUI_TEXT_RGB_A',0.8])"};
```
dedmen:
```suggestion
    colorText[] = {
        "(profileNamespace getVariable ['IGUI_TEXT_RGB_R',0])",
        "(profileNamespace getVariable ['IGUI_TEXT_RGB_G',1])",
        "(profileNamespace getVariable ['IGUI_TEXT_RGB_B',1])",
        "(profileNamespace getVariable ['IGUI_TEXT_RGB_A',0.8])"
    };
```
Doesn't hurt right?

addons/ui_units/functions/fnc_onAltLoad.sqf
```diff
@@ -0,0 +1,38 @@
...
+
+[{
+    (_this select 0) params ["_ctrl", "_alt", "_settingEH"];
+    if (isNull _ctrl) exitWith {
+        [_this select 1] call CBA_fnc_removePerFrameHandler;
+        ["CBA_SettingChanged", _settingEH] call CBA_fnc_removeEventHandler;
+    };
+    _ctrl ctrlSetFade 1;
+    _ctrl ctrlCommit 0;
+    private _altValue = (((getPosATL vehicle ACE_player) select 2) / GVAR(altRatio));
```
commy2: ATL is based on bottom of the sea height above water. Shouldn't it be sea level = 0 instead?
I don't suggest AGL, because that changes with the waves (so it is dependant on time), but surely this needs to be handled differently over the sea, right?

synixebrett: Hm, yeah that would be different than vanilla behaviour. Is there a way to determine if you are above water? I'm not familiar with one. It should switch to ASL in that case.

commy2: ATL above land would make sense. Predict height of when you have to open the chute. No idea what BI does, why don't we just read their value with `parseNumber ctrlText` from the hard coded control and convert that if need to our units?

synixebrett: We lose precision that way since we will be doing the conversions on rounded numbers. I'd like to ideally use the raw values so we get accurate results.

synixebrett: Wait, couldn't any negative values just use 0?

commy2: No.

synixebrett: `(getPosATL player) - (0 min (getTerrainHeightASL getPos player))` should work just fine

synixebrett: Fixed in 8ee50719e5be1d47c7a27d16dc707d0047126973

Max255PL:
Since this modifies the altitude display already, would it be possible to have both ASL and AGL measurements shown at the same time in all aircraft? 
This was a feature back in ACE2 (along with a UI variometer), random pic showing this: https://i.ytimg.com/vi/oLnRu10thKY/maxresdefault.jpg
(ASL middle right)

Max255PL:
Found a bug with this, when controlling UAVs the UI is still getting the player position instead of the drone. Result as expected.

commy2:
All `vehicle ACE_player` should be `cameraOn`.

synixebrett:
> Since this modifies the altitude display already, would it be possible to have both ASL and AGL measurements shown at the same time in all aircraft?
> This was a feature back in ACE2 (along with a UI variometer)

This PR is not really modifying the UI, and there is unfortunately not enough space to add these in. It's something I'd be interested in adding but it might be better suited for an aircraft orientated mod like AWEsome since it will require modifying the look of the UI to add room. At the very least it should be a different PR if it is to be added to ACE.

addons/arsenal/missions/Arsenal.VR/script_component.hpp
jonpas: Can be reverted!

addons/ui_units/initSettings.sqf
```diff
@@ -0,0 +1,47 @@
+[
+    QGVAR(landSpeed),
+    "LIST",
+    [localize LSTRING(Speed_DisplayName), localize LSTRING(Speed_Description)],
+    [format["ACE %1", localize LSTRING(Module_DisplayName)], localize LSTRING(Land_DisplayName)],
+    [[SPEED_KMH, SPEED_MPH, SPEED_KNOT, SPEED_MS], [LSTRING(Speed_KMH), LSTRING(Speed_MPH), LSTRING(Speed_KNOT), LSTRING(Speed_MS)], 0]
+] call CBA_settings_fnc_init;
```
jonpas: Is it really a good idea to have those local? What units a car shows you doesn't depend on the driver, but on the car. Additionally, I don't want to impose confusion when people run different units, so making those global settings would be ideal.

commy2: If you don't want to confuse people with different settings on the clients, force one or the other. But leave the option to not force this settings for communities that don't care.

PR #7047, opened 2019-06-08
Medical Treatment - Add medical facility classes config

mharis001:
- title, https://github.com/acemod/ACE3/pull/6933#discussion_r274387363

### DIFF
--- a/addons/medical_treatment/XEH_preInit.sqf
+++ b/addons/medical_treatment/XEH_preInit.sqf
@@ -34,10 +34,13 @@ GVAR(animDurations) = [] call CBA_fnc_createNamespace;
     ["AinvPknlMstpSnonWnonDnon_medic1", 10]
 ];
 
-// class names of medical facilities
+// class names of medical facilities (config case)
 GVAR(facilityClasses) = [];
 {
-    GVAR(facilityClasses) append getArray _x;
+    {
+        private _name = configName (configFile >> "CfgVehicles" >> _x);
+        if (_name != "") then { GVAR(facilityClasses) pushBackUnique _name; };
+    } forEach getArray _x;
 } forEach configProperties [configFile >> QEGVAR(medical,facilities), "isArray _x"];
 
 // array of medical items to replace and their ACE equivalents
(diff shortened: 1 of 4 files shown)


addons/medical_treatment/XEH_preInit.sqf
```diff
@@ -35,23 +35,10 @@ GVAR(animDurations) = [] call CBA_fnc_createNamespace;
...
-    "Land_A_Hospital",
-    "CDF_WarfareBFieldhHospital",
-    "GUE_WarfareBFieldhHospital",
-    "INS_WarfareBFieldhHospital",
-    "RU_WarfareBFieldhHospital",
-    "USMC_WarfareBFieldhHospital"
-];
+GVAR(facilityClasses) = [];
+{
+    GVAR(facilityClasses) append getArray _x;
```
PabstMirror:
```suggestion
    GVAR(facilityClasses) append ((getArray _x) select {isClass (configFile >> "CfgVehicles" >> _x)});
```

dedmen: Our checks are case sensitive, might also convert to config-case just to be safe?
Or just add to docs that it's case-sensitive and screw people who can't write correctly.

addons/medical_treatment/XEH_preInit.sqf
```diff
@@ -34,24 +34,14 @@ GVAR(animDurations) = [] call CBA_fnc_createNamespace;
...
-    "INS_WarfareBFieldhHospital",
-    "RU_WarfareBFieldhHospital",
-    "USMC_WarfareBFieldhHospital"
-];
+// class names of medical facilities (config case)
+GVAR(facilityClasses) = [];
+{
+    {
+        private _name = configName (configFile >> "CfgVehicles" >> _x);
+        if (_name != "") then { GVAR(facilityClasses) pushBackUnique _name; };
```
dedmen: Could also use apply+append
and then a single `x = x - [""]` at the end, instead of constantly checking != "".
But that array is small, and this is done once in preInit. Probably a microsecond or two of potential.

PR #7049, opened 2019-06-09
Medical Treatment - Scale bandage time based on amount of wound treated

PabstMirror:
Was scaling treatment time by `_amountOf`
e.g. if you have 10x wounds, a bandage would still only treat the `effectiveness` defined in the configs
so it would take longer, but you wouldn't actually be doing any more work

- `_amountOf > _effectiveness` - scale by 1
 - `_effectiveness > _amountOf` - scale down to a max of 2/3

also further increases min bandage time to 2.25 to help with animation problems

### DIFF
--- a/addons/medical_treatment/functions/fnc_getBandageTime.sqf
+++ b/addons/medical_treatment/functions/fnc_getBandageTime.sqf
@@ -37,7 +37,9 @@ private _category = (_classID % 10);
 private _bandageTime = [BANDAGE_TIME_S, BANDAGE_TIME_M, BANDAGE_TIME_L] select _category;
 
 // Scale bandage time based on amount left and effectiveness (less time if only a little wound left)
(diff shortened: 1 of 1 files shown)


PabstMirror:
This should be a minor effect, and really only have any effect when the amount left is very small

| Amount Left  | Bandage Effectiveness | Speed mod |
| ------------- | ------------- | ------------- |
| 1  | 1  | 1 |
| 0.5 | 1 | 0.833 |
| 1 | 0.5 | 1 |

alganthe:
This fixes both upper and lower limit issues, the lower limit still has quite a fast animation speed but it's not as ridiculous as it was.

PabstMirror:
`_bandageTime = _bandageTime * (linearConversion [0, _effectiveness, _amountOf, 0.666, 1, true]);`
that 2/3's is subjective and can be tweaked in the future if we want

PR #7052, opened 2019-06-14
Maptools - Allow repositioning the map display for vanilla gps

### LINKED ISSUE #7051: request: On Map GPS movable
If I may make a request,

Please allow the on map GPS to be moved around.
It currently covers the bottom right of the map (the contour interval)

Otherwise great mod and great work.

PabstMirror:
Close #7051
also tweaked default Y pos so it does not block contours/scale box

### DIFF
--- a/addons/maptools/CfgUIGrids.hpp
+++ b/addons/maptools/CfgUIGrids.hpp
@@ -1,0 +1,20 @@
+class CfgUIGrids {
+    class IGUI {
+        class Presets {
+            class Arma3 {
+                class Variables {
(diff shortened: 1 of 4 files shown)


commy2:
Why not make it drag and droppable like the map tools or the vanilla watch and compass?

PabstMirror:
Aren't those p3d objects?
Maybe we can simple just move it a bit up so we don't block the "scale" box on the map

commy2:
>Aren't those p3d objects?
Doesn't change anything.

PR #7057, opened 2019-06-17
Medical - Cleanup settings

mharis001:
- Cleanup medical settings
- Remove some unused stringtable entries

Cleaned up setting names and descriptions and sorted them into appropriate sub-categories.
Left some settings that control the general behavior of medical in the main category.


addons/medical_statemachine/functions/fnc_handleStateUnconscious.sqf
```diff
@@ -28,14 +27,22 @@ if (_painLevel > 0) then {
...
+// Handle spontaneous wake up from unconsciousness
 if (EGVAR(medical,spontaneousWakeUpChance) > 0) then {
     if (_unit call EFUNC(medical_status,hasStableVitals)) then {
-        private _lastWakeUpCheck = _unit getVariable [QEGVAR(medical,lastWakeUpCheck), CBA_missionTime];
+        private _lastWakeUpCheck = _unit getVariable [QEGVAR(medical,lastWakeUpCheck);
+
+        // Handle setting being changed mid-mission and still properly check
+        // already unconscious units, should handle locality changes as well
+        if (isNil "_lastWakeUpCheck") exitWith {
+            _unit setVariable [QGVAR(medical,lastWakeUpCheck), CBA_missionTime];
```
mharis001: Could also trigger a wake up check right away rather than having the 15 second wait in this situation?

addons/medical_damage/stringtable.xml
```diff
@@ -1,78 +1,18 @@
...
-                <Hungarian>Mennyi sérülést szenvedhet el egy AI, mielőtt meghal?</Hungarian>
-                <Italian>Quanto è il danno che un'IA può sostenere prima di essere uccisa?</Italian>
-                <Japanese>AI が死に始める前に損傷を受けるようにしますか?</Japanese>
-                <Korean>얼마정도의 부상을 인공지능이 죽기 전까지 버틸 수 있습니까?</Korean>
-                <Chinesesimp>AI 死亡前所能承受的伤害程度</Chinesesimp>
-                <Chinese>AI 死亡前所能承受的傷害程度</Chinese>
-            </Key>
-        </Container>
+        <Key ID="STR_ACE_Medical_Damage_PlayerDamageThreshold_DisplayName">
+            <English>Player Damage Threshold</English>
```
SilentSpike: Would be tempted to rename these damage settings as they look more like a health coefficient than a damage threshold.

addons/medical_gui/stringtable.xml
```diff
@@ -1,11 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project name="ACE">
     <Package name="Medical GUI">
-        <Key ID="STR_ACE_Medical_GUI_Category">
-            <English>ACE Medical GUI</English>
-            <German>Sanitätsmenü</German>
-            <Japanese>ACE 医療 GUI</Japanese>
-            <Russian>ACE Медицина - Интерфейс</Russian>
+        <Key ID="STR_ACE_Medical_GUI_SubCategory">
+            <English>GUI</English>
```
SilentSpike:
```suggestion
            <English>Medical GUI</English>
```

SilentSpike: Actually, I guess this is under medical subcategory already? In which case ignore

addons/medical_statemachine/functions/fnc_handleStateUnconscious.sqf
SilentSpike: I think it's good this way. That would introduce the possibility of going unconscious for less time than the check interval.

addons/medical_statemachine/stringtable.xml
```diff
@@ -1,99 +1,69 @@
...
+            <Russian>Смертельные ранения игрока</Russian>
+        </Key>
+        <Key ID="STR_ACE_Medical_Statemachine_FatalInjuryCondition_Description">
+            <English>Controls when a player can receive a fatal injury.</English>
+            <German>Definiert wann ein Spieler tödliche Verletzungen bekommen kann</German>
+            <Japanese>プレイヤーが致命傷から蘇生できるかどうか決定します</Japanese>
+            <Russian>Определяет, когда игрок может получить смертельное ранение</Russian>
+        </Key>
+        <Key ID="STR_ACE_Medical_Statemachine_FatalInjuryConditionAI_DisplayName">
+            <English>AI Instant Death</English>
```
SilentSpike: I think this is too ambiguous, it's also weird how this setting only does anything if AI unconsciousness is enabled (because otherwise they always die instantly upon receiving a wound which would push them into unconsciousness).

addons/medical_status/stringtable.xml
```diff
@@ -1,89 +1,87 @@
...
+            <Chinese>修改疼痛強度的係數</Chinese>
+        </Key>
+        <Key ID="STR_ACE_Medical_Status_IvFlowRate_DisplayName">
+            <English>IV Transfusion Flow Rate</English>
+            <German>Transfusions Fließrate</German>
+            <Japanese>IV 輸血の流量</Japanese>
+            <Russian>Скорость внутривенного переливания</Russian>
+        </Key>
+        <Key ID="STR_ACE_Medical_Status_IvFlowRate_Description">
+            <English>Controls how quickly IV Bags will have an effect.</English>
```
SilentSpike: Could use a minor rewording here (as it could be interpreted as how much time passes before the IV starts).

mharis001: Agreed. Maybe: `Controls how quickly fluid flows out of IV Bags.`? Could also describe how exactly the coefficient affects the calculation.

mharis001:
Merge #7069 before merging this.

addons/medical_statemachine/stringtable.xml
SilentSpike: Could actually be a 3 way combination of both settings:
- AI can't go unconscious
- AI can go unconscious
- AI can receive a fatal injury (and go unconscious)

Just hard to find an appropriate title/wording

addons/medical_damage/stringtable.xml
SilentSpike: Believe in meeting we discussed something like "Critical Damage Threshold" > "Damage event which will cause a unit to fall unconscious" (reword as desired).

addons/medical_statemachine/stringtable.xml
SilentSpike: 🤔 Maybe it's actually fine as is, this is a really annoying setting design pattern, but I guess the unconscious and fatal injury systems are separate ultimately.

Would still reword this so that it's clear user is enabling/disabling fatal injuries for AI. That way when we have a section in the docs explaining fatal injuries it's clearly related to this setting.

addons/medical_statemachine/functions/fnc_handleStateUnconscious.sqf
PabstMirror: this only applies to when the setting is changed mid-mission from 0% chance, 
so I think it's just fine

addons/medical_damage/stringtable.xml
```diff
@@ -2,16 +2,16 @@
 <Project name="ACE">
     <Package name="Medical_Damage">
         <Key ID="STR_ACE_Medical_Damage_PlayerDamageThreshold_DisplayName">
-            <English>Player Damage Threshold</English>
+            <English>Player Critical Damage Threshold</English>
         </Key>
         <Key ID="STR_ACE_Medical_Damage_PlayerDamageThreshold_Description">
-            <English>Sets the amount of damage a player can take before being killed.</English>
+            <English>Sets the amount of damage a player can receive before going unconscious.</English>
```
SilentSpike: 🤔 I still think the description could be better (because it sounds cumulative here, but it's really for a single damage event). However, we can tweak it later 👍

PR #7060, opened 2019-06-19
Medical - Rework CPR and Bleeding in cardiac arrest

PabstMirror:
- Runs vitals loop during cardiac arrest (replace #7013)
Lets us get rid of `calculateBlood` in medical treatment and bring it back to a single func in vitals
- Set a floor on the effects of cardiac output on bleeding 
- Remove  `cprCreatesPulse`  setting
- Add setting to control success rate of CPR
- Hide CPR for basic diagnose because they can't tell difference between regular uncon
- Providing CPR now decrease time flow of cardiac arrest timeout


SilentSpike:
> Do we want a cardiacArrestStopsBleeding setting; for groups that use cardiac arrest as more of a revive state?

I've thought about this more and am in favour of no setting. One of the goals of the rewrite has been to remove the need for an explicit revive state and I think by adding that setting we're basically including one under a different name. Cardiac arrest is already a fine recovery stage for dying units.

> Do we still want cprCreatesPulse as a setting?

I vote no setting and just have the behaviour follow realism (I presume that'd be a weak pulse).

> Before CPR showed when unconscious, but failed on progessBar immedietly? Should we show and allow on unconscious or only show when in cardiac arrest?

I think showing while unconscious makes sense as you actually need to check if they're in cardiac arrest or not before blindly starting CPR. Based on [a quick search](https://resus.org.au/faq/will-i-do-harm-if-i-give-chest-compressions-to-someone-with-a-beating-heart) no harm would actually be done to a patient with a beating heart if performing CPR on them (though apparently there is little data).

SilentSpike:
The other option for showing while unconscious is to only show in cardiac arrest when the advanced treatment setting is disabled (if that still exists - I believe it adds/remove the status check type actions).

addons/medical_statemachine/functions/fnc_enteredStateCardiacArrest.sqf
```diff
@@ -22,8 +22,10 @@ params ["_unit"];
 private _time = GVAR(cardiacArrestTime);
 _time = _time + random [_time*-0.1, 0, _time*0.1];
```
SilentSpike:
```suggestion
_time = _time + _time * random [-0.1, 0, 0.1];
```

Just a very minor optimisation

addons/medical_vitals/functions/fnc_handleUnitVitals.sqf
```diff
@@ -113,6 +113,7 @@ switch (true) do {
         TRACE_3("BloodVolume Fatal",_unit,BLOOD_VOLUME_FATAL,_bloodVolume);
```
SilentSpike: Did we want to remove the bleedout death condition as part of these changes?

Not sure we ever reached a decision on that

addons/medical_vitals/functions/fnc_updateHeartRate.sqf
```diff
@@ -22,7 +22,15 @@ params ["_unit", "_hrTargetAdjustment", "_deltaT", "_syncValue"];
 
 private _heartRate = GET_HEART_RATE(_unit);
 
-if !IN_CRDC_ARRST(_unit) then {
+if IN_CRDC_ARRST(_unit) then {
+    if (alive (_unit getVariable [QEGVAR(medical,CPR_provider), objNull])) then {
+        if (_heartRate == 0) then { _syncValue = true }; // always sync on large change
```
SilentSpike:
```suggestion
        _syncValue = (_heartRate == 0); // always sync on large change
```

addons/medical_vitals/functions/fnc_updateHeartRate.sqf
```diff
@@ -22,7 +22,15 @@ params ["_unit", "_hrTargetAdjustment", "_deltaT", "_syncValue"];
...
+        _heartRate = random [25, 30, 35];
+    } else {
+        if (_heartRate != 0) then { _syncValue = true }; // always sync on large change
```
SilentSpike:
```suggestion
        _syncValue = (_heartRate != 0); // always sync on large change
```

addons/medical_vitals/functions/fnc_updateHeartRate.sqf
SilentSpike: Thinking about it more, I guess the reason you did it this way is we don't want to overwrite a `true` input value.

PabstMirror: yeah I felt that on start/stop of CPR it needs to update immediately,
so someone else checking pulse can see it
`_syncValue = _syncValue && {(_heartRate == 0){;`
would also work but that should be similar to what's already there

severgun:
update TODO?
https://github.com/acemod/ACE3/issues/6913

Arcanum417:
> > Do we want a cardiacArrestStopsBleeding setting; for groups that use cardiac arrest as more of a revive state?
> 
> I've thought about this more and am in favour of no setting. One of the goals of the rewrite has been to remove the need for an explicit revive state and I think by adding that setting we're basically including one under a different name. Cardiac arrest is already a fine recovery stage for dying units.

Well, I am running a unit that kinda uses a revive state with way outdated medical-rewrite compiled version. I see no reason to include a setting that would replicate a "revive" state as was in the old medical. This is a fine state granted, but we work on a presumtion of no respawn or perma-death. On missions we always try to gather our fallen and set up a medevac to eventually bring them back to fight. It creates a valuable game-play element to have that asset usefull. Adding a switch so we don't have to cobble some way around this would be nice.

I see the rewrite as really separating options from each other. Before it was everything bundled under "revive" and with rewrite we can define a lot of the finer details, ie. now we can set up that death is possible from massive damage like mines etc. and still have regular injuries result in revivable cassualities that last a long time.

PR #7061, opened 2019-06-19
Medical StateMachine - Tweak dead state handling

PabstMirror:
2 unrelated things, so we can separate if any problems
- Add death reason on death transition
Pretty simple little hints about final cause of death, might be useful.

- Don't call setDammage inside HandleDamage EH

Killed EH fires twice, I think because the `setDamage` is called while we are still inside the handleDamage code block.
The first killed EH is (`[unit, unit]` / suicide)
The second is actually correct (`[unit, killer]`)

master
```
230483: ace_medical_death - [t1]
230483: ace_killed - [t1,"unknown",z2,z2]
230483: addEventHandler killed - [t1,t1,<NULL-object>,true]
230483: cba killed eh - [t1,t1,<NULL-object>,true]
230483: addEventHandler killed - [t1,z2,z2,true]
230483: cba killed eh - [t1,z2,z2,true]
```
After the fix you only get suicides via vanilla killed EH.
So I'm not positive if this is something we want to fix, double killed EH might break some things, but not being able to ID the killer might break others.
pr
```
221573: ace_medical_death - [t3]
221573: ace_killed - [t3,"unknown",z2,z2]
221574: addEventHandler killed - [t3,t3,<NULL-object>,true]
221574: cba killed eh - [t3,t3,<NULL-object>,true]
```


PabstMirror:
Another possibility is if we can "kill" the unit from returning damage in "ACE_HDBracket" then we should be able to get a single killed EH with correct killer, but I haven't had any luck
(`passThrough` didn't work)

PabstMirror:
This (ab)uses the fact that EH params are the same array passed to each EH
So we can "fix" the killer.
The choice is if we delay a frame to fix the double killed EH.

### No delay
Engine will see [Killed Suicide, Killed Normal]
So scoreboard and `rating` should be semi-correct
Killed EH will fire twice, but have corrected killer both times

### Delay
Engine will see [Killed Suicide]
Killed EH will fire once with corrected killer


Potential problems are not adding the EH early enough, and then there's killedMP which I don't know what to do with as vars will be local only.
There also is the `ace_killed` CBA event which will only fire once, and also have cause of death (like `"CardiacArrest:Timeout"`, might be nice for acex_killtracker). If mission makers have problems, we can always try to get them to use this event.

addons/medical_statemachine/functions/fnc_enteredStateDeath.sqf
```diff
@@ -16,10 +16,12 @@
...
 params ["_unit"];
+if (isNull _unit) exitWith {};
 
-// TODO: Probably also needs additional logic to deal with edge cases
+//IGNORE_PRIVATE_WARNING ["_thisOrigin", "_thisTransition"]; // vars provided by CBA_statemachine
+TRACE_3("enteredStateDeath",_this,_thisOrigin,_thisTransition);
 
-// Send a local event before death
-[QEGVAR(medical,death), [_unit]] call CBA_fnc_localEvent;
+private _causeOfDeath = format ["%1:%2",_thisOrigin,_thisTransition];
```
TheMagnetar:
```suggestion
private _causeOfDeath = format ["%1:%2", _thisOrigin, _thisTransition];
```

Issue #654, opened 2015-04-16
Only display medical interactions if the player holstered their weapon


esteldunedain:
I propose forcing the player to holster their weapons to enable various interactions, specially medical.

Reasons:
- It's a bit weird to see people pointing their guns at person or a body. They might be trying to treat them, but it just doesn't look that way at all, which detract from the whole 3d space. 
- Holstering/drawing the weapon is fast an I don't think it would be cumbersome.
- This would cut down 90% of the actions during normal combat, making the interaction system much for fluid to use in for actions in general (tapping shoulders, opening doors, taking prisoners, etc). 

Other things:
- Maybe the "check pulse" action should remain; maybe some sort of diagnosing too.
- Maybe the "holster weapon" key should be made more accesible; probably add it to self interaction too to prevent people who forgot their keybind from treating.
- To the very minimum the weapon should be lowered to allow treatment.
- I'm not sure you can holster while prone, so maybe should still be able to treat with your weapon on that stance.
- This could be a setting too

We already do this for surrendering, but I think this idea could be used for other things too.

jaynus:
I'd recommend forcing lowering instead of holstering. Double tapping ctrl is a known vanilla action, versus a whole other feature

dixon13:
+1 jaynus's idea

commy2:
Problem is, that it still plays that ugly weapon-to-back animation then.

meat147:
I prefer ed version can we make this optional default opt out ?

Codingboy:
+1 jaynus's idea

esteldunedain:
Obviously people have different opinions about this one. People preferences may also be related to what medical level they use. Safest option (not necesarilly the best) would be to provide the three alternatives: 
- Always show medical actions
- Hide medical actions when weapon is raised
- Hide medical actions with weapon in hands

The discussion would then be which we designate as default.

jaynus:
#2 default, 3 choice setting

jonpas:
Agreed on above.

alganthe:
with the new animation on dev branch (air stitching) you won't get the weapon on back animation loop so forcing weapon holstered by default would be a good idea on advanced.

KoffeinFlummi:
#2 or #3 should **not** be the default settings.

[...]

PR #7063, opened 2019-06-19
Medical - Option for treatment to require holstering/lowering weapon.

ddm999:
Adds Medical options (off by default) to disallow treatment while a weapon is active or raised.
--
Based on #654

What this does:
- Adds a "Holster Required" entry to the ACE Medical options (under Treatment).
- - When set to "Holstered only", medical treatments can only be performed if there is no active weapon (either by not having a weapon or by pressing the ACE Weapons > Holster Weapon control).
- - When set to "Lowered or Holstered", medical treatments can only be performed if there is no active weapon, or the weapon is lowered.
- - When set to "Disabled" (default), nothing changes.
- Adds a "Bypass Holster Required for Examination" entry to the ACE Medical options (also under Treatment, off by default).
- - When checked, treatments in the "examine" category are allowed regardless of weapon state and "Holster Required" setting.
- - This allows the Check Pulse, Check Blood Pressure, Response actions to be performed without lowering or holstering.

What this ***doesn't*** do (for potential consideration):
- Hide the medical interaction from limbs that have no available options (so it doesn't really 'de-clutter', other than preventing the treatment options from appearing and disappearing while you're looking around a body with weapon raised).
- - This means you can still see the Injuries and Triage Card display on the left by hovering over limbs, if "Enable Medical Actions" isn't set to "Disabled".
- Wait for any unequip / lower animations to finish before you're allowed to perform medical actions.
- Prevent access to the Medical Menu, even if no treatments are available.
- Prevent access to the Triage Card via Torso.
- Give the player an indication of why they can't perform medical actions if settings are enabled but they have their gun raised / equipped.
- Distinguish between treatments to allow some be allowed when weapon is lowered while others only allowed when no weapon is active.
- Distinguish between self-treatment and treating others.

Noted potential "issues":
- when raising weapon and opening the Medical Menu quickly, there's a delay before some treatments disappear (note: they can't/won't be performed if clicked)
- at a glance, "Holster Required: Holstered only" doesn't really make much sense. Might need to change the DisplayName.

Also, I realise that the request this is based on is very, very old, and I'd imagine some of the reasoning behind why this was wanted no longer applies.
I still made this PR because it seemed relatively simple for someone with little experience of the project. :)


ddm999:
Just noticed this doesn't play nice with vehicles (keeps your state when you entered it), changed title to WIP and am going to look into it tomorrow.

ddm999:
Fixed the issue with vehicles (holster setting is basically ignored when either medic or patient is in a vehicle: so pointing a gun at a vehicle while you're healing someone inside is allowed).
If anyone's interested in any of the stuff I mentioned in the PR description (under "what this (currently) doesn't do") - let me know, and specify what you'd personally prefer. I'll see if I can add it to this PR as another non-default option. :)

Otherwise, this small optional feature is ready, and I'll remove the WIP tag if there's no extra features wanted.

addons/medical_treatment/functions/fnc_canTreat.sqf
```diff
@@ -27,6 +27,14 @@ isClass _config
 && {_medic != _patient || {GET_NUMBER_ENTRY(_config >> "allowSelfTreatment") == 1}}
 && {[_medic, GET_NUMBER_ENTRY(_config >> "medicRequired")] call FUNC(isMedic)}
 && {
+    // Handle holster settings
+    GVAR(holsterRequired) == 0
+    || !(vehicle _medic isEqualTo _medic) // medic is in a vehicle, so weapon is considered holstered
+    || !(vehicle _patient isEqualTo _patient) // patient is in a vehicle, ^
```
dedmen: A simple != might be faster than a ! and isEqualTo.

ddm999: Ah ok, from what I'd read on the wiki it sounded like isEqualTo might've been faster. Will correct now.

bux:
One of the best PR descriptions of a first time contributor. Cheers. <3

addons/medical_treatment/functions/fnc_canTreat.sqf
dedmen: isEqualTo is a TINY bit faster than ==.
But ! (x isEqualTo y) are more script commands and is slower.

addons/medical_treatment/functions/fnc_canTreat.sqf
```diff
@@ -27,13 +27,13 @@ isClass _config
...
 && {
-    // Handle holster settings
+    // Handle holster settings [disabled, lowered, loweredExam, holster, holsterExam]
     GVAR(holsterRequired) == 0
-    || vehicle _medic != _medic // medic is in a vehicle, so weapon is considered holstered
-    || vehicle _patient != _patient // patient is in a vehicle, ^
-    || {GVAR(holsterNoReqExamine) && getText (_config >> "category") == "examine"} // if examine bypass is on
-    || currentWeapon _medic isEqualTo "" // weapon is holstered
-    || {GVAR(holsterRequired) == 1 && weaponLowered _medic} // if just lowered is allowed
+    || {vehicle _medic != _medic} // medic is in a vehicle, so weapon is considered holstered
```
jonpas: I would separate all those conditions into a separate function, for clarity. They are all here for holstering only.

Max255PL:
We were testing this PR lately and found an issue with healing while prone.
It still recognizes the weapon as "up" and you can't perform any medical actions. 
Yes, you can put it on the back but sometimes it's a suboptimal solution.

A good solution for this would be to add a setting to check for weapon safety on, be it with combination with other settings or by itself.

PabstMirror:
If you are prone, your weapon is always lowered 😄

PR #7064, opened 2019-06-20
Item replacement framework

BaerMitUmlaut:
- A function that lets you register item replacements by class, with or without inheriting classes, and item type

Focus was on performance, should be pretty fast. Intended to be used for medical and the compat free logistics (#7040).

### DIFF
--- a/addons/common/functions/fnc_registerItemReplacement.sqf
+++ b/addons/common/functions/fnc_registerItemReplacement.sqf
@@ -76,8 +76,7 @@ private _fnc_replaceItems = {
 };
 
 // Setup on first run
-private _firstRun = isNil QGVAR(itemReplacements);
-if (_firstRun) then {
+if (isNil QGVAR(itemReplacements)) then {
     GVAR(itemReplacements) = [] call CBA_fnc_createNamespace;
     GVAR(inheritedReplacements) = [];
     GVAR(oldItems) = [];
@@ -104,7 +103,7 @@ _oldReplacements append _newItems;
 GVAR(itemReplacements) setVariable [_oldItem, _newItems];
 
 // Force item scan when new replacement was registered in PostInit
-if (!isNull ACE_player || _firstRun) then {
+if (!isNull ACE_player) then {
     GVAR(oldItems) = [];
 
     // Exec next frame to ensure full scan only runs once per frame
--- a/addons/common/functions/fnc_registerItemReplacement.sqf
+++ b/addons/common/functions/fnc_registerItemReplacement.sqf
@@ -1,0 +1,115 @@
+#include "script_component.hpp"
+/*
+ * Author: BaerMitUmlaut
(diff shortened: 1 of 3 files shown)


addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -0,0 +1,115 @@
...
+private _firstRun = isNil QGVAR(itemReplacements);
+if (_firstRun) then {
+    GVAR(itemReplacements) = [] call CBA_fnc_createNamespace;
+    GVAR(inheritedReplacements) = [];
+    GVAR(oldItems) = [];
+    ["loadout", _fnc_replaceItems] call CBA_fnc_addPlayerEventHandler;
+};
+
+// Save item replacement
+// $ prefix is used for types (numbers) and replacements with inherition
```
dedmen:
```suggestion
// $ prefix is used for types (numbers) and replacements with inheritance
```

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -0,0 +1,115 @@
...
+    private _newItems = _items - GVAR(oldItems);
+    if (_newItems isEqualTo []) exitWith {
+        GVAR(oldItems) = _items;
+    };
+
+    _newItems sort true; // Sort so all items of current class can be replaced at once
+    private _replacements = [];
+    private _cfgWeapons = configFile >> "CfgWeapons"; // Microoptimization
+
+    for "_i" from 0 to count _newItems - 1 do {
```
dedmen: reason that forEach can't be used here?

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -0,0 +1,115 @@
+#include "script_component.hpp"
+/*
+ * Author: BaerMitUmlaut
+ * Registers an event handler that replaces an item or item type with one or
+ * more other items.
+ *
+ * Arguments:
+ * 0: Item or item type ID to replace <STRING or NUMBER>
```
dedmen: what is "item type ID"?

addons/common/functions/fnc_registerItemReplacement.sqf
BaerMitUmlaut: Skipping elements, see L62.

addons/common/functions/fnc_registerItemReplacement.sqf
BaerMitUmlaut: https://github.com/acemod/ACE3/blob/master/addons/main/script_macros.hpp#L63-L97

dedmen: Missed Line 47 I guess ^^

PabstMirror:
Merged as part of #7224 (guess git didn't auto-close because of squash?)

PR #7074, opened 2019-06-27
Stop AI firing at unconscious units

SilentSpike:
This just reimplemented the behaviour from #5864

I believe we thought using `setUnconscious` command would make this unnecessary, but testing seems to show that's not the case.


commy2:
Does this need resetting on respawn?

PabstMirror:
statusEffect_set should handle respawn (new object = reset everything)

commy2:
I should've never doubted you.

PR #3923, opened 2016-06-14
Grenade + Chemlight Sounds

voiperr:
**When merged this pull request will:**
- Remove ridiculous clinking sound when grenades spawn.
- Add:
  - Pin-pulling sound when player throw a grenade.
  - Snapping sound when player throws a chemlight.
  - Ignition sound when player throws handflare.
  - (Forward compat with https://github.com/acemod/ACE3/tree/grenadeThrowing)

PR #7075, opened 2019-06-27
Fix grenade impact sounds

### LINKED ISSUE #7073: Missing grenade impact sounds
**Arma 3 Version:** `1.92` (stable)
**CBA Version:** `3.11.2` (stable)
**ACE3 Version:** `3.12.6` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
With ACE running, you get grenade pin pull sounds, but no grenade impact sounds (when grenade hits the floor or wall for example). 
With no mods loaded, you get grenade impact sounds, but no pull pin sounds (since ACE adds that).

**Steps to reproduce:**
https://www.youtube.com/watch?v=vbAb25obxCM&feature=youtu.be
Comparison with, and without ACE

Everywhere

jonpas:
- Fix #7073 

Were removed in #3923 "ridiculous clinking sound", but it also removes impact sounds - this commit reverts that change.

### DIFF
--- a/addons/grenades/CfgAmmo.hpp
+++ b/addons/grenades/CfgAmmo.hpp
@@ -3,7 +3,6 @@ class CfgAmmo {
     class Default;
     class Grenade: Default {
         GVAR(pullPinSound)[] = {"A3\sounds_f\weapons\grenades\Grenade_PullPin.wss", 1.5, 1, 10};
-        impactGroundSoft[] = {};
     };
 
     class FlareCore;


jonpas:
This does not appear to be impact sounds at all!

It is just a priming when you pull the pin, there is still no impact sound. :thinking:

jonpas:
Never mind, it does fix impact sounds, there are just none if you throw it on the floor at your feet.

However, the sound played next to pin sound is pretty bad, I suggest you have a look @commy2 and see if you notice it now.

jonpas:
I would close this, neither solution is perfect, but the one in `master` is preferably I think.

commy2:
See: https://github.com/acemod/ACE3/issues/7073#issuecomment-544237582

PR #7076, opened 2019-06-28
Medical - epiBoostsSpontaneousWakeUp

PabstMirror:
Idea I had to help bridge the gap between 
- basic medication which always does a `WakeUp` on epi use
- advanced medication where the only thing that will wake a unit up is the spontaneous wake up chance

This boosts the spontaneous wake up chance when the unit has epi in their system
I wasn't sure the best way to implement this boost, right now it makes the check interval shorter, which is very different from just boosting the random chance.

Is it realistic?
I'm not sure, it does seem like epi is used during CPR, but that doesn't really apply to unconsciousness.
I mainly want it to be able to get units up faster if they are actively being treated by medics.


nk3nny:
From a gameplay perspective: In Nopryl.no we have a custom mod which essentially does just that.  I am happy to report that it is useful and good option when running advanced medical.  *thumbs up*

alganthe:
From research, a vasoconstrictor is recommended along oxygenated blood when cardiac arrest is induced via hypovolemia.

Epi is a vasoconstrictor, however the doses required are much lower than in a pen and they are administered via an IV.

I'd say keeping the epi-pen is a good in between reality and gameplay as we already have a tad too many items.

addons/medical_statemachine/initSettings.sqf
```diff
@@ -41,3 +41,13 @@
     [1, 3600, 30],
     true
 ] call CBA_settings_fnc_init;
+
+[
+    QGVAR(epiBoostsSpontaneousWakeUp), "SLIDER",
+    [LSTRING(epiBoostsSpontaneousWakeUpDisplayName), LSTRING(epiBoostsSpontaneousWakeUp_Description)],
+    _categoryArray,
```
mharis001:
```suggestion
    [ELSTRING(medical,Category), LSTRING(SubCategory)]
```

addons/medical_statemachine/initSettings.sqf
```diff
@@ -41,3 +41,13 @@
...
+    [1,15,1,1], // [min, max, default value, trailing decimals (-1 for whole numbers only)]
```
mharis001:
```suggestion
    [1, 15, 1, 1],
```

addons/medical_statemachine/initSettings.sqf
```diff
@@ -41,3 +41,13 @@
...
+    true, // isGlobal
```
mharis001:
```suggestion
    true,
```

addons/medical_statemachine/initSettings.sqf
mharis001: Is the max value tied to `SPONTANEOUS_WAKE_UP_INTERVAL`?

addons/medical_statemachine/initSettings.sqf
```diff
@@ -41,3 +41,13 @@
     [1, 3600, 30],
     true
 ] call CBA_settings_fnc_init;
+
+[
+    QGVAR(epiBoostsSpontaneousWakeUp), "SLIDER",
+    [LSTRING(epiBoostsSpontaneousWakeUpDisplayName), LSTRING(epiBoostsSpontaneousWakeUp_Description)],
```
mharis001: Need to still add stringtable entries?

addons/medical_statemachine/initSettings.sqf
```diff
@@ -41,3 +41,13 @@
     [1, 3600, 30],
     true
 ] call CBA_settings_fnc_init;
+
+[
+    QGVAR(epiBoostsSpontaneousWakeUp), "SLIDER",
```
mharis001: Separate lines, to match other settings.

addons/medical_statemachine/initSettings.sqf
```diff
@@ -41,3 +41,13 @@
...
+    [LSTRING(epiBoostsSpontaneousWakeUpDisplayName), LSTRING(epiBoostsSpontaneousWakeUp_Description)],
+    _categoryArray,
+    [1,15,1,1], // [min, max, default value, trailing decimals (-1 for whole numbers only)]
+    true, // isGlobal
+    {[QGVAR(epiBoostsSpontaneousWakeUp), _this] call EFUNC(common,cbaSettings_settingChanged)},
+    true // Needs mission restart
```
mharis001: Shouldn't require mission restart?

addons/medical_statemachine/initSettings.sqf
```diff
@@ -41,3 +41,13 @@
...
     true
 ] call CBA_settings_fnc_init;
+
+[
+    QGVAR(epiBoostsSpontaneousWakeUp), "SLIDER",
+    [LSTRING(epiBoostsSpontaneousWakeUpDisplayName), LSTRING(epiBoostsSpontaneousWakeUp_Description)],
+    _categoryArray,
+    [1,15,1,1], // [min, max, default value, trailing decimals (-1 for whole numbers only)]
+    true, // isGlobal
+    {[QGVAR(epiBoostsSpontaneousWakeUp), _this] call EFUNC(common,cbaSettings_settingChanged)},
```
mharis001: Don't think this is needed?

addons/medical_statemachine/initSettings.sqf
PabstMirror: It boosts by increasing how often it's checked
`SPONTANEOUS_WAKE_UP_INTERVAL (15 sec) / GVAR(epiBoostsSpontaneousWakeUp)`
So setting range could really be anything, I increased max to 30x

addons/medical/stringtable.xml
```diff
@@ -33,6 +33,12 @@
             <French>La probabilité pour qu'une unité en état stable puisse reprendre conscience (état vérifié toutes les 15 secondes).</French>
             <Portuguese>A probabilidade que uma unidade com vitais estabilizados possa recuperar consciências (verificado a cada 15 segundos)</Portuguese>
         </Key>
+        <Key ID="STR_ACE_Medical_spontaneousWakeUpEpinephrineBoost_DisplayName">
+            <English>Epinephrine Increases Wake Up Chance</English>
+        </Key>
+        <Key ID="STR_ACE_Medical_spontaneousWakeUpEpinephrineBoost_Description">
+            <English>Increases how often spontaneous wake up checks happen when patient has Epinephrine in their system</English>
```
mharis001:
```suggestion
            <English>Increases how often spontaneous wake up checks happen when patient has Epinephrine in their system.</English>
```

PR #7078, opened 2019-06-28
Movement - shorten transition 'run lowered rifle' -> 'jump with rifle'

commy2:
- Currently when running with rifle up and pressing Vault action (default V iirc), you immediately jump, however, when running with rifle lowered and pressing Vault, your character has to transition to the run with weapon raised animation before jumping. Due to how these animations transition, the delay before the character jumps becomes about 0 to ~2 seconds randomly (depening on how far into the animation you are). Also, after the animation your weapon is raised, because the "jump with rifle" animation is played as "rifle raised, running" action.

- This PR adds a copy of the jump animation that is played as "rifle lowered, running" action and links from and links back to the run with rifle lowered animation. Therefore you now jump immediately when pressing Vault even though you are running with the rifle lowered, and after the jump, the weapon is still lowered.

- PR also changes the action for "sprinting with rifle lowered" to the jump animation instead of the vault animation, so now you jump when sprinting and pressing Vault while the weapon is lowered instead of having to slow down to running to jump.


jonpas:
Awesome!

PabstMirror:
Best part of this PR
In a a normal run with rifle lowered, and you hit V; it will jump and then you return to rifle lowered

I remember several times getting over a fence and then forgetting to re-lower rifle and then having stamina problems

Issue #7077, opened 2019-06-28
Mine defusal role


anfo55:
Hi, no problem to report, just a question about an ACE functionality.
Presently, any blufor with a defusal kit, engineer or not, can defuse a trip mine. Is this by design?

mharis001:
The `ace_explosives_requireSpecialist` setting controls whether the unit must an EOD to be able to defuse an explosive. [reference](https://github.com/acemod/ACE3/blob/master/addons/explosives/functions/fnc_canDefuse.sqf#L29)

An EOD is a unit which has the `ACE_isEOD` variable set on it as `true` or has the `explosiveSpecialist` unit trait. [reference](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_isEOD.sqf#L23)

If you want to require being an EOD to be able to defuse, enable the require specialist setting and assign your EOD by setting the `ACE_isEOD` variable to true on it.

Hope that helps.

anfo55:
Thank you @mharis001 for your reply
I'm really sorry to ask since your explanation is worded in such way that I should understand, but I'm still a bit lost with where to put the information you've supplied. If 1 unit out of 8 is to be EOD, in practical terms, which code would go where please?

commy2:
Isn't this a checkbox in the editor?

mharis001:
@anfo55 you could put `this setVariable ["ACE_isEOD", true];` in the init box of the unit. #7080 will add a 3DEN attribute for this.

anfo55:
Thankyou @mharis001 , I hope my little question and the new plan to implement into 3DEN is a popular choice and worth your effort. :)
If this helps with your implementation, I applied the `this setVariable ["ACE_isEOD", true];` to an already ACE Engineer player intended for the EOD role, and he was able to defuse, as you would expect, a regular trip and APERS mine. However, a small issue appears to remain, that another player -without- the `this setVariable ["ACE_isEOD", true];` can still defuse.

commy2:
Replace true with false for the unit that is not supposed to diffuse.

anfo55:
Ah, whoops! I wrongly assumed the true attribute on one unit, would make all others instantly false. Thanks @commy2

anfo55:
Didn't work unfortunately. Fire away with more questions if it helps you get to the bottom of this :)

commy2:
In the server settings, enable ace_explosives_RequireSpecialist.

anfo55:
[...]

PR #7080, opened 2019-06-28
Explosives - Add ace_isEOD 3DEN attribute

mharis001:
- Add 3DEN attribute to control if unit is an ace EOD
- #7077

### DIFF
--- a/addons/explosives/stringtable.xml
+++ b/addons/explosives/stringtable.xml
@@ -1040,9 +1040,11 @@
         </Key>
         <Key ID="STR_ACE_Explosives_IsEOD_DisplayName">
             <English>Is EOD</English>
+            <German>Kampfmittelbeseitigung</German>
         </Key>
         <Key ID="STR_ACE_Explosives_IsEOD_Description">
-            <English>Controls whether the unit is an explosive specialist</English>
+            <English>Controls whether the unit is an explosive specialist.</English>
+            <German>Steuert, ob die Einheit ein Sprengstoffspezialist ist.</German>
         </Key>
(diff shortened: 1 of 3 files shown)


addons/explosives/stringtable.xml
```diff
@@ -1038,5 +1038,11 @@
             <Polish>Czasomierz Wybuchu</Polish>
             <Russian>Таймер взрывчатки</Russian>
         </Key>
+        <Key ID="STR_ACE_Explosives_IsEOD_DisplayName">
+            <English>Is EOD</English>
```
commy2:
```suggestion
            <English>Is EOD</English>
            <German>Kampfmittelbeseitigung</German>

File: addons/explosives/stringtable.xml [hunk 2]
```diff
@@ -1038,5 +1038,11 @@
...
+        </Key>
+        <Key ID="STR_ACE_Explosives_IsEOD_Description">
+            <English>Controls whether the unit is an explosive specialist</English>
```
commy2:
```suggestion
            <English>Controls whether the unit is an explosive specialist.</English>
            <German>Steuert, ob die Einheit ein Sprengstoffspezialist ist.</German>

addons/explosives/stringtable.xml
bux: Sounds like a civilian job description. Usually in the BW they are called the same as in English EOD

PabstMirror:
would it be better to use `setUnitTrait ["explosiveSpecialist", true]`?

PR #7084, opened 2019-07-02
Small med stuff

PabstMirror:
don't show distance warning if still in vehicle with target (must of failed for a different reason)
make distance setting non-client settable


dedmen:
>don't show distance warning if still in vehicle with target

That was already in pre-rewrite. distance is probably using the non interpolated position, meaning remote objects are lagging. And in a fast flying helicopter or a fast driving car that caused issues.

addons/advanced_fatigue/XEH_postInit.sqf
```diff
@@ -48,7 +48,7 @@ if (!hasInterface) exitWith {};
             linearConversion [0, 1, (_this getVariable [QEGVAR(medical,pain), 0]), 1, 1.1, true];
         }] call FUNC(addDutyFactor);
         [QEGVAR(medical,bloodVolume), { // 100->1.0, 90->1.1, 80->1.2
-            linearConversion [6, 0, (_this getVariable [QEGVAR(medical,bloodVolume), 100]), 1, 2, true];
```
dedmen: Need to update comment above?

PabstMirror:
`ace_medical_gui_fnc_canOpenMenu` already had the distance/same vehicle check logic
this was just for the warning hint

PR #7091, opened 2019-07-05
Medical - add dummy cfgVehicle entries for old modules

PabstMirror:


addons/medical/CfgVehicles.hpp
```diff
@@ -1,7 +1,22 @@
 class CfgVehicles {
+    // Left as dumb modules so that old missions don't error about missing vehicles
```
jonpas:
```suggestion
    // Backwards compatibility
    // Left as dumb modules so that old missions don't error about missing vehicles
```

PR #7092, opened 2019-07-06
Safemode - add public setWeaponSafety function

ColdEvul:
- Added safeWeapon function allowing the weapon to be set on safe if it is not.

A tiny function i use to set people's weapons on safe at spawn and respawn. So here is a little donation.

### DIFF
--- a/addons/safemode/functions/fn_safeWeapon.sqf
+++ b/addons/safemode/functions/fn_safeWeapon.sqf
@@ -1,25 +1,33 @@
 /*
  * Author: Brostrom.A
- * Safe the given weapon if unlocked.
+ * Safe or unsafe the given weapon based on weapon state; locked or unlocked.
  *
  * Arguments:
  * 0: Unit <OBJECT>
  * 1: Weapon <STRING>
  * 2: Muzzle <STRING>
+ * 3: State <BOOL>
  *
  * Return Value:
  * None
  *
  * Example:
- * [ACE_player, currentWeapon ACE_player, currentMuzzle ACE_player] call ace_safemode_fnc_safeWeapon
+ * [ACE_player, currentWeapon ACE_player, currentMuzzle Ace_player, true] call ace_safemode_fnc_safeWeapon
  *
  * Public: yes
  */
 
-params ["_unit", "_weapon", "_muzzle"];
+params [
+    ["_unit", objNull, [objNull]], 
+    ["_weapon", objNull, [objNull]],
+    ["_muzzle", objNull, [objNull]],
+    ["_state", true, [true]]
+];
 
 private _safedWeapons = _unit getVariable [QGVAR(safedWeapons), []];
 
-if !(_weapon in _safedWeapons) then { 
+_weapon = configName (configFile >> "CfgWeapons" >> _weapon);
+
+if (((_state isEqualTo true) && ((_safedWeapons findIf {_weapon == _x}) == -1)) || ((_state isEqualTo false) && (!(_safedWeapons findIf {_weapon == _x}) == -1)))then {
     [_unit, _weapon, _muzzle] call FUNC(lockSafety);
 };
--- a/addons/safemode/functions/fn_safeWeapon.sqf
+++ b/addons/safemode/functions/fn_safeWeapon.sqf
@@ -28,6 +28,6 @@ private _safedWeapons = _unit getVariable [QGVAR(safedWeapons), []];
 
 _weapon = configName (configFile >> "CfgWeapons" >> _weapon);
 
-if ((_state && _weapon in _safedWeapons) || (!_state && _weapon in _safedWeapons)) then {
+if !(_state isEqualTo (_weapon in _safedWeapons)) then {
     [_unit, _weapon, _muzzle] call FUNC(lockSafety);
 };
--- a/addons/safemode/functions/fn_safeWeapon.sqf
+++ b/addons/safemode/functions/fn_safeWeapon.sqf
@@ -19,8 +19,8 @@
 
 params [
     ["_unit", objNull, [objNull]], 
-    ["_weapon", objNull, [objNull]],
-    ["_muzzle", objNull, [objNull]],
+    ["_weapon", "", [""]],
+    ["_muzzle", "", [""]],
     ["_state", true, [true]]
 ];
 
@@ -28,6 +28,6 @@ private _safedWeapons = _unit getVariable [QGVAR(safedWeapons), []];
 
 _weapon = configName (configFile >> "CfgWeapons" >> _weapon);
 
-if (((_state isEqualTo true) && ((_safedWeapons findIf {_weapon == _x}) == -1)) || ((_state isEqualTo false) && (!(_safedWeapons findIf {_weapon == _x}) == -1)))then {
+if ((_state && ((_safedWeapons findIf {_weapon == _x}) == -1)) || (!_state && (!(_safedWeapons findIf {_weapon == _x}) == -1)))then {
     [_unit, _weapon, _muzzle] call FUNC(lockSafety);
 };
--- a/addons/safemode/functions/fn_safeWeapon.sqf
+++ b/addons/safemode/functions/fn_safeWeapon.sqf
@@ -28,6 +28,6 @@ private _safedWeapons = _unit getVariable [QGVAR(safedWeapons), []];
 
 _weapon = configName (configFile >> "CfgWeapons" >> _weapon);
 
-if ((_state && ((_safedWeapons findIf {_weapon == _x}) == -1)) || (!_state && (!(_safedWeapons findIf {_weapon == _x}) == -1)))then {
+if ((_state && _weapon in _safedWeapons) || (!_state && _weapon in _safedWeapons)) then {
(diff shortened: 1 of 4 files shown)


addons/safemode/functions/fn_safeWeapon.sqf
```diff
@@ -0,0 +1,25 @@
...
+ * [ACE_player, currentWeapon ACE_player, currentMuzzle ACE_player] call ace_safemode_fnc_safeWeapon
+ *
+ * Public: yes
+ */
+
+params ["_unit", "_weapon", "_muzzle"];
+
+private _safedWeapons = _unit getVariable [QGVAR(safedWeapons), []];
+
+if !(_weapon in _safedWeapons) then {
```
commy2: Public functions should not be case sensitive.

ColdEvul:
is there a quick way to convert a array to lower case?

addons/safemode/functions/fn_safeWeapon.sqf
ColdEvul: is there a quick way to convert a array to lower case?

PabstMirror: `if ((_safedWeapons findIf {_weapon == _x}) == -1) then {`
should do it

commy2: The script uses config case. So you can just convert random case to config case:
```sqf
_weapon = configName (configFile >> "CfgWeapons" >> _weapon);
```

Dystopian:
maybe this function should get `true/false` parameter and lock/unlock accordingly?

ColdEvul:
Like that idea two
I make sure to add that to the function as well.

ColdEvul:
Note: Not tested cause I'm on traveling foot.

Can be checked code wise anyways. if the solution for the bool param is agreeable.

addons/safemode/functions/fn_safeWeapon.sqf
```diff
@@ -0,0 +1,33 @@
...
+    ["_weapon", objNull, [objNull]],
+    ["_muzzle", objNull, [objNull]],
+    ["_state", true, [true]]
+];
+
+private _safedWeapons = _unit getVariable [QGVAR(safedWeapons), []];
+
+_weapon = configName (configFile >> "CfgWeapons" >> _weapon);
+
+if (((_state isEqualTo true) && ((_safedWeapons findIf {_weapon == _x}) == -1)) || ((_state isEqualTo false) && (!(_safedWeapons findIf {_weapon == _x}) == -1)))then {
```
commy2: `_state isEqualTo true` is just `_state`.
`_state isEqualTo false` is just `!_state`.
There is no point in using `findIf` if we convert to config case. Array will only ever contain config case.

addons/safemode/functions/fn_safeWeapon.sqf
```diff
@@ -0,0 +1,33 @@
...
+ * Example:
+ * [ACE_player, currentWeapon ACE_player, currentMuzzle Ace_player, true] call ace_safemode_fnc_safeWeapon
+ *
+ * Public: yes
+ */
+
+params [
+    ["_unit", objNull, [objNull]], 
+    ["_weapon", objNull, [objNull]],
+    ["_muzzle", objNull, [objNull]],
```
commy2: _weapon and _muzzle default and expected values should be "", not objNull.

addons/safemode/functions/fn_safeWeapon.sqf
```diff
@@ -0,0 +1,33 @@
...
+    ["_weapon", "", [""]],
+    ["_muzzle", "", [""]],
+    ["_state", true, [true]]
+];
+
+private _safedWeapons = _unit getVariable [QGVAR(safedWeapons), []];
+
+_weapon = configName (configFile >> "CfgWeapons" >> _weapon);
+
+if ((_state && _weapon in _safedWeapons) || (!_state && _weapon in _safedWeapons)) then {
```
commy2: This logic doesn't work.

S: `_state`
W: `_weapon in _safedWeapons`

(S ∧ W) ∨ (¬S ∧ W)
W ∧ (S ∨ ¬S)
W ∧ T
W

So the function would only ever call `FUNC(lockSafety)` if W = `_weapon in _safedWeapons` is true: the function can only ever unlock a weapon.

I think the condition is just supposed to say:
`!(_state isEqualTo (_weapon in _safedWeapons))`

ColdEvul: it was sopuse to have been like this:
`((_state && _weapon in _safedWeapons) || (!_state && !(_weapon in _safedWeapons)))`
major typo on my part. 

whould your statement make the weapon locked if true and unlock in false?

commy2: It would lock it if _state is true and it isn't already locked, and unlock it if _state is false and it is locked atm.

ColdEvul: I see the logic now weary nice!

ColdEvul:
All done

commy2:
- Should the function be named setWeaponSafety?
- There is no reason for the user not to set muzzle to `currentMuzzle _unit`, therefore I think this should not be argument of the function, but just a variable set to `currentMuzzle _unit`.

ColdEvul:
>- Should the function be named setWeaponSafety?

Yes agree.

>- There is no reason for the user not to set muzzle to currentMuzzle _unit, therefore I think this should not be argument of the function, but just a variable set to currentMuzzle _unit.

Agree with this two.
Added it cause it was in the other function.

commy2:
For the internal function it is used to switch back to the last muzzle. It is needed, because you can turn off safety with F. Otherwise it would switch to the primary muzzle from the secondary muzzle for example in addition to unlocking the weapon, which feels weird. Not needed for this public function though.

commy2:
`_weapon` should be kept as argument. _muzzle cannot be converted to config case like this, and it is not needed, because currentMuzzle reports config case already.

ColdEvul:
Test in progress

addons/safemode/functions/fn_setWeaponSafety.sqf
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Example:
+ * [ACE_player, true] call ace_safemode_fnc_setWeaponSafety
+ *
+ * Public: no
+ */
+
+params [
+    ["_unit", objNull, [objNull]], 
+    ["_weapon", objNull, [objNull]],
```
commy2:
```suggestion
    ["_weapon", "", [""]],
```

addons/safemode/functions/fn_setWeaponSafety.sqf
```diff
@@ -0,0 +1,33 @@
...
+ * None
+ *
+ * Example:
+ * [ACE_player, true] call ace_safemode_fnc_setWeaponSafety
+ *
+ * Public: no
+ */
+
+params [
+    ["_unit", objNull, [objNull]],
```
commy2:
```suggestion
    ["_unit", objNull, [objNull]],
```
White space at end of line :)

commy2: There still is a space at the end of the line.

ColdEvul: now it should be merged.

addons/safemode/functions/fn_setWeaponSafety.sqf
```diff
@@ -0,0 +1,33 @@
...
+ * 1: Weapon <STRING>
+ * 2: State <BOOL>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [ACE_player, true] call ace_safemode_fnc_setWeaponSafety
+ *
+ * Public: no
```
commy2:
```suggestion
 * Public: yes
```
:)

jonpas: Fixed it to capitalized for you.

ColdEvul:
Tested and did some final fixes to the script.

**Testprotocol:**
- Line used `[player, currentWeapon player, true] call ace_safemode_fnc_setWeaponSafety` & `[player, currentWeapon player, false] call ace_safemode_fnc_setWeaponSafety`
- [x] Tested manually applied function via debug menu
- [x] Tested trigger set and unset safemode.

veteran29:
If this function is used for something like eg. "safestart" and the unit does not have weapon it will try to lock "empty" weapon.

Maybe it would be wise to add:
`if (_weapon == "") exitWith {};`

Unsafing the "empty" weapon results in an [error ](https://cdn.discordapp.com/attachments/409293587017367553/617475112098332722/20190831234527_1.jpg).

addons/safemode/functions/fnc_setWeaponSafety.sqf
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Public: Yes
+ */
+
+params [
+    ["_unit", objNull, [objNull]],
+    ["_weapon", "", [""]],
+    ["_state", true, [true]]
+];
+
```
veteran29:
```suggestion

if (_weapon == "") exitWith {};

```

PR #7098, opened 2019-07-11
Medical - Fix hitpoints on VR units

PabstMirror:


commy2:
Should this test run auto for all soldiers?

PabstMirror:
There is a similar check on unit init: https://github.com/acemod/ACE3/blob/master/addons/medical_engine/XEH_postInit.sqf#L21

I tested this with cup and rhs and saw no problems so I don't think this PR's test will be needed to be run that much after this.

PR #7101, opened 2019-07-11
Update Load Order for 1.94 Contact DLC

jonpas:
- Title

**Merge after 1.94 releases!**

### DIFF
--- a/addons/main/config.cpp
+++ b/addons/main/config.cpp
@@ -7,7 +7,7 @@ class CfgPatches {
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {
-            "A3_Data_F_Tank_Loadorder",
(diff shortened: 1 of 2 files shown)


addons/main/config.cpp
```diff
@@ -7,7 +7,7 @@ class CfgPatches {
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {
-            "A3_Data_F_Tank_Loadorder",
+            "A3_Data_F_Enoch_Loadorder",
             "A3_Data_F_Mod_Loadorder",
```
jonpas: Do we need this @PabstMirror ?

PabstMirror: I think we still do, nothing else in vanilla arma seems to require it

PR #7103, opened 2019-07-15
Medical Engine - Report any fatal head damage

PabstMirror:
Shoot man with autocannon
` incoming: _allDamages=[[2.21103,3,"Head"],[5.88152,4,"Body"],[5.04804,1.15436,"LeftArm"],[5.30232,1.27997,"RightArm"],[20.4579,1.1125,"LeftLeg"],[35.2891,1.05652,"RightLeg"]], _damageStructural=0.398899`

2.2 damage to head should be fatal, but it instead picks the largest leg damage, which takes a while to kill

### DIFF
--- a/addons/medical_engine/functions/fnc_handleDamage.sqf
+++ b/addons/medical_engine/functions/fnc_handleDamage.sqf
@@ -82,6 +82,11 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
 
     _allDamages sort false;
     (_allDamages select 0) params ["_receivedDamage", "", "_woundedHitPoint"];
+    if (_damageHead >= HEAD_DAMAGE_THRESHOLD) then {
+        TRACE_3("reporting fatal head damage instead of max",_damageHead,_receivedDamage,_woundedHitPoint);
+        _receivedDamage = _damageHead;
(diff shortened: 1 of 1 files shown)


dedmen:
Only special case for head? 5x hit in torso is also more lethal than 30x hit in leg.

bux:
> 5x hit in torso is also more lethal than 30x hit in leg.

depends on where one hits

SilentSpike:
I guess what we should really be doing is normalising the incoming damage before picking the maximum (since it scales differently for each bodypart).

This engine scaling also influenced the code here:
https://github.com/acemod/ACE3/blob/a8580d0298a08961558c2a8a37f62fdcc0be0876/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf#L98

PabstMirror:
There is a lot more that could be done,
- we could look into passing actual incomming damage to all hitpoints, so non `_isSelectionSpecific` wounds could be assigned to actual wounds hit, maybe use `selectWeightedRandom`
- body hits can be fatal as well under some conditions, so maybe pass body depending on ratio of max/body

For now this PR should be an improvement for any kind of large explosive damage taken.

PR #7109, opened 2019-07-18
Medical - Improve loading/unloading of unconc patients

PabstMirror:
- Fix unit being in locked animation with switchMove (If unit was unloaded from a vehicle, they may be in anim "deadstate" instead of "unconscious")
- Add hint when unloading unit (on success)
- Don't allow loading uncon into vehicles with `ejectDeadCargo=1` (e.g. quad bike)
- Added error logging (on both machines)

Also, removed use of `getvariable [QGVAR(loaded_persons)`, didn't see any reason why it existed

### DIFF
--- a/addons/common/functions/fnc_unloadPersonLocal.sqf
+++ b/addons/common/functions/fnc_unloadPersonLocal.sqf
@@ -20,11 +20,13 @@
 #define GROUP_SWITCH_ID QFUNC(loadPerson)
 
 params ["_unit", "_vehicle", ["_unloader", objNull]];
-TRACE_3("params",_unit,_vehicle,_unloader);
+TRACE_3("unloadpersonLocal",_unit,_vehicle,_unloader);
 
 //This covers testing vehicle stability and finding a safe position
 private _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition);
 TRACE_1("findUnloadPosition",_emptyPos);
+if (_emptyPos isEqualTo []) then { _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition); TRACE_1("findUnloadPosition 2nd attempt",_emptyPos); };
+if (_emptyPos isEqualTo []) then { _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition); TRACE_1("findUnloadPosition 3rd attempt",_emptyPos); };
 
 if (count _emptyPos != 3) exitwith {
     WARNING_4("Could not find unload pos %1-ASL: %2 isTouchingGround: %3 Speed: %4",_vehicle, getPosASL _vehicle, isTouchingGround _vehicle, speed _vehicle);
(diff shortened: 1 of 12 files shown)


addons/common/functions/fnc_unloadPersonLocal.sqf
```diff
@@ -20,11 +20,13 @@
...
 
 params ["_unit", "_vehicle", ["_unloader", objNull]];
-TRACE_3("params",_unit,_vehicle,_unloader);
+TRACE_3("unloadpersonLocal",_unit,_vehicle,_unloader);
 
 //This covers testing vehicle stability and finding a safe position
 private _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition);
 TRACE_1("findUnloadPosition",_emptyPos);
+if (_emptyPos isEqualTo []) then { _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition); TRACE_1("findUnloadPosition 2nd attempt",_emptyPos); };
+if (_emptyPos isEqualTo []) then { _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition); TRACE_1("findUnloadPosition 3rd attempt",_emptyPos); };
```
mharis001: Duplicated line

addons/common/functions/fnc_unloadPersonLocal.sqf
```diff
@@ -20,11 +20,13 @@
 #define GROUP_SWITCH_ID QFUNC(loadPerson)
 
 params ["_unit", "_vehicle", ["_unloader", objNull]];
-TRACE_3("params",_unit,_vehicle,_unloader);
+TRACE_3("unloadpersonLocal",_unit,_vehicle,_unloader);
 
 //This covers testing vehicle stability and finding a safe position
 private _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition);
 TRACE_1("findUnloadPosition",_emptyPos);
+if (_emptyPos isEqualTo []) then { _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition); TRACE_1("findUnloadPosition 2nd attempt",_emptyPos); };
```
mharis001: Might be more readable as multi line. Also, no need for parentheses around `typeOf _unit`.

addons/common/functions/fnc_unloadPersonLocal.sqf
dedmen: Seems to be intended.

mharis001: Yes, missed the "3rd attempt" part.

PabstMirror: probably because line was too wide  :smile:

addons/common/functions/fnc_unloadPersonLocal.sqf
```diff
@@ -20,11 +20,19 @@
...
+TRACE_3("unloadpersonLocal",_unit,_vehicle,_unloader);
 
 //This covers testing vehicle stability and finding a safe position
 private _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition);
 TRACE_1("findUnloadPosition",_emptyPos);
+if (_emptyPos isEqualTo []) then {
+    _emptyPos = [_vehicle, (typeOf _unit), _unloader] call EFUNC(common,findUnloadPosition);
+    TRACE_1("findUnloadPosition 2nd attempt",_emptyPos);
+};
+if (_emptyPos isEqualTo []) then {
```
TheMagnetar: This second if condition can be moved inside the first one.

addons/medical_treatment/functions/fnc_loadUnit.sqf
```diff
@@ -33,8 +34,20 @@ if (_patient call EFUNC(medical_status,isBeingDragged)) then {
 
 private _vehicle = [_medic, _patient, _vehicle] call EFUNC(common,loadPerson);
 
-if (!isNull _vehicle) then {
-    private _patientName = [_patient, false, true] call EFUNC(common,getName);
-    private _vehicleName = getText (configFile >> "CfgVehicles" >> typeOf _vehicle >> "displayName");
-    [[LSTRING(LoadedInto), _patientName, _vehicleName], 3] call EFUNC(common,displayTextStructured);
+if (isNull _vehicle) then {
```
TheMagnetar: If you want to avoid the if / else combination you can write it like `if (isNull _vehicle) exitWith { TRACE_1("no vehicle found",_vehicle);};` which may be cleaner at the end.

PR #7110, opened 2019-07-20
Medical - Add setting for pain effect intensity

LachlanSneddon:
- Allow users to modify the intensity of the pain effect


addons/medical_feedback/initSettings.sqf
```diff
@@ -14,3 +14,12 @@
...
 ] call CBA_Settings_fnc_init;
+
+[
+    QGVAR(painIntensity),
+    "SLIDER",
+    [LSTRING(PainIntensity_DisplayName), LSTRING(PainIntensity_Description)],
+    [ELSTRING(medical,Category), LSTRING(SubCategory)],
+    [0, 25, 0.6 , 2],
+    true
+] call CBA_settings_fnc_init;
```
commy2:
```suggestion
] call CBA_fnc_addSetting;

```

addons/medical_feedback/stringtable.xml
```diff
@@ -69,6 +69,11 @@
                 <Chinesesimp>启用伤者的尖叫声</Chinesesimp>
                 <Chinese>啟用傷者的尖叫聲</Chinese>
             </Key>
+            <Key ID="STR_ACE_Medical_feedback_PainIntensity_DisplayName">
+                <English>Pain Intensity</English>
```
commy2:
```suggestion
                <English>Pain Intensity</English>
                <German>Schmerzintensität</German>
```

addons/medical_feedback/stringtable.xml
```diff
@@ -69,6 +69,11 @@
...
+            </Key>
+            <Key ID="STR_ACE_Medical_feedback_PainIntensity_Description">
+                <English>Controls how intense the pain effect is</English>
```
commy2:
```suggestion
                <English>Controls intensity of the pain effect.</English>
                <German>Steuert die Intensität des Schmerzeffekts.</German>
```

commy2:
>Allow users to modify the intensity of the pain effect

You made it a global setting (which is the right way to do it imo), so technically you only allow the mission maker or server admin to modify the pain effect intensity and not the user.

addons/medical_feedback/initSettings.sqf
```diff
@@ -14,3 +14,12 @@
         [true] call FUNC(initEffects);
     }
 ] call CBA_Settings_fnc_init;
+
+[
+    QGVAR(painIntensity),
+    "SLIDER",
+    [LSTRING(PainIntensity_DisplayName), LSTRING(PainIntensity_Description)],
+    [ELSTRING(medical,Category), LSTRING(SubCategory)],
+    [0, 25, 0.6 , 2],
```
commy2: Are those reasonable limits?

LachlanSneddon: At 1 pain I've noticed that pulsing blur starts to look more like a kaleidoscope as the intensity values get higher. For white flashing the maximum is still bearable, so it could be higher. Will update will photo's to illustrate it better when I get the chance.

SilentSpike:
I imagine @BaerMitUmlaut will want to review this since he was working on pain effects.

Personally, I'm not convinced it's necessary.

mharis001:
I think it is generally better if coefficients have a default value of 1. Also, the setting could be split into one for each pain effect type?

The name could also be improved to `Pain Effect Intensity` or maybe some word other than intensity since, status already has a similar setting ([reference](https://github.com/acemod/ACE3/blob/master/addons/medical_status/stringtable.xml#L59)).

BaerMitUmlaut:
I'm currently reworking pain effects. Instead of making this adjustable I'd rather have good defaults. I don't think it needs to be adjustable.

LachlanSneddon:
> I'm currently reworking pain effects. Instead of making this adjustable I'd rather have good defaults. I don't think it needs to be adjustable.

Fair enough @BaerMitUmlaut, would you like me to close the request?

BaerMitUmlaut:
I think we can keep it open for now and if we think we need it later we can merge it then.

Thank you for your effort either way 👍

LachlanSneddon:
No problem, I've gain so many hours of gameplay from your mod, I thought it only fitting I should try and give some back.

addons/medical_feedback/stringtable.xml
```diff
@@ -69,6 +69,13 @@
                 <Chinesesimp>启用伤者的尖叫声</Chinesesimp>
                 <Chinese>啟用傷者的尖叫聲</Chinese>
             </Key>
+            <Key ID="STR_ACE_Medical_feedback_PainIntensity_DisplayName">
+                <English>Pain Intensity</English>
+                <German>Schmerzintensität</German>
+            </Key>
+            <Key ID="STR_ACE_Medical_feedback_PainIntensity_Description">
+                <English>Controls intensity of the pain effect.</English>
```
synixebrett:
```suggestion
                <English>Controls the intensity of the pain effect.</English>
```

PR #7112, opened 2019-07-22
Inventory - Adjust for changes to medical

PabstMirror:
- Inventory: adjust medical list for rewrite changes, also add vanilla medical items
- disableUserInput used setting `QEGVAR(medical,AllowChatWhileUnconscious)` but I could find no use of it in current or old medical https://github.com/acemod/ACE3/blob/backupBeforeRewriteMerge/addons/medical/ACE_Settings.hpp
I guess there is no reason we couldn't keep as a getVar?


addons/inventory/XEH_preStart.sqf
```diff
@@ -36,14 +36,11 @@ _grenades_ItemList = _grenades_ItemList arrayIntersect _grenades_ItemList;
 uiNamespace setVariable [QGVAR(Grenades_ItemList), _grenades_ItemList];
 
 // generate list of medical items
-private _medical_ItemList = [];
+private _medical_ItemList = ["FirstAidKit", "Medikit"];
```
commy2: What about the gm items or other hard coded ones? Probably should walk the config instead.

PabstMirror: now adds from https://github.com/acemod/ACE3/blob/master/addons/medical_treatment/CfgReplacementItems.hpp

commy2: But this is a different component. It should be able to list all medical items, even when using vanilla medical, because one deleted the medical components.

PabstMirror: even if you delete medical, compats will still add the array to configs

commy2: Ok, then you still need to handle the GM items, or do we have a GM compat?

PabstMirror: adjusted for changes in #7224
GM items not handled in this PR

PR #7114, opened 2019-07-24
Zeus - Fix possible enum error with assignTeam

PabstMirror:
```
Error Foreign error: Unknown enum value: ""
```
`assignedTeam` can now return ""
`assignTeam ""` can throw enum error on "", but only units in your group 
units in your group shouldn't return "",
so this might not even be necessary, but doesn't hurt to be safe

### DIFF
--- a/addons/zeus/functions/fnc_moduleGroupSide.sqf
+++ b/addons/zeus/functions/fnc_moduleGroupSide.sqf
@@ -48,7 +48,7 @@ if (GETVAR(_unit,ACE_isUnconscious,false) && {GETMVAR(EGVAR(medical,moveUnitsFro
     {
         private _team = assignedTeam _x;
         [_x] joinSilent _newGroup;
-        _x assignTeam _team;
+        if (_team != "") then { _x assignTeam _team; };
     } forEach units _unit;
     deleteGroup _oldGroup;
 };


addons/zeus/functions/fnc_moduleGroupSide.sqf
```diff
@@ -48,7 +48,7 @@ if (GETVAR(_unit,ACE_isUnconscious,false) && {GETMVAR(EGVAR(medical,moveUnitsFro
     {
         private _team = assignedTeam _x;
         [_x] joinSilent _newGroup;
-        _x assignTeam _team;
+        if (_team != "") then { _x assignTeam _team; };
```
commy2: Shouldn't  "" be treated as "MAIN" or whatever the name for "white" is? With this, "" means "keep the previous", which I don't think is right.

PabstMirror: it's after joining a new group, so fireteam was reset
doing nothing will leave in main

commy2: ok

SilentSpike: I feel like there could be a CBA function for changing group side to properly handle this stuff in a common location.

PR #7115, opened 2019-07-28
Interaction Menu - fix a crash when loading savegames

commy2:
- Currently any mission with sufficiently enough soldiers placed crashes when loading a savegame, because the game runs out of memory when deserializing arrays.
- The addInteractionToClass function from interact_menu with allowInheritance flag set adds an (retroactively applied) Init event handler to the object class for every added action, so the actions can be added to descendant classes of the base class the action was added to.
- ACE Medical adds a lot of actions to the CAManBase base class.
- This means ~250 more Init events on CAManBase when Medical is loaded.
- The array is referenced on every CAManBase object and thus creates a big save file, which cannot be loaded without running out of memory.

- This PR rewrites the allowInheritance part of the addInteractionToClass function to drastically reduce the memory usage.
- Realistically this is just a BI bug we will keep running into, but we can avoid it with this memory optimization.

### DIFF
--- a/addons/interact_menu/XEH_preInit.sqf
+++ b/addons/interact_menu/XEH_preInit.sqf
@@ -84,9 +84,9 @@ if (isNil QGVAR(inheritedActions)) then {
     private _type = typeOf _object;
 
     {
-        _x params ["_addedClasses", "_class", "_typeNum", "_parentPath", "_action"];
+        _x params ["_addedClasses", "_objectType", "_typeNum", "_parentPath", "_action"];
 
-        if (_object isKindOf _class && {_addedClasses pushBackUnique _type != -1}) then {
+        if (_object isKindOf _objectType && {_addedClasses pushBackUnique _type != -1}) then {
             [_type, _typeNum, _parentPath, _action] call FUNC(addActionToClass);
         };
     } forEach GVAR(inheritedActions);
--- a/addons/interact_menu/XEH_preInit.sqf
+++ b/addons/interact_menu/XEH_preInit.sqf
@@ -83,7 +83,7 @@ GVAR(inheritedManClasses) = [];
     params ["_object"];
     private _type = typeOf _object;
 
-    if ((_object isKindOf "CaManBase") && {GVAR(inheritedManClasses) pushBackUnique _type != -1}) then {
+    if ((_object isKindOf "CAManBase") && {GVAR(inheritedManClasses) pushBackUnique _type != -1}) then {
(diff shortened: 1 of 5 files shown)


commy2:
Thanks @Dystopian :)

PabstMirror:
Most inherited actions were for CAManBase (from medical)
So I added special handling for just that class,
it lets us do one check of `isKindOf && pushBackUnique` per unit init instead of once for each action

117 Units
Before
Init: 0.871338s
Place 8 Units: 0.0100098s / 8 = 1.25122ms

After
Init: 0.585205s
Place 8 Units: 0.000976563s / 8 = 0.12207ms

PabstMirror:
This will break WIP https://github.com/acemod/ACE3/pull/6295
But I think it will actually be much easier to implement

addons/interact_menu/XEH_preInit.sqf
```diff
@@ -75,14 +75,20 @@ GVAR(lastTimeSearchedActions) = -1000;
...
-};
+GVAR(inheritedActions) = [];
+GVAR(inheritedManActions) = []; // Actions only for CAManBase
+GVAR(inheritedManClasses) = [];
 
 ["All", "InitPost", {
     params ["_object"];
     private _type = typeOf _object;
 
+    if ((_object isKindOf "CaManBase") && {GVAR(inheritedManClasses) pushBackUnique _type != -1}) then {
```
commy2:
```suggestion
  if ((_object isKindOf "CAManBase") && {GVAR(inheritedManClasses) pushBackUnique _type != -1}) then {
```

addons/interact_menu/functions/fnc_addActionToClass.sqf
```diff
@@ -28,21 +28,24 @@ if (!params [["_objectType", "", [""]], ["_typeNum", 0, [0]], ["_parentPath", []
 TRACE_4("params",_objectType,_typeNum,_parentPath,_action);
 
 if (param [4, false, [false]]) exitwith {
-    if (isNil QGVAR(inheritedActions)) then {
-        GVAR(inheritedActions) = [];
-    };
-
-    private _addedClasses = [];
-    GVAR(inheritedActions) pushBack [_addedClasses, _objectType, _typeNum, _parentPath, _action];
+    if (_objectType == "CaManBase") then {
```
commy2:
```suggestion
    if (_objectType == "CAManBase") then {
```

PR #4497, opened 2016-10-01
Add GAU-8 Enhancement

sargken:
This is the same concept done as here [#1976](https://github.com/acemod/ACE3/pull/1976). 
Adds in the Enhanced GAU-8 Mod
Credit goes to @AlexusDE and @ZabuzaW

Has been tested in a single player environment. 

The following list presents all added/changed features:
- changed the calibre to 30mm
- increased muzzle velocity to 1030 m/s
- increased direct and indirect hit damage
- increased bullet spread
- added 2 burst modes: “Low” (1.5s, 2100rpm) and “High” (1.5s, 4200rpm)
- magazine size of 1174 rounds
- changed weapon sound (featuring the typical GAU-8 “fart of death”)
- new impact sound
- smoke effect while shooting the gun
- new impact effects (more smoke, flashing light of explosions)

Further Info on the [mod](https://forums.bistudio.com/topic/178360-mighty-gau-8a-avenger/)

PR #7118, opened 2019-08-01
Changes A-10/A-164 Cannon Values

LorenLuke:
-Reduce A-10/A-164 Cannon dispersion to realistic levels
-Tweak current "Gatling_30mm_HE_Plane_CAS_01_F" damage values
-Add Armor Piercing ammo type for 30mm Gatling
-Add Combat Mix ammo types (4:1 and 5:1 AP:HEI) and magazines for 30 mm Gatling
-Allows full rate of fire at lower framerates than was otherwise possible

### DIFF
--- a/addons/aircraft/CfgAmmo.hpp
+++ b/addons/aircraft/CfgAmmo.hpp
@@ -39,7 +39,7 @@ class CfgAmmo {
 
     // also adjust tracer, "muh lightshow"; also adjust splash damage radius
     class Gatling_30mm_HE_Plane_CAS_01_F: BulletBase {
-        hit = 40; // default: 180
+        hit = 70; // default: 180
         indirectHit = 12; // default: 4
         indirectHitRange = 4; // default: 3
         caliber = 1.4; // default: 5
@@ -48,28 +48,44 @@ class CfgAmmo {
         tracerStartTime = 0.02; // default: 0.1
         timeToLive = 40; // default: 6
     };
-
-    // adjust damage and splash damage, closer to bluefor gatling with same caliber
-    class Cannon_30mm_HE_Plane_CAS_02_F: Gatling_30mm_HE_Plane_CAS_01_F {
-        hit = 65; // default: 150
-        indirectHit = 10; // default: 4
-        indirectHitRange = 3; // default: 3
-    };
-
+	
     // adjust damage and splash damage, AP Rounds
-    class Cannon_30mm_HE_Plane_CAS_03_F: Gatling_30mm_HE_Plane_CAS_01_F {
-        hit = 75; // default: 150
-        indirectHit = 1; // default: 4
-        indirectHitRange = 0.25; // default: 3
+    class ACE_Gatling_30mm_AP_Plane_CAS_01_F: Gatling_30mm_HE_Plane_CAS_01_F {
+        explosive = 0.05; // default: 0.4
+		hit = 110; // default: 180
+        indirectHit = 2.5; // default: 4
+        indirectHitRange = 1; // default: 3
+		fuseDistance = 0.2; // default: 10
         deflecting = 5;
     };
 
-    class SubmunitionBase;
-    class ACE_30mm_mix: SubmunitionBase {
-        simulation = "shotSubmunitions";
-        submunitionAmmo[] = {"Gatling_30mm_HE_Plane_CAS_03_F",0.8,"Cannon_30mm_HE_Plane_CAS_01_F",0.2};
+	// adds submunition logic, enabling multiple rounds per frame
+	class ACE_Gatling_30mm_Sub_HEI: SubmunitionBullet {
+        submunitionAmmo = "Gatling_30mm_HE_Plane_CAS_01_F";
         weaponType = "cannon";
-        model = "\A3\Weapons_f\empty";
+		submunitionConeType[] = {"poissondisccenter", 3};
+		submunitionConeAngle = 0.056; // in degrees, 0.055 ~= 0.001 mils minute, but present
+        model = "\A3\Weapons_f\Data\bullettracer\tracer_red.p3d";
         triggerTime = 0.005;
+	};
+
+	class ACE_Gatling_30mm_Sub_AP: ACE_Gatling_30mm_Sub_HEI {
+		submunitionAmmo = "ACE_Gatling_30mm_AP_Plane_CAS_01_F";
+	};
+	
+    class ACE_Gatling_30mm_Sub_CM41: ACE_Gatling_30mm_Sub_HEI {
+        submunitionAmmo[] = {"ACE_Gatling_30mm_AP_Plane_CAS_01_F",0.8,"Cannon_30mm_HE_Plane_CAS_01_F",0.2};
+    };
+
+    class ACE_Gatling_30mm_Sub_CM51: ACE_Gatling_30mm_Sub_HEI {
+        submunitionAmmo[] = {"ACE_Gatling_30mm_AP_Plane_CAS_01_F",0.83,"Cannon_30mm_HE_Plane_CAS_01_F",0.17};
+    };
+
+
+    // adjust damage and splash damage, closer to bluefor gatling with same caliber
+    class Cannon_30mm_HE_Plane_CAS_02_F: Gatling_30mm_HE_Plane_CAS_01_F {
+        hit = 70; // default: 150
+        indirectHit = 11; // default: 4
+        indirectHitRange = 3; // default: 3
     };
 };
--- a/addons/aircraft/CfgAmmo.hpp
+++ b/addons/aircraft/CfgAmmo.hpp
@@ -39,9 +39,9 @@ class CfgAmmo {
 
     // also adjust tracer, "muh lightshow"; also adjust splash damage radius
     class Gatling_30mm_HE_Plane_CAS_01_F: BulletBase {
-        hit = 80; // default: 180
+        hit = 40; // default: 180
         indirectHit = 12; // default: 4
-        indirectHitRange = 3; // default: 3
+        indirectHitRange = 4; // default: 3
         caliber = 1.4; // default: 5
         deflecting = 3; // default: 5
         fuseDistance = 3; // default: 10
@@ -51,8 +51,29 @@ class CfgAmmo {
 
     // adjust damage and splash damage, closer to bluefor gatling with same caliber
     class Cannon_30mm_HE_Plane_CAS_02_F: Gatling_30mm_HE_Plane_CAS_01_F {
-        hit = 70; // default: 150
-        indirectHit = 11; // default: 4
+        hit = 65; // default: 150
+        indirectHit = 10; // default: 4
         indirectHitRange = 3; // default: 3
     };
+
+    // adjust damage and splash damage, AP Rounds
+    class Cannon_30mm_HE_Plane_CAS_03_F: Gatling_30mm_HE_Plane_CAS_01_F {
+        hit = 75; // default: 150
+        indirectHit = 1; // default: 4
+        indirectHitRange = 0.25; // default: 3
+        deflecting = 5
+    };
+
+    
+    class SubmunitionBase;
+    class ACE_30mm_mix : SubmunitionBase {
+        simulation = "shotSubmunitions";
+        submunitionAmmo[] = {"Gatling_30mm_HE_Plane_CAS_03_F",0.8,"Cannon_30mm_HE_Plane_CAS_01_F",0.2};
+        weaponType = "cannon";
+        model = "\A3\Weapons_f\empty";
+        triggerTime = 0.005;
+    };
+    
+    
+    
 };
--- a/addons/aircraft/CfgAmmo.hpp
+++ b/addons/aircraft/CfgAmmo.hpp
@@ -1,6 +1,6 @@
 class CfgAmmo {
     class BulletBase;
-    class B_20mm : BulletBase {
+    class B_20mm: BulletBase {
         hit = 80;
         indirectHit = 12;
         indirectHitRange = 2; //2;
@@ -15,8 +15,8 @@ class CfgAmmo {
         model = "\A3\Weapons_f\Data\bullettracer\tracer_red";
     };
 
-    class ACE_20mm_HE : B_20mm {};
-    class ACE_20mm_AP : B_20mm {
+    class ACE_20mm_HE: B_20mm {};
+    class ACE_20mm_AP: B_20mm {
         hit = 50;
         indirectHit = 12;
         indirectHitRange = 0.3; //2;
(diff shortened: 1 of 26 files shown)


addons/aircraft/CfgWeapons.hpp
```diff
@@ -16,8 +16,16 @@ class CfgWeapons {
         magazineReloadTime = 0.1;
     };
 
-    // bigger mag for comanche
     class CannonCore;
+    // Fix attrocious A-10 Cannon Dispersion
+    class Gatling_30mm_Plane_CAS_01_F : CannonCore {
```
commy2: White space before :

commy2:
Lots of white space.

addons/aircraft/CfgWeapons.hpp
```diff
@@ -16,8 +16,16 @@ class CfgWeapons {
...
     };
 
-    // bigger mag for comanche
     class CannonCore;
+    // Fix attrocious A-10 Cannon Dispersion
+    class Gatling_30mm_Plane_CAS_01_F: CannonCore {
+        class LowROF: Mode_FullAuto {
+            burst = 1;
+            reloadtime = 0.015;
+            dispersion = 0.0028; //0.279508497 = 0.25 * sqrt(0.8^-1); (80%, 5mil. https://en.wikipedia.org/wiki/GAU-8_Avenger#Accuracy) - Luke
```
commy2: Huh, where does this formula come from?

addons/aircraft/CfgWeapons.hpp
```diff
@@ -16,8 +16,17 @@ class CfgWeapons {
         magazineReloadTime = 0.1;
     };
 
-    // bigger mag for comanche
     class CannonCore;
+    // Fix attrocious A-10 Cannon Dispersion
+    class Gatling_30mm_Plane_CAS_01_F: CannonCore {
+        class LowROF: Mode_FullAuto {
+            burst = 1;
+            reloadtime = 0.015;
```
commy2: You need over 60 FPS to fire at this rate. This is impossible in multiplayer, the fire rate will vary depending on your FPS. This is not desirable.
Vanilla has 0.03, which is just about 34 FPS for constant fire rate.

addons/aircraft/CfgMagazines.hpp
```diff
@@ -4,6 +4,11 @@ class CfgMagazines {
     class 1000Rnd_Gatling_30mm_Plane_CAS_01_F: VehicleMagazine {
         count = 1170;
     };
+    
+    class ACE_1000Rnd_Gatling_30mm_Plane_CAS_CM: VehicleMagazine {
```
commy2: What does "CM" stand for?

Drofseh: Combat Mix probably

addons/aircraft/CfgAmmo.hpp
```diff
@@ -51,8 +51,25 @@ class CfgAmmo {
...
+    class Cannon_30mm_HE_Plane_CAS_03_F: Gatling_30mm_HE_Plane_CAS_01_F {
+        hit = 75; // default: 150
+        indirectHit = 1; // default: 4
+        indirectHitRange = 0.25; // default: 3
+        deflecting = 5;
+    };
+
+    class SubmunitionBase;
+    class ACE_30mm_mix: SubmunitionBase {
+        simulation = "shotSubmunitions";
```
commy2: SubmunitionBase already inherits `simulation = "shotSubmunitions";` from `SubmunitionCore`.

addons/aircraft/CfgAmmo.hpp
```diff
@@ -51,8 +51,25 @@ class CfgAmmo {
...
+    // adjust damage and splash damage, AP Rounds
+    class Cannon_30mm_HE_Plane_CAS_03_F: Gatling_30mm_HE_Plane_CAS_01_F {
+        hit = 75; // default: 150
+        indirectHit = 1; // default: 4
+        indirectHitRange = 0.25; // default: 3
+        deflecting = 5;
+    };
+
+    class SubmunitionBase;
+    class ACE_30mm_mix: SubmunitionBase {
```
commy2: Can we get a more descriptive classname please? Also, wouldn't `SubmunitionBullet` be a better base class?

addons/aircraft/CfgAmmo.hpp
```diff
@@ -51,8 +51,25 @@ class CfgAmmo {
...
+        simulation = "shotSubmunitions";
+        submunitionAmmo[] = {"Gatling_30mm_HE_Plane_CAS_03_F",0.8,"Cannon_30mm_HE_Plane_CAS_01_F",0.2};
+        weaponType = "cannon";
+        model = "\A3\Weapons_f\empty";
+        triggerTime = 0.005;
```
commy2: Other classes like this have considerably more config entries. For example:
```cpp
		class B_762x51_Minigun_Tracer_Red: SubmunitionBullet
		{
			hit = 11.6;
			indirectHit = 0;
			indirectHitRange = 0;
			model = "\A3\Weapons_f\Data\bullettracer\tracer_red";
			caliber = 3.6;
			deflecting = 25;
			explosive = 0;
			triggerTime = 0.1;
			submunitionConeAngle = 0;
			submunitionCount = 1;
			submunitionAmmo = "B_762x51_Minigun_Tracer_Red_splash";
			cartridge = "FxCartridge_762";
			visibleFire = 5;
			audibleFire = 70;
			cost = 1.2;
			airLock = 1;
			typicalSpeed = 800;
			tracerScale = 0.6;
			tracerStartTime = 0.0075;
			tracerEndTime = 5;
			airFriction = -0.001;
			class CamShakeExplode
			{
				power = 2.82843;
				duration = 0.6;
				frequency = 20;
				distance = 8.48528;
			};
			class CamShakeHit
			{
				power = 5;
				duration = 0.2;
				frequency = 20;
				distance = 1;
			};
		};
```
____

`subMunitionCount` may be able to have us simulate two or three projectiles from just a single shot, which would could make us reduce the rof to something that works in MP. Could be combined with that one entry that makes the ammo count show a multiple of the actually remaining rounds.

addons/aircraft/CfgAmmo.hpp
```diff
@@ -51,8 +51,25 @@ class CfgAmmo {
...
     class Cannon_30mm_HE_Plane_CAS_02_F: Gatling_30mm_HE_Plane_CAS_01_F {
-        hit = 70; // default: 150
-        indirectHit = 11; // default: 4
+        hit = 65; // default: 150
+        indirectHit = 10; // default: 4
         indirectHitRange = 3; // default: 3
     };
+
+    // adjust damage and splash damage, AP Rounds
+    class Cannon_30mm_HE_Plane_CAS_03_F: Gatling_30mm_HE_Plane_CAS_01_F {
```
commy2: This is a new classname created by ACE. Classnames are in global space, so it needs to be tagged.

addons/aircraft/CfgWeapons.hpp
LorenLuke: `The GAU-8/A accuracy when installed in the A-10 is rated at "5 mil, 80 percent", meaning that 80 percent of rounds fired will hit within a cone with an angle of five milliradians; this equates to a 40-foot (12-meter) diameter circle at the weapon's design range of 4,000 feet (1,200 m).`

If 80% of all rounds fall within 5 mils, then dispersion is set such that 80% of the area of a given circle is within 5 mils. 
5mils = +/- 0.0025 radian; 
area of circle = pi * r^2; 
area must equal 1/0.8; 
r = 0.0025 *sqrt(0.8^-1);

commy2: π is gone in your formula.
Formula of type `sqrt (x^-1)` should be transposed to `1 / sqrt x`.

I got
r = 1/√(0.8*π)

Are we sure Arma dispersion goes up and down? (Thus halfing the 5 mils effectively).

Are we sure this Wikipage is talking about 1 circle = 6400 mils and not 2π*2000 = 6283 mils? I suppose that one makes little difference.

SilentSpike:
There's a history of PRs to make changes to the A-10 cannon, see #4497

Previously we ended up closing them after consulting someone knowledgeable as it wasn't clear that the changes were realistic (for some details the opposite seemed true). However, that's not to say this PR should suffer the same fate - I would just recommend providing some reference for these changes.

commy2:
I see references for the dispersion change. I don't see references for the ammo mix (ratio) and the functions of these two ammo types.
Undoubtedly the m134 gau-17 gatling gun has a higher fire rate than what is, with consideration of the game's simulation frame rate, set by the base game.
Most of my concerns regarding this PR (and probably all of the old ones) are on the technical side.

addons/aircraft/CfgWeapons.hpp
LorenLuke: But π is present in both circles. So let us use a distance measure as a replacement of an angular measure. So, 5 mils is (basically) 5 meters at 1000 meters, so we'll say 5 meters for our radius in our '80%' circle (K) when compared to our dispersion circle (C).

`π(r_C)^2 = 0.8 * π(r_K)^2
(r_C)^2 = 0.8 * (r_K)^2
(r_C)^2 / 0.8 =(r_K)^2
√[(r_C)^2 / 0.8] = r_K
r_C/√[0.8] = r_K
;Substitute 0.005 for r_C;
0.005/√[0.8] = r_K
;Substitute 0.8944 for √[0.8];
0.005/0.8944 = r_K
;Solve;
0.00559 = r_K`

So you, the dispersion *should* be doubled, as the above is radial value. I'll get on that.

addons/aircraft/CfgMagazines.hpp
LorenLuke: @Drofseh is correct.

LorenLuke:
>ammo mix ratio
https://en.wikipedia.org/wiki/GAU-8_Avenger#Design
The standard ammunition mixture for anti-armor use is a five-to-one mix of PGU-14/B Armor Piercing Incendiary, with a projectile weight of about 14.0 oz (395 grams or 6,096 grains) and PGU-13/B High Explosive Incendiary (HEI) rounds, with a projectile weight of about 13.3 oz (378 grams or 5,833 grains).[8] The PGU-14/B's projectile incorporates a lightweight aluminum body, cast around a smaller caliber depleted uranium penetrating core.`

https://fas.org/man/dod-101/sys/ac/equip/gau-8.htm
`A typical combat load for the GAU-8 would include 1,100 rounds of 30mm high explosive or armor piercing ammunition. The 30mm API is mixed with 30mm High Explosive Incendiary (HEI) at the factory and is called Combat Mix Ammunition. The ratio of API to HEI rounds in the Combat Mix is 4:1. Combat mix is a sequential mixture of DU and HEI rounds in which 1 HEI round followed by 4 DU rounds are fired by the AN/GAU-8 gatling gun. DU is the primary munition for the A/OA-10 in a combat environment.`

Drofseh:
I like the approach with the submunitions and support this being included.

Here are some things I would like to see.
- More verbose classnames, for these in particular `ACE_1000Rnd_Gatling_30mm_Plane_CAS_CM`, `ACE_30mm_mix`, and `Cannon_30mm_HE_Plane_CAS_03_F` (which should both be tagged with `ACE_` as commy says, and also I think should be AP, not HE).
- A single rate of fire. The original selectable rate was only available very early on and was replaced with just one 3900 rpm setting. 
- The sources I've seen for the combat mix either say 5:1 AP to HE or 4:1 AP to HE. I'm not sure which is correct and it's possible that it changed at some point, but I'd prefer 4:1 (which is currently what the code has).

LorenLuke:
Progress commit-
Changed some values back to before (misread a classname, oops!)
Tweaked: Values for HEI rounds
Tweaked: Added more verbose class names for ammunition and magazines
Tweaked: Time between shots now 3x longer, but fires in 3rnd submunition bursts. Now supports full, uninterrupted maximum ROF at ~20FPS or higher.
Added: Dedicated AP round and magazine for 30mm cannon
Added: Combat Mix 4:1 ammo and magazine
Added: Combat Mix 5:1 ammo and magazine

addons/aircraft/CfgAmmo.hpp
LorenLuke: Done and implemented for the 30mm cannon. Each bullet fired as a submunition is 3 rounds that spread out, and each time the cannon is fired, 3 bullets are removed from the magazine. To maintain 60 "rounds" per second, only ~20-22FPS are needed, instead of 60.

addons/aircraft/CfgMagazines.hpp
```diff
@@ -2,9 +2,28 @@ class CfgMagazines {
...
     class VehicleMagazine;
     class 1000Rnd_Gatling_30mm_Plane_CAS_01_F: VehicleMagazine {
+        ammo = "ACE_Gatling_30mm_Sub_HEI";
         count = 1170;
     };
 
+    class ACE_1000Rnd_Gatling_30mm_Plane_CAS_AP: 1000Rnd_Gatling_30mm_Plane_CAS_01_F {
+        ammo = "ACE_Gatling_30mm_Sub_AP";
+        displayName = "30mm DU Armor Piercing";
+        displayNameShort = "30mm DUAP";
```
commy2: Should be stringtable keys.

LorenLuke:
I believe this is review/playteat ready.

LorenLuke:
Review ready. Request player feedback of how it plays. Unsure if CM 4:1 should be loaded by default to A-10/A-164, or if the HEI should remain the default loaded magazine.

Blutze:
If you are making a difference between HE and AP, you should make sure they have different penetration values. Currently you are just inheriting the vanilla ammo's `caliber = 1.4`, giving you ~20mm of RHA at ~900 m/s impact velocity. For both ammo types.

LorenLuke:
Bumped caliber values for AP rounds to penetrate between 50 and 70mm of Rha based on figure given.

Drofseh:
> Unsure if CM 4:1 should be loaded by default to A-10/A-164, or if the HEI should remain the default loaded magazine.

I would have the CM loaded by default as it is much more versatile.

LorenLuke:
Ready for final review.

addons/aircraft/CfgAmmo.hpp
```diff
@@ -39,15 +39,53 @@ class CfgAmmo {
...
+       ACE_caliber = 1.4;
+   };
+
+   class ACE_Gatling_30mm_Sub_AP: ACE_Gatling_30mm_Sub_HEI {
+        submunitionAmmo = "ACE_Gatling_30mm_AP_Plane_CAS_01_F";
+        ACE_caliber = 1.4;
+   };
+   
+    class ACE_Gatling_30mm_Sub_CM41: ACE_Gatling_30mm_Sub_HEI {
+        submunitionAmmo[] = {"ACE_Gatling_30mm_AP_Plane_CAS_01_F",0.8,"Cannon_30mm_HE_Plane_CAS_01_F",0.2};
```
PabstMirror: RPT now throws error
```
21:43:13 No class Cannon_30mm_HE_Plane_CAS_01_F
21:43:13 No ammo class Cannon_30mm_HE_Plane_CAS_01_F
21:43:13 No class Cannon_30mm_HE_Plane_CAS_01_F
21:43:13 No ammo class Cannon_30mm_HE_Plane_CAS_01_F
```

I think this should be 
`Gatling_30mm_HE_Plane_CAS_01_F`
instead of 
`Cannon_30mm_HE_Plane_CAS_01_F`

PR #7119, opened 2019-08-03
Ballistics - Add shotgun shell projectile magazine variety

Drofseh:
- Change the vanilla buckshot from what appears to be No. 3 based on pellet count to No. 00, as No. 00 is the most common for combat shotguns.
- Add multiple additional types of shotgun projectiles, including Nos. 0, 1, 2, 3, 4 buckshot and No. 7 birdshot.
- Add magazines for them for the over/under and underbarrel shotguns.

* Changed to No4 birdshot for better performance. ~~The No. 7 bird shot has many more pellets/submunitions than the others and needs to be tested for performance issues~~.
* I haven't done extra magazines for the demining drone weapon, should they be added?
* I should probably do something with the `caliber` value for the ammo as shot does not penetrate armour well.

### DIFF
--- a/addons/ballistics/CfgAmmo.hpp
+++ b/addons/ballistics/CfgAmmo.hpp
@@ -6,6 +6,89 @@ class CfgAmmo {
         timeToLive=6;
     };
 
+    class B_12Gauge_Pellets_Submunition : BulletBase { //#00 Buckshot
+        //vanilla values have been left as comments for reference purposes
+        //cost = 1;
+        //hit = 20;
+        //simulationStep = 0.0001;
+        //cartridge = "";
+        //submunitionAmmo = "B_12Gauge_Pellets_Submunition_Deploy";
+        submunitionConeType[] = {"poissondisc", 9};  //#00 Buckshot generally has 9 pellets per shell
+        //submunitionConeType[] = {"poissondisc", 18};
+        //submunitionConeAngle = 0.8;
+        //triggerSpeedCoef[] = {0.85, 1};
+        triggerTime = 0.015; // Shot takes ~5-15 feet to start spreading out and the vanilla triggerTime is too short to allow that
+        //triggerTime = 0.001;
+    };
+    class B_12Gauge_Pellets_Submunition_Deploy : BulletBase {
+        airFriction = -0.0030;
+        //airFriction = -0.0067;
+        hit = 2.55; //vanilla hit is way too high
+        //hit = 6;
+        //indirectHit = 0;
+        //indirectHitRange = 0;
+        //typicalSpeed = 360;
+        //deflecting = 35;
+    };
+
+    class ACE_12Gauge_Pellets_Submunition_No0 : B_12Gauge_Pellets_Submunition {
+        submunitionAmmo = "ACE_12Gauge_Pellets_Submunition_No0_Deploy";
+        submunitionConeType[] = {"poissondisc", 9};
+        submunitionConeAngle = 0.81;
+    };
+    class ACE_12Gauge_Pellets_Submunition_No0_Deploy : B_12Gauge_Pellets_Submunition_Deploy {
+        airFriction = -0.0033;
+        hit = 2.27;
+    };
+    class ACE_12Gauge_Pellets_Submunition_No1 : B_12Gauge_Pellets_Submunition {
+        submunitionAmmo = "ACE_12Gauge_Pellets_Submunition_No1_Deploy";
+        submunitionConeType[] = {"poissondisc", 11};
+        submunitionConeAngle = 0.83;
+    };
+    class ACE_12Gauge_Pellets_Submunition_No1_Deploy : B_12Gauge_Pellets_Submunition_Deploy {
+        airFriction = -0.0038;
+        hit = 1.86;
+    };
+    class ACE_12Gauge_Pellets_Submunition_No2 : B_12Gauge_Pellets_Submunition {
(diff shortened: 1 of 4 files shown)


addons/ballistics/CfgMagazines.hpp
```diff
@@ -3,7 +3,99 @@ class CfgMagazines {
 
     class CA_Magazine;
     class VehicleMagazine;
-    
+
```
commy2: bom?

addons/ballistics/CfgAmmo.hpp
```diff
@@ -6,6 +6,88 @@ class CfgAmmo {
         timeToLive=6;
     };
 
+    class B_12Gauge_Pellets_Submunition : BulletBase { //#00 Buckshot
```
commy2: Delete the space before the :

addons/ballistics/CfgMagazines.hpp
Drofseh: honestly not sure

addons/ballistics/CfgAmmo.hpp
Drofseh: Half of this file does ` : ` and the other half does `: ` should I change all of them?

addons/ballistics/CfgMagazines.hpp
dedmen: bom in middle of file doesn't make sense :u

Drofseh:
I've changed the No7 Birdshot into No4 Birdshot. Here is a repeat of the commit message with my reasoning.

While No7 is one of the most common loads for target shooting and hunting small birds the extreme number of pellets is just too hard on FPS.
No4 is not typically used in target shooting, but it is a common load for hunting various birds so I think it's realistic to have it for civlian purposes and the number of pellets is substatially less so the FPS hit isn't so bad.

If someone else could test the No4 Birdshot to see how it feels to them I'd appreciate it.

Drofseh:
I am happy with this in it's current state and it is not longer WIP.

addons/ballistics/CfgMagazineWells.hpp
```diff
@@ -1,5 +1,38 @@
 class CfgMagazineWells {
 
+    class CBA_12g_2rnds {
+        ADDON[] = {
+                "ACE_2Rnd_12Gauge_Pellets_No0_Buck",
```
commy2: indent wrong

Drofseh: ty, fixed

addons/ballistics/CfgAmmo.hpp
```diff
@@ -6,6 +6,114 @@ class CfgAmmo {
...
+    };
+    class ACE_12Gauge_Pellets_Submunition_No4_Bird_Deploy: B_12Gauge_Pellets_Submunition_Deploy {
+        caliber = 0.2;
+        airFriction = -0.0800;
+        hit = 0.15;
+    };
+
+    class B_12Gauge_Slug: BulletBase {
+        //caliber = 3; //too high, ~20mm of RHA (450*3*15/1000=20), ~64mm metal
+        caliber = 1.037; //~7mm RHA, ~22.4mm metal, probably still too high though as RHA is hardened.
```
commy2: Where did you get all these numbers for hit and caliber from? And why not just round this line down to 1.0. It is not like numbers this precise change anything noticably. Decreasing caliber from 3 to 1 also should reduce damage vs. uniforms, vests and helmets a lot. Are we sure we want to touch this at all?

Drofseh: For `caliber` I used an online armour penetration calculator to figure out the real world penetration of all the different projectiles and then used BI's equation `velocity[m/s] * caliber * penetrability / 1000` to convert those to config values, with `velocity` taken from the config values of the magazines and `penetrability` taken from config values in BI's material files.

For `hit` I took the real world muzzle energy of 9mm, compared it to the hit of 9mm in game and then used that to calculate the hit values of various shot types based on their respective real world muzzle energies.

commy2: That equation is missing the equal sign?

I don't think this is a good method to determine hit.

Drofseh: `Penetration depth in mm = velocity[m/s] * caliber * penetrability / 1000`
Penetrability for RHA steel is 15, for concrete 80, for meat 250.
So `450[m/s] * 1.037[caliber] * 15[penetrability] / 1000 = 6.99975 mm` penetration of rolled homogeneous armour, a type of hardened steel armour plate.


For hit, I think my method worked well enough and during in game testing of the different shot types it felt good. I'm open to suggestions for a different method though.

commy2: Vanilla values have way less decimal places:
```cpp
    class B_9x21_Ball: BulletBase {
        hit = 5;
        caliber = 1.2;
    };
```
This PR just fools accuracy to this stuff that does not exist. I am not opposed to merging this (if formatting is ok), but I won't approve such changes as I don't want to bear the onus.

Drofseh:
I've been testing this a little bit with the medical rewrite beta build and I'm actually even happier with it in the new medical system.

Getting hit with one of the larger buckshot sizes will inflict Medium to Large wounds, while getting hit with smaller buckshot sizes will inflict more Small to Medium
wounds.

Body armour and helmets will offer protection, but the arms and legs are still vulnerable.

PR #7124, opened 2019-08-07
Cargo - fix tankf

PabstMirror:


addons/cargo/CfgVehicles.hpp
```diff
@@ -63,6 +63,7 @@ class CfgVehicles {
         GVAR(space) = 4;
         GVAR(hasCargo) = 1;
     };
+    class Tank_F: Tank {};
```
jonpas: Shouldn't `Tank: LandVehicle` be `Tank_F: LancVehicle`?

PabstMirror: `Tank_F: Tank` is correct

PR #7125, opened 2019-08-07
Changes shift+adjust scope behavior

LorenLuke:
Makes shift+adjust a whole minute of angle, rather than moving up/down to the nearest whole number.

### DIFF
--- a/addons/scopes/functions/fnc_adjustScope.sqf
+++ b/addons/scopes/functions/fnc_adjustScope.sqf
@@ -36,10 +36,20 @@ private _zeroing = _adjustment select _weaponIndex;
 _zeroing params ["_elevation", "_windage", "_zero"];
 
 (GVAR(scopeAdjust) select _weaponIndex) params ["_maxVertical", "_verticalIncrement", "_maxHorizontal", "_horizontalIncrement"];
+
 if (_majorStep) then {
-    _verticalIncrement = _verticalIncrement * 10;
-    _horizontalIncrement = _horizontalIncrement * 10;
+    if(1/_verticalIncrement == floor(1/_verticalIncrement)) then {
+        _verticalIncrement = 1;
+    } else {
+        _verticalIncrement = _verticalIncrement * 5;
+    };
+    if(1/_horizontalIncrement == floor(1/_horizontalIncrement)) then {
+        _horizontalIncrement = 1;
+    } else {
+        _horizontalIncrement = _horizontalIncrement * 5;
+    };
 };
+
 TRACE_4("",_maxVertical,_verticalIncrement,_maxHorizontal,_horizontalIncrement);
 
 switch (_turretAndDirection) do {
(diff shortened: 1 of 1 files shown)


Blutze:
That is not a minute of angle in any case. 

Also, the base increments aren't always 0.1 mrad. I've seen 0.25, and I could imagine someone doing a crazy rounding mess that adjusts by 0.5 MoA (we should technically do that for the RHS compat, its what one of their scopes does IRL).

LorenLuke:
You're right, not minute of angle, but mils. That said, it's basically meant to be 10 clicks instead of 'some number between one and ten based on how many you've done before'.

commy2:
Why 10?

LorenLuke:
Because all the scopes I've used were the ACE ones that clicked about in .1 mils, so depending on what your scope was currently set to, shift+adjust could be 1 milliradian, or 0.1 milliradian. This is meant to make shift-adjust a fixed value for any given scope. It is not meant to work with simplified zeroing.

Drofseh:
I've always though that nearest whole number was a little odd as well.

Blutze:
> Because all the scopes I've used were the ACE ones that clicked about in .1 mils

Thats what I was saying: that behaviour is NOT universal.

https://github.com/toadie2k/NIArms/blob/master/Compatibility/hlc_ace3_compat_SIGAMT/config.cpp#L25 
[Actual data sheet](https://www.kern-aarau.ch/fileadmin/user_upload/Aldo/Wehrtechnik/Zielfernrohr_4x24_d.pdf)
https://github.com/toadie2k/NIArms/blob/master/hlc_wp_FN3011/config.cpp#L305

Just two examples of something else being used. As you can see on the last page of that pdf, increments (Seiten-/Höhenverstellung) are in "1/4 A‰", the per-mille referring to the same ratio you'd use for range estimates with a mildot reticle.

dedmen:
> Also, the base increments aren't always 0.1 mrad. I've seen 0.25, and I could imagine someone doing a crazy rounding mess that adjusts by 0.5 MoA

The click angle is defined in the config of the scope.

So this does 10 clicks of adjust, instead of just one mrad? We should make the number of clicks configurable.

The previous 1mrad also could've caused issues with scopes who'se click value doesn't add up to 1.
For example a scope with per click adjust of 0.3, if you do normal adjust you get 0.3,0.6,0.9,1.2

If you do a shift+adjust you suddenly get onto 1, and thus can get to 0.7,0.4,1.3 and so on, which normally shouldn't be accessible.

addons/scopes/functions/fnc_adjustScope.sqf
```diff
@@ -36,6 +36,20 @@ private _zeroing = _adjustment select _weaponIndex;
 _zeroing params ["_elevation", "_windage", "_zero"];
 
 (GVAR(scopeAdjust) select _weaponIndex) params ["_maxVertical", "_verticalIncrement", "_maxHorizontal", "_horizontalIncrement"];
+
+if (_majorStep) then {
+    if(1/_verticalIncrement == floor(1/_verticalIncrement)) then {
```
dedmen: I think this needs explanation?

Blutze: `if(1%_verticalIncrement == 0) then`
Assuming, of course, that SQF even has modulo and that ^this is the correct syntax for it.

Blutze:
I kinda like this new approach. However, the constant 5x for unusual step sizes might still be a bit high, and could also be a CBA setting.

addons/scopes/functions/fnc_adjustScope.sqf
alganthe: SQF indeed has modulo (SQF is bad but not THAT bad, cmon').

PR #7128, opened 2019-08-10
Avoid damage on fnc_paradropitem

formtapez:
- avoid damage on big vehicles when dropped from a transportation helicopter.

When big tanks like T-100 will be unloaded in-flight and hang on a parachute, the chute damages the vehicle and it get stuck in the ground. This fix attaches the parachute a bit higher on the tank and avoids damage.

### DIFF
--- a/addons/cargo/functions/fnc_paradropItem.sqf
+++ b/addons/cargo/functions/fnc_paradropItem.sqf
@@ -66,7 +66,8 @@ _itemObject setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vecto
 
     private _velocity = velocity _item;
 
-    _item attachTo [_parachute, [0,0,1]];
+    _item allowDamage false;
+    _item attachTo [_parachute, [0,0,3]];
     _parachute setVelocity _velocity;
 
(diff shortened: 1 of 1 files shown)


addons/cargo/functions/fnc_paradropItem.sqf
```diff
@@ -89,6 +90,7 @@ _itemObject setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vecto
             private _smoke = "SmokeshellYellow" createVehicle [0,0,0];
             _smoke attachTo [_item, [0,0,0]];
         };
+        _item allowDamage true;
```
commy2: What if the object was invulnerable before? Now parachuting makes it vulnerable.

WGPSenshi: An `isDamageAllowed`-if-check would be the straightforward answer. Trouble is that this only works if vehicle is local :/ , which could be messy in MP.

Dystopian: looks like you should use `statusEffect` functions like here https://github.com/acemod/ACE3/blob/9eba55be219f1b06d05f9050ca493062a460366d/addons/cargo/functions/fnc_loadItem.sqf#L44

Mike-MF:

PR #7133, opened 2019-08-11
Cargo - setSpace checks for existing cargo

### LINKED ISSUE #7131: ACE_cargo_fnc_setSpace Sets available space, not maximum
**Arma 3 Version:** `1.94.145977` (stable)
**CBA Version:** `3.12.0.190708` (stable)
**ACE3 Version:** `3.12.6.43` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- The function `ace_cargo_fnc_setSpace` appears to be intended to set the total cargo capacity of its target, but it directly sets the variable `ace_cargo_space` which actually represents the _remaining_ space in the inventory

**Steps to reproduce:**
- Place a vehicle, e.g. HEMTT transport (cargo space 20)
- Check vehicle cargo (1x spare wheel, 19 space left)
- Execute `[vehicle, 20] call ace_cargo_fnc_setSpace`
- Check vehicle cargo (1x spare wheel, 20 space left)
- Unload the spare wheel to confirm (21 space left)

- Singleplayer confirmed

**Additional information:**
- Suggested fix: change fnc_setSpace.sqf to first sum the total cargo already in the target vehicle, and set the space to `_space - _cargo`

**RPT log file:**
- N/A

pterolatypus:
- Modifies ace_cargo_fnc_setSpace to check for cargo already in the vehicle and set the total space, rather than the remaining space
Fixes #7131

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -1,4 +1,4 @@
-# ACE3 CONTRIBUTOR LIST
+﻿# ACE3 CONTRIBUTOR LIST
 # If you contributed, but are not listed here, contact me:
 # koffeinflummi@gmail.com
 #
@@ -123,6 +123,7 @@ PiZZADOX <509thParachuteInfantry@gmail.com>
 Phyma <sethramstrom@gmail.com>
 pokertour
 Professor <lukas.trneny@wo.cz>
+Pterolatypus <pterolatypus@gmail.com>
 QuickDagger
 rakowozz
 ramius86 <pasini86@hotmail.com>
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -1,4 +1,4 @@
-﻿# ACE3 CONTRIBUTOR LIST
+# ACE3 CONTRIBUTOR LIST
 # If you contributed, but are not listed here, contact me:
 # koffeinflummi@gmail.com
 #
--- a/addons/cargo/config.cpp
+++ b/addons/cargo/config.cpp
@@ -8,7 +8,7 @@ class CfgPatches {
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {"ace_interaction"};
         author = ECSTRING(common,ACETeam);
-        authors[] = {"commy2", "Glowbal","Pterolatypus"};
+        authors[] = {"commy2", "Glowbal"};
         url = ECSTRING(main,URL);
         VERSION_CONFIG;
     };
--- a/addons/cargo/config.cpp
+++ b/addons/cargo/config.cpp
@@ -8,7 +8,7 @@ class CfgPatches {
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {"ace_interaction"};
         author = ECSTRING(common,ACETeam);
-        authors[] = {"commy2", "Glowbal"};
+        authors[] = {"commy2", "Glowbal","Pterolatypus"};
         url = ECSTRING(main,URL);
         VERSION_CONFIG;
     };
--- a/addons/cargo/functions/fnc_setSpace.sqf
+++ b/addons/cargo/functions/fnc_setSpace.sqf
@@ -42,10 +42,10 @@ _vehicle setVariable [QGVAR(hasCargo), _space > 0, true];
 private _loaded = _vehicle getVariable [QGVAR(loaded), []];
 private _totalCargo = 0;
 {
-    _totalCargo += _x call FUNC(getSizeItem)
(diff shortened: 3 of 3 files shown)


addons/cargo/functions/fnc_setSpace.sqf
```diff
@@ -35,10 +35,18 @@ if (
     {_space == _vehicle getVariable [QGVAR(space), CARGO_SPACE(typeOf _vehicle)]}
 ) exitWith {};
 
+private ["_loaded", "_cargo"];
```
jonpas: Use private keyword.

addons/cargo/config.cpp
```diff
@@ -8,7 +8,7 @@ class CfgPatches {
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {"ace_interaction"};
         author = ECSTRING(common,ACETeam);
-        authors[] = {"commy2", "Glowbal"};
+        authors[] = {"commy2", "Glowbal","Pterolatypus"};
```
jonpas: Fixing a bug or improving something doesn't make you an author.

addons/cargo/functions/fnc_setSpace.sqf
```diff
@@ -35,10 +35,18 @@ if (
...
 // Apply new space globally
 // Necessary to update value, even if no space, as API could be used again
 _vehicle setVariable [QGVAR(hasCargo), _space > 0, true];
-_vehicle setVariable [QGVAR(space), _space, true];
+//Account for cargo already in the vehicle
+_loaded = _vehicle getVariable [QGVAR(loaded), []];
+_cargo = 0;
+{
+  _cargo += _x call FUNC(getSizeItem)
```
jonpas: This wasn't tested at all, was it?

Also 4 spaces indentation, read coding guidelines.

addons/cargo/functions/fnc_setSpace.sqf
```diff
@@ -35,10 +35,18 @@ if (
     {_space == _vehicle getVariable [QGVAR(space), CARGO_SPACE(typeOf _vehicle)]}
 ) exitWith {};
 
+private ["_loaded", "_cargo"];
 // Apply new space globally
 // Necessary to update value, even if no space, as API could be used again
 _vehicle setVariable [QGVAR(hasCargo), _space > 0, true];
-_vehicle setVariable [QGVAR(space), _space, true];
+//Account for cargo already in the vehicle
```
jonpas: Space after comment delimiter.

addons/cargo/functions/fnc_setSpace.sqf
pterolatypus: I tested it by repeating the steps in #7131 and confirming the issue is resolved in that case.
Am I missing something?

jonpas: How? SQF doesn't have `+=`.

pterolatypus: ...oops
That will be because my test case is bad (I'm setting the value to what it _should_ already be, but if it errors then nothing happens so the problem doesn't arise).
I'll fix it and test more thoroughly. Not sure why the error didn't pop up...

pterolatypus: Should be fixed now

AUTHORS.txt
```diff
@@ -1,4 +1,4 @@
-# ACE3 CONTRIBUTOR LIST
+﻿# ACE3 CONTRIBUTOR LIST
```
dedmen: please don't change bom mark.

addons/cargo/functions/fnc_setSpace.sqf
```diff
@@ -31,14 +31,22 @@ TRACE_2("setSpace",_vehicle,_size);
 // Nothing to do here
 if (
     (isNil "_space") ||
-    {isNull _vehicle} ||
-    {_space == _vehicle getVariable [QGVAR(space), CARGO_SPACE(typeOf _vehicle)]}
+    {isNull _vehicle}
 ) exitWith {};
 
 // Apply new space globally
```
Dystopian: maybe this comment should belong to `setVariable` lines below

AUTHORS.txt
pterolatypus: Think Atom did this automatically.
Fixed

addons/cargo/functions/fnc_setSpace.sqf
pterolatypus: Rearranged the comments and added a couple of newlines for readability

PR #7134, opened 2019-08-11
Add self stitching setting

alganthe:
- Add self stitching setting for medical rewrite.

### DIFF
--- a/addons/medical_treatment/functions/fnc_canStitch.sqf
+++ b/addons/medical_treatment/functions/fnc_canStitch.sqf
@@ -18,7 +18,4 @@
 
 params ["", "_patient"];
 
-!(GET_BANDAGED_WOUNDS(_patient) isEqualTo []) && {
-    !(_patient isEqualTo ACE_player) ||
-    (GVAR(allowSelfStitch) == 1 && (_patient isEqualTo ACE_player))
-}
+!(GET_BANDAGED_WOUNDS(_patient) isEqualTo [])
--- a/addons/medical_treatment/functions/fnc_canStitch.sqf
+++ b/addons/medical_treatment/functions/fnc_canStitch.sqf
@@ -18,4 +18,7 @@
 
 params ["", "_patient"];
 
-!(GET_BANDAGED_WOUNDS(_patient) isEqualTo [])
+!(GET_BANDAGED_WOUNDS(_patient) isEqualTo []) && {
+    !(_patient isEqualTo ACE_player) ||
+    (GVAR(allowSelfStitch) == 1 && (_patient isEqualTo ACE_player))
+}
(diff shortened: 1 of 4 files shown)


addons/medical_treatment/functions/fnc_canStitch.sqf
```diff
@@ -18,4 +18,7 @@
 
 params ["", "_patient"];
 
-!(GET_BANDAGED_WOUNDS(_patient) isEqualTo [])
+!(GET_BANDAGED_WOUNDS(_patient) isEqualTo []) && {
```
mharis001: Don't need this, the `allowSelfTreatment` config entry handles this. [reference](https://github.com/acemod/ACE3/blob/master/addons/medical_treatment/functions/fnc_canTreat.sqf#L27)

addons/medical_treatment/stringtable.xml
```diff
@@ -3496,5 +3496,13 @@
             <Chinesesimp>个人急救包使用条件</Chinesesimp>
             <Chinese>個人急救包使用條件</Chinese>
         </Key>
+        <Key ID="STR_ACE_Medical_Treatment_AllowSelfStitch_DisplayName">
```
mharis001: Move strings to top with the other (new) settings, after L217. I guess these are left over entries that need to cleaned up eventually.

addons/medical_treatment/stringtable.xml
```diff
@@ -3496,5 +3496,13 @@
...
+            <English>Self Stitching</English>
+            <French>Suturer soi-même</French>
+        </Key>
+        <Key ID="STR_ACE_Medical_Treatment_AllowSelfStitch_Description">
+            <English>Can the surgical kit be used on yourself?</English>
```
SilentSpike: Please no more questions as descriptions 😛

mharis001: Maybe: 

1. "Enables individuals to stitch themselves using a surgical kit."
2. "Enables the use of surgical kits to stitch oneself." (match the self IV setting)
3. "Controls whether individuals can stitch themselves."

Not sure which is best. With the medical settings cleanup, tried to avoid using questions as descriptions.

alganthe: I went with the second option, it's consistent with what we have.

addons/medical_treatment/stringtable.xml
alganthe: Done

addons/medical_treatment/functions/fnc_canStitch.sqf
alganthe: Didn't know canTreat worked like that, niiice.

PR #7135, opened 2019-08-12
Fix slide order in slideshow

bux:
- Adjust the order of the interaction menu entries in a way that they are in the same order as the user entered them.

### DIFF
--- a/addons/slideshow/functions/fnc_addSlideActions.sqf
+++ b/addons/slideshow/functions/fnc_addSlideActions.sqf
@@ -22,12 +22,15 @@
 params ["_objects", "_images", "_names", "_controller", "_currentSlideshow"];
 
 private _actions = [];
+
+// fix action order by going through the array backwards
+for "_i" from (count _images - 1) to 0 step -1 do
 {
     _actions pushBack
     [
         [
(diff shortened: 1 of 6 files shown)


addons/slideshow/functions/fnc_addSlideActions.sqf
```diff
@@ -22,12 +22,15 @@
 params ["_objects", "_images", "_names", "_controller", "_currentSlideshow"];
 
 private _actions = [];
+
+// fix action order by going through the array backwards
+for "_i" from (count _images - 1) to 0 step -1 do
```
jonpas: This breaks adding actions, no `_x` anymore. 

Use `reverse ARRAY`.

addons/slideshow/functions/fnc_createSlideshow.sqf
```diff
@@ -10,12 +10,13 @@
  * 3: Action Names <ARRAY>
  * 4: Slide Duration <NUMBER> (0 disables automatic transitions)
  * 5: Set Name <STRING> (default: localized "Slides")
+ * 6: Cover Image Path <STRING>
```
jonpas: Missing default.

PR #7137, opened 2019-08-12
Rearm - Add distance setting

alganthe:
- Add a setting to change the distance a vehicle can be rearmed at.

Someone asked for it on slack and I figured that this could indeed be a setting, the default value is set to the current value (20m) minimum at 10 and maximum at 50.

### DIFF
--- a/addons/rearm/stringtable.xml
+++ b/addons/rearm/stringtable.xml
@@ -612,5 +612,13 @@
             <Italian>Il rifornimento delle munizioni (-1 per disabilitarlo)</Italian>
             <Russian>Объем боеприпасов для перевооружения (-1 для отмены)</Russian>
         </Key>
+        <Key ID="STR_ACE_Rearm_RearmSettings_distance_DisplayName">
+            <English>Rearm distance</English>
+            <French>Distance de réarmement</French>
+        </Key>
+        <Key ID="STR_ACE_Rearm_RearmSettings_distance_Description">
+            <English>The maximum distance a vehicle can be rearmed at</English>
+            <French>Distance maximale à laquelle un véhicle peut être réarmé</French>
+        </Key>
     </Package>
 </Project>
--- a/addons/rearm/stringtable.xml
+++ b/addons/rearm/stringtable.xml
@@ -615,6 +615,7 @@
         <Key ID="STR_ACE_Rearm_RearmSettings_distance_DisplayName">
             <English>Rearm distance</English>
             <French>Distance de réarmement</French>
+            <German>Aufrüstbereich</German>
         </Key>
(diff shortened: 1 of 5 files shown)


addons/rearm/stringtable.xml
```diff
@@ -612,5 +612,13 @@
             <Italian>Il rifornimento delle munizioni (-1 per disabilitarlo)</Italian>
             <Russian>Объем боеприпасов для перевооружения (-1 для отмены)</Russian>
         </Key>
+        <Key ID="STR_ACE_Rearm_RearmSettings_distance_DisplayName">
+            <English>Rearm distance</English>
+            <French>Distance de réarmement</French>
```
dedmen:
```suggestion
            <French>Distance de réarmement</French>
            <German>Aufmunitionierungs Reichweite</German>
```
Might be too long?
Might also do `Aufmunitionierungsreichweite` but pls no.

addons/rearm/stringtable.xml
```diff
@@ -612,5 +612,13 @@
...
+        </Key>
+        <Key ID="STR_ACE_Rearm_RearmSettings_distance_Description">
+            <English>The maximum distance a vehicle can be rearmed at</English>
+            <French>Distance maximale à laquelle un véhicle peut être réarmé</French>
```
dedmen:
```suggestion
            <French>Distance maximale à laquelle un véhicle peut être réarmé</French>
            <German>Die maximale Distanz, über die ein Fahrzeug Aufmunitioniert werden kann</German>
```

addons/rearm/stringtable.xml
alganthe: definitely too long

PabstMirror:
For `GVAR(level) = 0`: we need to modify distance check in ace_rearm_fnc_rearmEntireVehicle or progress bar will fail immediately
The other distance checks should be fine as level 1/2 you carry the magazine to the vehicle.

addons/rearm/stringtable.xml
dedmen: Maybe @commy2 has an idea

shukari: Aufmunitionierungsreichweite (28)
Nachschubsreichweite (20)
Nachschubsbereich (17)
Aufrüstungsbereich (17)

some ideas...

alganthe: first one could clip, the other three would work, just give me the best one and i'll add it.

dedmen: Nachschubbereich would be without s, as schubs is pushing someone, would mean after push area.
Personally I'd just keep Rearm probably :D 

```suggestion
            <French>Distance de réarmement</French>
            <German>Aufrüstbereich</German>
```

PR #7143, opened 2019-08-15
Update wiki description of interaction framework

Schwaggot:
- update the wiki description of the interaction menu framework
- adds a hint that functions will be not executed on server
- removes 'it is generally preferred to add actions via config' as suggested by jonpas on Slack

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -132,6 +132,7 @@ Riccardo Petricca <petriccarcc@gmail.com>
 Robert Boklahánics <bokirobi@gmail.com>
 ruPaladin <happyworm24@rambler.ru>
 Rutger "RedBery" Meijering <c.redbery@gmail.com>
+Schwaggot <tom.ryan@posteo.de>
 simon84 <badguy360th@gmail.com>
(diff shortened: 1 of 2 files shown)


Dystopian:
For Zeus there is `ace_interact_menu_fnc_addActionToZeus`. Also functions **will** be executed on server if it's not dedicated (server with interface).

jonpas:
I'd reword it to say they are all local and entire menu is client-side. Because if you add it on server, it'll be on server, you just can't use it.

Schwaggot:
I've updated it. My point was that the function actually exit if ```!hasInterface```, e.g., https://github.com/acemod/ACE3/blob/master/addons/interact_menu/functions/fnc_createAction.sqf#L31

jonpas:
Ooooh, in that case I prefer the first version. :grimacing:

TheMagnetar:
@Schwaggot can you modify it again so we can merge it?

Schwaggot:
@TheMagnetar sure, I will do so either later today or tomorrow.

PR #7147, opened 2019-08-24
Cookoff - Allow enabling only on player crewed vehicles

Cyruz143:
- Add an extra cookoff setting to allow cookoff to only occur on player crewed vehicles

I know cookoff is now disabled by default due to it's weird issues where vehicles could take excessive amounts of damage. For people that mainly do co-ops it would be nice to still have it work for player controlled vehicles.

Tested locally and seems to work as expected, open to input especially on names (I'm terrible at them)!

### DIFF
--- a/addons/cookoff/CfgEden.hpp
+++ b/addons/cookoff/CfgEden.hpp
@@ -12,7 +12,7 @@ class Cfg3DEN {
                         expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
                         typeName = "BOOL";
                         condition = "objectVehicle";
-                        defaultValue = QUOTE((GETMVAR(QGVAR(enable),0)) in [1,2]);
+                        defaultValue = QUOTE((GETMVAR(QGVAR(enable),0)) in [ARR_2(1,2)]);
                     };
                     class GVAR(enableAmmoCookoff) {
                         property = QGVAR(enableAmmoCookoff);
--- a/addons/cookoff/CfgEden.hpp
+++ b/addons/cookoff/CfgEden.hpp
@@ -14,16 +14,6 @@ class Cfg3DEN {
                         condition = "objectVehicle";
                         defaultValue = QUOTE(GETMVAR(QGVAR(enable),true));
                     };
-                    class GVAR(disableForAI) {
-                        property = QGVAR(disableForAI);
-                        control = "Checkbox";
-                        displayName = CSTRING(disableForAI_name);
-                        tooltip = CSTRING(disableForAI_tooltip);
-                        expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
-                        typeName = "BOOL";
-                        condition = "objectVehicle";
-                        defaultValue = QUOTE(GETMVAR(QGVAR(disableForAI),false));
-                    };
                     class GVAR(enableAmmoCookoff) {
                         property = QGVAR(enableAmmoCookoff);
                         control = "Checkbox";
--- a/addons/cookoff/CfgEden.hpp
+++ b/addons/cookoff/CfgEden.hpp
@@ -12,7 +12,7 @@ class Cfg3DEN {
                         expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
                         typeName = "BOOL";
                         condition = "objectVehicle";
-                        defaultValue = QUOTE(GETMVAR(QGVAR(enable),true));
+                        defaultValue = QUOTE((GETMVAR(QGVAR(enable),0)) in [1,2]);
                     };
                     class GVAR(enableAmmoCookoff) {
                         property = QGVAR(enableAmmoCookoff);
--- a/addons/cookoff/CfgEden.hpp
+++ b/addons/cookoff/CfgEden.hpp
@@ -14,6 +14,16 @@ class Cfg3DEN {
                         condition = "objectVehicle";
                         defaultValue = QUOTE(GETMVAR(QGVAR(enable),true));
                     };
+                    class GVAR(disableForAI) {
+                        property = QGVAR(disableForAI);
+                        control = "Checkbox";
+                        displayName = CSTRING(disableForAI_name);
+                        tooltip = CSTRING(disableForAI_tooltip);
+                        expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
+                        typeName = "BOOL";
+                        condition = "objectVehicle";
+                        defaultValue = QUOTE(GETMVAR(QGVAR(disableForAI),false));
+                    };
                     class GVAR(enableAmmoCookoff) {
                         property = QGVAR(enableAmmoCookoff);
                         control = "Checkbox";
--- a/addons/cookoff/XEH_postInit.sqf
+++ b/addons/cookoff/XEH_postInit.sqf
@@ -26,6 +26,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
         GVAR(cacheTankDuplicates) setVariable [_typeOf, _duplicateHitpoints];
     };
 
+    _vehicle setVariable [QGVAR(disableForAI), GVAR(disableForAI), true];
     _vehicle addEventHandler ["HandleDamage", {
         if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)]) then {
             ["tank", _this] call FUNC(handleDamage);
@@ -52,6 +53,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
         GVAR(cacheTankDuplicates) setVariable [_typeOf, _duplicateHitpoints];
     };
 
+    _vehicle setVariable [QGVAR(disableForAI), GVAR(disableForAI), true];
     _vehicle addEventHandler ["HandleDamage", {
         if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)]) then {
             ["tank", _this] call FUNC(handleDamage);
@@ -62,6 +64,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 ["Car", "init", {
     params ["_vehicle"];
 
+    _vehicle setVariable [QGVAR(disableForAI), GVAR(disableForAI), true];
     _vehicle addEventHandler ["HandleDamage", {
         if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)]) then {
             ["car", _this] call FUNC(handleDamage);
--- a/addons/cookoff/XEH_postInit.sqf
+++ b/addons/cookoff/XEH_postInit.sqf
@@ -83,7 +83,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 
 // blow off turret effect
 ["Tank", "killed", {
-    if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)] !=0) then {
+    if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)] in [1, 2, true]) then {
         if (random 1 < 0.15) then {
             (_this select 0) call FUNC(blowOffTurret);
         };
--- a/addons/cookoff/XEH_postInit.sqf
+++ b/addons/cookoff/XEH_postInit.sqf
@@ -27,7 +27,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
     };
 
     _vehicle addEventHandler ["HandleDamage", {
-        if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)] !=0 ) then {
+        if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)] in [1, 2, true]) then {
             ["tank", _this] call FUNC(handleDamage);
         };
     }];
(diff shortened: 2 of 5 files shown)


addons/cookoff/XEH_postInit.sqf
```diff
@@ -62,6 +64,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 ["Car", "init", {
     params ["_vehicle"];
 
+    _vehicle setVariable [QGVAR(disableForAI), GVAR(disableForAI), true];
```
commy2: Init event is globally executed, so why do you attempt to use setVariable public, which has global effects?
Aside from that, setVariable public is bugged in init events, it does not synchronize.

addons/cookoff/stringtable.xml
```diff
@@ -25,6 +25,12 @@
             <Japanese>誘爆の損傷処理と砲塔の爆発効果を変更します。</Japanese>
             <Russian>Изменяет обработку урона для возгорания и эффекта срыва башни</Russian>
         </Key>
+        <Key ID="STR_ACE_CookOff_disableForAI_name">
+            <English>Disable cook off for AI vehicles</English>
+        </Key>
+        <Key ID="STR_ACE_CookOff_disableForAI_tooltip">
+            <English>Cook off will only run on player crewed vehicles</English>
```
commy2: But that's only the case if the checkbox is actually set. Also missing a period at the end of the sentence.

addons/cookoff/stringtable.xml
```diff
@@ -25,6 +25,12 @@
             <Japanese>誘爆の損傷処理と砲塔の爆発効果を変更します。</Japanese>
             <Russian>Изменяет обработку урона для возгорания и эффекта срыва башни</Russian>
         </Key>
+        <Key ID="STR_ACE_CookOff_disableForAI_name">
+            <English>Disable cook off for AI vehicles</English>
```
commy2: Too verbose for the settings display name imo.

addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -21,6 +21,10 @@ _thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIn
 // it's already dead, who cares?
 if (damage _vehicle >= 1) exitWith {};
 
+// Check for players and exit if none found and the disable for AI setting is true
+private _playerCrew = (fullCrew [_vehicle, "", false]) findIf {isPlayer (_x #0)};
+if (_vehicle getVariable [QGVAR(disableForAI), GVAR(disableForAI)] && _playerCrew isEqualTo -1) exitWith {};
```
PabstMirror: ` && {}` (short circuit) would be good here, moving the `fullCrew` line into it

addons/cookoff/initSettings.sqf
```diff
@@ -10,6 +10,16 @@
...
 
+[
+    QGVAR(disableForAI), "CHECKBOX",
+    [LSTRING(disableForAI_name), LSTRING(disableForAI_tooltip)],
+    LSTRING(category_displayName),
+    false, // default value
+    true, // isGlobal
+    {[QGVAR(disableForAI), _this] call EFUNC(common,cbaSettings_settingChanged)},
+    true // Needs mission restart
+] call CBA_settings_fnc_init;
```
commy2: Okay, so why is this a "disable" type setting, when the others are "enable"?
And it is by default off, so the default is a double negative "don't deactive". Would it not make more sense to have it say "activate" and just invert the default value?

Also, there already is a enable setting. I'd rather change the enable setting from checkbox to list with the following options:
- enable
- disable
- enable for player controlled vehicles

because those are the only 3 logical distinct options (`disable, enableAI` and `disable, disableAI` have the same effect).

List type settings should be bw-compatible with setting values like false and true with
flase -> index 0
true -> index 1

if not I'd consider it a bug in cba_settings.

addons/cookoff/functions/fnc_handleDamage.sqf
commy2: `_playerCrew` is such an arbitrary variable name. It is not clear what the content of the variable is. If taken literally it is misleading, because the name would imply something like an array of all crew that happen to be players.

```sqf
private _hasPlayerInCrew = fullCrew [_vehicle, "", false] findIf {isPlayer (_x select 0)} != -1;
if (_vehicle getVariable [QGVAR(disableForAI), GVAR(disableForAI)] && !_hasPlayerInCrew) exitWith {};
```

addons/cookoff/XEH_postInit.sqf
commy2: Also by always setting this variable on every vehicle, using the global variable later as fallback setting (getVariable default value) becomes irrelevant. I would just delete this line completely as it also arbitrarily prevents the setting to take effect when changed in mission (but also only for vehicles that were created before the setting change).

I'd still keep it an object namespace variable that is by default undefined and falls back to the global setting, to allow mission makers (or anyone using class event handlers!) to enable or disable this for specific vehicles (vehicle types).

addons/cookoff/initSettings.sqf
jonpas: Also CBA_fnc_addSetting.

Cyruz143:
Rewritten as per comments, requires a CBA fix from commy for backwards compatability.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -27,7 +27,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
     };
 
     _vehicle addEventHandler ["HandleDamage", {
-        if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)]) then {
+        if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)] !=0 ) then {
```
mharis001: This will error if someone was using `_vehicle setVariable ["ace_cookoff_enable", true, true]`.

addons/cookoff/initSettings.sqf
```diff
@@ -1,10 +1,10 @@
 // CBA Settings [ADDON: ace_cookoff]:
 
 [
-    QGVAR(enable), "CHECKBOX",
+    QGVAR(enable), "LIST",
     [LSTRING(enable_hd_name), LSTRING(enable_hd_tooltip)],
     LSTRING(category_displayName),
-    false, // default value
+    [[0, 1, 2], ["STR_A3_OPTIONS_DISABLED", "STR_A3_OPTIONS_ENABLED", "STR_ACE_CookOff_enableForPlayers"], 0],
```
mharis001: Use `LSTRING`.

addons/cookoff/XEH_postInit.sqf
Cyruz143: Any thoughts on how to mitigate this? I don't know what the general thought process is on people directly setting variables rather than using CBA settings or Eden check boxes

mharis001: Could use in: `(_this select 0) getVariable [QGVAR(enable), GVAR(enable)] in [1, 2, true]`.

addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -21,6 +21,10 @@ _thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIn
 // it's already dead, who cares?
 if (damage _vehicle >= 1) exitWith {};
 
+// Check for players and exit if none found and the disable for AI setting is true
+private _hasPlayerInCrew = fullCrew [_vehicle, "", false] findIf {isPlayer (_x select 0)} != -1;
+if (_vehicle getVariable [QGVAR(enable), GVAR(enable)] isEqualTo 2 && !_hasPlayerInCrew) exitWith {};
```
PabstMirror: `if (_vehicle getVariable [QGVAR(enable), GVAR(enable)] isEqualTo 2 && {fullCrew [_vehicle, "", false] findIf {isPlayer (_x select 0)} == -1}) exitWith {};`

commy2:
Needs update in cba_settings to read a List type setting in config set to false as 0 and true as 1.

addons/cookoff/functions/fnc_handleDamage.sqf
commy2: That's very hard to read. What was wrong with the lines before?

PabstMirror: `findIf` is always checked, hd can run 30+ times for a single impact on vehicles with lots of ERA hitpoints

addons/cookoff/functions/fnc_handleDamage.sqf
```diff
@@ -21,6 +21,9 @@ _thisHandleDamage params ["_vehicle", "", "_damage", "_source", "_ammo", "_hitIn
 // it's already dead, who cares?
 if (damage _vehicle >= 1) exitWith {};
 
+// Check for players and exit if none found and the enable for players only setting is true
+if (_vehicle getVariable [QGVAR(enable), GVAR(enable)] isEqualTo 2 && {fullCrew [_vehicle, "", false] findIf {isPlayer (_x select 0)} == -1}) exitWith {};
```
commy2: I think it is weird that this variable is checked in the event handler added in postInit, but then also again in the called function. This can probably be simplified. To avoid copying the line over and over, I would move all checks in the function.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -89,7 +89,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 
 // blow off turret effect
 ["Tank", "killed", {
-    if ((_this select 0) getVariable [QGVAR(enable),GVAR(enable)]) then {
+    if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)] in [1, 2, true]) then {
```
commy2: This is a killed effect. 2 (player crewed only) is treated exactly the same as 1/true (enabled), even though no player may be inside the vehicle. Weird.

Cyruz143: That's my mistake, will fix.

addons/cookoff/functions/fnc_handleDamage.sqf
Cyruz143: It's checked in the EH because at this point it's deciding if it should be added at all and was updated to the `in` check to support people who've set the variable manually as per @mharis001's suggestion.

Surely we want to keep this so the EH isn't added if people have the setting disabled? The function call is checking if it's just enabled or enabled for players only so is a slightly different check.

If it's all moved in to the function, it'll add the EH even when the setting is disabled. I'm likely misunderstanding, grab me on Slack.

Cyruz143: Moved check to be in fnc. Only checking if disabled because if it's 1/true it'll run and if it's 2 it has it's logic check already built in.

addons/cookoff/XEH_postInit.sqf
Cyruz143: Switched logic to only check if disabled as doesn't matter if it's 1 or 2.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -89,7 +83,7 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 
 // blow off turret effect
 ["Tank", "killed", {
-    if ((_this select 0) getVariable [QGVAR(enable),GVAR(enable)]) then {
+    if ((_this select 0) getVariable [QGVAR(enable), GVAR(enable)] !=0) then {
```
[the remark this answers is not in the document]
Cyruz143: It was that in a previous version, Commy advise me to change it as per https://github.com/acemod/ACE3/pull/7147#discussion_r317375977

Cyruz143:
Been running this for a month in dedi MP sessions, so far now issues noted.

addons/cookoff/XEH_postInit.sqf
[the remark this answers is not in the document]
PabstMirror: if it's set to false it will throw:
```
Error in expression <ce_cookoff_enable", ace_cookoff_enable] !=0) then {
if (random 1 < 0.15) then {
>
   Error position: <!=0) then {
if (random 1 < 0.15) then {
>
   Error !=: Type Bool, expected Number,String,Not a Number,Object,Side,Group,Text,Config entry,Display (dialog),Control,Network Object,Team member,Task,Location
 File z\ace\addons\cookoff\XEH_postInit.sqf..., line 86
```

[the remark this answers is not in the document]
Cyruz143: Can I get away with `if (!(_this select 0) getVariable [QGVAR(enable), GVAR(enable)] isEqualTo 0)` @PabstMirror

Cyruz143:
> 
> 
> also need to updated the 3den attribute which is expecting a bool for
> `defaultValue = QUOTE(GETMVAR(QGVAR(enable),true));`
> so it always adds the attribute when just opening a tanks 3den properties

I'm a bit lost on this one Pabst, I assume it should now have a drop down with 0, 1, 2 values? It should default to 1?

addons/cookoff/CfgEden.hpp
```diff
@@ -12,7 +12,7 @@ class Cfg3DEN {
                         expression = QUOTE(if !(_value) then {_this setVariable [ARR_3('%s',_value,true)];};);
                         typeName = "BOOL";
                         condition = "objectVehicle";
-                        defaultValue = QUOTE(GETMVAR(QGVAR(enable),true));
+                        defaultValue = QUOTE((GETMVAR(QGVAR(enable),0)) in [1,2]);
```
PabstMirror:
```suggestion
                        defaultValue = QUOTE((GETMVAR(QGVAR(enable),0)) in [ARR_2(1,2)]);
```

PR #7154, opened 2019-08-29
Improve medical statemachine settings

mharis001:
- Improve display names and descriptions of state machine settings
- Replace "AI Instant Death" setting with "AI Fatal Injuries" that provides the same level of control over AI fatal injuries as the player equivalent
    - Before, this setting also controlled whether AI were able to enter cardiac arrest from something other than fatal injuries (which is not what one would assume a setting named "AI Instant Death" controls)
- Change cardiac arrest behaviour for AI - cardiac arrest requires unconsciousness
    - Before, AI were able go into cardiac arrest even if they were unable to go unconscious (this would result in them immediately dying if they reanimated from cardiac arrest)
    - Now, if an AI unit goes into cardiac arrest with AI unconsciousness disabled, they will die immediately

The descriptions may need to be tweaked further but I think they now better express exactly what the settings control and how they interact with each other.


addons/medical_statemachine/functions/fnc_conditionExecutionDeath.sqf
```diff
@@ -17,4 +17,4 @@
 
 params ["_unit"];
 
-(GVAR(fatalInjuryCondition) < 2) && {!(_unit getVariable [QEGVAR(medical,deathBlocked), false])}
+(isPlayer _unit && {GVAR(fatalInjuriesPlayer) != FATAL_INJURIES_NEVER} || {!isPlayer _unit && {GVAR(fatalInjuriesAI) != FATAL_INJURIES_NEVER}}) && {!(_unit getVariable [QEGVAR(medical,deathBlocked), false])}
```
SilentSpike: I wouldn't mind seeing this conditional broken across multiple lines. That's just a minor readability thing though.

addons/medical_statemachine/stringtable.xml
```diff
@@ -8,33 +8,17 @@
...
+            <English>Player Fatal Injuries</English>
         </Key>
-        <Key ID="STR_ACE_Medical_Statemachine_FatalInjuryCondition_Description">
-            <English>Controls when a player can receive a fatal injury.</English>
-            <German>Definiert wann ein Spieler tödliche Verletzungen bekommen kann</German>
-            <Japanese>プレイヤーが致命傷から蘇生できるかどうか決定します</Japanese>
-            <Russian>Определяет, когда игрок может получить смертельное ранение</Russian>
-            <French>Définit quand un joueur peut recevoir une blessure fatale.</French>
+        <Key ID="STR_ACE_Medical_Statemachine_FatalInjuriesPlayer_Description">
+            <English>Controls when players can receive fatal injuries. A fatal injury is caused by receiving an injury to the head or heart.</English>
```
SilentSpike: This may need to be reworked particularly if we do the amassed wounds thing. I guess just 
> A fatal injury is caused by significant damage to the head or body

encompasses the cases of both organ damage (headshot / heartshot) or amassed wounding.

SilentSpike:
👍 I like this direction. I'm not sure if the description string for the AI unconsciousness setting is perhaps too long for the UI though?

mharis001:
The AI Unconsciousness string is a bit long but the newlines help break it up and allow it to fit in the UI.

PR #7155, opened 2019-08-30
Realistic names - Contact DLC rifles added

Laid3acK:
- Add the translation of the Contact rifles (except shotguns).

References from the [MSBS Grot Wiki](https://en.wikipedia.org/wiki/MSBS_rifle) and the [manufacturer](http://fabrykabroni.pl/) for the Promet.
References from various sources and Wikis for the AK15 variants (AK12 chambered in 7.62x39 mm).

### DIFF
--- a/addons/realisticnames/stringtable.xml
+++ b/addons/realisticnames/stringtable.xml
@@ -3950,5 +3950,277 @@
             <Chinesesimp>口径: 5.7mm&lt;br /&gt;发数: 50&lt;br /&gt;使用于: P90</Chinesesimp>
             <Korean>구경: 5.7mm&lt;br /&gt;장탄수: 50&lt;br /&gt;사용됨: P90</Korean>
         </Key>
+        <Key ID="STR_ACE_RealisticNames_arifle_AKM">
+            <Chinese>AKM</Chinese>
+            <French>AKM</French>
+            <Spanish>AKM</Spanish>
+            <Italian>AKM</Italian>
+            <Polish>AKM</Polish>
+            <Russian>AKM</Russian>
+            <German>AKM</German>
+            <Czech>AKM</Czech>
+            <English>AKM</English>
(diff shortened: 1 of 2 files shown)


Drofseh:
English first for string tables?

addons/realisticnames/stringtable.xml
```diff
@@ -3950,5 +3950,549 @@
...
+        <Key ID="STR_ACE_RealisticNames_arifle_AKM">
+            <Chinese>AKM</Chinese>
+            <French>AKM</French>
+            <Spanish>AKM</Spanish>
+            <Italian>AKM</Italian>
+            <Polish>AKM</Polish>
+            <Russian>AKM</Russian>
+            <German>AKM</German>
+            <Czech>AKM</Czech>
+            <English>AKM</English>
```
commy2: Move English to first position in stringtable. We forgo the second English entry Original, because it is a duplicate, however, this means the first language is the fall back if the game language is missing.
Therefore English should be at first position.

Blutze:
Might wanna include info about the cartridge in the names for the MSBSs, as this is not the standard 5.56 variant. The older naming system described on the wiki seem to allow that already, otherwise just drop a `(6.5x39)` on them.

`MSBS-6.5B GL`
`Grot B20PSBI-M2 (6.5x39)`

Something along those lines.

Laid3acK:
Maybe should be like that but, for example, 
the Promet Marksman 3D model is a bullpup version with a 22" barrel which doesn't exist.
The real one is chambered in 7.62x56 and looks like that : [MSBS DMR rifle](https://fragoutmag.com/fabryka-broni-lucznik-radom-msbs-762n-first-pictures/)

So, an exact designation with a wrong caliber has no sens, as many rifles in this game.
The reason i chosen a compromise with the BI and the manufacturer names.

- All English entries first fixed.
- Contact pistol and launcher added.
- Minors updates.

PR #7157, opened 2019-08-30
Add goggle overlay support for Contact update Blindfolds

veteran29:
- Title

Texture by @BaerMitUmlaut 
<details>
  <summary>Images</summary>
  
image

</details>


commy2:
Should name the file _ca before converting with img2paa.

veteran29:
I have done that. Then renamed to `Blindfold.paa`. Conversion was done with TexView 2

jonpas:
> Then renamed to `Blindfold.paa`.

Why though?

veteran29:
> Why though?
Because all other textures in the component are named like that.

commy2:
They shouldn't.

veteran29:
Should I fix all of them then?

commy2:
Make sure to convert them when named _ca, so
- paa → png
- rename _ca.png
- png → paa

jones140:
Overlay for new respirator gear as well?

jonpas:
> just a reminder that goggles effects are now client settable, so you would need to enforce the setting if you need the effect for a mission

This should have an override then.

veteran29:
Merge? Goggles can be forced via setting. Option to force googles per goggle class should be separate PR anyway imho.

commy2:
Agreed. Goggles component needs a rewrite anyway. Once I find my life again, I will do it.

PR #7160, opened 2019-09-01
Fix magazines not listed in arsenal due to case-sensitive check and replace  code with cba functions

brainslush:
- Fix: Magazines on the virtual item list of the ace arsenal which are written in different cases than in the config wouldn't be listed in the arsenal.
- Replaced code with cba functions to streamline code.

### DIFF
--- a/addons/arsenal/functions/fnc_fillRightPanel.sqf
+++ b/addons/arsenal/functions/fnc_fillRightPanel.sqf
@@ -74,24 +74,15 @@ private _compatibleItems = [];
 private _compatibleMagazines = [[[], []], [[], []], [[], []]];
 {
     if (_x != "") then {
-        private _weaponConfig = (configFile >> "CfgWeapons" >> _x);
         private _index = _forEachIndex;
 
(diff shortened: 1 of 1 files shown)


brainslush:
Don't merge. There is more to sort out.

dedmen:
The bug was because we were using configCase instead of everything toLower.
Didn't we at some point move arsenal to using configCase, to not have to toLower everything on every iteration or smth?

alganthe:
That was indeed the case dedmen.

brainslush:
We could do it the other way around and get the configCase names for the items added via addvirtualItems by doing a config lookup. Since addvirtualItems shouldn't be called that often I think the performance impact can be ignored.

dedmen:
> and get the configCase names for the items added via addvirtualItems by doing a config lookup
I thought we were already doing that 🤔

brainslush:
I'll create soon a new PR as I need to tackle the problem from a different angle. I'm closing this PR for now.

PR #7164, opened 2019-09-03
CSW - Add drag after deploy setting, fix container storage

PabstMirror:


addons/csw/stringtable.xml
```diff
@@ -78,6 +78,12 @@
         <Key ID="STR_ACE_CSW_progressBarTimeCoefficent_description">
             <English>Scales time required to assemble and reload static weapons</English>
         </Key>
+        <Key ID="STR_ACE_CSW_dragAfterDeploy_displayName">
+            <English>Drag tripods after deploying</English>
+        </Key>
+        <Key ID="STR_ACE_CSW_dragAfterDeploy_description">
+            <English>After deploying a tripod, start dragging it to a precise location</English>
```
synixebrett: unnecessary comma

TheMagnetar: These are two sentences so i think it is fine

PR #7169, opened 2019-09-05
Added player team color now stored as variable

ColdEvul:
- Added player team color stored as variable 

Can be usefull for scripting. I intend to use it to ensure color is used on respawn for instance

### DIFF
--- a/addons/interaction/functions/fnc_joinTeam.sqf
+++ b/addons/interaction/functions/fnc_joinTeam.sqf
@@ -19,6 +19,7 @@
 params ["_unit", "_team"];
 
 ["CBA_teamColorChanged", [_unit, _team]] call CBA_fnc_globalEvent;
+SETVAR(_unit, QGVAR(player_team), _team);
 
 // display message
 if (_unit == ACE_player) then {
--- a/addons/interaction/functions/fnc_joinTeam.sqf
+++ b/addons/interaction/functions/fnc_joinTeam.sqf
@@ -19,7 +19,7 @@
 params ["_unit", "_team"];
 
 ["CBA_teamColorChanged", [_unit, _team]] call CBA_fnc_globalEvent;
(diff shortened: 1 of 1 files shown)


addons/interaction/functions/fnc_joinTeam.sqf
```diff
@@ -19,6 +19,7 @@
 params ["_unit", "_team"];
 
 ["CBA_teamColorChanged", [_unit, _team]] call CBA_fnc_globalEvent;
+SETVAR(_unit, QGVAR(player_team), _team);
```
jonpas: We don't really use that macro anymore. I'd suggest:
```suggestion
_unit setVariable [QGVAR(playerTeam), _team];
```

commy2:
Literally just https://community.bistudio.com/wiki/assignedTeam

I refuse having this as a work around, even if BI broke their shit for no reason.

jonpas:
> I refuse having this as a work around, even if BI broke their shit for no reason.

If the command is broken, why not provide a working solution? We can say the same about many other ACE and CBA systems.

commy2:
But this is so stupid it hurts. I don't want to take part in this embarrassment.

jonpas:
Add a comment what it's for.

PabstMirror:
Could also use the `"CBA_teamColorChanged"` event to set variables you need

ColdEvul:
I use a similar variable in our script package to make sure the team is reapplied on death.

I figured I just could export it to this function. Nothing serius or any game changing you just have a player team variable to play with.

Using the last team color changed to.

jonpas:
Hmm, it's a 1-liner to do what @PabstMirror said, that's not that hard for the few mods that want to use that.

ColdEvul:
This one can be closed then. :)

PR #7170, opened 2019-09-05
Medical Treatment - Add logging for splints

PabstMirror:
Adds activity log entry for applying splints
- [ ] should it go on triage card (I think not)?


TheMagnetar:
It should be loggedin the triage card in my opinion. It is a treatment afterall but I am no expert.

PR #7171, opened 2019-09-06
Use strncpy and disable MSVC unsafe warnings

jonpas:
- Use `strncpy` instead of `strncpy_s` everywhere.
- Disable MSVC `unsafe` warning
- Set 64-bit flag correctly in CMake from generator name.
- Disable `/arch:SSE2` flag on 64-bit builds - flag only exists in x86 MSVC, it is enabled in x64 by default.
- Fix indentation (mixed tabs and spaces) in root `CMakeLists.txt˙.

Merge into `medicalExtensionWork`.


extensions/CMakeLists.txt
```diff
@@ -18,19 +18,29 @@ option(USE_DIRECTX "USE_DIRECTX" OFF)
...
+
+    if(USE_64BIT_BUILD)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /Qpar-report:2") # Default SSE2
+    else()
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /arch:SSE2 /Qpar-report:2")
+    endif()
+
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS} /D _DEBUG /MTd /Zi /Ob0 /Od /RTC1")
+    set(CMAKE_CXX_FLAGS_MINSIZEREL     "${CMAKE_CXX_FLAGS} /MT /O1 /Ob1 /D NDEBUG")
+    set(CMAKE_CXX_FLAGS_RELEASE       "${CMAKE_CXX_FLAGS} /MT /O2 /Ob2 /D NDEBUG")
```
dedmen: Might aswell set /Ox for release?

jonpas: We always build `RelWithDebInfo` for releases.

Anyways, that's for separate PR should we do it.

PR #7174, opened 2019-09-07
Medical - Minor fixes

LachlanSneddon:
-  Switch BP in watchVariable to the standard order
- Correct Elastic Effectiveness on abrasionMinor (43 to 4)
- Correct incorrect selection of elements in findMostEffectiveWound


PabstMirror:
findMostEffectiveWound fix isn't that minor, good catch

PR #7181, opened 2019-09-14
Portuguese Language Updated

MatsuoYutani:
- Add about 600 lines translated to Portuguese
- Some lines were fixed and updated to make more sense

### DIFF
--- a/addons/advanced_ballistics/stringtable.xml
+++ b/addons/advanced_ballistics/stringtable.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project name="ACE">
     <Package name="Advanced_Ballistics">
         <Key ID="STR_ACE_Advanced_Ballistics_WindInfoKey">
--- a/addons/advanced_ballistics/stringtable.xml
+++ b/addons/advanced_ballistics/stringtable.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project name="ACE">
     <Package name="Advanced_Ballistics">
         <Key ID="STR_ACE_Advanced_Ballistics_WindInfoKey">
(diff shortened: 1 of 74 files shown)


addons/advanced_ballistics/stringtable.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
commy2: Which one is right?

Blutze: @MatsuoYutani What tool did you use to make these edits, and if it was bux's Tabler, which version?

MatsuoYutani: Hello @commy2  and @Blutze I used Bux's Tabler latest version 0.9.1

commy2: @bux

bux: The new one has BOM, looks fine to me

dedmen: > The new one has BOM, looks fine to me.
> Indeed the ace one is missing the BOM

But we want without bom.

bux: Then the other way round :D
tabler checks wether the original file has BOM and only adds it in that case.
https://github.com/bux/tabler/commit/c67fd0dc69655878d30be6c0002f7ebdde928a82#diff-17e621989897af9f01c2714076d01920R154

MatsuoYutani: So... Do I have to remake it?

Blutze: On the "files changed" tab here on the PR, you can see in which files you caused changes to the very first line. If you go through them on your fork and open them the GitHub webclient's editor, you should be able to see a little red dot as the first character that you can remove like a normal one.

commy2:
Thank you! :>

commy2:
https://github.com/acemod/ACE3/issues/367
Portuguese is at 100% 🎉

PR #7191, opened 2019-09-25
Spectator - Restore HUD when coming out of spectator

### LINKED ISSUE #7186: Leaving spectator does not restore HUD elements
- **Arma 3:** `1.94`
- **CBA:** `3.12.2`
- **ACE3:** `3.12.6`

**Description:**
When leaving ACE spectator all vanilla HUD elements stay hidden.

**Steps to reproduce:**
- Create mission
- place down player unit and some other units marked as playable
- use these CBA settings
```sqf
// ACE Spect
force ace_spectator_enableAI = false;
force ace_spectator_restrictModes = 1;
force ace_spectator_restrictVisions = 0;
```
- Enter and leave ace spectator
```sqf
[{
    [true] call ace_spectator_fnc_setSpectator;
}, [], 2] call CBA_fnc_waitAndExecute;
[{
    [false] call ace_spectator_fnc_setSpectator;
}, [], 4] call CBA_fnc_waitAndExecute;
```
- Doing this while alive seems to be important to reproduce this issue

**Expected behavior:**
Normally all HUD elements will be restored, in this case those will not

- Dedicated
- Editor (Singleplayer)

**Additional context:**
This happened during an OP of mine, players complained about missing HUD after respawn. Recreated the issue in 3den with steps above.

zharf:
Fixes #7186

- Restore HUD for player when coming out of spectator just in case it was hidden by `ace_spectator_fnc_cam_setCameraMode`

### DIFF
--- a/addons/spectator/functions/fnc_ui.sqf
+++ b/addons/spectator/functions/fnc_ui.sqf
@@ -187,4 +187,7 @@ if (_init) then {
 
     // Ensure chat is shown again
     showChat true;
+
+    // Restore HUD
+    [] call EFUNC(common,showHud);
 };


SilentSpike:
This should probably be done in the UI function (`fnc_ui.sqf`) for consistency

PR #7195, opened 2019-09-28
Medical - Death from sum of trauma

PabstMirror:
Loosely based on old medical - https://github.com/acemod/ACE3/blob/backupBeforeRewriteMerge/addons/medical/functions/fnc_determineIfFatal.sqf

edit:
Currently you can only die from large hits to chest or head,
certian modded weapons and armor make it impossible to reach these values, so you cannot kill a unit besides a slow bleedout
This adds a setting to allow death from "sum of trama" which is very similar to how old medical worked.

This also adds a setting `clearTraumaAfterBandage` which clears bodyPartDamage when fully bandaging a limb. This is somewhat similar to the old medical healHitpoint setting, but used differently now.


commy2:
How often does this come into play?

addons/medical_damage/functions/fnc_determineIfFatal.sqf
```diff
@@ -0,0 +1,60 @@
...
+    };
+    if (_part == 1 && {_woundDamage >= ORGAN_DAMAGE_THRESHOLD} && {random 1 < HEART_HIT_CHANCE}) exitWith {
+        // Fatal damage to torso has various results based on organ hit - Heart shot is lethal
+        TRACE_1("determineIfFatal: lethal heartshot",_woundDamage);
+        true breakOut "main";
+    };
+};
+if (EGVAR(medical,fatalDamageSource) in [1, 2]) then {
+    // Sum of trauma to critical areas can be fatal (e.g. many small hits)
+    private _damageThreshold = if (isPlayer _unit) then { EGVAR(medical,playerDamageThreshold) } else { EGVAR(medical,AIDamageThreshold) };
```
commy2: What about remote controlled by zeus?

addons/medical_damage/functions/fnc_determineIfFatal.sqf
```diff
@@ -0,0 +1,60 @@
...
+    private _headThreashhold = 1.25 * _damageThreshold;
+    private _bodyThreashhold = 1.5 * _damageThreshold;
+
+    _bodyPartDamage params ["_headDamage", "_bodyDamage"];
+
+    private _chanceFatal = 0;
+    if (_headDamage > _headThreashhold) then { _chanceFatal = _chanceFatal + 0.95 * (_headDamage - _headThreashhold); };
+    if (_bodyDamage > _bodyThreashhold) then { _chanceFatal = _chanceFatal + 0.75 * (_bodyDamage - _bodyThreashhold); };
```
commy2: I think the parameters should be in a config somehwere and I think this probability function should be displayed differently, with easer tweakable parameters. It should be a cumulative distribution function where you set some characteristic points instead of piecewise linear functions.

addons/medical_damage/functions/fnc_determineIfFatal.sqf
SilentSpike: Treated as AI intentionally

SilentSpike:
I like the direction you've gone with this (with the fatal damage sources setting) 👍

@commy2 The motivation behind this is to handle the case where a unit is downed and receiving lots of small arms fire. None of it will cause a single damage event capable of being fatal, but  you might expect the unit to die quickly because cumulatively the damage is fatal.

addons/medical_damage/functions/fnc_determineIfFatal.sqf
PabstMirror: What would that look like in sqf?

PR #7200, opened 2019-09-28
Change medical item masses to be same as old medical

mharis001:
- title, part of #6913


commy2:
But why?

mharis001:
Discussion in meeting concluded that the item masses be kept the same as old.

jonpas:
They are heavier than the first aid we replace.

We also found no explanation of why it was changed to 2 in the first place.

PR #7212, opened 2019-10-01
Remove Atropine

alganthe:
- Remove atropine

As per decided, atropine has been commented out for the moment, I feel that deleting the entries would just create more work later when we decide to add it back.

### DIFF
--- a/addons/medical_treatment/ACE_Medical_Treatment.hpp
+++ b/addons/medical_treatment/ACE_Medical_Treatment.hpp
@@ -605,6 +605,7 @@ class ADDON {
             maxDose = 6;
             incompatibleMedication[] = {};
         };
+        /*
         class Atropine {
             painReduce = 0;
             hrIncreaseLow[] = {-2, -5};
@@ -615,6 +616,7 @@ class ADDON {
             maxDose = 6;
             incompatibleMedication[] = {};
         };
+        */
         class PainKillers {
             painReduce = 0.1;
             timeInSystem = 600;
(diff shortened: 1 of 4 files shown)


SilentSpike:
>when we decide to add it back

Never again 😆

Drofseh:
I'd like to keep it personally.

Either way though, won't removing it like this break currently existing missions/loadouts.

jonpas:
We are removing it that has been decided. If anyone wants it, they can take this and make a patch mod that adds it. However there is little reason with epi and adenosine doing what they should.

dedmen:
> Either way though, won't removing it like this break currently existing missions/loadouts.

Missions no, they should just print a warning to RPT.
Saved loadouts in arsenal, not for ACE Arsenal, it just won't add that item when you load the loadout.
BI Arsenal loadouts might break?

alganthe:
BI arsenal loadouts break even if you don't do anything, I don't think we should worry about that.

SilentSpike:
We could just `scope=1` the item and keep it removed from the treatment config so it does nothing?

dedmen:
> I don't think we should worry about that.

Same thought.

> We could just scope=1 the item and keep it removed from the treatment config so it does nothing?

I did the same in TFAR worked fine until BI changed Arsenal to no longer accept scope=1 items, but they reverted that "fix" for now.

Drofseh:
I'd like a scope=1 item as well, maybe with strings that say it no longer has a use in Ace medical.

grenadier2510:
I am in favor of the return of atropine!

PR #7214, opened 2019-10-01
Medical - Increase severity of falling damage and Improve non-selectionSpecific wounds

PabstMirror:
Was: `thresholds[] = {{0.5, 5}, {0.3, 2}, {0.05, 1}};`
a serious fall would be divided into too many wounds and not trigger limping or fracture


PabstMirror:
-add comment as to why

PabstMirror:
https://github.com/acemod/ACE3/blob/master/addons/medical_engine/functions/fnc_handleDamage.sqf#L113
all the `_woundedHitPoint =` aren't doing what we think because collision is non-selectionSpecific

this now passes an actual array of weighted damage for random selection,
it previously did 
`floor random 6`
 now does
```
selectRandomWeighted [0, _damageHead, 1, _damageBody, 2, _damageLeftArm, 3, _damageRightArm, 4, _damageLeftLeg, 5, _damageRightLeg]
```
it should mean that other non-location specific wounds will go to more correct locations.
e.g. an explosion on your left should be more likely to goto left arm and leg

SilentSpike:
I like the weighted selection stuff for non-selection specific damage. Perhaps we should use macros for the indexes though just to make it more clear what is being done when that variable is changed.

PR #7215, opened 2019-10-01
Medical Feedback - effects overhaul

BaerMitUmlaut:
- Fix a black flash the first time a white pain flash would appear
- Intensify all pain effects
- Add a full screen blur effect when in high pain
- Readd chromatic aberration pain effect
- Add pain effect without visual effects besides high pain blur
- Fix and overhaul injury and pain sounds
- Make sure bloody screen instantly appears when getting hit for better visual feedback


addons/medical_feedback/XEH_postInit.sqf
```diff
@@ -80,10 +88,28 @@ GVAR(heartBeatEffectRunning) = false;
...
 }] call CBA_fnc_addPlayerEventHandler;
 
+// Forced say3D
+[QGVAR(forceSay3D), {
+    params ["_unit", "_sound", "_distance"];
+
+    if (ACE_player distance _unit > _distance) exitWith {};
+
+    if (vehicle _unit == _unit) then {
+        // say3D waits for the previous sound to finish, so use a dummy instead
```
PabstMirror: would normal `say` work ok here? we don't really need 3d positioning

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
```diff
@@ -26,60 +26,35 @@ params [["_unit", objNull, [objNull]], ["_type", "hit", [""]], ["_severity", 0,
...
-
-if (isNil "_sound") exitWith {
-    ERROR("No sound for this speaker");
-};
-
-// Delete leading slash.
-if (_sound select [0, 1] == "\") then {
-    _sound = _sound select [1];
+// Fallback if speaker has no associated scream/moan sound
+if (isNull (configFile >> "CfgSounds" >> format ["ACE_moan_%1_low_1"])) then {
```
PabstMirror: format missing %1 arg

addons/medical_feedback/XEH_postInit.sqf
BaerMitUmlaut: No, because say waits for a previous say.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
```diff
@@ -26,60 +26,35 @@ params [["_unit", objNull, [objNull]], ["_type", "hit", [""]], ["_severity", 0,
...
+} else {
+    [10, 15, 20] select _severity;
 };
+private _filter = format ["(configName _x) find 'ACE_%1_%2_%3' == 0", _type, _speaker, _variation];
+private _sounds = _filter configClasses (configFile >> "CfgSounds");
+private _sound = configName selectRandom _sounds;
 
-playSound3D [_sound, objNull, false, position _unit, _volume, _frequency, _distance];
+// Limit network traffic by only sending the event to players who can potentially hear it
+private _targets = (ASLToAGL getPosASL _unit) nearEntities ["CAManBase", _distance];
```
PabstMirror: nearEntities should be able to take object on left hand side

BaerMitUmlaut: I don't think we should filter for players. If it's not a player, then it will likely be a HC or server owned unit, and the CBA variable will be sent to the server either way. Extra traffic to the HC doesn't hurt because it should run on the same machine as the server.

addons/medical_feedback/XEH_postInit.sqf
```diff
@@ -80,10 +88,28 @@ GVAR(heartBeatEffectRunning) = false;
...
+        private _dummy = "#dynamicsound" createVehicleLocal [0, 0, 0];
+        _dummy attachTo [_unit, [0, 0, 0], "camera"];
+        _dummy say3D [_sound, _distance, 1, false];
+
+        [{deleteVehicle _this}, _dummy, 5] call CBA_fnc_waitAndExecute;
```
commy2: `detach` _dummy before `deleteVehicle`, otherwise the `attachedObjects` array on _unit builds up null entries.

addons/medical_feedback/functions/fnc_initEffects.sqf
```diff
@@ -82,6 +100,6 @@ GVAR(ppIncapacitationGlare) = [
 
 GVAR(ppIncapacitationBlur) = [
     "DynamicBlur",
-    814, // 135xx does not work
+    815, // 135xx does not work
```
commy2: These should be macros.

addons/medical_feedback/XEH_postInit.sqf
```diff
@@ -3,6 +3,10 @@
 [QEGVAR(medical,injured), {
     params ["_unit", "_painLevel"];
     [_unit, "hit", PAIN_TO_SCREAM(_painLevel)] call FUNC(playInjuredSound);
+
+    if (hasInterface) then {
```
PabstMirror: zeus or single-player
```suggestion
    if (hasInterface && {_unit == ace_player}) then {
```

addons/medical_feedback/functions/fnc_handleEffects.sqf
```diff
@@ -44,10 +45,7 @@ if ((!GVAR(heartBeatEffectRunning)) && {_heartRate != 0} && {(_heartRate > 160)
     true, linearConversion [BLOOD_VOLUME_CLASS_2_HEMORRHAGE, BLOOD_VOLUME_CLASS_4_HEMORRHAGE, _bloodVolume, 0, 1, true]
 ] call FUNC(effectBloodVolume);
 
-if (!_unconscious) then {
-    [true, _pain] call FUNC(effectPain);
-};
-
-[true, _bleedingStrength] call FUNC(effectBleeding);
+[!_unconscious, _pain, _manualUpdate] call FUNC(effectPain);
```
PabstMirror: not used
```suggestion
[!_unconscious, _pain] call FUNC(effectPain);
```

PabstMirror:
looks pretty good
imho later we might want to add a setting for moan frequency 
(I think it's too often, but might just be that it wasn't working before)

addons/medical_feedback/XEH_postInit.sqf
BaerMitUmlaut: `_unit == ace_player` would suffice.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
```diff
@@ -26,60 +26,35 @@ params [["_unit", objNull, [objNull]], ["_type", "hit", [""]], ["_severity", 0,
...
-    _sound = _sound + ".wss";
+// Select actual sound
+private _variation = ["low", "mid", "high"] select _severity;
+private _distance = if (_type == "hit") then {
+    [50, 60, 70] select _severity;
+} else {
+    [10, 15, 20] select _severity;
 };
+private _filter = toLower format ["(toLower configName _x) find 'ACE_%1_%2_%3' == 0", _type, _speaker, _variation];
+private _sounds = _filter configClasses (configFile >> "CfgSounds");
```
PabstMirror: `"(tolower configname _x) find 'ace_hit_male08eng_mid' == 0" configClasses (configFile >> "CfgSounds");`
is 3.9ms, probably worse with full modset (more CfgSounds entries)
I think we need to do caching at start

PabstMirror: switched to a while loop which takes about 0.05ms
could fully cache at game start, but this should be ok for now

addons/medical_feedback/XEH_postInit.sqf
PabstMirror: `attachedObjects player`: `[<NULL-object>,<NULL-object>,<NULL-object>,<NULL-object>,<NULL-object>...]`

PR #7218, opened 2019-10-02
Remove duplicate class declaration from laserpointer

Dahlgren:
- Fix laserpointer not building

https://github.com/acemod/ACE3/blob/e8b104fe6129f33c1fb2b981afdf4dcf38a2ee13/addons/laserpointer/CfgWeapons.hpp#L3


jonpas:
Regression from #7217.

dedmen:
Ah I see
https://github.com/acemod/ACE3/blob/master/addons/laserpointer/CfgWeapons.hpp#L3
This got converted to the new syntax, so we should fix that too?

PR #7224, opened 2019-10-03
Fast medical item replacement

BaerMitUmlaut:
- Use the new item replacement framework for medical items

Requires #7064.


veteran29:
Will this be documented on wiki?

BaerMitUmlaut:
The config or what exactly?

veteran29:
> The config or what exactly?

`registerItemReplacement.sqf ` should be mentioned somewhere on wiki so addon creators will be aware that they can register their own replacements.

commy2:
Does it handle GM mod items?

BaerMitUmlaut:
Yes.

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -0,0 +1,114 @@
...
+        {
+            if (_item isKindOf [_x, _cfgWeapons]) then {
+                private _inheritedReplacements = GVAR(itemReplacements) getVariable ["$" + _x, []];
+                _replacements append _inheritedReplacements;
+            };
+        } forEach GVAR(inheritedReplacements);
+
+        // Skip lookup for all following items of this class
+        private _count = 1;
+        while {_newItems#(_i + 1) == _item} do {
```
PabstMirror: This can go out of bounds for the array, but `while {nil}` fails silently so it's probably fine, maybe leave a note

addons/medical_treatment/XEH_preInit.sqf
```diff
@@ -44,9 +44,27 @@ GVAR(facilityClasses) = [];
...
+    {
+        _x params ["_item", "_count"];
+        for "_i" from 1 to _count do {
+            _replacements pushBack _item;
+        };
+    } forEach getArray _x;
+
+    // check if replacement is for item type or class name
+    private _configName = configName _x;
+    private _toReplace = if (_configName find "ItemType_" == 0) then {
```
PabstMirror: case-insensitive
```suggestion
    private _toReplace = if ((_configName select [0,9]) == "ItemType_") then {
```

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -0,0 +1,114 @@
...
+// $ prefix is used for types (numbers) and replacements with inheritance
+if (_replaceInherited) then {
+    GVAR(inheritedReplacements) pushBack _oldItem;
+    _oldItem = "$" + _oldItem;
+};
+if (_oldItem isEqualType 0) then {
+    _oldItem = "$" + str _oldItem;
+};
+if (_newItems isEqualType "") then {
+    _newItems = [toLower _newItems];
```
PabstMirror: why `toLower`?

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -0,0 +1,114 @@
...
+    private _items = items _unit;
+    if (_items isEqualTo GVAR(oldItems)) exitWith {};
+
+    private _newItems = _items - GVAR(oldItems);
+    if (_newItems isEqualTo []) exitWith {
+        GVAR(oldItems) = _items;
+    };
+
+    _newItems sort true; // Sort so all items of current class can be replaced at once
+    private _replacements = [];
```
PabstMirror: need to move this inside for-loop so it gets reset for each item type

PabstMirror:
ref #7212
think about converting atropine to epi

addons/medical_treatment/XEH_preInit.sqf
BaerMitUmlaut: Eh, whatever 👍

addons/common/functions/fnc_registerItemReplacement.sqf
BaerMitUmlaut: Right, that doesn't seem necessary here.

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -32,3 +32,32 @@ if (isServer) then {
 // Logging events
 [QGVAR(addToLog), LINKFUNC(addToLog)] call CBA_fnc_addEventHandler;
 [QGVAR(addToTriageCard), LINKFUNC(addToTriageCard)] call CBA_fnc_addEventHandler;
+
+// replace medical items with their ACE equivalents
+["ace_settingsInitialized", {
+    TRACE_1("ace_settingsInitialized EH",GVAR(convertItems)); // 0: Enabled 1: RemoveOnly 2:Disabled
+    if (GVAR(convertItems) == 2) exitWith {};
+    configProperties [configFile >> QEGVAR(medical,replacementItems), "isArray _x"] apply {
```
mharis001: This should be changed to a `forEach` loop now.

Issue #7227, opened 2019-10-04
Medical Rewrite: Duplicate functions


BaerMitUmlaut:
There are two versions of fnc_adjustPainLevel and fnc_setUnconscious.

In my opinion, it is fine to have a public API in medical which is named the same as the original, but they should simply do stricter params checking and delegate the call to the internal functions. In this case, they do different things though.

I marked this as critical, because since it serves as public API, it needs to be fixed before release.

PabstMirror:
adjustPain:
private one (medical_status) should be renamed to `increasePainTo` as it can only increase
public one should work same as before rewrite

setUncon
private one (medical_status) should be renamed `enteredUnconscious`
public one should work same as before rewrite

SilentSpike:
I tend to agree with the OP and also Pabst's comments

jonpas:
Well, this was released, so Public API is what it is now.

PR #7231, opened 2019-10-06
Rework pain simulation

### LINKED ISSUE #5511: Pain always descreases from an untreated wound.
Testing the medical rewrite with untreated wounds we found the following.

**Observations:**
Pain never increases from an untreated wound. 
It always goes down on its own even when wound is left untreated. 
Pain never stays unchanging or fixed once it is present.
Pain always decreases.

**Thoughts:**
If left this way morphine is useless. 

**Expected:** 
Pain once it exists should either remain steady or increase if the wounds that caused it are left untreated. Especially if movement continues.

BaerMitUmlaut:
- Move pain simulation from Medical Status to Medical Vitals
- Dynamically calculate pain from open wounds, tourniquets and custom pain from API
- Rewrite pain API (see #7227)
  - Using it like the old API is still possible
  - Allows mod or component specific pain
  - Allows creating non-decaying pain

Would like to get some feedback on both the new simulation as well as the API.

Closes #5511.


PabstMirror:
Wounds stay in array after being bandaged, this is for connection to wound opening (fnc_handleBandageOpening.sqf) so will never decrease pain


can solve this one with `selectMax (_x apply { if (_x2 > 0) then {_x#5} else {0} })` I think



also problem with bruises which show up in openWounds, but cannot be healed as they don't cause bleeding

not sure how to solve this; maybe move to bandagedWounds on creation?

addons/medical_vitals/functions/fnc_updatePain.sqf
```diff
@@ -0,0 +1,86 @@
...
+ * Example:
+ * [player, 0, 1, false] call ace_medical_vitals_fnc_updatePain
+ *
+ * Public: No
+ */
+
+params ["_unit", "_painSupressAdjustment", "_deltaT", "_syncValue"];
+
+// Pain value formula: maximum of
+// - most painful open wound * (1 + 0.1 * count of open wounds)
```
PabstMirror: `count of open wounds` not used

jonpas:
Close? Probably done in some other recent PR or is just oudated.

LinkIsGrim:
Outdated, but implementation could be adapted. I like the idea.

PR #7232, opened 2019-10-07
Common - Optimize arithmeticGetResult function

mharis001:
- Optimize `arithmeticGetResult` common function by using `selectMax` and `selectMin` commands rather than `count` loop
    - Will provide a small performance improvement
    - For the current `ACE_setCustomAimCoef` (2 sources) with the "max" operation: ~0.0234 ms to ~0.0207 ms
- Change `count` to `forEach` (faster than `count` + `true/false/nil`)
- Cleanup some unnecessary parentheses

### DIFF
--- a/addons/common/functions/fnc_arithmeticGetResult.sqf
+++ b/addons/common/functions/fnc_arithmeticGetResult.sqf
@@ -1,12 +1,12 @@
 #include "script_component.hpp"
 /*
  * Author: PabstMirror
- * Gets arithmetic result from a set.
+ * Returns the arithmetic result of performing the given operation on a set.
  *
  * Arguments:
- * 0: Namespace <OBJECT><LOCATION><MISSIONNAMESPACE>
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_arithmeticGetResult.sqf
```diff
@@ -18,51 +18,44 @@
  * Public: Yes
  */
 
-params ["_namespace", "_setID", "_op"];
-TRACE_3("arithmeticGetResult",_namespace,_setID,_op);
+params ["_namespace", "_setID", "_operation"];
+TRACE_3("arithmeticGetResult",_namespace,_setID,_operation);
 
 private _data = (_namespace getVariable _setID) param [2, []];
```
PabstMirror: I do NOT think this is necessary, but it would make the func safer as `selectMax []` is undefined (and div0 in "avg")
```suggestion
private _data = (_namespace getVariable _setID) param [2, [{0}]];
```

PR #7236, opened 2019-10-11
Medical - Make some constants variables

PabstMirror:
Changes some hard constants to variables set to default values
I think making these settings is overkill, 
but this still allows manual tweaking of ace if desired without having to build a custom pbo

I would like to replace macros with the actual gvars unless it would be better to keep them?


commy2:
Could we calculate the constants on the fly and make the 10% and 90% thing a setting?

addons/medical_damage/functions/fnc_determineIfFatal.sqf
```diff
@@ -49,7 +46,7 @@ if (EGVAR(medical,fatalDamageSource) in [1, 2]) then {
     _bodyPartDamage params ["_headDamage", "_bodyDamage"];
 
     private _vitalDamage = ((_headDamage - _headThreshhold) max 0) + ((_bodyDamage - _bodyThreshhold) max 0);
-    private _chanceFatal = 1 - exp -((_vitalDamage/WEIBULL_L)^WEIBULL_K);
+    private _chanceFatal = 1 - exp -((_vitalDamage/FATAL_SUM_DAMAGE_WEIBULL_L)^FATAL_SUM_DAMAGE_WEIBULL_K);
     TRACE_3("",_bodyPartDamage,_vitalDamage,_chanceFatal);
 
     if (_chanceFatal > random 1) exitWith {
```
commy2:
```sqf
    if (_chanceFatal > random 1 || _vitalDamage == 1) exitWith {
```
Since `_chanceFatal` will never reach 1 (and could be << 1 for x=1, if k and λ allow it), this would guarantee death.

addons/medical_engine/XEH_preInit.sqf
```diff
@@ -6,6 +6,19 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+// Define "Constants" variables
+HEAD_DAMAGE_THRESHOLD = HEAD_DAMAGE_THRESHOLD_DEFAULT;
```
dedmen: Maybe also note that both sides are macros. I know its clear from the allcaps.
But a beginner will just think that HEAD_DAMAGE_THESHOLD is a variable.

PabstMirror: I think we should just replace all use of `HEAD_DAMAGE_THRESHOLD` with it's var

addons/medical_damage/functions/fnc_determineIfFatal.sqf
PabstMirror: _vitalDamage isn't capped at 1
```
[ACE] (medical_damage) TRACE: 25999 : _bodyPartDamage=[3.68512,0,0,0,0,0], _vitalDamage=2.43512, _chanceFatal=1 z\ace\addons\medical_damage\functions\fnc_determineIfFatal.sqf:37

addons/medical_engine/XEH_preInit.sqf
PabstMirror: on second thought, I think it might be cleaner to just keep the macros

PR #7238, opened 2019-10-12
Medical Feedback - Change pain scream frequency based on intensity 

PabstMirror:
Uses `handleUnitVitals` as a "1 second clock", so we don't have to check pain and generate a moan event each frame.
Scales moan timeout based on intensity.


addons/medical_feedback/stringtable.xml
```diff
@@ -65,39 +65,11 @@
...
-                <Hungarian>Engedélyezi a sérült egységek kiáltásait</Hungarian>
-                <Italian>Abilita Grida da parte delle unità ferite</Italian>
-                <Japanese>負傷したユニットを叫ぶようにします。</Japanese>
-                <Korean>부상당한 인원이 소리지르는것을 활성화합니다</Korean>
-                <Chinesesimp>启用伤者的尖叫声</Chinesesimp>
-                <Chinese>啟用傷者的尖叫聲</Chinese>
-            </Key>
-        </Container>
+        <Key ID="STR_ACE_Medical_Feedback_painScreamFrequency_DisplayName">
+            <English>Pain scream frequency</English>
```
mharis001:
```suggestion
            <English>Pain Scream Frequency</English>
```

addons/medical_feedback/stringtable.xml
```diff
@@ -65,39 +65,11 @@
...
+        </Key>
+        <Key ID="STR_ACE_Medical_Feedback_painScreamFrequency_Description">
+            <English>How often a unit screams from being in pain</English>
```
mharis001:
```suggestion
            <English>Controls how often a unit will scream from being in pain.</English>
```

BaerMitUmlaut:
Could we make frequency dependant on pain amount?

SilentSpike:
>Could we make frequency dependant on pain amount?

I quite like this idea, not sure we need a setting as opposed to just finding an acceptable frequency that isn't annoying

PR #7244, opened 2019-10-20
Remove PhysX from ugl smoke grenades

commy2:
With default values, the only (apparent) differences between shotSmokeX and shotDeploy are:
- no smoke creation
- no smoke sound
- old physics with configurable bounce (`deflectionSlowDown` etc.)

shotSmoke has smoke creation, but no smoke sounds, which is why #4648 was rejected.

Best way to achieve old physcis, but with smoke sound was to attach a shotSmokeX to a shotShell/shotSmoke/shotDeploy. Of these shotDeploy is the only one that has no side effects (explosion of shotShell, [duplicate] smoke creation of shotSmoke) with default values.
There don't seem to be duplicated bounce sound effects, likely because (most) physics are disabled for attached objects.

Both, the "physics projectile" (shotDeploy) and the "smoke projectile" (shotSmokeX) are deleted after timeToLive.

### DIFF
--- a/addons/grenades/functions/fnc_throwGrenade.sqf
+++ b/addons/grenades/functions/fnc_throwGrenade.sqf
@@ -26,9 +26,9 @@ if (isNull _projectile) then {
 private _config = configFile >> "CfgAmmo" >> _ammo;
 
 // fix smoke bounce
-if (isText (_config >> QGVAR(attachedAmmo)) && {local _unit}) then {
+if (isText (_config >> QGVAR(attachedAmmo))) then {
     private _attachedAmmo = getText (_config >> QGVAR(attachedAmmo));
-    private _attachedProjectile = _attachedAmmo createVehicle [0,0,0];
+    private _attachedProjectile = _attachedAmmo createVehicleLocal [0,0,0];
     _attachedProjectile attachTo [_projectile, [0,0,0]];
 };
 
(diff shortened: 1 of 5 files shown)


addons/grenades/functions/fnc_throwGrenade.sqf
```diff
@@ -18,15 +18,22 @@
...
     _projectile = nearestObject [_unit, _ammo];
 };
 
 private _config = configFile >> "CfgAmmo" >> _ammo;
 
+// fix smoke bounce
+if (isText (_config >> QGVAR(attachedAmmo)) && {local _unit}) then {
+    private _attachedAmmo = getText (_config >> QGVAR(attachedAmmo));
+    private _attachedProjectile = _attachedAmmo createVehicle [0,0,0];
+    _attachedProjectile attachTo [_projectile, [0,0,0]];
```
PabstMirror: attachTo has global effects
probably want this inside if (local _unit

PabstMirror:
Doesn't work in mp, probably can't attach to simulation type
test

commy2:
It may be that every client has their own projectile, so it needs createVehicleLocal and global execution. I will change it, can you retry then @PabstMirror ?

PabstMirror:
floating in mid air
test2

also, I think i heard multiple "grenade pin" effects from shooters location, not positive (cvl might not be local)

jonpas:
We could provide an optional for this. We use @BaerMitUmlaut's version: https://github.com/Theseus-Aegis/Mods/blob/5ce562312c08d9a6d01e046886c29d09f874c991/addons/ballistics/CfgAmmo.hpp

But it's also so simple that any group that would use it probably has their own version of it already.

rautamiekka:
> We could provide an optional for this. We use @BaerMitUmlaut's version: https://github.com/Theseus-Aegis/Mods/blob/5ce562312c08d9a6d01e046886c29d09f874c991/addons/ballistics/CfgAmmo.hpp
> 
> But it's also so simple that any group that would use it probably has their own version of it already.

Accessibility is more important than how simple it is.

PR #7253, opened 2019-10-28
Medical Treatment - Reset canTreatCached on treatment

PabstMirror:
`#define CAN_TREAT_CONDITION_CACHE_EXPIRY 2`
which can lead to treatments showing up that are no longer relevant

This resets condition cache after treatment occurs
because treatment may happen on a different machine there is unknown delay 
(this client->server and server->patient), 
so we have to keep polling the variable to see when the treatment actually happened

An alternative way of fixing this is to run something like this on the patient's machine
`_patient setVariable [format [QGVAR(canTreat_%1_%2), _bodyPart, _classname], nil, true];`, 
this might be better as it updates canTreatCached on all machines (eg 2 medics treating) 
but generates extra network traffic

### DIFF
--- a/addons/medical_treatment/functions/fnc_bandageLocal.sqf
+++ b/addons/medical_treatment/functions/fnc_bandageLocal.sqf
@@ -21,10 +21,10 @@ params ["_patient", "_bodyPart", "_bandage"];
 TRACE_3("bandageLocal",_patient,_bodyPart,_bandage);
 
 private _partIndex = ALL_BODY_PARTS find toLower _bodyPart;
-if (_partIndex < 0) exitWith {false};
+if (_partIndex < 0) exitWith {};
 
 private _openWounds = GET_OPEN_WOUNDS(_patient);
-if (_openWounds isEqualTo []) exitWith {false};
+if (_openWounds isEqualTo []) exitWith {};
 
 // Figure out which injury for this bodypart is the best choice to bandage
 // TODO also use up the remainder on left over injuries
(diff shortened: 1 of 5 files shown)


PabstMirror:
The other way would look like:
```
// Reset canTreatCached globally
_patient setVariable [format [QGVAR(%1_removeTourniquet), _bodyPart], nil, true];
_patient setVariable [format [QGVAR(%1_applyTourniquet), _bodyPart], nil, true];
```
but even messier for bandages (but only when bleeding stops)
global way would possibly better for multi medics, but I really don't like syncing all these extra variables (even though they are nils, they still show up in `allVariables`)

SilentSpike:
I think network traffic wouldn't be the worst since it's event driven by treatment and treatment doesn't really happen that frequently or all the time.

PabstMirror:
Now using 
```
private _nearPlayers = (_patient nearEntities ["Man", 6]) select {_x call EFUNC(common,isPlayer)};
[QEGVAR(interact_menu,clearConditionCaches), [], _nearPlayers] call CBA_fnc_targetEvent;
```
so code only runs on very close units

addons/medical_treatment/functions/fnc_bandageLocal.sqf
```diff
@@ -77,4 +77,8 @@ if (GVAR(clearTraumaAfterBandage)) then {
     };
 };
 
-true
+if (_amountOf <= 0) then { // Reset treatment condition cache for nearby players if we stopped all bleeding
+    private _nearPlayers = (_patient nearEntities ["Man", 6]) select {_x call EFUNC(common,isPlayer)};
```
mharis001: `CAManBase`?

PR #7254, opened 2019-10-29
Add some editor previews

PabstMirror:
`0 = [nil,"all",[],[],[],["ace_medicalSupplyCrate"]] spawn BIS_fnc_exportEditorPreviews;`


jonpas:
Does it have correct background color? You need to set some graphical option to get vanilla-like result.

_Just my OCD._

PabstMirror:
jpgs look washed out, but they are fairly normal in game

PR #7255, opened 2019-10-29
Medical AI - Improve Healing Logic

LachlanSneddon:
-Allows AI to deal with wounded in the cardiac arrest state. 
-Refines AI usage of blood IV's to minimise wastage.

### DIFF
--- a/addons/medical_ai/functions/fnc_healingLogic.sqf
+++ b/addons/medical_ai/functions/fnc_healingLogic.sqf
@@ -55,7 +55,7 @@ switch (true) do {
         _treatmentArgs = [_target, _selection, "FieldDressing"];
     };
     case (IN_CRDC_ARRST(_target)): {
-        if(EGVAR(medical_treatment,cprSuccessChance) == 0) then {
+        if (EGVAR(medical_treatment,cprSuccessChance) == 0) then {
             _treatmentEvent = '#WaitForHeartRate';
             [QEGVAR(medical,CPRSucceeded), _target] call CBA_fnc_localEvent;
         } else {
--- a/addons/medical_ai/functions/fnc_healingLogic.sqf
+++ b/addons/medical_ai/functions/fnc_healingLogic.sqf
@@ -54,9 +54,19 @@ switch (true) do {
         _treatmentTime = 5;
         _treatmentArgs = [_target, _selection, "FieldDressing"];
     };
+    case (IN_CRDC_ARRST(_target)): {
+        _treatmentEvent = QEGVAR(medical_treatment,cprLocal);
+        _treatmentTime = 15;
+        _treatmentArgs = [_healer,_target];    
+    };
     case (_isMedic && {GET_BLOOD_VOLUME(_target) < BLOOD_VOLUME_CLASS_2_HEMORRHAGE}): {
-        private _bloodBags = _target getVariable [QEGVAR(medical,ivBags), []];
-        if ((count _bloodBags) >= 2) exitWith {
+        //check if patient volume + currentIV volume is enough to allow the patient to wake up
+        private _totalIvVolume = 0; //in ml
+        {
+            _x params ["_volumeRemaining"];
+            _totalIvVolume = _totalIvVolume + _volumeRemaining;
+        } forEach (_target getVariable [QEGVAR(medical,ivBags), []]);        
+        if ((GET_BLOOD_VOLUME(_target) + (_totalIvVolume/1000)) > BLOOD_VOLUME_CLASS_2_HEMORRHAGE) exitWith {
             _treatmentEvent = "#waitForBlood";
         };
         _treatmentEvent = QEGVAR(medical_treatment,ivBagLocal);
--- a/addons/medical_ai/functions/fnc_healingLogic.sqf
+++ b/addons/medical_ai/functions/fnc_healingLogic.sqf
@@ -57,7 +57,6 @@ switch (true) do {
     case (IN_CRDC_ARRST(_target)): {
         if (EGVAR(medical_treatment,cprSuccessChance) == 0) then {
             _treatmentEvent = "#WaitForHeartRate";
-            [QEGVAR(medical,CPRSucceeded), _target] call CBA_fnc_localEvent;
         } else {
             _treatmentEvent = QEGVAR(medical_treatment,cprLocal);
             _treatmentArgs = [_healer, _target];
--- a/addons/medical_ai/functions/fnc_healingLogic.sqf
+++ b/addons/medical_ai/functions/fnc_healingLogic.sqf
@@ -56,7 +56,7 @@ switch (true) do {
     };
     case (IN_CRDC_ARRST(_target)): {
         if (EGVAR(medical_treatment,cprSuccessChance) == 0) then {
-            _treatmentEvent = '#WaitForHeartRate';
+            _treatmentEvent = "#WaitForHeartRate";
             [QEGVAR(medical,CPRSucceeded), _target] call CBA_fnc_localEvent;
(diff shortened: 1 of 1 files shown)


addons/medical_ai/functions/fnc_healingLogic.sqf
```diff
@@ -54,9 +54,19 @@ switch (true) do {
         _treatmentTime = 5;
         _treatmentArgs = [_target, _selection, "FieldDressing"];
     };
+    case (IN_CRDC_ARRST(_target)): {
+        _treatmentEvent = QEGVAR(medical_treatment,cprLocal);
+        _treatmentTime = 15;
+        _treatmentArgs = [_healer,_target];
```
mharis001:
```suggestion
        _treatmentArgs = [_healer, _target];    
```

addons/medical_ai/functions/fnc_healingLogic.sqf
```diff
@@ -54,9 +54,19 @@ switch (true) do {
...
     case (_isMedic && {GET_BLOOD_VOLUME(_target) < BLOOD_VOLUME_CLASS_2_HEMORRHAGE}): {
-        private _bloodBags = _target getVariable [QEGVAR(medical,ivBags), []];
-        if ((count _bloodBags) >= 2) exitWith {
+        //check if patient volume + currentIV volume is enough to allow the patient to wake up
+        private _totalIvVolume = 0; //in ml
+        {
+            _x params ["_volumeRemaining"];
+            _totalIvVolume = _totalIvVolume + _volumeRemaining;
+        } forEach (_target getVariable [QEGVAR(medical,ivBags), []]);        
+        if ((GET_BLOOD_VOLUME(_target) + (_totalIvVolume/1000)) > BLOOD_VOLUME_CLASS_2_HEMORRHAGE) exitWith {
```
mharis001:
```suggestion
        if (GET_BLOOD_VOLUME(_target) + (_totalIvVolume / 1000) > BLOOD_VOLUME_CLASS_2_HEMORRHAGE) exitWith {
```

addons/medical_ai/functions/fnc_healingLogic.sqf
```diff
@@ -54,9 +54,19 @@ switch (true) do {
...
     };
+    case (IN_CRDC_ARRST(_target)): {
+        _treatmentEvent = QEGVAR(medical_treatment,cprLocal);
+        _treatmentTime = 15;
+        _treatmentArgs = [_healer,_target];    
+    };
     case (_isMedic && {GET_BLOOD_VOLUME(_target) < BLOOD_VOLUME_CLASS_2_HEMORRHAGE}): {
-        private _bloodBags = _target getVariable [QEGVAR(medical,ivBags), []];
-        if ((count _bloodBags) >= 2) exitWith {
+        //check if patient volume + currentIV volume is enough to allow the patient to wake up
```
mharis001:
```suggestion
        // Check if patient's blood volume + remaining IV volume is enough to allow the patient to wake up
```

addons/medical_ai/functions/fnc_healingLogic.sqf
```diff
@@ -54,9 +54,19 @@ switch (true) do {
         _treatmentTime = 5;
         _treatmentArgs = [_target, _selection, "FieldDressing"];
     };
+    case (IN_CRDC_ARRST(_target)): {
+        _treatmentEvent = QEGVAR(medical_treatment,cprLocal);
```
SilentSpike: What if cpr success chance is set to 0? Then they're just extending the cardiac arrest timer but not actually treating the patient (in such a case the only treatment is a PAK I believe).

LachlanSneddon: good point, but I believe the PAK needs requires isInStableCondition, so they need to be awake and have a heart rate to use it.

SilentSpike: I think that requirement might have been removed, but so much has changed I need to double check that 😆

In any case, we should add something to handle this case

LachlanSneddon: Would it be more prudent to just give them something along the line of:
```
if(GVAR(cprSuccessChance) == 0) exitWith {
_treatmentEvent = QEGVAR(medical_treatment,cprSuccess);
};
```
as a PAK might be a bit overkill on the AI healing system.

LachlanSneddon: Updated to cover this case. I don't know if it was the right way to go about it but it definitely seems to work.

addons/medical_ai/functions/fnc_healingLogic.sqf
```diff
@@ -54,9 +54,24 @@ switch (true) do {
         _treatmentTime = 5;
         _treatmentArgs = [_target, _selection, "FieldDressing"];
     };
+    case (IN_CRDC_ARRST(_target)): {
+        if(EGVAR(medical_treatment,cprSuccessChance) == 0) then {
```
mharis001:
```suggestion
        if (EGVAR(medical_treatment,cprSuccessChance) == 0) then {
```

addons/medical_ai/functions/fnc_healingLogic.sqf
```diff
@@ -54,9 +54,24 @@ switch (true) do {
...
+            _treatmentEvent = '#WaitForHeartRate';
```
mharis001:
```suggestion
            _treatmentEvent = "#WaitForHeartRate";
```

addons/medical_ai/functions/fnc_healingLogic.sqf
```diff
@@ -54,9 +54,24 @@ switch (true) do {
...
+    };
     case (_isMedic && {GET_BLOOD_VOLUME(_target) < BLOOD_VOLUME_CLASS_2_HEMORRHAGE}): {
-        private _bloodBags = _target getVariable [QEGVAR(medical,ivBags), []];
-        if ((count _bloodBags) >= 2) exitWith {
+        // Check if patient's blood volume + remaining IV volume is enough to allow the patient to wake up
+        private _totalIvVolume = 0; //in ml
+        {
+            _x params ["_volumeRemaining"];
+            _totalIvVolume = _totalIvVolume + _volumeRemaining;
+        } forEach (_target getVariable [QEGVAR(medical,ivBags), []]);
```
mharis001:
```suggestion
        } forEach (_target getVariable [QEGVAR(medical,ivBags), []]);
        
```

addons/medical_ai/functions/fnc_healingLogic.sqf
```diff
@@ -54,9 +54,25 @@ switch (true) do {
         _treatmentTime = 5;
         _treatmentArgs = [_target, _selection, "FieldDressing"];
     };
+    case (IN_CRDC_ARRST(_target)): {
+        if (EGVAR(medical_treatment,cprSuccessChance) == 0) then {
+            _treatmentEvent = "#WaitForHeartRate";
+            [QEGVAR(medical,CPRSucceeded), _target] call CBA_fnc_localEvent;
```
PabstMirror:
```suggestion
```
I'm not sure about this line, I think if the chance setting is set to 0, then it should fail
The AI should follow the same rules

commy2:
`IN_CRDC_ARRST(_target) && {EGVAR(medical_treatment,cprSuccessChance) > 0}`
as switch condition?

PabstMirror:
I think that makes sense, it should then just use the default of `_treatmentEvent = "#none";`
Death is inevitable, but I can't think of an easy way to make the medic "give up" on the patient

PR #7258, opened 2019-10-30
Common - Network safe claim

PabstMirror:
Usage:
```
["pickup", { 
    systemChat format ["pickup %1", _this]; 
 
["ace_common_claimSafe", [player, x, false, "pickup"]] call cba_fnc_serverEvent;
```

because of network delay, two clients could claim the same object
this allows claims to be made on server, guaranteeing synchronization


commy2:
Very good.

PabstMirror:
This doesn't actually change any usage yet
e.g. Changing cargo over won't be easy, any might not be needed

PR #7265, opened 2019-11-05
Only play repetitive sounds when game is focused

PabstMirror:


jonpas:
Wouldn't `isGamePaused` make more sense here? That would then mimic vanilla behaviour.

There are 3 states: normal running, unfocused and paused. Unfocused can still have simulation running (and thus sounds), paused has simulation stopped. `-noPause` startup parameter controls this.

synixebrett:
Vanilla audio is still muted when unfocused while using `-noPause`

PabstMirror:
The problem only really happens in development when alt-tabbing a lot (I use -noPause)
it seems to queue up sounds like heartbeats and then when I tab back in it plays 100x heartbeats at once and is extremely loud

jonpas:
Well that's just silly, such consistency in the game! /s

The this makes sense.

PR #7268, opened 2019-11-07
Tagging - Group spray paint items in Arsenal

synixebrett:
- "Black spray paint" -> "Spray Paint (Black)"

I was in the arsenal and this annoyed me, slight QOL improvement


jonpas:
Can we go one step further and do `Spray Paint (Black)`? That should follow vanilla naming, which has every word capitalized if I remember correctly.

Issue #6090, opened 2018-01-28
Mass of ACE Medical fluids is lower than real life counterparts.


Drofseh:
**ACE3 Version:** `3.12.1` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
-The mass of ACE Medical fluids doesn't correspond to the actual weigh of these fluids in real life.
Regardless of fluid type, ACE fluids always have 
`mass = 10;` for 1000 ml,
`mass = 5;` for 500 ml, and
`mass = 2.5;` for 250 ml.

These correspond to 1 lb, 0.5 lbs, and 0.25 lbs in the ACE stamina system and Arsenal.
In reality 1000 ml of saline weighs 1.0046 kg, or `mass = 22.15;` in ARMA/ACE3 terms.

Additionally, saline, plasma, and blood should all have distinct weights.
Using the following specific gravities for each fluid:
```
saline  1.0046
plasma  1.02575
blood   1.05635
```
We get the corresponding `mass` values for ARMA
```
        1000 ml   500 ml   250 ml
saline  22.15     11.08   5.54
plasma  22.61     11.31   5.65
blood   23.29     11.65   5.82
```
**Steps to reproduce:**
- Check medical cfgWeapons.hpp starting here:
https://github.com/acemod/ACE3/blob/master/addons/medical/CfgWeapons.hpp#L122
- Or open ACE Arsenal in game, add fluid bags and watch the weight change.

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

**Placed Modules:**
- None

alganthe:
mass in arma is not a representation of weight, we have some approximations, but don't expect objects to be based on real world weights.

Drofseh:
I know mass is an abstract in ARMA terms, but ACE ties it specifically to lbs and kg in its inventory and arsenal displays.

alganthe:
bigger mass means you can stuff less in your backpack / any container, it might not be a good idea to double mass to fit the real weight.

Drofseh:
> it might not be a good idea
agreed, it has a good chance to mess up loadouts.

I decided to make an issue anyways given how dramatically different `mass` is from what I expected it to be,

brainslush:
We changed it for our community to the real values (ACE Inventory weight) but also increased most container sizes. Maybe it could be part of an optional or of the realistic weights component

PR #7272, opened 2019-11-17
Medical - Increase IV Mass

LachlanSneddon:
* increase the mass of all IV's (blood, saline and plasma) have a more realistic weight, based on the function ace_common_fnc_getWeight.


Drofseh:
For reference #6090

PabstMirror:
You risk breaking old loadouts with these changes
We had a similar issue with bandage weights that got reverted

jonpas:
I vote close for the same reason we reverted bandage changes.

jonpas:
No further comments in 10 days. Thanks for the contribution @LachlanSneddon but we are not changing this in main ACE3.

https://github.com/acemod/ACE3/issues/6090#issuecomment-372571553 this possibility of an optional still exists though. :)

PR #7273, opened 2019-11-17
Optics - Convert Optics to use CBA Optics

### LINKED ISSUE #4908: MOS 2D and PIP wrong zoom levels
**CBA Version:** `3.1.2` (dev `dd3f3d0454fd6c65a77aaaa4e209ed0632685f85`)
**ACE3 Version:** `3.9.0` (dev `19d5fa0a4e2606ba5fe12b1c71293a772a028bc6`)

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
`MOS (2D)` and `MOS (PIP)` have LRPS' zoom levels instead of the MOS'.

Screenshots of smallest zoom level.

Vanilla:
mos
ACE3 2D:
mos_2d
ACE3 PIP:
mos_pip

**Steps to reproduce:**
- Use the ACE3 MOS and vanilla MOS scope.

- Editor (Singleplayer)

**Placed Modules:**
- None

**RPT log file:**
- N/A

commy2:
- With all the benefits that come with it, as promised.
- Made the Elcan, Hamr and Pitbull all 4x magnification instead of 2.9x.
- Added continuous zooming levels to the MOS scope.
- Reduced optics blur.
- Fixes #4908.
  - Changed magnification levels of MOS scope to be from 2.75-5.5x.
- ~~Requires https://github.com/CBATeam/CBA_A3/pull/1678.~~ can't work


jonpas:
Would likely need to be redone, closing. Feel free to re-open though.

johnb432:
In a future PR, I might add support for switching between the base scope and the ACE 2D/PIP variants, like CUP does.

addons/optics/script_component.hpp
```diff
@@ -16,17 +16,20 @@
...
+#define PIP(classname,parent)\
+class classname: parent {\
+    author = ECSTRING(common,ACETeam);\
+    scope = 1;\
+    class ItemInfo: ItemInfo {\
+        modelOptics = "\x\cba\addons\optics\cba_optic_big_pip.p3d";\
+    };\
+}
+
+#define PRELOAD\
```
PabstMirror, line 29:
```suggestion
#define PRELOAD \
```

PabstMirror:
small ring of non-pip on the outside of scope



seems like `2d` and `pip` are both using pip now, not sure if intentional?

johnb432:
> small ring of non-pip on the outside of scope

That sucks. I had reduced the size, thinking it would be fine. I'm on a craptop with a 1080p screen, so I expected the scaling not to matter. Out of curiosity, what screen resolution do you have?

> seems like `2d` and `pip` are both using pip now, not sure if intentional?

That's not intentional, I will investigate.

johnb432:
> small ring of non-pip on the outside of scope

Looks like it's something I oversaw on my end - I imagine I'll be able to find a good value that works for everyone.

> seems like `2d` and `pip` are both using pip now, not sure if intentional?

This, on the other hand, worries me because on my end it works fine.

johnb432 (approved):
I've done rigorous testing (checked if all sights could be accessed, if sizes are good, if 2D and PIP work ok, magnification levels are the same as the Vanilla counter parts) and LGTM. Given I've contributed to this PR in a not insignificant manner, it would be wise for someone else to review it before merging.

addons/optics/CfgWeapons.hpp
```diff
@@ -1,110 +1,51 @@
...
-                class Snip: Snip {
-                    opticsZoomMin = 0.05;
-                    opticsZoomMax = 0.3;
-                    opticsZoomInit = 0.3;
-                    discretefov[] = {};
-                    modelOptics[] = {};
-                };
+            reticleTexture = QPATHTOF(reticles\hamr-reticle65_ca.paa);
+            reticleTextureNight = QPATHTOF(reticles\hamr-reticle65Illum_ca.paa);
+            reticleTextureSize = 1 * 0.9;
```
PabstMirror, line 25: before we set it to a higher zoom than normal hamr (changed `opticsZoomMax`) 
(~2.9x instead of 2x from vanilla)
If we are reverting that, then we also need to change the scale of the reticule or the ranging won't match

scopes

this value should be something around `0.53`, 
(can test with `(cba_optics_OpticReticleDetailTextures # 0) set [2, 0.53]` and fire to redraw)
but I'd like to try to find a actual formula for converting old to new
also with the smaller size we'll have to redo the pip hole size

PabstMirror: `.75 * 6.00094 / 8.59571` = 0.523599 
0.75 from https://github.com/acemod/ACE3/blob/master/addons/optics/CfgRscTitles.hpp#L62
and then the change in FOV/zoom

looks pretty close on the windage marks
y

but the problem is this is using `modelOptics = "\x\cba\addons\optics\cba_optic_small_70.p3d";`
and it's still not small enough

I'm not sure, but we might have to add more zooms to cup optics,
or revert the zoom change and go back to the 2.9x?

johnb432: > or revert the zoom change and go back to the 2.9x?

Probably the easier solution, as I have no idea how to create new, smaller CBA optic models.
I imagine I'll have to redo the sizes again and try using different CBA optics. The worry I have is that the PIP optics have 2 models to choose from, which might be limiting.

PabstMirror: I think just setting all the sizes to `0.75` should work if we keep the zoom the same as before
based on
https://github.com/CBATeam/CBA_A3/blob/master/addons/optics/script_component.hpp#L42-L45
https://github.com/acemod/ACE3/blob/master/addons/optics/CfgRscTitles.hpp#L62-L75

johnb432: > I think just setting all the sizes to `0.75` should work if we keep the zoom the same as before based on https://github.com/CBATeam/CBA_A3/blob/master/addons/optics/script_component.hpp#L42-L45 https://github.com/acemod/ACE3/blob/master/addons/optics/CfgRscTitles.hpp#L62-L75

0.75 works fine for the HAMR, but there is a small gap with Elcan and Pitbull scopes on the top and bottom of the screens. Changing it to 0.8 for both makes the gap disappear, but that means we'd need to change the magnification to something higher.

johnb432:
I want to address #4908 in this PR.
The Nightforce seems to be fine with vanilla magnification values when zoomed in (1 large increment = 1 mil). In the first mode, the conversion factor is ~2500 (e.g. person (1.8m) at 500m => 1.8 / 9 "mils" * 2500 = 500m).
The MOS is fine in it's first zoom mode (1 large increment = 10 mils), but it's ~2000 for the conversion factor in its zoomed in mode.

I'm not a fan of having different magnification levels between Vanilla, 2D and PIP scopes, but in the case of the HAMR, Elcan and Pitbull there isn't a choice. Here, however, I'd argue there is and, imo, it works well enough as it is.

If it's not considered good enough of a solution, what would need to be done is to change the magnification levels of the 2D and PIP scopes back to what they were prior to this PR (so slightly different in the case of the Nightforce), but I'd want to find a different pair of magnification levels for the MOS (what I currently have in mind is using the lower magnification of the Nightforce and halving that value to use as the first magnification level, given they use the same reticle).

In numbers:
- MOS vanilla: 2.33x & 5x
- MOS ACE 3.17.1: 5.38x & 21.55x (the same as the Nightforce, which is the problem that 4908 describes)
- MOS new: 2.69x & 5.38x

- Nightforce Vanilla: 5.56x & 22.73x
- Nightforce ACE 3.17.1: 5.38x & 21.55x
- Nightforce new: 5.38x & 21.55x

---

One more thing: The Vanilla Nightforce has two magnification modes, whereas the ACE 3.17.1 has a continuous zoom. Currently this PR reverts that behaviour (so it's like the Vanilla one), but I can add it back. The question is: If I add it back, do we also want to add continuous zoom to the MOS?

johnb432:
I overhauled the magnification levels of all scripted optics: the HAMR, Elcan and Pitbull have been bumped up to 4x, as to make the mildot/ranging match in-game bullet drop-off. The HAMR isn't quite perfect, as it was either the mildot or the bullet drop-off.
Bullet drop-off was measured using the 6.5mm MX cartridge.
An added benefit of this change is that the scopes have the same magnification levels as their real-life counterparts.
The MOS and Nightforce scopes were adjusted to 2.75-5.5x and 5.5-22x respectively. I based myself off of the real-life Nightforce and adjusted the in-game version until something matched: now, when fully zoomed in, every wide line represents 1 mil.
For the MOS, I just took the lower magnification level of the Nightforce and used halved it to get the minimum. If at a magnification level of 5.5x, you need to multiply the mils by 4 to get the proper value.

addons/optics/CfgWeapons.hpp
```diff
@@ -115,95 +65,100 @@ class CfgWeapons {
...
+        scope = 2;
         displayName = CSTRING(arco);
-        weaponInfoType = "ACE_RscWeapon_Arco";
+
+        class CBA_ScriptedOptic {
+            opticsPPEffects[] = {"ACE_OpticsRadBlur1"};
+
+            reticleTexture = QPATHTOF(reticles\arco-reticle65_ca.paa);
+            reticleTextureNight = QPATHTOF(reticles\arco-reticle65Illum_ca.paa);
+            reticleTextureSize = 0.72;
```
PabstMirror, line 77:
```suggestion
            reticleTextureSize = 0.75 * 0.0872664626 / (0.25 / 4);
```

addons/optics/CfgWeapons.hpp
```diff
@@ -115,95 +65,100 @@ class CfgWeapons {
...
-        weaponInfoType = "ACE_RscWeapon_MRCO";
+        scope = 2;
+        displayName = CSTRING(mrco);
+
+        class CBA_ScriptedOptic {
+            opticsPPEffects[] = {"ACE_OpticsRadBlur1"};
+
+            reticleTexture = QPATHTOF(reticles\mrco-reticle556_ca.paa);
+            reticleTextureNight = QPATHTOF(reticles\mrco-reticle556Illum_ca.paa);
+            reticleTextureSize = 0.72;
```
PabstMirror, line 129:
```suggestion
            reticleTextureSize = 0.75 * 0.0872664626 / (0.25 / 4);
```

addons/optics/CfgWeapons.hpp
```diff
@@ -1,110 +1,60 @@
...
-                class Snip: Snip {
-                    opticsZoomMin = 0.05;
-                    opticsZoomMax = 0.3;
-                    opticsZoomInit = 0.3;
-                    discretefov[] = {};
-                    modelOptics[] = {};
-                };
+            reticleTexture = QPATHTOF(reticles\hamr-reticle65_ca.paa);
+            reticleTextureNight = QPATHTOF(reticles\hamr-reticle65Illum_ca.paa);
+            reticleTextureSize = 0.9;
```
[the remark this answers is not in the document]
PabstMirror: original:

now:


zeroing is still off by about 50%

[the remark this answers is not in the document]
johnb432: Where does `0.0872664626` come from? Doing it by hand at 25x (have the mildots match at 25x), then scaling it down to 4x gives me a total factor of 1.024, whereas you have 1.0472.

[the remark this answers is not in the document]
PabstMirror: old zoom `opticsZoomMin = 0.0872664626;` -> `opticsZoomMin = 0.25 / 4;`
and 0.75 comes from difference from old texture scaling

the old way may have been wrong, so 1.024 could be correct

PabstMirror (approved):

addons/optics/CfgWeapons.hpp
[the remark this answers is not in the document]
PabstMirror: final:

johnb432:
One last thing: As it stands, I've removed the `PIP` stringtables. However, `realisticnames` still renamed the PIP scopes.
Do I remove the renaming from `realisticnames` or do I restore the PIP stringtable entries?

PR #7276, opened 2019-11-20
Arsenal - Remove unused VRTraining class from Arsenal config

dedmen:
Ref: https://forums.bohemia.net/forums/topic/193161-ace3-a-collaborative-merger-between-agm-cse-and-ace/?do=findComment&comment=3387061

apparently doesn't exist with contact loaded, and we don't use it anyway.

### DIFF
--- a/addons/arsenal/RscDisplayMain.hpp
+++ b/addons/arsenal/RscDisplayMain.hpp
@@ -9,7 +9,6 @@ class RscDisplayMain: RscStandardDisplay {
 
             class Controls: Controls {
                 class Bootcamp;
-                class VRTraining;
                 class Arsenal;
                 class GVAR(mission): Arsenal {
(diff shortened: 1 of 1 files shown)


commy2:
Can you still Tab key through the controls in the right order?

dedmen:
> Can you still Tab key through the controls in the right order?
It was just a unused reference, and a one causing a error at that.
A proper config binarizer like Mikeros should even remove that useless reference 🤔

commy2:
That doesn't answer my question. Also, learn to quote on Github, you're a prefessional computer person.

jonpas:
> A proper config binarizer like Mikeros should even remove that useless reference

Hide the error instead of telling the user to fix it? Doesn't sound like Mikero to me.

PR #7280, opened 2019-11-23
Arsenal - Fix default loadout public func not checking base items

### LINKED ISSUE #7275: Arsenal Default Loadouts doesn't support items hidden from arsenal
- **Arma 3:** `1.96 stable`
- **CBA:** `3.12.2 stable`
- **ACE3:** `3.12.6 stable`
Mods downloaded from Steam Workshop and up-to-date.
No other mods were used for the test-case.

**Description:**
When creating a Default Loadout based on an unit with weapons or backpacks which are not visible in the Arsenal (no `scope = 2`, or using `baseWeapon = "other_class"`), the items will not show up in the loadout list and won't be given to the player when selected. Or is inconsistent in results (eg. some backpacks with contents do work, some don't. Weapons never seem to show up).

**Steps to reproduce:**
- In 3DEN place a Repair Specialist (B_soldier_repair_F) and a random container with ACE Arsenal:
```
// init
[this, true] call ace_arsenal_fnc_initBox;
```
- Load the mission (SP or MP).
- Check the current output of `getUnitLoadout` on the player:
```
// debug console
_loadout = (getUnitLoadout player);
_loadout
```
- You will see that the unit has weapon `arifle_MX_Holo_pointer_F` with 2 attachments (`acc_pointer_IR` and `optic_Holosight`).
- Add the current loadout of the player to the Default Loadouts:
```
// debug console
_loadout = (getUnitLoadout player);
["<TEST>", _loadout] call ace_arsenal_fnc_addDefaultLoadout;
```
- Open ACE Arsenal on the container and go to Loadouts >> Default Loadouts.
- You wil see a loadout called `<TEST>`, but does *not* have a primary weapon.
- Select the loadout, give it to yourself and exit ACE Arsenal.
- You will now see that the unit does not have a primary weapon, the rest of the items are present.
- Check the new output of `getUnitLoadout` on the player:
```
// debug console
_loadout = (getUnitLoadout player);
_loadout
```
- You will now see that the primary weapon. the attachments and loaded magazine are gone from the array. All other items are present.

- The log file does show the following message when opening the Default Loadouts tab in ACE Arsenal:
```
16:27:54 ACE_Arsenal - Loadout:
    Name: <TEST>
    Missing items: []
    Unavailable items: ["arifle_MX_Holo_pointer_F"]
```

**Expected behavior:**
I expect that the full loadout which is saved is also available and loaded correctly on the player, unless the items do not exist in the configs (eg. modded content while mod is disabled).
It should ignore visibility and scope (unless `scope = 0`), or find the parent (eg, with `BIS_fnc_baseWeapon`) and retrieve the correct item, attachments and contents.

- Editor (Singleplayer) / Editor (Multiplayer)
- not tested in other environments, but should be everywhere

**Log Files:**
https://pastebin.com/rC6dh0V7

**Additional context:**

**Screenshots:**

alganthe:
- Fix default loadouts public function not checking for base items.

close #7275

### DIFF
--- a/addons/arsenal/functions/fnc_addDefaultLoadout.sqf
+++ b/addons/arsenal/functions/fnc_addDefaultLoadout.sqf
@@ -23,6 +23,81 @@ if (isNil QGVAR(defaultLoadoutsList)) then {
     GVAR(defaultLoadoutsList) = [];
 };
 
+for "_dataIndex" from 0 to 10 do {
+    switch (_dataIndex) do {
+
+        case 0;
+        case 1;
+        case 2;
+        case 8: {
+            if (count (_loadout select _dataIndex) > 0) then {
+
+                private _weapon = (_loadout select _dataIndex) select 0;
+                if (_weapon != "") then {
+
+                    private _baseWeapon = _weapon call BIS_fnc_baseWeapon;
+                     if (_weapon != _baseWeapon) then {
+                        (_loadout select _dataIndex) set [0, _baseWeapon];
+                    };
+                };
+            };
+        };
+
+        case 3;
+        case 4;
+        case 5: {
+            if (count (_loadout select _dataIndex) > 0) then {
+                private _containerContents = (_loadout select _dataIndex) select 1;
+
+                if (count _containerContents > 0) then {
+
+                    {
+                        if (count _x == 2) then {
+
+                                if ((_x select 0) isEqualType "") then {
+
+                                    private _item = (_x select 0);
+                                    if (_item != "") then {
+
+                                        private _uniqueBaseCfgText = getText (configFile >> "CfgWeapons" >> _item >> "ace_arsenal_uniqueBase");
+                                        if (_uniqueBaseCfgText != "") then {
+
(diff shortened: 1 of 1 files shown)


commy2:
Uhm, you high?
```sqf
for "_dataIndex" from 0 to 10 do {
    switch (_dataIndex) do {
        case 0;
```

alganthe:
???

PabstMirror:
```
case a;
case b: { code };
```
means that a or b trigger that code (it's like an OR)

commy2:
My point is that
```sqf
for "_i" from 0 to 10 do {
    switch (_i) do {
        case 1: {call A};
        case 2: {call B};
    };
};
```
is the same as:
```sqf
call A;
call B;
```

bux:
Agree. The for and switch are superfluous.

PabstMirror:
```
case 0;
case 1;
case 2;
case 8: {
```
to
```
{
} forEach [0,1,2,8]
```
I'm not sure that's noticeably better

alganthe:
It's not and is less readable, which is why I decided against it in the first place.

PR #7287, opened 2019-11-28
Add tchinese translation,for almost every stringtable i can find

dfr1238:
i check every folder,seem didn't miss one,if miss comment below please
-add missing chinese string
-fix some werid string(not too much)
-make most unit(like mm,m)to chinese


commy2:
There is this issue which lists all components with missing strings __on master branch__: https://github.com/acemod/ACE3/issues/367

If your PR is complete, then it should show 100% after merge.

dfr1238:
yep,seem all string was translate now

commy2:
Now it will take a while for the bot to pick up the changes. Then the pinned translation issue should update.

PR #7289, opened 2019-11-29
Move readme to doc folder and use versioning system

PabstMirror:
putting it next to pl and de readmes
I guess we could move them to base folder, but it should just be consistent


jonpas:
Has to be added here for versioning to work at all on it:
https://github.com/acemod/ACE3/blob/master/tools/make.py#L76

And here for armake: https://github.com/acemod/ACE3/blob/master/Makefile#L6

PR #6900, opened 2019-03-27
Adapt to HEMTT (CI)

jonpas:
**When merged this pull request will:**
- Use HEMTT as build system (uses armake2)
- ~~Add scripts (`setup.bat` and `setup.sh`) to download project-specific HEMTT binary~~ -> #7292
- Add `build.bat` for quick Windows builds without command line
- Remove armake(1) tooling
- ~~Work around preprocessor define and comment on same line (https://github.com/KoffeinFlummi/armake2/issues/24)~~
- Use on GitHub Actions (replace armake which fails since `actions/checkout@2.0.0`)

**TODO ACE3:**
- [x] Use on ~~CircleCI~~ GitHub Actions
- [x] Git commit hash (`version` header extension and signs)
- [x] Zip name (when archiving in HEMTT)
  - [x] Change to short version

**TODO HEMTT:**
- [x] Git commit hash utility (templating) (https://github.com/synixebrett/HEMTT/pull/74)
- [x] Copy files wildcards/glob pattern (https://github.com/synixebrett/HEMTT/issues/68 -> https://github.com/synixebrett/HEMTT/pull/85)
- [x] Copy files from sub-folders (https://github.com/synixebrett/HEMTT/issues/69 -> https://github.com/synixebrett/HEMTT/pull/85)
- [x] Archiving (zip) (https://github.com/synixebrett/HEMTT/pull/9)
- [x] Expanded version templating (https://github.com/synixebrett/HEMTT/issues/80)
- ~~Fix update on Windows global install (https://github.com/synixebrett/HEMTT/issues/53#issuecomment-478237110)~~ (not really required)
- [x] Launchable optional mod folders (https://github.com/synixebrett/HEMTT/pull/132, https://github.com/synixebrett/HEMTT/pull/225)

**TODO armake2:**
- [x] Fix subconfig not processed correctly (https://github.com/KoffeinFlummi/armake2/issues/45)
- [x] Fix preprocessor define and comment on same line (https://github.com/KoffeinFlummi/armake2/issues/24)
  - [x] Remove temporary removal of comments

PR #7292, opened 2019-12-03
Adapt to HEMTT (User)

jonpas:
_Continuation of https://github.com/acemod/ACE3/pull/6900 for general use._

- Add scripts (`setup.bat` and `setup.sh`) to download project-specific HEMTT binary

**TODO ACE3:**
- [ ] Add release/publish procedure (ref. https://github.com/CBATeam/CBA_A3/pull/730#issuecomment-411018642) (HEMTT script)
- [ ] Documentation (reuse #6362)
- [ ] Use Windows Installer instead of project-specific binary and download script (https://github.com/synixebrett/HEMTT/pull/60) - requires update compatibility fix
- [ ] Switch to HEMTT release builds (final)

**TODO HEMTT:**
- [ ] Release proper 0.7.x (restructure)
- [ ] Update versions in files (like `make.py` or `Makefile`) (https://github.com/synixebrett/HEMTT/issues/16)

### DIFF
--- a/tools/setup.sh
+++ b/tools/setup.sh
@@ -1,0 +1,12 @@
+VERSION=0.6.1
+ARCH=$(uname -m)
+
+echo "=> Downloading HEMTT..."
+wget -q --show-progress https://github.com/synixebrett/HEMTT/releases/download/v$VERSION/hemtt-v$VERSION-$ARCH-unknown-linux-gnu.tar.gz
+
+echo "=> Extracting HEMTT..."
+tar -xzf hemtt-v$VERSION-$ARCH-unknown-linux-gnu.tar.gz -C ../.
+rm hemtt-v$VERSION-$ARCH-unknown-linux-gnu.tar.gz
+
+echo "=> Updating HEMTT..."
+../hemtt update
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ release/*
 releases/*
 hemtt
 hemtt.exe
+tools/hemtt*.tar.gz*
+tools/hemtt*.zip*
 tools/temp
 *.cache
 *.pbo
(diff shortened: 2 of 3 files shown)


tools/setup.sh
```diff
@@ -0,0 +1,12 @@
+VERSION=0.6.1
+ARCH=$(uname -m)
+
+echo "=> Downloading HEMTT..."
+wget -q --show-progress https://github.com/synixebrett/HEMTT/releases/download/v$VERSION/hemtt-v$VERSION-$ARCH-unknown-linux-gnu.tar.gz
```
Dahlgren: The script will continue even if this fails. Add `set -e` somewhere before to exit the script if any command fails.

tools/setup.sh
```diff
@@ -0,0 +1,12 @@
...
+
+echo "=> Extracting HEMTT..."
+tar -xzf hemtt-v$VERSION-$ARCH-unknown-linux-gnu.tar.gz -C ../.
+rm hemtt-v$VERSION-$ARCH-unknown-linux-gnu.tar.gz
+
+echo "=> Updating HEMTT..."
+../hemtt update
```
Dahlgren: Should it really auto update? Future changes in HEMTT might not work with the current project setup such as hemtt.toml.

jonpas: I am not sure if I'll keep those scripts around at all yet.

dedmen: A npm like update scheme might be good here? only update patches, but not minor?

Dahlgren: I think it's better to use a known working version and then update the HEMTT link once a newer version have been tested successfully.

BrettMayson:
Can this be closed with HEMTT 1.0.0 support already on master?

jonpas:
Yeah, although we should still update documentation and guide users to use HEMTT in the future.

PR #7307, opened 2019-12-15
Add Wipe glasses self interaction

veteran29:
- Wipe glasses self interaction
- Toggleable via settings (default: false)

### DIFF
--- a/addons/goggles/functions/fnc_canWipeGlasses.sqf
+++ b/addons/goggles/functions/fnc_canWipeGlasses.sqf
@@ -15,4 +15,4 @@
  * Public: No
  */
 
-GVAR(effects) != 2 && {!GETVAR(ACE_player,ACE_isUnconscious,false)} // return
+GVAR(effects) == 2 && {!GETVAR(ACE_player,ACE_isUnconscious,false)} // return
--- a/addons/goggles/functions/fnc_canWipeGlasses.sqf
+++ b/addons/goggles/functions/fnc_canWipeGlasses.sqf
@@ -4,10 +4,10 @@
  * Checks if player can wipe goggles.
  *
  * Arguments:
- * 0: Player <OBJECT>
+ * None
  *
  * Return Value:
- * Can goggles be wiped <BOOL>
+ * Can wipe goggles <BOOL>
  *
  * Example:
  * [] call ace_goggles_fnc_canWipeGlasses
@@ -15,7 +15,4 @@
  * Public: No
  */
 
-// Can only wipe if full effects setting is set
-if (GVAR(effects) != 2) exitWith {false};
-
-!GETVAR(ACE_player,ACE_isUnconscious,false) // return
+GVAR(effects) != 2 && {!GETVAR(ACE_player,ACE_isUnconscious,false)} // return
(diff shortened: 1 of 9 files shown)


veteran29:
There is one issue right now, despite having `exceptions[] = {"notOnMap"};` in config the interaction is visible on the map screen (wanted to hide it there to not clutter the map interaction).

mharis001:
It works the other way around, you don't want "notOnMap" to be an exception for the `canInteractWith` check.

jonpas:
Exceptions are double-negated. In this case "not not on map", resulting in "on map", meaning also show on map. Just remove it as mharis said.

addons/goggles/functions/fnc_canWipeGlasses.sqf
```diff
@@ -0,0 +1,21 @@
+#include "script_component.hpp"
+/*
+ * Author: veteran29
+ * Checks if player can wipe goggles.
+ *
+ * Arguments:
+ * 0: Player <OBJECT>
```
mharis001: Should be "None".

addons/goggles/functions/fnc_canWipeGlasses.sqf
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Return Value:
+ * Can goggles be wiped <BOOL>
```
mharis001: "Can wipe goggles" sounds better IMO.

addons/goggles/functions/fnc_canWipeGlasses.sqf
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Example:
+ * [] call ace_goggles_fnc_canWipeGlasses
+ *
+ * Public: No
+ */
+
+// Can only wipe if full effects setting is set
+if (GVAR(effects) != 2) exitWith {false};
```
mharis001: Combine these two conditions: `GVAR(effects) != 2 && {!(ACE_player getVariable ["ACE_isUnconscious", false])}`.

addons/goggles/initSettings.sqf
```diff
@@ -0,0 +1,8 @@
+[
+    QGVAR(showClearGlasses),
+    "CHECKBOX",
+    [LSTRING(SettingShowClearGlasses), LSTRING(SettingShowClearGlasses)],
+    localize LSTRING(SettingsName),
+    false, // default value
+    2 // isGlobal
```
mharis001: Would `0` be better since it will still be client settable but can also still be forced by mission/server.

veteran29: Thanks for review, will apply suggestions soon. It would be cool if you would use GH suggestion feature in future, makes life easier 😜

addons/goggles/functions/fnc_canWipeGlasses.sqf
```diff
@@ -0,0 +1,18 @@
...
+ * Return Value:
+ * Can wipe goggles <BOOL>
+ *
+ * Example:
+ * [] call ace_goggles_fnc_canWipeGlasses
+ *
+ * Public: No
+ */
+
+GVAR(effects) != 2 && {!GETVAR(ACE_player,ACE_isUnconscious,false)} // return
```
PabstMirror:
```suggestion
GVAR(effects) == 2 && {!GETVAR(ACE_player,ACE_isUnconscious,false)} // return
```

PR #7319, opened 2019-12-20
Addition to Czech translation

PointForward:
- Add missing Czech translation strings
- There should be no untranslated string anymore in tabler (for now)
- Part of https://github.com/acemod/ACE3/issues/367

Tried to make it as understandable as possible, sometimes taking very slight creative license due to the Eng-Cze differences.
Also tried to follow examples set by current translations to appear more uniform.


PabstMirror:
I will get merge conflicts sorted

PR #7321, opened 2019-12-21
French translation of the latest updates

Elgin675:
- Upgrade the French translation, by translating the keys added during the last updates.

### DIFF
--- a/addons/goggles/stringtable.xml
+++ b/addons/goggles/stringtable.xml
@@ -111,9 +111,5 @@
             <Chinese>在自我互動中顯示擦拭護目鏡的動作</Chinese>
             <French>Afficher l'interaction "Essuyer les lunettes"</French>
         </Key>
-        <Key ID="STR_ACE_Goggles_SettingShowClearGlassesDesc">
-            <English>Show Wipe Goggles self interaction</English>
-            <French>Affiche l'action "Essuyer les lunettes" dans le menu d'interaction personnel.</French>
-        </Key>
     </Package>
 </Project>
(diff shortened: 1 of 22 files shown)


addons/goggles/stringtable.xml
```diff
@@ -109,6 +109,11 @@
             <Czech>Ukaž Vyčistit brýle v menu Interakce (vlastní)</Czech>
             <Japanese>ゴーグルの拭き取りをセルフ インタラクションに表示</Japanese>
             <Chinese>在自我互動中顯示擦拭護目鏡的動作</Chinese>
+            <French>Afficher l'interaction "Essuyer les lunettes"</French>
+        </Key>
+        <Key ID="STR_ACE_Goggles_SettingShowClearGlassesDesc">
```
PabstMirror: <Key ID="STR_ACE_Goggles_SettingShowClearGlasses">
            <English>Show Wipe Goggles self interaction</English>
            <French>Afficher l'interaction "Essuyer les lunettes"</French>

        <Key ID="STR_ACE_Goggles_SettingShowClearGlassesDesc">
            <English>Show Wipe Goggles self interaction</English>
            <French>Affiche l'action "Essuyer les lunettes" dans le menu d'interaction personnel.</French>


I feel like many translations are based on english, so it feels weird to have 2 identical phrases.
I think the reason done here is that "self interaction" doesn't fit in setting display width.
Is there any way to write the first one differently so that it does fit?

Otherwise I'd say we could add a generic string to common for
"Adds action to self interaction menu"
and use that in a few places.
If that sounds ok, I can do the work on this PR to add it

PabstMirror: I pushed these changes
```
        <Key ID="STR_ACE_Common_showActionInSelfInteraction">
            <English>Show the action in the self-interaction menu</English>
            <French>Affiche l'action dans le menu d'interaction personnel.</French>
```

Elgin675: @PabstMirror : Sorry for the late response, but I was out of town for a few days.
Your idea is very good :) !
I totally agree ;) !

PR #7330, opened 2019-12-26
Arsenal - Register feature camera

mharis001:
- title


PabstMirror:
maybe do this in fnc_onArsenalOpen.sqf

PabstMirror:
CBA only checks every 0.5 seconds, so it's not that important,
but we could move this and some of the other cams to on-use instead of always
I might try to do that in another PR

PR #7332, opened 2019-12-26
Stringtable Validator - Check for tabs and correct spacing

PabstMirror:


tools/stringtable_validator.py
```diff
@@ -115,6 +115,22 @@ def check_stringtable(filepath):
...
 
+    # Check whitespace for tabs/spacing
+    with open(filepath, 'r', encoding="latin-1") as file:
+       spacingDepth = 0
+       for line in file:
+           if ("\t" in line):
+               print("ERROR: Tabs!")
+               errors += 1
+           lineClean = line.lstrip().lower()
+           if (lineClean.startswith("</key") or lineClean.startswith("</package") or lineClean.startswith("</project") or lineClean.startswith("</container")):
```
jonpas:
```suggestion
           if lineClean.startswith("</key") or lineClean.startswith("</package") or lineClean.startswith("</project") or lineClean.startswith("</container"):
```

tools/stringtable_validator.py
```diff
@@ -115,6 +115,22 @@ def check_stringtable(filepath):
...
+               spacingDepth -= 4
+           if ((len(line) - len(lineClean)) != spacingDepth):
```
jonpas:
```suggestion
           if len(line) - len(lineClean) != spacingDepth:
```

tools/stringtable_validator.py
```diff
@@ -115,6 +115,22 @@ def check_stringtable(filepath):
...
+               print("ERROR: Spacing not correct for {}".format(line))
+               errors += 1
+           if (lineClean.startswith("<key") or lineClean.startswith("<package") or lineClean.startswith("<project") or lineClean.startswith("<container")):
```
jonpas:
```suggestion
           if lineClean.startswith("<key") or lineClean.startswith("<package") or lineClean.startswith("<project") or lineClean.startswith("<container"):
```

tools/stringtable_validator.py
```diff
@@ -115,6 +115,29 @@ def check_stringtable(filepath):
                 print("  ERROR: Key '{}' is defined {} times.".format(id, count))
                 errors += 1
 
+    # Check whitespace for tabs and correct number of indenting spaces
+    with open(filepath, "r", encoding = "latin-1") as file:
```
mharis001: Should this be `utf-8`?

jonpas: Yes, `utf-8` should handle all languages just fine. Unicode (Python default) will however not.

PabstMirror: I think `utf-8` had problems with `utf-8 bom`, or something like that. and latin just worked
but we enforce correct bom so it shouldn't be an issue

PR #7335, opened 2019-12-27
CSW - Update 3den config for 1.98 changes

PabstMirror:
not even sure if this will make it into final 1.98, but on dev branch now
(making a PR so I don't forget)
```
Updating base class Controls->, by z\ace\addons\csw\config.cpp/Cfg3DEN/Attributes/Combo/Controls/ (original (a3\3den\config.bin - no unload))
```


PabstMirror:
Still needed for final 1.98

PR #7337, opened 2019-12-27
Add AGM-114L Active Radar Homing Hellfire

TheCandianVendingMachine:
- Implement Active Radar Homing seeker type
- Add the AGM-114L Hellfire

The AGM-114L is an active-radar guided missile. It is primarily fire and forget, but the launcher can keep a lock if they want to give more accurate target data for the missiles flight. The guidance model uses the vanilla ARMA radar system to select the target, but once you fire the new implementation takes over. If there is datalinked targets the donor will "give" target data to the Hellfire instead of the shooter. Although the missile will be able to track, once it goes active it can still lose track due to LOS loss even if the shooter is still locked on. Because of this its most useful in open fields instead of forests or towns.

In action
https://youtu.be/yeZ1yMzPXH4
https://youtu.be/hvWmyw7GD6U
https://youtu.be/O8G3_iZyWGY
https://youtu.be/0yGIX_HPxPk


addons/hellfire/CfgWeapons.hpp
```diff
@@ -29,4 +30,26 @@ class CfgWeapons {
         magazines[] = {"6Rnd_ACE_Hellfire_AGM114N", "PylonMissile_1Rnd_ACE_Hellfire_AGM114N", "PylonRack_1Rnd_ACE_Hellfire_AGM114N", "PylonRack_3Rnd_ACE_Hellfire_AGM114N", "PylonRack_4Rnd_ACE_Hellfire_AGM114N"};
         magazineWell[] += {QGVAR(N)};
     };
+    class GVAR(launcher_L): GVAR(launcher) {
+        displayName = "AGM-114L Hellfire ""Longbow""";
+        magazines[] = {"6Rnd_ACE_Hellfire_AGM114L", "PylonMissile_1Rnd_ACE_Hellfire_AGM114L", "PylonRack_1Rnd_ACE_Hellfire_AGM114L", "PylonRack_3Rnd_ACE_Hellfire_AGM114L", "PylonRack_4Rnd_ACE_Hellfire_AGM114L"};
+        magazineWell[] += {QGVAR(L)};
+        EGVAR(laser,showHud) = 1;
```
PabstMirror:
```suggestion
        EGVAR(laser,showHud) = 1; // Just to show the attack profile
```
I think this is why; maybe something we could look at in the future

addons/missileguidance/functions/fnc_ahr_onFired.sqf
```diff
@@ -0,0 +1,71 @@
...
+    _timeToActive = 0 max (_distanceUntilActive / _velocityAtImpact);
+};
+
+if (isNil "_target") then {
+    _timeToActive = 0;
+    _isActive = true;
+    _target = objNull;
+};
+
+private _shooterHasActiveRadar = {
```
PabstMirror: seems like data link must be enabled, so we could also check
`vehicleReportRemoteTargets vehicle _shooter` as well

addons/missileguidance/functions/fnc_seekerType_ARH.sqf
```diff
@@ -0,0 +1,104 @@
...
+            };
+        } forEach _nearestObjects;
+    };
+
+    _projectile setMissileTarget _target;
+} else {
+    // External radar homing
+    // if the target is in the remote targets for the side, whoever the donor is will "datalink" the target for the hellfire.
+    private _remoteTargets = listRemoteTargets side _shooter;
+    if ((_remoteTargets findIf { (_target in _x) && (_x#1 > 0) }) < 0) then {
```
PabstMirror: I think if target is not in list it should go active immediately?

addons/missileguidance/functions/fnc_ahr_onFired.sqf
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Example:
+ * [] call ace_missileguidance_fnc_SACLOS_onFired
+ *
+ * Public: No
+ */
+params ["_firedEH", "_launchParams", "", "", "_stateParams"];
+_firedEH params ["_shooter","","","","","","_projectile"];
+_stateParams params ["", "_seekerStateParams"];
+_launchParams params ["_target"];
```
PabstMirror: you are consistent, so it's not a problem but by default missile guidance puts shooter instead of target on first param
```
_launchParams params ["_shooter","_targetLaunchParams","_seekerType","_attackProfile","_lockMode","_laserInfo"];
_targetLaunchParams params ["_target", "_targetPos", "_launchPos"];
```

addons/missileguidance/functions/fnc_ahr_onFired.sqf
```diff
@@ -0,0 +1,71 @@
...
+ * Sets up SACLOS state arrays (called from missileGuidance's onFired).
+ *
+ * Arguments:
+ * Guidance Arg Array <ARRAY>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call ace_missileguidance_fnc_SACLOS_onFired
```
PabstMirror:
```suggestion
 * [] call ace_missileguidance_fnc_ahr_onFired
```

addons/missileguidance/functions/fnc_ahr_onFired.sqf
```diff
@@ -0,0 +1,71 @@
+#include "script_component.hpp"
+/*
+ * Author: Brandon (TCVM)
+ * Sets up SACLOS state arrays (called from missileGuidance's onFired).
```
PabstMirror:
```suggestion
 * Sets up AHR state arrays (called from missileGuidance's onFired).
```

addons/missileguidance/functions/fnc_seekerType_ARH.sqf
```diff
@@ -0,0 +1,104 @@
...
+        private _a1 = abs _projPitch;
+        private _a2 = 180 - ((_seekerAngle / 2) + _a1);
+        private _seekerBaseRadiusAtGround = ACTIVE_RADAR_MINIMUM_SCAN_AREA max (_distanceToExpectedTarget / sin(_a2) * sin(_seekerAngle / 2));
+        private _seekerBaseRadiusAdjusted = linearConversion [0, _seekerBaseRadiusAtGround, (CBA_missionTime - _lastTimeSeen) * vectorMagnitude _lastKnownVelocity, ACTIVE_RADAR_MINIMUM_SCAN_AREA, _seekerBaseRadiusAtGround, false];
+        // Look in front of seeker for any targets
+        private _nearestObjects = nearestObjects [_expectedTargetPos, ["AllVehicles"], _seekerBaseRadiusAdjusted, false];
+
+        _nearestObjects = _nearestObjects apply {
+            // I check both Line of Sight versions to make sure that a single bush doesnt make the target lock dissapear but at the same time ensure that this can see through smoke. Should work 80% of the time
+            if ([_projectile, getPosASL _x, _seekerAngle] call ace_missileguidance_fnc_checkSeekerAngle && { ([_projectile, _x, true] call ace_missileguidance_fnc_checkLOS) || { ([_projectile, _x, false] call ace_missileguidance_fnc_checkLOS) } }) then {
```
PabstMirror: FUNC(

addons/missileguidance/functions/fnc_seekerType_ARH.sqf
```diff
@@ -0,0 +1,104 @@
...
+        {
+            if ((_x distance2d _expectedTargetPos) < _closestDistance) then {
+                _closestDistance = _x distance2d _expectedTargetPos;
+                _target = _x;
+            };
+        } forEach _nearestObjects;
+    };
+
+    _projectile setMissileTarget _target;
+} else {
```
PabstMirror: only if you want, it's a little hackey, but helps with the visual debugging
```suggestion
} else {
    #ifdef DRAW_GUIDANCE_INFO
    _seekerTypeName = "AHR - EXT";
    #endif
```

addons/missileguidance/functions/fnc_seekerType_ARH.sqf
TheCandianVendingMachine: Sadly due to the way `listRemoteTargets` works the vehicle needs to have datalink enabled for them to appear in this list

addons/missileguidance/functions/fnc_ahr_onFired.sqf
TheCandianVendingMachine: I didn't do that because the datalink to a missile is different to the datalink to aircraft in real life so you don't need to enable datalink to guide a missile

addons/hellfire/CfgWeapons.hpp
TheCandianVendingMachine: Yeah that's why. I was thinking of ripping out the missileguidance stuff from laser but I decided against it for this PR

addons/missileguidance/functions/fnc_seekerType_ARH.sqf
PabstMirror: shoot and then turn radar off
- With data link enabled: missile immediately goes active
- With data link disabled: `listRemoteTargets is always []` so nothing happens until time hits _timeWhenActive

PabstMirror: ops, this is fine I was thinking it worked differently

addons/missileguidance/functions/fnc_seekerType_ARH.sqf
```diff
@@ -0,0 +1,119 @@
...
+    _seekerTypeName = "AHR - EXT";
+    #endif
+    // External radar homing
+    // if the target is in the remote targets for the side, whoever the donor is will "datalink" the target for the hellfire.
+    private _remoteTargets = listRemoteTargets side _shooter;
+    if ((_remoteTargets findIf { (_target in _x) && (_x#1 > 0) }) < 0) then {
+        // I check both Line of Sight versions to make sure that a single bush doesnt make the target lock dissapear but at the same time ensure that this can see through smoke. Should work 80% of the time
+        if (!_shooterHasRadar || { !isVehicleRadarOn vehicle _shooter } || { !alive vehicle _shooter } || { !([vehicle _shooter, _target, true] call FUNC(checkLOS)) && { !([vehicle _shooter, _target, false] call FUNC(checkLOS)) } }) exitWith {
+            _seekerStateParams set [0, true];
+            _expectedTargetPos
```
PabstMirror: this exitWith return don't seem to be used

optionals/compat_rhs_usf3/CfgMagazineWells.hpp
```diff
@@ -5,4 +5,7 @@ class CfgMagazineWells {
     class ace_hellfire_N {
         ADDON[] = {GVAR(pylon_mag_2rnd_hellfire_n)};
```
PabstMirror:
```suggestion
        ADDON[] = {QGVAR(pylon_mag_2rnd_hellfire_n)};
```

optionals/compat_rhs_usf3/CfgMagazineWells.hpp
```diff
@@ -5,4 +5,7 @@ class CfgMagazineWells {
...
     };
+    class ace_hellfire_L {
+        ADDON[] = {GVAR(pylon_mag_2rnd_hellfire_L)};
```
PabstMirror:
```suggestion
        ADDON[] = {QGVAR(pylon_mag_2rnd_hellfire_L)};
```

PR #7338, opened 2019-12-28
Repair - Handle turret hitpoints with no selection

### LINKED ISSUE #7245: RHS Compat. Stryker doesn't have repair points for turret and optics.
- **Arma 3:** `1.94` stable
- **CBA:** `3.12.2` stable
- **ACE3:** `3.12.6` stable

**Description:**
RHS Strykers does not have some repair points. For example turret and optics.
Maybe something else.

**Steps to reproduce:**
- Place `rhsusf_stryker_m1126_m2_wd`
- Place engineer with repair kit
- `stryker setHitPointDamage ["hitturret",1];`
- Try to find repair point

- Try `[] execVM "\z\ace\addons\repair\dev\draw_showRepairInfo.sqf";`

PabstMirror:
Fix #7245
see notes in https://github.com/acemod/ACE3/issues/7245#issuecomment-569429941

### DIFF
--- a/addons/repair/functions/fnc_addRepairActions.sqf
+++ b/addons/repair/functions/fnc_addRepairActions.sqf
@@ -37,10 +37,13 @@ private _hitPointsAddedAmount = [];
 private _processedSelections = [];
 private _icon = ["a3\ui_f\data\igui\cfg\actions\repair_ca.paa", "#FFFFFF"];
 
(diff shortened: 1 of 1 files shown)


PabstMirror:
This is the best solution I have right now; 
I don't think this will effect vanilla or CUP stuff (mainly just RHS?)
I tested as much as I could and it will log any use with an `INFO` so it should be easier to track down problems

PR #7360, opened 2020-01-02
Tools - Add Conditional PR Labeler

jonpas:
- Title

_Testing for now._


jonpas:
Don't like how this one works, no way of `OR` conditions (only `AND`) and no complex or mutually-exclusive conditions.

PR #7363, opened 2020-01-02
Medical - Add unconscious animations

diwako:
- Will add various new animations to unconscious units
- Each animation uses the same prefix for easy identification
- new function used to apply matching animation
- Each animation will take the position of the ragdoll into account
- Effects are local to each machine as `setUnconscious` has no synced ragdoll over network

Note:
`CBA_fnc_waitUntilAndExecute` is being used to wait for an unconscious animation to play. All animation event handlers have been tested yet none of them pick up the initial after ragdoll animation.


bux:
I don't like "anims" as a folder name :)

would prefer "data"

diwako:
Fine by me ¯\_(ツ)_/¯

addons/medical_engine/XEH_preInit.sqf
```diff
@@ -52,4 +58,21 @@ addMissionEventHandler ["Loaded", {
     } forEach GVAR(fixedStatics);
 }];
 
+["ace_unconscious", {
+    params ["_unit", "_active"];
+    if (_active) then {
+        if (_unit getVariable [QGVAR(waitForAnim), true]) then {
+            [{(animationstate _this) find QGVAR(face) != -1}, {
+                [_this, animationstate _this] call FUNC(applyAnimAfterRagdoll);
+            }, _unit, 20] call CBA_fnc_waitUntilAndExecute;
```
commy2: What if client disconnects in these 20 seconds?

commy2:
I don't like "unc" as file name.
That's the name for some toads. https://en.wikipedia.org/wiki/Fire-bellied_toad

PabstMirror:
- [ ] need to test how this will work for jips

addons/medical_engine/XEH_preInit.sqf
```diff
@@ -52,4 +58,21 @@ addMissionEventHandler ["Loaded", {
     } forEach GVAR(fixedStatics);
 }];
 
+["ace_unconscious", {
+    params ["_unit", "_active"];
+    if (_active) then {
+        if (_unit getVariable [QGVAR(waitForAnim), true]) then {
+            [{(animationstate _this) find QGVAR(face) != -1}, {
```
dedmen:
```suggestion
            [{(animationState _this) find QGVAR(face) != -1}, {
```

addons/medical_engine/XEH_preInit.sqf
```diff
@@ -52,4 +58,21 @@ addMissionEventHandler ["Loaded", {
...
+                [_this, animationstate _this] call FUNC(applyAnimAfterRagdoll);
```
dedmen:
```suggestion
                [_this, animationState _this] call FUNC(applyAnimAfterRagdoll);
```

addons/medical_engine/XEH_preInit.sqf
dedmen: animationState objNull returns `""`
find will always be -1, will run till timeout and fail.

diwako: After 20 seconds nothing should happen.

addons/medical_engine/CfgMoves.hpp
```diff
@@ -8,6 +8,76 @@ class CfgMovesMaleSdr: CfgMovesBasic {
             weaponIK = 0;
         };
 
+        class Unconscious;
+        class DeadState;
+        class ace_unc_1: DeadState {
```
jonpas: It's written with full word literally everywhere but this class and file name, might as well do it here, would be a lot more readable, I had no clue what "unc" was at first.

AUTHORS.txt
```diff
@@ -102,6 +102,7 @@ Karneck <dschultz26@hotmail.com>
 Kavinsky <nmunozfernandez@gmail.com>
 Keithen <Keithen.Neu@gmail.com>
 Kllrt <kllrtik@gmail.com>
+KokaKolaA3
```
jonpas: Do we have his permission?

diwako: When i stated the mod version back in the day I asked him if he would be okay to add his animations to ACE. I remember him having a talk with someone about licensing as well. Last I talked to him about it was on 1st April 2019 (what an unfortunate date). Asked him if he is still alright with them being in ace and them being renamed with the `ace_*` prefix

Conversation about it over discord, sadly in german
image

Without his OK I would have not made this pr

bux: Das ist doch super :)

addons/medical_engine/CfgMoves.hpp
diwako: just config names or also file names?

diwako: Files renamed and configs names changed

PabstMirror:
Will try to do testing soon,
- [ ] Check for problems with mp sync
- [ ] Check for problems with the recent physx uncon fix

diwako:
Tested with the physx uncon fix. Noticed it breaks as you are not in the default unconscious animation anymore. So, I changed the current fix to reflect the changes made in this pr.

diwako:
Also did a small test with another person on a dedicated server. Chased around a squad of units local to me, while ramming them with quads. After one keeled over not dead I asked the second person to report what they saw.
Also toggled unconscious on the second person and myself for tests.
Seems to work, tho I'd say it should be tested further in a live mission.

diwako:
A modified ace_medical_engine.pbo was provided to 3 units for testing now.

On another note, should a CBA setting be added that disables the applying of the animations? Behavior will then be the same as  it is with current release. However you will not be in animation state `unconscious` but what ever was chosen from the `QGVAR(face%)` states.

BaerMitUmlaut:
Should the animation maybe contain unconscious in it's name? Might make it more clear and would allow using find for both vanilla and our custom animations.

diwako:
Can be done, just means the name of the animation will be longer. Tho I would add it at the end, e.g. `QGVAR(faceUp_unconscious)`. Separated with a `_` so it is not camel case and unconscious is spelled fully lower case as the original animation is called.

jonpas:
Uhm... `gh` CLI did not work logically and merged it upstream...

I'll keep it merged since it only needed testing and it will be done in next 2 weeks.

diwako:
I'll keep the branch on my repo around till then.

jonpas:
So far zero problems in testing!

PabstMirror:
same here, feedback has been positive

diwako:
🎉

PR #7373, opened 2020-01-03
Goggles - Skip rain effects on snow maps

PabstMirror:
Not tested on GM map, which might use same effect system?


addons/goggles/XEH_postInit.sqf
```diff
@@ -133,12 +133,16 @@ if (!hasInterface) exitWith {};
 
         }] call CBA_fnc_addEventHandler;
 
+        // Handle "rain is snow" maps like cup's chenarus winter
```
jonpas: CUP's Chernarus

PabstMirror:
GM map should have config value of 0.8, so this fix will also work

PR #7384, opened 2020-01-04
Spanish translation - Captives

maxpipoka:


PabstMirror:
Merged as part of #7432

PR #7392, opened 2020-01-04
Medical - Combine advancedBandages and woundReopening settings

mharis001:
- Combine `advancedBandages` and `woundReopening` settings into a single 3 option setting
    - `woundReopening` was only used when `advancedBandages` was enabled - since `BasicBandage` can't reopen
- Remove "grey bandaged wounds" from displaying in injury lists
    - Caused confusion as these weren't actual bandaged wounds instead, open wounds with amount of zero
- Improve injury lists wound entry adding code

### DIFF
--- a/addons/medical_gui/functions/fnc_updateInjuryList.sqf
+++ b/addons/medical_gui/functions/fnc_updateInjuryList.sqf
@@ -104,45 +104,34 @@ if (_totalIvVolume >= 1) then {
 // Add entries for open, bandaged, and stitched wounds
 private _woundEntries = [];
 
-private _fnc_getWoundDescription = {
-    private _classIndex = _woundClassID / 10;
-    private _category = _woundClassID % 10;
-    private _className = EGVAR(medical_damage,woundsData) select _classIndex select 6;
-    private _suffix = ["Minor", "Medium", "Large"] select _category;
-    private _woundName = localize format [ELSTRING(medical_damage,%1_%2), _className, _suffix];
-    if (_amountOf >= 1) then {
-        format ["%1x %2", ceil _amountOf, _woundName];
-    } else {
-        format [localize LSTRING(PartialX), _woundName];
-    };
+private _fnc_processWounds = {
+    params ["_wounds", "_format", "_color"];
+
+    {
+        _x params ["_woundClassID", "_bodyPartN", "_amountOf"];
+
+        if (_selectionN == _bodyPartN && {_amountOf > 0}) then {
+            private _classIndex = _woundClassID / 10;
+            private _category   = _woundClassID % 10;
+
+            private _className = EGVAR(medical_damage,woundsData) select _classIndex select 6;
+            private _suffix = ["Minor", "Medium", "Large"] select _category;
+            private _woundName = localize format [ELSTRING(medical_damage,%1_%2), _className, _suffix];
+
+            private _woundDescription = if (_amountOf >= 1) then {
+                format ["%1x %2", ceil _amountOf, _woundName]
+            } else {
+                format [localize LSTRING(PartialX), _woundName]
+            };
+
+            _woundEntries pushBack [format [_format, _woundDescription], _color];
+        };
+    } forEach _wounds;
 };
 
-{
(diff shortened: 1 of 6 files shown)


addons/medical_treatment/stringtable.xml
```diff
@@ -43,6 +43,25 @@
             <Italian>Abilita le azioni di controllo del polso, controllo della pressione sanguigna e controllo della risposta al trattamento invece dell'azione generica Diagnose.\ Se disabilitata, l'azione di RCP sarà mostrata solo quando l'esecuzione della RCP è appropriata.\Le azioni abilitate da questa impostazione sono necessarie per determinare se una persona è in stato di incoscienza o in arresto cardiaco.</Italian>
             <Czech>Povoluje kontrolu srdečního tepu, krevního tlaku a reakce pacienta namísto univerzální diagnózy.\nKdyž je tato možnost vypnuta, CPR akce bude dostupná pouze pokud je vhodné ji provést.\nAkce které tato možnost zapíná jsou nutné k určení zda je pacient v bezvědomí nebo má srdeční zástavu.</Czech>
         </Key>
+        <Key ID="STR_ACE_Medical_Treatment_AdvancedMedication_DisplayName">
+            <English>Advanced Medication</English>
+            <German>Erweiterte Medikamente</German>
```
bux:
```suggestion
            <German>Erweiterte Medikation</German>
```

addons/medical_gui/functions/fnc_updateInjuryList.sqf
```diff
@@ -104,45 +104,34 @@ if (_totalIvVolume >= 1) then {
...
+};
 
-{
-    _x params ["_woundClassID", "_bodyPartN", "_amountOf"];
-    if (_selectionN == _bodyPartN && {_amountOf > 0}) then {
-        _woundEntries pushBack [format ["[S] %1", call _fnc_getWoundDescription], [0.7, 0.7, 0.7, 1]];
-    };
-} forEach GET_STITCHED_WOUNDS(_target);
+[GET_OPEN_WOUNDS(_target), "%1", [1, 1, 1, 1]] call _fnc_processWounds;
+[GET_BANDAGED_WOUNDS(_target), "[B] %1", [0.88, 0.7, 0.65, 1]] call _fnc_processWounds;
```
PabstMirror:
```suggestion
`if ((missionNamespace getVariable [QEGVAR(medical_treatment,advancedBandages), 1]) == 2) then {`
    [GET_BANDAGED_WOUNDS(_target), "[B] %1", [0.88, 0.7, 0.65, 1]] call _fnc_processWounds;
```
might save a little time as you can only get bandaged/stitched in mode2

PabstMirror:
Might just be me, but I will miss seeing all the old "bandaged" wounds

addons/medical_gui/functions/fnc_updateInjuryList.sqf
mharis001: I think we'd still want to show any bandaged wounds in case the setting was changed mid-mission because those would still be able to be stitched.

SilentSpike: I'd be for only listing them when reopening is active. Wouldn't they then appear when the UI is recreated if the setting is changed mid-mission?

mharis001: The injuries list is constantly refreshed. The concern is that if a unit originally has a bandaged wound that could reopen or be stitched (`advancedBandaged == 2`), changing the setting would cause the originally bandaged wound to not be shown but it can still reopen and be stitched. 

Bandaging open wounds after changing the setting would not produce any new bandaged wounds - not able to reopen or be stitched.

SilentSpike: Ah I see the distinction here now between "bandaged wounds" the array and "bandaged wounds" when re-opening isn't enabled. Missed that before.

Wonder if we could handle that setting changing better 🤔 Probably beyond the scope of this PR though.

PR #7396, opened 2020-01-04
Medical - Add fracture chance setting

mharis001:
- Add `fractureChance` setting, https://github.com/acemod/ACE3/issues/3594#issuecomment-570797080
- Cleanup unused strings

### DIFF
--- a/addons/main/script_mod.hpp
+++ b/addons/main/script_mod.hpp
@@ -11,7 +11,7 @@
 
 // MINIMAL required version for the Mod. Components can specify others..
 #define REQUIRED_VERSION 1.96
-#define REQUIRED_CBA_VERSION {3,12,2}
+#define REQUIRED_CBA_VERSION {3,14,0}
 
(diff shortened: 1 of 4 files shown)


PabstMirror:
What if we made the setting control `FRACTURE_DAMAGE_THRESHOLD`
or have magnitude of damage somehow influence the random chance
```
&& {_woundDamage > FRACTURE_DAMAGE_THRESHOLD}
&& {random 1 < (linearConversion [FRACTURE_DAMAGE_THRESHOLD, 2*FRACTURE_DAMAGE_THRESHOLD, _woundDamage, 1, 1.5, true] * EGVAR(medical,fractureChance))}
```

mharis001:
Can change to that, would need to update the tooltip to make that clear. Maybe something like: "The probability of a fracture causing wound resulting in a fracture. This value is increased by the amount of damage, maxing out at a 50% increase to this probability at double the minimum fracture damage threshold."

PR #7400, opened 2020-01-05
Changed: Scripted calls to putInEarPlugs/joinTeam won't show hint

ColdEvul:
- Added ability to suppress hint message for team switch
- Added ability to suppress hint message for earplugs

This ability can be useful when it comes to scripted auto solutions to avoid a heap of messages on spawn, respawn or recently logged in clients.


commy2:
The functions are marked as public: no.

Personally I would take the opportunity to set the default value of _displayHint to false and to pass true everywhere in ACE.

commy2:
Why did you remove the default value? Now the argument is no longer optional. It should be inverted, not removed.

ColdEvul:
Misunderstood reverted and changed it to false

PR #7405, opened 2020-01-05
Changed: Updated italian translation

Italiandrache:


Blutze:
While you're at it, please check the last two keys in Advanced Fatigue (resistance doesn't sound right, and either way it isn't consistent with older keys) and `STR_ACE_CSW_PlaceTripod_displayName` (those things do in fact have three legs).

Issue #7378, opened 2020-01-03
People don't wake up from unconsciousness


Brumes-Wolf:
**Is your enhancement related to a problem?**
Unlike before the rewrite people/AI seem to never wake up on their own, and also seem to go uncon quite easily

**Solution you'd like:**
A setting or toggle that allows people to wake up when they are not stable.

**Additional context:**
For a small non milsim group having people be forced to respawn as soon as they go uncon is not very fun nor good gameplay. The previous system of people waking up on their own if they weren't very damaged worked better for this use case.

Bummeri:
Can you adjust the  "seem to go uncon quite easily" by adjusting this setting?
![Uncon set](https://i.imgur.com/HBIYero.png)

Does the epipen not help with the chance of waking up? Especially if you adjust this setting higher?
![Epi set](https://i.imgur.com/Rio4epD.png)

Brumes-Wolf:
I adjusted the unconsciousness  wake up chance to 1.0 and people still did not seem to wake up. I have not tried changing the pain coefficient yet, nor did I change the Epi wake up chance. I will try these sometime soon.

But my main request here is for people to be able to wake up on their own without help from others. so epipens aren't relevant to that.

Bummeri:
> I adjusted the unconsciousness wake up chance to 1.0 and people still did not seem to wake up.

I guess the unit cannot wake up since it:
- Is bleeding still too much
- Has lost too much blood
- Is in too much pain

I dont think bleeding slows on its own?
Does pain go down on its own?

It might be that the system does not permit any random wakeups if nothing is made to assist the uncon unit?

Brumes-Wolf:
> > I adjusted the unconsciousness wake up chance to 1.0 and people still did not seem to wake up.
> 
> I guess the unit cannot wake up since it:
> 
> * Is bleeding still too much
> * Has lost too much blood
> * Is in too much pain
> 
> I dont think bleeding slows on its own?
> Does pain go down on its own?
> 
> It might be that the system does not permit any random wakeups if nothing is made to assist the uncon unit?

What I meant is that previously units could randomly wake up without assistance if they were hit only a few times and didn't lost too much blood previously. I would like a settings to return this behavior to the rewrite version

dedmen:
I think the default random wakeup chance is too low?
With advanced medication you seem to basically have to wait for it, and it has only a 5% chance, checked every 15 seconds.
you can give epi, but epi boost by default is 100%, so.. none.

[...]

PR #7407, opened 2020-01-05
Medical - Fix epinephrine wake-up chance boost string

jonpas:
- Title - ref. https://github.com/acemod/ACE3/issues/7378#issuecomment-570726654

Not sure if it's a direct multiplier due to connection with epinephrine effectiveness, so I used "boost" as is in the setting name.

Could also be improved with showing it as a percentage in the UI.

### DIFF
--- a/addons/medical/stringtable.xml
+++ b/addons/medical/stringtable.xml
@@ -42,14 +42,7 @@
             <Spanish>La probabilidad de que una unidad estabilizada recupere la conciencia y se levante (se comprueba cada 15 segundos)</Spanish>
         </Key>
         <Key ID="STR_ACE_Medical_spontaneousWakeUpEpinephrineBoost_DisplayName">
-            <English>Epinephrine Increases Wake Up Chance</English>
-            <Japanese>アドレナリンで覚醒確率上昇</Japanese>
-            <Chinese>腎上腺素甦醒機率增加</Chinese>
-            <French>L'épinéphrine augmente les chances de réveil</French>
-            <Italian>L'epinefrina aumenta la possibilità di rialzarsi</Italian>
-            <Czech>Epinefrin zvyšuje šanci na probuzení</Czech>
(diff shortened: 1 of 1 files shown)


addons/medical/stringtable.xml
```diff
@@ -42,14 +42,7 @@
...
         <Key ID="STR_ACE_Medical_spontaneousWakeUpEpinephrineBoost_DisplayName">
-            <English>Epinephrine Increases Wake Up Chance</English>
-            <Japanese>アドレナリンで覚醒確率上昇</Japanese>
-            <Chinese>腎上腺素甦醒機率增加</Chinese>
-            <French>L'épinéphrine augmente les chances de réveil</French>
-            <Italian>L'epinefrina aumenta la possibilità di rialzarsi</Italian>
-            <Czech>Epinefrin zvyšuje šanci na probuzení</Czech>
-            <Portuguese>Epinefrina Aumenta a chance de acordar</Portuguese>
-            <Spanish>Epinefrina, aumenta la posibilidad de despertar</Spanish>
+            <English>Epinephrine Wake Up Chance Boost</English>
```
thojkooi: Boosts wake up chance?

jonpas: This is a string for a setting name, it's not an explanation of what it does.

PabstMirror: ref https://github.com/acemod/ACE3/blob/master/addons/medical_statemachine/functions/fnc_handleStateUnconscious.sqf#L37-L42

It doesn't directly change the wake up chance, 
but instead decreases the time between doing wakeup random
I did it this way because epi is somewhat short lived, so I felt like this would be the most responsive

Not positive on the math, but the effect should be similar

jonpas: Yeah, math looks good, but the setting name was confusing.

PR #7415, opened 2020-01-07
Medical Engine - Apply damage to the correct hitpoint

### LINKED ISSUE #7205: Implement incoming damage normalization
A task discussed at our last meeting. The background is that we're pretty certain each body part receives differently scaled damage (by the engine) from identical incoming damage events.

The basic idea is to normalise these values onto the same scale for comparison and so that wound damage is consistent (we perform comparison between the damage to each body part to decide where a wound is applied for example).

If implemented this should be heavily tested to see the results produced by different damage sources compared to the current behaviour. Ideally they should remain roughly the same if not make more sense (wound size, etc.). As such, this task doesn't have to happen before the release of the rewrite.

Baer came up with these scaling values (divisors) for damage:
```
hithead: 1
hitbody: 1.00001
hitarms: 0.0540424
hitlegs: 0.333338
```
using the following code
```sqf
fn_test = {
    params ["_mempoint", "_hitpoint"];
    private _target = "B_Soldier_F" createVehicle position player;
    _target setUnitLoadout getUnitLoadout (configFile >> "EmptyLoadout");
    private _mempointPos = _target modelToWorldVisualWorld (_target selectionPosition _mempoint);
    private _rayOrigin = _mempointPos vectorAdd [0, -1, 0];
    private _intersects = lineIntersectsSurfaces [_rayOrigin, _mempointPos, objNull, objNull, true, 1, "FIRE"];
    private _intersectPos = _intersects#0#0;
    if (_intersects isEqualTo []) exitWith {
        systemChat ("No intersection with " + _mempoint);
    };
    [{
        params ["_intersectPos"];
        private _bullet = "B_556x45_Ball" createVehicle [0, 0, 10];
        _bullet setPosASL (_intersectPos vectorAdd [0, 0, 0]);
        _bullet setVelocity [0, 100, 0];
    }, [_intersectPos], 1] call CBA_fnc_waitAndExecute;
    [{
        params ["_hitpoint", "_target"];
        diag_log formatText ["%1: %2", _hitpoint, (_target getHitPointDamage _hitpoint) / 0.489123];
        systemChat format ["%1: %2", _hitpoint, (_target getHitPointDamage _hitpoint) / 0.489123];
        deleteVehicle _target;
    }, [_hitpoint, _target], 2] call CBA_fnc_waitAndExecute;
};
diag_log text "-------------------";
{
    [fn_test, _x, 3 * _forEachIndex] call CBA_fnc_waitAndExecute;
} forEach [
    ["pilot", "hithead"],
    ["pelvis", "hitbody"],
    ["relbow", "hitarms"],
    ["rknee", "hitlegs"]
];
```

Some notes:
  - Damage can be scaled as early as possible within the damage handler here (values will then pass through to wounding): https://github.com/acemod/ACE3/blob/a53a2725e9bd07551f0180c8dcc78af2b93e251f/addons/medical_engine/functions/fnc_handleDamage.sqf#L73-L80
  - Should allow using a single value for worst wounding here: https://github.com/acemod/ACE3/blob/a53a2725e9bd07551f0180c8dcc78af2b93e251f/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf#L98
  - May allow us to remove the head prioritisation code in the damage handler: https://github.com/acemod/ACE3/blob/a53a2725e9bd07551f0180c8dcc78af2b93e251f/addons/medical_engine/functions/fnc_handleDamage.sqf#L85-L89

pterolatypus:
- Change handleDamage to calculate a unit's armour and assign the damage location based on damage _before_ armour.
- This means the config must be searched for details of their equipment - this data is cached to improve performance

References:
- [Damage system](https://community.bistudio.com/wiki/Arma_3_Damage_Description)
- [PPE Config](https://community.bistudio.com/wiki/Arma_3_Soldier_Protection) 

~Still need to do caching for uniforms, as they work differently.~ done I think
~Not sure what values are used if no uniform is worn - currently I check the unit's base class, but maybe it should default to 1?~ Using the base armor value from the unit itself, no modifiers from uniform.
~Do I need to handle inheritance in the config lookups, or does the game do that for me?~ apparently it's handled if I use the right command

My first major PR so be gentle

- close #7205 ~commy2


addons/medical_engine/XEH_preInit.sqf
```diff
@@ -28,6 +28,15 @@ if (isNil QUOTE(FATAL_SUM_DAMAGE_WEIBULL_K) || isNil QUOTE(FATAL_SUM_DAMAGE_WEIB
...
 };
 
+// Cache for armour values of equipped items (vests etc)
+if (isNil QGVAR(armourCache_items)) then {
+    GVAR(armourCache_items) = false call CBA_fnc_createNamespace;
+};
+// Separate cache for uniforms because they work differently
+if (isNil QGVAR(armourCache_uniforms)) then {
+    GVAR(armourCache_uniforms) = false call CBA_fnc_createNamespace;
+};
```
veteran29: I don't see a point for these `isNil`s. the caches should just be created without a check.

SilentSpike:
Am I right in thinking this is basically a better way of achieving what we wanted to with #7205 (because it accounts for the varied armour values of equipment)?

It sounds reasonable to me, though I've never personally looked in-depth into how arma's armour system actually works so I can't comment on whether the calculations being done are correct at this moment. Some other maintainers are more familiar with that and might chime in here.

My primary concern with a change like this is performance hit, because the damage EH is a bit of a hot spot in the code. It may be better to scan the whole config at game start and cache all values for quick retrieval later even if a given mission might only use a small subset 🤔 @dedmen as the resident performance guru, any thoughts on that?

addons/medical_engine/XEH_preInit.sqf
pterolatypus: You're right. Fixed.

pterolatypus:
This should fix the issue Pabst highlighted there. The original thing shown by Baer's tests is separate I believe. I think that's actually caused by rounds partially-penetrating limbs and not dumping all their energy.

According to the debug console perf tester, looking up a single item in the config is about 0.077ms and a single call without caching does 1-3 lookups taking between 0.121ms and 0.264ms.
So if 100 units each wearing distinct gear get hit at the same time, that adds up to about 26ms.

The other thing is, caching all the equipment before-mission means I'd have to trawl the entire CfgWeapons including things like guns and cosmetic items that, afaik, can't have armour.

JonBons:
What about making CAManBase or something run some init code that does the config lookup and cache?

commy2:
That is a lot more code than I imagined. Some decissions seem questionable to me.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
```diff
@@ -0,0 +1,73 @@
...
+params ["_unit","_hitpoint",["_noCache",false]];
+
+// Structural damage is inferred differently, not needed here
+if (_hitpoint in ["", "#structural"]) exitwith {[1,1]};
+
+
+// Get base values from the unit class
+private _unitCfg = configFile >> "CfgVehicles" >> (typeOf _unit);
+private _armour = if (isNumber (_unitCfg >> "armor")) then {getNumber (_unitCfg >> "armor")} else {2};
+private _explosion = if (isNumber (_unitCfg >> "explosionShielding")) then {getNumber (_unitCfg >> "explosionShielding")} else {0.4};
```
commy2: Where do those hard coded numbers come from? Basic_Body? I feel like, since this is cached, it should read those from config instead of hard coding.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+// If the uniform is valid, use hitpoint values from the unit class it defines
+private _uniform = uniform _unit;
+if !(_uniform isEqualTo "") then {
+    private _cfg = configFile >> "CfgWeapons" >> _uniform >> "ItemInfo";
+    if (!isText (_cfg >> "uniformClass")) exitwith {};
```
commy2: Can this ever happen?

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
```diff
@@ -0,0 +1,73 @@
...
+ * 0: Armour <NUMBER>
+ * 1: ExplosionShielding <NUMBER>
+ *
+ * Example:
+ * [player, "HitChest"] call ace_medical_engine_fnc_getHitpointArmour
+ *
+ * Public: No
+ */
+
+params ["_unit","_hitpoint",["_noCache",false]];
```
commy2: What's the point of noCache? Seems like bloat to me. If you want to clear the cache for debugging, delete the namespaces.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
```diff
@@ -0,0 +1,73 @@
...
+    if (_cached isEqualTo []) then {
+        TRACE_2("Uniform armour cache miss",_uniformClass,_noCache);
+        // Scan config for uniform 'unit' hitpoints
+        private _hpConfig = configFile >> "CfgVehicles" >> _uniformClass >> "HitPoints";
+        _cached = [];
+        {
+            private _hpArmour = getNumber (_x >> "armor");
+            private _hpExpl = getNumber (_x >> "explosionShielding");
+            _cached pushBack [configName _x, [_hpArmour, _hpExpl]];
+        } foreach ("true" configClasses _hpConfig);
```
commy2: `configClasses` ignores inherited sub classes. This is needed for Turrets in vehicle config for example, but I am pretty sure it is wrong here.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
```diff
@@ -0,0 +1,73 @@
...
+        _explosion = _explosion * _cached#1#1;
+    }
+};
+
+// Add values from equipped vests and other gear
+{
+    private _values = [_x, _hitpoint, _noCache] call FUNC(getItemArmour);
+    _armour = _armour + _values#0;
+    _explosion = _explosion + _values#1;
+} foreach ([vest _unit, headgear _unit] select {!(_x isEqualTo "")});
```
commy2: forEach
Also, just make getItemArmor return 0 for "" instead of using two loops.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+// Return
+[_armour, _explosion]
```
commy2: For this script, consider putting:
```sqf
private _cfgWeapons = configFile >> "CfgWeapons";
```
at the beginning to save some commands.

Do we use `#` over `select` in ACE now?

addons/medical_engine/functions/fnc_getItemArmour.sqf
```diff
@@ -0,0 +1,58 @@
...
+ params ["_item","_hitpoint",["_noCache",false]];
+
+_hitpoint = toLower _hitpoint;
+
+// Return the cached value if found
+private _cached = if (_noCache) then {[]} else {GVAR(armourCache_items) getVariable [_item,[]]};
+if !(_cached isEqualTo []) exitwith {
+    private _index = _cached findIf {_x#0 isEqualTo _hitpoint};
+    if (_index == -1) then {[0,0]} else {(_cached select _index)#1};
+};
```
commy2: The cache could be more powerful by merging the item classname and hitpoint as key using `format`. Then you don't need a loop.
```
getVariable [format ["%1$%2", _item, _hitpoint], []];
```
$ is not a valid char in classnames, thus safe to use a separator.

addons/medical_engine/functions/fnc_getItemArmour.sqf
```diff
@@ -0,0 +1,58 @@
...
+TRACE_1("Item armour cache miss",_item);
+
+private _protection = configFile >> "CfgWeapons" >> _item >> "ItemInfo" >> "HitpointsProtectionInfo";
+// If the item has no protection, cache as empty
+if (!isClass _protection) exitwith {
+    GVAR(armourCache_items) setVariable [_item,[]];
```
commy2: Any benefit in setting this as empty array? Isn't this the default already? Could maybe save some RAM skiping this.

addons/medical_engine/functions/fnc_getItemArmour.sqf
```diff
@@ -0,0 +1,58 @@
...
+};
+
+// Read all armour values from config
+// TODO: do I need to worry about inheritance here?
+// TODO: if a hitpoint has multiple entries this will find them all, but only the first will be returned by findIf. Do I need to do something (add them together?) or is this fine?
+private _armourValues = [];
+{
+    private _protectionHitpoint = toLower getText (_x >> "hitpointName");
+    private _armour = if (isNumber (_x >> "armor")) then {getNumber (_x >> "armor")} else {0};
+    private _expl = if (isNumber (_x >> "explosionShielding")) then {getNumber (_x >> "explosionShielding")} else {0};
```
commy2: `getNumber` already reports 0 if the entry is not a number.

addons/medical_engine/functions/fnc_getItemArmour.sqf
```diff
@@ -0,0 +1,58 @@
...
+    _armourValues pushBack [_protectionHitpoint,[_armour,_expl]];
+} foreach ("isText (_x >> ""hitpointName"")" configClasses _protection);
```
commy2: forEach. Not sure about `configClasses` as it has the issue described above.

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -51,52 +55,61 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
...
-        [_damageLeftArm,  PRIORITY_LEFT_ARM,  "LeftArm"],
-        [_damageRightArm, PRIORITY_RIGHT_ARM, "RightArm"],
-        [_damageLeftLeg,  PRIORITY_LEFT_LEG,  "LeftLeg"],
-        [_damageRightLeg, PRIORITY_RIGHT_LEG, "RightLeg"]
+        [_damageHead#0,     _damageHead#1,      PRIORITY_HEAD,      "Head"],
+        [_damageBody#0,     _damageBody#1,      PRIORITY_BODY,      "Body"],
+        [_damageLeftArm#0,  _damageLeftArm#1,   PRIORITY_LEFT_ARM,  "LeftArm"],
+        [_damageRightArm#0, _damageRightArm#1,  PRIORITY_RIGHT_ARM, "RightArm"],
+        [_damageLeftLeg#0,  _damageLeftLeg#1,   PRIORITY_LEFT_LEG,  "LeftLeg"],
+        [_damageRightLeg#0, _damageRightLeg#1,  PRIORITY_RIGHT_LEG, "RightLeg"]
```
commy2:
```
         _damageRightLeg + [PRIORITY_RIGHT_LEG, "RightLeg"]
```
is less commands and should therefore be faster.

commy2: Alternatively just
```
[_damageRightLeg,  PRIORITY_RIGHT_LEG, "RightLeg"]
```
as I think `sort` handles nested arrays.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
commy2: Consider `getAllHitPoints` to get all valid hitpoints.

addons/medical_engine/functions/fnc_getItemArmour.sqf
pterolatypus: I need to distinguish between "item not in cache -> scan config" and "item in cache but no values -> return default"

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
pterolatypus: Each of these functions refer to CfgWeapons exactly once so I'm not sure how it would help. I could pass `configFile >> "CfgWeapons"` as a parameter to `fnc_getItemArmour`, but that seems silly because it'll be constant.

As for `select` vs `#`, I couldn't find specific guidelines on it - I've used `#` where the index is fixed because I feel it looks neater (more like C-style array access) and `select` where it's variable, but I'm happy to change it if `#` is considered bad practice.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
pterolatypus: Do you have a link for `getAllHitPoints` syntax? I can't find it on the biki.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
pterolatypus: These numbers are originally from "O_Soldier_F" because that's what I was looking at at the time
`armor=2` is the same for every unit in the game.
`explosionShielding` is either 0.4 (most opfor/indep) or 0.3 (most blufor)
So they seemed like sensible default values

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
commy2: https://community.bistudio.com/wiki/getAllHitPointsDamage

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
commy2: The problem I have with `select` is, that it introduces a command with elevated precendence. This is a diversion from the usual left to right execution.
I could see it being useful as transcended syntax refering to the element of an array like cpp arr[1], however, since `#` is implemented as binary command, it still has lower precendence than unary commands, which makes it pointless, a disappointment and therefore triggers disgust response in the brain of the reviewer.

addons/medical_engine/functions/fnc_getItemArmour.sqf
```diff
@@ -44,10 +46,10 @@ if (!isClass _protection) exitwith {
 private _armourValues = [];
 {
     private _protectionHitpoint = toLower getText (_x >> "hitpointName");
-    private _armour = if (isNumber (_x >> "armor")) then {getNumber (_x >> "armor")} else {0};
-    private _expl = if (isNumber (_x >> "explosionShielding")) then {getNumber (_x >> "explosionShielding")} else {0};
+    private _armour = getNumber (_x >> "armor");
+    private _expl = getNumber (_x >> "explosionShielding");
     _armourValues pushBack [_protectionHitpoint,[_armour,_expl]];
-} foreach ("isText (_x >> ""hitpointName"")" configClasses _protection);
+} forEach configProperties [_protection, "isClass _x && {isText (_x >> ""hitpointName"")}"];
```
commy2: Wait, why would there be a sub-class without hitpointName entry?

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
bux: we don't use `#` in ACE

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
commy2: Well, I mean, you could fall back to the naked model:
```sqf
private _type = typeOf _unit; //"B_Soldier_F"
private _nakedClass = _cfgVehicles >> getText (_cfgWeapons >> getText (_cfgVehicles >> "B_Soldier_F" >> "nakedUniform"));
private _armor = getNumber (_nakedClass >> "armor");
private _explosion = getNumber (_nakedClass >> "explosionShielding");
```

However, assuming we can't guarantee that these values are defined, we can't be any more sure those have the values either. And the game likely wouldn't use them as fallback anyway. It would just go with 0 and possibly complain or act in a undefined manner (like missile projectile models without mass/zero mass).

Alternatively you could read them from CAManBase.

armor entry is first defined in class All. explosionShielding is first defined in CAManBase. CAManBase also defines armor with a sensible value. All objects inherit from CAManBase and if some mod for some reason created their own complete class, huge parts of the scripting wouldn't work as everyone is using CAManBase as common base class for person objects.

It should therefore really just be `getNumber`. We know the entry is defined or inherited.

Variable name should be spelled "_armor".

pterolatypus: Fair enough. I suppose, similar to the point below, if it isn't defined then something else is going to break first so it's not really our responsibility to check.

> Variable name should be spelled "_armor"

Dammit, thought I'd got away with that :P

addons/medical_engine/functions/fnc_getItemArmour.sqf
pterolatypus: I don't think this will work, for the same reason as below. E.g. when I look up "V_PlateCarrierGL_rgr$HitLegs" and find no value; does that mean I haven't cached it yet and need to scan the config, or I have checked it but it has no value for that hitpoint?

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
pterolatypus: I don't think I can use that, as I only have a classname not an object

addons/medical_engine/functions/fnc_getItemArmour.sqf
pterolatypus: Tested and it seems that if you do have one the vest is broken, the game won't even create the item; so I guess it's not an issue.

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -51,52 +55,61 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
...
+        HITPOINT_INDEX_HEAD, _damageHead select 1, HITPOINT_INDEX_BODY, _damageBody select 1, HITPOINT_INDEX_LARM, _damageLeftArm select 1, 
+        HITPOINT_INDEX_RARM, _damageRightArm select 1, HITPOINT_INDEX_LLEG, _damageLeftLeg select 1, HITPOINT_INDEX_RLEG, _damageRightLeg select 1
     ];
 
     _allDamages sort false;
-    (_allDamages select 0) params ["_receivedDamage", "", "_woundedHitPoint"];
-    if (_damageHead >= HEAD_DAMAGE_THRESHOLD) then {
-        TRACE_3("reporting fatal head damage instead of max",_damageHead,_receivedDamage,_woundedHitPoint);
-        _receivedDamage = _damageHead;
+    (_allDamages select 0) params ["","_receivedDamage", "", "_woundedHitPoint"];
```
commy2:
```suggestion
    (_allDamages select 0) params ["", "_receivedDamage", "", "_woundedHitPoint"];
```

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -34,9 +34,13 @@ if (_hitPoint isEqualTo "") then {
...
 if !(isDamageAllowed _unit && {_unit getVariable [QEGVAR(medical,allowDamage), true]}) exitWith {_oldDamage};
 
-// Damages are stored for "ace_hdbracket" event triggered last
 private _newDamage = _damage - _oldDamage;
-_unit setVariable [format [QGVAR($%1), _hitPoint], _newDamage];
+// Get armor value of hitpoint and calculate damage before armor
+private _armor = [_unit, _hitpoint] call FUNC(getHitpointArmor);
+private _realDamage = _newDamage * (_armor select 0);
+// Damages are stored for "ace_hdbracket" event triggered last
+_unit setVariable [format [QGVAR($%1), _hitPoint], [_realDamage,_newDamage]];
```
commy2:
```suggestion
_unit setVariable [format [QGVAR($%1), _hitPoint], [_realDamage, _newDamage]];
```

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,71 @@
...
+
+
+// Get base values from the unit class
+private _unitCfg = configFile >> "CfgVehicles" >> (typeOf _unit);
+private _armor = getNumber (_unitCfg >> "armor");
+private _explosion = getNumber (_unitCfg >> "explosionShielding");
+
+// If the uniform is valid, use hitpoint values from the unit class it defines
+private _uniform = uniform _unit;
+if !(_uniform isEqualTo "") then {
```
commy2:
```suggestion
if (_uniform != "") then {
```
Looks cleaner and has a command less, thus is faster.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Public: No
+ */
+
+params ["_unit","_hitpoint",["_noCache",false]];
+
+// Structural damage is inferred differently, not needed here
+if (_hitpoint in ["", "#structural"]) exitwith {[1,1]};
+
+
```
commy2:
```suggestion
```
Remove double newline.

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,59 @@
...
+};
+
+// Read all armor values from config
+// TODO: if a hitpoint has multiple entries this will find them all, but only the first will be returned by findIf. Do I need to do something (add them together?) or is this fine?
+private _armorValues = [];
+{
+    private _protectionHitpoint = toLower getText (_x >> "hitpointName");
+    private _armor = getNumber (_x >> "armor");
+    private _expl = getNumber (_x >> "explosionShielding");
+    _armorValues pushBack [_protectionHitpoint,[_armor,_expl]];
```
commy2: It's named `_explosion` in the other script.

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,59 @@
...
+    [0,0] //return default
+};
+
+// Read all armor values from config
+// TODO: if a hitpoint has multiple entries this will find them all, but only the first will be returned by findIf. Do I need to do something (add them together?) or is this fine?
+private _armorValues = [];
+{
+    private _protectionHitpoint = toLower getText (_x >> "hitpointName");
+    private _armor = getNumber (_x >> "armor");
+    private _expl = getNumber (_x >> "explosionShielding");
```
commy2:
```suggestion
    private _explosion = getNumber (_x >> "explosionShielding");
```

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,59 @@
...
+
+_hitpoint = toLower _hitpoint;
+
+if (_item isEqualTo "") exitwith {[0,0]};
+
+// Return the cached value if found
+private _cached = if (_noCache) then {[]} else {GVAR(armorCache_items) getVariable [_item,[]]};
+if !(_cached isEqualTo []) exitwith {
+    private _index = _cached findIf {(_x select 0) isEqualTo _hitpoint};
+    if (_index == -1) then {[0,0]} else {(_cached select _index) select 1};
```
commy2:
```suggestion
    _cached param [_index, [nil, [0,0]]] select 1 // return
```

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,59 @@
...
+    _armorValues pushBack [_protectionHitpoint, [_armor, _explosion]];
+} forEach configProperties [_protection, "isClass _x"];
+
+if (!_noCache) then {
+    GVAR(armorCache_items) setVariable [_item, _armorValues];
+};
+
+// Return the value that was actually requested
+private _index = _armorValues findIf {(_x select 0) isEqualTo _hitpoint};
+if (_index == -1) then {[0,0]} else {(_armorValues select _index) select 1} //return
```
commy2:
```suggestion
_armorValues param [_index, [nil, [0,0]]] select 1 // return
```

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,59 @@
...
+if !(_cached isEqualTo []) exitwith {
+    private _index = _cached findIf {(_x select 0) isEqualTo _hitpoint};
+    if (_index == -1) then {[0,0]} else {(_cached select _index) select 1};
+};
+TRACE_1("Item armor cache miss",_item);
+
+private _protection = configFile >> "CfgWeapons" >> _item >> "ItemInfo" >> "HitpointsProtectionInfo";
+// If the item has no protection, cache as empty
+if (!isClass _protection) exitwith {
+    GVAR(armorCache_items) setVariable [_item,[]];
```
commy2:
```suggestion
    GVAR(armorCache_items) setVariable [_item, []];
```

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,70 @@
...
+        _armor = _armor * (_values select 1) select 0;
+        _explosion = _explosion * (_values select 1) select 1;
+    }
+};
+
+// Add values from equipped vests and other gear
+{
+    private _values = [_x, _hitpoint, _noCache] call FUNC(getItemArmor);
+    _armor = _armor + (_values select 0);
+    _explosion = _explosion + (_values select 1);
```
commy2:
```sqf
    [_x, _hitpoint, _noCache] call FUNC(getItemArmor) params ["_armorX", "_explosionX"];
    _armor = _armor + _armorX;
    _explosion = _explosion + _explosionX;
```
Optimizing every small thing for this.

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -34,9 +34,13 @@ if (_hitPoint isEqualTo "") then {
 // Damage can be disabled with old variable or via sqf command allowDamage
 if !(isDamageAllowed _unit && {_unit getVariable [QEGVAR(medical,allowDamage), true]}) exitWith {_oldDamage};
 
-// Damages are stored for "ace_hdbracket" event triggered last
 private _newDamage = _damage - _oldDamage;
-_unit setVariable [format [QGVAR($%1), _hitPoint], _newDamage];
+// Get armor value of hitpoint and calculate damage before armor
+private _armor = [_unit, _hitpoint] call FUNC(getHitpointArmor);
+private _realDamage = _newDamage * (_armor select 0);
```
commy2:
```sqf
private _armor = [_unit, _hitpoint] call FUNC(getHitpointArmor) select 0;
private _realDamage = _newDamage * _armor;
```
Looks cleaner to me. Possibly L1 as:
```sqf
[_unit, _hitpoint] call FUNC(getHitpointArmor) params ["_armor"];
```

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,70 @@
...
+    private _cached = if (_noCache) then {[]} else {GVAR(armorCache_uniforms) getVariable [_uniformClass, []]};
+    if (_cached isEqualTo []) then {
+        TRACE_2("Uniform armor cache miss",_uniformClass,_noCache);
+        // Scan config for uniform 'unit' hitpoints
+        private _hpConfig = configFile >> "CfgVehicles" >> _uniformClass >> "HitPoints";
+        _cached = [];
+        {
+            private _hpArmor = getNumber (_x >> "armor");
+            private _hpExpl = getNumber (_x >> "explosionShielding");
+            _cached pushBack [configName _x, [_hpArmor, _hpExpl]];
```
commy2: Am I stupid or why are these not named `_armor` and `_explosion` like in the rest of the script?

pterolatypus: in this function `_armor` and `_explosion` are the running totals for the current hitpoint - `_hpArmor` and `_hpExpl` are used for each hitpoint read from the config.

commy2: Just re-use the identifiers. Nothing will overwrite, that is the whole reason for the `private` command.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -34,36 +34,34 @@ private _uniform = uniform _unit;
 if (_uniform != "") then {
     // Check for previously cached values
     private _uniformClass = getText (configFile >> "CfgWeapons" >> _uniform >> "ItemInfo" >> "uniformClass");
-    private _cached = if (_noCache) then {[]} else {GVAR(armorCache_uniforms) getVariable [_uniformClass, []]};
-    if (_cached isEqualTo []) then {
```
pterolatypus: Got myself muddled; empty isn't the same as not found.

addons/medical_engine/functions/fnc_getHitpointArmour.sqf
pterolatypus: Changed every occurrence to `armor` for consistency

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,68 @@
...
+ * 0: Armor <NUMBER>
+ * 1: ExplosionShielding <NUMBER>
+ *
+ * Example:
+ * [player, "HitChest"] call ace_medical_engine_fnc_getHitpointArmor
+ *
+ * Public: No
+ */
+
+params ["_unit","_hitpoint"];
```
commy2:
```suggestion
params ["_unit", "_hitpoint"];
```

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,57 @@
...
+ * 0: Armor <NUMBER>
+ * 1: ExplosionShielding <NUMBER>
+ *
+ * Example:
+ * ["V_PlateCarrier_rgr", "HitChest"] call ace_medical_engine_fnc_getItemArmor
+ *
+ * Public: No
+ */
+
+ params ["_item","_hitpoint"];
```
commy2:
```suggestion
params ["_item", "_hitpoint"];
```

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,68 @@
...
+
+// Structural damage is inferred differently, not needed here
+if (_hitpoint in ["", "#structural"]) exitwith {[1,1]};
+
+// Get base values from the unit class
+private _unitCfg = configFile >> "CfgVehicles" >> (typeOf _unit);
+private _armor = getNumber (_unitCfg >> "armor");
+private _explosion = getNumber (_unitCfg >> "explosionShielding");
```
commy2: Is this actually how it works though? I am pretty sure that the uniformClass determines the armor and explosive (and armorStructural) values and overwrites the unit completely...

pterolatypus: You are correct, uniform armor does override base class (tested by modding a CSAT uniform up to 1000 armor and putting it on a blufor unit).
New version correctly uses `nakedUniform` for this instead.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,78 @@
...
+
+// Add values from relevant gear
+{
+    [_x, _hitpoint] call _fnc_getItemArmor params ["_armorX", "_explosionX"];
+    _armor = _armor + _armorX;
+    _explosion = _explosion + _explosionX;
+} forEach [vest _unit, headgear _unit];
+
+// Return
+[_armor, _explosion];
```
commy2: Missing newline @ eof

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,78 @@
...
+// Get values for the specific hitpoint
+private _key = format ["%1$%2", _uniform, _hitpoint];
+private _hitpointValues = GVAR(armorCache) getVariable _key;
+if (isNil "_armorValues") then {
+    TRACE_2("Cache miss", _uniform, _hitpoint);
+    private _uniformClass = getText (configFile >> "CfgWeapons" >> _uniform >> "ItemInfo" >> "uniformClass");
+    private _hitpointCfg = configFile >> "CfgVehicles" >> _uniformClass >> "HitPoints" >> _hitpoint;
+    _hitpointValues = [getNumber (_hitpointCfg >> "armor"), getNumber (_hitpointCfg >> "explosionShielding")];
+    GVAR(armorCache) setVariable [_key, _hitpointValues];
+};
```
commy2: These lines should be part of `_fnc_getItemArmor`. Specifically after L28 it should be determined if the item is a uniform or not. If uniform, use this routine and store to cache. If helmet or vest, use the routine from L29-34.

This gets rid of some duplicated code.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
commy2:
```sqf
[_armor, _explosion] // return
```
We leave out the semi-colon for return values (even though the engine does not care).

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
commy2: To do this you can check if `CfgWeapons/<item>/ItemInfo/type` is 801, or better yet `TYPE_UNIFORM`, which is already defined from here:
https://github.com/acemod/ACE3/blob/master/addons/main/script_macros.hpp#L96

BIS_fnc_itemType would also work for me, but it is obviously slower (although this is all cached).

dedmen:
> It may be better to scan the whole config at game start and cache all values for quick retrieval later even if a given mission might only use a small subset

Nice that the tag didn't really give me any special notification :thinking: 
If everything is cached on first hit, so that we only have one small lagspike, then it should be fine?
We could also iterate over all currently existing units at postInit, and cache them then? Less useless data, and likely that on hit the gear is already in cache?

I don't think we really profiled the already existing hit/damage handlers, so we might be able to optimize on others too.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,78 @@
...
+// Helper function to check items with HitpointsProtectionInfo
+private _fnc_getItemArmor = {
+    params ["_item", "_hitpoint"];
+    if (_item isEqualTo "") exitWith {[0,0]};
+    private _key = format ["%1$%2", _item, _hitpoint];
+    private _armorValues = GVAR(armorCache) getVariable _key;
+    if (isNil "_armorValues") then {
+        TRACE_2("Cache miss", _item, _hitpoint);
+        private _cfg = configFile >> "CfgWeapons" >> _item >> "ItemInfo" >> "HitpointsProtectionInfo";
+        private _condition = format ["getText (_x >> ""hitpointName"") == ""%1""",_hitpoint];
```
dedmen: would prefer 
```suggestion
        private _condition = format ["getText (_x >> 'hitpointName') == '%1'",_hitpoint];
```

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,78 @@
...
+if (_uniform isEqualTo "") then {
+    _uniform =  getText (configFile >> "CfgVehicles" >> typeOf _unit >> "nakedUniform");
+};
+
+// Uniform base values are a multiplier for all hitpoint values
+private _baseValues = GVAR(armorCache) getVariable (_uniform + "$armor");
+if (isNil "_baseValues") then {
+    TRACE_1("Cache miss", _uniform);
+    private _uniformClass = getText (configFile >> "CfgWeapons" >> _uniform >> "ItemInfo" >> "uniformClass");
+    private _unitCfg = configFile >> "CfgVehicles" >> _uniformClass;
```
dedmen: Considering this is performance relevant, you maybe shouldn't re-lookup CfgWeapons/CfgVehicles everytime, and instead do it once and store it in a local.
`_fnc_getItemArmor`  could also use that local then.

SilentSpike:
>If everything is cached on first hit, so that we only have one small lagspike, then it should be fine?

Yeah my inclination was leaning this way too, because caching all combinations for everything could be a nightmare. Let's stick with cache on first hit.

>We could also iterate over all currently existing units at postInit, and cache them then? Less useless data, and likely that on hit the gear is already in cache?

This is probably a good compromise. I also had the thought that we could preemptively cache with a unit init handler.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
dedmen: but also considering that stuff is cached, that might cause two useless lookups everytime..

commy2:
I agree with first hit cache. Should this cause issues, we can add prloading later. First however, I want this to get to a state where uniform, vest, helment and unit lookup are all in one function with a single line of getVariable and setVariable each. Then this becomes easier.

I also think there may be potential for optimization outside the cache by not returning arrays that require excessive `select` usage in the event handler.

Whigital:
Just a thought, having the HitPoint armour values as part of the `medical_engine`:

- Cache with HitPoint armour values only, populated on unit init (`updateHitPointArmorValues`).
- Do a lookup against that cache in `handleDamage`.
- Update the cache on `Respawn`, `InventoryClosed`, `ContainerClosed` etc, if equipment changed.

Might be hard keeping the cache up-to-date on loadlout changes, but would keep handleDamage cleaner with only cache lookup (getVar).

commy2:
Equipment cannot be tracked on non-Player units due to lack of event handlers.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
pterolatypus: I've done this by passing a flag to the function instead; it seems wasteful to do an extra lookup since we already know which item is the uniform.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,67 @@
...
+ * Example:
+ * [player, "HitChest"] call ace_medical_engine_fnc_getHitpointArmor
+ *
+ * Public: No
+ */
+params ["_unit", "_hitpoint"];
+
+// Helper function to check cache and/or read config
+private _fnc_getItemArmor = {
+    params ["_item", "_hitpoint", ["_isUniform", false]];
```
commy2: I read your comment about you feeling checking if the item is a uniform or not is wasteful and I understand the intend.
However, you are missing the point. You're underestimating the power of the cache. By passing this as parameter, even if this parameter is basically never checked after the first execution, you are still investing extra time in this params command and the setting of a local variable.
This event handler will be executed many many times more than the caching event will take place. The cached isUniform check is on average lighter than this variable.

pterolatypus: hmm I suppose I follow - one extra config lookup per item is better than extra overhead on _every_ call. I struggle to believe that one extra param is worth it, but I'll defer to you since I'm by no means an expert.
Currently I'm only reading the config for the selected hitpoint, which means this will end up being called like a dozen times per-item; should I move back to caching all the hitpoints at once on the first call, to limit the overhead?

pterolatypus: That would involve adding a flag in the cache, to avoid that "have I cached this?" problem I mentioned, which is an extra `getVariable` on every execution so probably not, right?

commy2: >I struggle to believe that one extra param is worth it, but I'll defer to you since I'm by no means an expert.

It's not a question of worth it. It's simply min-maxing what we can get out of this.

>Currently I'm only reading the config for the selected hitpoint, which means this will end up being called like a dozen times per-item; should I move back to caching all the hitpoints at once on the first call, to limit the overhead?

See? This is what I consider not worth it. This lookup is once per hitpoint per item. Completely neglible to the large number of repeated executions.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
pterolatypus: As per commy's points, we expect the majority of calls to hit the cache so it's better to leave as much as possible until after that. A couple of extra `>>` only on cache miss is better than wasted processing on every event.
It _could_ be declared in the parent scope and only actually set/used if we do a lookup; but that means an extra `isNil` check, and is less readable.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,66 @@
...
+[_uniform, _hitpoint] call _fnc_getItemArmor params ["_armor", "_explosion"];
+
+// Add values from relevant gear
+{
+    [_x, _hitpoint] call _fnc_getItemArmor params ["_armorX", "_explosionX"];
+    _armor = _armor + _armorX;
+    _explosion = _explosion + _explosionX;
+} forEach [vest _unit, headgear _unit];
+
+[_armor, _explosion] // return
```
commy2: `_explosion` seems to be never used by anything.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,66 @@
...
+ * [player, "HitChest"] call ace_medical_engine_fnc_getHitpointArmor
+ *
+ * Public: No
+ */
+params ["_unit", "_hitpoint"];
+
+// Helper function to check cache and/or read config
+private _fnc_getItemArmor = {
+    params ["_item", "_hitpoint"];
+    if (_item isEqualTo "") exitWith {[0,0]};
```
commy2: Optimization: this line should be moved into the cache. `"$HitHead"` should be cached as [0,0] or 0 or w/e from inside.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,66 @@
...
+    private _key = format ["%1$%2", _item, _hitpoint];
+    private _armorValues = GVAR(armorCache) getVariable _key;
+    if (isNil "_armorValues") then {
+        TRACE_2("Cache miss", _item, _hitpoint);
+        private _itemInfo = configFile >> "CfgWeapons" >> _item >> "ItemInfo";
+        if (getNumber (_itemInfo >> "type") == TYPE_UNIFORM) then {
+            private _unitCfg = configFile >> "CfgVehicles" >> getText (_itemInfo >> "uniformClass");
+            private _baseValues = GVAR(armorCache) getVariable (_item + "$armor");
```
commy2: Observation, here you do `(_item + "$armor")` twice, while previously you stored it in `_key`. Which one seems better?

commy2: Is `$armor` really unique? Couldn't you use a key with different special chars (or multiple) to make it rigorous?

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,66 @@
...
+            _armorValues = [(_baseValues select 0) * getNumber (_entry >> "armor"), (_baseValues select 1) * getNumber (_entry >> "explosionShielding")];
+        } else {
+            private _condition = format ["getText (_x >> 'hitpointName') == '%1'",_hitpoint];
+            private _entry = configProperties [_itemInfo >> "HitpointsProtectionInfo", _condition] param [0, configNull];
+            _armorValues = [getNumber (_entry >> "armor"), getNumber (_entry >> "explosionShielding")];
+        };
+        GVAR(armorCache) setVariable [_key, _armorValues];
+    };
+    _armorValues // return
+};
```
commy2: The formatting is wild in `_fnc_getItemArmor`. Could use some more empty lines (before ifs), as well as line breaks in the long expressions.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,66 @@
...
+
+// Initialise with values from uniform
+[_uniform, _hitpoint] call _fnc_getItemArmor params ["_armor", "_explosion"];
+
+// Add values from relevant gear
+{
+    [_x, _hitpoint] call _fnc_getItemArmor params ["_armorX", "_explosionX"];
+    _armor = _armor + _armorX;
+    _explosion = _explosion + _explosionX;
+} forEach [vest _unit, headgear _unit];
```
commy2: Is a loop really needed if the array size is constant as well as short?

pterolatypus: In vanilla the only items that have HitpointsProtectionInfo are vests and helmets, but I don't know if it's _possible_ for other gear to define it too. I wrote it like this so that things like `goggles _unit` could easily be added later on if we find we need them.
Also it looks nicer than writing it out twice.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
pterolatypus: So, really `_explosion` should be used instead of `_armor` when the damage is explosive (`indirectHit` as opposed to `hit`) but at this stage we don't have a good way of determining whether that's the case.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
pterolatypus: >previously you stored it

Personally I think `_item + "$armor"` is more readable: in the other case I stored it in a variable because it's using `format` which I find less readable, and is probably fractionally slower.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
commy2: It's also slower. But I guess only marginally so (one command).

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
commy2: Isn't that always the case when the projectile is ""?

I am pretty sure explosion damage ⇒ selection "".
I am not sure explosion damage ⇔ selection "", however, I can't think of any counter example, aside from shit like drowning or falling. Can we catch those?

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,85 @@
...
+            if (isNil "_baseValues") then {
+                _baseValues = [getNumber (_unitCfg >> "armor"), getNumber (_unitCfg >> "explosionShielding")];
+                GVAR(armorCache) setVariable [(_item + "$$base"), _baseValues];
+            };
+
+            private _entry = _unitCfg >> "HitPoints" >> _hitpoint;
+            _armorValues = [
+                (_baseValues select 0) * getNumber (_entry >> "armor"),
+                (_baseValues select 1) * getNumber (_entry >> "explosionShielding")
+            ];
```
commy2: I have entangled this part a bit:
```sqf
         if (getNumber (_itemInfo >> "type") == TYPE_UNIFORM) then {
            private _config = configFile >> "CfgVehicles" >> getText (_itemInfo >> "uniformClass");
            private _entry = _config >> "HitPoints" >> _hitpoint;

            private _key = format ["%1$$base", _item];
            GVAR(armorCache) getVariable _key params ["_armor", "_explosion"];

            if (isNil "_armor") then {
                _armor = getNumber (_config >> "armor");
                _explosion = getNumber (_config >> "explosionShielding");
                GVAR(armorCache) setVariable [_key, [_armor, _explosion]];
            };

            _armorValues = [
                _armor * getNumber (_entry >> "armor"),
                _explosion * getNumber (_entry >> "explosionShielding")
            ];
        };
```

But then it occured to me that this is all inside a cache anyway. The lookups are minimal.
Really this all should just be:
```sqf
         if (getNumber (_itemInfo >> "type") == TYPE_UNIFORM) then {
            private _uniformClass = configFile >> "CfgVehicles" >> getText (_itemInfo >> "uniformClass");
            private _hitpointEntry = _uniformClass >> "HitPoints" >> _hitpoint;

            _armorValues = [
                getNumber (_uniformClass >> "armor") * getNumber (_hitpointEntry >> "armor"),
                getNumber (_uniformClass >> "explosionShielding") * getNumber (_hitpointEntry >> "explosionShielding")
            ];
        };
```

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
commy2: `format` is a unary command, while `+` is a binary command with alternative syntaxes.
No way `format` is slower than `+`.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
pterolatypus: Done some more testing and it seems I was wrong about this, `explosionShielding` works differently:
* It's a multiplier not a divisor
* It only seems to apply to unit hitpoints (so uniforms?)
* Both it **and** `armor` are applied on explosive damage

Explosive damage seems to report the ammo classname correctly so we can use that; it might make the most sense to move `EFUNC(damage,getTypeOfDamage)` into this component and then only pass the result of that (bullet, grenade etc) to the `woundReceived` event.

pterolatypus: On further inspection, because explosive damage is non-selection-specific I think it should be ignored here.
I have some ideas on improving nss damage as well, but that's a matter for another PR.

commy2:
Well yeah, okay, I suppose this is one way to go about it. Good enough for now I guess.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
commy2: There is still an empty line here that doesn't really belong. No reason to start a new scope/indent with an empty line.

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,69 @@
...
+
+        private _itemInfo = configFile >> "CfgWeapons" >> _item >> "ItemInfo";
+
+        if (getNumber (_itemInfo >> "type") == TYPE_UNIFORM) then {
+            private _unitCfg = configFile >> "CfgVehicles" >> getText (_itemInfo >> "uniformClass");
+            private _entry = _unitCfg >> "HitPoints" >> _hitpoint;
+
+            _armor = getNumber (_unitCfg >> "armor") * getNumber (_entry >> "armor")
+        } else {
+            private _condition = format ["getText (_x >> 'hitpointName') == '%1'",_hitpoint];
```
commy2:
```suggestion
            private _condition = format ["getText (_x >> 'hitpointName') == '%1'", _hitpoint];

File: addons/medical_engine/functions/fnc_getHitpointArmor.sqf [hunk 42]
```diff
@@ -0,0 +1,69 @@
...
+private _fnc_getItemArmor = {
+    params ["_item", "_hitpoint"];
+
+    private _key = format ["%1$%2", _item, _hitpoint];
+    private _armor = GVAR(armorCache) getVariable _key;
+
+    if (isNil "_armor") then {
+
+        TRACE_2("Cache miss", _item, _hitpoint);
+        if (_item isEqualTo "" || _hitpoint isEqualTo "") exitWith {
```
commy2: This doesn't really matter as it is cached, but you could also use something like:
```sqf
"" in [_item, _hitpoint]
```

commy2:
I like how clean this has become.

We should consider some things though:
- _fnc_getItemArmor as it's own function, so the local function does not have to be redefined every execution
- Caching the armor value on the unit itself. Something like:
```sqf
private _uniform = uniform _unit;
private _vest = vest _unit;
private _helmet = headgear _unit;

// If unit is naked, use its underwear class instead
if (_uniform isEqualTo "") then {
    _uniform =  getText (configFile >> "CfgVehicles" >> typeOf _unit >> "nakedUniform");
};

private _rags = format ["%1$%2$%3", _uniform, _vest, _helmet];
_unit getVariable [QGVAR(armorCache), [""]] params ["_prevRags", "_armor"];

if (_rags != _prevRags) then {
    _armor = 0;

    {
        _armor = _armor + ([_x, _hitpoint] call _fnc_getItemArmor);
    } forEach [_uniform, _vest, _helmet];

    _unit setVariable [QGVAR(armorCache), [_rags, _armor]];
};
```



 With the key being the uniform, vest, backpack combo.

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,48 @@
...
+        private _condition = format ["getText (_x >> 'hitpointName') == '%1'",_hitpoint];
+        private _entry = configProperties [_itemInfo >> "HitpointsProtectionInfo", _condition] param [0, configNull];
+
+        _armor = getNumber (_entry >> "armor");
+    };
+
+    GVAR(armorCache) setVariable [_key, _armor];
+};
+
+_armor // return
```
commy2:
```suggestion
_armor // return 

```
Missing newline @ eof

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,48 @@
...
+
+if (isNil "_armor") then {
+
+	TRACE_2("Cache miss", _item, _hitpoint);
+	if (_item isEqualTo "" || _hitpoint isEqualTo "") exitWith {
+	    _armor = 0;
+	    GVAR(armorCache) setVariable [_key, _armor];
+	};
+
+    private _itemInfo = configFile >> "CfgWeapons" >> _item >> "ItemInfo";
```
commy2: Indentation broken.

commy2: Ah, those are tabs.

commy2:
lgtm

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,48 @@
...
+ */
+params ["_item", "_hitpoint"];
+
+private _key = format ["%1$%2", _item, _hitpoint];
+private _armor = GVAR(armorCache) getVariable _key;
+
+if (isNil "_armor") then {
+
+    TRACE_2("Cache miss", _item, _hitpoint);
+    if (_item isEqualTo "" || _hitpoint isEqualTo "") exitWith {
```
mharis001: Lazy eval.

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,48 @@
+#include "script_component.hpp"
+/*
+ * Author: Pterolatypus
+ * Reads cache or config to determine what armor an item provides to a particular hitpoint
```
mharis001:
```suggestion
 * Returns the armor value the given item provides to a particular hitpoint.
```

addons/medical_engine/functions/fnc_getHitpointArmor.sqf
```diff
@@ -0,0 +1,46 @@
+#include "script_component.hpp"
+/*
+ * Author: Pterolatypus
+ * Checks a unit's equipment to calculate the total armor on a hitpoint
```
mharis001:
```suggestion
 * Checks a unit's equipment to calculate the total armor on a hitpoint.
```

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -0,0 +1,48 @@
...
+ * Public: No
+ */
+params ["_item", "_hitpoint"];
+
+private _key = format ["%1$%2", _item, _hitpoint];
+private _armor = GVAR(armorCache) getVariable _key;
+
+if (isNil "_armor") then {
+
+    TRACE_2("Cache miss", _item, _hitpoint);
```
mharis001:
```suggestion
    TRACE_2("Cache miss",_item,_hitpoint);
```

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -15,15 +15,15 @@
...
 
 private _key = format ["%1$%2", _item, _hitpoint];
 private _armor = GVAR(armorCache) getVariable _key;
 
 if (isNil "_armor") then {
-
-    TRACE_2("Cache miss", _item, _hitpoint);
-    if (_item isEqualTo "" || _hitpoint isEqualTo "") exitWith {
+    TRACE_2("Cache miss",_item,_hitpoint);
+    if ("" in [_item, _hitpoint]) exitWith {
```
commy2: nice

Issue #7409, opened 2020-01-05
Morphine does not reduce pain when "Advanced Medication" is disabled


Amuff1n:
**Mods:**
- **Arma 3:** 1.96.146114
- **CBA:** 3.13.0.191116 stable
- **ACE3:** 3.13.0.45 stable

**Description:**
When "Advanced Medication" is disabled, morphine does not successfully suppress pain. Immediately after administering morphine to a patient in pain, the pain will quickly return to the state it was previously. This seems to happen regardless of the Pain Coefficient. In effect, the pain only goes away naturally. I have noticed this issue in singleplayer and during my unit's last play session on our dedicated server.

**Steps to reproduce:**
- Go to the editor and place down a medic (player) and another unit.
- Disable "Medic AI", "Advanced Diagnose", "Advanced Bandages", and "Advanced Medication" in the ACE Medical addon options.
- Enter the mission and shoot the other unit in the leg.
- Bandage the leg and stop the bleeding. Notice the unit is in pain.
- Give the unit morphine.
- Once morphine is administered, notice that the pain is only momentarily suppressed before immediately returning to its previous level.

No amount of morphine given seems to make the pain go away quicker than normal.

Compare this to having "Advanced Medication" enabled:
- Go to the editor and place down a medic (player) and another unit.
- Disable "Medic AI", "Advanced Diagnose", "Advanced Bandages", but leave "Advanced Medication" enabled.
- Enter the mission and shoot the other unit in the leg.
- Bandage the leg and stop the bleeding. Notice the unit is in pain.
- Give the unit morphine.
- Wait around 15 seconds.
- Notice that the pain is decreasing, as expected. 

**Expected behavior:**
Similar to before the medical rewrite, I would expect the pain to be suppressed and stay suppressed. Before the medical rewrite, this would happen immediately on basic medical settings. I do not know if this is supposed to be the case after the rewrite or not, but regardless of speed the pain should decrease.

**Where did the issue occur?**
- Dedicated / Singleplayer / Editor (Singleplayer) 

**Log Files:**
https://pastebin.com/HZ7mFXTS

**Additional context:**
It is possible I do not understand the new medical settings fully. I would expect disabling "Advanced Diagnose", "Advanced Bandages", and "Advanced Medication" to emulate the previous basic medical, at least in treatment. If this is not the case and it is possible to adjust the settings to fix this "issue", I would be very grateful.

**Video:**
[...]

PR #7417, opened 2020-01-07
Fix suppression for basic medication

diwako:
- Fix basic medication not suppressing pain

Addressing #7409

### DIFF
--- a/addons/medical_vitals/functions/fnc_updatePainSuppress.sqf
+++ b/addons/medical_vitals/functions/fnc_updatePainSuppress.sqf
@@ -20,15 +20,16 @@
 
 params ["_unit", "_painSupressAdjustment", "_deltaT", "_syncValue"];
 
-_unit setVariable [VAR_PAIN_SUPP, 0 max _painSupressAdjustment, _syncValue];
-
 // Handle continuous pain reduction
 private _pain = GET_PAIN(_unit);
 _unit setVariable [VAR_PAIN, 0 max (_pain - _deltaT / PAIN_FADE_TIME), _syncValue];
 
-// Handles simple medication
 if (isNil QEGVAR(medical_treatment,advancedMedication) || {!EGVAR(medical_treatment,advancedMedication)}) then {
+    // Handles simple medication
     private _painSupress = _unit getVariable [VAR_PAIN_SUPP, 0];
     _painSupress = _painSupress - _deltaT / PAIN_SUPPRESSION_FADE_TIME;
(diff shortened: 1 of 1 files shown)


commy2:
Please explain.

diwako:
> Please explain.

When advanced medication is disabled, the pain suppression variable will be always set to 0. This happens right at the beginning of the changed file with
```sqf
_unit setVariable [VAR_PAIN_SUPP, 0 max _painSupressAdjustment, _syncValue];
```

Turns out the value for `_painSupressAdjustment` is always 0 when on basic medication. This value comes from the `handluUnitsVital` function. 
https://github.com/acemod/ACE3/blob/master/addons/medical_vitals/functions/fnc_handleUnitVitals.sqf#L101

The value will be always 0 because advanced medication is not enabled, no medication will be set on the unit. In case of morphine, the pain suppression value will be altered and then you exit out of the medication function.
https://github.com/acemod/ACE3/blob/master/addons/medical_treatment/functions/fnc_medicationLocal.sqf#L30

This means in the `handleUnitVitals` there is no medication to iterate over and to add to the value for `_painSupressAdjustment`.

The fix is to rearrange when the `updatePainSuppress` function updates the current pain suppression value. It does not make sense to set the value and then set it again based on the already altered value for pain suppression on that unit. Especially if that altered value is always 0.
Hence why I moved the initial set of the variable into the else case, as only in the case of advanced medication being enabled the value for `_painSupressAdjustment` can be 0.

commy2:
But there is no function call between the previous and the new position of setVariable.

diwako:
Not that is misunderstand, there was none and there should not be any.

The pain suppression value is being set in 2 functions specifically. One is is `ace_medical_treatment_fnc_medicationLocal` specifically when advanced medication is disabled and the other is  `ace_medical_vitals_fnc_updatePainSuppress`.

For basic medication the first function is used to increase the value and the second is used to decrease the value of the pain suppression over time. The bug is `ace_medical_vitals_fnc_updatePainSuppress` sets the value to the given value of `_painSupressAdjustment` and then decreases the value as we are in basic medication. But the pain suppression value was already overwritten with the wrong value `0`, so the logic thinks you just have no morphine in your blood system at all. That is why i moved the first `setVariable`.

`ace_medical_vitals_fnc_updatePainSuppress`  is being called from `ace_medical_vitals_fnc_handleUnitVitals`  and `ace_medical_treatment_fnc_medicationLocal` when ever someone injects an auto injector.

PR #7439, opened 2020-01-10
release-drafter - Add category for translations

PabstMirror:
and remove `enhancement` so label to title 1-1 mapping

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -10,15 +10,16 @@ categories:
     labels:
       - 'kind/feature request'
       - 'kind/enhancement'
-      - 'area/translations'
   - title: '**FIXED:**'
     labels:
       - 'kind/bug fix'
   - title: '**CHANGED:**'
     labels:
-      - 'kind/enhancement'
       - 'kind/cleanup'
       - 'area/compatibility'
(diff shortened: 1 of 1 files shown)


jonpas:
Enhancement would make sense to go into separate IMPROVED header IMO. We can always manually group headers together for smaller releases.

.github/release-drafter.yml
```diff
@@ -10,15 +10,16 @@ categories:
...
-      - 'area/translations'
   - title: '**FIXED:**'
     labels:
       - 'kind/bug fix'
   - title: '**CHANGED:**'
     labels:
-      - 'kind/enhancement'
       - 'kind/cleanup'
       - 'area/compatibility'
+  - title: '**Translations:**'
```
mharis001:
```suggestion
  - title: '**TRANSLATIONS:**'
```

PabstMirror:
I think we can rename the existing labels,
I think it would be best if the labels are "What you see is what you get" to changelog categories

jonpas:
I personally never liked current label system, much prefer the old way, like ACEX still has.

It doesn't have to be one-one in terms of nanes, but feature - added etc. would indeed nice.

PR #7440, opened 2020-01-10
Update French Translations

Elgin675:
- Title


addons/medical/stringtable.xml
```diff
@@ -46,6 +46,7 @@
         <Key ID="STR_ACE_Medical_spontaneousWakeUpEpinephrineBoost_DisplayName">
             <English>Epinephrine Wake Up Chance Boost</English>
             <French>L'épinéphrine augmente les chances de réveil</French>
+            <Polish>Zwiększenie szansy na wybudzenie po podaniu epinefryny</Polish>
```
dedmen: > Translations: French

> <Polish>

🤔

Elgin675: @dedmen Apparently, it happened when I was merged master :
https://github.com/acemod/ACE3/blob/master/addons/medical/stringtable.xml#L47L48

What am I supposed to do now ?

PabstMirror: I don't think it's a problem, doesn't show on PR diff

PabstMirror: `Zwiększenie szansy na wybudzenie po podaniu epinefryny` is same as current master current

PR #7445, opened 2020-01-11
Arsenal - Filter non-scope 2 mags from being auto-added

### LINKED ISSUE #7352: ML300L - FakeMagazine's .paa missing
- **Arma 3:** `1.96.14611`
- **CBA:** `3.13.0`
- **ACE3:** `3.13.0 release`

**Description:**
When equipping the ML300L from the Arsenal, it is loaded with a `FakeMagazine`. Opening the Arsenal's magazine tab removes it, but skipping that step leaves the magazine in. Opening the inventory now throws an error message `Warning Message: Picture equip\m\m_fakemagazine.paa not found`, and the loaded magazine does not show an icon.

**Expected behavior:**
The flashlight also works without any magazine. Is there a way to always make it spawn unloaded?

Editor (Singleplayer) / Virtual ACE Arsenal

**Log Files:**
RPT: https://gist.github.com/Blutze/93259739d7ff740169c3ae7c9415547c

**Additional context:**
Not exactly related to the issue itself, but the maplight beam is smaller than the regular maglite's. Depending on the shape of the cone, that might be realistic, but gutfeeling would be expecting a more powerful effect from a heavier, bigger, dedicated item.

PabstMirror:
Fix #7352

`_compatibleMags` was undefined
so this also fixes re-adding same magazines when switching between weapons
e.g.
have rifle with drum mag, switch to different camo and it will keep same drum

### DIFF
--- a/addons/arsenal/functions/fnc_onSelChangedLeft.sqf
+++ b/addons/arsenal/functions/fnc_onSelChangedLeft.sqf
@@ -57,8 +57,12 @@ switch (GVAR(currentLeftPanel)) do {
                 call _fnc_clearPreviousWepMags;
 
                 private _compatibleItems = (_item call bis_fnc_compatibleItems) apply {tolower _x};
+                private _cfgMags = configFile >> "CfgMagazines";
+                private _compatibleMags = ([_item, true] call CBA_fnc_compatibleMagazines) select { getNumber (_cfgMags >> _x >> "scope") == 2 };
                 GVAR(center) addWeapon _item;
-                GVAR(center) addWeaponItem [_item, [(getArray (configfile >> "cfgweapons" >> _item >> "magazines")) select 0]];
+                if ((count _compatibleMags) > 0) then {
+                    GVAR(center) addWeaponItem [_item, [_compatibleMags select 0]];
+                };
 
                 {
                     if (tolower _x in _compatibleItems || {_x in _compatibleMags}) then {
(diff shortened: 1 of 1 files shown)


addons/arsenal/functions/fnc_onSelChangedLeft.sqf
```diff
@@ -95,8 +99,12 @@ switch (GVAR(currentLeftPanel)) do {
                 call _fnc_clearPreviousWepMags;
 
                 private _compatibleItems = (_item call bis_fnc_compatibleItems) apply {tolower _x};
+                private _cfgMags = configFile >> "CfgMagazines";
+                private _compatibleMags = ([_item, true] call CBA_fnc_compatibleMagazines) select { getNumber (_cfgMags >> _x >> "scope") == 2 };
                 GVAR(center) addWeapon _item;
-                GVAR(center) addWeaponItem [_item, [(getArray (configfile >> "cfgweapons" >> _item >> "magazines")) select 0]];
+                if ((count _compatibleMags) > 0) then {
```
mharis001:
```suggestion
                if !(_compatibleMags isEqualTo []) then {
```

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
```diff
@@ -57,8 +57,12 @@ switch (GVAR(currentLeftPanel)) do {
...
```
mharis001:
```suggestion
                if !(_compatibleMags isEqualTo []) then {
```

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
```diff
@@ -132,8 +140,12 @@ switch (GVAR(currentLeftPanel)) do {
...
```
mharis001:
```suggestion
                if !(_compatibleMags isEqualTo []) then {
```

jonpas:
Requested changes addressed in https://github.com/acemod/ACE3/pull/7543.

PR #7451, opened 2020-01-11
Adjusted weight of grenades

ColdEvul:
- Added matching real world weight to throwables


addons/realisticweights/CfgMagazines.hpp
```diff
@@ -0,0 +1,30 @@
...
+    class SmokeShellPurple: SmokeShell {
+        mass = 11.85;
+    };
+    class SmokeShellYellow: SmokeShell {
+        mass = 11.85;
+    };
+    class MiniGrenade: CA_Magazine {
+        mass = 2.02;
+    };
+};
```
jonpas:
```suggestion
};

```

PabstMirror:
SmokeShell: HandGrenade { 
vanilla mass = 4; 
this mass = 14.99;

jonpas:
Hmm, pretty large difference.

Does this take volume of the grenade into account? An that grenades are usually mounted in extra pouches on the vest, outside of backpack and so on?

This must be taken into account, as Arma inventory is a mix of weight and volume, and it has no "stacking" or "placement" knowledge. I feel like grenades are at a pretty good weight as they are.

ColdEvul:
**HandGrenade**
Vanilla mass: 10 (0.454kg (1.00lb))
New mass: 8.82 (0.400kg (0.88lb))

**SmokeShell**
Vanilla mass: 4 (0.181kg (0.40lb))
New mass: 14.99 (0.680kg (1.50lb))

**SmokeShell (Color)**
Vanilla mass: 4 (0.181kg (0.40lb))
New mass: 11.85 (0.538kg (1.19lb))

**MiniGrenade**
Vanilla mass: 6 (0.272kg (0.60lb))
New mass: 2.02 (0.092kg (0.20lb))

This is the weights based on mass values i used

ColdEvul:
My self is unsure about the smokes tho but for the others i think its a good change :)

BaerMitUmlaut:
Increasing weight could break existing loadouts.

severgun:
Arma does not have volume attribute. Increasing weight will make vests useless.

ColdEvul:
Removed smokes from this the grenade weights are still percent.

commy2:
Just here to repeat that the `mass` entry is not a weight, but a measure of generic inventory space; and that the whole realisticweights component is trash that should be deleted.

BaerMitUmlaut:
It still makes sense because the vanilla values are all over the place.

ColdEvul:
Is there a better way to 'fix' Arma mass vs weight vs. inventory space?

commy2:
Accept that weight is not a thing, and that the entry "mass" means inventory space.

ColdEvul:
Is there a more reasonable way of calculating mass?

commy2:
From what? mass is not to be calculated. You compare with the mass of similar items in the base game and then use those values.

Drofseh:
`mass` in ACE3 *is* weight though, as inventory and arsenal UI both show kg/lbs directly from the mass of the items you have.

jonpas:
But game doesn't support that, what the UI shows is technically a lie, or the best programmatical approximation.

Drofseh:
Sure if you're running a pure vanilla game then mass is an arbitrary unit.

In ACE3 several components define mass == weight, inventory and arsenal components and this component, *realistic weights*.

We should consistently apply that.

jonpas:
We do not define `mass == weight`, it's just a form of display and calling it one of the two. Otherwise you'll see `"Arbitrary Mass/Weight: 5 bananas"` everywhere.

ColdEvul:
Whould a `ace_weight` variable that effect advanced fatigue as well effect play in to the cargo system (if defined)?

Lighter stuff whoukd make you lighter and be able to move longer while more weight will slow you down and able to sprint shorter.

Making dropping your backpack a valid tactic.

commy2:
Yes it would, and the scripts that retrieve this value could fall back to mass (and the BS formula I cooked up for it) should the entry be undefined.
However, the scripts could no longer use the load commands, but would have to add up themselves.

ColdEvul:
I'll up cook a demo and framework for it. I will close this once I have it up.

Drofseh:
I'd love an ace inventory system with `ace_weight` and `ace_size` for items and corresponding `ace_volume` for containers. `weight` effecting stamina and `size` filling up inventory `volume`

PR #7455, opened 2020-01-12
Medical - Add ability for unit specific damage threshold

diwako:
- Allows mission makers to set damage threshold per unit
- added macro to get damage threshold
- if no variable is set on the unit the global CBA setting value will be used

The new macro differentiates between AI and player.

The variable name for this is currently `ace_medical_unitDamageThreshold`. This is the same name the old medical system used. However the value was an array of 3 values in the old system. Now it is just a number.
So, it is debatable which name for the variable should be used.

### DIFF
--- a/addons/medical_damage/CfgEden.hpp
+++ b/addons/medical_damage/CfgEden.hpp
@@ -1,0 +1,28 @@
+class Cfg3DEN {
+    class Attributes {
+        class Slider;
+        class GVAR(slider): Slider {
+            attributeLoad = "params [""_ctrlGroup""]; private _slider = _ctrlGroup controlsGroupCtrl 100; private _edit = _ctrlGroup controlsGroupCtrl 101; _slider sliderSetPosition _value; _edit ctrlSetText ([_value, 1, 1] call CBA_fnc_formatNumber); ";
+            attributeSave = "params [""_ctrlGroup""]; sliderPosition (_ctrlGroup controlsGroupCtrl 100); ";
+            onLoad = "params [""_ctrlGroup""]; private _slider = _ctrlGroup controlsGroupCtrl 100; private _edit = _ctrlGroup controlsGroupCtrl 101; _slider sliderSetRange [0, 10]; _slider ctrlAddEventHandler [""SliderPosChanged"", { params [""_slider""]; private _edit = (ctrlParentControlsGroup _slider) controlsGroupCtrl 101; private _value = sliderPosition _slider;  _edit ctrlSetText ([_value, 1, 1] call CBA_fnc_formatNumber); }]; _edit ctrlAddEventHandler [""KillFocus"", { params [""_edit""]; private _slider = (ctrlParentControlsGroup _edit) controlsGroupCtrl 100; private _value = ((parseNumber ctrlText _edit) min 10) max 0; _slider sliderSetPosition _value; _edit ctrlSetText str _value; }];";
+        };
+    };
+    class Object {
+        class AttributeCategories {
+            class ace_attributes {
+                class Attributes {
+                    class GVAR(threshold) {
+                        property = QUOTE(threshold);
+                        control = QGVAR(slider);
+                        displayName = CSTRING(Eden_threshold_DisplayName);
+                        tooltip = CSTRING(Eden_threshold_Description);
+                        expression = QUOTE(if (_value >= 0.1) then {_this setVariable [ARR_3(QQEGVAR(medical,damageThreshold),_value,true)]});
+                        typeName = "NUMBER";
+                        condition = "objectBrain";
+                        defaultValue = 0;
+                    };
+                };
+            };
+        };
+    };
+};
--- a/addons/medical_damage/CfgEden.hpp
+++ b/addons/medical_damage/CfgEden.hpp
@@ -2,9 +2,30 @@ class Cfg3DEN {
     class Attributes {
         class Slider;
         class GVAR(slider): Slider {
-            attributeLoad = "params [""_ctrlGroup""]; private _slider = _ctrlGroup controlsGroupCtrl 100; private _edit = _ctrlGroup controlsGroupCtrl 101; _slider sliderSetPosition _value; _edit ctrlSetText ([_value, 1, 1] call CBA_fnc_formatNumber); ";
-            attributeSave = "params [""_ctrlGroup""]; sliderPosition (_ctrlGroup controlsGroupCtrl 100); ";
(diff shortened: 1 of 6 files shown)


addons/medical_engine/script_macros_medical.hpp
```diff
@@ -179,6 +179,7 @@
 #define GET_OPEN_WOUNDS(unit)       (unit getVariable [VAR_OPEN_WOUNDS, []])
 #define GET_BANDAGED_WOUNDS(unit)   (unit getVariable [VAR_BANDAGED_WOUNDS, []])
 #define GET_STITCHED_WOUNDS(unit)   (unit getVariable [VAR_STITCHED_WOUNDS, []])
+#define GET_DAMAGE_THRESHOLD(unit)  (unit getVariable [QEGVAR(medical,unitDamageThreshold), [EGVAR(medical,AIDamageThreshold),EGVAR(medical,playerDamageThreshold)] select (isPlayer unit)])
```
mharis001:
```suggestion
#define GET_DAMAGE_THRESHOLD(unit)  (unit getVariable [QEGVAR(medical,damageThreshold), [EGVAR(medical,AIDamageThreshold), EGVAR(medical,playerDamageThreshold)] select isPlayer unit])
```

I think if the expected value is different changing the name is fine, and the "unit" part can be omitted.

diwako: Yeah i also was thinking to just set it to `ace_medical_damageThreshold`. It seems very close to the original one tho. So I am not sure.

veteran29:
WDYT about adding eden attribute for this?

diwako:
@veteran29 question is how. It would not make sense to do a 1:1 threshold translation.
Basically I imaging the default threshold is 1, then all units placed in editor would have 1 as their default value. It would always use the per unit set value instead of the cba setting,
Maybe it is because I never added an eden attribute to a class so I do not really know if the value is always set, thus I have my concerns.

diwako:
Added threshold slider to units
Min value 0, max value 10

dedmen:
> Basically I imaging the default threshold is 1, then all units placed in editor would have 1 as their default

attribute setters are code. Can check for == default and just not set the var

diwako:
> > Basically I imaging the default threshold is 1, then all units placed in editor would have 1 as their default
> 
> attribute setters are code. Can check for == default and just not set the var

I've done it now a bit differently, as the slider is more precise than what is shown on the number selector, anything below 0.1, which shows as 0 on the selector, will be seen as "use cba settings threshold value".

addons/medical_damage/CfgEden.hpp
```diff
@@ -0,0 +1,28 @@
+class Cfg3DEN {
+    class Attributes {
+        class Slider;
+        class GVAR(slider): Slider {
+            attributeLoad = "params [""_ctrlGroup""]; private _slider = _ctrlGroup controlsGroupCtrl 100; private _edit = _ctrlGroup controlsGroupCtrl 101; _slider sliderSetPosition _value; _edit ctrlSetText ([_value, 1, 1] call CBA_fnc_formatNumber); ";
+            attributeSave = "params [""_ctrlGroup""]; sliderPosition (_ctrlGroup controlsGroupCtrl 100); ";
+            onLoad = "params [""_ctrlGroup""]; private _slider = _ctrlGroup controlsGroupCtrl 100; private _edit = _ctrlGroup controlsGroupCtrl 101; _slider sliderSetRange [0, 10]; _slider ctrlAddEventHandler [""SliderPosChanged"", { params [""_slider""]; private _edit = (ctrlParentControlsGroup _slider) controlsGroupCtrl 101; private _value = sliderPosition _slider;  _edit ctrlSetText ([_value, 1, 1] call CBA_fnc_formatNumber); }]; _edit ctrlAddEventHandler [""KillFocus"", { params [""_edit""]; private _slider = (ctrlParentControlsGroup _edit) controlsGroupCtrl 100; private _value = ((parseNumber ctrlText _edit) min 10) max 0; _slider sliderSetPosition _value; _edit ctrlSetText str _value; }];";
```
commy2: line breaks with \

diwako: Done

diwako:
Can the label  "status/WIP" be removed? This was done some time ago.

addons/medical_damage/CfgEden.hpp
```diff
@@ -0,0 +1,49 @@
...
+                private _edit = (ctrlParentControlsGroup _slider) controlsGroupCtrl 101;\
+                private _value = sliderPosition _slider;\
+                _edit ctrlSetText ([_value, 1, 1] call CBA_fnc_formatNumber);\
+            }];\
+            _edit ctrlAddEventHandler [""KillFocus"", {\
+                params [""_edit""];\
+                private _slider = (ctrlParentControlsGroup _edit) controlsGroupCtrl 100;\
+                private _value = ((parseNumber ctrlText _edit) min 10) max 0;\
+                _slider sliderSetPosition _value;\
+                _edit ctrlSetText str _value;\
```
PabstMirror, line 27:
```suggestion
                _edit ctrlSetText (if (_value < 0.1) then { localize ""str_disp_default"" } else {[_value, 1, 1] call CBA_fnc_formatNumber});\
```

addons/medical_damage/CfgEden.hpp
```diff
@@ -0,0 +1,49 @@
...
+            sliderPosition (_ctrlGroup controlsGroupCtrl 100); ";
+            onLoad = "params [""_ctrlGroup""];\
+            private _slider = _ctrlGroup controlsGroupCtrl 100;\
+            private _edit = _ctrlGroup controlsGroupCtrl 101;\
+            _slider sliderSetRange [0, 10];\
+            _slider ctrlAddEventHandler [""SliderPosChanged"", {\
+                params [""_slider""];\
+                private _edit = (ctrlParentControlsGroup _slider) controlsGroupCtrl 101;\
+                private _value = sliderPosition _slider;\
+                _edit ctrlSetText ([_value, 1, 1] call CBA_fnc_formatNumber);\
```
PabstMirror, line 20:
```suggestion
                _edit ctrlSetText (if (_value < 0.1) then {localize ""str_disp_default""} else {[_value, 1, 1] call CBA_fnc_formatNumber});\
```

addons/medical_damage/CfgEden.hpp
```diff
@@ -0,0 +1,49 @@
+class Cfg3DEN {
+    class Attributes {
+        class Slider;
+        class GVAR(slider): Slider {
+            attributeLoad = "params [""_ctrlGroup""];\
+            private _slider = _ctrlGroup controlsGroupCtrl 100;\
+            private _edit = _ctrlGroup controlsGroupCtrl 101;\
+            _slider sliderSetPosition _value;\
+            _edit ctrlSetText ([_value, 1, 1] call CBA_fnc_formatNumber);";
```
PabstMirror, line 9:
```suggestion
            _edit ctrlSetText (if (_value < 0.1) then {localize ""str_disp_default""} else {[_value, 1, 1] call CBA_fnc_formatNumber});";
```

addons/medical_damage/CfgEden.hpp
```diff
@@ -0,0 +1,49 @@
...
+            class ace_attributes {
+                class Attributes {
+                    class GVAR(threshold) {
+                        property = QUOTE(threshold);
+                        control = QGVAR(slider);
+                        displayName = CSTRING(Eden_threshold_DisplayName);
+                        tooltip = CSTRING(Eden_threshold_Description);
+                        expression = QUOTE(if (_value >= 0.1) then {_this setVariable [ARR_3(QQEGVAR(medical,damageThreshold),_value,true)]});
+                        typeName = "NUMBER";
+                        condition = "objectBrain";
```
PabstMirror, line 42:
```suggestion
                        condition = "objectControllable";
```

PR #7460, opened 2020-01-13
Mine Detector - Updated mine detector audio from -20db to -2db

Sycholic:
Audio was always too soft in game, original file was this way add in audio changes in game and it basically made it almost footstep levels which with any kind of noise nearby would totally drown out any detection audio.

Updates mine detector audio volume levels to appropriate levels.  Original was about -20db adjusted amplitude to proper levels -2db (-14 LUFS)

WAV file, and encoded into BI WSS audio file type.


Sycholic:
Also might be possible to consider using a mono sound file in this case. for those who use surround or radios like tfa so its pushed out the center channel..  Kinda odd to hear the detector sound going to the read speakers when putting on the 'headphones mode'  if this maybe wanted I can easily convert it to mono if that is wanted.

dedmen:
> I can easily convert it to mono if that is wanted.

I don't see a reason why left or right channel should be different, so mono should be just fine? and we'll save some filesize

Sycholic:
mono when you using surround sound automatically pushes it to the center channel if existing or dual outputs both speakers evenly if the center channel isnt found or being used.  any voice overs or audio you want positionally centeredy should be recorded in mono for this reason.  This is how windows and almost any dolby 'anything' works.  example. ever run into a YT video where everything is pushed out the center speaker, thats only because its a mono audio track video.  YT dont even support > 2ch. audio playback (internally yes but not for playback)

This was the only reason I suggested and offering it, if desired.  its good as it but some might want some modernization.

Sycholic:
Ignore about the mono part just double looked at the original it is already mono channel. 44.1khz 16-bit some reason thought it was stereo.  
Again to be clear only thing I changed was increasing the amplitude to reasonable levels -20dB SNR in a audio file is quiet a large difference in volume.

PR #7173, opened 2019-09-06
Use CBA Versioning System properly

jonpas:
**When merged this pull request will:**
- Title

**Merge after https://github.com/CBATeam/CBA_A3/pull/1221 is released!**

TODO:
- [x] Copy updated `script_macros_common.hpp` on CBA release
  - With cleaned-up quoting `##`

PR #7479, opened 2020-01-18
Changed version to versionStr as fix for #7478

### LINKED ISSUE #7478: getVersion returns blank string as result
**Mods:**
- **Arma 3:** `1.96.146114 stable`
- **CBA:** `3.13.0.191116 stable`
- **ACE3:** `3.13.0.45 stable`

**Description:**
I'm attempting to support both pre-medical rewrite and medical rewrite versions of ace, I figured I could do this by checking version on preInit, and setting a flag of which methods to call in ace's medical API. Sadly it seems `ace_common_fnc_getVersion` only returns an empty string.

**Steps to reproduce:**
- Load any map
- Open Debug Console
- Check output of `[] call ace_common_fnc_getVersion`
- See that it returns a blank string

**Expected behavior:**
- `[] call ace_common_fnc_getVersion` should return version number

- Editor (Singleplayer) & Editor (Multiplayer) & Virtual Arsenal

**Log Files:**
- Can upload if need be, didn't see anything out of the ordinary

**Additional Context:**
- Worth noting that the version number does still appear in RPT, so I'll look around for where that's printed from, but I still think `getVersion` should return the version itself rather than having to look elsewhere for it.

- Checking the version print [here](https://github.com/acemod/ACE3/blob/7a8489f35eede399016712593fa0e63c2ba495b4/addons/common/functions/fnc_checkFiles.sqf#L21) which is working with the one in [getVersion](https://github.com/acemod/ACE3/blob/29a0732b0c788b25f47ec0e8ef121cd2614ad0fa/addons/common/functions/fnc_getVersion.sqf#L18)  it seems that it could either be `ACE_main` being used instead of `ace_main` as the root, or `version` not being populated while `versionStr` is.

**Screenshots:**
image

M3ales:
`ace_common_fnc_getVersion` currently returns "" due to using `getText()` on a number config entry. I'm unsure if using `versionStr` provides intended behavior since `version` is a shorter version of `versionStr`, this however provides a result which seems to be in line with the docstring above.

- Fix #7478 
- Changed `version` to `versionStr` based on the return type specified in docs, its likely that `version` used to contain `versionStr`'s information. If need be the alternative: `str getNumber ()` would also work.

### DIFF
--- a/addons/common/functions/fnc_getVersion.sqf
+++ b/addons/common/functions/fnc_getVersion.sqf
@@ -15,4 +15,4 @@
  * Public: Yes
  */
 
-getText (configFile >> "CfgPatches" >> "ACE_main" >> "version") // return
+getText (configFile >> "CfgPatches" >> "ACE_main" >> "versionStr") // return


PabstMirror:
ref #7173

PR #7497, opened 2020-01-29
CSW - Minor fixes for CSW + RHS Compats

Whigital:

`ace_csw`:
- Add Mk6 Mortar icon.

`ace_compat_rhs_usf3` / `ace_compat_rhs_afrf3`/ `ace_compat_rhs_gref3`:
- Add `WeaponSlotsInfo` to get rid of error picture error when opening inventory.
- Use `rhs_SPG9_INS` instead of `rhsgref_ins_SPG9` for `ace_compat_rhs_afrf3_spg9_carry`.
- Add `ACE_CSW` inheritance on `ace_compat_rhs_afrf3_spg9m_carry`.


addons/csw/CfgWeapons.hpp
```diff
@@ -148,6 +145,9 @@ class CfgWeapons {
         model = QPATHTOEF(apl,ACE_CSW_Bag.p3d);
         modes[] = {};
         picture = QPATHTOF(UI\StaticAT_Icon.paa);
+        magazines[] = {
+            "Titan_AT"
```
commy2: Nope. This is for their home config patch. Where the class originates from.

commy2:
Why would static weapons appear in the arsenal?

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -180,7 +180,15 @@ class CfgWeapons {
         scope = 2;
         model = QPATHTOEF(apl,ACE_CSW_Bag.p3d);
         modes[] = {};
-        picture = "\rhsafrf\addons\rhs_heavyweapons\data\ico\rhs_2b14_82mm_msv_ca.paa";
+        picture = QPATHTOF(ui\rhs_2b14_82mm_msv.paa);
```
commy2: What?

commy2:
>RHS compat icons extracted from RHS. If you believe there might be an licensing issue, we can swap them for the generic icons in csw.

There is no reason to do this. Just use the external filepath.

Tbh, for legal reasons, these commits should be rebased.

Whigital:
> Why would static weapons appear in the arsenal?

Isn't the whole point of `CSW`, to have a crew that serves a weapon platform ?

What is the point of the weapon and tripod bags if you can't equip them from the arsenal. Maybe I've completely misunderstood the purpose of the module.


> There is no reason to do this. Just use the external filepath.

When equipped, you get an error when looking in the inventory (ex. `Warning Message: Picture \rhsusf\addons\rhsusf_heavyweapons\data\ico\rhs_m252_x_ca.paa not found`). I'm suspecting that Arma tries to find a higher/lower res version of the picture to display as icon in the inventory. Removing the `_co` suffix seems to disable this behavior, so i copied and renamed the .paa.

> Tbh, for legal reasons, these commits should be rebased.

I'll point them to the generic icons in `csw` if thats ok ?

commy2:
>I'll point them to the generic icons in csw if thats ok ?

No, use the RHS paths. If they changed them, adjust the paths. Obviously you have to start RHS to start these components.

addons/csw/CfgWeapons.hpp
commy2: This does make them be able to load the magazines, doesn't it?

Whigital: > 
> 
> > I'll point them to the generic icons in csw if thats ok ?
> 
> No, use the RHS paths. If they changed them, adjust the paths. Obviously you have to start RHS to start these components.

I did, of course. With the RHS paths, the icon/picture shows up in the arsenal ok, but when you exit and look in your inventory, the error pops up and the launcher slot is blank.

Whigital:
> No, use the RHS paths. If they changed them, adjust the paths. Obviously you have to start RHS to start these components.

I did, of course. With the RHS paths, the icon/picture shows up in the arsenal ok, but when you exit and look in your inventory, the error pops up and the launcher slot is blank.

addons/csw/CfgWeapons.hpp
Whigital: Yes, if you have it (the weapon bag) equipped, you can drag a mag to it, but nothing happens.

commy2: That's terrible.

commy2:
Does the picture error happen in the release version of ACE?

commy2:
I don't understand what `type = 4;` does here. That value should be inherited from LauncherCore.
Rule of thumb: type in public class -> you dun goofed or pulled something genius (unlikely)

Whigital:
> 
> 
> Does the picture error happen in the release version of ACE?

Yepp, just tested with SW versions ACE + USAF/AFRF/GREF + compats:

Equipped a `ace_compat_rhs_usf3_mk19_carry` in the arsenal, exit and open inventory.

`0:40:43 Warning Message: Picture \rhsusf\addons\rhsusf_heavyweapons\data\ico\rhs_mk19_tripod_x_ca.paa not found`

https://github.com/acemod/ACE3/blob/master/optionals/compat_rhs_usf3/CfgWeapons.hpp#L546

commy2:
None of the classnames are quoted.

commy2:
>Yepp, just tested with SW versions ACE + USAF/AFRF/GREF + compats:

Separate issue then. The files are not meant for weapon classes, thus the variants for attachments are missing.

Solution is to setup empty icons in WeaponSlotsInfo.

Whigital:
> 
> 
> I don't understand what `type = 4;` does here. That value should be inherited from LauncherCore.
> Rule of thumb: type in public class -> you dun goofed or pulled something genius (unlikely)

Correct, did a boo boo there. Looked at the wrong classes. The actual static weapons that has a corresponding vehicle for them have 65535. Will remove those declarations.

Whigital:
Most of the requested changes done now. Added a `MuzzleSlot` placeholder to make the engine use the supplied image based on this info -> https://community.bistudio.com/wiki/Arma_3_Weapon_Config_Guidelines#Inventory_icons_overlay_method

Image now shows in player inventory without any errors ....

addons/csw/CfgWeapons.hpp
Whigital: For a Titan/Spike AT, i guess its pretty obvious what kind of mag you need for it, but adding compatible mags to the array is the only way to know what ammo you need for the weapon. Otherwise you have to scroll through the "All Magazines" list and pray you pick up the right one.

We can remove them, but the convenience of having the compatible mags listed under Magazines maybe outweighs the fact you can "placebo load" the weapon with them ?

veteran29:
> Why would static weapons appear in the arsenal?

Without ACE CSW static weapons are available as backpack variants, I see no reason to not have launcher slot variants in it too. Otherwise to get launcher slot variants you would need to assemble backpack versions and disassemble them via CSW to get them.

Ideally you would have backpack versions in arsenal when CSW is disabled and opposite if enabled.

addons/csw/CfgWeapons.hpp
veteran29: Maybe consider adding custom properties for magwells and magazine classes. Content of these properties could be read by ACE Arsenal and shown on the magazine list.

dedmen:
> RHS compat icons extracted from RHS.

RHS will rip your ass up and potentially cause issues for us if you do that. So no no no no no.
> Tbh, for legal reasons, these commits should be rebased.

yes. Definitely. Please do that, right now the RHS icons are still in history. Should be fine if we just squash&merge at the end though, if author doesn't want to fix it.

Whigital:
Would you like me to remove the mag. defs. and just keep the other fixes, or just rebase as is ?

PabstMirror:
Tag @TheCandianVendingMachine

TheCandianVendingMachine:
@Whigital I would say remove the mag defs for right now. I'll take a look at ACE Arsenal to try and find a way to have compatible magazines show up if you have the launcher equipped, but being able to drag magazines into the launcher slot is hacky and I just don't like it

Whigital:
There we go, squashed, baked and magazines gone.

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -263,8 +278,9 @@ class CfgWeapons {
         modes[] = {};
         picture = "\rhsafrf\addons\rhs_heavyweapons\data\ico\rhs_SPG9_INS_ca.paa";
     };
+
     class GVAR(spg9m_carry): GVAR(spg9_carry) {
-        class ACE_CSW {
+        class ACE_CSW: ACE_CSW {
```
commy2: Why?

commy2:
Can you explain why:
```cpp
            class MuzzleSlot {
                iconScale = 0.1;
            };
```
works?

Why list iconScale at all?
Why is it not set to 0 or 1?
Why is not an empty picture set instead?
What about the other attachment slots?

optionals/compat_rhs_afrf3/CfgWeapons.hpp
Whigital: It didn't inherit the deploy and pickup attributes, so i added inheritance. Can hardcode them there instead.

Whigital:
> 
> 
> Can you explain why:
> 
> ```c++
>             class MuzzleSlot {
>                 iconScale = 0.1;
>             };
> ```
> 
> works?
> 
> Why list iconScale at all?
> Why is it not set to 0 or 1?
> Why is not an empty picture set instead?
> What about the other attachment slots?

Basically just to trick the engine into using the weapon picture instead of looking for alternatives based on slots:

https://community.bistudio.com/wiki/Arma_3_Weapon_Config_Guidelines#Inventory_icons_overlay_method

- If there is no slot with positive value of iconScale property, the game falls back to old behaviour

From what i could gather, the picture error is based on the engine looking for a picture with `<classname>_x_ca.paa` as baseLayer, `<classname>_m_ca.paa` for muzzle etc., if there are no slots with the `iconScale` property.

Can add the other slots and use an empty picture for them if it makes more sense.

commy2:
It wouldn't make more sense to add the other entries, however, I think this warrants a comment in the source code.

optionals/compat_rhs_afrf3/CfgWeapons.hpp
commy2: Ah, ok, so this was a different bug. Seems weird to me though. However, this spg9m_carry class is also missing the author attribute, which cannot be inherited, yet it does inherit public scope. There seem to be more problems with this class.

Whigital:
Maybe add a macro for that class (`MuzzleSlot`) and a comment that ?

```
// Fake muzzle slot to get weapon picture to display, ref
// https://community.bistudio.com/wiki/Arma_3_Weapon_Config_Guidelines#Inventory_icons_overlay_method

#define FAKE_MUZZLE_SLOT class MuzzleSlot { iconScale = 0.1; };
```

commy2:
Not sure about macro. It would be enough if the first occurance in the config is commented.

I'd chose a better comment though, i.e.:
```
// Positive value for iconScale forces game to use icon overlay method.
// Required, because the inventory icon has no accessory variants.
```

FAKE_MUZZLE_SLOT is a poor name, imo. One should use the self documenting possibilities when chosing a macro name. I.e. FIX_INVENTORY_ICON;

Issue #7403, opened 2020-01-05
Units stuck in unconscious animation when hit by physx object


diwako:
**Mods (complete and add to the following information):**
- **Arma 3:** `1.96.146114 Stable`
- **CBA:** `3.13.0` 
- **ACE3:** `3.13.0` 

**Description:**
When a unit is hit by a physx object and was forced to ragdoll, the unit it not able to get up again anymore

**Steps to reproduce:**
- Open eden
- Place 2 units and a quad bike
- preview
- ram the AI with the quad bike

**Expected behavior:**
The unit will be forced to ragdoll, after the recover animation, it will be stuck in the animation `unconscious` and will not get back up anymore. Happens to AI and players.

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) 

**Additional context:**
Some tries to fix the issue has been tried already. This one will work, however it will run on every unit, and in case the correct condition is met this eventhandler fires multiple times. So it possibly is not an ideal solution.

```sqf
["CAManBase", "AnimDone", {
    params ["_unit", "_anim"];
    if (!alive _unit || {!local _unit || {!(_anim isEqualTo "unconscious") || {_unit getVariable ["ACE_isUnconscious", false] || {lifeState _unit == "INCAPACITATED"}}}}) exitWith {};
    [_unit, false] call ace_medical_engine_fnc_setUnconsciousAnim;
}, true, [], true] call CBA_fnc_addClassEventHandler;
```

jonpas:
Can confirm, happened to me yesterday, getting out of a helicopter and hitting a rearm box someone else was holding just outside, flung 10m away, with all interactions stuck in that animation.

PabstMirror:
Should be fixed by #7499

PR #7499, opened 2020-01-30
Medical - Fix being stuck in unconscious animation after ragdoll

diwako:
- Get units out of unconscious animation after a physix object hit them

Addressing issue #7403

Note:
This code might not be fully optimal as it runs on every local unit. Thanks to commy2 for show me the way of propositional calculus.

https://streamable.com/com1h

### DIFF
--- a/addons/medical_engine/XEH_postInit.sqf
+++ b/addons/medical_engine/XEH_postInit.sqf
@@ -63,7 +63,7 @@
     } forEach (crew _vehicle);
 }, true, ["ParachuteBase"]] call CBA_fnc_addClassEventHandler;
 
-// fixes units being stuck in unconscious animation when being knocked over by a physix object
+// Fixes units being stuck in unconscious animation when being knocked over by a PhysX object
 ["CAManBase", "AnimDone", {
     params ["_unit", "_anim"];
     if (local _unit && {_anim == "unconscious" && {lifeState _unit != "INCAPACITATED"}}) then {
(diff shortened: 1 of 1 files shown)


addons/medical_engine/XEH_postInit.sqf
```diff
@@ -62,3 +62,11 @@
...
     } forEach (crew _vehicle);
 }, true, ["ParachuteBase"]] call CBA_fnc_addClassEventHandler;
+
+// fixes units being stuck in unconscious animation when being knocked over by a physix object
+["CAManBase", "AnimDone", {
+    params ["_unit", "_anim"];
+    if (local _unit && {_anim == "unconscious" && {lifeState _unit != "INCAPACITATED"}}) then {
+        [_unit, false] call FUNC(setUnconsciousAnim);
+    };
+}, true, [], true] call CBA_fnc_addClassEventHandler;
```
commy2:
```suggestion
}] call CBA_fnc_addClassEventHandler;
```
These are the default values, except "apply retroactively", but that is only relevant for init and init post events.

addons/medical_engine/XEH_postInit.sqf
```diff
@@ -62,3 +62,11 @@
         [QEGVAR(medical,woundReceived), [_x, "Head", _lethality, _killer, "#vehiclecrash", [HITPOINT_INDEX_HEAD,1]], _x] call CBA_fnc_targetEvent;
     } forEach (crew _vehicle);
 }, true, ["ParachuteBase"]] call CBA_fnc_addClassEventHandler;
+
+// fixes units being stuck in unconscious animation when being knocked over by a physix object
```
jonpas:
```suggestion
// fixes units being stuck in unconscious animation when being knocked over by a PhysX object
```

PabstMirror:
I think this will fix the problem, but for performance reasons I don't really like the anim EH being run on all units, but until we find anything better this is good for now

Issue #7356, opened 2020-01-02
Cannot interact with respawn gui when dying from unconsciousness


Arcanum417:
**Mods:**
- **Arma 3:** `1.96.146114` [e.g. 1.00 stable, rc,  dev]
- **CBA:** `3.13.0`
- **ACE3:** `3.13.0`
_Made sure to reproduce the issue with only CBA and ACE3 on a newly created mission!_

**Description:**
When dying from unconsciousness you cannot interact with respawn UI. Unconsciousness restrictions on iteraction do not end on death. I consider this a game breaking bug.

**Steps to reproduce:**
- Run arma with ACE3 and CBA only
- (optional) create new profile so you can run default ace settings
- Create new multiplayer mission in editor
- Place one BLUFOR unit, set variable name `test_sld`
- create one marker with cariable name `respawn_west`
- create one zeus game master module set `owner` as `test_sld`
- ensure you have default ACE settings set
- save the mission
- Create `description.ext` in the folder of the mission
- Paste this in description.ext
```
respawn = 2;
respawnDelay = 1;
respawnTemplates[] = {"MenuPosition","MenuInventory","Tickets","Spectator"};
```
- run the mission in listen server multiplayer
- click on respawn
- spawn BLUFOR AI from zeus
- shoot your character once in the chest (make him go unconscious but not outright dead)
- wait 10s
- shoot your character until it dies
- be presented with the respawn screen but be unable to click respawn or anything else

**Expected behavior:**
Be able to click respawn after dying from unconsciousness.

**Where did the issue occur?**
- Dedicated and Self-Hosted Multiplayer

zharf:
Are you sure this isn't because you were remote controlling another character when you died? Try pressing zeus key on the map again to see if that helps

Arcanum417:
Nope, also tested on a dedicated server without any zeus involvement (killed by another player the same way).

JANXOL:
Can confirm, also experienced the issue. Tested on a dedicated server with multiple people, no remote control involved. When shooting an uncosncious character and killing him, that player is unable to respawn or interact with UI, and has to be kicked and rejoin.

TheRealSpaceHippo:
I'm experiencing the same issue.

chazbotic:
my unit and i are also experiencing this issue. this issue occurs regardless if the player dies from cardiac arrest or is instantly killed (by script or applied damage from a very large explosion, for example).

[...]

PR #7502, opened 2020-02-01
[medical_status] Enable user input between death and respawn

Whigital:
- Enables user input before respawn so that the respawn interface can be interacted with.

Addresses https://github.com/acemod/ACE3/issues/7356

### DIFF
--- a/addons/medical_status/functions/fnc_handleKilled.sqf
+++ b/addons/medical_status/functions/fnc_handleKilled.sqf
@@ -38,4 +38,10 @@ TRACE_3("killer info",_killer,_instigator,_causeOfDeath);
 
 if (_unit isEqualTo (_unit getVariable [QGVAR(killed), objNull])) exitWith {}; // ensure event is only called once
 _unit setVariable [QGVAR(killed), _unit];
+
+if (_unit getVariable ["ACE_isUnconscious", false]) then {
+    // Enable user input before respawn, in case mission is using respawnTemplates
+    ["unconscious", false] call EGVAR(common,setDisableUserInputStatus);
+};
+
 ["ace_killed", [_unit, _causeOfDeath, _killer, _instigator]] call CBA_fnc_globalEvent;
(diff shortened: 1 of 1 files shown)


addons/medical_status/functions/fnc_handleKilled.sqf
```diff
@@ -38,4 +38,10 @@ TRACE_3("killer info",_killer,_instigator,_causeOfDeath);
 
 if (_unit isEqualTo (_unit getVariable [QGVAR(killed), objNull])) exitWith {}; // ensure event is only called once
 _unit setVariable [QGVAR(killed), _unit];
+
+if (_unit getVariable ["ACE_isUnconscious", false]) then {
+    // Enable user input before respawn, in case mission is using respawnTemplates
+    ["unconscious", false] call EGVAR(common,setDisableUserInputStatus);
```
mharis001: `EFUNC`.

Whigital: Ahh, not a var. My bad ....

commy2:
lgtm

commy2:
still lgtm

Silentbytex:
So i have this respawn problem in arma 3 and while looking for a fix to the respawn freeze i reached here. However i have no idea nor experience with code so im very confused with what is presented here. I went to ace>addons>medicalstatus(i opened the pbo file with notepad) and then i'm at a deadend because i don't know what to do. Can anyone help me out?

commy2:
This is a pull request, a change request to the code of ACE. After this pull request is merged and the next release published, the fix will be in the code of ACE.

You're lost and off topic.

Silentbytex:
Oh, ok. Thanks for the info.

PabstMirror:
What if `_unit` isn't player? (e.g. singleplayer)

SilentSpike:
Good point, not sure why that didn't cross my mind.

We could actually just replace the `IS_UNCONSCIOUS` condition with a check for whether the unit was the player to make sure input is always enabled on death. I think this should run for `ACE_Player` death not just `player`, right?

commy2:
Just `player` I believe.

SilentSpike:
Yeah think you're right on second though, believe there may be a player changed EH somewhere which will already handle enabling input again if a remote controlled unit dies

commy2:
This should be fixed before a release.

SilentSpike:
#7510

Issue #7448, opened 2020-01-11
Players able to interact while unconscious after they left cardiac arrest state


HexTFAC:
**Mods (complete and add to the following information):**
- **Arma 3:** `1.96` [Stable]
- **CBA:** `3.13.0` [Stable]
- **ACE3:** `3.13.0` [Stable]

**Description:**
As a player leaves the cardiac arrest state and enters the unconscious state, they are able to perform self-aid through the ACE3 Self-Interactions Menu

**Steps to reproduce:**
- Render player into cardiac arrest
- Solve for cardiac arrest
- Render into an unconscious state
- Allow player to observe self-aid options

**Expected behavior:**
Player should be unable to render aid to self while unconscious.

**Where did the issue occur?**
- Dedicated Multiplayer

**Log Files:**
N/A

**Additional context:**
N/A

**Screenshots:**
Video currently in production.

LockAndLoafCat:
I've had the same issue. My screen will even remain black sometimes and I can still self-interact.

Whigital:
https://github.com/acemod/ACE3/blob/5c774f5fe6557299c3ee0685137e01fd63fb8eed/addons/medical_statemachine/functions/fnc_leftStateCardiacArrest.sqf#L25

https://github.com/acemod/ACE3/blob/5c774f5fe6557299c3ee0685137e01fd63fb8eed/addons/medical_status/functions/fnc_setCardiacArrestState.sqf#L32

Shouldn't that be set to `true` when leaving `cardiacArrest` state ?

SilentSpike:
@Whigital That could very well be the cause this issue - well spotted. 

I think on leaving cardiac arrest we actually just shouldn't call that function at all, unconscious state doesn't change because it's only possible to go from cardiac arrest to death or unconsciousness.

SilentSpike:
This is now hopefully fixed on master, will close once confirmed

PabstMirror:
Should be fixed by #7503

PR #7503, opened 2020-02-01
[medical_status] Only set unconscious upon entering Cardiac Arrest

Whigital:
- Set unconsciousness only when entering Cardiac Arrest.

Addresses https://github.com/acemod/ACE3/issues/7448

### DIFF
--- a/addons/medical_status/functions/fnc_setCardiacArrestState.sqf
+++ b/addons/medical_status/functions/fnc_setCardiacArrestState.sqf
@@ -28,7 +28,9 @@ if (_active isEqualTo IN_CRDC_ARRST(_unit)) exitWith { TRACE_2("no change",_acti
 _unit setVariable [VAR_CRDC_ARRST, _active, true];
 _unit setVariable [VAR_HEART_RATE, [40, 0] select _active, true];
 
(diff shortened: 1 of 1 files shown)


Whigital:
There's a check for unconsciousness in `ace_medical_status_fnc_setUnconsciousState`,
https://github.com/acemod/ACE3/blob/2feeec0d7f252209c48a79760d2dc129e5e80655/addons/medical_status/functions/fnc_setUnconsciousState.sqf#L24

so adding something like `&& {_unit getVariable [VAR_UNCON, false]}` feels redundant, or ?

commy2:
Does it feel redundant or is it actually redundant? :)

SilentSpike:
Would be redundant yeah 👍

SilentSpike:
Have you tested that this is the cause of #7448?

The change should be included either way since it's an unnecessary call, just good to confirm

Whigital:
Didn't manage to reproduce with this fix in. Then again, i didn't try to reproduce with released code either so, inconclusive :confused:

commy2:
Would be nice to have confirmation that after this merged, the issue will be gone as well. Then I can tag this PR as "bug fix" and close the issue.

Issue #7429, opened 2020-01-08
ACE Medical: Taking damage from backblast even though allowDamage is false


leopard20:
allowDamage is set to false, but I take damage from backblast of ATs (tested RPG-42) when, e.g. my back is against a wall.

diwako:
Can confirm.
Over pressure (launcher and tank shots) is using function `addDamageToUnit`, which passes the damage over an event to function `woundsHandlerActive`
Both `woundsHandler` functions do not check if a unit can receive damage.

commy2:
addDamageToUnit and woundsHandler should not check for allowed damage, but the backblast function should.

diwako:
Ironically it checks for allowed damage if ACE medical is not present. I guess this was left in as in the previous medical version you could disable this damage by setting the variable `ace_medical_allowDamage` on a unit to `false`.

commy2:
Yeah, it seems like an oversight from changing the damage handling internals.

PR #7505, opened 2020-02-02
Backblast - Fix damage disabled units receiving damage from backblast

diwako:
- Will not damage units from over pressure if damage is disabled

Fix for #7429

### DIFF
--- a/addons/overpressure/functions/fnc_overpressureDamage.sqf
+++ b/addons/overpressure/functions/fnc_overpressureDamage.sqf
@@ -57,7 +57,7 @@ TRACE_3("cache",_overpressureAngle,_overpressureRange,_overpressureDamage);
             if (_x == ACE_player) then {[_damage * 100] call BIS_fnc_bloodEffect};
 
             TRACE_1("",isDamageAllowed _x);
-            if (isDamageAllowed _x) then {
+            if (isDamageAllowed _x && {_x getVariable [QEGVAR(medical,allowDamage), true]}) then {
                 if (isClass (configFile >> "CfgPatches" >> "ACE_Medical")) then {
                     [_x, _damage, "body", "backblast", _firer] call EFUNC(medical,addDamageToUnit);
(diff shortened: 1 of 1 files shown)


addons/overpressure/functions/fnc_overpressureDamage.sqf
```diff
@@ -56,12 +56,13 @@ TRACE_3("cache",_overpressureAngle,_overpressureRange,_overpressureDamage);
...
             if (_x == ACE_player) then {[_damage * 100] call BIS_fnc_bloodEffect};
 
-            if (isClass (configFile >> "CfgPatches" >> "ACE_Medical")) then {
-                [_x, _damage, "body", "backblast", _firer] call EFUNC(medical,addDamageToUnit);
-            } else {
-                TRACE_1("",isDamageAllowed _x);
-                if (!isDamageAllowed _x) exitWith {}; // Skip damage if not allowed
-                _x setDamage (damage _x + _damage);
+            TRACE_1("",isDamageAllowed _x);
+            if (isDamageAllowed _x) then {
```
jokoho48: I would also try to catch the ACE Medial variable
```suggestion
            if (isDamageAllowed _unit && {_unit getVariable [QEGVAR(medical,allowDamage), true]}) then {
```

commy2: That variable is never read by the game anywhere else as far as I can tell.

It is only set here:
https://github.com/acemod/ACE3/blob/master/addons/spectator/functions/fnc_setSpectator.sqf#L107
but that also should be deleted, as it does nothing.

What even is the point of that variable?

commy2: _unit -> _x

jokoho48: ups copy-pasted it from somewhere else :P

PR #7510, opened 2020-02-05
Only force enable input when player is killed

SilentSpike:
A quick hotfix for the implementation of #7502


commy2:
Is the function save to call on a unit that never was unconscious?

Whigital:
It removes the `_id` (`"unconscious"`) from `DISABLE_USER_INPUT_COLLECTION`. If it doesn't exist, the array is unchanged. Should be safe in my eyes.

https://github.com/acemod/ACE3/blob/cdf02f629e657cb6ca7263243df1bdad348a3d10/addons/common/functions/fnc_setDisableUserInputStatus.sqf#L30-L32

Whigital:
To be extra clear, perhaps add `&& {IS_UNCONSCIOUS(_unit)}` to the check, but would be redundant.

SilentSpike:
@commy2 Yeah safe to call and I like that this guarantees input will never be disabled due to ace unconsciousness state while dead

PR #7512, opened 2020-02-05
Map Gestures - Use object itself as center for nearEntities in getProximityPlayers

tbeswick96:
- Use the given unit as centre for `nearEntities` in `ace_map_gestures_fnc_getProximityPlayers`
- Previous `(getPos _unit)` would result in an empty return when stood in a building as height is reported as ~0, so the search performs at ground level
- `nearEntities` requires `PositionAGL`, `getPos` gives `PositionAGLS`


jokoho48:
lgtm

PR #7519, opened 2020-02-10
Map - Allow enabling BFT setting after mission start

shukari:
- change to CBA_fnc_addSetting not the internal fnc CBA_settings_fnc_init
- some vars can edited in the mission
- if BFT was disabled at mission start an den enabled it now works...

### DIFF
--- a/addons/map/initSettings.sqf
+++ b/addons/map/initSettings.sqf
@@ -85,7 +85,7 @@
     [0, 30, 1, 1],
     true,
     {[QGVAR(BFT_Interval), _this] call EFUNC(common,cbaSettings_settingChanged)},
-    false
+    true // Needs mission restart
 ] call CBA_fnc_addSetting;
 
 [
--- a/addons/map/initSettings.sqf
+++ b/addons/map/initSettings.sqf
@@ -69,8 +69,7 @@
     {
         [QGVAR(BFT_Enabled), _this] call EFUNC(common,cbaSettings_settingChanged);
         
-        if (isNil QGVAR(BFT_markers)) then
-        {
+        if (isNil QGVAR(BFT_markers)) then {
             GVAR(BFT_markers) = [];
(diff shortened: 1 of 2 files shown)


commy2:
Can you also disable it on the fly?

shukari:
@commy2 disable was always possible because if false:
https://github.com/acemod/ACE3/blob/fabe625d815fa99720144cfcabd2e4042db4f8f3/addons/map/functions/fnc_blueForceTrackingUpdate.sqf#L27

addons/map/initSettings.sqf
```diff
@@ -66,9 +66,17 @@
...
     false,
     true,
-    {[QGVAR(BFT_Enabled), _this] call EFUNC(common,cbaSettings_settingChanged)},
-    true // Needs mission restart
-] call CBA_settings_fnc_init;
+    {
+        [QGVAR(BFT_Enabled), _this] call EFUNC(common,cbaSettings_settingChanged);
+        
+        if (isNil QGVAR(BFT_markers)) then
+        {
```
synixebrett: Brackets should be on the same line as the `if` statement

commy2:
[This](https://github.com/acemod/ACE3/blob/master/addons/map/XEH_postInitClient.sqf#L27-L31) is not superfluous and should be deleted.

addons/map/initSettings.sqf
```diff
@@ -78,8 +85,8 @@
     [0, 30, 1, 1],
     true,
     {[QGVAR(BFT_Interval), _this] call EFUNC(common,cbaSettings_settingChanged)},
-    true // Needs mission restart
-] call CBA_settings_fnc_init;
+    false
```
PabstMirror: once pfeh is running, there is no way to change this
```suggestion
    true // Needs mission restart
```

PR #7520, opened 2020-02-11
Zeus - Added ace_cargoLoaded event

shukari:
- the load in module wasnt triggering the ace_cargoLoaded module

### Question
Why is the ace_cargoLoaded event not in the https://github.com/acemod/ACE3/blob/master/addons/cargo/functions/fnc_loadItem.sqf fnc?


shukari:
@SilentSpike can you review as ace_zeus maintainer?

jonpas:
I would prefer this to be reworked, but this works just fine as well.

SilentSpike:
I feel it would make more sense if we moved the listenable events in cargo into the loadItem function itself:

https://github.com/acemod/ACE3/blob/82d7980ada32b2a1bd1a90dee6226f204afa60cf/addons/cargo/XEH_postInit.sqf#L10-L21

That way there's only one origin point in the code for this event. Same for the unload one found below the above code snippet.

PabstMirror:
Can cleanup events later, merging for now

PR #7521, opened 2020-02-11
Medical - Improve injured sound by exiting the function earlier if no player is around

Vdauphin:
I notice `ace_medical_feedback_fnc_playInjuredSound` take an average time of execution around 300µs (~1000 call) even when no player is around the injured AI.

Sans titre

The idea of this PR is to reduce this value by exiting the function earlier.

  - Exit the  `ace_medical_feedback_fnc_playInjuredSound` earlier if no player is around
  - Get the hearing distance 
  - Check if players are around
  - Exit if no players

### DIFF
--- a/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
+++ b/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
@@ -34,7 +34,7 @@ private _distance = if (_type == "hit") then {
 } else {
     [10, 15, 20] select _severity;
 };
-private _targets = allPlayers inAreaArray [getPosWorld _unit, _distance, _distance];
+private _targets = allPlayers inAreaArray [getPosWorld _unit, _distance, _distance, 0, false, _distance];
 if (_targets isEqualTo []) exitWith {};
 
 // Handle timeout
--- a/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
+++ b/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
@@ -28,6 +28,15 @@ if (!local _unit) exitWith {
 };
 if !(_unit call EFUNC(common,isAwake)) exitWith {};
 
+// Limit network traffic by only sending the event to players who can potentially hear it
+private _distance = if (_type == "hit") then {
+    [50, 60, 70] select _severity;
+} else {
+    [10, 15, 20] select _severity;
+};
+private _targets = _unit nearEntities ["CAManBase", _distance];
+if (_targets isEqualTo []) exitWith {};
+
 // Handle timeout
 if (_unit getVariable [QGVAR(soundTimeout) + _type, -1] > CBA_missionTime) exitWith {};
 private _timeOut = if (_type == "moan") then { TIME_OUT_MOAN # _severity } else { TIME_OUT_HIT };
@@ -46,11 +55,6 @@ if (isNull (configFile >> "CfgSounds" >> format ["ACE_moan_%1_low_1", _speaker])
 
 // Select actual sound
 private _variation = ["low", "mid", "high"] select _severity;
-private _distance = if (_type == "hit") then {
-    [50, 60, 70] select _severity;
-} else {
-    [10, 15, 20] select _severity;
-};
 
 private _cfgSounds = configFile >> "CfgSounds";
 private _targetClass = format ["ACE_%1_%2_%3_", _type, _speaker, _variation];
@@ -63,6 +67,4 @@ while {isClass (_cfgSounds >> (_targetClass + str _index))} do {
 private _sound = configName selectRandom _sounds;
 if (isNil "_sound") exitWith { WARNING_1("no sounds for target [%1]",_targetClass); };
 
-// Limit network traffic by only sending the event to players who can potentially hear it
-private _targets = _unit nearEntities ["CAManBase", _distance];
 [QGVAR(forceSay3D), [_unit, _sound, _distance], _targets] call CBA_fnc_targetEvent;
--- a/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
+++ b/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
@@ -35,7 +35,7 @@ private _distance = if (_type == "hit") then {
     [10, 15, 20] select _severity;
 };
 private _targets = _unit nearEntities ["CAManBase", _distance];
-if ((_targets findIf {isPlayer _x}) == -1) exitWith {};
+if (_targets findIf {isPlayer _x} == -1) exitWith {};
 
 // Handle timeout
 if (_unit getVariable [QGVAR(soundTimeout) + _type, -1] > CBA_missionTime) exitWith {};
--- a/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
+++ b/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
@@ -35,7 +35,7 @@ private _distance = if (_type == "hit") then {
     [10, 15, 20] select _severity;
 };
 private _targets = _unit nearEntities ["CAManBase", _distance];
-if (_targets isEqualTo []) exitWith {};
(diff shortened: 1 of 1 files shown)


addons/medical_feedback/functions/fnc_playInjuredSound.sqf
```diff
@@ -28,6 +28,15 @@ if (!local _unit) exitWith {
...
 if !(_unit call EFUNC(common,isAwake)) exitWith {};
 
+// Limit network traffic by only sending the event to players who can potentially hear it
+private _distance = if (_type == "hit") then {
+    [50, 60, 70] select _severity;
+} else {
+    [10, 15, 20] select _severity;
+};
+private _targets = _unit nearEntities ["CAManBase", _distance];
+if (_targets isEqualTo []) exitWith {};
```
commy2: There always is at least one unit near, the unit that is hurt. Also, the avatar should still hear himself moan, even if there is no one else nearby.

commy2: `nearEntities` also completely fails to pick up units inside vehicles.

commy2:
This does not seem like a good idea to me.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
Vdauphin: The main idea here, is to check if player is around. 
Then something like this should fit the intended behavior:
`allPlayers inAreaArray [getPosWorld _unit, _distance/2, _distance/2]`

PabstMirror: `if ((_targets findIf {isPlayer _x}) == -1) exitWith {};`

jonpas: > There always is at least one unit near, the unit that is hurt.

This is about AI moaning if I understand correctly.

> Also, the avatar should still hear himself moan, even if there is no one else nearby.

AI can't hear itself.

commy2: jonpas, this script is not for AI only, it is for everyone.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
```diff
@@ -28,6 +28,15 @@ if (!local _unit) exitWith {
...
 if !(_unit call EFUNC(common,isAwake)) exitWith {};
 
+// Limit network traffic by only sending the event to players who can potentially hear it
+private _distance = if (_type == "hit") then {
+    [50, 60, 70] select _severity;
+} else {
+    [10, 15, 20] select _severity;
+};
+private _targets = _unit nearEntities ["CAManBase", _distance];
+if ((_targets findIf {isPlayer _x}) == -1) exitWith {};
```
commy2:
```suggestion
if (_targets findIf {isPlayer _x} == -1) exitWith {};
```

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
jonpas: I know, but the goal of the PR (if I understood correctly) is to disable it when an AI moans but no player is in range to hear it.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
```diff
@@ -28,6 +28,15 @@ if (!local _unit) exitWith {
...
 if !(_unit call EFUNC(common,isAwake)) exitWith {};
 
+// Limit network traffic by only sending the event to players who can potentially hear it
+private _distance = if (_type == "hit") then {
+    [50, 60, 70] select _severity;
+} else {
+    [10, 15, 20] select _severity;
+};
+private _targets = _unit nearEntities ["CAManBase", _distance];
+if (_targets findIf {isPlayer _x} == -1) exitWith {};
```
commy2: Does not recognize vehicle crew.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
commy2: And the line I commented on does not work that way. This argument is stupid.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
Vdauphin: So the 
```sqf
private _targets = allPlayers inAreaArray [getPosWorld _unit, _distance, _distance]
```
is the solution ?

If any player is inside the area then the `_targets` will not be empty. This also handle the case where player is inside a vehicle or `_unit` is a player.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
dedmen: are you sure that nearEntities is cheaper, than just paying the 300us cost to let the function run?

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
dedmen: That sounds better yes. But I'm not sure if this actually improved performance

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
Vdauphin: It is, I checked 
You can see in the screen the time of the `nearEntities` in purple
Sans titre
If I remembrer it is around 50µs in this case.

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
```diff
@@ -28,6 +28,15 @@ if (!local _unit) exitWith {
 };
 if !(_unit call EFUNC(common,isAwake)) exitWith {};
 
+// Limit network traffic by only sending the event to players who can potentially hear it
+private _distance = if (_type == "hit") then {
+    [50, 60, 70] select _severity;
+} else {
+    [10, 15, 20] select _severity;
+};
+private _targets = allPlayers inAreaArray [getPosWorld _unit, _distance, _distance];
```
PabstMirror: `allPlayers` is a really shitty command and is often not synced correctly
it can take minutes after connecting for someone to show up in the list
tbh, that's probably ok as groans aren't super critical

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
commy2: Perf of `nearEntities` is irrelevant. Unless you do shenigans with `crew`, the command is unusable for this, as I said already.

PabstMirror: isn't filtering out crew desirable?
people inside of a tank shouldn't hear the cries of enemies lubricating their treads

dedmen: > You can see in the screen the time of the nearEntities in purple

*squint* sooo teeeny

addons/medical_feedback/functions/fnc_playInjuredSound.sqf
PabstMirror: 56 units, 24 in-range
```
allUnits select {(_x distance player) < 100};    0.0657 ms
allUnits inAreaArray [getPos player, 100, 100];  0.0085 ms

File: addons/medical_feedback/functions/fnc_playInjuredSound.sqf (hunk 4 above)
Comment (PabstMirror) [reply addons/medical_feedback/functions/fnc_playInjuredSound.sqf:37]: This should handle z axis, but so rare it's not really needed
```suggestion
private _targets = allPlayers inAreaArray [getPosWorld _unit, _distance, _distance, 0, false, _distance];
```

commy2:
I don't get the handling z part. We don't want to mute all moans if you are 70 or whatever meters above the ground. Just leave the z out and it will handle all untis in the cylinder.

PabstMirror:
it's a distance check between the audio source and a player
adding the extra params makes it a 3d sphere search, instead of a 2d cylinder

PR #7522, opened 2020-02-11
Docs/Wiki - Fix formatting for BIS evh warning

Whigital:
- Fix formatting for `BIS event handlers warning` under section `7.3`.

Some HTML snuck in there, breaking the markdown.


jonpas:
HTML didn't snuck in there, it creates the "note" frame. But that doesn't support coding blocks, so I guess this is fine too.

Issue #7528, opened 2020-02-15
Medical - `setVolume` not called on unconscious


jonpas:
**Mods (complete and add to the following information):**
- **Arma 3:** `1.96`
- **CBA:** `3.14.0`
- **ACE3:** `3.13.0`

**Description:**
ACRE2 and TFAR volume is not muffled/lowered on unconscious.

**Steps to reproduce:**
- Go unconscious (medical rewrite).
- Listen.

**Expected behavior:**
Voice should be muffled/lowered by [ace_common_fnc_setVolume](https://github.com/acemod/ACE3/blob/e2ac18a05d5f646bc7cbc043bcc148fde4c0f5dd/addons/common/functions/fnc_setVolume.sqf), but this function never gets called from medical.

**Where did the issue occur?**
- Dedicated

**Log Files:**
- N/A

**Additional context:**
https://github.com/IDI-Systems/acre2/issues/882

**Screenshots:**
/

PR #7529, opened 2020-02-15
Medical Feedback - Set volume and radio usage based current unit

Whigital:
- Set players volume based on unconscious state

Fix https://github.com/acemod/ACE3/issues/7528

### DIFF
--- a/addons/medical_feedback/XEH_postInit.sqf
+++ b/addons/medical_feedback/XEH_postInit.sqf
@@ -43,14 +43,7 @@ GVAR(bloodTickCounter) = 0;
         ACE_player switchCamera "INTERNAL";
     };
 
-    // Toggle unconscious player's ability to talk in radio addons
-    if (["task_force_radio"] call EFUNC(common,isModLoaded)) then {
-        _unit setVariable ["tf_voiceVolume", [1, 0] select _unconscious, true];
-        _unit setVariable ["tf_unable_to_use_radio", _unconscious]; // Only used locally
-    };
-    if (["acre_main"] call EFUNC(common,isModLoaded)) then {
-        _unit setVariable ["acre_sys_core_isDisabled", _unconscious, true];
-    };
+    [!_unconscious, _unit] call EFUNC(common,setVolume);
 
     // Greatly reduce player's hearing ability while unconscious (affects radio addons)
     [QUOTE(ADDON), VOL_UNCONSCIOUS, _unconscious] call EFUNC(common,setHearingCapability);
@@ -64,13 +57,9 @@ GVAR(bloodTickCounter) = 0;
     params ["_unit"];
 
     if (_unit != ACE_player) exitWith {};
-    if (["task_force_radio"] call EFUNC(common,isModLoaded)) then {
-        _unit setVariable ["tf_voiceVolume", 1, true];
-        _unit setVariable ["tf_unable_to_use_radio", false];
-    };
-    if (["acre_main"] call EFUNC(common,isModLoaded)) then {
-        _unit setVariable ["acre_sys_core_isDisabled", false, true];
-    };
+
+    [true, _unit] call EFUNC(common,setVolume);
+
     [QUOTE(ADDON), 1, false] call EFUNC(common,setHearingCapability);
 }] call CBA_fnc_addEventHandler;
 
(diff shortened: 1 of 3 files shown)


addons/medical_status/functions/fnc_setUnconsciousState.sqf
```diff
@@ -31,6 +31,11 @@ _unit setVariable [VAR_UNCON, _active, true];
 // Stop AI firing at unconscious units in most situations (global effect)
 [_unit, "setHidden", "ace_unconscious", _active] call EFUNC(common,statusEffect_set);
 
+// Set players volume according to unconscious state
+if (_unit == ACE_player) then {
+    !_active call EFUNC(common,setVolume);
```
jonpas:
```suggestion
    [!_active] call EFUNC(common,setVolume);
```

SilentSpike:
I don't think this is such a simple fix as assumed here because the unconscious state of the local machine can change without hitting the setUnconsciousState function (for example, upon death or upon unit switching).

See: https://github.com/acemod/ACE3/blob/5a4487f1eae7173d9551d2df9eff7f0ca675aca4/addons/medical_feedback/XEH_postInit.sqf#L35-L92

Whigital:
Moving the `EFUNC(common,setVolume)` call from `ace_medical_status_fnc_setUnconsciousState` -> `medical_feedback/XEH_postInit` (`ace_unconscious` EH and `unit` playerEH) should suffice ?

From what i can see, the only place the event gets raised is at the end of `ace_medical_status_fnc_setUnconsciousState`.

Volume gets restored upon respawn, so might not be necessary on `ace_medical_death` event ?
https://github.com/acemod/ACE3/blob/5a4487f1eae7173d9551d2df9eff7f0ca675aca4/addons/common/functions/fnc_resetAllDefaults.sqf#L24

SilentSpike:
@Whigital I would include on death because death does not always immediately lead to respawn and could be filled with other systems (most commonly spectator time). 

However, on death it would only apply when the `_unit == player` since `setVolume` updates the player's sound specifically (and death of any `ACE_player !== player` results in a unit change back to `player` which triggers the unit event).

In fact, it should probably be clarified how TFAR and ACRE handle the unit variables that function is setting (always on `player`), I would assume that they only apply for that unit and those addons aren't just using `player` as a storage vessel. If they are unit specific then that function would be much more useful for us here if you could pass in a unit parameter.

jonpas:
ACRE uses a similar system to ace_player.

SilentSpike:
Figured as much, then the way I would fix this is add an option parameter to the common function to pass in a unit, and just update all three events in feedback postInit to call that function and pass in the unit if it's `ace_player`

Whigital:
So update `ace_common_fnc_setVolume` with something like
```
params [
    ["_setVolume", false],
    ["_unit", player]
];
```
so it still defaults to player, and use `[<true|false>, ACE_player] call EFUNC(common,setVolume)` for each of the EH's in `medical_feedback/XEH_postInit` ?

Off on a tangent here, but perhaps also update the seVars in `ace_common_fnc_setVolume` with `acre_player` and `TFAR_currentUnit` instead of `player` ?

SilentSpike:
That sound right 👍 

Only thing is the `"unit"` event `_new` is `ACE_player` by definition so no need to check that. Also note you can replace the existing lines which set TFAR and ACRE variables in all those EHs.

Whigital:
Beautify the setVolume function and get rid of the if statement ?
```
private _vol = ([MUTED_LEVEL, NORMAL_LEVEL] select _setVolume);
private _voice = ([NO_SOUND, NORMAL_LEVEL] select _setVolume);

2 fadeSound _vol;

//TFAR
_unit setVariable ["tf_voiceVolume", _voice , true];
_unit setVariable ["tf_globalVolume", _vol];
_unit setVariable ["tf_unable_to_use_radio", !_setVolume];

// ACRE2
if (!isNil "acre_api_fnc_setGlobalVolume") then { [_vol^0.33] call acre_api_fnc_setGlobalVolume; };
_unit setVariable ["acre_sys_core_isDisabled", !_setVolume, true];
```

SilentSpike:
Feel free to do so, current changes look correct to me

I also notice we're calling the `setHearingCapability` function which looks like it has some overlap with this setVolume function (both interact with `fadeSound`, `tf_globalVolume` and `acre_api_fnc_setGlobalVolume`).

This area of the code could probably use some cleanup (would avoid that for this PR and just focus on the fix, just noting for other ace devs).

Whigital:
Roger. Will keep the changes in this PR to just fixing the issue at hand.

Whigital:
Think i'll leave this PR as is. Will verify functionality and report back.

Whigital:
Toggling unconscious lowers volume and disables the use of ACRE radio without issues. Not been able to test in MP, but *should* work the same, considering the previous code just got moved from the EH's to `ace_common_fnc_setVolume`.

addons/medical_feedback/XEH_postInit.sqf
```diff
@@ -64,13 +57,9 @@ GVAR(bloodTickCounter) = 0;
...
 
     if (_unit != ACE_player) exitWith {};
-    if (["task_force_radio"] call EFUNC(common,isModLoaded)) then {
-        _unit setVariable ["tf_voiceVolume", 1, true];
-        _unit setVariable ["tf_unable_to_use_radio", false];
-    };
-    if (["acre_main"] call EFUNC(common,isModLoaded)) then {
-        _unit setVariable ["acre_sys_core_isDisabled", false, true];
-    };
+
```
SilentSpike:
```suggestion
    // Players always able to hear for any systems that might run while dead (e.g. spectator)
```

PR #7531, opened 2020-02-16
fix double execution of killed event

commy2:
- I can't reproduce the error if I don't return values when the unit is dead.

Don't ask why this works.

### DIFF
--- a/addons/medical_engine/XEH_postInit.sqf
+++ b/addons/medical_engine/XEH_postInit.sqf
@@ -24,7 +24,10 @@
         // "exitWith", which would be broken with "HandleDamage" otherwise.
         _unit setVariable [
             QEGVAR(medical,HandleDamageEHID),
-            _unit addEventHandler ["HandleDamage", {_this call FUNC(handleDamage)}]
+            _unit addEventHandler ["HandleDamage", {
+                private _return = _this call FUNC(handleDamage);
+                if (alive _unit) then {_return};
+            }]
         ];
     };
 }, nil, [IGNORE_BASE_UAVPILOTS], true] call CBA_fnc_addClassEventHandler;
--- a/addons/medical_engine/XEH_postInit.sqf
+++ b/addons/medical_engine/XEH_postInit.sqf
@@ -27,7 +27,7 @@
             _unit addEventHandler ["HandleDamage", {
                 // Not returning anything for dead units prevents double execution of Killed events.
                 private _return = _this call FUNC(handleDamage);
-                if (alive _unit) then {_return};
+                if (alive param [0]) then {_return};
             }]
         ];
     };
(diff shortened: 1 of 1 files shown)


commy2:
Please test with:
```sqf
["CAManBase", "Init", { 
    params ["_unit"]; 
 
    _unit addEventHandler ["Killed", { 
        systemChat str _this;
    }]; 
}, nil, nil, true] call CBA_fnc_addClassEventHandler; 
```

, because:
```sqf
["CAManBase", "Killed", { 
    systemChat str _this;
```
is already covered by https://github.com/CBATeam/CBA_A3/pull/1291

jonpas:
Write a comment directly in the code, so it won't get cleaned up one day.

commy2:
Done.

PabstMirror:
Seems to report correct killer, will do mp testing soon

addons/medical_engine/XEH_postInit.sqf
```diff
@@ -24,7 +24,11 @@
         // "exitWith", which would be broken with "HandleDamage" otherwise.
         _unit setVariable [
             QEGVAR(medical,HandleDamageEHID),
-            _unit addEventHandler ["HandleDamage", {_this call FUNC(handleDamage)}]
+            _unit addEventHandler ["HandleDamage", {
+                // Not returning anything for dead units prevents double execution of Killed events.
+                private _return = _this call FUNC(handleDamage);
+                if (alive _unit) then {_return};
```
mharis001: `_unit` is `nil` in this scope.

PabstMirror: Sadly, this is what was making it "work", it our EH always just returned nil; which meant the unit just died from vanilla medical.

PabstMirror:
Still getting double killed EH 
```
Mission [47792]: ace_medical_death event [id: 1] - [a2]
"calling setStructuralDamage (setDamage 1)"
Mission [47792]: ace_medical_woundReceived event [id: 1] - [a2,"Head",3.30751,a1,"B_65x39_Caseless",[0,3.30751,1,0,2,0.0232521,3,0.111614,4,0,5,0]]
Mission [47792]: ace_killed event [id: 0] - [a2,"FatalInjury:Death",a1,a1]
Mission [47792]: BIS addEventHandler killed [id: 1] - [a2,a2,<NULL-object>,true]
Mission [47792]: CBA addClassEventHandler killed [id: any] - [a2,a2,<NULL-object>,true]
Mission [47792]: BIS addEventHandler killed [id: 1] - [a2,a1,a1,true] // (modifyKilledArray disabled in handleDamage)
```

commy2:
@PabstMirror What does it report with this?

PabstMirror:
both of these caused double killed
```
private _return = _this call FUNC(handleDamage);
if (!alive param [0]) exitWith {diag_log "ret nil"; nil};
_return
```

```
private _return = _this call FUNC(handleDamage);
if (alive param [0]) then { _return } else {nil};
```

Issue #6948, opened 2019-04-18
Very high AI accuracy caused by ACE


BlackAlpha:
**Arma 3 Version:** 1.90.145471 Stable
**CBA Version:** 3.10.1 (Steam Workshop)
**ACE3 Version:** 3.12.6 (Steam Workshop)

**Mods:**
- CBA_A3
- ace

**Description:**
The AI accuracy is very high with the ACE mod and it's not possible to give the AI low accuracy. This is a problem if, for example, you want to do a mission with low AI skill or you are running a suppression script that uses setSkill to modify AI accuracy.

It would be very, very helpful if the ACE AI accuracy was more similar to vanilla, allowing for more granular control over the accuracy setting.

**Steps to reproduce:**
- Download this test mission: https://www.dropbox.com/s/j46m247dd9b1n5w/aiAccuracyTest1.VR.rar?dl=0
- Start vanilla and load up the mission in the editor.
- Don't do anything and allow the AI to fire 200+ shots and take note of the results.
- Use the scroll wheel to change the AI accuracy skill and repeat the test.
- Repeat the test with the ACE mod. The AI Accuracy is tremendously higher with the ACE mod, super-human level.

I've done many tests and the following numbers give you a pretty good picture:

Distance: 300 meters
Weapon: Rifle without magnification
Shooter position: Prone on elevated platform
Target position: Prone on ground
Profile accuracy level: 0.5
Setskills: 0.9

Vanilla
Accuracy Skill: 0.9
Accuracy SkillFinal: 0.8125
Shots fired: 200
Hits: 47
Accuracy: 23.5%

Vanilla
Accuracy Skill: 0.1
Accuracy SkillFinal: 0.1
Shots fired: 200
Hits: 7
Accuracy: 3.5%

Vanilla
Accuracy Skill: 0.01
Accuracy SkillFinal: 0.099
Shots fired: 200
Hits: 4
Accuracy: 2%

ACE
Accuracy Skill: 0.9
Accuracy SkillFinal: 0.8125
Shots fired: 200
Hits: 182
Accuracy: 91%

ACE
Accuracy Skill: 0.1
Accuracy SkillFinal: 0.1
Shots fired: 200
Hits: 71
Accuracy: 35.5%

ACE
Accuracy Skill: 0.01
Accuracy SkillFinal: 0.099
Shots fired: 200
Hits: 60
Accuracy: 30%

**Where did the issue occur?**
- Dedicated server / Editor (Singleplayer)

[...]

PR #7532, opened 2020-02-17
Ballistics - Fix AI Accuracy on weapons

Freddo3000:
- Change `aiDispersionCoefX/Y` values to suit the lowered dispersion values set by Ballistics
- Change dispersion and aiDispersionCoefX/Y to `__EVAL()`

Related ticket: https://github.com/acemod/ACE3/issues/6948  
Possibly related pull: https://github.com/acemod/ACE3/pull/4484

<details>
<summary>Notes:</summary>

Comparing Vanilla and ACE hit ratios, using a Katiba without optic on Regular difficulty, 0.9 skill on all sub-skills, AI hit ~24% of shots.  
With ACE Ballistics, same scenario, AI hit ~86% of shots.
[Testing scenario](https://dropbox.com/s/j46m247dd9b1n5w/aiAccuracyTest1.VR.rar?dl=0)

I found this to be caused by that the `aiDispersionCoefX/Y` values to be left at default, on both firemode and weapon class.

To calculate new values I went with this assumption:  
`dispersion * aiDispersionCoef = effective dispersion`

Using that assumption I calculate `aiDispersionCoefX/Y` using:  
`(oldDispersion * oldAIDispersionCoef)/newDispersion = newAIDispersionCoef`

**Example GM6 Lynx:**  
Without ace: 0.00035 * 1.7 = 0.000595 effective dispersion
With ace: 0.00013 * 1.7 = 0.000221 effective dispersion, lower than the default dispersion

**With changes:**

((0.00035 * 1.7)/0.00013) = 4.577 new aiDispersionCoef  
(0.00035 * 1.7) = 0.000595 effective dispersion // Default  
(0.00013 * 4.577) = 0.000595 effective dispersion  // ACE w.Changes

The same formula was used on both firemode and weapon class, I'm not entirely sure how exactly they link to weapon accuracy, but both do appear to affect it. If I were to guess, it goes something along the lines of `weaponDispersionCoef * firemodeDispersionCoef * dispersion = effective dispersion`. [Relevant](https://forums.bohemia.net/forums/topic/150499-ai-discussion-dev-branch/?do=findComment&comment=3289475)

Changing values to be within `__EVAL()`s instead of macros appeared to lower accuracy by ~5%, though that might have been due to testing error.

**Result:**  
Using a Katiba without optic on Regular difficulty, 0.9 skill on all sub-skills, AI hit ~32% of shots, higher than vanilla but much less than 86%.

</details>

### DIFF
--- a/addons/main/script_macros.hpp
+++ b/addons/main/script_macros.hpp
@@ -134,7 +134,9 @@
 #define MRAD_TO_DEG(d) ((d) / 17.45329252) // Conversion factor: 9 / (50 * PI)
 #define MOA_TO_RAD(d) ((d) * 0.00029088) // Conversion factor: PI / 10800
 
-#define AI_DISPERSION(newDispersion,oldDispersion,oldCoef) ((oldDispersion * oldCoef) / (newDispersion))
+#define AI_DISPERSION(newDispersion,oldDispersion,oldCoefY,oldCoefX)\
+aiDispersionCoefX = (((oldDispersion) * (oldCoefX)) / (newDispersion));\
+aiDispersionCoefY = (((oldDispersion) * (oldCoefY)) / (newDispersion))\
 
 #define ZEUS_ACTION_CONDITION ([_target, {QUOTE(QUOTE(ADDON)) in curatorAddons _this}, missionNamespace, QUOTE(QGVAR(zeusCheck)), 1E11, 'ace_interactMenuClosed'] call EFUNC(common,cachedCall))
 
--- a/addons/main/script_macros.hpp
+++ b/addons/main/script_macros.hpp
@@ -133,10 +133,8 @@
 #define DEG_TO_MRAD(d) ((d) * 17.45329252) // Conversion factor: (50 * PI) / 9
 #define MRAD_TO_DEG(d) ((d) / 17.45329252) // Conversion factor: 9 / (50 * PI)
 #define MOA_TO_RAD(d) ((d) * 0.00029088) // Conversion factor: PI / 10800
-#define EVAL_MOA_TO_RAD(d) __EVAL(MOA_TO_RAD(d))
 
 #define AI_DISPERSION(newDispersion,oldDispersion,oldCoef) ((oldDispersion * oldCoef) / (newDispersion))
-#define EVAL_AI_DISPERSION(newDispersion,oldDispersion,oldCoef) __EVAL(AI_DISPERSION(newDispersion,oldDispersion,oldCoef))
 
 #define ZEUS_ACTION_CONDITION ([_target, {QUOTE(QUOTE(ADDON)) in curatorAddons _this}, missionNamespace, QUOTE(QGVAR(zeusCheck)), 1E11, 'ace_interactMenuClosed'] call EFUNC(common,cachedCall))
 
--- a/addons/main/script_macros.hpp
+++ b/addons/main/script_macros.hpp
@@ -135,8 +135,8 @@
 #define MOA_TO_RAD(d) ((d) * 0.00029088) // Conversion factor: PI / 10800
 
 #define AI_DISPERSION(newDispersion,oldDispersion,oldCoefY,oldCoefX)\
-aiDispersionCoefX = (((oldDispersion) * (oldCoefX)) / (newDispersion));\
-aiDispersionCoefY = (((oldDispersion) * (oldCoefY)) / (newDispersion))\
+aiDispersionCoefX = (oldDispersion) * (oldCoefX) / (newDispersion);\
+aiDispersionCoefY = (oldDispersion) * (oldCoefY) / (newDispersion)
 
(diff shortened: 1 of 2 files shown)


commy2:
Why __EVAL? Does __EVAL even work when binarized? I.e. Armake/make.py instead of build.py.

Freddo3000:
> Why __EVAL? Does __EVAL even work when binarized? I.e. Armake/make.py instead of build.py.

I did it that way because it looks nicer in the config viewer, being able to get a direct reading and also to narrow it down in case it didn't like strings for some reason. I did observer a 5% reduction in accuracy as mentioned, though that might just be due to random chance.

If it works using make.py I'm not sure, as I seemingly can't get it to run, build.py works however.

commy2:
It won't. build.py just packs, which is not good enough for release builds. EVAL adds nothing of value anyway. To make sure the string is read as expression, wrap it into call:
```
entry = QUOTE(call {/stuff/})
```
and test if dispersion is different from 0 in practice (ignore getNumber and config viewer for this).

commy2:
```hpp
 #define AI_DISPERSION(newDispersion,oldDispersion,oldX,oldY)\
    aiDispersionCoefX = (oldDispersion) * (oldX) / (newDispersion);\
    aiDispersionCoefY = (oldDispersion) * (oldY) / (newDispersion)
```
```hpp
    dispersion = MOA_TO_RAD(0.45);
    AI_DISPERSION(MOA_TO_RAD(0.45),0.00035,1.4,1.7)
```

or
```hpp
    #define DISPERSION 0.45
    dispersion = MOA_TO_RAD(DISPERSION);
    AI_DISPERSION(MOA_TO_RAD(DISPERSION),0.00035,1.4,1.7)
```
?

addons/main/script_macros.hpp
```diff
@@ -134,7 +134,9 @@
 #define MRAD_TO_DEG(d) ((d) / 17.45329252) // Conversion factor: 9 / (50 * PI)
 #define MOA_TO_RAD(d) ((d) * 0.00029088) // Conversion factor: PI / 10800
 
-#define AI_DISPERSION(newDispersion,oldDispersion,oldCoef) ((oldDispersion * oldCoef) / (newDispersion))
+#define AI_DISPERSION(newDispersion,oldDispersion,oldCoefY,oldCoefX)\
+aiDispersionCoefX = (((oldDispersion) * (oldCoefX)) / (newDispersion));\
+aiDispersionCoefY = (((oldDispersion) * (oldCoefY)) / (newDispersion))\
```
commy2: You don't need the parentheses not around single vars, because this expression is closed.

commy2: The last escaped newline is wrong. May break if the line below is deleted during a reformatting. In any case, you just ended a line without ; which is logged as warning in RPT.

Freddo3000: Noticed and fixed with last commit.

commy2:
lgtm
Can you vouch for correct numbers?

Freddo3000:
AI Accuracy is ~30% higher than what it should be, but unless you want to throw in a magic number in there then I think this is about the best that we can come up with. Certainly much better than what it is currently, with a more than 100% accuracy increase.

Unless there is a typo somewhere, it should all be correct.

jonpas:
~30% higher is good. We are used to 100% higher anyways. Could this in any way conflict with AI mods though?

Freddo3000:
> 
> 
> ~30% higher is good. We are used to 100% higher anyways. Could this in any way conflict with AI mods though?

It will conflict with mods changing aiDispersionCoef on these weapons and don't load after ACE, and possibly mods inheriting from these weapons.

jonpas:
Yes it goes for all, but we decided to not touch AI as much as possible in ACE3.

However, since this is closer to restoring vanilla, I think it would be fine.

commy2:
I mean, we could always remove all the config stuff and do the dispersion using changProjectileDirection. I am always for script > config in cases like this that have a lot of interactions.

Freddo3000:
> 
> 
> I mean, we could always remove all the config stuff and do the dispersion using changProjectileDirection. I am always for script > config in cases like this that have a lot of interactions.

Well until that solution is ready, this is a good enough stopgap solution other than outright removing the dispersion changes.

commy2:
I can write that in one hour  (next weekend).

Freddo3000:
Perhaps merge this then and remove it with a new pull request when its ready, so it doesn't just happen to fall to the side.

Laid3acK:
The problem isn't the `aiDispersionCoefX/Y`, the problem is the rifles' `dispersion` itself.
Why not revert to the vanilla values for all rifles, except maybe the marksman and sniper rifles ?

commy2:
I'd personally revert all rifles to vanilla and then consider a scripted solution for certain marksman and sniper rifles. Idk who decides here though.

jonpas:
2 options:
- Remove `dispersion` all-together, because it influences AI.
- Move `dispersion` changes to an optional. (recommended)

Will be included in 3.13.1 if done by release, otherwise future.

commy2:
Close?

jonpas:
Moved to optional in #7557 and #7558.

PR #7533, opened 2020-02-17
Medical - Unlock air controls when going uncon

PabstMirror:
When pilot of aircraft goes uncon then do "Unlock controls" user action 
(co-pilot will then have to do the "Take Controls" actions)

Recently saw a need for this 😄 
Could add a setting (or hidden getVar setting), don't really think it needs one.

### DIFF
--- a/addons/medical_status/functions/fnc_setUnconsciousState.sqf
+++ b/addons/medical_status/functions/fnc_setUnconsciousState.sqf
@@ -47,7 +47,7 @@ if (_active) then {
         };
     };
     // Unlock controls for copilot if unit is pilot of aircraft
-    if (((vehicle _unit) isKindOf "Air") && {_unit == driver vehicle _unit}) then {
+    if (vehicle _unit isKindOf "Air" && {_unit == driver vehicle _unit}) then {
         TRACE_1("pilot of air vehicle - unlocking controls",vehicle _unit);
         // Do "Unlock controls" user action, co-pilot will then have to do the "Take Controls" actions
         _unit action ["UnlockVehicleControl", vehicle _unit];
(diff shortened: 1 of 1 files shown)


jonpas:
This is great! No more need to remember to do that every time I have a copilot. Never thought of it though.

addons/medical_status/functions/fnc_setUnconsciousState.sqf
```diff
@@ -46,6 +46,12 @@ if (_active) then {
             closeDialog 0;
         };
     };
+    // Unlock controls for copilot if unit is pilot of aircraft
+    if (((vehicle _unit) isKindOf "Air") && {_unit == driver vehicle _unit}) then {
```
mharis001:
```suggestion
    if (vehicle _unit isKindOf "Air" && {_unit == driver vehicle _unit}) then {
```

addons/medical_status/functions/fnc_setUnconsciousState.sqf
```diff
@@ -49,7 +49,7 @@ if (_active) then {
     // Unlock controls for copilot if unit is pilot of aircraft
     if (((vehicle _unit) isKindOf "Air") && {_unit == driver vehicle _unit}) then {
         TRACE_1("pilot of air vehicle - unlocking controls",vehicle _unit);
-        // Do "Unlock controls" user action (co-pilot will then have to do the "Take Controls" actions)
+        // Do "Unlock controls" user action, co-pilot will then have to do the "Take Controls" actions
```
commy2: haha

Quailsnap:
ya'll physic or did somebody tell you?
https://youtu.be/jzXmyjPKmS0?t=26185

addons/medical_status/functions/fnc_setUnconsciousState.sqf
```diff
@@ -46,6 +46,12 @@ if (_active) then {
             closeDialog 0;
         };
     };
+    // Unlock controls for copilot if unit is pilot of aircraft
+    if (vehicle _unit isKindOf "Air" && {_unit == driver vehicle _unit}) then {
```
commy2: I think, since you do the Air check first, you can use as second check: _unit == driver _unit as well. (saves 1 command :P)

PabstMirror: image

commy2: ok :(

PR #7535, opened 2020-02-18
Don't mark help wanted as stale

dedmen:
Help-wanted tickets are easy to solve things, that the core team usually doesn't do because they are unimportant and to leave them to beginners.
Thus they often go "stale", but they shouldn't.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -14,6 +14,7 @@ exemptLabels:
   - status/cherrypick
   - status/merge on release
   - sticky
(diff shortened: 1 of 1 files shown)


commy2:
Interesting interpretation of the "help wanted" tag...

SilentSpike:
I would tend to use the tag for tickets I don't have the time or a strong motivation to pursue, which would tend to be more intermediate/hard things.

Still shouldn't become stale really.

PR #7536, opened 2020-02-20
Fastroping - Fix deployAI not unwinding ropes

mharis001:
- After introduction of different rope lengths, `deployRopes` does not unwind ropes when called from `deployAI` because a rope class is not provided
- This PR fixes that by using the old rope length of 34.5 when the config entry `GVAR(ropeLength)` is not a positive number

### DIFF
--- a/addons/fastroping/functions/fnc_deployRopes.sqf
+++ b/addons/fastroping/functions/fnc_deployRopes.sqf
@@ -34,7 +34,7 @@ if (_ropeLength <= 0) then {
 
 TRACE_3("",_ropeClass,_ropeLength,GVAR(requireRopeItems));
 
-if (GVAR(requireRopeItems)) then {
+if (GVAR(requireRopeItems) && {_ropeClass != ""}) then {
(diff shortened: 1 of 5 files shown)


addons/fastroping/functions/fnc_deployRopes.sqf
```diff
@@ -26,9 +27,15 @@ private _deployedRopes = _vehicle getVariable [QGVAR(deployedRopes), []];
...
 
 private _ropeLength = getNumber (configfile >> "CfgWeapons" >> _ropeClass >> QGVAR(ropeLength));
+
+if (_ropeLength <= 0) then {
+    _ropeLength = DEFAULT_ROPE_LENGTH;
+};
+
 TRACE_3("",_ropeClass,_ropeLength,GVAR(requireRopeItems));
+
 if (GVAR(requireRopeItems)) then {
```
PabstMirror:
```suggestion
if (GVAR(requireRopeItems) && {_ropeClass != ""}) then {
```

Issue #7534, opened 2020-02-18
RHS RKG-3M/BRK M79 grenades not compatible with advanced throwing.


severgun:
**Mods (complete and add to the following information):**
- **Arma 3:** `1.96` stable
- **CBA:** `3.14` stable
- **ACE3:** `3.13` stable

**Description:**
This scripted grenades explode on players position.

**Steps to reproduce:**
- Go to editor
- Spawn unit with rhssaf_mag_brk_m79 rhsgref_mag_rkg3em grenades
- Try to throw with adv. throwing

**Expected behavior:**
As regular throw. Fly, deploy parachute, explode on hit.

**Where did the issue occur?**
- Tested in Editor (Singleplayer). Occur everywhere.

Whigital:
Needs similar compat entries in config as in https://github.com/acemod/ACE3/pull/6879 ?

Whigital:
Pretty sure this has to do with the `rhs_fuseTime` config value and the WaitAndExecute delay here:
https://github.com/acemod/ACE3/blob/c73f64617fcc2b1211878423296e9558068eb8f6/addons/advanced_throwing/functions/fnc_throw.sqf#L83

Tested with `rhs_fuseTime[] = {0.20,0.25,0.3}` for `rhsgref_ammo_rkg3em` and `rhs_fuseTime[] = {0.45,0.45,0.45}` for `rhssaf_ammo_brk_m79`.

`rhsgref_ammo_rkg3em` almost consistently got deployed at your feet, while `rhssaf_ammo_brk_m79` followed the throwing arc.

severgun:
> Needs similar compat entries in config as in #6879 ?

There is no config entries in that PR. It actually remove some old dirty hacks.

> Tested with rhs_fuseTime[] = {0.20,0.25,0.3} for rhsgref_ammo_rkg3em and rhs_fuseTime[] = {0.45,0.45,0.45} for rhssaf_ammo_brk_m79.
Hmm. OK will try to increase this deploy range and test later today

Whigital:
True, the PR only hooks into whatever the "Throw" event has defined for the specific class.

In this instance, the `rhs_fuseTime` values should be incresed above the `0.3` delay used in `ace_advanced_throwing_fnc_throw`. Perhaps even have that `0.3` added to the default values.

Don't know what effect it has on the functionality of the nades though, if the `rhs_fuseTime` values are set too high.

[...]

PR #7537, opened 2020-02-20
Compats RHS GREF - Increase fuseTime values

Whigital:
- Updates `rhs_fuseTime` array for `rhsgref_ammo_rkg3em` grenades to account for delay used in `ace_advanced_throwing_fnc_throw`
- Partial fix for https://github.com/acemod/ACE3/issues/7534

### DIFF
--- a/optionals/compat_rhs_gref3/CfgAmmo.hpp
+++ b/optionals/compat_rhs_gref3/CfgAmmo.hpp
@@ -2,7 +2,9 @@ class CfgAmmo {
     class GrenadeHand;
     class rhsgref_ammo_rkg3em: GrenadeHand { // Scripted shaped charge
         ace_frag_force = 0;
+        rhs_fuseTime[] = {0.6, 0.65, 0.7}; // Up the fuseTime values to account for delay used in ace_advanced_throwing_fnc_throw
     };
+
     class BulletBase;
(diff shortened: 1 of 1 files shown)


commy2:
Where are these numbers from?

severgun:
> Where are these numbers from?
Check comments at issue ticket.
Native RHS config values are `rhs_fuseTime[] = {0.3,0.35,0.4};`
Add 0.3 sec delay to compensate
https://github.com/acemod/ACE3/blob/c73f64617fcc2b1211878423296e9558068eb8f6/addons/advanced_throwing/functions/fnc_throw.sqf#L83
Result: `rhs_fuseTime[] = {0.6,0.65,0.7};`

Whigital:
See the issue for reasoning/background. Could update the comment in the config with more details for future reference.

commy2:
>This PR can affect regular throw in bad way, but I think in most cases it will be OK.

Then it's a no go and needs something less hacky/more sophisticated.

severgun:
> Then it's a no go and needs something less hacky/more sophisticated.

Theoretically, what can happen is that grenade will hit target before metamorphosis and disappear.
First ammo have 
```
explosionTime = -1;
explosionEffects = "";
hit = 0.1;
indirectHit = 0;
```

PR #7541, opened 2020-02-22
Reload - Fix static weapon ammo check

### LINKED ISSUE #7540: For some non-vanilla statics, "Check ammo" doesn't work right
- **Arma 3:** `1.96.146114` (current stable)
- **CBA:** `3.14.0`
- **ACE3:** `3.13.0`
- [RHS:AFRF](https://steamcommunity.com/sharedfiles/filedetails/?id=843425103): `v0.5.0`

(plus the current version of TFAR, 1.0.325, but that shouldn't matter here)

**Description:**
On an XM307 (from the base game), the check ammo display looks like it does on rifles and pistols:

https://i.imgur.com/VEzriXV.png

i.e. it has Ammo type and a bar showing how much ammo is left. Doing the same on a KORD from RHS:AFRF yields only the ammo type, `1x` and no bar:

https://i.imgur.com/zzQp9p7.png

I am not sure if this is an overflow problem (the longer test with 1x in it pushing the bar to an invisible area) or if the RHS statics have weird magazine configurations.

**Steps to reproduce:**

Place a KORD in the editor, go into mission, use ACE interact to check the ammo of the KORD.

**Expected behavior:**

A bar showing the state of the KORD magazine.

- I can reproduce this in editor-launched, server side and single player missions.

TheCandianVendingMachine:
For some reason ARMA never returns any muzzle for crewed/non crewed static weapons. Work around this by always returning whatever weapon the static weapon has no matter what

fixes #7540

### DIFF
--- a/addons/reload/functions/fnc_displayAmmo.sqf
+++ b/addons/reload/functions/fnc_displayAmmo.sqf
@@ -42,6 +42,11 @@ if (_target isKindOf "StaticWeapon") then {
             };
         } forEach _magazines;
     };
+    
+    // For static weapons the muzzle seemingly never returns anything for static weapons with/without people inside
+    if (_muzzle == "") then {
+        _muzzle = _weapon;
+    };
 };
 
 if (_magazine == "") exitWith {};


commy2:
>For some reason ARMA never returns any muzzle for crewed/non crewed static weapons.

Probably the weapon config missing:
```cpp
muzzles[] = {"this"};
```
or having
```cpp
muzzles[] = {};
```
or something.

TheCandianVendingMachine:
> > For some reason ARMA never returns any muzzle for crewed/non crewed static weapons.
> 
> Probably the weapon config missing:
> 
> ```c++
> muzzles[] = {"this"};
> ```
> 
> or having
> 
> ```c++
> muzzles[] = {};
> ```
> 
> or something.

Yeah but this is the case for every single static weapon, vanilla and modded. If you go `currentMuzzle gunner vehicle player` it will return the correct muzzle if a person is on the weapon, but that isnt a good check if no ones in the weapon.

I'm personally going with: it worked before and an ARMA update changed behaviour. According to the biki `By default the inherited muzzle for a weapon is: muzzles[] = {"this"};` as well, so it seems like it was correct behaviour and now it isn't

PR #7544, opened 2020-02-22
Common - Add isNotUnconscious canInteractWith condition

mharis001:
- title, this was removed in rewrite


jonpas:
Should fix ability to use interaction menu while being unconscious.

PR #7547, opened 2020-02-23
Medical Statemachine - Add setting to disallow bleedout in cardiac arrest

BaerMitUmlaut:
- Title

CA timer is useless if you can die anyways.


addons/medical_statemachine/Statemachine.hpp
```diff
@@ -101,10 +101,6 @@ class ACE_Medical_StateMachine {
             condition = QFUNC(conditionExecutionDeath);
             events[] = {QEGVAR(medical,FatalInjury)};
         };
-        class Bleedout {
-            targetState = "Dead";
-            events[] = {QEGVAR(medical,Bleedout)};
```
SilentSpike: Believe this event may as well be removed entirely if we're making this change as its whole purpose was to kill a unit in cardiac arrest if they lost loads of blood (to the point of being non-recoverable where there's no point letting the timer run down).

jonpas:
> I assume this was a change approved at the meeting

Negative, @BaerMitUmlaut was not at the meeting. :P

addons/medical_statemachine/Statemachine.hpp
BaerMitUmlaut: We can keep it, but behind a setting, I'd be fine with that. However in this current state, there is no way to get a somewhat similar behaviour to the old revive state, where you can always get units back up if they get treated within a certain timespan.

SilentSpike: To be honest I'm not even sure if it fulfils its original purpose, do we have much feedback on this front?

PabstMirror:
Ref https://github.com/acemod/ACE3/pull/7060

Maybe we should just add a setting to control if blood loss is possible while in CA

addons/medical_statemachine/Statemachine.hpp
BaerMitUmlaut: People were asking about that in Slack, yes.

jonpas:
> **PabstMirror: kinda links back to this [PR], if they can't bleedout in CA, then AI will be unkillable for a minute or so

Whigital:
> 
> Maybe we should just add a setting to control if blood loss is possible while in CA

A little off topic, but would it be too complicated to lower the bleeding coef. when in CA, since no pressure is being applied to the vascular system ?

BaerMitUmlaut:
Blood loss already depends on heart rate but has a minimum.

bux:
Agree with pabst, a setting would make sense here.

BaerMitUmlaut:
Setting it is 👍

addons/medical_statemachine/Statemachine.hpp
Elgin675: My group and I are opposed to this change: being able to bleed to death while in cardiac arrest was a really positive point for the gameplay.

With this change, it's less interesting.

Might as well reduce the bleeding coefficient, if that's a problem?

BaerMitUmlaut:
Made it a setting, also fixed the translation of "cardiac arrest time" in German (previously said "time until cardiac arrest" which makes no sense) since I reused its name in the description of the new setting.

Not tested.

addons/medical_statemachine/stringtable.xml
```diff
@@ -124,5 +125,13 @@
             <Czech>V srdeční zástavě</Czech>
             <Polish>Zatrzymanie Akcji Serca</Polish>
         </Key>
+        <Key ID="STR_ACE_Medical_Statemachine_CardiacArrestBleedout_DisplayName">
+            <English>Bleedout during cardiac arrest</English>
```
mharis001, line 129: Title case to match other settings.

addons/medical_statemachine/stringtable.xml
```diff
@@ -124,5 +125,13 @@
...
+            <German>Ausbluten im Herzstillstand</German>
+        </Key>
+        <Key ID="STR_ACE_Medical_Statemachine_CardiacArrestBleedout_Description">
+            <English>Controls whether you can die in cardiac arrest by blood loss before the cardiac arrest time runs out.</English>
```
mharis001, line 133:
```suggestion
            <English>Controls whether a person can die in cardiac arrest by blood loss before the cardiac arrest time runs out.</English>
```

slight72:
Hello everybody. My first post here and sorry for my poor English! 
I completely agree with dying from exsanguination, it’s realistic, but downside of current setting, I think, is unit respawn. 
It is now impossible to prevent that, unless setting bleeding coefficient to zero, which is, of course, nonsense. 
I think it will be useful to have an option to choice between dying or staying in unconscious/revive state when “fatal amount of blood etc.” happens. 
In this way, if I choose “unconscious  state”, it will be possible, for example, to bring back bodies to base or to call medevac to bring them to hospital, or things like that, instead of respawning . 
Not better or worst, but simply a different way to play !.
I look forward to hearing your comments! Thanks for your attention !

slight72 (approved):

BaerMitUmlaut:
@slight72 It is already possible to stay in a revivable state indefinitely, use the fatal injuries and cardiac arrest settings.
If you have questions regarding this, it's probably best if you'd join our Slack or create an issue. This is a code change and not really the right place for asking questions, as nobody will see them.

slight72 (approved):

slight72:
Thanks for your answer ! I already know it's possible but I however can't avoid to die when I lost more then 50% of blood. According to title I was suggesting to add an option. As I said, my first time here, sorry :) What do I have to do to suggest this modify ? Thanks again !

TheParadoxNinja (approved):

PabstMirror:
Would we also want to stop bleeding while in CA as well in `ace_medical_status_fnc_getBloodLoss`
something like
`_cardiacOutput max ([0, 0.05] select EGVAR(x,cardiacArrestBleedoutEnabled))`

BaerMitUmlaut:
No, this would be the closest to old medical.

PabstMirror:
This is what I am worried about:
Bleeding
In CA, still bleeding
Stabilized, 
CPR successful
then die from bleed out

but I don't think it really matters, and this setting is more for people using CA more as a revive meta state?

BaerMitUmlaut:
If CPR is successful and you continue to lose blood, wouldn't you just go into the CA state again?

PabstMirror (approved):

mharis001 (approved):

PR #7548, opened 2020-02-23
NightVision - Don't apply AimDownSightsBlur on optics with integrated night vision

### LINKED ISSUE #6666: NVS scope has blur

commy2:
- fix #6666 super satan
- replace close #6942

To explain: If the Magnification is > 3, we can safely assume that we're not looking down iron/holo/cqb sights, but are instead dealing with an optic with actual (!= from RMB zoom) magnification and thus integrated night vision (potentially with its own blur effects).

We need to refresh the effect every time magnification is changed from <3 to >3 and vice versa, as there is no OpticsMode event.

### DIFF
--- a/addons/nightvision/XEH_postInit.sqf
+++ b/addons/nightvision/XEH_postInit.sqf
@@ -21,6 +21,7 @@ GVAR(ppeffectRadialBlur) = -1;
 GVAR(ppeffectColorCorrect) = -1;
 GVAR(ppeffectBlur) = -1;
 
+GVAR(isUsingMagnification) = false;
 
 ["ace_settingsInitialized", {
     TRACE_4("settingsInitialized",GVAR(disableNVGsWithSights),GVAR(fogScaling),GVAR(noiseScaling),GVAR(effectScaling));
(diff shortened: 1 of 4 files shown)


commy2:
This was tested with `BWA3_optic_ZO4x30_RSAS_NSV` (optic with hmd nvg compatible CQB holo optics mode, and internal nvg scope optics mode) and it works as expected: blur only applied on CQB.

PR #7556, opened 2020-02-25
Ballistics - Move dispersion to optional

Laid3acK:
according with https://github.com/acemod/ACE3/pull/7532#issuecomment-589994082

### DIFF
--- a/optionals/compat_ace_dispersion/$PBOPREFIX$
+++ b/optionals/compat_ace_dispersion/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\ace\addons\compat_ace_dispersion
--- a/optionals/compat_ace_dispersion/script_component.hpp
+++ b/optionals/compat_ace_dispersion/script_component.hpp
@@ -1,0 +1,6 @@
+#define COMPONENT compat_ace_dispersion
+#define COMPONENT_BEAUTIFIED ACE3 Dispersion Compatibility
+
+#include "\z\ace\addons\main\script_mod.hpp"
+
+#include "\z\ace\addons\main\script_macros.hpp"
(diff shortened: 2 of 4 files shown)


optionals/compat_ace_dispersion/script_component.hpp
```diff
@@ -0,0 +1,6 @@
+#define COMPONENT compat_ace_dispersion
```
jonpas: `realisticdispersion` maybe? Mimics `realisticnames`.

`ace` is already there as a prefix. It's not really a compatibility.

optionals/compat_ace_dispersion/script_component.hpp
```diff
@@ -0,0 +1,6 @@
...
+#define COMPONENT_BEAUTIFIED ACE3 Dispersion Compatibility
```
jonpas: It's not really compatibility, more that it provides realistic values.

optionals/compat_ace_dispersion/$PBOPREFIX$
```diff
@@ -0,0 +1 @@
+z\ace\addons\compat_ace_dispersion
```
jonpas:
```suggestion
z\ace\addons\compat_ace_dispersion

```

optionals/compat_ace_dispersion/script_component.hpp
Laid3acK: Nothing realistic with these dispersions, except for casuals.
Rename that as you want.

optionals/compat_ace_dispersion/script_component.hpp
Laid3acK: Same.

optionals/compat_ace_dispersion/script_component.hpp
jonpas: Wasn't your plan to improve them? I'd like to avoid renaming.

Laid3acK: Do it please Jonpas, no problem.
To bad connexion at this moment.

PR #7557, opened 2020-02-25
Ballistics - Remove dispersion changes

Laid3acK:
According with https://github.com/acemod/ACE3/pull/7532#issuecomment-589994082


Laid3acK:
Next `Rifle_Base_F` clean up.

addons/ballistics/CfgWeapons.hpp
```diff
@@ -8,11 +8,98 @@ class CfgWeapons {
...
+    // Rifle_Long_Base Sniper Marksman
     // GM6 Lynx
-    class GM6_base_F: Rifle_Long_Base_F {};
+    class GM6_base_F: Rifle_Long_Base_F {
+        ACE_barrelLength = 730;
+        ACE_barrelTwist = 381.0;
+        initSpeed = -1.0;
+        magazineWell[] = {
+            "CBA_50BMG_M107"
+        }; // empty in vanilla
```
dedmen: doesn't mean that other mods aren't adding to it. += is safer

Laid3acK:
Except if i mistaken somewhere, should fix this issue https://github.com/acemod/ACE3/issues/7559

commy2:
If it does, it does so poorly. Still using magazines[] instead of magwells. So I'd rather keep that issue open until a proper fix.

PR #7567, opened 2020-03-02
Ballistics - Fix DMR_03 (SIG556) initSpeed

Laid3acK:
- Fix the `DMR_03_base_F` `initSpeed`.
- Fix the missing `arifle_katiba_Base_F` 30 rnd IR-Dim magazine (KH2002 Sama variants).
- Fix the `ACE_30Rnd_556x45_Stanag_Tracer_Dim` picture (30 rnd magazine icon instead 20 rnd).

### DIFF
--- a/addons/ballistics/CfgWeapons.hpp
+++ b/addons/ballistics/CfgWeapons.hpp
@@ -57,7 +57,7 @@ class CfgWeapons {
     };
 
     class srifle_DMR_03_F: DMR_03_base_F {
-        initSpeed = -0.991536;
+        initSpeed = -0.991536; // Default vanilla value 940
     };
 
     // ASP-1 Kir
--- a/addons/ballistics/CfgWeapons.hpp
+++ b/addons/ballistics/CfgWeapons.hpp
@@ -7,7 +7,8 @@ class CfgWeapons {
 
     // Rifle_Long_Base_F Sniper Marksman
     // GM6 Lynx
-    class GM6_base_F: Rifle_Long_Base_F {
+    class GM6_base_F: Rifle_Long_Base_F {};
+    class srifle_GM6_F: GM6_base_F {
         ACE_barrelLength = 730;
         ACE_barrelTwist = 381.0;
         initSpeed = -1.0;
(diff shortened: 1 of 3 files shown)


addons/ballistics/CfgWeapons.hpp
```diff
@@ -51,12 +51,15 @@ class CfgWeapons {
...
         ACE_barrelLength = 508.0;
         ACE_barrelTwist = 254.0;
-        initSpeed = -0.991536;
         magazineWell[] += {
             "CBA_762x51_M14"
         }; // vanilla magazineWell[] = {"M14_762x51"};
     };
 
+    class srifle_DMR_03_F: DMR_03_base_F {
+        initSpeed = -0.991536;
```
commy2: Explain.

commy2:
Why all the empty lines?

addons/ballistics/CfgWeapons.hpp
Laid3acK: When i tested the RC, i saw that the `class srifle_DMR_03_F` has his own `initSpeed `, 940 if i remember correctly.

Laid3acK:
I believed it was a part of the Coding Guidelines. I'm wrong ?

addons/ballistics/CfgWeapons.hpp
commy2: Yeah, saw that. I think we should keep the value in both classes and put a comment after this line.

addons/ballistics/CfgWeapons.hpp
```diff
@@ -51,12 +51,15 @@ class CfgWeapons {
...
         ACE_barrelLength = 508.0;
         ACE_barrelTwist = 254.0;
-        initSpeed = -0.991536;
         magazineWell[] += {
             "CBA_762x51_M14"
         }; // vanilla magazineWell[] = {"M14_762x51"};
     };
 
+    class srifle_DMR_03_F: DMR_03_base_F {
+        initSpeed = -0.991536; // Default vanilla value 940
```
commy2: No, an explanation why this is here. And also keep the original change.

Laid3acK: I don't understand what do you mean with an explanation. There is no explanation with the old config.
>And also keep the original change
Same, i don't understand my mistake.

Laid3acK:
Reverted to the old format (better compatibility with third mods).

PR #7571, opened 2020-03-03
Realistic Names - Add 2D optics

TyroneMF:
- add the ace 2d optics to realistic names.


TyroneMF:
https://i.gyazo.com/7fbf67a38b9cf73b012a9aa377bacb0c.png

jonpas:
Just require `optics`. `optics` only contains added 2D and PIP optics, something one doesn't even have to use. It's not like one is forced to use MILs with it (`scopes` does that).

And for the rare case of someone wanting to use `realisticnames` but not `optics`, we can't satisfy every last 0.00001%.

PR #7575, opened 2020-03-05
Ballistics : add ACE magazines in vanilla magwells

Laid3acK:
- update some ACE magazines in vanilla magazineWell as https://github.com/acemod/ACE3/pull/7562 and remove useless CBA magazineWell in the CfgWeapons.

### DIFF
--- a/addons/ballistics/CfgWeapons.hpp
+++ b/addons/ballistics/CfgWeapons.hpp
@@ -11,10 +11,8 @@ class CfgWeapons {
         ACE_barrelLength = 730;
         ACE_barrelTwist = 381.0;
         initSpeed = -1.0;
-        magazines[] += { // No Vanilla magwell
-            "ACE_5Rnd_127x99_Mag",
-            "ACE_5Rnd_127x99_API_Mag",
-            "ACE_5Rnd_127x99_AMAX_Mag"
+        magazineWell[] = { // Vanilla magwell
+            "GM6_127x108"
         };
     };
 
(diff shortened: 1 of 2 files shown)


Laid3acK:
Really proper like that in my opinion. Thanks commy2. :)

addons/ballistics/CfgWeapons.hpp
```diff
@@ -11,19 +11,18 @@ class CfgWeapons {
         ACE_barrelLength = 730;
         ACE_barrelTwist = 381.0;
         initSpeed = -1.0;
-        magazineWell[] += {
-            "CBA_50BMG_M107"
```
dedmen: why remove a valid magazine well? This one should be added on CBA side I'd say

Laid3acK: The both work in-game. As you want.
Maybe the name of this CBA magwell should be changed,
it's not really appropriate to the vanilla game and this rifle.

Laid3acK:
I just saw that the Lynx has his 12.7x108 magwell in the AiO config,
i just add it in the cfgweapons (probably a BI forgetting).

commy2:
>not a forgetting, the Lynx can't have a magwell with the vanilla game.

Why would that be?

Laid3acK:
A weapon configured with two magazines without the same `initSpeed` included in the `magazines[]` and also the `magazineWell[]` could work correctly with the vanilla game ?
I forgot that the Lynx has an `initSpeed=0`.  So probably yes.

Mike-MF:

PR #7578, opened 2020-03-07
Map - Detect switchable flashlight attachments for map

### LINKED ISSUE #7130: Map light doesn't detect laser/light attachments
**Arma 3 Version:** `1.94` stable
**CBA Version:** `3.12.0.190708` stable
**ACE3 Version:** `3.12.6` stable

**Mods:**
```
- CBA_A3
- ace
- TONS OF OTTHER MODS YA FUQ
```

**Description:**
- You cannot use a Laser/Light weapon attachment as Maplight if it's not currently switched to Light mod.

**Steps to reproduce:**
- Take laser/light attachment onto gun
- Can't use it as maplight

**Additional information:**
- Cuz it checks only the current attachment, which is currently a laser only. Thought as you can switch directly over, it would make sense that you can just switch it. Instead of first having to switch to your pistol, then switch to flashlight, then switch back to main weapon, then go to map and then enable the light.

**RPT log file:**
N/A

veteran29:
- title

Fixes #7130 

Needs https://github.com/CBATeam/CBA_A3/pull/1294


addons/map/functions/fnc_isFlashlight.sqf
```diff
@@ -0,0 +1,46 @@
...
+        if (-1 != [
+            _weaponConfig >> "ItemInfo" >> "FlashLight",
+            _weaponConfig >> "FlashLight"
+        ] findIf {
+            isText (_x >> "ACE_Flashlight_Colour")
+            || {!(getArray (_x >> "ambient") in [[], [0,0,0]])}
+        }) exitWith {
+            true
+        };
+        false
```
commy2: :thonk:

PabstMirror: `findIf` here might be cleaner

PabstMirror:
- [ ] - Update required CBA version (when it releases)

jonpas:
> Backlog: Tasks that **have a specific end point**, but no due date.
> Ongoing: Tasks that do not have a specific due date and **are ongoing as circumstances change**.

addons/map/functions/fnc_isFlashlight.sqf
veteran29: 🙃

addons/map/functions/fnc_isFlashlight.sqf
```diff
@@ -0,0 +1,44 @@
...
+        private _weaponConfig = _cfgWeapons >> _x;
+
+        -1 != [
+            _weaponConfig >> "ItemInfo" >> "FlashLight",
+            _weaponConfig >> "FlashLight"
+        ] findIf {
+            isText (_x >> "ACE_Flashlight_Colour")
+            || {!(getArray (_x >> "ambient") in [[], [0,0,0]])}
+        } // return
+    });
```
mharis001:
```suggestion
    _isFlashlight = _items findIf {
        private _weaponConfig = _cfgWeapons >> _x;

        [
            _weaponConfig >> "ItemInfo" >> "FlashLight",
            _weaponConfig >> "FlashLight"
        ] findIf {
            isText (_x >> "ACE_Flashlight_Colour")
            || {!(getArray (_x >> "ambient") in [[], [0,0,0]])}
        } != -1 // return
    } != -1;
```

addons/map/functions/fnc_isFlashlight.sqf
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Public: No
+ */
+
+params [["_class", "", [""]]];
+
+private _isFlashlight = GVAR(flashlights) getVariable _class;
+
+if (isNil "_isFlashlight") then {
+
```
mharis001:
```suggestion
```

addons/map/functions/fnc_isFlashlight.sqf
```diff
@@ -0,0 +1,44 @@
+#include "script_component.hpp"
+/*
+ * Author: veteran29
+ * Check if given class is a flashlight
```
mharis001:
```suggestion
 * Checks if the given item is a flashlight.
```

addons/map/functions/fnc_isFlashlight.sqf
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Arguments:
+ * 0: Class to check <STRING>
```
mharis001:
```suggestion
 * 0: Item Classname <STRING>
```

PR #7594, opened 2020-03-11
Arsenal - Fix main menu mission

### LINKED ISSUE #7584: ACE3 Arsenal acts as SP mission
- **Arma 3:** `1.96` 
- **CBA:** `3.14.0`
- **ACE3:** `3.13.1`
_Make sure to reproduce the issue with only CBA and ACE3 on a newly created mission!_

**Description:**
The ACE3 Arsenal acts like a singleplayer mission in terms of functionality. Whilst the UI seems to be fine, the button "Try" has been replaced by "Close", and once trying out a loadout, there is no way to return to the Loadout Arsenal screen. The only option is to "Save and Quit". Moreover when in the loadout screen, the background is not empty; the virtual vehicles and obstacles are visible. When trying out the loadout, the AI targets are not present.

**Steps to reproduce:**
- Go to ACE3 Arsenal .
-> "Try" button is missing.
- Click "Close" to proceed testing the loadout.
- Click Escape.
- See that "Arsenal" button has been replaced with "Save and Quit".

**Expected behavior:**
The buttons "Try" and "Arsenal" as described above should be present to achieve Virtual Arsenal functionality.

ACE3 Virtual Arsenal

**Log Files:**
Attached. 

EDIT: Turns out the archive w/ RPT log wasn't attached after all. 
Pastebin link: https://pastebin.com/gE98MG1D

PabstMirror:
Fix #7584

- Macros broke loading the mission's XEH_preInit
- With double killed XEH, 2nd time it's now objNull and threw script error because of a
`missionNamespace setvariable [_varName`

edit: I still think setting `objNull` is the best solution to the double killed xeh, just need to watch out for odd uses like this one

### DIFF
--- a/addons/arsenal/missions/Arsenal.VR/CfgEventHandlers.hpp
+++ b/addons/arsenal/missions/Arsenal.VR/CfgEventHandlers.hpp
@@ -1,11 +1,11 @@
 class Extended_PreInit_EventHandlers {
     class ADDON {
-        init = QUOTE(call COMPILE_FILE(XEH_preInit));
+        init = QUOTE(call compile preprocessFileLineNumbers 'XEH_preInit.sqf');
     };
 };
 
 class Extended_PostInit_EventHandlers {
     class ADDON {
-        init = QUOTE(call COMPILE_FILE(XEH_postInit));
+        init = QUOTE(call compile preprocessFileLineNumbers 'XEH_postInit.sqf');
     };
 };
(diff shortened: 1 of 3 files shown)


commy2:
Why is COMPILE_FILE() broken?

jonpas:
It's a mission and doesn't include it?

PabstMirror:
file path is `addons/arsenal/missions/Arsenal.VR/CfgEventHandlers.hpp`
it's a mission inside of a addon
we just want the mission files, not the base addon

commy2:
Ah, I see. One is a folder from addon space, while the PR is from mission space (no leading slash + no file that matches the path in addon space).

Maybe could use a comment.

PR #7596, opened 2020-03-12
Arsenal - add custom right panel misc items buttons

shukari:
- it is possible to sort special misc items in extra right panel buttons

#### Infos
- with `[["ACE_bloodIV_500", "ACE_Banana"], "Stuff"] call ace_arsenal_fnc_addRightPanelButton` you can do an extra button there
- original idea is from @Schwaggot 
- ~icon is from https://material.io/~
    - icon from myself

#### Example for medical
With this code, you have a setting for ace_medical to add a custom button with all medical stuff.
If this is good, i can add this also to this PR.
```
[QGVAR(addMedicalArsenalButton), "CHECKBOX", "addMedicalArsenalButton", localize LSTRING(settingCategory), true, true, {
    if (_this && isNil QGVAR(medicalArsenalButton)) then {
        GVAR(medicalArsenalButton) = [["ACE_fieldDressing", "ACE_packingBandage", "ACE_elasticBandage", "ACE_tourniquet", "ACE_splint",
            "ACE_morphine", "ACE_adenosine", "ACE_atropine", "ACE_epinephrine", "ACE_plasmaIV", "ACE_plasmaIV_500", "ACE_plasmaIV_250",
            "ACE_bloodIV", "ACE_bloodIV_500", "ACE_bloodIV_250", "ACE_salineIV", "ACE_salineIV_500", "ACE_salineIV_250", "ACE_quikclot",
            "ACE_personalAidKit", "ACE_surgicalKit", "ACE_bodyBag"], "MedicalStuff", "\z\ace\addons\arsenal\data\iconCustom.paa"] call ace_arsenal_fnc_addRightPanelButton;
    };
}, true] call CBA_fnc_addSetting;
```


BaerMitUmlaut:
That's a great idea, however I would like to see the following changes:
- Icon should be settable
- Tab should appear above the misc tab, because that's the "everything else" button
- Dynamically add them (you could have the basic definition in config and then `ctrlCreate` them)

I know this would increase complexity, but you would just need to calculate a couple positions.

shukari:
> Icon should be settable

it is

> Tab should appear above the misc tab, because that's the "everything else" button

because there are 10 static atm i placed it under that misc tab

> Dynamically add them (you could have the basic definition in config and then ctrlCreate them)

because of the arsenal system and the defines i build it extra static that everything is defined and has a IDC etc. But yeah i can make it dynamic if this is wished.

addons/arsenal/defines.hpp
```diff
@@ -413,3 +451,15 @@ _buttonCurrentMag2Ctrl ctrlCommit FADE_DELAY;\
...
     _contentPanelCtrl lnbSetPicture [[_newRow, 8], getText (configFile >> "cfgWeapons" >> (_loadout select 6) >> "picture")];\
     _contentPanelCtrl lnbSetPicture [[_newRow, 9], getText (configFile >> "cfgGlasses" >> (_loadout select 7) >> "picture")];
+
+#define ADD_CUSTOM_BUTTON(NUMBER)\
+    class iconBackgroundCustom##NUMBER: iconBackgroundOptic {\
+        idc = __EVAL((IDC_iconBackgroundCustom1 - 2) + (NUMBER * 2));\
+        y = QUOTE(safezoneY + (88 + (10 * NUMBER)) * GRID_H);\
+    };\
+    class buttonCustom##NUMBER: buttonOptic {\
+        idc = __EVAL((IDC_buttonCustom1 - 2) + (NUMBER * 2));\
```
commy2: This can't be binarized by all tools. Just use a string.

addons/arsenal/functions/fnc_addRightPanelButton.sqf
```diff
@@ -0,0 +1,51 @@
...
+};
+
+private _position = 0 max ((GVAR(customRightPanelButtons) findIf {isNil "_x"}) min (count GVAR(customRightPanelButtons)));
+
+if (_override != -1 && {_override >= 0} && {_override <= 9}) then {
+    _position = _override;
+};
+
+private _return = if (_position >= 0 && _position <= 9) then {
+    private _configCfgWeapons = configFile >> "CfgWeapons";
```
commy2: "config config weapons"?

addons/arsenal/functions/fnc_addRightPanelButton.sqf
```diff
@@ -0,0 +1,51 @@
...
+    _items = _items select {
+        private _configItemInfo = _configCfgWeapons >> _x >> "ItemInfo";
+        
+        isClass (_configItemInfo) && ((getNumber (_configItemInfo >> "type")) in [TYPE_MUZZLE, TYPE_OPTICS, TYPE_FLASHLIGHT, TYPE_BIPOD] &&
+            {(_x isKindOf ["CBA_MiscItem", (_configCfgWeapons)])}) || {(getNumber (_configItemInfo >> "type")) in [TYPE_FIRST_AID_KIT, TYPE_MEDIKIT, TYPE_TOOLKIT]} ||
+            {(getText (_configCfgWeapons >> _x >> "simulation")) == "ItemMineDetector"}
```
commy2: Terrible formatting. After fixing it and remove a superfluous check, this looks like this:
```sqf
_x isKindOf ["CBA_MiscItem", _configCfgWeapons] && {getNumber (_configItemInfo >> "type") in [TYPE_MUZZLE, TYPE_OPTICS, TYPE_FLASHLIGHT, TYPE_BIPOD]} ||
{getNumber (_configItemInfo >> "type") in [TYPE_FIRST_AID_KIT, TYPE_MEDIKIT, TYPE_TOOLKIT]} ||
{getText (_configCfgWeapons >> _x >> "simulation") == "ItemMineDetector"}
```
And I don't think it is right.

addons/arsenal/functions/fnc_addRightPanelButton.sqf
shukari: only a copy of https://github.com/acemod/ACE3/blob/febefe627415f5964089f4333a4ece40e3404ce7/addons/arsenal/functions/fnc_addVirtualItems.sqf#L48

addons/arsenal/functions/fnc_addRightPanelButton.sqf
shukari: only a copy of
https://github.com/acemod/ACE3/blob/febefe627415f5964089f4333a4ece40e3404ce7/addons/arsenal/functions/fnc_addVirtualItems.sqf#L168-L172

addons/arsenal/functions/fnc_addRightPanelButton.sqf
```diff
@@ -33,13 +33,13 @@ if (_override != -1 && {_override >= 0} && {_override <= 9}) then {
...
-        private _configItemInfo = _configCfgWeapons >> _x >> "ItemInfo";
-        
-        isClass (_configItemInfo) && ((getNumber (_configItemInfo >> "type")) in [TYPE_MUZZLE, TYPE_OPTICS, TYPE_FLASHLIGHT, TYPE_BIPOD] &&
-            {(_x isKindOf ["CBA_MiscItem", (_configCfgWeapons)])}) || {(getNumber (_configItemInfo >> "type")) in [TYPE_FIRST_AID_KIT, TYPE_MEDIKIT, TYPE_TOOLKIT]} ||
-            {(getText (_configCfgWeapons >> _x >> "simulation")) == "ItemMineDetector"}
+        private _configItemInfo = _cfgWeapons >> _x >> "ItemInfo";
+           
+        _x isKindOf ["CBA_MiscItem", _cfgWeapons] && {getNumber (_configItemInfo >> "type") in [TYPE_MUZZLE, TYPE_OPTICS, TYPE_FLASHLIGHT, TYPE_BIPOD]} ||
+            {getNumber (_configItemInfo >> "type") in [TYPE_FIRST_AID_KIT, TYPE_MEDIKIT, TYPE_TOOLKIT]} ||
+            {getText (_cfgWeapons >> _x >> "simulation") == "ItemMineDetector"}
```
commy2: Remove one extra indent in both lines.

shukari: done

dedmen:
Is there a limit? What if users push too many buttons (I'd call that unlikely to happen so prolly no matter)

addons/arsenal/defines.hpp
```diff
@@ -304,7 +325,24 @@ _buttonCurrentMag2Ctrl ctrlCommit FADE_DELAY;\
...
+    } else {\
+        _x ctrlSetFade 1;\
+        _x ctrlShow false;\
+        _x ctrlCommit 0;\
+        _x = _display displayCtrl (_idc - 1);\
+        _x ctrlSetFade 1;\
+        _x ctrlShow false;\
+        _x ctrlCommit 0;\
+    };\
+} foreach [RIGHT_PANEL_CUSTOM_BUTTONS];
```
dedmen:
```suggestion
} forEach [RIGHT_PANEL_CUSTOM_BUTTONS];
```

addons/arsenal/functions/fnc_addRightPanelButton.sqf
```diff
@@ -0,0 +1,51 @@
...
+    GVAR(customRightPanelButtons) = [];
+};
+
+private _position = 0 max ((GVAR(customRightPanelButtons) findIf {isNil "_x"}) min (count GVAR(customRightPanelButtons)));
+
+if (_override != -1 && {_override >= 0} && {_override <= 9}) then {
+    _position = _override;
+};
+
+private _return = if (_position >= 0 && _position <= 9) then {
```
dedmen: uselesss variable? might aswell just return the if/then statement. Or just exitWith.

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -272,10 +272,41 @@ switch (_ctrlIDC) do {
...
+    case IDC_buttonCustom7;
+    case IDC_buttonCustom8;
+    case IDC_buttonCustom9;
+    case IDC_buttonCustom10 : {
+        private _data = GVAR(customRightPanelButtons) param [[RIGHT_PANEL_CUSTOM_BUTTONS] findIf {_x == _ctrlIDC}];
+        
+        if (!isNil "_data") then {
+            private _items = _data select 0;
+            {
+                ["CfgWeapons", _x, true]  call _fnc_fill_right_Container;
```
dedmen:
```suggestion
                ["CfgWeapons", _x, true] call _fnc_fill_right_Container;
```

shukari:
> Is there a limit? What if users push too many buttons (I'd call that unlikely to happen so prolly no matter)

Yeah only 10 Buttons not more
https://github.com/acemod/ACE3/blob/603e6b363e2a6310cc7d53f2e1657f7d2e51ceba/addons/arsenal/functions/fnc_addRightPanelButton.sqf#L35

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -272,10 +272,41 @@ switch (_ctrlIDC) do {
...
+    case IDC_buttonCustom2;
+    case IDC_buttonCustom3;
+    case IDC_buttonCustom4;
+    case IDC_buttonCustom5;
+    case IDC_buttonCustom6;
+    case IDC_buttonCustom7;
+    case IDC_buttonCustom8;
+    case IDC_buttonCustom9;
+    case IDC_buttonCustom10 : {
+        private _data = GVAR(customRightPanelButtons) param [[RIGHT_PANEL_CUSTOM_BUTTONS] findIf {_x == _ctrlIDC}];
```
dedmen: what is this findIf?
That is just a `find`

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -272,10 +272,41 @@ switch (_ctrlIDC) do {
...
+        if (!isNil QGVAR(customRightPanelButtons)) then {
+            private _blockItems = [];
+            {
+                if (!isNil "_x") then {
+                    _blockItems append (_x select 0);
+                };
+            } forEach GVAR(customRightPanelButtons);
+            
+            {
+                ["CfgWeapons", _x, false]  call _fnc_fill_right_Container;
```
dedmen:
```suggestion
                ["CfgWeapons", _x, false] call _fnc_fill_right_Container;
```

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -272,10 +272,41 @@ switch (_ctrlIDC) do {
...
+            } foreach ((GVAR(virtualItems) select 17) select {!((toLower _x) in _blockItems)});
```
dedmen:
```suggestion
            } forEach ((GVAR(virtualItems) select 17) select {!((toLower _x) in _blockItems)});
```

addons/arsenal/functions/fnc_fillRightPanel.sqf
shukari: is findIf not better because not all gets checked?

dedmen:
> Yeah only 10 Buttons not more

should read before commenting...

addons/arsenal/functions/fnc_fillRightPanel.sqf
commy2: find is not even a loop, so it is infinitely better than findIf, which is a loop.

addons/arsenal/functions/fnc_addRightPanelButton.sqf
shukari: this was on purpose for better readability... should i change it?

shukari:
@BaerMitUmlaut what should i do now https://github.com/acemod/ACE3/pull/7596#issuecomment-598197023
is this for this purpose ok? or should i make everything dynamic (i dont now how atm)?

shukari:
@BaerMitUmlaut what should i do now https://github.com/acemod/ACE3/pull/7596#issuecomment-598197023
is this for this purpose ok? or should i make everything dynamic (i dont now how atm)?

shukari:
> @BaerMitUmlaut what should i do now [#7596 (comment)](https://github.com/acemod/ACE3/pull/7596#issuecomment-598197023)
> is this for this purpose ok? or should i make everything dynamic (i dont now how atm)?

I really need input. If i should make it dynamic then i need help. Im not used to ctrlCreate and GUI stuff so the misc tab must be there or i need help. If i need to change somthing then tell me.

BaerMitUmlaut:
>If i should make it dynamic then i need help.

You need to create a class in the config root for your control. You can set all common features there like height, width and background color. Ideally, its name would be prefixed - simply use GVAR. Then you can just use `ctrlCreate [QGVAR(miscButton), -1]`, `ctrlSetPosition` and `ctrlCommit 0` to create and position your control. Instead of using an IDC, you should just save the controls in a variable and access them that way. In that huge switch case, instead of checking the IDC, you can just use default and check if the control is within that variable.

If you need further help it would be best to talk in our Slack 👍

shukari:
@BaerMitUmlaut 
If changed it now. The problem i need the fix idcs for the hiding stuff (there is a hide button, there is hide if another tab etc.). I tried it with a ctrlGroup, but the fillRightPanel needs a button with a idc-1 background an as parent the right pane...
I dont now how i should do it better 😞

shukari:
Can someone review please? @commy2 @dedmen @BaerMitUmlaut @alganthe 
There is no maintainer or readme in the arsenal addon that i can annoy.

BaerMitUmlaut:
I took a closer look at it in a call with shukari and would say this is fine for now. Working completely without IDCs would require quite a lot of refactoring in other parts of the arsenal and I believe it's reasonable to not do that for now.

dedmen:
> icon is from https://material.io/

I cannot see the Apache License in your changes? Or are you trying to just relicense Apache license content as GPL? I don't think that'll work out well.

shukari:
> I cannot see the Apache License in your changes? Or are you trying to just relicense Apache license content as GPL? I don't think that'll work out well.

You now that i didnt do that on purpose.
But i see it now that this is something i forgot to check.
I made my own 3 white dots.

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -273,9 +273,20 @@ switch (_ctrlIDC) do {
...
+        // hide custom button items
+        if (!isNil QGVAR(customRightPanelButtons)) then {
+            private _blockItems = [];
+            {
+                if (!isNil "_x") then {
+                    _blockItems append (_x select 0);
+                };
+            } forEach GVAR(customRightPanelButtons);
+            
+            {
```
PabstMirror, line 285: move forEach loop out of isNil scope so it always runs

shukari: fixed, good spot

addons/arsenal/defines.hpp
```diff
@@ -304,7 +306,38 @@ _buttonCurrentMag2Ctrl ctrlCommit FADE_DELAY;\
...
     IDC_blockRighttBackground\
-];
+];\
+if (!isNil QGVAR(customRightPanelButtons)) then {\
+    private _miscOffset = 0;\
+    {\
+        if (!isNil "_x") then {\
+            _x params ["", "_picture", "_tooltip"];\
+            _miscOffset = _forEachIndex + 1;\
+            private _ctrl = _display ctrlCreate ["ace_arsenal_customArsenalButton_Button", 61 + (_forEachIndex * 2)];\
```
commy2, line 316: QGVAR(customArsenalButton_Button)

addons/arsenal/defines.hpp
```diff
@@ -304,7 +306,38 @@ _buttonCurrentMag2Ctrl ctrlCommit FADE_DELAY;\
...
+            _ctrl ctrlSetPosition [\
+                safezoneW + safezoneX - 10 * GRID_W,\
+                safezoneY + (88 + (10 * _forEachIndex)) * GRID_H\
+            ];\
+            _ctrl ctrlSetText _picture;\
+            _ctrl ctrlSetTooltip _tooltip;\
+            _ctrl ctrlCommit 0;\
+            _ctrl = _display ctrlCreate ["ace_arsenal_customArsenalButton_Background", 60 + (_forEachIndex * 2)];\
```
commy2, line 324: Same.

commy2:
Aside from these two, lgtm.

veteran29:
Icon of custom category disappears for me when I'm focusing something in the items list:
image
image

shukari:
@veteran29 good spot... the background was set after the icon... is fixed.

PR #7600, opened 2020-03-13
Tools - Add dummy files for check external

PabstMirror:
These are dummy files for p drive to allow building with pboProject (check externals now always enabled)
Normally extracted a3 is still required


commy2:
urg

PR #7607, opened 2020-03-16
Respawn - Add save gear event

jonpas:
- Title

Required for ACRE2 to add compatibility for proper radio assignment in combination with ACE Respawn's Save pre-death gear.

Not sure what else to put into the event, since all the data is on the unit anyways.

I am also pretty sure adding the event on killed event is fine, as long as whatever uses the event runs in unscheduled to make sure it happens before respawn kicks in.

### DIFF
--- a/addons/respawn/functions/fnc_handleKilled.sqf
+++ b/addons/respawn/functions/fnc_handleKilled.sqf
@@ -22,6 +22,7 @@ params ["_unit"];
 if (ACE_player == _unit && {GVAR(SavePreDeathGear)}) then {
     _unit setVariable [QGVAR(unitGear), getUnitLoadout _unit];
     _unit setVariable [QGVAR(activeWeaponAndMuzzle), [currentWeapon _unit, currentMuzzle _unit, currentWeaponMode _unit]];
+    [QGVAR(saveGear), _unit] call CBA_fnc_localEvent;
 };
 
 if (missionNamespace getVariable [QGVAR(showFriendlyFireMessage), false]) then {


jonpas:
Temporary mission script to sanitize ACRE radios:
```sqf
// initPlayerLocal.sqf
["ace_respawn_saveGear", {
    params ["_unit"];

    // Loadout
    private _loadout = getUnitLoadout _unit;

    // Remove "ItemRadioAcreFlagged"
    if ((_loadout select 9) select 2 == "ItemRadioAcreFlagged") then {
        (_loadout select 9) set [2, ""];
    };

    // Set ACRE base classes
    private _replaceRadioAcre = {
        params ["_item"];
        // Replace only if string (array can be eg. weapon inside container) and an ACRE radio
        if (!(_item isEqualType []) && {[_item] call FUNC(isRadio)}) then {
            _this set [0, [_item] call FUNC(getBaseRadio)];
        };
    };
    if !((_loadout select 3) isEqualTo []) then {
        {_x call _replaceRadioAcre} forEach ((_loadout select 3) select 1); // Uniform items
    };
    if !((_loadout select 4) isEqualTo []) then {
        {_x call _replaceRadioAcre} forEach ((_loadout select 4) select 1); // Vest items
    };
    if !((_loadout select 5) isEqualTo []) then {
        {_x call _replaceRadioAcre} forEach ((_loadout select 5) select 1); // Backpack items
    };

    _unit setVariable ["ace_respawn_unitGear", _loadout];
```

commy2:
Uhm, pass the saved gear + muzzle as arg?

jonpas:
I don't know if there is a point. You have to set the unit variables for any effect anyways.

dedmen:
I thought Arsenals baseItem/baseWeapon thing would solve this?

mharis001:
Misclick.

Issue #6928, opened 2019-04-09
XM307 Static can't be rearmed, when settings are "rearm whole magazine"


madpat3:
**Arma 3 Version:** `1.90` (stable)
**CBA Version:** `3.10.1` (stable)
**ACE3 Version:** `3.12.6` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- Title.

**Steps to reproduce:**
- Place a man, an ammo truck and a static XM307 on map. Empty the magazine of the weapon by attributes. Set Ace-logistic-settings to "rearm whole magazine". Now start the mission and try to rearm the weapon. Well in fact, you can do all the interaction stuff and all progress bars run through, but in the end, there won't be ammo in the weapon.  

**Where did the issue occur?**
- Dedicated / Editor (Singleplayer)

**Additional information:**
- Happens to all versions of the weapon. With XM312 (HMG) it works just fine.

jonpas:
Fixed in #7609.

PR #7609, opened 2020-03-20
Rearm - Fix vehicles failing rearm due to case differences

redbery:
- Remove case sensitivity from class check in getTurretMagazineAmmo

Currently the check in getTurretMagazineAmmo compares the class used in cfgVehicles' magazines array (within the turret) to the config name in cfgMagazines (returned by magazinesAllTurrets). In some cases developers have not used the same capitalization in both of these places. Within most systems this is not a problem as config classes are not case sensitive, but in the case of getTurretMagazineAmmo it means the function fails leading to a failure of the rearm action.

### DIFF
--- a/addons/rearm/functions/fnc_getTurretMagazineAmmo.sqf
+++ b/addons/rearm/functions/fnc_getTurretMagazineAmmo.sqf
@@ -23,5 +23,5 @@
 
 params ["_vehicle", "_turretPath", "_magazineClass"];
 
-private _ammo = magazinesAllTurrets _vehicle select {(_x select 0) isEqualTo _magazineClass && {(_x select 1) isEqualTo _turretPath}} apply {_x select 2};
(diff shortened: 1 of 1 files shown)


addons/rearm/functions/fnc_getTurretMagazineAmmo.sqf
```diff
@@ -23,5 +23,5 @@
 
 params ["_vehicle", "_turretPath", "_magazineClass"];
 
-private _ammo = magazinesAllTurrets _vehicle select {(_x select 0) isEqualTo _magazineClass && {(_x select 1) isEqualTo _turretPath}} apply {_x select 2};
+private _ammo = magazinesAllTurrets _vehicle select {(_x select 0) == _magazineClass && {(_x select 1) isEqualTo _turretPath}} apply {_x select 2};
 _ammo
```
mharis001:
```suggestion
magazinesAllTurrets _vehicle select {(_x select 0) == _magazineClass && {(_x select 1) isEqualTo _turretPath}} apply {_x select 2}
```

redbery:
This should fix https://github.com/acemod/ACE3/issues/6928

PR #7613, opened 2020-03-25
Refuel - Add enable setting

Dystopian:
- Close #7040.
- add refuel functionality from #7040 (which looks hard to merge): add enable setting and remove the need for compat addons;
- add support for vehicles without ACE refuel config;
- add `Land_FuelStation_03_pump_F` config;
- fix terrain fuel pumps support (not implemented in #7040);
- ~restore big explosion and long burn of refuel vehicles~;
- optimize some code.

- [x] add terrain pumps locations
- [x] check if `setFuelCargo` is persistent for terrain pumps
- [x] search for workaround for damaged pumps explosion strength


addons/refuel/CfgVehicles.hpp
```diff
@@ -1,3 +1,5 @@
+#define XEH_INHERITED class EventHandlers {class CBA_Extended_EventHandlers: CBA_Extended_EventHandlers {};}
```
commy2: Is this really needed?

Dystopian: macro or its using?

commy2: Both.

Dystopian: Macro: 1 line instead of 4. `(4-1)*7 - 1 = ` -20 lines from file. I think it's worth it. I would really prefer this macro in CBA like XEH_ENABLED.
Using: it handles pumps created during mission with `createVehicle` after mission init.

addons/refuel/XEH_postInit.sqf
```diff
@@ -1,7 +1,28 @@
 #include "script_component.hpp"
+if (!GVAR(enabled)) exitWith {};
+
+["All", "InitPost", {
```
Dystopian: Will it be better to replace `All` with each `cacheRefuelClasses`? It could skip at least `Man` initPost. I don't know XEH well enough.

jokoho48: with params index 4 you can add excluded Classes maybe adding CAManBase to that to remove some bulk

PabstMirror (dismissed):
Tested in local MP on stratis looking at a map pump `[getfuelCargo cursorObject, typeOf cursorObject]`
server `[0,"Land_FuelStation_Feed_F"]`
client `[1,"Land_FuelStation_Feed_F"]` and can use vanilla refuel
I don't know if `setFuelCargo` will work on map objects the way we want

Dystopian:
Looks like terrain pumps have local fuel value and they reset it in JIP. I fixed it.

I'm implementing pump locations caching now because `_worldCenter nearObjects [_x, _worldSize]` is quite long operation on big maps.

PabstMirror:
_worldSize * sqrt(2) / 2 but I don't think it will help much 😄 

I'm still worried about if `setFuelCargo` will be persistent, 
from what I understand map objects can be "streamed" in and out of memory and changes to them may be lost

e.g. setVariable may work initially but might be lost if player moves away from the area

Dystopian:
Added pumps locations. I don't own Contact DLC so who wants please try `call compile preprocessFileLineNumbers "z\ace\addons\refuel\dev\exportTerrainRefuelPositions.sqf"` in Eden Editor console when Livonia is loaded.
Will check if `setFuelCargo` is persistent. Also will search for workaround for damaged pumps explosion strength.

Dystopian:
Now destroyed refuel vehicles burn much longer - like did in vanilla. I didn't find good enough way to catch deadly hit so I implemented it with `HandleDamage` and `alive` on next frame. It doesn't handle 100% but it's good enough.

> I'm still worried about if `setFuelCargo` will be persistent,
> from what I understand map objects can be "streamed" in and out of memory and changes to them may be lost

Just checked if `setFuelCargo` is persistent in 2 hours MP session on Chernarus 2020. No pumps lost their fuel cargo. I won't be able to check it in longer session. Script returns all pumps count on map (should be the same) and array of desynced pumps (should be empty):
```
private _a = [];
private _c = 0;
{
    _x params ["_class", "_positions"];
    {
        {
            if (0 != getFuelCargo _x) then {
                _a pushBack ASLToAGL getPosASL _x;
            };
            _c = _c + 1;
        } forEach (_x nearObjects [_class, 30]);
    } forEach _positions;
} forEach getArray (configFile >> "ace_refuel_positions" >> worldName);
[_c, _a]
```

bux:
> I don't own Contact DLC so who wants please try call compile preprocessFileLineNumbers "z\ace\addons\refuel\dev\exportTerrainRefuelPositions.sqf" in Eden Editor console when Livonia is loaded.

```    
    Enoch[] = { /*  Livonia  */
        {"Land_FuelStation_Feed_F", {{10208,2173,0}}},
        {"Land_FuelStation_03_pump_F", {{2008,7365,0},{6259,3949,0}}}
    };
```

PabstMirror:
I was curious about another way to get map objects using
`[2708,5788,0] nearestObject 65997`
which is some kind of terrain builder ID
but doesn't seem to be any better than
`[2708,5788,0] nearObjects ["Land_FuelStation_Feed_F", 30];`

addons/refuel/initSettings.sqf
```diff
@@ -1,9 +1,18 @@
...
+[
+    QGVAR(enabled),
+    "CHECKBOX",
+    ELSTRING(common,Enabled),
+    [LELSTRING(OptionsMenu,CategoryLogistics), localize "str_state_refuel"],
+    true,
+    true,
+    {[QGVAR(enabled), _this] call EFUNC(common,cbaSettings_settingChanged)},
+    true // Needs mission restart
+] call CBA_Settings_fnc_init;
```
commy2, line 10: CBA_fnc_addSetting.

addons/refuel/XEH_postInit.sqf
```diff
@@ -1,9 +1,77 @@
...
+    params ["_vehicle"];
+    if (getFuelCargo _vehicle <= 0) exitWith {};
+
+    if (local _vehicle) then {
+        _vehicle setFuelCargo 0;
+    };
+    if (-1 == _vehicle getVariable [QGVAR(HDEHID), -1]) then {
```
commy2, line 11: Please stop doing that.

```suggestion
    if (_vehicle getVariable [QGVAR(HDEHID), -1] == -1) then {
```
and anywhere else.

Dystopian: I would understand it for lines like `_fuel == 1` end maybe even `getFuelCargo _x < 1`. For such short conditions I usually use common conditions. But for long expression yoda condition looks much easier for me. I'm sorry but no.

addons/refuel/ACE_Refuel_Positions.hpp
LinkIsGrim, line 1: Tell me we have a better way of doing this.

johnb432: I imagine you have already seen [this](https://github.com/acemod/ACE3/pull/7613/files#diff-f835b458067331321d4b13bba6c5546c5f7390d1643230d40a28ac18839f9900R53-R66) - the version for if there is no config for a particular map.

I don't think there is a better way of doing it more quickly unfortunately. The downside to the config version is that if the map updates a fuel storage position (either changing existing ones or adding new ones), it will not account for said changes.

The dynamic version however always works. I'm tempted to say drop the config implementation, but on the other hand some maps are *very* stable (meaning no terrain changes) and the performance benefit is probably noticeable.
On Takistan with a rather large modset the execution time is 77ms for the dynamic version, whereas the config version takes 2 ms.

addons/refuel/XEH_postInit.sqf
```diff
@@ -1,9 +1,77 @@
...
+    // placed in editor vehicles both trigger init and synchronize fuel cargo
+    {
+        if (0 < getFuelCargo _x) then {
+            if (local _x) then {_x setFuelCargo 0;};
+            _x setVariable [QGVAR(HDEHID), _x addEventHandler ["HandleDamage", LINKFUNC(handleDamage)]];
+        };
+    } forEach allMissionObjects "";
+} else {
+    // here are both terrain and editor static objects
+    WARNING_2("World %1: no configured %2; can be loaded slower",worldName,QGVAR(positions));
```
johnb432, line 54:
```suggestion
    WARNING_2("World %1: %2 is not configured; can load slower",worldName,QGVAR(positions));
```

addons/refuel/ACE_Refuel_Positions.hpp
johnb432: @LinkIsGrim I don't think there is a better way. Do you still want to explore options or can we consider this comment as resolved?

addons/refuel/dev/exportTerrainRefuelPositions.sqf
```diff
@@ -0,0 +1,84 @@
+// call compile preprocessFileLineNumbers "z\ace\addons\refuel\dev\exportTerrainRefuelPositions.sqf"
```
LinkIsGrim, line 1:
```suggestion
// call compileScript ["z\ace\addons\refuel\dev\exportTerrainRefuelPositions.sqf"]
```

addons/refuel/ACE_Refuel_Positions.hpp
LinkIsGrim: Resolved. It's ugly, but it's fine as long as there's working scripts for config generation.

addons/refuel/XEH_preStart.sqf
```diff
@@ -4,27 +4,44 @@
...
+private _cacheRefuelCargo = createHashMap;
 
 {
-    private _fuelCargo = getNumber (_x >> QGVAR(fuelCargo));
+    private _transportFuel = getNumber (_x >> "transportFuel");
+    private _fuelCargo = [_x >> QGVAR(fuelCargo), "NUMBER", _transportFuel] call CBA_fnc_getConfigEntry;
     if (_fuelCargo > 0 || {_fuelCargo == REFUEL_INFINITE_FUEL}) then {
         private _sourceClass = configName _x;
+        private _noXEH = !isText (_x >> "EventHandlers" >> "CBA_Extended_EventHandlers" >> "init");
+        private _isPublic = 2 == getNumber (_x >> "scope");
```
LinkIsGrim, line 17:
```suggestion
        private _isPublic = getNumber (_x >> "scope") == 2;
```

addons/refuel/XEH_preStart.sqf
```diff
@@ -4,27 +4,44 @@
...
+            || {_sourceClass isKindOf "Static"} // CBA_fnc_addClassEventHandler doesn't support "Static" class
         ) then {
-            if (2 == getNumber (_x >> "scope")) then {
-                _staticClasses pushBackUnique _sourceClass;
+            if (_isPublic) then {
+                if (_noXEH) then {
+                    WARNING_3("Class %1: %2 [%3] needs XEH",_sourceClass,configName inheritsFrom _x,configSourceMod _x);
+                };
+                _staticClasses pushBack _sourceClass;
+                if (-1 == _baseStaticClasses findIf {_sourceClass isKindOf _x}) then {
```
LinkIsGrim, line 28:
```suggestion
                if (_baseStaticClasses findIf {_sourceClass isKindOf _x} == -1) then {
```

addons/refuel/XEH_preStart.sqf
```diff
@@ -4,27 +4,44 @@
...
+                    _baseStaticClasses pushBack _sourceClass;
+                };
             };
         } else {
-            if (-1 == _dynamicClasses findIf {_sourceClass isKindOf _x}) then {
-                _dynamicClasses pushBackUnique _sourceClass;
+            if (-1 == _baseDynamicClasses findIf {_sourceClass isKindOf _x}) then {
```
LinkIsGrim, line 33:
```suggestion
            if (_baseDynamicClasses findIf {_sourceClass isKindOf _x} == -1) then {
```

addons/refuel/XEH_postInit.sqf
```diff
@@ -1,9 +1,84 @@
 #include "script_component.hpp"
+if (!GVAR(enabled)) exitWith {};
```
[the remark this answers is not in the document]
LinkIsGrim: Then this entire file needs to be wrapped in the event

addons/refuel/XEH_postInit.sqf
```diff
@@ -1,9 +1,84 @@
...
+            _x setVariable [QGVAR(HDEHID), _x addEventHandler ["HandleDamage", LINKFUNC(handleDamage)]];
+        };
+    } forEach allMissionObjects "";
+} else {
+    // here are both terrain and editor static objects
+    WARNING_2("World %1: %2 is not configured; can load slower",worldName,QGVAR(positions));
+    private _halfWorldSize = worldSize / 2;
+    private _worldCenter = [_halfWorldSize, _halfWorldSize];
+    _halfWorldSize = _halfWorldSize * sqrt 2;
+    private _refuelMissionObjects = allMissionObjects "" select {getFuelCargo _x > 0};
```
LinkIsGrim, line 62: use https://community.bistudio.com/wiki/allObjects instead?

johnb432: On Altis `-1 allObjects 0` is about 30% slowed than `allMissionObjects ""`.

LinkIsGrim: `9 allObjects 0`?

johnb432: `9 allObjects 0` is about 12% faster than `allMissionObjects ""`, although `allMissionObjects "all"` is even faster than `9 allObjects 0`, albeit by a small margin.

I'm a bit worried about the note on the `allObjects` wiki page:

> Some terrain placed objects are not part of slow vehicles collection, use [nearestTerrainObjects](https://community.bistudio.com/wiki/nearestTerrainObjects) command to return these objects

That means there are some buildings that are part of it - which ones? We don't want terrain buildings, but do we want buildings that were placed down in the editor?

Dystopian: `allMissionObjects` returns only editor placed objects. Usually on init there are not so many objects. Why do you want to replace it with `allObjects`?

Dystopian:
I've updated `Chernarus 2020` config and added some more maps. And because export algo is not ideal I got `WRONG COUNT` message for `Chernarus 2020` and `Rosche` and `INCREASE DISTANCE` for `Farabad`. In all cases I found close placed positions and replaced them with manual position. Thanks @johnb432 for added sorting :)

johnb432:
>restore big explosion and long burn of refuel vehicles;

I was testing this. Self-hosted, with a second PC joining as a client.

When the host shot a gas pump on Altis, it would burn long for both host and client. However, when the client shot a pump, it was a long burn for the host but a short one for the client.

Dystopian:
> shot a gas pump on Altis

I always tested on Stratis and it worked. But it really doesn't work on Altis. I don't have more ideas how to handle this so I decided to remove that PR part. Also I removed it for vehicles because loop logic becomes too messy without terrain pumps. Maybe it would be implemented somehow in another PR but I don't think from me :-(

addons/refuel/functions/fnc_makeSource.sqf
```diff
@@ -25,19 +25,22 @@ if !(EGVAR(common,settingsInitFinished)) exitWith {
...
-private _fuelCargoConfig = getNumber (configOf _source >> QGVAR(fuelCargo));
+private _fuelCargoConfig = _source call FUNC(getFuelCargo);
+
+TRACE_4("makeSource",_source,_fuelCargo,_hooks,_fuelCargoConfig);
 
 if (
     isNull _source
     || {_fuelCargo < 0 && {!(_fuelCargo in [REFUEL_INFINITE_FUEL, REFUEL_DISABLED_FUEL])}}
-    || {_fuelCargo != 0 && {_fuelCargo == _fuelCargoConfig}}
+    || {_fuelCargo == REFUEL_DISABLED_FUEL && {_fuelCargoConfig == REFUEL_DISABLED_FUEL}}
```
johnb432, line 43: If you make something into a fuel source, then want to "unmake" it, it fails here.

I don't think the previous code supported it, but it's something that should be looked into.

johnb432 (approved):
LGTM. I've tested some, but it could use some more.

Dystopian:
> I don't have more ideas

I tried to `setFuelCargo 1` in `BuildingChanged` Mission Event Handler but it doesn't work even in SP.

johnb432:
> I tried to `setFuelCargo 1` in `BuildingChanged` Mission Event Handler but it doesn't work even in SP.

Don't bother with trying to find a solution. If you do, it's best to make another PR at this point imo.

LinkIsGrim (approved):
LGTM

PR #7619, opened 2020-04-01
CSW - Add configs for 1.98's M2 Turret

PabstMirror:


addons/csw/CfgWeapons.hpp
```diff
@@ -179,6 +179,20 @@ class CfgWeapons {
...
     };
+    class GVAR(staticM2ShieldCarry): GVAR(staticHMGCarry) {
+        class ADDON: ADDON {
+            class assembleTo {
+                GVAR(m3Tripod) = "B_HMG_02_high_F";
+                GVAR(m3TripodLow) = "B_HMG_02_F";
+            };
+        };
+        class WeaponSlotsInfo: WeaponSlotsInfo {
+            mass = 1000;
```
TheCandianVendingMachine, line 190: I think this means the mass is defined as 100lbs. In real life the mass is 84lbs according to wIkipedia

PabstMirror: I was thinking the shield should add some heft

TheCandianVendingMachine: yeah fair actually

PR #7622, opened 2020-04-04
Explosives - Remove "Activate mine" action menu item

Saborknight:
The "Activate mine" action menu item was still showing, but only after an explosive was set, then diffused.

- Remove the "Activate mine" action menu item on a diffused explosive/mine

### DIFF
--- a/addons/explosives/config.cpp
+++ b/addons/explosives/config.cpp
@@ -47,7 +47,7 @@ class CfgActions {
     class DeactivateMine:None {
         show = 0;
     };
-    class UseContainerMagazine:None {
+    class UseContainerMagazine: None {
         show = 0;
     };
 };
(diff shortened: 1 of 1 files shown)


addons/explosives/config.cpp
```diff
@@ -47,6 +47,9 @@ class CfgActions {
     class DeactivateMine:None {
         show = 0;
     };
+    class UseContainerMagazine:None {
```
commy2, line 50:
```suggestion
    class UseContainerMagazine: None {
```

commy2:
Do we know what other uses `UseContainerMagazine` may have, or is this name just especially terrible?

Saborknight:
I haven't tested, but the attribute doesn't have any other definition other than "Activate mine". This sounds very specific to me, even if the classname is particularly terrible

PR #7624, opened 2020-04-05
Reload - Fix check ammo display for empty GL launchers

PabstMirror:
After shooting a rifle's GL, hitting check ammo would show there was a loaded shell and an incorrect mag count.

Now on empty it displays the count of total compatible mags
That code wasn't reachable before because there is a `if (_magazine == "") exitWith {};` at the top but seems to be the intent.

### DIFF
--- a/addons/reload/functions/fnc_displayAmmo.sqf
+++ b/addons/reload/functions/fnc_displayAmmo.sqf
@@ -42,7 +42,7 @@ if (_target isKindOf "StaticWeapon") then {
             };
         } forEach _magazines;
     };
-    
+
     // For static weapons the muzzle seemingly never returns anything for static weapons with/without people inside
     if (_muzzle == "") then {
         _muzzle = _weapon;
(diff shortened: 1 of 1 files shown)


addons/reload/functions/fnc_displayAmmo.sqf
```diff
@@ -76,10 +76,14 @@ if (_muzzle == _weapon) then {
...
 
+    _ammo = _target ammo _muzzle;
+    if (_ammo == 0) then { _magazine = ""; };
+
     _count = if (_magazine != "") then {
         {_x == _magazine} count (magazines _target + [_magazine])
     } else {
-        {_x in getArray (configFile >> "CfgWeapons" >> _weapon >> _muzzle >> "Magazines")} count magazines _target
+        private _compatibleMagazines =  [configFile >> "CfgWeapons" >> _weapon >> _muzzle] call CBA_fnc_compatibleMagazines;
+        {_x in _compatibleMagazines} count magazines _target
```
commy2, line 86: What about the 3 round gl mags?

PabstMirror: This doesn't really change how they were handled
Maybe if magazine count > 1 then show the ammo bars like a normal magazine?

PabstMirror:
Primary rifle unchanged
1xGL Loaded: Show mag and count of similar mags
3xGL or 10x 50cal Loaded: show ammo count like primary mag
1-3x gl unloaded: show count of any compatible GL mags
10x 50 empty: show ammo count like primary mag

PR #7625, opened 2020-04-06
WIP: Maps - Add topographic map item

PabstMirror:
Adds a simple topo map that doesn't magically show every single building, wall, rock and tree

New:

Old (sat tex off):

Creates a new map ctrl over the old one
this breaks **anything** that uses drawEH on the old map (`dspCtrl 51`)
but surprisingly button down stuff still seems to works fine, same with markers for the most part
ace already has decent support for non "itemMap"

ideas:
- [ ] probably re-add hill numbers
- [ ] maybe show urban areas with some kind of shading?

### DIFF
--- a/addons/map/CfgWeapons.hpp
+++ b/addons/map/CfgWeapons.hpp
@@ -23,4 +23,8 @@ class CfgWeapons {
             };
         };
     };
+    class ItemMap;
+    class ace_TopographicMap: ItemMap {
+        displayName = "Topo test";
+    };
 };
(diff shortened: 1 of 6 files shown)


PabstMirror:
So this got a lot more complex than I thought, most of the diff is just moving code over to a new helper: `ace_common_fnc_addMapEventHandler` for adding events to maps
- proper save/load support (markers/maptool did not support)
- runs events on the correct map, since there may be more than one now

markers, maptools, gestures, flashlight effects all seem to work correctly
the only bug right now is vanilla line drawing does not work on the new map

jonpas:
> the only bug right now is vanilla line drawing does not work on the new map

Could add a note in a corner or something that drawing is not supported?

veteran29:
> Could add a note in a corner or something that drawing is not supported?

It could be re-implemented with "new" commands that allow creating polylines.

PR #7640, opened 2020-04-16
Dragging - Add possiblity to rotate a carried object

bux:
- title
- use CBA_fnc_addDisplayHandler
- fix wrong number of param in showMouseHint comment


bux:
Example: https://streamable.com/qfaxkv

addons/dragging/functions/fnc_onKeyDown.sqf
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Public: No
+ */
+
+// EH is always assigned. Exit and don't overwrite input if not carrying
+if !(ace_player getVariable [QGVAR(isCarrying), false]) exitWith {false};
+
+params ["", "_key", "_shift", "_ctrl", "_alt"];
+
+GVAR(ctrlHeld) = _ctrl;
```
PabstMirror, line 24: just use `cba_events_control`

PabstMirror: e.g. https://github.com/acemod/ACE3/blob/master/addons/explosives/functions/fnc_openTimerUI.sqf#L58

bux: Nice! That's so much better :)

addons/dragging/stringtable.xml
```diff
@@ -86,21 +86,9 @@
...
-            <Italian>Alza/Abbassa</Italian>
-            <Spanish>Subir/Bajar</Spanish>
-            <French>Lever/Baisser</French>
-            <Japanese>上げる/下げる</Japanese>
-            <Korean>높이기/낮추기</Korean>
-            <Chinese>提高/下降</Chinese>
-            <Chinesesimp>提高/下降</Chinesesimp>
-            <Turkish>Yukarı/Aşağı</Turkish>
+        <Key ID="STR_ACE_Dragging_RaiseLowerRotate">
+            <English>Raise/Lower | (Ctrl + Scroll) Rotate</English>
```
commy2, line 90:
```suggestion
            <English>Raise/Lower | (Ctrl + Scroll Wheel) Rotate</English>
```

addons/dragging/stringtable.xml
```diff
@@ -86,21 +86,9 @@
...
+            <German>Heben/Senken | (Strg + Scrollen) Drehen</German>
```
commy2, line 91:
```suggestion
            <German>Heben/Senken | (Strg + Scrollrad) Drehen</German>
```

Scroll is not a key. It shoul

bux: Same as here:
https://github.com/acemod/ACE3/blob/master/addons/advanced_throwing/stringtable.xml#L247

addons/dragging/functions/fnc_handleScrollWheel.sqf
```diff
@@ -17,35 +17,47 @@
...
 
-private _unit = ACE_player;
-
 // EH is always assigned. Exit and don't overwrite input if not carrying
-if !(_unit getVariable [QGVAR(isCarrying), false]) exitWith {false};
+if !(ace_player getVariable [QGVAR(isCarrying), false]) exitWith {false};
 
-// move carried item 15 cm per scroll interval
-_scrollAmount = _scrollAmount * 0.15;
+private _unit = ace_player;
```
commy2, line 23: ACE_player.

Why move it down here though?

bux: I don't want to create a variable each time the player uses it's scrollwheel without carrying an item.

addons/dragging/stringtable.xml
commy2: Then change it there as well.

addons/dragging/functions/fnc_handleScrollWheel.sqf
commy2: You're not going to save a single frame by this in 1000 hours of playing A3 while continiuously turning the mouse wheel.
I don't even think your mouse will survive before you saved one second.

addons/dragging/stringtable.xml
bux: No. Can be done in a separate PR.

jonpas: No, do not change it. Scroll is scroll wheel rolling there is also a Middle Mouse Button as key which is also used in Advanced Throwing.

PabstMirror:
Would this be any better, it matches what we use on acex_fortify:

bux:
I used the same as in adv. throwing. Might be for a future improvement where we change this to use a small key cap icon <kbd>Ctrl</kbd>

addons/dragging/functions/fnc_carryObject.sqf
```diff
@@ -63,7 +63,7 @@ _unit setVariable [QGVAR(ReleaseActionID), [
 if (_target isKindOf "CAManBase") then {
     [localize LSTRING(Drop), "", ""] call EFUNC(interaction,showMouseHint);
 } else {
-    [localize LSTRING(Drop), "", localize LSTRING(LowerRaise)] call EFUNC(interaction,showMouseHint);
+    [localize LSTRING(Drop), "", localize LSTRING(RaiseLowerRotate)] call EFUNC(interaction,showMouseHint);
```
PabstMirror, line 66:
```suggestion
    [localize LSTRING(Drop), "", LLSTRING(LowerRaise), [["control", LLSTRING(rotate)]]] call EFUNC(interaction,showMouseHint);
```
The advantage of doing this now is we don't have to drop our existing translations

bux: It's hideous, though

PR #7643, opened 2020-04-18
Medical - Improve the fix for being stuck in unconscious after ragdoll

ZDatOneGuy:
- Change the way units are put into the unconscious animation by instead using an AnimChanged event handler that only runs while the unit is unconscious
- Remove the class changes in CfgExtendedAnimation
- Remove the existing AnimDone fix that runs on all units

Note:
If the unit wakes up and setUnconscious is set to false before they are unragdolled they seem to get stuck on their back for awhile, this seems to be a problem in vanilla also. If a unit is already in setUnconcious before being knocked ACE unconscious then the animation won't change but it shouldn't be a major issue.


addons/medical_engine/functions/fnc_unconAnimEH.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * This function should only be called on by the EH because of the reliance on _thisType and _thisID
+ *
+ * Public: No
+ */
+ 
+params ["_unit", "_anim"];
+
+//When unit wakes up remove the event handler
+if(!(_unit getVariable ["ACE_isUnconscious", true])) exitWith {
+    _unit removeEventHandler [_thisType, _thisID];
```
commy2, line 23: No. Do not constantly add and remove event handlers.

addons/medical_engine/CfgExtendedAnimation.hpp
```diff
@@ -1,10 +0,0 @@
-
-// we want the face down animation every time
-class CfgExtendedAnimation {
-    class Revive {
-        left = "Unconscious";
-        right = "Unconscious";
-        front = "Unconscious";
-        back = "Unconscious";
-    };
-};
```
commy2, line 10: This is required.

ZDatOneGuy:
Hmm, I've not noticed anything wrong with having event handlers remove themselves as it's better than having them run 24/7 but most of my experience is exclusively mission scripting. As for CfgExtendedAnimation, it seems to be the cause of the issue but I'll see if I can find a way to keep it while finding a better way than slapping an AnimDone EH on everyone

ZDatOneGuy:
For questions and comments should I just post here?

SilentSpike:
@ZDatOneGuy Yeah best to keep all related discussion in one place for future readers

ZDatOneGuy:
Alright. So to test, I isolated the CfgExtendedAnimation into it's own mod and loaded just it and CBA. The problem doesn't exist in vanilla since Unconscious connects to Incapacitated. ACE removes that connection, most likely to stop people from changing out of the animation from before the medical rewrite correct? With the use of setUnconscious that should no longer be a problem so a simple change in CfgMoves in common solves this issue. Just to be sure I'll test a bit longer but I'm pretty sure setUnconscious is pretty strict in locking units down

ZDatOneGuy:
Seems like everything works just fine since the animation is only ever set alongside the ace unconscious state anyways. I'll adjust my pull request

commy2:
Now you deleted the fix required for handling ragdolling after crashing into a vehicle.

ZDatOneGuy:
Well that's not what I was trying to do XD

ZDatOneGuy:
Well I am quite unfamiliar with github here so any assistance would be great

Mike-MF:

PR #7650, opened 2020-04-19
Interaction - Show current team color in Team Management menu

veteran29:
- title

### DIFF
--- a/addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
+++ b/addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
@@ -1,13 +1,12 @@
 #include "script_component.hpp"
 /*
  * Author: veteran29
- * Modifies the ACE_TeamManagement action to show current group color.
+ * Modifies the ACE_TeamManagement or join team action with given group color and icon.
  *
  * Arguments:
- * 0: Target <OBJECT>
- * 1: Player <OBJECT>
- * 2: Args <ANY>
- * 3: Action Data <ARRAY>
+ * 0: Team color <STRING>
+ * 1: Icon to show on action <STRING>
+ * 2: Action Data to modifiy <ARRAY>
  *
  * Return Value:
  * None
--- a/addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
+++ b/addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
@@ -19,13 +19,13 @@
 params ["_target", "", "", "_actionData"];
 
 private _color = switch (assignedTeam _target) do {
-    case "RED": {"#FF0000"};
-    case "GREEN": {"#00FF00"};
-    case "BLUE": {"#0000FF"};
-    case "YELLOW": {"#FFFF00"};
-    default {"#FFFFFF"};
+    case "RED": {missionNamespace getVariable [QEGVAR(nametags,nametagColorRed), [221, 0, 0]]};
+    case "GREEN": {missionNamespace getVariable [QEGVAR(nametags,nametagColorGreen), [0, 221, 0]]};
+    case "BLUE": {missionNamespace getVariable [QEGVAR(nametags,nametagColorBlue), [0, 0, 221]]};
+    case "YELLOW": {missionNamespace getVariable [QEGVAR(nametags,nametagColorYellow), [221, 221, 0]]};
+    default {missionNamespace getVariable [QEGVAR(nametags,nametagColorMain), [255,255,255]]};
 };
 
-_actionData set [2, [QPATHTOF(UI\team\team_management_ca.paa), _color]];
+_actionData set [2, [QPATHTOF(UI\team\team_management_ca.paa), _color call BIS_fnc_colorRGBtoHTML]];
 
 nil
--- a/addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
+++ b/addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
@@ -16,9 +16,9 @@
  * Public: No
  */
 
-params ["_target", "", "", "_actionData"];
+params ["_teamColor", "_icon", "_actionData"];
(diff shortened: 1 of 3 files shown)


addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+params ["_target", "", "", "_actionData"];
+
+private _color = switch (assignedTeam _target) do {
+    case "RED": {"#FF0000"};
+    case "GREEN": {"#00FF00"};
+    case "BLUE": {"#0000FF"};
+    case "YELLOW": {"#FFFF00"};
+    default {"#FFFFFF"};
```
commy2, line 26: https://github.com/acemod/ACE3/blob/master/addons/nametags/initSettings.sqf#L12-L56

addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
```diff
@@ -0,0 +1,31 @@
...
+};
+
+_actionData set [2, [QPATHTOF(UI\team\team_management_ca.paa), _color]];
+
+nil
```
commy2, line 31: `set` already reports `nil` though.

addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
veteran29: This is not related to nametags. The assign team color icons are fixed color too. But I can use these if you want.

bux: https://community.bistudio.com/wiki/assignTeam

commy2: At least use softer colors than FF0000 etc.

addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
jonpas: Explicit is better than implicit. What if some more code gets added after the `set`, it's way too easy to miss that the function has to return `nil`.

addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
jonpas: Nametag ones are specifically lower contrast if I remember correctly.

I don't see an issue in using same colors for both though.

veteran29: Personally I think they should use same colors as rest of the icons. So I have two proposals:
1. I make 4 additional colored icons like in right side of my screenshot.
2. I remove colored icons for team assignment and tint them with nametag color. If no nametag component I will use fixed color.

jonpas: I would go for 2. No need to complicate this, doesn't have to be color blind safe either since there is text right next to it.

veteran29:
Changed both root and submenu to use nametags color if component is available.

addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+params ["_teamColor", "_icon", "_actionData"];
+
+private _color = switch (toUpper _teamColor) do {
+    case "RED": {missionNamespace getVariable [QEGVAR(nametags,nametagColorRed), [221, 0, 0]]};
+    case "GREEN": {missionNamespace getVariable [QEGVAR(nametags,nametagColorGreen), [0, 221, 0]]};
+    case "BLUE": {missionNamespace getVariable [QEGVAR(nametags,nametagColorBlue), [0, 0, 221]]};
+    case "YELLOW": {missionNamespace getVariable [QEGVAR(nametags,nametagColorYellow), [221, 221, 0]]};
+    default {missionNamespace getVariable [QEGVAR(nametags,nametagColorMain), [255,255,255]]};
```
jonpas, line 26:
```suggestion
    default {missionNamespace getVariable [QEGVAR(nametags,nametagColorMain), [255, 255, 255]]};
```

addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
```diff
@@ -0,0 +1,31 @@
+#include "script_component.hpp"
+/*
+ * Author: veteran29
+ * Modifies the ACE_TeamManagement action to show current group color.
+ *
+ * Arguments:
+ * 0: Target <OBJECT>
```
synixebrett, line 7: This doesn't match `params`?

veteran29: Yep, forgot to change.

addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
```diff
@@ -0,0 +1,31 @@
...
+ * 1: Player <OBJECT>
+ * 2: Args <ANY>
+ * 3: Action Data <ARRAY>
```
veteran29, line 10:
```suggestion
 * 0: Team color <STRING>
 * 1: Icon to show on action <STRING>
 * 2: Action Data to modifiy <ARRAY>
```

addons/interaction/functions/fnc_modifyTeamManagementAction.sqf
```diff
@@ -0,0 +1,31 @@
+#include "script_component.hpp"
+/*
+ * Author: veteran29
+ * Modifies the ACE_TeamManagement action to show current group color.
```
veteran29, line 4:
```suggestion
 * Modifies the ACE_TeamManagement or join team action with given group color and icon.
```

veteran29:
Should be good now.

PR #7656, opened 2020-04-21
Translations - Add French Dragging RaiseLowerRotate key

Elgin675:
Add french translation for the `STR_ACE_Dragging_RaiseLowerRotate` key.


bux:
I'd prefer if it would be `baiser` 😚

Elgin675:
> 
> 
> I'd prefer if it would be `baiser` 😚

I think military regulations would prohibit it :D ^^.

PR #7661, opened 2020-04-24
Common - Fix ACE displayTextStructured overlap CBA_fnc_notify

Vdauphin:
ACE displayTextStructured overlap `CBA_fnc_notify`. This pull request add `CBA_fnc_notify`. So all mod/mission using `CBA_fnc_notify` will not overlap with ACE notification.

- Use `CBA_fnc_notify` and not `EFUNC(common,displayTextStructured)` when possible

### DIFF
--- a/addons/ai/functions/fnc_garrison.sqf
+++ b/addons/ai/functions/fnc_garrison.sqf
@@ -30,12 +30,12 @@ private _currentUnitMoveList = missionNameSpace getVariable [QGVAR(garrison_unit
 
 if (_startingPos isEqualTo [0,0,0]) exitWith {
     TRACE_1("fnc_garrison: StartingPos error",_startingPos);
-    [LSTRING(GarrisonInvalidPosition)] call EFUNC(common,displayTextStructured);
+    [LSTRING(GarrisonInvalidPosition)] call CBA_fnc_notify;
 };
 
 if (count _unitsArray == 0 || {isNull (_unitsArray select 0)}) exitWith {
     TRACE_1("fnc_garrison: Units error",_unitsArray);
-    [LSTRING(GarrisonNoUnits)] call EFUNC(common,displayTextStructured);
+    [LSTRING(GarrisonNoUnits)] call CBA_fnc_notify;
 };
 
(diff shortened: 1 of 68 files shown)


jonpas:
A year later, but is this ready or why is it still a draft?

Vdauphin:
The main problem is that we can't use `CBA_fnc_notify` everywhere
Like here: https://github.com/acemod/ACE3/blob/master/addons/reload/functions/fnc_displayAmmo.sqf#L113-L114
because `CBA_fnc_notify` doesn't support Structured Text

I don't know if we want both function (`CBA_fnc_notify` and `EFUNC(common,displayTextStructured)`) 
I didn't get any input when I asked in slack

jonpas:
Hmm.

Maybe a better way is to modify `displayTextStructured` to call `CBA_fnc_notify` when it can? I am not sure really. Maybe there is a way to use `CBA_fnc_notify` for everything with minor modifications? I haven't looked at it in a while.

Vdauphin:
> Maybe a better way is to modify displayTextStructured to call CBA_fnc_notify when it can? I am not sure really. 

It is a public function so I don't know if I can do that without changing the behavior of the function. In particular with the size parameter which is also available in `CBA_fnc_notify`
I can try to add something like that:
```sqf
if isStructuredText then
       use the old behavior
else 
     deal with the string to display and size parameter
      call cba_fnc_notify

``` 
But sadly the `_width` parameter will be ignore in case the text is just a string

Or do it like it:
```sqf
if isStructuredText and `_width` parameter provided then
       use the old behavior
else 
     deal with the string to display and size parameter
      call cba_fnc_notify

``` 
> Maybe there is a way to use CBA_fnc_notify for everything with minor modifications?

I think the last option will use `CBA_fnc_notify` for minor notifications (when you just pass a string as parameter to the `displayTextStructured`)

Tell me what you prefer or close it as `CBA_fnc_notify` lack support of structured text

jonpas:
Hmm, yeah I see.

I'll think about it and hopefully someone pitches in in the meantime as well.

Mike-MF:

PR #7662, opened 2020-04-24
Common - fix player's name showing as "Unknown"

commy2:
- Should fix "Unknown" being reported by getName in certain situations.


PabstMirror:
https://github.com/acemod/ACE3/blob/a8856e0bc86cd3e61f91e7cb67f73fe4f15793e4/addons/common/CfgEventHandlers.hpp#L48

anyone know why we add it to `All` instead of `CaManBase`?

commy2:
There is no reason, but also no harm, as respawning vehicles don't raise the event.

PR #7669, opened 2020-04-27
Nightvision - Increase max amount of IR rays to 64

Neviothr:
- Increase the maximum amount of rendered IR rays to 64, as mentioned [here](https://youtu.be/sOrUP0Rq6Jg?t=498).

### DIFF
--- a/addons/nightvision/CfgIRLaserSettings.hpp
+++ b/addons/nightvision/CfgIRLaserSettings.hpp
@@ -1,3 +1,3 @@
 class CfgIRLaserSettings {
-    maxNumberOfRays = 64;
+    maxNumberOfRays = 64; // Default is 24.
 };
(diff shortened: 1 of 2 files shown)


veteran29:
What happens to dynamic lights if there's 64 rays? Will they be disabled?

Neviothr:
I haven't noticed anything different with dynamic lights. Are they supposed to be affected by this?

BaerMitUmlaut:
Lasers are not light sources, they're objects with an emissive material (which means they will always be drawn at full brightness and not be affected by shadows, they do not emit light). They should not affect light sources at all.

dedmen:
> They should not affect light sources at all.

Yes they don't.
Also for reference the default number of rays is 24.

Neviothr:
> > They should not affect light sources at all.
> 
> Yes they don't.
> Also for reference the default number of rays is 24.

Added as a comment just in case.

PR #7699, opened 2020-05-08
Advanced Ballistics - Improve Performance in handleFirePFH

jokoho48:
- Changes the method of how we "clean" the array

### DIFF
--- a/addons/advanced_ballistics/XEH_postInit.sqf
+++ b/addons/advanced_ballistics/XEH_postInit.sqf
@@ -23,7 +23,7 @@ if (!hasInterface) exitWith {};
     ["ace_firedPlayerNonLocal", DFUNC(handleFired)] call CBA_fnc_addEventHandler;
 
     // Register Perframe Handler
-    [{call FUNC(handleFirePFH)}, GVAR(simulationInterval)] call CBA_fnc_addPerFrameHandler;
+    [FUNC(handleFirePFH), GVAR(simulationInterval)] call CBA_fnc_addPerFrameHandler;
 
     //Add warnings for missing compat PBOs (only if AB is on)
     {
--- a/addons/advanced_ballistics/XEH_postInit.sqf
+++ b/addons/advanced_ballistics/XEH_postInit.sqf
@@ -22,6 +22,9 @@ if (!hasInterface) exitWith {};
     ["ace_firedPlayer", DFUNC(handleFired)] call CBA_fnc_addEventHandler;
     ["ace_firedPlayerNonLocal", DFUNC(handleFired)] call CBA_fnc_addEventHandler;
 
+    // Register Perframe Handler
+    [{call FUNC(handleFirePFH)}, GVAR(simulationInterval)] call CBA_fnc_addPerFrameHandler;
+
     //Add warnings for missing compat PBOs (only if AB is on)
     {
         _x params ["_modPBO", "_compatPBO"];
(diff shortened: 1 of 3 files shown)


dedmen:
Did you test that a copy plus engine iteration for deletion over every single element is actually faster than engine interation until first element, and deletion?

addons/advanced_ballistics/functions/fnc_handleFirePFH.sqf
```diff
@@ -29,9 +28,15 @@
...
     };
-    nil
-} count +GVAR(allBullets);
+} forEach GVAR(allBullets);
+
+if (_deleted) then {
+    GVAR(allBullets) = GVAR(allBullets) - [objNull];
+};
 
 if (GVAR(allBullets) isEqualTo []) then {
```
PabstMirror, line 41: I think you could move 41-44 inside the `_deleted` if statement
(only way it could be `[]` is if there was a deletion)

jokoho48: I changed it for now. but the question is if this is really required for "performance" or if we should add the PFH only once and never remove it because from my POV this PFH is only doing work while a bullet is in the array

addons/advanced_ballistics/functions/fnc_handleFirePFH.sqf
```diff
@@ -29,11 +28,16 @@
...
     };
-    nil
-} count +GVAR(allBullets);
+} forEach GVAR(allBullets);
 
-if (GVAR(allBullets) isEqualTo []) then {
-    [_this select 1] call CBA_fnc_removePerFrameHandler;
-    GVAR(BulletPFH) = nil;
+if (_deleted) then {
+    GVAR(allBullets) = GVAR(allBullets) - [objNull];
```
SilentSpike, line 38: Is this quick than `GVAR(allBullets) arrayIntersect GVAR(allBullets)` to remove `nil` (instead of `objNull`) elements?

Alternatively, shouldn't the bullets automatically become `objNull` once they've impacted? Why are we checking `alive` to begin with?

jokoho48: the GVAR(allBullets) does not contain bullets but a data array.

SilentSpike: Ah yeah I see now it's unpacked in the loop and we're just using `objNull` for the removal 👍

SilentSpike: In that case, what about doing a `select {}` and just running all the code in the select block and ending it with `alive _bullet`?

jokoho48: slower. this is currently the fastest method. we had that discussion what the fastest way of that type of filtering an array I think 2x already and we decided and performance checked that this was back than the fastest method. and I just updated those now because I think we missed the 2 back in the day. the same code also runs in the CBA for some PFH systems

addons/advanced_ballistics/functions/fnc_handleFirePFH.sqf
PabstMirror: I think this would be good, windDeflection pfeh works that way

addons/advanced_ballistics/XEH_postInit.sqf
```diff
@@ -22,6 +22,9 @@ if (!hasInterface) exitWith {};
     ["ace_firedPlayer", DFUNC(handleFired)] call CBA_fnc_addEventHandler;
     ["ace_firedPlayerNonLocal", DFUNC(handleFired)] call CBA_fnc_addEventHandler;
 
+    // Register Perframe Handler
+    [{call FUNC(handleFirePFH)}, GVAR(simulationInterval)] call CBA_fnc_addPerFrameHandler;
```
PabstMirror, line 26:
```suggestion
    [FUNC(handleFirePFH), GVAR(simulationInterval)] call CBA_fnc_addPerFrameHandler;
```

PR #7709, opened 2020-05-13
Tagging - Added ability to specify custom model

dedmen:

Give modders the ability to specify the Object that will be created when you create a tag.
Main intention is that I want 10x10m textures, but could either add a boolean for 1x1m/10x10m or just let the modder configure literally every possible variant (There is also 1x2m and maybe others in some mods)

It accepts a p3d file path, or a CfgVehicles classname.


PabstMirror:
merged as part of #7710

PR #7710, opened 2020-05-13
Tagging - Add support for mission Tags

dedmen:

Parse tags from description.ext
It will also handle mission/mod paths just like https://community.bistudio.com/wiki/Description.ext#CfgSounds

This is based on #7709 branch (sorry but they would conflict either way)

### DIFF
--- a/addons/tagging/functions/fnc_compileConfigTags.sqf
+++ b/addons/tagging/functions/fnc_compileConfigTags.sqf
@@ -17,7 +17,7 @@
 
 {
     private _class = configName _x;
-    private _result = [_x, false] call ace_tagging_fnc_parseConfigTag;
+    private _result = [_x, false] call FUNC(parseConfigTag);
 
     if !(_result isEqualTo []) then {
         _result params ["_tagInfo", "_requiredItem"];
@@ -29,7 +29,7 @@
 
 {
     private _class = configName _x;
-    private _result = [_x, true] call ace_tagging_fnc_parseConfigTag;
+    private _result = [_x, true] call FUNC(parseConfigTag);
 
     if !(_result isEqualTo []) then {
         _result params ["_tagInfo", "_requiredItem"];
--- a/addons/tagging/functions/fnc_compileConfigTags.sqf
+++ b/addons/tagging/functions/fnc_compileConfigTags.sqf
@@ -48,8 +48,13 @@
 
     private _icon = getText (_x >> "icon");
 
+    private _tagModel = getText (_x >> "tagModel");
+    if (_tagModel == "") then {
+        _tagModel = "UserTexture1m_F";
+    };
+
     if (!_failure) then {
-        GVAR(cachedTags) pushBack [_class, _displayName, _requiredItem, _textures, _icon, _materials];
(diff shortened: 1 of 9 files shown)


addons/tagging/functions/fnc_compileConfigTags.sqf
```diff
@@ -16,40 +16,25 @@
  */
 
 {
-    private _failure = false;
     private _class = configName _x;
+    private _result = [_x, false] call ace_tagging_fnc_parseConfigTag;
```
PabstMirror, line 20:
```suggestion
    private _result = [_x, false] call FUNC(parseConfigTag);
```

addons/tagging/functions/fnc_compileConfigTags.sqf
```diff
@@ -16,40 +16,25 @@
...
-        _failure = true;
+        GVAR(cachedTags) pushBack _tagInfo;
+        GVAR(cachedRequiredItems) pushBackUnique _requiredItem;
     };
+} forEach ("true" configClasses (configFile >> "ACE_Tags"));
 
-    private _materials = getArray (_x >> "materials");
+{
+    private _class = configName _x;
+    private _result = [_x, true] call ace_tagging_fnc_parseConfigTag;
```
PabstMirror, line 32:
```suggestion
    private _result = [_x, true] call FUNC(parseConfigTag);
```

docs/wiki/framework/tagging-framework.md
```diff
@@ -59,3 +63,24 @@ class ACE_Tags {
 2  | `"ACE_SpraypaintRed"` | Required item to have in the inventory to be able to spray your tag
 3  | `["tagTexture1.paa", "tagTexture2.paa"]` | List of texture variants (one is randomly selected when tagging)
 4  | `"icon.paa"` | Icon being displayed in the interaction menu
+
+
+### 2.2 Tags in description.ext
+
+Tags can also be configured in description.ext like shown above.
+File Paths will be relative to your mission, if you want to define Tags inside description.ext but use Addon paths for `Texture`/`Material`/`TagModel` you need to prefix the path with `@`.
```
PabstMirror, line 71: ref, this is a semi-standard used in vanilla
>Since Arma 3 v1.49.131710 it is possible to define AddOn sounds in mission config. In order to make engine look for the sound in AddOn, the sound path must start with @ (instead of \) for example:

addons/tagging/functions/fnc_tag.sqf
```diff
@@ -20,7 +21,8 @@
 params [
     ["_unit", objNull, [objNull]],
     ["_texture", "", [""]],
-    ["_material", "", [""]]
+    ["_material", "", [""]],
+    ["_tagModel", "UserTexture1m_F", [""]]
```
mharis001, line 25: This is not used anywhere in this function. Should be added to L117 so it is passed to `createTag`.

dedmen: Its not used in the params, but should still be in the header, and adding it to params will make sure that it prints error at correct place if you pass wrong datatype

docs/wiki/framework/tagging-framework.md
dedmen: Should that be written down in docs somehow? Don't know how

PR #7721, opened 2020-05-22
Fix inconsistent mm usage

synixebrett:
- No more "82 mm" or similar, all replaced with "82mm"
- I ran a `([.\d]+)\smm` -> `$1mm`  find and replace

Notes:
This seems to be the more widely used option, ACE is already mostly using the "no space" version, CUP is entirely without spaces


Elgin675:
Personally I prefer to have a space, at least that's how the typographical rules are for my country (France).

Why do you want to impose it absolutely for each language ? The important thing for me is that each language should be "homogeneous".

synixebrett:
Seems like it was ACE that was wrong, BI uses `6.5 mm` consistently. I'll make a new PR to have ACE match this

PR #7730, opened 2020-05-23
Reload - Fix display ammo count in vehicle

Dystopian:
- title;
- ~~fix no `Check ammo` interaction on statics when CSW ammo handling is disabled;~~
- fix and improve `Take` EH;
- optimize fnc_displayAmmo.

When drag magazine to your weapon in gear in vehicle it will display empty ammo count. To fix it `weaponState` should be used instead of `currentWeapon`, `currentMuzzle` and `currentMagazine`.

### DIFF
--- a/addons/csw/functions/fnc_staticWeaponInit.sqf
+++ b/addons/csw/functions/fnc_staticWeaponInit.sqf
@@ -63,13 +63,12 @@ if (hasInterface && {!(_typeOf in GVAR(initializedStaticTypes))}) then {
         [_typeOf, 0, ["ACE_MainActions"], _disassembleAction] call EFUNC(interact_menu,addActionToClass);
     };
 
-    if (GVAR(ammoHandling) == 0) exitWith {};
 
     private _ammoActionPath = [];
     private _magazineLocation = getText (configFile >> "CfgVehicles" >> _typeOf >> QUOTE(ADDON) >> "magazineLocation");
     private _condition = { //IGNORE_PRIVATE_WARNING ["_target", "_player"];
-        // If weapon assembly/disassembly is enabled we enable ammo handling
-        if !([false, true, true, GVAR(defaultAssemblyMode)] select (_target getVariable [QGVAR(assemblyMode), 3])) exitWith { false };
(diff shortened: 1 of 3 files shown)


addons/csw/functions/fnc_staticWeaponInit.sqf
```diff
@@ -63,12 +63,13 @@ if (hasInterface && {!(_typeOf in GVAR(initializedStaticTypes))}) then {
         [_typeOf, 0, ["ACE_MainActions"], _disassembleAction] call EFUNC(interact_menu,addActionToClass);
     };
 
+    if (GVAR(ammoHandling) == 0) exitWith {};
```
PabstMirror, line 66: this is not the same

Dystopian: Yeah, my bad. Misread `||` instead of `&&`.

Max255PL:
I'm guessing these shouldn't be happening? Current ACE master + CBA 3.15.1 only.

```16:55:09 Error in expression <_unit call ace_reload_fnc_displayAmmo}; \;>
16:55:09   Error position: <\;>
16:55:09   Error Invalid number in expression
16:55:09 Error in expression <_unit call ace_reload_fnc_displayAmmo}; \;>
16:55:09   Error position: <\;>
16:55:09   Error Invalid number in expression
16:55:09 Error in expression <_unit call ace_reload_fnc_displayAmmo}; \;},{}],true,[]],["CAManBase","Take",[{c>
16:55:09   Error position: <\;},{}],true,[]],["CAManBase","Take",[{c>
16:55:09   Error Invalid number in expression
16:55:09 Error in expression <_unit call ace_reload_fnc_displayAmmo}; \;},{}],true,[]],["CAManBase","Take",[{c>
16:55:09   Error position: <\;},{}],true,[]],["CAManBase","Take",[{c>
16:55:09   Error Invalid number in expression```

Dystopian:
looks like that HEMTT bug with `\;` at the end. @Max255PL Did you use HEMTT?

Max255PL:
Ah, indeed I did. I knew I saw something like this before but couldn't remember what was it about. Disregard then.

PR #7731, opened 2020-05-23
Medical - Prevent packing into bodybag for bodies in vehicle

veteran29:
- title

Bodybag spawned inside vehicle goes booom.


PabstMirror:
👍 for adding comment explaining why it has to be like that

Issue #7637, opened 2020-04-14
[Explosives] Self-Interaction Menu Breaking


GCodeman:
**Mods**
- **Arma 3:** `1.96.146114`
- **CBA:** `3.15.0`
- **ACE3:** `3.13.1`

**Description:**
After hovering your cursor over the Explosives portion of the ACE Self Interaction Menu then dying, the Explosives Tab will be permanently stuck with whatever options you had upon death. This persists regardless of dying again or dropping/gaining any explosives. Also, upon acquiring new explosives, the option to place them does not get added to your menu.

**Steps to reproduce:**
-Take any explosive
-Open self interaction menu and hover cursor over explosive, but do not place it
-Die in any fashion or press respawn from the menu

**Expected behavior:**
-Upon respawning, your explosives menu should reset but does not.

**Where did the issue occur?**
- Self-Hosted Multiplayer (with several other players)
- Editor (Hosted in multiplayer editor, but playing only by myself)

**Video**
https://www.youtube.com/watch?v=SfTBHNMH19g

GCodeman:
I just noticed that it also occurs when opening the self-interaction menu at all when an explosive is in your inventory, not just when hovering over an explosive.

bux:
RPT is missing

GCodeman:
ArmA updated to 1.98.146303 today, so I made a new RPT file.
https://pastebin.com/G7cEYt34

Whigital:
Can confirm this, i think it has to do with how `ACE_player` reference changes when killed/respawned. Depending on when the `cba_event_loadoutEvent` is triggered, `ace_common_fnc_eraseCache` might clear the variables on the ragdoll, since the reference changed.

<details>
  <summary>init.sqf</summary>

```sqf
Report_Data = {
    params [
        "_event"
    ];

    _log = {
        params [
            "_cap",
            "_msg"
        ];

        diag_log (format ["%1: %2", _cap, _msg]);
    };

    private _cache = missionNamespace getVariable ["ace_common_clearCache_cba_events_loadoutEvent", nil];
    private _actions = ACE_player getVariable ["ace_explosives_explosiveActions", []];
    
    diag_log (format ["########## %1 ##########", _event]);
    diag_log (format ["player == %1", player]);
    diag_log (format ["ACE_player == %1", ACE_player]);
    ["ace_common_clearCache_cba_events_loadoutEvent", _cache] call _log;
    ["ace_explosives_explosiveActions", _actions] call _log;
};

[...]

PR #7739, opened 2020-05-28
Explosives - Fix place explosives cache not being cleared on respawn (#7637)

Whigital:
- Fix `ace_explosives_explosiveActions` not being cleared at respawn.

Addresses issue #7637

It isn't pretty, but it works. Any improvement suggestions welcome ....

### DIFF
--- a/addons/explosives/functions/fnc_onRespawn.sqf
+++ b/addons/explosives/functions/fnc_onRespawn.sqf
@@ -1,0 +1,30 @@
+#include "script_component.hpp"
+/*
+ * Author: Whigital
+ * Clears cached variable containing placeable explosives.
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ * 1: Corspe <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ *
+ * Public: No
+ */
+
+params [
+    "_unit",
+    "_corpse"
+];
+
+TRACE_2("params",_unit,_corpse);
+
+if (_unit == player) then {
+    [
+        {ACE_player == player},
+        {[ACE_player, QGVAR(explosiveActions)] call EFUNC(common,eraseCache);}
+    ] call CBA_fnc_waitUntilAndExecute;
+};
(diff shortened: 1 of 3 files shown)


addons/explosives/functions/fnc_onRespawn.sqf
```diff
@@ -0,0 +1,30 @@
...
+    "_corpse"
+];
+
+TRACE_2("params",_unit,_corpse);
+
+if (_unit == player) then {
+    [
+        {ACE_player == player},
+        {[ACE_player, QGVAR(explosiveActions)] call EFUNC(common,eraseCache);}
+    ] call CBA_fnc_waitUntilAndExecute;
```
commy2, line 29: I can not make sense of this WUAE. Please explain.

Whigital: Placable explosives actions are generated by `ace_explosives_fnc_fnc_addExplosiveActions`, which wraps the code in a cached call to `ace_common_fnc_cachedCall`.
https://github.com/acemod/ACE3/blob/19523f028a2d1316e8a15a618d643b611b543cf7/addons/explosives/functions/fnc_addExplosiveActions.sqf#L42

If the inventory changes, var (`ace_explosives_explosiveActions`) that stores the cached actions gets cleared. All good so far, works as intended.

Somewhere in the process of getting killed and respawning though, the reference to `ACE_player` changes, and the `cba_events_loadoutEvent` that is triggered after respawn does not clear the `ace_explosives_explosiveActions` var.

Don't know if it has to do with the `ACE_player` var referencing the old corpse still, or if the var gets restored. Only way i was able to *fix* it was to manually delay the call that is made when a `cba_events_loadoutEvent` is triggered.
https://github.com/acemod/ACE3/blob/19523f028a2d1316e8a15a618d643b611b543cf7/addons/common/functions/fnc_cachedCall.sqf#L49

PabstMirror:
Great work tracking down the bug,
I am a little worried about the waitUntilAndExecute loop (e.g. AI respawns?)
I think this could also be solved with this in the bottom of postInit
```
["unit", {
    params ["_player"];
    [_player, QGVAR(explosiveActions)] call EFUNC(common,eraseCache);
```

Issue #2261, opened 2015-08-25
Improvement suggestions for medical treatment realism


sgtBrowncoat:
I like the system however I have a few tweaks in mind that relate to how we actually perform TCCC. 

There is no airway/breathing management. Introduction of respiratory distress and an NPA would be neat. If there is penetrating chest trauma and the NPA doesnt fix the respiratory distress it can cue the medic to perform needle thoracentesis. 

Adding this allows for a medic more closely follow the MARCH algorithm.

Changing the use of whole blood to Hextend (or any brand hetastarch). Carrying whole blood in the field would be a nightmare. It is required to be kept cold and then you need to make sure to cross match with the person receiving. Hextend is a plasma volumizer that is given in 500ml increments to increase a low blood pressure caused by hypovolemia. 

 Use of a sphygmomanometer is not a common practice in the field. Those should be limited to medical vehicles and aid stations. We determine need for volume expansion based on systolic over palpation. Locationalized palpable pulses tell the operator approximately what the systolic BP is at. 
~80 systolic at the radial (arm)
~70 systolic at femoral (leg)
~60 systolic at carotid (head)
also it would be fun if you could make a pulse appear absent on a limb with a TQ on it.

gienkov:
>  make a pulse appear absent on a limb with a TQ on it

I like this one.

sgtBrowncoat:
you wouldnt believe how many times while I was in SOCM that I got caught up in the moment and would check my patient's pulse and get no feedback and start freaking out that I had missed some major injury while in reality I was doing great. Just not paying attention.

gienkov:
How does blood pressure behaves on limb that have TQ on? Is the readback lowered or raised?

sgtBrowncoat:
If the TQ is applied correctly you would not get a BP. All blood flow to that limb is cut off and there is no blood going to it and the blood that was in it is equalized. You could possibly get a transient diastolic BP however you would not get a systolic pressure.


If applied incorrectly (not enough pressure, over a joint, ect) it can allow blood into the limb because arteries reside deeper and not allow blood out through the veinous system- this creates what is called a compartment syndrome; but I think that's a little more than you want/need. 


You should also not give medications in a limb with a TQ on it because it would not go anywhere and actually have a toxic effect because the concentration is too high. Trying to give IV fluids would result in the bag not flowing. 


[...]

PR #7740, opened 2020-05-29
WIP: Medical Treatment - Realistic pulse checking

PabstMirror:
https://www.youtube.com/watch?v=kGFEYOfDQZo

Pulsing to show target's pulse (**possible seizure warning?**)
There is no way you are going to check someone's pulse in 2.5 seconds
With this you get their actual pulse and a time source
- can quickly tell if pulse is very fast/slow
- can count for 6 sec and multiply pulse by 10 or whatever accuracy/speed you want

Also disables blood pressure checking for non-medics as this requires specialized gear that a basic rifleman is not going to be carrying.

- [ ] settings for all this?


BaerMitUmlaut:
> possible seizure warning

I have something in the pipeline that also might not be epileptic friendly. Should we introduce a setting in common for epileptic friendly UX?

I think the flashing is a bit too much, love the idea though. What if the user has no watch in his inventory? What about custom watches?

thojkooi:
The flashing seems a bit much.

bux:
Maybe instead of a dark blurred sphere use something like a small white dot increasing in size/pulsing

BaerMitUmlaut:
Why not a heart icon pulsing?

synixebrett:
What about a sound effect instead of / in addition to  pulsing?

smitt14ua:
Useful. I think it's should be better if you'll change size little bit smaller and add some icons instead of blinking.
W/o watch it's not possible to check pulse?

pterolatypus:
I love the idea of adding mini-games like this - if medical were less about navigating menus and watching progress bars it'd be a lot more engaging.
I second the suggestion of a heart icon that changes in size (and maybe colour?) as well as sound effect.
Maybe out of scope but modulating the effect based on bp (higher pulse pressure -> bigger heartbeats) and making the pulse absent when the systolic bp is too low (ref #2261) would be nice too/

jonpas:
I want this...

johnb432:
> I want this...

As far as I can tell, two things remain:
- Handling epilepsy, in which case we should just skip it
- A setting to enable/disable.

I'll commit what I think would be suitable for both cases.

addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
```diff
@@ -231,23 +231,33 @@ class GVAR(actions) {
...
+        callbackProgress = QFUNC(checkPulseProgress);
+        callbackFailure = QUOTE(QQGVAR(checkPulse) cutText [ARR_2('','PLAIN')]);
         animationMedicProne = "";
         animationMedicSelfProne = "";
     };
     class CheckBloodPressure: CheckPulse {
         displayName = CSTRING(Actions_CheckBloodPressure);
         displayNameProgress = CSTRING(Check_Bloodpressure_Content);
         allowedSelections[] = {"LeftArm", "RightArm", "LeftLeg", "RightLeg"};
+        treatmentTime = 15;
```
johnb432, line 246: 15 is too long imo. I think 6 would be more reasonable, given the estimates you can do after 6 seconds (I would also change the pulse checking time to 6 too when realistic pulse checking is disabled).

addons/medical_treatment/functions/fnc_checkPulseProgress.sqf
```diff
@@ -0,0 +1,53 @@
...
+} else {
+    private _pulse = GET_HEART_RATE(_patient);
+    _lastPulse = _display getVariable [QGVAR(lastPulse), -1];
+
+    private _nextPulse = _lastPulse + (if (_pulse == 0) then {1e99} else { 60 / _pulse });
+    if (CBA_missionTime > _nextPulse) then { _display setVariable [QGVAR(lastPulse), _nextPulse]; };
+    private _fade = (linearConversion [-0.1, 0, CBA_missionTime - _nextPulse, 0, 0.85, true] max linearConversion [0.35, 0, CBA_missionTime - _lastPulse, 0, 0.75, true]) ^ 2;
+
+    if (EGVAR(common,epilepsyFriendlyMode)) then {
+        _fade = _fade min 0.2;
```
johnb432, line 39: I'm unsure if this is enough to not trigger seizures.

Regardless, it would be better to implement an icon.

addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
```diff
@@ -231,23 +231,33 @@ class GVAR(actions) {
         displayName = CSTRING(Actions_CheckPulse);
         displayNameProgress = CSTRING(Check_Pulse_Content);
         allowedSelections[] = {"All"};
-        condition = QUOTE(GVAR(advancedDiagnose) != 0);
+        treatmentTime = QUOTE([ARR_2(2.5,15)] select GVAR(enableRealisticPulseChecking));
+        condition = QUOTE(GVAR(advancedDiagnose) != 0 && {!GVAR(enableRealisticPulseChecking) || {(_medic getSlotItemName TYPE_WATCH) isKindOf [ARR_2('ItemWatch',configFile >> 'CfgWeapons')]}});
```
johnb432, line 235: Do we want to require a watch for basic pulse checking or just have the watch requirement for realistic pulse checking?

PR #7744, opened 2020-05-31
Medical Feedback - Fix overriding hearing setting when unconscious

severgun:
- Title

Currently if `ace_hearing_unconsciousnessVolume` > 0.25 it will be ignored in favor of
https://github.com/acemod/ACE3/blob/b30f023c042d4182091218f39db92b98f34e24ba/addons/medical_feedback/script_component.hpp#L40


PabstMirror:
Looks good, I think we could even delete
```
// Reduce volume if player is unconscious
if (ACE_player getVariable ["ACE_isUnconscious", false]) then {
    _volume = _volume min GVAR(UnconsciousnessVolume);
};
```
https://github.com/acemod/ACE3/blob/master/addons/hearing/functions/fnc_updateVolume.sqf#L59-L62

from hearing, as the effect will be taken care of by medical_feedback now

PabstMirror:
We could also move the setting to medical_feedback

PR #7746, opened 2020-06-01
Dragging - Add Keybind for Carry

LachlanSneddon:
- Add a Carry/Release Object key bind, using the same implementation as the Drag/Release Object key bind

### DIFF
--- a/addons/dragging/XEH_postInit.sqf
+++ b/addons/dragging/XEH_postInit.sqf
@@ -53,8 +53,7 @@ if (isNil "ACE_maxWeightCarry") then {
 
     [ACE_player, _cursor] call FUNC(startDrag);
     false
-}, {
-}, [-1, [false, false, false]]] call CBA_fnc_addKeybind; // UNBOUND
+}, {}, [-1, [false, false, false]]] call CBA_fnc_addKeybind; // UNBOUND
 
 ["ACE3 Common", QGVAR(carry), (localize LSTRING(CarryKeybind)), {
     if (!alive ACE_player) exitWith {false};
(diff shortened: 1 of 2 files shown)


addons/dragging/XEH_postInit.sqf
```diff
@@ -56,3 +56,27 @@ if (isNil "ACE_maxWeightCarry") then {
...
+    };
+
+    private _cursor = cursorObject;
+    if ((isNull _cursor) || {(_cursor distance ACE_player) > 2.6}) exitWith {false};
+    if (!([ACE_player, _cursor] call FUNC(canCarry))) exitWith {false};
+
+    [ACE_player, _cursor] call FUNC(startCarry);
+    false
+}, {
+    false
```
mharis001, line 81: Change to an empty code block if it just returns `false`; that way an key up handler isn't even added.

LachlanSneddon: updated

addons/dragging/XEH_postInit.sqf
```diff
@@ -54,5 +54,27 @@ if (isNil "ACE_maxWeightCarry") then {
...
+        false
+    };
+
+    private _cursor = cursorObject;
+    if ((isNull _cursor) || {(_cursor distance ACE_player) > 2.6}) exitWith {false};
+    if (!([ACE_player, _cursor] call FUNC(canCarry))) exitWith {false};
+
+    [ACE_player, _cursor] call FUNC(startCarry);
     false
+}, {
```
mharis001, line 79: No newline.

PR #7756, opened 2020-06-13
Fastroping - Fix FRIES removal

mharis001:
- Fix FRIES removal when helicopter is killed or deleted
- `equipFRIES` function could be called on any machine and as a result
    - the killed event would not work as expected if the locality of the helicopter changed
    - the deleted event would not work as expected if added on a client that later disconnects

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -18,6 +18,7 @@ Janus
 jokoho482 <jokoho482@gmail.com>
 Jonpas <jonpas33@gmail.com>
 Kieran
+mharis001 <mhariszakar@gmail.com>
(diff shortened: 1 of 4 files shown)


AUTHORS.txt
```diff
@@ -114,7 +115,6 @@ Luigi "Luigium" Myrini <luigium@outlook.fr>
 Macusercom <macusercom@gmail.com>
 MarcBook
 meat <p.humberdroz@gmail.com>
-mharis001 <mhariszakar@gmail.com>
```
dedmen, line 117: Why? They were sorted alphabetically?

PR #7757, opened 2020-06-13
Common - Fix disableUserInput blocking already open Zeus display

mharis001:
- title, when the curator's unit went unconscious while in Zeus, would require pressing the Zeus key again to see/interact with editable object icons
- Fix some magic numbers


commy2:
urg more actionKeys

PR #7509, opened 2020-02-05
(Ready) Adds more laser functionality to aircraft

LorenLuke:
**When merged this pull request will:**
- [x] Create laser spot tracker functionality for pilot cam-equipped fixed wing aircraft
- ~~[ ] Enable the usage of a marker laser (https://youtu.be/hv2gBHVXZds?t=169) for indication by plane to point to a location. (MP Compatible)~~
- [x] Enable self laser tracking- When ground-tracking in the pilot camera, flicking the laser on/off slaves the ground-track to the lased point, enabling tracking of points on buildings or objects as you orbit, rather than the ground behind them (like here: https://youtu.be/IOZ18GNPI-c?t=75).
 - ~~[x] Adds marker laser to gunner stations (including helicopters and fixed wing turreted aircraft, i.e. Blackfish, 3rd party add-on second seat operators, e.g. F/A-18B/D).~~
    - [x] Find alternate handling for weird non-primary turret LDs on vehicles, e.g. for Blackfish.
 - ~~[ ] Adds marker laser option for handheld Laser designators (toggleable/keybind).~~

Known issues:
- ~~Currently available on Pilot camera fixed wing craft only. (Will fix)~~ Fixed.
- ~~Some amount of marker laser/laser seeker detachment from target pod orientation when targeting on water. (Cannot fix?)~~
- ~~Laser point tracking on certain buildings (e.g. Airstation Mike 26 Radome), where the geometry is transparent to the laser seeker does not generate a tracking point. (May attempt fix)~~
- Sometimes, moving a ground-track point with an active laser doesn't track the laser point on a building. Stopgap Fix: switch laser off and on to track point on building.
- ~~Position of laser doesn't correspond to targeting pod location while at speed. (Cannot fix?)~~ Fixed! Thanks @BaerMitUmlaut!

Intended Behavior, which may be mistaken for bugs:
- When tracking a laser point on a building, switching the laser off will still keep the same track point until the camera is moved.
- Laser Spot Tracker does not follow a laser point after being set to 'ground track'. This is intentional so as to have a quick way to move the laser tracker without it being constantly pulled towards a particular laser spot.
- ~~Marker laser doesn't shine when turned on, except when designating. This is primarily for code reasons. If there exists worry about laser designating the wrong target, switch your laser code.~~

Marking this as 'Review Ready'.

PR #7761, opened 2020-06-14
Add IR marking laser for aircraft

BaerMitUmlaut:
- Adds an IR marking laser for aircraft, controlled through the pilot camera
- Idea taken from #7509 but completely rewritten
- IR laser is a 10km long light emitting object that is resized to the appropriate length, so it doesn't clip through objects or terrain
- Laser orientation updates are sent over the network once per second and interpolated clientside

TODO:
- [ ] **Blocking issue**: setVectorUp with attached objects is not instant, thus the laser doesnt actually touch the ground and is quite shaky when the plane is moving - maybe @dedmen can fix that in engine?
- [ ] Eden checkbox or something similar to manually equip vehicles with it instead of automatically enabling it for every plane with a pilot camera
- [ ] Feedback for the pilot that the laser is active (currently appears invisible because the camera is inside of the laser object, an offset could work but might block the view too much)
- [ ] Epilepsy friendly setting to disable laser flickering
- [ ] Maybe add support for 5km+ line intersection

Screenshot:

Reference image (the only one I could find, would love to see more if anybody finds some):
/arc-anglerfish-arc2-prod-mco.s3.amazonaws.com/public/NE3S4FWKV5GMVE6HKIGMWX3U5M.jpg)
[Source](https://www.militarytimes.com/flashpoints/2018/03/09/how-much-longer-will-us-and-partner-forces-own-the-night-in-combat/)

Mainly publishing this as a POC for custom IR lasers, I think the concept is pretty neato.


addons/markinglaser/CfgVehicles.hpp
```diff
@@ -0,0 +1,71 @@
+class CBA_Extended_EventHandlers;
+
+class CfgVehicles {
+    class FloatingStructure_F;
+    class ACE_MarkingLaser: FloatingStructure_F {
+        class EventHandlers {
+            class CBA_Extended_EventHandlers: CBA_Extended_EventHandlers {};
```
commy2, line 7: Please do:
```cpp
 class CBA_Extended_EventHandlers_base;
            class CBA_Extended_EventHandlers: CBA_Extended_EventHandlers_base {};
```

dedmen:
> setVectorUp with attached objects is not instant

more details? I would expect delay by a few frames, and then a "instant" flop into position.

addons/markinglaser/functions/fnc_renderPFH.sqf
```diff
@@ -0,0 +1,63 @@
...
+
+    // Adjust length of laser so it doesn't go through objects
+    private _intersectPos = lineIntersectsSurfaces [
+        _originWorld,
+        _originWorld vectorAdd (_vector vectorMultiply LIS_LIMIT),
+        _laser
+    ] param [0, []] param [0, []];
+
+    if (_intersectPos isEqualTo []) then {
+        _laser animate ["adjustLength", 0, true];
```
dedmen, line 47: wiki says to use "animateSource" instead

veteran29: Both are Effect Global however the PFH seems to be supposed to be running local?

Neviothr:
https://forums.bohemia.net/forums/topic/140837-development-branch-changelog/?page=58&tab=comments#comment-3438862
> Added: irDotSize entry to weapon IR laser attachment config classes for the size of the IR dot on laser impact

Perhaps make the laser's dot bigger, to create more of an area marking, rather than a pinpoint laser?

BaerMitUmlaut:
Aircraft don't have weapon attachments...
We can make the laser as big as we want, since everything here is custom and not based on vanilla.

dedmen:
Here is my script with the hopefully 2.08 scripted lasers feature
https://gist.github.com/dedmen/61643bd7b7205c13b68bff4b214f5840#file-weaponlaser-sqf-L136

addons/markinglaser/stringtable.xml
```diff
@@ -0,0 +1,25 @@
...
+        </Key>
+        <Key ID="STR_ACE_Markinglaser_Attribute_Enabled_Tooltip">
+            <English>Equips the aircrafts main turret or pilot camera with an IR marking laser.</English>
+            <German>Rüstet das Hauptgeschütz oder die Pilotenkamera des Luftfahrzeugs mit einem IR Markierungslaser aus.</German>
+        </Key>
+        <Key ID="STR_ACE_Markinglaser_Setting_Enabled_DisplayName">
+            <English>Enable IR marking lasers for aircraft</English>
+            <German>IR Markierungslaser für Luftfahrzeuge aktivieren</German>
+        </Key>
+        <Key ID="STR_ACE_Markinglaser_Setting_Enabled_Tooltip">
```
PabstMirror, line 20:
```suggestion
        <Key ID="STR_ACE_Markinglaser_Setting_Enabled_Description">
```

addons/markinglaser/functions/fnc_onAircraftInit.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+if !(_aircraft getVariable [QGVAR(enabled), false]) exitWith {};
+
+private _config = configOf _aircraft;
+private _turretData = [_config] call FUNC(findTurret);
+private _hasPilotCamera = getNumber (_config >> "PilotCamera" >> "controllable") > 0;
+
+if ((_turretData isEqualTo []) && {!_hasPilotCamera}) exitWith {
+    _aircraft setVariable [QGVAR(enabled), false];
+    WARNING_1("Class %1 does not have a pilot camera nor a turret that could be equipped with an IR marking laser.");
```
PabstMirror, line 28:
```suggestion
    WARNING_1("Class %1 does not have a pilot camera nor a turret that could be equipped with an IR marking laser.",configName _config);
```

addons/markinglaser/functions/fnc_onAircraftInit.sqf
```diff
@@ -0,0 +1,73 @@
...
+ * Equips an aircraft with a marking laser.
+ *
+ * Arguments:
+ * 0: Aircraft <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [plane] call ace_markinglaser_fnc_equipLaser
```
PabstMirror, line 13:
```suggestion
 * [plane] call ace_markinglaser_fnc_onAircraftInit
```

addons/markinglaser/functions/fnc_onLaserOff.sqf
```diff
@@ -0,0 +1,33 @@
...
+ * Handles a plane turning its marking laser on.
+ *
+ * Arguments:
+ * 0: Aircraft <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [plane] call ace_markinglaser_fnc_onLaserOn
```
PabstMirror, line 13:
```suggestion
 * [plane] call ace_markinglaser_fnc_onLaserOff
```

addons/markinglaser/functions/fnc_onLaserOn.sqf
```diff
@@ -0,0 +1,56 @@
...
+ * Handles a plane turning its marking laser off.
+ *
+ * Arguments:
+ * 0: Aircraft <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [plane] call ace_markinglaser_fnc_onLaserOff
```
PabstMirror, line 13:
```suggestion
 * [plane] call ace_markinglaser_fnc_onLaserOn
```

PabstMirror:
The 3den attribute is weird, it's default value is true.
So if you put a plane down, nothing will be set
if you open attributes it will show being checked, but not set anything

I had to set it to false, which added the attribute to the sqm
then change it back to true to get it to work

PabstMirror:
UAV's hate having this thing attached, they kept going up in altitude to avoid it

I forget what I did but somehow I ended up with this "burning" spot left, but it eventaully went away after a few seconds

veteran29:
> UAV's hate having this thing attached, they kept going up in altitude to avoid it

Maybe we should wait for 2.08 and use drawLaser?

BaerMitUmlaut:
>UAV's hate having this thing attached, they kept going up in altitude to avoid it

How are you testing this? I tried it with the Yabhon drone and a loiter waypoint at 500 m altitude, which seems to work just fine. That was pretty much the use case I had in mind originally. We could switch to the scripted laser later, if that has the same capability (range would be my main concern), but I would prefer not delaying this PR further unless necessary. 2.06 is already taking quite a while, so I don't expect 2.08 to still arrive this year.

I've fixed the attribute by enabling the laser by default. The bright spot is odd, since that should be part of the model, which is clearly not visible. I'd assume this is caused by some sort of lazy update within the engine. If you can repro it, please let me know.

dedmen:
> I don't expect 2.08 to still arrive this year.

Thats indeed not planned. rather Q1 2022

severgun:
Just a reminder that 2.08 already released.

BrettMayson:
- Switched to `drawLaser` instead of an object
- Shows the laser in Zeus night vision

Because `drawLaser` fades over time, it is a little obvious where the command repeats to extend the rage beyond what the engine supports for a single `drawLaser` call, but I personally think this is pretty minor.
image

addons/markinglaser/README.md
```diff
@@ -0,0 +1,11 @@
...
+===================
+
+Adds a marking laser for aircraft.
+
+
+## Maintainers
+
+The people responsible for merging changes to this component or answering potential questions.
+
+- [BaerMitUmlaut](https://github.com/BaerMitUmlaut)
```
LinkIsGrim, line 11:
```suggestion
```

LinkIsGrim (approved):

johnb432 (changes requested):
Got loads of
```
21:49:32 Error in expression <EqualTo [] } do {
        _intersects = lineIntersectsSurfaces [_startPos, _endP>
21:49:32   Error position: <lineIntersectsSurfaces [_startPos, _endP>
21:49:32   Error 0 elements provided, 3 expected
21:49:32 File /z/ace/addons/markinglaser/functions/fnc_renderPFH.sqf..., line 357
21:49:32 Error in expression <ivate _elevation = _modelVector#2 atan2 vectorMagnitude [_modelVector # 0, _mode>
21:49:32   Error position: <vectorMagnitude [_modelVector # 0, _mode>
21:49:32   Error Type Any, expected Number
21:49:32 File /z/ace/addons/markinglaser/functions/fnc_renderPFH.sqf..., line 339
```
being spammed after my plane crashed and I teleported myself out with Zeus (not sure if the teleportation had much of an effect tbh).

PabstMirror:
Last time I tested this in mp there were big problems with laser position syncing
I might have some ideas how to redo, I'll try to look into this because it is a neat feature

johnb432 (approved):

addons/markinglaser/README.md
```diff
@@ -0,0 +1,5 @@
+ace_markinglaser
+===================
+
+Adds a marking laser for aircraft.
```
rautamiekka, line 4:
```suggestion
Adds an IR marking laser for aircraft.
```

rautamiekka (changes requested):

PabstMirror:
I know forcing both lasers to be on at the same time isn't ideal
we can always try to revert https://github.com/acemod/ACE3/pull/10351 and get it working with the orginal logic
if we can solve the flicker//stairstep issue

PR #7783, opened 2020-06-30
Allow AI to Fastrope from VTOLs with FRIES

jwatt-arma:
**Description:**

I am part of the dev team for the [Project Future Vertical Lift](https://steamcommunity.com/sharedfiles/filedetails/?id=1686321576) mod. We are in process of releasing an ACE compatabiltity mod that adds ACE fastroping capability to the V-280 Valor VTOL aircraft that we made. This works great with human-controlled passengers. 
However, the waypoints that ace fastroping uses to allow AI to fastrope from vehicles requires that the vehicle be of type "Helicopter". Since VTOLs are of type "Plane", it does not allow AI to fastrope from the V-280. This change will fix that and prevent this from being a barrier to fully enabling fastroping on other VTOLs.

- Allow AI to fastrope from vehicles of type"Plane"

### DIFF
--- a/addons/fastroping/functions/fnc_deployAI.sqf
+++ b/addons/fastroping/functions/fnc_deployAI.sqf
@@ -18,7 +18,7 @@
  */
 params [["_vehicle", objNull, [objNull]], ["_deploySpecial", false, [true]], ["_createDeploymentGroup", true, [true]]];
 
-if (isNull _vehicle || {!(_vehicle isKindOf "Helicopter")}) exitWith {
+if (isNull _vehicle || {!((_vehicle isKindOf "Helicopter") || (_vehicle isKindOf "Plane"))}) exitWith {
     if (hasInterface) then {
         // Note: BIS_fnc_guiMessage causes a CTD with call, so spawn is used instead.
         ["deployAI was called with an invalid or non-existant vehicle.", QFUNC(deployAI)] spawn BIS_fnc_guiMessage;
--- a/addons/fastroping/functions/fnc_deployAI.sqf
+++ b/addons/fastroping/functions/fnc_deployAI.sqf
@@ -18,7 +18,7 @@
  */
 params [["_vehicle", objNull, [objNull]], ["_deploySpecial", false, [true]], ["_createDeploymentGroup", true, [true]]];
 
-if (isNull _vehicle || {!((_vehicle isKindOf "Helicopter") || (_vehicle isKindOf "Plane"))}) exitWith {
+if (isNull _vehicle || {!((_vehicle isKindOf "Helicopter") || {_vehicle isKindOf "Plane"})}) exitWith {
     if (hasInterface) then {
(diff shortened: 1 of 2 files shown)


addons/fastroping/functions/fnc_deployAI.sqf
```diff
@@ -18,7 +18,7 @@
  */
 params [["_vehicle", objNull, [objNull]], ["_deploySpecial", false, [true]], ["_createDeploymentGroup", true, [true]]];
 
-if (isNull _vehicle || {!(_vehicle isKindOf "Helicopter")}) exitWith {
+if (isNull _vehicle || {!((_vehicle isKindOf "Helicopter") || (_vehicle isKindOf "Plane"))}) exitWith {
```
PabstMirror, line 21:
```suggestion
if (isNull _vehicle || {!((_vehicle isKindOf "Helicopter") || {_vehicle isKindOf "Plane"})}) exitWith {
```

PabstMirror:
Will this be for AI only? it seems there are other places in ace_fastropping that are hardcoded for "Helicopter", like actions

tfg-jwatt:
> Will this be for AI only? it seems there are other places in ace_fastropping that are hardcoded for "Helicopter", like actions

Fastroping with human pilot/crew appeared to work fine in our tests and AI fastroping via the waypoint worked as expected with this change. Not sure if the other references you are referring to are vehicle-specific callback functions which wouldn't be used on VTOLs or if there is some other functionality I've missed.

jrdnmdhl:
> Will this be for AI only? it seems there are other places in ace_fastropping that are hardcoded for "Helicopter", like actions

I did a quick check and did not find any scripts with hardcoded reference to "Helicopter". Can you point me to what you are referring to?

PabstMirror:
Vehicle self-actions: https://github.com/acemod/ACE3/blob/master/addons/fastroping/CfgVehicles.hpp#L41-L91
Fries cleanup (not sure if this applies): https://github.com/acemod/ACE3/blob/master/addons/fastroping/XEH_preInit.sqf#L11-L15

jrdnmdhl:
> Vehicle self-actions: https://github.com/acemod/ACE3/blob/master/addons/fastroping/CfgVehicles.hpp#L41-L91
> Fries cleanup (not sure if this applies): https://github.com/acemod/ACE3/blob/master/addons/fastroping/XEH_preInit.sqf#L11-L15

Apologies, yes I see what you mean now. The cleanup one I should definitely make sure also applies to Plane. For the actions, I had envisioned that authors who want to add fastroping to their VTOLs would define those actions in their VTOLs, but I suppose it is more consistent with ACE to add it to the base plane class and let the conditions on the actions do the work of figuring out whether they are actually displayed to players.

Mike-MF:
PR is stale, mod itself seems to be abandoned at this stage.

Also, Fast roping from any kind of VTOL is a really bad idea.

Issue #7785, opened 2020-06-30
No interactions with TM-100/200/500


GalNa:
**Mods (complete and add to the following information):**
- **Arma 3:** `1.98`
- **CBA:** `3.15.0`
- **ACE3:** `3.13.3` 

**Description:**
Explosives TM-100/200/500 from mod RHS not working, it is look like any interactions was missed. 

**Steps to reproduce:**
- take 'rhssaf_tm200_mag'
- try place it

Whigital:
What are the thoughts about adding an RHS:SAF compat ?

I can handle the explosives, but that's about it, unless *someone* wants contribute.

Haven't looked at any of the SAF configs, but since it depends on USAF & AFRF, weapons, ammo, mag, vehivles etc might get compat values inherited already.

Blutze:
There are already at least some SAF weapons referenced in the GREF compat component: https://github.com/acemod/ACE3/blob/master/optionals/compat_rhs_gref3/CfgWeapons.hpp#L65-L79
Somehow, loading that without SAF doesn't give any errors on loading into the game or into a mission, but I would assume anything that iterates over all weapons checking for a config property would error out, so... shouldn't the component require SAF already?

- M84 is included in GREF compat, and its belt inherits from the PKM mags
- Skorpion does not inherit from vz 61, but uses those magazines and ammo; theres no compat stuff for that one anyway
- RBR-M80 (singleshot disposable launcher) does not inherit from M72, but its magazine and ammo inherit from the RPG26
- CZ99 and its magazines do not inherit from M9, but it uses ammo from USAF
- BRK antitank handgrenade classes do not inherit from the ones in AFRF, except for the penetrator
- All RHS flashbangs etc. inherit ACE frag from the vanilla handgrenade, making them more less-lethal than non-lethal
- Other SAF grenades don't inherit from anything in the RHS packs, but do inherit sensible stats from vanilla items
- MRUD claymores can be ACE-set to a pressure plate trigger that is located arount the body of the mine for the tripwire variants; no clue where it is for the "remote" variant
- PMA and TMA mines inherit apparently correct pressure plate behavior from somewhere; TMA correctly ignores personnel
- The tracer mag for the M76 is in SAF for some reason, GREF only has ball
- No pilot helmets, goggles or NVGs, no wirecutter backpacks, at least none that are `scope=2` and visible in the arsenal
- Medics inherit from some very basic base class, not AAF/CSAT medics, but apparently we don't need to touch them anyway
[...]

PR #7793, opened 2020-07-02
RHS Compat - Add RHS:SAF compat

Whigital:

- Add optional RHS:SAF compat

Addresses issue #7785

- [x] `Explosives`
- [x] `Ballistics`


Whigital:
Don't think there's any point in adding any `CSW` configs, since they're the same as `USAF` and `AFRF` assets with different names ....

PR #7796, opened 2020-07-03
Medical Engine - Prevent issues with respawning and ragdoll animations

PabstMirror:
If unit was killed while unconc it will keep `QGVAR(waitForAnim), true])` setVars


diwako:
While this will help with respawning, however won't this block applying animations after the first time, as in it will only show the default animations.

PR #7797, opened 2020-07-06
Overpressure - Add support for funnel shape backblast

kuloodporny:
- Add support for weapon attribute **ace_overpressure_funnel**, which will allow people standing directly behind the launcher to be unaffected, if this is how given launcher works IRL.
- ace_overpressure_funnel is added to Launcher base class and set to 0 (not used).
- Small proposed optimization change, can be reverted given there is evidence that lazy eval of separate code blocks is faster than nested code blocks.


kuloodporny:
It's not ready yet, I'm still working out the maths.

jonpas:
You don't have to reopen a pull request, simply keep committing.

kuloodporny:
Continued in #7798

PR #7798, opened 2020-07-06
Overpressure - Add support for funnel shaped backblast

kuloodporny:
- Add support for weapon attribute **ace_overpressure_funnel**, which will allow people standing directly behind the launcher to be unaffected, if this is how given launcher works IRL. Example: RPG-76 Komar
image
- ace_overpressure_funnel is added to Launcher base class and set to 0 (not used).
- Small proposed optimization change

### DIFF
--- a/addons/overpressure/CfgWeapons.hpp
+++ b/addons/overpressure/CfgWeapons.hpp
@@ -7,6 +7,7 @@ class CfgWeapons {
         GVAR(angle) = 60;
         GVAR(range) = 10;
         GVAR(damage) = 0.7;
+        GVAR(funnel) = 0; // Funnel shaped backblast does not affect units standing directly behind the launcher. Example: RPG-76 Komar
     };
 
     class Launcher_Base_F: Launcher {};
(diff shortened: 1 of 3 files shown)


kuloodporny:
Indeed the A3 lazy evals are smarter than I used to know.

PabstMirror:
I think we also want to modify ace_overpressure_fnc_firedEHBB to handle backblast reflected to the shooter

not sure of best way, 
could send multiple `FUNC(getDistance)` ray checks with `_direction` rotated by angle of the funnel, but that's messy
could also just disable reflection when funnel is enabled

kuloodporny:
Hmm, I overlooked that. Maybe in that case it would be better for me if I called the function manually in my mod, after I determine directions of exhausts - one of them is tagged by number 3 on picture below:
image

kuloodporny:
So I implemented my own script calling ACE native code. No need for actual funnel backblast then, this pull request can be closed unless somebody really needs the solution I made.

image

jonpas:
It does seem rather specific to the weapon type. So unless there is a need for a proper framework, this can rest. Thank you anyways!

kuloodporny:
Thank you guys, the review was really helpful!

PR #7804, opened 2020-07-08
Extensions - Add compileExtension script and update cMake file

PabstMirror:
python script to run 32/64 cmakes (should this be in /tools?)
gets ace version from sqf side
`_x64.dll` rename from ACRE
added `RVExtensionVersion`  

edit:
adds clang-format file
adds `EXPORT void __stdcall RVExtensionVersion(char *output, int outputSize) ` to all exts


jonpas:
> should this be in /tools?

Yes

.github/workflows/extensions.yml
```diff
@@ -22,7 +22,9 @@ jobs:
     - name: Build
       shell: cmd
       run: |
-        cd extensions/build
+        cd extensions
+        mkdir build
```
jonpas, line 26: Why? This folder is Git tracked.

PabstMirror: changed build folders to 
extensions/vcproj32
extensions/vcproj64

tools/compileExtensions.py
```diff
@@ -0,0 +1,101 @@
...
+# vim: set fileencoding=utf-8 :
+
+# make.py
+# An Arma 3 addon build system
+
+###############################################################################
+
+# The MIT License (MIT)
+
+# Copyright (c) 2013-2014 Ryan Schultz
```
jonpas, line 11: ? :D
I don't think this header is correct haha. I added Extension building to make.py anyways, so you can get rid of the entire header.

jonpas (commented):
CMakeLists have tabs in various places.

PR #7812, opened 2020-07-17
MissileGuidance - Add config option for using pilot camera as primary SACLOS control method

### LINKED ISSUE #7811: HOT Missile Guidance for Rotory  - missileguidance/functions/fnc_seekerType_SACLOS.sqf
After asking for help on the Slack chat, I have been told to come here and create an issue regarding a problem I faced.

When using the HOT missiles on a rotory craft, it seems that the HOT missiles are hardcoded to only work with the plane pilot camera, but not check for a rotory pilot camera. (Line 29 in the file)

Currently, rotory assets have to use their vehicle's nose to aim a wire guided missile. Allowing them to use their pilot camera would improve the quality of wireguided airstrikes.

**Solution you'd like:**
Instead of only checking if the shooter is a plane, maybe it could check if the shooter is a plane or a helicopter. I think this could help,

An alternative to using wireguided missiles would be to use laserguided missiles by self-lasering.

TheCandianVendingMachine:
fix #7811 
doesn't automatically apply to vehicles that want to use the camera since that would require information that we can't infer. All vanilla helicopters use other means for turrets so I am still assuming that to be the default.

### DIFF
--- a/addons/missileguidance/functions/fnc_SACLOS_onFired.sqf
+++ b/addons/missileguidance/functions/fnc_SACLOS_onFired.sqf
@@ -18,16 +18,18 @@ params ["_firedEH", "", "", "", "_stateParams"];
 _firedEH params ["_shooter","_weapon","","","","","_projectile"];
 _stateParams params ["", "_seekerStateParams"];
 
-private _config = ([_projectile] call CBA_fnc_getObjectConfig) >> "ace_missileguidance";
-
-if (_shooter isKindOf "Plane" && { !hasPilotCamera _shooter }) exitWith { WARNING("SACLOS fired from planes without camera unsupported"); };
+private _usePilotCamera = getNumber (([_shooter] call CBA_fnc_getObjectConfig) >> "pilotCamera" >> QGVAR(usePilotCameraForTargeting)) == 1;
 
 private _turretPath = [_shooter, _weapon] call CBA_fnc_turretPathWeapon;
 private _turretConfig = [_shooter, _turretPath] call CBA_fnc_getTurret;
 private _memoryPointGunnerOptics = getText(_turretConfig >> "memoryPointGunnerOptics");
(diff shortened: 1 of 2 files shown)


LorenLuke:
Rather than put in a value, would it be possible to see whether the magazine is in the turret/gunner mags for turret control, versus the main vehicle for pilot control (and/or if the pilot shot the weapon versus the gunner doing so)?

TheCandianVendingMachine:
> Rather than put in a value, would it be possible to see whether the magazine is in the turret/gunner mags for turret control, versus the main vehicle for pilot control (and/or if the pilot shot the weapon versus the gunner doing so)?

Not the issue presented. The issue is that we can't determine if a pilot camera is a sling-load helper or a targeting camera. If the pilot camera is a sling-load camera I don't want the missile to follow that and act dumb. If the pilot camera is a targeting pod (which isn't on any vanilla helicopters) it's up to the mod maker/compat maker to ensure that this flag is selected so it is used in calculations

We can't determine if a pilot-camera is a targeting one or not because in some instances you may want to have a fixed pilot camera that is the guidance sight. Its easier to assume that vanilla behaviour is the "correct" way for helicopters to manage weapon camera and add this to allow for compatibility with mod makers going off-script

MediHound:
With me not being the best regarding code, what line would I need to add to a vehicle to use the pilot camera for the missiles once this gets merged?

TheCandianVendingMachine:
> With me not being the best regarding code, what line would I need to add to a vehicle to use the pilot camera for the missiles once this gets merged?

For the vehicle there should be a config entry called `class pilotCamera`. In there you would put `ace_missileguidance_usePilotCameraForTargeting = 1;`

MediHound:
Thank you very much for the very swift reply!
I'll make sure to put that into the Pilot Camera once this gets merged! :)

PR #7815, opened 2020-07-22
Ballistics - Add a 5.8mm magazine with the CMR-76 (QBU-88)

Laid3acK:
- add a  5.8x42 mm 10 round magazine  for the `DMR_07_base_F` according with the [QBU-88 specifications](https://en.wikipedia.org/wiki/QBU-88).
- add a specific 5.8x42 mm "heavy bullet" according with the [DBP88 specifications](https://en.wikipedia.org/wiki/5.8%C3%9742mm) for this rifle.
- update the `airFriction` according with the CMR-76 / QBU-88 muzzle velocity and BC at the normal conditions (gap with AB 0.1 mRad (1 click) to 950 m, 0.2 to 1000 m, 0.5 to 1100 m, 0.7 to 1200 m, 1 mRad to 1300 m) : [Range Cards](https://steamuserimages-a.akamaihd.net/ugc/1465310083890288236/744A14EF04B1332DA0CB42372D44B32E0DBA1FA9/).
- update the AtragMx GunList.
- update the `CfgAmmoReference.hpp` and the `CfgWeaponsReference.hpp`.

> ##########################################
Ammo Class: ACE_580x42_DBP88_Ball
_drag model: 7_
_ballistic coefficient: 0.21_
MaxRanges (m): [950, 950]
MuzzleVelocities (m/s): [890, 900]
Max. Velocity diff (m/s): 33.23
Max. Drop diff (cm): 10.9
Max. Tof diff (ms): 40.0
Optimal airFriction: 0.00096786
##########################################

Terminal ballistic checked with this [article](https://www.gunsandammo.com/editorial/dbp87-5-8x42mm-chinas-high-velocity-caliber/248020) : vanilla `B_570x28_Ball` `hit` and `caliber` values a little bit overpowered (4.2 - 4.5 mm at 1000 m instead 3.5 mm).


Laid3acK:
I play with a english ArmA 3 but i fixed the vanilla french "Library" translation. This rifle was a airgun with BI.

Laid3acK:
I made a mistake with the vanilla translations ?

BaerMitUmlaut:
While it definitely works, you didn't use the ACE prefix for the stringtable entries:
`STR_ACE_Ballistics_12Gauge_Pellets_No00_Buck_NameShort` vs.
`str_a3_cfgweapons_srifle_dmr_07_wdm_f0`

Laid3acK:
That means i can't fix the caliber directly to the base game ?
I must use, for the translation, something like that : https://github.com/acemod/ACE3/blob/master/addons/realisticnames/CfgWeapons.hpp#L855
or like that : 
https://github.com/acemod/ACE3/blob/master/addons/ballistics/CfgMagazines.hpp#L408

BaerMitUmlaut:
Yes. Also, I don't like that you changed the round count. The class name doesn't fit now. I'd prefer if you'd add a new magazine.

Laid3acK:
> I don't like that you changed the round count. The class name doesn't fit now. I'd prefer if you'd add a new magazine.

I don't like that too and i did, to my first tests, a new mag `ACE_10Rnd_580x42_DBP88_Mag_F` inherit to the `20Rnd_650x39_Cased_Mag_F `, the only one `DMR_07_base_F` mag.
The problem in this case, 
if i configure the `DMR_07_base_F` with `magazines[] = {"ACE_10Rnd_580x42_DBP88_Mag_F"};`,
if an ace3 player selects and plays the vanilla CSAT (Pacific) Marksman, he will have no magazine in his inventory.
It's exactly what happened when i tried the first time the QBU-88 in-game.
The raison i chosen this solution.

BaerMitUmlaut:
That's because you removed the ability to take 20 round mags but didn't adjust the unit's loadout.

Laid3acK:
No, i just removed the ability to take the 20 round mags with the wrong caliber.

Arma 3 has some accurate 3D models, like the Lynx, the Cheytac or the QBU-88.
The goal of this PR is just fixing the vanilla exotic caliber and magazine count to this rifle and restoring the correct ones as irl.

This tiny config tweak with the `20Rnd_650x39_Cased_Mag_F` `count` solves an issue in-game.

Why we should keep the 20 round mags with the wrong caliber ?

Laid3acK:
Without an update with the vanilla translations for this rifle, that makes no sens.
Let's go with a new 5.8mm magazine for this splendid airgun. :)

The new mag is not appreciated, don't know why. Probably my fault ?

BaerMitUmlaut:
Why did you close this?

The problem is that this will cause existing loadouts and existing units, like the vanilla units, to have broken loadouts.

Laid3acK:
i had a conflict problem with this PR, probably because i used a old fork.
I made a new proper one https://github.com/acemod/ACE3/pull/7820 without problem this time.

PR #7819, opened 2020-07-26
Tagging - Add: Support of localized string for tagging interaction

Vdauphin:
- This pull request check if a string is localize and return the localization if available.
- If check of ìsLocalize`


PabstMirror:
Might be a tiny bit better to do this in ace_tagging_fnc_applyCustomTag so `GVAR(cachedTags)` just has localized values

addons/tagging/functions/fnc_applyCustomTag.sqf
```diff
@@ -27,6 +27,8 @@ params ["_identifier", "_displayName", "_requiredItem"];
 if !(GVAR(cachedTags) select {_x select 0 == _identifier} isEqualTo []) exitWith {
     INFO_2("Tag with selected identifier already exists: %1 (%2)",_identifier,_displayName)
 };
+_displayName = if (isLocalized _displayName) then {localize _displayName} else {_displayName};
+_this set [1, _displayName];
```
mharis001, line 31:
```suggestion

if (isLocalized _displayName) then {
    _this set [1, localize _displayName];
};

```

PR #7827, opened 2020-07-30
Maptool - Improve GPS 3D model by reducing vertex count A LOT

bux:
- reduce gps model vertex count from 3623 to 200 while still looking the same

image
_left old, right new_


BaerMitUmlaut:
wtf

PR #7830, opened 2020-08-01
Common - Fix trenches on snow terrain on Thirsk Winter map

Mysteryjuju:
- Allow players to dig trenches on snow on Thirsk Winter map


jonpas:
Snowman

PR #7838, opened 2020-08-03
MicroDAGR - Improve item model

bux:
- use the correct dimensions from the microDAGR technical document
- simplified model

_thicc_

_And, yes, the screen is turned off._


BaerMitUmlaut:
The brand logo is probably copyrighted, could you replace it with a mock? Maybe something silly like RockmyCojones?

bux:
We already use the logo since ever:

PabstMirror:
https://www.hollywoodreporter.com/thr-esq/call-duty-wins-first-amendment-victory-use-humvees-1287882

BaerMitUmlaut:
Ah right. Fine by me then. We can still change it if it becomes an issue.

PR #7851, opened 2020-08-11
Cargo - Fix loaded cargo not being deleted when vehicle is destroyed

### LINKED ISSUE #7844: Objects loaded inside cargo space are not deleted when a vehicle gets destroyed
Arma 3: 1.98 stable
CBA: 3.15.0 stable
ACE3: 3.13.3 stable

- Description:
When a vehicle gets destroyed, all the objects loaded inside its cargo should be deleted. This does not happen because [here](https://github.com/acemod/ACE3/blob/master/addons/cargo/CfgEventHandlers.hpp) the serverKilled XEH should be just a killed XEH since the killed EH is local.

- Steps to reproduce:
1. Place a vehicle in the editor, and an object to load into it. Assign a variable name to the object, then start the mission;
2. Load the object inside the vehicle;
3. Somehow destroy the vehicle (for example by using a rocket, or with the debug console);
4. Once that the vehicle has exploded, check if the object is alive (or use attachedObjects).

- Expected behavior:
The object should be deleted when the vehicle gets killed, as shown [here](https://github.com/acemod/ACE3/blob/master/addons/cargo/functions/fnc_handleDestroyed.sqf).

- Where did the issue occur?
Dedicated

Thank you!

mharis001:
- title, Close #7844 
    - Don't think `validateCargoSpace` needs to run on server
- Also delete loaded cargo when vehicle is deleted

### DIFF
--- a/addons/cargo/functions/fnc_handleDestroyed.sqf
+++ b/addons/cargo/functions/fnc_handleDestroyed.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: Glowbal
- * Handle object being destroyed. Only runs on server.
+ * Handle object being destroyed.
  *
  * Arguments:
  * 0: Object <OBJECT>
@@ -10,7 +10,7 @@
  * None
  *
  * Example:
- * [object] call ace_cargo_fnc_handleDestroyed
+ * [_object] call ace_cargo_fnc_handleDestroyed
  *
  * Public: No
  */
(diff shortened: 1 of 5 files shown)


addons/cargo/functions/fnc_handleDestroyed.sqf
```diff
@@ -24,9 +24,9 @@ if (_loaded isEqualTo []) exitWith {};
 {
     // TODO Do we want to be able to recover destroyed equipment?
     if (_x isEqualType objNull) then {
+        detach _x;
         deleteVehicle _x;
```
veteran29, line 28: Will `setDamage` trigger killed EHs? If so I wonder if we should kill the cargo so `killed` handlers will fire on it and delete it frame later?

mharis001: It should however, I think it's unnecessary for us to do it here.

thecage88:
Just curious, as I've searched all over and cant seem to find an answer. Is it possible to flag objects in the editor to be exempt from being deleted as cargo in the event the vehicle they are riding in is destroyed? 

basically, recreating this bug on purpose for specific objects?

PR #7871, opened 2020-08-24
Translations - Update / Fix French Translations (Grenades and Advanced Throwing)

Elgin675:
- Add french translations in `grenades` module,
- Fix french translations in `grenades` module.
- Fix french translations in `advanced throwing` module.

### DIFF
--- a/addons/grenades/stringtable.xml
+++ b/addons/grenades/stringtable.xml
@@ -530,11 +530,11 @@
         </Key>
         <Key ID="STR_ACE_Grenades_Convert_Fuse">
             <English>Convert to short fuse</English>
-            <French>Retardateur (7 secondes)</French>
+            <French>Appliquer une mèche</French>
         </Key>
         <Key ID="STR_ACE_Grenades_Remove_Fuse">
             <English>Remove short fuse</English>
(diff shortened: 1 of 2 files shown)


addons/grenades/stringtable.xml
```diff
@@ -530,11 +530,11 @@
         </Key>
         <Key ID="STR_ACE_Grenades_Convert_Fuse">
             <English>Convert to short fuse</English>
-            <French>Mèche lente (7 secondes)</French>
+            <French>Retardateur (7 secondes)</French>
```
PabstMirror, line 533: `(7 secondes)` - We should either remove from french or add to english so they are consistent

Elgin675: What do you think is the best solution ?
From my point of view, I thought that indicating the duration could be interesting, because in French, it is complicated to translate this text correctly, in the space available.

PabstMirror: @Cyruz143

Cyruz143: I was trying to keep shortish since it's coming up in the inventory UI and I'm not sure how it will scale on big UI scales with lower resolutions etc

PR #7876, opened 2020-08-25
Map Gestures - reduce the size of the procedural texture and don't generate mipmaps

jokoho48:
- Title


dedmen:
And why? generating 8x8 procedural texture instead of loading a 32x32 from disk is better on memory usage and doesn't need disk access to load.

jokoho48:
after my tests, it was marginally faster to use the static texture instead of procedural textures but if you say its worse than I trust you

dedmen:
How did you test and how much faster?

jokoho48:
when I use a 1,1,1 texture it is the fastest so I am gonna switch that to that.
I draw 100x100 "icons" and average the values of 1000 frames
8x8,3 - 7ms
texture - 6ms
1x1,1 - 5ms

PR #7877, opened 2020-08-26
Arsenal - Fix sorting when using search

synixebrett:
- Changing the active sort would break the top item's name when using the search bar


synixebrett:
My editor seems to have done whitespace cleanup in `tools/build.py`, I can remove if wanted

jonpas (approved):
It's fine.

PR #7878, opened 2020-08-26
Arsenal - Fix fire rate sort caused by flashlight

synixebrett:
- The 0 fire rate value of the Flashlight caused a divide by 0 error

### DIFF
--- a/addons/arsenal/functions/fnc_sortStatement_rateOfFire.sqf
+++ b/addons/arsenal/functions/fnc_sortStatement_rateOfFire.sqf
@@ -22,6 +22,7 @@ private _fireRate = [];
 } foreach _fireModes;
 
 _fireRate sort true;
+_fireRate = _fireRate param [0, 0];
 
 if (_fireRate == 0) exitWith {0};
-round (60 / (_fireRate param [0, 0]))
+round (60 / (_fireRate))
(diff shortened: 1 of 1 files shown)


jonpas (approved):

addons/arsenal/functions/fnc_sortStatement_rateOfFire.sqf
```diff
@@ -23,4 +23,5 @@ private _fireRate = [];
 
 _fireRate sort true;
 
+if (_fireRate == 0) exitWith {0};
```
PabstMirror, line 26: I think `_fireRate` is an array?

synixebrett: ah yeah 🤦

addons/arsenal/functions/fnc_sortStatement_rateOfFire.sqf
```diff
@@ -22,6 +22,7 @@ private _fireRate = [];
 } foreach _fireModes;
 
 _fireRate sort true;
+_fireRate = _fireRate param [0, 0];
 
 if (_fireRate == 0) exitWith {0};
-round (60 / (_fireRate param [0, 0]))
+round (60 / (_fireRate))
```
jonpas, line 28:
```suggestion
round (60 / _fireRate)
```

PR #7886, opened 2020-08-30
Quick Mount: Check vehicle simulation

Cyruz143:
**Stops players getting to vehicles with simulation disabled**
- You can currently mount in a vehicle witih sim disabled and you get "stuck" inside. Vanilla addActions allow this too (I've asked Ded to see if he can fix). For now this at leasts stops people accidentally getting in to something they shouldn't with a bad key press.
- Reusing the locked message as the player doesn't need to know the difference between locked/sim disabled, they just need to know they can't get in.
- Possibly warrants an RPT line so if you get people complaining they cannot get in a vehicle, RPT shows why (#blamethemissionmaker) - Thoughts?

### DIFF
--- a/addons/quickmount/functions/fnc_canShowFreeSeats.sqf
+++ b/addons/quickmount/functions/fnc_canShowFreeSeats.sqf
@@ -30,6 +30,7 @@ GVAR(enabled)
 }
 && {alive _vehicle}
 && {2 > locked _vehicle}
+&& {simulationEnabled _vehicle}
 && {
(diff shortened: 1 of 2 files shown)


jonpas (approved):
I like it!

jonpas:
> Possibly warrants an RPT line so if you get people complaining they cannot get in a vehicle, RPT shows why (#blamethemissionmaker) - Thoughts?

I don't think it's necessary, this is more an inconsistency fix than anything else.

PR #7887, opened 2020-08-31
Gunbag - Add BI Arsenal Compatibility

mjc4wilton:
- Prevent the weapon from being cleared from the gunbag after opening the BI arsenal.

### DIFF
--- a/addons/gunbag/functions/fnc_BIArsenalClose.sqf
+++ b/addons/gunbag/functions/fnc_BIArsenalClose.sqf
@@ -18,7 +18,7 @@
 params ["_unit"];
 
 if (!isNil QGVAR(arsenalCache)) then {
-    (backpackContainer _unit) setVariable [QGVAR(gunbagWeapon),GVAR(arsenalCache), true];
+    backpackContainer _unit setVariable [QGVAR(gunbagWeapon), GVAR(arsenalCache), true];
 };
 
 GVAR(arsenalCache) = nil;
--- a/addons/gunbag/functions/fnc_BIArsenalClose.sqf
+++ b/addons/gunbag/functions/fnc_BIArsenalClose.sqf
@@ -4,7 +4,7 @@
  * Handle BI arsenal closing
  *
  * Arguments:
- * None
+ * 0: Unit <OBJECT>
  *
  * Return Value:
  * None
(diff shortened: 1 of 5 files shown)


PabstMirror (approved):
Tested, seems to work great

addons/gunbag/functions/fnc_BIArsenalClose.sqf
```diff
@@ -0,0 +1,22 @@
...
+ * None
+ *
+ * Example:
+ * [] call ace_gunbag_fnc_BIArsenalClose
+ *
+ * Public: No
+ */
+
+if (!isNil QGVAR(arsenalCache)) then {
+    (backpackContainer player) setVariable [QGVAR(gunbagWeapon),GVAR(arsenalCache), true];
```
commy2, line 19:
```sqf
    backpackContainer player setVariable [QGVAR(gunbagWeapon), GVAR(arsenalCache), true];
```
Can you open the arsenal while remote controlling AI as Zeus?

In that case, this variable would be set on the wrong unit, as the `player` command is used, instead of a entity being passed around by the functions.

addons/gunbag/functions/fnc_BIArsenalClose.sqf
```diff
@@ -0,0 +1,22 @@
...
+};
+
+GVAR(arsenalCache) = nil;
```
commy2, line 22: Function is returning an assignment, which is bad design in SQF.

addons/gunbag/functions/fnc_BIArsenalOpen.sqf
```diff
@@ -0,0 +1,20 @@
...
+ * None
+ *
+ * Example:
+ * [] call ace_gunbag_fnc_BIArsenalOpen
+ *
+ * Public: No
+ */
+
+if ([player] call FUNC(hasGunBag)) then {
+    GVAR(arsenalCache) = (backpackContainer player) getVariable [QGVAR(gunbagWeapon), []];
```
commy2, line 19: Same as above. Probably does not handle the remote controlled units.

In any case, I would pass around the actual unit instead of relying on global state.

addons/gunbag/functions/fnc_BIArsenalClose.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Example:
+ * [] call ace_gunbag_fnc_BIArsenalClose
+ *
+ * Public: No
+ */
+
+params ["_unit"];
+
+if (!isNil QGVAR(arsenalCache)) then {
+    (backpackContainer _unit) setVariable [QGVAR(gunbagWeapon),GVAR(arsenalCache), true];
```
commy2, line 21:
```suggestion
    backpackContainer _unit setVariable [QGVAR(gunbagWeapon), GVAR(arsenalCache), true];
```
White space and parentheses.

PR #7896, opened 2020-09-04
Arsenal - Add ACRE2 Notification System support

jonpas:
- Same as #7895 but for Arsenal.


jonpas:
Decided to do by IDD directly in ACRE2, a bit easier to work with. Ref. https://github.com/IDI-Systems/acre2/pull/1019

PR #7900, opened 2020-09-08
Medical - Add Treatment Time Addon Settings

dgibso29:
Adds treatment time addon settings for Tourniquets, IV Bags, Syringes, Splints, and Body Bags.

Each value has been defaulted to the current default time for each type, and has been set in a range [0.1,  (currentDefault * 4)].

I have tested all 5 of these new settings without issue.

image

Thank you.

### DIFF
--- a/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
+++ b/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
@@ -79,7 +79,7 @@ class GVAR(actions) {
         icon = QPATHTOEF(medical_gui,ui\tourniquet.paa);
         allowedSelections[] = {"LeftArm", "RightArm", "LeftLeg", "RightLeg"};
         items[] = {"ACE_tourniquet"};
-        treatmentTime = 7;
+        treatmentTime = QGVAR(treatmentTimeTourniquet);
         condition = QUOTE(!([ARR_2(_patient,_bodyPart)] call FUNC(hasTourniquetAppliedTo)));
         callbackSuccess = QFUNC(tourniquet);
         litter[] = {};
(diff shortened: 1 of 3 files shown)


Elgin675:
Good idea :) ! But it doesn't "disturb" too much the animations ?

I have a suggestion. Would it be possible to add coefficients for medics/doctors ?
I think there is already an hardcoded coefficient for bandaging.

Can I redo the French translation, so that it is consistent with the rest of the other French translations (on the form) ?

dgibso29:
@Elgin675 
> Good idea :) ! But it doesn't "disturb" too much the animations ?
Entirely possible, but ultimately that would be the user/community's choice were they to desire extraordinarily low treatment times. Default values mirror the current hard-coded values, so they'd have to go out of their way to choose something low.

This PR came about because our community generally does not uses tourniquets due to their application time often being similar to bandages. Unless you happen to be hit badly in one limb, it's generally the 'better' choice to just take the extra few seconds to apply 2 bandages. My goal is to promote tourniquet use by halving the application time. Exposing the other treatment times was a natural extension of this. Configurable values > hard-coded IMO! :)

> I have a suggestion. Would it be possible to add coefficients for medics/doctors ?
> I think there is already an hardcoded coefficient for bandaging.
I considered exposing all 3 of these but didn't want to make too many changes in one go. I'll take another look and see what it will take to do so.

> I'll redo the French translation if you don't mind, so that it's consistent with the rest of the mod's translations.
Please do!

Cyruz143:
I've been using a tweak to drop the tourniquet apply times to 1 second for a while, if the time is lower than some magic value, it just doesn't play the anim so it should be ok for people setting very low values.

Elgin675:
Understood, thank you @Cyruz143 !

PabstMirror (approved):

PR #7901, opened 2020-09-09
Updated to directly check for open wounds

dgibso29:
Updated per PR feedback


dgibso29:
Somehow ended up on acemod:master instead of anden3:master. Closing.

PR #7907, opened 2020-09-17
Allow toggling ACE viewdistance for clients

Freddo3000:
I can't really find any reason why it isn't allowed right now. If there is one, please enlighten me.

- Allow clients to enable/disable the ACE view distance system, instead of being forced by the server.


shukari:
Im against this PR. In my group we use this to ensure a max viewdistance for specific maps.
It should be the server right to set it or not set it... 
In the current state the server or mission can block it OR let the client decide... your change hindered my needs as server owner.

Freddo3000:
> 
> 
> Im against this PR. In my group we use this to ensure a max viewdistance for specific maps.
> It should be the server right to set it or not set it...
> In the current state the server or mission can block it OR let the client decide... your change blocks all customisation.

You're mixing it up I believe. I'm not disabling the server being able to force the setting, I'm just allowing the setting to be changed by clients without the server overriding it, just like the rest of the settings.

You'll still be able to force the setting via mission/server overrides, just like all the other settings if you look at the edited file.

shukari:
> > Im against this PR. In my group we use this to ensure a max viewdistance for specific maps.
> > It should be the server right to set it or not set it...
> > In the current state the server or mission can block it OR let the client decide... your change blocks all customisation.
> 
> You're mixing it up I believe. I'm not disabling the server being able to force the setting, I'm just allowing the setting to be changed by clients without the server overriding it, just like the rest of the settings.
> 
> You'll still be able to force the setting via mission/server overrides, just like all the other settings if you look at the edited file.

Check youre are right. I was thinking of isGlobal = 2 (https://github.com/CBATeam/CBA_A3/wiki/CBA-Settings-System#global-and-local-settings-_isglobal) and was in a shock state ^^

BaerMitUmlaut:
This still uses the old settings file. Would you mind converting this to CBA settings?

Freddo3000:
> 
> 
> This still uses the old settings file. Would you mind converting this to CBA settings?

I could do that

addons/viewdistance/XEH_preInit.sqf
```diff
@@ -6,4 +6,15 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+#include "initSettings.sqf"
+
+if (GVAR(enabled) && hasInterface && !is3DEN) then {
```
veteran29, line 11: Should be executed on `CBA_settingsInitialized` event. Otherwise the server overwrite setting might not be received yet and it will have no effect.

addons/viewdistance/functions/fnc_RscDisplayOptionsVideo.sqf
```diff
@@ -0,0 +1,55 @@
...
+ * Arguments:
+ * 0: Video options display <DISPLAY>
+ *
+ * Return Value:
+ * Nothing
+ *
+ * Public: No
+ */
+
+disableSerialization;
```
veteran29, line 18: Not needed in unscheduled.

veteran29 (changes requested):

addons/viewdistance/XEH_preInit.sqf
Freddo3000: Good to know

jonpas:
@veteran29

addons/viewdistance/XEH_preInit.sqf
```diff
@@ -6,4 +6,24 @@ PREP_RECOMPILE_START;
...
+
+            ["vehicle",{
+                [false] call FUNC(adaptViewDistance);
+            }] call CBA_fnc_addPlayerEventHandler;
+
+            ["ACE_controlledUAV", {
+                [false] call FUNC(adaptViewDistance);
+            }] call CBA_fnc_addEventHandler;
+
+        };
```
veteran29, line 24:
```suggestion
        if (!GVAR(enabled)) exitWith {};

        ["vehicle",{
            [false] call FUNC(adaptViewDistance);
        }] call CBA_fnc_addPlayerEventHandler;

        ["ACE_controlledUAV", {
            [false] call FUNC(adaptViewDistance);
        }] call CBA_fnc_addEventHandler;

```

addons/viewdistance/initSettings.sqf
```diff
@@ -0,0 +1,73 @@
+private _distArr = [
+    [0,                      1,     2,      3,      4,      5,      6,      7,      8,      9,      10,     11,     12,     13,     14],
+    [LSTRING(videosettings), "500", "1000", "1500", "2000", "2500", "3000", "3500", "4000", "5000", "6000", "7000", "8000", "9000", "10000"],
+    0
```
veteran29, line 4: There are no official guidelines on array formating IIRC but we don't do this in the codebase usually. Can you format the array "normally".

addons/viewdistance/initSettings.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+[
+    QGVAR(objectViewDistanceCoeff),
+    "LIST",
+    [LSTRING(object_DisplayName),LSTRING(object_Description)],
+    _categoryStr,
+    [
+        [0,                   1,                       2,                   3,                      4,                    5,                        6                       ],
+        [LSTRING(object_off), LSTRING(object_verylow), LSTRING(object_low), LSTRING(object_medium), LSTRING(object_high), LSTRING(object_veryhigh), LSTRING(object_fovBased)],
+        0
```
veteran29, line 67: Same as above.

veteran29 (changes requested):

veteran29 (approved):

johnb432:
I believe #8461 and #9140 addressed this.
Either way, imo too many merge conflicts. If this is something that still needs integrating, open a new PR.

PR #7922, opened 2020-10-01
remove useless file

bux:


mharis001 (approved):

jonpas:
Isn't the entire point of the lock file to be in source control? So whoever pulls it gets the EXACT same setup?

jonpas:
Never mind, this one was in the wrong folder, thanks @jokoho48.

PR #7924, opened 2020-10-01
Reload/Interaction - Fix 2 hemtt ingame errors

jokoho48:
- Fixes 2 issues that occur when building with hemtt

### DIFF
--- a/addons/interaction/ACE_ZeusActions.hpp
+++ b/addons/interaction/ACE_ZeusActions.hpp
@@ -46,8 +46,7 @@ class ACE_ZeusActions {
                 )]; \
                 bis_fnc_curatorObjectPlaced_mouseOver = [ARR_2('OBJECT',_unit)]; \
                 private _rc = group _target createUnit [ARR_5('ModuleRemoteControl_F',[ARR_3(0,0,0)],[],0,'NONE')]; \
-                _rc setVariable [ARR_2('BIS_fnc_initModules_disableAutoActivation',false)]; \
-            );
+                _rc setVariable [ARR_2('BIS_fnc_initModules_disableAutoActivation',false)];);
         };
 
         class GVAR(repair) {
--- a/addons/interaction/ACE_ZeusActions.hpp
+++ b/addons/interaction/ACE_ZeusActions.hpp
@@ -35,18 +35,7 @@ class ACE_ZeusActions {
             displayName = "$STR_A3_CfgVehicles_ModuleRemoteControl_F";
             icon = "\A3\Modules_F_Curator\Data\portraitRemoteControl_ca.paa";
             condition = QUOTE(ZEUS_ACTION_CONDITION && {-1 < (curatorSelected select 0) findIf {!isNull effectiveCommander _x}});
-            statement = QUOTE( \
-                private _units = curatorSelected select 0; \
-                private _unit = _units param [ARR_2( \
-                    _units findIf { \
-                        side _x in [ARR_4(east,west,resistance,civilian)] \
-                        && !(isPlayer _x) \
-                    }, \
-                    objNull \
-                )]; \
-                bis_fnc_curatorObjectPlaced_mouseOver = [ARR_2('OBJECT',_unit)]; \
-                private _rc = group _target createUnit [ARR_5('ModuleRemoteControl_F',[ARR_3(0,0,0)],[],0,'NONE')]; \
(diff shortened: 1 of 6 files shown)


addons/reload/CfgEventHandlers.hpp
```diff
@@ -27,8 +27,7 @@ class Extended_Take_EventHandlers {
                     && {GVAR(DisplayText)} \
                     && {_container in [ARR_3(uniformContainer _unit, vestContainer _unit, backpackContainer _unit)]} \
                     && {_item == currentMagazine _unit} \
-                ) then {_unit call DFUNC(displayAmmo)}; \
-            );
+                ) then {_unit call DFUNC(displayAmmo)};);
```
BaerMitUmlaut, line 30: Why is this a DFUNC?

bux (changes requested):
Please split the PR in two and add proper titles for the changelog describing the issue in more detail

addons/reload/CfgEventHandlers.hpp
jokoho48: dont ask me :P i just fixed the general error that happend when you build it with hemtt

jokoho48:
> Please split the PR in two and add proper titles for the changelog describing the issue in more detail

tbh I think this should be excluded from the changelog because it is not a thing that the user really interests or is currently an issue for the builds we release.

veteran29:
I don't like the idea of making the code less readable because the build tool fails, I guess it's hard to fix this on HEMTT side?

PabstMirror:
I generally dislike any complex code in configs, `ARR_X` macros and mixing quotes types is a real pain to read and maintain.
We could just make it a new sqf function and have a simple `call` in config

PabstMirror (approved):

addons/interaction/functions/fnc_doRemoteControl.sqf
```diff
@@ -0,0 +1,25 @@
+#include "script_component.hpp"
+/*
+ * Author: ?
```
veteran29, line 3:
```suggestion
 * Author: Dystopian
```

ref: https://github.com/acemod/ACE3/pull/6632

veteran29 (changes requested):

Dystopian (changes requested):
also function description should be filled

addons/interaction/functions/fnc_doRemoteControl.sqf
Dystopian, line 3:
```suggestion
 * Author: SilentSpike, joko
```
ref #1519

addons/interaction/ACE_ZeusActions.hpp
```diff
@@ -35,19 +35,7 @@ class ACE_ZeusActions {
...
-                        side _x in [ARR_4(east,west,resistance,civilian)] \
-                        && !(isPlayer _x) \
-                    }, \
-                    objNull \
-                )]; \
-                bis_fnc_curatorObjectPlaced_mouseOver = [ARR_2('OBJECT',_unit)]; \
-                private _rc = group _target createUnit [ARR_5('ModuleRemoteControl_F',[ARR_3(0,0,0)],[],0,'NONE')]; \
-                _rc setVariable [ARR_2('BIS_fnc_initModules_disableAutoActivation',false)]; \
-            );
+            statement = QUOTE(call FUNC(doRemoteControl));
```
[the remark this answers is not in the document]
jokoho48: we use below also FUNC and in the hole project, we use FUNC in Configs and not DFUNC.
there are currently 738 uses of FUNCs in Configs and 119 DFUNCs and until now I never saw any issues reported with the use of FUNC.
also, I don't see any reason that macro should not work in configs.

PabstMirror (approved):

addons/interaction/ACE_ZeusActions.hpp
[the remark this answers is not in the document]
PabstMirror: I FUNC/DFUNC is only relevant for the stack tracing which I don't think gets much use at all in ace

PabstMirror (approved):

addons/interaction/ACE_ZeusActions.hpp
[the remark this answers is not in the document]
jonpas: Use DFUNC if you need it for callstack, we weren't very consistent with it from the start, so it doesn't matter at all now.

bux (approved):

PR #7929, opened 2020-10-02
Utilize isNotEqualTo

Neviothr:
- title, added in dev 2.01


addons/markers/functions/fnc_onMouseButtonDown.sqf
```diff
@@ -23,7 +23,7 @@
 
 params ["_ctrlMap", "_button", "_x", "_y", "_shift", "_ctrl", "_alt"];
 
-if (_button != 0 || {!([_shift, _ctrl, _alt] isEqualTo [false, false, true])}) exitWith {};
+if (_button != 0 || {[_shift, _ctrl, _alt] isEqisNotEqualToualTo [false, false, true]}) exitWith {};
```
Vdauphin, line 26:
```suggestion
if (_button != 0 || {[_shift, _ctrl, _alt] isNotEqualTo [false, false, true]}) exitWith {};
```

addons/scopes/functions/fnc_inventoryCheck.sqf
```diff
@@ -123,19 +123,19 @@ private _newGuns = [primaryWeapon _player, secondaryWeapon _player, handgunWeapo
...
 } forEach [0, 1, 2];
 
 if (GVAR(correctZeroing) || GVAR(simplifiedZeroing)) then {
-    if (!(_unitBaseAngle isEqualTo (_player getVariable [QGVAR(baseAngle), [0,0,0]]))) then {
+    if (_unitBaseAngle isNotEqualTo (_player getVariable [QGVAR(baseAngle), [0,0,0]])) then {
         TRACE_2("syncing",_unitBaseAngle,_player getVariable QGVAR(baseAngle));
         _player setVariable [QGVAR(baseAngle), _unitBaseAngle, true];
     };
-    if (!(_unitBoreHeight isEqualTo (_player getVariable [QGVAR(boreHeight), [0,0,0]]))) then {
+    if (_unitBoreHeight isEquaisNotEqualTolTo (_player getVariable [QGVAR(boreHeight), [0,0,0]])) then {
```
Vdauphin, line 138:
```suggestion
    if (_unitBoreHeight isNotEqualTo (_player getVariable [QGVAR(boreHeight), [0,0,0]])) then {
```

Vdauphin (changes requested):
:)

addons/interact_menu/functions/fnc_collectActiveActionTree.sqf
```diff
@@ -48,7 +48,7 @@ _fullPath pushBack _actionName;
 private _activeChildren = [];
 
 // If there's a statement to dynamically insert children then execute it
-if !({} isEqualTo _insertChildrenCode) then {
+if ({} isNotEqualTo _insertChildrenCode) then {
```
Neviothr, line 51: Can anyone shed some light on why this fails in SQF validator?

addons/map/XEH_postInitClient.sqf
```diff
@@ -47,7 +47,7 @@ GVAR(flashlights) = [] call CBA_fnc_createNamespace;
                 private _unitLight = _player getVariable [QGVAR(flashlight), ["", objNull]];
                 _unitLight params ["_flashlight", "_glow"];
                 if (_mapOn) then {
-                    if (!(_flashlight isEqualTo "") && {isNull _glow}) then {
+                    if (_flashlight isNotEqualTo "") && {isNull _glow}) then {
```
Neviothr, line 50: Here, too.

addons/interact_menu/functions/fnc_collectActiveActionTree.sqf
jokoho48: because of this: https://github.com/acemod/ACE3/blob/master/tools/sqf_validator.py#L11
isNotEqualTo is not a valid command after Code. just add it in the array and the issue should be solved

addons/map/XEH_postInitClient.sqf
jokoho48: here the linter is failing because there is a bracket closing to much
```suggestion
                    if (_flashlight isNotEqualTo "" && {isNull _glow}) then {
```

addons/interact_menu/functions/fnc_collectActiveActionTree.sqf
commy2: Just reverse it to
```
_insertChildrenCode isNotEqualTo {}
``´

Makes more sense anyway.

jokoho48: this should still be added to the validator!

jokoho48:
I think isNotEqualTo should be added to the sqf_validator.

PabstMirror (approved):

PR #7935, opened 2020-10-03
Fastroping - Fix ARR macro

PabstMirror:


mharis001 (approved):

Dahlgren:
This breaks the fix made in https://github.com/acemod/ACE3/pull/7931

PR #7936, opened 2020-10-04
Fastroping - Fix inconsistent usage of ARR macro

Dahlgren:
- Change remaining conditions to use nested ARR macros like other statements in fastroping CfgVehicles
- Fix build of fastroping with newer mikero tools

### DIFF
--- a/addons/fastroping/CfgVehicles.hpp
+++ b/addons/fastroping/CfgVehicles.hpp
@@ -221,7 +221,7 @@ class CfgVehicles {
         class UserActions {
             class Ramp_Open;
             class Ramp_Close: Ramp_Open {
-                condition = QUOTE([ARR_5(this,'CargoRamp_Open',[[0],[1],[2]])] call FUNC(canCloseRamp));
+                condition = QUOTE([ARR_3(this,'CargoRamp_Open',[ARR_3([0],[1],[2])])] call FUNC(canCloseRamp));
             };
         };
     };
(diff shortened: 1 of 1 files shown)


PabstMirror:
there are 4 commas, so ARR_5
otherwise it throws
```
File z\ace\addons\fastroping\CfgVehicles.hpp, line 210: '/CfgVehicles/Heli_Transport_02_base_F/UserActions/Ramp_Close.condition': Missing ';' at the end of line
Error context ])] call ace_fastroping_fnc_canCloseRamp);
ErrorMessage: File z\ace\addons\fastroping\CfgVehicles.hpp, line 210: '/CfgVehicles/Heli_Transport_02_base_F/UserActions/Ramp_Close.2': ']' encountered instead of '='
```

addons/fastroping/CfgVehicles.hpp
```diff
@@ -221,7 +221,7 @@ class CfgVehicles {
         class UserActions {
             class Ramp_Open;
             class Ramp_Close: Ramp_Open {
-                condition = QUOTE([ARR_5(this,'CargoRamp_Open',[[0],[1],[2]])] call FUNC(canCloseRamp));
+                condition = QUOTE([ARR_3(this,'CargoRamp_Open',[[0],[1],[2]])] call FUNC(canCloseRamp));
```
commy2, line 224: This is wrong though. There are 4 commas, so 5 arguments. Macros don't operate in SQF syntax, it's a preprocessor of raw text.

Dahlgren:
Changed the inner arrays to use ARR macros. Now it's compatible with both newer pboproject and cfgconvert.

commy2:
Bad change. The current implementation better reflects what is happening during preprocessing. The new one is adding complexity and hides from the reader how the macros are resolved.

jonpas (approved):
This is fine for now if it assures compatibility with the tools (remember, we want ease of access, not more restrictions to building ACE).

Should also be reported to Mikero and this PR can be reverted if it ever gets fixed.

PabstMirror (approved):

PR #7939, opened 2020-10-04
Corpse carrying and dragging

BaerMitUmlaut:
- Allow carrying or dragging corpses

When starting to carry or drag a corpse, a clone without any AI is created, which you can carry or drag just fine. Once the clone is dropped, the original corpse is teleported and the clone is destroyed.

Needs MP testing.


addons/dragging/functions/fnc_dropClone.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Public: No
+ */
+params ["_clone"];
+
+private _unit = _clone getVariable [QGVAR(original), objNull];
+_unit setPosASL getPosASL _clone;
+_unit hideObjectGlobal false;
+
+detach _clone;
```
Dystopian, line 23:
```suggestion
// This is necessary before deleting. Unit will also be detached later
detach _clone;
```

addons/dragging/functions/fnc_dropClone.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Drops a draggable / carryable clone of a dead unit.
+ *
+ * Arguments:
+ * 0: Clone <OBJECT>
+ *
+ * Return Value:
+ * Original unit.
+ *
+ * Example:
+ * [player] call ace_dragging_fnc_createClone;
```
Dystopian, line 13:
```suggestion
 * [player] call ace_dragging_fnc_dropClone;
```

addons/dragging/functions/fnc_carryObject.sqf
```diff
@@ -26,6 +26,9 @@ private _direction = _target getVariable [QGVAR(carryDirection), 0];
 
 // handle objects vs persons
 if (_target isKindOf "CAManBase") then {
+    if (!alive _target) then {
+        _target = [_target] call FUNC(createClone);
+    };
```
Dystopian, line 31:
```suggestion
```
looks like this call is not necessary because there is another one in `fnc_startCarry` below. Not tested

Dystopian (changes requested):

addons/dragging/functions/fnc_createClone.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * [player] call ace_dragging_fnc_createClone;
+ *
+ * Public: No
+ */
+params ["_unit"];
+
+private _clone = QGVAR(clone) createVehicle [0, 0, 0];
+_clone setUnitLoadout getUnitLoadout _unit;
+_clone setVariable [QGVAR(original), _unit];
+_unit hideObjectGlobal true;
```
PabstMirror, line 22: needs to be run on server
```suggestion
[QEGVAR(common,hideObjectGlobal), [_unit, true]] call CBA_fnc_serverEvent;
```

addons/dragging/functions/fnc_dropClone.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Example:
+ * [player] call ace_dragging_fnc_dropClone;
+ *
+ * Public: No
+ */
+params ["_clone"];
+
+private _unit = _clone getVariable [QGVAR(original), objNull];
+_unit setPosASL getPosASL _clone;
+_unit hideObjectGlobal false;
```
PabstMirror, line 21:
```suggestion
[QEGVAR(common,hideObjectGlobal), [_unit, false]] call CBA_fnc_serverEvent;
```

PabstMirror:
Kill and drag clone
Rifle is left behind and can access inventory on it while clone is being dragged
Drop guy and you can access the dead's inventoy from both the gun (at spot of death) and his body (and they are synced 😕 )

image

BaerMitUmlaut:
@PabstMirror I can only replicate the "synched duplicate inventories" when the body is still nearby. The inventory action has a maximum distance of 3.5 m according to my tests. I can only access the gun beyond that.

Italiandrache:
It is unclear to me whether the weapon being left behind is or not intended behavior. I think it is reasonable for the weapon to be moved alongside the corpse.

jonpas:
It should probably be moved as realistically you would have some kind of a sling on you. However, I can see the implementation problem with trying to get those weapons (they are ground weapon holders at that point).

BaerMitUmlaut:
Moving the weapon holder with the body could cause a bunch more issues (garbage collections might fail, getting the correct weapon holder in itself is a problem, etc), so I'd rather not do that. I don't think it's that big of a deal tbh.

BaerMitUmlaut:
Interesting, thank you for testing. Sounds like the body falls through the map? Maybe setting the position remotely would work better?

Dystopian:
> 
> 
> Interesting, thank you for testing. Sounds like the body falls through the map? Maybe setting the position remotely would work better?

I don't know. Can't work on fix yet.
Looks like bug with dropped dead body weapon I met in #6029. But it was fixed since I created that PR so I deleted workaround [there](https://github.com/acemod/ACE3/pull/6029/commits/8603ae08dc78438710650b793f33add084bc957a).

Dystopian:
Doesn't work if JIP for men who are already dead when JIP. Looks like init EH doesn't fire for dead men.

johnb432:
Since https://github.com/acemod/ACE3/pull/9273 was merged, closing this one - can be reopened if necessary.

PR #7941, opened 2020-10-04
Medical - Allow placing unconscious in a body bag

BrettMayson:
- Adds a setting allowing the placing of unconscious units in a body bag

Makes it a little harder for medics to rapidly determine if someone is dead or unconscious

### DIFF
--- a/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
+++ b/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
@@ -245,9 +245,7 @@ class GVAR(actions) {
         medicRequired = 0;
         treatmentTime = 15;
         items[] = {"ACE_bodyBag"};
-        // 'vehicle _patient' always returns the body
-        // '_patient in _patient' always false for body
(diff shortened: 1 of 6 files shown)


PabstMirror:
ref https://github.com/acemod/ACE3/blob/master/addons/medical_treatment/functions/fnc_placeInBodyBag.sqf
```
if (alive _patient) then {
    TRACE_1("Manually killing with setDead",_patient);
    [_patient, "buried_alive"] call EFUNC(medical_status,setDead);
};
```
lol

jonpas:
We should think about how the user (medic) can figure out if someone is dead. Not by simple interaction but through vitals.

PabstMirror:
What if we always show the action on unconscious units and allow the progress bar to go to completion
but only do the hard killing if setting is enabled, 
otherwise show a little hint like "The body twitched and may not be dead"

jonpas:
Yes, I like that idea @PabstMirror !

BrettMayson:
I like that idea too, but would like it to be an on by default setting we can turn off to achieve the original functionality of this PR.

jonpas (approved):

PR #7947, opened 2020-10-07
Markers - Add timestamps

Freddo3000:
- Add a checkbox when creating/editing markers to add a timestamp suffix.
- Requires a watch
- Move IDCs to macros.

image
image

Credit to @milaq for https://github.com/milaq/arma3_timemarkers from which this was inspired.


BrettMayson:
Should instead be an option under direction, rather than tucked in the corner

addons/markers/functions/fnc_canTimestamp.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Return Value:
+ * Whether unit is able to timestamp <BOOL>
+ *
+ * Example:
+ * [ACE_Player] call ace_markers_fnc_canTimestamp
+ *
+ * Public: No
+ */
+
+params [["_unit",ACE_Player]];
```
BrettMayson, line 18:
```suggestion
params [["_unit", ACE_player]];
```

addons/markers/functions/fnc_initInsertMarker.sqf
```diff
@@ -213,6 +231,30 @@
     _buttonCancel ctrlSetPosition _pos;
     _buttonCancel ctrlCommit 0;
 
+    ////////////////////
+    // init marker timestamp cb
+
+    if (GVAR(timestampEnabled) && [ACE_player] call FUNC(canTimestamp)) then {
```
BrettMayson, line 237:
```suggestion
    if (GVAR(timestampEnabled) && {[ACE_player] call FUNC(canTimestamp)}) then {
```

Freddo3000:
> 
> 
> Should instead be an option under direction, rather than tucked in the corner

I figured that considering it ties into the description, it should be close to it. I also wanted to keep the vertical space down.

addons/markers/functions/fnc_initInsertMarker.sqf
```diff
@@ -213,6 +231,30 @@
...
+    if (GVAR(timestampEnabled) && {[ACE_player] call FUNC(canTimestamp)}) then {
+        _buttonOK ctrlAddEventHandler ['ButtonClick', {
+            if (GETUVAR(GVAR(timestampChecked),false)) then {
+                params ["_buttonOk"];
+                private _description = (ctrlParent _buttonOk) displayctrl IDC_INSERT_MARKER;
+
+                _description ctrlSetText format [ // Add timestamp suffix
+                    "%1 [%2]",
+                    ctrlText _description,
+                    [daytime, "HH:MM"] call BIS_fnc_timeToString
```
bux, line 246: maybe provide a setting for time format and 24/12 h display

BrettMayson:
> I figured that considering it ties into the description, it should be close to it. I also wanted to keep the vertical space down.

Vertical space is less of a concern than UX, maybe it could go under description then rather than at the bottom to keep it closer to related fields (Description)

addons/markers/functions/fnc_initInsertMarker.sqf
Freddo3000: image

Freddo3000:
image

addons/markers/functions/fnc_initInsertMarker.sqf
```diff
@@ -29,26 +30,28 @@
...
     //BIS Controls:
-    private _text = _display displayctrl 101;
-    private _picture = _display displayctrl 102;
-    private _channel = _display displayctrl 103;
-    private _buttonOK = _display displayctrl 1;
-    private _buttonCancel = _display displayctrl 2;
-    private _description = _display displayctrl 1100;
-    private _title = _display displayctrl 1001;
-    private _descriptionChannel = _display displayctrl 1101;
+    private _text =                 _display displayctrl IDC_INSERT_MARKER;
```
jokoho48, line 33: why do you add those spaces?

addons/markers/stringtable.xml
```diff
@@ -150,5 +150,48 @@
...
         </Key>
+        <Key ID="STR_ACE_Markers_TimestampEnabled">
+            <English>Allow Timestamps</English>
+        </Key>
+        <Key ID="STR_ACE_Markers_TimestampEnabledDescription">
+            <English>Whether to allow timestamps to be automatically applied to markers</English>
+        </Key>
+
+        <Key ID="STR_ACE_Markers_TimestampTooltip">
+            <English>&lt;t size='0.8'&gt;Add Timestamp:&lt;/t&gt;</English>
```
jokoho48, line 160: why is here styling in the stringtable?

addons/markers/functions/fnc_initInsertMarker.sqf
Freddo3000: I find it more readable, but I can revert it otherwise.

jokoho48: I think it breaks the readability but that a maintainer should decide. I think we have it nowhere else in ace. I am also not sure if it is against coding style.

addons/markers/stringtable.xml
Freddo3000: Skips needing to format it in script

jokoho48: but may break if we have community translators that don't know what that is and maybe change something. I think styling should never be in a translation medium.

addons/markers/functions/fnc_initInsertMarker.sqf
veteran29: We usually don't do that here.

PabstMirror:
https://github.com/Freddo3000/ACE3/pull/1
I want to prevent this when editing markers


I tested an idea I had that uses a special asci char (that just looks like a space) to indicate that a timecode follows.
fnc_initInsertMarker:95
```
    if !((markerText GVAR(editingMarker)) isEqualTo "") then {
        // fill text input with text from marker which is being edited
        private _originalText = markerText GVAR(editingMarker);
        private _timeIndex = _originalText find ((toString [129]) + "["); // 129 just looks like a space
        if (_timeIndex > 0 ) then { _originalText = _originalText select [0,_timeIndex]; }; // stip off timecode
        _text ctrlSetText _originalText;
    };
```

fnc_onButtonClickConfirm:30
```
        "%1%2[%3%4]",
        ctrlText _description,
        toString [129],....
```

This means when you edit a marker it will automatically update the timecode instead of appending it.

addons/markers/functions/fnc_initInsertMarker.sqf
Freddo3000: Though is it something that I will have to change for this to be approved?

veteran29: Personally I think yes, do we have any guidelines on stuff like this? @jonpas ?

Drofseh:
I could see it being useful to have a history sometimes `TEXT [1345] [1353] [1405]`.

Maybe a second button to add instead of replace?

addons/markers/functions/fnc_initInsertMarker.sqf
PabstMirror: informal slack poll had mixed to negative results for this type of formatting
probably ok to leave in the PR, but it might get cleaned up in later refactoring

Freddo3000:
> 
> 
> I could see it being useful to have a history sometimes `TEXT [1345] [1353] [1405]`.
> 
> Maybe a second button to add instead of replace?

At that point you might as well place a new marker, imo.

Drofseh:
> > I could see it being useful to have a history sometimes `TEXT [1345] [1353] [1405]`.
> > Maybe a second button to add instead of replace?
> 
> At that point you might as well place a new marker, imo.

yah maybe...

veteran29 (approved):
LGTM

addons/markers/functions/fnc_initInsertMarker.sqf
veteran29: Resolving the conversation then.

PabstMirror (approved):

jonpas (approved):

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -0,0 +1,42 @@
...
+if (cbChecked _aceTimestamp && {[ACE_player] call FUNC(canTimestamp)}) then {
+    private _time = daytime;
+    private _ampm = switch (true) do {
+        case (GVAR(timestampHourFormat) == 24): {""};
+        case (_time < 12): {" am"};
+        case (_time > 12): {SUB(_time,12); " pm"};
+    };
+
+    _description ctrlSetText format [ // Add timestamp suffix
+        "%1%2[%2%3]",
```
dedmen, line 34: This doesn't seem right?
description, space, space, time

you give 4 parameter but only use 3, and the space seems wrong

Freddo3000: Yeah looks like I made a typo, should probably be `"%1%2[%3%4]"`

commy2: Yes, am-pm is completely broken in this pull request. Even if this part was repaired, am-pm will not be shown for 12 and 24-hours, and even if that part is repaired, it will show the wrong one.

Should all be fixed here: https://github.com/acemod/ACE3/pull/8034/files#diff-4b395ce5b5e0cc52ea30d46cdc2cb9d5fea1eb22d6ca4cfe885b60ef68b1e45eR48

PR #7948, opened 2020-10-07
Arsenal - Provide reference to the current box

BrettMayson:
- Provide a reference to the current box

Useful for mods that modify the arsenal

### DIFF
--- a/addons/arsenal/functions/fnc_openBox.sqf
+++ b/addons/arsenal/functions/fnc_openBox.sqf
@@ -34,14 +34,14 @@ if (canSuspend) exitWith {
     [{_this call FUNC(openBox)}, _this] call CBA_fnc_directCall;
 };
 
-GVAR(currentBox) = _object;
-
 private _displayToUse = [findDisplay 46, findDIsplay 312] select (!isNull findDisplay 312);
 _displayToUse = [_displayToUse, findDisplay 313] select (is3DEN);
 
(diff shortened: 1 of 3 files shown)


bux:
why `box`?

BrettMayson:
The related functions are `initBox`, `openBox`, and `removeBox`

bux:
Yeah, I know, pretty unlucky choice of names ... shame on you alganthe 😄 

Imho, I'd prefer something like `currentArsenalObject` or similar.

BrettMayson:
If it were a new feature, yeah, I'd agree. I do think it is better however to stay consistent with the existing public functions.

addons/arsenal/functions/fnc_openBox.sqf
```diff
@@ -34,6 +34,8 @@ if (canSuspend) exitWith {
     [{_this call FUNC(openBox)}, _this] call CBA_fnc_directCall;
 };
 
+GVAR(currentBox) = _object;
```
mharis001, line 37: Should move below L44 since function might exit without creating display and then `onArsenalClose` won't run to reset the variable to `objNull`.

PabstMirror (changes requested):
see above

PabstMirror (approved):

veteran29 (approved):

PR #7951, opened 2020-10-08
Medical - Adjustable times for many treatment actions

diwako:
- Add CBA settings for most constant treatment times
  - Default values are the original values
- Add variable time needed to stitch one wound

Currently changed treatments:
- Tourniquet (apply and remove)
- Splint
- Autoinjectors
- IV bags
- CPR
- Stitching


Elgin675:
It's not the same ?
https://github.com/acemod/ACE3/pull/7900

diwako:
Yeah that does look the same. only difference is i am including cpr and stitching time, Could edit this PR and kick out the ones that are duplicate.

GlutenFreeVapes:
Can you please also make a slider to increase bandage power?

dgibso29:
If I'm not mistaken, there are already Addon Options controlling CPR & Stich time. That's why they weren't included in my PR.

diwako:
As far as I see (at least on master branch). I do not see any adjustable settings for those two. Stitching is controlled by a constant pre processor instruction and CPR is constantly defined in config at the moment.

Might be best if you add those two in your PR, then I will close this one as it has become fully redundant.

dgibso29:
> As far as I see (at least on master branch). I do not see any adjustable settings for those two. Stitching is controlled by a constant pre processor instruction and CPR is constantly defined in config at the moment.
> 
> Might be best if you add those two in your PR, then I will close this one as it has become fully redundant.

Ah, my mistake. I was thinking of the PAK coefficient.

Sure, will do when I have a moment.

PabstMirror:
Combined with changes from #7900 
now just changing cpr and stitch

PabstMirror (approved):

jonpas (approved):

PR #7952, opened 2020-10-09
Arsenal - Sort by protection (helm, uniform, vest)

PabstMirror:
Idea From @Elgin675 - https://github.com/acemod/ACE3/issues/3594#issuecomment-698439972

### DIFF
--- a/addons/arsenal/ACE_Arsenal_Sorts.hpp
+++ b/addons/arsenal/ACE_Arsenal_Sorts.hpp
@@ -62,10 +62,15 @@ class GVAR(sorts) {
         statement = QUOTE(_this call FUNC(sortStatement_magCount));
     };
 
-    class ACE_protection: sortBase {
+    class ACE_protectionBallistic: sortBase {
         scope = 2;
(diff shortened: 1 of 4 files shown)


Elgin675:
Thank you very much @PabstMirror !
After a quick test, it works well !
But in the case of some vests with a higher explosive resistance than bullet resistance (V_PlateCarrierGL_xxx), the order is not good.

Is it due to the fact that the filter takes into account the highest resistance, without distinction between ballistic and explosion ?

Tri

PabstMirror:
The formula is roughly `ballistic * 2 + explosive`; we could change coefficient to 10x to make explosive less important.

Elgin675:
I think it's a good idea ;) !
Or separate the sorting by ballistic protection and the sorting by explosive protection, but it seems a bit extreme :D ^^ !

BrettMayson:
Personally I was going to split this into two

Elgin675:
Work's fine, thank you :) !

jonpas (approved):

PR #7955, opened 2020-10-11
Medical AI - Fix AI healing itself too fast when hit

### LINKED ISSUE #7649: AI will heal right after being shot if they've not shot recently
- **Arma 3:** `1.98` (profiling)
- **CBA:** `3.15.0`
- **ACE3:** `3.13.1`
_Make sure to reproduce the issue with only CBA and ACE3 on a newly created mission!_

**Description:**
AI starts healing instead of fighting back if you suprise them and shot them before they can shoot back..

**Steps to reproduce:**
- See this video: https://streamable.com/emkfic

**Expected behavior:**
AI does not heal right after being shot.

- SP/MP

**Log Files:**
- [Arma3_x64_2020-04-19_16-02-43.rpt.txt](https://github.com/acemod/ACE3/files/4499245/Arma3_x64_2020-04-19_16-02-43.rpt.txt)

BaerMitUmlaut:
- Fix #7649

### DIFF
--- a/addons/medical_ai/XEH_postInit.sqf
+++ b/addons/medical_ai/XEH_postInit.sqf
@@ -9,6 +9,11 @@
         _unit setVariable [QGVAR(lastFired), CBA_missionTime];
     }] call CBA_fnc_addEventHandler;
 
+    ["CAManBase", "Hit", {
+        params ["_unit"];
+        _unit setVariable [QGVAR(lastHit), CBA_missionTime];
+    }] call CBA_fnc_addClassEventHandler;
+
     #include "stateMachine.sqf"
 }] call CBA_fnc_addEventHandler;
 
--- a/addons/medical_ai/XEH_preInit.sqf
+++ b/addons/medical_ai/XEH_preInit.sqf
@@ -8,4 +8,8 @@ PREP_RECOMPILE_END;
 
 #include "initSettings.sqf"
 
+// default time values for AI being ready to heal, used in fnc_isSafe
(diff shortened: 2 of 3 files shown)


veteran29 (approved):
LGTM

PabstMirror:
Would `[QEGVAR(medical,woundReceived)` be a better event? 
I think it might trigger less often

we could also just add this to medical engine to be made available for other things.  Engine_handleDamage: line 51
```
if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
    _unit setVariable [QEGVAR(medical,lastDamageSource), _shooter];
    _unit setVariable [QEGVAR(medical,lastInstigator), _instigator];
    _unit setVariable [QEGVAR(medical,lastHitTime), CBA_missionTime]; // <-- NEW
```

BaerMitUmlaut:
We're only setting a local object variable in the EH, I don't think we really need to optimize that, that's about as low impact as it gets. HandleDamage triggers more often than Hit, woundReceived should trigger about as often as Hit.

Something I forgot: I would like to bump the timer again with suppression, but the suppression EH will be a nice and fast solution so I want to wait for that rather.

PabstMirror (approved):

mharis001 (approved):

BrettMayson:
Would it be worth it to make it a setting?

BaerMitUmlaut:
I'd prefer if this stuff stays hidden from the average user and "just works".

BrettMayson:
Hidden global variable? I just find 30 seconds a little for my taste, but I also understand if it is too niche

PabstMirror (approved):

jonpas (approved):

PR #7959, opened 2020-10-16
Medical Engine - Prevent automatic unloading of dead or unconscious units

### LINKED ISSUE #7897: AI patients get kicked out of vehicle when player loads
**ACE3 and CBA3 tested along with noticing it with many other mods**
**Description:**
if you have an unconscious AI, stabilize them, load them on a vehicle for MEDEVAC they get kicked out if you get in the same vehicle, like it doesn't recognize when a loaded patient AI is occupying a slot

**Steps to reproduce:**
Load Patient onto vehicle
board vehicle
if you selected the same seat (or if you board the vehicle from the ramp where you don't get to pick spot) the patient gets dumped onto the ground and you take their seat

**Expected behavior:**
board the vehicle in the next available seat accounting for patients

Self hosted multiplayer, Eden Editor

BaerMitUmlaut:
- Fix #7897
- Prevent dead units from automatically getting unloaded when a player gets in


PabstMirror:
Make ai driver go unconc, then kill him, seat remains locked
Do we only need to run this on AI?
Maybe we could instead only lock the seat when loading an AI uncon

BaerMitUmlaut:
Yes, that is intentional. The original issue was about AI, but this would be another step towards making it harder to differentiate between unconscious and dead units.

If we'd only apply this to AI, I could see problems with Zeus controlled units.

PabstMirror:
Something needs to unlock the seat or the vehicle will be unusable for the rest of the mission

BaerMitUmlaut:
Interaction menu - unload patient

PabstMirror:
Arma Bug (maybe 2.00) unrelated to this PR but has breaking effects

Put AI Driver (`d1`) in Car (`c1`)
```
[] spawn {
    [d1, true, 9999] call ace_medical_fnc_setUnconscious;
    sleep 1;
    [d1, 20, "Head", "bullet"] call ace_medical_fnc_addDamageToUnit;
    sleep 1;
    systemChat format ["Dead %1-%2", vehicle D1, time];
    waitUntil {(vehicle d1) == d1};
    systemChat format ["WTF %1-%2", vehicle D1, time];
};
```
Dead @ 2.5 sec, WTF @ ~30 sec
since they aren't in a vehicle, `ACE_UnloadPatient` action does not show
but they still show in `fullCrew`?

image

jonpas:
Anyone tested this with 2.02 yet?

ZluskeN (approved):
I just tested this in SP and it still works as intended 👍

LinkIsGrim:
> Arma Bug (maybe 2.00) unrelated to this PR but has breaking effects
> 
> Put AI Driver (`d1`) in Car (`c1`)
> 
> ```
> [] spawn {
>     [d1, true, 9999] call ace_medical_fnc_setUnconscious;
>     sleep 1;
>     [d1, 20, "Head", "bullet"] call ace_medical_fnc_addDamageToUnit;
>     sleep 1;
>     systemChat format ["Dead %1-%2", vehicle D1, time];
>     waitUntil {(vehicle d1) == d1};
>     systemChat format ["WTF %1-%2", vehicle D1, time];
> };
> ```
> 
> Dead @ 2.5 sec, WTF @ ~30 sec since they aren't in a vehicle, `ACE_UnloadPatient` action does not show but they still show in `fullCrew`?
> 
> image

`vehicle` for dead units is unreliable. `objectParent` needs to be used instead (it's faster anyway).

LinkIsGrim (approved):

addons/interaction/functions/fnc_canPullOutBody.sqf
```diff
@@ -18,6 +18,9 @@
 
 params ["_body", "_unit"];
 
+// Defer to ACE Medical's unload patient if present
```
PabstMirror, line 21: I think we still need this for pulling out dead bodies

LinkIsGrim: UnloadPatient handles it now.

LinkIsGrim: Should have been handled by #9281

PR #7978, opened 2020-11-01
Repair - Fix spare wheels receiving fall damage

### LINKED ISSUE #6142: ACE_wheel object takes too much falling dammage
Just unloading a wheel via cargo will leave it with about 0.08 damage (unload pos is about 0.05 above).
Setpos it to 0.5 above pavement height will do about 0.25 damage.

HandleDamage for a 0.05m fall:
```
"[10d7a8100# 164012: ace_wheel.p3d,"""",0.0915974,<NULL-object>,"""",-1,<NULL-object>,""""]"
"[10d7a8100# 164012: ace_wheel.p3d,""mat_rim"",0.262721,<NULL-object>,"""",0,<NULL-object>,""HitBody""]"
```

BaerMitUmlaut:
- Fix #6142

The hitpoint is only used for visual effects but causes the tire to receive fall damage. Setting `minimalDamage` to 1 basically disables it taking damage on its own and thus disables fall damage completely. Other damage still works fine.

### DIFF
--- a/addons/repair/CfgVehicles.hpp
+++ b/addons/repair/CfgVehicles.hpp
@@ -377,13 +377,14 @@ class CfgVehicles {
         // can not take damage individually though, because of limitations of the thingX simulation type
         class HitPoints {
             class HitBody {
-                armor = 0.6;
+                armor = 1;
                 material = -1;
                 name = "mat_rim";
                 visual = "mat_rim";
                 passThrough = 1;
                 radius = 0.1;
                 explosionShielding = 1;
+                minimalHit = 1;
             };
         };
 


PabstMirror (approved):

Zumiii:
This is still an Issue. The Object will receive damage after unloading it from a vehicle or spawning it with fortify even if this Piece of Workaround-Code is added:

`["ACE_Wheel", "InitPost", {
	params ["_entity"];
	[
		{
			(_this select 0) setdamage 0;
		},
	 	[_entity],
		1
	] call CBA_fnc_waitAndExecute;
}, true, [], true] call CBA_fnc_addClassEventHandler;`

BaerMitUmlaut:
@Zumiii you are commenting on a pull request that was closed about half a year ago. Please create a new issue and provide the necessary info there.

PR #7980, opened 2020-11-02
Common / Medical - Fix loading patients into turret only seats

### LINKED ISSUE #2965: Can't load patient into the co-pilot seat of an AH-6
ACE3 Version: 3.4.0 stable

**Mods:**
- `@CBA_A3`
- `@ace`

**Placed ACE3 Modules:**
- Respawn Module

**Description:**
- You're not able to load a patient into the co-pilot seat of an AH-6 chopper. You get the option to load the patient when close to the chopper but nothing happens after selecting it.

**Steps to reproduce:**
- Shoot a dude to unconsciousness
- Have an AH-6 nearby
- Try to load the patient via the interaction menu -> "Load patient"
- Nothing happens

-> Expected result: The patient is loaded into the co-pilot seat

- Editor 

**RPT log file:**
http://pastebin.com/zC1W6iVy

### LINKED ISSUE #4937: fnc_loadperson.sqf / fnc_loadpersonlocal.sqf
**Arma 3 Version:** `x.xx` (stable / rc / dev) *ALL
**CBA Version:** `3.x.x` (stable / dev + commit hash) *ALL
**ACE3 Version:** `3.x.x` (stable / dev + commit hash)*ALL

**Mods:**
- `@CBA_A3`
- `@ace`
- '@rhs'

**Description:**
- Loading people into vehicles might not work that well for modded vehicles, (only checks for cargo and gunner.
- Rolling down windows on the RHS humvees will turn the seat from cargo into turret. (having to roll up to window to load a patient in is annoying) 

**Steps to reproduce:**
- fullcrew [rhs_uaz_open,"all",true] //[[<NULL-object>,"driver",-1,[],false],[<NULL-object>,"cargo",4,[],false],[<NULL-object>,"Turret",2,[0],true],[<NULL-object>,"Turret",5,[1],true],[<NULL-object>,"Turret",0,[2],true],[<NULL-object>,"Turret",1,[3],true],[<NULL-object>,"Turret",3,[4],true]]

So instead of using the method use right now, I offer a solution to use https://community.bistudio.com/wiki/fullCrew instead, with the combination of moveInTurret commad.

ALL SEATS IN VEHICLES ARE EQUAL. TREAT THEM AS EQUAL AND STOP TURRET SHAMING TODAY!

- EVERYWHERE

**Placed Modules:**
- none

**RPT log file:**
- not needed.

BaerMitUmlaut:
- Fix #2965 (from 2015 😄)
- Fix #4937
- Additionally, add a failsave when the eject action fails

I tested this with the AH-6 and eject only did not work before. This might fix this problem on other vehicles, too, not sure what causes it.

### DIFF
--- a/addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
+++ b/addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
@@ -22,5 +22,5 @@ private _nearVehicles = nearestObjects [_unit, ["Car", "Air", "Tank", "Ship_F", 
 _nearVehicles select {
     // Filter cargo seats that will eject unconscious units (e.g. quad bike)
     ((_x emptyPositions "cargo" > 0) && {!(_unit getVariable ['ACE_isUnconscious', false])} || {(getNumber (configFile >> "CfgVehicles" >> (typeOf _x) >> "ejectDeadCargo")) == 0})
-    || {_x emptyPositions "gunner" > 0}
+    || {{isNull (_x#0)} count fullCrew [_x, "", true] > _x emptyPositions "driver"}
 }
--- a/addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
+++ b/addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
@@ -5,22 +5,29 @@
  *
  * Arguments:
  * 0: Unit <OBJECT>
- * 1: Distance <NUMBER>
+ * 1: Distance <NUMBER><OPTIONAL>
+ * 2: Restriceted to cargo only <BOOL><OPTIONAL>
  *
  * Return Value:
  * Nearest vehicles with a free seat <ARRAY>
  *
  * Example:
- * [bob] call ace_common_fnc_nearestVehiclesFreeSeat
+ * [cursorObject] call ace_common_fnc_nearestVehiclesFreeSeat
  *
  * Public: Yes
  */
 
-params ["_unit", ["_distance", 10]];
+params ["_unit", ["_distance", 10], ["_cargoOnly", false]];
 
 private _nearVehicles = nearestObjects [_unit, ["Car", "Air", "Tank", "Ship_F", "Pod_Heli_Transport_04_crewed_base_F"], _distance];
 _nearVehicles select {
     // Filter cargo seats that will eject unconscious units (e.g. quad bike)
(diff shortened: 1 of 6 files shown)


jonpas (approved):

veteran29 (approved):

mharis001 (approved):

PabstMirror:
Not working with MBT_04

loadPersonLocal line 24 is wrong (I think this is my fault from an old PR), it should be:
```
if ((_vehicle emptyPositions "cargo" > 0) && {!(_unit getVariable ['ACE_isUnconscious', false]) || {(getNumber (configFile >> "CfgVehicles" >> (typeOf _vehicle) >> "ejectDeadCargo")) == 0}}) then {
```
this was causing cargo loading path to always happen based on ejectDeadCargo, ignoring if there was room

addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
```diff
@@ -22,5 +22,5 @@ private _nearVehicles = nearestObjects [_unit, ["Car", "Air", "Tank", "Ship_F",
 _nearVehicles select {
     // Filter cargo seats that will eject unconscious units (e.g. quad bike)
     ((_x emptyPositions "cargo" > 0) && {!(_unit getVariable ['ACE_isUnconscious', false])} || {(getNumber (configFile >> "CfgVehicles" >> (typeOf _x) >> "ejectDeadCargo")) == 0})
-    || {_x emptyPositions "gunner" > 0}
+    || {{isNull (_x#0)} count fullCrew [_x, "", true] > _x emptyPositions "driver"}
```
PabstMirror, line 25: this has same error as above (also line 24)
But after fixing it still won't be right because line 25 will ignore the ejectDeadCargo check

PabstMirror: I think something like this should work
```
_nearVehicles select {
    private _canSitInCargo = (!(_unit getVariable ['ACE_isUnconscious', false])) || {(getNumber (configFile >> "CfgVehicles" >> (typeOf _x) >> "ejectDeadCargo")) == 0};
    ((fullCrew [_x, "", true]) findIf {
        _x params ["_body", "_role", "_cargoIndex"];
        (isNull _body) && {_role != "DRIVER"} && {_canSitInCargo || {_cargoIndex == -1}}
    }) > -1
}
```

PabstMirror (approved):

PR #7984, opened 2020-11-04
Cargo - Add BI Vehicle in vehicle support

Vdauphin:
- This PR choose when possible to use the BI Vehicle in Vehicle feature (ViV)
- When loading an item, the function check if it is possible to load it with ViV
- When unloading an iten, the function check if the item has been loaded with ViV and spawn it accordingly 
- [ ] Need to handle the case where player unload a ViV item without ACE Cargo. ie: As driver, you can unload all ViV item

### DIFF
--- a/addons/cargo/functions/fnc_loadItem.sqf
+++ b/addons/cargo/functions/fnc_loadItem.sqf
@@ -35,11 +35,13 @@ _vehicle setVariable [QGVAR(space), _space - _itemSize, true];
 
 if (_item isEqualType objNull) then {
     detach _item;
-    _item attachTo [_vehicle,[0,0,-100]];
-    [QEGVAR(common,hideObjectGlobal), [_item, true]] call CBA_fnc_serverEvent;
+    if !(_vehicle setVehicleCargo _item) then {
+        _item attachTo [_vehicle,[0,0,-100]];
+        [QEGVAR(common,hideObjectGlobal), [_item, true]] call CBA_fnc_serverEvent;
 
-    // Some objects below water will take damage over time and eventualy become "water logged" and unfixable (because of negative z attach)
-    [_item, "blockDamage", "ACE_cargo", true] call EFUNC(common,statusEffect_set);
+        // Some objects below water will take damage over time and eventualy become "water logged" and unfixable (because of negative z attach)
+        [_item, "blockDamage", "ACE_cargo", true] call EFUNC(common,statusEffect_set);
+    };
 };
 
 true
--- a/addons/cargo/functions/fnc_unloadItem.sqf
+++ b/addons/cargo/functions/fnc_unloadItem.sqf
@@ -48,10 +48,15 @@ private _itemSize = [_item] call FUNC(getSizeItem);
 _vehicle setVariable [QGVAR(space), (_space + _itemSize), true];
 
 if (_item isEqualType objNull) then {
-    detach _item;
-    // hideObjectGlobal must be executed before setPos to ensure light objects are rendered correctly
-    // do both on server to ensure they are executed in the correct order
-    [QGVAR(serverUnload), [_item, _emptyPosAGL]] call CBA_fnc_serverEvent;
+    if (isVehicleCargo _item) then {
+        objNull setVehicleCargo _item;
+        _item setPosASL (AGLtoASL _emptyPosAGL);
+    } else {
+        detach _item;
+        // hideObjectGlobal must be executed before setPos to ensure light objects are rendered correctly
+        // do both on server to ensure they are executed in the correct order
+        [QGVAR(serverUnload), [_item, _emptyPosAGL]] call CBA_fnc_serverEvent;
+    };
 } else {
     private _newItem = createVehicle [_item, _emptyPosAGL, [], 0, "NONE"];
     _newItem setPosASL (AGLtoASL _emptyPosAGL);


veteran29:
> Need to handle the case where player unload a ViV item without ACE Cargo. ie: As driver, you can unload all ViV item

we could use engine events for ViV stuff.  🙏 @dedmen

Vdauphin:
I will go for a `[{isNull isVehicleCargo _this}, ace_cargo_fnc_unloadItem] call CBA_fnc_waitUntilAndExecute` I guess

Drofseh:
I like this conceptually but play mostly with ww2 mods where the modern style of cargo pallet will not be a great look.
Could you include a setting to define a class for the pallet?

Vdauphin:
Do you have a class name for the WW2 pallet?

Drofseh:
> Do you have a class name for the WW2 pallet?

I'll have a look for something I think is good, but not everyone will run the same ww2 mods as my unit. 
Some people only run IFA3 or only FOW. That's why I would like the setting to accept a class name.

Drofseh:
`Land_WoodenBox_02_F` would fit, it's a Livonia object too so not mod dependant.

addons/cargo/functions/fnc_canShowUnloadAllVehicles.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Public: No
+ */
+
+params [
+    "_vehicle",
+    ["_unit", player, [objNull]]
+];
+
+driver _vehicle isEqualTo _unit &&
+{isClass (configfile >> "CfgVehicles" >> typeOf _vehicle >> "VehicleTransport" >> "Carrier")} &&
```
veteran29, line 25:
```suggestion
{isClass (configOf _vehicle >> "VehicleTransport" >> "Carrier")} &&
```

addons/cargo/functions/fnc_unloadAllVehicles.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Example:
+ * [vehicle player] call ace_cargo_fnc_unloadAllVehicles
+ *
+ * Public: No
+ */
+
+params ["_vehicle"];
+
+private _loadedVehicles = getVehicleCargo _vehicle - (_vehicle getVariable [QGVAR(loaded), []]);
```
veteran29, line 20: `getVehicleCargo _vehicle - (_vehicle getVariable [QGVAR(loaded), []]);` is used in 2 places already, maybe move it to a function? `fnc_getVehicleCargo` to keep naming of the command that it's replacing basically.

addons/cargo/functions/fnc_unloadAllVehicles.sqf
```diff
@@ -0,0 +1,26 @@
...
+private _unloadingInterval = getNumber (configfile >> "CfgVehicles" >> typeOf _vehicle >> "VehicleTransport" >> "Carrier" >> "unloadingInterval");
```
veteran29, line 21:
```suggestion
private _unloadingInterval = getNumber (configOf _vehicle >> "VehicleTransport" >> "Carrier" >> "unloadingInterval");
```

dedmen:
> > Need to handle the case where player unload a ViV item without ACE Cargo. ie: As driver, you can unload all ViV item
> 
> we could use engine events for ViV stuff. pray @dedmen

if you make FT ticket now, and DM me on discord I may be able to get it into 2.02.
I'll be gone for a while now and won't read github :u

veteran29:
> if you make FT ticket now, and DM me on discord I may be able to get it into 2.02.
> I'll be gone for a while now and won't read github :u @veteran29

https://feedback.bistudio.com/T155799

Vdauphin:
This PR don't need ViV events

dedmen:
Did you check unloading in aircraft in air?
As ViV does its own parachute stuff, and we'd still want to keep the ACE attached smoke on the drop.
Don't know if vanilla has a smoke too?

And if vanilla spawns parachute, then ACE doing it too might interfere.


Ah I see you didn't touch the paradropItem function at all? Does detach work properly with ViV? Doesn't it just detach inside the aircraft and make it explode in the air?

Vdauphin:
Before this PR, the ViV allow loading vehicle visually (https://community.bistudio.com/wiki/Arma_3:_Vehicle_in_Vehicle_Transport).
Before this PR, ACE cargo system load item and hide them under the ground by using `attachTo`

With this PR, the ViV allow loading vehicle visually (https://community.bistudio.com/wiki/Arma_3:_Vehicle_in_Vehicle_Transport). When you unload this type of vehicle (ie: previously loaded by using normal vanilla scroll menu) all vanilla behavior still apply. I don't change anything here. I just overwrite the vanilla scroll menu `Unload all vehicles` to a ACE interaction version. This interaction unload only vehicles not loaded inside ACE cargo (https://github.com/acemod/ACE3/pull/7984/files#diff-cb230982fd4f4b1a16f435906c6ea7876b85e5c5860a2c818d6f1a119ba49b4cR4).
With this PR, ACE cargo system load item and if possible add them visually to the cargo of the vehicle by using `setVehicleCargo` but if this command fail I still hide them under the ground by using `attachTo`. ACE cargo unloading is still not affected here.

dedmen:
> When you unload this type of vehicle (ie: previously loaded by using normal vanilla scroll menu) all vanilla behavior still apply.

Yes thats what I mean, I don't know if vanilla air unloading also attaches chemlight and smoke like ACE does.

> This interaction unload only vehicles not loaded inside ACE cargo

Isn't that confusing? you select "unload all" and it only unloads half of the things 🤔 

So load vehicle in via ace cargo (uses ViV internally), and then paradrop it in air it will properly use normal ACE Paradrop and not the Vanilla ViV paradrop? That was my question

dedmen:
Here this is the exact same code, even including comments in 
`fnc_unloadItem.sqf` and `fnc_paradropItem.sqf`
But you only updated one of the two.

https://github.com/acemod/ACE3/pull/7984/files#diff-070cc5773079749210aa2488a49801c093617f64895c9dfdf5b20c135a8eab07L52

https://github.com/acemod/ACE3/blob/master/addons/cargo/functions/fnc_paradropItem.sqf#L44

I wonder if thats really intended and if paradrop really works?
I imagine you adjust the unload version because you need to adjust it, why do you not need to adjust the paradrop one?

Vdauphin:
> Isn't that confusing? you select "unload all" and it only unloads half of the things thinking

It is as before. The current behavior is when you "unload all vehicles" you don't unload ACE cargo objects  :thinking: If you want to unload object in ACE cargo you use the ACE interaction menu. Also it is pretty clear because the scroll wheel say: "unload all VEHICLES" as you can't load vehicle with ACE cargo the player know what this interaction will unload, ie: vehicles and not objects as objects can only be loaded with ACE cargo

> So load vehicle in via ace cargo (uses ViV internally),

No, you can't load vehicle with ACE cargo, only objects.

> Here this is the exact same code, even including comments in
> fnc_unloadItem.sqf and fnc_paradropItem.sqf
> But you only updated one of the two.

That is correct... I throught the paradrop was just a call of the `fnc_unloadItem.sqf` and an event handling the paradrop

> I wonder if thats really intended and if paradrop really works?

it looks like it is working

> I imagine you adjust the unload version because you need to adjust it, why do you not need to adjust the paradrop one?

Because I throught the unloading was handle only by the `fnc_unloadItem.sqf` and not also by the `fnc_paradropItem.sqf` :cry: 
I will adjust it. 

Thank you for the input and sorry for the confusion!

dedmen:
> as you can't load vehicle with ACE cargo

You can if the vehicle is configured to have cargo size. For example I load Quads and Bicycles into ACE Cargo quite often when I do Logistics.

>  as objects can only be loaded with ACE cargo

Huh? but your last screenshot shows a ammobox being loaded in ViV?
In fact all of your screenshots show non-vehicle Objects being loaded in ViV.


> Also it is pretty clear because the scroll wheel say: "unload all VEHICLES" as you can't load vehicle with ACE cargo the player know what this interaction will unload,

Yes but.. It says "unload all VEHICLES" even when you also have non-vehicle objects (wheels, ammoboxes like shown in your screenshots) loaded in ViV. So thats not clear at all if ViV "unload all VEHICLES" also includes non-vehicle objects??

Vdauphin:
> You can if the vehicle is configured to have cargo size.

You can do what we want in A3. We just need to be consistant

> So yes you can definitely load vehicles with ACE Cargo.

Sure but the intended behavior is:
- ACE cargo to load items/objects by using ACE interaction
- ViV load only vehicles, you drive the vehicle and use the scroll wheel
It is up to mission maker to change this behavior and change interaction name accordingly.

>   Huh? but your last screenshot shows a ammobox being loaded in ViV?
> In fact all of your screenshots show non-vehicle Objects being loaded in ViV.

Yes but there are loaded thought ACE cargo interaction not with vanilla scroll menu.

> So thats not clear at all if ViV "unload all VEHICLES" also includes non-vehicle objects??

No because it will unload only vehicles and not items/objects... Also the interaction is not show if only item/objects are loaded (https://github.com/acemod/ACE3/pull/7984/files#diff-9182bb8a0870e1121805b2943f97e5e04eec113876a41b3bd8877ab5f88213f0R4)
To unload items/objects the player open the ACE cargo and use the corresponding list.

dedmen:
> not with vanilla scroll menu.

Everything I'm talking about is exclusively ACE Cargo Loading/Unloading.
Not at all talking about vanilla ViV vehicle loading.

> To unload items/objects the player open the ACE cargo and use the corresponding list.

Yes but according to your screenshots, these objects are stored in ViV slot (instead of being attachTo'ed underground) and you have to get them outta there on unload

Vdauphin:
> Yes but according to your screenshots, these objects are stored in ViV slot (instead of being attachTo'ed underground) and you have to get them outta there on unload.

As you can't load item object with scroll menu, the only way to load them is by using ACE cargo. Then when the player will want to unload it, he will use the same path ie: ACE cargo interaction etc. If you use only ACE cargo then the interaction  "unload all VEHICLES" will never be shown, so no confusion here.

> You load ammobox via ACE Cargo, it appears inside ViV slot (because thats the feature you are adding, right?)
> Now what happens when you unload that ammobox, in-flight.

it happen this:




> You need to get it out of ViV, problem is, ViV has its own paradrop/unload in air implementation.

I use `detach` which is not triggering this paradrop implementation. Or I could check if a parachute is already attach and don't add the ACE one. As you want


> And I don't know if the ViV unload command (which you have to use to get the box out of ViV) immediately triggers vanilla parachute (according to BIF posts it does),

Yes it does.

>  which means it wouldn't have the ACE parachute and ACE Chemlight and ACE Smokegrenade.
Thats my concern.

No, it would have A3 parachute and ACE parachute and ACE Chemlight and ACE Smokegrenade without the last commit

dedmen:
> As you can't load item object with scroll menu, the only way to load them is by using ACE cargo.

So am I missunderstanding this ticket and loading a Ammobox via ACE Cargo doesn't make it visibly appear in ViV cargo?
I thought thats all this ticket is about

veteran29:
> So am I missunderstanding this ticket and loading a Ammobox via ACE Cargo doesn't make it visibly appear in ViV cargo?
I thought thats all this ticket is about

This PR makes ACE Cargo use ViV cargo space if available, nothing else. ACE Cargo functionality is separate from engine ViV. Only cargo space is shared.

Please do not use this place for discussions.

dedmen:
> This PR makes ACE Cargo use ViV cargo space if available, nothing else. 

Yes, and unloading a ViV cargo vehicle via engine command might trigger engine paradrop handling which will conflict with ours.
And seeing paradrop code wasn't modified, even though unloading code (which is the exact same block of code) was modified I sensed there might be something wrong.
Which is why I asked to make sure.

ZluskeN:
Is this blocked by something? I was considering making something like this but it seems the work has already been done.

BrettMayson:
@Vdauphin I'm good to review and test? Nothing else to be done here?

Vdauphin:
Yes sure but it definitely need some testing in "real" condition to find some edge cases

addons/cargo/functions/fnc_loadItem.sqf
```diff
@@ -35,11 +35,37 @@ _vehicle setVariable [QGVAR(space), _space - _itemSize, true];
 
 if (_item isEqualType objNull) then {
     detach _item;
-    _item attachTo [_vehicle,[0,0,-100]];
-    [QEGVAR(common,hideObjectGlobal), [_item, true]] call CBA_fnc_serverEvent;
+    if !(_vehicle setVehicleCargo _item) then {
+        private _itemsCargo = _loaded arrayIntersect getVehicleCargo _vehicle;
```
ZluskeN, line 39: Ideally, this _itemsCargo array should be sorted by the objects' surface area. Something like: 
`private _itemsCargo = [_loaded arrayIntersect getVehicleCargo _vehicle, [], {_x call FUNC(getSurfaceItem)}, "DESCEND"] call BIS_fnc_sortBy;` (Untested)
That should make things more predictable and maximize the number of objects that stay in ViV cargo.

Vdauphin: I didn't took a deep look to your suggestion but as `getVehicleCargo` :
> The returned array is in the order the cargo vehicles were added to containing vehicle.

So first loaded items will be stored as cargo net as you can see in PR description under the concept section. This make sure to have all cargo nets properly stacked
If we are starting to store items by taking them not in this order, cargo net will start to appear randomly in the ViV. At the end, you may see cargo nets with space between them (which is not "nice" but acceptable?)

I will take a look to see how this behave

ZluskeN: I did some testing and I didn't realize the other VIV items stay in place when the box gets added. 
In that case it is probably best to remove them in the order they were loaded like you sad, I withdraw my suggestion.

It might be a good idea to look into making some kind of dynamic rearrangement but that should probably be done in another PR.

dedmen:
Seems we are still missing: https://github.com/acemod/ACE3/pull/7984#issuecomment-721783599
The engine events were added 3 years ago. But they are not used in this code.

dedmen:
> Seems we are still missing: https://github.com/acemod/ACE3/pull/7984#issuecomment-721783599

But since (I completely missed this change back then)
> No because it will unload only vehicles and not items/objects... Also the interaction is not show if only item/objects are loaded 

And https://github.com/acemod/ACE3/pull/7984/files#diff-19cb528ab9cffb1ff448fcf3c1125e9fb154cfa1c9b49ca23cc7b27ac643f3ffR4
We don't need it anymore, because there is no concern about them being unloaded through non-ace means?


But there is still atleast one unhandled way to unload a box.

You can ACE-Carry pull it off the back of the truck.
IMO that's amazing and immersive, but is that handled? Skimming through the code I couldn't find anything about it.


Okey I tested it
We display cargo items in UI by GVAR(loaded), when you pull a box off a truck with ACE Carry, the loaded variable is not updated.
The item shows in cargo, even though its not on the vehicle.

Repro:
Load 3 medical boxes onto hemtt flatbed
Check cargo menu of truck, shows 3 boxes and a wheel.
Walk close to the side of hemtt, grab box with ACE carrying off of the back of the truck.
(Diag binary throws an assert, the truck doesn't like that a vehicle in its cargo, is not attached to it. This is not really a problem, Arma will fix itself up next frame and detect that something was detached. I'm not sure if manually detaching would help here, or whether unloading with setVehicleCargo might cause issues)
Player is now carrying box (its way too high up, but can just be scrolled down normally)
Drop box onto ground.
Check hemtt cargo menu, shows 3 boxes and a wheel. But only 2 boxes are on it.
Unload the other boxes.
Truckbed is empty, but still shows 3 boxes in cargo.
Open cargo menu, and unload box from there (A box that isn't actually there anymore)
The 10s timer runs and, it says you successfully unloaded the box. But there is nothing in your hand. It now disappeared from the cargo menu.

IMO we should keep this carrying feature and fix it up instead.


```
cursorObject addEventHandler ["CargoUnloaded", {
	params ["_parentVehicle", "_cargoVehicle"];
hint str _this;
systemChat str _this;
}];
```

This does NOT detect the unload. Because Carrying doesn't detach.
It does detect the ACE Cargo unload.
And it does NOT detect when you `detach cursorObject` the box from the truck.
But `objNull setVehicleCargo cursorObject` works.

IMO the detach is a bug. The vehicle will notice that the object isn't attached in the next simulation cycle and remove it from the internal list. The Eventhandler should fire there too when it notices it missing. I will fix it for 2.18.
But maybe we could handle this better if we had a general detach eventhandler?


The easy fix would be to use this eventhandler, and inside carrying, check if the thing you are picking up, is currently attached to anything, and if yes then detach it from that first.

johnb432 (changes requested):
Ok, I've reviewed the PR in it's current form and before I merged master into it recently.

The cargo net idea is neat, but frankly it's the root cause of several headaches:
- If the cargo net is unloaded via ViV (but not ACE Cargo), *something* should happen: Either it's replaced and the unloaded cargo net is deleted (making it impossible to unload cargo nets essentially) or the space left behind should be filled up by some of the contents of said cargo net. Currently nothing happens.
- Generally, when you unload something, the space left behind is not filled by items inside ACE cargo, which imo it should replenish - I can address that though. I'm a bit hesitant about this suggestion though, since you might want to load something ViV that isn't ACE cargo and if the thing replenishes automatically, you'll need to unload all of the stored ACE cargo (and some ViV cargo) before being able to store your ViV item.
- I'm trying to use the ViV EH, but given that ACE very dynamically loads & unloads stuff from ViV, I'm not sure how to prevent those EH from triggering when ACE is doing the loading/unloading vs a 3rd party.
- Bug: Once you have filled up all ViV spaces (including using the cargo nets), adding another cargo item will make the "loose" cargo items (the ones not in the form of a cargo net) go into ACE cargo, where they can't be seen.
E.g.: Load numerous wooden boxes into a flatbed HEMTT until you have 3 cargo nets (you might have to increase the cargo space of the vehicle). Load several more wooden boxes - the first ones will stay, but after a certain threshold the wooden boxes will disappear, being moved to ACE cargo. You can start loading more boxes and the cycle begins again.

If we get rid of the cargo nets, it would make things easier, as the only time something can be unloaded is when something is purposely unloaded and not put into ACE cargo immediately. We could use the `CargoUnloaded` EH and check is something is in the loaded array of a vehicle - if it is, remove it from the array.

It would also remove the cargo-net-type setting. I really don't like it being a setting, imo it should be config, but even better is if we don't need it at all.

dedmen:
New entity events in 2.18
Attached [attachedObj, parentObj, isReattach bool]
Detached [attachedObj, parentObj]

Detached also fires on ViV cargo unload.
We should be pushing a dev-branch next week, if you want it early for testing I can send it on discord.

> which imo it should replenish

Considering how many headaches it causes. Also there might be uses where you don't want it.

Say you take a box from the front out, to interact with it. And then you put it back in. It can go into the same spot as it was before.
If it replenishes, and some tiny object like a jerry can, takes the space where previously the big box sat at. And now the box doesn't fit onto ViV space at all and must go into hidden cargo.. Meh.

I think I would have more cases where I don't want to replenish.
Could make a button in Cargo UI to do it manually, but UI edits are more work.

johnb432:
> New entity events in 2.18 Attached [attachedObj, parentObj, isReattach bool] Detached [attachedObj, parentObj]
> 
> Detached also fires on ViV cargo unload. And could also handle the ACE carry issue. (tho it would throw asserts in diag and I don't like that :D) We should be pushing a dev-branch next week, if you want it early for testing I can send it on discord.

I believe I've resolved the issue in https://github.com/acemod/ACE3/pull/7984/commits/623e77dbefbd75c00d1939b142b6853ac8c1afed by just unloading from ViV if they are loaded in ViV.
Currently I haven't ViV unload EH to handle ViV unloading of such manner, but I'm waiting for feedback on what I've written above.

> Considering how many headaches it causes. Also there might be cases where you don't want it.
> 
> Say you take a box from the front out, to interact with it. And then you put it back in. It can go into the same spot as it was before. If it replenishes, and some tiny object like a jerry can, takes the space where previously the big box sat at. And now the box doesn't fit onto ViV space at all and must go into hidden cargo.. Meh.
> 
> I think I would have more situations where I don't want to replenish. Could make a button in Cargo UI to do it manually, but UI edits are more work. Can already do it manually anyways, just need to unload and re-load.

Yea, I won't add it then.

johnb432:
The more I think about it, the more I think this is whole PR is a bad idea as it's implemented currently. Please don't get me wrong, I really like the visual effect this adds, but I don't think it's done right.

> One more thing: Do we want to have the "unload all vehicles" action unload all ViV, including ACE cargo or do we exclude ACE cargo from that (PR has the latter implemented already)?

At the moment, if you load something, you have no control over if it goes into ViV or ACE cargo space directly. As the PR is currently, if you use the "Unload all vehicles" action, it will unload all ViV items except items that are part of ACE cargo, although for some reason it unloads the cargo nets (at least according to the code - I'm guessing it's a bug, if not, unexpected behaviour). As dedmen pointed out before:

>> This interaction unload only vehicles not loaded inside ACE cargo
>
> Isn't that confusing? you select "unload all" and it only unloads half of the things 🤔

I agree: It's confusing and inconsistent, especially seeing as at the moment you have no control over anything. The "unload all vehicles" action should truly unload all ViV items imo.

Solution to that would be to add a button inside the cargo menu that would allow items to be moved from ViV to ACE cargo space and vice versa. This would allow one to set specific items to be ACE cargo and others to be ViV.
Unfortunately I can imagine there being some hurdles to overcome, such as the `CargoUnloaded` EH firing and thinking the item was unloaded and therefore removing it from the list, whereas in reality it was just moved. I guess we can use some flags to deal with that, but honestly I'd prefer not having to.
Furthermore, if you want to use ACE cargo only (say you want to keep the ViV space free in case of an emergency pickup of whatever), you would have to move every single item you load into ViV into the ACE cargo space. Depending on how many items you have loaded, this is tedious work.
Finally, I have mixed feelings about mixing ViV and the ACE cargo mechanics, which work in different ways.

So: Why not keep both mechanics separate instead?

The idea I have in mind is to have two loading options for each item: one for regular ACE cargo and one for ViV if it exists. The ViV items added to vehicles wouldn't be stored as ACE cargo and would only use regular ViV mechanics, it would be completely independent of the ACE cargo system. This means that the "Unload all vehicles" action would remain as it is in Vanilla and truly unload all ViV objects.
To unload ViV items, either adding a new interaction with the same style of menu as the cargo menu or adding a button in the cargo menu to switch to ViV items and back are the ideas I had in mind (first suggestion is probably easier to do, but not as elegant as the second). The nice thing is that you could probably use the "unload" and "deploy" buttons as they are now (unload would let the game engine handle unloading, deploy would use the existing ACE cargo deployment system). I imagine ACE cargo's paradrop mechanic would also be reused if necessary.

Going with this suggestion would also avoid the whole cargo nets system.

addons/cargo/functions/fnc_paradropItem.sqf
```diff
@@ -39,64 +41,51 @@ if (_itemSize < 0) exitWith {
...
+    ) then {
+        private _velocity = velocity _object;
+        private _parachute = createVehicle ["B_Parachute_02_F", [0, 0, 0], [], 0, "CAN_COLLIDE"];
 
-    // Cannot use setPos on parachutes without them closing down
-    _parachute attachTo [_object, [0, 0, 0]];
-    detach _parachute;
+        // Prevent collision damage
+        [QEGVAR(common,fixCollision), _parachute] call CBA_fnc_localEvent;
+        [QEGVAR(common,fixCollision), _object] call CBA_fnc_localEvent;
```
johnb432, line 71:
```suggestion
        [QEGVAR(common,fixCollision), _object, _object] call CBA_fnc_targetEvent;
```

johnb432:
> So: Why not keep both mechanics separate instead?

I've come up with a first draft of this idea, see https://github.com/acemod/ACE3/tree/cargo-add-viv/addons/cargo.

- ACE cargo spaces and ViV cargo spaces are fully separate.
- To access the ViV cargo space, you go through the into the ACE cargo menu and press the bottom left button to switch from one cargo space to another.
- You can deploy and unload ViV items. ViV cargo that isn't ACE cargo compatible can't be unloaded via this menu (e.g vehicles).
- Missing: Handling of paradropping ViV cargo items, but that's easy enough to do. I just don't want to spent more time atm getting this thing fully functional, only to find out that nobody wants it.

Thing to sort out if people want this version:
Atm, the ViV system uses some ACE cargo config attributes, such as size and `canLoad` (size is used to determine if an item is not loadable (`-1` would be not loadable)). However, there are Vanilla attributes (https://community.bistudio.com/wiki/Arma_3:_Vehicle_in_Vehicle_Transport) such as `canBeTransported` that exist.
Do we fully want to separate ViV cargo from ACE cargo (could make it a separate component at that point) and make interactions follow Vanilla rules only or do we make/keep a somewhat hybrid system?

PR #7998, opened 2020-11-09
Medical Statemachine - Ensure bleedout condition string is code

PabstMirror:
Fix
```
 Error Params: Type Bool, expected String,code
 File x\cba\addons\statemachine\fnc_addEventTransition.sqf, line 38
```
tested with filepatching and with pboProject


jonpas (approved):

BaerMitUmlaut:
Hmm this should maybe be fixed in CBA as well.

PR #8000, opened 2020-11-10
Fix cargo logic, adjust function for captives

PabstMirror:
for pr #7980
captive used same logic but different loading logic that is hardcoded to reject non-cargo slots, so added optional param

### DIFF
--- a/addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
+++ b/addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
@@ -5,22 +5,29 @@
  *
  * Arguments:
  * 0: Unit <OBJECT>
- * 1: Distance <NUMBER>
+ * 1: Distance <NUMBER><OPTIONAL>
+ * 2: Restriceted to cargo only <BOOL><OPTIONAL>
  *
  * Return Value:
  * Nearest vehicles with a free seat <ARRAY>
  *
  * Example:
- * [bob] call ace_common_fnc_nearestVehiclesFreeSeat
+ * [cursorObject] call ace_common_fnc_nearestVehiclesFreeSeat
  *
  * Public: Yes
  */
 
-params ["_unit", ["_distance", 10]];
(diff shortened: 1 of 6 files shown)


addons/common/functions/fnc_nearestVehiclesFreeSeat.sqf
```diff
@@ -5,22 +5,29 @@
...
-params ["_unit", ["_distance", 10]];
+params ["_unit", ["_distance", 10], ["_cargoOnly", false]];
 
 private _nearVehicles = nearestObjects [_unit, ["Car", "Air", "Tank", "Ship_F", "Pod_Heli_Transport_04_crewed_base_F"], _distance];
 _nearVehicles select {
     // Filter cargo seats that will eject unconscious units (e.g. quad bike)
-    ((_x emptyPositions "cargo" > 0) && {!(_unit getVariable ['ACE_isUnconscious', false])} || {(getNumber (configFile >> "CfgVehicles" >> (typeOf _x) >> "ejectDeadCargo")) == 0})
-    || {{isNull (_x#0)} count fullCrew [_x, "", true] > _x emptyPositions "driver"}
+    private _canSitInCargo = (!(_unit getVariable ['ACE_isUnconscious', false])) || {(getNumber (configFile >> "CfgVehicles" >> (typeOf _x) >> "ejectDeadCargo")) == 0};
+    ((fullCrew [_x, "", true]) findIf {
```
BaerMitUmlaut, line 26: How about`fullCrew [_x, ["", "cargo"] select _cargoOnly, true]`? Would reduce the entries you have to iterate over.

BaerMitUmlaut: Nevermind, that would not find any seats in the Polaris DAGOR for example.

BaerMitUmlaut (commented):
Code looks good to me, would you like me to test this or do you think we're good?

PR #8013, opened 2020-11-22
Medical Treatment - Fix fnc_addToInventory call

Dystopian:
- title.

There is no error before this PR as well (`true` goes to `default` in `switch`). Not tested.


PabstMirror (approved):

PabstMirror:
just for ref, 3rd arg is 
`2: Container (uniform, vest, backpack) (default: "") <STRING>`

mharis001 (approved):

PR #8016, opened 2020-11-26
Grenades - Use fnc_addToInventory

Cyruz143:
**Make use of fnc_addToInventory:**
- When you convert an explosive to a grenade, it'll try and go back in to the same container to stop it magically jumping from backpack to vest etc.

If there is a better way to get the original container rather than abusing `splitString` let me know

### DIFF
--- a/addons/grenades/functions/fnc_addChangeFuseItemContextMenuOptions.sqf
+++ b/addons/grenades/functions/fnc_addChangeFuseItemContextMenuOptions.sqf
@@ -34,8 +34,8 @@
         ],
         {
             params ["_unit", "", "", "_slot", "_magArr"];
-            private _containerStr = toLower ((_slot splitString "_") select 0);
-            [_unit, (_magArr select 1), _containerStr, -1] call ace_common_fnc_addToInventory;
+            private _containerStr = toLower (_slot splitString "_" select 0);
+            [_unit, _magArr select 1, _containerStr] call EFUNC(common,addToInventory);
             false;
         },
         true,
--- a/addons/grenades/functions/fnc_addChangeFuseItemContextMenuOptions.sqf
+++ b/addons/grenades/functions/fnc_addChangeFuseItemContextMenuOptions.sqf
@@ -57,8 +57,8 @@
         ],
         {
             params ["_unit", "", "", "_slot", "_magArr"];
-            private _containerStr = toLower ((_slot splitString "_") select 0);
-            [_unit, (_magArr select 0), _containerStr, -1] call ace_common_fnc_addToInventory;
+            private _containerStr = toLower (_slot splitString "_" select 0);
+            [_unit, _magArr select 0, _containerStr] call EFUNC(common,addToInventory);
             false;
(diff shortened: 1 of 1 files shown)


addons/grenades/functions/fnc_addChangeFuseItemContextMenuOptions.sqf
```diff
@@ -33,8 +33,9 @@
             }
         ],
         {
-            params ["_unit", "", "", "", "_magArr"];
-            _unit addMagazine (_magArr select 1);
+            params ["_unit", "", "", "_slot", "_magArr"];
+            private _containerStr = toLower ((_slot splitString "_") select 0);
+            [_unit, (_magArr select 1), _containerStr, -1] call ace_common_fnc_addToInventory;
```
Dystopian, line 38:
```suggestion
            private _containerStr = toLower (_slot splitString "_" select 0);
            [_unit, _magArr select 1, _containerStr] call EFUNC(common,addToInventory);
```

addons/grenades/functions/fnc_addChangeFuseItemContextMenuOptions.sqf
```diff
@@ -55,8 +56,9 @@
             }
         ],
         {
-            params ["_unit", "", "", "", "_magArr"];
-            _unit addMagazine (_magArr select 0);
+            params ["_unit", "", "", "_slot", "_magArr"];
+            private _containerStr = toLower ((_slot splitString "_") select 0);
+            [_unit, (_magArr select 0), _containerStr, -1] call ace_common_fnc_addToInventory;
```
Dystopian, line 61:
```suggestion
            private _containerStr = toLower (_slot splitString "_" select 0);
            [_unit, _magArr select 0, _containerStr] call EFUNC(common,addToInventory);
```

Dystopian (changes requested):

addons/grenades/functions/fnc_addChangeFuseItemContextMenuOptions.sqf
```diff
@@ -33,8 +33,9 @@
             }
         ],
         {
-            params ["_unit", "", "", "", "_magArr"];
-            _unit addMagazine (_magArr select 1);
+            params ["_unit", "", "", "_slot", "_magArr"];
+            private _containerStr = toLower (_slot splitString "_" select 0);
```
commy2, line 37: What if the name already contains a _ ?

Cyruz143: It can only return these though can't it? https://github.com/CBATeam/CBA_A3/blob/master/addons/ui/fnc_addItemContextMenuOption.sqf#L20-L75

Cyruz143: Less terrible method implemented

commy2 (approved):

veteran29 (approved):

PR #8017, opened 2020-11-26
Tagging - Remove useless isClass check in configClasses

dedmen:
Why? Because engine already checks
firefox_2020-11-26_11-23-41

Also engine will get optimization for when condition == "true", so we'll benefit from that here too when it arrives


mharis001 (approved):

commy2:
First one would've made sense if it were lazy eval.

commy2 (approved):

PR #8023, opened 2020-11-29
Cargo - Add ability to add custom name to cargo objects

JasperRab:
Ever had more than 5 of the same ammo boxes in a vehicle? And you need to find that one with your favorite banana?
Well with this you can change the name of the boxes that are loadable into the cargo system.

- Add the ability to edit cargo objects (crates/boxes only) names. Either via in-game action on the object it self or via 3den attributes.
- I've added a single function which spawns the UI to edit the name and also edits it. This is an scripted UI, it could be changed to a config version if wanted, but as I'm not good in UI coding I left it on this simple version mainly for testing, and to test the idea.
- I've added 5 new stringtable entries, 3 for the action/UI and 2 for 3den attributes. _These require extra translations._ 
- I've changed the lines that got the displayName from config to a variant that checks for a object variable, if that is not present it defaults back to the config displayName.
- I've added 3den attribute
Note:
For defaultValue I tried using:
`defaultValue = QUOTE(GET_STRING(configFile >> 'CfgVehicles' >> typeOf _this >> 'displayName',''));`
However the game did not like it as it kept creating this from it:
`defaultValue = "GET_STRING(configFile >> 'CfgVehicles' >> typeOf _this >> 'displayName''')";`
So if someone could tell me why that happens, thanks...
- I've added GVAR(objectActions) and removed the old GVAR(objectAction) as I needed to add 2. However I still used select to add them to class/object as it's faster when there are only two. But if required I can move it to an forEach.

**Pictures:**
image
Cargo menu with custom names and default (some) names.

image
3den attribute where the custom name can be set by mission makers.

image
UI of the renaming interface, it's basic but does the job. I can have a look at remaking it (in config) if wanted.

### DIFF
--- a/addons/cargo/CfgEden.hpp
+++ b/addons/cargo/CfgEden.hpp
@@ -3,6 +3,18 @@ class Cfg3DEN {
         class AttributeCategories {
             class ace_attributes {
                 class Attributes {
+                    class GVAR(customName) {
+                        displayName = CSTRING(customName_edenName);
+                        tooltip = CSTRING(customName_edenDesc);
+                        property = QGVAR(customName);
+                        control = "Edit";
+
+                        expression = QUOTE(_this setVariable [ARR_3(QQGVAR(customName), _value, true)];);
+                        defaultValue = QUOTE(getText(configFile >> 'CfgVehicles' >> typeOf _this >> 'displayName'));
+
+                        condition = "objectHasInventoryCargo-objectVehicle";
+                        typeName = "STRING";
+                    };
                     class GVAR(space) {
                         displayName = CSTRING(space_edenName);
                         tooltip = CSTRING(space_edenDesc);
--- a/addons/cargo/XEH_PREP.hpp
+++ b/addons/cargo/XEH_PREP.hpp
@@ -19,3 +19,4 @@ PREP(startLoadIn);
 PREP(startUnload);
 PREP(unloadItem);
 PREP(validateCargoSpace);
+PREP(renameObject);
--- a/addons/cargo/XEH_postInit.sqf
+++ b/addons/cargo/XEH_postInit.sqf
@@ -11,7 +11,7 @@
 
     // Show hint as feedback
     private _hint = [LSTRING(LoadingFailed), LSTRING(LoadedItem)] select _loaded;
-    private _itemName = getText (configFile >> "CfgVehicles" >> typeOf _item >> "displayName");
+    private _itemName = if (_item isEqualType "") then {getText (configfile >> "CfgVehicles" >> typeOf _item >> "displayName")} else {_item getVariable [QGVAR(customName), getText (configfile >> "CfgVehicles" >> typeOf _item >> "displayName")]};
     private _vehicleName = getText (configFile >> "CfgVehicles" >> typeOf _vehicle >> "displayName");
 
     [[_hint, _itemName, _vehicleName], 3.0] call EFUNC(common,displayTextStructured);
@@ -32,7 +32,7 @@
 
     // Show hint as feedback
     private _hint = [LSTRING(UnloadingFailed), LSTRING(UnloadedItem)] select _unloaded;
-    private _itemName = getText (configFile >> "CfgVehicles" >> _itemClass >> "displayName");
+    private _itemName = if (_item isEqualType "") then {getText (configfile >> "CfgVehicles" >> _itemClass >> "displayName")} else {_item getVariable [QGVAR(customName), getText (configfile >> "CfgVehicles" >> _itemClass >> "displayName")]};
     private _vehicleName = getText (configFile >> "CfgVehicles" >> typeOf _vehicle >> "displayName");
 
     [[_hint, _itemName, _vehicleName], 3.0] call EFUNC(common,displayTextStructured);
@@ -78,28 +78,44 @@ GVAR(vehicleAction) = [
     }
 ] call EFUNC(interact_menu,createAction);
 
-GVAR(objectAction) = [
-    QGVAR(load), localize LSTRING(loadObject), "a3\ui_f\data\IGUI\Cfg\Actions\loadVehicle_ca.paa",
-    {
-        params ["_target", "_player"];
-        [_player, _target] call FUNC(startLoadIn);
-    },
-    {
-        //IGNORE_PRIVATE_WARNING ["_target", "_player"];
-        GVAR(enable) &&
-        {(_target getVariable [QGVAR(canLoad), getNumber (configFile >> "CfgVehicles" >> (typeOf _target) >> QGVAR(canLoad))]) in [true, 1]} &&
-        {locked _target < 2} &&
-        {alive _target} &&
-        {[_player, _target, ["isNotSwimming"]] call EFUNC(common,canInteractWith)} &&
-        {((nearestObjects [_target, GVAR(cargoHolderTypes), (MAX_LOAD_DISTANCE + 10)]) findIf {
-            private _hasCargoConfig = 1 == getNumber (configFile >> "CfgVehicles" >> typeOf _x >> QGVAR(hasCargo));
-            private _hasCargoPublic = _x getVariable [QGVAR(hasCargo), false];
-            (_hasCargoConfig || {_hasCargoPublic}) && {_x != _target} && {alive _x} && {locked _x < 2} &&
-            {([_target, _x] call EFUNC(interaction,getInteractionDistance)) < MAX_LOAD_DISTANCE}
-        }) > -1}
-    },
-    LINKFUNC(addCargoVehiclesActions)
-] call EFUNC(interact_menu,createAction);
+GVAR(objectActions)= [
+    [QGVAR(renameObject), localize LSTRING(renameObject), "", //TODO: add icon, maybe a pencil couldn't find it before.
+        {
+            params ["_target"];
+            [_target] spawn FUNC(renameObject);
+        },
+        {
+            //IGNORE_PRIVATE_WARNING ["_target", "_player"];
+            GVAR(enable) &&
+            {(_target getVariable [QGVAR(canLoad), getNumber (configFile >> "CfgVehicles" >> (typeOf _target) >> QGVAR(canLoad))]) in [true, 1]} &&
+            {alive _target} &&
+            {[_player, _target, ["isNotSwimming"]] call EFUNC(common,canInteractWith)} &&
+            {!((typeOf _target) in ["ACE_Wheel", "ACE_Track"])} && // Exclude Wheel and Track
+            {!(_target iskindOf "Land_CanisterFuel_F")} // Exclude Fuel Canisters
+        }
+    ] call EFUNC(interact_menu,createAction),
+    [QGVAR(load), localize LSTRING(loadObject), "a3\ui_f\data\IGUI\Cfg\Actions\loadVehicle_ca.paa",
+        {
+            params ["_target", "_player"];
+            [_player, _target] call FUNC(startLoadIn);
+        },
+        {
+            //IGNORE_PRIVATE_WARNING ["_target", "_player"];
+            GVAR(enable) &&
+            {(_target getVariable [QGVAR(canLoad), getNumber (configFile >> "CfgVehicles" >> (typeOf _target) >> QGVAR(canLoad))]) in [true, 1]} &&
+            {locked _target < 2} &&
+            {alive _target} &&
+            {[_player, _target, ["isNotSwimming"]] call EFUNC(common,canInteractWith)} &&
+            {((nearestObjects [_target, GVAR(cargoHolderTypes), (MAX_LOAD_DISTANCE + 10)]) findIf {
+                private _hasCargoConfig = 1 == getNumber (configFile >> "CfgVehicles" >> typeOf _x >> QGVAR(hasCargo));
+                private _hasCargoPublic = _x getVariable [QGVAR(hasCargo), false];
+                (_hasCargoConfig || {_hasCargoPublic}) && {_x != _target} && {alive _x} && {locked _x < 2} &&
+                {([_target, _x] call EFUNC(interaction,getInteractionDistance)) < MAX_LOAD_DISTANCE}
+            }) > -1}
+        },
+        LINKFUNC(addCargoVehiclesActions)
+    ] call EFUNC(interact_menu,createAction)
+];
 
 // find all remaining configured classes and init them, see XEH_preStart.sqf
 private _vehicleClassesAddAction = call (uiNamespace getVariable [QGVAR(initializedVehicleClasses), {[]}]);
@@ -110,7 +126,8 @@ GVAR(initializedVehicleClasses) append _vehicleClassesAddAction;
 
 private _objectClassesAddAction = call (uiNamespace getVariable [QGVAR(initializedItemClasses), {[]}]);
 {
-    [_x, 0, ["ACE_MainActions"], GVAR(objectAction)] call EFUNC(interact_menu,addActionToClass);
+    [_x, 0, ["ACE_MainActions"], GVAR(objectActions) select 0] call EFUNC(interact_menu,addActionToClass);
+    [_x, 0, ["ACE_MainActions"], GVAR(objectActions) select 1] call EFUNC(interact_menu,addActionToClass);
 } forEach _objectClassesAddAction;
 GVAR(initializedItemClasses) append _objectClassesAddAction;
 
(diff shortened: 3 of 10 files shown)


jonpas:
> I've added a single function which spawns the UI to edit the name and also edits it. This is an scripted UI, it could be changed to a config version if wanted, but as I'm not good in UI coding I left it on this simple version mainly for testing, and to test the idea.

Scripted UI is better, easier to work with.

> These require extra translations.

Just leave them empty. Translators will translate (add their own languages) whenever they will.

> However the game did not like it as it kept creating this from it:

`QUOTE` adds a double quote (`"`) around text, anything inside a double quote is literal (macros will not get resolved. You can either use a single quote (`'`) or another set of double quotes (eg. `QQGVAR˙ does that). Depends what your `GET_STRING` is.

jonpas:
I don't particularly like another interaction menu node just for that. Why not put it into the cargo menu itself? Load and then rename.

I would also keep the base object name and add custom part to the end.

JasperRab:
Well the issue we have in our unit is that we use crates a lot for rearming, and loading the crates in to just rename them feels like extra work. Also we use different boxes for different ammo type, currently I've created multiple empty boxes with different names. however this option would give mission makers/logistic workers the freedom to do stuff their own way, not being limited to the names.
However if others think it would better fit in the cargo menu I could have a look at it.

I could add the custom tag to the end of the normal name, however I initially didn't want to add too much text as it could be annoying.

jonpas:
> and loading the crates in to just rename them feels like extra work

But it's only visible in the cargo menu anyways? I would understand if it would be displayed somewhere else.

As an alternative, Rename action could be a child of the Load action to decrease clutter.

Drofseh:
I think a child action of Load would be good.

JasperRab:
> As an alternative, Rename action could be a child of the Load action to decrease clutter.

When adding it as a child action to Load it would mean that you can't edit the name when you can't load it. As it won't be visible then. I'm not 100% happy about this compromise, I do agree less clutter is better but if it would mean the action won't be available all the time I'd leave it at the mainaction.

> But it's only visible in the cargo menu anyways? I would understand if it would be displayed somewhere else.

True, but I prefer if it can be changed without needing to load an object, that way you can prepare a few crates for later.

jonpas:
2 things I want for this:
- Keep the base object name and add custom part to the end.
- Setting to hide the "Rename" interaction, so people can hide it if they don't want to use it.

JasperRab:
I've changed it so that the custom part is behind the base name (see picture below).
Also I've added the setting to enable/disable the action (default is visible), this can be changed to hidden if wanted.

image

I also saw that there are some conflicting files, I tried to merge from the master to my branch but that hasn't resolved it.
How can I fix this?

addons/cargo/CfgEden.hpp
```diff
@@ -3,6 +3,18 @@ class Cfg3DEN {
...
+                    class GVAR(customName) {
+                        displayName = CSTRING(customName_edenName);
+                        tooltip = CSTRING(customName_edenDesc);
+                        property = QGVAR(customName);
+                        control = "Edit";
+
+                        expression = QUOTE(_this setVariable [ARR_3(QQGVAR(customName), _value, true)];);
+                        defaultValue = "''";
+
+                        condition = "objectHasInventoryCargo-objectVehicle";
```
jonpas, line 15:
```suggestion
                        condition = "objectHasInventoryCargo - objectVehicle";
```

addons/cargo/functions/fnc_getNameItem.sqf
```diff
@@ -0,0 +1,32 @@
+#include "script_component.hpp"
+/*
+ * Author: JasperRab
+ * Gets the name of the item, and alternatively the custom name if requested and available
```
jonpas, line 4:
```suggestion
 * Gets the name of the item, and alternatively the custom name if requested and available.
```

addons/cargo/functions/fnc_getNameItem.sqf
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Arguments:
+ * 0: Target <OBJECT>
+ * 1: Add custom name part <BOOL>
+ *
+ * Return Value:
+ * The display name
+ *
+ * Example:
+ * [crate_7] call ace_cargo_fnc_getNameItem;
```
jonpas, line 14:
```suggestion
 * [crate_7] call ace_cargo_fnc_getNameItem
```

addons/cargo/functions/fnc_getNameItem.sqf
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Public: Yes
+ */
+
+params ["_object", ["_addCustomPart", false]];
+
+private _class = if (_object isEqualType "") then {_object} else {typeOf _object};
+private _displayName = getText (configfile >> "CfgVehicles" >> _class >> "displayName");
```
jonpas, line 22:
```suggestion
private _displayName = getText (configFile >> "CfgVehicles" >> _class >> "displayName");
```

addons/cargo/functions/fnc_getNameItem.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+if (!(_object isEqualType "") && _addCustomPart) then {
+    private _customPart = _object getVariable [QGVAR(customName), ""];
+
+    if (_customPart isNotEqualTo "") then {
+        _displayName = _displayName + " ["+_customPart+"]";
```
jonpas, line 28:
```suggestion
        _displayName = _displayName + " [" + _customPart + "]";
```

addons/cargo/functions/fnc_renameObject.sqf
```diff
@@ -0,0 +1,66 @@
...
+ * Shows rename UI and renames object once completed
+ *
+ * Arguments:
+ * 0: Target <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [crate_7] call ace_cargo_fnc_renameObject;
```
jonpas, line 13:
```suggestion
 * [crate_7] call ace_cargo_fnc_renameObject
```

addons/cargo/functions/fnc_renameObject.sqf
```diff
@@ -0,0 +1,66 @@
+#include "script_component.hpp"
+/*
+ * Author: JasperRab
+ * Shows rename UI and renames object once completed
```
jonpas, line 4:
```suggestion
 * Shows rename UI and renames object once completed.
```

jonpas (changes requested):
Fixed merge conflicts for you, some more styling suggestions.

JasperRab:
Just did a last test with your suggestions and the merge from master, all works a intended.
I've also added myself to the authors file.

jonpas (approved):
Somebody else look at the UI stuff.

addons/cargo/functions/fnc_renameObject.sqf
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Example:
+ * [crate_7] call ace_cargo_fnc_renameObject
+ *
+ * Public: Yes
+ */
+
+params ["_target"];
+
+GVAR(renameTarget) = _target;
```
mharis001, line 20: Save the `_target` as a variable on the display instead.

addons/cargo/XEH_postInit.sqf
```diff
@@ -110,7 +125,8 @@ GVAR(initializedVehicleClasses) append _vehicleClassesAddAction;
 
 private _objectClassesAddAction = call (uiNamespace getVariable [QGVAR(initializedItemClasses), {[]}]);
 {
-    [_x, 0, ["ACE_MainActions"], GVAR(objectAction)] call EFUNC(interact_menu,addActionToClass);
+    [_x, 0, ["ACE_MainActions"], GVAR(objectActions) select 0] call EFUNC(interact_menu,addActionToClass);
+    [_x, 0, ["ACE_MainActions"], GVAR(objectActions) select 1] call EFUNC(interact_menu,addActionToClass);
```
mharis001, line 129: I think `{...} forEach GVAR(objectActions);` instead for when more actions are added.

addons/cargo/XEH_postInit.sqf
```diff
@@ -78,28 +76,45 @@ GVAR(vehicleAction) = [
...
-            {([_target, _x] call EFUNC(interaction,getInteractionDistance)) < MAX_LOAD_DISTANCE}
-        }) > -1}
-    },
-    LINKFUNC(addCargoVehiclesActions)
-] call EFUNC(interact_menu,createAction);
+GVAR(objectActions)= [
+    [QGVAR(renameObject), localize LSTRING(renameObject), "", //TODO: add icon, maybe a pencil couldn't find it before.
+        {
+            params ["_target"];
+            [_target] spawn FUNC(renameObject);
```
mharis001, line 83: Why `spawn`?

addons/cargo/XEH_postInit.sqf
```diff
@@ -78,28 +76,45 @@ GVAR(vehicleAction) = [
...
-        {((nearestObjects [_target, GVAR(cargoHolderTypes), (MAX_LOAD_DISTANCE + 10)]) findIf {
-            private _hasCargoConfig = 1 == getNumber (configOf _x >> QGVAR(hasCargo));
-            private _hasCargoPublic = _x getVariable [QGVAR(hasCargo), false];
-            (_hasCargoConfig || {_hasCargoPublic}) && {_x != _target} && {alive _x} && {locked _x < 2} &&
-            {([_target, _x] call EFUNC(interaction,getInteractionDistance)) < MAX_LOAD_DISTANCE}
-        }) > -1}
-    },
-    LINKFUNC(addCargoVehiclesActions)
-] call EFUNC(interact_menu,createAction);
+GVAR(objectActions)= [
```
mharis001, line 79: Space before and after `=`.

addons/cargo/functions/fnc_getNameItem.sqf
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Public: Yes
+ */
+
+params ["_object", ["_addCustomPart", false]];
+
+private _class = if (_object isEqualType "") then {_object} else {typeOf _object};
+private _displayName = getText (configFile >> "CfgVehicles" >> _class >> "displayName");
+
+if (!(_object isEqualType "") && _addCustomPart) then {
```
mharis001, line 24: `_addCustomPart && {!(_object isEqualType "")}` for lazy eval.

addons/cargo/functions/fnc_getNameItem.sqf
```diff
@@ -0,0 +1,32 @@
...
+/*
+ * Author: JasperRab
+ * Gets the name of the item, and alternatively the custom name if requested and available.
+ *
+ * Arguments:
+ * 0: Target <OBJECT>
+ * 1: Add custom name part <BOOL>
+ *
+ * Return Value:
+ * The display name
```
mharis001, line 11: `Item Name <STRING>`

addons/cargo/XEH_PREP.hpp
```diff
@@ -19,3 +19,5 @@ PREP(startLoadIn);
 PREP(startUnload);
 PREP(unloadItem);
 PREP(validateCargoSpace);
+PREP(renameObject);
+PREP(getNameItem);
```
mharis001, line 23: Sort these alphabetically.

mharis001 (changes requested):

addons/cargo/functions/fnc_renameObject.sqf
```diff
@@ -0,0 +1,66 @@
...
+
+disableSerialization;
+private _display = findDisplay 46 createDisplay "RscDisplayEmpty";
```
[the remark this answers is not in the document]
JasperRab: How should I properly specify display 46? I checked other functions in ACE and 46 gets used quite a lot.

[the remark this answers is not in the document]
mharis001: You can use `IDD_MISSION` from `"\a3\ui_f\hpp\defineResincl.inc"` or `uiNamespace getVariable "RscDisplayMission"`.

addons/cargo/XEH_postInit.sqf
JasperRab: When using call it will crash the game, not entirely sure why. Spawn seems to not let it crash.

jonpas: Use `CBA_fnc_execNextFrame`. I think it's crashing in unscheduled because interaction menu closes in the same frame as you are trying to create your rename dialog.

JasperRab: Yea was my thought as well, will try.

dedmen:
Just saw that this is a thing! I wished for that for quite a while since my logistic boxes are all of same type, but have same fill.
Already thought about just making dummy classes. Great stuff!

JasperRab:
> 
> 
> Just saw that this is a thing! I wished for that for quite a while since my logistic boxes are all of same type, but have same fill.
> Already thought about just making dummy classes. Great stuff!

That's what I have done right now for my group as well, but this seemed easier and also allows the mission maker to decide a bit more.

BrettMayson:
Is the only holdout here the UI?

JasperRab:
> 
> 
> Is the only holdout here the UI?

As far as I know I've applied all the requested changes, if anything else needs to be changed be free to tell me.

jonpas:
@mharis001 re-review please.

addons/cargo/stringtable.xml
```diff
@@ -33,6 +33,21 @@
...
+            <English>Renamed to:&lt;br/&gt;%1</English>
+        </Key>
+        <Key ID="STR_ACE_Cargo_renameObjectUI">
+            <English>Set New Name:</English>
+        </Key>
+        <Key ID="STR_ACE_Cargo_ModuleSettings_enableRename">
+            <English>Enable Rename Action</English>
+        </Key>
+        <Key ID="STR_ACE_Cargo_ModuleSettings_enableRename_Description">
+            <English>Enables the rename action for renamable objects</English>
```
mharis001, line 49:
```suggestion
            <English>Enables the rename action for renamable objects.</English>
```

addons/cargo/stringtable.xml
```diff
@@ -239,6 +254,12 @@
             <Chinesesimp>%1&lt;br/&gt;无法被卸载</Chinesesimp>
             <Turkish>%1&lt;br/&gt; kargo boşaltılamadı</Turkish>
         </Key>
+        <Key ID="STR_ACE_Cargo_customName_edenName">
+            <English>Custom Name</English>
+        </Key>
+        <Key ID="STR_ACE_Cargo_customName_edenDesc">
+            <English>Set a custom cargo name used in the cargo interface</English>
```
mharis001, line 261:
```suggestion
            <English>Set a custom cargo name used in the cargo interface.</English>
```

addons/cargo/functions/fnc_getNameItem.sqf
```diff
@@ -0,0 +1,32 @@
+#include "script_component.hpp"
+/*
+ * Author: JasperRab
+ * Gets the name of the item, and alternatively the custom name if requested and available.
+ *
+ * Arguments:
+ * 0: Target <OBJECT>
+ * 1: Add custom name part <BOOL>
```
mharis001, line 8:
```suggestion
 * 1: Add custom name part <BOOL> (default: false)
```

addons/cargo/functions/fnc_renameObject.sqf
```diff
@@ -0,0 +1,67 @@
...
+ * Public: Yes
+ */
+
+params ["_target"];
+
+private _display = findDisplay IDD_MISSION createDisplay "RscDisplayEmpty";
+_display displayAddEventHandler ["Unload", {
+    params ["_display", "_exitCode"];
+    if (_exitCode isEqualTo 1) then { // only set if pressed on the Save button
+        private _ctrlEdit = (((allControls _display) select {ctrlClassName _x isEqualTo "RscEdit"}) select 0);
```
mharis001, line 25: A better way would be to just `setVariable` `_ctrlEdit` on the display and access it here using `getVariable`.

mharis001 (changes requested):
If you click the background in the GUI the edit box loses focus and cannot be clicked into. Only way to go back into the edit box is by using TAB. Also, if the edit box is not in focus, you can move the player around with WASD (use a dialog to fix).

JasperRab:
I've moved the UI to a config file so it can be used with createDialog.
New UI looks like this:
image

To make the UI I've taken the cargo menu as template and rebuild it with the correct sizes from there.
I've also rewritten the renameObject function as the UI stuff wasn't needed anymore.

JasperRab:
@mharis001 I've fixed/reworked the UI as stated in my last message, could you have another look at it?

BrettMayson:
@JasperRab Can you fix the conflict?

JasperRab:
@BrettMayson Done, wasn't aware about the conflict. so thanks for reminding.

PR #8025, opened 2020-11-30
Map Gestures - Fix briefing mode group

smitt14ua:
Fix error `Error append: Type Group, expected Array`.

### DIFF
--- a/addons/map_gestures/functions/fnc_getProximityPlayers.sqf
+++ b/addons/map_gestures/functions/fnc_getProximityPlayers.sqf
@@ -37,7 +37,7 @@ switch (_this) do {
         _proximityPlayers append allUnits;
     };
     case (1): { // Players in Group
-        _proximityPlayers append units group ace_player;
+        _proximityPlayers append (units group ace_player);
     };
     case (2): { // Players on Side
         _proximityPlayers append (allUnits select {side (group _x) == side (group ace_player)});
--- a/addons/map_gestures/functions/fnc_getProximityPlayers.sqf
+++ b/addons/map_gestures/functions/fnc_getProximityPlayers.sqf
@@ -37,7 +37,7 @@ switch (_this) do {
         _proximityPlayers append allUnits;
     };
     case (1): { // Players in Group
-        _proximityPlayers append group ace_player;
+        _proximityPlayers append units group ace_player;
     };
     case (2): { // Players on Side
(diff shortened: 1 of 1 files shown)


addons/map_gestures/functions/fnc_getProximityPlayers.sqf
```diff
@@ -37,7 +37,7 @@ switch (_this) do {
         _proximityPlayers append allUnits;
     };
     case (1): { // Players in Group
-        _proximityPlayers append group ace_player;
+        _proximityPlayers append units group ace_player;
```
jonpas, line 40:
```suggestion
        _proximityPlayers append (units group ace_player);
```

veteran29: `units` works also with just the unit.
`_proximityPlayers append (units ace_player);`

smitt14ua: > `units` works also with just the unit.
> `_proximityPlayers append (units ace_player);`

Sounds good.

PabstMirror (approved):

veteran29 (approved):
Not a fan of these parentheses as I believe SQF is beautiful without them but LGTM

bux:
> SQF is beautiful

...

addons/map_gestures/functions/fnc_getProximityPlayers.sqf
```diff
@@ -37,7 +37,7 @@ switch (_this) do {
         _proximityPlayers append allUnits;
     };
     case (1): { // Players in Group
-        _proximityPlayers append group ace_player;
+        _proximityPlayers append (units ace_player);
```
commy2, line 40:
```suggestion
        _proximityPlayers append units ace_player;
````

jonpas: Why? Parentheses help readability without a single downside. It is used widely in the harshest standards to assure correct execution of code in the entire industry. I am not a fan of full MISRA for this use case, but this is a perfect example where parentheses outright improve readability (and possibly allow easy changing of commands that might have higher priority instead).

Dystopian: I like commy's suggestion here. A matter of taste I think.

PR #8031, opened 2020-12-02
Advanced Throwing - Add object variable to disable pick up of attached throwables

### LINKED ISSUE #8030: Advanced Throwing - Ways to disable pickup on specific throwables
While working on a recent mission, I came across an incredibly tiny issue that I doubt many people have run into - creating chemlights attached to a specific object that I don't want to be removable. I don't want to disable the ability to pick up attached grenades, as my players have made use of it in the past with IR strobes, but at the same time I don't want players being able to pick up and toss away a chemlight that's been spawned through `createVehicle` at mission start and `attachTo`'d a specific mission objective. I can also see potential usage for things like scripted flares or smoke grenades, ones which are created through mission logic that mission designers might not want players to touch, but don't want to attach them to anything at all. If there's already a way to do this, then it doesn't seem to be documented anywhere on the wiki, and I couldn't find anything that might have such effects browsing through Github here (though admittedly, I'm a mediocre scripter and may have missed something obvious).

jonpas:
- Close #8030 
- Usage: `OBJECT setVariable ["ace_advanced_throwing_disablePickUp", true, true]`

### DIFF
--- a/addons/advanced_throwing/functions/fnc_renderPickUpInteraction.sqf
+++ b/addons/advanced_throwing/functions/fnc_renderPickUpInteraction.sqf
@@ -31,7 +31,7 @@
             _nearThrowables append (ACE_player nearObjects ["ACE_Chemlight_IR_Dummy", PICK_UP_DISTANCE]);
 
             {
-                if (!(_x in _throwablesHelped) &&
+                if (!(_x in _throwablesHelped) && {!(_x getVariable [QGVAR(disablePickUp), false])} &&
                     {!(_x isKindOf "SmokeShellArty")} && {!(_x isKindOf "G_40mm_Smoke")} && // All smokes inherit from "GrenadeHand" >> "SmokeShell"
                     {GVAR(enablePickUpAttached) || {!GVAR(enablePickUpAttached) && {isNull (attachedTo _x)}}}
                 ) then {
--- a/addons/advanced_throwing/functions/fnc_renderPickUpInteraction.sqf
+++ b/addons/advanced_throwing/functions/fnc_renderPickUpInteraction.sqf
@@ -31,7 +31,7 @@
             _nearThrowables append (ACE_player nearObjects ["ACE_Chemlight_IR_Dummy", PICK_UP_DISTANCE]);
 
             {
-                if (!(_x in _throwablesHelped) && {!(_x getVariable [QGVAR(disablePickUp), false])} &&
+                if (!(_x in _throwablesHelped) && {!((attachedTo _x) getVariable [QGVAR(disablePickUp), false])} &&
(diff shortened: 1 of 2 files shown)


jonpas:
Good point, can you even get a reference to it in all use cases? Otherwise a class binding could be used.

PabstMirror:
```
{ 
    _x setVariable ["ace_advanced_throwing_disablePickUp", true]; 
    systemChat format ["Blocking %1 = %2", _x, _x getVariable ["ace_advanced_throwing_disablePickUp", -1]]; 
} forEach (attachedObjects cursorObject)
```
chemlight = -1
guess you can't set any vars on ammo

addons/advanced_throwing/functions/fnc_renderPickUpInteraction.sqf
```diff
@@ -31,7 +31,7 @@
             _nearThrowables append (ACE_player nearObjects ["ACE_Chemlight_IR_Dummy", PICK_UP_DISTANCE]);
 
             {
-                if (!(_x in _throwablesHelped) &&
+                if (!(_x in _throwablesHelped) && {!(_x getVariable [QGVAR(disablePickUp), false])} &&
```
PabstMirror, line 34:
```suggestion
                if (!(_x in _throwablesHelped) && {(attachedTo _x) getVariable [QGVAR(allowPickup), true]} &&
```

PabstMirror: could put the var on the attached source, should still solve #8030

jonpas: I like this.

Couldn't decide between "disable" and "allow". I usually like the reverse (if disabled by default -> allow, if enabled by default -> disabled).

PabstMirror:
```suggestion
                if (!(_x in _throwablesHelped) && {!((attachedTo _x) getVariable [QGVAR(disablePickUp), false])} &&
```

PabstMirror: otherway saves a `!` and `()` 😄

jonpas: Yeah :joy:  It of course doesn't matter at the end. Will update documentation. Did you test this?

PabstMirror:
Updated to put the var on the vehicle that has ammo attached to it. and updated doc

PabstMirror (approved):

jonpas:
Seeing a commit on that branch name scared me that something broke in last Arma update, then I realized this is on my PR. Phew!

PR #8034, opened 2020-12-07
Fix Multiple Issues With Timestamps

commy2:
The pull request that added the auto timestamping to the marker menu was incredibly rushed and broken.

- properly aligns text with "Timestamp" checkbox vertically
- makes checkbox a square instead of a fucking rectangle
- fixes completely broken 12-hour clock
- properly handles 12:xx am (after noon) and 12:xx pm (after midnight); 00:00 am/pm is not a thing...
- avoids the broken space character, because it is not a space in most fonts, bad workaround
- instead, adds ace_marker_fnc_removeTimestamp function
- removes bad code formatting changes with mid line whitespace
- includes the vanilla header file in the common component header file instead of in 5 different functions...

before

after:

### DIFF
--- a/addons/markers/functions/fnc_initInsertMarker.sqf
+++ b/addons/markers/functions/fnc_initInsertMarker.sqf
@@ -1,5 +1,4 @@
 #include "script_component.hpp"
-#include "\A3\ui_f\hpp\defineResincl.inc"
 /*
  * Author: BIS, commy2, Timi007
  * Sets up the marker placement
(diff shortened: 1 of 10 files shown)


addons/markers/functions/fnc_initInsertMarker.sqf
```diff
@@ -93,17 +90,9 @@
     _display displayAddEventHandler ["KeyDown", {(_this select 1) in [200, 208]}];
 
     private _hasTimestamp = false;
-    if !((markerText GVAR(editingMarker)) isEqualTo "") then {
+    if !(markerText GVAR(editingMarker) isEqualTo "") then {
```
jonpas, line 93: Can `markerText` return anything but a string?
```suggestion
    if !(markerText GVAR(editingMarker) == "") then {
```

addons/markers/functions/fnc_initInsertMarker.sqf
```diff
@@ -93,17 +90,9 @@
...
-
-        private _originalText = markerText GVAR(editingMarker);
-        private _timeIndex = _originalText find (TIMESTAMP_SPACE + "[");
-        if (_timeIndex > 0 ) then {
-            // Shave off timestamp
-            _hasTimestamp = true;
-            _originalText = _originalText select [0,_timeIndex];
-        };
-
+        private _originalText = markerText GVAR(editingMarker) call FUNC(removeTimestamp);
```
jonpas, line 95: Calling `markerText` twice.

jonpas (commented):
Quick check. Formatting/alignment changes are completely redundant.

addons/markers/functions/fnc_initInsertMarker.sqf
commy2: No, but `isEqualTo` is slightly faster while being exactly as readable. It is used that way all over the code in CBA and ACE.

addons/markers/functions/fnc_initInsertMarker.sqf
commy2: The trade off is a variable. And that variable is only used once for every marker creation, and only the minimum two for the arguably much rarer marker editing. I think both methods are equal.

addons/markers/functions/fnc_initInsertMarker.sqf
jonpas: I disagree on the readability (it should have been `===` or something), but this is fine.

addons/markers/functions/fnc_initInsertMarker.sqf
jonpas: This line and the `if` statement above would become more readable. Also single source of value is always better.

addons/markers/functions/fnc_initInsertMarker.sqf
dedmen: also a candidate for isNotEqualTo, which will probably land before next ace update

addons/markers/functions/fnc_initInsertMarker.sqf
dedmen: markerText is much faster in 2.02 than it was before.

addons/markers/functions/fnc_removeTimestamp.sqf
```diff
@@ -0,0 +1,46 @@
...
+// @todo
+_string = toArray _string;
+reverse _string;
+_string = toString _string;
+
+if (_string select [0, 1] != "[") exitWith {_original};
+
+if !(_string select [1, 1] in DIGITS) exitWith {_original};
+if !(_string select [2, 1] in DIGITS) exitWith {_original};
+if (_string select [3, 1] != ":") exitWith {_original};
```
PabstMirror, line 42: Fails if using `"HH"` format

addons/markers/functions/fnc_removeTimestamp.sqf
```diff
@@ -0,0 +1,46 @@
...
+ * "[13:37]" call ace_markers_fnc_removeTimestamp // ""
+ * "xyz [123]" call ace_markers_fnc_removeTimestamp // "xyz [123]"
+ *
+ * Public: No
+ */
+#define DIGITS "0123456789"
+
+params ["_original"];
+
+// @todo, 2.02 reverse command will support STRING types
```
PabstMirror, line 23: `reverse` does not like unicode
image
copies as `"�ссѵЀѿ� �добовС�"`

addons/markers/functions/fnc_removeTimestamp.sqf
PabstMirror: I think we just need to drop this and the next few lines to support time codes like `[12]`
or just drop HH format 
(I'm also not sure I understand why we have `"HH:MM:SS:MM"` but that's unrelated)

PabstMirror (changes requested):
timecode search

addons/markers/functions/fnc_removeTimestamp.sqf
commy2: : (

PabstMirror (approved):

addons/markers/functions/fnc_removeTimestamp.sqf
PabstMirror: @commy2 - I commited a fix to this PR, let me know if it looks ok

commy2:
Thanks for finishing this.

addons/markers/functions/fnc_removeTimestamp.sqf
dedmen: With forceUnicode?

PR #8036, opened 2020-12-07
Interact Menu - Add settings for self interaction menu reordering

Dystopian:
- add `More` submenu to unit self interaction menu and settings to move any root menu items there;
- add settings to move any submenu items to root.

I often use self interaction menu. But some items I use very often and some ones rarely. E.g. I put earplugs in each time I enter static/vehicle weapon and put them out on exit (to hear better). On the other hand I use `Team management` only once during mission and `Create Zeus` once in several missions. I don't want do disable `Gestures` and `Check air temperature` but  they are better for me in `More` submenu. So I would like to move earplugs to root and hide others in `More` submenu.

All settings are disabled by default and require mission restart. ~To use them:~
~- Settings - ACE Interaction menu - ACE Interaction menu (Self) - Allow reordering;~
~- restart mission;~
~- Settings - ACE Interaction menu (Self) - More/Move to Root - check what you need;~
~- restart mission.~


jonpas:
Interesting!

I am wondering if we really need the overall "Allow Reordering" setting, it seems to just add complexitiy. We could just generate all relevant settings on start, I don't see a real downside.

@veteran29 ?

Dystopian:
> I am wondering if we really need the overall "Allow Reordering" setting, it seems to just add complexitiy. We could just generate all relevant settings on start, I don't see a real downside.

I had 2 reasons:
1. `Move to Root` menu is quite big and is opened with noticeable lag. Some time ago @dedmen said in ACE Slack that he found reason of this for game settings (as I understood) and if it's fixed in future this point is out.
1. this settings can be not very intuitive for newbies.

Personally I don't care - please you decide.

jonpas:
Maybe I wasn't clear enough or I don't know if I understood you. If I am reading this correctly, when user enables "Allow Reordering" setting and restarts the mission, user gets additional settings for additional actions "Move X to Root", "Move Y to Root", correct?

I am not sure what @dedmen's concerns were (can you enlighten us @dedmen?), but if it's not a few seconds of delay, then I am fine with there being many settings.

How many settings are there in total with all of ACE3 anyways?

Dystopian:
> If I am reading this correctly, when user enables "Allow Reordering" setting and restarts the mission, user gets additional settings for additional actions "Move X to Root", "Move Y to Root", correct?

Additional setting categories `Interaction menu (self) - More` and `Interaction menu (self) - Move to root` appear.

> I am not sure what @dedmen's concerns were (can you enlighten us @dedmen?), but if it's not a few seconds of delay, then I am fine with there being many settings.
>
> How many settings are there in total with all of ACE3 anyways?

I have a delay about 1.2s on my PC when opening `Move to root` category. There are about 80 entries there.

jonpas:
Is it that slow simply because there are 80 settings in there? You don't do anything on the fly, everything happens on the `newControllableObject` event. So it's only CBA UI being slow with that amount of settings? (Not saying that it should be faster.)

What about splitting that large menu into multiple smaller menus to make it faster? Maybe Equipment, Interaction, ..., Other?

Dystopian:
> Is it that slow simply because there are 80 settings in there? You don't do anything on the fly, everything happens on the `newControllableObject` event. So it's only CBA UI being slow with that amount of settings? (Not saying that it should be faster.)

Yes. 

> What about splitting that large menu into multiple smaller menus to make it faster? Maybe Equipment, Interaction, ..., Other?

There are many ACE categories in settings already. I don't think splitting this setting to submenu is cool. Also I think (at least from my experience) this setting is opened and used only once, so delay is not a real problem.

addons/interact_menu/functions/fnc_initMenuReorder.sqf
```diff
@@ -0,0 +1,89 @@
...
+
+    _node params ["_actionData", "_actionChildren"];
+    _actionData params ["_name", "_title", "", "", "_condition"];
+    private _varName = _parentVarName + "__" + _name;
+    private _titleFull = _titlePrefix + _title;
+    private _conditionFullString = format ["%1 && {%2}", _parentConditionString, _condition call EFUNC(common,codeToString)];
+
+    // don't add already added setting
+    if (isNil {missionNamespace getVariable _varName}) then {
+        [_varName, "CHECKBOX", _titleFull, [_settingCategory, _rootActionTitle], false, false, {}, true] call CBA_fnc_addSetting;
```
jonpas, line 40:
```suggestion
        [_varName, "CHECKBOX", _titleFull, [_settingCategory, _rootActionTitle], false, 2, {}, true] call CBA_fnc_addSetting;
```
Should we do this? So people don't force those settings when they export them all?

jonpas (commented):
I don't think the delay is a problem either. As long as it isn't too long to trigger ACRE/TFAR disconnect.

I say we get rid of the "Allow Reordering" setting. @PabstMirror @BaerMitUmlaut @mharis001 (and others) What's your take on this?

mharis001:
I agree with @jonpas . I think the "Allow reordering" setting is not needed; these additional settings should always be available.

I terms of performance, I didn't notice any delays other than the expected one when selecting the "Move to Root" settings category. Even that wasn't too bad - not longer than a second or so.

addons/interact_menu/functions/fnc_initMenuReorder.sqf
mharis001: I think this makes sense since these settings are purely individual preference.

dedmen:
> Some time ago @dedmen said in ACE Slack that he found reason of this for game settings (as I understood) and if it's fixed in future this point is out.

That was checkboxes in game settings menu making texture lookups each.
I recently greatly optimized texture lookup speed on profiling branch, that should mitigate that

jonpas (approved):

mharis001 (approved):

PR #8039, opened 2020-12-10
RHS Compat - Add ACE_GForceCoef value for DF-15 uniforms

YetheSamartaka:

- Add missing ACE_GForceCoef value for DF-15 uniforms.
- ACE_GForceCoef value is the same as for Vanilla counterparts(Many other mods are using these values for their uniforms as well)

### DIFF
--- a/optionals/compat_rhs_afrf3/CfgWeapons.hpp
+++ b/optionals/compat_rhs_afrf3/CfgWeapons.hpp
@@ -332,13 +332,14 @@ class CfgWeapons {
         modes[] = {};
         picture = "\rhsafrf\addons\rhs_heavyweapons\data\ico\rhs_Kornet_9M133_2_msv_ca.paa";
     };
+    
     class rhs_uniform_flora;
-	class rhs_uniform_df15: rhs_uniform_flora
-	{
-		ACE_GForceCoef=0.80000001;
-	};
-	class rhs_uniform_df15_tan: rhs_uniform_df15
-	{
-		ACE_GForceCoef=0.80000001;
-	};
+    class rhs_uniform_df15: rhs_uniform_flora
+    {
+        ACE_GForceCoef=0.8;
+    };
+    class rhs_uniform_df15_tan: rhs_uniform_df15
+    {
+        ACE_GForceCoef=0.8;
+    };
 };
--- a/optionals/compat_rhs_afrf3/CfgWeapons.hpp
+++ b/optionals/compat_rhs_afrf3/CfgWeapons.hpp
@@ -332,4 +332,13 @@ class CfgWeapons {
         modes[] = {};
         picture = "\rhsafrf\addons\rhs_heavyweapons\data\ico\rhs_Kornet_9M133_2_msv_ca.paa";
     };
+    class rhs_uniform_flora;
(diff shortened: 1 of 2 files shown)


mharis001 (commented):
Please fix the formatting:

- 4 spaces per indentation level.
- Opening brace on same line.
- Spaces around `=`.

And simplify the value to `0.8`.

YetheSamartaka:
> 
> 
> Please fix the formatting:
> 
>     * 4 spaces per indentation level.
> 
>     * Opening brace on same line.
> 
>     * Spaces around `=`.
> 
> 
> And simplify the value to `0.8`.

All bulletpoints "fixed", it should be all right.

commy2:
lgtm though idk if the class inheritance is correct

YetheSamartaka:
> 
> 
> lgtm though idk if the class inheritance is correct

Initialy I tried it with **Uniform_Base** for all uniforms, but it worked only for _saf3_ uniforms. I peeked into RHS configs and found out, that for afrf3, they are indeed inheriting it from **rhs_uniform_flora**. I tried it in-game, it worked without any issues.

YetheSamartaka:
> 
> 
> > lgtm though idk if the class inheritance is correct
> 
> Initialy I tried it with **Uniform_Base** for all uniforms, but it worked only for _saf3_ uniforms. I peeked into RHS configs and found out, that for afrf3, they are indeed inheriting it from **rhs_uniform_flora**. I tried it in-game, it worked without any issues.

Addition to this, I just looked into _saf3_ config, and they are inheriting it from **Uniform_Base**, so it should be all right - this is RHS mess.

commy2:
Don't guess inheritance. Always take it from the config you base it on.

YetheSamartaka:
> 
> 
> Don't guess inheritance. Always take it from the config you base it on.

I knew that class from the past, just double-checked it 😅

YetheSamartaka:
What will happen next? Will it get approved, some changes requiered to be performed, or casted into the void?

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -332,4 +332,12 @@ class CfgWeapons {
         modes[] = {};
         picture = "\rhsafrf\addons\rhs_heavyweapons\data\ico\rhs_Kornet_9M133_2_msv_ca.paa";
     };
+    
+    class rhs_uniform_flora;
+    class rhs_uniform_df15: rhs_uniform_flora {
+        ACE_GForceCoef = 0.8;
+    };
+    class rhs_uniform_df15_tan: rhs_uniform_df15 {
+        ACE_GForceCoef = 0.8;
```
PabstMirror, line 341: because it inherits we don't need the config, but it doesn't really hurt

PabstMirror (approved):

optionals/compat_rhs_afrf3/CfgWeapons.hpp
commy2: I'd say remove it. It is just a reskin. The less external config, the better.

mharis001 (approved):

PR #8045, opened 2020-12-13
MissileGuidance - Active Radar Homing: Use AGL for `nearestObjects` check

TheCandianVendingMachine:
- Fix bug where AGM-114L used `positionASL` instead of `positionAGL` to check for targets

When developing I always used Altis near sea level for moving targets. This caused me to not catch this. Whoops

### DIFF
--- a/addons/missileguidance/functions/fnc_seekerType_ARH.sqf
+++ b/addons/missileguidance/functions/fnc_seekerType_ARH.sqf
@@ -71,7 +71,6 @@ if (_isActive || { CBA_missionTime >= _timeWhenActive }) then {
             };
         };
         _nearestObjects = _nearestObjects select { !isNull _x };
-        
         // Select closest object to the expected position to be the current radar target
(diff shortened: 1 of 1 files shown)


addons/missileguidance/functions/fnc_seekerType_ARH.sqf
```diff
@@ -70,7 +71,7 @@ if (_isActive || { CBA_missionTime >= _timeWhenActive }) then {
             };
         };
         _nearestObjects = _nearestObjects select { !isNull _x };
-
+
```
jonpas, line 74:
```suggestion

```

jonpas (changes requested):

jonpas:
I meant the spaces on the empty line. Empty line itself is fine. :D

jonpas (approved):

PabstMirror (approved):

PR #8049, opened 2020-12-25
Disable pboproject warnings are errors in build script

Dahlgren:
Macros with unused variables are warnings in latest public version
This causes issues with logging macros being unused with disabled debug


Dahlgren:
`-Warnings` seems to be valid argument in previous public version from Fall 2020 as well

PabstMirror (approved):

PabstMirror:
The setting seems to be persistent (can change in pboProject gui)
but a lot easier to have set via cmd line

PR #8055, opened 2021-01-05
Markers - Add scale slider

frankplow:
- Add scale slider to marker placement dialog
- Update map EHs to reflect current selected scale
- Update marker placement scripts to incorporate scale


commy2:
conflicts with #8034

PR #8060, opened 2021-01-06
Interact Menu - Combine parent menu with single child

Dystopian:
- add setting to combine parent menu with single child;
- adapt `fastroping` `Cut ropes` action for the setting;
- clean up `chemlights` and `map` menu configs a little.

Setting is disabled by default. ~Still not sure about setting name and description.~

### DIFF
--- a/addons/interact_menu/functions/fnc_collectActiveActionTree.sqf
+++ b/addons/interact_menu/functions/fnc_collectActiveActionTree.sqf
@@ -102,7 +102,7 @@ if ((_activeChildren isEqualTo []) && {_statementCode isEqualTo {}}) exitWith {
 };
 
 if (GVAR(consolidateSingleChild) && {count _activeChildren == 1} && {_statementCode isEqualTo {}}) then {
-    _activeChildren select 0 params ["_childActionData", "_childChildren"];
+    _activeChildren select 0 params ["_childActionData", "_childChildren", "_childObject"];
     _childActionData params ["", "_displayNameChild", "_iconChild", "_statementChild", "", "", "_customParamsChild", "", "", "_paramsChild"];
     _origActionData = [
         _actionName,
@@ -117,6 +117,7 @@ if (GVAR(consolidateSingleChild) && {count _activeChildren == 1} && {_statementC
         _paramsChild
     ];
     _activeChildren = _childChildren;
+    _object = _childObject;
 };
 
 [_origActionData, _activeChildren, _object]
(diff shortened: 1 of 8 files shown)


veteran29 (approved):
works. Nice feature 👍

PabstMirror:
Really minor, but I want to test how this will work with https://github.com/acemod/ACE3/pull/6876
where I used the child as a "safety"
I think we can just add a dummy `statement` if it's a problem

addons/interact_menu/stringtable.xml
```diff
@@ -489,5 +489,9 @@
             <Spanish>Selector de color</Spanish>
             <Turkish>Seçici Renk</Turkish>
         </Key>
+        <Key ID="STR_ACE_Interact_Menu_combineParentWithSingleChild">
+            <English>Combine parent menu with single child item</English>
```
mharis001, line 493: I think title case for setting display names, maybe something like: "Consolidate Single Child Actions". And then the description can be: "Combines parent actions with only one child action together."

Dystopian: > "Combines parent actions with only one child action together."

I don't think camel case is good for settings especially for those looking like sentences. Also I replaced `parent actions` with `parent action`.

addons/interact_menu/functions/fnc_collectActiveActionTree.sqf
```diff
@@ -91,5 +101,22 @@ if ((_activeChildren isEqualTo []) && {_statementCode isEqualTo {}}) exitWith {
     []
 };
 
+if (GVAR(consolidateSingleChild) && {count _activeChildren == 1} && {_statementCode isEqualTo {}}) then {
+    _activeChildren select 0 params ["_childActionData", "_childChildren"];
```
PabstMirror, line 105:
```suggestion
    _activeChildren select 0 params ["_childActionData", "_childChildren", "_childObject"];
```

addons/interact_menu/functions/fnc_collectActiveActionTree.sqf
```diff
@@ -91,5 +101,22 @@ if ((_activeChildren isEqualTo []) && {_statementCode isEqualTo {}}) exitWith {
...
+        _iconChild,
+        _statementChild,
+        _conditionCode,
+        _insertChildrenCode,
+        _customParamsChild,
+        _position,
+        _distance,
+        _paramsChild
+    ];
+    _activeChildren = _childChildren;
```
PabstMirror, line 119:
```suggestion
    _activeChildren = _childChildren;
    _object = _childObject;
```

PabstMirror (changes requested):
We need this if child action has different object
e.g. a single passenger won't show any interactions because the object was still on the vehicle

PabstMirror (approved):

Dystopian:
@PabstMirror please review the last commit

PabstMirror (approved):

Issue #7898, opened 2020-09-06
Medical - Units inside exploding vehicle don't take enough damage


mjc4wilton:
**Is your enhancement related to a problem?**
If a player or AI is inside a vehicle when it explodes, they are mostly fine and often times can walk away, albeit limping. Same issue is across ground and air vehicles, most likely why helicopter / plane crashes are underwhelming.

**Solution you'd like:**
Increase the damage a crew receives when inside a vehicle.

**Alternatives you've considered:**
None

**Additional context:**
Does not matter if the player wears body armor or what vehicle they were in. Effect is still minimal

Vdauphin:
See https://github.com/acemod/ACE3/issues/7305

kymckay:
Closing in favour of keeping conversation to #7305

PR #8063, opened 2021-01-08
Improved damage handling in medical_engine and medical_damage

### LINKED ISSUE #7305: Medical - Define system for Vehicle Crashes / Explosions
We need a plan for how to handle vehicle crashes and explosions (vehicle killed)

Current code: https://github.com/acemod/ACE3/blob/master/addons/medical_engine/functions/fnc_handleDamage.sqf#L190-L208

Problems
 - Only triggers on velocity, stationary vehicles being hit is a problem, but more critical is that sometimes the velocity seems to be reduced before the HD triggers.
e.g. put chopper in air, and do `w setVelocity [0,0,-100];`
often the velocity in HD is `[0,0,-97.3566]`, but about 30% of the time it's a special value of `[0,0,-0.1]` which is not enough to trigger (seems like the velocity got reset before the frame starts?)
I have seen transport chopper go down hard and explode, and then AI hop out and fight like nothing happened.

- Inconstant damage. The raw HD is usually something like
`[wD,"",1,wD,"",-1,<NULL-object>,""]`
in vanilla this would always be fatal, but we can only translate it to a random hitpoint wound, which may not even cause unconsciousness.

- Being alive inside a destroyed vehicle may cause problems. 
 
There is an unconscious pilot in this picture 😄 but no way to see or interact with him. Should we auto-eject or trigger periodic damage for being inside a burning wreck? Right now you are actually safer inside the burning wreck because it seems to protect from "FuelExplosion" damage.

- Exploding cars is dumb. So I'm mostly glad wrecks are more survivable, perhaps we  can further reduce damage from "FuelExplosion" damage types?  But Tanks that are cooking off should be dangerous. Aircraft should be extremely dangerous.

Ref
https://github.com/acemod/ACE3/pull/6330 - Boats ejecting PR, somewhat related
https://github.com/acemod/ACE3/pull/7304 - Manual tweak to make AIrcraft Crashes more Lethal

gandjustas:
- Changed hit detection from using ACE_HDBracket to Hit EH, less special cases to handle
- Added death when inflicted structural damage is 1 or more (fixes #7305, #7898)
- Moved headshot detection to HandleDamage and added instant death by headshot (practicaly faceshot)
- Moved heart hit detection before assigning wounds
- Changed ACE limb hitpoints to armor=1 and using vanilla limb hitpoint damage to assign wounds (lowers damage to limbs)
- Added fatality check for total structural damage (10-15 hits to limbs will kill unit, even if there is no blood loss)
- Changed wounds handling, damage applies to all bodyparts (like vanilla), wounds created on selection-specific bodyparts or all damaged bodyparts if damage is non-selection specific (grenades creates a lot small wounds in all bodyparts)

Downsides:
- Wounds not assigned when unit is dead (Hit EH is not call on dead bodies)
- Probably multiple Killed events


commy2:
Scary

addons/medical_damage/functions/fnc_applyDamage.sqf
```diff
@@ -0,0 +1,220 @@
...
+                HITPOINT_INDEX_LARM, 0.1, HITPOINT_INDEX_RARM, 0.1, 
+                HITPOINT_INDEX_LLEG, 1, HITPOINT_INDEX_RLEG, 1
+            ];
+            private _damagedBodyPart = selectRandomWeighted _damageSelectionArray;
+            private _bodyPartDamage = [[_damagedBodyPart, (_damage min PENETRATION_THRESHOLD)  / (GVAR(structuralDamageToBodyPartCoeff) select _damagedBodyPart)]];
+
+            // Significant damage suggests high relative velocity
+            // Momentum transfers to body/head for worse wounding
+            // Higher momentum results in higher chance for head to be hit for more lethality
+            if (_damage > PENETRATION_THRESHOLD) then {
```
BaerMitUmlaut, line 71: Vehicle crashes are related to bullet penetration now?

addons/medical_damage/functions/fnc_applyDamage.sqf
```diff
@@ -0,0 +1,220 @@
...
+            };
+
+            [QEGVAR(medical,woundReceived), [
+                _unit, 
+                _damagedBodyPart, 
+                _damage, 
+                "vehiclecrash", 
+                _bodyPartDamage
+                ]
+            ] call CBA_fnc_localEvent;
```
BaerMitUmlaut, line 86: https://ace3mod.com/wiki/development/coding-guidelines.html#51-braces-placement

addons/medical_damage/functions/fnc_applyDamage.sqf
```diff
@@ -0,0 +1,220 @@
...
+            private _damagedBodyPart = selectRandomWeighted _damageSelectionArray;
+
+            TRACE_3("Unknown damage",_source,_damage,_damagedBodyPart);
+            [QEGVAR(medical,woundReceived), [
+                _unit, 
+                _damagedBodyPart, 
+                _damage, 
+                _typeOfDamage, 
+                [[_damagedBodyPart, _damage / (GVAR(structuralDamageToBodyPartCoeff) select _damagedBodyPart)]]
+                ]
```
BaerMitUmlaut, line 107: https://ace3mod.com/wiki/development/coding-guidelines.html#51-braces-placement

addons/medical_damage/functions/fnc_applyDamage.sqf
```diff
@@ -0,0 +1,220 @@
...
+        ] select { _x#1 > 0 } ;
+        TRACE_4("received",_receivedDamage,_woundedHitPoint,_bodyPartDamage,_damage);
+
+        private _sum = 0;
+        {
+            _sum = _sum + _x#1 * (GVAR(structuralDamageToBodyPartCoeff) select _x#0);
+        } forEach _bodyPartDamage;
+
+        // Sum of bodypart damage is too low
+        if(_damage/_sum > 2) then {
```
BaerMitUmlaut, line 192: Missing space

addons/medical_damage/functions/fnc_applyDamage.sqf
```diff
@@ -0,0 +1,220 @@
...
+            _bodyPartDamage pushBack [_additionalBodyPart, _additionalDamage / (GVAR(structuralDamageToBodyPartCoeff) select _additionalBodyPart)];
+        };
+
+        [QEGVAR(medical,woundReceived), [
+            _unit, 
+            _woundedHitPointIdx, 
+            _damage, 
+            _typeOfDamage, 
+            _bodyPartDamage]
+        ] call CBA_fnc_localEvent;
```
BaerMitUmlaut, line 218: https://ace3mod.com/wiki/development/coding-guidelines.html#51-braces-placement

addons/medical_damage/functions/fnc_determineIfFatal.sqf
```diff
@@ -5,52 +5,45 @@
...
 
     if (_chanceFatal > random 1) exitWith {
-        TRACE_1("determineIfFatal: lethal trauma",_woundDamage);
+        TRACE_2("determineIfFatal: lethal trauma",_headDamage,_bodyDamage);
+        true breakOut "main";
+    };
+
+    private _structuralDamage = _unit getVariable [QEGVAR(medical,structuralDamage), 0];
+    _vitalDamage = (_structuralDamage - _damageThreshold) max 0;
+    _chanceFatal = 1 - exp -((_vitalDamage/FATAL_SUM_DAMAGE_WEIBULL_L)^FATAL_SUM_DAMAGE_WEIBULL_K);
```
BaerMitUmlaut, line 42: Missing spaces

addons/medical_engine/functions/fnc_parseDamageEvents.sqf
```diff
@@ -0,0 +1,86 @@
...
+    private _cnt = 0;
+
+    // Read events from end, until hitPointIndex==-1
+    while { 
+        _i = _i-1;
+        _event = _events select _i; 
+        _cnt = _cnt + 1;
+        _event#0 > -1
+    } do {
+        if(_event#1 > 0) then {
```
BaerMitUmlaut, line 49: Missing space

addons/medical_engine/functions/fnc_parseDamageEvents.sqf
```diff
@@ -0,0 +1,86 @@
...
+    _events deleteRange [_i, _cnt];
+    private _newHit = [diag_frameNo];
+    _newHit append (_event select [1,4]);
+    _newHit pushBack _hpDamages;
+
+    //Hit format [_frameNo, _damage, _source, _instigator, _ammo, _hpDamages]
+    _newHits pushBack _newHit; 
+
+    // Try to consume head prefix from start of stream 
+    if(count _events > 0 && {_events#0#0 == 2}) then {
```
BaerMitUmlaut, line 63: Missing space

addons/medical_engine/functions/fnc_parseDamageEvents.sqf
```diff
@@ -0,0 +1,86 @@
...
+                _events#1#0 == -1 or _events#1#0 == 2
+            }) then {
+                _events deleteAt 0;
+            };
+        };
+    };
+};
+
+// Add parsed hits to collection, remove old hits not complemented by Hit event
+if( count _newHits > 0) then {
```
BaerMitUmlaut, line 79: Missing space

addons/medical_engine/functions/fnc_processLimbsDamage.sqf
```diff
@@ -0,0 +1,31 @@
...
+ * None
+ *
+ * Public: No
+ */
+ 
+params ["_unit", "_damages"];
+_damages params ((_unit getVariable QEGVAR(medical,HitPoints)) apply { ["_" + _x,0] });
+
+private _aceHitArms = _hitleftarm + _hitrightarm;
+if(_aceHitArms > 0) then {
```
BaerMitUmlaut, line 20: Missing space

addons/medical_damage/functions/fnc_applyDamage.sqf
```diff
@@ -0,0 +1,220 @@
...
+
+if (alive _unit) then { _unit setDamage 0; };
+
+private _typeOfDamage = _ammo call FUNC(getTypeOfDamage);
+TRACE_6("Received hit",_unit,_source,_ammo,_typeOfDamage,_damage,_hitpointDamages);
+
+// No damage to hitpoints - special cases or environmental damage
+if (_hitpointDamages isEqualTo []) then {
+    switch (_typeOfDamage) do
+    {
```
BaerMitUmlaut, line 32: https://ace3mod.com/wiki/development/coding-guidelines.html#51-braces-placement

addons/medical_damage/functions/fnc_applyDamage.sqf
```diff
@@ -0,0 +1,220 @@
...
+            private _damageSelectionArray = [HITPOINT_INDEX_BODY, 1, HITPOINT_INDEX_LLEG, 1, HITPOINT_INDEX_RLEG, 1];
+            private _damagedBodyPart = selectRandomWeighted _damageSelectionArray;
+
+            [QEGVAR(medical,woundReceived), [
+                _unit, 
+                _damagedBodyPart, 
+                _damage, 
+                "unknown", 
+                [[_damagedBodyPart, _damage / (GVAR(structuralDamageToBodyPartCoeff) select _damagedBodyPart)]]
+                ]
```
BaerMitUmlaut, line 45: https://ace3mod.com/wiki/development/coding-guidelines.html#51-braces-placement

addons/medical_damage/functions/fnc_applyDamage.sqf
```diff
@@ -0,0 +1,220 @@
...
+        case "drowning": {
+            TRACE_5("Drowning",_unit,_source,_instigator,_damage,_newDamage);
+            [QEGVAR(medical,woundReceived), [
+                _unit, 
+                1, 
+                _damage, 
+                "unknown", 
+                [[1, _damage / GVAR(structuralDamageToBodyPartCoeff)#1]]
+                ]
+            ] call CBA_fnc_localEvent;
```
BaerMitUmlaut, line 57: https://ace3mod.com/wiki/development/coding-guidelines.html#51-braces-placement

addons/medical_damage/functions/fnc_getArmor.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+private _gear = [
+    _uniform,
+    vest _unit,
+    headgear _unit
+];
+
+private _rags = _gear joinString "$";
+private _var = format [QGVAR(armorCache%1), _rags];
+(localNamespace getVariable [_var, [""]]) params ["_prevRags", "_armor"];
```
BaerMitUmlaut, line 34: Why localNamespace?

BaerMitUmlaut (changes requested):
Without looking at functionality, code formatting definitely needs to be improved.

addons/medical_damage/functions/fnc_applyDamage.sqf
gandjustas: AFAIR just suitable constant.

addons/medical_damage/functions/fnc_getArmor.sqf
gandjustas: Why not? 
Reduced size of save game at least.

BaerMitUmlaut: Because it reduces readability and comprehensibility of the code. Save game size is nothing you should be concerned about.

kymckay:
> Changed hit detection from using ACE_HDBracket to Hit EH, less special cases to handle

This is a very fundamental and significant change, could you perhaps explain how you're facilitating this change in the code? I feel like there was a significant reason we did it the way we did and would like to see some reassurance as to why this change is an improvement. As it's a totally new file it's hard to directly compare the code changes themselves.

kymckay:
> Added death when inflicted structural damage is 1 or more (fixes #7305, #7898)

Similarly I'd like to see some justification for this with the ACE system in mind. Does this not bypass medical altogether? Is structural damage significant to the changes to damage handling - and if so in what way?

gandjustas:
Main reason of this changes - make medical not ignore structural damage.

There are lot of damage processing in vanilla relies on structural damage: 
- damage caused by different caliber of ammo with same hit and speed (AP vs Ball bullets, especially in RHS)
- damage caused to vehicle crew when vehicle hit by projectile
- death of units in exploded vehicles
- environmental/physx damage

If you ignore structural damage everything in list above will be broken, and it was broken in ace.

Unfortunately `ace_hdbracket` way is incompatible with structural damage handling, because structural damage can not fire HandleDamage for `ace_hdbracket` hitpoint. There is a special case for vehicle crash in ace, working incorrectly.

That's why I've changed hit detection to Hit EH.

gandjustas:
> > Added death when inflicted structural damage is 1 or more (fixes #7305, #7898)
> 
> 
> 
> Similarly I'd like to see some justification for this with the ACE system in mind. Does this not bypass medical altogether? Is structural damage significant to the changes to damage handling - and if so in what way?

When unit dies in vehicle crash (helicopter crash) or vehicle explodes with alive crew inside, crew receives only handleDamage event with selection="", hitPointIndex=-1, damage=1
And no other events are raised, even for ace_hdbracked hitpoint.

Structural damage = 1 is very high damage. Most of rifle bullets and grenades in radius on indirectHitRange causes less than 1 of structural damage to unit with body armor.

If we will kill unit with structural damage equal or more than 1 (like vanilla does) unit will die in crashes, from big explosions (bombs, artillery) and from high caliber girl (.50+).

gandjustas:
Structural damage handling isn't complicated. Hit detection is complicated. It was complicated before medical rewrite, it was complicated in rewrite, it's complicated in this PR. 

AFAIR I've tried to add structural damage handling without changing `ace_hdbracket` with no success.

addons/medical_damage/functions/fnc_getArmor.sqf
gandjustas: `localNamespace` is intended for caching

BaerMitUmlaut:
>There are lot of damage processing in vanilla relies on structural damage:
>- damage caused by different caliber of ammo with same hit and speed (AP vs Ball bullets, especially in RHS)

I have tested this and this seems to be true - however, this has the opposite effect of what I expected from the title of this PR. AP ammo causes higher structural damage when the hitpoint damage is the exact same. I have tested this with 7N6 ball and 7N22 AP ammo from RHS and a naked unit.

AP ammo should however cause the same or less damage, since the wound channel would be similar or smaller. I don't see how this is an improvement.

gandjustas:
Arma does not simulate armor and penetration for infantry. Armor values for vests just decreases incoming damage for hitpoint.

Caliber increases structural damage, not hitpoint damage. 

AP ammo will kill armored units faster than ball ammo, but same is true for unarmored units.

It's a limitation of Arma simulation, but there is no reason to ignore this in ace_medical making AP and ball ammo almost equal.

Btw ammo with higher caliber have greater chance to fully penetrate body and damage will be about 1/3 of hit*bulletSpeed/typicalSpeed.

BaerMitUmlaut:
I'm well aware that unit armor does not have the same kind of penetration simulation as vehicle armor. It's a limitation of the engine and I don't see how this makes it any better. I will run some further tests but this isn't very convincing so far.

gandjustas:
With structural damage there is much less bullets need to kill unit with bodyarmor with AP ammo, that is the main point.

pterolatypus:
>damage caused by different caliber of ammo with same hit and speed (AP vs Ball bullets, especially in RHS)

I did a bit of testing here myself and got different results. With the RHS M855 vs M855A1: they both have `hit=9` and `initSpeed=typicalSpeed` so at point blank the only difference should be from `caliber` - M855 is `0.44` whereas the A1 is `0.65`

When shooting at the same spot of the upper leg, the A1 produced _significantly_ less damage (less than half as much) both to the legs and in structural damage. In fact, the standard M855 was able to 1-shot the naked guy. In both cases the ratio of hitpoint damage:structural damage was the same.
I also fired those two rounds at the torso of a guy wearing a basic IOTV and the damage produced was identical (to 4 decimal places).

@BaerMitUmlaut did you spawn projectiles from script or shoot them manually? I'd like to build some sort of reproduceable testing methodology for this stuff as there are a lot of variables so it's hard to compare results.

BaerMitUmlaut:
I teleported them in a fired EH to the pelvis selection and set their velocity. I can check those bullets as well.

gandjustas:
> When shooting at the same spot of the upper leg, the A1 produced _significantly_ less damage (less than half as much) both to the legs and in structural damage. In fact, the standard M855 was able to 1-shot the naked guy. In both cases the ratio of hitpoint damage:structural damage was the same.
M855A1 fully penetrates leg and inflicts less damage, than M855.

> I also fired those two rounds at the torso of a guy wearing a basic IOTV and the damage produced was identical (to 4 decimal places).
It's nonlinear dependency between `hit`, `caliber` and `structural damage`.
Test results https://docs.google.com/spreadsheets/d/1LNcAdJKkRjDexnOx5Hz1vjp-56YcONyy2szch_gPhPQ/edit?usp=sharing

JNoXK:
Update:- Issue still persistent. 
Placed down a hatchback with a civilian inside and launched a Titan AT missile at the car. The car exploded instantly with the civilian jumping out unaffected.

gandjustas:
> Update:- Issue still persistent. 
> 
> Placed down a hatchback with a civilian inside and launched a Titan AT missile at the car. The car exploded instantly with the civilian jumping out unaffected.
> 

What is in the log?

JNoXK:
> > Update:- Issue still persistent.
> > Placed down a hatchback with a civilian inside and launched a Titan AT missile at the car. The car exploded instantly with the civilian jumping out unaffected.
> 
> What is in the log?

It says `22:13:25 Obsolete direct damage value` (when car was stationary and got hit)



I've noticed that the missiles worked fine on moving vehicles, but if stationary, the units take no damage and jump out.

JNoXK:
It seems if the car has a speed of 22km/h or more, the person inside consistently dies like they are supposed to, with only some rare exceptions during the test runs. But if the speed is anything under that value, the person is unharmed and ejects from the completely destroyed vehicle. When this happens, there are 3 or 4 entries in the log called `Obsolete direct damage value`

Mike-MF:

Issue #8020, opened 2020-11-27
AI opponents will not attack enemy unconscious


ZeusWrath:
- **Arma 3:** `2.00` [stable]
- **CBA:** `3` [stable, last version, Steam subscribe]
- **ACE3:** `3.13.5` [stable]
- **A3 Thermal Improvement** 
- **Bullet Casing** 
- **Tactical Position Ready** 
- **Magazine Simulation A3** 

**Description:**

I watched the AI opponents for several days, but I never saw them shoot at the unconscious. They'll just ignore them
It doesn't look natural, especially with [this problem](https://github.com/acemod/ACE3/issues/8019).

**Steps to reproduce:**

Places in the editor of 2 squads of opposing infantry fighters, turn on the camera and watch the unconscious units.

**Expected behavior:**
AI has to shoot at the unconscious. If someone doesn't like it, you can make a selection setting in Medical Settings. Many other mods have this settings

**Where did the issue occur?**
- Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

BaerMitUmlaut:
This is intentional.

ZeusWrath:
> This is intentional.

If the enemy does not shoot at the unconscious, it looks unnatural. **Why can't the setting be turned on / off?  Some mods have a setting - to shoot unconscious or not.**

BaerMitUmlaut:
Because an unconscious unit is not a threat.

ZeusWrath:
OK, Thank you for answer

gripen888:
I have got exactly opposite problem. Unconscious unit is sometimes not changed to civilian side and ai will continue to fire at patient. Sometimes 1/4 of mag, other times 2 mags in row.

BaerMitUmlaut:
Unconscious units do not change to civilian anymore. It's likely that this is an AI script interfering. If you don't use such a thing and can replicate this with just CBA and ACE, please create a new issue.

PR #8065, opened 2021-01-08
Fix AI Targeting Unconscious Enemies

### LINKED ISSUE #8041: AI fire at unconscious enemy
- **Arma 3**:  `2.00` [steam]
- **CBA:** `3.15.1` [steam workshop]
- **ACE3:** `3.13.5` [steam workshop]

**Description:**
AI will **sometimes** ignore the fact that enemy (AI or Player) is unconscious and fire at him. Sometime 2-3 shots. Other times 3-4 mags.
Like the AI is sometimes not satisfied with unconscious enemy and want him dead *just to be sure*.
This create very unpleasant experience when "**Player Fatal Injuries**" set to **Never**. You end up with player patient having enormous number of wounds.

**Steps to reproduce:**
1. Create any new mission and place yourselft as zeus.
2. Spawn 2 squads (enemy to each other). 
3. You should see this behavior in about 33% of firefights.

**Expected behavior:**
If AI really wana *hard-kill* enemy and is not satisfied with unconscious, there should be some addon settings that let me adjust this behavior or idealy turn it off.

- any SP or MP mission

**Log Files:**
- https://pastebin.com/Fi8uweL3

**Additional context:**
I also noticed that sometimes the unconscious soldier is recolored as civilian. Not sure if this hase anny connection tho.

Edit1 - updated log

frankplow:
- Fix #8041 (AI targeting unconscious enemies), by adding extra side-switching logic
- Remove antiquated wake-up logic from old medical
- Update `setHidden` status effect to make unit inaudible as well as invisible

The current system for stopping AI engaging unconscious units relies on the vanilla `setUnconscious` command, which is very buggy. For some reason, it fails to change the side of the unit to civilian despite ragdolling them in certain circumstances (see procedure at bottom). To resolve this, this PR adds redundant side-switching logic following `setUnconscious`. This is done using `switchToGroupSide` rather than `setCaptive` to avoid possible issues arising with unconscious handcuffed units.

The logic around https://github.com/acemod/ACE3/blob/fe544a274d043480e557903a891f6bfe02277d3e/addons/common/XEH_postInit.sqf#L353
 appears to be a holdover from the old medical and so has been removed for the risk of it interfering with the new medical system and to consolidate the code around unconsciousness logic to solely the medical modules.

In addition, this PR makes the `setHidden` status effect make the player inaudible as well as invisible. If a unit was moving when downed, they could still be detected by the enemy, undermining https://github.com/acemod/ACE3/blob/8e7f9b6db53a279432a2372c04be311807d1193b/addons/common/XEH_postInit.sqf#L64 I think the use of the `setHidden` status effect in relation to unconsciousness was originally added to help mitigate the issue when `setUnconscious` had failed, so this could likely be removed now.

A comprehensive reproduction procedure for the issue is as follows:
1. Start mission as zeus (with zeus module name `zeus`), with instant death disabled for AI and a single BLUFOR rifleman named `inf` and init field `zeus addCuratorEditableObjects [[this], true];`
2. Spawn REDFOR rifleman in front of BLUFOR rifleman
3. Wait for BLUFOR rifleman to neutralize REDFOR rifleman
4. Run `[inf, true] call ace_medical_fnc_setUnconscious;` in console
5. Observe that the faction of the BLUFOR rifleman does not change

Intended behaviour can be observed by not completing steps 2 and 3. The underlying issue with `setUnconscious` can be seen by running the game without mods and then executing `inf setUnconscious true;` in place of the command in step 4.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -53,12 +53,18 @@
 [QGVAR(setHidden), {
     params ["_object", "_set"];
     TRACE_2("setHidden EH",_object,_set);
+
     // May report nil. Default to factor 1.
     private _vis = [_object getUnitTrait "camouflageCoef"] param [0, 1];
+    private _aud = [_object getUnitTrait "audibleCoef"] param [0, 1];
+
     if (_set > 0) then {
-        if (_vis != 0) then {
+        if (_vis != 0 || _aud != 0) then {
             _object setVariable [QGVAR(oldVisibility), _vis];
             _object setUnitTrait ["camouflageCoef", 0];
+            _object setVariable [QGVAR(oldAudibleCoef), _aud];
+            _object setUnitTrait ["audibleCoef", 0];
+
             {
                 if (side _x != side group _object) then {
                     _x forgetTarget _object;
@@ -68,6 +74,8 @@
     } else {
         _vis = _object getVariable [QGVAR(oldVisibility), _vis];
         _object setUnitTrait ["camouflageCoef", _vis];
+        _aud = _object getVariable [QGVAR(oldAudibleCoef), _aud];
+        _object setUnitTrait ["audibleCoef", _vis];
     };
 }] call CBA_fnc_addEventHandler;
 
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -21,7 +21,7 @@
 [QGVAR(setStatusEffect), {_this call FUNC(statusEffect_set)}] call CBA_fnc_addEventHandler;
 ["forceWalk", false, ["ace_advanced_fatigue", "ACE_SwitchUnits", "ACE_Attach", "ACE_dragging", "ACE_Explosives", "ACE_Ladder", "ACE_Sandbag", "ACE_refuel", "ACE_rearm", "ACE_Trenches", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
 ["blockSprint", false, ["ace_advanced_fatigue", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
-["setCaptive", true, [QEGVAR(captives,Handcuffed), QEGVAR(captives,Surrendered)]] call FUNC(statusEffect_addType);
+["setCaptive", true, [QEGVAR(captives,Handcuffed), QEGVAR(captives,Surrendered), QEGVAR(medical_status,Unconscious)]] call FUNC(statusEffect_addType);
 ["blockDamage", false, ["fixCollision", "ACE_cargo"]] call FUNC(statusEffect_addType);
 ["blockEngine", false, ["ACE_Refuel"]] call FUNC(statusEffect_addType);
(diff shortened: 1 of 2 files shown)


frankplow:
I've just found this comment: https://github.com/acemod/ACE3/issues/8020#issuecomment-742814840. As far as I can tell, units still do change side when they go unconscious - this seems to be the intended behaviour of `setUnconscious`, which ACE uses without any logic to change the units side back. This is also the behaviour you see in the intended behaviour procedure described above. The only time I get the issue of the AI 'magdumping' on an unconscious enemy is when `setUnconscious` fails to switch the side of the unit.

The comment https://github.com/acemod/ACE3/blob/fe544a274d043480e557903a891f6bfe02277d3e/addons/medical_status/functions/fnc_setUnconsciousState.sqf#L31
says that it is maybe instead the job of `setHidden` alone to prevent enemies from engaging unconscious units, hence the improvement to it in this PR, but even with that improvement (i.e. commit 399525d36afd3d2b48fb621cf880e0d2e17a03dc only from this PR), the issue is still present, even without complicating things with AI mods which fiddle with `knowsAbout`.

FrostedVert:
Just out of curiosity, I’ve had good results with the AI surrendering and treating wounded combatants using the [ZluskeN Hors de combat](https://steamcommunity.com/sharedfiles/filedetails/?id=1139830447) mod. Does it work much differently from the existing system, and could any innovations in the mod be integrated with the changes being discussed here?

BaerMitUmlaut:
Changing the group of the unit to a civilian group was specifically avoided with the medical rewrite, and should be avoided here, too. It causes too many unwanted side effects.

BaerMitUmlaut:
@FrostedVert adding custom AI behaviour is beyond the scope of this PR.

frankplow:
@BaerMitUmlaut Recent commit changes over to using the `setCaptive` status effect rather than changing groups. I've done some testing with combinations of unconscious/handcuffed/surrendering units and it seems robust.

I also removed the `setHidden` status effect as it's no longer necessary with these changes and using it runs code for every unit.

addons/common/XEH_postInit.sqf
```diff
@@ -350,14 +358,6 @@ addMissionEventHandler ["PlayerViewChanged", {
 [QGVAR(displayTextStructured), {_this call FUNC(displayTextStructured)}] call CBA_fnc_addEventHandler;
 [QGVAR(displayTextPicture), {_this call FUNC(displayTextPicture)}] call CBA_fnc_addEventHandler;
 
-["ace_unconscious", {
```
frankplow, line 353: Just had a quick look over this myself again, I think deleting this EH was a mistake and may cause bugs with loading unconscious units into vehicles. I don't have access to a PC capable of running ArmA at the moment though, so I can't verify that.

frankplow:
I've had the time to do a little testing on this and, as per my suspicions above, 9f0c5d9f4407ee8f9302a7701be6867742a4c706 introduced some bugs when a unit woke up after being loaded into a vehicle.

In testing, I found some other strange behaviour and bugs regarding loading casualties - it's one of the last places in the codebase involving explicit side switching. I figure any changes there are best left to another commit though.

kymckay:
We intentionally replaced similar if not the same logic in the rewrite because it was causing some kind of issues (sadly I forget exactly why, we should dig up the old PRs/discussion to review).

PR #8070, opened 2021-01-14
Repair - Add enable setting

Dystopian:
- add `Enable` setting to `Repair` based on #7040 but with some improvements and without item replacement;
- add support for vehicles without ACE repair config;
- remove outdated config for `B_Truck_01_fuel_F`;
- fix `ACE_isRepairFacility` boolean value handling, see #6278 and https://github.com/acemod/ACE3/blob/24b1a71fbbc7a458435461fff19d44357e7eeaa5/docs/wiki/framework/repair-framework.md#L40
- optimize and improve some things.


addons/repair/CfgEden.hpp
```diff
@@ -1,7 +1,8 @@
+#define VANILLA_ISREPAIRVEHICLE (if (getRepairCargo _this > 0) then {1} else {0})
```
veteran29, line 1:
```suggestion
#define VANILLA_ISREPAIRVEHICLE (parseNumber (getRepairCargo _this > 0))
```

addons/repair/CfgEden.hpp
```diff
@@ -78,7 +79,7 @@ class Cfg3DEN {
                         displayName = CSTRING(editorLoadedWheels);
                         tooltip = CSTRING(editorLoadedWheels_tooltip);
                         property = QGVAR(editorLoadedWheels);
-                        defaultValue = "[0,1] select (_this isKindOf 'Car')"; // must match pre init script
+                        defaultValue = "[0,1] select (_this isKindOf 'Car')"; // must match post init script
```
veteran29, line 82:
```suggestion
                        defaultValue = "parseNumber (_this isKindOf 'Car')"; // must match post init script
```

addons/repair/CfgEden.hpp
```diff
@@ -69,7 +70,7 @@ class Cfg3DEN {
                         property = QGVAR(editorLoadedTracks);
                         control = "Edit";
                         expression = "_this setVariable ['%s',_value];";
-                        defaultValue = "[0,1] select (_this isKindOf 'Tank')"; // must match pre init script
+                        defaultValue = "[0,1] select (_this isKindOf 'Tank')"; // must match post init script
```
veteran29, line 73:
```suggestion
                        defaultValue = "parseNumber (_this isKindOf 'Tank')"; // must match post init script
```

veteran29 (dismissed):
Minor optimization.

addons/repair/functions/fnc_addRepairActions.sqf
```diff
@@ -15,7 +15,7 @@
  * Public: No
  */
 
-if (!hasInterface) exitWith {};
+if (!hasInterface || !GVAR(enabled)) exitWith {};
```
LinkIsGrim, line 18:
```suggestion
if !(hasInterface && {GVAR(enabled)}) exitWith {};
```

addons/repair/initSettings.sqf
LinkIsGrim, line 1: Solve merge conficts. Should have parity with master here except for `QGVAR(enabled)`.

addons/repair/XEH_postInit.sqf
```diff
@@ -18,3 +18,62 @@
...
+    if (local _vehicle && {getRepairCargo _vehicle > 0}) then {
+        _vehicle setRepairCargo 0;
+    };
+}, true, ["Man"], true] call CBA_fnc_addClassEventHandler;
+
+["CAManBase", "InitPost", {
+    params ["_unit"];
+    if (local _unit && {_unit getUnitTrait "engineer"}) then {
+        _unit setUnitTrait ["engineer", false];
+        if (-1 isEqualTo (_unit getVariable ["ACE_IsEngineer", -1])) then {
```
LinkIsGrim, line 41:
```suggestion
        if ((_unit getVariable ["ACE_IsEngineer", -1]) isEqualTo -1) then {
```

addons/repair/CfgVehicles.hpp
LinkIsGrim, line 1: Merge upstream master and remove anything that was added after this was last updated.

LinkIsGrim (changes requested):
`setUnitTrait ["Engineer", true]` by any third party will break things, but that's a compromise we'll have to make for no item replacements.

addons/repair/functions/fnc_isInRepairFacility.sqf
LinkIsGrim, line 1: Also needs to be updated after #8417.

addons/repair/XEH_postInit.sqf
rautamiekka, line 41: This is better:
```suggestion
        if !(_unit getVariable ["ACE_IsEngineer", false]) then {
```

LinkIsGrim: No, unit variable can be set to false while maintaining engineer trait. -1 is more appropriate as it's only valid if the variable hasn't been set.

rautamiekka: In that sense yeah, since would broadcast for nothing.

Dystopian:
Will do also some tests after upstream merging.

addons/repair/XEH_postInit.sqf
```diff
@@ -18,3 +18,65 @@
...
+}, true, ["Man"], true] call CBA_fnc_addClassEventHandler;
+
+["CAManBase", "InitPost", {
+    params ["_unit"];
+    if !(local _unit && {_unit getUnitTrait "engineer"}) exitWith {};
+        _unit setUnitTrait ["engineer", false];
+    if (_unit getVariable ["ACE_IsEngineer", -1] isEqualTo -1) then {
+            _unit setVariable ["ACE_IsEngineer", true, true];
+        };
+    TRACE_3("setUnitTrait",_unit,typeOf _unit,_unit getUnitTrait "engineer");
```
LinkIsGrim, line 45:
```suggestion
    if !(local _unit && {_unit getUnitTrait "engineer"}) exitWith {};

    _unit setUnitTrait ["engineer", false];
 
    if (_unit getVariable ["ACE_IsEngineer", -1] isEqualTo -1) then {
        _unit setVariable ["ACE_IsEngineer", true, true];
    };
    TRACE_3("setUnitTrait",_unit,typeOf _unit,_unit getUnitTrait "engineer");
```

addons/repair/XEH_postInit.sqf
```diff
@@ -18,3 +18,65 @@
...
+
+// load tracks and wheels
+if (isServer) then {
+    private _fnc_addSpareItems = {
+        if (!GVAR(addSpareParts)) exitWith {};
+        params ["_vehicle"];
+
+        private _spareTracks = _vehicle getVariable QGVAR(editorLoadedTracks);
+        if (isNil "_spareTracks") then {
+            _spareTracks = [0, 1] select (_vehicle isKindOf "Tank"); // must match eden attribute default
```
LinkIsGrim, line 65:
```suggestion
            _spareTracks = parseNumber (_vehicle isKindOf "Tank"); // must match eden attribute default
```

addons/repair/XEH_postInit.sqf
```diff
@@ -18,3 +18,65 @@
...
+        };
+        if (_spareTracks > 0) then {
+            [_vehicle, _spareTracks, "ACE_Track"] call FUNC(addSpareParts);
+        };
+
+        private _spareWheels = _vehicle getVariable QGVAR(editorLoadedWheels);
+        if (isNil "_spareWheels") then {
+            _spareWheels = [0, 1] select (_vehicle isKindOf "Car"); // must match eden attribute default
```
LinkIsGrim, line 73:
```suggestion
            _spareWheels = parseNumber (_vehicle isKindOf "Car"); // must match eden attribute default
```

LinkIsGrim (approved):

addons/repair/XEH_postInit.sqf
```diff
@@ -1,9 +1,9 @@
 #include "script_component.hpp"
 
+if !GVAR(enabled) exitWith {};
```
PabstMirror, line 3: need to wrap it all in
```
["CBA_settingsInitialized", {

```

addons/repair/XEH_postInit.sqf
```diff
@@ -1,9 +1,13 @@
 #include "script_component.hpp"
 
+if !(EGVAR(common,settingsInitFinished)) exitWith {
+    EGVAR(common,runAtSettingsInitialized) pushBack [compileScript [__FILE__], _this];
+};
```
Dystopian, line 5: May I use such code instead of moving the whole file into `CBA_settingsInitialized` frame? Also this can be used:
```suggestion
if (param [0, true]) exitWith {
    ["CBA_settingsInitialized", {false call compileScript [__FILE__]}] call CBA_fnc_addEventHandler;
};
```

LinkIsGrim: I'm not a fan of either solution, would rather just wrap everything in the EH.

PabstMirror: yeah, it might make the diff ugly but I think the cleanest way is to just tab everything over in the eh

LinkIsGrim:
@mharis001 [ZEN Engineer Attribute](https://github.com/zen-mod/ZEN/blob/4dad54bc299cb44d120b7196f28b78fd7f55dba0/addons/attributes/initAttributes.sqf#L956) will need to be updated with this.

PR #8077, opened 2021-01-21
fix common_fnc_setAimCoef by making AF use it

nomisum:
- fix common_fnc_setAimCoef by making AF use it

### DIFF
--- a/addons/advanced_fatigue/functions/fnc_handleEffects.sqf
+++ b/addons/advanced_fatigue/functions/fnc_handleEffects.sqf
@@ -96,4 +96,4 @@ if (_overexhausted) then {
 _unit setVariable [QGVAR(aimFatigue), _fatigue];
 
 private _aimCoef = [missionNamespace, "ACE_setCustomAimCoef", "max"] call EFUNC(common,arithmeticGetResult);
-[_unit, "ACE_setCustomAimCoef", _aimCoef, true] call EFUNC(common,setAimCoef);
+[_unit, QUOTE(ADDON), _aimCoef, true] call EFUNC(common,setAimCoef);
--- a/addons/advanced_fatigue/functions/fnc_handleEffects.sqf
+++ b/addons/advanced_fatigue/functions/fnc_handleEffects.sqf
@@ -96,4 +96,4 @@ if (_overexhausted) then {
 _unit setVariable [QGVAR(aimFatigue), _fatigue];
 
 private _aimCoef = [missionNamespace, "ACE_setCustomAimCoef", "max"] call EFUNC(common,arithmeticGetResult);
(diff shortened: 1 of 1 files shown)


jonpas (approved):

addons/advanced_fatigue/functions/fnc_handleEffects.sqf
```diff
@@ -96,4 +96,4 @@ if (_overexhausted) then {
 _unit setVariable [QGVAR(aimFatigue), _fatigue];
 
 private _aimCoef = [missionNamespace, "ACE_setCustomAimCoef", "max"] call EFUNC(common,arithmeticGetResult);
-_unit setCustomAimCoef _aimCoef;
+[_unit, "ACE_setCustomAimCoef", _aimCoef, true] call EFUNC(common,setAimCoef);
```
BaerMitUmlaut, line 99: "ACE_setCustomAimCoef" is not a unique ID for AF. Would recommend `QUOTE(ADDON)` or something similar.

BaerMitUmlaut (changes requested):

addons/advanced_fatigue/functions/fnc_handleEffects.sqf
nomisum: pushed your suggestion

PabstMirror:
Medical and AdvFatigue both use
```
[missionNamespace, "ACE_setCustomAimCoef", QUOTE(ADDON), {
    //code that returns a number
}] call EFUNC(common,arithmeticSetSource);
```
so EFUNC(common,setAimCoef) is redundant as we already have the max

IMHO we should just drop setAimCoef as it's non-functional now
mod/mission makers can use "ACE_setCustomAimCoef"

we change it to be a wrapper like:
https://github.com/acemod/ACE3/blob/4669ec961d47eef78a14034590a569471366a5bf/addons/common/functions/fnc_setAimCoef.sqf

but the params don't line up  
- _unit is assumed to be player
- I don't like `compile str _setting`

BaerMitUmlaut:
It's pretty bad how inconsistent we handle these things. `arithmeticSetSource` is a decent system, but not ideal either because one component still has the authority because at least one needs to use the actual engine function.

Ideally we would have a centralised system in `common` - something like the status effects system.

PabstMirror:
We could just go back to using https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_setAimCoef.sqf
it gets the max and applies it, seems to do everything we want

jonpas:
@PabstMirror @BaerMitUmlaut status?

Mike-MF:
@PabstMirror @BaerMitUmlaut Re-open if you want this.

PabstMirror:
replaced with https://github.com/acemod/ACE3/pull/9226 I think

PR #8079, opened 2021-01-22
(WIP) hiding all actions under CfgActions

abramsba:
- Completely disable vanilla action menu items from appearing

I would like to disable vanilla actions in favor of ACE interaction menu. The existing configuration file only overruled a few of the actions.

### DIFF
--- a/optionals/noactionmenu/CfgActions.hpp
+++ b/optionals/noactionmenu/CfgActions.hpp
@@ -1,24 +1,1108 @@
-
 class CfgActions {
-    // to pistol
     class None;
-    class HandGunOn: None {
+    class ActivateBreathingBomb: None {
         show = 0;
+        showIn3D = 0;
+        showWindow = 0;
     };
-    // to rifle
+    class ActivateFins: None {
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
+    class ActivateMine: None {
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
+    class ActiveSensorOff: None {
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
+    class ActiveSensorOn: None {
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
+    class AddBag: None {
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
+    class AIAssemble: None {
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
+    class APUOff: None {
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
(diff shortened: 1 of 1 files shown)


Italiandrache:
Why? Would you really prefer to open a door, or the inventory, using ace interactions menu over the vanilla actions?

I feel that scrolling a wheel and pressing the spacebar is far less clunky the ACE's interaction menu, but it's just my personal preference. Maybe consider leaving it as an option.

bux:
Ctrl + Space opens a door in ACE ...
That's quite the opposite of clunky for my taste

PabstMirror:
I'm not sure if there are keybinds or ace-interaction for all of these actions?

abramsba:
> Why? Would you really prefer to open a door, or the inventory, using ace interactions menu over the vanilla actions?
> That's quite the opposite of clunky for my taste

There are keybinds available for specific actions, opening door being one of them. I personally think the action menu implementation in vanilla Arma is terrible. The worst offending thing I can think of are when actions become selected automatically in custom vehicles and such, or accidentally performing some action because some context selected it automatically. 

> I'm not sure if there are keybinds or ace-interaction for all of these actions?

If I knew which ones were relevant I could come up with ace action replacements for them, but I haven't been able to find which ones they were on my own. 

I don't quite understand though the why questions, why is there a PBO included for this if nobody likes how this sounds?

commy2:
The component exists to hide actions that are implemented in ACE instead. But hiding all actions, even those you don't know "are relevant" seems like a bad idea to me.

optionals/noactionmenu/CfgActions.hpp
```diff
@@ -1,24 +1,1108 @@
...
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
+    class ListLeftVehicleDisplay: None {
+        show = 0;
+        showIn3D = 0;
+        showWindow = 0;
+    };
+    class ListPrevLeftVehicleDisplay: None {
```
commy2, line 463: Also, some of these inheritance trees are wrong. `ListPrevLeftVehicleDisplay` inherits from `ListPrevRightVehicleDisplay`, not "`None`". Just as an example. Don't feel like checking all of these.

abramsba:
> The component exists to hide actions that are implemented in ACE instead. But hiding all actions, even those you don't know "are relevant" seems like a bad idea to me.

That makes sense. Is it documented which actions have already been replaced?

optionals/noactionmenu/CfgActions.hpp
abramsba: Added some classes that I missed and corrected for this.

commy2:
No.

abramsba:
I'll close this then and package it into something else. I made it under the incorrect assumption that this should have blocked the entire action menu.

PR #8083, opened 2021-01-26
Medical - Show "Inventory" action on unconscious units

### LINKED ISSUE #8024: Medical - Show "Inventory" action on unconscious units
With recent enhancement of body bag interaction always showing, it is one less obvious thing that someone is dead. And our entire goal is to make unconscious and death harder to distinguish instantly (new random animations and rag-doll transitions are great).

One last part of this identified in my community was the "Inventory" action (scroll-wheel) that shows up on dead people. Icon tends to appear without scrolling and is a dead giveaway someone is dead, can't avoid it.

Removing that annoyance would be last general step to bring it to a state where you would have to perform at least some basic checks in medical, with the ability to perform body bag action as a longer check (with default settings) or even without any ability and solely relying on medic's judgement (allow unconscious body bag setting enabled).

---

Removing the "Inventory" action is not feasible as it's the same one that's used on crates etc. and we ideally don't want to outright remove that (it is often easier to use than ACE interaction as it's usually the only option available).

@BaerMitUmlaut suggested to simply add "Inventory" action to unconscious units as well (`addAction`), and that would be the simplest change. It could open the normal inventory or Disarm inventory, doesn't matter. But it has to make sure same translations and parameters are used (eg. distance of 3.5m).

Additionally, on dead units, Disarm inventory shows up as well, so it would be ideal if that also shows up on unconscious. The main difference is that you can drag items to the ground with Disarm inventory.

_I am opening this because I mostly don't have time to implement this myself, but a contributor might find it interesting._

Salluci:
- Close #8024.
- Adds "Inventory" action to unconscious units. Uses base Arma 3 inventory, action looks identical to vanilla action.

### DIFF
--- a/addons/medical_status/XEH_preInit.sqf
+++ b/addons/medical_status/XEH_preInit.sqf
@@ -26,14 +26,17 @@ PREP_RECOMPILE_END;
 addMissionEventHandler ["EntityKilled", {_this call FUNC(handleKilledMission)}];
 
 //Add Inventory action to uncon units
-["CAManBase", "init",
-{
-  params ["_unit"];
-  private _id = _unit addAction [localize "STR_ACTION_GEAR", {
-    params ["_target","_caller","_actionId","_arguments"];
-    _caller action ["Gear", _target];
-  },nil,5.1,true,true,"", "(_target getVariable ['ACE_isUnconscious',false] && {alive _target})",3.5];
-  _unit setUserActionText [_id,localize "STR_ACTION_GEAR","<img image='\A3\ui_f\data\igui\cfg\actions\gear_ca.paa' size='2.5' shadow=2 />"];
-},nil,nil,true] call CBA_fnc_addClassEventHandler;
+["CAManBase", "init", {
+    params ["_unit"];
+
+    private _id = _unit addAction ["", {
+        params ["_target", "_caller", "_actionId", "_arguments"];
+
+        _caller action ["Gear", _target];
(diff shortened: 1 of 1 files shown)


BaerMitUmlaut (changes requested):
Please follow our [code style guidelines](https://ace3mod.com/wiki/development/coding-guidelines.html#5-code-style).

addons/medical_status/XEH_preInit.sqf
```diff
@@ -25,4 +25,15 @@ PREP_RECOMPILE_END;
 
 addMissionEventHandler ["EntityKilled", {_this call FUNC(handleKilledMission)}];
 
+//Add Inventory action to uncon units
+["CAManBase", "init",
+{
+  params ["_unit"];
+  private _id = _unit addAction [localize "STR_ACTION_GEAR", {
```
veteran29, line 32: No point executing localize two times.
```suggestion
  private _id = _unit addAction ["", {
```

veteran29 (changes requested):

BaerMitUmlaut (approved):

veteran29 (approved):

PabstMirror (approved):

PabstMirror:
```
Run actions/upload-artifact@v2-preview
  with:
    name: ace3-23f6e044a073cabd39523b9584e51ed0b7129851-nobin
    path: releases/ace3_*.zip
With the provided path, there will be 1 files uploaded
Error: read ECONNRESET
```
failed check not-relevant

PR #8089, opened 2021-01-30
Arsenal - Less magic numbers

veteran29:
- title

I got confused what is what while debugging something inside the component and decided that it would be good to have more descriptive defines for these indices


BrettMayson (approved):

commy2:
If most of these:

```sqf
} foreach (GVAR(virtualItems) select 4);
```
are changed to:
```sqf
#define IDX_VIRT_UNIFORM 4

} foreach (GVAR(virtualItems) select IDX_VIRT_UNIFORM);
```
why not go all the way and change them to this:
```sqf
#define IDX_VIRT_UNIFORM 4
#define VIRT_UNIFORM (GVAR(virtualItems) select IDX_VIRT_UNIFORM)

} foreach VIRT_UNIFORM;
```
instead?

veteran29:
> If most of these:
> ...

I was thinking about this but it's more work and changing of indexes to defines adds "enough" (that depends on point of view ofc.) readability already.

Will see if I will have enough time to do both.

jonpas (approved):
Is an improvement. I think @commy2's suggestion can still be done, but it doesn't have to delay this PR.

PabstMirror:
I not a big fan of code in macros as it can hide the cost and make things less clear

- `#define IDX_VIRT_WEAPONS 0` 
has no cost, it's just showing the value of the number
- `#define VIRT_UNIFORM (GVAR(virtualItems) select IDX_VIRT_UNIFORM)` 
has a runtime cost in the `select` which might not be apparent

PabstMirror (approved):

PR #8102, opened 2021-02-11
Arsenal - Fix Zeus UI lockup after closing arsenal

neilzar:
- After opening and closing the arsenal on a unit and then deleting the unit, the UI would lock up as soon as another dialog was opened. This fixes that issue. 
- Fixes https://github.com/zen-mod/ZEN/issues/152


addons/arsenal/functions/fnc_onArsenalClose.sqf
```diff
@@ -69,6 +69,8 @@ camDestroy GVAR(camera);
 if (!isNil QGVAR(moduleUsed)) then {
     GVAR(moduleUsed) = nil;
     objNull remoteControl GVAR(center);
+    player switchCamera cameraView;
+    GVAR(center) setVariable ["bis_fnc_moduleRemoteControl_owner", nil, true];
```
mharis001, line 73: Unnecessary. This variable is never set when the arsenal is opened in `moduleUsed` mode.

addons/arsenal/functions/fnc_onArsenalClose.sqf
```diff
@@ -69,6 +69,7 @@ camDestroy GVAR(camera);
 if (!isNil QGVAR(moduleUsed)) then {
     GVAR(moduleUsed) = nil;
     objNull remoteControl GVAR(center);
+    player switchCamera cameraView;
```
commy2, line 72: `player` or `ACE_player` ? How does this relate to L62?

neilzar: L62 is only in 3den and when not using Zeus, while the bug happens in any form of Zeus such as on dedicated servers.

neilzar:
I have moved the cameraview reset so only one is needed in the file. I have tested in 3den, Zeus and in normal gameplay and I never noticed any issues. I am not sure why it was only reset after exiting the arsenal when not a curator in normal gameplay before, maybe the maintainers can do some additional checks if this current alteration doesn't create/bring back any issues that I am unaware off.

mharis001 (approved):

PR #8103, opened 2021-02-11
Interaction - Add interaction with terrain objects

Dystopian:
- title;
- add dev function for replacing any terrain object with hotkey.

Terrain objects with `ace_interaction_replaceTerrainObject` config value == 1 are replaced with created object when interaction menu is being opened near them. This feature can be useful at least for `dragging`, `acex sitting` and `acex field_rations`.

### DIFF
--- a/addons/interaction/XEH_postInit.sqf
+++ b/addons/interaction/XEH_postInit.sqf
@@ -77,7 +77,24 @@ ACE_Modifier = 0;
     };
 }] call CBA_fnc_addEventHandler;
 
+if (isServer) then {
+    [QGVAR(replaceTerrainObject), FUNC(replaceTerrainObject)] call CBA_fnc_addEventHandler;
+};
+
 if (!hasInterface) exitWith {};
+
+[QEGVAR(interact_menu,renderNearbyActions), {
+    if !GVAR(interactWithTerrainObjects) exitWith {};
+    {
+        if (!isObjectHidden _x && {_x getVariable [QGVAR(terrainObjectNotReplaced), true]} && {typeOf _x isEqualTo ""}) then {
+            private _model = getModelInfo _x select 1;
+            private _class = GVAR(replaceTerrainClasses) getVariable [_model, ""];
+            if (_class isEqualTo "") exitWith {};
+            _x setVariable [QGVAR(terrainObjectNotReplaced), false];
+            [QGVAR(replaceTerrainObject), [_x, _class]] call CBA_fnc_serverEvent;
+        };
+    } forEach nearestTerrainObjects [ACE_player, [], 5, false];
+}] call CBA_fnc_addEventHandler;
 
 GVAR(isOpeningDoor) = false;
 
--- a/addons/interaction/XEH_preInit.sqf
+++ b/addons/interaction/XEH_preInit.sqf
@@ -6,11 +6,22 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+#include "initSettings.sqf"
+
 DFUNC(repair_Statement) = { // moved from config because of build problems
     TRACE_1("repair_Statement",_this);
     {
         if (_x isKindOf 'AllVehicles' && {!(_x isKindOf 'Man')}) then { _x setDamage 0; };
     } forEach (curatorSelected select 0)
+};
+
+if hasInterface then {
+    GVAR(replaceTerrainClasses) = call CBA_fnc_createNamespace;
+    private _cacheReplaceTerrainArray = call (uiNamespace getVariable [QGVAR(cacheReplaceTerrainClasses), {[[],[]]}]);
+    _cacheReplaceTerrainArray params ["_cacheReplaceTerrainModels", "_cacheReplaceTerrainClasses"];
+    {
+        GVAR(replaceTerrainClasses) setVariable [_x, _cacheReplaceTerrainClasses select _forEachIndex];
+    } forEach _cacheReplaceTerrainModels;
 };
 
 ADDON = true;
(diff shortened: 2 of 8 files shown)


jonpas:
What happens when terrain object is a large building and somebody is inside it?

Even on small objects, any chance for offsets, blowing things up inadvertendly etc.?

Dystopian:
> What happens when terrain object is a large building and somebody is inside it?

The largest "building" I tested and implemented is `Land_BusStop_02_shelter_F` (the bus stop on Chernarus 2020 and, I suppose, on Livonia), see acemod/ACEX#255. It works OK and nobody even see object replacing (including those who's inside it). Anyway I don't think replacing of large buildings is a good idea.


> Even on small objects, any chance for offsets, blowing things up inadvertendly etc.?

There is a small chance of objects colliding. At least I saw plastic chairs falling to ground when replaced. But I don't think it's a big issue. And also the setting is disabled by default.

addons/interaction/XEH_preInit.sqf
```diff
@@ -15,4 +15,8 @@ DFUNC(repair_Statement) = { // moved from config because of build problems
     } forEach (curatorSelected select 0)
 };
 
+if hasInterface then {
```
jonpas, line 18:
```suggestion
if (hasInterface) then {
```

addons/interaction/initSettings.sqf
```diff
@@ -29,3 +29,11 @@
     format ["ACE %1", LLSTRING(DisplayName)],
     true
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(interactWithTerrainObjects), "CHECKBOX",
+    "str_a3_modules_moduleomquest_defend_f_attributes_useterrainobject0",
+    format ["ACE %1", LLSTRING(DisplayName)],
```
jonpas, line 36: Should add a description that hints at what might go wrong with this.

jonpas (commented):
I think that's pretty neat then!

addons/interaction/initSettings.sqf
Dystopian: How about `Warning: some objects can collide with other objects.`?

jonpas: Works for me, pinging @mharis001.

addons/interaction/XEH_preStart.sqf
```diff
@@ -1,3 +1,25 @@
 #include "script_component.hpp"
 
 #include "XEH_PREP.hpp"
+
+if !hasInterface exitWith {};
```
jonpas, line 5:
```suggestion
if (!hasInterface) exitWith {};
```

addons/interaction/XEH_preStart.sqf
```diff
@@ -1,3 +1,25 @@
...
+
+private _replaceTerrainClasses = QUOTE( \
+    getNumber (_x >> 'scope') == 2 \
+    && {getNumber (_x >> QQGVAR(replaceTerrainObject)) > 0} \
+) configClasses (configFile >> "CfgVehicles");
```
jonpas, line 10: SQF Validator thinks it's missing a semi-colon somewhere here. :thinking:

Dystopian: already reported in #5979

addons/interaction/XEH_postInit.sqf
```diff
@@ -77,10 +77,30 @@ ACE_Modifier = 0;
...
+if (isServer) then {
+    [QGVAR(replaceTerrainObject), FUNC(replaceTerrainObject)] call CBA_fnc_addEventHandler;
+};
+
 if (!hasInterface) exitWith {};
 
 GVAR(isOpeningDoor) = false;
 
+[QEGVAR(interact_menu,renderNearbyActions), {
+    if !GVAR(interactWithTerrainObjects) exitWith {};
```
mharis001, line 89:
```suggestion
    if (!GVAR(interactWithTerrainObjects)) exitWith {};
```

mharis001 (approved):
Code looks good to me.

addons/interaction/XEH_preStart.sqf
```diff
@@ -1,3 +1,25 @@
 #include "script_component.hpp"
 
 #include "XEH_PREP.hpp"
+
+if (!hasInterface) exitWith {};
+
+private _replaceTerrainClasses = QUOTE( \
+    getNumber (_x >> 'scope') == 2 \
+    && {getNumber (_x >> QQGVAR(replaceTerrainObject)) > 0} \
+) configClasses (configFile >> "CfgVehicles");
```
jonpas, line 10:
```suggestion
private _replaceTerrainClasses = QUOTE(
    getNumber (_x >> 'scope') == 2
    && {getNumber (_x >> QQGVAR(replaceTerrainObject)) > 0}
) configClasses (configFile >> "CfgVehicles");
```
This should work just fine as it is not a macro definition.

Dystopian: you are right, it works. I also swapped conditions and got -10ms :)

jonpas (approved):

PabstMirror (approved):
This is a little rough because of physX.
On stratis beach the "sun chairs" went flying into the ocean because of collisions.
**note**: I forgot to test, but maybe adding a frame delay between hide and create would help?

but it also adds a lot of functionality, it is default off and the setting has a warning, so I think it's ok to merge

Dystopian:
> On stratis beach the "sun chairs" went flying into the ocean because of collisions.
> **note**: I forgot to test, but maybe adding a frame delay between hide and create would help?

I found these chairs and now I hate them. Neither delay (even for some seconds) nor any of `[fixCollision, fixPosition, fixFloating]` functions help.

Bug looks VERY strange. Repro:
1. place player on Marina Bay beach exactly near single chair but not close to other 2 chairs:
interaction with terrain objects 1
1. press interaction key, observe single chair replacing
1. go close enough to other 2 chairs and press interaction
1. **sometimes** observe chairs strange behavior: floating, flipping, flying for some meters.

If skip step 2 I can't repro.

PR #8106, opened 2021-02-11
Wirecutter - Replace fences with destroyed models when cut

BaerMitUmlaut:
- Add some more fence types for wirecutting
- Replace fences with destroyed models when cut where possible

Some large fences are replaced by multiple smaller fences, sometimes the whole net is removed and just a pole remains - I tried to make it look as realistic as possible. Unfortunately I could not find a suitable replacement model for one of the most used fences on Altis and Stratis (the razor wire fence just in front of the radar dome in the screenshots below).

Before             |  After
:-:|:-:
 | 
 |

### DIFF
--- a/addons/logistics_wirecutter/XEH_preInit.sqf
+++ b/addons/logistics_wirecutter/XEH_preInit.sqf
@@ -6,26 +6,28 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
-GVAR(replacements) = createHashMapFromArray [
-    ["gameprooffence_01_l_5m_f.p3d", [["Land_GameProofFence_01_l_d_F", [0, 0, 0], 0]]],
-    ["indfnc_3_f.p3d", [["Land_IndFnc_3_D_F", [0.039, -0.221, 0], 0]]],
-    ["indfnc_3_hole_f.p3d", [["Land_IndFnc_3_D_F", [0.042, -0.252, 0], 0]]],
-    ["indfnc_9_f.p3d", [["Land_IndFnc_3_F", [-3, -0.06, 0], 0], ["Land_IndFnc_3_D_F", [0.039, -0.281, 0], 0], ["Land_IndFnc_3_F", [3, -0.06, 0], 0]]],
-    ["indfnc_corner_f.p3d", [["Land_IndFnc_3_D_F", [0.116, -0.223, 0], 0]]],
-    ["mil_wiredfence_f.p3d", [["Land_Mil_WiredFenceD_F", [0, 0, 0], 0]]],
-    ["net_fence_8m_f.p3d", [["Land_Net_FenceD_8m_F", [0, 0.1, 0], 0]]],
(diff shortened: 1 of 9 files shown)


addons/logistics_wirecutter/functions/fnc_destroyFence.sqf
```diff
@@ -0,0 +1,41 @@
...
+
+// If fence cannot be replaced with destroyed model, just knock it over
+if !(_fenceModel in GVAR(replacements)) exitWith {
+    _fence setDamage 1;
+};
+
+// Remove old fence
+if ([_fence] call CBA_fnc_isTerrainObject) then {
+    _fence setDamage 1;
+    _fence hideObjectGlobal true;
```
mharis001, line 29: `hideObjectGlobal` requires server execution.

addons/logistics_wirecutter/functions/fnc_destroyFence.sqf
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [fence] call ace_logistics_wirecutter_fnc_destroyFence
+ *
+ * Public: No
+ */
+params ["_fence"];
```
mharis001, line 17: Blank line above `params`.

addons/logistics_wirecutter/XEH_preInit.sqf
```diff
@@ -6,4 +6,26 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+GVAR(replacements) = createHashMapFromArray [
```
PabstMirror, line 9: I guess this can be server only as well

addons/logistics_wirecutter/functions/fnc_destroyFence.sqf
BaerMitUmlaut: Why?

mharis001: For consistency with most of the other function files.

BaerMitUmlaut: Should put it in the guidelines then. There's a bunch of functions which don't have a blank line there. I can change it here, ofc.

BaerMitUmlaut:
Fence destruction has been moved to server side.

veteran29 (approved):

PabstMirror:
- Maps can place down objects with custom size scales (separate xyz) but I don't think it's that common to do for fences
- Fences can have a geo property `'placement' = 'slopelandcontact'` which makes the ground contacts all touch the terrain while keeping the poles vertical. (It sorta turns the fence into a parallelogram).  As far as I know it's not something you can replicate on a mission created object.

before:

after:


This would only be an problem on sloped terrain, so we could just fallback to `setDamage`.

PabstMirror (approved):

BaerMitUmlaut:
Very interesting, I did not know about the slope feature. We could fallback to `setDamage` if the slope is too steep.

PR #8116, opened 2021-02-16
Medical Treatment - Settings to limit IVs to certain locations

pterolatypus:
- Adds two settings to limit use of a) saline and b) blood and plasma to specific locations
- Settings are separate to allow more differentiation of the IV types in the future
- Needs translations

Additionally, I noticed that the Treatment subcategory in settings is getting pretty big. Would it be worth moving this to its own category?

### DIFF
--- a/addons/medical_treatment/initSettings.sqf
+++ b/addons/medical_treatment/initSettings.sqf
@@ -225,22 +225,13 @@
 ] call CBA_settings_fnc_init;
 
 [
-    QGVAR(locationSaline),
+    QGVAR(locationIV),
     "LIST",
-    [LSTRING(LocationSaline_DisplayName), LSTRING(LocationSaline_Description)],
+    [LSTRING(LocationIV_DisplayName), LSTRING(LocationIV_Description)],
     [ELSTRING(medical,Category), LSTRING(SubCategory_Treatment)],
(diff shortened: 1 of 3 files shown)


addons/medical_treatment/initSettings.sqf
```diff
@@ -224,6 +224,24 @@
...
 ] call CBA_settings_fnc_init;
 
+[
+    QGVAR(locationSaline),
+    "LIST",
+    [LSTRING(LocationSaline_DisplayName), LSTRING(LocationSaline_Description)],
+    [ELSTRING(medical,Category), LSTRING(SubCategory_Treatment)],
+    [[0, 1, 2, 3, 4], [ELSTRING(common,Anywhere), ELSTRING(common,Vehicle), LSTRING(MedicalFacilities), LSTRING(VehiclesAndFacilities), ELSTRING(common,Disabled)], 0],
+    true
+] call CBA_settings_fnc_init;
```
mharis001, line 234: `CBA_fnc_addSetting`.

mharis001 (commented):
Why keep plasma and blood together while splitting saline? I think it would be better if all three are controlled by one setting or each has their own.

pterolatypus:
> Why keep plasma and blood together while splitting saline?

It was because I plan to write a separate PR changing how the different fluids work, but on second thoughts I can just split them out if and when I actually do that.

addons/medical_treatment/initSettings.sqf
pterolatypus: changed

PabstMirror (approved):

PR #8122, opened 2021-02-22
cookOff - Fix: Disable ammunition cookoff and turret effect when skipping destruction effects

Vdauphin:
- When you apply `cursorObject setDamage [1, false]` to destroy an object by skipping destruction effects the ammunition still cookoff
- So now we check if we want to skip destruction effects or not before calling cookoff ammunition

### DIFF
--- a/addons/cookoff/XEH_postInit.sqf
+++ b/addons/cookoff/XEH_postInit.sqf
@@ -73,8 +73,11 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 
 // secondary explosions
 ["AllVehicles", "killed", {
-    params ["_vehicle"];
-    if (_vehicle getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmoCookoff)]) then {
+    params ["_vehicle", "", "", "_useEffects"];
+    if (
+        _vehicle getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmoCookoff)] &&
+        _useEffects
+    ) then {
         if (GVAR(ammoCookoffDuration) == 0) exitWith {};
         ([_vehicle] call FUNC(getVehicleAmmo)) params ["_mags", "_total"];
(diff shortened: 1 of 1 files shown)


jonpas:
Can we also check the same thing in normal cookoff? I didn't check, but I think if you do `VEHICLE setDamage [0.99, false]` there is still a chance for cookoff to happen.

Vdauphin:
When you do `VEHICLE setDamage [0.99, false]`.
The vehicle blow up after some second. I think this is an Arma behavior
Because the vehicle explode by itself, I don't see how we can know if we should use the explosion effect or not

Also you might want to set 0.99 damage without effect and let Arma behave normaly?

jonpas:
> Also you might want to set 0.99 damage without effect and let Arma behave normaly?

Well this is really besides the point.

If we can't, we can't.

addons/cookoff/XEH_postInit.sqf
```diff
@@ -73,8 +73,11 @@ GVAR(cacheTankDuplicates) = call CBA_fnc_createNamespace;
 
 // secondary explosions
 ["AllVehicles", "killed", {
-    params ["_vehicle"];
-    if (_vehicle getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmoCookoff)]) then {
+    params ["_vehicle", "", "", "_useEffects"];
+    if (
+        _vehicle getVariable [QGVAR(enableAmmoCookoff), GVAR(enableAmmoCookoff)] &&
+        _useEffects
```
veteran29, line 79: Using the shorter variable first would be a bit more readable imho `_useEffects &&`.

veteran29 (approved):

addons/cookoff/XEH_postInit.sqf
mharis001: Could also combine the `if` `exitWith` below into this as well.

PabstMirror (approved):

PR #8128, opened 2021-02-23
Apply needed pboProject settings in make.py via Windows Registry

Dahlgren:
- Set the required exclude from pbo settings as defined in https://ace3mod.com/wiki/development/setting-up-the-development-environment.html while running make.py

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -227,6 +227,17 @@ def find_bi_tools(work_drive):
     else:
         raise Exception("BadTools","Arma 3 Tools are not installed correctly or the P: drive needs to be created.")
 
+def mikero_windows_registry(path, access=winreg.KEY_READ):
+    try:
+        return winreg.OpenKey(winreg.HKEY_CURRENT_USER, r"Software\Mikero\{}".format(path), access=access)
+    except FileNotFoundError:
+        try:
+            return winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, r"Software\Mikero\{}".format(path), access=access)
+        except FileNotFoundError:
+            try:
+                return winreg.OpenKey(winreg.HKEY_CURRENT_USER, r"Software\Wow6432Node\Mikero\{}".format(path), access=access)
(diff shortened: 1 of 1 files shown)


tools/make.py
```diff
@@ -264,6 +263,17 @@ def find_depbo_tools():
 
     return requiredToolPaths
 
+def pboproject_settings():
+    """Use registry entries to configure needed pboproject settings."""
+    value_exclude = "thumbs.db,*.txt,*.h,*.dep,*.cpp,*.bak,*.png,*.log,*.pew,source,*.tga"
```
PabstMirror, line 268: I seem to remember needing to remove *.h, but I don't think it's for ace, probably fine as it matches our wiki

tools/make.py
```diff
@@ -264,6 +263,17 @@ def find_depbo_tools():
...
+
+    try:
+        k = mikero_windows_registry(r"pboProject\Settings", access=winreg.KEY_SET_VALUE)
+        winreg.SetValueEx(k, "m_exclude", 0, winreg.REG_SZ, value_exclude)
+        winreg.SetValueEx(k, "m_exclude2", 0, winreg.REG_SZ, value_exclude)
```
PabstMirror, line 273: `m_exclude2` seems to be the `-B` exclusions, which should be the only way we build, but it's probably safest to set both

PabstMirror (approved):

tools/make.py
Dahlgren: I should probably try this build and see if that causes any issues 🤔

jonpas: We don't use `.h` anywhere, always `.hpp`. :)

PabstMirror: `my/config.cpp: #includes \a3\editor_f\Data\Scripts\dikCodes.h, but is excluded from pbo`

was the error pboProject gives for my addon, but this doesn't effect ace
And it shouldn't even be an error....

jonpas (approved):

PR #8131, opened 2021-02-25
rhs compats - move into addons and load only if rhs is used

PabstMirror:
lol, this works


jonpas:
Interesting!

veteran29:
Have you tested it with binarized config? It will not work iirc. 

Also I wonder If game is fine with empty config, are there any errors or warnings in rpt?

Dystopian:
But we can disable binarizing of this.
As for empty config - looks like `CfgPatches` can be moved out of `#if __has_include` block.

veteran29:
> But we can disable binarizing of this.
> As for empty config - looks like `CfgPatches` can be moved out of `#if __has_include` block.

Yeah we can not binarize but this makes game startup slower :(

BrettMayson:
> Interesting!
> 
> _I guess HEMTT can't handle it yet @BrettMayson._

And it never would be able to

PabstMirror:
What if we drop a `$NOBIN$` file in the addon, could hemtt recognize that and just do a `pack` on that one pbo?

BrettMayson:
Yes

PabstMirror:
This was just a proof of concept, I'm going to close for now
ideally we could actually totally remove compats and integrate them directly into ace addons
this would solve the problem that the ace_compats are not modular and require multiple ace addons

e.g. ace_rearm could mix vanilla and rhs changes
```
// CfgPatches - ace_rearm {
requiredAddons[] = {"ace_common",
#if __has_include("\rhsafrf\addons\rhs_c_weapons\config.bin")
  "rhs_c_vic"
#endif
};

//cfgVics {
vanilla{}.....
#if __has_include("\rhsafrf\addons\rhs_c_weapons\config.bin")
rhs_vic {} ....
#endif

PR #8137, opened 2021-02-26
Tools - Add script to check sqf/config with sqfvm

PabstMirror:


jonpas:
That's a lot of pragmas. Can't we just ignore initSettings and initKeybinds globally for ACE?

jonpas (approved):
Can add to CI eventually.

PR #8138, opened 2021-02-26
CSW: Add ability to delete dead statics

Cyruz143:
- Allow players to delete dead static weapons.
- Currently when a static weapon is killed, it's stuck in place, you can't move it or disassemble it (even though the action sometimes shows still).
- It's a bit gamey but it's a nice QOL change so you can set up another CSW in the same place by removing the dead one.


PabstMirror:
Won't show for JIPs, it might be better to just add to staticWeapons with `!alive` condition

PabstMirror:
or we just make them draggable, that's probably less gamey

Cyruz143:
I think there's a better way of doing this... will re-PR when I'm happy with it

PR #8140, opened 2021-02-26
User Interface - Add setting for Development Build watermark

Dystopian:
- title;
- replace some numeric idcs with macro from `defineResincl.inc`;
- add addon name check and replace magic variable in setting handler;
- optimize a little `fnc_setAdvancedElement` (move idd check out of loop).

The setting appears on dev branch only. It hides `11400` control group (`class BcgWIP: RscWIP`) which contains 2 controls: `WIP_VersionText` (`idc = 1001`) and `WIP_VersionNumber` (`idc = 1000`). KK wrote about more controls [here](http://killzonekid.com/arma-scripting-tutorials-how-to-remove-beta-watermark-other-things/) but I think it was about A3 beta watermark.

### DIFF
--- a/addons/ui/functions/fnc_setAdvancedElement.sqf
+++ b/addons/ui/functions/fnc_setAdvancedElement.sqf
@@ -76,18 +76,15 @@ private _success = false;
 
     // Loop through IGUI displays as they can be present several times for some reason
     {
+        _success = true;
+
         private _control = _x displayCtrl _idc;
-        if (ctrlFade _control == _fade) then {
-            _success = true;
-            continue;
-        };
+        if (ctrlFade _control == _fade) then {continue};
 
         TRACE_4("Setting Element Visibility",_element,_fade,_idd,_idc);
 
         _control ctrlSetFade _fade;
         _control ctrlCommit 0;
-
-        _success = true;
     } count _displays;
     nil
 } count _elements;
--- a/addons/ui/functions/fnc_setAdvancedElement.sqf
+++ b/addons/ui/functions/fnc_setAdvancedElement.sqf
@@ -66,6 +66,7 @@ if (!_force) then {
     };
 };
 
+private _displays = (uiNamespace getVariable "IGUI_displays") + [findDisplay IDD_MISSION] select {_idd == ctrlIDD _x};
 private _fade = [1, 0] select _show;
 
 // Disable/Enable elements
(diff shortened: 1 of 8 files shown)


Dystopian:
Looks like this unhides watermark in stable build. Will investigate.

PabstMirror:
maybe only add setting `if  (((productVersion # 2) % 2) == 1)`

addons/ui/stringtable.xml
```diff
@@ -156,6 +156,10 @@
             <Spanish>Barra de grupo</Spanish>
             <Turkish>Grup Göstergesi</Turkish>
         </Key>
+        <Key ID="STR_ACE_UI_devWatermark">
+            <English>Development Build watermark</English>
```
jonpas, line 160:
```suggestion
            <English>Development Build Watermark</English>
```

jonpas:
This feels weird having in ACE3. That watermark is (probably) there to tell the user bugs are expected and it is not a stable release.

However, I see the point of it and the statement about bugs isn't really less true for stable builds. Let's be careful about this.

Dystopian:
Fixed, now setting appears only on dev branch.

addons/ui/functions/fnc_setAdvancedElement.sqf
```diff
@@ -66,6 +66,7 @@ if (!_force) then {
     };
 };
 
+private _displays = (uiNamespace getVariable "IGUI_displays") + [findDisplay IDD_MISSION] select {_idd == ctrlIDD _x};
```
jonpas, line 69: Does `+` have precedence over `select`? Otherwise this doesn't do the same thing as it used to.

jonpas:
I am still not sure about this, what do you gain by hiding this?

addons/ui/functions/fnc_setAdvancedElement.sqf
Dystopian: yes

Dystopian:
> I am still not sure about this, what do you gain by hiding this?

This setting (disabled by default) allows to hide watermark when you play dev version constantly. Sometimes dev version has cool features which can't be in performance build. So if you can't wait for the next release you can play dev version and enjoy new features/fixes. Watermark is quite annoying when you're playing more than some minutes. And disabled by default setting doesn't confuse a3 newbies. Also ACE3 `ui` component is the best candidate to handle the problem.

jonpas:
Aaah, I think I finally know which watermark this is. Good, good!

addons/ui/functions/fnc_setAdvancedElement.sqf
jonpas: Would prefer some parentheses to make that clear then.

jonpas (approved):

PabstMirror (approved):

PR #8151, opened 2021-03-06
Update linguist settings

veteran29:
- title
- improves how GitHub counts language usage in the repository
- https://github.com/github/linguist/blob/master/docs/overrides.md

`docker run -v %cd%:/pwd -w /pwd inloopx/linguist linguist`
Before:
```
55.76%  C++
37.61%  SQF
2.51%   C
1.84%   PAWN
1.76%   Python
0.28%   SourcePawn
0.22%   CMake
0.00%   Batchfile
```
After
```
65.69%  SQF
29.67%  C++
2.59%   C
1.82%   Python
0.23%   CMake
0.00%   Batchfile
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,7 +3,7 @@
 *.jpg binary
 *.paa binary
 
-# Change GitHub language categorisation
+# Change GitHub language categorization
 addons/**/*.cpp linguist-language=SQF
 addons/**/*.hpp linguist-language=SQF
 
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,13 @@
 *.png binary
 *.jpg binary
 *.paa binary
+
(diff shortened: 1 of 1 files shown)


PabstMirror (approved):
neat

.gitattributes
```diff
@@ -2,3 +2,13 @@
 *.png binary
 *.jpg binary
 *.paa binary
+
+# Change GitHub language categorisation
```
PabstMirror, line 6:
```suggestion
# Change GitHub language categorization
```

jonpas:
Nice! But is config SQF? :thinking:

veteran29:
> Nice! But is config SQF? 🤔

Yeah it's not but it's closer to SQF than it's to CPP. I'm not aware about any other lang we could categorize it with.

PR #8153, opened 2021-03-06
WIP: Complex vitals

pterolatypus:
Add more realistic calculation of stroke volume and blood pressure:
- SV determined by approximations for ventricular preload (blood volume, mean bp), afterload (systolic bp) and contractility (medication)
- Mean Arterial Pressure calculated as CO * SVR
- High and low BP determined from MAP by pulse pressure and systole fraction ([reference](https://www.researchgate.net/publication/14609651_Calculation_of_Mean_Arterial_Pressure_During_Exercise_as_a_Function_of_Heart_Rate))
- HR will now adapt to normalise blood pressure in more cases

This puts SV and BP into a feedback loop w.r.t each other, so we must be careful that it balances sensibly.
Currently, HR spikes causing CA much more quickly than before - I plan to tune the HR response and add a vasoconstriction response to mitigate.
I'd also like to move a lot of the vital conditions that are based on blood volume to use CO & BP instead, so that these vital stats are more actually representative of a patient's state.

Other plans (possibly for separate PRs) include:
- More integration with Adv Fatigue
- Link to ACEX Field Rations (dehydration)

The idea is to make managing vitals a little less predictable, as there are more factors involved. I'd like it to be tuned in such a way that simple strategies (more blood = good) work ~90% of the time, while deeper understanding can allow dedicated medics to optimise the remaining 10%.
There's still a lot of placeholder stuff and magic numbers, I know, just want to open it up for discussion.

### DIFF
--- a/addons/medical_engine/script_macros_medical.hpp
+++ b/addons/medical_engine/script_macros_medical.hpp
@@ -54,6 +54,11 @@
 #define BLOOD_VOLUME_CLASS_4_HEMORRHAGE 3.600 // lost more than 40% blood, Class IV Hemorrhage
 #define BLOOD_VOLUME_FATAL 3.0 // Lost more than 50% blood, Unrecoverable
 
+#define DEFAULT_BP_MEAN 107
+#define DEFAULT_BP_HIGH 120
+#define DEFAULT_BP_LOW 80
+#define DEFAULT_BLOOD_PRESSURE [DEFAULT_BP_LOW,DEFAULT_BP_HIGH,DEFAULT_BP_MEAN] // at default heart rate
+
 // IV Change per second calculation:
 // 250 ml should take 60 seconds to fill. 250 ml / 60 s ~ 4.1667 ml/s.
 #define IV_CHANGE_PER_SECOND 4.1667 // in milliliters per second
--- a/addons/medical_engine/script_macros_medical.hpp
+++ b/addons/medical_engine/script_macros_medical.hpp
@@ -43,6 +43,7 @@
 
 #define DEFAULT_HEART_RATE 80
 #define DEFAULT_PERIPH_RES 100
+#define DEFAULT_STROKE_VOLUME 95e-3
 
 // --- blood
 // 0.077 l/kg * 80kg = 6.16l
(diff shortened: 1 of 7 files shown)


bababoombatman:
I might be wrong so feel free to correct me, but the afterload is the mean BP. The systolic pressure is merely the pressure generated by the right ventricle. It need to overcome the mean BP (the diastolic one to be precise) in order to open the aortic valve (hence afterload). The preload is merely the volume at be beginning of the systole, it has nothing to do with the mean arterial blood pressure, the pulmunary veins are the ones bringing the blood to the left hearth (if you want to take pressure in account it should be the central venous pressure, even though that's still not entirely correct)

pterolatypus:
Thanks for the comments. I'm not a cardiologist, so a lot of the stroke volume calculations are basically guesswork based on what sounds reasonable and the limited information I have available (access to scientific papers is _expensive_) - I expect this PR to need a **lot** of tweaking. I'll explain my reasoning, but I welcome feedback/discussion/corrections. If you have any sources or reference material that'd be really useful too.

>afterload is the mean BP

What I've read is inconclusive, but from my understanding it should be based on the aortic pressure during systole; I've just assumed that this should be closely related to systolic BP. In resting conditions it doesn't make a difference, but changes to heart rate now affect the ratio of BPs/BPd/BPm. What might make more sense would be to estimate mean aortic pressure during systole as a fixed ratio of systolic + diastolic BP (as opposed to the mean arterial pressure during the whole cycle, the ratio of which varies based on heart rate).

>preload is merely the volume at be beginning of the systole

Well, they're proportional so basically; this is what I actually calculate (end-diastolic volume) and I don't use the word "preload" in the code. I'm basically modeling venous return as proportional to CVP and approximating that based _mostly_ (70%) off blood volume. The other 30% is based off arterial pressure to allow changes to SVR to increase venous return, but not linearly. Maybe it should instead be `blood volume * resistance^something`, or perhaps I need to model venous resistance separately from the total resistance. Either way I should look deeper into how the vasoconstriction response works - it's currently pretty basic and I'd rather it not be based directly off blood volume.

PR #8154, opened 2021-03-07
Interaction - Show all possible modes for an attachment

PabstMirror:
image


Dystopian:
Tested, works. I would suggest to move these actions from `Detach` submenu to between `Attach` and `Detach` actions. It's much easier to select actions this way at least when menu is in list mode.

mjc4wilton:
> Tested, works. I would suggest to move these actions from `Detach` submenu to between `Attach` and `Detach` actions. It's much easier to select actions this way at least when menu is in list mode.

Maybe make it so it goes Weapon Attachments -> [Array of attachments] -> [Mode, Detach, etc.]?

addons/interaction/functions/fnc_getWeaponAttachmentsActions.sqf
```diff
@@ -49,8 +49,30 @@ params ["_unit"];
     // "detach" actions
     {
         if (_x isEqualTo "") then {continue};
+        private _originalAttachment = _x;
 
-        private _config = _cfgWeapons >> _x;
+        private _switchableAttachments = [_originalAttachment] call CBA_fnc_switchableAttachments;
+        private _convertToActions = [];
+        {
+            if (_x == _originalAttachment) then {continue};
```
Dystopian, line 57:
```suggestion
            if (_x == _originalAttachment) then {continue};
            if (getText(_CBA_PIPItems >> _x) == _originalAttachment) then {continue};
```
CBA PIP optics should be ignored too. CUP has many scopes like this (e.g. ACOG and SUSAT).

addons/interaction/functions/fnc_getWeaponAttachmentsActions.sqf
```diff
@@ -49,8 +49,30 @@ params ["_unit"];
     // "detach" actions
     {
```
Dystopian, line 50:
```suggestion
    private _CBA_PIPItems = configFile >> "CBA_PIPItems";
    {
```

Dystopian (changes requested):

johnb432:
> Maybe make it so it goes Weapon Attachments -> [Array of attachments] -> [Mode, Detach, etc.]?

I fully agree with that, given we're adding some new actions. I went ahead and pushed the necessary changes.

Do we want to allow switching to PIP modes for scopes? To me it doesn't seem to be the worst idea, but I don't really use those scopes.

johnb432 (approved):
LGTM, but given I've rewritten part of the code, I'd like another person to take a second look at this.

PabstMirror:
image

IMHO this might helpreadability to make the detach/attach sub-actions simpler
so they stand out from the switch action?

instead of `SPIR` -> `Detach: SPIR`
now just  `SPIR` -> `Detach`

johnb432:
> IMHO this might helpreadability to make the detach/attach sub-actions simpler so they stand out from the switch action?
> 
> instead of `SPIR` -> `Detach: SPIR` now just `SPIR` -> `Detach`

Done.

PR #8160, opened 2021-03-11
Medical - Improve keybind to open med menu for vehicle occupant under cursor

ampersand38:
- Add lineIntersectsSurfaces check to medical menu keybind
- Allows it to be used to open medical menu for vehicle occupant being looked at
- https://youtu.be/noIHrXAi8qc

### DIFF
--- a/addons/medical_gui/XEH_postInit.sqf
+++ b/addons/medical_gui/XEH_postInit.sqf
@@ -33,13 +33,24 @@ GVAR(selfInteractionActions) = [];
 }] call CBA_fnc_addEventHandler;
 
 ["ACE3 Common", QGVAR(openMedicalMenuKey), localize LSTRING(OpenMedicalMenu), {
-    // Get target (cursorTarget and cursorObject), if not valid then target is ACE_player
+    // Get target (cursorTarget, cursorObject, and lineIntersectsSurfaces along camera to maxDistance), if not valid then target is ACE_player
     TRACE_3("Open menu key",cursorTarget,cursorObject,ACE_player);
     private _target = cursorTarget;
     if !(_target isKindOf "CAManBase" && {[ACE_player, _target] call FUNC(canOpenMenu)}) then {
         _target = cursorObject;
         if !(_target isKindOf "CAManBase" && {[ACE_player, _target] call FUNC(canOpenMenu)}) then {
-            _target = ACE_player;
+            private _start = AGLToASL positionCameraToWorld [0, 0, 0];
+            private _end = AGLToASL positionCameraToWorld [0, 0, GVAR(maxDistance)];
+            private _intersections = lineIntersectsSurfaces [_start, _end, ACE_player, objNull, true, -1, "FIRE"];
+            _target = {
+                _x params ["", "", "_intersectObject"];
+                if (_intersectObject != ACE_player && {_intersectObject isKindOf "CAManBase" && {[ACE_player, _intersectObject] call FUNC(canOpenMenu)}}) exitWith {
+                    _intersectObject
+                };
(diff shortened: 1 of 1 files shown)


addons/medical_gui/XEH_postInit.sqf
```diff
@@ -33,13 +33,24 @@ GVAR(selfInteractionActions) = [];
...
     TRACE_3("Open menu key",cursorTarget,cursorObject,ACE_player);
     private _target = cursorTarget;
     if !(_target isKindOf "CAManBase" && {[ACE_player, _target] call FUNC(canOpenMenu)}) then {
         _target = cursorObject;
         if !(_target isKindOf "CAManBase" && {[ACE_player, _target] call FUNC(canOpenMenu)}) then {
-            _target = ACE_player;
+            private _start = AGLToASL positionCameraToWorld [0, 0, 0];
+            private _end = AGLToASL positionCameraToWorld [0, 0, GVAR(maxDistance)];
+            private _intersections = lineIntersectsSurfaces [_start, _end, ACE_player, objNull, true, -1, "FIRE"];
+            _target = {
```
PabstMirror, line 45: _target will be nil if there are no intersections

ampersand38: _target

jokoho48:
if I see it correct with this PR you can Interact with Units through walls. the easiest way to fix that is the check if _target has Any Crew in it and only then do the lis

PabstMirror:
limiting `lineIntersectsSurfaces` object limit to 1 (instead of -1) should fix?
or add same vehicle check to line 47 `(vehicle _intersectObject) == (vehicle ace_player)`

jokoho48:
I think limiting the return amount to 1 would be the best solution I think.
checking If the player is in the same vehicle would not work because the interaction is meant to be usable also from outside of the vehicle(at least if I understood the video correctly)

ampersand38:
When in a vehicle, despite having `ACE_player` in in ignored object, the first return is still the player unit. Checking the `crew` is a great idea though, to only look "through" crewed vehicles.

addons/medical_gui/XEH_postInit.sqf
```diff
@@ -33,13 +33,26 @@ GVAR(selfInteractionActions) = [];
...
+            private _intersections = lineIntersectsSurfaces [_start, _end, ACE_player, objNull, true, -1, "FIRE"];
+            {
+                _x params ["", "", "_intersectObject"];
+                // Only look "through" player and player's vehicle
+                if (!(_intersectObject isKindOf "CAManBase") && {_intersectObject != vehicle ACE_player}) exitWith {};
+                if (_intersectObject != ACE_player && {_intersectObject isKindOf "CAManBase" && {[ACE_player, _intersectObject] call FUNC(canOpenMenu)}}) exitWith {
+                    _target =_intersectObject
+                };
+            } forEach _intersections;
+            if !(_target isKindOf "CAManBase" && {[ACE_player, _target] call FUNC(canOpenMenu)}) then {
```
commy2, line 53:
```suggestion
            if (!(_target isKindOf "CAManBase") || {!([ACE_player, _target] call FUNC(canOpenMenu))}) then {
```
I don't like the negation over conjunction. This is imo easier to read.

Double check if what you wrote is what was intended.

ampersand38: I reused the condition from L39 and L41. Would you like me to change those as well?

veteran29: Yes please do change them so it's consistent.

jonpas:
Very nice improvement!

PR #8162, opened 2021-03-12
Medical Treatment - Allow diagnosing death/CA (invalidates old setting)

PabstMirror:
Adds a 3rd setting level to advancedDiagnose that shows death/CA state

Doing CPR for 5 minutes on your buddy, not sure if he's alive or dead, might be heartbreakingly realistic
but I feel like some players will want a way to diagnose their state easily without having to drop down to basic diagnose

### DIFF
--- a/addons/medical_treatment/stringtable.xml
+++ b/addons/medical_treatment/stringtable.xml
@@ -3703,7 +3703,7 @@
             <Turkish>%1 tepki vermiyor</Turkish>
         </Key>
         <Key ID="STR_ACE_Medical_Treatment_Check_Response_CardiacArrest">
-            <English>%1 is not responsive, taking shallow gasps and seizuring</English>
+            <English>%1 is not responsive, taking shallow gasps and convulsing</English>
         </Key>
         <Key ID="STR_ACE_Medical_Treatment_Check_Response_Dead">
             <English>%1 is not responsive, motionless and cold</English>
--- a/addons/medical_treatment/stringtable.xml
+++ b/addons/medical_treatment/stringtable.xml
@@ -58,7 +58,7 @@
             <Spanish>Habilita la comprobación del pulso, presión sanguinea y rsepuesta a las acciones del tratamiento en lugar de la acción de diagnóstico genérico. \n Cuando se deshabilita, la acción RCP solo se mostrará cuando realizar un RCP sea apropiado. \n Las acciones activadas de este ajuste son necesarias para determinar si una persona está inconsciente o en parada cardiaca.</Spanish>
         </Key>
         <Key ID="STR_ACE_Medical_Treatment_AdvancedDiagnose_DiagnoseCardiacArrest">
-            <English>Enabled + Can diagnose Death/Cardiac Arrest</English>
+            <English>Enabled &amp; Can Diagnose Death/Cardiac Arrest</English>
         </Key>
         <Key ID="STR_ACE_Medical_Treatment_AdvancedMedication_DisplayName">
             <English>Advanced Medication</English>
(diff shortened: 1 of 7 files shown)


jonpas (approved):
Will this break use of `true/false` in the setting (eg. via settings file)?

addons/medical_treatment/stringtable.xml
```diff
@@ -57,6 +57,9 @@
             <Turkish>Genel Teşhis eylemi yerine Nabzı Kontrol Et, Kan Basıncını Kontrol Et ve Yanıtı Kontrol Et tedavi eylemlerini etkinleştirir. \ Devre dışı bırakıldığında, CPR eylemi yalnızca CPR gerçekleştirilmesi uygun olduğunda gösterilecektir.</Turkish>
             <Spanish>Habilita la comprobación del pulso, presión sanguinea y rsepuesta a las acciones del tratamiento en lugar de la acción de diagnóstico genérico. \n Cuando se deshabilita, la acción RCP solo se mostrará cuando realizar un RCP sea apropiado. \n Las acciones activadas de este ajuste son necesarias para determinar si una persona está inconsciente o en parada cardiaca.</Spanish>
         </Key>
+        <Key ID="STR_ACE_Medical_Treatment_AdvancedDiagnose_DiagnoseCardiacArrest">
+            <English>Enabled + Can diagnose Death/Cardiac Arrest</English>
```
mharis001, line 61:
```suggestion
            <English>Enabled &amp; Can Diagnose Death/Cardiac Arrest</English>
```
For consistency with `GVAR(advancedBandages)` setting.

PabstMirror:
So I thought that if we converted bool to list, it would do false=0, true=1, but that doesn't seem to be true in cba_settings 
(maybe it was for old ace_settings?)

So right now it will ignore any saved true/false settings and use the default?
Would it be better to add this as a new setting?

addons/medical_treatment/stringtable.xml
```diff
@@ -3699,6 +3702,12 @@
             <Chinese>%1 沒有反應</Chinese>
             <Turkish>%1 tepki vermiyor</Turkish>
         </Key>
+        <Key ID="STR_ACE_Medical_Treatment_Check_Response_CardiacArrest">
+            <English>%1 is not responsive, taking shallow gasps and seizuring</English>
```
PabstMirror, line 3706:
```suggestion
            <English>%1 is not responsive, taking shallow gasps and convulsing</English>
```

PabstMirror: I don't think seizuring is a real word

addons/medical_treatment/stringtable.xml
```diff
@@ -3699,6 +3702,12 @@
             <Chinese>%1 沒有反應</Chinese>
             <Turkish>%1 tepki vermiyor</Turkish>
         </Key>
+        <Key ID="STR_ACE_Medical_Treatment_Check_Response_CardiacArrest">
+            <English>%1 is not responsive, taking shallow gasps and convulsing</English>
```
JoramD0, line 3706: Gasping is fairly common in early cardiac arrest, not sure if convulsing is. I am basing this on civilian first aid training (that I luckily have not had to use yet), not military experience.

jonpas:
Anything to consider/check/test/review on this?

PabstMirror:
Is it ok to break the existing setting by changing it's type, or should I just add a 2nd bool?

jonpas:
There, noted in changelog now, I think that's fine. Not changing it won't cause a script error, right?

PabstMirror:
cba settings should check types and maybe complain in rpt, but should all be handled safely

mharis001:
Should be fine. We have made a similar change to the advanced bandages and bandage reopening settings.

PR #8170, opened 2021-03-15
Medical - Add Survivability Chance for Fatal Injuries

TyroneMF:
- Adds a percentage slider for surviving a fatal injury.

### DIFF
--- a/addons/medical_damage/functions/fnc_determineIfFatal.sqf
+++ b/addons/medical_damage/functions/fnc_determineIfFatal.sqf
@@ -26,12 +26,12 @@ scopeName "main";
 
 if (EGVAR(medical,fatalDamageSource) in [0, 2]) then {
     // Emulate damage to vital organs - Original rewrite logic, only powerfull headshots or random torso shots
-    if (_part == 0 && {_woundDamage >= HEAD_DAMAGE_THRESHOLD} && {random 1 < EGVAR(medical,deathChance)}) exitWith {
+    if (_part == 0 && {_woundDamage >= HEAD_DAMAGE_THRESHOLD} && {random 1 < EGVAR(medical,survivalChance)}) exitWith {
         // Fatal damage to the head is guaranteed death
         TRACE_1("determineIfFatal: lethal headshot",_woundDamage);
         true breakOut "main";
     };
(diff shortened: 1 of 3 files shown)


commy2:
This does not make sense to me. If something is fatal, you cannot by definition survive it.

BrettMayson:
I'm not really a fan of RNG, but I don't see an issue with adding it for those that want it

kymckay:
My gut instinct agrees with @commy2, can someone clarify why this is desirable behaviour?

If fatal injuries are happening too often then I think the solution is to change the event itself, not add another layer of decision making with another setting to confuse things.

> I'm not really a fan of RNG, but I don't see an issue with adding it for those that want it

In general, the issue with continually adding new things is it can dilute existing functionality and bloat the codebase (settings in particular are hard to balance, it's not always the right answer to add more). Quality > quantity.

BrettMayson:
> In general, the issue with continually adding new things is it can dilute existing functionality and bloat the codebase (settings in particular are hard to balance, it's not always the right answer to add more). Quality > quantity.

While overall I agree that too many settings tweaking simulation isn't a good thing, this setting isn't exactly confusing or adding much for code

TyroneMF:
> My gut instinct agrees with @commy2, can someone clarify why this is desirable behaviour?

It's desirable because right now your options are either dying in cardiac arrest or blood loss which is easily countered by competent medics. Or having fatal injuries and dying to (in my experience) the first bullet that hits you.

This just gives you a middle ground of letting people die to gunshot wounds with a setting people can control. By default it's set to 0% so if you don't want to use it you don't have to.

kymckay:
>It's desirable because right now your options are either dying in cardiac arrest or blood loss which is easily countered by competent medics. Or having fatal injuries and dying to (in my experience) the first bullet that hits you.

Thank you and well put, that helps me understand the motivation 🙂

I can see the need for a middle ground between too harsh and too forgiving, just not sure if I agree with this way of achieving it as the logic seems so antithesis to the idea of a fatal injury system. I also don't currently have a better suggestion though, so it may be fine (even as a compromise that could be removed later if anyone rebalances the medical systems/complexity).

TyroneMF:
Closing. Use #8192 instead.

PR #8093, opened 2021-02-04
Tools - Make.py don't Binarise ANY configs

PabstMirror:
On current tools (pboProject 2.89/dePbo7.97)
Project built successfully but has significant errors in-game

ace_rearm:
```
#define GET_NUMBER(config,default) (if (isNumber (config)) then {getNumber (config)} else {default})
#define DEFAULT_REARMCARGO GET_NUMBER(configFile >> 'CfgVehicles' >> typeOf _this >> QQGVAR(defaultSupply),-1)
defaultValue = QUOTE(DEFAULT_REARMCARGO);
// mikro: Missing else statement
defaultValue = "(if (isNumber (configFile >> 'CfgVehicles' >> typeOf _this >> ""ace_rearm_defaultSupply"")) then {getNumber (configFile >> 'CfgVehicles' >> typeOf _this >> ""ace_rearm_defaultSupply"")}";
// expected
defaultValue = "(if (isNumber (configFile >> 'CfgVehicles' >> typeOf _this >> ""ace_rearm_defaultSupply"")) then {getNumber (configFile >> 'CfgVehicles' >> typeOf _this >> ""ace_rearm_defaultSupply"")} else {-1})";
```
ace_zeus:
```
condition = QUOTE(!(isNil QQGVAR(zeus) || {isNull GVAR(zeus)}));
// mikro: missing ) at end
condition = "!(isNil ""ace_zeus_zeus"" || {isNull ace_zeus_zeus}";
// expected
condition = "!(isNil ""ace_zeus_zeus"" || {isNull ace_zeus_zeus})";
```

I don't feel like we can trust this tool anymore
This PR should disable binarization and pack raw configs.
Only downside I know of might be a trivial increase to loading speed

PR #8172, opened 2021-03-16
Documentation - Add Mikero binarization note

jonpas:
- Instruct user to do #8093 manually, as `-B` on command line does not seem to work properly.


Dystopian:
Maybe we should edit setting in registry like in #8128? Looks like setting name is `m_no_binarise`.

PabstMirror (approved):

PabstMirror:
that sounds ok to me if someone wants to do it

PR #8174, opened 2021-03-17
Name Tags - Add ability to set custom rank icon

Drift91:
- This pull request adds the ability to set custom rank icons on an individual unit via `setVariable`.
- Usage: `_unit setVariable ["ace_nametags_rankIcon", "<path to .paa file>"]`
- This should override the faction rank icons if set, otherwise the faction icons or defaults will be used as appropriate.
- An advantage to using the direct path to the .paa file is that it should allow loading from both addons and the mission file, if I'm not mistaken.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -71,6 +71,7 @@ Dharma Bellamkonda <dharma.bellamkonda@gmail.com>
 Dimaslg <dimaslg@telecable.es>
 diwako
 dixon13 <dixonbegay@gmail.com>
+Drift_91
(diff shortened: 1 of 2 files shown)


commy2:
`customRankIcon` -> `rankIcon` ?

Drift91:
@commy2 Done. I did find the variable name a bit long myself. I also updated my original post for the pull request description.

Nothing wrong with my code that you can see? This is my first pull request for a script written in SQF, so just checking that I haven't messed anything up.

jonpas (approved):

Issue #6200, opened 2018-03-18
Zoom in/out action of compass on map broken by ace_map


severgun:
**Arma 3 Version:** `1.80` (stable)
**CBA Version:** `3.5` (stable)
**ACE3 Version:** `3.12.1` (stable)

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
- Zoom in/out action of compass on map broken by ace_map. Once you double click and move it, you can't zoom in/out it again, because Arma overwrite position[] and positionBack[] at Arma3Profile on disk.

**Steps to reproduce:**
- Double click on compass, move it around, double click again and again and again.

**Where did the issue occur?**
- Everywhere

**Placed Modules:**
- none

**Solution**
- Seems like setting zoomDuration to **non zero** value will fix this. (I tried 0.1)
https://github.com/acemod/ACE3/blob/2cad71cd1e42338fb757ff1ea22b53d44d401bcc/addons/map/config.cpp#L103

commy2:
Isn't this intentional?

severgun:
What intentional? Line 102 make it bigger. But you can stuck with zoomed out which is not readable. And all your efforts in line 102 went to hell.

So IMHO author want make it bigger and disable animation of resizing. But zero value gives undefined behavior.

PabstMirror:
![compass](https://user-images.githubusercontent.com/9376747/37728321-f3edacb2-2d07-11e8-86e0-9ee297b95c2f.png)

Vanilla Small
Vanilla Big
ACE Normal (it's locked to this size)
ACE - Big (keep scale change, but allow zooming)
```
            scale = 0.7;
            // zoomDuration = 0;
```

So I guess we wanted to remove the compass zooming and just leave it the big size all the time.
ACE does make it as big as you could get from vanilla
There does seem to be some weirdness in the profile, I think people would have to change it to fix zooming.

severgun:
I can't get why not to allow zoom action? Why compass should stay at one size?
Also zoomDuration did not disable zoom action even if it was desired.

Now it can stuck at vanilla small.

jonpas:
Because zooming compass is broken in the game. It can get stuck in small mode.

commy2:
Isn't this also about the ruler being scaled correctly?

severgun:
BIS Fault?
How to reproduce stuck on vanilla?

commy2:
Never heard of that. It doesn't scale so it fits the map grid.

barman75:
[...]

PR #8176, opened 2021-03-19
Map - Fix stuck map compass size (add 0.1 zoom duration)

R3voA3:
See #6200

* This PR changes `zoomDuration = 0;` to `zoomDuration = 0.1;` in map config.

### DIFF
--- a/addons/map/config.cpp
+++ b/addons/map/config.cpp
@@ -100,7 +100,7 @@ class RscDisplayMainMap {
     class objects {
         class Compass: RscObject {
             scale = 0.7;
-            zoomDuration = 0.1;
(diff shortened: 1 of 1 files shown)


addons/map/config.cpp
```diff
@@ -100,7 +100,7 @@ class RscDisplayMainMap {
     class objects {
         class Compass: RscObject {
             scale = 0.7;
-            zoomDuration = 0;
+            zoomDuration = 0.1;
```
jonpas, line 103:
```suggestion
            zoomDuration = 0.1; // 0 (instant transition) breaks Arma saving position/size
```

jonpas (approved):

PR #8186, opened 2021-03-27
Interact Menu - Simplify fnc_renderActionPoints

Dystopian:
- title.

To understand how inline function `_fnc_renderNearbyActions` works I had to simplify it a little. Changes:
- use `continue` to decrease loop nesting depth;
- replace `_numInteractions` (int) with `_hasInteractions` (bool) because number is not used;
- fix `_classActions = _namespace getVariable typeOf _target` doesn't have default value;
- some small code optimizations.

It's better to view changes in IDE than here in GitHub.

### DIFF
--- a/addons/interact_menu/functions/fnc_renderActionPoints.sqf
+++ b/addons/interact_menu/functions/fnc_renderActionPoints.sqf
@@ -35,48 +35,47 @@ private _fnc_renderNearbyActions = {
         private _target = _x;
 
         // Quick oclussion test. Skip objects more than 1 m behind the camera plane
-        private _lambda = ((getPosASL _x) vectorDiff GVAR(cameraPosASL)) vectorDotProduct GVAR(cameraDir);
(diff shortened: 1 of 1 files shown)


addons/interact_menu/functions/fnc_renderActionPoints.sqf
```diff
@@ -35,48 +35,47 @@ private _fnc_renderNearbyActions = {
...
+        private _classActions = GVAR(ActNamespace) getVariable [typeOf _target, []];
+        {
+            private _action = _x;
+            // Try to render the menu
+            if ([_target, _action] call FUNC(renderBaseMenu)) then {
+                _hasInteractions = true;
+                GVAR(foundActions) pushBack [_target, _action, GVAR(objectActionList)];
             };
+            nil
+        } count _classActions;
```
dedmen, line 72: count+nil is slower than forEach without extra nil command

PabstMirror (approved):

PR #8187, opened 2021-03-27
Optionals - Add No Medical and No Realistic Names components

Dystopian:
In some missions ACE Medical can't be used. Also there were some players who wanted to disable Realistic Names component.
I think this PR is most correct way to disable ACE components without mod folder modifying and messing with own mods based on ACE. Empty config is correctly loaded and works as expected when such optional component is enabled.

The biggest problem is that build tool must support `__has_include` preprocessor command. As of now Mikero tools (DePbo 7.97.0.25) don't support it and even with `-B` argument they try to rapify/check syntax (fail as result). AddonBuilder builds successfully.

About `$NOBIN$`. It looks like Mikero tools with `-B (Don't binarise)` argument do not binarize `config.cpp` but binarize models etc. AddonBuilder with `-packonly` argument do not binarize anything. So both tools can't be used for this PR building :disappointed: 
I would suggest addding new file flag (e.g. `$NOBINCONFIG$`) to indicate that only config should not be binarized. And naturally build tool (HEMTT?) should support such mode. Config syntax should be checked anyway ignoring `__has_include`.
@BrettMayson could you implement such feature if ACE team agrees to approve this PR?

Making this PR as draft until there is a tool for building.


addons/medical/config.cpp
```diff
@@ -1,3 +1,7 @@
+#if __has_include("\z\ace\addons\nomedical\script_component.hpp")
```
commy2, line 1: Does this even work when ACE is binned?

Dystopian: no, medical components must be unbinned

commy2: That's pretty bad.

PabstMirror: we're talking about configs, not models 
the difference between loading a binned vs non-binned config is probably a milisecond

jokoho48: The bigger issue is that hemtt does not support this yet. so using this would break the build tools

BaerMitUmlaut:
What's the point of this? What is the advantage of having to add a file instead of having to remove it? This doesn't make sense to me, sorry.

commy2:
Probably something about Steam. I wouldn't know.

Dystopian:
> 
> 
> What's the point of this? What is the advantage of having to add a file instead of having to remove it? This doesn't make sense to me, sorry.

- you don't have to manage files in ACE directory on each ACE update;
- you can simply disable/enable ACE medical for some game sessions/missions with just loading/unloading optional addon.

BaerMitUmlaut:
> you don't have to manage files in ACE directory on each ACE update;

Does the optional PBO move itself?

commy2:
No, but you can un/check it in Steam I suppose.

Dystopian:
> Does the optional PBO move itself?

You don't have to move optionals since #5038.

PabstMirror:
Problem we've found is that `__has_include` only checks if the file can be loaded
but it applies to both the virtual filesystem (files from pbos) and real files in the arma folder
e.g.
`#if __has_include("\test\t.hpp")`
will return true if you have a `"F:\SteamLibrary\steamapps\common\Arma 3\test\t.hpp"`

it's just like #inclding a `\userconfig\` file
we really need a `__has_include` that only checks virtual filesystem

PabstMirror:
preprocessor should respect filepatching in 2.06

Superxpdude:
Are there any plans to continue development of this PR? I am interested in this as a feature, and can put in the time to get it working via another PR, but I don't want to duplicate the work in case there are still plans to finish this one up.

Mostly wondering because the limiting factor for this seems to be resolved as far as I can tell based on PR #9116.

Dystopian:
@Superxpdude I will try to update it this week.

jonpas:
What's still preventing Medical from having a setting to enable/disable it?

Dystopian:
> What's still preventing Medical from having a setting to enable/disable it?

IIRC there were changes in unit damage config. I didn't find better way to handle it.

LinkIsGrim:
> > What's still preventing Medical from having a setting to enable/disable it?
> 
> IIRC there were changes in unit damage config. I didn't find better way to handle it.

We could handle the limb armor via script in `fnc_handleDamage` (ugh), I guess. Will just adding ACE_HDBracket change vanilla behavior in anyway?

LinkIsGrim (approved):

addons/fire/addon.toml
```diff
@@ -0,0 +1 @@
+preprocess = false
```
PabstMirror, line 1: need to update syntax
I can do this, just making note

optionals/nomedical/$PBOPREFIX$
```diff
@@ -0,0 +1 @@
+z\ace\addons\nomedical
```
jonpas, line 1:
```suggestion
z\ace\addons\nomedical

```

optionals/norealisticnames/$PBOPREFIX$
```diff
@@ -0,0 +1 @@
+z\ace\addons\norealisticnames
```
jonpas, line 1:
```suggestion
z\ace\addons\norealisticnames

```

jonpas (changes requested):

optionals/norealisticnames/$PBOPREFIX$
LinkIsGrim: We don't have newlines at EOF for any other component. Sure about this?

jonpas: We should have. :thinking:

jonpas: Some components have it, some don't, all can and should have it though. Mostly older ones don't have it as those files never get touched.

addons/fire/config.cpp
```diff
@@ -1,5 +1,12 @@
 #include "script_component.hpp"
 
+#if __has_include("\z\ace\addons\nomedical\script_component.hpp")
+#define PATCH_SKIP "No Medical"
+#endif
+
+#ifdef PATCH_SKIP
+ACE_PATCH_NOT_LOADED(ADDON,PATCH_SKIP)
+#else
```
jonpas, line 9:
```suggestion
#if __has_include("\z\ace\addons\nomedical\script_component.hpp")
ACE_PATCH_NOT_LOADED(ADDON,PATCH_SKIP)
#else

```

Why not just this? :thinking:

Dystopian: This way can define `PATCH_SKIP` with several conditions. E.g.:
```
#if __has_include("\z\ace\addons\nomedical\script_component.hpp")
#define PATCH_SKIP "No Medical"
#endif
#if __has_include("\some\other_addon\config.bin")
#define PATCH_SKIP "Some Other Addon"
#endif
```
and then you use `PATCH_SKIP` just once to disable component with `ACE_PATCH_NOT_LOADED`. In this PR this is mostly a good template for other `__has_include` components. I liked @PabstMirror idea and implemeted it here too.
See also https://github.com/acemod/ACE3/blob/fdbcc62c74f1c51cf3ff36b784789a92fada4702/addons/compat_ws/config.cpp#L3-L9

Also you have `PATCH_SKIP` undefined in suggestion.

jonpas: Ok your example adds multiple conditions, but you can also do that like this:
```cpp
#if __has_include("\z\ace\addons\nomedical\script_component.hpp")
#if __has_include("\some\other_addon\config.bin")
ACE_PATCH_NOT_LOADED(ADDON,"Some Other Addon")
#else
ACE_PATCH_NOT_LOADED(ADDON,"No Medical")
#else
// default config
#endif
#endif
```
Although that is uglier and I wouldn't want to do it.

Not sure, we just need the opposite of `skipWhenMissingDependencies`!

_Fixed suggestion._

jokoho48: I am not fully sure about `#if`, but I seem to remember that preprocessor "commands" like that are not able to be nested(at least `#ifdef` is not able to be nested)

jonpas: Ah, and they also don't support logical operators I believe. Yeah, we need a `skipWhenLoaded`.

PabstMirror:
```
#if __has_include("\rhsafrf\addons\rhs_main\loadorder\config.bin")
#else
    #define PATCH_SKIP "RHS AFRF"
#endif

#if __has_include("\z\ace\addons\explosives\script_component.hpp")
#else
    #ifdef PATCH_SKIP
    #else
        #define PATCH_SKIP "ACE Explosives"
    #endif
#endif
```
works on game, mikro and hemtt

rautamiekka: What about [`#ifndef`](https://community.bistudio.com/wiki/PreProcessor_Commands#.23ifndef), is it working everywhere ? Cuz if it is, I don't see why do ifdef-else.

veteran29: >  but I seem to remember that preprocessor "commands" like that are not able to be nested(at least #ifdef is not able to be nested)

This should be fixed at least for a patch or two.

LinkIsGrim: #9294 seems to work fine with nesting.

jonpas (approved):

PR #8190, opened 2021-03-28
advanced_fatigue: PRONE and CROUCH are mixed up

shukari:
- CROUCH is atm better than PRONE seams mixed up


BaerMitUmlaut:
No, because the sway is already reduced in vanilla.

PR #8192, opened 2021-04-02
Medical - Add Alternative Survival Chance

BrettMayson:
Co-authored-by: @TyroneMF

Alternative to #8170

- Adds a percentage slider for surviving a fatal injury.

### DIFF
--- a/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
+++ b/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
@@ -122,7 +122,7 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
                 _critialDamage = true;
             };
             if ([_unit, _bodyPartNToAdd, _bodyPartDamage, _woundDamage] call FUNC(determineIfFatal)) then {
-                if (!isPlayer _unit || random 1 <= EGVAR(medical,deathChance)) then
+                if (!isPlayer _unit || {random 1 <= EGVAR(medical,deathChance)}) then
                     TRACE_1("determineIfFatal returned true",_woundDamage);
                     [QEGVAR(medical,FatalInjury), _unit] call CBA_fnc_localEvent;
                 };
--- a/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
+++ b/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
@@ -122,7 +122,7 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
                 _critialDamage = true;
             };
             if ([_unit, _bodyPartNToAdd, _bodyPartDamage, _woundDamage] call FUNC(determineIfFatal)) then {
-                if (!isPlayer || random 1 <= EGVAR(medical,deathChance)) then
+                if (!isPlayer _unit || random 1 <= EGVAR(medical,deathChance)) then
                     TRACE_1("determineIfFatal returned true",_woundDamage);
                     [QEGVAR(medical,FatalInjury), _unit] call CBA_fnc_localEvent;
                 };
--- a/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
+++ b/addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
@@ -122,7 +122,7 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
                 _critialDamage = true;
             };
             if ([_unit, _bodyPartNToAdd, _bodyPartDamage, _woundDamage] call FUNC(determineIfFatal)) then {
-                if (isPlayer _unit && {random 1 > EGVAR(medical,survivalChance)}) exitWith {
+                if (isPlayer _unit && {random 1 > EGVAR(medical,survivalChance)}) then {
                     TRACE_1("determineIfFatal returned true",_woundDamage);
                     [QEGVAR(medical,FatalInjury), _unit] call CBA_fnc_localEvent;
                 };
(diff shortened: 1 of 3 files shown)


BrettMayson:
I am aware as soon as I pushed that AI is now invincible. I am declaring this a feature, not a bug.

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -122,8 +122,10 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
                 _critialDamage = true;
             };
             if ([_unit, _bodyPartNToAdd, _bodyPartDamage, _woundDamage] call FUNC(determineIfFatal)) then {
-                TRACE_1("determineIfFatal returned true",_woundDamage);
-                [QEGVAR(medical,FatalInjury), _unit] call CBA_fnc_localEvent;
+                if !(isPlayer _unit && {random 1 > EGVAR(medical,deathChance)}) then {
```
jonpas, line 125: `true && true -> !true -> false`, so if player and death chance above, it won't cause fatal injury? This seems reversed.

kymckay: De Morgan's law, the code as is can be simplified to:

!isPlayer || random 1 <= deathChance

kymckay:
Even though it's effectively the same as #8170 but inverse logic, it does feel like this makes more sense intuitively and if there's demand for it I do prefer this method to the other.

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
jonpas: You are correct, I was still reading it as survivability chance. If death chance is 0.15 then random value has to be below that, not above, so you get `true && false` when it should happen actually. :D

I would prefer inverting that with De Morgan's law as you state, easier to read. But that may just be me still stuck with "survival chance" making more sense in code, but probably not to the user.

Italiandrache:
> I am aware as soon as I pushed that AI is now invincible. I am declaring this a feature, not a bug.

What do you mean?

BrettMayson:
> 
> 
> > I am aware as soon as I pushed that AI is now invincible. I am declaring this a feature, not a bug.
> 
> What do you mean?

It was a joke since the first commit had that oversight, it's good now

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
jonpas: @BrettMayson Can we reverse that? Then I am happy to merge.

BrettMayson: Yeah sure I can invert it

jonpas (approved):

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -122,8 +122,10 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
                 _critialDamage = true;
             };
             if ([_unit, _bodyPartNToAdd, _bodyPartDamage, _woundDamage] call FUNC(determineIfFatal)) then {
-                TRACE_1("determineIfFatal returned true",_woundDamage);
-                [QEGVAR(medical,FatalInjury), _unit] call CBA_fnc_localEvent;
+                if (!isPlayer || random 1 <= EGVAR(medical,deathChance)) then
```
veteran29, line 125:
```suggestion
                if (!isPlayer _unit || random 1 <= EGVAR(medical,deathChance)) then
```

BrettMayson: ah lol, thanks. I just copy pasted from the comment on my phone and didn't notice

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -122,7 +122,7 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
                 _critialDamage = true;
             };
             if ([_unit, _bodyPartNToAdd, _bodyPartDamage, _woundDamage] call FUNC(determineIfFatal)) then {
-                if !(isPlayer _unit && {random 1 > EGVAR(medical,deathChance)}) then {
+                if (!isPlayer || random 1 <= EGVAR(medical,deathChance)) then
```
kymckay:
```suggestion
                if (!isPlayer || random 1 < EGVAR(medical,deathChance)) then
```

Very minor thing, but something I alluded to in my other comment. Arma's random operation is exclusive of the upper bound and inclusive of the lower which is 0, so for the setting to be the true chance the comparison should just be less than (consider the edge cases and it'll become clear why).

jonpas: Added missing opening brace too.

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -122,8 +122,10 @@ private _bodyPartVisParams = [_unit, false, false, false, false]; // params arra
                 _critialDamage = true;
             };
             if ([_unit, _bodyPartNToAdd, _bodyPartDamage, _woundDamage] call FUNC(determineIfFatal)) then {
-                TRACE_1("determineIfFatal returned true",_woundDamage);
-                [QEGVAR(medical,FatalInjury), _unit] call CBA_fnc_localEvent;
+                if (!isPlayer _unit || {random 1 <= EGVAR(medical,deathChance)}) then {
```
kymckay, line 125:
```suggestion
                if (!isPlayer _unit || {random 1 < EGVAR(medical,deathChance)}) then {
```

veteran29 (approved):

kymckay (approved):

PR #8203, opened 2021-04-09
Cargo - paradrop fixes

### LINKED ISSUE #8200: ace_cargoUnload wrong _item value
- **Arma 3:** irrelevant - ace code issue
- **CBA:** ` irrelevant - ace code issue
- **ACE3:** master branch and release

**Description:**
"ace_cargoUnloaded" eh returns classname instead of an object if the item was added using [classname,vehicle,count] call ace_cargo_fnc_addCargoItem;

**Steps to reproduce:**
1. create eh to monitor "ace_cargoUnloaded" and display its arguments;
2. add item to cargo using aforementioned command;
3. paradrop the item
4. obvserve classname being returned instead of an object.
 
 **Expected behavior:**
object to be returned.

**Additional context:**
Can be statically detected.
dropped object is stored in _itemObject but eh returns _item which can contain classname.

LupusTheCanine:
- Cargo - Improve make event "ace_cargoUnloaded" always pass object instead of sometimes passing classname.
- Cargo - Change move "ace_cargoUnloaded" event invocation from "ace_unloadItem" eh to function unloadItem to facilitate aforementioned change.

Fixes #8200

### DIFF
--- a/addons/cargo/functions/fnc_paradropItem.sqf
+++ b/addons/cargo/functions/fnc_paradropItem.sqf
@@ -38,7 +38,7 @@ TRACE_1("",_distBehind);
 private _posBehindVehicleAGL = _vehicle modelToWorld [0, _distBehind, -2];
 
 
-private _object=_item;
+private _object = _item;
 if (_item isEqualType objNull) then {
     detach _item;
     // hideObjectGlobal must be executed before setPos to ensure light objects are rendered correctly
--- a/addons/cargo/functions/fnc_paradropItem.sqf
+++ b/addons/cargo/functions/fnc_paradropItem.sqf
@@ -107,6 +107,6 @@ if (_showHint) then {
 };
 
 // Invoke listenable event
-["ace_cargoUnloaded", [_item, _vehicle, "paradrop"]] call CBA_fnc_globalEvent;
+["ace_cargoUnloaded", [_itemObject, _vehicle, "paradrop"]] call CBA_fnc_globalEvent;
 
 true
--- a/addons/cargo/functions/fnc_paradropItem.sqf
+++ b/addons/cargo/functions/fnc_paradropItem.sqf
@@ -38,19 +38,18 @@ TRACE_1("",_distBehind);
 private _posBehindVehicleAGL = _vehicle modelToWorld [0, _distBehind, -2];
 
 
-private _itemObject = if (_item isEqualType objNull) then {
+private _object=_item;
+if (_item isEqualType objNull) then {
     detach _item;
     // hideObjectGlobal must be executed before setPos to ensure light objects are rendered correctly
     // do both on server to ensure they are executed in the correct order
     [QGVAR(serverUnload), [_item, _posBehindVehicleAGL]] call CBA_fnc_serverEvent;
-    _item
 } else {
-    private _newItem = createVehicle [_item, _posBehindVehicleAGL, [], 0, "NONE"];
-    _newItem setPosASL (AGLtoASL _posBehindVehicleAGL);
-    _newItem
+    _object = createVehicle [_item, _posBehindVehicleAGL, [], 0, "NONE"];
+    _object setPosASL (AGLtoASL _posBehindVehicleAGL);
 };
 
-_itemObject setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vectorDir _vehicle)) vectorMultiply -5));
+_object setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vectorDir _vehicle)) vectorMultiply -5));
 
 // open parachute and ir light effect
 [{
@@ -74,7 +73,7 @@ _itemObject setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vecto
         _light attachTo [_item, [0,0,0]];
     };
 
-}, [_itemObject], 0.7] call CBA_fnc_waitAndExecute;
+}, [_object], 0.7] call CBA_fnc_waitAndExecute;
 
 // smoke effect when crate landed
 [{
@@ -93,7 +92,7 @@ _itemObject setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vecto
         [_this select 1] call CBA_fnc_removePerFrameHandler;
     };
 
-}, 1, [_itemObject]] call CBA_fnc_addPerFrameHandler;
+}, 1, [_object]] call CBA_fnc_addPerFrameHandler;
 
 if (_showHint) then {
     [
@@ -107,6 +106,6 @@ if (_showHint) then {
 };
 
 // Invoke listenable event
-["ace_cargoUnloaded", [_itemObject, _vehicle, "paradrop"]] call CBA_fnc_globalEvent;
+["ace_cargoUnloaded", [_object, _vehicle, "paradrop"]] call CBA_fnc_globalEvent;
 
 true
--- a/addons/cargo/functions/fnc_paradropItem.sqf
+++ b/addons/cargo/functions/fnc_paradropItem.sqf
@@ -40,10 +40,10 @@ private _posBehindVehicleAGL = _vehicle modelToWorld [0, _distBehind, -2];
 
 private _object = _item;
 if (_item isEqualType objNull) then {
-    detach _item;
+    detach _object;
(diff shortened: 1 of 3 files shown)


addons/cargo/functions/fnc_unloadItem.sqf
```diff
@@ -47,14 +47,17 @@ private _space = [_vehicle] call FUNC(getCargoSpaceLeft);
...
+    _item;
 } else {
     private _newItem = createVehicle [_item, _emptyPosAGL, [], 0, "NONE"];
     _newItem setPosASL (AGLtoASL _emptyPosAGL);
+    _newItem;
 };
-
+// Invoke listenable event
+["ace_cargoUnloaded", [_itemObject, _vehicle, "unload"]] call CBA_fnc_globalEvent;
 true
```
commy2, line 63: 1. What is an "itemObject"? Why name it "itemX" when it explictily is not an "item" (i.e. classname), but an "object"?
2. Do not use the rv on control structures.

```suggestion
private _object = _item;

if (_item isEqualType objNull) then {
    detach _item;
    // hideObjectGlobal must be executed before setPos to ensure light objects are rendered correctly
    // do both on server to ensure they are executed in the correct order
    [QGVAR(serverUnload), [_item, _emptyPosAGL]] call CBA_fnc_serverEvent;
} else {
    _object = createVehicle [_item, _emptyPosAGL, [], 0, "NONE"];
    _object setPosASL (AGLtoASL _emptyPosAGL);
};

// Invoke listenable event
["ace_cargoUnloaded", [_object, _vehicle, "unload"]] call CBA_fnc_globalEvent;
true
```

LupusTheCanine: 1.1 _itemObject is a place for a reference to an object that is being unloaded, that was either just created or moved into position if it already existed.
1.2 I called the variable the same way it was called in paradropItem because it does exactly the same thing.
2. ok, should I fix paradropItem to follow this rule?

commy2: That would be sweet, because it sounds like my criticisms extend to the old code as well. Although I would be fine with just not writing new non-sense. Perhaps a second pull could be made to sanitize the other code, or maybe it can be made to fit into this one. Up to you if you want to do that.

LupusTheCanine: all done

addons/cargo/functions/fnc_paradropItem.sqf
```diff
@@ -38,19 +38,18 @@ TRACE_1("",_distBehind);
 private _posBehindVehicleAGL = _vehicle modelToWorld [0, _distBehind, -2];
 
 
-private _itemObject = if (_item isEqualType objNull) then {
+private _object=_item;
```
commy2, line 41:
```suggestion
private _object = _item;
```

addons/cargo/functions/fnc_unloadItem.sqf
```diff
@@ -47,14 +47,16 @@ private _space = [_vehicle] call FUNC(getCargoSpaceLeft);
 private _itemSize = [_item] call FUNC(getSizeItem);
 _vehicle setVariable [QGVAR(space), (_space + _itemSize), true];
 
+private _object=_item;
```
commy2, line 50:
```suggestion
private _object = _item;
```

addons/cargo/functions/fnc_paradropItem.sqf
```diff
@@ -74,7 +73,7 @@ _itemObject setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vecto
         _light attachTo [_item, [0,0,0]];
```
commy2, line 73: In this block, rename _item to _object.

addons/cargo/functions/fnc_paradropItem.sqf
```diff
@@ -38,19 +38,18 @@ TRACE_1("",_distBehind);
 private _posBehindVehicleAGL = _vehicle modelToWorld [0, _distBehind, -2];
 
 
-private _itemObject = if (_item isEqualType objNull) then {
+private _object = _item;
+if (_item isEqualType objNull) then {
```
commy2, line 42:
```suggestion
if (_object isEqualType objNull) then {
```

addons/cargo/functions/fnc_paradropItem.sqf
```diff
@@ -38,19 +38,18 @@ TRACE_1("",_distBehind);
...
     detach _item;
     // hideObjectGlobal must be executed before setPos to ensure light objects are rendered correctly
     // do both on server to ensure they are executed in the correct order
     [QGVAR(serverUnload), [_item, _posBehindVehicleAGL]] call CBA_fnc_serverEvent;
```
commy2, line 46: In this block, use _item in place of _object.

addons/cargo/functions/fnc_paradropItem.sqf
```diff
@@ -93,20 +92,20 @@ _itemObject setVelocity ((velocity _vehicle) vectorAdd ((vectorNormalized (vecto
         [_this select 1] call CBA_fnc_removePerFrameHandler;
     };
```
commy2, line 94: In this block, rename _item to _object.

addons/cargo/functions/fnc_unloadItem.sqf
```diff
@@ -47,14 +47,16 @@ private _space = [_vehicle] call FUNC(getCargoSpaceLeft);
 private _itemSize = [_item] call FUNC(getSizeItem);
 _vehicle setVariable [QGVAR(space), (_space + _itemSize), true];
 
+private _object = _item;
 if (_item isEqualType objNull) then {
     detach _item;
     // hideObjectGlobal must be executed before setPos to ensure light objects are rendered correctly
     // do both on server to ensure they are executed in the correct order
     [QGVAR(serverUnload), [_item, _emptyPosAGL]] call CBA_fnc_serverEvent;
 } else {
```
commy2, line 56: In this block, use _item in place of _object.

commy2:
lgtm besides the horrible variable naming that is not your fault.

addons/cargo/functions/fnc_unloadItem.sqf
LupusTheCanine: done

addons/cargo/functions/fnc_paradropItem.sqf
LupusTheCanine: done

addons/cargo/functions/fnc_paradropItem.sqf
LupusTheCanine: done

LupusTheCanine:
All naming fixed in unload and paradrop.

LupusTheCanine:
missed one spot where there was _item instead of _object, fixed.

commy2:
Somebody test and approve this.

LupusTheCanine:
> Somebody test and approve this.

I can do MP testing with friends, do you see any things we should focus besides
- event handlers
- crates behavior after drop

commy2:
Just watch the crates and maybe add the evs with simple systemChat's to check if they fire, fire with the correct args, and not double fire or something.

LupusTheCanine:
@commy2 
I found another inconsistency that might be hard to fix.
event "ace_cargoUnloaded"
should be QGVAR(unloaded) 
but this would cause expanded name to go from
"ace_cargoUnloaded" to "ace_cargo_unloaded" which might break backwards compatibility for paradrop, not an issue for unload as it didn't have the event in the first place.

LupusTheCanine:
> Just watch the crates and maybe add the ehs with simple systemChat's to check if they fire, fire with the correct args, and not double fire or something.

We did the testing and event handlers behave correctly. No double fire, properly called on all machines with correct arguments.
Crates behaved as expected, not inheriting velocity when dropped by non owner.

commy2:
> Then why "serverUnloaded" is done using a macro?

The one with the macro is private (only used in the component), the other public API.

LupusTheCanine:
@commy2 is it ready for merge?
I would like to make next pull request with a fix for drop velocity but it depends on this one.

commy2:
Sorry, ask someone on Slack with more time than me.

LupusTheCanine:
@Glowbal
I would like to ask for a review.

PabstMirror (approved):
lgtm

PR #8209, opened 2021-04-16
Night Vision - Add color variants to all generations

JoramD0:
- Add 3 default color variants (black, brown & green) to all gens/versions of NVGs.


jonpas (approved):
Looks good to me, but @PabstMirror double-check.

mharis001 (commented):
Should add new NVG class names to `CfgPatches` `weapons` array.

PR #8212, opened 2021-04-17
Medical Feedback - Fix Injured Sounds Not Playing At Altitudes Above 70

### LINKED ISSUE #8208: Injury sounds not playing at higher altitudes
When the player is above a certain altitude, the injury sounds from `ace_medical_feedback_fnc_playInjuredSound` do no longer play locally. They are probably also not broadcasted to other players in the vicinity.

The reason is that `inAreaArray` (introduced to the ACE function in commit e9c1f5e18fe99b65c88eb882fbd00f879f11d408) seems to take a 3DPosition in AGL format while being provided a 3DPosition in the raw world coordinate system. This causes the array of target players to possibly be completely empty if the altitude is higher than the distance the sound is supposed to be played at.

This can be confirmed by checking the content of the following array while standing on top of a higher building's roof or while swimming in deeper water and by changing the function to get the player's position:
`allPlayers inAreaArray [getPosWorld player, 2, 2, 0, false, 2];`
(replace `getPosWorld` with `getPos`, `getPosATL`, `getPosASL`, `getPosASLW`, ... - see https://community.bistudio.com/wiki/Position)

For a working solution this array should contain the player under all circumstances.

A solution might be a change from `getPosWorld _unit` to `ASLToAGL getPosASL _unit`, as this seems to be the easiest way to get a position in AGL format.

commy2:
- `inAreaArray` takes AGL if z-value is provided.
- `playInjuredSound` gives it ASL, which means only units above z/2 sea level are detected as nearby.
- This PR changes the provided position from ASL to AGL in the most straight forward way possible.
- fix #8208

As an aside, this should've never checked the z in the first place. You're wasting more cycles than you will ever save iterating over units 3 stories above you. Just let the distance attenuation handle heights...

### DIFF
--- a/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
+++ b/addons/medical_feedback/functions/fnc_playInjuredSound.sqf
@@ -33,7 +33,7 @@ private _distance = if (_type == "hit") then {
 } else {
     [10, 15, 20] select _severity;
 };
-private _targets = allPlayers inAreaArray [getPosWorld _unit, _distance, _distance, 0, false, _distance];
+private _targets = allPlayers inAreaArray [ASLToAGL getPosASL _unit, _distance, _distance, 0, false, _distance];
 if (_targets isEqualTo []) exitWith {};
 
 // Handle timeout


commy2:
Btw, the geometric shape the command checks is an elliptic cylinder, where the origin is at the center of the cylinder. This seems to me very unintuitive. Also, half of the checked volume will always be under ground.

This z-parameter is just whack and should, again, be ignored.

PabstMirror (approved):

PR #8216, opened 2021-04-19
MicroDAGR - Add mode switching keybinds

JoramD0:
- Add 2 keybinds to cylce through the MicroDAGR modes/pages.
- Changed ERROR to LOG in fnc_saveCurrentAndSetNewMode to stop it from spamming the RPT, internal function anyway.

### DIFF
--- a/addons/microdagr/XEH_clientInit.sqf
+++ b/addons/microdagr/XEH_clientInit.sqf
@@ -20,16 +20,16 @@ private _closeCode = {
 // Mode keybinds:
 ["ACE3 Equipment", QGVAR(previousMode), LLSTRING(previousMode), {
     private _newMode = GVAR(currentApplicationPage) - 1;
-    if (_newMode < 0) then {
-        _newMode = 6;
+    if (_newMode < APP_MODE_INFODISPLAY) then {
+        _newMode = APP_MODE_SETUP;
     };
     [_newMode] call FUNC(saveCurrentAndSetNewMode);
 }, ""] call CBA_fnc_addKeybind;
 
 ["ACE3 Equipment", QGVAR(nextMode), LLSTRING(nextMode), {
     private _newMode = GVAR(currentApplicationPage) + 1;
-    if (_newMode > 6) then {
-        _newMode = 0;
+    if (_newMode > APP_MODE_SETUP) then {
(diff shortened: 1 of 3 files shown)


commy2:
"Spamming the log"?

bux:
Why is it spamming the log? Better to find the issue for that than to silence a probably valid log.

JoramD0:
Poor choice of words, the line I changed is a check to see if there is a display, which used to write an error to the RPT if there wasn't any. I can't imagine it would be beneficial to do the same check twice so I changed it to LOG so it would still error with full debug mode enabled.

jonpas (approved):
That error is rather pointless there. Condition is now used to stop execution on keybinds if MicroDAGR is not opened, rather than duplicating it.

Simple and short.

mharis001 (commented):
Works well. 

Can we get this to keep cycling between modes once the "end" for the keybind's direction is reached? For example, for the next mode keybind, after mode 6, it goes back to mode 0 (and the opposite for previous mode keybind). Essentially, mimic the behaviour of ALT-TABing and ALT-SHIFT-TABing.

jonpas:
We could, I believe that was a design decision by @JoramD0. And I kind of like it, but I also see the point now that you could only use one button to cycle between all pages and wouldn't need to bind both. :thinking:

JoramD0:
It was indeed a design choice, I could do either but I honestly like the way it is currently.

Could also go all out and somewhat overcomplicate it and add a setting

mharis001 (approved):

bux:
Where does the magic number for the mode come from? A variable/constant/define would make sense for it.

JoramD0:
@bux Do you mean these? https://github.com/acemod/ACE3/blob/master/addons/microdagr/script_component.hpp#L26-L33

commy2:
Yes, although that could be considered out of scope for this PR. Up to you.

JoramD0:
Okay, I forgot to use the defines which was my last commit.

I am still confused as to what you guys want me to do.

bux:
Way better, thank you.

Having a seemingly random number appear in code makes the next person to work on it wonder, why someone used this number in particular.

PabstMirror (approved):

Issue #8221, opened 2021-04-21
[ACE Compat] RHS AFRF Compat RPK74M uses wrong ammo


Dabako:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.02 stable` 
- **CBA:** `3.15.1 stable` 
- **ACE3:** `3.13.6 stable` 
- **ACE Compat - RHS AFRF** `3.13.6`
- **RHSAFRF** `0.5.5 stable`

**Description:**
When I spawn an Automatic Rifleman with the RPK74M the Ammunition the Weapon uses wrong ammo.
It uses 7.62x54mm but it should use 5.45x39mm.
Without ACE Compat the RPK74M uses 5.45x39mm

**Steps to reproduce:**

1. Editor or Zeus
2. Spawn Automatic Rifleman MSV
3. Click play
4. Try to shoot or reload

**Expected behavior:**
You should be able to shoot or reload

**Where did the issue occur?**
- Dedicated / Editor (Singleplayer) 

**Log Files:**
- https://gist.github.com/Dabako/93d919412097ad238b5580aea2666174

**Additional context:**
I would try to solve the problem myself and make a pull request but I don't know where to start. If someone can give me information I could try to help

**Screenshots:**
![20210421073036_1](https://user-images.githubusercontent.com/15671777/115507724-c46c2c00-a27c-11eb-936e-445620625a32.jpg)
![20210421073027_2](https://user-images.githubusercontent.com/15671777/115507752-cfbf5780-a27c-11eb-907a-48b187bd2a84.jpg)

PabstMirror:
new rhs update yesterday, 
`Updating base class rhs_weap_rpk74_base->rhs_weap_pkp, by z\ace\addons\compat_rhs_afrf3\config.cpp/CfgWeapons/rhs_weap_rpk74/ `

Dabako:
Okay I did not even realize that it was part of an new RHS Update, I was trying to fix it myself for the last hour. But I can't seem to get the ACE Debug build to work. And I realized that I dont even know how to build the optionals only

bux:
@Dabako Better join our public chat for questions like that:
https://slackin.ace3mod.com/

topden:
you have the wrong inheritance of the rpk class. now it has changed. 
wrong:
    class rhs_weap_rpk74: rhs_weap_pkp {
        ACE_barrelTwist = 195.072;
        ACE_barrelLength = 589.28;
    };
    
[...]

PR #8223, opened 2021-04-21
RHS ARF Compat - Fix RPK74M wrong ammo bug

Dabako:
- This patch will fix the bug #8221
- Add of RPK74M in optionals/compat_rhs_afrf3/CfgWeapons.hpp

I have tested it locally and it worked as intended


reyhard:
class rhs_weap_rpk74: rhs_weap_pkp { 
is still there so it will be still broken, just slightly less

Dabako:
I changed that in the second commit, I just checked it and it should not be there

optionals/compat_rhs_afrf3/CfgWeapons.hpp
```diff
@@ -57,6 +57,10 @@ class CfgWeapons {
     class rhs_weap_svds_npz: rhs_weap_svds {
         ACE_RailHeightAboveBore = 4.3348;
     };
+    class rhs_weap_rpk74_base;
+    class rhs_weap_rpk74m : rhs_weap_rpk74_base {
```
PabstMirror, line 61: After looking at their configs I think `class rhs_weap_rpk74_base: rhs_weap_rpk_base` would be a better choice
and probably keep the `ACE_barrelTwist = 195.072;` config from before as well

PabstMirror (changes requested):

optionals/compat_rhs_afrf3/CfgWeapons.hpp
Dabako: Changed in newest commit, tested it and it works for me too

PabstMirror (approved):

PR #8225, opened 2021-04-21
Vehiclelock - Improve setting name

### LINKED ISSUE #6828: Vehiclelock -1 state on setting is misleading
**Arma 3 Version:** stable
**CBA Version:** stable
**ACE3 Version:** stable

**Mods:**
```
- CBA_A3
- ace
```

**Description:**
The first setting is "Default" but i should be named "vehiclelock off" or "off".
If you dont know that it didnt init the vehiclelock on vehicles where is the difference between "Default" and "As is". In my opinion "Default" is misleading.

JoramD0:
- Improve `Vehicle Starting Lock State` setting names.

Renamed the "Default" setting to "As Is" and changed the current "As Is" to "Remove Ambiguous Lock State" to avoid confusion.

closes #6828

### DIFF
--- a/addons/vehiclelock/initSettings.sqf
+++ b/addons/vehiclelock/initSettings.sqf
@@ -23,7 +23,7 @@
     QGVAR(vehicleStartingLockState), "LIST",
     [LSTRING(VehicleStartingLockState_DisplayName), LSTRING(VehicleStartingLockState_Description)],
     LSTRING(DisplayName),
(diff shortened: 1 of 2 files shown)


mharis001 (approved):

PabstMirror (approved):

Drofseh:
"Remove Ambiguous Lock State" doesn't make any sense to me.
What is the ambiguous lock state that gets removed? What replaces it?

JoramD0:
@Drofseh 
[Here](https://github.com/acemod/ACE3/blob/master/addons/vehiclelock/functions/fnc_handleVehicleInitPost.sqf#L25) it eliminates the ambiguous 1-"Default" and 3-"Locked for Player" states.

Basically, it relocks vehicles that have the lock state in 3den set to "Locked" and "Locked for Players".

PR #8226, opened 2021-04-21
Parachute - Add failure chance

JoramD0:
- Add setting to enable percentage based failing of parachutes (0% chance by default).
- Add config value to `ParachuteBase` to set the failure delay.
- Add documentation for the config value mentioned above.

This was [a request](https://github.com/acemod/ACE3/issues/3594#issuecomment-546740178) from @DevastatorCM

[A small demo](https://youtu.be/nrmtXdw9ooU)

### DIFF
--- a/addons/parachute/functions/fnc_handleFailureChance.sqf
+++ b/addons/parachute/functions/fnc_handleFailureChance.sqf
@@ -11,16 +11,16 @@
  * None
  *
  * Example:
- * [player, vehicle player] call FUNC(handleFailureChance);
+ * [player, vehicle player] call FUNC(handleFailureChance)
  *
  * Public: No
  */
 
 params ["_unit", "_vehicle"];
 
-if !(_vehicle isKindOf "ParachuteBase" || {GVAR(failureChance) > 0}) exitWith {};
+if !(GVAR(failureChance) > 0 || {_vehicle isKindOf "ParachuteBase"}) exitWith {};
 
 if (random 1 < GVAR(failureChance)) then {
-    private _failureDelay = getNumber (configFile >> "CfgVehicles" >> str typeOf _vehicle >> QGVAR(failureDelay))
+    private _failureDelay = getNumber (configOf _vehicle >> QGVAR(failureDelay));
     [{[(_this select 0), (_this select 1)] call FUNC(cutParachute)}, [_unit, _vehicle], _failureDelay] call cba_fnc_waitAndExecute;
 };
--- a/addons/parachute/functions/fnc_handleFailureChance.sqf
+++ b/addons/parachute/functions/fnc_handleFailureChance.sqf
@@ -5,22 +5,22 @@
  *
  * Arguments:
  * 0: Unit <OBJECT>
- * 0: Vehicle <OBJECT>
+ * 1: Vehicle <OBJECT>
  *
  * Return Value:
  * None
  *
  * Example:
- * [player, vehicle player] call FUNC(handleFailureChance)
+ * [player, vehicle player] call ace_parachute_handleFailureChance
  *
  * Public: No
  */
 
 params ["_unit", "_vehicle"];
 
-if !(GVAR(failureChance) > 0 || {_vehicle isKindOf "ParachuteBase"}) exitWith {};
+if !(_vehicle isKindOf "ParachuteBase") exitWith {};
 
 if (random 1 < GVAR(failureChance)) then {
     private _failureDelay = getNumber (configOf _vehicle >> QGVAR(failureDelay));
-    [{[(_this select 0), (_this select 1)] call FUNC(cutParachute)}, [_unit, _vehicle], _failureDelay] call cba_fnc_waitAndExecute;
+    [FUNC(cutParachute), [_unit, _vehicle], _failureDelay] call CBA_fnc_waitAndExecute;
 };
--- a/addons/parachute/functions/fnc_handleFailureChance.sqf
+++ b/addons/parachute/functions/fnc_handleFailureChance.sqf
@@ -1,0 +1,26 @@
+#include "script_component.hpp"
+/*
+ * Author: JoramD
+ * Handles percentage chance parachute failure.
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ * 0: Vehicle <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player, vehicle player] call FUNC(handleFailureChance);
+ *
+ * Public: No
+ */
+
+params ["_unit", "_vehicle"];
+
+if !(_vehicle isKindOf "ParachuteBase" && {GVAR(failureChance) > 0}) exitWith {};
+
+if (random 1 < GVAR(failureChance)) then {
+    private _failureDelay = getNumber (configFile >> "CfgVehicles" >> str typeOf _vehicle >> QGVAR(failureDelay))
+    [{[(_this select 0), (_this select 1)] call FUNC(cutParachute)}, [_unit, _vehicle], _failureDelay] call cba_fnc_waitAndExecute;
+};
--- a/addons/parachute/functions/fnc_handleFailureChance.sqf
+++ b/addons/parachute/functions/fnc_handleFailureChance.sqf
@@ -11,7 +11,7 @@
  * None
  *
  * Example:
- * [player, vehicle player] call ace_parachute_handleFailureChance
+ * [player, vehicle player] call ace_parachute_fnc_handleFailureChance
  *
  * Public: No
  */
--- a/addons/parachute/functions/fnc_handleFailureChance.sqf
+++ b/addons/parachute/functions/fnc_handleFailureChance.sqf
@@ -18,7 +18,7 @@
 
 params ["_unit", "_vehicle"];
 
-if !(_vehicle isKindOf "ParachuteBase" && {GVAR(failureChance) > 0}) exitWith {};
+if !(_vehicle isKindOf "ParachuteBase" || {GVAR(failureChance) > 0}) exitWith {};
 
 if (random 1 < GVAR(failureChance)) then {
     private _failureDelay = getNumber (configFile >> "CfgVehicles" >> str typeOf _vehicle >> QGVAR(failureDelay))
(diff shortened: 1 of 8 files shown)


addons/parachute/functions/fnc_handleFailureChance.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Example:
+ * [player, vehicle player] call FUNC(handleFailureChance);
+ *
+ * Public: No
+ */
+
+params ["_unit", "_vehicle"];
+
+if !(_vehicle isKindOf "ParachuteBase" || {GVAR(failureChance) > 0}) exitWith {};
```
jokoho48, line 21: I personally would flip those conditions because checking if a value is bigger than 0 should be faster than checking if a object is a type
```suggestion
if !(GVAR(failureChance) > 0 || {_vehicle isKindOf "ParachuteBase"}) exitWith {};
```

BrettMayson:
I think it'd be neat if the parachute could be faulty, rather than just get cut

addons/parachute/functions/fnc_handleFailureChance.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ * 0: Vehicle <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player, vehicle player] call FUNC(handleFailureChance);
```
jonpas, line 14:
```suggestion
 * [player, vehicle player] call FUNC(handleFailureChance)
```

addons/parachute/functions/fnc_handleFailureChance.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Public: No
+ */
+
+params ["_unit", "_vehicle"];
+
+if !(_vehicle isKindOf "ParachuteBase" || {GVAR(failureChance) > 0}) exitWith {};
+
+if (random 1 < GVAR(failureChance)) then {
+    private _failureDelay = getNumber (configFile >> "CfgVehicles" >> str typeOf _vehicle >> QGVAR(failureDelay))
```
jonpas, line 24:
```suggestion
    private _failureDelay = getNumber (configFile >> "CfgVehicles" >> typeOf _vehicle >> QGVAR(failureDelay))
```
`typeOf` already returns string.

jonpas (changes requested):

addons/parachute/functions/fnc_handleFailureChance.sqf
jokoho48: just use configOf
```suggestion
    private _failureDelay = getNumber (configOf _vehicle >> QGVAR(failureDelay))
```

addons/parachute/functions/fnc_handleFailureChance.sqf
JoramD0: Thank you for the suggestion!

DevastatorCM:
It would be nice if there were some variables which effects the failure ratio given initialy. For instance
- altitude of opening (or player's dive speed) . Assumption here is HAHO where lets say player opens the chute above 1500-2000m and as air density lets say is low, speed of the dive is high and it can increase the chance of failure 
- body position: if player is not stable but diving (via "W") chance to failure increases due to wrong body position
- hit to an object or another player while chute is open: Tangles the chute so failure rate if such hit happens to the chute is 100%

JoramD0:
@DevastatorCM 

- Possible however not worth it since it is a very specific case.
- Can't think of a way to do that without impacting performance for such a small feature.
- Technically not possible as far as I'm aware.

DevastatorCM:
1- This one is the easy one I think. Simple multiplier of the AGL value when chute is opened
2- This one can be done by checking if move forward key is being pressed when the open chute action kicked in
3- I didnt chek this but if chute is an object, then https://community.bistudio.com/wiki/Arma_3:_Event_Handlers#EpeContact can be used on the chute maybe

JoramD0:
@DevastatorCM You can always open a PR yourself, I currently don't have the time/want to implement all that (if even possible without a performance impact).

JoramD0:
@Drofseh 
Do you mean adding a backup parachute or the option to only have the primary parachute fail?

jokoho48:
> 
> 
> > I think this should really add some options for a backup chute.
> > Right now all this seems to do is provide a random chance to ruin a player's experience by killing them at mission start.
> 
> There is already reserve chutes within ACE, unless if I am misunderstanding you https://github.com/acemod/ACE3/blob/master/addons/parachute/CfgVehicles.hpp

yes there is a backup chute

addons/parachute/functions/fnc_handleFailureChance.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Example:
+ * [player, vehicle player] call FUNC(handleFailureChance)
+ *
+ * Public: No
+ */
+
+params ["_unit", "_vehicle"];
+
+if !(GVAR(failureChance) > 0 || {_vehicle isKindOf "ParachuteBase"}) exitWith {};
```
PabstMirror, line 21:
```suggestion
if !(GVAR(failureChance) > 0 && {_vehicle isKindOf "ParachuteBase"}) exitWith {};
```

PabstMirror: imho this is more readable
```suggestion
if (GVAR(failureChance) == 0 || {!(_vehicle isKindOf "ParachuteBase")}) exitWith {};
```

commy2: De Morgan says you need to change or to and.

commy2, line 21:
```suggestion
if !(_vehicle isKindOf "ParachuteBase") exitWith {};
```
There is no reason to exit early if chance is 0. That literally just adds overhead on average.

addons/parachute/functions/fnc_handleFailureChance.sqf
```diff
@@ -0,0 +1,26 @@
...
+
+if (random 1 < GVAR(failureChance)) then {
+    private _failureDelay = getNumber (configOf _vehicle >> QGVAR(failureDelay));
+    [{[(_this select 0), (_this select 1)] call FUNC(cutParachute)}, [_unit, _vehicle], _failureDelay] call cba_fnc_waitAndExecute;
```
commy2, line 25:
```suggestion
    [FUNC(cutParachute), [_unit, _vehicle], _failureDelay] call CBA_fnc_waitAndExecute;
```

addons/parachute/functions/fnc_handleFailureChance.sqf
```diff
@@ -0,0 +1,26 @@
+#include "script_component.hpp"
+/*
+ * Author: JoramD
+ * Handles percentage chance parachute failure.
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ * 0: Vehicle <OBJECT>
```
commy2, line 8: Two 0?

addons/parachute/functions/fnc_handleFailureChance.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player, vehicle player] call FUNC(handleFailureChance)
```
commy2, line 14: Iirc, examples used to use expanded function names?

docs/wiki/framework/parachute-framework.md
```diff
@@ -22,3 +22,13 @@ class CfgVehicles {
...
 ```
+
+## 2. Adding failure cut delay
+
+```cpp
+class CfgVehicles {
+    class BananaParachute {
+        ace_parachute_failureDelay = 2;  // Add delay before parachute fails (time in seconds)
+    };
+};
```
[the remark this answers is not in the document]
JoramD0: Would you like me to change it for both examples?

addons/parachute/functions/fnc_handleFailureChance.sqf
```diff
@@ -5,22 +5,22 @@
...
  * 0: Unit <OBJECT>
- * 0: Vehicle <OBJECT>
+ * 1: Vehicle <OBJECT>
  *
  * Return Value:
  * None
  *
  * Example:
- * [player, vehicle player] call FUNC(handleFailureChance)
+ * [player, vehicle player] call ace_parachute_handleFailureChance
```
jonpas, line 14:
```suggestion
 * [player, vehicle player] call ace_parachute_fnc_handleFailureChance
```

jonpas (changes requested):

docs/wiki/framework/parachute-framework.md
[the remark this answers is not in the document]
jonpas: Yes do it, the entire component relies on `ParachuteBase` anyways.

jonpas (approved):

PabstMirror (approved):
Maybe in a future PR we could add a sound effect or some way of communicating to the player that their parachute failed

commy2:
Lgtm, without having tested this.

One minor gripe I have with this is the function name `handleFailureChance`. `handleX` implies to me that there is some kind of event handler involved:
`handle<Event>` like `handleInfoDisplayChanged` for `ACE_InfoDisplayChanged`.

docs/wiki/framework/parachute-framework.md
```diff
@@ -16,9 +16,21 @@ version:
...
 };
 ```
+
+## 2. Adding failure cut delay
+
+```cpp
+class CfgVehicles {
+    class ParachuteBase;
+    class BananaParachute: ParachuteBase {
+        ace_parachute_failureDelay = 2;  // Add delay before parachute fails (time in seconds)
```
commy2, line 33: Since you're using `getNumber`, the token can also be a simple expression string:

```cpp
        ace_parachute_failureDelay = "1 + random 2";
```
This will be evaluated every time the `getNumber` command is used.

commy2: (because there is no automatic type conversion in Arma :^)

JoramD0: Cool, did not know that was possible. The value however is intentionally set to `2`, it's the point air starts getting under the main parachute (which I would think could cause a rope to fail or something).

PR #8232, opened 2021-04-25
Disarming - Add ace_disarming_dropItems now return the shown right text in the item

Vdauphin:
- The `ace_disarming_dropItems` event now return the shown right text in the item
- Use `lbTextRight` in an array as `[_data]`

### DIFF
--- a/addons/disarming/functions/fnc_openDisarmDialog.sqf
+++ b/addons/disarming/functions/fnc_openDisarmDialog.sqf
@@ -41,7 +41,7 @@ GVAR(disarmTarget) = _target;
     if (isNull GVAR(disarmTarget)) exitWith {ERROR("disarmTarget is null");};
 
     TRACE_2("Debug: Droping %1 from %2",_data,GVAR(disarmTarget));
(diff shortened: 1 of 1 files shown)


addons/disarming/functions/fnc_openDisarmDialog.sqf
```diff
@@ -41,7 +41,7 @@ GVAR(disarmTarget) = _target;
     if (isNull GVAR(disarmTarget)) exitWith {ERROR("disarmTarget is null");};
 
     TRACE_2("Debug: Droping %1 from %2",_data,GVAR(disarmTarget));
-    [QGVAR(dropItems), [ACE_player, GVAR(disarmTarget), [_data]], [GVAR(disarmTarget)]] call CBA_fnc_targetEvent;
+    [QGVAR(dropItems), [ACE_player, GVAR(disarmTarget), [_data], [lbTextRight [_idc, 0]]], [GVAR(disarmTarget)]] call CBA_fnc_targetEvent;
```
Vdauphin, line 44: I never really deal with UI so I am not sure about what the `0` is
BI doc say it is the index https://community.bistudio.com/wiki/lbTextRight but don't see any index in privates variables returned by the `ctrlAddEventHandler`

PR #8239, opened 2021-05-01
Parachute - Add sound effect when cutting

PabstMirror:
ref #8226 
although it applies to both random failure and intentional cutting


bux (approved):

commy2 (approved):

commy2:
No volume and range arguments needed?

mharis001 (approved):

PR #8242, opened 2021-05-04
Medical AI - Use suppressed EH to determine safety of AI

BaerMitUmlaut:
- Use the suppressed EH instead of the current suppression value to determine the safety of an AI unit

This should prevent AI from treating itself or others when being shot at just sporadically.
Not tested yet.


BrettMayson (approved):
Worked great for me in SP

LinkIsGrim (approved):

PR #8264, opened 2021-05-23
Tools - Fix make after hemtt pack

Dystopian:
`build.py` and `hemtt pack` create pbo files in `optionals` dir. `make.py` needs only dirs to copy and stops with error.

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -366,7 +366,7 @@ def copy_important_files(source_dir,destination_dir):
 
 
 def copy_optionals_for_building(mod,pbos):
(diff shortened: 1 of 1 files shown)


PabstMirror (approved):

PabstMirror:
I had come up with a different solution in sqfc pr
```
    for dir_name in src_directories:
            if (os.path.isfile(os.path.join(optionals_root,dir_name))): continue # ignore .pbos from hemmit
```
but I think this is cleaner (`src_directories` is only directories)

PR #8269, opened 2021-05-29
Medical - Hide cursor when unconscious

BaerMitUmlaut:
- Add a custom transparent cursor
- Use this cursor when blocking inputs

The cursor being visible whilst being unconscious always bothered me. The cursor reappears in the escape menu.
The new custom cursor has a single 99% transparent pixel and is otherwise completely transparent. The pixel is necessary, otherwise a black or white square appears as the cursor. No idea why.


veteran29 (approved):

addons/common/functions/fnc_disableUserInput.sqf
```diff
@@ -41,6 +41,10 @@ if (_state) then {
 
         private _display = uiNamespace getVariable QGVAR(dlgDisableMouse);
 
+        // Hide cursor by using custom transparent cursor
+        private _map = (allControls _display)#0;
```
commy2, line 45:
```suggestion
        private _map = allControls _display select 0;
```

addons/common/DisableMouseDialog.hpp
```diff
@@ -0,0 +1,18 @@
+class ctrlMapEmpty;
+class GVAR(DisableMouse_Dialog) {
+    idd = -1;
+    movingEnable = 0;
+    onLoad = QUOTE(uiNamespace setVariable [ARR_2(QUOTE(QGVAR(dlgDisableMouse)),_this select 0)];);
+    objects[] = {};
+    class controlsBackground {
+        // Transparent map allows setting custom cursor
+        class Background: ctrlMapEmpty {
+            idc = -1;
```
PabstMirror, line 10:
```suggestion
            idc = 101;
```

addons/common/functions/fnc_disableUserInput.sqf
PabstMirror, line 45:
```suggestion
        private _map = _display displayCtrl 101;
```

PabstMirror (commented):
normal IDC is probably safest
just in-case someone extends this display the order might change
e.g. adding a "Nearest Medic is 50m away" info box or something like that

addons/common/functions/fnc_disableUserInput.sqf
BaerMitUmlaut: No.

PabstMirror (approved):

addons/common/DisableMouseDialog.hpp
```diff
@@ -0,0 +1,18 @@
+class ctrlMapEmpty;
+class GVAR(DisableMouse_Dialog) {
+    idd = -1;
+    movingEnable = 0;
+    onLoad = QUOTE(uiNamespace setVariable [ARR_2(QUOTE(QGVAR(dlgDisableMouse)),_this select 0)];);
```
Dystopian, line 5:
```suggestion
    onLoad = QUOTE(uiNamespace setVariable [ARR_2(QQGVAR(dlgDisableMouse),_this select 0)];);
```
not tested

Dystopian (changes requested):

addons/common/DisableMouseDialog.hpp
PabstMirror: I really like `with` because you don't have to worry about nested quotes
```suggestion
    onLoad = QUOTE(with uiNameSpace do { GVAR(dlgDisableMouse) = _this select 0; };);
```

BaerMitUmlaut: That's much more readable. Should be improved by using # though.

PabstMirror: `_this # 0` or ` _this#0`

PabstMirror (approved):

PR #8270, opened 2021-05-30
Vehicle Lock - Increase vehicle key mass to 0.1

### LINKED ISSUE #8266: Change ACE_key_master mass to 1 for Antistasi Arsenal Compatibility
Currently, items can only be loaded from the Antistasi arsenal if they have a weight greater than 0.

**Solution you'd like:**
Changing the mass to equal 1 for ACE_key_master items will enable them to be added to loadouts from the Antistasi arsenal.

Modifying the Antistasi arsenal to permit loading items of 0 weight would solve the issue, however the Antistasi mod has multiple forks from multiple maintainers.

BaerMitUmlaut:
- Fix #8266

### DIFF
--- a/addons/vehiclelock/CfgMagazines.hpp
+++ b/addons/vehiclelock/CfgMagazines.hpp
@@ -5,7 +5,7 @@ class CfgMagazines {
         displayName = "ACE Vehicle Key";   //!!!CANNOT be localized!!!: because it is used as part of the magazineDetail string
         descriptionShort = CSTRING(Item_Custom_Description);
         count = 1;
-        mass = 0.1;
+        mass = 0;
         ACE_isUnique = 1;
     };
 };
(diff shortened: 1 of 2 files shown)


jonpas (approved):

addons/vehiclelock/CfgMagazines.hpp
```diff
@@ -5,7 +5,7 @@ class CfgMagazines {
         displayName = "ACE Vehicle Key";   //!!!CANNOT be localized!!!: because it is used as part of the magazineDetail string
         descriptionShort = CSTRING(Item_Custom_Description);
         count = 1;
-        mass = 0;
+        mass = 0.1;
```
PabstMirror, line 8:
```suggestion
        mass = 0;
```
these are scope=1 and are really only going to be added via script so it should be fine to keep at 0
the reason I set them to mass=0 is so the script that adds them could never fail to do `addMagazine`
e.g. player's loadout is full, you can still add a mass=0 mag

commy2: Can't you just use container addMagazineCargoGlobal instead? Never fails, even when full. You need to figure out the container though, but that doesn't sound hard.

BaerMitUmlaut: I increased their weight so they match with the other item. I don't mind undoing it.

PabstMirror (approved):

PR #8273, opened 2021-06-01
Speed limiter utalizes setCruiseControl

ColdEvul:
- Change speed limiter to utalize setCruiseControl

**Adjustments and code of note:**
- Kept setVelocity with a +2 to speed check to make vehicle slow down downhill and to slow it down when enabled.
- Still use the per frame due to ctrl+scroll adjustment check and active breake.

### DIFF
--- a/addons/vehicles/functions/fnc_speedLimiter.sqf
+++ b/addons/vehicles/functions/fnc_speedLimiter.sqf
@@ -21,6 +21,7 @@ params ["_driver", "_vehicle"];
 if (GVAR(isSpeedLimiter)) exitWith {
     [localize LSTRING(Off)] call EFUNC(common,displayTextStructured);
     playSound "ACE_Sound_Click";
+    _vehicle setCruiseControl [0, false];
     GVAR(isSpeedLimiter) = false;
 };
 
@@ -38,21 +39,25 @@ GVAR(speedLimit) = speed _vehicle max 5;
         private _uavControll = UAVControl _vehicle;
         if ((_uavControll select 0) != _driver || _uavControll select 1 != "DRIVER") then {
             GVAR(isSpeedLimiter) = false;
+            _vehicle setCruiseControl [0, false];
         };
     } else {
         if (_driver != driver _vehicle) then {
             GVAR(isSpeedLimiter) = false;
+            _vehicle setCruiseControl [0, false];
         };
     };
 
     if (!GVAR(isSpeedLimiter)) exitWith {
+        _vehicle setCruiseControl [0, false];
         [_idPFH] call CBA_fnc_removePerFrameHandler;
     };
 
+    _vehicle setCruiseControl [GVAR(speedLimit), false];
+
     private _speed = speed _vehicle;
 
-    if (_speed > GVAR(speedLimit)) then {
+    if (_speed > GVAR(speedLimit)+2) then {
         _vehicle setVelocity ((velocity _vehicle) vectorMultiply ((GVAR(speedLimit) / _speed) - 0.00001));  // fix 1.42-hotfix PhysX libraries applying force in previous direction when turning
     };
-
 }, 0, [_driver, _vehicle]] call CBA_fnc_addPerFrameHandler;


Drofseh:
I don't know if you want to do it in this PR, but someway to have autoThrust as true to do actual cruise control would be nice to have at some point

ColdEvul:
> I don't know if you want to do it in this PR, but someway to have autoThrust as true to do actual cruise control would be nice to have at some point

I can experiement with it in a nother pr.

jonpas:
https://forums.bohemia.net/forums/topic/140837-development-branch-changelog/?page=58&tab=comments#comment-3438862

Changes to cruise control from today.

ColdEvul:
> Tweaked: Cruise control now aborts the speed limit when you brake manually
above script reapply it if i am not mistaken. will check effects.
> Tweaked: Cruise control now lets you drive faster than the target speed when you apply manual thrust
should not effect this pr

dedmen:
> Kept setVelocity with a +2 to speed check to make vehicle slow down downhill and to slow it down when enabled.

speed limiter will apply brakes if you are too fast.
setVelocity is unrealistic and will stop you immediately, thats bad

ColdEvul:
The cruise control do not activly slow you down when your going down hill wich is the resoning why i kept the setVelocity as mentioned.
In normal cases you enable the speed limiter with delete key and a sudden slow down in speed whould never apply.

I did not have the setVelocity breaker in the begining. But after testing were i manage go over ~20km/h of the speed limit.
With the setVelocity, to no suprce never reached, over set speed +2.

dedmen:
> The cruise control do not activly slow you down when your going down hill wich is the resoning

It should actively brake though.
I just looked at the code again
if (abs(curSpeed-setpoint) > setpoint*0.2f && speed>setpoint)
brakeWanted = 0.5;

The first check is to limit the PID controller to a smaller range, outside of the range its manually controlled, inside its controlled by the PID.

so if your limit is 50kmh if your speed delta is more than 20% (10kmh) above setpoint it will apply 50% brakes and force thrust to 0.
Then when you come within 20% of setpoint the brakes will be removed and thrust will be controlled by the PID.

I only did 50% brakes to make it a little smoother, maybe thats not sufficient for your downhill test.


My main concern is the "exploit" where you just quickly mouse-wheel your speed down if you need to quickly stop to apply the "magic" brakes.

Maybe the "magic" brake can be disabled if you just reduced speed via mouse wheel in the last 2-5 seconds?

addons/vehicles/functions/fnc_speedLimiter.sqf
```diff
@@ -38,21 +39,25 @@ GVAR(speedLimit) = speed _vehicle max 5;
...
+            _vehicle setCruiseControl [0, false];
         };
     };
 
     if (!GVAR(isSpeedLimiter)) exitWith {
+        _vehicle setCruiseControl [0, false];
         [_idPFH] call CBA_fnc_removePerFrameHandler;
     };
 
+    _vehicle setCruiseControl [GVAR(speedLimit), false];
```
dedmen, line 56: Don't do this!
setCruiseControl script command resets the thrust PID Controller.

If you reset it every frame you'll completely mess up the auto thrust by resetting the PID basically every simulation cycle.

Only do this if you turn on/off or if the speedLimit actually changes.
Engine doesn't check that provided speed is actually different from current setpoint, it just applies what you give it.

dedmen: If you reset the PID every frame, that will likely also cause your mentioned issues with going too fast downhill.

dedmen:
Okey I just tested, because the PID controller only controls thrust and not brakes (while you are within the 10kmh delta) you will go 60kmh downhill when speedlimit is set to 50 (testing on malden mountain)
The PID controller can go negative so I will adjust the code to apply brakes too.

ColdEvul:
> Okey I just tested, because the PID controller only controls thrust and not brakes (while you are within the 10kmh delta) you will go 60kmh downhill when speedlimit is set to 50 (testing on malden mountain)
> The PID controller can go negative so I will adjust the code to apply brakes too.

feel free to adjust

dedmen:
Okey I fixed it.
Looks good to me. of course theres a little over/undershoot but looks good enough.

https://youtu.be/36od_7hFr7

log bottom right

Thrust, Brake, PID output

addons/vehicles/functions/fnc_speedLimiter.sqf
```diff
@@ -38,21 +39,25 @@ GVAR(speedLimit) = speed _vehicle max 5;
...
+
     private _speed = speed _vehicle;
 
-    if (_speed > GVAR(speedLimit)) then {
+    if (_speed > GVAR(speedLimit)+2) then {
         _vehicle setVelocity ((velocity _vehicle) vectorMultiply ((GVAR(speedLimit) / _speed) - 0.00001));  // fix 1.42-hotfix PhysX libraries applying force in previous direction when turning
     };
```
veteran29, line 62: Should not be necessary after fixes to cruise control.
```suggestion
```

addons/vehicles/functions/fnc_speedLimiter.sqf
veteran29, line 56: See @dedmen comment.
```suggestion
    getCruiseControl _vehicle params ["_currentSpeedLimit"];
    if (_currentSpeedLimit != GVAR(speedLimit)) then {
        _vehicle setCruiseControl [GVAR(speedLimit), false];
    };
```

veteran29 (changes requested):

jonpas:
I am confused, are @veteran29's and @Zakant's comments even related? One seems to speak about upper-most value that can be restricted from a perspective of a mission maker, and another speaks of value when you activate.

If I am not mistaken, what @Zakant is again is actually how it works before this PR. You can still increase it by step. Someone explain?

veteran29:
I think Zakant does not understand what I meant. His comment does not make sense to me.

veteran29:
What I meant about respecting mission maker set limit is basically this...

when speed limiter is engaged, outside of PFH:
```sqf
GVAR(maxSpeedLimit) = getCruiseControl _vehicle select 0;
if (GVAR(maxSpeedLimit) == 0) then {
    GVAR(maxSpeedLimit) = 1e10;
};
```

inside the PFH:
```sqf
    getCruiseControl _vehicle params ["_currentSpeedLimit"];
    if (_currentSpeedLimit != GVAR(speedLimit)) then {
        _vehicle setCruiseControl [GVAR(speedLimit) min GVAR(maxSpeedLimit), false];
    };
```

PabstMirror:
@veteran29 Does that last sound ok?

Issue #2818, opened 2015-11-04
Interaction Menu whilst flying


stubaker5002:
Hi

I lose control of the aircraft whilst flying in Arma 3 when I open the self interaction menu. 

For example I've been shot int he Torso entering a hot LZ and I need to bandage myself whilst extracting the LZ and flying at the same time. I launch  the self interaction menu and start bandaging - as soon as this happens all control inputs stop responding; my engine closes throttle / collective and the aircraft will maintain the attitude / pitch.

If I'm badly damaged it means that I have to climb and gain enough altitude to be able to bandage myself whilst plummeting back through the sky and hope that the process of bandaging myself is complete in enough time to avoid crashing into the ground and at the same time hope that I dont expose myself to any air or ground to air threats....

Hope you can help, and thanks for a great ARMA 3 Mod!

Stu

jonpas:
And how do you expect to bandage yourself while flying (holding the aircraft controls)?

stubaker5002:
I realize I'd need two hands to bandage myself - but why does the throttle / collective drop? That control should stay at the level that it's set at? These controls are not 'sprung' in real life, so they would not drop back to zero throttle / collective normally.... Another scenario would be injecting morphine - I can do that with one hand in real life so as long as the throttle is set I'd only need one hand to control pitch etc.....

BaerMitUmlaut:
>I realize I'd need two hands to bandage myself - but why does the throttle / collective drop? That control should stay at the level that it's set at?
That's Armas fault really, that's simply how it does aircraft controls. Did you use the basic or the advanced flight model?

stubaker5002:
I use the basic flight model at the moment. Do you think it might be different with the advanced flight model?

BaerMitUmlaut:
Possibly, yes. The basic flight model doesn't really model collective.
You should get a copilot so he can take over or bandage you if you want to keep playing with the basic flight model.

stubaker5002:
Thanks BaerMitUmlaut. I'll experiment with the advanced flight model.

Pergor:
If you're going at full speed at sufficient altitude, you should be able to bandage yourself without losing neither height nor speed. Or at least it worked for me so far.

dixon13:
Maybe this should be closed then, since it's really an issue with Arma and not really an issue at the same time, since it adds realism to the game?

stubaker5002:
[...]

PR #8275, opened 2021-06-01
Vehicles - Add Auto-Throttle to Planes

TheCandianVendingMachine:
- Add a speed-limiter type control to planes

Uses a basic PID Controller to control the throttle of a plane to keep it at some speed. PID was tuned in Simulink, so values work well. Once you have speed limiter on, you don't have to touch the throttle again.

[Video of it in action](https://youtu.be/cWYr35_laGw)

### DIFF
--- a/addons/vehicles/functions/fnc_autoThrottle.sqf
+++ b/addons/vehicles/functions/fnc_autoThrottle.sqf
@@ -37,7 +37,8 @@ GVAR(speedLimit) = (((velocityModelSpace _vehicle) select 1) * 3.6) max 5;
 [{
     params ["_args", "_idPFH"];
     _args params ["_driver", "_vehicle", "_autothrottleParameters"];
-    _autothrottleParameters params ["_lastVelocity", "_integralValue"];
+    _autothrottleParameters params ["_lastVelocity", "_integralValue", "_lastTime"];
+    private _deltaTime = CBA_missionTime - _lastTime;
 
     if (_driver != driver _vehicle) then {
         GVAR(isSpeedLimiter) = false;
@@ -47,6 +48,8 @@ GVAR(speedLimit) = (((velocityModelSpace _vehicle) select 1) * 3.6) max 5;
         [_idPFH] call CBA_fnc_removePerFrameHandler;
     };
 
+    if (_deltaTime == 0) exitWith {};
+
     private _forwardVelocity = (velocityModelSpace _vehicle) select 1;
     // convert from KM/H to M/S
     private _velocityError = (GVAR(speedLimit) / 3.6) - _forwardVelocity;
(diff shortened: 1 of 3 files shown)


PabstMirror:
Close https://github.com/acemod/ACE3/issues/2818 (I think)

jonpas:
Why only planes? Most cars today have cruise control / tempomat. I assume the PID tune would be different though, so if you don't want to do that, that's fine. :D

jonpas (approved):

PabstMirror:
ground vics are getting in-engine cruise in 2.06
https://community.bistudio.com/wiki/setCruiseControl

jonpas:
Ah right! We have another PR somewhere for that already as well.

TheCandianVendingMachine:
I don't think there is a way to control the throttle in non-plane vehicles, so I'd have to go into the config and `setVelocity` appropriately and it just turns into a mess

jonpas:
[`setCruiseControl`](https://community.bistudio.com/wiki/setCruiseControl) has `autoThrust`.

PabstMirror (approved):

addons/vehicles/functions/fnc_autoThrottle.sqf
```diff
@@ -0,0 +1,99 @@
...
+        _throttle = 0 max ((_p + _d) min 1);
+    };
+
+    _vehicle setAirplaneThrottle _throttle;
+
+    _autothrottleParameters set [0, _d];
+    _autothrottleParameters set [1, _i];
+    _autothrottleParameters set [2, CBA_missionTime];
+    _autothrottleParameters set [3, _throttle];
+    _args set [2, _autothrottleParameters];
```
PabstMirror, line 96:
```suggestion
```

addons/vehicles/functions/fnc_autoThrottle.sqf
```diff
@@ -0,0 +1,99 @@
...
+    if (_driver != driver _vehicle) then {
+        GVAR(isSpeedLimiter) = false;
+    };
+
+    if (_throttleLogValue == 0) then {
+        _throttleLogValue = 1;
+    };
+
+    private _currentThrottle = (airplaneThrottle _vehicle) ^ _throttleLogValue;
+    if (_lastThrottleValue != -1 && EPSILON < abs (_currentThrottle - _lastThrottleValue)) then {
```
PabstMirror, line 55:
```suggestion
    if (_lastThrottleValue != -1 && {EPSILON < abs (_currentThrottle - _lastThrottleValue)}) then {
```

dedmen:
Somewhat unrelated to this PR
> PID was tuned in Simulink

Engine setCruiseControl has hardcoded PID params that might not be optimal
kp = 0.5
ki = 0.3
kd = 0.2
maybe that can be tuned better.
If you can optimize that and find better values, feel free to contact me and i can transfer them into the engine vehicle PID, but it needs to work with all vehicles from quads to tanks

PR #8276, opened 2021-06-02
Aircraft - Drone Tools

PabstMirror:
- Allows quickly changing UAV waypoints from gunner seat
 Modify loiter radius, altitude and direction
 Setting defaults to on because this is all stuff you can do from the terminal, just easier
- Draws laser designator on the map
Close https://github.com/KoffeinFlummi/AGM/pull/1903 😄 
Setting defaults off. Most drones already list the grid cords, this is just QOL

https://www.youtube.com/watch?v=8vF_b2aJ12s


addons/aircraft/functions/fnc_droneAddMapHandler.sqf
```diff
@@ -0,0 +1,39 @@
...
+        if (!alive _laserTarget) exitWith {};
+
+        params ["_map"];
+        _map drawLine [getPos _laserTarget, getPos (ACE_controlledUAV select 0), [1,0,0,0.333]];
+        _map drawIcon [
+        "\A3\ui_f\data\igui\rscingameui\rscoptics\laser_designator_iconLaserOn.paa",
+        [1,0,1,1],
+        (getPos _laserTarget),
+        16, 16,
+        ((getDir _laserTarget) + 90)];
```
Dystopian, line 35:
```suggestion
            "\A3\ui_f\data\igui\rscingameui\rscoptics\laser_designator_iconLaserOn.paa",
            [1,0,1,1],
            (getPos _laserTarget),
            16, 16,
            ((getDir _laserTarget) + 90)
        ];
```

Dystopian (changes requested):
I think there is no reason to disable these tools so settings are not needed.

TheCandianVendingMachine:
Would it be possible to extend this for all aircraft types? I'd like to see a target on the map for helicopters and jets as well

jonpas (approved):

PR #8278, opened 2021-06-08
Medical - Rework wound handling

pterolatypus:
1. Alter how damage types and their relationships to wound types are configured.
2. Allow different damage types to influence wound properties (such as size or bleeding rate) or override wound handling completely
3. Change handling of non-selection-specific damage to be closer to vanilla (makes explosives much more powerful)

*Ok, but why?*
Mostly just because the current system feels weird to me. Apart from that, this should make it simpler to extend the medical system (such as implementing less-lethal munitions) while keeping compatibility.
In addition it's a recurring complaint that explosives are weak; these changes should address that while making it easier to adjust such things in the future.

**1. Alter how damage types and their relationships to wound types are configured.**
Instead of having each wound type define a list of damage types that are allowed to cause that wound, I have each damage type contain a class for each wound type it can cause.
**2. Allow different damage types to influence wound properties (such as size or bleeding rate)**
These classes can then contain multipliers for the incoming damage, "size" of the wound, bleeding, pain and fracture chance allowing these values to be different between different damage types. A bruise caused by a *relatively* weak bullet hit may be more painful than a bruise caused by a fairly large collision.

On top of that, the type of wound created is now given by configurable weighted random which can be adjusted based on the incoming damage.
For example, wound type `bullet` can produce avulsions, contusions and velocity wounds:
- Avulsion has a weight of 1 as long as the damage is above 0.01, 0 otherwise. 
- Contusion is weight 1 if the damage is *below* 0.35, 0 otherwise.
- Velocity wound is weight 1 is the damage is above 0.35, 0 otherwise.

Therefore if the incoming damage is 0.6, the weights are avulsion=1, contusion=0, velocity=1 so it has an equal chance to produce a velocity wound or avulsion, and no chance to produce a contusion.
These weights need not be equal, and can also be interpolated between values: e.g. a wound may produce *mostly* bruises and scrapes when the damage value is low, and *mostly* lacerations when damage is high, but keep a sliding scale of randomness in between.
**...or override wound handling completely**
The wound handler function to use is now read from config and can be overridden for each damage type. This allows other mods to extend wound handling by wrapping the existing function or replace it entirely, or add new damage types with custom handling (such as fire or CBRN threats) without interfering with existing systems.

**3. Change handling of non-selection-specific damage to be closer to vanilla (makes explosives much more powerful)**
In the current implementation, only the hitpoint that took the most damage is actually passed on to the wound system - all other damage is discarded. This damage is then spread *between* all damaged hitpoints somewhat randomly, resulting in very inconsistent incapacitation/lethality (it's common for the head and torso to be untouched) and overall low damage numbers from indirect hits, collisions and falling damage.
Instead I pass all incoming damage to the wound event and have the wound handler make that decision only for "selection-specific" damage, e.g. bullets & shrapnel which should only affect one body part at a time, whereas explosions etc. will do separate damage to each part.

Here are some images to demonstrate the difference.
All show a 155mm shell with a number of `B_Soldier_F` units standing between 30m and 120m from the impact point (3m increments).
<details>
<summary>With no mods</summary>

Instant kill out to about 90m, serious wounds to 110-ish.
</details>
<details>
<summary>With just CBA and current ACE (workshop)</summary>

You can see the inconsistency - units as close as 50m weren't even knocked out, very few instant kills, generally not very effective.
Remember this is a direct vertical impact with targets standing in the open - basically optimal conditions.
</details>
<details>
<summary>With this PR</summary>

Lethal effects out to about 70m, KO out to 110m.
A second image taken a couple of minutes later

Most of the unconscious guys wouldn't have been saved by a medic, and the ones that weren't knocked out tended to pass out from wounds before they were able to bandage.
</details>
Needs a bit of tweaking definitely and personally I think the vanilla damage is already a bit over the top, but it's much more consistent and with the other changes should be easy to adjust.
I should also note that I've set all the thresholds and wound weights to be the same as current release for comparison purposes; these should be re-balanced before merging.

This change does come with a performance cost as the wound handler function is being called 6 times instead of just 1 - in my SP testing this roughly tripled the weight of the whole damage handling routine (from ~150us per handleDamage call to ~450) and didn't produce any noticeable lag (unless you have logging on) but this was in a fairly sterile environment so I'd like to test it in the wild.
I'll be looking at optimisation, but for now I'm most interested in feedback - if you think it's a terrible idea please say so and I can avoid wasting my time :P


BaerMitUmlaut:
>The wound handler function to use is now read from config and can be overridden for each damage type. This allows other mods to extend wound handling by wrapping the existing function or replace it entirely, or add new damage types with custom handling (such as fire or CBRN threats) without interfering with existing systems.

I like this.

>Instead I pass all incoming damage to the wound event and have the wound handler make that decision only for "selection-specific" damage, e.g. bullets & shrapnel which should only affect one body part at a time, whereas explosions etc. will do separate damage to each part.

Why not keep the existing method for selection specific damage, such as bullets or shrapnel, and sum the damage for structural damage? This might also solve the performance concerns?

>I think the vanilla damage is already a bit over the top

Since we can assume that other mods are "balanced" around vanilla damage, I'd prefer if we would get close to vanilla results, at least for explosion damage, treating unrecoverable unconsciousness as death when doing comparisons (as you've already done). I would also like to add concussions sooner or later.

---

Just like the other damage handling PR rewrite, I would have liked to see a discussion before an implementation. And if we touch this now, we should also consider the caliber based damage that RHS seems to use a lot and we don't handle well right now.

commy2:
I was skeptical when I saw that the PR edits HandleDamage, however the way you represent data with those snail images is really cool and changes to how explosions are handled are badly needed. I really like this (without having tried it out).

pterolatypus:
>Why not keep the existing method for selection specific damage

It essentially does, it's just that it's being done within the `woundReceived` EH instead of in `handleDamage`. The potential performance impact is only with damage that hits a lot of hitpoints (like explosions) as the actual handler function is now called for each body part instead of just once.

>sum the damage for structural damage

It seems pointless to add all the damage together only to then split it up again.
I could restructure it a bit so that all of the damage is passed to one call of the handler function - basically move the loop inside the handler function, which might cut the overhead a bit.
Also the thresholds need to be tweaked. Currently where a shell (for example) was creating 5-7 wounds *total* it's instead creating up to 7 wounds *per part hit*; bringing this down to 2-3 per part should help a bit.

>assume that other mods are "balanced" around vanilla damage

That was my logic - use vanilla behaviour as the baseline and then tweak around it.

>I would also like to add concussions

My future plans include a custom handler for explosives that would handle blast and fragmentation effects separately (the latter could integrate with ace fragmentation somehow) and a kind of "stun" quantity that would decouple unconsciousness from lethal damage somewhat. 

>we should also consider the caliber based damage that RHS seems to use

I'm not familiar with that, but it sounds interesting.
Another thing I wanted to write off the back of this was better handling of body armour, since the current system doesn't line up with vanilla for direct hits because of ignoring passthrough, and I'd like to use caliber to model penetration for that.

>I was skeptical when I saw that the PR edits HandleDamage

You're gonna love my latest commit then XD. But yeah, part of the motivation for this was being able to clean up HD a bit and move some of the logic (like handling for burning & fall damage) out of medical_engine so that we can tweak numbers and balance without having to be reminded of that horror.

>Any chance the script used to make these could be put in ACE

At first I tried just dropping a bomb on a long line of guys, but I discovered that they do in fact shield each other from the blast which I never knew before. I think I stole the spiral idea from someone else but I can't remember who.
It's currently just [this mission file](https://drive.google.com/file/d/1nvzNCmPuNZiwG6_UPsoa8nCk1RxLTczF/view) but yeah, I don't see why not. If I made it spawn the ammo too it could be used for grenades or other things.

kymckay:
@pterolatypus I'd just first like to say a massive thanks for your continued work on reworking medical, it's clear to me you've taken the time to understand the systems as they are and the intentions behind them in order to see where to improve.

I haven't tested this PR (it's been quite some time since I truly booted up Arma), but I like all of your reasoning, justifications and the results presented.

>Instead of having each wound type define a list of damage types that are allowed to cause that wound, I have each damage type contain a class for each wound type it can cause.

Honestly I think this this makes more sense semantically and in terms of coupling too (the cause specifies the effect).

> These classes can then contain multipliers for the incoming damage, "size" of the wound, bleeding, pain and fracture chance allowing these values to be different between different damage types. A bruise caused by a relatively weak bullet hit may be more painful than a bruise caused by a fairly large collision.
>
> On top of that, the type of wound created is now given by configurable weighted random which can be adjusted based on the incoming damage.

Very nice, more logically driven variance.

>Instead I pass all incoming damage to the wound event and have the wound handler make that decision only for "selection-specific" damage

Again, very logical. I've suspected for a long while that our filtering down to the maximum damage hitpoint is what's been causing some of the weirdness with the more general (as opposed to regional/local) damage types. Glad to see it being addressed.

----------

I do agree with @commy2 it would've been nice to see some discussion before implementation, but I also understand discussion isn't as active as it once was.

If we get to the stage where we're going to merge this. I think we should be proactive and get documentation of the system (config changes for addon authors, but also some developer documentation on how the system now works) before merging. Something we've not been strict about in the past, leading to loss of established design intent/system knowledge and having to reverse engineer it from the source.

pterolatypus:
Thanks for all the positive feedback!

I've made a change to how burning is handled - now instead of waiting for a certain amount of damage before creating wounds, it waits a certain amount of time (currently 1s which works reasonably). It's also a good example of a custom damage handler in action.
The result of this is that the number of wounds created is proportional to time spent in contact with the fire and the size of the wounds is determined by the intensity, which I think makes sense. I'd like a better wound type to give (instead of cuts and scrapes) but I think that's out of scope for this PR.

The system seems to be basically functional now, so I've started cleaning up and hunting the less-obvious bugs. I'll also implement the testing function and start working on documentation (boooo). If anyone has time to test it out and report any errors or unexpected behaviour that'd be appreciated, as I tend to miss stuff. Should be ready for a proper review soon.

pterolatypus:
Testing function and first attempt at documentation are in - I've created a new framework page for this as it's information for developers rather than end-users. I could also add sections for some of the stuff in medical_treatment, but that might be a job for another PR.

I think we're review-ready, at least on a systems-level.

One thing I do want to change is the thresholds - as mentioned, because explosives now typically hit all body parts this leads to 5-6x as many wounds being created which is bad for both performance and balance reasons (makes them waaay harder to treat). This is kind of a balance question so needs discussion.

xfunnypigx:
> One thing I do want to change is the thresholds - as mentioned, because explosives now typically hit all body parts this leads to 5-6x as many wounds being created which is bad for both performance and balance reasons (makes them waaay harder to treat). This needs some discussion though. An idea I had was something like 1-2 wounds at high damage values and 0-1 at lower values, which would create some satisfying randomness rather than everyone in a 150m radius being covered in bruises.

I believe that it is fair to have damage all over the body. However, I do think that the way it could be balanced is that higher surface area body parts (read: chest) and potentially the head could see higher chances for higher damage while limbs could see lower chances for that, as well as generally smaller wounds. This would make explosions still extremely dangerous and costly to deal with, while not just outright applying 5 wounds on every body part and thus virtually killing a soldier by themselves.
In other words: Heavier wounds for the chest and head, lighter and fewer wounds for the limbs.

kymckay:
Just read closely through the documentation and content-wise it lgtm 👍 Also makes reviewing the design intent of this PR easier for those less familiar with these systems which is nice. Everything described still sounds like an improvement to me

Nice job and thanks for doing that

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -5,188 +5,173 @@
...
 
+        _bodyPartDamage set [_bodyPartNToAdd, (_bodyPartDamage select _bodyPartNToAdd) + _woundDamage];
+        _bodyPartVisParams set [[1,2,3,3,4,4] select _bodyPartNToAdd, true]; // Mark the body part index needs updating
 
-            // Damage to limbs/head is scaled higher than torso by engine
-            // Anything above this value is guaranteed worst wound possible
-            private _worstDamage = [2, 1, 4, 4, 4, 4] select _bodyPartNToAdd;
+        // Damage to limbs/head is scaled higher than torso by engine
+        // Anything above this value is guaranteed worst wound possible
+        private _worstDamage = [2, 1, 4, 4, 4, 4] select _bodyPartNToAdd;
```
pterolatypus, line 88: This is weird and I don't like it. First off the comment is wrong - most things tend to do more damage to the torso because:

- It's big and has low penetrability, so projectiles almost never over-penetrate
- It multiplies explosive damage by 6 (presumably to simulate the fact that it's big)

I wonder if whoever tested this did so with armored targets and scaled the values according to that, as it seems to line up with the standard plate carrier, but I don't think it should be balanced like that.

Secondly, because this only affects the wound size and not the underlying damage number it interacts weirdly with thresholds - damage to a limb creates the same number of wounds as if had hit the torso but they are 1/4 the size, so it always creates a lot of small wounds instead of one or two large ones.

Personally I don't think this part scaling is necessary at all, but if it is it should probably be a multiplier to damage before the number of wounds is calculated.

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -5,188 +5,173 @@
...
 
-            // More wounds means more likely to get nasty wound
-            private _countModifier = 1 + random(_i - 1);
+        // Config specifies bleeding and pain for worst possible wound
+        // Worse wound correlates to higher damage, damage is not capped at 1
+        private _woundSize = linearConversion [0.1, _worstDamage, _woundDamage * _sizeMultiplier, 0.25, 1, true];
```
pterolatypus, line 92: I don't feel great about the minimum size value being 0.25. This makes lower-damage wounds proportionally more dangerous (in the worst case, 1/40 the damage results in 1/4 the bleeding) which has a nasty interaction with the thing I mentioned above. Combined with the fact that creating multiple wounds is already punishing because it usually means more work to bandage, this makes controlling the lethality of explosives really complicated and kind of arbitrary.

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
kymckay: @pterolatypus This was added by myself back during the rewrite process, at the time I knew it was a hacky imperfect way to try and resolve the inconsistent wound damage scaling we were seeing (because we didn't understand what was causing it, the armour + over-penetration systems I'm now pretty confident were what we were observing)

The reason it affects wound size was to scale the bleeding rate accordingly since that's the main cause of death from wounding that isn't fatal itself. I have wanted to understand and remove this hacky "fix" ever since first adding it so please be my guest, I think the changes here are a good motivation to do so

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
```diff
@@ -5,188 +5,172 @@
...
-            // Worse wound correlates to higher damage, damage is not capped at 1
-            private _bleedModifier = linearConversion [0.1, _worstDamage, _woundDamage * _countModifier, 0.25, 1, true];
-            private _painModifier = (_bleedModifier * random [0.7, 1, 1.3]) min 1; // Pain isn't directly scaled to bleeding
+        private _bleeding = _woundSize * _bleedMultiplier * _injuryBleedingRate;
 
-            private _bleeding = _injuryBleedingRate * _bleedModifier;
-            private _pain = _injuryPain * _painModifier;
-            _painLevel = _painLevel + _pain;
+        // wound category (minor [0.25-0.5], medium [0.5-0.75], large [0.75+])
+        private _category = floor linearConversion [0.25, 0.75, _woundSize, 0, 2, true];
```
pterolatypus, line 99: On the same topic I think these thresholds could be improved. As it is, the smallest "large" wound `0.75` is only 50% bigger than the largest "small" wound `0.499` which makes these terms less useful for gameplay.
I'd like to change it to an exponential scale like:
- minor: 0.125 - 0.25
- medium: 0.25 - 0.5
- large: 0.5-1

Which results in medium wounds being about twice as bad as minors, and large being about twice as bad as medium. The exact scale factor depends on what the minimum value is, something like `0.216-0.36-0.6-1` would work as well. This would obviously result in more wounds being categorised as large, but I think this would be a more honest representation of how dangerous those wounds are.

addons/medical_damage/ACE_Medical_Injuries.hpp
```diff
@@ -7,125 +7,275 @@ class ACE_Medical_Injuries {
...
+        // if 1, wounds are only applied to the hitpoint that took the most damage. othewrise, wounds are applied to all damaged hitpoints
         selectionSpecific = 1;
+        
+        // this can also be overridden for each damage type
+        woundsHandler = QFUNC(woundsHandlerActive);
 
         class bullet {
-            // above damage, amount. Put the highest threshold to the left and lower the threshold with the elements to the right of it.
-            thresholds[] = {{0.1, 1}};
+            thresholds[] = {{0.1, 1}, {0.1, 0}};
```
pterolatypus, line 70: This has the effect that bullet wounds are kind of capped - even if it does 17.5 damage it'll only ever produce one wound (although it'll definitely kill you).
I had an idea to allow bullets to create multiple wounds under certain circumstances:
- only if damage>2
- divide in such a way that each wound is always >1
- write the wound weighting so that these wounds can only be avulsions

This way very large hits (like getting your arm blown off by a 50) would create a proportional _number_ of large avulsions, allowing bleeding and pain to scale up higher and making limb shots with these types of weapons more dramatic (since they can't be fatal on their own).

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
pterolatypus: For now I've replaced this with a flat 2 which I think it reasonable. I've also made a change to the hitpoint macro that affects limb damage.

addons/medical_damage/functions/fnc_woundsHandlerSQF.sqf
pterolatypus: Minimum size value is now determined by the thresholds

addons/medical_damage/ACE_Medical_Injuries.hpp
pterolatypus: I've implemented this with higher numbers (it can always be reverted):
a bullet hit with `damage>3` may create multiple wounds, they will always be over 1.5 and wounds over that threshold always result in large avulsions. With the limb armor change, however, this virtually never happens.

addons/medical_engine/script_macros_config.hpp
```diff
@@ -24,27 +26,21 @@
...
-        armor = leg_armor;\
+    class HitLeftLeg: HitLegs {\
         material = -1;\
         name = "leg_l";\
-        passThrough = 1;\
         radius = 0.1;\
-        explosionShielding = 1;\
         visual = "injury_legs";\
         minimalHit = 0.01;\
     };\
```
pterolatypus, line 46: Previously the values used here were set manually and often didn't match up to the vanilla hitpoints. In the interest of vanilla-consistency I think it makes more sense to just inherit them.

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -38,13 +38,37 @@ private _newDamage = _damage - _oldDamage;
...
+    {_ammo isEqualTo ""} &&
+    {_vehicle != _unit} &&
+    {vectorMagnitude (velocity _vehicle) > 5}
+    // todo: no way to detect if stationary and another vehicle hits you
+) exitWith {
+    TRACE_5("Crash",_unit,_shooter,_instigator,_damage,_newDamage);
+    [QEGVAR(medical,woundReceived), [_unit, [[_newDamage, _hitPoint, _newDamage]], _unit, "vehiclecrash"]] call CBA_fnc_localEvent;
+
+    0
+};
```
pterolatypus, line 71: These "special" damage sources are now evaluated first so their damage doesn't get saved to the unit - previously it was being saved and not cleared so it would be included in the next event that triggered `ace_hdbracket`. It didn't cause any problems because it's only `#structural` damage which isn't used anywhere, but could have in the future.

Saving of damage is now done last, so it gets skipped any time wounds are actually applied.

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -38,13 +38,64 @@ private _newDamage = _damage - _oldDamage;
...
+    {
+        private _armor = [_unit, _x] call FUNC(getHitpointArmor);
+        private _shielding = GET_NUMBER(configFile >> "CfgVehicles" >> _uniformClass >> "Hitpoints" >> _x >> "explosionShielding", 1);
+        _damages pushBack [_hit*_shielding/_armor, ALL_BODY_PARTS select _forEachIndex, _hit*_shielding]
+    } forEach ALL_HITPOINTS;
+    TRACE_6("Vehicle explosion",_unit,_shooter,_instigator,_damage,_newDamage,_damages);
+    [QEGVAR(medical,woundReceived), [_unit, _damages, _unit, _ammo]] call CBA_fnc_localEvent;
+
+    0
+};
```
pterolatypus, line 98: I've added this to manually calculate damage for people inside exploding vehicles, because I had a flash of inspiration while looking at the code. Might be out of scope for this PR so I can move it if desired.

addons/medical/functions/fnc_addDamageToUnit.sqf
```diff
@@ -22,8 +21,8 @@
  */
 // #define DEBUG_TESTRESULTS
 
-params [["_unit", objNull, [objNull]], ["_damageToAdd", -1, [0]], ["_bodyPart", "", [""]], ["_typeOfDamage", "", [""]], ["_instigator", objNull, [objNull]], ["_damageSelectionArray", [], [[]]]];
-TRACE_6("addDamageToUnit",_unit,_damageToAdd,_bodyPart,_typeOfDamage,_instigator,_damageSelectionArray);
+params [["_unit", objNull, [objNull]], ["_damageToAdd", -1, [0]], ["_bodyPart", "", [""]], ["_typeOfDamage", "", [""]], ["_instigator", objNull, [objNull]]];
+TRACE_56("addDamageToUnit",_unit,_damageToAdd,_bodyPart,_typeOfDamage,_instigator);
```
veteran29:
```suggestion
TRACE_5("addDamageToUnit",_unit,_damageToAdd,_bodyPart,_typeOfDamage,_instigator);
```

addons/medical_engine/functions/fnc_handleDamage.sqf
kymckay: I'm assuming it's dependent on the other functional changes in the PR? If so I think it's fine here as it's not a huge addition to review at the same time, I'd just avoid any further scope creep.

If it's not I'd prefer it in a separate PR.

addons/medical_engine/script_macros_config.hpp
kymckay: Good idea to include a comment above the macro to explain their absence just in case anyone thinks of adding them again in future (also for reviewers who may not see this)

addons/medical_engine/functions/fnc_handleDamage.sqf
pterolatypus: The format will be affected by these changes, but no I'll remove it and revisit after this pr is sorted

pterolatypus:
Bump. I don't think anything else needs adding to this (I have more ideas but they're best left to other PRs) so I mostly need more input and reviews.

addons/medical_engine/script_macros_config.hpp
```diff
@@ -7,44 +7,40 @@
...
     };
 */
 
 // Our method for adding left and right arm and leg armor. Uses those selections
-// that are used for animations and therefore exist in all third party units
-// usage: arm_armor and leg_armor are the armor values of of HitHands and
-// HitLegs respectively.
+// that are used for animations and therefore exist in all third party units.
+// This used to take the armor values as parameters; it now inherits them from
+// the existing hitpoints instead.
```
kymckay, line 21: I think this would benefit from being more explicit to avoid regression:

```suggestion
// that are used for animations and therefore exist in all third party units.
// This used to take the armor values as parameters; it now inherits the values
// of `armor`, `passThrough` and `explosionShielding` from the existing hitpoints
// for vanilla consistency.
```

kymckay:
I think a PR of this scale would benefit from some community playtesting alongside individual testing. @jonpas You're pretty up to date, do we still have communities that would be interested in playtesting system changes like this?

kymckay:
Did some individual testing, observable results are very promising.

I did find one bug (which I've since been unable to reproduce), I was shooting blufor units with the main cannon of a `"B_T_MBT_01_TUSK_F"` and kept getting the following script error:

```
21:09:44 Error in expression < params ["_damage", "_bodyPart"];


if (toLower _bodyPart isEqualTo "#structural>
21:09:44   Error position: <toLower _bodyPart isEqualTo "#structural>
21:09:44   Error tolower: Type Number, expected String
21:09:44 File z\ace\addons\medical_damage\functions\fnc_woundsHandlerSQF.sqf..., line 46
21:10:00 Error in expression < params ["_damage", "_bodyPart"];


if (toLower _bodyPart isEqualTo "#structural>
21:10:00   Error position: <toLower _bodyPart isEqualTo "#structural>
21:10:00   Error tolower: Type Number, expected String
21:10:00 File z\ace\addons\medical_damage\functions\fnc_woundsHandlerSQF.sqf..., line 46
```

I suspect this was something to do with the specific unit I was shooting at

kymckay (dismissed):
See comment above, script error under certain conditions results in no damage to unit

Salluci:
Script error is caused by non selection specific damage being applied. Array that gets passed from fnc_woundReceived to fnc_woundsHandlerSQF only has two elements instead of three.

addons/medical/functions/fnc_addDamageToUnit.sqf
```diff
@@ -49,7 +44,7 @@ private _startDmg = +(_unit getVariable [QEGVAR(medical,bodyPartDamage), [0,0,0,
 private _startPain = GET_PAIN(_unit);
 #endif
 
-[QEGVAR(medical,woundReceived), [_unit, _bodyPart, _damageToAdd, _instigator, _typeOfDamage, _damageSelectionArray]] call CBA_fnc_localEvent;
+[QEGVAR(medical,woundReceived), [_unit, [[_bodyPart, _damageToAdd]], _instigator, _typeOfDamage]] call CBA_fnc_localEvent;
```
Salluci, line 47:
```suggestion
[QEGVAR(medical,woundReceived), [_unit, [[_damageToAdd, _bodyPart, _damageToAdd]], _instigator, _typeOfDamage]] call CBA_fnc_localEvent;
```
Simplest fix.

pterolatypus: Good catch, thanks. I changed the format of woundReceived to also take the damage-before-armor value and apparently forgot to update it everywhere.

kymckay:
Previous issue resolved 👍 

Some more individual testing notes:

- I think the distribution of wound types and sizing is significantly improved (both in terms of variance and appropriateness)
- Some unscientific test results, tested on civilians with a single torso shot from close distance:
  - M2 HMG .50 - 6 wounds, all 2+ damage, all 0.1 bleed, all 1.0 pain
  - P99 handgun (9mm) - 1 wound, 2.53 damage, 0.1 bleed, 1.0 pain
  - MX rifle (6.5mm) - 2 wounds, 2.26 and 2.33 damage, both 0.1 bleed, both 1.0 pain
  - MXM rifle (6.5mm) - 2 wounds, 2.25 and 2.35 damage, both 0.1 bleed, both 1.0 pain
  - CTAR-21 rifle (5.56mm) - 1 wound, 4.26 damage, 0.1 bleed, 1.0 pain
  - M2000 Intervention rifle (.408) - 6 wounds, all ~2 damage, all ~0.1 bleed, all ~1.0 pain
  - 40mm GPR-T round from Mk44 Bushmaster II cannon - 6 wounds (five to head), all less than 1 damage, all 0.03-0.05 bleed, all 0.03-0.04 pain
    - A second test of this shot had similar results, but 3 wounds to each arm and none to head
  - 7.62mm round from MAG 58M machine gun - 3 wounds, all ~2 damage, all ~0.1 bleed, all ~1.0 pain

kymckay:
I notice that standing in a campfire seems to result in no damage with the change to handling here (or perhaps it just takes too long to accumulate?)

pterolatypus:
Standing in fire should cause damage, there's just no wound type for it currently so it gives cuts and scrapes instead.
It might be that it doesn't do enough damage in the interval to actually create a wound, I'll look into it.

pterolatypus:
I tracked down the issue: it's because the example I used to test (a burning vehicle) did reasonable damage to individual hitpoints so I used that for the new burning handling. I didn't realise that some sources (e.g. campfires) only hit the structural `""` hitpoint.

I've adjusted the handling now to only use structural damage and randomly assign it to the torso and legs, which is closer to the old method. It still collects it into wounds over time - every second currently.
Another problem is that fire just does a really small amount of damage, like 0.05/s, so it was also not hitting the threshold to create wounds. I've reduced this threshold now but I think in the long term some balance changes are needed.

kymckay (approved):
A shame about the selection specific burn damage limitation.

I'm happy to approve this PR and leave it to other maintainers to decide how in-depth we want to test it. I don't foresee any significant risk to including this in the next release.

The results, code logic and motivations all pass my inspection. I think this can be merged as is and it will facilitate further incremental improvement (such as the true vehicle explosion handling that will follow).

My main concern was whether this could have any negative impact on the end user gameplay - which, after testing, I don't think it can even if we decide some of the wound values need tweaking.

addons/medical_engine/script_macros_config.hpp
kymckay: @pterolatypus Does this rewording correctly reflect the reality and seem reasonable to you?

addons/medical_damage/ACE_Medical_Injuries.hpp
```diff
@@ -7,136 +7,299 @@ class ACE_Medical_Injuries {
...
+            //No related wounds as drowning should not cause wounds/bleeding. Can be extended for internal injuries if they are added.
             thresholds[] = {{0, 0}};
+            woundsHandler = "{}";
+        };
+        class fire {
+            // custom handling for environmental fire sources
+            // passes damage to "burn" so doesn't need its own wound stats
+            woundsHandler = QFUNC(woundsHandlerBurning);
+        };
+        class burn {
```
severgun, line 285: class burn defined twice. 
`make.py release` failed to build.

jonpas:
Needs some conflict fixing. 6th argument was added in #8599, but this one removes 5th, not sure how to assure backwards compatibility as this is public API.

addons/medical/functions/fnc_addDamageToUnit.sqf
```diff
@@ -29,7 +29,7 @@ params [
     ["_bodyPart", "", [""]],
     ["_typeOfDamage", "", [""]],
     ["_instigator", objNull, [objNull]],
-    ["_damageSelectionArray", [], [[]]],
+    "_",
```
pterolatypus, line 32: This isn't a neat solution to the parameter issue, but so far I can't think of anything better.

veteran29: You can use empty string to denote unused parameter.

pterolatypus:
Had another idea about this; stackable wound handlers.
Something like:
```
class bullet {
    class woundsHandler {
        my_addon = "my_fnc_woundsHandler";
        etc...
    };
};
```
Allow declaring multiple handlers, then probably call them in reverse order and allow them to block further handling by returning a value.
It adds complexity, but would make it easier to extend or selectively override wound handling without having to re-implement a load of default behaviour.

addons/medical_engine/script_macros_config.hpp
rautamiekka: It's impressive that you're willing to write a multiline explanation into the code using single-line comments, due to how laborious it is, I'll give you that, but I still think it's better to use the `/* ... */` instead XD

Unless it's a policy I missed.

severgun:
This is old code but...
1) Why this exist here and used once while can be used twice? Maybe eliminate it and just use QGVAR?
https://github.com/acemod/ACE3/blob/da82e250ebf947d7fba0461a50958b5ff4dcb904/addons/medical_engine/functions/fnc_handleDamage.sqf#L17

2) Maybe it should be `[0,0]`? 
https://github.com/acemod/ACE3/blob/da82e250ebf947d7fba0461a50958b5ff4dcb904/addons/medical_engine/functions/fnc_handleDamage.sqf#L79
Otherwise it will fail at `_damageStructural + [PRIORITY_STRUCTURAL, "#structural"]` with SCALAR + ARRAY

severgun:
Bullets with high `hit` values generate alot of structural damage when hit limbs. So, because of sorting,
This:
https://github.com/acemod/ACE3/blob/61d417418a796cf3523301734dfca02d769a80bb/addons/medical_damage/functions/fnc_woundReceived.sqf#L26-L28
And this:
https://github.com/acemod/ACE3/blob/61d417418a796cf3523301734dfca02d769a80bb/addons/medical_damage/functions/fnc_woundsHandlerBurning.sqf#L26-L28
Ends up in 0 damage

STEPS TO REPRODUCE
* Place 12.7 HMG.
* Place bots.
* Shoot full mag in legs or arms.
* 0 damage.

Maybe something like that would be ok?
```sqf
if (_selectionSpecific > 0) then {
    private _firstBodyPart = toLower (_allDamages select 0 select 1);
    if (_firstBodyPart isEqualTo "#structural") then {
        _allDamages = [_allDamages select 1];
    } else {
        _allDamages = [_allDamages select 0];
    };
};
```

addons/medical_engine/XEH_postInit.sqf
```diff
@@ -59,7 +59,7 @@
     private _lethality = linearConversion [0, 25, (vectorMagnitude velocity _vehicle), 0.5, 1];
     TRACE_2("air crash",_lethality,crew _vehicle);
     {
-        [QEGVAR(medical,woundReceived), [_x, "Head", _lethality, _killer, "#vehiclecrash", [HITPOINT_INDEX_HEAD,1]], _x] call CBA_fnc_targetEvent;
+        [QEGVAR(medical,woundReceived), [_x, [["Head", _lethality]], _killer, "#vehiclecrash"], _x] call CBA_fnc_targetEvent;
```
severgun, line 62: Our community member also found bug.
Wrong damage array format.
```suggestion
        [QEGVAR(medical,woundReceived), [_x, [[_lethality, "Head", _lethality]], _killer, "#vehiclecrash"], _x] call CBA_fnc_targetEvent;
```

addons/medical_damage/XEH_preInit.sqf
```diff
@@ -25,11 +17,14 @@ addMissionEventHandler ["Loaded",{
...
+addMissionEventHandler ["Loaded",{
+    INFO("Mission Loaded - Reloading medical configs for extension");
+    // Reload configs into extension (handle full game restart)
+    call FUNC(parseConfigForInjuries);
+}];
 
-    private _typeOfDamage = _ammo call FUNC(getTypeOfDamage);
-    [_unit, _woundedHitPoint, _receivedDamage, _typeOfDamage, _damageSelectionArray] call FUNC(woundsHandlerActive);
-}] call CBA_fnc_addEventHandler;
+[QEGVAR(medical,woundReceived), FUNC(woundReceived)] call CBA_fnc_addEventHandler;
```
Salluci, line 28:
```suggestion
[QEGVAR(medical,woundReceived), LINKFUNC(woundReceived)] call CBA_fnc_addEventHandler;
```

addons/medical_damage/functions/fnc_woundReceived.sqf
```diff
@@ -21,11 +21,7 @@ params ["_unit", "_allDamages", "_shooter", "_ammo"];
 
 private _typeOfDamage = _ammo call FUNC(getTypeOfDamage);
 if (_typeOfDamage in GVAR(damageTypeDetails)) then {
-    (GVAR(damageTypeDetails) get _typeOfDamage) params ["", "_selectionSpecific", "_woundHandlers", ""];
-
-    if (_selectionSpecific > 0) then {
-        _allDamages = [_allDamages select 0];
-    };
+    (GVAR(damageTypeDetails) get _typeOfDamage) params ["", "", "_woundHandlers", ""];
```
rautamiekka, line 24: Shouldn't need to specify the last elements as unused:
```suggestion
    (GVAR(damageTypeDetails) get _typeOfDamage) params ["", "", "_woundHandlers"];
```

Salluci:
Some thoughts from using this internally in my community:

- While the creation of multiple/larger wounds from high damage events being more frequent is much appreciated and remniscent of pre-rewrite medical, it could probably use some toning down (for gameplay purposes if nothing else). Perhaps reintroducing the separation of the `_worstDamage` value for limbs/critical areas in woundsHandlerSQF?

- Even with the latest commit, small-arms fire with high `hit` values causing high structural damage is still an issue. Example for today: a headshot with `B_127x108_APDS` fired from `srifle_GM6_F` on an FIA rifleman ended up creating wounds in the unit's arms. Could this possibly be addressed by further increasing Head and Body priorities?

- For regular use cases (non-specialty weapons in infantry combat), lethality in general is increased, even against problematic factions like RHS AFRF, which is appreciated. High explosives, artillery, and grenades are also much enjoyable to use, and closer to the vanilla game's balancing for them, in a good way.

kymckay:
Posted elsewhere, but I think if we can address the two feedback points (structural damage sounds most important) then I think we should aim to get this merged in and deal with any unseen consequences after the fact.

In an ideal world this would have taken place over multiple smaller PRs making it easier to manage, but I think enough eyes have crossed this to say the system itself is functionally sound. Plus the sooner the better as this really lays a foundation for further damage handling improvements thanks to the improved semantics and more consistent behaviour.

It's also large enough that I'd rather get it in before it becomes impossible to merge without reworking or that it blocks other PRs.

pterolatypus:
>small-arms fire with high hit values

This isn't directly related to structural damage, it's just the same old splash damage issue. Your solution effectively flips the sorting which could possibly cause the opposite problem.
Instead I've just reordered the array to actually use priority (it was mostly being ignored before) which seems to improve the situation, but the underlying problem is still the fact that these vests are balanced around `passthrough` which we don't understand or have a good way of accounting for.

>Limbs typically being unarmored

With this PR, unarmored limbs should actually take less damage due to the changes to the hitpoint macro - their base armor has gone from 1 to 3 which should more than offset the `_worstDamage` change.

>lethality in general is increased

>some "interesting" interactions

You're gonna have to be more specific. For small arms there shouldn't be any immediate difference because the underlying damage (i.e. instant-kill potential) is unchanged and bleed rates should be pretty similar, except for events that now create multiple wounds. Is that happening more often than it should?
Personally I find the bleeding rate in general to be excessive with default settings, so we could consider lowering that baseline.

severgun:
> Personally I find the bleeding rate in general to be excessive with default settings, so we could consider lowering that baseline.

I did not test PR with lastest commits. But first expression month ago was:
Explosives, UGLs, grenades - wow I like it.
Disabled/destructed vehicles with AT - mehh, same as master. Units jumpout without any damage.
Bullets - cause a lot of contusions and small abrasion where on master branch that would be avulsion and velocity. So it feels like it is harder to kill. After that we find out issue with limbs. Maybe now it is better. 
I also ran out of bandages more often. Maybe wounds chances should be tuned. (This can be done after merge)

kymckay:
From what I can tell on quickly inspecting the sum of trauma setting, the bruises must already be contributing to that before this PR. Looking at the changes here again, it seems this behaviour is just more apparent because bruises can now cause more damage.

Drofseh:
I don't think bruising contributing to sum of trauma is too unreasonable. It could account for things we don't stimulate like broken ribs.

kymckay:
After internal discussion, I plan to merge this PR on Thursday (to give time for anyone to object). So if you have a strong reason this shouldn't be merged, say it now :grin: :tada: 

To be clear, we think some balancing is needed, but are happy for that to take place after merge since this PR is more about the functionality. The bruising thing does make some sense, though will need to be considered in balance with treatment of bruising.

pterolatypus:
Glad to see that the explosives change is appreciated, I was concerned people would find it excessive. I do still think the damage could be reduced a little, but that can wait.

The thing about bruises/sum of trauma is the same on master, you might just not notice it as much because those small hits can also be avulsions whereas I have changed the minimum damage for avulsions to match velocity wounds (`0.35`) as this is described as the threshold for "penetrating" damage.
It's definitely worth looking into in relation to when trauma should be cleared, but not directly related to this PR. Side note; those small wounds _are_ ignored when checking incapacitation (KO) which should also probably be discussed, doesn't make a sense to me.

I think I need to see it in the wild myself to get a feel for the balance, since the feedback thus far is a bit inconsistent (some of it could be placebo effect).
If nothing seems to be broken then yeah let's get it merged and you can all yell at me if there are problems :P

PR #8282, opened 2021-06-17
Common - Fix gesture anim lockout for some launchers

PabstMirror:
Fix lost ammo bug on some RHS launcher.
```
class rhs_weap_rpg7 { reloadAction = "RHS_GestureReloadRPG7";
class CfgGesturesMale { class RHS_GestureReloadRPG7
```

- This changes the code to search both locations for launchers
(should it search both for all weapons? I think it's valid to put them either place?)


PabstMirror:
@Cyruz143 mentioned using new 2.06 commands `https://community.bistudio.com/wiki/weaponReloadingTime`
ie: watch and wait for reload to finish before allowing gestures

This might be safer as there are other things that can change reload time like custom anim coef
or https://community.bistudio.com/wiki/setWeaponReloadingTime
(I remember some "muzzle-loading musket" joke-mission used this to simulate long reload times)

but I'd like to just merge this for now as those require 2.06

jonpas (approved):

PR #5233, opened 2017-06-01
Fix IR detach after time interval

Phyma:
**When merged this pull request will:**
- Will remove the x_IR_Grenade from person or vehicle when it is about to die.
- Works both for players and vehicles.
- Fixes #4946

However: Do we want it to get removed when it is about to die or just refreshed?
I think it would be one of the charms that the grenade dies after 5 minutes. While the strobe lives longer?

PR #8283, opened 2021-06-18
Attach - Improve IR Strobe blinking pattern and increase lifetime

### LINKED ISSUE #4946: IR Strobe detaches after time interval

**Mods:**
- `@CBA_A3`
- `@ace`

**Description:**
The time seems to be around 10 minutes.
Bux said it might be the lifeTime of the strobe. But it still continues blinking after it detaches.
I sure sounds to me like the batteries of the strobe ran out and it should stop blinking. But it continues blinking for atleast 5 minutes.

After it detaches 
`player getVariable ["ace_attach_attached",true]`
returns `[[<NULL-object>,"B_IR_Grenade"]]`

**Steps to reproduce:**
- Attach vanilla Nato IR strobe
- Go afk for 10-20 minutes
- Move a few steps away and the IR strobe will be floating in the air behind you

- Dedicated

mharis001:
- Change IR Strobe blinking pattern to 0.1 s on, 0.9 s off
- Increase lifetime to 8 hours
- Close #4946 
- Replace #5233

### DIFF
--- a/addons/attach/CfgVehicles.hpp
+++ b/addons/attach/CfgVehicles.hpp
@@ -86,8 +86,8 @@ class CfgVehicles {
             dayLight = 0;
             onlyInNvg = 1;
             useFlare = 0;
-            maxLifetime = 8 * 60 * 60;
-            blinkingPattern[] = {0.1, 0.9}; // 0.1 s on, 0.9 s off
+            maxLifetime = "8 * 60 * 60";
+            blinkingPattern[] = {0.1, 1.1}; // 0.1 s on, 1.1 s off
             blinkingStartsOn = 1;
             blinkingPatternGuarantee = 1;
         };
--- a/addons/attach/CfgVehicles.hpp
+++ b/addons/attach/CfgVehicles.hpp
@@ -86,6 +86,10 @@ class CfgVehicles {
             dayLight = 0;
             onlyInNvg = 1;
             useFlare = 0;
+            maxLifetime = 28800; // 8 hours
+            blinkingPattern[] = {0.1, 0.9}; // 0.1 s on, 0.9 s off
+            blinkingStartsOn = 1;
+            blinkingPatternGuarantee = 1;
         };
 
         side = 7;//-1=NO_SIDE yellow box,3=CIV grey box,4=NEUTRAL yellow box,6=FRIENDLY green box,7=LOGIC no radar signature
--- a/addons/attach/CfgVehicles.hpp
+++ b/addons/attach/CfgVehicles.hpp
@@ -86,7 +86,7 @@ class CfgVehicles {
             dayLight = 0;
             onlyInNvg = 1;
             useFlare = 0;
-            maxLifetime = 28800; // 8 hours
+            maxLifetime = 8 * 60 * 60;
             blinkingPattern[] = {0.1, 0.9}; // 0.1 s on, 0.9 s off
             blinkingStartsOn = 1;
             blinkingPatternGuarantee = 1;


addons/attach/CfgVehicles.hpp
```diff
@@ -86,6 +86,10 @@ class CfgVehicles {
             dayLight = 0;
             onlyInNvg = 1;
             useFlare = 0;
+            maxLifetime = 28800; // 8 hours
```
commy2, line 89:
```suggestion
            maxLifetime = 8 * 60 * 60;
```
?
Does the engine even respect high lifetimes on this simulation type?

mharis001: It should in the next update. https://feedback.bistudio.com/T158909

PabstMirror:
Adding different strobes with custom patterns would be neat for IFF
also synchronized https://www.youtube.com/watch?v=bR_z-6m7g1Q
but I'm not sure how

addons/attach/CfgVehicles.hpp
BaerMitUmlaut: Shouldn't this be quoted?

mharis001:
We could provide a couple variants with different patterns. Simplest solution would be to add additional strobe items for each of the patterns but could also allow the user to select the pattern using actions. Not sure how to handle synchronized strobes or how feasible that is.

BrettMayson:
If we delete all the strobes that are wanted to be synchronized and recreate them all at once, would that do the trick?

BaerMitUmlaut:
Just attach local copies.

TheCandianVendingMachine:
What about avoiding creating the strobe until `sync_time_start % CBA_missionTime == 0`?

addons/attach/CfgVehicles.hpp
```diff
@@ -86,6 +86,10 @@ class CfgVehicles {
             dayLight = 0;
             onlyInNvg = 1;
             useFlare = 0;
+            maxLifetime = 8 * 60 * 60;
```
BaerMitUmlaut, line 89:
```suggestion
            maxLifetime = "8 * 60 * 60";
```

addons/attach/CfgVehicles.hpp
```diff
@@ -86,6 +86,10 @@ class CfgVehicles {
...
+            blinkingPattern[] = {0.1, 0.9}; // 0.1 s on, 0.9 s off
```
BaerMitUmlaut, line 90:
```suggestion
            blinkingPattern[] = {0.1, 1.1}; // 0.1 s on, 1.1 s off
```

According to the [MS-2000 manual](http://www.ostrich.co.jp/wps/wp-content/uploads/2016/11/manual-y1-03-0275b.pdf) the blink rate should be 50±10 flashes per minute. I've checked out your branch and I have to say I prefer `blinkingPattern[] = {0.1, 1.1};`, which would meet that spec, seems a little less "nervous".

addons/attach/CfgVehicles.hpp
```diff
@@ -86,6 +86,10 @@ class CfgVehicles {
...
+            blinkingStartsOn = 1;
+            blinkingPatternGuarantee = 1;
```
BaerMitUmlaut, line 92: This property doesn't seem to work currently, you get no blinking at all if you use an on time less than a frame's duration (@dedmen?). But it doesn't hurt adding it.

BaerMitUmlaut (changes requested):

BaerMitUmlaut:
We can do additional features like synched blinking later, I would prefer to get this merged for the next release. Unless @mharis001  is already working on that?

mharis001:
I briefly experimented with adding strobes with different patterns but that can be done later. Let's get this merged for the next release.

BaerMitUmlaut (approved):

PabstMirror:
Safe to merge now, but blinking won't take effect til 2.06?

PabstMirror (approved):

mharis001:
Yes safe to merge now. These config entries will not be used until 2.06.

addons/attach/CfgVehicles.hpp
dedmen: Dunno, should be same as for that other blinking thing :D

PR #8290, opened 2021-06-23
Add optionals to validation tools

Dahlgren:
- Fix current validation issues in optionals
- Validate optionals in future builds

### DIFF
--- a/tools/sqf_validator.py
+++ b/tools/sqf_validator.py
@@ -165,14 +165,15 @@ def main():
     parser.add_argument('-m','--module', help='only search specified module addon folder', required=False, default="")
     args = parser.parse_args()
 
-    # Allow running from root directory as well as from inside the tools directory
-    rootDir = "../addons"
-    if (os.path.exists("addons")):
-        rootDir = "addons"
+    for folder in ['addons', 'optionals']:
+        # Allow running from root directory as well as from inside the tools directory
+        rootDir = "../" + folder
+        if (os.path.exists(folder)):
+            rootDir = folder
 
-    for root, dirnames, filenames in os.walk(rootDir + '/' + args.module):
-      for filename in fnmatch.filter(filenames, '*.sqf'):
-        sqf_list.append(os.path.join(root, filename))
+        for root, dirnames, filenames in os.walk(rootDir + '/' + args.module):
+            for filename in fnmatch.filter(filenames, '*.sqf'):
+                sqf_list.append(os.path.join(root, filename))
 
     for filename in sqf_list:
         bad_count = bad_count + check_sqf_syntax(filename)
--- a/optionals/compat_rh_de/CfgAmmo.hpp
+++ b/optionals/compat_rh_de/CfgAmmo.hpp
@@ -1,9 +1,9 @@
 
 class CfgAmmo
 {
-	class BulletBase;
-	class RH_50_AE_Ball: BulletBase
-	{
+    class BulletBase;
+    class RH_50_AE_Ball: BulletBase
+    {
         ACE_caliber=12.7;
         ACE_bulletLength=28.194;
         ACE_bulletMass=21.06;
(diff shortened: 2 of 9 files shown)


jonpas (approved):
Excellent!

tools/stringtable_validator.py
```diff
@@ -146,9 +146,11 @@ def main():
...
 
     # Allow running from root directory and tools directory
-    root_dir = ".."
-    if os.path.exists("addons"):
-        root_dir = "."
+    for folder in ['addons', 'optionals']:
+        # Allow running from root directory as well as from inside the tools directory
+        root_dir = "../" + folder
+        if (os.path.exists(folder)):
+            root_dir = folder
```
mharis001, line 153: I don't think this is needed. Stringtable validator already validated all `stringtable.xml` files in the project. Now, in GitHub Actions it is only checking `stringtable.xml` files in `optionals`.

tools/sqf_validator.py
```diff
@@ -165,10 +165,11 @@ def main():
...
 
-    # Allow running from root directory as well as from inside the tools directory
-    rootDir = "../addons"
-    if (os.path.exists("addons")):
-        rootDir = "addons"
+    for folder in ['addons', 'optionals']:
+        # Allow running from root directory as well as from inside the tools directory
+        rootDir = "../" + folder
+        if (os.path.exists(folder)):
+            rootDir = folder
```
mharis001, line 172: Same here. Only 13 SQF files are validated.

```
Run python3 tools/sqf_validator.py
Validating SQF
------
Checked 13 files
Errors detected: 0
SQF validation PASSED
```

tools/stringtable_validator.py
Dahlgren: Fixed

tools/sqf_validator.py
Dahlgren: Fixed

PabstMirror:
```
    # Allow running from root directory as well as from inside the tools directory
    projectPath = os.getcwd()
    if (os.path.basename(projectPath) == "tools"): projectPath = os.path.dirname(projectPath)

    for folder in ['addons', 'optionals']:
        folderPath = os.path.join(projectPath, folder)
        if (args.module != ""): folderPath = os.path.join(folderPath, args.module)
        for root, dirnames, filenames in os.walk(folderPath):
            for filename in fnmatch.filter(filenames, '*.sqf'):
                sqf_list.append(os.path.join(root, filename))
```

PabstMirror (approved):

tools/sqf_validator.py
```diff
@@ -165,14 +165,15 @@ def main():
...
-        sqf_list.append(os.path.join(root, filename))
+    for folder in ['addons', 'optionals']:
+        # Allow running from root directory as well as from inside the tools directory
+        rootDir = "../" + folder
+        if (os.path.exists(folder)):
+            rootDir = folder
+
+        for root, dirnames, filenames in os.walk(rootDir + '/' + args.module):
+          for filename in fnmatch.filter(filenames, '*.sqf'):
+            sqf_list.append(os.path.join(root, filename))
```
mharis001, line 176: Rest of the file uses 4 spaces for each indentation level. Same for other files.

tools/stringtable_validator.py
```diff
@@ -145,24 +145,26 @@ def main():
     print("Validating Stringtables")
     print("-----------------------")
 
-    # Allow running from root directory and tools directory
```
mharis001, line 148: I think the changes to `stringtable_validator.py` are not needed since it already checked all `stringtable.xml` files in the project.

tools/sqf_validator.py
Dahlgren: Huh, good catch. Seems like this bad had wrong indentation before as well 🤔

tools/stringtable_validator.py
Dahlgren: It is not required but if someone where to add a stringtable.xml into includes folder it could start complaining. Ideally we should merge the folder selection logic into some helper script.

tools/sqf_validator.py
Dahlgren: Fixed

tools/stringtable_validator.py
mharis001: Yeah, I see how that could be a problem. Would be fine to re-revert the changes (sorry 😄) or keep as is then.

mharis001 (approved):

PR #8291, opened 2021-06-23
Common - Fix unloading person on non-local vehicles

jokoho48:
- Fixes issues with not being able to unload players from vehicles that are remote to the unloaded player

### DIFF
--- a/addons/common/functions/fnc_unloadPersonLocal.sqf
+++ b/addons/common/functions/fnc_unloadPersonLocal.sqf
@@ -47,17 +47,22 @@ unassignVehicle _unit;
 [_unit] orderGetIn false;
 
 TRACE_1("Ejecting", alive _unit);
-_unit action ["Eject", vehicle _unit];
+private _vehicle = vehicle _unit;
+if (local _vehicle) then {
+    _unit action ["Eject", _vehicle];
+    // Failsafe - sometimes eject alone doesn't work, but moveOut does
+    [{
+        params ["_unit"];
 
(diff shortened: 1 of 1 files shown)


addons/common/XEH_postInit.sqf
```diff
@@ -133,6 +133,11 @@ if (isServer) then {
     _this lock (_this getVariable [QGVAR(lockStatus), locked _this]);
 }] call CBA_fnc_addEventHandler;
 
+[QGVAR(EjectVehicle), {
```
jonpas, line 136:
```suggestion
[QGVAR(ejectVehicle), {
```

jokoho48: sorry i am still testing and noticed that this still does not bring the wanted effect

jonpas:
Is this still WIP?

jokoho48:
> 
> 
> Is this still WIP?

no. else i would have a WIP tag or WIP in the title 😁

jonpas:
Can you explain how this fixes it?

jokoho48:
> 
> 
> Can you explain how this fixes it?

the generall issue is that `_unit action ["Eject", _vehicle];` only works if the _vehicle AND _unit are local. moveOut works if _vehicle is not local. so we use the old code with `_unit action ["Eject", _vehicle];` if the _vehicle is local and do moveOut if _vehicle is not local

jonpas (approved):

jonpas:
Someone else review as well!

PabstMirror (approved):

PR #8297, opened 2021-06-26
Cargo: labels

BrettMayson:
- Add support for labels on cargo items `ace_cargo_label`
- Add labels to bodybags with the unit name
image


PabstMirror:
This is great, I remember someone having an issue with multiple cargo boxes and not having any way to tell the difference between them.

https://github.com/acemod/ACE3/blob/master/docs/wiki/framework/cargo-framework.md
```
### 4.7 Adding labels to a mission's cargo objects (Added in ACE3 3.14.0)

You can add labels to loadable cargo objects by using `setVariable`:

//```
this setVariable ["ace_cargo_label", "Crayons"]
//```
```

jonpas:
How does this differ from #8023?

BrettMayson:
It differs because I was unaware of that one

PR #8298, opened 2021-06-27
VSCode - Add test/build tasks

PabstMirror:
t

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ release/*
 releases/*
 extensions/vcproj32/*
 extensions/vcproj64/*
+.vscode/*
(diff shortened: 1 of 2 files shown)


PabstMirror:
I can also just throw this in `/extras` and add `/.vscode` to git-ignore if not everyone wants this

jonpas:
Can VSCode read this from a subfolder? This would ideally live in `tools/(.)vscode`

PabstMirror:
Moved to /tools and added to gitignore
People just need to copy it from /tools to root if they want to use it

jonpas (approved):

PR #8300, opened 2021-06-27
GPS inaccuracy

BrettMayson:
Note sure if there is interest in ACE for this, or if I should just do this in-house or as a standalone mod.

- Add a function to get the position of the player, but with an offset depending on view of the "satellites"

Some future notes:
- Would be disabled by default
- Would be disabled if the vanilla blufor tracking is on
- When disabled, the function would just return `getPos _object`

Compatibility:
- Not sure how / if I'll be able to get the vanilla GPS minimap to co-operate
- Mods like cTab or [FriendlyTracker](https://github.com/ArmaForces/FriendlyTracker) would have to choose to call the function when ACE is loaded

### DIFF
--- a/addons/gps/$PBOPREFIX$
+++ b/addons/gps/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\ace\addons\gps
--- a/addons/gps/CfgEventHandlers.hpp
+++ b/addons/gps/CfgEventHandlers.hpp
@@ -1,0 +1,17 @@
+class Extended_PreStart_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preStart));
+    };
+};
+
+class Extended_PreInit_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preInit));
+    };
+};
+
+class Extended_PostInit_EventHandlers {
+    class ADDON {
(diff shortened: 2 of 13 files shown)


ColdEvul:
Could it be possible to make the marker update in intervals instead of floating with the player. 

Cause what it feels like now is that if you just wait you can still get exact position rather easy on were you are regardless of the inaccuracy. By having a intervall update it could make it harder to pinpoint excaxt position.

Unless this is possible already by playing with the BRF settings update intervall. I'm rather bad on the BRF system and how it works.
All in all i like inaccuracies!

TheMagnetar:
This should also affect the height. Essentially taking the height of the new  "inaccurate" position. Note however that military positioning accuracy is down to 1cm.

Drofseh:
This will be great for missions set in the 80s and 90s

BrettMayson:
> 
> 
> This should also affect the height. Essentially taking the height of the new "inaccurate" position. Note however that military positioning accuracy is down to 1cm.

Do you have a source for this? I am unable to find anything suggesting military GPS is that accurate

BrettMayson:
The hold up here, is the GPS minimap. I can not find a way to modify the position of the player, and have had troubles also with hiding it an drawing a different map in its place in an attempt to bypass any engine code. If someone is able to figure out a way this PR could one day be merged, otherwise it'll likely be abandonded as I'm done trying to get something to work.

PR #8313, opened 2021-07-10
DAGR - Align underlining in waypoint editing page

### LINKED ISSUE #8293: Digits are underlined incorrectly in DAGR waypoint input window
- **Arma 3:** `2.04`
- **CBA:** `3.15.1`
- **ACE3:** `3.13.6`

**Description:**
The underlines are off, it seems that their horizontal boundaries are too big. On my example the first digit is selected, but the underline is off.

**Steps to reproduce:**
Go to WP list menu in DAGR, go through the digits.

**Screenshots:**
image

LupusTheCanine:
- DAGR - Fix #8293 

I simply took 2nd and 7th underlines as they are aligned with 1st and 8th digit respectively and recalculated x values based on these. It is not perfect but definitely good enough.
image

### DIFF
--- a/addons/dagr/Dialog.hpp
+++ b/addons/dagr/Dialog.hpp
@@ -357,9 +357,10 @@ class DAGR_Menu {
         sizeEx = 0.04;
     };
 
+//Waypoint adding and modification digits underline
     class DAGR_MENU_PSELECTION1 : DAGR_Menu_Pic {
         idc = 2668788;
-        x = 0.451;
+        x = 0.465;
         y = 0.352;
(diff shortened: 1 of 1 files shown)


jonpas (approved):
This seems fine to me, but I don't use the DAGR.

PabstMirror:
it was still a little off at 1280x1024xsmall
image
it might be a non-monospaced font or ui scaling issue

but overall a big improvement from master, so I think it's good for now

PabstMirror (approved):

PR #8318, opened 2021-07-18
Common - Improve vehicle icon caching implementation

kymckay:
- Should be more optimised using a native hashmap. This function is used in spectator and for vehicle actions.

### DIFF
--- a/addons/common/functions/fnc_getVehicleIcon.sqf
+++ b/addons/common/functions/fnc_getVehicleIcon.sqf
@@ -25,7 +25,7 @@ private _objectType = if (_object isEqualType objNull) then {
 } else {
     _object
 };
-private _cachedValue = GVAR(vehicleIconCache) getVariable _objectType;
+private _cachedValue = GVAR(vehicleIconCache) getOrDefault [_objectType];
 
 if (isNil "_cachedValue") then {
     private _vehicleValue = getText (configfile >> "CfgVehicles" >> _objectType >> "icon");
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_getVehicleIcon.sqf
```diff
@@ -25,7 +25,7 @@ private _objectType = if (_object isEqualType objNull) then {
 } else {
     _object
 };
-private _cachedValue = GVAR(vehicleIconCache) getVariable _objectType;
+private _cachedValue = GVAR(vehicleIconCache) getOrDefault [_objectType];
```
veteran29, line 28:
```suggestion
private _cachedValue = GVAR(vehicleIconCache) get _objectType;
```

veteran29 (commented):
After this change, the cache will be case-sensitive but this should not matter.

addons/common/functions/fnc_getVehicleIcon.sqf
kymckay: @veteran29 does just `get` return nil when the key is missing?

I assumed it didn't since the default value is optional in the other version.

kymckay: Wait... by definition it must. I'm out of SQF practice so was thinking there'd be an exception.

veteran29 (approved):

jonpas (approved):

addons/common/functions/fnc_getVehicleIcon.sqf
```diff
@@ -25,7 +25,7 @@ private _objectType = if (_object isEqualType objNull) then {
 } else {
     _object
```
dedmen, line 26: if you want to be save case-sensitive wise.

configName (configFile >> "CfgVehicles" >> _object)

PR #8321, opened 2021-07-20
Adds the indication of applied CATs/Splints

10Dozen:
- Adds indication of the applied CATs and Splints to Stance indicator (appropriate icon will be drawn when at least 1 item of the type is appied)
- Indication may be disabled via settings

Note:
CATs indicator is to remeber that tourniquet is applied (by player or by other person) and it should be removed when possible.
Splint indicator is pretty discussable, maybe it's better to show 'fractured' icon instead to give player notification that something is really bad with his body


10Dozen:
As discussed in Slack to be changed as:
- show fractured status icon if any
- show 'splinted' status icon if any AND splint effect settings is not 'fully heals' to indicate that player suffers some penalty

Also make icons a bit flexible by adding icon to the most top position, but sorted (1) fractures, (2) CATs, (3) splints

10Dozen:
Updated logic of indicators to the next: indicators fulfilled from top to bottom. 
Fractures are priority 1 and are always on top (if fractures enabled and player has any).
Tourniquets has priority 2, icon drawn if player has any applied.
Splinted fracture indicator has priority 3 and only shown when (a) fractures enabled, (b) fracture setting is 2 or 3 (penalty remains after usage of the splint), (c) player has any splinted fracture.

Indication

addons/medical_feedback/functions/fnc_handleGUIIndication.sqf
```diff
@@ -0,0 +1,63 @@
...
+
+// --- Removes any indication and exit
+if (_dropIndicator) exitWith {
+    {
+        _x ctrlSetText "";
+    } forEach _indicatorSlots;
+};
+
+// --- Tourniquets
+private _hasTourniquets = {_x > 0} count GET_TOURNIQUETS(ACE_player) > 0;
```
veteran29, line 33: Using `count` for "has" checks is not optimal, `findIf` stops iteration on first found element.
```suggestion
private _hasTourniquets = GET_TOURNIQUETS(ACE_player) findIf {_x > 0} != -1;
```

addons/medical_feedback/functions/fnc_handleGUIIndication.sqf
```diff
@@ -0,0 +1,63 @@
...
+private _tourniquetIcon = ["",ICON_TOURNIQUET_PATH] select _hasTourniquets;
+
+// --- Fractures and Splints
+private _fractureSettings = EGVAR(medical,fractures);
+private _fractureIcon = "";
+private _splintIcon = "";
+
+if (_fractureSettings > 0) then {
+    // --- Fractures enabled: check for fracture indication
+    private _hasFractures = {_x > 0} count GET_FRACTURES(ACE_player) > 0;
```
veteran29, line 43:
```suggestion
    private _hasFractures = GET_FRACTURES(ACE_player) findIf {_x > 0} != -1;
```

addons/medical_feedback/functions/fnc_handleGUIIndication.sqf
```diff
@@ -0,0 +1,63 @@
...
+    _fractureIcon = ["",ICON_FRACTURE_PATH] select _hasFractures;
+
+    if (_fractureSettings > 1) then {
+        // --- Fractures can not be fully healed: check for splint indication
+        private _hasSplints = {_x isEqualTo -1} count GET_FRACTURES(ACE_player) > 0;
```
veteran29, line 48:
```suggestion
        private _hasSplints = GET_FRACTURES(ACE_player) findIf {_x isEqualTo -1} != -1;
```

veteran29 (changes requested):

addons/medical_feedback/functions/fnc_handleGUIIndication.sqf
10Dozen: It's a bit unclear what syntax is preferred:
A) `X findIf {Y} != -1`
B) `X findIf {Y} > -1`

I see both A and B used across the repo.
Why not 'isNotEqualTo' instead of `!=`? Is it just for specific data types? (found nothing here https://ace3mod.com/wiki/development/coding-guidelines.html)

jonpas:
Should "CAT" on the status indicator be "TQ"? I don't know what is usually used, but CAT is a specific tourniquet trademarked by whichever company makes it.

10Dozen:
> Should "CAT" on the status indicator be "TQ"? I don't know what is usually used, but CAT is a specific tourniquet trademarked by whichever company makes it.

Same as QuikClot. Using 'CAT' is good in terms of different languages - most of translations include [CAT], but TQ is not so widely used.

Anyway text may be removed at all, I was not sure that the icon is self-explanatory. I can make new one if somebody provide an example (photo or icon). Maybe something like this? 
image
image
Silhouette will be pretty specific.

jonpas:
Yeah, that's fine then.

10Dozen:
Update icon for tourniquet, now icons are:
image

jonpas:
Nice and uniform! 👌

addons/medical_feedback/RscInGameUI.hpp
```diff
@@ -1,13 +1,36 @@
...
+        class GVAR(stateIndicator1): RscPictureKeepAspect {
+            onLoad = QUOTE(uiNamespace setVariable [ARR_2(QQGVAR(stateIndicator1), _this select 0)]);
+            x = IGUI_GRID_STANCE_X + IGUI_GRID_STANCE_WAbs * 3 / 4;
+            y = IGUI_GRID_STANCE_Y;
+            w = IGUI_GRID_STANCE_WAbs / 4;
+            h = IGUI_GRID_STANCE_HAbs / 4;
+        };
+        class GVAR(stateIndicator2): GVAR(stateIndicator1) {
+            onLoad = QUOTE(uiNamespace setVariable [ARR_2(QQGVAR(stateIndicator2), _this select 0)]);
+            x = IGUI_GRID_STANCE_X + IGUI_GRID_STANCE_WAbs * 3 / 4;
```
mharis001, line 27: This property is the same as the first indicator and will be inherited.

10Dozen: Removed 'x' definition from childs

addons/medical_feedback/functions/fnc_handleEffects.sqf
```diff
@@ -58,4 +58,9 @@ if ((!GVAR(heartBeatEffectRunning)) && {_heartRate != 0} && {(_heartRate > 160)
 [!_unconscious, _pain] call FUNC(effectPain);
 [!_unconscious, _bleedingStrength, _manualUpdate] call FUNC(effectBleeding);
 
+// - Tourniquets, fractures and splints indication ---------------------------------------
+if (GVAR(GUIIndication)) then {
```
veteran29, line 62: Not a fan of naming it `GUIIndication`, I think something like `HUDIndicators` would be more adequate. But that's just a suggestion.

veteran29 (approved):

addons/medical_feedback/functions/fnc_handleGUIIndication.sqf
veteran29: After a quick regex search, it seems that `>` is used more frequently than `!=` in the ACE repo. However, there's no guideline on that.

As for `isNotEqualTo`, it has no benefits over `!=` for number type and it's longer to read.

10Dozen: Thanks for explanations.

addons/medical_feedback/stringtable.xml
```diff
@@ -189,6 +189,14 @@
                 <Chinese>啟用傷者的尖叫聲</Chinese>
                 <Turkish>Yaralı birimler tarafından çığlık atmayı etkinleştir</Turkish>
             </Key>
+            <Key ID="STR_ACE_Medical_Feedback_enableHUDIndication_DisplayName">
+                <English>Enable Fractures/Tourniquets Indication</English>
```
mharis001, line 193: Agree with @veteran29 , "Indicators" sounds more natural for this feature (here and in other places) IMO.

10Dozen: Oh, missed the tailings. Yeah, sounds good.

veteran29 (approved):

PR #8327, opened 2021-07-22
Vehicles - Add configurable engine startup delay

severgun:
- Add configurable engine startup delay by config and script
- Add function to set delay per vehicle
- Set default delay to 1.3 as it was before
- Add vehicles framework wiki page

### DIFF
--- a/addons/vehicles/functions/fnc_startEngine.sqf
+++ b/addons/vehicles/functions/fnc_startEngine.sqf
@@ -20,6 +20,8 @@ params ["_vehicle", "_isEngineOn"];
 
 if (!_isEngineOn || {floor abs speed _vehicle > 0 || {!isNull isVehicleCargo _vehicle}}) exitWith {};
 
+private _startupDelay = getNumber(configOf _vehicle >> QGVAR(engineStartDelay));
+
 [{
     params ["_args", "_idPFH"];
     _args params ["_vehicle", "_time", "_direction"];
@@ -29,4 +31,4 @@ if (!_isEngineOn || {floor abs speed _vehicle > 0 || {!isNull isVehicleCargo _ve
     _vehicle setVelocity [0, 0, 0];
     _vehicle setVectorDirAndUp _direction;
 
-} , 0, [_vehicle, CBA_missionTime + STARTUP_DELAY, [vectorDir _vehicle, vectorUp _vehicle]]] call CBA_fnc_addPerFrameHandler;
+} , 0, [_vehicle, CBA_missionTime + _startupDelay, [vectorDir _vehicle, vectorUp _vehicle]]] call CBA_fnc_addPerFrameHandler;
--- a/addons/vehicles/functions/fnc_startEngine.sqf
+++ b/addons/vehicles/functions/fnc_startEngine.sqf
@@ -20,7 +20,7 @@ params ["_vehicle", "_isEngineOn"];
 
 if (!_isEngineOn || {floor abs speed _vehicle > 0 || {!isNull isVehicleCargo _vehicle}}) exitWith {};
 
-private _startupDelay = getNumber(configOf _vehicle >> QGVAR(engineStartDelay));
+private _startupDelay = getNumber (configOf _vehicle >> QGVAR(engineStartDelay));
 
 [{
     params ["_args", "_idPFH"];
--- a/addons/vehicles/functions/fnc_startEngine.sqf
+++ b/addons/vehicles/functions/fnc_startEngine.sqf
@@ -20,7 +20,7 @@ params ["_vehicle", "_isEngineOn"];
 
 if (!_isEngineOn || {floor abs speed _vehicle > 0 || {!isNull isVehicleCargo _vehicle}}) exitWith {};
 
-private _startupDelay = getNumber (configOf _vehicle >> QGVAR(engineStartDelay));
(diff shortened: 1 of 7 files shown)


BrettMayson:
It'd be nice to also add a variable to the vehicle object itself so mission makers can easily adjust it for specific vehicles.

Also perhaps have an option to scale it with engine health?

severgun:
> Also perhaps have an option to scale it with engine health?

Sounds interesting as a feature, but will look weird, because there is no sounds of failing engine.

Dystopian (changes requested):
I would also suggest to set value for `Truck_F` ~ 3. And I would want to handle tanks (with value ~ 7) but then eventhandlers should be edited too.

severgun:
> I would also suggest to set value for `Truck_F` ~ 3. And I would want to handle tanks (with value ~ 7) but then eventhandlers should be edited too.

I can make it. But this will change known behaviour for all ACE3 users. Approve needed.
You can make your custom addon with custom timings.

jonpas:
I think large or different wait times might be problematic from an already existing missions standpoint.

addons/vehicles/functions/fnc_startEngine.sqf
```diff
@@ -20,6 +20,8 @@ params ["_vehicle", "_isEngineOn"];
 
 if (!_isEngineOn || {floor abs speed _vehicle > 0 || {!isNull isVehicleCargo _vehicle}}) exitWith {};
 
+private _startupDelay = getNumber (configOf _vehicle >> QGVAR(engineStartDelay));
```
jonpas, line 23:
```suggestion
private _startupDelay = _vehicle getVariable [QGVAR(engineStartDelay), -1];
if (_startupDelay == -1) then {
    _startupDelay = getNumber (configOf _vehicle >> QGVAR(engineStartDelay));
};
```
This should satisfy @BrettMayson's request.

severgun: Check new commit, please. Maybe I should revert and apply your's

jonpas: What you have looks good. Only thing I would maybe name the public function `setVehicleStartDelay`, as you are setting it, not adding it.

severgun: Done

jonpas (approved):

addons/vehicles/functions/fnc_startEngine.sqf
```diff
@@ -20,6 +20,8 @@ params ["_vehicle", "_isEngineOn"];
 
 if (!_isEngineOn || {floor abs speed _vehicle > 0 || {!isNull isVehicleCargo _vehicle}}) exitWith {};
 
+private _startupDelay = _vehicle getVariable [QGVAR(engineStartDelay), getNumber (configOf _vehicle >> QGVAR(engineStartDelay))];
+
```
PabstMirror, line 24:
```suggestion
if (_startupDelay <= 0) exitWith {};

```

docs/wiki/framework/vehicles-framework.md
```diff
@@ -0,0 +1,54 @@
...
+    class MyElectricCar {
+        ace_vehicles_engineStartDelay = 0.1;
+    };
+};
+```
+
+### 1.2 Setting the startup delay by script
+
+`ace_vehicles_fnc_setVehicleStartDelay`
+
```
PabstMirror, line 39:
```suggestion
Has global effects, so can be called just on server.

```

addons/vehicles/functions/fnc_startEngine.sqf
severgun: Why 0 not allowed? What if someone want play spaceships

docs/wiki/framework/vehicles-framework.md
severgun: setVariable[.., .., true] will broadcast value from any PC. Isn't it?

veteran29:
I do not like the fact that it basically turns the vehicle into an immovable object for the duration of the engine startup.

What happens if you will drive off a cliff, disable and enable the engine?

`setCruiseControl` should be checked if it can be used as an alternative.

severgun:
> `setCruiseControl` should be checked if it can be used as an alternative.

This PR does not introduce new mechanics. Engine startup delay in such implementation was here for years.
This PR just replace magic constat with configurable value.

veteran29:
> This PR does not introduce new mechanics.

I've completely missed that it edits existing code, I've forgotten that we have such feature :D

docs/wiki/framework/vehicles-framework.md
PabstMirror:
```suggestion
Has global effects.

```
Yeah any machine would be fine, I thought it might be good to mention because a lot of people will probably but it in the init field. although there is no real harm if the setVar is rerun lots of times

PabstMirror:
@veteran29 - https://www.youtube.com/watch?v=Wpa5JgI_qOw
but it's mostly a joke, because we already check `floor abs speed _vehicle > 0` (so only works if you drop straight down)

addons/vehicles/functions/fnc_startEngine.sqf
PabstMirror: this is just to avoid adding and removing the `CBA_fnc_addPerFrameHandler`
if the delay is 0 it shouldn't do anything anyway

PabstMirror (approved):

PR #8329, opened 2021-07-22
Common - Fix `addToInventory` replacing existing weapon

### LINKED ISSUE #8328: Common - addToInventory replaces weapons
- **Arma 3:** `2.04 stable`
- **CBA:** `3.15.5`
- **ACE3:** `3.3.16`

**Description:**
If you add a weapon through [addToInventory](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_addToInventory.sqf) while having enough space for the weapon but not the available weapon slot it will replace the weapon currently in that slot. This is not happening when you add it to a specific container.

**Steps to reproduce:**
1. Get a big enough backpack to fit a weapon and equip weapons.
2. Add another weapon with [addToInventory](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_addToInventory.sqf).
3. See that currently equipped weapon got replaced

**Expected behavior:**
To have the weapon go into the inventory or drop on the ground instead.

- Editor (Singleplayer)

**Log Files:**
- [rpt](https://gist.github.com/JoramD0/032eebc18843d78bcb35e3c1d6748cf2)

**Additional context:**
Cause of the issue:
CBA returns the weapon can fit into the inventory ([here](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_addToInventory.sqf#L40)), then ACE adds the weapon with [addWeaponGlobal](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_addToInventory.sqf#L60) which replaces the currently equipped weapon.

I tried thinking of a solution myself but couldn't think of anything good.

**Screenshots:**
N/A

JoramD0:
- Fix the weapon replacing bug/issue in addToInventory.

fixes #8328

### DIFF
--- a/addons/common/functions/fnc_addToInventory.sqf
+++ b/addons/common/functions/fnc_addToInventory.sqf
@@ -78,7 +78,7 @@ switch (_type select 0) do {
                             _x params ["_parameters", "_container"];
 
                             if (_parameters call CBA_fnc_canAddItem) exitWith {
-                                _container addWeaponCargoGlobal [_classname, 1];
+                                _container addWeaponCargoGlobal [_classname, 1]; // addWeaponGlobal will replace the weapon currently in a slot
                             };
                         } forEach [
                             [[_unit, _classname, 1, false, false, true], backpackContainer _unit],
--- a/addons/common/functions/fnc_addToInventory.sqf
+++ b/addons/common/functions/fnc_addToInventory.sqf
@@ -50,6 +50,9 @@ switch (_container) do {
                 case "handgun": {
                     _canFitWeaponSlot = handgunWeapon _unit == "";
                 };
+                case "binocular": {
+                    _canFitWeaponSlot = binocular _unit == "";
+                };
             };
         };
     };
--- a/addons/common/functions/fnc_addToInventory.sqf
+++ b/addons/common/functions/fnc_addToInventory.sqf
@@ -40,8 +40,7 @@ switch (_container) do {
     default {
         _canAdd = [_unit, _classname] call CBA_fnc_canAddItem;
         if (_canAdd) then {
-            private _slot = [_classname] call FUNC(getItemType);
-            switch (_slot select 1) do {
+            switch (_type select 1) do {
                 case "primary": {
                     _canFitWeaponSlot = primaryWeapon _unit == "";
                 };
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_addToInventory.sqf
```diff
@@ -38,12 +39,26 @@ switch (_container) do {
     };
     default {
         _canAdd = [_unit, _classname] call CBA_fnc_canAddItem;
+        if (_canAdd) then {
+            private _slot = [_classname] call ace_common_fnc_getItemType;
```
PabstMirror, line 43:
```suggestion
            private _slot = [_classname] call FUNC(getItemType);
```

addons/common/functions/fnc_addToInventory.sqf
```diff
@@ -38,12 +39,26 @@ switch (_container) do {
     };
     default {
         _canAdd = [_unit, _classname] call CBA_fnc_canAddItem;
+        if (_canAdd) then {
+            private _slot = [_classname] call FUNC(getItemType);
```
mharis001, line 43: `FUNC(getItemType)` is already called on L24.

addons/common/functions/fnc_addToInventory.sqf
```diff
@@ -57,7 +72,34 @@ switch (_type select 0) do {
...
+                            case "vest": {
+                                (vestContainer _unit) addWeaponCargoGlobal [_classname, 1];
+                            };
+                            case "backpack": {
+                                (backpackContainer _unit) addWeaponCargoGlobal [_classname, 1];
+                            };
+                            case "uniform": {
+                                (uniformContainer _unit) addWeaponCargoGlobal [_classname, 1];
+                            };
+                        };
```
mharis001, line 101: I think you can get rid of `_containerWithSpace` and just use `uniformContainer _unit`, `vestContainer _unit`, and `backpackContainer _unit` instead of the "uniform", "vest", and "backpack" strings. Doing that, L83 would become: `_container addWeaponCargoGlobal [_classname, 1];`.

JoramD0: Would not have thought of that, thank you!

jonpas (changes requested):
Perfect, but I'd like to see a comment somewhere why this is necessary. A simple "`addWeaponGlobal` will replace the weapon currently in a slot" would do.

jonpas (approved):

addons/common/functions/fnc_addToInventory.sqf
```diff
@@ -38,12 +39,25 @@ switch (_container) do {
...
+            switch (_type select 1) do {
+                case "primary": {
+                    _canFitWeaponSlot = primaryWeapon _unit == "";
+                };
+                case "secondary": {
+                    _canFitWeaponSlot = secondaryWeapon _unit == "";
+                };
+                case "handgun": {
+                    _canFitWeaponSlot = handgunWeapon _unit == "";
+                };
```
PabstMirror, line 52:
```suggestion
                };
                case "binocular": {
                    _canFitWeaponSlot = binocular _unit == "";
                };
```

mharis001 (approved):

PabstMirror (approved):

PR #8338, opened 2021-07-25
Dragging - Allow run when carrying light-weight objects

Dystopian:
- title.

Only sprint is blocked for light-weight objects.
`MAX_WEIGHT_RUN==50` is enough to carry 1 MX rifle or about 10 MX magazines or 1 AT rocket or 2 HE rockets.

Should be merged alongside #6029 but no problems without it.


addons/common/XEH_postInit.sqf
```diff
@@ -20,7 +20,7 @@
 //Status Effect EHs:
 [QGVAR(setStatusEffect), {_this call FUNC(statusEffect_set)}] call CBA_fnc_addEventHandler;
 ["forceWalk", false, ["ace_advanced_fatigue", "ACE_SwitchUnits", "ACE_Attach", "ACE_dragging", "ACE_Explosives", "ACE_Ladder", "ACE_Sandbag", "ACE_refuel", "ACE_rearm", "ACE_Trenches", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
-["blockSprint", false, ["ace_advanced_fatigue", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
+["blockSprint", false, ["ace_advanced_fatigue", "ACE_dragging", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
```
jonpas, line 23:
```suggestion
["blockSprint", false, ["ace_advanced_fatigue", "ace_dragging", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
```
Should really be like that everywhere, but won't break either way.

Dystopian: was added like in `forceWalk` line :man_shrugging:

shukari:
Is there a possibility to get this not as define... so other people can change it in the mission?

Dystopian:
> 
> 
> Is there a possibility to get this not as define... so other people can change it in the mission?

can do like here https://github.com/acemod/ACE3/blob/27bd25c4b8c94e28bd7ea60f053be5365ec709f8/addons/dragging/XEH_postInit.sqf#L16

shukari:
That would be great. It is a shame that so many things are now marcos so it can't be changed easy.

Drofseh:
Having it as a setting would be nice

Dystopian:
> 
> 
> Having it as a setting would be nice

This I can't do, sorry.

severgun:
> This I can't do, sorry.

Something like that in `XEH_preInit.sqf` and corresponding stringtables entries.
```sqf
[
    "ACE_maxWeightCarry", "SLIDER",
    [LSTRING(maxWeightCarry_displayName), LSTRING(maxWeightCarry_description)],
    LSTRING(weight_category),
    [0,10000,600,0],
    true
] call CBA_fnc_addSetting;

[
    QGVAR(maxWeightCarryRun), "SLIDER",
    [LSTRING(maxWeightCarryRun_displayName), LSTRING(maxWeightCarryRun_description)],
    LSTRING(weight_category),
    [0,1000,50,0],
    true
] call CBA_fnc_addSetting;
```

Dystopian:
- create stringtable values for 3 names and descriptions and their translating;
- edit documentation;
- testing and fixing.

Don't get me wrong, I just don't need all this work. I even think this setting is really useless for anyone and global variable is more than enough. Please feel free to make your PRs for it.

severgun:
> edit documentation

You introduce new magic variable and need to do this as well.

> Please feel free to make your PRs for it.

Ok I will.

10Dozen:
Maybe some weight check inside PFH? As i understand it is possible to load more items into carried box/weapon holder by other players, ace_common_fnc_claim only blocks accessing items for drag/carry/disassemble.

BaerMitUmlaut:
That's a good point @10Dozen, but I would classify this as a general bug, not a problem specific for this PR. Maybe we should use [`lockInventory`](https://community.bistudio.com/wiki/lockInventory) to block inventory access?

10Dozen:
> Maybe we should use lockInventory to block inventory access?
TBH when carriable weaponHolders will be added - ability to access carried holder will be pretty useful. E.g. you can pick some AT launcher, scroll up to move it higher and other players (that are sitting on rooftop of a small house) will be able to take it right from your hands. 
So maybe InventoryClosed EH with quantity/mass re-calculation and checks?

addons/dragging/XEH_postInit.sqf
```diff
@@ -15,6 +15,9 @@ if (isNil "ACE_maxWeightDrag") then {
 if (isNil "ACE_maxWeightCarry") then {
     ACE_maxWeightCarry = 600;
 };
+if (isNil QGVAR(maxWeightCarryRun)) then {
+    GVAR(maxWeightCarryRun) = 50;
```
LinkIsGrim, line 19: This seems a bit low. Max weight of 10lbs (`FUNC(getWeight) multiplies `mass` by 0.5) while running? 100 would be more appropriate, I think.

addons/dragging/functions/fnc_startCarry.sqf
```diff
@@ -56,7 +57,15 @@ if (_target isKindOf "CAManBase") then {
     _unit action ["SwitchWeapon", _unit, _unit, 299];
     [_unit, "AmovPercMstpSnonWnonDnon", 0] call EFUNC(common,doAnimation);
 
-    [_unit, "forceWalk", "ACE_dragging", true] call EFUNC(common,statusEffect_set);
+    // objects other than containers have calculated weight == 0 so we use getMass
+    if (-1 == ["ReammoBox_F", "WeaponHolder", "WeaponHolderSimulated"] findIf {_target isKindOf _x}) then {
+        _weight = getMass _target;
+    };
```
LinkIsGrim, line 63: This can go. `FUNC(getWeight)` already does `getMass`.

LinkIsGrim (changes requested):

LinkIsGrim (approved):

jonpas (approved):

PR #8345, opened 2021-08-01
Medical - Refactor Clear Trauma Setting

Salluci:
- Add in a setting for clearing trauma after stitching a wound.
- Refactor clearing trauma after bandage to clear trauma for individual wounds as they're bandaged.
- Previous behaviour was to set damage on body part to 0 after all wounds on it were bandaged.


PabstMirror:
I would like to see an setting to keep the old logic
- if you are shot in the leg and limping, you will continue to limp until that limb has been fully treated

Salluci:
Could you elaborate? IIRC, post-rewrite limping only happens due to fractures or the limp setting, which this doesn't touch.

PR #8347, opened 2021-08-02
Repair - Add support for RHS optic hitpoints

### LINKED ISSUE #8304: RHS Compat - Bradley IFV lacks repair points for optics
- **Arma 3:** `2.04` [stable]
- **CBA:** `3.15.1` [stable]
- **ACE3:** `3.13.6` [stable]
- **RHSUSAF:** `0.5.6` [stable]
- **ACE Compat - RHS United States Armed Forces:** `3.13.6` [stable]

**Description:**
The Bradley IFV from RHS lacks repair points for its optics, and can only be fixed by fully repairing it. 

**Steps to reproduce:**
_Follow [https://ace3mod.com/img/wiki/user/issue_flowchart.png](this flowchart)!_

1. Spawn an M2A3 Bradley IFV - I used specifically the BUSK III for this
2. execute `[] execVM "\z\ace\addons\repair\dev\draw_showRepairInfo.sqf";` to see repair points

**Expected behavior:**
Repair points would show for the Bradley's optics, allowing crews to repair it in the field if damaged

- Multiplayer with more mods, and singleplayer with mods listed at the start to confirm 

**Log Files:**
- NA

**Additional context:**
The Bradley's optics already are very weak and can blackout/break really easily, and without a way to repair it in the field, can make it inoperable by a few stray bullets hitting the optics. An ability to repair the optics, or to carry some sort of replacement optics would be very helpful for people using it. 

I took the guide used from [this post on the RHS Stryker](https://github.com/acemod/ACE3/issues/7245) to find the hitpoints and figured that it was possible to do for the Bradley as well since the issue was fixed on the Stryker. 

**Screenshots:**
Hitpoints:

Salluci:
- Update fnc_addRepairActions to include hitpoints that influence turret view.
- Hopefully closes #8304.

### DIFF
--- a/addons/repair/functions/fnc_addRepairActions.sqf
+++ b/addons/repair/functions/fnc_addRepairActions.sqf
@@ -94,8 +94,8 @@ private _turretPaths = ((fullCrew [_vehicle, "gunner", true]) + (fullCrew [_vehi
                 private _hitpointsCfg = "configName _x == _hitpoint" configClasses _turretHitpointCfg;
                 if (_hitpointsCfg isNotEqualTo []) exitWith {
                     TRACE_2("turret hitpoint configFound",_hitpoint,_x);
-                     // only do turret hitpoints for now or we get some weird stuff
-                    if ((_hitpoint in ["hitturret", "hitgun"]) || {(getNumber (_hitpointsCfg # 0 >> "isGun")) == 1} || {(getNumber (_hitpointsCfg # 0 >> "isTurret")) == 1}) then {
+                     // only do turret hitpoints or stuff linked to visuals for now or we apparently get some weird stuff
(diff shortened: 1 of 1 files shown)


Salluci:
Side effects may include unnecessary hitpoints being shown when the vehicle is damaged through irregular means (basically anything using setDamage and affecting all hitpoints). This mostly manifests in the RHS M1126 having 7 commander periscope hitpoints, but at least they're all in the correct position, and those hitpoint don't seem to be damaged through regular gameplay.

I'm not sure if there's any cleaner way to implement this. Making patches using config-based hitpoint groups would be ideal, but would be more work for the compats. Other option would be to lock these "weird" hitpoints behind a client setting that determines whether they're shown or not.

Salluci:
> Can you still override "repairable hitpoints" through the config API we have after this addition?

Yes.

PR #8350, opened 2021-08-04
Medical - Add ability to customize Medical GUI colors

severgun:
**When merged this pull request will:**
- Add CBA Settings entries to customize color scheme used in medical menu
- Add English and Russian locale
- PR related to #7915

PR #8354, opened 2021-08-05
Change hemorrhage text colors

severgun:
- Change hemorrhage text color from always red to yellow-red grade.

There was feature request to change `Lost some blood` text color from red to white. I think yellow is better to show that there is some loss but not so "screaming" as red.

https://gka.github.io/palettes/#/4|s|ffff00,ff0000|ffffe0,ff005e,93003a|0|0
```
#ffff00
#ffaa00
#ff5500
#ff0000
```


severgun:
Or should I take colors from #8350 ?
For example 3, 5, 7, 9

veteran29 (approved):

PR #8362, opened 2021-08-06
WIP - Various - Use uniqueUnitItems

LinkIsGrim:
- Title.
- Add support for checking against multiple items in common_hasItem.
- Add support for checking against magazines and weapons in common_hasItem.
- Change medical_treatment_hasItem to use common_hasItem.
- Change repair_hasItems to use common_hasItem.
- Change disarming_getAllGearUnit to use uniqueUnitItems.
- Change various functions to use common_hasItem instead of presence of item in uniqueItems array.
- Fix common_useMagazine to return false if magazine is not in unit's inventory.

Requires Arma v2.06.

### DIFF
--- a/addons/arsenal/functions/fnc_onArsenalClose.sqf
+++ b/addons/arsenal/functions/fnc_onArsenalClose.sqf
@@ -70,6 +70,9 @@ if (!isNil QGVAR(moduleUsed)) then {
     objNull remoteControl GVAR(center);
 };
 
+// Update player unique items cache
+ACE_player call EFUNC(common,uniqueItems);
+
 ACE_player switchCamera GVAR(cameraView);
 
 if (isMultiplayer) then {
--- a/addons/artillerytables/functions/fnc_interactMenuOpened.sqf
+++ b/addons/artillerytables/functions/fnc_interactMenuOpened.sqf
@@ -96,7 +96,7 @@ TRACE_2("searching for new vehicles",_vehicleAdded,_rangeTablesShown);
             };
             private _condition = {
                 //IGNORE_PRIVATE_WARNING ["_player"];
-                ([_player, "ACE_artilleryTable"] call EFUNC(common, hasItem)) && {[_player, objNull, ["notOnMap", "isNotSitting", "isNotInside"]] call EFUNC(common,canInteractWith)}
+                ([_player, "ACE_artilleryTable"] call EFUNC(common,hasItem)) && {[_player, objNull, ["notOnMap", "isNotSitting", "isNotInside"]] call EFUNC(common,canInteractWith)}
             };
             private _displayName = format ["%1%2", getText (_vehicleCfg >> "displayName"),["","*"] select _advCorrection];
             private _action = [format ['QGVAR(%1)',_index], _displayName, QPATHTOF(UI\icon_rangeTable.paa), _statement, _condition, {}, _info] call EFUNC(interact_menu,createAction);
--- a/addons/artillerytables/functions/fnc_interactMenuOpened.sqf
+++ b/addons/artillerytables/functions/fnc_interactMenuOpened.sqf
@@ -19,7 +19,7 @@ params ["_menuType"];
 TRACE_1("interactMenuOpened",_menuType);
 
 if (_menuType != 1) exitWith {};
-if (!("ACE_artilleryTable" in (ace_player call EFUNC(common,uniqueItems)))) exitWith {};
+if !([ace_player, "ACE_artilleryTable"] call EFUNC(common,hasItem)) exitWith {};
 
 private _vehicleAdded = ace_player getVariable [QGVAR(vehiclesAdded), []];
 private _rangeTablesShown = ace_player getVariable [QGVAR(rangeTablesShown), []];
@@ -96,7 +96,7 @@ TRACE_2("searching for new vehicles",_vehicleAdded,_rangeTablesShown);
             };
             private _condition = {
                 //IGNORE_PRIVATE_WARNING ["_player"];
-                ("ACE_artilleryTable" in (_player call EFUNC(common,uniqueItems))) && {[_player, objNull, ["notOnMap", "isNotSitting", "isNotInside"]] call EFUNC(common,canInteractWith)}
+                ([_player, "ACE_artilleryTable"] call EFUNC(common, hasItem)) && {[_player, objNull, ["notOnMap", "isNotSitting", "isNotInside"]] call EFUNC(common,canInteractWith)}
             };
             private _displayName = format ["%1%2", getText (_vehicleCfg >> "displayName"),["","*"] select _advCorrection];
             private _action = [format ['QGVAR(%1)',_index], _displayName, QPATHTOF(UI\icon_rangeTable.paa), _statement, _condition, {}, _info] call EFUNC(interact_menu,createAction);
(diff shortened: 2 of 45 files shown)


addons/disarming/functions/fnc_getAllGearUnit.sqf
```diff
@@ -18,36 +18,6 @@
...
-    if (_index != -1) then {
-        _classnamesCount set [_index, ((_classnamesCount select _index) + 1)];
-    } else {
-        _uniqueClassnames pushBack _x;
-        _classnamesCount pushBack 1;
-    };
-} forEach _allItems;
-
-[_uniqueClassnames, _classnamesCount]
+[keys _uniqueItems, values _uniqueItems]
```
mharis001, line 23: `toArray uniqueUnitItems [_target, 1, 2, 2, 2, true]`

addons/medical_treatment/functions/fnc_hasItem.sqf
```diff
@@ -21,11 +21,4 @@
...
 
-private _fnc_checkItems = {
-    params ["_unit"];
-
-    private _unitItems = _unit call EFUNC(common,uniqueItems);
-    _items findIf {_x in _unitItems} != -1
-};
-
-_medic call _fnc_checkItems || {GVAR(allowSharedEquipment) != 2 && {_patient call _fnc_checkItems}}
+[_medic, _items] call EFUNC(common,hasItem) || {GVAR(allowSharedEquipment) != 2 && {[_patient, _items] call EFUNC(common,hasItem)}}
```
mharis001, line 24: Not the same behaviour. `_fnc_checkItems` checked if the unit has one of the items (not all of them).

addons/repair/functions/fnc_hasItems.sqf
```diff
@@ -19,14 +19,4 @@
...
-    if ((_x isEqualType []) && {({[_unit, _x] call EFUNC(common,hasItem)} count _x == 0)}) exitWith {
-        _return = false;
-    };
-    if ((_x isEqualType "") && {!([_unit, _x] call EFUNC(common,hasItem))}) exitWith {
-        _return = false;
-    };
-} forEach _items;
-
-_return;
+[_unit, _items] call EFUNC(common, hasItem);
```
mharis001, line 22: No space after comma in `EFUNC`.

addons/arsenal/functions/fnc_onArsenalClose.sqf
```diff
@@ -70,6 +70,9 @@ if (!isNil QGVAR(moduleUsed)) then {
     objNull remoteControl GVAR(center);
 };
 
+// Update player unique items cache
+ACE_player call EFUNC(common,uniqueItems);
+
```
mharis001, line 75: Does the "loadout" player event handler not handle this?

addons/medical_treatment/functions/fnc_useItem.sqf
```diff
@@ -24,13 +24,11 @@ scopeName "Main";
 private _useOrder = [[_patient, _medic], [_medic, _patient], [_medic]] select GVAR(allowSharedEquipment);
 
 {
-    private _unit      = _x;
-    private _unitItems = _x call EFUNC(common,uniqueItems);
-
+    params ["_unit"];
```
mharis001, line 27: This will only check the medic now. `_x` is the unit from `_useOrder`.

mharis001 (changes requested):

addons/arsenal/functions/fnc_onArsenalClose.sqf
Salluci: Only clears the variable.

Salluci:
Should be all of them.

addons/arsenal/functions/fnc_onArsenalClose.sqf
mharis001: I don't see why arsenal needs to update the cache. It will automatically updated next time the getter function is called.

Salluci: Doesn't *need* to, this is just preemptively doing it since exiting the arsenal has a milisecond or two of downtime. I don't particularly care either way for it, can be deleted if needed.

addons/artillerytables/functions/fnc_interactMenuOpened.sqf
```diff
@@ -19,7 +19,7 @@ params ["_menuType"];
 TRACE_1("interactMenuOpened",_menuType);
 
 if (_menuType != 1) exitWith {};
-if (!("ACE_artilleryTable" in (ace_player call EFUNC(common,uniqueItems)))) exitWith {};
+if !([ace_player, "ACE_artilleryTable"] call EFUNC(common,hasItem)) exitWith {};
```
dedmen, line 22:
```suggestion
if !([ACE_player, "ACE_artilleryTable"] call EFUNC(common,hasItem)) exitWith {};
```
just for OCD

addons/common/functions/fnc_hasItem.sqf
```diff
@@ -16,6 +17,18 @@
...
+if (_item isEqualType "") then {
+    _item = [_item];
+};
+
+private _return = false;
+private _unitItems = _unit call FUNC(uniqueItems);
+
+{
+    _return = _x in _unitItems;
+    if ([_return, !_return] select _checkAll) then {break};
```
dedmen, line 31: It would be more performant to have one if condition and swap out the code passed to forEach.
Instead of evaluating every time inside the loop

Then you can also just use findIf for the !_checkAll case..
or actually can't you use findIf for both?

addons/common/functions/fnc_hasItem.sqf
```diff
@@ -16,6 +17,18 @@
  * Public: Yes
  */
 
-params [["_unit", objNull, [objNull]], ["_item", "", [""]]];
+params [["_unit", objNull, [objNull]], ["_item", "", ["", [""]]], ["_checkAll", false, [false]]];
 
-_item in (_unit call EFUNC(common,uniqueItems))
+if (_item isEqualType "") then {
+    _item = [_item];
```
dedmen, line 23: could improve performance by just using this as a shorthand to do a single check in _unitItems
instead of creating extra array here and going through the forEach and _checkAll handling below.

addons/common/functions/fnc_hasItem.sqf
Salluci: is `findIf` faster than `in`?

dedmen: No.
findIf is a forEach with the exitWith/break already built in.
You're not replacing the in, you're replacing the if, then, break

Salluci: Gotcha.

Salluci:
@PabstMirror 'Twas a mistake on my end. Can switch `in` condition to `(_hash getOrDefault` [_item, 0]) > 0`, or `!(isNil {_hash get _item})`, both should be faster due to hashmap's constant lookup time.

Salluci:
There's also the issue of `hasItem` being an API function, while both `in` and `get` are case-sensitive, but current `master` branch implementation doesn't seem to cover that, so won't in this unless requested.

PabstMirror:
`[player, "ACE_morphine"] call ace_common_fnc_hasItem;` 0.00655 ms

`[player, player] call ace_attach_fnc_getChildrenActions` 
0.0687 ms -> 0.115088 ms
has to do a lot forEach loops over things that aren't items

PR #8365, opened 2021-08-06
WIP - Arsenal - Optimizations & Improved CBA Accessory Support

Salluci:
- Remove excessive use of toLower everywhere.
- Optimize loadout verification macros to use in instead of findIf.
- Prioritize classname casing over baseWeapon casing in case of matching case-insensitive strings.
- Add better support for CBA Accessories.
- Add baseWeapon config property to ACE laser pointers.

Use of toLower on classnames can be troublesome for third-party arsenal hooks, potentially causing duplicate classnames in arrays and other complications. Since all functions used return case-sensitive classnames of items, best to get rid of it. Also faster.

Use of `in` instead of `findIf` greatly reduces stutter when opening loadouts tab on slower CPUs. (Personally went from a 4s freeze to less than a second).

Better support for CBA Accessories via checking attachments' baseWeapon config property when saving loadouts. This'll allow, for example, for a loadout saved with ACE_acc_pointer_green_IR attachment to still be loaded despite having that item having scope = 1, as the loadout will save its baseWeapon property instead.

### DIFF
--- a/addons/arsenal/functions/fnc_addRightPanelButton.sqf
+++ b/addons/arsenal/functions/fnc_addRightPanelButton.sqf
@@ -42,14 +42,14 @@ if (_position >= 0 && _position <= 9) then {
     private _cfgWeapons = configFile >> "CfgWeapons";
     _items = _items select {
         private _configItemInfo = _cfgWeapons >> _x >> "ItemInfo";
-           
+
         _x isKindOf ["CBA_MiscItem", _cfgWeapons] && {getNumber (_configItemInfo >> "type") in [TYPE_MUZZLE, TYPE_OPTICS, TYPE_FLASHLIGHT, TYPE_BIPOD]} ||
         {getNumber (_configItemInfo >> "type") in [TYPE_FIRST_AID_KIT, TYPE_MEDIKIT, TYPE_TOOLKIT]} ||
         {getText (_cfgWeapons >> _x >> "simulation") == "ItemMineDetector"}
     };
-    
+
     _return = _position;
-    GVAR(customRightPanelButtons) set [_position, [_items apply {toLower _x}, _picture, _tooltip]];
+    GVAR(customRightPanelButtons) set [_position, [_items, _picture, _tooltip]];
 };
 
 _return
--- a/addons/arsenal/functions/fnc_addRightPanelButton.sqf
+++ b/addons/arsenal/functions/fnc_addRightPanelButton.sqf
@@ -47,6 +47,7 @@ if (_position >= 0 && _position <= 9) then {
         {getNumber (_configItemInfo >> "type") in [TYPE_FIRST_AID_KIT, TYPE_MEDIKIT, TYPE_TOOLKIT]} ||
         {getText (_cfgWeapons >> _x >> "simulation") == "ItemMineDetector"}
     };
+    _items apply {configName (_cfgWeapons >> _x)};
 
     _return = _position;
     GVAR(customRightPanelButtons) set [_position, [_items, _picture, _tooltip]];
--- a/addons/arsenal/functions/fnc_attributeAddCompatible.sqf
+++ b/addons/arsenal/functions/fnc_attributeAddCompatible.sqf
@@ -51,7 +51,7 @@ if (_category == 8) then {
             _itemsToAdd append _magazines;
 
             {
-                _itemsToAdd append ([_magazineGroups, _x] call CBA_fnc_hashGet);
+                _itemsToAdd append (_magazineGroups get _x);
             } forEach getArray (_muzzleConfig >> "magazineWell");
         } forEach getArray (_weaponConfig >> "muzzles");
     } forEach _attributeWeapons;
--- a/addons/arsenal/functions/fnc_attributeAddCompatible.sqf
+++ b/addons/arsenal/functions/fnc_attributeAddCompatible.sqf
@@ -51,7 +51,7 @@ if (_category == 8) then {
             _itemsToAdd append _magazines;
 
             {
-                _itemsToAdd append ([_magazineGroups, toLower _x] call CBA_fnc_hashGet);
+                _itemsToAdd append ([_magazineGroups, _x] call CBA_fnc_hashGet);
             } forEach getArray (_muzzleConfig >> "magazineWell");
         } forEach getArray (_weaponConfig >> "muzzles");
     } forEach _attributeWeapons;
--- a/addons/arsenal/functions/fnc_attributeAddCompatible.sqf
+++ b/addons/arsenal/functions/fnc_attributeAddCompatible.sqf
@@ -38,6 +38,7 @@ private _itemsToAdd = [];
 if (_category == 8) then {
     private _magazineGroups = uiNamespace getVariable QGVAR(magazineGroups);
     private _cfgMagazines = configFile >> "CfgMagazines";
+    private _cfgMagazineWells = configFile >> "CfgMagazineWells";
 
     {
         private _weaponConfig = _cfgWeapons >> _x;
@@ -52,7 +53,7 @@ if (_category == 8) then {
 
             {
                 _itemsToAdd append (_magazineGroups get _x);
-            } forEach getArray ((_muzzleConfig >> "magazineWell") apply {configName (_cfgMagazines >> _x)});
+            } forEach getArray ((_muzzleConfig >> "magazineWell") apply {configName (_cfgMagazineWells >> _x)});
         } forEach getArray (_weaponConfig >> "muzzles");
     } forEach _attributeWeapons;
 } else {
--- a/addons/arsenal/functions/fnc_attributeAddCompatible.sqf
+++ b/addons/arsenal/functions/fnc_attributeAddCompatible.sqf
@@ -52,7 +52,7 @@ if (_category == 8) then {
 
             {
                 _itemsToAdd append (_magazineGroups get _x);
-            } forEach getArray (_muzzleConfig >> "magazineWell");
+            } forEach getArray ((_muzzleConfig >> "magazineWell") apply {configName (_cfgMagazines >> _x)});
         } forEach getArray (_weaponConfig >> "muzzles");
     } forEach _attributeWeapons;
 } else {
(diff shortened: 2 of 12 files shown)


addons/arsenal/functions/fnc_addRightPanelButton.sqf
```diff
@@ -42,14 +42,14 @@ if (_position >= 0 && _position <= 9) then {
...
+
         _x isKindOf ["CBA_MiscItem", _cfgWeapons] && {getNumber (_configItemInfo >> "type") in [TYPE_MUZZLE, TYPE_OPTICS, TYPE_FLASHLIGHT, TYPE_BIPOD]} ||
         {getNumber (_configItemInfo >> "type") in [TYPE_FIRST_AID_KIT, TYPE_MEDIKIT, TYPE_TOOLKIT]} ||
         {getText (_cfgWeapons >> _x >> "simulation") == "ItemMineDetector"}
     };
-    
+
     _return = _position;
-    GVAR(customRightPanelButtons) set [_position, [_items apply {toLower _x}, _picture, _tooltip]];
+    GVAR(customRightPanelButtons) set [_position, [_items, _picture, _tooltip]];
```
dedmen, line 52: This is a API function.
You cannot really rely on the input classnames being correct case.
You should convert them to config-case

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
```diff
@@ -51,7 +51,7 @@ if (_category == 8) then {
             _itemsToAdd append _magazines;
 
             {
-                _itemsToAdd append ([_magazineGroups, toLower _x] call CBA_fnc_hashGet);
+                _itemsToAdd append ([_magazineGroups, _x] call CBA_fnc_hashGet);
```
dedmen, line 54: You cannot rely on these names being config case correct.
If you make this case sensitive, the same case should be sued everywhere. Best way to do that is to make them config case like is already done for the magazines here
`_magazines = _magazines apply {configName (_cfgMagazines >> _x)};`

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -86,7 +90,7 @@ private _compatibleMagazines = [[[], []], [[], []], [[], []]];
             // Magazine groups
             {
                 private _magazineGroups = uiNamespace getVariable [QGVAR(magazineGroups),["#CBA_HASH#",[],[],[]]];
-                private _magArray = [_magazineGroups, toLower _x] call CBA_fnc_hashGet;
+                private _magArray = [_magazineGroups, _x] call CBA_fnc_hashGet;
```
dedmen, line 93: Again, you cannot rely on the name inside "magazineWell" to be correct case.

Also on a seperate note, it might make sense to convert this stuff to "real" hashmaps

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
```diff
@@ -56,7 +56,7 @@ switch (GVAR(currentLeftPanel)) do {
             if ((GVAR(currentItems) select 0) != _item && {_baseWeapon != _item}) then {
                 call _fnc_clearPreviousWepMags;
 
-                private _compatibleItems = (_item call bis_fnc_compatibleItems) apply {tolower _x};
+                private _compatibleItems = _item call bis_fnc_compatibleItems;
```
dedmen, line 59:
```suggestion
                private _compatibleItems = _item call BIS_fnc_compatibleItems;
```

you cannot rely on BIS_fnc_compatibleItems returning config case, should sanitize this input

addons/arsenal/functions/fnc_scanConfig.sqf
```diff
@@ -140,15 +141,6 @@ private _putList = [];
     private _className = configName _x;
 
     switch true do {
-        // Rifle, handgun, secondary weapons mags
-        case (
-                ((getNumber (_x >> "type") in [TYPE_MAGAZINE_PRIMARY_AND_THROW,TYPE_MAGAZINE_SECONDARY_AND_PUT,1536,TYPE_MAGAZINE_HANDGUN_AND_GL,TYPE_MAGAZINE_MISSILE]) ||
-                {(getNumber (_x >> QGVAR(hide))) == -1}) &&
-                {!(_className in _grenadeList)} &&
```
dedmen, line 147: Oh wow, nice find

dedmen:
Sorry should've been review instead of single comments, didn't expect to find more than one thing to note

Salluci:
If I'm not crazy, GVAR(virtualItems) is only available while the arsenal is open and is set to nil as soon as it's closed. Not sure if that's classified as API. In any case, setting this as WIP for now, then.

jonpas:
I think this is the only function that modifies `GVAR(virtualItems)` and is public API.
https://github.com/acemod/ACE3/blob/76676eee462cb0bbe400a482561c148d8652b550/addons/arsenal/functions/fnc_addVirtualItems.sqf

dedmen:
Yeah items needs to be sanitized then

For example here

https://github.com/acemod/ACE3/blob/76676eee462cb0bbe400a482561c148d8652b550/addons/arsenal/functions/fnc_addVirtualItems.sqf#L205

_x could be any case, and can be wrong.

Between here  https://github.com/acemod/ACE3/blob/76676eee462cb0bbe400a482561c148d8652b550/addons/arsenal/functions/fnc_addVirtualItems.sqf#L79-L80
you could insert a
`_x = configName (_configCfgWeapons >> _x)`

But it is still dangerous as you need to not miss any input, otherwise you'll get obscure hard to troubleshoot problems down the line.

I also don't know if our loadout saving/loading saved lowercase names previously, if it did you also need to sanitize these when loading a saved loadout.

Salluci:
> 
> 
> I think this is the only function that modifies `GVAR(virtualItems)` and is public API.
> https://github.com/acemod/ACE3/blob/76676eee462cb0bbe400a482561c148d8652b550/addons/arsenal/functions/fnc_addVirtualItems.sqf

That's the specific box's GVAR(virtualItems) variable. missionNamespace GVAR(virtualItems) is compiled on arsenal open [here](https://github.com/acemod/ACE3/blob/76676eee462cb0bbe400a482561c148d8652b550/addons/arsenal/functions/fnc_openBox.sqf#L46-L52) and modified [here](https://github.com/acemod/ACE3/blob/76676eee462cb0bbe400a482561c148d8652b550/addons/arsenal/functions/fnc_updateUniqueItemsList.sqf)

addons/arsenal/functions/fnc_scanConfig.sqf
```diff
@@ -157,6 +149,13 @@ private _putList = [];
         case (_className in _putList): {
             (_cargo select 16) pushBackUnique _className;
         };
+        // Rifle, handgun, secondary weapons mags
+        case (
+                (getNumber (_x >> "type") in [TYPE_MAGAZINE_PRIMARY_AND_THROW,TYPE_MAGAZINE_SECONDARY_AND_PUT,1536,TYPE_MAGAZINE_HANDGUN_AND_GL,TYPE_MAGAZINE_MISSILE]) ||
```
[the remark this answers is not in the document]
kymckay: Personally I'd like to see this array split across lines for readability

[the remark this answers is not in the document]
dedmen: #define MaskSlotPrimary     0x00000001  // primary weapons
#define MaskSlotSecondary   0x00000004  // secondary weapons
#define MaskSlotItem        0x00000F00  // items
#define MaskSlotBinocular   0x00003000  // binocular
#define MaskHardMounted     0x00010000  // hard mounted
#define MaskSlotHandGun     0x00000002  // hand gun
#define MaskSlotHandGunItem 0x000000F0  // hand gun magazines
#define MaskSlotInventory   0x001E0000  // hand gun magazines

1536 -> 0x600
number of used inventory slots = (type & MaskSlotItem) >> 8


so eh... Actually not really related to what kind of magazine it is at all. messy old stuff.

Looking at the code that handles this stuff I just want to get out, nothing is commented, a huge confusing mess and I have no idea what it does.
So maybe its better to keep these as "magic" numbers.
I suspect the code here was taken from arsenal

its a bitmask of all kinds of stuff.

#define TYPE_MAGAZINE_PRIMARY_AND_THROW 256
#define TYPE_MAGAZINE_SECONDARY_AND_PUT 512 // mainly
#define TYPE_MAGAZINE_MISSILE 768

Are actually not types, but number of inventory slot counts.
first one is 1 slot, second is 2 slots, third is 3 slots, 5136 is 6 slots.
Probably messy stuff from old A2 inventory

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
Salluci: Already done in fnc_scanConfig.

dedmen: No not done.
_x comes from `} forEach getArray (_muzzleConfig >> "magazineWell");`

Salluci:
That should cover everything mentioned so far. As for loadout saving, it's case-sensitive based on config case, so should be good and backwards compatible on that front.

dedmen:
This isn't that easy to test as you need to deliberately create magazine wells with wrong case magazines, weapons with wrong case magazines and magazine wells, scripts to create limited arsenal with wrong case classnames, 
Wrong classnames in ace_arsenal_fnc_addDefaultLoadout too, for ex https://github.com/WolfCorps/gf_scripts/blob/master/addons/gf_scripts/XEH_postInit.sqf#L241


I don't know if this needs any sanitation? 
If the loadouts are sanitized when they are displayed it wouldn't, but that'd be inefficient as thats what you want to make fast.

https://github.com/acemod/ACE3/blob/master/addons/arsenal/functions/fnc_addDefaultLoadout.sqf
You didn't touch that file, dunno if it needs to. No time to test that all

Salluci:
Adding virtual items with wrong case classnames: working
Adding default loadouts with wrong case classnames: working
Initializing limited arsenals with wrong case classnames: working
Weapons with wrong case magazines: working
Weapons with wrong case magazineWells: working
Magazine wells with wrong case magazines: working

Only thing that won't work is a profileNamespace loadout with a wrong case classname. Shouldn't ever happen normally since `ace_arsenal_saved_loadouts` isn't exposed to API and loadout data for saving is pulled from getUnitLoadout, which returns config case.

Salluci:
Closing this, found better approach after using my two remaining brain cells.

PR #8369, opened 2021-08-08
Weather - Add more exceptions for checking wind info

Salluci:
- Title.

Doesn't make a lot of sense not being to check wind while sitting on an open helicopter or swimming. Obstacles check still handles occlusion properly.

### DIFF
--- a/addons/weather/functions/fnc_displayWindInfo.sqf
+++ b/addons/weather/functions/fnc_displayWindInfo.sqf
@@ -35,7 +35,7 @@ TRACE_1("Starting Wind Info PFEH", GVAR(WindInfo));
     disableSerialization;
     params ["", "_pfID"];
 
-    if ((!GVAR(WindInfo)) || {!([ACE_player, ACE_player, []] call EFUNC(common,canInteractWith))}) exitWith {
+    if ((!GVAR(WindInfo)) || {!([ACE_player, ACE_player, ["notOnMap", "isNotInside", "isNotSitting", "isNotSwimming"]] call EFUNC(common,canInteractWith))}) exitWith {
         TRACE_1("Ending Wind Info PFEH", GVAR(WindInfo));
         GVAR(WindInfo) = false;
         (["RscWindIntuitive"] call BIS_fnc_rscLayer) cutText ["", "PLAIN"];
--- a/addons/weather/XEH_postInit.sqf
+++ b/addons/weather/XEH_postInit.sqf
@@ -4,7 +4,7 @@ GVAR(WindInfo) = false;
 ["ACE3 Common", QGVAR(WindInfoKey), localize LSTRING(WindInfoKeyToggle),
 {
     // Conditions: canInteract
-    if !([ACE_player, ACE_player, []] call EFUNC(common,canInteractWith)) exitWith {false};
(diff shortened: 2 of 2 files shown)


BaerMitUmlaut:
I wouldn't be opposed to adding bux' suggestion while freefalling... 😉

Salluci:
I'm not familiar enough with wind simulation in current version, is vertical speed a thing, or is it purely horizontal? [wind](https://community.bistudio.com/wiki/wind) command on BiWiki says third parameter of array is always 0.

Drofseh:
> Doesn't make a lot of sense not being to check wind while sitting on an open helicopter...

I'm a little confused about this comment, it sounds like you're saying players should be able to check wind while in a helicopter

Salluci:
> > Doesn't make a lot of sense not being to check wind while sitting on an open helicopter...
> 
> I'm a little confused about this comment, it sounds like you're saying players should be able to check wind while in a helicopter 

That was my original justification for tackling this, but as implemented players can measure wind inside all vehicles. Since speed calculation is based on eyePos, occlusion checking works correctly, so wind measured inside of a tank will just be 0.

From personal experience, you can definitely feel the wind while riding the average military helicopter, particularly on a hover.

addons/weather/functions/fnc_displayWindInfo.sqf
```diff
@@ -85,23 +90,26 @@ TRACE_1("Starting Wind Info PFEH", GVAR(WindInfo));
...
-
-    for "_index" from 0 to (_beaufortNumber - 1) do {
-        private _ball = __dsp ctrlCreate ["RscPicture", _index];
-        _ball ctrlSetText QPATHTOF(UI\wind_dot_ca.paa);
-        _ball ctrlSetTextColor [1,1,1,1];
-        private _ballCenterX = _centerX - (_ballWidth / 2) + ((sin _windDir) * 0.013333) * (_index - 4.9) + ((cos _windDir) * 0.0125);
-        private _ballCenterY = _centerY - (_ballHeight / 2) - ((1 * cos _windDir) * 4/3*0.013333) * (_index - 4.9) + ((sin _windDir) * 0.0125);
-        _ball ctrlSetPosition [_ballCenterX, _ballCenterY, _ballWidth, _ballHeight];
-        _ball ctrlCommit 0;
+    if (!_isUsingChute) then {
```
Dystopian, line 93: looks like `exitWith` could be used here for early exit instead of big `then` block

Dystopian (commented):
small comment

Drofseh:
> > > Doesn't make a lot of sense not being to check wind while sitting on an open helicopter...
> > 
> > 
> > I'm a little confused about this comment, it sounds like you're saying players should be able to check wind while in a helicopter
> 
> That was my original justification for tackling this, but as implemented players can measure wind inside all vehicles. Since speed calculation is based on eyePos, occlusion checking works correctly, so wind measured inside of a tank will just be 0.
> 
> From personal experience, you can definitely feel the wind while riding the average military helicopter, particularly on a hover.

You can still feel time wind direction despite the down draft from the rotors?

jonpas:
Moving to next milestone, seems like this warrants a bit of a discussion.

PR #8373, opened 2021-08-09
Medical Feedback - Stop heart beat effect on camera change

### LINKED ISSUE #8258: Medical - When in spectator screen, heart beating sound continues playing after pressing respawn
**Mods:**
- **Arma 3:** `2.04` profiling
- **CBA:** `3.15.3` stable
- **ACE3:** `3.13.6` stable

**Description:**
When pressing respawn whilst being unconscious and having the fast heart beat sound playing (so just before going into cardiac arrest), it keeps playing the fast heart beat sound until respawned.

**Steps to reproduce:**

1. Make a new mission, place down 2 playable units. Play in multiplayer.
2. Have player 1 shoot player 2 until unconscious.
3. When player 2 hears heartbeat when unconscious, press respawn.
4. Player 2 will continue to hear fast heart beating in the spectator camera until respawned.

**Expected behavior:**
Removal of heart beat when entering the spectator screen.

- Editor (Multiplayer) & Dedicated

**Log Files:**
- https://pastebin.com/5EUDkTTV

**Additional context:**
This happened on a newly created mission with ACE and CBA only and on a heavily modded server.

Salluci:
- Stop heart beat effect playing on camera change (zeus display, spectator, death).
- Close #8258.


kymckay:
This PR is not doing what the title claims, it's stopping on the medical death event.

I'm also not convinced this is the correct solution, it looks like the conditional logic should already be there to implicitly exit the function so that's probably where the bug is (whereas this just introduces an additional override GVAR).

severgun:
> I'm also not convinced this is the correct solution, it looks like the conditional logic should already be there to implicitly exit the function so that's probably where the bug is (whereas this just introduces an additional GVAR override).

Yes. Here explicit exit
https://github.com/acemod/ACE3/blob/e6a0772adb86d6d4d4ee038570326625e48ccdd3/addons/medical_feedback/functions/fnc_effectHeartBeat.sqf#L20-L23
Here HR sets to 0 on death.
https://github.com/acemod/ACE3/blob/e6a0772adb86d6d4d4ee038570326625e48ccdd3/addons/medical_status/functions/fnc_setDead.sqf#L20
But seems like pressing respawn did not handled well.

Salluci:
> 
> 
> > I'm also not convinced this is the correct solution, it looks like the conditional logic should already be there to implicitly exit the function so that's probably where the bug is (whereas this just introduces an additional GVAR override).
> 
> Yes. Here explicit exit
> 
> https://github.com/acemod/ACE3/blob/e6a0772adb86d6d4d4ee038570326625e48ccdd3/addons/medical_feedback/functions/fnc_effectHeartBeat.sqf#L20-L23
> 
> Here HR sets to 0 on death.
> 
> https://github.com/acemod/ACE3/blob/e6a0772adb86d6d4d4ee038570326625e48ccdd3/addons/medical_status/functions/fnc_setDead.sqf#L20
> 
> But seems like pressing respawn did not handled well.

From quick testing, respawn button bypasses setDead (output from the HandleKilled and HandleKilledMission functions show causeOfDeath as #scripted). Likely moving the variable setting from setDead on to the EH's would be enough.

addons/medical_status/functions/fnc_handleKilled.sqf
```diff
@@ -41,6 +41,14 @@ if (_causeOfDeath != "#scripted") then {
         _this set [1, _killer];
         _this set [2, _instigator];
     };
+} else { // Handle setDead being skipped (forced respawn)
+    
+    _unit setVariable [VAR_HEART_RATE, 0, true];
+    _unit setVariable [VAR_BLOOD_PRESS, [0, 0], true];
+
+    _unit setVariable [QEGVAR(medical,causeOfDeath), _reason, true];
```
severgun, line 49: `_reason` undefined?
Why not to call `fnc_setDead` with reason `#scripted`?

Salluci: _reason being undefined is a typo, fixing

handleKilledMission runs whenever any entity dies, including remote entities
handleKilled only runs on a local unit's death

Both the EH's will run when a local unit dies, but handleKilled will only run when a local unit dies, so it's preferable that it handle things.

addons/medical_status/functions/fnc_handleKilled.sqf
```diff
@@ -41,6 +41,14 @@ if (_causeOfDeath != "#scripted") then {
...
         _this set [2, _instigator];
     };
+} else { // Handle setDead being skipped (forced respawn)
+
+    _unit setVariable [VAR_HEART_RATE, 0, true];
+    _unit setVariable [VAR_BLOOD_PRESS, [0, 0], true];
+
+    _unit setVariable [QEGVAR(medical,causeOfDeath), _causeOfDeath, true];
+
+    [QEGVAR(medical,death), [_unit]] call CBA_fnc_localEvent;
```
kymckay, line 51: This solution seems reasonable, but it would make sense to me to just call the setDead function here instead of duplicating the code 👍 

I see that it will result in an unnecessary call to `EFUNC(medical_engine,setStructuralDamage)`, but think that should be harmless.

Salluci:
Last commit was necessary because there was still no handling for Zeus/Splendid Camera. Should be all good to go now.

addons/medical_feedback/functions/fnc_effectHeartBeat.sqf
```diff
@@ -17,6 +17,11 @@
 
 private _heartRate = GET_HEART_RATE(ACE_player);
 
+if (EGVAR(common,OldIsCamera)) exitWith {
+    TRACE_2("Ending heart beat effect - scripted camera",_heartRate,EGVAR(common,OldIsCamera));
+    GVAR(heartBeatEffectRunning) = false;
+};
+
```
kymckay, line 24:
```suggestion
if (EGVAR(common,OldIsCamera)) exitWith {
    TRACE_2("Ending heart beat effect - scripted camera",_heartRate,EGVAR(common,OldIsCamera));
    GVAR(heartBeatEffectRunning) = false;
};

private _heartRate = GET_HEART_RATE(ACE_player);
```

kymckay (changes requested):
Just a minor tweak since heart rate is not needed if this condition is true.

I'm not 100% about the use of `OldIsCamera`, it might be nicer to just directly use the `"featureCamera"` CBA player event somehow or to use `CBA_fnc_getActiveFeatureCamera`

kymckay (approved):
Happy to approve this as is functionally.

See previous regarding use of `CBA_fnc_getActiveFeatureCamera` for cleaner code if inclined to tidy that up.

Salluci:
OldIsCamera is set on camera view event by checking output of CBA_fnc_activeCamera against `""`. Since this function will likely run more often than the camera is changed, using the variable might be faster, if less readable. Plus, keeps it consistent with use of the variable in handleEffects.

kymckay:
Ah I see, yeah that seems like reasonable logic to me 👍

veteran29 (approved):

addons/medical_feedback/XEH_postInit.sqf
```diff
@@ -78,6 +78,27 @@ GVAR(bloodTickCounter) = 0;
...
+    if (_newCamera == "") then { // switched back to player view
+        private _status = IS_UNCONSCIOUS(_unit);
+        [!_status, _unit] call EFUNC(common,setVolume);
+
+        [QUOTE(ADDON), _volume, _status] call EFUNC(common,setHearingCapability);
+
+        ["unconscious", _status] call EFUNC(common,setDisableUserInputStatus);
+    } else {
+        [false, _unit] call EFUNC(common,setVolume);
+        [QUOTE(ADDON), _volume, false] call EFUNC(common,setHearingCapability);
```
kymckay, line 98: Why should we mute audio for players in curator cam or spectator?

Doesn't this go against the intention of line 63 above

Salluci: Shouldn't, misread the function params

PR #8375, opened 2021-08-09
Arsenal - Handle Unique Attachments

### LINKED ISSUE #8372: Ability to take weapon attachments that are not available in the arsenal
- **Arma 3:** `2.04`
- **CBA:** `3.15.5`
- **ACE3:** `3.13.6`

**Description:**
You can take any attachment for the weapon, even if none of it will be available in the arsenal.

**Steps to reproduce:**

1. Set up an arsenal that doesn't have, for example, thermal imaging scopes.
2. Install any scope that is not in your arsenal (for example, thermal imaging on your weapon).
3. Take from your arsenal absolutely any number of scopes that are initially inaccessible in the arsenal, but appear there if they are in your inventory or on your weapon

**Expected behavior:**
The ability to take weapon attachments that are not available in the arsenal should be disabled.

**Log Files:**
- I dont think, what it can help, but ok. https://pastebin.com/Sx42SNYm

**Screenshots:**
- No screenshots, but i got video about it. https://youtu.be/4tXvoAXKIhY

Salluci:
- Close #8372.

Forced attachments/magazines not in current box's virtual items to be handled as unique.

### DIFF
--- a/addons/arsenal/defines.hpp
+++ b/addons/arsenal/defines.hpp
@@ -460,3 +460,6 @@ if (!isNil QGVAR(customRightPanelButtons)) then {\
     _contentPanelCtrl lnbSetPicture [[_newRow, 7], getText (configFile >> "cfgVehicles" >> ((_loadout select 5) select 0) >> "picture")];\
     _contentPanelCtrl lnbSetPicture [[_newRow, 8], getText (configFile >> "cfgWeapons" >> (_loadout select 6) >> "picture")];\
     _contentPanelCtrl lnbSetPicture [[_newRow, 9], getText (configFile >> "cfgGlasses" >> (_loadout select 7) >> "picture")];
+
+#define ALL_BOX_VIRTUAL_ITEMS\
+    (flatten (GVAR(currentBox) getVariable [QGVAR(virtualItems),[]]))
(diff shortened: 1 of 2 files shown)


PabstMirror (approved):

addons/arsenal/defines.hpp
```diff
@@ -460,3 +460,6 @@ if (!isNil QGVAR(customRightPanelButtons)) then {\
     _contentPanelCtrl lnbSetPicture [[_newRow, 7], getText (configFile >> "cfgVehicles" >> ((_loadout select 5) select 0) >> "picture")];\
     _contentPanelCtrl lnbSetPicture [[_newRow, 8], getText (configFile >> "cfgWeapons" >> (_loadout select 6) >> "picture")];\
     _contentPanelCtrl lnbSetPicture [[_newRow, 9], getText (configFile >> "cfgGlasses" >> (_loadout select 7) >> "picture")];
+
+#define ALL_BOX_VIRTUAL_ITEMS\
+    (flatten (GVAR(currentBox) getVariable [QGVAR(virtualItems),[]]))
```
mharis001, line 465: How expensive is `flatten`? `_fnc_fill_right_Container` is called for every list item that is added which then uses `ALL_BOX_VIRTUAL_ITEMS` every time. Might be better to just `flatten` the virtual items once.

Salluci: Good point.

PR #8379, opened 2021-08-13
Common - Add replacement items to proper containers

### LINKED ISSUE #8361: ACE Medical Items replacement fills from uniform first
- **Arma 3:** `2.04`
- **CBA:** `3.15.5`
- **ACE3:** `3.13.6`

**Description:**
When picking up a first aid kit or medikit and putting it into your backpack or vest, converts into items initially filling your uniform.

**Steps to reproduce:**

1. Equip an empty uniform, vest, and backpack. Ensure they can hold items.
2. Put a medikit/first aid kit in the backpack/vest
3. Notice that it fills up the uniform first, instead of the vest/backpack

**Expected behavior:**
To fill the inventory container where the replaced item initially was first. So if you add a first aid kit to your backpack, the converted contents should be in the backpack first.

Salluci:
- Title.
- Close #8361.

Adding FirstAidKit or MedKit to a backpack will now place the replacement items inside backpack, etc. inside of always going for uniform first.

### DIFF
--- a/addons/common/functions/fnc_registerItemReplacement.sqf
+++ b/addons/common/functions/fnc_registerItemReplacement.sqf
@@ -28,15 +28,21 @@ private _fnc_replaceItems = {
     if (_items isEqualTo GVAR(oldItems)) exitWith {};
 
     private _newItems = _items - GVAR(oldItems);
+    private _newItems = _newItems arrayIntersect _newItems; // Get unique items only
     if (_newItems isEqualTo []) exitWith {
         GVAR(oldItems) = _items;
     };
 
-    _newItems sort true; // Sort so all items of current class can be replaced at once
     private _cfgWeapons = configFile >> "CfgWeapons"; // Microoptimization
 
     for "_i" from 0 to count _newItems - 1 do {
         private _item = _newItems#_i;
+
+        // Get count of item in each container
+        private _containerCount = [];
+        {
+            _containerCount pushBack ({_x == _item} count _x)
+        } forEach [uniformItems _unit, vestItems _unit, backpackItems _unit];
 
         // Determine replacement items: direct replacements, ...
          private _replacements = GVAR(itemReplacements) getVariable [_item, []];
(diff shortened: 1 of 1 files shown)


severgun:
As you already touch it may be you will fix 
I think this is misstype in `setVariable`. Instead of `_newItems` it should be `_oldReplacements`. Or I'm wrong?
https://github.com/acemod/ACE3/blob/aae10a6d06dd7b7f9ac02f29b6e510fbff6177a6/addons/common/functions/fnc_registerItemReplacement.sqf#L102-L104

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -28,16 +28,22 @@ private _fnc_replaceItems = {
     if (_items isEqualTo GVAR(oldItems)) exitWith {};
 
     private _newItems = _items - GVAR(oldItems);
+    private _newItems = _newItems arrayIntersect _newItems; // Get unique items only
```
severgun, line 31: double private and same name?

Salluci:
To be honest, I'm not quite sure on the itemReplacements setVariable change. Previous behaviour was to overwrite previous when a new replacement was set, with this PR it'll add on to the old replacement, which might be problematic. Would like to get some input from members on this.

severgun:
> Previous behaviour was to overwrite previous when a new replacement was set, with this PR it'll add on to the old replacement, which might be problematic.

Maybe.
But then `private _oldReplacements` never used and append did nothing useful.

Author @BaerMitUmlaut review required

PabstMirror:
We were doing the `append` so it seems like it was the intention to keep both items, and I think it's a good idea for max compatiblity.

example:
```
 ["ACE_RangeCard", "ACE_plasmaIV"] call ace_common_fnc_registerItemReplacement;
 ["ACE_RangeCard", "ACE_salineIV"] call ace_common_fnc_registerItemReplacement;

player addItem "ACE_RangeCard";
```
previously this would only give saline, with this PR it gives both

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -54,21 +60,18 @@ private _fnc_replaceItems = {
...
-            for "_j" from 1 to _count do {
-                { [_unit, _x] call FUNC(addToInventory) } forEach _replacements;
-            };
+            {
+                if (_x == 0) then {continue};
+                private _container = ["uniform", "vest", "backpack"] select _forEachIndex;
+                for "_j" from 1 to _x do {
+                    { [_unit, _x, _container] call FUNC(addToInventory) } forEach _replacements;
+                }
+            } forEach _containerCount;
```
PabstMirror, line 74: This would be a little bit safer in case we don't have room in the original container
```
{
    if (_x == 0) then {continue};
    private _container = ["uniform", "vest", "backpack"] select _forEachIndex;
    for "_j" from 1 to _x do {
        { 
            if ([_unit, _x, 1, _container == "uniform", _container == "vest", _container == "backpack"] call CBA_fnc_canAddItem) then {
                [_unit, _x, _container] call FUNC(addToInventory)  // add to specific container
            } else {
                [_unit, _x, ""] call FUNC(addToInventory) // no room, add anywhere
            }
        } forEach _replacements;
    }
} forEach _containerCount;
```
e.g.
```
for "_i" from 0 to 9 do { ["ACE_RangeCard", "ACE_salineIV"] call ace_common_fnc_registerItemReplacement; };
player addItemToUniform "ACE_RangeCard";
```

PR #8380, opened 2021-08-14
Add vehicle towing

TheCandianVendingMachine:
- Implements 2.06's `setTowParent`
- Adds the ability to attach a rope to a car/tank and tow a given `carx` vehicle
- Moves `ACE_ropeX` from `ace_fastrope` to `ace_common`
- Adds 2 shorter ropes

Requires ARMA version 2.06. Any vehicle with simulation `TankX` or `CarX` can tow any vehicle with simulation `CarX` - this is an engine limitation and I don't really have any way to tell the user that is the case: any suggestions?

In Action: https://youtu.be/hSGIu9_rrCg


severgun:
> any suggestions?

`Attach Tow Rope` action throw hint / CBA_fnc_notify that `vehicle can not be towed` if vehicle simulation is not `CarX`.

veteran29:
Damn, I wanted to write something very similar this weekend ;-D

Nice one!

dedmen:
Btw with 2.06 you can now also make custom rope models/textures.
Though I think the normal rope is appropriate here.

The problem that you have to use the brakes to "loosen" the brakes of the towed vehicle is weird ish.
I had to implement a special hacky way to start the vehicles physx simulation by force-rotating the front wheels.

I think the issue here might be that the rope actually stops and pulls back the heavy truck (in your video), so at the point where you affect force on the towed vehicle, you instantly stop.
The engine thing only works if the front vehicle is accelerating, not when its stopped.

One possible thing could be to use addForce or setVelocity to push the towed vehicle slightly.
But the big problem there is finding out when to do it and how hard to do it.


In my testing I always attached rope first and then setTowParent, not 100% sure if its relevant though
```
myRope = ropeCreate [f1_1, [0,-2,-1], f3_1, [0,1,-1], 10];
f3_1 setTowParent f1_1;
```

dedmen:
Actually, I just retested vehicle towing in our internal build, the brake loosening doesn't happen at all. 
Something is broken.

dedmen:
Okey I fixed my interpolation and added a second factor.
If the rope is stretched more than 110%, it will force turn the wheels to "unstick" the vehicle, even if delta speed is low.
That should also help a bit when you don't gain enough speed to create high enough wheel torque for the towed vehicle.

rautamiekka:
Does that mean tanks and others with similar mass can't be towed by a single vehicle anymore ?

dedmen:
> Does that mean tanks and others with similar mass can't be towed by a single vehicle anymore ?

I don't understand the question, what do you mean by "anymore"? This is a new feature

TheCandianVendingMachine:
> TankX can also be towed, but in my tests most ropes were insufficient for the mass and just ripped.

I just got my info from the Biki page. It explicitly says that TankX can't be towed at all

dedmen:
Wiki is outdated then

severgun:
`TankX can also be towed, but in my tests most ropes were insufficient for the mass and just ripped.`
What defines rope strength and can it be increased?

dedmen:
> What defines rope strength and can it be increased?

You can make custom ropes that stretch less. If a rope stretches too far, it breaks.

```
class CfgVehicles {
	
	class Rope;
	class RopeDedmen : Rope {
		maxRelLenght = 1.1;
		maxExtraLenght = 20;
		springFactor = 1;
		segmentType = "RopeSegmentDedmen";	
        torqueFactor = 0.5;
		dampingFactor[] = {10,10,10};
	}
}
```
maxRelLenght -> max stretch before break
maxExtraLenght -> max stretch before break
springFactor  -> how much it stretches

PabstMirror:
I see the note about players being in the vic in https://community.bistudio.com/wiki/setTowParent
but currently it locks the vehicle even if players are inside which isn't right

I think we should only allow towing fully empty vehicles
or we don't touch lock at all and hope for the best (I guess my choice)


```
Ref to nonnetwork object 1442495c100# 166329: sign_sphere10cm_f.p3d
Ref to nonnetwork object 1444db2b580# 166465: sign_sphere10cm_f.p3d
Client: Nonnetwork object 55ff32e0.
Client: Nonnetwork object 3ee425c0.
```
which is probably from "local" vehicles, probably not critical but maybe something to look at in the future

overall it seems to work fairly well

addons/towing/XEH_postInit.sqf
```diff
@@ -0,0 +1,23 @@
+#include "script_component.hpp"
+["MouseButtonDown", LINKFUNC(onMouseButtonDown)] call CBA_fnc_addDisplayHandler;
+["MouseButtonUp", LINKFUNC(onMouseButtonUp)] call CBA_fnc_addDisplayHandler;
+GVAR(mouseLeft) = false;
+GVAR(mouseRight) = false;
+
+GVAR(cancel) = false;
+GVAR(canAttach) = false;
+
+[QGVAR(lockVehicle), {
```
Dystopian, line 10: there are already events for it https://github.com/acemod/ACE3/blob/ee4431980146091fa816b7b9da87dde6e3ca3094/addons/common/XEH_postInit.sqf#L127-L134

Dystopian (commented):
I would suggest to move all ropes data to new `ropes` component to not blow `common` up

PabstMirror:
what about logistics?

Drofseh:
`STR_ACE_logistics_rope_descriptionShort` should be modified to indicate the ropes can be used for towing.

654wak654:
Not a merge blocker but some missing features I noticed while testing:
- Ability to remove tow rope from the towing vehicle's side.
- Some way to block vehicles from getting towed (could be based on vehicle lock).
- Documentation (Doesn't need much).
- Handle either vehicle being deleted. Looks like the engine already handles them getting destroyed, so this might be @dedmen  territory. The issue is that one vehicle being deleted leaves the rope & helper object behind. Both being deleted leaves just the helper object behind. Attempting to get your rope back using it causes the following error:
```
 6:47:49 Error in expression <unit", "_parent", "_child"];
;

_parent removeEventHandler ["RopeBreak", _parent>
 6:47:49   Error position: <removeEventHandler ["RopeBreak", _parent>
 6:47:49   Error Type Any, expected Number
 6:47:49 File z\ace\addons\towing\functions\fnc_detach.sqf..., line 21
```

TheCandianVendingMachine:
> I would merge this now, but that script error is no good, we should cleanup helper/rope objects on vehicle deletion.

probably fixed

TheCandianVendingMachine:
> Is it possible to make some vehicles tow "heavier" vehicles? It's currently not possible for the recovery vehicle Nemmera (built on the Merkava Mk3 chassis) to tow e.g. the AAF Leopard.

I fixed the deleting issue finally: was a stupid reason, I forgot that the event handlers would still exist after you detach.

It's on the engine to make vehicles be able to tow other vehicles - as far as I know it has to do with the weight of the vehicles, and manually adjusting weights via script seems like a good way to cause issues

PR #8384, opened 2021-08-16
UI - Infantry speed indictator

veteran29:
- title

ace_speed_indicator


addons/advanced_fatigue/script_component.hpp
```diff
@@ -14,6 +14,7 @@
     #define DEBUG_SETTINGS DEBUG_SETTINGS_ADVANCED_FATIGUE
 #endif
 
+#include "\a3\ui_f\hpp\defineCommonGrids.inc"
```
jonpas, line 17: Should be below macros include.

addons/advanced_fatigue/RscInGameUI.hpp
```diff
@@ -0,0 +1,18 @@
...
+class RscInGameUI {
+    class RscStanceInfo {
+        controls[] += {QGVAR(speedIndicator)};
+        class GVAR(speedIndicator): RscPictureKeepAspect {
+            onLoad = QUOTE(uiNamespace setVariable [ARR_2(QQGVAR(speedIndicator),_this select 0)]);
+            x = IGUI_GRID_STANCE_X + IGUI_GRID_STANCE_WAbs / 2 - SPEED_INDICATOR_W / 2;
+            y = IGUI_GRID_STANCE_Y + IGUI_GRID_STANCE_HAbs - SPEED_INDICATOR_Y_OFFSET);
+            w = SPEED_INDICATOR_W;
+            h = SPEED_INDICATOR_H;
+
```
jonpas, line 15:
```suggestion
```

jonpas (changes requested):

addons/advanced_fatigue/script_component.hpp
veteran29: Copied from `medical_feedback` ;P
https://github.com/acemod/ACE3/blob/master/addons/medical_feedback/script_component.hpp#L17

jonpas (approved):

addons/advanced_fatigue/RscInGameUI.hpp
```diff
@@ -0,0 +1,17 @@
...
+#define SPEED_INDICATOR_Y_OFFSET (IGUI_GRID_STANCE_HAbs / 4.25)
+
+class RscPictureKeepAspect;
+class RscInGameUI {
+    class RscStanceInfo {
+        controls[] += {QGVAR(speedIndicator)};
+        class GVAR(speedIndicator): RscPictureKeepAspect {
+            onLoad = QUOTE(uiNamespace setVariable [ARR_2(QQGVAR(speedIndicator),_this select 0)]);
+            x = IGUI_GRID_STANCE_X + IGUI_GRID_STANCE_WAbs / 2 - SPEED_INDICATOR_W / 2;
+            y = IGUI_GRID_STANCE_Y + IGUI_GRID_STANCE_HAbs - SPEED_INDICATOR_Y_OFFSET);
```
PabstMirror, line 12:
```suggestion
            y = IGUI_GRID_STANCE_Y + IGUI_GRID_STANCE_HAbs - SPEED_INDICATOR_Y_OFFSET;
```

Drofseh:
I would actually really like this not to be in Advanced Fatigue, a lot of units don't use AF and this is still valuable info for the regular stamina system

veteran29:
> I would actually really like this not to be in Advanced Fatigue, a lot of units don't use AF and this is still valuable info for the regular stamina system

I have no intention (and motivation) to separate this from AF, feel free to do so in a PR after this gets merged.

Drofseh:
Yah I really think this should be in `UI`, it's not even related to fatigue

BaerMitUmlaut:
I don't see any reason why this needs to be bound to AF, since it doesn't access any AF functionality.

veteran29:
My main idea was that you do not need careful stamina management in vanilla system and such indicator will be not useful/needed in there.

veteran29:
Found some time to move it to UI.

addons/ui/functions/fnc_handleSpeedIndicator.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * Handles visual changes of the speed indicator.
+ *
+ * Arguments:
+ * Current animation state <STRING>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * ["amovpercmtacslowwrfldf_ver2"] call ace_advanced_fatigue_fnc_handleSpeedIndicator
```
PabstMirror, line 13:
```suggestion
 * ["amovpercmtacslowwrfldf_ver2"] call ace_ui_fnc_handleSpeedIndicator
```

PabstMirror (approved):

jonpas (approved):

PR #8385, opened 2021-08-16
Medical - Add treatment status events

interruptinuse:
Adds
- `ace_medical_treatment_treatmentStartedLocal`
- `ace_medical_treatment_treatmentSucceededLocal`, and
- `ace_medical_treatment_treatmentFailedLocal`
CBA events. All of them have the same six parameters passed unchanged from the treatment action:
`_medic`, `_patient`, `_bodyPart`, `_classname`, `_itemUser`, `_usedItem`.

Rationale:
I've been trying to write a ShackTac-style medical extension without forking ACE. When trying to add a "PlayerName is bandaging you"/"You are being assisted" hint, I've been stopped by the following drawbacks in existing `ace_medical_treatment_{TREATMENT}Local` (e.g. `ace_medical_treatment_bandageLocal`) events:
1) They are not a single convenient entry point;
2) They do not carry a `_medic` parameter.

[Here's an example of someone else trying to do this kind of hint with the existing API](https://github.com/fparma/fparma-mods/blob/6ffd86e17be98b4474f12fac453c21b1a73d322d/addons/common/XEH_postInitClient.sqf#L47-L53). I think this PR would improve the situation.

Existing "ace_treatmentSucceded" event seems grandfathered in from before the medical rewrite, so I'm leaving it alone.

### DIFF
--- a/addons/medical_treatment/functions/fnc_treatment.sqf
+++ b/addons/medical_treatment/functions/fnc_treatment.sqf
@@ -147,6 +147,8 @@ if (_callbackProgress isEqualTo {}) then {
 
 [_medic, _patient, _bodyPart, _classname, _itemUser, _usedItem] call _callbackStart;
 
+["ace_treatmentStarted", [_medic, _patient, _bodyPart, _classname, _itemUser, _usedItem]] call CBA_fnc_localEvent;
+
 [
     _treatmentTime,
     [_medic, _patient, _bodyPart, _classname, _itemUser, _usedItem],
--- a/addons/medical_treatment/functions/fnc_treatmentFailure.sqf
+++ b/addons/medical_treatment/functions/fnc_treatmentFailure.sqf
@@ -48,3 +48,5 @@ if (!isNil QEGVAR(advanced_fatigue,setAnimExclusions)) then {
 GET_FUNCTION(_callbackFailure,configFile >> QGVAR(actions) >> _classname >> "callbackFailure");
 
 _args call _callbackFailure;
+
+["ace_treatmentFailed", [_medic, _patient, _bodyPart, _classname, _itemUser, _usedItem]] call CBA_fnc_localEvent;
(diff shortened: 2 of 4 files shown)


addons/medical_treatment/functions/fnc_treatment.sqf
```diff
@@ -147,6 +147,8 @@ if (_callbackProgress isEqualTo {}) then {
 
 [_medic, _patient, _bodyPart, _classname, _itemUser, _usedItem] call _callbackStart;
 
+[QGVAR(treatmentStartedLocal), [_medic, _patient, _bodyPart, _classname, _itemUser, _usedItem]] call CBA_fnc_localEvent;
```
BrettMayson, line 150: These don't need to have `Local` at the end of them, that is implied by `localEvent` and is documented.

It looks like these should also be simplified to `ace_treatmentStarted` rather than `ace_medical_treatment_treatmentStartedLocal`, but I'd let someone else confirm that

BrettMayson (changes requested):

addons/medical_treatment/functions/fnc_treatment.sqf
interruptinuse: Quite a lot of events in ACE are QGVAR. Are they supposed to be internal then?

interruptinuse: - Updated `ace_treatmentSucceded` with `_itemUser` and `_usedItem` parameters;
- Renamed other two events to `ace_treatmentStarted` and `ace_treatmentFailed`

kymckay: Yes I believe we decided at some point that QGVAR events for internal only and public events can just use `ace_`, probably not documented anywhere though

PabstMirror (approved):
lgtm

PR #8387, opened 2021-08-22
Medical - update setDead to use setHitpointDamage instigator param

Salluci:
- Title.
- Fixes scoreboard not counting kills etc, Slack discussion.

Forward compatible, but only has full effect with 2.06 profiling branch or 2.08.
~~Change to `enteredStateDeath` can be omitted if necessary as it can break current scoreboard functionality (not that it tends to work anyway).~~ Removed frame change to enteredStateDeath, just make it use instigator now.


addons/medical_status/functions/fnc_setDead.sqf
```diff
@@ -22,8 +23,17 @@ _unit setVariable [VAR_BLOOD_PRESS, [0, 0], true];
...
 // Send a local event before death
 [QEGVAR(medical,death), [_unit]] call CBA_fnc_localEvent;
 
 // Kill the unit without changing visual apperance
-[_unit, 1] call EFUNC(medical_engine,setStructuralDamage);
+private _prevDamage = _unit getHitPointDamage "hitHead";
+
+_unit setHitPointDamage ["hitHead", 1, true, _instigator];
+
+_unit setHitPointDamage ["hitHead", _prevDamage, true, _instigator];
```
kymckay, line 39: I'm not sure if the instigator parameter makes semantic sense for the second call here to reset the damage.

Unlikely to have meaningful effect, but for the sake of clear code intentions I'd personally leave it off.

addons/medical_statemachine/functions/fnc_enteredStateDeath.sqf
```diff
@@ -23,5 +23,5 @@ TRACE_3("enteredStateDeath",_this,_thisOrigin,_thisTransition);
 
 private _causeOfDeath = format ["%1:%2", _thisOrigin, _thisTransition];
 
-// could delay a frame here to fix the double killed EH, but we lose it being a "native" kill (scoreboard / rating)
-[_unit, _causeOfDeath] call EFUNC(medical_status,setDead);
+// delay a frame here to fix the double killed EH
+[{[_this select 0, _this select 1] call EFUNC(medical_status,setDead)}, [_unit, _causeOfDeath]] call CBA_fnc_execNextFrame;
```
kymckay, line 27: I'm not sure if this should be introduced as don't know if we ever determined possible side effects of this. 

There was also another fix for the double killed EH if I recall correctly.

Believe @PabstMirror would be the person to know best

addons/medical_treatment/functions/fnc_placeInBodyBag.sqf
```diff
@@ -31,7 +31,7 @@ if (!local _patient) exitWith {
 
 if (alive _patient) then {
     TRACE_1("Manually killing with setDead",_patient);
-    [_patient, "buried_alive"] call EFUNC(medical_status,setDead);
+    [_patient, "buried_alive", _medic] call EFUNC(medical_status,setDead);
```
kymckay, line 34: Any possible unwanted effects of having medics considered the killer of teammates?

Does this new command variation influence rating (could cause AI to consider them sideEnemy)? @dedmen

Salluci: It does affects rating, per Dedmen's comment in A3 discord, #community_wiki.

addons/medical_status/functions/fnc_setDead.sqf
Salluci: I'll just drop the latter two parameters in the second call.

addons/medical_statemachine/functions/fnc_enteredStateDeath.sqf
PabstMirror: >Fixed: "Killed" Event Handler was sometimes firing twice - [FT-T149510](https://feedback.bistudio.com/T149510)
I'm assuming this will make it into 2.06

there also is some CBA code that can probably be removed: https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_preInit.sqf#L66

either way I think we should wait for at least 2.06 to be out

Salluci:
Need stuff from #8373 for a fix. Rebasing

Salluci:
I can't Git.

PR #8388, opened 2021-08-23
Captives - Disallow ACRE/TFAR radio usage on surrender/handcuff

Salluci:
- Title.
- Supersedes #4334.
- Remove setting of master radio control variable from common_fnc_setVolume.

If it's still necessary to have setVolume control radio transmission allowance, I can edit that to call the status effect instead.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -26,6 +26,7 @@
 ["blockEngine", false, ["ACE_Refuel"]] call FUNC(statusEffect_addType);
 ["blockThrow", false, ["ACE_Attach", "ACE_concertina_wire", "ACE_dragging", "ACE_Explosives", "ACE_Ladder", "ACE_rearm", "ACE_refuel", "ACE_Sandbag", "ACE_Trenches", "ACE_tripod"]] call FUNC(statusEffect_addType);
 ["setHidden", true, ["ace_unconscious"]] call FUNC(statusEffect_addType);
+["blockRadio", false, [QEGVAR(captives,Handcuffed), QEGVAR(captives,Surrendered), "ace_unconscious"]] call FUNC(statusEffect_addType);
 
 [QGVAR(forceWalk), {
     params ["_object", "_set"];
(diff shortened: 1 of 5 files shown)


addons/medical_status/functions/fnc_setUnconsciousState.sqf
```diff
@@ -31,6 +31,9 @@ _unit setVariable [VAR_UNCON, _active, true];
 // Stop AI firing at unconscious units in most situations (global effect)
 [_unit, "setHidden", "ace_unconscious", _active] call EFUNC(common,statusEffect_set);
 
+// Block radio on unconsciousness for compatibility with captive module
+[_unit, "blockRadio", "ace_unconscious", _active] call EFUNC(common,statusEffect_set);
```
kymckay, line 35: This is the wrong place for this. See postInit of medical_feedback

kymckay (changes requested):

kymckay:
(otherwise looks good 👍 )

addons/medical_status/functions/fnc_setUnconsciousState.sqf
Salluci: Non-player units also need to have radio disabled, though, in case of RC/switchUnit

kymckay: Oh true, I'm falsely equivalating two different things (the setVolume which belongs in feedback and this status effect which does belong here in status). This looks good, thanks for correcting me 👍

kymckay (approved):

PabstMirror (approved):

PabstMirror:
Added a minor optimization to not broadcast the setVars if the radio addon isn't loaded
Based on code from fnc_endTransmission `if (isClass (configFile >> "CfgPatches" >> "task_force_radio")) then {`

addons/common/XEH_postInit.sqf
```diff
@@ -71,6 +72,20 @@
...
+    params ["_object", "_set"];
+    TRACE_2("blockRadio EH",_object,_set);
+    if (_object isEqualTo ACE_Player && {_set > 0}) then {
+        call FUNC(endRadioTransmission);
+    };
+    if (isClass (configFile >> "CfgPatches" >> "task_force_radio")) then {
+        _object setVariable ["tf_unable_to_use_radio", _set > 0, true];
+    };
+    if (isClass (configFile >> "CfgPatches" >> "acre_main")) then {
+        _object setVariable ["acre_sys_core_isDisabled", _set > 0, true];
```
PabstMirror, line 85: This blocks direct speaking, not just radios.
It's no fun for people handcuffed

654wak654: `ace_mouthgag` when?

jonpas: Are you sure that actually works? I could swear a handcuffed person was speaking fine this weekend on our session.

BrettMayson: I can confirm, I used `player setVariable ["acre_sys_core_isDisabled", true, true];` yesterday to prevent player's from speaking during parts of a Halloween mission

veteran29: Yeah, yesterday we had self surrendering players which were muted, the character mimic was set to talking but they were unable to be heard.

jonpas: Ah I see, radio block for captive is completely new with this PR. We could add another variable to ACRE2 that only disables radio. Or API functions, even better.

PR #8390, opened 2021-08-23
Medical - Fix #8389

### LINKED ISSUE #8389: Medical Dependency Tree Issue
Just reviewing some of the items from the draft changelog. I don't believe this PR should have been merged as is, medical engine cannot depend on a setting from medical damage due to the dependency tree of those components.

Either medical engine must now depend on medical damage (creating a loop in the dependency graph) or this setting should be moved to medical engine where it takes effect.

Fwiw, there was some reason this setting was removed in the rewrite, but I forget what it was. I think it may be because there's no way to consistently detect all vehicle crashes (as reflected in the code comments).

_Originally posted by @kymckay in https://github.com/acemod/ACE3/issues/8149#issuecomment-893415500_

Salluci:
- Title.
- Close #8389

### DIFF
--- a/addons/medical_engine/initSettings.sqf
+++ b/addons/medical_engine/initSettings.sqf
@@ -1,0 +1,8 @@
+[
+    QEGVAR(medical,enableVehicleCrashes),
+    "CHECKBOX",
+    [LSTRING(EnableVehicleCrashes_DisplayName), LSTRING(EnableVehicleCrashes_Description)],
+    ELSTRING(medical,Category),
+    true,
+    true
+] call CBA_fnc_addSetting;
--- a/addons/medical_engine/stringtable.xml
+++ b/addons/medical_engine/stringtable.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ACE">
+    <Package name="Medical_Damage">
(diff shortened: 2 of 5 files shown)


mharis001 (approved):

addons/medical_engine/initSettings.sqf
```diff
@@ -0,0 +1,8 @@
+[
+    QEGVAR(medical,enableVehicleCrashes),
+    "CHECKBOX",
+    [LSTRING(EnableVehicleCrashes_DisplayName), LSTRING(EnableVehicleCrashes_Description)],
```
PabstMirror, line 4: I guess we have to move the stringtables over as well

addons/medical_engine/stringtable.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project name="ACE">
+    <Package name="Medical_Damage">
```
PabstMirror, line 3:
```suggestion
    <Package name="Medical_Engine">
```

mharis001 (approved):
Forgot about the stringtable entries. LGTM now.

PabstMirror (approved):

PR #8394, opened 2021-08-25
Medical - Add Pain Unconscious Threshold setting

severgun:
- Add slider to set PAIN_UNCONSCIOUS which currently is just magic variable not even mentioned in wiki
- Add English and Russian stringtables


addons/medical_damage/initSettings.sqf
```diff
@@ -34,6 +34,15 @@
     true
 ] call CBA_fnc_addSetting;
 
+[
+    QEGVAR(medical,const_painUnconscious),
```
BrettMayson, line 38: Rename this? Seems kinda weird to have a setting with the prefix `const`

BrettMayson (changes requested):

addons/medical_damage/initSettings.sqf
severgun: what if someone use it?
Or using undocumented variables is "at one's own risk"?

addons/medical_engine/script_macros_medical.hpp
```diff
@@ -73,7 +73,7 @@
 #define DAMAGE_TOTAL_COLORS 10
 
 // --- pain
-#define PAIN_UNCONSCIOUS EGVAR(medical,const_painUnconscious)
+#define PAIN_UNCONSCIOUS EGVAR(medical,painUnconsciousThreshold)
 #define PAIN_UNCONSCIOUS_DEFAULT 0.5
```
severgun, line 77: I don't know if we need this after adding CBA setting.
https://github.com/acemod/ACE3/blob/039af66a8f8ef253870c61ae53d800d5a0f7cc13/addons/medical_engine/XEH_preInit.sqf#L17

addons/medical_damage/stringtable.xml
```diff
@@ -705,6 +705,14 @@
             <Spanish>La probabilidad de que una persona caiga inconsciente cuando su dolor está por encima del umbral al haber recibido daño.</Spanish>
             <French>La probabilité pour qu'une personne perde connaissance lorsque la douleur ressentie est supérieure à son seuil de tolérance.</French>
         </Key>
+        <Key ID="STR_ACE_Medical_Damage_PainUnconsciousThreshold_DisplayName">
+            <English>Pain Unconscious Threshold</English>
+            <Russian>Порог боли для потери сознания</Russian>
+        </Key>
+        <Key ID="STR_ACE_Medical_Damage_PainUnconsciousThreshold_Description">
+            <English>Sets the amount of pain from the received wound, at which a unit can lose consciousness.</English>
```
Salluci, line 713:
```suggestion
            <English>Sets the threshold for severe pain, above which unit can lose consciouness when receiving a wound.</English>
```

More readable IMO and lets user know what they're looking for when changing the setting.

Salluci:
UX would benefit from switch the hardcoded values to the variable here: https://github.com/acemod/ACE3/blob/master/addons/medical_gui/functions/fnc_updateInjuryList.sqf#L74-L88

severgun:
> UX would benefit from switching the hardcoded values to the variable here:

Pain always in 0-1 range. I don't see who and when need to change this localization switch.
Also this PR considered undesirable.

olimolly:
Since the february pain system redesign update had consequences on `const_painUnconscious`.

This request should be reconsider to adapt players preferences.

Effect of what was at 0.5 default value is more like 0.35 or ~0.4 now. Which I recommand as default value.
It would fix https://github.com/acemod/ACE3/issues/9027 issue in the same time.

LinkIsGrim:
I think this is valid to have as a setting. We can move it to a separate settings menu if it ends up being an issue, I've been meaning to categorize the medical settings independently anyway.

LinkIsGrim (approved):

addons/medical_damage/initSettings.sqf
```diff
@@ -34,6 +34,15 @@
     true
 ] call CBA_fnc_addSetting;
 
+[
+    QEGVAR(medical,painUnconsciousThreshold),
```
PabstMirror, line 38:
```suggestion
    QUOTE(PAIN_UNCONSCIOUS),
```
technically it could be this I don't really like it and would rather keep it as is

PabstMirror (approved):

PR #8397, opened 2021-08-28
Medical - Update `setDead` to use `setHitpointDamage` instigator param

Salluci:
- Title.
- Fixes scoreboard not counting kills etc, Slack discussion.
- Count forced respawn or script kill while unit is unconscious as kill for last instigator.

Forward compatible, but only has full effect with 2.06 profiling branch or 2.08.
Needs #8373.


PabstMirror:
I think we could capture respawn button press by modifying https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_disableUserInput.sqf#L69

something like
```
_ctrl ctrlSetEventHandler ["buttonClick", QUOTE(closeDialog 0; [player, "respawn_button"] call EFUNC(medical_status,setDead); [false] call DFUNC(disableUserInput);)];
```

PabstMirror:
> 2.06.148309 new PROFILING branch with PERFORMANCE binaries, v1, server and client, windows 32/64-bit, linux server 32/64-bit
> - Added: Instigator parameter to setHitPointDamage/setHit/setHitIndex script commands

I guess changes to the command are still being worked on

PabstMirror:
Did some testing and I think this seems to be working fine
I made some changes in commit https://github.com/acemod/ACE3/commit/4304701a1a3adb18efee67bab19a05f8632bc343

the main change is we no longer need to handle `if (_causeOfDeath != "#scripted" || {IS_UNCONSCIOUS(_unit)}) then {` because it's handled by the button directly

addons/common/functions/fnc_disableUserInput.sqf
```diff
@@ -66,7 +66,11 @@ if (_state) then {
                 _ctrl ctrlSetTooltip "Abort.";
 
                 _ctrl = _dlg displayctrl ([104, 1010] select isMultiplayer);
-                _ctrl ctrlSetEventHandler ["buttonClick", QUOTE(closeDialog 0; player setDamage 1; [false] call DFUNC(disableUserInput);)];
+if (["ace_medical"] call FUNC(isModLoaded)) then {
```
Vdauphin, line 69: indentation is missing here?

Salluci:
Is anything needed from me on this? I'm a bit busy the next month, not sure if I can make time to work on anything.

PabstMirror (approved):

PabstMirror:
I pushed those changes to this branch, so it should be good to go

jonpas (approved):

veteran29 (approved):

PR #8410, opened 2021-09-01
WIP - Ballistics - Cleanup magazines and magazineWells

### LINKED ISSUE #8355: Ballistics - Bad News ULR .300 WM inherit from 7.62 magazines
.300 WM magazines for the Bad News ULR inherit from 7.62x51 magazines. This causes innacurate round count (20, should be 10, capacity for .300 WM lines up with capacity for .338 LM) as well as having icons for 7.62x51 magazines instead of the icon used for the Bad News' magazines. Additionally, the "7.62" name for the magazine can be confusing, and should probably be changed for .300 WM in keeping with the .338 LM magazines.

**Solution you'd like:**
Change inheritance on classes and classnames. This might break existing loadouts, which is why I've not made a PR for it yet.

Changing "count" and "icon" properties manually, but this'd make the round count in the classname not line up with the round count in the magazine.

**Additional context:**
Affected classes: https://github.com/acemod/ACE3/blob/master/addons/ballistics/CfgMagazines.hpp#L379-L404

Salluci:
- Change base class of ACE 10Rnd 7.62x51 magazines to use Contact added 10Rnd Mk14 magazine with new model and icon.
- Switch GM6 Lynx magazine well from M107 to AS50. GM6 Lynx uses single stack magazines, M107 uses double stack.
- Remove 30Rnd MX 6.5 Creedmoor magazines from CBA 6.5C AR10 magazine well. MX magazines won't fit the average AR-10.
- Close #8355 by changing base class of ACE .300 WM magazines and adding new 10Rnd magazines.
- Add 10Rnd .50BMG magazines for the CBA_M107 magazine well.

TODO:

- [ ] Add color variants for MX magazines

- [ ] Remove 6.5 creedmoor magazines from MX series OR

- [ ] Add 6.5 creedmoor for Katiba/MSBS series

### DIFF
--- a/addons/ballistics/CfgMagazineWells.hpp
+++ b/addons/ballistics/CfgMagazineWells.hpp
@@ -155,15 +155,16 @@ class CfgMagazineWells {
             "ACE_10Rnd_338_API526_Mag",
             "ACE_20Rnd_762x67_Mk248_Mod_0_Mag",
             "ACE_20Rnd_762x67_Mk248_Mod_1_Mag",
-            "ACE_20Rnd_762x67_Berger_Hybrid_OTM_Mag"
+            "ACE_20Rnd_762x67_Berger_Hybrid_OTM_Mag",
+            "ACE_10Rnd_762x67_Mk248_Mod_0_Mag",
+            "ACE_10Rnd_762x67_Mk248_Mod_1_Mag",
+            "ACE_10Rnd_762x67_Berger_Hybrid_OTM_Mag"
         };
     };
 
     class CBA_65C_AR10 {
         ADDON[] = {
-            "ACE_30Rnd_65_Creedmor_mag",
             "ACE_20Rnd_65_Creedmor_mag",
-            "ACE_30Rnd_65x47_Scenar_mag",
             "ACE_20Rnd_65x47_Scenar_mag"
         };
     };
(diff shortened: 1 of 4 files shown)


Salluci:
I just now realized the BI magazines' strings have a space before the mm, ie "5.56 mm" vs ACE mags' "5.56mm". My OCD is going to kill me.

Salluci:
Fair, but alphabetical sorting 'breaks' if kept like current master branch.

Drofseh:
I uhh, hate to tell you this, but at least in English, there is also no standardization on putting a space before 'mm'.
It is both calibre and region dependant, but trends to no space.
I also support no space for space saving reasons.

see
https://saami.org/technical-information/cartridge-chamber-drawings/
https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Firearms/Cartridges

rautamiekka:
> Fair, but alphabetical sorting 'breaks' if kept like current master branch.

FWIW, IIRC the vanilla spelling is with a space, so you're right. Would be unreasonable amounts of work creating a "compat" for every weapon/ammo mod, too, the Universal Ammo System mod being a particular offender in that regard with so many gun and ammo entries.

> I uhh, hate to tell you this, but at least in English, there is also no standardization on putting a space before 'mm' for a calibre. It is both calibre and region dependant, but trends to no space. I also support no space for space saving reasons.

Yep, I don't care what the other languages (than English; maybe I would if Finnish was possible) do. Thanks for the support.

Salluci:
> I uhh, hate to tell you this, but at least in English, there is also no standardization on putting a space before 'mm' for a calibre.
It is both calibre and region dependant, but trends to no space.

I'm very much aware of that and hate the way BI did it, but making ACE follow the same pattern is cleaner than editing base game stringtables. Still, the commit can be reverted and was only done for the sake of keeping alphabetical sorting consistent.

jonpas:
Which was does BI do it? We should follow that.

I personally prefer space due to ISO standard for units, but bullets are all over the place IRL, so we should follow BI way to keep consistency.

PR #8414, opened 2021-09-01
Attach - Add child options

Slatery:
Reason - clean up the attach item menu to allow for multiple similar items, or the same item with different options, to be attached, without having to have multiple different items in a player's inventory, and not cluttering the interaction tree

- adds child actions to attaching items
- allow for multiple effects to be attached from a single item in inventory

### DIFF
--- a/addons/attach/CfgWeapons.hpp
+++ b/addons/attach/CfgWeapons.hpp
@@ -2,17 +2,92 @@ class CfgWeapons {
     class ACE_ItemCore;
     class CBA_MiscItem_ItemInfo;
 
-    class ACE_IR_Strobe_Item: ACE_ItemCore {
-        ACE_attachable = "ACE_IR_Strobe_Effect";
+    class ACE_IR_Strobe_Base: ACE_ItemCore {
         author = ECSTRING(common,ACETeam);
-        scope = 2;
+        scope = 1;
         displayName = CSTRING(IrStrobe_Name);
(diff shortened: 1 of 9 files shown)


jonpas:
Tabs.

Slatery:
This method of creating dummy classes was the lowest friction solution, as it allows an indefinite amount of children, without having to refactor a majority of the code. I'll update the PR comment to reflect this

addons/attach/CfgWeapons.hpp
```diff
@@ -2,17 +2,92 @@ class CfgWeapons {
...
+    class ACE_IR_Strobe_Item_Three_Medium : ACE_IR_Strobe_Base{
+        selectName = "IR Mode 3 - Medium";
+        ACE_attachable = "ACE_IR_Strobe_three_medium_Effect";
+    };
+    class ACE_IR_Strobe_Item_Three_Low : ACE_IR_Strobe_Base{
+        selectName = "IR Mode 3 - Low";
+        ACE_attachable = "ACE_IR_Strobe_three_low_Effect";
+    };
+    class ACE_IR_Strobe_Item_Four : ACE_IR_Strobe_Base{
+        selectName = "IR Mode 3";
```
Drofseh, line 77: should be mode 4

Slatery: have resolved

PR #8415, opened 2021-09-01
Merge ACEX

jonpas:
- **Merge ACEX into ACE3**
- Keep backwards compatibility with any items, settings and config entries.
  - Add `X` set of macros (`XADDON`, `(QQ)XGVAR`).
  - Add ACEX-prefixed CfgPatches.
- Convert ACEX component's ACE Settings into CBA Settings

Idea of ACEX was good, but it still today isn't large enough to really warrant it being its own mod, it always felt like it added more work (maintenance, updating dependencies, doing another release etc.) than it was worth it.

Instead, after some internal discussion, this PR is opened as a proof of concept to merge ACEX and ACE3. All ACEX components come with settings to disable them, and any future components that would be deemed "out of scope for ACE3" and would instead go into ACEX, will be able to go into ACE3 either in this similar way (disabled by default) or optional, which we distribute separately on Steam and in launchable `@mod` folders anyways.

ACEX doesn't have the benefits it used to have, especially with CBA Settings, the way we now handle optionals in ACE3 and reduced workforce as Arma 3 is less and less interesting for the lack of a better word.

---

The idea to merge ACEX was approved by the ACE-Team, but this PR now needs some love from the community, testing it on old missions, verifying backwards compatibility is there and generally reviewing it.

---

Left to do after initial approvals/tests:
- [ ] Documentation refactor
- [ ] News post


jonpas:
One internal worry that was brought up by @BaerMitUmlaut is a potential problem with the `X` family of macros. Other components referencing those variables could be easily mistaken.

I don't think it's a big issue, PRs should be tested anyways. But as an alternative idea, we could also change do `ace_var = acex_var;` on setting change (each time setting changes). That would just require the settings to retain ACEX-prefix for backwards compatibility, but all use of it could use the standard set of macros.

As for config entries, we could also just ignore backwards compatibility for them, I am not sure how many people used those in their own mods, assuming not many.

jonpas:
Note: Remove "Maintainers" part from `README.md`s as well.

jonpas:
> fyi [acemod/ACEX#258](https://github.com/acemod/ACEX/pull/258) was never properly reviewed to my knowledge.

And it was not merged either as I said above.

PR #8419, opened 2021-09-04
Medical - Keep tourniquets when PAKing

Salluci:
- Title.

Sets tourniquet application time to PAK time if a tourniquet is applied on limb instead of removing.


mharis001:
What is the reasoning behind this change? I believe PAK treatment is meant to be a "full reset".

PR #8422, opened 2021-09-04
Medical - Add multiple uses to consumable Surgical Kit

Salluci:
- Add magazine version of surgical kit that has its ammo decreased by one with every wound closed.
- Refactor surgicalKitProgress and getStitchTime to use actual wound count instead of count of wound arrays.

Has no influence on non-consumable Surgical Kit. Needs #8362 (and by consequence, v2.06) or more specifically the ability to check for magazines as required items for medical treatment.


Salluci:
Bloody git diff

PR #8424, opened 2021-09-04
Medical - Add multiple uses to consumable Surgical Kit

Salluci:

- Add magazine version of surgical kit that has its ammo decreased by one with every wound closed.
- Refactor surgicalKitProgress and getStitchTime to use actual wound count instead of count of wound arrays.
- Change stitching to close one wound instead of one wound conglomerate every GVAR(woundStitchTime) seconds.

Charge system has no influence on non-consumable Surgical Kit. Needs #8362 (and by consequence, v2.06) or more specifically the ability to check for magazines as required items for medical treatment.

New stitching will close one wound every GVAR(woundStitchTime) seconds. Previous behaviour was to close every set of same ID wounds on same limb instead, so if patient had 4x Medium Avulsions on their left leg, the 4 would be closed at once. Now, 1 is closed and array is only deleted when post-stitch amount would be negative or 0.

### DIFF
--- a/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
+++ b/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
@@ -279,7 +279,7 @@ class GVAR(actions) {
         displayNameProgress = CSTRING(Stitching);
         icon = QPATHTOEF(medical_gui,ui\surgical_kit.paa);
         category = "advanced";
-        items[] = {"ACE_surgicalKit"};
+        items[] = {"ACE_surgicalKit_mag"};
         treatmentLocations = QGVAR(locationSurgicalKit);
         allowSelfTreatment = QGVAR(allowSelfStitch);
         medicRequired = QGVAR(medicSurgicalKit);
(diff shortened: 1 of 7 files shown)


BrettMayson:
What about having a new item, like stitches, and having those be consumable rather than the surgical kit itself.

Salluci:
Players would only see the magazine version of surgical kits with this, but I see your point, though that would also create redudancy with the "Consume Surgical Kit" setting. Easier and faster codewise since consumeItem is already a thing.

Drofseh:
> require more input from user (having to interact or use medical menu for each wound closed)

The stitching loop could just remove one item each time a wound is stitched.

BrettMayson:
> 
> 
> > require more input from user (having to interact or use medical menu for each wound closed)
> 
> The stitching loop could just remove one item each time a wound is stitched.

Yes, this was my thinking too

BaerMitUmlaut:
Why not give players multiple surgical kits? I think this adds unnecessary complexity and I don't see any advantage.

rautamiekka:
Multiple surgical kits weigh a lot.

severgun:
Reducing item weight is always safe change.

Salluci:
See #9020.

PR #8428, opened 2021-09-05
Common - Add settings for additional progress bar information

Salluci:
- Add settings to show time left and % done in progress bar.


PabstMirror:
I know grad trenches did something funny with ace's progressBar, but as far as I can tell this should work fine with it.

Salluci:
Was for keeping things consistent with existing strings. Will adjust.

Issue #8356, opened 2021-08-05
Common - Look at improving reload mutex in 2.06


PabstMirror:
Follow up to #8142

new in 2.06 should allow an even better fix
we can fully eliminate false postives and handle odd things like `setWeaponReloadingTime`

https://community.bistudio.com/wiki/weaponState 
> magazineReloadPhase: Number - current magazine reload phase from 1 to 0, 0 - reload complete. >0 - reload in progress (Since Arma 3 v2.05.147855)

or

https://community.bistudio.com/wiki/gestureState
> Returns the name of a unit's current gesture

johnb432:
I don't know if you have already looked into this, but it might be able to also help.
https://community.bistudio.com/wiki/Arma_3:_Event_Handlers#Weapon_Config_Muzzle_Events

PR #8432, opened 2021-09-08
Common - Switch Reload Mutex to Gesture Events

Salluci:
- Title.

Requires gestureChanged & gestureDone support in CBA & Arma 2.06


Dystopian:
Maybe better to use `weaponState`'s `magazineReloadPhase` value instead of `isReloading` var. Also if we ever need `reload` event `magazineReloadPhase` can be used in CBA player EH for it. Tag #8356.

veteran29:
This breaks https://github.com/IDI-Systems/acre2/pull/1031 by removing `ace_common_isReloading `

jonpas:
Please don't force push changes, it makes it incredibly hard to review or update reviews based on latest changes.

jonpas:
> This breaks [IDI-Systems/acre2#1031](https://github.com/IDI-Systems/acre2/pull/1031) by removing `ace_common_isReloading `

We can always update ACRE2 as well if this gets removed. I don't think it was considered public API anyway?

Salluci:
Commands are now available in current Arma patch.

Salluci:
I haven't had issues reported on community testing with this, but I understand it's a bit sneakier with any bugs that might show up. Can I get another set of eyes on this?

PabstMirror:
Not working for me, not sure if something changed in 2.08?
Place "Heavy Gunner" in editor
`diag_log text format ["Gesture changed: %2 - %3", _unit, _gesture, (weaponState ACE_Player)];`
```
Gesture changed: gesturereloadmmg02 - ["MMG_02_sand_RCO_LP_F","MMG_02_sand_RCO_LP_F","manual","130Rnd_338_Mag",130,0,0]
Gesture changed: aidlpercmstpsraswrfldnon_g01_player - ["MMG_02_sand_RCO_LP_F","MMG_02_sand_RCO_LP_F","manual","130Rnd_338_Mag",130,0,0]
Gesture changed: gesturereloadpistol - ["hgun_P07_F","hgun_P07_F","Single","16Rnd_9x21_Mag",17,0,0]
```
Reloading big gun and pistol both fail to trigger the mutex, it seems like weaponState # 6 is 0 at time the gesture event runs

Salluci:
Back to WIP it is.

LinkIsGrim:
weaponState is not viable for this (unless we use a PFH and just do `GVAR(isReloading) = ((weaponState ACE_Player) select 6) > 0` constantly), so this is the best we can get if we want to drop the loadout event.

johnb432:
https://discord.com/channels/105462288051380224/108187245529268224/1244681061535973457

hypoxia125:
> > [discord.com/channels/105462288051380224/108187245529268224/1244681061535973457](https://discord.com/channels/105462288051380224/108187245529268224/1244681061535973457)
> 
> Thanks, I hate it.

I'm just here to make more problems lol

johnb432:
@LinkIsGrim Can you take a look at the changes I made?

johnb432:
> What happens if weapons are switched mid-reload to an empty weapon (is that possible?)

I removed the weapon during reload and it broke. I'll see if I can find a fix.

johnb432:
Fix found and pushed.

johnb432 (approved):
It would be best if someone else also approved this, given the changes I've made.

PabstMirror:
```
[] spawn {
player reload [];
sleep 0.01;
systemChat str ace_common_isReloading;
"ace_gestures_Forward" call ace_gestures_fnc_playSignal;
};
```
prints false and magazine disappears
and `ace_common_isReloading` is locked to true (I think line 511 never completes)
the new way seems to be 1-2 frames behind the old way
but the simplification may still be worth it, it's pretty hard to time a reload and gesture keybind at the same time

also both old and new fail on
```
player reload [];
"ace_gestures_Forward" call ace_gestures_fnc_playSignal;
```
happening in same frame

johnb432:
> ```sqf
> player reload [];
> "ace_gestures_Forward" call ace_gestures_fnc_playSignal;
> ```

`magazineReloadPhase` from `weaponState` returns `>0` when run the code, oof. I'll need to think what should be done (maybe have a variable to check if `magazineReloadPhase` decreases over time?).

johnb432:
Managed to find a fix for gestures breaking the reload mutex.

johnb432 (approved):
Just added the same check for the reload phase getting stuck when units are switched.

PR #8433, opened 2021-09-09
Medical - Add dropping weapons on unconsciousness

AndreasBrostrom:
- Added setting to allow a change of dropping held weapon when going unconscious (Default 0% chance / off)
- Moved `throwWeapon` function from `hitreactions` to `common`.

### DIFF
--- a/addons/fire/functions/fnc_burnReaction.sqf
+++ b/addons/fire/functions/fnc_burnReaction.sqf
@@ -21,7 +21,7 @@ if (
     && {_unit in _unit && { !(currentWeapon _unit isEqualTo "") }}
     && {!isPlayer _unit || GVAR(dropWeapon >= 2)}
 ) then {
-    [_unit] call EFUNC(hitreactions,throwWeapon);
+    [_unit] call EFUNC(common,throwWeapon);
 };
 
 if (_unit isKindOf "CAManBase") then {
--- a/addons/medical_status/functions/fnc_setUnconsciousState.sqf
+++ b/addons/medical_status/functions/fnc_setUnconsciousState.sqf
@@ -42,6 +42,11 @@ if (_active) then {
         _unit setVariable [QEGVAR(medical,lastWakeUpCheck), _lastWakeUpCheck max CBA_missionTime];
     };
 
+    // Player drop weapon chanse
+    if (EGVAR(medical,dropWeaponUnconsciousChanse) != 0 && {_unit == ACE_player && {random 1 <= EGVAR(medical,dropWeaponUnconsciousChanse)}}) then {
+        _unit call EFUNC(hitreactions,throwWeapon);
+    };
+
     if (_unit == ACE_player) then {
         if (visibleMap) then {openMap false};
 
--- a/addons/medical_status/functions/fnc_setUnconsciousState.sqf
+++ b/addons/medical_status/functions/fnc_setUnconsciousState.sqf
@@ -44,7 +44,7 @@ if (_active) then {
 
     // Player drop weapon chanse
     if (EGVAR(medical,dropWeaponUnconsciousChanse) != 0 && {_unit == ACE_player && {random 1 <= EGVAR(medical,dropWeaponUnconsciousChanse)}}) then {
-        _unit call EFUNC(hitreactions,throwWeapon);
+        _unit call EFUNC(common,throwWeapon);
     };
 
     if (_unit == ACE_player) then {
--- a/addons/medical_status/functions/fnc_setUnconsciousState.sqf
+++ b/addons/medical_status/functions/fnc_setUnconsciousState.sqf
@@ -42,7 +42,7 @@ if (_active) then {
         _unit setVariable [QEGVAR(medical,lastWakeUpCheck), _lastWakeUpCheck max CBA_missionTime];
     };
 
-    // Player drop weapon channe
+    // Player drop weapon chance
     if (EGVAR(medical,dropWeaponUnconsciousChance) != 0 && {_unit == ACE_player && {random 1 <= EGVAR(medical,dropWeaponUnconsciousChance)}}) then {
         _unit call EFUNC(common,throwWeapon);
     };
--- a/addons/medical_status/functions/fnc_setUnconsciousState.sqf
+++ b/addons/medical_status/functions/fnc_setUnconsciousState.sqf
@@ -42,8 +42,8 @@ if (_active) then {
         _unit setVariable [QEGVAR(medical,lastWakeUpCheck), _lastWakeUpCheck max CBA_missionTime];
     };
 
-    // Player drop weapon chanse
-    if (EGVAR(medical,dropWeaponUnconsciousChanse) != 0 && {_unit == ACE_player && {random 1 <= EGVAR(medical,dropWeaponUnconsciousChanse)}}) then {
+    // Player drop weapon channe
+    if (EGVAR(medical,dropWeaponUnconsciousChance) != 0 && {_unit == ACE_player && {random 1 <= EGVAR(medical,dropWeaponUnconsciousChance)}}) then {
         _unit call EFUNC(common,throwWeapon);
     };
 
--- a/addons/medical_status/functions/fnc_setUnconsciousState.sqf
+++ b/addons/medical_status/functions/fnc_setUnconsciousState.sqf
@@ -42,16 +42,15 @@ if (_active) then {
         _unit setVariable [QEGVAR(medical,lastWakeUpCheck), _lastWakeUpCheck max CBA_missionTime];
     };
 
-    // Chance for player to drop weapon
-    if (EGVAR(medical,dropWeaponUnconsciousChance) != 0 && {_unit == ACE_player && {random 1 <= EGVAR(medical,dropWeaponUnconsciousChance)}}) then {
-        _unit call EFUNC(common,throwWeapon);
-    };
-
     if (_unit == ACE_player) then {
         if (visibleMap) then {openMap false};
 
         while {dialog} do {
             closeDialog 0;
+        };
+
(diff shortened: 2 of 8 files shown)


BrettMayson:
Do we also want a slider to set the chance of this happening?

addons/medical/functions/fnc_setUnconscious.sqf
```diff
@@ -46,6 +46,14 @@ if (currentWeapon _unit != primaryWeapon _unit) then {
     _unit selectWeapon primaryWeapon _unit;
 };
 
+if (QGVAR(dropWeaponOnUnconscious)) then {
+    if (_unit == ACE_player) then {
+        if (_knockOut) then {
+            player call EFUNC(hitreactions,throwWeapon);
```
commy2, line 52: Why should the Zeus throw their weapon when the remote controlled AI gets shot?

ColdEvul: Hehe, they should not.

Assuming that this should be suffisant.
```suggestion
            ACE_player call EFUNC(hitreactions,throwWeapon);
```
 also dont want ai to drop their weapons cause ai

ColdEvul:
> Do we also want a slider to set the chance of this happening?

Sure i can hammer one. The setting is off per default.

kymckay:
>Do we also want a slider to set the chance of this happening?

I'd take out the "also", boolean + slider should almost always just be a single slider setting that captures the "true" and "false" states anyway

kymckay:
I'd also point out that this PR introduces a dependency on hitreactions which is not reflected in the config

I'd also point out this is only the public function, it is not called from the statemachine and as such this behaviour won't be seen unless calling the public function, the setting doesn't really belong in this component which houses public APIs. It will need to effect the underlying unconscious state code.

ColdEvul:
> I'd also point out this is only the public function, it is not called from the statemachine and as such this behaviour won't be seen unless calling the public function, the setting doesn't really belong in this component which houses public APIs. It will need to effect the underlying unconscious state code.

were is this defined?

Salluci:
Logic for this needs to be moved/added to [addons/medical_status/functions/fnc_setUnconsciousState.sqf](https://github.com/acemod/ACE3/blob/master/addons/medical_status/functions/fnc_setUnconsciousState.sqf) so it's used internally and actually has an effect on default gameplay.

ColdEvul:
Will 

> Logic for this needs to be moved/added to [addons/medical_status/functions/fnc_setUnconsciousState.sqf](https://github.com/acemod/ACE3/blob/master/addons/medical_status/functions/fnc_setUnconsciousState.sqf) so it's used internally and actually has an effect on default gameplay.

Will adjust

addons/medical/functions/fnc_setUnconscious.sqf
```diff
@@ -46,6 +46,16 @@ if (currentWeapon _unit != primaryWeapon _unit) then {
...
 
+if (GVAR(dropWeaponOnUnconscious) != 0) then {
+    if (_unit == ACE_player) then {
+        if (_knockOut) then {
+            if (random 1 <= GVAR(dropWeaponOnUnconscious)) then {
+                ACE_player call EFUNC(hitreactions,throwWeapon);
+            };
+        };
+    };
+};
```
severgun, line 57: Why not to use &&? And there is already `if (_knockOut)` couple lines lower. So why not to put it inside?
```sqf
if (GVAR(dropWeaponOnUnconscious) != 0 && {_unit == ACE_player && {random 1 <= GVAR(dropWeaponOnUnconscious)}}) then {
    ACE_player call EFUNC(hitreactions,throwWeapon);
};
```

addons/medical/stringtable.xml
```diff
@@ -18,6 +18,12 @@
             <Turkish>ACE Medikal</Turkish>
             <Japanese>ACE 医療</Japanese>
         </Key>
+        <Key ID="STR_ACE_Medical_EnableDropWeaponOnUnconscious_DisplayName">
+            <English>Drop weapon on unconscious</English>
```
severgun, line 22: Should be Title Case and mention that this is **chance**

addons/medical/functions/fnc_setUnconscious.sqf
AndreasBrostrom: agree way better will adjust.

addons/medical/stringtable.xml
commy2: Drop Weapon When Falling Unconscious

Drop Weapon When Losing Consciousness

🤔

Drofseh: Drop Weapon On Loss of Consciousness

AndreasBrostrom:
ToDo
- [ ] Change name of setting display name
- [ ] Rewrite and move to: addons/medical_status/functions/fnc_setUnconsciousState.sqf
- [ ] Test

Salluci:
I don't like the idea of making `medical_status` dependent on `hitreactions`.

jonpas:
Please don't force push, it is impossible to follow your changes.

AndreasBrostrom:
> I don't like the idea of making `medical_status` dependent on `hitreactions`. Rather move `fnc_throwWeapon` to `common` component. Would be more useful there for this and other modules (throwing weapons when surrendering, for example).
> 
> Variable name/stringtable spelling is wrong. "Chance", not "Chanse".

I'll make the adjustment

> Please don't force push, it is impossible to follow your changes.

Did due to rebase, retore and squashing wip commits. It was eather that or a new PR.

addons/medical_status/stringtable.xml
```diff
@@ -111,5 +111,11 @@
             <Spanish>Controla la rapidez con que fluye el líquido de las bolsas intravenosas. El cambio de volumen de la bolsa IV se calcula como: \n intervalo (s) de tiempo * cambio iv por segundo (4.1667 mL/s) * velocidad de flujo (este coeficiente).</Spanish>
             <Turkish>IV Torbalardan sıvının ne kadar hızlı aktığını kontrol eder. IV Torba hacim değişikliği şu şekilde hesaplanır: zaman aralıkları iv saniye başına değişim (4.1667 mL / s) akış hızı (bu katsayı).</Turkish>
         </Key>
+        <Key ID="STR_ACE_Medical_Status_DropWeaponUnconsciousChanse_DisplayName">
+            <English>Chance to Drop Weapon On Loss of Consciousness</English>
+        </Key>
+        <Key ID="STR_ACE_Medical_Status_DropWeaponUnconsciousChanse_Description">
+            <English>The probability of a person of dropping their weapon on loss of concussion.</English>
```
rautamiekka, line 118: A thinking error there: `concussion` -> `consciousness`.

rautamiekka (changes requested):

addons/medical_status/stringtable.xml
AndreasBrostrom: was not conscious

AndreasBrostrom:
Updated header

LinkIsGrim (changes requested):
Need to update function call in https://github.com/acemod/ACE3/blob/49f4ea23e70dd4784b28248f605b14105f4a9a24/addons/fire/functions/fnc_burnReaction.sqf#L24

addons/medical_status/functions/fnc_setUnconsciousState.sqf
```diff
@@ -42,6 +42,11 @@ if (_active) then {
         _unit setVariable [QEGVAR(medical,lastWakeUpCheck), _lastWakeUpCheck max CBA_missionTime];
     };
 
+    // Player drop weapon chance
```
LinkIsGrim, line 45:
```suggestion
    // Chance for player to drop weapon
```

addons/medical_status/stringtable.xml
```diff
@@ -111,5 +111,11 @@
             <Spanish>Controla la rapidez con que fluye el líquido de las bolsas intravenosas. El cambio de volumen de la bolsa IV se calcula como: \n intervalo (s) de tiempo * cambio iv por segundo (4.1667 mL/s) * velocidad de flujo (este coeficiente).</Spanish>
             <Turkish>IV Torbalardan sıvının ne kadar hızlı aktığını kontrol eder. IV Torba hacim değişikliği şu şekilde hesaplanır: zaman aralıkları iv saniye başına değişim (4.1667 mL / s) akış hızı (bu katsayı).</Turkish>
         </Key>
+        <Key ID="STR_ACE_Medical_Status_DropWeaponUnconsciousChance_DisplayName">
+            <English>Unconscious Drop Weapon</English>
```
LinkIsGrim, line 115:
```suggestion
            <English>Unconsciousness Weapon Drop Chance</English>
```

addons/medical_status/stringtable.xml
```diff
@@ -111,5 +111,11 @@
...
+        </Key>
+        <Key ID="STR_ACE_Medical_Status_DropWeaponUnconsciousChance_Description">
+            <English>The probability of a player dropping their weapon on loss of consciousness.</English>
```
LinkIsGrim, line 118:
```suggestion
            <English>Chance for a player to drop their weapon when going unconscious.\nHas no effect on AI.</English>
```

LinkIsGrim (commented):
Either that or a better setting title. Otherwise it's good.

LinkIsGrim (approved):

addons/fire/functions/fnc_burnReaction.sqf
```diff
@@ -21,7 +21,7 @@ if (
     && {_unit in _unit && { !(currentWeapon _unit isEqualTo "") }}
     && {!isPlayer _unit || GVAR(dropWeapon >= 2)}
```
PabstMirror, line 22: it was fine but makes me very uncomfortable
```suggestion
    && {!isPlayer _unit || GVAR(dropWeapon) >= 2}
```

LinkIsGrim: that... shouldn't work?

PR #8434, opened 2021-09-11
Arsenal - Add stat to show how AI will use ammo

PabstMirror:
image

Mainly so I can find rockets that AI will actually shoot at infantry 
but it's not a real stat, so I don't know if it really makes to show
maybe just show in editor but not in-game?


veteran29:
> maybe just show in editor but not in-game?

I think it makes sense to show this in editor or when current player is a zeus.
```sqf
condition = QUOTE(is3DEN || !isNull getAssignedCuratorLogic player);
```

veteran29:
Should this still be a draft?

Drofseh:
Could we add a getVariable to the condition?
That way a misisonmaker could enable it as well, say for the ace arsenal mission in the main menu?

veteran29 (approved):

LinkIsGrim (approved):

PR #8437, opened 2021-09-12
Cargo - Fix can't load `Slingload_01_Base_F`

Vdauphin:
20ft containers (`Land_Cargo20_military_green_F`) can be loaded. But all children from `Slingload_01_Base_F` which share the same real size of a 20ft container can't be loaded (ex: `B_Slingload_01_Ammo_F`)

- This pull request enable the loading of  `Slingload_01_Base_F` childrens. It follow the current behavior of all 20 ft containers (https://github.com/acemod/ACE3/blob/e6a0772adb86d6d4d4ee038570326625e48ccdd3/addons/cargo/CfgVehicles.hpp#L534-L647)

### DIFF
--- a/addons/cargo/CfgVehicles.hpp
+++ b/addons/cargo/CfgVehicles.hpp
@@ -356,8 +356,8 @@ class CfgVehicles {
 
      //Huron 20ft containers
     class Slingload_01_Base_F: Slingload_base_F {
-        GVAR(canLoad) = 0;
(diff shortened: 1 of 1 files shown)


jonpas:
Can you add a comment next to the changed line indicating why it has to be like that (the point of this PR) for future source reference?

PabstMirror (approved):

PR #8444, opened 2021-09-17
Medical GUI - Add setting to show Damage

LinkIsGrim:
- Add Setting & Injury List entry for displaying qualitative % of Damage Threshold taken on a given body part.
- Change hardcoded pain tooltip values in updateInjuryList to painUnconsciousThreshold variable.
- Take Damage Threshold into account when displaying body part damage color.

Screenshot:


BrettMayson:
Setting for this? I personally would not want it

addons/medical_gui/initSettings.sqf
```diff
@@ -51,3 +51,12 @@
     [[0, 1, 2], [ELSTRING(common,Disabled), ELSTRING(Medical_Treatment,Anyone), ELSTRING(Medical_Treatment,Medics)], 1],
     false
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(showDamageToolTip),
```
BrettMayson, line 56: Also nickpicky, but this isn't really a tooltip is it? Should it be entry instead?

BrettMayson (changes requested):

addons/medical_gui/initSettings.sqf
Salluci: sure, why the hell not

BrettMayson (approved):
Seems fine to me, fractures do seem to send anything to 100% damage.

addons/medical_gui/functions/fnc_updateInjuryList.sqf
```diff
@@ -31,7 +31,26 @@ private _bodyPartName = [
...
+switch (true) do {
+    case (_selectionN > 3): {
+        _damageThreshold = LIMPING_DAMAGE_THRESHOLD * 4;
+    };
+    case (_selectionN > 1): {
+        _damageThreshold = FRACTURE_DAMAGE_THRESHOLD * 4;
+    };
+    case (_selectionN == 0): {
+        _damageThreshold = _damageThreshold / 2;
+    };
```
veteran29: IMO each `case` could use a comment to which body parts it applies, easier to follow.

addons/medical_gui/functions/fnc_updateInjuryList.sqf
```diff
@@ -31,8 +31,47 @@ private _bodyPartName = [
     LSTRING(RightLeg)
 ] select _selectionN;
 
+
+
```
BrettMayson, line 35: ?

BrettMayson (approved):

addons/medical_gui/functions/fnc_updateInjuryList.sqf
LinkIsGrim: Typos?

Rythoka:
Just an opinion, but the terminology used here isn't great and I don't think it would be immediately obvious to most people what the difference between "major" and "severe" is. Also, "chronic" has nothing to do with severity, it just means that a conditions is persistent or long-lasting. "Chronic trauma" would be something closer to Carpal Tunnel Syndrome than a gunshot wound.

IMO a better set of categories would be "Minor," "Moderate," "Severe," and "Critical."

PR #8453, opened 2021-09-24
Laser - Adds Aircraft Point Locking; Laser Spot Tracker

LorenLuke:
Trimmed down version of #7509.

For all Pilot Camera-equipped Vehicles:
- Automatically point lock one's own laser designator dot while it's on, allowing your vehicle to track buildings and such, rather than the terrain behind them.
- Add a laser spot tracker, toggled by the interaction menu.
- Automatically lock the pilot camera to any laser dot of matching code within a 25 degree cone, so long as the LST is enabled and the pilot camera is not locked to any target.


LorenLuke:
Requesting Review.

LinkIsGrim (approved):

addons/laser/XEH_postInit.sqf
```diff
@@ -33,7 +33,12 @@ if (hasInterface) then {
 ["ace_laserOn", {
     params ["_uuid", "_args"];
     TRACE_2("ace_laserOn eh",_uuid,_args);
+
     GVAR(laserEmitters) set [_uuid, _args];
+    private _unit = _args select 0;
+    if (hasPilotCamera _unit) then {
```
PabstMirror, line 39: don't we want some kind of locality check?

LinkIsGrim: `local _unit` should be safe enough, though this will require resetting the laser to work if locality changes.

mrschick:
Something must have changed since this was merged, in the way that the TGP behaves.
`laser_fnc_laserPointTrack` simply causes the TGP to jitter while designating in AREA track, both across terrain but also with the cursor on a building.

As far as I can tell, this is unfortunately not permitting the intended feature 👇🏻 to work, but also breaking otherwise functional vanilla behaviour.

> Automatically point lock one's own laser designator dot while it's on, allowing your vehicle to track buildings and such, rather than the terrain behind them.

I'll try to figure out a fix for this to make it fully functional again. Though if that isn't possible, I guess it would be best to deprecate that code, or alternatively provide a CBA Setting to disable its behaviour.

PR #8460, opened 2021-09-30
Repair - Fix claimed object abuse

Salluci:
- Remove ability to use claimed objects in multiple repairs.

Fix for the "wheel glitch" which lets two or more players repair multiple hitpoints with a single spare wheel/track.


jonpas:
Do we know of any 3rd party that uses `callbackProgress` that we can notify?

PR #8461, opened 2021-10-01
WIP - Various - Move to CBA Settings

Salluci:
- Move settings from ACE Settings framework to CBA Settings.


jonpas:
@PabstMirror Think we can automatically assume `movedToSQF = 1;` after this PR? #8415 also changes all ACEX (well ACE3 with the PR I guess) settings to CBA Settings. So if we will proceed with ACEX Merge, we should merge this after 3.14.0 together with ACEX Merge and clean up all `ACE_Settings.hpp` files.

jonpas:
Why the merge of name and type? That should be kept on separate lines.

Salluci:
More readable IMO, some components do it (nametags). Undo?

PR #8468, opened 2021-10-01
WIP - Various - Use engine hashmaps

Salluci:
- Switch CBA hashes to engine hashmaps.
- Minor optimization in arsenal_fnc_scanConfig.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -267,8 +267,7 @@ enableCamShake true;
 
 //FUNC(showHud) needs to be refreshed if it was set during mission init
 ["ace_infoDisplayChanged", {
-    GVAR(showHudHash) params ["", "", "_masks"];
-    if (_masks isNotEqualTo []) then {
+    if ((values GVAR(showHudHash)) isNotEqualTo []) then {
         [] call FUNC(showHud);
     };
 }] call CBA_fnc_addEventHandler;
--- a/addons/common/functions/fnc_showHud.sqf
+++ b/addons/common/functions/fnc_showHud.sqf
@@ -42,15 +42,15 @@ if (_reason != "") then {
     _reason = toLower _reason;
     if (_mask isEqualTo []) then {
         TRACE_2("Removing", _reason, _mask);
-        [GVAR(showHudHash), _reason] call CBA_fnc_hashRem;
+        GVAR(showHudHash) deleteAt _reason;
     } else {
         while {(count _mask) < 10} do { _mask pushBack true; };
         TRACE_2("Setting", _reason, _mask);
-        [GVAR(showHudHash), _reason, _mask] call CBA_fnc_hashSet;
+        GVAR(showHudHash) set [_reason, _mask];
     };
 };
 
-GVAR(showHudHash) params ["", "_reasons", "_masks"];
+(toArray GVAR(showHudHash)) params ["_reasons", "_masks"];
 private _resultMask = [];
 
 for "_index" from 0 to 9 do {
--- a/addons/finger/functions/fnc_incomingFinger.sqf
+++ b/addons/finger/functions/fnc_incomingFinger.sqf
@@ -28,7 +28,7 @@ private _fingerPos = if (_sourceUnit == ACE_player) then {
 TRACE_3("incoming finger:", _sourceUnit, _fingerPosPrecise, _fingerPos);
 
 private _data = [diag_tickTime, _fingerPos, ([_sourceUnit, false, true] call EFUNC(common,getName))];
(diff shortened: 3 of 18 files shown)


addons/finger/functions/fnc_incomingFinger.sqf
```diff
@@ -28,7 +28,7 @@ private _fingerPos = if (_sourceUnit == ACE_player) then {
 TRACE_3("incoming finger:", _sourceUnit, _fingerPosPrecise, _fingerPos);
 
 private _data = [diag_tickTime, _fingerPos, ([_sourceUnit, false, true] call EFUNC(common,getName))];
-[GVAR(fingersHash), _sourceUnit, _data] call CBA_fnc_hashSet;
+GVAR(fingersHash) set [_sourceUnit, _data];
```
mharis001, line 31: `HashMap`s don't support `OBJECT` as key type.

addons/common/XEH_postInit.sqf
```diff
@@ -267,8 +267,7 @@ enableCamShake true;
 
 //FUNC(showHud) needs to be refreshed if it was set during mission init
 ["ace_infoDisplayChanged", {
-    GVAR(showHudHash) params ["", "", "_masks"];
-    if (_masks isNotEqualTo []) then {
+    if ((values GVAR(showHudHash)) isNotEqualTo []) then {
```
dedmen, line 270: If you want to check for empty, use count or
`hash isEqualTo createHashMap`
Don't grab values, that allocates a new array and copies all values over if its not empty.

addons/common/functions/fnc_showHud.sqf
```diff
@@ -42,15 +42,15 @@ if (_reason != "") then {
...
     } else {
         while {(count _mask) < 10} do { _mask pushBack true; };
         TRACE_2("Setting", _reason, _mask);
-        [GVAR(showHudHash), _reason, _mask] call CBA_fnc_hashSet;
+        GVAR(showHudHash) set [_reason, _mask];
     };
 };
 
-GVAR(showHudHash) params ["", "_reasons", "_masks"];
+(toArray GVAR(showHudHash)) params ["_reasons", "_masks"];
```
dedmen, line 53: The code actually only uses reasons for logging.
Its a waste to always copy them out even though you only use it inside the logging.
Just use `keys GVAR(showHudHash)` down below in the TRACE_2 instead of always pulling them out even though TRACE is normally disabled.

addons/finger/functions/fnc_perFrameEH.sqf
```diff
@@ -15,30 +15,30 @@
...
         // Fade out:
         _drawColor set [3, ((_drawColor select 3) * ((_timeLeftToShow min 0.5) / 0.5))];
 
         drawIcon3D [QPATHTOF(UI\fp_icon2.paa), _drawColor, ASLtoAGL _pos, _iconSize, _iconSize, 0, _name, 1, 0.03, "RobotoCondensed"];
     };
-}] call CBA_fnc_hashEachPair;
+} forEach GVAR(fingersHash);
 
-if ((GVAR(fingersHash) select 1) isEqualTo []) then {
+if ((keys GVAR(fingersHash)) isEqualTo []) then {
```
dedmen, line 41: Again comparing to createHashmap is more efficient.
Just like you array variable to empty array, you should also compare hashmap variable to empty hashmap.

addons/overheating/script_component.hpp
```diff
@@ -3,7 +3,7 @@
 #include "\z\ace\addons\main\script_mod.hpp"
 
 // #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
+#define DISABLE_COMPILE_CACHE
```
dedmen, line 6:
```suggestion
// #define DISABLE_COMPILE_CACHE
```

addons/finger/functions/fnc_incomingFinger.sqf
```diff
@@ -28,7 +28,7 @@ private _fingerPos = if (_sourceUnit == ACE_player) then {
 TRACE_3("incoming finger:", _sourceUnit, _fingerPosPrecise, _fingerPos);
 
 private _data = [diag_tickTime, _fingerPos, ([_sourceUnit, false, true] call EFUNC(common,getName))];
-[GVAR(fingersHash), _sourceUnit, _data] call CBA_fnc_hashSet;
+GVAR(fingersHash) set [hashValue _sourceUnit, _data];
```
dedmen, line 31: Nice idea, I didn't think of that.
hashValue of object is basically BIS_fnc_netID that works in SP and MP

PR #8475, opened 2021-10-03
MagazineRepack - Add setting & hint for repacking equippped magazines

### LINKED ISSUE #8471: MagazineRepack - When repacking a weapon can be unloaded without the player knowing.
- **ACE3:** `3.13.6`

**Description:**
When repacking a magazine class, if a partial magazine is loaded into any weapon of the player, the loaded magazine gets removed without notice.
This can lead to situations where the player has repacked their magazines and tries to fire, but cannot as there is no magazine in the gun.

**Steps to reproduce:**
1. Have a weapon, 2 of the same class of magazine for that weapon, and available inventory space to hold at least 2 of that magazine class.
2. Fire some, but not all, of the rounds from one magazine, reload, fire some of the round in the second magazine.
3. Repack those magazines.
4. Attempt to fire weapon.
5. Open inventory, see 2 magazines in inventory, 0 magazines in weapon.

https://github.com/acemod/ACE3/blob/66ce0ba5ff79c9b43dea37b06b71649b720595b6/addons/magazinerepack/functions/fnc_startRepackingMagazine.sqf#L44-L46

**Expected behavior:**
The player should not be ignorant of unloading their own weapon.
If the player's weapon is unloaded, a hint or other notification should be displayed to inform them.

Alternatively, the loaded magazine should be ignored for repacking and should not be removed from the weapon.

Salluci:
- Title.
- Close #8471.

### DIFF
--- a/addons/magazinerepack/ACE_Settings.hpp
+++ b/addons/magazinerepack/ACE_Settings.hpp
@@ -1,26 +1,14 @@
 class ACE_Settings {
     //Time to move a round from one magazine to another
     class GVAR(timePerAmmo) {
-        category = CSTRING(DisplayName);
-        displayName = CSTRING(timePerAmmo);
-        value = 1.5;
-        typeName = "SCALAR";
-        sliderSettings[] = {1, 10, 1.5, 1};
+        movedToSQF = 1;
     };
     //Time to swap between magazines when repacking
     class GVAR(timePerMagazine) {
-        category = CSTRING(DisplayName);
-        displayName = CSTRING(timePerMagazine);
-        value = 2.0;
-        typeName = "SCALAR";
-        sliderSettings[] = {1, 10, 2, 1};
+        movedToSQF = 1;
     };
     //Time to relink 2 belts together
     class GVAR(timePerBeltLink) {
-        category = CSTRING(DisplayName);
-        displayName = CSTRING(timePerBeltLink);
-        value = 8.0;
-        typeName = "SCALAR";
-        sliderSettings[] = {1, 10, 8, 1};
+        movedToSQF = 1;
     };
 };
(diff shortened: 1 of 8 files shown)


Drofseh:
I'm happy with this solution.
I don't think there is a need for 0,1,2 though, just bool should be fine and always show the hint if it's turned on.
I think `repackLoadedMagazines` would be clearer than `repackEquippedMagazines`, but whatever.

Salluci:
Went with `equipped` because technically any magazine with ammo in it can be called 'loaded'. Equipped can only mean it's currently in use.

Drofseh:
`equipped` to me sounds like any magazine in my equipent/inventory
eg. "I am equipped with 10 magazines and a rifle."

Since arma doesn't have unloaded magazines, I still think `repackLoadedMagazines` sounds clearer

Dabako:
> 
> 
> I'm happy with this solution. I don't think there is a need for 0,1,2 though, just bool should be fine and always show the hint if it's turned on. I think `repackLoadedMagazines` would be clearer than `repackEquippedMagazines`, but whatever.

Do I understand that right. If there are no three options that would mean there is only the option to either don't repack the magazin that is loaded or do it with an hint? I am personally liking that touch that when you forget to reload your gun after repacking you are just without bullets. So it would be nice to have the option to repack that mag without a hint.

Dabako:
Yeah it would really be nicer if there would be an animation playing unloading the weapon so people could tell that you gun got unloaded. (I think to add something like an unloading animation would be a bit to complicated for this problem?)

But I was kinda happy that I unloaded it and that I got punished for not reloading my weapon. I understand that this behaviour can be seen as a bug. (What it kinda seems to be without an animation). 
I think it would be nice to give players that type of option to make it with our without hints because I personally just want as few hints as possible in my gameplay experience.

jonpas:
> Since arma doesn't have unloaded magazines, I still think `repackLoadedMagazines` sounds clearer

Agreed.

> If you really really want it though, I guess it's ok if it's included, but personally I think it's a bug not a feature.

Not a bug, just a missing animation as @Dabako said. I can see why it can be seen as a bug though, and that could be another solution in addition to this PR. I am ok with removing the "no hint" option for now for the sake of clarity.

PR #8480, opened 2021-10-04
Viewports

PabstMirror:
Adds usable viewports and display-screens for vehicles 

https://www.youtube.com/watch?v=Szv14eh6mS0

- Focus is mainly on passengers (as driver/gunner/commander are usually fine)
- Gives basic situational awareness to mounted infantry in a realistic manor.
- Simple "look to view" controls

Downsides:
- Not really relevant to most vanilla vics, this is really for other addons
- Config hell - using raw model offsets to define the viewports
The most complicated part of this addon is a dev tool to assist placing these

<details>
  <summary>Other options</summary>
  
- Why not make them all `RenderTargets` (e.g. CUP BTR-80 from front seat)?
Making them "screens" will reduce performance, looks worse imho
this is not how periscopes work:

cupBTR80

- Why not make them just windows (e.g. CUP LAV-25)?
This isn't bad, but it's often impossible to get your head in the correct position to see anything.
Also windows are not really the same as periscopes.
And modifying the model is not trivial (or possible)

I don't think there is a perfect solution, but this PR will improve the experience of mounted infantry.
</details>


addons/viewports/CfgVehicles.hpp
```diff
@@ -0,0 +1,169 @@
+class CfgVehicles {
+    class B_MBT_01_base_F;
+    class B_MBT_01_cannon_F: B_MBT_01_base_F { // Merkel
+        class ADDON {
```
dedmen, line 4: should ADDON macro really be used here?
As I understand it these macros are used so they auto update if you rename a component.
But if for some reason ace_viewports gets renamed in the future, we don't want this classname to change and break all mods that built upon this?

addons/viewports/CfgVehicles.hpp
```diff
@@ -0,0 +1,169 @@
...
+            class SLD_backLeftUpper {
+                type = "screen";
+                camLocation[] = {0,0,0.05};
```
dedmen, line 7: Would be good to also allow memory point in here.
Then future made vehicles can just directly provide the camera memory points inside the model

addons/viewports/CfgVehicles.hpp
```diff
@@ -0,0 +1,169 @@
...
+        };
+    };
+
+
+
+    class Wheeled_APC_F;
+    class Tank_F;
+
+     // 80 and 80A p3ds both use same modelPos
+    #if __has_include("\CUP\WheeledVehicles\CUP_WheeledVehicles_BTR80\CUP_BTR_80.p3d")
```
dedmen, line 34: Does HEMTT support leaving a config.cpp unbinarized?
Would be cool to add a 
#pragma NO_BINARIZE
or something like that, I can maybe add that to CfgConvert/AddonBuilder in 2.08

addons/viewports/CfgVehicles.hpp
jonpas: It's just to make things easier, this won't ever realistically get renamed.

addons/viewports/CfgVehicles.hpp
jokoho48: no, it does not and it does not even have support for #if nor __has_include

addons/viewports/CfgVehicles.hpp
simkas: I would second this, if this could use memory points, it would make adding this to new vehicles pretty easy and would be a very nice addition.

Drofseh:
With the "look to view" controls, can multiple people use the same viewport at once?

addons/viewports/functions/fnc_eachFrame.sqf
```diff
@@ -0,0 +1,77 @@
...
+/*
+* Author: PabstMirror
+* Runs each frame while inside of a vehicle with viewports
+*
+* Arguments:
+* 0: PFEH Args <ARRAY>
+*   0: Player <OBJECT>
+*   1: Vehicle <OBJECT>
+*   2: Viewport configuration <ARRAY>
+*   3: Viewport index showwn (-1 for none) <NUMBER>
```
veteran29, line 11:
```suggestion
*   3: Viewport index shown (-1 for none) <NUMBER>
```

addons/viewports/functions/fnc_eachFrame.sqf
```diff
@@ -0,0 +1,77 @@
...
+* [] call ace_viewports_fnc_eachFrame
+*
+* Public: No
+*/
+
+params ["_args", "_pfID"];
+_args params ["_player", "_vehicle", "_viewports", "_shownIndex", "_lastVisionMode"];
+
+private _newIndex = -1;
+if (cba_events_control) then {
```
veteran29, line 27: Is this for holding `control` key? Could be a normal hotkey.

veteran29 (approved):
LGTM

severgun:
Please, make docs more informative. 
```sqf
// camAttach [ARRAY or NUMBER] - Required
camAttach[] = {0,0};
camAttach = 55;
```
What is number and what is array? Oh, this is angle and turret path. Or not?
`// maxDistance [NUMBER] - Optional`
Is this activation distance?

severgun:
Sometimes viewports do not initialize. Maybe arma quirks.
* Open editor
* Place Merkava TUSK with crew
* Remove bots but keep playable.
* With context menu change seat to Passenger N
* Launch Preview
* No viewport display. Get out/get in does not fix it.
* Exit and launch again.
* Viewport works as intended

Is it possible to make viewport "screen" size, ratio and texture customizable by config?
If not, maybe hardcode bigger size? IRL you will lean quite close to periscope.

Also custom viewports could be used as "lean to window" functionality. As an example MI-8 windows.

PabstMirror:
finally looked into this and it's just a weird arma bug
`isTurnedOut player` is true when you do those specific mission steps (you can't even turn out from those cargo seats so this really makes no sense)

addons/viewports/dev/debugPoints.sqf
```diff
@@ -125,9 +125,11 @@ addMissionEventHandler ["Draw3D", {
     if (isNull _vehicle) exitWith {};
 
 
-    drawIcon3D ["#(argb,8,8,3)color(1,1,1,1)", [0,1,0,1], aslToAGL eyepos player, 0.1, 0.1, 0, "eye", 1, 0.02, "TahomaB"];
-    drawIcon3D ["#(argb,8,8,3)color(1,1,1,1)", [0,1,0,1], player modelToWorldVisual (player selectionPosition "pilot"), 0.1, 0.1, 0, "pilot", 1, 0.02, "TahomaB"];
-
+    drawIcon3D ["#(argb,8,8,3)color(1,1,1,1)", [1,1,0,1], _vehicle modelToWorldVisual [0,0,0], 0.1, 0.1, 0, "", 1, 0.02, "TahomaB"];
+    if (alive player) then {
```
Drofseh, line 129: ace_player instead?

PabstMirror:
```suggestion
    if (alive player) then { // not using ace_player so this works in 3den
```

PR #7039, opened 2019-06-07
Refueling fuel containers

BaerMitUmlaut:
**When merged this pull request will:**
- Cleanup some of the refuel code
  - Removed unnecessary macro
  - Removed duplicate code
  - More comments
  - Fixed a missing private
- Add refueling of containers (so you can refuel the fuel cargo of a fuel truck or a jerry can)

PR #8482, opened 2021-10-04
Refuel - Add Container Refueling

Salluci:
- Continuation of #7039.
- Switch to configOf and apply requested changes.
- Add check fuel interaction to jerry cans.
- Disable loading sink into cargo while nozzle is attached.
- Disable taking nozzles while another nozzle is attached.


Salluci:
Tried to break it every way I could think of, couldn't. MP syncing properly.

Salluci:
See #8981.

PR #8485, opened 2021-10-04
Grenades - Flashbang reaction fixes

BrettMayson:
- Debounce the flashbang reaction, which will make the 9 bang usable
- Fix the skill adjustment, so it doesn't lose sub skill values
- Store the flash strength on the unit, not used in ACE but I know I could use it in other mods personally

### DIFF
--- a/addons/grenades/functions/fnc_flashbangExplosionEH.sqf
+++ b/addons/grenades/functions/fnc_flashbangExplosionEH.sqf
@@ -64,7 +64,7 @@ _affected = _affected - [ACE_player];
             // Not used interally but could be useful for other mods
             _unit setVariable [QGVAR(flashStrength), _strength, true];
             {
-                _unit setSkill [_x, ((_unit skill _x) / 50)];
+                _unit setSkill [_x, (_unit skill _x) / 50];
             } forEach SUBSKILLS;
             [{
                 params ["_unit"];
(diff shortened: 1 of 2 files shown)


addons/grenades/functions/fnc_flashbangExplosionEH.sqf
```diff
@@ -47,28 +47,42 @@ private _affected = (ASLtoAGL _grenadePosASL) nearEntities ["CAManBase", 20];
...
-                [_unit, false] call EFUNC(common,disableAI);
-            };
-
-            _unit setSkill (skill _unit * 50);
-        }, [_x], 7 * _strength] call CBA_fnc_waitAndExecute;
+        private _dirToFlash = _unit getDir _grenadePosASL;
+        _unit setDir (_dirToFlash + linearConversion [0.2, 1, _strength, 40, 135] * selectRandom [-1, 1]);
+
+        private _flashReactionDebounce = _unit getVariable [QGVAR(flashReactionDebounce), 0];
+        _unit setVariable [QGVAR(flashReactionDebounce), CBA_missionTime + (7 * _strength)];
```
PabstMirror, line 62:
```suggestion
        _unit setVariable [QGVAR(flashReactionDebounce), _flashReactionDebounce max (CBA_missionTime + (7 * _strength))];
```
if later flashes come in at lower strength if could reduce the time

PabstMirror (approved):

mharis001 (approved):

Issue #8407, opened 2021-08-31
Advanced Fatigue - Medical overrides `swayFactor` setting


Whigital:
**Mods:**
- **Arma 3:** `2.04.147716`
- **CBA:** `3.15.5.210713`
- **ACE3:** `3.13.6` (https://github.com/acemod/ACE3/commit/8bd4922fccc86b53bae25d43a8c95b77c9b75f54)

**Description:**
Setting Adv. Fatigue sway factor lower than 1 gets overridden by hardcoded Medical sway factor

**Steps to reproduce:**

1. Enable Adv. Fatigue.
2. Set `ace_advanced_fatigue_swayFactor` to anything lower than `1`, `0.5` in this case.
3. Inspect the returns from the sway calcs:

```sqf
private _aimHash = (missionNamespace getVariable ["ACE_setCustomAimCoef", nil]);

{
    private _hashCode = [_aimHash, _x] call CBA_fnc_hashGet;

    diag_log text format ["%1 sway value: %2", _x, call _hashCode];
} forEach ([_aimHash] call CBA_fnc_hashKeys);

private _swayMax = [missionNamespace, "ACE_setCustomAimCoef", "max"] call ace_common_fnc_arithmeticGetResult;

diag_log text format ["Sway value from arithmeticGetResult: %1", _swayMax];
```

```
21:47:53 ace_advanced_fatigue sway value: 0.782006
21:47:53 ace_medical sway value: 1
21:47:53 Sway value from arithmeticGetResult: 1
```

**Expected behavior:**
Sway/amingCoef affected by `ace_advanced_fatigue_swayFactor` setting.

**Where did the issue occur?**
Anywhere

**Additional context:**
https://github.com/acemod/ACE3/blob/76676eee462cb0bbe400a482561c148d8652b550/addons/advanced_fatigue/XEH_postInit.sqf#L5-L20

https://github.com/acemod/ACE3/blob/76676eee462cb0bbe400a482561c148d8652b550/addons/medical/XEH_postInit.sqf#L8-L10

Overwriting the `ace_medical` `ACE_setCustomAimCoef` codeblock with `ace_advanced_fatigue_swayFactor` incorporated solves the issue, don't know if it's the best solution though:

```sqf
[missionNamespace, "ACE_setCustomAimCoef", "ace_medical", {
[...]

PR #8486, opened 2021-10-06
Advanced Fatigue - Fix swayFactor calculation/application

Whigital:
- Change the way `swayFactor` setting is applied
- Incorporate unit stance into aimingCoef

See https://github.com/acemod/ACE3/issues/8407 for full details, but the gist of it is that neither stance or swayFactor got incorperated into the aimCoef if it was below 1.

With these changes, stance and swayFactor is always taken into account, and any external coef extracted via `ACE_setCustomAimCoef`.

Fix https://github.com/acemod/ACE3/issues/8407


LinkIsGrim:
Superseded by #9226

PR #8488, opened 2021-10-07
Explosives - Fix Dial Tone Sound Position

654wak654:
- Replace `modelToWorldVisual` with `modelToWorldVisualWorld` so that `playSound3D` plays the dial sound at the correct position.

I wanted to implement dialing as requested [here](https://github.com/acemod/ACE3/issues/3594#issuecomment-919802127), but to my surprise there was already dial tone functionality in the module!

A quick search shows every other code that uses `playSound3D` with `modelToWorldVisual` first runs it through `AGLToASL`, so this bug was unique. **Edit:** I've made another PR (#8489) to get rid of this ambiguity everywhere else.

### DIFF
--- a/addons/explosives/functions/fnc_dialingPhone.sqf
+++ b/addons/explosives/functions/fnc_dialingPhone.sqf
@@ -22,7 +22,8 @@ params ["_args", "_pfID"];
 _args params ["_unit", "_i", "_arr", "_code"];
 
 if ((_i mod 4) == 0) then {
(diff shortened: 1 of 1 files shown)


654wak654:
I guess we could also get rid of the `_i mod 4` part, and add a setting for "IED Dialer Audible Chance" setting, but that is _definitely_ out of the scope of this fix :smile:.

PR #8491, opened 2021-10-07
Medical - Use amount of wounds for stitching

LinkIsGrim:
- Change stitching logic to close 1x of a wound every GVAR(woundStitchTime) seconds.

Current logic closes "sets" of wounds (i.e. 4x Med. Vel. wounds on left leg, same ID and same position) every GVAR(woundStitchTime).


BrettMayson:
We'll need to make the stitching kit smaller, and make the default stitch time shorter.

With the current implementation my group has run out of sutures several times in the field, this will make it even harder, especially in those cases when someone ends up with a whole bunch of wounds.

PR #8496, opened 2021-10-10
Hunter Killer (Commander Override) 

PabstMirror:
Allows commander to override gunner and slew the turret to the direction his view is facing.

https://www.youtube.com/watch?v=p5yoC9p35GQ

### How it works: 
- Stick `directionStabilized = 1;` on the puppet turret
- Normally you then manually turn it off 
    `_vehicle enableDirectionStabilization [false, _puppetTurret];`
- When commander override is triggered then enable it and then do
    `_vehicle lockCameraTo [_lookPoint, _puppetTurret];`

### Design Issues:

- Aim accuracy
The demo is setup with both low and high accuracy modes, but we should just pick one mode and stick with it

- Normal Precision
Uses eyeDirection/turretDirection on a non-local turret which is only roughly synced in mp.
e.g. at 1000m it will aim at the right building but not the right window.
As the commander tracks over it will kind of jump from rough point to point, not continously.

- High precision
We sync our local lookDir via setVar and it should perfectly match
I think this is fine, we just need to do a little work to make sure we don't overload network traffic

- Commander Firing
I tried `action//manualFire` and it was weird, it didn't seem to fire in the direction either turret was looking.
I suspect both FCS were conflicting.
Could try the other `fire` commands, but I'm ok leaving this feature to just slew and then handoff to gunner.

- Config hell (turret and optics inheritance is always a mess)


PabstMirror:
ll - yeah, we can probably use this same code to do slew-to-cue

2.08 will add 3rd arg, and won't require any `directionStabilized = 1;` configs
e.g. `(vehicle player) lockCameraTo  [w, [0], true]`
seems to have some smoothing to it that different from the other way that hard snaps, it also seems to works on AI

diwako:
Posting the summarization of the discussion about "missing featrures" from slack, before it gets lost in time.

- an option to disable that feature for yourself, so your view cannot be slaved
- gunner has the same keybinds and can do the same thing.

- Not all tanks allow it, so a config entry on the turrets makes sense to me.
`ace_hunterKiller = 0;`
With
```
0 - no hunter killer system (or -1)
1 - can be controlled
2 - can control
3 - both
```

Drofseh:
> * Not all tanks allow it, so a config entry on the turrets makes sense to me.
>   `ace_hunterKiller = 0;`
>   With
> 
> ```
> 0 - no hunter killer system (or -1)
> 1 - can be controlled
> 2 - can control
> 3 - both
> ```

I'd actually like to update my config entry suggestion and move the entry to the vehicle instead of the turret.

Have it be an array if arrays, with subarrays having the format [TURRER_INDEX, VALUE] 
```
ace_hunterKiller = [
    [[0], 1],
    [[1], 2]]
];
```

This will make adding it to compats a lot easier since it will avoid complicated turret inheritance issues, and it's a single config lookup instead of getting from each turret.

severgun:
> * gunner has the same keybinds and can do the same thing.

Should have possibility to turn off. As much as I remember IRL there is no such thing. Only commander can observe and force direction.

addons/hunterkiller/functions/fnc_keydown.sqf
```diff
@@ -0,0 +1,47 @@
+#include "script_component.hpp"
+/*
+ * Author: PabstMirror
+ * Slew keybind pressed
+ *
+ * Arguments:
+ * 0: Override if true, Observe if false <BOOL>
```
PabstMirror, line 7:
```suggestion
 * 0: Key mode ("override" or "observe") <STRING>
```

severgun:
[Add to all 2035 tanks](https://github.com/acemod/ACE3/pull/8496/commits/fdbb3698f388c0d960480896b87349cb11b316f1)

RHS AFRF Compat would be nice to have too.

nomisum:
would be nice to have some manual override/fnc to enable this for a non-supported tank quickly without the need for a compat mod

JonBons:
Just played around with this tonight and it seems to work fairly well.

addons/hunterkiller/functions/fnc_turretChangedEH.sqf
```diff
@@ -0,0 +1,51 @@
...
+            []
+        };
+    };
+};
+TRACE_1("",_hkArray);
+if ((count _hkArray) != 2) exitWith {};
+
+{
+    _x params ["_xTurret", "_xMode"];
+    TRACE_2("wtf",_playerTurret,_xTurret);
```
veteran29, line 45: Maybe put something more descriptive in here? 😂

docs/wiki/framework/hunterkiller-framework.md
```diff
@@ -0,0 +1,42 @@
...
+// OBSERVE = 1
+// OVERRIDE = 2
+// OBSERVE_AND_OVERRIDE = 3
+
+class CfgVehicles {
+    class MyTankA {
+        ace_hunterkiller = 1; // enable with default settings
+    };
+    class MyTankB {
+        ace_hunterkiller[] = {{{0}, 1}, {{0,0}, 3}}; // Default settings: (gunner can observer, commander can observe and override gunner)
```
veteran29, line 28:
```suggestion
        ace_hunterkiller[] = {{{0}, 1}, {{0,0}, 3}}; // Default settings: (gunner can observe, commander can observe and override gunner)
```

veteran29 (approved):

JonBons:
As an FYI, adding the APC base classes unlocks this ability on a lot of common mods. Might need compat changes to adjust scope of this feature.

severgun:
> or just to allow it on anything that has a gunner/commander turret team role.

pls no

BrettMayson (approved):

PR #8508, opened 2021-10-13
Repair - Use callback functions directly

LinkIsGrim:
- Switch callback lines in `ACE_Repair.hpp` to use functions directly in `fnc_repair` instead of call statement that'll need to be recompiled.


veteran29:
`QUOTE(LINKFUNC(name))` would be better for filepatching/dev.

PR #8513, opened 2021-10-14
ACEX - Final tweaks

PabstMirror:
define old funcs to keep old missions/doc usable
Apply `QXGVAR` for thrist/hunger -  https://github.com/acemod/ACE3/pull/8415#discussion_r728608132


jonpas:
Neat!

PR #8522, opened 2021-10-15
Overheating - Fix errors, improve documentation, change ACEX references for integration

Drofseh:
- title
- requires #8540

### DIFF
--- a/.github/workflows/pboproject.yml
+++ b/.github/workflows/pboproject.yml
@@ -1,0 +1,82 @@
+name: pboProject
+
+on:
+  push:
+    branches:
+    - master
+  pull_request_target:
+
+jobs:
+  pboproject:
+    runs-on: windows-2019
+    steps:
+    - name: Install Arma 3 Tools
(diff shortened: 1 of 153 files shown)


PabstMirror:
check overheating postinit
`acex_field_rations_helper` doesn't exist anymore and it should be `QEGVAR(field_rations,helper)`

PabstMirror:
https://github.com/acemod/ACE3/blob/master/addons/field_rations/XEH_postInit.sqf#L19
is `QGVAR(waterSource)`

https://github.com/acemod/ACE3/blob/master/addons/field_rations/XEH_postInit.sqf#L73
is `QXGVAR(waterSource)`

we really need to be careful mixing ace and acex
maybe we should maintain a list of all old acex_vars in the addon's readme.md

Drofseh:
Ok, I *think* I've got everything correct here.
I've also made a PR #8540 to add some external reference macros for AceX stuff that are now used here

PR #8532, opened 2021-10-16
Common - Fix calling getVar on null object in `canInteractWith`

PabstMirror:
Prime a grenade in adv_tossing and it spams rpt with
```
getVariable called on null object
 âž¥ Context: 	[] L49 (z\ace\addons\advanced_throwing\functions\fnc_prepare.sqf)
	[] L18 (z\ace\addons\advanced_throwing\functions\fnc_drawThrowable.sqf)
	[] L18 (z\ace\addons\advanced_throwing\functions\fnc_drawThrowable.sqf)
	[] L32 (z\ace\addons\advanced_throwing\functions\fnc_canPrepare.sqf)
	[] L33 (z\ace\addons\advanced_throwing\functions\fnc_canPrepare.sqf)
	[] L24 (z\ace\addons\common\functions\fnc_canInteractWith.sqf)
```
there is a call with null target
```
{[_unit, objNull, [...]] call EFUNC(common,canInteractWith)}
```

and in 2.06 calling getVar on a null throws rpt warning (but no popup even with showscript debug)

edit: this is a profiling branch thing
>- Added: Debug log message with context for when a script tries to getVariable on objNull/grpNull

### DIFF
--- a/addons/movement/functions/fnc_handleVirtualMass.sqf
+++ b/addons/movement/functions/fnc_handleVirtualMass.sqf
@@ -23,7 +23,9 @@ if (isNull _unit) exitWith {};
 private _virtualLoad = 0;
 
 {
-    _virtualLoad = _virtualLoad + (_x getVariable [QGVAR(vLoad), 0]);
+    if (!isNull _x) then {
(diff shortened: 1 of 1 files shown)


PabstMirror:
yeah

PR #8536, opened 2021-10-16
Towing - Fix you can still tow wreck even if you shouldn't

Vdauphin:
You can still try to attach rope on wreck which you shouldn't be allow according to the current implementation

- Little typo: `_target` is not define
- Change `_target` to `_parent`


Dabako:
Okay, but I think an implementation to have the ability to move wrecks in some way would also be really nice in the future.

PR #8544, opened 2021-10-17
Weather - Use apparent temp in `displayAirTemp`

PabstMirror:
Uses param temp as I think was the original intention
`displayTextPicture` was called twice, first one used undefined `_color` var

https://github.com/acemod/ACE3/blob/master/addons/weather/functions/fnc_getApproximateAirTemp.sqf 
is the only thing that calls this func

edit: Tag @LorenLuke for review


LorenLuke:
Looks good to me. Was L44 just an extra instance of L72?

PR #8566, opened 2021-10-20
Tools - Update sqf_linter.py to ignore parse/macro errors

PabstMirror:
tool has limited use because it struggles with macro usage and has many false positives, 
this limits the errors it reports to just local/private diagnostic


Dystopian:
close #5979?

PR #8578, opened 2021-10-22
Fire - Add drop weapon setting

veteran29:
- add settings for dropping weapon when on fire, set to AI only by default as it seems more sane
- Minor cleanup of stringtable to have same same casing as rest of the project
- use CBA_fnc_addSetting
- moved burn reaction effects to dedicated function


Drofseh:
> Is there a difference between being only slightly on fire and having every part of your skin render crispy? Would you drop your weapon in both cases?

I think having it as a random chance to drop with the amount of fire as a coef would be sensible.

PR #8584, opened 2021-10-22
Tools - Binarize configs in make.py

PabstMirror:
Run configs through `CfgConvert.exe` twice to get nice de-macroed configs

Ref we disabled bin in #8093

### DIFF
--- a/addons/rearm/XEH_preStart.sqf
+++ b/addons/rearm/XEH_preStart.sqf
@@ -1,3 +1,9 @@
 #include "script_component.hpp"
 
 #include "XEH_PREP.hpp"
+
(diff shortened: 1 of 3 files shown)


PabstMirror:
Can now use `$FORCE_ADDONMAKER$` to force pack with addon maker, which will allow us to use
`#if __has_include("` macros

the downside is it seems to add some `#include "BIS_AddonInfo.hpp"`
```
class BIS_AddonInfo
{
	author="";
	timepacked="1635942182";
};

```
junk to config.cpp

jonpas:
Any information on when `__has_include` support will be added to Mikero's free tools?

PR #8585, opened 2021-10-22
Fortify - Add Progress Bar

mjc4wilton:
- Add a progress bar when placing objects
- Time to build is dependent on cost and user-configurable coefficients to an Ax + b function
- Update documentation to match changes

Currently using GVAR for everything. Not sure if I should be using XGVAR or leave as is after the ACEX merge.


PabstMirror:
Just my opinion, but I'd rather the default time setting is 0, which would make it work the same as before.

mjc4wilton:
I'm not completely opposed to having it set to 0 by default, but at the same time since I feel like this is a long overdue improvement to the out-of-the-box experience.

Drofseh:
> Just my opinion, but I'd rather the default time setting is 0, which would make it work the same as before.

I agree

veteran29:
It feels bit weird to add stuff radically changing how the game works like Adv Vehicle Damage enabled by default but hesitating to have this enabled by default.

IMO it makes sense to need to wait a few seconds for fortification placement.

PabstMirror (approved):

jonpas:
> Currently using GVAR for everything. Not sure if I should be using XGVAR or leave as is after the ACEX merge.

GVAR is good, XGVAR only for things that need to be backwards compatible, which essentially means nothing new.

I agree default should be as is, changed from old, this makes sense.

addons/fortify/XEH_postInit.sqf
```diff
@@ -72,3 +72,32 @@ GVAR(objectRotationZ) = 0;
...
+// Place object event handler
+[QGVAR(deployFinished), {
+    params ["_args", "_elapsedTime", "_totalTime", "_errorCode"];
+    _args params ["_unit", "_side", "_typeOf", "_posASL", "_vectorDir", "_vectorUp"];
+
+    private _newObject = _typeOf createVehicle _posASL;
+    _newObject setPosASL _posASL;
+    _newObject setVectorDirAndUp [_vectorDir, _vectorUp];
+
+    // Server will use this event to run the jip compatible QGVAR(addActionToObject) event
```
jonpas, line 85:
```suggestion
    // Server will use this event to run the jip compatible QGVAR(addActionToObject) event and create the related map marker
```

jonpas (approved):

addons/fortify/initSettings.sqf
```diff
@@ -10,6 +10,22 @@
     ]
 ] call CBA_fnc_addSetting;
 
+[
+    QGVAR(timeCostCoefficient),
+    "SLIDER",
+    [LLSTRING(settingHint_timeCostCoefficient), LLSTRING(settingHintDesc_timeCostCoefficient)],
+    LLSTRING(settingsCategory),
+    [0, 10, 1, 2] // Min, Max, Default, Trailing Decimals, is Percentage
+] call CBA_fnc_addSetting;
```
PabstMirror, line 19: Should these settings be global?

mjc4wilton: Come to think of it: Definitely

Issue #8569, opened 2021-10-21
Weapon cook off does not work when unit is moving with weapon not up


diwako:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.06` [e.g. 1.00 stable, rc,  dev]
- **CBA:** `3.15.6` [e.g. 3.0.0 stable, commit hash]
- **ACE3:** `3.14.0` [eg. 3.0.0 stable, commit hash]

**Description:**
Weapon cook off does not make the gun fire when jogging or running with gun lowered. This is as in not using the combat pace option.

You still get the pop ups saying your gun just cooked off

**Steps to reproduce:**

1. Make weapon hot hot hot until it cooks off
2. start jogging or running

**Expected behavior:**
Gun fires anyways, jogging does not stop the carnage

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

Drofseh:
Using `forceWalk` isn't as bad as I feared it would be. When you are sprinting, are set to `forceWalk`, and then the `forceWalk` condition is removed, you go right back to sprinting.

There needs to be a slight delay between being `forceWalk` being turned on and the fired event though, to give time for the animation to change, or else the weapon doesn't actually fire.

Drofseh:
I think the best solution in the end will be to remove the use of `forceWeaponFire` and instead spawn the projectile and effects and play audio.

I've been looking into how to do this, but I'd like some help if anyone has time.

PR #8617, opened 2021-10-25
Overheating - Fix issues from release

### LINKED ISSUE #8642: Overheating - weapon heat not cleared on respawn
- **Arma 3:** `2.06 (stable and profiling)`
- **CBA:** `3.15.6 (workshop)`
- **ACE3:** `3.14 (workshop)`

**Description:**
When a unit dies, the heat level of their weapon is not reset. This is especially noticeable if your weapon has reached cookoff temp.

**Steps to reproduce:**

1. In any mission with respawn enabled
2. Fire until weapon heats up
3. Respawn
4. Check weapon temperature

**Expected behavior:**
Temperature of all weapons should be reset on respawn

- Dedicated & local MP

**Log Files:**
- https://gist.github.com/pterolatypus/fc531ad1dcdc8651a90dbb852cd48ecb

Drofseh:
- add `ace_overheating_fnc_canCoolWeaponWithItem` function
- add setting, `ace_overheating_suppressorCoef`, to adjust how much additional heat goes into the weapon when a suppressor is equipped
- add setting, `ace_overheating_coolingCoef`, to adjust how quickly weapons cool
- add exit to ammo temp loop if `ace_overheating_cookoffCoef` is changed to 0 mid-mission, this prevents an issue where all weapons start to cookoff regardless of temp, because required temp gets multiplied by `ace_overheating_cookoffCoef` which has been set to 0
- add public functions `ace_overheating_fnc_getAmmoTemperature`, `ace_overheating_fnc_setAmmoTemperature`, `ace_overheating_fnc_getWeaponTemperature`, `ace_overheating_fnc_setWeaponTemperature`
- add type of jam to `ace_weaponJammed` local event, close acemod#8637
- clear all weapon heat on death, fix #8642
- deprecate `ace_overheating_fnc_getBarrelMass` to be a wrapper for `ace_overheating_fnc_getWeaponData` that only returns barrel mass, this is cached and therefore faster
- improve documentation
- move `ace_overheating_heatCoef` check to a more sensible place
- move overheating cookoff into separate function `ace_overheating_fnc_cookoffWeapon`
- require mission restart when `ace_overheating_heatCoef` setting is changed (it gets cached per ammo type)
- use ambientTemperature as floor for weapon and ammo temp
- workaround for #8657


Drofseh:
My workaround for #8657 will check if they are surrendering or captive and not show the interaction.

I don't think this truly solves the issue though as other things added to the hand selection in the future will be able to overlap.

addons/overheating/functions/fnc_canCoolWeaponWithItem.sqf
```diff
@@ -0,0 +1,36 @@
...
+
+private _return = false;
+
+if (
+    GVAR(enabled)
+    && {isClass (configfile >> "CfgPatches" >> "acex_field_rations")}
+    && {[_unit, currentWeapon _unit] call FUNC(getWeaponTemperature) > (ambientTemperature select 0)}
+) then {
+    {
+        if (getNumber (configFile >> "CfgWeapons" >> _x >> QEXGVAR(field_rations,thirstQuenched)) > 0) exitWith {
```
Salluci, line 30: Another config lookup running every frame here

Drofseh: This should only run when the interaction menu is actually open

addons/overheating/functions/fnc_updateAmmoTemperature.sqf
```diff
@@ -47,58 +49,12 @@ if (_ammoTemperature < _barrelTemperature) then {
     _ammoTemperature = _barrelTemperature;
 };
 
-// check for cook off
+// cookoff if too hot
 if (_ammoTemperature > (GUNPOWDER_IGNITION_TEMP * GVAR(cookoffCoef))) then {
+    [_unit, _weapon, _canUnjam, _jamType] call ace_overheating_fnc_cookoffWeapon;
```
PabstMirror, line 54:
```suggestion
    [_unit, _weapon, _canUnjam, _jamType] call FUNC(cookoffWeapon);
```

addons/overheating/CfgVehicles.hpp
```diff
@@ -66,7 +66,7 @@ class CfgVehicles {
                 };
                 class GVAR(CoolWeaponWithItem) {
                     displayName = CSTRING(CoolWeaponWithItem);
-                    condition = QUOTE(GVAR(enabled) && {isClass (configfile >> 'CfgPatches' >> 'acex_field_rations')});
+                    condition = 'GVAR(enabled) && {[_player, _target] call FUNC(canCoolWeaponWithItem)} && {!(_unit getVariable [QEGVAR(captives,isSurrendering), false])} && {!(_unit getVariable [QEGVAR(captives,isHandcuffed), false])}';
```
PabstMirror, line 69:
```suggestion
                    condition = 'GVAR(enabled) && {[_player, _target] call FUNC(canCoolWeaponWithItem)} && {!(_target getVariable [QEGVAR(captives,isSurrendering), false])} && {!(_target getVariable [QEGVAR(captives,isHandcuffed), false])}';
```

PabstMirror: This line is getting a little long for configs,
we could move it to the func with something like
```

if ((_unit != _player) && {(_unit getVariable [QEGVAR(captives,isSurrendering), false]) || {_unit getVariable [QEGVAR(captives,isHandcuffed), false]}}) exitWith {
    false;
};
```

addons/overheating/XEH_postInit.sqf
```diff
@@ -76,8 +76,18 @@ if (hasInterface) then {
         }] call CBA_fnc_addClassEventHandler;
     };
 
+    // Reset all weapon heat to ambient on death to prevent cookoffs when a unit respawns.
+    ["CAManBase", "Killed", {
+        params ["_unit"];
+        {
+            _unit setVariable [_x, ambientTemperature select 0];
+        } forEach (_unit getVariable [QGVAR(trackedWeapons), []]);
```
PabstMirror, line 84: I think we could move all this to ace_overheating_fnc_handleRespawn

Drofseh:
I've had a request from @10Dozen to add a keybind to unload the magazine. Use case is that's faster to hit a key than it is to open the inventory and drag the mag from weapon to container.

He'd like it in ace_common so that it can still be used even when overheating isn't loaded, for example if a group has a "no loaded weapon in base" rule and they come back from a patrol.

I've written a function for it, but one thing I want to include is the possibilty for a custom unload animation `ACE_unloadAction` and unload sound `ACE_unloadSound` config entries from mod makers, similar to how there is a config entry for a custom jam clearning animation `ACE_clearJamAction`.

If this keybind goes into ace_common though, I'm not sure where to document these config entries.

Thoughts?

addons/overheating/XEH_postInit.sqf
Drofseh: Do you think it's safe to wait until respawn, rather than immediately on death?
I wouldn't want errors or weird things to happen while they are in spectator or otherwise waiting to respawn.

PabstMirror:
IMHO it could be added later to another PR if it's going to be in common
We'd really like to get these fixes done for 3.14.1

addons/overheating/XEH_postInit.sqf
Drofseh: @PabstMirror ?

PabstMirror: This is probably safer as is, so lets not change it now

Drofseh:
I think I'm done for this PR.
#8569 is still an issue, but I'm not sure how to solve that yet and I think it's fine to leave it for another PR.

addons/overheating/functions/fnc_canCoolWeaponWithItem.sqf
```diff
@@ -0,0 +1,36 @@
...
+            _return = true;
+        };
+    } forEach (_player call EFUNC(common,uniqueItems));
+};
+
+_return
```
mharis001, line 36: Can be written without the `if` and `findIf` instead.

PabstMirror:
```
GVAR(enabled)
&& {isClass (configfile >> "CfgPatches" >> "acex_field_rations")}
&& {[_unit, currentWeapon _unit] call FUNC(getWeaponTemperature) > (ambientTemperature select 0)}
&& {(((_player call EFUNC(common,uniqueItems)) findIf {getNumber (configFile >> "CfgWeapons" >> _x >> QEXGVAR(field_rations,thirstQuenched)) > 0}) != -1}
```

pterolatypus:
>require mission restart when ace_overheating_heatCoef setting is changed (it gets cached per ammo type)

Why not just clear the cache whenever the setting is changed? This seems like the kind of value people would want to tweak progressively without having to restart the mission.

addons/overheating/CfgVehicles.hpp
```diff
@@ -66,7 +66,7 @@ class CfgVehicles {
                 };
                 class GVAR(CoolWeaponWithItem) {
                     displayName = CSTRING(CoolWeaponWithItem);
-                    condition = QUOTE(GVAR(enabled) && {isClass (configfile >> 'CfgPatches' >> 'acex_field_rations')});
+                    condition = 'GVAR(enabled) && {[_player, _target] call FUNC(canCoolWeaponWithItem)} && {!(_target getVariable [QEGVAR(captives,isSurrendering), false])} && {!(_target getVariable [QEGVAR(captives,isHandcuffed), false])}';
```
PabstMirror, line 69: It seems like this actually compiles with pboProject,
but with dev (build.py/hemtt) builds it crashes game with
```
 File z\ace\addons\overheating\CfgVehicles.hpp, line 68: '/CfgVehicles/CAManBase/ACE_Actions/ACE_Weapon/ace_overheating_CoolWeaponWithItem.condition': Missing ';' prior '}'
 Error context & {!(_target getVariable ["ace_captives_isSurrendering", false])} && {!(_target getVariable ["ace_captives_isHandcuffed", false])}';
 ErrorMessage: File z\ace\addons\overheating\CfgVehicles.hpp, line 68: '/CfgVehicles/CAManBase/ACE_Actions/ACE_Weapon.': '&' encountered instead of '='
```

PabstMirror (approved):

Drofseh:
> > require mission restart when ace_overheating_heatCoef setting is changed (it gets cached per ammo type)
> 
> Why not just clear the cache whenever the setting is changed? This seems like the kind of value people would want to tweak progressively without having to restart the mission.

Makes sense, what would be the best way to do this?

docs/wiki/feature/overheating.md
```diff
@@ -53,12 +53,11 @@ Jams can be cleared in the following ways:
 - Select `Equipment`.
 - Select `Check weapon temperature`.
 
-**NOTE** When the bar is half full (yellow) it means the barrel is around 500°c.
-Your weapon will be even more prone to jams, and it'll get worse if you don't let the barrel cool down or swap it.
+**NOTE** Each section on the bar represents 100°c. When the bar reaches 2 sections weapons can start to cookoff. When it is half full (yellow) it means the barrel is around 500°c. Your weapon will be even more prone to jams, and it'll get worse if you don't let the barrel cool down or swap it.
```
jonpas, line 56:
```suggestion
**NOTE** Each section on the bar represents 100°C. When the bar reaches 2 sections weapons can start to cookoff. When it is half full (yellow) it means the barrel is around 500°C. Your weapon will be even more prone to jams, and it'll get worse if you don't let the barrel cool down or swap it.
```

addons/overheating/functions/fnc_setWeaponTemperature.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Public: Yes
+ */
+
+params ["_unit", "_weapon", "_temp"];
+
+private _ammoTempVarName = format [QGVAR(%1_ammoTemp), _weapon];
+
+_unit setVariable [_ammoTempVarName, _temp];
+
+nil
```
jonpas, line 26:
```suggestion
```
No need to return anything.

addons/overheating/functions/fnc_setAmmoTemperature.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Public: Yes
+ */
+
+params ["_unit", "_weapon", "_temp"];
+
+private _ammoTempVarName = format [QGVAR(%1_ammoTemp), _weapon];
+
+_unit setVariable [_ammoTempVarName, _temp];
+
+nil
```
jonpas, line 26:
```suggestion
```

addons/overheating/functions/fnc_getWeaponTemperature.sqf
```diff
@@ -0,0 +1,23 @@
+#include "script_component.hpp"
+/*
+ * Author: drofseh
+ * Get current temperaure of weapon.
```
jonpas, line 4:
```suggestion
 * Get current temperature of weapon.
```

addons/overheating/functions/fnc_getBarrelMass.sqf
```diff
@@ -17,4 +17,6 @@
 
 params ["_weapon"];
 
-METAL_MASS_RATIO * (getNumber (configFile >> "CfgWeapons" >> _weapon >> "WeaponSlotsInfo" >> "mass") / 22.0) max 1.0;
+([_weapon] call FUNC(getWeaponData)) params ["", "", "", "", "", "", "", "_barrelMass"];
```
jonpas, line 20:
```suggestion
([_weapon] call FUNC(getWeaponData)) param [7];
```
Better?

addons/overheating/functions/fnc_getAmmoTemperature.sqf
```diff
@@ -0,0 +1,23 @@
...
+ * [player, currentWeapon player] call ace_overheating_fnc_getAmmoTemperature
+ *
+ * Public: Yes
+ */
+
+params ["_unit", "_weapon"];
+
+private _ammoTempVarName = format [QGVAR(%1_ammoTemp), _weapon];
+
+_unit getVariable [_ammoTempVarName, ambientTemperature select 0];
```
jonpas, line 23:
```suggestion
_unit getVariable [_ammoTempVarName, ambientTemperature select 0]
```

addons/overheating/functions/fnc_getAmmoTemperature.sqf
```diff
@@ -0,0 +1,23 @@
+#include "script_component.hpp"
+/*
+ * Author: drofseh
+ * Get current temperaure of weapon's ammo.
```
jonpas, line 4:
```suggestion
 * Get current temperature of weapon's ammo.
```

addons/overheating/functions/fnc_coolWeaponWithWaterSource.sqf
```diff
@@ -59,7 +59,7 @@ private _fnc_condition = {
         };
 
         //Cool the weapon down
-        private _barrelMass = _weapon call FUNC(getBarrelMass);
+        ([_weapon] call FUNC(getWeaponData)) params ["", "", "", "", "", "", "", "_barrelMass"];
```
jonpas, line 62:
```suggestion
        ([_weapon] call FUNC(getWeaponData)) param [7];
```

addons/overheating/functions/fnc_coolWeaponWithItem.sqf
```diff
@@ -61,7 +61,7 @@ private _fnc_onSuccess = {
     };
 
     // cool the weapon
-    private _barrelMass = _weapon call FUNC(getBarrelMass);
+    ([_weapon] call FUNC(getWeaponData)) params ["", "", "", "", "", "", "", "_barrelMass"];
```
jonpas, line 64:
```suggestion
    ([_weapon] call FUNC(getWeaponData)) param [7];
```
This might be worth macro'ing.
Also why does this no longer just call `getBarrelMass`, I don't see a reason to deprecate that function, it's convenient wrapper for this magic param index.

jonpas (changes requested):

addons/overheating/functions/fnc_coolWeaponWithItem.sqf
Drofseh: wouldn't `([_weapon] call FUNC(getWeaponData)) param [7];` have to be this?

`private _barrelMass = ([_weapon] call FUNC(getWeaponData)) param [7];`

`_barrelMass` gets used on the next line in `_temperature = [_temperature, _barrelMass, _liquidAmount * 10] call FUNC(calculateCooling);`

jonpas: Ah yes, of course! Fixed in suggestion.

Drofseh: Part of my change to params was because it creates the private variable automatically, as well as this being the way getting caged weapon data is retrieved in other places in Overheating. 

Is doing it the way you suggest faster? Or do you just want it for readability?

jonpas: Lots of empty arguments in `params` I believe is slower?

But idk, I still prefer that helper function. Either way, good to merge I think.

jonpas (approved):

addons/overheating/functions/fnc_coolWeaponWithItem.sqf
veteran29: Why `param` instead of `select` if you're not using the default value?

Drofseh: 0.001 ms for `private _barrelMass = ["1","2","3","4","5","6","7"] select 6;`
0.0012 ms for `private _barrelMass = ["1","2","3","4","5","6","7"] param [7];`
0.0015 ms for `["1","2","3","4","5","6","7"] params ["", "", "", "", "", "", "", "_barrelMass"];`

So it's definitely slower, I'll switch it to `select` for when only one value is needed

Drofseh: Pabst merged just before I committed so I'll make another quick PR for this

Issue #7893, opened 2020-09-03
Medical stuck in uncon animation


dedmen:
**Mods:**
- **Arma 3:** `2.01` [stable]
- **CBA:** `3.15.0` [23fa433d]
- **ACE3:** `3.13.4` [434b0b9a]
~_Make sure to reproduce the issue with only CBA and ACE3 on a newly created mission!_~ (nu)

**Description:**
After waking up, staying stuck in uncon animation

**Steps to reproduce:**
- two people in mp
- player1 kills player2 to uncon
- player2 execute script `[player] call ace_medical_treatment_fnc_fullHealLocal;`
- repeat until player2 is awake, but player1 still sees him in uncon animation
- player 2 execute
```
{
	str [lifeState player2, simulationEnabled player2, animationState player2,  (player2 getVariable ["ACE_isUnconscious", false])] remoteExec ["systemChat", remoteExecutedOwner];
} remoteExec ["call", 2];
```
- player2 is not uncon, but animationState is "ace_medical_engine_facedown_unconscious"
- look at code
https://github.com/acemod/ACE3/blob/master/addons/medical_engine/functions/fnc_setUnconsciousAnim.sqf#L62
code should workaround, but code checks for "ace_unconscious_" name, which is not the case
- fix
player2 execute
```
{
    [player2, "AmovPpneMstpSnonWnonDnon", 2] call ace_common_fnc_doAnimation;	
} remoteExec ["call", 2];
```
or player1 execute
```
{
	if (
		!(_x getVariable ["ACE_isUnconscious", false]) && 
		{private _animation = animationState _x; _animation == "unconscious" || {_animation == "deadstate" || {_animation find "ace_medical_" != -1}}}
	) then {
		[_x, "AmovPpneMstpSnonWnonDnon", 2] call ace_common_fnc_doAnimation;
	};
} forEach allPlayers;
```

**Expected behavior:**
Not stuck in animation

**Where did the issue occur?**
- Dedicated

**Log Files:**
My poo is made of goo

**Additional context:**
Stuck bug happened about 20 times in 3 player 1 hour liberation play, and only took 3 tries (ala "steps to reproduce") to repro.

**Screenshots:**
![](https://cdn.discordapp.com/attachments/558666278957613057/751042044817834024/unknown.png)
player on left is alive but bugged

Whigital:
Would this be acceptable -> https://github.com/Whigital/ACE3/compare/medical_uncon-anims

[...]

PR #8623, opened 2021-10-25
Medical Engine - Fix unconscious animation state checks

Whigital:
- Unify `Medical Engine` unconscious animations with a common prefix (`ace_medical_engine_uncon_anim`).
- Use unified prefix in `animationState` check (`QGVAR(uncon_anim)`).

Fix https://github.com/acemod/ACE3/issues/7893


Fullback14:
This is breaking random unconscious anims for me. Units default to the standard prone anim instead.

PabstMirror (approved):
Seems to be working correctly

jonpas (approved):

PR #8624, opened 2021-10-25
Zeus - Add module for burning units

BaerMitUmlaut:
- Add a module for burning units


addons/zeus/functions/fnc_moduleBurn.sqf
```diff
@@ -0,0 +1,50 @@
...
+
+if !(local _logic) exitWith {};
+
+private _unit = attachedTo _logic;
+
+// Validate module target
+scopeName "Main";
+private _fnc_errorAndClose = {
+    params ["_msg"];
+    deleteVehicle _logic;
```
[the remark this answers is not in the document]
BaerMitUmlaut: I've simplified the code. It was mostly copied from another module, I probably should have taken a closer look.

jonpas (approved):

Drofseh (approved):

addons/zeus/config.cpp
```diff
@@ -25,6 +25,7 @@ class CfgPatches {
             QGVAR(moduleTeleportPlayers),
             QGVAR(moduleHeal),
             QGVAR(moduleSuicideBomber),
+            QGVAR(moduleBurn),
```
mharis001, line 28: Should be removed since it is in the units array of the `GVAR(fire)` class.

TheCandianVendingMachine (approved):

addons/zeus/config.cpp
PabstMirror: I was wondering about this, because other modules are listed in 2 different `units[]`

```suggestion
```

thojkooi (approved):

Issue #8630, opened 2021-10-26
Multiple units on fire makes it impossible to be put on fire by standing close to just one unit burning


Crowdedlight:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.06 stable`
- **CBA:** `v3.15.6.211004 stable` 
- **ACE3:** `master branch` commit hash: 08dff76  

**Description:**
If multiple units are on fire from the new fire module and you stand close to one, then you won't catch on fire yourself as the ``burnCounter`` keeps resetting between the different instances of the fire script spawned per unit on fire. 

**Steps to reproduce:**
1. spawn 2 units and yourself as player. disable ai ``move/path`` abilities to avoid them running away
2. watch burncounter in debug console: ``player getVariable ["ace_fire_burnCounter", 0]``
3. Put one unit on fire, stand next to it
4. before burn counter reaches 5, put other unit on fire
5. you should now see the ``burnCounter`` keep going between 0 and 1 and you never catch on fire.

**Expected behavior:**
I expected that I would be able to be put on fire standing close to one unit on fire, no matter how many other units in server is on fire

**Where did the issue occur?**
self-hosted multiplayer, if needed can validate on dedicated server. 

**Log Files:**
https://gist.github.com/Crowdedlight/6349771d7ea8036ae27b9c2d4e0ce5df

**Additional context:**
Based on the source code the issue is in ``fnc_burn.sfq`` line 166 to 176.  (https://github.com/acemod/ACE3/blob/master/addons/fire/functions/fnc_burn.sqf#L165-L177)   
The loop that increments the burn indicator reads from public variable that is set on your own player. This means that when you have multiple instances of the PFH running as multiple units are on fire, they will all save and read their individual ``burnCounter`` to the same public variable. 

I haven't tested it yet, but my immediate thought was to change so the ``burnCounter`` would be saved on the unit that is on fire not the ``ace_player``. As long as this is done without sync over network each instance should have their own ``burnCounter`` as it is saved on the unit burning not the ``ace_player``. 

I will test and if it works make a PR today or tomorrow.

Issue #8631, opened 2021-10-26
You can be put on fire when standing next to burning unit even if allowDamage is disabled


Crowdedlight:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.06 stable`
- **CBA:** `v3.15.6.211004 stable` 
- **ACE3:** `master branch` commit hash: 08dff76  

**Description:**
If unit is set on fire and you stand close to it while "allowDamage" is set to false, you will still be put on fire after a moment. 

**Steps to reproduce:**
1. spawn 1 units and yourself as player. disable ai ``move/path`` abilities to avoid it running away
2. disable damage for yourself
3. Put one unit on fire, stand next to it
4. After the normal interval you will be put on fire and take damage

**Expected behavior:**
I expect that if damage is disabled the fire would not propergate and catch myself on fire. 

**Where did the issue occur?**
self-hosted multiplayer, if needed can validate on dedicated server. 

**Log Files:**
https://gist.github.com/Crowdedlight/6349771d7ea8036ae27b9c2d4e0ce5df

**Additional context:**
The fire propagation code does not take into account if damage is allowed on the unit its about to spread to.

LinkIsGrim:
Fixed in #8632.

PR #8632, opened 2021-10-26
Fire - Fix fire spread to invincible units and not being able to be put on fire by proximity if multiple units are on fire

Crowdedlight:
This fixes https://github.com/acemod/ACE3/issues/8630 https://github.com/acemod/ACE3/issues/8631 

It adds a check for ``isDamageAllowed`` before it tries to spread fire to a nearby unit. So units with disabled damage won't be put on fire. 

It changes so the ``burnCounter`` is saved as a variable on the unit that is on fire, and not on the player. This means that multiple instances of the fire script (when multiple units is on fire) does not reset each others ``burnCounter`` and thus never allow one to be put on fire by proximity. 

The burnCounter fix has not been tested yet, will do later tonight or tomorrow when I got time.


addons/fire/functions/fnc_burn.sqf
```diff
@@ -162,8 +162,8 @@ if (_isBurning) exitWith {};
                         [QGVAR(burn), [ace_player, _intensity * (7 / 8), _instigator]] call CBA_fnc_globalEvent;
                     };
                 } else {
-                    if ((ace_player isKindOf "Man") && { !(_unit isEqualTo ace_player) }) then {
-                        private _burnCounter = ace_player getVariable [QGVAR(burnCounter), 0];
+                    if ((ace_player isKindOf "Man") && { !(_unit isEqualTo ace_player) } && isDamageAllowed ace_player) then {
```
commy2, line 165: While changing this, `x isNotEqualTo y` is now a thing that can replace `!(x isEqualTo y)`.

Crowdedlight: Changed all the instances of ``!( x isEqualTo y)`` in the file to ``x isNotEqualTo y``

addons/fire/functions/fnc_burn.sqf
```diff
@@ -162,8 +162,8 @@ if (_isBurning) exitWith {};
                         [QGVAR(burn), [ace_player, _intensity * (7 / 8), _instigator]] call CBA_fnc_globalEvent;
                     };
                 } else {
-                    if ((ace_player isKindOf "Man") && { !(_unit isEqualTo ace_player) }) then {
-                        private _burnCounter = ace_player getVariable [QGVAR(burnCounter), 0];
+                    if ((ace_player isKindOf "Man") && { _unit isNotEqualTo ace_player } && isDamageAllowed ace_player) then {
```
LinkIsGrim, line 165:
```suggestion
                    if ((ace_player isKindOf "Man") && { _unit isNotEqualTo ace_player } && {isDamageAllowed ace_player && {ace_player getVariable [QEGVAR(medical,allowDamage), true]}}) then {
```

LinkIsGrim (approved):

PR #8647, opened 2021-10-28
Explosives - Fix `objNull` reported as source and instigator in BI `handleDamage` event for explosive triggered by timer

Vdauphin:
title

~~This fix only works on client. I don't know if this function is supposed to be call server side.~~
~~As workaround, we could check if player is not null and pass _explosive as instigator.~~

- Pass the player setting the timer as instigator
- relevent function: 
  - https://github.com/acemod/ACE3/blob/06336b8dcfa8128b37ecb0ffb9961def8e02c22f/addons/explosives/functions/fnc_detonateExplosive.sqf#L58
  - https://github.com/acemod/ACE3/blob/87cfd9632e369dcd7529700f0f2e5bdd332697ab/addons/explosives/XEH_postInit.sqf#L28

### DIFF
--- a/addons/explosives/ACE_Triggers.hpp
+++ b/addons/explosives/ACE_Triggers.hpp
@@ -56,7 +56,7 @@ class ACE_Triggers {
         isAttachable = 1;
         displayName = CSTRING(timerName);
         picture = QPATHTOF(data\UI\Timer.paa);
-        onPlace = QUOTE([ARR_4(_this select 1, _this select 3 select 0, '#timer', _this select 0)] call FUNC(startTimer); false);
+        onPlace = QUOTE([ARR_4(_this select 1, _this select 3 select 0, nil, _this select 0)] call FUNC(startTimer); false);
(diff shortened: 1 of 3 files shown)


PabstMirror:
The only case it wouldn't be player is if someone is using the AI Speed Dial, which is probably rare
But it might be better to use https://github.com/acemod/ACE3/commit/c29622f42d1c3acb89f56765ca34b43bcc3666ac
and pass the specific unit to this func

PabstMirror:
@Dahlgren 
automaton checks for pboProject claim it failed with
```
onPlace = QUOTE([ARR_4(_this select 1, _this select 3 select 0, '#timer', _this select 0)] call FUNC(startTimer); false);
In File z\ace\addons\explosives\ACE_Triggers.hpp: circa Line 59 Unknown #
```
github action is using DePbo.7.97
latest free is 8.35 and it builds fine on that

we could also just change the 3rd arg to `nil`

PabstMirror (approved):

jonpas (approved):

PR #8648, opened 2021-10-29
Cookoff - Limit maximum cook intensity/duration

veteran29:
- title

**Reasoning:**
Advanced vehicle damage calculates the intensity (which indirectly also means duration) of the cook-off by getting the amount of the explosive ammo. This can result in ridiculously long cook-offs for vehicles that have autocannons with HE ammunition.

I've decided to use 10 intensity as maximum which roughly should limit it to ~7.5min (the intervals between cook-offs are randomised).

We could consider making this a setting but I'm not sure how to properly tell the user what it does.
  
  
**Reproduction of the issue:**
- spawn `O_APC_Tracked_02_cannon_F`
- spawn 'B_MBT_01_TUSK_F`
- shot IFV in the turret with the Tank Heat round
- ~15min cook-off

### DIFF
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -16,7 +16,7 @@
  * Public: No
  */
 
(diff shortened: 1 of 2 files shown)


PabstMirror:
we could also just do this without the changes to `params` (just keep the `min`) for now
because I think we'll want to come in at some point and take a deeper look at ammo cooking time

PabstMirror (approved):

jonpas (approved):

PR #8652, opened 2021-10-29
Cyrus magazine incorrectly configured as a belt

ZluskeN:
See Title, the Cyrus magazine for some reason inherits from the HK121 belt box


PabstMirror (approved):
that is some weird inheritance

PR #8659, opened 2021-11-01
Common - Block radio status effect without blocking direct voice (ACRE)

PabstMirror:
ref https://github.com/IDI-Systems/acre2/pull/1159 (will need to wait for next acre release)


addons/common/XEH_postInit.sqf
```diff
@@ -82,7 +82,7 @@
         _object setVariable ["tf_unable_to_use_radio", _set > 0, true];
     };
     if (isClass (configFile >> "CfgPatches" >> "acre_main")) then {
-        _object setVariable ["acre_sys_core_isDisabled", _set > 0, true];
+        _object setVariable ["acre_sys_core_unableToUseRadio", _set > 0, true];
```
PabstMirror, line 85:
```suggestion
        _object setVariable ["acre_sys_core_isDisabledRadio", _set > 0, true];
```

jonpas (approved):

jonpas:
Merged in ACRE2.

Slatery:
would just changing the variable here not block voices for unconscious players? If the variable name is to be believed it won't stop players from talking while unconscious wouldn't it?

Salluci:
No. This is just the `blockRadio` status effect, uncon voice control is handled elsewhere (`setVolume`).

jonpas:
Merge or wait for ACRE2 to have it in a release?

PabstMirror:
this is really subjective
I think it's worse to not be able to say anything than for a captive to be able to use the radio
but it's probably fine to wait for next acre

Slatery:
> No. This is just the `blockRadio` status effect, uncon voice control is handled elsewhere (`setVolume`).

had a play around yesterday (not realizing the ACRE change needed to go through) and ran into the issue of players being able to talk while unconscious, the setVolume function handles incoming sounds not outgoing. 
https://github.com/acemod/ACE3/blob/48912be73cd1610f4fb41fc37db485f3facfbde5/addons/medical_status/functions/fnc_setUnconsciousState.sqf#L35 is what's used to prevent talking, which uses this does it not? I can't find anywhere in the ACRE codebase either that handles the ACE unconscious status

veteran29 (approved):

Salluci:
> > No. This is just the `blockRadio` status effect, uncon voice control is handled elsewhere (`setVolume`).
> 
> had a play around yesterday (not realizing the ACRE change needed to go through) and ran into the issue of players being able to talk while unconscious, the setVolume function handles incoming sounds not outgoing.
> 
> https://github.com/acemod/ACE3/blob/48912be73cd1610f4fb41fc37db485f3facfbde5/addons/medical_status/functions/fnc_setUnconsciousState.sqf#L35
> is what's used to prevent talking, which uses this does it not? I can't find anywhere in the ACRE codebase either that handles the ACE unconscious status

See `medical_feedback` component. That's where that bit is handled.

PR #8669, opened 2021-11-07
Adv Fatigue - Make stamina penalty for weapon raised/ready scale with weapon inertia

pterolatypus:
- Title
- Makes the penalty apply to secondary weapons too

A setup with a full-length rifle, medium-range optic and flash hider has inertia around 0.9-1 so the stamina drain will be about the same in that case.
Gives lighter weapons an extra edge in CQB, as you can move faster while keeping them at the ready.


Salluci:
Why not cache weapons and attachments separately instead of their combinations?

pterolatypus:
> 
> 
> Why not cache weapons and attachments separately instead of their combinations?

I expect people to switch weapons more often than they change their attachments, so that's what I'm optimising for.
I could do both, but it means another hashmap lookup and condition block to save a getNumber which is already only reached when you switch attachments, so I thought it's probably not worth it.

addons/advanced_fatigue/functions/fnc_getWeaponInertia.sqf
```diff
@@ -0,0 +1,36 @@
...
+ * [ACE_player] call ace_advanced_fatigue_fnc_getWeaponInertia
+ *
+ * Public: No
+ */
+params [["_unit", ACE_player, [objNull]]];
+
+private _cache = GVAR(inertiaCache);
+private _weapon = currentWeapon _unit;
+private _accessories = _unit weaponAccessories _weapon;
+private _key = format (["%1$%2+%3+%4+%5", _weapon] + _accessories);
```
commy2, line 22:
```suggestion
private _key = str [_weapon, _accessories];
```

addons/advanced_fatigue/functions/fnc_getWeaponInertia.sqf
```diff
@@ -0,0 +1,36 @@
...
+
+private _inertia = _cache get _key;
+if (isNil "_inertia") then {
+    _inertia = 0;
+    {
+        if (isNumber (configFile >> "CfgWeapons" >> _x >> "inertia")) then {
+            _inertia = _inertia + getNumber (configFile >> "CfgWeapons" >> _x >> "inertia");
+        };
+    } forEach [_weapon] + _accessories;
```
commy2, line 31: I know + is raised priority from other binary commands, but you should still parantesis this one imo.

addons/advanced_fatigue/functions/fnc_getWeaponInertia.sqf
```diff
@@ -0,0 +1,36 @@
...
+private _cache = GVAR(inertiaCache);
+private _weapon = currentWeapon _unit;
+private _accessories = _unit weaponAccessories _weapon;
+private _key = format (["%1$%2+%3+%4+%5", _weapon] + _accessories);
+
+private _inertia = _cache get _key;
+if (isNil "_inertia") then {
+    _inertia = 0;
+    {
+        if (isNumber (configFile >> "CfgWeapons" >> _x >> "inertia")) then {
```
commy2, line 28: `isNumber` returns false for simple expressions, despite `getNumber` returning a NUMBER that can be != 0. This should also do check for `isText`. Consider putting the config path in a variable. Also consider putting `configFile >> "CfgWeapons"` in a higher scope variable.

Blutze: Can a CfgWeapons entry even be missing the inertia property in the first place?

commy2: You could probably make it so, but then the game likely would use 0. I think you're right and the check is completely unnecessary.

addons/advanced_fatigue/functions/fnc_getWeaponInertia.sqf
jonpas:
```suggestion
    } forEach ([_weapon] + _accessories);
```

addons/advanced_fatigue/functions/fnc_getWeaponInertia.sqf
pterolatypus: In fact, since this is a hashmap, it doesn't even need to be a string.

addons/advanced_fatigue/functions/fnc_getWeaponInertia.sqf
pterolatypus: Resolved as the + is now done elsewhere

addons/advanced_fatigue/functions/fnc_getWeaponInertia.sqf
pterolatypus: A handful of base classes have it undefined so it is possible. I wouldn't be too worried about it if it weren't for nil propagation - I'm a bit nervous about a modder screwing up their inheritance somehow breaking the whole of adv fatigue. This check also catches the case where `_x == ""`.

I don't know if this is allowed to be an expression (it never is in vanilla) but I've changed it to `getNumber` right off the bat and ignore if it's nil, which should cover all bases with fewer commands.

Salluci:
"Works on my machine"

commy2:
Bananenprinzip

LinkIsGrim:
> I'm interested in more feedback about this, atm I wonder if the scaling is a bit too much (stuff like MMGs can go up to like 2.5 inertia) so I'm considering using a fractional power, like `GVAR(inertia)^0.7` or something. Could also make it a setting.

Scaling should be fine. MMGs are hell to carry around, especially raised. Stamina consumption should reflect that.

LinkIsGrim (approved):
Still works on my machine.

PR #8599, opened 2021-10-24
Medical - Add option to prevent scripted injuries on invulnerable units

mjc4wilton:
**When merged this pull request will:**
- Add optional parameter (default: true for backward compatibility) to enable adding damage to invulnerable units.
- Add exit condition check to `ace_medical_fnc_addDamageToUnit`
- Resolve #8593

PR #8670, opened 2021-11-07
SOG Compat - Add `isDamageAllowed` check to punji damage handler

Zman6258:
- Title.

### DIFF
--- a/optionals/compat_sog/functions/fnc_handlePunjiTrapDamage.sqf
+++ b/optionals/compat_sog/functions/fnc_handlePunjiTrapDamage.sqf
@@ -19,7 +19,7 @@ params ["_trap"];
 if (isNull (configFile >> "CfgPatches" >> "ace_medical")) exitWith {};
 
 private _radius = getNumber (configOf _trap >> "indirectHitRange");
-private _affectedUnits = (_trap nearEntities ["CAManBase", _radius]) select {local _x};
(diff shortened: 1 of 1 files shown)


optionals/compat_sog/functions/fnc_handlePunjiTrapDamage.sqf
```diff
@@ -19,7 +19,7 @@ params ["_trap"];
 if (isNull (configFile >> "CfgPatches" >> "ace_medical")) exitWith {};
 
 private _radius = getNumber (configOf _trap >> "indirectHitRange");
-private _affectedUnits = (_trap nearEntities ["CAManBase", _radius]) select {local _x};
+private _affectedUnits = (_trap nearEntities ["CAManBase", _radius]) select {local _x && isDamageAllowed _x};
```
commy2, line 22: Maybe "lazy select", i.e. 
```suggestion
private _affectedUnits = (_trap nearEntities ["CAManBase", _radius]) select {local _x} select {isDamageAllowed _x};
```

Zman6258: Oh cool, I never thought of doing it like that

veteran29:
Could use the new parameter introduced in #8599 instead of the filter.

PabstMirror:
right now 8599 throws an `ERROR_1` (we might want to change?), 
but I still think it's better to handle in the calling function like this PR is doing

PabstMirror (approved):

PR #8671, opened 2021-11-07
Medical Engine - Synchronize random unconscious anims

PabstMirror:
Semi related to PR #8623

Makes all clients pick the exact same animation instead of purely random.
The actual randomness is constant per unit and it only has 64 possible values so the distribution won't be perfect, 
but I think it's more than good enough for this application.

### DIFF
--- a/addons/medical_engine/functions/fnc_applyAnimAfterRagdoll.sqf
+++ b/addons/medical_engine/functions/fnc_applyAnimAfterRagdoll.sqf
@@ -17,13 +17,16 @@
  */
 
 params ["_unit", "_anim"];
-TRACE_2("applyAnimAfterRagdoll",_unit,_unconsciousAnimation);
+TRACE_2("applyAnimAfterRagdoll",_unit,_anim);
 
 if !(IS_UNCONSCIOUS(_unit) &&                   // do not run if unit is conscious
     {alive _unit &&                             // do not run if unit is dead
     {isNull objectParent _unit}}) exitWith {};  // do not run if unit in any vehicle
 
-private _unconsciousAnimation = selectRandom (GVAR(animations) getVariable [_anim, [""]]);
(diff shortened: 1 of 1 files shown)


PabstMirror:
Without this pr, results of AI going unconc from perspective of 2 different players

before

jonpas (approved):

addons/medical_engine/functions/fnc_applyAnimAfterRagdoll.sqf
```diff
@@ -17,13 +17,16 @@
...
 params ["_unit", "_anim"];
-TRACE_2("applyAnimAfterRagdoll",_unit,_unconsciousAnimation);
+TRACE_2("applyAnimAfterRagdoll",_unit,_anim);
 
 if !(IS_UNCONSCIOUS(_unit) &&                   // do not run if unit is conscious
     {alive _unit &&                             // do not run if unit is dead
     {isNull objectParent _unit}}) exitWith {};  // do not run if unit in any vehicle
 
-private _unconsciousAnimation = selectRandom (GVAR(animations) getVariable [_anim, [""]]);
+private _animsArray = (GVAR(animations) getVariable [_anim, [""]]);
```
mharis001, line 26:
```suggestion
private _animsArray = GVAR(animations) getVariable [_anim, [""]];
```

addons/medical_engine/functions/fnc_applyAnimAfterRagdoll.sqf
```diff
@@ -17,13 +17,16 @@
...
+private _random = ((toArray (hashValue _unit)) param [0, 0]);
```
mharis001, line 27:
```suggestion
private _random = (toArray (hashValue _unit)) param [0, 0];
```

mharis001 (approved):

PR #8672, opened 2021-11-08
Overheating - Use `select` instead of `params`

Drofseh:
- switch from params to select for single value returns when calling getWeaponData


Drofseh:
@PabstMirror @jonpas

jonpas (approved):

PabstMirror (approved):

PR #8674, opened 2021-11-10
Map Tools - Add Vernier Scale and Improve resolution

LorenLuke:
- Adds a Vernier scale (4 mils) to the maptools, allowing visual use for slightly more accurate reading of milliradian directions.


veteran29:
Needs to make sure we're using only the highest mip of the texture to prevent blurry maptools on lower texture settings. (https://github.com/arma3/MipMapTool)

bux:
Could you add an example image to the description please :)

LorenLuke:
Worth noting, the actual point of rotation is ever so slightly offset, so sometimes the scale will not align exactly as it should, but it still will be read within +/-3 mils or so. Which is no more than 12 meters error at 4km, and still aligns to compass fine.

Center section (basically identical to both maptools):
Center
Large map tool oriented north:
Large
Small maptool oriented north:
Small
5468:

3542:

1652:

4748:

bux:
Could you try using another color than cyan. Imho it is hard to read, especially on the gray background.

LorenLuke:
@bux 
> Could you try using another color than cyan. Imho it is hard to read, especially on the gray background.

Are you referring to the numbers or the lines, or both? I personally fine the lines fine, and would like to keep them that color, as it contrasts to the white, the black, and the red, but if you want to have the numbers changed, I could do that easily (my only concern with those small numbers being black is associating them with the cyan lines). 

If you have another color you believe to work better for either the numbers or the lines, I'm more than willing to hear what you have in mind.

Drofseh:
I really don't like the cyan colour either. I think we should stay with red or black, probably red in this case.

LorenLuke:
Not committed at time of posting, but... here's it in red, and not on the VR map.

1874:

2686:

5318:

TymoteuszKazubski:
Red looks way more readable IMHO.

LorenLuke:
Ready in red.

dedmen:
The texture is just 2k, I think it might make sense here to go 4k for this?

LorenLuke:
@dedmen 
> 4k
While I don't disagree, my general workflow has been 
+ paa to png
+ edit in photoshop
+ save as interlaced png
+ img to paa

Because of this I didn't want to mess with scaling at all in case it threw it off, and am unsure how to go about making a higher resolution image while guaranteeing the whole thing isn't just going to be twice as big in scale as the original.

LinkIsGrim (approved):

PR #8676, opened 2021-11-11
Finger - Add scaling settings

Drofseh:
- add two settings to scale the size of the visual marker; `sizeCoef` will scale the size directly from the setting value, default 1x is the same as the current size; `proximityScaling` will scale the size based on the distance between the observer and the pointer, 2x the size at `maxRange`, 0.25x at 0m, interpolated in between.

### DIFF
--- a/addons/finger/functions/fnc_perFrameEH.sqf
+++ b/addons/finger/functions/fnc_perFrameEH.sqf
@@ -21,7 +21,7 @@ if !([ACE_player, ACE_player, ["isNotInside", "isNotSwimming"]] call EFUNC(commo
 // Make sure player is dismounted or in a static weapon:
 if ((ACE_player != vehicle ACE_player) && {!((vehicle ACE_player) isKindOf "StaticWeapon")}) then {GVAR(fingersHash) = createHashMap};
 
-private _iconBaseSize = GVAR(size) * BASE_SIZE * 0.10713 * (call EFUNC(common,getZoom));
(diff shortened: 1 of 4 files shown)


Salluci:
Nice workaround with `_sourceUnit`.

addons/finger/functions/fnc_perFrameEH.sqf
```diff
@@ -34,6 +34,11 @@ private _iconSize = BASE_SIZE * 0.10713 * (call EFUNC(common,getZoom));
         // Fade out:
         _drawColor set [3, ((_drawColor select 3) * ((_timeLeftToShow min 0.5) / 0.5))];
 
+        private _iconSize = _iconBaseSize;
+        if (GVAR(proximityScaling)) then {
```
PabstMirror, line 38: I think we could get away with just a single setting?
```suggestion
        if (GVAR(sizeCoef) != 1) then {
```

Drofseh: I want people to be able to set a custom scale if they want, and also to enable/disable distance based scaling

PabstMirror: nm I guess they do different things

Drofseh:
[Video of prox scaling](https://i.gyazo.com/415bfda72be4a95be108155501fd93db.mp4)

PabstMirror (approved):

PR #8685, opened 2021-11-16
Cookoff- Adds heating effect to cookoff

LorenLuke:
- Make tanks that cook-off actually heat up on IR.

### DIFF
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -74,6 +74,8 @@ if (_smokeDelayEnabled) then {
             };
         };
 
+        private _tiPars = getVehicleTIPars _vehicle; 
+
         private _lastFlameTime = _vehicle getVariable [QGVAR(lastFlame), 0];
         private _nextFlameTime = _vehicle getVariable [QGVAR(nextFlame), 0];
 
@@ -122,5 +124,8 @@ if (_smokeDelayEnabled) then {
                 _vehicle setVariable [QGVAR(nextExplosiveDetonation), random 60];
             };
         };
+        
+        _vehicle setVehicleTiPars [((_tiPars select 0) + (_intensity * 0.01))/1.005, ((_tiPars select 1) + (_intensity * 0.004))/1.002, ((_tiPars select 2) + (_intensity * 0.01)/1.005)];
+        
     }, 0.25, [_vehicle, _positions, _ammoDetonationChance, _detonateAfterCookoff, _instigator, _fireSource, _canRing, _smokeEffects]] call CBA_fnc_addPerFrameHandler
 }, [_vehicle, _positions, _intensity, _ammoDetonationChance, _detonateAfterCookoff, _instigator, _fireSource, _canRing], _delay] call CBA_fnc_waitAndExecute;


PabstMirror:
https://community.bistudio.com/wiki/setVehicleTIPars says local effect
so I'm not sure if this will sync correctly in mp?

LorenLuke:
@PabstMirror 
>Local Effect

It seems the whole thing is called and then immediately checked for locality, so I don't suspect it'll be an issue.

PabstMirror:
Untitled

LorenLuke:
I'm unsure what's going on here then, as the code states that the script shouldn't even start unless the vehicle is local. Does it still do all the smoking and flame effects otherwise? 

And, if so, I know it's been said to avoid `remoteExec`. What about `remoteExecCall` or `BIS_fnc_MP`?

jonpas:
No, use CBA Custom Events for any network "execution".

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -122,5 +124,8 @@ if (_smokeDelayEnabled) then {
                 _vehicle setVariable [QGVAR(nextExplosiveDetonation), random 60];
             };
         };
+        
+        [_vehicle, [((_tiPars select 0) + (_intensity * 0.01))/1.005, ((_tiPars select 1) + (_intensity * 0.004))/1.002, ((_tiPars select 2) + (_intensity * 0.01)/1.005)]] call CBA_fnc_globalEvent;
```
commy2, line 128: Seems wrong - no event name.
Also, please add some line breaks. This is almost 200 chars long...

pterolatypus:
>the script shouldn't even start unless the vehicle is local

Do it in `fnc_cookOffEffect` instead

TheCandianVendingMachine:
This sends an event across the network every 1/4 second. in `fnc_cookOffEffect.sqf` it would be called locally. Shove `_intensity` somewhere in there and move the code there

addons/cookoff/functions/fnc_cookOffEffect.sqf
```diff
@@ -183,5 +184,16 @@ if (isServer) then {
...
+    [QGVAR(setTIPars), [
+        _obj, [
+            ((_tiPars select 0) + (_intensity * 0.01))/1.005, 
+            ((_tiPars select 1) + (_intensity * 0.004))/1.002, 
+            ((_tiPars select 2) + (_intensity * 0.01))/1.005
+        ]
+    ]] call CBA_fnc_globalEvent;
+
+    
 }, 0, [_obj, _jet, _ring, _time, CBA_missionTime, _light, _fireSelection, _sound]] call cba_fnc_addPerFrameHandler;
```
pterolatypus, line 198:
```suggestion
}, 0, [_obj, _jet, _ring, _time, CBA_missionTime, _light, _fireSelection, _sound, _intensity]] call cba_fnc_addPerFrameHandler;
```

LorenLuke: I swear I put it in there... guess this is what I get for trying to use both desktop git and the online editor for the same thing.

addons/cookoff/functions/fnc_cookOffEffect.sqf
```diff
@@ -183,5 +184,16 @@ if (isServer) then {
...
-}, 0, [_obj, _jet, _ring, _time, CBA_missionTime, _light, _fireSelection, _sound]] call cba_fnc_addPerFrameHandler;
+    
+    private _tiPars = getVehicleTIPars _obj; 
+    [QGVAR(setTIPars), [
+        _obj, [
+            ((_tiPars select 0) + (_intensity * 0.01))/1.005, 
+            ((_tiPars select 1) + (_intensity * 0.004))/1.002, 
+            ((_tiPars select 2) + (_intensity * 0.01))/1.005
+        ]
+    ]] call CBA_fnc_globalEvent;
```
commy2, line 195: This raises a global events from globally executed script I think.

commy2:
What's the logic behind the math here? Refering to this part:
```
            ((_tiPars select 0) + (_intensity * 0.01))/1.005, 
            ((_tiPars select 1) + (_intensity * 0.004))/1.002, 
            ((_tiPars select 2) + (_intensity * 0.01))/1.005
```
Lots of magic hard coded numbers here.

addons/cookoff/functions/fnc_cookOffEffect.sqf
TheCandianVendingMachine: Yeah, this should be called locally now. This function is run per client

LorenLuke:
> magic numbers

The center value is the tracks/tires, while the other two are the engine and the cannon barrel, so at the outset it stands to reason that the thing having smoke exit from it and the thing with all the ammo burning inside of it would heat up faster than the tires/treads, so that's why the first and last values are equally heated up while the second one heats up more slowly.

The second half of the numbers is so as to 1) not be immediately just set to max, 2) not have a linear increase in the values as the heating would taper off, instead having a curve as it gets closer and closer to max, and 3) the math (by always going slightly more than half the distance with fixed values) allows for it to actually hit a maximum level, rather than just only get asymptotically close. 

The numbers don't correlate to any real world data other than 'I think when it's belching fire, it should at least look like it has been warmed up some' followed by 'I'm sure it'll be glowing on sensors in short order', and how I feel would best correlate to how quickly a tank would heat up by cooking off.

addons/cookoff/functions/fnc_cookOffEffect.sqf
```diff
@@ -183,5 +184,13 @@ if (isServer) then {
...
         ];
     };
-}, 0, [_obj, _jet, _ring, _time, CBA_missionTime, _light, _fireSelection, _sound]] call cba_fnc_addPerFrameHandler;
+    
+    private _tiPars = getVehicleTIPars _obj; 
+    _obj setVehicleTIPars [
+        ((_tiPars select 0) + (_intensity * 0.01))/1.005, 
+        ((_tiPars select 1) + (_intensity * 0.004))/1.002, 
+        ((_tiPars select 2) + (_intensity * 0.01))/1.005
+    ];
```
veteran29, line 193: `vectorAdd` will be faster than doing `select` three times.
```suggestion
    _obj setVehicleTIPars (_tiPars vectorAdd [
        (_intensity * 0.01) / 1.005, 
        (_intensity * 0.004) / 1.002, 
        (_intensity * 0.01) / 1.005
    ]);
```

addons/cookoff/functions/fnc_cookOffEffect.sqf
```diff
@@ -183,5 +184,13 @@ if (isServer) then {
...
+    
+}, 0, [_obj, _jet, _ring, _time, CBA_missionTime, _light, _fireSelection, _sound, _intensity]] call cba_fnc_addPerFrameHandler;
```
veteran29, line 195:
```suggestion
}, 0, [_obj, _jet, _ring, _time, CBA_missionTime, _light, _fireSelection, _sound, _intensity]] call CBA_fnc_addPerFrameHandler;
```

veteran29 (changes requested):

addons/cookoff/functions/fnc_cookOffEffect.sqf
```diff
@@ -183,5 +184,13 @@ if (isServer) then {
             [2 + random 1], 1, 0, "", "", _obj
         ];
     };
-}, 0, [_obj, _jet, _ring, _time, CBA_missionTime, _light, _fireSelection, _sound]] call cba_fnc_addPerFrameHandler;
+    
+    private _tiPars = getVehicleTIPars _obj; 
+    _obj setVehicleTIPars [
```
TheCandianVendingMachine, line 189: Please add comments as to what the magic numbers indicate

TheCandianVendingMachine (changes requested):

addons/cookoff/functions/fnc_cookOffEffect.sqf
PabstMirror: math is different
`(current + add) / 1.05` 
vs
`current + (add/ 1.05)`

PabstMirror (approved):

PR #8704, opened 2021-11-25
Misc - Cleanup unused namespace vars

PabstMirror:


mharis001 (approved):

addons/rearm/XEH_postInit.sqf
```diff
@@ -1,6 +1,5 @@
 #include "script_component.hpp"
 
-GVAR(hardpointGroupsCache) = [] call CBA_fnc_createNamespace;
 GVAR(configTypesAdded) = [];
```
commy2, line 3: Should rather replace this array with a HashMap. Would make these lookups O(1) instead of O(n).
https://github.com/acemod/ACE3/blob/master/addons/rearm/functions/fnc_initSupplyVehicle.sqf#L32-L33

Issue #8715, opened 2021-11-29
Setting To Completely Destroy Vehicles After Cooking-Off/Burning


LuizBarros99:
**Is your enhancement related to a problem?**
In a way it is yes related to a problem. In short the problem is the possibility to easily repair vehicles that have cooked-off/burnt (after the cook-off/burning ends)

**Solution you'd like:**
My suggestion is to completely destroy the vehicle after/once it cooks-off/burns. This would prevent the vehicle from being easily repaired.

**Alternatives you've considered:**
Some alternatives that I have considered relate to WHEN to completely destroy the vehicle. Alternatives are at the start of the cook-off, at the middle of cook-off, at the end of cook-off. These different alternatives could be considered to simulate the vehicle being damaged over time from burning/cooking-off, or even to simulate a sudden catastrophic ammo detonation.

**Additional context:**
This would simulate the vehicle being damaged beyond repair when cooking-off/burning, and would prevent it from being easily repaired.

ZluskeN:
there is already a variable called `_detonateAfterCookoff` here: [https://github.com/acemod/ACE3/blob/9c946727cce62913cf3ab0eb59e1b9521dbdd99d/addons/vehicle_damage/functions/fnc_handleCookoff.sqf#L34](url)
However it seems the probability is too low to ever realistically happen. A good way to improve this could be to make this probability approach 1 with subsequent cookoffs (or even consecutive hits). 

Additionally, this line [https://github.com/acemod/ACE3/blob/f75bfd4cef3617c9de45b274e82a9841ed293f9f/addons/cookoff/functions/fnc_cookOff.sqf#L73](url) could be altered to use alternative syntax `setDamage [1, false];` to destroy the vehicle without triggering a Catastrophic kill effect if there is no ammo left to cook off/explode

RufusVulpecula:
So, what would the line _detonateAfterCookoff needs to be in order for the cook-off to always result in a catastrophic kill?

ZluskeN:
Well, you could just change it to `private _detonateAfterCookoff = true;` but that means the `detonationDuringFireProb` config entry of the vehicles will be bypassed entirely. Maybe this should be a user setting?

RufusVulpecula:
I can try that, thanks. I believe this as well as a cumulative damage threshold for instant catastrophic kill should be an option. I admire the system for what it's trying to achieve but I don't think this implementation is the best way to go.

LuizBarros99:
[...]

PR #8734, opened 2021-12-10
Cookoff - Add setting to destroy vehicle after cooking off

ZluskeN:
- Add a setting to always destroy vehicles after cooking off
- The vehicle will be destroyed without explosion
- Catastrophic kill effect will trigger if `_detonateAfterCookoff` is true
- Addresses https://github.com/acemod/ACE3/issues/8715

### DIFF
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -69,8 +69,8 @@ if (_smokeDelayEnabled) then {
             _vehicle setVariable [QGVAR(isCookingOff), false, true];
             [_pfh] call CBA_fnc_removePerFrameHandler;
 
-            if (_detonateAfterCookoff) then {
-                _vehicle setDamage 1;
+            if (GVAR(destroyVehicleAfterCookoff) || _detonateAfterCookoff) then {
+                _vehicle setDamage [1, _detonateAfterCookoff];
             };
         };
 
--- a/addons/cookoff/initSettings.sqf
+++ b/addons/cookoff/initSettings.sqf
@@ -19,6 +19,16 @@
 ] call CBA_fnc_addSetting;
 
 [
+    QGVAR(destroyVehicleAfterCookoff), "CHECKBOX",
+    [LSTRING(destroyVehicleAfterCookoff_name), LSTRING(destroyVehicleAfterCookoff_tooltip)],
+    LSTRING(category_displayName),
+    true, // default value
+    true, // isGlobal
+    {[QGVAR(destroyVehicleAfterCookoff), _this] call EFUNC(common,cbaSettings_settingChanged)},
+    false // Needs mission restart
+] call CBA_fnc_addSetting;
+
+[
     QGVAR(enableAmmoCookoff), "CHECKBOX",
     [LSTRING(enableAmmoCookoff_name), LSTRING(enableAmmoCookoff_tooltip)],
     LSTRING(category_displayName),
--- a/addons/cookoff/initSettings.sqf
+++ b/addons/cookoff/initSettings.sqf
@@ -22,7 +22,7 @@
     QGVAR(destroyVehicleAfterCookoff), "CHECKBOX",
     [LSTRING(destroyVehicleAfterCookoff_name), LSTRING(destroyVehicleAfterCookoff_tooltip)],
     LSTRING(category_displayName),
-    true, // default value
+    false, // default value
     true, // isGlobal
     {[QGVAR(destroyVehicleAfterCookoff), _this] call EFUNC(common,cbaSettings_settingChanged)},
     false // Needs mission restart
--- a/addons/cookoff/stringtable.xml
+++ b/addons/cookoff/stringtable.xml
@@ -163,7 +163,7 @@
         <Key ID="STR_ACE_CookOff_destroyVehicleAfterCookoff_name">
             <English>Destroy Vehicles After Cook-Off</English>
         </Key>
-        <Key ID="STR_ACE_CookOff_destroyVehicleAfterCookoff_tooltip	">
+        <Key ID="STR_ACE_CookOff_destroyVehicleAfterCookoff_tooltip">
             <English>Vehicles will be destroyed after cooking off</English>
         </Key>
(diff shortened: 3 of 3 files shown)


veteran29 (approved):

Drofseh:
lol I was litterally about to PR exactly this change.

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -69,8 +69,8 @@ if (_smokeDelayEnabled) then {
             _vehicle setVariable [QGVAR(isCookingOff), false, true];
             [_pfh] call CBA_fnc_removePerFrameHandler;
 
-            if (_detonateAfterCookoff) then {
-                _vehicle setDamage 1;
+            if (GVAR(destroyVehicleAfterCookoff) || _detonateAfterCookoff) then {
+                _vehicle setDamage [1, _detonateAfterCookoff];
```
TheCandianVendingMachine, line 73: When detonating via variable, effects are desired

TheCandianVendingMachine (changes requested):

TheCandianVendingMachine (approved):

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -69,8 +69,8 @@ if (_smokeDelayEnabled) then {
             _vehicle setVariable [QGVAR(isCookingOff), false, true];
             [_pfh] call CBA_fnc_removePerFrameHandler;
 
-            if (_detonateAfterCookoff) then {
-                _vehicle setDamage 1;
+            if (GVAR(destroyVehicleAfterCookoff) || _detonateAfterCookoff) then {
```
rautamiekka, line 72: How about this ... ?
`{_detonateAfterCookoff}`

Granted, I dunno the language well enough to know if it matters.

TheCandianVendingMachine: No I was wrong, I changed my review. You're good with `setDamage [1, _detonateAfterCookoff]`

Drofseh (changes requested):

Drofseh (approved):

addons/cookoff/stringtable.xml
```diff
@@ -160,6 +160,12 @@
             <Portuguese>Multiplicador para a chance de cozinhamento. Valores mais altos aumentam as chances de ocorrer.</Portuguese>
             <Czech>Multiplikátor pro pravděpodobnost vznícení munice. Vyšší hodnota znamená vyšší šanci vznícení munice.</Czech>
         </Key>
+        <Key ID="STR_ACE_CookOff_destroyVehicleAfterCookoff_name">
+            <English>Destroy Vehicles After Cook-Off</English>
```
mharis001, line 164:
```suggestion
            <English>Destroy Vehicles After Cook-off</English>
```
To match above. It is also written without hyphen.

addons/cookoff/stringtable.xml
```diff
@@ -160,6 +160,12 @@
...
+        </Key>
+        <Key ID="STR_ACE_CookOff_destroyVehicleAfterCookoff_tooltip">
+            <English>Whether or not vehicles will always be destroyed after cooking off. Depending on the cookoff the vehicle may explode, or may simply be killed without effects</English>
```
mharis001, line 167:
```suggestion
            <English>Controls whether vehicles will always be destroyed after cooking off. Depending on the cookoff, the vehicle may explode or may simply be killed without effects.</English>
```

addons/cookoff/initSettings.sqf
```diff
@@ -18,6 +18,16 @@
     false // Needs mission restart
 ] call CBA_fnc_addSetting;
 
+[
+    QGVAR(destroyVehicleAfterCookoff), "CHECKBOX",
+    [LSTRING(destroyVehicleAfterCookoff_name), LSTRING(destroyVehicleAfterCookoff_tooltip)],
+    LSTRING(category_displayName),
+    true, // default value
```
TheCandianVendingMachine, line 25: Default false. Intended behaviour is for them to remain there not blown up

TheCandianVendingMachine (dismissed):

addons/cookoff/initSettings.sqf
ZluskeN: Changed it 👍

Drofseh:
imo having tile default as false is a mistake. 

Not having vehicles die breaks any currently existing script that counts vehicle destruction, this is a huge change from expected behaviour and has been one of the main complaints I've seen people have about this component.

addons/cookoff/initSettings.sqf
ZluskeN: > imo having tile default as false is a mistake.
> 
> Not having vehicles die breaks any currently existing script that counts vehicle destruction, this is a huge change from expected behaviour and has been one of the main complaints I've seen people have about this component.

Well I agree with this, otherwise I wouldn't have made this PR but I can't call myself an authority here. @TheCandianVendingMachine How do we resolve this?

veteran29: I agree with @Drofseh, the current default behaviour is a big change from vanilla game default behaviour and I believe it should be opt-in instead of opt-out.

veteran29: asked @acemod/maintainers for opinions.

veteran29: Maintainers have agreed the setting should remain off by default so the system works as it was designed, however, due to a big change from vanilla behaviour we want Advanced Vehicle Damage disabled by default.

Disabling of the system by default should be done in separate PR so it will be reflected in changelog.

veteran29:
> imo having tile default as false is a mistake.
> 
> Not having vehicles die breaks any currently existing script that counts vehicle destruction, this is a huge change from expected behaviour and has been one of the main complaints I've seen people have about this component.

See: https://github.com/acemod/ACE3/pull/8734#discussion_r787918244

addons/cookoff/initSettings.sqf
severgun: Can't get why after cookoff and full burnout vehicle should be not wrecked and even repairable?

PR #8735, opened 2021-12-11
Common - Add keybind and function to unload unit's weapon/muzzle

Drofseh:
- add keybind and function to unload weapon/muzzle of unit
- requires https://github.com/CBATeam/CBA_A3/pull/1527

**To Do:**

- [x] document new weapon config entries, `ACE_unloadAction`, `ACE_unloadSound`. I'm not sure which framework doc to put this in.
- [x] see if there's a way to interupt the reload animation if there is no `ACE_unloadAction`


Drofseh:
I haven't been able to find a way to interupt the default reload animation, playing another gesture (or no gesture) doesn't seem to do it.
However, even when the full reload animation is played, because the magazine disappears in the middle it looks like they are just clearing the chamber, so I'm happy with it as is.

Drofseh:
This is ready for review now that https://github.com/CBATeam/CBA_A3/pull/1527 has been merged

addons/common/functions/fnc_unloadUnitWeapon.sqf
```diff
@@ -0,0 +1,87 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+* [ACE_player, currentWeapon ACE_player, currentMuzzle ACE_player, 23, false] call ace_common_fnc_unloadUnitWeapon
+ *
+ * Public: No
+ */
+
+params ["_unit", "_weapon", ["_muzzle", _weapon], ["_ammoCount", ammo _muzzle _unit], ["_skipAnim", false]];
```
PabstMirror, line 22:
```suggestion
params ["_unit", "_weapon", ["_muzzle", _weapon], ["_ammoCount", _unit ammo _muzzle ], ["_skipAnim", false]];
```

PabstMirror:
I can mash the button and multiple sound effects play
but it doesn't dupe ammo so it's probably not a problem

PabstMirror (approved):

PabstMirror:
also maybe this should be in ace_reload?

jonpas:
I think I recommended `common` as I saw potential uses for it in other modules.

docs/wiki/framework/common-framework.md
```diff
@@ -0,0 +1,26 @@
...
+layout: wiki
+title: Common Framework
+description: Notes on ACE3 Common.
+group: framework
+order: 5
+parent: wiki
+mod: ace
+version:
+  major: 3
+  minor: 15
```
PabstMirror, line 11:
```suggestion
  minor: 14
```

docs/wiki/framework/common-framework.md
```diff
@@ -0,0 +1,26 @@
...
+  patch: 0
```
PabstMirror, line 12:
```suggestion
  patch: 2
```

PabstMirror (approved):

PR #8751, opened 2021-12-22
Cookoff - Minor improvement to Fire integration

LinkIsGrim:
- Change cookoff fire source key to `hashValue _object + "cookoff"` so both cookoff fire effect and other burn effects using vehicle as key can coexist.
- Change one instance of `CBA_fnc_getObjectConfig` to `configOf` directly.
- Change local/global events to serverEvent (readability).

### IMPORTANT


addons/fire/XEH_postInit.sqf
```diff
@@ -13,6 +13,19 @@
...
     if (!GVAR(enabled)) exitWith {};
 
+    ["AllVehicles", "Killed", {
+        params ["_vehicle", "_killer"];
+
+        if !(local _vehicle) exitWith {};
+
+        if (getText (configOf _vehicle >> "destrType") == "") exitWith {};
+
+        private _radius = 1.5 * ((boundingBoxReal _vehicle) select 2);
```
TheCandianVendingMachine, line 23: Radius should probably be the smallest axis in the xy-plane * 1.5 for gameplay reasons

addons/fire/XEH_postInit.sqf
```diff
@@ -13,6 +13,19 @@
...
+        private _intensity = (fuel _vehicle) * 10;
+
+        [QGVAR(addFireSource), [_vehicle, _radius, _intensity, _vehicle, {fuel (_this select 0) != 0}, [_vehicle]]] call CBA_fnc_serverEvent;
```
TheCandianVendingMachine, line 26: Will the fuel of the vehicle go down as it's destroyed?

addons/fire/XEH_postInit.sqf
```diff
@@ -13,6 +13,19 @@
     TRACE_1("settingsInit", GVAR(enabled));
     if (!GVAR(enabled)) exitWith {};
 
+    ["AllVehicles", "Killed", {
+        params ["_vehicle", "_killer"];
+
+        if !(local _vehicle) exitWith {};
+
+        if (getText (configOf _vehicle >> "destrType") == "") exitWith {};
```
TheCandianVendingMachine, line 21: I think this will also add a fire source for destroyed buildings. Probably best to explicitly define the vehicles which this happens rather than assume that you will get all of the edge cases

TheCandianVendingMachine (dismissed):

addons/fire/XEH_postInit.sqf
Salluci: Yes.

addons/fire/XEH_postInit.sqf
Salluci: I'm worried about that implementation for things like the HEMTT. One side is much longer to the point where touching the back of the vehicle might not burn the unit due to the distance from the model center.

Salluci:
Found a better implementation for burning units via medical engine side. Removing that bit.

addons/fire/XEH_postInit.sqf
TheCandianVendingMachine: I don't know, I thought it may look goofy if you aren't near the fire and you still burn. Worst case we change it later

addons/cookoff/functions/fnc_cookOffEffect.sqf
```diff
@@ -37,7 +37,7 @@ if (isServer) then {
     if (_ring) then {
         private _intensity = 6;
         private _radius = 1.5 * ((boundingBoxReal _obj) select 2);
-        [QEGVAR(fire,addFireSource), [_obj, _radius, _intensity, _obj]] call CBA_fnc_localEvent;
+        [QEGVAR(fire,addFireSource), [_obj, _radius, _intensity, str _obj + "_cookoff"]] call CBA_fnc_serverEvent;
```
PabstMirror, line 40: Not sure, would this be any better?
I think technically a vic could be named differently of different machines.
```suggestion
        [QEGVAR(fire,addFireSource), [_obj, _radius, _intensity, hashValue _obj + "_cookoff"]] call CBA_fnc_serverEvent;
```

johnb432 (approved):
LGTM

BrettMayson (approved):

johnb432:
> Change local/global events to serverEvent (readability).
I missed that part @LinkIsGrim. I don't think this is necessary - is it something that you still think is worth it?

LinkIsGrim:
Check diff and see if that still applies, I don't think it does

johnb432:
> Check diff and see if that still applies, I don't think it does

No it doesn't apply anymore, because I unintentionally reverted it when merging main into this branch yesterday.
Imo it's best left this way (as in with `CBA_fnc_localEvent`), but if you insist for readability, I can understand that.

PR #8760, opened 2021-12-24
UI - Make Speed Indicator independent of 'Show Stance Info' difficulty setting

### LINKED ISSUE #8758: UI - Make Speed Indicator Independent of Stance Indicator
Speed Indicator added in 3.14 is a sub-control of the Stance Indicator, which means its visibility is subject to the mission's difficulty settings having the stance indicator enabled. This also results in the Speed Indicator not being shown unless there's been a recent stance change if the stance indicator is set to fade-out.

**Solution you'd like:**
Tie the Speed Indicator to another class of RscInGameUI, or make it its own class entirely.

It's possible to disable the Stance Indicator via Selective UI while having it enabled in difficulty settings. This doesn't account for the fade-out possibility.

**Additional context:**
N/A.

Salluci:
- Move speed indicator control into its own separate container, like Advanced Fatigue's Stamina Bar.
- Make speed indicator control persist on screen when opening the interaction or medical menus.
- Close #8758.

### IMPORTANT


veteran29 (changes requested):
As the indicator is not tied to unit info container now, the indicator will stay there when you enter the vehicle while moving.
image

addons/ui/functions/fnc_createSpeedIndicator.sqf
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Public: No
+ */
+params ["_display"];
+
+private _speedIndicator = _display ctrlCreate [QGVAR(SpeedIndicatorContainer), -1];
+uiNamespace setVariable [QGVAR(speedIndicatorContainer), _speedIndicator];
+
+_speedIndicator ctrlSetFade 1;
+_speedIndicator ctrlCommit 0;
```
veteran29, line 23: Is there a need to fade this at all? When the indicator is disabled the whole container is hidden anyway.

addons/ui/functions/fnc_handleSpeedIndicator.sqf
```diff
@@ -24,5 +24,8 @@ private _isProne = _animState select [5, 3] isEqualTo "pne";
 
 private _icon = GVAR(speedIndicatorIconHash) getOrDefault [[_animSpeed, _isProne], ""];
 
-private _speedIndicator = uiNamespace getVariable [QGVAR(speedIndicator), controlNull];
-_speedIndicator ctrlSetText _icon;
+private _speedIndicatorContainer = uiNamespace getVariable [QGVAR(speedIndicatorContainer), controlNull];
+(_speedIndicatorContainer controlsGroupCtrl 10) ctrlSetText _icon;
+
+_speedIndicatorContainer ctrlSetFade 0;
+_speedIndicatorContainer ctrlCommit 0
```
veteran29, line 31: I see no need to "unfade" the indicator on every animation change. As this code runs "all the time" due to changing animations I would prefer for it to be as fast as possible. 

Also I would suggest to store the indicator itself in the uiNamspace so there will be no need for `controlsGroupCtrl `.

veteran29 (changes requested):

Salluci:
Working on modular alternate solution. Closing.

PR #8768, opened 2022-01-09
Translation - Korean v2.0

jtj5283449:

- Will translate over 1000 lines which wasn't translated before
- Minor adjustments and synchronization between some gears and items name so it can go smooth with vanilla items (Needs more time to overhaul the translation)


PabstMirror:
I don't understand the diff, did the encoding change?

bux:
@jtj5283449 There's something wrong with the files you've committed. 
image
That's a lot of changed lines :)

jtj5283449:
my apologies, I don't know how to github and as you can tell I got a brain made with cinderblock. I simply uploaded xml files into my clone and requested merge. I guess I should've done something else

PabstMirror (approved):

PabstMirror:
Mass replaced \r\n with \n
probably happened because files were uploaded
imho this is not a problem we should have had to deal with, github.com should be smart enough to handle that

PR #8769, opened 2022-01-10
Overheating - Improve cookoff by using dummy to fire when forceWeaponFire doesn't work

ampersand38:
- Idea by @Drofseh 
- When player is sprinting or rifle is on back, rifle that is in cookoff doesn't fire with `forceWeaponFire`
- This solution uses a cutdown character model, created as agent, to hold a copy of the cooking-off-weapon, and attaches it to the weapon proxy on the player. The shot is fired from that weapon, and player weapon's ammo is decreased.
- Works when sprinting https://www.youtube.com/watch?v=DzgIGUXO0DY
- Works for rifle on back but not holstered pistol, which may require spawning a simpleObject model of the vest to find where the holstered-pistol proxy is. Not sure if it's worth it to go to the trouble though https://www.youtube.com/watch?v=DzgIGUXO0DY

To-Do
- [ ] Setup conditions to check when dummy is necessary https://github.com/ampersand38/ACE3/commit/217536cebeb41a9a81c81a4135f37b2764ff49ee#diff-587d0df0e254c8dc43facbef60db2b16de73494d011e8ea1934493126d82f86aR52
- [ ] Check if next round will be cooked off as well, to decide whether dummy should be deleted https://github.com/ampersand38/ACE3/commit/217536cebeb41a9a81c81a4135f37b2764ff49ee#diff-81276afcbfd5e758aae4c577cffbcb9e84931eb0ac8e0fe288a4be9bfa8f4d6eR43


Drofseh:
> Setup conditions to check when dummy is necessary

depending on how expensive it is, it might be best to use the dummy with every cook off.
The ammo temp loop runs once per second, so under normal circumstances there is only one cook off per second (per weapon) or less, and I'm really not sure how we could detect every animation that prevents `forceWeaponFire` from working

> Check if next round will be cooked off as well, to decide whether dummy should be deleted

`_isNextRoundCool = 770 < (GVAR(cookoffCoef) * (([player, primaryWeapon player] call ace_overheating_fnc_getWeaponTemperature) max ([player, handgunWeapon player] call ace_overheating_fnc_getWeaponTemperature)));`

`770` is the temperature where, with default settings, the heat transfers to the ammo so fast that the gun runs away and cooks off at cyclic rate. Multiplying that by the setting for the ammo temp required to cook off should roughly scale the point at which the gun runs away and the dummy should be kept.

PabstMirror:
What if dummy kills someone?
What happens if player disconnects after dummy is created?

PR #8779, opened 2022-01-16
Medical GUI - Improve descriptions for blood loss and damage color settings

mharis001:
- title


PabstMirror:
Do we need to wipe all the non-english translations?
It's just better phrasing but the meaning hasn't really changed

PabstMirror (approved):

veteran29 (approved):

PR #8786, opened 2022-01-22
Arsenal - Improve magazine removal after weapon switch

johnb432:
Improve the removal of magazines when switching weapons in the arsenal.

Current behaviour:
- When a new weapon is selected, only magazines defined in the "magazines" attribute of the old weapon's class were deleted.
See [here](https://github.com/acemod/ACE3/blob/936dd5d9786cd32b3061cf702f9ce45b45fa852a/addons/arsenal/functions/fnc_onSelChangedLeft.sqf#L32).
This leads to inconsistent and unnecessary magazine removal upon weapon switching.
- When a new weapon is selected, it would consume a magazine from the loadout.
See [here](https://github.com/acemod/ACE3/blob/936dd5d9786cd32b3061cf702f9ce45b45fa852a/addons/arsenal/functions/fnc_onSelChangedLeft.sqf#L62).

Improved behaviour:
- When a new weapon is selected, it will remove all magazines from the previous weapon that are not compatible with the new weapon.
See [here](https://github.com/johnb432/ACE3/blob/43494d24bec80cae911ccbca0e2699a6b065ffe6/addons/arsenal/functions/fnc_onSelChangedLeft.sqf#L32).
This is also valid when selecting no weapon: It will remove all magazines of the previous weapon.

On a personal note, I think the above can be seen as good and bad behaviour: When selecting launcher ammunition when you have a launcher equipped means that when you unequip the launcher in the arsenal, it will delete the launcher magazines.
However for other weapon types, especially side arms in my opinion, it can be very handy to remove all of the side arm magazines along with the unequipped weapon.
This behaviour can be circumvented by selecting the desired magazines via the all magazines tab in the arsenal, but it might pose an inconvenience for some.
Either way, I'm open for discussion - a CBA setting could be added if needed.
- When a new weapon is selected, it does not consume a magazine from the loadout.
See [here](https://github.com/johnb432/ACE3/blob/43494d24bec80cae911ccbca0e2699a6b065ffe6/addons/arsenal/functions/fnc_onSelChangedLeft.sqf#L68).

-----
What *hasn't* changed is if you have magazines from a weapon that isn't involved in switching process, it still won't affect those magazines.
E.g: If you have an RPG-7 and you switch your primary weapon (e.g. basegame HK416), it will not remove the RPG-7 rounds you have in your loadout.

### IMPORTANT

### DIFF
--- a/addons/arsenal/functions/fnc_onSelChangedLeft.sqf
+++ b/addons/arsenal/functions/fnc_onSelChangedLeft.sqf
@@ -29,10 +29,15 @@ private _containerDefaultRightPanel = _display displayCtrl IDC_buttonMisc;
 private _selectCorrectPanelContainer = [_containerDefaultRightPanel, _display displayCtrl GVAR(currentRightPanel)] select (!isNil QGVAR(currentRightPanel) && {GVAR(currentRightPanel) in [RIGHT_PANEL_ITEMS_IDCS]});
 
 private _fnc_clearPreviousWepMags = {
-    private _compatibleMags = getArray (configfile >> "cfgweapons" >> _baseWeapon >> "magazines");
+    private _compatibleMagsBaseWeapon = [_baseWeapon, true] call CBA_fnc_compatibleMagazines;
+
+    if (_item != "") then {
+        _compatibleMagsBaseWeapon = _compatibleMagsBaseWeapon select {!(_x in _compatibleMags)};
+    };
+
     {
         GVAR(center) removeMagazines _x;
-    } foreach _compatibleMags;
+    } foreach _compatibleMagsBaseWeapon;
 
     GVAR(currentItems) set [15, uniformItems GVAR(center)];
     GVAR(currentItems) set [16, vestItems GVAR(center)];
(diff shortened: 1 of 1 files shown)


Salluci (approved):

PabstMirror (approved):

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
```diff
@@ -29,10 +29,15 @@ private _containerDefaultRightPanel = _display displayCtrl IDC_buttonMisc;
 private _selectCorrectPanelContainer = [_containerDefaultRightPanel, _display displayCtrl GVAR(currentRightPanel)] select (!isNil QGVAR(currentRightPanel) && {GVAR(currentRightPanel) in [RIGHT_PANEL_ITEMS_IDCS]});
 
 private _fnc_clearPreviousWepMags = {
-    private _compatibleMags = getArray (configfile >> "cfgweapons" >> _baseWeapon >> "magazines");
+    private _compatibleMagsBaseWeapon = [_baseWeapon, true] call CBA_fnc_compatibleMagazines;
```
PabstMirror, line 32: this also gets mags from other muzzles (e.g. grenade launcher), which I think is what we want

PR #8788, opened 2022-01-24
Medical Statemachine - Improve edge case death handling

### LINKED ISSUE #8772: Controls & Respawn UI Disabled When Dead Body is Damaged
**Mods:**
- **Arma 3:** `2.06` stable
- **CBA:** `3.15.6` stable
- **ACE3:** `3.14.1` stable

**Description:**
What I call "the overkill bug": when a player is killed and if their body takes further damage before they can respawn, their whole respawn UI will be closed, and all controls (with the exceptions noted below) are disabled.  This can affect both the host and any clients / other players on the server.  Server can be either a listen server hosted from the main menu of Arma, or a dedicated server.

The only controls that do not seem to be disabled are:
"/" for text chat
"," and "." to change channel
"y" to open the Zeus interface*

*note that, since the player is dead, the Zeus interface will immediately close; this does not fix the bug.

The only known solution is ALT+F4 and restart the game, which is not ideal when hosting a listen server.

**Steps to reproduce:**
   1. Open the Eden editor and create a MP scenario featuring at least one playable unit and some way to deal damage after force-respawn or death (i.e. another player's gun [a] or fall damage from ejecting from a helicopter [b]).
   2. Make sure to enable respawns on custom positions, selectable spawn position & inventory, but disable respawn counter.
   3. Set the respawn time to something reasonable (eg. 5 seconds) and place a respawn point for the side of the playable units.
   4. Host the scenario, either as a listen server or as a dedicated server.
   5. Select a playable unit in the role assignment screen.
   6. Start the scenario.

   7a. In the case of using another player to cause damage, force-respawn.
   8a. While dead and in the respawn menu, have the other player shoot at your corpse (ideally the torso or head).

   7b. In the case of using fall damage to cause damage, board the helicopter and get some altitude and momentum.
   8b. Once you're high enough and moving fast enough to be sure you will take a fatal amount of fall damage, eject from the aircraft.
   9b. Immediately force-respawn before you hit the ground (while still high enough to inflict fall damage).
   10b. Wait in the respawn menu until your body smacks the ground.
   
   Result: you will be "overkilled".  Might not happen every single time, but _very_ frequently - you're _lucky_ if it doesn't happen.

**Expected behavior:**
Taking damage while dead should have no effect; you should remain in the respawn menu, all controls still unlocked, and be able to return to the main menu or respawn, oblivious to your body taking further damage.

- Dedicated
- Self-Hosted Multiplayer

**Log Files:**
- https://gist.github.com/Punchington/c659204e723a30fd5fdc2f1a8f7dccc3

**Additional context:**
I've been using TADST to launch dedicated servers.

While my addon options / ACE settings are not their default values, I know this bug isn't very old; months ago I took a long break from Arma, and in that time I believe there may have been an update for ACE.  Upon returning to the game, with the same settings as before, this bug started occurring.  It's gone on long enough and no one else seems to have reported it.

Upon request I can provide my addon settings, especially if no one else is experiencing this bug and it's somehow my combo of settings causing this.  This bug has caused many sessions to come to a halt; I've made this GitHub account for this reason.

**Videos:**
https://youtu.be/cbRnzB44naw

Salluci:
- Fix #8772.
- Add additional `onStateEntered` function for unconscious state. (logging + condition checking inside statemachine component)
- Add manual statemachine transition to dead state in `setDead`. (handling for forced respawn on stable unit)
- Add additional `alive` check in `setUnconsciousState`. (dead units shouldn't change consciousness state)
- Improve/add condition checking in other statemachine functions.

### IMPORTANT


addons/medical_statemachine/functions/fnc_enteredStateCardiacArrest.sqf
```diff
@@ -17,6 +17,7 @@
  */
 
 params ["_unit"];
+if (isNull _unit || {!isNil {_unit getVariable QEGVAR(medical,causeOfDeath)}}) exitWith {};
```
veteran29, line 20: Is this happening commonly? If not might be worth adding a WARNING in the code block.

Same for the rest of the occurrences of this check.

Salluci: In theory, shouldn't ever happen. Clockwork is supposed to filter out null and dead units. In practice, will happen every time conditions for #8772 trigger, which hasn't been exactly common IME, but could be relatively often for fatal injuries set to always.

Haven't dealt with WARNING yet. How would that look like?

veteran29: You use it like any other logging macro:
https://github.com/acemod/ACE3/blob/master/include/x/cba/addons/main/script_macros_common.hpp#L268
```sqf
if (isNull _unit || {!isNil {_unit getVariable QEGVAR(medical,causeOfDeath)}}) exitWith {
    WARNING_1("Unit is dead or null - %1",_unit);
};
```
if it's not an expected behaviour I thought we might log it.

veteran29 (commented):
LGTM but I won't approve as I would like somebody more familiar with the medical_statemachine to take a look at this.

Salluci:
Wake Baer from cryosleep?

addons/medical_statemachine/functions/fnc_enteredStateDeath.sqf
```diff
@@ -16,13 +16,14 @@
  */
 
 params ["_unit"];
-if (isNull _unit) exitWith {};
+if (isNull _unit || {!isNil {_unit getVariable QEGVAR(medical,causeOfDeath)}}) exitWith {
+    WARNING_1("State transition on dead or null unit - %1",_unit);
```
PabstMirror, line 20:
```suggestion
    if ((_unit getVariable [QEGVAR(medical,causeOfDeath), ""]) == "#scripted") exitWith {};
    WARNING_1("State transition on dead or null unit - %1",_unit);
```
We don't want to show warning for scripted deaths (`cursorObject setDammage 1`)
Because of the manual transition setDead will run before this and have already set the var

addons/medical_statemachine/functions/fnc_enteredStateDeath.sqf
```diff
@@ -16,13 +16,14 @@
  */
 
 params ["_unit"];
-if (isNull _unit) exitWith {};
+if (isNull _unit || {!isNil {_unit getVariable QEGVAR(medical,causeOfDeath)}}) exitWith {
+    WARNING_1("enteredStateDeath: State transition on dead or null unit - %1",_unit);
```
PabstMirror, line 20:
```suggestion
    if ((_unit getVariable [QEGVAR(medical,causeOfDeath), ""]) == "#scripted") exitWith {};
    WARNING_1("enteredStateDeath: State transition on dead or null unit - %1",_unit);
```
We don't want to show warning for scripted deaths (`cursorObject setDammage 1`)
Because of the manual transition setDead will run before this and have already set the var

PabstMirror (approved):

PR #8794, opened 2022-01-26
Common - Add cache to `fnc_isModLoaded`

Salluci:
- Title.
- Move remaining manual CfgPatches lookups to use the function.
- Change a CfgVehicles >> typeOf lookup to configOf.
- Change overpressure damage to only show blood effect if damage is enabled.

### IMPORTANT

### DIFF
--- a/addons/common/functions/fnc_isModLoaded.sqf
+++ b/addons/common/functions/fnc_isModLoaded.sqf
@@ -1,10 +1,10 @@
 #include "script_component.hpp"
 /*
  * Author: Glowbal
- * Check in cfgPatches if modification is loaded
+ * Check in CfgPatches if modification is loaded
  *
  * Arguments:
- * 0: Mod Name or Classname of the mod in cfgPatches <STRING>
+ * 0: Mod Name or Classname of the mod in CfgPatches <STRING>
  *
  * Return Value:
  * if modification is loaded <BOOL>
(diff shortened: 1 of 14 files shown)


jonpas:
Is it actually faster? Config reading is pretty fast to begin with, and this only does one config read. Benchmarks would be nice.

Salluci:
Old:
```
Result:
0.0178 ms

Cycles:
10000/10000

Code:
["ACE_Medical"] call ace_common_fnc_isModLoaded
```

PR:
```
Result:
0.0068 ms

Cycles:
10000/10000

Code:
["ACE_Medical"] call isModLoadedCached
```

addons/common/functions/fnc_isModLoaded.sqf
```diff
@@ -1,10 +1,10 @@
 #include "script_component.hpp"
 /*
  * Author: Glowbal
```
PabstMirror, line 3:
```suggestion
 * Author: Glowbal, Salluci
```

PabstMirror (approved):

addons/common/functions/fnc_isModLoaded.sqf
Salluci: Ooooo I get my name on a thing, shiny!

PR #8797, opened 2022-01-26
Cargo - Make player carry/drag item on unload

Salluci:
- Title.

Player will start carrying or dragging item after unloading from cargo if possible. Includes weight checking.

### IMPORTANT

### DIFF
--- a/addons/cargo/functions/fnc_unloadCarryItem.sqf
+++ b/addons/cargo/functions/fnc_unloadCarryItem.sqf
@@ -16,13 +16,24 @@
  * Public: No
  */
 params ["_unloader", "_object"];
+TRACE_2("unloadCarryItem-start",_unloader,_object);
 
 if !(["ace_dragging"] call EFUNC(common,isModLoaded)) exitWith {};
 
-if ([_unloader, _object] call EFUNC(dragging,canCarry)) then {
-    [_unloader, _object] call EFUNC(dragging,startCarry);
-} else {
-    if ([_unloader, _object] call EFUNC(dragging,canDrag)) then {
+// When unloading attached objects, this code will run before server has finished moving object to the safe position
+[{
+    params ["_unloader", "_object"];
+    (_unloader distance _object) < 10
+}, {
+    params ["_unloader", "_object"];
+    TRACE_2("unloadCarryItem-unloaded",_unloader,_object);
+    if ([_unloader, _object] call EFUNC(dragging,canCarry)) exitWith {
+        [_unloader, _object] call EFUNC(dragging,startCarry);
+    };
+    if ([_unloader, _object] call EFUNC(dragging,canDrag)) exitWith {
         [_unloader, _object] call EFUNC(dragging,startDrag);
     };
-};
+}, _this, 1.0, {  // delay is based on how long it will take server event to trigger and take effect
+    // not a hard error if this fails, could have just unloaded to other side of vehicle because of findSafePos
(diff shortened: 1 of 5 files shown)


addons/cargo/functions/fnc_unloadCarryItem.sqf
```diff
@@ -0,0 +1,29 @@
...
+if !(["ace_dragging"] call EFUNC(common,isModLoaded)) exitWith {};
+
+switch (true) do {
+    case ([_unloader, _object] call EFUNC(dragging,canCarry)): {
+        [_unloader, _object] call EFUNC(dragging,startCarry);
+    };
+    case ([_unloader, _object] call EFUNC(dragging,canDrag)): {
+        [_unloader, _object] call EFUNC(dragging,startDrag);
+    };
+};
```
jonpas, line 29:
```suggestion
if ([_unloader, _object] call EFUNC(dragging,canCarry)) then {
    [_unloader, _object] call EFUNC(dragging,startCarry);
} else {
    if ([_unloader, _object] call EFUNC(dragging,canDrag)) then {
        [_unloader, _object] call EFUNC(dragging,startDrag);
    };
};
```
I find this way more readable.

Salluci: Ew, nested if blocks.

As you wish.

PabstMirror (changes requested):
doesn't work in mp

addons/cargo/functions/fnc_unloadCarryItem.sqf
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player, object] call ace_cargo_fnc_unloadCarryItem
+ *
+ * Public: No
+ */
+params ["_unloader", "_object"];
```
PabstMirror, line 18:
```
params ["_unloader", "_object"];
TRACE_2("unloadCarryItem-start",_unloader,_object);

if !(["ace_dragging"] call EFUNC(common,isModLoaded)) exitWith {};

// When unloading attached objects, this code will run before server has finished moving object to the safe position
[{
    params ["_unloader", "_object"];
    (_unloader distance _object) < 10
}, {
    params ["_unloader", "_object"];
    TRACE_2("unloadCarryItem-unloaded",_unloader,_object);
    if ([_unloader, _object] call EFUNC(dragging,canCarry)) exitWith {
        [_unloader, _object] call EFUNC(dragging,startCarry);
    };
    if ([_unloader, _object] call EFUNC(dragging,canDrag)) exitWith {
        [_unloader, _object] call EFUNC(dragging,startDrag);
    };
}, _this, 1.1, { TRACE_1("unloadCarryItem-failed to unload nearby player",_this); }] call CBA_fnc_waitUntilAndExecute;
```

When unloading a object type in mp the distance check in dragging was failing

`ace_cargo_fnc_unloadItem` calls `QGVAR(serverUnload)` and then immeditly calls `unloadCarryItem`
but it takes a bit for the server to actually move the object down to where the player is

this adds a little delay so all that can happen (only local mp it's 4 frames)
It's possible for this to fail if the unload position is away from the player, but that's just something that can happen and not really an error

also I was worried about collision and physx but it seemed to behave ok
I also switched the nested if statements to the superior `exitWith` style

Salluci: Don't have access to anything that can (conveniently) commit atm, but works for me. Thanks for testing.

PabstMirror (approved):

PR #8803, opened 2022-01-27
Medical Status - Add `isDamageAllowed` check to `setDead`

Salluci:
- Title.

### IMPORTANT

### DIFF
--- a/addons/medical_status/XEH_postInit.sqf
+++ b/addons/medical_status/XEH_postInit.sqf
@@ -2,6 +2,3 @@
 
 // Handle pain changes on injury
 [QEGVAR(medical,injured), LINKFUNC(adjustPainLevel)] call CBA_fnc_addEventHandler;
-
-// Remote setDead, internal usage only
-[QGVAR(setDead), LINKFUNC(setDead)] call CBA_fnc_addEventHandler;
--- a/addons/medical_status/XEH_postInit.sqf
+++ b/addons/medical_status/XEH_postInit.sqf
@@ -2,3 +2,6 @@
 
 // Handle pain changes on injury
 [QEGVAR(medical,injured), LINKFUNC(adjustPainLevel)] call CBA_fnc_addEventHandler;
+
+// Remote setDead
+[QGVAR(setDead), LINKFUNC(setDead)] call CBA_fnc_addEventHandler;
--- a/addons/medical_status/XEH_postInit.sqf
+++ b/addons/medical_status/XEH_postInit.sqf
@@ -3,5 +3,5 @@
 // Handle pain changes on injury
 [QEGVAR(medical,injured), LINKFUNC(adjustPainLevel)] call CBA_fnc_addEventHandler;
 
-// Remote setDead
+// Remote setDead, internal usage only
 [QGVAR(setDead), LINKFUNC(setDead)] call CBA_fnc_addEventHandler;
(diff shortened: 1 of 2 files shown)


kymckay:
Just to clarify because this is really hard to review without context:

1. What error in the existing logic is causing units to remain alive but be in the dead state? Where is this function being called from a remote location?
2. Is this function **intended** to run on remote units (i.e. is this fix valid)? It's been a while since I worked on the state machine, but does it not run on all machines only for the units local to each?
3. Why is the frame delay necessary? What's happening without it?

Salluci:
1. Good question. This shouldn't ever be called from a remote location under normal working conditions.
2. Function header would imply this was originally local effect only. Need to ask commy on what intention was. Yes, statemachine *should* only run for local units.
3. We're shooting in the dark here. In theory, setting damage twice on a unit in the same frame works perfectly fine: the unit is killed by the first call of `setHitPointDamage` then the second call returns damage to previous so as remove visual changes. No idea if this is the part that's failling.

Veteran's community has had issues with players being moved to dead state while remaining alive. As entering dead state would require this function to be called, I can only see three things failling: 
1. The unit has damage disabled on it. I find it hard to believe this is the case.
2. The unit isn't local. `setHitPointDamage` on a remote unit has no effect. Thus, the unit remains alive but is in the dead state.
3. The second call of `setHitPointDamage` on the unit cancels the first if done in the same frame. This doesn't seem to be the case in my testing, but I've also seen no negative change in delaying a frame anyway.

kymckay:
Thanks, that's perfect! :relaxed:  As a general bit of feedback on development habits: try to make exactly those intentions/assumptions clear in both the code comments and in the git history. It saves reviewers having to reverse engineer the coding intention (or take your word that it's sound), but also is vital for anyone who looks at it a year from now and wants to know why it's there.

Personally I'm unsure about these changes, if we have no evidence to suggest they're the issue then they seem like a bit of a red herring. I do seem to recall we used to have a delay on the damage resetting, may be worth checking git history to see why it was removed (if true) :thinking: While there would be no harm in adding remote support to the function, if the intention is that it only ever runs local to the unit then the right fix would be to figure out where it is running remotely (if it is).

kymckay:
Ah looks like previously when we had a frame delay it was back when we were using `setStructuralDamage`, during the rewrite I removed it - but it's not analogous to the delay here because it was applied to the actual kill damage, not the damage afterwards.

veteran29:
@kymckay see the linked PR from my unit mod, the PR was made before I was actually able to evaluate where the issue is. @Salluci can you convert this PR to the draft for now?

veteran29:
After further investigation, I found out that the medical issue we were facing in my unit is by us enabling BIS EG Spectator the moment unit dies, this started causing issues due to `setHitPointDamage` respecting that flag.

@Salluci could you simplify this PR to include only the:
```sqf
if (!isDamageAllowed _unit) then {
    WARNING_1("setDead executed on unit with damage blocked - %1",_unit);
    _unit allowDamage true;
};
```
bit.

veteran29 (approved):

PabstMirror (approved):

kymckay (changes requested):
@veteran29 Are you hooking into the local event of this function which runs before the `setHitPointDamage`?

In any case I think this check should be moved to after that since other code using that event could be falsely preventing damage, would also be good to see a code comment explaining that very reasoning for future reference.

kymckay (approved):

PR #8811, opened 2022-02-04
Ballistics/Realistic Weights - Improve `LMG_Mk200_F`

Salluci:
- Change barrel length to 15" as per KAC's website.
- Change weight to 11.4lbs as per KAC's website.

Reference: https://www.knightarmco.com/12627/shop/military/machine-gun/lamg

### IMPORTANT


BaerMitUmlaut:
The in game barrel length seems to match other weapons with a similar length barrel (HK 416, CTAR), so that's definitely more accurate.

I did not find any good references that it was ever called the Stoner 99, I believe it should be the Stoner 96. I have not found out where the 99 designation comes from, I found mentions of it in some blog comments and some airsoft related forum threads, but nothing solid. The origin within ACE seems to be https://github.com/KoffeinFlummi/AGM/pull/826 from @bux.

I am not a fan of just changing all translations, especially not the Japanese one. The Chinese one is untouched for some reason? I would suggest just changing the number, since that will still be more accurate but not as confusing to players and cause less translation issues.

bux:
I actually can't remember why we went for the 99 designation. It might have been something to do with the 2035 setting. But iirc there was a reason why we didn't use 96.

Salluci:
Alright. I'll drop the naming changes for now, then.

PR #8814, opened 2022-02-08
Advanced Ballistics - Add pre-computed extension data to improve performance on some maps

jetelain:
- Allow third party maps to provide per-computed advanced ballistics extension data
- Add a function to compute advanced ballistics extension data and copy it to clipboard
- Add an optional X-Cam Taunus pbo to avoid up to 10 minutes framerate drop in exchange of 0.5 to 1 seconds extra loading time.

### IMPORTANT


Salluci:
Pre-compile for vanilla maps as well?

optionals/compat_xcam_taunus/config.cpp
```diff
@@ -0,0 +1,17 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        name = COMPONENT_NAME;
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"xcam_taunus"};
```
jonpas, line 9: Is there an actual requirement? Seems like the config entry does not create one.

Maybe we should just have that in ACE3 directly if a compat is actually not needed.

PabstMirror: We should be able to just include the data, if the map exists or not
i.e. We're not adding it to cfgWorlds, but it's own custom config

veteran29: I'm not sure how much impact such big config entries could have on the performance/memory usage. Perhaps the data could be stored in the script file instead of config? With SQFC it should load fast enough.

jetelain: I will make a benchmark to see SQF vs SQFC vs config to see

dedmen:
Uff.
Huuuge config array is quite bad, less efficient than SQF and much less than SQFC indeed. And think about AIO config exports.. oof.. Please don't do that.
And then you iterate over that array in script and send it to extension.

Its a extension, its advanced as heck compared to script. Let it do the parsing.
Make the data be a large (compressed if possible) string, that is just passed as one blob to the extension.
Less filesize and much more efficient.
Especially once terrains start to add more precomputed things.

jetelain:
Hi,

Don't worry, it's clearly not suitable for large scale usage.

The ideal way would to be act like ACRE2 and reads a binary data file directly from PBO. It would avoid to do the heavy lift with script. But it has some drawbacks.

I think so, passing a large BLOB (likely as Base64) could the best compromise. It was initially designed like a separate mod, so I did not thought modifying the extension.

I'm currently working on other projects, and my current mod provides a suitable workaround for performance issues with the X-CAM Taunus map, despite not ideal, it's still largely more efficient that original behavior.

PR #8824, opened 2022-02-18
medical_damage - cleanup woundHandler naming

PabstMirror:
delete old woundHandler sqf for extension that's been unused for years


kymckay (approved):
Looks good, I'd perhaps rename the default handler to something like `woundsHandlerBase` rather than `Active`, but happy either way.

jonpas:
I vote `woundsHandlerBase` too. Or just `woundsHandler`, it's not like it's a "base class", it does its own thing and the others just add on it in parallel.

Why not delete the extension code as well? It's outdated anyways.

kymckay:
@jonpas They actually act in series:

>All wound handlers defined for a particular damage type will be called, and they are called in the opposite order they were added. The value returned by each handler is what will be passed as parameters to the next one - if the return is not valid (not an array, or not enough arguments) handling will be terminated and no other handler functions will be called. This allows custom wound handlers to selectively handle damage and either block the built-in handling or pass modified damage data.

jonpas:
Well yes, because you can't run them "in parallel" without effects in the game. I meant that you are not extending that base wound handler to create your own, you are **adding** to it.

mharis001 (approved):

PR #8827, opened 2022-02-21
Cargo - Add alternative unloading item from vehicle cargo

smitt14ua:
- New button in cargo menu to select place for unloading

**YouTube presentation:**
<details>
<summary>Open/close</summary>
.

[Arma 3 acemod item placement](https://youtu.be/LfhT4sAbN2o "Arma 3 acemod item placement")

[Arma 3 acemod item placement 2](https://youtu.be/HOo_5mGn6PI "Arma 3 acemod item placement 2")
</details>


severgun:
🤔 https://github.com/acemod/ACE3/pull/8797

smitt14ua:
> Duplicate?🤔 #8797

If player cannot carry unloaded object, it will be placed randomly. My solution is for big (not-carried) objects and prevents collision bugs after unloading.

addons/cargo/functions/fnc_deploy.sqf
```diff
@@ -0,0 +1,143 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call ace_cargo_fnc_deploy
+ *
+ * Public: No
+ */
+
+disableSerialization;
```
commy2, line 18: No.

smitt14ua: Do you think that better to have special function for getting selected item?

smitt14ua: I moved getting a selected item into separate function.

johnb432:
What needs to be finished or looked at for this to be pulled?

smitt14ua:
> What needs to be finished or looked at for this to be pulled?

Nothing, I think. commy2 didn't respond me

LinkIsGrim:
> What needs to be finished or looked at for this to be pulled?

General testing, MP, and code review.

johnb432:
I've taken the liberty to implement some fixes and improvements directly:
- Combined `handle` functions into a single function `FUNC(handleDeployInterrupt)`.
- Implemented `FUNC(handleDeployInterrupt)` being called (on opening the interact menu, vehicle and weapon switching, being killed and captured, surrendering, going unconscious and feature cameras changing.)
- Made sure that `setDir` is called before any position setting commands. Wiki suggests this is necessary and I did not verify this.
- Added authors from functions that were used as basis for the newly written ones. I *assume* they came from `sandbag`, as the comments are exactly the same. If I'm wrong, we can easily correct that.
- Removed 1 second delay after placing. This, in my opinion, does not make any sense and I will update `sandbag`'s code accordingly.

I have tested this both in SP and self-hosted MP, works in both cases.

My suggestion is to merge #9617 first (if it's accepted), then this PR could be updated to accommodate the changes.

smitt14ua:
Thx, @johnb432 ❤️

addons/cargo/functions/fnc_unloadItem.sqf
```diff
@@ -41,9 +46,18 @@ if (_itemSize < 0) exitWith {
...
-TRACE_1("findUnloadPosition",_emptyPosAGL);
+private _deployed = _deploy isNotEqualTo [];
+
+if (!_deployed) then {
+    // This covers testing vehicle stability and finding a safe position
+    for "_i" from 1 to 3 do {
+        _emptyPosAGL = [_vehicle, _item, _unloader] call EFUNC(common,findUnloadPosition);
+
+        if (_emptyPosAGL isNotEqualTo []) exitWith {};
+    };
```
LinkIsGrim, line 57: I'm not sure why this should be executed more than once?

LinkIsGrim (commented):
LGTM other than comment above

addons/cargo/functions/fnc_unloadItem.sqf
johnb432: Because `EFUNC(common,findUnloadPosition)` is not guaranteed to return a valid position, despite there one existing.

I'm not a fan of it, but I'm not sure how else to go about it - unless we modify `EFUNC(common,findUnloadPosition)` to always find an existing spot if one exists.

LinkIsGrim: > Because EFUNC(common,findUnloadPosition) is not guaranteed to return a valid position, despite there one existing.

Thanks, I hate it.

LinkIsGrim:
I've removed hiding the object when deployment is blocked since it's less confusing that way IMO.

Tweaked object radius to use firegeometry bounding box as that's closer to the real thing.

You can place objects inside other objects up to a certain point. If we want to prevent that, the intersection checks need to be changed to `lineIntersectsSurfaces` and begPos needs to be moved further along the line between eye/vehiclePos and the model's position. Guesstimation says half of the object radius should be good.

johnb432:
> If we want to prevent that, the intersection checks need to be changed to `lineIntersectsSurfaces` and begPos needs to be moved further along the line between eye/vehiclePos and the model's position. Guesstimation says half of the object radius should be good.

You'll need to run that by me again chief, I don't get it.

LinkIsGrim:
Math

johnb432:
I went ahead and simplified the deployment method, by essentially copying what carrying does. I feel it's more robust and allows for greater flexibility.
We can improve the deployment method later, but imo it's better if we use something simple for now.

LinkIsGrim (approved):
LGTM


https://github.com/acemod/ACE3/assets/69561145/bafba925-666c-4eee-8aa0-2e838f609331

addons/cargo/stringtable.xml
```diff
@@ -528,5 +562,12 @@
             <French>Active si les éléments de cargaison sont portés ou traînés après le déchargement.</French>
             <Portuguese>Controla se os itens de carga são carregados ou arrastados após a descarga.</Portuguese>
         </Key>
+
+        <Key ID="STR_ACE_Cargo_enableDeploy">
```
LinkIsGrim, line 566:
```suggestion
        <Key ID="STR_ACE_Cargo_enableDeploy">
```

PR #8752, opened 2021-12-22
WIP - Medical - Add damage handling for crew inside destroyed vehicles

LinkIsGrim:
**When merged this pull request will:**
- Title.

`_lethality` is just arbitrary.
### IMPORTANT

- [ ] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3mod.com/).
- [x] [Development Guidelines](https://ace3mod.com/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #8832, opened 2022-03-02
Medical Engine - Add vehicle explosion handling

pterolatypus:
- Add damage handling for units inside vehicles that explode

Implementation moved from #8278, also see #8752 

It's currently a bit [extreme](https://i.imgur.com/dAvhEOR.jpeg) but comparable to what a unit standing next to the vehicle experiences, and it's probably going to be an instakill anyway so not super important.

### IMPORTANT


addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -67,6 +67,34 @@ if (
     0
 };
 
+// Being inside an exploding vehicle doesn't trigger for each hitpoint
+// It seems to fire twice with ammo type "FuelExplosion" or "FuelExplosionBig"
+if (
+    _hitPoint isEqualTo "#structural" &&
+    {_ammo isKindOf "FuelExplosion"} &&
+    {_vehicle != _unit} &&
+    {_damage == 1}
```
severgun, line 76: Is it always 1?

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -67,6 +67,34 @@ if (
...
+    if (_uniform isEqualTo "") then {
+        _uniform = getText (configOf _unit >> "nakedUniform");
+    };
+    private _uniformClass = GET_STRING(configFile >> "CfgWeapons" >> _uniform >> "ItemInfo" >> "uniformClass", "U_BasicBody");
+    private _damages = [];
+    {
+        private _armor = [_unit, _x] call FUNC(getHitpointArmor);
+        // would be nice to move this into getHitpointArmor
+        private _shielding = GET_NUMBER(configFile >> "CfgVehicles" >> _uniformClass >> "Hitpoints" >> _x >> "explosionShielding", 1);
+        _damages pushBack [_hit*_shielding/_armor, ALL_BODY_PARTS select _forEachIndex, _hit*_shielding]
```
severgun, line 90: missing semicolon

Salluci:
Can't `configOf` be used on `_ammo`? Might save some microseconds.

addons/medical_engine/functions/fnc_handleDamage.sqf
pterolatypus: Yes, at least in all the testing I've done.
Also, not having this condition could lead to double-handling when a unit is inside a vehicle and another vehicle explodes nearby.

kymckay:
I notice #8752 includes a check to filter out destruction of vehicles which don't explode (iiirc rubber dinghy would be one), is that not needed?

Otherwise I like the logic here. Another thing that other PR includes is setting the crew on fire, but I'm not sure if that should occur from here or be picked up elsewhere using an event or similar.

pterolatypus:
I would ex

pterolatypus:
Can confirm, vehicles that don't explode don't trigger HandleDamage.
As for ejecting the crew, I might make a separate PR for that as I have some ideas.

IMO this is good to go.

tbeswick96:
What's the intended hit case for this? I'm testing it out as I've got an issue where vehicles fully blown up by a missile/bomb just have the units inside get out as if nothing happened and walk away.
I've messed around with this and found replacing the `{_ammo isKindOf "FuelExplosion"}` check with `{GET_NUMBER(configFile >> "CfgAmmo" >> _ammo >> "explosive",0) == 1}` and tweaking `{_damage == 1}` to `{_damage >= 1}` stops the above case, but I'm not sure if that was the point of this handling block.

LinkIsGrim:
Done in #9246

PR #8840, opened 2022-03-09
Vehicles - Mousewheel will set speed limiter value to even multiples

PabstMirror:
Before if you set speedlimter at 22
22->27->32

now
22->25->30


jonpas (approved):
Yeees!

jokoho48:
why only do that with 5? should this maybe be done always?

PabstMirror:
it's weird if the setting is something like `6.2`, you get floating point rounding errors

mharis001 (approved):

mharis001:
Maybe change the setting to always be round numbers (`__trailingDecimals` to `-1`). It is already set to `0` (don't show decimals).

mharis001 (approved):

PR #8852, opened 2022-03-20
Tagging - Allow tagging vehicles (overwrite "clan" tag)

PabstMirror:
https://www.youtube.com/watch?v=jaslhKPtHCw


addons/tagging/functions/fnc_createTag.sqf
```diff
@@ -20,14 +21,20 @@
...
 TRACE_5("createTag:",_tagPosASL,_vectorDirAndUp,_texture,_object,_unit);
 
 if (_texture == "") exitWith {
     ERROR_1("%1 is not a valid tag texture.",_texture);
     false
 };
 
+if (_isVehicleTag) exitWith {
+    TRACE_3("tagging vehicle",_object,typeOf _object,_texture);
+    _object setObjectTextureGlobal ["clan", _texture];
```
neilzar, line 34:
```suggestion
    _object setObjectTextureGlobal [getText (configOf _obj >> "selectionClan"), _texture];
```
In case any mod decides to not name the clan selection "clan".

PabstMirror:
uses same system as https://github.com/acemod/ACE3/pull/9018 but I don't think they conflict

BrettMayson (approved):
It does work fine on vehicles that support it. Does feel a little weird to me that it also works on vehicles without a selection for a clan tag, and that it uses self-interaction instead of the vehicle's interaction point.

PabstMirror:
Added filter for vehicles without a valid `selection`

BrettMayson (approved):

veteran29 (approved):
LGTM

PR #8857, opened 2022-03-22
Casings - Add infantry bullet casings

Cyruz143:
- It's a stripped down and basic version of (https://github.com/acemod/ACE3/pull/4132).
- Only impacts infantry weapons, no vehicles etc
- Fully local, no network sync
- Has some settings (on/off, maximum amount of brass)

I've been using a version of this since 2020. It works fairly well, will have odd floating cases, especially near the edges of objects and buildings etc. I've taken most of the vector bits from the original PR, math could probably be better if anyone wants to have a go!

I've left esteldunedain in the authors as I'm using parts of his original code. Also @diwako helped me with this a while ago, happy to add you on to this too bud!

### IMPORTANT

### DIFF
--- a/addons/casings/XEH_postInit.sqf
+++ b/addons/casings/XEH_postInit.sqf
@@ -1,0 +1,7 @@
+#include "script_component.hpp"
+
+if (!hasInterface || !(GVAR(casingsEnabled))) exitWith {};
+
+GVAR(casings) = createHashMap;
+GVAR(casingsArr) = [];
+["CAManBase", "FiredMan", {call FUNC(createCase)}] call CBA_fnc_addClassEventHandler;
--- a/addons/casings/XEH_postInit.sqf
+++ b/addons/casings/XEH_postInit.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 
-if (!hasInterface || !(GVAR(casingsEnabled))) exitWith {};
+if (!hasInterface || !(GVAR(enabled))) exitWith {};
 
 GVAR(casings) = createHashMap;
 GVAR(casingsArr) = [];
-["CAManBase", "FiredMan", {call FUNC(createCase)}] call CBA_fnc_addClassEventHandler;
+["CAManBase", "FiredMan", {call FUNC(createCasing)}] call CBA_fnc_addClassEventHandler;
--- a/addons/casings/XEH_postInit.sqf
+++ b/addons/casings/XEH_postInit.sqf
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 
-if (!hasInterface || !(GVAR(enabled))) exitWith {};
+if (!hasInterface || !GVAR(enabled)) exitWith {};
 
 GVAR(cachedCasings) = createHashMap;
 GVAR(casings) = [];
--- a/addons/casings/XEH_postInit.sqf
+++ b/addons/casings/XEH_postInit.sqf
@@ -2,6 +2,6 @@
 
 if (!hasInterface || !(GVAR(enabled))) exitWith {};
 
-GVAR(casings) = createHashMap;
-GVAR(casingsArr) = [];
+GVAR(cachedCasings) = createHashMap;
+GVAR(casings) = [];
 ["CAManBase", "FiredMan", {call FUNC(createCasing)}] call CBA_fnc_addClassEventHandler;
--- a/addons/casings/functions/fnc_createCasing.sqf
+++ b/addons/casings/functions/fnc_createCasing.sqf
@@ -17,7 +17,7 @@
  * Public: No
  */
 
-params ["_unit", "_weapon", "", "", "_ammo"];
+params ["_unit", "", "", "", "_ammo"];
 
 if (!isNull objectParent _unit) exitWith {};
 
--- a/addons/casings/functions/fnc_createCasing.sqf
+++ b/addons/casings/functions/fnc_createCasing.sqf
@@ -42,11 +42,14 @@ if (_modelPath isEqualTo "") exitWith {};
 
 private _unitPos = getposASL _unit;
 // Distant shooters don't produce as many cases
-if ((positionCameraToWorld [0,0,0]) vectorDistance _unitPos > 100 && {random 1 < 0.9}) exitWith {};
+if ((AGLToASL positionCameraToWorld [0,0,0]) vectorDistance _unitPos > 100 && {random 1 < 0.9}) exitWith {};
 
 private _weapDir = _unit weaponDirection currentWeapon _unit;
 private _ejectDir = _weapDir vectorCrossProduct [0, 0, 1];
-private _pos = _unitPos vectorAdd (_weapDir vectorMultiply (-0.5 + random 1.0 + random 1.0)) vectorAdd (_ejectDir vectorMultiply (0.2 + random 1.0 + random 1.0)) vectorAdd [0, 0, 0.005];
+private _pos = _unitPos
+    vectorAdd (_weapDir vectorMultiply (-0.5 + random 1 + random 1))
+    vectorAdd (_ejectDir vectorMultiply (0.2 + random 1 + random 1))
+    vectorAdd [0, 0, 0.005];
 
 [
     {
--- a/addons/casings/functions/fnc_createCasing.sqf
+++ b/addons/casings/functions/fnc_createCasing.sqf
@@ -60,10 +60,8 @@ private _pos = _unitPos
         _casing setDir (random 360);
         private _idx = GVAR(casings) pushBack _casing;
 
-        if (_idx >= GVAR(maxCasings)) then {
-            for "_" from 0 to (_idx - GVAR(maxCasings)) do {
-                deleteVehicle (GVAR(casings) deleteAt 0);
-            };
+        for "_" from 0 to (_idx - GVAR(maxCasings)) do {
+            deleteVehicle (GVAR(casings) deleteAt 0);
         };
     },
     [_modelPath,_pos],
--- a/addons/casings/functions/fnc_createCasing.sqf
+++ b/addons/casings/functions/fnc_createCasing.sqf
@@ -46,10 +46,7 @@ if ((positionCameraToWorld [0,0,0]) vectorDistance _unitPos > 100 && {random 1 <
 
 private _weapDir = _unit weaponDirection currentWeapon _unit;
 private _ejectDir = _weapDir vectorCrossProduct [0, 0, 1];
-private _pos = _unitPos vectorAdd
-                  (_weapDir vectorMultiply (-0.5 + random 1.0 + random 1.0)) vectorAdd
-                  (_ejectDir vectorMultiply (0.2 + random 1.0 + random 1.0));
-
+private _pos = _unitPos vectorAdd (_weapDir vectorMultiply (-0.5 + random 1.0 + random 1.0)) vectorAdd (_ejectDir vectorMultiply (0.2 + random 1.0 + random 1.0));
 _pos set [2, (_pos #2) + 0.005];
 
 [
--- a/addons/casings/functions/fnc_createCasing.sqf
+++ b/addons/casings/functions/fnc_createCasing.sqf
@@ -47,7 +47,7 @@ if ((AGLToASL positionCameraToWorld [0,0,0]) vectorDistance _unitPos > 100 && {r
 private _weapDir = _unit weaponDirection currentWeapon _unit;
 private _ejectDir = _weapDir vectorCrossProduct [0, 0, 1];
 private _pos = _unitPos
-    vectorAdd (_weapDir vectorMultiply (-0.5 + random 1 + random 1))
+    vectorAdd (_weapDir vectorMultiply (-0.5 + random 2))
     vectorAdd (_ejectDir vectorMultiply (0.2 + random 1 + random 1))
     vectorAdd [0, 0, 0.005];
 
--- a/addons/casings/functions/fnc_createCasing.sqf
+++ b/addons/casings/functions/fnc_createCasing.sqf
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 /*
- * Author: esteldunedain / Cyruz
+ * Author: esteldunedain / Cyruz / Diwako
  * Produces a casing matching the fired rifles caliber on the ground around the unit
  *
  * Arguments:
--- a/addons/casings/functions/fnc_createCasing.sqf
+++ b/addons/casings/functions/fnc_createCasing.sqf
@@ -56,7 +56,8 @@ private _pos = _unitPos
         private _casing = createSimpleObject [_modelPath, [0,0,0], true];
         _casing setDir (random 360);
         private _lisPos = (lineIntersectsSurfaces [_pos, _pos vectorAdd [0,0,-1e11], objNull, objNull, true, 1, "ROADWAY", "FIRE"]) #0;
-        _casing setPosASL _lisPos #0;
+        private _casing = createSimpleObject [_modelPath, _lisPos #0, true];
+        _casing setDir (random 360);
         _casing setVectorUp _lisPos #1;
         private _idx = GVAR(casings) pushBack _casing;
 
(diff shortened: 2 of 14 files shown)


addons/casings/initSettings.sqf
```diff
@@ -0,0 +1,19 @@
+[
+    QGVAR(enabled), "CHECKBOX",
+    [LSTRING(displayName), LSTRING(description)],
+    LSTRING(Settings_DisplayName),
+    true,
+    true,
+    {},
+    true
```
diwako, line 8: As this spawns objects in the end, it can be a performance hog for older system, I would suggest that this should be a client setting.

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,71 @@
...
+_pos set [2, (_pos #2) + 0.005];
+
+[
+    {
+        params ["_modelPath","_pos"];
+
+        private _casing = createSimpleObject [_modelPath, [0,0,0], true];
+        _casing setPosATL _pos;
+        _casing setDir (random 360);
+        GVAR(casingsArr) pushBack _casing;
```
diwako, line 59: pushback gives you an index, use that instead of using count afterwards

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+    private _cartridge = getText (configFile >> "CfgAmmo" >> _ammo >> "cartridge");
+    //Default cartridge is a 5.56mm model
+    _modelPath = switch (_cartridge) do {
+        case "FxCartridge_9mm":     { "A3\Weapons_f\ammo\cartridge_small.p3d" };
+        case "FxCartridge_65":      { "A3\weapons_f\ammo\cartridge_65.p3d" };
+        case "FxCartridge_762":     { "A3\weapons_f\ammo\cartridge_762.p3d" };
+        case "FxCartridge_127":     { "A3\weapons_f\ammo\cartridge_127.p3d" };
+        case "FxCartridge_slug":    { "A3\weapons_f\ammo\cartridge_slug.p3d" };
+        case "":                    { "" };
+        default { "A3\Weapons_f\ammo\cartridge.p3d" };
```
commy2, line 36: Could probably be optimized using HashMap.

Also, getText and building the config is kind of slow. Since this is basically every frame for fast firing weapons: caching?

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+    };
+    GVAR(casings) set [_ammo, _modelPath];
+};
+
+if (_modelPath isEqualTo "") exitWith {};
+
+private _unitPos = getposATL _unit;
+// Distant shooters don't produce as many cases
+if ((positionCameraToWorld [0,0,0]) vectorDistance _unitPos > 100 && {random 1 < 0.9}) exitWith {};
```
commy2, line 45: vector math must be done in ASL, because ATL does not have a flat topography which is necessary for vector commands to work accurately.

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+
+private _weapDir = _unit weaponDirection currentWeapon _unit;
+private _ejectDir = _weapDir vectorCrossProduct [0, 0, 1];
+private _pos = _unitPos vectorAdd (_weapDir vectorMultiply (-0.5 + random 1.0 + random 1.0)) vectorAdd (_ejectDir vectorMultiply (0.2 + random 1.0 + random 1.0));
+_pos set [2, (_pos #2) + 0.005];
+
+[
+    {
+        params ["_modelPath","_pos"];
```
commy2, line 54: space after comma

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+if (_modelPath isEqualTo "") exitWith {};
+
+private _unitPos = getposATL _unit;
+// Distant shooters don't produce as many cases
+if ((positionCameraToWorld [0,0,0]) vectorDistance _unitPos > 100 && {random 1 < 0.9}) exitWith {};
+
+private _weapDir = _unit weaponDirection currentWeapon _unit;
+private _ejectDir = _weapDir vectorCrossProduct [0, 0, 1];
+private _pos = _unitPos vectorAdd (_weapDir vectorMultiply (-0.5 + random 1.0 + random 1.0)) vectorAdd (_ejectDir vectorMultiply (0.2 + random 1.0 + random 1.0));
+_pos set [2, (_pos #2) + 0.005];
```
commy2, line 50: `vectorAdd [0, 0, 0.005]` in the expression above.

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+    {
+        params ["_modelPath","_pos"];
+
+        private _casing = createSimpleObject [_modelPath, [0,0,0], true];
+        _casing setPosATL _pos;
+        _casing setDir (random 360);
+        private _idx = GVAR(casingsArr) pushBack _casing;
+
+        if (_idx >= GVAR(maxCasings)) then {
+            for "_i" from 0 to (_idx - GVAR(maxCasings)) do {
```
commy2, line 62: throwaway index should probably be "_" or "-" instead of "_i".

addons/casings/functions/fnc_createCasing.sqf
commy2: Ah, nvm, it already is cached. Probably want to rename `GVAR(casings)` to something that says it's an cache (and replace GVAR(casingsArr) with just GVAR(casings)).

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Example:
+ * [player, currentWeapon player, "","", player ammo currentWeapon player] call ace_casings_fnc_createCase
+ *
+ * Public: No
+ */
+
+params ["_unit", "_weapon", "", "", "_ammo"];
+
+if (!isNull objectParent _unit || { _weapon in ["Throw","Put"] } ) exitWith {};
```
commy2, line 22: Excluding Throw and Put here seems like an overall loss. The casing would be "" anyway for these. I don't see why they are special compared to, e.g. a launcher or an ugl.

addons/casings/functions/fnc_createCasing.sqf
Cyruz143: Switched to cachedCasings for hashmap and casings for the array.

addons/casings/functions/fnc_createCasing.sqf
Cyruz143: TIL!

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,68 @@
...
+        default { "A3\Weapons_f\ammo\cartridge.p3d" };
+    };
+    GVAR(cachedCasings) set [_ammo, _modelPath];
+};
+
+if (_modelPath isEqualTo "") exitWith {};
+
+private _unitPos = getposASL _unit;
+// Distant shooters don't produce as many cases
+if ((positionCameraToWorld [0,0,0]) vectorDistance _unitPos > 100 && {random 1 < 0.9}) exitWith {};
```
commy2, line 45:
```SQF
AGLToASL positionCameraToWorld
```

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,68 @@
...
+
+private _weapDir = _unit weaponDirection currentWeapon _unit;
+private _ejectDir = _weapDir vectorCrossProduct [0, 0, 1];
+private _pos = _unitPos vectorAdd (_weapDir vectorMultiply (-0.5 + random 1.0 + random 1.0)) vectorAdd (_ejectDir vectorMultiply (0.2 + random 1.0 + random 1.0)) vectorAdd [0, 0, 0.005];
```
commy2, line 49: Would break up the line:
```sqf
 private _pos = _unitPos
    vectorAdd (_weapDir vectorMultiply (-0.5 + random 1.0 + random 1.0))
    vectorAdd (_ejectDir vectorMultiply (0.2 + random 1.0 + random 1.0)) 
    vectorAdd [0, 0, 0.005];
```

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,68 @@
...
+[
+    {
+        params ["_modelPath", "_pos"];
+
+        private _casing = createSimpleObject [_modelPath, [0,0,0], true];
+        _casing setPosASL _pos;
+        _casing setDir (random 360);
+        private _idx = GVAR(casings) pushBack _casing;
+
+        if (_idx >= GVAR(maxCasings)) then {
```
commy2, line 60: Upon thinking about this, the if is unnecessary.

diwako: how so? at some point you need to start deleting cases.

addons/casings/stringtable.xml
```diff
@@ -0,0 +1,20 @@
...
+<Project name="ACE">
+    <Package name="Casings">
+        <Key ID="STR_ACE_Casings_Settings_DisplayName">
+            <English>ACE Casings</English>
+        </Key>
+        <Key ID="STR_ACE_Casings_displayName">
+            <English>Casings Enabled</English>
+        </Key>
+        <Key ID="STR_ACE_Casings_description">
+            <English>Enable persistent casings</English>
```
diwako, line 11: I feel like this needs to be elaborated a bit more. The layman will not know what this exactly means. E.g. say that it will add new objects on the ground and it might be a performance hazard for older systems.

Cyruz143: Added a warning, will see if anyone has better wording. All the current wording was listed from the existing PR so might need a general review.

addons/casings/functions/fnc_createCasing.sqf
commy2: Keep the for loop. It already implies the if making it superfluous. Remember that `for "-" from -1 to 0` is a noop.

diwako: i see. in most cases it will be then  `from 0 to 0` which is one step.

Cyruz143: diwako explained it, I still don't get it but it works so it's fine...

addons/casings/XEH_postInit.sqf
```diff
@@ -0,0 +1,7 @@
+#include "script_component.hpp"
+
+if (!hasInterface || !(GVAR(enabled))) exitWith {};
```
rautamiekka, line 3:
```suggestion
if (!hasInterface || {!GVAR(enabled)}) exitWith {};
```
Eliminates the need to unnecessarily check both, and pretty sure the parentheses ain't needed.

commy2: This is slower on average for any mission with more than 1 player connected, regardless whether casings are enabled or not.

commy2: I do support axing the parentheses though.

rautamiekka: > This is slower on average for any mission with more than 1 player connected, regardless whether casings are enabled or not.

Huh. Complete contrast to what Biwiki said.

commy2: >BI Wiki

LOL
Where do you think you are?

Cyruz143: it's 0.0007ms vs 0.0008ms here guys for something that runs once. Binned parentheses .

rautamiekka: > > BI Wiki
> 
> LOL Where do you think you are?

Possibly remembered the spelling wrong but the point gets across.

jonpas (approved):
Short and neat.

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player, currentWeapon player, "","", player ammo currentWeapon player] call ace_casings_fnc_createCase
+ *
+ * Public: No
+ */
+
+params ["_unit", "_weapon", "", "", "_ammo"];
```
PabstMirror, line 20:
```suggestion
params ["_unit", "", "", "", "_ammo"];
```

PabstMirror:
some physx//gravity problems, but probably not easily fixable (might be because they are simple objects)

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+if (_modelPath isEqualTo "") exitWith {};
+
+private _unitPos = getposASL _unit;
+// Distant shooters don't produce as many cases
+if ((AGLToASL positionCameraToWorld [0,0,0]) vectorDistance _unitPos > 100 && {random 1 < 0.9}) exitWith {};
+
+private _weapDir = _unit weaponDirection currentWeapon _unit;
+private _ejectDir = _weapDir vectorCrossProduct [0, 0, 1];
+private _pos = _unitPos
+    vectorAdd (_weapDir vectorMultiply (-0.5 + random 1 + random 1))
```
PabstMirror, line 50:
```suggestion
    vectorAdd (_weapDir vectorMultiply (-0.5 + random 2))
```

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+    vectorAdd (_ejectDir vectorMultiply (0.2 + random 1 + random 1))
```
PabstMirror, line 51:
```suggestion
    vectorAdd (_ejectDir vectorMultiply (0.2 + random 2))
```

Cyruz143:
After many discussions in dev on Slack. I'm switched it to use an LIS check. Because of the performance concerns, especially for something visual, I've disabled it by default.

Cyruz143:
Ran this PR'd version through a session and got a few people of varying hardware specs to enable it. Even during some fairly big fire fights, no-one noticed any performance impact (as I was asking them to keep an eye on stuttering/fps etc). Obviously the arma butt dyno isn't the best performance metric but it's better than nothing!

jonpas:
I think this should be enabled by default. :)

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+    vectorAdd (_ejectDir vectorMultiply (0.2 + random 2));
+
+[
+    {
+        params ["_modelPath", "_pos"];
+
+        private _casing = createSimpleObject [_modelPath, [0,0,0], true];
+        _casing setDir (random 360);
+        private _lisPos = (lineIntersectsSurfaces [_pos, _pos vectorAdd [0,0,-1e11], objNull, objNull, true, 1, "ROADWAY", "FIRE"]) #0;
+        _casing setPosASL _lisPos #0;
```
PabstMirror, line 59:
```suggestion
        private _casing = createSimpleObject [_modelPath, _lisPos #0, true];
        _casing setDir (random 360);
```

addons/casings/functions/fnc_createCasing.sqf
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Arguments:
+ * 0: unit - Object the event handler is assigned to <OBJECT>
+ * 1: ammo - Ammo used <STRING>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player, "", "","", "B_556x45_Ball"] call ace_casings_fnc_createCase
```
PabstMirror, line 14:
```suggestion
 * [player, "", "","", "B_556x45_Ball"] call ace_casings_fnc_createCasing
```

PabstMirror (approved):

PabstMirror:
Bullets seem to be halfway underground
I think because it's a simple object it's placed via https://community.bistudio.com/wiki/setPosWorld (model center)
So I guess we might want to add the radius of the case to the z height (probably 4-5mm)
But it's not a big deal

PabstMirror (approved):

PR #8858, opened 2022-03-27
MapTools - Allow mod maps and compasses to use map tools

### LINKED ISSUE #8851: Aligning mapTools with current compass heading does only work for vanilla compass
The mapTools feature of aligning the tool with the current compass heading does only work if the player is equipped with a vanilla compass (`'ItemCompass'`). 

- In case you are playing with the GM DLC this will not work because it uses a different compass (might cause problems with other DLCs as well).
- A vehicle gunner with powerful optics but without a compass equipped does also not get the option to align the map tool with himself (which doesn't make sense IMHO).

So the AND condition in below mentiond LOC does more harm than good. **I recommend to remove it.** The only damage being done in doing so is that infanterists and low tech vehicle gunners with map tools but without a compass (who does that anyways) will be able to align the map tools with their own heading.

https://github.com/acemod/ACE3/blob/d1353713198720fda42a134d6d1a6e23e4c4d4c2/addons/maptools/CfgVehicles.hpp#L55

Let me know what you think. I can remove it since I am currently already working on mapTools, but I want to hear your feedback first.

Drofseh:
- close #8851

### DIFF
--- a/addons/maptools/CfgVehicles.hpp
+++ b/addons/maptools/CfgVehicles.hpp
@@ -52,7 +52,7 @@ class CfgVehicles {
                 };
                 class ACE_MapToolsAlignCompass {
                     displayName = CSTRING(MapToolsAlignCompass);
-                    condition = QUOTE((GVAR(mapTool_Shown) != 0) && {'ItemCompass' in assigneditems ACE_player});
+                    condition = QUOTE((GVAR(mapTool_Shown) != 0) && {getUnitLoadout ACE_player param [9, []] param [3, ""] != ""});
                     statement = QUOTE(GVAR(mapTool_angle) = getDir ACE_player;);
                     exceptions[] = {"isNotDragging", "notOnMap", "isNotInside", "isNotSitting"};
                     showDisabled = 1;
--- a/addons/maptools/CfgVehicles.hpp
+++ b/addons/maptools/CfgVehicles.hpp
@@ -52,7 +52,7 @@ class CfgVehicles {
                 };
                 class ACE_MapToolsAlignCompass {
                     displayName = CSTRING(MapToolsAlignCompass);
-                    condition = 'GVAR(mapTool_Shown) != 0 && {getUnitLoadout ACE_player param [9, []] param [3, ""] != ""}';
+                    condition = QUOTE(GVAR(mapTool_Shown) != 0 && {getUnitLoadout ACE_player param [ARR_2(9, [])] param [ARR_2(3, '')] != ''});
                     statement = QUOTE(GVAR(mapTool_angle) = getDir ACE_player;);
                     exceptions[] = {"isNotDragging", "notOnMap", "isNotInside", "isNotSitting"};
                     showDisabled = 1;
--- a/addons/maptools/CfgVehicles.hpp
+++ b/addons/maptools/CfgVehicles.hpp
@@ -52,7 +52,7 @@ class CfgVehicles {
                 };
                 class ACE_MapToolsAlignCompass {
                     displayName = CSTRING(MapToolsAlignCompass);
-                    condition = QUOTE((GVAR(mapTool_Shown) != 0) && {getUnitLoadout ACE_player param [9, []] param [3, ""] != ""});
+                    condition = QUOTE((GVAR(mapTool_Shown) != 0) && {getUnitLoadout ACE_player param [9, []] param [3, ''] != ''});
                     statement = QUOTE(GVAR(mapTool_angle) = getDir ACE_player;);
                     exceptions[] = {"isNotDragging", "notOnMap", "isNotInside", "isNotSitting"};
                     showDisabled = 1;
--- a/addons/maptools/CfgVehicles.hpp
+++ b/addons/maptools/CfgVehicles.hpp
@@ -52,7 +52,7 @@ class CfgVehicles {
                 };
                 class ACE_MapToolsAlignCompass {
                     displayName = CSTRING(MapToolsAlignCompass);
-                    condition = QUOTE((GVAR(mapTool_Shown) != 0) && {getUnitLoadout ACE_player param [9, []] param [3, ''] != ''});
+                    condition = 'GVAR(mapTool_Shown) != 0 && {getUnitLoadout ACE_player param [9, []] param [3, ""] != ""}';
                     statement = QUOTE(GVAR(mapTool_angle) = getDir ACE_player;);
                     exceptions[] = {"isNotDragging", "notOnMap", "isNotInside", "isNotSitting"};
                     showDisabled = 1;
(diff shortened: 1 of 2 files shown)


addons/maptools/CfgVehicles.hpp
```diff
@@ -52,7 +52,7 @@ class CfgVehicles {
                 };
                 class ACE_MapToolsAlignCompass {
                     displayName = CSTRING(MapToolsAlignCompass);
-                    condition = QUOTE((GVAR(mapTool_Shown) != 0) && {'ItemCompass' in assigneditems ACE_player});
+                    condition = QUOTE((GVAR(mapTool_Shown) != 0) && {getUnitLoadout ACE_player param [9, []] param [3, ''] != ''});
```
PabstMirror, line 55: don't we need the `ARR_x` macros

Drofseh: I can update this, but I'm not 100% where I should use the macro

Is this correct?
`getUnitLoadout ACE_player param ARR_2(9,[]) param ARR_2(3,'')`

Do I do anything about the `[]` default value?

b-mayr-1984: > Is this correct? `getUnitLoadout ACE_player param ARR_2(9,[]) param ARR_2(3,'')`

Well it is definitely better because with this change the build runs successfully on my machine. Before it didn't. 

> Do I do anything about the `[]` default value?

Do you mean wether you should do another substitution with something like `ARR_0` (which doesn't exist) as opposed to `[]` ?

Drofseh: > > Is this correct? `getUnitLoadout ACE_player param ARR_2(9,[]) param ARR_2(3,'')`
> 
> Well it is definitely better because with this change the build runs successfully on my machine. Before it didn't.
> 
> > Do I do anything about the `[]` default value?
> 
> Do you mean wether you should do another substitution with something like `ARR_0` (which doesn't exist) as opposed to `[]` ?

That's what was confusing me yeah, it's an array, but it's an empty array. ARR macros replace square brackets so do I need to replace those with something

b-mayr-1984: > ARR macros replace square brackets so do I need to replace those with something

As far as I understand, the ARR_x macros do **not** replace the square brackets (see [here](https://github.com/acemod/ACE3/blob/master/include/x/cba/addons/main/script_macros_common.hpp#L1066-L1084)). 
They only preserve the commas that would get erased by the other macro `QUOTE` (see https://community.bistudio.com/wiki/PreProcessor_Commands#Arguments). 
Hence the code wouldn't compile without the ARR_x macros.

My takeaway from this is that the square brackets are still needed (also _around_ the ARR_ macros, not just _within_ them).
`getUnitLoadout ACE_player param [ARR_2(9,[])] param [ARR_2(3,'')] != ''`

Can anyone with more experience in SQF confirm this?
(I am just extrapolating from my C/C++ background. But I am not good with SQF and potential differences between its preprocessor and that from C/C++.)

jonpas:
```suggestion
                    condition = 'GVAR(mapTool_Shown) != 0 && {getUnitLoadout ACE_player param [9, []] param [3, ""] != ""}';
```
There you go, stop worrying about `ARR_n`. :P

b-mayr-1984: Why is code always so hard to grasp as soon as you start using macros? 😅

jonpas: Because there is a good use of macros, that help with naming, etc., and there is a bad use, that needlessly complicates things.

b-mayr-1984 (approved):
Tested on my machine with 
- GM
- VN
- Vanilla 

PR changes work well in either case. Thumbs up from my side. 
Thank you @Drofseh @commy2 @jonpas :-)

addons/maptools/CfgVehicles.hpp
```diff
@@ -52,7 +52,7 @@ class CfgVehicles {
                 };
                 class ACE_MapToolsAlignCompass {
                     displayName = CSTRING(MapToolsAlignCompass);
-                    condition = QUOTE((GVAR(mapTool_Shown) != 0) && {'ItemCompass' in assigneditems ACE_player});
+                    condition = 'GVAR(mapTool_Shown) != 0 && {getUnitLoadout ACE_player param [9, []] param [3, ""] != ""}';
```
PabstMirror, line 55: crashes game on load
```
File z\ace\addons\maptools\CfgVehicles.hpp, line 54: '/CfgVehicles/CAManBase/ACE_SelfActions/ACE_MapTools/ACE_MapToolsAlignCompass.condition': Missing ';' prior '}'
Error context ;
ErrorMessage: File z\ace\addons\maptools\CfgVehicles.hpp, line 54: '/CfgVehicles/CAManBase/ACE_SelfActions/ACE_MapTools.': ''' encountered instead of '='
Application terminated intentionally
ErrorMessage: File z\ace\addons\maptools\CfgVehicles.hpp, line 54: '/CfgVehicles/CAManBase/ACE_SelfActions/ACE_MapTools.': ''' encountered instead of '='
```

```suggestion
                    condition = QUOTE(GVAR(mapTool_Shown) != 0 && {getUnitLoadout ACE_player param [ARR_2(9, [])] param [ARR_2(3, '')] != ''});
```

PabstMirror (changes requested):

addons/maptools/CfgVehicles.hpp
b-mayr-1984: Hm weird, it didn't crash my game when I tried it. 🤔

PabstMirror: did you do full build?  this was on a filepatching build which might work differently

b-mayr-1984: Hm, that might be it.

Drofseh: @PabstMirror I'm not sure why that would crash at all, but did it only crash with filepatching?

PabstMirror: I didn't test with a full (pboProject) build.
But I think i've seen this before, that tool can handle the macros and make a binned config that works, but arma itself can't

Drofseh: Just to confirm, it's `condition = 'GVAR(mapTool_Shown) != 0 && {getUnitLoadout ACE_player param [9, []] param [3, ""] != ""}';` that is crashing?

That has fewer macros, and the only macro present is also present in `condition = QUOTE(GVAR(mapTool_Shown) != 0 && {getUnitLoadout ACE_player param [ARR_2(9, [])] param [ARR_2(3, '')] != ''});`

PabstMirror (approved):

PR #8861, opened 2022-04-04
Medical Damage - Improve falling and collision damage

pterolatypus:
- Shift wound weighting so that bruises are more likely at low damage values
- Adjust wound thresholds so that low damage values have a chance to be ignored
- Add size multipliers to wounds

**Falling**
The overall result should be fewer, larger wounds and less bandaging required, particularly for small falls.
Longer falls (more than about 4-5m) will cause larger wounds, but KO/kill potential should be unchanged.
Pain incurred from falling should be increased across the board.

**Collisions**
As above, overall effect should be fewer wounds overall and much less bleeding for very small collisions.
~Apparently armor is taken into account for collisions, so I've added a custom handler to reduce its effect and scaled down the base damage to offset this. The numbers are all just tweaked by empirical testing, repeatedly running people over in the editor.~ Unfortunately not viable as falling damage is sometimes registered as collision and this was causing it to occasionally do much more damage than intended.

### IMPORTANT


kymckay:
Should falling damage also have reduced armor influence since it's a form of collision?

pterolatypus:
>Should falling damage also have reduced armor influence

It appears to have a different effect already, which is unexpected. I suppose collisions with terrain are handled differently from collisions with objects.

Results from testing a 5m fall

| Vest | Body Damage (Approx) |
| ---- | ---- |
| None | 0.36 |
| Carrier Lite | 0.36 |
| Carrier Special Rig | 0.36 |
| GA Carrier GL Rig | 0.23 |

As opposed to hitting a dude with a car, which appears to have the full effect of their gear's `armor` value.

addons/medical_damage/functions/fnc_woundsHandlerCollision.sqf
```diff
@@ -0,0 +1,28 @@
...
+ * Arguments:
+ * 0: Unit That Was Hit <OBJECT>
+ * 1: Damage done to each body part <ARRAY>
+ * 2: Type of the damage done <STRING>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player, [[0.5, "body", 1.5]], "collision"] call ace_medical_damage_fnc_woundsHandlerVehicleCrash
```
PabstMirror, line 15:
```suggestion
 * [player, [[0.5, "body", 1.5]], "collision"] call ace_medical_damage_fnc_woundsHandlerCollision
```

PabstMirror:
Still having issues getting large multipart damage https://www.youtube.com/watch?v=uhWiwO64HV8
We should get fractures, not bleed out from a fall like that

pterolatypus:
Unfortunately whats happening there is the falling damage is being registered as collision and the armor reduction is actually significantly increasing the damage, so I've reverted that change until falling detection can be improved.

PabstMirror:
seems a lot better imho
diff on github is weird but I think it will merge ok

PabstMirror (approved):

PR #8862, opened 2022-04-05
Laser - Fix seeker owner return value

zharf:
- Fixes an issue with laser seeker return value mentioned [here](https://github.com/acemod/ACE3/pull/8531#issuecomment-954865110)
- Chooses the owner with most hits instead of the last one in the hashmap iteration order


PabstMirror (approved):

PabstMirror:
Good fix,
I am curious what you're using this for.

IIRC long a go missile guidance was originally going to hand-off guidance to the laser owner
but as far as I can tell nothing in ace uses the owner value

zharf:
I also found that nothing used it in ACE however I traced an issue in ITC AIR to this issue
https://github.com/YannikH/itc_air/blob/master/itc_air_tgp/functions/lss/search.sqf

Not sure if they should be using this either but ITC seems to be abandoned nowadays

PR #8870, opened 2022-04-09
Add Missing German Translation

Nilia119:
- Added Missing Translation in CSW
- Added Missing Translation in field_radio
- Added Missing Translation in finger
- Added Missing Translation in noradio
- Added Missing Translation in overpressure


bux (approved):
Dankeschön dafür.

PR #8874, opened 2022-04-13
Medical Damage - Add scaling for number of body parts hit from non-specific wound types

PabstMirror:
This PR adds a (setting?) to control the number of wounds created for non-selection-specific wound types.

ref a 3m fall:
```
_allDamages=[
    [0.086638,"Body",1.03966],
    [0.086638,"RightArm",0.519828],
    [0.086638,"RightLeg",0.519828],
    [0.086638,"LeftLeg",0.519828],
    [0.086638,"LeftArm",0.519828],
    [0.0187293,"Head",0.0374586],
    [0.00418826,"#structural",0.0125648]], _typeOfDamage=collision
```

resulting in wounds on all body parts (5 bleeding)
for just a 3m fall this will take a while to treat, old medical would usually just be a single wound on a leg

Old medical roughly took `0.86638` (top dmg) worth of damage and then create a number of wounds spread among the different body parts
New medical handles each part with with it's own damage source, so you almost get ~5x as much damage applied.

The new setting @ `0.0` will behave just like new medical unchanged.
@ `1.0` it will approximate old medical (same dam applied, but not spread out as much)
imho `0.3 - 0.5` seems to work very well


pterolatypus:
Personally not a big fan of this, feels too much like reverting to the old behaviour when moving away from that was half the point of the rework.

>you almost get ~5x as much damage applied

This is kind of intentional, it was far too forgiving before. I do agree that it needs to be tweaked a bit though especially at the low end; forcing the players to spend this much time bandaging is too much inconvenience for a fairly minor actual threat.
See #8861 for alternative solution.

It's weird that this fall is registered as `collision` instead of `falling`, we should keep an eye out for ways to improve fall detection.

PR #8880, opened 2022-04-19
CSW - Fix missing magazine on assembling

Dystopian:
- fix missing magazine on assembling some launchers (I know only about CUP Metis);
- fix `fnc_reload_getVehicleMagazine` header arguments.

When you place loaded with rocket CUP Metis to its internal tripod (you don't need an external one) it's assembled but without rocket in it. Rocket is just disappeared from your inventory. Also when you disassemble static Metis to portable launcher it's empty and rocket(s) is placed near you. This PR makes new launcher loaded if it was loaded before (dis)assembling.

### DIFF
--- a/addons/csw/functions/fnc_assemble_deployTripod.sqf
+++ b/addons/csw/functions/fnc_assemble_deployTripod.sqf
@@ -21,12 +21,14 @@
 
     // Remove the tripod from the launcher slot
     private _secondaryWeaponClassname = secondaryWeapon _player;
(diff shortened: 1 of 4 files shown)


PabstMirror (approved):

PabstMirror:
Works great on the cup metis

It seems like it's hardcoded to 1 round which is just fine for rockets,
in the future we could adjust it to handle belts with variable ammo
but I think this PR is fine as is

PR #8891, opened 2022-04-24
Medical Vitals - Add event to vitals loop

pterolatypus:
- Add a local event in the main vitals loop so that extension mods can run updates in-sync without having to write their own loop.

### IMPORTANT

### DIFF
--- a/addons/medical_vitals/functions/fnc_handleUnitVitals.sqf
+++ b/addons/medical_vitals/functions/fnc_handleUnitVitals.sqf
@@ -165,4 +165,7 @@ if (!isPlayer _unit) then {
 
 END_COUNTER(Vitals);
 
+//placed outside the counter as 3rd-party code may be called from this event
+[QEGVAR(medical,handleUnitVitals), _unit] call CBA_fnc_localEvent;
+
 true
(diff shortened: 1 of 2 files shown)


addons/medical_vitals/functions/fnc_handleUnitVitals.sqf
```diff
@@ -165,4 +165,7 @@ if (!isPlayer _unit) then {
 
 END_COUNTER(Vitals);
 
+//placed outside the counter as 3rd-party code may be called from this event
+[QEGVAR(medical,handleUnitVitals), _unit] call CBA_fnc_localEvent;
```
mharis001, line 169: Thoughts on providing `_deltaT` here as well for convenience?

pterolatypus: good call

pterolatypus:
I feel like this should be documented, but I'm not sure where to put it as there isn't really any doc for vitals yet.

PabstMirror (approved):

jonpas:
@pterolatypus Can put it on the Events page, where there is a list of API events.

PR #8903, opened 2022-05-05
Medical - Add AI Unconsciousness exception (when using zeus module)

YetheSamartaka:
- Adds AI Unconsciousness exception option. It allows you to put AI into unconsciousness via Zeus Module even though the AI unconsciousness is disabled. This "feature" was possible before the medical rewrite and allowed to put selected AI units into unconsciousness even though the overall AI unconsciousness was disabled. This was very handy and many groups missed this ability including myself.
- Special thanks to Pterolatypus for consultation.


addons/zeus/functions/fnc_moduleUnconscious.sqf
```diff
@@ -38,6 +38,12 @@ if (isNil QEFUNC(medical,setUnconscious)) then {
                 [LSTRING(OnlyAlive)] call FUNC(showMessage);
             } else {
                 private _unconscious = GETVAR(_unit,ACE_isUnconscious,false);
+				if (ace_medical_statemachine_AIUnconsciousness_Exception) then {
+                _unit setVariable ["ACE_AI_Unconsciousness_ExceptionVar", true, true];
+					if (_unconscious) then {
+					_unit setVariable ["ACE_AI_Unconsciousness_ExceptionVar", false, true];	
+					};
```
rautamiekka:
```suggestion
                    _unit setVariable ["ACE_AI_Unconsciousness_ExceptionVar", true, true];
					if (_unconscious) then {
					    _unit setVariable ["ACE_AI_Unconsciousness_ExceptionVar", false, true];	
					};
```

Salluci:
I'd refactor this to handle the entirety of the statemachine condition inside a function if one is going to be called at all.

Don't particularly like the variable name, but meh.

Change in the module function should use EGVAR macro.

YetheSamartaka:
> I'd refactor this to handle the entirety of the statemachine condition inside a function if one is going to be called at all.
> 
> Don't particularly like the variable name, but meh.
> 
> Change in the Zeus module function should use EGVAR macro.

1) I tried to have that **_this getVariable and so on** inside the statemachine condition, but I encountered difficulties.
2) I'm aware that it is not the best variable name. I'm open to suggestions.
3) Do you mean replacing **ace_medical_statemachine_AIUnconsciousness_Exception** to the **EGVAR(medical_statemachine,AIUnconsciousness_Exception)**?

mharis001 (changes requested):
I don't think this should have a setting; it seems confusing. If the only goal is to allow for unconsciousness to be forced (for the Zeus module), let's add logic that specifically allows that without requiring a setting to be set.

kymckay:
> I don't think this should have a setting; it seems confusing. If the only goal is to allow for unconsciousness to be forced (for the Zeus module), let's add logic that specifically allows that without requiring a setting to be set.

Was about to comment the exact same, a setting adds no value here. If we add the logic and someone doesn't want to unconscious AI via zeus they can just not do it.

YetheSamartaka:
> > I don't think this should have a setting; it seems confusing. If the only goal is to allow for unconsciousness to be forced (for the Zeus module), let's add logic that specifically allows that without requiring a setting to be set.
> 
> Was about to comment the exact same, a setting adds no value here. If we add the logic and someone doesn't want to unconscious AI via zeus they can just not do it.

As requested, the addon option was removed. I've also adjusted the variable name.

addons/medical_statemachine/Statemachine.hpp
```diff
@@ -46,7 +46,7 @@ class ACE_Medical_StateMachine {
         onStateEntered = QFUNC(enteredStateUnconscious);
         class DeathAI {
             targetState = "Dead";
-            condition = QUOTE(!GVAR(AIUnconsciousness) && {!isPlayer _this});
+            condition = QUOTE(!(GVAR(AIUnconsciousness) || (_this getVariable [ARR_2(QQEGVAR(zeus,AIUnconsciousException), false)])) && {!isPlayer _this});
```
pterolatypus, line 49:
```suggestion
            condition = QUOTE(!(_this getVariable [ARR_2(QQGVAR(AIUnconsciousness), GVAR(AIUnconsciousness))]) && {!isPlayer _this});
```

This would allow the variable to override the setting in both directions.

addons/zeus/functions/fnc_moduleUnconscious.sqf
```diff
@@ -38,6 +38,10 @@ if (isNil QEFUNC(medical,setUnconscious)) then {
                 [LSTRING(OnlyAlive)] call FUNC(showMessage);
             } else {
                 private _unconscious = GETVAR(_unit,ACE_isUnconscious,false);
+                _unit setVariable [QGVAR(AIUnconsciousException), true, true];
+                    if (_unconscious) then {
+                    _unit setVariable [QGVAR(AIUnconsciousException), false, true];
+                    };
```
pterolatypus, line 44:
```suggestion
                if (_unconscious) then {
                    _unit setVariable [QEGVAR(medical_statemachine,AIUnconsciousness), nil, true];
                } else {
                    _unit setVariable [QEGVAR(medical_statemachine,AIUnconsciousness), true, true];
                };
```

No point setting it to true if you're immediately going to change it.

YetheSamartaka: That is to prevent the situation where you would use the Zeus module on an AI dropping it to unconsciousness and then after a while, you would wake up the same AI again using the Zeus module, and then afterward, someone would shoot the AI into unconsciousness(With the AI unconsciousness turned off) which would not kill the AI as it should.

addons/medical_statemachine/Statemachine.hpp
YetheSamartaka: This suggestion in this form is not working in-game. The part with ARR_2 is wrong, it is not using the correct variable that is set by the Zeus module. I do not have experience with comma operator, so for the time being, I' am unable to fix it.

kymckay (approved):

mharis001 (commented):
How does this work with the `CardiacArrest` state, which has a similar condition?

YetheSamartaka:
> How does this work with the `CardiacArrest` state, which has a similar condition?

This exception only targets the Zeus module for unconsciousness, so it should not affect the CardiacArrest at all.

PabstMirror (approved):

mharis001 (approved):

PR #8904, opened 2022-05-08
Common - Add separate blockSpeaking status for acre direct

tbeswick96:
- Add new status effect `blockSpeaking` to block acre direct voice
- Set `blockSpeaking` in `setUnconsciousState`

### IMPORTANT

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -27,6 +27,7 @@
 ["blockThrow", false, ["ACE_Attach", "ACE_concertina_wire", "ACE_dragging", "ACE_Explosives", "ACE_Ladder", "ACE_rearm", "ACE_refuel", "ACE_Sandbag", "ACE_Trenches", "ACE_tripod"]] call FUNC(statusEffect_addType);
 ["setHidden", true, ["ace_unconscious"]] call FUNC(statusEffect_addType);
 ["blockRadio", false, [QEGVAR(captives,Handcuffed), QEGVAR(captives,Surrendered), "ace_unconscious"]] call FUNC(statusEffect_addType);
+["blockSpeaking", false, ["ace_unconscious"]] call FUNC(statusEffect_addType);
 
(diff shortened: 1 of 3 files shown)


jonpas (approved):

addons/common/XEH_postInit.sqf
```diff
@@ -86,6 +87,14 @@
     };
 }] call CBA_fnc_addEventHandler;
 
+[QGVAR(blockSpeaking), {
+    params ["_object", "_set"];
+    TRACE_2("blockSpeaking EH",_object,_set);
+    if (["acre_main"] call FUNC(isModLoaded)) then {
```
tbeswick96, line 93: I thought about adding the tfar setVariables here too instead of in `setVolume`, but I'm not familiar enough with it to know whether `tf_voiceVolume` is hearing or speaking volume

PabstMirror: I moved the code over so things are consistant between the 2 mods
I tested by watching `[player, true, "guy"] call TFAR_fnc_preparePositionCoordinates;` 
and verified that the output changes
ref https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/master/addons/core/functions/fnc_preparePositionCoordinates.sqf#L103
(although this version is different from the one on steam, but both seem to work the same)

PabstMirror (approved):

addons/common/XEH_postInit.sqf
tbeswick96: Nice, thanks!

addons/common/functions/fnc_setVolume.sqf
```diff
@@ -31,7 +31,6 @@ if (_setVolume) then {
     2 fadeSound NORMAL_LEVEL;
 
     // TFAR
-    _unit setVariable ["tf_voiceVolume", NORMAL_LEVEL, true];
```
PabstMirror, line 34: also good to move this public setVar behind a isModLoaded check

PR #8912, opened 2022-05-15
Medical Treatment - Change stitching when wound reopening disabled

Freddo3000:
- Allow stitching when "Clear Trauma" is set to "On Stitch".
- This changes behaviour as such that trauma is cleared upon finishing stitching, instead of periodically, with time calculation based on `ΣbodyPartDamage * GVAR(woundStitchTime)`. This is due to some of the values otherwise used for stitching are only available when wound reopening is enabled.

### IMPORTANT

### DIFF
--- a/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
+++ b/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp
@@ -285,7 +285,7 @@ class GVAR(actions) {
         medicRequired = QGVAR(medicSurgicalKit);
         treatmentTime = QFUNC(getStitchTime);
         condition = QFUNC(canStitch);
-        callbackSuccess = "";
+        callbackSuccess = QFUNC(surgicalKitSuccess);
         callbackProgress = QFUNC(surgicalKitProgress);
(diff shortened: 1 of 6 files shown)


kymckay:
I think in terms of end user experience this isn't the ideal solution (since if action is interrupted that progress will be lost). However, it does identify a valid case that's not being handled and seems fine as an initial solution.

Having recently reacquainted myself with our stitching code to procude #8926, the way we handle stitching logic is not very intuitive and seems over complicated. I reckon if someone was motivated they could refactor the underlying system in a way which also accommodates this case with incremental progress.

Freddo3000:
> I think in terms of end user experience this isn't the ideal solution (since if action is interrupted that progress will be lost). However, it does identify a valid case that's not being handled and seems fine as an initial solution.
> 
> Having recently reacquainted myself with our stitching code to produce #8926, the way we handle stitching logic is not very intuitive and seems over complicated. I reckon if someone was motivated they could refactor the underlying system in a way which also accommodates this case with incremental progress.

This was essentially my thought on it as well, a stopgap solution until someone who is more acquainted with ACE produces something better.

PR #8915, opened 2022-05-17
Cargo - Limit rename text length

### LINKED ISSUE #8834: BUG - Potential Buffer Overflow or DOS when object cargo renaming.
- **Arma 3:** `2.06` [e.g. 1.00 stable, rc,  dev]
- **CBA:** `3.15.6.211004` [e.g. 3.0.0 stable, commit hash]
- **ACE3:  3.14.1

**Description:**

**Steps to reproduce:**
_Follow [https://ace3mod.com/img/wiki/user/issue_flowchart.png](this flowchart)!_

1. Go ace interact with any object that you can rename
2. Rename object with any size length of string -> any size but for example do a 2 gb size string
3. Arma starts to slow down
4. Arma has crashed on the client, and then server crashes

**Expected behavior:**
CHECK THE LENGTH OF THE STRING BEFORE SAVING IT IN NAMESPACE

- Dedicated 

**Log Files:**
Can't post logs here will dm someone the logs

**Additional context:**
Add any other context about the problem here. Rename group might have a similar problem

**Screenshots:**
Client crashed on the client who did it. https://cdn.discordapp.com/attachments/949162918270894110/949162930895745034/unknown.png

PabstMirror:
Fix #8834

### DIFF
--- a/addons/cargo/functions/fnc_renameObject.sqf
+++ b/addons/cargo/functions/fnc_renameObject.sqf
@@ -19,6 +19,7 @@ private _display = uiNamespace getVariable QGVAR(menuDisplay);
 if (isNil "_display") exitWith {};
 
 private _ctrlEditText = ctrlText 100;
+_ctrlEditText = _ctrlEditText select [0,32];
 if (_ctrlEditText isEqualTo "") then { // custom name has been removed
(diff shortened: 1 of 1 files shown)


jonpas (approved):

dedmen:
There is `maxChars` config entry for text boxes. That would be more obvious to the user when he cannot enter any further characters.

PR #8925, opened 2022-05-27
Advanced Vehicle Damage - Change default setting to disabled

ZluskeN:
As stated in https://github.com/acemod/ACE3/pull/8734#discussion_r787918244, Maintainers have decided that Advanced Vehicle Damage should disabled by default.

- Change default setting of advanced vehicle damage to false
- Allow https://github.com/acemod/ACE3/pull/8734 to be merged

### DIFF
--- a/addons/vehicle_damage/initSettings.sqf
+++ b/addons/vehicle_damage/initSettings.sqf
@@ -2,7 +2,7 @@
     QGVAR(enabled), "CHECKBOX",
     [ELSTRING(common,Enabled), LSTRING(setting_description)],
     LSTRING(category_displayName),
-    true, // default value
+    false, // default value
     true, // isGlobal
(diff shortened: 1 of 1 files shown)


veteran29 (approved):

PabstMirror:
I'm not sure we should be changing defaults now
anyone who wants it off will have already changed settings

Drofseh:
> 

ZluskeN is just doing what he was told, but it confuses me as well.
I would like to see Advanced Vehicle Damage enabled by default, but also have #8734 be enabled by default.
That way we still get advanced damage, but scripts that expect vehicles to be dead still function.

veteran29:
For the time being, we want the system off by default, it has some issues and brings compatibility problems.

PR #8926, opened 2022-05-29
Medical - Change medical to use hashmaps for wound storage

### LINKED ISSUE #6468: Optimise unit wound storage
The way we currently store wounds on units (in `QEGVAR(medical,openWounds)`) means we have to loop through the whole array at many points in the code to filter out the ones we aren't interested in.

There's already a comment to remove one element from the wounds array (the ID which is seemingly never used), but beyond that I believe we could be storing these in a much more efficient way if we were using some form sorting or key/value system to speed up retrieval.

This is fairly low priority and would be a nice refactor to do once functionality is all in place.

kymckay:
- We most frequently want to access wounds by body part, so this makes
  that a constant time lookup.
- The body part index is no longer stored in every wound since it's
  inherent in the wound storage.
- Using body part names as the keys of the hashmap to improve code
  clarity (no more magic numbers).

closes #6468

I've tested these relevant parts of medical which access the wound structures:

- [x] Wound creation still works.
- [x] Menu and action displays still work.
- [x] Damage effect updates (limping).
- [x] Bandaging still works.
- [x] Stitching still works (inc. testing incremental behaviour).
- [x] Wound reopening still works.
- [x] AI treatment still works.
- [x] Litter creation still works.

### IMPORTANT


kymckay:
The most awkward part to rework with hashmaps was stitching so give that code particular attention when reviewing. I've tested in pretty extensively though.

kymckay:
One consideration I just had is that this will break de-serialization across versions. I should probably add some migration code there to convert any array into a hashmap appropriately.

addons/medical_treatment/functions/fnc_surgicalKitProgress.sqf
LinkIsGrim, line 1: This whole file has gotten really confusing now if you don't know exactly what you're doing, but that's a problem for another day. Sutures consumed as intended without progress being broken now. Rest LGTM.

LinkIsGrim (approved):

addons/medical/functions/fnc_deserializeState.sqf
```diff
@@ -35,6 +35,22 @@ if !(_unit getVariable [QGVAR(initialized), false]) exitWith {
 
 private _state = [_json] call CBA_fnc_parseJSON;
 
+// Migration from old array wounding storage serialized in old versions (<= 3.15.0)
```
kymckay, line 38: Before merge this comment will need updating to whatever latest state is. Because I'm not currently involved in our release management strategy I think someone else who is should make the call on merging this.

tbeswick96:
The serialize and deserialize functions don't work with this due to CBA_fnc_parseJSON not handling native hashmaps properly
Try this code in debug console:
```sqf
private _inputData = call CBA_fnc_createNamespace; 
_inputData setVariable ["hashmap", createHashmap]; 

private _json = [_inputData] call CBA_fnc_encodeJSON;

private _outputData = [_json] call CBA_fnc_parseJSON; 
private _hashmap = _outputData getVariable "hashmap";
_hashmap
```
and it outputs
`Location CBA_NamespaceDummy at -1000, -1000`

Which results in the wound hashmaps being the wrong type on data load

PR #8928, opened 2022-05-31
Cookoff - Fix ammo not being removed from turrets with additional FFV

veteran29:
- Fix ammo not being removed from turrets which have "additional" FFV, eg. in tanks after turning out from gunner pos


TheCandianVendingMachine:
oh awesome I thought that was another command

TheCandianVendingMachine (approved):

PabstMirror (approved):

PR #8933, opened 2022-06-05
Cookoff - Fix smoke effects not getting cleaned up

JonBons:
- Adds a `Deleted` EH to vehicles such that any children effects get cleaned up if it's being removed mid-cookoff

### DIFF
--- a/addons/cookoff/XEH_postInit.sqf
+++ b/addons/cookoff/XEH_postInit.sqf
@@ -11,6 +11,16 @@
 [QGVAR(smoke), FUNC(smoke)] call CBA_fnc_addEventHandler;
 [QGVAR(cookOffBox), FUNC(cookOffBox)] call CBA_fnc_addEventHandler;
 
+// handle cleaning up effects when vehicle is deleted mid-cookoff 
+[QGVAR(addCleanupHandlers), {
+    params ["_vehicle"];
+
+    _vehicle addEventHandler ["Deleted", {
+        params ["_vehicle"];
+        [QGVAR(cleanupEffects), [_vehicle]] call CBA_fnc_globalEvent;
+    }];
+}] call CBA_fnc_addEventHandler;
+
 [QGVAR(cleanupEffects), {
     params ["_vehicle", ["_effects", []]];
 
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -30,6 +30,8 @@ TRACE_9("cooking off",_vehicle,_intensity,_instigator,_smokeDelayEnabled,_ammoDe
 if (_vehicle getVariable [QGVAR(isCookingOff), false]) exitWith {};
 _vehicle setVariable [QGVAR(isCookingOff), true, true];
 
+[QGVAR(addCleanupHandlers), [_vehicle]] call CBA_fnc_localEvent;
+
 // limit maximum value of intensity to prevent very long cook-off times
 _intensity = _intensity min _maxIntensity;
 


addons/cookoff/XEH_postInit.sqf
```diff
@@ -11,6 +11,16 @@
 [QGVAR(smoke), FUNC(smoke)] call CBA_fnc_addEventHandler;
 [QGVAR(cookOffBox), FUNC(cookOffBox)] call CBA_fnc_addEventHandler;
 
+// handle cleaning up effects when vehicle is deleted mid-cookoff 
+[QGVAR(addCleanupHandlers), {
+    params ["_vehicle"];
```
veteran29, line 16: There should be a guard to prevent the EH from being added multiple times, or the EH should be removed after cook off is done. Vehicle can cook off again if it's reparied/rearmed.

veteran29 (changes requested):

PabstMirror (approved):

addons/cookoff/XEH_postInit.sqf
```diff
@@ -11,6 +11,21 @@
...
+    
+    // Don't add a new EH if cookoff is run multiple times
+    if ((_vehicle getVariable [QGVAR(deletedEH), -1]) == -1) then {
+        private _deletedEH = _vehicle addEventHandler ["Deleted", {
+            params ["_vehicle"];
+            [QGVAR(cleanupEffects), [_vehicle]] call CBA_fnc_globalEvent;
```
veteran29, line 22: Isn't deleted firing everywhere it's added not only where the object is local? Shouldn't this be a local event?

JonBons: So my chain of logic for this was that the CBA `cookOff` event has its own `local _vehicle` check before executing `ace_cookoff_fnc_cookOff` which triggers a local event for `addCleanupHandlers` and thus the EH code will only exist and execute on the owner of the vehicle which is why the `CBA_fnc_globalEvent` is utilized here. I could add a locality check to the `addCleanupHandlers` if necessary?

JonBons: I'll go ahead and add it just incase somehow some code ends up here without being the owner.

kymckay (approved):

addons/cookoff/functions/fnc_cookOff.sqf
```diff
@@ -30,6 +30,8 @@ TRACE_9("cooking off",_vehicle,_intensity,_instigator,_smokeDelayEnabled,_ammoDe
 if (_vehicle getVariable [QGVAR(isCookingOff), false]) exitWith {};
 _vehicle setVariable [QGVAR(isCookingOff), true, true];
 
+[QGVAR(addCleanupHandlers), [_vehicle]] call CBA_fnc_localEvent;
```
veteran29, line 33: 1. zeus spawns vehicle
2. vehicle cookoffs
3. zeus disconnects
4. something deletes vehicle
5. effects stay there?

shouldn't this be global?

If so the locality check in deleted EH could be dropped and event should be local then.

veteran29 (changes requested):

addons/cookoff/functions/fnc_cookOff.sqf
kymckay: This is a good point, but isn't the simplest way to do this to add the EH on the server as a server event? This way no matter where the vehicle ends up local at time of deletion, the global cleanup event can be fired from a single location which will always exists for the duration of the session.

kymckay: To clarify that: The deleted EH needs to run once and send out a global event for cleanup to handle JIPs where it may not have been first added (unless I'm misremembering my Arma 3 locality logic).

veteran29: True, server event would be the best. Add cleanup handler on server, emit global event from server.

JonBons: I'll take a look at reworking this this weekend, totally slipped my mind it'd need to handle locality changing hands.

JonBons:
This should be good to go now.

PR #8945, opened 2022-06-24
Logistics Wirecutter - Add new fences models for the wirecutter

formtapez:
- make the wirecutter able to cut through all new wire fences. (there was some addition to the fences of vanilla ArmA (e.g. at the airport on Altis).
- fences for @cup_terrains_core are also included.


formtapez:
> Would you add any other missing fences (from the DLC's and other mods ACE3 has compats for) since we're on this, please ? I can look for the missing ones to help.

Sure, if you tell me which Mods i have to check.
I think the DLCs should be covered, because i searched "fence" in the editor and placed every object which made sense to cut,

PabstMirror (approved):

mharis001 (approved):

PR #8943, opened 2022-06-21
Add placeable and carryable flags

Timi007:
Hello everyone, I would like to donate a component of one of my mods to ACE! Since Brett already did something like this, I took his work as a template and rewrote my component.

**When merged this pull request will:**
- Add flags which can be placed or carried by a unit
  - White
  - Red
  - Blue
  - Green
  - Yellow
  - Orange
  - Purple
  - Black

flags

PR #8951, opened 2022-06-30
Marker Flags - Add placing system & icons

Timi007:
- Add placing system same as in #8943
- Add new fancy icons
- Adjust colors of flags so that my eyes don't hurt


BrettMayson:
Are we not able to use a single white icon here. and set the color value? Rather than an icon for each color?

Timi007:
We could but doesn't look as good. I also wanted it to be consistent with #8943.

BrettMayson (approved):
Really like the icons. Thanks for the contribution!

PR #8953, opened 2022-07-01
Cookoff - Add ability to disable fire jet effect in configs

JonBons:
The main purpose is to fix vehicles like a vodnik spawning the fire jet effect because it doesn't really make any sense and looks odd.

- Adds `ace_cookoff_canHaveFireJet` config number value that determines if its allowed to create a fire jet during cookoff (no set value defaults to enabled)

### DIFF
--- a/addons/cookoff/functions/fnc_cookOffEffect.sqf
+++ b/addons/cookoff/functions/fnc_cookOffEffect.sqf
@@ -32,8 +32,10 @@ private _sound = objNull;
 if (isServer) then {
     // ironically biggest performance hit is this. Creating a new sound source takes up aprox 400 milliseconds.
     // I dont think there is an alternative that takes into effect distance and whatever, but if you find one please fix!
-    private _soundName = selectRandomWeighted [QGVAR(Sound_low), 0.1, QGVAR(Sound_mid), 0.25, QGVAR(Sound_high), 0.65];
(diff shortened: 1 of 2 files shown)


veteran29 (approved):
LGTM

addons/cookoff/functions/fnc_cookOffEffect.sqf
```diff
@@ -32,8 +32,10 @@ private _sound = objNull;
 if (isServer) then {
     // ironically biggest performance hit is this. Creating a new sound source takes up aprox 400 milliseconds.
     // I dont think there is an alternative that takes into effect distance and whatever, but if you find one please fix!
-    private _soundName = selectRandomWeighted [QGVAR(Sound_low), 0.1, QGVAR(Sound_mid), 0.25, QGVAR(Sound_high), 0.65];
-    _sound = createSoundSource [_soundName, position _obj, [], 0];
+    if (_jet || _ring) then {
```
rautamiekka, line 35:
```suggestion
    if (_jet || {_ring}) then {
```

veteran29: Lazy eval is not always better, when you have 2 bools it's more expensive to create the code for lazy eval than just using them.

kymckay (approved):

PR #8963, opened 2022-07-09
Towing - Add rope to vehicle inventory

Dystopian:
- add setting to add rope to all `Car` and `Tank` vehicles inventory.

Cars get 6m rope; tanks, APCs and trucks get 12m rope. There is `ace_cargo_hasCargo` config value check to handle all those quadbike/cart/etc vehicles.

Setting needs restart only on disabling.

Not completely sure about:
- default setting value (enabled now);
- setting category (`Logistics` now, can be `Vehicles`);
- should add caching for vehicle classes (I don't think it's worth it);

At least some CUP trucks have `Car_F` parent instead of `Truck_F` so they get 6m rope but I don't think it's a big problem.

### DIFF
--- a/addons/towing/functions/fnc_addRopeToVehicle.sqf
+++ b/addons/towing/functions/fnc_addRopeToVehicle.sqf
@@ -15,6 +15,7 @@
  * Public: No
  */
 
+if (!GVAR(addRopeToVehicleInventory)) exitWith {};
 params ["_vehicle"];
 
 if (0 == getNumber (configOf _vehicle >> QEGVAR(cargo,hasCargo))) exitWith {};
--- a/addons/towing/functions/fnc_addRopeToVehicle.sqf
+++ b/addons/towing/functions/fnc_addRopeToVehicle.sqf
@@ -1,0 +1,26 @@
+#include "script_component.hpp"
+/*
+ * Author: Dystopian
+ * Adds rope to vehicle inventory.
+ *
+ * Arguments:
+ * 0: Vehicle <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * cursorObject call ace_towing_fnc_addRopeToVehicle
+ *
+ * Public: No
+ */
+
+params ["_vehicle"];
+
+if (0 == getNumber (configOf _vehicle >> QEGVAR(cargo,hasCargo))) exitWith {};
+
+private _ropeType = if (
+    -1 < ["Tank", "Wheeled_APC_F", "Truck_F"] findIf {_vehicle isKindOf _x}
+) then {"ACE_rope12"} else {"ACE_rope6"};
+
+_vehicle addItemCargoGlobal [_ropeType, 1];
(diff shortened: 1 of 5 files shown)


addons/towing/functions/fnc_addRopeToVehicle.sqf
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * cursorObject call ace_towing_fnc_addRopeToVehicle
+ *
+ * Public: No
+ */
+
```
PabstMirror, line 17: Adding this check should make setting fully changeable at runtime

```suggestion

if (!GVAR(addRopeToVehicleInventory)) exitWith {};
```

PabstMirror:
Looks good
default setting could go either way, I think true is probably fine

addons/towing/functions/fnc_addRopeToVehicle.sqf
Dystopian: Only for new vehicles created after setting disabling I suppose? Will it be consistent with existing vehicles which already have rope in inventory?

addons/towing/initSettings.sqf
```diff
@@ -0,0 +1,17 @@
...
+    true,
+    true,
+    {
+        if (!_this) exitWith {
+            [QGVAR(addRopeToVehicleInventory), _this] call EFUNC(common,cbaSettings_settingChanged);
+        };
+        if (!isServer || {!isNil QGVAR(addRopeToVehicleInventory_initialized)}) exitWith {};
+        GVAR(addRopeToVehicleInventory_initialized) = true;
+        ["Tank", "initPost", {call FUNC(addRopeToVehicle)}, true, [], true] call CBA_fnc_addClassEventHandler;
+        ["Car", "initPost", {call FUNC(addRopeToVehicle)}, true, [], true] call CBA_fnc_addClassEventHandler;
```
mharis001, line 14: For CBA XEH, don't need the `call` wrapper optimization like BI EHs. Can just use `FUNC` directly or `LINKFUNC` (preferred).

Dystopian: forgot about `LINKFUNC`. Will fix in 2 weeks (can't test atm) if you don't fix earlier

addons/towing/functions/fnc_addRopeToVehicle.sqf
veteran29: Ofc if rope was already added to vehicles it will stay there. I would not concern about "will it be consistent".

veteran29 (approved):
LGTM with Pabst suggestion.

addons/towing/functions/fnc_addRopeToVehicle.sqf
Dystopian: what about setting `needRestart` argument then?

veteran29: I don't see a need for it, the mission does not need to be restarted not to have the ropes being added to the vehicles.

Dystopian (commented):
not tested but looks fine

PabstMirror (approved):

PR #8966, opened 2022-07-13
Medical - Add an option to consider select AIs as players for medical system

RusComBear:
- Add a setting that allows to consider select AIs as players for medical system.
Available options are: disabled, all playable AIs, all AIs in player squad.
- Add an Eden attribute that allows to consider specific units as either players or AI for medical system. Attribute will override mission/server setting and won't affect players.

**More detailed explanation:**
Affected AIs:
- Will use player setting for critical damage threshold
- Will use player setting for fatal injuries
- Will be able to go unconscious regardless of AI settings
- Can survive fatal injury if "fatal injury death chance" setting is set lower than 100% (btw, this isn't mentioned anywhere but, according to the wound handler code, normally this setting is ignored for AI, hence this change)

This setting is meant to improve single player experience by making only select allies/squadmates able to go unconscious, make them tougher with damage threshold settings, etc. Two options for flexibility and easy integration into existing scenarios without any mission editing. As maintainers suggested, i will also look into adding this as an Eden attribute for specific units.
_Edit:_ Eden attribute has been added.

I already tested all the options in several missions hosted in singleplayer and local multiplayer: so far everything works as intended and there is no notable perfomance impact.
Might need to test it on dedicated server (i have reason to believe that player squadmates option might not work properly for clients in some cases), but since this feature meant for singleplayer, its not really important.

**TODO:**
- [x] Add an Eden attribute to override this setting for specific units (either always consider them as player or as AI)
- [ ] Add explanation of this setting in the ACE documentation(?)

### IMPORTANT

### DIFF
--- a/addons/medical_engine/functions/fnc_conditionIsConsideredPlayer.sqf
+++ b/addons/medical_engine/functions/fnc_conditionIsConsideredPlayer.sqf
@@ -17,9 +17,7 @@
 
 params ["_unit"];
 
-if (isPlayer _unit) then {
-	true
-} else {
+isPlayer _unit || {
 	// check if unit medical settings are overriden via eden atribute
 	switch (_unit getVariable [QEGVAR(medical,OverrideAIMedicalSettings), 0]) do {
 		case 0: {


addons/medical_engine/functions/fnc_conditionIsConsideredPlayer.sqf
```diff
@@ -0,0 +1,37 @@
...
+           } else {
+	          if (_unit in switchableUnits) then {true} else {false};
+                                   };
+                               };
+	    case 2:  { 
+           if (_unit in units player) then {true} else {false}
+                                        };
+	    default { false }; //Default code should never be entered, but just in case
+     };
+}
```
rautamiekka, line 37:
```suggestion
if (isPlayer _unit) then {
    true
} else {
    switch (EGVAR(medical,considerSelectAIPlayersForMedical)) do {
	    case 0: {false};
	    case 1: { 
            if (isMultiplayer) then {
                [false, true] select (_unit in playableUnits);
            } else {
                [false, true] select (_unit in switchableUnits);
            };
        };
	    case 2: { 
            [false, true] select (_unit in units ACE_player);
        };
	    default {false}; //Default code should never be entered, but just in case
    };
}
```

addons/medical_engine/functions/fnc_conditionIsConsideredPlayer.sqf
```diff
@@ -0,0 +1,37 @@
...
+params ["_unit"];
+
+if (isPlayer _unit) then {
+    true
+} else {
+    switch (EGVAR(medical,considerSelectAIPlayersForMedical)) do {
+	    case 0: {false};
+	    case 1: { 
+            if (isMultiplayer) then {
+                [false, true] select (_unit in playableUnits);
```
veteran29, line 27:
```suggestion
                _unit in playableUnits // return
```

addons/medical_engine/functions/fnc_conditionIsConsideredPlayer.sqf
```diff
@@ -0,0 +1,37 @@
...
+            } else {
+                [false, true] select (_unit in switchableUnits);
```
veteran29, line 29:
```suggestion
               _unit in switchableUnits // return
```

addons/medical_engine/functions/fnc_conditionIsConsideredPlayer.sqf
```diff
@@ -0,0 +1,37 @@
...
+            };
+        };
+	    case 2: { 
+            [false, true] select (_unit in units ACE_player);
```
veteran29, line 33:
```suggestion
            _unit in units ACE_player // return
```

veteran29 (changes requested):

addons/medical_engine/functions/fnc_conditionIsConsideredPlayer.sqf
rautamiekka: These are even better. Too obvious for me to realize.

RusComBear:
**Update:** as promised, i have added an Eden attribute. There is also minor fix for stringtable and general improvements to code readability for main function.

Eden attribute allows to always consider specific unit as either player or AI, overriding general mission/server settings. Attribute only affects AI. Theoretically, i can make it work for players too, so that mission makers could make specific players use AI medical settings, but i can't really imagine why anyone would want this, so i think its better to keep it as it is, to avoid unnecessary checks for players.
The harderst part was to came up with a good attribute name and description that won't be too complicated. I think it might ended up a bit too long, but i'm running out of ideas at this point. If you have any ideas to improve it, feel free to make a suggestion.

One last thing: i wanted to add description of new setting to ACE documentation, but i'm not sure what category i should put it in. It kind of affects both wounds and states? Or maybe add it to AI section? It doesn't really affects AI itself, but kinda related? I need advice on this one.

YetheSamartaka:
This is awesome. Would it be also possible to add it to Zeus(As a module or some elegant way in unit properties), so you could add/modify AI units on the fly?

addons/medical_engine/functions/fnc_conditionIsConsideredPlayer.sqf
```diff
@@ -0,0 +1,45 @@
...
+				case 2: { 
+					_unit in units ACE_player // return
+				};
+				default {false}; // failsafe, just in case
+			};
+		};
+		case 1: {true};
+		case 2: {false};
+	};
+}
```
mharis001, line 45:
```sqf
isPlayer _unit || {
    // ...
}
```

RusComBear:
> This is awesome. Would it be also possible to add it to Zeus(As a module or some elegant way in unit properties), so you could add/modify AI units on the fly?

I don't see any reason why it shouldn't be possible, just make a module that allows to change same variable eden attribute uses. Not sure about adding it to unit properties directly. Wouldn't that conflict with mods like Zeus Enhanced?
In any case, i don't have any experience modifying Zeus. Might look into it eventually, but no guarantees. Anyone more experienced with this is free to try.

YetheSamartaka:
> 

Zeus module would probably be the easiest way to go. It is not complicated, you can take a look at how it is done for other modules here in ACE.

kymckay:
Sounds out of scope for this PR and something that should be done as a follow up if this is merged

YetheSamartaka:
> 

I can't agree as there is already a 3DEN attribute for it. Imagine if the situation was reversed and the only way would be to do it in Zeus(Excluding the manual variable assignment for example via init field) using a module. Editor and Zeus are both essential.

LinkIsGrim:
Might be a good idea to cache the check in the unit after the first run?

johnb432:
> Might be a good idea to cache the check in the unit after the first run?

If the setting is player squad mates, I don't think it would be wise to cache it, as the player might change groups or gain/lose AI over time.

PR #8971, opened 2022-07-19
Refuel - Add Progress Bar Duration Setting

654wak654:
- Change the static `REFUEL_PROGRESS_DURATION` macro to a new setting called `progressDuration` so it can be changed as desired.
- Remove the `TIME_PROGRESSBAR` macro wrapping `REFUEL_PROGRESS_DURATION`, as the new setting can be used when debugging now.
- Also do a minor Turkish language fix.


addons/refuel/functions/fnc_moduleRefuelSettings.sqf
```diff
@@ -23,5 +23,6 @@ if !(_activated) exitWith {};
 
 [_logic, QGVAR(rate), "rate"] call EFUNC(common,readSettingFromModule);
 [_logic, QGVAR(hoseLength), "hoseLength"] call EFUNC(common,readSettingFromModule);
+[_logic, QGVAR(progressDuration), "progressDuration"] call EFUNC(common,readSettingFromModule);
 
-INFO_2("Refuel Module Initialized with flow rate: %1 - hoseLength: %2",GVAR(rate), GVAR(hoseLength))
+INFO_2("Refuel Module Initialized with flow rate: %1 - hoseLength: %2 - progressDuration: %3",GVAR(rate),GVAR(hoseLength),GVAR(progressDuration))
```
PabstMirror, line 28:
```suggestion
INFO_3("Refuel Module Initialized with flow rate: %1 - hoseLength: %2 - progressDuration: %3",GVAR(rate),GVAR(hoseLength),GVAR(progressDuration))
```

PabstMirror (approved):

addons/refuel/stringtable.xml
```diff
@@ -479,7 +479,11 @@
             <Portuguese>Comprimento da Mangueira de Combustível</Portuguese>
             <French>Longueur du tuyau</French>
             <Czech>Délka hadice na palivo</Czech>
-            <Turkish>Yakıt poması uzunluğu</Turkish>
+            <Turkish>Yakıt hortumu uzunluğu</Turkish>
+        </Key>
+        <Key ID="STR_ACE_Refuel_RefuelSettings_progressDuration_DisplayName">
+            <English>Progress bar duration</English>
```
PabstMirror, line 485: "Pump Interaction Time" might sound a little more natural, but it's fine as is

654wak654: It effects all overall pump, hose, vehicle interaction time so I wanted to make it obvious it' all interactions times. Could maybe call it just "Interaction Time", and then add a hint that explains it better?

kymckay:
Would it make more sense to present this as a coefficient applied to some sane default time instead?

If not, why not make the CBA setting of the time type?

654wak654:
Thanks @kymckay, I haven't touched arma in a while, didn't even know we got a time type! 😅 

Also updated the display name and added a description.

veteran29 (approved):

mharis001 (approved):

PR #8974, opened 2022-07-22
SOG Compat - Add 1.2 update support

veteran29:
- title

### DIFF
--- a/optionals/compat_sog/CfgAmmo/explosives.hpp
+++ b/optionals/compat_sog/CfgAmmo/explosives.hpp
@@ -10,8 +10,8 @@ class vn_mine_m18_x3_ammo: vn_mine_m18_ammo {
     EGVAR(explosives,explosive) = "vn_mine_m18_x3_ammo_scripted";
 };
 
-class APERSBoundingMine_Range_Ammo;
-class vn_mine_tripwire_m16_02_ammo: APERSBoundingMine_Range_Ammo {
+class vn_mine_m16_base;
+class vn_mine_tripwire_m16_02_ammo: vn_mine_m16_base {
     EGVAR(explosives,defuseObjectPosition)[] = {0, -0.01, 0.088};
 };
 
@@ -31,6 +31,7 @@ class vn_mine_m112_remote_ammo: DemoCharge_Remote_Ammo {
 
 // Disable engine damage of punji traps, script damage manually
 // This allows a tighter control of caused wounds and removes ear ringing
+class APERSBoundingMine_Range_Ammo;
 class vn_mine_punji_01_ammo: APERSBoundingMine_Range_Ammo {
     EGVAR(minedetector,detectable) = 0;
 
--- a/optionals/compat_sog/CfgAmmo/explosives.hpp
+++ b/optionals/compat_sog/CfgAmmo/explosives.hpp
@@ -62,3 +62,81 @@ class vn_mine_punji_03_ammo: vn_mine_punji_01_ammo {
         };
     };
 };
+
+class vn_mine_punji_04_ammo: APERSBoundingMine_Range_Ammo {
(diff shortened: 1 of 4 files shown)


optionals/compat_sog/CfgVehicles/explosives.hpp
```diff
@@ -140,3 +140,91 @@ class EXPLOSIVES_PLACE(punji_03): EGVAR(explosives,Place) {
...
+};
+
+class EXPLOSIVES_PLACE(punji_05): EGVAR(explosives,Place) {
+    displayName = "$STR_VN_MINE_PUNJI_05_MAG_DN";
+    model = "vn\weapons_f_vietnam_03\mines\punji_02\vn_mine_punji_05_mag";
+};
+
+// Bike mine
+class EXPLOSIVES_PLACE(bike): EGVAR(explosives,Place) {
+    displayName = $STR_VN_MINE_BIKE_MAG_DN;
```
rautamiekka, line 156: This and so many others have an inconsistency with the above ones: the dollar ref is unquoted here, double-quoted above.

veteran29: True, I've been copy pasting directly from the source and we don't quote stringtables by convention in here 😅

Drofseh:
For the Type 55A I found a source saying "When travelling at 28 knots the operational range is 500 nm", that gives it an endurance of 17.857 hours, you can use that and its in game fuel consumption to estimate the total fuel capacity it should have, although changing the vanilla fuelCapacity value might be required to match it up properly and I don't know if you want to do that.

I found another souce saying it has 1286 KG of diesel, this is about 1495 litres, but will depend on the specific type of diesel used and the tempurature.

optionals/compat_sog/CfgAmmo/explosives.hpp
```diff
@@ -61,3 +62,97 @@ class vn_mine_punji_03_ammo: vn_mine_punji_01_ammo {
...
+class vn_mine_bike_ammo: APERSMine_Range_Ammo {
+    EGVAR(explosives,explosive) = "vn_mine_bike_ammo_scripted";
+};
+
+class vn_mine_cartridge_ammo: APERSMine_Range_Ammo {
+    EGVAR(explosives,explosive) = "vn_mine_cartridge_ammo_scripted";
+
+    // bump range and damage slightly, default values do not work well with ACE Medical
+    indirectHit = 2;
+    indirectHitRange = 0.7;
```
veteran29, line 107: TODO this should be conditional depending on if ACE Medical is loaded, like punji traps changes.

PabstMirror (approved):

optionals/compat_sog/config.cpp
```diff
@@ -26,12 +26,19 @@ class CfgPatches {
...
+            QGVAR(mk18_carry),
             QGVAR(mortar_m29_carry),
             QGVAR(mortar_m2_carry),
             QGVAR(mortar_type53_carry)
         };
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"vn_weapons", "vn_emm_main_menu", "weapons_f_vietnam_c", "ace_trenches"};
+        requiredAddons[] = {
+            "ace_trenches",
+            "loadorder_f_vietnam"
```
PabstMirror, line 40: `Warning Message: Addon 'ace_compat_sog' requires addon 'loadorder_f_vietnam'`
I get this when testing with `!Workshop\@S.O.G. Prairie Fire - Creator DLC Compatibility Data for Non-Owners`

it does have vn_weapons but not loadorder_f_vietnam, so I think we should keep the the old way

veteran29: The loadorder will be there soon and I'm holding up the PR as draft until it's there ;)

PR #8981, opened 2022-07-25
Refuel - Added ability to refuel fuel sources

Quantx:

- Added ability to refuel fuel truck's tanks and jerry cans
- Added a separate flow rate option for filling fuel sources to account for the vastly larger volumes
- Fuel counter now behaves the same for both limited and unlimited fuel sources
- Improved fuel transfer process to ensure that no fuel is lost during the transfer
- Added action to jerry cans to check how much fuel is left
- Prevented fuel source objects from being dragged, carried or loaded into vehicles while they are being refueled or are refueling something

This PR is intended to supplant #7039 by @BaerMitUmlaut


Quantx:
I noticed that fuel trucks with big fuel containers can take ages to fill compared to vehicles with smaller tanks. Should we implement some sort of flow rate multiplier for refilling fuel trucks?

severgun:
From gameplay standpoint multiplayer definitely multiplayer should be implemented.
From IRL it should consider source pump performance, hose diameter and target fitting size. Aircrafts refuel at 1500-2200 L/min

LupusTheCanine:
> From gameplay standpoint multiplier definitely should be implemented. From IRL it should consider source pump performance, hose diameter and target fitting size. Aircrafts refuel at 1500-2200 L/min

It would be nice if each vehicle had max intake rate and sources had max flow rate. It would then pick lower of the two values.

Quantx:
I completely agree with both of the above comments. For now I've implemented a separate flow rate multiplier option for filling cargo fuel sources. It's 10x the standard refueling rate by default which should make filling huge tanks less of a pain. Adding dedicated input/output flow rates for each tank is a great idea and I'll probably implement something like this in t he future.

addons/refuel/CfgVehicles.hpp
```diff
@@ -71,35 +71,35 @@ class CfgVehicles {
                 icon = QPATHTOF(ui\icon_refuel_interact.paa);
                 class GVAR(PickUpNozzle) {
                     displayName = CSTRING(TakeNozzle);
-                    condition = QUOTE([ARR_2(_player,_target)] call FUNC(canTakeNozzle));
-                    statement = QUOTE([ARR_2(_player,_target)] call FUNC(takeNozzle));
+                    condition = QUOTE([ARR_2(_player,_target)] call DFUNC(canTakeNozzle));
+                    statement = QUOTE([ARR_2(_player,_target)] call DFUNC(takeNozzle));
```
BaerMitUmlaut, line 75: DFUNC should only be used when assigning code to a function, like this:
```sqf
DFUNC(foo) = compileFinal "...";
```

BaerMitUmlaut (changes requested):

addons/refuel/CfgVehicles.hpp
Quantx: I've implemented the changes you suggested. Didn't know that was the specific use case for DFUNC, thanks for clearing that up.

Dystopian: > DFUNC should only be used when assigning code to a function, like this:

No, [Coding Guidelines](https://ace3.acemod.org/wiki/development/coding-guidelines.html#21-modulepbo-specific-macro-usage) object. In some situations `FUNC` can become not quotable and using of `DFUNC` is legal in configs.

Drofseh:
@Quantx would you be able to make your variable names a little more verbose? Stuff like `_bb`, `_ll`, and `_rr` makes it a little hard to understand.

Quantx:
@Drofseh I didn't actually write the specific section of code you're referencing. However, I do intend to rewrite that section as part of a different PR after this one gets merged.

Drofseh:
@Quantx Oh my bad, if you're willing to do it that'd be great, but I don't think it has to be in this PR

addons/refuel/functions/fnc_disconnect.sqf
```diff
@@ -1,10 +1,10 @@
 #include "script_component.hpp"
 /*
  * Author: GitHawk
- * Disconnect a fuel nozzle.
+ * Disconnect a fuel nozzle and make unit pick it up.
  *
  * Arguments:
- * 0: Unit <OBJECT>
+ * 0: Unit <OBJECT> (optional)
```
Dystopian, line 7:
```suggestion
 * 0: Unit <OBJECT> (default: objNull)
```

addons/refuel/functions/fnc_getCapacity.sqf
```diff
@@ -0,0 +1,35 @@
+#include "script_component.hpp"
+/*
+ * Author: QuantX
+ * Get the capacity of a fuel source's tank
```
Dystopian, line 4:
```suggestion
 * Gets the capacity of a fuel source's tank.
```

addons/refuel/script_component.hpp
```diff
@@ -3,7 +3,7 @@
 #include "\z\ace\addons\main\script_mod.hpp"
 
 // #define DRAW_HOOKS_POS
-// #define DEBUG_MODE_FULL
+// #define DEBUG_ENABLED_REFUEL
```
Dystopian, line 6:
```suggestion
// #define DEBUG_MODE_FULL
```
this is template in all ACE addons

Dystopian (changes requested):
Also newline at the end of files and space after commands like `getNumber(` -> `getNumber (`

AUTHORS.txt
```diff
@@ -183,3 +183,4 @@ YetheSamartaka
 xrufix
 Zakant <Zakant@gmx.de>
 zGuba
+QuantX
```
BrettMayson, line 186: Place it up above so it is in alphabetical order

BrettMayson (changes requested):

AUTHORS.txt
Quantx: Done. I somehow missed that this was in alphabetical order.

addons/refuel/script_component.hpp
Quantx: I reverted this change.

addons/refuel/functions/fnc_getCapacity.sqf
Quantx: Fixed

addons/refuel/functions/fnc_disconnect.sqf
Quantx: Fixed

addons/refuel/functions/fnc_canConnectNozzle.sqf
```diff
@@ -0,0 +1,30 @@
+#include "script_component.hpp"
+/*
+ * Author: QuantX
+ * Check if an object can have a nozzle connected to it
```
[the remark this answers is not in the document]
Quantx: Fixed

addons/refuel/functions/fnc_disconnect.sqf
```diff
@@ -1,10 +1,10 @@
 #include "script_component.hpp"
 /*
  * Author: GitHawk
- * Disconnect a fuel nozzle.
+ * Disconnect a fuel nozzle and make unit pick it up.
```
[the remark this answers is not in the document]
Quantx: Fixed

addons/refuel/functions/fnc_turnOn.sqf
```diff
@@ -6,6 +6,7 @@
  * Arguments:
  * 0: Unit <OBJECT>
  * 1: Nozzle <OBJECT>
+ * 2: Refuel container (optional) <BOOL>
```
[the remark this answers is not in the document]
Quantx: Fixed

addons/refuel/functions/fnc_canTurnOn.sqf
```diff
@@ -6,6 +6,7 @@
  * Arguments:
  * 0: Unit <OBJECT>
  * 1: Nozzle <OBJECT>
+ * 2: Refuel container (optional) <BOOL>
```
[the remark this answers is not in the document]
Quantx: Fixed. I've gone through and added the space between all instances of `getNumber` and `isNumber` I could find. I'm pretty sure that should cover all of the instances of that specific spacing issue.

Quantx:
@Dystopian I've gone through and added the space between all instances of `getNumber` and `isNumber` I could find. I'm pretty sure that should cover all of the instances of that specific spacing issue.

Dystopian:
> @Dystopian I've gone through and added the space between all instances

At least you missed `isNull(` cases. Should probably use regex to find such instances.

Also there are still no newlines at the end of some files. You can see it in `Files changed` section of this PR (red circles at the end of file).

Quantx:
@Dystopian Fixed the `isNull(` cases and added newlines to every file.

Mirviriam:
I read thru the changes & it looks like both requests were fulfilled - is there something else holding feature back or just that we need to bump?

Quantx:
I have nothing left to add to this feature at this point and I've made all the requested changes. So unless there are any other suggestions or changes anyone wants to make, this should be good to merge.

addons/refuel/functions/fnc_makeJerryCan.sqf
```diff
@@ -22,12 +22,12 @@ if (isNull _target ||
...
     {_target getVariable [QGVAR(jerryCan), false]}) exitWith {};
 
-if (isServer) then {
-    [_target, _fuelAmount] call FUNC(setFuel);  // has global effects
-};
 _target setVariable [QGVAR(jerryCan), true];
 _target setVariable [QGVAR(source), _target];
 
+_target setVariable [QGVAR(capacity), _fuelAmount];
+[_target, _fuelAmount] call FUNC(setFuel);
```
PabstMirror, line 29:
```suggestion
if (isServer) then { [_target, _fuelAmount] call FUNC(setFuel); };
```
`setFuel` does `_source setVariable [QGVAR(currentFuelCargo), (_fuel min _capacity) max 0, true];` 
so I think anytime someone jips it would refill all jerry cans

PabstMirror (approved):
everything else looks good

addons/refuel/functions/fnc_makeJerryCan.sqf
Quantx: Ah, good catch!

BrettMayson (approved):

addons/refuel/functions/fnc_getFuel.sqf
```diff
@@ -19,11 +19,4 @@ params [["_source", objNull, [objNull]]];
...
 
-private _fuel = _source getVariable QGVAR(currentFuelCargo);
-
-if (isNil "_fuel") then {
-    _fuel = getNumber (configOf _source >> QGVAR(fuelCargo));
-    _source setVariable [QGVAR(currentFuelCargo), _fuel, true];
-};
-
-_fuel
+_source getVariable [QGVAR(currentFuelCargo), 0];
```
BaerMitUmlaut, line 22: This change is unnecessary and breaks public API.

BaerMitUmlaut (dismissed):

addons/refuel/functions/fnc_getFuel.sqf
Quantx: Okay, I'll revert this

Issue #8979, opened 2022-07-24
Rocket launcher back-blast incorrectly starts at front of launcher


darknightx33x:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.08 main/stable`
- **CBA:** `v3.15.7.220517 stable` 
- **ACE3:** `3.15.0 stable` 

**Description:**
The back blast cone of shoulder fired rocket launchers originates from the point where the rocket starts when fired. For many launchers, the rocket starts at front/muzzle of the launcher. This results in the back blast originating from the front of the launcher, instead of the back of it. Thus, you can hit people standing close by or shoulder to shoulder with the person firing the rocket with the back blast, and injuring them unintentionally. 

**Steps to reproduce:**
_Follow [https://ace3.acemod.org/img/wiki/user/issue_flowchart.webp](this flowchart)!_

Easiest/clearest example:
1. From Arma 3 menu, Go into ace arsenal 
2. Equip a loaded rocket launcher. i.e. rpg-32 (launchers with larger cones like the rpg-32 with a 60 degrees cone)
3. stand as close as possible next to or slightly in front of one/several test dummies with at least one being on your right side, next to the launcher. Roughly where an assistant loader should be. close, but not behind the back of the launcher.
4. Fire the rocket off into the distance to not harm anyone nearby.
5. Inspect the nearby dummies for wounds.
6. Additionally, note the beginning point of the rocket's path with the automatically generated tracer. (you may need to take several side steps to get it to not appear inside the launcher)

**Expected behavior:**
Back blast originates from the back of the launcher. In the example given above, the dummy standing next to you should not have any wounds on them. 

**Where did the issue occur?**
- Tested on: Dedicated Multiplayer /  Virtual Arsenal
- Expect it to be same for all modes.

**Log Files:**
No crash log generated during testing.

**Additional context:**
I had a short discussion on the discord server before posting this bug
https://discord.com/channels/976165959041679380/976224710385877002/1000815367301632121

**Screenshots:**
Attached image with rough lines drawn to show back blast cone on rpg-32 from projectile start. Note the blood on the ground indicating wounds from the 2 of 3 standing, and 3 of 3 crouched dummies. Several of them, very clearly out of the range of what the back blast should be.
![image](https://user-images.githubusercontent.com/44118889/180665315-aebc7e25-8445-454c-964b-418a1140ba6f.png)

PR #8985, opened 2022-07-26
Overpressure - Fix backblast originating at projectile

Whigital:
- Add backblast offset to shoulder launchers
- Change overpressure cache from missionNamespace vars to a hashMap
- Update docs with new `ace_overpressure_offset` attribute

Fix https://github.com/acemod/ACE3/issues/8979

### DIFF
--- a/addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
+++ b/addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
@@ -47,11 +47,12 @@ TRACE_1("ConfigPath",_config);
 private _return = [
     (getNumber (_config >> QGVAR(angle))),
     (getNumber (_config >> QGVAR(range))) * GVAR(distanceCoefficient),
-    (getNumber (_config >> QGVAR(damage)))
+    (getNumber (_config >> QGVAR(damage))),
+    (getNumber (_config >> QGVAR(offset)))
 ];
 
-private _varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
-missionNameSpace setVariable [_varName, _return];
-TRACE_2("Return",_varName,_return);
+private _key = format ["%1#%2#%3", _weapon, _ammo, _magazine];
+GVAR(cacheHash) set [_key, _return];
+TRACE_2("Return",_key,_return);
 
 _return
--- a/addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
+++ b/addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
@@ -51,7 +51,7 @@ private _return = [
     (getNumber (_config >> QGVAR(offset)))
 ];
 
-private _key = format ["%1#%2#%3", _weapon, _ammo, _magazine];
+private _key = [_weapon, _ammo, _magazine];
 GVAR(cacheHash) set [_key, _return];
 TRACE_2("Return",_key,_return);
 
--- a/addons/overpressure/functions/fnc_firedEHBB.sqf
+++ b/addons/overpressure/functions/fnc_firedEHBB.sqf
@@ -18,20 +18,22 @@
 //IGNORE_PRIVATE_WARNING ["_unit", "_weapon", "_muzzle", "_mode", "_ammo", "_magazine", "_projectile", "_vehicle", "_gunner", "_turret"];
 TRACE_10("firedEH:",_unit, _weapon, _muzzle, _mode, _ammo, _magazine, _projectile, _vehicle, _gunner, _turret);
 
-// Bake variable name and check if the variable exists, call the caching function otherwise
-private _varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
-private _var = if (isNil _varName) then {
-    [_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues);
-} else {
-    missionNameSpace getVariable _varName;
+// Bake key name and check if it exists, call the caching function otherwise
+private _key = (format ["%1#%2#%3", _weapon, _ammo, _magazine]);
+private _opData = (GVAR(cacheHash) getOrDefault [_key, []]);
+
+if (_opData isEqualTo []) then {
+    _opData = [_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues);
 };
-_var params ["_backblastAngle","_backblastRange","_backblastDamage"];
-TRACE_3("cache",_backblastAngle,_backblastRange,_backblastDamage);
+
+_opData params ["_backblastAngle", "_backblastRange", "_backblastDamage", "_offset"];
+TRACE_4("cache",_backblastAngle,_backblastRange,_backblastDamage,_offset);
 
 if (_backblastDamage <= 0) exitWith {};
 
 private _position = getPosASL _projectile;
 private _direction = [0, 0, 0] vectorDiff (vectorDir _projectile);
+_position = (_position vectorAdd (_direction vectorMultiply _offset));
 
 // Damage to others
 private _affected = (ASLtoAGL _position) nearEntities ["CAManBase", _backblastRange];
--- a/addons/overpressure/functions/fnc_firedEHBB.sqf
+++ b/addons/overpressure/functions/fnc_firedEHBB.sqf
@@ -18,11 +18,11 @@
 //IGNORE_PRIVATE_WARNING ["_unit", "_weapon", "_muzzle", "_mode", "_ammo", "_magazine", "_projectile", "_vehicle", "_gunner", "_turret"];
 TRACE_10("firedEH:",_unit, _weapon, _muzzle, _mode, _ammo, _magazine, _projectile, _vehicle, _gunner, _turret);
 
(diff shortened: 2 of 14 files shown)


kymckay (approved):
Nice, even updated the docs :+1:

Haven't verified that all of your updated config values make sense, but I trust you tested those to set them all.

addons/overpressure/functions/fnc_firedEHBB.sqf
```diff
@@ -18,20 +18,22 @@
...
-TRACE_3("cache",_backblastAngle,_backblastRange,_backblastDamage);
+
+_opData params ["_backblastAngle", "_backblastRange", "_backblastDamage", "_offset"];
+TRACE_4("cache",_backblastAngle,_backblastRange,_backblastDamage,_offset);
 
 if (_backblastDamage <= 0) exitWith {};
 
 private _position = getPosASL _projectile;
 private _direction = [0, 0, 0] vectorDiff (vectorDir _projectile);
+_position = (_position vectorAdd (_direction vectorMultiply _offset));
```
rautamiekka, line 36:
```suggestion
private _position = getPosASL _projectile vectorAdd (_direction vectorMultiply _offset);
private _direction = [0, 0, 0] vectorDiff (vectorDir _projectile);
```
AFAICT, in this case the `vectorAdd` statement can just instead be combined with the `getPosATL` statement it modifies.

Dystopian: @rautamiekka looks like you use `_direction` before defining

Whigital: Positioning statement simplified ....

addons/overpressure/functions/fnc_firedEHBB.sqf
```diff
@@ -18,20 +18,19 @@
...
-} else {
-    missionNameSpace getVariable _varName;
-};
-_var params ["_backblastAngle","_backblastRange","_backblastDamage"];
-TRACE_3("cache",_backblastAngle,_backblastRange,_backblastDamage);
+// Bake key name and check if it exists, call the caching function otherwise
+private _opData = (GVAR(cacheHash) getOrDefault [
+    (format ["%1#%2#%3", _weapon, _ammo, _magazine]),
+    ([_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues))
+]);
```
mharis001, line 25: Unnecessary parentheses.

addons/overpressure/functions/fnc_firedEHOP.sqf
```diff
@@ -18,20 +18,17 @@
...
-    [_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues);
-} else {
-    missionNameSpace getVariable _varName;
-};
-_var params ["_dangerZoneAngle","_dangerZoneRange","_dangerZoneDamage"];
+// Bake key name and check if it exists, call the caching function otherwise
+private _opData = (GVAR(cacheHash) getOrDefault [
+    (format ["%1#%2#%3", _weapon, _ammo, _magazine]),
+    ([_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues))
+]);
```
mharis001, line 25: Unnecessary parentheses.

addons/overpressure/functions/fnc_overpressureDamage.sqf
```diff
@@ -23,14 +23,13 @@
...
-    [_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues);
-} else {
-    missionNameSpace getVariable _varName;
-};
-_var params ["_overpressureAngle","_overpressureRange","_overpressureDamage"];
+// Bake key name and check if it exists, call the caching function otherwise
+private _opData = (GVAR(cacheHash) getOrDefault [
+    (format ["%1#%2#%3", _weapon, _ammo, _magazine]),
+    ([_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues))
+]);
```
mharis001, line 30: Unnecessary parentheses.

addons/overpressure/functions/fnc_cacheOverPressureValues.sqf
```diff
@@ -47,11 +47,12 @@ TRACE_1("ConfigPath",_config);
...
     (getNumber (_config >> QGVAR(range))) * GVAR(distanceCoefficient),
-    (getNumber (_config >> QGVAR(damage)))
+    (getNumber (_config >> QGVAR(damage))),
+    (getNumber (_config >> QGVAR(offset)))
 ];
 
-private _varName = format [QGVAR(values%1%2%3), _weapon, _ammo, _magazine];
-missionNameSpace setVariable [_varName, _return];
-TRACE_2("Return",_varName,_return);
+private _key = format ["%1#%2#%3", _weapon, _ammo, _magazine];
```
mharis001, line 54: For hash map don't need to `format`; could use array as key directly.

addons/overpressure/functions/fnc_firedEHBB.sqf
```diff
@@ -18,11 +18,11 @@
...
 
-// Bake key name and check if it exists, call the caching function otherwise
-private _opData = (GVAR(cacheHash) getOrDefault [
-    (format ["%1#%2#%3", _weapon, _ammo, _magazine]),
-    ([_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues))
-]);
+// Check if key exists, call the caching function otherwise
+private _opData = GVAR(cacheHash) getOrDefault [
+    [_weapon, _ammo, _magazine],
+    [_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues)
```
kymckay, line 24: One point of note, I believe this function is always being called regardless of whether the key exists or not (due to the way SQF commands work, this array is fully constructed before `getOrDefault` does anything).

Whigital: Use something like this perhaps, or just move the check into `ace_overpressure_cacheOverPressureValues` and always do that call ?
```sqf
private _key = [_weapon, _ammo, _magazine];
private _opData = (
    if (_key in GVAR(cacheHash)) then {
        GVAR(cacheHash) get _key;
    } else {
        _key call FUNC(cacheOverPressureValues);
    }
);
```

mharis001: To avoid code duplication, probably better to add a function like `getOverPressureValues` that takes care of the caching details as well.

addons/overpressure/functions/fnc_firedEHBB.sqf
rautamiekka: > @rautamiekka looks like you use `_direction` before defining

Oops. Knew I missed something.

addons/overpressure/functions/fnc_firedEHBB.sqf
Whigital: Was just thinking adding this at the beginning of `ace_overpressure_cacheOverPressureValues`:
```sqf
private _key = [_weapon, _ammo, _magazine];
private _return = [];

// Return cached entry if it exists
if (_key in GVAR(cacheHash)) exitWith {
    _return = GVAR(cacheHash) get _key;
    TRACE_2("CacheHit",_key,_return);
    _return
};
```

And then just do this in the other functions:
```sqf
// Call the caching function to retrieve overpressure data
private _opData = [_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues);
```

commy2: Standard implementation of a hash is such:
```sqf
private _value = GVAR(cache) getOrDefault [_key, nil];
if (isNil "_return") then {
    _value = _key call FUNC(expensiveFunction);
    GVAR(cache) set [_key, _value];
};
_value // return
```
A hash only makes sense if expensiveFunction is more expensive than traversing this snippet on average.

Whigital: So, I've added a `getOverPressureValues` function, which is basically just a wrapper for `cacheOverPressureValues`:
```sqf
private _values = GVAR(cache) get _this;

// Values not cached, call the caching function
if (isNil "_values") then {
    _values = _this call FUNC(cacheOverPressureValues);
};

_values
```

`private _opValues = [_weapon, _ammo, _magazine] call FUNC(getOverPressureValues);` then used in the EH functions.

Will this be satisfactory ?

PabstMirror: I think it's short enough you could just keep it inline in the 3 places it's used and not create a new func for it
```
private _opData = GVAR(cacheHash) get [_weapon, _ammo, _magazine];
if (isNil "_opData") then {
    _opData = [_weapon, _ammo, _magazine] call FUNC(cacheOverPressureValues);
};
```

kymckay: Personally I'd just rename the `chacheOverPressureValues` function to `getOverpressureValues` and then put the caching in there

PabstMirror (approved):

kymckay:
Re-reviewed, thank you for such a well-packaged contribution :smiley:

PR #9001, opened 2022-08-15
Medical - Disable blood spurts for fire, burn and drown damage types

veteran29:
- title
- Added `noBlood` config param to the damageTypes configs in ACE_Medical_Injuries
- Docs

### DIFF
--- a/addons/medical_blood/functions/fnc_handleWoundReceived.sqf
+++ b/addons/medical_blood/functions/fnc_handleWoundReceived.sqf
@@ -18,8 +18,11 @@
  * Public: No
  */
 
-params ["_unit", "_allDamages", "_shooter"];
-(_allDamages select 0) params ["_damage", ""];
+params ["_unit", "_allDamages", "_shooter", "_damageType"];
+(_allDamages select 0) params ["_damage"];
+
+// Don't bleed if damage type does not cause bleeding
+if (getNumber (configFile >> "ACE_Medical_Injuries" >> "damageTypes" >> _damageType >> "noBlood") == 1) exitWith {};
 
 // Don't bleed when players only and a non-player unit is wounded
 if (GVAR(enabledFor) == BLOOD_ONLY_PLAYERS && {!isPlayer _unit && {_unit != ACE_player}}) exitWith {};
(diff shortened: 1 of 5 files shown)


veteran29:
FIY @pterolatypus

addons/medical_blood/functions/fnc_handleWoundReceived.sqf
```diff
@@ -18,8 +18,11 @@
  * Public: No
  */
 
-params ["_unit", "_allDamages", "_shooter"];
-(_allDamages select 0) params ["_damage", ""];
+params ["_unit", "_allDamages", "_shooter", "_damageType"];
+(_allDamages select 0) params ["_damage"];
+
+// Don't bleed if damage type does not cause bleeding
+if (GVAR(noBloodDamageTypes) getOrDefault [_damageType, false]) exitWith {};
```
mharis001, line 25: I think use `in` here.

pterolatypus:
ideally I would like to see this be based on wound type rather than damage type, but that would require a bit of restructuring

addons/medical_blood/functions/fnc_handleWoundReceived.sqf
veteran29: Yeah, I forgot it works for HashMaps

veteran29:
> I'd be interested to know if there is actually any benefit to caching the config at prestart rather than just reading it fresh at preinit.

```sqf
// Execution Time: 0.0015 ms  |  Cycles: 10000/10000  |  Total Time: 15 ms
(call (uiNamespace getVariable "ace_medical_blood_noBloodDamageTypes"))
```

```sqf
// Execution Time: 0.0234 ms  |  Cycles: 10000/10000  |  Total Time: 234 ms
"getNumber (_x >> 'noBlood') == 1" configClasses (configFile >> "ACE_Medical_Injuries" >> "damageTypes") apply {[configName _x, nil]}
```

Is it worth it? Not sure but it's how we've been doing it in other components so doing it like this for consistency. Also, the more damage types the slower the config crawl so not doing that on every mission load is in theory beneficial if there's a lot of custom damage types (unlikely).

pterolatypus:
fair enough, that's more than I was expecting

mharis001 (approved):
Code LGTM.

PabstMirror (approved):

PR #9004, opened 2022-08-17
fix documentation

CantBeLucifer:
for some reason this variable is a number not a boolean. should really be fixed in code.


mharis001:
It can be either `BOOL` or `NUMBER`. https://github.com/acemod/ACE3/blob/131b0b9472931e6bc857f2f0b92e21f2656c8c9d/addons/repair/functions/fnc_isRepairVehicle.sqf#L24-L25

CantBeLucifer:
Doesn't work for me 
image

CantBeLucifer:
Issue is actually with ACE_isRepairFacility not ACE_isRepairVehicle

PR #9005, opened 2022-08-17
Medical Treatment - Change stitching action to be per-body-part

pterolatypus:
- Stitch each body part separately instead of all at once
- Gives medics a bit more control

### IMPORTANT


addons/medical_treatment/functions/fnc_getStitchableWounds.sqf
```diff
@@ -29,5 +30,6 @@ private _bleedingBodyParts = GET_OPEN_WOUNDS(_unit) select {
 GET_BANDAGED_WOUNDS(_unit) select {
     _x params ["", "_bodyPartN"];
 
-    !(_bodyPartN in _bleedingBodyParts)
+    (_bodyPart in [-1, _bodyPartN]) &&
+    {!(_bodyPartN in _bleedingBodyParts)}
```
mharis001, line 34:
```suggestion
    _bodyPart in [-1, _bodyPartN] && {!(_bodyPartN in _bleedingBodyParts)}
```

addons/medical_treatment/functions/fnc_canStitch.sqf
```diff
@@ -16,6 +17,7 @@
  * Public: No
  */
 
-params ["", "_patient"];
+params ["", "_patient", "_bodyPart"];
 
-(_patient call FUNC(getStitchableWounds) isNotEqualTo [])
+private _index = ALL_BODY_PARTS find toLower _bodyPart;
+([_patient, _index] call FUNC(getStitchableWounds) isNotEqualTo [])
```
mharis001, line 23:
```suggestion
[_patient, _index] call FUNC(getStitchableWounds) isNotEqualTo []
```

BrettMayson:
Should this be a setting? I can see this creating a lot of confusion, anger, and "bug reports" when this goes live

veteran29:
The progress bar should say which body part is stitched IMO.

pterolatypus:
>Should this be a setting?

I agree it might cause some confusion, but I think people will get used to it quickly enough and it's more consistent with other treatments.
Adding a setting could lead to more confusion in the long run as the interaction will be different between different servers - this is already a problem when trying to teach people how to use medical, because the behaviour can vary massively depending on the settings.

That said, I don't really care too much I just want this for my community so I'll put it in if people really want it.

>The progress bar should say which body part is stitched

I agree that could be good. Currently the progress par text is static for all medical actions, so it'd mean tweaking that and then it could be applied to virtually all treatments. I think it merits a separate PR though.

mharis001:
I'd agree with no setting. This change makes stitching consistent with the other treatments and just seems like more natural interaction behaviour.

kymckay:
Agree with no setting, each kit should probably get 5 uses (one for each body part) if we have a way to do that so that they're "equivalent" to before (and old loadouts are balanced as expected)

veteran29:
If we will make surgical kit into a magazine we can easily track amount of use. Not sure if changing its type will cause any issues at this point tho.

pterolatypus:
To be honest, consuming surgical kits is already weird with the current implementation because if you cancel before it finishes you can stitch all but 1 wound and then get your item back. I think it'd make more sense to have an item like "suturing thread" that consumes one per wound stitched.

BrettMayson:
> To be honest, consuming surgical kits is already weird with the current implementation because if you cancel before it finishes you can stitch all but 1 wound and then get your item back. I think it'd make more sense to have an item like "suturing thread" that consumes one per wound stitched.

I believe this was the solution in the other thread this was brought up

> I'd agree with no setting

Me too, just gotta make sure it's really obvious in the changelog. We'll still have people showing up in the Discord to complain it is broken anyway though

Drofseh:
Making a new item will require everyone to update any custom loadouts they have.

I'd prefer the existing surgical kit became multi use when consumable.

pterolatypus:
We could always convert the existing kits

BrettMayson:
Needs to be updated after #9020

BrettMayson (approved):

PR #8997, opened 2022-08-10
Cargo - Add setting for carry after cargo unload

tbeswick96:
**When merged this pull request will:**
- Add a setting to disable automatically carrying/dragging cargo on unload

### IMPORTANT

- [x] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9013, opened 2022-08-26
Cargo - Add a way to block auto carry after unload on classes

veteran29:
- title
- extension of #8997


markooff:
Thank You very much @veteran29 - this is exactly the solution I've mention about in https://github.com/acemod/ACE3/pull/8997

Thank You :)

PabstMirror (approved):

PR #9016, opened 2022-08-27
Fire - Increase effectiveness of rolling for putting out fire

veteran29:
- title

Reasoning:
In my experience when burning, players do not think about using the medical menu for putting themselves out, usually, they try to use rolling which is highly ineffective. This PR makes them a bit closer to each other in effectiveness, however "Pat down" action is still bit more effective.


jonpas (approved):

Drofseh:
I didn't even realize these had different levels of effectiveness.
But if they are going to, I think rolling ought to be *more* effective than patting down, as rolling is the recommended thing to do in every first aid course I've ever done.

veteran29:
If so I can make them exactly as effective, but the rolling decreases the intensity every 3s while pat down action does it after 5s.

TheCandianVendingMachine:
The issue with rolling vs patting down is that if rolling is more effective, patting down shouldn't be an option. I decided to make patting down more effective for the gameplay dynamic of people yelling at someone on fire to stop rolling so they can put them out faster

veteran29:
Rolling is most intuitive in such a situation, we can consider making the pat down more effective in case it is done by other people. But I believe that current rolling values are way too low.

rautamiekka:
> The issue with rolling vs patting down is that if rolling is more effective, patting down shouldn't be an option because it's just useless. I decided to make patting down more effective for the gameplay dynamic of people yelling at someone on fire to stop rolling so they can put them out faster

The'Pat Down' is more useful when there already are ppl around the victim, assuming everyone's 'Pat Down' has an effect just like multiple ppl patching the same body part is way faster.

TheCandianVendingMachine (approved):

mharis001 (approved):

PR #9019, opened 2022-08-31
Zeus - Add medical menu module

BrettMayson:
- Allows Zeus to open the medical menu on a unit
- Inspired by #6348


addons/medical_treatment/functions/fnc_treatment.sqf
```diff
@@ -54,88 +54,90 @@ private _userAndItem = if (GET_NUMBER_ENTRY(_config >> "consumeItem") == 1) then
 
 _userAndItem params ["_itemUser", "_usedItem"];
 
-// Get treatment animation for the medic
-private _medicAnim = if (_medic isEqualTo _patient) then {
-    getText (_config >> ["animationMedicSelf", "animationMedicSelfProne"] select (stance _medic == "PRONE"));
-} else {
-    getText (_config >> ["animationMedic", "animationMedicProne"] select (stance _medic == "PRONE"));
-};
+if !(_medic isEqualTo player && {!isNull findDisplay 312}) then {
```
rautamiekka, line 57:
```suggestion
if (_medic isNotEqualTo player && {!isNull findDisplay 312}) then {
```
If you meant "the medic ain't the Player and the current view is the Zeus interface", if I understand [`findDisplay`](https://community.bistudio.com/wiki/findDisplay) correctly.

BrettMayson: I want to show the animation unless:
1. the medic is the player
2. the player is in Zeus

Your change would mean that the animation would never be shown for the player. An alternative could be this, which I suppose could be easier to read.
```suggestion
if (_medic isNotEqualTo player || {isNull findDisplay 312}) then {
```

addons/medical_treatment/functions/fnc_treatment.sqf
```diff
@@ -54,88 +54,90 @@ private _userAndItem = if (GET_NUMBER_ENTRY(_config >> "consumeItem") == 1) then
...
 
-// Play a random treatment sound globally if defined
-if (isArray (_config >> "sounds")) then {
-    selectRandom getArray (_config >> "sounds") params ["_file", ["_volume", 1], ["_pitch", 1], ["_distance", 10]];
-    playSound3D [_file, objNull, false, getPosASL _medic, _volume, _pitch, _distance];
+    // Play a random treatment sound globally if defined
+    if (isArray (_config >> "sounds")) then {
+        selectRandom getArray (_config >> "sounds") params ["_file", ["_volume", 1], ["_pitch", 1], ["_distance", 10]];
+        playSound3D [_file, objNull, false, getPosASL _medic, _volume, _pitch, _distance];
+    };
```
rautamiekka, line 140:
```suggestion
    private _soundsConfig = _config >> "sounds";
    if (isArray _soundsConfig) then {
        selectRandom getArray _soundConfigs params ["_file", ["_volume", 1], ["_pitch", 1], ["_distance", 10]];
        playSound3D [_file, objNull, false, getPosASL _medic, _volume, _pitch, _distance];
    };
```

addons/medical_treatment/functions/fnc_treatment.sqf
rautamiekka: Yeah, that's much more readable.

jonpas:
Every action, no consuming.

BrettMayson:
Works fine in SP, very broken in MP. May need someone more familiar with medical to assist

johnb432:
> Works fine in SP, very broken in MP. May need someone more familiar with medical to assist

I took your branch and merged the current master branch into it. There are some merge conflicts, but they are easily solved.

I just tested it in self-hosted MP, with a second PC acting as a second client. I haven't come across any problems whatsoever.
Can you detail what sort of issues you stumbled across when you tested in MP? Granted, I haven't tested extensively, but applying tourniquets, bandaging, PAKing and stitching all worked fine for me.

---

I have a few suggestions:
- Add a Zeus check in `ace_medical_gui_fnc_handleToggle` to avoid toggling targets for the medical menu while in Zeus.
```SQF
// If in Zeus, ignore
if (!isNull findDisplay 312) exitWith {};
```
Adding the above at the start of the file did the trick in my tests.
- This is up for general debate I think, but I would like to see the treatment times be reduced drastically when in Zeus. As a Zeus, I don't want to have to wait 7s to apply a tourniquet. In my tests I forced all medical interactions to last 1s when in Zeus.
Maybe it's worth adding a CBA setting for that? It would allow players tune it to their liking. It could be a multiplier or a set value.

BrettMayson:
Feel free to finish it up, it's probably be a while before it was ok my radar again

johnb432:
Closed in favor of #9367

Issue #9035, opened 2022-09-14
ACE Arsenal - 'SHIFT + IMPORT' feature not importing data from clipboard


SkyeGuy:
**MODS USED:**
- **Arma 3**: Public release build, `version 2.10`
- **CBA:** `3.15.8`
- **ACE3:** `3.15.1`
-> Issue was replicated with only these two mods loaded on a new scratch-built 3DEN editor mission made on 9/14/22.

**Description:**
ACE Arsenal - the "SHIFT + IMPORT" feature is not importing DEFAULT loadout data from clipboard. "SHIFT + EXPORT" seems to be working correctly in exporting the data to clipboard.

**Steps to reproduce:**

1. Load mission in 3DEN editor with playable unit down. Right click -> Edit Loadout -> ACE Arsenal.
2. Click LOADOUTS -> DEFAULT LOADOUTS at bottom.
3. Save kit as loadout with name. Hit CLOSE to return to ACE Arsenal.
4. Open Notepad or other raw text editor program, to paste the data.
5. Hit "LEFT SHIFT + CLICK 'EXPORT'" button to export the saved default loadout to clipboard.
6. CTRL+V on Notepad to paste data.
7. Go back to ACE Arsenal; go into the DEFAULT LOADOUTS screen again and delete the saved loadout to test the import. Hit CLOSE.
8. Go to Notepad .txt file; hit 'CTRL+A' to select all, then 'CTRL+C' to copy data.
9. Back to Arsenal - hit "LEFT SHIFT + CLICK 'IMPORT'" to import the data from clipboard. 
10. Navigate to DEFAULT LOADOUTS screen. **No imported data is present.**

**Expected behavior:**
When importing previously saved & exported default loadout arrays, the SHIFT + IMPORT function should import the data.

**Where did the issue occur?**
- Editor (Singleplayer & Multiplayer), in ACE ARSENAL

**Log Files:**
- No logs applicable for this scenario that I am aware of, unless there are logs that ACE Arsenal functions output somewhere.

**Video:**
There is a video I recorded locally demonstrating the bug, see the Google Drive link below.
https://drive.google.com/file/d/1xtyjRXJnxQTLzhXufTBdj6v4aFwM76so/view?usp=sharing

johnb432:
[...]

Issue #8844, opened 2022-03-12
removeVirtualItems - Remove case sensitivity


johnb432:
**Is your enhancement related to a problem?**
The ACE Arsenal doesn't return case sensitive naming for weapon attachments (which is another thing that could be remedied, but that's not the point of the issue).
If one wants to remove virtual items from the ACE Arsenal, they must give the case sensitive config name of the items to be removed.

E.g

`[_this, ["optic_DMS"]] call ace_arsenal_fnc_removeVirtualItems` does not work, whereas
`[_this, ["optic_dms"]] call ace_arsenal_fnc_removeVirtualItems` does.

This means the user has to look it up in the configs. Frankly it's not a big deal, but it would be nice if that could be avoided.

**Solution you'd like:**
I would like to see it case insensitive. It would be a quality of life improvement.

**Proposed solution :**

https://github.com/acemod/ACE3/blob/f85ed213c59ba6926c230e0ef8ca1ead292b53af/addons/arsenal/functions/fnc_removeVirtualItems.sqf#L55

```SQF
// Make sure all items are in string form
_items = _items select {_x isEqualType "" && {_x != ""}};

// Find the items that are to be removed using non-case sensitive means
private _item = "";

private _fnc_remove = {
    {
        _item = _x;
        _this deleteAt (_this findIf {_x == _item});
    } forEach _items;

    _this;
};

{
    if (_forEachIndex isEqualTo 0) then {
        _cargo set [_forEachIndex, [(_x select 0) call _fnc_remove, (_x select 1) call _fnc_remove, (_x select 2) call _fnc_remove]];
    } else {
        if (_forEachIndex isEqualTo 1) then {
            _cargo set [_forEachIndex, [(_x select 0) call _fnc_remove, (_x select 1) call _fnc_remove, (_x select 2) call _fnc_remove, (_x select 3) call _fnc_remove]];
        } else {
            _cargo set [_cargo find _x, _x call _fnc_remove];
        };
    };
} forEach _cargo;
```
This is what I have come up with as a working solution. However, it's noticeably slower for large arrays:
Testing with about 700 entries in the items parameter made the game freeze for an estimated time of 1-2s.

I'm hoping others can (help) implement this enhancement.

rautamiekka:
Since I can't test I'll have to speculate, but would it be faster, if less RAM-efficient, to return a copy that's been created by adding only the wanted stuff, instead of deleting from the original array ?

commy2:
[...]

PR #9040, opened 2022-09-18
Arsenal - Add/Fix/Improve/Change numerous aspects

johnb432:
My aim was to cleanup the old code and fix a couple of things:

- Everything has been made to work with config case sensitive names.
- Added the ability to add/remove primary and secondary muzzle magazines to weapons where it wasn't previously possible:
    - Handguns: Added secondary muzzle magazines
    - Launchers: Added secondary muzzle magazines
    - Binoculars: Added primary and secondary muzzle magazines
    
    I added this for weapons like RHS' SMAW, which has two muzzles (primary and a spotting rifle) and laser designators, where you can ensure that a magazine has been loaded.
- Added: `ace_arsenal_uniqueBase` can be applied to items from "CfgWeapons", "CfgMagazines" and "CfgVehicles".
  I'm unsure about that change, but I felt it matched the documentation on it better. Maybe the documentation should be expanded/changed instead.
- Added: More caches to hopefully reduce some loading times. Sorting now caches results, reducing repeated sorting times drastically.
- Added: When searching, it will keep the currently selected item highlighted, unless it doesn't match the search results (works for items as well as loadouts).
- Added the ability to add items from "CfgMagazines" into the "Misc. items" or custom tabs (thanks @LinkIsGrim for the help).
- Added "baseWeapon" class support for weapon attachments. If a weapon attachment has the config property "baseWeapon" defined, it will take that item and show that in the arsenal.
- Added stronger filtering on item scopes: If `scopeArsenal` exists, it must be 2 and `scope > 0`, otherwise `scope == 2`.
- Added "descending" (default, as it is now) and "ascending" sort order as a drop down menu.
- Added: Unique backpacks in containers can now be removed with either the "-" or "clear all items" button.
- Added: FUNC(removeSort). You can't remove the default sort type (alphabetically) to avoid problems with the arsenal.
- Addresses #9035 ~~(it should in theory at least)~~.
- Addresses #8844.
- Fixed: Linked items being removed when switching to a new weapon ( #8786 removed that accidentally).
  Now it removes the linked items only if they aren't available to the arsenal.
- Fixed: Faces not being JIP compatible. Previously it would constantly overwrite the same JIP ID.
- Fixed: Weapons with attachments that were in containers losing their attachments when changing containers of the same type.
- Fixed: When sorting by a number, 2 decimal points have been added, so that when you sort by weight it returns the correct order.
- Fixed: Both FUNC(addStat) and FUNC(compileStats) actually take priority into account now. Because of that priority on several stats needed to be tweaked.
- Changed: When switching containers of the same type, it will no longer replenish magazines.
- Changed: If there are only items with no mass (e.g. ACE fortity tool), the "remove all items" button shows.
- Changed: Loadouts can be renamed case-insensitively ("Temp" -> "TemP", previously not possible).
- Changed: Converted the arsenal to partially work with hashmaps instead of arrays (for configItems and virtualItems, currentItems is still an array). Because of this, performance of FUNC(addVirtualItems) and FUNC(removeVirtualItems) has improved immensely.
- Changed: CBA disposable launchers are handled differently now: Within the arsenal, you can change weapon attachments on disposable launchers, but you can't change their magazines (primary or secondary). Item info on the right and the stats show correct information.
- Changed: FUNC(addSort) now checks if the new sorting method already exists and doesn't add it if it does.
- Changed: FUNC(removeStat) ensures that there are no gaps within the stat array. So from now on, if there is an empty spot in the stats page, it's because there is a stat, but the condition for it being shown hasn't been met.
- Changed: Insignia now need `scope == 2` or no scope property to be shown (like vanilla arsenal).
- Probably more, which I have forgotten at this point, as I have been working on this for several months.

~~This is still WIP and I have numerous questions for the ACE team as to what I should do in some cases. I'll open reviews for the areas in question.~~

~~I do have a couple of general questions though:~~
- ~~Do Macros for the arsenal need to be backwards compatible or can they be altered (or even deleted)?~~
- ~~What sort of scopes (private, protected or public) should be allowed for weapons in the arsenal?~~
- ~~Is `GVAR(currentItems)` public? Do mods rely on it?~~
- ~~I'm unsure whether the new functions I have added to `common` should be placed there or in the arsenal. I figured they could be useful in general, so I put them in common. `ace_common_fnc_addWeapon` can be removed if https://github.com/CBATeam/CBA_A3/pull/1557 gets merged.~~

---
I have tested it some, but mostly SP (with some 3DEN, but only a little). However, whatever is MP related I haven't tested at all.

### IMPORTANT


addons/arsenal/XEH_postInit.sqf
```diff
@@ -81,25 +79,46 @@ GVAR(lastSortRight) = "";
...
 ["CBA_loadoutSet", {
     params ["_unit", "_loadout", "_extendedInfo"];
+
+    // Set face
     private _face = _extendedInfo getOrDefault [QGVAR(face), ""];
+
     if (_face != "") then {
-        _unit setFace _face;
+        if (isMultiplayer) then {
+            private _id = [QGVAR(broadcastFace), [_unit, _face], QGVAR(centerFace_) + netId _unit] call CBA_fnc_globalEventJIP;
```
johnb432, line 98: I'm not a huge fan of setting the face every time `"CBA_loadoutSet"` is called, but I don't think there is a way around it.
On a more important note: Is `netId` the right command for the job? What this needs is a constant ID for the same unit (and I mean object, which means after respawning it must change).

addons/arsenal/functions/fnc_addVirtualItems.sqf
```diff
@@ -21,206 +21,203 @@
...
+        // Add onto existing items, in case some items that were already added aren't available by default in the arsenal
+        for "_index" from IDX_VIRT_ITEMS_ALL to IDX_VIRT_MISC_ITEMS do {
             (_cargo select _index) append (_configItems select _index);
             _cargo set [_index, (_cargo select _index) arrayIntersect (_cargo select _index)];
         };
     };
-
 } else {
+    // Make sure all items are in string form, then convert to config case (non-existent items return "")
+    _items = (_items select {_x isEqualType ""}) apply {_x call EFUNC(common,getConfigName)};
```
johnb432, line 57: Do we want to be able to add items that have `scope` < 2? If so, what are the limits? 1? 0? Should `scopeArsenal` be considered here?

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
```diff
@@ -23,51 +23,36 @@ private _category = lbCurSel (_controlsGroup controlsGroupCtrl IDC_ATTRIBUTE_CAT
...
-            // Only add existent magazines and ensure correct classname case
-            private _magazines = getArray (_muzzleConfig >> "magazines") select {isClass (_cfgMagazines >> _x)};
-            _magazines = _magazines apply {configName (_cfgMagazines >> _x)};
-            _itemsToAdd append _magazines;
-
-            {
-                _itemsToAdd append (_magazineGroups get (toLower _x));
-            } forEach getArray (_muzzleConfig >> "magazineWell");
-        } forEach getArray (_weaponConfig >> "muzzles");
+        _itemsToAdd append (compatibleMagazines _x);
```
johnb432, line 39: I understand if the ACE team wants `CBA_fnc_compatibleMagazines` to be used instead (because of ongoing discussions on the CBA GitHub).

addons/arsenal/functions/fnc_attributeImport.sqf
```diff
@@ -15,32 +16,43 @@
...
-            _filteredList pushBackUnique (_x select _index);
-        };
-    } forEach _configItemsFlat;
-} forEach _importList;
+    // Find common entries
+    _intersection = _x arrayIntersect _importList;
+
+    _filteredList append _intersection;
+
+    //////////////////////// Is this necessary/wanted?
```
johnb432, line 49: Would the code below be of any benefit?

addons/arsenal/functions/fnc_buttonCargo.sqf
```diff
@@ -16,77 +16,92 @@
...
-// Update progress bar status, weight info
-private _loadIndicatorBarCtrl = _display displayCtrl IDC_loadIndicatorBar;
-_loadIndicatorBarCtrl progressSetPosition _load;
-
-private _value = {_x == _item} count _items;
-_ctrlList lnbSetText [[_lnbCurSel, 2],str _value];
+// Find out how many items of that type there are and update the number displayed
+_ctrlList lnbSetText [[_lnbCurSel, 2], str ({_x == _item} count _containerItems)];
 
+//////////////////// Return NUMBER, as it did before or return BOOL, as the doc says is should ?
```
johnb432, line 103: As the comments asks: What should be done? As it is now it returns a bool, previously it returned a number. Doc says it returns bool.

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -42,101 +42,111 @@ if (!(ctrlShown _searchbarCtrl) || {ctrlFade _searchbarCtrl > 0}) then {
...
-        } foreach getArray (_weaponConfig >> "muzzles");
-    };
-} foreach [primaryWeapon GVAR(center), handgunWeapon GVAR(center), secondaryWeapon GVAR(center)];
-
-private _itemsToCheck = [];
-private _compatibleMagsPrimaryMuzzle = [];
-private _compatibleMagsSecondaryMuzzle = [];
+    // Get compatible magazines for primary & secondary muzzle (secondary muzzle is not guaranteed to exist)
+    // Assumption: One weapon can have two muzzles maximum
+    _compatibleMagazines pushBack [compatibleMagazines [_x, _muzzles param [0, ""]], compatibleMagazines [_x, _muzzles param [1, ""]]];
```
johnb432, line 88: Can be easily swapped out for `CBA_fnc_compatibleMagazines`

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -42,101 +42,111 @@ if (!(ctrlShown _searchbarCtrl) || {ctrlFade _searchbarCtrl > 0}) then {
...
 switch (GVAR(currentLeftPanel)) do {
-    case IDC_buttonPrimaryWeapon : {
+    // If weapons or binoculars are chosen, get their compatible magazines & items
+    // Primary weapon
+    case IDC_buttonPrimaryWeapon: {
         _compatibleMagsPrimaryMuzzle = _compatibleMagazines select 0 select 0;
         _compatibleMagsSecondaryMuzzle = _compatibleMagazines select 0 select 1;
-        _compatibleItems = (primaryWeapon GVAR(center)) call bis_fnc_compatibleItems;
-        _itemsToCheck = GVAR(currentItems) select 18;
+        _compatibleItems = compatibleItems (_weapons select 0)/* call CBA_fnc_compatibleItems*/;
```
johnb432, line 121: Can be easily swapped out for `CBA_fnc_compatibleItems`.

addons/arsenal/functions/fnc_fillSort.sqf
```diff
@@ -89,11 +93,17 @@ private _lastSort = [GVAR(lastSortLeft), GVAR(lastSortRight)] select _rightSort;
 private _sortIndex = 0;
 
 {
-    if (_x isEqualTo []) exitWith {};
+    if (_x isEqualTo []) then {
```
johnb432, line 96: I'm unsure about this change, but to me it seems rather odd to stop at an empty array (as it was previously).

addons/arsenal/functions/fnc_onArsenalClose.sqf
```diff
@@ -80,12 +80,13 @@ if (!isNull curatorCamera) then {
     curatorCamera setVectorDirAndUp _dirAndUp;
 };
 
-if (isMultiplayer) then {
-    [QGVAR(broadcastFace), [GVAR(center), GVAR(currentFace)], QGVAR(center) + "_face"] call CBA_fnc_globalEventJIP;
-    [QGVAR(center) + "_face", GVAR(center)] call CBA_fnc_removeGlobalEventJIP;
+// Make face and voice selection JIP compatible; 3DEN doesn't need this though
+if (isMultiplayer && {!is3DEN}) then {
+    private _id = [QGVAR(broadcastFace), [GVAR(center), GVAR(currentFace)], QGVAR(centerFace_) + netId GVAR(center)] call CBA_fnc_globalEventJIP;
```
johnb432, line 85: Same as previous comment on JIP for faces.

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
```diff
@@ -18,429 +19,631 @@ params ["_control", "_curSel"];
...
-private _containerDefaultRightPanel = _display displayCtrl IDC_buttonMisc;
-private _selectCorrectPanelContainer = [_containerDefaultRightPanel, _display displayCtrl GVAR(currentRightPanel)] select (!isNil QGVAR(currentRightPanel) && {GVAR(currentRightPanel) in [RIGHT_PANEL_ITEMS_IDCS]});
+// When having chosen a new category, see if the current right panel can be kept open, otherwise take default
+private _selectCorrectPanelWeapon = [_display displayCtrl IDC_buttonOptic, _display displayCtrl GVAR(currentRightPanel)] select (!isNil QGVAR(currentRightPanel) && {GVAR(currentRightPanel) in [RIGHT_PANEL_ACC_IDCS, IDC_buttonCurrentMag, IDC_buttonCurrentMag2]});
+private _selectCorrectPanelContainer = [_display displayCtrl IDC_buttonMisc, _display displayCtrl GVAR(currentRightPanel)] select (!isNil QGVAR(currentRightPanel) && {GVAR(currentRightPanel) in [RIGHT_PANEL_ITEMS_IDCS]});
 
+// Remove all magazines from the previous weapon that aren't compatible with the new weapon
 private _fnc_clearPreviousWepMags = {
-    private _compatibleMagsBaseWeapon = [_baseWeapon, true] call CBA_fnc_compatibleMagazines;
+    private _compatibleMagsBaseWeapon = compatibleMagazines _baseWeapon;
```
johnb432, line 31: Can be reverted if wished.

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
```diff
@@ -18,429 +19,631 @@ params ["_control", "_curSel"];
...
             TOGGLE_RIGHT_PANEL_HIDE
         } else {
-            if ((GVAR(currentItems) select 0) != _item && {_baseWeapon != _item}) then {
-                private _compatibleItems = (_item call bis_fnc_compatibleItems) apply {tolower _x};
-                private _cfgMags = configFile >> "CfgMagazines";
-                private _compatibleMags = ([_item, true] call CBA_fnc_compatibleMagazines) select { getNumber (_cfgMags >> _x >> "scope") == 2 };
+            // Check if a new primary weapon was selected
+            if ((GVAR(currentItems) select IDX_CURR_PRIMARY_WEAPON) != _item && {_baseWeapon != _item}) then {
+                // Get magazines that are compatible with the new weapon
+                private _compatibleMags = compatibleMagazines _item;
```
johnb432, line 69: Same comment as for previous `compatibleMagazines`.

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
```diff
@@ -18,429 +19,631 @@ params ["_control", "_curSel"];
...
+                private _compatibleMagIndex = _compatibleMags findAny (GVAR(virtualItems) select IDX_VIRT_ITEMS_ALL);
 
+                // Remove all magazines from the previous weapon that aren't compatible with the new one
                 call _fnc_clearPreviousWepMags;
 
-                [GVAR(center), _item] call CBA_fnc_addWeaponWithoutItems;
-                if (_compatibleMags isNotEqualTo []) then {
-                    GVAR(center) addWeaponItem [_item, [_compatibleMags select 0], true];
+                // Add new weapon without taking a magazine from the inventory
+                [GVAR(center), _item] call EFUNC(common,addWeapon);
```
johnb432, line 76: Can be replaced if that CBA PR I mentioned is merged.

addons/arsenal/functions/fnc_scanConfig.sqf
```diff
@@ -12,178 +13,223 @@
...
-    [ ], // WeaponThrow 15
-    [ ], // WeaponPut 16
-    [ ] // InventoryItems 17
-];
-
-private _configCfgWeapons = configFile >> "CfgWeapons"; //Save this lookup in variable for perf improvement
+private _configItems = EMPTY_VIRTUAL_ARSENAL;
 
+private _filterFunction = [{
+    isClass _x && {(if (isNumber (_x >> 'scopeArsenal')) then {getNumber (_x >> 'scopeArsenal')} else {getNumber (_x >> 'scope')}) == 2} && {getNumber (_x >> 'ace_arsenal_hide') != 1}
```
johnb432, line 19: Should items with `scope` < 2 and `scopeArsenal` = 2 be allowed?

addons/arsenal/functions/fnc_scanConfig.sqf
```diff
@@ -12,178 +13,223 @@
...
+} forEach configProperties [configFile >> "CfgVehicles", _filterFunction, true];
 
+// Get all facewear
 {
-    (_cargo select 7) pushBackUnique (configName _x);
-} foreach configProperties [(configFile >> "CfgGlasses"), "isClass _x && {(if (isNumber (_x >> 'scopeArsenal')) then {getNumber (_x >> 'scopeArsenal')} else {getNumber (_x >> 'scope')}) == 2} && {getNumber (_x >> 'ace_arsenal_hide') != 1}", true];
+    (_configItems select IDX_VIRT_GOGGLES) pushBackUnique (configName _x);
+} forEach configProperties [configFile >> "CfgGlasses", _filterFunction, true];
 
+// Get magazine groups (magazinewells)
```
johnb432, line 183: The magazine groups are no longer used by the arsenal. Do we still keep the code?

addons/arsenal/functions/fnc_updateRightPanel.sqf
```diff
@@ -1,57 +1,56 @@
...
 
-private _loadIndicatorBarCtrl = _display displayCtrl IDC_loadIndicatorBar;
-private _curSel = lnbCurSelRow _control;
-
-(lnbSize _control) params ["_rows"];
+private _loadRemaining = maxLoad _container - loadAbs _container;
 
-_maxLoad = getnumber (configfile >> "CfgVehicles" >> _maxLoad >> "maximumLoad");
-_maxLoad = _maxLoad * (1 - (progressPosition _loadIndicatorBarCtrl));
-_maxLoad = parseNumber (_maxLoad toFixed 2); // Required to avoid an issue where even though the typename returns "SCALAR" it doesn't act as one.
```
johnb432, line 26: `_display` was never defined anywhere, so that's why he kept getting `SCALAR`.

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -23,220 +23,289 @@ if (count _loadout == 2) then {
...
+                                    // Check if item exists
+                                    if (
+                                        isClass (_cfgWeapons >> _item) ||
+                                        {isClass (_cfgMagazines >> _item)} ||
+                                        {isClass (_cfgGlasses >> _item)} ||
+                                        {isClass (_cfgVehicles >> _item)}
+                                    ) then {
+                                        // Check if item is available in arsenal
+                                        if !(
+                                            ////////////////// use _item in GVAR(virtualItemsFlat) ?
```
johnb432, line 208: Could `GVAR(virtualItemsFlat)` be used instead of all of the individual comparisons? The individual comparisons don't look in the weapons and magazines - which it would for `GVAR(virtualItemsFlat)` - would that be a problem?

addons/arsenal/missions/Arsenal.VR/initPlayerLocal.sqf
```diff
@@ -118,33 +122,45 @@ private _coverObjects = [
...
 [_unit, true, false] call FUNC(initBox);
 
-[{!isNull findDisplay 46}, {
+/*
+[{!isNull findDisplay IDD_MISSION}, {
     [_this, _this, true] call FUNC(openBox);
 }, _unit] call CBA_fnc_waitUntilAndExecute;
+*/
+// Open arsenal when loading screen has finished
+["CBA_loadingScreenDone", {
```
johnb432, line 151: From my tests this works just as well, but should it be changed?

addons/common/XEH_postInit.sqf
```diff
@@ -558,11 +558,8 @@ GVAR(deviceKeyCurrentIndex) = -1;
 
 ["CBA_loadoutSet", {
     params ["_unit", "_loadout"];
-    // remove if with https://github.com/CBATeam/CBA_A3/pull/1548
```
johnb432, line 561: This has been merged, so code is no longer needed.

addons/common/XEH_preStart.sqf
```diff
@@ -1,3 +1,13 @@
 #include "script_component.hpp"
 
 #include "XEH_PREP.hpp"
+
+// These functions are used for the ace arsenal, one of which are called in preStart (ace_arsenal_fnc_scanConfig)
+// Cache for FUNC(getAddon)
+uiNamespace setVariable [QGVAR(addonCache), createHashMap];
```
johnb432, line 7: Is it guaranteed, that `common`'s preStart will run before `arsenal`'s? If not, I will have to add these cache definitions to the functions themselves.

addons/common/functions/fnc_baseWeapon.sqf
```diff
@@ -0,0 +1,62 @@
...
+    _className = configName _configBase;
+    (uiNamespace getVariable QGVAR(baseWeaponNameCache)) set [_cacheKey, _className];
+    _className
+};
+
+// Get first parent without any attachments
+scopeName "main";
+
+// Only take weapons available to the arsenal
+while {isClass _config && {getNumber (_config >> "scope") == 2}} do {
```
johnb432, line 51: What sort of scopes should we allow? Should it be different for the arsenal and general purpose?

addons/arsenal/functions/fnc_buttonImport.sqf
```diff
@@ -15,102 +15,72 @@
...
+    // Supports CBA extended loadout or getUnitLoadout arrays
     {
         if (
-                count _x == 2 &&
-                {_x select 0 isEqualType ""} &&
-                {_x select 0 != ""} &&
-                {_x select 1 isEqualType []} &&
-                {count (_x select 1) == 10}
+            count _x == 2 &&
+            {(_x select 0) isEqualType ""} &&
```
rautamiekka, line 31: There's no need to surround with parenthesis like this, but it does make it more readable.

addons/arsenal/XEH_postInit.sqf
dedmen: netId is specific to the object, so yes should be good.
Alternatively you could use hashValue of the object (but that also uses netId internally)

Important, keep the netId issues in singleplayer in mind, see wiki page

addons/arsenal/functions/fnc_addVirtualItems.sqf
dedmen: scope less than 2 should usually not be displayed.
scopeArsenal is good question, what does vanilla arsenal do?

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
dedmen: If we do essentially the same thing, its better to use CBA to not have duplicate code
https://community.bistudio.com/wiki/compatibleMagazines is also a thing now, don't know what CBA uses

addons/arsenal/functions/fnc_attributeImport.sqf
dedmen: It speeds up the array Intersect on next iteration, so yes

addons/arsenal/functions/fnc_fillSort.sqf
dedmen: Seems correct. continue didn't exist back then

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
dedmen: I looked at your CBA PR.
Assuming your PR is merged,
Con:
This would do the same thing but without cache (is this performance critical?) and filter (which isn't needed) so it would be faster/simpler not to use CBA.
Pro:
If any bug is found in compatibleMagazines command, CBA could temp fix it in their function. But you don't get access to that if you don't use it here.

dedmen: And one important thing on your CBA PR it was said that the engine command can return duplicates? (I should fix that btw)
That can be bad we don't want to display duplicate items in arsenal view.

addons/arsenal/functions/fnc_scanConfig.sqf
dedmen: If it really isn't used/needed anymore. I'm inclined to say no.
Other mods may be piggy-bagging on our code and use our scanned values from our cache.
But these are not marked as public/api so mods shouldn't.

addons/arsenal/functions/fnc_updateRightPanel.sqf
dedmen: But if it was undefined/nil, it wouldn't be able to be passed to typeName 🤔

addons/arsenal/functions/fnc_updateUniqueItemsList.sqf
```diff
@@ -13,106 +14,103 @@
...
-
         // Facewear
-        case (isClass (_configGlasses >> _x)): {
-            (GVAR(virtualItems) select 24) pushBackUnique _x;
+        case (isClass (_cfgGlasses >> _x)): {
+            (GVAR(virtualItems) select IDX_VIRT_UNIQUE_GOGGLES) pushBackUnique _x;
         };
     };
-} foreach _array;
+} forEach (uniformItems GVAR(center) + vestItems GVAR(center) + backpackItems GVAR(center));
```
dedmen, line 116: Imagine being a medic, backpackItems array now has hundreds of bandages and medical supplies.
99% of them being duplicates

You will forEach through an array that likely has 90+% of duplicates where you only need one.

uniqueUnitItems command/cba thing does exist, if really getting items from the center unit is required here.
(Also can't center be a box? a box wouldn't have uniform/vest/backpack?)

addons/arsenal/missions/Arsenal.VR/initPlayerLocal.sqf
```diff
@@ -60,11 +58,11 @@ for "_i" from 0 to 1 do {
     _target setSpeedMode "LIMITED";
 };
 
-//--- Armored vehicles
+// Armored vehicles to the rear of the player
 private _vehicles = [];
 
 private _step = 15;
-_position = [position _unit select 0,(position _unit select 1) + 30,0];
+_position = _unit getRelPos [30, 180];
```
dedmen, line 65: Not the same thing, unit may rotate (though probably at this step it doesn't have the chance to rotate yet?)

addons/arsenal/missions/Arsenal.VR/initPlayerLocal.sqf
```diff
@@ -93,23 +91,29 @@ _position = [position _unit select 0,(position _unit select 1) + 30,0];
...
+                _allDisabled = _allDisabled && ({!_x} count _hitAlive >= 2);
+
+                sleep 0.1;
+            } forEach _this;
 
-    setStatValue ["MarkMassVirtualDestruction", 1];
+            _allDisabled
+        };
+
+        setStatValue ["MarkMassVirtualDestruction", 1];
```
dedmen, line 112: Modded missions shouldn't be able to do this 🤔

addons/arsenal/missions/Arsenal.VR/initPlayerLocal.sqf
dedmen: Seems safer not to.
Or even, the old one fires sooner.
Which means arsenal is opened behind loading screen and player won't notice the short lag when it opens.

addons/common/XEH_preStart.sqf
dedmen: if common is in requiredAddons of arsenal (which it has to be) yes.

addons/arsenal/missions/Arsenal.VR/initPlayerLocal.sqf
johnb432: > Not the same thing, unit may rotate (though probably at this step it doesn't have the chance to rotate yet?)

I know it's not the same thing, but I figured it's cleaner and that the player couldn't have moved when this was executed, as the same structure is used multiple times further above in the code (e.g. `_position = _unit getRelPos [20, 90];`).

addons/arsenal/functions/fnc_fillSort.sqf
johnb432: > Seems correct. continue didn't exist back then

I would have imagined the person who made it back then would have just put the bracket like this:
```SQF
{
    if (count _x != 0) then { // or !(_x isEqualTo [])
        _x params ["_sortName", "_displayName", "", "_condition"];
        // Insert rest here
    };
} forEach _sorts;
```

I just don't understand why it exited instead of skipping the empty element.

addons/arsenal/functions/fnc_updateUniqueItemsList.sqf
johnb432: > Imagine being a medic, backpackItems array now has hundreds of bandages and medical supplies. 99% of them being duplicates
> 
> You will forEach through an array that likely has 90+% of duplicates where you only need one.
> 
> uniqueUnitItems command/cba thing does exist, if really getting items from the center unit is required here. (Also can't center be a box? a box wouldn't have uniform/vest/backpack?)

I must admit, I didn't think much about this.
I just "de-macrod" what was already there (as the macro was no longer needed) - in in turn means that I don't think it's possible for the center to be a box.

I fully agree and will check if a uniqueItems command/function is available for this purpose.

addons/arsenal/functions/fnc_addVirtualItems.sqf
johnb432: `BIS_fnc_addVirtualItemCargo` does not check anything.

In `BIS_fnc_arsenal`, there are two different types of checking:
```SQF
if (isNumber (_x >> "scopeArsenal")) then { getNumber (_x >> "scopeArsenal") } else { getNumber (_x >> "scope") }
```
for weapons, glasses, backpacks and more
and
```SQF
if (getNumber (_cfgMag >> "scope") == 2 || getNumber (_cfgMag >> "scopeArsenal") == 2) then
```
for magazines and weapon attachments.

addons/arsenal/XEH_postInit.sqf
TACHarsis: BIS_fnc_netId doe the same as netID but is fully SP compatible. netID does return nothing/nil in SP.

johnb432:
The latest commit should address some of the issues.

BrettMayson:
@johnb432 Can you handle the merge conflicts, and advise if this is ready for a review or if you have any further unresolved questions.

johnb432:
> @johnb432 Can you handle the merge conflicts, and advise if this is ready for a review or if you have any further unresolved questions.

All all of the questions that I have asked in the still open reviews and at the end of the PR description need replying.

When those have been answered, it will be close to review time.

addons/arsenal/defines.hpp
```diff
@@ -7,6 +7,14 @@
 #define WIDTH_GAP (WIDTH_TOTAL / 100)
 #define WIDTH_SINGLE ((WIDTH_TOTAL - 6 * WIDTH_GAP) / 5)
 
+// IDDs
+#define IDD_MISSION 46
+#define IDD_RSCDISPLAYCURATOR 312
+#define IDD_DISPLAY3DEN	313
```
Dystopian, line 13:
```suggestion
#define IDD_DISPLAY3DEN 313
```
> I can't find the tab that it's complaining about

it's right here :)

Dystopian (changes requested):
> Is GVAR(currentItems) public? Do mods rely on it?

Usually public vars have `ACE_` prefix so I suppose no.

addons/arsenal/functions/fnc_attributeAddItems.sqf
```diff
@@ -68,49 +72,51 @@ if (_category == -1) exitWith {
...
+    case (_category == 11): {_cfgVehicles};
+    case (_category == 12): {_cfgGlasses};
+    default {_cfgWeapons};
 };
 
+private _displayName = "";
+private _picture = "";
+private _symbol = SYMBOL_ITEM_NONE;
+private _alpha = 0;
+private _index = -1;
```
mharis001, line 101: What is the reasoning for this change?

addons/arsenal/defines.hpp
johnb432: I'm disappointed in myself that I hadn't found that - I was sure that I had tried to replace that with a whitespace.
Thanks!

addons/arsenal/functions/fnc_attributeAddItems.sqf
johnb432: As I see it and have tested it in the past, it helps with performance.

Instead of deleting and allocating memory space for a variable each cycle, it reuses the same memory address.
That being said I don't actually know, the previous sentence being an assumption.

addons/arsenal/functions/fnc_verifyLoadout.sqf
LinkIsGrim: Should be fine.

LinkIsGrim:
@johnb432 

> Is GVAR(currentItems) public? Do mods rely on it?

Not documented, so not public. I've used it for workarounds/jank in my stuff IIRC, but responsibility isn't ours at that point.
> Do Macros for the arsenal need to be backwards compatible or can they be altered (or even deleted)?

Need more specifics from you on that. IDCs shouldn't be touched. Search/filter macros can be altered, see above.
> Functions in common

Should be fine unless it's arsenal-specific functionality. I'd say moving to CBA if necessary can always be done later, but that's just me.
> Scopes

I'd say maintain current behavior: scopeArsenal == 2 if available, scope == 2 if not. General purpose should also use this, if intent is for arsenal to use it. baseWeapon filtering should handle any other edge cases like scopeArsenal == 2 and scope < 2. Though again, if it's arsenal-specific functionality, should be in arsenal.
> Event return

Keep the bool, update the doc. Better to maintain BWC at this point.

addons/arsenal/functions/fnc_attributeAddItems.sqf
LinkIsGrim: Checks out from my experience.

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
johnb432: This will be left aside until further noticed.

addons/arsenal/functions/fnc_addVirtualItems.sqf
johnb432: Addressed in [4c9eb8b](https://github.com/acemod/ACE3/pull/9040/commits/4c9eb8b480d2dc32c95f150a8fd040f6cc527af6).

addons/arsenal/functions/fnc_scanConfig.sqf
johnb432: Addressed in [4c9eb8b](https://github.com/acemod/ACE3/pull/9040/commits/4c9eb8b480d2dc32c95f150a8fd040f6cc527af6).

addons/arsenal/functions/fnc_verifyLoadout.sqf
johnb432: Updated to work with `GVAR(virtualItemsFlat)`.

addons/common/functions/fnc_baseWeapon.sqf
johnb432: Addressed in [4c9eb8b](https://github.com/acemod/ACE3/pull/9040/commits/4c9eb8b480d2dc32c95f150a8fd040f6cc527af6).

johnb432:
Firstly, thank you for looking at the PR.

---

> Need more specifics from you on that. IDCs shouldn't be touched. Search/filter macros can be altered/deleted, see above.

I have changed multiple macro definitions in `defines.hpp`, but these changes do not affect IDCs.

> Should be fine unless it's arsenal-specific functionality. I'd say moving to CBA if necessary can always be done later, but that's just me.

At this point it's been put on hold.

> I'd say maintain current behavior: scopeArsenal == 2 if available, scope == 2 if not. General purpose should also use this, if intent is for arsenal to use it. baseWeapon filtering should handle any other edge cases like scopeArsenal == 2 and scope < 2. Though again, if it's arsenal-specific functionality, should be in arsenal.

It should be addressed with the latest commit ([4c9eb8b](https://github.com/acemod/ACE3/pull/9040/commits/4c9eb8b480d2dc32c95f150a8fd040f6cc527af6)).

> Keep the number, update the doc. Better to maintain BWC at this point.

Also addressed in the latest commit.

---

One question remains: Should ACE make use of `compatibleItems` and `compatibleMagazines` instead of `CBA_fnc_compatibleItems` and `CBA_fnc_compatibleMagazines` respectively?

LinkIsGrim:
>One question remains: Should ACE make use of compatibleItems and compatibleMagazines instead of CBA_fnc_compatibleItems and CBA_fnc_compatibleMagazines respectively?

If `compatibleItems` and `compatibleMagazines` don't currently have any bugs or weird behavior we should be fine to use them directly.

addons/arsenal/functions/fnc_scanConfig.sqf
```diff
@@ -12,178 +13,207 @@
...
-];
-
-private _configCfgWeapons = configFile >> "CfgWeapons"; //Save this lookup in variable for perf improvement
+private _configItems = EMPTY_VIRTUAL_ARSENAL;
 
+// https://community.bistudio.com/wiki/Arma_3:_Characters_And_Gear_Encoding_Guide#Character_configuration
+// https://github.com/acemod/ACE3/pull/9040#issuecomment-1597748331
+private _filterFunction = [{
+    isClass _x && {(if (isNumber (_x >> "scopeArsenal")) then {getNumber (_x >> "scopeArsenal")} else {getNumber (_x >> "scope")}) == 2} && {getNumber (_x >> QGVAR(hide)) != 1}
+}] call EFUNC(common,codeToString);
```
LinkIsGrim, line 22:
```suggestion
private _filterFunction = toString {
    isClass _x && {(if (isNumber (_x >> "scopeArsenal")) then {getNumber (_x >> "scopeArsenal")} else {getNumber (_x >> "scope")}) == 2} && {getNumber (_x >> QGVAR(hide)) != 1}
};
```

[`toString`](https://community.bistudio.com/wiki/toString) supports `CODE` since 2.02.

johnb432:
> If `compatibleItems` and `compatibleMagazines` don't currently have any bugs or weird behavior we should be fine to use them directly.

Here's a [thread](https://github.com/CBATeam/CBA_A3/pull/1558) about the differences. In my opinion, the engine commands work just fine and should be sufficient for this use case, but ACE wants to use the CBA functions, then so be it.

mwpowellhtx:
Outside observer looking in, I think it makes perfect sense to leverage the API scaffold provided by CBA. Offers you an opportunity for things to be handled consistently, predictably, etc. Even if the core functionality does route through the scripting commands, it's fine. The benefits are the consistency, etc. Good luck! 🍻

LinkIsGrim:
> >One question remains: Should ACE make use of compatibleItems and compatibleMagazines instead of CBA_fnc_compatibleItems and CBA_fnc_compatibleMagazines respectively?
> 
> If `compatibleItems` and `compatibleMagazines` don't currently have any bugs or weird behavior we should be fine to use them directly.

Doesn't look like they do? At least nothing worse than CBA when doing bad configs on purpose.

LinkIsGrim:
> > If `compatibleItems` and `compatibleMagazines` don't currently have any bugs or weird behavior we should be fine to use them directly.
> 
> Here's a [thread](https://github.com/CBATeam/CBA_A3/pull/1558) about the differences. In my opinion, the engine commands work just fine and should be sufficient for this use case, but ACE wants to use the CBA functions, then so be it.

Use the engine commands directly. If we find a bug testing this related to that we'll go back to CBA.

addons/arsenal/functions/fnc_attributeAddCompatible.sqf
johnb432: Decision has been taken to move forward with engine commands instead of CBA. If bugs are found with engine commands, ACE will revert back to the CBA functions.

addons/arsenal/functions/fnc_fillRightPanel.sqf
johnb432: Decision has been taken to move forward with engine commands instead of CBA. If bugs are found with engine commands, ACE will revert back to the CBA functions.

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
johnb432: Decision has been taken to move forward with engine commands instead of CBA. If bugs are found with engine commands, ACE will revert back to the CBA functions.

addons/arsenal/functions/fnc_onSelChangedLeft.sqf
johnb432: Decision has been taken to move forward with engine commands instead of CBA. If bugs are found with engine commands, ACE will revert back to the CBA functions.

addons/arsenal/functions/fnc_fillRightPanel.sqf
johnb432: Decision has been taken to move forward with engine commands instead of CBA. If bugs are found with engine commands, ACE will revert back to the CBA functions.

johnb432:
I added something yesterday that I forgot to mention: In the ACE Arsenal mission (accessible from the main menu) it is no longer possible to access the arsenal if you are remote controlling a unit. This is to prevent weird behaviour.

---

I have been testing with this branch. There is an issue I noticed: Global Mobilisation's (and some other mods) base classes for binoculars are listed both in the BI and ACE arsenals, when they shouldn't be (`scope = 0`, `scopeArsenal = 2`) if Unsung is loaded. I think this is because Unsung does something weird with the base configs, but that's not really the point. Obviously ACE can't take every poorly made config into account, but perhaps some sort of mitigation might be possible:

Should ACE stick with the current filtering, see:
>I'd say maintain current behavior: scopeArsenal == 2 if available, scope == 2 if not. General purpose should also use this, if intent >is for arsenal to use it. baseWeapon filtering should handle any other edge cases like scopeArsenal == 2 and scope < 2. Though >again, if it's arsenal-specific functionality, should be in arsenal.

Or should ACE filter more strictly - i.e `(scopeArsenal == 2 and scope > 0)` if `scopeArsenal` is available, `scope == 2` if not?

If ACE adheres by [this guideline](https://community.bistudio.com/wiki/Arma_3:_Characters_And_Gear_Encoding_Guide#Character_configuration), then ACE might to consider filtering out anything that has `scope = 0`.

---

This testing has also resulted in a couple of bug fixes, such as dog tags giving errors when starting the arsenal mission, when you had dog tags on yourself when quitting the arsenal mission last. Now, instead of not putting any names down, it defaults to the name "Unknown".

Sorting is now guaranteed to always return the same sorting: Previously it just used the value returned by the sorting function and display name to make a sorting value. However, some items have the display name, making the sorting not always happen in the same order. With the latest commit ([f329fe1](https://github.com/acemod/ACE3/pull/9040/commits/f329fe1aab27f75ba7dbad7a716ebcfd8627b368)) it now takes both the display name and class name into account.

In the future, in another PR, either removing dog tags from your loadout whenever you are in the arsenal mission or simply saving the dog tag data upon quitting, then reloading it when you play the arsenal mission, are some more elegant solutions. But I don't think mixing multiple major component cleanups is a good idea, so it will be for another time.

LinkIsGrim:
We shouldn't be handling classes with scope = 0 at all, add the the check for scope > 0.

mwpowellhtx:
Last time I checked, IIRC:
```
2 = public
1 = protected (not public)
0 = private
```
But don't quote me on that. For sure `2` is key.

johnb432:
[6d8549d](https://github.com/acemod/ACE3/pull/9040/commits/6d8549d4b2292ca41f1efd4cf934a3c4b9d69c0a) changelog:
- Added the ability to add items from "CfgMagazines" into the "Misc. items" or custom tabs.
- Added "baseWeapon" class support for weapon attachments. If a weapon attachment has the config property "baseWeapon" defined, it will take that item and show that in the arsenal.
- Added stronger filtering on item scopes (scope > 0 at least for every item).
- Added "descending" (default, as it is now) and "ascending" sort order as a drop down menu.
- Unique backpacks in containers can now be removed with either the "-" or "clear all items" button.
- When sorting by a number, 2 decimal points have been added, so that when you sort by weight it returns the correct order.
---

I have been looking at `FUNC(addVirtualItems)`. Is [this type of filtering](https://github.com/johnb432/ACE3/blob/6d8549d4b2292ca41f1efd4cf934a3c4b9d69c0a/addons/arsenal/functions/fnc_addVirtualItems.sqf#L84C9-L84C9) necessary or even wanted anymore?
`GVAR(configItems)` should contain all possible valid entries - so should all items be checked through their configs (as it is now) or can I just use `GVAR(configItems)` to place items in the correct categories? The second would probably be faster and I imagine should result in the same, especially with the stricter scope filtering that I have just added.

addons/arsenal/functions/fnc_fillSort.sqf
johnb432: The updated code doesn't seem to pose a problem, so going with it for now.

mwpowellhtx:
Would you mind clarifying:

> Added stronger filtering on item scopes (scope > 0 at least for every item).

It is valid to include [_scope_](https://community.bistudio.com/wiki/CfgVehicles_Config_Reference#scope) `1` (_protected_) scope? I don't think so. `scope=2` should be what is usable to players. `1` is for internal class consistency, AFAIK. If there is some other interpretation of that, would be happy to hear it, but this is my understanding.

Re: `FUNC(addVirtualItems)`, that is essential, I think, aligning the virtual items buckets for use with an arsenal UI style layout, either the thing itself, or inspired by. I mean, unless you don't want the easy-to-use buckets? Could you explain why that is? What is your contingent suggestion in that case?

I am not familiar with `GVAR(configItems)`, per se, but just 'having' all the "valid" items, notwithstanding _public_ `scope` questions, does not necessarily achieve that, does it? So you are either aligning with the UI groupings there or factoring that elsewhere, and the work is already done 'there', so... Again, what is your contingenty recommendations where that is concerned?

Thanks for the efforts...

johnb432:
> Would you mind clarifying:
> 
> > Added stronger filtering on item scopes (scope > 0 at least for every item).
> 
> It is valid to include [_scope_](https://community.bistudio.com/wiki/CfgVehicles_Config_Reference#scope) `1` (_protected_) scope? I don't think so. `scope=2` should be what is usable to players. `1` is for internal class consistency, AFAIK. If there is some other interpretation of that, would be happy to hear it, but this is my understanding.

My changes are [here](https://github.com/acemod/ACE3/blob/6d8549d4b2292ca41f1efd4cf934a3c4b9d69c0a/addons/arsenal/functions/fnc_scanConfig.sqf#L21). If an item has `scopeArsenal` defined, them check if it's `2` and if `scope > 0`. If not, then check if `scope == 2`.

Generally speaking, `scope = 1` is used if you want to remove access from players to an item, but you might want to either spawn it in via a script or in units via the config. Take `arifle_ARX_ghex_ACO_Pointer_Snds_F` or even `arifle_AKM_FL_F` as examples.

My question is: Why is it bad to have items with `scope = 1` && `scopeArsenal = 2` available to players?

My concern is if you force `scope = 2` && `scopeArsenal = 2` it might break compatibility with other mods. Items in the arsenal with `scope = 0` is bad because it throws errors, so those ought to be filtered out, but while items with `scope = 1` && `scopeArsenal = 2` aren't necessarily good for the arsenal, they aren't bad for it either.

---
> Re: `FUNC(addVirtualItems)`, that is essential, I think, aligning the virtual items buckets for use with an arsenal UI style layout, either the thing itself, or inspired by. I mean, unless you don't want the easy-to-use buckets? Could you explain why that is? What is your contingent suggestion in that case?
> 
> I am not familiar with `GVAR(configItems)`, per se, but just 'having' all the "valid" items, notwithstanding _public_ `scope` questions, does not necessarily achieve that, does it? So you are either aligning with the UI groupings there or factoring that elsewhere, and the work is already done 'there', so... Again, what is your contingenty recommendations where that is concerned?
> 
> Thanks for the efforts...

I did not understand any of that. Let me rephrase my first text in case I wasn't clear:

I would like to remove [this type of filtering](https://github.com/johnb432/ACE3/blob/6d8549d4b2292ca41f1efd4cf934a3c4b9d69c0a/addons/arsenal/functions/fnc_addVirtualItems.sqf#L84C9-L84C9), as it's slow because it has to look up everything in the config.

What I would like to implement is the checking of items via `GVAR(configItems)` which contains all valid classes for the arsenal. It would be much quicker.

In my opinion, if an item isn't detected by `FUNC(scanConfig)` (which defines the content of `GVAR(configItems)`), then it's invalid for the arsenal and should not be able to be added via `FUNC(addVirtualItems)`. So to me it should be possible to use `GVAR(configItems)` to do the check, but I might have missed something.

LinkIsGrim:
>In my opinion, if an item isn't detected by FUNC(scanConfig) (which defines the content of GVAR(configItems)), then it's invalid for the arsenal and should not be able to be added via FUNC(addVirtualItems). So to me it should be possible to use GVAR(configItems) to do the check, but I might have missed something.

So long as updateUniqueUnitItems can handle the edge cases (like ACE Intel Items), this is fine. If someone wants to force an invalid virtual item into the arsenal, they can do so at their own risk overwriting the cache or `GVAR(virtualItems)` anyway.

mwpowellhtx:
Re: _scope_ did you read the docs? From the docs, it makes no sense to include _vegetable_ (`1`, `protected`), but it makes perfect sense to include the _Onions_, _purple_ or _green_, `2`, `public`. That's not me criticizing you, that is per the BIS guidance.

mwpowellhtx:
So to clarify, the _virtual item buckets_ are still imporant. It is simply the various `isClass` checks? i.e. with a logical reduction, ad infinitum, as long as the _arrayIntersect_ (or ad hoc equivalent thereof) confirms the class, should be valid?

johnb432:
[8a4a08f](https://github.com/acemod/ACE3/pull/9040/commits/8a4a08f30e1e04ed22e3408af13965e7322f5bc8) introduces the following:

- Converted the arsenal to partially work with hashmaps instead of arrays (for configItems and virtualItems, currentItems is still an array).
- Because of the above, performance of `FUNC(addVirtualItems)` and `FUNC(removeVirtualItems)` has improved immensely.
- Sorting now caches results, reducing repeated sorting times drastically.
- CBA disposable launchers are handled differently now: Within the arsenal, you can change weapon attachments on disposable launchers, but you can't change their magazines (primary or secondary). Item info on the right and the stats show correct information.
- `FUNC(addSort)` now checks if the new sorting method already exists and doesn't add it if it does.
- `FUNC(removeSort)` now exists. You can't remove the default sort type (alphabetically) to avoid problems with the arsenal.
- Both `FUNC(addStat)` and `FUNC(compileStats)` actually taken priority into account now. Because of that priority on several stats needed to be tweaked.
- `FUNC(removeStat)` ensures that there are no gaps within the stat array (so if there is an empty spot in the stats page, it's because there is a stat, but the condition for it being shown hasn't been met).
- Insignia now need `scope == 2` or no `scope` property to be shown (like vanilla arsenal).
---

Sorry that the scope of this PR keeps growing, but I really felt this is worth it.

LinkIsGrim:
TODO for testing:

- [ ] Compatible items.
- [ ] Compatible magazines.
- [ ] Switching between worn items.
- [ ] Switching between weapons with incompatible magazines.
- [ ] Switching between weapons with compatible magazines.
- [ ] Switching between weapons with compatible magazines only on the secondary muzzle.
- [ ] Saving and loading loadouts.
- [ ] Loading loadouts with unavailable/invalid items.
- [ ] Default Loadouts.
- [ ] Public Loadouts.
- [ ] CBA Extended Loadout info.
- [ ] Custom categories.
- [ ] Magazines in Misc Items.
- [ ] Magazines in Custom categories.
- [ ] `baseWeapon` filtering.
- [ ] `baseWeapon` property working for attachments.
- [ ] Unique `assignedItems` not being removed.
- [ ] Sorting and stats still work.
- [ ] Face/Insignia/Voice JIP compatibility.
- [ ] Weapons/Containers within containers when switching containers/loading loadouts.

LinkIsGrim:
Items not present in configItems should be able to be forcefully added via `FUNC(addVirtualItems)`, so long as they have `scope` >= 1. `scopeArsenal` and `baseWeapon` can be disregarded for items added this way.

Items with no `baseWeapon` (or themselves as `baseWeapon`) and outside of `configItems` are not shown in the cargo list. They should appear as unique items.

Performance-wise, this is better, but there's a new delay when switching right panel tabs for the first time. I assume this is due to the recent sorting changes. It's not major, but noticeable. Finding another way to handle that would be nice.

addons/arsenal/functions/fnc_addVirtualItems.sqf
LinkIsGrim, line 1: This should probably use `baseWeapon` property when adding the items as well, since only those get added to `configItems`.

addons/arsenal/functions/fnc_removeVirtualItems.sqf
LinkIsGrim, line 1: Same as `FUNC(addVirtualItems)`.

LinkIsGrim (changes requested):
Looking pretty good.

docs/wiki/framework/arsenal-framework.md
```diff
@@ -134,11 +134,13 @@ Examples:
...
 
-- `ace_arsenal_hide`: `0`(shown) or `1` (hidden), used to hide items from ACE Arsenal or `-1` (forced show), for vehicle magazines.
-- `ace_arsenal_uniqueBase`: Class name of the item that ACE Arsenal will replace it with when saving a loadout.
+- `ace_arsenal_hide`: `0`(shown) or `1` (hidden), used to hide items from ACE Arsenal or `-1` (forced show), for magazines.
+- `ace_arsenal_uniqueBase`: Class name of the item that ACE Arsenal will replace it with when saving a loadout and displaying it in the arsenal. Supports configs in "CfgWeapons" (e.g. Maps, Compasses, Watches, GPS / UAV Terminals, Radios, NVGs, Uniforms, Vests - however no Primary, Secondary, Handgun, Launcher weapons or weapon attachments are supported), "CfgMagazines" (any magazine) and "CfgVehicles" (e.g. Backpacks). Item that replaces must be of the same config type as the original item.
 
 Both of them are optional.
 
+- `baseWeapon`: Class name that is used to display an item in the arsenal. This property can be applied to any weapon or weapon attachment in "CfgWeapons".
+
```
LinkIsGrim, line 143:
```suggestion
- `baseWeapon`: Class name that is used to display an item in the arsenal. This property can be applied to any weapon or weapon attachment in `CfgWeapons`.
- `ACE_isUnique`: Items in `CfgMagazines` with this property set to `1` will be treated and shown by the Arsenal as Misc. Items. Used for items with attached data that needs to be kept track of, such as Notepads or Spare Barrels.

```

Or some other, better description. Either way, this needs to be documented.

addons/arsenal/functions/fnc_addVirtualItems.sqf
johnb432: > This should probably use `baseWeapon` property when adding the items as well, since only those get added to `configItems`.

It already does that [here](https://github.com/johnb432/ACE3/blob/dbdba9ae8f60ae8f418c070dc83325fbbb8190d3/addons/arsenal/functions/fnc_addVirtualItems.sqf#L89). Do you mean other items too? There shouldn't be any other types of items with `baseWeapon` defined/looked at.

mwpowellhtx: My two cents, the only thing is, class names are case insensitive, IIRC, i.e. `_x == _y` can return `true` for `"MyClAsSnAme"` as compared to `"myclassname"`, just an off the cuff example... however, `in` may not necessarily be, also IIRC. May be better off with a `GVAR(configItemsFlat) findIf { _x == _classname; }`.

LinkIsGrim: > My two cents, the only thing is, class names are case insensitive, IIRC, i.e. `_x == _y` can return `true` for `"MyClAsSnAme"` as compared to `"myclassname"`, just an off the cuff example... however, `in` may not necessarily be, also IIRC. May be better off with a `GVAR(configItemsFlat) findIf { _x == _classname; }`.

Items already get changed to their configCase names [here](https://github.com/johnb432/ACE3/blob/dbdba9ae8f60ae8f418c070dc83325fbbb8190d3/addons/arsenal/functions/fnc_addVirtualItems.sqf#L70)

johnb432:
I didn't really want to push this (hopefully) last commit, as a load of testing had already been done, but I felt it was worth it:
- Usage of `keys` when iterating over a hashmap with `forEach`, as it's much faster that way.
- Added partial support for weapons such as `"vn_m79_p"` (see [ticket](https://feedback.bistudio.com/T173880)). Currently you can add and remove magazines from primary and secondary muzzles. However, when saving a loadout, it will not take the secondary muzzle mag into account. I'm hoping `getUnitLoadout` will be fixed soonTM, otherwise I will make another PR fixing it later.
- Better support for unique items: Previously, `configItems` had e.g. magazines with `scope = 1`. As that has been changed, prior to this latest commit (but in this PR), magazines that were compatible with your weapons but had `scope = 1` were shown in the "Misc. items" tab. This commit fixes that sort of behaviour.

addons/arsenal/functions/fnc_fillRightPanel.sqf
```diff
@@ -207,146 +199,149 @@ switch (_ctrlIDC) do {
...
+                        ["CfgVehicles", _x, false, true] call _fnc_fillRightContainer;
+                    };
+                    // Unique goggles
+                    case (_x in (GVAR(virtualItems) get IDX_VIRT_MISC_ITEMS)): {
+                        ["CfgGlasses", _x, false, true] call _fnc_fillRightContainer;
+                    };
+                    // Unknown items
+                    case (_x in (GVAR(virtualItems) get IDX_VIRT_UNIQUE_UNKNOWN_ITEMS)): {
+                        ["CfgWeapons", _x, true, true, true] call _fnc_fillRightContainer;
+                    };
```
LinkIsGrim, line 342: Either I'm missing something or this didn't come out as expected.

mwpowellhtx: > Either I'm missing something or this didn't come out as expected.

?

rautamiekka: Yeah, the `case` statements are just direct copypastes, which won't do what the comments say.

johnb432: Yea, I screwed up when I copy-pasted. I will address this tonight.

LinkIsGrim (approved):
I haven't run this many tests since I was a QA intern.

addons/common/functions/fnc_uniqueUnitItems.sqf
```diff
@@ -0,0 +1,47 @@
...
+ * Returns list of items (including magazines, backpacks and other) in a unit's inventory.
+ * Number definition: 0: Do not return value, 1: Return container only, 2: Return container and content, 3: Return content only
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ * 1: Weapon items <NUMBER> (default: 2)
+ * 2: Uniform items <NUMBER> (default: 2)
+ * 3: Vest items <NUMBER> (default: 2)
+ * 4: Backpack items <NUMBER> (default: 2)
+ * 2: Assigned items <BOOL> (default: true)
```
PabstMirror, line 13:
```suggestion
 * 5: Assigned items <BOOL> (default: true)
```

addons/arsenal/functions/fnc_updateUniqueItemsList.sqf
```diff
@@ -13,106 +14,166 @@
...
+            };
         };
+        // Weapons
+        case (_isWeapon): {
+            _config = _cfgWeapons >> _x;
+            _simulationType = getText (_config >> "simulation");
+            _configItemInfo = _config >> "ItemInfo";
+            _hasItemInfo = isClass (_configItemInfo);
+            _itemInfoType = if (_hasItemInfo) then {getNumber (_configItemInfo >> "type")} else {0};
+            _isMiscItem = _className isKindOf ["CBA_MiscItem", _cfgWeapons];
```
PabstMirror, line 99: `_className` is undefined, probably just `_x` but not sure
```suggestion
            _isMiscItem = _x isKindOf ["CBA_MiscItem", _cfgWeapons];
```

addons/arsenal/functions/fnc_fillLoadoutsList.sqf
```diff
@@ -31,98 +29,106 @@ _contentPanelCtrl ctrlCommit FADE_DELAY;
...
 
             _contentPanelCtrl lnbSetColumnsPos [0, 0.15, 0.40, 0.50, 0.60, 0.70, 0.75, 0.80, 0.85, 0.90];
-            private _newRow = _contentPanelCtrl lnbAddRow [_playerName, _loadoutName];
+            _newRow = _contentPanelCtrl lnbAddRow [_playerName, _loadoutName];
 
-            _extendedLoadout params ["_loadout"];
             ADD_LOADOUTS_LIST_PICTURES
 
-            _contentPanelCtrl lnbSetData [[_newRow, 1], _playerName + _loadoutName];
+            _contentPanelCtrl lnbSetData [[_newRow, 1], _loadoutVar];
```
PabstMirror, line 117: `_loadoutVar` undefined in else block
maybe move line 102 above the if?

addons/arsenal/functions/fnc_updateUniqueItemsList.sqf
LinkIsGrim: `_x` is classname. That works.

addons/arsenal/functions/fnc_fillLoadoutsList.sqf
```diff
@@ -31,98 +29,106 @@ _contentPanelCtrl ctrlCommit FADE_DELAY;
...
             _contentPanelCtrl lnbSetPicture [[_newRow, 0], QPATHTOF(data\iconPublic.paa)];
             _contentPanelCtrl lnbSetValue [[_newRow, 0], 1];
         };
-    } foreach ([_data, +GVAR(defaultLoadoutsList)] select (ctrlIDC _control == IDC_buttonDefaultLoadouts));
+    } forEach ([profileNamespace getVariable [QGVAR(saved_loadouts), []], GVAR(defaultLoadoutsList)] select (ctrlIDC _control == IDC_buttonDefaultLoadouts));
 } else {
+    private _allPlayerNames = allPlayers apply {name _x};
 
     {
         _x params ["_playerName", "_loadoutName", "_loadoutData"];
```
LinkIsGrim, line 99:
```suggestion
        _x params ["_playerName", "_loadoutName", "_loadoutData"];
        private _loadoutVar = _playerName + _loadoutName;
```

addons/arsenal/functions/fnc_fillLoadoutsList.sqf
```diff
@@ -31,98 +29,106 @@ _contentPanelCtrl ctrlCommit FADE_DELAY;
...
 
-        if ((allPlayers apply {name _x}) find _playerName == -1) then {
-
+        // If player who shared loadout doesn't exist anymore, unshare loadout
+        if !(_playerName in _allPlayerNames) then {
             private _loadoutVar = _playerName + _loadoutName;
```
LinkIsGrim, line 102:
```suggestion
```

PabstMirror (approved):

mwpowellhtx:
Congrats @johnb432, pending merge and such. 👍

LinkIsGrim:
Scary.

mwpowellhtx:
Just my two cents, FWIW... doing something similar in my _inventory manager_ effort, presenting _addon mod logos_ and such, along similar lines as the pending [_ACE COMMON_ `'getAddon'`](https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getAddon.sqf) cache. However, I am noticing that the cohesion with the the prospect of a _cached addon_ (really, _mod_), is not great, with usage also appearing during various list filling algorithms, i.e. [`fnc_scanConfig`](https://github.com/acemod/ACE3/blob/master/addons/arsenal/functions/fnc_scanConfig.sqf#L225).

Anyway, the approach I'm taking is to cache the mod config itself, as opposed for every single virtual item class config, and either to have that config available, or a `HASHMAP` bundle of desired `modParams`. That's the first observation, potential optimization.

Second, rather than `"logo"`, not sure whether `"logoSmall"` would be more appropriate, just based upon [`modParams`](https://community.bistudio.com/wiki/modParams#Syntax) documentation. `TL;DR`, have not done the leg work to evaluate whether that is actually desirable, or whether `"logo"` itself is in fact better coverage there. Would be interested to learn what you discover.

And just a third ancillary incidental observation, is `uiNamespace` intentional? This caches variables in the UI namespace, which if I understand correctly, persists between 'mission' engagements, over the lifetime of the A3 session? Is that accurate? May or may not be desirable, I don't know. But in any event...

Good luck, good hunting!

PR #9052, opened 2022-10-01
Dragging - Improve weapon reselect behaviour

mharis001:
- Improve weapon reselect behaviour when dropping carried objects
- Previously, this would always reselect the primary weapon but now reselects whatever weapon (or none) was held when carrying was started

### DIFF
--- a/addons/dragging/functions/fnc_dropObject_carry.sqf
+++ b/addons/dragging/functions/fnc_dropObject_carry.sqf
@@ -48,11 +48,11 @@ if (_target isKindOf "CAManBase" || {animationState _unit in CARRY_ANIMATIONS}) 
 _unit removeWeapon "ACE_FakePrimaryWeapon";
 
 // reselect weapon and re-enable sprint
-private _previousWeapon = _unit getVariable [QGVAR(previousWeapon), primaryWeapon _unit];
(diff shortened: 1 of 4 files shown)


jonpas (commented):
Why public `setVariable`?

mharis001:
Public variable isn't specifically required but should allow for handling disconnecting players better.

addons/dragging/functions/fnc_dropObject_carry.sqf
```diff
@@ -48,7 +48,12 @@ if (_target isKindOf "CAManBase" || {animationState _unit in CARRY_ANIMATIONS})
 _unit removeWeapon "ACE_FakePrimaryWeapon";
 
 // reselect weapon and re-enable sprint
-_unit selectWeapon primaryWeapon _unit;
+private _previousWeapon = _unit getVariable [QGVAR(previousWeapon), primaryWeapon _unit];
+_unit setVariable [QGVAR(previousWeapon), nil, true];
+
+if (_previousWeapon != "") then {
+    _unit selectWeapon _previousWeapon;
+};
```
commy2, line 56: Does this preserve fire mode?

PabstMirror: handles firemode now

PR #9057, opened 2022-10-03
Weather - Add variable to temporarily pause wind simulation

PulsarNeutronStar:
This would enable mission maker to temporarily disable the execution of setWind every second. Currently, a temporary pause isnt possible.

adds a check on GVAR(windSimulation) to allow mission makers to (temporarily) disable Wind Simulation mid mission

### IMPORTANT


BaerMitUmlaut (changes requested):
Modifying a CBA setting at runtime through a script is bad practice and will break with JIPs.

PulsarNeutronStar:
> Modifying a CBA setting at runtime through a script is bad practice and will break with JIPs.

if i see it right, this specific script runs only on the server and will be spawned by an PFeh, but its handle isnt saved anywhere.

Is it possible to see a "list" of all currently running perframe-eh-handlers and would there be a way to "find" the right one and removing it temporarily instead?

johnb432:
> if i see it right, this specific script runs only on the server and will be spawned by an PFeh, but its handle isnt saved anywhere.
> 
> Is it possible to see a "list" of all currently running perframe-eh-handlers and would there be a way to "find" the right one and removing it temporarily instead?

We are not going to touch the PFH execution array.

However, what can be done more easily (but you'd still need to test it) is to set a "normal" global variable that is publicly available. `ace_weather_disableWindSimulation` would be my recommended variable name.

addons/weather/functions/fnc_updateWind.sqf
```diff
@@ -15,6 +15,8 @@
  * Public: No
  */
 
+if (!GVAR(windSimulation)) exitWith {};
```
johnb432, line 18:
```suggestion
// Public API to temporarily disable wind simulation
if (missionNamespace getVariable [QGVAR(disableWindSimulation), false]) exitWith {};
```

docs/wiki/framework/weather-framework.md
```diff
@@ -19,3 +19,10 @@ ACE3 Weather extends the existing weather by temperature, humidity and air press
 The additional wind simulation, which is also influenced by the season and the geographical location, can be deactivated if necessary.
 
 Cloud cover, rain and fog can still be set via the mission settings.
+
+
+## 2. Wind Simulation
+
+## 2.1 Temporarily Pause Wind Simulation
+
+When Wind Simulation is enabled at mission start, it can be temporarily disabled by setting `ACE_Weather_windSimulation = false`.
```
johnb432, line 28:
```suggestion
When Wind Simulation is enabled at mission start, it can be temporarily disabled by setting `ace_weather_disableWindSimulation = true`. To reenable wind simulation, the variable must either be set to `false` or `nil`.
```

LinkIsGrim (approved):

johnb432 (approved):

BaerMitUmlaut (approved):

PulsarNeutronStar:
only now i saw that this has been addressed and merged, thanks alot! :)

PR #9060, opened 2022-10-05
Cookoff - Move canHaveFireJet logic into vehicle_damage

JonBons:
**When merged this pull request will:**
- _Moves canHaveFireJet logic into vehicle_damage to allow for external calls to cookoff to modify this param bypassing the config check_

PR #9061, opened 2022-10-05
Cookoff - Delay full vehicle destruction

JonBons:

- Delay final explosion.

<details>
<summary>Old description</summary>

- _Restricted cookoff jet/ring flame effects to a single execution_
- _After the cookoff jet/ring flame effects have finished a finale can happen (based on random roll) that results in catastrophic destruction of the vehicle (ex. https://youtu.be/qfldUOrXCGk)_
  - _Adds `Destruction Finale Chance` CBA setting to adjust the random roll chance of finale_
  - _Adds `ace_vehicle_damage_canHaveFinale` to disable feature via configs_

Depends upon https://github.com/acemod/ACE3/pull/9060
</details>

### DIFF
--- a/addons/cookoff/functions/fnc_cookOff.sqf
+++ b/addons/cookoff/functions/fnc_cookOff.sqf
@@ -16,7 +16,7 @@
  * Public: No
  */
 
-params ["_vehicle", "_intensity", ["_instigator", objNull], ["_smokeDelayEnabled", true], ["_ammoDetonationChance", 0], ["_detonateAfterCookoff", false], ["_fireSource", ""], ["_canRing", true], ["_canJet", true], ["_maxIntensity", MAX_COOKOFF_INTENSITY, [0]]];
+params ["_vehicle", "_intensity", ["_instigator", objNull], ["_smokeDelayEnabled", true], ["_ammoDetonationChance", 0], ["_detonateAfterCookoff", false], ["_fireSource", ""], ["_canRing", true], ["_maxIntensity", MAX_COOKOFF_INTENSITY, [0]], ["_canJet", true, [true]]];
 
 if (GVAR(enable) == 0) exitWith {};
 if !(GVAR(enableFire)) exitWith {};
--- a/addons/vehicle_damage/functions/fnc_handleCookoff.sqf
+++ b/addons/vehicle_damage/functions/fnc_handleCookoff.sqf
@@ -44,7 +44,8 @@ if (!_alreadyCookingOff && { _chanceOfFire >= random 1 }) exitWith {
         _source = ["hit_engine_point", "HitPoints"];
     };
 
-    [QEGVAR(cookOff,cookOff), [_vehicle, _intensity, _injurer, _delayWithSmoke, _fireDetonateChance, _detonateAfterCookoff, _source, _canRing, _canJet]] call CBA_fnc_localEvent;
+    // sending nil for _maxIntensity (9th param) to use default value in ace_cookoff_fnc_cookOff
+    [QEGVAR(cookOff,cookOff), [_vehicle, _intensity, _injurer, _delayWithSmoke, _fireDetonateChance, _detonateAfterCookoff, _source, _canRing, nil, _canJet]] call CBA_fnc_localEvent;
     _vehicle setVariable [QGVAR(cookingOff), true];
     LOG_4("Cooking-off [%1] with a chance-of-fire [%2] - Delayed Smoke | Detonate after cookoff [%3 | %4]",_vehicle,_chanceOfFire,_delayWithSmoke,_detonateAfterCookoff);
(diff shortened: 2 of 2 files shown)


addons/cookoff/script_component.hpp
```diff
@@ -29,6 +29,9 @@
 #define MAX_TIME_BETWEEN_AMMO_DET 25
 #define MAX_COOKOFF_INTENSITY 10
 
+#define COOKOFF_FLAME_STAGE 1 // How many flame executions to allow
+#define COOKOFF_FINALE_STAGE 2 // When to trigger finale stage
+
```
TheCandianVendingMachine, line 34: This is just me, but I would like some variance here `#define COOKOFF_FLAME_STAGE (1 + floor 3)`
Purely just because I think it's fun for people to walk up to a vehicle they think is safe and have it explode in their face.
It does complicate the code a bit, and isn't blocking

JonBons: I think this should be a setting then because in my use case, I absolutely do not want more than one flameout stage

TheCandianVendingMachine (approved):
lgtm. sorry for not approving this earlier, I think I got caught up on the setting and ended up losing time in the year

johnb432:
Ngl the PR title is misleading. I thought it was *just* adding a chance to explode after the cook-off stages were complete, instead of also adding a hard cap of 1 flame-out stage.

Imo we should drop that hard cap in favor of #9758 's `GVAR(cookoffDuration)` setting. We can tweak 9768's implementation of the setting, but I think it should be done in a similar fashion.

addons/cookoff/initSettings.inc.sqf
```diff
@@ -67,3 +67,13 @@
     {[QGVAR(probabilityCoef), _this] call EFUNC(common,cbaSettings_settingChanged)},
     true // Needs mission restart
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(finaleChance), "SLIDER",
+    [LSTRING(finaleChance_name), LSTRING(finaleChance_tooltip)],
+    LSTRING(category_displayName),
+    [0,1,0.5,2,true], // [min, max, default value, trailing decimals (-1 for whole numbers only)]
```
Mike-MF, line 75:
```suggestion
    [0, 1, 0.5, 2, true], // [min, max, default value, trailing decimals (-1 for whole numbers only)]
```

Just a minor formatting thing.

johnb432: I fully agree with the change, but imo that can be done in #9758. Given the whole file is done in the current format, for consistency's sake I don't want to change it here.

johnb432:
> Imo we should drop the hard cap for flame-out stages in favor of #9758 's `GVAR(cookoffDuration)` setting. We can tweak 9758's implementation of the setting, but I think it should be done in a similar fashion.

This was echoed in internal discussion. I'm trying to implement such changes, but it's harder than I initially anticipated.


>     * _After the cookoff jet/ring flame effects have finished a finale can happen (based on random roll) that results in catastrophic destruction of the vehicle (ex. https://youtu.be/qfldUOrXCGk)_
>       
>       * _Adds `Destruction Finale Chance` CBA setting to adjust the random roll chance of finale_
>       * _Adds `ace_vehicle_damage_canHaveFinale` to disable feature via configs_

Why are we adding a separate config entry for this? We have https://github.com/acemod/ACE3/blob/dda6b9d2dc2d896a53b6e81001e595b17bee1596/addons/cookoff/functions/fnc_cookOff.sqf#L75

I understand it isn't quite the same (the only difference I can spot is the that is happens at a specific time and that an explosion is spawned), but can't we either live with it or tweak that to get the desired behaviour, instead of adding yet another config entry?

Assuming we scrap the above from this PR, the remaining thing this PR adds is:

>     * _Restricted cookoff jet/ring flame effects to a single execution_

But that would be also scrapped, in favor of 9758's cookoff duration setting.

---

I went ahead and pushed a commit with the changes I think are necessary to make it compatible with 9758's cookoff duration setting and added a time between cookoff ending and final explosion happening. The final explosion is controlled by `GVAR(destroyVehicleAfterCookoff) || _detonateAfterCookoff` solely, the new config entry `ace_vehicle_damage_canHaveFinale` was removed as I deemed it unnecessary.

These changes fix a bug with the current implementation: `QGVAR(intensity)` was never updated after the final stage, meaning the PFH runs infinitely long, until the variable is manually set or the vehicle is deleted.

johnb432 (approved):

TheCandianVendingMachine:
I worry that the changes changed the intention of the original PR, but if #9758 does what you say it does it sounds like this PR was going to be defunct anyway. With that in mind, I agree with your points and am approving this

TheCandianVendingMachine (approved):

PR #9069, opened 2022-10-11
Arsenal - Action Buttons

BrettMayson:
- Adds a moddable action system to the arsenal

[gunbag.webm](https://user-images.githubusercontent.com/3946739/195196697-f3b48c32-6b1b-4c11-bf71-f57122d5a222.webm)

### IMPORTANT

- [X] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [X] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [X] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.


addons/gunbag/functions/fnc_weaponName.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * Get gunbag weapon name
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [player] call ace_gunbag_fnc_status
```
rautamiekka, line 13:
```suggestion
 * [player] call ace_gunbag_fnc_weaponName
```

GeorgeBAS:
Could this potentially be extended to a "Weapon on back" action as well? As opposed to current method which is grab weapon > self interact > weapon on back > go back into arsenal and grab 2nd weapon

BrettMayson:
It can be extended by any mod

addons/arsenal/functions/fnc_itemInfo.sqf
```diff
@@ -73,6 +74,7 @@ if (isClass _itemCfg) then {
 
 } else {
     [QGVAR(displayStats), [_display, _control, -1, nil]] call CBA_fnc_localEvent;
+    [QGVAR(displayACtions), [_display, _control, -1, nil]] call CBA_fnc_localEvent;
```
johnb432, line 77: Minor typo here: `QGVAR(displayACtions)` should be `QGVAR(displayActions)`.

LinkIsGrim (approved):

addons/arsenal/functions/fnc_itemInfo.sqf
LinkIsGrim, line 77:
```suggestion
    [QGVAR(displayActions), [_display, _control, -1, nil]] call CBA_fnc_localEvent;
```

BrettMayson:
This should probably have a scripted way to add buttons as well, like stats and sorts. Right now it is only via config

addons/arsenal/functions/fnc_refresh.sqf
```diff
@@ -0,0 +1,63 @@
+#include "script_component.hpp"
+#include "..\defines.hpp"
+/*
+ * Author: Brett Mayson
+ * Refreshes the arsenal to show extenal changes
```
rautamiekka, line 5: Just a small typo:
```suggestion
 * Refreshes the arsenal to show external changes
```

rautamiekka (changes requested):

LinkIsGrim (changes requested):
This'll need some more hand-tuning now because of the changes in #9040.

Current issues:
- [ ] Eden Compatibility
- [ ] Weapons not in the arsenal can be duplicated when put into a gunbag (weapon gets added to `GVAR(virtualItems)` in `FUNC(onArsenalOpen)`)
- [ ] Weapons not in the arsenal are lost when pulled from a gunbag (see above, but opposite).

johnb432:
@LinkIsGrim Is `FUNC(refresh)` still needed though? It's only used once.

LinkIsGrim:
> @LinkIsGrim Is `FUNC(refresh)` still needed though? It's only used once.

It's less messy if that bit's in a function. Since actions can modify player's equipped loadout there needs to be a way to refresh everything. Plus it can be good API in the future.

johnb432:
> This'll need some more hand-tuning now because of the changes in #9040.
> 
>     * [ ]  Weapons not in the arsenal can be duplicated when put into a gunbag (issue with `FUNC(updateUniqueItemsList)`?)
> 
>     * [ ]  Weapons not in the arsenal are lost when pulled from a gunbag (see above).

`FUNC(updateUniqueItemsList)` does not handle items in the left panels, it only handles items in the right panels.

All items in the left tab from the player, even the ones that can't be duplicated, are added to `GVAR(virtualItems)` when the arsenal is opened.

Either `GVAR(virtualItems)` needs to be recreated every time an action is used or we add a way to store whether an item is unique or not.
Since `GVAR(virtualItems)` is a hashmap now, we could store the information as the value to the item (as of now, all values are [`nil`](https://github.com/acemod/ACE3/blob/ad1f50304fc68ff5ee4c01bd68360e0478f778e6/addons/arsenal/functions/fnc_onArsenalOpen.sqf#L107)). However, this would mean a performance hit when adding left panel items, as it would be iterating over a hashmap instead of an [array](https://github.com/acemod/ACE3/blob/ad1f50304fc68ff5ee4c01bd68360e0478f778e6/addons/arsenal/functions/fnc_fillLeftPanel.sqf#L58).

rautamiekka:
> Eit would be iterating over a hashmap instead of an [array](https://github.com/acemod/ACE3/blob/ad1f50304fc68ff5ee4c01bd68360e0478f778e6/addons/arsenal/functions/fnc_fillLeftPanel.sqf#L58).

Something with [`keys`](https://community.bistudio.com/wiki/keys) which returns an array of HM keys, [`get`](https://community.bistudio.com/wiki/get), maybe [`values`](https://community.bistudio.com/wiki/values), maybe [`toArray`](https://community.bistudio.com/wiki/toArray), should work:
```sqf
{
    //Stuff.
} forEach keys GVAR(virtualItems);
```
`get` returning a reference makes stuff fast, if kinda confusing.

johnb432:
[This](https://github.com/acemod/ACE3/commit/ab77b8d05cd3530dfa4d701bac2e00389427178e) should resolve the problem. I did some testing with gunbags and it worked. Still needs to be tested more however.

johnb432:
> [This](https://github.com/acemod/ACE3/commit/ab77b8d05cd3530dfa4d701bac2e00389427178e) should resolve the problem. I did some testing with gunbags and it worked. Still needs to be tested more however.

Cleaned up the code a bit: https://github.com/acemod/ACE3/commit/04c8a6cc13675419cacb1858b9a5d18e23057526

LinkIsGrim:
3DEN needed some touching: `ctrlEventHandlers` are processed in reverse order, so refresh needs to be added first. Also skipped waiting a frame for 3DEN. Other issues should be handled by #9287.

LinkIsGrim (approved):

PR #9082, opened 2022-10-24
Western Sahara Compatibility

Mike-MF:
- Add a compatibility addon for the Western Sahara CDLC
- Mostly just adds ballistics values to weapons and hearing protection to helmets that need it.

- Sidenote: the XMS rifles are a weird off-shoot of the XM8 rifles and there is no data to use for barrel lengths, compared to the vanilla CTAR it's around the same length barrel so the values are matched from that.

### DIFF
--- a/optionals/compat_ws/$PBOPREFIX$
+++ b/optionals/compat_ws/$PBOPREFIX$
@@ -1,0 +1,1 @@
+z\ace\addons\compat_ws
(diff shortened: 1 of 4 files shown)


jonpas:
> Sidenote: the XMS rifles are a weird off-shoot of the XM8 rifles and there is no data to use for barrel lengths, compared to the vanilla CTAR it's around the same length barrel so the values are matched from that.

Put that in a comment in the actual file.

jonpas (approved):

PabstMirror (approved):

PR #8776, opened 2022-01-15
CSW - Add reloading from nearSupplies

LinkIsGrim:
**When merged this pull request will:**
- Add ability for players to reload from nearby vehicles, friendly AI with ammo, boxes, ground holders, and dropped backpacks.
- Make AI reload upon entering an empty CSW.
- Move AI CSW reloading to its own function.
- Change a config typeOf lookup to configOf.
- Fix AI CSW reloading with magazines different than currently loaded magazine.
- Fix some function headers.
- Stop AI from being able to reload CSW's from players or enemies' inventory.
- Change a CBA namespace in CSW to native hashMap.
- Micro-optimizations in reload functions to avoid solving config lookups within loops.
- Improve `common_fnc_removeSpecificMagazine`'s interaction with WeaponHolders.
- Capitalize setting titles.
- Fix noun order in some portuguese strings.
- Add setting to change advanced assembly's ammo returning to use a holder item with vanilla ammo box model instead of a really messy GroundWeaponHolder. Holder is draggable/carryable, but items can't be inserted into it by players. It is still possible to pull magazines from it. Subjective change, but I've seen a fair few communities not use the feature because of this.

### IMPORTANT

- [ ] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3mod.com/).
- [x] [Development Guidelines](https://ace3mod.com/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9085, opened 2022-10-27
CSW - Add CSW ammo handling to wheeled vehicles

killerswin2:
- Add ammo handling to the vehicle's interaction menu
- Adds a few more ammo types that these vehicles require
- This support was extended to SOG and RHS compats. I would do more but I don't own any other CDLCs
- The new csw turrents all got custom load and unload times. 
- magazine locations were left on the vehicle, but they can be moved else were if someone has the time to find the correct name selections for about 75 objects (<- maybe someone who has a lot more free time then me)
- stringTables where added, translations will be needed later
- SOG got two more ui pictures, one for zpu and one for m40 ammo
- RHS slap ammo was added to CSW but only the 400 count for crows weapon systems

**Videos of use**
https://youtu.be/biR3zFmbXpU

**November 1st 2022 Update Video**
https://youtu.be/mUeDi7owfO4


killerswin2:
ARMA3
rhsafrf
rhsgref
rhssaf
rhsusaf
sog
here are svgs of the inheritance that I used

rautamiekka:
All the cDLC's have a **compatibility data** mod in the Workshop for those who don't have the cDLC, the diff being it's only good for joining the multiplayer and checking the stuff out in Arsenal, if the descriptions can be trusted.

addons/csw/CfgMagazineGroups.hpp
```diff
@@ -49,6 +49,36 @@ class GVAR(groups) {
...
+    class 200Rnd_556x45_Box_Tracer_F {
+        200Rnd_556x45_Box_Tracer_F = 1;
+    };
+
+    //LSV_AT weirdness
+    class Vorona_HEAT {
+        Vorona_HEAT = 1;
+    };
+
+    // SGP9 HEAT
```
rautamiekka, line 67:
```suggestion
    // SPG9 HEAT
```
Typo.

optionals/compat_sog/stringtable.xml
```diff
@@ -305,5 +305,11 @@
             <Korean>[CSW] 53식 박격포 포신</Korean>
             <Chinesesimp>[班组] 53式迫击炮炮管</Chinesesimp>
         </Key>
+        <Key ID="STR_ACE_Compat_SOG_m40a1rr_v_8_heat_mag">
+            <English>[CSW] 105x607mm HEAT round</English>
+        </Key>
+        <Key ID="STR_ACE_Compat_SOG_m40a1rr_v_8_apers_mag">
+            <English>[CSW]  105x607mm APERS round</English>
```
rautamiekka, line 312:
```suggestion
            <English>[CSW] 105x607mm APERS round</English>
```

killerswin2:
yeah I'm a bit iffy on compatibility data, the only one that I trust is the sog one.

severgun:
Without code changes this is dirty implementation.

Reload should be only from gunner seat. On some vehicles only turned out.
In your video at 1:03 SPG auto reloaded.
As I understand ammo not removed/unloaded from turret on spawn.
I think vehicle inventory can not be used for loading/unloading mags in this implementation. Will work only from/to backpack?

Also in theory some mags should be interchangeable between vehicle and player's weapon. For example M240.

killerswin2:
>I think vehicle inventory can not be used for loading/unloading mags in this implementation. Will work only from/to backpack?
 [bd396d1](https://github.com/acemod/ACE3/pull/9085/commits/bd396d1b6f8d8adece527cadd52fa7f17df6df0e) Added support for getting ammo out of the vehicle, but originally this work already if you had ammo in the vehicle because CWS uses nearestSupplies for searching. [PR here] (https://github.com/acemod/ACE3/pull/8776)

> Also in theory some mags should be interchangeable between vehicle and player's weapon. For example M240.
m240 is defined that way. https://github.com/acemod/ACE3/blob/a088e717af24b8988f7b1c41e713030b9d0b8cac/optionals/compat_rhs_usf3/CfgMagazineGroups.hpp#L27 There is actually multiple of those in the magazinegroups.

> Reload should be only from turret seat. On some vehicles only turned out. 
I disagree on that idea. If the weapon is out in the open and I can see it and physically touch it then why should I not be able to operate it like a regular static weapon. The difference is that one weapon is attached to wheel while the other is not.


> In your video at 1:03 SPG auto reloaded
Turrets auto reload when empty or unload. I believe that is a base game feature.

addons/csw/functions/fnc_reload_canLoadMagazine.sqf
```diff
@@ -45,6 +45,14 @@ private _ammoNeeded = _desiredAmmo min getNumber (_cfgMagazinesCarryMag >> "coun
 private _loadedMag = "";
 private _isBeltLinking = false;
 
+// Zeus Enhanced will keep the mags in the turrent and set the count to zero
+(magazinesAllTurrets _vehicle) apply {
+    _x params ["_xMag", "_xTurret", "_xAmmo"];
+    if (_xTurret isEqualTo _turret && _xAmmo isEqualTo 0) then {
```
rautamiekka, line 51:
```suggestion
    if (_xTurret isEqualTo _turret && {_xAmmo isEqualTo 0}) then {
```

addons/csw/functions/fnc_reload_handleReturnAmmo.sqf
```diff
@@ -44,6 +44,9 @@ if (isNull _container) then {
     _container = (nearestObjects [_unloadTo, [QGVAR(ammo_holder), "GroundWeaponHolder"], 10]) param [0, objNull];
 };
 
+// is the player in a vehicle? If so the container is the vehicle
+if (isNull _container && (!isNull (objectParent _unloadTo))) then {_container = objectParent _unloadTo;};
```
rautamiekka, line 48:
```suggestion
if (isNull _container && {!isNull (objectParent _unloadTo)}) then {_container = objectParent _unloadTo;};
```

severgun:
> I disagree on that idea. If the weapon is out in the open and I can see it and physically touch it then why should I not be able to operate it like a regular static weapon. The difference is that one weapon is attached to wheel while the other is not.

IRL you can't reach a gun from ground on any vehicle. Even on pickup. Some vehicles have multiple turrents. IRL some weapons can be reloaded only from certain seats. Reload from anywhere feels like exploit.

>ammo handling to wheeled vehicles

What is wrong with tracked?

killerswin2:
My intent was just Wheeled vehicles. I don't have plans to expand this pr. It's already big enough for review.

addons/csw/functions/fnc_getCarryMagazine.sqf
```diff
@@ -19,8 +19,19 @@ params ["_vehicleMag"];
...
-    private _groups = "getNumber (_x >> _vehicleMag) == 1 && {isClass (configFile >> 'CfgMagazines' >> configName _x)}" configClasses (configFile >> QGVAR(groups));
+    private _cfgMag = configFile >> "CfgMagazines";
+    private _ammoCount = getNumber (_cfgMag >> _vehicleMag >>  "count");
+    private _magazines = "getNumber (_x >> _vehicleMag) == 1 && {isClass (configFile >> 'CfgMagazines' >> configName _x)}" configClasses (configFile >> QGVAR(groups));
+
+    // try to find a mag that has the same count as the vehicleMag first, if that doesn't exist give the first element in the group array
     _carryMag = configName (_groups param [0, configNull]);
+    {
+        if (getNumber (_cfgMag >> configName _x >> "count") isEqualTo _ammoCount) then {
+            _carryMag = configName _x;
```
rautamiekka, line 30:
```suggestion
        private _configName = configName _x;
        if (getNumber (_cfgMag >> _configName >> "count") isEqualTo _ammoCount) then {
            _carryMag = _configName;
```
No clue if doing this with this and others will make them faster or slower, though.

killerswin2: Thanks for the suggestion. I think that it is slower tho, because we have to get the name every time. The if statement eliminates that in the current code.

BrettMayson:
imo I don't like the mix of the vanilla system of storing extra ammo. Spawning a vehicle with 3 extra mags, but only being able to unload them and never reload feels really weird to me. I think it'd be better to have a onInit handler that removes all the compatible default ammo and switch them to CSW mags in the inventory whenever the system is enabled.

killerswin2:
That's a genius idea!

Drofseh:
> imo I don't like the mix of the vanilla system of storing extra ammo. Spawning a vehicle with 3 extra mags, but only being able to unload them and never reload feels really weird to me. I think it'd be better to have a onInit handler that removes all the compatible default ammo and switch them to CSW mags in the inventory whenever the system is enabled.

Moving the magazines to inventory may break loadouts for existing missions.

BrettMayson:
> Moving the magazines to inventory may break loadouts for existing missions.

killerswin2:
yeah, I know addItemglobal will do that

Drofseh:
> > Moving the magazines to inventory may break loadouts for existing missions.
> 
> Was there not a way to go over the cargo space limit?

addItemCargoGlobal will do it as killerswin2 says, but that's still problematic as a full inventory prevents players from putting things into the inventory during gameplay.

For example, if the vehicle has a toolkit and is overfilled with CSW ammo, I take the toolkit to repair the vehicle, now I cannot put the toolkit back in the vehicle.

BrettMayson:
`setMaxLoad` to compensate for the additional items?

Still preferable to the current mix of vanilla and inventory

Drofseh:
> `setMaxLoad` to compensate for the additional items?
> 
> Still preferable to the current mix of vanilla and inventory

That could work, get normal inventory capacity, get mass of all vehicle magazines, setMaxLoad to the total, add the vehicle magazines

killerswin2:
Alright, implemented. It's a option. If players don't want it they can have it turned off.

addons/csw/functions/fnc_staticWeaponVehicleInit.sqf
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Example:
+ * [vehicle player] call ace_csw_fnc_staticWeaponVehicleInit;
+ *
+ * Public: No
+ */
+
+params["_vehicle"];
+// exit with the user doesn't want the ammo to be replaced
+if(!GVAR(replaceTurrentAmmoWithCSW)) exitwith {};
```
LinkIsGrim, line 21: "Turrent" > "Turret", everywhere

addons/csw/functions/fnc_staticWeaponVehicleInit.sqf
```diff
@@ -0,0 +1,49 @@
...
+
+        _vehicle removeMagazineTurret [_xMag, _xTurret];
+        private _cswMag = _xMag call FUNC(getCarryMagazine);
+        _addLoad = _addLoad + getNumber(_config >> _cswMag >> "mass");
+        _replacementMags pushback [_cswMag, _xAmmo];
+    };
+} forEach (magazinesAllTurrets _vehicle);
+
+// set the load, so that users can remove the mags and add them other items to the vehicle.
+[_vehicle, _addLoad] remoteExecCall ["setMaxLoad", 2];
```
LinkIsGrim, line 44: This should use CBA events. `remoteExec`/`remoteExecCall` is bad.

addons/csw/stringtable.xml
```diff
@@ -303,6 +303,36 @@
...
+            <French></French>
+            <Japanese></Japanese>
+            <Chinese></Chinese>
+            <Chinesesimp></Chinesesimp>
+            <Italian></Italian>
+            <Czech></Czech>
+            <Polish></Polish>
+            <Spanish></Spanish>
+            <Russian></Russian>
+            <Korean></Korean>
```
LinkIsGrim, line 319: This breaks translations. Remove languages with no translation so they show up in English at least.

addons/csw/stringtable.xml
```diff
@@ -303,6 +303,36 @@
...
```
LinkIsGrim, line 334: Same deal.

addons/csw/stringtable.xml
```diff
@@ -658,6 +688,18 @@
             <Russian>[CSW] Лента 20-мм гранат для ст. гранатомёта</Russian>
             <Korean>[CSW] 20mm 고속유탄발사기 탄띠</Korean>
         </Key>
+        <Key ID="STR_ACE_CSW_GMGBelt96_displayName">
+            <English>[CSW] 40mm Grenade GMG Belt 96 Count</English>
```
LinkIsGrim, line 692: Do we include ammo count like this in any other belts' displaynames? Seems weird to me.

addons/csw/stringtable.xml
```diff
@@ -658,6 +688,18 @@
...
+        </Key>
+        <Key ID="STR_ACE_CSW_SPGHEAT_displayName">
+            <English>[CSW] SPG9 HEAT Rounds</English>
+        </Key>
+        <Key ID="STR_ACE_CSW_SPGHE_displayName">
+            <English>[CSW] SPG9 HE Rounds</English>
+        </Key>
+        <Key ID="STR_ACE_CSW_65_39Belt_displayName">
+            <English>[CSW] 6.5X39mm Belt</English>
```
LinkIsGrim, line 701:
```suggestion
            <English>[CSW] 6.5x39 mm Belt</English>
```

BI mags add a space before the 'mm'. We should do the same for consistency. The 'X' is never capitalized.

addons/csw/initSettings.sqf
```diff
@@ -59,3 +59,18 @@ private _categoryArray = [format ["ACE %1", localize LSTRING(DisplayName)]];
     {[QGVAR(dragAfterDeploy), _this] call EFUNC(common,cbaSettings_settingChanged)},
     false // Needs mission restart
 ] call CBA_fnc_addSetting;
+
+
+[
+    QGVAR(replaceTurrentAmmoWithCSW), "CHECKBOX",
```
LinkIsGrim, line 65: "Turrent" > "Turret"

optionals/compat_rhs_usf3/stringtable.xml
```diff
@@ -97,5 +97,20 @@
             <Russian>[CSW] Mk. 19 40 мм M1001 Канистровый</Russian>
             <Spanish>[CSW] Mk. 19 40mm M1001 Bote de metralla</Spanish>
         </Key>
+        <Key ID="STR_ACE_Compat_RHS_USF3_mag_400rnd_127x99_displayName">
+            <English>[CSW] M2 12.7x99mm HMG Belt 400 Count</English>
```
LinkIsGrim, line 101: See other comment on ammo display names.

LinkIsGrim (changes requested):
Good compat work.

addons/csw/stringtable.xml
killerswin2: Ah that was done, because there a 48 count GMG belt.

addons/csw/XEH_postInit.sqf
```diff
@@ -20,6 +20,10 @@ GVAR(vehicleMagCache) = createHashMap;
 [QGVAR(returnAmmo), LINKFUNC(reload_handleReturnAmmo)] call CBA_fnc_addEventHandler;
 
 
+if (isServer) then 
+{
```
BrettMayson, line 24:
```suggestion
if (isServer) then {
```

BrettMayson (changes requested):

addons/csw/XEH_postInit.sqf
killerswin2: noooooooooooooo, my cpp style lol.

LinkIsGrim:
Linking a belt from a vehicle inventory deletes the magazine and breaks the link condition.

addons/csw/functions/fnc_staticWeaponVehicleInit.sqf
LinkIsGrim, line 1: Needs handling for AI reload (GetIn and Fired events). Also we might want at least one magazine to remain loaded? And take `GVAR(ammoHandling)` into account. Refactoring and then just using `FUNC(staticWeaponInit_handleExtraMags)` might be a good idea for all that.

LinkIsGrim (changes requested):

killerswin2:
I don't have anytime to do this pr for the next three weeks

killerswin2:
someone else, will have to implement changes and bring this up to standards

PR #9096, opened 2022-11-10
Medical Treatment - Improve Medical Actions underwater

SpicyBagpipes:
- Enable the player to apply tourniquets underwater
- No longer waste time trying to apply bandages underwater

### IMPORTANT


Drofseh:
Adding the swimming exception to fnc_treatment seems like it will mean that all medical treatments will be able to take place underwater now.

Some consideretion is going to have to be given to deciding to allow things like CPR, splints, stitching, etc, underwater, or if they should also be excluded with isSwimming checks. Probably in fnc_canStitch, fnc_canCPR, etc.

SpicyBagpipes:
> Adding the swimming exception to fnc_treatment seems like it will mean that all medical treatments will be able to take place underwater now.
> 
> Some consideretion is going to have to be given to deciding to allow things like CPR, splints, stitching, etc, underwater, or if they should also be excluded with isSwimming checks. Probably in fnc_canStitch, fnc_canCPR, etc.

Apologies for late reply. I agree, however at the moment at least in my opinion, bandaging is the only action that should be prohibited due to bandages being "soaked". I think for other actions community feedback would be best for determining this and such.

PabstMirror:
First thing we should do is define how we want the system to work.
IMHO only tournaquiting makes sense, bandages (especially quick clot) wouldn't really work, medication and cpr are probably also out.

Right now it's in a weird state where the medical menu shows and you can do any treatment though it
but can't do anything through the interaction menu

https://github.com/acemod/ACE3/blob/master/addons/medical_gui/CfgVehicles.hpp
`ACE_Medical` -> `exceptions[] = {"isNotInside", "isNotSitting"};` 
`ACE_Medical_Menu` -> `exceptions[] = {"isNotInside", "isNotSwimming"};` (note sitting vs swimming)

And I assumed treatment would fail once started via progress bar checking isSwimming but that check fails because
`ace_common_fnc_isSwimming` only check animation
so start treatment, animation changes, then `[player] call ace_common_fnc_isSwimming` = false

What I think we should do:
- add exception for `"isNotSwimming"` to the base interactions so the medical interactions can show
- add config `allowedUnderwater =1` for tourniquit action (default of 0 will be good for compatibility)
- add check to ace_medical_treatment_fnc_canTreat 
```
((getNumber (_config >> "allowedUnderwater")) == 1) || {!([_medic] call ace_common_fnc_isSwimming)}
```

SpicyBagpipes:
Done. Only thing is the docs do not have anywhere suitable to add the new `allowedUnderwater` flag so I have elected to not add it

BrettMayson (approved):

PR #9099, opened 2022-11-20
Refuel - Fix warnings when building with latest pboProject

Dahlgren:
- Fix macro and access property in refuel addon causing warnings with latest pboProject

### IMPORTANT

### DIFF
--- a/addons/refuel/CfgVehicles.hpp
+++ b/addons/refuel/CfgVehicles.hpp
@@ -38,7 +38,7 @@
                 icon = QPATHTOF(ui\icon_refuel_interact.paa); \
             }; \
         }; \
-    };
+    }
 
 class CBA_Extended_EventHandlers;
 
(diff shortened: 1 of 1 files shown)


veteran29 (approved):
What is the `access` property responsible even for?

PabstMirror:
mikro says
>access= anything is bis-only to protect thier classes.
Remove this line, it is no use to you, and deadly to the engine.

https://community.bistudio.com/wiki/Arma:_GUI_Configuration#Properties
has some info, doesn't seem like something we need to be touching
> 0 - ReadAndWrite - this is the default case where properties can still be added or overridden.
1 - ReadAndCreate - this only allows creating new properties.
2 - ReadOnly - this does not allow to do anything in deriving classes.
3 - ReadOnlyVerified - this does not allow to do anything either in deriving classes, and a CRC check will be performed

PabstMirror (approved):

Issue #9027, opened 2022-09-07
ACE3 medical: Unconsciousness miscalculated on limbs. (Unexpected behavior after february 2022 update)


olimolly:
**Mods (complete and add to the following information):**
- **Arma 3:** ` 2.10 Main branch latest `
- **CBA:** `3.15.7 latest`
- **ACE3:** `3.15.0 latest`

**Description:**
With ace medical setting "summa of traumatisme" (ace_medical_fatalDamageSource), the pain on limbs is miscalculated and not turning unconscious the AI. (ace_medical_painUnconsciousChance).
The consequence is that AI are ignoring damages if you don't touch their head or torso. It takes more or less 30 bullets for an AI to die aiming only at legs, even with (ace_medical_AIDamageThreshold) at minimum. 
I was using those settings before multiple times, and AI legs were never that much insensitives. That is why it makes me think it's a bug after those tests again, different weapons different units, armor, uniforms...
Aiming at the torso will make the same damage as usual (depending armor vest off course).

I tested with different settings but here is an export you can load:
[pastebin ace medical settings](https://pastebin.com/NgpU4kFm)

**Steps to reproduce:**
_Follow [https://ace3.acemod.org/img/wiki/user/issue_flowchart.webp](this flowchart)!_

1. load medical settings as this value being important for the test:
`
force ace_medical_fatalDamageSource = 1;
force ace_medical_AIDamageThreshold = 0.143474;
force ace_medical_painCoefficient = 1;
force ace_medical_painUnconsciousChance = 0.5;
force ace_medical_bleefingCoefficient = 0; // we care pain only.
`
2. Place a lot of AI on careless mode. (with a lot of ammo)
3. Shot an AI in the legs only, 

Optionnal: check the pain of the unit between shots to see if it goes moderate or intense, as I understood, Unconsciousness should depend of ace_medical_painUnconsciousChance but it's not anymore since the february update.

**Expected behavior:**
AI still standing after a full magazine in the legs.

**Where did the issue occur?**
- **On dedicated**, Editor (Singleplayer) / Editor (Multiplayer)
- Can't reproduce in Ace Virtual Arsenal (even with setting load in "server")
- Can't reproduce on SOG DLC and ace compat.

Log Files (useless):
- Link to [pastebin rpt](https://pastebin.com/zV0U6JyV)

**Problem:**

https://github.com/acemod/ACE3/commit/73a7dbdc1ed90d3bbb41d453817263f99fc247b0
[...]

PR #9100, opened 2022-11-20
#9027 Gives back ability to put AI unconscious by shooting at limbs

olimolly:
Fix related to #9027 

- Gives back ability to put AI unconscious by shooting at limbs with painUnconsciousChance setting. 
- Change `ace_medical_const_painUnconscious `to 0.35 (from 0.5).
- Change default `painUnconsciousChance `to 0.3 (from 0.1) to add a little dose of variation.

TIP: If you want your AI to keep being limbs insensitive:
- Just set force `ace_medical_fatalDamageSource` to 0 (head and so called "heart") which is exactly the goal of that setting already.

### IMPORTANT

- [ ] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.

### DIFF
--- a/addons/medical_damage/initSettings.sqf
+++ b/addons/medical_damage/initSettings.sqf
@@ -30,7 +30,7 @@
     "SLIDER",
     [LSTRING(PainUnconsciousChance_DisplayName), LSTRING(PainUnconsciousChance_Description)],
     ELSTRING(medical,Category),
-    [0, 1, 0.1, 2, true],
+    [0, 1, 0.3, 2, true],
     true
 ] call CBA_fnc_addSetting;
 
--- a/addons/medical_engine/script_macros_medical.hpp
+++ b/addons/medical_engine/script_macros_medical.hpp
@@ -74,7 +74,7 @@
 
 // --- pain
 #define PAIN_UNCONSCIOUS EGVAR(medical,const_painUnconscious)
-#define PAIN_UNCONSCIOUS_DEFAULT 0.5
+#define PAIN_UNCONSCIOUS_DEFAULT 0.35
 
 // Pain fade out time (time it takes until pain is guaranteed to be completly gone)
 #define PAIN_FADE_TIME EGVAR(medical,const_painFadeTime)


Salluci:
How does changing a setting's default value fix the underlying issue you're describing?

olimolly:
`ace_medical_const_painUnconscious ` is a constant and not meant to be customisable in settings.

The purpose of this constant is to be at the primary level. Other variables are reffering to this constant in order to have a functionnal logical system.
But it has been forgotten to adapt it after february update. The redesign of pain system forgot it. New pain system tone down the pain variable and this constant lost its purpose. It was a threshold that could not been reached anymore by limbs pain values.

Mike-MF:
Changing the setting default doesn't make much sense, virtually nobody runs ACE as default. And if they do raising the default after this long will just bring people asking questions about why it's happening more now.

Salluci:
> It is a threshold that can't be reached anymore by limbs pain values. 

Fair enough for the change on the constant. I wouldn't change the setting's default value, though.

severgun:
I don't know why but that was intended to hide and hardcode value. Read comments.
https://github.com/acemod/ACE3/pull/8394

olimolly:
> I still don't understand why, but that was intended to hide and hardcode value. Read comments. #8394

Oh wow I did not see that, but yes same issue. Would be a solution with a default balanced value off course. _(0.35 imo)._

It is an easy solution but best solution would be to add specific code to manage limbs pain, as head and torso having specific code.

LinkIsGrim:
> How does changing a setting and a magic var's default values fix the underlying issue you're describing?

Most firearm hits to limbs will already cause pain near 1 due to low limb armor. The only scenario where this isn't the case is against armored uniforms. Changing the constant's value will do absolutely nothing. Increase the chance for unconsciousness from pain (or the pain coefficient) if you want units to go unconscious more often.

olimolly:
> Increase the chance for unconsciousness from pain (or the pain coefficient) if you want units to go unconscious more often.

The problem is that it also affects players as well as AIs. So you would suggest having both unconscious chance for AI only and also player only.

But now if you merged the slider for painUnconsciousThreshold. It will allow armored AI to be killed with bullets on limbs only. Specifally requested for sniping with 12.7. Adding the slider will help.

PR #7950, opened 2020-10-08
Dragging - Add new dragging animations

Malbryn:
**When merged this pull request will:**
- Add new dragging animations when the player has a secondary weapon.
- Allow the player to fire their weapon while dragging an object.
- Add new CBA settings to turn on/off this feature.
- French, German, Hungarian and Russian translations of said CBA settings. (Thanks to Floofy Floof, SilberJojo and Vericht for helping me with the translations.)

 **Description:**
- The aim of this feature is to allow players to defend themselves while dragging objects during combat, for eaxmple dragging an unconscious friend to cover.
- If the _"Allow firing while dragging"_ option is enabled in the CBA settings (under ACE Dragging) and the player has a secondary weapon, then the player will drag the object while holding their secondary. If no secondary weapon can be found, the vanilla dragging animation will be used. In both cases the player can fire their gun.
- If the _"Allow firing while dragging"_ option is disabled, the current animation and release action will be used and the player can't fire.

**Demo:**
- [Imgur](https://i.imgur.com/ejySyVJ.mp4)

PR #9112, opened 2022-12-27
Dragging - Fix unknown DIK key in dragging code

killerswin2:
#7950 introduce a unknown DIK for dropping objects, this pr reassigns that value to be the Escape Key


killerswin2:
0xF1, is the left mouse button. Oof

PR #9113, opened 2023-01-04
Rearm - Add scripted way of creating rearm source to framework wiki

TACHarsis:
- Add documentation about how to use ace_rearm_fnc_makeSource to create rearm source object at runtime
- Removed outdated/wrong note

### IMPORTANT


TACHarsis:
I've guessed the version from which on it was included in ACE from the date it was committed and then took the next ACE release from the milestones. So if someone has a better way of determining that, lemme know.

I also tried to mirror how it's documented for refuel as much as possible, since it's been copied from there anyway, but the points on each page are not in the same order and putting it at the buttom like in refuel would have disconnected it from the config way of setting it up. I guess this could be expanded to make the two pages more alike, given their strong parallels?

The hint I removed simply did nothing ingame and I suspect it was how it worked before copying makeSource from refuel?

TACHarsis:
Opened hopefully better PR.

PR #9116, opened 2023-01-06
Optionals - Load dynamically based on presence of mod (`__has_include`)

PabstMirror:
ref discussion on https://github.com/acemod/ACE3/pull/8187

pboProject requires a cfgPatch so this adds a dummy one 
(does nothing and is named differently so shouldn't cause problems)

https://feedback.bistudio.com/T161113 is still a problem when using file patching but unlikely to cause issues


.github/workflows/pboproject.yml
```diff
@@ -35,18 +35,18 @@ jobs:
...
         repository: CBATeam/CBA_A3.git
     - name: Checkout ACE3
-      uses: actions/checkout@v2
+      uses: actions/checkout@master
       with:
         path: z\ace
         persist-credentials: false
     - name: Checkout pull request
-      uses: actions/checkout@v2
+      uses: actions/checkout@master
```
Dahlgren, line 49:
```suggestion
      uses: actions/checkout@v3
```

.github/workflows/pboproject.yml
```diff
@@ -72,13 +72,13 @@ jobs:
       env:
         PYTHONUNBUFFERED: 1
     - name: Archive logs
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@master
```
Dahlgren, line 75:
```suggestion
      uses: actions/upload-artifact@v3
```

tools/.vscode/tasks.json
```diff
@@ -72,21 +72,23 @@
...
             "group": {
                 "kind": "build",
                 "isDefault": true
             }
         },
         {
-            "label": "Build: Hemtt Release",
-            "command": "hemtt.exe",
+            "label": "Build: Hemtt New",
+            "command": "hemttNew.exe",
```
Dahlgren, line 85: This change in binary is not reflected in any docs, revert?

.github/workflows/arma.yml
```diff
@@ -35,14 +35,13 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout the source code
-      uses: actions/checkout@master
-    - name: Build using HEMTT
-      uses: arma-actions/hemtt@master
-      with:
-        command: build --release --ci
+    - uses: actions/checkout@master
```
Dahlgren, line 38:
```suggestion
    - uses: actions/checkout@v3
```

.github/workflows/arma.yml
```diff
@@ -35,14 +35,13 @@ jobs:
...
+      name: Setup HEMTT
+    - name: Build
+      run: hemtt build
     - name: Upload Artifact
-      uses: actions/upload-artifact@v2-preview
+      uses: actions/upload-artifact@v3
       with:
-        name: ace3-${{ github.sha }}-nobin
-        path: releases/ace3_*.zip
+        name: ace3-${{ github.sha }}
```
BrettMayson, line 46: This will still be no bin, as binarize isn't available on Linux

PabstMirror:
```suggestion
        name: ace3-${{ github.sha }}-nobin
```

Dahlgren: Should we run it on Windows instead then? Can use same prepare step as the pboproject job to have binarize etc setup

PabstMirror: right now we really just want it for the linting

but we could also use it to have a dev/master full build
we'd also need to enable script compiling

.github/workflows/arma.yml
```diff
@@ -35,14 +35,13 @@ jobs:
...
-      uses: actions/checkout@master
-    - name: Build using HEMTT
-      uses: arma-actions/hemtt@master
-      with:
-        command: build --release --ci
+    - uses: actions/checkout@v3
+    - uses: arma-actions/hemtt@v1
+      name: Setup HEMTT
+    - name: Build
+      run: hemtt build
```
jonpas, line 42:
```suggestion
      run: hemtt build
    - name: Rename build folder
      run: mv .hemttout/build .hemttout/@ace
```

.github/workflows/arma.yml
```diff
@@ -35,14 +35,13 @@ jobs:
...
     - name: Upload Artifact
-      uses: actions/upload-artifact@v2-preview
+      uses: actions/upload-artifact@v3
       with:
         name: ace3-${{ github.sha }}-nobin
-        path: releases/ace3_*.zip
+        path: .hemttout/build
```
jonpas, line 47:
```suggestion
        path: .hemttout/@*
```

jonpas (changes requested):
Made it directly-loadable by renameing `.hemttout/build` to `.hemttout/@ace` before uploading artifact.

PabstMirror:
artifact doesn't have any bisigns but seems fine otherwise
(compat_ws is not binned but everything else is)

veteran29 (approved):

jonpas (approved):

PR #9121, opened 2023-01-09
Markers - Add the option for UTC/System time timestamps

Daniel-H123:
- Allow people the option to enable In-game time (default), System time or UTC time for marker timestamps
- Allow people to set an offset for their timezone when UTC timestamps are enabled

Discussion: #9119 

### IMPORTANT


Daniel-H123:
Hope this fixed the "Tab detected at ..." problem. Otherwise I have no idea how to fix it. It has maybe to do with my VSCode sqf formatting plugin.

Daniel-H123:
Sorry for the amount of commits, but finaly found the problem. Changed the indenting tabs to spaces. Checked the sqf by running the validation myself and that seemed to work.
3665d81750ac5da7e4632dfcb9106161

addons/markers/initSettings.sqf
```diff
@@ -60,4 +80,4 @@ private _formatDescription = [
         ["HH", "HH:MM", "HH:MM:SS", "HH:MM:SS:MM"],
         1
     ]
-] call CBA_fnc_addSetting;
+] call CBA_fnc_addSetting;
```
LinkIsGrim, line 83: Newline at the end of the file.

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -1,28 +1,56 @@
...
-private _description = _display displayctrl IDC_INSERT_MARKER;
+private _description = _display displayCtrl IDC_INSERT_MARKER;
 private _aceTimestamp = _display displayCtrl IDC_ACE_INSERT_MARKER_TIMESTAMP;
 
 // handle timestamp
-if (cbChecked _aceTimestamp && {ACE_player call FUNC(canTimestamp)}) then {
-    private _time = daytime;
+if (cbChecked _aceTimestamp && {
+    ACE_player call FUNC(canTimestamp)
+}) then {
```
LinkIsGrim, line 26:
```suggestion
if (cbChecked _aceTimestamp && {ACE_player call FUNC(canTimestamp)}) then {
```

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -1,28 +1,56 @@
...
+    private _time = [];
+
+    switch (GVAR(timestampTimezone)) do
+    {
```
LinkIsGrim, line 30:
```suggestion
    switch (GVAR(timestampTimezone)) do {
```

Same with the `case`'s

LinkIsGrim (changes requested):

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -17,12 +17,37 @@
...
 private _aceTimestamp = _display displayCtrl IDC_ACE_INSERT_MARKER_TIMESTAMP;
 
 // handle timestamp
 if (cbChecked _aceTimestamp && {ACE_player call FUNC(canTimestamp)}) then {
-    private _time = daytime;
+    private _time = [];
+
+    switch (GVAR(timestampTimezone)) do {
+        default
+        {
```
LinkIsGrim, line 29:
```suggestion
        default {
```

See https://ace3.acemod.org/wiki/development/coding-guidelines.html#51-braces-placement

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -17,12 +17,34 @@
...
+        case 2: {
+            _offset = GVAR(timestampUTCOffset);
+            systemTimeUTC params["", "", "", "_hour", "_min", "_sec"];
+            _hour = _hour + round(_offset);
+            if (_hour < 0) then {
+                _hour = 24 + _hour;
+            };
+            if (_hour > 24) then {
+                _hour = _hour - 24;
+            };
```
LinkIsGrim, line 44: SQF has modulo operator right? Could do 
```suggestion
            _hour = _hour % 24;
```

LinkIsGrim (commented):
See comment, but it's fine as is too.

LinkIsGrim (approved):

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -17,12 +17,34 @@
...
+            _time = _hour + _min/60 + _sec/3600;
+        };
+    };
```
johnb432, line 47:
```suggestion
    private _time = switch (GVAR(timestampTimezone)) do {
        case 1: {
            systemTime params ["", "", "", "_hour", "_min", "_sec"];
            _hour + _min/60 + _sec/3600
        };
        case 2: {
            systemTimeUTC params ["", "", "", "_hour", "_min", "_sec"];
            _hour = _hour + round (GVAR(timestampUTCOffset));
            if (_hour < 0) then {
                _hour = 24 + _hour;
            };
            if (_hour >= 24) then {
                _hour = _hour - 24;
            };
            _hour + _min/60 + _sec/3600
        };
        default {
            dayTime
        };
    };
```

johnb432 (changes requested):
Modulo could also work, but it would need some testing - I tried to use it and it broke with big offsets.

addons/markers/functions/fnc_onButtonClickConfirm.sqf
Daniel-H123: Never used it with negative value's, but works great 👍

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -17,12 +17,29 @@
...
+            _offset = GVAR(timestampUTCOffset);
+            systemTimeUTC params["", "", "", "_hour", "_min", "_sec"];
+            _hour = _hour + round(_offset);
+            _hour = _hour % 24;
+            _time = _hour + _min/60 + _sec/3600;
+        };
+        default {
+            _time = dayTime;
+        };
+    };
```
johnb432, line 42:
```suggestion
    private _time = switch (GVAR(timestampTimezone)) do {
        case 1: {
            systemTime params ["", "", "", "_hour", "_min", "_sec"];
            _hour + _min/60 + _sec/3600
        };
        case 2: {
            systemTimeUTC params ["", "", "", "_hour", "_min", "_sec"];
            _hour = _hour + round (GVAR(timestampUTCOffset));
            _hour = _hour % 24;
            _hour + _min/60 + _sec/3600
        };
        default {
            dayTime
        };
    };
```

Daniel-H123:
> Modulo could also work, but it would need some testing - I tried to use it and it broke with big offsets.

Can you check this or elaborate? The maximum offset should be +- 12 hours so I don't understand where this can cause an error.

johnb432:
> Can you check this or elaborate? The maximum offset should be +- 12 hours so I don't understand where this can cause an error.

I'm sorry I wasn't clear the first time, I should have been.

I tested the `systemTimeUTC` "algorithm" on its own in the debug console, with large offsets outside of the -12 to 12 range. In the instance of -24, it failed.

Granted this type of failure should never happen, but it does worry me regardless. `%` in Arma can return negative numbers, so something like `-12 % 24` would result in `-12` (this is possible in the current implementation, if it's midnight and the offset is -12). I'll take a better look later today at a solution that I find more robust.

Daniel-H123:
> something like `-12 % 24` would result in `-12` (this is possible in the current implementation, if it's midnight and the offset is -12).

Did you test it with the updated modulo solution, because I didn't implement the modulo when you tested it, so it did not work with offsets greater than +-12. Also -12 % 24 should just return 12 right?

LinkIsGrim:
`abs (x % y)`

johnb432:
> `abs (x % y)`

Doesn't work. `abs (0 % -10)` gives `10`, when it should be `14`.

johnb432:
> Doesn't work. `abs (0 % -10)` gives `10`, when it should be `14`.
> 
> I'm working on a fix though.

Well... I got slightly carried away, but I have found a fix and have made several improvements. I'll make a separate PR with the changes though, so this PR doesn't grow in scope.

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -17,12 +17,27 @@
...
+    private _time = switch (GVAR(timestampTimezone)) do {
+        case 1: {
+            systemTime params ["", "", "", "_hour", "_min", "_sec"];
+            _hour + _min/60 + _sec/3600
+        };
+        case 2: {
+            systemTimeUTC params ["", "", "", "_hour", "_min", "_sec"];
+            _hour = _hour + round (GVAR(timestampUTCOffset));
+            _hour = _hour < 0 ? _hour + _hour % 24 : _hour % 24;
+            _hour = if (Hour < 0) then { 24 + Hour };
```
johnb432, line 34:
```suggestion
            _hour = _hour + round ((GVAR(timestampUTCOffset) % 24) + 24 % 24);
            _hour = _hour % 24;
```

Daniel-H123: Modulo goes before adding and subtraction right?

Daniel-H123: Implemented this with a simular approach

johnb432: I tested it rather thoroughly, and it worked.

Daniel-H123:
```sqf
_hour = _hour + round ((GVAR(timestampUTCOffset) % 24) + 24 % 24);
_hour = _hour % 24;
```

can be simplified to:
```sqf
_hour = (_hour + round (GVAR(timestampUTCOffset))) % 24;
```

so it's just the same as only using `_hour = _hour % 24`.

It has the same problem with mod returning just the negative value of the positive mod.



This would be the best approach in my opinion:

```sqf
_hour = (_hour + round (GVAR(timestampUTCOffset))) % 24;
_hour = if (_hour < 0) then { 24 + _hour } else { _hour };
```

Daniel-H123:
There is one thing that I am unsure of. There are a handfull amount of places that use timezones with an offset of 15min intervals, so UTC+5.45 is theoretically possible. I personaly find it a bit messy to allow decimal input in the slider. Maybe its best to add an extra minutes offset selectbox with 4 options (0, 15, 30, 45) to be safe.

Daniel-H123:
@johnb432 I'm finished. Hope this works. I've tested it and it worked great for me. Good luck 😄

addons/markers/functions/fnc_onButtonClickConfirm.sqf
johnb432: > ```sqf
> _hour = _hour + round ((GVAR(timestampUTCOffset) % 24) + 24 % 24);
> _hour = _hour % 24;
> ```
> 
> can be simplified to:
> 
> ```sqf
> _hour = (_hour + round (GVAR(timestampUTCOffset))) % 24;
> ```

No, it can't, because as you pointed out yourself, it would be as `_hour = _hour % 24`.

You could simplify it to:
```sqf
_hour = ((_hour + round GVAR(timestampUTCOffset)) % 24) + 24 % 24;
```

You remove all multiples of 24 with the inner most `%`. You then add 24 to make sure it's positive, but you ensure it's not larger than 24 by taking the modulo.

Seeing as you implemented it in the latest commit means it's been resolved.

addons/markers/initSettings.sqf
```diff
@@ -32,6 +32,37 @@ private _categoryName = format ["ACE %1", localize ELSTRING(map,Module_DisplayNa
...
+    true
+] call CBA_fnc_addSetting;
+
+[
+    QGVAR(TimestampUTCMinutesOffset), "LIST",
+    [LSTRING(TimestampUTCMinutesOffset), LSTRING(TimestampUTCMinutesOffsetDescription)],
+    [_categoryName, LLSTRING(Module_DisplayName)],
+    [
+        [0, 15, 30, 45],
+        [+00, +15, +30, +45],
```
johnb432, line 61: Seeing as they haven't been put into string format, might as well write the numbers in "regular" format.

```suggestion
        [0, 15, 30, 45],
```

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -17,12 +17,27 @@
...
-    private _time = daytime;
+    private _time = switch (GVAR(timestampTimezone)) do {
+        case 1: {
+            systemTime params ["", "", "", "_hour", "_min", "_sec"];
+            _hour + _min/60 + _sec/3600
+        };
+        case 2: {
+            systemTimeUTC params ["", "", "", "_hour", "_min", "_sec"];
+            _hour = _hour + round (GVAR(timestampUTCOffset));
+            _min = _min + GVAR(timestampUTCMinutesOffset);
```
johnb432, line 33: `GVAR(timestampUTCMinutesOffset)` should take the sign of `GVAR(timestampUTCOffset)` into account.

E.g [UTC−09:30](https://en.wikipedia.org/wiki/UTC%E2%88%9209:30) needs to be set to `-10` and `30` for the settings right now, which I don't think is intuitive.

```suggestion
            _min = _min + (GVAR(timestampUTCMinutesOffset) * GVAR(timestampUTCOffset) / abs GVAR(timestampUTCOffset));
```

johnb432: Alternatively, you can add the negative offsets to the settings.

Daniel-H123: Yeah thanks for pointing that out. It is definitly something I did not take in to account.

Daniel-H123: `GVAR(timestampUTCOffset) / abs GVAR(timestampUTCOffset)` does not work with an offset of 0, becauese of devide by zero.

addons/markers/functions/fnc_onButtonClickConfirm.sqf
```diff
@@ -17,12 +17,28 @@
...
 
 // handle timestamp
 if (cbChecked _aceTimestamp && {ACE_player call FUNC(canTimestamp)}) then {
-    private _time = daytime;
+    private _time = switch (GVAR(timestampTimezone)) do {
+        case 1: {
+            systemTime params ["", "", "", "_hour", "_min", "_sec"];
+            _hour + _min/60 + _sec/3600
+        };
+        case 2: {
```
LinkIsGrim, line 30: Add comment explaining why it's done this way

LinkIsGrim (changes requested):

addons/markers/functions/fnc_onButtonClickConfirm.sqf
Daniel-H123: Sorry you mean the switch statement as a whole or just the part you selected?

Daniel-H123: > Add comment explaining why it's done this way

Added comments to the switch. If you mean that specific part why the first 3 params are not included and the _min and _sec are devided, i'm not really sure how I would comment that without being to technical or wordy.

LinkIsGrim: Sorry, I meant case 2 specifically. It's fine if it's wordy, but that needs to be there lest an onlooker reverts it in the future.

Daniel-H123: Oke I think this makes sense.

```sqf
case 2: {
    systemTimeUTC params ["", "", "", "_hour", "_min", "_sec"];
    _hourOffset = round (GVAR(timestampUTCOffset));
    _hour = _hour + _hourOffset;

    // add or subtract minutes offset based on the negative or positive timezone
    _min = if (_hourOffset < 0) then { _min - GVAR(timestampUTCMinutesOffset) } else { _min + GVAR(timestampUTCMinutesOffset) };

    // prevent the timestamp from exceeding 24 hours or going below 0 hours
    _time = ((_hour + _min/60 + _sec/3600) % 24 + 24) % 24;
    _time
};
```

LinkIsGrim (approved):

PR #9128, opened 2023-01-23
Common - Hide cursor during progress bar

JonBons:
- _Hides the cursor while the progress bar dialog is drawn._


PabstMirror:
Might not even need the settings

PabstMirror (approved):

mharis001:
I don't think the setting adds value and the default is the better behaviour.

JonBons:
Yeeted.

mharis001 (approved):
Code LGTM.

PR #9131, opened 2023-01-25
Medical Engine - Ignore Reflector Hitpoints

PabstMirror:
Issue:
```
class class CAManBase: Man {
  class Reflectors {
      class CustomLight_1 {
          hitpoint         = "Light";
```
will cause
```
getAllHitPointsDamage player param [0, []]; ->
[..."hitrightleg","ace_hdbracket","#light"]
```
and break medical completely, this pr should at least allow medical to add the EH and the results seem normal
But I'm not really sure if this is a good fix, I don't have much experience working with `class Reflectors`

### DIFF
--- a/addons/medical_engine/XEH_postInit.sqf
+++ b/addons/medical_engine/XEH_postInit.sqf
@@ -13,6 +13,7 @@
     // Check if last hit point is our dummy.
     private _allHitPoints = getAllHitPointsDamage _unit param [0, []];
     reverse _allHitPoints;
+    while {(_allHitPoints param [0, ""]) select [0,1] == "#"} do { WARNING_1("Ignoring Reflector hitpoint %1", _allHitPoints deleteAt 0); };
 
     if (_allHitPoints param [0, ""] != "ACE_HDBracket") then {
         private _config = configOf _unit;


PabstMirror:
@commy2 - any thoughts on this?
adding to `class reflectiors` and it will still pass the `\dev\test_hitpointConfigs.sqf` test because hd_bracket is the last `class hitpoint` but just not the last entry in `getAllHitPointsDamage`

humans don't have lights built into them so I don't really get why this is being done...
but I think this PR will allow ace to be compatible
without this fix the handleDamage never gets added and it breaks completely

commy2:
Why not use `ARR select CODE` instead of while deleteAt?

addons/medical_engine/XEH_postInit.sqf
```diff
@@ -13,6 +13,7 @@
     // Check if last hit point is our dummy.
     private _allHitPoints = getAllHitPointsDamage _unit param [0, []];
     reverse _allHitPoints;
+    while {(_allHitPoints param [0, ""]) select [0,1] == "#"} do { WARNING_1("Ignoring Reflector hitpoint %1", _allHitPoints deleteAt 0); };
```
LinkIsGrim, line 16: Only downside to using select is not logging directly, but this isn't performance critical.

LinkIsGrim (approved):

PR #9120, opened 2023-01-09
Fix mismatch between rtm and model.cfg

Dahlgren:
**When merged this pull request will:**
- Fix errors when building with latest Mikero Tools

### IMPORTANT

- [ ] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [ ] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [ ] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

### TODO
- [ ] `dragon` requires `Hips`  in skeleton
- [ ] `medical_engine` requires `camo` in skeleton

PR #9135, opened 2023-01-27
Tools - Add dummy rtms to make ace buildable by current mikro tools

PabstMirror:
replace #9120 ref https://github.com/acemod/ACE3/pull/9120#issuecomment-1399817975

current free tools complain about case-sensitivity, I think that will be removed in the future so we don't need to modify the model.cfgs
the medical_engine animations still have problems with bones in the rtms that don't exist in the model (I think)
this does nothing to fix that problem, but will make the current version of ace continue to build the same as before


jonpas (approved):
Ugh, ohkay.

PR #9139, opened 2023-01-29
Common - Fixed CBA keybinds not working with progress bar

JonBons:
- _Due to the way the map control works we have to manually call into CBA to hook eventhandlers onto the display to work_

### DIFF
--- a/addons/common/functions/fnc_progressBar.sqf
+++ b/addons/common/functions/fnc_progressBar.sqf
@@ -33,7 +33,7 @@ createDialog QGVAR(ProgressBar_Dialog);
 private _display = uiNamespace getVariable QGVAR(dlgProgress);
 
 // Ensure CBA keybindings are hooked into the display
-[_display] call (uiNamespace getVariable ["CBA_events_fnc_initDisplayCurator", {}]);
+_display call (uiNamespace getVariable "CBA_events_fnc_initDisplayCurator");
 
 // Hide cursor by using custom transparent cursor
 private _map = _display displayCtrl 101;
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_progressBar.sqf
```diff
@@ -30,8 +30,12 @@ private _player = ACE_player;
 closeDialog 0;
 createDialog QGVAR(ProgressBar_Dialog);
 
-// Hide cursor by using custom transparent cursor
 private _display = uiNamespace getVariable QGVAR(dlgProgress);
+
+// Ensure CBA keybindings are hooked into the display
+[_display] call (uiNamespace getVariable ["CBA_events_fnc_initDisplayCurator", {}]);
```
veteran29, line 36: Is there a point in having fallback value? ACE can't be used without CBA anyway.
```suggestion
_display call (uiNamespace getVariable "CBA_events_fnc_initDisplayCurator");
```

JonBons: Just a matter of habit.

PabstMirror:
on current master (without this PR) I tested the following code and it worked as expected when running a progressbar
```

["testt", "test", "test", {
    systemChat format ["%1 - Down", time];
},
{
    systemChat format ["%1 - Up", time];
}, [47, [false, false, false]], false] call CBA_fnc_addKeybind; // v
```
So I'm not sure if this is needed?

JonBons:
@PabstMirror It only breaks with CBA binds on the numpad that are also used to move the map around in vanilla arma (Num 2/4/6/8, etc)

PabstMirror (approved):

PabstMirror:
confirmed needed for those keys
and cba_events should be able to handle the doubled events for normal keys

PR #9146, opened 2023-02-05
Cookoff - Remove flares from cookoff

Cyruz143:
- Omit flares from vehicle cook-off

Not sure how much of an impact this has to vanilla arma but anyone with **bright** flares will know the eye burning, soul scortching pain of a vehicle exploding when it's loaded with flares. The particles all start to break as 10 flares all cook off on one spot.

### IMPORTANT

### DIFF
--- a/addons/cookoff/functions/fnc_getVehicleAmmo.sqf
+++ b/addons/cookoff/functions/fnc_getVehicleAmmo.sqf
@@ -29,7 +29,7 @@ private _totalAmmo = 0;
     _x params ["_mag", "_turret", "_count"];
     // if the turret is an FFV seat, it takes magazines from the soldier
     if (_count > 0) then {
-        if (GVAR(removeFlares) && { _mag call FUNC(isMagazineFlare) }) then {continue};
+        if (GVAR(removeFlares) && {_mag call FUNC(isMagazineFlare)}) then {continue};
         private _ammo = getText (configFile >> "CfgMagazines" >> _mag >> "ammo");
         private _model = getText (configFile >> "CfgAmmo" >> _ammo >> "model");
         if (_model == "\A3\weapons_f\empty") exitWith {TRACE_3("skipping",_mag,_ammo,_model);};
@@ -42,7 +42,7 @@ private _totalAmmo = 0;
 {
     _x params ["_mag", "_count"];
     if (_count > 0) then {
-        if (GVAR(removeFlares) && { _mag call FUNC(isMagazineFlare) }) then {continue};
+        if (GVAR(removeFlares) && {_mag call FUNC(isMagazineFlare)}) then {continue};
         _ammoToDetonate pushBack [_mag, _count];
         _totalAmmo = _totalAmmo + _count;
     };
--- a/addons/cookoff/functions/fnc_getVehicleAmmo.sqf
+++ b/addons/cookoff/functions/fnc_getVehicleAmmo.sqf
@@ -29,6 +29,7 @@ private _totalAmmo = 0;
     _x params ["_mag", "_turret", "_count"];
     // if the turret is an FFV seat, it takes magazines from the soldier
     if (_count > 0) then {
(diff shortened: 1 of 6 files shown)


addons/cookoff/functions/fnc_isMagazineFlare.sqf
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Public: No
+ */
+
+params ["_magazine"];
+
+private _return = false;
+private _ammo = getText (configFile >> "CfgMagazines" >> _magazine >> "ammo");
+private _intensity = getNumber (configFile >> "CfgAmmo" >> _ammo >> "intensity");
+private _flare = getNumber (configFile >> "CfgAmmo" >> _ammo >> "ace_grenades_flare");
```
mharis001, line 23: `QEGVAR`.

addons/cookoff/functions/fnc_isMagazineFlare.sqf
```diff
@@ -0,0 +1,29 @@
...
+
+if (_intensity isNotEqualTo 0 || _flare isEqualTo 1) then {
+    _return = true;
+};
```
mharis001, line 27: Don't need the `_return` variable. And, personally, prefer `==` and `!=` for numbers.

addons/cookoff/functions/fnc_getVehicleAmmo.sqf
```diff
@@ -29,6 +29,7 @@ private _totalAmmo = 0;
     _x params ["_mag", "_turret", "_count"];
     // if the turret is an FFV seat, it takes magazines from the soldier
     if (_count > 0) then {
+        if (GVAR(removeFlares) && { _mag call FUNC(isMagazineFlare) }) then {continue};
```
mharis001, line 32: No spaces around braces - here and below.

addons/cookoff/functions/fnc_isMagazineFlare.sqf
```diff
@@ -0,0 +1,28 @@
...
+
+private _ammo = getText (configFile >> "CfgMagazines" >> _magazine >> "ammo");
+private _intensity = getNumber (configFile >> "CfgAmmo" >> _ammo >> "intensity");
+private _flare = getNumber (configFile >> "CfgAmmo" >> _ammo >> QEGVAR(grenades,flare));
+
+if (_intensity != 0 || _flare == 1) then {
+    true
+} else {
+    false
+};
```
mharis001, line 28: Put just the condition as the last line (don't need the `if`/`else`), and use lazy eval.

Cyruz143: Adjusted, not using lazy here as you have some cases where flare is 1 but intensity is 0. I.e. ACE hand flares.

veteran29: I'm pretty sure lazy eval will be slower in case of simple number comparison anyway.

PabstMirror:
Wasn't there a problems with flares floating in air during cookoff?

Cyruz143:
> Wasn't there a problems with flares floating in air during cookoff? -Meaning this might not need a setting.

Not sure tbh pabst, cookoff does spawn the ammo usually at the turret height (iirc) so potentially, yes

PabstMirror:
recent example of tracers just hovering over a burning wreck
(this can be handled another time)

but I also think we can just remove all these problematic ammo types without a setting

Drofseh:
>  recent example of tracers just hovering over a burning wreck (this can be handled another time)
> 
> but I also think we can just remove all these problematic ammo types without a setting

iirc this happens when the velocity of the spawned ammo is too low.

Personally though I've never been a fan of the ammo shooting out, especially on armored vehicles. A) it would be exploding inside the vehicle so it couldn't fly out, B) even if it could fly out of would only go a couple meters, not the dozens of meters that happens in ace cookoff

Cyruz143:
As per chat, removed setting

PabstMirror (approved):

PR #9154, opened 2023-02-18
Interaction - Fix opening locked doors for GM buildings

### LINKED ISSUE #9143: Locked doors of GM buildings can still be opened with open door ACE shortcut
**Mods:**
- **Arma 3:** `2.10`
- **CBA:** `3.15.8`
- **ACE3:** `3.15.2`

**Description:**
Locked doors of GM CDLC buildings can still be opened when using the open door ACE shortcut (`ctrl + space`) even though it is not possible to open them with the mousewheel actions.

**Steps to reproduce:**
1. Go into a mission on the Weferlingen map
2. Find a building and go to the front/main door
3. Execute `cursorObject setVariable ["bis_disabled_door_1", 1]`
4. Check that door cannot be opened with mousewheel action
5. Open door with ACE shortcut (`ctrl + space`)

**Expected behavior:**
ACE shortcut shouldn't open locked doors.

- Dedicated / Editor (Singleplayer) / Editor (Multiplayer)

**Log Files:**
- [RPT](https://pastebin.com/zRyrkwZg) (Nothing of interest)

**Additional context:**
Executing `[5] call ace_interaction_fnc_getDoor` returns `door_01` instead of the expected `door_1` as the door index. Only setting `cursorObject setVariable ["bis_disabled_door_01", 1]` will lock for the ACE shortcut but not for mousewheel actions.  One can set both variables `bis_disabled_door_1` and `bis_disabled_door_01` to fix this issue.

This bug also occurs when using the 3DEN module to lock doors instead of setting the variable manually.

PabstMirror:
Fix #9143

e.g. land_gm_euro_house_11_wL
```
_door = "door_01"; // from getDoor
class openDoor_1 {
    condition = "this animationSourcePhase 'door_1_sound_source' <= 0.5";
    statement = "([this, 1, 1] call BIS_fnc_Door);";
```

bis_fnc_door checks
```
// _door is a number
if ((_structure getVariable [format ["bis_disabled_Door_%1", _door], 0]) != 1) 
```

so if we get a "door_0X" then we just check both vars

### DIFF
--- a/addons/interaction/functions/fnc_openDoor.sqf
+++ b/addons/interaction/functions/fnc_openDoor.sqf
@@ -34,9 +34,14 @@ _getDoorAnimations params ["_animations"];
 if (_animations isEqualTo []) exitWith {};
 
 private _lockedVariable = format ["bis_disabled_%1", _door];
+private _lockedVariableAlt = _lockedVariable; // GM Buildings may have door names like door_01 but locking expects door_1
+if ((count _door == 7) && {(_door select [0, 6]) == "door_0"}) then { 
+    _lockedVariableAlt = format ["bis_disabled_door_%1", _door select [6, 1]]; // stip off the leading zero, the check both vars
+};
 
 // Check if the door can be locked aka have locked variable, otherwhise cant lock it
-if ((_house animationPhase (_animations select 0) <= 0) && {_house getVariable [_lockedVariable, 0] == 1}) exitWith {
+if ((_house animationPhase (_animations select 0) <= 0) && 
+    {(_house getVariable [_lockedVariable, 0] == 1) || {_house getVariable [_lockedVariableAlt, 0] == 1}}) exitWith {
     private _lockedAnimation = format ["%1_locked_source", _door];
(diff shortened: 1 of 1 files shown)


addons/interaction/functions/fnc_openDoor.sqf
```diff
@@ -34,9 +34,14 @@ _getDoorAnimations params ["_animations"];
 if (_animations isEqualTo []) exitWith {};
 
 private _lockedVariable = format ["bis_disabled_%1", _door];
+private _lockedVariableAlt = _lockedVariable; // GM Buildings may have door names like door_01 but locking expects door_1
+if ((count _door == 7) && {(_door select [0, 6]) == "door_0"}) then { 
+    _lockedVariableAlt = format ["bis_disabled_door_%1", _door select [6, 1]]; // stip off the leading zero, the check both vars
```
PabstMirror, line 39:
```suggestion
    _lockedVariableAlt = format ["bis_disabled_door_%1", _door select [6, 1]]; // stip off the leading zero then check both vars
```

jonpas (approved):

PR #9156, opened 2023-02-18
Common - Fix script error in fnc_getVehicleCargo

### LINKED ISSUE #9155: ace_common_fnc_getInPosition throwing an error
https://github.com/acemod/ACE3/blob/4de8487db0beeeafecbb316abebdf2941ebb508f/addons/common/functions/fnc_getInPosition.sqf#L177

The function is being called by passing the type of the vehicle to the function, where instead there should be the vehicle itself.

PabstMirror:
Close #9155

### DIFF
--- a/addons/common/functions/fnc_getVehicleCargo.sqf
+++ b/addons/common/functions/fnc_getVehicleCargo.sqf
@@ -15,15 +15,16 @@
  * Public: Yes
  */
 
-params [["_vehicle", "", ["objNull"]]];
+params [["_classname", "", [""]]];
 
-private _config = configFile >> "CfgVehicles" >> _vehicle;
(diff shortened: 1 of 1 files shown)


addons/common/functions/fnc_getVehicleCargo.sqf
```diff
@@ -15,9 +15,9 @@
  * Public: Yes
  */
 
-params [["_vehicle", objNull, [objNull]]];
+params [["_vehicle", "", ["objNull"]]];
```
commy2, line 18:
```suggestion
params [["_classname", "", ["#none"]]];
```
& change _vehicle -> _classname, so it doesn't happen again?

jonpas (approved):

PR #9162, opened 2023-03-01
Casings - Add missing vanilla casing types

Zman6258:
- Add missing vanilla shell casing models.

Most noticeable with 7.62x39, where they swap from steel to brass casings on landing.


PabstMirror (approved):

PabstMirror:
we might want to look into using a hashmap after this pr is merged

BrettMayson:
> we might want to look into using a hashmap after this pr is merged

With a HashMap also add a way for mods to define more?

BrettMayson (approved):

PR #9166, opened 2023-03-08
Respawn - fix script error

PabstMirror:
initRallypoint calls `["ace_rallypointMoved", [_rallypoint, _side]] call CBA_fnc_globalEvent;`
which calls `ace_respawn_fnc_updateRallypoint` but it's missing 3rd arg

causes (in spawned code) `Error Undefined variable in expression: _position`

### DIFF
--- a/addons/respawn/functions/fnc_updateRallypoint.sqf
+++ b/addons/respawn/functions/fnc_updateRallypoint.sqf
@@ -17,7 +17,8 @@
  * Public: No
  */
 
-params ["_rallypoint", "_side", "_position"];
+params ["_rallypoint", "_side"];
+private _position = param [2, getpos _rallypoint];
 
 if (!hasInterface) exitWith {};
 


addons/respawn/functions/fnc_updateRallypoint.sqf
```diff
@@ -17,7 +17,8 @@
  * Public: No
  */
 
-params ["_rallypoint", "_side", "_position"];
+params ["_rallypoint", "_side"];
+private _position = param [2, getpos _rallypoint];
```
veteran29, line 21: why not:
```suggestion
params ["_rallypoint", "_side", ["_position", getPos _rallypoint]];
```

PabstMirror: `_rallypoint` is undefined when making the array (the array is fully constructed and then passed to `params`)

veteran29 (approved):
LGTM

PR #9169, opened 2023-03-15
Arsenal - GM Compat - Add uniform details support

veteran29:
- title
- handle additional insignia and player/character name on the front pocket

### IMPORTANT

GM does it like this:
```sqf
[missionNamespace, "arsenalClosed", {player call gm_core_characters_fnc_updateUniformDetails}] call bis_fnc_addScriptedEventhandler;
```

### DIFF
--- a/optionals/compat_gm/XEH_preinit.sqf
+++ b/optionals/compat_gm/XEH_preinit.sqf
@@ -7,7 +7,7 @@ PREP_RECOMPILE_START;
 PREP_RECOMPILE_END;
 
 // handle GM uniform additonal insignia
-["ace_arsenal_displayClosed", {
+[QEGVAR(arsenal,displayClosed), {
     ACE_arsenal_center call gm_core_characters_fnc_updateUniformDetails;
 }] call CBA_fnc_addEventHandler;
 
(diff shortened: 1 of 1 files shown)


optionals/compat_gm/XEH_preinit.sqf
```diff
@@ -6,4 +6,9 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+// handle GM uniform additional insignia
+[QEGVAR(arsenal,displayClosed), {
+    ACE_arsenal_center call gm_core_characters_fnc_updateUniformDetails;
```
PabstMirror, line 11: maybe check just in-case the unit became null?

```suggestion
    if (alive EGVAR(arsenal,center)) then { EGVAR(arsenal,center) call gm_core_characters_fnc_updateUniformDetails; };
```

veteran29: The GM function does config checks if it can even run on the unit, in case of null it will get empty class so it should early exit, not sure if needed.

PabstMirror (approved):

optionals/compat_gm/XEH_preinit.sqf
mharis001, line 11:
```suggestion
    EGVAR(arsenal,center) call gm_core_characters_fnc_updateUniformDetails;
```

mharis001 (approved):

PR #9194, opened 2023-05-05
Vehicle Damage - Fix blank hit selections resulting in absurd damage to the driver

JonBons:
- _Change damage handling get the overall damage from the vehicle if the hit selection is blank_

**Repro steps**

Using `CUP_B_BMP2_CZ` for testing, some observations: 
- I can drive full-speed into pretty much anything without consequence, very rarely take a wound from a head-on collision with a tree, rock, or building 
- Those collisions will cause track damage usually, but they also damage the front lights (#l svetlo/#p svetlo)
- At some point after enough collisions I'll start taking damage, sometimes quite catastrophically, sometimes less so

I iterated on all kinds of different damage combinations and eventually discovered what seems to be one cause (I'm sure the logic means it applies in other ways as well): 
- If those front lights mentioned above are at damage state 1 (there are seemingly four selections, I only was able to damage two of them via scripting and did the other two via running into things), there's a very high chance that the driver will take a lot of damage from driving through bushes and similar obstacles

It may not happen on the first bush you go through, but it'll happen soon enough. It's probably just dependent on whatever you hit managing to hit one of those obliterated lights to trigger it.

Go into a mission, create the BMP with the following snippet
```sqf
private _bmp = "CUP_B_BMP2_CZ" createVehicle (getPos player);
player moveInDriver _bmp;
```
Damage the lights (this only damages some of the selections, the others need to be damaged by driving into a building)
```sqf
{(vehicle player) setHitpointDamage [_x, 1]} forEach ["#l svetlo","#p svetlo"];
```
Now drive into a building a few times to damage the other selections, after doing so drive over a bush or crash through a small wall and watch as the driver recieves a bunch of wounds.

If you need to repair the vehicle after slamming into a building too many times (resulting in disabled tracks/engine)
```sqf
{(vehicle player) setHitpointDamage [_x, 0]} forEach ["hithull", "hitturret", "hitengine", "hitfuel","hitrtrack","hitltrack"];
```

Here is a video of repro showing toggling the fix on/off: https://youtu.be/KNprZ4aezRY

### DIFF
--- a/addons/vehicle_damage/functions/fnc_handleDamage.sqf
+++ b/addons/vehicle_damage/functions/fnc_handleDamage.sqf
@@ -22,7 +22,15 @@
  * Public: No
  */
 
-params ["_vehicle", "", "_damage", "", "_projectile", "_hitIndex", "", "_hitPoint"];
+params ["_vehicle", "_hitSelection", "_damage", "", "_projectile", "_hitIndex", "", "_hitPoint"];
+TRACE_6("HandleDamage event inputs",_vehicle,_hitSelection,_damage,_projectile,_hitIndex,_hitPoint);
+
+private _returnHit = if (_hitSelection isNotEqualTo "") then {
+    _vehicle getHitIndex _hitIndex
+} else {
+    damage _vehicle
+};
+
 if !(_projectile in ["ace_ammoExplosion", "ACE_ammoExplosionLarge"]) then {
     if (local _vehicle) then {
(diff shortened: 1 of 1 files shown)


JonBons:
After adding debugging, seeing `(vehicle player) getHitIndex -1` returning `0.89` while `damage (vehicle player)` is `0.0411112` is the reason for changing the method here, fixes the huge wounding that can occur from minor damage events.

@TheCandianVendingMachine

LinkIsGrim (approved):

PR #9200, opened 2023-05-15
Tools - Make HEMTT releases equivalent to older releases

jonpas:
- Move translated READMEs to root in HEMTT build.
- Rename final zip file.

Makes the release equivalent to previous releases (except key/signature file names and each optional having its own key, which is a change for the better).

**Requires https://github.com/BrettMayson/HEMTT/pull/500 (Rhai refactor)!**


.hemtt/hooks/pre_build/01_set_version.rhai
```diff
@@ -0,0 +1,4 @@
+let modcpp = HEMTT_VFS.join("mod.cpp").open_file().read();
```
jonpas, line 1: `README.md` version won't be set, we should really set that and commit it before building a release though.

BrettMayson (approved):

.hemtt/hooks/pre_build/01_set_version.rhai
jonpas: @PabstMirror relevant for you mostly I guess.

Issue #9165, opened 2023-03-07
Global Mobilization border fences can't be cut with the wire cutter


Nomas-X:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.12` Stable
- **CBA:** `3.15.8` Stable
- **ACE3:** `3.15.2` Stable

**Description:**
Border fences on the Weferlingen maps in the Global Mobilization CDLC can not be cut with with the wire cutter as there is no interaction to do so.

**Steps to reproduce:**
_Follow [https://ace3.acemod.org/img/wiki/user/issue_flowchart.webp](this flowchart)!_

1. _Go into a mission on the Weferlingen map or place a unit in the editor with the wire cutter._
2. _Walk up to one of the border fences._
3. _Attempt to interact with it via ACE Interaction to cut it down._

**Expected behavior:**
I expected the "Cut Fence" interaction to appear somewhere on the fence, and to be able to execute said interaction to cut down the fence.

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Singleplayer

**Log Files:**
- https://gist.github.com/Nomas-X/0842c4ecd5c32056f2ee22942678f068

**Additional context:**
N/A

**Screenshots:**
No screenshots however I have a video:
https://cdn.discordapp.com/attachments/976224710385877002/1082611474704371722/2023-03-07_13-29-54.mp4

mdx97:
Looks like support was intended to be added by whitelisting the model's P3D ([see here](https://github.com/acemod/ACE3/blob/eab93f739a17a2a40a10042f08bbf8909de228a5/addons/logistics_wirecutter/script_component.hpp#L60-L71C31)), but with the way [`ace_logistics_wirecutter_fnc_isFence` is written](https://github.com/acemod/ACE3/blob/master/addons/logistics_wirecutter/functions/fnc_isFence.sqf), that list will not even be checked if the `configOf` the fence is not null. I tested it, and this is the problem.

I think there's two options to fix:
- Update `ace_compat_gm` to add `EGVAR(logistics_wirecutter,isFence) = 1;` to all the relevant fences.
- Update `isFence` to also check the P3D whitelist even if there is a config entry found for the fence and `ace_logistics_wirecutter_isFence` has not been set on it.

mdx97:
I opened a PR that implements option 2. Seems like the better choice here.

rautamiekka:
[...]

PR #9201, opened 2023-05-16
Logistics Wirecutter - Check if fence model is in `FENCE_P3DS` even if config check fails.

mdx97:
- Change the behavior of `ace_logistics_wirecutter_fnc_isFence` to check the `FENCE_P3DS` list even if `isFence` is unset (or set to 0) in the object's config.
   - It seems unlikely that `isFence` would get set to 0 on an object _and_ its model would be added to `FENCE_P3DS`. We don't even explicitly set this value to 0 anywhere anyways...

### IMPORTANT

closes https://github.com/acemod/ACE3/issues/9165

### DIFF
--- a/addons/logistics_wirecutter/functions/fnc_isFence.sqf
+++ b/addons/logistics_wirecutter/functions/fnc_isFence.sqf
@@ -22,8 +22,8 @@ TRACE_1("Checking if fence",_object);
 private _configOf = configOf _object;
 if !(isNull _configOf) then {
     // Check for isFence entry since we have valid configOf
-    getNumber (_configOf >> QGVAR(isFence)) == 1 // return
-} else {
-    // Check the p3d name against list (in script_component.hpp)
-    (getModelInfo _object select 0) in FENCE_P3DS // return
+    if (getNumber (_configOf >> QGVAR(isFence)) == 1) exitWith { true };
 };
+
+// Check the p3d name against list (in script_component.hpp)
+(getModelInfo _object select 0) in FENCE_P3DS // return
--- a/addons/logistics_wirecutter/functions/fnc_isFence.sqf
+++ b/addons/logistics_wirecutter/functions/fnc_isFence.sqf
@@ -20,10 +20,14 @@ params ["_object"];
 TRACE_1("Checking if fence",_object);
 
 private _configOf = configOf _object;
-if !(isNull _configOf) then {
+private _isConfigFence = if !(isNull _configOf) then {
     // Check for isFence entry since we have valid configOf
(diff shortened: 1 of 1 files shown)


addons/logistics_wirecutter/functions/fnc_isFence.sqf
```diff
@@ -22,8 +22,8 @@ TRACE_1("Checking if fence",_object);
 private _configOf = configOf _object;
 if !(isNull _configOf) then {
     // Check for isFence entry since we have valid configOf
-    getNumber (_configOf >> QGVAR(isFence)) == 1 // return
-} else {
-    // Check the p3d name against list (in script_component.hpp)
-    (getModelInfo _object select 0) in FENCE_P3DS // return
+    if (getNumber (_configOf >> QGVAR(isFence)) == 1) exitWith { true };
```
PabstMirror, line 25: this exitWith is inside a if/then so it won't have any effect

mathew-horner: Been too long since I've SQF'd - I'll refactor so that it works properly.

mdx97: Been too long since I've written SQF - my fault. Will fix!

mdx97: Fixed now. I added brackets around the secondary condition to get lazy evaluation.

addons/logistics_wirecutter/functions/fnc_isFence.sqf
```diff
@@ -20,10 +20,14 @@ params ["_object"];
...
 } else {
-    // Check the p3d name against list (in script_component.hpp)
-    (getModelInfo _object select 0) in FENCE_P3DS // return
+    false
 };
+
+_isConfigFence || {
+    // Check the p3d name against list (in script_component.hpp)
+    (getModelInfo _object select 0) in FENCE_P3DS
+}
```
mharis001, line 33:
```sqf
getNumber (configOf _object >> QGVAR(isFence)) == 1 || {(getModelInfo _object select 0) in FENCE_P3DS}
```

mdx97: Ah, based off the way it was written I thought that `getNumber` required the config entry to be not null, but now I see it just returns 0 when it's `configNull`. Cool, I'll simplify this bit.

PabstMirror:
we could also just add config entries for these fences
it would save having to do the extra `in` test on every object with a config
but it probably isn't that expensive to check both

PabstMirror (approved):

PR #9202, opened 2023-05-17
Reload/ReloadLauncher - Add notifications when buddy loading or linking

Drofseh:
- Add notifications to reloadLauncher
- Also set _reloadTime to `magazineReloadTime min 2.5` instead of just 2.5 if there is no `buddyReloadTime` set

### DIFF
--- a/addons/reloadlaunchers/stringtable.xml
+++ b/addons/reloadlaunchers/stringtable.xml
@@ -18,6 +18,12 @@
             <Chinese>裝載發射器</Chinese>
             <Turkish>Fırlatıcıyı Yükle</Turkish>
         </Key>
+        <Key ID="STR_ACE_ReloadLaunchers_LoadingStarted">
+            <English>is loading your launcher.</English>
+        </Key>
+        <Key ID="STR_ACE_ReloadLaunchers_LoadingAborted">
+            <English>stopped loading your launcher.</English>
+        </Key>
         <Key ID="STR_ACE_ReloadLaunchers_LoadingLauncher">
             <English>Loading launcher...</English>
             <French>Chargement du lanceur...</French>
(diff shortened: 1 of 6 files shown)


addons/reloadlaunchers/stringtable.xml
```diff
@@ -18,6 +18,12 @@
             <Chinese>裝載發射器</Chinese>
             <Turkish>Fırlatıcıyı Yükle</Turkish>
         </Key>
+        <Key ID="STR_ACE_ReloadLaunchers_LoadingStarted">
+            <English>is loading your launcher.</English>
+        </Key>
+        <Key ID="STR_ACE_ReloadLaunchers_LoadingAborted">
+            <English>stopped loading your launcher.</English>
+        </Key>
```
BrettMayson, line 26: This is not going to work for some languages, even though in this case I think it should for every language Arma supports. Not sure if we do this anywhere else, but probably want something like `%1 is loading your launcher`, and format on that.

BrettMayson (commented):
Personally I'd like an option to disable this, but I think it could be enabled by default

addons/reloadlaunchers/stringtable.xml
Drofseh: Yeah I can format it like that instead.

Drofseh:
> Personally I'd like an option to disable this, but I think it could be enabled by default

There are already messages for the loader/linker that give the status, but I don't mind adding a setting if you're really like.

BrettMayson:
In my opinion no reason not to add an option, but I'm fine if we don't want to. I just personally prefer having people say "I'm loading your launcher". Notifications about your own actions don't really matter as much.

Drofseh:
Merged into #9335 instead

PR #9216, opened 2023-06-13
Medical Engine - Improve damage calculation for explosive-resistant armor

LinkIsGrim:
**When merged this pull request will:**
- Scales high armor values using passthrough to prevent gear with high explosive resistance (armor) becoming immune to bullets. See #9063 and others.
- Close #9063.
- Close #9044.
- Close #8315.
- Close #8008.
- Close #7799.
- Close #8956.

~~I've only done limited testing but this doesn't seem to break fire/fall damage/drowning/explosives or have anything beyond the intended effect on wound creation.~~

Doesn't break anything.

This also improves protection from vests and helmets against pistol calibers, especially the lower tier tactical vests and armored uniforms as a side effect, but this is closer to expected performance when using vanilla as a reference.

image

### IMPORTANT

- [ ] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9217, opened 2023-06-14
Medical Damage - Add alternate armor penetration

LinkIsGrim:
- Add alternate damage calculation that takes ammo's penetrative ability into account (aka "caliber"). Includes CBA setting to disable if players prefer previous balance or have custom ammo configs. The idea is that AP ammo should actually make a difference against armored infantry, rather than just cover and vehicles.
- Pass ammo classname to wound handler data (validation for return data remains the same).

Requires #10573.

Armor is converted to RHAe ( pulled from `\a3\Data_F\penetration\*.bisurf`) based on its scaled value and ignored based on projectile's ability to penetrate material.

Balanced around ACE ammo values. We don't have to worry about 3rd party armor values because of #9216 and an armor soft-cap (at RHAe 110m) implemented in this PR. Overpenetration and angle of incidence are taken into account naturally by using impact damage to calculate impact speed, I actually trust the engine's handling of this more than whatever math we cook up.

### IMPORTANT


Salluci:
I'm happy with this for now. Ready for review.

addons/medical_damage/functions/fnc_getAmmoData.sqf
```diff
@@ -0,0 +1,32 @@
...
+ * Returns base damage value, penetration factor, and expected muzzle velocity of a given round, either from a cache or by reading the ammo config.
+ *
+ * Arguments:
+ * 0: Ammo <STRING>
+ *
+ * Return Value:
+ * Base damage value, penetration factor, muzzle velocity <ARRAY of NUMBER>
+ *
+ * Example:
+ * ["B_556x45_Ball"] call ace_medical_engine_fnc_getAmmoHit
```
LinkIsGrim, line 13:
```suggestion
 * ["B_556x45_Ball"] call ace_medical_engine_fnc_getAmmoData
```

addons/medical_damage/functions/fnc_woundsHandlerArmorPenetration.sqf
```diff
@@ -0,0 +1,63 @@
...
+ * Example:
+ * [player, [[0.5, "Body", 1]], "bullet"] call ace_medical_damage_fnc_woundsHandlerArmorPenetration
+ *
+ * Public: No
+ */
+
+// armor 2 with passthrough 0.8
+#define SCALED_UNPROTECTED_VALUE 4
+#define ENGINE_DAMAGE_INDEX 0
+
```
LinkIsGrim, line 25:
```suggestion
```

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -116,7 +116,7 @@ if (_hitPoint isEqualTo "ace_hdbracket") exitWith {
 
     _allDamages sort false;
     _allDamages = _allDamages apply {[_x select 2, _x select 3, _x select 0]};
-    
+
```
LinkIsGrim, line 119:
```suggestion
    
```

addons/medical_engine/script_macros_medical.hpp
```diff
@@ -62,6 +62,9 @@
 #define PENETRATION_THRESHOLD EGVAR(medical,const_penetrationThreshold)
 #define PENETRATION_THRESHOLD_DEFAULT 0.35
 
+// Baseline penetrability used for armor penetration calculation, see (https://community.bistudio.com/wiki/CfgAmmo_Config_Reference#caliber)
+#define ARMOR_PENETRABILITY 0.015
+
```
LinkIsGrim, line 67:
```suggestion
```

addons/medical_damage/functions/fnc_getAmmoData.sqf
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Return Value:
+ * Base damage value, penetration factor, muzzle velocity <ARRAY of NUMBER>
+ *
+ * Example:
+ * ["B_556x45_Ball"] call ace_medical_engine_fnc_getAmmoData
+ *
+ * Public: No
+ */
+
```
LinkIsGrim, line 17:
```suggestion
// Baseline penetrability used for armor penetration calculation, see (https://community.bistudio.com/wiki/CfgAmmo_Config_Reference#caliber)
#define ARMOR_PENETRABILITY 0.015

```

Drofseh:
"Overpenetration will cause lower damage"

This is generally true IRL as well. For bullets of the same calibre a bullet made of a harder material (such as steel for AP bullets) will not deform as much as bullets made of softer material.

IRL it does depends on the contruction of that specific brand/model of bullet, but as a rule of thumb I think we should be fine with this behavior in game.

rautamiekka:
> "Overpenetration will cause lower damage"

Besides less deforming, the projectile has lost at least some of its power by the time of overpenetration from the earlier penetration.

johnb432 (commented):
This thing might "break" (just wouldn't do anything) in the following scenario:
Third party mod adds a wound handler to `bullet`, but constructs the return array with `[_unit, _allDamages, _typeOfDamage]` instead of modifying `_this` by reference (both are allowed, as per docs). Said wound handler is called before this wound handler and "deletes" `_ammo` from args.

No idea how likely it's in practice though.

addons/medical_damage/functions/fnc_woundsHandlerArmorPenetration.sqf
```diff
@@ -0,0 +1,84 @@
+#include "..\script_component.hpp"
+/*
+ * Author: LinkIsGrim
+ * Custom wounds handler for armor penetration. Calculates damage based on round material penetration and unit armor
+ *
+ * Arguments:
+ * 0: Unit That Was Hit <OBJECT>
```
johnb432, line 7:
```suggestion
 * 0: Unit that was hit <OBJECT>
```

johnb432:
> pushBackUnique will duplicate if someone changes the casing of the ammo string but I see no reason to do that and it shouldn't break anything

What worries me in that scenario is if we decide to add more parameters in the future, where it could cause issues.

johnb432:
> Then I'd say the custom wound handler system is new enough that we can break BWC and help users in updating if that's the case. Still need a better approach though.

If it was added in #8278, then I'd agree it's recent enough, but other opinions are welcome.

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -18,53 +18,41 @@
...
-    // Scale armor using passthrough to fix explosive-resistant armor (#9063)
-    // Skip scaling for uniforms and items that don't cover the hitpoint to prevent infinite armor
-    if (_armor > 0) then {
-        if (_itemType == TYPE_UNIFORM) then {
-            _armorScaled = _armor;
-        } else {
-            _armorScaled = (log (_armor / (_passThrough ^ _passThroughEffect))) * 10;
+        // Scale armor using passthrough to fix explosive-resistant & stupid armor (#9063)
+        // Skip scaling for uniforms and items that don't cover the hitpoint to prevent infinite armor
+        if (_itemType != TYPE_UNIFORM && (_armor > _armorLevelStep * 6)) then {
```
PabstMirror, line 51: _armorLevelStep not defined

addons/medical_engine/functions/fnc_getItemArmor.sqf
```diff
@@ -18,53 +18,42 @@
...
-    // Scale armor using passthrough to fix explosive-resistant armor (#9063)
-    // Skip scaling for uniforms and items that don't cover the hitpoint to prevent infinite armor
-    if (_armor > 0) then {
-        if (_itemType == TYPE_UNIFORM) then {
-            _armorScaled = _armor;
-        } else {
-            _armorScaled = (log (_armor / (_passThrough ^ _passThroughEffect))) * 10;
+        // Scale armor using passthrough to fix explosive-resistant & stupid armor (#9063)
+        // Skip scaling for uniforms and items that don't cover the hitpoint to prevent infinite armor
+        private _armorLevelStep = [4, 2] select (_itemType == TYPE_HEADGEAR);
```
johnb432, line 51: Doesn't this change the balance compared to before?

I haven't looked at it in detail, so forgive me if I'm wrong.

johnb432 (approved):
Modded ammunition from RHS, NIArms and 3CB Factions performs all over the place with this enabled. I'm not going to pretend to know anything about penetration of body armor by small calibers, but balancing between mods is very noticeable.
I guess it's not an ACE issue though.

LinkIsGrim:
It's already a problem without this PR. NIArms ammo goes hard on hit but low on caliber for example. This at least finally sets a hard line in the sand for what works and what doesn't, IMO.

But thinking of fixing it:

- I don't like the idea of having more compats we have to manually patch for. Dynamic projectile replacement to use ACE ammo based on some lookup would be great but performance-intensive.
- This could be moved to a separate addon (`medical_armorpen`) with its own sub-compats for altering ammo stats (through hard patches). Hard-edits would disallow disabling this without removing the addon though.
- Hard patches could target magazines and change their `ammo` instead, but again it's compat work. Probably the most compatible option going for that though.
- Base penetration/damage on some other property, preferably from ballistics as that's mostly accurate across all mods, but I'm not a material expert either and there's some things you just can't derive from just ammunition weight, shape, and size.

johnb432:
>It's already a problem without this PR. NIArms ammo goes hard on hit but low on caliber for example. This at least finally sets a hard line in the sand for what works and what doesn't, IMO.

I was just comparing the behaviour of when the setting is off vs. on. I use `Sum of trauma` as the fatal damage source, which makes it, with alternate armor pen off, take more or less the same amount of hits to kill an enemy with the same caliber. From a balancing point of view that's neat, mod provenance doesn't matter; however, from a realism point of view it sucks, as AP or any enhanced ammo like EPR doesn't make a difference.

Turning alternate armor pen on makes EPR and AP shine, although the amount depends on from which mod the ammo is from - is what I originally meant to say.

> I don't like the idea of having more compats we have to manually patch for.

***If*** we add patches, then it would have to be compats via ammo. Mod makers don't always use inheritance to the full extent when it comes to magazines, i.e. readding the `ammo` attribute every time. If we edit the ammo class itself, it's far less to deal with imo.
But as you said. this isn't ideal either, as disabling it means needing to remove the PBO(s). I'm also not a fan of doing things at runtime - need to think about this more.

DJ-Reticuli:
It's actually worse than that.  With 7.62x51, for instance, currently with ACE, M993 AP penetrates body armor less than the default ball round, and M118 boat tail hollow point, Mk316 match king, and Mk319 all actually penetrate body armor way more than the default 308 round, almost like they're 300WM or even a 338 round.  It's literally the opposite of reality.  Should be closer to M993 > Mk319 >Mk316 > default round > M118bthp for body armor penetration.  Then the existing hit damage spec should take over in injury intensity.  If the person shot has no body armor, then just the hit damage should be dealt as it currently is.

As for how to scale the body armor penetration, I don't think it's necessary to have more patches to ammo packs to do any of this after ACE has been enhanced to support such a feature.  How could ACE be modded to do this?  Energy of the round is calculated at the moment of impact, right, and scales the damage, as does angle of impact?  ACE compatible ammunition already must have bullet length, bullet diameter, and ballistic coefficient that is a measure of aerodynamics and cross-sectional density.  All G7 rounds are pointy, though so are some G1 rounds.  At a given kinetic energy and all else being equal, the narrower and/or pointier round will penetrate more.  It knows the twist of the gun the bullets were fired from (tighter produces more spin, which increases penetration), the distance the bullet's been fired (further, reduces the spin), and whether it's passed below the transonic threshold outside the barrel that should give it a chance of tumbling (say, 50%) and producing zero penetration, only bruising.  Obviously, you wouldn't flip that coin for bullets that exited the barrel subsonic.  Bullet names also commonly have AP or HP in the name, which should further enhance or reduce penetration, respectively.  SD or FRAN in the title could also be treated like HP in regards to reduced penetration.

LinkIsGrim:
> Using round names for effect in damage calculations

No.

For the rest of the stuff if you can make it work consistently then open a PR.

LinkIsGrim:
> If we add patches, then it would have to be compats via ammo. Mod makers don't always use inheritance to the full extent when it comes to magazines, i.e. readding the ammo attribute every time. If we edit the ammo class itself, it's far less to deal with imo.
But as you said. this isn't ideal either, as disabling it means needing to remove the PBO(s). I'm also not a fan of doing things at runtime - need to think about this more.

Worst case we make compats under the mods requiring `ace_medical` to be loaded and thus they'll be disabled alongside medical.

addons/medical_damage/functions/fnc_woundsHandlerArmorPenetration.sqf
```diff
@@ -0,0 +1,80 @@
...
+(_ammo call FUNC(getAmmoData)) params ["_hit", "_penFactor", "_typicalSpeed"];
+
+// Skip bad ammo
+if (_hit <= 0) exitWith {_this};
+
+private _damageData = _allDamages select 0; // selection specific
+_damageData params ["_engineDamage", "_bodyPart", "_realDamage"];
+
+private _armorLevelStep = [2, 4] select (_bodyPart == "body");
+private _armor = (_realDamage/_engineDamage) - 2; // remove base armor
```
LinkIsGrim:
```suggestion
#define DAMAGE_SCALING_FACTOR 10
#define UNSCALED_BASE_ARMOR 2

params ["_unit", "_allDamages", "_typeOfDamage", "_ammo"];
TRACE_4("woundsHandlerArmorPenetration",_unit,_allDamages,_typeOfDamage,_ammo);

if (!EGVAR(medical,alternateArmorPenetration)) exitWith {_this};

// See (https://community.bistudio.com/wiki/CfgAmmo_Config_Reference#caliber),
// _penFactor is ammo "caliber" * RHA penetrability, armor plates according to BI are just made of RHAe material
(_ammo call FUNC(getAmmoData)) params ["_hit", "_penFactor", "_typicalSpeed"];

// Skip bad ammo
if (_hit <= 0) exitWith {_this};

private _damageData = _allDamages select 0; // selection specific
_damageData params ["_engineDamage", "_bodyPart", "_realDamage"];

private _armorLevelStep = [2, 4] select (_bodyPart == "body");
private _armor = (_realDamage/_engineDamage) - UNSCALED_BASE_ARMOR; // remove base armor
```

addons/medical_damage/functions/fnc_woundsHandlerArmorPenetration.sqf
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+// This gets close to vanilla values on FMJ ammo
+#define DAMAGE_SCALING_FACTOR 10
+#define UNSCALED_BASE_ARMOR 2
+
+params ["_unit", "_allDamages", "_typeOfDamage", "_ammo"];
+TRACE_4("woundsHandlerArmorPenetration",_unit,_allDamages,_typeOfDamage,_ammo);
+
+if (!EGVAR(medical,alternateArmorPenetration)) exitWith {_this};
```
johnb432: Could move this up before line 28.

addons/medical_damage/functions/fnc_getAmmoData.sqf
```diff
@@ -0,0 +1,36 @@
...
+GVAR(ammoCache) getOrDefaultCall [toLowerANSI _ammo, {
+    TRACE_1("Cache miss",_ammo);
+    private _ammoConfig = configFile >> "CfgAmmo" >> _ammo;
+
+    if (isNull _ammoConfig) then {
+        [0, 0, 0] // return
+    } else {
+        private _hit = getNumber (_ammoConfig >> "hit");
+        private _caliber = (getNumber (_ammoConfig >> "caliber"));
```
rautamiekka:
```suggestion
        private _caliber = getNumber (_ammoConfig >> "caliber");
```

addons/medical_damage/functions/fnc_woundsHandlerArmorPenetration.sqf
```diff
@@ -0,0 +1,81 @@
...
+// _penFactor is ammo "caliber" * RHA penetrability, armor plates according to BI are just made of RHAe material
+(_ammo call FUNC(getAmmoData)) params ["_hit", "_penFactor", "_typicalSpeed"];
+
+// Skip bad ammo
+if (_hit <= 0) exitWith {_this};
+
+private _damageData = _allDamages select 0; // selection specific
+_damageData params ["_engineDamage", "_bodyPart", "_realDamage"];
+
+private _armorLevelStep = [2, 4] select (_bodyPart == "body");
```
rautamiekka, line 43: _I'm hugely disappointed to have noticed in testing that even on my laggy machine the `==` is actually consistently faster than the `isEqualTo toLowerANSI` combo (especially with `toLower`), even when `_bodyPart`'s length is 6x as many letters, somehow `toLowerANSI` is adding quite a bit to the comparison. "Big sad", as they say. I wonder if the 'Arma 3 v2.20' update did that._

rautamiekka (changes requested):

PabstMirror (approved):

addons/medical_damage/functions/fnc_woundsHandlerArmorPenetration.sqf
```diff
@@ -0,0 +1,81 @@
...
+// Impact damage is hit * (impactSpeed / typicalSpeed): https://community.bistudio.com/wiki/CfgAmmo_Config_Reference#typicalSpeed
+// Impact damage is already lowered by engine based on hit angle, so speed and therefore penetration are also naturally lowered
+// Assume typicalSpeed < 1 means no damage dropoff
+private _impactSpeed = (_realDamage/_hit) * (_typicalSpeed max 1);
+
+private _penDepth = _penFactor * _impactSpeed;
+
+// Max damage is the config value, go down from there based on armor penetration
+private _finalDamage = (_hit * ((_penDepth/_armorThickness) min 1)) / DAMAGE_SCALING_FACTOR;
+_damageData set [0, _finalDamage];
```
PabstMirror, line 77: Could make the setting a scalar and do

```suggestion
_damageData set [0, linearConversion [0, 1, EGVAR(medical,alternateArmorPenetration), _hit, _finalDamage]];
```

which allows fully controlling scaling between untouched and adjusted

LinkIsGrim: Not a fan, personally.

Already too many questions about the damage passthrough effect setting

DJ-Reticuli:
I think ammo 'hit' and ammo (non-ACE) 'caliber' in the config.cpp's should be entirely ignored with ACE and relegated to legacy vanilla calculations when people aren't using ACE, and rather ACE should use (forgive me if I'm forgetting something) bullet mass, bullet diameter (smaller penetrates more), pointiness from BC (I believe G7 * 2 roughly equals G1 for this, or maybe G7's boat tail means no *2 multiplication is necessary), velocity at impact, rotations per second of the bullet at impact (not sure if ACE is already tracking bullet rotation in flight as a result of barrel twist and initial muzzle velocity, but bullet rotation is a storage of kinetic energy, especially at higher rates, quadratically with the square of the angular velocity, with wider and/or higher mass bullets storing even more at a given rpm, all other things being equal), and incidence angle all dictate penetration of obstacles, vehicles, and body armor, the latter being heretofore neglected.  

Actual medical damage after penetration would be mostly similar other than being attenuated by having just gone through armor, and except bullet diameter and pointiness would have the opposite effects as in penetration, assuming penetration of the obstacle, vehicle, or body armor had been successful or there was nothing to pass through and defeat in the first place.  If in the case of body armor there wasn't successful penetration, then you've still got bullet mass, velocity, and incidence angle to deal out a bruise of varying level underneath depending on how much armor it just had to defeat.  You'd probably want armor itself to have its own durability or hit points separate from the main medical damage hit points that get reduced for each section of the body armor that is depleted through successive hits.  All body armor has armor values already, so you could just use that and track their reduction for each area of the body separately from the medical damage & body points stuff available.

You could also add a supersonic bullet transonic-induced tumbling flag that is more likely the lower mass the bullet is and the less its rpm, and if a bullet that starts out supersonically eventually goes subsonic, then there's a certain probability the flag is tripped, and if the flag is tripped, the larger of bullet diameter and bullet length is now used instead of exclusively bullet diameter in the penetration and medical injury calculations for girth.  Again, high girth on impact worsens penetration, but deals more extensive soft tissue injury, all other things being equal.  The tumbling flag could even have a probability of being triggered upon impact with a soft target.  A tumbling bullet on an unarmored body, whether it began tumbling in the air or on impact with soft target, would also increase damage to unarmored bodies through energy transfer, at least in this case as a result of the now bigger cross section of the impact.  Heck, the tumbling flag if implemented would also be useful for ACE Advanced Ballistics, causing rapid deceleration, which has its own eventual consequences on later kinetic energy.  An interesting way these start to interact.

In the beginning of such an implementation, the exact magnitudes of each of these above factors in the equation is less important than just getting them worked into some calculation for penetration, post-penetration damage, and bruising. Obviously, you'd now be tracking bullet rotation if you're not already, be tracking the depletable armor values of each body armor part separate from body medical or health points, and you'd have that tumbling flag.  Once something is worked up, then comes the fun part: tweaking their influence on the outcomes until some of the different advantages / disadvantages of the rounds starts showing through and resembling what it should be doing without the need for 'hit' or 'caliber', hopefully first improving penetration realism, but maybe eventually helping injury modeling.  Heck, you could leave the exact magnitudes and even relationships of each factor open for editing in ACE Medical for people to mess with and report back on through the interwebs, effectively crowdsourcing final implementation.

Zantza:
Would it be possible to add a setting that acts as a multiplier to the round's penetration capability? To me it would seem by far the simplest way to eliminate discrepancies between how different mods configure their ammo. Or if people just don't want armor to be as impactful.

LinkIsGrim:
@Zantza sure, just add a new setting to initSettings and a coefficient to the penFactor

olimolly:
If I remember correctly, there was a slider to adjust the armor penetration in the Medical settings before the last update, 
There is now only the checkbox ace_medical_alternateArmorPenetration ?

olimolly:
image

olimolly:
First of all, thank you very much for your efforts in improving the Ace medical system. However, I have noticed some inconsistencies in the latest update. (and sorry third but last post after more investigations).

The first issue is that the default settings are configured to prevent damage to limbs, which seems unusual.
Once activated, the threshold of 5.0 seems much too high as it does not inflict any damage. This is less obvious this time.
Next, I can't see any difference between JHP and FMJ ammunition, despite this being the whole point of the feature.
Finally, it seems that vanilla ammunition overrides all settings.

Besides, the RHS ammunition is causing deadly bruising, but not bleeding? I assume it's a bug, given that there is blood on the ground.

I ran the following linked tests with JHP and FMJ bullets from RHS (visible in the top right corner of the HUD), but could not notice any differences. Also it seems the vanilla handgun is not affected by the update/settings.

The feature is seeding a lot of misunderstanding among the community. Some damage were already controversial before the update and this one is giving more confusion about settings and the damage system.

I think everyone would appreciate if you could clarify some settings or rerun more tests. 
Thank you in advance <3

https://www.youtube.com/watch?v=utSGyUVhAEM&feature=youtu.be

johnb432:
>The first issue is that the default settings are configured to prevent damage to limbs, which seems unusual.

No it doesn't. As you show yourself in your video, the vanilla P99 still damages limbs.

>Once activated, the threshold of 5.0 seems much too high as it does not inflict any damage. This is less obvious this time.

What threshold are you talking about? What do you mean by 'this time'- context?

>Next, I can't see any difference between JHP and FMJ ammunition, despite this being the whole point of the feature.

If you read the PR's description, this is in fact not the whole point of the feature. I'll quote directly:

>The idea is that AP ammo should actually make a difference against armored infantry, rather than just cover and vehicles.

You have tested none of these in your video. Test any AP vs non AP ammo on armor: e.g. base game tracer vs ACE AP, or even RHS M855 vs RHS M855A1 and you'll notice a difference in how many shots it will take to knock a person unconscious (and you'll likely see a difference in wounds created).

>Finally, it seems that vanilla ammunition overrides all settings.

What settings are you referring to - context?

---

>I ran the following linked tests with JHP and FMJ bullets from RHS (visible in the top right corner of the HUD), but could not notice any differences. Also it seems the vanilla handgun is not affected by the update/settings.

ACE balances for the base game. RHS balances things much differently than the base game and we don't take that into account. Hence why you are seeing weird things like:

>Besides, the RHS ammunition is causing deadly bruising, but not bleeding? I assume it's a bug, given that there is blood on the ground.

Why haven't we balanced RHS damage? It's simple: we haven't found a good approach for a solution yet.

olimolly:
Thanks a lot for direct answers !
I thought FMJ were AP but indeed not, I misunderstood that and the feature then, my bad. I'll then do more tests with your informations.

However, I genuinely still don't get everything yet. If the feature aims to change AP amo effect, should not we feel only changes for armored soldier ? Or is it like a needed global rewrite with side effects to be able to include this feature ?



> No it doesn't. As you show yourself in your video, the vanilla P99 still damages limbs.

I just don't get why limbs are damaged when the option in set to never have damage on limbs. 
With the information I have, it just makes me think vanilla is overriding the option since it's having different effect with the g17.

> > Once activated, the threshold of 5.0 seems much too high as it does not inflict any damage. This is less obvious this time.
> What threshold are you talking about? What do you mean by 'this time'- context?

The limbs damaged threshold is set too high by default to use the RHS G17 for example. 
It's obviously not the reference for mod testing but then, all mods amo and weapons needs re-test and patch ? It's like there is hidden armor in these pants as it shows only bruises damages. It's just confusing even if it has effect on killing the ai.

> ACE balances for the base game. RHS balances things much differently than the base game and we don't take that into account. Hence why you are seeing weird things like:

Fair, so same problem as the question above.

Again, i just want to understand and help by giving feedback.

johnb432:
> I just don't get why limbs are damaged when the option in set to never have damage on limbs.

I suggest reading the tooltip of the setting, because that's not at all what that setting does. Feel free to ask on discord if you don't understand something, because it's off topic for this thread.

> However, I genuinely still don't get everything yet. If the feature aims to change AP amo effect, should not we feel only changes for armored soldier ? Or is it like a needed global rewrite with side effects to be able to include this feature ?

It affects all armor, including armor that is in uniforms (which is more than you'd expect). I will try to take a look at what I can do to improve it.

olimolly:
Alright, I won't use my energy and time giving feedback for this anymore, noted. Understandable why more and more people are using KAT.
            
I read the tooltip and used Discord thanks. Maybe you should check how the last feature is appreciated.


            ID="STR_ACE_Medical_Damage_uselimbDamage_DisplayName"
            <English>Use Limb Damage</English>
ID="STR_ACE_Medical_Damage_uselimbDamage_Description"
            <English>Controls whether limb damage is taken into account for sum of trauma calculations.</English>

But yeah I must be the problem.

olimolly:
ID="STR_ACE_Medical_Damage_uselimbDamage_Description"
            Controls whether **limb damage** is taken into account for **sum of trauma calculations**.

PR #9220, opened 2023-06-19
Medical - Add arsenal category

LinkIsGrim:
- Add "ACE Medical" category to ACE Arsenal through API function

Contains all items in ACE_Medical_Treatment_Actions, compatible with third party mods that add medical items (OPTRE, KAM (though not all items from KAM are added apparently? This should be fixed on their end with API)). Should clean up Misc Items a bit.

Uses Medical GUI icon if that's available. Vanilla medical icon is ugly.

See also #9221.

image

### IMPORTANT

### DIFF
--- a/addons/medical_treatment/XEH_postInit.sqf
+++ b/addons/medical_treatment/XEH_postInit.sqf
@@ -1,5 +1,7 @@
 #include "script_component.hpp"
 
+#define ARSENAL_CATEGORY_ICON (["\A3\ui_f\data\igui\cfg\actions\heal_ca.paa", QPATHTOEF(medical_gui,data\categories\bandage_fracture.paa)] select (["ace_medical_gui"] call EFUNC(common,isModLoaded)))
+
 [QEGVAR(medical_status,initialized), {
     params ["_unit"];
 
@@ -61,3 +63,8 @@ if (isServer) then {
         [_toReplace, _replacements] call EFUNC(common,registerItemReplacement);
     } forEach (configProperties [configFile >> QEGVAR(medical,replacementItems), "isArray _x"]);
 }] call CBA_fnc_addEventHandler;
+
+// Custom Arsenal tab
+if (["ace_arsenal"] call EFUNC(common,isModLoaded)) then {
+    [uiNamespace getVariable [QGVAR(treatmentItems), []], localize ELSTRING(medical,Category), ARSENAL_CATEGORY_ICON] call EFUNC(arsenal,addRightPanelButton);
+};
--- a/addons/medical_treatment/XEH_postInit.sqf
+++ b/addons/medical_treatment/XEH_postInit.sqf
@@ -1,7 +1,5 @@
 #include "script_component.hpp"
 
-#define ARSENAL_CATEGORY_ICON (["\A3\ui_f\data\igui\cfg\actions\heal_ca.paa", QPATHTOEF(medical_gui,data\categories\bandage_fracture.paa)] select (["ace_medical_gui"] call EFUNC(common,isModLoaded)))
-
 [QEGVAR(medical_status,initialized), {
     params ["_unit"];
 
@@ -63,8 +61,3 @@ if (isServer) then {
         [_toReplace, _replacements] call EFUNC(common,registerItemReplacement);
     } forEach (configProperties [configFile >> QEGVAR(medical,replacementItems), "isArray _x"]);
 }] call CBA_fnc_addEventHandler;
-
-// Custom Arsenal tab
-if (["ace_arsenal"] call EFUNC(common,isModLoaded)) then {
-    [uiNamespace getVariable [QGVAR(treatmentItems), []], localize ELSTRING(medical,Category), ARSENAL_CATEGORY_ICON] call EFUNC(arsenal,addRightPanelButton);
-};
--- a/addons/medical_treatment/functions/fnc_scanMedicalItems.sqf
+++ b/addons/medical_treatment/functions/fnc_scanMedicalItems.sqf
@@ -31,4 +31,9 @@ private _fnc_isMedicalItem = toString {
     _list pushBack (configName _x);
 } forEach (_fnc_isMedicalItem configClasses (configFile >> "CfgWeapons"));
 
+// Some third-party mods use magazines as medical items
+{
+    _list pushBack (configName _x);
+} forEach (_fnc_isMedicalItem configClasses (configFile >> "CfgMagazines"));
+
 uiNamespace setVariable [QGVAR(treatmentItems), compileFinal str (_list arrayIntersect _list)]
--- a/addons/medical_treatment/functions/fnc_scanMedicalItems.sqf
+++ b/addons/medical_treatment/functions/fnc_scanMedicalItems.sqf
@@ -1,0 +1,24 @@
+#include "script_component.hpp"
+/*
+ * Author: Salluci
+ * Caches all item classnames used in ACE_Medical_Treatment_Actions
+ *
+ * Arguments:
+ * None
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * call ace_medical_treatment_fnc_getAllMedicalItems
+ *
+ * Public: No
+ */
+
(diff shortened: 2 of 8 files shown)


addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -61,3 +63,8 @@ if (isServer) then {
         [_toReplace, _replacements] call EFUNC(common,registerItemReplacement);
     } forEach (configProperties [configFile >> QEGVAR(medical,replacementItems), "isArray _x"]);
 }] call CBA_fnc_addEventHandler;
+
+// Custom Arsenal tab
+if (["ace_arsenal"] call EFUNC(common,isModLoaded)) then {
+    [uiNamespace getVariable [QGVAR(treatmentItems), []], localize ELSTRING(medical,Category), ARSENAL_CATEGORY_ICON] call EFUNC(arsenal,addRightPanelButton);
```
PabstMirror, line 69:
```suggestion
    [call (uiNamespace getVariable [QGVAR(treatmentItems), {[]}]), localize ELSTRING(medical,Category), ARSENAL_CATEGORY_ICON] call EFUNC(arsenal,addRightPanelButton);
```

addons/medical_treatment/functions/fnc_scanMedicalItems.sqf
```diff
@@ -0,0 +1,24 @@
...
+ * Public: No
+ */
+
+private _list = ["ACE_suture"]; // ugly workaround for sutures
+private _config = configFile >> QGVAR(actions);
+{
+    _list append (getArray (_x >> "items"));
+} forEach ("true" configClasses _config);
+
+uiNamespace setVariable [QGVAR(treatmentItems), _list arrayIntersect _list]
```
PabstMirror, line 24:
```suggestion
uiNamespace setVariable [QGVAR(treatmentItems), compileFinal str (_list arrayIntersect _list)]
```
this protects the var

in 2.14 we can actually `compileFinal` arrays - https://community.bistudio.com/wiki/compileFinal
(although it's broken on dev branch)

addons/medical_treatment/functions/fnc_scanMedicalItems.sqf
```diff
@@ -0,0 +1,24 @@
...
+ * Caches all item classnames used in ACE_Medical_Treatment_Actions
+ *
+ * Arguments:
+ * None
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * call ace_medical_treatment_fnc_getAllMedicalItems
```
PabstMirror, line 13:
```suggestion
 * call ace_medical_treatment_fnc_scanMedicalItems
```

addons/medical_treatment/XEH_postInit.sqf
PabstMirror: also this should probably be in preInit so it works in 3den

PabstMirror (changes requested):

addons/medical_treatment/functions/fnc_scanMedicalItems.sqf
```diff
@@ -0,0 +1,24 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * call ace_medical_treatment_fnc_scanMedicalItems
+ *
+ * Public: No
+ */
+
+private _list = ["ACE_suture"]; // ugly workaround for sutures
```
BrettMayson, line 18: There is surely going to be modded items that fall in this same category, perhaps we need an additional way to mark items as needed to go in this category?

Salluci: Config property good enough?

BrettMayson: I would say so yeah

YetheSamartaka:
Have you tried this change with your other PR (https://github.com/acemod/ACE3/pull/9221)? I had difficulties adding more than 10 categories (Seems to be capped at 10, but I'm not entirely sure)

LinkIsGrim:
There's a hard limit on 10 extra categories in the arsenal, yeah. Out of scope for this, but why someone would have more than 10 extra categories is beyond me.

YetheSamartaka:
> 

Well, when this PR and https://github.com/acemod/ACE3/pull/9221/files gets merged, only one of them will be displayed, because adding just one will reach 10 limit and any newly registered categories will not work.

LinkIsGrim:
> > 
> 
> Well, when this PR and #9221 gets merged, only one will be displayed, because adding just one will reach 10 limit and any newly registered categories will not work. So essentially, your other PR is for nothing because it will never be displayed if this gets merged as well due to reaching that limit.
> 
> I like this change and splitting things into categories. In KAM, we have our own Medical category, but your solution is better. However, when you add your other category for Field Rations, it won't work together. I wanted to add a separate category for GreenMag and a couple of other mods as well, so the limit should be addressed when this PR is tampering with the categories in this way.

Not the case on not doing anything, the categories work in addition to the base ones. Both PRs merged:


Category handling is out of scope for this one, and I don't see a good way to do it (adding a scroll bar there would be hell).

YetheSamartaka:
> > > 
> > 
> > 
> > Well, when this PR and #9221 gets merged, only one will be displayed, because adding just one will reach 10 limit and any newly registered categories will not work. So essentially, your other PR is for nothing because it will never be displayed if this gets merged as well due to reaching that limit.
> > I like this change and splitting things into categories. In KAM, we have our own Medical category, but your solution is better. However, when you add your other category for Field Rations, it won't work together. I wanted to add a separate category for GreenMag and a couple of other mods as well, so the limit should be addressed when this PR is tampering with the categories in this way.
> 
> Not the case on not doing anything, the categories work in addition to the base ones. Both PRs merged: 
> 
> Category handling is out of scope for this one, and I don't see a good way to do it (adding a scroll bar there would be hell).

I see. Then I must have made some kind of mistake when I was adding new category using the same way and it was not working for me (At that time, I only had 2 categories added as an extra). Thank you for testing it out together.

veteran29:
> Category handling is out of scope for this one, and I don't see a good way to do it (adding a scroll bar there would be hell).

Excess categories could be hidden under another "category" button (3 dots or something like in mobile apps) that shows overlay menu with them temporarily after being clicked.

addons/medical_treatment/functions/fnc_scanMedicalItems.sqf
```diff
@@ -0,0 +1,34 @@
...
+// get items in ACE_Medical_Treament_Actions, fallback for items without API config property
+{
+    _list append (getArray (_x >> "items"));
+} forEach ("true" configClasses _cfgActions);
+
+{
+    _list pushBack (configName _x);
+} forEach (_fnc_isMedicalItem configClasses (configFile >> "CfgWeapons"));
+
+uiNamespace setVariable [QGVAR(treatmentItems), compileFinal str (_list arrayIntersect _list)]
```
jonpas, line 34: Why are you storing code? I don't like this, this should store an array that gets cache-invalidated as necessary. Not sure how other things in Arsenal are done though.

jonpas: Do we do that for other variables in `ace_arsenal`?

LinkIsGrim: Nope.

jonpas: Then I really don't see the reason. @PabstMirror ?

PabstMirror: this is how we've been storing things computed at prestart
see https://github.com/acemod/ACE3/blob/master/addons/cargo/XEH_preStart.sqf

LinkIsGrim: I can swing either way, but arsenal just leaves things unprotected (performance concerns, I guess). I guess once 2.14 compileFinal hits this can be switch to a proper data type anyway.

PabstMirror (approved):

PR #9225, opened 2023-06-24
Dragging - Add additional weight checks on inventory modification

LinkIsGrim:
- Improve `FUNC(getWeight)` to take PhysX mass into account. This does mean you can't run carrying the vanilla ammo boxes, but they look unwieldy and heavy enough for that to be the case anyway.
- Add weight checking to block sprint or drop objects if their inventory is modified while being carried.
- Change some conditions for setting the status effects and weight checking. This looks and works better IMO, YMMV.
- Add settings for to disable running, not drop overweight items, and skip PhysX mass. Defaults are pre-#8338 behavior.

Should be merged alongside #8338.

As for backpacks, I don't think there's any case where this should be called directly on them, but I can readd that config lookup. Nothing else in CfgVehicles using Vanilla+RHS+CUP has mass property.

Settings:

### IMPORTANT


addons/common/XEH_postInit.sqf
```diff
@@ -20,7 +20,7 @@
 //Status Effect EHs:
 [QGVAR(setStatusEffect), {_this call FUNC(statusEffect_set)}] call CBA_fnc_addEventHandler;
 ["forceWalk", false, ["ace_advanced_fatigue", "ACE_SwitchUnits", "ACE_Attach", "ACE_dragging", "ACE_Explosives", "ACE_Ladder", "ACE_Sandbag", "ACE_refuel", "ACE_rearm", "ACE_Trenches", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
-["blockSprint", false, ["ace_advanced_fatigue", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
+["blockSprint", false, ["ace_advanced_fatigue", "ACE_dragging", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
```
jonpas, line 23:
```suggestion
["blockSprint", false, ["ace_advanced_fatigue", "ace_dragging", "ace_medical_fracture"]] call FUNC(statusEffect_addType);
```

addons/dragging/functions/fnc_startDrag.sqf
```diff
@@ -20,11 +20,14 @@ params ["_unit", "_target"];
...
-if (!GETVAR(_target,GVAR(ignoreWeightDrag),false) && {
-    private _weight = [_target] call FUNC(getWeight);
-    _weight > GETMVAR(ACE_maxWeightDrag,1E11)
-}) exitWith {
-    // exit if object weight is over global var value
+private _weight = if (_target getVariable [QGVAR(ignoreWeightDrag), false]) then {
+    0
+} else {
+    [_target] call FUNC(getWeight)
+};
```
jonpas, line 27:
```suggestion
private _weight = 0;
if !(_target getVariable [QGVAR(ignoreWeightDrag), false]) then {
    _weight = [_target] call FUNC(getWeight);
};
```
`_var = if (cond) then {val} else {val}` is ugly.

jonpas: Same for above file.

jonpas (changes requested):

jonpas:
Will re-review once #8338 is merged.

addons/common/XEH_postInit.sqf
LinkIsGrim: Should all of these be changed to lowercase?

jonpas: Yeah. There is some old stuff where we did some upper-case. It's not a problem, but all lower-case is preferred.

LinkIsGrim: I'll address that in another PR then.

jonpas:
This might break some missions relying on carrying certain boxes. And there isn't an easy way to disable the new added weight from what I can see?

LinkIsGrim:
> This might break some missions relying on carrying certain boxes. And there isn't an easy way to disable the new added weight from what I can see?

Got any ideas? Best I can think of is a param to remove the PhysX mass and remove dropping when overweight.

LinkIsGrim:
Settings will default to pre-#8338 behavior.

jonpas:
I would make the settings use the new behaviour, as I think it makes more sense. But providing them is good.

`dropOnOverweight` setting is probably unnecessary. That's a bug fix in my opinion as that's how it should have always been.

`allowRunWithLightweight` I am on the edge if it should be a setting. I'd say just allow it by default, but maybe someone else can pitch in.

`skipContainerWeight` setting is good, but should be disabled by default. :+1:

LinkIsGrim:
Intent was to *really* not break missions. Will adjust.

jonpas:
I mean, some things change, that's fine. My initial worry was that there was no way to "fix" that mission (regarding PhysX weight) without changing the contents and possibly flow with that. Also I don't trust PhysX so having an option to disable it is very welcome.

LinkIsGrim:
There's a few ways a mission can break:
1. It relied on the player being slow while carrying an item.
2. It relied on the item's PhysX weight not being taken into account for carrying.
3. It relied on the player not being forced to drop an overweight item (this should really only happen if another player accessed it though?). So I guess this covers all but `#3` now.

jonpas:
None of those are truly game-breaking and that's really all we care about, otherwise strive for improved functionality.

LinkIsGrim:
Then this should be good. MP works.

addons/dragging/functions/fnc_startCarry.sqf
```diff
@@ -58,29 +61,32 @@ if (_target isKindOf "CAManBase") then {
...
 // required for aborting animation
 _unit setVariable [QGVAR(carriedObject), _target, true];
 
+// Event Handler for changes to inventory: allows unit to start running if weight is removed or be forced to walk if weight is added, drop object if weight past ACE_maxWeightCarry
+private _ehID = _target addEventHandler ["ContainerClosed", {
+    params ["_container", ""];
+    private _owner = _container getVariable [QEGVAR(common,owner), objNull];
+    [QGVAR(carryingContainerClosed), [_container, _owner], _owner] call CBA_fnc_targetEvent;
+}];
+_target setVariable [QGVAR(carryingContainerClosedEh), _ehID, true];
```
Dystopian, line 88: Why public? It would have sense only if EH runs globally.

addons/dragging/functions/fnc_startCarry.sqf
```diff
@@ -58,29 +61,32 @@ if (_target isKindOf "CAManBase") then {
...
 
-
 // prevents draging and carrying at the same time
 _unit setVariable [QGVAR(isCarrying), true, true];
 
 // required for aborting animation
 _unit setVariable [QGVAR(carriedObject), _target, true];
 
+// Event Handler for changes to inventory: allows unit to start running if weight is removed or be forced to walk if weight is added, drop object if weight past ACE_maxWeightCarry
+private _ehID = _target addEventHandler ["ContainerClosed", {
```
PabstMirror, line 83: event only fires on player accessing inventory
but we only add EH on the carrier's machine
I really don't think it's a problem, just noting that the CBA_fnc_targetEvent probably won't ever fire

adding and removing the EH too all machines in mp would be a big headache
could add EH globably via xeh and check if being dragged and has owner
but I dont think it's a big deal if weight changes mid-carry

PabstMirror (approved):

addons/dragging/functions/fnc_startCarry.sqf
LinkIsGrim: We can switch to `Put` and `Take` right now, it'll only have actual effect in 2.14, but should work in MP.
I misread the locality on `ContainerClosed` originally.

LinkIsGrim: Moved to XEH.

PabstMirror:
added a few extra checks to make sure box is actually being carried/dragged
both events were firing because "owner" was set, 
I think owner could also be set if it was being loaded by cargo or used for repair

LinkIsGrim:
Good to go on my end. IMO we should add an exception to allow interaction on claimed objects for some actions, like [here](https://github.com/LinkIsGrim/ACE3/blob/0e9cd42b565956b0adbfd26858de67a9fe4004bf/addons/interaction/CfgVehicles.hpp#L642).

PR #9228, opened 2023-06-26
Intel Items - Change scope for better Arsenal filtering

LinkIsGrim:
- Title.

Found while testing #9040. If the intent is just to hide them from the arsenal, we should be using scopeArsenal.

### IMPORTANT

### DIFF
--- a/addons/intelitems/CfgMagazines.hpp
+++ b/addons/intelitems/CfgMagazines.hpp
@@ -6,11 +6,13 @@ class CfgMagazines {
         ACE_isUnique = 1;
         GVAR(intel) = 1;
         GVAR(control) = "";
+        scopeArsenal = 0;
     };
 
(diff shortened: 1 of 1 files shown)


addons/intelitems/CfgMagazines.hpp
```diff
@@ -20,6 +22,7 @@ class CfgMagazines {
 
     class XGVAR(document): GVAR(base) {
         author = ECSTRING(common,ACETeam);
+        scope = 2;
```
PabstMirror, line 25: notepad changes seem fine 
but I think scope = 1 should be ok for doc and photo

iirc there are still some other gear scripts that just use `scope == 2` checks and we really don't want these showing up

PabstMirror (approved):

PR #9231, opened 2023-06-27
Tools - Update setup.py for `hemmt dev` filepatching

PabstMirror:
ref https://ace3.acemod.org/wiki/development/setting-up-the-development-environment.html#411-file-patching-setup

e.g.
```
# Use Hemmt Dev Path for arma filepatching:
  y: F:\DEV\ACE3\.hemttout\dev
  n: F:\DEV\ACE3
(y/n):
```

### DIFF
--- a/tools/setup.py
+++ b/tools/setup.py
@@ -17,8 +17,9 @@ PROJECTDIR = "ace"
 CBA = "P:\\x\\cba"
 ##########################
 
+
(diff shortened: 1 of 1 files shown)


jonpas (approved):
Sure, although I don't really see the need. Just load the mod from wherever it's cloned?

PR #9241, opened 2023-07-02
Medical - Change default spontaneous wake up chance & epinephrine boost settings

LinkIsGrim:
- Title.

Epinephrine does nothing with default settings. This is counter-productive for understanding of that mechanic.

Some math for unconsciousness: on average, a stable, unconscious patient will take 3.5 minutes (0.95^14) to reach 50% chance to wake up. Bell curve means this will usually be higher. Setting to 10% halves this without making things easy.

### IMPORTANT


PabstMirror:
ref #7076
this was added more for gameplay reasons than realism
i'm not against this but we should be careful about changing default settings now

imho default settings lead to brutal wake up times, 
but most people have probably changed their settings to what they want

PabstMirror (approved):

PR #9246, opened 2023-07-09
Medical - Add damage handling for explosive damage while inside vehicles

LinkIsGrim:
- Add damage handling for all damage from explosives while inside a vehicle. This includes mines, artillery, explosive charges, AT, explosive projectiles like 40mm grenades, and most importantly catastrophic vehicle explosions.
- Change vehicle checks in `ace_medical_engine_fnc_handleDamage` to `objectParent`. It's faster and handles dead units.

Continuation of @pterolatypus #8832. Includes #9216, should be merged alongside/after.

I wasn't able to figure out why damage from vehicle explosions can be negative, but just using `abs` value puts wounds at where they should be anyway.

Doesn't completely guarantee lethality from catastrophic explosions because of the random body part roll. Should it? I feel like the chance for survival is low enough anyway (have to not roll head or body six times at damage = 1).

### IMPORTANT


PabstMirror:
might be able to remove https://github.com/acemod/ACE3/blob/master/addons/medical_engine/XEH_postInit.sqf#L64-L76 after this

addons/medical_damage/functions/fnc_woundsHandlerVehiclehit.sqf
```diff
@@ -0,0 +1,37 @@
...
+
+// damage can sometimes be negative (why?)
+// damage to structural is low unless it's a very large explosion, in which case it is typically >= 1
+private _damageToApply = (abs (_allDamages select 0 select 0));
+
+private _newDamages = [];
+
+// hitpoints are randomized, more damage means more wounds in different body parts
+for "_i" from 1 to (_damageToApply * 6) do {
+    _newDamages pushBack [_damageToApply * 6, selectRandom ALL_BODY_PARTS, _damageToApply * 6]
```
PabstMirror, line 33: I understand the 6x on line 32, but not sure I get where the 6x on 33 is coming from?

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -32,9 +32,14 @@ if (_hitPoint isEqualTo "") then {
...
 private _newDamage = _damage - _oldDamage;
-// Get armor value of hitpoint and calculate damage before armor
-private _armor = [_unit, _hitpoint] call FUNC(getHitpointArmor);
+// Get scaled armor value of hitpoint and calculate damage before armor
+// We scale using passThrough to handle explosive-resistant armor properly (#9063)
+// We need realDamage to determine which limb was hit correctly
+[_unit, _hitpoint] call FUNC(getHitpointArmor) params ["_armor", "_armorScaled"];
 private _realDamage = _newDamage * _armor;
+if (_hitPoint isNotEqualTo "#structural") then {
+    _newDamage = _newDamage * (_armor/_armorScaled);
```
PabstMirror, line 41: I think we could make this scaling a setting?
just a 0-1 lerp between old way and new

PabstMirror:
this is very subjective but I think it's ok if some explosions are survivable
e.g. `hatchback setDamage 1`, car's don't really explode, so I'm not sure if that should be fatal

addons/medical_engine/functions/fnc_handleDamage.sqf
LinkIsGrim: Why?

LinkIsGrim:
I'm okay with vehicle explosions being sometimes survivable, though that is mostly going to be handled by `selectRandom ALL_BODY_PARTS`.

addons/medical_damage/functions/fnc_woundsHandlerVehiclehit.sqf
LinkIsGrim: IIRC, scaling. Though looking at this again now I'd rather rewrite this loop.

LinkIsGrim:
This needs testing/compatibility with AVD, as that damages crew based on damage sustained by the vehicle. Need to check if it doesn't become excessive.

addons/medical_engine/functions/fnc_handleDamage.sqf
```diff
@@ -50,14 +55,16 @@ if (
...
 // Crashing a vehicle doesn't fire the EH for each hitpoint so the "ace_hdbracket" code never runs
 // It does fire the EH multiple times, but this seems to scale with the intensity of the crash
-private _vehicle = vehicle _unit;
 if (
     EGVAR(medical,enableVehicleCrashes) &&
     {_hitPoint isEqualTo "#structural"} &&
     {_ammo isEqualTo ""} &&
-    {_vehicle != _unit} &&
+    {_inVehicle} &&
     {vectorMagnitude (velocity _vehicle) > 5}
```
LinkIsGrim, line 68: undefined variable

addons/medical_damage/functions/fnc_woundsHandlerVehiclehit.sqf
LinkIsGrim: Remembered: damage here is only gotten from `#structural` hitpoint and only available in `0-1` range, typically very low. So it's multiplied by 6 to account for it representing damage to each limb, sorta. It feels right in game IMO.

veteran29 (approved):

veteran29:
Code looks ok, I did not test it.

addons/medical_damage/functions/fnc_woundsHandlerVehiclehit.sqf
```diff
@@ -0,0 +1,37 @@
+#include "script_component.hpp"
```
PabstMirror, line 1:
```suggestion
#include "..\script_component.hpp"
```

LinkIsGrim:
Yell at me if this breaks. Focus testing concern on RC.

PR #9265, opened 2023-07-14
Common - Add variable to block item replacement temporarily

### LINKED ISSUE #9104: fnc_registerItemReplacement have awkward behavior while in Arsenal
- **Arma 3:** `2.10 profiling`
- **CBA:** `3.15.8 stable`
- **ACE3:** `3.15.2 stable`

**Description:**
When a mission/mods uses fnc_registerItemReplacement to replace an item (A) with another (B)
Let's call Item we want to replace: A
and replaced item(s): B
arsenal will have weird behavior where the count of A stays at 1 and the count of B stays at 0 but the weight increase when you add more of A. It seems like item were converted but the count of B are not updated until you leave and re-enter the arsenal.

Furthermore, having count of A stays at 1 seems odd to the players if they want to add multiple of A. Or players accidentally add more than they actually want but they can't remove it directly because the item are already converted

An built in example of this would be vanilla IFAK which would be converted to bandages and some other medical stuff

**Steps to reproduce:**
_Follow [https://ace3.acemod.org/img/wiki/user/issue_flowchart.webp](this flowchart)!_

1. Enable replace vanilla items in ACE Medical Settings
2. Enter ACE Arsenal
3. Add First Aid Kit to the player

**Expected behavior:**
Not replacing items while user is in arsenal

- Self Hosted Multiplayer
- Single Player

**Log Files:**
- Link to ([gist](https://gist.github.com) or [pastebin](http://pastebin.com)) to the client and/or server RPT file. An instruction to find your RPT files can be found [here](https://community.bistudio.com/wiki/Crash_Files#Arma_3).
- https://pastebin.com/1GXLfcyw

**Additional context:**
Found this problem with KAT Advance Medical mod where painkillers are hard to access since it is in Magazines Catagory due to technical limitations. So the devs uses ace_common_fnc_registerItemReplacement to put a pseudo item in "misc" catagory and replace it with the real painkiller box.

**Screenshots:**
[If applicable, add screenshots to help explain your problem.]
https://user-images.githubusercontent.com/58496201/204175161-e7b4422a-4eab-41e7-8501-b8af895af0aa.mp4

LinkIsGrim:
- Title.
- Block item replacement on arsenal open, resume on arsenal close.
- Close #9104. (Potentially use #9069 to display item replacements later?)

### IMPORTANT


johnb432:
> Should this be a status effect instead?

I'd say it's depends: Do you want this to be publicly available, so that other mods can get and set it?

Personally I'd think it would be nice for it to be a status effect.

PabstMirror:
https://github.com/acemod/ACE3/commit/9a704f9033d0f8a90b195a2e898edeab38ed9353
any thoughts on just re-calling the replace code instead of the fake item

LinkIsGrim:
> [9a704f9](https://github.com/acemod/ACE3/commit/9a704f9033d0f8a90b195a2e898edeab38ed9353) any thoughts on just re-calling the replace code instead of the fake item

Awesome. `compileFinal` to be safe though.

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -20,8 +20,12 @@
 params [["_oldItem", "", [0,""]], ["_newItems", "", ["", []]], ["_replaceInherited", false, [false]]];
 TRACE_3("registerItemReplacement",_oldItem,_newItems,_replaceInherited);
 
+
+// Setup on first run
+if (isNil QGVAR(itemReplacements)) then {
+
 // CBA player event handler function
-private _fnc_replaceItems = {
+DFUNC(registerItemReplacement_replaceItems) = compileFinal toString {
```
PabstMirror, line 28: I didn't tab it over so the diff is readable, but can shift it over if it looks good

LinkIsGrim: LGTM.

PabstMirror (approved):

addons/common/functions/fnc_registerItemReplacement.sqf
```diff
@@ -20,76 +20,87 @@
...
+if (isNil QGVAR(itemReplacements)) then {
 
-    private _newItems = _items - GVAR(oldItems);
-    _newItems = _newItems arrayIntersect _newItems; // Get unique items only
-    if (_newItems isEqualTo []) exitWith {
-        GVAR(oldItems) = _items;
-    };
+    // CBA player event handler function
+    DFUNC(registerItemReplacement_replaceItems) = compileFinal toString {
+        params ["_unit"];
```
PabstMirror, line 29: this function could also possibly be useful for running gear replacement on an AI
if someone had a reason to do so

LinkIsGrim: move this to a proper function?

PR #9266, opened 2023-07-15
Common - Improve PBO checking

### LINKED ISSUE #9181: ACE PBO Check, when set to "Kick", can be bypassed using an Xbox Controller
- **Arma 3:** `2.12 stable` [e.g. 1.00 stable, rc,  dev]
- **CBA:** `3.15.8` [e.g. 3.0.0 stable, commit hash]
- **ACE3:** `3.x15.2` [eg. 3.0.0 stable, commit hash]

**Description:**
With `ace_common_CheckPBOsAction` set to 2 (kick), and with outdated / mismatched PBO's loaded, the server comes up with a warning dialog first, and only after clicking okay on that dialog do you get kicked.
The problem is that, if you plug in an Xbox controller, and instead of clicking okay, press B on the Xbox controller, it closes the dialog and you don't get kicked. Despite there being no cancel / close button on the dialog. 

**Steps to reproduce:**
_Follow [https://ace3.acemod.org/img/wiki/user/issue_flowchart.webp](this flowchart)!_

1. Set up a server running CBA and Ace with `ace_common_CheckPBOsAction = 2;`
2. Remove one of the PBOs from ace that the server has loaded from your client then join the server.
3. After you load in and are presented with a warning dialog saying you have missing addons, instead of clicking okay, press B on a connected Xbox controller
4. Observe that you didn't get kicked. 

**Expected behavior:**
You get kicked from server no matter what input you press.

- Dedicated server

**Additional context:**
Suggestion by veteran29 which could work would be to move the kick logic to Dialog Unload instead of Okay.

johnb432:
- Intended to close #9181.
However, I have not tested if #9181 has been fixed with this, as I do not own a controller.
- `FUNC(errorMessage)` has been modified to be a dedicated function for usage in PBO checking ~~and has been adapted to be executable on clients without any mods~~.
- Added events `"ace_versioning_clientCheckDone"` and `"ace_versioning_serverCheckDone"` for public usage. This would allow other mods to be notified when PBO checking has been completed and whether it has been completed successfully or not. Event names and arguments passed are up for debate: For now, `"ace_versioning_clientCheckDone"` passes a copy of `ACE_Version_ClientErrors`, `"ace_versioning_serverCheckDone"` passes a copy of `ACE_Version_ServerVersions`.
- `FUNC(checkPBOs)` uses the event `"ace_versioning_clientCheckDone"` instead of a PFH.
- `checkVersionNumber.sqf` now prints `"and x more."` at the end of an error message, to allow players to know how many PBOs they are missing.
- PFHs and `waitUntil` have been replaced with `addPublicVariableEventHandler`.
- `checkVersionNumber.sqf` is now a function.
- ~~Added checking for clients that don't have ACE 3.x.x loaded (version depending on if/when this is added). If a client has an older version, they will be kicked.~~ I was not satisfied by any of the solutions that we came up with, so I dropped it.
The way it checks might need some tweaking (changes are in `XEH_postInit.sqf`).
- Code cleanup.

<details>
<summary>Old description</summary>
- Intended to close #9181 by adding a new argument to `FUNC(errorMessage)`, which allows users to pass codes to both the "Ok" and "Cancel" buttons, without having said buttons being shown.
However, I have not tested if #9181 has been fixed with this, as I do not own a controller.
- `FUNC(errorMessage)`: "ok" and "cancel" codes are no longer global variables ~~, they are passed using `CBA_fnc_addBISEventHandler`~~. This prevents code from being overwritten, however I'm unsure if this is necessary at all.
- ~~PFHs and `waitUntil` used for checking public variables have had their intervals reduced from 1 second to 0.5 seconds. This should make loading into mission be a little snappier.~~
- Added events `"ace_versioning_clientCheckDone"` and `"ace_versioning_serverCheckDone"` for public usage. This would allow other mods to be notified when PBO checking has been completed and whether it has been completed successfully or not. Event names and arguments passed are up for debate: For now, `"ace_versioning_clientCheckDone"` passes a copy of `ACE_Version_ClientErrors`, `"ace_versioning_serverCheckDone"` passes a copy of `ACE_Version_ServerVersions`.
- `FUNC(checkPBOs)` uses the event `"ace_versioning_clientCheckDone"` instead of a PFH. I'm unsure about this change, as if the event is raised externally it could cause some issues - hence the `WIP` tag in the title.
If the event is not considered safe, then I will restructure `FUNC(checkPBOs)`: First I'd execute `checkVersionNumber.sqf` and `waitUntil` it's done, then execute the rest of `FUNC(checkPBOs)`, so as to remove a PFH check - or something else, see in the first point of the questions below.
- ~~`checkVersionNumber.sqf` no longer relies on the global variable `ACE_Version_Whitelist`, the whitelist is passed by `FUNC(checkPBOs)` directly. This is to avoid third party influence. `ACE_Version_ClientVersions` has also been changed for the same reason.~~ no longer valid
- `checkVersionNumber.sqf` now prints `"and x more."` at the end of an error message, to allow players to know how many PBOs they are missing.
- Code cleanup.

[8386a3c](https://github.com/acemod/ACE3/pull/9266/commits/8386a3cf66207b58c21eb5a853793a2988e1ce1d) includes:
- PFHs and `waitUntil` have been replaced with `addPublicVariableEventHandler`.
- `checkVersionNumber.sqf` is now a function.
- Added checking for clients that don't have ACE 3.x.x loaded (version depending on if/when this is added). If a client has an older version, they will be kicked.
The way it checks might need some tweaking (changes are in `XEH_postInit.sqf`).
- `FUNC(errorMessage)` has been adapted to be executable on clients without any mods.
</details>

<details>
<summary>Questions I had:</summary>
- ~~Is there any specific reason why `waitUntil` is used in `checkVersionNumber.sqf` instead of a PFH? `FUNC(checkPBOs)` runs when the event `"CBA_settingsInitialized"` has been raised, so `CBA_fnc_addPerFrameHandler` ought to be available.
Even more elegant imo would be to use a `addPublicVariableEventHandler` on `ACE_Version_ServerVersions` to check when it has been received by the client. Would there be any disadvantages to this?~~ answered
- ~~In `FUNC(checkPBOs)`: What does the `@todo` refer to? The point above?~~ answered
- ~~Currently, max. 11 PBOs are listed in the log if they are missing, wrong version etc. I added the `"and x more."` at the end of an error message. Should this be changed to something that reflects the total amount instead (e.g. `"(_x total)"`)? Or should ACE simply list all of the PBOs in the logs (probably not `systemChat` though)? Other ideas? I feel like the current info could be supplemented to help out people figure out what they need to load/unload.~~
- ~~For when 2.14 rolls out: Should `ACE_Version_ServerVersions` and `ACE_Version_ClientVersions` switch to hashmaps, for the sole purpose of hashmaps being able to be made final using [compileFinal](https://community.bistudio.com/wiki/compileFinal)? Unless - can arrays be made final? The disadvantage of hashmaps is the alphabetical order would be lost, which in this case would be less than ideal imo, so I don't really want to change to them. But at the same time, finalising the variables might outweigh that drawback.~~ answered
</details>
Looking forward to any input.

### IMPORTANT


commy2:
> Is there any specific reason why waitUntil is used in checkVersionNumber.sqf instead of a PFH?

Legacy code / used to run when remote executed without CBA and ACE installed.

> In FUNC(checkPBOs): What does the @todo refer to? The point above?

Yes. Todo: make it a function, not a script.

> Should ACE_Version_ServerVersions and ACE_Version_ClientVersions switch to hashmaps, for the sole purpose of hashmaps being able to be made final using [compileFinal](https://community.bistudio.com/wiki/compileFinal)?

Don't see the point. Also, whack that Dedmen didn't make them insertion ordered.

johnb432:
> Legacy code / used to run when remote executed without CBA and ACE installed.

What does this mean ultimately? Can this be updated or not?

commy2:
> What does this mean ultimately?

Nothing, it's already broken.

> Can this be updated or not?

Sure, but keep in mind that it has to be backwards compatible, meaning, it makes no sense if someone connects with an older version and then nothing shows up because the interface of the PBO checking keeps changing between updates.

PabstMirror:
btw 2.14 will add mod hash to https://community.bistudio.com/wiki/allAddonsInfo
this might be simpler, faster and have the ability to check mods that don't version
and also handle checking mods that re-sign with the same key

johnb432:
> btw 2.14 will add mod hash to https://community.bistudio.com/wiki/allAddonsInfo this might be simpler, faster and have the ability to check mods that don't version and also handle checking mods that re-sign with the same key

Would this new system replace the current one or compliment it?

PabstMirror:
don't need to do anything about it in this PR
just posting info incase it could be useful

addons/common/XEH_postInit.sqf
```diff
@@ -244,6 +244,35 @@ if (_currentVersion != _previousVersion) then {
...
+            params ["", "_uid", "", "", "_owner"];
+
+            // If ACE is not loaded, kick unit
+            if !(missionNamespace getVariable [QGVAR(aceLoaded_) + _uid, false]) then {
+                // Send function to client
+                _owner publicVariableClient QFUNC(errorMessage);
+
+                // Wait for function to broadcast, then kick client
+                [{
+                    ["[ACE] ERROR", "ACE is not present", {findDisplay 46 closeDisplay 0}, {findDisplay 46 closeDisplay 0}, nil, [true, false]] remoteExecCall [QFUNC(errorMessage), _this];
```
LinkIsGrim, line 263: This won't work if ACE isn't loaded on the client, only if it's a version from before this PR.

LinkIsGrim:
>Currently, max. 11 PBOs are listed in the log if they are missing, wrong version etc. I added the "and x more." at the end of an error message. Should this be changed to something that reflects the total amount instead (e.g. "(_x total)")? Or should ACE simply list all of the PBOs in the logs (probably not systemChat though)? Other ideas? I feel like the current info could be supplemented to help out people figure out what they need to load/unload.

All mismatched PBOs should be listed in logs. User error message (and that functionality) was changed in #9560, should update this PR after to see what overlaps/conflicts.

addons/common/XEH_postInit.sqf
```diff
@@ -251,6 +251,35 @@ if (_currentVersion != _previousVersion) then {
...
+// This handles clients joining that do not have ACE loaded
+if (isServer) then {
+    addMissionEventHandler ["PlayerConnected", {
+        [{
+            params ["", "_uid", "", "", "_owner"];
+
+            // If ACE is not loaded, kick unit
+            if !(missionNamespace getVariable [QGVAR(aceLoaded_) + _uid, false]) then {
+                // Send function to client
+                _owner publicVariableClient QFUNC(errorMessage);
```
LinkIsGrim, line 266: Neat.

LinkIsGrim (approved):

LinkIsGrim (changes requested):
I did *not* mean to approve that.

johnb432:
> All mismatched PBOs should be listed in logs. User error message (and that functionality) was changed in #9560, should update this PR after to see what overlaps/conflicts.

I have only addressed the merge conflicts so far, I haven't looked at compatibility between #9560 and this.

addons/common/XEH_postInit.sqf
johnb432: The only problem is that, as you pointed out in your deleted comment, `missionNamespace getVariable [QGVAR(aceLoaded_) + _uid, false]` will not be set by clients who have an outdated ACE.

The easiest fix would be to change the error message on [line 270](https://github.com/acemod/ACE3/pull/9266/files#diff-d241a0f897acd01a4b09998a038b0c1afd584435e7ae84d08c50755d9b914ce1R270) to `"ACE is not present or outdated"`, but if we want to fix it properly, I'm not sure how to go about it.

LinkIsGrim: Add build number to variable name and update the error message?

LinkIsGrim:
Haven't tested yet, but request after https://discord.com/channels/976165959041679380/1166914354051027024: addon mismatch message should show server's [`configSourceMod`](https://community.bistudio.com/wiki/configSourceMod) for `common` or `main`

johnb432:
> Haven't tested yet, but request after https://discord.com/channels/976165959041679380/1166914354051027024: addon mismatch message should show server's [`configSourceMod`](https://community.bistudio.com/wiki/configSourceMod) for `common` or `main`

Can you elaborate further on your suggestion? Do you mean only for ACE (so only in `FUNC(checkFiles)`) or for all mods (`FUNC(checkPBOs)`)? Do you want the server's `configSourceMod` for `main` to show up on the client's error message?

LinkIsGrim:
ACE only. Server's source for `main` displayed on client (should be good enough to just setVariable?)

addons/common/XEH_postInit.sqf
johnb432: > Add build number to variable name and update the error message?

I don't think I was clear with my explanation, as I don't understand your suggestion. By "outdated" I meant a version of ACE that does not support this feature, meaning it does not run `missionNamespace setVariable [QGVAR(aceLoaded_) + getPlayerUID player, true, 2]`.

How would adding the build number help in this specific case? Clients that don't have ACE loaded at all and clients with "outdated" ACE versions are indistinguishable as of the current implementation. I was wondering if there is something built into ACE that we could leverage, something that makes clients send information to the server upon connection.

LinkIsGrim: Just change the error message. We don't need to distinguish being outdated past the point of this PR or not having the mod.

LinkIsGrim:
Remaining question: between source of outdated addons and the `and x more` message, all bases should be covered. Potentially include source dirs for pbos not in the whitelist if pbo checking is enabled for all pbos, but that's about it.

Using the pbo hash from `allAddonsInfo` could be interesting but it needs to take filepatching allowance into account. Best done after this PR anyway

addons/common/XEH_postInit.sqf
```diff
@@ -263,6 +263,35 @@ if (_currentVersion != _previousVersion) then {
...
+missionNamespace setVariable [QGVAR(aceLoaded_) + getPlayerUID player, true, 2];
+
+// This handles clients joining that do not have ACE loaded
+if (isServer) then {
+    addMissionEventHandler ["PlayerConnected", {
+        [{
+            params ["", "_uid", "", "", "_owner"];
+
+            // If ACE is not loaded, kick unit
+            if !(missionNamespace getVariable [QGVAR(aceLoaded_) + _uid, false]) then {
```
LinkIsGrim, line 276: Either variable broadcast is not being done on time or something's wrong with this check. Getting kicked off dedicated MP hosted off my own machine.

LinkIsGrim: Increasing timeout on this WAE to 10 seconds fixed it, but I don't think a magic number exists and we're subject to player/server perfomance and network connection here.

LinkIsGrim: First load, timeout at 9 seconds: Kicked for no/outdated ACE.
Second load, same timeout: got in fine.

This isn't a reliable implementation.

LinkIsGrim:
cf50c78 fixes some undefined vars that caused version checking not to work, and a variable name change for clarity

LinkIsGrim:
I'm not a fan of "server missing addons" (the server is never wrong, IMO) but that's just flavor.

johnb432:
> I'm not a fan of "server missing addons" (the server is never wrong, IMO) but that's just flavor.

Same, but when I first made the PR, I didn't want to change it. However I think I will now.

addons/common/XEH_postInit.sqf
```diff
@@ -263,6 +263,40 @@ if (_currentVersion != _previousVersion) then {
...
+
+        // Wait for function to broadcast, then run function on client
+        [{
+            params ["_owner", "_fncNameCheckAcePresence", "_fncNameCheckErrorMessage"];
+
+            _fncNameCheckErrorMessage remoteExecCall [_fncNameCheckAcePresence, _owner];
+
+            // Delete the functions on the server
+            //missionNamespace setVariable [_fncNameCheckAcePresence, nil];
+            //missionNamespace setVariable [_fncNameCheckErrorMessage, nil];
```
johnb432, line 299: @LinkIsGrim I've completely reworked how the detection system works.

The idea is to make a system that is robust and cannot be hacked into.

What it does now is the server sends 2 functions, each with a unique name, to the connecting client and makes the client run `FUNC(checkAcePresence)`. `FUNC(checkAcePresence)` checks if ACE is properly installed (it can/should be expanded further to its current implementation). If ACE is detected, all is well, but if it isn't, then it runs `FUNC(checkErrorMessage)`. `FUNC(checkErrorMessage)` will spawn an error message that is very hard to remove without it quitting the game.

I've tested the system with ACE and ACE + TFAR on the host (because TFAR has an error message that makes the ACE one vanish, but if TFAR one is clicked away, it still quits the mission), client had no mods loaded.

However, there is a choice we have to make (if this system is adequate; I was thinking maybe we should implement a CBA setting to toggle this system): `FUNC(checkAcePresence)` and `FUNC(checkErrorMessage)` are both compiled with the `isFinal` flag being true, which means that any direct copy is also final. This means when we send the function over the network with `missionNamespace setVariable [_fncNameCheckAcePresence, FUNC(checkAcePresence), _owner];`, it will create a `missionNamespace` variable with the name `_fncNameCheckAcePresence`. This in turn means that every time someone connects, two new functions are created that cannot be deleted. I'm really not a fan of this, but currently I don't see any alternatives.

If a non-final version of the functions is sent, those copies can be deleted on the server. However, it also means that the client can overwrite them. There is a delay of about 1 second (in normal networking cases) on the client between receiving the functions and executing them, meaning there is ample time to scan all mission variables, finding the functions sent by the server and modifying them.

Any thoughts?

On a side note:
I'm getting
`10:46:27 Attempt to delete final function - ace_common_fnc_checkacepresence_953ef70819671adb14c96fce40c64a11f592a80bc5bb69378d5d5eeb1c520dfd` for an unknown reason, I will investigate this.

johnb432:
I've tested in self-hosted, I have not been able to test on a dedicated server.

johnb432:
I was able to test this on a dedicated server:

```
9:23:50 [ACE] johnb43: ERROR client missing addon(s): AR_AdvancedRappelling, AUR_AdvancedUrbanRappelling, SA_AdvancedSlingLoading, OCAP_main, OCAP_extension, OCAP_recorder
9:23:50 [ACE] johnb43: ERROR client additional addon(s): jsrs_soundmod_boats, jsrs_soundmod_bullethits, L_ES_ES_main, L_ES_ES_sounds, L_ES_ES_sys, jsrs_soundmod_cfg_rhs_afrf_weapons, jsrs_soundmod_cfg_rhs_afrf_vehicles, jsrs_soundmod_cfg_rhs_afrf_air_vehicles, jsrs_soundmod_cfg_rhs_usf_weapons, jsrs_soundmod_cfg_rhs_usf_vehicles, jsrs_soundmod_cfg_rhs_usf_air_vehicles, no_actions_main, tao_rewrite_main, turret_lock_main, Revo_EP, AIMEE_main, AIMEE_uav_terminal, AIMEE_vehicle_controls, AIMEE_vehicle_seats, bettinv_main, bettinv_main_ace, tfar_ace_extended_main, AIMEE_change_ammo, AIMEE_group, AIMEE_inventory, zeus_additions_main
```

It works as intended.

PR #9269, opened 2023-07-19
Dragging - Corpse carrying and dragging - continued

johnb432:
**When merged this pull request will:**
- Continuation of #7939.
- Code cleanup.

I had found out a way to implemented corpse carrying, but it was somewhat clunky. However, when trying to implement said solution, I found a better one and wanted to share it.

Pinging those were most heavily involved with the above mentioned PR:
@BaerMitUmlaut @PabstMirror @Dystopian 

I haven't tested this PR a lot, but the initial testing was promising imo.

### IMPORTANT

- [ ] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9271, opened 2023-07-20
Dragging - Code cleanup

johnb432:
- The code cleanup part of #9269.

### IMPORTANT


addons/dragging/functions/fnc_canCarry.sqf
```diff
@@ -22,24 +22,24 @@ if !(alive _target && {_target getVariable [QGVAR(canCarry), false]} && {isNull
...
-//#2644 - Units with injured legs cannot bear the extra weight of carrying an object
-//The fireman carry animation does not slow down for injured legs, so you could carry and run
+// #2644 - Units with injured legs cannot bear the extra weight of carrying an object
+// The fireman carry animation does not slow down for injured legs, so you could carry and run
 if ((_unit getHitPointDamage "HitLegs") >= 0.5) exitWith {false};
 
 // Static weapons need to be empty for carrying (ignore UAV AI)
 if (_target isKindOf "StaticWeapon") exitWith {
-    crew _target findIf {getText (configOf _x >> "simulation") != "UAVPilot"} == -1
+    (crew _target) findIf {getText (configOf _x >> "simulation") != "UAVPilot"} == -1
```
rautamiekka, line 31: I guess this would make sense in readability, in a way, otherwise pointless.

addons/dragging/functions/fnc_canCarry.sqf
```diff
@@ -22,24 +22,24 @@ if !(alive _target && {_target getVariable [QGVAR(canCarry), false]} && {isNull
...
 };
 
 // Units need to be unconscious or limping
 if (_target isKindOf "CAManBase") exitWith {
-    lifeState _target isEqualTo "INCAPACITATED"
-    || {_target getHitPointDamage "HitLegs" >= 0.5}
+    lifeState _target == "INCAPACITATED" ||
```
rautamiekka, line 36: Pointless and [lowers performance](https://community.bistudio.com/wiki/isEqualTo).

johnb432: > Pointless and [lowers performance](https://community.bistudio.com/wiki/isEqualTo).

It's for readability reasons. I'm well aware it performs faster, but I feel in these sorts of cases the readability outweighs the performance loss, especially since there is virtually none here.

addons/dragging/XEH_postInit.sqf
LinkIsGrim, line 1: Can you move keybinds to "initKeybinds.sqf"?

addons/dragging/XEH_postInit.sqf
```diff
@@ -15,67 +15,23 @@ if (isNil "ACE_maxWeightDrag") then {
...
-["ace_unconscious", {_this call FUNC(handleUnconscious)}] call CBA_fnc_addEventHandler;
+// Handle waking up dragged unit and falling unconscious while dragging
+["ace_unconscious", LINKFUNC(handleUnconscious)] call CBA_fnc_addEventHandler;
 
-// display event handler
-["MouseZChanged", {_this select 1 call FUNC(handleScrollWheel)}] call CBA_fnc_addDisplayHandler;
+// Display event handler
+["MouseZChanged", {(_this select 1) call FUNC(handleScrollWheel)}] call CBA_fnc_addDisplayHandler;
 
 //@todo Captivity?
```
johnb432, line 37: Does anyone know what this refers to?

addons/dragging/initKeybinds.sqf
```diff
@@ -0,0 +1,58 @@
+// Add Keybinds
+["ACE3 Common", QGVAR(drag), LLSTRING(DragKeybind), {
+    private _player = ACE_player;
+
+    if (!alive _player) exitWith {false};
+    if !([_player, objNull, ["isNotDragging", "isNotCarrying"]] call EFUNC(common,canInteractWith)) exitWith {false};
```
johnb432, line 6: As far as I can tell, you are meant to be able to drag while swimming.
I've never quite understood the logic of exclusions of `EFUNC(common,canInteractWith)` - Does this account for being able to drag while swimming?

addons/dragging/XEH_postInit.sqf
LinkIsGrim: I'm gonna guess it's handling for being captured while carrying/dragging an object.

addons/dragging/XEH_preInit.sqf
```diff
@@ -7,5 +7,6 @@ PREP_RECOMPILE_START;
 PREP_RECOMPILE_END;
 
 #include "initSettings.sqf"
+#include "initKeybinds.sqf"
```
LinkIsGrim, line 10: Keybinds should be in postInit.

addons/dragging/functions/fnc_canCarry.sqf
```diff
@@ -22,24 +22,24 @@ if !(alive _target && {_target getVariable [QGVAR(canCarry), false]} && {isNull
...
+    {_target getHitPointDamage "HitLegs" >= 0.5}
 };
 
 // Check max items for WeaponHolders
 if (["WeaponHolder", "WeaponHolderSimulated"] findIf {_target isKindOf _x} != -1) exitWith {
     (count (weaponCargo _target + magazineCargo _target + itemCargo _target)) <= MAX_DRAGGED_ITEMS
 };
 
-true // return
+true // Return
```
LinkIsGrim, line 45: We use `// return`, in lowercase, everywhere else in the project.

addons/dragging/functions/fnc_canDrag.sqf
```diff
@@ -24,18 +24,18 @@ if !([_unit, _target, ["isNotSwimming"]] call EFUNC(common,canInteractWith)) exi
...
+    {_target getHitPointDamage "HitLegs" >= 0.5}}
 };
 
 // Check max items for WeaponHolders
 if (["WeaponHolder", "WeaponHolderSimulated"] findIf {_target isKindOf _x} != -1) exitWith {
     (count (weaponCargo _target + magazineCargo _target + itemCargo _target)) <= MAX_DRAGGED_ITEMS
 };
 
-true // return
+true // Return
```
LinkIsGrim, line 41: Ditto.

addons/dragging/functions/fnc_carryObject.sqf
```diff
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: commy2
- * Carry an object.
+ * Carries an object.
```
LinkIsGrim, line 4:
```suggestion
 * Handles attaching and setting up a carried object.
```

Or something better.

addons/dragging/functions/fnc_dragObject.sqf
```diff
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 /*
  * Author: commy2, Malbryn
- * Drag an object. Called from ace_dragging_fnc_startDrag
+ * Drags an object. Called from ace_dragging_fnc_startDrag
```
LinkIsGrim, line 4: Should match `FUNC(carryObject)`.

addons/dragging/functions/fnc_getWeight.sqf
```diff
@@ -33,4 +33,4 @@ _weight = _weight + getNumber (configOf _object >> "mass");
 
 // Mass in Arma isn't an exact amount but rather a volume/weight value
 // This attempts to work around that by making it a usable value (sort of)
-_weight * 0.5;
+_weight * 0.5
```
LinkIsGrim, line 36:
```suggestion
_weight * 0.5 // return
```

addons/dragging/initKeybinds.sqf
```diff
@@ -0,0 +1,58 @@
+// Add Keybinds
+["ACE3 Common", QGVAR(drag), LLSTRING(DragKeybind), {
+    private _player = ACE_player;
+
+    if (!alive _player) exitWith {false};
```
LinkIsGrim, line 5:
```suggestion
    if !(alive _player) exitWith {false};
```

addons/dragging/initKeybinds.sqf
```diff
@@ -0,0 +1,58 @@
...
+
+    [_player, _cursorObject] call FUNC(startDrag);
+
+    false
+}, {}, [-1, [false, false, false]]] call CBA_fnc_addKeybind; // UNBOUND
+
+["ACE3 Common", QGVAR(carry), LLSTRING(CarryKeybind), {
+    private _player = ACE_player;
+
+    if (!alive _player) exitWith {false};
```
LinkIsGrim, line 34:
```suggestion
    if !(alive _player) exitWith {false};
```

LinkIsGrim (commented):
This will have merge conflicts with #9225

addons/dragging/initKeybinds.sqf
LinkIsGrim:
```suggestion
    if !([_player, objNull, ["isNotDragging", "isNotCarrying", "isNotSwimming"]] call EFUNC(common,canInteractWith)) exitWith {false};
```

addons/dragging/initKeybinds.sqf
BrettMayson: This is the opposite of the ACE Coding Guidelines (although I know both are used in the code)

https://ace3.acemod.org/wiki/development/coding-guidelines.html#55-brackets-around-code

johnb432: My rule of thumb is: If it's one statement that is easily readable (such as the above), I add it inside the brackets.

But then we have lines such as [these](https://github.com/acemod/ACE3/blob/52ed0fc6be5669521dd4acdd756e6afd18e4fe6d/addons/dragging/functions/fnc_canCarry.sqf#L21). I'm not sure how to interpret the ACE guidelines here - are these fine or should these types of statements be corrected?

LinkIsGrim (approved):
I'm not of fan of `==/!=` instead of `isEqualTo/isNotEqualTo` in conditions and PFHs, though.

addons/dragging/functions/fnc_handleAnimChanged.sqf
```diff
@@ -1,23 +1,21 @@
...
  * None
  *
  * Example:
- * [_unit, "amovpercmstpsnonwnondnon"] call ace_dragging_fnc_handleAnimChanged;
+ * [player, "amovpercmstpsnonwnondnon"] call ace_dragging_fnc_handleAnimChanged;
  *
  * Public: No
 */
 
-//IGNORE_PRIVATE_WARNING ["_thisArgs", "_thisID"]; // From CBA_fnc_addBISEventHandler;
```
PabstMirror, line 19: sqf_linter will use these hints to reduce false positives
tool hasn't been updated in a long time so it's not that useful
but going forward there really isn't any reason to remove these lines

johnb432: Should I add them back?

johnb432:
> I'm not of fan of `==/!=` instead of `isEqualTo/isNotEqualTo` in conditions and PFHs, though.

The reason is readability for me.
`isEqualTo` might be less prone to throwing errors, but it will fail just as much as `==` if the comparison contains a variable that is `nil`, which, I'd argue, is the most common type of problem.

rautamiekka:
> if the comparison contains a variable that is `nil`

That's why `isNil`, documentation & testing, exists. It ain't `iET`'s fault.

johnb432:
> It ain't `iET`'s fault.

Never claimed it to be...?
I was merely pointing out that both operators/commands were limited in the same way when it comes to variables that are `nil`.

PabstMirror:
probably won't be an issue, but iet also handles nulls a little differently
```
_unit = objNull;
(vehicle _unit) == _unit; // false
(vehicle _unit) isEqualTo _unit; // true
```

LinkIsGrim:
> > I'm not of fan of `==/!=` instead of `isEqualTo/isNotEqualTo` in conditions and PFHs, though.
> 
> The reason is readability for me. `isEqualTo` might be less prone to throwing errors, but it will fail just as much as `==` if the comparison contains a variable that is `nil`, which, I'd argue, is the most common type of problem.
> 
> I can change it though, it's not a problem.

I'm just a fan of saving nanoseconds in things that run every frame.

johnb432:
> I'm just a fan of saving nanoseconds in things that run every frame.

Understandable.

Any further opinions on the matter? `isEqualTo` seems to be favored atm, will change to that in a couple of days if nobody disagrees.

jonpas:
`isEqualTo` is so minisculy faster it is pointless to discuss. Just use what is more appropriate in this case. Comparing numbers / same types, `==` is completely fine.

LinkIsGrim:
`==` it is!

PR #9138, opened 2023-01-29
Dragging - Improve carrying by allowing carried object to be loaded into vehicle by key or mouse

ampersand38:
**When merged this pull request will:**
- allowing carried object to be loaded into vehicle by key or mouse
- carrying mouse hint updates to show possibility of loading
- dragging behaviour is not modified due to need to to move backwards and thus not immediately looking toward target vehicle
- https://youtu.be/hVRH48DeybQ

### IMPORTANT

- [x] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9272, opened 2023-07-20
Dragging - Fix unintended Cargo dependency and improve ragdoll/disconnection handling

johnb432:
- The improvement part of https://github.com/acemod/ACE3/pull/9269, plius a third one (last in the list below).

1. Makes `ace_dragging` depend on `ace_cargo`. There are functions and stringtable entries from `ace_cargo` that are called, see [here](https://github.com/acemod/ACE3/blob/aead9a8eb4949714ddcb5d7b07b57f0c798a4ea0/addons/dragging/functions/fnc_dropObject_carry.sqf#L105), [here](https://github.com/acemod/ACE3/blob/aead9a8eb4949714ddcb5d7b07b57f0c798a4ea0/addons/dragging/functions/fnc_carryObjectPFH.sqf#L61) and [here](https://github.com/acemod/ACE3/blob/aead9a8eb4949714ddcb5d7b07b57f0c798a4ea0/addons/dragging/functions/fnc_carryObjectPFH.sqf#L65). If `ace_dragging` isn't supposed to be dependent on `ace_cargo`, then `ace_dragging` will need to be adjusted accordingly.
2. Disconnection detection for units carrying or dragging was done previously with the `HandleDisconnect` mission EH. In my testing (self-hosted MP) `HandleDisconnect` fires too late and when trying to read the `QGVAR(isCarrying)/QGVAR(isDragging)` variables, it just returns `false`, regardless if a unit if carrying/dragging something or not. This makes it impossible for carried/dragged objects to be dropped properly.
`PlayerDisconnected` fires before and therefore makes it possible to drop carried/dragged objects correctly. However, I'm unsure if there are any disadvantages to `PlayerDisconnected`.
3. Added a check to `FUNC(canDrag)` and `FUNC(canCarry)`, where it disables dragging/carrying if a unit is in ragdoll animation. This is to prevent odd behaviour. As it is now, you can try to drag a unit in ragdoll, but they stay in place while you go into the drag animation.

### IMPORTANT

### DIFF
--- a/addons/common/stringtable.xml
+++ b/addons/common/stringtable.xml
@@ -1661,5 +1661,21 @@
             <Russian>Разрядить оружие</Russian>
             <Spanish>Descargar arma</Spanish>
         </Key>
+        <Key ID="STR_ACE_common_loadObject">
+            <English>Load</English>
+            <German>Beladen</German>
+            <Polish>Załaduj</Polish>
+            <Portuguese>Carregar</Portuguese>
+            <Russian>Загрузить</Russian>
+            <Czech>Naložit</Czech>
+            <Spanish>Cargar</Spanish>
+            <Italian>Carica</Italian>
+            <French>Charger</French>
+            <Japanese>積み込む</Japanese>
+            <Korean>싣기</Korean>
+            <Chinese>裝載</Chinese>
+            <Chinesesimp>装载</Chinesesimp>
+            <Turkish>Yükle</Turkish>
+        </Key>
     </Package>
 </Project>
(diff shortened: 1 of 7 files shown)


LinkIsGrim:
> If ace_dragging isn't supposed to be dependent on ace_cargo, then ace_dragging will need to be adjusted accordingly.

This wasn't the case [until recently](#9138). Should be adjusted using conditions. Rest is fine.

LinkIsGrim (approved):

johnb432:
Nice adjustments and additions @LinkIsGrim !

PR #9273, opened 2023-07-20
Dragging - Corpse carrying and dragging - continued

johnb432:
- Changes related to corpse carrying from #9269.
- Continuation of https://github.com/acemod/ACE3/pull/7939.

### IMPORTANT


addons/dragging/functions/fnc_createClone.sqf
```diff
@@ -0,0 +1,41 @@
...
+ * [player] call ace_dragging_fnc_createClone;
+ *
+ * Public: No
+ */
+params ["_target"];
+
+private _clone = QGVAR(clone) createVehicle [0, 0, 0];
+
+// Clone loadout
+_clone setUnitLoadout getUnitLoadout _target;
```
PabstMirror, line 22: I think this will cause problems with acre or other systems expecting unique ids

johnb432: > I think this will cause problems with acre or other systems expecting unique ids

Filter them out using `acre_api_fnc_filterUnitLoadout` and whatever TFAR has respectively wouldn't suffice?

PabstMirror:
encountered some weird invisibility and body duplication
https://www.youtube.com/watch?v=gvnRnQPhjf0

addons/dragging/functions/fnc_createClone.sqf
jonpas: Just use CBA's `CBA_fnc_getLoadout`.

johnb432:
Changelog of [b32ad48](https://github.com/acemod/ACE3/pull/9273/commits/b32ad48f825edb2f8938b825502679e6ace78a10):
- Added actions to dead units. I do not know of any downsides of using `allDeadMen` to get all dead units, so I went with that.
- Added some delays for clone creation and deletion, in order to get their corpses to be synced better. On my end the duplication of bodies is gone, but I haven't tested on a dedicated server.
- Corpse is removed from garbage collector, so that it can be kept. If it's deleted through other manners, the unit will just drop the clone (as opposed to before, when it kept the clone on their back).

LinkIsGrim (approved):
No issues observed when testing with a small player count (8 players). Code LGTM.

We'll never get full scale testing done, banana principle it is IMO.

johnb432:
> No issues observed when testing with a small player count (8 players). Code LGTM.
> 
> We'll never get full scale testing done, banana principle it is IMO.

I integrated this in my zeus mod a while back and so far I haven't received any complaints (doesn't mean much though, as sometimes broken things aren't reported at all). However, when testing it, I noticed that moving the corpses as little as possible was key to avoiding them being desynced (duh).
In the zeus mod, the corpses are hidden, but they remain above terrain level - which means simulation doesn't have to be touched. It also means that the corpse only has to be moved once.

I'll push a commit with the suggested changes, to show what I mean.

johnb432:
> I'll push a commit with the suggested changes, to show what I mean.

See https://github.com/acemod/ACE3/pull/9273/commits/d1755af35be1ffb5d49a27527b26d15ab0d1ba79.

LinkIsGrim:
Damage info is lost upon clone creation, which means injury textures disappear.

Keeping wounds isn't really necessary, as Medical can't be accessed while the clone is being carried, but the transition can be a bit jarring.

johnb432:
> Keeping the body aboveground and just hiding it means you can still access its inventory via the dropped weapon if there is one.

Reverting that part then.

LinkIsGrim (approved):

addons/dragging/functions/fnc_dropClone.sqf
```diff
@@ -0,0 +1,77 @@
...
+        _pos = _pos vectorAdd [0, 0, 0.05];
+    };
+
+    // Make sure position is not underground
+    if (_pos select 2 < 0.05) then {
+        _pos set [2, 0.05];
+    };
+
+    // Move the unit where it is local
+    [QGVAR(moveCorpse), [_target, getDir _unit + 180, ], _target] call CBA_fnc_targetEvent;
```
LinkIsGrim, line 43:
```suggestion
    [QGVAR(moveCorpse), [_target, getDir _unit + 180, _pos], _target] call CBA_fnc_targetEvent;
```

addons/dragging/functions/fnc_createClone.sqf
```diff
@@ -0,0 +1,101 @@
...
+
+    // Clone loadout (sometimes default loadouts are randomised, so overwrite those)
+    [_clone, _target call CBA_fnc_getLoadout] call CBA_fnc_setLoadout;
+
+    // Sets the facial expression
+    [[QGVAR(cloneCreated), [_target, _clone]] call CBA_fnc_globalEventJIP, _clone] call CBA_fnc_removeGlobalEventJIP;
+
+    // Release claim on corpse
+    [objNull, _target] call EFUNC(common,claim);
+}, [_clone, _target, _posATL], 0.25] call CBA_fnc_waitAndExecute;
```
LinkIsGrim, line 88: Can't this just be moved out of the WAE now? The delay on the loadout being applied to the clone is a bit jarring.

addons/dragging/functions/fnc_createClone.sqf
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * Public: No
+ */
+
+params ["_unit", "_target"];
+
+private _posATL = getPosASL _target;
+
+// Create clone
+private _clone = createVehicle [QGVAR(clone), ASLToAGL _posATL, [], 0, "CAN_COLLIDE"];
```
LinkIsGrim, line 24:
```suggestion
private _posATL = getPosATL _target;

// Create clone
private _clone = createVehicle [QGVAR(clone), _posATL, [], 0, "CAN_COLLIDE"];
```

According to wiki AGL is only relevant for amphibious vehicles.

addons/dragging/functions/fnc_startCarryLocal.sqf
```diff
@@ -34,6 +34,11 @@ if (_weight > GETMVAR(ACE_maxWeightCarry,1E11)) exitWith {
     [LLSTRING(UnableToDrag)] call EFUNC(common,displayTextStructured);
 };
 
+// Create clone for dead units
+if (!alive _target) then {
+    _target = [_unit, _target] call FUNC(createClone);
+};
+
```
LinkIsGrim, line 41: move below line 45

addons/dragging/functions/fnc_createClone.sqf
johnb432: I'm not sure if a small delay is needed for synchronisation, but I guess we'll see.

addons/dragging/XEH_postInit.sqf
```diff
@@ -54,6 +59,29 @@ if (isNil QGVAR(maxWeightCarryRun)) then {
...
+    // Set direction before position
+    _corpse setDir _dir;
+
+    // Bring corpse back to clone's position
+    _corpse setPosATL _pos;
+
+    // Sync the corpse
+    [QEGVAR(common,awake), [_corpse, true]] call CBA_fnc_globalEvent;
+    [QEGVAR(common,awake), [_corpse, false]] call CBA_fnc_globalEvent;
+    [QEGVAR(common,awake), [_corpse, true]] call CBA_fnc_globalEvent;
```
johnb432, line 82: This last is not necessary, but it makes the corpse ragdoll, which can be both good and bad: It looks better, but it might cause collisions.

LinkIsGrim:
According to @BrettMayson
>and yeah the carrying is still a little too buggy
at least a third of the time they snap back after being dropped

I think we'll have to concede defeat to the engine on this one.

johnb432:
> According to @BrettMayson
> 
> > and yeah the carrying is still a little too buggy
> > at least a third of the time they snap back after being dropped
> 
> I think we'll have to concede defeat to the engine on this one.

If it's the carrying that's causing the issue, then I'd say we drop that (pun not intended), but keep dragging - unless both are problematic. If so, then I agree and we'll need to put this PR down.

BrettMayson:
I didn't mean the carrying is buggy while dragging is not, both have the problem

Drofseh:
Could we add it but disabled by default,with a totip warning saying it's buggy?

rautamiekka:
> Could we add it but disabled by default,with a totip warning saying it's buggy?

Would be sufficient if the feature itself otherwise works, better than nothing that way.

jonpas:
If it is 1 out of 3 times bugged as @BrettMayson put it, then it should not be in core ACE3. It can be an optional component like `ace_particles` though.

Drofseh:
> If it is 1 out of 3 times bugged as @BrettMayson put it, then it should not be in core ACE3. It can be an optional component like `ace_particles` though.

having it the optionals would be ok with me!

BrettMayson:
It would be neat if we could just get someone to look at this on the engine side and see if it could be implemented...

When it does work, we really love it

LinkIsGrim:
Back to ongoing this goes

johnb432:
@BrettMayson Can you this test again? I tested this version on a dedicated server and found it worked.

Dystopian:
I did some tests on dedicated and didn't see any bugs.

johnb432:
> I did some tests on dedicated and didn't see any bugs.

I actually found some weird behaviour, which I forgot to report here:
When JIP, it seems to act weirdly when you try to walk over bodies, as if the body was standing up or had some collision at least (when it usually doesn't). I had enhanced movement enabled, as well as many other mods, so idk how that affects that.
It's not game breaking, but it's annoying.

johnb432:
> I actually found some weird behaviour, which I forgot to report here: When JIP, it seems to act weirdly when you try to walk over bodies, as if the body was standing up or had some collision at least (when it usually doesn't). I had enhanced movement enabled, as well as many other mods, so idk how that affects that. It's not game breaking, but it's annoying.

I've managed to address that issue with the latest commit. I tested it on a dedicated server and everything seemed to work fine.

addons/dragging/XEH_postInit.sqf
```diff
@@ -1,9 +1,72 @@
...
+
+    // Check if the corpse is already close to the target
+    // If so, don't teleport
+    if !(
+        (_currentPos select 0 <= _xPos + 0.25) &&
+        {_currentPos select 0 >= _xPos - 0.25} &&
+        {_currentPos select 1 <= _yPos + 0.25} &&
+        {_currentPos select 1 >= _yPos - 0.25} &&
+        {_currentPos select 2 <= _zPos + 0.25} &&
+        {_currentPos select 2 >= _zPos - 0.25}
```
jokoho48, line 20: ever heard of distance?

```suggestion
_currentPos distance _posATL > 0.25
```

johnb432: It's not the same, but there's no specific reason why the volume to be checked should be a cube and not a sphere - it's just the first thing I thought of.

docs/wiki/framework/events-framework.md
```diff
@@ -167,6 +167,13 @@ MenuType: 0 = Interaction, 1 = Self Interaction
 | `ace_headless_groupTransferPre` | [_group, _HC (OBJECT), _previousOwner, _idHC] | Target | Listen | Called just before a group is transferred from any machine to a HC. Called where group currently is local and on the HC, where group is going to be local.
 | `ace_headless_groupTransferPost` | [_group, _HC (OBJECT), _previousOwner, _idHC, _transferredSuccessfully] | Target | Listen | Called just after a group is transferred from a machine to a HC. Called where group was local and on the HC, where group is now local. `_transferredSuccessfully` is passed so mods can actually check if the locality was properly transferred, as ownership transfer is not guaranteed.
 
+### 2.18 Interaction (`ace_dragging`)
```
johnb432, line 170:
```suggestion
### 2.18 Dragging (`ace_dragging`)
```

addons/dragging/functions/fnc_canCarry.sqf
```diff
@@ -18,26 +18,30 @@
...
 // #2644 - Units with injured legs cannot bear the extra weight of carrying an object
 // The fireman carry animation does not slow down for injured legs, so you could carry and run
 if ((_unit getHitPointDamage "HitLegs") >= 0.5) exitWith {false};
 
+// Units need to be unconscious or limping; Units also need to not be in ragdoll if alive, as that causes desync issues
+if (_isPerson) exitWith {
+    !_alive ||
+    {(isAwake _target) && // not ragdolled if alive
+    {!(lifeState _target in ["HEALTHY", "INJURED"]) ||
+    {_target getHitPointDamage "HitLegs" >= 0.5}}}
```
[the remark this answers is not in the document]
johnb432: Imo should be done in another PR, then we could update all code to reflect that.

addons/dragging/functions/fnc_canCarry.sqf
```diff
@@ -18,26 +18,30 @@
...
 
 // #2644 - Units with injured legs cannot bear the extra weight of carrying an object
 // The fireman carry animation does not slow down for injured legs, so you could carry and run
 if ((_unit getHitPointDamage "HitLegs") >= 0.5) exitWith {false};
 
+// Units need to be unconscious or limping; Units also need to not be in ragdoll if alive, as that causes desync issues
+if (_isPerson) exitWith {
+    !_alive ||
+    {(isAwake _target) && // not ragdolled if alive
+    {!(lifeState _target in ["HEALTHY", "INJURED"]) ||
```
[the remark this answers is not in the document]
johnb432: Won't hurt, but I'm thinking if I can simplify it more.

addons/medical_status/functions/fnc_addInventoryActions.sqf
```diff
@@ -46,5 +46,6 @@ _unit addAction ["OpenBag", {
     {!lockedInventory _backpackContainer} &&
     {maxLoad _backpackContainer > 0} &&
     {getNumber (_backpackConfig >> "disableInventory") != 1} &&
+    {(typeOf _target) != QEGVAR(dragging,clone)} &&
```
PabstMirror, line 49: While dragging, you could access the backpack of the clone
action was being added by this because

`alive x = true`
`lifeState x = "DEAD"`

addons/medical_gui/functions/fnc_displayPatientInformation.sqf
```diff
@@ -40,6 +40,7 @@ if (isNull _display) then {
         if (ACE_player distance _target > MAX_DISTANCE && {vehicle _target != vehicle ACE_player}) exitWith {
             [_pfhID] call CBA_fnc_removePerFrameHandler;
             QGVAR(RscPatientInfo) cutFadeOut 0.3;
+            if (((getPosATL _target) # 2) < -9) exitWith {}; // handle dragging corpse/clone
```
PabstMirror, line 43: this just fixes the little popup that "target has moved"
it's kinda ugly but works...

addons/medical_status/functions/fnc_addInventoryActions.sqf
johnb432: Would `isKindOf` instead of `typeOf` make more sense? Or is it not worth it?

PabstMirror:
```suggestion
    {!(_target isKindOf QEGVAR(dragging,clone)} &&
```
this might be better if we add other mimics?
but I think typeOf is a tiny bit faster

PabstMirror:
super-rare but clones can drown - (`getOxygenRemaining` drops)
but I don't think it's a big deal, u really have to try to get them into this situation
image

addons/dragging/functions/fnc_canCarry.sqf
```diff
@@ -18,26 +18,30 @@
...
 
 if !([_unit, _target, []] call EFUNC(common,canInteractWith)) exitWith {false};
 
 // #2644 - Units with injured legs cannot bear the extra weight of carrying an object
 // The fireman carry animation does not slow down for injured legs, so you could carry and run
 if ((_unit getHitPointDamage "HitLegs") >= 0.5) exitWith {false};
 
+// Units need to be unconscious or limping; Units also need to not be in ragdoll if alive, as that causes desync issues
+if (_isPerson) exitWith {
+    !_alive ||
```
PabstMirror, line 34:
```suggestion
    ((!_alive) && {missionNamespace getVariable [QGVAR(canMoveDead), true]}) ||
```

addons/dragging/functions/fnc_canDrag.sqf
```diff
@@ -18,22 +18,26 @@
...
+private _alive = alive _target;
+private _isPerson = _target isKindOf "CAManBase";
+
+if !((_alive || _isPerson) && {_target getVariable [QGVAR(canDrag), false]} && {isNull objectParent _target}) exitWith {false};
 
 if !([_unit, _target, ["isNotSwimming"]] call EFUNC(common,canInteractWith)) exitWith {false};
 
+// Units need to be unconscious or limping; Units also need to not be in ragdoll if alive, as that causes desync issues
+if (_isPerson) exitWith {
+    !_alive ||
```
PabstMirror, line 30:
```suggestion
    ((!_alive) && {missionNamespace getVariable [QGVAR(canMoveDead), true]}) ||
```

PabstMirror:
Everything seems to work great in mp
I could see this possibly causing problems with some kind of mission scripts/triggers because the real body doesn't move
and acre radios on int-audio won't update their location as the body is moved

nothing that I think is serious, but would be nice to have a setVar just in-case people need it

PabstMirror (approved):

johnb432:
> super-rare but clones can drown - (`getOxygenRemaining` drops) but I don't think it's a big deal, u really have to try to get them into this situation image

Mostly solved by refilling the oxygen from time to time (<0.5 seems to be lethal with the medical system). If you manually set the oxygen to 0, it's possible to have the unit die.

PabstMirror (approved):

PR #9277, opened 2023-07-21
Tools - fix toml file read

PabstMirror:
can't just re-read() a file in python without moving cursor back or something
change to just read once


jonpas (approved):
Well yeah, why would you haha.

PR #9278, opened 2023-07-21
Arsenal - Make Zeuses able to save, delete, and edit Default Loadouts

LinkIsGrim:
- Title.
- Add `global` parameter to `FUNC(addDefaultLoadout)`.
- Switch to using `FUNC(addDefaultLoadout)` internally.
- Add `FUNC(removeDefaultLoadout)` and `FUNC(renameDefaultLoadout)`. What it says on the tin.
- Add documentation for above.

Loadout index in the local loadout save events remains the same, but that whole behavior's weird with default loadouts.
Loadout save/delete events aren't fired remotely in MP.
Doesn't have the same handling as public loadouts. Should it? I don't think there's much issue with the list not being updated until the player changes tabs. Default loadouts shouldn't be getting updated that often.

### IMPORTANT


LinkIsGrim:
Tested in SP and Eden, no idea if MP'll work (it should?).

PabstMirror (approved):
tested in local mp

PR #9283, opened 2023-07-22
Medical Engine - Fix unloading dead units in vehicles and handle deleted units vehicles

PabstMirror:
ref #7959
- fix unloading dead units
- fix position setting when unloading (non-dead units will now unload near you)
- lock seat when mounted unit dies
- unlock seat when unit is deleted


LinkIsGrim:
`FUNC(unloadPersonLocal)` needs to be switched to `objectParent` as well, that's why it's failing.

addons/common/functions/fnc_unloadPersonLocal.sqf
```diff
@@ -47,15 +47,19 @@ unassignVehicle _unit;
...
+
 if (local _vehicle) then {
     _unit action ["Eject", _vehicle];
     // Failsafe - sometimes eject alone doesn't work, but moveOut does
     [{
         params ["_unit"];
 
-        if (vehicle _unit != _unit) then {
-            WARNING_1("UnloadPersonLocal [%1] did not eject normally",_unit);
+        if (objectParent _unit != _unit) then {
```
LinkIsGrim, line 57:
```suggestion
        if (!isNull objectParent _unit) then {
```

addons/common/functions/fnc_unloadPersonLocal.sqf
```diff
@@ -66,15 +70,15 @@ if (local _vehicle) then {
 
 [{
     params ["_unit", "_emptyPos"];
-    (alive _unit) && {(vehicle _unit) != _unit}
+    (alive _unit) && {(objectParent _unit) != _unit}
```
LinkIsGrim, line 73:
```suggestion
    (alive _unit) && {!isNull objectParent _unit}
```

I don't really understand what this block is supposed to do, though: alive and in a vehicle as condition?

addons/medical_engine/XEH_postInit.sqf
```diff
@@ -94,3 +94,8 @@
         };
     };
 }] call CBA_fnc_addEventHandler;
+
+["CAManBase", "Killed", {
+    params ["_unit"];
+    [_unit] call FUNC(unlockUnconsciousSeat);
```
LinkIsGrim, line 100: This kinda defeats the point of #7959 IMO. Better to get the unloading working at 100%.

LinkIsGrim:
https://github.com/LinkIsGrim/ACE3/commit/179957c39d76850ac6021358302d59da340b18c5 should handle everything.

addons/common/functions/fnc_unloadPersonLocal.sqf
PabstMirror: I think it was designed to wait until the unit actually got out of the vehicle and then move them to the empty pos
but the logic was inverted and it's never done anything

addons/medical_engine/XEH_postInit.sqf
```diff
@@ -94,3 +94,19 @@
         };
     };
 }] call CBA_fnc_addEventHandler;
+
+["ace_killed", { // global event
+    params ["_unit"];
+    TRACE_3("unit Killed",_unit,objectParent _unit,local _unit);
+    if (!isNull objectParent _unit && {local objectParent _unit}) exitWith {
```
LinkIsGrim, line 101: We might be able to switch this to just
```suggestion
    if (local objectParent _unit) exitWith {
```
as `local objNull` seems to always be false, but it's a minuscule gain and less readable IMO.

addons/medical_engine/XEH_postInit.sqf
```diff
@@ -85,7 +85,7 @@
 
 ["ace_unconscious", {
     params ["_unit", "_unconscious"];
-
+    TRACE_3("unit uncon",_unit,objectParent _unit,local _unit);
     if (vehicle _unit != _unit && {local vehicle _unit}) then {
```
LinkIsGrim, line 89:
```suggestion
    if (!isNull objectParent _unit && {local objectParent _unit}) then {
```

LinkIsGrim (approved):

PR #9290, opened 2023-07-26
Arsenal - Improve stats display

LinkIsGrim:
- Remove early sorting of arsenal stats into groups of 5. Stats will be checked directly and the five with highest priority will be displayed.
- Set current actions page to `nil` when closing the Arsenal. (Drive-by change)
- Hide stats box completely when nothing is selected or there are no valid stats for the item (unlikely, but could happen).
- Do some code cleanup for hiding unused stat lines.

Currently, if more than 5 stats are present for a tab but an item doesn't meet display conditions for all of them, page controls will be enabled, and stats may not be shown despite still having space on the first page. This PR addresses that.

Stats box outline would still be displayed even if no item was selected or no valid stats were found. Now it's hidden.

Performance impact is negligible.

### IMPORTANT


addons/arsenal/functions/fnc_handleStats.sqf
```diff
@@ -43,222 +44,217 @@ private _hideUnusedFnc = {
...
-            } forEach [
-                _statsPreviousPageCtrl,
-                _statsNextPageCtrl,
-                _statsCurrentPageCtrl
-            ];
+    // If there are no stats to show (unlikely), just hide everything
+    if (_statsTab isEqualTo []) exitWith {
+        call _fnc_hideEverything
+    };
+
```
PabstMirror, line 101:
```suggestion

   GVAR(currentStatPage) = GVAR(currentStatPage) min floor ((count _statsTab) / 5); 
```
e.g. if you are on page 2 and go to a new item with only 3 stats total, it will jump back to a page with something on it

LinkIsGrim: This should really jump back to page 0 on item changes. Whoops.

PabstMirror (approved):

PR #9295, opened 2023-07-28
Common - Make latitude fallback to CfgWorlds value if no specific data is available

LinkIsGrim:
- Title.
- Add a minor optimization in `weather_fnc_getMapData`.

### IMPORTANT

### DIFF
--- a/addons/common/functions/fnc_getMapData.sqf
+++ b/addons/common/functions/fnc_getMapData.sqf
@@ -129,4 +129,4 @@ if (_map in ["xcam_prototype"]) exitWith { [35, 0] };
 if (_map in ["xcam_taunus"]) exitWith { [50, 0] };
 if (_map in ["zargabad"]) exitWith { [35, 2000] };
 
-[] //Return empty array if we have no specific data for the map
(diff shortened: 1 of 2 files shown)


addons/common/functions/fnc_getMapData.sqf
```diff
@@ -129,4 +129,4 @@ if (_map in ["xcam_prototype"]) exitWith { [35, 0] };
 if (_map in ["xcam_taunus"]) exitWith { [50, 0] };
 if (_map in ["zargabad"]) exitWith { [35, 2000] };
 
-[] //Return empty array if we have no specific data for the map
+[getNumber (configFile >> "CfgWorlds" >> _map >> "latitude"), 0] // Use terrain config if we have no data for this map
```
PabstMirror, line 132: I think `[]` is fine to return here

https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getMGRSdata.sqf#L22-L33
is the only thing that uses it and it handles empty array already
+besides a debug function to test which maps are missing entries in the look up

we at least want negative lat and `elevationOffset` here to be constant

LinkIsGrim:
No need for this. (Though some maps don't respect their config latitude despite not having specific data, need to investigate.)

PR #9015, opened 2022-08-27
Extensions: Rust

BrettMayson:
**When merged this pull request will:**
- Merge all extensions into one
- It's now in Rust though
- Use fancy `callExtension` features instead of `:` deliminated strings

- [x] advanced_ballistics
- [ ] artillerytables
- [ ] break_line
- [x] Clipboard
- [ ] fcs

- [ ] log from Rust to RPT

Medical and dynload are no longer needed

PR #9298, opened 2023-07-28
Artillery Tables - Add additional extension hooks for shot simulation and solution calculation

LorenLuke:
- Title.

To be used for a dedicated artillery computer.

### IMPORTANT


addons/artillerytables/functions/fnc_calculateElevation.sqf
```diff
@@ -0,0 +1,66 @@
...
+    private _shotResults = [_useAngle, _targetHeight, _muzzleVelocity, _airFriction, _crossWind, _tailWind, _temperature, _airDensity] call FUNC(simulateShot);
+    
+    _xDeviation = _shotResults select 0;
+    _resultDistance = _shotResults select 1;
+    _tof = _shotResults select 2;
+    _useDistance = (2 * _targetDistance) - _resultDistance;
+};
+
+private _angleOffsetDeg = _xDeviation atan2 _resultDistance;
+private _angleOffset = _angleOffsetDeg * 6400 / 360;
```
LinkIsGrim, line 64: I'd like this calculation (x * 17.777...) to be defined as a macro somwhere, similar to the MOA conversion we have in `ballistics`.

addons/artillerytables/functions/fnc_adjustFire.sqf
```diff
@@ -0,0 +1,41 @@
+#include "script_component.hpp"
+/*
+ * Author: LorenLuke
+ * Adjusts a target point north and east, and recalculates a solution in air based on atmospheric conditions
+ *
+ * Arguments:
+ * 0: Own Position ASL; 
+ * 1: Target Position ASL
+ * 2: adjustment to the East (negative is west)
```
LinkIsGrim, line 9: Please capitalize these everywhere. Adjustment should also include units in argument description.

addons/artillerytables/functions/fnc_adjustFire.sqf
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Public: No
+ */
+
+params ["_ownPos", "_targetPos", "_adjustEast", "_adjustNorth", "_adjustUp", "_muzzleVelocity", "_airFriction", ["_highAngle", true], ["_temperature", 15], ["_airDensity", 1.225], ["_windDir", 0], ["_windSpeed", 0]];
+
+//DEFAULT_AIR_FRICTION == -0.00006
+//MK6_82mm_AIR_FRICTION == -0.0001
+
+private _resultPos = [_adjustEast + _targetPos select 0, _adjustNorth + _targetPos select 1, _adjustUp + _targetPos select 2];
```
LinkIsGrim, line 37: Is this not being used?

addons/artillerytables/functions/fnc_adjustFire.sqf
```diff
@@ -0,0 +1,41 @@
...
+
+private _returns = ["_ownPos", "_targetPos", "_muzzleVelocity", "_highAngle", "_airFriction", "_temperature", "_airDensity", "_windDir", "_windSpeed"] call FUNC(calculateSolution);
+
+_returns;
```
LinkIsGrim, line 41:
```suggestion
_returns // return
```
`// return` comment is unecessary, but we don't use `;` for return values.

addons/artillerytables/functions/fnc_adjustFire.sqf
```diff
@@ -0,0 +1,41 @@
...
+ * Example:
+ * [myPos, targetPos, 20, 50, 0, 200, -0.0001, true, 15, 1.225, 225, 5] call ace_artilleryTables_fnc_adjustFire
+ *
+ * Public: No
+ */
+
+params ["_ownPos", "_targetPos", "_adjustEast", "_adjustNorth", "_adjustUp", "_muzzleVelocity", "_airFriction", ["_highAngle", true], ["_temperature", 15], ["_airDensity", 1.225], ["_windDir", 0], ["_windSpeed", 0]];
+
+//DEFAULT_AIR_FRICTION == -0.00006
+//MK6_82mm_AIR_FRICTION == -0.0001
```
LinkIsGrim, line 35: If this is intended to be a user-facing function then `_airFriction` should be handled by the function as well, IMO.

addons/artillerytables/functions/fnc_adjustFire.sqf
```diff
@@ -0,0 +1,41 @@
...
+ * 1: angle adjust left or right - milliradians
+ * 2: time of flight
+ *
+ * Example:
+ * [myPos, targetPos, 20, 50, 0, 200, -0.0001, true, 15, 1.225, 225, 5] call ace_artilleryTables_fnc_adjustFire
+ *
+ * Public: No
+ */
+
+params ["_ownPos", "_targetPos", "_adjustEast", "_adjustNorth", "_adjustUp", "_muzzleVelocity", "_airFriction", ["_highAngle", true], ["_temperature", 15], ["_airDensity", 1.225], ["_windDir", 0], ["_windSpeed", 0]];
```
LinkIsGrim, line 32: `_ownPos` would be better served as parameter for the vehicle firing, IMO.

addons/artillerytables/functions/fnc_adjustFire.sqf
```diff
@@ -0,0 +1,41 @@
...
+ * 11: speed of wind in m/s
+ *
+ * Return Value:
+ * array of returns <ARRAY>
+ * 0: angle of shot - milliradians
+ * 1: angle adjust left or right - milliradians
+ * 2: time of flight
+ *
+ * Example:
+ * [myPos, targetPos, 20, 50, 0, 200, -0.0001, true, 15, 1.225, 225, 5] call ace_artilleryTables_fnc_adjustFire
```
LinkIsGrim, line 27:
```suggestion
 * [getPosASL vehicle player, targetPos, 20, 50, 0, 200, -0.0001, true, 15, 1.225, 225, 5] call ace_artilleryTables_fnc_adjustFire
```

LinkIsGrim (changes requested):

addons/artillerytables/functions/fnc_adjustFire.sqf
LorenLuke: The problem with that is that it has to identify it as a mortar versus as an artillery shell, and I feel like that might fall into the weird 'special numbers' case, as it were, if any other mod were to use anything different.

(If you recall the demonstration with the shells falling short of predicted solutions, that was because it automatically added DEFAULT_AIR_FRICTION when I was using the mortar; I'd like to avoid that if possible just by making it enterable, and having some other function determine which value to use, as they were just included here for reference in testing).

addons/artillerytables/functions/fnc_adjustFire.sqf
LorenLuke: It isn't and it should be. Meant to replace `_targetPos` in line 39.

addons/artillerytables/functions/fnc_calculateElevation.sqf
LorenLuke: Already is; though that macro will be used here in next commit.

https://github.com/acemod/ACE3/blob/eb53e7a2fe089a3ddb11a2a54b57af285ce99734/addons/artillerytables/script_component.hpp#L15

LorenLuke: (Will also replace line 53 of this file, too.)

LinkIsGrim (approved):

PabstMirror:
still planning on 3.16 just didn't want it in RC because extension problems with battleeye

jonpas:
Do in #9015 and merge the Rust extensions after 3.16.0 release.

LorenLuke:
It's now 3.16.2... is this not happening?

LinkIsGrim:
It'll happen with #9015

johnb432:
#9015 has been merged, therefore closing. Thanks @LorenLuke!

PR #9303, opened 2023-08-01
Tools - document_functions.py: handle header sub indexs for arguments

PabstMirror:
changes header to 
```
 * Arguments:
 * 0: Tabs to add sort to <ARRAY>
 * - 0: Left Tab Indexes <ARRAY of NUMBERS>
```
we could also just keep
`  * 1.0: Mouse area control <CONTROL>`

### DIFF
--- a/addons/common/functions/fnc_watchVariable.sqf
+++ b/addons/common/functions/fnc_watchVariable.sqf
@@ -8,11 +8,11 @@
  * 1: Code to generate result (passed nothing, can return any) <CODE> (default: {})
  * 2: Array containing modifiers <ARRAY> (default: [])
  *   For Numbers:
- *     2.0: Show Delta change <BOOL> (default: true)
- *     2.1: Slider Min Value <NUMBER> (default: 0)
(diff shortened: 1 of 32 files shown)


addons/common/functions/fnc_watchVariable.sqf
```diff
@@ -8,11 +8,11 @@
...
  *   For Numbers:
- *     2.0: Show Delta change <BOOL> (default: true)
- *     2.1: Slider Min Value <NUMBER> (default: 0)
- *     2.1: Slider Max Value <NUMBER> (default: 0)
+ * - 0: Show Delta change <BOOL> (default: true)
+ * - 1: Slider Min Value <NUMBER> (default: 0)
+ * - 2: Slider Max Value <NUMBER> (default: 0)
  *   For Anything else:
- *     2.0: Number of structured text lines <NUMBER> (default: 1)
+ * + 0: Number of structured text lines <NUMBER> (default: 1)
```
PabstMirror, line 15: `+` so it doesn't get counted is kind of a hack, but it works for this one unusual case

jonpas (approved):

PR #9307, opened 2023-08-05
Arsenal - Favorites System

LinkIsGrim:
- Add a favorites system to the arsenal.
- Make `FUNC(refresh)` public, add parameter to skip updating current/unique items.
- Fix tooltips for the menu bar not showing unless a button or the menu bar area was clicked last (hovering over the buttons will kill search bar focus, though).

Default mode can be controlled by CBA Setting, so can the highlight color. I'd go for icons instead but we'd have to either overwrite mod icons if an item is favorited or make left panel a listnbox as well (and it would be cluttered on the right panel unless its width was increased).

EDIT: Now that this works, I've tried out using icons. The ones available in vanilla look ugly as sin, and clutter is still a problem, particularly with magazine names.

Favorites can be added/removed by double clicking an item while holding left shift and are saved to profileNamespace.
Still works with sorting/searching.
Current items (including inventory) are also displayed in favorites, without the highlight color.

Favorites are always saved in lowercase to prevent them being affected by configCase changes (mod pack changes or updates).

### IMPORTANT


addons/arsenal/defines.hpp
```diff
@@ -19,6 +19,9 @@
 #define ASCENDING 0
 #define DESCENDING 1
 
+// Favorites
+#define FAVORITES_COLOR (GVAR(favoritesColor) + [1])
```
LinkIsGrim, line 23: Since alpha is modified for the right panel (in case the item doesn't fit), I opted for this solution instead, but it's a bit dirty.

addons/arsenal/initSettings.sqf
```diff
@@ -33,26 +33,47 @@ private _category = LLSTRING(settingCategory);
     true
 ] call CBA_fnc_addSetting;
 
+[
+    QGVAR(defaultToFavorites),
+    "CHECKBOX",
+    [LSTRING(defaultToFavoritesSetting), LSTRING(defaultToFavoritesTooltip)],
+    _category,
+    false,
+    -2 // never overwrite the client
```
johnb432, line 42: See https://github.com/CBATeam/CBA_A3/wiki/CBA-Settings-System#create-a-custom-setting-for-mission-or-mod:

```suggestion
    2 // never overwrite the client
```

addons/arsenal/initSettings.sqf
```diff
@@ -33,26 +33,47 @@ private _category = LLSTRING(settingCategory);
...
+] call CBA_fnc_addSetting;
+
+[
+    QGVAR(favoritesColor),
+    "COLOR",
+    [LSTRING(favoritesColorSetting), LSTRING(favoritesColorTooltip)],
+    _category,
+    [0.9, 0.875, 0.6],
+    -2 // never overwrite the client
```
johnb432, line 51: Same here

addons/arsenal/initSettings.sqf
LinkIsGrim: Derp.

addons/arsenal/functions/fnc_updateRightPanel.sqf
```diff
@@ -22,15 +22,18 @@ private _rightPanelCache = uiNamespace getVariable [QGVAR(rightPanelCache), crea
...
+private _alpha = 1;
 
 // Grey out items that are too big to fit in remaining space of the container
 for "_row" from 0 to (lnbSize _control select 0) - 1 do {
     _mass = _rightPanelCache getOrDefault [_control lnbData [_row, 0], 0];
+    _color = _control lnbColor [_row, 1];
 
     // Lower alpha on color for items that can't fit
-    _color = [1, 1, 1, [0.25, 1] select (_mass <= _loadRemaining)];
+    private _alpha = [0.25, 1] select (_mass <= _loadRemaining);
```
johnb432, line 33:
```suggestion
    _alpha = [0.25, 1] select (_mass <= _loadRemaining);
```

LinkIsGrim:
That should be all the bugs handled (besides the search bar losing focus if you hover over a button in the menu bar).

PabstMirror (approved):

addons/arsenal/functions/fnc_onArsenalOpen.sqf
```diff
@@ -95,6 +95,9 @@ GVAR(currentActionPage) = 0;
 // Update current item list
 call FUNC(updateCurrentItemsList);
 
+[_display, _display displayCtrl IDC_buttonFavorites] call FUNC(buttonFavorites);
```
LinkIsGrim, line 98:
```suggestion
// Setup favorites button text and switch to default mode defined by setting
[_display, _display displayCtrl IDC_buttonFavorites] call FUNC(buttonFavorites);
```

BrettMayson (approved):

Issue #8927, opened 2022-05-31
Custom wound handlers cannot be added when given function nil during pre-init


diwako:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.08 Profiling/Performance branch`
- **CBA:** `3.15.7`
- **ACE3:** `3.15.0`

**Description:**
Custom wound handlers from a custom mod will not be added when given function name does not yet exist. Mostly happens if CBA prep is used for the custom mod and the medical damage module is in its `CfgPatches >> requiredAddons`

`ace_medical_damage_parseConfigForInjuries` runs on preinit and on mission loaded, however it appears that the prepped function for the custom wound handler is not yet defined.

**Steps to reproduce:**
_Follow [https://ace3.acemod.org/img/wiki/user/issue_flowchart.webp](this flowchart)!_

1. Create custom mod with wound handler defined
2. Add medical damage module as requiredAddon
3. Prep function in custom mod in pre start and pre init
4. Handler is not added to wound type

**Expected behavior:**
Wound handler is added to the wound type cache, even if the function does not yet exist as it will exist after pre init.

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

**Additional context:**
example config:
```cpp
class CfgPatches {
    class ADDON {
        requiredAddons[] = {"ace_medical_damage"};
    };
};
class ACE_Medical_Injuries {
    class damageTypes {
        class woundHandlers;
        class falling {
            class woundHandlers: woundHandlers {
                ADDON = QFUNC(handleFallDamage);
            };
        };
    };
};
```

It however works when wrapping the handler in a dummy function like so
```cpp
ADDON = QUOTE({ call FUNC(handleFallDamage) });
```

PR #9310, opened 2023-08-06
Medical Damage - Improve custom wound handling

johnb432:
- Addresses #8927 by wrapping entry in `call _fnc` block if `_fnc` does not exist when config is parsed.
I'm not sure how well this new way can account for code (see example below), but I'm not sure if this system was ever meant for handling code directly.

```cpp
class ACE_Medical_Injuries {
    class damageTypes {
        class woundHandlers;
        class bullet {
            class woundHandlers: woundHandlers {
                ADDON = QUOTE(systemChat str _damageData; _damageData);
            };
        };
    };
};
```

- If a wound handler returns an invalid return, it will be ignored. However it will no longer stop checking other wound handlers; it will continue on until it has gone through all wound handlers.
- Invalid wound handler return now includes `nil`.
- Minor code cleanup.

### IMPORTANT

### DIFF
--- a/addons/medical_damage/functions/fnc_parseWoundHandlersCfg.sqf
+++ b/addons/medical_damage/functions/fnc_parseWoundHandlersCfg.sqf
@@ -10,24 +10,35 @@
  * None
  *
  * Example:
- * [configFile >> "ace_medical_injuries" >> "damageTypes"] call ace_medical_damage_fnc_parseWoundHandlersCfg
+ * [configFile >> "ace_medical_injuries" >> "damageTypes" >> "woundHandlers"] call ace_medical_damage_fnc_parseWoundHandlersCfg
  *
  * Public: No
  */
 params ["_config"];
 
-// read all valid entries from config and store
+// Read all valid entries from config and store
 private _entries = [];
+private _entryResult = "";
+
 {
-    private _entryResult = call compile getText _x;
-    if !(isNil "_entryResult") then {
-        _entries pushBack _entryResult;
-    }
+    _entryResult = getText _x;
+
+    if (_entryResult != "") then {
+        _function = call compile _entryResult;
+
+        // If function doesn't exist yet, wrap in extra layer
+        if (isNil "_function") then {
+            _function = compile format ["call %1", _entryResult];
+        };
+
+        _entries pushBack _function;
+    };
 } forEach configProperties [_config, "isText _x", false];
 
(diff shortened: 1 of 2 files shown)


LinkIsGrim:
> However it will no longer stop checking other wound handlers; it will continue on until it has gone through all wound handlers.

I'm not sure about this. I don't think many mods have modified wound handling yet, but I still don't think it's a safe change.

johnb432:
> > However it will no longer stop checking other wound handlers; it will continue on until it has gone through all wound handlers.
> 
> I'm not sure about this. I don't think many mods have modified wound handling yet, but I still don't think it's a safe change.

I don't understand why it was designed this way in the first place: Why should the evaluation of all wound handlers be stopped if a single one fails?
A third party mod might easily return something invalid, which would prevent another 3rd party mod from having their wound handler from running.

In fact, as it is now it's safer imo: It checks `isNil`, which it didn't before, leading to possibly unpredictable behaviour.

LinkIsGrim:
@pterolatypus your code. Thoughts?

pterolatypus:
Appreciate I'm very late to this, been out of the game for a long time, but for the sake of information:
Having at least some way to stop handling was intentional, so that a custom handler can tell the system "this event has been dealt with" and not duplicate handling. Think of the return as "damage still left to be handled".
An example of this is `fnc_woundsHandlerBurning` which deals with the incoming small damage ticks and takes responsibility for passing them on later. Because it "handles" all of the damage, there is no damage left afterwards.
I suppose you could return something like `[_unit, [], _damageType]` which would have a similar effect, but I feel like the intention of returning `[]` is pretty clear and the current implementation lets us just skip subsequent handlers that probably wouldn't have done anything anyway.

If you're worried about someone's bad code causing a problem, you could make it only terminate when it gets exactly `[]` and continue in any other invalid case. Personally, though, I'd just tell the offending modder to read the docs and fix their code.

As to the bug fix: the current implementation does handle arbitrary expressions but I doubt anyone uses it as they're only actually called at PreInit, only the return value is stored. If you wrap it like that, the whole expression would get called on every wound event. I think that's fine, but you should probably just wrap it in every case so that it's consistent.

johnb432:
https://ace3.acemod.org/wiki/framework/medical-framework#44-wound-handler-function says that `[]` does indeed make it stop, so I've removed that change. I'm not going to change anything further, apart from considering `nil` as an invalid return.

PR #9313, opened 2023-08-08
AI - Add automatic NVG equip/unequip

jonpas:
- Adds functionality (disabled by default) where AI will get NVG automatically assigned during night time and unassigned during day time.

### IMPORTANT

---

Better naming ideas welcome.

### DIFF
--- a/addons/ai/XEH_postInit.sqf
+++ b/addons/ai/XEH_postInit.sqf
@@ -68,3 +68,8 @@
     params ["_unit", "_mode"];
     _unit enableGunLights _mode;
 }] call CBA_fnc_addEventHandler;
+
+
+if (isServer) then {
+    call FUNC(assignNVG);
+};
--- a/addons/ai/XEH_postInit.sqf
+++ b/addons/ai/XEH_postInit.sqf
@@ -71,5 +71,10 @@
 
 
 if (isServer) then {
-    ["CAManBase", "init", FUNC(assignNVG)] call CBA_fnc_addClassEventHandler;
+    ["CAManBase", "init", {
+        // wait for HMD to be assigned so `hmd _unit` works
+        [{
+            _this call FUNC(assignNVG);
+        }, _this, 1] call CBA_fnc_waitAndExecute;
+    }] call CBA_fnc_addClassEventHandler;
 };
--- a/addons/ai/XEH_postInit.sqf
+++ b/addons/ai/XEH_postInit.sqf
@@ -71,5 +71,5 @@
 
 
 if (isServer) then {
-    ["CAManBase", "init", FUNC(assignNVG), true, [], true] call CBA_fnc_addClassEventHandler;
+    ["CAManBase", "init", FUNC(assignNVG)] call CBA_fnc_addClassEventHandler;
 };
--- a/addons/ai/XEH_postInit.sqf
+++ b/addons/ai/XEH_postInit.sqf
@@ -71,5 +71,5 @@
 
 
 if (isServer) then {
-    call FUNC(assignNVG);
+    ["CAManBase", "init", FUNC(assignNVG), true, [], true] call CBA_fnc_addClassEventHandler;
 };
--- a/addons/ai/XEH_postInit.sqf
+++ b/addons/ai/XEH_postInit.sqf
@@ -73,8 +73,6 @@
 if (isServer) then {
     ["CAManBase", "init", {
         // wait for HMD to be assigned so `hmd _unit` works
-        [{
-            _this call FUNC(assignNVG);
-        }, _this, 1] call CBA_fnc_waitAndExecute;
+        [FUNC(assignNVG), _this, 1] call CBA_fnc_waitAndExecute;
     }] call CBA_fnc_addClassEventHandler;
 };
--- a/addons/ai/XEH_preInit.sqf
+++ b/addons/ai/XEH_preInit.sqf
@@ -6,4 +6,6 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+#include "initSettings.sqf"
+
 ADDON = true;
--- a/addons/ai/XEH_preInit.sqf
+++ b/addons/ai/XEH_preInit.sqf
@@ -6,11 +6,11 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
-#include "initSettings.sqf"
-
 if (isServer) then {
-    GVAR(assignNVGpfh) = -1;
+    GVAR(assignNVGthread) = false;
     GVAR(assignNVGstate) = false;
 };
+
+#include "initSettings.sqf"
 
 ADDON = true;
--- a/addons/ai/XEH_preInit.sqf
+++ b/addons/ai/XEH_preInit.sqf
@@ -8,4 +8,9 @@ PREP_RECOMPILE_END;
 
 #include "initSettings.sqf"
 
+if (isServer) then {
+    GVAR(assignNVGpfh) = -1;
+    GVAR(assignNVGstate) = false;
+};
+
 ADDON = true;
--- a/addons/ai/functions/fnc_assignNVG.sqf
+++ b/addons/ai/functions/fnc_assignNVG.sqf
@@ -2,7 +2,6 @@
 /*
  * Author: Jonpas
  * Assigns AI first found NVG in their inventory during night time and unassigns it during day time.
- * Calls itself periodically.
  *
  * Arguments:
  * None
@@ -18,28 +17,23 @@
 
 if (!GVAR(assignNVG)) exitWith {};
 
-private _equip = sunOrMoon < 1 || moonIntensity > 0.8;
+params ["_unit"];
 
-{
-    private _unit = _x;
-    if (alive _unit && !isPlayer _unit) then {
-        private _nvg = hmd _unit;
+if (alive _unit && !isPlayer _unit) then {
+    private _nvg = hmd _unit;
 
-        if (_equip) then {
-            if (_nvg == "") then {
-                private _items = [_unit, false, true, true, true, false, false] call CBA_fnc_uniqueUnitItems; // backpack, vest, uniform
-                {
-                    if (getText (configFile >> "CfgWeapons" >> _x >> "simulation") == "NVGoggles") exitWith {
-                        _unit assignItem _x;
-                    };
-                } forEach _items;
-            };
-        } else {
-            if (_nvg != "" && {currentVisionMode _unit == 0} && {_unit canAdd _nvg}) then {
-                _unit unassignItem _nvg;
-            };
+    if (GVAR(assignNVGstate)) then {
+        if (_nvg == "") then {
+            private _items = [_unit, false, true, true, true, false, false] call CBA_fnc_uniqueUnitItems; // backpack, vest, uniform
+            {
+                if (getText (configFile >> "CfgWeapons" >> _x >> "simulation") == "NVGoggles") exitWith {
+                    _unit assignItem _x;
+                };
+            } forEach _items;
+        };
+    } else {
+        if (_nvg != "" && {currentVisionMode _unit == 0} && {_unit canAdd _nvg}) then {
+            _unit unassignItem _nvg;
         };
     };
-} forEach allUnits;
-
-[FUNC(assignNVG), 300] call CBA_fnc_waitAndExecute;
+};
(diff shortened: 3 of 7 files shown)


addons/ai/functions/fnc_assignNVG.sqf
```diff
@@ -0,0 +1,45 @@
...
+ * Public: No
+ */
+
+if (!GVAR(assignNVG)) exitWith {};
+
+private _equip = sunOrMoon < 1 || moonIntensity > 0.8;
+
+{
+    private _unit = _x;
+    if (!isNull _unit && {!isPlayer _unit} && {alive _unit}) then {
```
veteran29, line 25: `alive objNull isEqualTo false`

Also not possible to get null here if it's unscheduled.
Not sure if that lazy eval is not slower.
```suggestion
    if (alive _unit && {!isPlayer _unit}) then {
```

veteran29 (changes requested):

addons/ai/initSettings.sqf
```diff
@@ -0,0 +1,10 @@
+private _category = format ["ACE %1", localize LSTRING(DisplayName)];
+
+[
+    QGVAR(assignNVG), "CHECKBOX",
+    [LSTRING(AssignNVG_DisplayName), LSTRING(AssignNVG_Description)],
+    _category,
+    false,
+    1,
+    FUNC(assignNVG)
```
veteran29, line 9: 1. Start the mission with setting enabled
2. Disabled the setting, enable the setting (before 300s passed)
3. Two loops will run

veteran29 (changes requested):

addons/ai/initSettings.sqf
jonpas: Indeed. To be fair, should probably just require a restart for this setting. Runtime changes don't really make sense, as we don't revert to whatever it was originally anyways. Thoughts?

johnb432: Why not use a PFH with a delay of 300s? That way you could save the ID and check if it exists or not. It would allow for a CBA setting that can be changed mid mission.

On another note: Having this functionality run on unit creation would be nice imo, that way you wouldn't have to wait up to 5 mins to get an AI to have its NVGs equipped.

On yet another note: If you assign NVGs to a unit, do the AI use them automatically?

addons/ai/initSettings.sqf
```diff
@@ -0,0 +1,10 @@
+private _category = format ["ACE %1", localize LSTRING(DisplayName)];
```
johnb432, line 1:
```suggestion
private _category = format ["ACE %1", LLSTRING(DisplayName)];
```

addons/ai/initSettings.sqf
jonpas: PFH would work yes.

First run is done without a delay, could also run on new unit creation once.

AI uses them automatically if it's assigned yes. But they don't take them automatically from their uniform, which is what this PR does. It's more a cosmetic thing so you don't constantly have AI with NVGs running around, especially prevalent with vanilla factions.

addons/ai/functions/fnc_assignNVG.sqf
```diff
@@ -0,0 +1,39 @@
+#include "script_component.hpp"
+/*
+ * Author: Jonpas
+ * Assigns AI first found NVG in their inventory during night time and unassigns it during day time.
+ *
+ * Arguments:
+ * None
```
Dystopian, line 7:
```suggestion
 * 0: Unit <OBJECT>
```

addons/ai/functions/fnc_assignNVG.sqf
```diff
@@ -0,0 +1,39 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call ace_ai_fnc_assignNVG
+ *
+ * Public: No
+ */
+
+if (!GVAR(assignNVG)) exitWith {};
```
Dystopian, line 18:
```suggestion
```
I think this check can be safely omitted.

addons/ai/functions/fnc_assignNVGpfh.sqf
```diff
@@ -0,0 +1,24 @@
...
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [] call ace_ai_fnc_assignNVGpfh
+ *
+ * Public: No
+ */
+
+if (!GVAR(assignNVG)) exitWith {};
```
Dystopian, line 18:
```suggestion
```
this too

Dystopian (changes requested):

addons/ai/functions/fnc_assignNVG.sqf
jonpas: No, it's used to prevent any changes from class EH (unit spawning).

addons/ai/XEH_postInit.sqf
```diff
@@ -68,3 +68,8 @@
     params ["_unit", "_mode"];
     _unit enableGunLights _mode;
 }] call CBA_fnc_addEventHandler;
+
+
+if (isServer) then {
+    ["CAManBase", "init", FUNC(assignNVG), true, [], true] call CBA_fnc_addClassEventHandler;
```
jonpas, line 74: First PFH run is immediate, can ignore retroactive application.

addons/ai/initSettings.sqf
```diff
@@ -0,0 +1,19 @@
...
+[
+    QGVAR(assignNVG), "CHECKBOX",
+    [LSTRING(AssignNVG_DisplayName), LSTRING(AssignNVG_Description)],
+    _category,
+    false,
+    1,
+    {
+        if (isServer) then {
+            params ["_enabled"];
+            if (_enabled) then {
```
johnb432, line 12: Personally I would add a check that prevents 2 PFHs running at the same time and trying to remove a non-existent PFH.
I'm not sure how necessary those checks would be though.

jonpas: Doesn't hurt indeed.

Mike-MF (approved):
Tested with preplaced and Zeus spawned AI, works perfectly.

acknowledge14:
How about a option to only add low tech(generation 1) NVG?

Drofseh:
Could make the setting be a number
0 - no assignment
1 - assign Gen 1
2 - assign Gen 2
3 etc.
4 etc.

addons/ai/XEH_postInit.sqf
```diff
@@ -68,3 +68,11 @@
     params ["_unit", "_mode"];
     _unit enableGunLights _mode;
 }] call CBA_fnc_addEventHandler;
+
+
+if (isServer) then {
+    ["CAManBase", "init", {
```
johnb432, line 74: Would using `"postInit"` instead of `"init"` make a difference when it comes to `hmd _unit` returning `""`?

jonpas: No, has been tested.

jonpas:
> How about a option to only add low tech(generation 1) NVG?


> Could make the setting be a number 0 - no assignment 1 - assign Gen 1 2 - assign Gen 2 3 etc. 4 etc.
> 
> Maybe do -1 for thermal goggles or something? idk

This feature does not add any NVGs at all. It merely puts them into the inventory during the day and puts them back on the head during the night. It is a visual feature. If the unit has a Gen 1 in loadouty it'll use that, if it has a Gen 2 it'll use that...

addons/ai/XEH_preInit.sqf
```diff
@@ -6,4 +6,11 @@ PREP_RECOMPILE_START;
 #include "XEH_PREP.hpp"
 PREP_RECOMPILE_END;
 
+#include "initSettings.sqf"
+
+if (isServer) then {
+    GVAR(assignNVGpfh) = -1;
+    GVAR(assignNVGstate) = false;
+};
```
johnb432, line 14: Could the CBA setting be set and run before it gets to here? If yes, `GVAR(assignNVGpfh)` would be overwritten.

jonpas: I don't think so, this is all unscheduled, so adding the setting and initializing this variable should happen right one after another. And as far as I know, settings initialize in `postInit`.

addons/ai/functions/fnc_assignNVG.sqf
```diff
@@ -0,0 +1,39 @@
+#include "script_component.hpp"
+/*
+ * Author: Jonpas
+ * Assigns AI first found NVG in their inventory during night time and unassigns it during day time.
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+
```
Mike-MF, line 8:
```suggestion
 *
```

jonpas: Interesting!

PabstMirror:
swapped pfeh to CBA_fnc_waitAndExecute which should be a tiny bit faster
(no time code checked each frame, just the first element in the ordered array)

PabstMirror (approved):

LinkIsGrim (approved):

LinkIsGrim:
> Better naming ideas welcome.

`AI - Add setting to automatically unequip NVGs during daytime`?

johnb432:
> `AI - Add setting to automatically unequip NVGs during daytime`?

`AI - Add setting to automatically equip/unequip NVGs depending on time of day`?

It's a bit lengthy though...

jonpas:
I meant more the naming of settings and description (strings in-game).

addons/ai/stringtable.xml
```diff
@@ -65,5 +81,11 @@
             <Czech>Nenalezena žádná budova.</Czech>
             <Turkish>Bir yapı bulunamadı</Turkish>
         </Key>
+        <Key ID="STR_ACE_AI_AssignNVG_DisplayName">
+            <English>Assign NVG</English>
```
johnb432, line 85: The only clarification I can think of:

```suggestion
            <English>Assign NVG automatically</English>
```

Drofseh: There's a difference in meaning between the plain English word assign (to allocate a thing to an entity) and the Arma assign command (to put a specific item in a valid slot).

To me, having asign in the name, while techinically correct from an Arma perspective, sounds like it's going to give NVGs to units that don't have them (which is the reason for my mistaken comment higher up), when it's actually just going to move them from inventory to NVG slot.

I think the only truely unambiguous English words might be don (put on [an item of clothing]) and doff (remove [an item of clothing]).

`AI Don and Doff NVGs Automatically` doesn't have a great ring to it though...

jonpas: Yeah, I had the same issue. Maybe adding "Does NOT give NVGs to AI!" to the description would help?

LinkIsGrim: 'Equip', maybe?

rautamiekka: > 'Equip', maybe?

Beat me to it by 5mins. +1.

jonpas: Equip can also be understood as both.

LinkIsGrim: 'Put on'

johnb432: 'Make AI wear/take off NVGs automatically'

addons/ai/stringtable.xml
```diff
@@ -65,5 +81,11 @@
...
+        </Key>
+        <Key ID="STR_ACE_AI_AssignNVG_Description">
+            <English>Assigns NVG in AI's inventory during night time and unassigns it during day time.</English>
```
LinkIsGrim, line 88:
```suggestion
            <English>Equips NVG in AI's inventory during night time and takes it off during day time.\nDoes not add NVGs to AI loadouts.</English>
```

addons/ai/stringtable.xml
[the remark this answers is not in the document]
jonpas: Isn't that too long for a setting title?

`AI` in front is unnecessary, category is already that.

Alternatives:
- `Put on and Take off NVGs automatically`
- `Automatically Put-on / Take-off NVG`
- `Automatically Wear NVG`
- `Auto-NVG`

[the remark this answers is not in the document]
LinkIsGrim: It's a checkbox, I don't think we need to worry about length. Updated suggestion.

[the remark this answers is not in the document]
LinkIsGrim: Either way naming is the only thing holding this up. We can come up with something better later if it's too confusing.

jonpas:
:+1: I am happy with this naming now, thanks @LinkIsGrim.

PR #9316, opened 2023-08-10
Arsenal - Improve performance of loadout verification

LinkIsGrim:
- Title.
- Remove amount of unavailable/nullItems, list is checked directly instead.

Switched to single iteration through the loadout array. See previous optimizations below.
<details><summary>Old</summary>
<p>
Optimization 1: Add items to nullItemsList along with configCase changes, since we know that `configName _x == ""` means the class doesn't exist anywhere.

Optimization 2: Replace null items with empty string directly when changing loadout contents to configCase. This removes the need to check for `isClass _x` on the second iteration through the loadout.

Optimization 3: Check all classnames against `GVAR(virtualItemsFlatAll)` instead of categories, since hashmap lookup time is constant. This allowed for removing the index checking in `_fnc_weaponCheck` and others.

Optimization 4: Change use of `pushBackUnique` to combo `pushBack` and `arrayIntersect` as that is faster in most cases now, particularly with large arrays.

Optimization 5: Switch use of count of unavailable/null items for checking the lists against a default value in relevant places.

Optimization 6: Replaced some loop logic (assignedItems now uses forEach, and container item checking was changed to use the `_containerItems` array already defined in `_x params`).

Optimization 7: Due to all of the above, config lookups and some variable assignments could be removed.

To make this work, the unavailable items list needs to be checked against [""]. Otherwise array modification or explicit checking for `_x != ""` would be needed for everything, which would reduce performance gain.
</p>
</details> 

End result, loading just CBA and ACE and checking against a full arsenal, is a 13% improvement:

```

testLoadout = [[["rhs_weap_m4a1_blockII_KAC_bk","rhsusf_acc_nt4_black","","rhsusf_acc_su230",["rhs_mag_30Rnd_556x45_Mk262_Stanag",30],[],"rhsusf_acc_grip2"],["rhs_weap_M136_hp","","","",[],[],""],["rhsusf_weap_glock17g4","","","",["rhsusf_mag_17Rnd_9x19_JHP",17],[],""],["VSM_Multicam_Crye_SS_od_shirt_Camo",[["ACE_CableTie",2],["ACE_EarPlugs",2],["ACE_EntrenchingTool",1],["ACE_epinephrine",3],["ACE_RangeCard",1],["ACE_splint",4],["ACE_tourniquet",4],["ACE_M26_Clacker",1],["ACE_MapTools",1],["ACE_Flashlight_XL50",1],["ACE_packingBandage",15],["ACE_IR_Strobe_Item",1],["ACE_morphine",3]]],["VSM_FAPC_Breacher_OGA_OD",[["rhs_mag_30Rnd_556x45_Mk262_Stanag",8,30],["HandGrenade",2,1],["ACE_M84",4,1],["rhs_mag_an_m14_th3",2,1],["SmokeShell",4,1],["SmokeShellBlue",1,1],["SmokeShellGreen",1,1],["SmokeShellRed",1,1],["rhsusf_mag_17Rnd_9x19_JHP",4,17]]],["VSM_OGA_OD_Backpack_Compact",[["DemoCharge_Remote_Mag",2,1]]],"VSM_OD_spray_OPS","VSM_Shemagh_Balaclava2_OD_Peltor_Goggles",["Laserdesignator_01_khk_F","","","",["Laserbatteries",1],[],""],["ItemMap","B_UavTerminal","ItemRadio","ItemCompass","ACE_Altimeter","ACE_NVG_Wide"]],[]]

Result:
0.768012 ms

Cycles:
1303/10000

Code:
[testLoadout] call ace_arsenal_fnc_verifyLoadoutOld

Result:
0.676349 ms

Cycles:
1479/10000

Code:
[testLoadout] call ace_arsenal_fnc_verifyLoadout
```

Loadout verification is already pretty fast now thanks to hashmaps but it's still the part of the arsenal where a user is most likely to notice stuttering or hang ups, particularly with large loadout lists and loadouts with many different item types.

Possible issues: 
- Items with the same classname in different config types (CfgWeapons and CfgVehicles, for example) could end up not getting registered as null. This is going to be *extremely* rare, though.
- Third parties might rely on the output of `FUNC(verifyLoadout)`? It's not public, though.

### IMPORTANT


addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -23,118 +23,63 @@ if (count _loadout == 2) then {
...
-private _cfgGlasses = configFile >> "CfgGlasses";
-
-private _weapons = GVAR(virtualItems) get IDX_VIRT_WEAPONS;
-private _attachments = GVAR(virtualItems) get IDX_VIRT_ATTACHMENTS;
-
 private _name = "";
-private _nullItemsAmount = 0;
-private _unavailableItemsAmount = 0;
 private _nullItemsList = [];
 private _unavailableItemsList = [];
```
johnb432, line 28:
```suggestion
private _nullItemsList = createHashMap;
private _unavailableItemsList = createHashMap;
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -23,118 +23,63 @@ if (count _loadout == 2) then {
...
-                } else {
-                    _x
-                };
-            } else {
-                _x
+            // If item doesn't exist in config, "" is returned
+            if (_name == "") then {
+                _nullItemsList pushBack _x;
             };
+            _name
```
johnb432, line 40:
```suggestion
                _nullItemsList set [_x, nil];
            };
            
            _name
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -23,118 +23,63 @@ if (count _loadout == 2) then {
...
-                    ) then {
-                        _unavailableItemsList pushBackUnique _x;
-                        _weaponArray set [_forEachIndex, ""];
-                        INC(_unavailableItemsAmount);
-                    };
-                } else {
-                    _nullItemsList pushBackUnique _x;
+                // Check if item is available in arsenal
+                if !(_x in GVAR(virtualItemsFlatAll)) then {
+                    _unavailableItemsList pushBack _x;
```
johnb432, line 70:
```suggestion
                    _unavailableItemsList set [_x, nil];
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -23,118 +23,63 @@ if (count _loadout == 2) then {
...
-                    if !(_magazine in (GVAR(virtualItems) get IDX_VIRT_ITEMS_ALL)) then {
-                        _unavailableItemsList pushBackUnique _magazine;
-                        _weaponArray set [_forEachIndex, []];
-                        INC(_unavailableItemsAmount);
-                    };
-                } else {
-                    _nullItemsList pushBackUnique _magazine;
+                // Check if item is available in arsenal
+                if !(_magazine in GVAR(virtualItemsFlatAll)) then {
+                    _unavailableItemsList pushBack _magazine;
```
johnb432, line 81:
```suggestion
                    _unavailableItemsList set [_magazine, nil];
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -151,151 +96,57 @@ for "_dataIndex" from IDX_LOADOUT_PRIMARY_WEAPON to IDX_LOADOUT_ASSIGNEDITEMS do
...
-                                case (_x isEqualTypeArray [[], 0]): {
-                                    [_x select 0] call _fnc_weaponCheck;
-                                };
-                                // Misc. items have 2 entries: Name and amount, containers have 2 entries: Name and isBackpack
-                                default {
-                                    _x params ["_item"];
+            if !(_item in GVAR(virtualItemsFlatAll)) then {
+                _unavailableItemsList pushBack _item;
+                _loadout set [_dataIndex, []];
+                continue
```
johnb432, line 110:
```suggestion
               _unavailableItemsList set [_x, nil];
                _loadout set [_dataIndex, []];
                
                continue;
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -151,151 +96,57 @@ for "_dataIndex" from IDX_LOADOUT_PRIMARY_WEAPON to IDX_LOADOUT_ASSIGNEDITEMS do
...
-                            };
-                        } forEach _containerItems;
-                    };
-                } else {
-                    _nullItemsList pushBackUnique _item;
-                    _loadout set [_dataIndex, []];
-                    INC(_nullItemsAmount);
+                if (_item isEqualType []) then {
+                    _x call _fnc_weaponCheck;
+                    continue
```
johnb432, line 118: This is more for comprehension than anything.

```suggestion
                   [_item] call _fnc_weaponCheck;
                   
                   continue;
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -151,151 +96,57 @@ for "_dataIndex" from IDX_LOADOUT_PRIMARY_WEAPON to IDX_LOADOUT_ASSIGNEDITEMS do
...
-                        _loadout set [_dataIndex, ""];
-                        INC(_unavailableItemsAmount);
-                    };
-                } else {
-                    _nullItemsList pushBackUnique _item;
-                    _loadout set [_dataIndex, ""];
-                    INC(_nullItemsAmount);
+
+                if !(_item in GVAR(virtualItemsFlatAll)) then {
+                    _unavailableItemsList pushBack _item;
```
johnb432, line 123:
```suggestion
                    _unavailableItemsList set [_item, nil];
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -151,151 +96,57 @@ for "_dataIndex" from IDX_LOADOUT_PRIMARY_WEAPON to IDX_LOADOUT_ASSIGNEDITEMS do
...
-                        _loadout set [_dataIndex, ""];
-                        INC(_unavailableItemsAmount);
-                    };
-                } else {
-                    _nullItemsList pushBackUnique _item;
-                    _loadout set [_dataIndex, ""];
-                    INC(_nullItemsAmount);
-                };
+            if !(_item in GVAR(virtualItemsFlatAll)) then {
+                _unavailableItemsList pushBack _item;
```
johnb432, line 134:
```suggestion
                _unavailableItemsList set [_item, nil];
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -151,151 +96,57 @@ for "_dataIndex" from IDX_LOADOUT_PRIMARY_WEAPON to IDX_LOADOUT_ASSIGNEDITEMS do
...
-                        _assignedItems set [_subIndex, ""];
-                        INC(_nullItemsAmount);
-                    };
-                };
-            };
+            {
+                if !(_x in GVAR(virtualItemsFlatAll)) then {
+                    _unavailableItemsList pushBack _x;
+                    _assignedItems set [_forEachIndex, ""];
+                }
```
johnb432, line 146:
```suggestion
                    _unavailableItemsList set [_x, nil];
                    _assignedItems set [_forEachIndex, ""];
                };
```

addons/arsenal/XEH_postInit.sqf
```diff
@@ -71,10 +71,10 @@ GVAR(lastSortDirectionRight) = DESCENDING;
         _contentPanelCtrl lnbSetData [[_newRow, 1], _playerName + _loadoutName];
 
         // Set color of row, depending if items are unavailable/missing
-        if (_nullItemsAmount > 0) then {
+        if (_nullItemsList isNotEqualTo []) then {
             _contentPanelCtrl lnbSetColor [[_newRow, 1], [1, 0, 0, 0.8]];
         } else {
-            if (_unavailableItemsAmount > 0) then {
+            if (_unavailableItemsList isNotEqualTo [""]) then {
```
johnb432, line 77:
```suggestion
            if (_unavailableItemsList isNotEqualTo []) then {
```

addons/arsenal/functions/fnc_fillLoadoutsList.sqf
```diff
@@ -50,10 +50,10 @@ if (GVAR(currentLoadoutsTab) != IDC_buttonSharedLoadouts) then {
             _loadoutCachedInfo = [_loadoutData] call FUNC(verifyLoadout);
             _contentPanelCtrl setVariable [_loadoutNameAndTab, _loadoutCachedInfo];
 
-            _loadoutCachedInfo params ["", "_nullItemsAmount", "_unavailableItemsAmount", "_nullItemsList", "_unavailableItemsList"];
+            _loadoutCachedInfo params ["", "_nullItemsList", "_unavailableItemsList"];
 
             // Log missing / nil items to RPT (only once per arsenal session)
-            if (GVAR(EnableRPTLog) && {(_nullItemsAmount > 0) || {_unavailableItemsAmount > 0}}) then {
+            if (GVAR(EnableRPTLog) && {(_nullItemsList isNotEqualTo []) || {_unavailableItemsList isNotEqualTo [""]}}) then {
```
johnb432, line 56:
```suggestion
            if (GVAR(EnableRPTLog) && {(_nullItemsList isNotEqualTo []) || {_unavailableItemsList isNotEqualTo []}}) then {
```

johnb432 (commented):
Mostly formatting changes in my review, but also some a very minor performance improvement suggestion.

johnb432:
> Optimization 4: Change use of `pushBackUnique` to combo `pushBack` and `arrayIntersect` as that is faster in most cases now, particularly with large arrays.

Are you sure about this? I had the impression the opposite was true.
That's neat information for the future if true.

jonpas:
Yeah, that is true IIRC. With `pushBackUnique` you check if the value already exists (so you iterate through the whole thing) each time you call it. With `pushBack` you don't do that, and you only do it once with `arrayIntersect` at the end.

LinkIsGrim:
> Mostly formatting changes in my review, but also some a very minor performance improvement suggestion.

I was under the assumption that creating a hashmap/setting a key is slower than using arrays, and I was specifically avoiding using `- [""]`. Have you profiled that to see if it's actually faster?

johnb432:
> I was under the assumption that creating a hashmap/setting a key is slower than using arrays, and I was specifically avoiding using `- [""]`. Have you profiled that to see if it's actually faster?

It really depends on what is being done:

Hashmaps will be roughly 80-90x faster if there are loads of different entries, like here:
```SQF
private _nullItemsList = createHashMap;

for "_i" from 1 to 10000 do {
	_nullItemsList set [_i, nil];
};

keys _nullItemsList
```

Arrays will be 2x faster if there are loads of the same entry:
```SQF
private _nullItemsList = [];

for "_i" from 1 to 10000 do {
	_nullItemsList pushback 1; // Not _i like in example above!
};

_nullItemsList arrayIntersect _nullItemsList
```

I'm guessing in this application it's more of the first case, however it doesn't make a measurable difference.
I'll edit/remove my suggestions if you want to stick with an array.
---
What if a loadout were to have all items available - does `_unavailableItemsList` still contain `""`? Looking at the code it wouldn't, so it should be removed imo.

johnb432:
> If a loadout were to have all items available `_unavailableItemsList` wouldn't contain `""`, so it has to be removed. Also feels like whack to have to check for `[""]` specifically instead of `[]` imo.

Updated my suggestions @LinkIsGrim

LinkIsGrim:
https://github.com/LinkIsGrim/ACE3/commit/e5950f2818d929af38766ad56a7ec0afc8af0862 switches to a single iteration through the loadout array (turning everything into configCase, checking for null and availability all at once).

Output for the same test loadout:
```
Result:
0.676349 ms

Cycles:
1479/10000

Code:
[testLoadout] call ace_arsenal_fnc_verifyLoadout
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -23,279 +23,50 @@ if (count _loadout == 2) then {
...
+                };
             };
+
+            _name
         } else {
             // Handle arrays
             if (_x isEqualType []) then {
-                _x call _fnc_toConfigCase
+                _itemArray = _x call _fnc_filterLoadout;
+                if ((_itemArray select 0) isEqualTo "") then {
```
LinkIsGrim, line 56: Kind of a hack here: `_itemArray select 0` can be a string (magazines, items, etc), or an array (weapons, primarily). So `isEqualTo` is necessary.

johnb432, line 56: This might fix it?

```suggestion
                if ((_itemArray select 0) isEqualTo "" && {!(_itemArray isEqualTypeAll "")}) then {
```

This one might be a bit safer:
```suggestion
                if ((_itemArray select 0) isEqualTo "" && {!(_itemArray isEqualTypeParams ["","","","","",""])}) then {
```

LinkIsGrim: That makes sense.

addons/arsenal/script_component.hpp
```diff
@@ -3,7 +3,7 @@
 #include "\z\ace\addons\main\script_mod.hpp"
 
 // #define DEBUG_MODE_FULL
-// #define DISABLE_COMPILE_CACHE
+#define DISABLE_COMPILE_CACHE
```
LinkIsGrim, line 6:
```suggestion
// #define DISABLE_COMPILE_CACHE
```

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -23,279 +23,51 @@ if (count _loadout == 2) then {
...
             };
+
+            _name
         } else {
             // Handle arrays
             if (_x isEqualType []) then {
-                _x call _fnc_toConfigCase
+                _itemArray = _x call _fnc_filterLoadout;
+                // isEqualTypeParams check is used for assignedItems array
+                if ((_itemArray select 0) isEqualTo "" && {!(_itemArray isEqualTypeParams ["","","","","",""])}) then {
```
johnb432, line 57: This is does not play well if it's a weapon array and the weapon is not available.

Generally speaking, what should this actually do?

LinkIsGrim: Containers are represented as ["containerClassname", [itemArray]], "containerClassname" is replaced by "" if it's unavailable, this is supposed to replace [itemArray] with []

johnb432: > Containers are represented as ["containerClassname", [itemArray]], "containerClassname" is replaced by "" if it's unavailable, this is supposed to replace [itemArray] with []

It's not necessary to remove it, at least for `setUnitLoadout`. It causes more issues than it's worth imo, better off dropping it unless we really need it?

johnb432:
```suggestion
                if (count _itemArray == 2 && {(_itemArray select 0) isEqualTo ""} && {(_itemArray select 1) isEqualType []}) then {
```

Testing with this has so far been successful.

`// If "" is given as a backpack, an error is thrown, therefore, filter out all unavailable/null containers` a better comment, but I can't include it with the suggestion I made here.

johnb432: I've taken the liberty to merge this suggestion.

addons/arsenal/functions/fnc_verifyLoadout.sqf
```diff
@@ -23,279 +23,51 @@ if (count _loadout == 2) then {
...
-                };
+            // If item doesn't exist in config, "" is returned
+            if (_name == "") then {
+                _nullItemsList pushBack _x;
             } else {
-                _x
+                // Check if item or its base weapon exist in the arsenal
+                if !(_name in GVAR(virtualItemsFlatAll)) then {
+                    _name = _name call FUNC(baseWeapon);
+                    if !(_name in GVAR(virtualItemsFlatAll)) then {
```
johnb432, line 44: Do we want `GVAR(virtualItemsFlatAll)` or `GVAR(virtualItemsFlat)`? I feel `GVAR(virtualItemsFlat)` is more appropriate, seeing as `GVAR(virtualItemsFlatAll)` includes unique items.

LinkIsGrim: I like the idea of `GVAR(virtualItemsFlatAll)` being more flexible, but it can be used to duplicate items, so `GVAR(virtualItemsFlat)` it is.

johnb432:
```suggestion
                if !(_name in GVAR(virtualItemsFlat)) then {
                    _name = _name call FUNC(baseWeapon);
                    if !(_name in GVAR(virtualItemsFlat)) then {
```

johnb432 (approved):
I've tested it, but I feel it could do with some more by others.

addons/arsenal/functions/fnc_fillLoadoutsList.sqf
```diff
@@ -50,10 +50,10 @@ if (GVAR(currentLoadoutsTab) != IDC_buttonSharedLoadouts) then {
             _loadoutCachedInfo = [_loadoutData] call FUNC(verifyLoadout);
             _contentPanelCtrl setVariable [_loadoutNameAndTab, _loadoutCachedInfo];
 
-            _loadoutCachedInfo params ["", "_nullItemsAmount", "_unavailableItemsAmount", "_nullItemsList", "_unavailableItemsList"];
+            _loadoutCachedInfo params ["", "_nullItemsList", "_unavailableItemsList"];
 
             // Log missing / nil items to RPT (only once per arsenal session)
-            if (GVAR(EnableRPTLog) && {(_nullItemsAmount > 0) || {_unavailableItemsAmount > 0}}) then {
+            if (GVAR(EnableRPTLog) && {(_nullItemsList isNotEqualTo []) || {_unavailableItemsList isNotEqualTo []}}) then {
```
rautamiekka, line 56: Doubt the lazy evals save anything here with checks this simple/quick.
```suggestion
            if (GVAR(EnableRPTLog) && (_nullItemsList isNotEqualTo [] || _unavailableItemsList isNotEqualTo [])) then {
```

rautamiekka (changes requested):

addons/arsenal/functions/fnc_fillLoadoutsList.sqf
LinkIsGrim: They do, lazy eval fail is still faster than creating an empty array.

rautamiekka: Interesting. Might be circumventable if pre-created, but, point.

PR #9270, opened 2023-07-20
Fastroping - Improve AI flight height

SpicyBagpipes:
New alt syntax forces AI to fly at specified height

**When merged this pull request will:**
- Modify the AI fast roping waypoint to use new alt syntax for flyInHeight introduced with 2.14

### IMPORTANT

- [x] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9327, opened 2023-08-16
Update fnc_deployAIWaypoint.sqf

PabstMirror:
ref #9270

### DIFF
--- a/addons/fastroping/functions/fnc_deployAIWaypoint.sqf
+++ b/addons/fastroping/functions/fnc_deployAIWaypoint.sqf
@@ -24,7 +24,7 @@ private _speedMode = speedMode _group;
 // - Approach -----------------------------------------------------------------
 if (_vehicle distance2D _position > 50) then {
     _group setSpeedMode "LIMITED";
-    _vehicle flyInHeight [20, true];
+    _vehicle flyInHeight ([20, [20, true]] select ((productVersion # 2) >= 213)); // remove selecet after 2.14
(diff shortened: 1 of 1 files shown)


addons/fastroping/functions/fnc_deployAIWaypoint.sqf
```diff
@@ -24,7 +24,7 @@ private _speedMode = speedMode _group;
 // - Approach -----------------------------------------------------------------
 if (_vehicle distance2D _position > 50) then {
     _group setSpeedMode "LIMITED";
-    _vehicle flyInHeight [20, true];
+    _vehicle flyInHeight ([20, [20, true]] select ((productVersion # 2) >= 213)); // remove selecet after 2.14
```
rautamiekka, line 27:
```suggestion
    _vehicle flyInHeight ([20, [20, true]] select ((productVersion # 2) >= 213)); // remove select after 2.14
```

rautamiekka (changes requested):

jonpas:
I don't expect us to release anything before 2.14, so let's refrain from merging this unless we really need it.

PabstMirror:
My group uses dev builds and doubt we're the only one
I'd need master to be usable with current arma

jonpas:
Ah fair

jonpas (approved):

SpicyBagpipes:
Need me to resubmit PR once this is merged?

PR #9325, opened 2023-08-14
Medical GUI - Add pagination for categories

SpicyBagpipes:
**When merged this pull request will:**
- Add pagination for categories

### IMPORTANT

- [x] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9332, opened 2023-08-18
Medical Gui - Dynamically create any number of action buttons

PabstMirror:
maybe alternate to #9325


SpicyBagpipes:
Imo is the better option between this and pagination, would just have the padding between buttons still and make scroll bar edge not overlap triage end
image

kymckay (approved):
Nice and clean, love a PR that reduces lines of config!

SpicyBagpipes:
Looks good now, should I close my PR?

LinkIsGrim (approved):

Issue #1119, opened 2015-05-12
Linking Belts - reloader will get wrong ammo count


PabstMirror:
After re-linking a belt, other units will get wrong count for check ammo.
The ammo will show as 0 to the reloader, so they won't be able to link belt again.

can verify by checking difference between `z1 ammo "LMG_Mk200_F"` on both shooter and reloader clients.

Problem seems to be `setAmmo` (local unit / global effects) isn't propagated/synced immediately 
(seems to take up to 40 seconds)

* A Starting Ammo 20.
* B Reloads ammo to 100
* A can shoot 15 more [A will see ammo 85, B will see amoo 5]

So clients seem to keep their own ammo count of other unit's weapons and decrements it on each fired event.
Wrong values will eventually resync to the correct value.

PR #9334, opened 2023-08-19
Reload Launchers - Code cleanup

johnb432:
- Title.
- Some changes :
    - If a launcher has two muzzles, e.g. RHS SMAW, and if the secondary muzzle were loaded but not the first, you could not use the reload action. Same happened if the launcher had a magazine in the primary muzzle with 0 ammo in it. This PR makes it so that it only checks the primary muzzle and checks if the ammo is 0.
    - The compatible magazines check did not check for magazines in magwells. This PR adds support for magwells.
    
The changes were included in this PR because I deemed them to be relatively small.

### IMPORTANT

### DIFF
--- a/addons/reloadlaunchers/functions/fnc_canLoad.sqf
+++ b/addons/reloadlaunchers/functions/fnc_canLoad.sqf
@@ -4,16 +4,16 @@
  * Check of the unit can reload the launcher of target unit.
  *
  * Arguments:
- * 0: Unit to do the reloading <OBJECT>
- * 1: Unit eqipped with launcher <OBJECT>
- * 2: weapon name <STRING>
- * 3: missile name <STRING>
+ * 0: Unit to execute the reload <OBJECT>
+ * 1: Unit equipped with the launcher <OBJECT>
+ * 2: Launcher name <STRING>
+ * 3: Missile name <STRING>
  *
  * Return Value:
  * None
  *
  * Example:
- * [bob, kevin, "weapon", "missile"] call ace_reloadlaunchers_fnc_canLoad
+ * [player, cursorTarget, "launch_RPG32_F", "RPG32_F"] call ace_reloadlaunchers_fnc_canLoad
  *
  * Public: No
  */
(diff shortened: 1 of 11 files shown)


addons/reloadlaunchers/functions/fnc_canLoad.sqf
```diff
@@ -4,38 +4,38 @@
...
+// Check if target has its secondary weapon equipped
 if !(_weapon in weapons _target) exitWith {false};
 
-// check if the target really needs to be reloaded
-if (count secondaryWeaponMagazine _target > 0) exitWith {false};
-
-// check if the launcher is compatible
-if (getNumber (configFile >> "CfgWeapons" >> _weapon >> QGVAR(enabled)) == 0) exitWith {false};
+// Check if the target's launcher's primary muzzle really needs to be reloaded
+if (((getUnitLoadout _target) select 1 param [4, []] param [1, 0]) != 0) exitWith {false};
```
LinkIsGrim, line 38: Oof. `getUnitLoadout` is *slow* for a condition. Maybe use `weaponsInfo` with 2.14 changes (or `weaponState`) instead?

addons/reloadlaunchers/functions/fnc_getLoadableMissiles.sqf
```diff
@@ -1,29 +1,23 @@
...
-// get available magazines of reloader, Note: "magazines" does not include currently loaded magazines
-private _magazines = magazines _unit;
-
-// case sensitvity
-_magazines = _magazines apply {toLower _x};
-
-// get reloaders magazine types compatible with targets launcher. No duplicates.
-getArray (configFile >> "CfgWeapons" >> _weapon >> "magazines") select {toLower _x in _magazines} // return
+// Look for primary muzzle magazines only
+(compatibleMagazines [_weapon, "this"]) arrayIntersect (magazines _unit)
```
LinkIsGrim, line 23: Again, slow, but faster than before I guess.

addons/reloadlaunchers/functions/fnc_canLoad.sqf
johnb432: I might be able to get away with `_target ammo _weapon`, since we are checking primary muzzles only, but I need to check more.

johnb432: > I might be able to get away with `_target ammo _weapon`, since we are checking primary muzzles only, but I need to check.

Testing suggests `_target ammo _weapon` works.

addons/reloadlaunchers/functions/fnc_getLoadableMissiles.sqf
johnb432: I don't think there is much we can do to speed this up.

Drofseh: > I don't think there is much we can do to speed this up.

Is it worth caching?

PabstMirror: `(compatibleMagazines [secondaryWeapon player, "this"]) arrayIntersect (magazines player);` is ~0.0043ms (with 16 mags)
I don't think it's a problem

addons/reloadlaunchers/functions/fnc_canLoad.sqf
johnb432: I was digging into `ace_reload`, as I'm cleaning up its code too. I stumbled across https://github.com/acemod/ACE3/issues/1119. I can confirm that the issue described in https://github.com/acemod/ACE3/issues/1119 still exists. It also happens when `player setUnitLoadout getUnitLoadout player` is used.
`weaponState` in those cases also returns incorrect information.

A ticket can be found here: https://feedback.bistudio.com/T167015, to which I'll add my findings.

But because of this, I'm not sure if using `_target ammo _weapon` is safe.

LinkIsGrim (approved):
Network sync issues aren't going to be any worse than what we currently have, IMO

PR #9294, opened 2023-07-28
Compat RHS - Make compats autoload if mod is present

LinkIsGrim:
**When merged this pull request will:**
- Title.
- General config cleanup.

Non-dependency creating changes (hearing protection, goggles, ballistics values, etc) are kept in the main component. Rest are moved to a sub-component requiring both the mod and their respective addon.

I've opted to only keep the breaking changes in the sub-components instead of all relevant entries, to make getting rid of compats easier in the future should they no longer be necessary.

### IMPORTANT

- [ ] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9204, opened 2023-05-23
Compat SOG - Make compat auto load if DLC is present

veteran29:
**When merged this pull request will:**
- title
- use `skipWhenMissingDependencies` for auto loading when DLC is available

### IMPORTANT

- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9339, opened 2023-08-21
Tools - Update publisher for compats moving to main

PabstMirror:
Update publish.py tool for compats we move to main addons

- [x] sog #9204
- [x] rhs #9294

Testing what we should do with old workshop entries:
retire
Publisher won't let you push an update with no pbos
but can be bypassed by putting a dummy in `\dumFolder\anything.pbo` instead of `\addons`
also put a file in `\addons\something.txt` or it will error

once that's pushed the arma launcher will complain the addon is corrupted

I think that and also renaming the workshop entry to something like
`ACE Compat - SOG` to `UNSUBSCRIBE/No Longer Used - ACE Compat - SOG` 
should work and hopefully minimize the number of "ace_x" is outdated issues we get


PabstMirror:
per grim, it might be a better idea to keep these around and just keep them updated with identical pbos to the main
loading 2 identical/signed pbos shouldn't actually cause any problems

will still need to update this tool for the new path locations of moved compats (or handle in each individual PR)

jonpas:
> PabstMirror: I think I changed my opinion, we should just delete then steam/arma launcher will do a good job deleting them it might cause a small warning with old modlists, but shouldn't be a real problem
> I tested with a test mod and when I deleted it just went away cleanly

PabstMirror:
we still need to remove stuff from the publish.py tool

jonpas (approved):

PR #9340, opened 2023-08-21
Weather - Fix number in `fnc_calculateWindChill.sqf`

PabstMirror:
`_z = 1,2;` is valid sqf 🤷


kymckay (approved):
Good catch, who'd have thought the little known second statement delimiter would cause bugs :laughing:

jonpas (approved):

PR #9343, opened 2023-08-26
Reload - Code cleanup and various improvements

johnb432:
- Title.
- Improvements/Changes:
  - Previously, when linking a belt, a magazine was removed upon the action start from the unit's (who was trying to link) inventory and readded in case of failure. This PR changes it: It now only removes the magazine at the end of the progressbar upon successful completion. It will check if it was able to remove the magazine; if not, the interaction is cancelled.
  - Upon completion or failure of a belt linking, a notification shows up. #9202 (or a follow-up PR) could be made to disable these notifications.
  - Previously, there was some very specific behaviour concerning UGLs and checking of their ammo. This has been expanded to where if any secondary muzzle has no ammo left, it will show the number of compatible magazines on the unit.
  - CBA events are now added to connected machines that don't have interfaces. I imagine this could cause issues if the unit you are trying to link a belt to is local to a dedicated server or headless client. I don't remember coming across problems stemming from this, but it's a safety measure that doesn't impact performance greatly.
  - If `displayNameShort` for static weapon magazines can't be found, `displayName` is used instead. I don't recall encountering an issue caused from the lack of a `displayNameShort` entry, but it's a small safeguard that doesn't cost much.

### IMPORTANT

### DIFF
--- a/addons/reload/functions/fnc_getAmmoToLinkBelt.sqf
+++ b/addons/reload/functions/fnc_getAmmoToLinkBelt.sqf
@@ -4,7 +4,7 @@
  * Check if the target has an MG equipped with belt system that a unit can link.
  *
  * Arguments:
- * 0: Unit to execute the reload <OBJECT>
+ * 0: Unit wanting to execute the reload <OBJECT>
  * 1: Unit equipped with the weapon <OBJECT>
  *
  * Return Value:
--- a/addons/reload/functions/fnc_getAmmoToLinkBelt.sqf
+++ b/addons/reload/functions/fnc_getAmmoToLinkBelt.sqf
@@ -1,11 +1,11 @@
 #include "script_component.hpp"
 /*
  * Author: esteldunedain, phyma
- * Check if the target has an MG equiped with belt system that the player can link
+ * Check if the target has an MG equipped with belt system that a unit can link.
  *
  * Arguments:
- * 0: Player <OBJECT>
- * 1: Target <OBJECT>
+ * 0: Unit to execute the reload <OBJECT>
+ * 1: Unit equipped with the weapon <OBJECT>
  *
  * Return Value:
  * Maximum ammo of magazine (-1 on error) <NUMBER>
@@ -16,12 +16,12 @@
  * Public: No
  */
 
-params ["_player", "_target"];
+params ["_unit", "_target"];
 
-if (vehicle _target != _target) exitWith {-1};
+if !(isNull objectParent _target) exitWith {-1};
 
-private _magazineType = currentMagazine _target;
-private _magazineCfg = configFile >> "CfgMagazines" >> _magazineType;
+private _magazine = currentMagazine _target;
+private _magazineCfg = configFile >> "CfgMagazines" >> _magazine;
 
 if (getNumber (_magazineCfg >> "ACE_isBelt") == 0) exitWith {-1};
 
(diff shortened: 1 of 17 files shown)


LinkIsGrim (approved):

addons/reload/functions/fnc_getAmmoToLinkBelt.sqf
```diff
@@ -1,11 +1,11 @@
 #include "script_component.hpp"
 /*
  * Author: esteldunedain, phyma
- * Check if the target has an MG equiped with belt system that the player can link
+ * Check if the target has an MG equipped with belt system that a unit can link.
  *
  * Arguments:
- * 0: Player <OBJECT>
- * 1: Target <OBJECT>
+ * 0: Unit to execute the reload <OBJECT>
```
rautamiekka, line 7: _(EDIT: Changed to the wording in `fnc_startLinkingBelt.sqf`)_

```suggestion
 * 0: Unit executing the reload <OBJECT>
```

addons/reload/initKeybinds.sqf
```diff
@@ -0,0 +1,16 @@
+#include "\a3\ui_f\hpp\defineDIKCodes.inc"
+
+// Add Keybinds
+["ACE3 Weapons", QGVAR(checkAmmo), LSTRING(checkAmmo), {
+    // Conditions: canInteract
+    if !([ACE_player, vehicle ACE_player, ["isNotInside", "isNotSwimming", "isNotSitting"]] call EFUNC(common,canInteractWith)) exitWith {false};
```
rautamiekka, line 6:
```suggestion
    if !([ACE_player, objectParent ACE_player, ["isNotInside", "isNotSwimming", "isNotSitting"]] call EFUNC(common,canInteractWith)) exitWith {false};
```
Copied from `fnc_getAmmoToLinkBelt.sqf` above. Dunno if that works as intended, but I don't really see why not.

rautamiekka (changes requested):

addons/reload/functions/fnc_getAmmoToLinkBelt.sqf
johnb432: The function is called before the actual reload, so I prefer the original wording.

addons/reload/initKeybinds.sqf
LinkIsGrim: `vehicle` is more appropriate in this case.

rautamiekka: How come ?

addons/reload/functions/fnc_getAmmoToLinkBelt.sqf
rautamiekka: The problem is the original is confusing.

How about this ?
```suggestion
 * 0: Unit that would execute the reload <OBJECT>
```

rautamiekka: Or this ?
```suggestion
 * 0: Unit wanting to execute the reload <OBJECT>
```

addons/reload/initKeybinds.sqf
LinkIsGrim: canInteractWith conditions could potentially rely on `_unit == _target` (if unit is on foot, `vehicle _unit` returns `_unit`

rautamiekka: Nothing can be simple in SQF ...

PR #9344, opened 2023-08-26
Common - Improve weapon unloading

johnb432:
- Previously, if you start unloading a weapon and switch to another, it will duplicate the magazine of the newly selected weapon. This is fixed with this PR.
- The unload action can be spammed, making it play the reload sounds as many times as the action was spammed.
- Previously, you could press the unload keybind in turrets. It would play the sound, but do nothing. This PR makes it so that you need to be able to use the weapon in order to unload it.
- Because of the fixes above, `ace_common_fnc_unloadUnitWeapon` now can take an additional argument: `_magazine`.
- Additionally, for the reloading sounds it also looks for files with the `.ogg` extension.

### IMPORTANT

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -540,21 +540,30 @@ GVAR(deviceKeyCurrentIndex) = -1;
 [0xC7, [true, false, false]], false] call CBA_fnc_addKeybind;  //SHIFT + Home Key
 
 
-["ACE3 Weapons", QGVAR(unloadWeapon), LLSTRING(unloadWeapon), {
-    // Conditions:
-    if !([ACE_player, objNull, ["isNotInside"]] call FUNC(canInteractWith)) exitWith {false};
+["ACE3 Weapons", QGVAR(unloadWeapon), LSTRING(unloadWeapon), {
+    private _unit = ACE_player;
 
-    private _currentWeapon = currentWeapon ACE_player;
-    if !(_currentWeapon != primaryWeapon _unit && {_currentWeapon != handgunWeapon _unit} && {_currentWeapon != secondaryWeapon _unit}) exitWith {false};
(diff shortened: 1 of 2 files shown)


LinkIsGrim (approved):

addons/common/XEH_postInit.sqf
```diff
@@ -540,21 +540,30 @@ GVAR(deviceKeyCurrentIndex) = -1;
...
-["ACE3 Weapons", QGVAR(unloadWeapon), LLSTRING(unloadWeapon), {
-    // Conditions:
-    if !([ACE_player, objNull, ["isNotInside"]] call FUNC(canInteractWith)) exitWith {false};
+["ACE3 Weapons", QGVAR(unloadWeapon), LSTRING(unloadWeapon), {
+    private _unit = ACE_player;
 
-    private _currentWeapon = currentWeapon ACE_player;
-    if !(_currentWeapon != primaryWeapon _unit && {_currentWeapon != handgunWeapon _unit} && {_currentWeapon != secondaryWeapon _unit}) exitWith {false};
+    // Conditions
+    if !([_unit, objNull, ["isNotInside"]] call FUNC(canInteractWith)) exitWith {false};
```
Drofseh, line 547: Does this prevent passengers from unloading their primary/secondary/handgun weapons?

johnb432: > Does this prevent passengers/FFV from unloading their primary/secondary/handgun weapons?

I tested with the RHS BMP-1 and RHS M1232: You can unload in FFV positions, but not when inside as a passenger.
You can't reload your weapon using the reload keybind or action menu, so why should you be able to unload using a keybind? That's my train of thought.

Drofseh: Ok, as long as FFV is supported that's fine

PR #9350, opened 2023-08-29
Captives - Add modded cable tie support

SpicyBagpipes:
- Title, child classes will be compatible

### IMPORTANT

### DIFF
--- a/addons/captives/CfgWeapons.hpp
+++ b/addons/captives/CfgWeapons.hpp
@@ -4,6 +4,7 @@ class CfgWeapons {
 
     class ACE_CableTie: ACE_ItemCore {
         author = ECSTRING(common,ACETeam);
+        ace_captives_cableTie = 1;
         displayName = CSTRING(CableTie);
         descriptionShort = CSTRING(CableTieDescription);
         model = QPATHTOF(models\ace_cabletie.p3d);
--- a/addons/captives/CfgWeapons.hpp
+++ b/addons/captives/CfgWeapons.hpp
@@ -4,7 +4,7 @@ class CfgWeapons {
 
     class ACE_CableTie: ACE_ItemCore {
         author = ECSTRING(common,ACETeam);
-        ace_captives_cableTie = 1;
+        GVAR(cableTie) = 1;
         displayName = CSTRING(CableTie);
         descriptionShort = CSTRING(CableTieDescription);
         model = QPATHTOF(models\ace_cabletie.p3d);
--- a/addons/captives/CfgWeapons.hpp
+++ b/addons/captives/CfgWeapons.hpp
@@ -4,7 +4,7 @@ class CfgWeapons {
 
     class ACE_CableTie: ACE_ItemCore {
         author = ECSTRING(common,ACETeam);
-        GVAR(cableTie) = 1;
+        GVAR(restraint) = 1;
         displayName = CSTRING(CableTie);
         descriptionShort = CSTRING(CableTieDescription);
         model = QPATHTOF(models\ace_cabletie.p3d);
--- a/addons/captives/XEH_preStart.sqf
+++ b/addons/captives/XEH_preStart.sqf
@@ -2,5 +2,5 @@
 
 #include "XEH_PREP.hpp"
 
-private _cableTies = (QUOTE(getNumber (_x >> QQGVAR(is_Cable_Tie)) > 0) configClasses (configFile >> "CfgWeapons") apply {configName _x});
+private _cableTies = (QUOTE(getNumber (_x >> QQGVAR(cableTie)) > 0) configClasses (configFile >> "CfgWeapons") apply {configName _x});
 uiNamespace setVariable [QGVAR(cableTies), compileFinal str _cableTies];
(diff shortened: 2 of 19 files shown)


addons/captives/functions/fnc_canApplyHandcuffs.sqf
```diff
@@ -20,7 +20,7 @@ params ["_unit", "_target"];
 //Check sides, Player has cableTie, target is alive and not already handcuffed
 
 (GVAR(allowHandcuffOwnSide) || {(side _unit) != (side _target)}) &&
-{"ACE_CableTie" in (_unit call EFUNC(common,uniqueItems))} &&
+{((_unit call EFUNC(common,uniqueItems)) select {_x isKindOf "ACE_CableTie"}) isNotEqualTo []} &&
```
LinkIsGrim, line 23: Not gonna work, need to specify CfgWeapons for the lookup.

SpicyBagpipes: Second time I've done that, my bad. Will fix

PabstMirror:
need to update ACE_captives_fnc_doApplyHandcuffs as well

SpicyBagpipes:
> need to update ACE_captives_fnc_doApplyHandcuffs as well

Just want first instance of kind of ace_cabletie or child found removed?

johnb432:
Why not go with the approach found here?
https://github.com/acemod/ACE3/blob/master/addons/trenches/functions/fnc_hasEntrenchingTool.sqf

SpicyBagpipes:
> Why not go with the approach found here? https://github.com/acemod/ACE3/blob/master/addons/trenches/functions/fnc_hasEntrenchingTool.sqf

Don't see any benefit to it in this instance as no backpacks etc for entrenching tools and anyone who has already tried to create their own likely inherits from cable tie

johnb432:
> Don't see any benefit to it in this instance as no backpacks etc for entrenching tools and anyone who has already tried to create their own likely inherits from cable tie

Reasonable.

The only benefit I see in adding a function is for public usage, but I'm not sure how needed such a function would be.

commy2:
I propose making it an attribute. I don't like using inheritance for this.

ace_captives_is_cable_tie = 1;

addons/captives/XEH_preStart.sqf
```diff
@@ -1,3 +1,6 @@
 #include "script_component.hpp"
 
 #include "XEH_PREP.hpp"
+
+private _cableTies = (QUOTE(getNumber (_x >> QQGVAR(is_Cable_Tie)) > 0) configClasses (configFile >> "CfgWeapons") apply {configName _x});
```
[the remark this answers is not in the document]
Drofseh: I think `isRestraint` might be better than `isCabletie`.
It's more generic and fits better for rope or handcuff items

Drofseh:
I think `isRestraint` might be better than `isCabletie`. 
It's more generic and fits better for rope or handcuff items

SpicyBagpipes:
> I think `isRestraint` might be better than `isCabletie`. It's more generic and fits better for rope or handcuff items

As is not exposed to end users I think `isCabletie` is better as is in keeping with ace item name

SpicyBagpipes:
Though not opposed to adding a rope restraint or something to expand on it? 🤷‍♀️

BaerMitUmlaut (changes requested):
Attribute is currently not set on the default cable ties.

addons/captives/CfgWeapons.hpp
```diff
@@ -4,6 +4,7 @@ class CfgWeapons {
 
     class ACE_CableTie: ACE_ItemCore {
         author = ECSTRING(common,ACETeam);
+        ace_captives_cableTie = 1;
```
veteran29, line 7: GVAR

veteran29 (changes requested):

addons/captives/functions/fnc_canApplyHandcuffs.sqf
```diff
@@ -20,7 +20,7 @@ params ["_unit", "_target"];
 //Check sides, Player has cableTie, target is alive and not already handcuffed
 
 (GVAR(allowHandcuffOwnSide) || {(side _unit) != (side _target)}) &&
-{"ACE_CableTie" in (_unit call EFUNC(common,uniqueItems))} &&
+{((_unit call EFUNC(common,uniqueItems)) arrayIntersect GVAR(cableTies)) isNotEqualTo []} &&
```
veteran29, line 23: `findAny` != 1 should be faster.

SpicyBagpipes: Done in latest commit

SpicyBagpipes:
Can do this for all the other "tools" currently present (think it's just the defusal kit left with no modded support) if desired, changes only add .0012ms to the condition.

Drofseh:
> > I think `isRestraint` might be better than `isCabletie`. It's more generic and fits better for rope or handcuff items
> 
> As is not exposed to end users I think `isCabletie` is better as is in keeping with ace item name

Right, I'm saying it should be different than the item name because if I add a handcuff item in the future it will be ofd to call it a cable tie.

SpicyBagpipes:
> Right, I'm saying it should be different than the item name because if I add a handcuff item in the future it will be odd to call it a cable tie.

Done

LinkIsGrim (approved):

PR #9354, opened 2023-08-30
Arsenal - Add option to show live search results

PabstMirror:


addons/arsenal/functions/fnc_onKeyDown.sqf
```diff
@@ -183,6 +178,14 @@ if (!isNull _loadoutsDisplay) then {
                 };
             };
         };
+        if (_updateSearchResults) then {
```
PabstMirror, line 181: the only hangup is here,
`handleSearchBar` will check `ctrlText _control`
if you are typing letters, the letters will show up in that as you expect
if you hit backspace, it apparently doesn't take effect until after this returns

normally I'd just use execNextFrame but this needs to work in 3den
any ideas?

LinkIsGrim: Isn't there an EH for CT_EDIT changes?

LinkIsGrim: https://community.bistudio.com/wiki/User_Interface_Event_Handlers#onEditChanged 2.14

LinkIsGrim: With above I'd say we can just have live search always on.

LinkIsGrim:
It works well but it would look better if the refresh animation was skipped with live search.

LinkIsGrim:
I'm all for just making this the new default.

LinkIsGrim (approved):

PabstMirror:
I'm not sure; 
testing with a big modset
typing is fine as it's not doing a full search each time
hitting backspace and it does a full search and it can be a little sluggish
but all the keypresses are "buffered" so it still works fine

maybe we should just get some real-world testing on dev branch and decide later about making it default

LinkIsGrim:
We could modify https://github.com/acemod/ACE3/blob/1492e4ab098419e8714e339e8d14ecb609798e1a/addons/arsenal/functions/fnc_addListBoxItem.sqf#L24-L40 to skip the item if it doesn't match search. That would be faster, but maybe a bit dirty?

PR #9356, opened 2023-08-30
Tools - make.py will try running pboProject once if it's not in registry

PabstMirror:
Maybe fix random failed builds


veteran29 (approved):

PabstMirror:
got a new fail after this https://github.com/acemod/ACE3/actions/runs/6073044406/job/16474190149?pr=9361

> pboProject found normally via registry

looks like failure is down in `pboproject_settings()` not `find_depbo_tools()`

PR #9364, opened 2023-09-05
Compat GM - Make compat autoload when mod is present

LinkIsGrim:
- Title.
- Cleanup: remove CSW dependency (was never implemented) and fix explosives properties.

Still needs separation into subcomponents, working on that.

### IMPORTANT


addons/compat_gm/config.cpp
```diff
@@ -6,15 +6,15 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"ace_csw", "ace_explosives", "ace_refuel", "gm_core"};
+        requiredAddons[] = {"gm_core"};
```
PabstMirror, line 9: I really don't know if we need this, but I added it for some of the other compats
```suggestion
        requiredAddons[] = {"ace_common", "gm_core"};
```

PabstMirror (approved):
tested working with `@Global Mobilization - Cold War Germany- Compatibility Data for Non-Owners`

addons/compat_gm/config.cpp
LinkIsGrim: Good to have I suppose

jonpas (approved):

PR #9367, opened 2023-09-06
Zeus - Add medical menu module

johnb432:
- Continuation of #9019. Credit goes to @BrettMayson.
- Additional changes (partially from https://github.com/acemod/ACE3/pull/9019#issuecomment-1637042455):
  - Added a CBA setting that acts as a multiplier for treatment times when in Zeus. 0 is a valid number and the treatment happens immediately.
  - Medical menu can be opened on the player who is in Zeus, regardless of consciousness state. In #9019 the player needed to be conscious.

Tested with self-hosted MP, with a second PC acting as a second client. Everything that was tried worked (bandaging, tourniquets, splints, IVs, PAK).

### IMPORTANT


addons/medical_gui/functions/fnc_canOpenMenu.sqf
```diff
@@ -18,8 +18,13 @@
...
 params ["_player", "_target"];
 
-alive _player
-&& {!IS_UNCONSCIOUS(_player)}
-&& {!isNull _target}
-&& {_player distance _target < GVAR(maxDistance) || {vehicle _player == vehicle _target}}
-&& {GVAR(enableMedicalMenu) == 1 || {GVAR(enableMedicalMenu) == 2 && {vehicle _player != _player || {vehicle _target != _target}}}}
+// If in Zeus
+if (!isNull findDisplay 312) exitWith {
+    !isNull _target &&
```
PabstMirror, line 23:
```suggestion
    !isNull _target && {EGVAR(medical_treatment,treatmentTimeCoeffZeus) != 10} &&
```
not sure, it might be nice to have a way to disable
and I don't want to add a brand new setting
this sorta works (disable if time set to max) but is a little ugly

BrettMayson: timeCoef of zero would make more sense to me

johnb432: > not sure, it might be nice to have a way to disable and I don't want to add a brand new setting this sorta works (disable if time set to max) but is a little ugly

@PabstMirror What do you mean by "disabled"? That the setting doesn't have an effect or that you can't open the medical menu via zeus at all?

If it's the latter, then I suggest extending the CBA setting slider from `[0, 10]` to `[-1, 10]` and if the coeff is below 0, it would be considered to be turned off. Personally I would like to have the ability to set the time to 0.
But I don't understand why you would want to be able to enable/disable zeus medical menu access though.

However, if it's the prior, setting the coeff to `1` means no change.

PabstMirror: overloading this setting was a bad idea and is just going to confuse everyone
and most people won't care if zeus heals
we could just add a simple getVar "setting" so it could be disabled if a mission maker/group desires

```suggestion
    !isNull _target && {missionNamespace getVariable [QGVAR(enableZeusModule), true]} &&
```

addons/medical_treatment/functions/fnc_canTreat.sqf
```diff
@@ -20,19 +20,23 @@
...
 private _config = configFile >> QGVAR(actions) >> _classname;
 
 isClass _config
 && {_patient isKindOf "CAManBase"}
-&& {_medic != _patient || {GET_NUMBER_ENTRY(_config >> "allowSelfTreatment") == 1}}
-&& {[_medic, GET_NUMBER_ENTRY(_config >> "medicRequired")] call FUNC(isMedic)}
-&& {[_medic, _patient, _config] call FUNC(canTreat_holsterCheck)}
+&& {_skipForZeus || {_medic != _patient || {GET_NUMBER_ENTRY(_config >> "allowSelfTreatment") == 1}}}
+&& {_skipForZeus || {[_medic, GET_NUMBER_ENTRY(_config >> "medicRequired")] call FUNC(isMedic)}}
+&& {_skipForZeus || {[_medic, _patient, _config] call FUNC(canTreat_holsterCheck)}}
```
LinkIsGrim, line 31: Early exit instead? This looks fugly.

johnb432: I updated it to be more efficient as well.

LinkIsGrim (approved):

PR #9276, opened 2023-07-21
Medical Treatment - Add Grave Digging for Corpse Disposal

Drofseh:
**When merged this pull request will:**
- Add the ability to dig a grave for a unit, requires an entrenching tool
- Create new funciton for both body bag and grave placement

59b3165dd9671229e8da14d951af871a

### IMPORTANT

- [X] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [X] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [X] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9371, opened 2023-09-08
Medical Treatment - Use Laws of War bodybag model and add colored bodybags

LinkIsGrim:
- Title

Needs translations if we want to add separate actions for bagging with a specific color.

### IMPORTANT


PabstMirror:
easy to fix conflict with https://github.com/acemod/ACE3/pull/9276
(could maybe put vehClass/rotation configs on the item)

not a fan of getting popups or end-screen dlc stuff
maybe just add new but leave the original untouched?

LinkIsGrim:
Should be fine if we just keep the inventory item the same. Object body bag shouldn't cause any popups.

PabstMirror:
I changed the original item back to using our item-model so we have at least one choice without dlc stuff
(the DLC item models also look weird when laying on the ground?)

PabstMirror (approved):

LinkIsGrim:
I'd rather use ACE's inventory item model and just retexture it. Couldn't get it to look good, though.

addons/medical_treatment/CfgVehicles.hpp
```diff
@@ -38,6 +38,25 @@ class CfgVehicles {
             };
         };
     };
+    class ACE_bodyBagObject_new: ACE_bodyBagObject {
+        hiddenSelections[] = {"camo"};
+        model = "\A3\Props_F_Orange\Humanitarian\Camps\Bodybag_01_F.p3d";
+    };
+
+    class ACE_bodyBag_black: ACE_bodyBagObject_new {
```
PabstMirror, line 46: I think we should just move this back to `ACE_bodyBagObject`?

just in-case people have mods/scripts that use that specific classname
like "ACE_bodyBagObject" in area detection or cargo loaded
maybe keep the old model as a new classname `ACE_bodyBagObject_original`

PabstMirror (approved):

Issue #9384, opened 2023-09-10
Arsenal - Better support CBA scripted attachments


LinkIsGrim:
`baseWeapon` property handling added in #9040 is pretty good but manually adding those in compats is a lot of maintenance.

For RHS we can use `rhs_optics_base` and `rhs_acc_combo` properties, they should contain the base attachment.

Unsure on CBA implementation.

LinkIsGrim:
#9386 handled RHS attachments. Need to find a solution for CBA/switchable attachments now.

PR #9386, opened 2023-09-11
Arsenal/Compat RHS - Improve support for RHS attachments

LinkIsGrim:
- Add event `ace_arsenal_weaponItemChanged`: params `_weapon, _item, _itemIndex`. `_itemIndex` is relative to `ace_arsenal_currentItems` attachment array. Documented. Fires on weapon attachment/magazine change.
- Use above event to call RHS's function for optic/attachment switching.
- Add baseWeapon property to RHS rail attachments.
- Add handling in `ace_arsenal_fnc_scanConfig` for RHS scripted optics. Overrides baseWeapon cache.

Partial address of #9384.

Prevents this:

And this:

### IMPORTANT


addons/compat_rhs_usf3/XEH_postInit.sqf
```diff
@@ -0,0 +1,31 @@
...
+        };
+        case ITEM_INDEX_BIPOD: {
+            // Need this call to make sure RHS's functions are set
+            call rhs_fnc_accGripod;
+            if (getText (configFile >> "CfgWeapons" >> _item >> "rhs_grip_type") == "") then {
+                call rhs_grip_deinitialize;
+            };
+        };
+    };
+}] call CBA_fnc_addEventHandler;
```
veteran29, line 31: not a fan of how this is duplicated multiple times, maybe we could have additional `compat_rhs` addon that just hosts the scripts and checks for presence of any of RHS mods in the script?

veteran29 (changes requested):

addons/compat_rhs_usf3/XEH_postInit.sqf
```diff
@@ -0,0 +1,31 @@
...
+[QEGVAR(arsenal,weaponItemChanged), {
+    params ["_weapon", "_item", "_itemIndex"];
+    if (EGVAR(arsenal,center) != ACE_player) exitWith {};
+
+    switch (_itemIndex) do {
+        case ITEM_INDEX_SIDE: {
+            call rhs_fnc_anpeq15_rail;
+        };
+        case ITEM_INDEX_OPTIC: {
+            [] call rhs_fnc_preferredOptic;
```
PabstMirror, line 21: with just rhsUS it's undefined but there is

```suggestion
            [] call RHS_fnc_preferedOptic
```
not sure

LinkIsGrim: I forgot to remove this bit. Optic handling isn't necessary as it'll get caught by `baseWeapon` and switched back by RHS on arsenal closing anyway.

addons/compat_rhs_usf3/XEH_postInit.sqf
LinkIsGrim: Could just throw it into arsenal?

PabstMirror (approved):

Issue #9385, opened 2023-09-10
Dragging - Script error in XEH_postInit


PabstMirror:
On dedicated server
```
18:23:23 Error in expression < _unit = (getUserInfo (_this select 5)) select 10;
        _unit call ace_dragg>
18:23:23   Error position: <select 10;
        _unit call ace_dragg>
18:23:23   Error Zero divisor
18:23:23 File /z/ace/addons/dragging/XEH_postInit.sqf..., line 8
```

jonpas:
Comes from this https://github.com/acemod/ACE3/commit/9505d4c47ef56e9c214b89b8d0768d986526c96b#diff-baccfba963e58cce5d824d01073bc7cc39f05a4a36030b6b9923904f6b7126c9 which added [`getUserInfo`](https://community.bistudio.com/wiki/getUserInfo), but that command "Returns an empty array if said player is not found".

This is the exact same issue as https://github.com/OCAP2/addon/pull/31. :smile:

jonpas:
@johnb432

johnb432:
> Comes from this [9505d4c#diff-baccfba963e58cce5d824d01073bc7cc39f05a4a36030b6b9923904f6b7126c9](https://github.com/acemod/ACE3/commit/9505d4c47ef56e9c214b89b8d0768d986526c96b#diff-baccfba963e58cce5d824d01073bc7cc39f05a4a36030b6b9923904f6b7126c9) which added [`getUserInfo`](https://community.bistudio.com/wiki/getUserInfo), but that command "Returns an empty array if said player is not found".
> 
> This is the exact same issue as [OCAP2/addon#31](https://github.com/OCAP2/addon/pull/31). 😄
> 
> It is not a reliable use, especially when players force disconnect or crash.

@jonpas

What do you propose? The fix you implemented for OCAP2 is a solid base, but should I revert the EH change instead? I think `HandleDisconnect` (old EH) *might* be able to handle force disconnections and crashed better than `PlayerDisconnected`, but I'm really not sure. I ran a test where I cut off the internet connection to the connected client and `HandleDisconnect` returned the correct unit (I did not check `PlayerDisconnected`).

The reason why I changed the EH is because I thought `HandleDisconnect` triggered too late. In reality it's fine, I had forgot to take something into account. So the change was unnecessary and it's probably a good idea to revert back to the old EH.

I'll make a PR reverting to the old EH if nobody objects.

jonpas:
[...]

PR #9387, opened 2023-09-11
Dragging - Fix for #9385

johnb432:
- Title (see https://github.com/acemod/ACE3/issues/9385).
- An alternative fix could be:

```SQF
// Release object on disconnection. Function is identical to killed
addMissionEventHandler ["PlayerDisconnected", {
    private _unit = (getUserInfo (_this select 5)) param [10, objNull];
    
    if (isNull _unit) exitWith {}; // This line wouldn't even be necessary, as FUNC(handleKilled) can accept objNull

    _unit call FUNC(handleKilled);
}];
```

`PlayerDisconnected` fires earlier than `HandleDisconnect`, but the comment I added in https://github.com/acemod/ACE3/commit/9505d4c47ef56e9c214b89b8d0768d986526c96b#diff-baccfba963e58cce5d824d01073bc7cc39f05a4a36030b6b9923904f6b7126c9 was wrong; it isn't too late, I just forgot to take something into account whilst testing.

Maybe one has better abnormal disconnection (force quitting, crashing, etc.) detection than the other, but I don't know.

Any input is appreciated.

### IMPORTANT

### DIFF
--- a/addons/dragging/XEH_postInit.sqf
+++ b/addons/dragging/XEH_postInit.sqf
@@ -1,14 +1,9 @@
 // by PabstMirror, commy2
 #include "script_component.hpp"
 
-// Release object on disconnection. Function is identical to killed
 if (isServer) then {
-    // 'HandleDisconnect' EH triggers too late
-    addMissionEventHandler ["PlayerDisconnected", {
-        private _unit = (getUserInfo (_this select 5)) select 10;
-
-        _unit call FUNC(handleKilled);
-    }];
+    // Release object on disconnection. Function is identical to killed
+    addMissionEventHandler ["HandleDisconnect", LINKFUNC(handleKilled)];
 
     // Handle surrending and handcuffing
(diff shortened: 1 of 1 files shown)


jonpas:
> `FUNC(handleKilled) can accept objNull`

Can it though?

---

We don't use `PlayerDisconnected` anywhere in ACE3 (apart from this one), so I assume there is either no difference or `HandleDisconnect` is preferred for simplicity.

jonpas:
Reverts part of #9272. Fixes #9385.

jonpas (approved):

johnb432:
> Can it though?

`_unit getVariable [QGVAR(isDragging), false]` and `_unit getVariable [QGVAR(isCarrying), false]` are checked within `FUNC(handleKilled)`, so if `_unit` is null, it returns `false` in both cases. Nothing else is in `FUNC(handleKilled)` that could be broken by `objNull`.

PR #9389, opened 2023-09-11
Dragging - Improve `featureCamera` handling

johnb432:
- Upon entering an arsenal, the player will drop his dragged or carried object.
- Applies to both BI and ACE arsenals.
- Only applies if the player opens the arsenal on himself.

The right click pop-up is annoying and distracting if you enter an arsenal.

That's why I'm thinking of something similar for when entering the Zeus interface, especially if you use ZEN which uses right clicks. Thoughts on that idea?

### IMPORTANT


LinkIsGrim:
`featureCamera` CBA vent is more generic, if you'd like to use that.

johnb432:
> `featureCamera` CBA event is more generic, if you'd like to use that.

Updated to include that change. Are there any cameras that should allow dragging/carrying?

jonpas:
Splendid Camera should probably allow it.

LinkIsGrim:
> > `featureCamera` CBA event is more generic, if you'd like to use that.
> 
> Updated to include that change. Are there any cameras that should allow dragging/carrying?

Player probably shouldn't drop items when opening Zeus, either.

rautamiekka:
> > > `featureCamera` CBA event is more generic, if you'd like to use that.
> > 
> > 
> > Updated to include that change. Are there any cameras that should allow dragging/carrying?
> 
> Player probably shouldn't drop items when opening Zeus, either.

Yeah, definitely not, will cause more problems than solves/prevents.

johnb432:
> Player probably shouldn't drop items when opening Zeus, either.

Any particular reason why?

Currently if you are dragging/carrying and go into Zeus, the mouse button hint is in the middle of the screen - not a huge deal, but unnecessary and annoying imo.
If you want to move your camera, you will drop the dragged/carried object anyway. I don't understand why you would want to limit yourself by not right-clicking, just to keep the dragged/carried object. Imo far easier to drop and pick up again afterwards.

---

An entirely different approach to this problem could be to "pause" dragging/carrying whilst in any camera other than default. By "pause" I mean remove the mouse button hint and remove the mouse button EH. When getting back into the default camera, those would be added back.

I'm heavily leaning towards this new approach, would be more elegant imo.

rautamiekka:
Dropping could have 2 effects:
1. Tip off the observing players to be unnecessarily cautious, cuz at minimum it looks strange that the Zeus drops something, leaves it there for at least a bit, then picks it up again, and likely rinse-and-repeat, as he's using the interface to see what the Players are doing.
2. Cause a script depending on the Zeus carrying/dragging to break, malfunction, or fire early.

In other words it would hurt or break the story. I don't see any reasons for this for the Zeus interface.

----
Yeah, the "pause" is a better way.

johnb432:
Pause has been added.

LinkIsGrim (approved):

PR #9394, opened 2023-09-12
Advanced Ballistics - Fix `_mapSize` sometimes 0 by using engine command

Eathox:
On certain maps, such as Virtual Reality, `getNumber (configFile >> "CfgWorlds" >> worldName >> "MapSize")` results in 0

### DIFF
--- a/addons/advanced_ballistics/functions/fnc_initializeTerrainExtension.sqf
+++ b/addons/advanced_ballistics/functions/fnc_initializeTerrainExtension.sqf
@@ -19,7 +19,7 @@ if (!hasInterface) exitWith {};
 if (!GVAR(enabled)) exitWith {};
 
 private _initStartTime = diag_tickTime;
-private _mapSize = getNumber (configFile >> "CfgWorlds" >> worldName >> "MapSize");
+private _mapSize = worldSize;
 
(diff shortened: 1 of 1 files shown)


LinkIsGrim:
Is the return from `worldSize` for Altis the same as the config lookup?

PabstMirror:
altis identical
celle2 has config size of 0 like VR
oddly Farabad has config 16125, but worldSize cmd of 10752 (probably inheritance)

LinkIsGrim (approved):

LinkIsGrim:
Farabad's seems to be more accurate with `worldSize`.

Eathox:
Seeing as one or the other is more accurate for different maps, would it be preferable to get both and use one that returns the bigger number?

BrettMayson:
I think `worldSize` is just generally the better option?

rautamiekka:
[`worldSize`](https://community.bistudio.com/wiki/worldSize):
> Returns the engine calculated size (terrain side length) of the current world ([BIS_fnc_mapSize](https://community.bistudio.com/wiki/BIS_fnc_mapSize) may return the same value, but is just a simple look up in [config](https://community.bistudio.com/wiki/configFile)).

PabstMirror (approved):

PR #9397, opened 2023-09-13
Captives - Add Reusable and Locking restraints

SpicyBagpipes:
- Add handcuffs
- Add ability for restraints to be reusable
- Add ability for restraints to be locked restraints
- Add restraint keys

### IMPORTANT

### DIFF
--- a/addons/captives/functions/fnc_setHandcuffed.sqf
+++ b/addons/captives/functions/fnc_setHandcuffed.sqf
@@ -7,6 +7,7 @@
  * 0: Unit <OBJECT>
  * 1: True to take captive, false to release captive <BOOL>
  * 2: Caller <OBJECT>
+ * 3: Selected Cuffs <STRING>
  *
  * Return Value:
  * None
@@ -17,7 +18,7 @@
  * Public: No
  */
 
-params ["_unit", "_state", ["_caller", objNull]];
+params ["_unit", "_state", ["_caller", objNull], "_selectedCuffs"];
 TRACE_3("params",_unit,_state,_caller);
 
 if (!local _unit) exitWith {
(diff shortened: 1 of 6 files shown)


SpicyBagpipes:
Looking for opinions on how cuff should be selected

addons/captives/functions/fnc_setHandcuffed.sqf
```diff
@@ -17,7 +18,7 @@
  * Public: No
  */
 
-params ["_unit", "_state", ["_caller", objNull]];
+params ["_unit", "_state", ["_caller", objNull], "_selectedCuffs"];
```
BrettMayson, line 21:
```suggestion
params ["_unit", "_state", ["_caller", objNull], "_selectedRestraint"];
```

Should use restraint instead of cuffs throughout the code, although we obviously can't change the events, unfortunately

BrettMayson (changes requested):
I think it should also be possible to change the restraint on someone without having to release them first. To switch from cable ties to handcuffs for example.

I could see a option to attempt to break out being wanted by some groups, making the different types of restraints have more varience.

SpicyBagpipes:
> I think it should also be possible to change the restraint on someone without having to release them first. To switch from cable ties to handcuffs for example.

Agreed, maybe have the radial menu for restraints never disappear and just modify the actions to always remove and readd the cuffs?


> I could see a option to attempt to break out being wanted by some groups, making the different types of restraints have more varience.

Imo that'd be too much complexity for this, handcuffs are easy enough to understand as having pros -- being reusable and not everybody can reuse them -- but making cable ties being able to be broken out of could make them pointless overall.

Perhaps if handcuff key is in restrainee's uniform they can break out of the cuffs though?

Drofseh:
Switching the type of restraint is an interesting option and could be added here. 

I think breaking out of things, if desired, should be a separate PR.

rautamiekka:
Breaking out without making the restraint pointless or the Players (hell, even the Zeus) paranoid, now that's a brain twister.

SpicyBagpipes:
> Breaking out without making the restraint pointless or the Players (hell, even the Zeus) paranoid, now that's a brain twister.

Yeah, agreed -- not within the scope of this PR.

Will attempt to do the changes to the action tomorrow/day after however am moving within this timeframe so may not be possible

SpicyBagpipes:
Forgotten where I even got with this, no longer have time or energy to continue PR.

PR #5827, opened 2017-11-28
Add manual reload for vehicle weapons

PabstMirror:
Adds ability to reload vehicle weapons (swapping to a fuller magazine)
I'm not sure why this isn't in the base game

Example:
HMG gunner on car, have 10 rounds left;
before you would have to mag dump to do a tactical reload

PR #9398, opened 2023-09-13
Reload - Add manual reload for vehicle weapons - continued

johnb432:
- Title. Continuation of #5827, credit to @PabstMirror.
- `FUNC(canSwapTurretMagazine)` has been altered slightly from its original form, but the functionality is the same.
- `FUNC(swapTurretMagazine)` has been significantly changed. Previously it removed all turret magazines, whereas now it uses the `"LoadMagazine"` action to change magazines.
- Added an icon to the action (icon is from base game).

### IMPORTANT


johnb432:
Instead of this being an ACE interaction, it could be made to trigger when pressing the reload key instead. Thoughts?

Drofseh:
> Instead of this being an ACE interaction, it could be made to trigger when pressing the reload key instead. Thoughts?

Both? Both!

johnb432:
> Both? Both!

Fair enough xD

On another note: Should there be a CBA setting enabling/disabling this feature?

Drofseh:
I don't really think it needs a setting

PabstMirror:
```
    class HMG_01;
    class HMG_static: HMG_01 {
        type = 1;
    };
```
will let the engine reload via R key

johnb432:
> ```
>     class HMG_01;
>     class HMG_static: HMG_01 {
>         type = 1;
>     };
> ```
> 
> will let the engine reload via R key

Should that be integrated into this PR? As in use that for static weapons, weapons on vehicles or vehicles (I'm unsure if it applies to `CfgVehicles `or `CfgWeapons`)?
Or it could check if ` type = 1` is set for the weapon/vehicle in the `addUserActionEventHandler`?

What do you suggest doing with that config entry?

PabstMirror:
We could just add `type = 1;` too all static weapons to make them reloadable via engine
I can't see any downside to this except that it's more config work

johnb432:
> ```
>     class HMG_01;
>     class HMG_static: HMG_01 {
>         type = 1;
>     };
> ```
> 
> will let the engine reload via R key

Are you sure about this? I have just tried loading CBA and ACE and using any static weapon (all vanilla static weapons have `type = 1`).

ampersand38:
Would [repacking magazines / relinking belts](https://github.com/ampersand38/reload-repack-turret-magazines/blob/main/addons/reload-repack-turret-magazines/fn_reloadTurret.sqf#L53) be in scope here?

johnb432:
> Would [repacking magazines / relinking belts](https://github.com/ampersand38/reload-repack-turret-magazines/blob/main/addons/reload-repack-turret-magazines/fn_reloadTurret.sqf#L53) be in scope here?

I think it would be best as a independent ACE interaction, instead of happening automatically when reloading (if I understood your code correctly).
But I would absolutely love it to be a part of ACE.

PabstMirror:
we've really put it off but https://github.com/acemod/ACE3/pull/9085
contains a lot of what we want for repacking/relinking

Drofseh:
I don't know if 9085 covers it or not, but any packing/linking for vehicles will need to keep the empty magazines or ace rearm can fail with scripted loadouts.

johnb432:
> I don't know if 9085 covers it or not, but any packing/linking for vehicles will need to keep the empty magazines or ace rearm can fail with scripted loadouts.

I was thinking if I would need to adjust this PR for that, but it doesn't look like it. The reload action only happens if there is some ammo available in the weapon (but not a full magazine).

johnb432:
We make it so that this is disabled when CSW is enabled for vehicles, when we get around to merging 9085.

addons/reload/functions/fnc_canSwapTurretMagazine.sqf
```diff
@@ -0,0 +1,52 @@
...
+private _turretPath = _vehicle unitTurret _unit;
+if (_turretPath in [[-1], []]) exitWith {false}; // skip driver / cargo
+if !(_vehicle turretLocal _turretPath) exitWith {false}; // just to be safe
+
+(weaponState [_vehicle, _turretPath]) params ["_weapon", "_muzzle", "", "_magazine", "_ammoCount", "_roundReloadPhase", "_magazineReloadPhase"];
+TRACE_5("",_weapon,_muzzle,_magazine,_ammoCount,typeOf _vehicle);
+
+if ((_weapon == "") || {_weapon != _muzzle}) exitWith {false}; // skip multi-muzzle (he/ap auto-cannons)
+if (_magazine == "") exitWith {false};
+if (_roundReloadPhase + _magazineReloadPhase != 0) exitWith {false}; // can't reload while already reloading
```
LinkIsGrim, line 31:
```suggestion
if (_roundReloadPhase + _magazineReloadPhase != 0) exitWith {false}; // can't reload while already reloading or while shooting
```

LinkIsGrim (approved):

PR #9402, opened 2023-09-17
Common - Add lockInventory status effect

JasperRab:
Adds [`lockInventory`](https://community.bistudio.com/wiki/lockInventory) status effect, for cross-mod compatibility possibilities.
Tested both in singleplayer and multiplayer settings.

### IMPORTANT


PabstMirror:
self note, we could probably use this to replace ACE_VehicleLock_fnc_onOpenInventory which is old/clunky

JasperRab:
The only thing I was still thinking about adding is that it would "kick" players that are currently in the inventory when it's being locked. That currently doesn't happen by vanilla.
However not sure if it's better to do that in a separate PR.

LinkIsGrim (approved):

JasperRab:
Found an issue with this, for JIP players, it won't be locked.
idk if the status effect stuff actually supports JIP stuff?

The setVariable is there on the object. But the code from the CBA event doesn't obviously run.

PR #9409, opened 2023-09-19
Backpacks - Code cleanup

LinkIsGrim:
- Title.

### IMPORTANT

### DIFF
--- a/addons/backpacks/XEH_postInit.sqf
+++ b/addons/backpacks/XEH_postInit.sqf
@@ -1,3 +1,3 @@
 #include "script_component.hpp"
 
-["ace_backpackOpened", {_this call FUNC(backpackOpened)}] call CBA_fnc_addEventHandler;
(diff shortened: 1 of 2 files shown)


PabstMirror:
Looks like you can do `(backpackContainer player) setMaxLoad 100` to add storage to a uav/static backpack, 
not sure if it's a good idea, but at least this code will handle it

PabstMirror (approved):

johnb432:
Unfortunately, this doesn't check the type of container if the passed argument is a container anymore. This means that `(uniformContainer player) call ace_backpacks_fnc_isBackpack` would return `true` (I can't verify this, don't have access to Arma atm).

I think something like this should fix it:

```SQF
if (_backpack isEqualType objNull) exitWith {
    maxLoad _backpack > 0 && {getText (configOf _backpack >> "vehicleClass") == "backpacks"}
};
```

PR #9434, opened 2023-09-25
Laser - Add Individual Laser Code Digit Control

LorenLuke:
- Refactor Laser Code controls to be controlled in a per-digit manner
- Set Default Laser Code to 1688
- Set Default Laser Wavelength to 1064 (Based on Nd:YAG laser designator frequencies; Src: https://en.wikipedia.org/wiki/Laser_designator; https://en.wikipedia.org/wiki/AN/PEQ-1_SOFLAM)
- Update language stringtables to accommodate new commands


PabstMirror (approved):

addons/laser/stringtable.xml
```diff
@@ -51,22 +51,56 @@
...
-            <Czech>Laser - Kód +</Czech>
-            <Italian>Laser - Cambia codice +</Italian>
-            <Japanese>レーザ - コードの数値を増やす</Japanese>
-            <Korean>레이저 - 코드 순환 위</Korean>
-            <Chinesesimp>激光—循环切换激光码 上</Chinesesimp>
-            <Chinese>雷射 - 循環切換雷射碼 上</Chinese>
-            <Turkish>Lazer - Çevrim Kodu Yukarı</Turkish>
+        <Key ID="STR_ACE_Laser_laserCodeUpOnes">
+            <English>Laser - Cycle Code Up - XXX#</English>
+            <German>Lasercode +XXX#</German>
```
PabstMirror, line 56: not sure why german/french are so different
should this be "groBcoden" or something?

LorenLuke: I honestly don't know, and I wanted to mess with existing translations as little as possible in this case if it could at all be avoided.

PR #9439, opened 2023-09-28
Compat SOG - Add barrel swap for MGs and belt linking for drum magazines

Esmeray6:
- Add barrel swapping to the following SOGPF weapons: MG42 and L4.
- Change the drum magazines for MG42 to be considered belts. This will allow belt linking with these magazines.

### IMPORTANT

### DIFF
--- a/addons/compat_sog/CfgMagazines/belts.hpp
+++ b/addons/compat_sog/CfgMagazines/belts.hpp
@@ -12,3 +12,8 @@ class vn_m16_mag_base;
 class vn_m63a_100_mag: vn_m16_mag_base {
     ACE_isBelt = 1;
 };
+class vn_mg42_50_mag: vn_lmgmag_base {
+    ACE_isBelt = 1;
+};
+class vn_mg42_50_t_mag: vn_mg42_50_mag {};
+class vn_rpd_125_mag: vn_rpd_100_mag {};
--- a/addons/compat_sog/CfgMagazines/belts.hpp
+++ b/addons/compat_sog/CfgMagazines/belts.hpp
@@ -15,5 +15,3 @@ class vn_m63a_100_mag: vn_m16_mag_base {
 class vn_mg42_50_mag: vn_lmgmag_base {
     ACE_isBelt = 1;
 };
-class vn_mg42_50_t_mag: vn_mg42_50_mag {};
-class vn_rpd_125_mag: vn_rpd_100_mag {};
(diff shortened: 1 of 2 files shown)


addons/compat_sog/CfgMagazines/belts.hpp
```diff
@@ -12,3 +12,8 @@ class vn_m16_mag_base;
 class vn_m63a_100_mag: vn_m16_mag_base {
     ACE_isBelt = 1;
 };
+class vn_mg42_50_mag: vn_lmgmag_base {
+    ACE_isBelt = 1;
+};
+class vn_mg42_50_t_mag: vn_mg42_50_mag {};
+class vn_rpd_125_mag: vn_rpd_100_mag {};
```
PabstMirror, line 19:
```suggestion
```
ok to drop these imho

Esmeray6: True. Not sure what I was thinking when I added them. Purposes of explicitness, I suppose.

Drofseh:
RPD does not have a removable barrel.

Drofseh:
Neither does the M1918 BAR

Esmeray6:
> RPD does not have a removable barrel.
> Neither does the M1918 BAR

Ah, right. I was thinking more in terms of player convenience at the time. I'll remove them.

addons/compat_sog/CfgMagazines/belts.hpp
```diff
@@ -12,3 +12,6 @@ class vn_m16_mag_base;
 class vn_m63a_100_mag: vn_m16_mag_base {
     ACE_isBelt = 1;
 };
+class vn_mg42_50_mag: vn_lmgmag_base {
```
LinkIsGrim, line 15: I'm not a fan of this. Can MG42 drums be linked?

Esmeray6: If I understand it correctly, [Wikipedia](https://en.wikipedia.org/wiki/MG_42#:~:text=Feeding,-Method%20of%20joining&text=The%20MG%2042%20could%20use,housed%20in%20the%20feed%20cover.) says they cannot. Only through additional connectors, possibly.

Of course, it's not a standard practice in that case. Most of the time they would be loaded with 50-round drum magazines on the go.

Drofseh: They can, but only in 50 round segments (ei. a 200 round belt would have 4 segments).

I would keep them as linkable, it's not significantly less realistic than how linking any other belt can't exceed the `count` of the magazine class.

https://cdn11.bigcommerce.com/s-817fb/images/stencil/500x659/products/5331/69865/IMG_5654__11640.1694970702.jpg

LinkIsGrim (approved):

PR #9442, opened 2023-09-29
Medical Treatement - Add Grave Digging To Body Bags

Drofseh:
- Adds an action to dig a grave for a body bag.

### IMPORTANT


addons/medical_treatment/functions/fnc_placeInBodyBagOrGrave.sqf
```diff
@@ -34,10 +34,17 @@ if (alive _patient) then {
     [_patient, "buried_alive", _medic] call EFUNC(medical_status,setDead);
 };
 
-private _headPos = _patient modelToWorldVisual (_patient selectionPosition "head");
-private _spinePos = _patient modelToWorldVisual (_patient selectionPosition "Spine3");
-private _direction = (_headPos vectorFromTo _spinePos) call CBA_fnc_vectDir;
 private _position = getPosASL _patient;
+private _direction = 0;
+
+if (typeOf _patient == "ACE_bodyBagObject") then {
```
LinkIsGrim, line 40: isKindOf

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -62,6 +62,27 @@ if (isServer) then {
...
+                [[_this#1, _this#0], missionNameSpace getVariable [QGVAR(graveClassname), "Land_Grave_dirt_F"], [0,0,0], missionNameSpace getVariable [QGVAR(graveRotation), -90]] call FUNC(placeInBodyBagOrGrave);
+            },
+            {},
+            LLSTRING(DiggingGrave)
+        ] call EFUNC(common,progressBar);
+    },
+    {[_this#1] call FUNC(canDigGrave)}
+] call EFUNC(interact_menu,createAction);
+
+["ACE_bodyBagObject", 0, ["ACE_MainActions"], _buryBodyBag] call EFUNC(interact_menu,addActionToClass);
```
LinkIsGrim, line 84: Would rather this be added in config

Drofseh: I'll give it a try

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -80,15 +101,21 @@ if (["ace_trenches"] call EFUNC(common,isModLoaded)) then {
             [1.05, 0.02, 0.3] //position in centre of cross
         ] call EFUNC(interact_menu,createAction);
 
-        ["Land_Grave_dirt_F", 0, [], _checkHeadstoneAction] call EFUNC(interact_menu,addActionToClass);
+        [missionNameSpace getVariable [QGVAR(graveClassname), "Land_Grave_dirt_F"], 0, [], _checkHeadstoneAction] call EFUNC(interact_menu,addActionToClass);
```
PabstMirror, line 104: I'm not sure about this
the action has a specific offset for that model and a "t" icon that we might not want to show on generic grave models

Drofseh: Yeah the offset and icon will have to be updated when the new grave model @Marstruc is adding is merged.

Drofseh: @PabstMirror a headstone icon like this suit you?
5dcec0595d1b770f62bb2095c2a6e266

addons/medical_treatment/CfgVehicles.hpp
```diff
@@ -29,6 +29,13 @@ class CfgVehicles {
                 statement = "";
                 icon = "\a3\ui_f\data\IGUI\Cfg\Actions\eject_ca.paa";
                 selection = "";
+                class GVAR(buryBodyBag) {
+                    displayName = CSTRING(DigGrave);
+                    distance = HANDCUFFS_DISTANCE;
```
LinkIsGrim, line 34: Just use the number here. Don't create a dependency.

addons/medical_treatment/CfgVehicles.hpp
```diff
@@ -29,6 +29,13 @@ class CfgVehicles {
...
+                    condition = QUOTE([_this#1] call FUNC(canDigGrave));
```
LinkIsGrim, line 35: Won't this need tweaking because of the isAwake check?

addons/medical_treatment/XEH_postInit.sqf
```diff
@@ -67,28 +67,31 @@ if (["ace_trenches"] call EFUNC(common,isModLoaded)) then {
...
     };
 
     if (isServer) then {
         ["ace_placedInBodyBag", {
             params ["_target", "_restingPlace"];
             TRACE_2("ace_placedInBodyBag eh",_target,_restingPlace);
 
-            private _targetName = [_target, false, true] call EFUNC(common,getName);
+            private _targetName = "";
+            if (_patient isKindOf "ACE_bodyBagObject") then {
```
LinkIsGrim, line 89: _patient or _target?

LinkIsGrim (changes requested):

addons/medical_treatment/CfgVehicles.hpp
Drofseh: I should've put the isAwake check here now that I think about it.

Drofseh: Moved to the unit action in #9451

LinkIsGrim (approved):

PR #9361, opened 2023-09-04
Captives - Blindfold captive units

mrschick:
**When merged this pull request will:**
- Add an interaction that puts a Contact DLC blindfold from one's own inventory onto a captive unit;
- If the captive unit already has some form of goggles on they will be preserved by being moved to the capturer's inventory before being replaced with a blindfold;

PR #9463, opened 2023-10-04
Captives - Allow disarming and blindfolding of captives in vehicles

mrschick:
- Allow disarming of captives in vehicles, the items will then be dropped in the vehicle's inventory;
- Allow blindfolding or removing the blindfold from captives in vehicles;

This is somewhat of a continuation of #9361.
To allow manipulating captives in vehicles I removed some `isNull objectParent _captive` exit conditions that check if the captives are not in any vehicle. Not sure what they were needed for and if they shouldn't be removed, but everything seems to work fine from my testing.

### DIFF
--- a/addons/captives/functions/fnc_doBlindfoldCaptive.sqf
+++ b/addons/captives/functions/fnc_doBlindfoldCaptive.sqf
@@ -69,6 +69,12 @@ if (_state) then { // Blindfold target
 // Handle for things that need to be dropped to the ground
 if (_dropGoggles) then {
     private _weaponHolder = nearestObject [_target, "WeaponHolder"];
+
+    // if _target is in a vehicle, use vehicle inventory as container
+    if ((vehicle _target) != _target) then {
+        _weaponHolder = vehicle _target;
+    };
+
     if (isNull _weaponHolder || {_target distance _weaponHolder > 2}) then {
         _weaponHolder = createVehicle ["GroundWeaponHolder", [0, 0, 0], [], 0, "NONE"];
         _weaponHolder setPosASL getPosASL _target;
--- a/addons/captives/functions/fnc_doBlindfoldCaptive.sqf
+++ b/addons/captives/functions/fnc_doBlindfoldCaptive.sqf
@@ -69,13 +69,13 @@ if (_state) then { // Blindfold target
 // Handle for things that need to be dropped to the ground or in a vehicle inventory
 if (_dropGoggles) then {
     private _weaponHolder = nearestObject [_target, "WeaponHolder"];
-
     // if _target is in a vehicle, use vehicle inventory as container
-    if ((vehicle _target) != _target) then {
-        _weaponHolder = vehicle _target;
+    private _inVehicle = !isNull objectParent _target;
+    if (_inVehicle) then {
+        _weaponHolder = objectParent _target;
     };
 
-    if (isNull _weaponHolder && {_target distance _weaponHolder > 2}) then {
+    if (!_inVehicle && {isNull _weaponHolder || {_target distance _weaponHolder > 2}}) then {
         _weaponHolder = createVehicle ["GroundWeaponHolder", [0, 0, 0], [], 0, "NONE"];
         _weaponHolder setPosASL getPosASL _target;
     };
--- a/addons/captives/functions/fnc_doBlindfoldCaptive.sqf
+++ b/addons/captives/functions/fnc_doBlindfoldCaptive.sqf
@@ -66,7 +66,7 @@ if (_state) then { // Blindfold target
     removeGoggles _target;
 };
 
-// Handle for things that need to be dropped to the ground
+// Handle for things that need to be dropped to the ground or in a vehicle inventory
 if (_dropGoggles) then {
     private _weaponHolder = nearestObject [_target, "WeaponHolder"];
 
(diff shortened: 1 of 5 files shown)


addons/captives/functions/fnc_doBlindfoldCaptive.sqf
```diff
@@ -69,6 +69,12 @@ if (_state) then { // Blindfold target
 // Handle for things that need to be dropped to the ground
 if (_dropGoggles) then {
     private _weaponHolder = nearestObject [_target, "WeaponHolder"];
+
+    // if _target is in a vehicle, use vehicle inventory as container
+    if ((vehicle _target) != _target) then {
+        _weaponHolder = vehicle _target;
+    };
+
     if (isNull _weaponHolder || {_target distance _weaponHolder > 2}) then {
```
PabstMirror, line 78: I think this distance check could fail on big vics?

mrschick: Does this check distance? Doesn't it just return the vehicle a unit is in?

LinkIsGrim: IIRC it'll return model center which could be more than 2m away from the passenger.

mrschick: I'm still confused, isn't all that matters that `_weaponHolder` is assigned a "vehicle" which has an inventory that can be filled, be it a `GroundWeaponHolder` or any other enterable vehicle?

LinkIsGrim: There's a distance check on line 78. That's what Pabst mentioned.

`_target distance vehicle _target` might return a number higher than 2 depending on `_target`'s position relative to the center of the vehicle's model, in which case the distance check could fail.

PabstMirror: `(crew vehicle player) apply {_x distance vehicle _x}`
in the vanilla chinook, last 2 rows (first to load captives) are `3.1073` and `2.63218`

_target = the captive
_weaponHolder = their vehicle

so when `_target distance _weaponHolder > 2` is checked, it will return true
this means it will create a brand-new weapon holder, inside of the vehicle

mrschick: > There's a distance check on line 78. That's what Pabst mentioned.

Ah, yes. I had a feeling that's what was being referred to, but since only my change is highlighted by the review it didn't click.

> so when `_target distance _weaponHolder > 2` is checked, it will return true
> this means it will create a brand-new weapon holder, inside of the vehicle

Got it, when writing the change I thought that the first check at line 78 would prevent that issue, overlooking that it's a `||` 😆

Changing that to `&&` should be sufficient, no?

mrschick: Fixed it.

addons/captives/functions/fnc_doBlindfoldCaptive.sqf
```diff
@@ -66,10 +66,16 @@ if (_state) then { // Blindfold target
...
 if (_dropGoggles) then {
     private _weaponHolder = nearestObject [_target, "WeaponHolder"];
-    if (isNull _weaponHolder || {_target distance _weaponHolder > 2}) then {
+
+    // if _target is in a vehicle, use vehicle inventory as container
+    if ((vehicle _target) != _target) then {
+        _weaponHolder = vehicle _target;
+    };
+
+    if (isNull _weaponHolder && {_target distance _weaponHolder > 2}) then {
```
LinkIsGrim, line 78:
```suggestion
    // if _target is in a vehicle, use vehicle inventory as container
    private _inVehicle = !isNull objectParent _target;
    if (_inVehicle) then {
        _weaponHolder = objectParent _target;
    };

    if (!_inVehicle && {isNull _weaponHolder || {_target distance _weaponHolder > 2}}) then {
```

addons/disarming/functions/fnc_disarmDropItems.sqf
```diff
@@ -39,6 +39,11 @@ if (_doNotDropAmmo && {({_x in _listOfItemsToRemove} count (magazines _target))
 
 private _holder = objNull;
 
+// if _target is in a vehicle, use vehicle inventory as container
+if ((vehicle _target) != _target) then {
+    _holder = vehicle _target;
```
LinkIsGrim, line 44:
```suggestion
if (!isNull objectParent _target) then {
    _holder = objectParent _target;
```

LinkIsGrim (changes requested):

addons/captives/functions/fnc_doBlindfoldCaptive.sqf
LinkIsGrim: Fix is in review. Changing the condition from OR to AND doesn't handle being more than 2m away from a valid, found weaponHolder.

LinkIsGrim (approved):

addons/captives/functions/fnc_doBlindfoldCaptive.sqf
mrschick: Ah, indeed. Good catch.

PR #9467, opened 2023-10-06
Medical GUI - Peek Medical Info on Hit

amsteadrayle:
- Temporarily show the player's patient information display when hit

### IMPORTANT

### DIFF
--- a/addons/medical_gui/XEH_postInit.sqf
+++ b/addons/medical_gui/XEH_postInit.sqf
@@ -81,7 +81,7 @@ GVAR(selfInteractionActions) = [];
 
 [QEGVAR(medical,woundReceived), {
     params ["_unit", "_allDamages", ""];
-    if !(GVAR(peekMedicalOnHit) && ACE_player == _unit) exitWith {};
+    if !(GVAR(peekMedicalOnHit) && {_unit == ACE_player}) exitWith {};
 
     private _bodypart = toLower (_allDamages select 0 select 1);
     private _bodypartIndex = ALL_BODY_PARTS find _bodypart;
--- a/addons/medical_gui/XEH_postInit.sqf
+++ b/addons/medical_gui/XEH_postInit.sqf
@@ -88,9 +88,9 @@ GVAR(selfInteractionActions) = [];
 
     [ACE_player, _bodypartIndex] call FUNC(displayPatientInformation);
 
-    if (CBA_missionTime - GVAR(peekOnHitLastOpenedOn) > 3) then {
+    if (CBA_missionTime - GVAR(peekOnHitLastOpenedOn) > GVAR(peekMedicalOnHitDuration)) then {
         [{
-            CBA_missionTime - GVAR(peekOnHitLastOpenedOn) > 3
+            CBA_missionTime - GVAR(peekOnHitLastOpenedOn) > GVAR(peekMedicalOnHitDuration)
         }, {QGVAR(RscPatientInfo) cutFadeOut 0.3}] call CBA_fnc_waitUntilAndExecute;
     };
     GVAR(peekOnHitLastOpenedOn) = CBA_missionTime;
--- a/addons/medical_gui/XEH_postInit.sqf
+++ b/addons/medical_gui/XEH_postInit.sqf
@@ -79,16 +79,19 @@ GVAR(selfInteractionActions) = [];
     QGVAR(RscPatientInfo) cutFadeOut 0.3;
 }] call CBA_fnc_addEventHandler;
 
-ACE_player addEventHandler
-["Hit",
-    {
-        [ACE_player, 0] call FUNC(displayPatientInformation);
+[QEGVAR(medical,woundReceived), {
+    params ["_unit", "_allDamages", ""];
+    if !(ACE_PLAYER == _unit) exitWith {};
 
-        if (CBA_missionTime - GVAR(peekOnHitLastOpenedOn) > 3) then {
-            [{
-                CBA_missionTime - GVAR(peekOnHitLastOpenedOn) > 3
-            }, {QGVAR(RscPatientInfo) cutFadeOut 0.3}] call CBA_fnc_waitUntilAndExecute;
-        };
-        GVAR(peekOnHitLastOpenedOn) = CBA_missionTime;
-    }
-];
+    private _bodypart = toLower (_allDamages select 0 select 1);
+    private _bodypartIndex = ALL_BODY_PARTS find _bodypart;
+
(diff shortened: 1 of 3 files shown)


amsteadrayle:
Like with my other PR, the goal with this is to provide the player with immediate feedback about their medical condition without interrupting gameplay.

Is `Hit` the best event to attach this to? Would `HandleDamage` or `medical_injured` be different or better in any way?

TODO:

- [ ] Add setting to enable/disable
- [ ] Add setting for duration
- [ ] Fix flickering on subsequent hits before fadeout finishes

- I think this can be fixed by changing the `CBA_fnc_waitAndExecute` to `CBA_fnc_waitUntilAndExecute` and checking the time elapsed since the last hit. Is there an existing, accepted implementation of that? Would there be a problem with multiple identical `CBA_fnc_waitUntilAndExecute` in this situation?

- [ ] Show damaged body part

- Is there a good way to do this? I've searched around a bunch and I haven't been able to figure anything out that would be straightforward. This is more of a hopeful goal at the moment.

LinkIsGrim:
Event should be a matter of preference (as long as execution order is right). Tying to medical's injured event would be most appropriate and consistent with the rest of the components.

amsteadrayle:
This now uses the `medical_woundReceived` event to know about the damaged body part.

Here's a quick demo, tweaked to still show my own info when remote controling a unit.
https://www.youtube.com/watch?v=VGBjroxGD70

addons/medical_gui/XEH_postInit.sqf
```diff
@@ -76,3 +78,20 @@ GVAR(selfInteractionActions) = [];
 ["ace_interactMenuClosed", {
     QGVAR(RscPatientInfo) cutFadeOut 0.3;
 }] call CBA_fnc_addEventHandler;
+
+[QEGVAR(medical,woundReceived), {
+    params ["_unit", "_allDamages", ""];
+    if !(GVAR(peekMedicalOnHit) && ACE_PLAYER == _unit) exitWith {};
```
rautamiekka, line 84:
```suggestion
    if !(GVAR(peekMedicalOnHit) && ACE_player == _unit) exitWith {};
```

addons/medical_gui/XEH_postInit.sqf
```diff
@@ -76,3 +78,20 @@ GVAR(selfInteractionActions) = [];
...
+
+    private _bodypart = toLower (_allDamages select 0 select 1);
```
rautamiekka, line 86: There're only ANSI chars, right ?
```suggestion
    private _bodypart = toLowerANSI (_allDamages select 0 select 1);
```

rautamiekka (changes requested):

addons/medical_gui/XEH_postInit.sqf
amsteadrayle: Probably, but I was just copying the existing usage from here:
https://github.com/acemod/ACE3/blob/78bb0514d1233712428da023cdc732c037ed8aeb/addons/medical_damage/functions/fnc_woundsHandlerBase.sqf#L44

rautamiekka: Ya, seen exactly that one before and it annoys me a lot.

LinkIsGrim: Why this? We don't do it anywhere else.

addons/medical_gui/XEH_postInit.sqf
```diff
@@ -76,3 +78,20 @@ GVAR(selfInteractionActions) = [];
 ["ace_interactMenuClosed", {
     QGVAR(RscPatientInfo) cutFadeOut 0.3;
 }] call CBA_fnc_addEventHandler;
+
+[QEGVAR(medical,woundReceived), {
+    params ["_unit", "_allDamages", ""];
+    if !(GVAR(peekMedicalOnHit) && ACE_player == _unit) exitWith {};
```
LinkIsGrim, line 84:
```suggestion
    if !(GVAR(peekMedicalOnHit) && {_unit == ACE_player}) exitWith {};
```

addons/medical_gui/initSettings.sqf
```diff
@@ -126,3 +126,21 @@ private _categoryColors = [ELSTRING(medical,Category), format ["| %1 |", LELSTRI
...
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(peekMedicalOnHit),
+    "CHECKBOX",
+    [LSTRING(PeekMedicalOnHit_DisplayName), LSTRING(PeekMedicalOnHit_Description)],
+    [ELSTRING(medical,Category), LSTRING(SubCategory)],
+    true,
+    false // isGlobal
+] call CBA_fnc_addSetting;
```
LinkIsGrim, line 137: This should not be local IMO.

Some might prefer defaulting to off.

amsteadrayle: I set it local to allow servers to leave it to player preference. Particularly thinking about if somone wants to disable it while a server allows it. But I don't feel too strongly about it.

I do agree with defaulting to off, on second thought.

veteran29: Local can always be overwritten.

LinkIsGrim:
Looks good, solve merge conflicts.

addons/medical_gui/initSettings.sqf
```diff
@@ -135,3 +135,21 @@ private _categoryColors = [ELSTRING(medical,Category), format ["| %1 |", LELSTRI
...
+    false,
+    false // isGlobal
+] call CBA_fnc_addSetting;
+
+[
+    QGVAR(peekMedicalOnHitDuration),
+    "TIME",
+    [LSTRING(PeekMedicalOnHitDuration_DisplayName), LSTRING(PeekMedicalOnHitDuration_Description)],
+    [ELSTRING(medical,Category), LSTRING(SubCategory)],
+    [1, 10, 3],
```
LinkIsGrim, line 153:
```suggestion
    [0, 5, 1],
```

LinkIsGrim (approved):

amsteadrayle:
Those setting limits were intentional. 0 seconds is too short to actually parse any information, and 1 second is still very short. The upper limit being 10 seconds was just an extreme in case anyone felt like looking for that long.
The timing is different from the manual peek because you don't control when or how long it appears.

LinkIsGrim:
There's no harm in reducing the lower end of the time range.

10 seconds is too much time IMO. 5 is an eternity during a firefight already.

I guess default time could've been kept as it was but since this feature is off by default, anyone using it will be looking at settings anyway.

PR #9469, opened 2023-10-07
Medical GUI - Display qualitative bleed rate in injury list

amsteadrayle:
- Specify the severity of bleeding in the injury list with an adjective and color

### IMPORTANT

### DIFF
--- a/addons/medical_gui/functions/fnc_updateInjuryList.sqf
+++ b/addons/medical_gui/functions/fnc_updateInjuryList.sqf
@@ -73,18 +73,22 @@ if (GVAR(showDamageEntry)) then {
 // Indicate if unit is bleeding at all
 if (IS_BLEEDING(_target)) then {
     // Give a qualitative description of the rate of bleeding
-    private _bleed_rate = GET_BLOOD_LOSS(_target);
+    private _cardiacOutput = [_target] call EFUNC(medical_status,getCardiacOutput);
+    private _bleedRate = GET_BLOOD_LOSS(_target);
+    private _bleedRateKO = BLOOD_LOSS_KNOCK_OUT_THRESHOLD * (_cardiacOutput max 0.05);
+    // Use nonzero minimum cardiac output to prevent all bleeding showing as massive during cardiac arrest
+
     switch (true) do {
-        case (_bleed_rate <= BLEED_RATE_SLOW): {
+        case (_bleedRate < _bleedRateKO * BLEED_RATE_SLOW): {
             _entries pushBack [localize LSTRING(Bleed_Rate1), [1, 1, 0, 1]];
         };
-        case (_bleed_rate <= BLEED_RATE_MODERATE): {
+        case (_bleedRate < _bleedRateKO * BLEED_RATE_MODERATE): {
             _entries pushBack [localize LSTRING(Bleed_Rate2), [1, 0.67, 0, 1]];
         };
-        case (_bleed_rate <= BLEED_RATE_SEVERE): {
+        case (_bleedRate < _bleedRateKO * BLEED_RATE_SEVERE): {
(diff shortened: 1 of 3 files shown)


amsteadrayle:
The goal with this is to indicate to the player how immediate of a threat the bleeding is. In my experience, with the combination of the visual feedback (blood on screen) and the injury list simply saying "Bleeding" in bright red, I tend to treat all bleeding as an urgent problem. This change will help to clarify when bleeding is an urgent problem and when it can be handled later.

I roughly based the bleed rate thresholds on the amount of time to reach fatal blood loss and the intended situation.
| Label | Time to fatal | Description |
|--------|--------|--------|
| Slow | >8 minutes | Not an immediate concern |
| Moderate | 3-8 minutes | Important but can be handled by one person |
| Severe | 1-3 minutes | Critical and likely requiring treatment from multiple people |
| Massive | <1 minute | Probably going to die |

These thresholds are just what feel right to me, so input is appreciated. These could definitely be different based on a group's settings and playstyle. Should they be exposed as settings? (Of course, all of these would be massive bleeding in real life. This is just estimating for the time scale of gameplay.)

https://github.com/acemod/ACE3/assets/2516219/d4070c0e-725d-4a90-ab1c-8eebb3499c6c

LinkIsGrim:
Would prefer a setting, and potentially tying to cardiac output * blood loss knock out threshold, so it can serve as a visual indicator of https://github.com/acemod/ACE3/blob/master/addons/medical_status/functions/fnc_hasStableVitals.sqf#L23-L25

amsteadrayle:
Basing it on the blood loss knock out threshold makes sense, for example with massive bleeding being 100% of the threshold, severe being 50%, etc. But I think tying it to cardiac output would cause problems when the cardiac output is very low or zero, where any bleeding at all would register as massive bleeding.

Is there a recommendation for how to organize the settings? Is there a good way to do a group of settings where, for example, the second must be higher than the first but lower than the third?

LinkIsGrim:
Settings will be displayed in the order they were added, so any prioritization has be done manually. Not really any other way to group them other than the header, that I'm aware of.

amsteadrayle:
The bleed rate thresholds are now relative to the thresholds that determine "unstable" and "critical" vitals.
| Label | % of critical | Description |
|--------|--------|--------|
| Slow | <10% | Not an immediate concern |
| Moderate | 10-50% | Important but can be handled by one person |
| Severe | 50-100% | Urgent and likely requiring treatment from multiple people |
| Massive | >100% | High chance of death |

The thresholds also take cardiac output into account, so the same wounds should register the same severity of bleeding regardless of heart function.

LinkIsGrim (approved):

LinkIsGrim:
merge conflicts again, you know the drill

addons/medical_gui/functions/fnc_updateInjuryList.sqf
```diff
@@ -24,7 +24,26 @@ private _nonissueColor = [1, 1, 1, 0.33];
...
+        case (_bleedRate < _bleedRateKO * BLEED_RATE_SLOW): {
+            _entries pushBack [localize LSTRING(Bleed_Rate1), [1, 1, 0, 1]];
+        };
+        case (_bleedRate < _bleedRateKO * BLEED_RATE_MODERATE): {
+            _entries pushBack [localize LSTRING(Bleed_Rate2), [1, 0.67, 0, 1]];
+        };
+        case (_bleedRate < _bleedRateKO * BLEED_RATE_SEVERE): {
+            _entries pushBack [localize LSTRING(Bleed_Rate3), [1, 0.33, 0, 1]];
+        };
+        case (_bleedRate >= _bleedRateKO * BLEED_RATE_SEVERE): {
```
PabstMirror, line 43:
```suggestion
        default {
```

PR #9470, opened 2023-10-07
Arsenal - Add stats for reloadable launchers and belt linking

PabstMirror:
I guess part bug report: where is the `800 rpm` from?

### DIFF
--- a/addons/reload/ACE_Arsenal_Stats.hpp
+++ b/addons/reload/ACE_Arsenal_Stats.hpp
@@ -1,0 +1,13 @@
+class EGVAR(arsenal,stats) {
+    class statBase;
+    class ACE_isBelt: statBase {
+        scope = 2;
+        priority = -1;
+        stats[] = {"ACE_isBelt"};
+        displayName = CSTRING(LinkBelt);
+        showText = 1;
+        textStatement = QUOTE(localize QUOTE(ELSTRING(Common,Enabled)));
+        condition = QUOTE(params[ARR_2('_stat', '_config')]; (getNumber (_config >> _stat select 0)) == 1);
+        tabs[] = {{}, {4}};
+    };
+};
--- a/addons/reloadlaunchers/ACE_Arsenal_Stats.hpp
+++ b/addons/reloadlaunchers/ACE_Arsenal_Stats.hpp
@@ -1,0 +1,13 @@
+class EGVAR(arsenal,stats) {
+    class statBase;
+    class ADDON: statBase {
+        scope = 2;
+        priority = -1;
+        stats[] = {QGVAR(enabled)};
+        displayName= CSTRING(LoadLauncher);
+        showText = 1;
+        text = ECSTRING(Common,Enabled);
+        condition = QUOTE(params[ARR_2('_stat', '_config')]; (getNumber (_config >> _stat select 0)) == 1);
+        tabs[] = {{2}, {}};
+    };
+};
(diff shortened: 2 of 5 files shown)


addons/reloadlaunchers/ACE_Arsenal_Stats.hpp
```diff
@@ -0,0 +1,13 @@
+class EGVAR(arsenal,stats) {
+    class statBase;
+    class ADDON: statBase {
+        scope = 2;
+        priority = -1;
+        stats[] = {QGVAR(enabled)};
+        displayName= CSTRING(LoadLauncher);
```
rautamiekka, line 7:
```suggestion
        displayName = CSTRING(LoadLauncher);
```

rautamiekka (changes requested):

johnb432:
> I guess part bug report: where is the `800 rpm` from? 

Your config is invalid. `text` is not supported by the [arsenal framework](https://github.com/acemod/ACE3/blob/master/docs/wiki/framework/arsenal-framework.md#51-adding-stats-via-config).

However, it might not be wise to return some other random garbage if nothing is found, although personally I don't think it's a big issue.

addons/reloadlaunchers/ACE_Arsenal_Stats.hpp
```diff
@@ -0,0 +1,13 @@
...
+        showText = 1;
+        text = ECSTRING(Common,Enabled);
```
johnb432, line 9:
```suggestion
        textStatement = QUOTE(localize QUOTE(ELSTRING(Common,Enabled)));
```

addons/reloadlaunchers/ACE_Arsenal_Stats.hpp
```diff
@@ -0,0 +1,13 @@
+class EGVAR(arsenal,stats) {
+    class statBase;
+    class ADDON: statBase {
+        scope = 2;
+        priority = -1;
+        stats[] = {QGVAR(enabled)};
+        displayName = CSTRING(LoadLauncher);
```
LinkIsGrim, line 7: This could easily be misinterpreted as an indication of the launcher being disposable. Alternate stringtable maybe? (`Buddy loading`?)

addons/reload/ACE_Arsenal_Stats.hpp
```diff
@@ -0,0 +1,13 @@
+class EGVAR(arsenal,stats) {
+    class statBase;
+    class ACE_isBelt: statBase {
+        scope = 2;
+        priority = -1;
+        stats[] = {"ACE_isBelt"};
+        displayName = CSTRING(LinkBelt);
+        showText = 1;
+        textStatement = QUOTE(localize QUOTE(ELSTRING(Common,Enabled)));
```
LinkIsGrim, line 9:
```suggestion
        textStatement = ECSTRING(Common,Enabled);
```

Will this work?

PabstMirror: it will look like 
`call compile "ace_common_x"`
and return nil, which is why I got the "800 rpm" weirdness

PabstMirror: arsenal action does 
```
        private _text = getText (_x >> "text");
        private _textStatement = getText (_x >> "textStatement");

        private _type = switch (false) do {
            case (_text == ""): {
                _statement = format ["""%1""", _text];
                ACTION_TYPE_TEXT
            };
            case (_textStatement == ""): {
                _statement = _textStatement;
                ACTION_TYPE_TEXT
            };
            case (_statement == ""): {
                _statement = _statement;
                ACTION_TYPE_BUTTON
            };
            default {
                -1
            };
        };
```
which sorta makes both compatible, but I think johnb432's code is fine

LinkIsGrim (approved):
Note to future self: add `text` property to stats.

PR #9474, opened 2023-10-09
Medical GUI - Show number of available treatment items in medical menu

amsteadrayle:
**When merged this pull request will:**
- Display the number of treatment items available on each treatment button in the medical menu 

### IMPORTANT

- [ ] If the contribution affects [the documentation](https://github.com/acemod/ACE3/tree/master/docs), please include your changes in this pull request so the documentation will appear on the [website](https://ace3.acemod.org/).
- [x] [Development Guidelines](https://ace3.acemod.org/wiki/development/) are read, understood and applied.
- [x] Title of this PR uses our standard template `Component - Add|Fix|Improve|Change|Make|Remove {changes}`.

PR #9475, opened 2023-10-10
Medical GUI - Show warning if tourniquet will interfere with medical menu action

amsteadrayle:
- Show a warning tooltip if a tourniquet will interfere with a medical menu action

### IMPORTANT


amsteadrayle:
https://github.com/acemod/ACE3/assets/2516219/a41a8e70-9e43-4f50-a774-096b05afc7db

To be clear, this does not prevent or obstruct medical actions. It only adds a tooltip to make the player aware of the situation.

This overlaps with #9474, and the intent is that this tooltip will override the item count tooltip.

addons/medical_gui/functions/fnc_updateActions.sqf
```diff
@@ -50,6 +50,16 @@ private _shownIndex = 0;
         _ctrl ctrlSetPositionY POS_H(1.1 * _shownIndex);
         _ctrl ctrlCommit 0;
 
+        // Show warning if tourniquet will interfere with action
+        if (((_category in ["examine", "medication"]) || (_items findIf {"IV" in _x}) > -1)
```
johnb432, line 54: Isn't there an IV category? The `_items findIf {"IV" in _x}) > -1` check is too specific imo.

amsteadrayle: IVs are in the advanced category, shared with things like stitching and CPR. Checking the item name was the most succinct way I could figure out without having to pass extra information about the actions.

amsteadrayle:
I've elected to not add a setting for this. To me, the warning is universal and low-impact enough to just have it always on. If you all disagree, then I'm fine with adding a setting. I just didn't want to go ahead and add yet another setting if it wasn't needed.

(By universal I mean that it's applicable regardless of what settings someone might use, as far as I'm aware.)

LinkIsGrim:
I'm not sure on this one. It is definitely a UX improvement for casual players, but it feels like giving up one of the finer details that makes the medical system unique (even if it's still there, some of the magic is gone with this being shown).

I think this isn't necessary with the addition of the outline, honestly.

amsteadrayle:
That does makes sense for the people who are interested in engaging with a detailed medical system - making them pay closer attention. But this warning would be a good way to teach new and casual players about one of the more subtle and otherwise invisible mistakes that can be made. Especially since that mistake can directly lead to player death and a generally negative experience.

I'd like it as a setting that's on by default and not global.

jonpas:
If you want to teach a new player something, it could be a one-time thing. But I don't think we have anything else like that in ACE3.

amsteadrayle:
I shouldn't have worded it as teaching. Thinking on it more, it's about having a user interface where, if a button is not going to do what it normally does, the user should be told that.

Drofseh:
Do this and the tooltip for # of items interfere/overlap or anything?

jonpas:
> I shouldn't have worded it as teaching. Thinking on it more, it's about having a user interface where, if a button is not going to do what it normally does, the user should be told that.

I understand, but in this case, the user can do it, as you can do it in real life. That button will do just fine what it says it will do. The resulting effects are a different story and have nothing to do with this UX in particular.

amsteadrayle:
> Do this and the tooltip for # of items interfere/overlap or anything?

This warning will override the tooltip for # of items.

amsteadrayle:
This should be all set to merge now, whether or not you decide to add it.

addons/medical_gui/stringtable.xml
```diff
@@ -1348,5 +1348,14 @@
         <Key ID="STR_ACE_Medical_GUI_TreatmentItemCount_Vehicle">
             <English>in vehicle's inventory</English>
         </Key>
+        <Key ID="STR_ACE_Medical_GUI_TourniquetWarning">
+            <English>No effect while tourniqeted</English>
```
Drofseh, line 1352: I think phrasing it as `No effect until tourniqet removed` might be better.
You can still perform the action, and it will have an effect, it just gets delayed until after the TQ is taken off.

addons/medical_gui/functions/fnc_updateActions.sqf
```diff
@@ -58,8 +58,17 @@ private _shownIndex = 0;
...
+        _ctrl ctrlSetTooltipColorText [1, 1, 1, 1];
         _ctrl ctrlSetTooltip _countText;
 
+        // Show warning if tourniquet will interfere with action
+        if (GVAR(tourniquetWarning)
+        && {(_category in ["examine", "medication"]) || (_items findIf {"IV" in _x}) > -1}
+        && {HAS_TOURNIQUET_APPLIED_ON(GVAR(target),GVAR(selectedBodyPart))}) then {
+            _ctrl ctrlSetTooltipColorText [1, 1, 0, 1];
+            _ctrl ctrlSetTooltip localize LSTRING(TourniquetWarning);
+        }
```
LinkIsGrim, line 70:
```suggestion
        if (
            GVAR(tourniquetWarning) &&
            {(_category in ["examine", "medication"]) || (_items findIf {"IV" in _x}) > -1} &&
            {HAS_TOURNIQUET_APPLIED_ON(GVAR(target),GVAR(selectedBodyPart))}
        ) then {
            _ctrl ctrlSetTooltipColorText [1, 1, 0, 1];
            _ctrl ctrlSetTooltip localize LSTRING(TourniquetWarning);
        };
```

code style

addons/medical_gui/initSettings.sqf
```diff
@@ -162,3 +162,12 @@ private _categoryColors = [ELSTRING(medical,Category), format ["| %1 |", LELSTRI
...
     false
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(tourniquetWarning),
+    "CHECKBOX",
+    [LSTRING(TourniquetWarning_DisplayName), LSTRING(TourniquetWarning_Description)],
+    [ELSTRING(medical,Category), LSTRING(SubCategory)],
+    false,
+    false
```
LinkIsGrim, line 172:
```suggestion
    true
```
Global

addons/medical_gui/functions/fnc_updateActions.sqf
```diff
@@ -58,8 +58,17 @@ private _shownIndex = 0;
...
         };
+        _ctrl ctrlSetTooltipColorText [1, 1, 1, 1];
         _ctrl ctrlSetTooltip _countText;
 
+        // Show warning if tourniquet will interfere with action
+        if (GVAR(tourniquetWarning)
+        && {(_category in ["examine", "medication"]) || (_items findIf {"IV" in _x}) > -1}
+        && {HAS_TOURNIQUET_APPLIED_ON(GVAR(target),GVAR(selectedBodyPart))}) then {
+            _ctrl ctrlSetTooltipColorText [1, 1, 0, 1];
+            _ctrl ctrlSetTooltip localize LSTRING(TourniquetWarning);
```
LinkIsGrim, line 69:
```suggestion
            _ctrl ctrlSetTooltip LLSTRING(TourniquetWarning);
```
localize macro

addons/medical_gui/functions/fnc_updateActions.sqf
```diff
@@ -58,8 +58,17 @@ private _shownIndex = 0;
             private _counts = [_items] call FUNC(countTreatmentItems);
             _countText = _counts call FUNC(formatItemCounts);
         };
+        _ctrl ctrlSetTooltipColorText [1, 1, 1, 1];
         _ctrl ctrlSetTooltip _countText;
 
+        // Show warning if tourniquet will interfere with action
+        if (GVAR(tourniquetWarning)
+        && {(_category in ["examine", "medication"]) || (_items findIf {"IV" in _x}) > -1}
```
LinkIsGrim, line 66:
```suggestion
        && {(_category in ["examine", "medication"]) || (_items findIf {"IV" in _x}) > -1}
```

Still don't like the IV check, but anything else requires config lookups/caching items.

LinkIsGrim (changes requested):

addons/medical_gui/initSettings.sqf
amsteadrayle: Oh really? It seems like something that would be left to the player unless admins specifically care. That's my assumption for UI stuff in general.

johnb432: > Oh really? It seems like something that would be left to the player unless admins specifically care. That's my assumption for UI stuff in general.

I agree with this. @LinkIsGrim Why would you want to enforce it globally?

addons/medical_gui/functions/fnc_updateActions.sqf
johnb432: I don't really have the means to properly investigate the solution I am proposing, but maybe if https://github.com/acemod/ACE3/blob/46473601d60a674f9d6c8ff8d02b1b535d1aea01/addons/medical_gui/functions/fnc_updateActions.sqf#L41 gives the statement defined e.g. here https://github.com/acemod/ACE3/blob/f774eaf787f32670926259f470abd5b0303d0357/addons/medical_treatment/ACE_Medical_Treatment_Actions.hpp#L157 we could use that instead?

addons/medical_gui/initSettings.sqf
LinkIsGrim: Given the effect, I'd rather this be *really* deliberate on when it's enabled.

That being said, vox populi and all that.

addons/medical_gui/functions/fnc_updateActions.sqf
LinkIsGrim: It's still specific, but the medical system wasn't designed with this in mind (and I don't see a good way to implement it). It's fine. I think most 3rd party IVs will have "IV" in their classname.

LinkIsGrim (approved):

addons/medical_gui/functions/fnc_updateActions.sqf
rautamiekka: If anything, enforcing the " IV" (notice the leading space) name requirement would be 1 way to ensure support with ACE outside config, assuming the 3rd-party dev gives a fuck.

addons/medical_gui/initSettings.sqf
Drofseh: I don't think this needs to be global. Communits admins can force if they want to.

LinkIsGrim:
we ball²

PR #9320, opened 2023-08-12
Field Manual - Add Framework

hypoxia125:
-- Created framework for addition of the Field Manual --

Created %PBOPREFIX%
Created CfgHintCategories.hpp
Created CfgHints.hpp
Created config.cpp
Created script_component.hpp

Will continue to add parts in additional PR - small incriments
closes #3416

PR #9477, opened 2023-10-12
Medical GUI - Add keyboard shortcuts for selecting medical menu categories and body parts

amsteadrayle:
- Set up keyboard handling for medical menu
- Add keyboard shortcuts to select categories and body parts

### IMPORTANT

### DIFF
--- a/addons/medical_gui/functions/fnc_onKeyDown.sqf
+++ b/addons/medical_gui/functions/fnc_onKeyDown.sqf
@@ -1,0 +1,88 @@
+#include "..\script_component.hpp"
+#include "\a3\ui_f\hpp\defineDIKCodes.inc"
+/*
+ * Author: AmsteadRayle
+ * Handles keyboard inputs in medical menu.
+ *
+ * Arguments:
+ * 0: Menu display <DISPLAY>
+ * 1: Key being pressed <NUMBER>
+ * 2: Shift state <BOOL>
+ * 3: Ctrl state <BOOL>
+ * 4: Alt state <BOOL>
+ *
+ * Return Value:
+ * None
+ *
+ * Public: No
+*/
+
+#define NUMBER_KEYS [DIK_1, DIK_2, DIK_3, DIK_4, DIK_5, DIK_6, DIK_7, DIK_8, DIK_9, DIK_0]
+
+params ["", "_args"];
+_args params ["_display", "_keyPressed", "_shiftState", "_ctrlState", "_altState"];
+
+private _return = true;
+
+private _fnc_isCategoryVisible = {
+    params ["_category"];
+    (GVAR(actions) findIf {_category == _x select 1}) > -1
+};
+
+private _visibleCategories = [
+    "bandage","medication","airway","advanced","drag"
+] select {[_x] call _fnc_isCategoryVisible};
+
+private _allCategories = ["triage", "examine"] + _visibleCategories + ["toggle"];
+
+// Use categories as keys in hashmap because there are fewer,
+// otherwise the hashmap is padded with nil
+private _keyCategoryPairs = _allCategories createHashMapFromArray NUMBER_KEYS;
+
+private _temp_category = "";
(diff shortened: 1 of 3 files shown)


amsteadrayle:
A couple questions about this:
Is the airway category ever actually used? I wasn't able to find anything that would enable it. There's a bunch of extra work in this to handle a variable number of categories in case airway is enabled. If airway is definitely never going to show up, then the code can be simplified a lot.

Is it possible to rebind these kinds of keybinds? I don't think this would need rebinding, but I'm curious.

How should this functionality be communicated to the player? Would adding a simple hint to the tooltip be fine? Like "Examine Patient (2)" and "Select Left Arm (D)"

addons/medical_gui/functions/fnc_onKeyDown.sqf
```diff
@@ -0,0 +1,84 @@
...
+private _allCategories = ["triage", "examine"] + _visibleCategories + ["toggle"];
+
+// Use hashmap as a shortcut to "zip" two arrays together
+// Use categories as keys in hashmap because there are fewer,
+// otherwise the hashmap is padded with nil
+private _keyCategoryPairs = _allCategories createHashMapFromArray NUMBER_KEYS;
+
+private _temp_category = "";
+
+switch (true) do {
```
LinkIsGrim, line 44: Can't this just be `switch _keyPressed`?

amsteadrayle: Probably. But it looks like `switch (true)` allows for using modifier keys, looking at the arsenal implementation, which this is drawn from.
https://github.com/acemod/ACE3/blob/176482381ad2222502157933c0cd622f2a049da2/addons/arsenal/functions/fnc_onKeyDown.sqf#L32-L41

addons/medical_gui/functions/fnc_onKeyDown.sqf
```diff
@@ -0,0 +1,84 @@
...
+ * 2: Shift state <BOOL>
+ * 3: Ctrl state <BOOL>
+ * 4: Alt state <BOOL>
+ *
+ * Return Value:
+ * None
+ *
+ * Public: No
+*/
+// TODO: Is the airway category ever visible? Can the dynamic category stuff be removed?
```
LinkIsGrim, line 19: It will be eventually, plus this maintains support for KAM. Keep it.

addons/medical_gui/functions/fnc_onKeyDown.sqf
LinkIsGrim: Are modifier keys going to be used here though? `_keyPressed` is still limited to DIK key codes.

addons/medical_gui/functions/fnc_onKeyDown.sqf
amsteadrayle: What is KAM?

LinkIsGrim:
> Is it possible to rebind these kinds of keybinds? I don't think this would need rebinding, but I'm curious.
Yes but it's absolutely not worth it.

>How should this functionality be communicated to the player? Would adding a simple hint to the tooltip be fine? Like "Examine Patient (2)" and "Select Left Arm (D)"

We have this great [Field Manual](https://github.com/acemod/ACE3/pull/9320) thing now. A Medical Menu section there would be fitting.

amsteadrayle:
> We have this great Field Manual thing now.

Oh nice! Is that something I should be incorporating here? And would you have that be instead of the tooltip hints, or in addition to them?

LinkIsGrim:
Instead of tooltips to reduce clutter. Doing it in this PR is fine.

addons/medical_gui/functions/fnc_onKeyDown.sqf
LinkIsGrim: 3rd party medical addon

LinkIsGrim (approved):

PR #9487, opened 2023-10-13
Medical GUI - Add sound and conditional switching to Medical Menu keybinds

LinkIsGrim:
- Prevent switching to disabled or selected categories via keybinds
- Add sound feedback.

### IMPORTANT

### DIFF
--- a/addons/medical_gui/functions/fnc_onKeyDown.sqf
+++ b/addons/medical_gui/functions/fnc_onKeyDown.sqf
@@ -5,14 +5,18 @@
  * Handles keyboard inputs in medical menu.
  *
  * Arguments:
- * 0: Menu display <DISPLAY>
- * 1: Key being pressed <NUMBER>
- * 2: Shift state <BOOL>
- * 3: Ctrl state <BOOL>
- * 4: Alt state <BOOL>
+ * 1: Args <ARRAY>
+ * - 0: Menu display <DISPLAY>
+ * - 1: Key being pressed <NUMBER>
+ * - 2: Shift state <BOOL>
+ * - 3: Ctrl state <BOOL>
+ * - 4: Alt state <BOOL>
  *
  * Return Value:
  * None
+ *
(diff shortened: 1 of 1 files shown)


addons/medical_gui/functions/fnc_onKeyDown.sqf
```diff
@@ -81,4 +87,8 @@ switch (true) do {
     };
 };
 
+if (_return) then {
+    playSoundUI ["\a3\ui_f\data\sound\rscbuttonmenu\soundClick.wss", 0.09, 1];
```
LinkIsGrim, line 91: Volume will be scaled off UI volume slider and thus might not match clicking the buttons themselves. Is there a CfgSounds entry for this? (or does playSound support filenames?)

veteran29: Some BI utility script use `playSound ["SoundClick", true];`

addons/medical_gui/functions/fnc_onKeyDown.sqf
```diff
@@ -40,39 +45,44 @@ private _allCategories = ["triage", "examine"] + _visibleCategories + ["toggle"]
...
-            }
+        _temp_idc = IDC_TRIAGE + (ALL_CATEGORIES find _temp_category) * 10;
+        case (_keyPressed == _y && {GVAR(selectedCategory) != _temp_category}): {
+            if (ctrlEnabled _temp_idc) then {
+                if (_temp_category == "toggle") then {
+                    call FUNC(handleToggle);
+                } else {
+                    GVAR(selectedCategory) = _temp_category;
+                };
+            };
```
PabstMirror, line 62: If you hit a button for a category that is not enabled then it won't play a sound
e.g. Medication when head selected
```suggestion
            } else {
                _return = false;
            };
```

PabstMirror (approved):

PR #9493, opened 2023-10-15
Common - Add support for non-blocking progress bar

LinkIsGrim:
- Title.
- Make Cargo's airdropping use it.

This does not replace the current progress bar functionality (dialog).

Pressing Escape/opening a dialog/opening the interaction menu still closes the progress bar. Is there anything else that needs to be handled by this?

### IMPORTANT


amsteadrayle:
Is there a reason that the progress bar intercepts mouse movement? Not being able to freelook while in progress has been a little pet peeve of mine for a long time.

LinkIsGrim:
The progress bar is a UI dialog. Dialogs and displays intercept mouse movement. That's about it.

LinkIsGrim:
I can't find a way to avoid the fade-in, "PLAIN NOFADE" has no effect on cutRsc.

Didn't edit the function header. I'm lazy.

Interactions with animations will still block player movement even with the new param. Freelook is unaffected.

Interactions without animations will allow player movement/vehicle control freely. Anyone using this will need to embed that in condition.

Again, this has to be used explicitly. Default behavior is the dialog.

johnb432:
How well does it handle death?

veteran29 (approved):
LGTM

LinkIsGrim:
Gonna edit the header. Hold on merge till then.

LinkIsGrim:
> How well does it handle death?

Should be same as the regular progress bar handles it, but haven't tested for that specifically.

LinkIsGrim:
Note for changelog, add note on airdrop change.

addons/common/functions/fnc_progressBar.sqf
```diff
@@ -9,9 +10,10 @@
  * 1: Arguments, passed to condition, fail and finish <ARRAY>
  * 2: On Finish: Code called or STRING raised as event. <CODE, STRING>
  * 3: On Failure: Code called or STRING raised as event. <CODE, STRING>
- * 4: (Optional) Localized Title <STRING>
- * 5: Code to check each frame (Optional) <CODE>
- * 6: Exceptions for checking EFUNC(common,canInteractWith) (Optional)<ARRAY>
+ * 4: Localized Title (default: "") <STRING>
+ * 5: Code to check each frame (default: {true}) <CODE>
+ * 6: Exceptions for checking ace_common_fnc_canInteractWith (default: []) <ARRAY>
+ * 7: Create progress bar as dialog, this blocks user input (default: true) <BOOL>
```
johnb432, line 16: There are both versions in ACE, but there are about 5 times more of the type I'm suggesting:

```suggestion
 * 4: Localized Title <STRING> (default: "")
 * 5: Code to check each frame <CODE> (default: {true})
 * 6: Exceptions for checking ace_common_fnc_canInteractWith <ARRAY> (default: [])
 * 7: Create progress bar as dialog, this blocks user input <BOOL> (default: true)
```

LinkIsGrim: Yeah I thought that looked weird while typing it

addons/common/functions/fnc_progressBar.sqf
```diff
@@ -22,22 +24,34 @@
...
-_map ctrlMapCursor ["", QGVAR(blank)];
+if (_dialog) then {
+    private _map = _display displayCtrl 101;
+    _map ctrlMapCursor ["", QGVAR(blank)];
+} else { // Add key handler for ESC to cancel
+    [DIK_ESCAPE, [false, false, false], {
+        QGVAR(progressBarDisplay) cutText ["", "PLAIN"];
+        [QGVAR(progressBarKeyHandler), "keydown"] call CBA_fnc_removeKeyHandler;
+        true
+    }, "keydown", QGVAR(progressBarKeyHandler)] call CBA_fnc_addKeyHandler;
```
johnb432, line 53: They key handler is only removed if ESC is pressed. Could this be problematic if the display is closed by other means? The key handler would still be active.

LinkIsGrim: Good catch.

LinkIsGrim:
Removing the key handler twice doesn't cause any issues, thankfully.

johnb432 (approved):
Noticed during testing that pressing the "Airdrop" button would sometimes fail silently. I can't get a proper repro though,
I assume this will be merged before #9617, which will give us more time to determine what the issue is.

LinkIsGrim:
Likely failed due to flight conditions, notification should pop up though.

PR #9499, opened 2023-10-16
Arsenal - Add stat for capabilities of pointer attachments

PabstMirror:
image


LinkIsGrim:
Gonna be honest, I don't understand what the suffixex mean.

PabstMirror:
maybe going to 3 chars will help, but then some with many modes will get cut off

LinkIsGrim:
We can always make the stat window larger

PabstMirror:
For most attachment it works just fine
On the few it doesn't you still know it has a lot of modes
image

LinkIsGrim (approved):

PR #9502, opened 2023-10-16
Cargo - Add Cargo UI to the Inventory UI

jokoho48:
- Currently Adds an Additional way of checking the Cargo of Vehicles
- Adds a Dirty Workaround for all Objects to have a "Inventory" Action

grafik

### TODO
- [ ] Reopen the Correct UI depending on what was used for Unloading/Paradrop if ReopenOnX is used
- [ ] Fully Replace the other UI Element else add a Setting to toggle between or have both?
- [ ] test with Different Resolutions

### Discussion Points
Should this be a replacement or an alternative maybe with an extra Setting?
If we go for a full replacement of the current UI we need to be clear that 

### IMPORTANT

### DIFF
--- a/addons/cargo/functions/fnc_onOpenInventory.sqf
+++ b/addons/cargo/functions/fnc_onOpenInventory.sqf
@@ -1,0 +1,136 @@
+#include "..\script_component.hpp"
+/*
+ * Author: joko // Jonas, BadGuy
+ * Handles Inventory UI when vehicle has a Cargo Inventory
+ *
+ * Arguments:
+ * 0: Unit <OBJECT>
+ * 1: Container <OBJECT>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * [vehicle] call ace_cargo_fnc_addCargoItem
+ *
+ * Public: No
+ */
+
+if !(GVAR(enable)) exitWith {};
+
+params ["_unit", "_container"];
+
+// This is the 2nd Part of the Hack for Vehicles that dont have an Inventory
+if (_container isKindOf "GroundWeaponHolder") then {
+    private _target = cursorObject;
+    if (!(_target getVariable [QGVAR(hasInventory), true]) && ((ACE_Player distance _target) < MAX_LOAD_DISTANCE) && !lockedInventory _target) then {
+        _container = _target;
+    };
+};
+
+if !(_container getVariable [QGVAR(hasCargo), getNumber (configOf _container >> QGVAR(hasCargo)) == 1]) exitWith {};
+
+GVAR(interactionVehicle) = _container;
+
+GVAR(interactionParadrop) = _container isKindOf "Air" && {
+                                private _turretPath = _unit call CBA_fnc_turretPath;
+                                (_unit == (driver _container)) || // pilot
+                                {(getNumber (([_container, _turretPath] call CBA_fnc_getTurret) >> "isCopilot")) == 1} || // coPilot
+                                {_turretPath in (getArray (configOf _container >> QGVAR(loadmasterTurrets)))} // loadMaster turret from config 
+                            };
+
+[{!isNull (findDisplay 602)}, {
+    private _display = (findDisplay 602); // Find Inventory Display
+    private _gY = ((((safeZoneW / safeZoneH) min 1.2) / 1.2) / 25);
+    private _gX = (((safeZoneW / safeZoneH) min 1.2) / 40);
+
+
+    private _hasInventory = GVAR(interactionVehicle) getVariable [QGVAR(hasInventory), true];
+    private _xOffset = [1, -5.25] select (_hasInventory);
+
+    if (_hasInventory) then {
+        {
+            private _pos = ctrlPosition _x;
+
+            _x ctrlSetPosition [
+                (_pos select 0) + 6.25 * _gX,
+                (_pos select 1)
+            ];
+            _x ctrlCommit 0;
+        } forEach allControls _display;
+    } else {
+        {
+            (_display displayCtrl _x) ctrlSetFade 1;
+            (_display displayCtrl _x) ctrlCommit 0;
+        } forEach [1001, 632, 6554, 6307, 6385, 6321];
+    };
+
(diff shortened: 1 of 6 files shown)


LinkIsGrim:
I think a replacer is fine if this matches the inventory UI design (pictures and thicker line, no need for quantity).

jokoho48:
> I think a replacer is fine if this matches the inventory UI design (pictures and thicker line, no need for quantity).

well it matches as far as we can match. we don't have/can't produce all the images we would need for the inventory items for those I think so it is currently just a list similar to the current cargo UI just directly integrated in the Inventory UI

> Need to make sure it works the same though (like passengers in aircraft not being able to unload cargo, although, I've seen requests for that).

That should already be handled by those lines https://github.com/acemod/ACE3/pull/9502/files#diff-079026a24a4965fcf78d16170da41a373fa8639fc8fb424d745abb63e56ebd53R35-R40

addons/cargo/functions/fnc_onOpenInventory.sqf
```diff
@@ -0,0 +1,136 @@
...
+            _list lbAdd _displayName;
+        } forEach _cargoItems;
+
+        private _cargoCapacity = GVAR(interactionVehicle) getVariable [QGVAR(spaceMax), getNumber (configOf GVAR(interactionVehicle) >> QGVAR(space))];
+        private _usedCargoCapacity = _cargoCapacity - ([GVAR(interactionVehicle)] call FUNC(getCargoSpaceLeft));
+
+        _loadBar progressSetPosition (_usedCargoCapacity / _cargoCapacity);
+
+    }, 1, [_loadBar, _list]] call CBA_fnc_addPerFrameHandler;
+}] call CBA_fnc_waitUntilAndExecute;
```
LinkIsGrim, line 136:
```suggestion
}] call CBA_fnc_waitUntilAndExecute;

```
newline at EOF

addons/cargo/functions/fnc_onOpenInventory.sqf
```diff
@@ -0,0 +1,136 @@
...
+
+if !(_container getVariable [QGVAR(hasCargo), getNumber (configOf _container >> QGVAR(hasCargo)) == 1]) exitWith {};
+
+GVAR(interactionVehicle) = _container;
+
+GVAR(interactionParadrop) = _container isKindOf "Air" && {
+                                private _turretPath = _unit call CBA_fnc_turretPath;
+                                (_unit == (driver _container)) || // pilot
+                                {(getNumber (([_container, _turretPath] call CBA_fnc_getTurret) >> "isCopilot")) == 1} || // coPilot
+                                {_turretPath in (getArray (configOf _container >> QGVAR(loadmasterTurrets)))} // loadMaster turret from config
```
LinkIsGrim, line 39: I didn't know this was a thing. Neat.

jokoho48: copied from here: https://github.com/acemod/ACE3/blob/9873bf566ca06f5f88538a216ac0d8fb61f22245/addons/cargo/functions/fnc_initVehicle.sqf#L75-L78 ^^

jokoho48:
From my testing, having the UI Reopen automatically when the player has Carray After Unload enabled does not make sense because Reopening the Inventory forces a player Animation to happen and resets the carry animation.

jonpas:
I don't think we need to match Inventory UI exactly, as it is a different purpose.

> One more question is if we want to re-add a "text" form of the filling of the Vehicle as currently it is handled via the bar at the bottom

Only text, it makes more sense when the entire system is designed around discrete values.

I assume the loaded cargo item text is the same (including name etc.) as on the old cargo menu?

johnb432:
(Possible) bugs:

1) Are you sure that https://github.com/acemod/ACE3/pull/9502/files#diff-1282e82f9ea9eb3a3eca2f12059eb2783232bde9e0d23543c332533db0e01e67R73 includes all conditions for an object to have an inventory? Doesn't the `maximumLoad` attribute also play a role?
2) It doesn't take the UI size into account - see the `ace_inventory_inventoryDisplaySize` CBA setting.
3) Zeus module for unloading cargo is broken.

---

> I think we can keep the cargo menu action and just add the inventory display, that way we can drop the hack. I'm not a fan of it.

I'm not a fan of integrating the cargo UI into the inventory display, for the following reasons:

1) I personally *hate* the Vanilla scroll wheel menu with a passion. For me, it's clunky, easy to overshoot the desired action (probably more of a skill issue on my part though) and hard to access (I tried using the scroll wheel interaction on a Merkava - the interaction isn't available for the front third of the vehicle, which imo is suboptimal).
2) If other mods change the inventory layout/size (I'm thinking specifically Better Inventory, which I personally use a lot), it will be screwed up. Not really ACE' problem, but see below.
3) Lastly, I don't see the point of adding the cargo menu to the inventory display. You can't move items from one window to another - what's the benefit?

jokoho48:
> I'm not a fan of integrating the cargo UI into the inventory display, for the following reasons:
> 
>     1. I personally _hate_ the Vanilla scroll wheel menu with a passion. For me, it's clunky, easy to overshoot the desired action (probably more of a skill issue on my part though) and hard to access (I tried using the scroll wheel interaction on a Merkava - the interaction isn't available for the front third of the vehicle, which imo is suboptimal).
> 
>     2. If other mods change the inventory layout/size (I'm thinking specifically Better Inventory, which I personally use a lot), it will be screwed up. Not really ACE' problem, but see below.
> 
>     3. Lastly, I don't see the point of adding the cargo menu to the inventory display. You can't move items from one window to another - what's the benefit?

1. this does not just work for the Scroll wheel menu but also just when pressing `I`. I am quite the opposite the current ACE interaction menu is often quite difficult to reach and feels Clunky
2. yes that is true. i original had it implemented so that it only does add it as a additional setting. but all feedback i got was to replace the old menu
3. It makes it easier to access and is not "hidden" in an often quite difficult-to-reach Interaction point. Additionally, because it integrates into the Vanilla UI, it is Friendlier for newer Players who don't have Experience with ACE but with Arma 3 Vanilla.

PR #9513, opened 2023-10-18
Zeus - Fix trench module on clients

PabstMirror:
```
TRACE: 92958 moduleLayTrench: _logic=L Alpha 1-2:1, _logic getVariable "test"=any z\ace\addons\zeus\functions\fnc_moduleLayTrench.sqf:23
TRACE: 92962 moduleLayTrench: _logic=L ACE Utility:1, _logic getVariable "test"=92958 z\ace\addons\zeus\functions\fnc_moduleLayTrench.sqf:23
```

I put one module down on a client and get `function` runs twice for the same module, separated by ~5 frames?
I have no idea why

### DIFF
--- a/addons/zeus/functions/fnc_moduleLayTrench.sqf
+++ b/addons/zeus/functions/fnc_moduleLayTrench.sqf
@@ -27,6 +27,9 @@ if !(["ace_trenches"] call EFUNC(common,isModLoaded)) exitWith {
     [LSTRING(RequiresAddon)] call FUNC(showMessage);
 };
 
+if (_logic getVariable [QGVAR(ran), false]) exitWith { WARNING("double run?") };
+_logic setVariable [QGVAR(ran), true];
+
 private _drawCode = {
     params ["_object", "_mousePos"];
     private _startPos = getPos _object;


PabstMirror:
Might be due to `BIS_fnc_moduleInit` and `BIS_fnc_moduleExecute`
I think it's doing
```
	if (ismultiplayer && _isGlobal && !is3DEN && time > 0) then {		
		[_arguments,_function,-2] spawn BIS_fnc_MP; 
	};	
	_arguments spawn (missionnamespace getvariable _function);
```
so the server runs it once and every other client runs it twice
once normally and a 2nd time after network delay, which would match the 5 frames I'm seeing
wtf, or am I reading this wrong?

mharis001:
Yes, you are correct. BI's module framework has been bugged since the start and is the main reason why ZEN adds its own module initialization function.

veteran29:
I'm not even sure why our Zeus modules are global.

addons/zeus/functions/fnc_moduleLayTrench.sqf
```diff
@@ -27,6 +27,9 @@ if !(["ace_trenches"] call EFUNC(common,isModLoaded)) exitWith {
     [LSTRING(RequiresAddon)] call FUNC(showMessage);
 };
 
+if (_logic getVariable [QGVAR(ran), false]) exitWith { WARNING("double run?") };
```
PabstMirror, line 30:
```suggestion
if (_logic getVariable [QGVAR(ran), false]) exitWith {};
```

PabstMirror:
we need global because we need the func to run on zues's machine
I guess we can add run-guards on every zeus func
but in most cases they are deleted right away so it didn't matter

LinkIsGrim (approved):

jonpas (approved):

PR #9519, opened 2023-10-18
Documentation - Update jquery, remove polyfills, don't use foundation js

bux:
- see title
  - old jquery is being marked as bad by lighthouse
  - had to remove old foundation js, because of breaking changes in v3 compared to v2
  - browsers have matured by 7 years since introducing the polyfills...............................


bux:
the css file is a built file...
the validate task should ignore the "docs/css" & "docs/js" folders

jonpas:
> the css file is a built file... the validate task should ignore the "docs/css" & "docs/js" folders

@veteran29

bux:
@PabstMirror  manually fixing the bom is - of course -  possible :)

jonpas:
I don't think we want only `addons` checked though. :thinking:

PabstMirror:
add another action for optionals?

PabstMirror:
or do the bom check in our own script

veteran29:
Or make PR to the action and improve it for everyone ;)

bux:
merge?

jonpas:
I would preferable bom-check everywhere and we just remove it from the generated file for now.

PabstMirror:
this will check BOM with exclude settings if we want?
```
import os
import sys
import codecs
import concurrent.futures

addon_base_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
paths_to_ignore = [x.lower() for x in ["node_modules", ".git"]]

def get_files_to_process(base_path):
    arma_files = []
    for root, _dirs, files in os.walk(base_path):
        for file in files:
            filePath = os.path.join(root, file)
            if any(p in filePath.lower() for p in paths_to_ignore): continue
            arma_files.append(os.path.join(root, file))
    print(f"-- Found {len(arma_files)} files")
    return arma_files

def process_file(filePath):
    with open(filePath, 'rb') as f:
        raw = f.read(4)
        # Note some binary files may randomly happen to start with same BOM prefix
        if (raw.startswith(codecs.BOM_UTF8)): return filePath
    return None

def main():
    print(f"-- Checking BOM")
    error_count = 0
    arma_files = get_files_to_process(addon_base_path)
    with concurrent.futures.ThreadPoolExecutor(max_workers=12) as executor:
        for fileError in executor.map(process_file, arma_files):
            if fileError is not None:
                error_count += 1
                print(f"BOM {fileError}")

    print("-- Errors: {}".format(error_count))
    return error_count


if __name__ == "__main__":
    sys.exit(main())
```

jonpas:
I will look into adding it to `bom-check` action, it's just bash after-all.

PR #9539, opened 2023-10-19
Hit Reactions - Add weapon dropping on arm hit

SpicyBagpipes:
- ~~Gun will be dropped when it is hit by a bullet if probability met.~~
- Gun will be dropped when arm is hit by a bullet if separate probability met.
- AI will retrieve gun after a couple of seconds.

### IMPORTANT


SpicyBagpipes:
https://github.com/acemod/ACE3/assets/100206101/fdb49ece-b332-4dde-a8f9-e53bf4872e01

Doesn't happen much if at all with vanilla weapons, mostly a modded thing. Vanilla weapons just have bullets pass through

LinkIsGrim:
Do we want to add weapon dropping on hand/arm hits as well? Can just check against selection

SpicyBagpipes:
> Do we want to add weapon dropping on hand/arm hits as well? Can just check against selection

Possibly, linked to different slider? Can then move condition for dropping stuff into the events?

addons/hitreactions/initSettings.sqf
```diff
@@ -15,3 +15,11 @@ private _category = [LELSTRING(common,categoryUncategorized), QUOTE(COMPONENT_BE
...
     1
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(weaponDropChanceArmHit), "SLIDER",
+    LSTRING(weaponDropChanceArmHit_displayName),
+    _category,
+    [0, 1, 0.5, 1],
+    1
+] call CBA_fnc_addSetting;
```
LinkIsGrim, line 25: newline

addons/hitreactions/initSettings.sqf
```diff
@@ -7,3 +7,19 @@ private _category = [LELSTRING(common,categoryUncategorized), QUOTE(COMPONENT_BE
...
+
+[
+    QGVAR(weaponDropChanceArmHit), "SLIDER",
+    LSTRING(weaponDropChanceArmHit_displayName),
+    _category,
+    [0, 1, 0.5, 1],
```
LinkIsGrim, line 23:
```suggestion
    [0, 1, 0.5, true],
```

Display as %

addons/hitreactions/stringtable.xml
```diff
@@ -17,5 +17,11 @@
             <Turkish>Düşmeyi tetikleyen min hasar</Turkish>
             <Spanish>Daño mínimo para provocar la caída</Spanish>
         </Key>
+        <Key ID="STR_ACE_HitReactions_weaponDropChanceGunHit_displayName">
+            <English>Chance of Weapon Drop on Gun Hit. 0 to disable.</English>
```
LinkIsGrim:
```suggestion
            <English>Weapon Drop Chance (Gun Hit)</English>
```
Add a description entry

addons/hitreactions/stringtable.xml
```diff
@@ -17,5 +17,11 @@
...
+        </Key>
+        <Key ID="STR_ACE_HitReactions_weaponDropChanceArmHit_displayName">
+            <English>Chance of Weapon Drop on Arm Hit. 0 to disable.</English>
```
LinkIsGrim:
```suggestion
            <English>Weapon Drop Chance (Arm Hit)</English>
```
Same as above

addons/hitreactions/initSettings.sqf
```diff
@@ -7,3 +7,19 @@ private _category = [LELSTRING(common,categoryUncategorized), QUOTE(COMPONENT_BE
     [-1, 1, 0.1, 1],
     1
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(weaponDropChanceGunHit), "SLIDER",
+    LSTRING(weaponDropChanceGunHit_displayName),
+    _category,
+    [0, 1, 0.5, 1],
```
LinkIsGrim, line 15:
```suggestion
    [0, 1, 0.5, true],
```
Display as %

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,37 @@
...
+ * ["", [], [1,1,1]] call ACE_hitreactions_fnc_checkWeaponDrop
+ *
+ * Public: No
+ */
+
+params ["_surfaceType", "_components", "_position"];
+
+if (!(_surfaceType isEqualTo "" && _components findAny GVAR(armComponents) isEqualTo -1)) exitWith {}; // Could be potential for edge cases here however couldn't find any from testing.
+
+private _unit = nearestObject [_pos, "CAManBase"];
```
LinkIsGrim, line 24:
```suggestion
private _unit = nearestObject [_position, "CAManBase"];
```

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,37 @@
...
+private _weapon = currentWeapon _unit;
+
+if (_weapon in GVAR(undroppableGuns)) exitWith {};
+
+private _thrownGun = _unit call ace_common_fnc_throwWeapon;
```
LinkIsGrim, line 29:
```suggestion
private _thrownGun = _unit call EFUNC(common,throwWeapon);
```

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,35 @@
...
+    "rightarmroll",
+    "righthand",
+    "rightarm"
+];
+
+["ace_firedNonPlayer", {
+    params ["", "", "", "", "", "", "_projectile"];
+    private _willDropGun = (random 1) > GVAR(weaponDropChanceGunHit);
+    private _willDropArm = (random 1) > GVAR(weaponDropChanceArmHit);
+    if (!(_willDropGun && _willDropArm)) exitWith {};
```
LinkIsGrim, line 19:
```suggestion
    private _roll = random 1;
    private _willDrop = _roll < GVAR(weaponDropChanceGunHit) || {_roll < GVAR(weaponDropChanceArmHit)};
    if (!_willDrop) exitWith {};
```
Was this working before? Am I going crazy?

I'm not a fan of the concept but adding the EH on fired is the best way to do it.

LinkIsGrim (changes requested):

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: you're not going crazy, was me being stupid.

SpicyBagpipes: lazy eval there is unnecessary, will cost more performance than the test

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
SpicyBagpipes: odd it was working beforehand, thanks

addons/hitreactions/initSettings.sqf
SpicyBagpipes: 1 is for decimals, true comes after

addons/hitreactions/stringtable.xml
SpicyBagpipes: Common sense in hindsight

addons/hitreactions/initSettings.sqf
```diff
@@ -7,3 +7,18 @@ private _category = [LELSTRING(common,categoryUncategorized), QUOTE(COMPONENT_BE
     [-1, 1, 0.1, 1],
     1
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(weaponDropEnabled), "CHECKBOX",
+    LSTRING(weaponDropEnabled_displayName),
+    _category,
+    true
```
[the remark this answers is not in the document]
SpicyBagpipes: outdated, using flat probabilities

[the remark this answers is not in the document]
SpicyBagpipes: will set arm hit to 0 by default though

addons/hitreactions/initSettings.sqf
SpicyBagpipes: Github playing up, what do you want as newline? Can only see L22-25 which is same as I already have in my env

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -14,19 +14,21 @@ GVAR(armComponents) = [
...
 }] call CBA_fnc_addEventHandler;
 
 ["ace_firedPlayer", {
     params ["", "", "", "", "", "", "_projectile"];
-    private _willDropGun = (random 1) > GVAR(weaponDropChanceGunHit);
-    private _willDropArm = (random 1) > GVAR(weaponDropChanceArmHit);
+    private _roll = random 1;
+    private _willDropGun = _roll > GVAR(weaponDropChanceGunHit);
+    private _willDropArm = _roll > GVAR(weaponDropChanceArmHit);
     if (!(_willDropGun && _willDropArm)) exitWith {};
```
LinkIsGrim, line 32: KJW's conditions

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -14,19 +14,21 @@ GVAR(armComponents) = [
 
 ["ace_firedNonPlayer", {
     params ["", "", "", "", "", "", "_projectile"];
-    private _willDropGun = (random 1) > GVAR(weaponDropChanceGunHit);
-    private _willDropArm = (random 1) > GVAR(weaponDropChanceArmHit);
-    if (!(_willDropGun && _willDropArm)) exitWith {};
-    _projectile addEventHandler ["HitPart", { 
+    private _roll = random 1;
+    private _willDropGun = _roll > GVAR(weaponDropChanceGunHit);
+    private _willDropArm = _roll > GVAR(weaponDropChanceArmHit);
```
LinkIsGrim, line 19: if lazy eval makes no sense, the extra variable allocation doesn't either

addons/hitreactions/initSettings.sqf
```diff
@@ -7,3 +7,19 @@ private _category = [LELSTRING(common,categoryUncategorized), QUOTE(COMPONENT_BE
...
+    1
+] call CBA_fnc_addSetting;
+
+[
+    QGVAR(weaponDropChanceArmHit), "SLIDER",
+    LSTRING(weaponDropChanceArmHit_displayName),
+    _category,
+    [0, 1, 0.5, 1, true],
+    1
+] call CBA_fnc_addSetting;
```
LinkIsGrim, line 25:
```suggestion
] call CBA_fnc_addSetting;

```
EOF newline

addons/hitreactions/initSettings.sqf
LinkIsGrim:
```suggestion
] call CBA_fnc_addSetting;

```
EOF newline

SpicyBagpipes:
Don't know why it's started a review for me, guh

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Example:
+ * ["", [], [1,1,1]] call ACE_hitreactions_fnc_checkWeaponDrop
+ *
+ * Public: No
+ */
+
+params ["_surfaceType", "_components", "_position"];
+
+if (!(_surfaceType isEqualTo "") && !(_components findAny GVAR(armComponents) isEqualTo -1)) exitWith {}; // Could be potential for edge cases here however couldn't find any from testing.
```
rautamiekka, line 22: Alternatively, since it exists (and a huge group of ppl dunno about it):
```suggestion
if (_surfaceType isNotEqualTo "" && (_components findAny GVAR(armComponents) isNotEqualTo -1)) exitWith {}; // Could be potential for edge cases here however couldn't find any from testing.
```

rautamiekka (changes requested):

PabstMirror:
Have you tested at all in mp?
IIRC hitPart is a little weird and there are reasons we didn't use it in medical
I think the eh will only fire on the shooter's machine, but we want throwWeapon to be where the target is local

SpicyBagpipes:
> Have you tested at all in mp? IIRC hitPart is a little weird and there are reasons we didn't use it in medical I think the eh will only fire on the shooter's machine, but we want throwWeapon to be where the target is local

Projectile hitpart has no issues like regular hitpart eh. havent tested in mp just yet but if throwweapon needs to be local can just target event it?

Drofseh:
@SpicyBagpipes thoughts on rifle sling and pistol lanyard items
If the unit has a sling in inventory they always retain or get a bonus to retain the primary weapon, lanyard does the same for secondary (handgun) weapons. 

Items wouldn't need a model, just use the generic ammo pouch, so all that'd really be needed is a UI image. 


Could require the sling to put a weapon on back too, but that'd be for a different PR.

SpicyBagpipes:
> @SpicyBagpipes thoughts on rifle sling and pistol lanyard items?
> 
> If the unit has a sling in inventory they always retain or get a bonus to retain the primary weapon, lanyard does the same for sidearm weapons.
> 
> Items wouldn't need a model, just use the generic ammo pouch, so all that'd really be needed is a UI image.
> 
> Could require the sling to put a weapon on back too, but that'd be for a different PR.

Interesting idea, though feels too complex for ace and more in place with unit specific mods imo

SpicyBagpipes:
https://github.com/acemod/ACE3/assets/100206101/0ecd045d-1863-45ba-91e8-d1532a5cfc6d

LinkIsGrim:
AI will be *really* vulnerable in CQB if they drop their weapon and don't switch to their pistol.

SpicyBagpipes:
> AI will be _really_ vulnerable in CQB if they drop their weapon and don't switch to their pistol.

They switch to pistol if needed

LinkIsGrim:
> > AI will be _really_ vulnerable in CQB if they drop their weapon and don't switch to their pistol.
> 
> They switch to pistol if needed. Won't swap back to rifle unless needed too.
> 
> Can always cba waituntilandexecute findnearestenemy being 5 meters away or so

They'll play the pickup animation anyway, won't they?

I like that solution, but it should keep the small delay it currently has as well.

SpicyBagpipes:
> > > AI will be _really_ vulnerable in CQB if they drop their weapon and don't switch to their pistol.
> > 
> > They switch to pistol if needed. Won't swap back to rifle unless needed too.
> > 
> > Can always cba waituntilandexecute findnearestenemy being 5 meters away or so
> 
> They'll play the pickup animation anyway, won't they?
> 
> I like that solution, but it should keep the small delay it currently has as well.

Ah, true.

I'll add it later today -- just go straight through the check if the unit doesn't have either a pistol or a primary?

veteran29:
I think the feature should have a toggle setting and it should be off by default. It's really strong against AI and also adds potential issue with creation of scripted ground holders that could not be picked up by various "garbage collectors".

The Event handlers should early exit when the feature if disabled to reduce perf cost.

SpicyBagpipes:
> I think the feature should have a toggle setting and it should be off by default. It's really strong against AI and also adds potential issues with the creation of scripted ground holders that could not be picked up by various "garbage collectors".
> 
> The Event handlers should early exit when the feature is disabled to reduce performance cost.

Is setting probability to 0 not sufficient? Can early exit when equals 0 if you want. If not then will do and move settings to their own category.

From my (albeit limited) testing I couldn't find any issues with GWHs as they have a bisurf set by default from what I can tell but probable I've missed some cases.

veteran29:
> From my (albeit limited) testing I couldn't find any issues with GWHs as they have a bisurf set by default from what I can tell but probable I've missed some cases.

I mean about creating the bodies. When you delete the body its weapon will be gone too even if it fell out of their hand. If you shoot the AI in the gun, the GWH will stay after deletion of the unit.

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,68 @@
...
+    "lefthand",
+    "leftforearm",
+    "leftarmroll",
+    "rightforearm",
+    "rightarmroll",
+    "righthand",
+    "rightarm"
+];
+
+GVAR(safePickupDistance) = 8;
```
LinkIsGrim, line 15: Valhe should be #define'd in script_component.hpl

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,68 @@
...
+
+["ace_firedNonPlayer", {
+    params ["", "", "", "", "", "", "_projectile"];
+    if (GVAR(weaponDropChanceGunHit) + GVAR(weaponDropChanceArmHit) == 0) exitWith {};
+    private _roll = random 1;
+    private _willDropGun = _roll < GVAR(weaponDropChanceGunHit);
+    private _willDropArm = _roll < GVAR(weaponDropChanceArmHit);
```
LinkIsGrim, line 22: I wouldn't do variable allocation and just throw these into the condition block directly

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+private _didHitArm = _components findAny GVAR(armComponents) isNotEqualTo -1;
+private _didHitGun = _surfaceType isEqualTo "" && _components isEqualTo [];  // Could be potential for edge cases here however couldn't find any from testing.
+
+if (!_didHitArm && !_didHitGun) exitWith {}; // Exit if it didn't hit an arm and also didn't hit a gun.
+
+private _unit = nearestObject [_position, "CAManBase"];
+private _weapon = currentWeapon _unit;
+
+if (_weapon in GVAR(undroppableGuns)) exitWith {};
```
LinkIsGrim, line 30: Where is this GVAR's default value set?

LinkIsGrim (changes requested):

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: More readable imo

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
LinkIsGrim: I don't think so?

Standby. There's a safer way to define that var.

addons/hitreactions/XEH_postInit.sqf
LinkIsGrim: Performance is a concern if this will be used on fired event.

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
LinkIsGrim: Ref https://github.com/acemod/ACE3/blob/eee86a4baccf07626d035892e55132ea135fc05e/addons/medical_engine/XEH_preInit.sqf#L12

That way we don't overwrite user's modifications but still define the variable. Can do this in postInit.

```
if (isNil QGVAR(undroppableGuns)) then {GVAR(undroppableGuns = []};
```

SpicyBagpipes: done

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: Done

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+[QGVAR(dropGun), [_unit], _unit] call CBA_fnc_targetEvent;
```
LinkIsGrim, line 32:
```suggestion
[QGVAR(dropGun), [_unit], _unit] call CBA_fnc_targetEvent;

```
EOF newline

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Public: No
+ */
+
+params ["_surfaceType", "_components", "_position"];
+
+private _didHitArm = _components findAny GVAR(armComponents) isNotEqualTo -1;
+private _didHitGun = _surfaceType isEqualTo "" && _components isEqualTo [];  // Could be potential for edge cases here however couldn't find any from testing.
+
+if (!_didHitArm && !_didHitGun) exitWith {}; // Exit if it didn't hit an arm and also didn't hit a gun.
```
LinkIsGrim, line 25:
```suggestion
if !(_didHitArm || _didHitGun) exitWith {};
```
I prefer this condition style but others might disagree. Comment can go, condition is self-explanatory

LinkIsGrim (approved):
LGTM

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,65 @@
...
+    "leftforearm",
+    "leftarmroll",
+    "rightforearm",
+    "rightarmroll",
+    "righthand",
+    "rightarm"
+];
+
+GVAR(safePickupDistance) = DEFAULT_PICKUP_DISTANCE;
+if (isNil QGVAR(undroppableGuns)) then {GVAR(undroppableGuns) = DEFAULT_UNDROPPABLE_GUNS};
```
jonpas, line 16: How would it not be `nil`?

This and everything above fits `preInit` more.

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,65 @@
...
+    "lefthand",
+    "leftforearm",
+    "leftarmroll",
+    "rightforearm",
+    "rightarmroll",
+    "righthand",
+    "rightarm"
+];
+
+GVAR(safePickupDistance) = DEFAULT_PICKUP_DISTANCE;
```
jonpas, line 15: Why not just use the define? Do you want that changeable on the fly? Should use `missionNamespace` specifically then.

addons/hitreactions/XEH_postInit.sqf
LinkIsGrim: user-defined in before postInit

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,65 @@
...
+if (isNil QGVAR(undroppableGuns)) then {GVAR(undroppableGuns) = DEFAULT_UNDROPPABLE_GUNS};
+
+["ace_firedNonPlayer", {
```
PabstMirror, line 18: we should conditionally add these event handlers based on settings

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: @LinkIsGrim just move that and everything above to preinit and skip isnil check?

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: yes -- is missionnamespace not implied in func?

addons/hitreactions/XEH_postInit.sqf
LinkIsGrim: Moving to preInit is fine, isNil check should be kept. What if user's addon loaded first and defined that in preInit as well?

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
SpicyBagpipes: Edited for it, imo previous one is more readable but second is likely slightly more performant

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: Done

addons/hitreactions/XEH_preInit.sqf
```diff
@@ -9,3 +9,18 @@ PREP_RECOMPILE_END;
...
+    "leftforearm",
+    "leftarmroll",
+    "rightforearm",
+    "rightarmroll",
+    "righthand",
+    "rightarm"
+];
+
+GVAR(safePickupDistance) = DEFAULT_PICKUP_DISTANCE;
+GVAR(undroppableGuns) = DEFAULT_UNDROPPABLE_GUNS;
```
LinkIsGrim, line 26:
```suggestion
if (isNil QGVAR(safePickupDistance)) then {GVAR(safePickupDistance) = DEFAULT_PICKUP_DISTANCE};
if (isNil QGVAR(undroppableGuns)) then {GVAR(undroppableGuns) = DEFAULT_UNDROPPABLE_GUNS};
```

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,50 @@
...
+
+[QGVAR(dropGun), {
+    params ["_unit"];
+    private _weapon = currentWeapon _unit;
+    private _thrownGun = _unit call EFUNC(common,throwWeapon);
+    if (isPlayer _unit) exitWith {}; // Don't make the player pick its own gun up.
+    [
+        {
+            params ["_thrownGun", "_weapon", "_unit"];
+            handgunWeapon _unit == "" && primaryWeapon _unit == "" || {(_unit distance (_unit findNearestEnemy _unit)) > GVAR(safePickupDistance)}
```
rautamiekka, line 35:
```suggestion
            handgunWeapon _unit isEqualTo "" && primaryWeapon _unit isEqualTo "" || {(_unit distance (_unit findNearestEnemy _unit)) > GVAR(safePickupDistance)}
```
For consistency.

rautamiekka (changes requested):

Mike-MF (commented):
I'm seeing a serious lack of documentation for this.

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: consistency with rest of addon, uses ==

SpicyBagpipes:
> I'm seeing a serious lack of documentation for this.

Doing once I've got finalised state for pr

addons/hitreactions/XEH_postInit.sqf
rautamiekka: Fair enough.

SpicyBagpipes:
Which is now... Would this be better suited under frameworks or features?

LinkIsGrim:
Features -> Hit Reactions

SpicyBagpipes:
Anything else required for docs? Thinking safe pickup distance should be omitted as doesn't really need to be modified

LinkIsGrim:
Should indicate classnames are case-sensitive and don't use inheritance

Up to you on pickup distance, I'd document it though.

SpicyBagpipes:
Will omit pickup distance as any change to value will rarely be seen

LinkIsGrim (approved):

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,32 @@
...
+ * Arguments:
+ * 0: surfaceType <STRING>
+ * 1: components <ARRAY>
+ * 2: position <ARRAY>
+ *
+ * Return Value:
+ * None
+ *
+ * Example:
+ * ["", [], [1,1,1]] call ACE_hitreactions_fnc_checkWeaponDrop
```
johnb432, line 15:
```suggestion
 * Checks if the unit should drop their weapon based on projectile hit info.
 *
 * Arguments:
 * 0: Surface type <STRING>
 * 1: Components <ARRAY>
 * 2: Position <ARRAY>
 *
 * Return Value:
 * None
 *
 * Example:
 * ["", [], [1, 1, 1]] call ace_hitreactions_fnc_checkWeaponDrop
```

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Public: No
+ */
+
+params ["_surfaceType", "_components", "_position"];
+
+private _didHitArm = _components findAny GVAR(armComponents) isNotEqualTo -1;
+private _didHitGun = _surfaceType isEqualTo "" && _components isEqualTo [];  // Could be potential for edge cases here however couldn't find any from testing.
```
johnb432, line 23:
```suggestion
private _didHitGun = _surfaceType isEqualTo "" && _components isEqualTo [];  // for potential edge cases
```

addons/hitreactions/initSettings.sqf
```diff
@@ -7,3 +7,19 @@ private _category = [LELSTRING(common,categoryUncategorized), QUOTE(COMPONENT_BE
...
+    _category,
+    [0, 1, 0, 1, true],
+    1
+] call CBA_fnc_addSetting;
+
+[
+    QGVAR(weaponDropChanceArmHit), "SLIDER",
+    LSTRING(weaponDropChanceArmHit_displayName),
+    _category,
+    [0, 1, 0, 1, true],
```
LinkIsGrim, line 23:
```suggestion
    [0, 1, 0, 2, true],
    1
] call CBA_fnc_addSetting;

[
    QGVAR(weaponDropChanceArmHit), "SLIDER",
    LSTRING(weaponDropChanceArmHit_displayName),
    _category,
    [0, 1, 0, 2, true],
```

LinkIsGrim (approved):

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,50 @@
+#include "script_component.hpp"
+
+if (GVAR(weaponDropChanceGunHit) + GVAR(weaponDropChanceArmHit) == 0) exitWith {};
```
johnb432, line 3: Either you need to add the `_needRestart` parameter (see https://github.com/CBATeam/CBA_A3/wiki/CBA-Settings-System#arguments-of-cba_fnc_addsetting) to the settings or this check has to go, as it could be changed mid-mission by the server.

Personally, I prefer if you don't have to restart the mission just to change a CBA setting, so I'd like to see the check removed.

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,50 @@
...
+}] call CBA_fnc_addEventHandler;
+
+if (!hasInterface) exitWith {};
+
+["ace_firedPlayer", {
+    params ["", "", "", "", "", "", "_projectile"];
+    private _roll = random 1;
+    if (!(_roll < GVAR(weaponDropChanceGunHit) || _roll < GVAR(weaponDropChanceArmHit))) exitWith {};
+    _projectile addEventHandler ["HitPart", { 
+        params ["_projectile", "_hitEntity", "_projectileOwner", "_pos", "_velocity", "_normal", "_components", "_radius" ,"_surfaceType"];
```
johnb432, line 22:
```suggestion
    private _roll = random 1;
    if !(_roll < GVAR(weaponDropChanceGunHit) || _roll < GVAR(weaponDropChanceArmHit)) exitWith {};
    
    params ["", "", "", "", "", "", "_projectile"];
    
    _projectile addEventHandler ["HitPart", {
        params ["", "", "", "_pos", "", "", "_components", "" ,"_surfaceType"];
        [_surfaceType, _components, _pos] call FUNC(checkWeaponDrop);
    }];
}] call CBA_fnc_addEventHandler;

if (!hasInterface) exitWith {};

["ace_firedPlayer", {
    private _roll = random 1;
    if !(_roll < GVAR(weaponDropChanceGunHit) || _roll < GVAR(weaponDropChanceArmHit)) exitWith {};
    
    params ["", "", "", "", "", "", "_projectile"];
    
    _projectile addEventHandler ["HitPart", { 
       params ["", "", "", "_pos", "", "", "_components", "" ,"_surfaceType"];
```

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,50 @@
...
+            handgunWeapon _unit == "" && primaryWeapon _unit == "" || {(_unit distance (_unit findNearestEnemy _unit)) > GVAR(safePickupDistance)}
+        },
+        {
+            params ["_thrownGun", "_weapon", "_unit"];
+            [
+                {
+                    params ["_thrownGun", "_weapon", "_unit"];
+                    _unit action ["TakeWeapon", _thrownGun, _weapon];
+                },
+                [_thrownGun, _weapon, _unit],
```
johnb432, line 44:
```suggestion
            [
                {
                    params ["_thrownGun", "_weapon", "_unit"];
                    _unit action ["TakeWeapon", _thrownGun, _weapon];
                },
                _this,
```

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Example:
+ * ["", [], [1, 1, 1]] call ace_hitreactions_fnc_checkWeaponDrop
+ *
+ * Public: No
+ */
+
+params ["_surfaceType", "_components", "_position"];
+
+private _didHitArm = _components findAny GVAR(armComponents) isNotEqualTo -1;
```
johnb432, line 22: For readability's sake:
```suggestion
private _didHitArm = _components findAny GVAR(armComponents) != -1;
```

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
```diff
@@ -0,0 +1,32 @@
...
+private _didHitGun = _surfaceType isEqualTo "" && _components isEqualTo [];  // for potential edge cases
+
+if !(_didHitArm || _didHitGun) exitWith {};
+
+private _unit = nearestObject [_position, "CAManBase"];
+private _weapon = currentWeapon _unit;
+
+if (_weapon in GVAR(undroppableGuns)) exitWith {};
+
+[QGVAR(dropGun), [_unit], _unit] call CBA_fnc_targetEvent;
```
johnb432, line 32:
```suggestion
[QGVAR(dropGun), _unit, _unit] call CBA_fnc_targetEvent;
```

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,50 @@
+#include "script_component.hpp"
+
+if (GVAR(weaponDropChanceGunHit) + GVAR(weaponDropChanceArmHit) == 0) exitWith {};
+
+["ace_firedNonPlayer", {
+    params ["", "", "", "", "", "", "_projectile"];
+    private _roll = random 1;
+    if (!(_roll < GVAR(weaponDropChanceGunHit) || _roll < GVAR(weaponDropChanceArmHit))) exitWith {};
```
johnb432, line 8: There is no point in having 2 separate settings when you combine them.

Either you need to check the probabilities in `FUNC(checkWeaponDrop)` or you have one setting.

I tested this by setting `Weapon Drop Chance (Gun Hit)` to 100 % and the other to 0%. When a unit is shot in the arm, he drops the weapon.

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,50 @@
...
+    _projectile addEventHandler ["HitPart", { 
+        params ["_projectile", "_hitEntity", "_projectileOwner", "_pos", "_velocity", "_normal", "_components", "_radius" ,"_surfaceType"];
+        [_surfaceType, _components, _pos] call FUNC(checkWeaponDrop);
+    }];
+}] call CBA_fnc_addEventHandler;
+
+[QGVAR(dropGun), {
```
johnb432, line 27: This can happen locally to a machine without an interface. You need to add tihs EH to all machines.

johnb432 (changes requested):

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: Holdover from early exit, removed the check

addons/hitreactions/functions/fnc_checkWeaponDrop.sqf
SpicyBagpipes: Committed however `!=` is slower than `isNotEqualTo` in some cases. Don't think that is the case in this one.

addons/hitreactions/XEH_postInit.sqf
SpicyBagpipes: Resolved in latest commit, think there may be a faster way to do it though

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,47 @@
...
+                {
+                    params ["_thrownGun", "_weapon", "_unit"];
+                    _unit action ["TakeWeapon", _thrownGun, _weapon];
+                },
+                _this,
+                random [1,2,3]
+            ] call CBA_fnc_waitAndExecute;
+        },
+        [_thrownGun, _weapon, _unit]
+    ] call CBA_fnc_waitUntilAndExecute;
```
johnb432, line 46: I'm not a fan of the check running every frame, without a timeout. Imo, this should be replaced by a PFH and a timeout should be defined.

SpicyBagpipes: If unit has been shot to have its gun dropped is it not reasonable to presume they are going to die imminently anyways?

johnb432: Imo no, this is too unpredictable to be assumed.

rautamiekka: > If unit has been shot to have its gun dropped is it not reasonable to presume they are going to die imminently anyways?

Not reasonable.

johnb432:
`QGVAR(dropGun)` is being triggered twice in some instances.

Repro steps:
1. Place down a unit that has a primary and a handgun (e.g. `"B_Soldier_A_F"`), and set both probabilities to 100%.
2. Shoot unit in arm. The unit drops his primary and picks it up again, but switches to handgun.
3. Shoot unit in arm again. See that both his handgun and primary weapon are dropped.

My uneducated guess is that the `"HitPart"` EH is firing twice on the projectile, and as the unit quickly changes from having no weapon because the pistol has been dropped, it also drops the primary.

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -0,0 +1,47 @@
...
+}] call CBA_fnc_addEventHandler;
```
johnb432, line 47: This is what I was thinking would be suited for checking if a unit should pick up their weapon.

```suggestion
[QGVAR(doMove), {
    params ["_unit", "_pos"];

    _unit setDestination [_pos, "LEADER PLANNED", true];
    _unit doMove _pos;
}] call CBA_fnc_addEventHandler;

[QGVAR(dropGun), {
    params ["_unit"];

    private _weapon = currentWeapon _unit;
    private _thrownGun = _unit call EFUNC(common,throwWeapon);

    if (isPlayer _unit) exitWith {}; // Don't make the player pick its own gun up.

    [{
        (_this select 0) params ["_unit", "_weapon", "_thrownGun", "_timeout"];

        // If the unit has been deleted or dead, if the gun doesn't exist anymore or if it's been too long, stop
        if (!alive _unit || {isNull _thrownGun} || {CBA_missionTime >= _timeout}) exitWith {
            (_this select 1) call CBA_fnc_removePerFrameHandler;
        };

        // If the unit has no essential weapons, force them to get their weapon, otherwise wait until no enemies are present
        if ((primaryWeapon _unit == "" && {handgunWeapon _unit == ""}) || {(_unit distance (_unit findNearestEnemy _unit)) > GVAR(safePickupDistance)}) then {
            // If the unit is too far away, make them move closer
            if (_unit distance _thrownGun >= 4) exitWith {
                [QGVAR(doMove), [_unit, getPosATL _thrownGun]] call CBA_fnc_targetEvent;
            };

            (_this select 1) call CBA_fnc_removePerFrameHandler;

            [{
                params ["_unit", "_weapon", "_thrownGun"];

                _unit action ["TakeWeapon", _thrownGun, _weapon];
            }, [_unit, _weapon, _thrownGun], random [1, 2, 3]] call CBA_fnc_waitAndExecute;
        };
    }, 5, [_unit, _weapon, _thrownGun, CBA_missionTime + 300]] call CBA_fnc_addPerFrameHandler;
}] call CBA_fnc_addEventHandler;
```

johnb432:
Bug: When the shooter is close to target when firing, it can make the shooter drop his weapon.

The `_hitEntity` from https://community.bistudio.com/wiki/Arma_3:_Event_Handlers#HitPart_2 should to be taken into account, at least for when the arms are hit.

---
@SpicyBagpipes Are you going to keep working on this?

SpicyBagpipes:
> Bug: When the shooter is close to target when firing, it can make the shooter drop his weapon.
> 
> The `_hitEntity` from https://community.bistudio.com/wiki/Arma_3:_Event_Handlers#HitPart_2 should to be taken into account, at least for when the arms are hit.
> 
> @SpicyBagpipes Are you going to keep working on this?

Yes have just been swamped in uni work, need to find the time to get ahead of it and do some proper work. HitEntity for gun hits doesn't give a value but can do for arm hits

SpicyBagpipes:
> Bug: When the shooter is close to target when firing, it can make the shooter drop his weapon.
> 
> The `_hitEntity` from https://community.bistudio.com/wiki/Arma_3:_Event_Handlers#HitPart_2 should to be taken into account, at least for when the arms are hit.

Latest commit does the latter, need to find a good way to do the former

SpicyBagpipes:
2.16 appears to have broken gun hits as surfaceType is now returning bisurfs

SpicyBagpipes:
Shooter dropping weapon on arm hit should no longer happen, and unless we get a weapon component I think it's unlikely that gun hits will work again; I can note out the gun hits part of the code or delete it?

johnb432:
> Shooter dropping weapon on arm hit should no longer happen, and unless we get a weapon component I think it's unlikely that gun hits will work again; I can note out the gun hits part of the code or delete it?

I'm fine with it - @LinkIsGrim what do you think?

Drofseh:
@SpicyBagpipes could to do separate % settings for players and AI?

`docs/wiki/feature/hitreactions.md ` will have to be updated if gun hits don't function anymore.

johnb432:
> `QGVAR(dropGun)` is being triggered twice in some instances.
> 
> Repro steps:
> 
>     1. Place down a unit that has a primary and a handgun (e.g. `"B_Soldier_A_F"`), and set both probabilities to 100%.
> 
>     2. Shoot unit in arm. The unit drops his primary and picks it up again, but switches to handgun.
> 
>     3. Shoot unit in arm again. See that both his handgun and primary weapon are dropped.
> 
> 
> I confirmed it isn't `EFUNC(common,throwWeapon)` causing the issue, it's definitely the event that is being raised twice.
> 
> My uneducated guess is that the `"HitPart"` EH is firing twice on the projectile, and as the unit quickly changes from having no weapon because the pistol has been dropped, it also drops the primary.

Fixed in latest commit.

---

- Removed gun hit part.
- Updated wiki accordingly.
- Removed getting closest entity when a projectile hits something. Why was it added in the first place @SpicyBagpipes?
- Unconscious units won't attempt to drop their weapons. I don't think it worked, but I've added some safeguards regarding that.
  However, units transitioning to unconscious animations can still pick up weapons. I would like to address that, but I'm not sure how/if I have the motivation to find a solution to that.

SpicyBagpipes:
> > `QGVAR(dropGun)` is being triggered twice in some instances.
> > 
> > Repro steps:
> > 
> >     1. Place down a unit that has a primary and a handgun (e.g. `"B_Soldier_A_F"`), and set both probabilities to 100%.
> > 
> >     2. Shoot unit in arm. The unit drops his primary and picks it up again, but switches to handgun.
> > 
> >     3. Shoot unit in arm again. See that both his handgun and primary weapon are dropped.
> > 
> > 
> > I confirmed it isn't `EFUNC(common,throwWeapon)` causing the issue, it's definitely the event that is being raised twice.
> > 
> > My uneducated guess is that the `"HitPart"` EH is firing twice on the projectile, and as the unit quickly changes from having no weapon because the pistol has been dropped, it also drops the primary.
> 
> Fixed in latest commit.
> 
> ---
> 
> - Removed gun hit part.
> - Updated wiki accordingly.
> - Removed getting closest entity when a projectile hits something. Why was it added in the first place @SpicyBagpipes?
> - Unconscious units won't attempt to drop their weapons. I don't think it worked, but I've added some safeguards regarding that.
>   However, units transitioning to unconscious animations can still pick up weapons. I would like to address that, but I'm not sure how/if I have the motivation to find a solution to that.

Entity was only way to get any reference to the weapon "owner" (whoever is holding what gun got hit) as there was no other reliable method. As of 2.16 _something_ seems to have changed behaviour in hitpart return so it no longer functions so no need any more.


Is lifestate not incapacitated whilst in transition to uncon?

SpicyBagpipes:
> @SpicyBagpipes could to do separate % settings for players and AI?
> 
> `docs/wiki/feature/hitreactions.md` will have to be updated if gun hits don't function anymore.

Will do seperate % if John hasn't already, haven't checked as still early here

johnb432:
> Will do seperate % if John hasn't already, haven't checked as still early here

I didn't do it, as I forgot that. I'll let you do it.

> Is lifestate not incapacitated whilst in transition to uncon?

Possibly. Can you investigate that?

SpicyBagpipes:
> > Will do seperate % if John hasn't already, haven't checked as still early here
> 
> I didn't do it, as I forgot that. I'll let you do it.
> 
> > Is lifestate not incapacitated whilst in transition to uncon?
> 
> Possibly. Can you investigate that?

Tested with `allUnits select {lifeState _x != "HEALTHY"}`, unit is picked up the second they start falling once uncon

johnb432 (approved):

johnb432:
In a future PR (potentially): Look at adding this to all projectiles (I'm thinking mostly about fragmentation).

addons/hitreactions/XEH_postInit.sqf
```diff
@@ -70,6 +70,14 @@
             (_this select 1) call CBA_fnc_removePerFrameHandler;
 
             _unit action ["TakeWeapon", _thrownWeapon, _weapon];
+
+            // "TakeWeapon" action bugs out unit where they don't switch weapons
+            // Resetting loadout fixes bugged state
+            [{
+                (_this select 0) hasWeapon (_this select 1)
+            }, {
+            	(_this select 0) setUnitLoadout getUnitLoadout (_this select 0);
```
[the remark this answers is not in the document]
PabstMirror: `setUnitLoadout` will break all things using unique magazines (intel items)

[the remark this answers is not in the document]
johnb432: > Do we have any other way of doing this?

I tried with `CBA_fnc_addWeaponWithoutItems` and adding items individually with `addWeaponItem`, but that didn't work, the AI would reload when receiving the new weapon.

I can try to use `addWeapon` and replacing the magazine that's inserted into the gun, but maybe even then the AI will reload, but I haven't checked yet.

> `setUnitLoadout` will break all things using unique magazines (intel items)

I wasn't aware of this because I added it in the arsenal. It's executed when you change a container (uniform, vest or backpack).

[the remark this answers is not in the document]
johnb432: I'm working on a method that doesn't use `setUnitLoadout`, but it's currently 100 extra lines of code.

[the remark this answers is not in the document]
johnb432: > `setUnitLoadout` will break all things using unique magazines (intel items)

Do we even care about that for AI?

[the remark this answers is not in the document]
johnb432: Fair - I have come up with a far better solution now.

johnb432 (changes requested):
I'm going to make another PR that needs to be merged before this one, for the changes in `ace_common_fnc_addWeapon`.

addons/hitreactions/XEH_postInit.sqf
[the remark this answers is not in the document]
johnb432: I retested it without the fix, just to make sure it was necessary.
I retested it several times, with several game restarts and it's now not necessary. I'm certain that yesterday's testing showed it was necessary, so I'm at a loss for words.

johnb432 (approved):

PR #9542, opened 2023-10-20
Arsenal - Add support for modifying Virtual Items while opened

LinkIsGrim:
- Title.
- Move updating `GVAR(virtualItemsFlat)` to function
- Update English "No virtual item" stringtable to plural, rest are fine as they refer to "None" items being available.

### DIFF
--- a/addons/arsenal/functions/fnc_addVirtualItems.sqf
+++ b/addons/arsenal/functions/fnc_addVirtualItems.sqf
@@ -123,3 +123,8 @@ if (_items isEqualType true) then {
 };
 
 _object setVariable [QGVAR(virtualItems), _cargo, _global];
+
+private _currentBox = !isNil QGVAR(currentBox) && {GVAR(currentBox) isEqualTo _object};
+if (_currentBox) then {
+    [true, true] call FUNC(refresh);
+};
--- a/addons/arsenal/functions/fnc_addVirtualItems.sqf
+++ b/addons/arsenal/functions/fnc_addVirtualItems.sqf
@@ -124,7 +124,7 @@ if (_items isEqualType true) then {
 
 _object setVariable [QGVAR(virtualItems), _cargo, _global];
 
-private _currentBox = !isNil QGVAR(currentBox) && {GVAR(currentBox) isEqualTo _object};
-if (_currentBox) then {
+// If the arsenal is already open, refresh arsenal display
+if (!isNil QGVAR(currentBox) && {GVAR(currentBox) isEqualTo _object}) then {
     [true, true] call FUNC(refresh);
 };
--- a/addons/arsenal/functions/fnc_refresh.sqf
+++ b/addons/arsenal/functions/fnc_refresh.sqf
@@ -6,6 +6,7 @@
  *
  * Arguments:
  * 0: Update current and unique items lists <BOOL> (default: true)
+ * 1: Update virtual items list <BOOL> (default: false)
  *
  * Return Value:
  * None
@@ -15,7 +16,7 @@
  *
  * Public: Yes
 */
-params [["_updateItems", true, [true]]];
+params [["_updateItems", true, [true]], ["_updateVirtualItems", false, [false]]];
 
 if (_updateItems) then {
     // Update current item list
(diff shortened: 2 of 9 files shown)


LinkIsGrim:
Should notify the user the arsenal was closed because it was updated.

addons/arsenal/functions/fnc_addVirtualItems.sqf
```diff
@@ -123,3 +123,8 @@ if (_items isEqualType true) then {
 };
 
 _object setVariable [QGVAR(virtualItems), _cargo, _global];
+
+private _currentBox = !isNil QGVAR(currentBox) && {GVAR(currentBox) isEqualTo _object};
+if (_currentBox) then {
```
johnb432, line 128:
```suggestion
// If the arsenal is already open, refresh arsenal display
if (!isNil QGVAR(currentBox) && {GVAR(currentBox) isEqualTo _object}) then {
```

addons/arsenal/functions/fnc_refresh.sqf
```diff
@@ -15,7 +16,9 @@
  *
  * Public: Yes
 */
-params [["_updateItems", true, [true]]];
+params [["_updateItems", true, [true]], ["_updateVirtualItems", false, [false]]];
+
+TRACE_2("",_updateItems,_updateVirtualItems);
 
 if (_updateItems) then {
```
johnb432, line 23: As we are going to be updating the display if `_updateVirtualItems` is true, no need to do something expensive twice.
```suggestion
if (_updateItems && !_updateVirtualItems) then {
```

addons/arsenal/functions/fnc_refresh.sqf
```diff
@@ -25,6 +28,21 @@ if (_updateItems) then {
     call FUNC(updateUniqueItemsList);
```
johnb432, line 28: As we are going to be updating the display if _updateVirtualItems is true, no need to do something expensive twice.
```suggestion
    if (!_updateVirtualItems) then {
        call FUNC(updateUniqueItemsList);
    };
```

addons/arsenal/functions/fnc_refresh.sqf
```diff
@@ -25,6 +28,21 @@ if (_updateItems) then {
...
 };
 
+private _virtualItems = GVAR(currentBox) getVariable QGVAR(virtualItems);
+if (isNil "_virtualItems") exitWith {
+    [LLSTRING(noVirtualItems), false, 5, 1] call EFUNC(common,displayText);
```
johnb432, line 33: `STR_ACE_Arsenal_noVirtualItems` needs to be updated to `No virtual items available`

johnb432:
`ace_arsenal_fnc_removeBox` should also be updated.

In fact, all of your 
```
if (_currentBox) then {
    call _fnc_closeArsenal;
};
```
calls are just after a `ace_arsenal_fnc_removeBox` call. Imo move the above mentioned code from `ace_arsenal_fnc_removeVirtualItems` into `ace_arsenal_fnc_removeBox`.

LinkIsGrim:
Bugs with adding virtual items atm

LinkIsGrim:
Fixed script errors, works now.

PabstMirror (approved):

addons/arsenal/functions/fnc_refresh.sqf
```diff
@@ -15,13 +16,37 @@
...
     call FUNC(updateCurrentItemsList);
 
     // This takes care of unique inventory items (arsenal doesn't have it whitelisted)
+    if (!_updateVirtualItems) then {
+        call FUNC(updateUniqueItemsList);
+    };
+};
+
+private _virtualItems = GVAR(currentBox) getVariable QGVAR(virtualItems);
+if (isNil "_virtualItems") exitWith {
```
PabstMirror, line 39: this ends up being `nil` in 3den when editing a unit's inventory

the displayText and execNextframe both don't work in 3den so I guess it's ok
but it feels a little dangerous

johnb432: I think it might be possible to not run this function at all when in 3den, so we could add a condition at the very top to exit.

PR #9543, opened 2023-10-20
Medical - Change default Fatal Injuries Source setting to Either

LinkIsGrim:
- Title.


LinkIsGrim:
I know changing default settings is bad but this will save us some headaches from "hey I shot 200 rounds into this guy and he won't die"

johnb432 (approved):

PR #9546, opened 2023-10-20
Arsenal - Add param to skip refresh animation

LinkIsGrim:
- Title.

IMO `FUNC(refresh)` should skip by default. Best done after #9542 to avoid conflicts

### IMPORTANT

### DIFF
--- a/addons/arsenal/functions/fnc_refresh.sqf
+++ b/addons/arsenal/functions/fnc_refresh.sqf
@@ -30,4 +30,4 @@ if (!isNull findDisplay IDD_loadouts_display) exitWith {};
 
 private _display = findDisplay IDD_ace_arsenal;
 
-[_display, _display displayCtrl GVAR(currentLeftPanel)] call FUNC(fillLeftPanel);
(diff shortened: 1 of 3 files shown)


addons/arsenal/functions/fnc_refresh.sqf
```diff
@@ -53,6 +54,11 @@ if (_updateVirtualItems) then {
 // Don't refresh left panel if in loadout tab
 if (!isNull findDisplay IDD_loadouts_display) exitWith {};
 
+if (!_animate) then {
+    GVAR(refreshing) = true;
+    [{GVAR(refreshing) = false}, nil, 3] call CBA_fnc_execAfterNFrames;
```
PabstMirror, line 59: execAfterNFrames won't work in 3den, but I guess it won't ever be reached so it's ok?
ref https://github.com/acemod/ACE3/pull/9542/files#r1369643874

PabstMirror: Maybe on line 38 we add `if (is3den) exitWith {};`?

LinkIsGrim:
Updated action docs and added special case handling for 3DEN.

PabstMirror (approved):

PR #9560, opened 2023-10-22
Common - Diagnose ace install if there are version problems

PabstMirror:


LinkIsGrim:
Ref above: https://steamcommunity.com/sharedfiles/filedetails/?id=1744095754

LinkIsGrim:
image

Logging works. I'd edit the "ace_component is outdated" message box to be something like this instead as well:
"ace_component is outdated (VERSION.NUMBER.NUMBER), common version is (COMMON.VERSION.NUMBER)"

LinkIsGrim:
image
Thoughts?

PabstMirror:
BTW this isn't meant to block or discourage custom builds or repacking/reuploading ace
that's still totally fine to do based on our license

it's just going to print more debug info on mod folders / workshop source and only when there is a problem

LinkIsGrim (approved):

jonpas (approved):

addons/common/functions/fnc_checkFiles.sqf
```diff
@@ -39,29 +39,47 @@ if ([_cbaRequiredAr, _cbaVersionAr] call cba_versioning_fnc_version_compare) the
...
+        call FUNC(checkFiles_diagnoseACE);
 
         if ((_x select [0, 10]) != "ace_compat") then {
             if (hasInterface) then {
-                ["[ACE] ERROR", _errorMsg, {findDisplay 46 closeDisplay 0}] call FUNC(errorMessage);
+                _oldAddons pushBack _x;
             };
         } else {
-            _oldCompats pushBack _x;  // Don't block game if it's just an old compat pbo
+            _oldCompats pushBack [_x, _addonVersion];  // Don't block game if it's just an old compat pbo
```
PabstMirror, line 58:
```suggestion
            _oldCompats pushBack [_x, _addonVersion, _addonSource];  // Don't block game if it's just an old compat pbo
```
I think? not sure but looks like line 78 uses it?

LinkIsGrim: Forgot to remove after testing. Will just edit line 78.

PR #9582, opened 2023-10-25
WIP - Arsenal - Add filters to arsenal 

amsteadrayle:
- Add functionality to filter the lists in the arsenal

### IMPORTANT


amsteadrayle:
https://github.com/acemod/ACE3/assets/2516219/03d49d67-4c2f-4307-acac-01f93eb7219b

Still very WIP and lots to do.

johnb432:
Awesome! I haven't looked at the code in detail but it looks good.
However, I'm not quite convinced by the specific way the filtering for weapon types is implemented. It works for Vanilla, but how well does it fare with modded weapons? If it fares well, then ignore the rest below.

I used [this](https://github.com/johnb432/Zeus-Additions/blob/740dce862a888983a038887d85ad3df2b57a846a/addons/main/functions/fnc_getRole.sqf#L44) technique to try to get a unit's role, but I know for a fact this is not 100 % reliable. Therefore, I unfortunately can't suggest a better solution, but maybe someone else can?

amsteadrayle:
I'm not fond of the weapon type filtering either, but I've search pretty extensively for another way to categorize weapons. I'm not sure what its real purpose is, but the `cursor` value seems to determine the category in the Equipment Storage menu*. From what I've seen, mods do a pretty good job of using the correct `cursor` value and putting their weapons in those categories.
image

*The vanilla `smg` and `sgun` weapons are in the assault rifle category in this menu. So it's possible that the assault rifle category is actually "assault rifle + other".

LinkIsGrim:
Feature request for a "filters applied" local event.

amsteadrayle:
What should be hooked up to the event?

LinkIsGrim:
Nothing. Just API for 3rd party use.

PR #9588, opened 2023-10-26
Laser - Revert default code to 1111

BroBeansCPG:
- Revert default laser code change in v3.16.0 from 1688 to 1111
- Reverts arbitrary change to _default_ sourced from DCS
- Ensures backwards compatibility with mods/missions that never expected this number to change

If anything, 1111 is more realistic as made aware by US aviators, 1688 isn't a _standard_


jonpas:
@LorenLuke

BaerMitUmlaut:
I found this document: https://www.bits.de/NRANEU/others/jp-doctrine/jp3_09_1%2899%29.pdf
According to it, 1688 is the fixed laser code of the AC-130H. It also mentions that LGB laser codes are set by ground crew and cannot be changed in flight, which I found interesting. Doesn't mention defaults.

I'd rather have a solid source in writing than "I know a guy", regardless of which default is chosen.

BroBeansCPG:
I'd also rather not argue who's source is better, the consensus was there is no default, it's set on a per mission basis. However, if a number was likely to be defaulted to, it'd more likely be 1111 than 1688. 

My main concern is backwards compatibility 


Ironically, I was bored and decided to link some sources that were not of "trust me bro" origin

akpws

You can see here codes are set manually on the ordnance itself, so there's no one number fits all solution. 

https://www.globalsecurity.org/military/library/policy/usmc/mcwp/3-16/appk.pdf

Another source which effectively says lower is better, alluding to no standard, however the importance of codes being briefed. 

another, 
https://info.publicintelligence.net/MTTP-JFIRE.pdf

Again, discussing limitations of some ordnance and air frames, suggesting no standard. 

> The AH-1W can designate codes 1111-1788, but has max effectiveness from 1111-1148

> Paveway II – Laser-guided, free-fall weapon. Laser codes are pre-flight
selectable (code 1511-1788). 

https://irp.fas.org/doddir/dod/jp3_09_3.pdf

> The Hellfire system should use PRF codes in the range of 1111-to-1488 to achieve the highest probability of hit.

BaerMitUmlaut:
I found this drawing from a manual of the A-10:
grafik
As you can see, the laser code is set to 1688. The text does not mention this specific code as far as I can tell though. It could very well be that this is the default code and thus shown.

PabstMirror:
just post full technical manuals so we can settle this debate

LinkIsGrim:
just buy an A10 to settle so we can settle this debate

YannikH:
I just wanted to jump in to back up brobeans from the perspective from the Hatchet H-60 mod, and also the AH-64D mod that the reason this change to 1688 was problematic for us, is that we have custom hellfire implementations and custom cockpit systems that change the laser codes, that currently assume the default is 1111.

I think the above discussion sufficiently proved that neither 1111 nor 1688 are some kind of universal standard laser code when it comes to real world usage of PRFs. The important thing for us is that 1111 is already supported by our mods, and possibly other mods (for example ITC Air Systems, which is still in use by many users, but hasn't been actively maintained for many years, or potentially the F/A-18E/F Super Hornet mod).

I understand that of course every other mod that adds their own functionality with the ACE laser system could fix their code to work with this new standard, and from the perspective of the Hatchet team, we'll definitely make sure in the future to build code more robust so that this kind of change won't cause a problem again, but for now I think it'd be much nicer to just revert to the old (equally realistic or unrealistic) standard of 1111.

jonpas:
Agreed with @YannikH, but would still like to hear @LorenLuke's argument for the change, it is not something I thought of when seeing his PR.

LorenLuke:
> It's not **_really_** a standard frequency

But it's plastered in various places, including in some of the material provided above. If ACE markets itself as a realism-enhancing mod, then any step to approach the real world, even as small as this, is justification itself. 

You could argue that it's all anecdotal and you'd probably be correct. To say that no particular laser code is no more 'correct' than any of the others could easily be true. But the sheer mountain of this evidence (including that you and others have put) shows that this particular code is the primary example that's used and most likely can be considered 'default' of some kind. 


> Lower numbered PRF codes provide higher quality designation due to faster pulse repetition.

> The AH-1W can designate codes 1111-1788, but has max effectiveness from 1111-1148

These statements are probably true for the same reasons.  A higher frequency illumination signal is easier to filter over a short amount of time than a lower frequency one. 


> The Hellfire system should use PRF codes in the range of 1111-to-1488 to achieve the highest probability of hit.

Same point as above, although up to 1688 is cited to work with Hellfires for 'desired probability of hit'.
Pg. 4-62
https://www.globalsecurity.org/jhtml/jframe.html#https://www.globalsecurity.org/military/library/policy/usmc/mcwp/3-23-1/ch4.pdf|||  


> Paveway II – Laser-guided, free-fall weapon. Laser codes are pre-flight selectable (code 1511-1788).

If you're offering this as evidence that 1111 should be default, then I'm reading this as the 1111 laser code shouldn't work with the ingame GBU-12s.  

----

Re: @YannikH 

There's many mods that depend on ACE and with many updates I've seen lots of people ask for reversions or carve-outs or exceptions or alterations to features so as to keep those dependent mods from breaking in some way. Those people are either invited to adapt their code to fit, or welcomed to open a PR. I've never seen a member of the ACE team ever imply other mod's maintenance or adaptation (or lack thereof) was ACE's responsibility. It's an odd thing to me that this appears something different.

The only reason (right or wrong) I believe that this reversion is entertained as an idea at all is because it can be done in three keystrokes. But if literally those 3 tiny keystrokes can break so much in the way of downstream, then I suggest do what's best for most and breaks least.

BroBeansCPG:
> There's many mods that depend on ACE and with many updates I've seen lots of people ask for reversions or carve-outs or exceptions or alterations to features so as to keep those dependent mods from breaking in some way. Those people are either invited to adapt their code to fit, or **welcomed to open a PR**. 

We've already implemented a more robust check on our end, yes our update schedule sucks but we understand that's an us issue, but I'm not asking you to bend backwards or make core alterations to features to ensure compatibility. I've since **opened a PR** in the hopes that fewer mods/missions require updating. 
>I've never seen a member of the ACE team ever imply other mod's maintenance or adaptation (or lack thereof) was ACE's responsibility. It's an odd thing to me that this appears something different.

Backwards compatibility is heavily ingrained in the history of Ace, and fundamentally arma modding. I never once said it was ACE's responsibility, however, considering this number is arbitrary (there is no PRF simulation) I don't think asking it to be reverted to what it was, is out of the question. 


>If you're offering this as evidence that 1111 should be default, then I'm reading this as the 1111 laser code shouldn't work with the ingame GBU-12s.

I think you missed the main jist of my point I tried to make, and that was, there is no one code fits all situations. Looking at the documents we've all shared, the only constant is that numbers change all the time dependent on the air frame, conditions, mission & branch. 


We could go back and forth all day sharing more sources which are largely example illustrations and not entirely fact, at the end of the day, I've submitted a PR, do what you will.

PulsarNeutronStar:
I suggest 1337 as default so both sides are equally annoyed.

BroBeansCPG:
> Should we revert wavelength change as well? 

Thanks, lol, you just found another issue in hatchet with our LST. 

What else is wavelength used to simulate? if anything at all? 

Obviously realism is the goal, but if simulation is absent the number becomes arbitrary and if that is the case. At this stage in the life cycle I guess the question does need to be asked, is it for the better or is it in-fact detrimental.

jonpas:
Thought I posted this already, but apparently not.

The team has decided to revert the laser code with this PR, as we do not care what number it is (after reviewing provided documents and arguments), if keeping 1111 means less old mods broken, that is the way to go.

jonpas (approved):

jonpas:
Ref. 9434

PR #9592, opened 2023-10-26
Configs - Fix case

PabstMirror:
These addons never get linted because of the has_include
probably a good idea to occasionly replace

`#if __has_include\(.*\)` with `#ifdef NEVERDEFINED`
and 
```
[rapify]
enabled = false
```
to true

so hemtt can take a look at them


jonpas:
Could @BrettMayson HEMTT lint them by just skipping any `__has_include`, but keep it unrapified for packing?

LinkIsGrim (approved):

